PR #1544, opened 2017-05-17
More factory refactoring

gravitystorm:
Factory factory factory.

### DIFF
--- a/test/models/way_test.rb
+++ b/test/models/way_test.rb
@@ -13,7 +13,7 @@ class WayTest < ActiveSupport::TestCase
     node = create(:node)
     visible_way = create(:way)
     create(:way_node, :way => visible_way, :node => node)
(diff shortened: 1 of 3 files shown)


test/models/way_test.rb
```diff
@@ -10,11 +10,16 @@ def test_db_count
...
   def test_bbox
-    node = current_nodes(:used_node_1)
-    [:visible_way,
-     :invisible_way,
-     :used_way].each do |way_symbol|
-      way = current_ways(way_symbol)
+    node = create(:node)
+    visible_way = create(:way)
+    create(:way_node, :way => visible_way, :node => node)
+    invisible_way = create(:way)
```
tomhughes: That way doesn't look very invisible to me?

gravitystorm: Good catch!

PR #1547, opened 2017-05-22
Include car profile for GraphHopper routing

karussell:
We now have a fast routing that is able to consider turn restrictions (and costs). And so we would like to add `car` as a further option.

### DIFF
--- a/app/assets/javascripts/index/directions/graphhopper.js
+++ b/app/assets/javascripts/index/directions/graphhopper.js
@@ -1,4 +1,4 @@
-function GraphHopperEngine(id, vehicleType) {
+function GraphHopperEngine(id, vehicleType, disableCH) {
   var GH_INSTR_MAP = {
     "-3": 7, // sharp left
     "-2": 6, // left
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/index/directions/graphhopper.js
```diff
@@ -79,5 +80,6 @@ function GraphHopperEngine(id, vehicleType) {
   };
 }
 
-OSM.Directions.addEngine(new GraphHopperEngine("graphhopper_bicycle", "bike"), true);
-OSM.Directions.addEngine(new GraphHopperEngine("graphhopper_foot", "foot"), true);
+OSM.Directions.addEngine(new GraphHopperEngine("graphhopper_car", "car", true), true);
```
tomhughes: I assume that `disableCH` refers to contraction hierarchies, in which case I would have thought long distance car routing would be exactly the case where you would want them so I'm surprised that you are disabling them for car routing but leaving them enabled for other things?

karussell: We disable contraction hierarchies for car as we currently have not implemented CH with turn restrictions. But disabling CH now falls back to [a still fast enough algorithm](https://github.com/graphhopper/graphhopper/pull/780) that supports turn restrictions. For the others vehicle profiles: there is no reason yet to not use CH.

tomhughes: Well I would have thought that if needed to be disabled for the car profile then that should be something the server did automatically as part of the profile rather than forcing the client to deal with implementation details...

karussell: This is true but we have to enable CH for backward compatibility reasons for now

PR #1550, opened 2017-05-27
Beautify 404 page

comzeradd:
I stumbled upon the 404 page today and I thought I should give it some love.

I guess inline style and no external css on that page is intentional, so I kept this pattern.

### DIFF
--- a/app/assets/stylesheets/errors.scss
+++ b/app/assets/stylesheets/errors.scss
@@ -1,8 +1,6 @@
-.logo {
-  float: left;
-  margin: 10px;
-}
-
-.details {
-  float: left;
+.error {
+  max-width: 100%;
+  width: 500px;
+  margin: 20px auto;
+  text-align: center;
 }
--- a/app/views/errors/forbidden.html.erb
+++ b/app/views/errors/forbidden.html.erb
@@ -1,3 +1,3 @@
-<h1>Forbidden</h1>  
+<h1>Forbidden</h1>
 <p>The operation you requested on the OpenStreetMap server is only available to administrators (HTTP 403)</p>
-<p>Feel free to <a href="http://wiki.openstreetmap.org/wiki/Contact" title="Various contact channels explained">contact</a> the OpenStreetMap community if you have found a broken link / bug. Make a note of the exact URL of your request.</p>
+<p>Feel free to <a href="https://wiki.openstreetmap.org/wiki/Contact" title="Various contact channels explained">contact</a> the OpenStreetMap community if you have found a broken link / bug. Make a note of the exact URL of your request.</p>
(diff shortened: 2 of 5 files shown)


public/404.html
```diff
@@ -2,11 +2,11 @@
    "http://www.w3.org/TR/html4/loose.dtd">
 <html>
 <body>
-  <img src="/assets/osm_logo.png" style="float:left; margin:10px">
-  <div style="float:left;">
-    <h1>File not found</h1>  
-    <p>Couldn't find a file/directory/API operation by that name on the OpenStreetMap server (HTTP 404)</p>
-    <p>Feel free to <a href="http://wiki.openstreetmap.org/wiki/Contact" title="Various contact channels explained">contact</a> the OpenStreetMap community if you have found a broken link / bug. Make a note of the exact URL of your request.</p>
+  <div style="max-width:100%;width:500px;margin:auto;text-align:center;">
+    <h1>Page does not exists :(</h1>
```
iandees: Perhaps use:

> Page does not exist.

tomhughes:
The last thing we need is new graphics on the page, at least until somebody figures out how to fix it properly (which is hard if I recall) because they won't work without manual hacking on the live servers.

Once the page is fixed to be built from a template and use the asset pipeline there will also be no need for any CSS to be inline.

HolgerJeromin:
We could include the svg file directly to the html if we want an image here and don't want to fix the asset problem.

comzeradd:
Yes, I guess on the long-term there should be a proper 404 page including project's css and assets.

I see this as a temporary fix, mainly because the current one is not very helpful for the users. It doesn't have a link to go back to the home page, and it mostly gives the impression of some generic server error rather than a 'wrong link' error.

I included the svg inline so we don't have to include any new asset.

HolgerJeromin:
The svg could be optimized for size (remove meta data) but this is perhaps not so important for this page. 
You can optimize it online 
https://jakearchibald.github.io/svgomg/

tomhughes:
Well it's not like anybody should be hitting it so I'm not sure why it needs to be especially helpful... Do you know of some problem that causes people to see it?

comzeradd:
Thanks @HolgerJeromin. Nice tip!

@tomhughes Some might hit it though. It could even be a case of an old url that is not there anymore. It doesn't hurt to slightly improve the UX for the users ;) Especially when it doesn't mess too much with the current code.

matkoniecz:
> Do you know of some problem that causes people to see it?

For me it is typically a malformed link - typically something like http://www.openstreetmap.org/copyrightwith_attached_text_because_space_was_missed

comzeradd:
I rebased the commit to meet the changes that moved the error templates. I remove the custom graphic, since the same one is used on all error pages.

gravitystorm:
I like in particular the idea that you should be able to click from the error page and get back to the home page. I've extracted this idea into [a separate PR](#2362), which just makes the logo clickable, rather than adding more words, and avoids hardcoding the URL.

There's a few other things that I would suggest changing on this PR, in particular the ASCII smiley, which is unlikely to translate well into other languages, and the CSS too. However I think that would mean there's not much left here to work with. So I'm going to close this in favour of the other PR.

Thanks for the proposals, in particular the concept of linking back to the home page, which I agree would be an improvement.

PR #1556, opened 2017-06-01
Remove fixtures

gravitystorm:
This PR removes the test fixtures, and associated helper methods. It also fixes some tests that were implicitly relying on the test fixtures being in the database.


test/controllers/changeset_controller_test.rb
```diff
@@ -2036,37 +2046,46 @@ def test_list_max_id
   ##
   # This should display the last 20 changesets closed
   def test_feed
+    changeset = create(:changeset, :num_changes => 1)
+    _empty_changeset = create(:changeset, :num_changes => 0)
+
```
tomhughes: I think we should create a changeset that is still open here, to make sure it doesn't get included in the feed result - that was (admittedly very non-obviously) being tested before,

gravitystorm: I think this might be a misleading comment - the controller code doesn't appear to check for open vs closed. In fact the [comment in the code](https://github.com/openstreetmap/openstreetmap-website/blob/4cd2d22788cbc8acf411c9f9436eee85a4590c5f/app/controllers/changeset_controller.rb#L251) suggests that it should only be /open/ changesets, which I think is a second misleading comment!

The changeset in the test here is open anyway, but I can add a closed one to test the behaviour.

Do you think I should update both comments to simply reference "changesets" rather than open or closed?

tomhughes: Ah yes you're right... I misread the code in `check_feed_result` and thought it as filtering out open changesets.

tomhughes: Updating the comments sounds good though as they are very misleading.

gravitystorm: Done in 060230f

PR #1558, opened 2017-06-05
Transform the web site into a fullscreen web app

plarus:
I made the necessary changes to turn OSM.org into fullscreen web app on iOS and android

Sources : 

https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html
https://developer.chrome.com/multidevice/android/installtohomescreen

Issue from https://trac.openstreetmap.org/ticket/5218

### DIFF
--- a/app/assets/favicons/manifest.json.erb
+++ b/app/assets/favicons/manifest.json.erb
@@ -37,5 +37,7 @@
 			"type": "image\/png",
 			"density": "4.0"
 		}
-	]
+	],
+	"start_url": "index.html",
+	"display": "standalone"
 }
--- a/app/assets/favicons/manifest.json.erb
+++ b/app/assets/favicons/manifest.json.erb
@@ -38,6 +38,6 @@
 			"density": "4.0"
 		}
 	],
-	"start_url": "index.html",
-	"display": "standalone"
+	"start_url": "/",
+	"display": "minimal-ui"
 }


tomhughes:
What is a "full screeen web app" exactly? How is it different to visiting the page in a web browser?

tomhughes:
Having scanned the Android link I'm not at all sure this is a good idea as it will, as I understand it, hide all the normal navigation controls and I'm not sure the site would be very usable if that happened?

tomhughes:
Also as far as I can see both Chrome and Firefox on Android offer "install to home screen" options even without any of this.

plarus:
A full screen web app is a website displayed out of the browser (without the top screen browsing bar).

Yes the add to home screen button also exist without this code. But with this code if the website is launched from the home screen (from a previously created link) it is launched in an out of browser full screen mode.

One example : https://www.viamichelin.com

tomhughes:
Right, but that really requires that the web site is designed to be used in that way (ie that it's basically a single page site) which ours isn't.

plarus:
OSM is a single page website for most users => The map page

Multiple pages website is not a problem as explain here : https://developer.chrome.com/multidevice/android/installtohomescreen#work
And OSM is already a mobile compatible website!

The objective is to get more space for the map (especially on smartphones).
Who actually use the browser address bar on smartphones?

tomhughes:
See that's exactly where I thought this was coming from - the idea that the map page is the main page people use.

It's explicitly not our goal to provide an end-user site - that is to say a google maps style site for end users. Our primary audience is intended to be mappers, who will use  much more of the site.

app/views/layouts/_head.html.erb
```diff
@@ -20,6 +20,9 @@
   <%= favicon_link_tag "favicon-16x16.png", :rel => "icon", :sizes => "16x16", :type => "image/png" %>
   <%= tag("link", { :rel => "mask-icon", :href => asset_path("tag-icon.svg"), :color => "#7ebc6f" }) %>
   <%= tag("link", { :rel => "manifest", :href => asset_path("manifest.json") }) %>
+  <%= tag("meta", { :name => "mobile-web-app-capable", :content => "yes" }) %>
```
HolgerJeromin: This is [not used anymore](https://developer.chrome.com/multidevice/android/installtohomescreen) in Chrome 39.

plarus:
I have updated my commit only keeping "manifest.json" updates.
This will only have an impact when saving a site bookmark to a device's home screen.

Web App Manifest fundamentals are described here: https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/

zepatrik:
I have thought of something like this too, maybe there could also be more offline capability with service workers: https://developers.google.com/web/fundamentals/getting-started/primers/service-workers
Posting this here because you are in the subject already but I can open an issue too. Haven't done something like this but I'm interested in this and therefore I maybe can do this.
Goal would be to have some improved map viewer, for both end users and mappers.

HolgerJeromin:
The full version with service worker should be discussed or at least merged seperate. 
They are imo primary for notification when not having the webpage open. For example a news site with a ticker. 
But I am not sure if osm.org needs such logic

zepatrik:
So should I start a separate discussion in an issue? I thought of handling things like cache with a service worker, so opening the website on an device being offline would work with cached tiles.

k-yle:
Is it possible for this change to be merged please? It makes a big difference on cheaper phones with small screens, as the browser uses up a lot of screen space with the URL bar, menu bar, etc., making it harder to use the map. 

It's 2020 now and there are next to no disadvantages of this change. I've now come to expect websites to open in fullscreen when I pin them to my phone's home screen. 

I'd be happy to help if it's more complicated than just merging this pull request (_which also addresses issue #1250_).

Thanks a lot,
Kyle

gravitystorm:
> It's 2020 now and there are next to no disadvantages of this change.

I'd like to check one thing in particular, which is the lack of browser navigation buttons (e.g. the 'back' button). The android link above has a section:

> The CSS display-mode media query indicates how the PWA was launched, either in a browser tab, or as an installed PWA. This makes it possible to apply different styles depending on how the app was launched. For example, always hide the install button and **provide a back button when launched as an installed PWA**.

There's a few parts of our site where the browser back button is necessary, for example on forms that don't (yet) have a cancel  button, or when browsing features like nodes and ways etc. Would enabling PWA mode cause problems in these situations?

HolgerJeromin:
> There's a few parts of our site where the browser back button is necessary, for example on forms that don't (yet) have a cancel button, or when browsing features like nodes and ways etc. Would enabling PWA mode cause problems in these situations?

In my own (home automation) web app I use 
`"display": "minimal-ui"`
https://developer.mozilla.org/en-US/docs/Web/Manifest/display
This is perhaps the better choice compared to suggested 'standalone' for osm.org

app/assets/favicons/manifest.json.erb
```diff
@@ -37,5 +37,7 @@
 			"type": "image\/png",
 			"density": "4.0"
 		}
-	]
+	],
+	"start_url": "index.html",
```
HolgerJeromin, line 41: This should probably be "/" because `https://www.openstreetmap.org/index.html` is no canonical url.

gravitystorm:
I would be happy with "/" and "minimal-ui" as the values for these two settings. @HolgerJeromin could you make a fresh PR for review please? Or @plarus could you update this one?

plarus:
@gravitystorm Updated it as expected

gravitystorm:
Thanks @plarus !

PR #1561, opened 2017-06-11
Update to rails 5.0.x

tomhughes:
Update to rails 5.0.x (currently 5.0.3) and fix up things that have been deprecated or changed as necessary.

There's a test site at https://rails50.apis.dev.openstreetmap.org/ so please shout if you find any problems.


db/structure.sql
```diff
@@ -49,8 +49,7 @@ SET search_path = public, pg_catalog;
 
 CREATE TYPE format_enum AS ENUM (
     'html',
-    'markdown',
-    'text'
```
gravitystorm: If removing the text option from format_enum is deliberate, then I think we need a migration too.

tomhughes: No that's just an accident - that branch is using a different database and apparently it had that value missing for some reason. I've fixed it (and the random field order changes in user) now and also rebased over current master.

tomhughes:
Does anybody have any more questions/issues/comments before I merge this?

PR #1563, opened 2017-06-12
iD 2.2.2 related additions

bhousel:
* Remove traffico from the asset pipeline - this is Mapillary's legacy traffic sign library that is no longer included or used
* Disable `removeUselessStrokeAndFill` svgo plugin (see openstreetmap/iD#4105) - I think it is removing the `#000000` fill on an icon.


bhousel:
Thanks @tomhughes!  I think you might also need to recompile the assets.  Not sure....

tomhughes:
@bhousel what do you mean? assets are precompiled as part of the deployment.

bhousel:
> assets are precompiled as part of the deployment.

I still see the old iD-sprite that's missing the color, but not sure whether that is because it is cached or because rails asset pipeline didn't precompile it (because the sprite itself didn't change, only the options to svgo did)

screenshot 2017-06-12 20 31 47

tomhughes:
Oh it will only recompile things that have changed yes, and there is likely no dependency on the options.

PR #1566, opened 2017-06-23
Rework coordinates to avoid scientific formatting of small numbers

### LINKED ISSUE #1509: Scientific notation in coordinates
In addition to #341 there are other places where coordinates can be shown in XML in scientific notation. These include:

* Changeset bounding boxes
* User home locations
* Tracepoint locations
* Node historical versions (oops, should have done this in #1507 )
* Diary entry rss entries
* Notes

Before I scatter `format("%.7f", foo.lat)` calls everywhere, is there a better approach to solving this problem?

gravitystorm:
This turned out to be slightly more complex than I expected, but it does work using the delegate method. Fixes #1509 

In the json builder, an explicit `.to_f` is required since the builder calls `to_s` on objects that it doesn't recognise, and we would end up with strings instead of floats in the json.

In lib/potlatch, again we need to be explicit, since there's a hardcoded shortcut for built-in types when running `[n].pack("G")`, but adding `.to_f` sorts it out too.

The rest of the changes are in the tests, where testing coordinates as strings needs the corresponding changes.

### DIFF
--- a/lib/geo_record.rb
+++ b/lib/geo_record.rb
@@ -1,4 +1,19 @@
+require "delegate"
+
 module GeoRecord
+  # Ensure that when coordinates are printed that they are always in decimal degrees,
+  # and not e.g. 4.0e-05
+  # Unfortunately you can't extend Numeric classes directly (e.g. `Coord < Float`).
+  class Coord < DelegateClass(Float)
+    def initialize(obj)
+      super(obj)
+    end
+
+    def to_s
+      format("%.7f", self)
+    end
+  end
+
   # This scaling factor is used to convert between the float lat/lon that is
   # returned by the API, and the integer lat/lon equivalent that is stored in
   # the database.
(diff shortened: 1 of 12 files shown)


lib/geo_record.rb
```diff
@@ -1,4 +1,19 @@
+require "delegate"
+
 module GeoRecord
+  # Ensure that when coordinates are printed that they are always in decimal degrees,
+  # and not e.g. 4.0e-05
+  # Unfortunately you can't extend Numeric classes directly (e.g. `Coord < Float`).
+  class Coord < DelegateClass(Float)
```
tomhughes: What's the benefit of using delegation here over just inheriting from `Float` and overriding the `to_s` method?

If we did that then I think we wouldn't need to special case the AMF encoding (because Coord would match Numeric) and we wouldn't need `to_f` on the pack etc?

gravitystorm: Do you mean something like `Coord < Float`? Ruby (or MRI, specifically) doesn't let you do that unfortunately, see e.g. http://www.virtuouscode.com/2011/08/17/you-cant-subclass-integers-in-ruby/ or http://flylib.com/books/en/2.44.1.41/1/  . In short, some built-in classes are special-cased and don't behave like normal ruby classes. The delegate library is there to provide an alternative approach.

tomhughes: Yes that's what I meant, and I guess that explains why ;-)

PR #1568, opened 2017-06-29
Fix layer choice panel scroll issue on mobile

renziver:
Made the layer choice panel scrollable on mobile as mentioned in issue 
#1551 

screen shot 2017-06-29 at 10 05 38 am

Issue #1551, opened 2017-05-28
Cannot scroll layers choice panel on mobile


jidanni:
On mobile, one cannot choose the lowest layers of the list. It's beyond the edge of the screen and it cannot be accessed.

jidanni:
![screenshot_20170528-125632](https://cloud.githubusercontent.com/assets/1230959/26526217/7913be96-4338-11e7-9181-3c55c314f61c.jpg)

jidanni:
Seen in Firefox. This apparently fixed position list cannot be scroll down.

tomhughes:
What screen size triggers this? Only I'm struggling to reproduce it...

jidanni:
Okay, it's working now.

gravitystorm:
I've been able to reproduce this issue, both on Firefox (Ubuntu desktop) and Safari (iOS).

 The problem is when both the `sidebar` (normally on the left in full size version) and the `map-ui` (normally on the right) panels are open at the same time, then the lower `map-ui` panel can't be completely scrolled. This happens with any combination of the two panel types. For example, with your browser narrow enough to trigger the small screen css, open the "New Note" panel (`sidebar`) and the Legend (`map-ui`). The legend panel can be scrolled but not all the way to the bottom. Other combinations (like search results in `sidebar` and layer switcher in `map-ui`) also show the same behaviour. If only one of the two panels is open, scrolling works fine for both panel types.

PR #1569, opened 2017-06-29
Fix scroll layer choice problem on mobile

renziver:
Addressed Issue #1568 
screen shot 2017-06-29 at 10 05 38 am

### DIFF
--- a/app/assets/stylesheets/small.scss
+++ b/app/assets/stylesheets/small.scss
@@ -101,7 +101,7 @@ body.small {
   &.map-layout {
     #sidebar, #map {
       position: relative;
-      overflow-x: hidden;
+      overflow-x: scroll;
       width: 100%;
       height: 50%;
     }
@@ -111,7 +111,7 @@ body.small {
         position: absolute;
         width: 300px;
         height: auto;
-        overflow: hidden;
+        overflow: scroll;
       }
 
       #map {
@@ -122,7 +122,8 @@ body.small {
     #map-ui {
       z-index: 9999;
       width: 100%;
-      overflow-y: scroll;
+      overflow: scroll !important;
+      height:50%;
     }
   }
 
@@ -142,6 +143,7 @@ body.small {
     height: 200px;
     max-height: none;
     min-height: auto;
+    overflow: scroll !important;
   }
 
   /* Rules for the login form */
@@ -169,7 +171,7 @@ body.small {
 
   .user_map {
     width: 100% !important;
-    height: 300px !important;
+    height: 100px !important;
   }
 
   #userinformation .deemphasize {
--- a/app/assets/stylesheets/small.scss
+++ b/app/assets/stylesheets/small.scss
@@ -101,7 +101,7 @@ body.small {
   &.map-layout {
     #sidebar, #map {
       position: relative;
-      overflow-x: scroll;
+      overflow-x: hidden;
       width: 100%;
       height: 50%;
     }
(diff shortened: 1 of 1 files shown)


tomhughes:
There seem to be a lot of changes here which have nothing to do with scrolling the layer switcher?

You're changing the scroll attributes for a number of things which don't seem to be related to the layer switcher and the height of two things, neither of which seems to be related to the layer switcher?

renziver:
@tomhughes tried to fix it again. thank you.

padi:
This addresses #1551 instead of the one referenced in the description.

Since the fix is just a one-liner, you can fix your commit history by using using git's interactive rebase tool: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

padi:
This addresses #1551 instead of the one referenced in the description.

Since the fix is just a one-liner, you can fix your commit history by using using git's interactive rebase tool: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

@renziver

tomhughes:
So @jidanni seems to think this is no longer a problem? Certainly I couldn't reproduce it...

gravitystorm:
I've reopened #1551 since I've been able to reproduce the reported problem.

This PR does fix the problem. However, it does so by making the `map-ui` panel always 50% of screen height, even if it is the only open panel (Currently if the `map-ui` panel is shown alone, it is full height). For reference, the `sidebar` is always 50% height on small screens.

So it fixes the problem described, but with the downside that the `map-ui` panels are always smaller (and so need to be scrolled more often, and options on the layer switcher will be usually hidden). 

Perhaps there is an alternative approach that allows the `map-ui` to remain 100% when shown alone, but 50% when the sidebar is active? I'm not strong enough with CSS to know if that's feasible.

gravitystorm:
> This PR does fix the problem. However, it does so by making the `map-ui` panel always 50% of screen height, even if it is the only open panel (Currently if the `map-ui` panel is shown alone, it is full height). For reference, the `sidebar` is always 50% height on small screens.
> 
> So it fixes the problem described, but with the downside that the `map-ui` panels are always smaller (and so need to be scrolled more often, and options on the layer switcher will be usually hidden).

Given the above mentioned limitations, I'm going to close this PR. Alternative solutions to #1551 are of course welcome.

jidanni:
> So @jidanni seems to think this is no longer a problem? Certainly I couldn't reproduce it...

@jidanni won't get notified, as you have blocked him.

gravitystorm:
> @jidanni won't get notified, as you have blocked him.

Nobody has blocked your notifications. We don't have control over the notifications sent to you, that's entirely between your own Github notification settings, and your junk mail filters.

tomhughes:
Actually he may be right. It's true that I personally have blocked him but I have no idea what effect that has on notifications of comments I make to him.

jidanni:
https://catless.ncl.ac.uk/Risks/31.46#subj35
https://catless.ncl.ac.uk/Risks/31.47#subj13

jidanni:
It's a shame that so many relationships here are exacerbated simply due to GitHub's mistaken implementation of blocking. Making user A think user B is ignoring them when all along they don't get the notification in the first place, etc.. That's why I think OSM should inform its GitHub users of the problem, as GitHub is not going to fix it.

gravitystorm:
@jidanni @tomhughes OK, thanks both for explaining the situation to me, now I understand what's going on.

PR #309, opened 2013-06-18
fixing reload-languages script to work with Rails 3

drewda:
"Railties now deprecates: RAILS_ROOT in favor of Rails.root"

http://edgeguides.rubyonrails.org/3_0_release_notes.html#railties

PR #1574, opened 2017-07-05
Remove unused script

gravitystorm:
Refs #309, where the comments shows that it hasn't been used "for
several years", and that comment was several years ago.

### DIFF
--- a/script/locale/reload-languages
+++ b/script/locale/reload-languages
@@ -1,5 +1,0 @@
-#!/usr/bin/env ruby
-
-require File.dirname(__FILE__) + "/../../config/environment"
-
-Language.load(RAILS_ROOT + "/config/languages.yml")


tomhughes:
Well it hasn't been used, but it probably should have been... Of course `script/locale/update-languages` would need to work so we could update the yaml file, and that currently fails because the page it fetches from at Translatewiki no longer exists.

I have to say my memory was that the language list had come from wikipedia, or the ISO list or something and I don't remember getting it from Translatewiki at all!

Of course really we probably shouldn't have this in the database at all, though I think the diary entries have a foreign key against it.

There are several other bits of ancient cruft in `script/locale` that are even better candidates for cleanup than this for sure.

gravitystorm:
> There are several other bits of ancient cruft in script/locale that are even better candidates for cleanup than this for sure.

I'm sure there are! But I was just going on the discussion in the referenced PR.

PR #97, opened 2012-09-17
Remove the date people accepted Contributor terms on OSM personal page

Gnonthgol:
Cleans up the user page a bit. This is not needed any longer for people who have accepted.

PR #1575, opened 2017-07-05
Hide contributor terms details, unless the user hasn't agreed

gravitystorm:
It's no longer important to show when users agreed to the contributor terms.
For users that haven't agreed, we still show their status on their profile.

Logic was also re-written to avoid lots of `if not` things, and tests added.

Replaces #97


tomhughes:
@simonpoole I suspect you're the person that uses this most, so is this OK by you?

simonpoole:
"important" who knows (as I still diff the accepted/not accepted files every day I get the information in a roundabout way in any case), but it is removing the information completely from the public view for every body else (the API only returns accepted / not accepted).

So, I can live with the change, if anybody else cares ... unlikely.

PR #1268, opened 2016-08-14
Allow reporting of abusive users to moderators or admins

grischard:
Working base for merging #841

Issue #841, opened 2014-11-20
Allow reporting of abusive users to moderators or admins


gileri:
Right now there is no "direct" way to report user who behave incorrectly, apart from the [Spam wiki page](http://wiki.openstreetmap.org/wiki/Spam), which isn't easy to manage.

I tried to start implementing this, but having no prior RoR experience slow me down much, so I figured I should describe here what could be implemented :

A simple "Report" link added next to "Add a friend" on the user profile, leading to page where a logged-in user could select a reason from a drop-down list, and maybe add a comment, to report an abusive user.

The Report model would contain these fields : 
- emitter (User who click "report")
- reported (User reported)
- date
- reason (enum of reasons)
- comment (text)

Possible reasons : 
- diary spam
- changeset spam
- diary comments spam
- changeset comment spam
- GPX traces spam
- abusive private messages
- offensive avatar
  It can be possible to group spam into one reason to simplify.

The deletion of an user (either emitter or reported) must cascade and delete Reports.

A controller/view must also be added in order to allow admins/moderators to view and act on pending reports, grouping them by reported users.

rickmastfan67:
Will it also be possible to report anonymous 'notes' when the notes are completely inappropriate and need to be 'hidden' when just closing them wouldn't do?

emacsen:
Hiding notes is a moderator task, but eventually I'd like to see this functionality expanded to moderators as well as administrators.

grischard:
If someone wants to work on this, some work has already been done:

https://www.openstreetmap.org/user/sbagroy986/diary/35257

https://github.com/sbagroy986/openstreetmap-website

kocio-pl:
@grischard Is it still a proof of concept or there's a chance to have it deployed?

tomhughes:
It's pending review, completion and integration.

It's not a proof of concept, it's an incomplete GSOC project.

pnorman:
If anyone wants to take up completing the GSOC project I rebased it earlier this year to end up with https://github.com/pnorman/openstreetmap-website/tree/moderation_rebased

@woodpeck added a commit with https://github.com/woodpeck/openstreetmap-website/tree/moderation_rebased

openstreetmap-website:
It wasn't incomplete.

It was pending review by you (Tom) and final changes.

The difficulty I had in getting this done is a big reason why I didn't
volunteer for any more OSM related programming projects.

- Serge

kocio-pl:
[...]

PR #1576, opened 2017-07-05
Allow reporting of abusive users to moderators or admins

gravitystorm:
Supersedes PR #1268 , see also issue #841 and [this diary entry](https://www.openstreetmap.org/user/sbagroy986/diary/35257).

This PR builds on the work by @sbagroy986, @zerebubuth, @woodpeck and @grischard as we try to shepherd this GSoC project through review and into production.

I have cleaned up some aspects of this PR, purely to minimise the distractions when comparing diffs and trying to understand how it works. Please avoid nitpicking the code for now - we need to consider bigger questions first. For the avoidance of doubt, this PR not in a fit state for merging!


gravitystorm:
So I've had a look through this PR today for a few hours, and it's nowhere near ready to merge or deploy. It'll need a ton of work, but I'm prepared to tackle it. Help is always welcome!

The first thing I want to establish is whether the basic data model is what we need, since it's quite different from how I was considering tackling the problem. My approach was going to be along the lines of just one table, let's call it "flags", with `object_id, object_type, user_id, comment, timestamps`. The moderators can then be given a list of objects, ordered by the most flags. They can then press "delete" or "ignore" and all the flags are removed from the database.

This PR creates three new objects:
### Issues
You get one of these for every object that has been reported:
```
      t.string :reportable_type, :null => false
      t.integer :reportable_id, :null => false
      t.integer :reported_user_id, :null => false
      t.integer :status
      t.string :issue_type
      t.datetime :resolved_at
      t.integer :resolved_by
      t.datetime :created_at
      t.datetime :updated_at
      t.integer :updated_by
      t.integer :report_count, :default => 0
      t.timestamps :null => false
```
### Reports
These contain the individual reports from users, so each Issue can have many Reports
```
    create_table :reports do |t|
      t.integer :issue_id
      t.integer :reporter_user_id
      t.text :details
      t.datetime :created_at
      t.datetime :updated_at

      t.timestamps :null => false
    end
```
### Issue Comments
Allows moderators/admins to add comments about an issue
```
    create_table :issue_comments do |t|
      t.integer :issue_id
      t.integer :commenter_user_id
      t.text :body
      t.datetime :created_at
      t.boolean :reassign
      t.timestamps :null => false
    end
```

I'm not sure if all of these things are necessary, or over-engineering? Thoughts welcome.

gravitystorm:
The report types is one of the gnarliest bits. This allows the user to report each object for one or more reason (see the final image in https://www.openstreetmap.org/user/sbagroy986/diary/35257 ). The reasons are [listed in the translations](https://github.com/gravitystorm/openstreetmap-website/blob/9046b5ba8d07cd8907685ab8da7f73829cb692c5/config/locales/en.yml#L946-L1023) and are usually one of "spam", "offensive", "threat" or "other", but also range into "Undiscussed Import" and "Mechanical Edit" for changesets. Do we want this? 

The PR supports creating reports and issues for this list of objects: `DiaryEntry DiaryComment User Changeset Note`. Too many? Too few?

simonpoole:
@gravitystorm ChangsetComment should probably be included too. 

There is a more philosophical question about scope though, while I think having just one reporting mechanism is a good thing, stuff like ""Undiscussed Import" and "Mechanical Edit" would seems to be out of scope for operations, and if not present, we should probably have a way of handling such stuff differently.

woodpeck:
Great to see this being dug up by someone who cares. 

"Undiscussed import" and "mechanical edit" are the odd ones out because they refer to the geodata itself and not the "meta" bits. These are the (only) parts in the list where the community is - at least theoretically - able to self-police. No special privileges are required to revert an undiscussed import. Such self-policing is made difficult by the lack of a central, easy-to-monitor issue list. It would be a great improvement for community self-policing if, in addition to writing changeset comments, one could also flag a changeset to appear in a central "directory of discussion-worthy stuff" (and if there was some way to reference a multitude of similar changesets from one discussion thread). That, however, is firmly out of scope for this feature, and external sites exist that allow you to monitor changeset discussion entries which goes some way in that direction.

The other items that can be reported on are items where community self-help is not possible. Still I think that transparency would be good so that people can see that something has already been reported (even if that would perhaps negate the usefulness of "handle cases with most flags first" since people might not add flags once they see something has been reported already). Maybe users should even be able to comment on issues instead of raising a new issue on the same subject? Then again, would be be inventing yet another forum?

The data model *does* look overly complicated, however the idea of being able to comment on something sounds sensible, at least unless we want to force moderators/admins to open a ticket in a separate ticket system for every case where some discussion or fact-finding is necessary.

I wonder if this change also introduces a more fine-grained set of permissions than "admin" and "moderator"? What remedies are available to act on reports - who can remove offensive language from the various items under consideration (or can they just be suppressed altogether)? What kind of "paper trail" would be left by such actions - is it conceivable that we e.g. give the right to remove offensive remarks from changeset comments to 100 people speaking different languages without vetting every one of them like we would if we'd give them full moderator rights?

gravitystorm:
> ChangesetComment should probably be included too.

Good suggestion.

> [Reporting undiscussed imports], however, is firmly out of scope for this feature

I tend to agree. I want to make sure that this focusses on flagging spam that needs intervention from someone with website powers, even if it might be a basis for a more complex system in the future. From a UI perspective it will be a good idea to integrate everything, but that's probably a phase 2 thing.

> Then again, would we be inventing yet another forum?

I would prefer to keep the issues and issue comments as a tool just for moderators/admins for now. It reduces the scope, and can always be changed in the future when we get more experience. I think where the comments will prove useful is when an issue (e.g. diary comment moderation) is reassigned from the moderators to the administrators in order to close a whole user account, for example. The PR includes the concept of reassigning issues from one role to another (but issues aren't assigned to individuals).

> I wonder if this change also introduces a more fine-grained set of permissions than "admin" and "moderator"

Not as yet, still only moderators and admins. It has the idea that issues are assigned by default to a particular role based on the object type (which is reasonable e.g. DiaryEntry -> moderator) and can be subsequently reassigned, but then has some assumptions like people only belong to one or other group, there's only two groups, etc, which will need some refactoring. 

> What remedies are available to act on reports

Nothing, beyond whatever exists already. This PR just adds the reporting mechanism. I suspect further PRs will be required to e.g. allow moderators to edit ChangesetComments or whatever remedies are required.

> What kind of "paper trail" would be left by such actions

Nothing, beyond whatever exists already or is implemented in the future. See also [paper_trail](https://github.com/airblade/paper_trail).

> give the right to remove offensive remarks from changeset comments to 100 people speaking different languages without vetting every one of them like we would if we'd give them full moderator rights

My feeling is to make this a phase 2 concern, until we know how many reports per week we get. But I'll keep an eye open that we might need more user roles (and that might need a proper authorisation framework, like [cancancan](https://github.com/CanCanCommunity/cancancan), but that's even more scope creep!).

grischard:
First of all, thank you Andy for picking this up!

I agree that the current database structure sounds heavily engineered, and would indeed tend to go with just one table like you've suggested. I would add one flag (approved/removed/spam?) if the reports have been acted upon, and one flag to objects to ignore further reports.

Showing the number of reports is maybe not desired. If the reports are a black box, it encourages duplicate reports, which make it easier to prioritise moderator actions, or even [remove objects automatically](https://www.reddit.com/r/AutoModerator/comments/41xzms/possible_to_have_automoderator_remove_posts_that/), which I do on the few subreddits  I moderate on reddit, based on a few criteria like user age or posting history.

Reports that have been acted upon should not be deleted. They are useful to respond to user complaints, and make interesting statistics.

Could the report interface be a modal dialog to preserve user flow? What's the interface like from the moderators' perspective?

The comment should not be mandatory. Spam, for example, requires no further comment.

gravitystorm:
I've started working on the PR, tackling things more or less at random. There's lots to do.

One question I have is regarding refactoring the migrations. Should I add further migrations, e.g. to rename columns? Or since it's not deployed anywhere, is it cleaner if I just combine the existing migrations into one and make my changes there? I'd prefer to have a PR with just one end-result migration but I don't want to break something that I'm not aware of.

tomhughes:
I'd say it's fine to combine them.

gravitystorm:
I've started writing some feature tests, which let me do some more major refactoring. I'm splitting out the issues_controller into three - one for creating reports, one for viewing issues and one for managing issue comments. That makes certain things more logical, like reports#new and issue_comments#delete and so on.

Along with that I'm also fixing more bits and pieces at random when I spot them! I'll need another few stabs at this, then I'll create a task list so that we can monitor progress more easily.

tomhughes:
I'm not sure I really want to try and introduce capybara as part of a big feature like this I have to say. I'd much rather we introduced that as a separate PR rather as a side effect of another project.

That said my memory was that we were blocked on that until we were able to go to rails 5.1?

gravitystorm:
> I'm not sure I really want to try and introduce capybara as part of a big feature like this I have to say. I'd much rather we introduced that as a separate PR rather as a side effect of another project.

Oh sure - I'd introduce it as a separate PR. I just needed some feature tests right now to help work on this PR (and since I'm allergic to writing tests that interrogate forms with complex css selectors and then just do raw posts with potentially different form data, I'm avoiding controller tests for now :-) )

> That said my memory was that we were blocked on that until we were able to go to rails 5.1?

`minitest-rails-capybara` lets me write capybara-based "feature" tests in 5.0, and rails 5.1 includes built-in capybara-based "system" tests. It should just be a case of dropping the gem and changing some boiler plate because the "feature" and "system" tests have the same syntax and behaviour. There's a good chance we'll move to 5.1 before this PR is ready anyway!

gravitystorm:
Making some progress here, but still a lot to do. In no particular order, an incomplete TODO list would be:

* Rework the "issue_type" to be a user_role_enum, and restrict in the model to admin/moderator. This should be set for all issues, and makes it easier for reassigning and searching. The defaults can continue to be driven by the model type
* Find and convert all the hardcoded English strings in the views
* Rework so that users who have both admin and moderator roles can see all of the issues
* Handle "report types" via attributes. These were crammed into the details using string joining and driven by checking the translations for the available list of types.
* Make the issue reassigning work
* Various workflow issues. Currently working through the list of issues and resolving them would be timeconsuming since it involves a lot of clicking through.

gravitystorm:
This is getting close to being ready now. If anyone fancies having a look at the code, the fundamentals are all there and working. My updated todo list is still in a comment above.

grischard:
You've done a lot of work here. Thank you so much.

Does [this](https://github.com/openstreetmap/openstreetmap-website/pull/1576/files#diff-88abeec4ba9bdccfec0d0ef524bcd854R20) mean that once a moderator has marked an issue as resolved, it won't be reopened even if new reports come in? I see elsewhere that issues can be marked as 'ignored' or 'reopened', and yet elsewhere that the reopen event marks the issue as open.

Would you please have screenshots of what it will act like?

gravitystorm:
> You've done a lot of work here. Thank you so much.

You're welcome! It feels a bit like I've accidentally signed up for a mini-GSoC all of my own, but hopefully it'll be worthwhile.

> Does this mean that[...]

Sorry, I should have mentioned more clearly that anything marked "FIXME" in the codebase I intend to resolve before submitting for full review. I've added that to my comment above for clarity.

> Would you please have screenshots of what it will act like?

Sure, I can do that too.

tomhughes:
I have updated the test site at https://moderation.apis.dev.openstreetmap.org/ to be based on this branch instead of the original GSOC code.

mmd-osm:
Just adding a few random thoughts after some testing on the dev instance

### Reporting Changesets

* `Before sending in a report for official action, be sure that:`   --> Not clear what "official action" refers to. Maybe adding the term "Data Working Group" might be helpful here, if that's the group handling those reports.

* `You are unable to fix the problem yourself` --> You are unable to fix the problem yourself, or with the help of your fellow community members. (or something similar).

* `This changeset contains a newbie or an editor error` -> Not sure if this is something a moderator wants to handle. I thought this feature is more geared towards _serious_ issues, not something you could discuss and easily resolve on a mailing list or forum.
Side effect of such a category could be that people start confusing this feature with OSMCha's changeset flagging, and moderators get flooded with such reports.


### Reporting a user

grafik

Flag is very hard to spot at the right border. Could this be possibly moved a bit closed to the user name?


### General remarks

* Other sites like StackOverflow only allow for a certain number of flags to be raised per day (iirc, that's 10). I guess that's in place so that user don't abuse this feature, and moderators won't get overwhelmed with feedback that would be better handled by the community.

* User doesn't get any feedback, if a certain flag was already raised by themselves. Maybe a message like "User/Changeset/Note was already reported by you" could be added.

gravitystorm:
@mmd-osm thanks for your feedback, very useful. As a short response to each of your points:

* The "official action" can definitely be reworded, although referring to "moderators" in general is probably better than DWG specifically
* Good point on the community members
* We definitely need to review the list of reasons for each object, along with whether we want to flag changesets through this mechanism at all.
* I can move the flag for users to be in a more similar position to other reportable things, but I don't want these flags to be too dominant since they will be very rarely used.
* I'll leave flags-per-day as a follow-on, until it's needed
* Preventing multiple reports from the same user is a good idea.

Thanks again for the feedback!

tilmanb:
The "save changes" button on the "report user" screen should be renamed to "send report" or such. "Save changes" is what happens on the server side, but the user really wants to send/register/file a report.

simonpoole:
@gravitystorm small nitpick, the current flag icon is just a bit too small. Making it rather hard to spot with less that perfect eyesight, this is further compounded by the flag being the same colour as the headings it is often next to. I suspect a 10-20% size increase would already be an improvement.

gravitystorm:
@tilmanb @simonpoole thanks both for the feedback!

gravitystorm:
I think I've done about everything that I plan for this pull request, so it's now available for a more thorough review. Note that the recent batch of changes includes a change to the migration, so will need re-running.

In particular, I haven't implemented the following:
* Prevent users from submitting multiple reports
* Automatically close issues when the problem is dealt with

The latter in particular is the most pressing thing. This would mean that if e.g. a diary entry is removed by an admin, then any corresponding issue would be automatically closed. But I don't want to expand the PR any further than it already is, so these things can wait either until this PR is approved or rejected.

All feedback is welcome!

simonpoole:
I just managed to get a server error (500) by reporting a anon note. Unluckily no further information available.

tomhughes:
@simonpoole if @gravitystorm really meant he had edited a migration rather than adding a new one then the live site will be broken until I go and hack the database again.

gravitystorm:
@simonpoole @tomhughes yes, I've edited the migration - so the database on the dev server will need fiddling.

tomhughes:
@gravitystorm can you point at the edit so I know what to change?

gravitystorm:
https://github.com/openstreetmap/openstreetmap-website/pull/1576/commits/80a241f7985fec6ec853e48a25823446b59b7682#diff-a050e5b3cf7f858d5a561369f1274710 - remove the "not null" constraint on `issues.reported_user_id`

tomhughes:
OK that should be fixed now.

simonpoole:
Yes, works now, thanks!

simonpoole:
Colour me more confused than normal, but wasn't it possible to comment on individual changesets and changeset comments before? Somehow this no longer seems to be possible. 
grafik

tilmanb:
@simonpoole Yes it was, and it has been removed: https://github.com/openstreetmap/openstreetmap-website/pull/1576/commits/3006ce9713da119ecbe9e0e1f915df53ff1a898f .
I think users would use this interface all the time to report simple data problems, and I understand the rationale for removing it.

mikelmaron:
How should abusive behavior in changesets comments be handled, if it can't be reported through this mechanism?

tomhughes:
@mikelmaron if you look at the commit that @tilmanb found you'll see it wasn't about reporting abusive behaviour (I don't think any of this is about that at present) but mostly about reporting editing errors which I agree is probably out of scope of this for now.

Mostly we're at the point now where we want to get an MVP out and then worry about gold plating it with extra features.

tomhughes:
Specifically https://github.com/openstreetmap/openstreetmap-website/commit/3006ce9713da119ecbe9e0e1f915df53ff1a898f#diff-3f97cbb8726f3310e0b0c386dbe89e22L36 was the list of reasons for reporting a changset.

mikelmaron:
>  it wasn't about reporting abusive behaviour (I don't think any of this is about that at present)

Yes I looked at the commit, but I'm confused. Name of this ticket is "Allow reporting of abusive users to moderators or admins".

Agree this functionality shouldn't be to report editing errors, but there will be a need to report abusive changeset **comments** (eventually)

> Mostly we're at the point now where we want to get an MVP out and then worry about gold plating it with extra features.

Agreed don't consider this a blocker myself. Good to know it will be taken up eventually

tomhughes:
I'm not making any promises that anything in particular will be taken up, just commenting on general strategy at this point.

Right now we don't really have anybody to report abusive behaviour to and I suspect we're going to need some infrastructure work to support better routing of reports before we're in a position to offer things like that.

mikelmaron:
> Right now we don't really have anybody to report abusive behaviour to 

Any discussion on how to approach this yet? Or even that needs to be taken up? What's the possible timeline for deployment of this feature?

simonpoole:
@tilmanb understood, as said I was just confused by the functionality going away. Outside of that I would agree with @tomhughes that we should get this shipped.

gravitystorm:
> Agree this functionality shouldn't be to report editing errors, but there will be a need to report abusive changeset comments (eventually)

There's definitely scope to widen the list of things that can be reported. I would certainly consider adding changeset comments, note comments and traces in followup PRs, but I'm particularly cautious about expanding the scope until we know that the workflow is OK.

> Any discussion on how to approach this yet? Or even that needs to be taken up?

We'll need to suck-it-and-see, but we might need to add more website moderators, if the list of reports is too big for the existing group of moderators and/or administrators to handle. Currently moderator status on the website is a pretty powerful position since it includes being able to create user_blocks, which I'd prefer to tie more closely to DWG membership status. So if we find we need to add a few dozen new website moderators just to handle notes, diary_entries etc, then we'll need more fine-grained roles and permissions. Some foundational code and refactoring would be needed for that, see e.g. https://github.com/openstreetmap/openstreetmap-website/issues/1626

Of course, it's hard to know whether there will actually be an increase in reports being dealt with, since we currently handle most spam etc through existing out-of-band channels (e.g. using IRC, or via DWG) - this just makes the reporting in-band instead of out-of-band. Like I said, probably best to start with a small list of reportable items and see how it goes.

> What's the possible timeline for deployment of this feature?

* Other developers now review this PR. It's sort of half-reviewed since technically the work I was doing was just improving the existing GSoC PR, with a view to taking care of things I would have asked the student to fix (if I had a time machine).
* I can make any changes requested.
* Gets deployed pretty quickly after that.

tomhughes:
I think if anything "website moderator" will need to be even more powerful than the current moderator.

Whoever is dealing with spam reports for diary entries and user profiles really needs to be able to cancel accounts completely., or at least that is how things work now.

I'll try and have a look at the code this weekend but I suspect the biggest challenges are going to be procedural rather than technical.

mikelmaron:
> cautious about expanding the scope until we know that the workflow is OK.

That makes sense. We do need to be clear about the whatever the current scope happens to be, I was confused myself until finding this in code https://github.com/openstreetmap/openstreetmap-website/pull/1576/files#diff-3f97cbb8726f3310e0b0c386dbe89e22R31. This should be in  human documentation on the website, along with a description of the procedures, and how to get involved in moderating. Perhaps another short static page like https://www.openstreetmap.org/welcome, linked from /help and from relevant places in the reporting workflow. 

I'm happy to take a try at putting that together, if this makes sense.

> We'll need to suck-it-and-see, but we might need to add more website moderators, if the list of reports is too big for the existing group of moderators and/or administrators to handle. ... So if we find we need to add a few dozen new website moderators just to handle notes, diary_entries etc, then we'll need more fine-grained roles and permissions. 

Will there be anything to help track the number and kind of reports, and how they're being handled?

There's also the issue that type of reported issue (spam vs personal/offensive), may need to be handled differently, with different guidance to moderators, and different processes for selecting moderators.

Is there a process currently for granting moderator status that we can build off, from the DWG?

simonpoole:
Just so that it is clear I was thinking of, reporting spam and vandalism, which is typically associated with changesets and should eventually go to the DWG or a variant of the DWG, not on installing the OSM equivalent of Blockwarts.

woodpeck:
@mikelmaron I would very much prefer to soft-launch this so we can correct any issues that we see instead of creating a flurry of documentation about it right away. I think having a "report this" button is self-explanatory enough to start, and we can document the process properly once it has settled. As long as things are not widely documented, it is easier to make changes.

mikelmaron:
Then for soft launch, we should remove reasons other than spam and vandalism from the reportable categories.

screen shot 2018-03-01 at 11 41 15 am

mikelmaron:
Or leave them for soft launch, but have a clear place where we can discuss what's happening with the reports and what the process should be in the future.

mikelmaron:
> All of the problems can be reported already

Where is the guidance on how someone should report a threat in a changeset comment to the DWG? What constitutes a threat, and what process does the DWG do to address it? Why is the DWG responsible for something which is not data, but communication related?

gravitystorm:
> Would [DWG] continue with the same workflow?

> Why is the DWG responsible for something which is not data, but communication related?

With both of the topics I quote above, we're in danger of going significantly off-topic! Certainly worth discussing, but best not in the comments for this pull request.

pnorman:
> > Would [DWG] continue with the same workflow?

> With both of the topics I quote above, we're in danger of going significantly off-topic! Certainly worth discussing, but best not in the comments for this pull request.

I think how we get the data out of the site is within scope. If we're able to have reports generate emails, then we can have go to OTRS and we don't need to worry about it on the website side. If we have can only see current reports through the website, then we need to figure out how we're going to get it out for activity reports.

app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,36 @@
+class ReportsController < ApplicationController
+  layout "site"
+
+  before_action :authorize_web
+  before_action :require_user
+
+  def new
+    if create_new_report_params.present?
```
tomhughes: When would the parameters not be present, and what should happen then? It looks to me like the view would just bomb because `@report` would not exist, so presumably it's not supposed to happen?

app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def create
+    @report = current_user.reports.new(report_params)
+    @report.issue = Issue.find_or_initialize_by(:reportable_id => params[:report][:issue][:reportable_id], :reportable_type => params[:report][:issue][:reportable_type])
+
+    if @report.save
+      @report.issue.save
+      @report.issue.reopen! unless @report.issue.open?
+      redirect_to root_path, :notice => t("issues.create.successful_report")
```
tomhughes: Should we not be trying to remember where the user was when they made a report and then send them back there rather than dumping them back at the home page?

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
+# == Schema Information
+#
+# Table name: issues
+#
+#  id               :integer          not null, primary key
+#  reportable_type  :string           not null
+#  reportable_id    :integer          not null
+#  reported_user_id :integer
```
tomhughes: Shouldn't this be `not null` here? Is it ever possible to have a report that doesn't map to a user?

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
...
+#  status           :integer
```
tomhughes: Why is this an integer rather than an enum? Is this a requirement for using the state machine?

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
...
+#
+# Indexes
+#
+#  index_issues_on_reportable_id_and_reportable_type  (reportable_id,reportable_type)
+#  index_issues_on_reported_user_id                   (reported_user_id)
+#  index_issues_on_updated_by                         (updated_by)
+#
+# Foreign Keys
+#
+#  issues_reported_user_id_fkey  (reported_user_id => users.id) ON DELETE => cascade
```
tomhughes: I'm not sure I like having `ON DELETE CASCADE` on these (and other) foreign keys. None of our existing ones have that and it's quite dangerous.

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
...
+#  issues_updated_by_fkey        (updated_by => users.id) ON DELETE => cascade
```
tomhughes: Shouldn't there be a foreign key constraint on `resolved_by` as well?

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
...
+#
+
+class Issue < ActiveRecord::Base
+  belongs_to :reportable, :polymorphic => true
+  belongs_to :reported_user, :class_name => "User", :foreign_key => :reported_user_id
+  belongs_to :user_updated, :class_name => "User", :foreign_key => :updated_by
```
tomhughes: Should we have a `belongs_to` for `resolved_by` as well?

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  before_validation :set_default_assigned_role
+  before_validation :set_reported_user
+
+  # Check if more statuses are needed
+  enum :status => %w[open ignored resolved]
+
+  scope :with_status, ->(issue_status) { where(:status => statuses[issue_status]) }
+
+  def read_reports
```
tomhughes: Would `resolved_reports` be a better name here? That seems to be what it is actually doing? Same goes for the inverse version of course.

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
...
+    resolved_at.present? ? reports.where("updated_at < ?", resolved_at) : nil
+  end
+
+  def unread_reports
+    resolved_at.present? ? reports.where("updated_at >= ?", resolved_at) : reports
+  end
+
+  include AASM
+  aasm :column => :status, :no_direct_assignment => true do
```
tomhughes: There's probably no right or wrong answer here, but do we really gain anything from pulling in `aasm` here and doing a state machine rather than just defining three fairly simple methods?

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
...
+                         when "Note"
+                           reportable.author
+                         else
+                           reportable.user
+                         end
+  end
+
+  def set_default_assigned_role
+    role = %w[Note].include?(reportable.class.name) ? "moderator" : "administrator"
+    self.assigned_role = role if assigned_role.blank?
```
tomhughes: This seems to be written in quite an obtuse way. I'd be inclined to go for something like:

```
if assigned_role.blank?
  self.assigned_role = case reportable.class.name
                       when "Note" then "moderator"
                       else "administrator"
                       end
end
```

app/models/issue.rb
```diff
@@ -0,0 +1,99 @@
...
+#  resolved_at      :datetime
+#  resolved_by      :integer
+#  updated_by       :integer
+#  reports_count    :integer          default(0)
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#
+# Indexes
+#
+#  index_issues_on_reportable_id_and_reportable_type  (reportable_id,reportable_type)
```
tomhughes: It probably doesn't matter, unless we're going to be looking up issues by type without an id, but I would have put type before id on this index.

app/models/report.rb
```diff
@@ -0,0 +1,40 @@
+# == Schema Information
+#
+# Table name: reports
+#
+#  id         :integer          not null, primary key
+#  issue_id   :integer
```
tomhughes: I assume this should have a `not null` constraint?

app/models/report.rb
```diff
@@ -0,0 +1,40 @@
...
+#  user_id    :integer
```
tomhughes: I assume this should have a `not null` constraint?

app/models/report.rb
```diff
@@ -0,0 +1,40 @@
...
+#  details    :text             not null
+#  category   :string           not null
```
tomhughes: Should this be an enum? The categories are bounded right?

app/models/user.rb
```diff
@@ -73,6 +73,12 @@ class User < ActiveRecord::Base
 
   has_many :roles, :class_name => "UserRole"
 
+  has_many :issues, :class_name => "Issue", :foreign_key => :reported_user_id
+  has_one :issue, :class_name => "Issue", :foreign_key => :updated_by
```
tomhughes: Does this make any sense? It's on `updated_by` so it's linking to issues updated by this user, but is `has_one` which doesn't seem to be right?

app/views/browse/note.html.erb
```diff
@@ -17,6 +17,11 @@
       <br/>
       <%= note_event(@note.status, @note.closed_at, @note_comments.last.author) %>
     <% end %>
+    <% if current_user && current_user != @note.author %>
```
tomhughes: I suspect that this whole block, which is going to be repeated a number of times, should probably be in a helper. I guess it would need to be given the object and the title and could probably figure out everything else from that?

app/views/browse/note.html.erb
```diff
@@ -17,6 +17,11 @@
...
+      <%= link_to new_report_url(reportable_id: @note.id, reportable_type: @note.class.name), :title => t('browse.note.report') do %>
+          &nbsp;&#9872;
```
tomhughes: I really don't like the use of that unicode flag character - is anybody likely to recognise what it means? I don't think it's commonly used elsewhere? There's also the question of how widely it will render correctly...

If we can't come up with anything better then maybe we should just use text like `Report user` etc instead?

app/views/issues/_comments.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="issue-comments">
+  <% comments.each do |comment| %>
+    <div class="comment">
+      <div style="float:left">
+        <%= link_to user_thumbnail(comment.user), :controller => :user, :action =>:view, :display_name => comment.user.display_name %>
+      </div>
+      <b> <%= link_to comment.user.display_name, :controller => :user, :action =>:view, :display_name => comment.user.display_name %> </b> <br/>
+      <%= comment.body %>
```
tomhughes: Shouldn't this be going through the rich text renderer? Either that or we shouldn't be using a `richtext_area` to collect the comments...

app/views/issues/_reports.html.erb
```diff
@@ -0,0 +1,19 @@
+<% reports.each do |report| %>
+  <div class="reports">
```
tomhughes: As this is inside the loop, and hence only contains one report, this seems like the wrong class name to use?

app/views/issues/show.html.erb
```diff
@@ -0,0 +1,64 @@
...
+      <%= render 'reports', reports: @unread_reports %>
+    </div>
+    <% end %>
+    <br/>
+  </div>
+
+  <% if @issue.reported_user %>
+    <div class="related-block">
+      <h3><%= t ".other_issues_against_this_user" %></h3>
+      <div class="unread-reports">
```
tomhughes: I think we could do with some sort of styling here to separate entries in the "other issues" list as currently it's just a mass of link text (which sometimes wraps) and it's not clear where one ends and the next begins.

app/views/issues/index.html.erb
```diff
@@ -0,0 +1,44 @@
+<% content_for :heading do %>
+  <h1><%= t ".title" %></h1>
+<% end %>
+
+<%= form_tag(issues_path, :method => :get) do %>
```
tomhughes: I think this should be a POST form as currently it leads to very ugly URLs which doesn't seem very nice?

Also the form doesn't seem to preserve state so I can't tell what I'm currently filtering on and I have to start from scratch each time if I want to adjust my filter.

app/views/issues/index.html.erb
```diff
@@ -0,0 +1,44 @@
...
+      <td><b><%= t ".reported_user" %></b></td>
+      <td><b><%= t ".last_updated_by" %></b></td>
+      <td><b><%= t ".last_updated_at" %></b></td>
+    </tr>
+  </thead>
+  <tbody>
+    <% @issues.each do |issue| %>
+      <tr>
+        <td><%= t "issues.states.#{issue.status}" %></td>
+        <td><%= link_to t(".reports_count", :count => issue.reports_count), issue %></td>
```
tomhughes: Currently this column can, and does, wrap between the number and the word reports. I think it would be better marked as nowrap to stop that.

app/views/issues/index.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        <td><%= link_to reportable_title(issue.reportable), reportable_url(issue.reportable) %></td>
+        <td><%= link_to issue.reported_user.display_name, :controller => :user, :action => :view, :display_name => issue.reported_user.display_name if issue.reported_user %></td>
+        <td><% if issue.user_updated %> <%= issue.user_updated.display_name %> <% else %> - <% end %></td>
+        <td><%= l(issue.updated_at.to_datetime, :format => :friendly) %></td>
```
tomhughes: I wonder if we should make this a friendly date, and maybe merge it with the previous column as "Last Updated" with entries like "3 minutes ago by fred" which might save some horizontal space?

app/views/issues/_reports.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <div style="float:left">
+      <%= link_to user_thumbnail(report.user), :controller => :user, :action =>:view, :display_name => report.user.display_name %>
+    </div>
+    <%= t(".reported_by_html", :user_name => report.user.display_name, :user_url => url_for(:controller => :user, :action => :view, :display_name => report.user.display_name)) %> <br/>
+    <span class="deemphasize">
+      <%= t(".updated_at", :datetime => l(report.updated_at.to_datetime, :format => :friendly)) %>
+    </span>
+    <br/>
+    <span class="deemphasize">
+      <%= t ".category", category: report.category %>
```
tomhughes: I'm tempted to get rid of this as a separate line and merge into the first line as "Reported as %{category} by %{user}" to save a bit of space and rid of the line with the ugly "Category:" prefix.

app/views/reports/new.html.erb
```diff
@@ -0,0 +1,42 @@
+<% content_for :heading do %>
+  <h1><%= t ".title_html", :link => link_to(reportable_title(@report.issue.reportable), reportable_url(@report.issue.reportable)) %></h1>
+<% end %>
+
+<div class="report-disclaimer">
```
tomhughes: I think this box could do with a bit more padding, especially left and right?

app/views/reports/new.html.erb
```diff
@@ -0,0 +1,42 @@
...
+      <%= issue_form.hidden_field :reportable_type %>
+    <% end %>
+
+    <div class='form-row'>
+      <p><%= t('issues.new.select') %></p>
+      <ul>
+      <% Report.categories_for(@report.issue.reportable).each do |c| %>
+        <li>
+          <%= radio_button :report, :category, c %>
+          <%= label_tag "report_category_#{c}", t("reports.categories.#{@report.issue.reportable.class.name}.#{c}") %> <br/>
```
tomhughes: I'm inclined to say we should run `underscore` on the class name here to make the locale keys more typical?

config/locales/en.yml
```diff
@@ -959,6 +962,106 @@ en:
...
+    states:
+      ignored: Ignored
+      open: Open
+      resolved: Resolved
+  reports:
+    new:
+      title_html: "Report %{link}"
+    categories:
+      DiaryEntry:
+        spam: This Diary Entry is/contains spam
```
tomhughes: I don't think "Diary Entry" should be capitalised here - same goes for diary comments and users. Oddly notes is already lower case, which is inconsistent anyway.

config/locales/en.yml
```diff
@@ -959,6 +962,106 @@ en:
...
+        other: Other
+      User:
+        spam: This User profile is/contains spam
+        offensive: This User profile is obscene/offensive
+        threat: This User profile contains a threat
+        vandal: This User is a vandal
+        other: Other
+      Note:
+        spam: This note is spam
+        vandalism: This note is vandalism
```
tomhughes: Not sure how a note can be vandalism? What exactly is it vandalising?

pnorman:
Because there's a UX question about the flag, I looked at what other sites are doing, both with the initial button, and any followup button or page.

Github
image
image

Facebook
image
image
FB allows multiple selections

Google+
image
image

Twitter
image
image

app/models/issue.rb
gravitystorm: Anonymous notes (and if added in future, anonymous note comments).

config/locales/en.yml
```diff
@@ -959,6 +962,106 @@ en:
...
+      reported_user: Reported User
+      reports_count:
+        one: "1 Report"
+        other: "%{count} Reports"
+      reported_item: Reported Item
+    create:
+      successful_report: Your report has been registered sucessfully
+      provide_details: Please provide the required details
+    update:
+      new_report: Your report been registered sucessfully
```
tilmanb: missing "has"

config/locales/en.yml
tilmanb: Maybe if it's intentionally misleading for non-spam reasons (just "for the lulz" stating "There is a post box here")?

tomhughes: But that would only become vandalism if somebody went ahead and added it to the map without validating it?

app/models/issue.rb
gravitystorm: It's a mechanism for highlighting reports which come in after the issue has been resolved. Presumably it's meant for issues that are being ignored, which might be reported again with further information. "read" and "unread" aren't the best names, something like "reports_created_before_issue_was_resolved" and "reports_created_after_issue_was_resolved" would be more accurate.

I have my doubts as to whether this is even a worthwhile concept, compared to just listing all the reports all the time.

app/models/issue.rb
gravitystorm: I'd like to keep the state machine. We other methods for free, such as issue.open?, and it'll likely come into its own when we get more serious about notifications and other state transitions.

gravitystorm:
Thanks @tomhughes and @tilmanb for your reviews. I've address all the issues raised, except for the following:

* is it possible to use a postgres enum column with aasm? Is it a good idea?
* concept of read_reports and unread_reports - rename or remove?
* helper for report links
* format of report links (flag or something else)
* use post for search form, and preserving state while searching
* more tests

More feedback on any topic is welcome! 

Although please bear in mind that I *really* don't want to do much more on this PR, I'd much rather get it out the door and more easily share the workload on additional features.

app/models/issue.rb
tomhughes: I don't think it's work such crazy names as that certainly so I guess we just leave it if we keep the thing at all.

app/models/issue.rb
tomhughes: Okay, just thought I'd ask the question.

tomhughes:
I've just tried it and it does seem to be possible to use an enum for the status. I converted the status field on issues to a enum in my test database and after I commented out the rails level enum declaration in the model all the tests passed.

gravitystorm:
@tomhughes thanks for investigating!

gravitystorm:
Latest batch of changes are now in - just want to highlight that ee1a863 involved another database change.

I don't have anything left on my todo list, so further reviews welcome.

mmd-osm:
The only thing I found slightly inconvenient during testing was the disconnect between hiding an entry (say a blog post), and setting all related issues to "resolved". Currently, these are two separate steps. Maybe a future enhancement could be to explicitly ask the moderator, if related reports should be automatically set to resolved.

Quite oddly even moderators can't see hidden blog posts anymore. So when checking my list of (open/rejected/resolved) issues, some list entries might show some dead links to already hidden posts. This might be inconvenient, when someone wants to check why a particular entry was hidden in the first place, but can't really access it anymore now.

Otherwise this looks very nice, great work.... let's 🚀

gravitystorm:
@mmd-osm thanks for the review, I really appreciate the time you've taken to try this out.

Resolving issues automatically when the item is "dealt with" (e.g. if a moderator hides a diary entry while viewing it, automatically close the issue, if it exists) is something that would be very good for a follow-on pull request.

You make a good point about moderators / admins being able to view hidden posts, I hadn't thought about that. Again, would be a good follow-on enhancement.

pnorman:
I sent the following email for discussion among the DWG

> The "moderation" feature of the website is nearly complete. As the people with the moderator role and those in charge of data-related disputes, we'll be handling some of the reports. It's important to make sure we test the report handling, as we don't want the feature to roll out and reports not get processed.
> 
> I'm not sure what the best way to do this testing is, and am open to suggestions. The GitHub issue tracking the new feature is https://github.com/openstreetmap/openstreetmap-website/pull/1576. We should be focusing on the side the public doesn't see, since it's what we'll be using. In particular, we should look at the interface between getting a report and action by the DWG.
> 
> Any suggestions on how to review this?

mmd-osm:
> I'm not sure what the best way to do this testing is, and am open to suggestions.

For my local tests I have created a number of accounts to simulate different persona, e.g. some spammer/abusive user, a normal user reporting issues and a moderator. This way you can easily simulate the UI and workflow for different users, and get a better feeling how this all works, and what each user groups'  experience looks like . I didn't extensively test passing issues around from one mod account to another.

When registering new users, a single email account is absolutely sufficient. Simply add a different "+" suffix for each account as usual (e.g. myemail+devspammer@my.org). Also, someone needs to add a moderator flag to one of your account on https://moderation.apis.dev.openstreetmap.org/

SomeoneElseOSM:
@pnorman is there anything that describes the functionality that we're supposed to see here?  

I can login in as "user1", create a diary entry https://moderation.apis.dev.openstreetmap.org/user/user1/diary/5 , report it as a different user (dwg3 FWIW).  I can't see anything for that entry in https://moderation.apis.dev.openstreetmap.org/issues - I'm guessing that that's because it's being dealt with by admins not moderators (as diary spam is now) - is that correct?

gravitystorm:
> I'm guessing that that's because it's being dealt with by admins not moderators (as diary spam is now) - is that correct?

@SomeoneElseOSM yes, that's correct. Currently issues with Notes are assigned to moderators, and everything else to admins. This can of course be changed if required.

test/system/report_diary_comment_test.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    click_on I18n.t("diary_entry.diary_comment.report")
+    assert page.has_content? "Report"
+    assert page.has_content? I18n.t("issues.new.disclaimer.intro")
+
+    choose I18n.t("reports.categories.diary_comment.spam")
+    fill_in "report_details", :with => "This comment is spam"
+    click_on "Create Report"
+
+    assert page.has_content? "Your report has been registered sucessfully"
```
tilmanb: Does the string also need to be translated via I18n.t(...) ?

tomhughes: Not in the tests, no - those always run in English. If it's not translated at the point where it is generated then that would be an issue.

tilmanb: Ok, thanks. Then just out of curiosity - why do line 14 and 24 of the test use the internationalisation?

tomhughes: I suspect they don't need to, but it won't matter.

gravitystorm:
@tomhughes please let me know if you'd like any changes to this PR, or if there's anything that I can do to make the review easier.

SomeoneElseOSM:
I'm currently getting an error 500 trying to log in with "user2" at https://moderation.apis.dev.openstreetmap.org .  "user1" oddly seems fine.  Is that something that needs to be fixed here or elsewhere?

tomhughes:
I've been through everything again and pushed some changes, which are now live on the test site.

Unless anybody spots new problems I plan to merge this soon...

don-vip:
Is there an API if we want to be able to report abusive users from JOSM?

tomhughes:
No.

PR #1583, opened 2017-07-12
Use current_user to represent the currently logged in user

gravitystorm:
This is already used by the oauth plugin, and is a general rails convention.

I've left out the users_controller and the parts of the application_controller, since they will need a lot more scrutiny when changed. 

This helps make the controller code feel more familiar to new users, and is preparatory work for moving to a standard login system at some point in the future.


tomhughes:
There's a whole load of `@user` references in the views as well ;-)

PR #1585, opened 2017-07-14
Fix icons on RTL/Arabic About page

mapmeld:
OSM uses the R2 gem to "flip" CSS for right-to-left languages. I have filed [this bug](https://github.com/mzsanford/R2rb/issues/17) in their repo as well.

On the about page (screenshot attached) only one of the icons is visible for Arabic or other rtl languages. This is happening because R2 "flips" the background-position of the icons to an invalid value. One icon escapes because its background-position does not match the RegEx used by the gem.  Even if this were valid CSS, we would not want to change/flip our sprite coordinates.

My recommendation is to use 0 or possibly ```-0px``` instead of ```0px``` as the x-coordinate here until R2 has some change to avoid sprites.

OSM is already using this syntax (intentionally or not, I don't know) on icon.search which has ```background-position: 0 0;```

screen shot 2017-07-13 at 3 58 57 pm

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -2776,11 +2776,11 @@ input.richtext_title[type="text"] {
     vertical-align: middle;
     background: 40px 40px image-url('about/sprite.png') no-repeat;
 
-    &.local        { background-position: 0px    0px; }
-    &.community    { background-position: 0px  -40px; }
-    &.open         { background-position: 0px  -80px; }
-    &.partners     { background-position: 0px -120px; }
-    &.infringement { background-position: 0px -160px; }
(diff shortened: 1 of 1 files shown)


tomhughes:
I think we should let this be fixed properly upstream rather than using fragile local hacks to try and trick it...

mapmeld:
@tomhughes the gem suggests adding this feature as a more general solution, thoughts?

https://github.com/mzsanford/R2rb/compare/osm_sprite

tomhughes:
Whatever really, so long as it's not just relying on implementation specific details to trick it.

mapmeld:
closing, will replace with PR for updated gem and sprite-tag

PR #1580, opened 2017-07-11
Add FAQ and information about creating a banner image

bhousel:
This tells people how to create a pull request to add a banner image for their event.

PR #1587, opened 2017-07-17
add banner for SotM 2017

althio:
thank you @bhousel for the detailed instructions @ https://github.com/openstreetmap/openstreetmap-website/pull/1580


tomhughes:
That image looks very big, as in much bigger than previous banners?

tomhughes:
Confirmed - previous banners have been roughly square at 350x350 while that is 376x586 so it's both too wide (that column is only 350 pixels wide) and also a lot taller than normal.

Also if you're going to remove banner.yml entries please remove the referenced images as well.

tomhughes:
So it's actually being scaled (because of the column width) to 350x545 currently.

althio:
376 wide was an error, thank you for reviewing.
The image is now natively 350x554.

Please add to https://github.com/openstreetmap/openstreetmap-website/pull/1580 the requirement about recommended height.

althio:
The image is now 350x350.

bhousel:
Nice, sorry for the confusion on image size - I will update #1580 with some more info.

PR #1591, opened 2017-07-21
Update to Leaflet.Locate 0.62.0

c960657:
Leaflet.Locate 0.54.0 is incompatible with Leaflet 1.1.0.

This causes the following error when clicking _Show my location_:
> Uncaught TypeError: t.DomUtil.removeClasses is not a function

Leaflet.Locate 0.62.0 and later includes [the fix](https://github.com/domoritz/leaflet-locatecontrol/commit/690891f93f161653cd6897704f00b16bd5aa7040) for this incompatibility.

### DIFF
--- a/Vendorfile
+++ b/Vendorfile
@@ -31,7 +31,7 @@ folder 'vendor/assets' do
       folder 'img', 'src/img'
     end
 
-    from 'git://github.com/domoritz/leaflet-locatecontrol.git', :tag => 'v0.54.0' do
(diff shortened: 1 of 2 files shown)


c960657:
AFAICT the version of leaflet.locate.js currently in master is identical to the upstream 0.54.0 release. It seems the local changes were removed in f4a4632b45af73f051f846bc01f4d45c58f77f8b.

tomhughes:
Oh you're quite right, looks like I did manage to get rid of it...

PR #1595, opened 2017-07-26
Further refactoring to avoid @user

gravitystorm:
This follows on from #1583 and refactors the remaining controller code, and all of the views, to use the `current_user` approach for accessing the currently logged in user.

The `self.current_user =` feels a bit clumsy, but it's necessary to avoid creating local variables with the same name as the controller method. I'm sure those will disappear when we eventually migrate to a devise or similar.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,9 +7,9 @@ class ApplicationController < ActionController::Base
 
   def authorize_web
     if session[:user]
-      @user = User.where(:id => session[:user]).where("status IN ('active', 'confirmed', 'suspended')").first
+      self.current_user = User.where(:id => session[:user]).where("status IN ('active', 'confirmed', 'suspended')").first
 
-      if @user.status == "suspended"
+      if current_user.status == "suspended"
         session.delete(:user)
         session_expires_automatically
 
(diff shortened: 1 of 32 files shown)


routes.txt
```diff
@@ -0,0 +1,270 @@
+                               Prefix Verb     URI Pattern                                                  Controller#Action
```
tomhughes: Did you intend to add this file?

gravitystorm: D'oh.

gravitystorm: OK, force-updated the PR without that file now.

app/controllers/application_controller.rb
```diff
@@ -47,7 +49,7 @@ def require_user
   end
 
   def require_oauth
-    @oauth = @user.access_token(OAUTH_KEY) if @user && defined? OAUTH_KEY
+    @oauth = @user.access_token(OAUTH_KEY) if current_user && defined? OAUTH_KEY
```
tomhughes: You missed one in this line ;-)

As far as I could tell last night that was the only one in any view or controller that had been missed, which is probably pretty good doing.

gravitystorm: Oh dear, I really thought I'd found them all!

PR #1597, opened 2017-07-27
Complete @user to current_user transition

tomhughes:
This (hopefully) completes the `@user` to `current_user` transition by fixing a few remaining issues and then removing the `@user` variable altogether.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
 
   before_action :fetch_body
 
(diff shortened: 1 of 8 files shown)


gravitystorm:
Thanks! I'd done the helpers stuff on a local branch, but hadn't figured out the missing User.new in user_controller to get the forms working smoothly. But I was doing the same thing of changing the `@user` to something else to flush out the remaining references :-)

I'll give it a while (it's not exactly a high priority) before changing the `@this_user` stuff to `@user`, just in case there's any lingering PRs that might blow up on rebasing.

PR #1614, opened 2017-08-18
Refactor trace api_update to avoid duplicating tags

### LINKED ISSUE #1600: PUT on gpx_file in API does not remove old tags
I wanted to use the API in order to update a gpx_file details.
The following is my PUT command body:
```xml
<osm version='0.6' generator='OpenStreetMap server' copyright='OpenStreetMap and contributors' attribution='http://www.openstreetmap.org/copyright' license='http://opendatacommons.org/licenses/odbl/1-0/'>
      <gpx_file id='2466659' name='2014_02_15_1001_______________20140215_1001.gpx' lat='31.7165' lon='35.4505' visibility='private'>
          <description>2014-02-15 1001 עדינות צוקים__20140215_1001.gpx</description>
          <tag>tag0</tag>
      </gpx_file>
</osm>
```

When using the above on a gpx file that does not have tags everything works as expected but when the using it on a gpx file that already has a tag the previous tag is not removed.

The same effect can be achieved when running this command twice - `tag0` will be added twice.

gravitystorm:
This PR refactors the trace api_update method to work on existing trace objects, rather than creating temporary 'shadow' trace objects. In doing so it refactors the `from_xml` and `from_xml_node` model methods to be instance methods instead of class methods.

Advantages:
* No longer duplicate tags (i.e. fixes #1600)
* makes updating a trace via xml more similar to updating via forms (which don't use shadow objects)

Concerns:
* Doesn't avoid deleting and recreating tag objects when the tags are the same (but neither does the `tagstring=` method, afaict)
* Other classes (Node, Changeset etc) all use the class method approach for `from_xml`, rather than instance methods. I don't know whether they should be refactored too, or whether there's some advantage of the class method approach that I haven't understood.

### DIFF
--- a/app/controllers/trace_controller.rb
+++ b/app/controllers/trace_controller.rb
@@ -263,7 +263,7 @@ class TraceController < ApplicationController
     trace = Trace.visible.find(params[:id])
 
     if trace.user == current_user
-      trace.from_xml(request.raw_post)
+      trace.update_from_xml(request.raw_post)
       trace.save!
 
       head :ok
--- a/app/controllers/trace_controller.rb
+++ b/app/controllers/trace_controller.rb
@@ -263,15 +263,7 @@ class TraceController < ApplicationController
     trace = Trace.visible.find(params[:id])
 
     if trace.user == current_user
-      new_trace = Trace.from_xml(request.raw_post)
-
-      unless new_trace && new_trace.id == trace.id
-        raise OSM::APIBadUserInput.new("The id in the url (#{trace.id}) is not the same as provided in the xml (#{new_trace.id})")
-      end
-
-      trace.description = new_trace.description
-      trace.tags = new_trace.tags
-      trace.visibility = new_trace.visibility
+      trace.from_xml(request.raw_post)
       trace.save!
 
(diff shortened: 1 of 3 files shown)


tomhughes:
Assigning to `tags` is enough to trigger deletion of the old tags I take it?

There is of course a fundamental difference between traces and other objects in that we don't keep multiple versions of traces - so it makes sense to have `from_xml` on a node be a constructor that creates the new object we are going to save that replaces the current version. Though I guess we don't keep history for changesets either.

I certainly think that if we go this route we should change the method name to `update_from_xml` or something so that we don't different models using the same name in different ways.

gravitystorm:
> Assigning to tags is enough to trigger deletion of the old tags I take it?

Yes. From http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many :

"collection=objects :  Replaces the collections content by deleting and adding objects as appropriate."

> I certainly think that if we go this route we should change the method name to update_from_xml or something so that we don't have different models using the same name in different ways.

OK, I can do that.

PR #1632, opened 2017-09-05
Update about text

Klumbumbus:
Change
OpenStreetMap powers map data on thousands of web sites, mobile apps, and hardware devices
to
OpenStreetMap provides geographic data for thousands of maps, routers and other services.

as this is more universal and better fits the today's situation. OpenStreetMap is a base not only for maps but also for routers and other people evaluating the data in any way, e.g. creating statistics etc.

https://wiki.openstreetmap.org/wiki/Main_Page also uses the wording geographic data.

The current wording is used as reference to map for maps only as other use cases are not mentioned at the about page (see discussion at http://www.openstreetmap.org/changeset/51405110).

This PR also fixes two grammar typos.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1195,7 +1195,7 @@ en:
   about_page:
     next: Next
     copyright_html: <span>&copy;</span>OpenStreetMap<br>contributors
-    used_by: "%{name} powers map data on thousands of web sites, mobile apps, and hardware devices"
+    used_by: "%{name} provides geographic data for thousands of maps, routers and other services."
     lede_text: |
       OpenStreetMap is built by a community of mappers that contribute and maintain data
       about roads, trails, cafés, railway stations, and much more, all over the world.


tomhughes:
Well "routers" is certainly not right - if we really wanted to say that then "routing engines" would probably be the correct phrase to use.

But really this seems like a trivial change - neither version looks particularly better than the other to me. To start with I think you're reading "map data" too literally as just meaning pretty pictures rather than the raw data that power all sorts of services.

What do other people think? With the above correction regarding routing is this better than what we have now?

tomhughes:
I'd love to know what the two grammar errors are that you think you've corrected - the only thing I can see is that "powers map data" is quite a clumsy phrase and probably not really correct.

Klumbumbus:
>you're reading "map data" too literally

not me but others
>the two grammar errors

I removed the comma before the word "and" and added a dot at the end.

Klumbumbus:
OK, reading english grammar again it seems it is not wrong to use a comma before the and.

tomhughes:
Oh that's an "oxford comma" and is a matter of great debate - there's no real right or wrong about it ;-)

mmd-osm:
No reason to be shy here, OSM run on **millions** of mobile devices thanks to Maps.Me, OsmAnd and friends.

simonpoole:
Completely off-topic, but @mmd-osm the numbers reported as "downloads" on the various play stores are exactly that, net numbers are not exposed, and if this wasn't a marketing exercise, I would say that we don0t have the slightest idea how many users there are a out there. But as it is a marketing exercise, you could easily use "a 100 million mobile devices" 1/2 :-).

Klumbumbus:
>I think the original version reads better.

If we don't make the change like in the PR, then let's at least change "map data" to "geographic data" or "geodata".
This statement at the about page must no longer be misused as alibi to map for renderers.

gravitystorm:
I support changing the "map data" to "geographic data", but I'd prefer to leave the rest of the sentence alone. The websites, mobile apps etc are an easier end-user concept than abstract technical things like routing engines. I also think that "powers ... data on" is a bit clumsy. As for the Oxford comma - well, the rest of the page has them, so lets leave that topic alone too!

So:

> used_by: "%{name} powers map data on thousands of web sites, mobile apps, and hardware devices"

becomes:

> used_by: "%{name} provides geographic data for thousands of web sites, mobile apps, and hardware devices"

simonpoole:
It is rather silly that the techies are discussing marketing copy @harry-wood does the CWG have a better  suggestion? Maybe drop the number, and simply say ".... all over the world."?

zerebubuth:
@simonpoole, I think it's a great idea to ask for additional suggestions from CWG and others. However, I don't think it's okay to dismiss people's opinions as "silly" by pigeon-holing them as "techies". If you don't like the suggestions, please critique the suggestions rather than the people making them.

harry-wood:
hey! Are you saying I'm not techie? \*shakes fist\* :-)

Lot of different ideas just for this one sentence. I guess it's quite a prominent one, so worth thinking about. I'm afraid I'm just going add to the range of ideas

'map data' vs 'geographic data' I'm slightly preferring map data actually. For me "geographic" conjures thoughts of more broad zoomed out data on climate or population (a lot of stuff we actually *don't* include in OpenStreetMap), while "map data" is actually a reasonable short-hand for describing the kind of data we have, even if it risks people reading it too literally.

I guess we can all agree that "provides data for" is better than "powers data on"

simonpoole:
@zerebubuth I didn't dismiss peoples opinions as "silly" that's negative conjecture on your part, I've even made two suggestions on this discussion myself. What is silly is not asking for input from the experts, which I did.

gravitystorm:
Replaced by #2689

PR #1637, opened 2017-09-09
Point at new OSRM demo server

willwhite:
The OSRM Demo Server is down and we're now using the Mapbox API as the official OSRM demo. This PR switches to point at the Mapbox Directions API, which has nearly an identical API. The only thing we don't support right now is the `hints` parameter, so I removed that in this PR. The access token included here should give plenty of headroom to avoid interruptions in the future.

### DIFF
--- a/app/assets/javascripts/index/directions/osrm.js
+++ b/app/assets/javascripts/index/directions/osrm.js
@@ -1,12 +1,10 @@
 // OSRM car engine
-// Doesn't yet support hints
 
 function OSRMEngine() {
-  var cachedHints = [];
 
   return {
     id: "osrm_car",
-    creditline: '<a href="http://project-osrm.org/" target="_blank">OSRM</a>',
+    creditline: '<a href="https://www.mapbox.com/navigation/" target="_blank">Mapbox</a>',
     draggable: true,
 
     _transformSteps: function(input_steps, line) {
@@ -110,18 +108,11 @@ function OSRMEngine() {
     getRoute: function (points, callback) {
 
       var params = [
+        { name: "access_token": value: "pk.eyJ1Ijoib3NybSIsImEiOiJjajdiOHNna28wcGVhMnFvOTVkOWlpNWUzIn0.7V3lSL-vVZ_RSf8_Pc1Ebw" },
         { name: "overview", value: "false" },
         { name: "geometries", value: "polyline" },
         { name: "steps", value: true }
       ];
-
-
-      if (cachedHints.length === points.length) {
-        params.push({name: "hints", value: cachedHints.join(";")});
-      } else {
-        // invalidate cache
-        cachedHints = [];
-      }
 
       var encoded_coords = points.map(function(p) {
         return p.lng + ',' + p.lat;
@@ -132,10 +123,6 @@ function OSRMEngine() {
       var onResponse = function (data) {
         if (data.code !== 'Ok')
           return callback(true);
-
-        cachedHints = data.waypoints.map(function(wp) {
-          return wp.hint;
-        });
 
         var line = [];
         var transformLeg = function (leg) {
--- a/app/assets/javascripts/index/directions/osrm.js
+++ b/app/assets/javascripts/index/directions/osrm.js
@@ -108,7 +108,7 @@ function OSRMEngine() {
     getRoute: function (points, callback) {
 
       var params = [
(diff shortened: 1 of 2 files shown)


tomhughes:
Do you intended that access token to be shared by all deployments of this code?

We normally make such tokens configurable in `config/application.yml` and then deploy the specific token for the main site via our chef recipes.

willwhite:
Yes, sharing that access token across deployments is ok and hardcoding it make the switchover easier. If you'd rather make it a configurable I'm happy to implement that too.

willwhite:
@tomhughes any strong opinion either way? I think this is ready to merge from my perspective if you are ok with the token being there.

tomhughes:
@willwhite sorry, there aren't any significant technical issues, but I had concerns around switching to what is essentially a commercial service as the default so I've been taking soundings offline around that issue.

mmd-osm:
@tomhughes : would you feel more comfortable to have one of the other two routing engines as a default until this is sorted out? Not sure, if GraphHopper or Mapzen could handle the additional load, so maybe this is not feasible.

Some users on the forums reported that routing on the main page doesn't work anymore since a couple of weeks and that they're somewhat unhappy with the current situation.

tomhughes:
Well those are both commercial as well - the GraphHopper case is confusing because it's the name of both the software and also now of a service built on that software.

mmd-osm:
@tomhughes: I see your point. As all available routers are basically commercial services, we could maybe 
- have no default at all have have the user choose his/her favorite routing engine first. 
The dropdown box would initially show something like "Please choose your routing engine".
- pick one of the 3 engines at random, although this may be a bit confusing.

Both options are not ideal from a UX perspective, though it could restore routing on the main page without giving preference to a single commercial service.

karussell:
Another possibility is to pick an engine randomly.

Another possibility would be to pick ORS as it is a university, but they now offer also commercial services based on their routing engine ;)

> As all available routing engines are basically commercial services, we could maybe

You have to differentiate here: all routing engines are truly open source with an open source license, they just have a commercial hosted service *based* on the open source engine. Of course I see your point but naming the engines commercial services is plain wrong. Also it might be interesting to discuss: if there is an engine without a company in the background - would this be really more sustainable/good for OSM?

> the GraphHopper case is confusing because it's the name of both the software and also now of a service built on that software

Might be, still our commercial solution is the "GraphHopper Directions API".

joostschouppe:
Couldn't we temporarily merge this? Or: isn't the damage to the OSM brand of having a non-functional router on the front page bigger than the damage from giving undue promotion to a company? We can take all the time in the world to argue about the principles, but preferably while the frontpage is functional.

tomhughes:
@karussell My logic is that where the routing service being used is the commercial version, with the same routing rules (and potentially additional third party data) then we should name it as what it is, namely the name of the commercial service.

If it's just using the same software but is otherwise independent then I have no problem using the name of the software, as we have been doing with OSM until now.

karussell:
> @karussell My logic is that where the routing service being used is the commercial version, with the same routing rules

Understood, then indeed "GraphHopper" could be misleading, but "GraphHopper Directions API" might be too long. Maybe "GraphHopper API"?

Komzpa:
Another way to solve it is to change design to show <foot> <cycle> <car> choice, making requests to all three engines and showing them as alternative routes, labeling them as 'provided by OSRM / Valhalla / GraphHopper' in route instructions.

Nakaner:
Does anyone know if Mapbox's routing service uses unmodified OSM data? Or do they use a modified version of OSM as [they use for their rendered maps](https://help.openstreetmap.org/questions/59826/no-more-update-between-openstreetmap-and-strava-maps#59873)?

mmd-osm:
Original demo server was restarted by @TheMarex! :1st_place_medal:  Many thanks!!

See https://lists.openstreetmap.org/pipermail/talk-de/2017-October/114296.html

daniel-j-h:
Small update from the Open Source Routing Machine folks re. demo server:

https://lists.openstreetmap.org/pipermail/osrm-talk/2017-October/001545.html

pnorman:
> Small update from the Open Source Routing Machine folks re. demo server:
> 
> https://lists.openstreetmap.org/pipermail/osrm-talk/2017-October/001545.html

Do you know if the Mapbox freeflow traffic data is open, or if its a closed product?

danpat:
This PR can be closed, no need to merge - `router.project-osrm.org` should be significantly more stable going forward, and will only run on stable OSRM releases.

PR #1651, opened 2017-10-16
Add banner for OSMF Membership Drive

bhousel:
This adds a banner for the current OSMF 2017 membership drive.
cc @mikelmaron 

screenshot 2017-10-16 01 25 09


systemed:
Helvetica pls

PR #1653, opened 2017-10-17
fix locate control cursor

obama:
in chrome 61 the cursor for the locate button for some reason was not the hand, but simple mouse

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -524,6 +524,7 @@ body.compact {
 .control-locate .control-button,
 .control-share .control-button {
   border-radius: 0 0 0 4px;
+  cursor: pointer;
 }
 
 /* Rules for the sidebar and main map area */


tomhughes:
Well "for some reason" isn't very helpful - did you manage to find out the reason?

tomhughes:
Also, what has changed in chrome 61 to make this happen? Though I'm seeing it in Firefox 56 as well, so are you sure it's a change in chrome 61?

obama:
no idea, im just saying what my browser is 
it looks like chrome does not apply the standart rule for <a> elements, but i dont know why

HolgerJeromin:
This is not related to an update of a browser. Even the ancient opera 12 (non chromium based) and firefox do not show the `cursor:pointer;`  Probably no one noticed this till now.

The real difference is a missing `href="#"` in the geolocation `a`-element.

PR #1654, opened 2017-10-18
Add missing href=# in location button

HolgerJeromin:
With this the cursor should render identical to the other buttons. 
It is a proper `a` element where the browser default style is a pointer.

closes #1653


tomhughes:
That only fixes the map on the user profile page - you need the same fix in `index.js` for the main map.

PR #1656, opened 2017-10-21
 allow sorting user blocks by created_at or ends_at

woodpeck:
This pull request makes the list of user blocks sortable by the end time (currently the only choice) or creation time. Sorting by creation time is useful because in practice https://www.openstreetmap.org/user_blocks is crowded out by long-time blocks, and newly created blocks first show up on page 3. To make obvious what happens, the pull request also introduces a new column for the creation time. I have another pull request that does not introduce the extra column, but it was confusing to the user to sort the table by a column that was not visible.


pierzen:
Will this info be accessible through the OSM API 0.6 ?

woodpeck:
I have no use case for retrieving user blocks programmatically but if I ever have, I'll certainly try to make a pull request to add an API call.

tomhughes:
Isn't adding two more columns to this view going to make it break rather badly? It already wraps some columns rather nastily on my laptop with the current columns...

tomhughes:
I'm confused anyway - you seem to be adding two new columns to the heading but only one to the body... Isn't that going to throw everything off?

woodpeck:
I'm only adding one new column, "created at", and allow sorting by that. The "ends at" column has always been there, just that it's not called "ends at", it's called "status". My first attempt at scratching my particular itch was not adding any new columns at all, just making the table heading for "status" toggle between the "ends at" time shown in the status column, and the not-shown "created at" time. This saves screen space, however 100% of the 2 users I showed that to were confused about what it was supposed to do!

Could the wrapping effect that you observe on your laptop perhaps be remedied by changing the rather lengthy "... and after the user has logged in" text in the status column to something like "[newline](login required)"?

tomhughes:
Actually it was the wrapping of "(not revoked)" that really leapt out like a sort thumb:

screenshot from 2017-10-31 18-45-40

But really the main problem (other than me having the Revoke! link making it even wider) is that we're trying to show far too much - this is meant to be a summary screen but we're trying to show every single detail!

What for the example is the point of showing the description, giving that we appear to be truncating it after only a couple of dozen characters so that you can't read anything meaningful anyway?

woodpeck:
Kind of depends what our target audience is. As a DWG person using this "for work" I curse the non-responsiveness of the page and would like the table to grow with the browser window, and I'd show more of the description not less. I have no idea who else uses this page and what for. My hunch is, though, that whoever is using it regularly must be more interested in seeing the newest blocks than in seeing those they've seen a hundred times before so even those users should benefit from the ability to sort. If the "Reason for block" is a problem, we could perhaps move it to a tool tip over the "show" link.

PR #1658, opened 2017-10-22
Avoid using or comparing explicit model ids

gravitystorm:
The code is easier to read using higher-level concepts. This PR isn't comprehensive but it fixes a few obvious cases in the controllers.

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -411,8 +411,8 @@ class UserController < ApplicationController
     if @new_friend
       if request.post?
         friend = Friend.new
-        friend.user_id = current_user.id
-        friend.friend_user_id = @new_friend.id
+        friend.befriender = current_user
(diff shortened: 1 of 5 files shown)


app/controllers/user_controller.rb
```diff
@@ -411,8 +411,8 @@ def make_friend
     if @new_friend
       if request.post?
         friend = Friend.new
-        friend.user_id = current_user.id
-        friend.friend_user_id = @new_friend.id
+        friend.befriender = current_user
+        friend.befriendee = @new_friend
```
HolgerJeromin: Imo here the old names were easier to understand.

gravitystorm: Yep, those aren't ideal but they are already in use elsewhere. I'm open to suggestions for changing them. I also intend to rename the `Friend` model to `Friendship`, which will make things clearer. But none of that is in this PR.

Issue #1570, opened 2017-06-29
Missing feature tests


gravitystorm:
Our test suite is heavily focussed on controller and model tests. Unfortunately this means that very little of the site is actually tested, that is, very few of the tests render the views, or check that links are working and pointing in the right place etc. The controller tests don't render views, and each controller method is directly tested, rather than using the routes.

This means that problems crop up even though all the tests pass - see e.g. https://github.com/openstreetmap/openstreetmap-website/commit/117f0e82260efba3c6862e3aefa4b015b062a109

The advice I've read over the years is to focus less on controller tests, and more on feature tests - where a web browser is used to render the html, click on links etc. I find this works well in my other projects.

I would suggest we do some or all of the following, but of course I'm open to suggestions:
* Add feature tests to catch all the problems in the commit mentioned above
* Ensure that the signup flow works using feature tests
* Consider replacing the non-API, non-XML/RSS controller tests as feature tests

tomhughes:
Adding tests for the things in that commit is on my to do list for this evening ;-) I only pushed it without tests to get it out ASAP.

tomhughes:
It's not true that controller tests don't render views by the way - they absolutely do.

gravitystorm:
> It's not true that controller tests don't render views by the way - they absolutely do.

My mistake. Minitest (and rails, by default) have different setup and naming to what I'm used to with RSpec.

The tests were added in fe1e28b4f45f6aa4bb53dc5163c187c5206b66a7 . I will make sure to add "[system](http://guides.rubyonrails.org/testing.html#system-testing)" tests instead of controller tests in future pull requests (e.g. `visit login_url`, `fill_in :email, :with => user.email` etc). But I won't keep this issue open here given there's no further specific actions required.

tomhughes:
Hmm I thought you were referring to integration tests - system tests (which I wasn't aware of) use capybara and rely on an actual browser of some sort so there needs to be some thought and investigation before we start using those.

We do have a handful of experimental javascript tests that worked in that way although I think they may be broken since the rails 5 upgrade as I had to comment out konacha in the gemfile - those tests aren't run by default.

tomhughes:
Also system tests are new in rails 5.1 and we're only on 5.0 at the moment as composite_primary_keys doesn't support 5.1 yet. So I guess we need to wait on https://github.com/composite-primary-keys/composite_primary_keys/pull/404 before we can consider this.

PR #1664, opened 2017-10-30
Add framework for system tests

tomhughes:
This is a first attempt at setting things up so we can run system tests, as previously discussed in #1570.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,3 +28,4 @@ script:
   - bundle exec rubocop -f fuubar
   - bundle exec rake jshint
   - bundle exec rake test:db
+  - bundle exec rake test:system
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,4 +28,3 @@ script:
   - bundle exec rubocop -f fuubar
   - bundle exec rake jshint
   - bundle exec rake test:db
-  - bundle exec rake test:system
--- a/Gemfile
+++ b/Gemfile
@@ -119,9 +119,10 @@ end
 
 # Needed in development as well so rake can see konacha tasks
 group :development, :test do
+  gem "capybara", "~> 2.13"
   gem "coveralls", :require => false
   gem "factory_bot_rails"
   gem "jshint"
-  #  gem "konacha"
   gem "poltergeist"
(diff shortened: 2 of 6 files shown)


gravitystorm:
I'm definitely pro system tests in general, but I found two issues with this PR to discuss:

* The screenshots appear as base64 output in my terminal (Ubuntu 16.04). From [a related pull request](https://github.com/rails/rails/pull/28185), it seems some rails devs are aware that the inline images aren't supported by certain terminals. Should we disable them here, or report upstream?
* The system tests aren't run by default. The [rails docs confirms this](http://guides.rubyonrails.org/testing.html#system-testing), but doesn't offer any way to change the behaviour. I think it's important that they are run alongside the model, controller and integration tests. What do you think?

Beyond these, all looks good to me.

tomhughes:
I didn't see any screenshots, or rather I recall them being images that I had to view. I've never heard of a terminal that supports images anyway?

I agree about running them, but there doesn't seem to be a good solution as far as I could see - it seems to be entirely deliberate (presumably because system tests are relatively slow) that they aren't run by default. I'll investigate some more though...

tomhughes:
The basic problem is https://github.com/rails/rails/blob/master/railties/lib/rails/test_unit/runner.rb#L44 which, after taking patterns from the command line (defaulting to all) and finding matching tests then explicitly removes the system tests.

So it's very hard to override - you would probably have to monkey patch the `load_tests` method to remove that exclude.

tomhughes:
Ah actually it only excludes if there were no user patterns, so there might be some way...

tomhughes:
Turns out you can append to a rake task, so it was pretty easy to make it run them as a separate step in the task...

gravitystorm:
LGTM. I'll report the base64 screenshot upstream and see what they say, but we can merge this in the meantime.

gravitystorm:
Added as 9e929de53091ab5db24be602546e525a72548dbd and  d581fbfaf12c21145985cbc98f41cd8ac4e0a88e and upstream bug report at https://github.com/rails/rails/issues/31159

Issue #1662, opened 2017-10-28
Using Rails point to my local title server,when I sign up or log in to the server this happens:ActionController::InvalidAuthenticityToken


gleide:
I use opensreetmap to point my local server. I can see my local map in web browser when i type localhost:3000.But When I try to sign up, The error occours:
Parameters: {"utf8"=>"✓", "authenticity_token"=>"+DRjYos1zxLzd5iDC9AZyUX5UqN3lhiUUcLmHmU5FAy2r6TRQoarVoC2ci/Fe4w9p/u9JA4x05BwRlhAJCPatA==", "user"=>{"email"=>"yuanyahui11@gmail.com", "email_confirmation"=>"yuanyahui11@gmail.com", "display_name"=>"yeager", "auth_provider"=>"", "pass_crypt"=>"[FILTERED]", "pass_crypt_confirmation"=>"[FILTERED]"}, "commit"=>"Sign Up"}
Can't verify CSRF token authenticity.
How can I sovle this,Hope can get help .Thanks!

tomhughes:
Sorry, but this is not a problem in our code so you will need to debug your installation and figure out why it isn't working.

tomhughes:
At a guess I'd say there's some problem with your session storage so the authenticity token is not being retained by the server, but that is only a guess and there's no way for us to debug this from the error message - you'll need to get your hands dirty and figure it out.

drewda:
@gleide I've encountered the same issue recently when setting up a fresh copy of the Rails app locally. Here are two solutions that you may find useful:

- Create a blank file at `tmp/caching-dev.txt`. This will turn on a local in-memory cache store, which will be used to store your user session data.
- Install a local [memcache](https://memcached.org/) server. Set `memcache_servers: [127.0.0.1]` in your `application.yml`.

@tomhughes I've proposed a small improvement that may guard against issues like this in the future. Here it is, in case you and other maintainers think it may be useful: https://github.com/openstreetmap/openstreetmap-website/pull/1666

gleide:
Thanks very much! It perfectly solved my problem!

drewda:
@gleide you're welcome. Have fun with your local OSM app!

PR #1666, opened 2017-11-02
when no cache store is set, use cookies for sessions

drewda:
This will help provide "training wheels" to keep new users from situations as reported in https://github.com/openstreetmap/openstreetmap-website/issues/1662

I've tested all three conditions and found them to work -- at least locally in the Rails `development` environment on a Mac.


tomhughes:
I suspect that will cause more problems than it solves - is the cookie store actually able to cope with our sessions given the size limits that typically apply to it?

drewda:
I've tested registration flow, sign in, and sign out against the cookie store. Everything stored by those actions appears to fit with the ~4Kb storage limit. Not sure how other uses of session store behave.

Another alternative is to just throw an exception or log a warning when neither Memcache nor Rails cache store are available. Do you have a preference?

drewda:
👍 thanks, @tomhughes.

PR #1668, opened 2017-11-03
Update site to use jQuery 3

tomhughes:
I did't realise that jquery-rails offers all three major versions of jQuery and you have to choose which one you want, so we've been sat on 1.x for a long time...

I've tried to test as much as I can but it wouldn't hurt if other people played with it before we merge it - there's a test site deploying now at https://tomh.apis.dev.openstreetmap.org/.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,4 +1,4 @@
-//= require jquery
+//= require jquery3
 //= require jquery_ujs
 //= require jquery.timers
 //= require jquery.cookie


HolgerJeromin:
This "probably" drop support ie 6-8 which was fully supported by jquery 1
https://jquery.com/browser-support/

I have no idea how many are using this old versions with this page. Perhaps browsing the map should be working. (i can test that in a few days)

tomhughes:
Well IE8 is next to nothing (just 8 visits yesterday) and anything earlier than that doesn't even register and is likely already broken anyway.

There is an underlying motivation here (which is what led to me discovering this) which is that I'd like to move ahead with deploying a content security policy but jquery 1 triggers alerts with some browsers using our proposed policy.

PR #1670, opened 2017-11-05
Remove Potlatch 1 note from README.md

kocio-pl:
Just a small fix - Potlatch 1 editor is not available on website now.


tomhughes:
If potlatch 1 isn't available then that's a bug (as far as I know it's also not true).

It's not on the menu anymore but it's still present if you have the URL.

Issue #1678, opened 2017-11-15
Add 'swap from and to' to directions


jidanni:
http://www.openstreetmap.org/directions?engine=osrm_car&route=24.16453%2C120.86340%3B24.16506%2C120.86437#m
is great. But they forgot a most basic functionality: swap from and to.

jidanni:
Using the mouse to cut and paste one into the other... because of some address expansion engine working, results in what was 115 meters becoming 48 meters.
I.e., one cannot do a clean swap via other methods easily.

jidanni:
The user would probably have to edit the string seen in the URL bar, avoiding the %2C's etc.

jguthrie100:
I took the opportunity to add a pull request @ #1693

jidanni:
@jguthrie100 Thanks.

tomhughes:
This was fixed in #1693.

PR #1693, opened 2017-12-01
Added 'Reverse Directions' link

jguthrie100:
Added a text link that reverses the directions of the planned route

See issue #1678

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -32,6 +32,8 @@ OSM.Directions = function (map) {
   function Endpoint(input, iconUrl) {
     var endpoint = {};
 
+    endpoint.input = input;
+
     endpoint.marker = L.marker([0, 0], {
       icon: L.icon({
         iconUrl: iconUrl,
@@ -110,6 +112,20 @@ OSM.Directions = function (map) {
 
     return endpoint;
   }
+
+  $(".directions_form .reverse_directions").on("click", function() {
+    var input_from = endpoints[0].input.val();
+    var input_to = endpoints[1].input.val();
+    var latlng_from = endpoints[0].latlng;
+    var latlng_to = endpoints[1].latlng;
+
+    endpoints[0].setLatLng(latlng_to);
+    endpoints[1].setLatLng(latlng_from);
+    endpoints[0].input.val(input_to);
+    endpoints[1].input.val(input_from);
+
+    getRoute();
+  });
 
   $(".directions_form .close").on("click", function(e) {
     e.preventDefault();
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1002,6 +1002,11 @@ header .search_forms,
       vertical-align: middle;
     }
   }
+
+  a.reverse_directions {
+    cursor: pointer;
+    margin: 0px 0px 5px 25px;
+  }
 }
 
 /* Rules for the map key which appears in the popout sidebar */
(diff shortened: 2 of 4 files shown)


jguthrie100:
Its very basic from a design principle, but maybe someone else wants to amend it to make the design a bit better (maybe use an arrows.png image like google maps does, instead of just having a basic 'Reverse Directions' text link)

Added 'Reverse Directions' to the en.yml locale which will need translating

jguthrie100:
image

app/assets/javascripts/index/directions.js
```diff
@@ -111,6 +113,20 @@ OSM.Directions = function (map) {
...
+    var latlng_from = endpoints[0].latlng;
+    var latlng_to = endpoints[1].latlng;
+
+    endpoints[0].setLatLng(latlng_to);
+    endpoints[1].setLatLng(latlng_from);
+    endpoints[0].input.val(input_to);
+    endpoints[1].input.val(input_from);
+
+    $(".directions_form .routing_go").trigger("click");
+  });
```
tomhughes: I think it would be better to do this in the same way we do it in other places, along these lines:

```
OSM.router.route("/directions?" + querystring.stringify({
  from: $("#route_to").val(),
  to: $("#route_from").val()
}));
```

jguthrie100: I originally did it that way, but it doesn't work when you paste a link with just coords in the to and from inputs.

It also jumps the marker to the nearest feature, rather than keeping it pin pointed in the exact coordinate (which leads to the problem OP mentioned about the true coord location not being retained). I'm not sure if it's just standard OSM behaviour to jump the marker to the nearest feature though.

jguthrie100: Basically .val() forces through an Geolocation API call which can mess with the actual user-submitted value or coord

tomhughes: What about if you include a `route` parameter in the URL with the actual lat/lon values? I think in that case it will use those rather than rerunning the geocode? So something like:

```
OSM.router.route("/directions?" + querystring.stringify({
  from: $("#route_to").val(),
  to: $("#route_from").val(),
  route: from_lat + "," + from_lon + ";" + to_lat + "," + to_lon
}));
```

although I haven't looked at how to get hold of the lat/lon values yet.

jguthrie100: I've recommitted with some changes.

The OSM.router.route method fails to work in the example given in the linked Issue (I think its because the latlng values of each marker don't get set when linking straight from the URL, rather than after clicking the "Go" button)

Instead of making things too complicated, I've made a simple call to getRoute(), which handles the setting of all latlng values and includes a call to OSM.router.route.

It now works with the example in the linked Issue as well as normal situations when clicking the "Go" button

jguthrie100:
I've recommitted with some changes.

The OSM.router.route method fails to work in the example given in the linked Issue (I think its because the latlng values of each marker don't get set when linking straight from the URL, rather than after clicking the "Go" button)

Instead of making things too complicated, I've made a simple call to getRoute(), which handles the setting of all latlng values and includes a call to OSM.router.route.

It now works with the example in the linked Issue as well as normal situations when clicking the "Go" button

tomhughes:
I think I left it because I wanted to investigate whether my suggested approach could be made to work.

I've merged this now and applied aea57cdfc321edd68467e0b17bc5335f84a512b9 on top to make it go through the router.

zerebubuth:
Great, thanks!

mmd-osm:
Somehow this doesn't work here. When I click on "Reverse Directions", it just shows the same route again, without swapping start and end.

jguthrie100:
I have the same problem. The addresses in the start and end fields also get switched to coordinates..
@tomhughes @zerebubuth

jguthrie100:
Not sure whats going on because it works when running locally, just not on the live site..

PR #1698, opened 2017-12-06
Adding lxml2 related hint to MacOS installation section

mvexel:
On High Sierra, `libxml-ruby` has trouble finding `lxml2` headers when `lxml2` was installed using homebrew without the config binary. The internet has some steps to redo the installation so that everything works OK.


tomhughes:
I can't comment on the accuracy of Mac installation instructions but I'm not sure I like the idea linking to random gist comments like that.

mvexel:
Fair enough. How about adding explicit installation instructions for libxml2 instead?

gravitystorm:
@mvexel Thanks! I don't have access to any Macs (nor do I know anything about best practises for things like homebrew) so thanks for helping improve our docs.

PR #1701, opened 2017-12-08
Revoking administrator role on current user should fail

### LINKED ISSUE #1697: Granting / revoking roles for own user
I wonder if we really want to allow a user to grant and revoke roles on his/her own user. My suggesting would be to disallow such an operation. Maybe this should be further restricted to the administrator role?

mmd-osm:
See https://github.com/openstreetmap/openstreetmap-website/issues/1697 for further discussion:

> maybe stop them removing their own administrator role

That's exactly what I wanted to accomplish. As there's no additional confirmation popup, I accidentally locked myself out of being local admin, and had to head for the rails console.

Second attempt is now restricted to the *administrator* role. Code style and naming conventions may not be ideal - will happily apply suggestions for improvement.

Change introduced in this pull request:

grafik

Hit orange star (revoke administrator role):

grafik

Action rejected, user is still admin

### DIFF
--- a/app/controllers/user_roles_controller.rb
+++ b/app/controllers/user_roles_controller.rb
@@ -8,7 +8,6 @@ class UserRolesController < ApplicationController
   before_action :require_valid_role
   before_action :not_in_role, :only => [:grant]
   before_action :in_role, :only => [:revoke]
-  before_action :not_revoke_admin_current_user
 
   def grant
     @this_user.roles.create(:role => @role, :granter => current_user)
@@ -16,7 +15,12 @@ class UserRolesController < ApplicationController
   end
 
   def revoke
-    UserRole.where(:user_id => @this_user.id, :role => @role).delete_all
+    # checks that administrator role is not revoked from current user
+    if current_user == @this_user && @role == "administrator"
+      flash[:error] = t("user_role.filter.not_revoke_admin_current_user")
+    else
+      UserRole.where(:user_id => @this_user.id, :role => @role).delete_all
+    end
     redirect_to :controller => "user", :action => "view", :display_name => @this_user.display_name
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/user_roles_controller.rb
```diff
@@ -8,6 +8,7 @@ class UserRolesController < ApplicationController
   before_action :require_valid_role
   before_action :not_in_role, :only => [:grant]
   before_action :in_role, :only => [:revoke]
+  before_action :not_revoke_admin_current_user
```
tomhughes: This should only be applied to the `revoke` method but in fact this is so specific that I don't think there's any point making it a filter - the point of filters is that they can be shared across multiple methods but these even has the method name in it's name so clearly it only ever makes sense to apply it to revoke.

As such I would just put the logic in the `revoke` method and get rid of the filter.

test/controllers/user_roles_controller_test.rb
```diff
@@ -134,5 +134,9 @@ def test_revoke
     end
     assert_redirected_to user_path(target_user.display_name)
     assert_equal "The string `no_such_role' is not a valid role.", flash[:error]
+
+    # Revoking administrator role from current user should fail
+    post :revoke, :params => { :display_name => administrator_user.display_name, :role => "administrator" }
+    assert_redirected_to user_path(administrator_user.display_name)
```
tomhughes: Can we assert that the flash message is shown here... I think there some examples of doing it in other tests.

app/controllers/user_roles_controller.rb
mmd-osm: Good point, I moved the check to the `revoke` method now.

test/controllers/user_roles_controller_test.rb
mmd-osm: I added the flash msg assert as well now.

mmd-osm: Travis CI reports some spurious issue, which doesn't appear related to this PR in any way. Local test was ok, though (with a different seed value).

tomhughes: If a test with a different seed value is ok it means there is some dependency between tests so that they only work when run in a certain order.

tomhughes: Actually the problem is that test tries to fetch an arbitrary node ID expecting it not to exist so fails if a previous test happens to have created it.

PR #1704, opened 2017-12-16
Allow (un)subscribing discussions also when changeset still open

### LINKED ISSUE #1627: unable to watch changesets that are not yet closed but there is a "watch" button
Noticed on https://www.openstreetmap.org/changeset/51641815? (unclosed at this moment)

Comment box is missing (I guess that it is intentional, though there is no explanation why comments are impossible), I see "watch" button that may be clicked resulting i no change - after reload it still "watch" button, not "stop watching button".

Initial:
selection_003

after clicking "Obserwuj" (watch in Polish)
selection_004

after reload
selection_005

mmd-osm:
Fixes #1627 

There’s one rubocop max class length message left in the Travis CI build, which i have no idea how to address properly. Supposedly, I need to run `rubocop --auto-gen-config`, but this creates a new file with many unrelated "fixes".

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -82,7 +82,7 @@ Metrics/BlockNesting:
 # Offense count: 63
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 1796
+  Max: 1797
 
 # Offense count: 71
 Metrics/CyclomaticComplexity:
--- a/app/controllers/changeset_controller.rb
+++ b/app/controllers/changeset_controller.rb
@@ -357,7 +357,6 @@ class ChangesetController < ApplicationController
 
     # Find the changeset and check it is valid
     changeset = Changeset.find(id)
-    raise OSM::APIChangesetNotYetClosedError, changeset if changeset.is_open?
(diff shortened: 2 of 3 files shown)


tomhughes:
Before we consider this does anybody know *why* the current behaviour is what it is? Presumably there was a good reason?

tomhughes:
BTW to fix the rubocop warning just edit `.rubocop_todo.yml` and increase the value.

mmd-osm:
> Before we consider this does anybody know why the current behaviour is what it is? 

That's a really good question. Originally, my idea was to remove the "Subscribe" button on the UI. However, @pnorman's comment on https://github.com/openstreetmap/openstreetmap-website/issues/1627#issuecomment-350596170 makes sense after all. So we keep the button and remove the check for subscribe/unsubscribe in this PR instead.

That behaviour has already been introduced in the very first commit: https://github.com/openstreetmap/openstreetmap-website/pull/772 and never changed since then. It wasn't documented on the API 0.6 wiki page (until I added it last week, will remove it again once this is merged). In addition, the UI never matched the backend, see https://github.com/openstreetmap/openstreetmap-website/issues/1627#issuecomment-350635196

PR #1705, opened 2017-12-16
Discussions on open changesets

### LINKED ISSUE #845: better feedback for comments with an open changeset
Accidentally i uploaded a lot of temp data. I corrected all of them and had luck, the changeset was still open so i could update the description.
But it would be nice if i could comment a changeset, in case someone come across this in the future.

mmd-osm:
This pull request adds an explanation why the comment box is not visible on open changesets, see https://github.com/openstreetmap/openstreetmap-website/issues/1627#issue-254635501

### Changeset still open

grafik

--> additional comment added

--- 

Fixes #845


tomhughes:
Basic idea is fine but I'm not sure I like the wording - not sure what it should be though. Maybe something like:

> Changeset still open - discussion will open once the changeset is closed.

But maybe somebody has a better suggestion.

PR #1723, opened 2018-01-25
Add links to the trademark policy

simonpoole:
The trademark policy has now been approved by the OSMF board, link to it 

- point to the TM policy instead of the LWG on the Copyright page
- add short text on TMs to the About page


simonpoole:
Thanks @tomhughes

PR #1731, opened 2018-02-04
Add marker at trace start coordinate

### LINKED ISSUE #1676: Add marker at trace start coordinate
1. Upload a fresh trace.
2. Now on that trace,
```
Start coordinate:	24.1804; 120.864 (map / edit)
```
click "map".
In such cases wouldn't it be best to put a marker on the map at the start coordinate?

biswesh456:
Fixes #1676 

I have added mlat and mlon parameters in map link of trace. It now marks the starting coordinate of the trace.

### DIFF
--- a/app/views/trace/_trace.html.erb
+++ b/app/views/trace/_trace.html.erb
@@ -16,7 +16,7 @@
       <% end %>
       ... <%= t'trace.trace.ago', :time_in_words_ago => time_ago_in_words(trace.timestamp) %></span>
       <%= link_to t('trace.trace.more'), {:controller => 'trace', :action => 'view', :display_name => trace.user.display_name, :id => trace.id}, {:title => t('trace.trace.trace_details')} %> /
-      <%= link_to_if trace.inserted?, t('trace.trace.map'), {:controller => 'site', :action => 'index', :anchor => "map=14/#{trace.latitude}/#{trace.longitude}"}, {:title => t('trace.trace.view_map')} %> /
+      <%= link_to_if trace.inserted?, t('trace.trace.map'), {:controller => 'site', :action => 'index', :mlat => "#{trace.latitude}", :mlon => "#{trace.longitude}", :anchor => "map=14/#{trace.latitude}/#{trace.longitude}"}, {:title => t('trace.trace.view_map')} %> /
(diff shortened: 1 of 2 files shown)


tomhughes:
The first chunk looks like it is used for a trace that hasn't been loaded yet (where `trace.inserted` is false) and in that case the trace won't have a known start position yet so does that cause a problem at all?

zerebubuth:
It looks like [`link_to_if`](https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to_if) isn't going to evaluate the link when `trace.inserted?` is false. I thnk that means that the trace will have non-nil lat/lon whenever `trace.inserted?` is true, so it shouldn't cause a problem?

There are some [other places in the code](https://github.com/openstreetmap/openstreetmap-website/blob/609cedc4998805d38d1607e68c67e491ce589114/app/models/trace.rb#L182-L183) where we're assuming `inserted?` implies lat/lon are present.

tomhughes:
Ah I missed it was `link_to_if` there. That's fine then - if it has inserted successfully there should always be a start point.

app/views/trace/_trace.html.erb
```diff
@@ -16,7 +16,7 @@
       <% end %>
       ... <%= t'trace.trace.ago', :time_in_words_ago => time_ago_in_words(trace.timestamp) %></span>
       <%= link_to t('trace.trace.more'), {:controller => 'trace', :action => 'view', :display_name => trace.user.display_name, :id => trace.id}, {:title => t('trace.trace.trace_details')} %> /
-      <%= link_to_if trace.inserted?, t('trace.trace.map'), {:controller => 'site', :action => 'index', :anchor => "map=14/#{trace.latitude}/#{trace.longitude}"}, {:title => t('trace.trace.view_map')} %> /
+      <%= link_to_if trace.inserted?, t('trace.trace.map'), {:controller => 'site', :action => 'index', :mlat => "#{trace.latitude}", :mlon => "#{trace.longitude}", :anchor => "map=14/#{trace.latitude}/#{trace.longitude}"}, {:title => t('trace.trace.view_map')} %> /
```
tomhughes: I've fixed it for this change but for future reference there's no point doing string substitution for a single value like that - it will just slow things down. Just use `:mlat => trace.latitude` etc instead.

PR #1732, opened 2018-02-04
Remove npemap as a source for UK postcodes

tomhughes:
As discussed on the mailing list the other week there probably isn't much point continuing to use npemap as a geocoder backend for UK postcodes given that nominatim consults codepoint open data anyway.


Firefishy:
+1 sounds good.

PR #1734, opened 2018-02-11
Allow admins and moderators to delete traces

### LINKED ISSUE #1625: Allow admins and/or moderators to delete traces
At the moment traces can only be deleted by the user that uploaded them. It would be useful to allow admins and/or moderators to do the same.

biswesh456:
Fixes #1625


gravitystorm:
This PR shows the link, but it won't actually work. In trace_controller.rb#delete method, only the trace owner can delete the trace.

It is best to create a test (in test/controllers/trace_controller_test.rb) that logs in as an admin and tries to delete the trace. That would highlight the problem that I mention.

biswesh456:
@gravitystorm Extremely sorry that I didn't actually delete it to check if it worked or not. This time I created an admin and checked if the link works or not. It worked for me. Is it fine now?

gravitystorm:
The code looks OK, but you still need to write the tests. That way we make sure that the code keeps working, even when we do updates and refactoring in the future.

biswesh456:
I have trying to add the test case but it keeps on giving me this error:

Failure:
TraceControllerTest#test_delete [/Users/bisweshmohapatra/Desktop/gitHub/openstreetmap-website/test/controllers/trace_controller_test.rb:692]:

Expected response to be a <3XX: redirect>, but was a <404: Not Found>

But when I created a user and gave it admin role using UserRole model it was able to delete it. 
I haven't written any test case before. I have given my test case below. Can you look at it and tell me the mistake?

https://pastebin.com/9PqKXkkP

gravitystorm:
It's hard to be sure since that's not a complete test case, but check here:

```
:params => { :display_name => administrator_user.display_name, :id => public_trace_file.id }
```
Is `public_trace_file` actually owned by `administrator_user`? I would expect the `display_name` to be a regular user here (e.g. `public_trace_file.user.display_name`). You are trying to log in as an administrator (the session part of the command) but the trace you are trying to delete should by owned by a normal user, for the test to be correct.

biswesh456:
I tried it. But it gave the same result. Should I commit the changes so that you can see the entire file?

gravitystorm:
@biswesh456 sure, that might be easier.

gravitystorm:
As far as I can see, you're now trying to delete the same trace twice. So it gets deleted by the user who owns it, and then when you try deleting it again as an administrator, you get a 404 response.

You could create a separate test case for deleting the trace as an admin user, that would be one way to solve the problem.

biswesh456:
@gravitystorm I have made the changes after your suggestion and now it passed all the test cases.

biswesh456:
I am creating a new pull request #1765 for this and closing this pull request.

PR #1738, opened 2018-02-11
Fixed motorway directions when using OSRM directions

jguthrie100:
- OSRM directions now have named motorways instead of naming them all as unnamed roads.
 - GraphHopper roads are now written in bold, to bring them in alignment with OSRM styling

### DIFF
--- a/app/assets/javascripts/index/directions/graphhopper.js
+++ b/app/assets/javascripts/index/directions/graphhopper.js
@@ -47,7 +47,14 @@ function GraphHopperEngine(id, vehicleType) {
             var instr = path.instructions[i];
             var instrCode = (i === len - 1) ? 14 : GH_INSTR_MAP[instr.sign];
             var instrText = "<b>" + (i + 1) + ".</b> ";
-            instrText += instr.text;
+
+            var directionText = instr.text;
+            if (directionText.includes(" onto ")) {
+              var directions = directionText.split(" onto ");
+              directionText = directions[0] + " onto <b>" + directions[1] + "</b>";
+            }
+
+            instrText += directionText;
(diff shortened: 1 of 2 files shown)


tomhughes:
There's no need to open an issue and then immediately open a PR to fix it - it just creates noise.

tomhughes:
I don't think parsing the text returned from the API is sensible - the correct approach is to get upstream to return structured information.

tomhughes:
And why on earth is a PR titled "Fixed motorway directions when using OSRM directions" changing the graphhopper interface?

jguthrie100:
Apologies for opening an issue and a PR request. I assumed it was always best to link a PR to an issue.

The graphhopper was an oversight that I didn't think worthy of opening another PR request for. Should it be in two separate PR requests, or just a single PR request with a broader scope/title?

tomhughes:
Well I'm not sure it really matters given that both changes seem to to rely on parsing the returned directions, which seems very fragile to me so I'm not inclined to accept either change.

jguthrie100:
The unnamed roads for motorways seems like a really big thing to leave as an unnamed road when the motorway name is listed in `step.ref` without needing to do any text parsing.

To have directions that say _"turn left onto unnamed road"_ when unnamed road is a national highway, seems like a massive oversight, especially when the fix is so easy!

The parsing out of the $east or $west doesn't need to happen, it just makes the output look a lot cleaner.

i.e. `Turn left onto Sunbelt Highway (I5)` instead of `Turn left onto Sunbelt Highway (I5 $east)`.

Even leaving it with the $east tag is **much** better than leaving it as _unnamed road_, although removing the $east tag is better in my opinion although I can understand your reservations.

PR #1739, opened 2018-02-11
OSRM directions now include motorway/freeway names

jguthrie100:
At the moment, most motorways are listed as unnamed roads.

This is because, for some reason, the OSRM API returns normal road names (i.e. Savile Street) in the `step.name` attribute in the JSON response, however, includes motorway/freeway names (i.e. M1) in the `step.ref` attribute.
Sometimes motorways are named (i.e.  _L'Autoroute du Nord_ heading north out of Paris), however, it would still be useful to have the numeric reference too (in this case, _A1_)

At the moment, the _L'Autoroute du Nord_ out of Paris is referenced just by name (i.e. _"Turn left onto **L'Autoroute du Nord**"_, however, it would be very beneficial when planning routes if the route number is indicated too (i.e. _"Turn left onto **L'Autoroute du Nord (A1)**"_), which this pull request does.

Likewise, the M1 in Britain is an unnamed road, so at the moment directions are in the form of _Turn left onto unnamed road (110km)_, which seems like a massive oversight when referencing a national highway system!
My changes change this so it now says _Turn left onto **M1**_.

In addition, main trunk roads are also referenced by road number too, so "Turn left onto **Tottenham Court Road**" becomes "Turn left onto **Tottenham Court Road (A400)**", which again is very useful for route planning.

_____

I have tested the changes across the USA, Africa, Europe, Japan, China etc, and in all cases times where you are directed along 1000s of kms of unnamed road, it now directs you along 750km of E1 or 300km of N23 for example.

### DIFF
--- a/app/controllers/export_controller.rb
+++ b/app/controllers/export_controller.rb
@@ -19,7 +19,7 @@ class ExportController < ApplicationController
       format = params[:mapnik_format]
       scale = params[:mapnik_scale]
 
-      redirect_to "http://render.openstreetmap.org/cgi-bin/export?bbox=#{bbox}&scale=#{scale}&format=#{format}"
+      redirect_to "https://render.openstreetmap.org/cgi-bin/export?bbox=#{bbox}&scale=#{scale}&format=#{format}"
     end
   end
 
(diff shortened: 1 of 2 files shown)


jguthrie100:
@tomhughes I've taken care not to parse out data from any strings. The only manipulation is using data already included in the API response, which I use in full.

Issue #1573, opened 2017-07-03
OSRM directions omit motorway_link destinations


1ec5:
If I get OSRM (car) directions over [this `motorway_link`](http://www.openstreetmap.org/way/11501858), [the resulting route step](http://www.openstreetmap.org/directions?engine=osrm_car&route=39.22313%2C-76.66095%3B39.22177%2C-76.65376) has the instruction:

> Take the ramp on the right onto unnamed road

By comparison, the [Project OSRM demo site](http://map.project-osrm.org/?z=18&center=39.222768%2C-76.660208&loc=39.222475%2C-76.660479&loc=39.222415%2C-76.653072&hl=en&alt=0) say:

> Take the ramp on the right towards MD 295 North: Baltimore

/ref #1262
/cc @TheMarex

Penegal:
Support of directions by OSRM on OSM.org seems totally missing or broken: directions for a route using [this `secondary` way](https://www.openstreetmap.org/way/242805694) only say to _take 2nd exit onto Avenue Victor Hugo (D 266)_, and then to _exit roundabout onto Avenue Victor Hugo (D 266)_. OSRM as used on OSM.org is not using its `destination` tag. In fact, I can't remember seeing the `destination(:*)` tags used with OSRM as proposed on OSM.org, when routing on ways I know these tags were present.

1ec5:
The original issue in https://github.com/openstreetmap/openstreetmap-website/issues/1573#issue-240218282 regarding motorway links has been fixed. The issue in https://github.com/openstreetmap/openstreetmap-website/issues/1573#issuecomment-389142630 has not been fixed. It doesn’t reproduce [on the Project OSRM demo site](http://map.project-osrm.org/?z=18&center=48.307820%2C6.133580&loc=48.308404%2C6.133100&loc=48.307156%2C6.133615&hl=en&alt=0).

PR #1740, opened 2018-02-12
Fixed on ramps and off ramps leading onto unnamed roads

jguthrie100:
Fixes issue #1573  

The directions text on on and off ramps now says (for eg) 'Take the ramp on the left towards Baltimore, I58' instead of 'Take the ramp on the left onto unnamed road'

Tested on motorways/highways across USA and Europe


jguthrie100:
without_directions option removes the unnamed road text when its redundant and just leaves the instruction at `Take the ramp on the left`.

Matches the OSRM site more closely

tomhughes:
If I'm understanding this right then this doesn't actually limit itself to unnamed roads right? It changes all on and off ramps to be described in terms of the destination (if any) and not to mention the road you are moving on to?

Wouldn't it be better to show both if they are known? or at least to show the road name if that is known and the destination isn't?

jguthrie100:
Yeah that makes more sense.

Is there any chance you could have a look at #1741 , because when/if that gets committed, it'll change the logic required for checking whether an off/onramp has a valid name..

PR #1743, opened 2018-02-12
Fix the upcoming issue with EasyList blocking

Zverik:
Changing `<div id="banner">` to `<div id="event">` to better reflect the contents of banners and to circumvent ad blockers that use EasyList.


mxa:
You could randomize the id string to make blocking a little harder.

Zverik:
> You could randomize the id string to make blocking a little harder.

The plan is not to circumvent any blocking. It's just to make OpenStreetMap project stand out in the black-list. Then we can share the story of blocking open and non-profit initiatives with the media.

tomhughes:
Whilst I appreciate the aim all this is going to do is start an arms war, and that isn't going to end well for anybody.

PR #1745, opened 2018-02-12
Map pans to improved position when clicking on directions

jguthrie100:
Selected direction & popup now pan to the center of the screen, whereas before sometimes it would have the highlighted junction & popup tucked away right in the corner..

As an example, go to https://www.openstreetmap.org/directions?engine=osrm_car&route=55.9496%2C-3.1915%3B52.2033%2C0.1249
- Zoom in close to street level at a location far away from the route (i.e. zoom into central Paris)
- Then click on of the direction instructions on the left, and the map will pan to the selected direction point, but it will be right in the corner of the screen most likely.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -7,7 +7,7 @@ OSM.Directions = function (map) {
   var dragging;        // true if the user is dragging a start/end point
   var chosenEngine;
 
-  var popup = L.popup();
+  var popup = L.popup({autoPanPadding: [100, 100]});
 
   var polyline = L.polyline([], {
     color: '#03f',


tomhughes:
Isn't using `panTo` a bit drastic? All we really want is to make sure the location is visible, so maybe something like `panInsideBounds` would be better?

To be honest I'm slightly surprised that opening a popup doesn't force it to pan so the poup is visible....

tomhughes:
So as you say it does already pan to make the popup visible, which seems fine to me. I don't see any need to force it to the centre - that just means that if you click on a series of junctions it will be constantly panning.

jguthrie100:
It just seems really unfunctional to me to have the popup tucked away in the corner.

image

To me the above seems **much** more intuitive and user-friendly than the below image, which is almost going out of its way to be as hidden as possible!

image

I really feel makes the most sense from a UI point of view, even if it requires a panTo (which shouldn't be too much of an overkill bearing in mind how infrequently directions are clicked and how powerful most computers and even mobiles are these days)

tomhughes:
You did deliberately pick the worst possible case there - more commonly what happens is that the route runs across the visible view and the popup moves along it as you click on different steps.

The only way to get that is to deliberately pan away from the route and then click on a step, which is hardly a normal way to interact with a route.

tomhughes:
Here's my suggestion - it comes in two steps:
* When the popup is created specify `autoPanPadding: [100, 100]` or similar in the options so that there is some padding from the edge of the map when it opens.
* Rather than panning to the popup after opening, pan to the route before doing so by doing `panInsideBounds(polyline.getBounds().pad(0.05))` on the map so that the map pans back to the route but without changing zoom.
That way we ensure the user is looking at the route if they have panned away, and we prevent popups being to close to the edge, but without constantly panning the map to keep it dead centre.

jguthrie100:
Added the autoPanPadding which works.

panInsideBounds doesn't help and just makes the map jump around erratically for some reason.

PR #1747, opened 2018-02-13
Added km & miles distance selection

jguthrie100:
Added the option to select between miles and kilometers on the directions page

There is no change in output to what is currently outputted, except for a small change in the rounding on distances between 5000 and 10000m.

Currently 6800m will be displayed as 7km, but with my changes it becomes 6.8km. This is the same for all distances between 5000 and 10000m.

I can make changes so that the formatting is entirely consistent with the current formatting, but it will require a number of if() statements which are probably more hassle than its worth.

image


tomhughes:
So the obvious problem here is that it mixes metric and imperial units - surely if you've chosen miles it should use feet or yards for shorter distances?

jguthrie100:
I did wonder about that.

I can switch it to yards easily enough. I only know that in my personal experience, I think in metres and miles, which is maybe a bit strange.

What would be a good scale for using yards and feet? Would it be best to just ignore one of the two and just have every distance under 1 mile being measured in feet (or yards), rather than switching from feet to yards to miles.

bhousel:
> What would be a good scale for using yards and feet? Would it be best to just ignore one of the two and just have every distance under 1 mile being measured in feet (or yards), rather than switching from feet to yards to miles.

A common way to handle this in Imperial units is to display the distance in miles down to about 0.2 mile (rounded to nearest tenth, e.g. "1.3 miles"), then display in feet below that..  Definitely avoid using yards for this.

tomhughes:
I think there's a bit of an i18n thing there - as @bhousel says I believe feet is most common which I think is because that is what the US normally does. In the UK it would be normal to use yards for short distances on road signs though ;-)

Feet is probably the best answer though.

jguthrie100:
Definitely a i18n issue.

I'm pretty certain on sat nav (in the UK at least), you'll often hear "In 300 yards, turn left".
I don't think I've ever heard it in feet though..

jguthrie100:
Added a locale check so that en-GB uses yards and everywhere else uses feet.

And a quick question about the CI not allowing for default params (` 'default parameters' is only available in ES6 (use esnext option) `)
Is allowing for ES6 an option, or do I need to rewrite that part?

tomhughes:
Seriously, no. If you want to support both then do it via a key in the locale file so that other locales can choose what to use.

app/assets/javascripts/index/directions.js
```diff
@@ -137,13 +138,36 @@ OSM.Directions = function (map) {
...
-    } else if (m < 10000) {
-      return (m / 1000.0).toFixed(1) + "km";
+  $(".directions_form .distance_units a").on("click", function() {
+    $(".directions_form .distance_units a").removeClass("selected");
+    $(this).addClass("selected");
+    distanceUnits = $(this).data("unit");
+    getRoute();
+  });
+
+  function formatDistance(m, distanceUnits = "km") {
```
bhousel: I'd avoid the ES6 Javascript here - it won't work on IE11.

jguthrie100:
@tomhughes

Lonkmapp:
Please add this poropsal to the website ! It’s so needed ! Many people use miles instead of kilometers , I love openstreetmaps and really would be so thrilled to see this ! Please support this idea ! please please please

Lonkmapp:
Please please please p

AntonKhorev:
Was reimplemented in #5915.

PR #1751, opened 2018-02-14
Stop map zooming out after dropping Route end marker

jguthrie100:
PR to stop the map zooming out after a new route has been dragged.

It brings it into line with how Google, Bing etc handle route dragging (just leaving the map view as it is), which seems like the most user friendly way to go about it :)

Map still zooms out when searching a new route via the 'Go' button


tomhughes:
I'm not entirely sure what's going on but this doesn't seem to work reliably for me in Firefox - it does stop many of the pans and zooms but not all.

I think, but I'm not entirely sure, that it's when a drag is quite quick that it doesn't work.

tomhughes:
I think the problem is when there is a `getRoute` running from the dynamic drag and it finished with an error after `dragEnd` is set this clearing that before the real `getRoute` finishes.

I have an alternate fix that avoids global state which I'm going to push.

PR #1754, opened 2018-02-14
Stop setValue() always changing input text to coords

jguthrie100:
Stops the input text always getting set to coords.

Was unsure to use `endpoints[0].value` or `$("#route_from").val()`. Went with the jQuery because the first time you hit Go, `endpoint.value` is still the non-geocoded text (despite the input box text being geocoded)

Also not sure of positioning of `input.val(value)` in the `setValue()` method, and whether it needs to be copied into the else statement too (or whether it needs some different refactoring), but the general idea works.

Geocoded values are now saved in the url and stay as geocoded string values when reversing directions and pressing Go etc.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -62,11 +62,12 @@ OSM.Directions = function (map) {
     endpoint.setValue = function(value, latlng) {
       endpoint.value = value;
       delete endpoint.latlng;
-      input.val(value);
 
       if (latlng) {
         endpoint.setLatLng(latlng);
+        input.val(value);
       } else {
+        input.val(value);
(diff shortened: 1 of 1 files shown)


tomhughes:
I don't think it's a good idea to keep them in the URL all the time - it makes the URL very long.

I only put them in when doing the reverse in order to preserve them across the swap though I notice that doesn't seem to be working - sure it was when I wrote it though!

tomhughes:
You definitely need to set the value in the other branch by the way, but before the `getGeocode`.

I think I didn't realise `setLatLng` would replace the UI text as well as setting the internal values, so I was expecting the value that had just been set to be preserved.

jguthrie100:
@tomhughes

PR #1755, opened 2018-02-14
Fix route path 'hovering' when marker positions are changed

jguthrie100:
If you have a route already on screen, and then type in a new location in either the start or end destinations and click anywhere except for basically the Go button, then the marker position moves, but the route remains going to the old position.

Likewise, if you then type in a new destination for the other marker and click anywhere except the Go button, then you end up with 2 markers in their respective positions and a route path just gliding across the screen unconnected from everything!

I propose that getRoute() is called everytime the input text is changed because chances are, the marker will move and a new route will need to be calculated.

I do this by simply setting `awaitingGeocode = true` when the input text changes.

___
I also included in some refactoring which should probably be in another PR.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -54,6 +54,8 @@ OSM.Directions = function (map) {
     });
 
     input.on("change", function (e) {
+      awaitingGeocode = true;
+      
       // make text the same in both text boxes
       var value = e.target.value;
       endpoint.setValue(value);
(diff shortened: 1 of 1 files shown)


tomhughes:
We can't be constantly firing route requests as users type though - it may even be in breach of the terms for some of them.

I think the only router where we could even think of doing this is OSRM as I think that's the only one where do real time updates during a marker drag?

jguthrie100:
I wasn't really clear.. it only fires off a request when the focus leaves the input box. So you type in "Calgary" and then change focus (click in the other box, or tab out or something) and at that point it makes a request.
Definitely doesn't do it for every letter typed!

tomhughes:
So what exactly does the "other refactoring" do?

jguthrie100:
Deletes `endpoint.latlng = L.latLng(json[0]);
endpoint.marker
  .setLatLng(endpoint.latlng)
  .addTo(map);`
and makes a call to `endpoint.setLatLng()` instead

jguthrie100:
To be honest, the refactoring isn't really required. I can remove it and just leave the specific line of code in the `on input change` method if that's preferred?

PR #1758, opened 2018-02-16
Update country.rb

ardean80:
Without adding "Rails.root.join(...)" (or something else returning a complete path to the countries.xml file) this error occurred: 
"No such file or directory @ rb_sysopen - config/countries.xml" 
I think it happened because it could not find that file through a relative path.

### DIFF
--- a/lib/country.rb
+++ b/lib/country.rb
@@ -19,7 +19,7 @@ class Country
 
   def self.load_countries
     countries = {}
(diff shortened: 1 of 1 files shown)


gravitystorm:
Thanks! I couldn't reproduce the problem you report, but the solution looks good to me in any case.

tomhughes:
It will only happen if the current directory isn't the same as the rails root, which probably depends on the server you are using.

PR #1759, opened 2018-02-17
Missing autoconf dependency

simonpoole:
Seems as if the fii (whatever that does) gem requires autoconf to install/build.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -28,7 +28,7 @@ of packages required before you can get the various gems installed.
 These can be installed on Ubuntu 16.04 or later with:
 
 ```
-sudo apt-get install ruby2.3 libruby2.3 ruby2.3-dev \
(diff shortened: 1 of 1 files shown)


tomhughes:
You're reading it wrong - what it needs is libffi-dev which is why I added that to the production servers the other day.

gravitystorm:
If it now needs libffi-dev then don't we need to add that the installation notes?

tomhughes:
Well sure, that was my point, that it needs that adding, not autoconf.

simonpoole:
Just as a clarification I was updating an existing installation, so was working from a slightly different situation as a completely new install (I simply added what was obviously missing the install of the gem), but naturally adding the correct dependency to the doc makes more sense

tomhughes:
Well you should never need autoconf - that is only needed by a developer who needs to rebuild `configure` from `configure.in`.

In any case I don't think ruby modules even use that do they? They use mkmk and extconf.rb?

Issue #1762, opened 2018-02-23
old_name key name not linked to wiki without clear reason


matkoniecz:
see https://www.openstreetmap.org/way/34499083#map=19/49.96249/18.69686

name tag is linked, old name is not linked

I see no differences between https://wiki.openstreetmap.org/w/index.php?title=Key:name&action=edit and https://wiki.openstreetmap.org/w/index.php?title=Key:old_name&action=edit that should matter

old_name was not edited recently (see https://wiki.openstreetmap.org/w/index.php?title=Key:old_name&action=history ) so that it should not be a cache issue

mmd-osm:
See #1736: the automated script removed the old_name tag the last time it ran. There were no error messages whatsoever. Maybe try and run the script again to see if it makes a difference.

matkoniecz:
I will try running https://raw.githubusercontent.com/mmd-osm/openstreetmap-website/master/script/misc/update-wiki-pages - for now I am installing dependencies of thing necessary to install dependencies if this script.

matkoniecz:
For people attempting to figure out similar problems: this Perl script relies on https://www.mediawiki.org/wiki/API:Embeddedin and run queries similar to https://wiki.openstreetmap.org/w/api.php?action=query&list=embeddedin&eititle=Template:KeyDescription&eilimit=500 (further pages may be request by passing eicontinue in &eicontinue= parameters),

At least in this case wiki API failed to include page with a template that exists on it, so it is a wiki bug. I will try to clear Wiki cache by making edit to this page and rerun this script.

matkoniecz:
There is also https://wiki.openstreetmap.org/w/index.php?title=Special%3AWhatLinksHere&limit=500&target=Template%3AKeyDescription&namespace=0 that is probably using the same internal calls to display similar list and is easier to paginate for humans.

PR #1763, opened 2018-02-24
make cpanp command help working without requiring changes from user

matkoniecz:
https://raw.githubusercontent.com/mmd-osm/openstreetmap-website/master/script/misc/update-wiki-pages has

>         print "   sudo cpanp MediaWiki::API YAML::XS\n";

From my own experience and reading http://search.cpan.org/~bingos/CPANPLUS-0.9172/bin/cpanp it seems that it should be 

> sudo cpanp i MediaWiki::API YAML::XS

### DIFF
--- a/script/misc/update-wiki-pages
+++ b/script/misc/update-wiki-pages
@@ -9,7 +9,7 @@ use Getopt::Long ();
 BEGIN {
     eval "require MediaWiki::API; require YAML::XS;" or do {
         print "You have to install some modules via CPAN to run this:\n";
-        print "   sudo cpanp MediaWiki::API YAML::XS\n";
+        print "   sudo cpanp i MediaWiki::API YAML::XS\n";
         exit 1;
     };
 }


tomhughes:
Well there's no one way to install a perl module and I'm not sure we should be picking sides between os packages vs cpan vs cpanp never mind privileged vs unprivileged installation.

matkoniecz:
> I'm not sure we should

At least for me (I attempting to fix #1762) this instruction was really helpful.

> os packages vs cpan vs cpanp

in this case cpanp is already selected, but provided command is invalid. AFAIK this PR is not changing anything as this competition goes.

gravitystorm:
I'm happy to fix the script, rather than keep it as-is and broken. I don't see that cpan vs cpanp is relevant to this PR. But is it really just adding 'i' or should there be a dash i.e. '-i'?

tomhughes:
The script isn't really broken as such - the problem is that the script tries to be overly helpful if you don't have the right libraries installed by trying to tell you how to install them. The problem is that there is no one right way to do that. For me on Fedora my preferred approach would be:

```
sudo dnf install 'perl(MediaWiki::API) perl(YAML::XS)'
```

But that is specific to me because I happen to have the MediaWiki::API module packaged locally as it isn't in the Fedora repos.

For Ubuntu my preference would be:

```
sudo apt-get install libmediawiki-api-perl libyaml-libyaml-perl
```

Of course you can always install directly from the CPAN repository either using the older `cpan` client or the newer `cpanp` client - in both cases you are likely to wind up being asked a lot of confusing questions if you've never used it before. You also then have the question of whether to install system wide using sudo or to your user account.

gravitystorm:
Alright, how about we add the packages to INSTALL.md instead? I know that people won't run this script very often (if at all), but it's perhaps better to include them in the existing installation instructions, rather than having more installation instructions hidden away in this script?

Or if that's still a cross-platform packaging hassle, we could add the mediawiki gem to the Gemfile and rewrite the script in ruby!? :smiley:

matkoniecz:
> Alright, how about we add the packages to INSTALL.md instead?

I am not sure is it preferable - at least in my case that triggered this PR I used this script without running entire website or any other code.

matkoniecz:
ping?

tomhughes:
I've made my views clear - for the tiny set of people that are ever going to run this script there's really no need to make it idiot proof. I say we stop trying to specify how to do it and just require the modules and let the user decide how best to install them.

tomhughes:
I mean I don't understand why we are trying to hand hold for those modules but not for Getopt::Long, Pod::Usage or Test::More...

matkoniecz:
If you think that it should not be merged as it makes situation worse feel free to close it (and least I prefer to get my PRs/issues rejected and WONTFIXed rather than stuck in a limbo).

PR #1765, opened 2018-02-24
Allow admins and moderators to delete traces

### LINKED ISSUE #1625: Allow admins and/or moderators to delete traces

biswesh456:
Fixes #1625

### DIFF
--- a/app/controllers/trace_controller.rb
+++ b/app/controllers/trace_controller.rb
@@ -186,7 +186,7 @@ class TraceController < ApplicationController
 
     if !trace.visible?
       head :not_found
-    elsif current_user.nil? || trace.user != current_user
+    elsif current_user.nil? || (trace.user != current_user && !current_user.administrator? && !current_user.moderator?)
       head :forbidden
     else
       trace.visible = false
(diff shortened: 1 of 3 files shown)


biswesh456:
@gravitystorm I have opened this new pull request since I kind of messed up with the previous branch while rebasing 😅. I have made changes to the test cases as you had suggested.

gravitystorm:
Thanks! It looks good. I would ask you to make two more changes:

1) In the test, you have

> post :delete, :params => { :display_name => **admin**.display_name, :id => public_trace_file.id }, :session => { :user => admin }

The `display_name` parameter (i.e. the name used in the URL) should be for the user of the trace, not the admin doing the deleting. I know that the `display_name` is currently ignored by the code, but it's best to change it to: 

> post :delete, :params => { :display_name => **public_trace_file.user**.display_name, :id => public_trace_file.id }, :session => { :user => admin }

2) When an admin or moderator deletes a trace, I think it is best for them to be redirected to the list of traces for the user that owned that trace, rather than redirecting the admin to their own list of traces. This can be achieved by changing (in app/controllers/trace_controller.rb)

>  redirect_to :action => :list, :display_name => **current_user**.display_name

to 

> redirect_to :action => :list, :display_name => **trace.user**.display_name

and adjusting the test accordingly.

biswesh456:
@gravitystorm Sorry for the late response as I was having my university exams. I have changed the code as suggested by you.

gravitystorm:
Great! I made one small simplification to the redirection, and then merged your changes. Thanks!

PR #1769, opened 2018-02-27
Consistently use "trace" everywhere

### LINKED ISSUE #1644: Be consistent: track vs. trace
https://www.openstreetmap.org/traces says

```
Public GPS traces
Browse recent GPS track uploads Upload a trace See your traces
```

To be consistent that should be

```
Public GPS traces
Browse recent GPS trace uploads Upload a trace See your traces
```

Same with the ```See your traces``` page.

aaronlidman:
Fixes #1644


tomhughes:
Fixing potlatch 1 is possibly overkill ;-)

PR #1773, opened 2018-02-27
Display coordinates on notes pages

### LINKED ISSUE #1355: Add coord for every note
When you open http://www.openstreetmap.org/note/294569 - coordinates are not shown to user (http://www.openstreetmap.org/note/294569#map=15/56.8874/65.3966).

mavl suggested to move map a little to get this information http://www.openstreetmap.org/user/mavl/diary/39791

But 1. coordinates would be imprecise 2. **website can do this for user**

aaronlidman:
untitled*before and after*

Fixes #1355.

Clicking note coordinates simply zooms to the area at Z18, just like node pages behave. I also cleaned up the warning, which impacted export and note creation pages. Here they are now.

screen shot 2018-02-27 at 12 43 52 pm
screen shot 2018-02-27 at 12 44 13 pm

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1214,8 +1214,9 @@ tr.turn:hover {
   }
 
   .warning {
-    margin: 0 0 $lineheight/2 0;
-    padding: 0 $lineheight/2;
+    margin: 0 10px;
+    padding: 10px;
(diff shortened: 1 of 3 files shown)


gravitystorm:
The warnings part of this PR were reimplemented in #2711 and the link to the map reimplemented in #2712 

Thanks for the original PR!

PR #1774, opened 2018-02-28
Add counts for notes and diary comments to user profiles

### LINKED ISSUE #1643: Add note and comment counts too the user profile
On Profile, one sees
```My Edits 508 My Notes My Traces 20 My Diary 0 My Comments My Settings```
this should be
```My Edits 508 My Notes 2 My Traces 20 My Diary 0 My Comments 0 My Settings```
else it looks broken.

### PR DESCRIPTION
**Pull request by aaronlidman**
new counts

Fixes #1643.

### DIFF
--- a/app/models/note_comment.rb
+++ b/app/models/note_comment.rb
@@ -25,7 +25,7 @@
 
 class NoteComment < ActiveRecord::Base
   belongs_to :note, :foreign_key => :note_id, :touch => true
-  belongs_to :author, :class_name => "User", :foreign_key => :author_id
+  belongs_to :author, :class_name => "User", :foreign_key => :author_id, :counter_cache => :notes_count
 
   validates :id, :uniqueness => true, :presence => { :on => :update },
                  :numericality => { :on => :update, :integer_only => true }
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1,10 +1,3 @@
---
--- PostgreSQL database dump
---
-
--- Dumped from database version 9.5.7
--- Dumped by pg_dump version 9.5.7
-
 SET statement_timeout = 0;
 SET lock_timeout = 0;
 SET client_encoding = 'UTF8';
@@ -130,7 +123,7 @@ CREATE TYPE user_status_enum AS ENUM (
 
 CREATE FUNCTION maptile_for_point(bigint, bigint, integer) RETURNS integer
     LANGUAGE c STRICT
-    AS '$libdir/libpgosm', 'maptile_for_point';
+    AS '/srv/openstreetmap-website/db/functions/libpgosm.so', 'maptile_for_point';
 
 
 --
@@ -139,7 +132,7 @@ CREATE FUNCTION maptile_for_point(bigint, bigint, integer) RETURNS integer
 
 CREATE FUNCTION tile_for_point(integer, integer) RETURNS bigint
     LANGUAGE c STRICT
-    AS '$libdir/libpgosm', 'tile_for_point';
+    AS '/srv/openstreetmap-website/db/functions/libpgosm.so', 'tile_for_point';
 
 
 --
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by tomhughes** [comment]
These need to be cached in the same way as the other counts.

**Comment by aaronlidman** [comment]
@tomhughes how's it look now?

The changes to `db/structure.sql` look a little scary, not sure if that substantive of a change is expected.

File: app/models/note_comment.rb [hunk 1]
```diff
@@ -25,7 +25,7 @@
 
 class NoteComment < ActiveRecord::Base
   belongs_to :note, :foreign_key => :note_id, :touch => true
-  belongs_to :author, :class_name => "User", :foreign_key => :author_id
+  belongs_to :author, :class_name => "User", :foreign_key => :author_id, :counter_cache => :notes_count
```
tomhughes: What was the reason for overriding the name here? Would the default of `note_comments_count` not be more accurate?

aaronlidman: I think `note_comments_count` is less accurate because we're not counting the number of comments but rather the number of authored notes. So `note_comments_count` isn't exactly right because a user can comment multiple times on a note but maybe `note_authored_count` is a better name?

tomhughes: Are you sure it's counting the number of notes? I'm pretty sure that's going to count the total number of comments isn't it?

It's on the relationship between user and note comments, and the documentation explains:

> The counter cache is incremented when an object of this class is created and decremented when it's destroyed. 

So it will count every object of class `NoteComment` belonging to a given user.

aaronlidman: Ah that's right, I misunderstood.

Since there's no distinction of an author in the `Note` class, do you have a recommendation for how to count just the first comment on a note? That logic would have to be within the `NoteComment` class right?

tomhughes: I don't think you can using the builtin counter caches, or at least I tried various things without any success.

The general suggestion seems to be https://github.com/magnusvk/counter_culture which allows for conditional counters.

db/structure.sql
```diff
@@ -130,7 +123,7 @@ CREATE TYPE user_status_enum AS ENUM (
 
 CREATE FUNCTION maptile_for_point(bigint, bigint, integer) RETURNS integer
     LANGUAGE c STRICT
-    AS '$libdir/libpgosm', 'maptile_for_point';
+    AS '/srv/openstreetmap-website/db/functions/libpgosm.so', 'maptile_for_point';
```
stefanb: These changes likely do not belong here, as they are the result of setting up your local environment.
Make the changes include only relevant lines (adding `notes_count` and `diary_comments_count` in `users` table). Not sure why this file gets so vastly different - i noticed this problem also while setting up my dev environment.

tomhughes:
Because it is sensitive to things like where you have the 3GL functions installed. It can also vary in non-meaningful ways if you're running a different version of postgres as it's basically just the output of `pg_dump`.

db/structure.sql
tomhughes: Because it is sensitive to things like where you have the 3GL functions installed. It can also vary in non-meaningful ways if you're running a different version of postgres as it's basically just the output of `pg_dump`.

stefanb: Then it would make things easier and to just `.gitignore` the whole file.... unless you use the diff for checking/verifying the effects of the changes on the actual DB.

tomhughes: Oh I'd love not to have it, but people insisted on having a schema...

gravitystorm:
Aside from the unwanted changes to the structure.sql file, it might also be worth removing the notes count for now since it's more complex to get right. The diary comments count part doesn't need to wait!

gravitystorm:
I think this PR has been abandoned, and it's not in a mergeable state. 

I'd still be happy to see a fresh PR for the diary comments count, as per my comment from 2018. I think the notes/note_comments bit would be best addressed separately.

PR #1779, opened 2018-03-05
Added color preview box in tag browser sidebar

stefanb:
Applied to the colour tags (https://wiki.openstreetmap.org/wiki/Key:colour) in the sidebar, having:
* key either:
  * colour
  * building:colour
  * ref:colour
  * roof:colour
* and value either:
  * 3 or 6 character hex code prefixed by "#", case insensitive (eg "#ffaabb" or "#fff" or "#FF0505") or
  * w3c colour name (case insensitive, eg "deeppink" or "DEEPPINK" or "DeepPink"...), full list from https://www.w3.org/TR/css-color-3/#svg-color
 
In such cases a 20px x 20px color preview box is floated to the right of the tag value field to be as unintrusive as possible, but still allowing the color preview and confirming the editors that the tag was entered correctly.

Example:
image


tomhughes:
Not sure I'm really the right person to review this given I don't think those tags belong in the database in the first place...

stefanb:
Well, these tags are well documented in wiki and there are many such tags in use in the DB:
* https://taginfo.openstreetmap.org/keys/?key=colour#values
* https://taginfo.openstreetmap.org/keys/?key=building%3Acolour#values
* https://taginfo.openstreetmap.org/keys/?key=ref%3Acolour#values
* https://taginfo.openstreetmap.org/keys/?key=roof%3Acolour#values

app/helpers/browse_helper.rb
```diff
@@ -188,4 +190,16 @@ def telephone_link(_key, value)
 
     "tel:#{value_no_whitespace}"
   end
+
+  def colour_preview(key, value)
+    return nil unless (key =~ /^(|building:|ref:|roof:)colour$/ || key =~ /^(int_)?ref:colour(_(bg|tx))?$/) && !value.nil?
```
Zverik: Maybe make it more generic? [Taginfo shows](https://taginfo.openstreetmap.org/search?q=%3Acolour) many other colour tags, including three for seamarks and longer variants like `building:roof:colour`.

And `int_ref:colour` options are in hundreds, why make a clause specifically for these?

stefanb: If we just look at the existing data and disregard the wiki, we should 
* also make the "u" in "colour" optional. - https://taginfo.openstreetmap.org/search?q=color
* allow any optional prefix and suffix, not requiring any special separators
* drop the case sensitivity

Which makes the regex boild down to simple:
```
/colou?r/i
```
which can be understood that the editor wanted to tag some colour property and if the value indeed looks like a colour (hex code or valid w3c value) we can provide the preview.

Or we could require some common key structure. According to https://taginfo.openstreetmap
require some common key structure. According to https://taginfo.openstreetmap.org/search?q=colour we could generalize it to:
* alow any optional prefix, separated by colon (:) or underscore (_)
* allow any optional suffix, separated by colon (:) or underscore (_)
* insist on british spelling with "u"
* require lowercase
```
/^(.+[:_])?colour([:_].+)?$/
```

Both generalizations would allow us to preview all sorts of future colour tags, even unforeseen enforcements:
image

What level of generalization did you have in mind? 
* any tag or any tag that looks like it is related to colour
* any structure or some specific separators
* any spelling (british or us also)
* any case (upper, lower, whatever)

gravitystorm: I'm very much opposed to anything that shows tagging mistakes as if they are correctly tagged. It's part of the guidelines established for the openstreetmap-carto project, and I'd like to apply similar guidelines to the browse pages:

"It's an important feedback mechanism for mappers to validate their edits and helps to prevent unfavorable fragmentation of tag use."

So no "color" tags and nothing that would show e.g. "buidling:colour" (note misspelling).

gravitystorm: (I've also removed an image from your comment - I know the joke, but it's not obviously a joke when shown out of context)

Zverik: I agree that accepting `color` is a bit too much.

stefanb: Yes, it is very important to indicate whether tagging was done *correctly* or not.... as much as possible.

The colour tag **value** is well defined and thoroughly checked and preview box is only shown if the value is a valid colour. User can then validate the colour correctness visually (editors could offering color pickers).

**Key** validation on the other hand is performed in the key column of the sidebar. Currently the only visual indicator is the presence of the link to wiki. That link is unfortunately missing for many valid keys (eg when key legitimately contains some variable or the wiki page is not existing or redirecting to a more general key). There might be also cases when the key was deprecated, but the link still exists because the page remained in wiki for reference. Key validity indication could be improved, to not just existence of wiki page but
- also some other metadata from wiki (redirect, status...) or 
- link to most suitable wiki page (eg it the key is "building:levitating" which does not exist in wiki, just link to the first part "building" to the "key:building" wiki page which would give editors more hints about the topic and how to fix the tagging) or 
- OSM data (eg order tags by key occurance count)
- some keys are only valid on specific elements (relations, ways, nodes), 
- some keys are requiring presence of related tags (eg drink:* tag requires appropriate shop or amenity tag)

Key validation is a complex topic to be tackled separately.

The question here is to what extent should the key validity influence the value validation, given that OSM has an evolving schema where no key can be "totally wrong", just maybe discouragd and less (or not at all) supported by some tools (renderers, routing engines...), and the community tries to reach consensus before introducing new keys, and educating newcomers of established schema. 

Some minimal influence (key must at least contain "colour") is needed to prevent color preview boxes showing up occasionally in the name and other totally irrelevant tags.

It would not be clear to users to know why they don't see the colour preview box for their mispelled "buidling:colour=#aaa" tag. Absence of color preview box may say that *something* is wrong, but it would not point to the key, but would most likely leave user wondering whether should it be "#aaaaaa" or "lightgrey" or "lightgray" or "paledesertwhite" or ...? 

----

I have played a bit with [100 most common keys sontaining "colour"](https://taginfo.openstreetmap.org/api/4/keys/all?query=colour&sortname=count_all&sortorder=desc&page=1&rp=100) and came up with a rather long regex: 
```
/^(?>(?:roof|building(?>:(?>roof|facade))?|seamark:[a-z0-9_:]+|(?:cycleway:[a-z0-9_:]+:)?surface|(?>int_)?ref|destination|light):)?colour(?>(?>_(?>bg|tx)|:(?>arrow|back|bonnet|cap|text)))?$/
```
Which matches roughly 50% of these keys and some unrealistic fake ones (added by me at the end) - You can experiment yourself: http://rubular.com/r/nAjRBYjvUZ
(Please note that in the above regex the ```(?: ... )``` and ```(?> ... )``` are non-capturing and atomic-grouping performance optimisations, as we don't need individually matched components.)

Or we can stick to the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and use simple (cheap to execute and easy to maintain) regex:
```
/colour/
```
as the first filter and indication that user might have wanted to tag a colour value, and only then we try to interpret that given value as a colour.

Zverik: Thanks for trying different options. The long regular expression looks interesting, but it is not future-proof. I'd propose to use `/^(?:.+:)?colour$/` so that only `colour` or `whatever:colour` tags are matched.

I don't like variants with more suffixes, like `colour_tx` — that will catch wrong tags like `:colour_pattern` or apply only to local non-discussed tagging. Leading underscores are only used only in undocumented rare tags, so let's look only for semicolons.

stefanb:
In 1fc4fb9f6384d1ad9151910abe7bb3a0cd7d0057 I have simplified the RegEx as suggested by @Zverik.

gravitystorm:
I'm still not keen on accepting arbitrary keys, as per my comment on the earlier diff. Yes, maintaining lists of "accepted" tags is hard, and not something that can be easily automated ("top 100" or "more than 100 uses" or "has a wiki page" etc). But such is life with OpenStreetMap, and editing software and rendering software has to make similar decisions, particularly if they are concerned (as I am) about reducing tag fragmentation and being part of the mapper feedback loop.

stefanb:
@gravitystorm, current solution is not "accepting" any key, it is merely taking a hint from the key that the editor might have tried to tag some colour and tries to visualize it if the value is a valid colour.

As I wrote above in the code reviewing discussion:
> Key validation on the other hand is performed in the key column of the sidebar. Currently the only visual indicator is the presence of the link to wiki. That link is unfortunately missing for many valid keys (eg when key legitimately contains some variable or the wiki page is not existing or redirecting to a more general key). There might be also cases when the key was deprecated, but the link still exists because the page remained in wiki for reference. Key validity indication could be improved, to not just existence of wiki page but
> 
> also some other metadata from wiki (redirect, status...) or
> link to most suitable wiki page (eg it the key is "building:levitating" which does not exist in wiki, just link to the first part "building" to the "key:building" wiki page which would give editors more hints about the topic and how to fix the tagging) or
> OSM data (eg order tags by key occurance count)
> some keys are only valid on specific elements (relations, ways, nodes),
> some keys are requiring presence of related tags (eg drink:* tag requires appropriate shop or amenity tag)
> Key validation is a complex topic to be tackled separately.

and

> It would not be clear to users to know why they don't see the colour preview box for their mispelled "buidling:colour=#aaa" tag. Absence of color preview box may say that something is wrong, but it would not point to the key, but would most likely leave user wondering whether should it be "#aaaaaa" or "lightgrey" or "lightgray" or "paledesertwhite" or ...?

TL;DR: Key "validation" (or better: level of approval) should be indicated next to the key in the left column of the browser sidebar.

Zverik:
Isn't validation better left to sites like Osmose and KeepRight? Maintaining them, updating to reflect new tagging schemes and generallay improving is much easier than this one.

stefanb:
I have added a ticket #1782 to improve indicating the key correctness and linking to wiki in the key column of the browser sidebar.

mmd-osm:
Tried to test this coding locally, but somehow I'm getting escaped html. Not exactly sure where this comes from. Test on 701334bcc6f9cd31831fc8e62b1321ded4491c97 + merged stefanb/tag-colour-preview (1fc4fb9f6384d1ad9151910abe7bb3a0cd7d0057) into it.

grafik

stefanb:
@mmd-osm tnx for noticing. In 23fa74823a66df95330f75f9bfd5cb2bffcb2505 i have fixed that by marking the html markup string as `html_safe` to prevent ruby from html escaping it.
It works:
image

...but rubocop [complains](https://travis-ci.org/openstreetmap/openstreetmap-website/builds/355233870) with:
```
app/helpers/browse_helper.rb:80:183: C: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
      %( <div class="colour-preview-box" style="background-color:#{h(value)}" title="#{h(t('browse.tag_details.colour_preview', :colour_value => colour_value))}"></div>#{h(value)} ).html_safe
```
Not sure how to fix that properly. @tomhughes, is it ok to add the exception to the whole file to `.rubocop*.yml` or is there a better way (adding exception for just this line or preventing the unwanted html escaping in some other way)?

tomhughes:
@stefanb Well the issue isn't really disabling the cop, it's using `html_safe!` in the first place - that should be an absolute last resort to be used only if there is no way to do the job properly.

tomhughes:
To be clear in this case you can almost certainly do the job properly by constructing the HTML with ruby element helper methods instead of just embedding it as a string.

tomhughes:
On top of which it has inline styles (not acceptable and soon likely to be banned by security policy) and is using a `div` where I suspect a `span` is more appropriate given I doubt we're in a block context where that is being used.

Basically that whole line is a massive trigger.

tomhughes:
That's worse really - inline disablements clutter the code up.

stefanb:
All mentioned problems should now be fixed properly. Thanks for the valuable hints.
There is no sane workaround for inline style in this case, so that will have to be handled using a `nonce` during CSP implementation.

tomhughes:
Our CSP does currently allow inline styles but I think only because I haven't got around to tracking them all down and replacing them yet.

I'm not sure if a nonce will work - will need to look it up. There are certainly other ways though, like a data attribute that triggers javascript to set a style.

HolgerJeromin:
> I'm not sure if a nonce will work - will need to look it up. There are certainly other ways though, like a data attribute that triggers Javascript to set a style.

Smells like a workaround for a unwanted lint error.

tomhughes:
What are you on about? A CSP policy violation is not a "lint error".

stefanb:
But still, CSP has `nonce`s for exactly such purposes to allow some specific use instead of general `unsafe-inline` directive. 
Resorting to javascript is a possible workaround, but also has to be done properly (eg to not just `eval()` the data attribute :) )

And of course, it also depends on the framework's support for CSP and how easy it is to use `nonce`s.

secure_headers gem seems to support this just fine, but as soon as the `nonce` is present in CSP `style-src` browsers will stop allowing `unsafe-inline`, breaking the styling in such places, so it needs to be done trough-out the website at once or have the CSP in report-only mode for a while until all instances are tackled.

tomhughes:
I know the nonce will work in principle - the question is whether there is a good way for use to use one that doesn't negate the purpose. We almost certainly can't just commit the nonce in this repository for example as that would defeat the whole purpose.

So at the very least it probably needs to be a configuration option that can be set from chef. Even then I'm not sure that it helps much given a bad actor can just access the source and copy the nonce.

mmd-osm:
After some testing, the new colours somehow feel a bit, well how should I put it, annoying. Please don't get me wrong, I still find the idea cool, yet it could be made bit less obtrusive, i.e. giving a user some indication there's some colour without adding too much distraction.

The only idea I could come up with was to use different opacity value depending on where the mouse is. This may be a stupid idea to start with, and add even more confusion for a normal user. Maybe other don't find it distracting as it is right now?

grafik

Opacity value of 0.1 instead of 0.4:

grafik


When focusing on the key/value pair block, the colours are shown with opacity 1 again, like before.

grafik



```
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1219,9 +1219,17 @@ tr.turn:hover {
       margin: 0px;
       border: 1px solid rgba(0, 0, 0, .1); 
       // add color via inline css on element: background-color: <tag value>;
+
+      transition: opacity 0.1s ease-out;
+      opacity: 0.4;
     }
   }
 
+  .browse-tag-list:hover .colour-preview-box {
+     display:block;
+     opacity: 1;
+  }
+
   .warning {
     margin: 0 0 $lineheight/2 0;
     padding: 0 $lineheight/2;
```

HolgerJeromin:
Perhaps make the div smaller.
Opacity changes the color. On mobile we have no :hover

stefanb:
@tomhughes: "nonce" = number used _once_, not a long term secret configured somewhere in chef

@mmd-osm: you made a good point with overly-intrusive colours, but in reality most building colours are not as vivid/saturated as in our test cases, but usually much more pale. Having it shown more pale than the actual value might encourage users to enter more aggressive/vivis/saturated colours, so that they would look nice in the browser.

@HolgerJeromin: valid point about mobile not having hover, here is a sample with smaller boxes: size 12x12px, same as 12px font for in the table, with 4px margin on top and bottom to keep them vertically centered:
image
Would that be better?

stefanb:
Replaced the inline style attribute with data attribute + javascript and made the boxes smaller (same size as text)

stefanb:
@tomhughes, any other concerns?

gravitystorm:
I've been reviewing this PR today, and despite my [unresolved objections](https://github.com/gravitystorm) I'm willing to go ahead and accept this PR.

However, in the intervening time the PR has bit-rotted and now has conflicts, mainly due to the [split of browse_helper and browse_tags_helper](https://github.com/openstreetmap/openstreetmap-website/commit/74d2c4336bab2efe44c6bdadf688d26b18e320b3) last year. @stefanb if you're willing to rebase this onto a recent master, that would be helpful. If not, no worries, I can do that in a few weeks time.

stefanb:
@gravitystorm thanks for picking this up, I can check it once i restore my OSM dev environment on a rainy day :)

stefanb:
@gravitystorm, I have resolved the conflicts and fixed the code style problem that appeared wit more strict linting during the building process. Please review and test it.

gravitystorm:
@stefanb Great, thank you! I've merged this now. I look forward to seeing it on real-life data, to see if there's any iterations we need to make.

Thanks again for rebasing and making the linting changes to go with it.

stefanb:
Thanks for checking, merging and deploying!

Some examples:
https://www.openstreetmap.org/way/176487479 (building)
https://www.openstreetmap.org/relation/59655 (road route relation)

PR #1795, opened 2018-03-28
Silence image_optim missing worker warnings during tests

gravitystorm:
There's two parts to this - firstly, removing the forced loading of the config file, and letting image_optim load the config itself. Secondly, this then enables per-environment image_optim config files, and so we can disable the various workers in the test environment.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -43,8 +43,5 @@ module OpenStreetMap
       config.logstasher.logger_path = LOGSTASH_PATH
       config.logstasher.log_controller_parameters = true
     end
-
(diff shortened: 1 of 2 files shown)


tomhughes:
After looking at the `image_optim` source I added f9df59a8a49d2aed057eaee0b7249eea329d34c5 to simplify things - when there is a base config it reads that and then merges the environment specific configuration so only the things that need to be changed need to be included.

tomhughes:
Ignore that... I was talking nonsense and have removed it.

PR #1796, opened 2018-03-28
Refactor routes.rb

gravitystorm:
This PR refactors some of the routes.rb file. I aim to slowly move towards standard "resourceful" routing, but that will take a while and I think is best done gradually. This PR includes a few general refactorings that touch on larger parts of the file.


tomhughes:
What was the basis for considering an alias unnecessary? Is it that those routes were already getting that name implicitly? or just that we're not using the name anywhere?

gravitystorm:
They were already getting that name implicitly.

Issue #1163, opened 2016-02-26
»My« prefixes unnecessary (»My messages«, »My profile«, »My edits«, etc.)


jancborchardt:
In the user menu on the top right as well as in the tab navigation bar in the user profile, every single item has the prefix »My«. See this (it’s French, but same thing):
![capture du 2016-02-26 13-51-04](https://cloud.githubusercontent.com/assets/925062/13352789/08023f62-dc90-11e5-8cc9-bb0de507bad2.png)
Also lower down the section is called »Your friends«.

These prefixes are unnecessary and should be removed because they have several drawbacks:
- they make the items significantly longer to scan
- entries look very similar because they all start with the same word
- they take up more space which can result in layout problems depending on the language
- it’s pretty obvious that a dropdown with my name on it will have »my« settings. Probably not other people’s settings. Same for my own profile. Yeah, I will probably see »my« edits. ;)
- it just feels plain weird nowadays to have these prefixes used. Check this article about the topic on UXmatters: http://www.uxmatters.com/mt/archives/2009/07/generating-ideas-your-versus-my-in-user-interfaces.php#wording

d1g:
+1 

just note: Firefox and Chrome doesn't use "my settings" idiom

HolgerJeromin:
There are translations without that issue. For example German.

mcepl:
Czech translation has the problem as well.

GaspardPO:
What do we do with this issue?
Should we close it, or should we remove "My" in labels?
(and all the translations)

PR #1799, opened 2018-03-28
Be more consistent with 'my' vs 'your'

gravitystorm:
...for objects that belong to the user. We use 'my' almost everywhere, and this PR takes care of a few that use 'your' inconsistently.

I'm not making any comment on whether 'my' or 'your' or 'neither' is the best (see #1163), just trying to be consistent.


tomhughes:
I agree that being consistent is a minimal requirement so I have merged this.

That said I have to say that some of the results seem very ugly to me and I'm pretty sure I prefer "my" over "your" here with the caveat that in fact I think the prefix should be removed altogether where it's not necessary to distinguish what is meant, as suggested in #1163.

PR #1800, opened 2018-03-29
Unlocateable place now highlighted in red and specified in error message

### LINKED ISSUE #1790: Directions: "Sorry - couldn't find that place": Say WHICH place
Kindly when ever saying "Sorry - couldn't find that place",
be sure to say if it was the From place,
or if it was the To place that was bothering it.
And / or print the place that was causing the trouble.'
No it is NOT always 100% obvious.
We might have tried entering various things in both boxes etc.

jguthrie100:
Fixes #1790 
Added a new locale translation where you can specify a specific location in the 'not found' error message
Made it so that the textbox containing the error-ed value gets highlighted in red

image

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -53,6 +53,10 @@ OSM.Directions = function (map) {
       }
     });
 
+    input.on("keydown", function() {
+      input.css("background-color", "#fff");
+    });
+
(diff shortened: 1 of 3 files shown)


tomhughes:
Merged with a few additional tweaks in 7e6e1dfbaf2d9112aa0c86a2546fa1780dd69023.

PR #1802, opened 2018-03-31
Converted roundabouts to take "1st, 2nd, 3rd" exit etc

jguthrie100:
Made a change to OSRM routing directions to say `Take 1st exit onto xx road`, rather than `Take exit 1`, which is a bit more applicable to how one would word the instructions in real life (and matches the OSRM template)


mmd-osm:
This doesn't really work for non English languages, ordinal numbers need their own i18n rules. Here's a short example for German:

Current text: `Im Kreisverkehr Ausfahrt 2 auf xyz nehmen`
Changed in this PR to: `Im Kreisverkehr 2nd Ausfahrt auf xyz nehmen`
Correct would be: `Im Kreisverkehr 2. Ausfahrt auf xyz nehmen`

tomhughes:
Yes this probably needs to be done with `%{count}` instead of `%{exit}` in the string and then subkeys can be used to do the correct translations.

jguthrie100:
I tried using the `%count` technique, but its just for simple pluralisations (zero, one, other), so I've been a bit hacky and added a list of 1st, 2nd, 3rd values in the locale file and a numToWord function in the directions JS file. I couldn't think of a better way to implement it.

Also stopped at 10. If a roundabout has 11+ exits, it'll just omit specifying the exit number and just specify the road name instead, but I can't see any roundabouts having more than 10 exits!

gravitystorm:
It's interesting to note that ["ordinalizing" numbers has just received a patch in rails](https://github.com/rails/rails/pull/32168) - obviously we would need to wait until it's in a released version, but it's worth beaning in mind that there's now a standard approach to this problem.

tomhughes:
Well we would need it in i18n-js anyway, rather than in the server side ruby version of i18n.

gravitystorm:
> Well we would need it in i18n-js anyway, rather than in the server side ruby version of i18n.

Ah, good point.

tomhughes:
I think the main problem is that if a roundabout has more than 10 exits this will wind up saying "11 exit" rather than "exit 11". The correct grammar (in English at least) is different when using ordinals vs non-ordinals.

jguthrie100:
With 11 exits it would just pass in an empty string, so the message would be `"Take exit onto High Street"`

jguthrie100:
I've changed it now so that with 10 or fewer exits the new `exit_ordinal` locale string is used, where if there are 11+ exits, the original exit string is used

PR #1806, opened 2018-04-02
Allow searching via UK GSS code.

EdwardBetts:
The UK Office for National Statistics maintains a series of codes to
represent geographical areas of the UK. These codes follow a standard
format and are recorded in OpenStreetMap using the ref:gss tag.

This change lets map users search the map by UK GSS code. It uses the
Overpass API to look for a relation with the code provided by the user.

Example search: E06000001 will return Hartlepool (unitary authority).

### DIFF
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -23,6 +23,8 @@ class GeocoderController < ApplicationController
       elsif @params[:query] =~ /^[A-Z]\d[A-Z]\s*\d[A-Z]\d$/i
         @sources.push "ca_postcode"
         @sources.push "osm_nominatim"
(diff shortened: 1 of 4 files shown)


HolgerJeromin:
This should be integrated into nominatim

EdwardBetts:
@HolgerJeromin The team behind nominatim are not keen on search by identifier, they think that a database lookup is more appropriate than using a geocoder. See Nominatim#714 for example.

csmale:
Be aware that the GSS codes can be volatile. If there is a boundary change for any reason, the areas concerned will be issued with new codes. Changes which do not affect the boundary (e.g. name changes) will not result in a new code however.

tomhughes:
Sorry, not only is this country specific (and we don't really want to be adding more country specific search methods) but it's also incredibly niche so we won't be adding this.

PR #1805, opened 2018-04-01
Fix destinations and roundabouts locale

jguthrie100:
Done a few changes here.

I refactored the with_destination locale so its just a string that one can append onto the end of an instruction. Hopefully it will still work with different grammar structures in different languages!

I also fixed some ambiguous roundabout instructions..
i.e. https://www.openstreetmap.org/directions?engine=osrm_car&route=53.7338%2C-1.5258%3B53.6832%2C-1.4999#map=13/53.7086/-1.5171

Instructions 7-16 are a bit weird - so I've fixed it to specify `"Exit straight/slight right/right from roundabout"` when exiting roundabouts

PR #1808, opened 2018-04-02
Fix roundabout exit text

jguthrie100:
This is basically a re-commit of #1805 , except I've removed the bit that separated out the `%{destinations}` template.

It now basically just fixes the ambiguous roundabout exit templating and refactors some of the code that builds the template name
i.e. https://www.openstreetmap.org/directions?engine=osrm_car&route=53.7338%2C-1.5258%3B53.6832%2C-1.4999#map=13/53.7086/-1.5171

(See instructions 7-16 - its not clear that every other instruction is an instruction to _leave_ the roundabout!)

### DIFF
--- a/app/assets/javascripts/index/directions/osrm.js
+++ b/app/assets/javascripts/index/directions/osrm.js
@@ -130,7 +130,7 @@ function OSRMEngine() {
 
         if (namedRoad) params.name = name;
         if (Object.keys(params).length > 0) {
-          template = template + "_with_" + Object.keys(params).join("_");
+          template = template + "_with_" + Object.keys(params).sort().join("_");
         }
         instText += I18n.t(template, params);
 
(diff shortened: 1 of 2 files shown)


tomhughes:
I can't even begin to think how to validate that this is sensible - the changes are so significant that it's going to be very hard to reason about.

I'm also not at all sure that it makes sense to talk about "slight left" off a roundabout. It's not normal even to talk about left/right when leaving a roundabout - all navigation systems I have ever encountered just tell you to take the nth exit or, when on the roundabout, to take the next exit.

jguthrie100:
Basically the mass of changes is simply renaming `instruction_without_exit` to `instruction_with_name`, and then copying your code (that builds up the `ìnstruction_with_exit_directions` etc)

As for the `slight left` and `turn right off roundabout`, at the moment it says the same thing, except much more confusingly!
At the moment it says `At roundabout continue straight onto A650` - All I have done is change the text to read `Exit straight onto A650`.

OSRM says `Go straight onto A650`.

Either way, having `At roundabout take exit 2`, followed immediately by `At roundabout continue straight onto A650` is a really confusing way of telling you to enter and leave the roundabout! It sounds like its two separate roundabouts

jguthrie100:
Would it be better if I cancelled all the refactoring bits and just reworded the roundabout text?

PR #1809, opened 2018-04-02
Reworded roundabout exit text

jguthrie100:
I've cut back all the complicated refactoring from the PR I've just closed, and just left in the roundabout wording changes.

Basically I've just reworded the instructions to be the same as whats used on OSRM..

See: https://www.openstreetmap.org/directions?engine=osrm_car&route=53.7320%2C-1.5165%3B53.7254%2C-1.5161

vs: http://map.project-osrm.org/?z=17&center=53.730483%2C-1.512321&loc=53.731819%2C-1.514997&loc=53.728976%2C-1.512133&hl=en&alt=0

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2377,11 +2377,11 @@ en:
         against_oneway_without_exit: Go against one-way on %{name}
         end_oneway_without_exit: End of one-way on %{name}
         roundabout_with_exit: At roundabout take exit %{exit} onto %{name}
-        turn_left_with_exit: At roundabout turn left onto %{name}
-        slight_left_with_exit: At roundabout slight left onto %{name}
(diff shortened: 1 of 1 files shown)


tomhughes:
Unless I've misunderstood when these instructions are issued this change doesn't make any sense to me. Removing "at roundabout" is a significant regression in terms of the information being imparted.

tomhughes:
So having looked at your example it is clear that "at roundabout" is wrong for the second case, but so really is "straight on".

The first case looks correct to me - in an actual turn by turn navigation system it would presumably be read out as you approached that point and then "at roundabout take the 3rd exit makes perfect sense".

The second case I would expect to be something like "take the exit towards X" and talking about going straight on makes no real sense with or without "at roundabout".

PR #1813, opened 2018-04-07
Update xid_to_int4.c

dim3dro1:
Made changes to avoid error when executing "CREATE FUNCTION xid_to_int4(xid) RETURNS int4 AS '`pwd`/db/functions/libpgosm', 'xid_to_int4' LANGUAGE C STRICT".

### DIFF
--- a/db/functions/xid_to_int4.c
+++ b/db/functions/xid_to_int4.c
@@ -1,10 +1,13 @@
 #include <postgres.h>
 #include <fmgr.h>
 
-int xid_to_int4(TransactionId xid)
+Datum
+xid_to_int4(PG_FUNCTION_ARGS)
 {
-   return xid;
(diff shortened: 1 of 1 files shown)


tomhughes:
More details please - what sort of error?

If you're referring to the error in postgres 9.7 in later then that is known about and can't be fixed by this or anything else which is why we are stuck on 9.6 until we re-engineer things.

tomhughes:
Sorry I meant 9.6 or later - we are stuck on 9.5 as a result.

pnorman:
Wouldn't the proposed changes substantially change xid_to_int4 and break it, no longer converting an xid to int4?

tomhughes:
Well I have no idea what they do, but we know that function is doing is not a good idea anyway. Basically if you want to do incremental dumps (which is the only reason for needing that function) you need to stick to postgres 9.5 or earlier.

Issue #1807, opened 2018-04-02
Add OpenLocationCode to OSM website


bjohas:
Mobile OSM apps (e.g. OSMAnd) already support OpenLocationCode, as does Google maps, and it would be great if the OSM website did too. There's a Ruby implementation, https://github.com/google/open-location-code/blob/master/ruby/lib/plus_codes/open_location_code.rb. 

I.e. a search for 6PH57VP3+PQ, https://www.openstreetmap.org/search?query=6PH57VP3%2BPQ should return the area indicated by the OLC.

tomhughes:
No, just no.

We've refused many such codes in the past and I don't see how this is any different.

bjohas:
I agree that OSM cannot support every single code under the sun. However, I'd say it's different, because OLC is already supported by OSMAnd, on the roadmap for Maps.Me (AFAIK), and supported by Google Maps (website and app). So users may come to expect it. Plus it's really simple to implement, and (unlike many other codes) fully open.

Firefishy:
I think OpenStreetMap.org search should support PlusCodes. To be more blunt, we should actively try support these Open schemes to reduce the value of the close commercial alternatives.

@tomhughes OK if we re-open this ticket?

openstreetmap-website:
That is all fine and dandy, but not the OSMFs remit. And even if it was, then why start with a not really important service? Obviously we should start with putting all commercial map tile services out of business.

Am 30. September 2018 23:03:19 MESZ schrieb Grant <notifications@github.com>:
>I think OpenStreetMap.org search should support PlusCodes. To be more
>blunt, we should actively try support these Open schemes to reduce the
>value of the close commercial alternatives.
>
>@tomhughes OK if we re-open this ticket?
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/openstreetmap/openstreetmap-website/issues/1807#issuecomment-425751867

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit Kaiten Mail gesendet.

Firefishy:
> Obviously we should start with putting all commercial map tile services out of business.

I should have been clearer, I believe that location data (address, coordinates) should fundamentally be open and it is morally objectionable for a company to try close them up for commercial advantage.

systemed:
> [@simonpoole] That is all fine and dandy, but not the OSMFs remit.

[...]

PR #1818, opened 2018-04-11
Added PlusCode functionality to search

jguthrie100:
Implements #1807

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -86,6 +86,9 @@ gem "faraday"
 # Load geoip for querying Maxmind GeoIP database
 gem "geoip"
 
+# Load plus_codes to generate OpenLocationCodes
+gem "plus_codes"
+
 # Load rotp to generate TOTP tokens
 gem "rotp"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -218,6 +218,7 @@ GEM
     parser (2.5.0.5)
       ast (~> 2.4.0)
     pg (0.21.0)
+    plus_codes (0.2.1)
     poltergeist (1.17.0)
       capybara (~> 2.1)
       cliver (~> 0.3.1)
@@ -392,6 +393,7 @@ DEPENDENCIES
   openstreetmap-deadlock_retry (>= 1.3.0)
   paperclip (~> 5.2)
   pg (~> 0.18)
+  plus_codes
   poltergeist
   psych
   puma (~> 3.7)
@@ -415,4 +417,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.15.4
+   1.16.1
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -2,6 +2,7 @@ class GeocoderController < ApplicationController
   require "cgi"
   require "uri"
   require "rexml/document"
+  require "plus_codes/open_location_code"
 
   before_action :authorize_web
   before_action :set_locale
@@ -23,6 +24,8 @@ class GeocoderController < ApplicationController
       elsif @params[:query] =~ /^[A-Z]\d[A-Z]\s*\d[A-Z]\d$/i
         @sources.push "ca_postcode"
         @sources.push "osm_nominatim"
+      elsif PlusCodes::OpenLocationCode.new.valid?(@params[:query])
+        @sources.push "plus_code"
       else
         @sources.push "osm_nominatim"
         @sources.push "geonames" if defined?(GEONAMES_USERNAME)
@@ -236,6 +239,29 @@ class GeocoderController < ApplicationController
     render :action => "results"
   rescue StandardError => ex
     @error = "Error contacting api.geonames.org: #{ex}"
+    render :action => "error"
+  end
+
+  def search_plus_code
+    # get query parameters
+    code = params[:query]
+
+    olc = PlusCodes::OpenLocationCode.new
+    lat_lon = olc.decode(code)
+
+    @results = []
+    @results.push(:lat => lat_lon.latitude_center, :lon => lat_lon.longitude_center,
+                  :zoom => PLUSCODE_ZOOM,
+                  :name => params[:query])
+
+    render :action => "results"
+  rescue StandardError => ex
+    unless olc.full?(params[:query])
+      @error = "#{params[:query]} is not a full Plus Code address"
+    else
+      @error = ex.message
+    end
+
     render :action => "error"
   end
 
--- a/config/example.application.yml
+++ b/config/example.application.yml
@@ -38,6 +38,8 @@ defaults: &defaults
   postcode_zoom: 15
   # Zoom level to use for geonames results from the geocoder
   geonames_zoom: 12
+  # Zoom level to use for pluscode results from the geocoder
+  pluscode_zoom: 16
   # Timeout for API calls in seconds
   api_timeout: 300
   # Timeout for web pages in seconds
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -373,6 +373,7 @@ en:
         geonames: 'Results from <a href="http://www.geonames.org/">GeoNames</a>'
         osm_nominatim_reverse: 'Results from <a href="https://nominatim.openstreetmap.org/">OpenStreetMap Nominatim</a>'
         geonames_reverse: 'Results from <a href="http://www.geonames.org/">GeoNames</a>'
+        plus_code: 'Result from <a href="https://plus.codes/">Plus Code</a>'
     search_osm_nominatim:
       prefix_format: "%{name}"
       prefix:
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -252,6 +252,7 @@ OpenStreetMap::Application.routes.draw do
   get "/geocoder/search_geonames" => "geocoder#search_geonames"
   get "/geocoder/search_osm_nominatim_reverse" => "geocoder#search_osm_nominatim_reverse"
   get "/geocoder/search_geonames_reverse" => "geocoder#search_geonames_reverse"
+  get "/geocoder/search_plus_code" => "geocoder#search_plus_code"
 
   # directions
   get "/directions" => "directions#search"
--- a/test/controllers/geocoder_controller_test.rb
+++ b/test/controllers/geocoder_controller_test.rb
@@ -33,6 +33,10 @@ class GeocoderControllerTest < ActionController::TestCase
       { :path => "/geocoder/search_geonames_reverse", :method => :get },
       { :controller => "geocoder", :action => "search_geonames_reverse" }
     )
+    assert_routing(
+      { :path => "/geocoder/search_plus_code", :method => :get },
+      { :controller => "geocoder", :action => "search_plus_code" }
+    )
   end
 
   ##
@@ -420,6 +424,19 @@ class GeocoderControllerTest < ActionController::TestCase
       results_check :name => "England", :suffix => ", United Kingdom",
                     :lat => 51.7632, :lon => -0.0076
     end
+  end
+
+  ##
+  # Test the plus code search
+  def test_search_plus_code
+    get :search_plus_code, :xhr => true,
+                                  :params => { :query => "6PH57VP3+PQ" }
+    results_check :name => "6PH57VP3+PQ",
+                  :lat => 1.2868125, :lon => 103.85443749999999
+
+    get :search_plus_code, :xhr => true,
+                                  :params => { :query => "X4HP+M5" }
+    results_check_error "X4HP+M5 is not a full Plus Code address"
   end
 
   private


jguthrie100:
Sorry, didn't see: https://github.com/tomhughes/openstreetmap-website/commit/2e0a2c67caf64df732f1e14160d5ead96c73a656

tomhughes:
Sorry yes I have already knocked this, but I still haven't decided whether to offer it up or not as people are divided on whether it's a good idea.

jguthrie100:
Seems like a good idea to me. Helpful to people who use PlusCodes and totally non-disruptive for people who don't..

govvin:
I would find this useful. OpenLocationCodes (or PlusCodes, as Google would like to call it) is already supported in the OsmAnd app.

sommerluk:
> before you know it, 13 states in Africa have switched their addressing to the proprietary vendor

A valid concert. Many African countries do not have a working address system in use. Most streets don’t have names and even when they have names, the houses don’t have house numbers. There are some efforts for example in Benin, with financial support from France, to create public address systems ([street names + house numbers based on mathematical measurement](https://wiki.openstreetmap.org/wiki/FR:WikiProjet_B%C3%A9nin#Le_nouveau_syst.C3.A8me_d.E2.80.99adressage_de_Cotonou)) but they are limited to some of the bigger cities. On the other hand, online-ordering and logistics are growing a lot currently in these countries, including small villages. So the post company of Ivory Coast is working since two years with what3words, which is proprietary and creates a vendor-look-in and can only be used by authorized apps and so on – something that in my opinion is not really helpful for the development in Ivory Coast.

jguthrie100:
> something that in my opinion is not really helpful for the development in Ivory Coast.

From a wholly academic perspective its not very useful, but I bet for the guy sitting in his house in the middle of a slum on the outskirts of Abidjan, and the postman looking to deliver his new microwave, that its an absolute godsend, whether or not its proprietary or not!

If its so bad for public services to use proprietary technology then its up to the opensource standards community to come up with a better alternative, of which I'm not sure there are any that can rival what3words!

woodpeck:
The proprietary systems are not brilliant inventions or engineering feats; the only thing that sets them apart is their marketing and sales budget. Which they have precisely because they promise big profits down the line. Profits that the happy guy with his microwave will be paying for, one way or the other. Hence any rules about adding address/geocoding providers we come up with *must* include an open source requirement so that we don't become complicit to exploitation.

dieterdreist:
sent from a phone

> On 13. Jul 2018, at 13:33, Jamie Guthrie <notifications@github.com> wrote:
> 
> of which I'm not sure there are any that can rival what3words


so clearly you have not yet heard about WTF

sommerluk:
> a better alternative, of which I'm not sure there are any that can rival what3words!

Well, in countries where much more than half of the territory do not have neither Internet coverage nor even mobile phone coverage and many villages not even electric power supply (they charge their smartphones – yes, they have smartphones – with photovoltaic) PlusCodes seem to be clearly the better choice than what3words – also from a technical point of view: No need for neither a 10 MB word database that works only for a particular language nor for an Internet connection. Instead light code that works also on older smartphones.

sabas:
I attended the Plus Codes presentation at SotM by @drinckes (there was also @janetchapman); it could be good to support this coding since it has an open license and as far as I understood is used in some places.
Apart the 'long' version,  I wonder if it would be worth to support also the short version (this would need to resolve the name part to nominatim to calculate the area code I think).

It's already supported in GMaps and Maps.me.

Vaomatua:
The potential for Plus Codes is great, the addresses for every location on earth already exist, they can be derived offline, are concise, and recognizable. I've looked at many systems and feel that Plus Codes is the first one I've studied that makes sense anywhere in the world.
The Tanzania Development Trust has calculated the Plus Codes for 17 million building points in Tanzania.  We did a test upload in the [Zeze village](https://www.openstreetmap.org/changeset/59213224) as a way to see how it could work in OSM.  

zeze

woodpeck:
Frankly, this is the worst idea I have heard in years. Plus codes can be computed from coordinates, and coordinates can be computed from plus codes. The way to add plus codes to OSM is *not* adding plus code tags to objects in OSM (and you have added THREE of them). What you have done here is the informational equivalent of adding a "latitude" and "longitude" tag to every building. Please do not do that again.

Vaomatua:
Interesting perspective @woodpeck . If the tools existed to help people in Africa communicate their location easily this wouldn't be necessary. As it is, this is a better option that anything that exists, and certainly better than doing nothing.

woodpeck:
I think we may be miscommunicating. Can you explain in which way the plus codes that you have imported to OSM are used by the people in Africa? Are you loading the data from OSM onto a device that is not plus code capable, or are you printing the plus codes on maps, or what is the use case? What I want to say is that there is no use case that *I* can think of where it makes sense to have the data *in OSM*.

jguthrie100:
What OSM needs really is a Right Click -> Get Lat/Lon and Right Click -> Get PlusCode

dieterdreist:
sent from a phone

> On 10. Aug 2018, at 00:20, Jamie Guthrie <notifications@github.com> wrote:
> 
> What OSM needs really is a Right Click -> Get Lat/Lon and Right Click -> Get PlusCode


you get lat/long when you choose “show address” after right click

jguthrie100:
If PlusCodes get implemented, then the PlusCode should be dropped in there too then!

RobJN:
> The proprietary systems are not brilliant inventions or engineering feats; the only thing that sets them apart is their marketing and sales budget. Which they have precisely because they promise big profits down the line.

Um, you've lost me on that one @woodpeck. What bid profits? "Plus codes are free. There are no licensing fees or other costs. The technology is open-sourced." This isn't W3W!

As @drinckes says: It takes a lot of money to countries to go about naming all their roads. Plus codes is available for free now (and always) and it doesn't prevent a government from naming all their roads and creating a postal address system when time and money allows. 

See Doug's full talk from SotM 2017 here:
https://2017.stateofthemap.org/2017/open-source-street-addresses/

simonpoole:
@RobJN plus codes are calculated from the coordinates, it doesn't make any sense in any conceivable way to store them in the DB and it is just as silly as storing the coordinates as tags on the objects would be  (which has been known to happen).

The other issue is that they use Latin script letters which is a non-starter for anything intended for global use.

RobJN:
Yeah, don't put them in the database as key=value tags. Implement the ability to search them in Nomination (as per the PR). It sounds like OSMAnd, MapsMe and GoogleMaps have already done this. Why not Nominatim too?

> The other issue is that they use Latin script letters which is a non-starter for anything intended for global use.

Um, but so does every url (or has that changed now). Sounds like an excuse in my opinion.

RobJN:
Also a good tweet on addresses (or the lack of them) from SotM:

> Geoffrey [@kateregga1] from @mapuganda sharing about mapping in Northern Uganda. Refugees describe their address based on the nearest water tank “I live in tank 11”. #sotm
> https://twitter.com/mapkibera/status/1023913370972372992

I don't know if @kateregga1 saw the Plus Codes lightning talk (or the [2017 talk](https://2017.stateofthemap.org/2017/open-source-street-addresses/)), but it would be nice to get his view as plus codes seem like they could help improve that situation.

simonpoole:
@RobJN why would you "Implement the ability to search them in Nomination" that makes 0 sense, if you have a plus code you simply convert it to the grid square/centroid coordinates no searching necessary or even possible. 

Another way of looking at it is simply a coded format for coordinates. with reduced precision. 

See https://en.wikipedia.org/wiki/Internationalized_domain_name for i18n domain names, a different problem, but not an excuse to repeat the same mistakes.

RobJN:
> Why not Nominatim too?

Sorry I got this wrong. It looks like the PR is to add them direct to the OSM.org search as a new provider alongside nominatim. My bad for not understanding the technical side properly.

sabas:
Sarah has said this shouldn't be implemented by Nominatim, see https://github.com/openstreetmap/Nominatim/issues/914

drinckes:
Hi - I'm the lead of the plus codes project. I was at SotM this year in Milan but had to unexpectedly leave early, I was hoping to have some of this discussion in person but unfortunately it didn't work out.

@woodpeck makes a good point - the plus code is just a formatting of the lat/lng so it doesn't need to be stored with the object*. What could be worth storing or associating with the object is the _locality_, because computing the "best" locality associated with a point can be expensive. (It isn't necessarily the closest, because you want to respect country and state borders, and you might have a tiny village close and a large city slightly further, but the city may make more sense.)

Of course, that assumes the plus code address is the center of the object. But you might want it to indicate the doorway, or the drive way (if it's a long way off the road). In that case you will want the code separate but I agree that will be a nice to have.

@simonpoole makes the good point that plus codes are based on latin characters. We were worried about this, so did some surveys with people in different script countries, and most of them made the point that @RobJN made - basically "umm, how do you think we type www.google.com?" We launched the codes in Maps on Android a year ago, and we see high usage in non-latin script countries (arabic, cyrillic, japanese etc). At the moment, our thoughts are that there is more to be gained by having a single, consistent script for the code part, but at least we can reconsider this in the future.

I would love to see search support in OSM web, and _especially_ in geocoders (so that not everyone has to implement it in their front end). The challenge for geocoders is that they don't turn an address into a location, so much as into the best matching object in their database. A plus code like "27W4+MR Kintampo" is an exact location (8.046747, -1.742878), but geocoders return a nearby road or even just a locality.

I'd be excited to see geocoders consider this problem, not just for support of plus codes, but also to support the kinds of address that are really common in poorly addressed areas - like "50m past the bus stop", "kilometer 10, MG road" etc.

It's going to be a long time before any of us map every building in the world. It would be great if someone in Paraguay, Mali, Canada, or anywhere, can get their plus code address, tell it to someone else and that person can get to the location, using any map app, and without having to wait for someone to add their building to maps.

johnwhelan:
Since OSMand already supports them I think that deals with the religious arguments and adding them to the search would reduce the number of people adding them to the addr: fields in OSM.  There is a thread in osm-talk where Frederick has written a fairly long email outlining why adding them separately is a bad idea and I get the impression from his post adding them to the search would be the sensible approach.

So can we just enable this since it is coded already?

Thanks John

RobJN:
Over on the ML (and I stress, I have not read it all), @systemed suggests there is sufficient consensus around @tomhughes' implementation. See: https://lists.openstreetmap.org/pipermail/talk/2018-August/081220.html

@woodpeck mentioned the featured [tile layer policy](https://wiki.openstreetmap.org/wiki/Featured_tile_layers/Guidelines_for_new_tile_layers). If we assume that as a starting point (i.e. copy-paste) then we tick off all 6 of the "Must" criteria and most of the "Should" list (albeit that the Unique and Interesting aspects could be debated seemingly forever!). The last "Should" item isn't really relevant in this use case.

Based on that, can I be so bold to suggest that this is now ready for inclusion?

sommerluk:
I’m also in favour of supporting it now.

der-stefan:
By the way: radio amateurs heavily use a locator system for decades, see [Maidenhead Locator System on Wikipedia](https://en.wikipedia.org/wiki/Maidenhead_Locator_System) (also called QTH locator).
OpenTopoMap handles QTH locators for years: https://opentopomap.org/#qth=JN59MO

``<troll>``Why not map QTH locators additionally to every building in OSM?``</troll>``

pangoSE:
@drinckes Thanks for commenting and for leading the project. I agree with all your points wholeheartedly.
@der-stefan good idea lets add that too!
@jguthrie100 I agree adding it to "show adress" and updating this linktext to "Show coordinates, geocodes and address for this point"

From my perspective it would be a plus if osm.org either directly or via a geocoder support every single location code system on the planet that is open and free to implement. As long as someone is willing to implement it I say "go for it".

bjohas:
I’d very much love to have this in OSM. We’re usng OLC with health workers in Zambia, and it would be great if OSM supported this.

johnwhelan:
How are you using OLC in Zambia that means you can't use them with OSM?

OSMand for example allows them for searches.

Thanks John

bjohas:
Sure, but it means that across the organisation, it's got poor usability. E.g. people using the desktop (back in the office) cannot use them, at least not with OSM. So that makes it difficult to use OLC to communicate locations between the field and head office. E.g. a field worker communicates the location of a small settlement to the head office (via phone) then how would the head office worker get the location? Sure, it's possible, but it's a cumbersome process. We're trying to make rural health work smoother, rather than more cumbersome. Similarly, you could just have a decoding website, that feeds coordinates into OSM. Technologically it's simple, but in practice it means an extra step that just confuses things.

johnwhelan:
The information for a olc location is already encoded in OSM.  Adding it means you may have a conflict so technically using what is there already means one less thing to go wrong.

What exactly are your requirements?  How do you know the OLC to start with?  What software is used in the field and the office?  Are you using printed maps?

Thanks John

bjohas:
Sure - This is not about adding OLC to 
the OSM metadata. We have a clustering algorithm that determines clusters of houses. We label these with OLC (outside OSM), acting like a postcode, as we don't have names (nor do local health workers). So the OLC provides a logical label, that can be printed on maps, or sets of bookmarks in maps.me. I would also like maps.me/Google maps to be able to generate OLC for sharing. I think OSMand can do this. We can then collect village names over time, or communicate specific needs for villages using OLC. However, if OSM cannot look up the OLC is problematic for us.

Vaomatua:
What is OLC?
Plus Codes are the most practical scheme for communicating addresses
(locations) I have encountered in my 30 years of geospatial work.

johnwhelan:
Note to Bjoern OpenStreetMap is a database nothing else.  It doesn't look up anything.  There are various programs that can access the data.  One is OSMand, another transforms or renders the data to a visible map which is what you see on www.openstreetmap.org.  OpenStreetMap Nominatim is used to search the map.  As previously noted the code has been developed to add in to support OLC but my understanding is it has not been pushed to the published version.

OLC and Plus codes are the same thing.  It was first called Open Code Location and developed by Google Zurich I think Google now prefers it to be called Plus Codes.

RobJN:
I think Bjeoern is using "OSM" to mean the OSM.org website. In which he case his request is for the search box to allow use of OLCs/PlusCodes as per the title of this issue (on the OSM website repo).

Personally I set why I'd be happy to see this in my earlier post:
https://github.com/openstreetmap/openstreetmap-website/pull/1818#issuecomment-412689177

Since then 3 more people have started support for this (the issue as explained in the title - adding plus code tags to OSM database is a bad idea and off topic in this thread and repo).

Vaomatua:
Bjoern,
Remember that OSM is Open Street Map, not Closed Street Database
You can simply add addr:housenumber=8FXCP9J7+8G and the MAP will become
significantly more useful for places where addresses do not exist.

johnwhelan:
"Bjoern, Remember that OSM is Open Street Map, not Closed Street Database You can simply add addr:housenumber=8FXCP9J7+8G and the MAP will become significantly more useful for places where addresses do not exist."

Please read through the other posts.  There are a number of reasons why this should not be done which are listed there.  First the information is redundant which means when reentered it might be mistyped.  The others are listed in the posts above.  What Bjoern needs is the code enabling in Nominatim.

Cheerio John

simonpoole:
@bjohas could you explain how you are using plus codes, from your brief description it would seem to be at odds with its design (a fixed grid system). 

@Vaomatua that is very broken. Plus codes are a grid system, like many before and likely many after, at best they are a method of indicating a rough geographic location, they definitely are not building addresses. The only reason anybody even has this on their agenda because of a different grid system being pushed at the same markets.

bjohas:
@simonpoole Yes, it's a fixed grid, one that designates areas. So let's say we have a cluster around Mkuishi hospital. We can label this as https://plus.codes/5GRF99MW+ (which represents a small area). I.e. as mentioned above, we basically map houses first from sat imagery; we then cluster (essentially giving concave areas of usually a few houses), and we label these with OLC, such as 5GRF99MW+. Of course, the small clusters may need a couple of adjacent OLC codes if needed (as our geographically determined clusters don't map 1:1 onto OLC - in practice, it's not really an issue, as the 8-character OLC is at a good scale). We collect village names over time, and usually several of these OLCs then correspond to a certain locality, that we then enter on OSM (e.g. as a village). Does that make sense?

simonpoole:
@Vaomatua the issues with using plus codes (or other similar systems) as house numbers in OSM instead of grid cells has been discussed in depth on the mailing lists, definitely no need to repeat it here.

johnwhelan:
@Vaomatua The database is not referenced directly but indirectly by software.  

I knew them as programs you might know them as apps.  To search for a location on the web site displayed map what you enter is sent to Nominatim which returns the lat and long and the appropriate bit of the map.  The OLC or plus codes have been coded in Nominatim but have not yet been released as has been stated earlier in this thread.

If you add a tag with the OLC code unless you are using JOSM it will be invisible since Nominatim won't know about it or how to interpret it or use it.  OSMand understands OLC in relation to OSM and other software may also understand OLC codes.  For printed maps Maperitive can be used to show the OLC grid but you need to set up the rules to do it.

Bug me Johnwhelan through OSM directly.  Computer Databases cannot normally be directly read by humans.  I understand that the Oxford English Dictionary can be considered a database of words and can be read by a human but that is not what we are talking about.  You need software to get at the data.  The information needed for an OLC is already encoded in the database it just needs software to access it.

Cheerio John

bjohas:
Hello all, and thanks for comments. Referring back to  @RobJN 's post, about several people expressing interest, would it be possible to make the decision to move ahead with "the PR [which] is to add them direct to the OSM.org search as a new provider alongside nominatim". @tomhughes @jguthrie100 @simonpoole. Failing that, can it be put to a vote somewhere?

tomhughes:
Have you considered the possibility that it's Christmas and the relevant people might just be away and not be paying attention to you lot competing to see who can write the longest essay on this bug?

I was once on the verge of merging this, but it became clear that there wasn't good consensus so I stopped. I will review that when I have time but really the amount of noise here doesn't actually help at all - just repeating the same arguments ad infinitum doesn't add anything any neither does trying to stage some sort of self selecting popular vote.

I would also note for the record that at one point there really were people talking about adding OLC codes to OSM objects so it wasn't a totally mad thing for somebody to raise.

bjohas:
Many thanks all. I've started a page here: https://wiki.openstreetmap.org/wiki/Proposal_Open_Location_Code. Please contribute!

PR #1824, opened 2018-04-17
Set default_url_options for action_mailer

gravitystorm:
This saves having to repeat the same host and protocol options throughout the notifiers.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -43,8 +43,5 @@ module OpenStreetMap
       config.logstasher.logger_path = LOGSTASH_PATH
       config.logstasher.log_controller_parameters = true
     end
-
-    # Set the host and protocol for all action mailer urls
-    config.action_mailer.default_url_options = { :host => SERVER_URL, :protocol => SERVER_PROTOCOL }
   end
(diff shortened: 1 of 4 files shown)


config/application.rb
```diff
@@ -43,5 +43,8 @@ class Application < Rails::Application
       config.logstasher.logger_path = LOGSTASH_PATH
       config.logstasher.log_controller_parameters = true
     end
+
+    # Set the host and protocol for all action mailer urls
+    config.action_mailer.default_url_options = { :host => SERVER_URL, :protocol => SERVER_PROTOCOL }
```
tomhughes: I think this would be better in `config/initializers/action_mailer.rb` if possible. I like to restruct changes to `application.rb` to what can't go elsewhere as it makes updates easier.

gravitystorm: Done. There's an order-of-loading issue with using `Rails.application.config` to achieve this in the initializer, so I just went with overriding ActionMailer::Base.default_url_options instead. That seems reasonable and matches the smtp_settings approach too.

Issue #1289, opened 2016-09-15
Best method for renaming translation keys?


gravitystorm:
I'm doing some work on refactoring controller methods, to bring them into line with standard rails naming (i.e. new/create/edit/update/show/index instead of view, list etc).

Part of this involves renaming translation keys, e.g. `diary_entry.view.xxx` becomes `diary_entry.show.xxx`. This means that all the translations are lost, and translators have to re-translation strings that haven't actually changed.

What is the best approach for dealing with this? Can we get translatewiki to handle this situation? Should we add the new translation keys first, to allow translators to copy+paste themselves? Do we just push on and let the translators deal with it? Or something else?

I'd definitely prefer the first - we should be able to fix the translation keys without losing any translations or asking the translators to do make-work. But I don't know what's practical.

tomhughes:
@Nikerabbit How big an issue is if it we rename keys in the non-english translations? Can you easily resync to TranslateWiki when you next push changes?

Nikerabbit:
I would prefer to get list of key renames, so I can execute them in translatewiki.net preserving history.

PR #1827, opened 2018-04-18
Use lazy translation lookups for the user_blocks partials

gravitystorm:
Lazy translations are easier to read, to write, and easier to deal with when refactoring things like partial names, controller methods and controller names. 

This is just a first gentle step in a wider project.

@Nikerabbit as per https://github.com/openstreetmap/openstreetmap-website/issues/1289 this PR will involve changing translation keys on Translatewiki, if it is merged.

### DIFF
--- a/app/views/user_blocks/_block.html.erb
+++ b/app/views/user_blocks/_block.html.erb
@@ -11,14 +11,14 @@
   <td class="<%= c1 %>"><%=h block_status(block) %></td>
   <td class="<%= c1 %>">
     <% if block.revoker_id.nil? %>
-      <%= t('user_block.partial.not_revoked') %>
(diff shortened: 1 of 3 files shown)


Nikerabbit:
Thanks for the heads up.

tomhughes:
As you've only changed the english messages I don't think this will affect @Nikerabbit as such, but it does mean those messages will have to be translated again.

Nikerabbit:
After merged, I will apply the renames to also translations so they don't actually need to be translated again.

tomhughes:
OK. I've merged it now if you want to do your magic.

gravitystorm:
It looks like everything was renamed on translatewiki - thanks @Nikerabbit !

PR #1830, opened 2018-04-19
Mention error message in OAuth setup for iD

zerebubuth:
My usual workflow for trying to fix errors that I don't understand is to paste the error message into a search engine, and I think many other people do the same. Putting the error message here would help me, and other people, find the step missed during setup.

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -86,6 +86,8 @@ potlatch2_key: "8lFmZPsagHV4l3rkAHq0hWY5vV3Ctl3oEFY1aXth"
 
 Follow the same process for registering and configuring iD (`id_key`) and the website/Notes (`oauth_key`), or to save time, simply reuse the same consumer key for each.
 
+**NOTE:** If you forget to set up OAuth, then you will get an error message similar to `uninitialized constant ActionView::CompiledTemplates::ID_KEY`.
+
 ## Troubleshooting
 
 Rails has its own log.  To inspect the log, do this:


HolgerJeromin:
Perhaps this situation can be detected on runtime and handled with a clear error info.

gravitystorm:
> Perhaps this situation can be detected on runtime and handled with a clear error info.

Sounds good - @HolgerJeromin could you try making a pr for that?

HolgerJeromin:
I have no idea how integration of the website code and iD works. Analyzing the correct place for this warning (and learning ruby along the way) is not possible for me right now 8-/

mmd-osm:
Mapping a non existing ID_KEY to a blank value would show the following error popup when starting iD:

grafik


```ruby
diff --git a/app/views/site/_id.html.erb b/app/views/site/_id.html.erb
index 867b148..57a8967 100644
--- a/app/views/site/_id.html.erb
+++ b/app/views/site/_id.html.erb
@@ -1,7 +1,7 @@
 <%= javascript_include_tag "edit/id" %>
 
 <div id="map">
-  <% data = { :key => ID_KEY } -%>
+  <% data = { :key => defined?(ID_KEY) ? ID_KEY : "" } -%>
   <% data[:lat] = @lat if @lat -%>
   <% data[:lon] = @lon if @lon -%>
   <% data[:gpx] = trace_data_url(params[:gpx], :format => :xml) if params[:gpx] -%>


```

PR #1836, opened 2018-04-25
Use lazy translation lookups for redactions

gravitystorm:
cc @Nikerabbit 

The following translations are renamed in this PR

* `redation.*` to `redactions.*`


Nikerabbit:
Thanks for the heads up. I'll rename translations once merged.

PR #1842, opened 2018-04-26
Friendly error msg for missing iD key in application.yml

mmd-osm:
Fall back to an empty iD key in case it hasn't been maintained in application.yml yet.

This way, a user will see an _"iD has not been configured"_ popup when starting iD, instead of a stack trace with "uninitialized constant ActionView::CompiledTemplates::ID_KEY" on it.

Fixes #1830

### DIFF
--- a/app/views/site/_id.html.erb
+++ b/app/views/site/_id.html.erb
@@ -1,7 +1,7 @@
 <%= javascript_include_tag "edit/id" %>
 
 <div id="map">
-  <% data = { :key => ID_KEY } -%>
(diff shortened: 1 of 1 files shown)


mmd-osm:
I would say this obsoletes https://github.com/openstreetmap/openstreetmap-website/commit/b73825efe78dd1f77db8ad36709b833f592543e5, which could probably be reverted, then?

mmd-osm:
Obsoleted by #2177

PR #1851, opened 2018-05-09
More lazy translation lookups for traces

gravitystorm:
A few more lazy translations, to make future refactoring easier.

@Nikerabbit  This PR contains translation key renaming as follows:

* `trace.trace_form.*` is combined into the existing `trace.create.*`
* `trace.trace_header.traces_waiting.*` becomes `trace.create.traces_waiting.*`
* Other `trace.trace_header.*` is combined into the existing `trace.list.*`


Nikerabbit:
I've renamed these on wiki. Thanks again for providing the rename patterns.

Issue #1847, opened 2018-05-03
Add "My Places" to allow more than just one "Go to Home Location"


jidanni:
```Go to Home Location``` is great, but assumes the user only has one Home, and no Office, and no Favorite Bowling Alley etc.

Therefore please add a My Places, where one can store more than just one location.

tomhughes:
This sounds like something is primarily an end user feature, which would be out of scope, but I guess it is potentially useful to mappers as well.

jidanni:
Also the user would no longer need to rely on browser bookmarks for each of his favorite places.
Thus no matter what device he logged in from, there his favorite places would still be.

No need to add another level: Want to Go, Family, Custom____, .... just yet that is. Just give him a way to save more than one (Home location). I mean it is rather a cultural assumption, that for all seasons of the year, all people live in the same place.

mmd-osm:
Similar https://trac.openstreetmap.org/ticket/807 (Allow people to be multi-homed)

PR #1868, opened 2018-05-16
Add "Set home location to here" button on context menu

jguthrie100:
I thought it would be helpful to make setting the user's home location much easier, and save having to use the slightly clunky Settings page interface.

Now the user can just right-click anywhere on the map and select "Set home location", and their home location will be updated (provided they are logged in)

The backend process might need reworking a bit (not sure how happy you'll be using an Ajax call - However creating a new valid OSM.Router.route path seemed like a bit of an overkill!)

### DIFF
--- a/app/assets/javascripts/index/contextmenu.js
+++ b/app/assets/javascripts/index/contextmenu.js
@@ -72,6 +72,21 @@ OSM.initializeContextMenu = function (map) {
     }
   });
 
+  map.contextmenu.addItem({
+    text: I18n.t("javascripts.context.set_home_location"),
+    callback: function setHomeLocation(e) {
+      var precision = OSM.zoomPrecision(map.getZoom()),
+          latlng = e.latlng.wrap(),
+          lat = latlng.lat.toFixed(precision),
+          lng = latlng.lng.toFixed(precision);
+console.log(lng);
+$.ajax({url: "/api/0.6/set_home_loc?lat=" + lat + "&lon=" + lng, success: function(result){
+        console.log("TREST");
+    }});
+      //OSM.router.route("/set_home_loc?lat=" + lat + "&lon=" + lng);
+    }
+  });
+
   map.on("mousedown", function (e) {
     if (e.originalEvent.shiftKey) map.contextmenu.disable();
     else map.contextmenu.enable();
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/index/contextmenu.js
```diff
@@ -72,6 +72,22 @@ OSM.initializeContextMenu = function (map) {
...
 
+  map.contextmenu.addItem({
+    text: I18n.t("javascripts.context.set_home_location"),
+    callback: function setHomeLocation(e) {
+      var precision = OSM.zoomPrecision(map.getZoom()),
+          latlng = e.latlng.wrap(),
+          lat = latlng.lat.toFixed(precision),
+          lng = latlng.lng.toFixed(precision);
+      
+      $.ajax({url: "/set_home_loc?lat=" + lat + "&lon=" + lng, success: function(){
```
tomhughes: I wonder if this should be a `POST` request? It is technically idempotent I guess, but having `GET` requests modify state doesn't really seem right.

config/routes.rb
```diff
@@ -256,6 +256,9 @@
   # directions
   get "/directions" => "directions#search"
 
+  # set home location
+  match "/set_home_loc" => "user#set_home_location", :via => [:get, :post]
```
tomhughes: Whatever the decision is about the request type to use, this should only match what is needed and not the other one.

mmd-osm:
> and their home location will be updated (provided they are logged in)

What happens if I'm not logged in? Does it make sense to include this item in the context menu in this case?

tomhughes:
No, it doesn't - you can tell client side by checking `OSM.user` which will be the user ID if they are logged in.

jguthrie100:
> How does this approach scale to multiple locations, as per #1847? I guess we'd still only update that one home location, right?

I guess it would still work normally, unless the "Home location" is saved as "`location[0]`" or something along those lines - even then, it would be an easy fix!

jguthrie100:
The tests are passing, its just failing on some unrelated, badly translated locales?

tomhughes:
A nasty side effect of the way you are adding the marker is that it causes the map to pan and zoom which is probably not sensible.

Fixing that isn't going to be trivial though, as it require stashing the marker somewhere that both scripts can find it, or using a custom event or something, rather than faking a click, which was always a pretty nasty thing to to.

jguthrie100:
I've created a common method, although it could maybe be placed in a more appropriate file (is there a global helper functions library anywhere, or is osm.js ok?)

PR #1871, opened 2018-05-17
Fix a browse icon typo

hikemaniac:
This PR fixes a typo in the `browse.scss` file, where one file had the invalid extension ".pn" instead of ".png".


gravitystorm:
Thanks!

PR #1872, opened 2018-05-17
Avoid using jsonp for GraphHopper

karussell:
`jsonp` should be avoided as error reporting is not always possible. The necessary magic should be done from jquery automatically. (is there a possibility to test this on an openstreetmap.org domain before going production?)

### DIFF
--- a/app/assets/javascripts/index/directions/graphhopper.js
+++ b/app/assets/javascripts/index/directions/graphhopper.js
@@ -33,7 +33,7 @@ function GraphHopperEngine(id, vehicleType) {
           point: points.map(function (p) { return p.lat + "," + p.lng; })
         },
         traditional: true,
-        dataType: "jsonp",
+        dataType: "json",
         success: function (data) {
           if (!data.paths || data.paths.length === 0)
             return callback(true);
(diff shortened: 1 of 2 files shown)


tomhughes:
Well as far as I know we were only using jsonp because XHR wasn't possible - we certainly prefer not to use it. Is there a suitable cross domain policy in place that will allow access?

Also you will need to update https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/application_controller.rb#L411 to move graphopper from `script_src` to `connect_src`.

karussell:
> Is there a suitable cross domain policy in place that will allow access?

Yes, it was reported it works for others :)

For https://graphhopper.com/api/1/info?key=_KEY_ you can see:

```
access-control-allow-origin=*
access-control-expose-headers=X-RateLimit-Limit,X-RateLimit-Remaining,X-RateLimit-Reset,X-RateLimit-Credits
```

And 
`
curl -v -XOPTIONS 'https://graphhopper.com/api/1/info?key=_KEY_'
`
reports some more Access-Control headers
> to move graphopper from script_src to connect_src

done

karussell:
Thanks!

karussell:
When will it be deployed so that I can test if it still works?

tomhughes:
It's been tested, and has been deployed - it should be live by now.

karussell:
Wow, that was fast. Thanks a bunch - also for removing the other jsonp parameter! Works :+1:

PR #1880, opened 2018-05-24
Wrap the permission label with a <label> tag

EdwardBetts:
Gives the user a larger target to toggle a permission.

### DIFF
--- a/app/views/oauth/authorize.html.erb
+++ b/app/views/oauth/authorize.html.erb
@@ -12,7 +12,7 @@
   <p><%= t '.allow_to' %></p>
   <ul>
   <% @token.client_application.permissions.each do |perm| %>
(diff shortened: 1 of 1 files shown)


EdwardBetts:
Currently the user has to click the check box to toggle the option, with this change they can also click the label. The HTML changes to look like this:

`<li><input type="checkbox" name="allow_read_prefs" id="allow_read_prefs" value="yes" checked="checked" /><label for="allow_read_prefs">read your user preferences.</label></li>`

gravitystorm:
Thanks @EdwardBetts ! This fixes a pet peeve of mine. There's lots more situations like this throughout the codebase, a simple search of `app/views` shows lots of label tags that aren't associated with any form fields. Even the ones that have "for" attributes are often wrong - e.g. app/views/user/new.html.erb where none of the label "for" attributes match the id of the form field. If you get the chance, fixing some of these, and moving them to rails tag helpers, would be appreciated too!

PR #1883, opened 2018-05-27
Change language immediately after updating settings

jguthrie100:
Makes sure that the correct locale is used immediately after updating settings.

At the moment, changing the user's preferred language on the settings page doesn't take effect immediately, and the old locale language is used on the "Your settings have been successfully updated" page.

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -712,7 +712,7 @@ class UserController < ApplicationController
     end
 
     if user.save
-      set_locale
+      set_locale(true)
 
       if user.new_email.blank? || user.new_email == user.email
         flash.now[:notice] = t "user.account.flash update success"
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -118,7 +118,6 @@ class UserController < ApplicationController
   end
 
   def account
-    @title = t "user.account.title"
     @tokens = current_user.oauth_tokens.authorized
 
     if params[:user] && params[:user][:display_name] && params[:user][:description]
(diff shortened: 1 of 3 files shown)


app/controllers/user_controller.rb
```diff
@@ -126,6 +126,7 @@ def account
          (params[:user][:auth_provider] == current_user.auth_provider &&
           params[:user][:auth_uid] == current_user.auth_uid)
         update_user(current_user, params)
+        @title = t "user.account.title"
```
tomhughes: Would it not be more sensible just to move setting of the title to later in the method rather than duplicating it here?

app/controllers/user_controller.rb
```diff
@@ -712,6 +713,7 @@ def update_user(user, params)
     end
 
     if user.save
+      @preferred_languages = current_user.preferred_languages
```
tomhughes: What is this doing? Does anything actually use that member of the controller?

Was it an early attempt to reset the cache on the model that has since been replaced with the change in the model?

app/models/user.rb
```diff
@@ -189,7 +189,7 @@ def preferred_language
   end
 
   def preferred_languages
-    @preferred_languages ||= Locale.list(languages)
+    Locale.list(languages)
```
tomhughes: Rather than just getting rid of the memoization maybe we should use an `after_save` filter to set it to `nil` so that it will be recomputed on next read?

app/controllers/user_controller.rb
jguthrie100: It makes much more sense - I didn't realise that the `redirect_to` method didn't immediately exit the method

app/controllers/user_controller.rb
jguthrie100: It was a hack, that in combination with the removed memoization in `user.rb` resulted in the preferred_languages being recomputed

PR #1897, opened 2018-06-09
Usermess

jimnanney:
Make user#view into more resourceful user#show


tomhughes:
Could you rework this so that it's limited to just the user changes - if you  want to propose other changes then fine but please do it in separate PRs rather than mashing a whole load of things together.

In fact it doesn't look like those other changes are even from you, so you have maybe based this on somebody else's tree? If you have then please rebase it on our master branch.

Issue #1314, opened 2016-10-06
Move the non-db quad_tile functions to an extension?


gravitystorm:
Documenting the existence of the non-db quad_tile functions is useful (#1313) but it makes the instructions longer than they already are.

It would be good to find an alternative, so that the compiled versions of the functions are available automatically, without developers needing to care about them. Of course, the ruby fall-back means that they already don't need to care, but only until they waste a lot of time running the test suite!

As far as I'm aware, the easiest way to make the compilation happen automatically is through bundler when installing gems. The code only relies on `<math.h>` so there's no fancy dependencies. Would it make sense to move the quad_tile code to a gem? Or is there a way to put the extension compilation instructions into the openstreetmap-website/Gemfile instead?

http://guides.rubygems.org/gems-with-extensions/

Issue #1852, opened 2018-05-09
Integrate the high-performance GPX importer


gravitystorm:
For a long time the OSMF has used a [high performance GPX importer](https://github.com/openstreetmap/gpx-import) to process uploaded traces. This runs as a separate daemon on the servers, unrelated to this code.

This has a number of drawbacks:

* There is a lot of ruby code in this repo for importing traces that is [unused and bitrotting].(https://github.com/openstreetmap/openstreetmap-website/issues/281)
* The emails sent by the GPX importer [are not translated](https://github.com/openstreetmap/gpx-import/tree/master/templates).
* There is no queue management for pending traces, so only one import process (on one server) can run at any time.
* The daemon needs its own installation and configuration, its own connections to the database, and its own ability to send mail.

Having a daemon running background tasks is fairly standard in the modern rails world. So I'd like to propose some changes:

* We use the [ActiveJob](http://guides.rubyonrails.org/active_job_basics.html) framework for managing the queue of traces that need importing.
* We rework the C code as a ruby gem, so that we can declare a dependency in the Gemfile and it gets installed and compiled for 'free' with bundler.
* We refactor so that only the parsing of the trace files is handled by the C code, and the email notifications and any other scaffolding comes from this rails app. This will allow us to support translations.
* We ensure that as many worker processes as are necessary can be run in parallel, to improve import speeds.

I'm not familiar at all with the C code, nor the operations of it in production, so any advice on this proposal is welcome!

mmd-osm:
Out of curiosity, what was the main bottleneck motivating the switch to a C implementation?

Inserting gps_points is notoriously slow to the use of single insert, but would probably benefit quite a bit from a bulk insert operation via `activerecord-import `(-> affects https://github.com/openstreetmap/openstreetmap-website/blob/master/app/models/trace.rb#L292-L307). So this part could be covered by some Rails code with decent performance.

I'm not sure if parsing GPX files as such is really an issue. Once I disable `tp.save!` in trace.rb, processing traces via gpx_import.rb is really fast.

The calculation of the `tile` value is already exposed via another Gem (native implementation), and is only duplicated in the GPX importer.

Then there's some image generation which I don't know much about. Could this be the only thing left worthwhile keeping in a native gem?

gravitystorm:
I had a conversation with @zerebubuth two weeks ago, and he was also suggesting that it might not be worth keeping the C implementation, and instead just refactoring the ruby version to be performant.

[...]

PR #1899, opened 2018-06-13
Remove quad_tile library and extract to gem

benreyn:
Quad tile functions are now installed via a gem with native extension
automatically. This improves the run time of the test suite by ~30%

Ref openstreetmap/openstreetmap-website#1314


benreyn:
Hi @gravitystorm 👋 This was a nice self contained PR that we can submit from ruby-for-good 2018. Theres lots of conversation to have around how we want to get this in, mainly who should be the owner of the `quad-tile` gem and how it should be released to ruby-gems. This PR seems as good a place as any to get that conversation started.

gravitystorm:
@benreyn Great! I think the gem should probably be under the [openstreetmap github repo](https://github.com/openstreetmap/), which we can arrange if you're happy with that. I'm also happy for you to give us any guidance on how to release it to ruby-gems. I have a couple of gems there but not a lot of experience with the process. Should we set up an organisation or similar on rubygems.org?

As a side note, the .travis.yml needs a slight tweak to remove the quad_tile build step, as part of this PR.

benreyn:
@gravitystorm,

Re: transfer to the osm org on github, yeah that’s exactly what I had in mind. Ill take a look at the Travis build steps tomorrow, and, I think your right about he rubygems setup. Given this and few other potential c extension gems that might get built in The future, (re: #1852) an osm account on rubygems.org would probably be good. I’ll do some digging tomorrow and find out what the recommended approach is.

tomhughes:
We can't use git gems, so the gem definitely needs to be published and the Gemfile changed to reflect that, and we need control of both the gem source and the publication of the gems.

benreyn:
It looks like github won't let me transfer ownership unless I have permissions to create repositories on the OSM organization.

tomhughes:
Ah yes I forgot about that annoying "feature" of github...

benreyn:
Shouldn't there be a way for one of either two parties to "request" a transfer that requires the other to "approve" it?

tomhughes:
It works like that if you're transferring to another person, but not if you're transferring to an organisation :-(

benreyn:
Its a bit of a roundabout, but I could transfer to you and then you could transfer to the OSM org. (thanks github)

tomhughes:
As you like. I did send you a member invite for the openstreetmap org as well.

benreyn:
👍 That would be MUCH less brain damage.

benreyn:
It looks there are still specific permission I need _within_ the organization to make the transfer. Specifically the ability to create new repositories.

tomhughes:
Yes you either need to be an owner or I need to enable repo creation for ordinary members, which is what I have just done.

benreyn:
Wonderful! Transfered. Ill let you handle publishing to ruby gems, and then I can clean up this PR to point to that source, and fix the TravisCI failures.

tomhughes:
I've activated travis for it, but does the travis config actually do anything? It doesn't look like it runs any tests? Not that I have managed to make them run locally either...

benreyn:
`bundle exec rake test` should do the trick. What happening for you locally?

tomhughes:
Oh I wasn't going through bundle, and I have rake 12 not rake 10 so it was erroring out requiring `rake/extensiontask`. Is there a reason we're asking for an old version of rake?

Hmm and now apparently I broken it by doing what `gem build` told me to do and changing the `minitest` dependency!

tomhughes:
Ah rake 12 does work if you have rake-compiler.

benreyn:
Yeah, the steps should be a `bundle install` to get the development dependencies, followed by a `rake build` and `rake test` (apologies, should have probably been more explicit, in the README)

tomhughes:
I think a bigger problem is that you've changed the ruby code so that it unconditionally replaces the compiled versions of the methods with the ruby versions which will kind of defeat the whole point ;-)

The previous structure was designed so that it would still work even if somebody hadn't built the extension and would just be slower - if we're moving to a gem we can probably assume it will be compiled and delete the ruby versions.

benreyn:
Am I reading the diff wrong? Im fairly certain that the ruby versions of those methods are deleted [here](https://github.com/openstreetmap/openstreetmap-website/pull/1899/files#diff-601230e3ecf5411db5c29ccd4b55f0cb)

tomhughes:
Right, but they still exist in the gem in `lib/quad_tile.rb` after the load of the C versions and if you stick a `raise` in one you will find the tests hit it because the ruby versions have replaced the C versions.

benreyn:
Oh! yes your right! I think those accidentally got copied over when I brought over the  sql methods.

benreyn:
Thanks for taking care of removing those @tomhughes. Ill clean up the travis config and gem source on this branch this afternoon. (still morning here and I need to get to work) Ill ping you here when its done.

db/functions/Makefile
```diff
@@ -1,7 +1,7 @@
 PG_CONFIG ?= pg_config
 DESTDIR ?= .
 
-QTDIR=../../lib/quad_tile
+QTDIR=./
```
tomhughes: I was already trying to figure how best to make this find the header in the quad_tile gem ;-)

I mean duplicating it like this is an option but I wonder if it's better to avoid duplication by having the gem export the path to the header?

Gemfile
```diff
@@ -55,6 +55,9 @@ gem "record_tag_helper"
 gem "rinku", ">= 1.2.2", :require => "rails_rinku"
 gem "validates_email_format_of", ">= 1.5.1"
 
+#
+gem "quad_tile"
```
tomhughes: This needs to be `~> 1.0.1` as the first version doesn't actually build.

db/functions/Makefile
benreyn: I was wondering about this myself. I can take a spin at how that might look. Ive always been curious what the best way to handle shared dependencies within C code is.

benreyn: Ahh I think I see what your saying now, the Gem could expose this in some way. Ill add that over in the quad_tile repo

benreyn: Turns out bundler has a hand way to do this with `bundle show`. Since we `bundle install` before this step we can assume its installed at this step. 👍

Gemfile
benreyn: Addressed in 0a08659

db/functions/Makefile
benreyn: Would be nice however if we could figure out a way to extract this to the gem as well (possibly renaming to something like `osm_extensions`) and avoid any manual compiling step in the install. Might be a clever way to do this with the pg gem. Ill look into it.

PR #1904, opened 2018-06-17
Initial cut at authorization patterns

cflipse:
This is an initial run at adding a centralized authentication mechanism by the team from RubyForGood

During the pre-RubyForGood conversations, we converged on `CanCanCan`  because of it's centralized authorization configuration, the ability to ensure authorization happens with a single controller filter, and it's default-deny policy.

This PR represents a few initial passes at making this work within the existing system.  The basics of the authorization controls have been added to the application controller, and a small number of controllers have been shifted to the new system, to demonstrate a couple of different authorization scenarios:

For a controller that requires admin privledges, forcing a login for an unauthenticated user, while raising an error for an authenticated user who happens to not be an administrator.  The `Diary Entry` controller's `deny_access` shows how we would be able to override failure behavior when needed.

CanCanCan also provides view helpers that can check permissions, and so avoid rendering links to unauthorized resources, but we have not added any of those to the view templates.

There is something very important to note here:  The existing code leans towards the friendly, describing to users why access has been denied to a particular resource.  A default-deny system like `cancancan` _cannot_ provide this information on it's own. As a result, either some of the error messages will have to become less informative, or it will become necessary to duplicate some subset of the access logic in the controllers in order to provide more informative failure messages.

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -103,7 +103,9 @@ class SiteController < ApplicationController
     @locale = params[:copyright_locale] || I18n.locale
   end
 
-  def welcome; end
+  def welcome
+    require_user
+  end
 
   def help; end
 
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -6,9 +6,10 @@ class SiteController < ApplicationController
   before_action :set_locale
   before_action :redirect_browse_params, :only => :index
   before_action :redirect_map_params, :only => [:index, :edit, :export]
-  before_action :require_user, :only => [:welcome]
   before_action :require_oauth, :only => [:index]
   before_action :update_totp, :only => [:index]
+
+  authorize_resource :class => false
 
   def index
     session[:location] ||= OSM.ip_location(request.env["REMOTE_ADDR"]) unless STATUS == :database_readonly || STATUS == :database_offline
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -17,9 +17,7 @@ class Ability
 
       can [:create, :edit, :comment, :subscribe, :unsubscribe], DiaryEntry
 
-      if user.administrator?
-        can [:hide, :hidecomment], [DiaryEntry, DiaryComment]
-      end
+      can [:hide, :hidecomment], [DiaryEntry, DiaryComment] if user.administrator?
     end
     # Define abilities for the passed in user here. For example:
     #
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -3,9 +3,10 @@ class Ability
 
   def initialize(user)
     can :index, :site
+    can [:permalink, :edit, :help, :fixthemap, :offline, :export, :about, :preview, :copyright, :key, :id, :welcome], :site
 
     if user
-      can :welcome, :site
+      can :weclome, :site
     end
     # Define abilities for the passed in user here. For example:
     #
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -1,0 +1,37 @@
+class Ability
+  include CanCan::Ability
+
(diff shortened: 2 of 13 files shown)


tomhughes:
When you say that this represents an "initial pass" does that mean it is (in principle) mergable as is just that there will then be more controllers to convert, or will there be more work to do to convert other controllers before it is fully usable?

In any case I've asked @gravitystorm to review as I think he has more experience in this area and has certainly already been considering the options.

cflipse:
In theory, this is a fair exemplar of what the controllers will change to, and subsequent controllers can be adapted following this pattern.

For some of the controllers, there will be more work to do, _depending_ on how you want to adjust to this reality:  CanCanCan will not tell you _why_ you have failed authorization, only that you have.  It's inherent in the default-deny style.  Currently before filters are taking away access to something, and they can report _why_ they are taking away the access.  Default-deny can't do that, because instead of taking away access for a reason, you never had it in the first place.

`DiaryEntriesController` addresses this by overriding the `deny_access` handler and essentially replicating the knowledge that `hide` and `hidecomment` _require_ an adminstrator role.  Something simliar would have to be done for missing capabilies, checking against the `granted_capabilities` helper to attempt to produce a nicer denial message.  Note that this is only necessary if you want to provide _specific reasons_ for access refusal.

gravitystorm:
Hi Chris - thanks for the work and sorry it's taken me a while to review. In general it looks good to me, and I hadn't considered how the tokens would work so that's very helpful.

I have two questions at present, which I'll mention as code comments. Otherwise I plan to take this branch and build on top of it further.

app/controllers/site_controller.rb
```diff
@@ -102,7 +103,9 @@ def copyright
     @locale = params[:copyright_locale] || I18n.locale
   end
 
-  def welcome; end
+  def welcome
+    require_user
```
gravitystorm: Could you explain this bit for me? My thoughts were that using CanCanCan would be to remove uses of our home-grown require_user/admin/moderator from the code. My approach would be to change the authorisation rules to have `can [:welcome], :site` in the "if user" section, rather than putting it in the 'all users' section and then adding a restriction here. Have I overlooked something?

app/models/capability.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    end
+  end
+
+  private
+
+  # If a user provides no tokens, they've authenticated via a non-oauth method
+  # and permission to access to all capabilities is assumed.
+  def capability?(token, cap)
+    token.nil? || token.read_attribute(cap)
```
gravitystorm: The original `token.nil?` check was here when you were combining both the Ability and Capability stuff, so for normal web stuff it was ok if the token was nil. But it's been refactored since, so that it's only used for token-based capability checks. I worry that this opens up a foot-gun situation, since as far as I can tell there's no reason for calling `capability?` without a token -  what do you think?

gravitystorm:
I should add that if the answer to my questions is "I can't remember, that's too long ago" then that's fine too :-)

app/models/ability.rb
```diff
@@ -0,0 +1,49 @@
...
+    can :index, :site
+    can [:permalink, :edit, :help, :fixthemap, :offline, :export, :about, :preview, :copyright, :key, :id, :welcome], :site
+
+    can [:list, :rss, :view, :comments], DiaryEntry
+
+    can [:search, :search_latlon, :search_ca_postcode, :search_osm_nominatim,
+         :search_geonames, :search_osm_nominatim_reverse, :search_geonames_reverse], :geocoder
+
+    if user
+      can :weclome, :site
```
mmd-osm: Typo?

gravitystorm:
Thanks for your work on this @cflipse it's been really useful! I've built on top of this further, and opened an updated PR at #2023

PR #1905, opened 2018-06-19
SotM 2018 banner

frafra:
This PR adds the banner for the 2018 State of the Map.


tomhughes:
That image is nearly 1500 pixels wide - banner images need to be 350 pixels wide.

The browser will resize it, but it would be better not to send one that was three or four times bigger than necessary.

frafra:
Sorry; I will talk with the designer (I just did the pull request for her). I will fix as soon as possible.

frafra:
Any news? Is it ok or should we change anything? Thank you :)

tomhughes:
Well I didn't know you had pushed a new version because you didn't say anything ;-)

frafra:
Sorry @tomhughes, I didn't knew rebasing/amending would not send notifications :D

tomhughes:
I thought it did now but I didn't notice it - they're quite subtle I think.

I'll be pushing it as 45a0666523595f930187bdc26aef327e728abd82 shortly - that's just your commit with the GIF changed to PNG because it's the 21st century ;-)

tomhughes:
It's live, so it should be there. Note that if you close it that will stop it appearing again, so check your cookies.

frafra:
Ok, I fount the issue. It looks like ad blockers detect the world "banner" in the URI, so they do not show it.
I could whitelist openstreetmap.org, but I think that many users could have the same issue. Should we rename "banners" to "events"?

tomhughes:
We've been around this - if we do that the adblockers will just match that instead.

We tried to convince them this wasn't adverts but they refused to listen.

bhousel:
Ah yeah, I can confirm that AdBlock Plus is blocking the banners. 😢

frafra:
EasyList is blocking openstreetmap.org because of default regex/keyword, they are not targeting openstreetmap.org specifically (openstreetmap.org is on their whitelist for social buttons).

From their policy:

> Self-promotion should not be specifically removed by EasyList, although equally should not be whitelisted if it is blocked.

I think we should just change the path so we will not be blocked. They will not introduce a specific rule for us for this specific case, no whitelist, no blacklist.

bhousel:
looking at easylist now.. it looks like our `#banner` is explicitly blocked:

screenshot 2018-06-25 10 12 07

tomhughes:
We have literally discussed this with them and they were adamant that our banners didn't fit as self promotion.

tomhughes:
See #1743 and https://github.com/easylist/easylist/pull/906.

bhousel:
Thanks for context @tomhughes  - I'll look into a workaround later today.

frafra:
Thank @bhousel, this is not shown in their main repository.
If the issue is that the link points to an external website, couldn't we just refer to our wiki and ask them to unblock us?

Like pointing to https://wiki.openstreetmap.org/wiki/State_of_the_Map_2018

tomhughes:
Well the actual SOTM one uses an openstreetmap.org domain anyway, but banners for local chapter conferences often don't.

frafra:
So we are not violating their policy at the moment, we could ask to be removed from the blacklist and refer to the wiki page for future events (if an official/internal mini-site is missing).

simonpoole:
I have to say that when I was following the original discussion with the maintainers of the list the clear impression that it really didn't matter what we would do, the banners would get blocked in any case (if necessary by making up a new rule) and I would recommend against expending any effort on a wild goose chase to try and work around it.

frafra:
As far as I understand, the explicitly said that auto-advertising is fine if done on the same domain and if it not intrusive. https://github.com/easylist/easylist/pull/906#issuecomment-366183612

We are following this rule at the moment I suppose. We could just add that in the future we are going to use the same domain. I would be happy to spend half an hour with them even if there could be a low chance of success, because we could not reach almost a third of our users.

Firefishy:
@frafra Go for it. It is worth a try. I'm burnt on the issue.

bhousel:
@Firefishy I'm going to just insert the banner with javascript, so it can't get blocked.

Firefishy:
@bhousel I am unsure if the javascript route is best, it could become an arms war.

My personal idea would be something like div `#osmorg-banner` vs `#external-banner` or similar.

Zverik:
There was also this rejected pull request: #1743.

bhousel:
> @bhousel I am unsure if the javascript route is best, it could become an arms war.
> My personal idea would be something like div #osmorg-banner vs #external-banner or similar.

Yeah I did some digging today into how ABP works, and I don't think I can circumvent it after all.  The rules are pretty sophisticated so whatever we pick, I can imagine a regex that would match it (even for dynamically generated content).  I'll think about it some more.

bhousel:
> Changing the ID of the announcement area is something we should consider - it's not really something I've ever thought about as a banner, so #banner stuff catches it unintentionally. Aren't we using the same div for downtime announcements?

The `#banner` id is not triggering unintentionally.. openstreetmap.org's `#banner` is explicitly blocked in easylist.

If we change the div id to something else, easylist will change their block to match whatever we change the div id to.

frafra:
I am not sure they will block it. We are not violating any rule at the moment. I think we should just ask them to re-evaluate our block.

PR #1914, opened 2018-06-24
add icon for OSMF membership to user page

datendelphin:
Add an icon for OSMF membership, similar to user roles.
Creates a new table for memberships, and adds an optional setting
on the user settings page to select if the user wants the
membership to be displayed on to user page or not.


tomhughes:
Why create a new table? I thought the plan was to use a role?

Is there a workable plan for keeping this in sync yet?

datendelphin:
The query to extract the osm user names from civicrm looks something like this:

https://join.osmfoundation.org/wp-content/plugins/civicrm/civicrm/extern/rest.php?entity=Membership&action=get&api_key=userkey&key=sitekey&json={"sequential":1,"return":"join_date,contact_id.custom_1","membership_type_id":{"IN":["Normal Member","Associate Member"]},"status_id":{"IN":["Current","Grace"]},"options":{"limit":10000}}

but with the correct keys of course. It outputs json. I can provide a script to update the osm usernames with this query if desired. Updating daily should probably be enough.

datendelphin:
Adding a new table for the membership looked a bit cleaner to me, but if desired I can put it in the existing role table. It does not perfectly match as the role table also has a granter and a date when it was granted.

mmd-osm:
When using the roles table, we might want to adjust the explanation text a bit, as an OSMF membership is not only shown as an icon on the user page, but also exposed as role via the ` /api/0.6/user/#id` endpoint, meaning other applications can also find out about it.

test/integration/user_membership_test.rb
```diff
@@ -0,0 +1,67 @@
+require "test_helper"
```
GaspardPO: Will it be possible to split the method in different tests?

GaspardPO: different functions so if the first assert fails, we still test the others ones.
So we know if a regression affects all features, or just one.

app/helpers/user_membership_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def membership_icon(user, membership)
+    if user.show_membership?(membership)
+      image = "membership/#{membership}"
+      alt = t("user.view.membership.#{membership}")
+      title = t("user.view.membership.#{membership}")
+      id = "badge-membership-#{membership}"
+    end
+
+    if image
```
gravitystorm: This will always evaluate to true, if the `user.show_membership?` evaluated to true. So I think it would be better to combine the two if statements

gravitystorm: Oh, I see now this is done this way to match the user_roles_helper logic, which is quite similar.

gravitystorm:
I was asked to review this today. Apart from my inline comments, my thoughts are:

* I would suggest using lowercase throughout for 'OSMF', such as in enums, translations keys, factory names etc. If we add another membership in the future with more complex capilitization (FoOBaR) then it would be less to remember to just have everything in lowercase.
* I would prefer to see the tests split up as described into smaller blocks that are a bit more clear as to what is being tested each time. But I would accept the PR as-is since it's fairly similar to the other tests we have.
* Some of the functionality here is squeezed into the user account page, so we end up with a not-very-rails-like way of iterating over the memberships to save the 'show' status. I'd prefer to see this refactored, but that can come later, since the whole account page needs refactoring anyway.
* I'm fine with seeing these in a different table, since I can see that they have a different set of attributes from the user_roles table, and are only being used for badges rather than as part of the permissions system.
* I would consider using a mixture of controller and system tests, instead of an integration test, but I would accept this PR as-is.

I think the final call on the new table vs user_roles table should be made by TomH rather than me.

gravitystorm:
> I can provide a script to update the osm usernames with this query if desired. Updating daily should probably be enough.

If this can be provided as a rake task, that would be ideal. It's best to make the code within the rake task as small as possible so that tests can be more easily written. 

For example, in other situations I have written rake tasks that simply enqueue an ActiveJob job, which is what does the actual work to fetch updates, update tables, etc, since the jobs are easier to test and come with built-in retries, error reporting etc.

mmd-osm:
Maybe that’s also a topic for microcosms: your  user profile page could display a list of groups/orgs you belong to, similar to a GitHub profile page. Membership and icon for a OSMF group would then be centrally managed.

Special casing the OSMF membership like implemented in this PR may not be ideal for such a use case.

brazilsingh:
Reading through the open PRs and wanted to add a note on this one. The idea, an OSMF membership badge is a good one, but the PR looks stalled: it's well behind master now, the new-table-vs-user_roles question was never given a final call, and the CiviCRM sync still needs reworking into a rake task + ActiveJob job as @gravitystorm  suggested. @mmd-osm 's later point about a more general groups/orgs-on-profiles approach also seems worth weighing, since that might be a better home for this than special-casing OSMF membership.

Given all that, my suggestion would be to either close it (with the idea revived later as a fresh PR aligned with the broader groups concept) or, if @datendelphin  is still interested, have them rebase and address the open review points. Either way it'd be good to get it off the stale pile. Just one reviewer's view, leaving the call to the maintainers.

Issue #1543, opened 2017-05-16
Forbid anonymous comments for notes


Zverik:
We in Russia have encountered a person that is leaving anonymous one-letter comments to all of the notes. That got me thinking, why do we do that, allowing anonymous comments? I understand anonymous reports: people can help improving the map without registering. But comments cannot add anything helpful: reporters who commit, register.

HolgerJeromin:
I had seen in the past anonymous comments like: "yes, shop xy is really closed". 
This was useful data. 
But on the other side there are many comments "oh, in the last comment i forgot got login. It was me."

VladimirSlavik:
What happens if the anonymous reporter comes back to answer a question?

TakutoRU:
Of hundreds closed by me notes I found maybe 5 when really anonymous reporter answered question. Others were OSM users who forget to login, or "hit and run" notes, which was never revisited again by author.
Maybe some captcha on comments help with mass bot spamming? False alarms from spam comments to my notes start to be annoyance.

RicoElectrico:
While I've seen these single character comments in Gdańsk, in my experience legitimate anonymous comments are not that rare.

Still, the system is too loose indeed. We could require a captcha for more than a few comments from the same person.

rory:
IME i haven't seen many anonymous comments, it's rare in Ireland. But I have seen cases where there has been follow up feedback provided in an anonymous comment.

gravitystorm:
From the above comments it looks like there are both unwanted anonymous comments along with useful ones. So it's not clear to me that we should block them. Rather than basing it on the experiences of a few people it might be worth examining the wider situation.

We publish all the notes including comments at http://planet.openstreetmap.org/notes/ . Someone might be willing to run some analysis of these, to see how many anonymous comments vs regular comments, whether there's any correlation between types (or lengths) of comments vs how quickly they are resolved, and so on.

Alternatively, we can approach this a different way, and instead make it easier for users to flag unhelpful comments, whether anonymous or not, see e.g. #1576 

I realise the originally reported problem is genuine, I just don't want us to [throw the baby out with the bathwater](https://en.wikipedia.org/wiki/Don%27t_throw_the_baby_out_with_the_bathwater).

Zverik:
[...]

PR #1926, opened 2018-07-17
do not allow anonymous users to comment on notes

woodpeck:
Implements #1543

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -5,7 +5,7 @@ class NotesController < ApplicationController
   before_action :check_api_readable
   before_action :authorize_web, :only => [:mine]
   before_action :setup_user_auth, :only => [:create, :comment]
-  before_action :authorize, :only => [:close, :reopen, :destroy]
+  before_action :authorize, :only => [:close, :reopen, :destroy, :comment]
   before_action :require_moderator, :only => [:destroy]
   before_action :check_api_writable, :only => [:create, :comment, :close, :reopen, :destroy]
   before_action :require_allow_write_notes, :only => [:create, :comment, :close, :reopen, :destroy]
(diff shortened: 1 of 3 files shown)


woodpeck:
Apparently the travis check for this PR failed due to some travis-internal error. Can I re-trigger the check without making an unnecessary dummy commit?

woodpeck:
I noticed that LWG have suggested to make any acess to notes contingent on a login, including creating and reading notes. If implemented, that would make this pull request redundant.

gravitystorm:
> I think the only question that remains is the policy one of whether we want to do this.

I think it should be merged. The evidence presented is convincing to me.

It's also a reversible change - if worst comes to worst and there's a compelling reason later on to allow anonymous comments again, we can easily do that.

simonpoole:
And ..... what about announcing a significant behaviour change of the website?

tomhughes:
It's not live yet.

tomhughes:
That said we will obviously want to put it live soon so if people feel a communication strategy is necessary then now is the time to arrange it.

I'm not sure how many of the people impacted such a strategy might reach though - in reality I doubt many people other than deliberate abusers are going to be impacted at all.

simonpoole:
I was more considering the "normal" OSM contributors as an audience for an announcement @harry-wood

Zverik:
Awesome, thank you Andy and Tom! I will announce the change to the Russian part of the community at least :)

mmd-osm:
Should we add a small one liner explanation that "you need to log on to comment on this note"?

matkoniecz:
With "log on" being a link?

RicoElectrico:
> Should we add a small one liner explanation that "you need to log on to comment on this note"?

As well as "log in to receive notifications about note status" if creating a note.

While we're at it, it'd be nice to link to the https://wiki.openstreetmap.org/wiki/Notes pages in the note creation view as well Something like: "Learn more about Notes".

PR #1942, opened 2018-07-26
Added distance of traces to model and view

jguthrie100:
- Migrated a new column (trace-length)
 - Added 'Distance' row in trace view
 - Calculates distance between each point using haversine formula

### DIFF
--- a/app/models/trace.rb
+++ b/app/models/trace.rb
@@ -2,17 +2,18 @@
 #
 # Table name: gpx_files
 #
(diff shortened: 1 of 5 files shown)


tomhughes:
So firstly, what exactly is the benefit to adding this - how does it help mappers create the map?

More importantly, the code you have edited isn't actually used - the ruby importer is far too slow so we used the C one (https://github.com/openstreetmap/gpx-import) in production and have done for many years.

PR #1953, opened 2018-08-25
Add an API endpoint to get the latest notes

ENT8R:
This PR adds an additional endpoint to the Map Notes API which allows to get the latest notes which are available. It can be accessed at `/api/0.6/notes/latest` and takes the same parameters as e.g. the search endpoint (`limit`, `closed`). The data can be returned as `.xml`, `.json`, `.rss` or `.gpx`.
A possible use case for this endpoint is that applications could request the latest notes and show them on a map/list and users can resolve the notes or ask questions because I think the faster someone comments/resolves a note you created not that long ago the higher is the possibility that the creator of the note wants to answer the question and did not forget what the note was all about.
In the end it might also help to reduce the amount of created notes at all and thus making the map a better map 😄

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -275,6 +275,24 @@ class NotesController < ApplicationController
   end
 
   ##
+  # Return a list of the latest notes
+  def latest
+    # Get any conditions that need to be applied
+    @notes = closed_condition(Note.all)
+
+    # Find the notes we want to return
+    @notes = @notes.order("updated_at DESC").limit(result_limit).preload(:comments)
+
+    # Render the result
+    respond_to do |format|
+      format.rss { render :action => :index }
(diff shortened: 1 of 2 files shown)


tomhughes:
Why not just add a time filter (if there isn't one already) to the existing search endpoint?

ENT8R:
Didn't do this because there is no similar filter for any other endpoint (yet). Maybe this would be better as a separate PR because this would affect all query endpoints... What do you think? Anyway, this is actually a pretty good idea! :+1:

ENT8R:
Oops... Sorry I misunderstood your comment...
I added this as an extra endpoint because I thought it might be more modular this way... But if you think that a time filter would be better, I can instead add this, of course.

ENT8R:
Closing in favor of a new PR...

Issue #832, opened 2014-11-07
Allow sorting and filtering of notes displayed in user profile


rudolfmayer:
Not sure if this was already reported in some way, at least I couldn't find anything in the open issues.

My request is as follows - the user page displaying the notes created, e.g. mine at http://www.openstreetmap.org/user/Rudolf%20Mayer/notes, should allow for
- sorting by creation date, solved date, name, ...
- filtering; here especially I would like to be able to display only the unresolved notes, as these are the ones I'd like to work on improving.

Thanks for your consideration,
Rudolf

mnalis:
I absolutely agree, especially on filtering by unresolved notes. 

I've dozen of "my notes" pages, and if a note ever need work that I cannot do immediately (eg. it need resurveying), it will fall behind first few pages and with current situation I'll likely never see (and hence will not give attention to) that note again...

Or at least, firstly show all unresolved notes, and only after them all resolved notes. That one should literally be single line change in the code (modifying ORDER BY clause) so extremely easy to do, should not stop on anybody toes, and would increase functionality a lot.

mnalis:
since I couldn't wait for this to be implemented, as a temporary workaround I made it myself: 
http://my-notes.osm-hr.org/ 
( or go grab the source yourself at https://github.com/osm-hr/my-osm-notes )

You can specify one (or several) users, and it will show only unresolved notes that they created or participated in (commented, closed, reopened etc). It uses Notes planet dump, so only gets updated once a day. Real solution in openstreetmap website would be much better (and operate in realtime), so hopefully this ticket would get some attention. Until it does, you're welcome to use instance above.

mnalis:
see also https://trac.openstreetmap.org/ticket/5205

letypequividelespoubelles:
any developement coming sson ? It seems it's an expected feature !

gravitystorm:
> any developement coming sson ? It seems it's an expected feature !

Volunteers welcome! Otherwise, it's like many other features and is waiting in the list.

letypequividelespoubelles:
@gravitystorm :) Unfortunetaly I don't have the dev skills :(

mnalis:
@gravitystorm You say volunteers are welcome... Hypothetically speaking, if one were to invest the effort into investigating codebase and writing a pull request to implement sorting by note status / date, what are the chances they'd be getting feedback/hints from OSM people with commit privileges and PR eventually being accepted (if it is written up to the standards, of course)? 

[...]

PR #1954, opened 2018-08-25
Add an API endpoint to get notes of a specific user

### LINKED ISSUE #1194: Add API endpoint for getting notes related to an user as JSON
It would be very useful if all notes related to an user could be accessed programmaticaly. The data sent would be essentialy the same available at `/user/:username/notes` route accessible at profile area, but in JSON format.

ENT8R:
This PR adds an additional endpoint to the Map 🗺 Notes 📝 API which allows to get all notes created by a specific user 👨 (either by the `id` of the user of its `display_name`). It can be accessed at `/api/0.6/notes/user` and takes the same parameters as e.g. the search 🔍 endpoint (`limit`, `closed`). The data can also of course be returned as `.xml`, `.json`, `.rss` or `.gpx`.
This PR would fix #1194 and it is at least a step towards #832...
Originally I planned to add the query parameter to this endpoint too (`?q=*`) but I had some problems ❓ doing this: each time I tested it with a word which was contained in 3 of 5 notes it resulted in 0 results... This is because I commented it out 😞... Maybe somebody else knows why this does not work 😃

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -260,10 +260,43 @@ class NotesController < ApplicationController
 
     # Get any conditions that need to be applied
     @notes = closed_condition(Note.all)
-    @notes = @notes.joins(:comments).where("to_tsvector('english', note_comments.body) @@ plainto_tsquery('english', ?)", params[:q])
+    @notes = filter_notes(@notes, params[:q])
 
     # Find the notes we want to return
     @notes = @notes.order("updated_at DESC").limit(result_limit).preload(:comments)
+
+    # Render the result
+    respond_to do |format|
+      format.rss { render :action => :index }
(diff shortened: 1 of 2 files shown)


tomhughes:
Again why not just add an extra filter to the existing search endpoint?

ENT8R:
Hmm... Because here this would make no sense IMO. Adding a time filter to the endpoint I added now could make sense but by adding only a time filter to the already existing endpoint doesn't solve the problem to get notes created by a specific user.

tomhughes:
Well the point would be to add a user filter as well? I think some other similar endpoints like changesets already have both?

ENT8R:
Ah, so you want everything in the already existing `search` endpoint...
Then it might be better to close both PRs and open a new one once I finished...

PR #1955, opened 2018-08-25
Enhanced the notes search endpoint

### LINKED ISSUE #1194: Add API endpoint for getting notes related to an user as JSON

ENT8R:
Follow-up of #1953 and #1954:
The `search` 🔍 endpoint was enhanced to add the following features 🎉:
- get notes of a specific user 👨 (either by the `display_name` 📛 or the  `id` 🆔)
- get notes of a specific time ⌚️ range (if no `to` (end time) parameter was given, it is automatically set to the current time)
- get the latest notes 📝 if no parameter at all was specified

I still could not figure out why the query filter does not work for the notes of a specific user 🤷‍♂️... (see [the code 💻](https://github.com/ENT8R/openstreetmap-website/blob/notes-search/app/controllers/notes_controller.rb#L275#L281) for more information ℹ️) Maybe someone else knows the issue just by looking at the code 😃...
This PR would fix #1194 and it is at least a step towards #832...


Nakaner:
Did you take privacy in account and limit searching by user ID and display name to authenticated users?

ENT8R:
No, not yet... Should I do this?

tomhughes:
I don't see why - we don't do it for any other APIs currently.

That may change as a result of GDPR but that can be handled as and when we have a final decision on what needs to be done for that.

test/controllers/notes_controller_test.rb
```diff
@@ -876,6 +876,45 @@ def test_search_success
     end
   end
 
+  def test_search_by_user_success
```
GaspardPO: I see the unit test for display name, but is there also a unit test for filter by id?

ENT8R: I've added some more tests with the last commit

ENT8R:
I fixed now all remaining issues, all checks are finishing successful, so is there anything which prevents it from being merged or is everything done?

app/controllers/notes_controller.rb
```diff
@@ -255,15 +255,52 @@ def destroy
...
+    elsif params[:id]
+      @user = User.find_by(:id => params[:id])
+    end
 
-    # Get any conditions that need to be applied
-    @notes = closed_condition(Note.all)
-    @notes = @notes.joins(:comments).where("to_tsvector('english', note_comments.body) @@ plainto_tsquery('english', ?)", params[:q])
+    if @user
+      @notes = @user.notes
+      @notes = closed_condition(@notes)
```
tomhughes: I'd merge these into one line so that it follows the same pattern as the all notes case.

app/controllers/notes_controller.rb
```diff
@@ -255,15 +255,52 @@ def destroy
...
+    elsif params[:display_name] || params[:id]
+      # Return an error message because obviously the user could not be found
+      raise OSM::APIBadUserInput, "The user could not be found"
```
tomhughes: I think we should split this in two and make the text `User XXX not known` where XXX is the id or name we searched for. The comment isn't really necessary as it just states what is obvious from the error.

app/controllers/notes_controller.rb
```diff
@@ -255,15 +255,52 @@ def destroy
...
+    else
+      @notes = closed_condition(Note.all)
+    end
+
+    # Filter by a given string
+    if params[:q]
+      @notes = @notes.joins(:comments)
+      @notes = if @user
+                 @notes.where("to_tsvector('english', comments_notes.body) @@ plainto_tsquery('english', ?)", params[:q])
```
tomhughes: Does the file really change name to `comment_notes` in this case? That is kind of worrying and makes this part of the code very messy... I wonder if we should always start from `Note.all` and then filter by user instead of coming at the notes from the user - that should hopefully avoid having to make this condition extra complicated.

app/controllers/notes_controller.rb
```diff
@@ -255,15 +255,52 @@ def destroy
...
+      begin
+        from = Time.parse(params[:from])
+        to = if params[:to]
+               Time.parse(params[:to])
+             else
+               Time.now
+             end
+      rescue ArgumentError
+        # return a more generic error so that everybody knows what is wrong
+        raise OSM::APIBadUserInput, "The date is in a wrong format"
```
tomhughes: Which date? You've parsed two dates but as there both in the same block you have no idea which one failed... This should probably be split up so the error can be more helpful like `Date XXX not recognised` or something.

app/controllers/notes_controller.rb
```diff
@@ -255,15 +255,52 @@ def destroy
...
+      end
+
+      @notes = @notes.where("(created_at > '#{from}' AND created_at < '#{to}')")
```
tomhughes: There's no need for parentheses here and in any case there's no need to resort to raw SQL as you can do `where(:created_at => from .. to)` instead.

app/controllers/notes_controller.rb
```diff
@@ -255,15 +255,52 @@ def destroy
...
+    end
 
     # Find the notes we want to return
-    @notes = @notes.order("updated_at DESC").limit(result_limit).preload(:comments)
+    @notes = @notes.order("updated_at DESC").distinct.limit(result_limit).preload(:comments)
```
tomhughes: Why was a distinct sort needed here? Under what conditions do we get duplicates where we didn't before?

app/controllers/notes_controller.rb
tomhughes: I definitely think there is a case for starting form `Note.all` and then using doing the user filter as `.joins(:comments).where(:note_comments => { :author => @user })` so that we can do this condition in one consistent way.

ENT8R: > and then using doing the user filter as `.joins(:comments).where(:note_comments => { :author => @user })`

Well, but there is no `author` existing in the table: https://github.com/openstreetmap/openstreetmap-website/blob/master/app/models/note_comment.rb#L9-L10 we can only search by `author_ip` or `author_id`... Is the latter one maybe achievable through `.joins(:comments).where(:note_comments => { :author_id => @user.id })`?

tomhughes: Sorry, that's what I meant. I was playing with with various things and copied the wrong one - you don't need `.id` on the user object though as rails will do that anyway.

mmd-osm:
@ENT8R : please take a moment to update the documentation on the [Wiki page](https://wiki.openstreetmap.org/wiki/API_v0.6#Search_for_notes_on_text_and_comments:_GET_.2Fapi.2F0.6.2Fnotes.2Fsearch) once this gets merged.

ENT8R:
Yepp, that's on my list...

tomhughes:
Merged - note that I renamed the `id` parameter to `user` as it would be very confusing to have `id` mean a user ID in a notes method! It also makes it consistent with changesets. Sorry I didn't notice that before.

ENT8R:
Thank you very much :tada:

ENT8R:
> @ENT8R : please take a moment to update the documentation on the Wiki page once this gets merged.

I updated it now: https://wiki.openstreetmap.org/wiki/API_v0.6#Search_for_notes%3A_GET_.2Fapi.2F0.6.2Fnotes.2Fsearch

ENT8R:
BTW: How long does it take that the changes are live on https://www.openstreetmap.org and not just the testing APIs?

tomhughes:
Half an hour once I deploy it, which I just did.

mmd-osm:
@ENT8R : thanks for updating the Wiki! One thing I'm not entirely sure of it the date format. https://ruby-doc.org/stdlib-2.1.1/libdoc/time/rdoc/Time.html mentions some heuristic to detect date format. However, it would probably be better to recommend one format in the documentation, e.g. iso8601.

PR #1958, opened 2018-08-27
Add parsing coordinates in lon, lat format

Zverik:
The website allows for searching by a coordinate in "lat, lon" format. Quite often the order is opposite, with longitude coming first. GeoJSON, WKT, OpenLayers and many other formats and services use "lon, lat". To search by such coordinate on the OSM website, one needs to manually swap the numbers.

This pull request modifies the `search_latlon` function so that it adds not one, but two lines to its results: both "lat, lon" and "lon, lat" coordinates.

### DIFF
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -39,17 +39,23 @@ class GeocoderController < ApplicationController
   def search_latlon
     lat = params[:lat].to_f
     lon = params[:lon].to_f
-    if lat < -90 || lat > 90
-      @error = "Latitude #{lat} out of range"
-      render :action => "error"
-    elsif lon < -180 || lon > 180
-      @error = "Longitude #{lon} out of range"
+    @results = []
+
+    if lat >= -90 && lat <= 90 && lon >= -180 && lon <= 180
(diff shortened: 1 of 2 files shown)


tomhughes:
I think this is working at the wrong level - it would probably be better done by having the initial search trigger multiple calls to the backend.

The reason I say that is that `normalize_params` parses many things into lat/lon pairs but in most cases the order is clear as there are N/S/E/W markers - only the pure numeric cases can be ambiguous.

As this stands an entirely unambiguous search string will produce two results...

Zverik:
You are absolutely right with NSEW markers — I have added a check for that now. As for the multiple calls, it would complicate the output with multiple result sections showing a coordinate each, so I am trying to do all the work inside a single source.

Zverik:
Awesome, thank you!

mmd-osm:
Interesting side effect: once I click on "Where is this?", two results are presented underneath "Internal", in lat/lon and lon/lat format. Shouldn't the web site know where I am?

grafik

grafik

Zverik:
Thanks @mmd-osm, working on a fix in #1968.

PR #1976, opened 2018-09-06
Fix any_relations always being false

jiaxuyang:
In relation.members, the order of columns is [type, id, role]. And the value of type is among "Node", "Way" and "Relation".


tomhughes:
This should have a test.

Thrimbda:
seems like test is added, any progress on this?

mmd-osm:
Did you encounter any existing changeset in OSM where this is a real issue, i.e. the bounding box was calculated incorrectly?

PR #1979, opened 2018-09-07
Add banner against EU Copyright Directive

Nakaner:
The EU Copyright Directive will harm the project because it requires us to set up automatic upload filters. I hope that I don't have to explain in detail why this is harmful.

OpenStreetMap Germany has gone further. Our tile server tile.openstreetmap.de responds with a black map tile to every tenth tile request.

@woodpeck


gravitystorm:
As @Nakaner points out on his [mailing list post](https://lists.openstreetmap.org/pipermail/talk/2018-September/081390.html), this is a political statement rather than a routine banner. I have my own opinions on whether we should show the banner, but I also believe that it's more appropriate for the OSMF Board to make the decision on this topic.

fredao:
A politicalha statement? It is to protector outros data! Política is madeira in Brussels.

systemed:
> I hope that I don't have to explain in detail why this is harmful

If you don't want to explain it yourself, could you perhaps provide a link to a detailed explanation (ideally with references to the proposed legislation itself) of how this will be harmful to OpenStreetMap in particular? I've seen a lot of commentary about its impact on Wikipedia, sure, but Wikipedia has a very different attitude to copyright to OSM.

Nakaner:
@systemed The English translation of https://www.openstreetmap.de/uf/ is not finished yet (pending review) but you can find the preliminary version at https://michreichert.de/osm/openstreetmap.de/uf/english.html  (source code at https://github.com/fossgis/openstreetmap.de/blob/uploadfilter-english/uf/english.html)

systemed:
@nakaner Thank you.

PR #1983, opened 2018-09-10
Use full url escaping when required in trace tests

gravitystorm:
The same approach is used elsewhere in the test suite.


tomhughes:
I did what I did because none of the recommended  routines worked - originally I used `URI.escape` but rubylint told me that was deprecated and none of the alternatives it suggested actually worked right for what we needed...

PR #1984, opened 2018-09-10
Rename user#list to user#index

gravitystorm:
More resourceful renaming.


tomhughes:
@Nikerabbit this renames `user.list` to `user.index`

PR #1985, opened 2018-09-10
Rename diary_entry#list to #index

gravitystorm:
Even more resourceful route renaming


tomhughes:
@Nikerabbit this renames `diary_entry.list` to `diary_entry.index`

PR #1918, opened 2018-06-29
[Discussion] Refactors from RFG 2018

benreyn:
This PR contains all of our larger refactor efforts from Ruby for Good 2018. Just oppening this PR to discuss some of the changes. Anything that individually looks good to merge as is we can cheery pick onto master. 

Highlights indclude...

1. Renaming some `view` controller actions to the more Rails standard `show`
2. Rubocop fixes
3. Renaming `UserController` to more standard naming of `UsersController` 
4. Various other flattening of routes and steps toward a more resourceful routing strategy.

Ill answer as many questions about these changes as I can, but feel free to also ping the original authors.

PR #1987, opened 2018-09-10
Selected rubocop changes (ruby for good)

gravitystorm:
This PR pulls in three of the rubocop-related changes from the Ruby For Good refactoring PR (#1918). One of them was already fixed in the meantime.


tomhughes:
I think I mostly hadn't done those because having parentheses on some asserts and not others is horribly ugly :-(

PR #1990, opened 2018-09-10
Reverse welcome banner show/hide logic

### LINKED ISSUE #1951: Welcome box flickers
Usually, the welcome box is shown for new users and should never appear again as soon as the welcome box gets closed (and a _osm_welcome cookie set). The way it is implemented now, the welcome box is always shown, and in case it should be closed, it gets closed about 100ms later. This leads to some annoying flickering on the screen.

Can we change this in a way that the welcome box is always hidden, and only displayed if needed, i.e. turn around today's logic? 

<details>
<summary>Screencast (click to open)</summary>

flicker
</details>

I tried the following approach. However, it's not perfect for mobile, as show() gets immediately overruled by a `!important` css setting in small.scss, resulting again in some flickering :(
Maybe this could be combined with something like: `window.matchMedia("only screen and (max-width:768px)").matches`

<details>

```javascript
diff --git a/app/assets/javascripts/index.js b/app/assets/javascripts/index.js
index ccf4e561c..64aae48bd 100644
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -187,8 +187,8 @@ $(document).ready(function () {
     $.cookie('_osm_location', OSM.locationCookie(map), { expires: expiry, path: '/' });
   });
 
-  if ($.cookie('_osm_welcome') === 'hide') {
-    $('.welcome').hide();
+  if ($.cookie('_osm_welcome') !== 'hide') {
+     $('.welcome').show();
   }
 
   $('.welcome .close-wrap').on('click', function() {
diff --git a/app/assets/stylesheets/common.scss b/app/assets/stylesheets/common.scss
index f9214b24d..dd4a70693 100644
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -613,7 +613,7 @@ body.compact {
 
     #banner,
     .welcome {
-      display: block;
+        display: none;
     }
 
     #sidebar_content {

```

</details>

mmd-osm:
Fixes #1951

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -187,8 +187,10 @@ $(document).ready(function () {
     $.cookie('_osm_location', OSM.locationCookie(map), { expires: expiry, path: '/' });
   });
 
-  if ($.cookie('_osm_welcome') === 'hide') {
-    $('.welcome').hide();
+  if ($.cookie('_osm_welcome') !== 'hide' &&
+     (typeof window.matchMedia !== "function" ||
+     !window.matchMedia("only screen and (max-width:768px)").matches)) {
+    $('.welcome').show();
   }
 
   $('.welcome .close-wrap').on('click', function() {
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -613,7 +613,7 @@ body.compact {
 
     #banner,
     .welcome {
-      display: block;
+      display: none;
     }
 
     #sidebar_content {


app/assets/javascripts/index.js
```diff
@@ -187,8 +187,10 @@ $(document).ready(function () {
     $.cookie('_osm_location', OSM.locationCookie(map), { expires: expiry, path: '/' });
   });
 
-  if ($.cookie('_osm_welcome') === 'hide') {
-    $('.welcome').hide();
+  if ($.cookie('_osm_welcome') !== 'hide' &&
+     (typeof window.matchMedia !== "function" ||
+     !window.matchMedia("only screen and (max-width:768px)").matches)) {
```
tomhughes: This doesn't appear to have anything to do with resolving #1951 so perhaps you could explain what it's about?

How widely supported is that function? I mean I can see you're concerned it might not exist so it's obviously not universal - checking it's type is probably overkill in any case unless we're expecting somebody to create `window.matchmedia` with some other type to annoy us!

mmd-osm: Sure. This may really be kind of an overkill, and it's only useful for mobile devices anyway. I'm perfectly fine to remove those two lines 191+192 again. Their only purpose is to skip `$('.welcome').show()`, if the screen is too small. 

Reason being that small.scss defines the following `display:none` with `!important`. So whatever we try to do in terms of show(), it gets overruled immediately (with some flickering on the mobile screen again).

```
  #sidebar .welcome,
  #sidebar #banner {
    display: none !important;
  }
```

According to https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia#Browser_compatibility, window.matchMedia support is quite good. Line 191 should cover those browsers, where it's not yet supported.

mmd-osm:
Ah, that's way better and avoids the pointless Javascript fiddling in my patch altogether. Thanks!

Issue #1710, opened 2017-12-22
Why are some diff uploads slow?


zerebubuth:
As pointed out in [this comment](https://github.com/openstreetmap/operations/issues/154#issuecomment-353475540) by @mmd-osm, diff uploads are buffered in memory before being processed. However, [another comment](https://github.com/openstreetmap/operations/issues/154#issuecomment-353289168) points out that some transactions can take up to 2 hours to process. This seems strange, given that it should just be parsing the XML (which is slow, but not that slow) and talking to the nearby database (which is also slow, but not that slow).

To investigate further, we need to find a recent diff upload that took a long time to process. That should give us some clues to start with.

Komzpa:
Can we get a slowlog published?

```
shared_preload_libraries = 'auto_explain'
auto_explain.log_min_duration = '100ms'
auto_explain.log_nested_statements = on
auto_explain.log_analyze = on
```

It also seems we can have transactions up to 1M seconds, or 12 days: 
https://munin.openstreetmap.org/openstreetmap/karm.openstreetmap/postgres_querylength_openstreetmap_9_5_main.html

It can also be a reading transaction, say, a forgotten open psql on terminal.

`statement_timeout` and `idle_in_transaction_session_timeout` can help limit it, so it becomes application error traceable in application logs.

To get distribution of queries and keep sanity reading endless postgres logfiles: https://github.com/dalibo/pgbadger

tomhughes:
Please don't confuse things @Komzpa - as you yourself admit there are lots of reasons for long postgres transactions that have nothing to do with changeset uploads.

A slow log isn't going to help unless the delay is down to a single statement and in any case trying to sanitise a slow log for publication is not a reasonable request.

I'm already working on @zerebubuth request so how about we cut out the back seat debugging long enough to let me investigate...

tomhughes:
So to get back to the point, the longest upload I can see yesterday was about 503 seconds or just over 8 minutes.

That was changeset 54833679 and it had a total elapsed time of 503868ms and an activerecord time of 237165.8ms.

tomhughes:
I've now gone back over all the logs we have, which is basically a week, and the longest I see is changeset 54716666 which has an elapsed time of 1028520ms or around 17 minutes and an activerecord time of 524142.0ms.

mmd-osm:
[...]

PR #1995, opened 2018-09-17
add bulk_upload

jiaxuyang:
Key optimization
1. Reduce redundant check
    Since all changes in an upload request belong to the same changeset. It's not necessary to query 
    changesets table for every object in the osmChange file.
2. Skip uniqueness rails validation
    Skip uniqueness validation and that in associated model. Delegate it to database unique index.
3. Bulk check
    Bulk select data from database. Check it in batch.
    There can be dependencies between objects in one batch operation, which result in new problems 
    to solve. Basically, we divide one batch into multi batch.
    a. batch create relation
        If relation1 is a member of relation2, and they are in same create batch. We can't save them in one 
        batch. We have to save relation1 first to get its id, then use this id to replace the placeholder id 
        and save relation2.
    b. batch delete relation
        If relation1 is a member of relation2, and they are in same delete batch. We just delete them all 
        and don't raise an exception for the "used constraint".
4. Bulk save
    a. Delete
        delete from ... where id in (...)
    b. Create
        Use activerecord-import gem to implement conveniently.
    c. Modify
        Postgres supports INSERT ... ON CONFLICT DO UPDATE
5. update changeset bbox
    a. Before: query longitude and latitude for every object, update bbox for every object.
    b. After: query max, min of longitude and latitude once, update bbox once.

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -2,14 +2,14 @@
 #
 # Table name: current_nodes
 #
-#  id           :integer          not null, primary key
+#  id           :bigint(8)        not null, primary key
 #  latitude     :integer          not null
 #  longitude    :integer          not null
-#  changeset_id :integer          not null
+#  changeset_id :bigint(8)        not null
 #  visible      :boolean          not null
 #  timestamp    :datetime         not null
-#  tile         :integer          not null
-#  version      :integer          not null
+#  tile         :bigint(8)        not null
+#  version      :bigint(8)        not null
 #
 # Indexes
 #
@@ -46,8 +46,9 @@ class Node < ActiveRecord::Base
   has_many :containing_relation_members, :class_name => "RelationMember", :as => :member
   has_many :containing_relations, :class_name => "Relation", :through => :containing_relation_members, :source => :relation
 
+  attr_accessor :skip_uniqueness
   validates :id, :uniqueness => true, :presence => { :on => :update },
-                 :numericality => { :on => :update, :integer_only => true }
+                 :numericality => { :on => :update, :integer_only => true }, :unless => :skip_uniqueness
   validates :version, :presence => true,
                       :numericality => { :integer_only => true }
   validates :changeset_id, :presence => true,
@@ -57,7 +58,7 @@ class Node < ActiveRecord::Base
   validates :longitude, :presence => true,
                         :numericality => { :integer_only => true }
   validates :timestamp, :presence => true
-  validates :changeset, :associated => true
+  validates :changeset, :associated => true, :unless => :skip_uniqueness
   validates :visible, :inclusion => [true, false]
 
   validate :validate_position
@@ -137,6 +138,13 @@ class Node < ActiveRecord::Base
     BoundingBox.new(longitude, latitude, longitude, latitude)
   end
 
+  def self.update_changeset_bbox_bulk(changeset, node_ids)
+    rows = Node.select("min(longitude) as min_lon, min(latitude) as min_lat, max(longitude) as max_lon, max(latitude) as max_lat")
+               .where(:id => node_ids.uniq)
+    temp_bbox = BoundingBox.new rows[0]["min_lon"], rows[0]["min_lat"], rows[0]["max_lon"], rows[0]["max_lat"]
+    changeset.update_bbox!(temp_bbox)
+  end
+
   # Should probably be renamed delete_from to come in line with update
   def delete_with_history!(new_node, user)
     raise OSM::APIAlreadyDeletedError.new("node", new_node.id) unless visible
@@ -164,6 +172,61 @@ class Node < ActiveRecord::Base
     end
   end
 
+  def self.delete_with_history_bulk!(nodes, changeset, if_unused = false)
+    node_hash = nodes.collect { |node| [node.id, node] }.to_h
+    skipped = {}
+    Node.transaction do
+      # check if node ids exists
+      node_ids = node_hash.keys
+      old_nodes = Node.select("id, version, visible").where(:id => node_ids).lock
+      raise OSM::APIBadUserInput, "Node not exist. id: " + (node_ids - old_nodes.collect(&:id)).join(", ") unless node_ids.length == old_nodes.length
+      old_nodes.each do |old|
+        unless old.visible
+          # already deleted
+          raise OSM::APIAlreadyDeletedError.new("node", old.id) unless if_unused
+          # if-unused: ignore and do next.
+          # return old db version to client.
+          node_hash[old.id].version = old.version
+          skipped[old.id] = node_hash[old.id]
+          node_hash.delete old.id
+          next
+        end
+        # check if client version equals db version
+        new = node_hash[old.id]
+        new.changeset = changeset
+        old.check_consistency(old, new, changeset.user)
+      end
+      # discover nodes referred by ways or relations
+      way_nodes = WayNode.select("node_id, way_id").where(:node_id => node_hash.keys).group_by(&:node_id)
+      raise OSM::APIPreconditionFailedError, "Node #{way_nodes.first[0]} is still used by ways #{way_nodes.first[1].collect(&:way_id).join(',')}." unless way_nodes.empty? || if_unused
+      way_nodes.each_key do |node_id|
+        skipped[node_id] = node_hash[node_id]
+        node_hash.delete node_id
+      end
+      rel_members = RelationMember.select("member_id, relation_id").where(:member_type => "Node", :member_id => node_hash.keys).group_by(&:member_id)
+      raise OSM::APIPreconditionFailedError, "Node #{rel_members.first[0]} is still used by relations #{rel_members.first[1].collect(&:relation_id).join(',')}." unless rel_members.empty? || if_unused
+      rel_members.each_key do |node_id|
+        skipped[node_id] = node_hash[node_id]
+        node_hash.delete node_id
+      end
+      # modify columns to delete
+      to_delete_nodes = node_hash.values
+      to_delete_nodes.each do |node|
+        # set lat lon not null, won't really modify latitude, longitude, tile.
+        # just a work-around to skip not-null constraint.
+        node.latitude = 0 if node.latitude.nil?
+        node.longitude = 0 if node.longitude.nil?
+        node.tags = {}
+        node.visible = false
+      end
+      # update changeset bbox
+      update_changeset_bbox_bulk(changeset, to_delete_nodes.collect(&:id))
+      # save
+      save_with_history_bulk!(to_delete_nodes, changeset, true)
+    end
+    skipped
+  end
+
   def update_from(new_node, user)
     Node.transaction do
       lock!
@@ -189,6 +252,28 @@ class Node < ActiveRecord::Base
     end
   end
 
+  def self.update_from_bulk(nodes, changeset)
+    Node.transaction do
+      nodes.sort_by!(&:id)
+      node_ids = nodes.collect(&:id)
+      # get id, version to check
+      # get lat, lon to update changeset
+      # lock for update
+      old_nodes = Node.select("id, latitude, longitude, version").where(:id => node_ids).order(:id).lock
+      node_ids.length.times do |i|
+        nodes[i].changeset = changeset
+        old_nodes[i].check_consistency(old_nodes[i], nodes[i], changeset.user)
+        # update changeset bbox with *old* position first
+        changeset.update_bbox!(old_nodes[i].bbox)
+        nodes[i].visible = true
+
+        # update changeset bbox with *new* position
+        changeset.update_bbox!(nodes[i].bbox)
+      end
+      save_with_history_bulk!(nodes, changeset)
+    end
+  end
+
   def create_with_history(user)
     check_create_consistency(self, user)
     self.version = 0
@@ -198,6 +283,15 @@ class Node < ActiveRecord::Base
     changeset.update_bbox!(bbox)
 
     save_with_history!
+  end
+
+  def self.create_with_history_bulk(nodes, changeset)
+    nodes.each do |node|
+      node.version = 0
+      node.visible = true
+      changeset.update_bbox!(node.bbox)
+    end
+    save_with_history_bulk!(nodes, changeset)
   end
 
   def to_xml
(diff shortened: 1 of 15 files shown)


mmd-osm:
It's pretty much the same performance optimization I described in https://github.com/openstreetmap/openstreetmap-website/issues/1710 and also implemented in cgimap. I tested a few easy cases via url rewriting (upload --> bulk_upload), and they seemed ok. The current code seems to be duplicating quite a lot of stuff, i.e. @gravitystorm probably wants to to through the code.

The general concept of bundling many small sql statments into larger ones works, regardless of which language issues the requests. Still there's some more overhead in Rails processing (about factor 5 in one test case).

tomhughes:
Well my point was why is there nothing in the PR to explain what it's about? or even in the commit messages...

Above all else, if it's just optimisations why is it a new method rather then just improving the existing method?

jiaxuyang:
> Well my point was why is there nothing in the PR to explain what it's about? or even in the commit messages...
> 
> Above all else, if it's just optimisations why is it a new method rather then just improving the existing method?

Sorry for making you confused, I will write some descriptions later...
The new bulk_upload method is slightly different from old upload method on response in some case, but won't break original logic. If it's OK, I will make a commit to replace the old one. Thanks!

jiaxuyang:
> It's pretty much the same performance optimization I described in #1710 and also implemented in cgimap. I tested a few easy cases via url rewriting (upload --> bulk_upload), and they seemed ok. The current code introduces quite a bit of duplication, i.e. @gravitystorm probably needs to to through it in much more detail.
> 
> The general concept of bundling many small sql statments into larger ones works regardless of which language issues the sql requests. Still there's some more overhead in Rails processing vs. C++ (about factor 5 in one test case). In any case, we will see at least a 30x speedup.

Yes, I have seen your analysis and effort on cgimap. But it seems the new feature bulk_upload not finished? I just can't wait...

mmd-osm:
> But it seems the new feature bulk_upload not finished? 

It's functionally complete and available for testing on https://upload.apis.dev.openstreetmap.org/
If you find some issue, please create an issue on Github, thanks!

tomhughes:
@jiaxuyang please don't spend a lot of time on this until we can understand what your goals are and decide whether this is a direction we want to pursue - the general expectation has been that @mmd-osm's cgimap work is the way forward.

jiaxuyang:
> until we can understand what your goals are
My goal: improve the performance of diff upload api on rails.
@mmd-osm I will follow any new progress in cgimap project. Hope to use cgimap in the future.

mmd-osm:
It’s probably not a bad idea to have a roughly similar logic in both rails and cgimap, so I wouldn’t discard the idea right from the start based on what’s going on in cgimap. There’s a number of corner cases, and cross validating different implementations would be use case which comes to mind.

tomhughes:
Well the goal was to have rails use the cgimap code as a library I thought?

But sure if it's API compatible then we could take it, but it sounded like it wasn't as there was talk of changes in the response?

Bigger problem is I'm not sure who would review it - the upload code is not something I have any experience of - it was all @zerebubuth's work I think.

mmd-osm:
>> The new bulk_upload method is slightly different from old upload method on response in some case, but won't break original logic.

> if it's API compatible then we could take it, but it sounded like it wasn't as there was talk of changes in the response?

I think the new logic in this pull request first processes all create, modify and finally the delete operations, so it could potentially re-order the sequence in which elements are being processed.

While this isn't an issue for e.g. iD (requests are already sorted in this sequence), other API users might send multiple create/modify/delete blocks in more or less random order. If they expect the {old_id,new_id, version} in the response to appear in the same order as in the request, this may cause some issues. Those cases would have to be validated in more detail. This equally applies to other use cases which work in today's implementation (which works step-by-step from top to bottom). One such use case is already mentioned in point 3. above.


> Bigger problem is I'm not sure who would review it

Indeed, that's the most challenging issue. I struggle quite a bit with the Rails syntax, and can't really add much value here. The feedback so far is mostly based on testing, less so from trying to figure out what the code actually does.

app/models/node.rb
```diff
@@ -164,6 +172,65 @@ def delete_with_history!(new_node, user)
...
   end
 
+  def self.delete_with_history_bulk!(nodes, changeset, if_unused = false)
+    node_hash = nodes.collect { |node| [node.id, node] }.to_h
+    skipped = {}
+    Node.transaction do
+      # check if node ids exists
+      node_ids = node_hash.keys
+      old_nodes = Node.select("id, version, visible").where(:id => node_ids).lock
+      raise OSM::APIBadUserInput, "Node not exist. id: " + (node_ids - old_nodes.collect(&:id)).join(", ") unless node_ids.length == old_nodes.length
```
mmd-osm: Non-existing object should return HTTP 404 (Not found) instead of HTTP 400 (Bad Request)

app/models/relation.rb
```diff
@@ -207,6 +218,54 @@ def delete_with_history!(new_relation, user)
...
+  def self.delete_with_history_bulk!(relations, changeset, if_unused = false)
+    relation_hash = relations.collect { |r| [r.id, r] }.to_h
+    skipped = {}
+    # need to start the transaction here, so that the database can
+    # provide repeatable reads for the used-by checks. this means it
+    # shouldn't be possible to get race conditions.
+    Relation.transaction do
+      relation_ids = relation_hash.keys
+      old_relations = Relation.select("id, version, visible").where(:id => relation_ids).lock
+      raise OSM::APIBadUserInput, "Relation not exist. id: " + (relation_ids - old_relations.collect(&:id)).join(", ") unless relation_ids.length == old_relations.length
```
mmd-osm: Non-existing object should return HTTP 404 (Not found) instead of HTTP 400 (Bad Request)

app/models/way.rb
```diff
@@ -248,6 +300,55 @@ def delete_with_history!(new_way, user)
...
+  def self.delete_with_history_bulk!(ways, changeset, if_unused = false)
+    way_hash = ways.collect { |way| [way.id, way] }.to_h
+    skipped = {}
+    # need to start the transaction here, so that the database can
+    # provide repeatable reads for the used-by checks. this means it
+    # shouldn't be possible to get race conditions.
+    Way.transaction do
+      way_ids = way_hash.keys
+      old_ways = Way.select("id, version, visible").where(:id => way_ids).lock
+      raise OSM::APIBadUserInput, "Way not exist. id: " + (way_ids - old_ways.collect(&:id)).join(", ") unless way_ids.length == old_ways.length
```
mmd-osm: Non-existing object should return HTTP 404 (Not found) instead of HTTP 400 (Bad Request)

mmd-osm:
The following test message currently produces an HTTP 500 Internal Server Errror - _API threw unexpected NoMethodError exception: undefined method `check_consistency' for nil:NilClass_

Expected result would be HTTP 404 not found

```xml
<?xml version="1.0" encoding="UTF-8"?>
<osmChange version="0.6" generator="iD">
  <modify>
    <node id="1200000000000" lon="-11" lat="-46" version="1" changeset="1216"/>
 </modify>
</osmChange>
```

mmd-osm:
I have found another bug:

```
<?xml version="1.0" encoding="UTF-8"?>
<osmChange version="0.6" generator="iD">
  <create>
    <node id="-1" lat="1" lon="2" changeset="1248"/>
    </create>
   <create>
    <node id="-1" lat="1" lon="2" changeset="1248"/>
    </create>
</osmChange>
```

Expected Result:

HTTP 400: Placeholder IDs must be unique for created elements.


Actual Result:

HTTP 200 OK

```
<?xml version="1.0" encoding="UTF-8"?>
<diffResult version="0.6" generator="OpenStreetMap server" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
    <node old_id="-1" new_id="5003081839" new_version="1"/>
    <node old_id="-1" new_id="5003081840" new_version="1"/>
</diffResult>
```

jiaxuyang:
> I have found another bug:
> 
> ```
> <?xml version="1.0" encoding="UTF-8"?>
> <osmChange version="0.6" generator="iD">
>   <create>
>     <node id="-1" lat="1" lon="2" changeset="1248"/>
>     </create>
>    <create>
>     <node id="-1" lat="1" lon="2" changeset="1248"/>
>     </create>
> </osmChange>
> ```
> Expected Result:
> 
> HTTP 400: Placeholder IDs must be unique for created elements.
> 
> Actual Result:
> 
> HTTP 200 OK
> 
> ```
> <?xml version="1.0" encoding="UTF-8"?>
> <diffResult version="0.6" generator="OpenStreetMap server" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
>     <node old_id="-1" new_id="5003081839" new_version="1"/>
>     <node old_id="-1" new_id="5003081840" new_version="1"/>
> </diffResult>
> ```

This bug was bring in duo to the support for multiple action blocks. I have added unit test for this case. Thx.

mmd-osm:
There's another issue with backwards compatibility: the current implementation assumes that every operations occurs from top to bottom. This way, you cannot reference objects which occur only later in the file. Although your implementation is not incorrect from a technical point of view, it allows for scenarios, which are forbidden in the current implementation. If we ever need to switch back to the current implementation, this will no longer be possible.

Here's an example:


```
<?xml version="1.0" encoding="UTF-8"?>
<osmChange version="0.6" generator="iD">
  <create>
    <node id="-1" lat="1" lon="2" changeset="1251"/>
    <way id="-1"  changeset="1251">
        <nd ref="-1"/>
        <nd ref="-2"/>
    </way>
    <node id="-2" lat="1" lon="2" changeset="1251"/>
    </create>
</osmChange>
```

Expected result:

HTTP 400 - Placeholder node not found for reference -2 in way -1

Actual result:

HTTP 200 OK

```
<?xml version="1.0" encoding="UTF-8"?>
<diffResult version="0.6" generator="OpenStreetMap server" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
    <node old_id="-1" new_id="5003081852" new_version="1"/>
    <node old_id="-2" new_id="5003081853" new_version="1"/>
    <way old_id="-1" new_id="4000212054" new_version="1"/>
</diffResult>
```

Furthermore, due to the internal resorting of objects, the `diffResult` result message no longer shows the object in the exact same sequence as they occured in the `osmChange` message. I also consider this to be a breaking change which needs to be corrected.

mmd-osm:
@jiaxuyang  : thanks for the update, looks good so far!

@tomhughes : the functional issues I have reported earlier seem to be fixed now,. I will continue testing the code, though. If you like, you could maybe have a look through the code to see if the coding style is ok, overall code complexity is still reasonable, etc.

tomhughes:
Just to clarify this now replaces the existing upload endpoint rather than adding something in parallel right? Only I seem to see a lot more added code than removed code?

Realistically reviewing this is going to be a huge job, especially for somebody who doesn't really understand the current upload code - my best guess is that is likely a day's work to review it, if not a weekend's. So don't expect me to be able to do it immediately...

mmd-osm:
> Just to clarify this now replaces the existing upload endpoint rather than adding something in parallel right? 

Yes, that's exactly right, it's a drop in replacement for the existing code, only with performance improvements. The results (and error messages) should be exactly the same as in today's version.

> Only I seem to see a lot more added code than removed code?

Bulk operations introduced a lot more complexity due to the additional packaging logic, so this makes sense.

> So don't expect me to be able to do it immediately...

That's perfectly fine, there's no hurry at all. I'm also waiting on @zerebubuth to review the corresponding cgimap implementation.

Overall, this may seem to be doubling efforts. Nevertheless, I still find it useful to also improve the Rails port by using bulk operations (similar to how the cgimap implementation works). cgimap will be faster in the end, but the difference won't be in the 100x range anymore, which is also good news for people who don't use cgimap as acceleration layer.

gravitystorm:
I've had a look through this today. It's a large and complex PR touching one of the most complex parts of the codebase, so it's hard to review. But I want to say that I completely support the idea of refactoring the changeset upload processing to import entities in bulk, in order to improve performance.

I've merged with the current master and fixed the conflicts, since there's been some controller refactoring since this PR was made. You can see the results at https://github.com/gravitystorm/openstreetmap-website/tree/changeset_bulk . I haven't updated this PR, since the tests no longer pass.

It turns out that there is a problem somewhere between this PR and the updated version of [activerecord-import](https://github.com/zdennis/activerecord-import) that we are now using. This PR uses `v0.25.0` and we now use `v1.0.1`. I think it's caused by our use of [composite-primary-keys](https://github.com/composite-primary-keys/composite_primary_keys) which allows belongs_to relations to have a foreign key specified as an array instead of a string, e.g. 

```
belongs_to :old_node, :foreign_key => [:node_id, :version]
```
activerecord-import then blows up [when trying to convert this array to a symbol](https://github.com/zdennis/activerecord-import/blob/38cc26b21969a7c76a3e5b88cb79d6591ee09035/lib/activerecord-import/import.rb#L51). This [was introduced](https://github.com/zdennis/activerecord-import/commit/3e7cd50eac766c09fb8b1ff5f372459248ae2bd1) in activerecord-import `0.28.1`.

So I think there are two things that we should work on here now:

1) Extract the additional changeset upload tests into their own PR. They contain useful checks that @mmd-osm and @jiaxuyang worked on as part of this PR, and will come in useful at some point even if this PR isn't merged.

2) Figure out the best way forward with this foreign key problem. This could involve a combination of 
    * rewriting this PR
    * changes to activerecord-import to restore compatibility with composite-primary-keys
    * rewriting this PR to work with the [upcoming rails 6 "insert_all"](https://github.com/rails/rails/commit/91ed21b304c468db8ce9fd830312c151432935d0)
    * something else entirely

pnorman:
I recommend closing this, as it's had a changes requested label for over three years with no response.

gravitystorm:
I'm not expecting any updates from the OP any more, but there's some useful stuff in here (particularly from my point 1 above) that I'm hoping not to lose into the void.

mmd-osm:
The only relevant test cases seem to be in test/controllers/changeset_controller_test.rb:

copy & paste candidates:

* https://github.com/jiaxuyang/openstreetmap-website/blame/performance-optimization/test/controllers/changeset_controller_test.rb#L1071-L1088
* https://github.com/jiaxuyang/openstreetmap-website/blame/performance-optimization/test/controllers/changeset_controller_test.rb#L1091-L1158
* https://github.com/jiaxuyang/openstreetmap-website/blame/performance-optimization/test/controllers/changeset_controller_test.rb#L1401-L1534

Other tests depend on *bulk methods, which don't exist on master.

mmd-osm:
Testcases are now available as #3821.

PR can be closed.

PR #1996, opened 2018-09-17
Build fix: Removed bogus translations with unknown interpolation variable exit

stefanb:
It seems that some Turkish translations in translatewiki brings in the unknown interpolation variable.

### DIFF
--- a/config/locales/tr.yml
+++ b/config/locales/tr.yml
@@ -2625,8 +2625,6 @@ tr:
           %{directions} yönüne doğru'
         offramp_right_with_name: Sağdaki rampayı kullanarak %{name} girin
         offramp_right_with_directions: '%{directions} yönünde sağdaki rampadan'
-        offramp_right_with_name_directions: '%{exit}''den çıkış yap soldaki %{name}
-          %{directions} yönüne doğru'
         onramp_right_without_exit: Rampadan sağa dönerek %{name} girin
         onramp_right_without_directions: Rampaya doğru sağa dönün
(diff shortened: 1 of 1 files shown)


tomhughes:
I already removed those once, and will continue to do so as I deploy each change if they come back - there's no need to submit PRs for this.

gravitystorm:
@stefanb Thanks for picking this up. To elaborate on Tom's point, this needs to be properly fixed upstream in Translatewiki in the turkish translations.

tomhughes:
I did already let @Nikerabbit know when I fixed it up last week.

Nikerabbit:
I had messages the translator, but no response.

I can put the English text there and keep it marked as outdated, if that helps?

Nikerabbit:
I also filed https://phabricator.wikimedia.org/T204568 to address the root cause.

tomhughes:
In this case I think the best thing is to reset it to the "untranslated" state in the hope that somebody will redo it properly.

I'm not sure if that means removing it or resetting and marking as outdated? I'm not clear on the details of how an untranslated string is represented - from out point of view the ideal is to have that string no longer output in the tr locale so that we will fall back to en until it is retranslated.

Nikerabbit:
Removing translation is hard. It needs a synchronized removal both in the repo and in the wiki. Perhaps I can try that on next import.

stefanb:
@Nikerabbit, thanks for reporting the bug upstream!
@tomhughes, thank you for fixing it in an identical way :)

Issue #1944, opened 2018-08-05
Remove hosting reference to Imperial College


pnorman:
We've moved the site away from there. `partners_ic` is the name of the string.

tomhughes:
This is tricky to fix because of the way the interpolation works... We really need to fix it in all languages simultaneously :-(

tomhughes:
Oh sorry, didn't mean to close it - must have hit the wrong button.

pnorman:
> This is tricky to fix because of the way the interpolation works... We really need to fix it in all languages simultaneously :-(

Ya, this is why I didn't just go and edit it, I couldn't easily figure it out.

> To show a bit of goodwill, I'm wondering if IC could be moved to some kind of "Alumni" partner section, e.g. partners, who have supported OSM over many years in the past.

They're listed on the hardware thanks page at https://hardware.openstreetmap.org/thanks/ but the main place the website shows the list is to new users in the notice box, where space is *very* limited.

![image](https://user-images.githubusercontent.com/1190866/43714170-017a6448-9931-11e8-815b-12317aaa0151.png)

mmd-osm:
> the main place the website shows the list is to new users in the notice box, where space is very limited

Ok, I think I didn't quite get your original intent somehow. I really had the _about page_ in mind, which also uses that same  `partners_ic` text: https://www.openstreetmap.org/about . There's also a direct link from the osm main page.

PR #2003, opened 2018-09-21
Add links to ToU

simonpoole:
This is WIP . The intent is to do a first pass at linking to the ToU (About page done, a good solution for the signup page needs to be found). This is not a replacement for actually integrating the ToU in to the signup process, it simply serves to make them discoverable.


app/views/site/about.html.erb
```diff
@@ -34,7 +34,6 @@
     <h2><div class='icon partners'></div><%= t ".partners_title" %></h2>
     <p><%= t 'layouts.partners_html',
              :ucl => link_to(t('layouts.partners_ucl'), "https://www.ucl.ac.uk"),
-             :ic => link_to(t('layouts.partners_ic'), "https://www.imperial.ac.uk/"),
```
tomhughes: As previously discussed in #1944 it's not safe to remove IC like this as it will break every language that still has a `%{ic}` marker in the translation.

simonpoole: I would suggest simply renaming layouts.partners_html in that case (the version without IC), not the most elegant solution, but in the grand scheme of things I don't think that really matters.

gravitystorm:
I'm of course happy to add a link to the terms of use, but I think this PR needs to be closed and a fresh one opened with just that specific change in it. 

This PR unfortunately contains two unrelated changes (adding an 'android_apps' view, and the crediting changes @tomhughes picked up on) and has conflicts and test failures.

tomhughes:
Note that the credit change has now been made in https://github.com/openstreetmap/openstreetmap-website/commit/ad3ca54e286d8bb733afb96b64d421c07faa358b.

simonpoole:
a) this was explicitly declared as work in progress and wasn't intended to be merged as is
b) the "unrelated" text changes were motivated by the need reduce the amount of text, primarily because we should be adding the text to the welcome "box" too where space is a concern.

PR #2006, opened 2018-09-26
Added /notes/mine url path

### LINKED ISSUE #1997: Link '/notes/mine' to my own notes
Hi,

It would be very convenient to have 'osm.org/notes/mine' refer to my (the logged in user) notes - just like osm.org/traces/mine.

Is this possible?

jguthrie100:
Converted `notes#mine` to work more like `traces#mine`

Renamed `notes#mine` to `notes#user` and changed associated validations etc

Fixes #1997


test/controllers/notes_controller_test.rb
```diff
@@ -121,7 +121,11 @@ def test_routes
 
     assert_routing(
       { :path => "/user/username/notes", :method => :get },
-      { :controller => "notes", :action => "mine", :display_name => "username" }
+      { :controller => "notes", :action => "user", :display_name => "username" }
+    )
+    assert_routing(
+      { :path => "/traces/mine", :method => :get },
+      { :controller => "traces", :action => "mine" }
```
tomhughes: I think you mean `notes` here ;-)

app/controllers/notes_controller.rb
```diff
@@ -276,12 +277,20 @@ def search
...
     if params[:display_name]
       if @user = User.active.find_by(:display_name => params[:display_name])
         @params = params.permit(:display_name)
-        @title = t "notes.mine.title", :user => @user.display_name
-        @heading = t "notes.mine.heading", :user => @user.display_name
+
+        # set title/heading
+        if current_user && current_user.display_name == @user.display_name
+          @title = t "user.show.my notes"
+          @heading = t "user.show.my notes"
```
tomhughes: I know that traces does something similar, but I wonder if this is sensible/worthwhile...

I mean `/mine` is basically a shortcut to asking for a particular user that isn't even preserved in the URL because we redirect, so giving this page a different title if you're logged in seems a little odd and not really necessary?

jguthrie100: I'm only copying what it does in the traces controller.. it basically just says "My Notes" if you're logged in instead of "Harry's Notes"

test/controllers/notes_controller_test.rb
jguthrie100: Oops, yes I do!

app/controllers/notes_controller.rb
jguthrie100: Excerpt from traces controller:

```
@title = if target_user.nil?
               t ".public_traces"
             elsif current_user && current_user == target_user
               t ".my_traces"
```

jguthrie100: Just noticed that you referenced the traces page already, but for me I think its a worthwhile addition.
Throughout the whole site there are references to `My Profile` and `My Traces` and `My Settings` etc, so I'd say its only apt to apply the same to the Notes page!

gravitystorm:
This PR was superseded by https://github.com/openstreetmap/openstreetmap-website/pull/2699

tomhughes:
It's not really true that #2699 superseded this, because it didn't actually implement `/notes/mine` as a route!

PR #2009, opened 2018-10-02
Turn changeset border darker orange on hover

### LINKED ISSUE #1515: Change border colour when highlighting changeset in history
When you mouseover a changeset in the history list, the fill-opacity of the changeset is adjusted in the main map. However when you have various changesets on top of each other, and varying in size, it's sometimes difficult to see the opacity change.

I'd like to suggest the orange border is changed to a different colour to make it more obvious which changeset you are highlighting - perhaps red or blue?

jguthrie100:
Fixes #1515

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -38,12 +38,12 @@ OSM.History = function(map) {
   };
 
   function highlightChangeset(id) {
-    group.getLayer(id).setStyle({fillOpacity: 0.3, color: "#0095AA"});
+    group.getLayer(id).setStyle({fillOpacity: 0.3, color: "#FF6600", weight: 3});
     $("#changeset_" + id).addClass("selected");
   }
 
   function unHighlightChangeset(id) {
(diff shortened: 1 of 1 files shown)


jguthrie100:
Fixes #1515

jguthrie100:
No hover:
image

Hover:
image

tomhughes:
That's kind of going from "subtle" to "slap in the face with a house brick" really though...

It also looks a bit weird having a blue border with a yellow fill like that I think?

jguthrie100:
What would be preferable?
A slightly darker shade of orange? A thicker border?

It makes sense to change it - just to what..

boothym:
The issue is when you have many similarly sized changesets in the same area - what's "subtle" no longer works at highlighting which changeset is the one you've moused over in the history list. Changing the fill won't really work as it's covered by other changesets, it's the border which is important.

I suggested blue so that it's contrasting enough and shouldn't cause problems those who are colour blind.

PR #2010, opened 2018-10-02
Set minimum changeset boundary size

### LINKED ISSUE #2008: Minimum size for changeset box
If a changeset is only an edit of a single node, then it's basically impossible to select this changeset from the map. There should be a minimum 10-20px padding on the changeset box around the node so that it can be clicked on.

It's a problem especially when viewing a user's edits list and you see this small orange dot - but you can't click on it. (I did manage eventually when using the zoom function at 200%)

Would also be useful to fix in conjunction with #1515 to make changesets more obvious.

jguthrie100:
Fixes #2008

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -97,17 +97,26 @@ OSM.History = function(map) {
     $("[data-changeset]").each(function () {
       var changeset = $(this).data('changeset');
       if (changeset.bbox) {
-        var latModifier = 0;
-        var lonModifier = 0;
+        var latWidth = changeset.bbox.maxlat - changeset.bbox.minlat;
+        var lonWidth = changeset.bbox.maxlon - changeset.bbox.minlon;
+        var minLatWidth = 0.0004;
+        var minLonWidth = 0.0008;
 
-        if (changeset.bbox.minlat === changeset.bbox.maxlat || changeset.bbox.minlon === changeset.bbox.maxlon) {
-          latModifier = 0.0001;
-          lonModifier = 0.0002;
+        var bounds = [[changeset.bbox.minlat, changeset.bbox.minlon],
+          [changeset.bbox.maxlat, changeset.bbox.maxlon]];
+
+        if (latWidth < minLatWidth) {
+          bounds[0][0] -= ((minLatWidth - latWidth) / 2);
+          bounds[1][0] += ((minLatWidth - latWidth) / 2);
+        }
+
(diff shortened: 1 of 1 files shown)


jguthrie100:
Current:
image

Modified:
image

tomhughes:
That doesn't really set a minimum size though, does it?

To start with it only operates if the size is zero, not if the size is greater than zero but below the minimum.

Also it changes both dimensions even if only one dimension is too small.

jguthrie100:
@tomhughes 

How should the actual changeset rendering be handled, once the link has been clicked through from the history page. At the moment, its still the minimum size: https://www.openstreetmap.org/changeset/63058122

From what I can tell, the changeset is rendered automatically using internal Leaflet GeoJSON code, based on data returned from the OSM API. So the API returns the tiny changeset bounds, and GeoJSON renders it automatically.

So to fix it, the actual data from the API has to be manipulated, either in the controller (so if the changeset has equal minlat/maxlat or minlon/maxlon values, then a minimum sized changeset is sent back instead), or by manipulating the xml data that is returned.

How would you handle it? Is it worth changing the API output so that the minimum changeset bounds are changed in the data layer before being sent back from the API, rather than being manipulated in the view layer?

Or should it just be left as it is and ignored for now?

tomhughes:
Can't you just manipulate the feature array after it is parsed by `buildFeatures` and before it is added to the map with `addData`?

gravitystorm:
If possible I'd prefer to see the minimum box size specified in pixels, since that's what we're really interested in for viewing and clicking. If the bounding box is e.g. 0.0004 degrees across, then that will be still to small to see on some zoom levels, unnecessarily large on other zoom levels, and varied widths depending on the latitude.

mmd-osm:
Minimum pixel size would indeed have some advantages...

See this example https://upload.apis.dev.openstreetmap.org/history#map=19/47.39142/8.37422 

Once I click on the bbox, it immediately shrink to half its size: https://upload.apis.dev.openstreetmap.org/changeset/109

jguthrie100:
I wrote a fix for the history tab in #2018 
If it's approved I can replicate the changes for the actual changeset as well

mmd-osm:
Oh, sorry, missed that one... :(

uswoods:
This change is unwanted from the community: See https://forum.openstreetmap.org/viewtopic.php?id=64109

mmd-osm:
@uswoods :  that’s not what was said that forum post.

PR #2018, opened 2018-10-06
Set minimum changeset size in pixels

jguthrie100:
Changed the minimum changeset width to be in pixels rather than degrees.

When zooming out, the minimum pixel width is maintained, so even when max-zoomed out, the pixel width will be 40px (width could be easily edited)

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -97,23 +97,27 @@ OSM.History = function(map) {
     $("[data-changeset]").each(function () {
       var changeset = $(this).data('changeset');
       if (changeset.bbox) {
-        var latWidth = changeset.bbox.maxlat - changeset.bbox.minlat,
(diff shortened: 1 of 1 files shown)


jguthrie100:
image

image

image

image


To be honest, I'm not sure if its an improvement or not. It seems a bit unintuitive when zoomed out

jguthrie100:
image

Its not so bad at distance actually when you consider how small the different changesets would be

jguthrie100:
Actually, I take it back. There is a considerable improvement to the current implementation which shows just tiny dots!

image

uswoods:
_To be honest, I'm not sure if its an improvement or not. It seems a bit unintuitive when zoomed out._

Indeed, sir! You should better revert this change.

tomhughes:
I wasn't aware that two people in a forum thread were "the community". Is there anything else they'd like a veto on?

Issue #1626, opened 2017-08-30
Use an authorisation framework


gravitystorm:
Our codebase currently has a home-grown permissions system. The logic of who can do what is built into every controller action, which is becoming unwieldy, particularly when we want to allow multiple types of user to do something (e.g. allow the user, moderators and admins to delete something). We also have a "default allow" system, where each action is permitted to all users, unless we add permissions checks. I'd like to make it easier to add more roles (e.g. like allowing people to moderate diaries without being able to block users). And as with much of what I work on here, I'd like to tack towards standard solutions.

So I'd like to move to a third-party authorisation framework. I'd like to see in particular:
* A default deny policy, so that actions aren't accidentally accessible.
* A centralised authorisation configuration, so that permissions aren't scattered all over the site.
* A system that's powerful enough that we can write complex authorisation rules.

I'm particularly fond of [decl_auth](https://github.com/stffn/declarative_authorization/) but unfortunately that is now unmaintained and none of the forks have gathered any momentum. The best alternative that I've found is [CanCanCan](https://github.com/CanCanCommunity/cancancan), which behaves similarly to decl_auth, but their Ability class looks a bit clunky compared to the decl_auth authorisation_rules DSL.

https://www.ruby-toolbox.com/categories/rails_authorization.html has a list of alternatives.

Because this will be a bit of an invasive surgery, I'd like to get other people's thoughts before jumping in at the deep end.

gravitystorm:
For a non-trivial but also not-over-the-top example of a central rules file, I've previously worked on this one (using decl_auth):

https://github.com/cyclestreets/cyclescape/blob/staging/config/authorization_rules.rb

tomhughes:
Sounds like a good idea to me, and I agree that CanCanCan looks like it's probably the best option from what I've seen so far.

I have to say that having read quite a bit of the CanCanCan documentation now I still don't entirely understand how the ability objects get created and attached to the request so that he controllers and views can see them, but I'm sure there's something entirely obvious that I've missed that you will be able to figure out ;-)

gravitystorm:
For reference I [made a start](https://github.com/gravitystorm/openstreetmap-website/tree/cancancan) on integrating CanCanCan a few weeks ago, and the proof of concept worked (just allowing sign in and viewing the /welcome page). 

What became quickly apparent is that the whole thing will be much easier to implement if we refactor the majority of the routes first, to use resourceful routing much more widely, so that's what I'm working towards.

[...]

PR #2019, opened 2018-10-07
move #require_administrator to application controller

### LINKED ISSUE #1825: Move require_administrator to application_controller
There's a require_moderator in application_controller, but no corresponding require_administrator. That hasn't stopped people repeatedly wanting it...

```
$ grep -rn 'def require_administrator' app/
app/controllers/user_controller.rb:740:  def require_administrator
app/controllers/user_roles_controller.rb:27:  def require_administrator
app/controllers/diary_entry_controller.rb:236:  def require_administrator
```

wgoh:
Fixes #1825

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -140,7 +140,7 @@ class ApplicationController < ActionController::Base
   def require_administrator
     if current_user
       unless current_user.administrator?
-        flash[:error] = t("user.filter.not_an_administrator")
+        flash[:error] = t("users.filter.not_an_administrator")
         redirect_to user_path(current_user)
       end
     else
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -135,6 +135,20 @@ class ApplicationController < ActionController::Base
   end
 
   ##
+  # require that the user is a administrator, or fill out a helpful error message
+  # and return them to the user/login page.
+  def require_administrator
+    if current_user
+      unless current_user.administrator?
+        flash[:error] = t("user.filter.not_an_administrator")
+        redirect_to user_path(current_user)
+      end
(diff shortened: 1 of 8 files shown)


wgoh:
@gravitystorm is this the intended behaviour? if it is I will move on to fix the tests. also, let me know if I am not following the correct procedure as this is my first time contributing here. :)

app/controllers/application_controller.rb
```diff
@@ -134,6 +134,20 @@ def require_moderator
     end
   end
 
+  ##
+  # require that the user is a administrator, or fill out a helpful error message
+  # and return them to the user/login page.
+  def require_administrator
+    if current_user
+      unless current_user.administrator?
+        flash[:error] = t("user.filter.not_an_administrator")
```
tomhughes: That needs to be `users.filter.not_an_administrator` as that key doesn't exist - looks like you chose to copy the one that is broken since @gravitystorm renamed `user` to `users` last week ;-)

The `user_role` version can also be removed from `config/locales/en.yml` as it will no longer be used once this is merged.

tomhughes: Incidentally the fact we hadn't picked that up means that the "need administrator" path in the diary entry controller is not being tested, so if you feel like adding a test for that it would be great ;-)

wgoh:
@tomhughes thanks for your feedback. could you advise on how to fix the tests for `test/integration/user_roles_test.rb`? thanks!

gravitystorm:
Unfortunately other refactoring work means that this PR is no longer required. We've already removed all the require_administrator methods as part of the project to [use CanCanCan](https://github.com/openstreetmap/openstreetmap-website/issues/1626) 

Thanks anyway @wgoh !

PR #2023, opened 2018-10-10
Move to CanCanCan for authorization

### LINKED ISSUE #1626: Use an authorisation framework

gravitystorm:
Resolves #1626. Builds on and replaces #1904 

I've taken #1904, brought it up to date, and resolved a couple of things that I'd noticed and added a few more refactorings, including the first use of the `can?` in the views.

At this point, do we want to merge what we have already and then refactor the rest of the controllers in subsequent PRs, or should we wait until we're ready with a comprehensive PR that covers all controllers?


app/models/ability.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+class Ability
```
tomhughes: Is `app/model` the right place for this? It doesn't look like a model and the CanCanCan documentation just refers to is as a "class" so I was expecting it to be in `lib` but I guess this is where the generator put it?

app/models/ability.rb
```diff
@@ -0,0 +1,57 @@
...
+  include CanCan::Ability
+
+  def initialize(user)
+    can :index, :site
+    can [:permalink, :edit, :help, :fixthemap, :offline, :export, :about, :preview, :copyright, :key, :id], :site
```
tomhughes: Can these not be merged into one?

Is the ability to use a symbol as the last argument to refer to actions in a controller that does correspond to a model documented anywhere? Everything I've seen just refers to using a class name there to authorize access to objects of that class?

app/controllers/application_controller.rb
```diff
@@ -466,6 +468,23 @@ def better_errors_allow_inline
     raise
   end
 
+  def current_ability
+    Ability.new(current_user).merge(granted_capability)
```
tomhughes: This has lost the memoization of the default method, which has `@current_ability ||=` as a prefix - is that deliberate?

app/controllers/application_controller.rb
```diff
@@ -466,6 +468,23 @@ def better_errors_allow_inline
...
+  end
+
+  def granted_capability
+    Capability.new(current_user, current_token)
+  end
+
+  def deny_access(_exception)
+    if current_user
```
tomhughes: This should be `current_token` here I think? Otherwise we wind up talking about OAuth when it wasn't used - the current code is guarded by a token test at least.

If we did that and then had a suitably generic case after it for `current_user` then I think we could get rid of many of the special case overrides for `deny_access` in the other controllers - most of those are pretty much impossible to hit anyway so don't really need to give the precise detail they do now.

app/models/capability.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class Capability
```
tomhughes: Obviously similar comments apply here as with `Ability` but this one I think is something of your invention as I don't see it mentioned in the documentation?

I wonder if this is even needed - given we are overriding `current_ability` could we not pass both the user and token to `Ability` and have it do everything?

app/models/ability.rb
gravitystorm: Yes, this is where the generator puts it. I'm happy to put it where ever suits - perhaps in config? I'm not sure about lib, I consider that somewhere where stuff goes that (in theory) could be extracted into a gem, but I can see the logic there since we put various classes into lib already.

From https://github.com/CanCanCommunity/cancancan/blob/develop/lib/generators/cancan/ability/USAGE: 

>The cancan:ability generator creates an Ability class in the models
  directory. You can move this file anywhere you want as long as it
is in the load path.

app/models/ability.rb
gravitystorm: I'll merge these, good spot. The symbol usage and `class => false` is from https://github.com/CanCanCommunity/cancancan/wiki/Non-RESTful-Controllers

app/models/ability.rb
tomhughes: It's tricky... I'm not sure rails even creates a `lib` directory these days, and I know it was removed from the default load but we put it back in `config/application.rb`.

app/controllers/application_controller.rb
gravitystorm: Not deliberate by me. But adding in the memoization causes the tests to fail, and I would need to investigate why.

app/controllers/application_controller.rb
gravitystorm: I'm in general happy with telling people that they don't have permission (rather than what permission they need) to avoid all the special case overrides. I agree there's something not quite right on handling tokens vs current_user here.

app/models/capability.rb
gravitystorm: I don't have any views on this, beyond what @cflipse wrote in the [commit message](https://github.com/openstreetmap/openstreetmap-website/pull/2023/commits/91fc65a2e3ad47414344a6d5dc03cb5fb2a08ce1) when he split them out.

To be blunt, I don't yet have a good understanding of how all the tokens stuff works, so I'm happy to take direction here. My experience elsewhere is just with straightforward approaches around having a current_user with particular roles, and our token handling here is more complex.

app/controllers/application_controller.rb
gravitystorm: The tests fail because in our controller tests, the controller state (e.g. instance variables) is not reset between each "request" - see https://github.com/rails/rails/issues/24566 . Calling `get` in a controller test isn't really making an http request, it just calls the methods on the controller object which exists for the length of the test.

So for a simple example, `site_controller_test#test_welcome` calls `get` twice. The memoized result from the first call means that the second logged-in call is also forbidden. This is, ultimately, because you're not really meant to make multiple requests in a single controller testcase - that should be in integration or system tests. But we do this a **lot** in our tests.

gravitystorm:
I've pushed a few more changes, in particular a slight reworking of the token handling which I think makes the behaviour more obvious.

I also tried to convert another couple of controllers, but realised that there's a few edge cases in each case which deserve their own PR, so I don't think I'll expand the scope of this any further yet!

gravitystorm:
My brain is melting from trying to understand all this, but I'm almost on top of it now. I've used the same check as setup_user_auth, for consistency, and removed the user check from Capabilities. I hope this is a valid approach!

tomhughes:
I think so, and long term it should mean we can get rid of `setup_user_auth` and having to remember to call it on requests that may need to be authorised and just have the CanCan setup trigger the checks automatically.

Fully achieving that will require also handling basic auth when setting up abilities. Plus we'll have to figure out what to do with the user blocks check...

gravitystorm:
I think this is ready for final review and/or merging now.

app/models/ability.rb
benreyn: Just weighing in, FWIW, Ive seen a folder created for these and seen them put in `app/abilities/` before. It would also make [splitting the ability file](https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities:-Best-Practices#split-your-abilityrb-file) (which are likely to want to do later) much easier.

cflipse: `/lib` is almost certainly the wrong place, as this isn't external / library code.  I could see putting it in config, but it's active code that gets checked, and config isn't the first place I'd look.  Ben's thought about app/abilities appeals, if the plan is to deconstruct the large object at some point ...

app/models/capability.rb
cflipse: It's been a while.  IIRC, the capabilities reflect permissions granted to the application -- so, if it's making a request to access your GPS, and you say "yes", then you've granted the app that capability.  Another common example is when you OAuth login and the system asks for permission to read your contact lists.

This is, more or less, inverse of the CanCanCan's normal idea of an Ability, which is the app deciding if the user has permission to do something; Capability is the user deciding if the app has permission to do something.  (Capability was an inherited name, I suspect that something better could be determined)  The end result of the calculation is the same, but separating them helps to keep them from getting too crossed.

app/models/ability.rb
tomhughes: I did do some googling before and there seems to be approximately no consensus on where to put things that aren't one of the "standard" types of class.

I'd always seen `lib` as being for random bits of non-classifiable code and not just for external things (which wouldn't normally be in the repo at all) and in the early days when `lib` was on the autoload path I think that was a more common view but then it got removed.

I didn't even know could just create random directories under `app` to be honest - how does that work? Does rails add every directory there to the autoload path?

cflipse: It does.

there are _some_ additional app directories that are something approaching a standard:  `app/presenters`, `app/forms`, `app/services`  (especially the last one) but it's pretty much open season for organizing things under the top layer of app.  Rails will loop through those directories and add them to autoload at boot time (so, if you add a _new_ `app/foo` class, you'd need to reboot your dev server, but otherwise, classes underneath are subject to Rails' loading magicks)

Issue #1878, opened 2018-05-24
redesign /copyright to unify the wiki Contributors page


andrewharvey:
pulling out the discussion from #1480 about changes to the /copyright page to support displaying a large number of contributors/attributions.

>...Essentially we simply need to code a solution for this that will remove the current very artificial two tier system, which likely means we will have collapse-able / expandable lists on the copyright page and some way import the contents...

/cc @simonpoole

PR #2026, opened 2018-10-14
add australia copyright notice for PSMA Australia

andrewharvey:
The [Australian community wants to import](https://lists.openstreetmap.org/pipermail/talk-au/2018-October/012085.html) suburb, locality and municipality admin boundaries into OSM from a government CC BY 4.0 source, as [documented](https://wiki.openstreetmap.org/wiki/Import/Catalogue/PSMA_Admin_Boundaries), and we'll need this notice in place before the import can happen.

The source did agree to and complete the the OSMF's CC BY waiver, unfortunately it was the first revision which mentioned openstreetmap.org/copyright instead of the wiki contributors page. It took 6 months passing through multiple bureaucrats, so I'd rather not try to re-negotiate this to the wiki contributors page.

Hopefully it's okay to add a new contributor to this page for this substantial contribution, until we can address #1878

### DIFF
--- a/app/views/site/copyright.html.erb
+++ b/app/views/site/copyright.html.erb
@@ -56,6 +56,7 @@
 <p><%= t ".legal_babble.contributors_intro_html", :locale => @locale %></p>
 <ul id="contributors">
   <li><%= t ".legal_babble.contributors_at_html", :locale => @locale %></li>
+  <li><%= t ".legal_babble.contributors_au_html", :locale => @locale %></li>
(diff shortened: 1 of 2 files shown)


tomhughes:
I believe LWG are generally opposed to adding anything new here cc @simonpoole

simonpoole:
Yes, but in this specific case we, (the LWG), screwed up a bit by having it wrong in the waiver so I think an exception is OK. WHich doesn't mean that we wont address the issue in general when there is time.

PR #2028, opened 2018-10-20
Add links to the ToU and include them in signup

simonpoole:
- add links to the ToU in the Welcome box and on the About page
- add tou_agreed to users table
- require checking a checkbox that the ToU have been read for signup

Missing:

~~- [ ] tests~~

### DIFF
--- a/app/assets/javascripts/user.js
+++ b/app/assets/javascripts/user.js
@@ -126,7 +126,6 @@ $(document).ready(function () {
   });
 
   $("#read_tou").on("click", function () { 
-    var agreebtn = document.getElementById('agree');
-    agreebtn.disabled=!this.checked;
+    $("#agree").prop("disabled", !$(this).prop("checked"));
   });
 });
--- a/app/assets/javascripts/user.js
+++ b/app/assets/javascripts/user.js
@@ -125,11 +125,11 @@ $(document).ready(function () {
     $("#contributorTerms").load(url);
   });
 
-  $("#read_ct").on("click", function () { 
+  $("#read_ct").on("click", function () {
     $("#continue").prop("disabled", !($(this).prop("checked") && $("#read_tou").prop("checked")));
   });
 
-  $("#read_tou").on("click", function () { 
+  $("#read_tou").on("click", function () {
     $("#continue").prop("disabled", !($(this).prop("checked") && $("#read_ct").prop("checked")));
   });
 });
--- a/app/assets/javascripts/user.js
+++ b/app/assets/javascripts/user.js
@@ -125,7 +125,11 @@ $(document).ready(function () {
     $("#contributorTerms").load(url);
   });
 
+  $("#read_ct").on("click", function () { 
+    $("#continue").prop("disabled", !($(this).prop("checked") && $("#read_tou").prop("checked")));
+  });
+
   $("#read_tou").on("click", function () { 
-    $("#agree").prop("disabled", !$(this).prop("checked"));
+    $("#continue").prop("disabled", !($(this).prop("checked") && $("#read_ct").prop("checked")));
   });
 });
--- a/app/assets/javascripts/user.js
+++ b/app/assets/javascripts/user.js
@@ -124,4 +124,9 @@ $(document).ready(function () {
     $("#contributorTerms").html("<img src='" + OSM.SEARCHING + "' />");
     $("#contributorTerms").load(url);
   });
+
+  $("#read_tou").on("click", function () { 
+    var agreebtn = document.getElementById('agree');
+    agreebtn.disabled=!this.checked;
+  });
 });
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -53,6 +53,7 @@ class UsersController < ApplicationController
     elsif current_user
       unless current_user.terms_agreed?
         current_user.consider_pd = params[:user][:consider_pd]
+        current_user.tou_agreed = Time.now.getutc
         current_user.terms_agreed = Time.now.getutc
         current_user.terms_seen = true
 
@@ -73,6 +74,7 @@ class UsersController < ApplicationController
         current_user.creation_ip = request.remote_ip
         current_user.languages = http_accept_language.user_preferred_languages
         current_user.terms_agreed = Time.now.getutc
+        current_user.tou_agreed = Time.now.getutc
         current_user.terms_seen = true
 
         if current_user.auth_uid.blank?
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -36,7 +36,7 @@ class UsersController < ApplicationController
   def save
     @title = t "users.new.title"
 
-    if params[:decline]
+    if params[:decline] || !params[:read_tou] || params[:read_tou] == 0
       if current_user
         current_user.terms_seen = true
 
@@ -47,8 +47,10 @@ class UsersController < ApplicationController
         else
           redirect_to :action => :account, :display_name => current_user.display_name
         end
-      else
+      elsif params[:decline]
         redirect_to t("users.terms.declined")
+      else
+        redirect_to :action => :terms
       end
     elsif current_user
       unless current_user.terms_agreed?
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -36,7 +36,7 @@ class UsersController < ApplicationController
   def save
     @title = t "users.new.title"
 
-    if params[:decline] || !params[:read_tou]
+    if params[:decline] || !(params[:read_tou] && params[:read_ct])
       if current_user
         current_user.terms_seen = true
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -36,7 +36,7 @@ class UsersController < ApplicationController
   def save
     @title = t "users.new.title"
 
-    if params[:decline] || !params[:read_tou] || params[:read_tou] == 0
+    if params[:decline] || !params[:read_tou] || params[:read_tou] == "0"
       if current_user
         current_user.terms_seen = true
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -36,7 +36,7 @@ class UsersController < ApplicationController
   def save
     @title = t "users.new.title"
 
-    if params[:decline] || !params[:read_tou] || params[:read_tou] == "0"
+    if params[:decline] || !params[:read_tou]
       if current_user
         current_user.terms_seen = true
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -34,6 +34,7 @@
 #  image_content_type  :string
 #  auth_provider       :string
 #  home_tile           :integer
+#  tou_agreed          :datetime
 #
 # Indexes
 #
--- a/app/views/users/terms.html.erb
+++ b/app/views/users/terms.html.erb
@@ -9,7 +9,12 @@
 
 <%= form_tag({ :action => "save" }, { :class => " inner22 standard-form fillL" }) do %>
   <!-- legale is <%= @legale %> -->
+  <p class="deemphasize"><%= t '.read and accept' %></p>
+  <label class="standard-label">
+    <%= t '.heading_ct' %>
+  </label>
   <div class='form-row horizontal-list'>
+    <p class="deemphasize"><%= t '.contributor_terms_explain' %></p>
     <label class="standard-label">
       <%= t ".legale_select" %>
     </label>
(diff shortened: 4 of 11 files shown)


simonpoole:
http://upc.poole.ch:3000/ is running this right now (against a local test database) if anybody wants to see if the way that this works now is OK.

simonpoole:
@tomhughes the change on the about page is on line 1274 on en.yml (I needed some squinting at the screen to refind it :-)).

app/views/users/terms.html.erb
```diff
@@ -28,27 +30,36 @@
...
+  <label class="standard-label">
+    <%= t 'layouts.tou' %>
+  </label>
+  <p class="deemphasize"><%= t '.tou_explain' %></p>
   <div class="form-row">
-    <label for="user_consider_pd">
-      <%= check_box('user', 'consider_pd') %>
-      <%= t '.consider_pd' %>
+    <label for="read_tou">
+      <%= check_box_tag 'read_tou',1,false,:onclick=>"agreebtn = document.getElementById('agree');agreebtn.disabled=!this.checked;"%>
```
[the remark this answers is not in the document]
simonpoole: I was expecting so much, is there a preferred js file to put the handler in?

config/locales/en.yml
```diff
@@ -1989,10 +1991,14 @@ en:
...
       terms declined url: https://wiki.openstreetmap.org/wiki/Contributor_Terms_Declined
     terms:
-      title: "Contributor terms"
-      heading: "Contributor terms"
-      read and accept: "Please read the agreement and press the agree button to confirm that you accept the terms of this agreement for your existing and future contributions."
-      consider_pd: "In addition to the above agreement, I consider my contributions to be in the Public Domain"
+      title: "Terms"
+      heading: "Terms"
+      heading_ct: "Contributor terms"
+      read and accept: "Please read the contributor agreement and the terms of use, check the checkbox when done and then press the agree button to confirm that you accept the terms of both documents."
```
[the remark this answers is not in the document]
simonpoole: We were moving the text around a lot and I expect some more changes when we've got some feed back. I'll give all the changed strings a once over and change the names of those that we can clearly not reuse as soon as I have a good feeling about the whole thing.

tomhughes:
@simonpoole I did actually try looking at it on your test instance and couldn't see the link but now that I understand the change that was likely because I was seeing the en-GB version of the string that hadn't changed.

app/views/users/terms.html.erb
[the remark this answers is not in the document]
tomhughes: Well `app/assets/javascripts/user.js` looks like an obvious choice given it is already loaded in that page.

mmd-osm:
I'm wondering how this new flag affects existing users. Will editing be blocked unless they have ticked some checkbox?

simonpoole:
Sorry, I wanted to explain this in the comment but in the usual hackweekend environment I forgot.

The current changes do not effect existing accounts at all with the exception of those that haven't accepted the contributor terms.
Making the Tou acceptance mandatory for other existing accounts is way down the road, if ever, and will require additional coding not to mention discussion. This is in line with the LWG plans.

simonpoole:
Feedback from @systemed 

- doesn't like the new position of the PD checkbox (surprise :-))
- suggests not having the ToU link in the Welcome box and instead moving it in to the attribution "footer"

config/locales/en.yml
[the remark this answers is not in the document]
simonpoole: I've renamed the 'read and accept' string, the rest would seem to be harmless (aka the title will be different till the translations have been fixed, but that is no big issue),.

tomhughes:
Well https://github.com/openstreetmap/openstreetmap-website/blob/master/test/integration/user_creation_test.rb#L75 tests the signup flow including accepting the terms?

If you actually want to check what happens to the button on the client then you will need a system test though obviously you want to check that the backend works right as well as you don't want to trust client side javascript ;-)

mmd-osm:
> The current changes do not effect existing accounts at all with the exception of those that haven't accepted the contributor terms.
Making the Tou acceptance mandatory for other existing accounts is way down the road, if ever, and will require additional coding not to mention discussion. This is in line with the LWG plans.

@simonpoole : somehow this doesn't fit well with what is stated here: https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services:

> rare case: For a transition period, there will be old user accounts that have not yet accepted the ToU. These would not be given access to "sensitive" material. 

In the context of GDPR there needs to be some mechanism to accept the ToU even for existing users, otherwise you won't see any of the to-be hidden fields anymore even as logged on user.

simonpoole:
@mmd-osm I was saying that in the context of making acceptance mandatory with this PR. As you point out we need to either:

- make acceptance mandatory latest before the API restrictions go in to force (when the code is available and so on, I'm not holding my breath on that),

- or provide a mechanism with which people could accept when necessary.

The later doesn't seem to make much sense as there are other issue than just GDPR compliance that are addressed in the ToU and at least -in principle- we would want everybody to agree to them. I'm just being a bit pragmatic about the whole thing as there are going to be a lot more signups going forward than there are now, and we are not in a hurry to force acceptance as long as there is no sign of the GDPR related code being worked on.

app/models/user.rb
```diff
@@ -34,6 +34,7 @@
 #  image_content_type  :string
 #  auth_provider       :string
 #  home_tile           :integer
+#  tou_agreed          :datetime
```
pnorman: What does this map in PostgreSQL? We should be using `timestamp with time zone` for a purpose like this.

tomhughes: All out servers run on UTC anyway so it won't make any real difference.

simonpoole: As @tomhughes says and we use UTC for timestamps in UTC in other places of the code most particularly for CT acceptance so I would find it a bit confusing to treat this differently.

gravitystorm: We should stick with `timestamp (without time zone)` for consistency in this PR. But as @pnorman indicates, we should migrate all timestamp fields to `timestamp with time zone` as a separate task.

mmd-osm: @gravitystorm : please keep in mind that changing to "timestamp with time zone" possibly also has some impact on cgimap, it's not a rails only thing.

tomhughes: I don't think it should matter given that we always use UTC anyway. I equally don't see why we would want to change either but that's not really a matter for this PR.

simonpoole: I note that the ongoing discussion currently makes it impossible to finalize this PR, it would be really nice if this could be done in this year.

simonpoole:
> 
> 
> Well https://github.com/openstreetmap/openstreetmap-website/blob/master/test/integration/user_creation_test.rb#L75 tests the signup flow including accepting the terms?
> 
> If you actually want to check what happens to the button on the client then you will need a system test though obviously you want to check that the backend works right as well as you don't want to trust client side javascript ;-)

I'll punt on the client side javascript :-) (assuming that we don't actually have any kind of test harness that directly simulates user interaction), But since the requirement that the checkbox is checked is the only functional aspect that could be tested separately I'm a bit at a loss to find any sensible additional testable condition, so IMHO I would let things stand as they are now (when and if we add code to force ToU acceptance for existing users that can be revisited).

app/assets/javascripts/user.js
```diff
@@ -124,4 +124,9 @@ $(document).ready(function () {
     $("#contributorTerms").html("<img src='" + OSM.SEARCHING + "' />");
     $("#contributorTerms").load(url);
   });
+
+  $("#read_tou").on("click", function () { 
+    var agreebtn = document.getElementById('agree');
+    agreebtn.disabled=!this.checked;
```
tomhughes: I'd rather do this with jQuery like everything else - something like this I think:

```
$("#agree").prop("disabled", !$(this).prop("checked"))
```

app/views/users/terms.html.erb
```diff
@@ -28,27 +30,36 @@
...
+      <%= t '.read_tou_html', :tou_link => link_to(t('layouts.tou'), "https://wiki.osmfoundation.org/wiki/Terms_of_Use", :target => :new) %>
     </label>
-    <span class="minorNote">(<%= link_to(t('.consider_pd_why'), t('.consider_pd_why_url'), :target => :new)%>)</span>
 
     <%= hidden_field_tag('referer', h(params[:referer])) unless params[:referer].nil? %>
 
     <div class="buttons form-row inner20 clearfix">
-      <p class="deemphasize"><%= t '.read and accept' %></p>
-      <%= submit_tag(t('.agree'), :name => "agree", :id => "agree") %>
+      <%= submit_tag(t('.agree'), :name => "agree", :id => "agree", disabled: true)  %>
```
tomhughes: Please use the `=>` syntax for hashes.

app/views/users/terms.html.erb
```diff
@@ -28,27 +30,36 @@
...
+  <label class="standard-label">
+    <%= t 'layouts.tou' %>
+  </label>
+  <p class="deemphasize"><%= t '.tou_explain' %></p>
   <div class="form-row">
-    <label for="user_consider_pd">
-      <%= check_box('user', 'consider_pd') %>
-      <%= t '.consider_pd' %>
+    <label for="read_tou">
+      <%= check_box_tag 'read_tou',1,false %>
```
tomhughes: There's no need for the last two arguments - those are the default values.

app/views/users/terms.html.erb
simonpoole: Fixed in https://github.com/openstreetmap/openstreetmap-website/pull/2028/commits/5dfbc92ef60112c55462bab75898c25d31d6bc84

app/views/users/terms.html.erb
simonpoole: Fixed in https://github.com/openstreetmap/openstreetmap-website/pull/2028/commits/5dfbc92ef60112c55462bab75898c25d31d6bc84

simonpoole:
> 
> 
> In addition to the inline comments, don't we need something on the server side to check that they actually agreed?
> 
> It looks to me like currently it relies on the submit button being disabled until javascript enables it when the checkbox is ticked but that leaves it open to somebody to hack a post of the form with `read_tou` unset and then claim they have never agreed...

I now protect against this, not really sure if "users_controler.save" is the best place, and redirect back to the terms page. Further I've added a test that tests for the redirct if the check box parameter is missing/not checked (not that I actually understand how it works :-)).

app/controllers/users_controller.rb
```diff
@@ -41,7 +41,7 @@ def terms
   def save
     @title = t "users.new.title"
 
-    if params[:decline]
+    if params[:decline] || !params[:read_tou] || params[:read_tou] == "0"
```
tomhughes: You don't actually need both tests here as "truthiness" means that `"0"` is false and hence the first test will be true...

simonpoole:
@tomhughes ping this can't proceed before there's feedback on the mock up.

tomhughes:
That's certainly much better. Just a couple of comments...
* I think I would say "...above contributor terms" in the prompt with the first check box to make it clear it's referring to the text above.
* The ToU section is a bit weird still because it looks like they're missing but actually you're expected to follow the link that is part of the check box text. I think I would prefer to move the link up into the body text somehow (not sure of the exact wording) and then word the question more like the first one referring to "...above terms of use" if that makes sense?

pnorman:
I'd rather get rid of the PD checkbox since we've determined it's legally meaningless than use up space on the redesign.

simonpoole:
https://github.com/openstreetmap/openstreetmap-website/pull/2028/files/1657dad61c9220a638638d6eeb75ab2a239f6672..0e0078cef30556f056ae5e1c9ad1556e76b645e7 adds the CT check box for real and makes some text changes along the lines of what @tomhughes suggested.

simonpoole:
@tomhughes should I the current state available on a local test instance as I did in October?

simonpoole:
I can rebase this to get rid of the new conflict, but it only makes sense to do so if there is a chance of it getting merged.

tomhughes:
What are we actually waiting on? I didn't really understand your previous comment?

simonpoole:
> 
> 
> What are we actually waiting on? I didn't really understand your previous comment?

Well in October I was running a test instance for public review, see https://github.com/openstreetmap/openstreetmap-website/pull/2028#issuecomment-431669165 I was just offering to do it again (with the current layout), if it could move this forward. 

In any case from my pov I don't see anything further that needs doing except if we want to in some way display the ToU text, but IMHO there simply isn't enough screen real estate to do that in any meaningful way.

tomhughes:
You could do, or I could put up a dev instance.

config/locales/en.yml
```diff
@@ -1089,6 +1089,8 @@ en:
     partners_ucl: "UCL"
     partners_bytemark: "Bytemark Hosting"
     partners_partners: "partners"
+    intro_tou_html: "Use of this site is subject to our %{tou_link}."
```
tomhughes: Is this actually used anywhere?

simonpoole: This was the text in the welcome box in the 1st version that @systemed didn't want, and that I then removed. Fixed now.

simonpoole:
Thanks (in the name of the LWG)!

PR #2029, opened 2018-10-20
Don't duplicate note comments in RSS feeds for new notes

amandasaurus:
Currently, in the RSS feed of note activity, when a note is created, the note is included twice, because it includes the comment on the note, and then the whole note. With only one comment, the latter is redudant.

e.g.:
> new note (near Dolphin's Barn, Ushers E ED, Dublin 8, Dublin, County Dublin, Leinster, D08 WN40, Ireland)
> 20 October 2018, 17:13
> Comment
> Created 10 minutes ago by ManAboutCouch
> closed
> Full note
> Created 10 minutes ago by ManAboutCouch
> closed

This is my attempt to fix that. I know very little about Ruby, or Rails, so please excuse any beginner mistakes. :) Feel free to rip this apart and do it properly, if needed. This is my first patch to the OSM-website. :)

I think this should also apply to the email content as well, right?

### DIFF
--- a/app/views/notes/_entry.html.erb
+++ b/app/views/notes/_entry.html.erb
@@ -1,4 +1,6 @@
 <h2><%= t ".comment" %></h2>
 <%= render :partial => "comment", :object => entry %>
+<% if entry.note.comments.count > 1 %>
 <h2><%= t ".full" %></h2>
 <%= render :partial => "description", :object => entry.note %>
+<% end %>
--- a/test/controllers/notes_controller_test.rb
+++ b/test/controllers/notes_controller_test.rb
@@ -1066,13 +1066,13 @@ class NotesControllerTest < ActionController::TestCase
     assert_equal "application/rss+xml", @response.content_type
     assert_select "rss", :count => 1 do
       assert_select "channel", :count => 1 do
-        assert_select "item", :count => 2;
+        assert_select "item", :count => 2
 
         # There is only 1 note, which has no comments, so it shouldn't include the "full" section
         assert_select "item" do
-            assert_select "description", :text => /This is note comment 2/
-            assert_select "description", :count => 0, :text => /Full note/
-            assert_select "description", :count=>0, :text => /This is note comment 2.*This is note comment 2.*/m
+          assert_select "description", :text => /This is note comment 2/
+          assert_select "description", :count => 0, :text => /Full note/
+          assert_select "description", :count => 0, :text => /This is note comment 2.*This is note comment 2.*/m
         end
       end
     end
@@ -1085,12 +1085,12 @@ class NotesControllerTest < ActionController::TestCase
     assert_equal "application/rss+xml", @response.content_type
     assert_select "rss", :count => 1 do
       assert_select "channel", :count => 1 do
-        assert_select "item", :count => 2;
+        assert_select "item", :count => 2
         assert_select "item" do
-            assert_select "description", :text => /Full note/
-            assert_select "description", :text => /This is note comment 6/
-            assert_select "description", :text => /This is note comment 5/
-            assert_select "description", :text => /Comment.*This is note comment 6.*Full note.*This is note comment 5.*This is note comment 6.*/m
+          assert_select "description", :text => /Full note/
+          assert_select "description", :text => /This is note comment 6/
+          assert_select "description", :text => /This is note comment 5/
+          assert_select "description", :text => /Comment.*This is note comment 6.*Full note.*This is note comment 5.*This is note comment 6.*/m
         end
       end
     end
--- a/test/controllers/notes_controller_test.rb
+++ b/test/controllers/notes_controller_test.rb
@@ -1067,13 +1067,30 @@ class NotesControllerTest < ActionController::TestCase
     assert_select "rss", :count => 1 do
       assert_select "channel", :count => 1 do
         assert_select "item", :count => 2;
+
+        # There is only 1 note, which has no comments, so it shouldn't include the "full" section
         assert_select "item" do
-            assert_select "description", {:text => %r{<h2>Comment</h2>
-<div class="note-comment" style="margin-top: 5px">
-  <div class="note-comment-description" style="font-size: smaller; color: #999999">Created <span title=".*">less than a minute</span> ago</div>
-  <div class="note-comment-text">This is note comment 2</div>
-</div>
-</div>}}
+            assert_select "description", :text => /This is note comment 2/
+            assert_select "description", :count => 0, :text => /Full note/
+            assert_select "description", :count=>0, :text => /This is note comment 2.*This is note comment 2.*/m
+        end
+      end
+    end
+
+    # Create a note with 2 comments and ensure the full part is in the RSS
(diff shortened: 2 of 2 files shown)


tomhughes:
Granted this is redundant but you could equally argue that it's always redundant to include the recent comment because it's in the full note.

As it stands this would for example break any downstream user of the feed that chose just to ignore the comment and always show the full version of the note so it seems to be to be quite a risky change to make?

rory:
AFAIK this doesn't change the notes API? If you were doing something with notes, surely you'd use that rather than parsing the HTML in this? (if you are parsing the HTML then you're doing it wrong and shouldn't complain if this changes (IMO)) 

On 21 October 2018 20:44:19 CEST, Tom Hughes <notifications@github.com> wrote:
>Granted this is redundant but you could equally argue that it's always
>redundant to include the recent comment because it's in the full note.
>
>As it stands this would for example break any downstream user of the
>feed that chose just to ignore the comment and always show the full
>version of the note so it seems to be to be quite a risky change to
>make?

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

rory:
Yes my goal is to change the HTML of how a note is displayed inside the feed of notes.

On 21 October 2018 22:26:27 CEST, Tom Hughes <notifications@github.com> wrote:
>Is that in HTML inside the feed then? Sorry, but it was pretty hard to
>figure out what exactly you were changing and what the effect would
>be...

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tomhughes:
I confess that I don't really understand the notes feed, and I'm not sure who uses it and what for so I'm finding it hard to evaluate this.

The core problem is that it's a feed of notes but it is triggered by comments so that notes reappear when they have a new comment right? But it has this weird dual thing of showing both the latest comment (that caused it to be included) and the whole thread?

But it seems to me that you change doesn't really fix that because we still get the duplication the second time round - why is it a problem for a new note and not for subsequent comments?

rory:
> I'm not sure who uses it and what for

I do! 🙂 I have it in my RSS reader so I can monitor the notes in an area.

> so that notes reappear when they have a new comment right?

Correct.

> But it has this weird dual thing of showing both the latest comment (that caused it to be included) and the whole thread?

Yes. When there's a "comment" on a note (including the opening of the note), there's an entry in the feed, which includes that comment, and then all the comments in/on that note. So when a note is opened, the first comment is displayed twice.

> you change doesn't really fix that because we still get the duplication the second time round - why is it a problem for a new note and not for subsequent comments?

My ruby skills are underdeveloped 😉, so I didn't try too much. This patch is an improvement, while not making things perfect.

Yes when there's the (e.g.) 5th comment on a note, that comment will appear twice (at the top and as part of the 'whole note'). IMO this isn't so bad. Including the whole note (w/ all comments in chronological order) in each feed entry is good. And including, at the top, the comment that caused this feed is good too. So I think removing the 5th comment from the 'whole note' section is bad, and removing it from the top is bad too.

rory:
> I'm not sure who uses it and what for

I do! 🙂 I have it in my RSS reader so I can monitor the notes in an area.

> so that notes reappear when they have a new comment right?

Correct.

> But it has this weird dual thing of showing both the latest comment (that caused it to be included) and the whole thread?

Yes. When there's a "comment" on a note (including the opening of the note), there's an entry in the feed, which includes that comment, and then all the comments in/on that note. So when a note is opened, the first comment is displayed twice.

> you change doesn't really fix that because we still get the duplication the second time round - why is it a problem for a new note and not for subsequent comments?

My ruby skills are underdeveloped 😉, so I didn't try too much. This patch is an improvement, while not making things perfect.

Yes when there's the (e.g.) 5th comment on a note, that comment will appear twice (at the top and as part of the 'whole note'). IMO this isn't so bad. Including the whole note (w/ all comments in chronological order) in each feed entry is good. And including, at the top, the comment that caused this feed is good too. So I think removing the 5th comment from the 'whole note' section is bad, and removing it from the top is bad too.

rory:
> I'm not sure who uses it and what for

I do! 🙂 I have it in my RSS reader so I can monitor the notes in an area.

> so that notes reappear when they have a new comment right?

Correct.

> But it has this weird dual thing of showing both the latest comment (that caused it to be included) and the whole thread?

Yes. When there's a "comment" on a note (including the opening of the note), there's an entry in the feed, which includes that comment, and then all the comments in/on that note. So when a note is opened, the first comment is displayed twice.

> you change doesn't really fix that because we still get the duplication the second time round - why is it a problem for a new note and not for subsequent comments?

My ruby skills are underdeveloped 😉, so I didn't try too much. This patch is an improvement, while not making things perfect.

Yes when there's the (e.g.) 5th comment on a note, that comment will appear twice (at the top and as part of the 'whole note'). IMO this isn't so bad. Including the whole note (w/ all comments in chronological order) in each feed entry is good. And including, at the top, the comment that caused this feed is good too. So I think removing the 5th comment from the 'whole note' section is bad, and removing it from the top is bad too.

rory:
> I'm not sure who uses it and what for

I do! 🙂 I have it in my RSS reader so I can monitor the notes in an area.

> so that notes reappear when they have a new comment right?

Correct.

> But it has this weird dual thing of showing both the latest comment (that caused it to be included) and the whole thread?

Yes. When there's a "comment" on a note (including the opening of the note), there's an entry in the feed, which includes that comment, and then all the comments in/on that note. So when a note is opened, the first comment is displayed twice.

> you change doesn't really fix that because we still get the duplication the second time round - why is it a problem for a new note and not for subsequent comments?

My ruby skills are underdeveloped 😉, so I didn't try too much. This patch is an improvement, while not making things perfect.

Yes when there's the (e.g.) 5th comment on a note, that comment will appear twice (at the top and as part of the 'whole note'). IMO this isn't so bad. Including the whole note (w/ all comments in chronological order) in each feed entry is good. And including, at the top, the comment that caused this feed is good too. So I think removing the 5th comment from the 'whole note' section is bad, and removing it from the top is bad too.

rory:
> I'm not sure who uses it and what for

I do! 🙂 I have it in my RSS reader so I can monitor the notes in an area.

 > so that notes reappear when they have a new comment right?

Correct.

 > But it has this weird dual thing of showing both the latest comment 
(that caused it to be included) and the whole thread?

Yes. When there's a "comment" on a note (including the opening of the 
note), there's an entry in the feed, which includes that comment, and 
then all the comments in/on that note. So when a note is opened, the 
first comment is displayed twice.

 > you change doesn't really fix that because we still get the 
duplication the second time round - why is it a problem for a new note 
and not for subsequent comments?

My ruby skills are underdeveloped 😉, so I didn't try too much. This 
patch is an improvement, while not making things perfect.

Yes when there's the (e.g.) 5th comment on a note, that comment will 
appear twice (at the top and as part of the 'whole note'). IMO this 
isn't so bad. Including the whole note (w/ all comments in chronological 
order) in each feed entry is good. And including, at the top, the 
comment that caused this feed is good too. So I think removing the 5th 
comment from the 'whole note' section is bad, and removing it from the 
top is bad too.

Issue #2033, opened 2018-10-24
Redaction title can be blank.


RoyallDesigns:
@1v0ryh4t and I found that when creating a new redaction under `/redactions/new`, the redaction title is allowed to be blank. When attempting to view said redactions under `/redactions`, the new redaction is not present or viewable under the list as its title or name is an empty string. I would like to add a validation to a `redaction` so that when creating or editing the `redaction`, the name can never be blank. (This seems like a decent `good-first-issue` for myself.)

Please see the attached images.

![notitleredactioncreation](https://user-images.githubusercontent.com/41215187/47397923-97684e00-d6ff-11e8-8e0c-b1267067f9a8.png)
![redactionsscreen](https://user-images.githubusercontent.com/41215187/47397929-9b946b80-d6ff-11e8-9e90-46f08170ace2.png)

All redactions (`/redactions`):

![redactions_list](https://user-images.githubusercontent.com/41215187/47398054-3a20cc80-d700-11e8-97ca-7b6729bb9671.png)

gravitystorm:
I think it's reasonable to require a title, should we also consider making the description required too?

I checked the [original pull request](https://github.com/openstreetmap/openstreetmap-website/pull/39) for adding redactions, and I couldn't find any discussion as to whether titles and/or descriptions should be optional. I'd be interested in feedback from any openstreetmap.org moderators, who have the most experience with creating redactions. Personally I think it would be fine to require both attributes to be present.

tomhughes:
I think it's clear a title should be required and I'd be inclined to say a description should be as well.

I'm surprised I didn't insist on it at the time - it must have slipped by me when I was reviewing this.

rileythomp:
Does this still need to be worked on?

SomeoneElseOSM:
> I'd be interested in feedback from any openstreetmap.org moderators

Creating new redactions isn't something that we need to do that often, thankfully -the last one I did was a couple of years ago I think - https://www.openstreetmap.org/redactions/102 .  I'd always expect to use a title and description.

sonukushwaha403:
is this issue open can i work on it

gravitystorm:
> is this issue open can i work on it

Yes, the issue is open and can be worked on. Please also read through the [previous pull request](https://github.com/openstreetmap/openstreetmap-website/pull/2034) for more information.

BLSowmya:
is this issue open  and can i work on it ???

gravitystorm:
[...]

PR #2034, opened 2018-10-24
Added a validation to the Redaction model for the title.

RoyallDesigns:
Now, a `Redaction`'s `title` must be present for it to be stored in the database.
The next question is if a test should be written to verify this in an automatic fashion. It does not appear that any of the unit tests for the `Redaction` test for the responsibility of a `Redaction` to not store invalid entries in the database. Where do we draw the line at responsibility of the `rails` developer / the `Redaction` model and `rails` itself?


tomhughes:
If you're going to open a PR there's no need to open a bug first only to then open a PR that fixes it - that just causes the information to be fragmented over two tickets when one would do.

And if you do open a bug first then please at least mark the PR as closing the bug in your commit message so that we don't have to close it manually...

gravitystorm:
@RoyallDesigns Thanks for your PR. If we decide to make the redaction title required, then there are three things that should be changed:

* The database needs to have a "not null" constraint added on the `title` column. This will require a [migration](https://edgeguides.rubyonrails.org/active_record_migrations.html) to be added. The migration might also need to check for existing records with blank titles, and set a default title, before adding the constraint.
* The model needs to have a validation, which is what you've got here
* The model test (in test/models/redaction_test.rb) should have a new test, asserting that the redaction is not valid when the title is empty. This ensures that any refactoring mistakes that happen to the model in future get caught in the test suite.

gravitystorm:
> If you're going to open a PR there's no need to open a bug first only to then open a PR that fixes it - that just causes the information to be fragmented over two tickets when one would do.
> 
> And if you do open a bug first then please at least mark the PR as closing the bug in your commit message so that we don't have to close it manually...

I don't think these chastisements are necessary, nor particularly welcoming for a first time contributor.

tomhughes:
It was intended as advice rather than a chastisement, so apologies if it came over wrongly.

RoyallDesigns:
This addresses issue #2033.

RoyallDesigns:
Resolves issue #2033.

db/migrate/058_name_unnamed_redactions.rb
```diff
@@ -0,0 +1,19 @@
+class NameUnnamedRedactions < ActiveRecord::Migration[5.0]
+  def self.up
+    Redaction.find_each do |redaction|
+        if redaction.title.empty?
```
tomhughes: Indentation should be the standard two spaces please.

db/migrate/058_name_unnamed_redactions.rb
```diff
@@ -0,0 +1,19 @@
...
+            redaction.title = redaction.id.to_s
+            redaction.save!
+        end
+    end
```
tomhughes: After you fixed any blank titles you need to use `change_column` to add a "not null" constraint to it.

db/migrate/058_name_unnamed_redactions.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def self.down
+    Redaction.find_each do |redaction|
```
tomhughes: There's no need to change the titles back here, but you do need to remove the constraint with `change_column` here.

RoyallDesigns:
Is there an easy way to contribute to this branch on my end, privately, to make changes and then synchronize said changes with this pull request? This way, as I am working on changes, I am not making anyone think that the commits are "merge-worthy".

RoyallDesigns:
I should have been a bit more clear. I meant to say, from your experience, it is possible to make changes with regard to one's own branch on one's own GitHub repo (say: https://github.com/RoyallDesigns/openstreetmap-website/tree/redaction-title-issue-2033) without me automatically requesting to make changes to this Pull Request? I want to be able to push small changes for others I am working with to see and use (off of my repo) without necessarily attempting to request a review of them.

tomhughes:
You would have to push them to a different branch - while this PR is open it will pick up any changes to the branch you used to open it automatically.

That doesn't really matter though - just tell us when you thnik you're done and we can ignore changes until then.

RoyallDesigns:
@tomhughes Thank you for the assistance.

johnkirby2:
Hello @tomhughes. I'm trying to help @RoyallDesigns with documentation, and I was wondering if you had suggestions about comment styling for code specific to openstreetmap? I saw that you mentioned indentation edits to @RoyallDesigns for 058_name_unnamed_redactinos.rb. I'm a bit of a beginner with this, so I appreciate your patience and you reading this. Thanks!

tomhughes:
Not sure what you mean by comment styling - in general we only add comments when there is something to say that is not obvious from the code.

More generally, make sure that `rubocop` passes - that would have caught the indentation.

db/migrate/058_name_unnamed_redactions.rb
```diff
@@ -0,0 +1,19 @@
+class NameUnnamedRedactions < ActiveRecord::Migration[5.0]
+  def self.up
+  	# after titles are fixed we change the column so to not be null
+    change_column_null "redactions", "title", false
```
gravitystorm: I think you might need to fill in the titles first, and then change the column constraint afterwards.

db/migrate/058_name_unnamed_redactions.rb
```diff
@@ -0,0 +1,19 @@
+class NameUnnamedRedactions < ActiveRecord::Migration[5.0]
```
gravitystorm: The migration file should use a timestamp in the filename, see recent migrations that start with db/migrate/2018[...]

db/migrate/058_name_unnamed_redactions.rb
```diff
@@ -0,0 +1,19 @@
...
+  def self.up
+  	# after titles are fixed we change the column so to not be null
+    change_column_null "redactions", "title", false
+    # gets each redaction
+    # sets redaction title equal to a string of the id if there is no title
+    Redaction.find_each do |redaction|
+      if redaction.title.empty?
+        redaction.title = redaction.id.to_s
```
gravitystorm: There's no need to use `.to_s` here, rails will take care of the conversion for you.

johnkirby2:
Hello. I've re-pushed the 058_name_unnamed_redactions.rb file with some changes added. Thank you @tomhughes and @gravitystorm for your recommendations. Those of us who are working on the redaction-title-issue-2033 branch have been experiencing an issue with openstreetmap where every time we attempt to use vagrant within our own version of the project and run the rails server we get this error message in the browser:

syntaxerrorscreenshot

The way we've been able to find a workaround for this involves copying and pasting the needed files over to the right directory, and then re-running rails server. Using these commands in our local version of the repo:

`cd app/assets/stylesheets/ltr`
`cp -p ../common.scss ../small.scss .`
`cd ../rtl`
`cp -p ../common.scss common.r2.scss`
`cp -p ../small.scss small.r2.scss`
`cd ../../../../`
`vagrant ssh`
`cd /srv/openstreetmap-website`
`rails server`

Is there possibly a more long-term remedy for this?

RoyallDesigns:
@gravitystorm and @tomhughes, do you recommend any tests for the migration that we have created for unnamed `redactions`? I do not see any tests for migrations in the current setup, so I wonder if they are necessary?

RoyallDesigns:
If a test is not required for a migration, then I believe we are ready to merge in when you approve, @gravitystorm and @tomhughes. Thank you for all of the assistance.

gravitystorm:
First of all, I'm sorry that it has taken so long to respond here.

I've reviewed this PR a couple of times over the last few months, and mulled over the options, and there's a number of issues with it, including:

* Tests are added for both `title` and `description`, but the migration only covers the `title`.
* The model tests can be simplified by using e.g. `redaction.valid?` instead of attempting to save and then asserting against the errors.
* A lot of comments are added to existing tests, but they mostly just restate exactly what the code already says.
* Due to changes in the meantime, this PR now conflicts with master.

But the big thing is that we've introduced the 'strong_migrations' gem, which flags up this migration as potentially locking the database while it runs. So the migration will have to be substantially re-written to deal with this.

My recommendation for reworking this PR is to split it into two parts - one for the additional model tests (which only test the model validations anyway, not the database constraints) and a separate one for the migration, rewritten to be compatible with `strong_migrations`.

PR #2038, opened 2018-10-31
Use deliver_later for email sending

gravitystorm:
This PR builds on #2037 and configures email sending as a background task.

Note that this depends on having the setup for delayed_job (#2037) already merged, along with having workers running in production (https://github.com/openstreetmap/chef/issues/201), before this can be deployed - so I'm marking it as WIP until those prerequisites are met.


gravitystorm:
Both the prerequisites have been finished, so this is ready to go.

PR #2042, opened 2018-11-03
Limit note comment size to 2000 characters

mmd-osm:
Follow up for #1543 

There's essentially no upper limit to notes comments, leaving the upload more or less unprotected to nonsense stuff like https://upload.apis.dev.openstreetmap.org/note/3

PR attempts to set the limit to 2000 characters both in the model, as well in the textarea (to avoid dealing with an unhappy backend). I'm not sure if anything else needs to be done to deal with existing (too long) comments.


HolgerJeromin:
We should check if every existing note comments are inside the new limit

tomhughes:
It doesn't really matter - the limit will only be enforced if the comment is changed for any reason and about the only way to do that is to hide it so that's about the only thing which might break.

PR #2044, opened 2018-11-04
Specifiy invalid chars in username error message

### LINKED ISSUE #157: Error: Display Name is invalid
The error "Display Name is invalid" while registering might not be too helpful. Instead, display which characters are not allowed (I'm not completely sure about the handling of `\/`):

``` diff
diff --git a/app/models/user.rb b/app/models/user.rb
index 3b43130..27662f1 100644
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -36,7 +36,7 @@ class User < ActiveRecord::Base
   validates_length_of :display_name, :within => 3..255, :allow_nil => true
   validates_email_format_of :email, :if => Proc.new { |u| u.email_changed? }
   validates_email_format_of :new_email, :allow_blank => true, :if => Proc.new { |u| u.new_email_changed? }
-  validates_format_of :display_name, :with => /^[^\/;.,?%#]*$/, :if => Proc.new { |u| u.display_name_changed? }
+  validates_format_of :display_name, :with => /^[^\/;.,?%#]*$/, :message => "contains invalid characters (\/;.,?%#)" :if => Proc.new { |u| u.display_name_changed? }
   validates_format_of :display_name, :with => /^\S/, :message => "has leading whitespace", :if => Proc.new { |u| u.display_name_changed? }
   validates_format_of :display_name, :with => /\S$/, :message => "has trailing whitespace", :if => Proc.new { |u| u.display_name_changed? }
   validates_numericality_of :home_lat, :allow_nil => true
```

jguthrie100:
Fixes #157

I've extracted out the regex list of invalid characters to a constant (to avoid errors in duplication) and added a locale translation. This should make the error message a lot more helpful!


tomhughes:
We should probably go further an extract it to a shared concern or something as there are other models which use the same validation the are various ways (https://guides.rubyonrails.org/active_record_validations.html#common-validation-options) of creating a custom validation that can be shared.

jguthrie100:
If this works as intended, then if you merge this, I can create a new PR with a custom validator (and with changes to a few models) based on the updated `master` branch which might work out cleaner.

tomhughes:
They're not ASCII chars anyway, they're unicode code points ;-)

Note that the first group is the ones which have special meaning in URLs and are hence banned in fields which might become part of a URL while the second group are (or should be) just banned everywhere.

So I would tend to make the names more like `INVALID_URL_CHARACTERS` and `INVALID_CHARACTERS` rather than what you have now (technically characters is wrong but it will make more sense to a casual reader).

tomhughes:
By the way, feel free to rebase on master if you want, or to rebase the set of patches to cleanup the history rather than just adding more commits - there's absolutely no need to preserve the history on a branch that isn't merged yet if rewriting it would give a cleaner and more meaningful set of commits. You'll have to force push to update the branch but that's fine for an unmerged branch.

jguthrie100:
To be honest, I'm not entirely sure how to go about rebasing etc.

Instead I've just made a commit with some new validator methods and I've applied the validations to every input text file d I could find in the models.

Is there any chance you could reorganise everything if the commits/branches need reorganising, because I'm completely out of my depth git-wise!

tomhughes:
I've replaced your merge with a rebase now and also taken out some other unrelated changes to the model comments.

To rebase on master you can just do `git rebase master` rather than `git merge master` which will replay your commits on the current tip of master, stopping if any of the commits won't apply cleanly and needs fixing up, much as merge does.

tomhughes:
Argh. You've bought back all the stuff I cleaned up last night!

tomhughes:
Why have you merged this branch (or at least a version of it) to itself anyway? Did you not pull my rebased version before you triyed to change it?

tomhughes:
I've added your two new commits to my version of the branch and repushed to fix things again - if you want to make more changes please make sure you pull the new version first.

jguthrie100:
Apologies! I'm not sure at all whats happened and why I didn't have the latest changes. My rebasing skills definitely need some work!

Unfortunately I've only just seen your message, but have already amended and forced some changes through (to undo the bigint(8) changes etc), and I may have messed up your branch again.

If its still ok though, then everything that needed to be implemented has been implemented, so its good to be merged and deleted when you're happy with the changes.

Apologies again!

PR #2049, opened 2018-11-07
Update leaflet.locate.js

domoritz:


tomhughes:
Update it with what? To do what? When opening a PR please try and explain the purpose of the change....

In any case as a third party vendored asset this should be updated upstream, not here.

tomhughes:
Ah I see this is an update from upstream...

tomhughes:
There's no real need to push these at us by the way - it's in our vendor file so will get pulled whenever I run a update of vendored resources.

domoritz:
Sweet, I didn't know that. Thanks for the comments.

tomhughes:
It turns out the version is actually hardcoded in `Vendorfile` so needs to be updated to get a new release. I made that change to your patch this morning before I merged it.

What I typically do is to check for new versions of any plugins when I'm updating leaflet itself and update `Vendorfile` as appropriate but always happy to take new versions in between.

PR #2051, opened 2018-11-07
CanCanCan for user blocks and redactions

gravitystorm:
This PR refactors user blocks and redactions to use CanCanCan. 

This does change the response slightly for forbidden actions, rather than redirecting to :index with a flash message. But I'm happy with the simplification and I doubt anyone will notice the difference.


tomhughes:
I added some extra code in ccdec3ed4c0f7c472f6db7e5eff2dfc1579f3dd1 to try return nicer errors when the client is a browser... In practice I don't think that there is any path (other than deliberate mucking about) that should get you to those errors because you should only see the links if you are logged in as a moderator.

Much the same goes for the custom `deny_access` for diary entries, issues and issue comments in that I don't think they can generally happen so we should probably just let them hit the default.

gravitystorm:
Thanks @tomhughes , nice solution there. I'll look at removing the custom deny_access methods, since I don't think the extra code complexity is providing any real-world benefit.

PR #2056, opened 2018-11-12
add links to Welcome Mat on /welcome and /help

mikelmaron:
Work on the [Welcome Mat](https://welcome.openstreetmap.org/) has stabilized and it's ready to link from other sites in OSM. This PR adds links on /welcome and /help


tomhughes:
Having two "welcome" links seems like it will be confusing?

mikelmaron:
Yes missed that. Will think of a different label and update

PR #2060, opened 2018-11-14
Remove custom deny_access handlers

gravitystorm:
Since these pages are not accessed by normal users, except for url fiddling, it's fine to respond with a generic access denied.


tomhughes:
There seems to be a test failure here:

```
Failure:
IssuesTest#test_view_issues_normal_user [/home/tom/src/osm/rails/test/system/issues_test.rb:15]:
Expected false to be truthy.
```

gravitystorm:
Oops! Fixed and PR updated.

PR #2062, opened 2018-11-14
Return 401 Unauthorized when user is not logged in

gravitystorm:
This PR changes the `require_user` and `deny_access` methods to return 401 Unauthorized to web requests when the user hasn't yet logged in, and updates the affected tests. This brings us better in line with the HTTP specs.

I came across this while doing some more work on CanCanCan refactoring. The `authorize` method, used by all the API actions, has always returned 401 in this situation, so this PR makes the behaviour the same for non-API actions and simplifies some other work-in-progress that I have.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -490,10 +490,10 @@ class ApplicationController < ActionController::Base
     elsif request.get?
       respond_to do |format|
         format.html { redirect_to :controller => "users", :action => "login", :referer => request.fullpath }
-        format.any { head :forbidden }
+        format.any { head :unauthorized }
       end
     else
-      head :forbidden
+      head :unauthorized
     end
   end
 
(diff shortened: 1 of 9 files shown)


tomhughes:
I wondered about that, but I'm not at all sure - it encourages the user of basic auth which we really want to deprecate and we should in any case only be doing it to API users and not to browsers which is hard to determine in many cases.

gravitystorm:
Oh, I see. I didn't realise that 401 responses also require a "WWW-Authenticate" header, and that there's (apparently) no way to have that header explain that you should go to the login page to get yourself a cookie before retrying the (POST|PUT|DELETE) request.

I certainly don't want to add more Basic Auth stuff, so I guess for these non-GET requests we should keep returning 403 Forbidden since we can just return that without any consequences.

I though that this change would be helpful to some (edge case) users. I thought that a browser would default to a slightly different (and more helpful) message to the user for 401 vs 403, but it just opens cans of worms so I'll withdraw this PR.

PR #2063, opened 2018-11-14
Fossgis OSRM server support

simonpoole:
This fixes https://github.com/openstreetmap/operations/issues/237 by adding support for the FOSGIS OSRM server and removing the previous OSRM based routing service.


tomhughes:
This goes way beyond what that ticket is requesting - that ticket is about adding a new routing engine and does propose removing one.

As I said on the operations ticket, the code changes are not the holdup here. The holdup is making the policy decision.

mmd-osm:
I believe a CSP entry for the new endpoint is missing here: https://github.com/openstreetmap/openstreetmap-website/blob/ccdec3ed4c0f7c472f6db7e5eff2dfc1579f3dd1/app/controllers/application_controller.rb#L417

simonpoole:
> 
> 
> I believe a CSP entry for the new endpoint is missing here:
> 
> [openstreetmap-website/app/controllers/application_controller.rb](https://github.com/openstreetmap/openstreetmap-website/blob/ccdec3ed4c0f7c472f6db7e5eff2dfc1579f3dd1/app/controllers/application_controller.rb#L417)
> 
> Line 417 in [ccdec3e](/openstreetmap/openstreetmap-website/commit/ccdec3ed4c0f7c472f6db7e5eff2dfc1579f3dd1)
>  :connect_src => %w[nominatim.openstreetmap.org overpass-api.de router.project-osrm.org graphhopper.com],

Fixed in https://github.com/openstreetmap/openstreetmap-website/pull/2063/commits/181b078d052a2b65e69ce920d3a04825edc3a931 (/me mutters something about configuration should be in configuration).

tomhughes:
Well if the code that accesses the site is hard coded in the source why not the security rule? What would be the point of allowing the user to change to allow a different site if that wasn't the one the code was going to contact...

Where a site name used by the code comes from configuration we also populate the security rules from the same configuration.

That all seems to make sense to me, but maybe I am missing something?

mmd-osm:
> Where a site name used by the code comes from configuration we also populate the security rules from the same configuration.

This seems to be the case for PIWIK only at this time? It doesn't seemt to happen automagically, and requires changes to another location:

https://github.com/openstreetmap/openstreetmap-website/blob/1285bcbd712044720c15e2de2e6bacaf4b0b0a92/config/initializers/secure_headers.rb#L21

tomhughes:
Yes, to `config/piwik.yml` which is part of the piwik configuration ;-)

Yes splitting that between a flag in the main config file and a separate file might not be ideal but it is all configuration, and the name of the site is always coming from one place wherever it is used.

simonpoole:
@tomhughes the URls in question (external services) are in application.yml now it is true that the way things are done, those entries are used in a hardcoded fashion in code and there a many many ways to skin this particular car, but for example extracting the domains and adding them to a list (for constructing the CSP) could at least be localized in the actual code that uses the URL.

tomhughes:
Oh in that case you are absolutely right.

pnorman:
> Change looks good to me. As a member of OWG this gets my thumbs up.

Likewise, except I have not done a code review.

config/locales/en.yml
```diff
@@ -2441,13 +2441,15 @@ en:
     directions:
       ascend: "Ascend"
       engines:
+        fossgis_osrm_bike: "Bicycle (FOSSGIS OSRM)"
+        fossgis_osrm_car: "Car (FOSSGIS OSRM))"
+        fossgis_osrm_foot: "Foot (FOSSGIS OSRM))"
```
tomhughes: We seem to have a superfluity of parentheses here...

config/example.application.yml
```diff
@@ -102,6 +102,7 @@ defaults: &defaults
   graphhopper_url: "https://graphhopper.com/api/1/route"
   mapquest_directions_url: "https://open.mapquestapi.com/directions/v2/route"
   osrm_url: "https://router.project-osrm.org/route/v1/driving/"
```
tomhughes: This line can be removed I think?

app/assets/javascripts/index/directions/fossgis.js
```diff
@@ -119,9 +119,9 @@ function OSRMEngine() {
           namedRoad = false;
         }
 
-        if (step.maneuver.type.match(/^exit (rotary|roundabout)$/)) {
+        if (step.maneuver.type.match(/exit (rotary|roundabout)/)) {
```
tomhughes: What's the reason for removing the anchors from this, and the other matches? Is this server producing different directions that have something else before and after?

config/locales/en.yml
simonpoole: the parentheses should go, as to the naming: no strong feelings historically we've indicated the routing engine in one form or the other

config/example.application.yml
simonpoole: Assuming MQ doesn't rise from the semi-dead yes

app/assets/javascripts/index/directions/fossgis.js
simonpoole: No reason, because it is simply bitrot: the code is based on an old version, pre this commit https://github.com/openstreetmap/openstreetmap-website/commit/e9abea3dd79a11a6d756a9e3748ee6cc75911a10 of the file. I'll give bring the code up to the equivalent state.

config/example.application.yml
tomhughes: I meant `osrm_url` actually - the MQ one does still work if you have an API key though I guess it may just be using the non-open routing now?

config/locales/en.yml
simonpoole: Changed to ..... (OSRM) as we have attribution of FOSGISS in the footer of the routing results.

config/example.application.yml
simonpoole: I actually removed both and other lingering bits referring to MQ. Easy enough to resurrect in the unlikely case of them ever coming back.

app/assets/javascripts/index/directions/fossgis.js
```diff
@@ -1,12 +1,12 @@
 // OSRM car engine
```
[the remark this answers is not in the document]
simonpoole: @mmd-osm fossgis.js is not identical with osrm.js. At the time I originally authored it I was playing with the idea of supporting all in one (needs some special casing because of the slightly different URLS and lack of different profiles), but given that the MB instance is unlikely to come back, I punted.

simonpoole:
> 
> 
> > Other than the inline comments I have made I think my only outstanding question is how we want to refer to this - principally whether to include "OSRM" in the name or just refer to it as FOSSGIS?
> 
> Is "car (OSRM @ FOSSGIS)" short enough? Luckily both the engine and the operator of the instance have short names.

The current version of the PR only uses (OSRM) as Fossgis is attributed in the results.

pnorman:
> Other than the inline comments I have made I think my only outstanding question is how we want to refer to this - principally whether to include "OSRM" in the name or just refer to it as FOSSGIS?

In the past we had Mapzen, not Valhalla.

Issue #2064, opened 2018-11-14
Problem with deny_access and request formats during tests


gravitystorm:
I'm attempting to refactor changeset_comments_controller to use CanCanCan, and there's a problem with how our tests run, and how they interact with with our `deny_access` handler, and API request format handling in general.

Until now, you can call `post "api/0.6/changeset/comment/:id/hide"` with any request format you like, either html or xml or plain text or whatever, and since there's no payload it doesn't really matter very much. The action ignores the request format and always responds with XML (`render :xml => comment.changeset.to_xml.to_s`).

We take advantage of this in our tests, which are mostly of the format `post :destroy, :params => { :id => comment.id }` which is sending a text/html request behind the scenes. You can see the difference between "accept anything, always respond in XML" and "only accept XML requests" if you change the controller action to e.g.

```ruby
    # Return a copy of the updated changeset
    respond_to do |format|
      format.xml { render :xml => comment.changeset.to_xml.to_s }
    end
```
This would now throw an error on any request that's not XML, including the requests from the test suite.

The immediate problem is that our `deny_access` handler now uses the request format to decide whether to redirect requests to the login page (for html requests, assumed to be website requests) or not (for any other type of request). So when porting these API methods to CanCanCan, instead of receiving a 403 response (when you're logged in but don't have permissions, e.g. not a moderator) you get redirected to an html error page instead. This is because the test suite makes all its requests as text/html by default. Previously that hasn't mattered, but now it does.

Possible solutions include:

* Just change the test suite and accept the redirects. For example,
   ```diff
       # not a moderator
       post :restore, :params => { :id => comment.id }
  -   assert_response :forbidden
  +   assert_response :redirect
  +   assert_redirected_to :controller => :errors, :action => :forbidden
  ```
  But I think this is very misleading, since that's not how we're designing the XML API. I don't think it's a good idea to add in the redirect for real API responses either, since not all libraries follow redirects.

* Just change the test suite. For example, 
  ```diff
       # not a moderator
  -    post :restore, :params => { :id => comment.id }
  +    post :restore, :params => { :id => comment.id, :format => "xml" }
       assert_response :forbidden
  ```
  This could be tedious and repetitive though. A variation would involve e.g. `api_post` as a wrapper around `post` to set the format, perhaps. Although it makes the tests pass, any real-world API user sending text/html (are there any? Who knows?) would still be affected and get a redirect.

* Set a default format in a `before_action` handler in the controller
   ```diff
[...]

PR #2065, opened 2018-11-14
Set default format for all API routes

tomhughes:
This makes the change I proposed in #2064 to add a default format to all API routes.

### DIFF
--- a/test/controllers/nodes_controller_test.rb
+++ b/test/controllers/nodes_controller_test.rb
@@ -6,23 +6,23 @@ class NodesControllerTest < ActionController::TestCase
   def test_routes
     assert_routing(
       { :path => "/api/0.6/node/create", :method => :put },
-      { :controller => "nodes", :action => "create" }
+      { :controller => "nodes", :action => "create", :format => "xml" }
     )
     assert_routing(
       { :path => "/api/0.6/node/1", :method => :get },
-      { :controller => "nodes", :action => "read", :id => "1" }
+      { :controller => "nodes", :action => "read", :id => "1", :format => "xml" }
     )
     assert_routing(
       { :path => "/api/0.6/node/1", :method => :put },
-      { :controller => "nodes", :action => "update", :id => "1" }
+      { :controller => "nodes", :action => "update", :id => "1", :format => "xml" }
     )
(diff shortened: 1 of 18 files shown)


tomhughes:
This may not be such a good idea... It seems if you default the format that way then while an explicit extension will still override it an `Accept` header won't.

I think this probably needs more research...

test/controllers/nodes_controller_test.rb
```diff
@@ -6,23 +6,23 @@ class NodesControllerTest < ActionController::TestCase
...
     )
     assert_routing(
       { :path => "/api/0.6/node/1", :method => :put },
-      { :controller => "nodes", :action => "update", :id => "1" }
+      { :controller => "nodes", :action => "update", :id => "1", :format => "xml" }
     )
     assert_routing(
       { :path => "/api/0.6/node/1", :method => :delete },
-      { :controller => "nodes", :action => "delete", :id => "1" }
+      { :controller => "nodes", :action => "delete", :id => "1", :format => "xml" }
```
mmd-osm: PUT and DELETE related requests in this controller (as well as in the ways and relations controller) return plain text as a response (see https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/api/nodes_controller.rb#L54-L62).

I have some doubts if it's correct to assume format :xml in this case, and would only do so in case the response really uses XML format. Maybe I'm missing something here, though.

tomhughes: Do those methods actually return anything at all?

tomhughes: Just the version number, which isn't XML.

Oh hang on, this is an ancient and known-not-working PR of mine, so I'm not sure what your point is? I has assumed it was your XML rework...

mmd-osm: Oh, right, sorry for the confusion. It felt a bit silly to have so much duplication in my XML rework, started looking around, and came across this old PR. I was wondering wether that's something worth considering.

gravitystorm:
Given the conflicts and the lack of urgency in figuring out how to handle the Accepts header stuff, I'm going to close this PR.

PR #2068, opened 2018-11-19
Move tracepoint ordering logic into tracepoint model, and add tests

### LINKED ISSUE #2059: Tests for tracepoint ordering
https://github.com/openstreetmap/openstreetmap-website/commit/cdb42d2a6ce539be3f3e50edebb88240dfd16e04 was a hotfix for an important privacy problem, but hasn't yet received any tests.

I would also recommend refactoring the complex logic out of the controller and into the TracePoint model, which will make it easier to test too.

krubokrubo:
Closes #2059 

This should behave exactly as before, just moving the logic into the model.

### DIFF
--- a/app/models/tracepoint.rb
+++ b/app/models/tracepoint.rb
@@ -39,4 +39,12 @@ class Tracepoint < ActiveRecord::Base
     el1 << (XML::Node.new("time") << timestamp.xmlschema) if print_timestamp
     el1
   end
+
+  # Return tracepoints in bbox but hide the order of non-trackable tracepoints
+  def self.in_bbox(bbox)
+    ordered_points = self.bbox(bbox).joins(:trace).where(:gpx_files => { :visibility => %w[trackable identifiable] }).order("gpx_id DESC, trackid ASC, timestamp ASC")
+    unordered_points = self.bbox(bbox).joins(:trace).where(:gpx_files => { :visibility => %w[public private] }).order("gps_points.latitude", "gps_points.longitude")
+    ordered_points.union_all(unordered_points)
+  end
+
 end
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -30,9 +30,7 @@ class ApiController < ApplicationController
     end
 
     # get all the points
-    ordered_points = Tracepoint.bbox(bbox).joins(:trace).where(:gpx_files => { :visibility => %w[trackable identifiable] }).order("gpx_id DESC, trackid ASC, timestamp ASC")
-    unordered_points = Tracepoint.bbox(bbox).joins(:trace).where(:gpx_files => { :visibility => %w[public private] }).order("gps_points.latitude", "gps_points.longitude", "gps_points.timestamp")
-    points = ordered_points.union_all(unordered_points).offset(offset).limit(TRACEPOINTS_PER_PAGE)
+    points = Tracepoint.in_bbox(bbox).offset(offset).limit(TRACEPOINTS_PER_PAGE)
 
     doc = XML::Document.new
     doc.encoding = XML::Encoding::UTF_8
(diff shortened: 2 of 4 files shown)


tomhughes:
I must admit that I wasn't really sure what @gravitystorm though was the benefit of moving this (rather special case) login into the model was...

Certainly `in_bbox` doesn't seem to be a full description of what it does, or rather it implies something much simpler than what it really is.

krubokrubo:
I see what you mean. I can give it a longer name like `in_bbox_with_ordering_privacy`. Also it looks like I committed a bunch of rubocop offenses. I'll get those fixed. Meanwhile maybe @gravitystorm will let us know what he had in mind.

tomhughes:
For what it's worth my immediate thought at the time was to maybe add `ordered` and `unordered` scopes on the model  but still do the union in the controller. Not sure where the bbox goes though - could be in the controller or could be an argument on the scopes.

krubokrubo:
That makes sense to me.  I moved the tracepoint privacy ordering into scopes on the model now, doing the union in the controller.  I'm just chaining them with the existing bbox scope, which seems to me like the simplest way.  I had some mixups with my commits (you can see my force-push history here), but cleared up the rubocop offenses and other issues I could find.  Let me know if it's suitable?

gravitystorm:
> Meanwhile maybe @gravitystorm will let us know what he had in mind.

It was just a general point. Moving complex logic to models makes it easier to test, among other things. For example, it's easier to check that the code selects the correct objects from the database in a model test, rather than having to parse the XML output of a controller to check if the correct objects were retrieved (and e.g. not extra objects).

Tom's suggestion of using scopes to abstract away the complexity from the controller is the kind of thing that I had in mind. The scopes are a good way to ensure that we don't end up making complex queries in multiple places, as well as being a natural home for custom Tracepoint finder methods that go being straightforward `find ... order ... limit` that you normally get in controllers.

Anyway, it was more of a "if you're working on this then moving the code would help" rather than any major problem.

app/models/tracepoint.rb
```diff
@@ -39,4 +39,10 @@ def to_xml_node(print_timestamp = false)
     el1 << (XML::Node.new("time") << timestamp.xmlschema) if print_timestamp
     el1
   end
+
+  # Return points of trackable traces in original order
+  scope :trackable_ordered, -> { joins(:trace).where(:gpx_files => { :visibility => %w[trackable identifiable] }).order("gpx_id DESC, trackid ASC, timestamp ASC") }
+
+  # Hide the order of points of non-trackable traces for privacy
+  scope :non_trackable_unordered, -> { joins(:trace).where(:gpx_files => { :visibility => %w[public private] }).order("gps_points.latitude", "gps_points.longitude") }
```
tomhughes: Can you move these to the top please, for consistency with other models - normally we put scopes immediately after the associations at the top of the model.

As to names I agree with @gravitystorm that these names may be confusing but I'm not sure the ones I suggested are perfect either so I'm not really sure what's best. The goal of my names was just to separate point which could be ordered (as `ordered`) from those which can't (as `unordered`) but I realise that may not be very clear from the names.

krubokrubo: Ah, I see I missed the model format earlier. Yes, I moved the new scopes to the right location in the model now.

I'm not sure what to suggest for names.  As a newcomer it makes sense to me that identifiable and trackable traces both have the property trackable?==true and are both included in a trackable scope.  `unordered` is trickier---perhaps the ones that have trackable?==false could be called `anonymized` or `non_trackable_anonymized` or just `non_trackable`?  But as I study it more, "public" traces still puzzle me---I don't see the value of anonymizing the points on the map while the trace remains publicly available for download.  I'm guessing this has already been discussed elsewhere.

PR #2071, opened 2018-11-24
Show tel: links on multiple phone numbers separated by ;

krubokrubo:
I actually implemented the feature for multiple phone numbers in afbb28aa.  This increased [`browse_helper.rb`](https://github.com/krubokrubo/openstreetmap-website/blob/afbb28aa11f5e59f61b464ccb095122595b235c1/app/helpers/browse_helper.rb) by 8 lines of code.

Unfortunately this raised a rubocop offense for ModuleLength.  So I made the current commit, 52694d23, to split [`browse_helper.rb`](https://github.com/krubokrubo/openstreetmap-website/blob/52694d23f38ef2a586c14a11badac8f992dd80ec/app/helpers/browse_helper.rb) and a new file, [`format_links_helper.rb`](https://github.com/krubokrubo/openstreetmap-website/blob/52694d23f38ef2a586c14a11badac8f992dd80ec/app/helpers/format_links_helper.rb).  Now rubocop doesn't find any offense, but please check if this is an appropriate way to do it.


tomhughes:
This is really hard to review because you seem to have made a lot of non-essential changes to the tests.

It would probably have been better to leave the existing tests (bar the changes needed to adapt to the API change) and add new tests for the semicolon separated lists to the end so that comparison tools are able to show the changes more clearly.

I think I'm right in saying that the only function change is that `/` is no longer allowed in numbers? Why did you choose to do that?

Issue #2069, opened 2018-11-22
Properly mark `tel:` link on multiple phone numbers in the POI


nachanon:
In Thailand, it is very common for moderate-sized stores and SME companies to provide multiple phone numbers rather than using single-number trunk line. (And you are expected to try the alternate numbers if you got a busy signal) These occurences are documented as such in my mapping journey by using the usual [semicolon-separated values](https://wiki.openstreetmap.org/wiki/Semi-colon_value_separator).

The problem is, main OpenStreetMap interface **does not seem to mark _any_ `tel:` link on POI's [`phone`](https://wiki.openstreetmap.org/wiki/Key:phone) key if this field contains semicolon separator(s)**. This leaves numbers in affected POIs as plain text, and user must manually select and copy each of them into a dialer when they need to call-- a minor incovenience for desktop, but very painful for mobile.

Examples and screenshots:

- [Nan Wan](https://www.openstreetmap.org/way/637578269)
	
		phone = +66 86 394 4382;+66 62 559 6156
	
	![](https://u.cubeupload.com/xwindows/dA5GhV.png)

- [Bangkok Art and Culture Centre](https://www.openstreetmap.org/way/48966995)
	
		phone = +66 2 214 6630;+66 2 214 6631;+66 2 214 6632;+66 2 214 6633;+66 2 214 6634;+66 2 214 6635;+66 2 214 6636;+66 2 214 6637;+66 2 214 6638
	
	![](https://u.cubeupload.com/xwindows/tVufps.png)

- [Krungdeb Co-Operative Store](https://www.openstreetmap.org/node/5304732122)
	
		phone = +66 2 281 8228;+66 2 282 2489;+66 2 629 4265; +66 2 629 4266
	
	![](https://u.cubeupload.com/xwindows/T9ixX0.png)

You would see that the telephone numbers are not autolinked. See _Notes_ below for more (many, many more) examples.

Compare and contrast with POI bearing a single phone number:

- [Pat Vintage Cafe](https://www.openstreetmap.org/way/234856667)
	
	`phone = +66 2 116 9966`
	
	![](https://u.cubeupload.com/xwindows/6DpQNU.png)
	
	You would see from the tooltip that phone number field is properly linked to [tel:+6621169966](tel:+6621169966).

Please investigate and fix.

Browser: Opera 11.64 (Win32)
System: Microsoft Windows XP Professional SP3

-----

### Notes ###

[...]

PR #2072, opened 2018-11-27
Show tel: links for multiple phone numbers separated by ;

krubokrubo:
This is the feature for multiple phone numbers (#2069) again.  I'm now following the existing code in accepting the / symbol in phone number links.  I also adjusted the tests so the comparison is clean.

(Note: I'm submitting this as a new PR so that the commit history is cleaner than my first try (#2071).  github won't let me reopen the same PR since I force-pushed to the same branch.)

The feature is now implemented in 5ba64efd, which led to rubocop offenses that I resolved in d44ae2ef and 4826a6e3.  Let me know if this is suitable.


tomhughes:
You don't need to open a new PR actually - you can just force push to the branch for the previous one.

tomhughes:
What was the logic behind "fixing" the `AssignmentInCondition` offenses? It doesn't look like that would have triggered at the point so I'm guessing the issue was that moving them to the new helper would cause them to start triggering unless you excluded that?

For the record the reason I never fixed that is that I didn't see it as a major problem and I find the result extremely ugly :-(

That said there are a number of other places where I have similar feelings but have given in and held my nose and let rubocop have it's way so maybe we should do the same here - either that or just decide we don't like that rule and disable it globally.

app/helpers/format_links_helper.rb
```diff
@@ -0,0 +1,138 @@
+require "cgi"
+
+module FormatLinksHelper
```
tomhughes: If we're going to split this (the alternative being to just increase the limit as we've never worried about the rubocop length limits much) then I thing something like `BrowseTagsHelper` would be better given these helper methods are about dealing with formatting tag keys and values for the data browser rather than more general link formatting.

krubokrubo: Sounds good, this is renamed now.

I'm fairly new to Ruby and rubocop, so I don't yet have an opinion about the rubocop rules, I was just trying to resolve the cascade of offenses. (First I got ModuleLength, then when I split the module it led to AssignmentInCondition, although I realize I made this unclear by committing the fixes in reverse order.)

If you're not concerned about those offenses, I'm happy to leave them as they were.  In that case, we could just look at commit 5ba64ef for this feature.

tomhughes:
I've removed the `AssignmentInCondition` change from this branch - if we want to condisder that I think we should discuss it separately.

krubokrubo:
Glad to see this is working! Thanks for your guidance on it.

PR #2083, opened 2018-12-12
Use only token capabilities when a token is provided

gravitystorm:
The Authenticate#allow? method (from oauth-plugin) sets `current_user` as a side effect of checking the token. But this allowed a valid token to access all actions that are available to that user, beyond the capabilities for that token.

This PR changes that, so that if there is a token, only the permissions from the Capability class are used. I've added a couple of tests that illustrate the change, to show that with the new approach you definitely need the allow_read_prefs permission on a token.

I went for the @request.env fiddling to make the request, since creating a fully signed request (as used in test/integration/oauth_test) is a bit more complex than I would like. But I'm open to other suggestions.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -446,9 +446,9 @@ class ApplicationController < ActionController::Base
   end
 
   def current_ability
-    # Add in capabilities from the oauth token if it exists and is a valid access token
+    # Use capabilities from the oauth token if it exists and is a valid access token
     if Authenticator.new(self, [:token]).allow?
-      Ability.new(current_user).merge(Capability.new(current_token))
+      Capability.new(current_token)
     else
       Ability.new(current_user)
     end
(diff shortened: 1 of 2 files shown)


tomhughes:
I had to partially revert this in eb7c4cdedd4f18ceb7362b0f1766f4a8b0dfca62 because we still need to allow actions which don't require a login.

I'm not sure if my solution there is the best one or not, or if we should separate out the common actions that don't require a login into another class?

gravitystorm:
Ah, sorry about that. I think your approach is fine for now, because the tokens are only set up for particular actions (via `authorize` vs `authorize_web`). 

I want to avoid allowing tokens (and basic auth) to be used for non-api methods by mistake, so technically they shouldn't get all the permissions from Ability.new(nil) since that includes a lot of non-API methods. But we can come back to this later, since (afaict) it doesn't matter yet.

PR #2098, opened 2019-01-03
Add parsing of bare #zoom/lat/lon style hashes

JesseWeinstein:
mapbox-gl-js creates bare hashes without a querystring; this (very small) change will enable OSM to respond correctly to them, if people paste them in by accident.

I've also made a PR for the reverse: https://github.com/mapbox/mapbox-gl-js/pull/7725

### DIFF
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -148,20 +148,25 @@ OSM = {
       return args;
     }
 
-    hash = querystring.parse(hash.substr(i + 1));
+    let hash_map;
+    if (/\d/.test(hash.substr(i + 1, 1))) {
+      hash_map = hash.substr(i + 1);
+    } else {             
+      hash = querystring.parse(hash.substr(i + 1));
 
-    var map = (hash.map || '').split('/'),
-      zoom = parseInt(map[0], 10),
-      lat = parseFloat(map[1]),
-      lon = parseFloat(map[2]);
+      if (hash.layers) {
+        args.layers = hash.layers;
+      }
+      hash_map = (hash.map || '');
+    }
+    var map = hash_map.split('/'),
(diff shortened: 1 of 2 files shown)


tomhughes:
Well there's no need for both PRs surely, and our URL format is long established so I'm not sure why we should be expected to add support for endless lists of formats produced by other pieces of software - if they want to generate links to osm.org they should use the correct OSM URL format.

grischard:
I like this format, it's shorter.

JesseWeinstein:
Thank you both for your feedback. The need for both PRs is because it is courteous to avoid barfing on common input (not an "endless list of formats"), and that applies to both sides. 

This is not about *generating* links (for which, I agree, it's trivial to just generate the canonical format), but about interactive use by humans, who often fail to remember fine distinctions (like whether or not a four character prefix is required in a particular context). 

I do not intend to be implying that you "should be expected" to do this, or anything else -- merely to point out that, should you wish to respond more kindly to users making this common mistake, here is the code necessary to do so.

gravitystorm:
@JesseWeinstein Thanks for the suggested feature. However, I don't believe it's a common enough use-case for us to add and maintain the additional code.

JesseWeinstein:
Ten lines and about one more branch is too much "additional code"? That ... does not seem like a plausible justification. 

I can sympathize with the idea that adding this additional (trivial) variant format would excessively complicate the API -- in the sense that we would need to document (and test) both formats, and ensure they keep working. Is that more like what you were referring to?

gravitystorm:
> I can sympathize with the idea that adding this additional (trivial) variant format would excessively complicate the API -- in the sense that we would need to document (and test) both formats, and ensure they keep working. Is that more like what you were referring to?

Yep, that's a reasonable way to explain it.

PR #2101, opened 2019-01-08
Update zh-CN.yml

tianyinlee:
Correct some obvious translation errors


tomhughes:
Translation updates need to be made using TranslateWiki - see https://wiki.openstreetmap.org/wiki/Website_internationalization#How_to_translate for more information.

gravitystorm:
We get fresh updates from translatewiki every few days - you can see from the commit history at https://github.com/openstreetmap/openstreetmap-website/commits/master

Issue #2103, opened 2019-01-09
Nested lists are incorrectly displayed in user messaging interface


nachanon:
I ran into this problem when I was writing a rather long reply to another mapper's message. I have drafted my message with [original Markdown][1]; but when I previewed the message before sending, I found several problems with [OpenStreetMap.org](https://www.openstreetmap.org/)'s Markdown implementation around its interpretation of nested numbered/bullet list.

This resulted in me having fumble for several minutes to find _partial_ workaround (see below), before I could actually send the message in a relatively unmangled form.

Exact steps to reproduce:

- Start a browser with a new private browsing session (or with cookies cleared) and JavaScript enabled.
- Visit your own OSM user page as anonymous user.
- Click _Send Message_.
- You will be prompted to log in: enter your OSM user and password.
- <q>Send a new message</q> page will appear.
- Enter `These are nested lists...` in the _Subject_ field.
- **Copy** text from [this Gist](https://gist.githubusercontent.com/nachanon/320b8a0473ac0658a8f2fe8ddd9da6c7/raw/8c94d376792af491c71795a2f225e5212ed046eb/nestedlist.md) **and paste** it into the _Body_ field. (Warning: Tab indentation on blank lines are totally intended)
- Click the _Preview_ button.

The preview would show something like this:
![](http://u.cubeupload.com/xwindows/ZMoga4.png)

Then, compare it to an output from [the original Markdown][1]:
![](http://u.cubeupload.com/xwindows/0qf3OL.png)

You would see that on OSM messaging interface...

- _Bullet_ lists that are nested inside numbered list **are mistakenly displayed as _numbered_ lists**. **(No workaround)**
- Sub lists with _equally-indented_ blank lines before and after them **are mistakenly displayed as code snippet _and_ excluded from the list**. (Workaround: remove these lines before sending the message)

Please investigate and fix.

Browser: Opera 11.64 (Win32)  
System: Microsoft Windows XP Professional SP3

### Notes ###

- This is not specific to preview: the same message sent to myself also exhibit the same problem when viewed in the on-site inbox.
- This is not specific to browser: result obtained from Pale Moon 25.8.1 (GNU/Linux 32-bit) is the same.  
- The word _Markdown_ on the messaging page's help panel actually links to [the original Markdown][1]. So, compatibility is expected.

[1]: https://daringfireball.net/projects/markdown/

tomhughes:
[...]

PR #2111, opened 2019-01-09
Switch to using kramdown for markdown formatting

tomhughes:
As suggested by @gravitystorm in #2103 this switches our markdown formatting to use kramdown instead of redcarpet.

As kramdown doesn't have the builtin sanitisation features of redcarpet I have passed the resulting HTML through the sanitiser used for other formats and have also added a few extra restrictions to links in user generated content.

Note that this doesn't fix all the issues in #2103 as bullet lists nested in numbered lists still don't seem to work as the reporter there is expecting.

Also I noticed that redcarpet was turning thinks which look like links but have no link markup into links which kramdown doesn't - we could fix that by running our autolinker if we wanted.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,7 +77,7 @@ gem "omniauth-openid"
 gem "omniauth-windowslive"
 
 # Markdown formatting support
-gem "redcarpet"
+gem "kramdown"
 
 # For status transitions of Issues
 gem "aasm"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -161,6 +161,7 @@ GEM
       jsonify (< 0.4.0)
     jwt (2.1.0)
     kgio (2.11.2)
+    kramdown (1.17.0)
     libv8 (3.16.14.19)
     libxml-ruby (3.1.0)
     listen (3.1.5)
@@ -303,7 +304,6 @@ GEM
       ffi (~> 1.0)
     record_tag_helper (1.0.0)
       actionview (~> 5.x)
-    redcarpet (3.4.0)
     ref (2.0.0)
     request_store (1.4.1)
       rack (>= 1.4)
@@ -412,6 +412,7 @@ DEPENDENCIES
   json
   jsonify-rails
   kgio
+  kramdown
   libxml-ruby (>= 2.0.5)
   listen
   logstasher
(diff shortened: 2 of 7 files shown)


tomhughes:
I've added in the auto linkification for now - if we decide to remove it then some tests would need to be fixed.

gravitystorm:
OK, looks good to me!

> Note that this doesn't fix all the issues in #2103 as bullet lists nested in numbered lists still don't seem to work as the reporter there is expecting.

That's a problem with our stylesheets - the html outputted by kramdown seems correct as it's outputting `<ul>` tags for the bullet lists. But [in our stylesheets](https://github.com/openstreetmap/openstreetmap-website/blob/fbbabeff1e3bc473720e6eaece83825a9c349ba8/app/assets/stylesheets/common.scss#L2349-L2351) we have:

```
ol li {
  list-style: decimal;
}
```
... and as far as I can tell, that means that any `<ul>` instead of a `<ol><li>` will inherit the decimal list style. Perhaps we could fix that as a followup.

tomhughes:
Good spot! Added a fix for that in 81b37f9263ec0d76b60b3273e5f9d9a4345bfbfe.

mmd-osm:
There's an ongoing discussion in a diary post about kramdown causing some regressions for some people. More specifically it seems that images with links to other sites somehow get rendered in a different way, and in some cases no longer work, see https://www.openstreetmap.org/user/MKnight/diary/40702 as an example.

Using the following style (http://www.disturbancesinthewash.net/journal/2012/8/11/how-to-add-an-image-with-link-in-markdown.html) seems to be ok, though.

tomhughes:
If you want anybody to investigate then you'll have to provide some actual details of what doesn't work, rather than what does work, which isn't very interesting.

tomhughes:
Sorry I assumed that the "discussion" you referred to (which I didn't try and translate) had come some conclusion ;-)

Some of the other ones are worse in that they do `alt text (link url)` with no proper link syntax at all - that was never valid.

tomhughes:
I think I've fixed them all now.

mmd-osm:
>  I assumed that the "discussion" you referred to (which I didn't try and translate) had come some conclusion I think there was no real conclusion, other than the previous syntax didn't work anymore. 

> I think I've fixed them all now.

Issue #281, opened 2013-05-14
Status of picture/icon methods in lib/gpx.rb


gravitystorm:
The two methods mentioned try calling "Magick::Image" and "Magick::ImageList", which I guess are supposed to be RMagick calls. And I also guess that since rmagick was removed some time ago, these calls aren't actually being used in production, i.e. the standalone gpx-importer is making the images.

So as written, these methods throw errors when called from the console.

```
gpx = GPX::File.new('')
gpx.picture(0,0,1,1,10)
NameError: uninitialized constant GPX::File::Magick
```

Should we keep these methods, write some tests for them and fix the rmagick dependency? This way, developers could upload GPX traces for testing, without needing to install the gpx-importer. Or should we remove them, since the code isn't used in production, they don't work, and they don't have any tests?

mmd-osm:
Just a thought: could we backport the image.c code in gpx-import to ruby, using https://github.com/dark-panda/gd2-ffij (or maybe the original gd2 library)? This would only introduce a dependency on gd2, rather than some Imagemagick stuff.

https://github.com/openstreetmap/gpx-import/blob/master/src/image.c

mmd-osm:
Rendering a GPX trace icon isn't too difficult using gd2-ffij:

```ruby
    require 'gd2-ffij'

[...]

    def icon(min_lat, min_lon, max_lat, max_lon)
      width = 50
      height = 50
      proj = OSM::Mercator.new(min_lat, min_lon, max_lat, max_lon, width, height)

      image = GD2::Image::IndexedColor.new(width, height)

      black = image.palette.allocate(GD2::Color[0, 0, 0])
      white = image.palette.allocate(GD2::Color[255, 255, 255])

      image.draw do |pen|
        pen.color = white
        pen.rectangle(0, 0, width, height, true)
      end

      image.draw do |pen|
        pen.color = black
        pen.anti_aliasing = true
        pen.dont_blend = false

        oldpx = 0.0
        oldpy = 0.0

        first = true

        points do |p|
          px = proj.x(p.longitude)
          py = proj.y(p.latitude)

          pen.line(px, py, oldpx, oldpy) unless first

          first = false
          oldpy = py
          oldpx = px
        end

      end
  
      image.gif
    end
  end
```

gravitystorm:
This has been resolved by #2120, which now uses the [gd2-ffij](https://github.com/dark-panda/gd2-ffij) library for creating pictures and icons.

PR #2120, opened 2019-01-16
Optional Trace import using ActiveJob

gravitystorm:
Coming out of the conversations in #281 and #1852, I've started working on a mechanism to import and delete traces using the job queue, and avoiding the need for the [external gpx importer](https://github.com/openstreetmap/gpx-import) completely.

My first aim is to provide people with an option to import traces in development, by providing a configuration option to enable it, and fixing the icon creation to work again. This is what I've done so far, and I'm putting it up here for initial comments. It doesn't have any tests yet, hence the WIP label. But I think it could be merged as-is, since it's off by default and the icon methods have been broken for years anyway.

Beyond this stage, I would see the plan (in no particular order) as roughly:
* Write some tests to make sure this all works (done)
* Get the animated gifs working (see https://github.com/dark-panda/gd2-ffij/issues/18)
* Get OSMF to set the option to false in the chef scripts, then flip the option to true in the defaults, so that the job queue works by default for new developers / other deployments
* Fix the importing to use [activerecord-import](https://github.com/zdennis/activerecord-import), to cut down the number of inserts per trace

OSMF can then switch off the external importer and use this one instead.

Initial thoughts/comments/etc welcome. Big thanks to @mmd-osm for investigating both gd2-ffij and activerecord-import


gravitystorm:
I think this is ready for review/merging, now that the tests are written. There's obviously two major limitations with this code right now (the lack of the animated gif for the large picture, and the 1-sql-insert-per-tracepoint thing), but I'd rather see those fixed those in future PRs rather than doing everything in one PR.

test/jobs/trace_importer_job_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    gpx = Minitest::Mock.new
+    def gpx.actual_points
+      5
+    end
+
+    trace.stub(:import, gpx) do
+      perform_enqueued_jobs do
+        TraceImporterJob.perform_now(trace)
+      end
+    end
```
tomhughes: I'm struggling to understand this, or to find any documentation of the `stub` method that would help me. I assume it's a FactoryBot thing but although I find occasional references to it on google the actual documentation never seems to mention it.

I assume it stubs the `import` method but I don't see when this test would ever call that method? or why calling import which would normally be called from the job should actually run the job?

I'm probably just being dense but I'd like to understand what's going on ;-)

app/jobs/trace_destroyer_job.rb
```diff
@@ -0,0 +1,10 @@
+class TraceDestroyerJob < ApplicationJob
+  queue_as :default
+
+  def perform(trace)
+    trace.destroy
+  rescue StandardError => ex
+    logger.info ex.to_s
+    ex.backtrace.each { |l| logger.info l }
```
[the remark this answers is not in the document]
gravitystorm: Just copy+pasted from the previous daemon code. I agree, it makes sense to remove the exception handling from here.

test/jobs/trace_importer_job_test.rb
gravitystorm: It's from minitest/mock. 

https://www.rubydoc.info/gems/minitest/4.2.0/Object:stub

The stubbing takes a block, and the stub is only effective inside the block. So when we call `TraceImporterJob.perform_now(trace)`, the `trace.import` method has been overwritten to just return the (fake) gpx object and avoid doing anything else (parsing xml, creating images on disk etc).

I did all this before realising that I needed the FakeFS for other stuff anyway. So if you prefer, I could rework this to create a fully working `trace` object that actually runs `trace.import` for real, with suitable setup to put a valid xml file in the right place and clean up the side-effects of running the `.import` code. Or we can stick with mocking/stubbing, which is probably a better way to do it, but perhaps a bit impenetrable?

mmd-osm:
> There's obviously two major limitations with this code right now (the lack of the animated gif for the large picture, and the 1-sql-insert-per-tracepoint thing), but I'd rather see those fixed those in future PRs rather than doing everything in one PR.

Animated gif support may be a bit tricky, as the repo owner hasn't been all that responsive lately. Maybe we'd have to fork it, and run our version of it, and find someone to finialize that animated gif part? 

Bulk INSERT is quite crucial from a performance pov, otherwise this thing is barely usable. Given that we already have that code lying around (https://github.com/openstreetmap/openstreetmap-website/issues/1852#issuecomment-427116280), and @tomhughes already reviewed a few iterations, I don't see why we couldn't integrate this part right away.

gravitystorm:
> Bulk INSERT is quite crucial from a performance pov, otherwise this thing is barely usable

I worked on it last week too, after I made the pull request, and it was [pretty straightforward](https://github.com/gravitystorm/openstreetmap-website/commit/74e1b98ea2701b722ebbb124e03f272848c51676). I noticed that the way the import method currently works involved multiple passes at the XML parsing, so I want to look at that too. But in general, I'm a fan of incremental pull requests, to make the reviews more straightforward.

I'll work on this further next week.

mmd-osm:
Some folks on Stackoverflow reported an issue with gd2-ffij on MacOS: "Couldn't load the gd2 library". This may be some documentation issue in INSTALL.md. I asked them to create a new issue here, as I can't reproduce this locally.

Issue #2124, opened 2019-01-29
Various parts of website broken in older browsers, errors in console


Shepard:
A recent change seems to have introduced some problems with the website when using it with older browsers, like IE11. Features like searching, add note and query features don't work. Opening the browser console reveals some error messages:

```
Uncaught TypeError: n.findIndex is not a function
Uncaught TypeError: Cannot read property 'route' of undefined
Unrecognized Content-Security-Policy directive 'worker-src'.
jQuery.Deferred exception: n.findIndex is not a function TypeError: n.findIndex is not a function
```

It seems like commit 48e886784e742ec28e33c32b3a0e0bf7cf6d21e7 might be at least partly responsible for the errors seen in the console. According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex `findIndex` is not supported in IE11.

Shepard:
This error reappeared: "Uncaught TypeError: n.findIndex is not a function".
Should I open a new ticket?

tomhughes:
What browser?

Shepard:
Sorry for the late response. This was on Chromium 43.

HolgerJeromin:
Using a four year old browser seems not a good idea.

tomhughes:
Just too old as well. By the looks of it 45+ would work... I've pushed https://github.com/openstreetmap/openstreetmap-website/commit/d74dd80540824b86fc93fc9573f412bbbcaefbc4 now to tighten things up a bit.

PR #2125, opened 2019-01-30
Replace augment.js with a modern set of polyfills where needed

tomhughes:
We've had `augment.js` in our vendor tree for some time now which provides polyfills for a number of ES5 methods but it looks pretty dead upstream.

Now we could quite possibly get away without it now - only really ancient browsers like IE8 are really affected as anything even vaguely modern has fairly complete ES5 support. As I discovered last night however (with `Array.prototype.findIndex`) people certainly notice is you rely on ES6 though even that is supported by most browsers (excluding all IEs obviously).

So what this does is to replace `augment.js` with a reasonably complete set of ES5 and ES6 polyfills (from `polyfill.io`) but it uses browser detection to avoid sending them to modern browsers.

Currently it assumes that WebKit/Firefox/Safari/Edge are OK which certainly seems to be the case for current versions and we can always tune it if it turns out that some older versions need help.


HolgerJeromin:
ref #2124

tomhughes:
I deliberately used a whitelist, after some consideration.

You of course correct that it's impossible to know the set of browsers that don't need it. It's equally impossible to know the set that do however, and as the polyfills do nothing when not needed I decided it was safer to err on the side of caution and send them when they might be needed - not sending them is basically just a performance optimisation.

I would also suggest that a substantial proportion of "other browsers" that would count as modern enough are in fact WebKit based anyway and would therefore be covered by this whitelist unless they mask that fact from their user agent.

gravitystorm:
I'm no javascript expert, but overall looks good to me.

On the browser whitelist/blacklist point, I can see both sides. This is, after all, just a small optimisation so it doesn't really matter - we could just send the polyfills to everyone.

I thought about suggesting `browser.modern?` as a way to let upstream deal with new browsers (since they will have their finger on the pulse much more than we do). But I'm not certain that it will help much since it doesn't tie into specific features like es6 support - so the upstream definition of `.modern?` might change significantly as the years go by.

PR #2131, opened 2019-02-06
Use activerecord-import for bulk importing tracepoint records

gravitystorm:
Non-rigourous testing shows a significant speedup, even on ssds.

### DIFF
--- a/app/models/trace.rb
+++ b/app/models/trace.rb
@@ -289,6 +289,8 @@ class Trace < ActiveRecord::Base
     # If there are any existing points for this trace then delete them
     Tracepoint.where(:gpx_id => id).delete_all
 
+    # Gather the trace points together for a bulk import
+    tracepoints = []
     gpx.points do |point|
       if first
         f_lat = point.latitude
@@ -303,8 +305,15 @@ class Trace < ActiveRecord::Base
       tp.timestamp = point.timestamp
       tp.gpx_id = id
       tp.trackid = point.segment
-      tp.save!
+      tracepoints << tp
     end
+
+    # Run the before_save and before_create callbacks, and then import them in bulk with activerecord-import
+    tracepoints.each do |tp|
+      tp.run_callbacks(:save) { false }
+      tp.run_callbacks(:create) { false }
+    end
+    Tracepoint.import(tracepoints)
 
     if gpx.actual_points.positive?
       max_lat = Tracepoint.where(:gpx_id => id).maximum(:latitude)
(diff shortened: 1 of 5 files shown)


gravitystorm:
One consideration here is around memory usage. The gpx.points yields tracepoints one at a time, which keeps memory usage low. But to insert them in bulk, we obviously need many tracepoints available at the same time, so we collect them up before bulk inserting.

Thoughts welcome! Do you think this will work as-is, or do we need to be more defensive about building the tracepoint array?

app/models/trace.rb
```diff
@@ -303,9 +305,16 @@ def import
...
-      tp.save!
+      tracepoints << tp
     end
 
+    # Run the before_save and before_create callbacks, and then import them in bulk with activerecord-import
+    tracepoints.each do |tp|
+      tp.run_callbacks(:save) { false }
+      tp.run_callbacks(:create) { false }
+    end
+    Tracepoint.import(tracepoints)
```
tomhughes: Shouldn't this be `import!` to raise an exception on error? The old save calls were written that way?

tomhughes:
As far as memory goes, the largest trace we have seems to have about fifteen million points, which raises a few issues - both memory as you say but also the size of the SQL if they are all inserted in one go!

Maybe we should do them in batches of a thousand or something?

app/models/trace.rb
mmd-osm: Re. batch size: https://github.com/zdennis/activerecord-import#batching mentions some parameter to control the max number of rows per INSERT.

mmd-osm:
I wonder if we should introduce maximum gpx file size limit to control memory consumption and processing time. 10-20 MB after decompression (and still before parsing) could be large enough for most users, and still reasonably small.

gravitystorm:
> We also need to improve zip file handling in general here, so people can't kill the server by uploading funny zip bombs.

@mmd-osm Could you open a separate issue for this please?

app/models/trace.rb
gravitystorm: I fixed this in 4d45a35f15c7f92c125ff070bf4757adc064ef9f

tomhughes: Well that stops us generating one mammoth SQL statement yes, but it still creates in-memory objects for all the points. Not sure whether that is actually an issue or not?

gravitystorm: Good point, but I don't know, tbh.

I can look at doing something clever to yield the points in batches too.

gravitystorm: OK, I've found a way to use `enum_for` to allow a simple way to pull points out of gpx.points using `.each_slice`. That means we will do multiple bulk_imports but without having more than 1,000 points in memory at one time.

PR #2161, opened 2019-02-28
Move API controllers and methods into their own namespace

gravitystorm:
This PR builds on https://github.com/openstreetmap/openstreetmap-website/pull/2160 and moves the other api controllers into the api namespace. It also splits up "mixed" controllers, where existing controllers have a mixture of website and api methods, into two separate controllers.

This removes a ton of `before_action ... except ...` and `before_action ... only ...` from the controllers, which is a big win for understandability and avoiding errors.

I plan in future work to refactor all the api controllers to inherit from a general ApiController, so that certain things (like before_actions) are common to all methods in all api controllers. However, I avoiding doing so here since the PR is already quite big and complex.


tomhughes:
@Nikerabbit note that this moves some keys (`comment`, `rss` and `entry`) from under `notes` to `notes.api` in the translations.

PR #2164, opened 2019-03-01
move changeset xml generation to view

woodpeck:
This pull request gets rid of the to_xml method in the changeset model, and replaces it with XML builder code in the view which is the canonical way of generating output. This paves the way to parametrising XML output later. Some changes to the tests were necessary to make up for the lost to_xml functionality in the model. The plan is to get rid of other, similar to_xml methods one by one.


tomhughes:
What do you mean by "parameterising" the output?

Is it worth spending time on this if the goal is to get rid of this code altogether and share the cgimap code instead?

woodpeck:
By "parameterising" the output I mean that at a later time I want to make it dependent on who is asking - those who have accepted the ToU will get a more verbose XML and those who haven't will see less. This would be inelegant to do inside the model's to_xml code.

I would love to skip this work and concentrate on cgimap instead in the knowledge that cgimap is going to be used as a module inside the rails code, but I lack the capabilities of making that wish come true and I fear that it might be years away!

gravitystorm:
> Is it worth spending time on this if the goal is to get rid of this code altogether and share the cgimap code instead?

I think it's worth spending the time on this, since as Frederik says it's stuff that can be actioned now.

I also think it's something that we'll do for things I don't expect to ever see in cgimap, like traces and tracepoints and user_preferences, so we might as well be consistent with the other models that currently have the `to_xml` or `to_xml_node` methods.

mmd-osm:
While you're at it, maybe also think about JSON output for a minute, as it would help us push the respective code on cgimap side to production.

gravitystorm:
> While you're at it, maybe also think about JSON output for a minute, as it would help us push the respective code on cgimap side to production.

I see moving the xml generation code to views as the key enabler of json output, since it allows us to switch output formats in the controllers more idiomatically.

tomhughes:
I wasn't necessarily trying to be negative I was just genuinely curious as to the motivation given that this is a significant structural change whose overall consequences are hard to assess.

On the face of it this is indeed clearly the correct approach, to separate the view properly, allowing multiple formats etc.

Performance is obviously a concern and is hard to assess, though for production purposes most of the places where that is a concern are likely already run through cgimap anyway.

My other main concern is having to still keep the `to_xml` methods (but moved to the tests) so that we now have duplication of effort. Then again maybe it's good that we don't have the slightly circular situation where we are using the models to test themselves?

gravitystorm:
> Then again maybe it's good that we don't have the slightly circular situation where we are using the models to test themselves?

Yeah, I don't think the XML payloads sent to the API should be generated from the models, since as you say that becomes a bit circular. Whether the XML test payload should just be inline HEREDOC style ([example](https://github.com/openstreetmap/openstreetmap-website/blob/94b59f440321846760d430fee378f05a6a2803ee/test/controllers/api/changesets_controller_test.rb#L533-L556)) or via a fixture (e.g. [like traces](https://github.com/openstreetmap/openstreetmap-website/blob/master/test/gpx/fixtures/a.gpx)) or via helper methods (like this PR) - I'm not quite sure which is most preferable.

gravitystorm:
There are 6 rubocop warnings that need to be fixed before this can be merged.

gravitystorm:
I've pushed some fixes to this PR, and I think it's ready to merge.

PR #2168, opened 2019-03-06
Use ERB Lint to improve erb templates

gravitystorm:
[ERB Lint](https://github.com/Shopify/erb-lint) brings a useful amount of linting to our html.erb templates. It also allows us to run rubocop on the code embedded in the templates, which is useful too.

As a proof of concept, I've disabled most of the erb linters and many of the rubocop linters too, and just made whitespace changes. These are trivial but were also drowning out everything else, so it's nice to fix them first.

This adds erblint to the travis config, and updates the contributing guide to refer to it too.

The intention is to work through everything marked TODO in the config file, and fix those issues later. The rubocop-related configurations not marked as TODO are those that the ERB Lint docs suggest disabling permanently.

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -1,0 +1,55 @@
+---
+linters:
+  SelfClosingTag:
+    enabled: false
+  SpaceInHtmlTag:
+    enabled: false # TODO
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      inherit_from:
+        - .rubocop.yml
+      Layout/InitialIndentation:
+        Enabled: false
+      Layout/TrailingBlankLines:
+        Enabled: false
+      Layout/TrailingWhitespace:
+        Enabled: false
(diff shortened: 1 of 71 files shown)


gravitystorm:
Ah, there's a problem with erb-lint. It says it supports ruby 2.3, but uses a method (`String.match?`) that's only available in 2.4+. Let me investigate.

tomhughes:
We're on 2.5 anyway, and rails 6 is going to require 2.5 so I say we just up our requirement?

gravitystorm:
It's [fixed upstream](https://github.com/Shopify/erb-lint), just not in a released version of that gem yet.

I'd be happy to bump the required ruby version. It means that we drop support for 16.04, but ruby 2.5 is in 18.04 as far as I can tell - and as you say, it'll soon be required for rails 6 anyway.

Issue #2169, opened 2019-03-06
Reduce papercuts when changing configuration settings


gravitystorm:
From time to time we change the required configuration settings that are expected to be found in `config/application.yml`, for example by adding a new key. This can cause trivial-but-annoying errors for developers. In particular, the updated application might run fine until you hit one particular page that requires the new key. This isn't picked up on in the tests, since they use the `config.application.example.yml` file, nor on application boot. When you hit the problem, the error is a bit cryptic too.

I've investigated using off-the-shelf configuration management gems, with an eye to replacing our [custom preinitializer-and-constants](https://github.com/openstreetmap/openstreetmap-website/blob/bd79c197cc8919e566f4563aafe1f01d9cfc94bc/config/preinitializer.rb) based approach. Criteria include:

* Support for configuration settings from yaml files
* Support for overriding settings using environment variables
* Support for our custom all-in-one existing yaml approach
* Support for non-string settings (like integers and floats)
* Validation of settings (e.g. flagging up required values at boot, ensuring certain settings are numerical)
* Allow settings to be missing (e.g. the oauth_keys, where you need to boot the app to register for consumer tokens)
* Is the gem maintained etc

I've had previous experience with [Figaro](https://github.com/laserlemon/figaro), but it treats all settings as strings, and doesn't have any built-in conversions or validation. The solution for validation and conversion could then be to use an initializer to populate the [Rails custom settings](https://guides.rubyonrails.org/configuring.html#custom-configuration). Figaro also lacks any obvious way to convert our existing application.yml into the expected yaml format.

I've been exploring the [Config](https://github.com/railsconfig/config/) gem today, and I can get it mostly working, but there's a few rough edges in there too. I'll open a PR shortly.

I'd also like to hear if anyone else has thoughts on different approaches to this problem, or different criteria that should be considered, or things that I've overlooked.

PR #2170, opened 2019-03-06
Use Config gem to read and validate configuration settings

gravitystorm:
This is a partial solution to #2169 using the [Config](https://github.com/railsconfig/config/) gem. It features:

* validating particular settings, and what format those settings should be, at boot time
* supports our existing config file structure (through some hackery), including special settings for tests
* works with existing application.yml files
* supports environment variable overrides (but see below)

However:

* It needs to work in two stages, using initializers/settings.rb to revisit the setup with our custom application.yml file. This isn't ideal, since running everything in initializers/config.rb would mean settings are processed at the [before_configuration](https://github.com/railsconfig/config/blob/3a28b646a69c16f87d6539b4e1c4189aecb34c84/lib/config/integrations/rails/railtie.rb#L21) stage, whereas settings.rb runs later.
* It still requires our preinitializer, in order that the constants required in `application.rb` are available. `application.rb` runs before the code in `initializers/settings.rb`
* Although it supports environment variables, it can't support the same naming convention as we currently use. For example, we currently support e.g. 'OSM_API_TIMEOUT=foo'. If we set the prefix to 'OSM' and the env_separator to '_', then that parses to 'Settings.api.timeout' instead of 'Settings.api_timeout'. I'm really not sure if this is a big deal or not.
* There's a corner case involving environment variables that look-like-numbers-but-aren't. For example, with `env OSM.api_version=0.7 bundle exec ...`, the 0.7 gets converted to a float and then fails the string validation.

If we can ignore environment variables (can we?) then this still is an improvement on what we have already, due to the validations. However, it could be even better if there are other refactorings that can remove the need for the preinitializer code entirely, and I welcome any feedback.


tomhughes:
Having done some testing I think most of the above problems go away if we just bite the bullet and move our settings to `settings.yml` with per-environment overrides as necessary.

Doing that, and removing the preinitiailiser and your `settings.rb` initialiser and with various extra conversions to use `Settings.` it is possible to get to the console and the rest is just a question of more work really.

The only problem (given that the `config.rb` initialise is deliberately loaded very early) is the very first use of `STATUS` in `config/application.rb` which occurs before it is loaded. Maybe we should just say that status is always set from the environment and not use settings for it?

As far as I general environment support is concerned I doubt it is ever used. I had forgotten we supported it myself! So yes we can likely get rid of it.

gravitystorm:
> Having done some testing I think most of the above problems go away if we just bite the bullet and move our settings to `settings.yml` with per-environment overrides as necessary.

I was mulling this over last night, before I read your message, and I came to a similar conclusion. I realised that it's the only sensible approach to using settings in other initializers, which is exactly why the gem loads values so early. With this PR's approach of loading settings in a regular initializer, it would be hard to get it working at all.

> The only problem (given that the `config.rb` initialise is deliberately loaded very early) is the very first use of `STATUS` in `config/application.rb` which occurs before it is loaded. Maybe we should just say that status is always set from the environment and not use settings for it?

If that's workable with you, then great. I had a brief look around to see if there's anyone else with a "database offline" mode in rails, and how they manage it, but my googling was just picking up unrelated scenarios.

> So yes we can likely get rid of it.

Great. I'd like to support environment variables since it's useful when e.g. deploying to Heroku, but it's much easier if we can break compatibility with the existing ones if they are unused. I had no idea that we supported them either, until yesterday.

I'll close this PR, and work on the "bite the bullet" approach as a replacement.

Issue #2146, opened 2019-02-17
Error 500 on showing note with no comments


westnordost:
### Observed behavior
Displaying a note with no comments yields an error 500.

### How to reproduce
Display this note: https://www.openstreetmap.org/note/940980
the note via the API: https://api.openstreetmap.org/api/0.6/notes/940980

### Expected behavior
1. The OSM Api should reject creating a note without an initial comment
2. Either all current notes with no comments are deleted from DB or displaying a note with no comment should not yield an error 500

tomhughes:
I don't think that's true is it. I thought @Firefishy established that the cause was that the creator of the note had been deleted because they were a spammer?

tomhughes:
So the immediate cause might be that there are no (visible) comments, but the conclusion that the API allows creation of a note without a comment is not - the only bug here is the exception when a note has no visible comments.

westnordost:
That might as well be, I can't see whether it is invisible or never existed, I only reported what I saw. Anyway, if the (only) comment was deleted, then error 500 would still not be appropriate.

tomhughes:
Sorry I assumed it was you that had been discussing it with @Firefishy on IRC last night but I now see that it wasn't and in fact that was a different note.

tomhughes:
The cause is essentially the same - the user that opened this note deleted their account causing their note comments to be hidden.

This wasn't a spam note though, which raises the question of whether we should in fact hide comments from deleted users or not, or whether we should just treat them as anonymous,

westnordost:
I guess a good approach would be to treat it the same as map data and GPS traces a deleted user contributed.

Dimitar5555:
This issue is still present.
https://api.openstreetmap.org/api/0.6/notes/2523764
https://www.openstreetmap.org/note/2523764

tomhughes:
That would be why this issue is still open yes. Did have something useful to add?

Dimitar5555:
The note that westnodorst linked seems to no longer results in a HTTP 500 error and currently it has content (i.e. it's not empty).

Discostu36:
Sorry for not finding this issue before opening mine. I think you should hide/delete these notes. There is no use for empty notes leading to error pages, it only makes maintenance harder.

tomhughes:
We already have a plan (#3831) to fix this which we have started work on.

kmpoppe:
[...]

PR #2176, opened 2019-03-13
Raise error when note has no comments

jguthrie100:
Temporary fix for #2146 

- Raises `RecordNotFound` exception when note has no comments (avoiding 500 Error)

### DIFF
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -84,6 +84,8 @@ class BrowseController < ApplicationController
     else
       @note = Note.visible.find(params[:id])
       @note_comments = @note.comments
+
+      raise ActiveRecord::RecordNotFound if @note.comments.empty?
     end
   rescue ActiveRecord::RecordNotFound
     render :action => "not_found", :status => :not_found


tomhughes:
This definitely needs a test but more significantly I'm not sure this is really the right fix. I have a feeling that the `visible` scope on notes should exclude these but that turns out to be non-trivial to do...

Still thinking about the best way to deal with it really.

jguthrie100:
The problem is that `@note_comments` is nil, but in the view, a request is made for `@note_comments.first`, which then causes a crash.

So either the view needs to be rewritten to handle a case where there are no comments, or the view needs to not be called in the first place..

jguthrie100:
The line that crashes is:

https://github.com/openstreetmap/openstreetmap-website/blob/874fddf49945443e5af47812fa3a30e96375ce9c/app/views/browse/note.html.erb#L10-L13

jguthrie100:
Unless a new error is raised: `ActiveRecord::RecordDeleted`, which opens the comments page, but notes the comments that have been deleted

jguthrie100:
> This definitely needs a test but more significantly I'm not sure this is really the right fix. I have a feeling that the `visible` scope on notes should exclude these but that turns out to be non-trivial to do...

What do you mean by "exclude these"? What should be excluded?

mmd-osm:
#3608 supersedes this pull request.

PR #2177, opened 2019-03-13
Use Config gem to read and validate configuration settings

### LINKED ISSUE #2169: Reduce papercuts when changing configuration settings

gravitystorm:
Fixes #2169. See also #2170 for an earlier attempt.

This PR replaces our application.yml based settings management with the [Config](https://github.com/railsconfig/config/) gem.

* Default settings are now in `config/settings.yml`
* You can override these settings by creating `config/settings.local.yml` with the particular settings you need to override.
* We can therefore add new settings in the future and they will just work
* Some settings are now validated. More could be validated in the future.
* The STATUS constant is now only read from the environment, since we need to know it before loading rails and before the configuration settings are read. You can run e.g. `STATUS=database_offline bundle exec rails server`
* The application will not start if it finds `config/application.yml`. You'll receive an helpful error message explaining what to do.

I've included this guard in case developers aren't aware of these changes, to avoid them running the updated code and not realising that their custom settings in application.yml are being completely ignored. If you want to have a zero-downtime change to the new way of configuring, then:
* Before deploying this code, configure your deployment environment to create `config/settings.local.yml` - this only needs the settings you want to override. It will be ignored by older codebases.
* Also, configure your deployment environment to remove `config/application.yml` if the default settings file `config/settings.yml` exists - it will appear after this PR is merged.
* Then you should be fine to deploy versions of the app both before and after this change as you see fit.


gravitystorm:
Oh, I thought the settings.local.yml was optional, but it fails when that's missing. I'll need to update this PR.

PR #2181, opened 2019-03-17
Add banner against EU Copyright Directive

Nakaner:
See my [email](https://lists.openstreetmap.org/pipermail/osmf-talk/2019-March/006027.html) on OSMF-Talk mailing list for details.

This pull request is based on #1979.

Set end date to 2019-03-28 because that's the end of the sittings period March II. The draft aganda says that the topic will be tabled on 2019-03-26. If this becomes final, I will change the end date to 26 March.


tomhughes:
It's premature to open this at the same time as seeking approval - this is not something we from a technical point of view would be able to accept without OSMF approval.

Issue #2132, opened 2019-02-06
Animations for traces


gravitystorm:
Traces can now be imported without requiring the external gpx-import daemon. However, the animation in the `GPX::File.picture` method haven't yet been implemented. This issue is to track progress on this topic and make sure we don't forget to fix this!

We are currently waiting on [dark-panda/gd2-ffij#18](https://github.com/dark-panda/gd2-ffij/issues/18).

mmd-osm:
I believe one thing that could be done in the meantime is to render one distinct image for each frame (=10 different images), which later get passed on to gdImageGifAnimBegin, gdImageGifAnimAdd and gdImageGifAnimEnd to eventually produce the animated gif.

https://github.com/openstreetmap/gpx-import/blob/master/src/image.c#L72-L105

mmd-osm:
By the way, you can use `gem "gd2-ffij", :git => 'git@github.com:mmd-osm/gd2-ffij.git', :branch => 'animated_gif'` to play around with the prototype.

I still get segmentation faults when I'm try to put some kind of loop around the image creation, and I have absolutely no idea why. Somehow Ruby does really strange stuff under the hood. The following creates a gif file with 2 images, though. Maybe there's some GC kicking in, or scoping rules destroy parts of the FFI stuff when iterating in a loop?!

```
    def picture(min_lat, min_lon, max_lat, max_lon, _num_points)
      frames = 10
      width = 250
      height = 250
      proj = OSM::Mercator.new(min_lat, min_lon, max_lat, max_lon, width, height)

      image = []

      image[0] = GD2::Image::IndexedColor.new(width, height) 
      black = image[0].palette.allocate(GD2::Color[0, 0, 0])
      white = image[0].palette.allocate(GD2::Color[255, 255, 255])
      grey = image[0].palette.allocate(GD2::Color[187, 187, 187])

      image[0].draw do |pen|
        pen.color = grey
        pen.rectangle(0, 0, width, height, true)
      end

      image[1] = GD2::Image::IndexedColor.new(width, height)
      black = image[1].palette.allocate(GD2::Color[0, 0, 0])
      white = image[1].palette.allocate(GD2::Color[255, 255, 255])
      grey = image[1].palette.allocate(GD2::Color[187, 187, 187])

      image[1].draw do |pen|
        pen.color = white
        pen.rectangle(0, 0, width, height, true)
      end

      image[1].draw do |pen|
        pen.color = black
[...]

PR #2190, opened 2019-03-27
Enable trace jobs by default

gravitystorm:
I think it's a good idea to use the trace jobs queue by default now, even while we wait on #2132.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -60,9 +60,9 @@ spam_threshold: 50
 # Default legale (jurisdiction location) for contributor terms
 default_legale: GB
 # Use the built-in jobs queue for importing traces
-# Leave as false if you are using the external high-speed gpx importer
+# Set to false if you are using the external high-speed gpx importer
 # https://github.com/openstreetmap/gpx-import
-trace_use_job_queue: false
+trace_use_job_queue: true
 # Location of GPX traces and images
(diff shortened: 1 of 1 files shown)


tomhughes:
Animations is the only reason I hadn't gone ahead and thrown the switch...

I did do all the preparatory work on the chef side to be able to stand up a separate queue runner for this on machines that can see the GPX store.

tomhughes:
I realised that this doesn't actually affect the production site because chef is forcing that to false anyway so we might we well merge this now.

I'm thinking I may switch the production site in the next few days as part of decommissioning spike-{01,02,03} and replacing them with spike-{06,07,08} given that spike-01 is currently running the import daemon.

gravitystorm:
> I realised that this doesn't actually affect the production site because chef is forcing that to false anyway so we might we well merge this now.

Yep, I wasn't intending to flick the switch for production yet, just to change the default for new deployments, and in particular change the default so that gpx importing works by default for developers.

I would like to see the animations fixed, but there's no progress on #2132 yet and I'm not sure what the way forward is on that one.

mmd-osm:
> I would like to see the animations fixed, but there's no progress on #2132 yet and I'm not sure what the way forward is on that one.

No worries, I think we can get this thing working:

new_foo

PR #2191, opened 2019-03-27
Refactor notes#mine

gravitystorm:
This is some minor refactoring of a view that I noticed. I felt it was more complicated than necessary.


tomhughes:
@Nikerabbit note that this renames `notes.mine.subheading` to `notes.mine.subheading_html`.

PR #2196, opened 2019-04-01
Replace jshint with eslint

tomhughes:
This PR attempts to replace jshint with eslint - as a first pass it uses a config that tries to replicate the jshint options we had enabled although eslint does seem to have a a few more things that I have cleaned up in a separate commit.

Once this is in we can think about enabling more checks - probably starting with some of those that iD is using.


gravitystorm:
Looks good to me, everything worked when I tested it locally but I'm not the world's biggest javascript expert. I have used eslint in other projects (though nothing fancy), and it's a good idea to align with iD too.

PR #2207, opened 2019-04-12
Move gpx_file (trace) xml generation out of model and into view

woodpeck:
This continues the work from #2164 for the "trace" class.

### DIFF
--- a/app/controllers/api/traces_controller.rb
+++ b/app/controllers/api/traces_controller.rb
@@ -16,14 +16,9 @@ module Api
     around_action :api_call_handle_error
 
     def show
-      trace = Trace.visible.find(params[:id])
+      @trace = Trace.visible.find(params[:id])
 
-      if trace.public? || trace.user == current_user
-        @traces = [trace]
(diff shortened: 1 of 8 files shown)


tomhughes:
This all lost fine for the most part but it does create a slightly confusing naming situation.

We wind up with a `trace` view that actually takes an argument named `@traces` and renders multiple traces so that it can also support a separate action on the user controller.

One option would be just to rename it to `traces` so that things are consistent, but that means we have a view that doesn't directly relate to any action. The only other approach I can see is to have separate top level views though they could still share the common partial where most of the work is done.

gravitystorm:
> The only other approach I can see is to have separate top level views though they could still share the common partial where most of the work is done.

That would be my preferred approach. Each action (show, data, index, etc) should have a correspondingly named view, even if that view is little more than including the partial. This keeps everything where a new developer would expect to find it.

gravitystorm:
I've added another commit to this PR based on Tom's feedback and my own, and I think it's good to go now.

PR #2209, opened 2019-04-14
Replace eslint-rails-ee with direct calls to eslint

tomhughes:
Sadly while `eslint-rails-ee` is better than the original `eslint-rails` it turns out that the version of `eslint` it is bundling is still ancient and I'd really like to use something newer.

So this PR tries a new approach, installing eslint directly with yarn and then using a simple rake task to invoke it.

I chose to use yarn rather than npm simply because that is the preferred tool for managing javascript dependencies in rails although it does make it a bit more of a pain for people to setup.


lib/tasks/eslint.rake
```diff
@@ -0,0 +1,27 @@
...
+  def yarn_path
+    Rails.root.join("bin", "yarn").to_s
+  end
+
+  def config_file
+    Rails.root.join("config", "eslint.json").to_s
+  end
+
+  def js_files
+    require File.dirname(__FILE__) + "/../../config/environment"
```
gravitystorm: Instead of manually requiring the environment, rake tasks can be marked as depending on the rails environment e.g.

`task :check => :environment do`

.travis.yml
```diff
@@ -26,8 +26,9 @@ before_script:
   - touch config/settings.local.yml
   - bundle exec rake db:migrate
   - bundle exec rake i18n:js:export
+  - bundle exec rake yarn:install
 script:
   - bundle exec rubocop -f fuubar
-  - bundle exec rake eslint:run_all
+  - bundle exec rake eslint
```
gravitystorm: We should add this to CONTRIBUTING.md in the Coding Style section. I mean, I expect a change like this to have a corresponding change to the documentation, but the documentation mentioning eslint is missing :smile:

PR #2210, opened 2019-04-15
changes reply to this entry to Contact the author

### LINKED ISSUE #2166: The meaning of "Reply to this entry" is not clear
Its not really clear what "reply to this entry" does. It sounds like its replying to diary entry, but its sending the person a private message. So it would be clearer if it if said something like "Write user message" or something similar that explicitly says your sending them a message. Instead of just leaving one on the diary entry like "comment on this entry" does. 

Also, the usage of "reply" is kind of ambiguous in this situation because the communication hasn't actually been initiated yet. So your not really "replying" to anything in the traditional sense.

ConSou:
attempts to resolve #2166

### DIFF
--- a/config/locales/en-GB.yml
+++ b/config/locales/en-GB.yml
@@ -340,7 +340,7 @@ en-GB:
     diary_entry:
       posted_by: Posted by %{link_user} on %{created} in %{language_link}
       comment_link: Comment on this entry
-      reply_link: Reply to this entry
+      reply_link: Contact the author
       comment_count:
         zero: No comments
         one: '%{count} comment'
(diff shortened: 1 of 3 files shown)


gravitystorm:
Thanks for the pull request @ConSou! Unfortunately I can't merge this yet, for a couple of reasons:

* This PR contains changes to multiple translation files. You only need to change `en.yml`, the other files are managed by [Translatewiki](https://translatewiki.net/wiki/Translating:OpenStreetMap)
* I've updated the original issue with the terminology I'd prefer to see ('Send a message to the author') since 'contact' is a bit vague
* There's also another place where this should be fixed, in the diary comment notification emails.

If you'd like to rework your pull request then that would be really helpful.

ConSou:
Hello @gravitystorm, I appreciate the feedback, I can make the changes you mentioned!

gravitystorm:
Closing in favour of #2710 which is a version of this PR based on my original review.

PR #2221, opened 2019-05-05
[WIP] XML generation: nodes, ways, relations, map controller → view; JSON output

mmd-osm:
As the issue title suggests, this is still very much WIP with the intention to gather some early feedback, go through a number of iterations to fix issues, etc.

Topics in this PR:

(1) Move the XML generation from the controller to the view for the following controllers:

* nodes_controller
* ways_controller
* relations_controller
* old_controller
* map_controller

(2) Adding JSON output according to the Overpass API JSON format spec for nodes, ways, relations, and the map call.

TODO:

- [ ] Remove XML generation methods in models
- [x] Finish off remaining methods in all 5 controllers
- [ ] Add "redaction" flag in XML response, check if anything else if missing
- [ ] Test cases
- [ ] A good approach to keep using changeset_cache, user_display_name_cache

Follow up for https://github.com/openstreetmap/iD/issues/3765

### DIFF
--- a/app/controllers/api/map_controller.rb
+++ b/app/controllers/api/map_controller.rb
@@ -35,10 +35,8 @@ module Api
         return
       end
 
-      doc = OSM::API.new.get_xml_doc
-
       # add bounds
-      doc.root << bbox.add_bounds_to(XML::Node.new("bounds"))
+      @bounds = bbox
 
       # get ways
       # find which ways are needed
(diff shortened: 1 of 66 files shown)


mmd-osm:
Short update: this pull request now covers all aspects of CGImap in terms of JSON output capabilities. Also, all test cases pass again with XML output. Due to lack of knowledge, I had to temporarily remove the user/changeset caching.

I think that would be a great opportunity for a first code review, and then continue working on the rough edges.

mmd-osm:
Ok, I think this PR is too difficult to review all in one step.

I created #2223 now with XML related changes only, and once we're clear on that one, we can continue with the JSON part in a new PR afterwards. Let's do one step after the other...

PR #2223, opened 2019-05-07
XML generation: nodes, ways, relations, map controller → view

mmd-osm:
As the issue title suggests, this is still very much WIP with the intention to gather some early feedback, go through a number of iterations to fix issues, etc.

Topic in this PR:

Move the XML generation from the controller to the view for the following controllers:

- nodes_controller
- ways_controller
- relations_controller
- old_controller
- map_controller

This change is a prerequisite for both GDPR and JSON output.

Maybe obsolete:

- [ ] Add "redaction" flag in XML response, check if anything else if missing  --> not available in CGImap either
- [x] ~A good approach to keep using changeset_cache, user_display_name_cache~


mmd-osm:
Ping @gravitystorm : could you please take a look if this is going in the right direction. I want to push JSON output for Rails port, so we can enable in CGImap for everyone to use. This pull request is a prerequisite for this change, as is for our planned GDPR changes @woodpeck was suggesting earlier on.

gravitystorm:
Sorry for not providing feedback more quickly. I've had a look at this, and here are my suggestions:

* Avoid the topic of default rendering formats in this PR. It creates a breaking API change in some edge cases (as per #2065) so best dealt with separately. This means your controllers shouldn't contain `respond_to` blocks and the tests then won't need changes.

* Don't wrap objects in arrays unnecessarily. For example `@nodes = [node]`. The relevant view should just handle an `@node` object instead.

* Try to avoid creating arrays for no reason. For example, after refactoring your code becomes:

```
      @elems = []

      visible_elements.each do |element|
        @elems << element
      end
```
... which isn't doing anything useful!

* If you're not sure how to handle the caches for users etc (and to be honest, I don't know either) then just leave these methods and come back to them in a separate PR. I'd prefer to see one PR that covers all the straightforward refactoring and is easy to review, and a second PR that covers tough things but is nice and short.

* I think there's need for `.empty?` guards when using `.each` (in  app/views/api/map/_node.xml.builder)

* Avoid using `@variables` in partials. The partials are easier to reuse/understand if you pass in the objects at the point they are rendered. For example, you have `@bounds` in the _bounds.builder partial, but it would be better to call it from the index.xml.builder with e.g. `  osm << (render(:partial => "bounds", :object => @bounds) || "")` and then just use `bounds` in the bounds partial.

* Check out https://github.com/openstreetmap/openstreetmap-website/pull/2278 since it applies to many of the partial renderings that you have in this PR.

* Do you really need `|| ""` in all these render calls? I suspect (but haven't tested) that most calls will work without them.

mmd-osm:
>  This means your controllers shouldn't contain respond_to blocks and the tests then won't need changes.

Please keep in mind that there's a seperate pull request coming for JSON support. How would you handle an additional format without respond_to blocks?

gravitystorm:
> Please keep in mind that there's a seperate pull request coming for JSON support. How would you handle an additional format without respond_to blocks?

Let's deal with that separately, since it affects all routes, all API tests, and not just the routes that you're refactoring in this PR.

gravitystorm:
> Let's deal with that separately, since it affects all routes, all API tests, and not just the routes that you're refactoring in this PR.

I mean, I think it's going to involve adding `respond_to` blocks, and likely going to cause chaos around handling requests with missing or confusing Accepts headers, and headaches around writing tests without tons of repetition, etc etc - but if we're going make those kind of changes they should be made for the whole API at the same time, not for only a few methods in this PR. So that's why I suggest removing the respond_to sections from this particular PR and dealing with that topic separately.

mmd-osm:
Just a side note: do you know if we will have Ruby for Good folks are improving the world again for OSM this year? If so, it would probably make sense to delegate this to the "pros" (rather than me fiddling with this).

gravitystorm:
> Just a side note: do you know if we will have Ruby for Good folks are improving the world again for OSM this year? If so, it would probably make sense to delegate this to the "pros" (rather than me fiddling with this).

I haven't heard anything about Ruby for Good this year. But it's well worth you working on this in any case, you know plenty about the subject matter and we're all talented amateurs here :smiley:

mmd-osm:
> I think there's need for .empty? guards when using .each (in app/views/api/map/_node.xml.builder)

I didn't quite get this comment. Currently the code reads as:

```ruby
if node.tags.empty?                          <<<< check is already here
  xml.node(attrs)
else
  xml.node(attrs) do |nd|
    node.tags.each do |k, v|
      nd.tag(:k => k, :v => v)
    end
  end
end
```
So this should already be covered. I didn't spot any other usage of "each" in this partial.

mmd-osm:
> Do you really need || "" in all these render calls? I suspect (but haven't tested) that most calls will work without them.

If you take `/map` as an example, the call might actually fail with "_API threw unexpected ActionView::Template::Error exception: no implicit conversion of nil into String_", in case there are only nodes and ways, but no relations in the bbox, and you don't supply ` || "" ` as a fallback.

Pretty much the same applies to /nodes, /ways, and /relations, which could return some empty result. HIstory and version calls might not return anything in case of redacted object versions.

mmd-osm:
> Avoid using @variables in partials

Good point, fixed in https://github.com/openstreetmap/openstreetmap-website/pull/2223/commits/45e7c9a32e08446fd087580d84b2e11eadc1683d

mmd-osm:
> Check out #2278 since it applies to many of the partial renderings that you have in this PR.

Not 100% sure, if my understanding is right, so let's try `/map/index.xml.builder`:

```
xml.instruct!

  osm << (render(:partial => "bounds", :object => @bounds) || "")
  osm << (render(@nodes) || "")
  osm << (render(@ways) || "")
  osm << (render(@relations) || "")
end
```

Triggers:

`API threw unexpected ActionView::Template::Error exception: Missing partial api/nodes/_node with {:locale=>[:en], :formats=>[:xml], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jsonify, :coffee]}. `

I believe this kind of optimization doesn't work due to the way I'm keeping all node, way and relation partials under the umbrella of the /map call.

#2278 works in that other pull request, as all partials are located in the directory named after the model, so the direct lookup is possible. Somehow I'm working against that convention in my coding...

I think he only way to make this work would be to move /api/map/_{node,way,relation}.xml.builder to /api/{nodes,ways,relations} respectively.

gravitystorm:
> I think there's need for .empty? guards when using

Sorry, that should have read 'I think there's **no** need for' the guards.



> The reason why this is outside of `xml.node(attrs) do |nd|` is that it fails horribly otherwise - either producing invalid XML or simply crashing, I don't recall at the moment.

Oh, then perhaps there is. That's a pity since I would have thought the `each` simply doesn't do anything when there's no tags.

tomhughes:
> > The reason why this is outside of `xml.node(attrs) do |nd|` is that it fails horribly otherwise - either producing invalid XML or simply crashing, I don't recall at the moment.
> 
> Oh, then perhaps there is. That's a pity since I would have thought the `each` simply doesn't do anything when there's no tags.

I just tried removing the `empty?` special case and it seems to work fine for me for nodes with no tags - it would odd if it didn't given that ways and relations aren't special cased like that and they presumably work.

mmd-osm:
IIRC some consuming app weren't happy, otherwise I wouldn't have added this special case. Need to check later, what exactly the issue was.

tomhughes:
> IIRC some consuming app weren't happy, otherwise I wouldn't have added this special case. Need to check later, what exactly the issue was.

Ah it does cause you to get an element with no content (`<node...></code>`) rather than an empty element (`<node/>`) but any client that can't cope with that is just buggy!

mmd-osm:
> Ah it does cause you to get an element with no content (<node...></node>) rather than an empty element (<node.../>) but any client that can't cope with that is just buggy!

Ah, I was wrong. The idea was really to have the response in line to what we have today, meaning, this extra logic was meant to re-create the exact same response, and avoid the extra `</node>`. This also happens to save a few bytes as well. Client should be able to handle either version.

mmd-osm:
> If you're not sure how to handle the caches for users etc (and to be honest, I don't know either) then just leave these methods and come back to them in a separate PR. I'd prefer to see one PR that covers all the straightforward refactoring and is easy to review, and a second PR that covers tough things but is nice and short.

I'm seeing lots of **CACHE** entries on the Rails console. Does that mean that the builder framework is clever enough to do its own caching? If so, I would just remove the already commented `changeset_cache` and `user_display_cache` initialization, and we're done with the topic.

```
  CACHE Changeset Load (0.0ms)  SELECT  "changesets".* FROM "changesets" WHERE "changesets"."id" = $1 LIMIT $2  [["id", 1479], ["LIMIT", 1]]
  ↳ app/views/api/ways/_way.xml.builder:7
  CACHE User Load (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ↳ app/views/api/ways/_way.xml.builder:7
  Node Load (0.6ms)  SELECT "current_nodes".* FROM "current_nodes" INNER JOIN "current_way_nodes" ON "current_nodes"."id" = "current_way_nodes"."node_id" WHERE "current_way_nodes"."way_id" = $1 ORDER BY "current_way_nodes"."sequence_id" ASC  [["way_id", 4000260693]]
  ↳ app/views/api/ways/_way.xml.builder:12
  CACHE Changeset Load (0.0ms)  SELECT  "changesets".* FROM "changesets" WHERE "changesets"."id" = $1 LIMIT $2  [["id", 1481], ["LIMIT", 1]]
  ↳ app/views/api/ways/_way.xml.builder:7
  CACHE User Load (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ↳ app/views/api/ways/_way.xml.builder:7
  Node Load (0.4ms)  SELECT "current_nodes".* FROM "current_nodes" INNER JOIN "current_way_nodes" ON "current_nodes"."id" = "current_way_nodes"."node_id" WHERE "current_way_nodes"."way_id" = $1 ORDER BY "current_way_nodes"."sequence_id" ASC  [["way_id", 4000281620]]
  ↳ app/views/api/ways/_way.xml.builder:12
  CACHE Changeset Load (0.0ms)  SELECT  "changesets".* FROM "changesets" WHERE "changesets"."id" = $1 LIMIT $2  [["id", 1480], ["LIMIT", 1]]
  ↳ app/views/api/ways/_way.xml.builder:7

```

mmd-osm:
> Don't wrap objects in arrays unnecessarily. 

This issue has been fixed in https://github.com/openstreetmap/openstreetmap-website/pull/2223/commits/cf3cf5216fa3697ad2e3c060fced6a7bbb16f4a4

app/controllers/api/map_controller.rb
```diff
@@ -92,13 +91,16 @@ def index
 
       # this "uniq" may be slightly inefficient; it may be better to first collect and output
       # all node-related relations, then find the *not yet covered* way-related ones etc.
+      @relations = []
```
gravitystorm: `@relations = relations.uniq`
 
Or perhaps rename the `relations` variable right from the start to `@relations`, and then `@relations.uniq!` at this point

app/controllers/api/nodes_controller.rb
```diff
@@ -31,7 +31,12 @@ def show
       response.last_modified = node.timestamp
 
       if node.visible
-        render :xml => node.to_xml.to_s
+        @node = node
```
gravitystorm: Perhaps `@node` throughout

gravitystorm: There's a bunch of other places like this too - generally if you find yourself needing to rename `variable` to `@variable` it probably should have been `@variable` from the start. But it's not critical, we can always refactor this later.

mmd-osm: Right, I also found this somewhat awkward... makes total sense.

mmd-osm:
(travis ci fails due to https://github.com/openstreetmap/openstreetmap-website/commit/c1c277186ef0c519bed5b07e1b67ab8681010f15)

mmd-osm:
> Avoid the topic of default rendering formats in this PR

Done in  https://github.com/openstreetmap/openstreetmap-website/pull/2223/commits/2bfb976b8e42f6ab0e87c053e91b2bc88538dba6 https://github.com/openstreetmap/openstreetmap-website/pull/2223/commits/fb953006a1a8fe2949f376e167dda963dfd784bd  https://github.com/openstreetmap/openstreetmap-website/pull/2223/commits/84faaaf7950e61b5f87019398aeda0ead8d41c2c

gravitystorm:
Great, this now looks good to me.

I'd like @tomhughes just to give it a quick glance, specifically to check that

```diff
- render :xml => doc.to_s
+ render :formats => [:xml]
```
are equivalent and don't have any side-effects around content-types etc that I might have missed.

tomhughes:
I haven't managed to find any documentation for `:formats` but we have other places (eg https://github.com/openstreetmap/openstreetmap-website/blob/master/app/views/traces/georss.rss.builder#L28) where we use it to force rendering in a specific format so I assume it's fine.

app/views/api/map/index.xml.builder
```diff
@@ -0,0 +1,8 @@
+xml.instruct!
+
+xml.osm(OSM::API.new.xml_root_attributes) do |osm|
+  osm << (render(:partial => "bounds", :object => @bounds) || "")
+  osm << (render(@nodes) || "")
+  osm << (render(@ways) || "")
+  osm << (render(@relations) || "")
```
tomhughes: Why do we need all the `|| ""` stuff here? What does adding a string to the XML even do?

mmd-osm: That’s a logical or operator. If render returns nil, an empty string is added to the result as a fallback. See https://github.com/openstreetmap/openstreetmap-website/pull/2223#issuecomment-506019631

mmd-osm: Documentation https://guides.rubyonrails.org/layouts_and_rendering.html says:

> In the event that the collection is empty, render will return nil, so it should be fairly simple to provide alternative content.
> 
> ```
> <h1>Products</h1><%= render(@products) || "There are no products available." %>
> ```

tomhughes:
Sure I understand what `||` is and what that expression is doing and how you would use it in an ERB template - what I'd like to understand is what it does in builder though.

I did try to look last night but the builder doco wasn't working - will try again.

mmd-osm:
Ok, that's what I thought. I stumbled upon the "ActionView::Template::Error exception: no implicit conversion of nil into String" error message, and found this as a workable solution somewhere. Unfortunately, I don't recall where that was.

gravitystorm:
> Unfortunately, I don't recall where that was.

https://github.com/openstreetmap/openstreetmap-website/commit/925bd812a2c9000c4901384b70f20e65b0f53cd4

PR #2225, opened 2019-05-15
Allow user_blocks to be created for longer periods

gravitystorm:
This avoids admins from having to manually change end periods in the database.

Most of this diff is the helper to display durations in the dropdown menu in more friendly terms than e.g. "8766 hours". The hours are carefully picked to show up as rounded durations. It's possible to pick particular hours that don't fit the pattern, e.g. 36 hours, but handling those situations was becoming an exercise in over-engineering.

Screenshot from 2019-05-15 11-44-47

 I also wanted to preserve the "0 hours" label since that's how those blocks are referred to in common parlance, and to avoid e.g. "about 1 month" and similar vagueness, so that precluded using the standard rails time_to_words helpers.

### DIFF
--- a/app/helpers/user_blocks_helper.rb
+++ b/app/helpers/user_blocks_helper.rb
@@ -22,4 +22,19 @@ module UserBlocksHelper
       I18n.t("user_blocks.helper.time_past", :time => friendly_date(last_time)).html_safe
     end
   end
+
+  def block_duration_in_words(duration)
+    parts = ActiveSupport::Duration.build(duration).parts
+    if duration < 1.day
+      I18n.t("user_blocks.helper.block_duration.hours", :count => parts[:hours])
+    elsif duration < 1.week
+      I18n.t("user_blocks.helper.block_duration.days", :count => parts[:days])
+    elsif duration < 1.month
+      I18n.t("user_blocks.helper.block_duration.weeks", :count => parts[:weeks])
+    elsif duration < 1.year
+      I18n.t("user_blocks.helper.block_duration.months", :count => parts[:months])
+    else
+      I18n.t("user_blocks.helper.block_duration.years", :count => parts[:years])
+    end
+  end
 end
--- a/app/views/user_blocks/edit.html.erb
+++ b/app/views/user_blocks/edit.html.erb
@@ -18,7 +18,7 @@
   </p>
   <p>
     <%= label_tag "user_block_period", t(".period") %><br />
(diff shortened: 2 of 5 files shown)


Zverik:
This is good, although I don't like durations that are too long: what's the difference for an admin between 6 months and a year? Would you consider replacing 6+ months with "Forever" (100k hours)?

Or better, shorten the list by half to "0, 3, 24, 24*3, 24*7, 24*30, forever".

gravitystorm:
To be honest I don't know why there's already a choice of 3 hours vs 6 hours, but I don't hand out blocks and I don't know which are most frequently used. I'm happy to take guidance from DWG if they aren't happy with the list of options.

SomeoneElseOSM:
(to give a bit of info about block usage) the lengths I normally use are:

0 hours (as a "message that needs to be read before continuing to edit").  On a per-human-user basis, by far the most common.

4 days (often used for sockpuppets and "please email the DWG before continuing")

1 day and 2 days - used to enforce "cooling down"periods.

1 hour - very rare - used sometimes to indicate to paid mappers that we do really mean business and yes, they really should do what we have asked them to do.

Going forward I'd expect I'd mostly use any blocks longer than 4 days alongside a message "please email the DWG before continuing" and to interfere with vandalism rather than a "difference of opinion about how to map".  There always needs to be a "way back in" even for people who have been guilty of bad actions in the past.  Currently (when dealing with determined sockpuppets of vandals) its a pain to keep rolling forward 4-day blocks until they can be banned (as an example, all of those at the top of https://www.openstreetmap.org/user/SomeoneElse/blocks_by started out as 4 day blocks).

simonpoole:
Just a reminder on the side given that somebody with more Rails foo than me is looking at the matter:

- if a user is logged in when they get a block there is no way that they will notice or be notified (I had a look at terminating the session, but iirc there is no straightforward way to do that, see "foo"),

-  further they can, for example start iD (or already be editing and receive a block) and to continue to edit as normal, just to get a "the sky is falling" moment when they try to save, see  https://github.com/openstreetmap/iD/issues/5400

gravitystorm:
> Just a reminder on the side given that somebody with more Rails foo than me is looking at the matter:

Thanks, I'd noticed that mentioned in a different thread. At the moment the blocked status is only checked during the authentication step of the API calls. I agree there's better ways to do this and I'll get round to that eventually.

mmd-osm:
There's  /api/0.6/user/details in place to find out if you've been blocked in the meantime. Maybe that's a good first step already (see the iD issue 5400 above for more details).

tomhughes:
I've merged this as is for - if DWG want different lengths then just ask, or send a PR to change them.

PR #2226, opened 2019-05-15
Allow moderators to hide diary entries and comments

gravitystorm:
This arises from discussions with DWG.

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -44,6 +44,7 @@ class Ability
         can [:account, :go_public, :make_friend, :remove_friend], User
 
         if user.moderator?
+          can [:hide, :hidecomment], DiaryEntry
           can [:index, :show, :resolve, :ignore, :reopen], Issue
           can :create, IssueComment
           can [:new, :create, :edit, :update, :destroy], Redaction
@@ -51,7 +52,7 @@ class Ability
         end
 
         if user.administrator?
-          can [:hide, :hidecomment], [DiaryEntry, DiaryComment]
+          can [:hide, :hidecomment], DiaryEntry
           can [:index, :show, :resolve, :ignore, :reopen], Issue
           can :create, IssueComment
           can [:set_status, :delete, :index], User
--- a/app/views/diary_entries/_diary_comment.html.erb
+++ b/app/views/diary_entries/_diary_comment.html.erb
@@ -7,7 +7,7 @@
   </p>
 
   <div class="richtext"><%= diary_comment.body.to_html %></div>
-  <% if current_user && current_user.administrator? %>
+  <% if can? :hidecomment, DiaryEntry %>
     <span>
       <%= link_to t(".hide_link"), hide_diary_comment_path(:display_name => diary_comment.diary_entry.user.display_name, :id => diary_comment.diary_entry.id, :comment => diary_comment.id), :method => :post, :data => { :confirm => t(".confirm") } %>
     </span>
--- a/app/views/diary_entries/_diary_entry.html.erb
+++ b/app/views/diary_entries/_diary_entry.html.erb
@@ -37,7 +37,7 @@
       </li>
     <% end %>
 
-    <% if current_user && current_user.administrator? %>
+    <% if can? :hide, DiaryEntry %>
       <li>
         <%= link_to t(".hide_link"), hide_diary_entry_path(:display_name => diary_entry.user.display_name, :id => diary_entry.id), :method => :post, :data => { :confirm => t(".confirm") } %>
       </li>
--- a/test/abilities/abilities_test.rb
+++ b/test/abilities/abilities_test.rb
@@ -23,7 +23,6 @@ class GuestAbilityTest < AbilityTest
 
     [:create, :edit, :comment, :subscribe, :unsubscribe, :hide, :hidecomment].each do |action|
       assert ability.cannot?(action, DiaryEntry), "should not be able to #{action} DiaryEntries"
-      assert ability.cannot?(action, DiaryComment), "should not be able to #{action} DiaryEntries"
     end
   end
 
@@ -54,7 +53,6 @@ class UserAbilityTest < AbilityTest
 
     [:hide, :hidecomment].each do |action|
       assert ability.cannot?(action, DiaryEntry), "should not be able to #{action} DiaryEntries"
-      assert ability.cannot?(action, DiaryComment), "should not be able to #{action} DiaryEntries"
     end
 
     [:index, :show, :resolve, :ignore, :reopen].each do |action|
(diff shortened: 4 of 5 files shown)


tomhughes:
Well sure, because it gives them new powers ;-)

Are we sure we're happy with granting them that power?

If we're going to do that should reports of diary issues also go to the moderator queue rather than the admin queue?

gravitystorm:
> Are we sure we're happy with granting them that power?

That's not for me to decide, since I'm not in any of the groups [website moderators, DWG members, sysadmins, website administrators]. :-)

But it does seem reasonable to me. They already deal with notes and changeset comments, neither of which are geodata.

> If we're going to do that should reports of diary issues also go to the moderator queue rather than the admin queue?

I thought about that, but I don't have visibility into the current reports and who would be best to deal with the majority of them. For example, I don't know whether the reports are mostly spam (admins) or behaviour complaints (moderators) but I guess spam.

This PR just gives moderators abilities, which allows the relevant issues to be divided up based on context. The default assigned_role could be changed, or we could do more complex things like assign based on report category.

tomhughes:
My main concern is how we ensure a consistent standard with regard to what is and is not considered acceptable if we suddenly grant this power to a lot more people.

As I said in an email the other day I tend to adopt a very conservative attitude and block those things which are obviously inappropriate but many of the reporters are looking for a much harder line and I worry that we might wind up with very different standards being applied depending on who is looking at the issue.

woodpeck:
I think that the admins should not be burdened with the decision of what is inappropriate and what isn't. If the admins were more than just @tomhughes then they would face the same problem (setting consistent standards) among themselves; the only reason this is a no-brainer today is that only on person does it, and that, at the same time, is also undesirable (because you have said countless times that you abhor having to make decisions like that). Having the group of moderators - essentially, the DWG - do this can certainly lead to an uneven implementation at first but a standard will emerge eventually. If someone is unhappy with what the DWG does they can *still* seek reprieve from the OSMF board to overrule DWG.

gravitystorm:
@tomhughes Anything that I can do to move this along? I agree with @woodpeck's point of view, and also more people taking care of bits of normal spam here and there can help us focus on the bigger issues.

tomhughes:
So I was thinking about this over the weekend and I think there's actually a bigger problem here.

Currently when people spam the diary we don't just remove the diary entry, we close the account to stop them being able to do it again. The "hide" option for diary entries and comments is actually used fairly rarely, mostly for things which I think are wildly off topic or inappropriate in some way but which are from established users.

So if we go ahead with this and moderators start just hiding the diary entries it might actually make things worse because the users will just be able to repost and it will be next to impossible for people like me with the administrative role to do anything about it.

gravitystorm:
> So if we go ahead with this

Remember that this is aimed mostly at letting the moderators manage community behaviour. Help with spam is just a side effect and not the main purpose of this PR.

> it will be next to impossible for people like me with the administrative role to do anything about it.

I don't understand this. How does a moderator hiding a spam post prevent an admin from later closing the account? And how would leaving administrators with exactly the same permissions prevent them from doing something that they do now? If it's a workflow issue (where leaving spam available for hours makes it easier for administrators to block the account), then let me know more details about this.

tomhughes:
It prevents it because I will have no idea that the spam ever existed, so no idea that I need to close the account.

Even if somebody was to report the user I would have to go and dig into the database with SQL to verify the report before I could take any action and that is totally unscalable.

woodpeck:
I don't fully understand the problem but if it is a real issue then it could be addressed this in one of two ways: (1) by process: simply ask the moderators to use the "hide" feature only on accounts that have edits and/or non-spam diary entries and thus are not a candidate for outright account deletion; or (2) by code: only offer a "hide" option when outright account deletion has become unlikely because the account has edits.

smsm1:
Another option could be for administrators to be able to see the hidden spam posts so that they can make the decision to close the account. Having the number of spam items on an account could be useful to moderators, should be a statistic shown to them, possibly with dates of when they posted the recently hidden spam, or maybe even shown the whole spam items.

tomhughes:
That is certainly an option, but personally I'm not keen of having tens of thousands of chinese spam posts cluttering up my diary view just because I'm me ;-)

SomeoneElseOSM:
> Another option could be for administrators to be able to see the hidden spam posts

(for the benefit of those not familiar) this is exactly how "hidden" status works now for moderators with e.g. hidden changeset comments - moderators can see the content, along with "hidden by X" below.

smsm1:
> That is certainly an option, but personally I'm not keen of having tens of thousands of chinese spam posts cluttering up my diary view just because I'm me ;-)

The default view would be as now, then an additional view or filter would add the extra items with the hidden state.

tomhughes:
Yes that could work.

My real point here is that we need to consider the entire workflow and how we expect reports to be handled and ensure that the tools exist to allow that to happen in the easiest way possible without getting in the way of normal use of OSM.

gravitystorm:
@tomhughes Have you had the chance to develop/consider the workflow you desire? I'm happy to code whatever you need. 

But otherwise, I would prefer to proceed with merging this as-is.

tomhughes:
Sorry that was more of a general comment about how the report stuff is more work for the moderator and admins than it needs to be.

I'm not sure there is any workflow which really solves my concerns here. Really my concerns here can't be solved technically, only by agreements and then getting humans to stick those agreements.

I guess we might as well merge it. What's a bit more pain for me.

I mean yes Shaun's idea of a way to opt in to viewing hidden diary posts is certainly useful, but unless I'm going to be expected to proactively go and search for hidden diary posts and consider whether the user needs to be blocked it doesn't actually help.

I honestly don't know what the answer is - my model for what the moderation tools should look like is stackoverflow where the report is presented to the moderator with as much relevant information as possible in a single view and there are simple action buttons. I have no idea how SO goes about things like translating a moderator deciding a post is spam into actually blocking accounts though - maybe they just do it automatically if another posts are confirmed as spam?

gravitystorm:
> unless I'm going to be expected to proactively go and search for hidden diary posts and consider whether the user needs to be blocked it doesn't actually help.

So I expect the moderators to be using these abilities in two situations

* To deal with community stuff, e.g. someone being deliberately offensive. Nothing needs to be done by the administrators in these situations.
* Helping out with spam problems. In this case, if the account also needs closing, we could ask the moderators to report the user after hiding the spam. That will put an issue in the admins list, right?

Would that reduce the work (or at least, not increase the work) for the admins?

tomhughes:
That's probably about the best we can do yes, though I do really need an easy way to be able to see the user's diary posts or comments with deleted entries included.

gravitystorm:
So #2248 was merged last week. What, if anything, still blocks this?

tomhughes:
Probably nothing from a policy point of view, beyond ensuring that DWG understand what is happening and what is expected.

I haven't done a technical review - one question I guess is whether the change from #2248 should apply to moderators as well.

gravitystorm:
>  one question I guess is whether the change from #2248 should apply to moderators as well.

I think if they ask for it we can add it later, but for now I don't think it's worth adding.

SomeoneElseOSM:
>> one question I guess is whether the change from #2248 should apply to moderators as well.

> I think if they ask for it we can add it later, but for now I don't think it's worth adding.

Some way of viewing hidden diary entries would make sense, even if only to "unhide" after clicking a button in error.

PR #2229, opened 2019-05-19
Add static code checks in Travis CI

mmd-osm:
This pull request adds Brakeman as Static Analysis Security Tool, ~as well as bundle-audit to check for vulnerable gems~.

Documentation:

- Brakeman: https://brakemanscanner.org/docs/
- ~bundle-audit: https://www.rubydoc.info/gems/bundler-audit/frames~

Some ideas taken from: https://rietta.com/blog/2017/10/03/automate-security-scans-with-continuous-integration/

### DIFF
--- a/.brakeman.ignore.json
+++ b/.brakeman.ignore.json
@@ -1,0 +1,1529 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Redirect",
+      "warning_code": 18,
+      "fingerprint": "06e17db22db1f94acfd9a66ed0960c933a0534e75a143e2835b9b61064fdb97e",
+      "check_name": "Redirect",
+      "message": "Possible unprotected redirect",
+      "file": "app/controllers/messages_controller.rb",
+      "line": 122,
+      "link": "https://brakemanscanner.org/docs/warning_types/redirect/",
+      "code": "redirect_to(params[:referer])",
+      "render_path": null,
+      "location": {
+        "type": "method",
+        "class": "MessagesController",
+        "method": "destroy"
+      },
+      "user_input": "params[:referer]",
+      "confidence": "High",
+      "note": ""
+    },
+    {
+      "warning_type": "File Access",
+      "warning_code": 16,
+      "fingerprint": "12c528c1d324860a0d5c574a620ecc4be31bd4d947e4dac959cc57022266b10d",
+      "check_name": "FileAccess",
+      "message": "Model attribute used in file name",
+      "file": "app/controllers/api/traces_controller.rb",
+      "line": 142,
+      "link": "https://brakemanscanner.org/docs/warning_types/file_access/",
+      "code": "FileUtils.rm_f(Trace.new(:name => file.original_filename.gsub(/[^a-zA-Z0-9.]/, \"_\"), :tagstring => tags, :description => description, :visibility => visibility, :inserted => true, :user => current_user, :timestamp => Time.now.getutc).trace_name)",
+      "render_path": null,
+      "location": {
+        "type": "method",
+        "class": "Api::TracesController",
+        "method": "do_create"
+      },
+      "user_input": "Trace.new(:name => file.original_filename.gsub(/[^a-zA-Z0-9.]/, \"_\"), :tagstring => tags, :description => description, :visibility => visibility, :inserted => true, :user => current_user, :timestamp => Time.now.getutc).trace_name",
+      "confidence": "Medium",
+      "note": ""
+    },
+    {
+      "warning_type": "Cross-Site Scripting",
+      "warning_code": 2,
+      "fingerprint": "1fab7706b9ed0587e8b87ede43ec6f9aa9bab1b3bcfb83be1c2f6efd68b37c98",
+      "check_name": "CrossSiteScripting",
+      "message": "Unescaped model attribute",
+      "file": "app/views/browse/_relation_member.html.erb",
+      "line": 8,
+      "link": "https://brakemanscanner.org/docs/warning_types/cross_site_scripting",
+      "code": "t(\".entry_role\", :type => t((\".type.\" + Relation.new.member_type.downcase)), :name => link_to(printable_name(Relation.new.member), { :action => Relation.new.member_type.downcase, :id => Relation.new.member_id.to_s }, :title => link_title(Relation.new.member), :rel => link_follow(Relation.new.member)), :role => h(Relation.new.member_role))",
+      "render_path": [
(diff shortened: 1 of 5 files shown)


tomhughes:
We already get alerts from github automatically for vulnerabilities and I don't see how something run by travis helps because that only runs on activity - checking for vulnerabilities needs to happen all the time.

mmd-osm:
Did you take a look at the Brakeman results for this pull request? This isn’t part of what Github reports, as it only covers gems with known issues.

tomhughes:
So we can't possibly take this with dozens of outstanding failures - at the very least you would need to "fix" those first before we can consider this.

That said, there is a reason I put fix in quotes, because many of it's complaints look pretty dodgy to me. I suspect there is a high false positive rate here...

tomhughes:
My first comment was directed at the auditing - the second addresses my initial thoughts on the brakeman reports.

mmd-osm:
Seems I was a bit too quick to respond. In general I agree that many findings might be questionable, still I managed to use one of the reported unprotected redirects in ways which were probably not intended. This might be a bit of work initially, but hopefully that’s  useful to catch some issues in the future before they hit production.

tomhughes:
Well maybe, but just because you can make it do something unintended doesn't mean there is a security issue. I'm sure there are lots of way to make it do silly things if you feed it unintended output but if those things are harmless then I'm not really worried - the only away to avoid that would be to add vast amount of detailed validation of inputs.

mmd-osm:
As an example for the redirect, someone might post an innocently looking osm.org link which in fact takes an unsuspecting user to a malware site without further ado. Those are the kind of issues I’d like to see fixed.

mmd-osm:
I changed the command line settings for brakeman to avoid non-zero rc for warnings. This way, we can already run the checks as part of the build, and work on the issues as time permits.

gravitystorm:
I've had a look through the output, and while sure some are false positives, it's picking up a lot of dodgy coding practices that we've otherwise overlooked. Particularly the SQL injection stuff, where some small changes to the code would reduce the opportunity for mishaps later.

So I'm in principle happy to add Brakeman, the question is the best way to do it. If we add it with the return code disabled (as now), then we haven't achieved much since we could add new vulnerabilities without triggering a build failure. So I'd prefer to approach it the way we do with rubocop and erblint, that is to create a configuration that ignores existing problems, but will alert on new problems. We can then work our way through the todo list.

From looking at [the brakeman docs](https://brakemanscanner.org/docs/options/), we can create a config file and ignore the failing tests for now.

For this PR, it would also be best to just focus on the installation and configuration of brakeman. The regexp fixes would be valid in a standalone PR.

gravitystorm:
Oh. Licensing.

[Brakeman is proprietary software](https://github.com/presidentbeef/brakeman/blob/master/LICENSE.md). It appears from reading the licence (IANAL etc) that we'd be covered by the free-beer usage, and any conceivable use in our context would be fine (e.g. developer being paid to contribute to OpenStreetMap, or this codebase being used in a commercial context). But it's still something to consider.

An open-source fork named "[Railroader](https://github.com/david-a-wheeler/railroader)" exists, forked from the last MIT licensed version of Brakeman. The Railroader project is currently 4 months old.

See https://github.com/presidentbeef/brakeman/issues/1312 and https://github.com/david-a-wheeler/railroader/issues/2 for more context.

So would we be happy using Brakeman?

tomhughes:
I agree that I think there's nothing in the licence that would stop us using it - the issue come down to whether we want to do so, or if the OSMF policy will allow us to, assuming we consider ourselves bound by that.

The fork appears to have activity at least, although it's fairly new as you say so hard to know how it will pan out long term.

woodpeck:
The OSMF policy would certainly not *forbid* using proprietary software, it would however ask everyone to consider whether a suitable open source equivalent is available, and to regularly repeat this assessment. The willingness of users to use something "perhaps not yet quite as good as the free-beer commercial offering" is what has made OSM into the success it is today. It behooves us to show the same willingness towards other open projects but of course there are limits.

gravitystorm:
@mmd-osm what are your thoughts on:

* Using railroader instead of brakeman
* Setting up a config file to ignore existing problems, but fail the travis build on new problems (e.g. similar to `.rubocop-todo.yml`)

gravitystorm:
@mmd-osm just checking in on this - what are your thoughts for/against my previous suggestions?

mmd-osm:
I don't have a strong opinion on the first point, as we're only using this tool part part of the CI chain, and it doesn't introduce any new runtime dependencies. I'm also using tools like coverity for cgimap to check for issues in c++ code, even though they're not open source.

Regarding the second point, I haven't really seen anything in the config settings like a rubocop style exception definition to disable checks on file/method level. This seems a bit unfortunate as we would skip some checks globally, even for newly introduced issues. Maybe I'm missing something here in the documentation and more fine grained control is possible somehow.

gravitystorm:
OK, lets stick with Brakeman. I've also figured out how to ignore the current issues using a config file, so I believe this is now ready to merge. 

(Sorry for the force-pushing, I was aiming to preserve the original commit while also resolving conflicts and adding the new config file to this existing PR.)

tomhughes:
Is there no better way of suppressing warnings than that monstrosity of a JSON file that lists each one in excruciating detail?

I mean ideally we want to suppress whole classes of warning until we decide what we want to do about that rather than suppressing specific examples, which means that new code will still trigger complaints about things that we may well not be interested in fixing.

In other words we want something more like the way rubocop works...

gravitystorm:
> Is there no better way of suppressing warnings than that monstrosity of a JSON file that lists each one in excruciating detail?

I've investigated the alternative options. Brakeman only supports one config file at a time - so even though it reads `./config/brakeman.yml` by default, if you pass another one on the command line, the original file is ignored. This makes it impossible to work the way we have with rubocop, where we have separate files for TODO and 'permanent' ignores. But the config file is yml, which supports comments, so we can split it into two sections using comments if required.

In 6e62531b68ff0ff62eb0a944f0c6d35054b13c27 I removed the ignore json file, and ignored all the checks that are currently causing warnings, but marked them as todo using a comment. Again, it's not as flexible as rubocop, so there's no way to disable a class of warning just for specific files, so this approach runs the risk of introducing similar bugs in new commits while the check is globally disabled.

What do you think? I'm happy with either option.

gravitystorm:
Updated and merged as #2723

Thanks @mmd-osm for this original PR and for bringing Brakeman to our attention in the first place.

PR #2236, opened 2019-05-29
Update annotations to correctly show bigint types

gravitystorm:
In https://github.com/ctran/annotate_models/pull/515 the annotation gem started to distinguish `bigint` columns from `integer` columns. This PR updates the relevant annotations.


gravitystorm:
I've added a second commit with updates to the index annotations arising out of https://github.com/ctran/annotate_models/pull/482

PR #2237, opened 2019-05-29
Use strong_migrations 

gravitystorm:
… to help developers avoid problems in production database schema changes.

I've used this gem in other projects and I find it really useful. For many developers it's not clear which migrations can be done near-instantly (e.g. adding a new column) and which migrations will lock a table, potentially for hours (e.g. adding a new column with a default value).

This gem highlights bad migrations during development, so that the developers can learn what the problems might be before the sysadmins find out in production. I've set it to ignore all existing migrations since some of them won't have been safe, but that no longer matters.

### DIFF
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,0 +1,1 @@
+StrongMigrations.start_after = 20190518115041 # rubocop:disable Style/NumericLiterals
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,7 @@ gem "rack-cors"
 gem "rails-i18n", "~> 4.0.0"
 gem "record_tag_helper"
 gem "rinku", ">= 2.0.6", :require => "rails_rinku"
+gem "strong_migrations"
 gem "validates_email_format_of", ">= 1.5.1"
 
 # Native OSM extensions
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -405,6 +405,8 @@ GEM
       actionpack (>= 4.0)
       activesupport (>= 4.0)
       sprockets (>= 3.0.0)
+    strong_migrations (0.3.1)
+      activerecord (>= 3.2.0)
     term-ansicolor (1.7.1)
       tins (~> 1.0)
     terrapin (0.6.0)
@@ -510,6 +512,7 @@ DEPENDENCIES
   sanitize
   sassc-rails
   secure_headers
+  strong_migrations
   therubyracer
   uglifier (>= 1.3.0)
   validates_email_format_of (>= 1.5.1)


tomhughes:
Even adding a new column can be a problem for a table like `users` because the contention on it is so high and any sort of change to table requires a lock on the table at least for a few seconds...

tomhughes:
I will need to look at this in more detail but it strikes me that this is a big grab bag of stuff, much of it well intentioned, but some of it quite worrying as well in terms of what it is suggesting.

Do we at least know if it has the sense to recognise when you are doing things (like creating an index) in the same migration that creates a table and not complain about them and suggest you do crazy things in those cases?

gravitystorm:
> Do we at least know if it has the sense to recognise when you are doing things (like creating an index) in the same migration that creates a table and not complain about them

Maybe the test suite is useful if you want to review this? Probably clearer than reading the code.

https://github.com/ankane/strong_migrations/blob/11391b500c62e8aa0bb8660c45be5c2f49ca4be0/test/strong_migrations_test.rb#L252

I haven't found any craziness in my other projects. If there ever was a conflict of opinions, then the 'safety_assured' wrapper could be used.

gravitystorm:
Interestingly, due to my fat-fingering, it already picked up that the most recent migration (adding indexes to acls) could have been improved by creating them concurrently, instead of locking the acls table for the duration of the index build. Probably not a big deal for that table, but it would have been an easy fix, and would come in more useful for the larger tables too.

https://github.com/openstreetmap/openstreetmap-website/blob/master/db/migrate/20190518115041_add_acl_indexes.rb

tomhughes:
The table thing is good at least and I certainly got the impression that the author know what they're doing when it comes to progres.

On the other hand there's all those suggestions to disable transactions which like makes me mega nervous.

config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1 @@
+StrongMigrations.start_after = "20190518115041".to_i
```
tomhughes: Why do this as a string and then convert it to integer? Why not just write it as an integer?

gravitystorm: It’s a hacky way to avoid rubocop complaining about underscores in large integers  🙂 perhaps I should be more sensible and use a magic rubocop disable-comment instead.

tomhughes: Oh I thought I had turned that off because I kind of hate it myself ;-)

gravitystorm: Oh, I'm fine with the underscores for actual numbers (like 50_000), but this string of digits isn't really representing 20 trillion, 190 billion, 518 million and so on somethings, so the underscores are meaningless here - e.g. 20_190_518_115_041. 

Anyway, I've used a rubocop disable comment instead, which is probably the best approach.

tomhughes: Oh I was going to just merge it as is, just didn't get around to it last night. I'm not a great fan of disable comments personally.

PR #2239, opened 2019-05-29
Allow a configurable delay for diary entry feeds

gravitystorm:
* This is to provide another tool to help spam fighting. It is, of course, not a panacea.
* The configuration value is set to zero hours in order to be 'least surprising' for anyone else deploying the code.
* You can change the setting without interfering with your test suite.

If this is merged, then it will be up to the sysadmins what value they want to set for osm.org. I would suggest 6 hours as a reasonable compromise between unnecessary delays and allowing enough time to allow admins (and/or moderators, see #2226) to take action. Of course, it's a setting that can be adjusted, and can be set back to 0 hours in the future.

### DIFF
--- a/app/controllers/diary_entries_controller.rb
+++ b/app/controllers/diary_entries_controller.rb
@@ -166,6 +166,10 @@ class DiaryEntriesController < ApplicationController
     else
       @entries = DiaryEntry.joins(:user).where(:users => { :status => %w[active confirmed] })
 
+      # Items can't be flagged as deleted in the RSS format.
+      # For the general feeds, allow a delay before publishing, to help spam fighting
+      @entries = @entries.where("created_at < :time", :time => Settings.diary_feed_delay.hours.ago)
+
       if params[:language]
         @entries = @entries.where(:language_code => params[:language])
         @title = t("diary_entries.feed.language.title", :language_name => Language.find(params[:language]).english_name)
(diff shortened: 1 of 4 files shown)


gravitystorm:
>  it will be up to the sysadmins what value they want to set for osm.org. I would suggest 6 hours

I just wanted to highlight this bit again, since it's still set to zero (i.e. immediately publish to the feeds) on the OSMF servers.

tomhughes:
I know it is, because by the time we got this in the major problem had passed, so it's a tool we are holding in reserve for now, especially since we now have @harry-wood's fix for the aggregator in place.

harry-wood:
There's a small caveat to my fix, as I just mentioned here: https://github.com/gravitystorm/blogs.osm.org/issues/17#issuecomment-503933732  We do of course still retain spam in the time window of 30 minute between updates. This means spam which is zapped within a minute on the diaries website by our vigilant moderators, might linger for 29 minutes on blogs.openstreetmap.org  . 

Another factor is _other_ users of the diaries RSS feed. This delay trick would help keep their feed clean.

...of course delaying the feed is also a compromise in use-ability. Feeds are better when they're up to-to-date ... somewhat ....not sure how important it is.

gravitystorm:
> might linger for 29 minutes on blogs.openstreetmap.org .

Yep, and then my [feed reader](http://newsblur.com/) pulls them in from the aggregator's feed, and I still see the spam since other feed consumers don't have any understanding of deleting items. So removing them from blogs.openstreetmap.org is great, and worthwhile, since it makes that site usable. But they still leak onwards in the feeds.

Basically, anything that happens after the spam reaches the first feed is, while worthwhile, not going to stop me from seeing the spam. So it's imperative to do what we can to avoid the spam getting into the feeds in the first place.

PR #2248, opened 2019-06-05
Show deleted diaries and diary comments to admins

gravitystorm:
In https://github.com/openstreetmap/openstreetmap-website/pull/2226 @tomhuges wrote:

> I do really need an easy way to be able to see the user's diary posts or comments with deleted entries included. 

So here we go! This PR shows deleted diary entries to administrators, so long as the user hasn't also been deleted. It also shows deleted diary comments to administrators. It shows these deleted things deemphasized, and allows admins to restore them if desired.

There's a small difference in functionality between entries and comments, because comments are always shown (i.e. not filtered by user status). However, that was more complexity and we're already getting into some deep hypothetical weeds here.

### DIFF
--- a/test/controllers/diary_entries_controller_test.rb
+++ b/test/controllers/diary_entries_controller_test.rb
@@ -94,6 +94,10 @@ class DiaryEntriesControllerTest < ActionController::TestCase
       { :controller => "diary_entries", :action => "hidecomment", :display_name => "username", :id => "1", :comment => "2" }
     )
     assert_routing(
+      { :path => "/user/username/diary/1/unhidecomment/2", :method => :post },
+      { :controller => "diary_entries", :action => "unhidecomment", :display_name => "username", :id => "1", :comment => "2" }
+    )
+    assert_routing(
       { :path => "/user/username/diary/1/subscribe", :method => :post },
       { :controller => "diary_entries", :action => "subscribe", :display_name => "username", :id => "1" }
     )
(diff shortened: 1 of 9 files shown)


test/controllers/diary_entries_controller_test.rb
```diff
@@ -85,6 +85,10 @@ def test_routes
       { :path => "/user/username/diary/1/hide", :method => :post },
       { :controller => "diary_entries", :action => "hide", :display_name => "username", :id => "1" }
     )
+    assert_routing(
+      { :path => "/user/username/diary/1/unhide", :method => :post },
+      { :controller => "diary_entries", :action => "unhide", :display_name => "username", :id => "1" }
+    )
```
tomhughes: Shouldn't there be a routing test for `unhidecomment` as well?

test/system/diary_entry_test.rb
```diff
@@ -15,4 +15,50 @@ def setup
...
+  test "deleted diary comments should be hidden for regular users" do
+    @deleted_comment = create(:diary_comment, :diary_entry => @diary_entry, :visible => false)
+
+    sign_in_as(create(:user))
+    visit diary_entry_path(@diary_entry.user, @diary_entry)
+
+    assert_not page.has_content? @deleted_comment.body
+  end
+
+  test "deleted diary comments should be show to administrators" do
```
tomhughes: I assume `show` should be `shown` here...

test/controllers/diary_entries_controller_test.rb
gravitystorm: Good catch

gravitystorm:
I've fixed both reviews and rebased since it clashed with #2250

PR #2257, opened 2019-06-12
Adjust submit labels to use built-in i18n handling

gravitystorm:
I noticed this while working on the diary entries, so now I'm rolling it out more widely. In many cases we've been avoiding using the built-in translation keys for submit buttons, and hard-coding translations based on the file the form is in (rather than based on the model for the form). This has made some form refactoring harder than it needs to be, as well as just being a bit of a waste of time. Sometimes this has been done just because of copying other places where it was done first.

There's ample opportunity to improve these labels in the future - too many saves and not enough updates in my opinion! But I'll leave the strings mostly unchanged for now.

@nikerabbit this PR contains a few translation key renames:

* `diary_entries.show.save_button` -> `helpers.submit.diary_comment.create`
* `messages.new.send_button` -> `helpers.submit.message.create`
* `oauth_clients.new.submit` -> `helpers.submit.client_application.create`
* `oauth_clients.edit.submit` -> `helpers.submit.client_application.update`
* `redactions.new.submit` -> `helpers.submit.redaction.create`
* `redactions.edit.submit` -> `helpers.submit.redaction.update`
* `traces.new.upload_button` -> `helpers.submit.trace.create`
* `traces.edit.save_button` -> `helpers.submit.trace.update`
* `user_blocks.new.submit` -> `helpers.submit.user_block.create`
* `user_blocks.edit.submit` -> `helpers.submit.user_block.update`

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -18,4 +18,10 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
       .to_return(:status => 404)
     super(*args)
   end
+
+  # Phantomjs can pick up browser Accept-Language preferences from your desktop environment.
+  # We don't want this to happen during the tests!
+  setup do
+    page.driver.add_headers("Accept-Language" => "en")
+  end
 end
--- a/app/views/diary_entries/show.html.erb
+++ b/app/views/diary_entries/show.html.erb
@@ -19,9 +19,9 @@
 
     <%= error_messages_for "diary_comment" %>
 
-    <%= form_for :diary_comment, :url => { :action => "comment" } do |f| %>
+    <%= form_for @entry.comments.new, :url => { :action => "comment" } do |f| %>
       <%= richtext_area :diary_comment, :body, :cols => 80, :rows => 15 %>
-      <%= submit_tag t(".save_button") %>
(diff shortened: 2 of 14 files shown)


test/application_system_test_case.rb
```diff
@@ -18,4 +18,10 @@ def initialize(*args)
       .to_return(:status => 404)
     super(*args)
   end
+
+  # Phantomjs can pick up browser Accept-Language preferences from your desktop environment.
+  # We don't want this to happen during the tests!
+  setup do
+    page.driver.add_headers("Accept-Language" => "en")
+  end
```
tomhughes: As much as we may want this I think it at least needs to be in a separate commit... I assume it got added accidentally with something else?

gravitystorm: It was deliberately in this commit, since the change to the issue comments submit button exposed the problem with the language handling in the system tests. So I thought it was useful to put them both in the same commit to fully illustrate why I was doing it and what sort of test it affects - but I can break it into two commits if you prefer.

tomhughes: Ah ok. I hadn't actually even worked out which commit it came from, I was just surprised to see it.

PR #2266, opened 2019-06-17
Add user_<uid> class to diary entry div

systemed:
This adds a class containing the user ID to the div element enclosing a diary entry.

So, for example, a diary entry I write would get the class `user_165`.

This enables a site viewer to add basic killfile-like functionality with Firefox's usercontent.css or extensions like Stylus.

### DIFF
--- a/app/views/diary_entries/_diary_entry.html.erb
+++ b/app/views/diary_entries/_diary_entry.html.erb
@@ -1,4 +1,4 @@
-<div class='diary_post<%= " deemphasize" unless diary_entry.visible %>'>
+<div class='diary_post<%= " deemphasize" unless diary_entry.visible %> user_<%= diary_entry.user.id %>'>
   <div class='post_heading clearfix'>
     <% if !@user %>
       <%= user_thumbnail diary_entry.user %>


tomhughes:
Actually adding per-user kill file rules might be reasonable, but adding something to the CSS that is not used by the site but exists just to support external kill files seems like a bad idea.

If only because some day somebody will notice it is "unused" and remove it...

Nakaner:
I agree with @tomhughes. Some day people will have forgotten this feature. What about using a [data attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) instead?

tomhughes:
I'm not sure that makes the analysis any different...

systemed:
A proper killfile would be great, but obviously a lot of engineering effort. I think this is a low-impact fix for a fairly niche case. If it breaks it breaks and I promise to re-add it when it does!

mmd-osm:
uBlock Origin:

```
##div.diary_post:if( .clearfix.post_heading > .deemphasize > [href="/user/alexkemp"])
```

Good bye, my friend!

gravitystorm:
@systemed Does the suggestion by @mmd-osm work for the tools you are using?

systemed:
It's a great suggestion. Unfortunately I don't use uBO (the Safari port has been abandoned and was in any case slower than Safari's native content-blockers).

gravitystorm:
I've ummed and aahhhed about this for weeks, since I both object to it ([for the same reasons as @tomhughes](https://github.com/openstreetmap/openstreetmap-website/pull/2266#issuecomment-502678401)) and also support it since it's pretty trivial and [nobody's going to mind if we break it in future](https://github.com/openstreetmap/openstreetmap-website/pull/2266#issuecomment-502684786). But it's not worth me considering it for any longer.

So it just comes down to having a slight bias for change. But if this gets stripped out again in a future refactoring, so be it.

Issue #2255, opened 2019-06-12
Better translations for time_ago_in_words


gravitystorm:
(Moving https://trac.openstreetmap.org/ticket/4968 here for better visibility.)

In rails there is a [`time_ago_in_words`](https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-time_ago_in_words) helper (aka `distance_of_time_in_words_to_now`) that we use extensively. The strings it returns are actually just the same as the [`distance_of_time_in_words`](https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-distance_of_time_in_words) helper, e.g. "3 years" or "about 1 hour"; that is, they doesn't come with 'ago' attached.

The [logic behind the helper](https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/actionview/lib/action_view/helpers/date_helper.rb#L95) is fairly complex. The [default translations come from the rails-i18n](https://github.com/svenfuchs/rails-i18n/blob/86942a876dfe5c53782dcc1b117acb946099c674/rails/locale/en.yml#L64) gem, we don't control them ourselves.

We then use these "distance_in_words" strings to create what I'll refer to as "distance_in_words_ago" strings. In English this is very easy, since every possible output of the helper can just have "ago" tacked on the end. For example

* English: "3 years" -> "3 years **ago**"

However, this is not the case in other languages, where to be gramatically correct, in certain cases the original "distance_in_words" needs tweaking. For example:

* Hungarian: "2 óra" -> "2 ór**ája**"
* German: "2 Tage" -> "**vor** 2 Tage**n**"

Other people identified the problem many years ago, and so since rails 3.2 the `time_ago_in_words` helper accepts a `:scope`, which allows you to pick a different i18n scope for the localisation. For example, instead of using the default `en.datetime.distance_in_words` from the rails-i18n gem, we could define our own `distance_in_words_ago` hierarchy:

```
en:
  datetime:
    distance_in_words_ago:
      about_x_hours:
        one: about 1 hour ago
        other: about %{count} hours ago
      about_x_months:
        one: about 1 month ago
        ...
```

We would then update our views, so instead of using the output from the helper in a '.ago' translation, we would set the i18n scope in the helper, and use the output directly. For example:

```diff
       <% end %>
-      ... <%= t ".ago", :time_in_words_ago => time_ago_in_words(trace.timestamp) %></span>
[...]

PR #2271, opened 2019-06-19
Use a translation scope for time_ago_in_words

gravitystorm:
This represents step 3 in the plan at #2255. I've marked it as WIP since we need to wait for #2270 to be merged, and give the translators some time to make their updates, but it's actually a finished PR. I thought it would be useful to have here now for review since I've done the work to make sure it's all going to work out as intended, even if it's not for immediate merging.

This PR updates our use of `distance_of_time_in_words_to_now` and `time_ago_in_words` rails helpers to use a translation scope. It also introduces a `friendly_date_ago` helper which should be used in place of the `friendly_date` helper for these 'ago' translations.


gravitystorm:
The translation strings are now available in 18 other languages in master, so I'm happy to see this reviewed and merged now.

PR #1290, opened 2016-09-15
Enable the app to be built using Docker

paulsturgess:
Hi, I saw your request for help on this app and I thought maybe getting it working via Docker might make local dev more appealing.

This is a work in progress PR, not ready for merging just yet. I thought it would be good to open it to get some early feedback.

TODO:
- [ ] Investigate why some specs are failing
- [ ] Test the app manually

PR #2272, opened 2019-06-19
Add docker support

fazlerabbi37:
Docker is one of the most popular container platform that makes it easier to build and deploy application. This pull request will add docker support for development setup of openstreetmap-website. 

This PR is build on top of an old abandoned PR (#1290) and follows the [Installation](https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md) guide provided with this project. 

Please keep in mind that I am just starting out with Docker and feel free to suggest and discuss anything that comes to mind. I will try my best to answer them. :smile: 

Todo:
- [x] make docker file for the rails app
- [x] make docker file for the PostgreSQL database
- [x] make a docker-compose file to tie up both containers
- [x] fix deployment when docker files are on docker directory
- [x] fix `rake test:db` error
- [x] populating the database
- [x] documentation 
- [ ] configure an unprivileged user


docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ruby:2.5-slim
+
+#install packages
+#fixes dpkg man page softlink error while installing postgresql-client [source: https://github.com/debuerreotype/debuerreotype/issues/10#issuecomment-438342078]
+RUN mkdir -p /usr/share/man/man1 && mkdir -p /usr/share/man/man7
+RUN apt-get update && apt-get install curl -y
```
pnorman: Each time apt-get is run you should update and then remove the APT cache files at the end of the command.

docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN curl -sL https://deb.nodesource.com/setup_10.x -o nodesource_setup.sh && bash nodesource_setup.sh && rm nodesource_setup.sh
+RUN apt-get install -y --no-install-recommends ruby-dev libarchive-dev libmagickwand-dev libxml2-dev libxslt1-dev build-essential libpq-dev libsasl2-dev imagemagick libffi-dev locales postgresql-client-9.6 nodejs && apt-get clean && rm -rf /var/lib/apt/lists/*
```
pnorman: The `rm -rf /var/lib/apt/lists/*` doesn't reclaim the space since the space is still used by earlier layers of the dockerfile.

docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN npm install yarn -g
```
pnorman: Docs state `bundle exec rake yarn:install`. If there's a reason to differ, can you add a comment?

docker/Dockerfile
tomhughes: Why?

docker/Dockerfile
tomhughes: This all looks horribly wrong anyway - why is is creating directories by hand?

And downloading and installing node from upstream using a dodgy "run this shell script as root" install method isn't going to fly. Why not just use the ubuntu packaged one?

docker/Dockerfile
tomhughes: Different thing - you need to install yarn first (unfortunately it's not packaged in Ubuntu like it is in Fedora) and then `bundle exec rake yarn:install` runs yarn to install the modules listed in `package.json`.

docker/Dockerfile
pnorman: So that the cache files don't make it into the docker image

tomhughes: Isn't this just building a container for local use though? Not anything that's going to be published?

pnorman: Sure, but what I've read indicates it as a best practice in general

mmd-osm:
Worthwhile taking a look at: https://github.com/developmentseed/osm-seed

docker/Dockerfile
fazlerabbi37: > This all looks horribly wrong anyway - why is is creating directories by hand?

as I have mentioned on the comment above the command, while installing `postgresql-client`, because `update-alternatives` can't create symbolic link, `dpkg` throws an error and exits with code 100. A quick search pointed to [this](https://stackoverflow.com/a/52655008/5350059) stack answer which solves the problem. 

> And downloading and installing node from upstream using a dodgy "run this shell script as root" install method isn't going to fly. Why not just use the ubuntu packaged one?

the image `ruby:2.5-slim` usages  Debian 9 (stretch) as base image. Though I could install `nodejs` using the office Debian repo , I couldn't install `npm` using the `apt-get` command as it gives `E: Unable to locate package npm` error. I followed the official download [instruction](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions-enterprise-linux-fedora-and-snap-packages) which leads to NodeSource's GitHub [instruction](https://github.com/nodesource/distributions/blob/master/README.md#debinstall) and installed `nodejs` and `npm`. 

If there is any better way to solve this, I would be happy to hear. :blush:

gravitystorm: > If there is any better way to solve this, I would be happy to hear. blush

I'm also not a Docker expert, but it might be worth considering starting from a different base image, like Ubuntu 18.04, rather than starting from a ruby base image. This would more closely match the installations instructions, and also the Vagrant build.

gravitystorm:
> This pull request will add docker support for development setup of openstreetmap-website.

While I'm happy to add Docker support, I feel like we're ending up with a lot of duplication in installation instructions. We have the [human-readable INSTALL.md instructions](https://github.com/openstreetmap/openstreetmap-website/blob/31edc7dff6f4305366d0f93d35ba3fb1956bb71f/INSTALL.md), the [travis config](https://github.com/openstreetmap/openstreetmap-website/blob/master/.travis.yml), the [Vagrant provisioning scripts](https://github.com/openstreetmap/openstreetmap-website/blob/master/script/vagrant/setup/provision.sh) and now a Dockerfile - all of these cover similar ground. 

Is there any way that we can simplify things, or at least reduce some of the duplication? It's becoming increasingly hard to make sure these are all in line when we make changes.

docker/Dockerfile
fazlerabbi37: > I'm also not a Docker expert, but it might be worth considering starting from a different base image, like Ubuntu 18.04, rather than starting from a ruby base image. 

Neither am I. As far as I understand the purpose of using a ruby image it to remove the legwork required to install ruby on an Ubuntu base image. Maybe someone with better docker understanding can explain it better? ping @pnorman 

I am building on top of the previous PR (#1290) and at first I wanted to see if I can build the `web` and `db` images using docker with the new code because the code base and the installation procedure has changed since that PR. If we all agree on the pros, we can change the base image to Ubuntu 18.04.

Firefishy:
Please give me some time to review this over the weekend. I have a very good understanding of docker.

fazlerabbi37:
So it would seem I have hit a wall. :disappointed: I can build the application and run database migrations and nodejs package installation with `yarn` when the docker files are on the root directory but the database test command fails giving the following error:

```
The HashDiff constant used by this gem conflicts with another gem of a similar name.  As of version 1.0 the HashDiff constant will be completely removed and replaced by Hashdiff.  For more information see https://github.com/liufengyun/hashdiff/issues/45.
rake aborted!
PG::ConnectionBad: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

However, when the docker files are moved to docker directory the `docker-compose build` command fails with `docker_web_1 exited with code 10` error and I can't find any reference of the error code. :no_mouth: 

I have changed the `context` when I moved the docker file to docker directory. If anyone can shade some light on this, would appreciate it. :smile:

gravitystorm:
> database test command fails giving the following error:

This is probably because the default settings for connecting to the database are to use a Unix domain socket, whereas I suspect the Docker config is set up to use networking between the containers. Unfortunately I can't provide any guidance here as to what is best from a Docker point of view!

gravitystorm:
There's a few commandline things which are pretty important to having a useful development environment, such as:

* Running the test suite
* Running rake commands (e.g. for database migrations)
* Running the console
* Connecting to the database with psql

I'd like to make sure all of these things are possible in the Dockerized development setup, since at the moment I only see mention of running the webserver. With Vagrant, for example, you can just `vagrant ssh` into the virtual machine and do all of these things.

If they are possible already (great!) then it would be worthwhile documenting how that works, since it's not obvious to me at least!

fazlerabbi37:
ok I have some good news. I was able solve the deployment problem mentioned on my previous comment. I forgot to change the volume mount point when I moved the `docker-compose.yml` to docker directory from root directory and docker was unable to find the files necessary. Silly mistake ruined my weekend :no_mouth:. But unfortunately haven't made any progress with `rake test:db` error yet.

>There's a few commandline things which are pretty important to having a useful development environment, such as:

>Running the test suite
>Running rake commands (e.g. for database migrations)
>Running the console
>Connecting to the database with psql

>I'd like to make sure all of these things are possible in the Dockerized development setup, since at the moment I only see mention of running the webserver. With Vagrant, for example, you can just vagrant ssh into the virtual machine and do all of these things.

@gravitystorm  they all are possible but for some reason I have setback regarding the `rake test:db` command. Might have something to do with the UNIX socket vs Docker tpc connection as you have mentioned in your comment. And yes you can also use `bash` on all of them.

> If they are possible already (great!) then it would be worthwhile documenting how that works, since it's not obvious to me at least!

I just went through the old PR (#1290) and found that @paulsturgess made a nice [doc](https://github.com/paulsturgess/openstreetmap-website/blob/b1f90c5219f3d103e9d9d9942230861de3203327/DOCKER.md) for docker usage. Will try to re-use/re-purpose that and push it asap. Cheers!

mmd-osm:
The developmentseed repo I mentioned earlier in this issue uses TCP connections instead of unix sockets to connect to he postgresql database.

fazlerabbi37:
> The developmentseed repo I mentioned earlier in this issue uses TCP connections instead of unix sockets to connect to he postgresql database.

come to think of it, wouldn't that UNIX socket vs Docker tpc connection issue be applicable for `rake db:migrate` command as well? As far as I understand it also connects to the `db` PostgreSQL image and creates tables and function and it works ok. If so why it fails only on `rake test:db`. While I investigate that may be @gravitystorm and @tomhughes and help me understand the process better!?

robbyoconnor:
You should also configure an unprivileged user for the Docker images.

fazlerabbi37:
> You should also configure an unprivileged user for the Docker images.

hi @robbyoconnor. I don't know a lot about unprivileged user in docker but saw `bundle` complaining about it being running as `root` so thanks for the unprivileged user creation pointer. I am still stuck with the basics ie. running the test properly and populating the database. So for now I will add this to do and try to finish it after the basic tasks are done. 

---
note to self: This medium [article](https://medium.com/better-programming/running-a-container-with-a-non-root-user-e35830d1f42a) seems to have a good explanation. Might come handy.

robbyoconnor:
@fazlerabbi37 here's the thing: Docker does have isolation HOWEVER as root -- you can escape it...in theory is it likely..probably not but I'm not willing to chance it. Always treat root in a docker container as equal to root on the host...you are root, you can escape namespaces.

docker/Dockerfile.postgres
```diff
@@ -1,4 +1,4 @@
-FROM postgres:9.4
+FROM postgres:11
```
mmd-osm: Production still uses 9.5 as of now.

tomhughes:
Only because of the xid index issues though, which relate to how we do the diff feed. I use 11 for my development setup.

docker/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: '2'
+services:
+  web:
+    image: openstreetmap-website:v1
+    build:
+      context: ..
```
robbyoconnor: Why not put the Dockerfiles in this directory?

docker/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      - "3000:3000"
+    command: bundle exec rails s -p 3000 -b '0.0.0.0'
+    depends_on:
+      - db
+    environment:
+      DATABASE_URL: postgres://postgres@db:5432
+  db:
+    image: openstreetmap-db:v1
+    build:
+      context: ..
```
robbyoconnor: Same here as above?

docker/docker-compose.yml
fazlerabbi37: Which directory? Right now all docker related file are in one directory named `docker`.

robbyoconnor: Oh wait, nevermind.

fazlerabbi37:
I have a question regarding this [line](https://github.com/openstreetmap/openstreetmap-website/blob/master/db/functions/Makefile#L4) of the [`Makefile`](https://github.com/openstreetmap/openstreetmap-website/blob/master/db/functions/Makefile). In this line the `shell bundle show quad_tile` command gets the installation path of `quad_tile` gem. What I am asking is if it is possible to do that without using `bundle`. 

The reason for using docker is not only ease of development setup but also isolation of different container like the `web` and `db`. The `db` container's concern is handling the database operation so it can work without a `ruby` or `rails` environment. On the `Makefile` the 4th line makes that job a bit hard as it both needs `ruby` or `rails` environment and the installation of all the gems of `Gemfile` and both of this are unnecessary once the `libpgosm.so` has been complied. 

cc: @tomhughes @gravitystorm

tomhughes:
Well you can just `gem install` it but you will still need ruby and rubygems, and there is then a risk that you will get a version that doesn't match. It doesn't change much though.

fazlerabbi37:
I think I asked a really stupid question on my previous comment :unamused:. For compiling `libpgosm.so` we must need `quadtile.o` file which we obtained though the `gem install` (right!? :confused: ) so `ruby` or `rails` environment is a must. To remove the `ruby` or `rails` environment dependency we need a way to get the `quadtile.o` file and get it's location without `bundle`.

---

I hacked a dirty-solution on my machine which seems to be able to compile the `libpgosm.so` and gets all the tasks successfully done. So a bit stuck here.

fazlerabbi37:
> Well you can just `gem install` it but you will still need ruby and rubygems, and there is then a risk that you will get a version that doesn't match. It doesn't change much though.

yes, tried and failed due to version mismatch as you said.

tomhughes:
I don't see how - the bundle is calling for the latest version and that is what `gem install` should give you as well. It would only be an issue if the bundle was asking for an older version for some reason.

tomhughes:
Also you don't need any `.o` files from the gem, just the header file, which is used to compile `quadtile.o` in the database functions.

fazlerabbi37:
I am really sorry for not being able to push any update since July but I have good news to share. Docker support is now ready for review. :tada:

goonux:
> I am really sorry for not being able to push any update since July but I have good news to share. Docker support is now ready for review. 
> 
> /cc: @mmd-osm @robbyoconnor have a look. can't re-request review for you guys.

Hi @fazlerabbi37 , I'm a newbie in osm, I have tried you docker-compose file and I got the error

ActiveRecord::PendingMigrationError (

Migrations are pending. To resolve this issue, run:

        bin/rails db:migrate RAILS_ENV=development

):

Then, I have ran

docker exec -it openstreetmap-website /bin/bash -c "bin/rails db:migrate RAILS_ENV=development"

But, it's still error.

fazlerabbi37:
@goonux did you followed this [instructions](https://github.com/fazlerabbi37/openstreetmap-website/blob/docker/docker/DOCKER.md) to build it? If not please give it a try. In the meantime I will try to do a fresh build with latest `master` branch code. :blush:

goonux:
> @goonux did you followed this [instructions](https://github.com/fazlerabbi37/openstreetmap-website/blob/docker/docker/DOCKER.md) to build it? If not please give it a try. In the meantime I will try to do a fresh build with latest `master` branch code.

I'm sorry, I didn't read it yet, I will try now. Thanks.

mmd-osm:
TODO: create dedicated user & group to run the server, don't just use root.

fazlerabbi37:
> TODO: create dedicated user & group to run the server, don't just use root.

@mmd-osm I am making very slow progress on this due to [SotM Asia 2019](http://stateofthemap.asia). meanwhile, it would be great if you can point me towards a few resource on this topic.

jalessio:
I took a run at improving on this PR by adding my own knowledge of Docker best practices along with some of the recommendations in this PR. I created PR https://github.com/openstreetmap/openstreetmap-website/pull/2406 which allows you to very quickly have the Rails app up and running using docker-compose. 

Please have a look - thanks.

gravitystorm:
Closing in favour of #2409 , feel free to reopen if necessary.

PR #2296, opened 2019-07-02
Switch user images to use Active Storage

tomhughes:
This is a first draft for moving the user images to use Active Storage and is setup so that in production we can use an S3 backend for the storage.

New images are saved to Active Storage, and images from there are preferred, falling back to the old images if there isn't a suitable attachment.

Once this is merged and deployed we can use a background job to migrate the existing images and then remove the old paperclip based image support.


gravitystorm:
Great! All looks good to me. Hopefully mini_magick doesn't cause as many installation problems as RMagick used to.

Only thing is around cleanups after the tests. The way the tests are currently written the ActiveStorage files get purged, since in the same test after attaching an avatar we switch to gravatar. But if the test was re-written (which it should, really) or new tests added, we could easily end up with files left over on disk, accumulating after test runs.

The [rails docs](https://edgeguides.rubyonrails.org/active_storage_overview.html#discarding-files-stored-during-system-tests) don't have an example for controller tests unfortunately, but going by the others it's not something taken care of automatically.

mmd-osm:
We're currently discussing this on Slack, as it seems to affect some apps out there.

Previously, http://localhost:3000/api/0.6/user/1 returned absolute urls, e.g. ` image lookup fails.

bhousel:
oh yeah these look broken in the iD notes interface too..

Screenshot 2019-07-16 17 01 46

(cc @quincylvania )

tomhughes:
I can easily change it but as far as I can tell (and it's hard now as the code has changed) the old code gave relative URLs as well.

Certainly the test (https://github.com/openstreetmap/openstreetmap-website/blob/master/test/helpers/user_helper_test.rb#L57) is expecting to get a relative value.

mmd-osm:
I still have a Rails port running locally which is about 3 weeks old, and it returns some absolute urls.

tomhughes:
I'm sure you're right - it's what I was expecting. I'm just now sure how it was doing it ;-)

mmd-osm:
https://github.com/tomhughes/openstreetmap-website/blob/75e60acf661e39f3ab01d486ed17616bab036f73/app/helpers/user_helper.rb#L53

Is this maybe some side effect of using `url_for` in user_image_url?

tomhughes:
Well it is but I'm more interested in the case after that, which is the old code, and how that worked and why in the test is gave a relative link.

The fix is deployed now anyway.

PR #2313, opened 2019-07-15
Send trace import result notification immediately

### LINKED ISSUE #2312: GPX import failure emails are not being sent
GPX import failure emails are not being sent because the message template tries to access the trace but because the email sending is done as a delayed job by the time it is processed the failed trace has been deleted.

tomhughes:
If we delay sending failure notifications then they will fail because the trace will no longer exist.

As trace imports are running in the background anyway there doesn't seem to be any good reason to defer the emails.

Fixes #2312

### DIFF
--- a/app/jobs/trace_importer_job.rb
+++ b/app/jobs/trace_importer_job.rb
@@ -5,15 +5,15 @@ class TraceImporterJob < ApplicationJob
     gpx = trace.import
 
     if gpx.actual_points.positive?
-      Notifier.gpx_success(trace, gpx.actual_points).deliver_later
+      Notifier.gpx_success(trace, gpx.actual_points).deliver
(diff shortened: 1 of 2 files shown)


gravitystorm:
Looks fine to me - I'm happy with emails being delievered synchronously since the task is being done asynchronously. 

It's a pity this problem wasn't picked up in the tests, but that's because there the jobs are run immediately. I could design some kind of state machine for this - only delete when pending notifications have been sent - but that seems cumbersome.

Thinking about it some more, we could redesign the Notifier.gpx_failure calls to take the data they need, rather than a trace object. See https://ricostacruz.com/til/rails-mailer-race-conditions . But I'm not sure if it's worth the trade-off in this situation.

PR #2318, opened 2019-07-17
Update annotations

gravitystorm:
The annotations for the user model changed after 91a37235ac748df7dc241b215790d6cbecd652f3 due to ignoring these columns.


tomhughes:
I've already got a commit ready to go in a day or two that removes the fields from the database complete and includes this - it's only because the stupid strong migrations stuff that it's a mess at the moment. I assume it's also why travis is screaming although I couldn't really figure out WTF was going on there.

PR #2320, opened 2019-07-17
Models in migrations

gravitystorm:
A recent change to the user model triggered breakages in the database migrations. This is because the migrations have, until now, been using the live model in old migrations. When there is a table name change, or certain column changes, in the live model, this can blow up the old migration.

Instead, it's best practise that if a model is needed during the migration, the model is defined within the migration. This ensures that the model picks up the table name, columns etc that match what's in the database at that stage.

### DIFF
--- a/db/migrate/20120219161649_add_user_image_fingerprint.rb
+++ b/db/migrate/20120219161649_add_user_image_fingerprint.rb
@@ -1,4 +1,7 @@
 class AddUserImageFingerprint < ActiveRecord::Migration[4.2]
+  class User < ActiveRecord::Base
+  end
+
   def up
     add_column :users, :image_fingerprint, :string, :null => true
 
--- a/db/migrate/20150111192335_subscribe_old_changesets.rb
+++ b/db/migrate/20150111192335_subscribe_old_changesets.rb
@@ -1,4 +1,7 @@
 class SubscribeOldChangesets < ActiveRecord::Migration[4.2]
+  class Changeset < ActiveRecord::Base
+  end
+
   def up
     Changeset.find_each do |changeset|
       changeset.subscribers << changeset.user unless changeset.subscribers.exists?(changeset.user.id)
--- a/db/migrate/20161011010929_subscribe_authors_to_diary_entries.rb
+++ b/db/migrate/20161011010929_subscribe_authors_to_diary_entries.rb
@@ -1,4 +1,7 @@
 class SubscribeAuthorsToDiaryEntries < ActiveRecord::Migration[4.2]
+  class DiaryEntry < ActiveRecord::Base
+  end
+
   def up
     DiaryEntry.find_each do |diary_entry|
       diary_entry.subscriptions.create(:user => diary_entry.user) unless diary_entry.subscribers.exists?(diary_entry.user.id)
--- a/db/migrate/005_tile_tracepoints.rb
+++ b/db/migrate/005_tile_tracepoints.rb
@@ -1,6 +1,10 @@
 require "migrate"
 
 class TileTracepoints < ActiveRecord::Migration[4.2]
+  class Tracepoint < ActiveRecord::Base
+    self.table_name = "gps_points"
+  end
+
   def self.up
     add_column "gps_points", "tile", :bigint
     add_index "gps_points", ["tile"], :name => "points_tile_idx"
--- a/db/migrate/006_tile_nodes.rb
+++ b/db/migrate/006_tile_nodes.rb
@@ -1,6 +1,14 @@
 require "migrate"
 
 class TileNodes < ActiveRecord::Migration[4.2]
+  class Node < ActiveRecord::Base
+    self.table_name = "current_nodes"
+  end
+
+  class OldNode < ActiveRecord::Base
+    self.table_name = "nodes"
+  end
+
   def self.upgrade_table(from_table, to_table, model)
     if ENV["USE_DB_FUNCTIONS"]
       execute <<-SQL
--- a/db/migrate/013_add_email_valid.rb
+++ b/db/migrate/013_add_email_valid.rb
@@ -1,4 +1,7 @@
 class AddEmailValid < ActiveRecord::Migration[4.2]
+  class User < ActiveRecord::Base
+  end
+
   def self.up
     add_column "users", "email_valid", :boolean, :default => false, :null => false
     User.update_all("email_valid = (active != 0)") # email_valid is :boolean, but active is :integer. "email_valid = active" (see r11802 or earlier) will fail for stricter dbs than mysql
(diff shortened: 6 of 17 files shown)


tomhughes:
Oh come on, this is just insane.

Where exactly is this "best practice" advice coming from? Please don't tell me it's `strong_migrations`...

tomhughes:
I'm certainly more convinced than ever by this that the `strong_migrations` advice to handle removing columns via two commits where you first deploy something to ignore them and then deploy a migration to remove them (which still needs to be marked as unsafe by the way!) is just crazy.

The fundamental problem is that it ties together the git history and the migration history so you have to execute them in sync with each other - this idea here is one result, and a second is that you can't go back and forth through such a migration without also rolling git back and forth at the same time.

tomhughes:
By the way my commit to drop the columns was also going to remove the ignore line from the class which would unbreak the tests here. It was unclear if that was what `strong_migrations` intended or if they wanted the class cluttered up with ever growing lists of ignored fields forever more.

gravitystorm:
> Oh come on, this is just insane.

No it's not, it's perfectly normal.

> Where exactly is this "best practice" advice coming from? Please don't tell me it's `strong_migrations`...

It's got nothing to do with strong migrations. See e.g. http://railsguides.net/change-data-in-migrations-like-a-boss/ https://github.com/rubocop-hq/rails-style-guide#define-model-class-migrations http://blog.testdouble.com/posts/2014-11-04-healthy-migration-habits.html and elsewhere.

This PR is exactly to decouple the old migrations from the current state of the models. We should never have been relying on the current models in the old migrations, right from the moment they were written. If we ever e.g. remove a model, or rename a table when an old migration depends on the model, etc et, then the migrations break.

tomhughes:
Frankly I'm astonished that those minimal models even work - some of our migrations do pretty extreme things and I'm surprised none of them are relying on the details of the models.

I rather suspect that maybe they only work with an empty database, and that some of the ones that try and mutate existing data might fail if there was any data to change?

db/migrate/20120219161649_add_user_image_fingerprint.rb
```diff
@@ -1,4 +1,7 @@
 class AddUserImageFingerprint < ActiveRecord::Migration[4.2]
+  class User < ActiveRecord::Base
+  end
+
```
tomhughes: I don't think this will work if there are any users in the database because the migration relies on being able to access the images attached to the user via the paperclip attachment, which has not been declared here.

That of course gets even more fun once we remove paperclip from the Gemfile...

db/migrate/20150111192335_subscribe_old_changesets.rb
```diff
@@ -1,4 +1,7 @@
 class SubscribeOldChangesets < ActiveRecord::Migration[4.2]
+  class Changeset < ActiveRecord::Base
+  end
+
```
tomhughes: I think this will fail if there are changesets in the database as the migration relies on the `subscribers` and `user` associations on the changeset.

gravitystorm:
> I'm certainly more convinced than ever by this that the `strong_migrations` advice to handle removing columns via two commits where you first deploy something to ignore them and then deploy a migration to remove them (which still needs to be marked as unsafe by the way!) is just crazy.

OK, we're getting off topic slightly here, but it's not crazy. If you want to drop a column, you need to first ensure that no running copy of the codebase expects that column to be there. Since each copy reads the database when it boots, and caches the table structure in memory, and makes it part of every copy of that model that it handles, that causes a situation with a few options:

* Drop the column. Get lots of errors in production from the instances that are still trying to use that column. Restart all running app instances as quickly as possible.
* Stop all running app instances. Drop the column. Restart all instances and live with the downtime that just happened.
* Configure the models to ignore the column. Restart the instances with that code. Drop the column at your leisure.

That's why strong_migrations suggests the third option.

db/migrate/20161011010929_subscribe_authors_to_diary_entries.rb
```diff
@@ -1,4 +1,7 @@
 class SubscribeAuthorsToDiaryEntries < ActiveRecord::Migration[4.2]
+  class DiaryEntry < ActiveRecord::Base
+  end
+
```
tomhughes: Again the migration relies on the `subscriptions` and `user` associations.

gravitystorm:
> Frankly I'm astonished that those minimal models even work - some of our migrations do pretty extreme things and I'm surprised none of them are relying on the details of the models.
> 
> I rather suspect that maybe they only work with an empty database, and that some of the ones that try and mutate existing data might fail if there was any data to change?

Oh almost certainly. But to be honest I'm not going to spend any time on that. If the migrations now work for an empty database, and are more robust to future changes, so much the better.

From a broader perspective, we should really ditch our old migrations. I'd be happy replacing everything before e.g. 2017 with one migration called "2017010101_current_structure" since they are full of dodgy stuff regarding segments and c extensions and whatnot that literally nobody needs to use in 2019. If anyone is truly interested in how they worked, then there's a git history. Otherwise we need to keep maintaining them for no gain.

db/migrate/20120219161649_add_user_image_fingerprint.rb
gravitystorm: So what should we do? Breakages from removing paperclip in the near future is the kind of fragility that comes with relying on current model code in migrations.

Of course one option is to reimplement the migration so that it doesn't depend on paperclip at all, i.e. writing some custom code to handle the fingerprint generation. This allows us to remove paperclip yet keeping this migration 100% functional. But it hardly seems worthwhile us of time. 

The second option would be to assume (like I do) that nobody is running that old a version of the database so it's all hypothetical anyway and keep what is written here. It'll work for an empty database but not one with data in it. Same situation will happen when we remove paperclip, even without this change.

And the third option, if we think migrations from this age of database are all hypothetical anyway, is to just ditch these old migrations completely.

tomhughes:
Well I guess we could merge this for now.

Long term I'm not sure I have any better ideas than your one of rolling up all the old migrations, though that is a lot of work to get right... The final result would have to reuse an existing timestamp to make sure existing users didn't try and apply it.

db/migrate/20120219161649_add_user_image_fingerprint.rb
tomhughes: Well I guess we could merge this for now.

Long term I'm not sure I have any better ideas than your one of rolling up all the old migrations, though that is a lot of work to get right... The final result would have to reuse an existing timestamp to make sure existing users didn't try and apply it.

tomhughes:
I've merged this but also added 227c402ad10059ec35394f2a960a620e5cda73af to remove the code to add fingerprints to images - it is irrelevant for anybody migrating up to the latest revision anyway.

I also added 826b70fdc5f44ab2f6bde94b8ab67bf98ee43981 to move the two subscription ones (which are basically optional to enable subscription for existing entries) into rake tasks that people can run as needed if updating an old installation.

Issue #2327, opened 2019-07-25
Directions geocoding doesn't take current location into account


mtmail:
**Issue**

While the search endpoint (`GeocoderController`) sends the current map bounding box to the geocoder (Nominatim), the geocoding AJAX request in `directions.js` doesn't.

**To reproduce**

Normal search
1. Search for Manchester https://www.openstreetmap.org/search?query=manchester
2. Search for King Street https://www.openstreetmap.org/search?query=king+street
3. a King Street in Manchester is found (https://www.openstreetmap.org/way/4961636)

vs

Directions search
1. Search for Manchester https://www.openstreetmap.org/search?query=manchester
2. Open directions dialog, it shows start and destination fields. Start is now prefilled with `Manchester, Greater Manchester, North West England, England, United Kingdom`
3. In the destination field type in "King Street" and press `[TAB]` or unfocus the input field
4. a King Street in Derby is found, far away

**Expected behavior**

King Street near Manchester should be set as destination.

tomhughes:
The only thing I wonder about here is that it's probably quite common for one of the end points not to be in view, or even nearby, when setting up a route...

That said this does only give preference to locations near the view box so it's probably reasonable.

PR #2328, opened 2019-07-25
geocode request inside directions.js: add &viewbox parameter

mtmail:
fixes https://github.com/openstreetmap/openstreetmap-website/issues/2327


tomhughes:
There's no need to open an issue if you're immediately going to open a PR to fix it - just describing the problem in the PR is fine.

Issue #2336, opened 2019-08-02
Set <title> upon search


jidanni:
Comparing the browser tab title of
1. https://www.google.com/maps/search/電信局前
1. https://www.openstreetmap.org/search?query=電信局前

OSM should set the HTML '<title>' too.

samrjenkins:
@jidanni I have created PR #2347 for this issue

PR #2347, opened 2019-08-10
set <title> upon search

samrjenkins:
This is a response to issue: [#2336](https://github.com/openstreetmap/openstreetmap-website/issues/2336)

Prepend search query string to HTML <title>

Example:
If query = `New York`
Title content will be set to: `New York | OpenStreetMaps`

### DIFF
--- a/app/views/geocoder/search.html.erb
+++ b/app/views/geocoder/search.html.erb
@@ -1,3 +1,5 @@
+<% set_title(@params[:query]) %>
+
 <h2>
     <a class="geolink" href="<%= root_path %>"><span class="icon close"></span></a>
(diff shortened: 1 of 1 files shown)


tomhughes:
Can you give an example of what the resulting title looks like?

tomhughes:
In fact have you checked that this works?

Only I wouldn't have thought setting it there would work in most cases - it will work if you load a URL with a search query in it, but it won't update when you do a search in an already loaded page.

tomhughes:
Yes I know it sets `X-Page-Title` but I didn't think `search.html` would be rendered for an XHR request. I might be wrong though - will have a look later.

Issue #2348, opened 2019-08-12
Multi Fetch GET with /full option?


bhousel:
Per https://github.com/openstreetmap/iD/issues/6668#issuecomment-520489274
iD is using the [Multi Fetch GET](https://wiki.openstreetmap.org/wiki/API_v0.6#Multi_fetch:_GET_.2Fapi.2F0.6.2F.5Bnodes.7Cways.7Crelations.5D.3F.23parameters) API call in a few situations to fetch ways and relations:
- downloading additional member ways for correct handling of multipolygons, routes, etc.
- checking and dealing with error 409 conflicts at save time

Unfortunately we need to still perform more calls after this to get all the related nodes (and this can be a lot of nodes).

It would be really great if we could have a version of this call that works like the [`/full` Single Fetch GET call](https://wiki.openstreetmap.org/wiki/API_v0.6#Full:_GET_.2Fapi.2F0.6.2F.5Bway.7Crelation.5D.2F.23id.2Ffull) that returns all the related nodes in the response.

cc @mmd-osm

gravitystorm:
Seems reasonable to me.

mmd-osm:
As I mentioned in the linked issue, we still need to sort out, how the new "multi fetch with full option" is exposed via the API. Should we extend the existing endpoint (if so, how exactly should that look like?), or would it make more sense to define a new endpoint?
That's more or less the same level of detail you would need to update the OSM API documentation in the WIki.

pnorman:
> As I mentioned in the linked issue, we still need to sort out, how the new "multi fetch with full option" is exposed via the API. Should we extend the existing endpoint (if so, how exactly should that look like?), or would it make more sense to define a new endpoint?

Any opinions on this from those who want the new endpoint?

Issue #2162, opened 2019-02-28
Capabilities call and versioning


gravitystorm:
I'm working on adding support for running multiple versions of the API in parallel. This is because I think the old way we used to do API upgrades (e.g. switch off openstreetmap for a few days, switch it back on again with the new version) might not be tolerated in 2019 :smile: 

I started work on this at the Karlsruhe Hack Weekend and if you want to see some WIP and get a better idea what I'm thinking about then have a look at https://github.com/gravitystorm/openstreetmap-website/commit/f574838ddabf3dc871da25ae9ccc331c2537d0c1 (specifically the config/routes.rb file).

One topic came up that stumped me. We currently have two capabilities calls, that currently return the same thing:

* [`api/capabilities`](https://www.openstreetmap.org/api/capabilities) (the 'unversioned' call)
* [`api/0.6/capabilities`](https://www.openstreetmap.org/api/0.6/capabilities) (the 'versioned' call)

A hypothetical `api/0.7/capabilities` response could indicate that it is version 0.7 (in the root `osm` element) but 0.6 is still available (in the `versions` element), for example:

```
<osm version="0.7" generator="OpenStreetMap server" ...>
  <api>
    <version minimum="0.6" maximum="0.7"/>
    <area maximum="0.25"/>
    ....
```

But my immediate quandary is - for the unversioned `api/capabilities` call - what version should be shown in the root element when multiple versions are available? `<osm version="???" ...>`

It also got me thinking - should we even have an unversioned capabilities call? What happens if we want to change that response in future? Should we instead have an `api/versions` call that just indicates supported versions, and clients can then call `api/X/capabilities` for the one they want to use? Should we remove the version information from `api/X/capabilities` since it's not really a capability of that version of the API?

Does anything (e.g. editors) actually use the current version information anyway? Do requests for capabilities generally use the versioned or unversioned calls?

Discuss!

bhousel:
> Does anything (e.g. editors) actually use the current version information anyway? Do requests for capabilities generally use the versioned or unversioned calls?

iD currently calls `api/capabilities` (unversioned) but we should change this.

It makes more sense to me to just drop the unversioned endpoint and have separate resources for:
* `api/0.6/capabilities`
* `api/0.7/capabilities`

> Should we instead have an api/versions call that just indicates supported versions, and clients can then call api/X/capabilities for the one they want to use? 

seems fine 👍 

[...]

PR #2353, opened 2019-08-21
Supporting multiple API versions

gravitystorm:
I'd like to support multiple API versions, so that we can deploy API 0.7 in parallel to API 0.6. It's not something we've ever done before, but I think it's the only reasonable way to handle API version changes these days!

I'm opening this PR just to receive any initial feedback on this proposed approach to the code changes. So far this approach demonstrates the basic concepts, including:

* Supporting different scenarios, e.g. dropping old api calls, adding new api calls, api calls that are the same in each version, api calls that are similar but return different results (e.g. different response codes)
* Optimised for the assumption that almost all api calls are the same between different versions.
* Being able to choose which versions are deployed. This allows the codebase to support multiple API versions during development and when running the tests, and so it avoids long-running branches.
* Only code that is different between api versions is in any way duplicated. Otherwise the same code is used for each api version.
* Establishes a naming convention for controllers that behave differently in different API versions. The general idea is that old code lives in specific controllers, e.g. `API::V06::CapabilitiesController`, and new code lives in the normal place e.g. `API::CapabilitiesController`. This makes future upgrades easier, since the assumption is that new code will be used in version N+1 too.
* It's all designed to allow multiple, non-contiguous versions. For example, if we yank a future version 0.8, it'll cope fine with e.g. deployed_versions = ["0.6", "0.7", "0.9"]

If you are interested in seeing how it works, then the changes to "config/routes.rb" along with the output from "bundle exec rake routes" are the best way to see the overall idea.

The biggest drawbacks so far are around the tests. 

First, I think it makes sense that every API version that the codebase knows about is fully tested, i.e. even if the result is expected to be the same, a given api test should run once for each api version. This leads to the indentation of all the tests changing, since we need to add a "all_api_versions" loop around almost every test. So the diffs and git blame are horrible.

Secondly, the controller methods involve lots of changes like this:

```diff
- get :show
+ get :show, :params => { :api_version => version }
```
It leads to a lot of extra `params => ...` to skim read, which isn't ideal. I've tried working around this but [the workarounds have their own drawbacks](https://github.com/gravitystorm/openstreetmap-website/issues/28#issuecomment-523427402).

The tests don't all pass yet, so this is not yet in shape to be committed, but I hope to get it there soon. In the meantime, and before I make any further changes that you might not be happy with, any feedback is very welcome!


mmd-osm:
I haven't looked through all the details yet, so please bear with me. I was wondering a bit how you would handle evolutionary or even revolutionary changes to the data model. Typically this is one of the most troubling and complicated aspects of new API versions, and I think it would be good to have some basic concept in place as well here.

gravitystorm:
>  I was wondering a bit how you would handle evolutionary or even revolutionary changes to the data model.

I can sum this up with the phrase "it depends"!
* Different versions of the API can use different views, so if we added something to a future api (for example "node coordinate precision" or somesuch) it can be shown in the latest version and ignored in previous versions, or shown in a different way (e.g. as tags).
* Different api versions can use different controllers too, as already demoed in this PR for the capabilities controller. This allows more subtantial changes in behaviour between api versions, and changes like fixing response codes.
* We could use different models too. I'm not sure why we'd want to, but e.g. api/0.7/node/1 could call a node_controller that fetches data from the ShinyNewNode model.

But it all really depends. I think the underlying question might be "how do we handle drastically non-backwards-compatible changes, like converting all closed ways into an area type (or if we introduce an area type into API N+1, how does that work with API N clients); or how would a change like removing segments work with two api versions running in parallel. I don't have an answer for these major structural changes, except to say that if it's logically possible at all to access the data between versions, the code approach in this PR will be able to handle it. And perhaps there'll be some change to the datastructure that prevents parallel API versions and it'll trigger a hard cutover between versions.

But lets not get stuck on the biggest problem, and one that's not yet in hand. In the meantime, there's a ton of backwards-compatible but needs-API-bump changes that have been stuck for years, so we can at least sort out those ones :smile:

gravitystorm:
I should have said, if there's specific API changes that you're thinking about, even if they are just to illustrate a point, let me know and I'll see how they fit.

simonpoole:
> But lets not get stuck on the biggest problem, and one that's not yet in hand. In the meantime, there's a ton of backwards-compatible but needs-API-bump changes that have been stuck for years, so we can at least sort out those ones 😄

Likely making myself very unpopular:

- where are all these "backward-compatible but needs-API-bump changes that have been stuck for years changes" documented?
- in general I oppose all of this on the grounds that the GDPR related changes need to be implemented first and I'm actually surprised that "nice to have" API changes are even being given a seconds thought in the current situation.

gravitystorm:
>     * where are all these "backward-compatible but needs-API-bump changes that have been stuck for years changes" documented?

Find your favourite API-0.7 wishlist, ignore the stuff about areas, and there's your list. :smile: 

Different people have created different lists over the years. I'm not intending to implement many changes, just the ones that I've been personally complaining about since API 0.6 was released over a decade ago (like incorrect http status codes, and plain-text responses, and stuff like that).

> in general I oppose all of this on the grounds that the GDPR related changes need to be implemented first

I know that you **want** them first, but that doesn't mean that they **need** to be done first. 

In particular, if the GDPR-related changes can be made without changing the API version then they can be implemented in parallel to this work, and so they are not interdependent. If they need to break API compatibility, then this work will need to come first anyway. So either way, they don't need to come first.

bhousel:
> I should have said, if there's specific API changes that you're thinking about, even if they are just to illustrate a point, let me know and I'll see how they fit.

Just off the top of my head, we'd see some immediate performance improvement on the iD side from:
- JSON API (#2221)
- Multi-fetch GET with /full (#2348)

(these require some coordination from the CGImap side too, but I don't think that's a blocker)

simonpoole:
Except, naturally, that neither of #2221 and #2348 depend on this PR. 

Without it (this PR) we would have simply added them as a 0.6 feature, as we have done so many times before and not doing that raises the whole question of versioning of the OSM API in general which is a rabbit hole of its own ,

gravitystorm:
I've updated this PR so that the test suite passes, and it's now ready for review.

Currently only a few routes have been adapted for multiple API version support, namely:

* capabilities
* permissions
* changesets

I intend to work on the rest of the routes in subsequent PRs. The settings in this PR ensure that only 0.6 is deployed by default, so it can be merged without any side effects.

simonpoole:
Could we get it on the sandbox first?

gravitystorm:
> Could we get it on the sandbox first?

First? Do you mean before code review?

Or if you want it before merging, to what end? So far this PR is just internal code refactoring, there's no changes to the API (other than dropping one line from the api/0.7/capabilities response) and even if this is merged there will still be no changes since 0.7 is disabled. So we can ask Tom to set up a sandbox "first", but I'm not sure what you would want to do with it?

Of course, it'll be worth having a sandbox available later on, but I don't think it's worthwhile effort at this stage.

If you still have concerns, let me know what I can do to help.

simonpoole:
> 
> 
> > Could we get it on the sandbox first?
> 
> First? Do you mean before code review?

Before deployment, which in our case implies before merger.
> 
> Or if you want it before merging, to what end? So far this PR is just internal code refactoring, there's no changes to the API (other than dropping one line from the api/0.7/capabilities response) and even if this is merged there will still be no changes since 0.7 is disabled. 

Famous last words. In reality there are always things that might break, for example as when the  authorisation refactoring was deployed.

Being to test against a deployment, while not a panacea, at least gives us a fighting chance to ferret out any assumptions that no longer hold true and so on.

mmd-osm:
>  I'm not intending to implement many changes, just the ones that I've been personally complaining about since API 0.6 was released over a decade ago (like incorrect http status codes, and plain-text responses, and stuff like that).

That's the bit I least like about the current multiple API version idea: it seems to focus on something I would describe as _cosmetic changes_ only. That's ok, except for it only adds work to consumers of the API while offering no real value to them at all. Since they already have a working API integration, they would need to spend extra development time and effort to get back to a status quo. 

Isn't there anything more compelling to do that would give API consumers more of an incentive to move to the latest and greatest version?

gravitystorm:
> adds work to consumers of the API

*existing* consumers of the API. Some of these headaches and quirks need to be solved by every API consumer, and the total number of future API consumers yet to be written vastly exceeds the ones written so far. So the sooner we fix them, the better, and it's a shame they've been known about and unfixed for so long already.

> Isn't there anything more compelling to do that would give API consumers more of an incentive to move to the latest and greatest version?

Maybe we will want to put some of the more compelling things in v0.8, or v0.9, or something. But I'm determined to avoid getting into the same situation as has happened over, and over, and over again, where the scope of v0.7 expands inexorably until it collapses under its own weight! I'd rather break up the logjam and work on smaller, more frequent improvements (e.g. every 18-36 months, rather than 10+ years and counting) to the API. And I'd rather keep the API changes small so that a developer can upgrade their app with a small amount of code changes that's feasible in a weekend of hacking, rather than some significant API upgrade that puts them off doing any of it and leaves them stuck on v0.6. And making life easier for the developers is the whole point of adding multiple version support, so that they can upgrade when it suits them best.

Anyway, let's talk about this further in a future issue, since we're burying the point of this PR ("is this the best code approach to multiple version support? Can you see a better way of coding the tests?") in wider discussions.

test/controllers/api/capabilities_controller_test.rb
```diff
@@ -7,27 +7,54 @@ class CapabilitiesControllerTest < ActionController::TestCase
     def test_routes
       assert_routing(
         { :path => "/api/capabilities", :method => :get },
-        { :controller => "api/capabilities", :action => "show" }
+        { :controller => "api/v06/capabilities", :action => "show", :api_version => "0.6" }
```
mmd-osm: Say we want to use semantic versioning like 1.2.0, how would this be reflected specifically in directory names like `v06`?

app/assets/javascripts/osm.js.erb
```diff
@@ -10,7 +10,7 @@ OSM = {
   MAX_REQUEST_AREA:        <%= Settings.max_request_area.to_json %>,
   SERVER_PROTOCOL:         <%= Settings.server_protocol.to_json %>,
   SERVER_URL:              <%= Settings.server_url.to_json %>,
-  API_VERSION:             <%= Settings.api_version.to_json %>,
+  API_VERSION:             <%= Settings.api_versions.min_by(&:to_f).to_json %>,
```
mmd-osm: Why is this value set to "min_by", and what are the implications of it? Does `&:to_f` play nice with semver (e.g. 1.2.0)?

simonpoole:
@gravitystorm you might want to consider making a statement as to versioning of the API (for example  if it follow semver semantics going forward), or will we have to continue to assume that every version change is breaking as it is now.

config/routes.rb
```diff
@@ -1,24 +1,58 @@
 OpenStreetMap::Application.routes.draw do
+  # Our api versions are decimals, but controllers cannot start with a number
+  # or contain punctuation
+  def v_string(version)
+    "v#{version.delete('.')}"
```
mmd-osm: Thinking of semver again, would it be better to replace "." by "_" maybe, so it's `v0_7 ` and `v1_2 `? How about patch version number? Ignore or include?

test/controllers/api/changesets_controller_test.rb
```diff
@@ -267,1035 +292,1086 @@ def test_close_not_found
...
+        changeset = create(:changeset, :user => user)
+        node = create(:node)
+        way = create(:way_with_nodes, :nodes_count => 2)
+        relation = create(:relation)
+
+        basic_authorization user.email, "test"
+
+        # simple diff to create a node way and relation using placeholders
+        diff = <<CHANGESET.strip_heredoc
+          <osmChange>
```
mmd-osm: At one point we'd also need to include a version number in `<osmChange>` (maybe not now)

mmd-osm:
> I'd rather break up the logjam and work on smaller, more frequent improvements (e.g. every 18-36 months, rather than 10+ years and counting) to the API.

This brings up the topic how long you want to support previous API versions. Effectively, we need a Versioning Policy and a clear messaging to API consumers about EOL dates for each version well ahead of time.

test/controllers/api/capabilities_controller_test.rb
simonpoole: @mmd-osm a reasonable assumption IMHO would be to only actually have different "directory" names for major versions (aka breaking changes), the client can then determine from the capabilities which minor version is actually supported and from that determine which backwards compatible features are present.

gravitystorm:
> every version change is breaking as it is now

That will always be the case. If it's not breaking, then there's no need to change the API version in the first place.

I'm sure you're aware that the current abilities of API 0.6 are quite expanded from the initial version back in 2009, but so far it's had to be backwards compatible to avoid breaking clients (with some minor exceptions, like the swf endpoint). Because we've only (been able to|chosen to) deploy backwards-compatible changes, we haven't incremented the version number.

https://stackoverflow.com/questions/27901549/semantic-versioning-of-rest-apis explains it clearly - if we moved to semantic versioning, then only the major number is of interest to API consumers. So that's what we'd use in the URLs. If we wanted to move to semantic versioning for the next version of the API, then I would suggest calling it "API v7". 

We'd never need to have an endpoint like `/api/v7.1.0/node/1` since, by the definition of semantic versioning, any call to that would be backwards-compatible with any other API 7.x endpoint. If it wasn't compatible, we'd be on API v8, not v7.1. 

I'd be happy to move to using integers instead of what we have now, but I suggest we park this for the release-after-next.

app/assets/javascripts/osm.js.erb
gravitystorm: It's mainly just min_by to keep everything working for now, since it will pick 0.6 unless the site operator chooses to only deploy 0.7. 

It's used for the bits of the website that talk to the API, like notes and changeset comments. Since there's no changes yet, it's more of a "pick either" situation.

On a wider point, I'd rather work on refactoring those bits of the site to just be regular webpages like diary entry comments, but that's a different project!

config/routes.rb
gravitystorm: As discussed in the comments, semver would only use major version numbers here. So e.g. `v7` or `v124`. But it's a great point to raise, thank you.

test/controllers/api/changesets_controller_test.rb
gravitystorm: Yeah, I noticed that during the refactoring, but lets leave that for now since we're not planning any changes to the osmchange format (afaik).

gravitystorm:
@mmd-osm Thanks for looking at the code! That's really helpful and I appreciate it. If you have any suggestions or more feedback I'd love to hear it.

simonpoole:
> 
> 
> > every version change is breaking as it is now
> 
> That will always be the case. If it's not breaking, then there's no need to change the API version in the first place.
> 
> I'm sure you're aware that the current abilities of API 0.6 are quite expanded from the initial version back in 2009, but so far it's had to be backwards compatible to avoid breaking clients (with some minor exceptions, like the swf endpoint). Because we've only (been able to|chosen to) deploy backwards-compatible changes, we haven't incremented the version number.
> 
> https://stackoverflow.com/questions/27901549/semantic-versioning-of-rest-apis explains it clearly - if we moved to semantic versioning, then only the major number is of interest to API consumers. So that's what we'd use in the URLs. If we wanted to move to semantic versioning for the next version of the API, then I would suggest calling it "API v7".
> 

Neither the original semver definition nor the SO article say that. 

We agree that in the calls, only the major version needs to be used, but, because semver allows backwards compatible additions you still need some way of determining what version you -exactly- have in front of you to know if a specific minor version addition is present or not. 

In the case of the OSM API that is easily achieved by  returning the actual semver (so if I want to use feature X any minor version higher than the one it was introduced in are OK). That we can't do that currently with 0.6 is the major issue with all the additions that have happened over the years.

test/controllers/api/changesets_controller_test.rb
simonpoole: As I pointed out above, we return the data model version in all kinds of places, not just in OSC format. That needs to be decoupled from the API version.

gravitystorm:
> you still need some way of determining what version you -exactly- have in front of you

So let's take a feature like "notes" which was added in during the 0.6 era. I think your use-case is along the lines of "I'm an editor developer, and I want to know if notes are available on this endpoint, which might be osm.org or might be some other deployment that is running an older codebase that doesn't have notes yet, so that I can show/hide the relevant UI". If I've misunderstood let me know.

You're suggesting that we should therefore introduce the minor API version in responses. I suspect you don't need to know the minor version in every response (e.g. `/api/X/ways/1` -> `<osm version="7.9" ...`), but you need a response from the server to let you know what the server is capable of doing. So maybe just in the `/api/X/capabilities` call? And then you can show/hide notes functionality depending on whether X.Y is greater than the version number where notes were introduced.

But then, why bother with lists of features and their corresponding minor version numbers? That's a bit indirect, since we could just put the additional features into the `capabilities` response e.g. `<feature name="notes" />` to indicate that the endpoint now supports notes. Would that work?

gravitystorm:
> This will change the data model version in sync with the API version. Which a) normally wouldn't be the case, b) will lead to all kind of breakage in the XML parsers.

I've read this a few times, but I don't understand what you mean. Could you give an example?

simonpoole:
Consider for example the output of https://api.openstreetmap.org/api/0.6/map?bbox=-122.083819,37.4217991,-122.0834335,37.4222209

Which starts with 

`<osm version="0.6" generator="CGImap 0.7.5 (14135 thorn-03.openstreetmap.org)" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
<bounds minlat="37.4217991" minlon="-122.0838190" maxlat="37.4222209" maxlon="-122.0834335"/>
`

Giving the version of the OSM XML document as 0.6, as this is the only indication of the format in use and OSM does not refer to any external schema docs, you will find that a lot of stuff will bork if you change that 0.6 to 0.7 or whatever.

mmd-osm:
I tend to agree that we have multiple lifecycles/versioning depending on what you look at. One the one hand, we add features over time to the API in an incremental/non breaking way by either extending existing endpoints or adding new ones. The downside of the current approach is that we have no mechanism to communicate those changes to the outside world. I think this would be a good ft for some semantic versioning, and now would be a good time to seriously think about it, rather than introducing some 0.7 version with some fuzzy requirements. 

Now, regarding the XML documents: everything that has been done in the recent years mostly left those unchanged. Some minor tweaks like adding an `if-unused` attribute in the OsmChange message happened on a message that isn't versioned anyway. For the OSM XML format, I think we never did any changes (?!), and any proposed changes (apart from introducing an area model) don't seem to force us to create a new version of the OSM XML format.

Re. the future OSM JSON format (which will be using the Overpass API JSON format), I'm not sure what the best approach would be. Technically, it's still using the same OSM API 0.6 data model underneath, so we might just call that 0.6 as well?

I think it would be good to  establish an independent versioning for the different parts.

simonpoole:
@mmd-osm well the only data model change that is in a stage of serious consideration (changing way geometry representation), would require a complete switch over (as a backwards compatibility would be very very difficult to provide), so IMHO not worth considering in the context of this PR. 

With other words the version in the documents could just as well be hardwired to 0.6 for now.

bhousel:
Probably worth mentioning that semantic versioning guidelines [explicitly allow a 0.x major version](https://semver.org/#spec-item-5) as a signal that anything can change at any time. 

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

So we are, in a sense, already doing semver the way it's intended.  If the OSM API ever gets to v1 then we need to start thinking more about backward compatibility and breakage.

simonpoole:
At the time the 0.6 api was released semver was still two years away. Trying to retroactively apply senver semantics doesn't really work.

bhousel:
> trying to retroactively apply senver semantics doesn't really work.

It's a number.  it either works or it doesn't.  (in this case, it does)

simonpoole:
> 
> 
> > trying to retroactively apply senver semantics doesn't really work.
> 
> It's a number. it either works or it doesn't. (in this case, it does)

You were implying to that the *number" gives us licence to break everything at will, not only is that not supported by the facts (as said the "number" predates the concept of semver), it relveals a very hostile attitude toward other developers,

mmd-osm:
> You were implying to that the *number" gives us licence to break everything at will,

It may be a red flag to other developers that things _may_ change, and we _could_ break things at our own discretion. In fact our Terms of Usage even support that approach. _IX Disclaimer_ says:

> We may add, change, replace, remove, or discontinue the features and functions of the Services, including APIs. It is your responsibility to ensure that calls or requests you make to the Services are compatible with then-current APIs and authentication methods, and that you handle responses properly.

Of course, we're trying to be nice, but we're in no way _obliged_ to do so.

simonpoole:
@mmd-osm given that I'm one of the authors of that text, I can assure you that it simply is designed to  limit the liability of the OSMF if we or somebody else screws up. It is completely orthogonal to any ethical, moral and professional obligation to behave reasonably.

gravitystorm:
>Consider for example the output of https://api.openstreetmap.org/api/0.6/map?bbox=-122.083819,37.4217991,-122.0834335,37.4222209
>Which starts with
> `<osm version="0.6"`

Right, and I would expect the output for any requests made to api/0.7 will start `<osm version="0.7">`. If a client is making requests to the api, then it would expect the response version to match the request version. If it wants a 0.6 response, it should make a request to the 0.6 endpoint.

Now if that response is saved somewhere, then it's useful to know which version of the API originally created that document, so that you know how to parse it. The version in the document is the version of the response. It's not related to some underlying abstract data model, if there is such a thing.

For example, imagine if we keep this underlying abstract data model exactly the same, but change the response format. This could be* changing tags to have "key" and "value" attributes instead of "k" and "v". Then it seems obvious that the response should be `<osm version="0.7" ... <tag key="..." value="...">`, since software will need to know the version of the response to know how to parse it. But this illustrates my point - the version in the document is tied to the **representation** of the data, not the underlying abstract data model, which hasn't changed.

So this PR follows what I expect most other people expect - that the version in the returned document matches the version of the API that requested the document.

(* purely for example; I'm not proposing this)

gravitystorm:
So, trying (desperately) to drag the conversation back to the contents of this PR - does anyone have any comments on handling the indentation of the tests? I really dislike PRs that combine widespread indentation with additional changes to the code, since I find it very hard to spot where the real changes are and which lines are just indented with no other change.

I've been considering whether I should split out the indentation into a separate PR, with a dummy `indent` method like

```ruby
def indent
  yield
end
```
and then make a PR with the methods indented but with no other changes
```diff
  def test_something
-   code
-   code
-   post :foo
-   code
+   indent
+    code
+    code
+    post :foo
+    code
+   end
  end
```

and then in the second PR it will be more obvious which of the lines have real changes. 

```diff
  def test_other_thing
-   indent
+   all_api_versions.each do |version|
      code
      code
-     post :foo
+     post :foo, params => { etc }
      code
    end
end
```

Or am I worrying about this too much, and the PR is fine the way it is now?

iandees:
@gravitystorm If it helps, one of the options in the GitHub pull request view is to ignore whitespace changes. It's in one of the hamburger menus when viewing the diffs.

simonpoole:
@gravitystorm you are essentially making an argument for versioning the representation too. Up to now the abstract underlying data representation has essentially been documented (if you so want) by the XML representation and has been in lock step with that. 

Now clearly we could change the representation in isolation, as your example  suggests, without changing the API nor the underlying abstract data model and this would clearly require changing the API version too, because without that you would not know that you are getting a non-backwards compatible XML doc. But the ITU lies in the direction of doing and supporting that kind of change, so I hope we are not venturing there.

But the other way around does not imply the same thing, for example all the changes that have been made to the API to date work just fine with 0.6 XML documents.  And for example it would be -very- surprising if implementing https://github.com/openstreetmap/openstreetmap-website/issues/2348 and jacking up the API version would suddenly result in output that is not parseable by tools (at least those that do proper validation of their input) that would in principle work just fine if the gratious change to the "representation version" hadn't been made.

gravitystorm:
> you are essentially making an argument for versioning the representation too.

No, I'm explaining that the `version` is the API response version, not a version of the data model. For example, nothing changed in the `trace` responses when we last changed the API version, but the responses to `api/0.6/gpx/nnn` are all `<osm version="0.6">` and not 0.4 or 0.3 or whenever that part of the underlying data model changed. So in API 0.7, if they have the same underlying data model, or even the same representation, they will definitely have 0.7 in their responses.

 So I'll say it again, the `version` in the documents indicates the version of the API used, not the version of any underlying data model.

> Up to now the abstract underlying data representation has essentially been documented (if you so want) by the XML representation and has been in lock step with that.

No, it hasn't. There's been lots of changes to the underlying data model in the last 10 years, none of which have had a change in API version number - because even though the data model has changed, the API request/responses have only changed in backwards compatible ways. For example, we added notes. That was a backwards-incompatible change to the data model (you can't fit notes into a database that doesn't have a table for them, for example) but a backwards-compatible change to the API, hence no version number change.

> And for example it would be -very- surprising if implementing #2348 and jacking up the API version

Nobody has suggested that adding a new backwards-compatible API call needs a new API version - except for you, when you wrote "Without it (this PR) we would have simply added them as a 0.6 feature". So I'm not going to rebut something of your own creation that you are now arguing against. 

If you have another example, that would be helpful.

> if the gratious change to the "representation version" hadn't been made.

I'm not proposing any "gratuitous" changes to the API version. The only reason I'm proposing to change the API version is to allow us to make backwards-incompatible changes that can't otherwise be made. The whole point of this PR is to allow us to run multiple API versions in parallel - it's even in the title! So any tools that can only parse one version can keep using that version. Gratuitous change would be to yank the old version when the new one is deployed, or to bump version numbers for backwards-compatible changes, and I'm proposing neither of those.

gravitystorm:
> @gravitystorm If it helps, one of the options in the GitHub pull request view is to ignore whitespace changes. It's in one of the hamburger menus when viewing the diffs.

Thanks @iandees, that's really helpful. Took me a while to find it, even after you'd told me it was there somewhere!

simonpoole:
> If you have another example, that would be helpful.
OK, you change the error responses to be returned in a structured format and not the currently plain text.

gravitystorm:
@tomhughes I'd love to hear your feedback on this PR - for example on the indentation question, or the overall approach for how to support multiple versions

mmd-osm:
I still very much disagree with the overall approach to tightly couple the OSM XML version returned by the API version to the version number that is part of the URL. I found [this blog post](https://www.mnot.net/blog/2012/12/04/api-evolution.html) raises a valid point here:

_Another conversation that I have sometimes is about how to relate format changes to API versions. In short, they should be completely separate; formats can have lives of their own, and to get the most value out of them, they should do so. It’s fine to say “Version 2 of the API requires the foo resource to support version 5 of the bar format,” of course._

Let's be fair, changing the osm version header from 0.6 to 0.7 would cause some breakage for no good reason, where in reality you're maybe only trying to change the HTTP response codes or produce some nice error messages to be returned by the API. In both cases there's no valid reason at all to also change the OSM XML version number.

I still haven't seen an answer to a point I raised earlier on, how the long term evolution of new version numbers should look like, in particular, for how long old versions will be supported. Without a clear strategy in place, you would keep on adding more and more versions over time, to accommodate for API consumers unable or unwilling to move to a newer version. Today, they don't think about "sunsetting" a version, but in the future they will have to, and they need to be aware of that up front. 

Handling multiple versions in parallel will add some mental strain when working with the code, even when  they share a large part of the same code base. Please keep that in mind so the additional complexity won't kill the code in the long run.

pnorman:
> long old versions will be supported

It'd have to depend on the what is a reasonable time for clients to move to a new version, and how much dev work is involved in maintaining the old version. It might be necessary to make an API version read-only too

gravitystorm:
> to tightly couple the OSM XML header version number to the version number that is part of the URL

But these are, and always have been, the same number. The number in the response *is* the API version number, full stop. There's no such thing as an "OSM XML header version number" as an independent concept as you are describing, one that provides a "format version" that could stay the same as the API behaviour and version number changes.

Consider the "api/X/gpx/1" response. Every time the API version number changed, the number in the trace response changed, even when there's no other change in the XML. Therefore that number is the API version number, and not some independent "XML format version" which just happens to be 0.7 too.

Anyway, this is not a new concept being introduced by this PR, so I'm going to try to avoid debating it further here. Perhaps there is a need for a "OSM XML format version" (or, as previously discussed, an "OSM data model version", which is a third distinct concept), but that would be separate from the API version currently contained in the responses.

> changing the OSM XML version header from 0.6 to 0.7 would cause some breakage for no good reason

Given that we haven't decided what the final list of features are in the next version, it's a bit premature to say that it's going to be for "no good reason"!

One of the points of this PR is to introduce a mechanism that allows us to implement whichever changes we see fit, independently of when they are deployed, by using a "feature flag" concept. That way we can implement a bunch of different improvements, and it gives us flexibility to decide when enough things are implemented to make the deployment of 0.7 "worth it". Until that point, it all lives behind the `Settings.deployed_api_versions` flag.

Also, in reality few things are going to break when 0.7 is released, since they can keep working with 0.6. Again, one of the purposes of this PR is to allow us to run multiple versions side by side. So no clients will talk to 0.7 until the developers make them compatible with 0.7.

For software that is consuming OSM data without interacting with the API (e.g. osm2pgsql), sure, some of them wouldn't understand today what to do with e.g. an osm file saved from a `api/0.7/map` call. But that's fine, since we haven't decided what changes there will be yet! By the time we finally release 0.7, those changes could be either trivial or non-trivial to adapt to, and most software will either be adapted already, or users can get their data from a 0.6 source (e.g. `api/0.6/map`), or run it through an `osm07to06` utility, which for some elements like nodes/ways/relations might be a noop, or for notes/traces/changesets might be non-trivial. Who knows yet?

So we'll need to see at that point whether it's for "no good reason" or not, it's not something I can decide on before we've even started implementing anything.

>  for how long old versions will be supported.

As Paul says, "that depends". I'm not going to debate here how long to keep 0.6 running after 0.7 is released, since at this rate we're never going to see 0.7 in the first place! 

> Handling multiple versions in parallel will add some mental strain when working with the code, even when they share a large part of the same code base.

Absolutely, that's a big factor I considered while implementing this PR. If you have any comments on the approach used in this PR, I'd like to hear them. I'm very much open to suggestions as to how to streamline having multiple versions in the codebase. I'm happy with the current approach but alternative suggestions are valuable.

simonpoole:
Just as a data point: we have 100'000s of files, if not millions, on planet.openstreetmap.org that have nothing to do with the API that reference a version 0.6, not even to mention the ubiquitous PBF format that currently has a "OsmSchema-V0.6" field.

gravitystorm:
> Just as a data point: we have 100'000s of files, if not millions, on planet.openstreetmap.org that have nothing to do with the API that reference a version 0.6

Yes, and there are files there that reference, 0.5, 0.4, and even 0.3, so we've survived previous version changes.

I don't know what your question might be, and I'm not going to guess.

joto:
I have to agree with others here. Changing the version number in the XML and PBF files will break a lot of software, including mine. This software is out there and, even if we change new versions now to accept a new version number, old version of this software will be out there for years. There is absolutely no way we can break the compatibility if it is not absolutely necessary.

@gravitystorm As you mentioned yourself, 0.6 has been around for a very long time, so comparing the situation to 0.5 and earlier versions doesn't make sense. When we changed from 0.5 to 0.6 OSM was much smaller. And writting an `osm07to06` program that would just change the `0.7` to `0.6` so old software can read it, that doesn't sound like a sensible solution either. Of course this would be different if the files actually contain data in a format that needs to be different to transport this data, if we had areas or so. :-)

I think the only way forward here is to decouple API and file format versions. And I don't see why this is a problem really. Keep the `0.6` in the API URLs as "legacy", but the next number will be `1`, `2`, etc. (has already been discussed, we only need "major" version for API versions anyway). The next version of the files can be `0.7` or anything else we like, sometimes new API versions will mean new file versions, sometimes not. Maybe API version `3` will have an extra parameter to support version `0.6` and version `0.7` files or whatever. You could even have new file versions without new API versions if that specific file format isn't created by the API at all but only available as download.

We might also need to think not only about file versions and API versions but versions of the (abstract) data model behind it. I am not sure myself whether they are the same or what exactly their relationship is.

BTW: We already have some variants of the XML file format (JOSM, Overpass, ...) and they are not really compatible so should have gotten some kind of identifier, but that's a totally different issue again.

mmd-osm:
> But these are, and always have been, the same number. The number in the response is the API version number, full stop. There's no such thing as an "OSM XML header version number" as an independent concept as you are describing,

Claiming that both numbers have been the same in the past is technically not incorrect, but to a large extent ignores the reasons why this is so: the last two API changes (0.4 ⟶ 0.5 and 0.5 ⟶ 0.6) involved both API changes and incompatible changes to the OSM XML format. In both cases it was inevitable that the OSM XML header version number had to change along with the API version number change. I don't want to go looking into 0.3 or earlier versions, as they're probably even less relevant for the discussion at hand.

However, based on previous experience, you cannot draw the conclusion that an API change from 0.6 ⟶ 0.7 should be reflected in an otherwise unchanged OSM XML format. We simply didn't have that situation in the past.

gravitystorm:
> The next version of the files can be `0.7` or anything else we like, sometimes new API versions will mean new file versions, sometimes not

I'd like to check what you're proposing here. For example, lets say in the next version of the API, the nodes/ways/relations output is the same, but the format of the notes output is different[1]. Are you proposing

a) all endpoints should continue to output `version=0.6`
b) nodes/ways/relations should output `version=0.6` but notes should output `version=0.7`
c) something else?

Secondly, what if we introduce a new OSM document in the API? Until now, we've just used the existing number. For example, if we add a <diary_entry> API[2]. What number would that then output? Do you think it would be best to

a) start again from 0.1 for fresh OSM document types
b) start from the lowest existing number in use
c) start from the highest existing number in use
d) something else?

[1] This is a genuine plan of mine, to make notes have a description instead of a special first comment, so I'm not making a strawman
[2] Also something I genuinely want to add, but it's not dependent on having a new API version so could come before or after these changes.

simonpoole:
IMHO if you start considering stuff outside of just the core OSM data and API it becomes tricky because there has never been any consensus on carving up the API so that essentially independent parts can actually be run independent of each other (see @zerebubuth suggestions way back). This is particularly noticeable with the Notes API which is really a third party service bolted on to the existing data and user API (btw just to make things complicated notes use a different XML document for on disk storage than what is returned from the API).
 
In any case if we really want to indulge in 2nd system syndrome, I would suggest separating at least the documents and potentially the APIs for core user data, osm data, notes and for any social media functionality. Starting with 0.6 for everything that exists right now.

gravitystorm:
> IMHO if you start considering stuff outside of just the core OSM data and API it becomes tricky

Unfortunately that's what I have to do - everything under `/api/` is affected by these discussions.

What do you consider as the limits of "core OSM data"? I assume at least nodes, ways, relations, and the output of the map call (e.g. the <bounds> element). Would you also include the users and changesets output? Both are referenced from a map response but only by id.

> carving up the API so that essentially independent parts can actually be run independent of each other

I haven't seen much discussion of that, if any. I know there were discussions in the past about running all of `/api/` separately from other non-api stuff, but not about splitting the api itself into multiple projects. But in any case, I've already been told off in this thread for referencing things that happened a while ago, so I won't delve too deeply into historical discussions.

I don't think it would solve any of the problem at hand though. If we split the API up into multiple software projects, it could still be run transparently at `/api/...` with an apache redirects config file. But then  we'll still have all the same discussions about which number to put inside of which responses.

> In any case if we really want to indulge in 2nd system syndrome

Since I think given that splitting things up doesn't solve the topic of what numbers to put in the responses, I'm going to avoid this "2nd system syndrome" and declare it out of scope for this PR.

mmd-osm:
>> carving up the API so that essentially independent parts can actually be run independent of each other
> I haven't seen much discussion of that, if any.

Yes, that conept sounds familiar, see my comment here: https://github.com/openstreetmap/openstreetmap-website/issues/2162#issuecomment-470207138

Quoting myself: _Today, we have so many different API endpoints under the umbrella of the API 0.6 that don't really belong there. Examples could be changeset discussions, gps traces, user data, and probably also map notes. Any incompatible changes here don't really warrant a new OSM API version in a strict sense IMHO, and this further complicates overall API evolution. Maybe that's worth exploring in another issue._

gravitystorm:
So it's clear that this PR has got stuck, and that I need to find some way to move this forward. I've been trying to figure out what that would be for the last few weeks.

This PR was (and still is) just a tiny first step towards general support for multiple versions, never mind any decisions as to how the versions will differ from each other when finally deployed. However, it's reasonable that people want to pitch in with their thoughts on the wider concepts around the API versions. Two recurring themes come up:

* The topic of what the next version number in the url should be (e.g. `0.7` vs `7` and related concepts of semantic versioning for the API)
* The topic of what version number(s) should be in the responses

The latter is particularly what has got this PR stuck. We can't have a fully informed decision on what number(s) to put into the responses until we see more examples of what's actually going to change. But we can't code any of the specific changes until I have implemented general support for multiple versions. And this is just the first in ~20 PRs that will be needed for the general support, never mind the ~?? PRs which will be about actually changing any API responses.

So my proposal is:

* To create a new issue discussing what the next version number in the urls should be
* To create a new issue for continuing the discussion on version numbers in the responses
* Meanwhile, to open a fresh PR to replace this one, with conflicts resolved
* To continue to work on the general support for multiple versions
* To then work on specific changes to the API

I think this will allow us to unblock the general development work while the wider issues are discussed more fully.

tomhughes:
I do have this in my queue for a technical review as I think you requested a while ago, but I kind of held of because it sort of blew up again.

I don't expect to find any major technical issues though - if support for multiple API versions in parallel is necessary then I'm sure this is broadly speaking the correct approach.

You and I disagree on whether support for multiple API versions in parallel is required - actually that's not quite true as I have no problem with the idea in principle but I just don't think there is any realistic way we can achieve it.

Separately a lot of people think that the object model should be versioned separately to the API version, which you don't feel is necessary, but which I tend to agree with I think. In fact that ties in very much with my thoughts about multiple API versions in that I think there is no problem with multiple API versions where the object model remains the same but what people really want to do is to enhance the object model and I very doubtful that we can support multiple versions of that in parallel.

gravitystorm:
> enhance the object model and I very doubtful that we can support multiple versions of that in parallel

I agree there are changes that will be hard or impossible to support in parallel. But that shouldn't stop us from making all the other changes that can.

gravitystorm:
Closing this PR as per https://github.com/openstreetmap/openstreetmap-website/pull/2353#issuecomment-561601385

mmd-osm:
In their recent S2S meeting the OSMF Board discussed how they can support rewriting of the API  ("_Supporting API: confirm goals and identify next steps - supporting Andy Allen’s rewrite of the API (Paul, Joost, Guillaume)_")  -> https://wiki.osmfoundation.org/wiki/Board/Minutes/2020-10-S2S

Maybe this is referring to some previous @gravitystorm blog post on supporting multiple versions? Anyone have some insights what this is all about?

simonpoole:
> ... Anyone have some insights what this is all about?

The board doing random stuff without coordination with anybody else? It's the norm, see also a budget without asking the WGs for their requirements and many other things.

woodpeck:
> The board doing random stuff without coordination with anybody else?

Don't get carried away. You know as well as I do that it could very well be a harmless question discussed internally ("can/should we use resources on this") and only after a preliminary nod would the board approach third parties.

mmd-osm:
In a first step, I really wanted to get some better understanding what the intended scope of those requirements were, irrespective of effort, feasibility, etc.

Agreed, communication could be a bit more transparent at times (otherwise I wouldn't be asking those questions), but this issue isn't a good place for that discussion.

gravitystorm:
> In a first step, I really wanted to get some better understanding what the intended scope of those requirements were, irrespective of effort, feasibility, etc.

Just for clarity - I haven't spoken recently with any board members on this topic, so I don't have any insights as to this particular conversation or what was intended (or even what they mean by 'rewrite of the API'). But I am always happy to have more help, either on the narrow topic of the API or the wider topic of the everything else covered by this repo.

grischard:
Frederik supposes correctly. We're seeing how important and Andy's work is, how progress there makes the work of others easier, and discussed if and how we could support it. There was a confusion between rails port and API, and we ended up chatting mostly about the rails port. We haven't decided to fund any projects, or indeed even had a chat with Andy recently (hi!).

We're currently running the microgrants, three software Grant pilot projects (osm2pgsql, Nominatim, Potlatch) and paying Quincy to work on iD full-time. We intend experience from those to support a reactivated EWG who would be in charge of managing the projects and allocating a budget.

We're very interested in hearing from anyone who would like their projects to be supported by the OSMF, or anyone who would be interested in joining EWG.

mmd-osm:
@grischard : thank you for the clarification, the meeting minutes make a whole lot more sense now.

PR #2354, opened 2019-08-21
Overpass query bbox to use map precision

mmd-osm:
15 decimal places for bbox lat/lon values seemed a bit excessive.


tomhughes:
Well sure there may not be any good reason to send 15 but is there any good reason not to?

Does adding this extra code solve some technical issue? I don't believe that URL is visible anywhere so I don't think there's any UX concern here is there?

mmd-osm:
It's really a matter of consistency, as we're already using a limited number of digits for `(around:xx,lat,lon)` in the same query.

PR #2358, opened 2019-08-24
promote yourself to superuser

openbrian:
If your database already has this user you'll need to promote the user to superuser.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -148,13 +148,22 @@ instructions below as appropriate.
 
 ### PostgreSQL account setup
 
-We need to create a PostgreSQL role (i.e. user account) for your current user, and it needs to be a superuser so that we can create more databases.
+We need to create a PostgreSQL role (i.e. user account) for yourself for development, and it needs to be a superuser so that we can create more databases.  Superuser is also needed to run automated tests which will CREATE EXTENSIONs in the database.
 
 ```
 sudo -u postgres -i
 createuser -s <username>
 exit
 ```
+
+If you already have an account in your PostgreSQL database, promote yourself to a superuser.
+
+```
+sudo -u postgres -i
+psql template1 -c "alter role <username> with superuser"
+exit
+```
+
 
 ### Create the databases
 


tomhughes:
No, deciding whether to promote your main development user to a database user is a personal decision that should be made with an understanding of the issues involved and not one we should encourage people to make unthinkingly.

woodpeck:
Perhaps we should then at least write something like "note, some automated tests that try to execute CREATE EXTENSION commands on the database will fail when not run as a privileged database user"?

tomhughes:
Well we need to quantify the actual problem before we can formulate a solution - in other words under what circumstances in the permission required.

There probably aren't any good solutions sadly.

openbrian:
Before I take a second stab at this can you tell me what you mean by "your current user".  For me that's brian@localhost on a debian system.  The problem was when I ran automated tests, they failed with the CREATE EXTENSION command.

Perhaps the issue is with these words "your current user".  Instead it should be "a newly created user that will own the openstreetmap database".  Thoughts?

Side note, you don't need superuser "so that we can create more databases".  That can be granted to a user.

tomhughes:
Well the general point is that you shouldn't be throwing around postgres superuser privileges without understanding what you are doing.

Ideally there would be a single step that you could execute as the postgres user but as I say I'm not sure if that's actually possible.

It may be that we should recommend an initial run of `rake test:db` as a database superuser to populate the test database but as I say it needs some research to see if that is good enough or under what circumstances rails may try and recreate the test database from scratch.

openbrian:
I found a way around this issue.  It's to use a template database.  PG 9.5 add the IS_TEMPLATE option to CREATE DATABASE.  If true, anyone with CREATEDB privilege can clone it.  If the template has the btree_gist extension, the new database will also have it.

openbrian:
I've done just a little research and found that rails does have some support for this.  See activerecord-5.2.3/lib/active_record/connection_adapters/postgresql/schema_statements.rb, line 30.  Though I'm not sure how easy this is to use.

PR #2367, opened 2019-09-04
Update some old information

ismailarilik:


gravitystorm:
I don't even understand what those sentences mean, never mind whether they are still relevant. 

Can somebody explain why we should keep that paragraph, or whether we should delete it?

tomhughes:
So basically it's all about how vagrant is a complete shit show...

A box is a VM image and https://app.vagrantup.com/ is a site where you can search for suitable base images to use but the complication is that vagrant supports multiple backend virtualisation systems (virtualbox, libvirt, parallels, etc) and not all images support all the backends.

In particular the official Ubuntu images published by Canonical (eg https://app.vagrantup.com/ubuntu/boxes/bionic64 recommended here) only support virtualbox, so if you're using some other virtualisation system then you're out of luck with those.

There are alternatives (for example https://app.vagrantup.com/generic/boxes/ubuntu1804) which support a wider range of backends.

gravitystorm:
So it sounds like this bit of the documentation is out of date then, and needs changing rather than just the updating in this PR.

If I understand things correctly (and that's far from guaranteed) then we could either:

a) change the documentation to say that if you don't want to use the virtualbox provider, other boxes are available at (insert link), or
b) change the box used at https://github.com/openstreetmap/openstreetmap-website/blob/5dfa747823e0940b86d8c70423269b57cffe10ed/Vagrantfile#L7 to be one that supports the other providers, like `generic/ubuntu1804`, and then remove this paragraph from the docs entirely.

Are there more options? And any preferences between options a and b?

tomhughes:
Well (a) is basically what it is saying I think? It gives the URL to search for alternates?

gravitystorm:
> Note than until there are suitable _bionic64_ vagrant boxes for other providers

But there *are* suitable vagrant boxes for other providers, right? If so, this sentence makes little sense.

tomhughes:
True that is not ideal phrasing...

gravitystorm:
Closing in favour of #2400, which is a more thorough update.

PR #2373, opened 2019-09-22
Add link to my messages to users/show navigation

tordans:
The my messages page can be accessed via the user dropdown, but not via the user navigation on my logged in user-profile page. This adds the link when logged in to have a consistent access to all my sub-pages.

# Current header

Bildschirmfoto 2019-09-22 um 19 47 49

# New header

Bildschirmfoto 2019-09-22 um 19 47 57

--- 

# Open questions

- [ ] The new navi uses the same translation string as the dropdown (as do all those links). The german translation for the `.my messages` only says "Nachrichten" (EN: Messages). This is consistent in the dropdown, but not on the user-page. Suggestion: Duplicate the translations in all languages so the dropdown can have it's own variation. Right now the dropdown forces the translation with `<%= t("users.show.my messages") %>`

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -7,6 +7,10 @@
         <!-- Displaying user's own profile page to themself -->
         <ul class='secondary-actions clearfix'>
           <li>
+            <%= link_to t(".my messages"), inbox_messages_path %>
+            <span class='count-number'><%= number_with_delimiter(current_user.new_messages.size) %></span>
+          </li>
+          <li>
             <%= link_to t(".my edits"), :controller => "changesets", :action => "index", :display_name => current_user.display_name %>
             <span class='count-number'><%= number_with_delimiter(current_user.changesets.size) %></span>
           </li>


tomhughes:
I think we're fairly reluctant to add more things to that line to be honest - it's already far too long really.

tordans:
For me, the simplicity of the UI is less in the count of elements but more in finding the elements where I expect them. I expect my messages to be part of the my-* links. I find myself confused about what I find in the dropdown and what in the user profile navigation.

harry-wood:
Been pondering this. Maybe the profile page should have a second navigation element, maybe a second strip of links to include "My messages", "My Settings", and "Logout" (doubling up on links available in the top-right drop-down). Bit cluttering maybe, but it would be a rationalisation since "My settings" is an odd one out at the moment. It's the only one currently doubled-up. It's also the only one which is not a list of things which form part of the profile.

A related problem: when viewing somebody else's profile this same strip of links contains  "lists of things which form part of the profile" mixed with three actions. So here again a second strip of links might make it tidier.

gravitystorm:
> Maybe the profile page should have a second navigation element, maybe a second strip of links

I think a better a

pnorman:
> I think a better a

It looks like this message was cut off

gravitystorm:
I honestly can't remember what I was writing here, nor why I closed the pull request - so reopening seems like the best bet.

Issue #2372, opened 2019-09-22
Installation: couldn't find file 'settings.local.yml'


tordans:
I am following https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md#macosx and got to "Running the tests" https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md#running-the-tests which tells me:

```
Error:
BrowseControllerTest#test_read_node:
ActionView::Template::Error: couldn't find file 'settings.local.yml'
Checked in these paths: 
```

So following https://github.com/openstreetmap/openstreetmap-website/blob/c162c065c6f6db5dc584599f5639b296b627a165/config/initializers/config.rb#L9 I added an empty file in `config/settings.local.yml` which resolved the issue.

Should I check this file in or add a line to create it to the installation instructions?

tomhughes:
Yes this is a known issue, but it's not clear what the best solution is - what you did certainly works but ideally it shouldn't be necessary and we're still looking for a way to resolve that. There are already tickets about this I believe.

HolgerJeromin:
> There are already tickets about this I believe.

ref #2185

Issue #2345, opened 2019-08-08
INSTALL.md has minor quirks


YuriiNskyi:
I'm trying to install `openstreetmap-website` on clean `Ubuntu 18.04 LTS`.

After spending some time on solving [this issue](https://github.com/openstreetmap/openstreetmap-website/issues/2295), by executing these commands:

```
sudo apt-get -y install npm
sudo npm install yarn -g
```

here are the new problems:

First has occurred when executing `bundle install`:

>Traceback (most recent call last):
	1: from /usr/local/bin/bundle:23:in `<main>'
/usr/local/bin/bundle:23:in `load': cannot load such file -- /usr/share/rubygems-integration/all/gems/bundler-1.16.1/exe/bundle (LoadError)

After googling, this problem is simply resolves:

```
sudo gem update --system
sudo gem uninstall bundler
sudo gem install bundle
```

The next steps works as needed, until we stops at `bundle exec rails server`:

> couldn't find file 'settings.local.yml'

This is happily resolved by executing `cp config/settings.yml config/settings.local.yml`.

Finally, the site works.

There are multiple proposed solutions:

1. Add section with Yarn installation into [INSTALL.md](https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md).
2. Do something with bundler (it's up to you, I'm not a professional Ruby developer)
3. Do not require `settings.local.yml` file as, I think, it is only needed to override `settings.yml`. What if I don't need any overrides and totally agreed with a default? 
4. Do something with these errors, while running the server:

>pngcrush worker: `pngcrush` not found; please provide proper binary or disable this worker (--no-pngcrush argument or `:pngcrush => false` through options)
advpng worker: `advpng` not found; please provide proper binary or disable this worker (--no-advpng argument or `:advpng => false` through options)
optipng worker: `optipng` not found; please provide proper binary or disable this worker (--no-optipng argument or `:optipng => false` through options)
pngquant worker: `pngquant` not found; please provide proper binary or disable this worker (--no-pngquant argument or `:pngquant => false` through options)
jhead worker: `jhead` not found, `jpegtran` not found; please provide proper binary or disable this worker (--no-jhead argument or `:jhead => false` through options)
jpegoptim worker: `jpegoptim` not found; please provide proper binary or disable this worker (--no-jpegoptim argument or `:jpegoptim => false` through options)
[...]

Issue #2292, opened 2019-06-30
Error in bundle exec rake db:create


caduguedess:
I'm trying to install OSM-Website in a Ubuntu Bash in Windows 10 and when I run the command:
`bundle exec rake db:create`
It gives me error:
![dbcreate](https://user-images.githubusercontent.com/51020710/60390993-d8003e80-9aba-11e9-950c-37398b64a5a4.png)
There is also a txt file uploaded for the full error with the --trace command.
[dbcreate.txt](https://github.com/openstreetmap/openstreetmap-website/files/3342427/dbcreate.txt)
Thanks for your time.

tomhughes:
Your postgres server isn't running so you will need to consult the poistgres documentation on how to fix that.

You environment is not something that any of us have any experience in I'm afraid - we know Ubuntu but not any special issues with running it in the linux subsystem on Windows.

Normally installing postgres would have created and started an instance but it looks like that hasn't happened for you - you can try `sudo systemctl start postgresql` but if that doesn't help then I really don't know and you may need to start consulting postgres resources about whether there are any known issues running in your slightly unusual environment.

caduguedess:
I just installed Ubuntu 18.04 in my computer, erase the Windows 10, so I will try to install in the most ideal way.
I noticed that when I installed postgre with the following code:
`sudo apt-get install postgresql postgresql-contrib libpq-dev postgresql-server-dev-all`
I think something got wrong, because I got a red message in the Terminal.
![postgre](https://user-images.githubusercontent.com/51020710/60510870-e7dc7600-9ca6-11e9-9936-dbb1c28c33c8.png)
Maybe I have to fix it before continue installing the other parts?

tomhughes:
I honest can't remember if that is normal or not but I wouldn't worry about it unless you encounter other problems.

gravitystorm:
I think it's fine, but you might need to run `sudo service postgresql start`.

tomhughes:
Oh yes I sure that (or, to be modern `sudo systemctl start postgresql`) is the most that will be needed but I suspect that was probably already done and that message is just an implementation detail of it reporting status after creating the database and before starting it.

caduguedess:
> Oh yes I sure that (or, to be modern `sudo systemctl start postgresql`) is the most that will be needed but I suspect that was probably already done and that message is just an implementation detail of it reporting status after creating the database and before starting it.

[...]

Issue #2185, opened 2019-03-20
Sprockets vs optional config files


gravitystorm:
In [app/assets/javascripts/osm.js.erb](https://github.com/openstreetmap/openstreetmap-website/blob/53cc1d2d11d358c171ee5beab55a3d20b79e9dab/app/assets/javascripts/osm.js.erb) we need to tell sprockets to recompile the output whenever the settings change. To do this we mark that file as depending on `config/settings.yml` and `config/settings.local.yml`.

However, this throws up two problems:

* `settings.local.yml` is supposed to be optional, but Sprockets throws an error if the file can't be found.
* The config gem also supports a [bunch of other files](https://github.com/railsconfig/config/blob/3a28b646a69c16f87d6539b4e1c4189aecb34c84/lib/config.rb#L60-L70), which should also trigger a recompile if they exist and then get changed.

I've checked through the sprockets documentation, but I can't see any obvious guidance for approaches to this situation.

Potential solutions include:

* Remove the dependency on `settings.local.yml`. This will make the 'out-of-the-box' situation better for developers, but could lead to problems later on when changes to that file are not reflected in the compiled javascript (without manual asset recompilation).
* Ask the upstream sprockets developers what they think about depending on optional files, or if they have any suggestions.
* Create an initializer which outputs the computed settings to a file, and depend on that output file instead.

gravitystorm:
For anyone who gets `ActionView::Template::Error: couldn't find file 'settings.local.yml'` or similar when first setting up your environment, then this issue is the root cause. You need to create an empty file in `config/settings.local.yml`, for example by running:

```bash
$ touch config/settings.local.yml
```

mmd-osm:
How about:
* Create an empty settings.local.yml file in config.rb, in case no local settings have been created
* If that doesn't work, print out an error message and exit, asking the dev to create an empty file settings.local.yml file themselves.

mmd-osm:
I'm not seeing any error messages on Rails 8 anyone in case config/settings.local.yml is missing. Is this working as expected? Maybe we could remove the respective section in the readme file now.

pablobm:
For a new datapoint: I cloned the repo today and can see the errors if I run the test suite without the `config/settings.local.yml` file. Eg:

```
$ docker compose run --rm web bundle exec rails test:all
[+] Creating 1/1
 ✔ Container openstreetmap-website-db-1  Running                                                                                                                                                                                       0.0s
[...]

PR #2374, opened 2019-09-23
Add reference to settings.local.yml-file in installation.md

tordans:
There seems to be an easy solution that does not have negative side effects. So I suggest to add it to the installation instructions until https://github.com/openstreetmap/openstreetmap-website/issues/2185 has a more permanent solution. 

Once https://github.com/openstreetmap/openstreetmap-website/issues/2185 is solved, this commit should be reverted.

The same issue came up in https://github.com/openstreetmap/openstreetmap-website/issues/2372, https://github.com/openstreetmap/openstreetmap-website/issues/2292#issuecomment-508531378, https://github.com/openstreetmap/openstreetmap-website/issues/2345#issue-478341245

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -124,6 +124,14 @@ We use [Yarn](https://yarnpkg.com/) to manage the Node.js modules required for t
 bundle exec rake yarn:install
 ```
 
+## Prepare local settings file
+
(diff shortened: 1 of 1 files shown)


gravitystorm:
Thanks!

Issue #2380, opened 2019-10-02
Use created_at instead of updated_at for searching notes through the API


ENT8R:
Is there a specific reason why in
https://github.com/openstreetmap/openstreetmap-website/blob/83983bfc6a31c2da0af6f18137cd31ba40519191/app/controllers/api/notes_controller.rb#L301
the notes (via the `/api/0.6/notes/search` call) are sorted by their `updated_at` value?
Actually I can't really think of any reason right now as sorting it by `created_at` sounds a lot more plausible to me and that was actually the behaviour I would expect from that API call...

If there is a reason why `updated_at` is rather used than `created_at`, maybe it would be feasible to add another parameter specifying the wanted sorting value?

tomhughes:
It's done so that notes which have been recently commented on rise to the top - basically so that the most recently active notes are returned.

ENT8R:
I understand the use case for the main openstreetmap website, but there might be potentially other data consumers which rather want only to get notes sorted by their created_at value. 
So, would it maybe feasible to add another parameter to the API call specifying whether to sort the notes by their created_at value rather than updated_at?

tomhughes:
Well of course it's feasible - patches welcome.

PR #2381, opened 2019-10-03
Sort and order notes

ENT8R:
Adds two new parameters to the notes search API to be able to
- specify the value the notes should be sorted by (either `created_at` or `updated_at`, `updated_at` remains the default value)
- specify the order of the returned notes (either `ASC` or `DESC`, `DESC` remains the default)

A new call to the notes search API may look like this:
[/api/0.6/notes/search?**sort_by=created_at&order_by=ASC**](https://api.openstreetmap.org/api/0.6/notes/search?sort_by=created_at&order_by=ASC)

See also #2380 for the reason of this pull request.

### DIFF
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -294,10 +294,14 @@ module Api
           raise OSM::APIBadUserInput, "Date #{params[:to]} is in a wrong format"
         end
 
-        @notes = params[:sort] == "created_at" ? @notes.where(:created_at => from..to) : @notes.where(:updated_at => from..to)
+        @notes = if params[:sort] == "updated_at"
+                   @notes.where(:updated_at => from..to)
+                 else
+                   @notes.where(:created_at => from..to)
+                 end
       end
 
-      # Find the notes we want to return
+      # Choose the sort order
       @notes = if params[:sort] == "created_at"
                  if params[:order] == "oldest"
                    @notes.order("created_at ASC")
@@ -312,6 +316,7 @@ module Api
                  end
                end
 
+      # Find the notes we want to return
       @notes = @notes.distinct.limit(result_limit).preload(:comments)
 
       # Render the result
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -294,11 +294,14 @@ module Api
           raise OSM::APIBadUserInput, "Date #{params[:to]} is in a wrong format"
         end
 
-        @notes = @notes.where(:created_at => from..to)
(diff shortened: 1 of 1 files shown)


tomhughes:
There's no need to expose raw field names or SQL terms in the API like that - just use `order=newest` and `order=oldest` or something.

Also try to avoid text interpolation into queries like that unless it's absolutely necessary - yes I know it's safe here, but I'd rather just see a conditional block that chains on the right order statement to the query for each possible sort order.

ENT8R:
> I'd rather just see a conditional block that chains on the right order statement to the query for each possible sort order.

That was actually my first implementation, but then I thought, that this looks a little bit inflated. 😄
So now I changed it again to the old version...

app/controllers/api/notes_controller.rb
```diff
@@ -294,11 +294,25 @@ def search
...
+                 end
+               else
+                 if params[:order] == "oldest"
+                   @notes.order("updated_at ASC")
+                 else
+                   @notes.order("updated_at DESC")
+                 end
+               end
+
+      @notes = @notes.distinct.limit(result_limit).preload(:comments)
```
ENT8R: BTW: I also added `.distinct` which was missing before but might also be useful because returning the same note multiple times is not really helpful...

rugk: This does fix a big problem with duplicate notes, see https://github.com/ENT8R/NotesReview/issues/43

ENT8R:
@tomhughes Just wondering, is there still something which has to be done in order for this pull request to be merged?

app/controllers/api/notes_controller.rb
```diff
@@ -294,11 +294,25 @@ def search
           raise OSM::APIBadUserInput, "Date #{params[:to]} is in a wrong format"
         end
 
-        @notes = @notes.where(:created_at => from..to)
+        @notes = params[:sort] == "created_at" ? @notes.where(:created_at => from..to) : @notes.where(:updated_at => from..to)
```
tomhughes: This will change the behaviour of existing API calls - you should make `created_at` the default to preserve the existing behaviour when no sort is specified.

I'd also prefer it if this was written like the sort order code with a multiline if rather than using the ternary operator, which is much less readable.

app/controllers/api/notes_controller.rb
```diff
@@ -294,11 +294,25 @@ def search
...
       end
 
       # Find the notes we want to return
```
tomhughes: This comment should be moved down to stay with the line it was originally attached to as it's incorrect where it is now.

If you want a comment here then it should say `Chose the sort order` or something.

app/controllers/api/notes_controller.rb
ENT8R: You're right. I didn't notice that one...

PR #2383, opened 2019-10-04
Remove need for shared library PostgreSQL functions

zerebubuth:
This repo contains code for three functions to be loaded into PostgreSQL as a shared library:

1. `maptile_for_point`, which is used only by the `/changes` API call. This API call is little-used and IMHO should be deprecated and removed. However, even now it's hardly on the hot path for most development activities.
2. `tile_for_point`, which is used only in migrations. At this point, it seems unlikely that anyone will be doing a migration on existing data which would call this function (most developers will be running migrations on an empty database, to set it up).
3. `xid_to_int4`, which is only used for replication using Osmosis and isn't used in the Rails code at all. Hopefully this will be replaced Real Soon Now, but until then it's a quite advanced feature that most developers won't need.

Therefore, this patch proposes to replace the above three shared library functions with SQL implementations of the first two. These are _much_ slower - by a factor of about 30x, however this makes no difference when they're run on a completely empty database. In return, we're able to drop a dependency on the PostgreSQL server development package, and clean a few lines out of the installation instructions.

It's still possible to make and install the shared library functions, and I've included instructions about how to do that - although it shouldn't be necessary for the vast majority of `openstreetmap-website` developers.

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -131,5 +131,5 @@ If you want to deploy The Rails Port for production use, you'll need to make a f
 * The included version of the GPX importer is slow and/or completely inoperable. You should consider using [the high-speed GPX importer](https://git.openstreetmap.org/gpx-import.git/).
 * Make sure you generate the i18n files and precompile the production assets: `RAILS_ENV=production rake i18n:js:export assets:precompile`
 * Make sure the web server user as well as the rails user can read, write and create directories in `tmp/`.
-* If you want to use diff replication then you will need to install the shared library versions of the special SQL functions (see the bottom of [INSTALL.md](INSTALL.md)).
+* If you want to use diff replication then you will need to install the shared library special SQL functions for the `xid_to_int4` function, for which there is no pure SQL alternative. (See the bottom of [INSTALL.md](INSTALL.md))
 * If you expect to serve a lot of `/changes` API calls, then you might also want to install the shared library versions of the SQL functions.
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -131,3 +131,5 @@ If you want to deploy The Rails Port for production use, you'll need to make a f
 * The included version of the GPX importer is slow and/or completely inoperable. You should consider using [the high-speed GPX importer](https://git.openstreetmap.org/gpx-import.git/).
 * Make sure you generate the i18n files and precompile the production assets: `RAILS_ENV=production rake i18n:js:export assets:precompile`
 * Make sure the web server user as well as the rails user can read, write and create directories in `tmp/`.
+* If you want to use diff replication then you will need to install the shared library versions of the special SQL functions (see the bottom of [INSTALL.md](INSTALL.md)).
+* If you expect to serve a lot of `/changes` API calls, then you might also want to install the shared library versions of the SQL functions.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -31,8 +31,8 @@ sudo apt-get update
 sudo apt-get install ruby2.5 libruby2.5 ruby2.5-dev bundler \
                      libmagickwand-dev libxml2-dev libxslt1-dev nodejs \
                      apache2 apache2-dev build-essential git-core phantomjs \
-                     postgresql postgresql-contrib libpq-dev postgresql-server-dev-all \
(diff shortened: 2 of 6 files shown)


mmd-osm:
Good idea, I was about to propose the same, but haven't gotten to it. :+1:

INSTALL.md
```diff
@@ -174,20 +174,12 @@ psql -d openstreetmap -c "CREATE EXTENSION btree_gist"
 
 ### PostgreSQL Functions
 
-We need to install special functions into the PostgreSQL databases, and these are provided by a library that needs compiling first.
+There are special database functions required by a (little-used) API call, the migrations and diff replication. The former two are provided as *either* pure SQL functions or a compiled shared library. It is recommended that you start with the pure SQL versions, as described below, and only install the compiled shared library if you are running a production server making a lot of `/changes` API calls or need the diff replication functionality.
```
gravitystorm: I would suggest rephrasing this to avoid the user from having to read, understand, or make a decision at this point. So for example, I think it's enough to say:

"We need to install some special functions into the PostgreSQL databases"

... and leave it at that. The explanation of the advantages of the compiled shared library can then live in the later section of the docs, and if someone gets that far and wants to use them, it's only a total of 3 lines of code to copy+paste (1 here for installing, 2 later for removing) so no big deal.

INSTALL.md
```diff
@@ -225,3 +217,42 @@ Note that the OSM map tiles you see aren't created from your local database - th
 # Configuration
 
 After installing this software, you may need to carry out some [configuration steps](CONFIGURE.md), depending on your tasks.
+
+# Installing compiled shared library database functions
```
gravitystorm: Add `(optional)` to this title to highlight (and reassure) that it's not necessary for most people.

INSTALL.md
```diff
@@ -225,3 +217,42 @@ Note that the OSM map tiles you see aren't created from your local database - th
...
+
+You probably only need to do this if you are running a large, production instance of openstreetmap-website.
```
gravitystorm: Additional explanation from above can go here

INSTALL.md
zerebubuth: Good points, thanks. I've made those changes in 2423d86. Please let me know what you think.

gravitystorm: Perfect

CONFIGURE.md
```diff
@@ -131,3 +131,5 @@ If you want to deploy The Rails Port for production use, you'll need to make a f
 * The included version of the GPX importer is slow and/or completely inoperable. You should consider using [the high-speed GPX importer](https://git.openstreetmap.org/gpx-import.git/).
 * Make sure you generate the i18n files and precompile the production assets: `RAILS_ENV=production rake i18n:js:export assets:precompile`
 * Make sure the web server user as well as the rails user can read, write and create directories in `tmp/`.
+* If you want to use diff replication then you will need to install the shared library versions of the special SQL functions (see the bottom of [INSTALL.md](INSTALL.md)).
```
mmd-osm: This sounds a bit as if there was a non-shared library version of the xid_to_int4 function available, which isn't the case. To make this a bit clearer maybe state that diff replication depends on a function xid_to_int4, which is only being made available as a shared library. Additional bonus information would be to state that this function is only supported up to and including Postgres 9.5.

tomhughes: Well the point is that the `xid_to_int4` function is something that as far as I know only we use and we would simply arrange to carry on using the 3GL version without it being part of the standard installation.

mmd-osm: I wasn't even aware of a 3GL version, but you're right: osmosis ships one: https://github.com/openstreetmap/osmosis/blob/master/package/script/contrib/apidb_0.6_osmosis_xid_indexing.sql

zerebubuth: To try and avoid confusion, I've changed that line in 7fb5bcaf3 to explicitly say there's no pure SQL version of `xid_to_int4`.

grischard:
Just out of curiosity: what did `/changes` do?

tomhughes:
It gets a list of tiles at a given zoom level that have changed in a given time period - by default if does z12 for the last hour but both can be changed.

It only looks at nodes though so it won't notice changes to ways or relations that cross a tile.

PR #2386, opened 2019-10-11
Make 'back' button not look disabled.

das-g:
When viewing a message in the inbox, the "back" button looks paler than the other buttons below the message. This can be mistaken for that button to be disabled, and thus there being no way (other than the browser's own "back" button or accessing it anew from the logged-in-user drop down menu) to get back to the list of messages in the inbox. (Actually, the button is just "`deemphasize`d", whatever that's supposed to mean.)

This change makes it no longer deemphazized, thus displaying it in the same style as the other buttons there.

### DIFF
--- a/app/views/messages/show.html.erb
+++ b/app/views/messages/show.html.erb
@@ -36,5 +36,5 @@
 
 <% end %>
 
-  <%= link_to t(".back"), outbox_messages_path, :class => "button deemphasize" %>
+  <%= link_to t(".back"), outbox_messages_path, :class => "button" %>
   </div>


gravitystorm:
We use "deemphasized" buttons in a few places, but many of them are for moderator actions so you don't see them. This is one of the few examples where they are visible to normal users.

I think what we want is to have (at least) three different button styles, including
* main buttons, for whatever the primary purpose of the form is
* secondary buttons, for other form actions that are less likely, e.g. hiding comments
* disabled buttons, for obvious reasons

I tend to take design guidance from the [bootstrap project](https://getbootstrap.com), since it's fairly modern and a lot of the web is built on it or variations of it. They have [multiple different approaches to buttons](https://getbootstrap.com/docs/4.3/components/buttons/), including:

* button styles (primary, secondary, success etc).
* Outline buttons, as an alternative
* Disabled buttons
* and more

I think the disabled button here is the problem. Although we've used a "lighter less saturated blue" to mean "deemphasized" (or "secondary"), bootstrap uses "lighter less saturated" to mean "disabled". And you'll see a lot more bootstrap sites than OSM-themed sites around the web!

So we need to move away from "lighter less saturated" for secondary buttons.

gravitystorm:
So that was the general point, but I want to make specific comments about this situation. In general, it's a very bad idea to have things that look like buttons but aren't really buttons i.e. they don't change state. The "Back" link in question is really just a link, and so it shouldn't be a button in the first place, never mind discussing whether it's emphasised or not.

Of course, from a design point of view it's nice if they all line up, have the same margins, and so on. It's very common at the bottom of a form to have a cancel "button", even when it's not a form submission, to make life easy and make things look consistent. (OSM sadly lacks these in most places).

Going back to bootstrap - they have a standard "link" button. It works well in my experience, and it's what we need here, if you ask me. 

So we should create a button styling that works in the same way as bootstrap's "btn btn-lnk", and use it here. Meanwhile we should change the "demphasized" button type to avoid making it look like a disabled button too.

gravitystorm:
Time has passed, and we're now moving most things to bootstrap. I've made an alternative version of this PR at #2688 using the link buttons that I described in this ticket, so I'm closing this PR in favour of the approach in #2688.

PR #2403, opened 2019-10-18
Fix installation instructions for Mac OS X

### LINKED ISSUE #2371: Installation: `Error: invalid option: --with-xml2-config`
I am following https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md#macosx

> Install libxml2: brew install libxml2 --with-xml2-config

Which causes the error below.
Could it be, that this option is only avaliable for `gem install libxml-ruby`?

The issue is solved now, but maybe someone who understands more about this could update the docs?

--- 

Btw, if someone googels this error, for me I could install `libxml-ruby 3.1.0` after doing https://stackoverflow.com/a/52624622.

---

```terminal
brew install libxml2 --with-xml2-config
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (homebrew/cask).
No changes to formulae.

Usage: brew install [options] formula

Install formula. Additional options specific to formula may be appended to
the command.

Unless HOMEBREW_NO_INSTALL_CLEANUP is set, brew cleanup will then be run for
the installed formulae or, every 30 days, for all formulae.

    -d, --debug                      If brewing fails, open an interactive
                                     debugging session with access to IRB or a
                                     shell inside the temporary build directory.
        --env                        If std is passed, use the standard build
                                     environment instead of superenv. If super
                                     is passed, use superenv even if the formula
                                     specifies the standard build environment.
        --ignore-dependencies        An unsupported Homebrew development flag to
                                     skip installing any dependencies of any
                                     kind. If the dependencies are not already
                                     present, the formula will have issues. If
                                     you're not developing Homebrew, consider
                                     adjusting your PATH rather than using this
                                     flag.
        --only-dependencies          Install the dependencies with specified
                                     options but do not install the formula
                                     itself.
        --cc                         Attempt to compile using the specified
                                     compiler, which should be the name of the
                                     compiler's executable, e.g. gcc-7 for GCC
                                     7. In order to use LLVM's clang, specify
                                     llvm_clang. To use the Apple-provided
                                     clang, specify clang. This option will
                                     only accept compilers that are provided by
                                     Homebrew or bundled with macOS. Please do
                                     not file issues if you encounter errors
                                     while using this option.
    -s, --build-from-source          Compile formula from source even if a
                                     bottle is provided. Dependencies will still
                                     be installed from bottles if they are
                                     available.
        --force-bottle               Install from a bottle if it exists for the
                                     current or newest version of macOS, even if
                                     it would not normally be used for
                                     installation.
        --include-test               Install testing dependencies required to
                                     run brew test formula.
        --devel                      If formula defines it, install the
                                     development version.
        --HEAD                       If formula defines it, install the HEAD
                                     version, aka. master, trunk, unstable.
        --fetch-HEAD                 Fetch the upstream repository to detect if
                                     the HEAD installation of the formula is
                                     outdated. Otherwise, the repository's HEAD
                                     will only be checked for updates when a new
                                     stable or development version has been
                                     released.
        --keep-tmp                   Retain the temporary files created during
                                     installation.
        --build-bottle               Prepare the formula for eventual bottling
                                     during installation, skipping any
                                     post-install steps.
        --bottle-arch                Optimise bottles for the specified
                                     architecture rather than the oldest
                                     architecture supported by the version of
                                     macOS the bottles are built on.
    -f, --force                      Install without checking for previously
                                     installed keg-only or non-migrated
                                     versions.
    -v, --verbose                    Print the verification and postinstall
                                     steps.
        --display-times              Print install times for each formula at the
                                     end of the run.
    -i, --interactive                Download and patch formula, then open a
                                     shell. This allows the user to run
                                     ./configure --help and otherwise
                                     determine how to turn the software package
                                     into a Homebrew package.
    -g, --git                        Create a Git repository, useful for
                                     creating patches to the software.
    -h, --help                       Show this message.
Error: invalid option: --with-xml2-config
```

zerebubuth:
The installation of `libxml-ruby` was broken because it was picking up the system version of the libxml2 library. By setting a `bundler` config item, it's possible to point it to the right place.

Also updated the instructions for `phantomjs` -- do we even use this anymore? Anyway, Homebrew has moved to a cask and it is now not officially supported, but can be installed. Finally, added some of the extra command line tools that are used (`pngcrush`, `optipng`, etc...)

Hopefully this fixes #2371 - if anyone else has a Mac OS X machine, please test it out and let me know. For reference, I'm running 10.15 (Catalina), so it would be super helpful if anyone with other versions could test the instructions out.


tomhughes:
We don't normally bother telling people to install all the optional image compressors (most of which will never be used anyway) but I guess it doesn't really matter.

zerebubuth:
I got about half a screen's worth of warning messages about them, so figured they were worth installing. If there's some way to disable them in the example config (I assume they're not actually used in the tests), would that be better?

zerebubuth:
Sure. `rbenv` is actually how I did it... but figured it was an added level of complexity. If you prefer, I can switch over to `rbenv`. (Although I should note that the Ubuntu installation instructions also say `sudo gem2.5 install bundler`  - perhaps we want to change both?)

Firefishy:
Use `brew install rbenv`.

zerebubuth:
Thanks - I know that's how Homebrew works :stuck_out_tongue_winking_eye: 

My point was more that:

1. `brew install rbenv && rbenv install $VERSION && rbenv use $VERSION` takes more steps _and_ considerably longer time, since `rbenv install` is from source, not from Homebrew's pre-compiled bottles.
2. `sudo gem2.5 install bundler` is what the install docs currently recommend on Ubuntu.

I think `rbenv` is a better way to go - and it's what I use personally. I just wanted to check from everyone, including @gravitystorm, that everyone agrees that the usability trade-off is worth it. And that we should change it for both Mac OS X and Ubuntu.

gravitystorm:
>  I just wanted to check from everyone, including @gravitystorm, that everyone agrees that the usability trade-off is worth it.

No, I don't agree that the trade-off is worth it. Our codebase has a minimum version of ruby, but no particular needs for that specific version (i.e. 2.6 is fine, 2.7 will be fine, etc). Modern bundler sorts out making sure all the gems are in order (i.e. not running a copy of a gem compiled against a different ruby version), so that major feature of rbenv is unnecessary too.

I haven't run any kind of ruby manager since the whole ruby1.8 / 1.9 / 2.0 mess, and I work on plenty of different projects with different minimum ruby versions!

Of course, if someone prefers rbenv, be my guest. But like non-standard postgresql configs, I don't think we should confuse our installation notes with non-standard ruby setups or setups that require multiple optional steps.

>     2\. `sudo gem2.5 install bundler` is what the install docs currently recommend on Ubuntu.

This needs a review. I'm not sure why we're installing the bundler package and then installing bundler again via rubygems - maybe for hysterical raisins? If the Xenial version of bundler is too old (not sure that it is) then we shouldn't be installing it anyway - but if it's fine, there's no need for the gem install version.

> Also updated the instructions for `phantomjs` -- do we even use this anymore?

Yes, in the [system tests](https://github.com/openstreetmap/openstreetmap-website/blob/master/test/application_system_test_case.rb). But migrating to one of the headless webdriver things is another topic worth working on.

zerebubuth:
Added a section at the end about `rbenv` in commit 5895ae62a. Hopefully this is a way of keeping the instructions as simple as possible while also pointing out to anyone interested in the digression that many of us would recommend `rbenv` to manage Ruby installs and gems in non-admin user accounts.

gravitystorm:
> If there's some way to disable them in the example config (I assume they're not actually used in the tests), would that be better?

[The warnings should already be excluded when running the tests](https://github.com/openstreetmap/openstreetmap-website/blob/master/config/image_optim/test.yml). Did you see the warnings during development?

gravitystorm:
Other than that, it looks fine to me. I think the rbenv stuff is a big digression in the install notes, but I'm willing to merge it.

zerebubuth:
> > If there's some way to disable them in the example config (I assume they're not actually used in the tests), would that be better?
> 
> [The warnings should already be excluded when running the tests](https://github.com/openstreetmap/openstreetmap-website/blob/master/config/image_optim/test.yml). Did you see the warnings during development?

Yes, although during `bundle exec rake db:migrate` (and perhaps `db:create` too, I can't remember), not during the tests.

Issue #2401, opened 2019-10-17
Can't proceed after bundler install


icemagno:
I can't go further after 
`cd openstreetmap-website`
`bundle install`

```
root@36022329ca97:/home/openstreetmap-website# bundle install
Traceback (most recent call last):
        1: from /usr/local/bin/bundle:23:in `<main>'
/usr/local/bin/bundle:23:in `load': cannot load such file -- /usr/share/rubygems-integration/all/gems/bundler-1.16.1/exe/bundle (LoadError)
```
I'm trying to create a Docker image from the install procedure.

mmd-osm:
You may try #2272 instead of creating your own Docker image.

tomhughes:
Other than that it appears your bundler install is broken but you've provided no details of how you installed that or even what instructions you are following, what your platform us, or anything else.

I'm not even sure how @mmd-osm has concluded that you're using docker...

mmd-osm:
It’s that root@36022329ca97 in the prompt, and the op wrote he’s working on a Docker image.

icemagno:
@tomhughes ... good Lord. I WROTE it. Can you READ the entire text?

By the way... Can't find the Dockerfile in the code.

tomhughes:
Well if you don't want my help then fine. I'll sign off from this ticket. Good luck.

mmd-osm:
It's a Pull Request, which means the Dockerfile is not yet in the master branch. It still needs to be merged.

See https://github.com/fazlerabbi37/openstreetmap-website/tree/docker

icemagno:
Many thanks. Consider this as solved. You can close.

icemagno:
@mmd-osm after follow the instructions from https://github.com/fazlerabbi37/openstreetmap-website/tree/docker I have now a database container running and an image from the website but I cant put the website container to run.

Actualy I do not ran the docker compose file. Just build the two steps (database and website) one after other. I can put the database to run and can access it from outside Docker.

I can build the website image but I can't run it because when I try to run `bundle exec rails s -p 3000 -b '0.0.0.0'` I got the help for `bundle exec new`:

```
Usage:
  rails new APP_PATH [options]

Options:
      [--skip-namespace], [--no-skip-namespace]            # Skip namespace (affects only isolated applications)
  -r, [--ruby=PATH]                                        # Path to the Ruby binary of your choice
                                                           # Default: /usr/local/bin/ruby
[...]

PR #2409, opened 2019-10-22
Add Docker Compose Support for Development Environment

jalessio:
This is a cleaned up version of PR #2406 which was built upon the work of @fazlerabbi37 in PR #2272. 

- Introduces the usage of `docker-compose` for creating a development environment for this project
- Adds `DOCKER.md` for Docker-specific installation instructions
- ~Adds a `Makefile` for providing shortcuts for running common `docker-compose` commands~

There is certainly still room for improvement here, including pruning unneeded dependencies in the top-level `Dockerfile`, but I propose merging this working version and iterating from that point _(particularly since many of the changes will apply regardless of the installation method)_. I think we will benefit from feedback of a few people actually using this before making too many more changes.


DOCKER.md
```diff
@@ -0,0 +1,97 @@
...
+
+    cp config/example.storage.yml config/storage.yml
+
+### Database
+
+    cp config/docker.database.yml config/database.yml
+
+### App configuration
+
+    cp config/settings.yml config/settings.local.yml
```
gravitystorm: No, please don't do this. The settings.local.yml file should only contain values that you want to override.

See https://github.com/openstreetmap/openstreetmap-website/blob/edd49e85823079d6de19cbc9f89c6abc0568382b/INSTALL.md#prepare-local-settings-file

DOCKER.md
```diff
@@ -0,0 +1,97 @@
...
+    docker-compose up -d
+
+This will launch one Docker container for each 'service' specified in `docker-compose.yml` and run them in the background. There are two options for inspecting the logs of these running containers:
+
+- You can tail logs of a running container with a command like this: `docker-compose logs -f web` or `docker-compose logs -f db`.
+- Instead of running the containers in the background with the `-d` flag, you can launch the containers in the foreground with `docker-compose up`. The downside of this is that the logs of all the 'services' defined in `docker-compose.yml` will be intermingled. If you don't want this you can mix and match - for example, you can run the database in background with `docker-compose up -d db` and then run the Rails app in the foreground via `docker-compose up web`.
+
+### Migrations
+
+While the `db' service is running, open another terminal windows and run:
```
gravitystorm: * mismatched quotes, should be `db`.
* terminal window should be singular

DOCKER.md
jalessio: 👍 Fixed by copying content from `INSTALL.md`. 

For the record, I don't really like all the duplication of content from `INSTALL.md` but having a complete set of Docker-specific instructions seems valuable instead of making people piece together instructions from multiple files. If someone has a better idea on how to reduce the duplication please let me know.

DOCKER.md
jalessio: 👍 Fixed

jalessio:
@gravitystorm any thoughts on how to move this forward? I believe it's in a usable state and offers a quick way to get up and running with this project locally. I've been using it locally for over a week now without any issues.

I could definitely benefit from a few people trying it out from scratch and providing feedback, but perhaps that's more likely to occur once it's on `master`? 

@zerebubuth / @Firefishy / @mmd-osm - would any of you be willing to give this branch a quick run-through locally since you were all active in the previous PR (#2406)? Thanks!

SomeoneElseOSM:
@jalessio Maybe write an OSM diary entry (or similar) explaining what this is and how it would be useful to people - explain why would they benefit from having a local copy of this code, and what Docker is and how that helps them to do that?  I appreciate that there's some of that at the top of https://github.com/openstreetmap/openstreetmap-website/pull/2272 , but a bit more info would also be useful.

mmd-osm:
Test instructions are here: https://github.com/openstreetmap/openstreetmap-website/issues/2401#issuecomment-546763439

mmd-osm:
I tested the set up yesterday, and it looks good. It took forever to download phantomjs and I ended up removing it from the Dockerfile (it's deprecated / archived anyway and should be replaced in the master branch as wel).

@jalessio : as most of the editing API calls run via CGImap these days in production, we might consider (optionally) including CGImap https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/Dockerfile as well,. Dispatching to Rails port / cgimap could be done using lighttpd (https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/lighttpd.conf) as a reverse proxy.

gravitystorm:
> @gravitystorm any thoughts on how to move this forward?

I'll have another look later in the week - it looks like you've already taken care of a few other things that I was going to query.

> as most of the editing API calls run via CGImap these days in production, we might consider (optionally) including CGImap

No, please don't. I'm happy to see stuff in this repository to help with setting up development environments, but we deliberately leave production choices out of consideration.

mmd-osm:
> we deliberately leave production choices out of consideration

We'll add it to the CGImap repo then, no worries. I don't really care which repo hosts this config, as long as we offer this option at all.

icemagno:
As a suggestion i would like to say that you could build your own image and publish at Docker Hub. So we would just run it.  Maybe the database image and the website image.  The only flaw I see is regarding the configuration file. The OSM guys could accept database  configuration via environment variables and not only by YAML. I think the OSM WEBSITE is not very flexible and was build in a very 
monilithic old fashion.

DOCKER.md
```diff
@@ -0,0 +1,100 @@
...
+This will launch one Docker container for each 'service' specified in `docker-compose.yml` and run them in the background. There are two options for inspecting the logs of these running containers:
+
+- You can tail logs of a running container with a command like this: `docker-compose logs -f web` or `docker-compose logs -f db`.
+- Instead of running the containers in the background with the `-d` flag, you can launch the containers in the foreground with `docker-compose up`. The downside of this is that the logs of all the 'services' defined in `docker-compose.yml` will be intermingled. If you don't want this you can mix and match - for example, you can run the database in background with `docker-compose up -d db` and then run the Rails app in the foreground via `docker-compose up web`.
+
+### Migrations
+
+Run the Rails database migrations:
+
+    docker-compose run --rm web rake db:migrate
```
gravitystorm: For consistency and reliability, we should have `... bundle exec rake db:migrate` here

DOCKER.md
```diff
@@ -0,0 +1,100 @@
...
+
+Once you have data loaded for Washington, DC you should be able to navigate to `http://localhost:3000/#map=12/38.8938/-77.0146` to begin working with your local instance.
+
+### Additional Configuration
+
+See `CONFIGURE.md` for information on how to manage users and enable OAuth for iD, JOSM etc.
+
+### Tests
+
+    docker-compose run --rm web rake test:db
```
gravitystorm: bundle exec

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ADD Gemfile* /app/
+RUN bundle install
+
+# Setup local
+RUN sed -i -e 's/# en_GB.UTF-8 UTF-8/en_GB.UTF-8 UTF-8/' /etc/locale.gen && \
+    echo 'LANG="en_GB.UTF-8"'>/etc/default/locale && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=en_GB.UTF-8
+
+ENV LANG en_GB.UTF-8
```
gravitystorm: There needs to be a bit about `bundle exec rake yarn:install` somewhere, either here in the docker config, or elsewhere in the manual installation notes.

icemagno:
What about config the database connection using environment vars?

gravitystorm:
> What about config the database connection using environment vars?

@icemagno What's your usecase? Would this be for your development environment, or for production deployments?

icemagno:
@gravitystorm I'm developing a distributed software with about 50 microservices in Spring boot and must follow some Standards in the continuous delivery and infrastructure deployment.  I already have it running and using a old OSM server I've created. Now I have a Docker conteiner running a very nice and rich styled OSM and want to create a software to create and edit OSM and custom features ( from the Brazilian Gov. Metadata Standard ). So I'm trying some options mostly (actualy a must) in Docker. The installation way I seen here is good and working but I need some more flexible because I need to document the installation process to the future and it must follow the current standard wich is one script for each container and must be possible to replace any container without affect others. So I need one script to build and run the database and another to do the same with the website. No docker compose is allowed even if it is the easy way to do it. Each build and run process must be documented alone.

mmd-osm:
This blog post https://blog.codeship.com/running-rails-development-environment-docker/ should give you some idea how to set up database.yml to fetch database user and password from environment variables. Besides the requirements you describe seem rather out of scope for this pull request.

icemagno:
@mmd-osm thanks. I'll give a look but you may consider use it anyway because this is the best way to offer a Docker image. Under my viewpoint it is not a good idea to create hard coded configurations in a Docker deploy. It is extremely boring editing files before building and sometimes I believe the Docker premisse of "fire-and-go" is broken. And the user must be free to choose database names, hosts and etc.

DOCKER.md
mmd-osm: `docker-compose run --rm web bundle exec rake db:migrate` didn't work here, while `docker-compose run --rm web rake db:migrate` did.

Error message:
```
docker-compose run --rm web bundle exec rake db:migrate
Creating network "osm-microcosms_default" with the default driver
Creating osm-microcosms_db_1 ... done
rake aborted!
PG::ConnectionBad: could not connect to server: Connection refused
	Is the server running on host "db" (172.19.0.2) and accepting
	TCP/IP connections on port 5432?
/usr/local/bundle/gems/pg-1.1.4/lib/pg.rb:56:in `initialize'
/usr/local/bundle/gems/pg-1.1.4/lib/pg.rb:56:in `new'
/usr/local/bundle/gems/pg-1.1.4/lib/pg.rb:56:in `connect'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/postgresql_adapter.rb:692:in `connect'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/postgresql_adapter.rb:223:in `initialize'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/postgresql_adapter.rb:48:in `new'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/postgresql_adapter.rb:48:in `postgresql_connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:811:in `new_connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:855:in `checkout_new_connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:834:in `try_to_checkout_new_connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:795:in `acquire_connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:523:in `checkout'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:382:in `connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:1014:in `retrieve_connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_handling.rb:118:in `retrieve_connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/connection_handling.rb:90:in `connection'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/tasks/database_tasks.rb:172:in `migrate'
/usr/local/bundle/gems/strong_migrations-0.4.1/lib/strong_migrations/database_tasks.rb:4:in `migrate'
/usr/local/bundle/gems/activerecord-5.2.3/lib/active_record/railties/databases.rake:60:in `block (2 levels) in <top (required)>'
/usr/local/bundle/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
/usr/local/bin/bundle:30:in `block in <main>'
/usr/local/bin/bundle:22:in `<main>'
Tasks: TOP => db:migrate
```

mmd-osm:
>>    as most of the editing API calls run via CGImap these days in production, we might consider (optionally) including CGImap
> No, please don't. I'm happy to see stuff in this repository to help with setting up development environments, but we deliberately leave production choices out of consideration.

FYI. A complete stack consisting of Rails port, Postgres db, cgimap and lighttpd is now described in more detail in https://github.com/zerebubuth/openstreetmap-cgimap/pull/213.

It leverages the Postgres and Rails port set up outlined in this issue, and extends docker-compose-yml by two additional services for cgimap and lighttpd.

jalessio:
TODO - possibly switch to Debian Bionic base image to match other parts of this repo. See https://github.com/openstreetmap/openstreetmap-website/pull/2473

migurski:
I just ran through the complete process of running this Docker environment, to test the user experience of a random passerby who’d like to contribute an edit to the site. @jalessio walked me through setting up Docker-Compose, initializing a clean new database, importing data using Osmosis (from a small PBF extract), and verifying that everything worked at `localhost:3000`.

I made the smallest change I could think of: updating the header text in a way that was obviously visible (history/schmistory, etc.). Rails hot-reloaded everything, and I was able to see my change immediately:

Screen Shot 2020-02-15 at 5 44 50 PM

As a Rails novice who doesn't expect to become an expert, this has been a super useful way to understand the moving parts of the website and test potential changes. I can imagine helping a designer implement an idea here and test it out live, useful to getting new ideas into OSM.

I know that Docker configs can fall out of date over time, but as we’ve seen with https://github.com/openstreetmap/chef/pull/266 tests for this can be added to our Travis configuration and images can be published to Docker Hub where a casual developer can try them out.

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    build:
+      context: .
+      dockerfile: docker/postgres/Dockerfile
+    ports:
+      - "54321:5432"
+    environment:
+      POSTGRES_HOST_AUTH_METHOD: trust
+      POSTGRES_DB: openstreetmap
+    volumes:
+      - ./docker-db-data:/var/lib/postgresql/data
```
Firefishy: This should really be a data volume, rather than a file mapping.
See: https://docs.docker.com/compose/compose-file/compose-file-v2/#volume-configuration-reference

Dockerfile
```diff
@@ -0,0 +1,52 @@
+FROM ruby:2.5
+
+# fixes dpkg man page softlink error while installing postgresql-client [source: https://stackoverflow.com/a/52655008/5350059]
+RUN mkdir -p /usr/share/man/man1 && \
+    mkdir -p /usr/share/man/man7
+
+# npm is not available in Debian repo so following official instruction [source: https://github.com/nodesource/distributions/blob/master/README.md#debinstall]
```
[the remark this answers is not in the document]
Firefishy: Best to use `FROM: ruby:2.5` rather than `FROM: ruby:2.5-buster`

DOCKER.md
```diff
@@ -0,0 +1,98 @@
...
+
+    docker-compose run --rm web bundle exec rake db:migrate
+
+### Tests
+
+Run the test suite by running:
+
+    docker-compose run --rm web bundle exec rake test:db
+
+### Loading an OSM extract
```
jalessio: Since loading an extract via osmosis stopped working after PR https://github.com/openstreetmap/openstreetmap-website/pull/2432 I'm inclined to remove this section until I can offer a working solution.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN mkdir -p /app
+WORKDIR /app
+
+# Install Ruby packages
+ADD Gemfile Gemfile.lock /app/
+RUN bundle install
+
+# Install NodeJS packages
+ADD package.json yarn.lock /app/
+RUN yarn
```
jalessio: @gravitystorm I realize you've been consistently asking for using the `bundle exec` form of commands for consistency _(which I appreciate)_, but I'll try to convince you of this exception... 

I toiled with getting this to work via `bundle exec rake db:migrate`, but I couldn't get it to run without adding a big chunk of the repository into the image during build. It ended up adding a bunch of extra files to the image *just* so that I could run `bundle exec rake db:migrate` instead `yarn` directly.  I believe the result is the same in the end so I went with the simpler option.

That said, perhaps someone else could compile a small list of files necessary to successfully run `rake` commands during the image build?

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+      nodejs \
+      osmosis \
+      postgresql-client \
+      ruby-dev && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# install npm packages
+RUN npm install -g --unsafe-perm \
+      phantomjs-prebuilt \
```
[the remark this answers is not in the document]
jalessio: @gravitystorm I moved to installing `phantomjs` via apt.

DOCKER.md
```diff
@@ -0,0 +1,100 @@
...
+            user="openstreetmap" \
+            validateSchemaVersion="no"
+
+Once you have data loaded for Washington, DC you should be able to navigate to `http://localhost:3000/#map=12/38.8938/-77.0146` to begin working with your local instance.
+
+### Additional Configuration
+
+See `CONFIGURE.md` for information on how to manage users and enable OAuth for iD, JOSM etc.
+
+### Tests
```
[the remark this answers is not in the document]
jalessio: 👍 Done

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+      yarn
```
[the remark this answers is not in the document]
jalessio: @gravitystorm Agreed! I'm now installing `yarn` via apt.

Dockerfile
jalessio: I'm not calling the `yarn` installer during Docker image build.

jalessio:
I made a number of changes to this PR and I believe I have addressed all the comments / requests for changes _(except one! see below)_. I have also rebased it against master.

## TODO

- [ ] Update docker-compose config to use a data volume per @Firefishy (https://github.com/openstreetmap/openstreetmap-website/pull/2409#pullrequestreview-362092729)
- [ ] Remove instructions for using osmosis? _(since they don't work any more)_

DOCKER.md
mmd-osm: There’s a work songs available: https://github.com/openstreetmap/openstreetmap-website/issues/2543#issuecomment-587889501

Just add it to the description

migurski: It’s a super-ugly workaround, almost to the point that I think it’d be harmful just to recommend that people monkey with the schema pre- and post-import. Let’s just remove the section for now?

mmd-osm: It’s as ugly as having to tweak sequences. My point was, it was already ugly before, we’re not adding much to that by this workaround.

Dockerfile
tomhughes: You're already done `bundle install` which should install everything `bundle exec` might need - if you don't use `bundle exec` then you have no idea if things are using the expected modules, or even if they will work at all.

migurski:
I added tests for the docker-compose configuration in a separate Travis job.

DOCKER.md
migurski: Osmosis 0.47.1 now correctly writes to the DB so this section should work fine.

francois2metz:
Since a few weeks, the postgres image on the docker hub has changed and if no password is set, you have to set the `POSTGRES_HOST_AUTH_METHOD` environment variable to `trust`.
Also instead of mounting the local directory for the postgres data, I just used the default volume of the postgres image.

Here is my patch
```
diff --git a/docker-compose.yml b/docker-compose.yml
index 70730febf..5d7943e1c 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,5 @@ services:
     ports:
       - "54321:5432"
     environment:
+      POSTGRES_HOST_AUTH_METHOD: trust
       POSTGRES_DB: openstreetmap
-    volumes:
-      - ./docker-db-data:/var/lib/postgresql/data
```

Except that it worked flawless

migurski:
Update: I rebased everything to `master` to support work on #2735.

hiddewie:
I had to bump the `ruby` Docker base image from 2.5 to version 2.7 to get the `bundle install` to work.

migurski:
Update: I rebased everything to `master`.

frodrigo:
I test this PR and there is isse with apt install package signature.
Then I need to run `docker-compose run --rm web bundle exec rake yarn:install` to have the js working.

migurski:
Thanks @frodrigo! I rebased to try to reproduce the apt error, but passed cleanly: https://travis-ci.org/github/openstreetmap/openstreetmap-website/jobs/748014637

Can you describe your error in more detail here?

frodrigo:
> Thanks @frodrigo! I rebased to try to reproduce the apt error, but passed cleanly: https://travis-ci.org/github/openstreetmap/openstreetmap-website/jobs/748014637
> 
> Can you describe your error in more detail here?

Tested today on a other setup. And think good with APT. Both setups with fresh docker image download. Nevertheless the `rake yarn:install` step still required.

migurski:
Glad to hear the Apt parts are working. I’ll ensure that the Yarn step is well-reflected in the docs.

migurski:
Update: I rebased this PR on master branch added a note about running `yarn`.

I separately tried a few options for putting the Node module installation steps right into the `Dockerfile` ([here](https://travis-ci.org/github/migurski/openstreetmap-website/builds/748427428), [here](https://travis-ci.org/github/migurski/openstreetmap-website/builds/749307618), and [here](https://travis-ci.org/github/migurski/openstreetmap-website/builds/749313792)) but could not figure out a way to install globally and recognize the results. If any has advice on this, I’d love to hear it.

migurski:
Thanks for #3005, @MegaMind77-coder. This PR is up-to-date with the default branch and should work for you. Hope we can get it merged and supported to smooth the experience of new contributors.

MegaMind77-coder:
> Thanks for #3005, @MegaMind77-coder. This PR is up-to-date with the default branch and should work for you. Hope we can get it merged and supported to smooth the experience of new contributors.

Thanks, @migurski  will try to get it up and running 
Very excited to contribute to this amazing project.

mmd-osm:
It would be good to use a dedicated user inside the Dockerfile. Running stuff as root inside a container isn't recommended.

migurski:
Thanks for the suggestion @mmd-osm. Could you point to any examples that you’d recommend we borrow from?

hiddewie:
There is a nice article on https://thelurkingvariable.com/2019/12/17/ruby-with-docker-compose-as-a-non-root-user/ which explains the process step by step.

fazlerabbi37:
> Could you point to any examples that you’d recommend we borrow from?

I think @mmd-osm meant this one: [Run the Docker daemon as a non-root user](ttps://docs.docker.com/engine/security/rootless)

mmd-osm:
No, this is about explicitly defining a USER in the Dockerfile, according to https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-2-set-a-user

migurski:
The `USER` definition in the Dockerfile seems like overkill to me: bundle and yarn installs are the only things that might come after it. Running the site as a non-root user (@hiddewie’s link) does make sense if it can be made to match OSM’s production installation. Any advice on that, @mmd-osm?

mmd-osm:
According to https://github.com/openstreetmap/chef/blob/master/cookbooks/web/recipes/rails.rb#L75 I would assume that rails port is deployed and later run by user&group “rails”. So the idea is that the Docker setup should also reflect this.

gravitystorm:
> So the idea is that the Docker setup should also reflect this.

No, a development setup for this repo should have nothing to do with whatever the production deployment is for the OSMF. No need to share usernames, or groups, and of course, different software is used too (e.g. shouldn't use apache/passenger for development, for example). 

So please don't do anything to artificially make it similar to OSMF production, instead just focus on whatever makes the best development environment.

mmd-osm:
@gravitystorm : I guess you’re not running your rails port dev environment as a root user either. This makes me wonder why a container should be set up differently. Maybe I’m missing something here, I don’t see the overhead.

gravitystorm:
>  I guess you’re not running your rails port dev environment as a root user either.

Of course - but I'm not running it as rails:rails either.

migurski:
Running as root has the advantage of making the Docker config portions shorter, which might positively influence @tomhughes or @gravitystorm’s decision to commit this PR, supporting an easy dev environment in the default branch, and trusting that changes made by participants using Docker are trustworthy.

mmd-osm:
> I'm not running it as rails:rails either.

That’s perfectly fine. What I was trying to get at is that some non-root user would be typically used, and that should also apply in case of a dev environment.

migurski:
I’m expecting to leave this PR as-is for the moment, not seeing a compelling recommendation on how to change it just yet.

Andy, I think the suggestions from last year leading to the `changes requested` label have been mostly addressed at various points. I still see an open request from @Firefishy to use data volumes. What else is keeping that label active?

mmd-osm:
Another thing to keep in mind is that whatever is written to the /app directory inside the container will end up in your own filesystem with "root" ownership. This may be somewhat inconvenient and could cause some other issues down the road. Ideally, the container would use your own uid and gid. We'll leave that to the next iteration, until someone else complains about it.

migurski:
So far, that’s the `node_modules` directory. I switched this to global installation [in a few attempts above](https://github.com/openstreetmap/openstreetmap-website/pull/2409#issuecomment-743807851), but I wasn’t successful in getting the running Rails app to read from those.

mmd-osm:
`bundle install` also needs to deploy a few gems in a fresh install, so I would assume this is relevant here as well. 

In a quick test drive, I created a non-root user with uid 1000 right before:

```
# Setup app location
RUN mkdir -p /app
```

(uid 1000 matches my Ubuntu user id)

Then I  changed the ownership of that /app directory to the new user using chown, followed by `USER xyz`.
`bundle install` and `yarn install` would then run as non-root, and that seemed to work ok.

migurski:
The UID matching seems a bit brittle, since it’ll be different for different users. My inclination here is to trust that someone running Docker to test changes to the OSM website will be able to handle a local directory permission change. I think it’s a lower bar to clear than running this on the underlying system.

docker/postgres/openstreetmap-postgres-init.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -ex
+
+# Create 'openstreetmap' user
+psql -v ON_ERROR_STOP=1 -U "$POSTGRES_USER" <<-EOSQL
+    CREATE USER openstreetmap PASSWORD 'openstreetmap';
+    GRANT ALL PRIVILEGES ON DATABASE openstreetmap TO openstreetmap;
+EOSQL
```
gravitystorm, line 8: This is where the `openstreetmap` role is created - see my other review note.

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+      libffi-dev \
+      libmagickwand-dev \
+      libpq-dev \
+      libsasl2-dev \
+      libxml2-dev \
+      libxslt1-dev \
+      locales \
+      nodejs \
+      default-jre-headless \
+      phantomjs \
```
gravitystorm, line 23: We don't use phantomjs any more, see #2544 and #2598 

Without geckodriver, the test suite doesn't work. I've had a look at packages, but I could only find [a request on debian to include geckodriver](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=8742070 and it hasn't had much activity. Perhaps someone with better search-fu can find a debian package for geckodriver?

gravitystorm (changes requested):
I've added the specific review points from my testing today. I also note that Firefishy raised an issue about data volumes, but I'm not enough of a Docker expert to know if it's a nice-to-have or a must-have.

The geckodriver issue raises the biggest point for me. The ruby images this Dockerfile are based on are themselves based on Debian, rather than Ubuntu. Our installation notes and Vagrantfile are both based on Ubuntu. I don't know if the geckodriver issue is easy to fix or not, but in the longer term, I'm concerned that there may be other packages that we use from Ubuntu that aren't available in Debian, or whichever debian release the ruby image of the day is based on. So I wonder if it's better to align these, and base this Dockerfile on an Ubuntu-based image instead?

.travis.yml
```diff
@@ -1,38 +1,58 @@
...
+    script:
+      - bundle exec rubocop -f fuubar
+      - bundle exec rake eslint
+      - bundle exec erblint .
+      - bundle exec brakeman -q
+      - bundle exec rake db:structure:dump
+      - sed -e "/idle_in_transaction_session_timeout/d" -e 's/ IMMUTABLE / /' -e "/^--/d" db/structure.sql > db/structure.actual
+      - diff -uw db/structure.expected db/structure.actual
+      - bundle exec rake test:db
+  - name: "Docker-Compose Configuration"
```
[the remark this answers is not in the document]
migurski: Would it make sense to instead rebase this PR on top of #3002 entirely?

Dockerfile
migurski: This is pretty far outside my experience, so hopefully someone can chime in.

migurski:
Thanks for your detailed notes, Andy. I’ll make updates where you suggested and wait for additional input on some of the places where we have questions. Do you think the move to GH Actions is a foregone conclusion now?

gravitystorm:
> Do you think the move to GH Actions is a foregone conclusion now?

It's not a done deal until it gets merged. But I suspect it'll be clear what's happening with Github Actions before the geckodriver stuff here is resolved, so I thought I should flag it now anyway.

migurski:
I restored the paragraph and switched DB users. The superuser setup seemed unnecessary so I skipped it. I’d appreciate some advice on the phantomjs thing, not really sure how it’s used here.

It also seems that #3002 is now merged, so I assume this will all have to move to GH actions?

gravitystorm:
> The superuser setup seemed unnecessary so I skipped it.

So how do the test database(s) get created? The development database is created as part of the environment variables passed in docker-compose.yml (I assume that the container is set up to read the environment and create that database) but I can't see where either the test databases are created, nor enough permissions given to the `openstreetmap` role to do so.

> I’d appreciate some advice on the phantomjs thing, not really sure how it’s used here.

phantomjs was used by the tests, but now it's firefox-geckodriver. Both are ways of mimicking a web browser in the test suite. Without geckodriver installed, the tests won't pass. And having the tests pass is a key part of the development environment. :smile:

We either need someone to suggest how to get firefox-geckodriver installed on debian as part of the image build process, or to advise on whether switching to an Ubuntu-based image would be better.

migurski:
> So how do the test database(s) get created? 

The DB is created by specifying a config ENV var for the [postgres Docker image](https://hub.docker.com/_/postgres); that happens at this line of [docker-compose.yml](https://github.com/jalessio/openstreetmap-website/blob/801d5b6261be6da4c1d47fc9679a8c74ec699370/docker-compose.yml#L23).

> And having the tests pass is a key part of the development environment.

lol 👍 

This might be putting too much pressure on the Docker setup. The [test config we’ve defined here](https://github.com/jalessio/openstreetmap-website/blob/801d5b6261be6da4c1d47fc9679a8c74ec699370/.travis.yml#L3-L41) continues to run the application tests outside of Docker [just as in the default branch, pre-3002](https://github.com/openstreetmap/openstreetmap-website/blob/1b4a06adef58262a3e9eba47049a6ca212d88be4/.travis.yml). The [Docker-specific test](https://github.com/jalessio/openstreetmap-website/blob/801d5b6261be6da4c1d47fc9679a8c74ec699370/.travis.yml#L42-L59) just checks that the dev environment results in a functional Rails site so that casual contributors can test their changes. As an occasional contributor, I never run the test suite locally and rely on the CI to do it for me. 

I’ll read up on Geckodriver and PhantomJS.

gravitystorm:
Ah, sorry, perhaps I wasn't clear. I wasn't referring to any CI-related matters. I mean that when you use this Docker configuration to set up a local development environment, you need to have multiple databases:

* A development database, called `openstreetmap`, that I know is created by the env vars
* One or more (probably more) test databases, so that you can run the test suite (or parts of it) locally before committing changes.

So my concern is that they way this has been written, where the env vars set up only one database, and they `openstreetmap` postgres role is not a super user, means that the test suite will not run on your laptop.

For our INSTALL.md and Vagrant approaches, the postgres role used in the test configuration has enough permissions to create whichever databases are required when the test suite runs

> As an occasional contributor, I never run the test suite locally and rely on the CI to do it for me.

Well I hope that this will be made easier by having a Docker-based development environment for you :smile: As per the notes, running `docker-compose run --rm web bundle exec rake test:db` should be enough.

To be double-clear, I'm not asking that the full test suite runs in Docker as part of the CI system. I'm not interested in that. But I'm asking that the test suite works when I run the above command on my laptop.

migurski:
I’m working on getting the tests running. Quick question about the proper Ubuntu image to use: the Chef repo [seems to specify 20.04](https://github.com/openstreetmap/chef/blob/a24470a145847920ffe3cb6764c93d9192ed8ca0/.github/workflows/test-kitchen.yml#L105), but the [brightbox/ruby-ng PPA](https://github.com/openstreetmap/chef/blob/65902c2f5a35a64a2159a4a9bb19c39a5f5bd097/cookbooks/apt/recipes/default.rb#L61-L64) where Ruby 2.7 comes from [does not offer 20.04 Focal packages](https://launchpad.net/~brightbox/+archive/ubuntu/ruby-ng). I can make 2.7 work on Ubuntu 18.04, but that leads to a number of other incompatibilities and workarounds such as NodeJS and PG version mismatches, installing tzdata manually, and [something to do with bootsnap](https://github.com/Shopify/bootsnap/issues/262). 

Is there another way to use 20.04 and Ruby 2.7 compatible with how we do things in Chef?

migurski:
Meanwhile, I’ve just rebased this PR to cover the latest Github Actions configuration on the default branch.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Build Docker Image
+      run: |
+        docker-compose build
+    - name: Start Docker-Compose
+      run: |
+        docker-compose up -d
+        sleep 15 # let the DB warm up a little
+    - name: Prepare Database
+      run: |
+        docker-compose run --rm web rake db:migrate
```
tordans, line 26:
```suggestion
        docker-compose run --rm web rails db:migrate
```

Since we are on a fairly new version of Rails, this is the same and looks a bit nicer and is more up to date. 

(I will comment only this one random instance, but there are a few of them)

DOCKER.md
```diff
@@ -0,0 +1,104 @@
...
+    docker-compose run --rm web osmosis \
+        -verbose    \
+        --read-pbf district-of-columbia-latest.osm.pbf \
+        --write-apidb \
+            host="db" \
+            database="openstreetmap" \
+            user="openstreetmap" \
+            validateSchemaVersion="no"
+
+Once you have data loaded for Washington, DC you should be able to navigate to `http://localhost:3000/#map=12/38.8938/-77.0146` to begin working with your local instance.
```
tordans, line 86:
```suggestion
Once you have data loaded for Washington, DC you should be able to navigate to [`http://localhost:3000/#map=12/38.8938/-77.0146`](http://localhost:3000/#map=12/38.8938/-77.0146) to begin working with your local instance.
```

DOCKER.md
```diff
@@ -0,0 +1,104 @@
...
+
+Install required front-end Node modules:
+
+    docker-compose run --rm web bundle exec rake yarn:install
+
+### Loading an OSM extract
+
+This installation comes with no geographic data loaded. You can either create new data using one of the editors (Potlatch 2, iD, JOSM etc) or by loading an OSM extract. Here an example for loading an OSM extract into your Docker-based OSM instance.
+
+For example, let's download the District of Columbia from Geofabrik:
```
tordans, line 71:
```suggestion
For example, let's download the District of Columbia from Geofabrik or [any other region](https://download.geofabrik.de):
```

DOCKER.md
```diff
@@ -0,0 +1,104 @@
...
+            host="db" \
+            database="openstreetmap" \
+            user="openstreetmap" \
+            validateSchemaVersion="no"
+
+Once you have data loaded for Washington, DC you should be able to navigate to `http://localhost:3000/#map=12/38.8938/-77.0146` to begin working with your local instance.
+
+### Additional Configuration
+
+See `CONFIGURE.md` for information on how to manage users and enable OAuth for iD, JOSM etc.
```
tordans, line 90:
```suggestion
See [`CONFIGURE.md`](CONFIGURE.md) for information on how to manage users and enable OAuth for iD, JOSM etc.
```

migurski:
Update: @Firefishy helped me understand that the Brightbox PPA in our Chef config is a red herring; I’ve pushed an update running on Ubuntu 20.04 with Ruby 2.7 and Node 10.

docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+      context: .
+    volumes:
+      - .:/app
+      # don't mount tmp directory (https://github.com/Shopify/bootsnap/issues/177)
+      - /app/tmp
+    ports:
+      - "3000:3000"
+    environment:
+      # https://github.com/Shopify/bootsnap/issues/262
+      ENABLE_BOOTSNAP: 'false'
```
mmd-osm, line 15: I don't see where ENABLE_BOOTSNAP would be used right now. Did you maybe forget to check in your changes to config/boot.rb here?

tordans:
First of all: Thanks a lot for this. This was the easiest setup I did, yet :-).

Here are a few things that happened:

### db:migrate

I get this response. Is this something important or can it be ignored?

```
$ docker-compose run --rm web bundle exec rails db:migrate
Creating openstreetmap-website_web_run ... done
pngcrush worker: `pngcrush` not found; please provide proper binary or disable this worker (--no-pngcrush argument or `:pngcrush => false` through options)
advpng worker: `advpng` not found; please provide proper binary or disable this worker (--no-advpng argument or `:advpng => false` through options)
optipng worker: `optipng` not found; please provide proper binary or disable this worker (--no-optipng argument or `:optipng => false` through options)
pngquant worker: `pngquant` not found; please provide proper binary or disable this worker (--no-pngquant argument or `:pngquant => false` through options)
jhead worker: `jhead` not found, `jpegtran` not found; please provide proper binary or disable this worker (--no-jhead argument or `:jhead => false` through options)
jpegoptim worker: `jpegoptim` not found; please provide proper binary or disable this worker (--no-jpegoptim argument or `:jpegoptim => false` through options)
jpegtran worker: `jpegtran` not found; please provide proper binary or disable this worker (--no-jpegtran argument or `:jpegtran => false` through options)
gifsicle worker: `gifsicle` not found; please provide proper binary or disable this worker (--no-gifsicle argument or `:gifsicle => false` through options)
svgo worker: `svgo` not found; please provide proper binary or disable this worker (--no-svgo argument or `:svgo => false` through options)
```

The migrations did run.

### rake test:db

Something is wrong:

```
$ docker-compose run --rm web bundle exec rake test:db

Creating openstreetmap-website_web_run ... done

rake aborted!
ActiveRecord::StatementInvalid: PG::InsufficientPrivilege: ERROR:  permission denied to create database
/usr/local/bundle/gems/activerecord-6.0.3.4/lib/active_record/connection_adapters/postgresql/database_statements.rb:92:in `exec'
…
Tasks: TOP => db:test:load => db:test:purge
```

### osmosis

This is still running – maybe. Or it got stuck. Will update with results later …

```
docker-compose run --rm web osmosis \
                                                                   -verbose    \
                                                                   --read-pbf berlin-latest.osm.pbf \
                                                                   --write-apidb \
                                                                       host="db" \
                                                                       database="openstreetmap" \
                                                                       user="openstreetmap" \
                                                                       validateSchemaVersion="no"
```

### Suggestions/Questions

- Should we document in more detail how to create a local user? Thinks like inspecting log to find the activation URL and cleaning this up to use it …
- Should we document in more detail how to "look into the docker DB with a local db client"? / How does this work?
- Should we suggest to setup a local domain? And use this for links in emails for example (the emails dont use the localhost that I use to open the page)?

### My current state

- I can load the website locally
- I can create a user
- I cannot edit via iD (alert: "iD was not configured"), but that is not a priority for me
- I cannot run tests yet (with test:db)

docker-compose.yml
migurski: This is me chasing down some Ruby 2.5/2.7 vs. Ubuntu 18/20 weirdness when migrating away from the `ruby:2.7` base image per @gravitystorm’s recommendation above. I might be able to back this out if the latest Ubuntu 20.04 + Ruby 2.7 commit builds cleanly.

migurski:
Thanks for your comments @tordans! I’m really happy this is useful for you.

I believe the `db:migrate` warnings can be ignored; I see them too but it’s not clear to me how to suppress them.

You’re right that `rake test:db` does not currently work. I’m most of the way through migrating to an Ubuntu base that will make this possible, then we’ll be able to fiddle the permissions to get the tests to run cleanly.

Berlin’s a big town, I would expect the Osmosis import to take a while.

I like all of your other suggestions, and I’d be happy to expand this PR to include them.

mmd-osm:
> I cannot edit via iD (alert: "iD was not configured"), but that is not a priority for me

A step by step description is included here: https://github.com/openstreetmap/openstreetmap-website/blob/master/CONFIGURE.md#oauth-consumer-keys

docker-compose.yml
migurski: Some details on why Bootsnap. Without these references in `docker-compose.yml`, we see this output:

```
+ docker-compose run --rm web bundle exec rake db:migrate
Starting openstreetmap-website_db_1 ... done
Errno::ENOENT: No such file or directory - bs_fetch:atomic_write_cache_file:chmod
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/compile_cache/iseq.rb:30:in `fetch'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/compile_cache/iseq.rb:30:in `fetch'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/compile_cache/iseq.rb:47:in `load_iseq'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `require'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `block in require_with_bootsnap_lfi'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/loaded_features_index.rb:92:in `register'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require_with_bootsnap_lfi'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:31:in `require'
/var/lib/gems/2.7.0/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:324:in `block in require'
/var/lib/gems/2.7.0/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:291:in `load_dependency'
/var/lib/gems/2.7.0/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:324:in `require'
/var/lib/gems/2.7.0/gems/bootstrap-4.5.3/lib/bootstrap.rb:61:in `register_rails_engine'
/var/lib/gems/2.7.0/gems/bootstrap-4.5.3/lib/bootstrap.rb:11:in `load!'
/var/lib/gems/2.7.0/gems/bootstrap-4.5.3/lib/bootstrap.rb:75:in `<main>'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `require'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `block in require_with_bootsnap_lfi'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/loaded_features_index.rb:92:in `register'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require_with_bootsnap_lfi'
/var/lib/gems/2.7.0/gems/bootsnap-1.5.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:31:in `require'
/var/lib/gems/2.7.0/gems/bundler-2.2.3/lib/bundler/runtime.rb:66:in `block (2 levels) in require'
/var/lib/gems/2.7.0/gems/bundler-2.2.3/lib/bundler/runtime.rb:61:in `each'
/var/lib/gems/2.7.0/gems/bundler-2.2.3/lib/bundler/runtime.rb:61:in `block in require'
/var/lib/gems/2.7.0/gems/bundler-2.2.3/lib/bundler/runtime.rb:50:in `each'
/var/lib/gems/2.7.0/gems/bundler-2.2.3/lib/bundler/runtime.rb:50:in `require'
/var/lib/gems/2.7.0/gems/bundler-2.2.3/lib/bundler.rb:174:in `require'
/app/config/application.rb:19:in `<top (required)>'
/app/rakefile:4:in `require_relative'
/app/rakefile:4:in `<top (required)>'
/usr/share/rubygems-integration/all/gems/rake-13.0.1/exe/rake:27:in `<top (required)>'
(See full trace by running task with --trace)
```

With these lines included, `rake db:migrate` runs to completion.

migurski: Clarification: the error output above happens on my local machine (MacOS, Docker 2.3). Same process seems happy on GH Actions: https://github.com/migurski/openstreetmap-website/runs/1618711125

mmd-osm: Yeah, on Github, it's pretty much a no-op, as this environment variable isn't being used anywhere b/c config/boot.rb isn't part of this pull request. I also haven't seen this before on Ubuntu 20.04 host.

.github/workflows/docker.yml
migurski: I tested locally and this seems to work well! I’m hesitant to apply this change since we seem to have the “rake” and “Rakefile” naming convention throughout the repo. Maybe we can change this globally in another PR?

docker-compose.yml
migurski: Any advice? I don’t know what Bootsnap is or does, I’m just blindly searching error messages here.

mmd-osm: In the Github issue you've linked to, they also replaced one line in a file named  `config/boot.rb`:

On my local box, there's one line to load the bootsnap/setup module, which is used to speed up Rails application loading:
```
require "bootsnap/setup" # Speed up boot time by caching expensive operations.
```

They replace that line by making the loading process depend  on your new environment variable ENABLE_BOOTSNAP:

```
require 'bootsnap/setup' if ENV.fetch("ENABLE_BOOTSNAP", "true") == "true"
```

migurski: That makes sense — I’ve added this check in `config/boot.rb` and it seems to now work both locally and on GH. Thanks!

migurski:
Update on the test suite: [when running under GH Actions](https://github.com/migurski/openstreetmap-website/runs/1619876640?check_suite_focus=true), we see 7 failures, 18 errors, 0 skips. When running locally, I see 9 failures, 254 errors, 0 skips.

mmd-osm:
You need to add this library to your list of Ubuntu libs to install: https://github.com/openstreetmap/chef/blob/bd1a856797211001b32f47b89d947be41c07e9b4/cookbooks/web/resources/rails_port.rb#L97

migurski:
Great, that brings it down to [just one failure](https://github.com/migurski/openstreetmap-website/runs/1620503777?check_suite_focus=true), the trace test above.

migurski:
Update: tests all passed!

The one remaining failure above was due to `system()` not throwing an exception when `unzip` was missing from the filesystem. "d41d8cd98f00b204e9800998ecf8427e" is the MD5 of an empty string.

migurski:
I believe this pretty ready to merge. @gravitystorm what do you think? Are you okay keeping the full test run in the CI under Docker even though it adds a few minutes of overhead?

docker/postgres/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM postgres:11
+
+# Add db init script to install OSM-specific Postgres functions/extensions.
+ADD docker/postgres/openstreetmap-postgres-init.sh /docker-entrypoint-initdb.d/
+
+# Custom database functions are in a SQL file.
+ADD db/functions/functions.sql /usr/local/sbin/osm-db-functions.sql
```
mmd-osm, line 7: A couple of comments: 

* /usr/local/sbin seems to be a somewhat unusual location for sql scripts. 

* Also, I'm not 100% sure if we need this script at all. At least maptile_for_point (for the /changes endpoint) and xid_to_int4 (for the old osmosis based replication) are likely candidates to be removed in the near future. tile_for_point functionality is probably part of the quad_tile, and might also no longer be needed.

migurski: Do you have an alternate suggestion for a location?

Tests fail when these functions are excluded.

mmd-osm: Do you recall which test did in fact fail?

Regarding the location, can't you use /app/db/functions/functions.sql as is?

migurski: Here’s the failing test suite that was fixed by adding functions.sql: https://github.com/migurski/openstreetmap-website/runs/1619876640?check_suite_focus=true

We don’t use `/app` in the DB container, so it’s not there by default.

mmd-osm: Ah, yes, right, the script need to go to another container. Assumption confirmed, it was the /changes endpoint. I really want to get rid of this code, as it's outdated and not used by anyone except for some single unknown user. (-> re-opened #2486). Once this is done, we don't need the sql script anymore.

```
 Failure:
Api::ChangesControllerTest#test_changes_hours_valid [/app/test/controllers/api/changes_controller_test.rb:91]:
Expected response to be a <2XX: success>, but was a <500: Internal Server Error>
Response body: ActiveRecord::StatementInvalid: PG::UndefinedFunction: ERROR:  function maptile_for_point(integer, integer, integer) does not exist
LINE 1: SELECT COUNT(*) AS count_all, maptile_for_point(latitude, lo...
```

tordans:
@migurski how to I migrate to the newest version? I did a rerun of docker-compose build after rebasing and that ran through. But I see errors afterwards:

### 1

```
docker-compose run --rm web bundle exec rake test:db
Creating openstreetmap-website_web_run ... done
rake aborted!
ActiveRecord::StatementInvalid: PG::InsufficientPrivilege: ERROR:  permission denied to create database
/var/lib/gems/2.7.0/gems/activerecord-6.0.3.4/lib/active_record/connection_adapters/postgresql/database_statements.rb:92:in `exec'
```

### 2

```
$ cp config/docker.database.yml config/database.yml

cp: config/docker.database.yml: No such file or directory
```

### 3

```
$ docker-compose up -d

ERROR:
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml
```

tordans:
I still have issues when running the tests:

```
$ docker-compose run --rm web bundle exec rake test:db
Creating openstreetmap-website_web_run ... done
rake aborted!
ActiveRecord::StatementInvalid: PG::InsufficientPrivilege: ERROR:  permission denied to create database
```

I just re-run the build command. And rebased first. Anything else I need to do?

tordans:
Addition to my Suggestions/Questions-Section above:

* Explain in DOCKER.md how to "uninstall" it all.

migurski:
Thanks for sticking with this @tordans. Sounds like you have a prior version of the Postgres DB from a few versions back. As we’ve been developing this branch locally, these are the two commands I run to clear out the database and containers after every use:

    docker-compose down
    docker volume rm openstreetmap-website_db-data

Does that help?

tordans:
@migurski yes. I removed everything after https://github.com/openstreetmap/openstreetmap-website/pull/2409#issuecomment-752983066 and did a clean reinstall. Tests are running now 🥳 .

tordans:
@migurski I am having trouble uploading a GPX trace. Could that be a docker issue (in the sense, that docker does not have a file system to write to or something like this?) – or do I need to look somewhere else?

**Testcase:**
- http://localhost:3000/traces/new
- Use file https://www.openstreetmap.org/trace/3460088/data (from https://www.openstreetmap.org/user/tordans/traces/3460088)
=> Upload

**Details about the error:**
<details>

```
NoMethodError in TracesController#create
undefined method `id' for nil:NilClass

108 109 110 111 112 113 | end        if @trace.id         flash[:notice] = t ".trace_uploaded"         flash[:warning] = t ".traces_waiting", :count => current_user.traces.where(:inserted => false).count if current_user.traces.where(:inserted => false).count > 4
-- | --

Rails.root: /app

Application Trace | Framework Trace | Full Trace
app/controllers/traces_controller.rb:110:in `create'
app/controllers/application_controller.rb:333:in `better_errors_allow_inline'
config/initializers/compressed_requests.rb:27:in `call'
config/initializers/cors.rb:9:in `call'

Request
Parameters:

{"utf8"=>"✓",
 "authenticity_token"=>"ObG7QWOGKxtg0N3mbIILp7lQXHaOlZba3XKgscyyEAguc5wlb5ImS6xqtnw1qPZvaYIFnCpRafNF8CcjDgu9+g==",
 "trace"=>
  {"gpx_file"=>
    #<ActionDispatch::Http::UploadedFile:0x00007f69da1ac840
     @content_type="application/octet-stream",
     @headers="Content-Disposition: form-data; name=\"trace[gpx_file]\"; filename=\"3460088 (1).gpx\"\r\n" + "Content-Type: application/octet-stream\r\n",
     @original_filename="3460088 (1).gpx",
     @tempfile=#<File:/tmp/RackMultipart20210102-1-imhec7.gpx>>,
   "description"=>"Test 123123123123",
   "tagstring"=>"",
   "visibility"=>"public"},
 "commit"=>"Hochladen"}
```

</details>

mmd-osm:
You need to either create the respective directory first or charge the gpx location here: https://github.com/openstreetmap/openstreetmap-website/blob/master/config/settings.yml#L69

migurski:
We should probably have those directories set up by default in the Docker config. @tordans if you run this command, does it work to upload a trace?

    docker-compose run --rm web mkdir -pv /home/osm/traces /home/osm/images

tordans:
> … if you run this command, does it work to upload a trace?

No unfortunately not; same error message as before.

```
$ docker-compose up -d
Creating network "openstreetmap-website_default" with the default driver
Creating openstreetmap-website_db_1 ... done
Creating openstreetmap-website_web_1 ... done
$ docker-compose run --rm web mkdir -pv /home/osm/traces /home/osm/images
Creating openstreetmap-website_web_run ... done
mkdir: created directory '/home/osm'
mkdir: created directory '/home/osm/traces'
mkdir: created directory '/home/osm/images'
```

migurski:
Right, that just made the directories in a temporary container. I’ve added a call to `mkdir` in the Dockerfile and rebased everything to the default branch. Try updating your local copy and re-building from this newest version? I uploaded a test trace in my local copy and it seems to have worked.

Screen Shot 2021-01-02 at 3 08 34 PM

mmd-osm:
Any thoughts on storing /home/osm contents outside the container? `gpx_files` db table will reference gpx traces in the filesystem that will be gone next time the container gets rebuilt.

migurski:
Sounds like a good idea. I just moved the DB contents into a Docker volume per @Firefishy's earlier suggestion. Would another Docker volume be appropriate for this? I believe they persist between ups/downs.

gravitystorm:
I'm keen to see this finished and merged, but I've run out of time this week for a second review. 

What I should mention is that when I tried it previously it slightly messed up my regular development environment. This is, as far as I can figure out, because when I ran the docker container it wrote some cache files to the `project_repo/tmp` directory *as root* (:nauseated_face:). When I then tried to run rails outside of the container, everything failed since my regular user couldn't update those files. I had to do some `find . -group root | sudo xargs rm` outside of the container to get rid of them.

I'm not sure if it's possible to avoid this, i.e. configure it so that whenever a process within the container writes to the non-containerised filesystem, for it to do so as my user instead of root. But I thought I would mention it in case any Docker experts know how to do this.

migurski:
I think our recent change to follow Grant’s advice on using Docker volumes as well as Jamie’s suggestion to prevent mounting `node_modules` solves this. Previously, `docker-db-data` and `node_modules` were created outside the Docker filesystem, but they no longer are. No other directories are created outside Docker.

migurski:
@mmd-osm I’ve updated this PR with the `/home/osm` directories as persistent Docker volumes based on our conversation above.

migurski:
I’ve rebased this PR to keep it updated. @gravitystorm I’ve confirmed that the directory litter you pointed out in a previous review has been fixed since late last year. I believe we’ve addressed all feedback.

gravitystorm (approved):

gravitystorm:
Great, thanks @migurski and everyone else who contributed!

I did find a first-run issue with the tests, involving i18n-js. But during my investigations (checking the web interface, running the tests again) everything worked, so something triggered the file generation. And now I can't recreate it, so I just mention it in case it comes up again for other people at some point. But I didn't think it was a showstopper, so I'm happy to merge.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,36 @@
...
+        cp config/example.storage.yml config/storage.yml
+        cp config/docker.database.yml config/database.yml
+        touch config/settings.local.yml
+    - name: Build Docker Image
+      run: |
+        docker-compose build
+    - name: Start Docker-Compose
+      run: |
+        docker-compose up -d
+        sleep 15 # let the DB warm up a little
```
tomhughes, line 23: That's a failure waiting to happen... Is there really not a way to actually detect when it's ready instead of guessing?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  - push
+  - pull_request
+jobs:
+  test:
+    name: Docker
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout source
+      uses: actions/checkout@v1
```
tomhughes, line 11: This is out of date - there is a v2 now.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,36 @@
...
+        docker-compose run web bundle exec rake i18n:js:export
+        docker-compose run --rm web osmosis --rx docker/null-island.osm.xml --wd host=db database=openstreetmap user=openstreetmap password=openstreetmap validateSchemaVersion=no
+    - name: Test Basic Website
+      run: |
+        curl -siL http://127.0.0.1:3000 | egrep '^HTTP/1.1 200 OK'
+        curl -siL http://127.0.0.1:3000 | grep 'OpenStreetMap is the free wiki world map'
+        curl -siL http://127.0.0.1:3000/api/0.6/node/1 | grep 'Null Island'
+    - name: Test Complete Suite
+      run: |
+        docker-compose run --rm web bundle exec rails test:db
```
tomhughes, line 36: I always use `rake test:db` but I admit to not being able to keep up with the constant changes about what is run via `rails` and what is run via `rake`...

config/boot.rb
```diff
@@ -1,4 +1,4 @@
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
 
 require "bundler/setup" # Set up gems listed in the Gemfile.
-require "bootsnap/setup" # Speed up boot time by caching expensive operations.
+require "bootsnap/setup" if ENV.fetch("ENABLE_BOOTSNAP", "true") == "true" # Speed up boot time by caching expensive operations.
```
tomhughes, line 4: I'm sorry but there has to be a better solution to the bootsnap problem that polluting core rails setup files with this nonsense.

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+      - db
+
+  db:
+    build:
+      context: .
+      dockerfile: docker/postgres/Dockerfile
+    ports:
+      - "54321:5432"
+    environment:
+      POSTGRES_HOST_AUTH_METHOD: trust
```
tomhughes, line 31: If you're doing trust auth why did you put a password in the database configuration?

docker/postgres/Dockerfile
tomhughes: I'm sorry but I agree that `/usr/local/sbin` is daft - it's not an executable and even if it was why sbin and not bin? I'd suggest lib or share for this.

docker/postgres/openstreetmap-postgres-init.sh
tomhughes: As you appear to be using trust auth there shouldn't be any need to set a password and it would also be best not to set superuser unless have to - the only thing that is likely to be needed for installing the extension which os done as the postgres user anyway.

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+      locales \
+      nodejs \
+      postgresql-client \
+      ruby2.7 \
+      ruby2.7-dev \
+      tzdata \
+      unzip \
+      yarnpkg \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
```
tomhughes, line 31: Why is this removing part of apt's internal state?

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# Install current Osmosis
+RUN curl -OL https://github.com/openstreetmap/osmosis/releases/download/0.47.2/osmosis-0.47.2.tgz \
+ && tar -C /usr/local -xzf osmosis-0.47.2.tgz
```
tomhughes, line 35: Hard coding an omsosmis version number is a bad idea as it will need updating when osmosis changes. Why are we wasting time downloading osmosis anyway? How many people will use it? It's not like it even works very well for loading an API database which I assume is what you have it here for...

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+WORKDIR /app
+
+# Install Ruby packages
+ADD Gemfile Gemfile.lock /app/
+RUN gem install bundler \
+ && bundle install
+
+# Install NodeJS packages
+ADD package.json yarn.lock /app/
+RUN yarnpkg install
```
tomhughes, line 50: This should be done as `bundle exec rake yarn:install` and not by running yarn directly.

config/docker.database.yml
```diff
@@ -0,0 +1,20 @@
+# This configuration is tailored for use with docker-compose. See DOCKER.md for more information.
+
+development:
+  adapter: postgresql
+  database: openstreetmap
+  username: openstreetmap
+  password: openstreetmap
```
tomhughes, line 7: If you're using trust auth (which you appear to be) then you shouldn't need a password.

tomhughes (commented):
Apologies for submitting this after the PR has been merged but I hadn't realised the PR was ready for review.

migurski:
Thanks @gravitystorm, so excited to see this merged! @tomhughes, I’ll follow up on your points in follow-up work and I’ll continue to keep a close eye on any Docker-related issues that emerge in this repo.

docker/postgres/Dockerfile
mmd-osm: Now that the /changes endpoint has been removed, do we need the SQL script at all?

gravitystorm: Unfortunately yes, we still need it. The `tile_for_point` SQL function is not needed for day-to-day operations (and so not required for the tests), since we calculate the tile using the QuadTile gem before storing it, and we again use the QuadTile gem to generate the (giant) select queries.

But there's also the database migrations where the columns were added, and those migrations backfill the tile columns using the SQL function. 

I have much else to say about ancient migrations, but for now, we can't just drop the SQL function entirely.

gravitystorm:
> Apologies for submitting this after the PR has been merged but I hadn't realised the PR was ready for review.

And I'm sorry for merging it without checking if you were happy with it. It's one of the downsides of PRs with 100+ comments, I guess - maybe we should get better at closing and reopening PRs after several rounds of changes. 

And I'm kicking myself about the `sbin` bit, since I'd flagged that up in my own head, but then forgot all about it by the time I'd built and rebuilt the containers a few times :frowning_face:

docker/postgres/Dockerfile
mmd-osm: That's a bit odd that you need those migrations on an empty database inside a container. Is there some way to start fresh with a more modern state of the database sql scripts, and skip any ancient stuff?

gravitystorm: To answer your second question, yes, you run those old migrations - the instructions have `docker-compose run --rm web bundle exec rake db:migrate`

But the wider topic is whether we should keep those old migrations, whether we see the database defined by either a) the sum of all the migrations or b) the structure.sql file (opinions vary), and finally whether if we do drop old migrations should we have a 'fake' starter migration, or should we just keep a certain number of migrations and change all the documentation to use rake db:structure:load (or similar), and finally (finally) if we don't keep all the migrations, how do we ensure that the structure file matches the sum of the migrations that have been run production. So that's not a simple set of decisions.

mmd-osm: Somehow I assumed a new database would be set up using db/structure.sql, and then the contents of the migrations table would define, which additional migrations would be needed, by comparing that table's contents against what's inside db/migrate. Maybe the mechanism works in a completely different way, but then I don't really know what's the purpose of that db/structure.sql is in the first place (sorry, I know this all sounds much like noob questions).

mmd-osm: In any case, running the migration passes on a fresh install even without the db scripts, so I'm even more confused now why they would be needed.

gravitystorm: Let's discuss this on a new issue - otherwise this conversation will never be found!

PR #2228, opened 2019-05-19
Remove SWF gpx trackpoints feature for Potlatch1

mmd-osm:
This pull request removes the SWF gpx trackpoints feature which is only used to fetch gpx trackpoints in Potlatch 1. The code showed up in a static code check tool, and I was able to trigger segfaults.

Due to its limited use today, I didn't see much reason to even try and fix it. Potlatch1 would still continue to function, e.g. for people using with the Undelete function. However, the "G" (for gpx background) will cease to work. I think that's totally acceptable given the alternatives available.

PR #2412, opened 2019-10-23
Remove node/way/relation Search API

gravitystorm:
This API clearly shows its age, having never been fully implemented and simply returning 100 ways (or 2000 relations) at random matching the provided tags.

I'm sad to see this go, since it's my favourite part of the API, given its obvious age, lack of functionality, arbitrary limits and so on. As I said in #2228 I would normally fight tooth and nail against making backwards-incompatible changes to the API without bumping the version (or doing some well-publicised deprecation and removal) but, in this case, I don't really think the additional effort is worth it.

If anyone can show this API is useful, or is actually being used, please speak up!


pnorman:
https://www.openstreetmap.org/api/0.6/search returns a 503 as do the by-type pages

tomhughes:
Well that URL is blocked because it includes nodes and node searching is blocked and has been for years.

Way and relation search do work, but only if you specify value to search for.

mmd-osm:
> If anyone can show this API is useful

I'm also in favor of removing it. The random nature in which results are collected from the DB make it impractical for almost all use cases. The only thing which comes to mind is a check for extremely infrequent tags. Then again, **taginfo** does a much better job at identifying such rare key/values pairs. Querying those (rare) objects on a global scale would still be possible via Overpass API.

tomhughes:
Seems like there is general agreement, and nobody claiming that they are using it, which tallies with the logs on the live service...

PR #2419, opened 2019-10-30
add "gem update --system" before bundle install

phillipcheng:


tomhughes:
No, that's not something we want to recommend because our instructions tell people to install gem using system packages and doing `gem update --system` will install it using gem which creates potential conflicts and confusion.

Issue #737, opened 2014-05-05
Method to exit edit mode is unintuitive


jameshfisher:
After posting [a bug report on the iD project](https://github.com/openstreetmap/iD/issues/2216), @tyrasd informed me that this would be a better place.

As you'll see, I was unable to find how to exit edit mode. I'm now aware that I can "click the logo", but I'm suggesting that that isn't a good solution.

By having an "Edit" button while in "View" mode, and the same "Edit" button bolded while in "Edit" mode, the current UI tells me that the "Edit" button is something like a toggle button: I can click it to toggle in and out of edit mode.

When I saw that this wasn't the case, I then looked for a "close"/"cross" button in a corner, based on the intuition that I had been "dropped into" edit mode, and so I was looking for a way to "escape" it.

When I saw that this also wasn't the case, I gave up. and examined the URLs in order to manually construct it.

Since this process was entirely driven by web conventions/intuitions I've built up over time, I expect that an average web user would follow a similar process and get slightly frustrated like I did.

Normally, "clicking the logo" takes you to the homepage of whatever site you're on. The behavior I expected in the context of OSM is that the logo would take me to `http://www.openstreetmap.org/`, not to `http://www.openstreetmap.org/#map=A/B/C`. The intuition here is that the logo usually acts like a "reset" button.

IMO, an explicit "exit edit mode" button would be more intuitive.

I wonder whether anyone has done any [hallway testing](https://en.wikipedia.org/wiki/Usability_testing#Hallway_testing) on this?

123MT:
I encountered the same frustration and had to google "osm id exit edit" to determine how to exit edit mode.  Why not a "view" button that replaces the edit button+dropdown after edit mode is entered? This would follow same convention as the side menus which toggle menus on and off with subsequent clicks of the icon.  I agree fully with the original poster that this is a very frustrating portion of the UI of an otherwise intuitive map editor.

I will now click on the logo to exit edit mode which makes no sense to me or anyone else except perhaps the original developer.

pgirolami:
Googling "osm how to exit edit mode in id" because I was stuck too brought me here... that must mean something...

timmc:
@tomhughes I think it's pretty clear from this thread that the current user interface is confusing. Clicking the logo of a site *always* means "go back to the home page, lose all navigation state".

Would you accept a PR adding a View tab?

GuidoS:
[...]

PR #2420, opened 2019-11-01
Add "view" tab to main navigation

BrandonXLF:
This is one of my two solutions to the problem of the UI not presenting a clear way to return to the "view" page. I finally realized you can just click the logo, but this isn't very clear and I've seen several people online face the same issue. This solution works by adding a "View" tab to the main navigation menu. Here's an example of it working: viewtab.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -323,6 +323,10 @@ $(document).ready(function () {
       return map.getState();
     };
 
+    page.unload = function () {
+      $("#view_tab").removeClass("current");
+    }
+
     return page;
   };
 
@@ -356,7 +360,6 @@ $(document).ready(function () {
 
     page.unload = function () {
       map.removeObject();
-      $("#view_tab").removeClass("current");
     };
 
     return page;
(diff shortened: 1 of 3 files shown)


HolgerJeromin:
On small screens I dont think we have enough space for another tab.

BrandonXLF:
> On small screens I dont think we have enough space for another tab.

The website will auto-resize (the primary navigation will collapse). With the change, this seems to happen for screens below 715px in width with the change. Currently, this happens at 645px (for me at least). It's a difference for sure, but I don't think there are many devices that would be impacted (according to chrome devtools iPads should be fine).

tomhughes:
I've never really understood why people have such a problem with this - clicking on the site logo to return the home page has been standard web site design for like twenty years now.

Our position has always been that we're so short of screen real estate that we don't want to waste any but I guess we can consider this.

mmd-osm:
Related issue: #737

gravitystorm:
I'm happy with the proposed tab from a UX perspective. But from a technical PoV, I don't know enough about how the javascript stuff works to review it confidently.

AntonKhorev:
Since we merged https://github.com/openstreetmap/openstreetmap-website/pull/5151 we are less likely to merge this because the goal was to save space in the header. Although it still may be possible, then it's going to be a trade of *History* and *Export* for *View* which may still save some space.

PR #2421, opened 2019-11-01
Allow primary tabs to "toggle" states

BrandonXLF:
This is my second solution (the first one being [pull request #2420](https://github.com/openstreetmap/openstreetmap-website/pull/2420)) for the problem of the UI not presenting a clear way to return to the "view" page from edit, history, and export. I finally realized you can just click the logo, but this isn't very clear and I've seen several people online face the same issue. This solution works by having the enabled tab in the primary menu return to the root page. I was expecting this behaviour to be already present since the tabs resemble buttons that can be toggled since, but I can also see the menu as more of a multiple-choice menu, making my first option preferable. The buttons current serve no purpose when they are active, so making them act as a link to the root page (allowing them to be "dissected") makes sense. Here's an example of it working: toggletabs

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -307,7 +307,6 @@ $(document).ready(function () {
     var page = {};
 
     page.pushstate = page.popstate = function () {
-      $("#view_tab").addClass("current");
       map.setSidebarOverlaid(true);
       document.title = I18n.t("layouts.project_name.title");
     };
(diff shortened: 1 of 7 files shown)


BrandonXLF:
https://github.com/openstreetmap/openstreetmap-website/pull/2420

tomhughes:
I think this one is a non-starter - nobody is going to expect tabs to work like that so it defeats the whole of object of making it easier for people.

PR #2427, opened 2019-11-09
Remove expand_bbox endpoint

### LINKED ISSUE #2316: Deprecate and remove expand_bbox call
_Expanding Bounding Box_ API endpoint was once added to work around some performance issues during changeset upload. Today, it isn't necessary in any way from a functional point of view. Could we possibly deprecate and remove it from the API code?

By looking at the production logs, are there any applications still using this endpoint? If it isn't used anymore, could we do this even for API 0.6?

Link: https://wiki.openstreetmap.org/wiki/API_v0.6#Expand_Bounding_Box:_POST_.2Fapi.2F0.6.2Fchangeset.2F.23id.2Fexpand_bbox

mmd-osm:
This pull request removes the `expand_bbox` call, see discussion in #2316 for details. I removed more test cases than I originally anticipated, so a bit of extra scrutiny would be advisable.

Note: Analysis of current production log files to spot potential (non-mainstream) users of this endpoint is still pending at this time.

Closes #2316

### DIFF
--- a/app/abilities/api_ability.rb
+++ b/app/abilities/api_ability.rb
@@ -37,7 +37,7 @@ class ApiAbility
         can [:read, :read_one, :update, :update_one, :delete_one], UserPreference
 
         if user.terms_agreed?
-          can [:create, :update, :upload, :close, :subscribe, :unsubscribe, :expand_bbox], Changeset
+          can [:create, :update, :upload, :close, :subscribe, :unsubscribe], Changeset
           can :create, ChangesetComment
           can [:create, :update, :delete], Node
           can [:create, :update, :delete], Way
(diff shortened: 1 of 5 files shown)


gravitystorm:
Code changes look fine to me. 

We'll need to wait on the sysadmins/OWG letting us know if the call is actually in real-world use.

gravitystorm:
Thanks @mmd-osm !

mmd-osm:
I updated the Wiki page as well to reflect this change: https://wiki.openstreetmap.org/w/index.php?title=API_v0.6&type=revision&diff=1928943&oldid=1922149

PR #2428, opened 2019-11-09
 add SOTM Africa 2019 banner 

enockseth:


tomhughes:
Arg. I do wish people wouldn't open these at the last minute...

enockseth:
Sorry and thanks @tomhughes! Better late than never... :)

Issue #2397, opened 2019-10-16
Possible points needs pluralisation


gravitystorm:
From https://trac.openstreetmap.org/ticket/5296

`notifier.gpx_notification.success.loaded_successfully` cannot be properly translated, since it doesn't allow plural forms. In the original ticket, Tom notes that there are two number interpolations in this string, although that might not be an issue since there's only one object ("points") that needs to be pluralised.

loaded_successfully: loaded successfully with %{trace_points} out of a possible %{possible_points} points.

rubynho:
Hey @gravitystorm, I'd like to solve this issue!

gravitystorm:
> Hey @gravitystorm, I'd like to solve this issue!

Great! Please make a pull request, and feel free to ask any questions.

gravitystorm:
Fixed by #2431

PR #2431, opened 2019-11-12
Notifier message pluralization

rubynho:
### Feature
Correctly pluralize the word point in the notification message based on the given possible points.

The `count` parameter is needed in order to provide the necessary information to i18n choose the right message.

Solves #2397

### DIFF
--- a/app/views/notifier/gpx_success.html.erb
+++ b/app/views/notifier/gpx_success.html.erb
@@ -3,7 +3,5 @@
 <p>
   <%= render :partial => "gpx_description" %>
   <%= t("notifier.gpx_notification.success.loaded_successfully",
-        :trace_points => @trace_points,
-        :possible_points => @possible_points,
-        :count => @possible_points) %>
+        :trace_points => @trace_points, :possible_points => @possible_points, :count => @possible_points) %>
 </p>
--- a/app/views/notifier/gpx_success.html.erb
+++ b/app/views/notifier/gpx_success.html.erb
@@ -2,5 +2,8 @@
 
 <p>
   <%= render :partial => "gpx_description" %>
-  <%= t "notifier.gpx_notification.success.loaded_successfully", :trace_points => @trace_points, :possible_points => @possible_points %>
+  <%= t("notifier.gpx_notification.success.loaded_successfully",
+        :trace_points => @trace_points,
+        :possible_points => @possible_points,
+        :count => @possible_points) %>
 </p>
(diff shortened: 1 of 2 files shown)


gravitystorm:
Thanks for the PR @rubynho !

There is one change required to your PR. The travis build reports this error, from our `erblint` checks:

```
Style/HashSyntax: Use hash rockets syntax.
In file: app/views/notifier/gpx_success.html.erb:5
```

This means use e.g. `:trace_points => @trace_points` and not e.g. `trace_points: @trace_points`

If you can, please make this change, squash your commits and force-push to your branch. This will ensure there is only one commit in the history for this PR.

rubynho:
Hey @gravitystorm, I think the errors on Travis will be solved when updating the other i18n files. Am I right?

tomhughes:
Well yes, but we don't know when that will be, so we can't merge with it like this.

You will need to change the name of the message so that other languages fall back to the english one until they are retranslated - you can't just change the structure of a message like you have done here.

rubynho:
Can someone run the Travis for me? It seems that didn't recognize the forced push.

tomhughes:
I've rerun it (https://travis-ci.org/openstreetmap/openstreetmap-website/builds/611332229) and it has passed but it doesn't seem to have managed to notify github properly.

rubynho:
I pushed another commit to start Travis. I can delete or squash into one commit if you want after the build succeeds.

simon04:
@gravitystorm, @tomhughes, is anything holding us back from merging this PR? I think we shouldn't let first-time contributors wait for four months for getting their PR merged.

tomhughes:
Aside from the fact that it won't work? It's trying to pluralise but that only works if the variable is called `count` and neither variable here is called that.

The core problem that is that there are two numeric variables here and either one could be singular or plural but pluralisation only supports a single controlling variable. Because the two here are related it might just be possible but I can imagine there might be languages where it would be a problem.

gravitystorm:
> Aside from the fact that it won't work? It's trying to pluralise but that only works if the variable is called `count` and neither variable here is called that.

The `:count` is being passed, and the pluralisation will be triggered, even though `%{count}` doesn't appear in the string itself. 

I think this is a neat workaround to avoid changing the parameter names in the existing strings.

> The core problem that is that there are two numeric variables here and either one could be singular or plural but pluralisation only supports a single controlling variable.

Yes, this is a limitation. I've done some research and found very little on the topic ([1](https://stackoverflow.com/questions/24303896/i18n-pluralization-of-a-sentence-with-few-counts) [2](https://stackoverflow.com/questions/45199525/rails-5-can-you-pluralize-two-words-in-a-string-with-two-different-counts)). But I think that in many cases, there only needs to be one noun to pluralise, and so I think we're better off merging this as-is, rather than waiting for perfection.

Issue #2417, opened 2019-10-30
Purpose of user.nearby database column?


gravitystorm:
We currently have a "nearby" column in the users table, added in 2007:

```
#  nearby              :integer          default(50)
```
https://github.com/openstreetmap/openstreetmap-website/commit/0a8c26e596b40da6d428a5c91db0d815d13c89e4#diff-10aaef61b133f90c8313930f89fe7dac

Is this required for anything? The `User.nearby` code [doesn't appear to use it](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/models/user.rb#L199).

gravitystorm:
Closed by https://github.com/openstreetmap/openstreetmap-website/pull/2439

mmd-osm:
Related https://trac.openstreetmap.org/ticket/436

PR #2432, opened 2019-11-13
Ignore the users.nearby column

gravitystorm:
This is the first step of removing the column, see #2417. It needs to be deployed before a migration to remove it, since the columns are cached in ActiveRecord and things break if objects exist in memory
that expect the database column to be there.

Thanks to `strong_migrations` for pointing this out when I tried creating the migration first!

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -46,6 +46,8 @@
 class User < ActiveRecord::Base
   require "xml/libxml"
 
+  self.ignored_columns = ["nearby"]
+
   has_many :traces, -> { where(:visible => true) }
   has_many :diary_entries, -> { order(:created_at => :desc) }
   has_many :diary_comments, -> { order(:created_at => :desc) }


tomhughes:
Well if you really want to pander to `strong_migrations` extremist nonsense...

gravitystorm:
> Well if you really want to pander to `strong_migrations` extremist nonsense...

I'm surprised by your attitude here @tomhughes, it's not very helpful. 

Without doing this, if we just ran the migration to drop the column, then anyone in the middle of doing anything with a User object (i.e. already read from the database and then writing to the database) can receive an error. Creating a user, updating their preferences, etc. So you run the migration, and people see errors. I want to avoid that.

I don't consider this to be "pandering to ... extremist nonsense" and it's a) a  pretty trivial solution and b) not something that I'm burdening you with.

If I haven't yet convinced you that `strong_migrations` is useful, then fine, you don't have to agree. But I find it useful, and I like doing migrations properly, so I'll continue to use it.

tomhughes:
Yes it might cause a breakage for the couple of minutes between me running the migration and then doing a coordinated deployment on the web servers.

I just think that this, like many of `strong_migrations` suggestions, creates a lot of work and litters the source with residual noise, for very minimal gains. I believe I made my views on this known when we originally added it ;-)

gravitystorm:
> I just think that this ... creates a lot of work

Creates a lot of work? The two options are: 

a) run a migration, which breaks things in production, or
b) apply this one-line patch.  Later, run exactly the same migration as in a). Finally, remove this one-line from the source.

Seems like an easy choice to me.

tomhughes:
I was assuming you would leave it in. Yes I know that doesn't actually help but it's easy to think it does unless you think it through carefully ;-)

PR #2434, opened 2019-11-14
Switch to using jbuilder for JSON templates

tomhughes:
It looks like `jsonify-rails` is pretty dead upstream while `jbuilder` is much more alive and is the suggested solution added to all new rails projects by default.

It does use a slightly different format for timestamps which I have currently accepted for the notes API but maybe we should force the old format?


gravitystorm:
I think we should keep the date format matching, to avoid breaking compatibility with anything. I checked and a straightforward `json.date_created note.created_at.to_s` (and `to_s` for the other dates, of course) appears to be sufficient.

It would be good to move all the notes dates (including in the XML) to the same "xmlschema" format as the rest of the API, but let's make that part of the next API version.

tomhughes:
Updated to force the old style timestamps for all relevant fields...

PR #2435, opened 2019-11-14
Proposal for a more refreshing layout

7RST1:
I got kinda bored of the kinda flat and crammed design the website currently has so I made some small changes to the css and below are the results.
bilde
I am by no means expecting this to be implemented directly. This proposed pull request has bugs, like the editing drop-down menu appearing under the search menu. I was just hoping to spark some interest towards a design refresh, with more "breathing room".

I don't know if i added this pull request correctly. I am kinda a noob when it comes to GitHub and Gits overall, but I tried my best. :)


app/assets/stylesheets/common.scss
```diff
@@ -251,9 +251,15 @@ table {
...
-  position: relative;
+  position: absolute;
+  border-radius: 10px;
+  right: 10px;
+  left: 10px;
+  top: 10px;
+  box-shadow: 0px 0px 15px rgba(0,0,0,0.3);
   z-index: 1001;
   font-size: 14px;
+  background-color: white;
```
7RST1: raises the header above website content and changes to make it more "floating"

gravitystorm:
Thanks @7RST1 for your proposal! It's nice to have someone interested in the style of the site, since it hasn't changed much in the last few years.

For this specific proposal though, I don't think that I prefer it to what we have now. I find it hard to provide design feedback, so I'm sorry in advance. But rather than just saying "I don't like it", maybe it's helpful for me to try to explain my thought process:

* It looks a bit strange to me to have a floating menu bar on the other pages on the site. And I think it would be inconsistent to have it look different on some pages and not the others.
* It takes up more vertical space than the current menu, so takes up more room on small screens
* I think it works to even further over-emphasis the map on the front page, which I don't like. The rendered map is just a small part of what OSM is all about, and I would prefer to see the emphasis move towards other aspects of the project. This approach makes the map stretch to every corner.
* I think the floating menu looks like a series of map controls, since they now within the boundary of the map. I don't think that it looks like a menu for the other parts of the site.
* Finally, I don't think it looks particularly modern. Most sites have a flat menu bar across the top, and don't have many (if any) drop shadows around it. I browsed [some website themes](https://themes.getbootstrap.com/) to compare and they have lots of menus, but none like this.

I'm definitely interested in hearing what other people think, and also if you have alternative ideas for the design - either for the front page, or for the main menu, or for other parts of the site too.

7RST1:
Thank you for such thorough feedback!
I am aware myself that the design is far from perfect, and I agree with your points. Though, I am still hoping that the website could be refreshed somehow, and someone better at desining and developing than me would be inspired to do that job.

map93-osm-2:
I definitely agree some design updates would be nice, and I really like the floating search bar; I think it does help somewhat to separate the project from the map itself. I do agree with @gravitystorm's criticism regarding the floating navigation; it would be good to prevent that from taking up too much space.

Regarding the website right now, the largest issues I see with it are:
* There isn't an easy way to discover all the resources available to learn more about OSM, and as a result, it's easy to get stuck. Things like the iD tutorial and https://www.openstreetmap.org/about cover the basics pretty well, but after that the learning curve gets steep really quickly - it'd be good to find ways to make that easier.
* For people who aren't already familiar with OSM, the website doesn't really answer the question of "why" - Why should I care about an open map, and what can I do with it? Answering that question takes a decent amount of research and articles from other websites, which not everyone is going to be willing to do.
* The website is fairly difficult to use on mobile, which is now [52% of all internet traffic](https://gs.statcounter.com/platform-market-share/desktop-mobile-tablet).

With that in mind, I'd propose something like this:

osm-concept

For the navigation items:

* "Map data" (probably with a different name) would help to answer "why" and "how". It could include:
  * Copyright info
  * Cool examples of what other people have done with the data
  * Instructions on how to download the data
  * Commonly-used software to process the data

Alternatively, this could just be a list of links to existing resources (switch2osm, the copyright notice, whatever wiki articles there are on this). But I think it's important to have all of these in a centralized place that's easily accessible to new visitors.

* The wiki is probably the most comprehensive guide to OSM available, it would be really nice to have a link to that be easily visible.

* "Start contributing" would link to the existing sign up page. 

* I'm not sure exactly where the GPS trace page would go. It doesn't seem to be frequently-used enough to justify having a link in the main navigation, but there probably should be a link to it somewhere.

Making this work well on mobile is a separate issue; I'll try to spend some more time coming up with a design for that.

Let me know what you think! I think there's definitely a lot of different arguments that could be made about what the most important resources are to include in the navigation. The UI details could definitely use some more refinement as well; this is just something I threw together in the browser devtools in a few minutes.

7RST1:
Looks much better!
Though, I was thinking about what you said, that OSM is not all about the rendered map on the official website. After a bit of afterthought, maybe this shouldn't be the main website. To shine some light on the data itself, maybe this would do;

- OpenStreetMap.org takes you to a modern looking info page (more animated and fluid than the existing one). This would have info about how to download the data, a bit about the licence, a bit about mapping, and a bit about the humanitarian purposes of OSM.
- Ideally there would be a silent video in the background showcasing different uses of the map, renders and using it for humanitarian purposes.
- There would be a big button that says "Take me to the map!" which directs you to OpenStreetMap.org/map. This sould look like the existing main page. People who would like to not be disturbed by the info page could set a shortcut directly to the rendered map.

This would appeal more towards the potential users that possibly have no experience with maps and such.

tomhughes:
Proposals like that have been discussed to death many, many times in the past...

7RST1:
Okay. Sorry for not doing my research. Thanks though for the answers

Adamant36:
I like the idea of there being a link to the wiki on the main page at least. Having in it on the help page seems a bit wrong since its not just for getting "help" on things. Plus, if it's more prominent maybe it will lead to more contributors.

tomhughes:
I'm not saying it's a bad idea - in fact I think it has some merit. I'm just warning that when it has been suggested before not everybody has agreed with the strategy.

dieterdreist:
there used to be a link to the wiki on the map page before it got
redesigned some years ago. The wiki was then put as the last item on the
secondary "help" page, and this decision was strongly defended in the
following discussions, perceivedly in an attempt of de-emphasizing of the
wiki as a whole (for the avoidance of doubt, this is not directed towards
Tom, Grant or Andy). I agree we should bring it back to the first page.

PR #2439, opened 2019-11-20
Remove the users.nearby column

gravitystorm:
Refs #2417 and #2432


migurski:
Could we consider re-adding this column, as a dummy? Osmosis requires it to exist and fails if it does not. I’ve got a PR open to fix this in https://github.com/openstreetmap/osmosis/pull/54, but there are many copies of Osmosis floating around in packaging managers that won't see this change. 

I found this incompatibility when testing the Docker-compose PR for this repo: https://github.com/openstreetmap/openstreetmap-website/pull/2409#issuecomment-586659968

PR #2442, opened 2019-11-20
Update colours of landuse icons, add icon for natural=scrub

hikemaniac:
This PR updates the colours of most landuse icons from OpenStreetMap Carto in browse.scss and also includes an icon for natural=scrub.

Also, I removed the "school.png" file because it wasn't used in browse.scss.


tomhughes:
Updates colours to what? for what purpose?

Are you saying the colours have change in the stylesheet and this is updating them to match?

hikemaniac:
Indeed the colour scheme of OpenStreetMap Carto has changed a lot over the past five years (this is when most of the browse.scss icons were last updated). With this pull request, I am just matching the colour scheme to the latest version of the map stylesheet as this discrepancy is something that confuses me often when I open the browse view. For some of the tags, the difference is more subtle but I decided to update all of them in one go.

PR #2444, opened 2019-11-23
Make /about/LANG work

### LINKED ISSUE #2402: /copyright/LANG works but /about/LANG doesn't
The copyright page (`/copyright`) can be translated into other languages with a URL. i.e. https://www.openstreetmap.org/copyright/de is it in German. But the About page ( `/about`) doesn't support this feature.

Is it possible to add that, so that one can provide a language specific 'about OSM' link?

tankhiwale:
1. Modified about routes
2. Accept the new params in site#about
3. Update about.html.erb


tomhughes:
Please try to give PRs meaningful titles explaining what they do, don't just reference another ticket.

simon04:
The canonical way already working on every(?) page/route is to append `?locale=XX`, e.g., https://www.openstreetmap.org/about/?locale=de

tomhughes:
That is a test hack for development purposes and is not to be used or promoted for public use.

Issue #2445, opened 2019-11-23
Edit with Remote Control blocked in Safari


pyrog:
Editing with remote editor failed. It display the following message:
>Editing failed - make sure JOSM or Merkaartor is loaded and the remote control option is enabled

JOSM is running. It works fine with Firefox and Chrome.
Clicking directly on the following link work too (display Ok).

Console display this error message:
>Unrecognized Content-Security-Policy directive 'worker-src'.
>[blocked] The page at about:blank was not allowed to display insecure content from http://127.0.0.1:8111/load_and_zoom?left=-70.6644307209015&top=41.95700512499653&right=-70.65615190734864&bottom=41.95621870428267.

Safari Version 13.0.2 (14608.2.40.1.2) macOS Mojave (10.14.6)

See also https://github.com/hotosm/tasking-manager/issues/927

tomhughes:
Sounds like a bug in Safari to me - the `worker-src` directive (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src) specifies valid sources for Worker, SharedWorker, or ServiceWorker scripts and as we don't use any of those we set it to `none`.

I have no idea why Safari has decided to apply it to an iframe load - we do add `http://127.0.0.1:8111` to both `frame-src` and `child_src` in our security policy which is the policy that should be used for such loads (`frame-src` is a newer replacement for `child-src`).

tomhughes:
Note that the error message reported at https://github.com/hotosm/tasking-manager/issues/927#issuecomment-557806742 is different to what you have quoted here, namely:

> [blocked] The page at about:blank was not allowed to display insecure content from http://127.0.0.1:8111/load_and_zoom?left=-70.6644307209015&top=41.95700512499653&right=-70.65615190734864&bottom=41.95621870428267.

That message is refusing to do an http load from an https page which is also a bug/limitation in Safari - other browsers specifically exempt 127.0.0.1 from that restriction.

mmd-osm:
Someone on the forum suggested, that Safari is keeping a HSTS pinning list, and once you've visited a localhost address via HTTPS, localhost will be added to that list. From that point onwards, Safari enforce HTTPS in the future and you will no longer be able to access localhost via HTTP.

The following blog posts describes a way to remove localhost in the HSTS pinning list. I haven't tried it, so ymmv: https://dev.to/tuacker/clear-saved-hsts-pinnings-in-safari-and-macos-9hd

[...]

PR #2446, opened 2019-11-26
Always use http to talk to josm; see #2445

grischard:


tomhughes:
We can drop bowser altogether I think once this is removed?

PR #2448, opened 2019-11-27
Parallel tests

gravitystorm:
I spent some time trying to get the parallel tests to work. The main problem is surrounding the trace tests, since they make assumptions about removing files during the teardown stage. Since all the tests share the same directory, this causes test failures when one test is tearing down while the other is still using the dir.

I've reworked it so that the tests each get their own unique directory, using Dir.mktmpdir.

 However, this appears to introduce some flakeyness to the test suite, and I'm not sure why. I'm sharing this WIP to see if other people also see the flakey tests, see what Travis says, and see if anyone has any better ideas!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -138,7 +138,6 @@ end
 
 # Gems needed for running tests
 group :test do
-  gem "fakefs", :require => "fakefs/safe"
   gem "minitest", "~> 5.1", :platforms => [:ruby_19, :ruby_20]
   gem "rails-controller-testing"
   gem "rubocop"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -206,7 +206,6 @@ GEM
     factory_bot_rails (5.1.1)
       factory_bot (~> 5.1.0)
       railties (>= 4.2.0)
-    fakefs (1.0.0)
     faraday (1.0.0)
       multipart-post (>= 1.2, < 3)
     ffi (1.12.2)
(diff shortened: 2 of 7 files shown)


tomhughes:
Looks to have worked in travis but that may well be single core anyway.

I saw some EEXIST errors from the trace tests on my machines but did see a 3x speedup on a four core machine.

gravitystorm:
Hmm, I've come back to this again this week, but I can't trigger any failures in the tests today :-/

tomhughes:
I didn't get very far with trying to diagnose the errors I was seeing but my best guess is that it's related to the mocking of the filesystem as there didn't seem to be any real EEXIST errors in the strace output so my guess is that they were returned by the mock.

I wonder if the mocking is needed if we're creating unique temp dirs? It has been quite problematic in general because it can only see those things which happen from ruby code...

gravitystorm:
> I wonder if the mocking is needed if we're creating unique temp dirs?

The main thing is that we create symlinks for the images (and traces), and when we run the importer then we overwrite the fixtures. So I was considering just making the factories copy the fixtures rather than making symlinks, which might allow us to ditch all the Fakefs stuff entirely at the cost of lots of copying files around during the tests.

gravitystorm:
> So I was considering just making the factories copy the fixtures rather than making symlinks, which might allow us to ditch all the Fakefs stuff entirely at the cost of lots of copying files around during the tests.

I think this is the most straightforward solution, so I've rebased and made this change too.

Issue #2321, opened 2019-07-17
Better styling for Getting Help page


gravitystorm:
The [Getting Help](https://www.openstreetmap.org/help) page is currently rather unappealing. 

Back in 2014 [we had a PR from @danstowell](https://github.com/openstreetmap/openstreetmap-website/pull/728) with suggestions for improvement. Unfortunately the PR had some unresolved questions around the images, and since then more links have been added to the page and the PR is no longer mergeable.

However, I feel like the basic suggestion of moving to what [Bootstrap describes as "cards"](https://getbootstrap.com/docs/4.3/components/card/) is a valid idea. To test it out further, without having to recreate everything that bootstrap offers, [I crowbarred Bootstrap 4 into the site, and made some mockups](https://github.com/gravitystorm/openstreetmap-website/commit/d4cf08f837965cde30ee741355cf8c491b921ec3).

# Basic cards
![Screenshot from 2019-07-17 17-53-53](https://user-images.githubusercontent.com/360803/61391868-f25d5900-a8bd-11e9-8a26-283ded81203e.png)

# Cards with URL footers
![Screenshot from 2019-07-17 17-54-12](https://user-images.githubusercontent.com/360803/61391870-f25d5900-a8bd-11e9-9ab8-906045de9d83.png)

# Cards with images (and URL footers)
![Screenshot from 2019-07-17 17-55-18](https://user-images.githubusercontent.com/360803/61391871-f25d5900-a8bd-11e9-9770-2d3973b1bfba.png)

I think the overall approach of cards (or something similar) works well. I also experimented with regular lists, and definition lists, without much success.

For the implementation details, we could simply duplicate all the CSS rules from Bootstrap 4 related to the cards. Or we could come up with our own CSS rules that do similar things, but that's outside my own wheelhouse. My preference would be to switch to Bootstrap entirely, and get this kind of thing for free, but that's a much larger topic. So suggestions and advice around this are requested.

Alternatively, other people might have better ideas for sprucing up this page that take a different approach entirely! Again, alternative suggestions would be welcome.

mmd-osm:
I think this topic has some overlap with https://welcome.openstreetmap.org/ although target groups differ a bit. Could we use the Welcome Mat design as a baseline, maybe even some of its contents?

gravitystorm:
[...]

PR #2452, opened 2019-12-04
RFC Use Bootstrap components for styling

gravitystorm:
I'm interested to hear what people think about this idea. I've discussed using bootstrap to guide our design in #2386 and also using the card components to spruce up the help page in #2321. I think a wider goal of replacing our custom (unmaintained) reset/style/grid system with bootstrap would be useful. I'm not suggesting here that we change our colours or layout, just that there's a lot of stuff that we can get (and easily upgrade in future) for free by using it.

I tried the approach of just throwing the whole of bootstrap into our stylesheet, which mostly worked but caused some glitches in rendering dropdowns and tooltips. So this approach instead has most of bootstrap disabled, and only certain components enabled, and keeps most of our styling intact. I started with the help page restyling to demonstrate how we can use one of the components.

The long-term goal with this would be to enable components individually, resolve any conflicts with the existing styles, and remove our custom framework/styling code for stuff that bootstrap can handle for us.

I'm not sure why autoprefixer-rails was constrained in https://github.com/openstreetmap/openstreetmap-website/commit/7c7ff4fdf71ae8f450299ddfdcfc4aa137338d28 so any comments on that are welcome too.


tomhughes:
I think the issue was that autoprefixer-rails 9.x required sprockets 4 and we were still on 3 but we aren't any more so I can likely change that - will have a look later.

gravitystorm:
> I think the issue was that autoprefixer-rails 9.x required sprockets 4 and we were still on 3 but we aren't any more so I can likely change that - will have a look later.

Thanks, I've rebased this branch now that is taken care of.

gravitystorm:
OK, if nobody has any comments about either this PR or using an off-the-shelf style framework in general, then I think we should merge this and move on.

The next steps would include removing our custom reset classes, reworking forms (see e.g. pictures at https://github.com/gravitystorm/openstreetmap-website/issues/37), working though the list of bootstrap components and enabling them all, and working though our stylesheets to see what can be removed and left to bootstrap instead.

app/assets/stylesheets/_bootstrap-custom.scss
```diff
@@ -0,0 +1,44 @@
+/*!
+ * Bootstrap v4.3.1 (https://getbootstrap.com/)
+ * Copyright 2011-2019 The Bootstrap Authors
+ * Copyright 2011-2019 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ */
```
tomhughes: If this is upstream code shouldn't it be in `vendor` instead of `app`? I guess in truth it's a kind of halfway house in that this is a template that you're expected to edit?

Is this expected to go away once we are able to just enable the whole thing?

app/assets/stylesheets/common.scss
```diff
@@ -1,4 +1,5 @@
 @import "parameters";
+@import "bootstrap-custom";
```
tomhughes: How does this find the other file when the file starts with an underscore? Is there some rule where the scss preprocessor searches for files with the underscore?

What's the logic behind preferring a CSS level `@import` over a sprockets level require? Then again we already seem to have a mix of both and I'm not sure how the split was decided...

app/assets/stylesheets/screen-ltr.css
```diff
@@ -1,5 +1,5 @@
 /*
  *= require ltr/common
- *= require bootstrap
+ *= require bootstrap-tooltips
```
tomhughes: Presumably a fairly short term goal will be to replace the existing vendored tooltip and dropdown support with something based on the bootstrap gem instead? I assume you've tested that there isn't any issue with mixing the two in the short term...

app/assets/stylesheets/_bootstrap-custom.scss
gravitystorm: Yes, it's a temporary thing. The [guidelines suggest copy+pasting](https://github.com/twbs/bootstrap-rubygem/blob/ef343b781ca1316ba3953295222578a65bd576c0/README.md#sass-individual-components) that file and working though it.

app/assets/stylesheets/common.scss
gravitystorm: I'm just following the documentation, I don't have any particular views either way.

app/assets/stylesheets/screen-ltr.css
gravitystorm: Yes, I would expect to replace the vendored tooltip stuff and just use the upstream version. Currently the tooltips appear to work exactly as before (specifically black text on white background, bootstrap has vice-versa).

PR #2458, opened 2019-12-11
Use squiggly heredocs instead of rails strip_heredoc

gravitystorm:
Squiggly heredocs were introduced in ruby 2.3 and I prefer the ability to indent the closing heredoc marker.

### DIFF
--- a/test/controllers/api/changesets_controller_test.rb
+++ b/test/controllers/api/changesets_controller_test.rb
@@ -282,7 +282,7 @@ module Api
 
       # simple diff to change a node, way and relation by removing
       # their tags
-      diff = <<CHANGESET.strip_heredoc
+      diff = <<~CHANGESET
         <osmChange>
          <modify>
           <node id='#{node.id}' lon='0' lat='0' changeset='#{changeset_id}' version='1'/>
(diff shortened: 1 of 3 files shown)


tomhughes:
What I'm wondering is why the `Layout/HeredocIndentation` cop wasn't complaining about this, and still doesn't for the remaining `<<-` style here documents in the code...

In rubocop 0.77 it should be on by default and enforcing squiggly but it doesn't seem to be, even after I manually configure it on :-(

tomhughes:
I think the cop doesn't quite do what I expect, or what it's documentation suggests - looking at https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/layout/heredoc_indentation.rb#L75 it will bail out on a non-squiggly here doc if it has any identation. So it will only complain about unindented here docs and suggest indenting them with a squiggly here doc.

gravitystorm:
See also https://github.com/rubocop-hq/rubocop/issues/6512

tomhughes:
I fixed a few more in the AMF code ;-) https://github.com/openstreetmap/openstreetmap-website/commit/7531060a12fcdcadc05cba4e2bf803a87bab66f1

Issue #2430, opened 2019-11-11
Blog post subscription notification emails lacks In-Reply-To header field


mmd-osm:
I'm subscribed to a blog post and receive lots of emails by users replying to that post. As those emails don't include an `In-Reply-To` header, my inbox fills up with pages and pages of messages, rather than producing a threaded view of all emails. Not sure if Rails suppors this at all, but it would be quite handy.

tomhughes:
Well what value would you give it? I'm not even sure what generates the message IDs, ie whether rails does it or whether it leaves it for exim to do, but I am sure it doesn't record them anywhere so it has no idea what value to use in the future.

There's also no way to know which comment a future comment refers to, so would you just have all comments as replies to the original entry? But that never has a message ID at all because no email was ever sent for it... Even if we invented one I'm not sure what a mail client would do with a bunch of messages that claim to reply to a message it has never seen?

mmd-osm:
As an example, GitHub uses an `In-Reply-To: <openstreetmap/openstreetmap-website/pull/2229@github.com>` style header for replies. IIRC, this also works when you hit the "Subscribe" button where you never received the original message for an issue/pull request.

We could use a similar fake header, like `In-Reply-To: <openstreetmap-diary-123456@openstreetmap.org>`. We know that the message should be created for a diary comment, and each diary has a unique numeric identifier.

mmd-osm:
First idea for an implementation would be something along those lines:

```ruby
diff --git a/app/mailers/notifier.rb b/app/mailers/notifier.rb
index b12599981..98d4946d8 100644
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -99,6 +99,8 @@ class Notifier < ApplicationMailer
 
       attach_user_avatar(comment.user)
 
+      headers['In-Reply-To'] = "osm-diary-#{comment.diary_entry.id}@#{Settings.server_url}"
+
       mail :from => from_address(comment.user.display_name, "c", comment.id, comment.digest, recipient.id),
            :to => recipient.email,
            :subject => I18n.t("notifier.diary_comment_notification.subject", :user => comment.user.display_name)
@@ -136,6 +138,8 @@ class Notifier < ApplicationMailer
       @author = @commenter
       attach_user_avatar(comment.author)
 
+      headers['In-Reply-To'] = "osm-note-#{comment.note.id}@#{Settings.server_url}"
+
       subject = if @owner
[...]

PR #2462, opened 2019-12-12
Add In-Reply-To, References headers to email

mmd-osm:
Closes https://github.com/openstreetmap/openstreetmap-website/issues/2430

Context: _I'm subscribed to a blog post/changeset discussion/note and receive lots of emails by users replying to that post. As those emails don't include an In-Reply-To & References header, my inbox fills up with pages and pages of unrelated messages, rather than producing a nice threaded view of all emails._

### DIFF
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -94,13 +94,11 @@ class Notifier < ApplicationMailer
       @readurl = diary_entry_url(comment.diary_entry.user, comment.diary_entry, :anchor => "comment#{comment.id}")
       @commenturl = diary_entry_url(comment.diary_entry.user, comment.diary_entry, :anchor => "newcomment")
       @replyurl = new_message_url(comment.user, :message => { :title => "Re: #{comment.diary_entry.title}" })
-      @ref = "osm-diary-#{comment.diary_entry.id}@#{Settings.server_url}"
       @author = @from_user
 
       attach_user_avatar(comment.user)
 
-      headers["In-Reply-To"] = @ref
-      headers["References"]  = @ref
+      set_references("diary", comment.diary_entry)
 
       mail :from => from_address(comment.user.display_name, "c", comment.id, comment.digest, recipient.id),
            :to => recipient.email,
@@ -129,7 +127,6 @@ class Notifier < ApplicationMailer
       @comment = comment.body
       @owner = recipient == comment.note.author
       @event = comment.event
-      @ref = "osm-note-#{comment.note.id}@#{Settings.server_url}"
 
       @commenter = if comment.author
                      comment.author.display_name
@@ -140,8 +137,7 @@ class Notifier < ApplicationMailer
       @author = @commenter
       attach_user_avatar(comment.author)
 
-      headers["In-Reply-To"] = @ref
-      headers["References"]  = @ref
+      set_references("note", comment.note)
 
       subject = if @owner
                   I18n.t("notifier.note_comment_notification.#{@event}.subject_own", :commenter => @commenter)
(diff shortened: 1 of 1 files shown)


app/mailers/notifier.rb
```diff
@@ -94,11 +94,14 @@ def diary_comment_notification(comment, recipient)
...
       @commenturl = diary_entry_url(comment.diary_entry.user, comment.diary_entry, :anchor => "newcomment")
       @replyurl = new_message_url(comment.user, :message => { :title => "Re: #{comment.diary_entry.title}" })
-
+      @ref = "osm-diary-#{comment.diary_entry.id}@#{Settings.server_url}"
       @author = @from_user
 
       attach_user_avatar(comment.user)
 
+      headers["In-Reply-To"] = @ref
+      headers["References"]  = @ref
```
tomhughes: I think we could reduce the amount of duplication here by having a `set_references` helper function and then just doing `set_references("diary", comment.diary_entry)` etc in each of the three methods?

mmd-osm: Good point, added a new helper method now.

mmd-osm:
I tested this change  on master.apis.dev now. Results look good in Thunderbird: comments to a note from different people, close, reactive, are all in one thread, in they share the same In-Reply-To header.

However, Google Mail now starts merging emails that share the same subject, although the HTTP References header differs. Even worse, different types of messages are now grouped, even they belong to different objects.

I think one way of addressing this would be to use less generic email title (see my comment here: https://github.com/openstreetmap/openstreetmap-website/issues/2430#issuecomment-552435433).

tomhughes:
That sounds very odd... I thought gmail used the Message-ID and any references to link threads.

If anything you'd expect it to fall back to title matching where there is no reference information and adding that ought to improve it!

tomhughes:
Actually is that the problem, that there is no root message with that ID for it to anchor the thread on?

mmd-osm:
The Github notification mails I receive for this pull request don't have a root message, only those two header fields. However, all mails share the same title (e.g.` [openstreetmap/openstreetmap-website] Add In-Reply-To, References headers to email (#2462)`), and that may be the reason why threading works with Github notifications in gmail. Current osm emails don't include anything specific except for the user name that sent the email maybe.

I'm a bit reluctant to push this change as it is now to production, given that probably a fair amount of people uses gmail inbox. Can we somehow refine this on master for a bit without pushing to production, or have it inactive in production?

tomhughes:
I haven't pushed it to production yet.

mmd-osm:
I think https://gsuiteupdates.googleblog.com/2019/03/threading-changes-in-gmail-conversation-view.html —> additional details would be relevant here: besides the header, we need to make sure that messages that should belong together have the same email subject.

mmd-osm:
Before changing any of the existing messages (which is a lot of work), I'd like to try two other things:

* Add new header `X-Entity-Ref-ID`  -> set value to `ref`, check if this fixes threading
* Add new header `X-Entity-Ref-ID`  -> set value to some random value to turn off threading in gmail altogether

tomhughes:
I have https://tomh.apis.dev.openstreetmap.org/ running with `X-Entity-Ref-ID` set to the `ref` value and that seems to stop gmail threading them at all...

mmd-osm:
Thanks! Indeed, this turns threading off in gmail, which I think is ok for the time being.

Assuming we change some of the wording of those messages, we could remove the `X-Entity-Ref-ID `header at a later point in time. In particular there's no need to include `%{user}` in the email subject, as we already have that very same information in the "From:" header (at least on production).

tomhughes:
I've merged that to master now. Doesn't this mean we're just back where we started though?!?

tomhughes:
I guess it may thread for other mail clients.

mmd-osm:
Exactly. I just checked Thunderbird, and it's still threading as expected.

PR #2464, opened 2019-12-16
Improve textarea behaviour

hikemaniac:
This PR adds the `resize: vertical` CSS property to comment textareas to fix a problem, where they could be dragged beyond the sidebar.
Furthermore, I've added a `min-height` to all textareas to not let them be too small to enable the display of a scrolling bar.
I've also added a `border-radius` to form fields, to make them more consistent with what we do with buttons and tables.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -2106,9 +2106,6 @@ textarea {
   padding: 5px;
   width: 100%;
   min-height: 50px;
-}
-
-textarea.comment {
   resize: vertical;
 }
 
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -2092,6 +2092,7 @@ textarea {
   color: #222;
   background-color: #fff;
   border: 1px solid $grey;
+  border-radius: 3px;
   padding: 2px 5px;
   margin: 0;
   width: 200px;
@@ -2104,6 +2105,11 @@ textarea {
 textarea {
   padding: 5px;
   width: 100%;
+  min-height: 50px;
+}
+
+textarea.comment {
+  resize: vertical;
 }
 
 /* Rules for user images */


app/assets/stylesheets/common.scss
```diff
@@ -2092,6 +2092,7 @@ textarea {
   color: #222;
   background-color: #fff;
   border: 1px solid $grey;
+  border-radius: 3px;
```
gravitystorm: This should use the `$border-radius` variable

app/assets/stylesheets/common.scss
```diff
@@ -2104,6 +2105,11 @@ textarea {
 textarea {
   padding: 5px;
   width: 100%;
+  min-height: 50px;
+}
+
+textarea.comment {
+  resize: vertical;
```
gravitystorm: I'd prefer for this to apply to all `textarea` since other places also go haywire when they are resized horizontally. Simply moving this line to the `textarea` section immediately above will be enough.

app/assets/stylesheets/common.scss
hikemaniac: I was initially reluctant to use the variable, because it is only used for the navigation bar buttons in the CSS. All the other items that make use of `border-radius` instead have the 3px hardcoded. I thought that `$border-radius` is only used for navigation buttons and if their border-radius is changed in the future, it might look weird if suddenly all sorts of things (e.g. the browse tag list) also have a larger `border-radius`. Should I change it anyway?

app/assets/stylesheets/common.scss
hikemaniac: Thank you, I just added it with my latest commit. 👍

app/assets/stylesheets/common.scss
gravitystorm: I think this could be debated either way, but it's not super important so let's leave it as you've proposed.

PR #2469, opened 2019-12-19
CSS fixes for headings

gravitystorm:
This fixes a number of issues with the headings, by removing some reset rules (that are acting after the bootstrap reset + defaults and so are messing things up) and by removing some custom heading font sizes and allowing the bootstrap values to be used.

I've overridden the heading sizes for the sidebar temporarily (roughly making h2 the same as bootstrap h3, and h3/h4 the same as bootstrap h5) since there's a bunch of specific style rules preventing me from changing e.g. h2 -> h3. For example, the h2 in sidebar gets extra left padding, since it isn't included in the sidebar browse-section. Fixing all that will come in a future PR.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -5,11 +5,10 @@
 
 /* Minimal CSS reset */
 
-html, body, ul, ol, li, form, fieldset, legend, h1, h2, h3, h4, h5, h6, p, input {
+html, body, ul, ol, li, form, fieldset, legend, input {
   margin: 0;
   padding: 0;
   border: 0;
-  font-size:100%;
 }
 
 fieldset,img { border: 0; }
@@ -542,8 +541,6 @@ body.compact {
   #sidebar {
     float: left;
     width: $sidebarWidth;
-    background: #fff;
-    font-size: 12px;
 
     #sidebar_loader {
       display: none;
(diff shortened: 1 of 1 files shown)


tomhughes:
I've merged this for now but it's definitely going to need some more work as I think things kind of swings things back the other way - the h3/h4 items in the side bar are increased from 13px to 20px in size which makes them nearly 50% larger than before.

app/assets/stylesheets/common.scss
```diff
@@ -542,8 +541,6 @@ body.compact {
   #sidebar {
     float: left;
     width: $sidebarWidth;
-    background: #fff;
```
tomhughes: Removing the background colour here means the welcome message for new users has no background and the text is overlaid directly on the map.

Other uses of the sidebar seem to work somehow, despite not having an explicit backgound colour - they seem to inherit the browser default I think but that doesn't happen with the welcome message?

gravitystorm: OK, I'll investigate and fix this.

woodpeck:
FWIW, the new headings appear quite big to me too; any heading like "Node: Somename (12345)" will now wrap on my screen, and any changeset comment with more than 6 words will wrap too. Useful changeset comments can easily span four lines; if you look at someone's edit history, the sidebar is practically one single rather-big-font headline. I guess one can get used to it but it definitely takes some getting used to...

HolgerJeromin:
Ref #2492

enockseth:
Agree with @woodpeck.  Font sizes way too big for me too.

PR #2473, opened 2019-12-22
travis: change dist to bionic

mmd-osm:
We're using bionic in production, so I figured it would be good to do so in CI, too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: bionic
 language: ruby
 rvm:
   - 2.5.3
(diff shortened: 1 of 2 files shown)


mmd-osm:
Ok, test cases fail with a pretty extensive list of failures, which is quite good, because I can't get those tests to pass on my local bionic machine either. 

```
Error:
UserDiariesTest#test_showing_create_diary_entry:
ActionView::Template::Error: 
    app/views/layouts/_header.html.erb:105
    app/views/layouts/site.html.erb:5
    app/controllers/diary_entries_controller.rb:21:in `new'
    config/initializers/compressed_requests.rb:27:in `call'
    config/initializers/cors.rb:9:in `call'
    test/integration/user_diaries_test.rb:29:in `test_showing_create_diary_entry'
```

PR #2475, opened 2019-12-23
Translate units used by leaflet-locatecontrol

simon04:
This fixes the rather awkward strings "Du bist innerhalb von 2080 meters um diesen Punkt" (de), or "Вы находитесь в 2080 meters от этой точки" (ru).

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -109,6 +109,8 @@ $(document).ready(function () {
     iconLoading: "icon geolocate",
     strings: {
       title: I18n.t("javascripts.map.locate.title"),
+      metersUnit: I18n.t("javascripts.map.locate.metersUnit"),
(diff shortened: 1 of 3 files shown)


tomhughes:
That looks like "lego" translation to me, which is unlikely to work properly in many languages.

I mean I know we're limited here by what leaflet can do, and this is probably better than what we have now, but it's not going to be perfect for all languages.

simon04:
(I have enjoyed playing with LEGO since my childhood :wink:)

Currently, the library doesn't offer other way. However, I've submitted https://github.com/domoritz/leaflet-locatecontrol/pull/258 which would allow to plug-in your i18n system.

tomhughes:
Actually I hadn't realised we were also providing the template that those strings are substituted  into - that makes it less of an issue though the function would have the advantage of allowing plurals to be handled properly.

PR #2485, opened 2019-12-28
JSON output nodes, ways, relations, map

mmd-osm:
This is a follow up for #2221, this time using jbuilder. To avoid large scale changes to routes.rb and test cases, I added some method in the respective controllers to set a default output format "XML".

---

Mandatory motivation slide from _API 0.7; or, The Whale_, by Matt Amos:

xml_yuck


mmd-osm:
Really disappointing. This framework is unable to produce floating numbers with 7 decimal digits without wrapping it in quotation marks. Some people on StackOverflow suggest to monkey patch BigDecimal - whatever they mean by that. Oh well.

This is what we need:

```
  "bounds": {
  "minlat": 0.9924642,
  "minlon": 1.1463021,
  "maxlat": 1.0044411,
  "maxlon": 1.1633345
 },
```

This is what jbuilder comes up with:

```
 "bounds"=>{"minlat"=>"6.9000000", "minlon"=>"6.9000000", "maxlat"=>"7.1000000", "maxlon"=>"7.1000000"}
```

app/controllers/api/map_controller.rb
```diff
@@ -5,6 +5,13 @@ class MapController < ApiController
     before_action :check_api_readable
     around_action :api_call_handle_error, :api_call_timeout
 
+    before_action :default_format_xml
+
+    # Set format to xml unless client requires a specific format
+    def default_format_xml
```
gravitystorm: There's no need to repeat the function definition in each controller. You can move the definition to `app/controllers/api_controller.rb`, since all the controllers in `app/controllers/api/` are subclassed from that one.

app/controllers/api/map_controller.rb
```diff
@@ -5,6 +5,13 @@ class MapController < ApiController
     before_action :check_api_readable
     around_action :api_call_handle_error, :api_call_timeout
 
+    before_action :default_format_xml
```
gravitystorm: We need to be careful about this. Previously in #2065 we ran into problems with accidentally ignoring `Accepts` headers. This approach is different, so perhaps avoids that problem, but we should have test-suite coverage for whatever we are trying to support.

app/views/api/nodes/_node.json.jbuilder
```diff
@@ -0,0 +1,15 @@
+json.type "node"
+json.id node.id
+if node.visible
+  json.lat format("%.7f", node.lat.to_f)
```
gravitystorm: In order to get the XML output the way we want it, we have created our own "Coord" class. This allows us to precisely control what the output is whenever the coord is converted to a string, without having to write "format(%.7f)" all over the codebase.

https://github.com/openstreetmap/openstreetmap-website/blob/c524cbf096d9f67936a5ec6050e5c83fd49077cd/app/models/concerns/geo_record.rb#L4-L15

In that PR I used `.to_f` in the json views for notes, to avoid the builder calling `.to_s` and getting quoted strings. Would that avoid the quoting problem you have here too?

app/controllers/api/map_controller.rb
mmd-osm: Yup, moved that one to the api_controller for now.

app/views/api/nodes/_node.json.jbuilder
mmd-osm: Well,  .to_f omits the quotation issue. However, it will no longer create an output with 7 fixed decimal places, which all other implementations of this format do.

app/controllers/api/map_controller.rb
mmd-osm: I think proper support for `Accept` header is still open. I don't know where the in the processing this information gets lost.

app/views/api/nodes/_node.json.jbuilder
gravitystorm: We might be able to precisely control the output by creating a .to_json method, or similar. It'll probably require investigating exactly how the builder library code determines what to do with these custom objects though.

mmd-osm:
Just one more comment regarding the "visible" flag - this is not present in the original Overpass JSON format. To keep the /map result in line with this format, I also drop the visible flag, in case an object in visible. Only objects with are in fact invisible have an additional `"visible": false` attribute.

app/views/api/nodes/_node.json.jbuilder
mmd-osm: That sounds good. I pretty much gave up on that point after some wasted hours on Stack Overflow.

mmd-osm: I moved formatting to Coord now, and added an _as_json_ method. It's a bit better now, quotation marks are gone, the number has at most 7 decimal places. What is missing is the fixed number of 7 decimal places.

app/controllers/api/map_controller.rb
mmd-osm: default_format_xml now handles an `Accept: application/json` header correctly and returns JSON format in this case.

mmd-osm:
Meanwhile, a first version of a JSON parsing logic has been merged into iD. Once enabled, we're seeing significant speed ups by using JSON, like anticipated back in 2013. As an example > 10s loading + rendering time on XML vs. 3s using JSON.

app/controllers/api/map_controller.rb
mmd-osm: Added a number of unit tests for Accept headers now, so this can probably be set to done now.

app/controllers/api_controller.rb
```diff
@@ -3,6 +3,52 @@ class ApiController < ApplicationController
 
   private
 
+  ##
+  # Set default request format to xml unless a client requests a specific format,
+  # which can be done via (a) URL suffix and/or (b) HTTP Accept header, where
+  # the URL suffix always takes precedence over the Accept header.
+  def set_default_request_format
```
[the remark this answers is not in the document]
mmd-osm: >The edge-case is that if the header exists, and is parseable, but doesn't contain either xml, json, or '/', then request.format isn't set. I'm not sure if that has any consequences though.

That's not really an edge case, leaving request.format as is is intended behaviour and covered by the last Accept header unit test  "text/json is in invalid format, ActionController::UnknownFormat error is expected". If a user requests a single format we don't support at all, we can't fall back to XML here, as it would be violating the respective RFC.

In any case, it would be better to return 406 Not Acceptable instead of HTTP 500. Somehow the Rails framework doesn't handle this properly. In any case, that's outside of anything I control.

>  assume request.format = "xml" on the first line of the function. 

No, don't do that. As I said, it would introduce incorrect behaviour.

> , if you've had enough of refactoring this function already, I understand.

Right. No more refactoring from my side. I needed some code I can still read in a week's time. You're of course free to make it look more Rails like. It would take me more than an hour and the result would still be poor.

app/views/api/nodes/index.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+json.partial! "api/map/root_attributes"
```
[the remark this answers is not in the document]
mmd-osm: I had lots of issues with that fallback logic. Can you suggest something concrete which works here without having me to go through another number of iterations to get it right?

test/controllers/api/map_controller_test.rb
```diff
@@ -23,6 +23,82 @@ def test_routes
...
+      assert_routing(
+        { :path => "/api/0.6/map.json", :method => :get },
+        { :controller => "api/map", :action => "index", :format => "json" }
+      )
+    end
+
+    ##
+    # test http accept headers
+    def test_http_accept_header
+      node = create(:node, :lat => 7, :lon => 7)
```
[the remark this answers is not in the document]
mmd-osm: Ok, I can remove those, np.

app/controllers/api_controller.rb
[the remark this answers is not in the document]
mmd-osm: Wait, I really wanted to add a test case with "`http_accept_format("image/jpeg")`", which behaves in the way like text/json (which is not a valid format). Will to that. The expected outcome would still be HTTP 500 (instead of HTTP 406), and not returning XML.

app/controllers/api_controller.rb
```diff
@@ -3,6 +3,52 @@ class ApiController < ApplicationController
...
+          request.format = "json"
+          break
+        end
+
+        # Any format, not explicitly requesting XML or JSON -> assume XML as default
+        if mime == "*/*"
+          request.format = "xml"
+          break
+        end
+      end
```
mmd-osm: TODO: Add a comment here, that we deliberately don't set request.format, in case not XML, JSON or */* was requested, so that an HTTP error is returned back to the client.

gravitystorm: But this is a breaking change to the API, right? Now, if you request a resource with an Accept header of "text/plain" then you get the XML response. This PR would change that behaviour. That's why I discussed having the default being XML.

Of course, I would propose to fix this in API 0.7 and behave more sensibly wrt Accept headers in the new API version, but that will need to wait until there's progress with that topic.

mmd-osm: That's how the future behaviour looks like vs. what we have on osm.org/master.apis:

`curl -v 'https://master.apis.dev.openstreetmap.org/api/0.6/map?bbox=-61.083984375,45.5371366803986,-61.07849121093749,45.540984218050774' -H "Accept: text/plain`

Returns: HTTP 406 (also for www.openstreetmap.org).

`curl -v 'http://localhost:3000/api/0.6/map?bbox=-61.083984375,45.5371366803986,-61.07849121093749,45.540984218050774' -H "Accept: text/plain"`

500 Internal Server Error - ActionController::UnknownFormat: ActionController::UnknownFormat

gravitystorm:
> On the other hand, some of those methods return plain text, rather than XML.

So that's a big mistake in the API design that I don't want us to make again! I would like to see all json requests getting a json response (and of course fix the plain-text responses for XML requests in 0.7).

If I understand this PR correctly, it adds json endpoints for only a subset of the API, and only for GET requests, is that correct? Do you plan to cover the rest of the API methods in followup pull requests, like changeset upload, and other controllers like traces etc?

app/controllers/api_controller.rb
gravitystorm: Ah, is because cgimap is doing this? I have different results for e.g. my own localhost, or other rails port deployments like OpenHistoricalMap

I think this might be where cgimap is doing so, based on the error messages:
https://github.com/zerebubuth/openstreetmap-cgimap/blob/ee2d8078db851505888ddad275419066a4015cef/src/choose_formatter.cpp#L236

So it is a change in behaviour for the rails port, but given that cgimap already does it's own thing here, it's less of an issue.

app/controllers/api_controller.rb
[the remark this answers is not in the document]
gravitystorm: > It would be better to return 406 Not Acceptable instead of HTTP 500. Somehow the Rails framework doesn't handle this properly. but unfortunately that's outside of anything I control.

I've investigated this, and I believe it's because we wrap all api calls in our own `api_call_handle_error`. You might be able to solve this by adding

```
rescue ActionController::UnknownFormat
   head :not_acceptable
```
to the top of the list of handled exceptions at https://github.com/openstreetmap/openstreetmap-website/blob/f7f73322b3abae8648d6462914eea032a93263fd/app/controllers/application_controller.rb#L186-L204

app/controllers/api_controller.rb
mmd-osm: In both cases the response was generated by cgimap, so yes, Rails port and cgimap behaved differently for many years here.

In general, Accept header support in Rails is quite a delicate topic. There's a number of issues out there, like adding `*/*` would break multi format output, or as in your example, requesting text/plain would happily return XML.

It was the main reason for another round of Accept header parsing in _set_default_request_format_, rather than relying on the framework to do the right thing.

app/controllers/api_controller.rb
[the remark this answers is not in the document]
mmd-osm: Ah, thanks! That's much better, I will change this in a minute.

app/views/api/nodes/index.json.jbuilder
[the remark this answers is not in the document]
mmd-osm: Ok, this wasn't too hard to fix, moved the file one level up and changed the partial to "api/root_attributes"

gravitystorm:
Sorry that I didn't get to finish the review this week, this is still at the top of my todo list.

mmd-osm:
Thanks for the update. By the way, I'm planning to release a new CGImap version 0.8.0 soon, along with a new PPA file for production that includes json support. Can you finish the review in the next couple of weeks maybe?

gravitystorm:
Thanks for your patience - today was the first day for a few weeks that I've been able to spin up the code and poke around with it locally (rather than just reviewing via the website). I wanted to  check a few things in the output before approval.

Everything looks fine to me, so I'm going to merge this. In the (unlikely) event that there's a showstopping bug in the format that none of us have spotted, then I think we should reserve the right to make changes over the next month, even if that breaks compatibility. After that, we'll reserve breaking changes for api version increments, as normal.

tomhughes:
Ah... I would have liked to have given that the once over before we merged it given that some of it is pretty tricky stuff I think.

tomhughes:
I assume @gravitystorm has looked at the actual JSON being generated and is happy with that so I'm not going to do into that in any detail, and the basics of the changes to the controllers mostly look sensible enough.

My real concerns are around `set_default_request_format` both from the point of view of having to call it in all the controllers and also the actual code that is in it - there really ought to be a better solution and I was planning to do some investigation of the problem but it will require some serious deep diving in the rails code I suspect.

At the very least I want to go through what is there with a fine toothed comb and make sure I understand what it is doing before we think about deploying this.

gravitystorm:
> Ah... I would have liked to have given that the once over before we merged it given that some of it is pretty tricky stuff I think.

Ah, sorry about that! I'm happy for you to revert the merge and reopen the PR until you've had the chance to review it.

mmd-osm:
> there really ought to be a better solution and I was planning to do some investigation of the problem but it will require some serious deep diving in the rails code I suspect.

Here are some of the challanges I was facing:
- Ensure backwards compatibility for clients sending broken Accept headers (i.e. JOSM). 
- Ensure backwards compatibility for clients that are not requesting any particular format (*/*)
- In general, XML is always more important than JSON (also backwards compatibility)
- Avoid large scale changes to existing unit tests (setting some default format on API level would not work)
- URL Path suffix is more important than Accept header

We have test cases covering all of those special cases.

I haven't found any approach to achieve those goals without additional logic outside of the Rails framework. People reported a few issues re. Accept header processing on the Rails repo. Hopefully those somewhat more fragile bits of the framework won't bite us anymore now.

tomhughes:
So have you completely ignored the rails processing and done your own from scratch? I'm not sure that isn't even more scary... At the very least it means I now have to read and understand the RFC and check it's right :-(

Can you explain more about the JOSM issue? Does it only affect old versions or does it still affect the current version? Has it been reported?

Large scale changes to the tests may be annoying, but avoiding them is not a good reason to use a suboptimal solution to something - sometimes refactoring is the right thing to do.

All the other points seem reasonable though I don't know yet what that actually means in terms of our options. I hope to investigate over the weekend.

mmd-osm:
> So have you completely ignored the rails processing and done your own from scratch?

I wouldn't call it doing it all from scratch. I'm reusing the same MIME parsing logic the framework also uses and set a default format in case of XML, JSON or `*/*` for selected methods only.

> Can you explain more about the JOSM issue? Does it only affect old versions or does it still affect the current version? Has it been reported?

Sure. What they are currently sending is the following:

```
text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2
```

There are a number of issues here:
- `*` is not a valid mime type, the spec requires `<MIME_type>/<MIME_subtype>`
- The additional space after the semicolon is not permitted
- The number format in q=.2 is invalid, is should be q=0.2

I haven't reported those yet. Given that old versions will linger around for a longish time, we'd need to deal with this in any case.

>  but avoiding them is not a good reason to use a suboptimal solution to something 

I think this would only affect the initial part of the set_default_request_format method, where neither an Accept header, not an url suffix were set (as in the case of unit tests only).

```
      accept_header = request.headers["HTTP_ACCEPT"]
      if accept_header.nil?
        # e.g. unit tests don't set an Accept: header by default, force XML in this case
        request.format = "xml"
        return
      end
```

tomhughes:
I've done a bit of a rework in aaf9d15d5679f97447be0a09790aa63fef0b9022 which is partly just making it more idiomatic ruby and partly slightly changing the way it interacts with rails.

The big issues requiring us to take so much control are (a) the JOSM issue and (b) the fact that rails doesn't really have a way to set a default format as such - it has a hardwired default of `:js` for XHR requests and `:html` for everything else. In fact that second one is the real problem because it we could set a default of `:xml` then it would be used when the JOSM accept failed to parse...

simon04:
The issue has been addressed in JOSM:

https://josm.openstreetmap.de/changeset/15968/josm
> fix `#18812` - HttpClient: specify Accept=*/* to prevent Java from adding Accept=text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2

https://josm.openstreetmap.de/changeset/15969/josm
> see `#18812` - OsmApi, OsmServerReader, NameFinder: specify Accept=application/xml, */*;q=0.8

mmd-osm:
@tomhughes : thanks a lot for the review.  I'm happy that my initial approach wasn't completely off and I haven't missed anything really obvious in the Rails framework.

@simon04 : thanks a lot for the quick fix! While going through your changes, I noticed that CGimap wasn't handling application/xml as expected. This triggered a follow on patch in https://github.com/zerebubuth/openstreetmap-cgimap/pull/221.

PeWu:
FYI, this change broke osm-history: https://github.com/PeWu/osm-history/issues/13

mmd-osm:
@PeWu : as you wrote in the issue, you are simply sending the wrong Accept header:

"Unfortunately, osm-history sends an accept: application/json header"

By the way, JOSM did _not_ have this issue. We have implemented a special case handling to bypass the issue.

gravitystorm:
> @PeWu : as you wrote in the issue, you are simply sending the wrong Accept header:

I suspect that there could be other applications, mainly nodejs ones, that were sending application/json headers without really meaning it (e.g. the http request library they are using does that by default, but the developer was expecting only an XML response since that's the only thing we offered).

I think that's just something each developer will have to adjust to though, I can't imagine a clean way of handling these situations.

mmd-osm:
If people are unable to override their Accept header, there's also the option to append an `.xml` suffix to the URL. 

Some examples: 
- https://api.openstreetmap.org/api/0.6/way/4972913.xml
- https://api.openstreetmap.org/api/0.6/way/4972913/history.xml

tomhughes:
In this case they probably want to convert to JSON anyway - currently they are getting the XML and turning it into JSON immediately!

anatoliegolovco:
Are there plans to implement json response for following endpoint: /api/0.6/changesets ?
We use it for activity monitoring.

tomhughes:
In the general sense that we plan to make the entire API available? Sure. There are no concrete plans that I know of however.

gravitystorm:
> In the general sense that we plan to make the entire API available? Sure. There are no concrete plans that I know of however.

Yep, it just needs volunteers.

prusswan:
Just to confirm, this only adds the _possibility_ of JSON responses from those endpoints? Looks like iD is still requesting XML due to openstreetmap/iD#3765

mmd-osm:
Newest iD v3 uses json.

mmd-osm:
iD 2.18.3 that was deployed to production yesterday is the first release to use JSON.

PR #2490, opened 2020-01-02
Remove the dashed underline from prose headings

gravitystorm:
No other headings are underlined, and the underlines look more like a section separator (e.g. dashed hr) and so visually separated the heading from its content.


tomhughes:
Note that the only place which uses the `prose` class is `oauth_clients#show` and that doesn't have any headings anyway so rich text is what will be affected here.

PR #2494, opened 2020-01-05
Avoid using `raw` with translation strings

gravitystorm:
We use `raw` extensively with translation strings, mainly so that the end result can contain links. This is mostly fine, but sometimes risks a security problem, when any part of the string might have come from user input (e.g. display names) and therefore needs careful handling.

A generally safer way to do this is to use translation keys ending in 'html'. This tells rails that we want to avoid escaping the html, while at the same time allowing rails to still manage potentially unsafe text, like display names, and escaping those properly automatically.

This PR refactors most places where we have previously used `raw`. (There is still raw used in the notifications, in the terms page, and for pagination links.)

@translatewiki This PR contains many translation key renames. Is it helpful for me to list them here for you, or is that not worth doing?

### DIFF
--- a/app/views/browse/_containing_relation.html.erb
+++ b/app/views/browse/_containing_relation.html.erb
@@ -1,7 +1,7 @@
 <li><%= linked_name = link_to h(printable_name(containing_relation.relation)), :action => "relation", :id => containing_relation.relation.id.to_s
         if containing_relation.member_role.blank?
-          raw t ".entry", :relation_name => linked_name
+          t ".entry_html", :relation_name => linked_name
         else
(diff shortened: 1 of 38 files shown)


Abijeet:
> This PR contains many translation key renames. Is it helpful for me to list them here for you, or is that not worth doing?

That should not be necessary. I see that the content for the messages have not been modified, so the new rename functionality should automatically match the renames.

Abijeet:
These renames have been imported into translatewiki.net

gravitystorm:
Thanks both @Abijeet and @tomhughes

PR #2495, opened 2020-01-06
Remove fallback to hostip.info

tomhughes:
We haven't used hostip.info in production for years, except for when we did so by accident, leading to a two hour partial outage when it went down.


Firefishy:
👍

tomhughes:
I was mostly just giving it some time to see of anybody had any objections or other comments.

gravitystorm:
Looks good to me, and always good to remove some complexity in the tests too :smile:

Issue #2488, opened 2019-12-29
Session.id unit test failures 


mmd-osm:
Could the commit below (security fix for CVE-2019-16782) be the reason why our unit tests are failing? We’re using session.id in a header erb and it triggers an error when generating the logout link. I’m seeing this `request.session.id&.public_id` in one of the unit tests: 

https://github.com/rails/rails/commit/92ec9f270d696a17ba7604b3d02931f07c3e7c6a#diff-ddd3ea55daa7e105c86ed353964cfd5b

tomhughes:
Yes I've been looking at it this evening and got as far as working out that `session.id` sometimes fails with the latest `rack` but I'm still working on the details...

tomhughes:
So the story is that `session.id` still works when using the memcache session store but fails when using the in memory cache which is the default - in that case you need `session.id.public_id` but that fails in memcache mode!

So it's working for me locally as I have memcache configured. Travis is also supposed to have it configured but apparently it isn't working...

tomhughes:
I've fixed the travis config in c524cbf096d9f67936a5ec6050e5c83fd49077cd so that it uses memcache as intended which has fixed travis but it is still broken for developers using the default settings.

It's hard to fix until upstream fixes that CVE for memcache, or at least makes `public_id` work for memcache.

mmd-osm:
Thanks! So if anyone comes across the following Runtime error in  `<%= link_to t("layouts.logout"), logout_path(:session => session.id, :referer => request.fullpath), :class => "geolink" %>
`
```
RuntimeError - :
  app/views/layouts/_header.html.erb:105
  app/views/layouts/site.html.erb:5
  app/views/layouts/map.html.erb:87
  app/controllers/application_controller.rb:321:in `better_errors_allow_inline'
  config/initializers/compressed_requests.rb:27:in `call'
  config/initializers/cors.rb:9:in `call'

```

you need to enable memcached by adding the following to config/settings.local.yml and enabling the memcached service:

```
memcache_servers: 
                  - 127.0.0.1
```

tomhughes:
Well you also need to have memcached running ;-)

Legolash2o:
How do you setup/enable the memcached service?

gravitystorm:
I've been digging into this, since I found it strange that we need to use the session id in logout links in the first place. I found that the reason was to avoid people being logged out by simple get requests: https://trac.openstreetmap.org/ticket/2792 (10 years ago).

[...]

PR #2496, opened 2020-01-08
Use a post link to logout

gravitystorm:
This avoids needing to access the session id, which is currently only working with the memcache store.

The fallback page is preserved for anyone who wants to logout without using javascript.

Refs #2488

### DIFF
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -344,27 +344,13 @@ class UsersControllerTest < ActionController::TestCase
   end
 
   def test_logout_without_referer
-    get :logout
-    assert_response :success
-    assert_template :logout
-    assert_select "input[name=referer][value=?]", ""
-
-    session_id = assert_select("input[name=session]").first["value"]
-
-    get :logout, :params => { :session => session_id }
+    post :logout
     assert_response :redirect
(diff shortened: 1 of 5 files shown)


test/controllers/users_controller_test.rb
```diff
@@ -344,27 +344,13 @@ def test_save_referer_params
...
 
   def test_logout_without_referer
-    get :logout
-    assert_response :success
-    assert_template :logout
-    assert_select "input[name=referer][value=?]", ""
-
-    session_id = assert_select("input[name=session]").first["value"]
-
-    get :logout, :params => { :session => session_id }
```
tomhughes: I'm inclined to leave in the get test in method which checks that GET returns the login page - obviously get rid of the assertion that the session ID is in the returned form.

I know you're doing it in the system tests as well, but the cases here do check extra things like that the fallback page preserves the referer?

gravitystorm: Sure, I've added those bits of the tests back in. I've also added a system test that clicks the button and checks the redirect, as well as a system test that checks the referer works for normal behaviour.

(I added the last two as system tests, since I dislike writing controller tests where you check the form, and then try submitting a facsimile of what you hope the browser sends)

Issue #2492, opened 2020-01-03
Headings are way too large


Prince-Kassad:
We had the opposite problem (headings were way too small) up until today. Now they are way too large. This affects all pages where headings are used (i. e. history, user page, etc.)

Using latest Firefox on Win7.

dieterdreist:
sent from a phone

> On 3. Jan 2020, at 10:48, Prince-Kassad <notifications@github.com> wrote:
> 
> We had the opposite problem (headings were way too small) up until today. Now they are way too large. This affects all pages where headings are used (i. e. history, user page, etc.)


does this apply to large screens only? On mobile current heading size seems ok to me.

Prince-Kassad:
Please see the attached screenshot:

![websitebug1](https://user-images.githubusercontent.com/48839736/71771656-2cd7b280-2f3f-11ea-83b3-971a64229a8e.png)

* The changeset comment should not be rendered as a heading
* The font size for the normal text is disproportionate compared to the font size used for the changeset attributes
* The large font means changeset comments can exceed the width of the left pane, overflowing into nothingness where you can't read the changeset comment
* The text container attributes are also wrong and result in lots of unnecessary padding (try using the query tool and you will see what I mean)

As you can see from the top pane (which is unchanged) this is not a DPI/system setting problem. It is a CSS bug in the website itself.

gravitystorm:
Yes, there are a few obvious problems here and I intend to fix them:

* The use of h2 on the sidebars is often too big, but the layout of the sidebar explicitly depends on the first thing being a header, and it being specifically h2. For example, if you use h3 then padding and margins aren't applied. I intend to refactor this so that we can choose headings as we see fit, by making the whole of the sidebar content within a padded div.
* I've no idea why the changeset comment tag was ever in a heading. Previously the heading was changed to the same size as the normal text and made bold. I don't know why a header was therefore used. I think it would be better as regular text.

I will also look into the changeset comments problem. 

>     * The text container attributes are also wrong and result in lots of unnecessary padding (try using the query tool and you will see what I mean)

Can you explain this further please; I don't understand what you are referring to here.

Prince-Kassad:
Please see this second screenshot:

![websitebug2](https://user-images.githubusercontent.com/48839736/71778712-88428880-2fb1-11ea-8341-a800932323db.png)

Pretty sure this didn't happen before. The query results has a lot more (unnecessary) padding than e. g. the history page.

[...]

PR #2497, opened 2020-01-08
Various CSS fixes

gravitystorm:
This PR fixes a few CSS issues that were highlighted in #2492 and others that I noticed while working on that. In summary:

* Uses bootstrap's list-group-flush to replace custom css, for sidebars with lists-with-lines-between
* Uses h6 for the changeset comment tag text
* Removes some unnecessary custom css rules
* Tweaks the rss icon to line it up with other icons

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -214,17 +214,15 @@ OSM.Query = function (map) {
 
           if (interestingFeature(element)) {
             var $li = $("<li>")
-              .addClass("query-result")
+              .addClass("query-result list-group-item")
               .data("geometry", featureGeometry(element))
-              .appendTo($ul);
-            var $p = $("<p>")
               .text(featurePrefix(element) + " ")
-              .appendTo($li);
+              .appendTo($ul);
 
             $("<a>")
               .attr("href", "/" + element.type + "/" + element.id)
               .text(featureName(element))
-              .appendTo($p);
+              .appendTo($li);
           }
         }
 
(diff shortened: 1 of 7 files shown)


tomhughes:
I've added one extra tweak in ffb671b80d3f9105bfbb11ba9f4ef8cf5d0f262a to pull the query tool results back to the left as this change had, probably unintentionally, led to them being indented like this:

before

My change moves them back to the left in line with where they were before:

after

The geocoder search results could probably do with being converted to this new format as well...

gravitystorm:
> probably unintentionally

It was actually deliberate, but I'm not too fussed either way :-) I think the horizontal lines that stretch beyond the left-most character of the heading makes it look like the heading is part of one of the items.

```
  heading

  stuff
---- divider ---
  stuff

  heading

  stuff
---- divider ---
  stuff
---- divider ---
  stuff
```
The above looks like a single list of four items with three dividers to me.
```
  heading

    stuff
  ---- divider
    stuff

  heading

    stuff
  ---- divider
    stuff
  ---- divider
    stuff
```
The above looks more like two lists, with a header before each list and dividers between the list items. 

But like I say, there's perhaps bigger things to worry about! Just thought I should explain my reasoning.

tomhughes:
I think it was the lines sticking out beyond the text that I found odd - the actual indentation wasn't particularly unreasonable.

Actually I think it's the combination - the lines sticking out would work if all the text was indented but having some text align with the lines and some not winds up looking a bit weird to me somehow?

PR #2500, opened 2020-01-08
Avoid double-escaping display names through the translation system

gravitystorm:
They are escaped properly without the h() call, and doing that just double-escapes them.

As part of #2494 I set my local users to have all kinds of weird usernames, and today I spotted some places where they are being double-escaped. I suspect that almost all our uses of h() have been redundant since rails 3.x, but I haven't checked the rest.


tomhughes:
I found another one as well, fixed in 1c062165f695a8fb07653d6331bbb914052c5faf.

PR #2519, opened 2020-01-25
Fix typo: change 'noreferer' to 'noreferrer'

nuntius35:
as the correct rel attribute in external links.


tomhughes:
Are you sure about this? The header is definitely spelt `Referer` even though that is not correct english so are you sure the attribute is different?

nuntius35:
Yes, I'm sure that the name of the header is the only place which is misspelled. Here is one reference I can offer:  https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer
-- 
Andreas Geyer-Schulz
gsa@posteo.de

PR #2521, opened 2020-01-26
Prevent exceptions while reversing directions

HolgerJeromin:
When reversing a route with full or partial empty `to` and `from` inputs, we get an exception.

Steps to reproduce.

- open directions, 
- enter something in `to` input.
- hit reverse directions
- the variable `to` is undefined, but the properties `lat` and `lng` are accessed.

image

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -145,10 +145,10 @@ OSM.Directions = function (map) {
         to = endpoints[1].latlng,
         routeFrom = "",
         routeTo = "";
-    if(from){
+    if (from){
       routeFrom = from.lat + "," + from.lng;
     }
-    if(to){
+    if (to){
       routeTo = to.lat + "," + to.lng;
     }
 
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -142,12 +142,20 @@ OSM.Directions = function (map) {
 
   $(".directions_form .reverse_directions").on("click", function () {
     var from = endpoints[0].latlng,
-        to = endpoints[1].latlng;
+        to = endpoints[1].latlng,
+        routeFrom = "",
+        routeTo = "";
+    if(from){
+      routeFrom = from.lat + "," + from.lng;
+    }
+    if(to){
+      routeTo = to.lat + "," + to.lng;
+    }
 
     OSM.router.route("/directions?" + querystring.stringify({
       from: $("#route_to").val(),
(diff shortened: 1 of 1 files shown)


tomhughes:
I don't think it's this simple - the reverse thing tries to be careful to preserve the actual descriptive names rather than just the latitude and longitude but sometimes that doesn't work properly because things don't roundtrip through the geocoder correctly which is when the problem you are seeing probably occurs. Just using the coordinates is not the right answer though.

HolgerJeromin:
The current code fetches the old strings (even if empty) and fetches the coordinates when valid.
Ignoring invalid coordinates seems to be a good approach.

tomhughes:
I just know there are a whole load of bugs about this stuff and nobody has dared touch it because it's very fragile.

tomhughes:
As far as I can tell the difference is that the old code takes the geocoded results from the input fields and the new code takes the position of the actual map markers but I'm not sure why one would ever exist without the other and, if they're not tied together properly, how we know which is more reliable.

Really I want to understand what is happening and how we get to this state before we can determine if this is the best approach.

tomhughes:
Oh I see you explained that it's when only one endpoint has been entered. That doesn't explain why you are switching the source of the coordinates though?

HolgerJeromin:
I have not (or did not wanted) to change anything when coordinates are valid.
I only checked the validity of the objects and used a default (`""`) in that case.

HolgerJeromin:
I quite changed
`route: to.lat + "," + to.lng + ";" + from.lat + "," + from.lng`
to 
`route: to ? (to.lat + "," + to.lng) : "" + ";" + from ? (from.lat + "," + from.lng) : ""`

But this looks ugly

app/assets/javascripts/index/directions.js
```diff
@@ -142,12 +142,20 @@ OSM.Directions = function (map) {
...
+    if (from) {
+      routeFrom = from.lat + "," + from.lng;
+    }
+    if (to) {
+      routeTo = to.lat + "," + to.lng;
+    }
 
     OSM.router.route("/directions?" + querystring.stringify({
       from: $("#route_to").val(),
       to: $("#route_from").val(),
```
HolgerJeromin: Note that THESE `to` things is the property for the object of the `stringyfy` and not the variable six lines above

HolgerJeromin: We probably should rename the variables if they are too confusing here. But I wanted a minimal fix in the first place.

PR #2524, opened 2020-01-29
Use bootstrap for dropdowns and tooltips

gravitystorm:
This PR switches to using bootstrap for dropdowns and tooltips. It allows us to drop the CSS and Javascript files that we've vendored.

I wanted to do these as two steps, but they both depend on the same javascript files and so it was easier to do them together. Additionally, there was a bit of rework required for the button groups and dropdowns, both because of differing html requirements between the two versions, and also a bit of CSS overrides since our main navigation uses different colours than the rest of the site. Also, three menus are tricky to get right at the same time, particularly when playing with screen widths!

Anyway, I covered every edge case that I could find, and it all seems to be working as intended. After this, I think we can enable the rest of the bootstrap components without any conflicts.


app/assets/javascripts/application.js
```diff
@@ -3,8 +3,8 @@
 //= require jquery.timers
 //= require jquery.cookie
 //= require jquery.throttle-debounce
-//= require bootstrap.tooltip
-//= require bootstrap.dropdown
+//= require popper
+//= require bootstrap-sprockets
```
tomhughes: What's the story here? I assume `bootstrap-sprockets` is loading the bootstrap javascript from the gem but what's `popper` and where is that coming from?

gravitystorm: I'm just following the [README](https://github.com/twbs/bootstrap-rubygem/blob/ef343b781ca1316ba3953295222578a65bd576c0/README.md#a-ruby-on-rails):

> Bootstrap tooltips and popovers depend on popper.js for positioning. The bootstrap gem already depends on the popper_js gem.

> Add Bootstrap dependencies and Bootstrap to your application.js:

> `//= require jquery3`
> `//= require popper`
> `//= require bootstrap-sprockets`

> While bootstrap-sprockets provides individual Bootstrap components for ease of debugging, you may alternatively require the concatenated bootstrap for faster compilation:

> `//= require jquery3`
> `//= require popper`
> `//= require bootstrap`

mmd-osm:
I've noticed that some of the colours changed a bit, e.g. "Log In" and "Sign Up" not have a much darker background when hovered. Similar for tooltips:

Should I expect this kind of change in this PR?

tooltip

tomhughes:
There are a variety of minor changes but I didn't think any of them were a block to merging.

mmd-osm:
I think it wasn't clear to me if  _"I covered every edge case that I could find, and it all seems to be working as intended"_ would imply that the styling didn't change as much.

Maybe my expectation was wrong here, and those changes were intentional (which is ok, I just could figure that out by reading the initial post).

mmd-osm:
Just adding that I’m seeing a huge improvement on my mobile with those bootstrap changes. For the first time since a couple of years the map is properly shown.

gravitystorm:
> I think it wasn't clear to me if _"I covered every edge case that I could find, and it all seems to be working as intended"_ would imply that the styling didn't change as much.

So the edge cases were things like "If I click on the edit menu dropdown, then the border colour reverts to blue" and "If I click on the edit menu dropdown, then with the dropdown open I click outside the browser area, then the dropdown button background reverts to blue" - those are things that were clearly ugly and are the types of edge cases that found and fixed. 

Other things like the colour scheme for tooltips or exactly how grey a background colour is on the signup button - those kind of changes I would call 'semi-deliberate'. By that I mean, they are changes to the defaults that I'm aware of and that I'm happy to accept since I think they look fine.

I'm always happy to hear feedback on these things though, there's so many edge cases (particularly with dropdowns) that I'm sure some styling bugs will slip through now and then.

mmd-osm:
One thing I noticed were dropdowns in RTL languages: https://www.openstreetmap.org/?locale=ar - somehow the white dropdown box seems a bit too narrow so that the text won't fit in there.

PR #2525, opened 2020-01-30
Use an unstyled list for the login buttons

gravitystorm:
Follow fix for a bug I introduced in #2508

### DIFF
--- a/app/views/users/login.html.erb
+++ b/app/views/users/login.html.erb
@@ -51,7 +51,7 @@
 
         <p class='standard-label'><%= t ".with external" %></p>
 
-        <ul class='clearfix' id="login_auth_buttons">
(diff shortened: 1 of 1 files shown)


gravitystorm:
Screenshot from 2020-01-30 10-27-56

mmd-osm:
Apologies for hijacking this PR. could you do the same change for a way's list of nodes: https://www.openstreetmap.org/way/297000506

Those extra black bullets weren't present before, c.f. https://www.openhistoricalmap.org/way/198189321

gravitystorm:
> Apologies for hijacking this PR. could you do the same change for a way's list of nodes: https://www.openstreetmap.org/way/297000506

Looks like there's a bunch of places that I missed. Rustling up a PR now.

PR #2532, opened 2020-02-12
Add dark theme for OSM site and ID editor, depending on color theme 

bezdna:
image

### DIFF
--- a/app/assets/stylesheets/_dark_theme.scss
+++ b/app/assets/stylesheets/_dark_theme.scss
@@ -258,7 +258,7 @@
   #sidebar_content .browse-section {
     border-color: #444;
   }
-
+  #sidebar_content .query-results ul li.selected,
   #sidebar .search_results_entry ul li.selected {
     background: #333;
   }
--- a/app/assets/stylesheets/_dark_theme.scss
+++ b/app/assets/stylesheets/_dark_theme.scss
@@ -1,0 +1,1630 @@
+//  Author: AlexPS
+//  License: CC0 - Public Domain
+//  Origin file: https://userstyles.org/styles/173774/openstreetmap-dark-theme
+
+@media (prefers-color-scheme: dark) {
+  ::-webkit-scrollbar {
+    overflow: visible;
+    background: #282828;
+    border-color: #444;
+  }
+
+  ::-webkit-scrollbar-track {
+    background-clip: padding-box;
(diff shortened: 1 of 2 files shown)


HolgerJeromin:
109 values with `!important` feels very wrong.

mmd-osm:
You probably want to do this via bootstrap, rather than uploading some CSS file than has been tweaked via userstyles.org.

mmd-osm:
_dark_theme.scss will be basically unmaintainable. You should probably look into how this can be achieved via bootstrap instead.

bezdna:
@HolgerJeromin if you know rules cascade,  you not say about this problem.
Because inline styles have highest specificity, and to implement other rules, need use important
button.bar-button.save[style="background: rgb(255, 255, 253);"] {
    background: #333332 !important;
 }

You can read more here https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

bezdna:
@mmd-osm this file change only colores, and this feather work if user set in system or in browser dark mode. I think it better if user waiting dark mode in site he get it.

mmd-osm:
I’m thinking more about projects like https://github.com/ForEvolve/bootstrap-dark

Any additional line of css to support dark mode feels like unneeded extra effort.

HolgerJeromin:
>  if you know rules cascade, you not say about this problem.

I know CSS pretty well. 
If this feature (which I really like) needs 109 hacks it is probably the wrong approach.

I know you got the code from userstyles, but they have no other ways to solve this problem. 
But here we should solve it maintainable.

bezdna:
If take a closer look, you see that this project use all css style from original bootstrap, only use others colours. I suggest use extra css that change only colours.

OSM website not very flexible product, and not lot of changes in UI.  And i think not need a lot of time, to support dark mode.

gravitystorm:
Thanks for the PR, but I'm going to decline this approach. This approach, of having lots of overrides and !important flags and so on is fine when a user style is the only available option. But that's not a good approach when we have full control over the CSS sent to the user.

Also, 1630 lines of CSS is not something I want to maintain. To implement a dark mode, I would ideally like to see out-of-the-box support from bootstrap itself. If that's not available, then maybe 10-20 lines of code to set some colour variables, but little more than that. Anything more complex will be too much to maintain for a low-priority feature like this.

PR #2533, opened 2020-02-13
Change UI element size in small screen

bezdna:
Remove some static size text and margin and padding and change them to REM, now many UI size depending on one class on HTML tag (small, compact, etc)
image
image
image

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -81,11 +81,11 @@ $(document).ready(function () {
     var windowWidth = $(window).width();
 
     if (windowWidth < compactWidth) {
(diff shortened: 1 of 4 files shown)


gravitystorm:
Thanks for your work on this, but I'm going to decline this PR. The current aim of refactoring the stylesheet is to remove most of our custom CSS and use bootstrap defaults instead. This PR edits (but not removes) some of our custom CSS, as well as introducing new rules, so it doesn't fit with our wider plans.

bezdna:
I hope that on mobile screens it will look neat, not horsepower, and I hope that you will not do it for long.

PR #2534, opened 2020-02-13
Fix help page, grid card layout on mobile

bezdna:
before: 
image
 
after: 
image


tomhughes:
@gravitystorm this seems to fix the problem (which I was already aware of) so I have merged - shout if you don't think this was the correct solution.

@bezdna thanks for the fix!

gravitystorm:
Looks good to me!

PR #2535, opened 2020-02-13
add box shadow for header in mobile

bezdna:
I think need separete 2 overlays becase not see what open, they merge, because the background is the same, and this minimal change will help to understand what is open and the main menu too.

**before:**
image
not see what open main menu.

**after:**
image
image
image


bezdna:
image

but it seems to me when any block is open, not the menu, the shadow looks necessary, which separates the header from the block with additional content

gravitystorm:
I think this PR has been abandoned, but feel free to re-open with the requested changes.

PR #2537, opened 2020-02-14
fix welcome message

bezdna:
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -403,7 +403,7 @@ body.compact {
     }
 
     h2 {
-      padding: $lineheight $lineheight*2.5 $lineheight/2 $lineheight;
+      padding: $lineheight 0 $lineheight/2 $lineheight;
       font-size: 1.35rem;
     }
 
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -493,6 +493,18 @@ body.compact {
       }
     }
   }
+  .welcome-header {
+    display: flex;
+    h2 {
+      flex: 1 1 0;
+    }
+    .button {
+      width: auto;
+      border: none;
+      background-color: transparent;
+      padding: 1em;
+    }
+  }
 
   #banner {
     display: none;
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -404,7 +404,7 @@ body.compact {
 
     h2 {
       padding: $lineheight $lineheight $lineheight/2;
-      font-size: 1.5rem;
+      font-size: 1.35rem;
     }
 
     h3, h4 {
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -403,7 +403,7 @@ body.compact {
     }
 
     h2 {
-      padding: $lineheight $lineheight $lineheight/2;
+      padding: $lineheight $lineheight*2.5 $lineheight/2 $lineheight;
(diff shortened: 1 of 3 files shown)


gravitystorm:
We're trying to move away from having custom css, and in particular custom header sizes, so I'm not keen on this solution.

How does this message compare with the other sidebars that have a header and a close icon? Does the same problem appear there? Should the positioning of the close icon be improved or use a different approach?

bezdna:
> We're trying to move away from having custom css, and in particular custom header sizes, so I'm not keen on this solution.

OK.


> How does this message compare with the other sidebars that have a header and a close icon? 
Does the same problem appear there? 

not testing.

> Should the positioning of the close icon be improved or use a different approach?

need use different position type, all elements need one flow, and remove all position absolute
https://getbootstrap.com/docs/4.0/components/modal/#modal-components good example

boothym:
@gravitystorm if you're not going to use this solution then could you please another way to fix it as shouldn't be left as it is.

None of the other sidebar headings interfere with the close icon, because the headings are shorter. So either you need to make the text smaller, made the sidebar wider or move the close icon.

gravitystorm:
> So either you need to [...]

@boothym Patches welcome. You're just as able to help as anyone else is.

bezdna:
WelcomeARABIC_long
WelcomeARABIC
WelcomeEN_focus
WelcomeEN

app/views/layouts/map.html.erb
```diff
@@ -40,8 +40,12 @@
...
     <% unless current_user %>
       <div class="welcome">
-        <h2><%= t "layouts.intro_header" %></h2>
-        <div class="close-wrap"><span class="icon close"></span></div>
+        <div class="welcome-header">
+          <h2><%= t "layouts.intro_header" %></h2>
+          <button type="button" class="button close" aria-label="Close">
+            <span class="icon close"></span>
+          </button>
+        </div>
```
gravitystorm: It might be easier to just put the close icon within the h2, and adjust the javascript handler to match. For example, the search sidebar close button is done like this:

```
<h2>
    <a class="geolink" href="<%= root_path %>"><span class="icon close"></span></a>
    <%= t("site.sidebar.search_results") %>
</h2>
```
It doesn't have any problems with the text overlapping the icon, since it's floated to the right, rather than absolutely positioned.

app/assets/stylesheets/common.scss
```diff
@@ -493,6 +493,18 @@ body.compact {
...
+    h2 {
+      flex: 1 1 0;
+    }
+    .button {
+      width: auto;
+      border: none;
+      background-color: transparent;
+      padding: 1em;
+    }
+  }
```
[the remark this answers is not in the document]
bezdna: h2.flex-grow-1

app/views/layouts/map.html.erb
bezdna: 1. In terms of semantics, a button is not a title element
2. in terms of accessibility, if the button is in a block with a heading, this implies that the button will close the heading, not the block.
3. from the layout point of view, managing floats in RTL and LTR is more difficult, and this should be monitored.

gravitystorm:
@bezdna Thanks for the hints about `d-flex` and `flex-grow-1` - I used your suggestion in #2901 which (hopefully!) fixes this issue.

bezdna:
> @bezdna Thanks for the hints about `d-flex` and `flex-grow-1` - I used your suggestion in #2901 which (hopefully!) fixes this issue.

I'm glad I was helpful and that this task will be closed.

PR #2540, opened 2020-02-17
Improve macOS install notes

saerdnaer:


gravitystorm:
This PR undid the change made in #2539 by re-inserting the ')' into the link. So after this PR, the link is broken again.

tomhughes:
Oh I misread that as being a duplicate change... Will fix...

PR #2542, opened 2020-02-17
Show mailto: links for email values in browse

### LINKED ISSUE #1689: Make email values a link in object detail mode
It would be very helpful and time saving to have the email values of objects being a link, so when you click them, your preferred mail program opens. This is basically a html command (mailto:example@example.org) and this is already what OSM does with phone and fax values (tel:12345). Would it be possible to implement this feature?

boubascript:
Resolves #1689.
This will recognize most, but not all, valid email values and create mailto: links for them.
First used a regex suggested by the [W3C HTML spec](https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address) directly,  but changed to using the built-in `URI::MailTo::EMAIL_REGEXP` in 5d7b09e3bd0a072835d992abc4a5229c3fcb10f1.

A subset of valid emails containing comments, whitespace characters, and quoted strings may not be recognized.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,6 +29,9 @@ Metrics/ClassLength:
   Exclude:
     - 'test/**/*'
 
+Metrics/ModuleLength:
+  Max: 150
+
 Naming/FileName:
   Exclude:
     - 'script/deliver-message'
--- a/app/helpers/browse_tags_helper.rb
+++ b/app/helpers/browse_tags_helper.rb
@@ -21,6 +21,12 @@ module BrowseTagsHelper
       link_to h(wmc[:title]), wmc[:url], :title => t("browse.tag_details.wikimedia_commons_link", :page => wmc[:title])
     elsif url = wiki_link("tag", "#{key}=#{value}")
       link_to h(value), url, :title => t("browse.tag_details.wiki_link.tag", :key => key, :value => value)
+    elsif emails = email_links(key, value)
+      # similarly, email_links() returns an array of emails
+      emails = emails.map do |e|
+        link_to(h(e[:email]), e[:url], :title => t("browse.tag_details.email_link", :email => e[:email]))
+      end
+      safe_join(emails, "; ")
     elsif phones = telephone_links(key, value)
       # similarly, telephone_links() returns an array of phone numbers
       phones = phones.map do |p|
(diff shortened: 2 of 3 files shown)


boubascript:
Adding email to the BrowseTagsHelper caused a rubocop ModuleLength offense, sorry for not catching that before making the pull request.
I'm having a little trouble approaching splitting up the module. Should I move the wiki related tag helpers to a separate module, move the phone and email tag helpers, or something else entirely?

tomhughes:
Don't try and split it up, just change the limit in `.rubocop.yml`.

boubascript:
Thanks! Just confirming, is it preferred to change the Max for Metrics/ModuleLength rather than exclude this specific file?

tomhughes:
That's what we normally do, yes.

gravitystorm:
Generally I prefer to avoid adding lots of code, and gigantic regexps count as lots and lots of code!

Ruby has a built-in regexp for email validation, namely `URI::MailTo::EMAIL_REGEXP` which is recommended for most situations. That way, someone else can be in charge of maintaining and debugging the regexp.

https://stackoverflow.com/questions/22993545/ruby-email-validation-with-regex

I'm not sure if that will work with the semi-colon delineated approach, but it's worth seeing if it can be made to work since I think it's much simpler, easier to read, and less to maintain.

boubascript:
I didn't think of huge regexps counting as lots of code before, but yes using `URI::MailTo::EMAIL_REGEXP` definitely makes a lot more sense here. Thanks for bringing it up,  I don't know how I didn't come across it before. (Side Note: Interestingly enough, it actually currently uses the same exact implementation for the [expression matching an email](https://github.com/ruby/ruby/blob/master/lib/uri/mailto.rb#L56)).

The problem with using it in a semi-colon delineated approach is that it contains the modifiers `\A` and `\z` at the beginning and end respectively, which means it only matches single emails with no trailing whitespace at either end. Building around the regexp using string interpolation directly (i.e using  `#{URI::MailTo::EMAIL_REGEXP}` within the match regex) doesn't work because of this.

I am considering two approaches to dealing with this so semi-colon delineated lists of emails (with possible whitespace around the semicolons) are matched:

* My first thought was to manipulate the built-in regexp literal by converting it to a string and using substring indexing so the `\A` and `\z` are just not included.  The full regex would look like:
`\A\s*(?:#{URI::MailTo::EMAIL_REGEXP.to_s[9..-4]})\s*` \
`(;\s*(?:#{URI::MailTo::EMAIL_REGEXP.to_s[9..-4]})\s*)*\z`.
This feels very hacky and may not be much easier to read, though. 

* Another possible approach would be splitting the value on ';' first, then stripping each possible email and checking if it matches the built-in regexp directly. In the loop, nil would be returned right away if an unmatched email occurs. The only issue I see with this approach is the fact that it may do some unnecessary work when given a semi-colon delineated list containing an invalid email towards the end, since it iterates through the array created by .split() before checking validity.

Thoughts?

mmd-osm:
I ran the following Overpass query on an April 2019 planet:

```
[out:csv(::count, ::"count:nodes", ::"count:ways", ::"count:relations")];
nwr["email"];out count;
nwr["email"~"@"];out count;
nwr["email"~";"];out count;
nwr["contact:email"];out count;
nwr["contact:email"~"@"];out count;
nwr["contact:email"~";"];out count;
```

Test case    | @count | @count:nodes | @count:ways | @count:relations | 
|----|--------|--------------|-------------|------------------| 
| All email | 321586 | 223544       | 93158       | 4884             | 
| @ in email | 320877 | 223116       | 92896       | 4865             | 
| ; in email | 939    | 754          | 167         | 18               | 

Test case    | @count | @count:nodes | @count:ways | @count:relations | 
|----|--------|--------------|-------------|------------------| 
| All contact:email | 106861 | 72930        | 32617       | 1314             | 
| @ in contact:email | 106687 | 72805        | 32572       | 1310             | 
| ; in contact:email | 414    | 318          | 84          | 12               | 


My conclusion would be  that semicolons in email addresses are a fairly niche requirement as they don't even affect 0,5% of all objects. I would just ignore those, it's not worth the effort.

boubascript:
Thanks so much for gathering those stats ! 
I can't really argue with those numbers, it is pretty niche.
I'll make the changes necessary to switch to exclusively picking up single emails.

gravitystorm:
Great, thanks @boubascript and @mmd-osm for your work on this!

Issue #2035, opened 2018-10-26
Unfortunate use of screen estate on mobile


dieterdreist:

mmd-osm:
Point C is already covered in https://github.com/openstreetmap/openstreetmap-website/issues/1663

HolgerJeromin:
On my phone the area A is default collapsed and only visible when touching the menu icon.

malnvenshorn:
It should also be possible to hide the routing directions...

![photo_2019-02-22_23-06-19](https://user-images.githubusercontent.com/13801805/53274198-ab947a00-36f6-11e9-8e48-2ab33711fb74.jpg)

malnvenshorn:
Any plans to make openstreetmap usable on mobile devices?

gravitystorm:
> Any plans to make openstreetmap usable on mobile devices?

@malnvenshorn Please be careful with the way that you phrase your question, since it comes across as a bit rude. The site is already usable on mobile devices, but of course it could be better too. We're a very small group of volunteer developers, and so many issues will wait for a long time. Asking curt questions doesn't make it happen any faster and doesn't help encourage other people to join in.

malnvenshorn:
My question wasn't meant to be rude, but because there was no feedback I took the liberty to ask. Sorry for that.

gravitystorm:
@malnvenshorn no worries. If you want to get started with learning how the code base works and potentially even working on this topic yourself, please let us know!

malnvenshorn:
I'd like to but I have no experience with Ruby/Rails and to many other projects I'm working on to find time to get familiar with the project and used frameworks.

Issue #1864, opened 2018-05-15
Incorrect panning direction when clicking History on mobile browser


andrzej-r:
When I browse osm.org using a mobile browser (portrait orientation) and click "History", the map is shifted to the right (east). It looks like the intention was to make space for the history pane, except that in the portrait mode the pane is displayed above the map, not to the left of it.

This issue does not occur in the landscape mode.

tomhughes:
The only one I've found so far is #729 but it's know that there are a lot of issues around the panning of the map and the opening and closing of various things but nobody has ever managed to drink enough to think about tackling it.

smaprs:
https://user-images.githubusercontent.com/29050750/40054830-88cb0386-581c-11e8-913c-308285352e0a.jpg

smaprs:
I guess this is also about what I've wrongly asked in the  
https://github.com/gravitystorm/openstreetmap-carto/issues/3237  
and so I was told to post it here:  
  
Expected behavior

Rendering standard OSM tiles normally on mobile (cellular android) on history mode.

Actual behavior

Renders only some standard OSM tiles, but blank tiles for the rest, on mobile, on history mode:
like in openstreetmap.org/history#map...
They got truncated at half screen more or less.
Also it displaces from the center of where it was.
Also it shows not well the changeset frames in the area, it shows just as a thiny trace, not as an area.
So can't chose a changeset to verify.
It happens only on mobile and on history mode.
When disabled history mode, rendering returns to normal.
Some other fellow mappers have reported same problem.

The picture link above shows it.
Thanks

tomhughes:
@smaprs I don't see how that is related to this ticket at all so I've going to remove those comments - please open a separate ticket for each issue you you have, but also be aware that there is nobody much working on the mobile view so there is unlikely to be a quick response.

andrzej-r:
What would you expect to be causing it? This could be a good starting issue for new contributors (myself included).

tomhughes:
I'm pretty sure it's not a good starting issue ;-)

Things which are definitely involved are the stylesheet and the javascript that handles routing requests and opening and closing the sidebar. Leaflet may also be involved in that the javascript does odd things to keep the map centred as the sidebar opens and closes.

Issue #1663, opened 2017-10-30
Highlighting objects on mobile


mmd-osm:
I’m opening https://www.openstreetmap.org/way/26904597 on an iPhone 6s in Safari/Chrome. Instead of displaying the object in the map center, it’s shown in the lower left corner. When panning the map the object marker even disappears. In addition there’s a largish grey area where I’d expect a map. Please see the two screenshots below.

![06230697-30b8-4e46-b5d9-66e3db14de16](https://user-images.githubusercontent.com/5842757/32163712-7b1141ce-bd5d-11e7-8d9a-8ef5e5076eed.jpeg)
![3681b74e-5ad4-4942-809c-996f74a0f88d](https://user-images.githubusercontent.com/5842757/32163714-7b315f40-bd5d-11e7-8747-9afa98a6a380.jpeg)

mmd-osm:
BTW: No actual iPhone is required to reproduce this, Firefox Screen size simulation (Ctrl-Shift-M) will do as well. Be sure to select iPhone 6s:

![grafik](https://user-images.githubusercontent.com/5842757/32215647-d25aa900-be22-11e7-8a61-bbf14dd15013.png)

HolgerJeromin:
The disapearing outline is there in chrome/opera on android (blink engine), too. But not in firefox mobile (gecko).

HolgerJeromin:
This is fixed now. Browser update?

mmd-osm:
No it’s not fixed, still the same issue on chrome 63.0 on iPhone

HolgerJeromin:
You are right, sorry. Fixed in Opera 44, but not chrome 63 on android.
But on iOS all browsers uses the safari engine.

bryceco:
In case it’s not clear: this is still broken on safari in iOS.

dieterdreist:
Checking now, it works for notes on iOS (all tiles displayed). There is still an issue if you look at a specific object (/way/123) and close the details the whole map pans approximately one screen width to the left, so that the object you were looking at is off screen

carstenhag:
Broken on chrome beta 77 @ android.

dieterdreist:
it is still broken in iPhone7 with latest Safari and iOS13.1. The tile is not requested, instead a tile is requested that isn't visible at all (only 3 tiles, all of the same column, are requested). This is the whole tile-container:

[...]

PR #2547, opened 2020-02-21
Mobile layout is triggered not by JS, but by media queries, now some map…

### LINKED ISSUE #2349: Tiles not loading properly when opening notes link on phone
Hey, usually I do not have issues with the website, but the tiles, *never* load as they should on my phone. On my Mac, they loas fine.

Using Chrome Beta on a One Plus 5 with Android 9.

When loading the page, only the tiles at the left load. The right stays gray. Panning around the map does not load tiles. When going to another tab in chrome and then revisiting the OSM tab, all tiles load properly.

Example: https://www.openstreetmap.org/note/1885992#map=14/38.5414/-0.1020

Screenshot_20190817-102808

bezdna:
… bugs have gone away themselves

fix #2349 #2035 #1864 #1663

**no blank space:**
image

### DIFF
--- a/app/assets/stylesheets/small.scss
+++ b/app/assets/stylesheets/small.scss
@@ -1,183 +1,187 @@
 @import "parameters";
 
 /* Styles specific to a small screen, such as iPhone, Android, etc... */
+@media (max-width: 640px) {
+  body {
 
-body.small {
-
-  input[type="submit"],
-  input[type="text"] {
-    -webkit-appearance: none;
-  }
-
-  .column-1 {
-    width: 100%;
-  }
-
-  #menu-icon {
-    display: inline-block !important;
-  }
-
-  nav.primary,
-  nav.secondary {
-    float: none !important;
-    position: relative;
-    display: block;
-    clear: both;
-  }
-
-  header {
-    height: auto;
-    min-height: $headerHeight;
-    background: #fff;
-
-    h1 {
-      padding-bottom: 15px;
+    input[type="submit"],
+    input[type="text"] {
+      -webkit-appearance: none;
     }
 
-    &.closed nav {
+    .column-1 {
+      width: 100%;
+    }
+
+    #menu-icon {
+      display: inline-block !important;
+    }
+
+    nav.primary,
+    nav.secondary {
+      float: none !important;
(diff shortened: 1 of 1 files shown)


gravitystorm:
I agree with the general approach here. I hadn't realised until now, but the switch to mobile layout (for the whole page) is based on the rendered width of the menus. If you change one of the menu labels to be really long, you can see that we move the sidebars around even on really wide screens.

I think we should decouple this, and switch the layout based only on the width of the screen, and change only the menus based on the rendered width of the menus.

I can't merge this PR yet, because there's two things that we need to address:

* Choice of breakpoint width. This PR uses a width of 640px, but that's not one of [the breakpoints used by bootstrap](https://getbootstrap.com/docs/4.4/layout/overview/). It would be good to align them, for consistency.
* There's a problem where the secondary nav reappears when the window is shrunk. The aim is that there are three stages - all menus fully shown, then 'compact' mode where the 'more' dropdown appears, and then 'small' where all three menus are combined into one hamburger. But with this PR, there's a width of screen (640px-720px) where the small rules are not applied (because the window is greater than 640), but there's not yet enough width for the compact menu, and so the menus are broken:

Screen Shot 2020-02-26 at 16 37 14

So I would suggest expanding this PR slightly, to fix the menu issue. It's likely fixable just by finding the menu-related style rules that are now inside the `@media` section, and moving them back into a `body.small` section.

It might also be worth moving the javascript-applied classes off of the `<body>` element and onto the `<header>` (in app/views/layouts/_header.html.erb) since they are only going to affect the menu display from now on.

bezdna:
@gravitystorm recent spot changes do not fix the problem well enough?

bezdna:
@gravitystorm two things that you said took into account.
why don't you want to accept pull request?

gravitystorm:
> why don't you want to accept pull request?

No reason, just a lack of time. I don't get much time to work on this each week, and I'm juggling a lot of different topics and trying to make progress on each. I'll get to this soon.

gravitystorm:
Great, I've reviewed this and it works well now. It fixes some, but not all, of the linked issues, so thank you very much for this!

PR #2548, opened 2020-02-26
Rename internal variable to avoid confusion

HolgerJeromin:
The rename is only a refactoring preventing confusion like in https://github.com/openstreetmap/openstreetmap-website/pull/2521#discussion_r370989660

The variables `to` and `from` are not the same as the properties a few lines below.


HolgerJeromin:
friendly ping :-)

PR #2558, opened 2020-03-14
Fix border-radius for search query input field

hikemaniac:
With this PR, I fixed a minor visual bug with the search query input field. The border-radius caused the field to not touch the neighbouring button at its corners, which looked weird in my opinion.

**Before**
Searchbar Before
**After**
Searchbar After

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -764,10 +764,6 @@ header .search_forms,
     overflow: hidden;
     border-radius: 2px 0 0 2px;
   }
-  
-  input#query {
-    border-radius: 3px 0px 0px 3px;
-  }
 
   input[type=text] {
     width: 100%;
@@ -777,6 +773,7 @@ header .search_forms,
 
   input[type=text].overflow {
     border-right: none;
+    border-radius: 3px 0px 0px 3px;
   }
 
   input:focus {
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -764,6 +764,10 @@ header .search_forms,
     overflow: hidden;
     border-radius: 2px 0 0 2px;
   }
+  
+  input#query {
+    border-radius: 3px 0px 0px 3px;
+  }
 
   input[type=text] {
     width: 100%;


pnorman:
Have you tested this with LTR languages?

hikemaniac:
@pnorman You are right, I haven't 😅, thank you for pointing that out.
I fixed it with my latest commit, but I'm not sure if this is the proper way to do this.

tomhughes:
Are you sure you need that and that `r2` won't take care of it? Explicit direction tests like that should be a last resort...

app/assets/stylesheets/common.scss
```diff
@@ -753,6 +753,14 @@ header .search_forms,
   display: none;
 }
 
+[dir=ltr] input#query {
```
tomhughes: As I thought `r2` will take of the RTL stuff so there's no need to do it yourself.

Also please move this inside the `.search_form` block below - as it is it will match any input element with that name anywhere on the site.

tomhughes:
Thinking about it some more why are we making this input field a special case? Is this something we should be applying more widely?

app/assets/stylesheets/common.scss
hikemaniac: Thank you, I just fixed it. 👍

hikemaniac:
> Thinking about it some more why are we making this input field a special case? Is this something we should be applying more widely?

As far as I know, this is the only field, where there is a button directly next to it, so this special case is necessary here.

app/assets/stylesheets/common.scss
```diff
@@ -764,6 +764,10 @@ header .search_forms,
     overflow: hidden;
     border-radius: 2px 0 0 2px;
   }
+  
+  input#query {
```
tomhughes: Now that you've moved I see we already have rules for the input field with the `input[type=text]` selector - this rule should probably just be in there unless there is a good reason not to? Does that other rule select other things?

hikemaniac: I don't see any reason not to. Currently we only have this one field with the `overflow` class and even if there was another field it would make sense to let it also behave like that, so I moved the CSS rule there.

PR #2563, opened 2020-03-18
WIP: Convert database config to support reading from replicas

gravitystorm:
https://guides.rubyonrails.org/active_record_multiple_databases.html

To enable splitting reads and writes we need to convert the database config files. This will involve everyone having to make changes to their local database.yml file too.

I generally don't like using yaml anchors and aliases, but I think this is the easiest way to avoid new developers from getting confused if they change e.g. the hostname for `development -> primary -> database` and forget to make the same change to the `primary_replica` entry too.

For situations where there's no replication (e.g. local development) then the example configs are fine. If you have a replicated database setup, then you can change just the bits you need, for example:

```
  primary_replica:
    <<: *production_db
    host: read-replica.example.com
    user: osmreadonly
    replica: true
```

I made some local tests and it seemed to work out of the box. With this PR, all reads come from the replica, unless the same user session has written to the database in the last 2 seconds. I had a look at the [munin graphs](https://munin.openstreetmap.org/openstreetmap.org/katla.openstreetmap.org/postgres_replication_9_5_main.html) and 2 seconds seems fine to me.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,3 +1,5 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  connects_to :database => { :writing => :primary, :reading => :primary_replica }
 end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -113,9 +113,9 @@ Rails.application.configure do
   # DatabaseSelector middleware is designed as such you can define your own
   # strategy for connection switching and pass that into the middleware through
   # these configuration options.
-  # config.active_record.database_selector = { delay: 2.seconds }
-  # config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
-  # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
+  config.active_record.database_selector = { :delay => 2.seconds }
+  config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
+  config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
 
   # Enable autoloading of dependencies.
   config.enable_dependency_loading = true
--- a/config/example.database.yml
+++ b/config/example.database.yml
@@ -2,28 +2,40 @@
 # See https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md#database-setup for detailed setup instructions.
 #
 development:
-  adapter: postgresql
-  database: openstreetmap
-#  username: openstreetmap
-#  password: openstreetmap
-#  host: localhost
-  encoding: utf8
+  primary: &development_db
+    adapter: postgresql
+    database: openstreetmap
+    # username: openstreetmap
+    # password: openstreetmap
+    # host: localhost
+    encoding: utf8
+  primary_replica:
+    <<: *development_db
+    replica: true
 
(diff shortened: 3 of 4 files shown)


mmd-osm:
From what I’ve seen the mechanism to decide which connection to use is largely based on the http verb: in particular GET would be directed to the replica database while POST/PUT/... hit the main database. I think this assumes that GET involves no database. I’ve read about some other apps where this isn’t the case and wonder wether we have the same issue. I’m thinking here about things like the user confirmation link we send out to new users, and similar.

gravitystorm:
> I’m thinking here about things like the user confirmation link we send out to new users via email, and similar.

Yes, these things might need checking (and at worst, we'll use the "manual connection switching" to a write connection, as described in the docs).

From a brief check of the `users_controller#confirm` and `#confirm_email` methods then both already require a form to be submitted (as should be the case, since email clients and spam filters can auto-crawl links in emails).

I can't off-hand think of other situations where we are currently GETing something that involves database changes, but I'll have a look and see what I can find. If you can think of any then let me know!

mmd-osm:
Here's one candidate: reading a message (via GET /message/id) updates the message_read flag:

```ruby
  # Show a message
  def show
    @title = t ".title"
    @message = Message.find(params[:id])

    if @message.recipient == current_user || @message.sender == current_user
      @message.message_read = true if @message.recipient == current_user
      @message.save            <<<<<<<
    else
```

image

gravitystorm:
> Here's one candidate: reading a message (via GET /message/id) updates the message_read flag:

That's a perfect example! Thanks for finding it. So we have two things we need to do:

* Most importantly, figure out how to run the tests so that they use read-only connections when they are using the `reading` connection. If we don't do this, then there's a risk that the tests will pass and stuff will break in production.
* Figure out what to do on a case-by-case basis. I suspect if we just find a few methods like this (where the write to the database is a side effect of a larger purpose that's read only) then we can wrap the line or two in a block that uses the `writing` connection. 

If anyone has any suggestions on the first topic, I'm all ears. I've marked this WIP in the meantime.

gravitystorm:
> If anyone has any suggestions on the first topic, I'm all ears

OK, I think I know how to make these show situations up in the tests. First, we need to enable the database_selector middleware in the test environment (and we might as well do so for development, too). 

However, the `controller` tests bypass the middleware entirely (partly why controller tests are deprecated - in many ways they don't really reflect an actual response). So only system tests are effective at picking up this problem, since they run the middleware.

Converting `controller` tests to `system` tests would be welcome, but it's a giant project! I'm not sure what to do in the meantime. Given the advantages of having reads-from-replicas enabled, it's worth taking the risk and fixing bugs if and when they come up?

gravitystorm:
> Converting `controller` tests to `system` tests would be welcome, but it's a giant project!

Further investigation shows that there's a middle ground option - updating our controller tests to inherit from `ActionDispatch::IntegrationTest` instead of `ActionController::TestCase` will also work. This is anyway recommended by the rails guide for controller tests (not just integration tests) and also [the documentation for `ActionController::TestCase`](https://api.rubyonrails.org/v6.0.2.2/classes/ActionController/TestCase.html).

It would involve a bunch of refactoring of the tests, but is much less work than rewriting them into system tests, even though system tests are the long term goal.

gravitystorm:
I've started refactoring the tests, but it could take a while. To minimise the number of open PRs, I'm going to close this PR, and reopen a fresh one when the refactoring is complete.

PR #2570, opened 2020-03-25
Use end-of-string termination check when validating urls

gravitystorm:
Flagged up by brakeman (#2229). 

While we debate the preferred configuration format for that tool, it makes sense to fix the problems it flags up in the meantime!

### DIFF
--- a/app/models/client_application.rb
+++ b/app/models/client_application.rb
@@ -39,9 +39,9 @@ class ClientApplication < ApplicationRecord
 
   validates :key, :presence => true, :uniqueness => true
   validates :name, :url, :secret, :presence => true
-  validates :url, :format => %r{\Ahttp(s?)://(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(/|/([\w#!:.?+=&%@!\-/]))?}i
-  validates :support_url, :allow_blank => true, :format => %r{\Ahttp(s?)://(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(/|/([\w#!:.?+=&%@!\-/]))?}i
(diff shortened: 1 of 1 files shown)


tomhughes:
Well maybe I'm missing something, but is this actually fixing anything? and if so what?

As far as I can tell it is actually breaking things, by limiting all three URLs to only a single path component with a maximum of one character?

tomhughes:
Hmm it does actually still match but only because the whole thing is massively broken and winds up matching the whole path against the capture that is only really intended to capture the host, as this [example](https://regex101.com/r/KyhJWj/1) shows.

tomhughes:
If you fix the hostname match to not match delimiters like `:` and `/` then it does [break](https://regex101.com/r/sZ7gEN/1).

tomhughes:
It would probably be more sensible to use something like https://github.com/perfectline/validates_url rather than rolling our own terribly broken regex...

gravitystorm:
> It would probably be more sensible to use something like https://github.com/perfectline/validates_url rather than rolling our own terribly broken regex...

I'm totally in favour of moving away from maintaining our own regexps! I didn't even realise that this one was broken; I was just trying to make the minimal necessary patch.

I'll close this PR and open a fresh one.

> is this actually fixing anything? and if so what?

So my line of thinking is that we have a validator to make sure that it's only valid URLs in these attributes, but due to a flaw in the regexp (no `\z`), you can have inputs that aren't valid URLs. Therefore, we should fix this. Whether that's a security problem or just a user experience problem, I don't think is important.

The side benefit is that this makes one few thing to think about in #2229!

PR #2575, opened 2020-03-31
Remove vendor/assets/openlayers

simon04:
Nowhere used. No longer needed since 2012-08 due to 30ad58c01ff8acae377cfd974fcc94d6898722da.


HolgerJeromin:
I remember many sites were including the js from osm.org
They should not, but removing the asset would break them.

tomhughes:
It is needed because @HolgerJeromin says lots of sites still try to load it from us.

grischard:
Do we have access logs that show how many sites load it from us? We could send them a redirect to a js cdn if a lot of people think that we are one.

tomhughes:
Only if we can find a CDN that still serves OL 2.x...

simon04:
cdnjs has it: https://cdnjs.com/libraries/openlayers/2.12

Firefishy:
The main file accessed is `/openlayers/OpenStreetMap.js` roughly 80k requests per day.

PR #2578, opened 2020-04-01
Ensure that urls are only valid if the entire string is a url

gravitystorm:
Replacement PR for #2557 

The previous regexps only checked that the string started with a url, but would accept further input.

The validates_url gem works well, but only with the inclusion of https://github.com/perfectline/validates_url/pull/85 which is not yet in a release. I'm not sure if it's better to wait for a release, or go with the git link for the gem.

There's also the fact that validates_url doesn't appear to support arbitrary schemes (e.g. `myapp://`) so I've used the underlying regexp for that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,7 @@ gem "rack-cors"
 gem "rails-i18n", "~> 4.0.0"
 gem "rinku", ">= 2.0.6", :require => "rails_rinku"
 gem "strong_migrations"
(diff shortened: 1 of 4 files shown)


tomhughes:
Pulling from git doesn't work in our current deployment environment so it's not an available option.

gravitystorm:
> Pulling from git doesn't work in our current deployment environment so it's not an available option.

I should have remembered that!

PR #2579, opened 2020-04-01
Ensure that urls are only valid if the entire string is a url

gravitystorm:
Replacement PR for #2570 and #2578 

This replaces our homegrown regexps (that didn't quite work) with ruby built-in regexps, and uses the termination anchor to ensure that the entire string, not just the first part, is validated.

### DIFF
--- a/app/models/client_application.rb
+++ b/app/models/client_application.rb
@@ -39,9 +39,9 @@ class ClientApplication < ApplicationRecord
 
   validates :key, :presence => true, :uniqueness => true
   validates :name, :url, :secret, :presence => true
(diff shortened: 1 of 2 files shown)


tomhughes:
I hadn't done anything with this because I wasn't sure there was any point spending time on a temporary solution when we plan to switch to `validates_url` once there is a new release..

gravitystorm:
> I hadn't done anything with this because I wasn't sure there was any point spending time on a temporary solution when we plan to switch to `validates_url` once there is a new release..

The last release was over 12 months ago, and the PR that we need was merged over 6 months ago. I don't think `validates_url` has a release cadence worth waiting for!

PR #2583, opened 2020-04-09
Add blob to CSP frame-src. Fixes #2582

### LINKED ISSUE #2582: Download of osmChange file blocked in iD editor
The current Content Security Policy HTTP header of https://www.openstreetmap.org/edit (iD editor) prevents using the "download osmChange file" link in Firefox. The link is accessible when clicking on the "save" button. Clicking on the link currently triggers no change in the GUI but an error message in the browser console is logged:
 
    Content Security Policy: The page’s settings blocked the loading of a
    resource at blob:https://www.openstreetmap.org/<uuid> ("frame-src").
 
The current CSP is:
 
    default-src 'self'; ... frame-src 'self'; ...
 
I propose that `blob:` should be added after `frame-src`.

grischard:
Fixes #2582


mmd-osm:
btw: putting "Fixes #2582" in the body of this issue (rather than in the header as is now) would auto close issue #2582 once this pull request gets merged.

tomhughes:
I think we need to understand the consequence of this before we add it as global policy - what exactly does that allow?

PR #2596, opened 2020-04-22
Refactor site controller tests to inherit from ActionDispatch::IntegrationTest

gravitystorm:
This is the modern way of writing controller tests, since it uses the rails middleware (among other things) and leads to more realistic tests.

Refs #2563

If we're all happy with this, then I can refactor all the remaining controller tests in order to unblock #2563

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -180,7 +180,7 @@ OpenStreetMap::Application.routes.draw do
   match "/auth/:provider" => "users#auth", :via => [:get, :post], :as => :auth
 
   # permalink
-  get "/go/:code" => "site#permalink", :code => /[a-zA-Z0-9_@~]+[=-]*/
+  get "/go/:code" => "site#permalink", :code => /[a-zA-Z0-9_@~]+[=-]*/, :as => :permalink
 
   # rich text preview
   post "/preview/:type" => "site#preview", :as => :preview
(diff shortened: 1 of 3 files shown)


tomhughes:
Can I just clarify - are the other changes actually required when changing the inheritance or is that just general cleanup?

gravitystorm:
> Can I just clarify - are the other changes actually required when changing the inheritance or is that just general cleanup?

They are actually required. The `get` etc methods now takes either a `foo_path`, `foo_url` or `'/foo/bar'` instead of a symbol, so I went for the `foo_path` style. The session handling needs to be changed too since there's no `:session` keyword for those methods any more.

http://blog.napcs.com/2016/07/03/upgrading-rails-5-controller-tests/ has a few more details. 

There is a choice of whether to pass the parameters to the `get` method, or to the `_path`, e.g.

```
get foo_path, :params => { :bar => :baz }   # both are pretty much
get foo_path(:bar => :baz)                  # equivalent
```
This generally doesn't matter for url query parameters. However the parameters that are created by matching in the routes file (e.g. `code` in `get "/go/:code" => "site#permalink"` defined in config/routes.rb) must be given to the `_path` method e.g.

```
# this doesn't work, since permalink_path needs a `code` param:
get permalink_path, :params => { :code => 'wBz3--', :node => 1}

# this works, since the `code` path parameter is provided to permalink_path:
get permalink_path(:code => 'wBz3--'), :params => { : node => 1 } 

# this also works, since any spare params are treated as query parameters:
get permalink_path(:code => 'wBz3--', :node => 1) 
```

... so I went with the third option throughout, for consistency and less typing/reading.

PR #2597, opened 2020-04-22
Use Open3.capture2 instead of backticks, to avoid command line injection risks

gravitystorm:
In this situation, `trace_name` can be trivially checked as legitimate, but this removes any lingering risks from interpolating into a command line instead of passing parameters explicitly.

Refs #2229

### DIFF
--- a/app/models/trace.rb
+++ b/app/models/trace.rb
@@ -117,7 +117,7 @@ class Trace < ApplicationRecord
   end
 
   def mime_type
-    filetype = `/usr/bin/file -Lbz #{trace_name}`.chomp
+    filetype = Open3.capture2("/usr/bin/file", "-Lbz", trace_name).first.chomp
     gzipped = filetype =~ /gzip compressed/
     bzipped = filetype =~ /bzip2 compressed/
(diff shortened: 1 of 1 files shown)


app/models/trace.rb
```diff
@@ -117,7 +117,7 @@ def trace_name
   end
 
   def mime_type
-    filetype = `/usr/bin/file -Lbz #{trace_name}`.chomp
+    filetype = Open3.capture2("/usr/bin/file", "-Lbz", trace_name)[0].chomp
```
tomhughes, line 120: I tend to prefer `.first` over `[0]` in cases like though I'm not sure there's a really good reason for doing so.

gravitystorm: Me too! I'm not sure why I overlooked this.

gravitystorm:
Anyone any ideas why travis and codecoverage are missing from this PR?

simonpoole:
> 
> 
> Anyone any ideas why travis and codecoverage are missing from this PR?

I can see them ....

gravitystorm:
> I can see them ....

OK, glitch in the matrix it seems.

PR #2598, opened 2020-04-22
Switch to using selenium for system tests

tomhughes:
This switches to using selenium with a headless firefox to run system tests. I would have preferred to use chrome (possibly running firefox as well in travis at least) but https://bugs.chromium.org/p/chromium/issues/detail?id=775911 means that headless chrome is a pain as we can't control `Accept-Language` fully.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ addons:
   postgresql: 9.5
   apt:
     packages:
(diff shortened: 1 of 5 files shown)


tomhughes:
Oh I use firefox myself, but I thought chome might be better for this. I would be nice to be able to test both at some point.

PR #2621, opened 2020-05-09
Request wikidata and osm sophox to linkify some tag value

### LINKED ISSUE #986: Display a link for mapillary keys
Hi,

Could you add a link to `mapillary=*` key of objects (as you do for wikimedia keys) ?

Thanks

### LINKED ISSUE #2405: Link to openplaques
Hi,

Could you add a link to [OpenPlaques](http://openplaques.org) for keys `openplaques:id` ?
See [Key:openplaques:id](https://wiki.openstreetmap.org/wiki/Key:openplaques:id)

Thanks,

https://github.com/tyrasd/overpass-turbo/issues/443

francois2metz:
This is an early PR to gather feedbacks about the refactoring of the value handling when I wanted to fix #2405.

I extracted each formatter to a class. Each class has a format method that returns the html or `nil` if nothing is matching.

I believe the formatters would need to be moved elsewhere. Any suggestions about it? Tests also need to be splitted/moved.

Fix #986


mmd-osm:
The issue I see with this kind of code is that it doesn't really scale. Next week another groups wants to have their special tag du jour linked on osm.org, and we add up with more and more code.

My suggestion would be:

* have pattern matching and replacement rules stored for each tag in Wikidata (on osm wiki)
* write a script to manually pull relevant tag information 
* store updated config.yml file it as part of this repository

We already do something similar for  config/wiki_pages.yml. Maybe you could take a look how that is implemented. When different groups want their tag linkified we would only send them to Wikidata, and run the script afterwards. No more manual fiddling with Rails code.

francois2metz:
This sounds like a good idea. I was digging in the tag2link feature of JOSM, and they use that:

- query to the osm wiki: https://sophox.org/#SELECT%20%28CONCAT%28%22Key%3A%22%2C%20%3Fpermanent_key_ID%29%20as%20%3FOSM_key%29%20%3Fformatter_URL%20%3Fitem%20WHERE%20%7B%0A%09%20%20%3Fitem%20osmdt%3AP2%20osmd%3AQ7.%0A%09%20%20%3Fitem%20osmdt%3AP16%20%3Fpermanent_key_ID.%0A%09%20%20%3Fitem%20osmdt%3AP8%20%3Fformatter_URL.%0A%09%7D
- query to wikidata https://w.wiki/QRF

If we are ok with that, I'm happy to implement it.

tomhughes:
Pretty sure this has already been suggested and discussed in another ticket...

We certainly don't want to be querying anything like that "live" as I assume JOSM does, but that isn't what @mmd-osm suggested.

I'm still wary that it will more or less eliminate any sort of review of what links are desirable in favour of just taking hordes of rules from some random web site. I realise in principle the changes would come here via PRs but what would be the real appetite for reviewing them and adding long term filtering of the results?

config/tag2link.yml
```diff
@@ -0,0 +1,73 @@
+---
+3dmr: https://3dmr.eu/model/$1
+fhrs:id: https://ratings.food.gov.uk/business/en-GB/$1
+ref:bag: https://bagviewer.kadaster.nl/lvbag/bag-viewer/index.html#?searchQuery=$1
+ref:edubase: https://get-information-schools.service.gov.uk/Establishments/Establishment/Details/$1
+ref:ruian:building: https://vdp.cuzk.cz/vdp/ruian/stavebniobjekty/$1
+website: https://web.archive.org/web/*/$1/
```
mmd-osm, line 7: This one looks strange

config/tag2link.yml
```diff
@@ -0,0 +1,73 @@
...
+mdb_id: https://molendatabase.nl/nederland/molen.php?nummer=$1
+ref:FR:LaPoste: http://www.laposte.fr/particulier/outils/trouver-un-bureau-de-poste/bureau-detail/$1/$1
+ref:FR:MemorialGenWeb: http://www.memorialgenweb.org/memorial3/html/fr/resultcommune.php?idsource=$1
+contact:yelp: https://www.yelp.com/biz/$1
+flag:wikidata: https://www.wikidata.org/wiki/$1
+ref:FR:museofile: https://www.pop.culture.gouv.fr/notice/museo/$1
+wikidata: https://www.wikidata.org/wiki/$1
+contact:phone: tel:$1
+phone: tel:$1
+image: https://commons.wikimedia.org/wiki/File:$1
```
mmd-osm, line 21: Same here

config/tag2link.yml
```diff
@@ -0,0 +1,73 @@
...
+ref:gnis: https://geonames.usgs.gov/apex/f?p=gnispq:3:::NO::P3_FID:$1
+ref:nrhp: https://npgallery.nps.gov/AssetDetail/NRIS/$1
+ref:isil: https://w3id.org/isil/$1
+seamark:radio_station:mmsi: https://www.marinetraffic.com/ais/details/ships/$1
+seamark:virtual_aton:mmsi: https://www.marinetraffic.com/ais/details/ships/$1
+ref:woonplaatscode: https://bag.basisregistraties.overheid.nl/bag/id/woonplaats/$1
+ref:arlhs: http://wlol.arlhs.com/lighthouse/$1.html
+ref:nga: https://tools.wmflabs.org/wikidata-externalid-url/?url=https%3A%2F%2Fmsi.nga.mil%2FqueryResults%3Fpublications%2Fngalol%2Flights-buoys%3Fvolume%3D%251%26featureNumber%3D%252%26includeRemovals%3Dfalse%26output%3Dhtml&exp=(%5Cd%7B3%7D)-(.*)&id=$1
+ref:uscg: https://tools.wmflabs.org/wikidata-externalid-url/?url=https%3A%2F%2Fmsi.nga.mil%2FqueryResults%3Fpublications%2Fuscgll%3Fvolume%3D%251%26featureNumber%3D%252%26includeRemovals%3Dfalse%26output%3Dhtml&exp=(%5Cd)-(.*)&id=$1
+ref:ccg: http://google.com/search?q=$1+site:www.notmar.gc.ca/publications/list-livre
```
mmd-osm, line 51: Google search to resolve a link?

mmd-osm:
So reviewing the list should be feasible, especially once we've established an initial set of links and only add small deltas later on.

I wonder if we should introduce a blacklist for things like `image`, `ref:ccg`, and `website` - either because we already have rules for them, or because they resolve to things like a web search.

config/tag2link.yml
```diff
@@ -0,0 +1,73 @@
...
+wikimedia_commons: https://commons.wikimedia.org/wiki/File:$1
+contact:tripadvisor: https://www.tripadvisor.com/$1
+ref:at:gkz: http://www.statistik.at/blickgem/gemDetail.do?gemnr=$1
+mapillary: https://www.mapillary.com/map/im/$1
+ref:FR:SIRET: https://www.infogreffe.fr/societes/entreprise-societe/$1
+de:amtlicher_gemeindeschluessel: http://www.statistik-portal.de/Statistik-Portal/gemeindeverz.asp?G=$1
+twitter: https://twitter.com/$1
+contact:facebook: https://www.facebook.com/$1
+nist:fips_code: http://www.census.gov/quickfacts/table/PST045215/$1,00
+contact:google_plus: https://web.archive.org/web/20190401000000/https://plus.google.com/$1
```
mmd-osm, line 32: Google Plus is dead.

script/misc/update-tag2link
```diff
@@ -0,0 +1,60 @@
...
+    exit 1
+  end
+end
+
+filecontent = results.flatten({}) do |memo, result|
+  memo[result["OSM_key"]["value"].slice("Key:".size...)] = result["formatter_URL"]["value"]
+  memo
+end.to_yaml
+
+File.write(ARGV[0], filecontent)
```
mmd-osm, line 58: Can you sort the results by key before generating the file? This way it's easier to merge changes later on.

script/misc/update-tag2link
```diff
@@ -0,0 +1,60 @@
...
+
+#
+# Run with ./script/misc/update-tag2link config/tag2link.yml
+#
+
+require 'net/http'
+require 'json'
+require 'psych'
+
+QUERIES = {
```
mmd-osm, line 11: If possible, add location to original queries in JOSM source code, so it's easier to keep them in sync.

script/misc/update-tag2link
francois2metz: Done.

script/misc/update-tag2link
francois2metz: done

.dockerignore
```diff
@@ -0,0 +1 @@
+docker-db-data
```
mmd-osm, line 1: This commit c3108bd54d009357d5d7d8e76f9655be89ee54f2 doesn't really belong here...

francois2metz: I know. This PR is still in draft :pray:

config/tag2link.yml
francois2metz: All contact:* have been removed (mostly because they don't follow the rules of the wiki)

francois2metz:
The code is mostly finished and ready for review. I'll still working to find a way to stabilize the output of the `update-tag2link` script as some keys are duplicated on wikidata.

francois2metz:
I slightly edited the query to remove the third party formatters in cdd1a48decaa64cd6be9b0ae6900a1a49db642d1. This should be much better. I'll squash it once the review is done.

gravitystorm:
> The code is mostly finished and ready for review.

I've reviewed the code today, and my only suggestions are:

* I would prefer to see the update script as a rake task, and have the target file hardcoded, to make it easier to find and run in the future. 

But these are just a suggestions, not required for merging. I don't have any other comments.

francois2metz:
>     * I would prefer to see the update script as a rake task, and have the target file hardcoded, to make it easier to find and run in the future.

Good idea, I'm working on it.

francois2metz:
I update the PR to use a rake task `tag2link:update`

francois2metz:
Hi there. Is there is anything I can do to make it in an mergeable state? Or this is something you are no interested to follow up?

pyrog:
> I'm still wary that it will more or less eliminate any sort of review of what links are desirable in favour of just taking hordes of rules from some random web site. I realise in principle the changes would come here via PRs but what would be the real appetite for reviewing them and adding long term filtering of the results?

The review mechanism could be implemented in Data Items.
It already exist in Wikidata.

pyrog:
> Pretty sure this has already been suggested and discussed in another ticket...

At least in https://github.com/openstreetmap/openstreetmap-website/issues/2405#issuecomment-548957060

matkoniecz:
Note that sophox.org used in this script is down and continues to be down for some time.

> have pattern matching and replacement rules stored for each tag in Wikidata (on osm wiki)
> write a script to manually pull relevant tag information from Wikidata as SPARQL query (or whatever is suitable to fetch the data)
> Run script from time to time and check in updated config.yml file as part of this repository

Fortunately this info can be also stored directly on OSM Wiki.

See https://wiki.openstreetmap.org/wiki/Template:KeyDescription#Parameters and

`| url_pattern = http://openplaques.org/plaques/[value]`

example on that page.

Would you be interested in script using that method? Would be OK to do it using Python script (I implemented it recently for unrelated task) or is it mandatory to use Ruby?

config/tag2link.yml
```diff
@@ -0,0 +1,60 @@
+# THIS FILE IS AUTOGENERATED WITH THE script/misc/update-tag2link
+# PROGRAM DO NOT MANUALLY EDIT IT
+
+---
+3dmr: https://3dmr.eu/model/$1
+contact:facebook: https://www.facebook.com/$1
```
matkoniecz, line 6: Looking at https://taginfo.openstreetmap.org/keys/contact%3Afacebook#values - it seems to be incorrect

matkoniecz (changes requested):

mmd-osm:
Sophox should be back in the near future from what I read. Until then, let’s keep this one open. Adding a label „waiting for upstream“ would be good.

app/helpers/browse_tags_helper.rb
```diff
@@ -169,6 +171,17 @@ def telephone_links(_key, value)
...
   end
 
+  def tag2link(key, value)
+    return unless TAG2LINK.key? key
+
+    url = if value =~ %r{^https?://}
+            value
+          else
+            TAG2LINK[key].gsub("$1", value)
+          end
```
kmpoppe, line 181: In the not-so-unlikely-case that a mapper puts multiple values in there, wouldn't it be sensible to just display the value if it contains a semicolon?
Because a link to `facebook com / value1;value2` would almost certainly not yield an usable result either, would it?

tomhughes:
It's been pointed out on #5200 that this could be simplified by using the `tag2link` npm package and loading the JSON data it includes - we'd probably want to only link things with a rank of `preferred` if only to avoid having more than one link for the same key!

danieldegroot2:
Superseded by #6197.

PR #2637, opened 2020-06-03
Remove ineffective query cache disable initializer

gravitystorm:
I don't know why the query cache was originally intended to be disabled, since the [original commit](0d8e4b4337e9b94b6a9ed85095005a779f69583c) doesn't give any reason or links to tickets.

However, this initializer seems to be ineffective, since queries are being successfully cached on my local machine, and as far as I can tell, there's no middleware with that name any longer. Perhaps this is all just historical stuff?

So this PR just removes the initializer.

### DIFF
--- a/config/initializers/query_cache.rb
+++ b/config/initializers/query_cache.rb
@@ -1,1 +1,0 @@
-Rails.configuration.middleware.delete ActiveRecord::QueryCache


tomhughes:
I suspect it was either concern about memory with things like large map calls, or about stale results for API calls.

As you say it's no longer working because the query cache now works via executor hooks instead of installing itself as middleware.

PR #2641, opened 2020-06-03
Convert lib files to model concerns

gravitystorm:
This PR converts two files from the `lib` dir to be model concerns.


tomhughes:
I thought I'd done all the east ones of these...

I suspect I didn't take the time to figure out how to do the `included` thing in `Redactable` and I think I wanted to try and find a better name for `ConsistencyValidations` as I wasn't convinced that made grammatical sense as a concern name.

PR #2648, opened 2020-06-09
README.kor.md translates for Korean

suuubbb:
I create README.kor.md!

while I interpreting , I realized that translation put on the wiki site.
but I think it difficult to find README translation on wiki site.
so  I want to get a request for README.kor.md from master branch.
Is it okay to pull request this?


tomhughes:
Sorry but I really don't think we want to go down that route - we would wind up with a hundred translated README files cluttering up the root directory of the repository and most of them would be likely be horribly out of date as we would have no way to update them when the english version changes.

Issue #2654, opened 2020-06-11
Use a form builder for bootstrap forms


gravitystorm:
Our forms are quite varied, for various historical reasons. Some use the rails input helpers (e.g. `f.text_field`) and some are completely hand-crafted (with raw html, e.g. `<input type=`).

We also have a custom set of CSS styling rules for forms, some of which I dislike (e.g. making text inputs 50% wide, even on small screens) and some of which is confusing (e.g. our use of deemphasised buttons).

Finally, we have inconsistent behaviours between different forms, such as when and where to show error messages, location of cancel buttons (if any), marking dangerous actions with different button colours, and all these sort of things that people subliminally expect with modern websites.

So I'd like to move our forms to use Bootstrap styling, and gain all the nice things that come with it like hints, placeholders, buttons, error field highlights etc. But building those forms manually, even with rails helpers, involves a lot of work to get the errors and hints in the right places, and with the right CSS classes to get styled correctly. This is where [simple_form](https://github.com/heartcombo/simple_form) comes in. I've used it, and other form builders like [formtastic](https://github.com/justinfrench/formtastic), in other projects, and simple_form ties in best with bootstrap. It greatly simplifies the form building and increases the chance of ending up on the "happy path" with errors hints etc all working smoothly.

I'm opening this issue just to explain what I'm planning, since this is another one of these things that will take multiple PRs to implement. I have some WIP code to share shortly, but here's a preview of the trace upload form.
![Screen Shot 2020-06-11 at 12 12 59](https://user-images.githubusercontent.com/360803/84373578-0fd59880-abdd-11ea-8b2e-d2ca39c7d013.png)

gravitystorm:
So we have two options - #2655 (using the `simple_form` gem) and #2667 (using the `bootstrap_form` gem). Overall there's not much to choose between them, but as I said in #2667 I lean slightly towards the bootstrap_form option.

Today I've been tackling integrating the richtext input (that we use in diary entries and elsewhere). This would be nice to have as a custom input, so that we could do `f.richtext_field :body` and get the label, textarea, help, preview buttons etc automatically. After a bit of trial and error I [found a way](https://github.com/gravitystorm/openstreetmap-website/commit/aeaf0404e7d805871b7fe3c4252adec92223d2b9) that works reasonably well.

[...]

PR #2655, opened 2020-06-11
WIP: simple_form

gravitystorm:
Refs #2654 

This is a draft implementation of using `simple_form` and Bootstrap. I might split this PR into stages, e.g. a separate PR for refactoring the existing CSS, and I haven't finished noodling around to find any remaining forms that I broke.

In the meantime, you can see how simple_form work for the trace upload form:

```
<%= simple_form_for @trace, :url => { :action => "create" }, :html => { :multipart => true } do |f| %>
  <%= f.error_notification %>
  <%= f.error_notification message: f.object.errors[:base].to_sentence if f.object.errors[:base].present? %>
  <%= f.input :gpx_file, :as => :file, required: true %>
  <%= f.input :description %>
  <%= f.input :tagstring %>
  <%= f.input :visibility, :collection => [:private, :public, :trackable, :identifiable],
                           :include_blank => false,
                           :hint => link_to(t(".visibility_help"), t(".visibility_help_url")) %>
  <%= f.button :submit %>
  <%= link_to t(".help"), t(".help_url"), :class => "btn btn-link" %>
<% end %>
```

So for example, `f.input` takes care of checking the i18n for hints, checks the model for errors, looks up labels, and adds all those elements automatically.


lib/templates/erb/scaffold/_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<%# frozen_string_literal: true %>
```
tomhughes, line 1: Is this actually used by anything?

config/locales/en.yml
```diff
@@ -116,6 +117,20 @@ en:
...
+      trace:
+        gpx_file: Upload GPX File
+        visibility: Visibility
+    options:
+      trace:
+        visibility:
+          private: Private (only shared as anonymous, unordered points)
+          public: Public (shown in trace list and as anonymous, unordered points)
+          trackable: Trackable (only shared as anonymous, ordered points with timestamps)
+          identifiable: Identifiable (shown in trace list and as identifiable, ordered points with timestamps)
```
tomhughes, line 133: With these descriptions moved here do we still need the old ones?

app/views/traces/edit.html.erb
```diff
@@ -4,52 +4,15 @@
...
-      <label class='standard-label'><%= t ".visibility" %></label>
-      <%= f.select :visibility, [[t("traces.visibility.private"), "private"], [t("traces.visibility.public"), "public"], [t("traces.visibility.trackable"), "trackable"], [t("traces.visibility.identifiable"), "identifiable"]] %> (<a href="<%= t ".visibility_help_url" %>"><%= t ".visibility_help" %></a>)
-    </div>
-  </fieldset>
-
-</div>
-
-<%= f.submit %>
-
+<%= simple_form_for @trace do |f| %>
```
tomhughes, line 7: This seems to have lost all the extra information that was displayed before the editable fields - is that a deliberate decision?

lib/templates/erb/scaffold/_form.html.erb
gravitystorm: I'm not sure yet - I think it might be used by `rails scaffold`, but I haven't checked. I lean towards removing it.

config/locales/en.yml
gravitystorm: Possibly - I think the existing translations are used for the `:show` action. That could be refactored to use the hints.

app/views/traces/edit.html.erb
gravitystorm: Yeah, no final decision on that. I'm not sure why e.g. number of points was shown on the edit form, it's not like you could change any of that metadata so why have it shown disabled on the form. Feedback welcome on this though.

I started playing around with the simple_form `disabled` option and the bootstrap `.form-control-plaintext` class but haven't finished that line of enquiry.

gravitystorm:
> The initializers, especially the bootstrap one, are a little disturbing ;-)

Yeah, I was surprised when I first used it and came across it being an initializer instead of hidden away within a gem. But I suspect it's a sort of pressure-relief valve for the maintainers - i.e. push much of the html generation code into the end application, in order to avoid getting swamped with everyone needing more and more configuration options.

tomhughes:
> > The initializers, especially the bootstrap one, are a little disturbing ;-)
> 
> Yeah, I was surprised when I first used it and came across it being an initializer instead of hidden away within a gem. But I suspect it's a sort of pressure-relief valve for the maintainers - i.e. push much of the html generation code into the end application, in order to avoid getting swamped with everyone needing more and more configuration options.

My main concern is maintenance of it as the gem and/or bootstrap develop over time...

gravitystorm:
> My main concern is maintenance of it as the gem and/or bootstrap develop over time...

Indeed, same here.

tordans:
Hey! We started migrating away from simpleform at @betterplace_org. We are now using https://github.com/bootstrap-ruby/bootstrap_form in some cases. The short story is, that bootstrap in simple form was not well maintained but introduced very specific code that is hard to improve (the bootstrap initializer mainly) and hard(er) to migrate away from in case it is needed. 
I can write more and add some links if you want.

gravitystorm:
> I can write more and add some links if you want.

Yes please! Your advice would be very valuable. I'm particularly interested in your comment about simple_form-bootstrap not being well maintained, along with any thoughts on using a bootstrap-specific form builder vs a form builder that fully supports multiple outputs.

tordans:
Hi Andy, I hope this helps and is useful.

I don't think simple_form is a bad option; I just don't think it's the best option out there anymore. There was a time, when they where the only good option to quickly build nice forms  in rails (and they still hold that image/brand), but today there are more specialized options.

> I'm particularly interested in your comment about simple_form-bootstrap not being well maintained, …

This opinion is probably tainted by my experience with simple_form + bootstrap4. But I think most of it is still true today.

Back then, there was no initializer for the new bootstrap and it was not possible to even create one without changing the simple_form core to work around requirements that bootstrap had ([Example](https://github.com/heartcombo/simple_form/pull/1552)) and kind-of-bugs in simple_form ([Example](https://github.com/heartcombo/simple_form/pull/1369#issuecomment-312208316)). However, IMO the process to get those changes going took too long for a project that I would like to base my forms upon for the years to come.

When we tried to contribute to simple_form, understanding and creating a configuration for the initializer did not feel great. It became very complex very quickly. So even the "I will do it myself"-option did not work out in a sane timeframe.

I started a initializer for the bootstrap4-alpha at https://github.com/heartcombo/simple_form/pull/1337 (Dec 2015) and there are a few re-tries over the years to get this general issue forward in this thread. 

One think I noticed is, that there wasn't a strong community of community-contributions to collaborate on this issue even though a lot of people had the same problem. For a while all that happened was people pasting their WIP-initializers in issue-comments. 

Also, it was quite hard to get input or support from the maintainers; thinks they wrote in 2017 suggested to me that they see the project in maintanance mode (https://github.com/heartcombo/simple_form/pull/1337#issuecomment-350709787) and that the initilizer setup is not what they would build today (https://github.com/heartcombo/simple_form/pull/1337#issuecomment-350889176). At that time the simple_form repo was hosted at plataformatec/simple_form and [they seem to to Elixir for a living now](https://plataformatec.com/en/) which suggested to me, that they moved on. However, I see now that the ownership of the repo has changed to heartcombo/simple_form, so maybe there is new vigor on the project now.
However, the public communication still seems off to me. The repo-Link in the header of https://github.com/heartcombo/simple_form still [links to a Blog with the last entry about bootstrap from 2014](http://blog.plataformatec.com.br/tag/simple_form/). And the readme still just say "bootstrap" without referencing the actual version.

> …along with any thoughts on using a bootstrap-specific form builder vs a form builder that fully supports multiple outputs.

I really like the way simple_form allows to write form – especially the hint-Attribute!

However, the form builder is specific to simple_form, which will result in a situation like I referenced above: If you want to update bootstrap, you need a new initializer which makes simple_form a blocking dependency. And updating the initializer is hard.

In contrast – and this is where my technical understanding becomes even muddier – the form build syntax of https://github.com/bootstrap-ruby/bootstrap_form is the same as the default-rails syntax https://guides.rubyonrails.org/v5.2/form_helpers.html; it's just the `form_for` that is replaced and thereby changes the output of the rails form builder.

In case an update to a new bootstrap is needed, one still needs to understand [the gigantic form builder code](https://github.com/bootstrap-ruby/bootstrap_form/blob/master/lib/bootstrap_form/form_builder.rb), but at least there is no additional complexity due to the abstraction of an initializer and framework. And the files only need to work for the new bootstrap (and not also Zurb Foundation …).

gravitystorm:
> Hi Andy, I hope this helps and is useful.

It's super useful. I think I'll try out bootstrap_form and based on your feedback I lean towards using it instead of simple_form. I'll have a better conclusion after trying it out though.

The big unknown is how either project will respond to the upcoming bootstrap 5 release, but there's no urgency on that topic from us, so long as it is supported eventually.

gravitystorm:
Closing in favour of bootstrap_form, see #2667

PR #2661, opened 2020-06-21
Change tile licence (pls merge 2020.7.1)

simonpoole:
This changes the text describing the tile licence as per decision of the OSMF board starting July 1st 2020. 

There might be a bit of fine tuning before this is finalised, in any case it should not go live before July 1st.


simonpoole:
An instance with the changed text is available on upc.poole.ch:3000/copyright

PR #2667, opened 2020-06-24
Use bootstrap_form for formbuilding

gravitystorm:
Based on feedback from @tordans, here is an alternative to #2655 that uses the [bootstrap_form](https://github.com/bootstrap-ruby/bootstrap_form) gem instead of simple_form.

It covers the same ground as the earlier PR, namely converting the trace forms, so it's easy to compare the two options. Beyond the general advice given by @tordans wrt to maintainability of the two gems, my feelings on `bootstrap_form` are:

Pros:
* Stays closer to the standard rails forms approach. Doesn't invent its own grammar e.g. `f.text_field` (the rails way) instead of `f.input` (the simple_form / formtastic approach).
* Much easier to use the different form types, like horizontal forms, if we want to.

Cons:
* No automatic detection of form fields types. You need to specify e.g. `f.text_field :description` instead of `f.input :description`. So given this topic is both a pro and a con, take your pick.
* Unsophisticated i18n for `select` field options. `simple_form` allows you to pull select option labels straight from the i18n system, e.g. `f.input :visibility, :collection => [:private, :public, :trackable, :identifiable]` will look up `en.simple_form.options.trace.visibility.private` etc for the labels. `bootstrap_form`, like rails, leaves you to do this by hand on each form afaict e.g. `f.select :visibility, [[t("traces.visibility.private"), "private"], ...]` which is clunky.

Headaches:
* Edge case with file upload fields - you have to go full hog and use the bootstrap custom file widget, including javascript. If you use `f.file_field_without_bootstrap`, then you get the standard browser button (like we have now), but you have to do everything by else hand, such as label tags and error displays. `simple_form` was half-way in between. But since I've already figured out how to get all the javascript working (eventually) then I'm happy to go with it.

Overall, I favour this PR over #2655, but I'm interested to see what everyone else thinks.


tomhughes:
I was concerned about the automatic type detection anyway, at least where strings are concerned, specifically with the difference between text areas verses ordinary text fields.

I think I'm right in saying it was relying on field lengths to decide, but postgres doesn't really need lengths on string fields - they are just syntactic sugar and I suspect many of our newer ones will just be generic text fields although I might be wrong.

package.json
```diff
@@ -1,7 +1,9 @@
 {
   "name": "openstreetmap",
   "private": true,
-  "dependencies": {},
+  "dependencies": {
+    "bs-custom-file-input": "^1.3.4"
+  },
```
tomhughes, line 6: We haven't been using yarn to manage javascript for the actual web site to date, although I have experimented with it. I was slightly surprised it worked but I see `node_modules` has crept on to the include path at some point, and the other infrastructure change I made for it has also been made for other reasons.

I guess there's no reason not to do it, other than trying to keep everything in one place - although is other stuff we could manage via yarn there are some things where we can't at the moment so would have to stick with vendorer.

It does mean that we will need to run yarn as part of the deployment though and to date we don't have any tooling for that.

gravitystorm:
I've fixed all the linter checks, and I think we're good to go here. If this is merged, I intend to convert all of our form handling in subsequent PRs.

tomhughes:
Merging is probably going to be delayed by the need to sort out the node module deployment for the production servers - if possible I'd like to wait until we get them switched to Ubuntu 20.04 as that will mean yarn is available from the repos.

gravitystorm:
> Merging is probably going to be delayed by the need to sort out the node module deployment for the production servers - if possible I'd like to wait until we get them switched to Ubuntu 20.04 as that will mean yarn is available from the repos.

No worries, no rush from my side :-)

PR #2670, opened 2020-06-26
Add OPNVKarte

pnorman:
The layer has been approved by the OWG, I just took longer than expected to add it.

image

There was some discussion with @gravitystorm about naming in IRC a few weeks back. I think this naming is consistent with what we've done with other layers, using the layer name rather than a description. We've used that for MapQuest, HOT, Transport, and close to that to OpenCycleMap.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2498,6 +2498,7 @@ en:
         cycle_map: Cycle Map
         transport_map: Transport Map
         hot: Humanitarian
+        opnvkarte: ÖPNVKarte
       layers:
         header: Map Layers
         notes: Map Notes
@@ -2509,6 +2510,7 @@ en:
       donate_link_text: "<a class='donate-attr' href='%{donate_url}'>Make a Donation</a>"
       terms: "<a href='%{terms_url}' target='_blank'>Website and API terms</a>"
       thunderforest: "Tiles courtesy of <a href='%{thunderforest_url}' target='_blank'>Andy Allan</a>"
+      opnvkarte: "Tiles courtesy of <a href='%{memomaps_url}' target='_blank'>MeMoMaps</a>"
       hotosm: "Tiles style by <a href='%{hotosm_url}' target='_blank'>Humanitarian OpenStreetMap Team</a> hosted by <a href='%{osmfrance_url}' target='_blank'>OpenStreetMap France</a>"
     site:
       edit_tooltip: Edit the map
--- a/vendor/assets/leaflet/leaflet.osm.js
+++ b/vendor/assets/leaflet/leaflet.osm.js
@@ -35,6 +35,14 @@ L.OSM.TransportMap = L.OSM.TileLayer.extend({
   }
 });
 
+L.OSM.OPNVKarte = L.OSM.TileLayer.extend({
+  options: {
+    url: 'https://tileserver.memomaps.de/tilegen/{z}/{x}/{y}.png',
+    maxZoom: 18,
+    attribution: '© <a href="https://www.openstreetmap.org/copyright" target="_blank">OpenStreetMap</a> contributors. Tiles courtesy of <a href="http://memomaps.de/" target="_blank">MeMoMaps</a>'
+  }
+});
+
 L.OSM.HOT = L.OSM.TileLayer.extend({
   options: {
     url: 'https://tile-{s}.openstreetmap.fr/hot/{z}/{x}/{y}.png',
--- a/app/assets/javascripts/embed.js.erb
+++ b/app/assets/javascripts/embed.js.erb
@@ -35,6 +35,8 @@ window.onload = function () {
     new L.OSM.CycleMap(thunderforestOptions).addTo(map);
   } else if (args.layer === "transportmap") {
     new L.OSM.TransportMap(thunderforestOptions).addTo(map);
+  } else if (args.layer === "opnvkarte") {
+    new L.OSM.OPNVKarte().addTo(map);
   } else if (args.layer === "hot") {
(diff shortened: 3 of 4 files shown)


config/locales/en.yml
```diff
@@ -2498,6 +2498,7 @@ en:
         cycle_map: Cycle Map
         transport_map: Transport Map
         hot: Humanitarian
+        opnvkarte: ÖPNVKarte
```
tomhughes, line 2501: Shouldn't this be ÖVPNKarte?

pnorman: Isn't that what it is? I don't see a typo

tomhughes: Oh you're quite right, it's just very hard to see in this font, at least without my glasses...

gravitystorm:
> . I think this naming is consistent with what we've done with other layers, using the layer name rather than a description. We've used that for MapQuest, HOT, Transport, and close to that to OpenCycleMap.

This is the exact opposite of what I was trying to say! I think the user interface should show what is useful to the user, i.e. what the topic of the map layer is, not whatever the codename of the project is or which organisation is behind it.

For example, the "Standard" layer shows a roughly standard map - it doesn't have a narrower topic than that. We don't call this layer "openstreetmap-carto" since that would be meaningless to most people. We (thankfully) no longer call it Mapnik. And if the project that implements this is changed, again, perhaps to openstreetmap-cartographic, then I would expect it still to be called "Standard" in the user interface.

The "Humanitarian" layer is another one - it shows a map with the topic of "Humanitarian", but not "HDM-CartoCSS" (the project name) nor HOT (the organisation behind it).

The "Transport Map" layer shows a Transport themed map. The project is called Thunderforest Transport, by Thunderforest, but we don't use either of those in the layer picker.

The "Cycle Map" layer is named like it is for historical reasons, and could be better named something like Cycling. But again, it's neither OpenCycleMap nor mentioning Thunderforest.

In summary, my view of what we should be doing showing the topic, not the implementation details, of each layer.

| Topic (shown to user) | Project Name (not shown)     | Hosting Organisation| Notes                         |
|-----------------------|-------------------------|----------------------------------|-------------------------------|
| Standard              | openstreetmap-carto     | OSMF                             |                               |
| Cycle Map        | OpenCycleMap            | Thunderforest                    | Topic could be renamed to "Cycling" |
| Transport Map             | Thunderforest Transport | Thunderforest                    | Topic could be renamed to "Transport"  |
| Humanitarian          | HDM-CartoCSS            | HOT                              |                               |

(For the avoidance of doubt, as well as being a maintainer here, I'm also the creator of two of the existing styles. I'm wearing my OpenStreetMap hat here, not my Thunderforest hat, and I'm not asking for the Thunderforest brand name, or the project names of either of these two styles, to be shown in the user interface - quite the opposite.)

As for the MapQuest layer, or the tiles@home layer, or the Mapnik label - yes, those existed, and I think they were poor choices of label. I think we should learn from our previous mistakes, and continue to use topics as the label shown to the user.

tomhughes:
Oh I completely misunderstood what you meant on IRC then as I thought you were meaning the the current cycle later should use the OpenCycleMap name, and was in fact planning to rename it as such.

The problem we have going forward is that we're going to have two cycling layers and two public transport layers, so we will need to have some way to distinguish them!

vendor/assets/leaflet/leaflet.osm.js
```diff
@@ -35,6 +35,14 @@ L.OSM.TransportMap = L.OSM.TileLayer.extend({
   }
 });
 
+L.OSM.OPNVKarte = L.OSM.TileLayer.extend({
```
HolgerJeromin, line 38: This file cames from upstream, doesn't it?

tomhughes: It does but we define our own instead I think because we change the attribution relative to upstream?

simonpoole:
May I point out that we already have precedent on how to handle this, namely the routing provider selection.

So I would suggest a scheme roughly like

_Topic_ - _Provider_

or

_Topic_ (_Provider_)

it may be appropriate to exchange "Provider" for a layer name if the provider prefers it that way, I son't think we really care about that aspect.

So we could have

**Standard**
**Cycling (Thunderforest)**
**Cycling (xyz)**
**Transport (Thunderforest)**
**Transport (ÖPNVKarte)**
**Miscellaneous (Humanitarian)**

(I would argue that the label Humanitarian is demeaning towards the providers and creators of the other layers and shouldn't be used but that is a different topic)

vendor/assets/leaflet/leaflet.osm.js
pnorman: Yes, this is an upstream vendored file. The changes are already made upstream, this was a `vendorer update` where I just updated leaflet-osm.

pnorman:
I don't really have a strong opinion on this, so if the maintainers will let me know which is preferred I'll change it.

tomhughes:
This is what the layer switcher would look like with the names @simonpoole suggested:

Screenshot_2020-06-28 OpenStreetMap

So we will definitely need to change the font size or something if we go for such long names.

simonpoole:
> 
> This is what the layer switcher would look like with the names @simonpoole suggested:
...
> 
> So we will definitely need to change the font size or something if we go for such long names.
...
Obviously there is potential for endless bike sheding. For example we could have the categories as headings of some kind (and not repeat them on the button), or a myrid of other ways of doing this. That IMHO somebody with some design chops should have a look at.

pnorman:
Could we split up the discussion on what to call this layer in this PR and any long-term renaming?

PR #2695, opened 2020-07-08
Turn on turn restriction support for graphhopper

simonpoole:
No way of actually testing if this works, but it is based on https://docs.graphhopper.com/#operation/getRoute/parameters/point and https://www.graphhopper.com/blog/2020/07/08/turn-restriction-support-for-graphhoppers-directions-api/

### DIFF
--- a/app/assets/javascripts/index/directions/graphhopper.js
+++ b/app/assets/javascripts/index/directions/graphhopper.js
@@ -29,6 +29,7 @@ function GraphHopperEngine(id, vehicleType) {
           "ch.disable": vehicleType === "car",
           "elevation": false,
           "instructions": true,
+          "turn_costs": true,
           "point": points.map(function (p) { return p.lat + "," + p.lng; })
         },
         traditional: true,


tomhughes:
Doesn't work I'm afraid - all routes with a 400 response and "Disabling CH not allowed on the server-side" as the message.

simonpoole:
The blog post would kind of imply that this should not disable CH support, maybe somebody from graphhopper could way in. From a marketing pov it would seem to be worth it as this would get rid of the rather frequent complaints from mappers that the adding of turn restrictions doesn't get reflected in the routing instructions when using graphhopper.

easbar:
Yes `turn_costs=true` does not disable CH, but it seems the problem is that requests with `disable.ch=true` are redirected to another server of the GraphHopper infrastructure which does not support turn costs. Why this line in `graphhopper.js`:

```javascript
"ch.disable": vehicleType === "car"
```
? @karussell maybe?

karussell:
There seems to be a bug or misconfiguration on our side now as `ch.disable=true` should already have turn costs support for car. Will investigate it.

> Doesn't work I'm afraid - all routes fail with a 400 response and "Disabling CH not allowed on the server-side" as the message.

We will try to make this working in the next weeks, but here it should not be necessary. We can either keep using `ch.disable=true` which already supports turn costs (except for the above bug) or we use `turn_costs=true`. Please note that for `turn_costs=true` the update cycle is currently much slower as indicated in the blog post. E.g. instead of 1-2 days it can take 3-7 days until a change in OSM data is considered in the routing.

easbar:
The reason we use `"ch.disable": vehicleType === "car"` is that this way we enforced another/slower algorithm that already supported turn restrictions. However this is broken at the moment because of a GH configuration issue (will be fixed soon) and on openstreetmap.org we get GraphHopper routing *without* turn restrictions support nonetheless.

There are two options:

1) `turn_costs=true&ch.disable=true`: This does not work at the moment but will be fixed soon. It will result in routing that does support turn restrictions but is a bit slower.

2) `turn_costs=true` (remove the `ch.disable` parameter): This already works and results in faster routing than the first option. The only drawback is the longer data update cycle @karussell mentioned.

To avoid confusion in the blog article maybe I should have written that turn costs were already supported before but only for `/route` (not `/matrix`) and only using a slower algorithm.

karussell:
> turn_costs=true&ch.disable=true: This does not work at the moment but will be fixed soon. It will result in routing that does support turn restrictions but is a bit slower than the second option

This also works now and the slower query speed might be less important than the much faster map update frequency.

BTW: if we do not use ch.disable=true we could also support [fast alternative routes](https://graphhopper.com/maps/?point=n%C3%BCrnberg&point=berlin&algorithm=alternative_route).

karussell:
@tomhughes thanks for merging this! We see now several requests that time out. This is one drawback that I forgot to mention: for long routes the calculation might take too long and as we have a limitation they then time out, which gives an ugly experience. I now leaning towards removing `ch.disable=true` and accepting the longer update time but giving a better user experience. Let me know if you agree and I can create a pull request.

tomhughes:
I've dropped the `ch.disable` now...

karussell:
Thanks!

PR #2696, opened 2020-07-08
Use bootstrap text-muted in place of custom deemphasze class

gravitystorm:
This avoids changing anything related to forms, since they will be switched over to bootstrap separately.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -101,6 +101,10 @@ a {
   }
 }
 
+.text-muted a {
+  color: $blue;
+}
+
 /* Rules for the header */
 
 #menu-icon {
@@ -1417,7 +1421,7 @@ tr.turn:hover {
     margin-top: $lineheight/2;
   }
 
-  &.deemphasize {
+  &.deleted {
     background-color: #fee;
   }
 
@@ -1472,7 +1476,7 @@ tr.turn:hover {
       padding-top: $lineheight;
       border-top: 1px solid $grey;
     }
-    &.deemphasize {
+    &.deleted {
       background-color: #fee;
     }
     p {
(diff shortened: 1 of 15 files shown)


app/assets/stylesheets/common.scss
```diff
@@ -101,6 +101,10 @@ a {
   }
 }
 
+.text-muted a {
```
tomhughes, line 104: So `text-muted` is a bootstrap thing but we still need a rule for it? Is that because bootstrap just fades it but we want to set the colour or something?

Also are the `deemphasize` rules above this still needed?

app/assets/stylesheets/common.scss
```diff
@@ -1417,7 +1421,7 @@ tr.turn:hover {
     margin-top: $lineheight/2;
   }
 
-  &.deemphasize {
+  &.deleted {
```
tomhughes, line 1424: What's the logic behind keeping this but changing the class it matches?

app/assets/stylesheets/common.scss
gravitystorm: In bootstrap `text-muted` there is no highlight colour for links, just an underline on hover. We use muted text with multiple links in various places, so I decided to keep the link colouring for now. 

The deemphasize rules are still needed for standard-form help text, but will be removed after forms are refactored.

app/assets/stylesheets/common.scss
gravitystorm: I think the class name is better as "deleted", since that's what has happened to the entry/comment. The associated css (setting a red background) isn't de-emphasizing anything, so it's a bit misleading to keep the original name.

tomhughes: Ah I missed that you had added that class in to some things.

PR #2704, opened 2020-07-11
Make layer naming more consistent and indicative

mikini:
* Use "Transport" for both layers showing public transport cartographies
* Indicate cartography name of transport layers in parenthesis to distinguish
* Also eliminate "map" in layer names, which seems redundant

This is a follow-up to a comment regarding naming on blog post
https://blog.openstreetmap.org/2020/07/10/opnvkarte-a-new-featured-layer-on-www-openstreetmap-org/
(still unmoderated as of this commit).

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2499,10 +2499,10 @@ en:
           other: You are within %{count} feet of this point
       base:
         standard: Standard
-        cycle_map: Cycle Map
-        transport_map: Transport Map
+        cycle_map: Cycle
+        transport_map: Transport (TF)
         hot: Humanitarian
-        opnvkarte: ÖPNVKarte
+        opnvkarte: Transport (ÖPNV)
       layers:
         header: Map Layers
         notes: Map Notes


gravitystorm:
See also https://github.com/openstreetmap/openstreetmap-website/pull/2670#issuecomment-650037548 where I describe what I think the naming conventions should be based on.

I should note here that "TF" is not a good contraction of Thunderforest, since Thunderforest is just one word and the f is never capitalised. 

Also, ÖPNV is neither the name of the map style (ÖPNVKarte), nor of the provider (MeMoMaps), so I think it's worth changing that to something else too.

mmd-osm:
So what is the final conclusion here? Adjust the pull request or maybe just close it?

mikini:
I have rebased the branch and added a new commit reflecting the discussions here and in #2670 as pr. my understanding of consensus. Space for longer texts were added by simply decreasing the font size of the layer name label.

The selector now looks like below (screenshot is of browser manipulated production site, did not succeed yet in getting dev environment running to test the code, so might have missed something);

Screenshot from 2023-01-13 01-58-41

mmd-osm:
Some recent changes to the layer list: ÖPNVKarte was removed and Tracestrack Topo added:

image

PR #2705, opened 2020-07-11
Link to English site for Memomaps (ÖPNV) attribution

mikini:
This is a follow-up to a comment regarding this on blog post
https://blog.openstreetmap.org/2020/07/10/opnvkarte-a-new-featured-layer-on-www-openstreetmap-org/
(still unmoderated as of this commit).


gravitystorm:
Thanks for the PR!

We can only link to one page for attribution, so we have the same link for all of our visitors. Therefore we need to think carefully before linking directly to a language-specific page.

Also, when I visit https://memomaps.de/ I get automatically redirected to http://memomaps.de/en/homepage/ - I've checked and that is based on browser language headers and is saved into a cookie. So I think it's fine to keep the link to the homepage, since memomaps is already doing language detection anyway.

PR #2714, opened 2020-07-17
Update embed.js.erb to use CDN test URLs

iandees:
This switches the embed.html page to use a temporary Fastly CDN URL to test configuration. This is related to ongoing work with the Operations Working Group to alleviate some of the load on our tile servers.


iandees:
If it's preferred, I can list the Fastly domains explicitly instead of using a wildcard.

config/initializers/secure_headers.rb
```diff
@@ -7,7 +7,7 @@
   :form_action => %w['self'],
   :frame_ancestors => %w['self'],
   :frame_src => %w['self'],
-  :img_src => %w['self' data: www.gravatar.com *.wp.com tile.openstreetmap.org *.tile.openstreetmap.org *.tile.thunderforest.com tileserver.memomaps.de *.openstreetmap.fr],
+  :img_src => %w['self' data: www.gravatar.com *.wp.com tile.openstreetmap.org *.tile.openstreetmap.org *.tile.thunderforest.com tileserver.memomaps.de *.openstreetmap.fr *.global.ssl.fastly.net],
```
tomhughes, line 10: I'd make this an override in the `embed` method of the export controller rather than enabling it globally.

Firefishy: @tomhughes Any examples of this? What is required?

tomhughes: Well there is already an override there, it just needs an `img_src` rule adding. You can just add that one source and it will be merged with the global list.

Firefishy:
Quick background for others not in the OSM Operations communications feedback loop: This is to test moving a small proportion of the `tile.openstreetmap.org` traffic to a 3rd party commercial CDN provider. We have been discussing it over the last few fortnightly OSM Ops meetings.

The point of the test is to evaluate:

1) How well the CDN performs?
2) What config we need to consider / set / create?
3) Gaining some metrics to help further decisions / evaluation?
4) Other observations?

config/initializers/secure_headers.rb
iandees: I think I properly added the override rule.

Firefishy:
@tomhughes Look ok to you?

tomhughes:
Oh I think it's fine technically - it's just an operational decision now about when we're ready to deploy it.

Firefishy:
I mentioned to @iandees: If it uses more traffic than expected or has any negative effects it would likely be best to revert the change. With that in mind, might be best to do a squish merge.

iandees:
For what it's worth, I'm OK with the change and am ready to watch the Fastly stats to make sure things are working. I can also prep a revert PR for us to merge in quickly if it would help.

tomhughes:
The fastly config is not currently acceptable and I am unable to fix it - absent some assistance with figuring out how to remove the "default condition" that it is adding to the VCL we are not able to proceed.

tomhughes:
To explain further, this is the bit of VCL that chooses the upstream:

```
# Snippet Choose Backend : 100
if ( client.geo.continent_code == "OC" ) {
  set req.backend = australia;
} else if ( client.geo.continent_code == "NA" ||
            client.geo.continent_code == "SA" ) {
  set req.backend = america;
} else {
  set req.backend = europe;
}

  # default conditions
  set req.backend = F_Odin___Amsterdam;
    # end default conditions
```

The first bit is my custom VCL which replaces your previous condition based attempt with a version that does better fallback.

But that that choice is overwritten by some fastly generated VCL that I can't work out how to get rid of.

iandees:
Gotcha. I submitted a support ticket asking for help there. In the meantime I'm trying to figure out how to upload custom VCL.

tomhughes:
If you're talking to support it might be worth asking about IPv6 and HTTP2 as well - I believe we have to ask support for them to be enabled.

iandees:
I was able to upload a custom VCL that inserts the backend selection snippet you added after the "default condition". It doesn't seem like the best way to do it, though.

I'll ask about IPv6 and HTTP2.

iandees:
Support gave a slightly better answer for overriding the default backend behavior. They said to add a "false" condition to all the backends. That seems really hacky to me, but it works and allows easier interaction with the VCL through the web console instead of writing custom VCL.

tomhughes:
This is not reassuring me that fastly is a sane solution...

iandees:
Yep, I agree that their friendly interface for modifying VCL leaves a bit to be desired and should be something we think about when making decisions about the full load of tile.osm.org. I think we'll probably want to store the VCL separately in version control and use their API to modify the configuration it rather than using their website.

For now, let's give it a try on this smaller amount of traffic and see what other weirdness we discover.

tomhughes:
Whilst I would normally agree about keeping it in version control I'm not sure I really want us to take responsibility for all that boilerplate and maintain a completely custom file.

As I understand it you can't anyway - if you go full custom then you still have to include a bunch of their boilerplate, which means you need some way of keeping that in sync with changes they might make, which sounds like a nightmare.

I've added the conditions now anyway...

tomhughes:
By the way another limitation is that you can apparently only define five upstreams through the web interface.

I'm assuming you could do more if you did it with custom VCL fragments, though I haven't tried yet, but they explicitly recommend using the web interface for that.

Firefishy:
@iandees You ok this going live? Do we have enough credits available?

iandees:
Yep, every account gets $50/mo in free credits and I have $47 left for the rest of July. Let's give it a shot.

tomhughes:
How many days^Whours do we expect $50 to last ;-) Just wondering not just about tomorrow but about what happens at midnight tomorrow when we start a new month...

tomhughes:
No, but I can tell you there were about 1.5 million loads of embed.html yesterday - don't know what the average number of tiles each one uses is though.

Firefishy:
Wow, 1.5 million loads is a lot more than I initially expected. I'd expect each page load to maybe load a guesstimate of 40x tiles at around 9KB each.

~700 tiles per second.
~6000 KBytes/s.
~3% of the total tile.openstreetmap.org traffic.

iandees:
40 tiles per embed.html load seems high to me, but based on my estimates that would be around $100/day. I'll reach out to Fastly to see if we can get enough credits to try at least this for at least a week.

pnorman:
I checked logs from Squid in late feb and got 4.8M requests per day, 0.8% of total.

iandees:
@pnorman that's 4.8M squid requests per day with embed.html referrer? That's 56 requests per second, so I'm confused where you got the 260M requests per day from.

pnorman:
> @pnorman that's 4.8M squid requests per day with embed.html referrer? That's 56 requests per second, so I'm confused where you got the 260M requests per day from.

56 TPS on squid, not on the CDN. We have a two-level cache system.

260M is 350000 * 0.0086 * 3600 * 24

Firefishy:
We have a 3 tier cache system:
1. nginx caching (edge)
2. squid caching
3. mod_tile / renderd caching (render backend)

Currently we only log 2. , but plan is to move logging to 1.

gravitystorm:
My only comment here is that I would be very cautious about using a non-openstreetmap.org dns name in production. It's likely that some people will copy it and use it directly, and the longer the trial (i.e. the more successful the trial) the more likely that is to happen.

If it's possible to use a CNAME alias for a domain that OSMF owns (e.g. tile-dev-cdn.openstreetmap.org or similar) then that would be an option. But perhaps that has been considered already in private discussions.

iandees:
Yep, agree on the domain. In order to use a custom domain we'd have to upgrade to a paid monthly account and set up SSL with the domain on Fastly. I didn't push for that because it'd increase the time to get started and cost to run the test. We wouldn't be able to use a free account and test quickly. My hope was to find a slice of traffic that would fit in the existing $50/mo free credits but we haven't found that yet 😄.

I need to ask the Fastly folks about more credits to test this, so I will also ask about SSL.

mmd-osm:
I have two questions on cache validity and expiry mechanisms on fastly:
*  How long will fastly keep cached tiles around before they expire, and new tiles will be fetched from other caches or the rendering server?
* What are options to selectively purge cache contents on fastly in case of vandalism?

iandees:
> How long will fastly keep cached tiles around before they expire, and new tiles will be fetched from other caches or the rendering server?

Fastly [respects the Cache-Control headers](https://docs.fastly.com/en/guides/configuring-caching) that the backend sends. For example [this tile](https://tile-openstreetmap-org.global.ssl.fastly.net/5/6/0.png) specifies `Cache-Control: max-age=111272` (plus some other bits), so Fastly will cache it up to 111,272 seconds (~1d6h). That header came from the backend, so we can control that behavior. It's the same header that Squid pays attention to, so I'm sure it has been tweaked already.

We can control the Fastly cache TTL separately from the browser TTL using the shared max-age part of the Cache-Control header.

More info on MDN's [`Cache-Control` page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control).

> What are options to selectively purge cache contents on fastly in case of vandalism?

Fastly has a [cache purge API](https://developer.fastly.com/reference/api/purging/) that we can use through the web interface or via an API once the backend is ready to re-render a corrected tile.

iandees:
An update here: the formalities to get Fastly credits to support this and future tests/go-live are done. They're not showing on the account yet, so I'm waiting for confirmation that they're applied and I'll post here when we're ok to merge from that standpoint.

iandees:
I had some back and forth with Fastly earlier this week and we should be good to go with credits. Their support person helped me get SSL set up, too. The couple DNS pull requests linked in above are related to that.

I modified this pull request to use the new `tile.openstreetmap.org`-based domain we created and have tested it to confirm that it supports HTTP/2 and resolves to a valid SSL cert.

tomhughes:
I went with https://github.com/openstreetmap/openstreetmap-website/commit/685561687022262a1af5fc09c5617757d8529f3c instead of this, so that we can change the URL for the production site without hard coding it here.

mmd-osm:
I haven’t seen any discussion on Fastly‘s privacy policy nor any update to our own privacy policy. Does the following still apply to Fastly? Is everyone happy with this?

> Map tiles are provided by a global network of cache servers, which tile server your browser or app access is determined dynamically by geolocation of the IP address and selection of the cache server "nearest" to you.

simonpoole:
I quote form a mail (on the topic of GDPR) to the OSMF board and the OWG from February this year:

> I need to stress, that If our corporate structure or operations change
> in a significant way this will have to be revisited, as in general we
> need to gauge the privacy impact of any changes that affect personal
> data. This includes adding or changing the location of data centres,
> storing data with third party providers, utilizing software that exposes
> personal data to third parties (for example iD) and giving new
> volunteers access to our systems.

So you can guess the happiness level.

iandees:
Let's move off-topic discussion to a different ticket or to a mailing list thread.

simonpoole:
There is no discussion, it's the boards head on the block, and they seem to be happy with the situation, So I really don't care.

mmd-osm:
Ok, sounds good to me, let the board figure it out, I don’t want to discuss anything either.

PR #2723, opened 2020-07-22
Use Brakeman for static code analysis

gravitystorm:
This is my preferred version of #2229 

I've also reviewed the currently ignored warnings. Many of them are in code that will soon be either removed (amf_controller) or refactored (trace file handling) so I'm not prioritising fixing them at the moment. But at least having Brakeman included will prevent other errors from being accidentally introduced.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ script:
   - bundle exec rubocop -f fuubar
   - bundle exec rake eslint
   - bundle exec erblint .
+  - bundle exec brakeman -q
   - bundle exec rake db:structure:dump
   - sed -e "/idle_in_transaction_session_timeout/d" -e 's/ IMMUTABLE / /' -e "/^--/d" db/structure.sql > db/structure.actual
   - diff -uw db/structure.expected db/structure.actual
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,14 @@ You can view test coverage statistics by browsing the `coverage` directory.
 The tests are automatically run on Pull Requests and other commits with the
 results shown on [Travis CI](https://travis-ci.org/openstreetmap/openstreetmap-website).
 
+## Static Analysis
+
+We also perform static analysis of our code. You can run the analysis yourself with:
+
+```
+bundle exec brakeman -q
+```
+
 ## Comments
 
 Sometimes it's not apparent from the code itself what it does, or,
--- a/Gemfile
+++ b/Gemfile
@@ -138,6 +138,7 @@ end
 
 # Gems needed for running tests
 group :test do
+  gem "brakeman"
   gem "capybara", ">= 2.15"
   gem "coveralls", :require => false
   gem "erb_lint", :require => false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -110,6 +110,7 @@ GEM
       autoprefixer-rails (>= 9.1.0)
       popper_js (>= 1.14.3, < 2)
       sassc-rails (>= 2.0.0)
+    brakeman (4.8.2)
     browser (4.2.0)
     builder (3.2.4)
     bzip2-ffi (1.0.0)
@@ -484,6 +485,7 @@ DEPENDENCIES
   binding_of_caller
   bootsnap (>= 1.4.2)
   bootstrap (~> 4.5.0)
+  brakeman
   browser
   bzip2-ffi
   cancancan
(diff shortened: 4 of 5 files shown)


tomhughes:
So it's all very well enabling it and just turning off all the "difficult" checks but what I really want to know in order to evaluate this is what happens once we start trying to turn them back on again...

I know in the original I expressed concern about the false positive rate though I'm not sure what I was using to evaluate that because I can't see any sort of human readable report anywhere on that ticket, presumably because they had again all been suppressed, all be it in a different way.

If we accept that false positives are unavoidable and that they will need to be suppressed, presumably using the mechanism from the original pull request, then my question, how robust is that suppression mechanism? Is it tightly tied to the current state of the code so that it will keep retriggering and having to be suppressed again when the code changes? or on the flip side is it going to to continue suppressing warnings when code is changed and needs to be re-evaluated to be sure that a false positive really is still a false positive?

I mean I can see it stores a source location, and the source text, and a fingerprint. What's not clear is what feeds into the fingerprint and which attributes exactly it compares when considering whether to suppress a warning.

The alternative to suppressing false positives of course is to change the code to avoid them but it may be hard to do that in ways that aren't extremely intrusive. Is there are documentation on suggested approaches (beyond the comically trivial) to writing code that avoids triggering these warnings?

tomhughes:
Incidentally if we are likely to need a suppression file at some point then it probably make sense to create a rake task to run this so that we can hide the details rather than adding more arguments to the command in the readme.

gravitystorm:
> I know in the original I expressed concern about the false positive rate though I'm not sure what I was using to evaluate that because I can't see any sort of human readable report anywhere on that ticket, presumably because they had again all been suppressed, all be it in a different way.

So IIRC in the original PR the command was set to not fail the build, so you could see all the output in the travis logs. That might be where you saw the list?

Many of the reports have been fixed already (e.g. in #2570 #2597), and more likely to disappear due to refactoring, so I think any list of remaining false positives, if there are ever any, is going to be small. 

> If we accept that false positives are unavoidable and that they will need to be suppressed, presumably using the mechanism from the original pull request, then my question is how robust is that suppression mechanism?

I've no idea. This PR takes a different approach of switching off types of test as a whole. So another option is to avoid the false-positives fingerprints list altogether, and just switch off the type of test if there are any false positives found in that type. Even then, we'll still be getting a lot of benefit from all the other active types.

tomhughes:
Out of interest I did a run with all the checks turned back on to see what they found... My quick summary is:

#### Command Injection: 6

These I believe all relate to interpolating trace file names into commands are are all safe as the names in question have been generated by us. Some of the easier ones have been fixed but the ones which redirect output are harder I suspect.

#### Cross-Site Scripting: 18

These all relate to using the `legale` parameter on the terms page to lookup the correct languages/ It's safe in that the worse you can do is pass a bogus language which we would then fail to lookup. I'm not really sure what it is about that use of a parameter (over all others) that it doesn't like, or what a fix might look like.

#### Dynamic Render Path: 3

These all appear to relate to it just not understanding the `render @collection` construct in rails and thinking that the collection is in fact an action which is dependent on a parameter when in fact the action is fixed it's the set of objects to be rendered that depends on the parameter.

#### File Access: 6

This is the trace parameter file names again - these ones in particular should go away once the trace files are moved to attachments.

#### Redirect: 10

This is a mixture of cases where we use a `referer` parameter to preserve an original URL that we later redirect back to and cases where we preserve a subset of input parameters across a redirect.

It's not immediately clear to me if these are really a problem or what would be considered a "safe" way of writing them.

I mean clearly somebody could say publish a link to our login page with a referer parameter pointing to a third party site but it's not clear what it gets them - in principle maybe allows whitewashing by making it look like a link to us but it's one where the user would have to have an OSM account to proceed through to the target which makes it seem unlikely to be a popular target.

We could validate that the referer pointed to openstreetmap.org but would that silence brakeman?

#### SQL Injection: 6

A few of these are just silly and can be easily fixed, others are deliberate cases of injecting code (generated by us, so safe) to do complicated selections and are hard to avoid.

gravitystorm:
> It's not immediately clear to me if these are really a problem or what would be considered a "safe" way of writing them.

Brakeman has documentation on how to fix them at https://brakemanscanner.org/docs/warning_types/redirect/ . 

But they are disabled in this PR, so why does it matter?

tomhughes:
Well I was trying to understand what sort of things it is able to tell us about and how useful those reports are how they can be fixed.

Given that everything else has been fixed already the only way I could generate any reports to evaluate was to re-enable those checks, which as I understand it is a long term goal anyway.

Thanks for the link anyway - that's exactly the sort of thing I was interested in seeing.

PR #2725, opened 2020-07-22
Avoid testing against empty db queries

### LINKED ISSUE #2614: Some tests still rely on fixtures
Some of our tests are written in a way that despite the fixtures being removed (in https://github.com/openstreetmap/openstreetmap-website/pull/1556) they still pass. For example:

* [api/ways_controller_test#test_full](https://github.com/openstreetmap/openstreetmap-website/blob/8b588ee1f9eef3ee1ae7deddf05a3914185c8d5c/test/controllers/api/ways_controller_test.rb#L66-L90) - this test iterates over `Way.all` but there are no ways in the database, so this does nothing.
* [changesets_controller_test#test_index](https://github.com/openstreetmap/openstreetmap-website/blob/8b588ee1f9eef3ee1ae7deddf05a3914185c8d5c/test/controllers/changesets_controller_test.rb#L33-L47) - this checks to ensure that some changesets have rendered, but there are none in the database and checking against an empty `Changesets.all` returns no errors
* similar for `test_index_xhr`, `test_index_bbox`
* `test_index_max_id` - I can't even figure out what this is doing, but it involves checking the id of changesets without any being in the database.

I think the common thread is these tests are all using database queries to build the list of expected objects, instead of stating them explicitly. This allows empty databases to pass (since the list of items shown on the page == list of items in query == 0). It's much better to be explicit about which items to expect.

So we should fix the above tests. And we should double-check if there are any other tests that involve queries to build the list of expected objects. I've grepped around a bit but maybe there's a better way.

gravitystorm:
This PR fixes #2614. Minor refactoring of tests to create relevant objects, and test against them. Previously these tests were asserting against empty database queries, since the fixtures were removed a few years ago.

### DIFF
--- a/test/controllers/changesets_controller_test.rb
+++ b/test/controllers/changesets_controller_test.rb
@@ -146,6 +146,7 @@ class ChangesetsControllerTest < ActionDispatch::IntegrationTest
     private_user = create(:user, :data_public => true)
     friendship = create(:friendship, :befriender => private_user)
     changeset = create(:changeset, :user => friendship.befriendee, :num_changes => 1)
+    _changeset2 = create(:changeset, :user => create(:user), :num_changes => 1)
 
     get friend_changesets_path
     assert_response :redirect
@@ -169,7 +170,9 @@ class ChangesetsControllerTest < ActionDispatch::IntegrationTest
   def test_index_nearby
     private_user = create(:user, :data_public => false, :home_lat => 51.1, :home_lon => 1.0)
     user = create(:user, :home_lat => 51.0, :home_lon => 1.0)
+    far_away_user = create(:user, :home_lat => 51.0, :home_lon => 130)
     changeset = create(:changeset, :user => user, :num_changes => 1)
+    _changeset2 = create(:changeset, :user => far_away_user, :num_changes => 1)
 
     get nearby_changesets_path
     assert_response :redirect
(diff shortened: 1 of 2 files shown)


test/controllers/changesets_controller_test.rb
```diff
@@ -131,7 +145,7 @@ def test_index_user_not_found
   def test_index_friends
     private_user = create(:user, :data_public => true)
     friendship = create(:friendship, :befriender => private_user)
-    create(:changeset, :user => friendship.befriendee)
+    changeset = create(:changeset, :user => friendship.befriendee, :num_changes => 1)
```
tomhughes, line 148: Shouldn't this create some "other" changesets to make sure we only get the expected one?

test/controllers/changesets_controller_test.rb
```diff
@@ -147,15 +161,15 @@ def test_index_friends
...
+    check_index_result([changeset])
   end
 
   ##
   # Checks the display of the nearby user changesets listing
   def test_index_nearby
     private_user = create(:user, :data_public => false, :home_lat => 51.1, :home_lon => 1.0)
     user = create(:user, :home_lat => 51.0, :home_lon => 1.0)
-    create(:changeset, :user => user)
+    changeset = create(:changeset, :user => user, :num_changes => 1)
```
tomhughes, line 172: Same here?

test/controllers/changesets_controller_test.rb
gravitystorm: Done! Thanks for picking this up

test/controllers/changesets_controller_test.rb
gravitystorm: Also done

PR #2728, opened 2020-07-22
Add 404.png used by sites using old OpenLayers example

Firefishy:
Add a message in the old OpenLayer 404.png


Firefishy:
We respond with around 8 million 410 - GONE responses for this 404.png tiles a day. Having the file available and responding with cache headers will hopefully reduce the number of requests.

tomhughes:
The requests don't really cost anything, and this certainly doesn't belong here - it's an operational issue for osm.org not an issue for this repo.

Firefishy:
8 million requests a day / 100 requests a second (>32 million fd stats and >16 million fs reads) does cost and is not insignificant. We already have legacy OpenLayers files in this repo, adding 1 static file will mostly get rid of the OpenLayers DDOS failure request loop that has been hitting us for years.

Would you prefer the 404.png file was dropped by chef?

Firefishy:
~2.8GB of access.log logs per day just to store the 410 request responses.

Firefishy:
This PR is now made redundant by updated to https://github.com/openstreetmap/chef/pull/325

PR #2729, opened 2020-07-22
Remove legacy OpenLayers files

Firefishy:
Remove legacy OpenLayers code, which for www.openstreetmap.org is now managed by https://github.com/openstreetmap/chef/tree/master/cookbooks/web/files/default/static/openlayers


gravitystorm:
Awesome, thank you!

PR #2735, opened 2020-07-27
Updated About Page Community Knowledge section for en and en-GB locales

migurski:
Implements [Martin Koppenhoefer’s suggestions for the _About Page_](https://lists.openstreetmap.org/pipermail/talk/2020-July/085163.html) for just `en` and `en-GB` locales.

Preview:

Screen Shot 2020-07-27 at 1 15 39 PM

### DIFF
--- a/config/locales/en-GB.yml
+++ b/config/locales/en-GB.yml
@@ -1311,8 +1311,8 @@ en-GB:
         is accurate and up to date.
       community_driven_title: Community Driven
       community_driven_html: |-
-        OpenStreetMap's community is diverse, passionate, and growing every day. Our contributors include enthusiast mappers, GIS professionals, engineers running the OSM servers, humanitarians mapping disaster-affected areas, and many more.
-        To learn more about the community, see the <a href='https://blog.openstreetmap.org'>OpenStreetMap Blog</a>, <a href='%{diary_path}'>user diaries</a>, <a href='https://blogs.openstreetmap.org/'>community blogs</a>, and the <a href='https://www.osmfoundation.org/'>OSM Foundation</a> website.
+        OpenStreetMap's community is diverse, passionate, and growing every day. Our contributors include people like you: enthusiast mappers, GIS professionals, engineers running the OSM servers, humanitarians mapping disaster-affected areas, and many more.
+        To learn more about the community, see <a href="https://wiki.openstreetmap.org/wiki/How_to_contribute">How To Contribute</a>, the <a href='https://blog.openstreetmap.org'>OpenStreetMap Blog</a>, <a href='%{diary_path}'>user diaries</a>, <a href='https://blogs.openstreetmap.org/'>community blogs</a>, and the <a href='https://www.osmfoundation.org/'>OSM Foundation</a> website.
       open_data_title: Open Data
       open_data_html: |-
         OpenStreetMap is <i>open data</i>: you are free to use it for any purpose
--- a/config/locales/en-GB.yml
+++ b/config/locales/en-GB.yml
@@ -1311,8 +1311,8 @@ en-GB:
         is accurate and up to date.
       community_driven_title: Community Driven
       community_driven_html: |-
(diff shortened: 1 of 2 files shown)


mmd-osm:
That Wiki page "How to contribute" hasn't seen much of an update in a couple of years, and looks a bit abandoned (not to mention that it doesn't look all that exciting either). It this really a page we'd like to mention on the main page? 

https://welcome.openstreetmap.org/ looks more engaging. Are contents there still up-to-date and relevant?

tomhughes:
Yes I'm somewhat reluctant to add links to wiki pages because we have no control over what happens to them - the content may be reasonable now (though it's not clear that is the case here) but it could change at any time.

We already have things like https://www.openstreetmap.org/fixthemap and https://welcome.openstreetmap.org/ which are more controlled as introductions.

migurski:
Sounds good, I’ve modified the links to point to the Welcome Mat instead.

Firefishy:
Psst (whispers) The top page on OpenStreetMap.org is https://www.openstreetmap.org/copyright, it unsurprisingly also has the highest bounce rate. It would be good to also have a think about how we could improve that page and get move people who arrive on /copyright converted into mappers.

migurski:
Good point, I made a new issue: https://github.com/openstreetmap/openstreetmap-website/issues/2736

config/locales/en-GB.yml
```diff
@@ -1311,8 +1311,8 @@ en-GB:
         is accurate and up to date.
       community_driven_title: Community Driven
       community_driven_html: |-
-        OpenStreetMap's community is diverse, passionate, and growing every day. Our contributors include enthusiast mappers, GIS professionals, engineers running the OSM servers, humanitarians mapping disaster-affected areas, and many more.
```
gravitystorm, line 1314: From CONTRIBUTING.md

> If you make a change that involve the locale files (in config/locales) then please only submit changes to the en.yml file. The other files are updated via Translatewiki and should not be included in your pull request.

config/locales/en.yml
```diff
@@ -1367,11 +1367,12 @@ en:
...
       community_driven_html: |
         OpenStreetMap's community is diverse, passionate, and growing every day.
-        Our contributors include enthusiast mappers, GIS professionals, engineers
+        Our contributors include people like you: enthusiast mappers, GIS professionals, engineers
         running the OSM servers, humanitarians mapping disaster-affected areas,
         and many more.
-        To learn more about the community, see the
-        <a href='https://blog.openstreetmap.org'>OpenStreetMap Blog</a>,
+        To learn more about the community, see the OpenStreetMap
+        <a href='https://welcome.openstreetmap.org/'>Welcome Mat</a>,
```
gravitystorm, line 1374: The "welcome mat" is a page designed specifically for an organisational audience, and is not aimed at individual contributors.

So if we're aiming this text at "people like you: <list of individuals>" then I think we're mixing up our messaging by having the welcome mat as the first link. I agree that the first link should cover similar ground to some of the topics shown on the welcome mat, but it should be written for an audience of individuals. I'm open to suggestions for what content is already out there which fills that purpose.

dieterdreist:
sent from a phone

> On 27. Jul 2020, at 23:35, Tom Hughes <notifications@github.com> wrote:
> 
> We already have things like https://www.openstreetmap.org/fixthemap and https://welcome.openstreetmap.org/ which are more controlled as introductions


the welcome mat is clearly aiming at organizations, and „contributing“ on this page seems to suggest importing data, at least that’s the main focus as I see it.

The fix the map link is better suited, and has links to community resources like the wiki and the beginner‘s guide.

Regarding the potential problem of someone vandalizing the howto wiki page: while it is definitely possible as with any wikipage, the solution would be detecting and reverting it quickly. Editing the page could (?) be restricted to autoconfirmed users maybe?

config/locales/en-GB.yml
migurski: Thanks. Maybe this could be an automated check? https://github.com/openstreetmap/openstreetmap-website/pull/2740 is a nudge in this direction that might be interesting.

migurski:
What happens next?

migurski:
Closing this PR since there doesn't seem to be an obvious path forward.

PR #2740, opened 2020-07-28
Add authorship check for config/locales

migurski:
Andy [pointed out the contribution rules for locales](https://github.com/openstreetmap/openstreetmap-website/pull/2735#discussion_r461356191): only `en.yml` is meant to be changed in PRs while other locales are edited viw TranslateWiki.

This PR adds a parallel Travis test to look for cases where this contribution rule has been violated. It lets the primary unit tests run to completion but offers quick feedback on authorship. 

Some changes to locale files are structural and exempt from the restriction. A check failure would need to be ignored in that case, or somehow suppressed. Would core project contributors like Andy or Tom be included in the TranslateWiki exception? Maybe this would only apply to external PRs?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,41 @@
-dist: bionic
-language: ruby
-rvm:
-  - 2.5.3
-cache:
-  - bundler
-addons:
-  postgresql: 9.5
-  apt:
-    packages:
-      - firefox-geckodriver
-      - libarchive-dev
-      - libgd-dev
-      - libffi-dev
-      - libbz2-dev
-services:
-  - memcached
-before_script:
-  - sed -e 's/ IMMUTABLE / /' -e "/^--/d" db/structure.sql > db/structure.expected
-  - psql -U postgres -c "CREATE DATABASE openstreetmap"
-  - psql -U postgres -c "CREATE EXTENSION btree_gist" openstreetmap
-  - psql -U postgres -f db/functions/functions.sql openstreetmap
-  - cp config/travis.database.yml config/database.yml
-  - cp config/example.storage.yml config/storage.yml
-  - touch config/settings.local.yml
-  - echo -e "---\nmemcache_servers:\n  - 127.0.0.1" > config/settings/test.local.yml
-  - bundle exec rake db:migrate
-  - bundle exec rake i18n:js:export
-  - bundle exec rake yarn:install
-script:
-  - bundle exec rubocop -f fuubar
-  - bundle exec rake eslint
-  - bundle exec erblint .
-  - bundle exec brakeman -q
-  - bundle exec rake db:structure:dump
-  - sed -e "/idle_in_transaction_session_timeout/d" -e 's/ IMMUTABLE / /' -e "/^--/d" db/structure.sql > db/structure.actual
-  - diff -uw db/structure.expected db/structure.actual
-  - bundle exec rake test:db
+jobs:
+  include:
+  - name: "Website Code"
+    dist: bionic
+    language: ruby
+    rvm:
+      - 2.5.3
+    cache:
+      - bundler
+    addons:
+      postgresql: 9.5
+      apt:
+        packages:
+          - firefox-geckodriver
+          - libarchive-dev
+          - libgd-dev
+          - libffi-dev
+          - libbz2-dev
+    services:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,38 +1,47 @@
...
-  - bundle exec erblint .
-  - bundle exec brakeman -q
-  - bundle exec rake db:structure:dump
-  - sed -e "/idle_in_transaction_session_timeout/d" -e 's/ IMMUTABLE / /' -e "/^--/d" db/structure.sql > db/structure.actual
-  - diff -uw db/structure.expected db/structure.actual
-  - bundle exec rake test:db
+jobs:
+  include:
+  - name: "Locale Authorship"
+    install: skip
```
mmd-osm, line 4: This one executes on "xenial". For consistency, maybe add "dist: bionic" as well?

mmd-osm:
Sometimes, maintainers need superpowers: https://github.com/openstreetmap/openstreetmap-website/commit/045ff92ac9d3cd824002be6fe67d636b9ba12b0b#diff-9dd965712449996048bac1647804859e

Locking them out probably won't make them all that happy.

tomhughes:
Well why make it a separate job at all when every other check runs in the one job? The whole thing just looks like massive overkill to me given that as @migurski has already identified it's as simple as saying they're not allowed.

migurski:
Tom, a separate job matches the pattern we’re already using with Coveralls, where “lint-like” tests can be kept separate from the lengthy core application logic tests. In a case where someone wants to override the failure, there’s still positive feedback from the primary tests. I’ve found this to be a great pattern in other repos, hopefully interesting for here.

tomhughes:
There are any number of lint like checks in the existing job. Also coveralls isn't separate except in as much as the results are uploaded to coveralls at the end of the run which then analyses them and reports a result to github.

migurski:
Okay. Are you suggesting that it should be integrated into the main Travis test loop, or just asking for clarification?

tomhughes:
Well if it's worth doing then yes, but I'm not convinced it is worth doing.

I mean we can argue about whether the various different checks should be done in parallel but that's a separate discussion and we should either do everything that way or do everything together as we do now.

The more important question is whether, given this isn't a major issue, we can find a way to make it reliable.

migurski:
Cool, I’ve moved the locale authorship check into a single Travis test block.

config/locales/check-authors.sh
```diff
@@ -0,0 +1,28 @@
...
+# Assume everything will be fine
+STATUS=0
+
+for FILE in *.yml; do
+  # Don't check authorship of en.yml
+  if [ $FILE != en.yml ]; then
+    HUMAN_AUTHOR=`git log --format='%ae' $COMMIT_RANGE -- $FILE | grep -v @translatewiki.net | head -n1`
+    if [ $HUMAN_AUTHOR ]; then
+      # Mark for failure if changes were made by anyone other than translatewiki.net
+      echo Found $HUMAN_AUTHOR in $FILE
```
Firefishy, line 22: Likely a good idea to add more output here to explain the policy reason for the failure.

migurski:
I took your suggestion, Grant. I moved the authorship check down below the other tests: if it’s going to be something we want to ignore from time to time, we’ll want the tests that can’t be ignored to complete beforehand. A parallel test run is better in this situation.

mmd-osm:
So the restriction to pull requests would be still missing, I guess. Docs suggest something like:

```
  - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash ./travis/run_on_pull_requests; fi'
```

migurski:
Good suggestion, here it is in action for an active PR:

Screen Shot 2020-07-29 at 11 43 41 AM

This is what it looks like for a normal push:

Screen Shot 2020-07-29 at 11 44 04 AM

migurski:
I think this PR addresses all the relevant feedback. I’m curious if there’s interest in merging it; if not I will close it out.

mmd-osm:
Just leave it open, it won't cost you a thing.

migurski:
I’m going to close this out. There’s a distributed cost to stale PRs that I’d like to avoid inflating. We can re-open this if it becomes interesting again!

PR #2777, opened 2020-08-16
Black list 2010 and 2016 imagery for the Canton Schaffhauen

simonpoole:
Sorry that this is necessary but as iD is currently dead in the water this seems to be the only expedient way of stopping people to use these two layers, which are not available on OSM compatible terms. Hopefully in the long term this can be removed again. 

PS: tested locally with iD.


simonpoole:
Thanks @tomhughes

dieterdreist:
there’s a typo in the title, maybe it can still be fixed for beauty reasons

Issue #2732, opened 2020-07-25
The comma in the Swedish coordinate format should be a period


zhouhana:
In the Swedish version of https://www.openstreetmap.org/ , coordinates are erroneously displayed with a comma instead of a period. This makes some other programs – for instance Gramps – unable to handle coordinate values that are copied from the Swedish version of the site.

(Note that in Sweden/Swedish, the decimal separator is still a comma instead of a period. For example, the American number 1,000.50 is in Sweden/Swedish written as 1.000,50 or 1 000,50 or 1000,50. I guess this is the source of the comma/period error.)

![openstreetmap](https://user-images.githubusercontent.com/493629/88463725-272ec380-ceb5-11ea-8d22-fa401180625a.png)

mmd-osm:
There's a very old trac ticket requesting the same for the German translation: https://trac.openstreetmap.org/ticket/5082 ("Coordinate display for node browe pages: Decimal separator should not be translated").

tomhughes:
Well I think we need to understand the issue first - the fact that another program can't parse it isn't in itself evidence that we are wrong, and I say that as a Gramps user myself...

If I understand correctly the claim being made is that, despite numbers generally being written with a comma as the decimal separator, for coordinates the decimal point is normally used, at least in Swedish and German? Does this hold more widely for other european countries that would normally use comma?

Or is the problem here that we'd using comma both as the decimal point and to separate the coordinates, thereby confusing programs like Gramps that try to parse a coordinate pair?

HolgerJeromin:
> for coordinates the decimal point is normally used, at least in Swedish and German?

Not in the normal environment. See here an example of an coordinate at a [residential house](https://wohnglueck.de/service/fertighaus/haus/viebrockhaus-vertriebs-gmbh-and-co-betrieb-kg/maxime-300):
![image](https://user-images.githubusercontent.com/2410353/88477690-6ce59d80-cf42-11ea-8975-3fd7512b98b3.png) ...

> Or is the problem here that we'd using comma both as the decimal point and to separate the coordinates, thereby confusing programs like Gramps that try to parse a coordinate pair?

... but in the computer environment you find more software products which does not support the three commas.

mmd-osm:
ISO 6709 “Standard representation of geographic point location by coordinates is the international standard for representation of latitude, longitude and altitude for geographic point locations.” mentions:

The decimal point is a part of the value, thus must usually be configured by the operating system

[...]

PR #2783, opened 2020-08-19
Avoid lego translations for coordinates

gravitystorm:
Use the i18n system for joining coordinate values together, while avoiding making any changes or standardisation involving the exact symbol used in each situation.

Refs #2732

This was mentioned in passing in the linked issue, so I thought I could at least fix that minor point. It doesn't change anything wrt decimal separators, and it's worth noting that we've been inconsistent in the past with `,` vs `;` within pairs of coordinates and this PR deliberately doesn't change either option, just moves them into the i18n files.


tomhughes:
Ha I was considering that as a solution the other day ;-)

Issue #2184, opened 2019-03-19
Link secondary WIkipedia tags - brand:wikipedia, operator:wikipedia, subject:wikipedia etc


matkoniecz:
Currently wikipedia article specified in `wikipedia` tag is linked in tag list.

But there are also secondary Wikipedia tags like `subject:wikipedia` that are not linked. It would be desirable to create link for both cases.

![Selection_006](https://user-images.githubusercontent.com/899988/54638391-3e70cc00-4a8b-11e9-9406-4d9127cfbc80.png)

![Selection_007](https://user-images.githubusercontent.com/899988/54638430-4cbee800-4a8b-11e9-823a-910631fa9b4a.png)

See https://www.openstreetmap.org/node/5167698292 and https://www.openstreetmap.org/node/1350073645

See https://wiki.openstreetmap.org/wiki/Key:wikipedia#Secondary_Wikipedia_links

See https://taginfo.openstreetmap.org//search?q=%3Awikipedia

`brand:wikipedia`, `operator:wikipedia`, `species:wikipedia`, `subject:wikipedia`, `name:wikipedia`, `architect:wikipedia`, `artist:wikipedia`, `name:etymology:wikipedia` are probably ones worth linking

HolgerJeromin:
Ref #788

nachanon:
This bug is still reproducible.

I have recently added an OpenStreetMap node-type POI representing a memorial statue ([`historic=memorial`](https://wiki.openstreetmap.org/wiki/Tag:historic=memorial),[`memorial=statue`](https://wiki.openstreetmap.org/wiki/Tag:memorial=statue)) with information on the person statue represents in the  in the [`subject:wikipedia`](https://wiki.openstreetmap.org/wiki/Key:subject:wikipedia) key, using the same format as [`wikipedia`](https://wiki.openstreetmap.org/wiki/Key:wikipedia) tag.

But as soon as I tried to verify the entered entry on main OpenStreetMap interface, I found that the [`subject:wikipedia`](https://wiki.openstreetmap.org/wiki/Key:subject:wikipedia) key's name was properly linked to its OSM Wiki description, **but the key's value is not linked to Wikipedia article in question**...

* [Field Marshall Bhanurangsi Savangwongse statue](https://www.openstreetmap.org/node/6567173391):
	
	![](https://u.cubeupload.com/xwindows/giEvia.png)
	
[...]

PR #2812, opened 2020-09-05
Add species to wikidata_links helper

tordans:
I propose to add the key https://wiki.openstreetmap.org/wiki/Key:species:wikidata to the wikidata_links helper, so it may be linked on osm.org as well.

The key status is "Status: de facto" with ~9k uses.

### DIFF
--- a/app/helpers/browse_tags_helper.rb
+++ b/app/helpers/browse_tags_helper.rb
@@ -105,7 +105,7 @@ module BrowseTagsHelper
         :title => value
       }]
     # Key has to be one of the accepted wikidata-tags
-    elsif key =~ /(architect|artist|brand|name:etymology|network|operator|subject):wikidata/ &&
+    elsif key =~ /(architect|artist|brand|name:etymology|network|operator|species|subject):wikidata/ &&
           # Value has to be a semicolon-separated list of wikidata-IDs (whitespaces allowed before and after semicolons)
           value =~ /^[Qq][1-9][0-9]*(\s*;\s*[Qq][1-9][0-9]*)*$/
       # Splitting at every semicolon to get a separate hash for each wikidata-ID


tordans:
Testcases:

- Example: https://www.openstreetmap.org/node/346751757
- List: https://sophox.org/#%23List%20all%20OSM%20objects%20with%20a%20place%20tag%0ASELECT%20%3FosmId%20WHERE%20%7B%0A%20%20%3FosmId%20osmt%3Aspecies%3Awikidata%20%3Fspecies%20%20.%0A%0A%20%20service%20%3Chttps%3A%2F%2Fquery.wikidata.org%2Fsparql%3E%20%7B%0A%20%20%20%20OPTIONAL%20%7B%20%3Fspecies%20wdt%3AP225%20%3Fscientific.%20%7D%0A%20%20%20%20SERVICE%20wikibase%3Alabel%20%7Bbd%3AserviceParam%20wikibase%3Alanguage%20%22%5BAUTO_LANGUAGE%5D%2Cen%22.%7D%0A%20%20%7D%0A%20%20%0A%7D%20LIMIT%2050

mmd-osm:
Wasn’t the idea to automate link generation via #2621 ?

boothym:
Would it be possible for you to add some others as described in #2184?

tordans:
@boothym this PR is about Wikidata. The way I read the issue you referenced, it's about Wikipedia.￼

For this PR I am still waiting for guidance by the maintainers about how to proceed. ￼

hiddewie:
Good change :+1:, this key is also listed explicitly in the OpenStreetmap wiki: https://wiki.openstreetmap.org/wiki/Key:wikidata#Secondary_Wikidata_links and https://wiki.openstreetmap.org/wiki/Wikidata#As_a_tag_qualifier

pyrog:
See #2621

matkoniecz:
That would be useful as many people keep adding such info as `wikidata`/`wikipedia`

It is clear https://wiki.openstreetmap.org/wiki/Tagging_for_the_renderer but linking such objects would help to steer them the correct way

mmd-osm:
#3691 supersedes this pull request.

AntonKhorev:
Maybe the maintainers think that these links are a [bad idea](https://github.com/openstreetmap/openstreetmap-website/issues/2405#issuecomment-544203726)?

gravitystorm:
Superceded by #3691, which is now merged. Thank you @tordans in any case.

gravitystorm:
> Maybe the maintainers think that these links are a [bad idea](https://github.com/openstreetmap/openstreetmap-website/issues/2405#issuecomment-544203726)?

So my point of view is not that they are inherently bad, but I'm still not super keen on them:

* There is a more or less unbounded list of tags that could be interpreted in a similar manner. It's hard to know where any line could be drawn, and we could end up with 100+ different destinations for tags that link to external resource lookups (i.e. the tag is some kind of id, that links to a different external site).
* `browse_tags_helper` is fairly complex already. It's already been split into its own file, and at the current rate of change might need further splitting. All this code needs maintaining to some extent, so it's not zero-cost to add more and more complexity and support more tags.
* The external links are not just as easy as put-an-id-into-a-url. `wikidata_links` shows this already, where there's a whole section on handling semi-colon delineated ids (with and without whitespace...), and presumably some other future types of ids could bring their own similar complexities.

Beyond all that, I don't like that this is the major (only?) part of this codebase that closely interprets OSM tags. The rest of the website and API is (mostly, n.b. changeset `comment` tags) tagging agnostic. This is a powerful part of OpenStreetMap, that tags can be invented, changed, rethought and discarded without needing anything changed in the API. Except that here, in the browse_tags_helper, we *do* hard-code the meaning of certain tags. This is something that I would prefer to avoid, both to maintain the power and flexibility for the OSM community, and also because we know that this codebase is used for other projects, who get stuck with the same hardcoding.

I hope this provides some explanation. Obviously I've just merged #3691 so I'm not completely opposed to such tags but there are reasons for me to be cautious about them.

Identitaet:
I mean when were talking about maintaining those tags in general there's 
https://www.wikidata.org/wiki/Property:P1282
A wikidata propperty which links back to the OSM tag or key
I guess there's a potentitial for automating that

But does anyone care enough to do such a thing ? (Probably not)

PR #580, opened 2013-12-01
Move changeset comment to just after changeset info, & style as if a quote

danstowell:
The current browse pane for a way (or node etc) looks a bit odd with the changeset comment coming first, as if the comment was about the object one is looking at, which is often not the case. This could be rather confusing. In this PR I have moved it to just under the changeset metadata, and also styled it as a quote (in italiics, with quotemarks).

PR #2838, opened 2020-09-16
Style changeset descriptions with italics on browse pages

gravitystorm:
Alternative to #580

This reworks changeset descriptions to use italics instead of header elements, in order to make them look more like quotes.

I've always found it strange that they were headers, particularly on the node/way/relation pages, since they are more similar to a text paragraph or a quote than a description of the element being viewed. But it turns out to be quite tricky to refactor, since the descriptions appear in multiple contexts, and the partials are repeated in history pages and look strange unless the first item is a header.

So this is my proposed solution, feedback welcome.

Screenshot from 2020-09-16 18-14-59 Screenshot from 2020-09-16 18-15-31 Screenshot from 2020-09-16 18-13-57

### DIFF
--- a/app/views/browse/_common_details.html.erb
+++ b/app/views/browse/_common_details.html.erb
@@ -1,21 +1,21 @@
-<h4>
+<h4 class="details">
+  <%= t "browse.version" %>
+  #<%= common_details.version %>
+</h4>
+
+<p class="font-italic">
   <% if common_details.changeset.tags['comment'].present? %>
     <%= linkify(common_details.changeset.tags["comment"]) %>
   <% else %>
     <%= t "browse.no_comment" %>
   <% end %>
-</h4>
+</p>
 
 <div class="details">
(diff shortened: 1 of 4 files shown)


danstowell:
As the original author of #580: this looks good to me. Improves on the current styling, in consistency and scrutability.

tomhughes (approved):
I'm not really a fan of large blocks of italic but I don't have a better plan.

The version number heading above it on the object details page and the change time and changeset number below it could maybe be smaller but it looks like all that stuff is still pending being "bootstrapped" anyway so will likely change again so I don't think there's any need to block this.

matkoniecz:
Note that this resulted (at least to me) in changeset comments presented as less important than closing time.

Not sure why, but it seems that changeset comment text is grayed-out unlike closing date.

Maybe changeset comment was actually made grayish? Or it is side effect of italics? Or FF bug? Or an optical illusion?

screen05

PR #2864, opened 2020-09-30
Use the db:seed task to load languages

gravitystorm:
This is seed data that should reasonably be loaded into every database, even those that are otherwise empty (e.g. no geo data). Using the seeds process means that it will often be loaded by default (depending on exactly which db tasks are run), and is easier to find for existing rails developers.

Replaces #309

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,7 +37,6 @@ Metrics/ModuleLength:
 Naming/FileName:
   Exclude:
     - 'script/deliver-message'
-    - 'script/locale/reload-languages'
     - 'script/update-spam-blocks'
 
 Naming/MethodParameterName:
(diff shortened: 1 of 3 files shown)


tomhughes:
Doesn't `db/migrate/034_create_languages.rb` already do this as part of migrating the database?

gravitystorm:
I'm sure that it does, but I'm not entirely sure that it should! 

Most developers wouldn't expect seed data to be added as a side effect of the migrations, and wouldn't rely on it either. It starts getting more complex/flakey when setting up the database with e.g. `db:structure:load`, which of course only has the structure and not this language data. I think it's best to have it in the seeds file, where it belongs. But I'm happy to leave the migration stuff untouched.

As implemented I don't foresee any problems, since Language.load can be run multiple times without errors.

tomhughes:
Well I was mostly asking if it was safe to run it multiple times, although `db:structure:load` is not a supported way to setup our database so anybody doing that gets to keep the pieces anyway ;-)

PR #2865, opened 2020-09-30
changes Update .erb-lint.yml

bktheory123:


tomhughes:
If you're going to propose a change you really need to explain the purpose the change or why you think it should be made not just "update X".

mmd-osm:
Other commits of this user look very similar to #2866, so this is probably just random spam again.

tomhughes:
Indeed...

PR #2870, opened 2020-10-01
btter Update FAQ.md

pajeet-patel:


Coltor-L:
Beautifol is spelled incorrectly.

PR #2872, opened 2020-10-01
Welcome message: fix display of 'close' button

### LINKED ISSUE #2530: Welcome message hiding X to close
The X link to the close the welcome message is hidden by "Welcome to OpenStreetMap" on both Chrome and Firefox on Win10. 
Screenshot_2020-02-10 OpenStreetMap

Noticed a couple of people (presumably new users) mention this on twitter: 

https://twitter.com/ericghill/status/1226633066723962880

https://twitter.com/leftistthot420/status/1226198567087415296

### LINKED ISSUE #2850: welcome banner close button (x) hidden behind text
Clipboard03

my screen size/resolution is 1280x720

Chrome on Windows 10

hiddewie:
Add padding to the welcome message, because the 'close' button overlaps with the title.

This fixes #2530, fixes #2850.

Reference #2537, https://twitter.com/EugenesDIYDen/status/1276804381132812291

Screenshots for many locales: https://github.com/openstreetmap/openstreetmap-website/pull/2872#issuecomment-703289569


tomhughes:
Changing the width of the entire sidebar is a rather extreme way to resolve those bugs... Especially since we no idea how much longer the title that is overlapping the X might be in other languages...

hiddewie:
@tomhughes You are right. I reverted the sidebar width changes (although I still think they are a good idea). I added padding specifically for the header, as the cross is always 'behind' that line.

English

image

German

image

Locale `tl` (longest I could find)

image

Locale `te` (shortest I could find)

image

dieterdreist:
The 2 blue buttons should have the same height. If one is on 2 lines (like for example in German), the other should have the same height (and vertically center the label)

hiddewie:
@dieterdreist I am willing to fix that, but the change is unrelated to the issues that the close button is behind the header text in some languages.

The same goes for the blue Go button in the search bar in the `te` locale

gravitystorm:
Thanks for the PR, I reviewed it yesterday. This proposal is similar to #2537, and it has similar problems to #2537 , namely

* It introduces more custom CSS classes, which we're trying to remove
* It only addresses the welcome message, and not the other similar situations like the sidebar panels.

Based on some of the discussions on #2537 by @bezdna , I've started work on replacing the h2+close button by a flex-box based solution. This in particular addresses the point made about the close button not being related to the header itself (and so it shouldn't appear within the h2 element). I ran out of time this week before finishing it, but you can see what I'm thinking at https://github.com/gravitystorm/openstreetmap-website/commit/b3fc5bc71a4fd7a5b13aee3cad1c527de244da4e - feedback is welcome. I intend to use the same solution (maybe not the exact same partial, but at least the same divs and flexbox layout) for all of our current close buttons, so that they behave consistently.

hiddewie:
@gravitystorm The commit gravitystorm@b3fc5bc looks good to fix the issue more consistently across all sidebar titles. A (draft) pull request would be nice to see that the problem is being worked on structurally. The close button has been invisible for some locales for a long time, which makes the site hard to use on mobile browsers (and some desktops).

I will close this PR.

Issue #1455, opened 2017-02-22
Do not ask for application OAuth permissions repeatedly


Zverik:
Below is a screenshot of an excerpt from my OAuth authorized apps page. These three apps have the same id, and the website knows that I had granted some permissions to these apps before. Facebook, Google and most other OAuth providers do not ask for permissions the second time: a user chooses what they want to share only once. It would be great if osm.org did the same, and did not produce duplicate entries in that list as a result.

![2017-02-22 12 11 48](https://cloud.githubusercontent.com/assets/766031/23204627/c64866a8-f900-11e6-8d18-9e305ed414f4.png)

tomhughes:
Well the client application is supposed to keep using the same token, not keep on reauthenticating new tokens!

Zverik:
I agree, but the practive shows that our website keeps state much better than clients.

simonpoole:
Doesn't guessing that it is the same device/app open up a security hole about as big as the whole universe plus some?

Zverik:
Why? A user is authenticated and has granted permissions to an app. I often see Facebook remembering an app instead of asking for permissions again and again, and nobody has complained.

tomhughes:
Do google, facebook etc really convert the new request token into the same access token? or do they just collapse multiple live access tokens to one line in the UI with the revoke button revoking all active tokens for that application?

tomhughes:
Of course the other thing is that they're doing OAuth 2.0 so the details may be somewhat different anyway.

Zverik:
An oauth2 provider must return a new token each time. Revoking an app removes all the tokens. With oauth2, tokens have a limited lifetime, so old ones are cleaned automatically. In our case we can limit the number of tokens and e.g. remove all but 20 newest.

The token list page presented above does not give a user an undestranding of what each token is for, so collapsing these is a reasonable first step.

Zverik:
This is a screenshot of app management in Facebook. It is a simple list of apps with a visibility indicator. You can either remove an app, or adjust settings when you click an app's name.

simonpoole:
Well we are not doing OAuth 2.0. 

In our case OAuth 1.0a each app only has the Consumer Key for identification, the association between user and the access token is established during the handshake. As a result there is no way I can see how you could go from the Consumer Key to the access token without re-authenticating the user and re-authorizing the app.

HolgerJeromin:
Kind of related #1074

Zverik:
[...]

PR #2889, opened 2020-10-17
Improve the OAuth index page

hiddewie:
Some minor improvement for the OAuth token/application page:
- Open authorized application links in a new tab
- Make some English titles use `OAuth` instead of `oauth`.
- Add a line about the expiration of tokens.
- Use monospace for displaying OAuth tokens

~~It might be a good idea to hide (or even delete tokens) that are no longer valid.~~

Screenshot:

image

### DIFF
--- a/app/views/oauth_clients/index.html.erb
+++ b/app/views/oauth_clients/index.html.erb
@@ -10,13 +10,15 @@
     <tr>
       <th><%= t ".application" %></th>
       <th><%= t ".issued_at" %></th>
+      <th><%= t ".valid_to" %></th>
       <th>&nbsp;</th>
     </tr>
   </thead>
   <% @tokens.each do |token| %>
     <tr>
-      <td><%= link_to token.client_application.name, token.client_application.url %></td>
+      <td><%= link_to token.client_application.name, token.client_application.url, target: "_blank" %></td>
       <td><%= token.authorized_at %></td>
+      <td><%= token.valid_to %></td>
       <td>
         <%= form_tag({ :controller => "oauth", :action => "revoke" }, { :class => "standard-form" }) do %>
           <%= hidden_field_tag "token", token.token %>
--- a/app/views/oauth_clients/index.html.erb
+++ b/app/views/oauth_clients/index.html.erb
@@ -10,7 +10,6 @@
     <tr>
       <th><%= t ".application" %></th>
       <th><%= t ".issued_at" %></th>
-      <th><%= t ".valid_to" %></th>
       <th>&nbsp;</th>
     </tr>
   </thead>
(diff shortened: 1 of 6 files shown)


tomhughes:
What tokens do you think we are showing that are no longer valid?

hiddewie:
I assumed that all tokens had an expiration date. So any granted access would automatically be revoked after a certain time. This seems not to be the case.

Many applications ask (For example the HOT tasking manager) ask every for a new access token every X days (not sure how many). That is why I thought the tokens expired automatically, and I wanted to see how long each token would be valid.

This would be good to mention in text on the page.

mmd-osm:
I guess what you're seeing here with the HOT tasking manager has been discussed here before: https://github.com/openstreetmap/openstreetmap-website/issues/1455

Also, note that we're still on OAuth 1.0a at this time.

hiddewie:
Hmm indeed, thanks for finding that issue.

OK I removed the `token.valid_to` visualization because the tokens do not expire. I added a line of text under the table that tells how long the tokens are valid.

app/views/oauth_clients/index.html.erb
```diff
@@ -15,7 +15,7 @@
   </thead>
   <% @tokens.each do |token| %>
     <tr>
-      <td><%= link_to token.client_application.name, token.client_application.url %></td>
+      <td><%= link_to token.client_application.name, token.client_application.url, target: "_blank" %></td>
```
gravitystorm, line 18: We rarely use `target: "_blank"` for links - any particular reason to add it to this one?

Also this line causes the CI failures for both hash rocket syntax, and noopener when using target: "_blank"

gravitystorm (changes requested):

app/views/oauth_clients/index.html.erb
hiddewie: I would argue that any link that does not remain in the current application should be opened in a new tab. This makes users less likely to lose their way if they click on the OAuth application name. There are many more external links which should maybe also be opened in a new tab. It would be even better to annotate them with an icon like below to notify users of such links.

image

hiddewie:
@gravitystorm I added the noopener and fixed the linting.

simonpoole:
A tiny bit of nitpicking with the original text on the page that should be fixed in a follow up PR. The text currently says

`The following tokens have been issued to applications in your name:`

and then goes on to not display tokens but an application list. 

`Tokens have been issued in your name for the following applications;`

which would be more correct (as always this will require changing the id of the string, not a simply text change, to invalidate translations).

mmd-osm:
The list is indeed based on individual tokens and shows the respective application rather than any token details:

https://github.com/openstreetmap/openstreetmap-website/blob/master/app/views/oauth_clients/index.html.erb#L16

mmd-osm:
As we have OAuth 2.0 in the meantime and OAuth 1.0a will be deprecated and removed eventually, I‘d vote to close this one.

If there’s anything worth improving on the OAuth 2 pages, focus on that instead.

hiddewie:
>  I‘d vote to close this one.

Fine for me :+1:

danieldegroot2:
Switch to OAuth 2, from OAuth 1 and HTTP Basic Auth, is being tracked at https://github.com/openstreetmap/operations/issues/867

PR #2902, opened 2020-10-21
Link to official OWG Banner Policy page

seav:
- Link to the official Banner Policy page on the OWG website instead of a GitHub ticket discussion
- Reformat Markdown lists so that the rendered HTML results in properly nested lists


gravitystorm:
Thanks!

PR #2905, opened 2020-10-25
Bump copyright string version to force re-translation

simonpoole:
This should force re-translation of one section in the copyright page that didn't get its "version" bumped.

### DIFF
--- a/app/views/site/copyright.html.erb
+++ b/app/views/site/copyright.html.erb
@@ -48,7 +48,7 @@
 
   <h3><%= t ".legal_babble.credit_title_html", :locale => @locale %></h3>
   <p><%= t ".legal_babble.credit_1_html", :locale => @locale %></p>
-  <p><%= t ".legal_babble.credit_2_1_html", :locale => @locale %></p>
+  <p><%= t ".legal_babble.credit_2_2_html", :locale => @locale %></p>
   <p><%= t ".legal_babble.credit_3_1_html", :locale => @locale %></p>
   <p><%= t ".legal_babble.credit_4_html", :locale => @locale %></p>
   <p><%= image_tag("attribution_example.png",
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1617,7 +1617,7 @@ en:
         credit_1_html: |
           We require that you use the credit &ldquo;&copy; OpenStreetMap
           contributors&rdquo;.
-        credit_2_1_html: |
+        credit_2_2_html: |
           You must also make it clear that the data is available under the Open
           Database License. You may do this by linking to
           <a href="https://www.openstreetmap.org/copyright">this copyright page</a>.


simonpoole:
PS: note that I did this correctly in the original PR, it seems as if the translators simply C&P'd the old text back in, so we might be in for multiple iterations of this.

tomhughes:
If we've already bumped it once and the text hasn't changed since then there is clearly no point in just bumping it again - people that do the right thing will just have to repeat work they've already done and people that don't still won't.

simonpoole:
Well I originally thought it was my mistake, but the thing is that outside of getting people to actually look at the text again there is no way forward.

simonpoole:
Maybe @translatewiki has a suggestion on how to deal with this.

gravitystorm:
> This should force re-translation of one section in the copyright page that didn't get its "version" bumped.

Note that this definitely won't work as intended. Translatewiki now handles "translation key renames" automatically, which was partly driven by the refactoring work that we've been doing (e.g. renaming controllers leading to mass renaming of translation keys). So translatewiki will recognise this as a translation key rename, since the value hasn't changed, and automatically copy over all the existing translations from the old key to the new key.

gravitystorm:
I'm going to close this PR, since it can't be merged, and to avoid confusion. I think that getting translators to re-translate an existing string needs to be managed through the translatewiki interface, ideally for only those translations that are known to be defective. If you want to open an issue to describe or discuss or track the progress in fixing the translations, that would be fine with me.

simonpoole:
> 
> Note that this definitely won't work as intended. Translatewiki now handles "translation key renames" automatically, which was partly driven by the refactoring work that we've been doing (e.g. renaming controllers leading to mass renaming of translation keys). So translatewiki will recognise this as a translation key rename, since the value hasn't changed, and automatically copy over all the existing translations from the old key to the new key.

So this was likely the problem to start with? As in the translation system is now completely broken?

tomhughes:
It's "broken" for this use case but "fixed" for some other use cases. So maybe we should all just calm down a bit and lets work out how we can address this instead of getting agro...

mvl22:
As background: This arose from our request:

https://twitter.com/cyclestreets/status/1320462262440701955

simonpoole:
@mvl22 the problem is that it effects all translations (~100) not just the, slightly comical, en-GB one, which could naturally be fixed manually as a one off.

mvl22:
Could someone kindly clarify what the way forward is on getting the copyright statement updated? (I regret I wasn't totally clear from Andy's posting, not being familiar with how things are set up.) It seems pretty critical to me that the tiles are being advertised as the wrong license now.

(I know that reviewers at PLOS are still rejecting author maps unnecessarily because it states these are still the old license.)

tomhughes:
The statement has been updated. I mean I know PLOS are idiots (or we wouldn't be here in the first place) but surely they aren't relying on the translated string on the web site instead of going and reading the actual license?

tomhughes:
But basically if you want it "fixed" in a particular translation that head over to Translatewiki and fix it.

mvl22:
> The statement has been updated.

Can you point me where?

https://www.openstreetmap.org/copyright
still says that the tiles are licensed under Creative Commons.

I am trying to help out a researcher who is dealing with PLOS, but sending them this discussion or the blog post isn't going to cut it, when the official copyright page says precisely the opposite.

Sorry that I don't understand the technicalities of how pages are generated so cannot update this myself, nor would I feel confident making any such a critical content change even if I did.

tomhughes:
Right at the top of that page it gives you a link to the canonical english version (https://www.openstreetmap.org/copyright/en) if that's not the version you're looking at. That version says:

```
The map tiles in the “standard style” at www.openstreetmap.org are a Produced Work by the OpenStreetMap Foundation using OpenStreetMap data under the Open Database License. If you are using these tiles please use the following attribution: “Base map and data from OpenStreetMap and OpenStreetMap Foundation”. 
```

Which does not mention CC anywhere.

mvl22:
> Right at the top of that page it gives you a link to the canonical english version (https://www.openstreetmap.org/copyright/en) if that's not the version you're looking at.

Thanks - that makes sense now.

I will point them to that.

mvl22:
Well, guess what after all that...

"Please note, however, that the Open Database License is not compatible with PLOS ONE's copyright license either. [...]"

simonpoole:
@mvl22 well I don't think we can provide a solution to people being wrong.

simonpoole:
@mvl22 a good source (PLOS ONEs boss :-)) has now confirmed that the licence change for tiles has been reviewed by their lawyers and is now considered compatible with their licence. Naturally that doesn't mean that is going to trickle down to the reviewers any time soon.

mvl22:
I see that PLOS ONE have now published the change of policy:

https://journals.plos.org/plosone/s/figures#loc-maps


> Several sites provide maps that are compatible with CC BY licensing, including:
> [...]
> OpenStreetMap: OpenStreetMap map tiles are free to use as long as they are accompanied by the following attribution statement: “Base map and data from OpenStreetMap and OpenStreetMap Foundation”. Maps created using OpenStreetMap data must be accompanied by the following attribution statement: "Contains information from OpenStreetMap and OpenStreetMap Foundation, which is made available under the Open Database License.”

which is great news.

PR #2921, opened 2020-10-28
Prefer string interpolation to concatenation

gravitystorm:
Partial work on another `rubocop_todo` item

### DIFF
--- a/app/helpers/title_helper.rb
+++ b/app/helpers/title_helper.rb
@@ -6,7 +6,7 @@ module TitleHelper
   def set_title(title = nil)
     if title
       @title = TitleHelper.coder.decode(title.gsub("<bdi>", "\u202a").gsub("</bdi>", "\u202c"))
-      response.headers["X-Page-Title"] = ERB::Util.u(@title + " | " + t("layouts.project_name.title"))
+      response.headers["X-Page-Title"] = ERB::Util.u("#{@title} | #{t('layouts.project_name.title')}")
     else
       @title = title
       response.headers["X-Page-Title"] = ERB::Util.u(t("layouts.project_name.title"))
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -189,7 +189,20 @@ Style/OptionalBooleanParameter:
     - 'test/models/trace_test.rb'
     - 'test/models/tracetag_test.rb'
 
-# Offense count: 28
+# Offense count: 19
 # Cop supports --auto-correct.
 Style/StringConcatenation:
-  Enabled: false
+  Exclude:
+    - 'app/helpers/browse_tags_helper.rb'
+    - 'app/helpers/user_roles_helper.rb'
+    - 'app/jobs/trace_importer_job.rb'
+    - 'app/models/client_application.rb'
+    - 'app/views/changesets/index.atom.builder'
+    - 'lib/migrate.rb'
+    - 'lib/potlatch.rb'
+    - 'script/cleanup'
+    - 'script/deliver-message'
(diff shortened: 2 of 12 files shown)


app/helpers/title_helper.rb
```diff
@@ -6,7 +6,7 @@ def self.coder
   def set_title(title = nil)
     if title
       @title = TitleHelper.coder.decode(title.gsub("<bdi>", "\u202a").gsub("</bdi>", "\u202c"))
-      response.headers["X-Page-Title"] = ERB::Util.u(@title + " | " + t("layouts.project_name.title"))
+      response.headers["X-Page-Title"] = ERB::Util.u("#{@title} | #{t('layouts.project_name.title')}")
```
tomhughes, line 9: This is getting quite messy - might `format` be a better choice here?

tomhughes (commented):
I think I did look at this one before but I was concerned that where some of the things being joined were non-trivial expressions it might all become a bit unreadable.

As is often the case the style guide is unfortunately lacking in any rationale as to why you should prefer interpolation - my natural assumption would be that concatenation would generally be higher performance for example not that it matters most of the time.

I notice that `format` is an acceptable approach though, which would be easier to read where some of the things being interpolated are more complicated expressions.

app/helpers/title_helper.rb
gravitystorm: > my natural assumption would be that concatenation would generally be higher performance

I think concatenation is generally slower, since for `string1 + string2 + string3` you end up needing even more strings, since each `+` allocates a new string to hold the result of adding the two together, but...

> not that it matters most of the time

Indeed! :smile: So I'm much more interested in the consistency, and developer experience, unless there's some catastrophic performance difference.

> it might all become a bit unreadable.

Yeah, it's certainly not straightforward to read. My local editor has different syntax highlighting than on github, and it's definitely easier on the eye. So overall, I'm still happy with the PR, but also happy to come up with an alternative approach.

I thought I'd try out the format() approach that you suggested, e.g.

```diff
- response.headers["X-Page-Title"] = ERB::Util.u("#{@title} | #{t('layouts.project_name.title')}")
+ response.headers["X-Page-Title"] = ERB::Util.u(format("%s | %s", @title, t('layouts.project_name.title')))
```
... but that triggers another rubocop warning about labelling the string tokens in the format string. And with labelled tokens, we need to supply a hash of values. So in the end we get something like this:

```diff
- response.headers["X-Page-Title"] = ERB::Util.u("#{@title} | #{t('layouts.project_name.title')}")
+ response.headers["X-Page-Title"] = ERB::Util.u(format("%<title>s | %<project_title>s", :title => @title, :project_title => t("layouts.project_name.title")))
```
I find this even harder to read, since the format string `"%<title>s | %<project_title>s"` looks sort of but not quite like an interpolated string e.g. `"#{title} | #{project_title}"`, and the line ends up being quite long too.

However, one final option is then to use an interpolated string, but to avoid putting complex stuff inside it, by unrolling it slightly e.g.

```diff
- response.headers["X-Page-Title"] = ERB::Util.u("#{@title} | #{t('layouts.project_name.title')}")
+ project_title = t 'layouts.project_name.title'
+ response.headers["X-Page-Title"] = ERB::Util.u("#{@title} | #{project_title}")
```

What do you think?

tomhughes: I'd be happy with the format. I suspect you're just not used to seeing the named argument style but we do have other places using it. That said I've merged this and added 45d4242f9d29b00e69aeb016e99a0ce230eb71d9 on top using the variable based approach.

PR #2926, opened 2020-10-30
Updated OpenSearch image & attribution + updated Facebook logo.

AntoninCurtit:


tomhughes:
Updated to what? and why?

Please try and keep to one change per PR and explain what the change is you're making and why you want to make it.

AntoninCurtit:
As you can see, there's just 2 main changes.
1. As I said, I have updated the Facebook logo, who was obsolet.
2. As I said, in OpenSearch I've updated OpenStreetMap image (who was obsolet) and I have updated the attribution (who, it seems, was obsolet too).

To me the changes were already simple and clear enough, plus, with of quantity of changes, 2 PR...
Anyway I let you review you changes, it will be pretty faster than reading this. 😉

tomhughes:
Well strangely I couldn't tell from comparing two strings of base64 what change you had made...

AntoninCurtit:
That's why I wrote "Updated OpenSearch image"... 🙃

tomhughes:
Yes and I know what the opensearch image is, but as far as I can see the one on osm.org looks about right.

Look when I've got time I'll pull your patch down and see if I can compare the images to see what the difference is since you obviously don't want to make life easy by posting them here.

AntoninCurtit:
Good lord...
Well, that's the image from osm.org.
image


The new one comes OR from the 16x16 hosted on osm.org homepage OR from a SVG → PNG (16x16) I've made, initially obtained from OSM Wiki.

_I mean, it's always like this in this repo ? Just 2 lines of simples changes requires an eternity..._ 🙈

Issue #2882, opened 2020-10-13
Map note icon "X" too scary to click on


jidanni:
(We) users are nervous about clicking on anything with an "X" in the fear that it clicking on it will "destroy" it (even before being able to "open it to read its details.")
Therefore consider redesigning the "map note" icon,  
![X3](https://user-images.githubusercontent.com/1230959/95870092-b1911000-0d9e-11eb-98f7-fdf4763278ea.jpg)
Perhaps simply use the same icon (minus the "+") that is already used in the side panel,
![X4](https://user-images.githubusercontent.com/1230959/95870658-56135200-0d9f-11eb-88dc-905262a4b69a.jpg)
I mean let's face it, 999 out of 1000 "X"s on webpages are for button that close/destroy/remove things.
Sure, we can "remember that this one vs. the other 999 is safe, don't worry"... but it still gives us goose bumps.
(E.g., #2850 is one of the 999.)

gravitystorm:
I also dislike the X, but not for the same reason. I find it hard to intuitively remember whether the red X means that there is an issue, or the issue has been resolved (since red X is often used to mean closed).

I propose changing the note to a yellow exclamation mark, and keeping the resolved note as a green tick.

ScriptForAll:
Personally, I don't feel like that when I came across with an open issue note. That's simply due to the fact that when see something red with a cross my brain perceives it as "something's wrong". 
However, I do agree that using an yellow exclamation mark would make it more clear although it conflicts with the iD editor's adding a new note colour which is also yellow, as shown below:

![Screenshot_2020-10-30 OpenStreetMap](https://user-images.githubusercontent.com/36014883/97724678-206fa800-1ac5-11eb-9d7c-13f5476f73ff.png)

GoodClover:
If we're changing the note icons I suggest also making them SVG.
The current icons are low-quality and appear badly compressed (at least for me).

HolgerJeromin:
> And maybe a bit off topic, but it might be worth having different types of icons for different "types" of Notes. 

ref for tags on notes #5294

kmpoppe:
This is already fixed: https://github.com/openstreetmap/openstreetmap-website/issues/2882#issuecomment-731780021

I tried designing new markers - with the existing ones as comparison:
<img width="907" height="245" alt="Image" src="https://github.com/user-attachments/assets/291520c4-5768-4d03-81d4-377ede026ed0" />

A PR should be relatively easy:
[...]

PR #2946, opened 2020-11-09
Changed open_note_marker from red 'X' to yellow '!'

nlaratta:
Addresses issue #2882 
Before: 
open_note_marker@2x
After:
open_note_marker@2x


gravitystorm (changes requested):
Looks good! Can I ask for a few small changes:

* There is a file NoteMarker.svg in the same directory, which contains the SVG sources for the note markers. Could you please add your new marker to that file too, that way it's easier in future if more changes need to be made.
* Have you tried with a slightly more vibrant colour? I found this particular yellow hard to see when I tested it locally. The markers are translucent so the map background shines through - this means that markers need to be strongly coloured to really stand out.
* The .png files here are much much bigger (10x) than the originals. It looks like photoshop has added large amounts of metadata to the files. If you can remove that, great, but if not we'll do that ourselves. (e.g. `pngcrush -rem text open_note_marker.png test.png` reduces from 19170 to 2274 bytes, original was 1883).

gravitystorm:
image

dieterdreist:
sent from a phone

> On 11. Nov 2020, at 15:11, Andy Allan <notifications@github.com> wrote:
> 
> ﻿
> 
> 

the shadow is missing, it will make it easier to spot with a drop shadow

matkoniecz:
> Have you tried with a slightly more vibrant colour? I found this particular yellow hard to see when I tested it locally.

Also, it is a bit sad/dead/depressing/dirty. I would support more vibrant also for that reason.

mnalis:
@nlaratta are you still interested on working on this PR to change Note marker for OSM?

mmd-osm:
https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/images/NoteMarker.svg hasn't been updated as requested by @gravitystorm - also no user feedback since two years.

Closing this PR seems inevitable.

mmd-osm:
Closing now, as suggested in previous comment.

PR #2955, opened 2020-11-11
Prefer keyword arguments when method has optional boolean arguments

gravitystorm:
This PR fixes some cases where we are using optional boolean parameters, to use keyword arguments instead.

 I think this is particularly useful where the method definition lies elsewhere, since it's not obvious what an isolated 'true' or 'false' means.

https://docs.rubocop.org/rubocop/1.0/cops_style.html#styleoptionalbooleanparameter

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -185,9 +185,6 @@ Style/OptionalBooleanParameter:
     - 'app/models/trace.rb'
     - 'app/models/tracepoint.rb'
     - 'app/models/way.rb'
-    - 'test/models/diary_entry_test.rb'
-    - 'test/models/trace_test.rb'
-    - 'test/models/tracetag_test.rb'
 
 # Offense count: 28
 # Cop supports --auto-correct.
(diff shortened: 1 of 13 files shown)


tomhughes:
Looks like you missed something:

```
Error:
DiaryEntryTest#test_diary_entry_validations:
ArgumentError: wrong number of arguments (given 2, expected 1)
    test/models/diary_entry_test.rb:51:in `diary_entry_valid'
    test/models/diary_entry_test.rb:13:in `test_diary_entry_validations'


rails test test/models/diary_entry_test.rb:9
```

tomhughes:
Maybe that test failure wasn't down to this as travis has passed and I've just seen that failure in a checkout without this merged.... Guess there me some non-deterministic failure :-(

tomhughes:
No it is from this branch - my checkout was messed up.

tomhughes:
Ah and you had fixed it anyway.

gravitystorm:
> Ah and you had fixed it anyway.

Sorry! I force-pushed, and then I was waiting to double-check that travis passed before commenting, but I got distracted and forgot to come back.

PR #2959, opened 2020-11-13
Add table borders for legibility in richtext diary posts

migurski:
Introduce new table border treatment for richtext areas in diary entries. In diary entries with tables like [*Calculating label points with PostGIS*](https://www.openstreetmap.org/user/pnorman/diary/394676), [*2020 Editor Usage*](https://www.openstreetmap.org/user/Glassman/diary/394740), and [*Querying OpenStreetMap Changesets with Amazon Athena*](https://www.openstreetmap.org/user/Jennings%20Anderson/diary/394762) table elements appear squashed and hard to read.

This change slightly expands padding within table cell elements and adds thin, light-gray lines between rows and cells to make wider and more complex tables easier to read. Screenshots below are from Firefox 82.

## Current Site

Table elements are pressed together and rows & columns can be difficult to see easily.

Screen Shot 2020-11-13 at 3 59 31 PM

## This Change

Table elements are spaced further apart and rows & columns are easy to follow visually.

Screen Shot 2020-11-13 at 4 00 03 PM

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1965,53 +1965,6 @@ div.secondary-actions {
     margin: 0;
     color: $darkgrey;
   }
-
-  /* Default table appearance: no background color, thin gray borders, narrow padding */
-
-  table tr td,
-  table tr th
-  {
-    border: none;
-    border-top: 1px solid #ccc;
-    border-left: 1px solid #ccc;
-    padding: .25em .5em;
-  }
-
-  /* Use thicker borders at start of tbody and tfoot table sections */
-
-  table tbody tr:first-child td,
-  table tfoot tr:first-child td
-  {
-    border-top-width: 3px;
-  }
-
-  /* Remove all outside table borders */
-
-  table *:first-child tr:first-child td,
-  table *:first-child tr:first-child th
-  {
-    border-top: none;
-    padding-top: 0;
-  }
-
-  table * tr td:first-child,
-  table * tr th:first-child
-  {
-    border-left: none;
-    padding-left: .25em;
-  }
-
-  table *:last-child tr:last-child td,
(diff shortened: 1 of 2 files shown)


gravitystorm (changes requested):
Thanks for the PR. I'm happy to see more styling being added to tables within diary entries and other places where we have our "richtext" content.

However, we need to change the approach used. We've moved to Bootstrap for most of our styling, and so we should not be adding more custom styling rules to common.scss any longer. Bootstrap gives us various table styling options (https://getbootstrap.com/docs/4.5/content/tables/) out of the box, and so we should use one of them.

Since bootstrap tables are opt-in, i.e. require a class to be added to the `<table>` tag, then that will need to be implemented as part of the RichText.to_html process (in [`lib/rich_text.rb`](https://github.com/openstreetmap/openstreetmap-website/blob/4e78f3ae5b0d96f599840e181c1a4be4335dded9/lib/rich_text.rb))

migurski:
Thanks Andy. I’ll see if that ERB change is something I can make.

(Aside: I’ve mostly found Ruby really hard to deal with from a systems/installation perspective. I’ve been able to make and test small changes like this due to the Docker support in #2409 that @jalessio implemented. I’ve been keeping that PR updated but it’s been in limbo for over a year. I would love to see it merged and supported in order to allow other non-Ruby developers like myself to propose small changes)

migurski:
I’ve backed out the custom CSS and applied Bootstrap table styles instead:

Screen Shot 2020-11-18 at 9 39 36 PM

tomhughes:
Thanks for the idea and the proof of concept @migurski but I'm not hugely keen on editing the generated HTML so I've opened #2971 as an alternative implementation that works by overriding the kramdown converter to add the classes as the HTML is generated.

PR #2960, opened 2020-11-14
Add updated authentication logos

mxdanger:
Made changes to the authentication logos on the sign in screen:

* Replace previous outdated authentication logos with new SVG logos.
* Add border-radius for consistent shape between all logos.

Side by side difference between the SVG icons and the old PNG icons:
Windows File Explorer screenshot
Windows File Explorer screenshot

This pull request needs testing.

### DIFF
--- a/app/models/oauth_nonce.rb
+++ b/app/models/oauth_nonce.rb
@@ -2,7 +2,7 @@
 #
 # Table name: oauth_nonces
 #
-#  id         :integer          not null, primary key
+#  id         :bigint           not null, primary key
 #  nonce      :string
 #  timestamp  :integer
 #  created_at :datetime
--- a/db/migrate/20201214144017_expand_nonce_id.rb
+++ b/db/migrate/20201214144017_expand_nonce_id.rb
@@ -1,0 +1,7 @@
+class ExpandNonceId < ActiveRecord::Migration[6.0]
+  def change
+    safety_assured do
+      change_column :oauth_nonces, :id, :bigint
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


tomhughes:
Is going SVG only safe from a compatibility point of view? Everywhere where we have used SVG to date we have done so using a `picture` element with a PNG fallback.

mxdanger:
> Is going SVG only safe from a compatibility point of view? Everywhere where we have used SVG to date we have done so using a `picture` element with a PNG fallback.

You raise a good point, however, what exactly is the minimum browser target for the website so we can get an idea? Depending on what we need to support, is there really any reason to keep using png fallback?

Every major browser dating back ten or so years supports SVG and all of the xml data in the minified svg files.

mxdanger:
> Is going SVG only safe from a compatibility point of view? Everywhere where we have used SVG to date we have done so using a `picture` element with a PNG fallback.

Going SVG only is very safe from a compatibility point of view. Browsers support for SVG is [99.82% on desktop and 100% on mobile](https://caniuse.com/svg). There's just no practical reason to need the fallback method as that would mean needing to maintain both the vector and raster image.

mxdanger:
All that's left to do is wait for someone to test this branch and provide feedback on any potential issues.

HolgerJeromin:
We should/could drop the `<picture>` elements in the future. Not really needed nowadays anymore.

gravitystorm:
Merged in #3009, with just a small update to avoid using custom CSS and to use a bootstrap utility class instead.

PR #2970, opened 2020-11-18
Refactor trace notifications to reduce lego translations

### LINKED ISSUE #2268: Remove lego translations for trace notification emails
Moving https://trac.openstreetmap.org/ticket/5032 here for visibility.

The trace notification emails are made from 'lego' translations, that is fragments of sentences stuck together with other data in between each fragment. You can get the idea of how they are hard to translate from the [relevant part of en.yml](https://github.com/openstreetmap/openstreetmap-website/blob/31edc7dff6f4305366d0f93d35ba3fb1956bb71f/config/locales/en.yml#L1149-L1163):

```
    gpx_notification:
      greeting: "Hi,"
      your_gpx_file: "It looks like your GPX file"
      with_description: "with the description"
      and_the_tags: "and the following tags:"
      and_no_tags: "and no tags."
      failure:
        subject: "[OpenStreetMap] GPX Import failure"
        failed_to_import: "failed to import. Here is the error:"
        more_info_1: "More information about GPX import failures and how to avoid"
        more_info_2: "them can be found at:"
        import_failures_url: "https://wiki.openstreetmap.org/wiki/GPX_Import_Failures"
      success:
        subject: "[OpenStreetMap] GPX Import success"
        loaded_successfully: loaded successfully with %{trace_points} out of a possible %{possible_points} points.
```
They should be refactored to use full sentences with the additional data in string interpolation variables e.g. `%{data}`. This allows translators to rearrange the word order, if necessary, as well as making life easier than when trying to translate fragments in isolation (e.g. 'them can be found at:'!).

gravitystorm:
Fixes #2268.

Unfortunately there's still a lego translation when both halves of the sentence are joined together, but this at least reduces the number of combinations to translate.

### DIFF
--- a/app/views/user_mailer/_gpx_description.html.erb
+++ b/app/views/user_mailer/_gpx_description.html.erb
@@ -1,7 +1,7 @@
 <% trace_name = tag.strong(@trace_name) %>
 <% trace_description = tag.em(@trace_description) %>
 <% if @trace_tags.length > 0 %>
-  <% tags = @trace_tags.map { tag.tag.rstrip.join(" ") } %>
+  <% tags = @trace_tags.map(&:tag).join(" ") %>
   <%= t ".description_with_tags_html", :trace_name => trace_name, :trace_description => trace_description, :tags => tags %>
 <% else %>
   <%= t ".description_with_no_tags_html", :trace_name => trace_name, :trace_description => trace_description %>
--- a/app/views/user_mailer/_gpx_description.html.erb
+++ b/app/views/user_mailer/_gpx_description.html.erb
@@ -1,12 +1,8 @@
-<%= t "user_mailer.gpx_notification.your_gpx_file" %>
-<strong><%= @trace_name %></strong>
-<%= t "user_mailer.gpx_notification.with_description" %>
-<em><%= @trace_description %></em>
-<% if @trace_tags.length>0 %>
-  <%= t "user_mailer.gpx_notification.and_the_tags" %>
-  <em><% @trace_tags.each do |tag| %>
-    <%= tag.tag.rstrip %>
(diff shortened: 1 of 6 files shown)


app/views/user_mailer/_gpx_description.html.erb
```diff
@@ -1,12 +1,8 @@
...
-<em><%= @trace_description %></em>
-<% if @trace_tags.length>0 %>
-  <%= t "user_mailer.gpx_notification.and_the_tags" %>
-  <em><% @trace_tags.each do |tag| %>
-    <%= tag.tag.rstrip %>
-  <% end %></em>
+<% trace_name = tag.strong(@trace_name) %>
+<% trace_description = tag.em(@trace_description) %>
+<% if @trace_tags.length > 0 %>
+  <% tags = @trace_tags.map { tag.tag.rstrip.join(" ") } %>
```
tomhughes, line 4: Did you mean to put that `join` after the `map` rather than inside it? Also should it be `safe_join` or is interpolating it into an `_html` resource going to have much the same effect>

gravitystorm: Two good catches; the join should be after the map, and I'll look into the `safe_join` aspect too. And maybe add some tests :smile:

gravitystorm: > and I'll look into the `safe_join` aspect too

Actually, I don't think we want the safe_join here, since `tag.tag` isn't made up of existing safe strings (e.g. already-built html) - `tag.tag` is just a regular string. So `join` is the best approach and the whole string will be properly escaped when interpolated into the erb.

Of course, `join` is still in the wrong place!

PR #2971, opened 2020-11-19
Add bootstrap classes to markdown tables

tomhughes:
This is an alternative implementation of #2959 that overrides the kramdown converter to add the classes instead of trying edit the generated HTML with a search and replace.


tomhughes:
Actually I think an even better solution (now pushed) is to get the sanitizer to add the classes - that has an easy way to add classes without overriding any user ones and will work for all types of rich text not just markdown.

gravitystorm (approved):
I like the approach, and I've also checked out the visual appearance so I'm happy with the proposed html classes too.

Issue #222, opened 2013-03-29
Glitch in browse way page


sabas:
There's a little problem on lower width viewports, at least in the italian version, see http://www.openstreetmap.org/browse/way/25984170 with a viewport of less than 900px
The Comment part goes under the changeset number (in italian the h4 is "Nel gruppo di modifiche" which spans 2 rows).

A fix would be
.browse-section div {
clear: left;
}

(of course if the class is used only there, I tried via chrome dev tools )

PR #2993, opened 2020-12-02
Mailer refactoring

gravitystorm:
This PR contains some translation-related refactoring for the user_mailer.

It took me two attempts to get this right, since there's some subtleties about how the `_html` translations interact with plain text mail templates, and so I had to revisit and refactor some (unsubmitted) work that was done last week when I realised what was going on. But I've added a test that would have caught my mistake earlier.

In short, `_html` translations take the translation options (e.g. something like `t(".foo", :username => "Jack & Jill")`, html_escapes the options, interpolates them into the translation, and returns the whole thing as a SafeBuffer to avoid further html_escaping in the template. Putting that straight into an html template is fine, since it a) is already a safebuffer and b) is already escaped. But for plain text templates, we don't want any html_escaping of the input options, so we have to avoid using `_html` translation keys. Hence the duplication between the `_html` and non-`_html` keys.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -83,7 +83,7 @@ class UserMailer < ApplicationMailer
 
       mail :from => from_address(message.sender.display_name, "m", message.id, message.digest),
            :to => message.recipient.email,
-           :subject => I18n.t("user_mailer.message_notification.subject_header", :subject => message.title)
+           :subject => t(".subject", :message_title => message.title)
     end
   end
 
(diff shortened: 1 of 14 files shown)


woodpeck:
There seems to be a small regression here. Since 06 December, when I get emails through the web site, the initial text ("so-and-so has sent you a message through OpenStreetMap") is HTML escaped, so that the anchor/link around the user name how appears in all its a target="_blank" rel="noopener" style="text-decoration: none; color: #222" glory directly in the message text!

gravitystorm:
> when I get emails through the web site, the initial text ("so-and-so has sent you a message through OpenStreetMap") is HTML escaped

argh, I'll look into this.

Issue #3000, opened 2020-12-09
Alternative to Travis


gravitystorm:
I've become aware that our Travis CI configuration will stop working soon. This is because they are switching off travis-ci.org (which we use) on December 31st.

https://travis-ci.community/t/extremely-poor-official-communication-of-the-org-shutdown/10568

https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/

Theoretically, we can move to travis-ci.com as they offer open source credits. However, it seems that there are [some limitations with their own ability to hand out credits](https://twitter.com/james_hilliard/status/1336081776691843072), and I'm not sure it's worth the hassle. I have no idea where this repo stands with regards to a non-recurring 10,000 minute quota, since I can't find anything useful on travis-ci.org, but I suspect we'd be having to ask frequently, and I wouldn't want to just run out one day and not be able to get more.

So I think we have a few choices:

* Switch to travis-ci.com and hope they give us enough credits
* Switch to using github actions
* Switch to another CI system, for example [Gitlab CI](https://about.gitlab.com/solutions/github/), without moving the repo

tomhughes:
I do already have a branch somewhere that uses github action...

tomhughes:
Yes https://github.com/tomhughes/openstreetmap-website/tree/actions is the branch though it needs bringing up to date I expect.

migurski:
Would it be an option to pay Travis commercial rates?

Glad you have a GH Actions branch in your back pocket, Tom!

gravitystorm:
Fixed by #3002

PR #3002, opened 2020-12-10
Switch CI to github actions

tomhughes:
This is a candidate solution for resolving #3000 by switching the CI to use github actions.

I've created two workflows, one to run the lint tasks and one to run the tests.

We probably want to consider what combination of platforms we want to run the tests on - currently I've got it running on 18.04 and 20.04 with both ruby 2.5 and 2.7.


gravitystorm:
This is the first time I've dealt with github actions, so I'm not sure I'm in the best place to review! But from what I see it looks good to me, and I'm happy to use it to replace travis. Some notes:

* I was surprised to see the coveralls gem removed, but I guess whatever functionality is required is included in their action
* I don't understand the restore_key stuff for the caching, and I can't find any explanation for how restore_keys with dangling hyphens matches (prefix matching? Ordered by time or alphabetically?). However, it shows the same thing in the documentation, so :shrug: 
* I haven't found any examples of splitting up the jobs into a build stage followed by multiple parallel linting stages, using the system generated by the build stage. It seems a waste to have every job duplicating the bundle install in parallel. Perhaps there is a trick involving a build stage to install and cache, with the expectation that the cache will then be usually available for every lint stage. But I guess this is only an edge case, since they only duplicate each other on the first run after a Gemfile.lock change, so is not a big problem anyway.
* I'm also used to being able to run CI from a custom image, rather than a fixed menu of operating systems (my experience is with Gitlab CI). So I normally set up a separate repo to make custom images containing all the `apt install` prerequisites, which tend to change infrequently. That makes all the CI runs a bit quicker. But my skim of the github actions doesn't make it obvious how to do this, and in any case it looks like the github actions version of ubuntu has a bunch of stuff already installed.

So overall, a big thumbs up from me! I'm happy to see this merged if you're happy that it's ready.

tomhughes:
I suspect the answer to your last point involves running a container on top of the base OS which can be done, and that could likely help with the third as well if you built a test container on top of the base OS container. That's mostly speculation though. I'd have been happy just being able to avoid some of the duplication in the config...

Issue #2927, opened 2020-11-01
Use Labels for Issues on this Repo


lectrician1:
I'm having a hard time going through this repo's issues regarding ones that deal with changesets (there are a lot). I notice that most issues do not have labels. Could the maintainers please label the New and Open issues?

Quincy in the iD repo labels Issues very well. I can easily find what I'm looking for. 

There is already a good selection of labels created. I think more specific labels could be created as well (there seems to be a lot of unique feature requests).

tomhughes:
This does not appear to be reporting an issue in the software contained in this repository, or an issue with any definitive end point.

lectrician1:
I would have contacted you Tom through another way about this request, but I couldn't find a other contact method so I decided to just make this an issue. 
I understand that this has no definitive end point, but can my request for a commitment to labeling be guaranteed in any way?

tomhughes:
No of course it can't - you're asking volunteers to do work. We have no way to force anybody to do anything so no way to guarantee anything.

gravitystorm:
> I'm having a hard time going through this repo's issues regarding ones that deal with changesets (there are a lot). I notice that most issues do not have labels.

Hi Seth - Out of interest, what would labelling issues like this help you with? I'm curious as to why you are looking for issues related to changesets, and I want to know more about what kind of situation you are trying to solve.

I do label issues and pull requests from time to time, but generally with either high-level labels (like 'dx' or 'changes requested') and generally not with a label for the particular feature that it corresponds to.

tomhughes:
I have to say I'm curious about that as well - thinking about this more overnight my main inclination would actually be to get rid of many of our labels, which are mostly historic and have unclear meaning, and keep only a few like `support` and `changes request` that have a clear meaning and add others as needed and as we have an objective meaning for them.

But I am interested in how labels are perceived to help contributors as I've always thought of them mostly as a tool for maintainers to use to divide up issues - that might sometimes include a label that is used to publish a list (like `good first issue`) to contributors but it's not immediately obvious that they are more generally useful to contributors.

tuckerrc:
Labels for the specific components in the application such as diaries, notes, map, admin/moderation, and changesets would help contributors find things they would be comfortable working on. When I learn a new codebase I usually pick a specific area (such as "changesets") and branch out from there.

[...]

PR #3011, opened 2020-12-18
Add labels documentation to CONTRIBUTING.md

tuckerrc:
This pull request includes some new documentation in the CONTRIBUTING file. Any suggestions for new labels or better ways to organize them are welcome. I'm not sure if this is the best place to discuss new labels but it's a start for what I had in mind. 

(I pulled some of the documentation from the iD repository).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,7 +152,7 @@ Developer Experience. The things that help with development (Docker, Vagrant, Ra
 Work in Progress.  Don't start work on these, somebody else already did!
 
 [good-first-issue]: http://labl.es/svg?text=good%20first%20issue&bgcolor=0e8a16
-[help-wanted]: http://labl.es/svg?text=help%20wanted&bgcolor=0e8a16
+[help-wanted]: http://labl.es/svg?text=help%20wanted&bgcolor=1D76DB
 [awaiting_upstream]: http://labl.es/svg?text=awaiting%20upstream&bgcolor=E0D05C
 [enhancement]: http://labl.es/svg?text=enhancment&bgcolor=E105D8
 [changes_requested]: http://labl.es/svg?text=changes%20requested&bgcolor=5319E7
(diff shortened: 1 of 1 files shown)


tuckerrc:
Changes in response to #2927

tomhughes:
As labels can't be set by reporters I'm not sure it's helpful to document them, or at least not here.

tomhughes:
Also I wouldn't just document what we have because that's almost certainly not a good set of labels - it's probably better to tidy them up first and then document them.

The current set probably needs cutting down and we certainly need to decide if multi word labels are going to use space, underscore or hyphen to join words...

lectrician1:
Definitely agree. Could the labels be given descriptions within GitHub itself as well please?

tuckerrc:
Reporters may not be able to add labels but someone looking for an issue to work on may find this documentation helpful. I am okay with the documentation being on the [labels page](https://github.com/openstreetmap/openstreetmap-website/labels) or the CONTRIBUTING.md file. 

A few benefits of putting the documentation here is @lectrician1 and I can make suggestions that only require accepting Pull Requests and anyone looking to contribute will find them when reading CONTRIBUTING.md.

tomhughes:
My main point was questioning whether it should be in the contributor documentation rather than whether it should be documented at all - basically whether there should be separate documentation for maintainers.

tuckerrc:
I think having both can be useful. A simplified one for contributors and a detailed one for maintainers to use when doing triage.

gravitystorm:
@tuckerrc I want to clarify the first tag I've noticed. 'work-in-progress' (and [WIP] in subject lines of pull requests) means that the author knows that their own proposed code is not yet ready to merge (maybe the tests fail, or something else needs to be finished first), but the author of the code is (likely) sharing it in order to gain early feedback. It doesn't mean 'Don't start work on these, somebody else already did!'

CONTRIBUTING.md
```diff
@@ -108,6 +108,65 @@ other commits, namely:
...
+* <sub>[![awaiting upstream][awaiting_upstream]][awaiting_upstream_link]</sub> -
+Awaiting Upstream. These issues won't be resolved until changes are made upstream.
+* <sub>[![enhancement][enhancement]][enhancement_link]</sub> -
+Enhancements. These are features that improve already supported .
+* <sub>[![changes requested][changes_requested]][changes_requested_link]</sub> -
+Changes Requested. There are more changes needed before a pull request will be accepted.
+* <sub>[![dx][dx]][dx_link]</sub> -
+Developer Experience. The things that help with development (Docker, Vagrant, Rake, etc.)
+* <sub>[![work-in-progress][wip]][wip_link]</sub> -
+Work in Progress.  Don't start work on these, somebody else already did!
```
pnorman, line 152: Draft PRs should be used for this.

gravitystorm:
I think this has been superceeded by the work in #3043 where we discussed the current + proposed labels, and also added descriptions in the Github user interface for all of the labels.

I think having the descriptions there are sufficient, and we don't need to duplicate the descriptions in the contributing file - they are likely to get out of sync, and it is also additional reading that new contributors don't really need to worry about so I don't want the whole document to become (more) intimidating.

Thanks for opening the PR though - I think we ended up with an even better result via implementing #3043.

tuckerrc:
> I think we ended up with an even better result via implementing #3043.

I agree. Thanks for adding the new labels!

PR #3017, opened 2020-12-24
Refactor Diary Entry forms to use a custom form builder

gravitystorm:
This PR refactors the Diary Entry forms to use bootstrap, including using a custom form builder to create a `f.richtext_field` form component that does everything we need it to.

I've kept all the javascript the same, and the relevant ids the same, and the previews etc worked for me. The new form fields are built on the bootstrap form layout, so it's responsive as the screen width changes.

I did consider making a form component for the latitude/longitude/use_map stuff, since we also have lat/lon inputs elsewhere (e.g. user home location). But I left that aside for now, since having a form control related to two fields is a bit tricky, and the javascript and exact behaviour differs between diary entries and user home locations anyway.

This should unlock the conversion of most of our forms to bootstrap.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1281,8 +1281,6 @@ tr.turn:hover {
 
 .diary_entries {
   #map {
-    position: relative;
-    width: 90%;
     height: 400px;
     border: 1px solid $grey;
     display: none;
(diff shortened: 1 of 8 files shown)


tomhughes:
Excellent! That unblocks the other form conversions as well ;-)

PR #3018, opened 2020-12-27
Break long URLs in blog post show

tordans:
This uses https://getbootstrap.com/docs/4.4/utilities/text/#word-break to break long URLs on the blog post show view.

Example case: https://www.openstreetmap.org/user/tordans/diary/395215

Screenshot before:

Bildschirmfoto 2020-12-27 um 14 16 15 Kopie

Screenshot after:

Bildschirmfoto 2020-12-27 um 14 20 40 Kopie

An alternativ solution would be this custom CSS:

```
.diary_post .richtext {
    margin-top: $lineheight;
    // Prevent long URLs from breaking out of the container
    // https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/
    overflow-wrap: break-word;
    word-wrap: break-word;
    word-break: break-word;  
  }
```

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -24,7 +24,7 @@ module ApplicationHelper
     tag.div(:id => "#{id}_container", :class => "richtext_container") do
       output_buffer << tag.div(:id => "#{id}_content", :class => "richtext_content") do
         output_buffer << text_area(object_name, method, options.merge("data-preview-url" => preview_url(:type => type)))
-        output_buffer << tag.div("", :id => "#{id}_preview", :class => "richtext_preview richtext")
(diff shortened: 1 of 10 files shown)


tordans:
JFYI, I noticed that I committed only one case where .richtext was used; I updated the PR to add the helper class to all cases. IMO there is no downside to this since all those places could have too long URLs

gravitystorm:
> JFYI, I noticed that I committed only one case where .richtext was used; I updated the PR to add the helper class to all cases.

Excellent, thank you! This is exactly what I would look for in a PR like this - figuring out the correct bootstrap class to us (in preference to custom CSS) and also applying the same solution to all similar situations elsewhere in the codebase.

PR #3019, opened 2020-12-27
Add button styles to edit-diary-entry-button

tordans:
This adds a button style and also a bit of spacing.

Before:

Bildschirmfoto 2020-12-27 um 14 26 21 Kopie

After:

Bildschirmfoto 2020-12-27 um 14 26 11 Kopie

Disclaimer: I did not run the code since my setup is not operational ATM.

The style looks different from the comments-buttons (example https://www.openstreetmap.org/user/tordans/diary/395215) but it uses the default bootstrap style so I guess we are moving everything to this style step by step, right?

### DIFF
--- a/app/views/diary_entries/_diary_entry.html.erb
+++ b/app/views/diary_entries/_diary_entry.html.erb
@@ -23,7 +23,7 @@
     <%= render :partial => "location", :object => diary_entry %>
   <% end %>
 
-  <ul class='secondary-actions clearfix'>
+  <ul class='secondary-actions clearfix pt-3'>
     <% if params[:action] == 'index' %>
       <li><%= link_to t(".comment_link"), diary_entry_path(diary_entry.user, diary_entry, :anchor => "newcomment") %></li>
       <li><%= link_to t(".reply_link"), new_message_path(diary_entry.user, :message => { :title => "Re: #{diary_entry.title}" }) %></li>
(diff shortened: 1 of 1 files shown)


tordans:
The tests are failing at 

```
Failure:
DiaryEntriesControllerTest#test_edit [/home/runner/work/openstreetmap-website/openstreetmap-website/test/controllers/diary_entries_controller_test.rb:321]:
Expected exactly 1 element matching "a[href='/user/User%202002/diary/16/edit']", found 0..
Expected: 1
  Actual: 0

rails test test/controllers/diary_entries_controller_test.rb:250
```

I wasn't able to get the dev env running again, yet, so I cannot look into it ATM.

mmd-osm:
I haven't tested it, but I'm suspecting that ,` :class => "btn btn-primary mt-3" ` is added as an additional URL parameter to the hyperlink, which is probably not what you're expecting. The test fails as it is no longer matching the expected hyperlink.

tomhughes:
Yes, if you want HTML options those will need to be in a separate hash.

tordans:
I am not sure I needed this journey into the history of Rails URL pattern :-D, but I hope the current version will work.

mmd-osm:
Technically speaking this seems to be working ok now. However, the list of diaries UX seems a bit inconsistent to me now:

pr3019

gravitystorm:
> Technically speaking this appears to be working ok now. However, the list of diaries UX seems a bit inconsistent to me now:

Ha, I was just taking that same screenshot too!

gravitystorm (commented):
This seems like a straightforward PR, but unfortunately it opens a can of worms.

As @mmd-osm notes, this proposal breaks the diary list pages, which uses the same partial. I'm not sure what the best solution is here, since I'm happy to move from links to buttons to denote 'actions' (and trying to edit a diary entry seems like an 'action' to me). Perhaps one solution is to use two different partials, and make the diary list a "summary" of each entry (e.g. with a max word limit) rather than including the entire entry. Or alternatively, perhaps we keep the partial just for the summaries, and move the "full display" directly into the view.

Beyond that, one more minor point:
* The need for pt-3 and mt-3 is not clear. I think the pt-3 is because the location partial is "bare", i.e. text not wrapped in a paragraph, and so I would look into changing the html first, to avoid needing any explicit padding/margins.

tordans:
Thanks. Will look into it more once I have a Dev env again (and the time).

Issue #2344, opened 2019-08-07
Vagrant and Docker installtions are unmaintained


tomhughes:
Bother vagrant and docker installations suffer from constant bitrot where we add new dependencies or change configuration files but don't update the relevant setup scripts.

Unfortunately we have no way of detecting this until some unfortunate end user tries to use them because none of the main developers uses them.

We either need to find a way of testing them (is that even possible in travis? or is there some other way) or we need to admit defeat and remove them.

tomhughes:
So vagrant can't be done in travis apparently (https://github.com/travis-ci/travis-ci/issues/6060) so we can forget that at least.

openbrian:
I'd prefer to develop in a docker environment.  I can work on creating a new Dockerfile.  Do you have any requirements?  My plan is to create 2 images, one for db and one for rails.  For me this may have to wait until after SotM US.

mmd-osm:
@openbrian : see https://github.com/openstreetmap/openstreetmap-website/pull/2272

tomhughes:
Yes this ticket is probably a bit confusing because I opened it thinking we already had a docker config but apparently my memory was faulty... Though it's definitely true that it would be good to have it tested to stop it breaking. That's something that I think travis can do, unlike vagrant.

gravitystorm:
So I don't think we really need the vagrant/virtualbox bit tested, so much as the rest of the setup script i.e the contents of [/script/vagrant/setup/provision.sh](https://github.com/openstreetmap/openstreetmap-website/blob/master/script/vagrant/setup/provision.sh). Could that be tested on travis?

Of course, both the travis config and the vagrant setup (and the Docker setup) cover a lot of the same ground - installing postgres, bundle install, etc. So if there's a way to refactor things to share some code, that would be great too.

fazlerabbi37:
just find this [page](https://blog.fossasia.org/testing-docker-deployment-using-travis/) where is seems docker builds can be tested with travis. Though needs to take a closer look at this.

--- 

> I'd prefer to develop in a docker environment. I can work on creating a new Dockerfile. Do you have any requirements? My plan is to create 2 images, one for db and one for rails. For me this may have to wait until after SotM US.

@openbrian please give your feedback on #2272 as I am just a novice on docker. :blush:

jalessio:
For reference, just adding a note here that I'm attempting to address the Docker part of this in PR https://github.com/openstreetmap/openstreetmap-website/pull/2409

[...]

Issue #2972, opened 2020-11-19
Vagrantfile does not allocate enough memory to run tests


tuckerrc:
I kept getting OOM errors when I tried to run the tests after following the Vagrant provisioning instructions. I increased the memory to 4096 and the issue went away. 4GB might be more than necessary but 1GB was not enough in my testing.

PR #3020, opened 2020-12-27
VAGRANT.md: add a note about maintenance status

tordans:
This adds a note about maintenance status by linking to https://github.com/openstreetmap/openstreetmap-website/issues/2344

I just tried to use this setup but it failed. I then searched and found the reason for it in https://github.com/openstreetmap/openstreetmap-website/issues/2344.

This link on the page would have helped, so here it is …

### DIFF
--- a/VAGRANT.md
+++ b/VAGRANT.md
@@ -1,3 +1,7 @@
+# Currently unmaintaned 
+
+The Vagrant setup is currently unmaintaned. See https://github.com/openstreetmap/openstreetmap-website/issues/2344 for more.
+
 # Installing Vagrant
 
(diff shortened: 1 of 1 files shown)


mmd-osm:
Someone else was successful only a week ago after bumping the available memory, see #2972. Also, there's not much anyone can do to help you based on "I just tried to use this setup but it failed.". You need to be much more specific and provide detailed error messages. I guess you already knew that.

tomhughes:
Quite. That is not a "reason" for any particular failure, it's just noting that there is no routine validation that the vagrant configuration is working in any particular environment..

PR #3023, opened 2020-12-30
Update secondary actions for diary entries

tordans:
This is a second take on https://github.com/openstreetmap/openstreetmap-website/pull/3019.

# Next steps

- [ ] Feedback on the UI / UX changes; and changes in general – especially by @gravitystorm 
- [ ] I will then look at possible test failures (especially once I get them running locally :))

# What changed:

>**Diary entry: move hide-button in separate box …**
699ed66
>This is only for admins, so we can keep it in a separate box that stands out from other actions.
>- add box
>- add button style

>**Diary entry: Reorder and restyle secondary actions …**
6eb624b
>- remove custom class and use bootstrap default classes instead https://getbootstrap.com/docs/4.5/content/typography/#inline
>- add button link to edit action – small on #index, regular on #show
>- make edit action first item in list so it may look better as a button
>- use edit_diary_entry_path for edit action
>- place new-comment and comment-count links next to eachother by moving reply-link elsewhere
>- make reply-link visible on #show as well (was #index only, for some reason)
>- make reply-link visible to non-current-users only

> **Diary entry: Remove check for presence of current_user …**
0b5a2fa
>"if nil == diary_entry.user" is false, so no need the presence of current_user first.

>**Diary entry: Remove the xml:lang attribute …**
7b3233e
>https://www.w3.org/International/questions/qa-html-language-declarations says we dont need it for html5 documents. This was the only place that had it.

>**Diary entry: Spacing if location given …**
e4d66c0
>The location box did not have any spacing bottom, yet.


tordans:
A question that is a bit off topic: Having all those double quotes is by design, yes?

```
Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
In file: app/views/diary_entries/_diary_entry.html.erb:32

```

PR #3024, opened 2020-12-30
Diary entries #index: Migrate to bootstrap pagination

tordans:
Use https://getbootstrap.com/docs/4.5/components/pagination/ for the pagination.

Side note: I tried using link_to_if but this does not work well in this case since multiple changes need to happen for the disabled state.

The pagination css class cannot be removed, yet, since there is quite a bit more custom pagination in the codebase that will be a lot harder to migrate. Maybe for those, using a pagination gem that also uses bootstrap would be the better approach.

# Screenshots

Before

Bildschirmfoto 2020-12-30 um 14 20 37

After 

Bildschirmfoto 2020-12-30 um 14 20 45

### DIFF
--- a/app/views/diary_entries/index.html.erb
+++ b/app/views/diary_entries/index.html.erb
@@ -28,21 +28,29 @@
 
   <%= render @entries %>
 
-  <div class="pagination">
-    <% if @entries.size < @page_size -%>
-      <%= t(".older_entries") %>
-    <% else -%>
-      <%= link_to t(".older_entries"), @params.merge(:page => @page + 1) %>
-    <% end -%>
+  <nav>
+    <ul class='pagination'>
+      <% if @entries.size >= @page_size -%>
+        <li class='page-item'>
+          <%= link_to t('.older_entries'), @params.merge(:page => @page + 1), :class => 'page-link' %>
+        </li>
+      <% else -%>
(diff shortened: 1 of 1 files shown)


gravitystorm (commented):
Thanks! I merged this with a small fix (removing an unnecessary `link_to_if`).

The pagination looks much better, and using bootstrap markup and classes is exactly what I am looking for in these types of PR. Thanks again.

mmd-osm:
A quick comment on this change: _Diary_ and _GPS traces_ don't seem to agree on the position of "old" and "new". I think it would make sense to use the same pattern everywhere, but I don't know which one makes more sense. Maybe a topic for a new issue.

#### Diary
pages_diary

#### GPS Trace
pages_gps

tordans:
Yes, let's move this to another issue. I wanted to look into using a pagination gem like Kaminari (spelling?) as well which would allow for more feature with less code in the controller… - and also create consistency.

PS: those PRs are not forgotten by me, just paused; will update soon.

tomhughes:
If anybody is looking at reworking pagination in general then the most important thing is that we need to move away from pagination based on counting (using `page=N` with an OFFSET/LIMIT query) and to pagination based on ID values where the query selects the next N records before/after a given ID value.

Issue #2308, opened 2019-07-14
Improve GPS trace upload UX


SilentSpike:
I've uploaded [quite a few traces now](https://www.openstreetmap.org/user/TheEditor101/traces) and have a suggestion to make the process easier to understand (especially for new users) and more transparent.

Instead of the whole "identifiable, public, trackable, private" category system, you could have check-boxes to enable/disable the details which these categories represent.

Example:
  [ ] Publicly associate trace with your account
  [ ] Publicly serve trace with ordered timestamps

In this way, it's much more clear to the user (I think) which privacy controls they have. You could still use the resulting combination to gives traces a category if required elsewhere.

GoodClover:
I agree with this, the current system is a bit confusing.
I don't know how the backend works (and it's in RoR which I don't know), but I'm going to have a look at just clarifying the options in the GPS menus.

GoodClover:
I've been looking at some ways to just clarify the existing drop down, but there are inconsistencies between the wiki and the website on the meaning of `Public`.

The current upload page says that `Public` traces are "shown in the trace list", "anonymous" and "unordered points".
However the wiki says they are linked to you user, they are not anonymous:
![image](https://user-images.githubusercontent.com/27222523/103353595-296afe80-4aa1-11eb-94eb-a9cedf047160.png)
And then, down the page, it agrees with what the website says about them being anonymous:
> **Public** means [...] Data served via the API does not reference your trace page. [...]


Can anyone clarify the correct meaning?
Are they linked to your user or not?

GoodClover:
Ok I tested it and the website is wrong, the table on the wiki is correct.

`Public` traces are shared as ***identifiable***, unordered points. And are therefore shown in trace lists.

tomhughes:
I have rechecked the source code and as far as I can tell that is not correct - for points served via the trackpoints API only points from `identifiable` tracks are linked to a track.

tomhughes:
The code is https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/api/tracepoints_controller.rb#L63 and only tracks which are both `trackable?` and `identifiable?` will reach the code that adds a link to the track node in the XML response.

GoodClover:
Random recent `Public` trace:
https://www.openstreetmap.org/user/gabriele_sani/traces/3541218
It's definitely linked to it's user.
[...]

PR #3026, opened 2020-12-30
Correct defninition of `Public` trace and make the definitions clearer.

GoodClover:
I have made the definition of trace visibilities clearer and corrected the definition of a `Public` trace.
See #2308

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2057,10 +2057,10 @@ en:
           by dragging. Add your message, then click save, and other mappers will investigate.
   traces:
     visibility:
-      private: "Private (only shared as anonymous, unordered points)"
-      public: "Public (shown in trace list and as anonymous, unordered points)"
-      trackable: "Trackable (only shared as anonymous, ordered points with timestamps)"
-      identifiable: "Identifiable (shown in trace list and as identifiable, ordered points with timestamps)"
+      private: "Private (Anonymous Points)  -  Shared as anonymous, unordered points."
+      public: "Public (Points)  -  Shared as identifiable, unordered points. Shown in trace lists."
+      trackable: "Trackable (Anonymous)  -  Shared as anonymous, ordered points (lines) with timestamps."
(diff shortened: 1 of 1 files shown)


tomhughes:
As I have just explained in #2308 your conclusion appears to be wrong - if you think that the tracepoints API is exposing that information then I believe it's an critical bug that should be fixed rather than changing the documentation but you have not as yet provided any evidence to support your claim.

tomhughes:
I've pushed 786b8cce651d31afc9a3ae920f25f0d13b0314c0 now which extends the tests to verify that no identifying information is returned.

Firefishy:
Typo in PR: `Reccomended`

tomhughes:
It doesn't really matter any way - we shouldn't be marking any particular option as recommended. That's a decision for the user.

Marking public as deprecated might be reasonable but I don't think we should make any other value judgements here.

tomhughes:
There are also practical limits to how much text you can reasonably show in an option list drop down...

GoodClover:
Oh sorry about the spelling, I must have just mistyped.

I copied the bracketed things from what the wiki said about each option, it says it's the recommended option on there.

About the limits of text, maybe just remove the long explanation altogether and leave that for the user to go to the wiki?

tomhughes:
I think there is a link to the wiki isn't there? Though I'm not very keen on that because people can make changes that might not be accurate and it might not be noticed.

tomhughes:
Yes - there is a "what does this mean" link under the control that links to the wiki page.

GoodClover:
I think having it marked as the "Full" option would be good if it's not marked as recommended.
To a new person, if just the names were shown, it's confusing as to what they mean.

Should I remove the notes in brackets then and leave the long text?
Or what else should be done?

GoodClover:
As Tom said, the text is impractical; it barely fits for Identifiable.
image

Maybe showing descriptions separately on the page would be better?

That's beyond my ability though as RoR/`.erb` is not my thing.

Issue #2332, opened 2019-07-28
Dark Mode


fishcharlie:
The CSS prefers-color-scheme media query allows for users to use dark mode with websites. Although the [browser adoption](https://caniuse.com/#feat=prefers-color-scheme) is still low now, I assume it will jump quite a bit when Chrome 76 and iOS 13 are released later this year.

OpenStreetMap should adopt this new standard and allow for dark mode on the website.

tomhughes:
Are you volunteering to do it?

Only we don't really have anybody with CSS or design skills so it is unlikely to happen unless somebody steps up to do the work.

dieterdreist:
sent from a phone

> On 28. Jul 2019, at 07:04, Charlie Fish <notifications@github.com> wrote:
> 
> OpenStreetMap should adopt this new standard and allow for dark mode on the website.


particularly for the map this would require an alternative rendering style and additional tiles, right?

tomhughes:
No idea but I guess if you took it to an extreme then yes - that is not likely to happen until vector rendering is in use for sure!

fishcharlie:
> Are you volunteering to do it?

I could for sure look into it and see what I can do. I haven't really looked at this codebase before. But sure, I can volunteer some time to it, not sure how far I'll get tho. Also fair warning: I'm not the best at design and color choices either. But I know enough about CSS to at least get the ball rolling maybe.

---

In terms of vector rendering, I think that would be nice for this project. Is that something that is going to happen soon? Maybe this is something we could do in stages? Have dark mode for the majority of the site, and wait for vector rendering for enabling dark mode on the map itself?

---

If I were to work on this, is it something that has enough support to be accomplished? I just don't wanna spend a lot of time working on it, and then it's decided that it's a bad idea and shouldn't be implemented. I understand that the changes made have to be approved and all that. But from a high level perspective, is it something that has enough support to be worked on without it being a total waste?

mmd-osm:
Maybe you should try the following user script first: https://userstyles.org/styles/173774/openstreetmap-dark-theme and see how it goes (as reported in http://www.weeklyosm.eu/en/archives/12261/)

tomhughes:
Everybody wants vector rendering but so far nobody is doing the work to make it happen.

As to everything else, it doesn't interest me, but if it can be done without creating a massive maintenance headache then I have no objection to it.

[...]

PR #3028, opened 2021-01-01
Introduce Dark Mode (WIP)

migurski:
This PR offers an alternative approach to browser dark mode, noted by @fishcharlie in #2332 and initially implemented by @bezdna in #2532. Before I spend more than a few hours on this, I’m seeking feedback on an approach that accommodates @gravitystorm’s notes on the prior PR:

> … having lots of overrides and !important flags and so on is fine when a user style is the only available option. But that's not a good approach when we have full control over the CSS sent to the user. … To implement a dark mode, I would ideally like to see out-of-the-box support from bootstrap itself. If that's not available, then maybe 10-20 lines of code to set some colour variables, but little more than that. – https://github.com/openstreetmap/openstreetmap-website/pull/2532#issuecomment-588163551

The alternative implementation here defines `*-color` and `*-invcolor` parameters as suggested though they’re somewhat duplicative because `common.scss` had a large number of bare color literals. I migrated all of these to the parameters file without attempting to interpret their meaning. To support dynamic color mode switching these are referenced in media queries near the default definition of each class. Example:

```
body {
  font-family: 'Helvetica Neue',Arial,sans-serif;
  font-size: $typeheight;
  line-height: 1.6666;
  color: $common-body-color;
  background-color: $common-body-background-color;
  margin: 0px;
  padding: 0px;
  text-align: left;
  height: 100%;
}

@media (prefers-color-scheme:dark)
{
  body {
    color: $common-body-invcolor;
    background-color: $common-body-background-invcolor;
  }
}
```

I’m also interested in help from anyone who’d like to collaborate on this PR to make it suitable for use! Some sample items left to be done are:

- [ ] Invert or adjust map colors per @pkrasicki’s suggestion – https://github.com/openstreetmap/openstreetmap-website/issues/2332#issuecomment-727266980
- [ ] Figure out why `.dropdown-menu` is not yet working
- [ ] Decide on visual approach for map toolbar
- [ ] Update custom CSS on pages like `/help` and `/about`

Here’s a short video showing how this implementation of dark mode behaves.

https://user-images.githubusercontent.com/58730/103447801-daf66500-4c44-11eb-9d9d-b175c9ec046d.mov


pkrasicki:
darkmode
You shouldn't use black color for backgrounds. Use dark gray, it will be easier to read the page.

You will need to make those green buttons lighter and link color (both purple and blue) is too dark as well.

Links in top right need to be a little lighter too. Perhaps use white color for current page.

pkrasicki:
Instead of creating additional variables and adding `invcolor` to their names I think you should just override default variables in @media(prefers-color-scheme: dark), if that's possible to do in SASS.

migurski:
That was the first thing I tried, @pkrasicki! It doesn’t work because the variable definitions are interpreted at compile time while the media queries need to happen at render time. There’s a different potential approach using CSS variables which may interfere with the [5% of users who can’t see them](https://caniuse.com/css-variables); the [@media query has less support](https://caniuse.com/prefers-color-scheme) but it also defaults to existing behavior.

I’ll apply your suggestion of lightening everything slightly.

gravitystorm:
When I start to review this and see "Large diffs are not rendered by default" and parameters.css getting to 150 lines, I don't think that fits my definition of "maybe 10-20 lines of code to set some colour variables, but little more than that." :smile:

This approach has a couple of major drawbacks. The first is that you have changed the name of colour variables like `$blue` to `$blue-color`, but bootstrap uses some of these colour variables too! This is how the primary button colour on the bootstrap forms matches our house blue.

The second is that you are having to invent an entire colour scheme from scratch, as can be seen with the talk about black backgrounds and coloured buttons. I'd prefer to offload those design choices elsewhere, if possible. For example, to an existing "bootstrap-dark" colour scheme where someone else has evaluated what the primary button colours should be, or just how different should the greys used in the background vs `btn btn-light` in order to meet accessibility guidelines. I'm really not keen on us making that up from scratch, but if it needs to be done then it needs to be done for all of the colours used in bootstrap. Have a look at https://github.com/twbs/bootstrap-rubygem/blob/4.5-stable/assets/stylesheets/bootstrap/_variables.scss to see what variables are available.

On a more minor point, I don't like the 'inv' naming since they aren't inverted - I don't think we'll be highlighting danger messages in cyan, for example!

You are right to pick out that there are many colours defined in custom.css that need addressing. However, I'd prefer to see them sorted out first in one (or more) separate PRs. For example, by using the bootstrap variables instead of specific RGB values, or even better by refactoring the code to use bootstrap instead of having any custom CSS in the first place. This will at least bring some structure, and reduce the maintenance burden for supporting dark mode later on.

As for the exact mechanism for how to support dark mode - I can't assess if this is the best option. But if you need to define `@media (prefers-color-scheme:dark)` over and over again, then I don't think it's maintainable. I also don't see how this changes the colour of anything (like form buttons or dropdown menus) where the colours are defined by bootstrap.

Refs https://github.com/twbs/bootstrap/issues/27514

migurski:
Okay, thanks Andy! I don't think there’s a way to implement dark mode in a way that respects browser settings without a minimal amount of `@media` querying, since it’s a client-side consideration and not one that Bootstrap will currently support. I’m closing this WIP, and I appreciate your quick feedback.

pkrasicki:
Maybe detecting system theme is not the right way then? Instead you could have a button to toggle between default and dark mode. It would require a little bit of JavaScript, but you need it anyway to change map colors in Leaflet.

Since Bootstrap doesn't have a dark theme, is it possible to not use custom CSS?

tomhughes:
We are not going to have a button to select dark mode... That's just insane.

bhousel:
For anyone curious, here's how I implemented the insane dark mode slider on https://nsi.guide
https://github.com/osmlab/name-suggestion-index/commit/2f68fe032e2388e576b66ed4e55128f53ff46358

It's really just a set of css rules that take effect if `.dark` class is added to the document root.
You don't need to add all the media queries to the .css because JavaScript has a [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) function that can give you that answer.

This implementation respects the user's dark mode preference the first time, then uses whatever is stored in local storage every time after that.

pkrasicki:
@bhousel I did the same in https://issviewer.com.
dark-theme-toggle

Seems like a reasonable alternative to me.

Issue #2851, opened 2020-09-25
Missing cannonical URL for search engines on blog index pages


tordans:
Hi, unfortunately I don't have the time to fully debug this, but I will post my findings and maybe someone can pick it up.

**Issue:**

I was looking for this blog post https://www.openstreetmap.org/user/Supaplex030/diary/393423 via the Google Image Search https://www.google.com/search?q=supaplex+parkpl%C3%A4tze+openstreetmap&sxsrf=ALeKk02JDhRGUouiU8tP9ztjwoVIuWM0tg:1601064638585&source=lnms&tbm=isch&sa=X&ved=2ahUKEwi8ta3IjoXsAhWC_KQKHd9vBt8Q_AUoAXoECAwQAw&biw=1403&bih=798 which shows the image of the blogpost that I am looking for. 

But: The search links me to this page https://blogs.openstreetmap.org/?viewType=Print&viewClass=Print

I don't know what this page is, but its not the blogpost details page and it looks like some special case page that should not be in the index.

**Thoughts:**

As I said, I did not debug further, but IMO the ideal case would be:

1. make sure google only sees the show-, not the index-views:

  * IMO the best practice would be, to have the blog-index pages on `<meta name="robots" content="follow,noindex">`
  * Only the show-pages should be `<meta name="robots" content="follow,index">` (which is the default)

2. make sure google understands the relation between outputs that show the same content and only has the primary output in the index:

  * All non-primary-output formats, should have a `<link rel="canonical" href="https://example.com/kleider/gruene-kleider"/>` to the primary output format
https://support.google.com/webmasters/answer/139066?hl=de

Update: In this case, this seems to boil down to informing google that those params `?viewType=Print&viewClass=Print` should not be in the search result.

Maybe https://search.google.com/search-console can help understanding an debugging this.

Update: Maybe this can also be solved via this https://support.google.com/webmasters/answer/6080548?hl=de

tomhughes:
If you have a problem with what Google link their search results to maybe you should complain to Google.

At least if you're going to complain to us try and explain what it is you want us to change and why. You use a lot of words here which I'm sure mean something to you but are entirely meaningless to me, not that I have any great inclination to alter our output to panders to Google's latest desires anyway.

tomhughes:
I'm particularly unclear what the index pages have to do with anything since you presumably wanted it to take you to the detail page not the index page?

tomhughes:
[...]

PR #3029, opened 2021-01-02
Add noindex meta tag to diary index pages

tordans:
Google (and other search engines) should follow all links on this page, but only index the actual diary_entry#show and not the #index. This meta tag tells google to do just that (https://developers.google.com/search/docs/advanced/crawling/block-indexing?hl=en).

Part of https://github.com/openstreetmap/openstreetmap-website/issues/2851

### DIFF
--- a/app/views/diary_entries/index.html.erb
+++ b/app/views/diary_entries/index.html.erb
@@ -1,3 +1,4 @@
+<% content_for :head, tag(:meta, :name => :robots, :content => :noindex) %>
 <% content_for :heading do %>
   <div <% if @user %> id="userinformation"<% end %>>
     <% if @user %>


gravitystorm:
@tomhughes I'm happy with the concept, but not sure if we should do this via meta tags or by adding to `public/robots.txt`

tomhughes:
I'm not sure I have a view - it's no something I know anything about.

Is it even possible to do this via `robots.txt` though? Won't that block crawling altogether?

I'm not sure what any of this has to do with canonical URLs though?

Frankly I had to check to make sure we didn't have the diary indexes blocked already as I think we did at one point because the pagination makes crawling them insanely expensive... They don't seem to be blocked at the moment though.

tordans:
I always thought robots.txt an noindex-metatags would just be the same but different methods. However, I find the meta tag easier to understand and validate its correctness. However, reading on the google help page, it sounds like the robots.txt way is actually seen as something different by google and they recommend using the meta-tag (or http-verbs).

So IMO this is the way to go …

https://developers.google.com/search/docs/advanced/robots/intro?hl=en
> You should not use robots.txt as a means to hide your web pages from Google Search results. This is because, if other pages point to your page with descriptive text, your page could still be indexed without visiting the page. **If you want to block your page from search results, use another method such as password protection or a noindex directive.**
> https://developers.google.com/search/reference/robots_meta_tag

mnalis:
I agree with @tordans that using meta tag is the right way to go here. Are there any blockers to pull this PR?

tomhughes:
Well I'm still none the wiser as to what the goal is - at no point has anybody explained **why** we don't want that page indexed.

mnalis:
@tomhughes  Basically, you want search engines to index the detailed page (like for example [this page](https://www.openstreetmap.org/user/Hans%20Thompson/diary/396516)) and not index page (like for example [this page](https://www.openstreetmap.org/user/Hans%20Thompson/diary)).  The `noindex` implemented in this PR is a hint to search engine that "_this is a duplicate page that should not be indexed, as real page with same or better data is available elsewhere_".

Main reason why would you want that behavior is that index page will often differ between the time they where indexed by search engine, and the time users visit them (eg. user will publish new diary entries which would push older entries to second subpage). 
Net result of that fact is that **users searching for some term on search engine will often land on index page that no longer contains the search term they searched for**, which is annoying to users (and we presumably want to make users happy). Hinting to search engines that they should not index such index pages solves the problem, as only detailed pages get cached then.

There are also other reasons why detail ("leaf") page (instead of index page) should always be preferred:
- index page might contain shortened text instead of full text
- detailed page might get downranked if index page is also indexed by search engine
- having users pick which of the several very similar versions of page they want to visit is wasting their time
- having same content appear on multiple pages might make the whole site look more spammy (especially when some users do put spam content in diaries!)
- it leads to less load on web servers and more up-to-date data in search engines (as with hint they only index portion of the pages, they can do faster, and do not need to reindex unchanged detail pages)
- etc.

More background is explained in linked https://github.com/openstreetmap/openstreetmap-website/issues/2851#issuecomment-761879964 and rest of that ticket, and link to several [SEO](https://en.wikipedia.org/wiki/Search_engine_optimization) pages which explain the issue in more detail.

So helping search engines decide what pages are important (and which are not) helps not only search engines, but also users trying to find OSM content, as well as OSM infrastructure load / bandwidth.

PR #3030, opened 2021-01-02
Add seed data for users, diary entries and iD

tordans:
This will seed a dev database with initial data.

To use this with the WIP Docker setup https://github.com/openstreetmap/openstreetmap-website/pull/2409, run

```
docker-compose run --rm web bash
rails db:seed
```

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,9 +1,156 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
 
 Language.load(Rails.root.join("config/languages.yml"))
+
+# Variables
+
+def lorem_ipsum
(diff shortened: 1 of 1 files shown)


tomhughes:
What's the purpose of creating random noise though? I might accept a way to create an initial user and the OAuth clients needed for the builtin oauth clients but I don't see any advantage to going beyond that.

tordans:
> … but I don't see any advantage to going beyond that.

While working on CSS updates and such I noticed I need blogposts and comments by different users that allow me to see long list (for pagination) and different cases of text length and user permission. Instead of logging in as separate users manually to create the content I used those seed.

In this spirit this PR is not finished; I would create a separate PR for additional seed data once I need them for other parts of the app.

Having the seed create some of each objects in different configurations on a Dev setup sounds like a good and helpful idea to me. Using the seed data is optional, after all.

gravitystorm:
We should have a way to separate out two concepts:

* The first is seed data, as defined in most of the documentation that I can find, meaning data that you will want in the database on any/all deployments. The usual example is a list of states or countries, and in our context that would include the languages table I guess. I would maybe put an iD editor registration in there, but not development data.
* The second is "development data", like this PR covers. The kind of stuff that shouldn't appear in a production database. I'm wary of putting this in seeds.rb without at least an `environment` guard.

For the development data, I would like to see this myself! Adding stuff to an empty database is no fun for new developers. But:

* It should be idempotent. When you add the next thing (e.g. traces) and I re-run it, I don't want to see more and more diary entries every time I run the script.
* I would consider using FactoryBot to generate the data, as we use for the tests, along with something like Faker to generate the boilerplate.
* We should consider making this a separate rake task, to isolate it from the db:seed functionality entirely.

I wrote all that, and then I found these posts which probably explain it better than me!

* https://medium.com/@henrytabimagiraldo_41792/stop-using-the-seed-file-to-populate-your-dev-data-in-rails-bb241130c7da
* https://thoughtbot.com/blog/priming-the-pump

gravitystorm:
Although it's gone quiet on this topic recently, I'm still in favour of the approach I used in #3101 as an alternative to this PR. So I'm closing this in favour of #3101 and we can continue discussion (if anything is needed) over there.

Issue #3031, opened 2021-01-02
Sync more hash-url-params for iD


tordans:
This is the corresponding issue to https://github.com/openstreetmap/iD/issues/8114 in the iD Repo. This the solution will – as far as I understand – be part of this repos code, the topic should (also) be discussed here IMO.

I looked into this issue and will create a draft-PR after this Issue.

Short summary of the issue:

iD has 14 url params (+ the location/zoom params) that allow to configure the editor via URL.
Those params work fine to initialise iD on osm.org but get removed right after the editor loaded.
(This is https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/javascripts/edit/id.js.erb#L23-L36)
Changes in the editor config are not represented in the hash URL on osm.org.
It is therefore not possible to share the configuration of an iD setup (apart from location+zoom) via URL.

Ideal solution:

The iD Dev-preview shows a different behaviour: Changes to the editor config are represented in the hash-URL.
Ideally, this behaviour is also possible for OSM.org (where we need to work around the iframe-setup).

tomhughes:
If changes to the iD integration are needed then the iD authors normally provide them as part of the PR when they do a new release.

tomhughes:
As a general point we don't expect people to deep link to iD on osm.org beyond simply copying the standard edit link - we don't really want to document or support secret URL parameters that people have to add manually because that's not a good user experience.

tordans:
> …we don't expect people to deep link to iD on osm.org beyond simply copying the standard edit link - we don't really want to document or support secret URL parameters that people have to add manually because that's not a good user experience.

I agree and this is what this issue is about. Right now, the URL parameter are not documented and need do be added manually. They only work to initialise the editor but then disappear from the URL.

Once fixed, a user can use iD without knowing anything about the URLs since it updates by itself and represents a shareable configuration of what the user sees at this point (eg. which map layer is active, which mapillary filter is active and such).

gravitystorm:
My concern with this is that "[cool URIs don't change](https://www.w3.org/Provider/Style/URI)". We've supported directly linking to the default editor for years, and providing a way to link to the editor at a certain position is something that is both useful, and will continue to make sense as we change editors (P1 -> P2 -> iD -> ???). 

[...]

PR #3033, opened 2021-01-02
Sync more hash-url-params for iD

tordans:
This commit includes a few console.logs since it is still WIP.

It solved part of the issue described in https://github.com/openstreetmap/openstreetmap-website/issues/3031:

Whenever I pan the map in iD, the osm-hash-params get updated with the current iD-in-iframe-configuration.
However, click actions in the iD iframe (eg changing the map layer) don't trigger this update until I also pan the map afterwards.

This is all I could contribute with my limited knowledge of the code. I am pretty sure there is a more elegant solution to this. However, I also think this would already be a bit better than what we have right now …

### DIFF
--- a/app/assets/javascripts/edit/id.js.erb
+++ b/app/assets/javascripts/edit/id.js.erb
@@ -20,6 +20,8 @@ $(document).ready(function () {
       params.map = (mapParams.zoom || 17) + '/' + mapParams.lat + '/' + mapParams.lon;
     }
 
+    console.table(hashParams);
+
     if (hashParams.background) params.background = hashParams.background;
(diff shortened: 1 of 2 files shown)


tordans:
Lets close this draft for now, it is too old to be useful.

I am still hoping for some improvements here in the future … like

- one set of url params `?`/`&` that are the same for all/most editors and part of the URL and synced by iD
- all the other editor state params as part of the URL hash `#`

That should allow for the website to have a clear and shared URL search params API … but still editors can have custom state to make the UI sharable.

There was some work on syncing the editor state with the URL in https://github.com/facebook/Rapid/issues/1404

PR #3034, opened 2021-01-02
Traces#index: Introduce tab navigation, fix tag-filter

tordans:
- introduce bootstrap tabs to switch the views
- introduce `content_for :heading_class` to remove the padding below the bootstrap tabs
- update the rss-image to use a svg, adopted from https://icons.getbootstrap.com/icons/rss/ (without the outer border)
- move rss- and new-button away from the view-switching actions
- the `@tag` logic was broken. introduce new link to remove the tag-filter; the tabs keep the filter once given
- use `&.` syntax nil-safety so we can remove `@display_name`

I will wait for a first review of this.

Still to do:

- [ ] Double Check the usage of `params[:tag]` in the views. Did I introduce a security problem here? If so, what is the smartest way to solve it? – Will need to look into this; but maybe someone already has the answer …
- [ ] Check specs
- [ ] Add translations for the two new strings (hint about how to this is welcome – just add it to en.yml and the other translations will come automatically?)

The remove-tag-feature is not ideal IMO, but on the other hand it was broken ATM and nobody noticed so it was probably not that important and IMO not worth adding even more complexity.

## Screenshots
**gps – logged-in – someone-else**
gps--logged-in--someone-else

**gps – logged-in – all**
gps--logged-in--all

**gps – logged-in – my**
gps--logged-in--my

**gps – logged-out – someone-else**
gps--logged-out--someone-else

**gps – logged-out – all**
gps--logged-out--all

**gps – logged-in – someone-else – tag filter active**
gps--logged-in--someone-else--tag

**gps – logged-in – all – tag filter active**
gps--logged-in--all--tag

**gps – logged-in – my – tag filter active**
gps--logged-in--my--tag

**gps – logged-out – someone-else – tag filter active**
gps--logged-out--someone-else--tag

**gps – logged-out – all – tag filter active**
gps--logged-out--all--tag

### DIFF
--- a/app/controllers/traces_controller.rb
+++ b/app/controllers/traces_controller.rb
@@ -67,7 +67,6 @@ class TracesController < ApplicationController
 
     # final helper vars for view
     @target_user = target_user
-    @display_name = target_user.display_name if target_user
   end
 
   def mine
--- a/app/views/layouts/_content.html.erb
+++ b/app/views/layouts/_content.html.erb
@@ -5,7 +5,7 @@
     <%= render :partial => "layouts/flash", :locals => { :flash => flash } %>
     <% if content_for? :heading %>
       <div class="content-heading">
-        <div class="content-inner">
+        <div class="content-inner <%= yield :heading_class %>">
           <%= yield :heading %>
         </div>
       </div>
--- a/app/views/traces/index.html.erb
+++ b/app/views/traces/index.html.erb
@@ -1,25 +1,61 @@
+<% content_for :heading_class, 'pb-0' %>
 <% content_for :heading do %>
   <h1><%= @title %></h1>
-  <ul class='secondary-actions clearfix'>
-    <li><%= t(".description") %></li>
-    <li><%= rss_link_to :action => "georss", :display_name => @display_name, :tag => @tag %></li>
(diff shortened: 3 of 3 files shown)


gravitystorm:
> * (hint about how to this is welcome – just add it to en.yml and the other translations will come automatically?)

Yes :smile:

gravitystorm:
>     * Double Check the usage of `params[:tag]` in the views.

From a quick review, I don't see anything obviously wrong. Rails is very good about avoiding XSS and similar problems, so long as you avoid using any of the utilities that mess around with string safety (like `raw`, `.html_safe`, etc).  I can double-check when the PR is ready for review.

gravitystorm:
I've marked this as draft due to the test failures.

gravitystorm:
I've included the patch from @migurski that got the tests working, along with some changes from myself (instead of adding them here as review notes) in #3226 . 

Therefore I'm closing this as superceded by #3226

Thanks for the initial work @tordans , I think the tabs look great and I can see us using them in other situations too!

migurski:
Nice to see this make it in!

tordans:
Yes, thanks for finishing this!

PR #3035, opened 2021-01-02
Traces: Migrate pagination to bootstrap

tordans:
- Change pagination to use bootstrap default styles
- … but with empty span-tags instead of non-active a-tags
- refactor partial to not use @-variables but get the variables via
the render-call, simplify render-call-syntax
- remove the additional `if @traces.size > 1` inside the pagination
partial since that caused the pagination to disappear on the last page.
The partial will be visible with inactive links.
- Add bootstrap spacer-classes to overwrite commons-css so removing. Eg.
commons also has a pagination class same as bootstrap.

### DIFF
--- a/app/views/traces/_trace_paging_nav.html.erb
+++ b/app/views/traces/_trace_paging_nav.html.erb
@@ -1,18 +1,27 @@
-<p>
+<nav>
+  <ul class="pagination my-3 py-0">
+    <% if page > 1 %>
+      <li class="page-item">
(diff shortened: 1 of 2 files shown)


gravitystorm:
In general I would avoid the bootstrap spacer classes wherever possible. Since they are used here to "fight against" the CSS rule in common.css, my approach would be to rename that class to avoid clashing with bootstrap. But as far as I can tell we only use `.pagination` in two places, both of which you have refactored already, so I've removed it, along with the spacer classes added in this PR.

Refactoring to avoid `@variables` in partials is good to see too.

Thank you!

PR #3036, opened 2021-01-03
Traces: Migrate list to bootstrap

tordans:
Changes:

- switch from tables to bootstrap media since it makes the spacing around the image easier and also its not a table in a semantic way
- remove the unless in #index since it is covered by the if-size>0 above

refactor the list
- reorder elements
- make the pending-image a blank square and move the pending notice to the first line
- move the actions to the right and use the previous title-text as link-text since those explain a lot better what the links actually do (I thought "edit" would edit the trace-db-entry, not open the map-editor)
- remove the two translations that are not used anymore from en.yml
- use if-present for tags-link-list since its easier to read
- use the same styling for the trace description as on the changeset description (italic text)

I will wait for a first review of this.

Still to do:

- [ ]  Check specs
- [ ]  Add translations for the two new strings (hint about how to this is welcome)

### Screenshots

BEFORE
traces-before

AFTER 

traces-after

AFTER / PENDING
traces-pending

### DIFF
--- a/app/views/traces/_trace.html.erb
+++ b/app/views/traces/_trace.html.erb
@@ -1,35 +1,47 @@
-<tr>
-  <td>
-    <% if Settings.status != "gpx_offline" %>
+<li class="media mb-1 p-2 <%= cycle('', ' bg-light') %>">
+  <% if Settings.status != "gpx_offline" %>
+    <% if trace.inserted %>
+      <a href="<%= url_for :controller => "traces", :action => "show", :id => trace.id, :display_name => trace.user.display_name %>">
+        <img src="<%= url_for :controller => "traces", :action => "icon", :id => trace.id, :display_name => trace.user.display_name %>" alt="" class="img-thumbnail mr-3" />
+      </a>
+    <% else %>
+      <div style="width: 60px; height: 60px" class="img-thumbnail bg-white mr-3"></div>
+    <% end %>
+  <% end %>
+
+  <div class="media-body">
+    <% if trace.inserted %>
+      <p class="float-right">
+        <%= link_to t(".view_map"), { :controller => "site", :action => "index", :mlat => trace.latitude, :mlon => trace.longitude, :anchor => "map=14/#{trace.latitude}/#{trace.longitude}" }, { :class => "mr-2 border-right pr-2" } %>
+        <%= link_to t(".edit_map"), { :controller => "site", :action => "edit", :gpx => trace.id } %>
+      </p>
+    <% end %>
+
+    <p class="mb-0">
+      <strong class="mr-2">
+        <%= link_to trace.name, :controller => "traces", :action => "show", :display_name => trace.user.display_name, :id => trace.id %>
+      </strong>
       <% if trace.inserted %>
-        <a href="<%= url_for :controller => "traces", :action => "show", :id => trace.id, :display_name => trace.user.display_name %>"><img src="<%= url_for :controller => "traces", :action => "icon", :id => trace.id, :display_name => trace.user.display_name %>" border="0" alt="" /></a>
+        <%= t ".count_points", :count => trace.size %>
       <% else %>
-        <span class="text-danger"><%= t ".pending" %></span>
+        <span class="text-warning"><%= t ".pending" %></span>
       <% end %>
-    <% end %>
-  </td>
-  <td><%= link_to trace.name, :controller => "traces", :action => "show", :display_name => trace.user.display_name, :id => trace.id %>
-    <span class="text-muted" title="<%= trace.timestamp %>"> ...
-      <% if trace.inserted %>
-        (<%= t ".count_points", :count => trace.size %>)
-      <% end %>
-      ... <%= time_ago_in_words(trace.timestamp, :scope => :'datetime.distance_in_words_ago') %></span>
-      <%= link_to_if trace.inserted?, t(".map"), { :controller => "site", :action => "index", :mlat => trace.latitude, :mlon => trace.longitude, :anchor => "map=14/#{trace.latitude}/#{trace.longitude}" }, { :title => t(".view_map") } %> /
-      <%= link_to t(".edit"), { :controller => "site", :action => "edit", :gpx => trace.id }, { :title => t(".edit_map") } %>
-
       <% badge_class = case trace.visibility
-                       when "public", "identifiable" then "success"
-                       else "danger"
-                       end %>
-      <span class="badge badge-<%= badge_class %> text-white"><%= t(".#{trace.visibility}") %></span>
+                        when "public", "identifiable" then "success"
+                        else "danger"
+                        end %>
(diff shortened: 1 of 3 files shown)


app/views/traces/_trace.html.erb
```diff
@@ -1,35 +1,44 @@
...
-                       else "danger"
-                       end %>
-      <span class="badge badge-<%= badge_class %> text-white"><%= t(".#{trace.visibility}") %></span>
+    <p class="mb-0 text-muted" title="<%= trace.timestamp %>">
+      <%= time_ago_in_words(trace.timestamp, :scope => :'datetime.distance_in_words_ago') %>
+      <%= t ".by" %> <%= link_to trace.user.display_name, user_path(trace.user) %>
+      <% if trace.tags.present? %>
+        <%= t ".in" %>
+        <%= safe_join(trace.tags.collect { |tag| link_to_tag tag.tag }, ", ") %>
+      <% end %>
```
mmd-osm, line 39: This bit here looks like a Lego-style translation, as it translates the text word by word. Obviously I can't tell if that's causing some issue for any of the ~97 languages we support.

Typically you would define one locale string that includes multiple variables and let the translators figure out what the proper text would be in their language.

tordans: I try not to touch the output – translations and things like this join – in all my PRs since my focus is on refactoring the html/css to bootstrap and while doing this moving the elements to a – in my opinion – better place in the view. I hope this will make it easier to merge since I expect less discussion about the right wording and less re-translating needs to be done.

If we where to change this specific translation, I would vote for changing the meaning of "in" to something like "categoriesed with" or "tagged with". The "in" term does not make much sense IMO, since the tags are not the main "folder" where the traces are "in", nor are the tags the main navigation element on the traces page.

tomhughes: Yes that's a separate issue and should be addressed (if there is indeed a problem) in a separate PR.

app/views/traces/_trace.html.erb
```diff
@@ -1,35 +1,44 @@
...
+    <% if trace.inserted %>
+      <a href="<%= url_for :controller => "traces", :action => "show", :id => trace.id, :display_name => trace.user.display_name %>">
+        <img src="<%= url_for :controller => "traces", :action => "icon", :id => trace.id, :display_name => trace.user.display_name %>" alt="" class="img-thumbnail mr-3" />
+      </a>
+    <% else %>
+      <div style="width: 60px; height: 60px" class="img-thumbnail bg-white mr-3"></div>
+    <% end %>
+  <% end %>
+
+  <div class="media-body">
```
gravitystorm, line 12: The media component has been removed from bootstrap 5. See https://github.com/twbs/bootstrap/pull/28265. So to avoid having to refactor this soon, we should avoid using it now.

app/views/traces/_trace.html.erb
```diff
@@ -1,35 +1,44 @@
-<tr>
-  <td>
-    <% if Settings.status != "gpx_offline" %>
-      <% if trace.inserted %>
-        <a href="<%= url_for :controller => "traces", :action => "show", :id => trace.id, :display_name => trace.user.display_name %>"><img src="<%= url_for :controller => "traces", :action => "icon", :id => trace.id, :display_name => trace.user.display_name %>" border="0" alt="" /></a>
-      <% else %>
-        <span class="text-danger"><%= t ".pending" %></span>
-      <% end %>
+<li class="media mb-1 p-2 <%= cycle("", " bg-light") %>">
```
gravitystorm, line 1: I count 10 different uses of margin and padding utilities - seems a bit high? If there is any way to reduce them, that would make the layout more robust to future changes. In particular, I can't see any system behind choosing different sizes e.g. `-1`, `-2` etc.

gravitystorm, line 1: We recently moved away from `cycle` in templates, although this is less intrusive than what we had previously. See https://github.com/openstreetmap/openstreetmap-website/commit/5fdada204cf92d8ca99300927713724aaaf1b2bd for the change.

This proposes changing the colour for the stripes, which makes it inconsistent with the other striped tables/lists e.g. messages.

Is there a different way to achieve this? Or alternatively, does writing more CSS to support attempting to stripe a list suggest that we should be sticking with a table?

app/views/traces/_trace.html.erb
```diff
@@ -1,35 +1,44 @@
...
+  <% if Settings.status != "gpx_offline" %>
+    <% if trace.inserted %>
+      <a href="<%= url_for :controller => "traces", :action => "show", :id => trace.id, :display_name => trace.user.display_name %>">
+        <img src="<%= url_for :controller => "traces", :action => "icon", :id => trace.id, :display_name => trace.user.display_name %>" alt="" class="img-thumbnail mr-3" />
+      </a>
+    <% else %>
+      <div style="width: 60px; height: 60px" class="img-thumbnail bg-white mr-3"></div>
```
gravitystorm, line 8: I don't think our CSP allows inline styling like this. But if we're making empty divs a specific width and height, to push other content around, then we should be looking at using a grid or table instead.

app/views/traces/_trace.html.erb
```diff
@@ -1,35 +1,44 @@
...
+    <% end %>
+  <% end %>
+
+  <div class="media-body">
+    <% if trace.inserted %>
+      <p class="float-right">
```
gravitystorm, line 14: I'm not sure if floating is the right approach here, but I haven't investigated alternatives like grid.

app/views/traces/_trace.html.erb
```diff
@@ -1,35 +1,44 @@
...
       <% if trace.inserted %>
-        (<%= t ".count_points", :count => trace.size %>)
+        <%= t ".count_points", :count => trace.size %>
+      <% else %>
+        <span class="text-warning"><%= t ".pending" %></span>
       <% end %>
-      ... <%= time_ago_in_words(trace.timestamp, :scope => :'datetime.distance_in_words_ago') %></span>
-      <%= link_to_if trace.inserted?, t(".map"), { :controller => "site", :action => "index", :mlat => trace.latitude, :mlon => trace.longitude, :anchor => "map=14/#{trace.latitude}/#{trace.longitude}" }, { :title => t(".view_map") } %> /
-      <%= link_to t(".edit"), { :controller => "site", :action => "edit", :gpx => trace.id }, { :title => t(".edit_map") } %>
+      <% badge_class = trace.visibility.in?("public", "identifiable") ? "success" : "danger" %>
```
gravitystorm, line 29: This needs to be `in?(Array)` i.e. `trace.visibility.in?(["public", "identifiable"])` (note added `[ ]`)

gravitystorm (changes requested):
Overall I like parts of this PR, e.g. removing the `...` and making the link text clearer. But I'm not convinced by the list vs tables - not yet, but still a possibility!

PR #3062, opened 2021-01-20
Adding banner for 2021-01 community survey

dorotheakazazi:
Suggesting banner for the current OSM community survey, created by the OSMF board and accessible at https://osmf.limequery.org/281662 until 14 February 2021. Announcement at https://blog.openstreetmap.org/2021/01/18/community-survey/


tomhughes:
I don't see how anybody would have any clue that image is inviting them to answer a survey...

dorotheakazazi:
Hi Tom :)

I didn't want to have the text "survey" on the banner, as that would be readable only by people who read English. Tried to convey that it is a survey with the image, but maybe it doesn't work :)

A solution could be to have the word "survey" in all the languages used in the survey (or at least in many of them), but that would take a bit of time to check it with the translators and the board wanted something quickly. If someone else wants to have a go at it, feel free, I currently don't have the time (at least not today) :)

mikelmaron:
@tomhughes something like this pass muster?

OSM Foundation Survey (4)

let me know if generally ok, and if it needs any tweaks. if 👍  then i can create a fresh PR, or let me know if another way.

mikelmaron:
as discussed in Board chat with @dorotheakazazi it would be great to provide multilingual banners, but we've never done so before and best solution would be to localize. that's out of scope right now.

tomhughes:
Well that's certainly a lot better yes.

pnorman:
Have you tested that image to make sure it doesn't overlap with the x button?

A printed link seems unnecessary as the image itself is a link, isn't it?

mikelmaron:
Have not tested the image @pnorman. What is the positioning of the X button exactly?

Yes link is unnecessary, but guess the idea was to emphasize where you're going.

mikelmaron:
@tomhughes I put in a new PR here https://github.com/openstreetmap/openstreetmap-website/pull/3066

pnorman:
> Have not tested the image @pnorman. What is the positioning of the X button exactly?

[The top 60x60 pixels](https://github.com/openstreetmap/openstreetmap-website/blob/415a05f96ac962e27e674a5ef1b1ec46d4be8977/FAQ.md)

PR #3065, opened 2021-01-21
Avoid recommending installation of the Ubuntu bundler package

### LINKED ISSUE #3013: bundle exec rake yarn:install - Error
I try install openstreetmap-website.
In the following command:
bundle exec rake yarn:install

screen
Screenshot_1

I used recommendation tomhughes
sudo apt-get install npm
sudo npm install -g yarn
But the result remained the same

My system is hosted on VirtualBox Version 6.1.16 r140961 (Qt5.6.2) (windows)
Screenshot_2

Description:	Ubuntu 20.04.1 LTS
Release:	20.04

What am I doing wrong?

gravitystorm:
Due to a problem with namespacing and vendoring, the Ubuntu package leads to namespace conflicts with Thor 1+

Fixes #3013

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -28,7 +28,7 @@ These can be installed on Ubuntu 20.04 or later with:
 
 ```
 sudo apt-get update
-sudo apt-get install ruby2.7 libruby2.7 ruby2.7-dev bundler \
+sudo apt-get install ruby2.7 libruby2.7 ruby2.7-dev \
                      libmagickwand-dev libxml2-dev libxslt1-dev nodejs \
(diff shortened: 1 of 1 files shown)


tomhughes:
As you say in #3013 this isn't a guaranteed "fix" and I'm not sure it is even with gem installations of bundler - what would happen if bundler installed by gem needed a different version of thor to that required by our Gemfile?

It's not something I can get excited about though...

tomhughes:
Ah I see you have addressed that more general point in your upstream comments ;-)

PR #3068, opened 2021-01-24
Move banner close icon so we can see it

harry-wood:
We sometimes have banner graphics (like the current "community_survey" image) which uses space quite far into the top right. This can make the close "x" icon hard to see. Move it towards the corner where it is less likely to overlap with banner graphic text.

**Before and after:**

Screenshot 2021-01-24 at 00 58 53

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -394,14 +394,14 @@ body.compact-nav {
       position: absolute;
       top: 0;
       right: 0;
-      width: 60px;
-      height: 60px;
+      width: 26px;
+      height: 26px;
 
       .icon.close {
         pointer-events: none;
         position: absolute;
-        right: 20px;
-        top: 20px;
+        right: 3px;
+        top: 3px;
       }
     }
 


pnorman (changes requested):
If we're removing it then the reserved area stated in FAQ.md should be adjusted

harry-wood:
Ah yeah. The image is not really following those [FAQ.md](https://github.com/openstreetmap/openstreetmap-website/blob/master/FAQ.md) instructions hey?

I suggest we merge this as a quick fix, and revert when this we're done with this image. And we can always use the same commit again in future if we get an image with this problem. All a bit fiddly but...

One way or another we should avoid the current hidden "x" situation.

tomhughes:
I'm sorry but suggesting we keep this around and merge it every time somebody leaves it until the last minute to produce a banner and then doesn't have time to produce a working one is just ridiculous.

People that want banners for upcoming events need to start thinking about it a few weeks in advance, not a few hours, or a few days after the event has started - that way there will be time for us to help them produce a workable banner.

harry-wood:
Fair enough I suppose. So the right way to fix the hidden "x" situation would be to change the image hey?

mikelmaron:
I can fix the banner image later today in a few hours, so it reserves 60x60 corner space

however, why not move the “x” and reduce the reserved space permanently? Seems like a good idea to have more space for banner content

mikelmaron:
Ok another [PR here](https://github.com/openstreetmap/openstreetmap-website/pull/3069) with image that reserves 60x60 space for the "x"

I still think that @harry-wood fix here is a good one and should be accepted, after adding edit of the FAQ to the PR

gravitystorm:
> however, why not move the “x” and reduce the reserved space permanently? Seems like a good idea to have more space for banner content

But then it becomes inconsistent with the size and positioning of all the other X buttons on the sidebar, first-time visitor dialog, etc. We're doing a lot of work to remove the 2500+ lines of custom CSS that's full of inconsistencies like this! So I don't want to introduce more one-off positioning rules.

For accessibility reasons, we should move away from having banners with the text embossed into an image anyway, but that's a different topic.

dieterdreist:
with the current code it would work by changing the colors:
make the x sprite white and add a color to the background: #bbb
border-radius: 10px; 
opacity: 0.8
On hover the background would become darker, opacity go to 1.0 and the sprite could remain the same.

this would render a round button and be visible also in case the banner design doesn’t take care of leaving space for the close button

mikelmaron:
Ah need for consistency makes sense @gravitystorm 

Agree on accessibility. Imagine most of our banners have included text. One long term solution is to allow localization of the banners.

tomhughes:
Yes that's a nice idea in theory but in practice most banners are supplied at very short notice and the chances of doing any meaningful localisation of images in a short period of time are minimal.

PR #3074, opened 2021-01-27
Refactor account form to use bootstrap

gravitystorm:
I've made this a separate PR since it's our biggest and most complex form.

There's definitely room for further refactoring, particularly around the image radio boxes and their names and values, but I didn't want to do too much at the same time.

### DIFF
--- a/app/views/users/account.html.erb
+++ b/app/views/users/account.html.erb
@@ -21,7 +21,7 @@
   <fieldset class="form-group">
     <label for="user_auth_provider"><%= t(".external auth") %></label>
     <div class="form-row">
-      <%= f.select(:auth_provider, Auth.providers, :hide_label => "true", :wrapper => { :class => "col-auto mb-0" }) %>
+      <%= f.select(:auth_provider, Auth.providers, :hide_label => true, :wrapper => { :class => "col-auto mb-0" }) %>
       <%= f.text_field(:auth_uid, :hide_label => true, :wrapper => { :class => "col mb-0" }) %>
     </div>
     <small class="form-text text-muted">(<a href="<%= t ".openid.link" %>" target="_new"><%= t ".openid.link text" %></a>)</small>
--- a/app/views/users/account.html.erb
+++ b/app/views/users/account.html.erb
@@ -10,158 +10,96 @@
   </ul>
 <% end %>
 
-<%= error_messages_for current_user %>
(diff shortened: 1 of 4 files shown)


app/views/users/account.html.erb
```diff
@@ -10,166 +10,104 @@
...
+  <%= f.text_field :display_name %>
+  <%= f.email_field :email, :disabled => true, :label => t(".current email address") %>
+  <%= f.email_field :new_email, :autocomplete => :off %>
+  <%= f.password_field :pass_crypt, :value => "", :autocomplete => :off %>
+  <%= f.password_field :pass_crypt_confirmation, :value => "", :autocomplete => :off %>
+
+  <fieldset class="form-group">
+    <label for="user_auth_provider"><%= t(".external auth") %></label>
+    <div class="form-row">
+      <%= f.select(:auth_provider, Auth.providers, :hide_label => "true", :wrapper => { :class => "col-auto mb-0" }) %>
```
tordans, line 24:
```suggestion
      <%= f.select(:auth_provider, Auth.providers, :hide_label => true, :wrapper => { :class => "col-auto mb-0" }) %>
```

To keep in sync with one line below.

app/views/users/account.html.erb
```diff
@@ -10,166 +10,104 @@
...
+      <%= f.text_field(:auth_uid, :hide_label => true, :wrapper => { :class => "col mb-0" }) %>
```
tordans, line 25: Optionally: to make it more accessible you could add the sr-only class to the label and a descriptive label text.

tordans (commented):
Just two thinks I noticed. Great refactoring.

app/views/users/account.html.erb
gravitystorm: I've added better descriptive labels in 3cabed9 (although auth_uid is still quite a technical concept, and I'm not even sure it's clear what the box is for in non-sr situations, but that's not a focus of this refactoring).

The `:hide_label` feature of bootstrap_form does exactly what you suggest - the label is still in the html tree, just marked with the sr-only class, but otherwise there and linked to the relevant input field :+1:

gravitystorm:
> Just two thinks I noticed.

Thanks for the review! Let me know if you spot anything else.

app/views/users/account.html.erb
tordans: Ah, thanks, did not know this. Good feature.

PR #3078, opened 2021-02-02
Update to Omniauth 2.x

tomhughes:
This requires converting all use of the /auth endpoints to use the POST method as GET is no longer supported.

Unfortunately I had to change all the `follow_redirect!` calls in the tests as in rails 6.0 that always uses GET regardless of the response status - in rails 6.1 that has actually been fixed but we can't move to that yet.


gravitystorm:
Looks good to me - although I know so little about the omniauth stuff, I'm not sure I'm adding much beyond checking that the tests pass :smile:

Issue #3081, opened 2021-02-03
Address post-merge Docker Compose cleanup issues


migurski:
Tom H identifies a few desired changes in https://github.com/openstreetmap/openstreetmap-website/pull/2409#pullrequestreview-582662631:

- [ ] Investigate alternative to Postgres sleep
- [x] Update checkout to v2
- [ ] Investigate bootsnap-related pollution in Rails setup files
- [ ] Move DB functions to either `lib` or `share`
- [ ] Determine if a DB password is needed when using trust auth
- [ ] Investigate whether `yarnpkg install` can move under `bundle exec`

I’ll look at each of these in a follow up PR and determine which needs a followup change vs. just an explanation.

tomhughes:
The good news is that it all seems to work using `podman-compose` in place of `docker-compose` in Fedora 33.

mmd-osm:
Re " Move DB functions to either lib or share" : /changes endpoint was removed yesterday, we don't need those DB functions anymore.

tomhughes:
Also #3082 has bumped the checkout action version.

gravitystorm:
> Re " Move DB functions to either lib or share" : /changes endpoint was removed yesterday, we don't need those DB functions anymore.

I know you'll have seen my comment, but just for other people's reference, we still need the function for the migrations: https://github.com/openstreetmap/openstreetmap-website/pull/2409#discussion_r570115583

PR #3083, opened 2021-02-04
Post-merge Docker Compose cleanup issues

migurski:
Follow up to Tom’s review comments in https://github.com/openstreetmap/openstreetmap-website/pull/2409#pullrequestreview-582662631:

- Checkout was updated to v2 upstream, so no need to do that in this PR
- After testing under a variety of platforms, we haven’t been able to reproduce last year’s bug so ENABLE_BOOTSTRAP is now gone.
- Moved DB functions to `/usr/local/share/` since they’re still necessary per https://github.com/openstreetmap/openstreetmap-website/pull/2409#discussion_r570115583
- DB password and superuser privilege are both still needed when even when using trust auth because the database is needed for setup, migrations, normal usage, and data imports
- `yarnpkg install` is still required since `bundle exec` will not work without additional app code present during Docker build step. The old option here was to leave the yarn step for the end-user after app installation, but I think it’s better more of the build process to be automated within the Dockerfile

Closes https://github.com/openstreetmap/openstreetmap-website/issues/3081

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -71,6 +71,7 @@ You can now use Docker to load this extract into your local Docker-based OSM ins
     docker-compose run --rm web osmosis \
         -verbose    \
         --read-pbf district-of-columbia-latest.osm.pbf \
+        --log-progress \
         --write-apidb \
             host="db" \
             database="openstreetmap" \
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM ubuntu:20.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 
-# Install system packages
+# Install system packages then clean up to minimize image size
 RUN apt-get update \
  && apt-get install --no-install-recommends -y \
       build-essential \
@@ -30,7 +30,7 @@ RUN apt-get update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
-# Install current Osmosis
+# Install compatible Osmosis to help users import sample data in a new instance
 RUN curl -OL https://github.com/openstreetmap/osmosis/releases/download/0.47.2/osmosis-0.47.2.tgz \
  && tar -C /usr/local -xzf osmosis-0.47.2.tgz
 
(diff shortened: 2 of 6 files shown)


tomhughes:
Blimey I just realised what you're trying to do and why you can't use rake to do the yarn install - you're trying to cherry pick what files to map into the container!

Yes of course that won't work, but it's completely unsuppported to try and install the gems and node modules in this way.

The docker configuration needs to follow the normal installation instructions to build the image, not make up crazy stuff of it's own...

gravitystorm:
>     * since `bundle exec` will not work without additional app code

What app code is needed? I would have expected `bundle exec rake yarn:install` to work after bundle is installed and `bundle install` has added all the gems.

tomhughes:
Well the first thing that fails is that `Rakefile` is missing, then if you map that in it fails because all the rails configuration files are missing.

Basically it's working in an environment where only `package.json` and `yarn.lock` have been mapped in to the container.

tomhughes:
I even wonder where it's installing the node modules to? Normally they get installed locally to `node_modules` in the source tree but that isn't mapped into the container at the point of the install here so are they being installed as system global modules?

migurski:
Yes, that’s right: it wants the Rakefile, then it wants more. When I tried to resolve this chain, it was transforming the Dockerfile from a working install of the host’s local Ruby code for development purposes to a self-contained deployment container that wouldn't help developers make and preview changes. There might be a role for the latter but I can’t tell if it’s possible to support in the same container!

Node modules go into a non-global `node_modules` directory which is kept out of the host directory tree based on ~December feedback from Andy: https://github.com/openstreetmap/openstreetmap-website/blob/afce133d863a9d2f1070c095c1c7feadbf9df019/docker-compose.yml#L9-L11

tomhughes:
I can't reproduce the bootsnap issues, either using podman on Fedora or using real docker on Ubuntu. I can remove all the BOOTSNAP_ENABLE stuff and it still works...

migurski:
I just tested the old ENABLE_BOOTSNAP problem on Mac and under Github Actions, and the `rake db:migrate` step seems to pass just fine. I can’t tell what’s changed since it was causing grief last year!

mmd-osm:
We'll probably take the _db function_ topic to a new issue, as @gravitystorm raised some concerns why they still would be needed. By the way, migrations did pass even without db functions in my tests earlier today.

migurski:
I rebased this PR to keep it current.

gravitystorm:
Thanks @migurski . I'm going to merge this PR since everything here looks like an improvement to me - even if the yarn:install stuff is still not ideal.

migurski:
thanks @gravitystorm!

Issue #3041, opened 2021-01-07
Add Wiki Link to Heading


lectrician1:
The Wiki is a major resource used by the OSM community. Having a dedicated link to it in the heading would be useful.

Here's a suggested placement:
![image](https://user-images.githubusercontent.com/22477298/103848396-e6dc8f80-5067-11eb-9454-79bdff9bfd24.png)

gravitystorm:
The wiki is already linked to from as part of the Help pages. I prefer this approach, since it is more flexible and easier to expand with multiple (and future) useful external links.

We can't add everything to the menu, so we draw the line somewhere. 

And I'm not in favour of linking to external sites in the menu bar either.

lectrician1:
The Wiki is not _really_ an external site. It is hosted by OSMF and the root of the OSM community. No mapper should have to figure out that the Wiki is in box at the _bottom_ of the help page. 

Also, I've found the Wiki to be my **most-used OSM resource** as a mapper other than the site itself (for iD and carto). There should be immediate access for anyone who does or doesn't know about it.

The Wiki also acts as a community hub. Currently, we have OSMcal, regional mapping projects, links to other communication channels, and of course, tag pages - where mappers can learn more and discuss on talk pages about tags. **All** of these things are nonexistent on the main site itself and are vital to holding together the community.

I'm also a bit frustrated that you closed this based on your opinion alone, when the feature request clearly has support by numerous others.

gravitystorm:
> The Wiki is not _really_ an external site. It is hosted by OSMF and the root of the OSM community.

I'm not debating who hosts it, or whether it's part of the community. I mean that it's an external site in that you click the link and you end up on a different site. You can notice this when the top bar changes and all the links change and it everything looks different. The other links in that menu are not external sites, since you click on them and you are still on the same site, with the same menu and it looks the same and so on.

It's great that you find the wiki useful, but that doesn't mean that it should automatically get a link in that menu - there are more things to consider than that.

And as for closing the issue - I've explained my reasoning, and I'm happy to explain my reasoning in more detail if you want. Or you can offer alternative considerations to persuade me otherwise. But please don't complain about a decision based on 'clear support by numerous others' since that ignores my reasoning entirely. One of the burdens of a maintainer is to make decisions, sometimes unpopular ones - but that's why I always explain things.

lectrician1:
Yes, then please explain.

gravitystorm:
> Yes, then please explain.

[...]

PR #3086, opened 2021-02-07
Add Wiki link to header

lectrician1:
Here is the PR for adding the Wiki link to header (#3041).

It opens the Wiki homepage in another tab (not language specific however).

I wasn't sure to post this PR yet since I never got a response to my question about it on the talk mailing list.

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -31,7 +31,6 @@
         </ul>
       </div>
       <%= link_to t("layouts.history"), history_path, :class => "btn btn-outline-primary geolink flex-grow-1 current_page_class(history_path)", :id => "history_tab" %>
-      <%= link_to t("layouts.export"), export_path, :class => "btn btn-outline-primary geolink current_page_class(export_path)", :id => "export_tab" %>
     </div>
   </nav>
   <nav class='secondary'>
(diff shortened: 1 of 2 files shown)


gravitystorm:
I'm going to close this PR since I've also closed the original issue. 

Just to help you with future PRs, I'll mention a couple of things that I noticed:

* This PR removed the Export link, without providing a replacement or explanation. I'm not sure if that was intended?
* The PR contains too many 'fixup' commits. In general PRs can contain multiple commits (and I prefer if they do) but each commit should be a logical group of work, rather than a blow-by-blow record of winning the battle with rubocop :-) My recommendation is to use a combination of `git rebase -i` (the 'fixup' option is great for combining such fixup commits into the original commit, and I use this every week)  and then force-push the branch to update the PR with your preferred list of commits.

lectrician1:
> This PR removed the Export link, without providing a replacement or explanation. I'm not sure if that was intended?

That was not intended. This probably just made it in since I was experimenting with things. 

> The PR contains too many 'fixup' commits.

I messed with my Ruby dependencies and somehow broke`thor` which caused the local tests not being able to run. I didn't know how to fix it on that branch, so I decided to just test via the github actions, which I shouldn't have done. My tests work on my main branch, so I should be good from now on.

gravitystorm:
> so I decided to just test via the github actions, which I shouldn't have done

That's a totally reasonable thing to do, so don't worry about that. I do the same when it comes to things like testing CI. But when you've got it all working, then a final bit of commit squashing/fixups/rearranging can make it easier to review.

PR #3091, opened 2021-02-09
Fix unexpected spacing on note buttons when overflowed

### LINKED ISSUE #3075: the look of notes
You've recently changed the look of buttons on Notes.
But the buttons are now too big and they don't fit on the screen:
notes

When adding a new note everything is fine:
notes2

mxdanger:
Resolves #3075

Depending on the current language, the buttons for closing and commenting on notes may overflow resulting in no spacing between the buttons. This change improves accessibility as it allows the user to easily distinguish between the two buttons and reduces accidental presses on the wrong button.

Example of the issue:
before image

Example after the fix:
after image

May require additional testing.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -977,6 +977,12 @@ tr.turn:hover {
   }
 }
 
+.btn-wrapper {
+  > .btn {
+    margin-bottom: $lineheight/4;
+  }
+}
+
 /* Rules for export sidebar */
 
 .export_form {
--- a/app/views/browse/note.html.erb
+++ b/app/views/browse/note.html.erb
@@ -53,10 +53,10 @@
         </div>
         <div>
           <% if current_user.moderator? -%>
-            <input type="submit" name="hide" value="<%= t("javascripts.notes.show.hide") %>" class="btn btn-light" data-note-id="<%= @note.id %>" data-method="DELETE" data-url="<%= note_url(@note, "json") %>">
+            <input type="submit" name="hide" value="<%= t("javascripts.notes.show.hide") %>" class="btn btn-light mb-2" data-note-id="<%= @note.id %>" data-method="DELETE" data-url="<%= note_url(@note, "json") %>">
           <% end -%>
-          <input type="submit" name="close" value="<%= t("javascripts.notes.show.resolve") %>" class="btn btn-primary" data-note-id="<%= @note.id %>" data-method="POST" data-url="<%= close_note_url(@note, "json") %>">
-          <input type="submit" name="comment" value="<%= t("javascripts.notes.show.comment") %>" class="btn btn-primary" data-note-id="<%= @note.id %>" data-method="POST" data-url="<%= comment_note_url(@note, "json") %>" disabled="1">
+          <input type="submit" name="close" value="<%= t("javascripts.notes.show.resolve") %>" class="btn btn-primary mb-2" data-note-id="<%= @note.id %>" data-method="POST" data-url="<%= close_note_url(@note, "json") %>">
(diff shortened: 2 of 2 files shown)


gravitystorm:
This is a reasonable fix. I think it has two drawbacks though.

First is the size of the gap. When there are three buttons it looks like the vertical gap is much bigger than the gap between the buttons:

Screenshot from 2021-02-10 16-23-20

So I think `mb-1` looks better:

Screenshot from 2021-02-10 16-23-45

The second drawback is that we have the same problem in an unknown number of other places. It's common to have multiple buttons, and I think anywhere we have two or more buttons runs the risk of wrapping on small screens, depending on the language. And I don't want to us all to have to remember to add the margins utilities to every set of buttons, keep them consistent, and for new developers to have to figure out why they are there (wrapping in small screens in certain languages isn't so obvious, and we can't add a comment to ever situation).

Ideally, bootstrap would just do the right thing when buttons wrap, so that we don't have to think about it. So I did some research today to find out why they haven't. It's been raised at least twice:

* https://github.com/twbs/bootstrap/issues/16912
* https://github.com/twbs/bootstrap/issues/24997

The concern seems to be that since buttons are so widely used, they shouldn't have a bottom margin by default. I'm not sure that I would agree with that, but I respect their decision.

In the [bootstrap buttons documentation](https://getbootstrap.com/docs/4.6/components/buttons/) however, their buttons work fine (I'd still prefer a smaller gap of course :smile:):

Screenshot from 2021-02-10 16-58-27

It turns out they are "cheating" slightly, and have a special set of margins for the documentation:

https://github.com/twbs/bootstrap/blob/ce8af15ad1a676239d08576138b7796cae6f133f/site/assets/scss/_component-examples.scss#L181-L186

So I think we could do something similar, by having a `buttons-wrapper` class that adds a margin to the buttons, and then we can use that wrapper every time we have a set of buttons. Or we could apply the bottom margin to all buttons by default, and override the few times that we don't want it?

Finally, if we set the margin in CSS, it would be nice to pick a size that adjusts automatically if the site-wide spacing utilities are adjusted (e.g. if a designer comes along and wants to change mb-1 to be a different width). So I think we can do that with something like:

```scss
// (explanatory comment about buttons and wrapping)
.button-wrapper {
  > .btn {
    margin-bottom: $spacer * .25;
  }
}
```

My SCSS skills aren't good enough to [directly pluck the computed size out of the `$spacers` array](https://github.com/twbs/bootstrap-rubygem/blob/e9e5b38bbb88b7b28b902b172ed8c6382478ebd2/assets/stylesheets/bootstrap/_variables.scss#L130-L148) (something like `margin-bottom: $spacers[1]` if this was ruby - does that work in SCSS?) but further refinements would be welcome.

What do you think? Should we apply this to every button, or have a wrapper div? And what approach do you think is best?

mxdanger:
> What do you think? Should we apply this to every button, or have a wrapper div? And what approach do you think is best?

I believe that using the `.button-wrapper` class you proposed would be the best option for simplicity and scalability. It is better than my band aid solution of adding the bootstrap class to each affected button. 

> Should we apply this to every button...

No, in doing so we would create lots of problems elsewhere making those buttons end up being out of place with the unexpected button margins.

gravitystorm:
> I believe that using the `.button-wrapper` class you proposed would be the best option for simplicity and scalability. It is better than my band aid solution of adding the bootstrap class to each affected button.

OK, let's do that then. Do you want to update this PR or create a fresh one?

mxdanger:
> OK, let's do that then. Do you want to update this PR or create a fresh one?

Alright, I've updated this PR with a new commit.

gravitystorm:
Thanks! I've tweaked the calculation to be based on the bootstrap spacer variable, and merged this PR.

PR #3097, opened 2021-02-16
Upload profile picture with standard image formats

Revathyne:
I have made profile pictures to be uploaded only with PNG, JPG, GIF, and JPEG formats, before which images can be uploaded with any extension such as .rb, .pdf, etc.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -107,6 +107,7 @@ class User < ApplicationRecord
 
   validates_email_format_of :email, :if => proc { |u| u.email_changed? }
   validates_email_format_of :new_email, :allow_blank => true, :if => proc { |u| u.new_email_changed? }
+  validates :image, allow_blank: true, format: { with: %r{.(gif|jpg|png|jpeg)\Z}i, message: 'must be a URL for GIF, JPG ,PNG or JPEG image.' }
 
   after_initialize :set_defaults
   before_save :encrypt_password


tomhughes:
Why? What's wrong with the other formats we support?

Revathyne:
While adding images we can add files with all extensions. Sometimes while adding files with other extensions can affect the application security.

tomhughes:
Do you have any evidence that this is creating a security hole? That we're somehow handling those files in a way that  is impacting security?

At the very least there are obvious extensions like BMP, TIFF and SVG missing, all of which I am pretty sure that people are using.

Better would be to ask the storage pipeline if it is able to handle the file as an image.

tomhughes:
Ideally it should call https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-image-3F but without trying it I don't know exactly how that will interact with validations in terms of when the blob object for the field is created.

Revathyne:
Currently, I don't have any evidence about creating a security hole, but hackers can add vulnerable files with any extensions. Yes extensions like BMP, TIFF, and SVG can also be added. 
I agree with you that without trying we cannot make sure how things will work. Hope you will find some fix for this. Thanks.

grischard:
We currently support a lot of crap that imagemagick does. Having a wide variety of formats open increases the possibility of having malicious javascript masquerading as an image file.

tomhughes:
None of which changes the fact that this PR doesn't actually even work, because `image` is not a known attribute on the user model so you just get an exception when it tries to run the validations...

tomhughes:
I pushed 3c4f32a76075e91d60efd82c91bd64ebff5b6082 which will hopefully do the job - it's a little tricky as there is no official way to work out if an attachment has changed.

Revathyne:
3c4f32a works perfectly fine. :)

PR #3104, opened 2021-02-17
Format openstreetmap.org/stats/data_stats

fredrik-lindseth:
It would be nice to be able to link to specific sections on the stats page on osm.org. 

If the headings would have the IDs proposed here one could link to them like `

```
https://www.openstreetmap.org/stats/data_stats.html#number-of-editors
```
I also 

- changed the topmost heading to be h1 instead of h2 since each page should have a h1
- Put `OpenStreetMap stats`on its own and on the next line `report run at 2021-02-16 23:00:05 +0000`

The final result looks like this 
image

### DIFF
--- a/script/statistics
+++ b/script/statistics
@@ -11,7 +11,8 @@ puts "<title>OpenStreetMap Statistics</title>"
 puts "<style>th { text-align: left }</style>"
 puts "</head>"
 puts "<body>"
-puts "<h2>OpenStreetMap stats report run at #{start_time}</h2>"
+puts "<h1>OpenStreetMap stats</h1>"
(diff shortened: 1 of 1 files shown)


tomhughes:
We should just delete it really - it's a pain to manage and ridiculously inefficient these days. It dates from a time when the database was much smaller and we didn't have lots of third party tools doing a far better job.

gravitystorm:
I'd be happy to see the stats page removed - I've thought about converting it previously to a rake task, but of course that wouldn't sort out any fundamental problems with how it works.

I'll leave that decision to @tomhughes though, since he deals with the operational side of things and I can't comment on that. In the meantime, this PR makes minor improvements, so I'm happy to merge it.

PR #3118, opened 2021-03-03
Update to iD v2.19.6

quincylvania:
#### :sparkles: Usability & Accessibility
* Update lists of Backgrounds and Overlays


mmd-osm:
The pull request adds a new `vendor/assets/swfobject/swfobject.js` file. Seems a bit unexpected now that Flash is dead. Originally, this has been removed in https://github.com/openstreetmap/openstreetmap-website/commit/0ff89c31e4f8a0c3fc31772c4cf914c4fd7f5656.

quincylvania:
> This pull request adds a new vendor/assets/swfobject/swfobject.js file.

Hmm I thought I just based this on `master` and ran `bundle install` like usual. Definitely not intentional. 🤔

mmd-osm:
For some reason, swfobject is still part of the Vendorfile. This might be the reason why this is getting pulled in again.

https://github.com/openstreetmap/openstreetmap-website/blob/master/Vendorfile#L50-L54

tomhughes:
It got missed is all. There seems to be a bug in vendorer though in that even if you only ask it update iD it installs other things if they're completely missing.

mmd-osm:
@quincylvania : the old dependency is gone now. Maybe you could try again based on the current master.

PR #3124, opened 2021-03-09
Added natural=coastline and natural=peninsula to i18n file

mortenbruhn:
It turns out that neither `natural=peninsula` or `natural=coastline` are available for translation using [Translatewiki](https://translatewiki.net/wiki/Translating:OpenStreetMap).

https://wiki.openstreetmap.org/wiki/Tag:natural%3Dpeninsula

https://wiki.openstreetmap.org/wiki/Coastline

Let me know if this is the way to add it so that it will be available in Translatewiki.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -969,6 +969,7 @@ en:
           cape: "Cape"
           cave_entrance: "Cave Entrance"
           cliff: "Cliff"
+          coastline: "Coastline"
           crater: "Crater"
           dune: "Dune"
           fell: "Fell"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -987,6 +987,7 @@ en:
           moor: "Moor"
           mud: "Mud"
           peak: "Peak"
+          peninsula: "Peninsula"
           point: "Point"
           reef: "Reef"
           ridge: "Ridge"


tomhughes:
Those nominatim related values are normally updated periodically by @lonvia doing some sort of export from nominatim I think.

lonvia:
I only take care to occasionally check the list and add values that might show up in search results. 'natural=coastline' will certainly never be among them because you can't search for it. Given that you use this list also for 'nearby lookups', the list of frequent values you need might be different though.

mortenbruhn:
> I only take care to occasionally check the list and add values that might show up in search results. 'natural=coastline' will certainly never be among them because you can't search for it. Given that you use this list also for 'nearby lookups', the list of frequent values you need might be different though.

That is exactly the use case - the "Find objects" arrow in the tools find all sorts of data - including `natural=coastline`. It is possible, however, to come by `natural=peninsula` (or indeed `place=peninsula`) by just searching for the name of the peninsula:

https://www.openstreetmap.org/search?query=Jylland#map=7/56.305/9.517

`place=peninsula` does _not_ have "approved" status, so I was planning on migrating some into `natural=peninsula` instead.

Anyway..

@lonvia: Is this PR the correct way of adding such terms so that they will be picked up by Translatewiki? Should I withdraw this PR or amend it in some way - or some third option?

tomhughes:
I was mostly worried about things getting removed again buy a future sync but if @lonvia only adds things then that's not a worry.

mortenbruhn:
@tomhughes @lonvia : ping :smiley: 

(are there any impediments or objections to merging this? If not, please approve)

Issue #3110, opened 2021-02-24
Future of db/functions?


mmd-osm:
Now that we have migrated from osmosis to osmdbt for replication diffs, and quadtile calculation resides in its own Gem, we could  think about removing db/functions altogether, like shown in https://github.com/mmd-osm/openstreetmap-website/commit/d5c3828db15f5aa1e0fd90baef486e8966fdfaa4 

Also, we started a discussion on this topic in https://github.com/openstreetmap/openstreetmap-website/issues/3081#issuecomment-773210432, which can continue here now.

Before creating a pull request for this topic, I'd like to get some early feedback, if this really makes sense.

Some points to consider:

* People might still want to try out osmosis replication in their dev environment. Is this in scope for this repo?
   * Alternative option to install sql function via https://github.com/openstreetmap/osmosis/blob/master/package/script/contrib/apidb_0.6_osmosis_xid_indexing.sql  instead of using the respective db function in this repo.
   * There are some CREATE INDEX statements in there that might no longer work on Postgres > 9.5, i.e. that file might not be suitable as is today.
* Impact on Rails migrations:
    * do we need to clean up old db functions? This also applies to the already deleted maptile_for_point function.
    * do we still need those old db functions to run old migrations? (topic of https://github.com/openstreetmap/openstreetmap-website/issues/3081#issuecomment-773210432)

tomhughes:
Please don't remove anything yet - the production database still has those functions for now.

gravitystorm:
> Before creating a pull request for this topic, I'd like to get some early feedback, if this really makes sense.

I'm always in favour of cleanup, and removing complexity in particular :smile:

>     * People might still want to try out osmosis replication in their dev environment. Is this in scope for this repo?

We should push the concept of setting up replication out of this repo. I think the best approach is for osmosis to ship the functions necessary for osmosis-based replication, and osmdbt to ship whatever it needs too. We should update CONFIGURE.md to indicate there are two options to consider (osmosis and osmdbt) depending on your PostgreSQL verison and point people to the right place, but otherwise remove ourselves from that.

>     * do we still need those old db functions to run old migrations?

This is the core of the discussion, I think. We need to keep the db functions for as long as they are used in the old migrations.

We could refactor the old migrations, to remove the functions. But this raises the question of what the old migrations are there for? Two possibilities:

* They a there to help people who have a (really old) database with data in it. So we need to keep the old migrations and the functions too.
[...]

PR #3126, opened 2021-03-10
Automatically enable the btree_gist extension

gravitystorm:
This simplifies the install instructions. Loading from structure.sql is already handled.

The more I thought about what I wrote in #3110 the more I realised it was worth doing anyway.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,7 +48,6 @@ jobs:
         sudo systemctl start postgresql
         sudo -u postgres createuser -s $(id -un)
         createdb openstreetmap
-        psql -c "CREATE EXTENSION btree_gist" openstreetmap
         psql -f db/functions/functions.sql openstreetmap
     - name: Configure rails
       run: |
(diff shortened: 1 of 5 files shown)


gravitystorm:
> This simplifies the install instructions.

... and the github CI, vagrant and Docker scripts too.

tomhughes:
I think the reason we didn't do that before is that enabling `btree_gist` can only be done as a postgres superuser which means it often needs to be done from a different user to the one listed in `database.yml`. Certainly that would be true for the production database.

gravitystorm:
Ah, I hadn't thought of that. 

But is this an edge case that we need to worry about? In that situation they are already off the beaten track when it comes to the installation instructions, and will probably know what to do if the migration fails.

mmd-osm:
NB: at some point in the future when using pg13, that extension would be considered trustworthy, and wouldn’t require superuser powers anymore:

_This module is considered “trusted”, that is, it can be installed by non-superusers who have CREATE privilege on the current database._

https://www.postgresql.org/docs/13/btree-gist.html

gravitystorm:
Given the above comments, I'm still in favour of this change.

pnorman (approved):

tomhughes:
I've confirmed that current postgres no longer requires superuser to install this so given that there is clearly no reason not to do this now.

PR #3146, opened 2021-03-23
Show the crafted duration of a block independent of the reading time

### LINKED ISSUE #3143: Expired blocks do not show for how long they were crafted
An expired block shows when it was created, and in the status when it ended. 
When the user reads the message after the expiry date, the block shows the reading date as its end. 

For example, block 3168 started 2019-09-19 and ended 2021-01-25 (visible only after hovering manually over the dates #907) which is about 16 months. Thus it probably ended when the user read the message, since block cannot be crafted for this span. 

Looking at blocks retrospectively, it makes a difference if a user read a 0h-block after a year, or a 1y-block just expired.

polarbearing:
Fixes #3143.
Not tested by myself. Treat like pseudocode, I don't speak Ruby.

Intends to append the "Created" line with the difference ```duration = (ends_at - created_at)```, so it would read e.g.
```Created: over 1 year ago (4 hours)``` avoiding a new text string

### DIFF
--- a/app/views/user_blocks/show.html.erb
+++ b/app/views/user_blocks/show.html.erb
@@ -28,7 +28,10 @@
 </p>
 <% end %>
 
-<p><b><%= t ".created" %></b>: <%= friendly_date_ago(@user_block.created_at) %></p>
+<p>
+  <b><%= t ".created" %></b>: <%= friendly_date_ago(@user_block.created_at) %>
+  (<%= block_duration_in_words(@user_block.ends_at - @user_block.created_at) %>)
+</p>
 
 <p><b><%= t ".status" %></b>: <%= block_status(@user_block) %></p>
 


gravitystorm:
> Not tested by myself. Treat like pseudocode, I don't speak Ruby.

@polarbearing are there any changes we can make to make it easier for you to test yourself? We have vagrant and docker development environments, and I'm happy to work on either them or whatever else that would help.

Unfortunately untested code PRs (both not manually tested, and also without any unit tests) put a lot of burden on the maintainers, so I want to help contributors like you be able to help share that part of the workload!

polarbearing:
@gravitystorm - as discussed via PM, indeed both time and local hardware resources are currently limited for me in the specific case. 
In general, an online sandbox environment might be helpful.

gravitystorm:
I tweaked this slightly by making a separate line for the duration - I tried it out locally on a few user blocks, and I didn't think the brackets were clear enough (as well as them being a bit problematic from an i18n point of view).

PR #3147, opened 2021-03-24
Create a SessionsController

gravitystorm:
UsersController is quite complex, and so I would like to split it up into different controllers. This also allows us to (more easily) move to standard new/create/edit/update methods, as well as simplifying our largest controller. We also have a long-time desire to move towards using [Devise](https://github.com/heartcombo/devise) for some of the user-related features.

This PR moves the login/logout functionality to a new SessionsController. This is modelled on the Devise SessionsController. There are many ways that we could split up UsersController but I think [following Devise controller naming conventions](https://github.com/heartcombo/devise/tree/master/app/controllers/devise) will make things easier down the road.

Unfortunately a bunch of utility methods from UsersController are also used by the oauth methods, so either need to be duplicated, or (as I've done here) moved to a concern to be shared between the two controllers. Perhaps future refactoring will simplify this further.

I split moving tests into a new file into a separate commit, so that it's one commit that has the test changes, and a second commit that simply moves them, rather than mixing together changes and moves into the same diff.


test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def test_login
+    user = create(:user)
+
+    get login_path
+    assert_response :redirect
+    assert_redirected_to login_path(:cookie_test => true)
+    follow_redirect!
+    assert_response :success
+    assert_template "sessions/new"
```
tomhughes, line 42: Presumably these template assertions could drop the `sessions` prefix now they been moved?

tomhughes (commented):
Presumably at some point thing like the omniauth callbacks that use those shared routines should themselves move to the session controller given they are concerned with creating sessions? Though with the complication that they're also used in account creation :-(

gravitystorm:
> Presumably at some point thing like the omniauth callbacks that use those shared routines should themselves move to the session controller given they are concerned with creating sessions?

I took that approach during development, but I didn't like it, since the Devise::SessionsController doesn't have any omniauth things in it. However, there is an [omniauth callbacks controller](https://github.com/heartcombo/devise/blob/master/app/controllers/devise/omniauth_callbacks_controller.rb) for devise - I haven't checked yet, but they might go in there, but I really don't know much about omniauth yet so I'm not sure if they are doing the same thing.

test/controllers/sessions_controller_test.rb
gravitystorm: This can go either way. The trouble with `assert_template "new"` is that it matches any template named `new` - it's not scoped to the controller under test. So if you end up rendering e.g. "users/new" somehow, the test will still pass. 

So two choices:
* Keep as-is, to ensure we are rendering the sign-in form and not e.g. the user registration form.
* Remove the scoping, to be consistent with all our other tests, and accept that the risk is minimal.

I'm happy either way.

tomhughes: Let's leave it as it is in that case - we should probably aim to move to that syntax more generally!

PR #3148, opened 2021-03-24
Bundle update for mimemagic yankage

gravitystorm:
mimemagic 0.3.5 was yanked earlier today, so this updates to 0.3.6

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -260,7 +260,7 @@ GEM
       mimemagic (~> 0.3.2)
     maxminddb (0.1.22)
     method_source (1.0.0)
-    mimemagic (0.3.5)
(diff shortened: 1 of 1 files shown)


Firefishy:
Damn you are fast.

tomhughes:
Too fast - there is 0.4.0 now.

gravitystorm:
> Too fast - there is 0.4.0 now.

That was released first, but there's a somewhere-in-rails dependency on `~> 0.3.2` which doesn't match `0.4.0` which is why there was also a `0.3.6` release made slightly later.

Firefishy:
Reference: https://github.com/rails/rails/issues/41750

PR #3149, opened 2021-03-24
Sanitize classes from outputs

gravitystorm:
There are an endless amount of shenanigans that you can get up to using CSS classes. Examples are available privately on request.

This PR makes a few related changes:

* Adds tests for adding classes to table elements
* Refactors the configuration setup for the sanitize gem (although it turned out not to be necessary for this particular task)
* Strip away all class attributes from the output, by extending our custom transformer, to prevent shenanigans.


gravitystorm:
Thanks to @mmd-osm for inadvertently highlighting this to me! I liked what you'd done with your user profile on osm.org, but your 'back to the map' link gave me too many ideas for other possibilities :smile:

Issue #3160, opened 2021-04-04
Feature request: Add support for user preferences to be returned as JSON formatted response (API 0.6)


donalhunt:
Discovered that while the `/api/0.6/user/details` endpoint will provide a json response, the `/api/0.6/user/preferences` endpoint does not. :(

**GET /api/0.6/user/details with "Accept: application/json" header:**
`{"user":{"id":12345,"display_name":"firstname surname","account_created":"1970-01-01T18:18:36Z","description":"","contributor_terms":{"agreed":true,"pd":false},"img":{"href":"https://www.gravatar.com/avatar/00000000000000000000000000.jpg"},"roles":[],"changesets":{"count":0},"traces":{"count":0},"blocks":{"received":{"count":0,"active":0}},"languages":["en-GB", "en"],"messages":{"received":{"count":0,"unread":0},"sent":{"count":0}}}}`

**GET /api/0.6/user/preferences with "Accept: application/json" header:**
```
<?xml version="1.0" encoding="UTF-8"?>
<osm version="0.6" generator="OpenStreetMap server" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
  <preferences>
<preference k="diary.default_language" v="en"/>
  </preferences>
</osm>
```

Looking at the codebase, this doesn't look particular hard to implement but I wanted to a) flag it as an issue and b) check if it would be worthwhile investing the time implementing it (e.g. by submitting a PR).

gravitystorm:
Yep, at the moment only some of the API calls have json responses. I'm happy to see PRs to include more of the endpoints, or ideally all of them.

migurski:
Can you point toward an example that we can copy, Andy?

mmd-osm:
This user added some JSON stuff back in Sep 2020: https://github.com/openstreetmap/openstreetmap-website/commits?author=mmd-osm

PR #3164, opened 2021-04-07
Added JSON support to user preferences

migurski:
I’m unsure about this, because I haven’t seen an example with more than one preference.

Test plan:

```
GET /api/0.6/user/preferences?format=json

{"preference":{"k":"diary.default_language","v":"en"}}
```

```
GET /api/0.6/user/preferences?format=xml

<osm version="0.6" generator="OpenStreetMap server" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright"license="http://opendatacommons.org/licenses/odbl/1-0/">
<preferences>
<preference k="diary.default_language" v="en"/>
</preferences>
</osm>
```

Closes https://github.com/openstreetmap/openstreetmap-website/issues/3160

### DIFF
--- a/app/views/api/user_preferences/index.json.jbuilder
+++ b/app/views/api/user_preferences/index.json.jbuilder
@@ -1,0 +1,1 @@
+json.partial! @user_preferences
--- a/app/views/api/user_preferences/index.json.jbuilder
+++ b/app/views/api/user_preferences/index.json.jbuilder
@@ -1,1 +1,3 @@
-json.partial! @user_preferences
+json.partial! "api/root_attributes"
+
+json.preferences @user_preferences.map { |pref| [pref.k, pref.v] }.to_h
--- a/app/controllers/api/user_preferences_controller.rb
+++ b/app/controllers/api/user_preferences_controller.rb
@@ -7,12 +7,17 @@ module Api
 
     around_action :api_call_handle_error
 
+    before_action :set_request_formats
+
     ##
     # return all the preferences as an XML document
     def index
       @user_preferences = current_user.preferences
 
-      render :formats => [:xml]
+      respond_to do |format|
+        format.xml
+        format.json
+      end
     end
 
     ##
--- a/app/views/api/user_preferences/_user_preference.json.jbuilder
+++ b/app/views/api/user_preferences/_user_preference.json.jbuilder
@@ -1,0 +1,6 @@
+attrs = {
+  "k" => user_preference.k,
+  "v" => user_preference.v
+}
+
+json.preference(attrs)
--- a/app/views/api/user_preferences/_user_preference.json.jbuilder
+++ b/app/views/api/user_preferences/_user_preference.json.jbuilder
@@ -1,6 +1,5 @@
 attrs = {
-  "k" => user_preference.k,
-  "v" => user_preference.v
+  user_preference.k => user_preference.v
 }
 
 json.preference(attrs)
--- a/app/views/api/user_preferences/_user_preference.json.jbuilder
+++ b/app/views/api/user_preferences/_user_preference.json.jbuilder
@@ -1,5 +1,0 @@
-attrs = {
-  user_preference.k => user_preference.v
-}
-
-json.preference(attrs)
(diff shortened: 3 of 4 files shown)


pnorman (changes requested):
> {"preference":{"k":"diary.default_language","v":"en"}}

I think `{"preference": {"diary.default_language": "en"}}` would be better, unless duplicate keys are possible.

migurski:
Oh good call. Updated in fd75268:

```
GET /api/0.6/user/preferences?format=json

{"preference":{"diary.default_language":"en"}}
```

tomhughes:
Is there a reason not to just drop the `preference` containers and do a simple dictionary of key value pairs in the JSON output?

The `preference` is needed in XML to avoid having arbitrary elements but a dictionary would be the natural approach for JSON I think.

app/views/api/user_preferences/index.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.partial! @user_preferences
```
gravitystorm, line 1: This should also contain the `root_elements`, similar to the XML response. This is already implemented for e.g. showing relations - see https://github.com/openstreetmap/openstreetmap-website/blob/master/app/views/api/relations/show.json.jbuilder

When those are included, it becomes clearer that there needs to be a root `preferences` (n.b. plural) key, to match the naming of the `<preferences>` element in the XML response. 

So something like this:

```
{
    "version": "0.6",
    "preferences": {
        "foo": "bar",
        "another_pref": "something else",
        "more preferences": "go here"
    }
}
```

gravitystorm (changes requested):
This PR also needs tests. Similarly to the previous JSON-response PRs, it doesn't need to duplicate every test of the XML output, but just enough to ensure that the routes exist, and that when json is requested a suitable json-formatted response is received.

Have a look at test/controllers/api/users_controller_test.rb and search for 'json' to see what I mean.

app/views/api/user_preferences/index.json.jbuilder
gravitystorm: I should also say that the preferences are analogous to "tags" on nodes/ways/relations, in that they are already implemented in the nodes/way/relations responses so you can see how it's done there, and for json we already decided to drop the "k/v" stuff so as @pnorman mentioned it's fine to use object attributes directly for them here too.

gravitystorm:
There is also the topic of the other methods (show, update, update_all, delete) and implementing a json API for each of them. But that opens a small can of worms, since those existing methods are terrible (e.g. plaintext responses to xml payloads, empty responses for updating a user_preference vs repeating the payload when updating a node) and so I'm happy if you want to back away slowly from that part of the discussion.

app/views/api/user_preferences/index.json.jbuilder
migurski: Thanks, I’ll add these! Do you know of a way to see multiple preferences? So far I’ve only been able to view the editor language, and I’d like to have additional key/value pairs.

migurski:
I may back away slowly from that part for now.

app/views/api/user_preferences/index.json.jbuilder
pnorman: You'll have to upload preferences using [the api](https://wiki.openstreetmap.org/wiki/API_v0.6#Preferences_of_the_logged-in_user)

migurski:
Updated in a95931ea3:

```
GET /api/0.6/user/preferences?format=json

{
    "preferences": {
        "coffee": "cream, no sugar",
        "diary.default_language": "en"
    },
    "version": "0.6"
}
```

```
GET /api/0.6/user/preferences?format=xml

<osm version="0.6" generator="OpenStreetMap server" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
  <preferences>
    <preference k="diary.default_language" v="en"/>
    <preference k="coffee" v="cream, no sugar"/>
  </preferences>
</osm>
```

migurski:
Tests coming soon, while I learn how tests work in Rails.

migurski:
Tests done, all feedback addressed. I think this is ready to merge.

gravitystorm:
> Tests done, all feedback addressed. I think this is ready to merge.

Great, thanks! I also spotted a couple of minor issues with the tests, but nothing major. You can see my changes at 6bb18436a31bc192c20f3df5caf15ac5888471b5 - adding a test for the routes, adding an assertion to ensure the output is in the correct format, and then tweaking a comment that drifted out of date.

Thanks again!

PR #3169, opened 2021-04-09
Display message when the there are no notes/comments to display

harry-wood:
Display more specific/friendly message when the there are no notes/comments to display, rather than empty tables.

### DIFF
--- a/test/controllers/notes_controller_test.rb
+++ b/test/controllers/notes_controller_test.rb
@@ -73,4 +73,11 @@ class NotesControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
     assert_select "table.note_list tr", :count => 11
   end
+
+  def test_empty_page
+    user = create(:user)
+    get user_notes_path(:display_name => user.display_name)
+    assert_response :success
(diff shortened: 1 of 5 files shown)


test/controllers/notes_controller_test.rb
```diff
@@ -73,4 +73,11 @@ def test_index_paged
     assert_response :success
     assert_select "table.note_list tr", :count => 11
   end
+
+  def test_empty_page
```
harry-wood, line 77: I created a test method for this scenario. Would your normally split it out like that or bundle this into one of the existing test methods? I'm more accustomed to creating lots of separate `it` blocks in nested `context` blocks in rspec

gravitystorm: I'm happy with smaller tests. 

I also agree with you with the rspec topic - I use it on all my other projects, and I much prefer having the context blocks with individual tests, rather than ending up with our normal mega-test that are dozens of assertions and 'paragraphs' long. But I'm not sure if there's much appetite for rewriting the test suite yet, until we have a few more developers at least!

PR #3177, opened 2021-04-13
Add support for OAuth 2

### LINKED ISSUE #1408: OAuth2 support?
Hi,

Any change to support OAuth 2 protocol. I currently have a security issue due to client side cuonsumer key vulnerability with OAuth 1. Also it's getting harder to support this old format in new packages (Mainly in .Net server side code).

P.S. I'm the co-founder and main maintener of https://israelhiking.osm.org.il and I have decided to use the OSM users as my site users to encourage people to contribute to OSM mapping.

tomhughes:
This is a first pass at OAuth 2 support using doorkeeper.

I've kept the OAuth 2 configuration separate from the OAuth 1 configuration - it probably could be merged if we really wanted but hopefully we can get rid of OAuth 1 at some point...

The URL endpoints have been moved into `/oauth2/` to avoid conflicts with the OAuth 1 endpoints.

There's a test instance at https://oauth2.apis.dev.openstreetmap.org/.

Fixes #1408


mmd-osm:
A few hints for people who want to try this out using Postman:

Create new application at https://oauth2.apis.dev.openstreetmap.org/oauth2/applications, note down client id + secret for later

In Postman, create a new request, switch to the Authorization tab
demo


* Authorization Type OAuth 2,0
* Add authorization data to _Request Headers_
* Grant type: Authorization Code / or / Authorization Code (With PKCE)
* (some callback url)
* Auth URL: `https://oauth2.apis.dev.openstreetmap.org/oauth2/authorize`
* Access Token URL: `https://oauth2.apis.dev.openstreetmap.org/oauth2/token`
* Client ID + Client Secret -> see previous app registration step
* Scope: see below, needs to be a subset of the scopes used for application registration

Scope names:

- read_prefs:  read their user preferences
- write_prefs: modify their user preferences
- write_diary: create diary entries, comments and make friends
- write_api:   modify the map
- read_gpx:    read their private GPS traces
- write_gpx:   upload GPS traces
- write_notes: modify notes

config/locales/en.yml
```diff
@@ -73,7 +73,7 @@ en:
       way_node: "Way Node"
       way_tag: "Way Tag"
     # Translates all the model attributes, which is used in error handling on the web site
-    # Only the ones that are used on the web site are translated at the moment
+    # Only the ones that are used on theweb site are translated at the moment
```
gravitystorm, line 76: typo

.rubocop_todo.yml
```diff
@@ -141,6 +141,10 @@ Rails/HelperInstanceVariable:
   Exclude:
     - 'app/helpers/title_helper.rb'
 
+Rails/LexicallyScopedActionFilter:
+  Exclude:
+    - 'app/controllers/oauth2_applications_controller.rb'
```
gravitystorm, line 146: Alternatively, we could add something like

```
def show
  super
end
```

for each of the affected methods

app/controllers/application_controller.rb
```diff
@@ -345,13 +345,13 @@ def current_ability
...
   def deny_access(_exception)
-    if current_token
+    if doorkeeper_token || current_token
       set_locale
       report_error t("oauth.permissions.missing"), :forbidden
     elsif current_user
       set_locale
       respond_to do |format|
-        format.html { redirect_to :controller => "errors", :action => "forbidden" }
+        format.html { redirect_to :controller => "/errors", :action => "forbidden" }
```
gravitystorm, line 354: I'm curious about the additional forward slash. I assume it's some kind of scoping thing, but it's not something that I've seen before. Is `:controller => :errors` and alternative?

app/models/user.rb
```diff
@@ -68,6 +68,8 @@ class User < ApplicationRecord
   has_many :client_applications
   has_many :oauth_tokens, -> { order(:authorized_at => :desc).preload(:client_application) }, :class_name => "OauthToken"
 
+  has_many :oauth2_applications, :class_name => Doorkeeper.config.application_model.name, :foreign_key => :owner_id
```
gravitystorm, line 71: Although we currently have it disabled in rubocop, we should probably start adding dependent definitions to our `has_many` relationships. https://docs.rubocop.org/rubocop-rails/cops_rails.html#railshasmanyorhasonedependent As it stands, it's not clear what should happen to the `:oauth2_applications` when a user is destroyed, or whether having the applications should block the destroy from completing.

db/migrate/20201004105659_create_doorkeeper_tables.rb
```diff
@@ -0,0 +1,53 @@
...
+      t.text :redirect_uri, :null => false
+      t.datetime :created_at, :null => false
+      t.datetime :revoked_at
+      t.string :scopes, :null => false, :default => ""
+      t.column :code_challenge, :string, :null => true
+      t.column :code_challenge_method, :string, :null => true
+    end
+
+    add_index :oauth_access_grants, :token, :unique => true
+    add_foreign_key :oauth_access_grants, :users, :column => :resource_owner_id, :validate => false
```
gravitystorm, line 33: I think we'll need a follow-on migration to validate these, or alternatively some safety_assured work given that we know the tables will be empty when the foreign key is added.

app/models/user.rb
```diff
@@ -68,6 +68,8 @@ class User < ApplicationRecord
...
+
```
gravitystorm, line 72: At the bottom of https://doorkeeper.gitbook.io/guides/ruby-on-rails/getting-started it suggests adding more has_many definitions (to access_grants and access_tokens), mainly it seems so that the `:dependent` option can be set. Do we need these too?

gravitystorm (commented):
I'm not in a great position to review this PR since I have no idea how either OAuth 2 nor Doorkeeper work! So I've read through the changes and commented on things that jumped out to me. 

I also intend to review the front-end pages and forms, and maybe see if I can rearrange the translations slightly (I like having e.g. form buttons auto-translated, although I know it can be a faff to discover the right i18n key for them), but I don't think any of that would be a blocker. 

I don't have enough knowledge about the deeper questions of how the fundamental OAuth2 stuff works to comment on that.

config/locales/en.yml
tomhughes: No idea how that crept in... will fix it.

.rubocop_todo.yml
tomhughes: I tried that ;-) The problem is that just triggers a different rubocop warning:

```
app/controllers/oauth2_applications_controller.rb:14:3: W: [Correctable] Lint/UselessMethodDefinition: Useless method definition detected.
  def show ...
  ^^^^^^^^
```

app/controllers/application_controller.rb
tomhughes: Not sure how that crept in either - it doesn't seem to be necessary so I'll fix it.

app/models/user.rb
tomhughes: Well we never destroy users to it's all a bit academic...

app/models/user.rb
tomhughes: Well we don't have an immediate use for them but they won't do any harm so I'll add them.

db/migrate/20201004105659_create_doorkeeper_tables.rb
tomhughes: Well it's a new table so there's nothing to validate - new records will still have to obey the constraint.

It's true that the database won't assume it's valid for all records but I'm not sure when that would matter.

I'm not really sure where that came from... It doesn't look like it came from doorkeeper's generator so I guess I added it for some reason, likely to shut safety_assured up because it didn't realise that the table was empty so complained that validation would be blocking.

tomhughes:
I realise you probably not well placed to evaluate the actual OAuth side of things but I figured you were best placed to look at UI issues and whether it was good rails in the decisions around models, controllers etc.

mmd-osm:
Regarding updating an application, I found out that this has no impact on existing tokens. While this seems reasonable in case of adding more scopes (you could otherwise trick a user into accepting only a minimum set of scopes and extend later on without them knowing), I'm not sure about removing scopes:
As an example, I'm still able to read user preferences (read_prefs is include in my scopes), although this scope has been withdrawn in the oauth application in the menatime. When calling the user details endpoint, my token scope has higher precedence than the application scopes.

tomhughes:
Yes that is how OAuth works - you have authorised those scopes and the token will continue to have them. The application scopes just limit what an application can ask for and once a token is granted it has whatever scopes you allowed it.

I know you're only trying to be helpful but all we're really implementing here is the UI part of things and integration into our authorization - we're not really going to get into redesigning the OAuth protocol or rewriting the internals of doorkeeper.

mmd-osm:
You probably realize that I'm asking those questions, because I need to implement this logic in cgimap as well, and wanted to make sure that we have the same understanding.

tomhughes:
I think things should actually be a lot simpler in cgimap than for OAuth 1 because all you really need to do is take the token from the URL arguments or authorization header, look it up in the database and check the scopes?

mmd-osm:
Totally agree, it's a lot simpler, and boils down to a single sql statement: https://github.com/mmd-osm/openstreetmap-cgimap/blob/feature/oauth2/src/backend/apidb/oauth_store.cpp#L214-L222

I already included token expiry, as this might come up as a requirement later on. Once I have completed remaining bits on unit test coverage, we can deploy this branch to the dev instance for testing.

tomhughes:
Note that doorkeeper also allows the token to be passed using the `access_token` or `bearer_token` parameters in the URL query string as well as via the authorization header.

mmd-osm:
> [Access token in  URL query string]

According to https://tools.ietf.org/html/rfc6750#section-5.3, this approach isn't recommended:

>  Don't pass bearer tokens in page URLs:  Bearer tokens SHOULD NOT be
>       passed in page URLs (for example, as query string parameters).
>       Instead, bearer tokens SHOULD be passed in HTTP message headers or
>       message bodies for which confidentiality measures are taken.
>       Browsers, web servers, and other software may not adequately
>       secure URLs in the browser history, web server logs, and other
>       data structures.  If bearer tokens are passed in page URLs,
>       attackers might be able to steal them from the history data, logs,
>       or other unsecured locations.

Adding this logic would be easily doable for sure. It feels a bit like encouraging bad practices, though.

tomhughes:
We can certainly change it - what we currently have is the default and it is configurable but I mostly worked on the assumption that upstream know more than me and not to change defaults without a good reason.

I assume the concern is that some frameworks may make it hard to add custom headers.

tomhughes:
I've removed support for URL parameters for now - we can always add them if we come up with a good reason and that's easier than removing them somewhere down the road.

mmd-osm:
Could we add an exemption in `force_ssl_in_redirect_uri` to allow http traffic targeting 127.0.0.1?

tomhughes:
I don't recognise that name - can you provide more information?

tomhughes:
Oh I see it's an option in the configuration. We can probably just turn it off because apache already takes care of that.

mmd-osm:
It's right here in doorkeeper.rb: https://github.com/tomhughes/openstreetmap-website/blob/doorkeeper/config/initializers/doorkeeper.rb#L249

I'm trying to set up a redirect to http://127.0.0.1 ... and the UI won't let me do it.

tomhughes:
Ah but that's for the redirect, not traffic to us - what's the use case though? Are you thinking of having JOSM redirect to it's remote edit endpoint or something?

mmd-osm:
Yes, exactly, JOSM might be one use case. I'm trying one OAuth 2.0 client lib, that has http://127.0.0.1 ... as redirect all over in the config settings. I've seen some lambda expression (don't know how exactly they're called in Ruby), that would check for the hostname part, and deactivate https in case of hostname == 127.0.0.1

tomhughes:
Yes I understand the idea, and I can't see any obvious problem but I'm always nervous about changing security things that have been designed by people more expert than myself in case there is some obscure edge case I haven't considered.... I'll think about it.

mmd-osm:
I think https://tools.ietf.org/html/rfc8252#section-7.3 would be relevant here. Note the following requirement regarding 127.0.0.1:

> The authorization server MUST allow any port to be specified at the
>    time of the request for loopback IP redirect URIs, to accommodate
>    clients that obtain an available ephemeral port from the operating
>    system at the time of the request.

I'm not sure if doorkeeper already allows any port for loopback communication.

Section 7.1 ("Private-Use URI Scheme Redirection") also seems interesting for mobile apps. It would be good if more mobile app devs could join this discussion, to see what requirements they have.

tomhughes:
Yes that's how mobile apps will normally do it - do you have some reason to think that won't work at the moment? I haven't knowingly restricted what URIs can be used (other than http implicitly because it gets redirected).

tomhughes:
I've exempted 127.0.0.1 from the redirect now.

gravitystorm:
> I figured you were best placed to look at UI issues and whether it was good rails in the decisions around models, controllers etc.

I made some progress on this today, but not as much as I'd hoped due to other commitments. One thing that stood out is the need to get the doorkeeper translations https://github.com/doorkeeper-gem/doorkeeper/blob/main/config/locales/en.yml into our own translation setup. I'm not sure which is better from a maintenance point of view - copy+paste them into en.yml? Or keep them in a separate file and configure translatewiki appropriately?

tomhughes:
That's why I did neither in the end and perpetrated a [bit of hackery](https://github.com/tomhughes/openstreetmap-website/blob/doorkeeper/config/initializers/doorkeeper.rb#L6) instead - basically what doorkeeper-i18n already does and what doorkeeper itself should probably do.

gravitystorm:
> That's why I did neither in the end and perpetrated a [bit of hackery](https://github.com/tomhughes/openstreetmap-website/blob/doorkeeper/config/initializers/doorkeeper.rb#L6) instead - basically what doorkeeper-i18n already does and what doorkeeper itself should probably do.

That doesn't make them available to translatewiki though, does it? 

Also I wanted to override one of the strings ("Application Created" -> "Application Registered" to fit with the use of "Registered" on buttons, titles etc) and I can't do that if we're using the file directly from the gem.

tomhughes:
Well in the same way rails strings aren't so they have to go via rails-i18n the doorkeeper ones have to go via doorkeeper-i18n.

The alternative as you say to paste then all into our en.yml and then somehow try to keep that up to date as upstream adds and removes them and I didn't fancy that.

highfalutin:
Hi @tomhughes, great work. As you may have heard, I'm working on a tool that will use OAuth2 to validate OSM users when they sign up for memberships at join.osmfoundation.org. Would you consider adding the [openid_connect extension](https://github.com/doorkeeper-gem/doorkeeper-openid_connect)? It would reduce the number of calls back and forth between the "join" site and your oauth server.

tomhughes:
We can consider it in the future but I don't want to delay the basics and I will need to do some reading to understand OpenID Connect before we can make a decision on that anyway.

highfalutin:
Thanks for the quick reply.

mmd-osm:
CGImap would be ready for some OAuth2 testing as well. Code still lives in its own feature branch. More details: https://github.com/zerebubuth/openstreetmap-cgimap/issues/238

gravitystorm:
> Well in the same way rails strings aren't so they have to go via rails-i18n the doorkeeper ones have to go via doorkeeper-i18n.
> 
> The alternative as you say to paste then all into our en.yml and then somehow try to keep that up to date as upstream adds and removes them and I didn't fancy that.

Ah, now I understand what's going on there, thanks. I haven't tried it yet, but perhaps we can skip loading the i18n files from the doorkeeper gem? doorkeeper-i18n includes a copy of en.yml as far as I can see.

tomhughes:
> Ah, now I understand what's going on there, thanks. I haven't tried it yet, but perhaps we can skip loading the i18n files from the doorkeeper gem? doorkeeper-i18n includes a copy of en.yml as far as I can see.

Ah yes I hadn't realised it was duplicated. I've removed that hack now and we can just use the i18n version.

gravitystorm:
> Scope names:
> 
>     * read_prefs:  read their user preferences

This is one thing that I ran into earlier while I was testing out the various screens. The dummy OAuth client that I was using to trigger the token-related screens needed to know the scope names (e.g. `read_prefs`) but they aren't shown anywhere and I failed to guess them. I think the developers always need to know them, is that right? If so, we should show them on the user interface somewhere, so that after an application has been registered the developer can find them out. That leads me to think that on the oauth_application#show (and maybe #index) page, where the human-descriptions of the scopes are shown. I first thought about doing them as hover-text but this might be hard to find, so perhaps as another column in the table or as `formatted code` after the names, or similar. But I thought I should check if this information is commonly needed, or just an artefact of how I was testing it.

mmd-osm:
One topic I found a bit unfortunate during testing was that you cannot easily find out who's responsible for an OAuth2 application, or even what a particular application name refers to.

This applies both to the list of "my authorized applications", and also when you need to confirm access to some application.

Other sites provide more details, like "App1 managed by _user1234_"  with a hyperlink pointing to the user1234. In the old OAuth 1.0a implementation, the application names comes with a hyperlink pointing to some URL. providing a bit more context to a user.

Whatever works best implementation wise, any additional context would be really helpful.

oauth2

tomhughes:
Yes the OAuth2 models don't have any equivalent to that. I mean we could add it but it means deviating a bit from the upstream defaults and I wasn't really sure that anybody used it.

tomhughes:
> > Scope names:
> > ```
> > * read_prefs:  read their user preferences
> > ```
> 
> This is one thing that I ran into earlier while I was testing out the various screens. The dummy OAuth client that I was using to trigger the token-related screens needed to know the scope names (e.g. `read_prefs`) but they aren't shown anywhere and I failed to guess them. I think the developers always need to know them, is that right? If so, we should show them on the user interface somewhere, so that after an application has been registered the developer can find them out. That leads me to think that on the oauth_application#show (and maybe #index) page, where the human-descriptions of the scopes are shown. I first thought about doing them as hover-text but this might be hard to find, so perhaps as another column in the table or as `formatted code` after the names, or similar. But I thought I should check if this information is commonly needed, or just an artefact of how I was testing it.

How does this look?
image
That's the `show` view but we could do the same for `index` easily enough.

tomhughes:
This is the equivalent for the `index` view:
image

mmd-osm:
> OAuth2 models don't have any equivalent to that. I mean we could add it but it means deviating a bit from the upstream defaults and I wasn't really sure that anybody used it.

What I really had in mind was something like:

oa2

app/views/oauth2_authorized_applications/_application.html.erb:

```erb
  <td class="align-middle">
   <%= User.find_by(:id => application.owner_id).display_name %>
  </td>
```

tomhughes:
I don't the OSM username of the person that created the application tells you much though - if you do that for iD or Potlatch then I think it points at me for example!

highfalutin:
This is working great in the application I'm building. I notice that the tokens don't have a `expires_in` property. Are the tokens truly non-expiring?

tomhughes:
Yes, they are.

mmd-osm:
I'm trying to update the documentation on the /api/0.6/permissions endpoint (https://wiki.openstreetmap.org/wiki/API_v0.6#Retrieving_permissions:_GET_.2Fapi.2F0.6.2Fpermissions).

Somehow results don't seem to depend on any assigned scopes. The behaviour looks very similar to what we would see with Basic Auth. Is this expected?

tomhughes:
I have no idea. I didn't even know that endpoint existed.

tomhughes:
I mean obviously it should work but as I didn't know it existed I made no attempt to make it do so.

tomhughes:
Fixed in 608aafbf14da7d4372842828125a8a6eb492a61d. Unfortunately for compatibility it has to continue reporting `allow_x` rather than the actual scope names... I'll deploy it later.

mmd-osm:
I also started updating https://wiki.openstreetmap.org/wiki/OAuth to include more OAuth 2.0 details. I'm sure there's still lots of details missing. Please feel free to add/revise the content there as needed.

don-vip:
@tomhughes it seems fully automatic OAuth1 authentication is failing now (but semi-automatic works). We received this bug report, confirmed by two users:
https://josm.openstreetmap.de/ticket/21065

And there is a similar bug report at Gnome Maps:
https://gitlab.gnome.org/GNOME/gnome-maps/-/issues/386

tomhughes:
Well fully automatic mode should never have been implemented in the first place, but if you think something is broken then open a ticket with details of what is wrong.

mmd-osm:
Issue is in POST /oauth/authorize

JOSM sends hardcoded values in automated mode: &allow_write_api=yes&allow_read_prefs=yes&commit=Save+changes&allow_write_gpx=yes&allow_write_notes=yes&allow_read_gpx=yes&allow_write_prefs=yes

Website uses: &allow_read_prefs=0&allow_read_prefs=1&allow_write_prefs=0&allow_write_prefs=1&allow_write_diary=0&allow_write_diary=1&allow_write_api=0&allow_write_api=1&allow_read_gpx=0&allow_read_gpx=1&allow_write_gpx=0&allow_write_gpx=1&allow_write_notes=0&allow_write_notes=1&

tomhughes:
I understand why they send `yes` but it was never right (the rails code does say the value should be `yes` but that seems to be ignored for checkboxes so they always send 1/0) and now that we actually check the value it will fail yes.

mmd-osm:
Right, JOSM needs to fix the fully automated mode, there's nothing left to do in this repo.

simon04:
Fixing the parameter values in JOSM is easy. However, it might take a while
till all users update to the new JOSM version. All previous versions will
forever be broken regarding the authentication method we've been
recommending for novice users for years. Isn't there another option?

tomhughes:
Well I could make the web site accept `yes` there but I'm reluctant because that endpoint is an internal implementation detail of the web site - the only thing that should ever be submitting to it is our web site code.

The fact that JOSM has a "fully automatic" mode that completely ignores how OAuth is supposed to work and abuses internal endpoints is the root causes here.

mmd-osm:
The Gnome maps app mentioned earlier on uses the same pattern, and sends a POST request to the /oauth/authorize endpoint:
https://gitlab.gnome.org/GNOME/gnome-maps/-/blob/master/src/osmConnection.js#L314-337

PR #3193, opened 2021-05-11
Update to rails 6.1

tomhughes:
Finally all the pieces are in place to upgrade to rails 6.1 so here's my offering...

A test instance is running at https://rails61.apis.dev.openstreetmap.org/.


gravitystorm (approved):
Looks good to me. I worked on the same stuff for a different project yesterday, so I recognise the `app:update` changes in all the config files!

gravitystorm:
I think we'll be able to enable most of the 6.1 framework defaults without any hassle, but we may as well wait until after this is merged since those are not blocking anything in particular - and will make bisecting easier if we do run into any issues.

mmd-osm:
Is the Docker build failure expected? I'm asking b/c some potential contributors to this site explicitly mentioned they're intending to use Docker to bootstrap their development process.

tomhughes:
The docker build passes or fails based on the output of some random generator so I just ignore those failures.

gravitystorm:
> Is the Docker build failure expected?

I've found the docker tests to be both slow and flaky, ever since it was introduced. I must say that I'm in the bad habit of ignoring the failures already.

gravitystorm:
> > Is the Docker build failure expected?
> 
> I've found the docker tests to be both slow and flaky, ever since it was introduced. I must say that I'm in the bad habit of ignoring the failures already.

I should expand on that and say that most times I see it happen, I rerun the tests and it passes. So for my own PRs I do this (some times a couple of times) so that we get green lights across the board, and to save anyone else from diving into it or to remove any reluctance to merge. But when I see that everything is green and the docker test is red, I pretty much assume it's a false alarm.

mmd-osm:
I was just wondering if there's some change that still needs to be reflected in the Dockerfile and related files somehow. If that's not the case, then all is fine from my pov.

PR #3201, opened 2021-05-19
WIP: Use database replicas for read requests

### LINKED ISSUE #2634: Use database replicas for read requests
We realised in #2563 that we needed to update our controller tests, so that they use the DatabaseSelector middleware in the test suite and pick up any cases of trying to write to the database during read requests. Now that the refactoring is done I thought I would revisit the multiple databases setup.

However, last week I found a new problem when running the test suite, and it's taken many hours to narrow down the cause (since it lies somewhere deep deep within ActiveRecord, database connections and rails middleware) into a simplified test case, and even then I'm still not sure what's causing it our how to fix/workaround it. I think it's just in the test suite since the rails app appears to work fine for me, and even throws the right error in the right places. But until the problem with the test suite is sorted, this is on hold.

[I've reported the problem I found to the rails team](https://github.com/rails/rails/issues/39449), maybe someone there can point us in the right direction! This issue here is just to share progress with everyone and avoid anyone else having to duplicate the work I've done. If you want to get involved, you can see it (not) working on my [`multiple_databases_2`](https://github.com/gravitystorm/openstreetmap-website/tree/multiple_databases_2) branch, or have a look at the testcase app linked from the rails issue.

gravitystorm:
Fixes #2634

The connection switching is managed using the standard Resolver, which chooses the connection based on the HTTP verb. In the small number of cases where we update the database during GET requests, we override the default choice of connections.

Additionally, after any write in a particular user session, the resolver will continue to use the same (write) connection for two seconds for subsequent reads. This gives time for the changes to be replicated before switching back.

Unfortunately [there's a bug in rails when running migrations](https://github.com/rails/rails/issues/41855) so we can't merge this yet, but I thought I would push this now for review and discussion anyway. It contains a lot fewer connection overrides than when I last looked at this, since there's been some refactoring in the meantime as we move more things to standard new/create/update/etc.

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -115,7 +115,10 @@ class ApiController < ApplicationController
                           elsif username == "token"
                             User.authenticate(:token => passwd) # preferred - random token for user from db, passed in basic auth
                           else
-                            User.authenticate(:username => username, :password => passwd) # basic auth
+                            # User.authenticate might try to update the saved password
+                            ActiveRecord::Base.connected_to(:role => :writing) do
+                              User.authenticate(:username => username, :password => passwd) # basic auth
+                            end
                           end
       # log if we have authenticated using basic auth
       logger.info "Authenticated as user #{current_user.id} using basic authentication" if current_user
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -186,8 +186,10 @@ class ApplicationController < ActionController::Base
 
   def set_locale
     if current_user&.languages&.empty? && !http_accept_language.user_preferred_languages.empty?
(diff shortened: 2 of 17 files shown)


pnorman:
Upstream bug has been fixed.

gravitystorm:
> Upstream bug has been fixed.

Great! I'll update this PR and check if everything is working now.

gravitystorm:
I'm going to close this PR. If we decide to support read-only replicas in the future (I'm still in favour!) then I think it's probably be easier to start afresh, rather than rebase this PR.

pablobm:
If starting afresh, would it be a matter of applying the same sort of changes shown here, or are there any new details to watch out for?

Issue #2785, opened 2020-08-20
Missing map key in the bicycle layer for violet bicycle ways


Sovxx:
Hi,
There is no explanation for what violet way is :
![violet](https://user-images.githubusercontent.com/12556923/90828916-2287ee80-e33f-11ea-8d8c-6cae7210c62c.png)

mnalis:
I would be interested to make PR to fix those colors in legend as identified by @systemed, any objections?

PR #3208, opened 2021-05-25
fix map key color for regional OpenCycleMap

mnalis:
as it changed from light cyan to purple on the map.

Closes: https://github.com/openstreetmap/openstreetmap-website/issues/2785


gravitystorm:
Thanks!

PR #3212, opened 2021-06-04
fixed nodejs package name for fedora

Robbendebiene:
There is no "js" package, so this is probably a typo and should be "nodejs" (https://nodejs.org/en/download/package-manager/#centos-fedora-and-red-hat-enterprise-linux)


tomhughes:
Actually there did used to be a `js` package (https://src.fedoraproject.org/rpms/js) but it is not retired.

Robbendebiene:
Oh you are right. I didn't know about that.

Issue #3215, opened 2021-06-08
Stray <br/> in 'confirmation email resent'-message


pietervdvn:
I created a new account on master.dev.apis. As the email confirmation didn't come through, I clicked the 'resend confirmation email'. The message I got then got an extra <br/><br/> (see screenshot)

![image](https://user-images.githubusercontent.com/1466478/121197049-44704480-c871-11eb-9f8f-41a4b1b00efa.png)

tomhughes:
Are you saying this only happens on that test instance?

tomhughes:
Ah no I can reproduce it on my test instance.

pietervdvn:
> Are you saying this only happens on that test instance?

No, I'm saying that I only tested it there - but I didn't try any other instance.

gravitystorm:
I wasn't aware of this particular error, but I've wrestled with html in our flash messages before and it's a bit of a nightmare. 

Rails flash [only allows](https://api.rubyonrails.org/classes/ActionDispatch/Flash.html) `String`, `Array` and `Hash` as flash objects, notably not `SafeBuffer`. So if we output the raw string (via `raw` or marking it as `.html_safe`) then we get problems like https://github.com/openstreetmap/openstreetmap-website/issues/1027 . If we let rails escape the `String`, then we get problems like this issue.

I'm not sure what the best approach is, so any suggestions are welcome.

mmd-osm:
This second paragraph was added back in 2009, and in case confirmation emails are longer relevant to a majority of users in 2021, we might as well just drop it.

tomhughes:
That's an interesting theory - shame it has no obvious basis in reality.

Do you know something I don't about what proportion of users use a social login from a provider where we accept the email without verification?

mmd-osm:
Well, in that case, we could rephrase the last sentence to something like "In case of issues, make sure to add support@osm to your safe senders list", which would be shorter, thereby avoiding the line break issue and hopefully is still to the point.

mmd-osm:
As a side note, why is the flash message even mentioning support@openstreetmap.org as an email address? Maybe I'm missing something here, but the automated confirmation links are currently sent out using "web@noreply.openstreetmap.org".

PR #3232, opened 2021-06-23
Use partials to support complex flash messages

gravitystorm:
This PR addresses #3215. I did a bunch of research to see what alternative approaches people were using - mostly, it seems, just opening themselves up to security errors by using .html_safe everywhere, going by the blogs and stackoverflow comments that I found. 

However, an approach I saw mentioned a couple of times and in particular used by a mysociety application is to use the ability of flash messages to store Hashes. Then you can use this Hash to render a template, either to include html or just to use the translation features. Since the only things we pass to the flash object is some strings that will need escaping (and we don't store any html anywhere in the flash object) everything works nicely. I'm quite pleased with this approach.

I also took the opportunity to move the flash messages to use a flex layout instead of custom CSS, which uncovered a small issue involving padding on floated elements in the header, so that was fixed too.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1404,25 +1404,6 @@ tr.turn:hover {
   &.notice {
     background-color: #CBEEA7;
   }
-
-  div.message {
-    display: inline-block;
-    margin-left: $lineheight / 2;
-    vertical-align: middle;
-
-    p {
-      margin-top: $lineheight * 0.5;
-      margin-bottom: $lineheight * 0.5;
-
-      &:first-child {
-        margin-top: 0px;
-      }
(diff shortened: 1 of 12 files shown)


tomhughes:
Were the flash messages you converted here just he ones that were causing a problem, or is this mostly infrastructure and we should be looking to convert he others over time as well?

gravitystorm:
As far as I could find yesterday, these are the only two flash messages that contain html - the rest are just plain text. It's fine to leave all the plain text ones as they are (much more simple for the majority case), but perhaps we'll want more flash messages with html (e.g. links) in future, and they should use the partial approach.

If you spot any other flash messages with links or other html in them, let me know!

PR #3237, opened 2021-06-27
Switch web site to use OAuth 2

tomhughes:
This switches the web site to use OAuth 2 for authentication.


mmd-osm:
Just wanted to double check: the embedded iD editor would still receive all needed OAuth 1.0a parameters, because id.html.erb uses `current_user.access_token(Settings.id_key)` to get those values, right?

tomhughes:
Correct - that can be converted separately later.

gravitystorm:
Looks good to me.

Issue #3167, opened 2021-04-07
Split User settings into different forms?


gravitystorm:
We currently have one big form for many things a user can change, including:

* Passwords and email addresses
* Profile description and avatar
* Preferred language and editor
* Home location

In my mind, these cover too many disparate things, and the page suffers from complexity both for the user and also for the developers (e.g. have a look at users_controller and see what goes on when you press save).

I think we should split this form up. My first instinct is to separate out things that are public (avatar, profile information, home location) from things that are private (email address, password, etc), and have "Edit Profile" and "Edit Settings" as two distinct forms. This will allow us to explain to users more clearly that certain settings are public information, even if they don't realise it (like home location).

But that still leaves two problems, firstly that the remaining settings are still a bit of a grab-bag between 'account' settings like email / password, and things that are more like 'preferences' such as language and preferred editor. Secondly, we already have another sub-page on the My Settings page, for OAuth, which is hidden away but still counts. So maybe we need:

* Account settings - (Display Name, Email Address, Password, External Authentication, Public editing, Contributor Terms, Go Public)
* Profile settings - (Profile description, avatar, home location)
* Preferences - (Preferred Language, Preferred Editor)
* OAuth Settings - no changes here, but maybe rename to Application Settings?

I also intend to rework viewing your own profile, to make it clearer what shows up for other people. The Users#show currently contains 20+ `if` statements since it's used for both public profiles and also showing extra things to yourself. A separate dashboard page for your own stuff would be more useful and understandable.

Before I try any of this out, I'm interested to hear what other people think, both about the underlying concept of splitting things up a bit, as well as the details of what could go where.

gravitystorm:
I've started working on this, beginning with moving the "preferences" to a separate form. When that's finished I'll be looking to split the account settings from the profile settings.

simonpoole:
While I completely understand the desire to split things up and to bring some order to chaos here, I'm not convinced that basing the split on quirks of the English language is going to work.

Right now in German there are "Settings" and "My User Settings" no indication what the difference would be to the end user, and essentially semantically that is exactly the same thing, given there are no global setting for the whole site that a normal user can change.

gravitystorm:
> While I completely understand the desire to split things up and to bring some order to chaos here,

Great! :smiley: 

> I'm not convinced that basing the split on quirks of the English language is going to work.

[...]

PR #3257, opened 2021-07-14
Separate forms for preferences and profile

gravitystorm:
This forms the first part of implementing #3167 , by creating separate forms for preferences and for changing your profile information. 

This helps make the users#account form (and update logic) simpler, although it's still very complex. The preferences page is currently a bit sparse, but could be expanded as we add more things to the site, and could be improved by much better descriptions of how the language preferences are supposed to work! Again, these can be follow-on tasks.


app/controllers/preferences_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  before_action :check_database_readable
+  before_action :check_database_writable, :only => [:update]
+
+  def show; end
+
+  def edit; end
+
+  def update
+    current_user.languages = params[:user][:languages].split(",")
```
tomhughes, line 17: I think this needs a `set_locale` call to ensure that any changes are active? You removed it from the user settings but don't seem to have added it back here...

app/views/preferences/edit.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  <h1><%= t ".title" %></h1>
+<% end %>
+
+<%= bootstrap_form_for current_user, :url => { :action => :update } do |f| %>
+  <%= f.select :preferred_editor, [[t("editor.default", :name => t("editor.#{Settings.default_editor}.name")), "default"]] + Editors::AVAILABLE_EDITORS.collect { |e| [t("editor.#{e}.description"), e] } %>
+
+  <%= f.text_field :languages %>
+
+  <%= f.primary t(".save") %>
+  <%= link_to t(".cancel"), preferences_path, :class => "btn btn-link" %>
```
tomhughes, line 11: Would `btn-outline-secondary` be a better a choice here? Having a link alongside a button looks a little odd to me?

app/views/profiles/edit.html.erb
```diff
@@ -0,0 +1,61 @@
...
+    </div>
+    <div class="form-check">
+      <input class="form-check-input" type="checkbox" name="updatehome" value="1" <% unless current_user.home_lat and current_user.home_lon %> checked="checked" <% end %> id="updatehome" />
+      <label class="form-check-label" for="updatehome"><%= t ".update home location on click" %></label>
+    </div>
+    <%= tag.div "", :id => "map", :class => "content_map set_location" %>
+  </fieldset>
+
+  <%= f.primary t(".save") %>
+    <%= link_to t(".cancel"), user_path(current_user), :class => "btn btn-link" %>
```
tomhughes, line 60: Same comment here about the class for the cancel button/link.

app/views/users/show.html.erb
```diff
@@ -165,6 +165,12 @@
 
     <div class="user-description richtext text-break"><%= @user.description.to_html %></div>
 
+    <% if current_user and @user.id == current_user.id %>
+      <div class="my-3">
+        <%= link_to t(".edit_profile"), edit_profile_path, :class => "btn btn-outline-primary" %>
```
tomhughes, line 170: I'm not sure about the placement of this - it looks a bit odd to me. That may be partly because it's above the extra user data line that administrators like me get but I think it would stick out a bit without that.

Could we maybe put it on the right at the top? or would that be an issue on small screens?

app/views/preferences/edit.html.erb
gravitystorm: I'm happy with the cancel link being just a link, since it reinforces the idea that it's not going to change any state on the server. 

We also use this pattern elsewhere already (traces, user blocks and messages) so I'd prefer to keep it consistent here, and if we do decide to change it, do so everywhere in a separate PR.

app/views/users/show.html.erb
gravitystorm: I can move it around, but to be fair I think the whole page is already a mess. My plan is a followup PR that 

* Moves the profile content out of the (grey) page header and into the main part of the page
* Remove the strange override that makes the profile page the only edge-to-edge page
* Moves the admin stuff into the header along with the rest of the stuff that's there already
* Moves the non-profile stuff (like lists of friends, links that aren't shown to other people etc) off that page and into a separate dashboard

I think at that point the edit button will sit more naturally under the profile content.

app/controllers/preferences_controller.rb
gravitystorm: There's a small discussion of that in the commit message for 9b8f2bb, but the main reason for removing it is that we no longer render a page if the update is successful. We redirect instead, which means on the subsequent page load, the user object is freshly loaded and the page will show in the new language. If the save is unsuccessful, the form is rendered again but I think that's appropriate to show the original language.

The one tiny edge case is that the flash message shown will be in the original language, so I could change that if you think the complexity is worth it?

tomhughes: If that only needs the `set_locale` then I guess it might be reasonable but if it needs all the other nonsense you removed in that commit then probably not... I gave up even trying to understand the subtly of all that reset stuff.

app/views/preferences/edit.html.erb
tomhughes: Yes I did notice that pattern was being used in other places - maybe I'll just do a PR to fix them all then.

app/views/users/show.html.erb
tomhughes: Sounds reasonable to me...

app/controllers/preferences_controller.rb
gravitystorm: I just pushed a fix for this using `flash[:notice] = t(".success", :locale => Locale.available.preferred(Locale.list(current_user.languages)))` but I've just realised that a much better approach would be to use the "flash partials" approach from #3232 so give me a moment to refactor.

app/views/preferences/show.html.erb
```diff
@@ -0,0 +1,28 @@
...
+  <% else %>
+    <dd class="col-sm-8"><%= t("editor.default", :name => t("editor.#{Settings.default_editor}.name")) %></dd>
+  <% end %>
+
+  <dt class="col-sm-4"><%= t ".preferred_languages" %></dt>
+
+  <dd class="col-sm-8">
+    <ul>
+      <% current_user.preferred_languages.each do |locale| %>
+        <li><%= locale %></li>
```
tomhughes, line 19: I thought I had mentioned this in my previous review but apparently I missed it...

I don't know about you but I'm not particularly keen on the bulleted list here... What do you think to just a simple list? (I did `list-style: none; padding-left: 0` to test it)

config/locales/en.yml
```diff
@@ -1644,6 +1644,42 @@ en:
       reset: "Reset Password"
       flash changed: "Your password has been changed."
       flash token bad: "Did not find that token, check the URL maybe?"
+  preferences:
+    show:
+      title: My Preferences
```
tomhughes, line 1649: This, and the other titles are in title case but I think we generally only capitalise the first word of page titles? Though I'm not sure I don't prefer it this way...

gravitystorm: Oh, I've never noticed. We seem to be inconsistent:

* Initial - GPS traces, Users' diaries, My settings
* Title case - Copyright and License, Getting Help, My Messages

So I think we should keep the title case here and converge on that in the future.

app/views/preferences/show.html.erb
gravitystorm: Done.

app/controllers/preferences_controller.rb
gravitystorm: OK, I've refactored (and force-pushed) so that we use a partial to ensure the successful flash message is shown in the updated language (simply by deferring the lookup of the flash message to the subsequent page load). I've added tests, but the important one is disabled until there's actually another translation to test against!

PR #3267, opened 2021-07-22
Use btn-outline-secondary for cancel buttons

tomhughes:
As discussed in the #3257 review this changes all cancel buttons to use `btn-outline-secondary` instead of `btn-link` because to my mind at least having a row of buttons and then what looks like a link is quite odd.

In all cases the existing buttons are all non-outline so the cancel button retains a distinctive look by being an outline button.

### DIFF
--- a/app/views/messages/new.html.erb
+++ b/app/views/messages/new.html.erb
@@ -8,5 +8,5 @@
   <%= f.richtext_field :body, :cols => 80, :rows => 20 %>
 
   <%= f.primary %>
-  <%= link_to t(".back_to_inbox"), inbox_messages_path, :class => "btn btn-link" %>
+  <%= link_to t(".back_to_inbox"), inbox_messages_path, :class => "btn btn-outline-secondary" %>
 <% end %>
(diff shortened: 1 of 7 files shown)


gravitystorm:
I've spent longer than I should thinking about this! I still lean 60/40 towards keeping the cancel links as looking like links. 

I tried to find any UX guides on this, but most of them said that for standard HTML forms there's no need to have them in the first place! They are only necessary on popups, modals, and forms that appear inline without a page reload, none of which apply here. 

* https://www.nngroup.com/articles/reset-and-cancel-buttons/
* https://ux.iu.edu/writings/buttons-vs-links-basic/

The one place where I'm more strongly against the button look is for the traces#new form, where it's a link to "Help" rather than a cancel link. I think this should remain as a link, even if the rest of them become secondary buttons.

tomhughes:
I tend to agree that many if not all of those cancel buttons should probably be removed...

I think what I find odd is having links alongside a button - ie where you expect to see (if anything) more buttons. A link underneath like the examples in the second link you gave would be fine for example.

PR #3284, opened 2021-08-02
Require login for User blocks

mmd-osm:
This PR addresses one of the low hanging fruits on the GDPR list by requiring users to log in for anything User Block related.

See https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services#User_blocks

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -24,7 +24,6 @@ class Ability
       can [:new, :create, :destroy], :session
       can [:index, :show, :data, :georss, :picture, :icon], Trace
       can [:terms, :new, :create, :save,  :show, :auth_success, :auth_failure], User
-      can [:index, :show, :blocks_on, :blocks_by], UserBlock
       can [:index, :show], Node
       can [:index, :show, :full, :ways_for_node], Way
       can [:index, :show, :full, :relations_for_node, :relations_for_way, :relations_for_relation], Relation
@@ -51,6 +50,7 @@ class Ability
         can [:new, :create], Report
         can [:mine, :new, :create, :edit, :update, :destroy], Trace
         can [:account, :go_public], User
+        can [:index, :show, :blocks_on, :blocks_by], UserBlock
 
         if user.moderator?
           can [:hide, :hidecomment], DiaryEntry
--- a/test/controllers/user_blocks_controller_test.rb
+++ b/test/controllers/user_blocks_controller_test.rb
@@ -63,7 +63,15 @@ class UserBlocksControllerTest < ActionDispatch::IntegrationTest
     expired_block = create(:user_block, :expired)
     revoked_block = create(:user_block, :revoked)
 
+    # Check that the user block page requires us to login
     get user_blocks_path
+    assert_redirected_to login_path(:referer => user_blocks_path)
+
+    # Login as a normal user
+    session_for(create(:user))
+
+    get user_blocks_path
+
     assert_response :success
     assert_select "table#block_list", :count => 1 do
       assert_select "tr", 4
@@ -78,7 +86,15 @@ class UserBlocksControllerTest < ActionDispatch::IntegrationTest
   def test_index_paged
     create_list(:user_block, 50)
 
+    # Check that the user block page requires us to login
     get user_blocks_path
+    assert_redirected_to login_path(:referer => user_blocks_path)
+
+    # Login as a normal user
+    session_for(create(:user))
+
+    get user_blocks_path
+
     assert_response :success
     assert_select "table#block_list", :count => 1 do
       assert_select "tr", :count => 21
@@ -97,6 +113,13 @@ class UserBlocksControllerTest < ActionDispatch::IntegrationTest
     active_block = create(:user_block, :needs_view)
     expired_block = create(:user_block, :expired)
     revoked_block = create(:user_block, :revoked)
+
+    # Check that viewing a block requires us to login
+    get user_block_path(:id => 99999)
+    assert_redirected_to login_path(:referer => user_block_path(:id => 99999))
+
+    # Login as a normal user
+    session_for(create(:user))
 
     # Viewing a block should fail when a bogus ID is given
     get user_block_path(:id => 99999)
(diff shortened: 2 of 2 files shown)


tomhughes:
Why does it make any difference from a GDPR point of view whether a user is logged in or not?

Either it's OK to reveal that a user is blocked to another unrelated person or it isn't. I don't see how whether or not you're a registered user makes any difference?

pnorman:
> See https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services#User_blocks

This is a proposed list. The discussion on the proposal was intended to take place after a PR, so there would be something to review. Now that there's a PR (this one), we can discuss something concrete.

mmd-osm:
That would be a question to @woodpeck who created that Wiki page originally based on LWG recommendations.

According my understanding of the "Terms of Use" table on the wiki page, a logged on user would be a prerequisite to show user blocks. 

At some point in the future, a user would have to accept some new "Terms of Use" in addition. As we don't have this part in place yet, I have covered the first step for now.

simonpoole:
New users have had to accept the ToU for something like two years.

Requiring old users to accept them for continued use of the website is still waiting on the OSMF board to formulate a communication strategy and schedule for the cut over.

And yes the basic idea is that sensitive information can only be accessed by accounts that have agreed to the ToU.

simonpoole:
> (Edit after Simon's comment below: we can't just use the ToU flag as of yet, as long as we don't have a working cut over plan for old users.)
> 

In principle basing such access on the combination off logged in user and accepted ToU would be quite acceptable. It simply wasn't something that was considered at the time (because it was assumed that the temporary state we've been in now for two years would be much shorter). Given that nobody is holding their breath for a board decision 'soon' however this might be a more viable way forward.

mmd-osm:
> To be clear: from a legal pov, from a community and practical side I would have some **doubts.**

@simonpoole : I'm not sure I'm following you here. Could you be a bit more explicit in your description what you see as next steps? I'm not very good at reading between the lines, there are way too many allusions here.

As a reminder, I'm not claiming this PR will be the final solution, but only the first building block.

simonpoole:
> 
> 
> > To be clear: from a legal pov, from a community and practical side I would have some **doubts.**
> 
> @simonpoole : I'm not sure I'm following you here. Could you be a bit more explicit in your description what you see as next steps? I'm not very good at reading between the lines, there are way too many allusions here.

Well I don't think, at least long term, having different classes of contributors is a good thing from a community cohesion pov. It would simply be a way to lessen the perceived impact of the undoubtedly unpopular "forced" acceptance of the ToU, and by that make it more palatable for the board.   

The practical downside is that we would have to make the ToU accepted flag available via the API, because just as for the OSMF, third party sites need to know if the ToU have been accepted or not, with other words extra work there too.

woodpeck:
There was some confusion initially because we equated "logged-in user" and "user who has agreed to the ToU". For a transition period this would not be the case. 

Nonetheless, my suggestion would be:

* don't design for a two-tier world; assume that all users who can log in will also have agreed to the ToU. Do not spend any effort on making the "has this user agreed to the ToU" flag externally available since it is temporary; do not spend any effort on keying certain functionality to th "has this user agreed to the ToU" flag.
* start restricting access to some things to users who are logged in (even if, currently, this doesn't mean much)
* separately and independently, work on a timeline and communications strategy to deny login to users who have not agreed to the ToU (along the lines of: announce on the lists, 3 months later show a message upon login that asks people to agree to the ToU but they can click on "not now", another 3 months later remove the "not now" option).

And regarding the latest point, I wouldn't bee overly cautious about what @simonpoole dreads to be "unpopular". Just say these terms of use have been in effect for 2yrs+ now and it's time that everyone sign up to them, it's just a formality, click here and you're done. 

Summing up, I think @mmd-osm's approach here is exactly right.

FlyingSpitz:
Can you please remove the email updates as I no longer use this email for
GitHub
Thanks

mmd-osm:
Hey, our friends on the osmf board have put this PR on their next board meeting's agenda: https://wiki.osmfoundation.org/w/index.php?title=Board/Minutes/2021-08-13#OSM_website_Terms_of_Use

amandasaurus:
I (as board member) added it to the agenda of the mid month chat (it's not a board meeting). I merely want to talk about it with the rest of the board. We cannot make decisions at that meeting. I predict this will only be an initial board conversation. 🙂 

But please tell us/me anything I/we need to know before Friday. 🙂

mmd-osm:
@amandasaurus :  meeting minutes for the mid month chat still seem somewhat incomplete at this point. Can you maybe give us an update what’s next on ToU topic? Do we need to wait for anyone to make some progress on this pull request?

amandasaurus:
I've talked to several OSMers, including OWG members about this. @tomhughes has confirmed that board action is not needed for this, or any other PR which would change non-ToU accepting people's use of the webite. I am happy to leave it to them. I have said if the board needs to vote on something, we can do that.
The rest of the board doesn't know much about this topic. The board hasn't taken a formal vote, but I think everyone on the OSMF board is happy to leave this to the decisions of the existing OWG & osm-website maintainers to decide.
While I am able to programme, I do not have the time myself to implement this type of patch. I don't even know where to begin. If someone thinks the board should take some action, then they can present a possible resolution/motion for the board to vote on (email board@osmfoundation.org).
Sorry for the delay in telling yous. I was thinking about this, but I appreciate @mmd-osm reminding me like this. 🙂

simonpoole:
"Your honor nobody produced a pull request, so we are clearly not guilty."

Just to re-re-re-point this out: making acceptance of the ToU mandatory for existing users was a separate point of the recommendations of the LWG to the OSMF board. Some of those recommendations the board acted on (unsuccessfully) to put in to place (for example the API changes) overs it left to later, the issue in question is one of them.

And, be any measure, the GDPR related research, the necessary actions and everything around it are by far the best, most detailed and well explained of any thing in the whole history of the OSMF. The board simply has no excuse at all to not get it's act together.

tomhughes:
Frankly I've long ago lost track of the discussion on this PR but I suspect it's almost entirely irrelevant and unnecessary. My point of view is:

* I have no particular problem with this change, or with requiring existing users to accept the terms of service and as far I know neither do any other maintainers of the web site. Equally I'm not especially interested in writing the code to do that, or any of the GDPR stuff, myself.
* This PR serves no purpose unless we first have the PR to require existing users to accept the terms of use.
* Any decision to require existing users to accept the terms of use ought to be accompanied by some publicity/explanation from the board or CWG or whatever when that goes into effect.

mmd-osm:
> This PR serves no purpose unless we first have the PR to require existing users to accept the terms of use.

Well, this isn't exactly what @woodpeck (and I believe also @simonpoole) suggested before: this PR is independent of any ToU discussion, i.e. we don't have any prerequisite for users to accept the ToU first. We can even completely ignore the ToU discussion for this PR.

tomhughes:
Yes it's independent in a technical sense, but it achieves nothing.

It makes it harder to access some information for no reason.

The stated reason is to ensure that only people that have accepted the ToU can access it, but that will only be true once we require existing users to accept them.

mmd-osm:
Closing now, as this PR isn't going anywhere.

PR #3285, opened 2021-08-03
Update attribution for Australia

andrewharvey:
Per request from the copyright holder to DWG updated to match requested attribution per https://data.gov.au/data/dataset/bdcf5b09-89bc-47ec-9281-6b8e9ee147aa

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1869,10 +1869,10 @@ en:
           <a href="https://www.vorarlberg.at/vorarlberg/bauen_wohnen/bauen/vermessung_geoinformation/weitereinformationen/services/wmsdienste.htm">Land Vorarlberg</a> and
           Land Tirol (under <a href="https://www.tirol.gv.at/applikationen/e-government/data/nutzungsbedingungen/">CC BY AT with amendments</a>).
         contributors_au_html: |
-          <strong>Australia</strong>: Contains data sourced from
-          <a href="https://www.psma.com.au/psma-data-copyright-and-disclaimer">PSMA Australia Limited</a>
+          <strong>Australia</strong>: Incorporates or developed using Administrative Boundaries &copy;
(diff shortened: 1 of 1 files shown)


tomhughes:
Has this been cleared by LWG?

andrewharvey:
The original request went to legal-questions, which I got forwarded since I organised the original waiver, so I've asked legal-questions if they can review here.

andrewharvey:
@tomhughes I asked legal-questions if they could review, but haven't heard anything yet. The copyright holder has just asked if this could be updated again, so I don't see how this would be an issue given it's per their request. Thanks.

simonpoole:
@andrewharvey I don't believe that the legal-questions OTRS queue is actively being worked through, at least not for multiple months. I know and then point out stuff that could have real legal consequences (as in this case), but that's the limit of my involvement. You probably should contact the LWG directly (legal@osmfoundation.org), or @mackerski

mackerski:
Hi folks,

As Simon notes, LWG is rather backlogged in OTRS, a situation we hope to
resolve. I've identified the ticket in question and I have added this topic
to tomorrow's LWG agenda. We will be back in touch once we have reviewed
things.

Thanks,
Dermot

andrewharvey:
@mackerski was this fine? I can't see any issues with updating the attribution, the data provider is asking for this to be updated.

openstreetmap-website:
Hi Andrew, and sorry for the delay.

Yes, the LWG is happy that this change be made where we provide this kind
of attribution. But we also recognised that this same change of name means
that the existing CC waiver we have (which refers to the old name) should
also be refreshed. So as part of our communication that we are making our
change, we would like you to ask that they in return endorse a revised
version of the waiver (perhaps that's easiest if you update the text of the
existing one as appropriate for their approval).

Thanks!
Dermot

andrewharvey:
Thanks @mackerski, I'll request an updated waiver with the new attribution.

PR #3286, opened 2021-08-05
Fix broken `libpgosm.so`

hummeltech:
After the removal of [db/functions/maptile.c](https://github.com/openstreetmap/openstreetmap-website/commit/4e6d729529c94486a50cf135dfbbdcf841790200#diff-c3f52ef93a4aaf1f148e626e39ec31e4de87ce44154fbb7f5ea6e321c4e195a4L34-L36) in 4e6d729529c94486a50cf135dfbbdcf841790200, the `PG_MODULE_MAGIC` macro no longer gets called and therefore causes the compiled shared library to no longer be compatible. Resolved by creating a new `.c`/`.o` file whose only function is to call `PG_MODULE_MAGIC`.

Also:
  * Added `libpgosm.so` build to new `Tests (libpgosm.so)` workflow, here are links the relevant jobs:
    * Before:
      https://github.com/hummeltech/openstreetmap-website/runs/3256812366?check_suite_focus=true
    * After:
      https://github.com/hummeltech/openstreetmap-website/actions/runs/1103419155

### DIFF
--- a/.github/workflows/tests-libpgosm.yml
+++ b/.github/workflows/tests-libpgosm.yml
@@ -1,0 +1,107 @@
+---
+name: Tests (libpgosm.so)
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  test:
+    name: Ubuntu ${{ matrix.ubuntu }}, Ruby ${{ matrix.ruby }}
+    strategy:
+      matrix:
+        ubuntu: [18.04, 20.04]
+        ruby: [2.7, 3.0]
+    runs-on: ubuntu-${{ matrix.ubuntu }}
+    env:
+      RAILS_ENV: test
+      OPENSTREETMAP_MEMCACHE_SERVERS: 127.0.0.1
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2.3.4
+      - name: Setup ruby
+        uses: actions/setup-ruby@v1.1.3
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Cache gems
+        uses: actions/cache@v2.1.6
+        with:
+          path: vendor/bundle
+          key: bundle-ubuntu-${{ matrix.ubuntu }}-ruby-${{ matrix.ruby }}-${{ hashFiles('Gemfile.lock') }}
+          restore-keys: |
+            bundle-ubuntu-${{ matrix.ubuntu }}-ruby-${{ matrix.ruby }}-
+      - name: Cache node modules
+        uses: actions/cache@v2.1.6
+        with:
+          path: node_modules
+          key: yarn-ubuntu-${{ matrix.ubuntu }}-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            yarn-ubuntu-${{ matrix.ubuntu }}-
+      - name: Install packages
+        run: |
+          sudo apt-get -yqq update
+          sudo apt-get -yqq install memcached
+      - name: Install gems
+        run: |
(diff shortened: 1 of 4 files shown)


tomhughes:
I could have sworn I already fixed this... I just added it to one of the other files though rather than creating a new one.

It doesn't look like I ever sent it upstream though.

tomhughes:
I'm not convinced it warrants doubling the testing matrix though.

We should get rid of `xid_to_int4` now anyway, and I'm not sure the other one actually gets much use in our production deployment now that the API largely goes through cgimap.

hummeltech:
> I'm not convinced it warrants doubling the testing matrix though.

Yes, you are correct, however since they are run in parallel, the difference in actual duration at or near zero.

* Without the addition: (5m 51s)
  https://github.com/openstreetmap/openstreetmap-website/actions/runs/1101212774

* With the addition: (5m 31s)
https://github.com/hummeltech/openstreetmap-website/actions/runs/1103028358

tomhughes:
It's only partly about the run time, it's also about the huge level of extra complication it adds to the workflow definition...

hummeltech:
Yes, it's true. I do think it would be good to test it though, unless it is decided to remove it altogether. I created an issue for discussion #3288

hummeltech:
Let me know what you suggest here. I can just create an additional workflow that only runs on `push` to `master`, or only on `pull_request`, whichever you prefer (or if you have a better idea.)

Thanks!

hummeltech:
Alright, I've moved everything into a new workflow (`Tests (libpgosm.so)`) & made it only run on:
```YAML
  push:
    branches:
      - master
```
https://github.com/hummeltech/openstreetmap-website/actions/runs/1103419155

mmd-osm:
Do we still need the shared lib at all now? The remaining quad function has been moved to its own gem years ago.

mmd-osm:
Also see https://github.com/openstreetmap/openstreetmap-website/issues/3110

hummeltech:
There seem to still be at least some projects still relying on `xid_to_int4`:
https://github.com/openstreetmap/osmosis/blob/4b1af52a5d9a5eca76e0550b464ea4ce11260e71/osmosis-apidb/src/main/java/org/openstreetmap/osmosis/apidb/v0_6/impl/EntityDao.java#L374-L376

mmd-osm:
We have moved to osmdbt in the meantime. Osmosis is no longer used here in production for this purpose.

hummeltech:
Yeah, I'm not really sure about the status of `tile_for_point`. I know we still use `osmosis`, but I understand it is being moved away from.

mmd-osm:
No really, we don’t use this function in osmosis anymore.

hummeltech:
> No really, we don’t use this function in osmosis anymore.
Yes, I understand, I meant more the community at large.

> Tile for point is superseded by https://github.com/openstreetmap/quad_tile
Thanks for the link!

mmd-osm:
So the old osmosis style diff replication relied on a special db index for performance, that is no longer supported with Postgres 9.6+. Essentially that means the whole osmosis based replication story is dead by now. Almost all osmosis users will use the tool for other purposes, and they won't be affected by a missing xis_to_int4 db function.

hummeltech:
Thank you for the explanation, yes I understand and I agree. I was just answering your question.

mmd-osm:
Bottom line: we can probably close this pull request, as the db function code will go away anyway. No need to spend time fixing dead code.

hummeltech:
I agree!

PR #3291, opened 2021-08-12
Avoid using raw html in translations for links.

gravitystorm:
This is a followup to 26698d6 which introduced the html. It's better to use interpolation for links, since this avoids the translations introducing any html syntax errors.

I had to change the translation key, since changing the interpolation variables alone would lead to breakages.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -191,7 +191,7 @@
     <div class="col-md order-md-last">
       <% if @user.home_lat.nil? or @user.home_lon.nil? %>
         <div id="map" class="content_map">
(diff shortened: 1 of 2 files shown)


tomhughes:
I had deliberately merged the translations (given it was an HTML resource anyway) in order to avoid potential lego issues...

gravitystorm:
Lego translations are things like "foo" + "bar", i.e. don't give translators full flexibility to rearrange the words in the sentence. "%{foo} bar" can be rearranged into "baz %{foo}" if necessary, so that's not a lego translation and therefore causes no problems.

I had guessed you'd done it deliberately! But if it had been a PR I would have commented that the original approach is fine and is preferred to the raw html approach.

PR #3297, opened 2021-08-18
Refactor profile page and create a dashboard

gravitystorm:
The profile page was unusual for three main reasons:

* It was full-width, unlike any other non-map page
* The user's profile was put into the (grey) header, instead of being shown in the main part of the page
* The profile page added some not-really-related information when you viewed your own profile (like friendship activities) without any indication whether this information was public or not.

To resolve these three things I've refactored the profile page, so that it only shows the profile information and does so in the main content part instead of within the header. I've moved the other parts to a new Dashboard page, which I think in future can be expanded with more information that is useful for mappers to do their work.


tomhughes:
The obvious aesthetic issue with narrowing the profile is that even with a reasonable wide browser you get ugly wrapping:
image
The extra admin only line underneath is a bit ugly as well but I suspect that wouldn't look quite as bad if it wasn't for the wrapping and in any case that only affects a couple of users like me.

gravitystorm:
There's two different wrapping parts here:

* First, the administrator lines ("Deactivate... " and "Email address...") currently appear on a line on their own. But I could move these into the same column as the rest of the text, like this:

Screenshot from 2021-08-19 12-17-07

Would that be preferred?

* Second, the wrapping of the links isn't great, with the wrapped line starting with a separator bar. It's what happens to all of the existing `secondary-actions` stuff that we have elsewhere too. I'm not sure what the best approach is for improving the `secondary-actions` lists. Bootstrap offers [horizontal list groups](https://getbootstrap.com/docs/4.6/components/list-group/#horizontal) but I don't think they look right here. It also offers [horizontal navs](https://getbootstrap.com/docs/4.6/components/navs/#horizontal-alignment) which sort of work but I'm not sure it's an improvement:

Screenshot from 2021-08-19 12-35-17

We could tighten up the padding (and add back the count indicators, of course), but would this look alright? Or do we need to come up with something else for the `secondary-actions` lists to make sure they look better when wrapped?

The final line with "Email address..." also looks weird since it's sort of like the "Mapper since..." line but has different spacing and bold keys - probably best if they are consistent.

tomhughes:
The really annoying bit is the behaviour of the separator bars when it wraps, but it seems that is really hard to avoid... With a bit of effort I did manage to get this:

image

That is based on https://stackoverflow.com/a/31732902/805631 and uses this modified CSS for the secondary menu:

```css
ul.secondary-actions {
  font-style: normal;
  margin-bottom: 0;
  margin-left: 0;
  padding: 0;
  &.pager {
    display: inline-block;
    margin-right: 60px;
  }
  > li {
    display: inline;
    list-style: none;
    white-space: nowrap;
    &::before {
      content: " "
    }
    &::after {
      content: " ";
      whote-space: normal;
      word-spacing: 1em;
      background-image: linear-gradient(
        to right,
        transparent 0 calc(50% - 0.5px),
        $grey 0 calc(50% + 0.5px),
        transparent 0
      );
    }
    &:first-child {
      border-left: 0;
      padding-left: 0;
    }
    &:last-child {
      margin-right: 0px;
    }
  }
}
```

There's a nasty catch though, which is that you've have any whitespace before the `</li>` in the HTML so the view has to be modified in a nasty way - if you do have whitespace then it collapses with the whitespace in the `::after` and you don't get the background image :-(

tomhughes:
After some more playing I think the solution from https://stackoverflow.com/a/31732902/805631 is better as it avoids being sensitive to the white space in the page though it does require wrapping each list in another element.

I've opened https://github.com/gravitystorm/openstreetmap-website/pull/179 against your branch which implements that and gives me:

image

gravitystorm:
I've added your negative-margin-and-wrapper approach - tricksy, but I like it! I've also reworked the layout slightly, so that nothing wraps underneath the profile picture, and the two small+muted lines look the same.

Screenshot from 2021-09-15 13-20-14

I think this is ready to go now.

PR #3300, opened 2021-08-24
Add support for privileged OAuth 2 applications

tomhughes:
This adds a general concept of OAuth 2 scopes that are only available to applications created by administrators and which grant special privileges. It further adds two such scopes:

* `skip_authorization` which allows the application to be automatically authorised without asking the user to confirm the authorisation
* `read_email` which allows the application to receive the user's registered email address as part of a user details API response

The ultimate goal is to support single signon for other openstreetmap.org services and specifically for the https://community.openstreetmap.org/ discourse instance by allowing authentication without a confirmation screen and also allowing access to email addresses.


mmd-osm:
The overall approach to use a admin-only scopes seems pretty much in line with what I discussed with @Firefishy some time ago.

I was wondering a bit, if the current implementation is matching your sysadmin requirements. Say you want to restrict the user/details endpoint with email details to specific client IPs on Apache side.

Right now, this seems a bit difficult to accomplish. I could think of some URL parameter like `?include_email=true` for better visibility also in log files. Does that make sense?

mmd-osm:
I wanted to try this out on the community osm Discourse site with the email field included in the config.

For some reason, the authorize step on osm.org isn't working anymore. Error message says "The requested redirect uri is malformed or doesn't match client redirect URI.". Has there been some recent config change maybe?

tomhughes:
Won't you need @Firefishy to enable the new features on the OAuth application though? I assume he owns it?

tomhughes:
The URL that fails seems to be using http for the callback but the registered callback URL for the application is https.

mmd-osm:
@Firefishy fixed it, it was some setting in Discourse to force https. Fetching and updating users' email addresses also works as expected.

Issue #3256, opened 2021-07-13
Communities page


atomoil:
Hi,

I’m a member of the Local Chapters and Communities Working Group and this issue is to say hi and open a discussion over some work we’d like to undertake on the site (to be clear we intend to do the hands on dev work for this).

We’d like to add a “Communities” top level navigation link, which points to a new Communities page on the site. 

![Combined](https://user-images.githubusercontent.com/184608/125505558-629f88bb-dde9-411a-82ef-d11718273986.jpg)

The Communities page will list local chapters and provide links to the wiki for other groups. The local chapters list would ideally be data driven, and osm-community-index could be useful in this regard.

![Screenshot 2021-06-28 at 16 59 10](https://user-images.githubusercontent.com/184608/125504582-57c1731c-447c-41e3-a491-edf47368e145.jpg)

Looking to the future, when Microcosms is ready to be released there would be an additional block containing relevant microcosms with a link to find more (see mocked up images). This has been discussed with Brian who is leading Microcosms development.

![Screenshot 2021-06-28 at 17 08 45](https://user-images.githubusercontent.com/184608/125504586-7b14f163-b8df-4f27-8f0e-6c471aff2aa4.jpg)

We’re trying to keep the scope of this work small and targeted, so are not proposing to remove or replace other links or otherwise restructure the existing navigation (as that should probably include a review of existing usage and a wider consultation on changes which we don’t have the resource to do at this time).

This is my first PR on the site, so I’d particularly like to know any systems or patterns or technical things we should bear in mind to help long term maintenance and the eventual PR gain approval (eg is there a particular way you’d prefer we interface with osm-community-index or a better way of achieving this goal).

This piece of work has been discussed on the Local Chapters and Communities Working Group GitLab (https://gitlab.com/osmfoundation/lccwg/-/issues/14) so you can find more thinking and discussion there.

Thanks in advance for any thoughts or guidance you (or anyone reading this issue) can share.

Adam

tomhughes:
You might want to start by reviewing previous discussion about adding new top level links there - given how limited that space is you're going to need to make a very good case.

Examples of how it works in English aren't very useful - try it with something like German where the translations are much longer ;-)

How would a data driven solution work? I understand you have a geocoded index, but what would you use as the key? The current map position? For a map that's not longer displayed once you switch to the community tab? The user's home location?

atomoil:
Hi @tomhughes,

Thanks for the speedy reply!

[...]

PR #3301, opened 2021-08-24
Add Communities page

atomoil:
Here's a draft PR for the Communities page as discussed on https://github.com/openstreetmap/openstreetmap-website/issues/3256

I would love to get any and all thoughts / questions / concerns - particularly on:
* how it's integrating with the `osm-community-index` data
* and also on how to handle the localisation (for this draft it's manually pulling it from the osm-community-index i18n files, but is this the best approach?)

Tests are still to come, but so any thoughts there are appreciated, but I'd like to make sure the approach is ok first.

(apologies for the lack of capitalisation on some of the commits - I'm happy to fix that if it's a problem)


app/models/communities.rb
```diff
@@ -0,0 +1,59 @@
+class Communities
+  require "yaml"
+
+  @local_chapters = {}
+
+  def self.local_chapters(locale)
+    @local_chapters[locale] = local_chapter_for(locale)
```
tomhughes, line 7: If that is meant to be memoising the value then you want `||=` not just `=`.

app/models/communities.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  class << self
```
tomhughes, line 10: Why was `self.` good enough before but now you want to resort to black magic to define these class methods?

app/models/communities.rb
```diff
@@ -0,0 +1,59 @@
...
+    protected
```
tomhughes, line 11: Is this the reason for the black magic? Because you can't have a protected class method so you'd making them instance methods on the metaclass instead? Why do you want protected anyway - is something going to inherit from this?

app/models/communities.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def local_chapter_for(locale)
+      @local_chapters_index = load_local_chapters
```
tomhughes, line 14: See previous comment about memoisation.

app/controllers/site_controller.rb
```diff
@@ -106,6 +106,11 @@ def about
     @locale = params[:about_locale] || I18n.locale
   end
 
+  def communities
+    @locale = params[:communities_locale] || I18n.locale
```
tomhughes, line 110: Is language the correct way to be selecting a community? Shouldn't it be location driven?

app/models/communities.rb
```diff
@@ -0,0 +1,59 @@
+class Communities
```
tomhughes, line 1: Why is this a model? It doesn't seem to act like one... As far as I can see you never actually even create an object of this class... Rather the class methods that look like they should be constructors for the model are actually just returning plain hashes?

tomhughes (commented):
This is just a few high level thoughts - the view stuff looks reasonably sane but while I haven't tried to understand the model stuff in detail yet it all looks pretty odd to me.

atomoil:
@tomhughes apologies for the delay getting back to this - I've refactored the data stuff in a way that I _think_ is an improvement. At any rate I'm a bit less embarrassed by it now. I'm not sure if `lib` is the right place to put these, but it seems better than `models`. 

I'm still not really sure how to add tests to it, considering how dependent it is on the npm lib / data files, but I'll take a look at the tests that exist already and see if I can find any similar patterns to be inspired by.

Anyway I'd love to get any / all feedback you have (and that applies to anyone else reading this too 👋 )

mmd-osm:
Just posting a screenshot how this looks like so that others don't have to try out your code:

pr_3301_1

tomhughes:
I thought I had when this was first opened, and it should only need doing once, so not sure why it was asking again but I've approved it.

tomhughes:
I don't know what causes that test error - it just happens sometimes when they run in actions :-( Don't think I've ever reproduced it locally.

I've rerun them now and they all completed successfully.

I'll try and do a proper review soon...

atomoil:
hey @gravitystorm, I hope you are well. While we're waiting for @tomhughes to have some free time to review this PR, would it be possible to share any examples of moving the json+yaml parsing to an initialiser? The one thing I'm not sure about is how we would load the localisations - the code at that point doesn't know what localisations are supported and so loads them and memoizes them on demand, so I'd love to get any thoughts on how we could get that to work on startup instead.

Also - if it helps, I'm happy to do a call to discuss this PR, or possibly even meet up in person as I _think_ you are both in the UK - I'm outside London but can get there or Oxford or Birmingham very easily.

gravitystorm:
I had a look at this yesterday, but I ran out of time before having anything to share. For the initializers, we have [config/initializers/banners.rb](https://github.com/openstreetmap/openstreetmap-website/blob/master/config/initializers/banners.rb) that loads data from a yaml file (albeit it doesn't freeze the result, which we should really do for constants). But I got sidetracked investigating other approaches than using global constants.

I also had a brief look at ways of loading data into the i18n system, but again I don't have anything to share yet. I'll be working on both of these next time I get the chance (unless someone else beats me to it!).

mmd-osm:
Tbh, I 'm still a bit worried that we're introducing a new runtime dependency, when we've already have an existing approach to ingest third party data from yml files which even supports i18n

It would be a simple script that runs once every half year (or whenever there's a new local chapter), and updates some yml file we'd consume during runtime. I'm ok to extract the information from the JSON files as of today, but maybe can we do this at dev time only..

See https://github.com/openstreetmap/openstreetmap-website/tree/master/script/misc and https://github.com/openstreetmap/openstreetmap-website/blob/master/config/wiki_pages.yml for reference.

gravitystorm:
> I also had a brief look at ways of loading data into the i18n system, but again I don't have anything to share yet.

I think the answer to this bit is `I18n.backend.store_translations`, which is [described in the i18n documentation](https://guides.rubyonrails.org/i18n.html). For example:

```
I18n.backend.store_translations :en, inbox: {
  zero: 'no messages', # optional
  one: 'one message',
  other: '%{count} messages'
} 
```

> It would be a simple script that runs once every half year (or whenever there's a new local chapter), and updates some yml file we'd consume during runtime.

I'm not super-keen on this approach, even for the existing wiki_pages.yml file. It reminds me of the old "vendoring" approach to code dependencies, where we ended up with copies of source code in our repo even though we weren't maintaining it. The current wiki_pages approach does the same, but for data. As a rule of thumb, I don't like having files in the repo where [the commit history just consists of "copy information from external source"](https://github.com/openstreetmap/openstreetmap-website/commits/master/config/wiki_pages.yml) over and over again.

So I'd rather have a gem-like (or yarn-like) approach, where third-party (code / data / yml / javascripts / CSS / whatever) is maintained elsewhere rather than in-repo, and we pull in copies through our dependency managers.  Since the communities data is coming from a node module, I think using yarn to manage this dependency is the right approach.

My concern is then just to ensure that the community data gets loaded during application boot, and not at request time.

mmd-osm:
I'm not sure if you could compare a list of 17 local chapters with pulling in large amounts of code using the old "vendoring" approach. Besides the occasional wiki updates, we're also updating translations in pull requests that end up with plenty of " _Localisation updates from https://translatewiki.net._" commits. This isn't exactly the same use case, still somewhat related to be considered as another data point.

We haven't seen any unit tests for this pull request yet, but my impression is that is all adding complexity for data which isn't changing frequently by any means. My concern is that this adds long term maintenance burden. I'm happy to be proven wrong here, though. :)

atomoil:
> I think the answer to this bit is `I18n.backend.store_translations`, which is [described in the i18n documentation](https://guides.rubyonrails.org/i18n.html).

Amazing - `I18n.backend.store_translations` looks like exactly what I was hoping existed 🤩

atomoil:
I've refactored to use `I18n.backend.store_translations` and it's massively reduced the complexity of the code. I'm so much happier with it now.

app/controllers/site_controller.rb
```diff
@@ -106,6 +106,11 @@ def about
     @locale = params[:about_locale] || I18n.locale
   end
 
+  def communities
+    OsmCommunityIndex::LocalChapter.add_to_i18n # this should be called on app init
```
atomoil, line 110: One thing tho - I would really appreciate any advice on where to move this to, so that it's called on initialisation of the website.

lib/osm_community_index/local_chapter.rb
```diff
@@ -0,0 +1,59 @@
...
+          url = resource["strings"]["url"]
+          local_chapters.push(LocalChapter.new(id, url))
+        end
+      end
+      local_chapters
+    end
+
+    def self.add_to_i18n
+      community_index = OsmCommunityIndex.community_index
+      files = Dir.children(Rails.root.join("node_modules/osm-community-index/i18n/"))
```
gravitystorm, line 31: `Dir.glob(Rails.root.join("node....i18n/*")` (note the *) will return an array of file paths, so then you can skip the path rebuilding bit two lines further down.

lib/osm_community_index/local_chapter.rb
```diff
@@ -0,0 +1,59 @@
...
+      files.each do |file|
+        path = Rails.root.join("node_modules/osm-community-index/i18n/#{file}")
+        locale = File.basename(file,".yaml")
+        community_index_yaml = YAML.safe_load(File.read(path))[locale]
+        # rails wants en-GB but osm-community-index has en_GB
+        locale_rails = locale.split("_").join("-")
+
+        community_index["resources"].each do |id, resource|
+          resource.each do |key, value|
+            next unless key == "type" && value == "osm-lc" && id != "OSMF"
```
gravitystorm, line 41: This local chapter filtering is duplicating the logic elsewhere in the file. So I can see two choices:

* Either make another method that returns an array of local chapters (and maybe use `filter` / [`select`](https://ruby-doc.org/core-2.7.0/Enumerable.html#method-i-select) / `find_all` / similar rather than an `each` loop, since you want to transform an array into another smaller array)
* Or use the existing `self.local_chapters` to get the array of LocalChapter objects
* Alternatively, consider moving the i18n loading to the main community index file, and load the names for all the resources, rather than just local chapters.

lib/osm_community_index/local_chapter.rb
```diff
@@ -0,0 +1,59 @@
...
+            strings = community_index_yaml[id] || {}
+            # if the name isn't defined then fall back on community,
+            # as per discussion here: https://github.com/osmlab/osm-community-index/issues/483
+            strings['name'] = strings['name'] || resource["strings"]["name"] || resource["strings"]["community"]
+
+            data = {}
+            data["osm_community_index"] = {}
+            data["osm_community_index"]["local_chapter"] = {}
+            data["osm_community_index"]["local_chapter"][id] = strings
+            I18n.backend.store_translations locale_rails, data
```
gravitystorm, line 52: I would move this out of the inner loop, so that you only call `store_translations` once per locale. 

The hash can be built up with rails `Hash#deep_merge!`, so if you do `data = {}` before the loop, then inside the loop you could do 

```
data.deep_merge!({"osm_community_index" => {"local_chapter" => {id => strings}}})
```
and you should (hopefully!) get the structure you are looking for.

app/controllers/site_controller.rb
atomoil: @gravitystorm I'd still really appreciate your input on where this can be moved to

atomoil:
Hi @gravitystorm 

I've implemented all of the suggested changes 😎

Before I move it to a full PR I'd really appreciate your thoughts on:

* What kind of tests I should add (eg what can I look at for best practise / inspiration)
* Where to put the `OsmCommunityIndex::LocalChapter.add_to_i18n` that's currently in `site_controller.rb`

Also, I'd really love to get @tomhughes thoughts too.

atomoil:
hey @gravitystorm and @tomhughes,

It's been a couple of weeks, so I just wanted to see if you had any thoughts on the above so that I can progress this and get it closer to being finished. If it helps, then I'm happy to do a call or maybe even try to meet physically as I _think_ we are in the same part of the world.

lib/osm_community_index/local_chapter.rb
```diff
@@ -52,12 +53,7 @@ def self.add_to_i18n
           # as per discussion here: https://github.com/osmlab/osm-community-index/issues/483
           strings['name'] = strings['name'] || resource["strings"]["name"] || resource["strings"]["community"]
 
-          data = {}
-          data["osm_community_index"] = {}
-          data["osm_community_index"]["local_chapter"] = {}
-          data["osm_community_index"]["local_chapter"][id] = strings
-          I18n.backend.store_translations locale_rails, data
```
mmd-osm: Removing `I18n.backend.store_translations locale_rails, data`  breaks i18n.

gravitystorm:
> hey @gravitystorm and @tomhughes,
> 
> It's been a couple of weeks, so I just wanted to see if you had any thoughts on the above so that I can progress this and get it closer to being finished. 

Sorry, I haven't had as much time to work on openstreetmap-website in the last few weeks as I would have hoped. This is still on my radar but I can't give an ETA for review. I've been trying to prioritise the account deletion work, since getting that finished will also help free up some administrator time.

So please bear with me!

atomoil:
> Sorry, I haven't had as much time to work on openstreetmap-website in the last few weeks as I would have hoped. This is still on my radar but I can't give an ETA for review. I've been trying to prioritise the account deletion work, since getting that finished will also help free up some administrator time.
> 
> So please bear with me!

Hi @gravitystorm, I appreciate you're really busy, and I'm sorry to keep hassling you, but as almost a month has passed, I just thought I'd see if you or @tomhughes are any closer to feeding back on this PR?

lib/osm_community_index/local_chapter.rb
```diff
@@ -0,0 +1,64 @@
...
+  class LocalChapter
+    attr_reader :id, :url
+
+    def initialize(id, url)
+      @id = id
+      @url = url
+    end
+
+    def self.local_chapters
+      @chapters = init_local_chapters
```
gravitystorm, line 11: This has two problems - it's not actually memoising (should be `||=`) and it's using a class variable which is not thread safe. Although that doesn't matter here (perhaps), if we include it someone else will copy this approach elsewhere at some point.

lib/osm_community_index/local_chapter.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def self.init_local_chapters
+      raw_local_chapters = load_raw_local_chapters
+      local_chapters = []
+      raw_local_chapters.each do |chapter|
```
gravitystorm, line 17: `map` rather than `foo = [] ... each ... foo.push(bar)` to make this easier to read

lib/osm_community_index/local_chapter.rb
```diff
@@ -0,0 +1,64 @@
...
+        id = chapter[:id]
+        url = chapter[:resource]["strings"]["url"]
+        local_chapters.push(LocalChapter.new(id, url))
+      end
+      local_chapters
+    end
+
+    def self.load_raw_local_chapters
+      community_index = OsmCommunityIndex.community_index
+      raw_local_chapters = []
```
gravitystorm, line 27: `map ... .compact` rather than `foo = [] ... each ... unless x; foo.push(bar)` to make this easier to read. Or if we were on ruby 2.7+ (which we aren't yet) we could use [`Enumerable.filter_map`](https://blog.saeloun.com/2019/05/25/ruby-2-7-enumerable-filter-map.html)

lib/osm_community_index/local_chapter.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def self.add_to_i18n
+      raw_local_chapters = load_raw_local_chapters
+      files = Dir.glob(Rails.root.join("node_modules/osm-community-index/i18n/*"))
+      files.each do |file|
+        locale = File.basename(file, ".yaml")
+        community_index_yaml = YAML.safe_load(File.read(file))[locale]
+        # rails wants en-GB but osm-community-index has en_GB
+        locale_rails = locale.split("_").join("-")
```
gravitystorm, line 45: [`String#tr`](https://ruby-doc.org/core-3.1.1/String.html#method-i-tr) instead of `String.split(foo).join(bar)`

config/locales/en.yml
```diff
@@ -2224,6 +2225,30 @@ en:
...
+      local_chapters:
+        title: Local Chapters
+        about_text: |
+          Local Chapters are country-level or region-level groups that have taken the formal step of
+          establishing not-for-profit legal entities. They represent the area's map and mappers when
+          dealing with local government, business, and media. They have also formed an affiliation
+          with the OpenStreetMap Foundation (OSMF), giving them a link to the legal and copyright
+          governing body.
+        list_text: |
+          So far we have the following formally established foundation Local Chapters:
```
gravitystorm, line 2244: Avoid 'so far', prefer 'Currently' or similar

lib/osm_community_index/local_chapter.rb
tomhughes: We are on 2.7+ as far as I know - it's what INSTALL says and it's what rubocop is targetting.

gravitystorm: Good point, I didn't actually check

package.json
```diff
@@ -7,6 +7,7 @@
     "js-cookie": "^3.0.0",
     "leaflet": "^1.6.0",
     "leaflet.locatecontrol": "^0.75.0",
+    "osm-community-index": "^5.1.3",
```
mmd-osm, line 10: Also, current version is 5.1.6.

mmd-osm:
https://gitlab.com/osmfoundation/lccwg/-/issues/14#note_727900169 mentioned: "... _emailed the Communications working group to get their feedback on the text of the page._" back in Nov'21.

Did you receive any feedback, has this been incorporated already?

atomoil:
> I'll leave some review notes but I'm happy to implement these myself.

Thanks for taking the time to review @gravitystorm and thanks for implementing the changes yourself. Please don't hesitate to let me know if there's anything I can do to help. (I'm particularly interested what kind of tests would work best, given the dependency on the external files).

lib/osm_community_index/local_chapter.rb
atomoil: 🙄 I removed the memoising so I could test changes I was making, and once again forgot to add it back before pushing 🙄

config/locales/en.yml
atomoil: Sounds like a good improvement 😎

package.json
atomoil: The ^ at the start means compatible with (expecting semver), so running `npm i` would install 5.1.6
(for more info see: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies)

atomoil:
> https://gitlab.com/osmfoundation/lccwg/-/issues/14#note_727900169 mentioned: "... _emailed the Communications working group to get their feedback on the text of the page._" back in Nov'21.
> 
> Did you receive any feedback, has this been incorporated already?

There was no specific feedback on the copy itself. There was a note to

> Please add lots of HTML links, to help with SEO & to help people find out more. 

and later (when asked for examples) 

> It's not a blocker for me, when you write some text, I'll send a patch.

mmd-osm:
> I'm particularly interested what kind of tests would work best, given the dependency on the external files

osm-community-index uses mock data for tests. Maybe you can set up something similar using a minimal json with 2-3 LCs and maybe 2 languages and validate results.

tomhughes:
> > Please add lots of HTML links, to help with SEO & to help people find out more.

Relevant and appropriate links I hope - we're not in the business of adding links just to juice google results thank you very much.

gravitystorm:
I've covered most of what I want to do with this PR before the next review, but the big remaining issue is around thread safety. I want to make sure that everything is set up in the initializer, which runs before the separate threads for processing requests kicks in.

I think it actually all works fine at the moment, but I'm looking for a clearer way to show this and make it robust to refactoring. So for example, we aren't using any constants or any freeze/deep freeze on the reading the json file, and so leaving things mutable opens the door for problems later. Also this file reading is currently done via the initializer, but only as a side effect of loading the translations, so that's not super obvious or robust.

I've been struggling to find good guidance on what we should be doing here, so if anyone has any rails knowledge about how to do some processing in an initializer and making the results available as a constant, and ideally in a way that doesn't raise red flags about globals and thread-unsafe class variables etc, then let me know! I'm not sure if there's anything available via activesupport that would help, or guidance to use the rails configuration system to store the data, or a gem that covers this, or anything like that.

atomoil:
hey @gravitystorm thanks for all of your work on this! I've only just spotted your message from a couple of weeks ago. I am asking around to see if I know anyone who can help on this, so _hopefully_ more soon!

lib/osm_community_index.rb
```diff
@@ -0,0 +1,12 @@
+module OsmCommunityIndex
+  require "yaml"
+
+  def self.community_index
+    @community_index ||= community_index_from_json
+  end
+
+  def self.community_index_from_json
+    json_file = Rails.root.join("node_modules/osm-community-index/dist/resources.json")
```
openbrian, line 9: Is this the best way to link the dataset?  Does OCI have an API we could call?

openbrian: Thinking about it a bit, that would be more complicated to sync.  Maybe OSM should be the system of record for communities.

gravitystorm:
>  or a gem that covers this

I've come across https://github.com/byroot/frozen_record today, which is really interesting - creating read-only models from yaml and json files, so at a high level it's what we're looking for. I'll have a look and think about whether it will help, since we don't have a one-to-one mapping between the layout of the data in the json files and the structure of the data that we need. But it could be promising.

lib/osm_community_index.rb
atomoil: Thanks for taking a look @openbrian (and sorry for the delay replying 😬). My intention with using Open Community Index was to follow the "single source of truth" principle. OCI already have the data we need to render the page on osm.org, so seems rude not to try to use it and share data across the OSM ecosystem. I do appreciate that a shorter-term solution of just hardcoding the data would have been easier to achieve, and we could always fall back to that if this approach doesn't pan out.

> Maybe OSM should be the system of record for communities.

Do you mean OCI or do you have something else in mind (if so I'd love to hear what you're thinking)?

atomoil:
> > or a gem that covers this
> 
> I've come across https://github.com/byroot/frozen_record today, which is really interesting - creating read-only models from yaml and json files, so at a high level it's what we're looking for. I'll have a look and think about whether it will help, since we don't have a one-to-one mapping between the layout of the data in the json files and the structure of the data that we need. But it could be promising.

ooo yes this looks very interesting. This is exactly what I was hoping existed (creating rails-compatible data models but using a static file instead of a database). Open Community Index has far more data than we need to model, but from reading that link it sounds like a custom backend could be used to produce smaller records. It _should_ be a case of shifting the existing code to a different context, so in theory shouldn't be a big job. I am away until start of August but I'm happy to take a look when I'm back.

gravitystorm:
I've added a commit to this branch which switches to FrozenRecord, and it seems to work well. In particular, I like being able to do things like:

```ruby 
@local_chapters = Community.where(:type => "osm-lc").where.not(:id => "OSMF")
```
... it feels much more rails-y this way, and makes it easier to expand in future. It also chops out a lot of parsing code - loading the communities is now only two lines of our own code, and the only significant processing logic that remains is loading the i18n.

I took the approach that the correct model is a "community" (not a local chapter) and that local chapters can be selected from the full list of communities, as above. My hope is that it makes it easier to implement other things in future, for example we could perhaps show more communities based on your language, or something like that?

My next tasks will be:
* Consider generalising the i18n parsing, so that it works for all communities and not just local chapters
* Review the communities page lede, since I've just been restyling those elsewhere
* Consider whether we should be using accessor methods (like I've done with `url`), whether that should instead be done during parsing, or whether we should just leave the raw data structure as-is (e.g. use `strings["url"]` in the views)

Any comments on all this are very welcome!

gravitystorm:
OK, I think I've covered everything that I want to here. Given how much coding I've now done, I think it's moved beyond me "reviewing" this PR. I'll ask @tomhughes to review it, to avoid me approving my own work.

atomoil:
This is great, thanks so much @gravitystorm

I also hope this can get extended in the future so I like the use of community over local chapter as that should make it clearer.

I hope we can get Tom to take a look and move this closer to going live.

Thanks so much for your time and dedication!

lib/osm_community_index.rb
```diff
@@ -0,0 +1,26 @@
+module OsmCommunityIndex
+  def self.add_to_i18n
+    communities = Community.all
+    files = Dir.glob(Rails.root.join("node_modules/osm-community-index/i18n/*"))
```
tomhughes, line 4: I would make that match `*.yaml` as an extra protection.

lib/osm_community_index.rb
```diff
@@ -0,0 +1,26 @@
...
+    files.each do |file|
+      locale = File.basename(file, ".yaml")
+      community_locale_yaml = YAML.safe_load(File.read(file))[locale]
+      # rails wants en-GB but osm-community-index has en_GB
+      locale_rails = locale.tr("_", "-")
```
tomhughes, line 9: Presumably this affects more than just `en-GB` so maybe the comment should be more generic, like `language-COUNTRY` or something?

Are we sure this is the only mismatch? Has anybody done a complete cross check between our language list and the OCI one?

lib/osm_community_index.rb
```diff
@@ -0,0 +1,26 @@
...
+      data = {}
+
+      communities.each do |community|
+        id = community[:id]
+
+        strings = community_locale_yaml[id] || {}
+        # if the name isn't defined then fall back on community,
+        # as per discussion here: https://github.com/osmlab/osm-community-index/issues/483
+        strings["name"] = strings["name"] || community["strings"]["name"] || community["strings"]["community"]
```
tomhughes, line 18: We should be consistent in whether we index `community` using symbols or strings - we use strings here but a symbol a few lines above.

In fact given community is a model can't we use member syntax? So `community.id` and `community.strings`?

lib/osm_community_index.rb
```diff
@@ -0,0 +1,26 @@
...
+    communities = Community.all
+    files = Dir.glob(Rails.root.join("node_modules/osm-community-index/i18n/*"))
+    files.each do |file|
+      locale = File.basename(file, ".yaml")
+      community_locale_yaml = YAML.safe_load(File.read(file))[locale]
+      # rails wants en-GB but osm-community-index has en_GB
+      locale_rails = locale.tr("_", "-")
+      data = {}
+
+      communities.each do |community|
```
tomhughes, line 12: I'd use `each_with_object` here rather than initialising `data` before and then using it in the loop.

lib/osm_community_index.rb
```diff
@@ -0,0 +1,26 @@
...
+        id = community[:id]
+
+        strings = community_locale_yaml[id] || {}
+        # if the name isn't defined then fall back on community,
+        # as per discussion here: https://github.com/osmlab/osm-community-index/issues/483
+        strings["name"] = strings["name"] || community["strings"]["name"] || community["strings"]["community"]
+
+        data.deep_merge!({ "osm_community_index" => { "communities" => { id => strings } } })
```
tomhughes, line 20: Does this actually need the outermost `{}` around the hash?

lib/osm_community_index.rb
```diff
@@ -0,0 +1,26 @@
+module OsmCommunityIndex
+  def self.add_to_i18n
+    communities = Community.all
```
tomhughes, line 3: Tricky question, but should we limit this using the same filter we use when choosing what to display? It would vastly limit how many translations we are pulling in...

tomhughes (changes requested):
Beyond my specific technical points that I've raised I guess my main question is whether loading about 680 records and potentially 37000 translations (currently) is a reasonable way of displaying a list of 17 local chapters...

lib/osm_community_index.rb
gravitystorm: Done.

lib/osm_community_index.rb
gravitystorm: Updated the comment, and move the complete cross check discussion to #3653

lib/osm_community_index.rb
gravitystorm: Updated to use `community.id` etc

lib/osm_community_index.rb
gravitystorm: That's much nicer - done.

lib/osm_community_index.rb
gravitystorm: Removed

lib/osm_community_index.rb
gravitystorm: Fair point. I'm happy to limit them for now, and the filter can be expanded if/when we start using other communities elsewhere on the site.

mmd-osm:
OCI and official names on https://wiki.osmfoundation.org/wiki/Local_Chapters differ in some cases, eg. Switzerland. 

As the text talks about legal entities, it would be good to use the official names, rather than some generic term.

atomoil:
> OCI and official names on https://wiki.osmfoundation.org/wiki/Local_Chapters differ in some cases, eg. Switzerland.
> 
> As the text talks about legal entities, it would be good to use the official names, rather than some generic term.

hey @mmd-osm thanks for spotting this. I'm going to work through it with the Local Chapters and Communities Working Group and make sure both are correct and aligned (in terms of legal names, the UK chapter's name on the wiki isn't the legal name) and if we don't want to list legal names then I'll get the copy within this PR amended to remove the issue.

atomoil:
As a follow-up on the note about community names:

After discussion with the local chapters and communities working group, I have slightly changed the text to highlight the name of the community over some idea of using a legal name, which isn’t so useful to the intended audience.

It is important to use the name the community uses to identify itself. In the case of Switzerland, it seems like different channels use different names, eg Twitter (https://twitter.com/SwissOSM) uses “Swiss OpenStreetMap” and the website (https://sosm.ch/) uses “Swiss OpenStreetMap Association”. That’s up to the community to decide on, and they can do this within osm-community-index, which would then lead to consistency with all the places that use osm-community-index.

atomoil:
@gravitystorm I could be wrong, but it _looks_ like you've done all of the things that you noted and that @tomhughes requested to be changed, so I've switched this from a draft to a full PR and fixed the 3 conflicts the Github highlighted. (I seem to be fighting with GitHub to re-request a review from both of you, so apologies if that's ended up spamming your inboxes)>

I know the Local Chapters and Communities Working Group is keen to get this PR merged in, so please don't hesitate to let me (or any of them) know what we can do to move this to a place where you're happy to merge it and put it into production.

Thanks again for all of your input on this. I'm really happy with FrozenRecord and generally where we've gotten to with it, and I sincerely hope the foundations that this provides can be built on in the future.

app/views/site/communities.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<% end %>
+
+<p class="lead"><%= t ".lede_text" %></p>
+
+<h2><%= t ".local_chapters.title" %></h2>
+<p><%= t ".local_chapters.about_text" %></p>
+<p><%= t ".local_chapters.list_text" %></p>
+<ul>
+  <% @local_chapters.each do |chapter| %>
+    <li><a href="<%= chapter.url %>"><%= t "osm_community_index.communities.#{chapter.id}.name" %></a></li>
```
openbrian, line 12: Might be susceptible to supply chain XSS attack.

tomhughes: Do you want to tell us what a "supply chain XSS attack" is, how you think this is vulnerable, and how your proposed fix helps?

gravitystorm: A supply chain XSS attack would be where something we rely on but don't control (in this case, the community index repo, and the translations therein) has nefarious html that we incorporate into our site. 

But the rails translation system will automatically escape any html found in the translation strings, unless steps are taken to output the translated string in raw format (e.g. by appending `_html` to the key). So unless I'm wrong, and it which case I'd love to see a more detailed explanation, a PoC, or steps to recreate, then this is fine and no different from what we do already with our own translations.

mmd-osm: I think chapter.url as in `href="<%= chapter.url %>"` is taken from the original JSON files, without going through any translation step, no?

gravitystorm: Good point!

openbrian: Yeah, Brakeman just called me out on this issue on my other PRL.  I did a little research on it.  Keep in mind the url could start with "javaScript:" or "data:".  The solution I did, which satisfied Brakeman, is "https://" + url.slice!(”https://"). I can fetch the error later, on a phone here.

tomhughes: But just changing the protocol is nonsense - surely the correct solution is to have the code that loads the data reject any entries with URLs it doesn't like?

openbrian: In the mirocosm PR, I use https://rubygems.org/gems/validate_url and add validation in the model 
```
  validates :url, :presence => true, :length => 1..255, :url => { :schemes => ["https"] }
```
Not sure that will work here.  I agree.  Prevent bad data from entering the app.  But the validation alone would not appease Brakeman.  Seems like it needed to see a static and trusted string there.  Shrug.  Just to be clear the issue is with the href, not the text.

For reference here's the finding.  https://github.com/openbrian/osm-microcosms/runs/8294617963?check_suite_focus=true#step:4:30

seav:
This PR has been linked from a United Nations SDG web page: https://ideas.unite.un.org/sdg11/Page/Challenge6

tomhughes:
Is that supposed to excite us? That some idiotic third party has taken it upon themselves to try and drive our development process without so much as consulting us?

Frankly I've lost track of what the current status of this is... I see it says it's waiting on my review but I thought I had reviewed it.

seav:
> Is that supposed to excite us?

I don't think so. I just mentioned it as an FYI so that if new people joined the conversation here then this can hopefully explain why.

> That some idiotic third party has taken it upon themselves to try and drive our development process without so much as consulting us?

Maybe I'm just dense but I didn't really see anything on the page that implied that they know a better development process for improving the website than the maintainers? In fact, they caution: "Remember that the website maintainers are volunteers." (Well except for @Firefishy, of course.)

tomhughes:
But why on earth would they seek to people to work on this project, which already has a PR that is well on it's way to being merged and a developer working on the feature?

Maybe you also missed the bit where they say that they are going to review the PRs!

In any case it's off topic for this ticket so we should move discussion elsewhere if discussion is necessary.

mmd-osm:
https://ideas.unite.un.org/sdg11/Page/Overview suggests the submission deadline is end of October already. That's less than 4 weeks...

grischard:
The UN and EU people of course talked to me, and I should have talked to Tom. This PR was linked from the challenge as an example of how we work and how the website works.

app/views/site/communities.html.erb
tomhughes: No because the HTML tags in that would be automatically escaped.

atomoil: @gravitystorm is this the only thing blocking getting this PR approved and merged? I'm happy to do the leg work on implementing whatever needs to be done, but I might need some pointers on where to look and what to do. If there are other things, then please let me know, as I am very keen to get some closure on this (I never expected it to take over a year!)

gravitystorm: @atomoil Yes, as far as I remember that's the only thing now. I'm sorry it's taken so long to get this over the line, I've not had any time to work on openstreetmap-website in the last few weeks. I want to take a look at this url validation, and consider it alongside our existing validation and the gem that's suggested above, so that we don't end up with three different approaches to validation.

atomoil: @gravitystorm ironically I put together some code to do this last night but ran out of time to test it. I’m not precious about it and more than happy to bin it in favour of something that is consistent with the rest of the site. Saying that - ideally any validation would not reject an entry with a missing/invalid URL, it would still list the chapter but with either a generic URL (eg the wiki page on local chapters) or no link (anchor tag).

gravitystorm:
I worked on the validation topic today, but I don't have something that I'm happy with.

I thought of two approaches. One is to write custom validation code in `ResourceBackend.load`, e.g. that uses a regexp to check the URLs are OK. But then when I started choosing schemes, and doing blank handling, then this feels a bit cludgy since we normally put attribute validation into the models. Reinventing validations feels wrong. If we can declare the validation in the model, then we can do a normal `validates...` declaration instead. So I tried that approach instead, and used the `validate_url` gem too. You can see my commit at https://github.com/gravitystorm/openstreetmap-website/commit/7b5df0daaf7b5a230f180ec914b02db66bfa2fee which I haven't added to this PR.

This threw up some issues:
* FrozenRecord doesn't have anything where it checks validity, nor can I see any way to crowbar that into the loading of each record. So instead I fail loudly during object initialization if there's an invalid URL. This should be fine, in theory, since we'll only update to versions of the node package from time to time, and if there's one with invalid urls then I think it's fine for the tests to fail (and the app to not boot). 
* OCI already contains various schemes, not just http and https but also mailto and xmpp. I wasn't expecting that.
* OCI also contains invalid urls :facepalm: For example, `https://matrix.to/#/#openstreetmap-cat:matrix.org` is [invalid](https://stackoverflow.com/a/26119120/105451) (albeit accepted by browsers) since the `#` isn't permitted inside fragments.

The last of these is a bit of a hard fail for this approach, since we have to deal with what OCI gives us, and this is not a malicious URL and failing to handle it gracefully is not reasonable.

So trying to do things "the right way" hasn't worked as well as I hoped. If anyone has any suggestions to build on this approach then I'm all ears, otherwise my next attempt will be to "reinvent validations" by hand in the `.load` method (sigh).

pnorman:
> * OCI already contains various schemes, not just http and https but also mailto and xmpp. I wasn't expecting that.

Is that an issue for validation? They should all be valid URLs

> * OCI also contains invalid urls 🤦 For example, `https://matrix.to/#/#openstreetmap-cat:matrix.org` is [invalid](https://stackoverflow.com/a/26119120/105451) (albeit accepted by browsers) since the `#` isn't permitted inside fragments.

I think this is an OCI bug. One of the `#` should be percent-encoded, and I've opened https://github.com/osmlab/osm-community-index/issues/592 about it.

gravitystorm:
> Is that an issue for validation? They should all be valid URLs

The issue is more about maintaining the list of schemes (more likely to need revisions than if it was just http(s)), and what to do with ones we don't want to display (mailto, xmpp), what to do with ones we haven't yet allowed, whether to strip them out, drop the record, emit a warning during parsing etc. If it was just http/https/nil that was expected, then the out-of-course handling would need less careful thought.

> I think this is an OCI bug. One of the `#` should be percent-encoded, and I've opened [osmlab/osm-community-index#592](https://github.com/osmlab/osm-community-index/issues/592) about it.

Thanks. It's worth noting that our two approaches to URL validation (`validate_url` gem, `URI::DEFAULT_PARSER.make_regexp`) behave differently - one accepts the second `#`, one doesn't! So I don't want to imply too much that it's a bug in OCI, rather than just something we need to deal with ourselves.

gravitystorm:
I've worked on this again today, and after more false starts with the `validates_url` gem, I've now got the validation working and I'm (mostly) happy with it.

@tomhughes This is done from my POV, please let me know your comments.

tomhughes (approved):
Looks good to me from a technical point of view.

mmd-osm:
I just checked the new community page using a locale parameter as shown below. It seems translations for the 17 local chapter names are only working in 4 out of 17 cases.

- https://www.openstreetmap.org/communities?locale=es
- https://www.openstreetmap.org/communities?locale=it
- https://www.openstreetmap.org/communities?locale=de
- https://www.openstreetmap.org/communities?locale=ar

As we've discussed earlier on, I still assume the community index is the original source for translations (rather than translatewiki). Also I'd expect more translations to be available, but I have no idea why that's not the case.

gravitystorm:
> Also I'd expect more translations to be available, but I have no idea why that's not the case.

Following up this in #3814

Issue #3312, opened 2021-09-06
Docker db problem


lectrician1:
I followed all of the instructions in DOCKER.md and started both the web and db containers, but I get this error when first visiting the website: 
```
config/initializers/cors.rb:9:in `call'

Started GET "/" for 172.18.0.1 at 2021-09-05 15:10:06 +0000

  

ActiveRecord::NoDatabaseError (FATAL:  role "openstreetmap" does not exist

):
```

And therefore it cannot start. I just need some help to get it working.

tomhughes:
Well unfortunately none of us knows anything about docker so you're probably not going to find much help here...

tomhughes:
In any case there's pretty much nothing anybody can do with that tiny fragment of information - whatever the actual problem is happened long before that so we need to see the log from when the database was built and populated.

tomhughes:
Specifically where the database container executed the `docker/postgres/openstreetmap-postgres-init.sh` script which does:

```
psql -v ON_ERROR_STOP=1 -U "$POSTGRES_USER" <<-EOSQL
    CREATE USER openstreetmap SUPERUSER PASSWORD 'openstreetmap';
    GRANT ALL PRIVILEGES ON DATABASE openstreetmap TO openstreetmap;
EOSQL
```

lectrician1:
This first line here seems to be the problem. 
```
/usr/local/bin/docker-entrypoint.sh: /docker-entrypoint-initdb.d/openstreetmap-postgres-init.sh: /bin/bash^M: bad interpreter: No such file or directory

2021-09-05 15:04:35.412 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432

2021-09-05 15:04:35.412 UTC [1] LOG:  listening on IPv6 address "::", port 5432

2021-09-05 15:04:35.417 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"

2021-09-05 15:04:35.430 UTC [26] LOG:  database system was interrupted; last known up at 2021-09-05 14:58:13 UTC

2021-09-05 15:04:35.686 UTC [26] LOG:  database system was not properly shut down; automatic recovery in progress

2021-09-05 15:04:35.688 UTC [26] LOG:  redo starts at 0/16506A0

2021-09-05 15:04:35.688 UTC [26] LOG:  invalid record length at 0/1650790: wanted 24, got 0

[...]

PR #3313, opened 2021-09-07
Add gitattribute to fix line endings

Firefishy:
Force line endings using a `.gitattribute` file and and `.editorconfig` file

Likely fix for: #3312

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


tomhughes:
I looked into that earlier but I couldn't figure out the correct settings for what we want, or even if there is actually a way to do it.

tomhughes:
Naturally I didn't think of trying to enumerate everything that might be text because that would be crazy...

Firefishy:
> Naturally I didn't think of trying to enumerate everything that might be text because that would be crazy...
`* text=auto eol=lf` would work reasonably well if everyone is using git >=2.10.0

Firefishy:
https://github.com/git/git/blob/master/Documentation/RelNotes/2.10.0.txt#L248

Firefishy:
Interesting example from Drupal: https://git.drupalcode.org/project/drupal/blob/HEAD/.gitattributes

tomhughes:
I switched it to autodetection in b11957a3957282de3949fb80e4d6cf4576eadc1d as git 2.10 is five years old now so I doubt it's a significant issue.

PR #3318, opened 2021-09-09
Added a couple of keys under `natural` to en.yml

mortenbruhn:
As the title suggests.

See also:
 
- #3251
- #3142
- #3124


mortenbruhn:
Also added `crossing`.

PR #3323, opened 2021-09-15
Temporary pin on rubocop-rails until next bugfix release

gravitystorm:
There is a bug in the Rails/ContentTag cop that causes rubocop to crash. The bug has already been fixed and so should be available in the next release.

See https://github.com/rubocop/rubocop-rails/issues/543

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -145,7 +145,7 @@ group :test do
   gem "rubocop"
   gem "rubocop-minitest"
   gem "rubocop-performance"
-  gem "rubocop-rails"
+  gem "rubocop-rails", "~> 2.11.3" # Bug in 2.12.0, fixed in master after 2.12.2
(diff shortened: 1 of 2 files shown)


tomhughes:
I already did the upgrade yesterday, and disabled that test for the problem files because of that bug.

gravitystorm:
Ah, I didn't notice the disabled file in `.rubocop_todo.yml`. But there's only one file in that list, and CI on master is failing due to the others.

tomhughes:
I didn't actually see the crash, what I saw was it trying to change `Locale.tag` calls even though that is nothing to do with AR classes.

I'm still not actually clear if the upstream fix is actually going to stop that as they seem to have concentrated on "top level" rather than on only looking at AR derived classes.

PR #3384, opened 2021-11-26
bug issue template: removed browsers

HolgerJeromin:


tomhughes:
Mobile is tricky - possibly we want `iOS` as an option and then `Other Mobile` with chrome and firefox options? Because people who think they're using chrome or firefox on iOS are really using safari with a different shell.

gravitystorm:
I'd much rather keep the list short, and importantly only use consumer-facing names. It's up to us, as the technical people, to know the details about which rendering engines are common to which browsers and I don't think we should be making that part of the decisions.

If the template doesn't narrow it down sufficiently, then we can always ask for more details.

I'm happy to add Other to the list.

But in general, we have so few browser-dependent bugs (compared with iD, which the list is copied from) that I could be more easily persuaded to just remove the list entirely, since it's not relevant to almost all the issues created in this repo.

HolgerJeromin:
I stumbled over the only company name which I wanted to align. 
Then the list started to get bigger and bigger. :-)
But perhaps you are right and we should get rid of it.

PR #3386, opened 2021-11-29
Notes ordering: show unresolved notes first

mnalis:
This simple PR changes ordering when showing user notes on the public website (`https://www.openstreetmap.org/user/USERNAME/notes`)

Previously, notes were sorted by date of last update (newest first), regardless of their status `open`/`closed`.

With this, it displays `open` notes first (as that is most likely what the user is interested in), followed by all `closed` notes.
Apart from that, the notes remain sorted as before - by date of last update (newest first).

While not completely solving it, this should alleviate most grave problems of https://github.com/openstreetmap/openstreetmap-website/issues/832#issuecomment-829383983

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -20,7 +20,7 @@ class NotesController < ApplicationController
         @page_size = 10
         @notes = @user.notes
         @notes = @notes.visible unless current_user&.moderator?
-        @notes = @notes.order("updated_at DESC, id").distinct.offset((@page - 1) * @page_size).limit(@page_size).preload(:comments => :author)
+        @notes = @notes.order("status, updated_at DESC, id").distinct.offset((@page - 1) * @page_size).limit(@page_size).preload(:comments => :author)
       else
         @title = t "users.no_such_user.title"
         @not_found_user = params[:display_name]


mnalis:
As there were not any negative comments either in this simple PR nor in linked https://github.com/openstreetmap/openstreetmap-website/issues/832#issuecomment-829383983 after several months, perhaps this could be considered for merging @gravitystorm? (or any feedback, of course, if you think it's not ready!)

tomhughes:
Well having just been looking at this query for #3443 the fact that we're ordering on `notes` but selecting on `note_comments` is not ideal from a performance point of view given that we have a limit as we have to find all the records and order them before applying the limit.

This change doesn't actually change that though as the problem already exists so it's not a reason not to do this.

More concerning probably is that it won't be obvious to users what the ordering is given the statuses are all grouped together.

Dimitar5555:
Would it be feasible to separate them in two different tabs (Open and Closed) like it's done in the messages page?

mnalis:
@tomhughes I would think most users are interested in `open` (unresolved) notes (which they likely want to solve), which would be displayed first by this PR, so IMHO there would be less confusion (as users would be shown firstly what there are most likely interested in). 
Also, visual cues (first all red, then all green) should help few users who are looking for their closed notes figure out new ordering. 

(One could verify that theory by correlating web-server logged user clicks on notes by their status, but that is quite likely overkill.)

Although if you still think users might be confused by improved ordering, we can update the text shown, something like:

```diff
-Notes submitted or commented on by %{user}
+Notes submitted or commented on by %{user} are ordered by first showing all open notes, followed by closed notes
```

That would help keep the patch simple and correct. Would you prefer that, @tomhughes ?

mnalis:
Anyway, as nowadays one can filter notes by status (thus forcing only unresolved notes to show), the need for this has diminished significantly for me, so I'll close it.

PR #3391, opened 2021-12-06
Added `highway=turning_circle` to en.yml

mortenbruhn:
Added `highway=turning_circle` to en.yml - value was not available in TranslateWiki for translation.

A precedent has already been set that labels found through the "Find objects" tool should be made available for translation too (see discussion on topic in #3124)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -829,6 +829,7 @@ en:
           trailhead: "Trailhead"
           trunk: "Trunk Road"
           trunk_link: "Trunk Road"
+          turning_circle: "Turning Circle"
           turning_loop: "Turning Loop"
           unclassified: "Unclassified Road"
           "yes" : "Road"


tomhughes:
What is the relevance of those other PRs? Do you plan to add an ever growing list of previous changes to every PR you make?

When you open a PR the description should be explaining the change the PR is making and why you think it would an enhancement, not just listing previous changes you have made.

mortenbruhn:
Duly noted. 

I was trying to reference a precedent that these key-value pairs in `en.yml` should usually just be merged. The references were to try to save time. I see now that the description i ended up doing only ended up saving time for me, not the reviewers. My apologies - will amend.

mortenbruhn:
@tomhughes desciption amended (see previous comment).

Thanks for quick review.

Issue #3123, opened 2021-03-05
Search forms redesign


mjourdan:
On www.openstreetmap.org, the search forms (both single location and directions) have a few issues...

## Problems

- picking up a search results clears the search field
- pressing the "directions" icon also clears the search field
- button to reverse directions is visually disconnected from directions, and placed after the "go" button
- the routing engines are displayed prominently, while they are technical details most people won't understand
- input fields lacks of readability once locations are found
- search mode and destination mode feel disconnected (forms have different width and no clear alignment)
- lots of gray area eating estate from the map

Also, maybe adding some extra features could satisfy common expectations, like:
- multiple destinations
- public transports

## Screenshots

![current search forms](https://user-images.githubusercontent.com/5346620/110166578-dee56580-7df4-11eb-893a-be4e41020421.png)

mjourdan:
## Proposal

Please find below two static mockups. The first one only aims to fix the issues previously mentionned, while the second one also adds the features mentionned:

![image](https://user-images.githubusercontent.com/5346620/110166114-2c150780-7df4-11eb-9684-a9aa53945e12.png)

Feel free to tell what you think!

tomhughes:
Just having mode selectors mean we have to make a policy decision to prefer one engine over another which whilst it may be good UX is bad politics.

You should also be aware that routing only exists to allow for connectivity testing - it is explicitly not a goal to provide an end user google maps replacement.

systemed:
> public transports

OSM doesn't store timetable information so can't really provide a public transport routing service. See https://help.openstreetmap.org/questions/78247/transport-options .

mjourdan:
There must be other ways to play good politics and still improve the ux. For example, the mockups above could allow sending requests alternately to one provider or the other, unless people picked their choice in the options.

Regarding the expectations about the routing service (and more widely about the website), the thing is OpenStreetMap is sometimes advertised as a privacy-friendly alternative to Google Maps. See [here](https://www.wired.com/story/the-best-privacy-friendly-alternatives-to-google-maps/) or [there](https://restoreprivacy.com/google-alternatives/). I know this is inaccurate, still it contributes to forge people expectations. 

As of public transports routing, the wiki also highlights OSM is not meant to store timetables information, but also states OSM [could provide basic routing services](https://wiki.openstreetmap.org/wiki/Public_transport). Still if such service provider existed, it could make sense to integrate here. Nothing mandatory though.

gravitystorm:
[...]

PR #3400, opened 2021-12-18
WIP: search forms redesign

mjourdan:
This branch is have an example of how #3123 could be implemented. Here is a screenshot:

image


gravitystorm:
Thanks for the mockup. I didn't realise until reviewing that it's a mockup rather than a proposal (i.e. there's a hardcoded options list, none of the buttons work) so I've converted this to a draft.

I think this highlights that we need to change the secondary button colour, since it's quite a dark shade of grey compared to our blue and green colours.

I don't think primary is the correct choice for the type-of-route buttons, since that's not actually an action on the form, rather it's a toggle choice. I would like to keep the primary colour for the main action (i.e. "click here to make something happen now").

* Talking of which, there's no button to make the routing happen now. Is that deliberate?
* How do you propose to handle situations where there are either more or fewer types of routing available? Different engines might have different lists of options (e.g. off-road cycling, skating, only car etc).
* Where do the icons come from? There's no attribution afaict. 
* The icons seem a bit inconsistent e.g. the weight of the arrows compared to the weight of the cross symbol. Do they all come from the same icon set?

(For icon sets, perhaps https://icons.getbootstrap.com/ would be useful.)

mjourdan:
Yes, this was meant as a prototype rather than a fully fledged implementation. Thanks for taking a look at it.

> Talking of which, there's no button to make the routing happen now. Is that deliberate?

Yes, filling in the two fields and pressing enter should be sufficient to get the results

> How do you propose to handle situations where there are either more or fewer types of routing available? Different engines might have different lists of options (e.g. off-road cycling, skating, only car etc).

The options would be updated according to the selected engine.

> Where do the icons come from? There's no attribution afaict.
> The icons seem a bit inconsistent e.g. the weight of the arrows compared to the weight of the cross symbol. Do they all come from the same icon set?

 It's true they are inconsistent. Arrows come from bootstrap, but I had to get the others from elsewere (ie gnome), as [evocated here](https://github.com/openstreetmap/openstreetmap-website/issues/3123#issuecomment-803141517).

Here is a small update after switchingfrom btn-primary to btn-light:

image

gravitystorm:
@mjourdan Thanks for your design suggestions. I'm not sure what the next steps would be, since this is only a prototype and I don't know if anyone is going to attempt to implement it. We already have the issue open - do we need to keep this PR open too?

mjourdan:
Closing this PR and keeping the issue open makes sense.

PR #3409, opened 2022-01-05
Fix display of suspension message when a user is suspended mid-session 

gravitystorm:
Without the ability defined, the user is still logged out, but then the `deny_access` check redirects to the login page. The re-login attempt would then fail anyway, with an error message, but let's fix the abilities and use the intended page.

I think this has been broken since the abilities were introduced, since there's no `suspended` action defined in the controller. I was a bit surprised that actions without definitions worked, but it seems that's a legitimate way to do things in Rails!

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -23,7 +23,7 @@ class Ability
       can [:index, :show], Redaction
       can [:new, :create, :destroy], :session
       can [:index, :show, :data, :georss, :picture, :icon], Trace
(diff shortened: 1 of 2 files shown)


simonpoole:
Does this have any bearing on https://github.com/openstreetmap/iD/issues/5400 ?

gravitystorm:
@simonpoole I don't think so - that's discussing user blocks (and how to handle those responses when using the API) whereas this is about user suspensions.

I can bet that there's a lack of tests for how mid-session user blocks are handled, although there's not a huge amount within the rails port that a user block would affect.

PR #3411, opened 2022-01-06
Use `assert_content` instead of `assert page.has_content?`

gravitystorm:
This PR switches to using `assert_content` (and `assert_link`) from Capybara's list of assertions, instead of using a plain assert on the Capybara node matchers (e.g. assert page.has_content?). 

This makes little difference when the tests are working, but the error messages are much more helpful when the tests fail (e.g. during development or refactoring). The previous approach led to "Expected false to be truthy" which is not super helpful to debug!

Example for `assert_content`:
```diff
  Failure:
  DiaryEntrySystemTest#test_reply_to_diary_entry_should_prefill_the_message_subject [/test/system/diary_entry_test.rb:15]:
- Expected false to be truthy.
+ expected to find text "Receive a new message" in "OpenStreetMap\nEdit\nHistory\nExport\nGPS Traces\nUser Diaries\nCopyright\nHelp\nAbout\nUser 5\nSend a new message to User 4\nSubject\nBody\nParsed with kramdown\nHeadings\n# Heading\n## Subheading\nUnordered list\n* First item\n* Second item\nOrdered list\n1. First item\n2. Second item\nLink\n[Text](URL)\nImage\n![Alt text](URL)\nBack to inbox"
```

Example for `assert_link`:
```diff
  IssuesTest#test_issue_index_with_multiple_roles [/test/system/issues_test.rb:116]:
- Expected false to be truthy.
+ expected to find link "5 Reports" with href "/issues/446" but there were no matches
```

https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Minitest/Assertions

### DIFF
--- a/test/system/issues_test.rb
+++ b/test/system/issues_test.rb
@@ -5,21 +5,21 @@ class IssuesTest < ApplicationSystemTestCase
 
   def test_view_issues_not_logged_in
     visit issues_path
-    assert page.has_content?(I18n.t("sessions.new.title"))
+    assert_content I18n.t("sessions.new.title")
   end
 
   def test_view_issues_normal_user
     sign_in_as(create(:user))
 
     visit issues_path
-    assert page.has_content?("Forbidden")
+    assert_content "Forbidden"
   end
 
   def test_view_no_issues
     sign_in_as(create(:moderator_user))
 
     visit issues_path
-    assert page.has_content?(I18n.t("issues.index.issues_not_found"))
+    assert_content I18n.t("issues.index.issues_not_found")
   end
 
   def test_view_issues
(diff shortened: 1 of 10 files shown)


test/system/issues_test.rb
```diff
@@ -113,7 +113,7 @@ def test_issue_index_with_multiple_roles
 
     visit issues_path
 
-    assert page.has_link?(I18n.t("issues.index.reports_count", :count => issue1.reports_count), :href => issue_path(issue1))
-    assert page.has_link?(I18n.t("issues.index.reports_count", :count => issue2.reports_count), :href => issue_path(issue2))
+    assert_link I18n.t("issues.index.reports_count", :count => issue1.reports_count), :href => issue_path(issue2)
```
tomhughes, line 116: Why does this change from `issue1` to `issue2`?

gravitystorm: Great catch! That was a mistake caused by me messing around trying to get the friendly error message to show up. I've force-pushed the corrected version.

PR #3414, opened 2022-01-10
Update to rails 7.x

tomhughes:
This is work in progress to update to rails 7 which I'm opening as all tests are now passing.

We're still waiting on updates to a few gems we use:

- [x] annotate
- [x] canonical-rails
- [x] delayed_job_active_record

Two of them are currently disabled and the third (canonical-rails) would need to be if I was using 7.0.1 as it doesn't work beyond 7.0.0 (yes really, it has a `<= 7` constraint).


gravitystorm:
Canonical rails looks like it could be updated fine - the maintainer is still active there. Annotate isn't crucial (I'd live without it for now, if it was the only blocker), and it looks like the maintainer might be awol.

`delayed_job_active_record` and the parent `delayed_job` might be more problematic, since they are looking slightly more abandoned that the others. Both have PRs for rails 7 support, and both have queries about the maintenance status. So they appear to be well used enough that some alternative solution (e.g. community fork) might appear, or since they are maintained by a company either they will find some time after the holidays to work on them again, or perhaps they can be approached about doing a release as some contract work.

The fact that the tests pass without delayed_job_active_record makes me smile nervously! I mean, it's helpful that we use the in-memory job queue for tests, but makes me pause given that it wouldn't work in production.

tomhughes:
The `delayed_job_active_record` update is out now so that's one less issue.

tomhughes:
The other two gems have been updated now so I think this is good to go.

lib/tasks/auto_annotate_models.rake
```diff
@@ -1,7 +1,7 @@
 # NOTE: only doing this in development as some production environments (Heroku)
 # NOTE: are sensitive to local FS writes, and besides -- it's just not proper
 # NOTE: to have a dev-mode tool do its thing in production.
-if Rails.env.development?
```
gravitystorm, line 4: I don't think we need this change now that annotate_models is working again

gravitystorm (approved):
Aside from the annotate models thing, everything else looks good to me. I've tested it locally and I get a few deprecation warnings but we can fix them later.

lib/tasks/auto_annotate_models.rake
tomhughes: Quite right... I've removed that again.

tomhughes:
The main deprecation warning you're probably seeing is:

```
DEPRECATION WARNING: Rails 7.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead.
```

which I fixed in https://github.com/doorkeeper-gem/doorkeeper/pull/1545 which has been merged and will be in the next doorkeeper release.

Issue #3242, opened 2021-07-01
Filter changesets in History by bounding box area


lectrician1:
A feature such as a slider that can filter the changesets by their bounding box area could help mappers sort through changesets in the area they are looking at without seeing the annoying larger changesets that cover the entire globe.

tordans:
@lectrician1 do you know osmcha? You can setup filter which only show changes inside a specified bbox.
You need to also set "bbox size bound to 1" (see https://github.com/mapbox/osmcha-frontend/search?q=bbox+size&type=issues for more).
You can setup an rss feed and such as described in https://www.openstreetmap.org/user/wille/diary/47903.

nitramwin:
@lectrician1 this would be really really great. I have just submitted an issue for iD to have a new 'check' to reduce the number of unintentional large-area edits https://github.com/openstreetmap/iD/issues/8590

Lee-Carre:
@tordans

> osmcha? You can setup filter which only show changes inside a specified bbox.

I **have** to use OSMCha with the filtering you describe, because the [osm.org/history](http://www.openstreetmap.org/history#map=10/49.2087/-2.1235) list is flooded with so many non-local changesets as to make it unusable.

The problem, then, is that the behaviour is not as you describe.

While changesets with bounding boxes larger than my area of interest are indeed excluded, also excluded are changesets which *do* make local changes.

Besides, a tool-dev (@mmd-osm) has [said](http://www.openstreetmap.org/changeset/110409629) that
> Today there's no workable solution available to analyze large changesets.

So, even with OSMCha, I either have to manually wade through dozens of completely irrelevant changesets (which make zero local changes), or have to be unaware of larger bounding-box changesets which do make some local change(s).

While OSMCha is a tool which I generally like, it's also v0, meaning that the developers think it's not yet ready for prime-time (which would be v1+).

mmd-osm:
For avoidance of doubt, I'm not involved in OSMCha development. The prototype I have presented uses alternative Overpass backends, whjile leveraging OSMCha's changeset display components to demo improvements.

boothym:
Yes, a tick box for "only show changesets in view" would be very useful - so if you are looking at one town or city you can see just the ones wholly contained in your current viewport.

mikelmaron:
> While OSMCha is a tool which I generally like, it's also v0, meaning that the developers think it's not yet ready for prime-time (which would be v1+).

[...]

PR #3417, opened 2022-01-12
Hide changesets from history list that don't fit into view

kallejre:
Not full but partial fix for issue #3242. While the original issue suggested to use a slider for filtering changesets based on size, i encountered some issues with trying to implement [version with a slider](https://github.com/openstreetmap/openstreetmap-website/issues/3242#issuecomment-1006709194) and decided to use variant with simple checkbox.

Those problems revolved mostly around having bit too large impact to be made without getting maintainers' opinion, such as adding multiple extra parameters for API, defining the extremes of the slider and how it should transition from one to another. In other words, what should be effect of setting slider to one end, to other and and how it behaves at any intermediate step. 
There were also some usability issues (some mobile devices didn't work smoothly with slider) and fears about having too severe performance impact on server. One of considered solutions (discussed outside of github) was to calculate area of every changeset and compare bbox coverage percentages, but that would have been computationally too expensive.

This PR adds following:

- Checkbox on `/history` pages allowing to hide all changesets that don't fully fit into current view.
- Parameter `fit_bbox` (boolean) for both API and history listing backends.
- Related test and i18n localization.
- Function `check_boolean` intended to be reused for parsing parameters `opened` and `closed`
- Fix for history view's bounding box creation on low zoom levels - previously use of Leaflet's `.wrap()` method on bbox caused bounding boxes' latitudes to be cropped too small (e.g if zoom to lvl 1 zoom and resize window in a way that left and right edge of map match, then only changesets from that gap would be shown).

Here's image of how new history page looks like:
osm-animated

I think i should mention that the checkbox is unchecked (`false`) by default and API also treats missing `fit_bbox` parameter as `false`. This also tries to cater for concerns about users possibly filtering out some local changesets, trying to ensure no-one accidentaly hides changesets by leaving old behaviour as default action.

PS. Weakly related question from my original comment on the issue, slightly rephrased:
> Probably I don't know how Ruby works because i have never used it, but shouldn't API `changesets` endpoint also support URL parameters `open/closed=false`? Currently API seems to only check if parameter is present, not it's value. Therefore code from that part is probably not reusable for the changeset filtering methods. Was such decision to ignore parameter values bug or a feature?


ZeLonewolf:
I would suggest inverting the logic and making the checkbox say:
[ ] Hide oversize changesets

kallejre:
Yes, that might be more understandable. I tried to get rid of negative-sounding "don't show" while at same time tried to keep the message as short and accurate as possible. I had dilemma about "oversize" vs "oversized", but shorter option was selected.

I'm still thinking what would be good effective algorithm for integrating slider into changeset filtering.

app/controllers/changesets_controller.rb
```diff
@@ -78,18 +78,35 @@ def feed
   # utility functions below.
   #------------------------------------------------------------
 
+  ##
+  # The API has few situations, where parameter is supposed
+  # to have boolean value (true/false), but previously their
+  # values were ignored and script checked only null/not null
+  # state.
+  # If string value is true, return boolean true, otherwize false.
+  def check_boolean(value)
```
tomhughes, line 87: If this is needed by multiple controllers then it should be moved up to `application_controller.rb` instead of duplicating it.

tomhughes (changes requested):
I think the main operational question here is, what is the performance impact of this likely to be - basically how many additional changesets may the database have to consider to try and find ones which fit in the view?

PR #3419, opened 2022-01-12
Use a state machine for user status

gravitystorm:
This came up during work on #3398 when considering all the ways state can be changed, and there being nowhere really to document what the states means and what it means to change them (e.g. confirmed isn't related to confirming the email, difference between an admin hiding and deleting an account, etc, which are a bit unintuitive). 

That made me think about AASM since you can define the states, the transitions, and also easily restrict which transitions between states are permitted. I think this will also come in useful when we think about removing "still pending" accounts and other things like that.

If this is approved, then I'll rework #3398 to be based on top of this.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -45,6 +45,7 @@
 
 class User < ApplicationRecord
   require "digest"
+  include AASM
 
   has_many :traces, -> { where(:visible => true) }
   has_many :diary_entries, -> { order(:created_at => :desc) }
@@ -158,6 +159,51 @@ class User < ApplicationRecord
     user
   end
 
+  aasm :column => :status, :no_direct_assignment => true do
+    state :pending, :initial => true
+    state :active
+    state :confirmed
+    state :suspended
+    state :deleted
+
+    # A normal account is active
+    event :activate do
+      transitions :from => :pending, :to => :active
+    end
+
+    # Used in test suite, not something that we would normally need to do.
+    event :deactivate do
+      transitions :from => :active, :to => :pending
+    end
+
+    # To confirm an account is used to override the spam scoring
+    event :confirm do
+      transitions :from => [:pending, :active, :suspended], :to => :confirmed
+    end
+
+    event :suspend do
+      transitions :from => [:pending, :active], :to => :suspended
+    end
+
+    # Mark the account as deleted but keep all data intact
+    event :hide do
+      transitions :from => [:pending, :active, :confirmed, :suspended], :to => :deleted
+    end
+
+    event :unhide do
+      transitions :from => [:deleted], :to => :active
+    end
+
+    # Mark the account as deleted and remove personal data
+    event :soft_destroy do
+      before do
+        remove_personal_data
+      end
+
+      transitions :from => [:pending, :active, :confirmed, :suspended], :to => :deleted
+    end
+  end
+
   def description
     RichText.new(self[:description_format], self[:description])
   end
@@ -241,8 +287,8 @@ class User < ApplicationRecord
   end
 
   ##
-  # destroy a user - leave the account but purge most personal data
-  def destroy
+  # remove personal data - leave the account but purge most personal data
+  def remove_personal_data
     avatar.purge_later
 
     self.display_name = "user_#{id}"
@@ -253,7 +299,6 @@ class User < ApplicationRecord
     self.new_email = nil
     self.auth_provider = nil
     self.auth_uid = nil
-    self.status = "deleted"
 
     save
   end
@@ -279,7 +324,7 @@ class User < ApplicationRecord
   ##
   # perform a spam check on a user
   def spam_check
-    update(:status => "suspended") if status == "active" && spam_score > Settings.spam_threshold
+    suspend! if may_suspend? && spam_score > Settings.spam_threshold
   end
 
   ##
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -181,8 +181,19 @@ class User < ApplicationRecord
       transitions :from => [:pending, :active, :suspended], :to => :confirmed
     end
 
+    # To unconfirm an account is to make it subject to future spam scoring again
+    event :unconfirm do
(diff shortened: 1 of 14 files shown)


tomhughes (changes requested):
I don't have any problem with this in principle but it looks like it has removed the ability to deactivate a user by moving them back to pending status - in itself that probably isn't very important as I'm not sure I ever use it directly.

I do use it as a first step to move a user from suspended to active though, because that isn't currently possible to do directly, and still isn't with this, but now there is no workaround either.

app/models/user.rb
```diff
@@ -158,6 +159,51 @@ def self.authenticate(options)
...
+    end
+
+    # Mark the account as deleted but keep all data intact
+    event :hide do
+      transitions :from => [:pending, :active, :confirmed, :suspended], :to => :deleted
+    end
+
+    event :unhide do
+      transitions :from => [:deleted], :to => :active
+    end
```
tordans, line 195: This signals to me that hiding is not really a state but a flag on the user. A bit like act_as_paranoid ([and similar](https://github.com/rubysherpas/paranoia)) did back in the day. 

Using a flag would clarify the `:unhide` event. Right now, a `pending` user that gets deleted would become `active` without ever running through the activation steps.

Another way to solve this would be to transition to `pending` within `unhide` and ask the user to re-take the activation steps.

---

However, I wonder what the use case for hiding an account would be. I had assumed that `remove_personal_data` was applied to all deleted users. Which IMO is desirable since we don't want personal info lying around on accounts that are not used and not visible, right? Could we maybe remove the `hide` usecase altogether?

app/models/user.rb
```diff
@@ -158,6 +159,51 @@ def self.authenticate(options)
     user
   end
 
+  aasm :column => :status, :no_direct_assignment => true do
```
tordans, line 162: Since the user model file is already very long, [this aasm section](https://github.com/openstreetmap/openstreetmap-website/blob/1a11c4dc191d93b18fcf5aa917448c8cd6d2556b/app/models/user.rb#L162-L205) would be a good candidate to extract into a concern IMO. However, skimming over https://codeclimate.com/blog/7-ways-to-decompose-fat-activerecord-models/ again suggests that a Concern is also smelly, so maybe there is a better way to separate the large file(?).

tordans (commented):
This makes the code a lot better 👍 .

app/models/user.rb
tomhughes: It's used for spam accounts, to close them without removing the evidence.

app/models/user.rb
```diff
@@ -158,6 +159,51 @@ def self.authenticate(options)
...
+
+    # A normal account is active
+    event :activate do
+      transitions :from => :pending, :to => :active
+    end
+
+    # Used in test suite, not something that we would normally need to do.
+    event :deactivate do
+      transitions :from => :active, :to => :pending
+    end
```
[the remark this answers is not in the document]
tomhughes: Well if it was only used for test cases then we should just remove those test cases as by definition we would be testing functionality that wasn't used in real life - that's better than polluting the source with comments.

As I said in my review though currently there is a production use case, it's just that this PR currently removes the ability to use it. There's probably no need for it though, if I was able to move people from suspended/confirmed to active though.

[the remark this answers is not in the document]
gravitystorm: It's only used in the factory, not in test cases. In the tests we want to default to having users in the "active" state for the vast, vast majority of test cases, but then there's no way to get a user back into the "pending" state for those tests that need it, because direct attribute assignment is disabled.

I think rather than a guard I might change it to being wrapped in the check, so that if you e.g. make a diagram using aasm-diagram it doesn't show up.

I'm open to better ideas - I thought that there might be a way in FactoryBot to set the attribute without aasm complaining but I couldn't find out how. See also https://github.com/aasm/aasm/issues/523

gravitystorm:
> I do use it as a first step to move a user from suspended to active though, because that isn't currently possible to do directly, and still isn't with this, but now there is no workaround either.

If we need another transition then I'm happy to add one. The `confirm` transition already allows going from `:suspended` to `:confirmed`. So do you also want an `unsuspend (:from => [:suspended], :to => :active)` ?

tomhughes:
> > I do use it as a first step to move a user from suspended to active though, because that isn't currently possible to do directly, and still isn't with this, but now there is no workaround either.
> 
> If we need another transition then I'm happy to add one. The `confirm` transition already allows going from `:suspended` to `:confirmed`. So do you also want an `unsuspend (:from => [:suspended], :to => :active)` ?

I think we definitely need that, yes. The only question is whether we also need to be able to go from confirmed to active - it's certainly wouldn't be a common thing to do most likely only something you'd use to reverse an accidental confirmation.

gravitystorm:
> This signals to me that hiding is not really a state but a flag on the user. A bit like act_as_paranoid (and similar) did back in the day.

Yep, there's big questions about whether hiding and suspending and activating (confirming email) should be separated out here into different columns. For example, being suspended for having a spammy profile doesn't reflect on whether you have confirmed your email or not, but combining the two concepts (as we have now) leads to information loss.

But that might be digging myself a bit of a hole here. On one side, I'm just trying to implement self-deletion of accounts. But on the other, we might not need AASM at all if we moved some of these states into flags. Thoughts and advice welcome.

mmd-osm:
I didn't look into any of the details, making this is a mostly uninformed heads up: In case any of those changes have some impact on the user's database table, more specifically the status column, you should probably let me know... ([here's why](https://github.com/zerebubuth/openstreetmap-cgimap/blob/cf3fb37a3441ffeb952278a7b5b209cc2b7c109e/src/backend/apidb/readonly_pgsql_selection.cpp#L748-L756))

jmgarnier:
Nice to bump into known people 😸 👋🏻 Hey @tordans !

gravitystorm:
I've added the two extra transitions, and made it so that deactivate is only available in the test environment.

tomhughes (approved):
Looks good now!

PR #3420, opened 2022-01-13
Add dir="auto" to search to improve right-to-left language support

mapmeld:
Improve right-to-left language search experience by adding HTML attribute `dir="auto"` to the input. Browsers will align left or right based on the input text.

No effect for left-to-right language input.

I am open to reordering the tag, or limiting this to just the primary search input.

### DIFF
--- a/app/views/layouts/_search.html.erb
+++ b/app/views/layouts/_search.html.erb
@@ -5,7 +5,7 @@
         <div class="input-group flex-nowrap">
           <div class='query_wrapper position-relative flex-grow-1'>
             <%= link_to t("site.search.where_am_i"), "#", :class => "describe_location position-absolute", :title => t("site.search.where_am_i_title") %>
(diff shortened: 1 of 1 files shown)


tomhughes:
I guess the question is, do people expect an input field like that to follow the rules of the locale for the rest of the site, or for the language they are typing...

Basically do people using the site in an RTL language expect an english name to be left or right aligned in the input field?

Obviously the question is also valid the other way round but I suspect most LTR users don't have enough experience of entering RTL values to have an opinion.

mapmeld:
I'd cite the W3C that this is the best practice 
- on forms https://www.w3.org/International/questions/qa-html-dir#quickanswer
- on input element https://www.w3.org/International/questions/qa-html-dir#autoinput

I tested the current behavior when Arabic is your language on OSM.org ; it forces any text input to be right-to-left, which is also incorrect and fixed by this PR.

Screen Shot 2022-01-18 at 11 35 04 AM

PR #3429, opened 2022-01-19
Add dir=auto to user generated content

### LINKED ISSUE #3428: Set dir="auto" on all user-generated-content
### Description

I think we have a reasonably complex scenario, since we have lots of "mixed-direction" situations - where someone has the site set to a language in one direction (e.g. RTL) but is viewing content in another direction (e.g. LTR). This makes it different to many websites which are usually either one or the other, and so extra care is needed.

"Mixed-direction"? For example, an Arabic speaker looking at map notes, or a French speaker viewing a Hebrew diary entry, or many other things like this. It's particularly hard for RTL users since much of the website user-generated-content (e.g. tag values) are in LTR text, and we currently set the whole page to be treated as RTL which then messes things up:

Screenshot from 2022-01-19 16-26-27

Note how the placement of the punctuation is scrambled. The logic used by browsers is to work through the text, and if it's recognisably LTR characters it will keep them in the right order, but placing each chunk right-to-left. So it places following characters (e.g. punctuation) "after" the text, which makes it look garbled.
bitmap

The first way that we can fix this is by putting `dir=auto` around any user-generated-content. We don't need to know or store language used for the UGC, since browsers will auto-detect based on the first indicative character found. Note that this means it happens element-by-element, so e.g. a diary entry with a paragraph in Arabic and another in English (or a blockquote, etc) will then work fine since the browser will auto-detect each individually. Setting `dir=auto` in this case gets us:

Screenshot from 2022-01-19 16-31-53

That's already better. The text-alignment is still incorrect, but that'll be dealt with in a different issue.

### Screenshots

_No response_

tomhughes:
Fixes #3428


tomhughes:
I was wondering about usernames, but they are displayed in a lot of places, and often in contexts where I'd be quite worried about the layout effects of them leaping to the far side of the page.

tomhughes:
Usernames are also often interpolated in longer strings - doe we just wrap them in a span with dir=auto within the wider string? Mark the whole wider string as auto?

gravitystorm:
> I was wondering about usernames, but they are displayed in a lot of places, and often in contexts where I'd be quite worried about the layout effects of them leaping to the far side of the page.

Hopefully they won't jump around much, due to the [text-align issue](https://github.com/openstreetmap/openstreetmap-website/issues/3432) if nothing else! But I think there aren't many cases were usernames are on their own with much freedom to roam around.

> Usernames are also often interpolated in longer strings - doe we just wrap them in a span with dir=auto within the wider string? Mark the whole wider string as auto?

I think just the username. If we mark the wider string as dir=auto and the username comes first, then it'll rearrange the remainder of the text too if the wider string contains punctuation. 

Contrived example:
`username commented yesterday! - block` could become `block! - username commented yesterday` if that username was in arabic, since the dir=auto would see the first character as arabic, and layout the wider string in three chunks RTL (`username commented yesterday ` first, then `! - `, then `block`). If just the username is dir=auto, then it won't scramble the rest of wider string.

tomhughes:
I've added a commit that makes a start on the usernames but as I thought it's quite involved - a number of languages resource have to be converted to html support to start with.

I think I've done most of the views now, but there's also some in controllers that I need to check out - those are hard to fix because helpers are accessible from controllers :-(

NeatNit:
I know this hasn't seen any action in 3 years, but if this is ever picked up again you should know:

My PR #5835 applies `dir="auto"` to each paragraph individually in rich text content. This produces better results, in line with every other platform used today (e.g. this very comment, or your favorite instant messaging app). It also makes some - though definitely not all - of the changes in this PR redundant.

Non-rich text like usernames still needs this PR, but rich text that converts to HTML with `to_html` is better handled "upstream" by that conversion.

NeatNit:
Thank you @tomhughes for merging #5840 today! Let me know if there's any way I could help with this one.

In particular, there are small things I've missed in my PRs because I didn't have the capacity to test properly. Namely:

- Title of diary entries, both in [lists of diary entries](https://master.apis.dev.openstreetmap.org/diary/he) and in [the diary entry itself](https://master.apis.dev.openstreetmap.org/user/NeatNit_Dev/diary/39) (the `!` should appear at the left end of the title)
- Changeset description (`comment` tag) in history view, both [general](https://www.openstreetmap.org/history) and [user-specific](https://www.openstreetmap.org/user/NeatNit/history?after=163583490) - see screenshots:

Wrong:
image
Right:
image

I think this PR might already deal with at least one of those. Bug again, no capacity for me to test.

Also, semi-relatedly, I noticed that the title of a diary entry has "&amp;lrm;" in it. For example, with the website in Hebrew the title of [this page](https://www.openstreetmap.org/user/pehar_11/diary/406634) is:

```html
<title>היומן שלpehar_11 &amp;rlm; | [u tijeku] Uređenje centra Rijeke | OpenStreetMap</title>
```

Clearly something needs fixing there. Haven't looked into it yet.

app/controllers/traces_controller.rb
```diff
@@ -33,7 +34,7 @@ def index
              elsif current_user && current_user == target_user
                t ".my_gps_traces"
              else
-               t ".public_traces_from", :user => target_user.display_name
+               t ".public_traces_from_html", :user => tag.bdi(target_user.display_name)
```
AntonKhorev, line 37: Doesn't work as expected for some reason. That's what I see in the upper left corner on `/user/:name/traces` pages:

image

tomhughes: So I've failed to find out exactly what's going on here but something deep in the rails output capture magic is going wrong and causing most of the output inside the `tag.head` block in `app/views/layouts/_head.html.erb` to be output directly instead of being captured and then wrapped in the tag.

It turns out that `bdi` tags in a title element don't work anyway though as the browser will just render them as text in the title bar, so I've replaced them with the equivalent unicode characters which is a trick we already use in `set_title`.

app/controllers/traces_controller.rb
```diff
@@ -1,6 +1,7 @@
 class TracesController < ApplicationController
   include UserMethods
   include PaginationMethods
+  include ActionView::Helpers::TagHelper
```
AntonKhorev, line 4: Is this include still needed after the fix https://github.com/openstreetmap/openstreetmap-website/pull/3429#discussion_r2106266400?

Issue #3427, opened 2022-01-19
Improve our handling of RTL


gravitystorm:
### Description

After considering #3420 further, I think there are several steps that we can take to improve the RTL experience. In doing this research I can heartily recommend the [rtlstyling](https://rtlstyling.com/posts/rtl-styling) site, which goes into a lot of nuance in this area that I haven't found well explained elsewhere.

I'm going to split this into multiple issues, so this will be an overall issue to show how everything will join up together. Current list of topics is:

 * Use dir="auto" for all user-generated-content
 * Use dir="auto" for most form inputs
 * Avoid using text-align on body element
 * (Still under consideration) Refactor CSS to remove the need for using r2
 * Update to bootstrap 5 (for non-text reasons, mainly button corners and menus)

I hope to make issues for each of the above in the next few days

PR #3430, opened 2022-01-19
Add dir=auto to text inputs

tomhughes:


HolgerJeromin:
ref #3427

gravitystorm:
Instead of putting this on every `text_field` and `richtext_field` individually, can we do that somewhere centrally? I haven't researched how easy rails makes it to add your own attributes to the field helpers.

Bootstrap_form would still (hopefully!) work fine since it uses those rails helpers underneath.

Might need to find (and convert?) anywhere still using hand-built html forms that aren't using the rails helpers. Adding changeset comments for example? And notes?

I don't think there's any forms with text input that we build in javascript (although we build certain other forms that way).

tomhughes:
Well as that site you pointed at says there are some fields (like email addresses) which shouldn't have dir=auto so if we did find a way to make it automatic we would need to be able to override it when we don't want it.

I mean I guess we might be using `email_field` for email but there are probably other text fields (latitude and longitude for user positions for example_ which shouldn't have it?

gravitystorm:
> we would need to be able to override it when we don't want it.

Yep. I expect if there was a default set via the rails helpers it would still be overridden by the options hash when the helper is used, so we can add the `dir` to the one or two cases that need it (basically the inverse of what we have here).

> (latitude and longitude for user positions for example)

text_fields used for numbers should be fine, since with dir="auto" they are actually better laid out for negative numbers than currently. In dir="rtl" the negative sign is shown to the right (before) the number, but in arabic the [negative should be shown to the left](https://en.wikipedia.org/wiki/Modern_Arabic_mathematical_notation). Both dir="auto" and dir="ltr" do the right thing here.

(If anyone has a contra-example for an RTL language that shows regular numerals differently, then I'd love to hear about it - I can't speak or read any RTL languages myself so my knowledge is limited to what I read online).

gravitystorm:
> I haven't researched how easy rails makes it to add your own attributes to the field helpers.

I've been looking into this today, and I haven't found any particularly neat way of doing this, other than monkey-patching the TextField initializer to add the option, e.g.

```ruby
# Apply dir=auto to all text fields
# Also includes fields subclassed from TextField, like PasswordField, EmailField etc
module ActionView
  module Helpers
    module Tags
      class TextField
        def initialize(object_name, method_name, template_object, options = {})
          super(object_name, method_name, template_object, options.reverse_merge({ :dir => :auto }))
        end
      end
    end
  end
end
```
Reading the source code for [TextField](https://github.com/rails/rails/blob/7-0-stable/actionview/lib/action_view/helpers/tags/text_field.rb) and [Base](https://github.com/rails/rails/blob/7-0-stable/actionview/lib/action_view/helpers/tags/base.rb) doesn't give me any hints that there are configurable defaults for the options hashes, only that they are expected to be passed in by the caller.

I guess a non-monkey-patching approach would be to use our own FormBuilder that subclasses bootstrap_form's [FormBuilder](https://github.com/bootstrap-ruby/bootstrap_form/blob/main/lib/bootstrap_form/form_builder.rb), and add the default html attributes at that stage. I haven't explored that approach yet to see if there's any gotchas there.

Issue #1934, opened 2018-07-22
Add user interface for moderators to hide note comments


woodpeck:
The data model already supports hiding individual note comments, however this currently requires setting the `visible` flag on the comment to false using SQL in the database since there is no API or user interface for it. It would be great if moderators could hide individual comments through the user interface. (Currently they can only hide a note completely.)

mmd-osm:
There's in fact an API endpoint available which is only available to moderators. For some reason it has never been documented on the Wiki page. I've added that missing documentation now: [link](https://wiki.openstreetmap.org/wiki/API_v0.6#Hide_changeset_comment:_POST_.2Fapi.2F0.6.2Fchangeset.2Fcomment.2F.23comment_id.2Fhide)

Unfortunately, it's a bit difficult to find out the changeset comment id as a user, as it's only exposed as `<li id="c33"/>` on the changeset browse page for changeset comment id 33. But after all, it's only an API...

Feel free to correct/enhance as needed.

mmd-osm:
(please ignore previous post, this issue is about note comments, not changeset comments...)

mmd-osm:
Note to EWG: Probably nothing left to do here except merging the pull request.

PR #3438, opened 2022-01-28
Added (un)hide note comments API endpoint

mmd-osm:
Adding the API part to address #1934 to hide/unhide single note comments for moderators. The code is very closely following the existing changeset comment hide / destroy implementation.

NotesController `comment` method could be moved to the new NoteCommentsController, to more closely follow the overall design in ChangesetCommentsController. I left that for another request to keep this one as small as possible.

UI part is covered in #3441, providing  the familiar "hide" / "unhide" feature:

modnotecomment

- [ ] Resolve discussion in https://github.com/openstreetmap/openstreetmap-website/issues/3450#issuecomment-1034913872 on how to handle "opening" events

Previous discussion:

> An "opened" event is used exclusively when creating a new note. https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/api/notes_controller.rb#L76
> 
> I think what needs to be clarified first is whether we want the opening event to be hidden at all. Depending on this decision other parts of the code need to be adjusted to handle this case properly.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,8 +32,16 @@ Metrics/ClassLength:
   Exclude:
     - 'test/**/*'
 
+Metrics/CyclomaticComplexity:
+  Exclude:
+    - 'app/abilities/api_capability.rb'
+
 Metrics/ModuleLength:
   Max: 150
+
+Metrics/PerceivedComplexity:
+  Exclude:
+    - 'app/abilities/api_capability.rb'
 
 Naming/FileName:
   Exclude:
--- a/app/abilities/api_ability.rb
+++ b/app/abilities/api_ability.rb
@@ -46,6 +46,7 @@ class ApiAbility
 
         if user.moderator?
           can [:destroy, :restore], ChangesetComment
+          can [:destroy, :restore], NoteComment
           can :destroy, Note
 
           if user.terms_agreed?
--- a/app/abilities/api_capability.rb
+++ b/app/abilities/api_capability.rb
@@ -29,6 +29,7 @@ class ApiCapability
 
       if user&.moderator?
         can [:destroy, :restore], ChangesetComment if scope?(token, :write_api)
+        can [:destroy, :restore], NoteComment if scope?(token, :write_api)
         can :destroy, Note if scope?(token, :write_notes)
         if user&.terms_agreed?
           can :redact, OldNode if scope?(token, :write_api)
--- a/app/controllers/api/note_comments_controller.rb
+++ b/app/controllers/api/note_comments_controller.rb
@@ -1,0 +1,52 @@
+module Api
+  class NoteCommentsController < ApiController
+    before_action :authorize
+
+    authorize_resource
+
+    before_action :check_api_writable
+    before_action :check_api_readable
+    around_action :api_call_handle_error
+    around_action :api_call_timeout
+
+    ##
+    # Sets visible flag on note comment to false
+    def destroy
+      # Check the arguments are sane
+      raise OSM::APIBadUserInput, "No id was given" unless params[:id]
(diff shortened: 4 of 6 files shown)


gravitystorm:
One of the problems with special-casing the "opened" event is that it's not always the first visible comment. For example, if a user is deleted, the new first visible comment is then the original first reply, and so on as more comments are hidden and/or users deleted.

mmd-osm:
Looking at https://github.com/openstreetmap/openstreetmap-website/blob/master/app/views/browse/note.html.erb#L38 I think, "opened" is never a visible comment. It seems to be some technical artifact to store additional data (like author_ip) for notes created by anonymous, as well as known users.

mmd-osm:
I think, "opened" notes never qualify to be displayed, as they are used to store additional data for the note (like ip address or user id). The question what happens to other comments (even if they come from deleted users) is somehow unrelated, imho.

AntonKhorev:
> One of the problems with special-casing the "opened" event is that it's not always the first visible comment.

Opening comment is either the first (description) one or is not shown. It can't be in any other place. What's possible is that the visible "description" is not the opening comment, but that's unrelated to the API changes here. And it won't happen once #3608 is merged.

The UI PR #3441 doesn't have a button to hide the description. If we want to be able to do this we can add it later.

mmd-osm:
@AntonKhorev : I have no more plans to follow up on this pull request (as well as #3441). If you like you can cherry pick my changes and create a new pull request. As you've done a number of related changes, it makes most sense to have it all done by one dev. Closing here.

gravitystorm:
I'd really rather these were left open, because I do mean to review them and if they are closed they will drift away into the void.

But you've closed them twice now, and so I'll respect that and leave them as you wish.

mmd-osm:
The original issues are still around, and the topic is on the DWG wishlist. So it's unlikely that topic will disappear.

As I'm not following the repo much these days, and there's lots of related activity in other pull requests I can't keep up with at the moment, closing seemed unavoidable. I really hope @AntonKhorev can take over somehow.

AntonKhorev:
Is there any related activity in pull requests other than https://github.com/openstreetmap/openstreetmap-website/pull/3608?

mmd-osm:
The biggest challenge for me was that the discussion in #3450 and https://github.com/openstreetmap/openstreetmap-website/pull/3438#issuecomment-1044512922 was never resolved.

Then we have some related activity in #3608 as you've noted. There may be in one of your pull requests. I haven't looked through all of them, so may be missing some context here.

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/pull/3438#issuecomment-1044512922 - see https://github.com/openstreetmap/openstreetmap-website/pull/3438#issuecomment-1193133163

https://github.com/openstreetmap/openstreetmap-website/pull/3608 is a fix for both https://github.com/openstreetmap/openstreetmap-website/issues/3450 and https://github.com/openstreetmap/openstreetmap-website/pull/3438#issuecomment-1044512922

There's no "lots" of related activity.

AntonKhorev:
All I can do here is wait for https://github.com/openstreetmap/openstreetmap-website/pull/3608 to get merged and if that happens, do https://github.com/openstreetmap/openstreetmap-website/pull/3608#issuecomment-1191269573 in a new pull request.

kmpoppe:
> All I can do here is wait for https://github.com/openstreetmap/openstreetmap-website/pull/3608 to get merged and if that happens, do https://github.com/openstreetmap/openstreetmap-website/pull/3608#issuecomment-1191269573 in a new pull request.

@Anton, #3608 was closed due to being superseded, are you interested in continuing from here?

Issue #1711, opened 2017-12-28
Limits on number of tags and relation members


mmd-osm:
While reviewing the diff upload code, I noticed that there's no max limit on the number of key/value pairs per object, as well as on the number of relation members in a relation. It would probably be a good idea to have such limits, given that we already have a max number of objects per changeset, and max number of nodes per way.

woodpeck:
Data point: actual maximum number of tags per object is ~ 430, on http://www.openstreetmap.org/node/424298167

tomhughes:
If you're going to propose adding an arbitrary limit you really need to give a good reason, not just say it would be a "good idea".

mmd-osm:
Today it's difficult to reason about the diff size and the corresponding memory requirements.

Let's assume we create a changeset with 30 nodes and 10'000 tags each. Each tag can have up to 255 characters in both key and value. That's already adding up to 146MB worth of data, giving both the upload and downstream apps like JOSM a hard time. Memory requirements will be far less with 5 characters for keys+value (<3MB), for a mapper it's still not helpful to have objects with that many tags.

I see this more as a safeguard to prevent silly things from happening. It shouldn't get in the way of any normal mapping activity, though. I don't have a strong opinion on the actual limit. 500 will for sure be too low, 10'000 seems unreasonable. Maybe around 1000-1500? I'm leaving that open for further discussion.

nurfarma:
[**Obat Pembesar Penis**](https://www.kampungobat.com/)
Sedia berbagai produk obat pembesar alat vital permanen di jual dengan harga paling murah.
Berikut obat pembesar penis yang kami jual:
- [Vimax Canada Asli](https://www.kampungobat.com/vimax-canada-asli/)
- [Klg Herbal Pills](https://www.kampungobat.com/klg-herbal-pills/)
- [Hammer Of Thor](https://www.kampungobat.com/hammer-of-thor/)
- [Hendel Forex](https://www.kampungobat.com/obat-hendel-forex/)
- [Vigrx Plus](https://www.kampungobat.com/vigrx-plus-usa-asli/)
- [Titan Gel](https://www.kampungobat.com/titan-gel/)
- [TestoUltra Testosterone](https://www.kampungobat.com/testoultra-testosterone-enhancer/)
- [Anabolic Rx24](https://www.kampungobat.com/jual-anabolic-rx24/)
- [Casturex](https://www.kampungobat.com/obat-caturex-herbal/)
[...]

PR #3440, opened 2022-01-30
Introduce relation member limit

mmd-osm:
See #1711 

* Check relation member limit when saving relations
* settings.yml: new parameter `max_number_of_relation_members: 32000`
* `/api/0.6/capabilities` now includes a new element `relationmembers`
* New error message "You tried to add #{provided} members to relation #{id}, however only #{max} are allowed" is already being used by CGImap


gravitystorm (approved):
I'm happy with this, but given the impact I'd like @tomhughes to approve it too before we merge.

test/models/relation_test.rb
```diff
@@ -230,4 +230,24 @@ def test_changeset_bbox_delete_relation
     assert_equal 39, changeset.min_lat
     assert_equal 116, changeset.max_lat
   end
+
+  # Check that the preconditions fail when you are over the defined limit of
+  # the maximum number of members in a relation.
+  def test_max_members_per_relation_limit
+    # Speed up unit test by using a small relation member limit
+    Settings.max_number_of_relation_members = 20
```
tomhughes, line 238: This should probably be reset to the original value after the test - it's probably not an issue here but changing global state in a test needs to be done carefully in case it affects later tests.

tomhughes (changes requested):

test/models/relation_test.rb
mmd-osm: Good point, should be fixed now.

tomhughes (approved):

mmd-osm:
@matkoniecz : https://github.com/zerebubuth/openstreetmap-cgimap/pull/174#issuecomment-466824718 has those details.

matkoniecz:
So it seems to be safe to ignore it if editor does not allow opening external data or automatic inflating data? Or mass adding elements to a relation?

mmd-osm:
@matkoniecz : can you do a follow up of this topic on the StreetComplete repo?

matkoniecz:
@mmd-osm If you think that there is chance that SC should be in some way modified to handle this change then yes.

mmd-osm:
Sorry, I simply don't know in detail how SC works with relations. I know that a Vespucci user managed to create even larger relations, and obviously JOSM. Even for iD I cannot tell if this can realistically happen. After all, the issue is rather niche for map editors. Downstream data consumers are much more in focus for this change.

gravitystorm:
Thank you @mmd-osm for the code here, and the corresponding code in cgimap, and for the various issues and announcements that you made too!

mmd-osm:
@tomhughes : just a reminder that CGImap requires additional settings in the Chef repo, see https://github.com/openstreetmap/openstreetmap-website/pull/3440#issue-1118447495

PR #3441, opened 2022-01-30
Added (un)hide note comments UI

### LINKED ISSUE #975: Notes - it'd be nice to be able to hide an individual note comment, not the whole note
For example, something like:

http://api06.dev.openstreetmap.org/note/912

Currently the only option is to hide the entire note, and re-add the sensible part as a new note.  This could become a bit of a pain if there are more than a couple of comments.

(background - Currently we're playing whack-a-mole with someone who's using anonymous note comments to post a phone number unrelated to the notes online.  Issue 921 also applies to these of course).

mmd-osm:
UI part for #1934 , depends on API backend in #3438

This pull request provides the same "hide" / "unhide" feature  already known from changeset comments:

modnotecomment

Code is very similar to https://github.com/openstreetmap/openstreetmap-website/blob/master/app/views/browse/changeset.html.erb#L33-L67

Closes https://github.com/openstreetmap/openstreetmap-website/issues/1934
Closes #975

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,8 +32,16 @@ Metrics/ClassLength:
   Exclude:
     - 'test/**/*'
 
+Metrics/CyclomaticComplexity:
+  Exclude:
+    - 'app/abilities/api_capability.rb'
+
 Metrics/ModuleLength:
   Max: 150
+
+Metrics/PerceivedComplexity:
+  Exclude:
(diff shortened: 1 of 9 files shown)


mmd-osm:
Closing here as per comment in https://github.com/openstreetmap/openstreetmap-website/pull/3438#issuecomment-1304502068

PR #3445, opened 2022-02-02
Added some key-value pairs under `railway=`

mortenbruhn:
These labels turn up in the 'Find objects' tool. 
They should be made available for translation - hence this PR.

See discussion in the following PRs:

- Original: #3124
- Supplemental: #3391

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1143,6 +1143,7 @@ en:
           "yes": "Place"
         railway:
           abandoned: "Abandoned Railway"
+          buffer_stop: "Buffer Stop"
           construction: "Railway under Construction"
           disused: "Disused Railway"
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -1156,6 +1157,7 @@ en:
           platform: "Railway Platform"
           preserved: "Preserved Railway"
           proposed: "Proposed Railway"
+          rail: "Rail"
```
mortenbruhn, line 1160: would "Railway track" be more accurate?

The page for `railway=rail` does not mention "track" much.

[https://wiki.openstreetmap.org/wiki/Tag:railway%3Drail](https://wiki.openstreetmap.org/wiki/Tag:railway%3Drail)

PR #3453, opened 2022-02-12
Adds advancedcomp to the list of macOS Homebrew dependencies

spilth:
This PR adds `advancecomp` to the list of macOS Homebrew dependencies which provides the required `advpng`.

## Background

When running the test suite and `assets:precompile` I noticed the following warning:

```bash
$ rake assets:precompile
advpng worker: `advpng` not found; please provide proper binary or disable this worker (--no-advpng argument or `:advpng => false` through options)
```

This can be alleviated by installing `advancecomp` which provides `advpng`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -90,7 +90,7 @@ Installing other dependencies:
 
 * Install Homebrew from https://brew.sh/
 * Install the latest version of Ruby: `brew install ruby`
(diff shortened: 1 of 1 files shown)


tomhughes:
Well advpng isn't really required - it's an optional extra that (if available) may be used for compressing images in production. It's not used in development at all.

spilth:
@tomhughes I'm a fan of trying to get rid of all warning messages if possible, lest people start ignoring them and bigger issues start being skipped because of it. And this seemed like an easy enough fix. Thanks for the merge!

PR #3455, opened 2022-02-12
Fix/remove some bits of contributing docs

harry-wood:
Fix/remove some bits of CONTRIBUTING.md docs
* Remove the reference to Travis CI and re-describe it
* Remove the section about `rake doc:app` which no longer works


tomhughes:
There's HTML documentation?

PR #3459, opened 2022-02-15
Add yalphabetize for yaml normalisation

samrjenkins:
This PR introduces [yalphabetize](https://github.com/samrjenkins/yalphabetize) to the project. I would love to hear your thoughts/feedback.

It provides us with the ability to quickly normalise/alphabetise YAML files such as locales and to maintain this normalisation into the future.

Advantages of alphabetising YAML include:
- easier to locate specific keys in large files
- easier to spot duplicate keys
- easier to compare equivalent locale YAMLs for different languages

Yalphabetize can also be integrated into CI to ensure that future PRs maintain the alphabetisation of locale files.

I am slightly unsure how to handle alphabetisation of the other languages because the CONTRIBUTION guidelines state that these files are maintained via a 3rd party. Is it ok to sort these other locales in the same way as en.yml?

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -113,3 +113,27 @@ jobs:
         bundle install --jobs 4 --retry 3
     - name: Run brakeman
       run: bundle exec brakeman -q
+  yalphabetize:
+    name: Yalphabetize
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2.4.0
+    - name: Setup ruby
+      uses: actions/setup-ruby@v1.1.3
+      with:
+        ruby-version: ${{ env.ruby }}
+    - name: Cache gems
+      uses: actions/cache@v2.1.7
+      with:
+        path: vendor/bundle
+        key: bundle-${{ env.os }}-${{ env.ruby }}-${{ hashFiles('Gemfile.lock') }}
+        restore-keys: |
+          bundle-${{ env.os }}-${{ env.ruby }}-
+    - name: Install gems
+      run: |
+        gem install bundler
+        bundle config set deployment true
+        bundle install --jobs 4 --retry 3
+    - name: Run yalphabetize
+      run: bundle exec yalphabetize
--- a/.yalphabetize.yml
+++ b/.yalphabetize.yml
@@ -1,0 +1,2 @@
+only:
+  - config/locales/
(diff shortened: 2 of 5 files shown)


tomhughes:
I'm not sure we actually want to fully alphabetize the locales - one example where we might not is that we currently have all the standard rails stuff at the top before our custom stuff.

If we do go for it then we would certainly need to encourage users to alphabetize before submitting or I can see this going off constantly in CI.

tomhughes:
To answer your second question, no you can't change any other locale as they are maintained via TranslateWiki.

gravitystorm:
Thanks @samrjenkins for the PR, I like the idea of having consistency in ordering in our en.yml file. It's something that I try to do when I'm adding new keys, but without any linting it's hard to maintain (and unreasonable to ask other people to follow undocumented manual ordering). So I'm happy to add something like yalphabetize.

But I think the "consistency in ordering" is more important than "strictly alphabetical". For example, you ask:

> Do you think there is much advantage in having the rails translations at the top?

Yes, I do. There are around 5 top-level keys that are used globally and are not related to controllers. So we put them at the top of the file. I think that makes the layout more logical than having e.g. the globally-useful "helpers" translations in among the controllers.

But this principle also extends to other keys. I would prefer the controller method names (the second-level keys, for most top level keys) to follow a specific order (e.g. `index show new create edit update destroy`) so that they (hopefully) match what is found in the controller files. And I would like for the view-related keys (usually the third/final level key) that they are alphabetical, except I always like to put `title` first.

I saw in the yalphabetize documentation that we can configure the order, but only for complete sets.
> numbers_extended:
> \# The keys in this mapping do not match an allow order or a subset of an allowed order (notice the extra `zero` key which does not appear in the allowed order of numbers).
> \# They are ordered alphabetically.
> one: 1
> three: 3
> two: 2
> zero: 0

If we could have something similar but where members of a configured set come first, in that order, and then everything else comes after, in alphabetical order, I think that might be perfect, e.g.:

> ordered_first
> \- [html activerecord helpers time datetime]
>  \- [index show new create edit update destroy]
> \- [title]

> users:
>   index: ...
>   update: ...
>   anything: ...
>   else: ...
>   in: ...
>   order: ...

gravitystorm:
@samrjenkins Any news on yalphabetize? If we can get the behaviour described in https://github.com/openstreetmap/openstreetmap-website/pull/3459#issuecomment-1040373207 then I'd be happy to start using it.

gravitystorm:
Closing for now, but I'm happy to reopen when there's some new features from upstream.

PR #3467, opened 2022-02-22
Make sidebar resizable

Strubbl:
This commit makes the sidebar horizontally resizable. It is not resizable on small screens.

Is my assumption right, that the `small.scss` file is the CSS for small screens? Because i want the sidebar to not be resizable on small screens, because it is above the map then. Would it make sense to make the sidebar vertically resizable on small screens?

In the `common.scss` i use `overflow: auto;`, but i saw [some lines above](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/stylesheets/common.scss#L349) that there are other overflow rules. Does this interfere with my lines? I do not know how to optimize this.

The result of this commit shall be the same as the result of my [Greasemonkey script](https://gitlab.com/Strubbl/userscripts/-/blob/master/osm_sidebar_resizer.user.js), which i wrote first. In the German forum i [was recommended](https://forum.openstreetmap.org/viewtopic.php?pid=856287#p856287) to propose this change to this repository.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -354,6 +354,8 @@ body.compact-nav {
     float: left;
     width: $sidebarWidth;
     background: #fff;
+    resize: horizontal;
+    overflow: auto;
 
     #sidebar_loader {
       display: none;
--- a/app/assets/stylesheets/small.scss
+++ b/app/assets/stylesheets/small.scss
@@ -35,6 +35,10 @@ body.small-nav {
     display: none;
   }
 
+  #sidebar {
+    resize: none;
+  }
+
   nav.primary {
     padding: 0;
 


tomhughes:
Yes `small.scss` is for small screens.

The big problem with this is that expanding the sidebar pushes the map to the side instead of expanding over it in the way the existing sidebar does when it opens.

gravitystorm:
Yeah, unfortunately this change doesn't quite work as I think was intended. In particular it can cause breakage on the map, where if you e.g. expand the sidebar, shrink the browser window, and then shrink the sidebar then the map tiles stop loading on part of the screen.

I think this is because resizing the sidebar, and so resizing the map, confuses leaflet as to its extents and/or centre of the map. Normally when the sidebar appears we need to do various things in javascript to recalculate the size of the map - see the [setSidebarOverlaid function](https://github.com/openstreetmap/openstreetmap-website/blob/45487e1ff8e7d334792440a2b5b3b1067269bfaf/app/assets/javascripts/leaflet.map.js#L302-L313). In fact, since there are dimensions hardcoded into that javascript, that might by and underlying cause of the problems. So without additional changes to the JS, a just-CSS approach isn't going to work, or at least won't work without some refactoring first.

I think also we would need to discuss the "why" part of this. I mean, I'd personally like a wider sidebar, and maybe one that was even wider on a very wide screen. But I wouldn't want one that resets back to narrow on every page load - I would want it to persist. And personally I don't need full freedom to choose arbitrary widths, I'd rather it just worked(tm). And finally I'd probably lean towards some kind of minimum width, since this PR allows you to resize to tiny widths and everything goes a bit broken at that point.

So @Strubbl thank you for the PR, although I'm not going to merge it. I'd be happy to look at a new PR which, unfortunately, would have to deal with the JS/map-related consequences of resizing the sidebar. And even if you don't want to go down that route, I'm interested to know the "why" for resizing the sidebar, i.e. what doesn't work right now for you, to see if there are alternative ways to fix it that might be less complex than diving into the JS and Leaflet behaviour.

Strubbl:
Thanks for the feedback. Using more space in the sidebar is nice, when you have got a wide screen. This way you need less scrolling in some situations, where the content of the sidebar is in focus instead of the map - so when i am a map editor instead of a "simple" map user. It also helps while reading and writing changeset comments in case the discussion consists of some more sentences/paragraphs.

PR #3469, opened 2022-02-23
Remove default values from id columns

gravitystorm:
In both the case of primary keys, and also foreign key references, there's no need to set a default value.

This doesn't have a big impact in routine situations, but can be very confusing when debugging corner cases.


tomhughes:
Granted it's a reasonable change but migrating those tables may be a challenge...

gravitystorm:
I thought this would be safe since it only changes the default values used for future INSERTS, it shouldn't be modifying any existing values.

tomhughes:
I don't think it will take long but it probably needs to take a brief lock on the table to change the metadata which can be a challenge if the contention for the table is high.

At least it's not the user table - that one really is a pain to migrate!

PR #3470, opened 2022-02-23
Enable active_record.belongs_to_required_by_default

gravitystorm:
This switches the logic so that `belongs_to` parent objects must exist by default, and marks the optional ones explicitly. This is reflected in the null/not_null status on the relevant db columns.

Doing this exposed a small corner case where we were relying on the previous default (of everything being optional) in order for trace objects to pass validation. When the `belongs_to` was made required by default, rails couldn't autodetect the inverse relationship, and so thought that all tracetags for new traces were invalid. Adding inverse_of fixed this. See also https://guides.rubyonrails.org/association_basics.html#bi-directional-associations for a better explanation.

### DIFF
--- a/app/models/access_token.rb
+++ b/app/models/access_token.rb
@@ -36,8 +36,8 @@
 #
 
 class AccessToken < OauthToken
-  belongs_to :user
-  belongs_to :client_application
+  belongs_to :user, :optional => true
+  belongs_to :client_application, :optional => true
 
   scope :valid, -> { where(:invalidated_at => nil) }
 
--- a/app/models/client_application.rb
+++ b/app/models/client_application.rb
@@ -31,7 +31,7 @@
 #
 
 class ClientApplication < ApplicationRecord
-  belongs_to :user
+  belongs_to :user, :optional => true
   has_many :tokens, :class_name => "OauthToken", :dependent => :delete_all
   has_many :access_tokens
   has_many :oauth2_verifiers
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -32,9 +32,9 @@
 
 class Issue < ApplicationRecord
   belongs_to :reportable, :polymorphic => true
-  belongs_to :reported_user, :class_name => "User"
-  belongs_to :user_resolved, :class_name => "User", :foreign_key => :resolved_by
-  belongs_to :user_updated, :class_name => "User", :foreign_key => :updated_by
+  belongs_to :reported_user, :class_name => "User", :optional => true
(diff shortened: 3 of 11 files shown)


app/models/access_token.rb
```diff
@@ -36,8 +36,8 @@
 #
 
 class AccessToken < OauthToken
-  belongs_to :user
-  belongs_to :client_application
+  belongs_to :user, :optional => true
+  belongs_to :client_application, :optional => true
```
tomhughes, line 40: Are these really optional? A token that isn't associated with a user or application?

app/models/client_application.rb
```diff
@@ -31,7 +31,7 @@
 #
 
 class ClientApplication < ApplicationRecord
-  belongs_to :user
+  belongs_to :user, :optional => true
```
tomhughes, line 34: Again I'm pretty sure all applications should be associated with a user.

app/models/issue.rb
```diff
@@ -32,9 +32,9 @@
 
 class Issue < ApplicationRecord
   belongs_to :reportable, :polymorphic => true
-  belongs_to :reported_user, :class_name => "User"
-  belongs_to :user_resolved, :class_name => "User", :foreign_key => :resolved_by
-  belongs_to :user_updated, :class_name => "User", :foreign_key => :updated_by
+  belongs_to :reported_user, :class_name => "User", :optional => true
```
tomhughes, line 35: Is this not set for all issues?

app/models/oauth_token.rb
```diff
@@ -36,8 +36,8 @@
 #
 
 class OauthToken < ApplicationRecord
-  belongs_to :client_application
-  belongs_to :user
+  belongs_to :client_application, :optional => true
+  belongs_to :user, :optional => true
```
tomhughes, line 40: As with `AccessToken` these both look like they should be required?

app/models/issue.rb
tomhughes: So `reported_user` can be null if an anonymous note is reported.

app/models/issue.rb
```diff
@@ -32,9 +32,9 @@
...
+  belongs_to :user_resolved, :class_name => "User", :foreign_key => :resolved_by, :optional => true
+  belongs_to :user_updated, :class_name => "User", :foreign_key => :updated_by, :optional => true
```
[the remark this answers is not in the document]
tomhughes: Look like the initial report doesn't count so this can be null.

tomhughes:
That only leaves the various oauth ones which are mostly as they are because that was the default from a third party component.

We can always fix them later if we want so I've merged it as is.

PR #3479, opened 2022-02-28
Fix opensearch icon & update attribution

### LINKED ISSUE #3478: OSM opensearch config uses an old/ugly logo
### URL

https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/opensearch/osm.xml

- open openstreetmap.org on Firefox
- click in the search bar
- press the OSM logo with the green '+', under the section titled 'This time, search with:'

Screenshot 2022-02-28 at 11 11 15

endim8:
Closes #3478 
this PR fixes the Opensearch icon and metadata

icon used to look like this:
Screenshot 2022-02-28 at 11 11 15
now icon looks like this: 
Screenshot 2022-02-28 at 13 14 54

Signed-off-by: Harry Bond <endim8@pm.me>


tomhughes:
Why change from PNG format to the massively larger X icon format?

endim8:
> Why change from PNG format to the massively larger ICO format?

Ah, I didn't realise ICO was bigger, I used it because the most suitable icon I could find was [osm.org/favicon.ico](https://www.osm.org/favicon.ico). The image is also bumped up to 48px instead of 16px though, which will result in a bigger file.

I'll try reducing the size 👍

endim8:
> Why change from PNG format to the massively larger ICO format?

Alright, I've made it as small as possible. The image is ~2.8x the filesize, however I think a 32px icon is worth that, considering that without compression it would be 4x the filesize.

PR #3486, opened 2022-03-02
Switch to using vips for image processing

tomhughes:
This switches to using vips (the rails 7 default) for image processing.

It also drops Ubuntu 18.04 support as the libvips it has is too old - we're coming up on the next LTS release anyway so it's probably time to let one go.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ RUN apt-get update \
       libmagickwand-dev \
       libpq-dev \
       libsasl2-dev \
+      libvips-dev \
       libxml2-dev \
       libxslt1-dev \
       locales \
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,9 @@ RUN apt-get update \
       default-jre-headless \
       file \
       firefox-geckodriver \
-      imagemagick \
       libarchive-dev \
       libffi-dev \
       libgd-dev \
-      libmagickwand-dev \
       libpq-dev \
       libsasl2-dev \
       libvips-dev \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN apt-get update \
       libmagickwand-dev \
       libpq-dev \
       libsasl2-dev \
+      libvips-dev \
```
gravitystorm, line 20: Do we still need the imagemagick and libmagickwand-dev packages?

gravitystorm:
I've tested it out with removing the imagemagick libraries, and the tests pass (specifically on docker, which I think doesn't have the imagick libraries installed by default). So I'm fairly confident that we don't need those libraries when we're using vips. But I'm all ears if there's some reason to keep them.

I also noticed we need to add vips to the brew install list.

My additional commits can be found at https://github.com/gravitystorm/openstreetmap-website/commits/vips-nomagick

tomhughes:
The problem is that rails still pulls in rmagick even if you're not using it (as it does vips) which means you probably need the libraries to be able to install rmagick?

It's different for vips because the ruby bindings use ffi so open libvips.so at run time only if it's used.

tomhughes:
I'm surprised that it's able to install with imagemagick but I guess the tests don't lie...

gravitystorm:
> I'm surprised that it's able to install with imagemagick but I guess the tests don't lie...

I think it's pulling in mini_magick instead of rmagick, and mini_magick doesn't appear to have any c/ffi stuff in it, rather it's a wrapper around the Imagemagick command line utils. So the mini_magick gem installs cleanly, and if it's not in use then the missing binaries aren't a problem.

PR #3488, opened 2022-03-03
Make account deletion remove the email address when possible

tomhughes:
This is a first draft for extending the new account deletion logic to remove the email address when the user has no edits.

As suggested to me by @simonpoole it doesn't attempt to offer the user a choice and instead just always removes the email when the user has no edits.

There is one tricky issue which is that changing `remove_personal_data` revealed that the use of `soft_destroy` by the user factory is problematic as it assumes that the user ID is set and uses it to set the display name (and now email) but when a user is created via a factory the user will not have been saved and won't have an ID yet.

Currently I've changed the factory to use `hide` instead which only sets the status but that's possibly not ideal.


simonpoole:
I think the big question here is if other content (that is messages, blog posts and so on) should lead to us retaining the e-mail or not. As @tomhughes knows I'm very much for only retaining the data in very narrowly defined scenarios but just keeping it if there has been data contributions maybe dangerous.

Assume for example that the account has been used to threaten other users via the message system. Not retaining the e-mail in such cases makes hit and run mischief essentially untraceable. But this is something that needs input from the DWG @woodpeck (maybe simply checking if the user has been reported might be good enough).

app/views/account/deletions/show.html.erb
```diff
@@ -21,6 +21,9 @@
 <ul>
   <li><%= t ".delete_profile" %></li>
   <li><%= t ".delete_display_name" %></li>
+  <% unless current_user.changesets.exists? -%>
```
gravitystorm, line 24: My initial thought is that we should abstract this slightly, e.g. `unless current_user.must_retain_email?` or similar. As per simonpoole's comment, there might be other situations where we need to retain the email.

gravitystorm (commented):
Are we making this too complicated? This was on my todo list but I haven't fully thought it out yet. My initial reaction is that I thought we were going to end up with two states:

a) We need to retain the account, and so we retain the email address
b) We don't need to retain the account, and so the email address disappears at the point that we "properly destroy" the account.

But here we are working where we need to keep the account but don't need to keep the email address. I guess this is to avoid notes, diary entries etc disappearing. I guess we should define exactly why we need to retain the email, which I think includes for relicensing issues, right? If so, do we need to include "has a trace?", since that's user generated data the we currently license too, that's not covered by changesets.

My other thought is whether some of this could be implemented with a delay, to give time for abuse investigations, e.g. disable the account immediately and then properly destroy it after 14 days.

But again, we should probably make a list of any reasons for retaining the email address, since that might guide the implementation.

tomhughes:
So we have always said that we need to keep the email if there are edits in case of any later questions about sources etc.

I believe @simonpoole did suggest we should extend that to accounts with any content, so include diary entries etc.

simonpoole:
> Are we making this too complicated? This was on my todo list but I haven't fully thought it out yet. My initial reaction is that I thought we were going to end up with two states:
> 
> a) We need to retain the account, and so we retain the email address 
>b) We don't need to retain the account, and so the email address disappears at the point that we "properly destroy" the account.

I believe this is what we are saying, aka we need to implement b). The criteria for "We need to retain the account," is a bit open, but I don't think there's disagreement on anything else.

Hedaja:
What will happen to accounts that have been deleted before this new method will be implemented?
Because right now I have the issue that I created an account with a typo inn the name and wanted to correct it by deleting and newly registering it but the email address is blocked now.

tomhughes:
Well contact support@openstreetmap.org and they will assist you.

What to do with pre-existing records is really an operations matter and not a matter for this PR or this repository.

app/views/account/deletions/show.html.erb
tomhughes: I've rebased this on the current master and introduced the `must_retain_email?` method to centralise the logic around when it should be retained as suggested.

tomhughes:
What do we need to progress this so that I can finally get saymine.com off my back?

SomeoneElseOSM:
At https://github.com/openstreetmap/openstreetmap-website/issues/4018#issuecomment-1532654435 @tomhughes  wrote "Maybe I should say I'll agree to this if we can get some action on 3488 in return?"

Just to be clear - who would you like a response from, and what question would you like answered?  I'm asking because @simonpoole above said "But this is something that needs input from the DWG" above, in case it's us...

tomhughes:
I'd just like to get it merged somehow, so whatever it takes to get that agreed!

PR #3492, opened 2022-03-08
OAuth updates to CONFIGURE doc + cosmetic tweak

harry-wood:
Update the instructions in CONFIGURE.md on how to set-up the built-in apps OAuth settings.

Also a small cosmetic tweak in OAuth 1 settings button:

before & after:

oauth-before-and-after

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -54,21 +54,25 @@ $ bundle exec rails console
 
 ## OAuth Consumer Keys
 
-Three of the built-in applications communicate via the API, and therefore need OAuth consumer keys configured. These are:
+There are two built-in applications which communicate via the API, and therefore need OAuth consumer keys configured. These are:
 
 * iD
 * The website itself (for the Notes functionality)
 
-For example, to use the iD editor you need to register it as an OAuth application.
+To use the iD editor you need to register it as an OAuth 1 application.
 
 Do the following:
 * Log into your Rails Port instance - e.g. http://localhost:3000
 * Click on your user name to go to your user page
 * Click on "my settings" on the user page
-* Click on "oauth settings" on the My settings page
+* Click on "OAuth 1 settings" on the My settings page
 * Click on 'Register your application'.
-* Unless you have set up alternatives, use Name: "Local iD" and URL: "http://localhost:3000"
-* Check the 'modify the map' box.
+* Unless you have set up alternatives, use Name: "Local iD" and Main Application URL: "http://localhost:3000"
+* Check boxes for the following Permissions
+  * 'read their user preferences'
+  * 'modify the map'
+  * 'read their private GPS traces'
+  * 'modify notes'
 * Everything else can be left with the default blank values.
 * Click the "Register" button
 * On the next page, copy the "consumer key"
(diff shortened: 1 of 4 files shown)


app/views/oauth_clients/index.html.erb
```diff
@@ -34,10 +34,14 @@
 <p><%= t(".no_apps_html", :oauth => link_to(t(".oauth"), "https://oauth.net")) %></p>
 <% else %>
 <p><%= t ".registered_apps" %></p>
-<% @client_applications.each do |client| %>
-  <div class="client_application">
-    <%= link_to client.name, :action => :show, :id => client.id %>
-  </div>
-<% end %>
+<p>
```
gravitystorm, line 37: Should be a list (which has automatic bottom padding) rather than wrapping in a paragraph

app/views/oauth_clients/index.html.erb
```diff
@@ -34,10 +34,14 @@
...
+  <% @client_applications.each do |client| %>
+    <div class="client_application">
+      <%= link_to client.name, :action => :show, :id => client.id %>
+    </div>
+  <% end %>
+</p>
 <% end %>
-<h3><%= link_to t(".register_new"), :action => :new %></h3>
+<p>
```
gravitystorm, line 45: No need to wrap a button in a paragraph tag (and it's not really correct anyway, since the contents aren't a paragraph of text). 

The final button on the page doesn't need any bottom padding anyway.

gravitystorm (changes requested):
Just a couple of small layout/css tweaks, but otherwise looks good to me.

app/views/oauth_clients/index.html.erb
harry-wood: I was copying off the oauth2 page (another dubious use of <p&gt; there), but yeah it didn't solve my spacing issue anyway. I'll get rid of it again.

app/views/oauth_clients/index.html.erb
harry-wood: Yeah. It probably isn't a list because people registering more than one application is very rare, but ok. Let me give that a try.

gravitystorm (approved):

CONFIGURE.md
```diff
@@ -81,11 +85,32 @@ An example excerpt from settings.local.yml:
...
+* Go to "[OAuth 2 applications](http://localhost:3000/oauth2/applications)" on the My settings page.
+* Click on "Register new application".
+* Use Name: "OpenStreetMap Web Site" and Redirect URIs: "http://localhost:3000"
+* Check boxes for the following Permissions
+  * 'Modify the map'
+  * 'Modify notes'
+* On the next page, copy the "Client Secret" and "Client ID"
+* Edit config/settings.local.yml in your rails tree
+* Add the "oauth_application" configuration with the "Client ID" as the value
+* Add the "oauth_key" configuration with the "Client Secret" as the value
```
tyrasd, line 103: is this really necessary? I don't immediately find anything in the code which uses this setting (b4a1e41968e1befdf6fc14715d95549f5537abca appears to have removed previous occurrences where it was used for oauth1) and the `settings.yml` now only mentions the [`oauth_application`](https://github.com/openstreetmap/openstreetmap-website/blob/73f0be96472a8d021a6347953192ca7b6ec509d7/config/settings.yml#L82-L83) setting. I tried running the website locally without the `oauth_key` setting and it seems to work just fine.

PR #3503, opened 2022-03-17
Fix SVG query icon font

mxdanger:
The question mark on the query icon is using a font that does not exist and is defaulting to serif. This fix uses the built-in web safe Helvetica with sans-serif fallback. This should match the PNG version of the icon.

Issue:
image


tomhughes:
Is the SVG actually used in production? or is it just the source that was used to generate the bitmaps?

mxdanger:
Yes, by default most icons on the website are SVG with PNG fallback.

PR #1267, opened 2016-08-13
Blacklist all Google imagery with one regex

pnorman:
This expression catches 100% of Google imagery seen in `imagery_used` with no false positives.

It won't stop someone who tries to load "https://www.google.com/maps" as imagery, but that will fail to get imagery.

Tested with JOSM, and it blocks some of the recent problems we've seen.

@pnorman
For the Data Working Group

PR #3512, opened 2022-03-26
google imagery blacklist entry update

hlfan:
Updated google imagery blacklist entry to include the pb implementation of the vt endpoint for URLs like these:
[https://www.google.at/maps/vt?pb=!1m4!1m3!1e{z}!2i{x}!3u{y}!2m2!1e1!2sm!3b0](https://www.openstreetmap.org/id#background=custom:https://www.google.at/maps/vt%3Fpb%3D!1m4!1m3!1e{z}!2i{x}!3u{y}!2m2!1e1!2sm!3b0)
The extension of the regular expression is specifically targeted to googles proprietary parameter design, as explored in [this stack overflow post](https://stackoverflow.com/questions/47017387/decoding-the-google-maps-embedded-parameters).

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -86,8 +86,8 @@ default_editor: "id"
 # Imagery to return in capabilities as blacklisted
 imagery_blacklist:
   # Current Google imagery URLs have google or googleapis in the domain
-  # with a vt or kh endpoint, and x, y and z query parameters
-  - ".*\\.google(apis)?\\..*/(vt|kh)[\\?/].*([xyz]=.*){3}.*"
+  # with a vt or kh endpoint, and x, y and z or pb-based query parameters
+  - ".*\\.google(apis)?\\..*/(vt|kh)[\\?/].*([xyz]=.*|!\d[a-z]\d+){3}.*"
   # Blacklist VWorld
   - "http://xdworld\\.vworld\\.kr:8080/.*"
   # Blacklist here


pnorman:
Are URLs like that technically viewable in editors right now?

gravitystorm:
> Are URLs like that technically viewable in editors right now?

Yes, I didn't realise it but in the original comment the link actually launches iD and demonstrates the problem.

gravitystorm:
So I'm happy to merge this, but I would like to revisit some of the conversation in https://github.com/openstreetmap/openstreetmap-website/pull/1267 , namely

> If we were happy to only match the domain before then why not stick with that but broaden the domain match? Why start trying to (badly) match the rest of the URL as well, which just means it will break every time google tweak their URLs... 

The currently proposed regex is even more complex than the complex one we had already. So in my view we should choose between these:

* If we're trying to stop everything from the google/googleapis domains, then we don't need the complexity in the path matching. 
* If we're trying to block the paths selectively, while permitting certain other paths to continue to work, then we need to add unit tests to a) illustrate which paths are supported and b) ensure these continue to work when the regex is next refactored.

I suspect we can just block all paths though, and therefore simplify the regex substantially. Thoughts?

hlfan:
Since the main objective here is to disallow unintentional copyright violation (this won’t stop deliberate maneuvering around the blacklist), blocking all Google related imagery could be split in two parts:
1. All data from the domain could be blocked: `".*\.google(apis)?\..*/.*"`, since I couldn't come up with reasonable URLs which would additionally be blocked because of this. Notice that [most links just including "google"](https://www.openstreetmap.org/id#background=custom:https://gibs.earthdata.nasa.gov/wmts/epsg3857/all/VIIRS_SNPP_CorrectedReflectance_TrueColor/default/2022-04-08/GoogleMapsCompatible_Level9/{z}/{y}/{x}.jpg) aren't matched.
2. Additionally, in a second entry, Google's embedded parameters could also be matched: `".*pb=.*(!\d[a-z]\d+){3,}.*"`, which could serve as a failsafe if IP addresses, redirects or other tricks out of scope are used.

This double filter would only let the vt or kh endpoints slip through, if the URL is deliberately crafted excluding the blacklisted domains.
I'm providing variant A (matching domain and parameters, preferred) and variant B (only matching the domain) in reverse order.
So I suggest merging version A, and if someone does find permitted imagery/data matched by these filters they can be reduced to version B or discussed again.

gravitystorm:
I've gone for the (less complex) option of just blocking by domain. If we need the extra complexity of either:
* selectively blocking endpoints in the google domains, or
* blocking those endpoints in other domains

... we can reassess (and add some tests).

ramSeraph:
the latest change blocks imagery hosted by users in google cloud storage..  they usually take the form "https://storage.googleapis.com/bucket_name/obj_name"

hlfan:
So reassessment time it is!

As @gravitystorm noted, a unit test for specificly allowing the [storage.google.com subdomain](https://www.osm.org/id#background=custom:https://storage.googleapis.com/soi_data/export/tiles/{z}/{x}/{y}.webp&map=6/22/80) is needed.
A negative lookaround in `.*(?<!storage)\.google(apis)?\..*/.*` seems to allow that.
That said, [Safari (and IE) doesn't support lookbehinds yet](https://caniuse.com/js-regexp-lookbehind).
Assuming this regex works for most OSM-contributers needing this exeption and there isn't a less complex option I'm missing, the regex above should do it.

ramSeraph:
Will the lack of backtracking regex actually make ID/openstreetmap website unusable from safari/IE? In that case maybe a new whitelist can be introduced and we can let the other clients pick up the whitelist at their own pace?

planemad:
> Assuming this regex works for most OSM-contributers needing this exeption and there isn't a less complex option I'm missing, the regex above should do it.

@hlfan This seems like the case and would be the best option to implement.

PR #3514, opened 2022-03-26
Various enlightments

tiziodcaio:
- Update to Yarn 3
- Minor changes to workflows
- Integrate default suggested config


Dockerfile
```diff
@@ -47,4 +47,4 @@ RUN gem install bundler \
 # Install NodeJS packages using yarnpkg
```
tordans, line 47: That is outdated then?

tiziodcaio: Not really, yarn v3 uses as a base yarn classic (v1). The js huge code you have found is executed by yarn classic and so using `yarnpkg` (which is simlinked to yarn) will, at the same way, work

tomhughes:
Please don't mix multiple unrelated changes in a single pull request - open a separate request for each change please so that the they can be reviewed and discussed on their own merits.

Issue #3275, opened 2021-07-28
Resolve problem with js-cookie 3.x and node engines


gravitystorm:
js-cookie 3.x has a dependency of nodejs >= 12. However, Ubuntu 20.04 ships with nodejs 10. We generally target the latest Ubuntu LTS, and this is also what we use for CI.

The upgrade to js-cookie 3.x was in 6feb125f4f0393bcfd319c426adb0b5f7e3c1c0f and as far as I'm aware was just for dependabot reasons rather than any required feature or bugfix. 102f3a6668665774f069a8116a251a28c3c5c0cc patched bin/yarn to ignore engines, and I can make a PR to the Dockerfile to do the same there.

I just wanted to open this discussion to check if ignoring the engines is the right approach, or if there are other workflows that might break (e.g. for other developers). As far as I can tell, for the js-cookie node module we're just using yarn/nodejs for code distribution and not otherwise using node for anything clever. But we use other node modules where we actually run the code in node (e.g. `eslint` in development, or `terser` in production) and where the engine used is more likely to be important. By ignoring the node engines, are we setting ourselves up for a problem later on e.g. when eslint or terser uses features from node 16 or 18 and we don't realise because we've disabled the engine check?

So options include:
* Pinning js-cookie to 2.x and re-enabling the engine check?
* Disabling the engine check in the Docker CI to match what we're doing in bin/yarn?
* Re-enabling engine checks when we move to Ubuntu 22.04?

tomhughes:
Well the node engine is irrelevant for something that's going to be sent to the browser...

tomhughes:
Pinning `js-cookie` is no long term solution unless upstream are going to maintain that branch - if a security issue is discovered and only fixed in 3.x we might be forced to upgrade at short notice.

You could argue it's a bit bizarre for `js-cookie` to require a given node engine given it's code that is only ever likely to be executed in a browser and not server side with node...

tiziodcaio:
I think the best thing to do is to upgrade node, the version 10 is in EOL... I cannot understand why focal fossa still uses this version now that the dependencies' requisites are higher

tomhughes:
We don't have a choice about node versions - we get what the operating system give us.

We're not going to start installing node from upstream just for this!

tiziodcaio:
Why not using the tag rolling which provides ubuntu 21.10

tiziodcaio:
[...]

PR #3515, opened 2022-03-28
Upgrade to Yarn v3

tiziodcaio:
Please see #3514


gravitystorm:
Thanks for splitting this into a separate PR, since changes to yarn are unrelated to change to bundler.

At first glance I have some serious concerns about this PR, so it would be good to update this PR with a lot more information. For example, what is the purpose of the upgrade? Does it fix anything in particular, is the old version of yarn blocking you from something, or is it just a maintenance task, or something else? (All are valid answers!)

Additionally:
* What is the 2.0MB of javascript in .yarn/releases ? Is this a full copy of yarn? If so, why do we need to have it in our own repository? We try to avoid copy+pasting external dependencies into this repo, and aim to only have code here that we intend to maintain ourselves. Why not use distribution packages for yarn like we do now?
* What implications are there for moving away from node_modules? Are all of our modules going to work under the alternative yarn system? Will it impact us when we want to add more modules in the future?
* Why is the --ignore-engines flag removed? It was added for a reason (see https://github.com/openstreetmap/openstreetmap-website/issues/3275 ) and so removing it should explain why it's no longer necessary.

tiziodcaio:
1. Yarn 1 could be deprecated in any moment and it is in maintenance mode, and the new lockfile format can be more safe: in fact it doesn't use the complete urls of the remotes, but it pulls everything from npm's database; if you want to use custom repos you have to use the complete link in the package.json. So yes it could be even a maintenance task 
2. The v3 and v2 versions of yarn are ugly for this: if you upgrade you have to integrate the js code of yarn itself in your repository. In fact for upgrade i used `yarn set version berry` which pulls the js code from online to your repo. At the moment the only alternative should be [corepack](https://nodejs.org/dist/latest-v17.x/docs/api/corepack.html).
3. No, you can use node_modules as usual it is setted in the way to be retrocompatible
4. I will write now in this issue, because, if I'm right, nodejs 10 is in EOL at the moment. The flag it is removed from yarn, so it gives error trying

tiziodcaio:
I checked the workflows before start push, so they all work... Now we should understand why docker fails... I haven't so much experience with it and I might take a bit longer to understand...

tomhughes:
So if the node modules are still getting installed into `node_modules` why have you stopped the actions from caching the installed modules?

tomhughes:
Also unless I'm mistaken all of this only works for the docker version, which for some reason seems to be running `yarnpkg` directly instead of doing `bundle exec rails yarn:install` which is what it should be doing and which would run whichever `yarn` was on the path.

gravitystorm:
I've been doing some research on how Debian (and therefore Ubuntu) intend to package yarn 2/3 in the future, so that we can align with them. Two useful links I found are:

* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=980316#24
* https://github.com/nodejs/corepack/issues/76

So it looks like eventually we can expect corepack to be packaged, and it will take care of yarn 2/3 in the usual manner (i.e. corepack treats yarn a bit like other node modules, on a per-repo basis). Unfortunately the Debian Node 16 package (in experimental) doesn't come with corepack, and I can't find any indications in https://salsa.debian.org/groups/js-team or on the other debian bug tracker that indicate that anyone is working on it. So it could be a long time before corepack is available through Ubuntu packages.

Since yarn 1.x is still supported (i.e. getting security fixes, and is available in Ubuntu packages), I think we should pause this upgrade. My plan would be to stick with yarn 1.x from packages until we move to Node 16, and then use yarn via corepack. (This means we skip over the intermediate stage, where we need to use yarn 1 to install yarn 2/3 and then commit yarn 2/3 to this repo.)

I'm happy to hear alternative views though!

tiziodcaio:
> So it looks like eventually we can expect corepack to be packaged, and it will take care of yarn 2/3 in the usual manner (i.e. corepack treats yarn a bit like other node modules, on a per-repo basis). Unfortunately the Debian Node 16 package (in experimental) doesn't come with corepack, and I can't find any indications in https://salsa.debian.org/groups/js-team or on the other debian bug tracker that indicate that anyone is working on it. So it could be a long time before corepack is available through Ubuntu packages.
> 
> Since yarn 1.x is still supported (i.e. getting security fixes, and is available in Ubuntu packages), I think we should pause this upgrade. My plan would be to stick with yarn 1.x from packages until we move to Node 16, and then use yarn via corepack. (This means we skip over the intermediate stage, where we need to use yarn 1 to install yarn 2/3 and then commit yarn 2/3 to this repo.)

I know corepack obviously.... Because I cannot upgrade nodejs I preferred chose a compromise between parts: corepack will be installed using npm, so we don't have to store the yarn's cjs huge blob!

tiziodcaio:
>     * Why is the --ignore-engines flag removed? It was added for a reason (see [Resolve problem with js-cookie 3.x and node engines #3275](https://github.com/openstreetmap/openstreetmap-website/issues/3275) ) and so removing it should explain why it's no longer necessary.

At the moment (in yarn berry) engines checks are disabled by default and I have to add a plugin to re-enable them... yarnpkg/berry#1177

gravitystorm:
@tiziodcaio Thanks for all your work so far on this. However, I'm going to put this upgrade on hold for now, and close this PR, since the upgrade currently causes more problems than it fixes.

* We should wait until dependabot supports yarn 2/3. We use dependabot on this project for the ruby side of things, and we're familiar with it from other projects too. Also given the linked issue has Github staff say they are working on it, then we should wait rather than either disabling it or moving to another provider.
* The ignore-engines flag (or rather, the fact that yarn 2/3 doesn't support node engine restrictions) is a bit of a red flag for me that yarn 2/3 isn't mature. If we need to add plugins to yarn (e.g. yarn-plugin-engines) then that's extra complexity since yarn 1 provides the same functionality built-in.
* I think the approach of "install npm to install corepack to install yarn to install the packages" is both a) the right approach and also b) a complete mess. As I said before, we should wait either until we are using a version of node that has corepack built-in (debian/ubuntu packaging caveats apply), or until there's a stronger reason to go down this route. Node is a small part of what we deal with and so all this installation complexity is bad. I want our project-wide installation to be as easy as possible for our contributors, very few of whom are dealing with anything node-related.
* We have no need to move away from yarn 1 yet.
* There is no compelling reason to move to yarn 2/3 yet.

There's other things I noticed in reviewing this PR, but I'm not asking for them to be fixed now given my comments above:

* There is a `nodeLinker: node-modules` configuration, but the github actions caching has all been changed so `node_modules` won't get cached. That should be resolved.
* There is a change to the eslint version, but that's not explained.
* The multi-platform installation notes (INSTALL.md) and the vagrant provisioning script (script/vagrant/setup/provision.sh) also need to be aligned with all these installation changes.

As I said at the top, thank you for your work on this, and I think it will come in useful in the future when the situation changes.

PR #3521, opened 2022-04-03
add banner for Russia-Ukraine war

bezdna:


Firefishy:
We (the operations group) don't yet have a policy for carrying a message like this, but I think it is an important request from the Ukraine mappers.

We will need some time to discuss it.

bezdna:
> We (the operations group) don't yet have a policy for carrying a message like this, but I think it is an important request from the Ukraine mappers.
> 
> We will need some time to discuss it.

Time is people's lives.
Either people stop the war, or they support it.
Alas, there is no other tone.

tomhughes:
I don't really think operations should be the ones to decide, and certainly the maintainer of this repository shouldn't be.

woodpeck:
I think that it would be detrimental to Ukrainian if editing in Ukraine were stopped. Humanitarian mappers say that OSM saves lives during crises and I am inclined to believe them. I think it is more likely that the average Ukrainian citizen benefits from up-to-date information in OSM, and less likely that the invading army benefits.

bezdna:
> I think that it would be detrimental to Ukrainian if editing in Ukraine were stopped. Humanitarian mappers say that OSM saves lives during crises and I am inclined to believe them. I think it is more likely that the average Ukrainian citizen benefits from up-to-date information in OSM, and less likely that the invading army benefits.

There is no benefit from mapping military bases, airports, oil depots and the like.
namely, this is sufficiently plotted on the OSM map.
The humanitarian mission of the OCM will be possible when the war ends, and during the active phase, the OCM becomes an assistant to the war!

bezdna:
file:///home/anton/Загрузки/25db0b8f4c62742a343cc32dea85f9be02b1c63b_1920_1080.jpg.webp-portal

bezdna:
25db0b8f4c62742a343cc32dea85f9be02b1c63b_1920_1080 jpg

woodpeck:
> that's enough to think about 

If the the map in the background of that picture is supposed to "prove" anything then it proves that the Russians are using a map generated from OSM data no newer than November 2018 (which is when OSM changed the border in the Sea of Asov (https://pewu.github.io/osm-history/#/way/218559335). If that's the case, Ukrainians can happily go on mapping and use OSM to their advantage - the changes will not appear on *that* map.

woodpeck:
> There is no benefit from mapping military bases, airports, oil depots and the like.

Your pull request does not ask people to stop mapping these features - it asks people to stop mapping anything.

bezdna:
> 



> Your pull request does not ask people to stop mapping these features - it asks people to stop mapping anything.

yes, because mapping even some tree line next to an airport or other infrastructure facility, or just at some dominant height can be quite critical.
In this video, you can clearly see what they were hiding in the tree line:
https://www.youtube.com/watch?v=en-ga97tk0A 

and there can be a huge number of such objects on the map, which can now be used by the military.

bezdna:
> If the the map in the background of that picture is supposed to "prove" anything then it proves that the Russians are using a map generated from OSM data no newer than November 2018 (which is when OSM changed the border in the Sea of Asov (https://pewu.github.io/osm-history/#/way/218559335). If that's the case, Ukrainians can happily go on mapping and use OSM to their advantage - the changes will not appear on _that_ map.

recent mapping of meat depots and military facilities before launching missile strikes suggests that OSM can be used just to update current tasks.

So the picture may be old, and the data may be new.

And if there is even the slightest hint of the use of OSM for the military, then I would insist that OSM does more harm than good.
Since you can find out where the store or help center for refugees is, simply in human language, and not using a map.

gravitystorm:
I defer the decision on adding this banner to the OSMF, either to the OWG or the Foundation Board, if they think that is a more suitable group to make the decision. It's well outside of our normal banner policy, so I think it's right for OSMF to decide on this particular case. I would suggest contacting both groups via email, if you haven't done so already. They can then make a speedy decision on this matter.

Beyond that, please do not discuss the pros and cons of mapping during wartime here, this is not a suitable venue for that discussion and nothing constructive will come of it. This conversation should be restricted to approval from the OSMF, and any technical matters regarding this PR.

Andygol:
I just leave this link for visibility.
https://wiki.openstreetmap.org/wiki/Russian-Ukrainian_war

gravitystorm:
Since it has been 3 weeks now without any word from the OSMF Board / OWG, and there's nothing else that I can action here, I'm going to close this PR. 

I'm very happy to re-open and/or merge this if/when they make any decisions.

PR #3523, opened 2022-04-06
Rubocop autofix: ambiguous operator precedence

gravitystorm:
This simply adds braces to clarify which mathematical operator comes first.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -21,19 +21,6 @@ require:
 Layout/LineLength:
   Max: 270
 
-# Offense count: 62
-# Cop supports --auto-correct.
-Lint/AmbiguousOperatorPrecedence:
-  Exclude:
-    - 'app/controllers/geocoder_controller.rb'
-    - 'app/models/user.rb'
-    - 'lib/bounding_box.rb'
-    - 'lib/osm.rb'
-    - 'lib/rich_text.rb'
-    - 'lib/short_link.rb'
-    - 'test/controllers/api/old_nodes_controller_test.rb'
-    - 'test/lib/short_link_test.rb'
-
 # Offense count: 34
 # Configuration parameters: AllowSafeAssignment.
 Lint/AssignmentInCondition:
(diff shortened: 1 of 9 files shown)


tomhughes:
I deliberately avoided doing this because it just annoys me to add such ridiculously verbose nonsense in case somebody somewhere doesn't know basic arithmetic precedence rules.

I mean sure there are cases that people don't remember and where it's wise to be explicit, but a chain of identical arithmetic operators is not that case!

gravitystorm:
Sure, I can see that and I can usually remember the rules too. But I think it becomes more useful here to use brackets when we're not talking about simple mathematical symbols. There's a fair amount of complexity when reading things like:

```
... captures[1].to_f 
... captures[1].to_f + captures[2].to_f
... captures[1].to_f + captures[2].to_f / 60
```
I mean, you need to read the captures bit, then think about the array access, then converting it to a float, and then keep on reading in case there's a higher-precedence operator still to follow. Similarly for the max_friends calculations:

```
max_friends = account_age_in_hours.ceil + recent_friends - active_reports * 10
```
It's a lot of stuff to read and mentally juggle before you get to the multiplication, because each part of the calculation is not just a simple symbol like `x y z i j k` etc.

So I think the braces are worthwhile in giving a heads-up that something is being done out-of-order, given the complexity of reading some of the components in these lines. And I don't think the extraneous braces make the other lines any harder to read.

tomhughes:
I think I misread some of the changes the other day - it doesn't actually bracket chains of identical ops as I suggested.

I have to disagree about the readability though - the huge numbers of parentheses actually slows me down when I try and read those expressions as I have to parse them all and try and understand what they're doing and where the matching open is when you get three or four closes in a row only to find it doesn't really do anything.

That may be me though - because I'm not used to using them in basic arithmetic expressions when I do see them I assume something unusual is going on and start trying to understand how the default precedence is being overridden.

Mostly I just find the results ugly though so I guess I don't really have a good argument for not doing it.

PR #52, opened 2012-07-09
Meaningful and disambiguating class/type labels

dittaeva:
Many of the class/type labels do not make sense for the user unless they are well familiar with the OSM map features tagging scheme. I have gone through the list and changed the most obvious ones that do not make sense on their own, or need disambiguating from other labels. Have also made some other adjustments.

Most are problematic because they aren't nouns. It does not make sense to have a name on "Residential" or "Conservation", and "Construction" on its own sounds more like a building than an area.

I've left out the following problematic labels:
`amenity:bureau_de_change` You could probably write English with only French words, but it doesn't make it understandable. Let the French translation use French.
`amenity:hall` what is this, a building?
`amenity:preschool` [On the wiki, this is said to be equal to kindergarten](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dkindergarten).
`historic:*` The first historic items could probably benefit from a "Historic " prefix.
`waterway:lock` The same as lock gate?

PR #3536, opened 2022-04-25
Meaningful and disambiguating class/type labels (from 2012)

JesseWeinstein:
I've fixed the merge conflicts for the ancient (2012 (!)) PR by @dittaeva https://github.com/openstreetmap/openstreetmap-website/pull/52 ; it seems to me that the name corrections are still helpful.


dittaeva:
Thanks Jesse! I remember doing the patch, and it is really nice that it finally is put to use.

config/locales/en.yml
```diff
@@ -661,12 +661,12 @@ en:
           driving_school: "Driving School"
           embassy: "Embassy"
           events_venue: "Events Venue"
-          fast_food: "Fast Food"
+          fast_food: "Fast Food Restaurant"
```
gravitystorm, line 664: This should be removed, since amenity=fast_food also covers food trucks and takeaway only outlets.

config/locales/en.yml
```diff
@@ -931,9 +931,9 @@ en:
           brownfield: "Brownfield Land"
           cemetery: "Cemetery"
           commercial: "Commercial Area"
-          conservation: "Conservation"
-          construction: "Construction"
-          farm: "Farm"
+          conservation: "Conservation Area"
+          construction: "Construction Area"
+          farm: "Farmland"
```
gravitystorm, line 936: This key is deprecated and should be removed rather than changed. There's only a few instances of it remaining in the database in 2022 anyway.

config/locales/en.yml
```diff
@@ -1228,8 +1228,8 @@ en:
           deli: "Deli"
           department_store: "Department Store"
           discount: "Discount Items Shop"
-          doityourself: "Do-It-Yourself"
-          dry_cleaning: "Dry Cleaning"
+          doityourself: "Do-It-Yourself Shop"
+          dry_cleaning: "Dry Cleaner"
```
gravitystorm, line 1232: The "Dry cleaning" term is the one exclusively used in the OSM wiki, so I think it's fine to keep the original

gravitystorm (changes requested):
I'm generally happy with the PR, and neutral on most of the changes, but I've flagged a few that need further attention.

JesseWeinstein:
Reverted the four that you flagged (thanks)! Let me know if there are any other changes requested.

gravitystorm (approved):

PR #3539, opened 2022-04-29
Kramdown formatting on reports text

harry-wood:
Pass the text of reports ('details' field) through the RichText formatter to give us kramdown formatting support.

Users don't generally get to see their reports. Only a few admins ever see them displayed, but DWG folks working with reports have complained of a **"wall of text"** problem when looking at some longer reports. This solves it, particularly with basic paragraphs rendered from double-newlines. Other kramdown tricks are a bonus:

before-after

We don't need to do support of alternative HTML formatting as we did for legacy diary entries. I don't think we need any special treatment for old reports in the database. We can bung them all through the kramdown formatter and it will either make no difference or will improve the display.

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -36,4 +36,8 @@ class Report < ApplicationRecord
     else %w[other]
     end
   end
+
+  def details
+    RichText.new("markdown", self[:details])
+  end
 end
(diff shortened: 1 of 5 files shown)


mmd-osm:
NB: EWG put this on their list of projects they wanted to fund. Added your pull request to their list, as it’s already done now.

https://pads.ccc.de/3kCiDBchei

PR #3542, opened 2022-05-04
Fix verification redirect on missing report category

harry-wood:
Fix an issue with the report verification, causing it to lose the text of a report if the user fails to fill in a category. In the absence of the required field 'category' it redirects to the new report form. This redirect should retain the value of the 'details' field.

### DIFF
--- a/test/controllers/reports_controller_test.rb
+++ b/test/controllers/reports_controller_test.rb
@@ -63,10 +63,12 @@ class ReportsControllerTest < ActionDispatch::IntegrationTest
       category = "other"
       post reports_path(:report => {
                           :category => category,
-                          :issue => { :reportable_id => 1, :reportable_type => "User" }
+                          :issue => { :reportable_id => target_user.id, :reportable_type => "User" }
                         })
     end
-    assert_response :redirect
+    assert_response :success
+    assert_template :new
+    assert_match(/Please provide the required details/, flash[:notice])
 
     assert_equal 1, issue.reports.count
   end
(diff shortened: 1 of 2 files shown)


harry-wood:
One inelegant thing about this fix, is that the textbox value (which can easily be very long) will end up passed in the URL after the redirect.
Screenshot 2022-05-04 at 13 25 34
I think it may be possible to avoid doing a `redirect_to` but instead just a `render`. I could look into that, but... maybe it's no big deal.

tomhughes:
My immediate inclination was to do a `render` to avoid the parameters in URL issue certainly but @gravitystorm probably knows better than me what the correct rails way to handle this is given it's normal to separate the form creation and form response processing into separate methods with rails.

tomhughes (changes requested):
This has a failing test due to a missing translation, though I suspect it's the result of an invalid interpolation of a `nil` value as the missing translation is `en.reports.new.categories.nil_class.other_label`.

test/controllers/reports_controller_test.rb
```diff
@@ -63,10 +63,12 @@ def test_new_report_with_incomplete_details
       category = "other"
       post reports_path(:report => {
                           :category => category,
-                          :issue => { :reportable_id => 1, :reportable_type => "User" }
+                          :issue => { :reportable_id => target_user.id, :reportable_type => "User" }
```
harry-wood, line 66: For some reason swapping to `render` instead of redirect causes our test user to get created (or _more often_ get created) with some id other than 1
but anyway it should probably have always been `target_user.id` to make it less brittle.

gravitystorm:
I just wanted to say this is a favourite PR of mine - a simple problem, with a small obvious fix, but where the final fix ended up involved using a more idiomatic approach, less overall complexity, and fixed other things (like the error highlighting and the fragile testing) as a side effect

 :+1:

harry-wood:
Always pleasing when making things better happens by making things simpler :-D

Issue #3551, opened 2022-05-15
Guard/prevent dangerous actions like API token deletion better


rugk:
## Problem

The API token on the website is only guarded by one `confirm()` (alert/prompt) prompt.
See this screenshot:

![grafik](https://user-images.githubusercontent.com/11966684/168497274-2483835f-c5b6-43a2-9fba-0f06174eba19.png)

If you accidentally click "Delete", this has potentially _severe consequences_. All your users of the app are force logged-out and need to login. _And_ you likely need to update the code and ship an update of your application, which again may take some time. This all results in a very bad UX impediment for your users. E.g. users could be blocked for days from using your app. (until they get that update.)
For (potential) apps that make money based on the features they offer, this can result in a sales loss, in the worst case.

Note this happened IRL: https://github.com/streetcomplete/StreetComplete/issues/4028#issuecomment-1126990804

### Variant 1: Undo

UX wise I'd propose two common solutions (you may keep the current dialog, or replace them by that new ones):
* add an “Undo” feature and do not delete the token right away, but after a “cooldown period” in which users can think whether they really wanted to do what they wanted…

The thing one has to keep in mind is that, in our case, we also need to be _fast_ deleting the API token, because the reason for deletion may also be that your token was stolen or so and someone already misused it and is doing edits etc.

### Screenshots
![grafik](https://user-images.githubusercontent.com/11966684/168497556-ce2ed947-7066-479c-ab49-3fddaddf38f0.png)

```html
<div class="flash warning row mx-0 p-3 align-items-center">
    <div class="col-auto">
      <picture>
        <source srcset="/assets/notice-8562cd4b85576531716bcc1543e08f82c82b6aea8244393974b5f4f306f85897.svg" type="image/svg+xml">
        <img srcset="/assets/notice-8562cd4b85576531716bcc1543e08f82c82b6aea8244393974b5f4f306f85897.svg" src="/assets/notice-d058d399b5d6f7b617d3cf8ff36c71b1f83c51cd4de6dd8ef1a194a451d48236.png">
      </picture>
    </div>
    <div class="" style="margin-right: 10px;">Application API key will be deleted in 10s.</div>
  <a class="btn btn-outline-primary" href="/oauth2/applications/new">Undo</a>
</div>
````

Maybe better mentioning the application name:
![grafik](https://user-images.githubusercontent.com/11966684/168497640-d5683797-578f-42f2-8b93-fe4ea79eed0e.png)

[...]

PR #3552, opened 2022-05-15
Add cause to feature request issue template

rugk:
When writing https://github.com/openstreetmap/openstreetmap-website/issues/3551 I noticed that your template is somewhat problematic.

What users really should write is how/why they have the feature request they want, i.e. what (problem) is “behind” a feature request. Similar to how user stories should always go like ”I as a [who] want [what], because [why]”, here the “why” aspect was missing.
The “Description” does not really tangle that, it only tangles _what_ the user wants.

Screenshots may also be better inline inside of a text description, but I see you want to nudge the user to add screenshots or mockups, so that's fine IMHO.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature.yml
+++ b/.github/ISSUE_TEMPLATE/feature.yml
@@ -7,6 +7,14 @@ body:
   - type: textarea
     id: idea
     attributes:
+      label: Problem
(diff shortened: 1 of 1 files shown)


gravitystorm:
@rugk Thanks for your suggestion! I've merged this now.

PR #3568, opened 2022-06-10
Margins for 'comment from' / 'reported by' lines on issues page

AntonKhorev:
Don't remove margins with `mb-0` class on reported-by lines.

After enabling markdown in issue reports/comments, they can consist of several paragraphs. There's spacing between paragraphs provided by their bottom margins. But then there's a reported-by line above and it doesn't have margins. This line is displayed next to the report text without any spacing which looks strange for multi-paragraph reports.

Before merging #3566 there was a kind-of margin provided by an empty paragraph resulting from invalid html. Now that paragraph is gone.

### DIFF
--- a/app/views/issues/_comments.html.erb
+++ b/app/views/issues/_comments.html.erb
@@ -5,7 +5,7 @@
         <%= link_to user_thumbnail(comment.user), user_path(comment.user) %>
       </div>
       <div class="col">
-        <p class="text-muted mb-0">
+        <p class="text-muted">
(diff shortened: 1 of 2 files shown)


tomhughes:
Can you provide some example before/after images?

I did attempt to verify that #3566 wasn't having any noticeable effect on the layout before I merged it but apparently I failed....

AntonKhorev:
A similar line in diary comments also doesn't have a bottom margin (removed by `comment-heading` class). But changeset/note comments do (actually the paragraph below them has a top margin).

AntonKhorev:
And the diary entry byline has margins provided by its container.

PR #3575, opened 2022-06-16
Update DOCKER.md

victorovento:
Need to run docker-compose run --rm web bundle exec rake db:create before running the tests, otherwise, it will trow the next error:

We could not find your database: osm_test. Which can be found in the database configuration file located at config/database.yml. (ActiveRecord::NoDatabaseError)

To resolve this issue:

- Did you create the database for this app, or delete it? You may need to create your database.
- Has the database name changed? Check your database.yml config has the correct database name.

To create your database, run:

        bin/rails db:create

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -51,6 +51,8 @@ This will launch one Docker container for each 'service' specified in `docker-co
 Run the Rails database migrations:
 
     docker-compose run --rm web bundle exec rake db:migrate
+    
+    docker-compose run --rm web bundle exec rake db:create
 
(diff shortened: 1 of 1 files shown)


tomhughes:
How did you try to run the tests? As i recall there's a special command you need to use the first time you run the tests to create test version of the database...

lectrician1:
I also have this problem. I just followed the Docker instructions exactly and got to `docker-compose run --rm web bundle exec rails test:all` and that error showed. This PR should be pushed @tomhughes .

PR #3577, opened 2022-06-19
Fix map edit controls min-zoom behaviour

### LINKED ISSUE #3456: Edit map Dropdown and tooltip text remain displayed after zooming out.
### URL

https://www.openstreetmap.org/

MacOS 12.0.1
Chrome 98.0

See screenshot. I zoom in somewhere to edit an area. This enables the Edit button and the dropdown button next to it. I then click the dropdown to see two options available. I click the dropdown button again to close it and then zoom back out. This is supposed to disable the dropdown button as it was when I was initially zoomed out but when I click the dropdown now I'm still seeing the options available. Clicking either option doesn't seem to do anything, so these shouldn't be displayed to the user when they zoom back out.

EditDropdown

faebebin:
The dropdown and tooltip of the map edit controls should behave similar when zooming in and out. This is fixed by disabling the button-group and the buttons. Fixes  #3456

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -59,12 +59,14 @@ window.updateLinks = function (loc, zoom, layers, object) {
   var editDisabled = zoom < 13;
   $("#edit_tab")
     .tooltip({ placement: "bottom" })
-    .off("click.minzoom")
-    .on("click.minzoom", function () { return !editDisabled; })
-    .toggleClass("disabled", editDisabled)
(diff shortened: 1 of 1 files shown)


faebebin:
Need to mention slight styling changes (no more hover effect) of the disabled edit buttons. This could of course be addressed with css if wished by admins. But IMO it is even more clear like this, that those buttons are disabled. BTW I would like to contribute regularly in the future and thus will start with some bug fixing.

PR #3578, opened 2022-06-20
Note update date in xml, json, gpx api output formats

AntonKhorev:
Adds `update_date` to notes api output.

Internally, notes have updated_at, created_at and closed_at dates. Create and close dates are available through the api, update date is not. Usually you can look at the date of the last note comment to figure out update_date. Except it's not always possible. If the last comment is hidden the date is going to be incorrect.

Why do you need update date? Note search results can be sorted by update date. The number of returned notes is limited, the default limit is 100. What if you want to get the next 100 notes? You can add a `from` or `to` parameter to your search call with the date of a last note you got as a value.* That way you can skip the first 100 notes you already received.**

I named the added key/element `date_updated`, similar to existing `date_created` and `date_closed` keys/elements, if we ignore one naming inconsistency in json where `closed_at` is used instead of `date_closed`. I didn't add update date to rss, because there's no good possible element inside `<item>` for it.

\* actually date + 1 second for `to`
\** except for same-second notes

### DIFF
--- a/app/views/api/notes/_note.gpx.builder
+++ b/app/views/api/notes/_note.gpx.builder
@@ -20,6 +20,7 @@ xml.wpt("lon" => note.lon, "lat" => note.lat) do
     end
 
     xml.date_created note.created_at
+    xml.date_updated note.updated_at
     xml.status note.status
 
     xml.date_closed note.closed_at if note.closed?
--- a/app/views/api/notes/_note.json.jbuilder
+++ b/app/views/api/notes/_note.json.jbuilder
@@ -17,6 +17,7 @@ json.properties do
   end
 
   json.date_created note.created_at.to_s
+  json.date_updated note.updated_at.to_s
   json.status note.status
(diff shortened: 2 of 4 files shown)


mmd-osm:
A few notes on endpoint consumers:

* At least JOSM doesn't break when adding the extra element. Last change calculation still based on comments. -> ok
*  [StreetComplete](https://github.com/westnordost/osmapi/blob/master/libs/notes/src/main/java/de/westnordost/osmapi/notes/NotesParser.java) seems to be ok with the change   
*  [Vespucci](https://github.com/MarcusWolschon/osmeditor4android/blob/master/src/main/java/de/blau/android/tasks/Note.java#L316-L327), like JOSM, last change calculation based on comments.

Change seems ok. 

TODO: update OSM API 0.6 wiki page once merged.

tomhughes (approved):
Looks good to me.

Issue #3071, opened 2021-01-26
Usability issues with the "report" feature regarding notes


woodpeck:
As one of the people processing the "reports" that are made on the OSM web site, I notice a frequent pattern where people "report" a note instead of just closing it. DWG frequently receives reports of "this note is spam", "this note is useless", "this note refers to something already mapped", "this note has a wrong name". Sometimes, looking at the history of the note, it seems that the reporter had already resolved/closed the note, then re-activated it just to report it. 

I think that there is a two-fold issue here:

1. Many users seem not to understand that, if they encounter a useless note, they can just write a comment and close it, and instead they choose to "report" it (this may partly be a translation issue, I have a hunch that we receive more "useless note" reports from e.g .France and Asia than from English or German speaking countries);
2. Some users do close a note and are then confused by it not going away.

It would be great if we could improve user guidance here. For the 2nd issue, maybe it is possible to display a flash notice after closing a note that explains "Thank you for closing this note. It will remain on the map for another 24 hours and then go away automatically." or so. For the 1st issue, perhaps the "report this note" link should be surrounded with an explanation: "If this note contains sensitive information that needs to be removed by an admin, click here to report it. For all other cases, you can put a comment below and close the note yourself."

tomhughes:
Closed notes stay for a week don't they, not 24 hours.

harry-wood:
> re-activated it just to report it

Actually users can still report a closed note (which seems like a good thing), so that shouldn't be a reason for them to reactivate it.

I suppose reporting rather than resolving a spammy note is something someone might do if they (quite reasonably) believe that the note will be better dealt with that way. They may hope the spam will be fully deleted including from notes API calls and data dumps. Presumably we _do_ need to try to fully delete some types of bad notes (copyright violations and other illegal material), but I guess we leave a lot of spammy notes in our notes history? Is the most useful thing to do both actually? We could invite users who have reported a note to also resolve it however...

Another thing is, someone might be uncomfortable with being identified as the person who resolves a note, but happy to anonymously report it. So I don't think we could make it auto-resolve a reported note. We could make it auto-report a note which is resolved with the word "spam". More reports! which bring me on to...

Are you able to deprioritise worrying about spam reports on any resolved notes? Is that a feature idea for your view of the reports? (not sure how that works) I feel like data on reporting of old resolved notes, is still data that might have some value. It _might_ in future be acted upon to remove the spam from history, no?

woodpeck:
[...]

PR #3581, opened 2022-06-27
Wording to discourage note reporting

harry-wood:
Add some wording wrapped around the 'Report this note' link to explain to users to resolve the note instead.

Also re-arrange to place this down at the bottom with less prominence than the comment/resolve form.

Tackling the issue reported by DWG, that too many note reports are coming to them, where the users could deal with it themselves: https://github.com/openstreetmap/openstreetmap-website/issues/3071

notes-reporting-text2

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -421,7 +421,7 @@ en:
       reopened_by_html: "Reactivated by %{user} <abbr title='%{exact_time}'>%{when}</abbr>"
       reopened_by_anonymous_html: "Reactivated by anonymous <abbr title='%{exact_time}'>%{when}</abbr>"
       hidden_by_html: "Hidden by %{user} <abbr title='%{exact_time}'>%{when}</abbr>"
-      report: Report this note
+      report: report this note
       coordinates_html: "%{latitude}, %{longitude}"
     query:
       title: "Query Features"
@@ -2883,6 +2883,7 @@ en:
         reactivate: Reactivate
         comment_and_resolve: Comment & Resolve
         comment: Comment
+        report_link_html: "If this note contains sensitive information that needs to be removed, you can %{link}. For all other problems with the note, please resolve it yourself with a comment."
     edit_help: Move the map and zoom in on a location you want to edit, then click here.
     directions:
       ascend: "Ascend"
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -841,6 +841,10 @@ tr.turn:hover {
     }
   }
 
+  form.note-comment-resolve {
+    margin: 10px 0px 40px 0px;
+  }
+
   .subscribe-buttons input {
     font-size: 90%;
     line-height: 15px;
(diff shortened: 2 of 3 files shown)


config/locales/en.yml
```diff
@@ -2883,6 +2883,7 @@ en:
         reactivate: Reactivate
         comment_and_resolve: Comment & Resolve
         comment: Comment
+        report_link_html: "If this note contains sensitive information that needs to be removed by an admin, you can %{link}. For all other cases, please resolve the note yourself with a comment"
```
pnorman, line 2886: This should contain a trailing period. Additionally, I'd remove "by an admin" as it doesn't matter to the user who removes it.

pnorman (changes requested):

AntonKhorev:
Sometimes users add a note, then resolve it (without making any changes to the map), then reopen, then report.

Probably there should also be some discouragement for resolving, so users don't close the notes *because* the information in notes is correct.

harry-wood:
Yeah I guess users incorrectly "Resolve" due to all sorts of misunderstandings. Maybe the meaning of "Resolve" can be made clearer somehow. Should we have different buttons for different outcomes? Should we be allowed to resolve without commenting? Not sure. I'm not going to tackle that in this PR though.

"For all other cases" is the wording @woodpeck was suggesting (https://github.com/openstreetmap/openstreetmap-website/issues/3071). I was basically questioning this when I mentioned spam. Isn't it better to _report_ spam rather than _resolve_ it? Answer: no! At least not from the point of view of people dealing with too many reports! The wording is deliberately discouraging reporting quite strongly.

AntonKhorev:
Resolving without commenting makes sense when you resolve your own note after you made the changes to the map.

"For all other cases" wording in #3581 was for "1. ... if they encounter a useless note" situation, not for every note someone might find.

AntonKhorev:
What I just wrote basically says replace "For all other cases" with "If the note is useless", but maybe even that is not enough. This would require the user to know which notes are useless, and the kind of a user who'd report the note when they don't need to probably doesn't know that.

harry-wood:
hmm. How about instead of "For all other cases..." we could say "For all other problems..." ?

AntonKhorev:
Users have to know that there are *problems with the map* and *problems with the notes*. In case of *problems with the map* they should keep the notes about these problems open. In case of *problems with the notes* they should resolve the notes.

AntonKhorev:
"For all other problems..." -> "For all other problems *with the note*..." is a better improvement.

gravitystorm:
Thanks for the PR. I've rebased it, and reworked the styling to use bootstrap spacer utilities rather than custom CSS. And I've put the text in a `<small>` tag too, since it's not normally required reading for dealing with the note.

I hope this reduces the number of reports that DWG needs to deal with!

harry-wood:
Thanks Andy! One `<small>` step @woodpeck !

woodpeck:
Just wanted to say thank you, I think it is already having an effect!

harry-wood:
Excellent!
I was just working on a follow-up tweak https://github.com/openstreetmap/openstreetmap-website/pull/3676 (Something else to help people realise they don't need to report)

PR #3589, opened 2022-07-05
Enable brotli compression for assets

tomhughes:


tomhughes:
Ah cool. I did do some googling but failed to find that... Will have a look at reworking things.

tomhughes:
I've reworked this using that gem now.

PR #3601, opened 2022-07-16
add banner for sotm 2022

tyrasd:


pnorman:
Per the [FAQ](https://github.com/openstreetmap/openstreetmap-website/blob/master/FAQ.md) please indicate when the banner should start being shown, which must be at least 7 days from when the PR is created.

tyrasd:
Sorry, I forgot about that. Please start to show the banner on July 23.

pnorman:
I think we missed checking the contrast against the X
image

dieterdreist:
I can provide a different version, but ideally we could add a class for a darker X (have alternative css for bright and dark x)?  The hamburger icon for example is much darker and easier to see on bright backgrounds, the same color for the x would be better IMHO.   Generally the light X is not very visible on white, and almost invisible on pastel colors

PR #3605, opened 2022-07-19
Add date range parameters to notes index API call

AntonKhorev:
Make [index (bounding box) notes API call](https://wiki.openstreetmap.org/wiki/API_v0.6#Retrieving_notes_data_by_bounding_box:_GET_/api/0.6/notes) be able to receive `from` and `to` parameters similar to [search call](https://wiki.openstreetmap.org/wiki/API_v0.6#Search_for_notes:_GET_/api/0.6/notes/search).

## Use case

With date range parameters you can request more notes by excluding the already received ones. Let's say the limit was 100, you requested notes for some bbox and got 100 notes with most recent update dates. Now you know there might be more notes inside this bbox and you want to get another portion. You could increase the limit but then you'd have to redownload the notes you already received. With date range parameters you can set the upper limit* to the update date** of the last note you received and repeat the request. This way you'll skip redownloading most of the notes.

\* You also have to set the lower limit, see below.

\** Of course it would help to know the true update date: #3578.

## Details

The logic when `from`/`to` parameters present is the same as in search queries, even though I find it strange. `from` parameter is *required* for range check, but `to` is not. `to` alone is silently ignored.

Is there going to be a performance impact when a date range check is added to bbox query? Maybe not because the query already sorts notes by update_date.


mmd-osm:
Use case makes sense, and nicely fits #3578. I don't have enough data here to test the performance impact.

AntonKhorev:
> Now the question is, do we have some use cases which depend on the created at timestamp instead? How would we differentiate both?

This API call sorts only by update dates. If you want the ability to sort by creation dates, you'd need another parameter.

AntonKhorev:
#4222 was merged and does what I wanted to achieve here. But maybe someone still wants `from`/`to` on index?

PR #3606, opened 2022-07-20
Add testing on Ubuntu 22.04

tomhughes:
This is more painful that it ought to be for a number of annoying reasons...

Firstly https://github.com/ruby/setup-ruby currently only supports ruby 3.1 on 22.04 even though 22.04 actually ships with 3.0 as the system ruby - apparently 3.1 is the first version that builds with openssl 3 out of the box so I guess Ubuntu must have patched 3.0 for the system build.

Secondly the forced move to a snap for Firefox means that geckodriver has (currently at least) gone away ([discussion here](https://bugs.launchpad.net/ubuntu/+source/firefox/+bug/1968266)) so we need to switch to Chrome and hope we won't have the same problem there once they force a move to snap for that!

Finally because chrome in headless mode doesn't support browser preferences ([discussion here](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1925)) we have to set the preferred language in the environment before starting it instead.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,8 +10,8 @@ jobs:
     name: Ubuntu ${{ matrix.ubuntu }}, Ruby ${{ matrix.ruby }}
     strategy:
       matrix:
-        ubuntu: [20.04]
-        ruby: [2.7, 3.0]
+        ubuntu: [20.04, 22.04]
+        ruby: [2.7, 3.0, 3.1]
     runs-on: ubuntu-${{ matrix.ubuntu }}
     env:
       RAILS_ENV: test


tomhughes:
The docker run is failing because chromium has in fact been moved to snap only in 22.04 but snaps can't be installed in a docker container so that's that for docker.

gravitystorm:
> The docker run is failing because chromium has in fact been moved to snap only in 22.04 but snaps can't be installed in a docker container so that's that for docker.

For docker, can we just run `bundle exec rails test` i.e. skip the system tests? I'd be happy with that, since running the tests there in CI is more to make sure that the development environment is set up correctly, rather than running all the tests per se.

But in saying that, what's the point in having a development environment where you can't run the system tests? I guess some for some tasks that will be sufficient, but it does rather imply that some developers won't be able to write them or run them. So I guess we need some solution so that devs who want to develop using our docker configuration, can still run all the tests.

> Secondly the forced move to a snap for Firefox means that geckodriver has (currently at least) gone away

Hmm, I wonder if either a) we should wait for geckodriver-and-snap-based solution to appear (although that might still have docker implications) or b) should we switch to using a deb package instead of snap? The linked [askubuntu](https://askubuntu.com/questions/1399383/how-to-install-firefox-as-a-traditional-deb-package-without-snap-in-ubuntu-22) thread highlights the [mozillateam PPA](https://launchpad.net/%7Emozillateam/+archive/ubuntu/ppa) but I don't know what the implications are of doing that. And I guess even then we'll have the problem of most developers (like me) will have a local system on which firefox is installed as a snap.

Oh what a headache!

tomhughes:
I did try a mozilla PPA from one of the bugs but it turned out not to have builds for jammy but I see that one does.

tomhughes:
I spent considerable time on this and I'm coming to the conclusion that sadly you might be right about need to disable system tests in docker.

The first I tried was the firefox repo you mentioned, and that actually works in docker but I can't get it to work for the main tests - trying to add the repo bombs out with an error about held packages but I can't figure out what they are. Presumably something to do with how github build their base images and install firefox in them.

I then tried switching the docker container to use Debian instead of Ubuntu as Debian has a non-snap version of chrome but that still doesn't get on with being run inside a docker container. I'm not sure exactly what is going on but I think there are fundamental conflicts with chromium's sandboxing trying to run inside the docker sandbox, even with `--no-sandbox` passed to chrome.

tomhughes:
It's been reported on https://bugs.launchpad.net/ubuntu/+source/firefox/+bug/1968266 that geckdriver has been fixed and my tests seem to confirm that it is now working, and that ruby-setup now supports 2.7 and 3.0 on 22.04 which fixes all the problems here.

PR #3607, opened 2022-07-20
Consider the first author of all note comments to be the note author

AntonKhorev:
<s>This is one of the prerequisites for not crashing when showing a note with all comments hidden.</s> - not anymore for #3608, I removed access to `note.author` there

### DIFF
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -82,12 +82,12 @@ class Note < ApplicationRecord
 
   # Return the author object, derived from the first comment
   def author
-    comments.first.author
+    all_comments.first.author
   end
 
   # Return the author IP address, derived from the first comment
   def author_ip
-    comments.first.author_ip
+    all_comments.first.author_ip
   end
 
   private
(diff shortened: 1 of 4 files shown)


tomhughes:
The problem with that is that it can be a deleted user and there are open questions around how we should handle those because currently we are inconsistent about the extent to which we acknowledge their existence.

tomhughes:
Is the first one OK though? We absolutely can't send email notifications to a deleted user.

AntonKhorev:
* Email notification only compares the recipient with the note author to decide on the subject later:
https://github.com/openstreetmap/openstreetmap-website/blob/e2af9941a8b00da813eb3e143056b9a9bed25011/app/mailers/user_mailer.rb#L129
There is a check if the recipient is visible:
https://github.com/openstreetmap/openstreetmap-website/blob/3871d61be43857436d1eaefe9e84841e8a7d0f4f/app/controllers/api/notes_controller.rb#L384
* In note rss I removed the `dc:creator` element if the note author is a deleted user.

AntonKhorev:
No longer necessary after #5609 and preceding PRs.

Issue #3396, opened 2021-12-14
When creator of a note is deleted, the first commentator looks like the creator


Strubbl:
### URL

https://www.openstreetmap.org/note/1598467

### How to reproduce the issue?

1. create a note with a user
2. comment the same note with a different user
3. delete the user who has created the not
4. now i looks like the user from 2. has created the note

### Screenshot(s) or anything else?

I the german osm forum this discussion came up: https://forum.openstreetmap.org/viewtopic.php?id=74374

It is confusing, that the first commentator and his/her comment looks like it was the original note text when the person who has created the note, is deleted in osm.

AntonKhorev:
Fixed in #5609

Issue #3450, opened 2022-02-09
Notes throw errors if there are no visible comments


gravitystorm:
It's not possible to view a note if there are no visible comments. This can come about either:

* There's only one comment (the original) and the user is deleted (see #3398)
* There's only one comment (the original) and that comment is hidden (see #3438)

In both cases, you get errors, albeit different ones in each case. But the point remains that there's a lot of (untested) logic that assumes there's at least one note comment visible.

Related:
* #3396 
* https://github.com/gravitystorm/openstreetmap-website/issues/6

mmd-osm:
I thought I tested a scenario where all notes were hidden, and I don't remember seeing an error message for non-moderators. Hmm.

mmd-osm:
Good point. What would be your suggestion here? Should the new API endpoint block any attempts to delete the first comment?

gravitystorm:
> Good point. What would be your suggestion here?

I don't know. I think relying on the fact that the first comment has the "opened" event, and that there's no other "opened" events, seems fragile - but I'm not entirely sure why I feel that. But maybe that's a workaround until we sort things out properly?

mmd-osm:
Hmm. An "opened" event is used exclusively when creating a new note. https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/api/notes_controller.rb#L76

I think what needs to be clarified first is whether we want the opening event to be hidden at all, and other parts of the code need to be adjusted to handle this case properly.

gravitystorm:
I've moved the discussion of the "opened" comment to the PR in #3438 , since this issue is to deal with the wider problems of the current implementation.

AntonKhorev:
Fixed in #5609

PR #3608, opened 2022-07-20
Show notes with no visible comments; show only first comment as note description

AntonKhorev:
* Note description is not shown if there are no visible comments or if the first visible comment is not the first actual comment. In the latter case the first visible comment is shown below "Created by", "Location" etc along with regular comments.
* Since reading `note.author` may cause a crash, don't read it, use `note.all_comments.first.author` instead. `note.all_comments.last.author` was already in use. If this user is deleted, the comment will be shown as "by deleted".

This should allow not to crash on notes without comments, fixing #2146 (and #2176 should not be required).

That's not a complete fix for notes without comments because some client-side js expects comments to be present.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -43,6 +43,7 @@ class NotesController < ApplicationController
       @note = Note.visible.find(params[:id])
       @note_comments = @note.comments
     end
+    @note_description_comments_count = !@note_comments.empty? && @note_comments[0] == @note.all_comments[0] ? 1 : 0
   rescue ActiveRecord::RecordNotFound
     render :template => "browse/not_found", :status => :not_found
   end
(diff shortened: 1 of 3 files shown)


mmd-osm:
Did you already check how this fits with #3441 ?

AntonKhorev:
Also should fix #3396.

AntonKhorev:
> Did you already check how this fits with #3441? I'd be happy if we could avoid some (larger) rework here, and see how different pieces could fit together nicely.

Seems to work ok with one line line of code to be changed manually, the one that decides how many comments to drop to get regular comments:

    <% @note_comments.drop(1).each do |comment| %>

to

    <% @note_comments.drop(@note_description_comments_count).each do |comment| %>

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/issues/2146 fixed in this PR has another unclosed duplicate https://github.com/openstreetmap/openstreetmap-website/issues/3450 (as mentioned [here](https://community.openstreetmap.org/t/error-trying-to-access-note/2199))

AntonKhorev:
No longer necessary after https://github.com/openstreetmap/openstreetmap-website/pull/5609 and preceding PRs.

PR #3615, opened 2022-07-22
Uncomment asserts for namespaced elements in note rss test

AntonKhorev:
They were commented out probably because they weren't working. They weren't working because `:` had to be replaced by `|` in element name selectors.


tomhughes:
Is that some standard CSS thing or some special assert_select hack?

tomhughes:
Ah yes it's a CSS thing.

HolgerJeromin:
You always learn something new. Thanks

https://developer.mozilla.org/en-US/docs/Web/CSS/@namespace

gravitystorm:
@AntonKhorev thanks for this, that's new to me too!

PR #3617, opened 2022-07-23
Don't add title to note marker if note has no visible comments

AntonKhorev:
Fixes javascript errors when the note layer is enabled and there's a note without comments.


HolgerJeromin (approved):

gravitystorm:
Why are there even notes available with no visible comments? What does that even mean for a note, conceptually?

I would have thought if the special first comment was hidden, the concept would be for the entire Note to be hidden. When you hide a diary entry, for example, we don't show a blank entry with the list of comments remaining.

maro-21:
Can you merge it?

nenad-vujicic (approved):
This works great on my side. Thanks!

AntonKhorev:
Shouldn't be necessary after #5576.

Issue #1115, opened 2015-12-21
Language switcher on the homepage


planemad:
Got a query from a teacher in India on how to get his students to sign up on localized version of the OSM website.

Currently this does not seem possible and everyone is forced to use the English version till they are logged in and can access the language settings.

How about integrating the [jquery universal language selector](https://github.com/wikimedia/jquery.uls) into the website. This will allow easy switching between any language in the world in a more usable way than having to know your language ISO code.

The uls is available on all Wikimedia sites and has been well optimized for performance.

tomhughes:
We have long wanted to include a selector, the problem is not the technology to do so, it's finding screen real estate for the selector.

In any case, we use the browser language preferences when users aren't signed in, so I'm not sure why you are seeing such a problem unless you are dealing with poorly configured browsers.

tomhughes:
That jquery thing certainly looks much more heavyweight than anything I was imagining.

planemad:
Some of the features in the ULS:
- Supports all Mediawiki supported languages
- Languages grouped according to region and script type
- Multilingual search
- Intelligent list based on user location
- Extensively tested for usability with users who dont know English

To my knowledge its the most usable language selection list anyone has made till date. It could certainly use some trimming though, but the core works well.

planemad:
> so I'm not sure why you are seeing such a problem unless you are dealing with poorly configured browsers

Yes it is a case of poorly configured browsers. Unfortunately if the only language you know well is an Indic language and are a new computer user, its not very easy to navigate to these settings.

tomhughes:
I did look at it, but it's quite hard even to figure out which bits of it are source and which bits are the files we would actually need to deploy. I think it's just the src, css and images directories we would need, but it's by no means clear.

I also found it quite hard to use myself, because many of the languages are only shown in their native scripts, which is fine for a native speaker but not very helpful for somebody like me trying to test it ;-)

In any case it's a major project to add it. Aside from working our the exactly how to import it into our vendor tree there needs to be somebody to design a control to trigger it, and space needs to be found for that control in our layout. We would also need to add support for a cookie based language override for non-logged in users and something to use that in preference to the browser derived language as the default when an account is created.

planemad:
Its great that its on the radar atleast. Such features makes a huge difference when talking with local government officials.

tomhughes:
I've put up a basic test site at http://locale.apis.dev.openstreetmap.org/ jquery.uls integrated. Issues include:

[...]

Issue #3547, opened 2022-05-08
Ability to change the language of the map


Ahmadhosseinbor:
### Description

I want to see the map in Persian and the names of countries, cities, POIs and streets should also be displayed in Persian, those that do not have a Persian name (tag name: fa) should be displayed in their local language

HolgerJeromin:
This is the wrong repo here.
With the many languages we need vector tiles for that.
See https://github.com/gravitystorm/openstreetmap-carto/issues/3201 for more information

Issue #2287, opened 2019-06-29
Simpler login page for people using login via Google, especially on mobile


matkoniecz:
Someone deciding to log in via Google account is assuming (at least among people that I tested) to skip most of registration form and use Gmail email as email, to not fill password.

Even selecting displayed user name may be surprising.

Would it be possible to offer simpler registration form for someone deciding to use a third party provider? For example there is no need to ask about email for someone registering via Google account, there is no need for them to create password (the may still reset password to setup one if it would be required for some purpose).

Discovered to be a major barrier during UX testing of one of OSM editors. My sample size is extremely small but people were extremely surprised to see password, email, username form after deciding to authenticate with Google.

note that on mobile unneeded forms take valuable screen space - see https://user-images.githubusercontent.com/899988/60381710-6e9c1380-9a59-11e9-98e5-d5a7ee4005dc.png

tomhughes:
We do prefill things based on the OAuth, including the email for trusted providers like gmail, and setting a password is optional in that case.

deevroman:
It seems that the results of the study were ignored https://github.com/streetcomplete/StreetComplete/issues/1470
I especially recommend paying attention to the comment https://github.com/streetcomplete/StreetComplete/issues/1470#issuecomment-526317818
I can also confirm from the experience of my friends that the current behavior is annoying. https://www.openstreetmap.org/user/TrickyFoxy/diary/399621

And I did not understand what the problem is to ask the user only the nickname and not show a huge form that the user does not expect.

matkoniecz:
> It seems that the results of the study were ignored https://github.com/streetcomplete/StreetComplete/issues/1470

Given that it was small scale, was not requested by developers here and was not very rigorous, this is not too surprising by itself.

tomhughes:
I'm not sure why we would even be expected to be aware of it given as far as I know it was never bought to our attention, or why UX testing of an unrelated app would be relevant to us.

deevroman:
Translated it inaccurately. Let's mark the conversation about this as an offtop and get back to the problem.

deevroman:
Stop-stop:) Return my first message. I corrected the translation error about the study. This did not cancel the question.

tomhughes:
Make up your damned mind. This is a closed bug anyway and the explanation given when it was closed still applies.

deevroman:
> We do prefill things based on the OAuth

[...]

PR #3618, opened 2022-07-24
Add a locale selector

tomhughes:
In light of https://www.openstreetmap.org/user/TrickyFoxy/diary/399621 I have resurrected my locale selector branch from many years ago that attempted to address #1115.

The rebased and updated version is running at https://locale.apis.dev.openstreetmap.org/ as it has been for the last N years...

Outstanding issues that we need to think about before we can consider merging this:

* Where exactly to put the trigger control and how to present it - currently I have it showing the language code to save space while Wikimedia normally show the full language name.
* What should take priority for logged in users? or maybe we just hide the control for logged in users?
* Do we need a way to clear the override (other that restarting browser, which loses the session cookie) or not? Possibly if we ignore the cookie and hide the control for logged in users then we don't?


deevroman:
There is another possible problem.
If you show this button on the main page, where there is a map, then the user can expect that the labels on the map will also change.

tomhughes:
Well they don't change for any of the existing ways of changing the locale so while that might be an issue it is not related to this PR and is off topic here.

deevroman:
Issue like this make me think so https://github.com/openstreetmap/openstreetmap-website/issues/3547

tomhughes:
Yes we are perfectly well aware that's a requested feature, but not one that is something we as authors of this repository can do anything about as we do not provide the map tiles. Hence it is out of scope of this discussion.

gravitystorm:
I'm not super keen on this, since browsers are in charge of dealing with language requests. There should (in theory, I know I know I know) be no need for language selectors. And so I find it surprising and a bit disappointing that in the linked diary post, the OP complains loudly about seeing a page in English when they have their browser set up to request pages in English. :shrug: 

However, if we choose to add one, then
* It should only appear for logged out users. We already have a system for customising language preferences for logged in users.
* I don't think there's a need for logged out users to be able to reset the override, they can just choose a different language
* We should reconsider how the logged-in language preferences work, since we'll have a nicer UI for logged out users (a dropdown) than for logged in users (a somewhat cryptic text field).

I also think we should stopped filling in the language preferences automatically. I don't agree with the logic of grabbing the browser preferences supplied when they first create an account, and using that to override their browser prefs from that point on in perpetuity. If I switch my browser to request FR then I expect the site to respect that by default, unless I've specifically chosen differently. So I think:
* We should leave the language preference blank by default, and only set it when the user does so explicitly
* For logged in users, we should use: prefs (if set) -> browser (if set) -> en 
* For logged out users we should use: override (if set) -> browser (if set) -> en

deevroman:
> Yes we are perfectly well aware that's a requested feature, but not one that is something we as authors of this repository can do anything about as we do not provide the map tiles.

You understand, I understand. But the user, as in issue, will most likely expect something else.

---

How about using the differences of the user's country (for example, by GeoIP) and browser headers for the trigger of showing language switching?

---

> when they have their browser set up to request pages in English. 🤷

From post:  _At first, for some reason, he is greeted by a page in English. <b>Well, maybe the problem is in the browser.</b> But why can’t I switch the language on the page?_

> the OP complains loudly

You say it like it was the whole post.

The lack of language switching was just a small thing that complicated the registration. There is quite a lot of text in the registration form and it is unpleasant to read it in a foreign language https://github.com/openstreetmap/openstreetmap-website/issues/2287#issuecomment-1193315720 (But there is also a Help section and everything will be in a foreign language there too...)

And if you carefully re-read the post, you can understand that the situation arose in the built-in StreetComplete browser, which does not have any settings.

---

tomhughes:
Superseded by #5201.

PR #3620, opened 2022-07-25
Arrange base layers menu in a grid

AntonKhorev:
Make layer selection buttons take less space on smaller screens, otherwise they are too stretched horizontally. And we may need more space for other things like note/data/trace checkboxes that are too small on mobile devices.

Screenshot_20220725-184122


mmd-osm:
I'm wondering why this change didn't go into small.scss, if it's mainly targeting small screens/mobile devices.

gravitystorm:
> I'm wondering why this change didn't go into small.scss, if it's mainly targeting small screens/mobile devices.

I'd like to get rid of small.scss entirely at some point.

The way that bootstrap works (and one of the aspects of "mobile-first" design in general) is that you make everything work on small screens, and then you do extra things on larger screens where necessary. This is why all the [bootstrap utils](https://getbootstrap.com/docs/5.1/utilities/spacing/) work on a "size-or-greater" approach. So you have e.g. `mx-3 mx-md-6` meaning that a horizontal margin is considered as 3 by default, and then 6 on "medium-or-larger" screens. It gets mentally confusing if you consider `mx-3 mx-md-6` meaning a margin of six by default, with a margin of 3 on smaller screens.

So our approach should be to have everything currently in small.css considered as the default, and then a "large.scss" for any larger-screen overrides. But these are few so it probably doesn't warrant a separate file, so my plan is to combine the remains of small.scss into common.scss . And many of the screen-size dependent changes are no longer in our custom scss files anyway, and the remaining custom css is being reduced too.

gravitystorm (changes requested):
Thanks for the design idea, which I'm happy with on initial review.

However the implementation is based on additional custom css rules, including a custom grid and specific pixel widths for gaps etc. I can't support that approach. 

We're using Bootstrap for all of our layouts now, and this PR should do the same. See e.g. [Bootstrap Grids](https://getbootstrap.com/docs/5.1/layout/grid/)

AntonKhorev:
> I'm wondering why this change didn't go into small.scss, if it's mainly targeting small screens/mobile devices.

Because there's nothing that is screen size-specific in this code.

> However the implementation is based on additional custom css rules, 

> including a custom grid

See below for reasons.

> and specific pixel widths for gaps etc. I can't support that approach.

There are two pixel values:

- gap width that you already have in `margin-bottom: 8px;` which I replaced with `gap`
- minimal width for layer buttons. I tried different widths and picked the one that I think works better for various screen sizes.

> We're using Bootstrap for all of our layouts now, and this PR should do the same. 

If you want more complicated code where you alter the html to declare rows and use screen-size based breakpoints, you'll do that. You'll have to declare a container, then rows, then put probably two columns in each row for buttons. Then you'll have to make sure that rows are disabled on small screens because they are narrow, and on large screens because the layers menu becomes narrow again ([with a *specific pixel value* of 250 for width](https://github.com/openstreetmap/openstreetmap-website/blob/2cbf6062fc1d9a1f169c36f98d5ff383575cef0a/app/assets/stylesheets/common.scss#L449)). Note that number of declared columns is kind of arbitrary because what you actually want is for buttons not to be too small. So the number of columns per row is going to be an indirect way to set the minimum button width. And screen-size breakpoints are going to be an indirect way to have a condition on the layers menu width.

> See e.g. [Bootstrap Grids](https://getbootstrap.com/docs/5.1/layout/grid/)

And you'll see there that Bootstrap grids are not [css grids](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout), they are based on [flexbox](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout). I use css grid here in order *not do declare rows and columns manually* because there's no intrinsic meaning behind putting this button in this row&column and that button in that row&column. That's what "custom grid" is for in this case. If you actually going to argue that let's say *Cycle map* layer button should be in the right column for some reason, you'll create that column manually and put the button there. But I see no such reason.

gravitystorm:
I'm not going to argue the merits of one grid system against another - I'm not an expert in either! I'll note that bootstrap 5 has experimental support for CSS grids, that's disabled by default. Maybe that will become default in the future or we'll switch to it or something, I have no idea.

What I will say though is that we've spent a lot of time and effort getting rid of lots of our old custom CSS and using bootstrap instead. This brings unification to our style (e.g. no longer having arbitrary pixels everywhere), opens up the possibilities for easier customisation, and gives us free updates - the effort that goes into bootstrap far exceeds any effort we can put in to maintaining our own custom CSS, and the number of people familiar with bootstrap is far greater than the number of people familiar with our custom CSS.

I'm happy to still use custom CSS where it's necessary, but I really want to avoid it where there's an alternative.

In this case, we already have "grids of things that rearrange themselves based on screen size" elsewhere on the site, so it needs some pretty compelling reason to have this particular "grid of things" implemented entirely differently than the others.

AntonKhorev:
> In this case, we already have "grids of things that rearrange themselves based on screen size" elsewhere on the site, so it needs some pretty compelling reason to have this particular "grid of things" implemented entirely differently than the others.

Is *not being based on screen size but being based on a container size* a compelling reason?

AntonKhorev:
> I'm not going to argue the merits of one grid system against another - I'm not an expert in either! 

I'm not arguing about merits of grid system. I'm arguing about [explicitly specified grids vs implicitly specified ones](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Basic_Concepts_of_Grid_Layout#the_implicit_and_explicit_grid). Let's say you have a commenting UI where you have a textarea and on the left side and a box with kramdown hints on the right side. You want textarea on the left and kramdown hints on the right. Therefore you declare that there's a row with two columns, the left column contains the textarea and the right one contains the hints.

Now you have a set of layer switching buttons. You *don't care which rows and columns they occupy*. Why would you go and write out those rows and columns if you don't care about them?

> I'll note that bootstrap 5 has experimental support for CSS grids, that's disabled by default. Maybe that will become default in the future or we'll switch to it or something, I have no idea.

You mean this thing?
https://getbootstrap.com/docs/5.1/layout/css-grid/

You still have to declare rows and columns manually with it, right? That's still an *explicit* grid. The reason to do the grid differently for layer buttons is because it's an *implicit* grid.

AntonKhorev:
> @AntonKhorev to be honest, it pains me to read the last few messages.

Let's clear this up. Are you telling that I should have used an explicit grid for layer buttons?

> Andy gave a clear guidance on how he thinks the code should be structured.

In case you think that I should have used implicit grid:
A clear guidance would be telling me how to do implicit grids in Bootstrap. I'm not a Bootstrap expert and I don't know how to do it. I searched for the relevant css properties in bootstrap source code and found nothing: https://github.com/twbs/bootstrap/search?q=grid-template , at least in the current version. Maybe there's some plugin?

In case you think that explicit grid is better: why? Is it just about not using any css features that are not defined in Bootstrap? And why is that? For maintenance because the code will get more complicated otherwise? My proposed change here adds three css properties and removes one. You hardly can do this change any simpler than that.

gravitystorm:
> You _don't care which rows and columns they occupy_. Why would you go and write out those rows and columns if you don't care about them?

We've used https://getbootstrap.com/docs/5.1/layout/grid/#row-columns elsewhere on the site, which allows us to avoid being explicit, and also set different numbers of columns for different screen sizes.

> Is _not being based on screen size but being based on a container size_ a compelling reason?

It's a nice reason, but on its own isn't compelling - partly because the container size is based on screen size in this case.

AntonKhorev:
> > You _don't care which rows and columns they occupy_. Why would you go and write out those rows and columns if you don't care about them?
> 
> We've used https://getbootstrap.com/docs/5.1/layout/grid/#row-columns elsewhere on the site, which allows us to avoid being explicit, and also set different numbers of columns for different screen sizes.

But that way you still specify the number (and contents) of columns. But why would you care about the number of columns for layer buttons? Why not saying "however many columns possible as long as they are equal width and wider than this size"? That's what you can specify with one css property.

> > Is _not being based on screen size but being based on a container size_ a compelling reason?
> 
> It's a nice reason, but on its own isn't compelling - partly because the container size is based on screen size in this case.

And then you'll have this weird condition to check where on small screens the container is narrow, on wider screens it's wider and on yet wider screens it's narrow again.

AntonKhorev:
> However the implementation is based on additional custom css rules, including a custom grid and specific pixel widths for gaps etc. I can't support that approach.

I still don't understand if you are against specific widths *in pixels* or widths in any units. Since the limiting factor for layer buttons are labels, it would make sense to specify the minimum width in character-based units, unless you don't want to specify it in any way of course.

gravitystorm:
> I still don't understand if you are against specific widths _in pixels_ or widths in any units.

In general I'm against specific widths appearing in custom.css (in fact, I'd be quite happy to remove most of what remains in common.scss entirely, but that's an ongoing project). 

So for example, `.layers-ui.base-layers li` contains `margin-bottom: 8px` - this is an example of a style rule that should be removed, and replaced with either a) nothing, since bootstrap applies appropriate margin-bottoms to most components anyway or b) a bootstrap utility, e.g. mb-3. This way, margins are kept consistent all around the site, and are easy to change by customising bootstrap centrally. Same with the border-radius - these buttons don't need a custom border radius that's different from every other border radius used on the site; they should be using a [bootstrap corner radius](https://getbootstrap.com/docs/5.1/utilities/borders/#border-radius) utility class.

So the same applies to grids, grid spacing, grid margins etc - no singular grid should have its own widths set, since that means they are inconsistent with any other grid used in the site.

This might seem unimportant, but small things like these often cause unexpected problems. For example, we used to have various `h1` and `h2` elements around the site with their own style rules and custom font sizes, originally in px and more recently in rem. But even those are now causing problems, since these hardcoded style rules interfere with the recently adopted [RFS](https://getbootstrap.com/docs/5.1/getting-started/rfs/) - see #3630. 

I hope that explains things.

AntonKhorev:
> So for example, `.layers-ui.base-layers li` contains `margin-bottom: 8px` - this is an example of a style rule that should be removed, and replaced with either a) nothing, since bootstrap applies appropriate margin-bottoms to most components anyway

But that's not going to happen in our case because:

- layer selection buttons are not Bootstrap components
- we want horizontal gap when two columns are visible - actually there is a default horizontal gap but it's too wide

> or b) a bootstrap utility, e.g. mb-3.

In #3629  I added `g-2` class to rows for 8px gap:
https://github.com/openstreetmap/openstreetmap-website/pull/3629/files#diff-87e33ea7d5e625dd851a78a3192719681de20caea8edd278a10f5342585c98e7R37

> This way, margins are kept consistent all around the site, 

Not in this case because I added a class just in this particular place.

> and are easy to change by customising bootstrap centrally. 

Not in this case unless you are going to redefine `g-2` globally.

> So the same applies to grids, grid spacing, grid margins etc - no singular grid should have its own widths set,

But this grid has its own spacing etc.

> since that means they are inconsistent with any other grid used in the site.

What this grid is supposed to be consistent with if there's no similar grid in the site?

Was adding `g-2` wrong in this case? If it was should I have used the default values instead (larger horizontal gaps and no vertical gaps)? But they look ugly. If it wasn't, why? There's no consistency / centrality in this case and it's still specifying the gap but in a different place in different units.

AntonKhorev:
Looking at other grid code I guess the answer to "Was adding g-2 wrong in this case?" should be yes, because there are margin classes with sizes etc.

AntonKhorev:
I expected merge conflicts here and there wasn't any, but I updated this PR anyway.

> However the implementation is based on additional custom css rules

now there's only one rule. Of course the most likely scenario under which this will get accepted is some new version of Bootstrap adds an utility class that does exactly this thing. Looks like 5.3 still doesn't have it.

> including a custom grid

now it's Bootstrap's `.d-grid` (which is the same as `display: grid`)

> and specific pixel widths for gaps

not it's Bootstrap's `.gap-2` (which is the same as that specific pixel value unless the spacer gets changed; it wasn't me put that specific value in css)

PR #3628, opened 2022-07-26
Update INSTALL.md

victorovento:
When running bundle exec rails server, the site can be only accessed from the local machine, and no other devices in network. It can generate confussion when user runs the dev server in a ubuntu virtual machine. Adding -b 0 at the end should work.


tomhughes:
Only being able to access it locally is entirely deliberate - it should be a conscious decision on the part of the user that they are ready to expose it to the wider world.

PR #3629, opened 2022-07-26
Put layer selection buttons into bootstrap row to have two cols on smaller screens

AntonKhorev:
"Politically correct" version of #3620 where instead of changing css we change javascript. Costs extra wrapper element for each button to utilize Bootstrap's gutters.

There's an assumption that small screens are not too small and are able to handle two buttons per row. Handling even smaller screens would require a custom breakpoint. Actually requiring two buttons per row is kind of backwards because we probably want the buttons themselves not to be too small, but this way we avoid providing widths values.


tomhughes:
When you're asked to make changes to a PR you can just push those changes to he same branch (force push if you're replacing the existing commits rather than adding to them) and it will change the PR rather than having to open a new one.

Also adding snide comments really doesn't help - you have been doing good work and contributing useful changes but you've just set your reputation back several steps in my estimation.

AntonKhorev:
> When you're asked to make changes to a PR you can just push those changes to he same branch (force push if you're replacing the existing commits rather than adding to them) and it will change the PR rather than having to open a new one.

I don't want to push them to that branch. I opened this branch so you could compare the changes here and the changes there and pick the one that's better. And I don't think that this PR is better.

> Also adding snide comments really doesn't help - you have been doing good work and contributing useful changes but you've just set your reputation back several steps in my estimation.

You mean calling this a "politically correct version"? It is a version that is not technically better than the other one. It could be if we somehow get rid of most of css, then there would be one language less in use, but I don't think that's going to happen. Otherwise how is it easier to write presentational code in imperative languages all across the codebase, instead of writing in a declarative language in one place?

Or are you talking about "kind of backwards" about button widths? They have a label over them which is sourced from localized strings. This label might get larger than you expect. Is it still going to fit? You could have a minimal width specified and incerease it in case it doesn't work. But now instead I have to fix a number of columns to satisfy the *no specific widths* requirement. Now what if the width turns out to be insufficient? It's either custom css to make text smaller or custom breakpoints for smaller screen sizes to go back to the single-column layout.

tomhughes:
Sorry I thought "politically correct" bit referred to Andy's request to do it using bootstrap and that you were trying to express an objection to that.

AntonKhorev:
If you really want to avoid depending on some specific minimal safe button width, a possible fix for that is to gradually decrease the label font size. That's more custom css however.

You'd change font size here
https://github.com/openstreetmap/openstreetmap-website/blob/e9887f88edf37ec164bb4373f7bcb86262b781b6/app/assets/stylesheets/common.scss#L509
to something like

    font-size: clamp(12px, calc(12px + 4/500 * (100vw - 268px)), 16px);

which can be made more readable by declaring variables (4 is a difference between font sizes, 500+268 is the breakpoint for layers panel to go full screen width).

Or do a less weird thing and put one custom width media query when the button becomes smaller than ~180px and make the font size smaller. That's about 420px. This is assuming that whoever made translations for layer names tried to make them fit in the original button size. Anyway the label font size now is quite large - 16px - which becomes noticeable once it moves under the OSM logo with font size 18px.

AntonKhorev:
I decreased the font size of layer labels on small screens:

VirtualBox_ubuntu1_28_07_2022_17_51_46

PR #3630, opened 2022-07-27
Various font size changes

gravitystorm:
Now that we're on Bootstrap 5 we are using RFS, which implements responsive font sizes. Therefore it's worth refactoring places where we have font-sizes declared in our common.scss, particularly for headers, because these custom definitions are not responsive.

I've also refactored the about, welcome and fixthemap pages slightly, to change the headers used, use the bootstrap '[lead](https://getbootstrap.com/docs/5.1/content/typography/#lead)' paragraph styling, and move some more CSS into bootstrap utils.

### DIFF
--- a/app/views/site/welcome.html.erb
+++ b/app/views/site/welcome.html.erb
@@ -6,66 +6,66 @@
   <h1><%= t ".title" %></h1>
 <% end %>
 
-<p><%= t ".introduction_html" %></p>
+<p class="lead"><%= t ".introduction_html" %></p>
 
-<h3><%= t ".whats_on_the_map.title" %></h3>
+<h2><%= t ".whats_on_the_map.title" %></h2>
 
 <div class='row'>
   <div class='col'>
-    <div class='center'>
-      <span class='sprite small check'></span>
+    <div>
+      <span class='sprite small check mx-auto'></span>
     </div>
     <p><%= t ".whats_on_the_map.on_html" %></p>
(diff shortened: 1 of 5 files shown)


app/views/site/welcome.html.erb
```diff
@@ -6,66 +6,66 @@
...
 
 <p><%= t ".basic_terms.paragraph_1_html" %></p>
 
-<div class='clearfix icon-list'>
+<div class='clearfix'>
   <div class='clearfix'>
-    <span class='sprite small term editor'></span>
-    <p><%= t ".basic_terms.editor_html" %></p>
+    <span class='sprite small term editor float-start'></span>
+    <p class="align-middle"><%= t ".basic_terms.editor_html" %></p>
```
tomhughes, line 35: This `align-middle` is not working because it's on a block element - it's also missing from the other three entries below.

tomhughes (changes requested):

app/views/site/welcome.html.erb
gravitystorm: oops, I had been doing some experiments, and that wasn't intended for this PR.

gravitystorm: fixed

tomhughes:
I've merged this but I think we'll need to look at the alignment of those items on the welcome page as the old version while not perfect did look better.

PR #3634, opened 2022-07-28
Fix map key tooltip

### LINKED ISSUE #3632: No tooltips for i/note/query buttons
### URL

https://openstreetmap.org

1. Hover over the buttons at the top right of the map.
2. When hovering over the Layers button, there is a 'Layers' tooltip
3. When hovering over the i/note/query buttons, there is no tooltip

image
image

AntonKhorev:
Fixes #3632

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -59,7 +59,7 @@ window.updateLinks = function (loc, zoom, layers, object) {
   var editDisabled = zoom < 13;
   $("#edit_tab")
     .tooltip({ placement: "bottom" })
-    .attr("data-original-title", editDisabled ?
+    .attr("data-bs-original-title", editDisabled ?
       I18n.t("javascripts.site.edit_disabled_tooltip") : "")
     // Disable the button group and also the buttons to avoid
     // inconsistent behaviour when zooming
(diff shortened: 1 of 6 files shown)


tomhughes (commented):
Thanks for the fix but maybe we should fix the other four places that are doing the same thing at the same time? This was the result of searching for data-original-title:

```
File: app/assets/javascripts/application.js
  62  17     .attr("data-original-title", editDisabled ?

File: app/assets/javascripts/leaflet.key.js
  66  21         .attr("data-original-title",

File: app/assets/javascripts/leaflet.layers.js
 174  27           item.attr("data-original-title", disabled ?

File: app/assets/javascripts/leaflet.query.js
  23  21         .attr("data-original-title", I18n.t(isDisabled ?

File: app/assets/javascripts/leaflet.note.js
  20  21         .attr("data-original-title", I18n.t(disabled ?
```

AntonKhorev:
They are not doing the same thing. The one I replaced swaps one tooltip text for another. Others are trying to enable/disable tooltips. If you change `data-original-title` to `data-bs-original-title` open the website with a zoom level where "zoom in to to stuff" tooltips are not shown then zoom out they sometimes fail.

gravitystorm:
What @tomhughes meant was that there were several tooltips broken, and the fix for all the tooltips was the same. So the same fix could be applied (`data-original-title` -> `data-bs-original-title`) in each case.

No matter. I have fixed the others, and added some tests. Thanks @AntonKhorev for this PR and in particular writing some tests for them, which were easy for me to extend to other tooltips and will prevent them from breaking in future.

The original cause of the breakage was me not noticing that they broke during #3610

PR #3642, opened 2022-08-01
Add Link and text for the new community forum

Firefishy:
Add link to the new [community forum](https://community.openstreetmap.org), replacing the link to the original "legacy" forum.

Intentionally replaced the translation keys to ensure replacement in the translated text.

### DIFF
--- a/app/views/site/help.html.erb
+++ b/app/views/site/help.html.erb
@@ -4,7 +4,7 @@
 
 <p class='introduction'><%= t ".introduction" %></p>
 
-<% sites = %w[beginners_guide help mailing_lists forums irc switch2osm welcomemat wiki] %>
+<% sites = %w[beginners_guide help mailing_lists forums community irc switch2osm welcomemat wiki] %>
 <% sites.prepend("welcome") if current_user %>
 
 <div class="row row-cols-sm-3 g-4 mb-3">
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2125,8 +2125,12 @@ en:
         description: Ask a question or discuss interesting matters on a wide range of topical or regional mailing lists.
       forums:
         url: https://forum.openstreetmap.org/
-        title: Forums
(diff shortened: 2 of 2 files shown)


gravitystorm:
A quick check shows that the "legacy" forums has a similar amount of activity (just looking at the number of active threads over the last day or two). I think it might be worth linking to both until the new one is significantly more used - what do you think?

Firefishy:
The plan is to [completely migrate the content of the old forum to the community site](https://github.com/openstreetmap/operations/issues/604) as soon as practical.

The [Forum Governance team would like to encourage the promotion of the new community site](https://community.openstreetmap.org/t/forums-governance-team-meeting-2022-07-14/1971#forums-promotion-3).

tordanik:
Our goal is to turn off the old forum once its contents have been migrated and to redirect existing links to the new site. The two platforms aren't really intended to co-exist for an extended period of time; this situation is mostly a result of delays with the content migration. So I agree with the approach chosen in this pull request: Replace the old link with the new one.

gravitystorm (changes requested):
I've read through various threads on community.osm.org to try to figure out what is going on, and I see forum.osm.org is still active and there's been no migration of any content yet. So there's still a reasonable need for mappers to visit forum.osm.org.

I understand the goals that you have. But I think it's better to add a link to the new site, and also keep the link to the existing forum.osm.org site for now. When the migration has been completed and there are no ongoing reasons to visit the old site (i.e. it's archived or redirected or disabled or whatever) then that would be the right time to remove the link. But while it's in active use, and the only source for the historical discussions, and it's effectively a free choice for a mapper which site to go to, we should link to both.

Firefishy:
I have amended the PR to keep the legacy forum now.

gravitystorm (approved):

gravitystorm:
Thanks for making the changes!

PR #3645, opened 2022-08-03
Fix link after AND rebranded GeoJunxion

harry-wood:
It looks like AND went through a rebrand a couple of years ago. Now known as GeoJunxion

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1974,7 +1974,7 @@ en:
           Direction Générale des Impôts.
         contributors_nl_html: |
           <strong>Netherlands</strong>: Contains &copy; AND data, 2007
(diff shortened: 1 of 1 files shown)


simonpoole:
IIRC attribution of AND was part of an agreement with the company, with other words an unilateral change is a nono, they need to agree with it.

gravitystorm:
@harry-wood Thanks for the PR. We usually take the view that any changes to these attribution strings have to be prior-approved by LWG, since there are often issues (like @simonpoole has just mentioned) that we, as maintainers, aren't best placed to resolve.

I'm going to close this PR for now, but feel free to discuss and agree any changes with LWG - and then we can merge the agreed PR.

PR #3657, opened 2022-08-17
Correct name in graphhopper.js

Mortein:
When using directions, the footer says "Directions courtesy of Graphhopper", it should instead say _GraphHopper_.


gravitystorm:
Thanks for the PR @Mortein !

PR #3660, opened 2022-08-24
Upgrade to bootstrap 5.2

gravitystorm:
The upgrade exposed a problem with the way we were calling the mixins, so use named arguments instead. This is probably a  better approach in any case.

Also remove a background declaration that is not required.


gravitystorm:
There's something weird going on with sass in CI, so I've converted this to a draft while I go spelunking on that.

gravitystorm:
> There's something weird going on with sass in CI, so I've converted this to a draft while I go spelunking on that.

https://github.com/sass/sassc-rails/issues/174 is the only mention I've found so far, doesn't hint at any workarounds.

Firefishy:
https://github.com/sass/sassc-ruby/pull/233 worked for me, I cannot comment if it is a pretty solution.

Firefishy:
There are also relevant comments here: https://github.com/sass/sassc-ruby/issues/220

gravitystorm:
@Firefishy great, thanks for the links.

Since this isn't urgent, and figuring out "what to do with sass" is a much bigger and more complex situation than I expected, I'm going to close this PR for now. I'll revisit the bootstrap 5.2 upgrade in a few weeks/months and see what the wider state of play is by that point.

PR #3666, opened 2022-08-27
Position close buttons with flexbox

AntonKhorev:
Implemented this:

> https://github.com/openstreetmap/openstreetmap-website/pull/3631#issuecomment-1203797930
> Many of the close buttons https://github.com/openstreetmap/openstreetmap-website/pull/2901 recently to use flexbox for layout, rather than floating, but there's still some (as found in this PR) where the icons are floating around. These could be refactored to follow the convention used in the sidebar.

Also refactored common code in Leaflet sidebars that wrote close buttons among other things.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -1,5 +1,6 @@
 //= require_self
 //= require leaflet.sidebar
+//= require leaflet.sidebar-pane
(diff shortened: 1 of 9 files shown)


gravitystorm:
Looks good to me! Thank you.

PR #3672, opened 2022-08-31
Remove geolink and close button nesting

AntonKhorev:
Fixes buttons within geolinks issue: https://github.com/openstreetmap/openstreetmap-website/pull/3631#issuecomment-1203797930

For the directions sidebar I removed the geolink because its close button had its own click handler that prevented default actions.


gravitystorm:
This is great, thank you!

PR #3674, opened 2022-09-01
Use Bootstrap toggle buttons for layers menu

AntonKhorev:
Enables keyboard controls for base layers section of layers sidebar.

What I didn't bother doing is making colors exactly the same as before. There was a green border on active button but every other control in the sidebar had blue active color which is also the default for bootstrap's primary buttons. So I kept the default color.

Updated version:
image

The update was done to eliminate pixel values from css at the expense of having wider borders.

Version using only the first two commits:
VirtualBox_ubuntu1_01_09_2022_18_16_56

### DIFF
--- a/app/assets/javascripts/leaflet.layers.js
+++ b/app/assets/javascripts/leaflet.layers.js
@@ -15,13 +15,13 @@ L.OSM.layers = function (options) {
         .prop("checked", map.hasLayer(layer))
         .appendTo(baseSection);
 
-      var mapContainer = $("<div>");
+      var mapContainer = $("<div class='position-absolute'>");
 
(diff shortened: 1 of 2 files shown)


mmd-osm:
I currently don't see a difference in the text color for the map layers. On your screenshot "ÖPNVKarte" and "Humanitarian" are shown in blue, while "Standard" and "CyclOSM" are in black. Is this somehow related this change?

grafik

tomhughes:
The blue is actually the standard bootstrap colour for a "primary" button and the black is then an override for the selected button, again from the bootstrap styles.

AntonKhorev:
On my screenshot one button is currently enabled layer, one is focused button, and two are neither enabled not focused.

HolgerJeromin:
can be closed?

Issue #1020, opened 2015-07-28
Let history viewer respect middle-click and Ctrl+click


ghost:
Please let openstreetmap.org/user/.../history respect middle-clicks and Ctrl+clicks: open them in a new window.

On the links itself in the list it works since they are just your old HTML `<a>`'s, but the rest of the list item is a JavaScript link. Also clicking the bounding box in the map should respect this.

This is especially important since the history list does not remember state, i.e. each time I go back in the browsing history, only the 20 first changesets are shown.

tomhughes:
I believe it already does, to the maximum extent possible - browsers do not allow all of these click types to be intercepted.

tomhughes:
See 528d4a040e593f6ce2171167c436dbdf27511711 and 77b4ff6ceb81a326b7ed3c80cf5d65ec2a2b18d5 for previous work on this.

M1dgard:
It really doesn't work in Firefox. Seems like jQuery-simulate isn't doing its job properly then.

IMHO, it's a bit overkill as well when you could just do something like

```js
function clickChangeset(id, e) {
  var url = "/changeset/"+id;
  if (e.button===1 || e.ctrlKey) {
    window.open(url);
  } else {
    window.location = url;
  }
}
```

tomhughes:
No, the problem is that it isn't possible. Middle click events are simply never delivered by the browser, so the click handler never even runs, and ctrl-click can't simulated in Firefox at least - see https://bugzilla.mozilla.org/show_bug.cgi?id=812202 for details.

tomhughes:
Oh and your solution is not acceptable as it assumes what middle mouse and/or ctrl will do - that is up to the browser UI designer and/or the user configuring the browser and we have no way of knowing what they have chosen. That is why we simulate the event instead so that the correct browser action can be triggered.

M1dgard:
I see, but now nothing works in Firefox ... I don't know about other browsers. On my phone it doesn't work either.

Actually my first thought was to make the `<a>` a big block element with all the stuff in it (and make all that stuff in it inline objects like `<span>` since `<a>` can't have block elements inside of it, and restore their block with CSS) but then you lose semantics. And it doesn't solve the map view.

M1dgard:
You could also provide an opt-in for the 'bad' behaviour in the user settings.

PR #3675, opened 2022-09-01
Stretch changeset links in history lists

AntonKhorev:
Use Bootstrap's stretched link instead of simulating clicking a normal link on [history](https://www.openstreetmap.org/history) pages.

Fixes https://github.com/openstreetmap/openstreetmap-website/issues/1020


tomhughes:
Great solution - thanks!

Issue #3663, opened 2022-08-24
Resolved note text


boothym:
The text at the bottom of a resolved note says (bold added by me):

> If this note contains sensitive information that needs to be removed, you can [Report this note](https://www.openstreetmap.org/reports/new?reportable_id=3226636&reportable_type=Note). For all other problems with the note, please **resolve** it yourself with a comment.

Since the note is already resolved, the bold text does not make sense - it should say something like "please reactivate the note and leave a comment", or be removed.

gravitystorm:
cc @harry-wood who created #3581 and might have some thoughts on how to handle this.

boothym:
Remembered I'd also opened a note about asking users to log in to comment on a note - see #2761 which would presumably change the same file for this.

AntonKhorev:
> "please reactivate the note and leave a comment"

"please reactivate the note" -> user reactivates the note -> "please resolve it" -> user resolves it -> "please reactivate the note" -> ...

PR #3676, opened 2022-09-02
Display how long until a note will disappear

harry-wood:
Modify the message displayed at the bottom of the notes sidepanel. For already resolved notes we don't want to say "please resolve it". Fixes https://github.com/openstreetmap/openstreetmap-website/issues/3663  .

Instead explain that it will disappear from the map (so reporting is probably not necessary) and display how long to go until that happens. Tackling https://github.com/openstreetmap/openstreetmap-website/issues/3071

_"For all other problems, resolving is sufficient. This resolved note will disappear from the map in X days"_

### DIFF
--- a/app/helpers/note_helper.rb
+++ b/app/helpers/note_helper.rb
@@ -23,4 +23,9 @@ module NoteHelper
       link_to h(author.display_name), link_options.merge(:controller => "/users", :action => "show", :display_name => author.display_name)
     end
   end
+
+  def disappear_in(note)
+    date = note.freshly_closed_until
+    tag.span(distance_of_time_in_words(date, Time.now.utc), :title => l(date, :format => :friendly))
+  end
 end
--- a/app/views/browse/note.html.erb
+++ b/app/views/browse/note.html.erb
@@ -71,6 +71,21 @@
   <% end %>
 
   <% if current_user && current_user != @note.author %>
-    <p><small class="text-muted"><%= t "javascripts.notes.show.report_link_html", :link => report_link(t(".report"), @note) %></small></p>
+    <p>
+      <small class="text-muted">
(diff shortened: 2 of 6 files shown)


app/helpers/note_helper.rb
```diff
@@ -23,4 +23,9 @@ def note_author(author, link_options = {})
       link_to h(author.display_name), link_options.merge(:controller => "/users", :action => "show", :display_name => author.display_name)
     end
   end
+
+  def disappear_date(closed_date)
+    date = closed_date + 7.days
```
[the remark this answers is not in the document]
harry-wood: I've made this a CONSTANT now, as well as rewriting it a bit

app/views/browse/note.html.erb
```diff
@@ -71,6 +71,21 @@
...
+        <% elsif @note.status == "closed" %>
+          <%= t "javascripts.notes.show.other_problems_resolved" %>
+        <% end %>
+      </small>
+    </p>
+  <% end %>
+
+  <% if @note.freshly_closed? %>
+    <small class="text-muted">
+      <%= t("javascripts.notes.show.disappear_date", :disappear_in => disappear_in(@note)).html_safe %>
```
gravitystorm, line 88: Avoid using `html_safe`, since that opens up potential for security issues (it just marks the string as html safe, it doesn't do any sanitisation). 

Instead, use a translation key that ends in `_html`. For similar examples, see https://github.com/openstreetmap/openstreetmap-website/pull/2898

gravitystorm (changes requested):

app/views/browse/note.html.erb
harry-wood: Ah yes. I had a feeling there might be a way to avoid html_safe. That's better. So it's called `disappear_date_html` in en.yml now.

gravitystorm:
Thanks!

PR #3678, opened 2022-09-02
Remove top margins from close buttons

AntonKhorev:
The margins mentioned in https://github.com/openstreetmap/openstreetmap-website/pull/3631#issuecomment-1203797930 are likely no longer necessary. They were leftovers from floated buttons put inside headers, probably to align the buttons with the text baseline. And they only affect the button position in the left sidebar.

| With margin-top | Without margin-top |
| --- | --- |
| b0 | b1


gravitystorm:
Thanks!

PR #3679, opened 2022-09-02
Link to user profile from user dropdown

AntonKhorev:
When logged in, there's a dropdown menu in the upper left corner with avatar/username. When clicked, the dropdown menu with *My Dashboard*, *My Messages* etc opens. It's also a link, so it can be middle-clicked and opened in a new tab. But the link is dummy, with href="#". Since it has a username maybe it should link to the user's profile?

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -80,7 +80,7 @@
     </ul>
     <% if current_user && current_user.id %>
       <div class='d-inline-flex dropdown user-menu logged-in clearfix'>
-        <a class='dropdown-toggle btn btn-outline-secondary px-2 py-1 flex-grow-1' data-bs-toggle='dropdown' href="#">
+        <a class='dropdown-toggle btn btn-outline-secondary px-2 py-1 flex-grow-1' data-bs-toggle='dropdown' href="<%= user_path(current_user) %>">
           <%= user_thumbnail_tiny(current_user, :width => 25, :height => 25, :class => "user_thumbnail_tiny rounded-1") %>
           <%= render :partial => "layouts/inbox" %>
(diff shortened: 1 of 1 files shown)


gravitystorm:
I think it would be better to refactor the button to be a <button> instead. It's not clear that it needs to be a link at all, it might only be so for historical reasons?

If we keep it as a link, we would then need to debate what page it should link to (profile or dashboard?) and it doesn't seem super-intuitive that left click makes it work like a button and middle click makes it work like a link anyway.

AntonKhorev:
> It's not clear that it needs to be a link

When it's a link, it can make a user page to be one (middle) click away instead of two (open dropdown - click menu link).

> (profile or dashboard?)

That depends on what people use more often. Dashboard is only useful if you have set your coordinates or added friends. And it doesn't have *My edits*, *My notes* etc links.

> it doesn't seem super-intuitive that left click makes it work like a button and middle click makes it work like a link anyway

Anyone who doesn't figure out that it's a link will just have to do the usual two clicks.

mmd-osm:
I think this feature has a rather poor discoverability. I would have never occured to me, that I could use a middle-click on a dropdown list.

AntonKhorev:
You can see that it's a link because the pointer changes over it and you know you can open links in a new tab with a middle click.

gravitystorm:
I've considered this further, and I'm still of the opinion that this isn't a behaviour that we should add. 

Instead, I've refactored the dropdown buttons to be html buttons, in #3857, which removes any suggestion that they should link to somewhere. I think this is also more in line with their primary purpose (which is to be a dropdown button) without causing any confusion for screen readers etc.

Thank you for the suggestion nonetheless.

Issue #1089, opened 2015-11-18
/user/<username>/history needs browser history memory


mvl22:
The /user/<username>/history page is painful when trying to find a changeset made a while back.

Ideally it would have an auto-scroll (like Facebook) when you reach the end of the current pagination, rather than requiring clicking a button.

But the main issue is that clicking on an entry, then clicking back, loses the current pagination position, so requires lots of re-clicking to get back to the same position.

polarbearing:
Auto-scroll is not ideal either, trying to find and refer to a sequence of changeset in midst of thousands is impossible. We should go back to classic pagination, and start with page one on the oldest changesets, as it is typical in forums.

gileri:
@polarbearing I think that the latest changesets of a user are the most relevant when looking at a user's history. Older changeset have more chance to have been overwritten by the same contributor or another user.

polarbearing:
How you number the pages is independent from the relevance. You can start showing the page with the highest number first, of course. Just the page number remains predictable in my proposal.

gileri:
You're right, nothing forbid to display the last (most recent) page by default, sorry for the noise.

polarbearing:
> @tomhughes said in #933: it wouldn't get merged, because it requires every view to count all the matching changesets.

where is the difficulty? The user page ```www.openstreetmap.org/user/<name>``` gives the number of edits already. On a user history page, the most recent page number, in numbered pagination, is =round_up(edits/page_length). 

With the current endless paging, it is practically impossible to find older pages of a power user, since I cannot make a manual binary search in his pages. I can only press more and more and more until my browser explodes.

There was numbered pagination a while ago, why has it been given up?

tomhughes:
That value on the user page is actually a special cache, not the real count.

Numbered pagination needs to go pretty much everywhere, precisely because it requires all the records to be counted to determine how many pages there are and then when page N is visited you have to read through all the records in order to find the right ones. That's why it gets noticeably slower the further down the list you go, and it's far worse when a search engine decides to visit all the pages...

The correct solution is to do pagination based on an indexed field (either the changeset ID or the timestamp in this case) so that viewing the hundredth page can be done by an indexed lookup. That is in fact exactly how this particular already works I believe, it just doesn't put that marker value in the URL at the moment.

Issue #647, opened 2013-12-09
Viewing a changeset loses context of previous changeset list


SomeoneElseOSM:
1) View changesets for a user.
2) Hit "load more" a couple of times
3) imagine that eventually you see one of interest.
4) Click on the changeset to display it.
5) Press back
6) You're now at (1) above rather than (3) - you have to go through the "load more" procedure in order to get back to where you were.

tomhughes:
It seems that "fixing" this involves an awful lot of complicated and fragile javascript yet we've never really discussed if this is a problem that's worth solving - my natural inclination if I was the person described in the above scenario would be to open the changesets I was interested in in new tabs rather than clicking backward and forward in the same tab.

AntonKhorev:
> yet we've never really discussed if this is a problem that's worth solving

Yes we did. See your email on 2021-06-09.

gravitystorm:
> It seems that "fixing" this involves an awful lot of complicated and fragile javascript

That's my feeling too. I wonder if there's a better approach that relies on less javascript, rather than more.

If the basic behaviour was that pagination worked properly without javascript, then we could use a small amount of javascript to enhance this (e.g. viewing more items without a full page reload). But what we have is javascript-only psuedo-pagination, and then we're looking at adding even more javascript to make it behave more like regular page loads. This seems the wrong strategy.

I also think adding more javascript that's changing urls, hooking into the browser page reloads etc, is either duplicating or will eventually conflict with [turbo](https://turbo.hotwired.dev/), which broadly covers this kind of functionality. I'm not saying we should use turbo to as part of the fix (although I'd happily review that) but what I want to avoid is making our lives harder in future.

tomhughes:
> > yet we've never really discussed if this is a problem that's worth solving
> 
> Yes we did. See your email on 2021-06-09.

I assume you are referring to where I wrote this:

> I wasn't aware there was a problem with that and the browser history - that just sounds like a bug but I'm not aware of anybody reporting it. Happy for it to be fixed. 

which was just a quick initial view based on the wishlist document rather than on this ticket and was also based on the assumption that a fix would be relative simple, probably by ensuring the correct URL parameters were in the history so the page would reload correctly.

AntonKhorev:
> > It seems that "fixing" this involves an awful lot of complicated and fragile javascript

~350 lines added, ~50 lines removed, 4 files and this is including tests.

[...]

PR #3682, opened 2022-09-09
Remember loaded changesets and scroll position in user edit history

AntonKhorev:
This is:

- "Improve User Edit History" on DWG wishlist
- #647
- #1089

Is saves changeset list html pieces to sessionStorage and remembers scroll position. Going to a history page, clicking some link then pressing *back* (or *forward*) should bring back the viewed history state along with everything that was loaded by pressing *Load more*. Reloading the page by pressing *reload* button throws away cached data.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -37,6 +37,7 @@ $(document).ready(function () {
     var title = sidebarContent[0];
     var html = sidebarContent[1];
 
+    map.setSidebarOverlaid(false);
     $("#flash").empty();
 
     var content = $(html);
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -414,4 +414,8 @@ $(document).ready(function () {
       e.preventDefault();
     }
   });
+
+  $(document).on("click", "#sidebar_content .btn-close", function () {
+    OSM.router.route("/" + OSM.formatHash(map));
+  });
 });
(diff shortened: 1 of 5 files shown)


mmd-osm:
I tested this on Firefox:

* I pressed "Load more" a few times
* clicked on a changeset further down in the list, and 
* pressed the back button to return to the list of changesets.
result: previous state was restored.

However, in some cases, when I click on the very first entry in the list of changesets, and then hit the back button, the previous state is lost somehow, and I only see the very first page of changesets. I cannot reliably reproduce the issue, it might be some timing issue.

Storing all the data as HTML fragments feels a bit heavy to me. As a browser only solution, you're a bit limited in what you can do client side  (Rails  returns the data as rendered HTML only.)  At least you have added a max entry limit of 10 lat/lon keys to avoid accumulating too much data.

mmd-osm:
We should also double check if we want home grown caching vs. https://github.com/hotwired/turbo-rails maybe.

AntonKhorev:
> We should also double check if we want home grown caching vs. https://github.com/hotwired/turbo-rails maybe. (The latter might not a good fit, after all)

https://github.com/openstreetmap/openstreetmap-website/issues/647#issuecomment-1298340289

mmd-osm:
Yes, that comment is actually the reason for my remark. If there's some chance we could offload all the caching to the framework, this may be easier to maintain in the long run. On the other hand, it's really not clear, if this framework would be a good fit at all.

PR #3688, opened 2022-09-13
Fix "Zoom in" tooltips on Edit button and Data/Notes layer checkboxes when they have to appear after not being shown initially

AntonKhorev:
Continuation of #3634 which I didn't have time to finish in August...

[As I said](https://github.com/openstreetmap/openstreetmap-website/pull/3634#issuecomment-1199687741) there's a different bug when tooltips are enabled/disabled and it needs a different fix. You can trigger the bug like this:

1. reload the website when the map is zoomed in enough for the control element (for example, the Edit button above the map) to be enabled and tooltip to be disabled
2. hover the mouse over the control element - nothing should happen at this moment
3. zoom out until the element is disabled
4. hover the mouse over the control element again - now you should get the error from this Bootstrap issue: https://github.com/twbs/bootstrap/issues/35375

The fix is to use [.disable()](https://getbootstrap.com/docs/5.1/components/tooltips/#disable) method on tooltip objects instead of trying to set the `data-bs-original-title` to empty string. `data-bs-original-title` is undocumented anyway.


tomhughes:
I had to add 6ffe31795ba8c116e612f3ad747f0522874de23f to wait for the tooltip to appear as the test was failing otherwise, both locally and in github actions.

I don't really understand why it's needed there and not in other tests which are doing similar things but the failure was very consistent though obviously it wasn't when this was originally opened :-(

Issue #3545, opened 2022-05-08
Welcome message buttons don't adjust to long translations


deevroman:
### URL

openstreetmap.org

### Screenshot(s) or anything else?

![SCR-20220508-pdo](https://user-images.githubusercontent.com/25708359/167303094-4724428f-c707-498a-8afc-a8f733a3ec07.png)

gravitystorm:
It might be possible to improve this using bootstrap button groups, but that will need some experiments.

PR #3690, opened 2022-09-14
Stick search input and Go button together with flex instead of input group

AntonKhorev:
There is an attempt to use an [input group](https://getbootstrap.com/docs/5.1/forms/input-group/) to combine the search input and the *Go* button. It doesn't work however. You can see the controls being rounded where they touch each other and this shouldn't happen:

search-before

The problem is trying to combine an input group with one input being put into a container that is required to position the *Where this is?* link. You have to remove the container but then positioning the link becomes more difficult. I could come up only with positioning using javascript, which is an unnecessary complication.

The simpler way is not to use an input group and specify which corners are to be rounded by hand:

search-after

### DIFF
--- a/app/views/layouts/_search.html.erb
+++ b/app/views/layouts/_search.html.erb
@@ -2,14 +2,12 @@
   <form method="GET" action="<%= search_path %>" class="search_form px-1 py-2">
     <div class="row gx-2 mx-0">
       <div class="col">
-        <div class="input-group flex-nowrap">
-          <div class='query_wrapper position-relative flex-grow-1'>
+        <div class="d-flex">
+          <span class='position-relative flex-grow-1'>
             <%= link_to t("site.search.where_am_i"), "#", :class => "describe_location position-absolute", :title => t("site.search.where_am_i_title") %>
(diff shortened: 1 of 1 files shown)


AntonKhorev:
Looking at the screenshot in #3545 this used to work correctly. It was broken probably by the Bootstrap update.

mmd-osm:
Does this also work with rtl languages? https://www.openstreetmap.org/?locale=ar

AntonKhorev:
No, it doesn't probably because r2 doesn't work correctly with the updated Bootstrap.

gravitystorm:
Thanks for the PR! 

This looks better than before, but as you point out, doesn't quite work on rtl pages (like other button groups e.g. edit/history/export). I suggest we tackle that issue separately.

AntonKhorev:
I had another version of this fix that kept input group and positioned the "where is this" link differently. Since it's not even a link I turned it into a button, which avoids the need for vertical positioning. All of that worked fine until I discovered that the search bar is used twice, one time inside the primary navigation. And the primary navigation has its primary button color redefined to green. Then I thought about sorting out color schemes, which led me to #4250. I didn't say anything here because I didn't expect this year-old pull request to get merged.

gravitystorm:
No worries, we can always revisit this when there are better options available, e.g. after reworking colours.

Issue #3422, opened 2022-01-15
install by docker,  Error: Invalid CSS after ".": expected 1 selector or at-rule, was "../common.scss"


windcatcher:
### 1. install it by docker.md
[https://github.com/openstreetmap/openstreetmap-website/blob/master/DOCKER.md](url)

### 2. open the website ,find the problem
**the website url** is [http://10.10.201.81:3000/#map=12/38.8938/-77.0146](url)
Showing /app/app/views/layouts/map.html.erb where line # raised:

Error: Invalid CSS after ".": expected 1 selector or at-rule, was "../common.scss"
        on line 1:1 of app/assets/stylesheets/ltr/common.scss
>> ../common.scss

### 3. run the command ,find this error info 
`bundle exec image_optim --info`

_root@7227613449f6:/app# bundle exec image_optim --info
image_optim v0.31.1
config:
  verbose: true
nice: 10
threads: 8
pack: false
skip_missing_workers: false
allow_lossy: false
cache_dir:
cache_worker_digests: false
timeout:
PATH: /var/lib/gems/2.7.0/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/var/lib/gems/2.7.0/gems/image_optim-0.31.1/vendor
Resolved jpegrescan 1a762f62 at /var/lib/gems/2.7.0/gems/image_optim-0.31.1/vendor/jpegrescan
Bin resolving errors:
pngcrush worker: `pngcrush` not found; please provide proper binary or disable this worker (--no-pngcrush argument or `:pngcrush => false` through options)
pngout worker: `pngout` not found; please provide proper binary or disable this worker (--no-pngout argument or `:pngout => false` through options)
advpng worker: `advpng` not found; please provide proper binary or disable this worker (--no-advpng argument or `:advpng => false` through options)
optipng worker: `optipng` not found; please provide proper binary or disable this worker (--no-optipng argument or `:optipng => false` through options)
pngquant worker: `pngquant` not found; please provide proper binary or disable this worker (--no-pngquant argument or `:pngquant => false` through options)
oxipng worker: `oxipng` not found; please provide proper binary or disable this worker (--no-oxipng argument or `:oxipng => false` through options)
jhead worker: `jhead` not found, `jpegtran` not found; please provide proper binary or disable this worker (--no-jhead argument or `:jhead => false` through options)
[...]

PR #3692, opened 2022-09-16
Add further Docker installation instructions for Windows users

lectrician1:
Adds documentation to tell Windows users to enable symlinks before cloning the repo so that they don't run into #3422.

Also gives alternatives for a few UNIX commands that aren't supported on Windows.


gravitystorm:
Thanks! I made a minor change to the markdown, but nothing significant. 

Thanks for figuring out the symlink stuff, it's definitely not something I would know how to address.

PR #3693, opened 2022-09-16
Clarify Windows installation bad + alternatives

lectrician1:
State some ruby gems might not be supported and Docker is easiest solution.


gravitystorm:
Thanks! I suspect the "order of easiest" might be a personal thing, but it seems reasonable to me.

PR #3694, opened 2022-09-17
Use Bootstrap button group for share tabs

AntonKhorev:
Removes all custom css under `.share-tabs`.

share


gravitystorm:
Excellent!

PR #3700, opened 2022-09-19
Replace <h5> used to set the font size with Bootstrap class h5

AntonKhorev:
This `<h5>` creates an [incorrect document outline](https://validator.w3.org/nu/?showoutline=yes&doc=https%3A%2F%2Fwww.openstreetmap.org%2Fabout).

### DIFF
--- a/app/views/site/about.html.erb
+++ b/app/views/site/about.html.erb
@@ -3,7 +3,7 @@
     <div class='row'>
       <div class='col-sm-7 user-image'></div>
       <div class='col-sm-5 px-5 py-3 byosm'>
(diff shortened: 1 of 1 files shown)


tomhughes:
By "incorrect document outline" do you mean the fact that `h5` follows `h1` with no intervening heading levels? As far as I know that has never been a requirement...

AntonKhorev:
I don't think this `<h5>` is a correct heading for anything on this page.

gravitystorm:
I'm not sure if every intermediate header needs to be included or not. But in any case, I'm happy to change this away from being a header, since it's not a section title or anything like that. If you were navigating the page with a screen reader, then having this appear in the (equivalent of) a table of contents would not be helpful.

PR #3701, opened 2022-09-19
Vertically center about page icons with flex align-items

AntonKhorev:
Icons on [the about page](https://www.openstreetmap.org/about) were shifted down too much because they were centered relative to headings with bottom margins. Also they were inside the headings which they shouldn't be, at least not as `<div>`s inside `<h2>`s.

Before:

about-before

After:

about-after

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1606,8 +1606,6 @@ dl.dl-inline {
   .icon {
     width: 30px;
     height: 30px;
-    margin-right: 10px;
-    vertical-align: middle;
     background: 40px 40px image-url('about/sprite.png') no-repeat;
 
     &.local {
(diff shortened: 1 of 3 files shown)


tomhughes:
Merged, though to my eyes the image now looks too high - it appears be be aligned on the baseline of the text rather then centred with it.

i think maybe it's because the descenders on the text are being included in the calculation, but the eye doesn't really see them when considering whether the image and text appear to be centred :-(

AntonKhorev:
Looks like it depends on the font. With Ubuntu's *Ubuntu Medium* font icons look centered correctly. On Windows *Segoe UI* is used and icons look higher relative to the text.

gravitystorm:
> i think maybe it's because the descenders on the text are being included in the calculation, but the eye doesn't really see them when considering whether the image and text appear to be centred :-(

Yeah, I had left adjusting these icons out of my previous refactoring on that page since I hadn't come up with a good solution. Maybe a proper designer would know the answer, but I don't know what the best solution would be - maybe adjusting the size of the icons to match the [baseline to cap height](https://en.wikipedia.org/wiki/Baseline_(typography)) or something like that. But then we're deep in the weeds of font-specific solutions, and then checking that the icon positioning is still appropriate for e.g. chinese characters, and I think that's too much to consider, given our limited resources.

AntonKhorev:
> maybe adjusting the size of the icons to match the [baseline to cap height](https://en.wikipedia.org/wiki/Baseline_(typography))

Font size is variable, depending on the window width.

PR #3704, opened 2022-09-19
Group map controls explicitly

AntonKhorev:
Avoid setting borders on specific map controls in css.

Why do  `.control-locate`, `.control-key` etc have their borders/margins altered in css? Because they are first or last buttons in groups, but this is not obvious from looking at css only. Here this is replaced by the code that actually groups controls. Now it's possible to add/regroup/remove controls in index.js without having to set their borders and margins manually.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -109,8 +109,14 @@ $(document).ready(function () {
       control.addTo(map);
     });
 
-    var lastContainer = controls[controls.length -1].getContainer();
-    $(lastContainer).addClass("leaflet-control-group-end");
+    var firstContainer = controls[0].getContainer();
+    $(firstContainer).find(".control-button").first()
+      .css("border-start-start-radius", "4px");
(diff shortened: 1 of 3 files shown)


gravitystorm:
Thanks for the PR, and thanks for your patience too.

A couple of things that I thought of while reviewing this PR:
* I don't know if there's any way to avoid specifying the classes for the first and last item in the group? Bootstrap button groups work (somehow, I haven't investigated) just based on the position of the button in the button group
* The rounding would be even better if we used the SASS variables from bootstrap - e.g. `border-start-start-radius: $border-radius;` so that they match the rest of the site automatically (although given the small sizes, it's not currently noticeable
* The rounding would be even better than that, if we could use bootstrap classes to control the rounding (so that if e.g. we have responsive rounding radii, it works automatically). But I can't find any description of how to round only one corner with the bootstrap docs, so maybe that's not possible.

Thanks again!

PR #3709, opened 2022-09-22
Show/hide query intro text depending on query mode

AntonKhorev:
When you use *Query features* map tool, a sidebar shows with this message "Click on the map to find nearby features." The message had a `query-intro` css class added in https://github.com/openstreetmap/openstreetmap-website/commit/078059b76b. Then the class was removed in https://github.com/openstreetmap/openstreetmap-website/commit/c8f0a81eb7043c297e050c1f3ff8a8cf7ff613bc. However there's still the js code that tries to hide this message based on its class. This happens as soon as the query runs.

But the query runs as soon as the sidebar is shown. If hiding is restored, the user won't have time to read the message, and then there's no point in writing it. On the other hand, if the message is not hidden, it will tell the user to click the map even when clicking will have no effect.

Here I'm showing/hiding the message depending on if it's possible to click the map and get a query to run.

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -21,19 +21,19 @@ OSM.Query = function (map) {
 
     if (queryButton.hasClass("active")) {
       disableQueryMode();
+      deactivateQueryMode();
     } else if (!queryButton.hasClass("disabled")) {
+      activateQueryMode();
       enableQueryMode();
     }
   }).on("disabled", function () {
     if (queryButton.hasClass("active")) {
-      map.off("click", clickHandler);
-      $(map.getContainer()).removeClass("query-active").addClass("query-disabled");
+      disableQueryMode();
       $(this).tooltip("show");
     }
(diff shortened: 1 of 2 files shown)


tomhughes:
As you say the query runs as soon as the sidebar opens, so is the sidebar intended to open as soon as you activate the tool, and then remove the message once you start he query?

You say you've hiding/showing it based on whether you can trigger a query with a click, but how exactly do you get to a state where the sidebar is open and you can trigger a query with a click?

AntonKhorev:
> but how exactly do you get to a state where the sidebar is open and you can trigger a query with a click?

When the query runs or when it is finished you are still in that state. You can click again and run another query. To exit the state you need to either close the sidebar (that "deactivates" the query mode) or zoom out (that "disables" the queries - I'm naming these states after css classes the query button gets). If you zoom out, you get into "activated disabled" state where you can zoom in and get back to "activated enabled" state.

PR #3712, opened 2022-09-22
Fix margin

danieldegroot2:
Fix unexpected margin introduced in https://github.com/openstreetmap/openstreetmap-website/commit/34edfdb9bae297144331b46dfe52319d13e519a3#diff-7085085f26b51cc428bf0a4dce47f5399822c07573a8eaa0c85ac64ce1cb0387R1107

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1128,7 +1128,7 @@ tr.turn:hover {
   min-width: 200px;
   p {
     padding: 0 0 5px 0;
-    margin-top: 0 0 0 60px;
+    margin: 0 0 0 60px;
     font-size: 12px;
   }
 }


tomhughes:
In what way is it "unexpected" exactly? What part of the site do I look at to see the change which this makes?

danieldegroot2:
The different margin values (top, bottom, left) should've been merged to `margin`, not `margin-top`. `margin-top` expects a single value for the top margin only.
From looking at the code, I think it is for user pop-ups on the 'nearby users' map on the "My Dashboard" page. Not sure what exactly.
https://github.com/openstreetmap/openstreetmap-website/blob/f62bd9f10df8511f498603236aecc41329781e6a/app/assets/stylesheets/common.scss#L1125

tomhughes:
So are you saying you're basing this change on a code review of a change from nearly ten years ago rather than any observed problem with the site? I guess you just noticed the odd `margin-top` with four values?

danieldegroot2:
Yes, I found it through the [W3C CSS validator](https://jigsaw.w3.org/css-validator/validator?uri=https%3A%2F%2Fwww.openstreetmap.org%2F&profile=css3svg&usermedium=all&warning=1&vextwarning=&lang=nl). If it doesn't break anything, feel free not to fix it, of course.

tomhughes:
Merged, but I don't think that rule ever has any effect because https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/stylesheets/leaflet-all.scss#L19 overrides it.

PR #3714, opened 2022-09-23
Allow setting JSON return type by HTTP ACCEPT header for notes API

Robbendebiene:
Currently the notes API json return type can only be specified by appending the .json file extension at the end of the URL path. This is inconsistent to other JSON API entry points which also allow specifying the return type via the HTTP ACCEPT header.

I guess the routing tests (see example below) do not catch these cases because they do not allow to specify headers.

```
      assert_routing(
        { :path => "/api/0.6/notes/1.json", :method => :get },
        { :controller => "api/notes", :action => "show", :id => "1", :format => "json" }
      )
```

### DIFF
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -9,6 +9,7 @@ module Api
 
     before_action :set_locale
     around_action :api_call_handle_error, :api_call_timeout
+    before_action :set_request_formats, :except => [:feed]
 
     ##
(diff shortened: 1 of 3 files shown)


tomhughes:
This causes a test failure - the routing test needs to be updated to remote the format.

Robbendebiene:
@tomhughes I've removed all xml format checks from the notes controller test where the format was not explicitly defined in the request/path.

PR #1189, opened 2016-03-18
Support version for /nodes, /ways, /relations

simon04:
Fixes #1067.

Adds the following API methods:
- `/api/0.6/nodes/versions?nodes=123v4,56v7`
- `/api/0.6/ways/versions?ways=123v4,56v7`
- `/api/0.6/relations/versions?relations=123v4,56v7`

PR #3715, opened 2022-09-23
Support versions in elements multi fetch

AntonKhorev:
As supported by cgimap. Related to https://github.com/openstreetmap/openstreetmap-website/pull/1189 which is a different implementation with different API entry points.

### DIFF
--- a/app/controllers/api/nodes_controller.rb
+++ b/app/controllers/api/nodes_controller.rb
@@ -67,15 +67,18 @@ module Api
       render :plain => node.version.to_s
     end
 
-    # Dump the details on many nodes whose ids are given in the "nodes" parameter.
+    # Dump the details on many nodes whose ids and optionally verions are given in the "nodes" parameter.
     def index
-      raise OSM::APIBadUserInput, "The parameter nodes is required, and must be of the form nodes=id[,id[,id...]]" unless params["nodes"]
+      raise OSM::APIBadUserInput, "The parameter nodes is required, and must be of the form nodes=ID[vVER][,ID[vVER][,ID[vVER]...]]" unless params["nodes"]
 
-      ids = params["nodes"].split(",").collect(&:to_i)
+      id_ver_strings, id_strings = params["nodes"].split(",").partition { |iv| iv.include? "v" }
+      id_vers = id_ver_strings.map { |iv| iv.split("v", 2).map(&:to_i) }
+      ids = id_strings.map(&:to_i)
 
       raise OSM::APIBadUserInput, "No nodes were given to search for" if ids.empty?
 
       @nodes = Node.find(ids)
+      @nodes += OldNode.find(id_vers) unless id_vers.empty?
 
       # Render the result
       respond_to do |format|
--- a/app/controllers/api/nodes_controller.rb
+++ b/app/controllers/api/nodes_controller.rb
@@ -78,7 +78,16 @@ module Api
       raise OSM::APIBadUserInput, "No nodes were given to search for" if ids.empty?
 
       @nodes = Node.find(ids)
-      @nodes += OldNode.find(id_vers) unless id_vers.empty?
+      unless id_vers.empty?
+        @nodes += OldNode.find(id_vers)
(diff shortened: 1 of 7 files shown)


mmd-osm:
ElementsApiController seems like a good way to maximize code reuse. However, I find things like  `current_model = controller_name.classify.constantize` and `old_model = "Old#{controller_name.classify}".constantize` too difficult to read for a low skilled dev. Given that it would only save us a few lines of code, it feels a bit over-engineered to me.

I wouldn't mind to see a bit more code duplication here, for the sake of better code readability.

AntonKhorev:
> Given that it would only save us a few lines of code, it feels a bit over-engineered to me.

With a proper abstract base class, `NodesController` can be reduced to basically nothing.

> things like current_model = controller_name.classify.constantize and old_model = "Old#{controller_name.classify}".constantize too difficult to read

Looks like Rails still doesn't provide anything better than that to get model classes without naming them directly. But if that's really too scary, I can make a couple of methods in child controller classes that return their model classes.

mmd-osm:
> But if that's really too scary, I can make a couple of methods in child controller classes that return their model classes.

I also found `The parameter #{controller_name} is required` a bit too extreme. Yes, it's super generic, but might be too fragile going forward (think: translations, maybe) . To be honest, I would completely remove the ElementsApiController, and add around 10 lines to each of three controllers. Perfectly fine for me.

Let's wait what both maintainers think about it, before doing any changes.

AntonKhorev:
> I also found The parameter #{controller_name} is required a bit too extreme. Yes, it's super generic, but might be too fragile going forward (think: translations, maybe)

How do you think the translation string would look like? I think it would be about as extreme. And again, I can do abstract method returning the message.

> I would completely remove the ElementsApiController

Why not remove ApiController too?

> and add around 10 lines to each of three controllers.

I'd remove around 80 lines from each.

I already managed to sneak in an abstract class removing triple copy-paste, although it was in javascript.

tomhughes (commented):
I was wondering how on earth cgimap wound up with an API extension relative to rails and it it turns out the history is far older and more baroque than I was expecting.

The cgimap version came in with https://github.com/zerebubuth/openstreetmap-cgimap/pull/129 which also added the single object versions calls which rails does have - that claims to be implementing #1189 for cgimap but actually is using a subtly different syntax.

That is probably a rather unfortunate syntax in fact because unlike #1189 it doesn't maintain the separation between main controller methods and history controller methods which then leads to the code here having to a complicated dance with the main controller accessing history objects and also to the non-deterministic uniqification where if you ask for the current version of an object in both versioned and unversioned form you don't know which one will be returned.

Now of course they should always be the same but it's an interesting wart.

I don't see any obvious problem with the code other than the rather complex nature of the abstract parent as @mmd-osm  noted which is not ideal but then equally neither is duplicating everything.

Possibly it would better to add new tests rather than add lots of cases to the existing ones. I suspect @gravitystorm will think so though he'd probably also say we should split the existing one.

AntonKhorev:
Updates:
- moved the generic elements controller into api module; this module already had the generic old elements controller `OldController`
- put version multifetch tests into different methods
- replaced `.classify.constantize` with abstract `current_model`/`old_model` methods

mmd-osm (approved):

AntonKhorev:
Removed `controller_name`, using `type_plural = current_model.model_name.plural` instead.

AntonKhorev:
Updated for resourceful routes.

AntonKhorev:
I decided to address this completely:

> Possibly it would better to add new tests rather than add lots of cases to the existing ones. I suspect @gravitystorm will think so though he'd probably also say we should split the existing one.

and split all `index` action tests, exisitng and added in this PR. Then I discovered that I never checked if old versions of elements were redacted. Added a fix for that too.

The expected behavior if one of the requested versions is redacted wasn't documented anywhere, so I copied what cgimap did, which is 404 like for missing versions and missing elements.

AntonKhorev:
For old elements tags can't be preloaded for the same reasons as in #4569.

Issue #957, opened 2015-04-17
For remote control add info about notes


Klumbumbus:
It would be nice if you use remote control to download mapdata in e.g. JOSM and a note ( "/note/123456" in the URL) or the notes layer ("&layers=N" in the URL) is active, JOSM downloads also the note(s). According to JOSM ticket https://josm.openstreetmap.de/ticket/9907#comment:17 a modification of the remote control at osm.org is required for this.

tomhughes:
Is there some documentation somewhere on the new parameters that have been added to the remote control?

tomhughes:
Looks like https://github.com/iandees/josm-notes/issues/23 is still open, so presumably this hasn't been implemented on the JOSM side yet?

Klumbumbus:
iandees/josm-notes#23 is about the old josm notes plugin, which was replaced by a josm core notes feature solution. Recently notes download via remote control was added (see https://josm.openstreetmap.de/changeset/8197/josm).

simon04:
One idea would have been to evaluate the referer header sent. However, it is always `Referer: http://www.openstreetmap.org/` (omitting the content after the hash).

The easiest solution would probably be to add the `layers` information from the hash to the remote control request?
```diff
diff --git a/app/assets/javascripts/index.js b/app/assets/javascripts/index.js
index 085b615..419ec73 100644
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -214,6 +214,7 @@ $(document).ready(function () {
         "https://127.0.0.1:8112/load_and_zoom?" :
         "http://127.0.0.1:8111/load_and_zoom?",
         query = {
+          layers: '', // TODO obtain from location.hash
           left: bbox.getWest() - 0.0001,
           top: bbox.getNorth() + 0.0001,
           right: bbox.getEast() + 0.0001,
```

tomhughes:
I wasn't planning to do layers - there is no precedent for that.

My immediate concern was loading the specific note that is active, as we already do that for other object types. It turns out to be complicated however, and requires some significant re-engineering of the code.

simon04:
Precedent: I noticed that. However, determining whether a specific layer was active is a matter of a simple string-contains or regexp test.

Hm, is there an other way to determine whether notes are being displayed?

Klumbumbus:
Similar to notes there is also a request for downloading a gpx trace. See https://josm.openstreetmap.de/ticket/13749

Issue #1215, opened 2016-05-06
When looking at a hidden note, you can't then navigate to another note.


DeeHants:
After I've [hidden a note](http://www.openstreetmap.org/note/506777), I can't then navigate to any other notes by clicking on them.
I'm a member of the DWG so this may differ for people without moderator access.

Using IE11 on Windows 7.

harry-wood:
people without moderator access (like me) can't see hidden notes. We just see _"Sorry, #506777 could not be found"_. So I guess that makes this a lower priority issue

Issue #546, opened 2013-11-28
Note to history transitions after hiding notes


pnorman:
Steps to reproduce
1. Log in on a moderator account on master.apis
2. Add a note
3. Click on the note
4. Hide the note (requires moderator role)
5. Click on History

The note disappears from the right-hand map, but history is not shown. Additionally the sidebar cannot be closed with the 'X' button.

Issue #507, opened 2013-10-09
Missing hidden_note_marker


pnorman:
http://www.openstreetmap.org/images/hidden_note_marker.png is missing. Used on note browse pages for notes that have been hidden by a moderator.

RicoElectrico:
What do you think about this marker (last one)?
I tried a dustbin and a crossed out eye, both aren't really legible at that size given marker style they need to fit into.
![obraz](https://user-images.githubusercontent.com/1409116/178623306-8061cc9c-e05e-4faf-b1ce-652f919332fe.png)

PR #3718, opened 2022-09-24
Load notes with remote control

AntonKhorev:
Does part of #957: loads a note when running a remote control edit from `/note/*id*` page.

Normally you do remote control with `load_and_zoom` rc command, and it doesn't support notes. But it's not a problem because you can run another rc command after that. Notes can be loaded by `import` command with a url parameter pointing to note data.

### DIFF
--- a/app/assets/javascripts/index/note.js
+++ b/app/assets/javascripts/index/note.js
@@ -1,7 +1,6 @@
 OSM.Note = function (map) {
   var content = $("#sidebar_content"),
-      page = {},
-      halo, currentNote;
+      page = {};
 
   var noteIcons = {
     "new": L.icon({
@@ -35,9 +34,9 @@ OSM.Note = function (map) {
     });
   }
 
-  page.pushstate = page.popstate = function (path) {
+  page.pushstate = page.popstate = function (path, id) {
     OSM.loadSidebarContent(path, function () {
-      initialize(function () {
+      initialize(id, function () {
         var data = $(".details").data(),
             latLng = L.latLng(data.coordinates.split(","));
         if (!map.getBounds().contains(latLng)) moveToNote();
@@ -45,11 +44,11 @@ OSM.Note = function (map) {
     });
   };
 
-  page.load = function () {
-    initialize(moveToNote);
+  page.load = function (path, id) {
+    initialize(id, moveToNote);
   };
 
-  function initialize(callback) {
+  function initialize(id, callback) {
     content.find("input[type=submit]").on("click", function (e) {
       e.preventDefault();
       var data = $(e.target).data();
@@ -70,28 +69,14 @@ OSM.Note = function (map) {
 
     content.find("textarea").val("").trigger("input");
 
-    var data = $(".details").data(),
-        latLng = L.latLng(data.coordinates.split(","));
+    var data = $(".details").data();
 
-    if (!halo || !map.hasLayer(halo)) {
-      halo = L.circleMarker(latLng, {
-        weight: 2.5,
-        radius: 20,
-        fillOpacity: 0.5,
-        color: "#FF6200"
-      });
-      map.addLayer(halo);
-    }
-
-    if (currentNote && map.hasLayer(currentNote)) map.removeLayer(currentNote);
-
-    currentNote = L.marker(latLng, {
-      icon: noteIcons[data.status],
-      opacity: 1,
-      interactive: true
+    map.addObject({
+      type: "note",
+      id: parseInt(id, 10),
+      latLng: L.latLng(data.coordinates.split(",")),
+      icon: noteIcons[data.status]
     });
-
-    map.addLayer(currentNote);
 
     if (callback) callback();
   }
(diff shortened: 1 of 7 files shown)


tomhughes:
It would be really helpful if you could keep your PRs to what the subject says they do - if you want to refactor things then it's better to do that as a separate PR rather than throwing them in with a functional change.

I started to review this thinking it was a small change only to find it's much more complicated than than and is going to need significant time to review it properly :-(

AntonKhorev:
I didn't want to refactor anything but the code is written in such a way...

The biggest change here is that the map object also gets to treat notes as "current objects". Before that it used to do this with elements and changesets.

Why did I do this:
1. I want to open the note with remote control when the edit link is clicked. But the edit link doesn't contain any information about the note.
2. Then I need to change the code that updates the link. I discover that it takes the "current object" from the map and adds it as a parameter. But the note is never added as a "current object" to the map.
3. Then I have to add notes as "current objects". That's the most significant change.

If you want all this to be split in smaller chunks, the first one would be the "current object". Actually it's the first commit, except for the test cases (https://github.com/openstreetmap/openstreetmap-website/pull/3718/commits/02ab1d338cfb8837c4623e7cebfc25ec0ae360b2) + the second commit to fix how the note id is passed when commenting/closing/reopening he note (https://github.com/openstreetmap/openstreetmap-website/pull/3718/commits/591a7464127f5a0e884e34cc4480d257ebc44cb9). Do you want a separate PR with that?

AntonKhorev:
Or I can make the note layer loading first. This is going to be a smaller change because it doesn't require dealing with the "current object".

app/assets/javascripts/index/note.js
```diff
@@ -20,40 +19,41 @@ OSM.Note = function (map) {
       iconAnchor: [12, 40]
     })
   };
+  noteIcons.hidden = noteIcons.closed; // TODO replace with actual "hidden" icon when it's added
```
tomhughes, line 22: Why do we want an icon for hidden notes? They are supposed to be hidden...

app/assets/javascripts/osm.js.erb
```diff
@@ -37,7 +37,8 @@ OSM = {
   SEARCHING:               <%= image_path("searching.gif").to_json %>,
 
   apiUrl: function (object) {
-    var url = "/api/" + OSM.API_VERSION + "/" + object.type + "/" + object.id;
+    var apiType = object.type === "note" ? "notes" : object.type;
```
tomhughes, line 40: This is kind of ugly... I'm half inclined to say why not just use `notes` as the object type, then you wouldn't need to jump through hoops here to fix things up.

Then again that means internally we are inconsistent about whether the object type is singular or plural, which also offends me ;-)

tomhughes (commented):
I really don't understand some of the breakdown of changes into commits here - the first and last commits are fairly obvious but 591a7464127f5a0e884e34cc4480d257ebc44cb9 and 4f9d7fbe7f36b449c9ed544969a8313763240f7d have descriptions that don't really explain what the commit is doing and for the second one especially I don't really understand what either of the two small changes have to do with each other or with the title of the commit.

app/assets/javascripts/index/note.js
AntonKhorev: Do you want to hide notes from moderators too or do you want to keep this bug?
https://github.com/openstreetmap/openstreetmap-website/issues/546
https://github.com/openstreetmap/openstreetmap-website/issues/1215

I fixed it here to check if other things are working. Of course I can put it back if you want.

I'm using the "closed" icon because I assume that the actual "hidden" icon, like introduced here https://github.com/openstreetmap/openstreetmap-website/issues/507 , still doesn't exist.

app/assets/javascripts/osm.js.erb
AntonKhorev: Not using apiUrl() is an option. Then the check for object.type will go to the calling code. "note" as a type is used by iD embedding code. While iD won't display a note, it's being directed to open the note location with links like https://www.openstreetmap.org/edit?note=1000 .

AntonKhorev:
The breakdown by commit is not ideal because I didn't know what I'd have to change when I started writing. Everything here can be broken down into six independent parts:

1. making notes to be "current objects" of the map
2. using note current objects to update edit links
3. calling the remote control commands - these 3 parts are required to see any working changes
4. test - previous 3 parts are required unless you want to tolerate failing tests
5. using the "closed" note marker as the "hidden" one too - not required, as long as you want javascript errors to keep happening when the "Hide" button is pressed
6. removing note id data attributes from the buttons in the view template - this is not required, but they are not used anywhere because the path is also written for each button and it contains the id. I removed the ids while I was thinking how to pass one to the edit link. The code before my changes didn't require these ids at all, they were ignored. Then there's also a client-side router for pages that contain a map. It is set up here: https://github.com/openstreetmap/openstreetmap-website/blob/f398c1e5eb35fa9e355544fb2860bc2ca9cfbe67/app/assets/javascripts/index.js#L360 It also passes the id to the note route handler - and it was also ignored. I decided to use it.

Issue #282, opened 2013-05-14
Create a rake task for populating the database


gravitystorm:
Currently the installation notes spend about 90 lines explaining how to populate the database from an extract. A substantial portion of this concerns osmosis options and resetting sequences.

This could be simplified by creating a rake task that calls osmosis with the configured database parameters and takes care of the sequences.

hanchao:
@danstowell Error occurred. Any help?

[hanchao07@dx-wmarch-map-routing01 ~]$ ./osmosis/bin/osmosis --truncate-apidb host="waimai-pgsql-wmmapdata01-write.vip.sankuai.com" database="wmgis" user="wmgis_rw" password="3ab8CPjB0C1mx6q7fdsuOuLkp" validateSchemaVersion="no"
九月 19, 2016 5:14:58 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Osmosis Version 0.45
九月 19, 2016 5:14:58 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Preparing pipeline.
九月 19, 2016 5:14:58 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Launching pipeline execution.
九月 19, 2016 5:14:58 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Pipeline executing, waiting for completion.
九月 19, 2016 5:14:59 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Pipeline complete.
九月 19, 2016 5:14:59 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Total execution time: 674 milliseconds.

[hanchao07@dx-wmarch-map-routing01 ~]$ ./osmosis/bin/osmosis --read-pbf file=china-latest.osm.pbf --write-apidb host="waimai-pgsql-wmmapdata01-write.vip.sankuai.com" database="wmgis" user="wmgis_rw" password="****" validateSchemaVersion="no"
九月 19, 2016 5:15:11 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Osmosis Version 0.45
九月 19, 2016 5:15:11 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Preparing pipeline.
九月 19, 2016 5:15:11 下午 org.openstreetmap.osmosis.core.Osmosis run
信息: Launching pipeline execution.
九月 19, 2016 5:15:11 下午 org.openstreetmap.osmosis.core.Osmosis run
[...]

PR #3719, opened 2022-09-25
add some notes about importing data to INSTALL.md

saerdnaer:
relates to #282

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -214,17 +214,13 @@ bundle exec rake db:migrate
 
 ### Optional: Database content
 
-If you want real world data in your API DB, you can download some data and import it with [Osmosis](https://wiki.openstreetmap.org/wiki/Osmosis#How_to_install). Data can be downloaded as OSM XML via JOSM or as OSM PBF extract from e.g. https://download.geofabrik.de.
+If you want real world data in your API DB, you can download some data and import it with [Osmosis](https://wiki.openstreetmap.org/wiki/Osmosis#How_to_install). Sample data can for example be downloaded per JOSM as OSM-XML and imported via:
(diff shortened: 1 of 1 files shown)


mmd-osm:
I think this has been documented elsewhere already: https://github.com/openstreetmap/openstreetmap-website/blob/master/DOCKER.md#loading-an-osm-extract

Maybe it would be good to consolidate this a bit?

danieldegroot2:
(also at [CONFIGURE.md#populating-the-database](https://github.com/openstreetmap/openstreetmap-website/blob/b8d83738158e3efc0792ef2c66bb41e846ed28ec/CONFIGURE.md#populating-the-database) )

saerdnaer:
For now I just cross-linked the other pages, the link to `CONFIGURE.md` was on really visible to me.

tomhughes:
Well I think the idea is to separate core things that everybody needs from detailing every obscurity of interest only to a few people.

tomhughes:
Note that I at least don't recommend trying to load data into an API database because it's very hard to do right and there aren't really the tools to do it well.

gravitystorm:
The documentation is deliberately separated into INSTALL.md and CONFIGURE.md, so things like populating the database should appear in the CONFIGURE.md - which it already does!

I would recommend that any changes you want to make should be put there, rather than being duplicated between multiple locations. Additionally, you should confirm that the stuff about disabling indexes is accurate - I don't like having our documentation saying "probably", when we could test it and find out for sure.

Also, you mentioned that you hadn't seen the documentation in CONFIGURE.md before creating this PR, so if you can think of any way to make that documentation more obvious, that would be great.

PR #3720, opened 2022-09-25
Fix formatting for 'et cetera'

danieldegroot2:
Fix formatting for 'et cetera'. Excludes vendor/licensed files.


tomhughes:
I think we can manage with the grammar police thanks. I have no particular objection to oxford commas or a trailing dot on etc but neither do I see any particular need to add them just to confirm to some individual's idea of what is correct.

PR #3728, opened 2022-09-30
Added some language codes for Pakistan to countries.xml

bgo-eiu:
More could be added, but the ones I have included are particularly likely to have translations become available via Translatewiki. This includes the addition of `pnb`, which now has `pnb.yml` in the repository as of this week.

The locale code for `pnb` in Transifex is `pa-PK`, so I have to put both in my settings in order to see iD in Punjabi Shahmukhi. I am not sure if there is a way to alias these in the locale settings

### DIFF
--- a/config/countries.xml
+++ b/config/countries.xml
@@ -3375,7 +3375,7 @@
 <areaInSqKm>803940.0</areaInSqKm>
 <population>184404791</population>
 <currencyCode>PKR</currencyCode>
-<languages>ur-PK,en-PK,pa,sd,ps,brh</languages>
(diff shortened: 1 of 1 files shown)


tomhughes:
That is completely pointless - we do not use the language codes in that file for anything.

bgo-eiu:
Was not sure, but `countries.xml` is still in one of the Ruby files

tomhughes:
Yes, but we only use the bounding box data, not anything else. It is used for mapping a country code from IP geolocation to a default map view for new visitors.

The file itself comes from https://www.geonames.org/export/ so has all sorts of other data we don't use.

bgo-eiu:
Do you know if the locale in the preferences can be aliased if different for the website and iD?

tomhughes:
I thought we did but apparently not though a user can list both codes in their preferred languages and that should work.

PR #3729, opened 2022-09-30
Punjabi + Luba-Katanga updates to languages.yml

bgo-eiu:
This PR adds `pnb` name and native label and updates `pa` name and native label, as these locale codes are for the same language in different scripts; `pa` being Gurmukhi used in India, `pnb` being Shahmukhi used in Pakistan.

I also noticed that Luba-Katanga was missing a native label and just had `~`, so I looked this up and added the native name Kiluba.

### DIFF
--- a/config/languages.yml
+++ b/config/languages.yml
@@ -312,7 +312,7 @@ lt:
   native: Lietuvių
 lu: 
   english: Luba-Katanga
-  native: ~
+  native: Kiluba
 lv: 
   english: Latvian
   native: Latviešu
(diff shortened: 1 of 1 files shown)


tomhughes:
That file is not manually maintained - it is generated using `script/locale/update-languages` so if it needs updating you should run that and submit a PR using the results.

bgo-eiu:
OK - the TranslateWiki name for pnb is incorrect; they are aware of this so I will do that whenever they fix it

PR #3731, opened 2022-10-03
Fixes delimiter direction in map key

bgo-eiu:
Screenshot_20221003-182747

This screenshot of the map key in Punjabi above shows the situation this PR is intebded to address, where "and" and opposite direction commas appear interspersed with the localized map key. This is because of the localized strings being joined with `to_sentence`, which is not aware of every locale and how to delimit a list for each one. Rather than asking all translators to translate "and" and provide a comma, this just uses either a left-to-right comma or a right-to-left comma depending on the known direction in order to keep things simple.

### DIFF
--- a/app/views/site/key.html.erb
+++ b/app/views/site/key.html.erb
@@ -7,7 +7,18 @@
             <%= image_tag "key/#{name}/#{entry['image']}" %>
           </td>
           <td class="mapkey-table-value">
-            <%= Array(t(".table.entry.#{entry['name']}")).to_sentence %>
+            <% if I18n.t("html.dir") = "rtl" %>
+              <%= Array(t(".table.entry.#{entry['name']}")).to_sentence(
+                words_connector: '، ',
+                two_words_connector: '، ',
+                last_word_connector: '، ') %>
+            <% end %>
+            <% if I18n.t("html.dir") = "ltr" %>
+              <%= Array(t(".table.entry.#{entry['name']}")).to_sentence(
+                words_connector: ', ',
+                two_words_connector: ', ',
+                last_word_connector: ', ') %>
+            <% end %>
           </td>
         </tr>
       <% end %>


tomhughes:
If `to_sentence` isn't working correctly then the solution is to fix the upstream rails code and/or translations, not to render it pointless by overriding all it's decisions locally.

bgo-eiu:
`to_sentence` likely gets its data from CLDR which is not interested in adding support for Punjabi based on my contact wirh them offering to add info through their survey tool earlier in the year.

It's also not the best choice to use "and" in this context in every language. In Punjabi, "and" and "on" are homographs,  and "on" is used instead of "in" to describe location. So if we want to say for example, "parks and gardens," we would say پارک تے باغ ("paark te baag) and if we want to say "gardens in parks" this is also پارک تے باغ. This is not a problem in a paragraph because it is usually obvious what is meant, but in a map key it is not actually clear what is meant. So even if `to_sentence` was able to generate an  "and" sentence for Punjabi, we would still have to override it for this language at least in order for the key to be unambiguous.

bgo-eiu:
Ampersand is also not an option as it is not supposed to be used in most Arabic-based scripts (it faces the wrong direction). A right to left ampersand technically exists ۽ but it is only used in Sindhi readers of other languages don't know this character

tomhughes:
The strings for `to_sentence` are in the [rails-i18n](https://github.com/svenfuchs/rails-i18n/) repository on github, for example the English ones are https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml#L206.

You'll see from the many accepted PRs there that they are happy to take changes and while CLDR may be an influence it's certainly not just a mechanical import.

I can't speak to the experience you had with CLDR but while it's not a quick process I've always found them helpful.

bgo-eiu:
Ah, glad to know I am wrong about that, recently in translating various software I've had to justify overriding aomething or other from CLDR, but I will make a PR for them. (CLDR inexplicably suggests Punjabi isn't a real language multiple times in their "Picking the Right Language" documentation. Who knows why, but I'm not the only person who has pointed this out to them.)

In this particular context, I would still suggest making the map key comma delimited for the ambiguity reason above. There are likely other languages where "and" is an ambiguous word (it usually is a very short one after all).

gravitystorm:
Why are we even trying to build sentences from translated strings in the first place? Wouldn't it be easier for everyone to have the full sentence in the original translation key, and then the translators can do the right thing in their language.

For example, we have keys like 

```
tram:
  - Light rail
  - tram
```
... but what is the advantage of having the sentence split into component parts, translating the parts, and then attempting to assemble a sentence? Why not just have `tram: "Light rail and tram"` instead?

bgo-eiu:
That sounds like a good idea to me. I assumed the main concern is creating a lot of additional work for translators, but maybe there is an efficient way to make that change.

The caveat would be that it should be a list of features rather than a word which can describes both as these are meant to describe particular similar but distinct keys. This can be made clear in the message documentation however.

bgo-eiu:
If some adjustment of the translation strings is on the table, I have created #3740 to track problems related to asking translators to provide fragments of sentences generally, as there are at least a few places where this could be improved

bgo-eiu:
OK - I think allowing translation of the full strings is the way to go, so I will close this and open a new PR when I have it ready.

Further, I had missed that there are currently inaccuracies in the map key anyway - allotments are rendered in light green, not brown, for example. Translating these is kind of pointless until this is fixed first

gravitystorm:
> OK - I think allowing translation of the full strings is the way to go, so I will close this and open a new PR when I have it ready.

Great, thanks in advance!

Note that for technical reasons, we can't use the same translation key if we change how it works. Changing the text is fine, but things like changing placeholders, or in this case, changing arrays to strings won't work. This is because there's a period of time where some translations haven't been updated to use the new format. So for example, if you want to change the `en.site.foo.bar` key from an array to a string, you need to change the key to `en.site.foo.something_else` too.

bgo-eiu:
Thank you for the pointers, I had a look at some of the other locales to see what the out of date translations look like and I see what you mean in this regard

PR #3733, opened 2022-10-04
Remove unused note id data attributes

AntonKhorev:
See 6. in https://github.com/openstreetmap/openstreetmap-website/pull/3718#issuecomment-1261480992

### DIFF
--- a/app/views/browse/note.html.erb
+++ b/app/views/browse/note.html.erb
@@ -49,10 +49,10 @@
         </div>
         <div class="btn-wrapper">
           <% if current_user.moderator? -%>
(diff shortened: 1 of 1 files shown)


mmd-osm:
Just a general comment on those data attributes: some people use them to filter out stuff they’re not interested in.

While that’s probably not a concern for note ids, maybe keep in mind that there might be users outside of the repo whose workflow might be impacted.

if I’m not completely mistaken @systemed  mentioned this at some point, but I’m unable to find it right now.

AntonKhorev:
These attributes are on form buttons here. The same id repeated on every button and never read.

PR #3738, opened 2022-10-05
install bundler v2.3.7

openbrian:
Use the same version as seen in Gemfile.lock to be consistent.


tomhughes:
There's really no point - that version will change routinely depending on what version I happen to have installed when I do an update and we will never remember to keep the Dockerfile the same.

PR #3742, opened 2022-10-08
Communities membership

openbrian:
Stage 2 of the communities PRs.  This PR adds user membership of communities.


danieldegroot2:
For failure messages, I would suggest to explicitly say it failed to do a given action. (see also other failure messages on the website)

> Failed to [verb of action] [subject]. [Optional sentence to give a reason for the failure]

Alternatively, present the failure in past tense
> [subject] could not be [verb of action in past tense]. [Optional sentence to give a reason for the failure]

If it 'did not do [something] succesfully', it may imply it did do something, but it did it improperly.

openbrian:
@danieldegroot2 I applied your suggestion.  I'll continue this pattern on the later stages.

gravitystorm:
I'm marking this as a draft since it depends on #3717 being reviewed and merged first.

PR #3745, opened 2022-10-08
Refer to the receiver in wrong user notification

danieldegroot2:
You -the receiver- "have _been_ asked" to read something, not someone else ("have asked").
(assuming this string is still in use)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1693,7 +1693,7 @@ en:
       no_sent_messages_html: "You have no sent messages yet. Why not get in touch with some of the %{people_mapping_nearby_link}?"
       people_mapping_nearby: "people mapping nearby"
     reply:
-      wrong_user: "You are logged in as `%{user}' but the message you have asked to reply to was not sent to that user. Please login as the correct user in order to reply."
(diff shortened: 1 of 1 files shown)


mmd-osm:
I think your fix is not correct. The scenario in this case is  that you as a user have received some notification from another user, and you're now asking the system to process your reply to that notification.

However, as you're not logged on as the original recipient of the notification, the system rejects your "ask" to reply to another user. You can't reply on behalf of the original recipient of a message.

danieldegroot2:
> I think your fix is not correct. The scenario in this case is that you as a user have received some notification from another user, and you're now asking the system to **process your reply** to that notification.

How about the second one? You can't be asking the system to solely read the message if it were a human, right? (Besides, it would mean these system messages were formulated rather oddly.)

tomhughes:
As @mmd-osm says the original message here is correct.

That text appears when you click on a read or reply link in an email but you are not logged in as the user the email was sent to so it is you that has asked to access the message (by clicking on the link).

danieldegroot2:
I see, sorry. I would suggest to say "the [subject] you would like to [verb of action] was not [...]", but that's a minor annoyance.

PR #3746, opened 2022-10-09
Display icon for current note only if this icon is defined

AntonKhorev:
An alternative to displaying some other icon instead of the missing "hidden" icon. Don't display any icon for hidden and other unknown note states, only show a "halo" circle marker.

Previous version with "closed" icon: https://github.com/openstreetmap/openstreetmap-website/pull/3718#discussion_r981484149

Fixes https://github.com/openstreetmap/openstreetmap-website/issues/1215 - when you have note layer enabled you can click on other notes after hiding a note.

Fixes https://github.com/openstreetmap/openstreetmap-website/issues/546 - although this bug doesn't fully manifests itself in current website versions. Before the fix clicking "history" works, but it's a full page reload. What is supposed to happen is a partial reload of the sidebar.

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -261,11 +261,13 @@ L.OSM.Map = L.Map.extend({
 
       L.circleMarker(object.latLng, haloStyle).addTo(this._objectLayer);
 
-      L.marker(object.latLng, {
-        icon: object.icon,
-        opacity: 1,
-        interactive: true
-      }).addTo(this._objectLayer);
+      if (object.icon) {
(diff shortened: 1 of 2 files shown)


tomhughes:
I don't have a particular problem with having an icon for hidden notes - my comment on your original PR was just asking when that could actually happen but you've at least partly answered that here by reference to what happens when you hide a note.

Thinking about it I imagine there's also the case where you visit a note URL from outside after that note has been hidden so maybe we should have an icon for it?

tomhughes:
Ah but hidden notes are invisible to normal users anyway so this really only matters to a handful of moderators so not worth spending huge effort on.

PR #3751, opened 2022-10-10
Communities events

openbrian:
Stage 3 - events and event organizers

See https://wiki.openstreetmap.org/wiki/Communities

This PR is against osm/master, so it's cumulative.  If you want to see the delta from stage 2 (membership) to stage 3 (events) see this PR https://github.com/openbrian/osm-microcosms/pull/204


gravitystorm:
I'm marking this as a draft since it depends on other PRs being reviewed and merged first.

PR #3752, opened 2022-10-12
Removed CSS which causes backwards rtl heading

bgo-eiu:
This problem is on the register page. Compare the heading, which is shown backwards,
‮,like this
‬‎to the top left button which is shown in the right direction.

Currently:
image

With removal of this CSS rule:
image

The only other thing this seems to change is the direction Earth guy is pointing, but this does not really matter

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1089,10 +1089,6 @@ tr.turn:hover {
   }
 }
 
-[dir=rtl] .header-illustration {
-  transform: scaleX(-1);
-}
-
 #content.maximised {
   top: 0;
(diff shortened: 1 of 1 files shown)


tomhughes:
Changing the direction of the image (and not breaking the "arm" part of it) is the whole point of that rule.

Surely the solution is to fix the rule to only apply to the image and not to the text?

tomhughes:
Ah the problem is that the image is a background image and the text is a descendant of that element, so we actually need a rule to flip the text back again...

bgo-eiu:
Oh, to be honest I thought the arm was supposed to be detached. Thanks for the fix!

PR #3753, opened 2022-10-14
Add OpenMapTiles vector layer

zdila:
This pull request adds OpenMapTiles vector layer among the featured layers. To ensure compatibility with the current stack based on Leaflet, it uses [maplibre-gl-leaflet](https://github.com/maplibre/maplibre-gl-leaflet) library.

The main advantage for users is support for labels in multiple languages: users can easily switch by changing language preferences in settings.


tomhughes:
What happened to all the discussions we had at SOTM about how to do this correctly?

That is to say first do a PR to add the technology for vector layers, then follow the procedure at https://wiki.openstreetmap.org/wiki/Featured_tile_layers to get OWG to approve a particular layer, then submit a PR to add that layer.

Also this PR looks incredibly noisy - it can't possibly reasonably need that number of commits to do this.

As it stands I'm afraid this is not mergeable and needs to redone as outlined above.

zdila:
@tomhughes  I was expecting it to be squash-merged which would got rid of all the commit noise.

zdila:
Thanks for the explanation.

Firefishy:
@zdila Great to see this starting out. If you'd like to discuss with the team, a good place is https://webchat.oftc.net/ in channel `#osm-dev`. Answers can be slow, so hang around.

PR #3755, opened 2022-10-17
Fix malformed tags table in changeset atom feed

malteger:
Prior to this change, the feed contained XHTML like
```xhtml
              <table cellpadding="0">
                <tr>
                  <td/>
changesets_count = 1                </tr>
                <tr>
                  <td/>
comment = Test                </tr>
                <tr>
                  <td/>
created_by = iD 2.22.0                </tr>
                <tr>
                  <td/>
host = http://localhost:3000/edit                </tr>
                <tr>
                  <td/>
imagery_used = Bing Maps Aerial                </tr>
                <tr>
                  <td/>
locale = en-US                </tr>
              </table>
```

after this change this is correctly rendered as
```xhtml
<table cellpadding="0">
  <tr>
    <td>changesets_count = 1</td>
  </tr>
  <tr>
    <td>comment = Test</td>
  </tr>
  <tr>
    <td>created_by = iD 2.22.0</td>
  </tr>
  <tr>
    <td>host = http://localhost:3000/edit</td>
  </tr>
  <tr>
    <td>imagery_used = Bing Maps Aerial</td>
  </tr>
  <tr>
    <td>locale = en-US</td>
  </tr>
</table>
```

### DIFF
--- a/app/views/changesets/index.atom.builder
+++ b/app/views/changesets/index.atom.builder
@@ -62,7 +62,7 @@ atom_feed(:language => I18n.locale, :schema_date => 2009,
                 td.table :cellpadding => "0" do |tag_table|
                   changeset.tags.sort.each do |tag|
                     tag_table.tr do |tag_tr|
-                      tag_tr.td << "#{tag[0]} = #{linkify(tag[1])}"
+                      tag_tr.td "#{tag[0]} = #{linkify(tag[1])}"
                     end
                   end
                 end


tomhughes:
It would be good to extend [`check_feed_result`](https://github.com/openstreetmap/openstreetmap-website/blob/915aeccabb77c6563f755b0504ae1f1743511fa4/test/controllers/changesets_controller_test.rb#L323) to check for the correct formatting...

I think when that is called there should always be a changeset in the feed with a `website=http://example.com/` tag.

malteger:
@tomhughes I just tried to write a test for this, but I am unable to access any elements in the feeds `content` tag using an `assert_select` selector. Just checking for the presence of the `website=http://example.com/` tag-text is not achieving this result, as the problem was the malformed XHTML. I would suspect that content of the `content` tag is parsed and stripped of HTML tags or something like that.

As I have no real ruby knowledge I am not able to dig into this further.

tomhughes:
It was tricky it turns out because the HTML is in a different namespace but the feed wasn't actually declaring a name for that namespace... I managed to do something in https://github.com/openstreetmap/openstreetmap-website/commit/140b296f59c1dbdb5747f40ec2dc38f1402d9a01 though.

malteger:
Thanks a lot!

PR #3764, opened 2022-10-22
another trailing space

openbrian:
My editor does this too.  There's one more trailing space in this area of the file.


gravitystorm:
Thank you! I've no idea how I missed that.

PR #3767, opened 2022-10-25
Don't display error messages when second scheduled rc command fails

AntonKhorev:
If we care about errors when using editors like Potlatch,
rc note loading gives an error in Potlatch anyway because Potlatch doesn't have the `import` command. In case of Potlatch there's no way to check what commands are supported. Potlatch won't even tell that it's Potlatch, unlike JOSM that sends this header: `Server: JOSM RemoteControl`. And we can't assume that it's Potlatch when there's no server header because we don't get to read the response as long as we make cors-avoiding requests (https://github.com/openstreetmap/openstreetmap-website/pull/3760).

The error message that is displayed when the note import fails is not helpful. It tells the user to check if the editor is launched and rc enabled. But we already know that it's the case because otherwise the first (`load_and_zoom`) command would have failed. It's better not to display any message at all in this case.

We can't know if `load_and_zoom` fully worked either. For example, on element pages `load_and_zoom` gets a parameter to select the element. This parameter is ignored by Potlatch and there's no error message about it. It makes sense then not to display the message about a failed note load.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -236,20 +236,20 @@ $(document).ready(function () {
           });
 
     if (object && object.type !== "note") query.set("select", object.type + object.id); // can't select notes
-    sendRemoteEditCommand(remoteEditHost + "/load_and_zoom?" + query, function () {
-      if (object && object.type === "note") {
-        const noteQuery = new URLSearchParams({ url: osmHost + OSM.apiUrl(object) });
-        sendRemoteEditCommand(remoteEditHost + "/import?" + noteQuery);
-      }
-    }, function () {
-      // eslint-disable-next-line no-alert
-      alert(I18n.t("site.index.remote_failed"));
-    });
+    sendRemoteEditCommand(remoteEditHost + "/load_and_zoom?" + query)
+      .then(() => {
+        if (object && object.type === "note") {
+          const noteQuery = new URLSearchParams({ url: osmHost + OSM.apiUrl(object) });
+          sendRemoteEditCommand(remoteEditHost + "/import?" + noteQuery);
+        }
+      })
+      .catch(() => {
+        // eslint-disable-next-line no-alert
+        alert(I18n.t("site.index.remote_failed"));
+      });
 
-    function sendRemoteEditCommand(url, callback, errorCallback) {
-      fetch(url, { mode: "no-cors", signal: AbortSignal.timeout(5000) })
-        .then(callback)
-        .catch(errorCallback);
+    function sendRemoteEditCommand(url) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/index.js
```diff
@@ -254,25 +254,27 @@ $(document).ready(function () {
...
+          alert(I18n.t("site.index.remote_failed"));
+        }
         iframe.remove();
       }, 5000);
 
       iframe
         .hide()
         .appendTo("body")
-        .attr("src", url)
+        .attr("src", url) // Open url without having to worry about CORS because we never read the response. This is for editors like Potlatch that don't send Access-Control-Allow-Origin header.
```
tomhughes, line 277: This comment doesn't seem to have anything to do with this PR and also creates an extremely long line.

app/assets/javascripts/index.js
```diff
@@ -254,25 +254,27 @@ $(document).ready(function () {
...
         };
 
-    if (object && object.type !== "note") query.select = object.type + object.id; // can't select notes
-    sendRemoteEditCommand(remoteEditHost + "/load_and_zoom?" + Qs.stringify(query), function () {
+    if (object && object.type !== "note") query.select = object.type + object.id; // It's not possible to select notes with this parameter.
+    sendRemoteEditCommand(remoteEditHost + "/load_and_zoom?" + Qs.stringify(query), true, function () {
       if (object && object.type === "note") {
         var noteQuery = { url: osmHost + OSM.apiUrl(object) };
-        sendRemoteEditCommand(remoteEditHost + "/import?" + Qs.stringify(noteQuery));
+        sendRemoteEditCommand(remoteEditHost + "/import?" + Qs.stringify(noteQuery), false); // This is an enhancement, don't display an error if it fails.
```
tomhughes, line 261: There's no need for the first part of this comment - the part after the comma is the only bit that is relevant. I'd also probably say `on failure` rather than `if it fails` because it avoids questions of what exactly `it` refers to.

app/assets/javascripts/index.js
tomhughes: Well if you really feel it's necessary, but as we're working to replace that code with XHR anyway I wouldn't bother.

AntonKhorev: I usually don't care how long the line with comment is.

mmd-osm:
Potlatch 3 currently has a "market share" of 0.15% (or 328 users) in 2022 so far. We don't see many users complaining about this issue (or the one mentioned in #3760). 

Since this issue is not impacting functionality (you can still edit using remote control), I would label it as a "nice to have" with "low prio". Even closing it would be perfectly fine, I believe.

tomhughes:
Looks good to me, thanks.

Issue #2973, opened 2020-11-22
Local site navigation


lectrician1:
**Initial Problem**
Currently, the Query features tool/page has no ability to navigate back to the Query Features list without re-requesting the site/Overpass-API. 

**Example**
![Untitled_ Nov 21, 2020 7_32 PM](https://user-images.githubusercontent.com/22477298/99891575-fecf8f80-2c30-11eb-9d46-eef248cf74f5.gif)

**Explanation**
Here, I query a point and click on an element. Then, I have to use the back-button for my browser to go back to the query list. Of course, this re-requests osm.org and in-turn, Overpass.

**Reasoning & Request**
Sometimes I like to look at and browse a lot of elements (especially in element-dense cities) using the query tool. 
After I look at an element, I want to be able to go back to the link fast, without having to wait for the web page to reload.
A dedicated back button that leads back to the Query page would do this.

**Other Uses**
Other left-pane-using pages on the site could also use this functionality.
For example, after viewing a changed element on a Changeset page, you should be able to easily go back to the Changeset you were viewing without having to use the back button for your browser.

A forward button could also be implemented.

tomhughes:
So basically you want us to cache all the search results and implement the navigation ourselves with a full history inside javascript rather than leveraging the browser history as we do now?

It think that's pretty unlikely - it sounds like it would be a huge amount of work.

I mean if somebody wants to write it we could review it but I worry that it would be so large and complex that I would be concerned about merging it.

gravitystorm:
> javascript rather than leveraging the browser history as we do now?

The problem is that we don't cache the results in the browser history at all, as @lectrician1 demonstrates (and as I found recently too). When you hit the back button, it makes a second overpass request.

Personally I think the solution is to reduce the javascript in the first place, and use more server-side stuff. I think the results caching would work better if the call to overpass was moved server-side, and the browser is just requesting page fragments from rails, rather than talking to overpass directly. That way, the html is all cached normally in the browser, since it's not dependent on any javascript re-running after hitting the back button.

tomhughes:
Well if you're lucky that might help but @lectrician1 is really asking for the opposite, to move everything client side so the client can keep a history and go back and forth without making calls to any servers.

mmd-osm:
Server side rendering might be quite challenging due to the IP-based rate limiting on the overpass instance.

1ec5:
[...]

PR #3769, opened 2022-10-26
Merge changeset <ol>s on Load more

AntonKhorev:
Since the pull request https://github.com/openstreetmap/openstreetmap-website/pull/3682 is not going anywhere I'll start taking it apart. Here are the changes in https://github.com/openstreetmap/openstreetmap-website/pull/3682/commits/10319cac0d54ddb9e8f1f467cf16928646474730.

On any history page that displays a list of changesets the "Load more" button loads more changesets. This button creates another list element for those changesets. You get as many lists as the number of times you pressed the button plus the original list. There's a line between list items but no line between lists. To the user it looks like some dividing lines are missing. Here you can see that there's no line between the second and the third changesets:
history-ols

This PR makes "Load more" to merge the changesets it loads into the original html list so there's only one list.

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -43,6 +43,18 @@ OSM.History = function (map) {
     $("#changeset_" + id).find("a.changeset_id").simulate("click", e);
   }
 
+  function displayFirstChangesets(html) {
+    $("#sidebar_content .changesets").html(html);
+  }
+
+  function displayMoreChangesets(html) {
+    $("#sidebar_content .changeset_more").replaceWith(html);
+    var oldList = $("#sidebar_content .changesets ol").first();
+    var newList = oldList.next("ol");
+    newList.children().appendTo(oldList);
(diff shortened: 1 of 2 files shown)


tomhughes:
Just because we haven't had time to look at something yet doesn't mean it's "not going anywhere" you know, but breaking things that make sense on their own out into separate PRs can certainly help as it makes them easier to review.

tomhughes:
More important by the way is that you've provided a clear explanation of the motivation behind this part of the change, something that wasn't mentioned at all in the original PR by the looks of it.

AntonKhorev:
> Just because we haven't had time to look at something yet doesn't mean it's "not going anywhere" you know

I need some kind of caching of changeset lists and you have different opinions about caching:

- https://github.com/openstreetmap/operations/issues/306#issuecomment-502488608
- https://github.com/openstreetmap/openstreetmap-website/issues/2973#issuecomment-731733501

PR #3773, opened 2022-10-27
Introduce sidebar content get/set functions

AntonKhorev:
Part of https://github.com/openstreetmap/openstreetmap-website/pull/3682/commits/c21b5eef3db4ed2ac95e7f5e3e417f27a2672cb1.
OSM.getSidebarContent() is used in later commits of https://github.com/openstreetmap/openstreetmap-website/pull/3682 to store changeset lists.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -34,6 +34,35 @@ $(document).ready(function () {
     worldCopyJump: true
   });
 
+  OSM.setSidebarContent = function (sidebarContent) {
+    var title = sidebarContent[0];
+    var html = sidebarContent[1];
+
+    $("#flash").empty();
+
+    var content = $(html);
+
+    if (title) {
+      document.title = decodeURIComponent(title);
+    }
+
+    $("head")
+      .find("link[type=\"application/atom+xml\"]")
+      .remove();
+
+    $("head")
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/index.js
```diff
@@ -61,25 +90,8 @@ $(document).ready(function () {
...
-        $("head")
-          .find("link[type=\"application/atom+xml\"]")
-          .remove();
-
-        $("head")
-          .append(content.filter("link[type=\"application/atom+xml\"]"));
-
-        $("#sidebar_content").html(content.not("link[type=\"application/atom+xml\"]"));
-
+        OSM.setSidebarContent([xhr.getResponseHeader("X-Page-Title"), xhr.responseText]);
```
tomhughes, line 94: What's the idea between packing these arguments into an array and then unpacking them inside the function rather than just passing two arguments?

Is it so you can pass the array returned by the get function into the set function in a later patch?

AntonKhorev: The idea is for the get function to be able to receive as an argument whatever was returned by the set function.

Get() needs to return two things. They need to be packed into something before being returned. Currently it's easy to imagine how you can unpack them and pass as individual arguments to set(). What if I add more things? What if I decide that with more things it's better to return an object from get()? I don't want to unpack such objects for set() arguments.

PR #3774, opened 2022-10-28
Introduce "reload" js controller method

AntonKhorev:
Part of https://github.com/openstreetmap/openstreetmap-website/pull/3682/commits/2d1862120c1b7c18ce9493139fb874a6ef08a0e1.

This is a method that gets called
- for controllers that have it defined
- when the "Reload current page" button is pressed in the browser
- if it's possible to detect that this button is pressed

Otherwise the usual "load" method is called.

If any kind of client-side caching is used, js controllers can flush the cache in this method.

### DIFF
--- a/app/assets/javascripts/router.js
+++ b/app/assets/javascripts/router.js
@@ -18,6 +18,10 @@
        as arguments the URL path plus any matching arguments for placeholders
        in the path template.
 
+     * The `reload` method, when defined, is called by the router when a page
+       reload is requested by the user. When not defined, the `load` method is
+       called instead.
+
      * The `pushstate` method is called when a page which matches the route's path
        template is loaded via pushState. It is passed the same arguments as `load`.
 
@@ -79,6 +83,10 @@ OSM.Router = function (map, rts) {
       params = params.concat(Array.prototype.slice.call(arguments, 2));
 
       return (controller[action] || $.noop).apply(controller, params);
+    };
+
+    route.has = function (action) {
+      return action in controller;
     };
 
     return route;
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/router.js
```diff
@@ -179,8 +187,20 @@ OSM.Router = function (map, rts) {
   };
 
   router.load = function () {
-    var loadState = currentRoute.run("load", currentPath);
+    var loadState = loadOrReload();
```
tomhughes, line 190: What's the reason for putting this logic in a nested function if this is the only user?

app/assets/javascripts/router.js
```diff
@@ -179,8 +187,20 @@ OSM.Router = function (map, rts) {
...
     router.stateChange(loadState || {});
+
+    function loadOrReload() {
+      if (currentRoute.has("reload") && window.performance) {
+        var wasReloaded = window.performance.getEntriesByType("navigation").some(function (entry) {
+          return entry.type === "reload";
+        });
```
tomhughes, line 197: I'm not familiar with `window.performance` but from a quick bit of reading on MDN it seems this if testing if there are any navigation events of type `reload` in the history but what's the scope of that? The comment you added says we're interest in whether this request was a reload?

app/assets/javascripts/router.js
AntonKhorev: Are you asking why it's nested or why it's defined when there's only one caller?

The short answer is to initialize var loadState when it's declared.

tomhughes: I'm asking why put the logic in a function rather than just having it inline.

Are you saying that function is only called once rather than each time the load function is invoked?

app/assets/javascripts/router.js
AntonKhorev: I think it's for the current page only, not for the history.

https://developer.mozilla.org/en-US/docs/Web/API/Performance : "The Performance interface provides access to performance-related information for the current page".

app/assets/javascripts/router.js
AntonKhorev: Of course loadOrReload is called as many times as router.load.

AntonKhorev: Changed it to extra variable instead of extra function.

PR #3775, opened 2022-10-29
Have only one changeset link in chageset list, give it a Bootstrap button outline

AntonKhorev:
History pages have two links to every changeset: one from a comment and one form an id. Why? Maybe because there's a chance that user wrote a comment that is difficult to click, but there's no way to make an unclickable chageset id. However that's no longer relevant because the link is stretched to the entire box with the changeset info.

Second link is in the way of keyboard navigation. When tabbing through the changesets you get from the first link to the second one leading to the same place. On user's changesets page you have to press tab twice to get to the next changeset.

Here I removed the second link. Also I replaced the standard link outline that the comment has when it's keyboard-focused with a Bootstrap outline* for inputs covering the entire stretched link.

\* actually it's a box-shadow and I made it inset to keep it inside the comment box

changeset-outline

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -817,6 +817,14 @@ tr.turn:hover {
       position: relative;
       z-index: 2; /* needs to be higher than Bootstrap's stretched link ::after z-index */
     }
+
+    a.stretched-link:focus-visible {
+      outline: none;
+    }
+
+    a.stretched-link:focus-visible::after {
+      box-shadow: inset $input-btn-focus-box-shadow;
+    }
   }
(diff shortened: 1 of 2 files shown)


tomhughes:
To add a bit of context the two links being talked about here are the fact that the whole block containing the comment is clickable (that isn't strictly a link as it works via a click handler on the block) and the actual link as an A element inside the block which is redundant.

I'm not going to merge this right now as I'd like @gravitystorm to look at the style parts of it.

AntonKhorev:
Also it seems that with two links the click event handler gets called twice under some circumstances.

PR #3779, opened 2022-10-31
Replace searching.gif images with Bootstrap spinners

AntonKhorev:
https://getbootstrap.com/docs/5.1/components/spinners/

One problem with the current "loading" image is that it has a long url that may change:
https://www.openstreetmap.org/assets/searching-3343ef26a638a661762242d1f68b505ce78f59cf7a60d15a1ca81bd64da1166c.gif

This is inconvenient if I start caching html like I want to do in https://github.com/openstreetmap/openstreetmap-website/pull/3682. I have a [special fix](https://github.com/openstreetmap/openstreetmap-website/pull/3682/commits/2b5d1efa04db45f634a753fd29bc3a46ab8efe72) for this which I wouldn't need with Bootstrap spinners because they don't have any links to assets in their html.

Later I want to remove spinners from html templates and write them with javascript. It's already happening to various degrees in different places of the code.


gravitystorm:
Merged, thank you!

I'm not sure I like the sound of this bit:

> Later I want to remove spinners from html templates and write them with javascript. It's already happening to various degrees in different places of the code.

I think we should be moving more of our html out of javascript and into html, not vice versa! But that's something we can discuss later.

AntonKhorev:
> I think we should be moving more of our html out of javascript and into html, not vice versa! But that's something we can discuss later.

Spinners don't constitute any meaningful html aside from indicating that some js-initiated request has started. And after that they are cleaned up by javascript. Why should they be written outside of javascript?

AntonKhorev:
> > Why should they be written outside of javascript?
> 
> Because there is html involved, and there are CSS style rules applied to that html. So if I want to make changes to that html, or the layout of where the spinner appears,

You know that one of the various methods of placing these spinners is having a hidden spinner stashed somewhere then copying it to the place where it needs to appear with javascript. You are not controlling *where* with html in this case.

> or add an extra div or add padding or anything like that, then I should be looking in the `app/views` directory to find the html code and css classes.

Make a spinner template, embed it in `osm.js.erb`, use it when you need a spinner to appear. This file already had some spinner-related code (https://github.com/openstreetmap/openstreetmap-website/commit/8bc7793d4233b9836e0d3513583ec76762128dd9).

Here's the template:

```
  <div class="text-center">
    <div class="spinner-border" role="status">
      <span class="visually-hidden"><%= t("browse.start_rjs.loading") %></span>
    </div>
  </div>
```

> We can by all means use JavaScript to implement the interactivity, but we shouldn't be using it for constructing html layouts or managing CSS classes.

Are you saying that we shouldn't remove spinners with javascript? Or removing them is ok because it's somehow not constructing/managing but placing them is not ok?

Continuing https://github.com/openstreetmap/openstreetmap-website/issues/647#issuecomment-1308694592, where would you put loading indicators on a page that works without javascript? Where would you put loading indicators on pages that work with turbo?

PR #3781, opened 2022-11-02
Enable the Action Order cop for remaining controllers

gravitystorm:
Where actions were reordered, the rails standard actions were also moved to the top of each controller.

My personal preference is for index/new/create/edit/update/destroy (i.e. pair off new/create and edit/update) but I guess it's better to follow the default order, which is based on the rails scaffolding order too.

I would be delighted at some point if we get to the point of having *only* these actions, but that's work for another day. :smiley: 

I've also fixed a comment (by deleting most of it) that I noticed while reordering actions - it referred to the old approach of having multi-purpose actions which was refactored a while ago.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -112,27 +112,6 @@ Rails/ActionControllerFlashBeforeRender:
     - 'app/controllers/user_blocks_controller.rb'
     - 'app/controllers/users_controller.rb'
 
-# Offense count: 18
-# This cop supports safe autocorrection (--autocorrect).
-# Configuration parameters: ExpectedOrder, Include.
-# ExpectedOrder: index, show, new, edit, create, update, destroy
-# Include: app/controllers/**/*.rb
-Rails/ActionOrder:
-  Exclude:
-    - 'app/controllers/api/changesets_controller.rb'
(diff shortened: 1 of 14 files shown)


tomhughes:
The only reason I had left those disabled instead of letting auto correct fix them was that it didn't manage to move the comments so needed a lot of manual intervention that I didn't have time for.

The flash one I've left disabled because it's very buggy and all of them are actually false positives...

gravitystorm:
> it didn't manage to move the comments so needed a lot of manual intervention that I didn't have time for.

Yep. Doing it manually also let me rearrange the non-restful ones too.



> The flash one I've left disabled because it's very buggy and all of them are actually false positives...

Fixed upstream by https://github.com/rubocop/rubocop-rails/pull/844 and so we just need to wait for the next rubocop-rails release.

tomhughes:
> > The flash one I've left disabled because it's very buggy and all of them are actually false positives...
> 
> Fixed upstream by [rubocop/rubocop-rails#844](https://github.com/rubocop/rubocop-rails/pull/844) and so we just need to wait for the next rubocop-rails release.

That's one of them yes - there are about three or four that I was tracking as I wasn't entirely sure which one(s) we were hitting...

PR #3799, opened 2022-11-16
Relax strict version constraint on rails gem

gravitystorm:
This lets us receive 7.0.x patch releases without having to edit the Gemfile.

I considered making this `~> 7.0` since point releases are usually safe (in particular, if you run `bundle update` and the tests pass, then all good) but there's often a "new framework defaults" file that needs to be generated. So there's an argument that minor 7.x releases should still be treated differently.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 
 # Require rails
-gem "rails", "7.0.4"
+gem "rails", "~> 7.0.0"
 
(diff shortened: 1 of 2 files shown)


tomhughes:
I'm not sure it really gains us anything as I normally see the new release announcement and update it but it doesn't do any harm either though I would have gone with `~> 7.0.4` to preserve the minimum personally.

PR #3802, opened 2022-11-17
Improve install / configuration docs

saerdnaer:
followup of #3719

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -8,7 +8,7 @@ Many settings are available in `config/settings.yml`. You can customize your ins
 
 ## Populating the database
 
-Your installation comes with no geographic data loaded. You can either create new data using one of the editors (iD, JOSM etc) or by loading an OSM extract.
+Your installation comes with no geographic data loaded. You can either create new data using one of the editors (iD, JOSM etc), download data as OSM-XML via JOSM or by loading an OSM extract.
 
 After installing but before creating any users or data, import an extract with [Osmosis](https://wiki.openstreetmap.org/wiki/Osmosis) and the [``--write-apidb``](https://wiki.openstreetmap.org/wiki/Osmosis/Detailed_Usage#--write-apidb_.28--wd.29) task.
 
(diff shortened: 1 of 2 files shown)


CONFIGURE.md
```diff
@@ -18,10 +18,16 @@ osmosis --read-pbf greater-london-latest.osm.pbf \
   user="openstreetmap" password="" validateSchemaVersion="no"
 ```
 
-Loading an apidb database with Osmosis is about **twenty** times slower than loading the equivalent data with osm2pgsql into a rendering database. [``--log-progress``](https://wiki.openstreetmap.org/wiki/Osmosis/Detailed_Usage#--log-progress_.28--lp.29) may be desirable for status updates.
+ ```
+ osmosis --read-xml local_area.osm --write-apidb database=openstreetmap validateSchemaVersion=no
+ ```
+
+Loading an apidb database with Osmosis is about **twenty** times slower than loading the equivalent data with osm2pgsql into a rendering database. [``--log-progress``](https://wiki.openstreetmap.org/wiki/Osmosis/Detailed_Usage#--log-progress_.28--lp.29) may be desirable for status updates. You could improve speed by [temporarily disabling index updating](https://fle.github.io/temporarily-disable-all-indexes-of-a-postgresql-table.html) and reindex all tables after the import.
```
mmd-osm, line 25: Have you tried out to "temporarily disabling index updating" and seen any difference in runtime? If not, I probably wouldn't this kind of recommendation here.

I believe osmosis is doing lots of single inserts and updates rather than bulk operations, which should be the main reason for its poor performance. As osmosis is mostly unmaintained these days, it's fairly unlikely to see a complete rewrite of the importer to speed things up.

gravitystorm:
@saerdnaer Thanks for the revised PR. As I mentioned in the previous PR, and as @mmd-osm has mentioned above, I'm not happy with merging the recommendation to switch off indexing without some evidence that it makes a significant speed up. 

I also notice that your additional command line example for osmosis is very similar to the existing one (swapping a pbf extract import for an xml extract import) while also changing the other parameters, as well as changing the formatting of those parameters. It's not clear to me why this would be useful, or if any of the differences between the two command lines are actually significant.

PR #3810, opened 2022-11-19
Use Puma and File cache on dev env for good performance

yuiseki:
## What

- To speed up performance in the local development environment, I propose the following
  - Using `puma` on `docker compose up`
  - Using File cache on `docker compose up`

## Why

- This pull request proposes two changes
  - The first is related to the caching mechanism, which I think is important for this Rails application
    - I feel there are performance issues, especially when using the iD editor in a development environment
    - When you try to use the iD editor, Rails requests all the nodes, ways and relations in the bbox of the area the browser is displaying to the DB and generates XML
    - These nodes, ways, and relations seem to be properly cached as fragments
    - However, the cache mechanism is not effective in the development environment!
    - For that reason, I would like to enable cache properly in the development environment as well!
  - The second point concerns the number of concurrent connections
    - I think the number of concurrent connections is as important for this Rails application as the caching mechanism
    - When you drag in the iD editor, the browser will abort the request and try a new connection
    - However, the development environment seems to have only one concurrent connection!
    - Therefore, I propose to change it so that requests can be accepted in parallel in the development environment by using puma

## How

- I have changed only about 15 lines
- Moreover, it is clear that they only affect the development environment

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -17,7 +17,7 @@ Rails.application.configure do
   # Enable server timing
   config.server_timing = true
 
-  if ENV['LOG_STDOUT']
+  if ENV['RAILS_LOG_STDOUT']
     config.logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new($stdout))
   end
 
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -22,8 +22,9 @@ Rails.application.configure do
   if Rails.root.join("tmp/caching-dev.txt").exist?
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
+    config.action_controller.page_cache_directory = Rails.public_path
 
-    config.cache_store = :memory_store
+    config.cache_store = :file_store, Rails.root.join('tmp', "cache")
     config.public_file_server.headers = {
       "Cache-Control" => "public, max-age=#{2.days.to_i}"
     }
@@ -32,6 +33,8 @@ Rails.application.configure do
 
     config.cache_store = :null_store
   end
+
+  config.logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT))
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -31,11 +31,6 @@ Rails.application.configure do
     config.public_file_server.headers = {
       "Cache-Control" => "public, max-age=#{2.days.to_i}"
     }
-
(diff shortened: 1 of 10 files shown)


config/environments/development.rb
```diff
@@ -33,6 +34,8 @@
     config.cache_store = :null_store
   end
 
+  config.logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new($stdout))
```
tomhughes, line 37: Your PR does not explain why you are making this change, which affects all users not just docker users.

config/example.database.yml
```diff
@@ -8,6 +8,7 @@ development:
 #  password: openstreetmap
 #  host: localhost
   encoding: utf8
+  pool: <%= ENV["DB_POOL"] || ENV['RAILS_MAX_THREADS'] || 5 %>
```
tomhughes, line 11: Your PR does not explain why you are making this change, which affects all users not just docker users.

config/environments/development.rb
```diff
@@ -22,8 +22,9 @@
   if Rails.root.join("tmp/caching-dev.txt").exist?
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
+    config.action_controller.page_cache_directory = Rails.public_path
```
tomhughes, line 25: Your PR does not explain why you are making this change, which affects all users not just docker users.

config/environments/development.rb
```diff
@@ -22,8 +22,9 @@
...
 
-    config.cache_store = :memory_store
+    config.cache_store = :file_store, Rails.root.join("tmp/cache")
```
tomhughes, line 27: Why is a file cache better than a memory cache? You say it gives better performance but normally memory is faster than than something that is disk backed? Bearing in mind that this affects all users is the performance benefit in fact only when running under docker?

yuiseki:
Thanks for careful reviewing my pull request!

As you pointed out, there were some points that were not explained well enough and had too much impact, so I have put the pull request on draft for now.

I will explain the background and process that led to my decision to make this change and try to limit the impact. Please wait.

gravitystorm:
@yuiseki Please let us know if your PR is now ready for review, or if you are planning to make more changes.

Additionally, please avoid "fixup" (or "revert") commits in your PR. From [CONTRIBUTING.md#pull-requests](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests)

> The commit history is important. This is a large codebase, developed over many years by many developers. We frequently need to read the commit history (e.g. using git blame) to figure out what is going on. So small, understandable, and relevant commits are important for other developers looking back at your work in future.

You currently have 17 commits for this PR, most of which have "revert", "fix" or "tweak" in the title!

PR #3821, opened 2022-11-30
Add more testcases to changesets_controller

mmd-osm:
Extracted from pull request #1995

Closes #1995


mmd-osm:
I have no idea why unit tests are failing with random javascript issues.

Here's what I get on my local machine:

Finished in 67.037668s, 19.0490 runs/s, 9321.1626 assertions/s.
1277 runs, 624869 assertions, 1 failures, 0 errors, 0 skips

(The one failure is due do a failure in application_helper_test.rb:60, which doesn't correctly run on non UTC-0 environments)

Issue #3782, opened 2022-11-03
 OSM map no longer loads in Safari on iOS 9.3.6


saper:
### URL

https://www.openstreetmap.org/

### How to reproduce the issue?

Load openstreetmap.org on something like iPad Mini 1 generation running iOS 9.3.6. This was the last 32-bit release so it cannot be upgraded.

I understand that no one from developers may still have access to something like this, but I am happy to help as this device is still useful during travel.

This broke few months ago, not recently.

### Screenshot(s) or anything else?

The screenshots are exactly the same as on https://github.com/openstreetmap/openstreetmap-website/issues/3387 except that my language is set to Polish ("pl").

I have fired up [Google's ios-webkit-debug-proxy](https://github.com/google/ios-webkit-debug-proxy/releases/download/1.7.1) and it show the first exception is at https://www.openstreetmap.org/assets/i18n/pl-f04b993d4547b4842aeb34cf05df4886ef128649fd5375e4be846a4b4bc67aa7.js line 1 pointing at the dot after first `I18n`:

````js
I18n.translations||...
````

![Developer tools screenshot pointing at the exception in the pl-f04b993d4547b4842aeb34cf05df4886ef128649fd5375e4be846a4b4bc67aa7.js file](https://user-images.githubusercontent.com/248374/199803933-54c4641e-4f64-433c-9eff-1fd1bce2c4ed.png)

Looks like application.js has not been loaded yet, maybe this is a problem with an order of script loading?

pnorman:
What version of Safari do you have?

tomhughes:
Given that the developers here have no access to Safari somebody will need to provide the errors from the console but most likely there will be nothing we can do - it's probably just too old for leaflet and/or bootstrap to support any more.

tomhughes:
It would also help to know when it last worked, but if it was a while ago then this may well be the same issue as Internet Explorer which was triggered by the bootstrap 5 upgrade.

saper:
User-Agent: Mozilla/5.0 (iPad; CPU OS 9_3_6 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13G37 Safari/601.1 

which is Safari 9

I can try to build the openstreetmap UI myself and try to find out.

saper:
> somebody will need to provide the errors from the console

[...]

PR #3828, opened 2022-12-07
Remove fallback code for undefined history.pushState()

AntonKhorev:
There's a check if `window.history.pushState` is defined. If not, there's fallback code that uses `window.location` instead of `window.history`. Given our [current browser support](https://github.com/openstreetmap/openstreetmap-website/issues/3782) the fallback is unnecessary. `window.history.pushState` is supported by [ten years old browsers](https://caniuse.com/mdn-api_history_pushstate).

### DIFF
--- a/app/assets/javascripts/router.js
+++ b/app/assets/javascripts/router.js
@@ -101,53 +101,43 @@ OSM.Router = function (map, rts) {
 
   var router = {};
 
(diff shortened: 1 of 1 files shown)


HolgerJeromin (approved):

saper:
It wasn't my intention with #3782 to define "new browser support" :( quite the contrary

Issue #933, opened 2015-03-18
"Load more" on user's changesets doesn't alter the url (ie doesn't keep state)


emacsen:
When looking at a user's changesets, you can click "Load More" in order to go back further in their history.

If you find the set of changesets you want, you don't appear able to send that URL to someone else, (or refer to later) because clicking the url on the page doesn't change, ie it doesn't reflect the "state" of the page, ie which set of changesets you're looking at- so you're stuck saying "Go here and hit Load More a dozen times".

tomhughes:
Well because of the way it works there is no direct URL which would encode that view, and I would be a little bit concerned about adding one because it would have to allow the loading of an arbitrary (and possibly massive) list of changesets.

emacsen:
Another way to handle this would be for the view to just only include the parameters and then refresh, ie "startAt" "num", so you have to page through them.

polarbearing:
Using classical pagination as discussed in #1089 would solve this issue.

tomhughes:
Except it wouldn't get merged, because it requires every view to count all the matching changesets.

tomhughes:
It's also completely irrelevant - we have working pagination here, the only problem is that we are paginating a partial not the whole page, so more work is needed to track the state of that partial in the URL of the enclosing page.

PR #3829, opened 2022-12-07
Make "Load more" button on history pages work as a link to the next page

AntonKhorev:
"Pagination working properly without javascript" from https://github.com/openstreetmap/openstreetmap-website/issues/647#issuecomment-1298624196.

The link actually exists even without this PR. You can open any changeset list page, middle-click the "Load more" button and you'll get the next page but it won't have the correct layout.

This PR makes the link work as a proper webpage. Also it changes the link from `.../history?max_id=(id)` to `.../history/(id)`. This makes it possible to construct permalinks to some point in time of someone's edit history.

The plan is later to update the id in the url when the sidebar is scrolled, similar to how [the new forum](https://community.openstreetmap.org/) updates its urls with comment numbers. This would roughly do what was requested in https://github.com/openstreetmap/openstreetmap-website/issues/933. Also it would do https://github.com/openstreetmap/openstreetmap-website/issues/647 but without caching.

Smaller parts of this PR:

- [ ] https://github.com/openstreetmap/openstreetmap-website/pull/3834
- [ ] https://github.com/openstreetmap/openstreetmap-website/pull/3835


app/assets/javascripts/index/history.js
```diff
@@ -173,7 +178,14 @@ OSM.History = function (map) {
...
 
-    loadFirstChangesets();
+    var maxId;
+    var paramsIndex = path.indexOf("?");
+    if (paramsIndex >= 0) {
+      var params = Qs.parse(path.substring(paramsIndex + 1));
+      maxId = params.max_id;
+    }
+
+    loadFirstChangesets(maxId);
```
tomhughes: I don't understand the point of this, indeed of any of this commit, when you remove it all again in the next commit?

app/views/changesets/index.html.erb
```diff
@@ -4,7 +4,7 @@
   </ol>
 <% if @changesets.size == 20 -%>
   <div class="changeset_more mt-3 text-center">
-    <%= link_to t(".load_more"), url_for(@params.merge(:max_id => @changesets.last.id - 1)), :class => "btn btn-primary" %>
+    <%= link_to t(".load_more"), "#{url_for}/#{@changesets.last.id - 1}", :class => "btn btn-primary" %>
```
tomhughes, line 7: That way of constructing the URL is not a good idea - doesn't `url_for(:max_id => @changesets.last.id - 1)` work? If not then there will something which does and which doesn't rely on manually gluing the component together like that.

tomhughes (changes requested):

AntonKhorev:
Superseded by #5876

Issue #3819, opened 2022-11-26
Misleading "Click on the map to find nearby features" messages in query features mode


AntonKhorev:
### URL

https://www.openstreetmap.org/query

### How to reproduce the issue?

When the query sidebar is open there's a message "Click on the map to find nearby features". The problem is that often it's displayed when clicking the map won't run any query. In those cases the message is misleading. You can get those cases by doing any of the following:

- opening an [empty query page](https://www.openstreetmap.org/query)
- opening a [complete query page](https://www.openstreetmap.org/query?lat=60&lon=30)
- right-clicking the map and selecting "query features" from the context menu
- clicking the "query features" button on the right side, clicking the map to get the query running then zooming out so the query mode gets disabled
- clicking the "query features" button, clicking the map then clicking the "query features" button again to deactivate the query mode

I made an [attempt of fixing that](https://github.com/openstreetmap/openstreetmap-website/pull/3709) and now it needs updating. But before updating I'd like to know how I'm supposed to fix this problem. Possible solutions are:

- just remove the message
- rewrite the message, maybe telling the user to click "query features" first if it's deactivated
- show/hide the message depending on whether clicking will actually run the query
- same as above but show the message as a tooltip on the "query features" button - this is already done with the "Zoom in to query features" message when the user zooms out too far while the query mode is enabled

lectrician1:
When you're zoomed out and can't query, it's also impossible to close the "Query" sidebar.

I think we should show "Zoom in to query features" as the message until the user zooms to a level at which they can query. Then we should change the message to "Click on the map to find nearby features." and automatically enable the Query tooltip.

AntonKhorev:
> When you're zoomed out and can't query, it's also impossible to close the "Query" sidebar.

Are you saying that the "X" button on the sidebar doesn't work?

> I think we should show "Zoom in to query features" as the message until the user zooms to a level at which they can query. Then we should change the message to "Click on the map to find nearby features." and automatically enable the Query tooltip.

Do you want both messages as tooltips?

AntonKhorev:
> > Are you saying that the "X" button on the sidebar doesn't work?
> 
> Yes.
> 

Does it happen only on this exact page https://www.openstreetmap.org/query without the lat/lon parameters in the url? I'm not sure if this page is supposed to be accessible without the parameters.

lectrician1:
> Does it happen only on this exact page https://www.openstreetmap.org/query without the lat/lon parameters in the url?

[...]

PR #3833, opened 2022-12-08
Don't rely on /query page receiving correct lat/lon parameters

AntonKhorev:
If you open the empty query page https://www.openstreetmap.org/query , you'll get [javascript errors](https://github.com/openstreetmap/openstreetmap-website/issues/3819#issuecomment-1341369521). Maybe it's not supposed to be accessible this way, but there's no 404 for missing parameters. You can also pass invalid lat/lon parameters that Leaflet won't recognize and won't construct a valid LatLng object. Leaflet may fail by throwing an exception or by returning `undefined`.

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -352,8 +352,11 @@ OSM.Query = function (map) {
   };
 
   page.load = function (path, noCentre) {
-    var params = Qs.parse(path.substring(path.indexOf("?") + 1)),
-        latlng = L.latLng(params.lat, params.lon);
+    var params = Qs.parse(path.substring(path.indexOf("?") + 1));
+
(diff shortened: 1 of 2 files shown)


tomhughes:
There's no 404 for missing parameters because we don't expect people to make up URLs and if they do they get to pick up the pieces - is there actually anyway to get the site to refer to that URL other than by manually entering it into the URL bar?

AntonKhorev:
So it's not an error? The same thing is going to happen anywhere with L.latLng() accepting arbitrary input. For example, typing something that is not a number in one of the [export](https://www.openstreetmap.org/export) inputs.

tomhughes:
Sure if there's a way to cause an error by typing something unexpected into an input field on the site then that would be totally different, but I haven't been able to reproduce that - anything typed in those fields is passed to the map call which then returns an error?

PR #3834, opened 2022-12-08
Make max_id part of url path on history pages

AntonKhorev:
Smaller change from https://github.com/openstreetmap/openstreetmap-website/pull/3829. 

Enables links like `/user/username/history/123` where `123` is treated like `max_id` parameter. Since those links are not yet used anywhere, there are no changes that are noticeable during normal website use.

Originally I didn't want to make this url change along with https://github.com/openstreetmap/openstreetmap-website/pull/3829 but it turned out that it was easier to do it rather than to work with `?max_id=123` parameters.


tomhughes:
What is the point of this when you've already included these changes in #3829 and they don't actually do anything useful on their own?

tomhughes:
Partial duplicate of #3829.

PR #3835, opened 2022-12-08
Update place history url test to recognize paths with max_id

AntonKhorev:
A small change from https://github.com/openstreetmap/openstreetmap-website/pull/3829.

Adds a function to check if the currently opened history page is a place history - those are treated specially, by updating the bbox parameters when the map view changes. The condition checked is modified to take into account possible changes from https://github.com/openstreetmap/openstreetmap-website/pull/3834.

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -58,7 +58,7 @@ OSM.History = function (map) {
   function update() {
     var data = { list: "1" };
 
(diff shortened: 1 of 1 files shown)


tomhughes:
Please stop with the duplicate PRs. I plan to review #3829 tomorrow so if you could just wait another day we can all stop wasting our time on these duplicates.

Issue #3843, opened 2022-12-14
Integrate Valhalla Routing (again)


nilsnolde:
### Problem

Implement support for another great FOSS routing engine.

### Description

We've been hosting a public Valhalla instance for the past year, funded by FOSSGIS and maintained by our company. I just got confimation that it'll be funded in the future as well, just like OSRM. The frontend is available here: https://valhalla.openstreetmap.de/.

If you agree, we'd like to PR the implementation in this repo soon-ish (Cc @chrstnbwnkl), so it can be used on osm.org next to Graphhopper & OSRM. @kevinkreiser mentioned that Valhalla was already supported back in the days. It has been added in this commit (https://github.com/openstreetmap/openstreetmap-website/commit/e54eadea95296a68044bd3f2b092b978a808f1b7), but I guess sufficiently many changes went over both Valhalla & the code in this repo since then.

tomhughes:
I don't think much will have changed in this repo though I can't speak for valhalla.

I'd start by reverting 9340c461732ef7387dd453ea3fcf5e64ea782ab4 though as that represents the state when it was removed - obviously it will need renaming as fosgiss_valhalla or something.

Probably we also need to rename the existing fossgis routing references in the code to make it clear they are the osrm version.

nilsnolde:
> I don't think much will have changed in this repo though I can't speak for valhalla.

Yeah, seems very straight-forward. 

> Probably we also need to rename the existing fossgis routing references in the code to make it clear they are the osrm version.

Agreed.

Ok cool, then you'd accept a PR it seems. I'll get the ball rolling in that case.

PR #3848, opened 2022-12-16
Re add valhalla

chrstnbwnkl:
Fixes https://github.com/openstreetmap/openstreetmap-website/issues/3843

Renamed the FOSSGIS routing engine and the `fossgis.js` file, and added the Valhalla routing  engine (based off previous work back from the mapzen days).


systemed:
Had a play around with this and it looks great.

One suggestion: the Graphhopper and OSRM instructions write out an index (in bold) at the start of each step. So:

> **3.** Turn right onto **Enstone Road**	90m

rather than just

> Turn right onto **Enstone Road**	90m

For consistency and for usability it would be good to include this in the Valhalla instructions. (It's particularly useful when discussing a shared route, e.g. "why is Valhalla sending me along Frog Street at step 7 on this route".)

Otherwise looks good to go.

app/assets/javascripts/index/directions/valhalla.js
```diff
@@ -0,0 +1,112 @@
...
+          callback(true);
+        }
+      });
+    }
+  };
+}
+
+  OSM.Directions.addEngine(new ValhallaEngine("valhalla_car", "auto"), true);
+  OSM.Directions.addEngine(new ValhallaEngine("valhalla_bicycle", "bicycle"), true);
+  OSM.Directions.addEngine(new ValhallaEngine("valhalla_foot", "pedestrian"), true);
```
tomhughes, line 112: Can we add a `fossgis_` prefix to the engine names here please?

tomhughes (commented):
My comments are all cosmetic, around making naming consistent.

I'm open to arguments that we shouldn't have the prefix, but we should be consistent in either having it or not having it - whether to have it really comes down to whether we're likely to have multiple instances of OSRM or Valhalla at any point...

chrstnbwnkl:
Thanks for the comments, I added the manoeuvre numbering and renamed the routers!

nilsnolde:
fwiw, I just realized the OSM homepage might be impacted a bit as well: the fossgis valhalla instance is down for a few days, see https://github.com/valhalla/valhalla/issues/6067#issuecomment-4480589599. we had repeated hardware failures and are considering options.

PR #3853, opened 2022-12-21
Wait for the button to be disabled, before checking the tooltip

gravitystorm:
Rather than sleeping for a fixed period, we can use Capybara's built-in ability to wait for elements to be present. Since we're only changing the state of existing elements though, we need to wait for their "disabled" property to be set.

Fixes some timing-related problems with the test suite.


gravitystorm:
Merged as 6f471c72c93d67bf086c18e6a3cc24fd882fbb2c

PR #3854, opened 2022-12-21
Fix some German translations

Nadyita:
There are 2 German translations that stand out at being wrong:
1. Clinic is translated to the same meaning as Hospital. This is definitely wrong, because the correct translation is either "Tagesklinik", "Poliklinik" or "Ärztehaus"
2. The `amenity=ranger_station` is translated to (literally) Rangerstation, when in fact, the Wiki already suggests calling it "visitor facility building", or "Besucherzentrum"/"Besucherstation".

I fixed both and here's my PR for it

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -774,7 +774,7 @@ de:
           charging_station: Ladestation
           childcare: Kinderbetreuung
           cinema: Kino
-          clinic: Krankenhaus
+          clinic: Ärztehaus
           clock: Uhr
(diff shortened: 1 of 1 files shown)


Firefishy:
Translations are not managed here. Please submit the fixes via: https://translatewiki.net/wiki/Translating:OpenStreetMap

tomhughes:
As @Firefishy says these will need to be fixed through TranslateWiki.

PR #3861, opened 2022-12-28
Load the community model when the database is offline

### LINKED ISSUE #3858: OsmCommunityIndex fails when database_offline status is set
I've added an entry `status: "database_offline"` to my settings.local.yml and wanted to start up the Rails server. For some reason, the server wouldn't start up anymore and fails with the following error message:

openstreetmap-website/lib/osm_community_index.rb:4:in `add_to_i18n': uninitialized constant OsmCommunityIndex::Community (NameError)

Maybe this is some issue in my local set up. Once I remove the database_offline setting again, everything is working fine.

gravitystorm:
Normally all the models are loaded automatically, but when the database is offline the app/models directory is ignored. The community model can still be used offline, since it doesn't depend on activerecord.

Fixes #3858

I'm not 100% happy with this, since I would prefer to keep the mechanics of the database_offline and file loading confined to one place (i.e. `config/application.rb`). But I couldn't find a way to deal with the eager_loading on a file-by-file basis, so that we could say "ignore all the files in this directory, but then for this one file..."

```ruby
# Don't eager load models when the database is offline
if Settings.status == "database_offline"
  config.paths["app/models"].skip_eager_load!
  config.paths["app/models/community.rb"].eager_load!  # <- yeah that doesn't work
  config.paths.add("app/models/community.rb", :eager_load => true)# <- yeah that doesn't work either
end
```

I'm also not 100% on the subtleties of "eager load" vs "auto load" here. I would have thought that marking the directory as not eligible for eager loading would mean that in production they aren't loaded during application boot, but were still auto loaded when required, and in any case were auto loaded during development. But the failure to auto-load from app/models when they are marked as "skip_eager_load!" suggests that I don't fully understand how auto_load and eager_load interact.

So based on that, adding the `require_relative` to the code run from the initializer a relatively nonintrusive way to fix this.

### DIFF
--- a/lib/osm_community_index.rb
+++ b/lib/osm_community_index.rb
@@ -1,4 +1,6 @@
 module OsmCommunityIndex
+  require_relative "../app/models/community" if Settings.status == "database_offline"
+
   def self.add_to_i18n
     # Filter the communities here to avoid loading excessive numbers of translations
     communities = Community.where(:type => "osm-lc").where.not(:id => "OSMF")


tomhughes:
I didn't think we even loaded `ActiveModel` in offline mode. My assumption was we would have to just disable community index in offline mode but I only just got back and haven't had time to look yet.

tomhughes:
You can't add an eager load for `app/models/community.rb` because the code which builds the list of eager loads from the path list drops anything which is covered by a parent path.

Fortunately I think I've found a better solution so I've opened #3874 for it.

gravitystorm:
Superseded by #3874

PR #3863, opened 2022-12-28
Update wiki pages

mmd-osm:
script/misc/update-wiki-pages hasn't been run in a while, so here's an updated version. The number of deleted rows seems a bit larger than usual. I can't tell for sure if the script is still working as expected and the results is fine, or if the underlying Wiki representation has been changed in a way that it breaks the update script.

Due to a large number of MediaWiki API calls and resulting issues with rate limiting, breaking network connections, and general stability issues, I've added another commit to define retry and timeout values for the API calls.

### DIFF
--- a/config/wiki_pages.yml
+++ b/config/wiki_pages.yml
@@ -37,11 +37,13 @@ ca:
     bicycle: Ca:Key:bicycle
     board:title: Ca:Key:board:title
     building: Ca:Key:building
+    city_limit: Ca:Key:city limit
     contact:facebook: Ca:Key:contact:facebook
     created_by: Ca:Key:created by
     facebook: Ca:Key:facebook
     lgbtq: Ca:Key:lgbtq
     name: Ca:Key:name
+    name:ca: Ca:Key:name:ca
     openplaques:id: Ca:Key:openplaques:id
     operator:type: Ca:Key:operator:type
     phone: Ca:Key:phone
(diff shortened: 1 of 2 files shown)


matkoniecz:
> The number of deleted rows seems a bit larger than usual.

I spot checked two cases and both were result of new handling of documented prefixes. See https://wiki.openstreetmap.org/w/index.php?title=Cs:Key:contact&action=history and https://wiki.openstreetmap.org/w/index.php?title=DE:Key:contact:phone&action=history

That should result mistaken linking to pages without matching documentation, though currently it seems that redirects pointing to KeyPrefixDescription are being skipped.

I  am anyway confused by https://github.com/openstreetmap/openstreetmap-website/blob/0b76b22a2be65c2579a7eac92b56414f1c40a0b2/script/misc/update-wiki-pages#L58 - have we ever got `Template:CzKeyDescription` or `Template:CzValueDescription` ? (from looking at https://wiki.openstreetmap.org/wiki/Special:Log?type=&user=&page=Template%3ACzValueDescription&wpdate=&tagfilter=&wpfilters%5B%5D=newusers and https://wiki.openstreetmap.org/wiki/Special:Log?type=&user=&page=Template%3ACzKeyDescription&wpdate=&tagfilter=&wpfilters%5B%5D=newusers answer seems "no")

mmd-osm:
`contact:phone` as an example now links to the generic English page  https://wiki.openstreetmap.org/wiki/Key:contact:phone?uselang=de instead of https://wiki.openstreetmap.org/wiki/DE:Key:contact:phone?uselang=de 

image

So it seems we have a bit of a regression here. The perl script fails to  handle KeyPrefixDescription correctly like you've already mentioned.

mmd-osm:
I've added the KeyPrefixDescription template in the Perl script. I think we're seeing fewer deletions now and at the same time much more new tags included in the results.

PR #3868, opened 2023-01-01
fix formatting of tag values with semicolon-separated URLs

k-yle:
If a tag value currently contains two URLs separated by a semicolon, the website creates a broken clickable link. 

[This example node](https://osm.org/node/9335292678) can be viewed on osm.org and in the Washington, D.C. example from [the installation guide](https://github.com/openstreetmap/openstreetmap-website/blob/master/DOCKER.md#loading-an-osm-extract).

<table>
<tr>
<td><strong>Before
<td><strong>After
<tr>
<td>
<td>
</table>

Closes openstreetmap/iD#9444 which was opened in the wrong issue tracker.

p.s. Please reconsider merging #3575, it's impossible for first time contributors to run the tests without digging through the issue tracker to find that rejected PR.


tomhughes:
Can you fix the rubocop warnings that the CI has reported please?

PR #3874, opened 2023-01-05
Allow models to load when the database is offline

### LINKED ISSUE #3858: OsmCommunityIndex fails when database_offline status is set

tomhughes:
This is an alternative to #3861 to fix #3858 and allow offline mode to be used with the community index.

It turns out that while it used to be necessary to stop `ActiveRecord` loading because it would immediately try and connect the database that is no longer the case as the introduction of switchable database connections means it now only connects when a model is used for the first time.

So we can in fact just get rid of a load of nasty code, part of which involves copying a list of loads from the rails code which then gets out of sync with rails over time.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,18 +1,6 @@
 require_relative "boot"
 
-if ENV.fetch("OPENSTREETMAP_STATUS", nil) == "database_offline"
-  require "active_model/railtie"
-  require "active_job/railtie"
-  require "active_storage/engine"
-  require "action_controller/railtie"
-  require "action_mailer/railtie"
-  require "action_view/railtie"
-  require "action_cable/engine"
-  require "sprockets/railtie"
-  require "rails/test_unit/railtie"
-else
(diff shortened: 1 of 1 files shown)


gravitystorm (approved):
Looks good to me! In fact, it looks excellent.

One of my major concerns with how to test offline/readonly mode functionality has been around booting the app into the different modes, or somehow unloading ActiveRecord while running the offline mode tests. But with this change, there's no difference in the booting or the class loading, and so we can write accurate tests much, much more easily.

PR #3878, opened 2023-01-07
Add database checks to issue and report controllers

### LINKED ISSUE #3875: Reporting issues doesn't respect database_readonly mode
### URL

https://www.openstreetmap.org/reports/new?reportable_id=000&reportable_type=DiaryEntry

If you boot the application in database_readonly mode, then links to report issues still appear, and the report issue form is shown and is active. It will then error after submission.

This is because there are no `check_database_writable` checks in reports_controller. There's none in issues_controller either, which is used by moderators for opening/closing/reassinging issues, all of which also require writeable database.

It might be worth going through the other controllers to check that they have appropriate `check_database_writable` and `check_database_readable` checks too.

_No response_

tomhughes:
Fixes #3875

### DIFF
--- a/app/controllers/issue_comments_controller.rb
+++ b/app/controllers/issue_comments_controller.rb
@@ -3,8 +3,11 @@ class IssueCommentsController < ApplicationController
 
   before_action :authorize_web
   before_action :set_locale
+  before_action :check_database_readable
 
   authorize_resource
+
+  before_action :check_database_writable, :only => [:create]
 
   def create
     @issue = Issue.find(params[:issue_id])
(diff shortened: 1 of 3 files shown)


app/controllers/issue_comments_controller.rb
```diff
@@ -3,6 +3,7 @@ class IssueCommentsController < ApplicationController
 
   before_action :authorize_web
   before_action :set_locale
+  before_action :check_database_readable
```
gravitystorm, line 6: I think this needs to be `writeable` for `create` to work

gravitystorm (changes requested):

gravitystorm (approved):

PR #3879, opened 2023-01-07
Make error pages translatable

### LINKED ISSUE #3877: Error pages are not translatable
### URL

https://www.openstreetmap.org/500

The error pages used in the errors controller are not translatable.

They used to be static html pages, before being fully rendered pages when they were moved to the errors controller. But we should move the text to the i18n system.

_No response_

tomhughes:
Fixes #3877

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -573,6 +573,21 @@ en:
       applications:
         create:
           notice: Application Registered.
+  errors:
+    contact:
+      contact_url: https://wiki.openstreetmap.org/wiki/Contact
+      contact_url_title: Various contact channels explained
+      contact: contact
+      contact_the_community_html: Feel free to %{contact_link} the OpenStreetMap community if you have found a broken link / bug. Make a note of the exact URL of your request.
+    forbidden:
+      title: Forbidden
+      description: The operation you requested on the OpenStreetMap server is only available to administrators (HTTP 403)
+    internal_server_error:
+      title: Application error
+      description: The OpenStreetMap server encountered an unexpected condition that prevented it from fulfilling the request (HTTP 500)
+    not_found:
+      title: File not found
+      description: Couldn't find a file/directory/API operation by that name on the OpenStreetMap server (HTTP 404)
   friendships:
     make_friend:
       heading: "Add %{user} as a friend?"
--- a/app/views/errors/_contact.html.erb
+++ b/app/views/errors/_contact.html.erb
@@ -1,0 +1,3 @@
+<p><%= t ".contact_the_community_html",
+         :contact_link => link_to(t(".contact"), t(".contact_url"),
+                                  :title => t(".contact_url_title")) %></p>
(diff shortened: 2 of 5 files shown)


config/locales/en.yml
```diff
@@ -572,6 +572,19 @@ en:
       applications:
         create:
           notice: Application Registered.
+  errors:
+    contact:
+      html: Feel free to <a href="http://wiki.openstreetmap.org/wiki/Contact" title="Various contact channels explained">contact</a> the OpenStreetMap community if you have found a broken link / bug. Make a note of the exact URL of your request.
```
gravitystorm, line 577: We should avoid including raw html tags in translations - these can be hard for translators to get right, among other things. Instead I think it's better to have each component of the paragraph as separate keys, e.g.
```
errors:
  contact:
    contact_url: http:/...
    contact_url_title: Various...
    contact: contact
    contact_the_community_html: Feel free to %{contact_link} the OpenStreetMap ....
```

... and the appropriate `link_to` when calling the translation.

gravitystorm, line 577: `https` for the link too

config/locales/en.yml
```diff
@@ -572,6 +572,19 @@ en:
...
+    forbidden:
+      title: Forbidden
+      paragraph_1: The operation you requested on the OpenStreetMap server is only available to administrators (HTTP 403)
+    internal_server_error:
+      title: Application error
+      paragraph_1: The OpenStreetMap server encountered an unexpected condition that prevented it from fulfilling the request (HTTP 500)
+      paragraph_2: This may be a problem in our Ruby On Rails code. 500 occurs with exceptions thrown outside of an action (like in Dispatcher setups or broken Ruby code)
```
gravitystorm, line 584: I think the second sentence here is jargon that we shouldn't show to end users. I mention it here because I don't know what "Dispatcher setups" are, in this context, and so I wouldn't be able to advise a translator what it means either! :smile:

config/locales/en.yml
```diff
@@ -572,6 +572,19 @@ en:
       applications:
         create:
           notice: Application Registered.
+  errors:
+    contact:
+      html: Feel free to <a href="http://wiki.openstreetmap.org/wiki/Contact" title="Various contact channels explained">contact</a> the OpenStreetMap community if you have found a broken link / bug. Make a note of the exact URL of your request.
+    forbidden:
+      title: Forbidden
+      paragraph_1: The operation you requested on the OpenStreetMap server is only available to administrators (HTTP 403)
```
gravitystorm, line 580: Eh, not strictly true, but whatever.

gravitystorm (changes requested):

config/locales/en.yml
tomhughes: Done.

tomhughes: Done.

config/locales/en.yml
tomhughes: Agreed... It's not really helpful so I've removed it.

tomhughes:
Apologies for merging this - it wasn't my intention and I must have pushed to the wrong place. I assume you're happy with it now I addressed your comments?

gravitystorm:
No worries - I was happy with the updated version!

Issue #1881, opened 2018-05-25
Sort search results from GeoNames


jidanni:
https://www.openstreetmap.org/search?query=南富村

```
Results from GeoNames
Nanfucun , China

Nanfu , Taiwan

Nanfu , Taiwan

Dananbu , Taiwan

Siwan , Taiwan

Shi’ergu , Taiwan

Shalao , Taiwan

Nanfu , Taiwan

Nanfu , Taiwan

Sheliao , Taiwan

Shezijiao , Taiwan

Nanfucun , China

Nanfucun , China

Nanfu , Taiwan

Xiaohushan , Taiwan

Haizikou , Taiwan

Waisiwan , Taiwan

Shatu Shan , Taiwan

Shangpingjiao , Taiwan

Houshixiao , Taiwan
```

1. Please sort the list, any way you like. But at least sort it.
2. No need for space before comma.
3. No need for empty lines after each line
4. Show a third component, when lines are the same. Don't just hope
the user will mouuseover to see the coordinates.

tomhughes:
Frankly we should probably just get rid of the geonames results now.

tomhughes:
There isn't a blank line by the way - there is a thing grey separator. That's the same for all search results and it would be inconsistent to make geonames different.

As to sorting I've always assumed geonames returns them in some sort of order of what it thinks is most likely - this is an unusual case because normally most of the results would start with the string you entered so sort wouldn't do much. I'm not sure how it's chinese search is working that it returns such diverse responses...

gravitystorm:
I'd be happy to see Geonames removed too - OpenStreetMap has good enough placename coverage in most of the world nowadays.

tomhughes:
Geonames support has now been removed by #3886 so this is no longer relevant.

Issue #80, opened 2012-09-04
Write and use jsonp api for nominatim for the rails port


tmcw:
Right now the rails port dispatches requests to nominatim ( https://github.com/openstreetmap/openstreetmap-website/blob/master/app/helpers/geocoder_helper.rb#L24 ) and parses the XML, and then formats. This could be more efficient if the browser hit nominatim directly and got JSONp that could be easily formatted into page content.

PR #3886, opened 2023-01-09
Remove Geonames and geocoder.ca

systemed:
As a historical legacy, we currently return search results from Geonames and geocoder.ca.

We did this in the early days of OSM as a convenience to help people navigate round the map - i.e. if Frogsbury hadn't been mapped yet, it was useful to be able to navigate to Frogsbury and start contributing there. Now that OSM's coverage is so much greater, there's no need for this any more. It's counter-intuitive to return non-OSM data when searching osm.org, and IMX Geonames usually returns "No results found" in any case.

This PR simply removes the support for Geonames and geocoder.ca. The only remaining geocoder engines are Nominatim (forward/reverse) and the simple latlon reader.

Ultimately this is a very small step towards replacing GeocoderController with client-side parsing of Nominatim responses, which would make them easier to use in other contexts (e.g. the placename search for routing), as per #80. #1881 also refers.


gravitystorm (approved):
I'm happy to see this change, so thumbs up from me. I'll also await feedback from @tomhughes in case he's aware of any reason that we should keep these.

gravitystorm:
> Ultimately this is a very small step towards replacing GeocoderController with client-side parsing of Nominatim responses, which would make them easier to use in other contexts

On this topic, I'm not in favour. Moving more things to our home-grown javascript isn't an approach that I support. In particular, we end up creating own own (poor) html templating approaches, a client-side translation system, and this just duplicates what we already have serverside. My preferred approach for these things is to supply the client with "[html over the wire](https://hotwired.dev/)", use rails views to construct the html, and so on.

PR #3890, opened 2023-01-11
Use bin/yarn consistently for installations

### LINKED ISSUE #3579: Yarn install broken - INSTALL.md out of date?
I followed the instructions in INSTALL.md with a fresh virtualbox of ubuntu server 20.04. In the yarn step I get this result:
```
user@hostname:~/openstreetmap-website$ bundle exec rake yarn:install
/usr/lib/ruby/2.7.0/net/protocol.rb:66: warning: already initialized constant Net::ProtocRetryError
/var/lib/gems/2.7.0/gems/net-protocol-0.1.3/lib/net/protocol.rb:68: warning: previous definition of ProtocRetryError was here
/usr/lib/ruby/2.7.0/net/protocol.rb:206: warning: already initialized constant Net::BufferedIO::BUFSIZE
/var/lib/gems/2.7.0/gems/net-protocol-0.1.3/lib/net/protocol.rb:208: warning: previous definition of BUFSIZE was here
/usr/lib/ruby/2.7.0/net/protocol.rb:503: warning: already initialized constant Net::NetPrivate::Socket
/var/lib/gems/2.7.0/gems/net-protocol-0.1.3/lib/net/protocol.rb:504: warning: previous definition of Socket was here
yarn install failed to execute.
Ensure yarn is installed and `yarn --version` runs without errors.
user@hostname:~/openstreetmap-website$
```
I also found this line in the dockerfile that says this is expected to not work. Seems the INSTALL.md needs updating to match the Dockerfile.
https://github.com/openstreetmap/openstreetmap-website/blob/master/Dockerfile#L47

gravitystorm:
Since rails 7, the `rails yarn:install` command no longer uses bin/yarn and so it
- doesn't detect yarn binaries named "yarnpkg" (e.g. Ubuntu), and
- doesn't allow us to pass any flags to yarn, (e.g. "--ignore-engines").

Since we want to avoid making the installation more difficult (by e.g. installing yarn from other sources, or requiring particular node versions) we can instead just use our existing bin/yarn script.

The reason Dockerfile did its own thing was to avoid pulling in big chunks of rails, but bin/yarn is a standalone ruby script.

Fixes #3579

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -54,7 +54,7 @@ jobs:
         restore-keys: |
           yarn-${{ env.os }}-
     - name: Install node modules
-      run: bundle exec rails yarn:install
+      run: bundle exec bin/yarn install
     - name: Create dummy database configuration
       run: cp config/example.database.yml config/database.yml
     - name: Run eslint
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,7 +54,7 @@ jobs:
     - name: Export javascript strings
       run: bundle exec rails i18n:js:export
     - name: Install node modules
-      run: bundle exec rails yarn:install
+      run: bundle exec bin/yarn install
     - name: Compile assets
       run: bundle exec rails assets:precompile
     - name: Run tests
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ ADD Gemfile Gemfile.lock /app/
 RUN gem install bundler \
  && bundle install
 
-# Install NodeJS packages using yarnpkg
-# `bundle exec rails yarn:install` will not work
+# Install NodeJS packages using yarn
 ADD package.json yarn.lock /app/
-RUN yarnpkg --ignore-engines install
+ADD bin/yarn /app/bin/
+RUN bundle exec bin/yarn install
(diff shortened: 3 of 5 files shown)


tomhughes:
It's not really our `bin/yarn` script... I'm pretty sure it came from rails and I just hacked it to add the switch and now rails has stopped using it.

I'm kind of reluctant to switch away from upstream's documented way of doing things, though of course upstream have now purged the whole thing from their memories and in any case it was never clear why you should use rails to do the install when all other manipulation of yarn packages had to be done directly.

gravitystorm:
> It's not really our `bin/yarn` script... I'm pretty sure it came from rails and I just hacked it to add the switch and now rails has stopped using it.

I meant that it's ours now :smile: Yes, it came from upstream rails, but they've stopped creating it for new applications. In different circumstances we would be deleting it as no longer required.

> I'm kind of reluctant to switch away from upstream's documented way of doing things,

So am I, but I think the alternatives are worse.
* If we do `bundle exec rails yarn:install` then we need to require non-Ubuntu packaged yarn, and we need to require everyone installs Nodejs 16+ (for no good reason).
* If we do `yarn install --ignore-engines`, then we still need to explain that it's `yarnpkg install --ignore-engines` on Ubuntu, which is additional complexity. 
* Alternatively we do `bundle exec bin/yarn install` as per this PR.

PR #3903, opened 2023-01-18
Add a "copy coordinates" entry to the right click menu

### LINKED ISSUE #3900: Pushing Cursor Coordinates to Clipboard as Soon as Possible
### Problem

I am using OSM to get cursor coordinates in Clipboard. Doing it quite often takes a lot of **time (Clicks)**.

### Description

Here is OSM Process:

- Right Mouse Click On MAP to open Context Menu,
- The option Show address is given,
- Clicking it opens the Right Panel with Search Results,
- Move the Cursor over the lat/long,
- Click it using the Right Mouse button,
- Context Menu opens,
- Click Copy Link Address.

The link is finally in the Clipboard.

It would be great (and probably not too hard to implement) if Copy Link Address is available in the Context menu after the first Right Mouse Button Click

### Screenshots

_No response_

tomhughes:
Fixes #3900

### DIFF
--- a/app/assets/javascripts/index/contextmenu.js
+++ b/app/assets/javascripts/index/contextmenu.js
@@ -74,6 +74,18 @@ OSM.initializeContextMenu = function (map) {
     }
   });
 
+  map.contextmenu.addItem({
+    text: I18n.t("javascripts.context.copy_coordinates"),
+    callback: function copyCoordinates(e) {
+      var precision = OSM.zoomPrecision(map.getZoom()),
+          latlng = e.latlng.wrap(),
+          lat = latlng.lat.toFixed(precision),
+          lng = latlng.lng.toFixed(precision);
+
+      navigator.clipboard.writeText(lat + ", " + lng);
+    }
+  });
+
   map.on("mousedown", function (e) {
     if (e.originalEvent.shiftKey) map.contextmenu.disable();
     else map.contextmenu.enable();
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3032,6 +3032,7 @@ en:
       show_address: Show address
       query_features: Query features
       centre_map: Centre map here
+      copy_coordinates: Copy coordinates
   redactions:
     edit:
       heading: "Edit Redaction"


gravitystorm:
I've spent far too much time thinking about this, considering the lack of obvious mapper-related use cases! In all my years mapping for OSM I've never needed to copy coordinates while browsing the map, and the original issue gives no indication how this feature would help mappers or fits into any mapping workflow either.

From a technical point of view it's mostly fine, but it reopens the can of worms that is coordinate formatting. I worry that merging this PR as-is might then lead to a succession of debates, hassle and/or followup PRs because:
* The coordinate formatting in this PR doesn't match the rest of the site. JavaScript `toFixed` is not locale-aware, so many users will get decimal points in the clipboard instead of the decimal commas that we show elsewhere. While I don't have a strong view as to which is right, it would be nice to be consistent. See also #2732
* The coordinate delineator is not translatable. See #2783

In saying all that, I don't feel particularly strongly for any of the three outcomes - either to merge as-is, make the PR more complex for a not-important cause, or leave the feature off entirely.

dusko23:
Hi Andy, here is my explanation. I hope you will understand what I am talking about.
I group of enthusiasts created a bespoke windows desktop application for land search and rescue. Detailed topographic maps, detailed aerial images, and detailed Lidar maps rule here.
The application is mission critical and must work without the internet.
However, if the internet is available existing map sources can give some ideas about how mapping at a specific point (zoomed in) looks like. For example, there might be a fork in the path where the person being searched might decide where to go. If the fork is not available on local topographic maps then getting a coordinate on web maps source is important.
In order to implement that feature I have added a Chromium component browser to browse genuine OSM, Google Maps, and Bing Maps. All these maps give the impression of being an integral part of the software giving the user a single point of experience. 
image

image

image

Google maps, Bing maps, and OSM use the right mouse button click to expose the context menu.
Google Maps and Bing Maps have a Context menu option to copy coordinate to Clipboard.
OSM opens the left Panel when Show Address is clicked.
I understand that implementing an option to populate the clipboard when the context menu item is clicked is difficult.
Therefore I have suggested a simple solution to copy Coordinate to the clipboard when the left panel coordinate is clicked.
Or adding Copy option below the coordinate could also be a good solution.
Like this
image

Thinking differently always brings new use cases :-)
I appreciate your work for OSM and I hope you appreciate that there might be new use cases emerging.
That is why I send an initiative to OSM in the first place.

dusko23:
You could just Copy a Link to a clipboard. Exactly the same as OSM is doing now by Context Menu Copy Link Item.
Inside our software, we manage all local issues that might arise from writing and parsing coordinates or links.

tomhughes:
The problem (from out point of view) is that www.openstreetmap.org is not intended as an end user replacement for Google Maps, rather our goal is to support mapping and we leave displaying end user maps to other people, and as such your use is not really in scope.

dusko23:
Hi Tom, I am not talking about any replacement here. Not sure how you got this impression.  I am talking about browsing different web mapping sources in order to spot any differences from local topographic maps. And that is important because users do use different map sources when going out. And one of the SAR Team's missions is to give the best efforts to guess what might have been going on out there.  And that brings OSM maps into an entirely new domain.

pnorman:
I'm often copying coordinates from osm.org for two purposes
- viewing in another map or style development tool. For this, I'm better off copying the URL hash, because that's what the other map normally needs; and
- copying the coordinate of something to compare with another coordinate for debugging

gravitystorm:
Based on the technical headaches that I mentioned in https://github.com/openstreetmap/openstreetmap-website/pull/3903#issuecomment-1403500956 and the lack of use-case, I think we can leave this one. I don't think it's worth the effort to resolve the technical stuff with coordinate formatting, and also maintain and/or have tests, given everything else we have on our plates.

PR #3907, opened 2023-01-22
Facelift `offline.html` and use Bootstrap classes for "notifications" under the search bar

Dimitar5555:
This PR makes the following changes:
- Notices, warnings and error messages which appear under the search bar will use Bootstrap's classes (instead of the custom `flash` classes)
- The `offline.html` page gets a facelift (red background around the message)
- The `edit.html` page also gets facelift although I wasn't able to test it since it redirects me to `/offline/`.

Note that `alert-success` uses `#D4EDDA` while `flash flash-notice` uses `#CBEEA7` for background colour. It w

image
image

### DIFF
--- a/app/views/layouts/_flash.html.erb
+++ b/app/views/layouts/_flash.html.erb
@@ -1,5 +1,5 @@
 <% if flash[:error] %>
-  <div class="flash error row mx-0 p-3 align-items-center">
+  <div class="alert alert-danger row mx-0 mb-0 p-3 rounded-0 align-items-center">
     <div class="col-auto">
       <picture>
         <source srcset="<%= image_path "notice.svg" %>" type="image/svg+xml" />
@@ -11,7 +11,7 @@
 <% end %>
 
 <% if flash[:warning] %>
-  <div class="flash warning row mx-0 p-3 align-items-center">
+  <div class="alert alert-warning row mx-0 mb-0 p-3 rounded-0 align-items-center">
     <div class="col-auto">
       <picture>
         <source srcset="<%= image_path "notice.svg" %>" type="image/svg+xml"></source>
@@ -23,7 +23,7 @@
 <% end %>
 
 <% if flash[:notice] %>
-  <div class="flash notice row mx-0 p-3 align-items-center">
+  <div class="alert alert-success row mx-0 mb-0 p-3 rounded-0 align-items-center">
     <div class="col-auto">
       <picture>
         <source srcset="<%= image_path "notice.svg" %>" type="image/svg+xml"></source>
--- a/app/views/site/edit.html.erb
+++ b/app/views/site/edit.html.erb
@@ -1,8 +1,12 @@
 <% content_for :content do %>
   <% if Settings.status == "database_offline" or Settings.status == "api_offline" %>
-    <p><%= t "layouts.osm_offline" %></p>
+    <div class="alert alert-warning text-center">
+        <p class="my-2"><%= t "layouts.osm_offline" %></p>
+    </div>
   <% elsif Settings.status == "database_readonly" or Settings.status == "api_readonly" %>
-    <p><%= t "layouts.osm_read_only" %></p>
+    <div class="alert alert-warning text-center">
+        <p class="my-2"><%= t "layouts.osm_read_only" %></p>
+    </div>
   <% elsif !current_user.data_public? %>
     <p><%= t ".not_public" %></p>
     <p><%= t ".not_public_description_html", :user_page => (link_to t(".user_page_link"), edit_account_path(:anchor => "public")) %></p>
--- a/app/views/site/offline.html.erb
+++ b/app/views/site/offline.html.erb
@@ -1,7 +1,0 @@
-<% if Settings.status == "database_offline" %>
-<p><%= t "layouts.osm_offline" %>
-</p>
-<% else %>
-<p><%= t "layouts.osm_read_only" %>
-</p>
-<% end %>
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1253,22 +1253,6 @@ tr.turn:hover {
   display: none;
 }
 
-/* Rules for "flash" notice boxes shown at the top of the content area */
-
-.flash {
-  &.error {
-    background-color: #ff7070;
-  }
-
-  &.warning {
-    background-color: #ffe0cc;
-  }
-
-  &.notice {
-    background-color: #CBEEA7;
-  }
-}
-
 /* Rules for highlighting fields with rails validation errors */
 
 .formError {
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -113,7 +113,14 @@ class SiteController < ApplicationController
 
   def export; end
 
-  def offline; end
+  def offline
+    flash.now[:warning] = if Settings.status == "database_offline"
+                            t("layouts.osm_offline")
+                          else
+                            t("layouts.osm_read_only")
+                          end
+    render :html => nil, :layout => true
+  end
 
   def preview
     render :html => RichText.new(params[:type], params[:text]).to_html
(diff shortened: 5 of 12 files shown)


tordans:
Is it possible space out the top and bottom padding of the sentence in the box? Eg. by using the padding/margin utility classes https://getbootstrap.com/docs/5.0/layout/utilities/#margin-and-padding?

Dimitar5555:
It is possible. 

Here is the result with `class="my-3"` on the paragraph.
image

`class="my-2"`
image

app/views/layouts/_flash.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if flash[:error] %>
-  <div class="flash error row mx-0 p-3 align-items-center">
+  <div class="alert alert-danger row mx-0 mb-0 p-3 rounded-0 align-items-center">
```
gravitystorm, line 2: I'm happy to see the move from custom CSS to use the built-in bootstrap alert panels and colours. There's a couple of things here that need addressing though:

* If the `flash` CSS is no longer required elsewhere, then it should be removed from `common.scss`
* Large numbers of tests are failing, since they are looking for the `flash` and `error` css classes. The tests need to be updated.

app/views/site/edit.html.erb
```diff
@@ -1,8 +1,12 @@
 <% content_for :content do %>
   <% if Settings.status == "database_offline" or Settings.status == "api_offline" %>
+    <div class="alert alert-danger text-center">
     <p><%= t "layouts.osm_offline" %></p>
+    </div>
```
gravitystorm, line 5: * The contents of this `div` need to be indented. The same with other places where you are wrapping existing code with additional `divs`.

app/views/site/edit.html.erb
```diff
@@ -1,8 +1,12 @@
...
   <% elsif Settings.status == "database_readonly" or Settings.status == "api_readonly" %>
+    <div class="alert alert-danger text-center">
```
gravitystorm, line 7: Offline and read_only modes are not "danger" messages, since there's nothing that the user can or has done wrong. 

If we look at the `map_layout` helper in https://github.com/openstreetmap/openstreetmap-website/blob/e443d99edd9c9b68d6c0fa41b028bb9a5cdb07c5/app/controllers/application_controller.rb#L258-L263 we can see that both are treated as `warning` messages

app/views/site/offline.html.erb
```diff
@@ -1,7 +1,9 @@
+<div class="alert alert-danger text-center">
```
gravitystorm, line 1: For both the offline and edit pages, we end up with "alert"-style banners, but these don't match the flash banners used on the rest of the site, namely:

* They don't stretch across the whole page
* They don't have the icon alongside the message

It would be nice to make them completely consistent, rather than partially consistent. So for example, I would consider reworking this page to use the existing flash messages system to render the alert, for example by using `flash.now[:warning] = t("layouts.osm_offline")` in the controller. That way any changes made to flash messages elsewhere (e.g. adjusting margins or paddings) will apply automatically to these offline alert messages.

gravitystorm (changes requested):
Thank you for the PR! I've made some requests as inline comments, and I'm happy to discuss them further if required.

app/views/layouts/_flash.html.erb
Dimitar5555: > If the `flash` CSS is no longer required elsewhere, then it should be removed from `common.scss`

There are quite a few tests which require the old classes and they were passing on the first test. Would it be OK if I update all instances of the current `flash` classes to `alert` (as well as their tests) or it would it be better to do it in a separate PR?

> Large numbers of tests are failing, since they are looking for the `flash` and `error` css classes. The tests need to be updated.

I've updated them. There should be only one failing about a `title` on the "Send message" page.

app/views/site/offline.html.erb
Dimitar5555: I'm not exactly sure what exactly needs to be done here. Would this work or the changes need to be made somewhere else?

```
<% if Settings.status == "database_offline" %>
<p><%= flash.now[:warning] = t("layouts.osm_offline") %>
</p>
<% else %>
<p><%= flash.now[:warning] = t("layouts.osm_read_only") %>
</p>
<% end %>
```

app/views/layouts/_flash.html.erb
Dimitar5555: You are right. I was confused since a few tests succeeded even though I didn't update them. They are now fixed.

app/views/site/offline.html.erb
gravitystorm: > Would this work

Something similar to this actually does work - the trick is that you can add flash messages from a view (since it appears the view is processed before the layout) but you need to avoid also writing the message at the same time. So no `<p>` tags and no `<%=` outputs. This works:

```
<% if Settings.status == "database_offline" %>
  <% flash.now[:warning] = t("layouts.osm_offline") %>
<% else %>
  <% flash.now[:warning] = t("layouts.osm_read_only") %>
<% end %>
```

However, I find that a bit unintuitive, since it depends on a timing issue with rendering views vs layouts, and usually flash messages are set in the controller before rendering starts. So I would avoid using `flash.now` in any views. When I said:

> for example by using `flash.now[:warning] = t("layouts.osm_offline")` **in the controller**

... I was thinking about changing the `site#offline` method (in SiteController) something like this:

```diff
-  def offline; end
+  def offline
+    flash.now[:warning] = if Settings.status == "database_offline"
+                            t("layouts.osm_offline")
+                          else
+                            t("layouts.osm_read_only")
+                          end
+  end
```
At this point, the view can be empty, and this will work. Having an empty view is a bit strange, so I would either think of something to put in there (for example, further explanation, but I'm not sure if that is useful) or else remove the view entirely and be explicit in the controller that there's nothing else to render. [I found some examples of how to do this](https://stackoverflow.com/a/45581659/105451), e.g.

```ruby
def offline
  flash.now[...]
  ...
  render :html => nil, :layout => true
end
```

gravitystorm:
Also, a request from [CONTRIBUTING.md](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md) to consider please:

app/views/site/offline.html.erb
Dimitar5555: I did it but it broke that test:
```
  # Test the offline page
  def test_offline
    get offline_path
    assert_response :success
    assert_template "offline"
  end
```
Do I have to delete the test, since I've deleted `offline.html.erb`?

Dimitar5555:
> Also, a request from [CONTRIBUTING.md](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md) to consider please:
> 
> > Avoid including "fixup" commits. If you have added a fixup commit (for example to fix a rubocop warning, or because you changed your own new code) please combine the fixup commit into the commit that introduced the problem. git rebase -i is very useful for this.

I had Git on my PC but it broke after I reinstalled the OS.. I've managed to squash all changes into a single commit on another branch using Github Desktop (and as a side "bonus" I made a mess in this one). Are you able to somehow clean it up? If not, I can close this PR and open a new one.

app/views/site/offline.html.erb
gravitystorm: Keep the test, it's still required. The final assertion no longer makes sense and needs to be adjusted - I would change it to no longer asserting that a particular template is used (since we aren't using a view), but to change it to ensure that the layout is still rendered, since we need the layout to make sure the flash message is displayed. 

There are examples of asserting the layout elsewhere in the test directory.

gravitystorm:
> If not, I can close this PR and open a new one.

You generally don't need to open a new one. If you have two branches (e.g. patch-1 and patch-2) and you have opened a PR using one of them (e.g. patch-1) then you can [rename the branches](https://stackoverflow.com/questions/6591213/how-do-i-rename-a-local-git-branch), e.g.

* rename `patch-1` to somethign like patch-1-backup or patch-3
* rename `patch-2` to `patch-1`
* force-push `patch-1` to github, and it will automatically update this PR

It's possible for maintainers to squash commits and rework commit messages and all that kind of thing, but it helps us a lot if we don't have to do so!

app/views/site/offline.html.erb
Dimitar5555: I've updated the test and it seems to be working.

gravitystorm:
Thanks for making all the requested changes @Dimitar5555 - that was very helpful!

Issue #3831, opened 2022-12-07
What should we do about note comments?


gravitystorm:
This isn't the first issue on notes and note comments, and I doubt it will be the last either!

The original implementation of notes has, for unimportant reasons, left us with the situation where Notes surprisingly don't have an author or body attribute, but instead every Note has a "first comment" that we treat as special. However, this is not how anyone really thinks of a note - we all think that the note belongs to someone, and has some body text associated with the note itself. The implementation that we have though is as if a note is created out of thin air, with no text and nobody associated to it, whenever someone makes that first comment. This conceptual mismatch, or this leaky abstraction, I think has led to more bugs and errors than any other part of the codebase, and is confusing and also doesn't seem to provide any useful benefits.

There's a whole bunch of open issues and open pull requests that try to fix the various bugs that crop up due to this leaky abstraction. However, many of these are fixing the symptoms one at a time, by making the views or API code more and more complex, rather than trying to fix the underlying model. This has generally made me averse to reviewing and/or merging them, since I think they are going in the wrong direction. I'm aware however that we haven't discussed what the right destination should be!

Personally, I think the long term goal should be to remove the concept of this "special first comment" entirely, and end up with a more straightforward model where we have:

* A Note, which has a body and an author. A note can exist without any comments.
* NoteComments, none of which are "special" and each can be individually added, removed, hidden etc without affecting the validity of the Note.

This would align note comments with our other comment models, for example changeset comments and diary comments. We don't implement diary entries by having a blank DiaryEntry and a special first DiaryComment containing the text of the diary entry!

If we are all agreed on the destination, then we can start planning out the refactoring required to get there. This will also help assess the existing pull requests and open issues on this topic.

gravitystorm:
The way that I would approach this refactoring would be roughly:

* Corral all the complexity and special casing into the models, rather than having the logic spread out over controllers and views as we currently have. So if you ask for e.g. `@note.comments` you would not get the special first comment in the list returned. Other things like `current_user.note_comments` would also behave as expected, with no special first comments that then need to be ignored.
* Rework the controllers and views to act as we would expect for the end stage, e.g. when creating a Note you provide the body and author to `Note#create` rather than the two-step process currently in api/notes_controller
* Create a migration to add the author and body to the Note database tables, start to use them when they are available, and start backfilling them from the special first comments. Destroy the unnecessary special first comments when they are no longer needed.
[...]

PR #3911, opened 2023-01-26
Refactor notes to use resourceful routing

gravitystorm:
This PR refactors the two notes routes, namely:
* browse#new_note -> notes#new
* browse#note -> notes#show

This brings them into line with resourceful routing naming conventions.

The PR is slightly more complex than I anticipated, since the api methods for notes was already hogging the `notes` resourceful routing (e.g. `note_path` was used for /api/0.6/notes/123), so I renamed that to `api_notes` which is a convention already used for other models that have both api and non-api paths. Secondly, the i18n needed a little bit of unpicking, since the views were using fully qualified instead of relative translation strings. And finally, a surprising number of other tests use notes as examples (e.g. for reporting users) and so they needed updating too (e.g. to use `note_path` instead of `browse_note_path`).

Refs #3831


config/locales/en.yml
```diff
@@ -2915,22 +2928,6 @@ en:
...
-      new:
-        intro: "Spotted a mistake or something missing? Let other mappers know so we can fix it. Move the marker to the correct position and type a note to explain the problem."
-        advice: "Your note is public and may be used to update the map, so don't enter personal information, or information from copyrighted maps or directory listings."
-        add: Add Note
-      show:
-        anonymous_warning: This note includes comments from anonymous users which should be independently verified.
-        hide: Hide
-        resolve: Resolve
-        reactivate: Reactivate
-        comment_and_resolve: Comment & Resolve
```
tomhughes, line 2927: Actually `comment` and `comment_and_resolve` are used from javascript in `note.js` where the button value is updated as text is entered...

tomhughes (changes requested):

Issue #3895, opened 2023-01-14
text too close to left side


maro-21:
I think the text is too close to the left edge. Shouldn't there be some spacing?

https://www.openstreetmap.org/user/test/history
https://www.openstreetmap.org/history/nearby

![00](https://user-images.githubusercontent.com/57260466/212485180-e32c1196-2b26-4bca-916e-45f29dfcc66d.jpg)

mehul071:
Hello, 
contributing for the first time.
I solved this issue and created a pull request can you please check if its correct . @tomhughes @maro-21

tomhughes:
If you want to ask about the status of your PR then please do it on the PR and not here, but also please wait more than one day before chasing!

mehul071:
Sure, I'll keep in mind

AntonKhorev:
fixed in https://github.com/openstreetmap/openstreetmap-website/commit/b7ad61560c498adf46c97746593506141de97e47

gravitystorm:
Thanks @AntonKhorev - I knew I'd seen this somewhere but I couldn't find it yesterday.

PR #3916, opened 2023-02-02
fixed the text position in sidebar

mehul071:


HolgerJeromin:
ref #3895

gravitystorm:
Unfortunately it's not quite this simple!

We have a sidebar that includes padding by default. Some contents of the sidebar, however, are designed to stretch right to the edges, and so they require a negative padding. This is usually where we have a list of items separated by horizontal lines, for example geocoding search results, or changeset lists (like here).

If you remove the negative padding, then neither the horizontal separator line, nor the hover highlighting, meet the edge of the sidebar, and also the content gets double-padded. See this image:

Screenshot from 2023-02-08 14-50-24


What you need to do to fix this is change the error messages, so that if any of them are shown the message is wrapped in an element with the correct padding.

gravitystorm:
The original problem was later fixed in #4234, therefore I'm going to close this PR.

PR #3918, opened 2023-02-03
Add WKT point formatting to Geocoder

sabas:
Hello,
first time contributing directly, I hope it is correct.
I was trying to search coordinates in WKT (from a DB dump) directly from the main search box and I noticed this wasn't recognised as a format.

### DIFF
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -211,6 +211,9 @@ class GeocoderController < ApplicationController
       elsif latlon = query.match(/^([+-]?\d+(\.\d*)?)(?:\s+|\s*,\s*)([+-]?\d+(\.\d*)?)$/)
         params.merge!(:lat => latlon[1].to_f, :lon => latlon[3].to_f).delete(:query)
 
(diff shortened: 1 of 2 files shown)


tomhughes:
I think that's a pretty obscure thing to support. I mean once you go there why limit yourself to POINT geometries? or to WKT come to that - are we going to allow GeoJSON features etc?

sabas:
Perhaps it can be generalized if there's a more generic gem?
I actually don't know the Ruby ecosystem but I was wondering something that takes care of converting different cases where it's only a matter of algorithmic conversion

tomhughes:
The point wasn't to generalise it - the point was that this is probably going to far as it is!

It's not likely you can't copy the coordinate from the POINT geometry string easily enough and they would already work I think.

sabas:
> The point wasn't to generalise it - the point was that this is probably going to far as it is!
> 
> It's not likely you can't copy the coordinate from the POINT geometry string easily enough and they would already work I think.

Yeah that's what I was doing actually, removing POINT and the parentheses lead to the two Internal results X,Y and Y,X

I was just observing the Geocoding box could do more. :)

Issue #3910, opened 2023-01-19
RTL UI has some reversed features


yarons:
### How to reproduce the issue?

Simply login and switch to any RTL language, you'll see that the search box has overlapping words and that the buttons on the top right are clipped on the wrong side (due to hard coded left/right style arguments).

### Screenshot(s) or anything else?

![image](https://user-images.githubusercontent.com/406826/213380301-561567e5-944f-4cee-a77a-9b73871dcead.png)

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### Which browsers are you seeing this problem on?

Firefox

biswajit-k:
I think it can be resolved by adding styles for placeholder for RTL language, @gravitystorm could you please let me know if this is correct? I can work on this.

yarons:
CSS supports such behavior:
https://stackoverflow.com/questions/22511460/css-dir-rtl-vs-style-directionrtl (The resolving answer)

I can try and tinker with the CSS in my browser and take a screenshot of it if it helps.
Thanks!

tomhughes:
It can be fixed yes, my draft fix is to replace https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/stylesheets/common.scss#L29 with:

```css
[dir=rtl] {
  /* no-r2 */ text-align: right;

  ::placeholder {
    /* no-r2 */ text-align: right;
  }
}

[dir=ltr] {
  /* no-r2 */ text-align: left;

  ::placeholder {
    /* no-r2 */ text-align: left;
  }
}
```

That works, though in my (english) browser the cursor winds up at the wrong end but I'm guessing that won't be the case if your browser is configured for an RTL language?

biswajit-k:
For me, it still shows the placeholder text on the left, I think it is because `dir => "auto"` attribute is set on the input field. Also, `dir` set to `auto` always displays placeholder text on the left irrespective of the language. So, I think we should remove the `dir` attribute for the input fields and let the input and placeholder text be left or right-aligned based on the language selected.
I referred below-
[A similar issue](https://github.com/whatwg/html/issues/4903) and [an attempted fix](https://github.com/wikimedia/codex/commit/a25d22fd8d93a18a2259b297ba62e7556c542131)

yarons:
I'm not sure if we can do it but we can use CSS vars but it may solve this issue, we just simply need to avoid the "auto" configuration and set the style for this field instead.
[...]

PR #3922, opened 2023-02-09
Vel-12/issue#3910 Resolved 

Vel-12:
isssue#3921 is resolved. the fix was inside [dir=rtl]  in common.scss


gravitystorm:
@vel-12 this PR contains two commits, the second commit is fixing a mistake that you made in the first commit.

From [CONTRIBUTING.md](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md)

> The commit history is important. This is a large codebase, developed over many years by many developers. We frequently need to read the commit history (e.g. using git blame) to figure out what is going on. So small, understandable, and relevant commits are important for other developers looking back at your work in future.

I would also like to see more of an explanation in your pull request, to show that by changing the direction of form controls that you aren't introducing any side effects. See also the continuing discussion at #3910.

Vel-12:
@gravitystorm 
1. if u apply text-align:right in form-control. it only apply in the input tag of the website.
2. when we change the language from right to left or from left to right, text overlapping occurs only in the search input tag and not anywhere else in the wesbsite.
3. From prevsious solutions, the disadvantage of using ::placeholder is that it also affects the cursor.

Overall no problems.

PR #3925, opened 2023-02-09
Vel-12/issue#3910

Vel-12:
#3910
@gravitystorm

1) if u apply text-align:right in form-control. it only apply in the input tag of the website.
2) when we change the language from right to left or from left to right, text overlapping occurs only in the search input tag and 
    not anywhere else in the wesbsite.
3) From prevsious solutions, the disadvantage of using ::placeholder is that it also affects the cursor.
    Overall no problems.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -26,7 +26,7 @@ small, aside {
   margin-right: $lineheight/4;
 }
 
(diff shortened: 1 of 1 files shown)


mmd-osm:
#3910 as been closed by #4353 in the meantime. Closing this issues, since it seems redundant now.

PR #3933, opened 2023-02-14
Add limit parameter to api changesets query

AntonKhorev:
Similar to notes query limit parameter.

Defaults to 100 changesets, which was the only possible value before. Max value is also 100, feel free to change it.

Use example: I need to get the latest changeset of a user. Without the limit parameter I have to download hundred changesets, although I only need one.

### DIFF
--- a/app/controllers/api/changesets_controller.rb
+++ b/app/controllers/api/changesets_controller.rb
@@ -19,6 +19,9 @@ module Api
     # Helper methods for checking consistency
     include ConsistencyValidations
 
+    DEFAULT_QUERY_LIMIT = 100
+    MAX_QUERY_LIMIT = 100
+
     ##
     # Return XML giving the basic info about the changeset. Does not
     # return anything about the nodes, ways and relations in the changeset.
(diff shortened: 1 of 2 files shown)


saerdnaer (approved):
LGTM

gravitystorm:
Thanks, looks good to me too. Merged.

If you have a chance for a followup PR, there's some refactoring that could help here (and also for notes):

* Move the magic numbers to settings.yml, so that they are in the same place as other api paging stuff like tracepoints_per_page
* Expose them in the capabilities document, since that allows API consumers to avoid hard-coding matching magic numbers into their own code
* Refactor the tests to use the values, e.g. instead of checking for `get changesets_path(:limit => "101")` check for `get changesets_path(:limit => Settings.foo + 1)`
* When testing limits, I like to make sure we have a pair of tests for highest-permitted and lowest-forbidden. This avoids off-by-one errors. So an additional test for `:limit => Settings.foo`. You can see that e.g. in the tests where we check "255" works and "256" fails.
* Add similar tests for notes, I couldn't find any tests for these limits there.

Issue #3189, opened 2021-05-03
Change label for leisure=horse_riding


Hufkratzer:
Currently in en.yml and (en-GB.yml) the label for tag leisure=horse_riding is "Horse Riding":
<img width="183" alt="nearbyf" src="https://user-images.githubusercontent.com/5958184/116935468-b0560080-ac66-11eb-9c9c-cca6a66d8a36.png">

This is inappropriate since "horse riding" is an activity but the tag leisure=horse_riding stands for a physical place (like the other leisure tags usually do). The label should better be 
* "Horseback Riding Facility" in American English (en.yml) and 
* "Horse Riding Facility" in British English (so after changing in en.yml the string to "Horseback Riding Facility" this should be translated in Translatewiki to "Horse Riding Facility"and imported from there to en-GB.yml, right?)

Compare with
* leisure=dance -> label; "Dance Hall"
* leisure=fishing -> label: "Fishing Area"
* https://dictionary.cambridge.org/dictionary/english/horse-riding
* https://dictionary.cambridge.org/dictionary/english/horseback-riding

iD:
<img width="225" alt="iD" src="https://user-images.githubusercontent.com/5958184/116935171-463d5b80-ac66-11eb-9ca4-c9ea0cf679f3.png">

JOSM:
<img width="373" alt="JOSM" src="https://user-images.githubusercontent.com/5958184/116935213-548b7780-ac66-11eb-8647-b377d80b7077.png">

tomhughes:
You realise that the JOSM menu is nothing to do with this repository right?

Hufkratzer:
Yes, like iD and https://dictionary.cambridge.org; this was just for information.

Hufkratzer:
The name for leisure=horse_riding was changed in iD; it's now:

| en | en-GB |
| ----- | ------- |
|  ![en](https://user-images.githubusercontent.com/5958184/171515488-c67f1a33-0088-4ebe-8987-13979f01bd9d.png) | ![en-GB](https://user-images.githubusercontent.com/5958184/171515642-33a0a920-656a-4e30-92d2-c37bf8a53a39.png) |

I propose to use that on the website too.

Hufkratzer:
The name for leisure=horse_riding was also changed in JOSM (18646, 2023-01-30); it's now:
![josm-hrc](https://user-images.githubusercontent.com/5958184/218221457-3d7935dd-d30c-42d7-acf2-0f0ae63570a3.png)
[...]

PR #3935, opened 2023-02-18
Change label for leisure=horse_riding in en.yml

Hufkratzer:
This PR just changes the label for tag [leisure=horse_riding](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dhorse_riding)
**from "Horse Riding" to "Horse Riding Centre"** in en.yml and thus resolves https://github.com/openstreetmap/openstreetmap-website/issues/3189

The motivation is to make it clearer to the user what the object tagged in this way is. Simply "hoirse riding" is too generic and an activity rather than a location for that activity. Other tags with suboptimal names also have more descriptive labels, e.g.
leisure=dance : "Dance Hall"
leisure=fishing : "Fishing Area"

The presets for leisure=horse_riding in the main editors iD and JOSM are now also named "Horse Riding Centre" (or "Horseback Riding Center" in US English, see https://github.com/openstreetmap/openstreetmap-website/issues/3189#issuecomment-1144229876 f).

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -970,7 +970,7 @@ en:
           fitness_station: "Fitness Station"
           garden: "Garden"
           golf_course: "Golf Course"
-          horse_riding: "Horse Riding"
+          horse_riding: "Horse Riding Centre"
           ice_rink: "Ice Rink"
           marina: "Marina"
           miniature_golf: "Miniature Golf"


gravitystorm:
Thanks @Hufkratzer for your PR. In particular, thank you for including all the context and explanations - it has really helped make it easier for me to make a decision on this, without having to go and do all the research myself!

PR #3936, opened 2023-02-22
Avoid using html in translations

gravitystorm:
This is the first batch of changes to avoid using html in our translation strings. It reworks various views to make sure that all html (e.g. links, paragraphs, bold tags etc) are defined in the view, and that the translations only contain the words that need to be translated. The aim is, among other reasons, to make it easier to translate since it avoids having to get the brackets, quoting etc correct to ensure valid html.

The primary challenge with this refactoring is, as usual, to make sure that the translation keys change whenever there is a difference with interpolations, which leads to a little strain around picking new translation keys.

### DIFF
--- a/app/views/accounts/_go_public.html.erb
+++ b/app/views/accounts/_go_public.html.erb
@@ -1,0 +1,14 @@
+<hr class="mb-3" />
+<a name="public"></a>
+<h2><%= t ".heading" %></h2>
+<p>
+  <%= t ".currently_not_public" %>
+  <strong><%= t ".only_public_can_edit" %></strong>
+  <%= t ".find_out_why_html", :link => link_to(t(".find_out_why"),
+                                               t(".find_out_why_url")) %>
+</p>
+<ul>
+  <li><%= t ".email_not_revealed" %></li>
+  <li><%= t ".not_reversible" %></li>
+</ul>
+<%= button_to t(".make_edits_public_button"), user_go_public_path, :class => "btn btn-primary" %>
--- a/app/views/accounts/edit.html.erb
+++ b/app/views/accounts/edit.html.erb
@@ -65,18 +65,5 @@
 <% end %>
 
 <% unless current_user.data_public? %>
-  <hr class="mb-3" />
-  <a name="public"></a>
-  <h2><%= t ".public editing note.heading" %></h2>
-  <p>
-    <%= t ".public editing note.currently_not_public" %>
-    <strong><%= t ".public editing note.only_public_can_edit" %></strong>
-    <%= t ".public editing note.find_out_why_html", :link => link_to(t(".public editing note.find_out_why"),
-                                                                     t(".public editing note.find_out_why_url")) %>
-  </p>
-  <ul>
-    <li><%= t ".public editing note.email_not_revealed" %></li>
-    <li><%= t ".public editing note.not_reversible" %></li>
-  </ul>
-  <%= button_to t(".make edits public button"), user_go_public_path, :class => "btn btn-primary" %>
+  <%= render :partial => "go_public" %>
 <% end %>
--- a/app/views/accounts/edit.html.erb
+++ b/app/views/accounts/edit.html.erb
@@ -65,8 +65,18 @@
 <% end %>
 
 <% unless current_user.data_public? %>
-<a name="public"></a>
-<h2><%= t ".public editing note.heading" %></h2>
-<%= t ".public editing note.html" %>
-<%= button_to t(".make edits public button"), user_go_public_path, :class => "btn btn-primary" %>
+  <hr class="mb-3" />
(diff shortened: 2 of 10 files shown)


app/views/accounts/_go_public.html.erb
```diff
@@ -0,0 +1,14 @@
+<hr class="mb-3" />
+<a name="public"></a>
+<h2><%= t ".heading" %></h2>
+<p>
+  <%= t ".currently_not_public" %>
+  <strong><%= t ".only_public_can_edit" %></strong>
+  <%= t ".find_out_why_html", :link => link_to(t(".find_out_why"),
+                                               t(".find_out_why_url")) %>
```
tomhughes, line 8: Is this a bit too close to being lego? Are we sure splitting the sentence into pieces like this is going to be OK in all translations?

tomhughes (commented):
So a lot of the translations that no longer have HTML in still have an `_html` suffix which I guess is necessary because we substituting HTML into them via variables?

gravitystorm:
> So a lot of the translations that no longer have HTML in still have an _html suffix which I guess is necessary because we substituting HTML into them via variables?

Yes, that's the case. When the conversions are finished I'm planning on creating a linter-type test to ensure that no `<>` characters are found in both the `en.yml` file and also the received translations from translatewiki.

app/views/accounts/_go_public.html.erb
gravitystorm: Lego translations are where we stick two translations together side-by-side, like `"string" + "string"` which prevents translators from rearranging the sentence to meet any grammatical needs. Interpolating is fine since they can still put the words in whichever order they like, e.g. moving words from after the link text to before the link text.

```erb
<%= t(".first_part") + link_to(foo) + t(".second_part") %> # lego translation
vs
<%= t(".whole_sentence"), :link => link_to(foo) %> # no problem
```

We already use the interpolation approach to links in numerous places.

There are two things we should consider with the interpolation approach:
* Don't be tempted to reuse the interpolated link text between different pages, since that text itself might depend on the rest of the sentence
* There's a risk that splitting the sentence into two keys makes it harder to translate, e.g. if you are looking at the link text, it could be harder to discover what context it's being used in. So we should keep the names of the keys similar (and keep them in the same naming hierarchy) so that it's more obvious to translators how the keys are related. I try to follow the convention of "blah blah %{foobar_link}", with the `foobar` key being the link text and `foobar_url` if the url is translateable.

tomhughes: I understand all that - my point was that what was previously a single piece of text is now three pieces. Sure it's not concatenated with `+` but three adjacent translation lookups in HTML amounts to the same thing.

So if you have a language where the "find out why" bit needs to be before the "only public can edit" bit for some reason you can no longer easily do that - you probably could just move the text from one translation to the other but people aren't going to realise that.

gravitystorm: I think it's worth the tradeoff in any slight translation difficulty to be able to avoid translators having to get the html right. 

The approach here, of having the link text in a separate string interpolated into the whole sentence, is exactly what we need to use when the urls are being generated - e.g. we can't put raw html with path helpers like `foo bar <a href="new_diary_entry_path(user)>link text</a>` into a translation. So there's also an argument that for consistency, all sentences with links should follow the same interpolation pattern, even if the link could be inlined into the translation.

tomhughes: Right, but we could do what you do with some of the others and make both the link and the only public can edit part substitutions into a single top level translation?

tomhughes: Well yes part of this is probably that I hate the whole way that is written, especially the final "sentence" that is entirely in parentheses ;-)

gravitystorm: Option 1 - put the parenthetical into the bolded sentence:

> Currently your edits are anonymous and people cannot send you messages or see your location. To show what you edited and allow people to contact you through the website, click the button below. **Since the 0.6 API changeover, only public users can edit map data** ([find out why](https://wiki.openstreetmap.org/wiki/Anonymous_edits)).

Option 2 - remove the parentheses from the final sentence, and capitalise

> Currently your edits are anonymous and people cannot send you messages or see your location. To show what you edited and allow people to contact you through the website, click the button below. **Since the 0.6 API changeover, only public users can edit map data.** [Find out why](https://wiki.openstreetmap.org/wiki/Anonymous_edits).

I'm happy to implement whichever is preferred, or any another option too. If it helps, I lean slightly towards option 2.

tomhughes: Honestly I think it's probably not worth spending a lot of time worrying about it - hardly anybody will even see that screen.

Issue #3932, opened 2023-02-13
Notes from website should include an indication the note was created here


angoca:
### Problem

Notes are created with an API call via a URL. Many applications offer the option to create notes. Most of them append a hashtag with the application's name (StreetComplete, OrganicMaps, Maps.me, OnOSM), which allows one to identify the source of the note. Also, notes can be created directly with a URL by providing all the necessary parameters.

When solving notes, knowing the note's origin could give an idea of its creation and what the user wanted to provide. Also, some notes are spam because they are created in a bulk way by using the URL directly.

### Description

I want notes created from the website to include a hashtag or an extra test indicating they were created. This allows us, the resolvers, to identify these notes from other notes.

### Screenshots

![image](https://user-images.githubusercontent.com/294317/218484154-4c40c525-b17a-4f89-b1e5-7ebc767b6d55.png)

![image](https://user-images.githubusercontent.com/294317/218484281-6ffbcdc1-5d86-4838-9c06-79d57109a3f8.png)

![image](https://user-images.githubusercontent.com/294317/218484346-f8501b27-a9c8-440a-84d1-62356c70ea4b.png)

tomhughes:
Notes are deliberately lightweight and do not include a lot of metadata and it really shouldn't matter what created them.

angoca:
Just appending a #osm.org could be enough to identify the source.

I have solved thousands of notes, and it is always good to know how the note was created if it was from an app (on the ground) or on a computer (the location could not be very precise).
If the note was from a mobile app, then the location could be very accurate or just a few meters far. Instead, from a computer (a website), the note could not be that accurate, because some people do not pay attention to it.

This is relevant if the note was created in the middle of nothing, and the text indicates there is, for example, a new shop. With a mobile app, I will think this is true, but from a website, this could be a misplaced note.

angoca:
In this moment, there is an issue with Iranian notes in the middle of the ocean. It seems they were created with an OnOSM fork. But thanks to the hashtag, it was easy to identify the error. If something happens with the notes from the website, with a hashtag it will be easier to identify the source of the issue.

https://github.com/osmlab/onosm.org/issues/103

matheusgomesms:
It is very useful to know where the note came from (and I actually believe this info should be somewhat mandatory). For example, notes created from Instagram (not sure how they do it) contain a lot of errors, and we are not sure if the notes came from there (except when the note creator mentions it).

[...]

PR #3938, opened 2023-02-22
Include hashtags in notes

angoca:
This allows to identify notes created from website, from other notes, which is useful when filtering and solving notes.

### DIFF
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -50,7 +50,7 @@ OSM.NewNote = function (map) {
       data: {
         lat: location.lat,
         lon: location.lng,
-        text: $(form.text).val()
+        text: $(form.text).val() + "#osm.org"
       },
       success: function (feature) {
         noteCreated(feature, marker);
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -50,7 +50,7 @@ OSM.NewNote = function (map) {
       data: {
         lat: location.lat,
         lon: location.lng,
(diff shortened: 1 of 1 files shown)


tomhughes:
If we're going to do this then we should do it properly as meta data on the model and not by expecting people to guess the meaning of bits of the note body.

gravitystorm:
@angoca thank you for your pull request. 

I agree with @tomhughes on this, and I have several other objections to this approach, as [I've commented in the related issue](https://github.com/openstreetmap/openstreetmap-website/issues/3932#issuecomment-1440222022). So I'm going to decline this pull request.

I'm going to review this PR anyway, with the aim of helping you with making future PRs. So here are my notes:

* I'm not sure that `#website` is particularly descriptive, especially if other hashtags are in use for other reasons than just marking what creating the notes (e.g. if there are hashtags for topics or challenges). It doesn't clearly indicate which website created it, for example, and other websites might start using `#website` too. Instead I would consider using the "generator" from our configuration, or perhaps something from the translations (`layouts.project_name`?) that is a) more descriptive b) more precise and c) likely to be configured differently for other installations of the software.
* If we are formally adding a structure to what was previously free-form text, we would need to indicate some kind of parsing rules for applications that want to consume it. For example, what characters are valid in a hashtag, what characters delineate one, etc.
* From [CONTRIBUTING.md#pull-requests](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests) - please don't include fixup commits. One of the commits in this PR is entirely replacing the other, so they should be combined together (using e.g. `git rebase -i`) into a single clean commit.

PR #3942, opened 2023-02-25
Improve the preview of diary entries on social media sites (using OpenGraph HTML tags)

amandasaurus:
Currently the “open graph” tags for social media sites are very uniform across the website. The description is always a short blurb about OSM. This patch starts customizing them, with a diary entry page.

I am new to Ruby, and might have made mistakes. Please correct me if I've done something wrong.

### DIFF
--- a/app/controllers/diary_entries_controller.rb
+++ b/app/controllers/diary_entries_controller.rb
@@ -66,6 +66,24 @@ class DiaryEntriesController < ApplicationController
     if @entry
       @title = t ".title", :user => params[:display_name], :title => @entry.title
       @comments = can?(:unhidecomment, DiaryEntry) ? @entry.comments : @entry.visible_comments
+      @og_tags = {
+        "og:type" => "article",
+        "og:locale" => @entry.language_code,
+
+        # helpers.user_image_url is a hack, and should be done Properly™
+        "og:image" => helpers.user_image_url(@user),
+        "og:image:secure_url" => helpers.user_image_url(@user),
+        # user_image_url returns a 100×100 image
+        "og:image:width" => "100",
+        "og:image:height" => "100",
+
+        # Should/Could this be translated?
+        "og:image:alt" => "Profile image of #{@user.display_name} on OpenStreetMap.org",
+
+        # Including the entire body sounds bad. Arbitrarily choose 300 chars as the limit.
+        # It would be better to strip out all HTML, or Markdown, syntax here.
+        "og:description" => @entry.body.to_text[0..300]
+      }
     else
       @title = t "diary_entries.no_such_entry.title", :id => params[:id]
       render :action => "no_such_entry", :status => :not_found
--- a/app/helpers/open_graph_helper.rb
+++ b/app/helpers/open_graph_helper.rb
@@ -1,6 +1,7 @@
 module OpenGraphHelper
-  def opengraph_tags(title = nil)
-    tags = {
+  def opengraph_tags(title, og_tags)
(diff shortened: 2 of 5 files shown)


gravitystorm (changes requested):
We discussed this in person at the Karlsruhe Hack Weekend, so this is only a summary of that review. Namely:

* We can rework the default handling in opengraph_tags to use a more conventional ruby approach of using `tags = {}` in the method signature, and `tags.reverse_merge(defaults)`
* However, we need to rethink how the function handles the title. It's a specific parameter at the moment, but it makes sense to combine that with other overrides somehow
* You're going to add the user profile image too, which is important because that uses a view helper, and those aren't normally available in controllers. Although we can bodge that in (by using `helpers` in the controller) it's an indication that we're mixing up the separation of concerns between controllers and views
* For the body, we need to check whether implementations handle "og:description" as html, or whether we need to pass only textual content to that
* entry.body will return raw markup, whether html or markdown, and we probably don't want that in the description
* We need some tests for the three different diary entry body types
* We need some rational for why 100 characters, and not e.g. 200 
* Is there a better way to truncate the text, rather than just characters. I couldn't find an example in the codebase but perhaps there is either a rails helper or something similar to intelligently truncate. 
* We need some other tests just to make sure that the type, title etc is what we expect so that it survives future refactoring.

gravitystorm:
Beyond that, I would like to find a better approach where the views pass these overrides into the opengraph tags, rather than using instance variables in the controllers. I believe the view is the right place for declaring these tags, since choosing "how something is shown" is the concern of the view, rather than the controller (which is more about fetching the data). So I would expect all the choices about titles, length of text etc to be written in app/views/diary_entries/show.html.erb and somehow make their way to the right place in the layout. For example, we use `content_for` for passing html to the `<head>` already, but  it's not quite the same here since we want to pass overrides to the opengraph defaults.

amandasaurus:
I have made the requested changes (that I am able to do), and force pushed to the branch.

I agree that having html generation & formatting should probably be in the HTML section (“views”?), but I don't know enough ruby on rails to know how to do that, Please feel free to refactor it mercifullessly. You may also give me an example which I can copy and paste.

gravitystorm:
There's a few things that come up here on further review:

* As hinted at in the comments, I think it's worth removing the markup from the description, rather than outputting it. The PR currently uses "to_text" which just returns the raw content, rather than a text-only representation. So I would prefer to see "something <b>important</b>" be outputted as "something important". We could use the `[strip_tags](https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-strip_tags) helper on the `to_html` output to do so
* The description is shortened by calling `[0..300]`, but we could use the rails `[truncate](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-truncate)` helper to do this more cleanly.

I think we can also look at the [meta-tags](https://github.com/kpumuk/meta-tags) gem, which doesn't cover the above points, but it could help us with setting defaults etc without having to build our own `set_meta_tags` equivalent. That could be done in a "prerequisites" pull request (i.e. another PR that introduces that gem, and then this PR refactored to use it).

tomhughes:
We should probably just improve what `to_text` in the rich text renderers does - the main place it is used at the moment is for plain text versions of emails and stripping markup there would be fine.

amandasaurus:
Yous have some good improvement suggestions here. But alas, my Ruby skill is not good enough to implement it. and unfortunately, I don't have the time to learn that.

PR #3952, opened 2023-03-01
add IDs to note and changeset mail subjects

Strubbl:
It would be very cool to have the note ID in the subject of emails, when a note was updated. This would also be helpful for changeset mails. I only did update the locale file for en, because i do not know where and how this needs also adjustments in the code to fill the ID place holders with the actual ID. This shall at least give an idea of i want to achieve.

At the moment the ID is only given indirectly via the `noteurl` or `changeset_url` in the `user_mailer.rb`. Is it anyhow easily possible to access the ID directly in the corresponding notification functions? Would it be ok for you to improve the mail subject to contain the ID?

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1585,8 +1585,8 @@ en:
       anonymous: An anonymous user
       greeting: "Hi,"
       commented:
-        subject_own: "[OpenStreetMap] %{commenter} has commented on one of your notes"
-        subject_other: "[OpenStreetMap] %{commenter} has commented on a note you are interested in"
+        subject_own: "[OpenStreetMap] %{commenter} has commented on your note %{note_id}"
+        subject_other: "[OpenStreetMap] %{commenter} has commented on note %{note_id} you are interested in"
         your_note: "%{commenter} has left a comment on one of your map notes near %{place}."
         your_note_html: "%{commenter} has left a comment on one of your map notes near %{place}."
         commented_note: "%{commenter} has left a comment on a map note you have commented on. The note is near %{place}."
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -1585,8 +1585,8 @@ en:
       anonymous: An anonymous user
       greeting: "Hi,"
       commented:
-        subject_own: "[OpenStreetMap] %{commenter} has commented on one of your notes"
-        subject_other: "[OpenStreetMap] %{commenter} has commented on a note you are interested in"
+        subject_own: "[OpenStreetMap] %{commenter} has commented on your note %{note_id}"
+        subject_other: "[OpenStreetMap] %{commenter} has commented on note %{note_id} you are interested in"
         your_note: "%{commenter} has left a comment on one of your map notes near %{place}."
```
gravitystorm, line 1590: You will also need to change the other `subject_own` and `subject_other` translations further down this file - there are separate subjects for "commented" "closed" and "reopened" events.

config/locales/en.yml
```diff
@@ -1611,8 +1611,8 @@ en:
       hi: "Hi %{to_user},"
       greeting: "Hi,"
       commented:
-        subject_own: "[OpenStreetMap] %{commenter} has commented on one of your changesets"
-        subject_other: "[OpenStreetMap] %{commenter} has commented on a changeset you are interested in"
+        subject_own: "[OpenStreetMap] %{commenter} has commented on your changeset %{changeset_id}"
+        subject_other: "[OpenStreetMap] %{commenter} has commented on changeset %{changeset_id} you are interested in"
         your_changeset: "%{commenter} left a comment at %{time} on one of your changesets"
```
gravitystorm, line 1616: Unfortunately, for technical reasons explained below, you can't change the interpolation variables in a translation, without also changing the key.

The technical reason is that the interpolation variables must be exactly the same in all translations for the same key. It can take several weeks, or longer, for all the translators to update their translations, and in the meantime the code would not work in those languages. So if you change the interpolations in a value, you need to change the key (e.g. `subject_own`) to something else too.

gravitystorm (changes requested):
@Strubbl thank you for your PR. I've marked this as a draft, since it is only partially implemented and not suitable for merging.

You asked where the interpolation values would come from, and one example is here:

https://github.com/openstreetmap/openstreetmap-website/blob/e9e4b10473fa536cdb840303ee4afccf71c69329/app/mailers/user_mailer.rb#L143-L147

You can see how the `commenter` variable is set. You need to do something similar for e.g. `note_id`. You can access the note id in this context in various ways, e.g. `comment.note.id`.

You will also need to adjust the tests, or write new ones, to make sure the subject line is filled in correctly.

PR #3966, opened 2023-03-10
Add space between form and Publish diary button

willemarcel:
It was missing some space between the form and the Publish button:
image

With this change:
image


gravitystorm:
Thanks!

Issue #1828, opened 2018-04-18
Change Windows Live to Microsoft Account


gravitystorm:
I found this request [from one of the Translatewiki translators](https://translatewiki.net/wiki/Thread:Translating_talk:OpenStreetMap/Change_Windows_Live_to_Microsoft_account) and it seems reasonable to me!

---

The Windows Live brand has been long discontinued. You can find a list of what Windows Live services were renamed to [here](https://en.wikipedia.org/wiki/Windows_Live#Renamed_services).

I believe the replacement word should be 'Microsoft account'. Please update the icon on [the login page](https://www.openstreetmap.org/login) and the following strings to reflect this change: `user.login.auth_providers.windowslive.title` and `user.login.auth_providers.windowslive.alt`

tomhughes:
I'm not clear what change they are asking for where the icon is concerned? As far as I can tell other sites are using the "Microsoft Account" wording but are still using the windows flag icon like us...

gravitystorm:
We're using an old Windows logo ([Vista or Windows 7 I think](http://logos.wikia.com/wiki/Microsoft_Windows)), whereas the Microsoft logo now includes [the coloured squares without the wave effect](http://logos.wikia.com/wiki/Microsoft). I would go for the coloured squares part of the Microsoft logo since that's what's shown on the following page.

tomhughes:
Well that's an interesting question...

Probably yes, despite that fact that the "windowslive" name also appears throughout the code.

The only other thing we might want to do is change the provider name, so the URLs don't have it in, but they only appear briefly and changing that will mean running a migration and also starts to make the source inconsistent unless we change it everywhere.

tomhughes:
I mean that the code is littered with that text especially in the tests.

The provider name is the value with goes in the database to identify which auth provider is in use for a given user, and which appears in the auth and callback urls.

neboduus:
Ok all clear! I will try to solve this little issue and do a pull request. If I will need suggestions I will use this issue discussion to ask. If you everyone have others things to be taken into consideration let me know.

noloman:
hi! is this issue still up for grabs? I'd like to start collaborating! 
Thanks!

tomhughes:
Yes it is still up for grabs.

SachitSankhe:
Hi, is this issue still up for resolving

faebebin:
Also stumbled across and would like to implement this. Think the logo is already updated:
[...]

PR #3970, opened 2023-03-14
Changed windowslive user strings to Microsoft login

konrad-wila:
#1828


gravitystorm:
Thanks @konrad-wila for the PR! A couple of quick points:

1. For the translations files, please only make changes to the en.yml file.

    https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#i18n

2. Your second commit will also need merging into the first, since it's just undoing the changes to the Gemfile.

    https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests (Avoid including "fixup" commits.)

tomhughes:
You don't have to close a PR just because we ask for changes, it's much better to force push an update and keep all the history in once place than to open a new PR with the modified version.

tomhughes:
I assume from your comment on #1828 that you're planning to change this again to only change the UI text?

tomhughes:
Please don't delete `yarn.lock` and `Gemfile.lock`.

Issue #3972, opened 2023-03-15
Update i18n strings to use the count placeholder instead of the number 1


gravitystorm:
I spotted this in another repository, and it applies to us too:

> `one` is used for numbers other than 1.
> A common mistake is to think that "one" is only for only the number 1.
> Instead, "one" is a category for any number that behaves like 1.
> So in some languages, for example, one → numbers that end in "1" (like 1, 21, 151) but that don't end in 11 (like 11, 111, 10311).
> - http://cldr.unicode.org/index/cldr-spec/plural-rules

We have a mixture of examples in en.yml, e.g.

```yaml
one: 1 way
one: "%{count} new message"
```
Some of our translators have figured out that you can always use `%{count}` in these circumstances. For example, in Ukrainian they have translated `one: 1 way` as `one: '%{count} лінії'` and this will work fine for e.g. 51 ways. Other languages can also do the same, if they are affected by this.

However, we should standardise on using the %{count} placeholder in en.yml, to help guide our translators to let them know this is possible. Not all translators will realise that they can use the `%{count}` placeholder, since it's not shown in the original string.

It would also have a small benefit if we start validating that each translation takes the same placeholders as the original.

gravitystorm:
Fixed by #3973

PR #3973, opened 2023-03-15
Include count in messages for the "one" plural category

tomhughes:
This fixes almost all the cases referred to in #3972.

I think the only remaining case is `users.index.showing` which is kind of weird and is only ever seen by administrators.


gravitystorm:
Thanks!

gravitystorm:
> I think the only remaining case is `users.index.showing` which is kind of weird and is only ever seen by administrators.

Yeah, that's a weird one. I think that falls into the "diminishing returns" category for refactoring!

PR #3975, opened 2023-03-20
Title and alt text changed from Windows Live to Microsoft Account

wtrbtl03:
Fixes Issue #1828 
Changed title and alt text of the button.
No variables were renamed, so functioning of auth should stay unaffected.
Internally the identifier "windowslive" is being used (as it was), just the UI shows "Log in with Microsoft" instead.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -213,7 +213,7 @@ en:
       openid: OpenID
       google: Google
       facebook: Facebook
-      windowslive: Windows Live
+      windowslive: Microsoft
       github: GitHub
       wikipedia: Wikipedia
   api:
@@ -1780,8 +1780,8 @@ en:
           title: Login with Facebook
           alt: Login with a Facebook Account
         windowslive:
-          title: Login with Windows Live
-          alt: Login with a Windows Live Account
+          title: Login with Microsoft
+          alt: Login with a Microsoft Account
         github:
           title: Login with GitHub
           alt: Login with a GitHub Account


tomhughes:
Thanks, but @konrad-wila has already been working on this in #3970.

Yes I know it's closed right now but as far as I know they are still working on it and will likely reopen it at some point.

tomhughes:
I'm a bit confused about why there are suddenly two people trying to change this.

I actually thought you were somebody from Microsoft when you opened your one last week because I knew they wanted to fix this and was expecting a PR from them - it was only when I looked at your profile this morning that I realised you weren't.

I do want to see if I can reach out to them though to double check exactly what it was they wanted before we merge this.

konrad-wila:
I can understand your confusion regarding the sudden appearance of two people trying to change the same thing. It's great that you thought I was from Microsoft, but I'm just an independent contributor trying to make a positive contribution to the project.

It's good to hear that you're planning to reach out to Microsoft for clarification. This will help ensure that we're all on the same page and that the changes we make align with their goals.

Please let me know if there's anything I can do to help in the meantime.

wtrbtl03:
Sure, we should wait till @tomhughes clarifies from Microsoft. They might suggest stopping the use of the [omniauth-windowslive](https://github.com/joel/omniauth-windowslive) gem that is currently being used and to handle the auth a different way.

Leaving the PR open till then.

tomhughes:
They can request stopping using omniauth as much as they like but it won't be happening!

tomhughes:
This will also resolve #1828.

wtrbtl03:
> Fixes Issue #1828

Yes as mentioned in first comment.

gravitystorm:
@tomhughes does this still need to wait for anything? I'm happy to merge it as-is (although a better commit message would be nice) even without hearing back from Microsoft.

tomhughes:
I think my only question would be verifying that is in fact the correct (current) name according to Microsoft.

tomhughes:
I struggled to find any concrete documentation on Microsoft recommendations but this has to be an improvement on the long dropped Windows Live naming. The best I found was actually [wikipedia](https://en.wikipedia.org/wiki/Windows_Live#Services) which says Microsoft Account is the new name.

Issue #3988, opened 2023-03-25
Share button unusable on a cell phone


jidanni:
When one wants to share their location, one taps the Share button on the right panel. Alas, on a cell phone the Share dialogue completely covers the entire screen! For instance, if we want to add a marker, we  want to see where the marker is going to be.

(More and more people are using cell phones these days.)

ayushanand308:
@jidanni A possible solution can be to limit the size of share dialogue and make it appear on the top instead of right side same as we get the popup on clicking the hamburger icon on top. What do you suggest?

jidanni:
Well all I know is e.g., https://maps.nlsc.gov.tw/
even has a PWA that makes using the site on a cell phone a breeze.

Prathm-s:
Hello @jidanni, I am intreseted in this issue, Could you please assign it to me?

jidanni:
I would love to but I'm not in charge here.

PR #3992, opened 2023-03-31
issue:#3988 Updated common.scss 

Prathm-s:


tomhughes:
Please when you are writing a commit explain what you are doing in the commit message - don't just write "updated <file>" as your commit message as it tells us, and future maintainers, nothing about what your update is supposed to do or why you wanted to make it.

Much the same goes for the PR itself - you need to explain a little bit about what you are trying to do.

Of course if you write a good commit message and the PR is a single commit then the PR description will default to that comment message meaning we get both things at the same time.

Prathm-s:
Fixes Issue #3988  

When we try to share location from mobile view the whole screen get blocked by sidebar layout so user can't see the map view.  I just edited sidebar width : 100% to 50%, so map can be visible while sharing.

PR #4000, opened 2023-04-04
Update URLs to use https

amire80:
The two gov.si URLs redirected to
the https URLs to which I am changing these.
Simply changing the current URLs to https
didn't work.

In the other URLs, simply changing to https.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2020,9 +2020,9 @@ en:
           (public information of Slovenia).
         contributors_si_slovenia: Slovenia
         contributors_si_gu: Surveying and Mapping Authority
-        contributors_si_gu_url: http://www.gu.gov.si/en/
+        contributors_si_gu_url: https://www.gov.si/en/state-authorities/bodies-within-ministries/surveying-and-mapping-authority/ 
         contributors_si_mkgp: Ministry of Agriculture, Forestry and Food
-        contributors_si_mkgp_url: http://www.mkgp.gov.si/en/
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -2020,26 +2020,26 @@ en:
...
-        contributors_es_scne_url: http://www.scne.es/
+        contributors_es_scne_url: https://www.scne.es/
         contributors_es_cc_by: CC BY 4.0
         contributors_es_cc_by_url: https://creativecommons.org/licenses/by/4.0/
         contributors_za_credit_html: |
           %{south_africa}: Contains data sourced from %{ngi_link}, State copyright reserved.
         contributors_za_south_africa: South Africa
         contributors_za_ngi: "Chief Directorate: National Geo-Spatial Information"
-        contributors_za_ngi_url: http://www.ngi.gov.za/
+        contributors_za_ngi_url: https://www.ngi.gov.za/
```
tomhughes, line 2042: That URL results in a certificate warning. Did you mean to use https://ngi.dalrrd.gov.za/ instead? That is where the http one redirects to?

tomhughes (changes requested):

config/locales/en.yml
Firefishy: Yes, https://ngi.dalrrd.gov.za/ is now the correct URL for NGI.

config/locales/en.yml
Denfeq, line 1: Дубликат #

Denfeq (approved):
[https://ngi.dalrrd.gov.za/]()

config/locales/en.yml
amire80: Thanks, fixed.

PR #4007, opened 2023-04-12
Add some validations for i18n values

gravitystorm:
These only apply to the en.yml file for now, but can be expanded in due course.

I had a long search for any third-party utilities that could do this for us, but I didn't find anything appropriate. None of the generic yaml validators that I found allowed custom validations of values (most were just focussed on validating the type of the value, int/string/hash etc against a fixed schema). I also checked for any i18n-specific validators, but again I didn't find any that could validate the values.

I expect to expand these checks to cover the rest of the translations, but there are currently a bunch of translations in the output that contain html elements (that aren't included in en.yml, which is interesting). I also expect to expand these check to add more validations, perhaps validating urls point to the same domain, and things like that.

### DIFF
--- a/test/lib/i18n_test.rb
+++ b/test/lib/i18n_test.rb
@@ -55,6 +55,20 @@ class I18nTest < ActiveSupport::TestCase
     end
   end
 
+  def test_en_for_raw_html
+    en = YAML.load_file(Rails.root.join("config/locales/en.yml"))
+    assert_nothing_raised do
+      check_values_for_raw_html(en)
+    end
+  end
+
+  def test_en_for_nil_values
+    en = YAML.load_file(Rails.root.join("config/locales/en.yml"))
+    assert_nothing_raised do
+      check_values_for_nil(en)
+    end
+  end
+
   private
 
   def translation_keys(scope = nil)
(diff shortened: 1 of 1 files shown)


test/lib/i18n_test.rb
```diff
@@ -55,6 +55,20 @@ class I18nTest < ActiveSupport::TestCase
...
   end
 
+  def test_en_for_raw_html
+    en = YAML.load_file(Rails.root.join("config/locales/en.yml"))
+    assert_nothing_raised do
+      check_values_for_raw_html(en)
+    end
+  end
+
+  def test_en_for_nil_values
```
tomhughes, line 65: What is this actually checking for? How would a `nil` arise?

gravitystorm: I had a `nil` value in the branch I was working on when developing this PR, due to some fat fingering in a different commit. I had left a key with nothing after the colon while refactoring.

The `nil` value didn't make it past any pull review (I fixed it before submitting that other PR), so it's slightly hypothetical. But it did cause the other test to error out, so I thought it might be worth adding separately, so that you get a clean error message.

I'm happy to remove it if you think it's unnecessary or potentially confusing.

PR #4013, opened 2023-04-20
Update docker compose to solve stray PID problem

danrademacher:
Resolves https://github.com/OpenHistoricalMap/issues/issues/516 but not sure if it might cause other problems? Also not sure why no one upstream at OSM has ever had to deal with this issue, since it seems at least somewhat known in the Rails world, based on https://ieftimov.com/posts/docker-compose-stray-pids-rails-beyond/

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,8 @@ services:
   web:
     build:
       context: .
+    environment:
+      PIDFILE: /tmp/pids/server.pid
     volumes:
       - .:/app
       # Prevent these directories from mounting so they're not shared between host OS and Docker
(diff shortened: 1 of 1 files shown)


danrademacher:
Ah, drat. I mean to file this against our OHM fork! Well, if it's useful upstream, that would be great. But if not, also OK to close this out

gravitystorm:
> Well, if it's useful upstream, that would be great. But if not, also OK to close this out

PRs are always welcome! I haven't reviewed this, but if it's been causing problems for developers of OHM then a fix is likely to be welcome here too.

PR #4038, opened 2023-05-20
add installation instructions for Ubuntu 23.04

SteveC:
Added instructions for Ubuntu 23.04 and tested them.

* Moved geckodriver to explicit download and install as there is no longer a package
* ruby versions aren't needed
* All the image helper warnings when doing db:create fixed with relevant packages
* Node/yarn all fail with the default packages, moved to installing nvm
* svgo is a snap

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -39,6 +39,34 @@ sudo apt-get install ruby3.0 libruby3.0 ruby3.0-dev \
 sudo gem3.0 install bundler
 ```
 
+For Ubuntu 23.04:
+
+```
+sudo apt-get install ruby libruby ruby-dev ruby-bundler \
+                     libvips-dev libxml2-dev libxslt1-dev \
+                     build-essential git-core pngcrush \
+                     postgresql postgresql-contrib libpq-dev libsasl2-dev \
+                     libffi-dev libgd-dev libarchive-dev libbz2-dev \
+                     libjpeg-turbo-progs advancecomp optipng pngquant jhead \
+                     jpegoptim gifsicle
+
+snap install svgo
+
+# geckodriver APT package no longer exists, so install maually:
+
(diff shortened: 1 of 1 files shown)


tomhughes:
I'm not sure it's practical to have separate instructions for specific versions of specific operating systems - there's just no way to even try and keep them up to date.

I'd prefer to keep using an LTS release as our baseline, and yes I know we need to fix the 22.04 instructions and doing so would quite likely give us something that works for 23.04 as well.

The problem is how to fix it - telling people to pipe the output of curl into a shell is just not acceptable. I know it's what node upstream tell people to do but it's still an absolutely terrible idea from a security point of view to tell people to do that.

gravitystorm:
Thanks @SteveC for your updated instructions.

I've been working on #4048 today, which includes updates our installation instructions to 22.04, and as part of that I think we've resolved many of the things you covered here. But let me work through them:

* I agree with @tomhughes , I'd prefer to maintain instructions only for Ubuntu LTS releases. I'm happy to add extra information for non-LTS releases (like adding a line or two explaining any differences, rather than copy+pasting lots of nearly-identical things) if necessary.
* I've dropped the version numbers from the ruby packages in #4053, so the same apt-install command works on both 22.04 and 23.04
* You don't need to install geckodriver separately, it's included in the firefox snap package.
* While I know that the yarn package in 22.04 is broken, the (admittedly ancient) version of node in 22.04 is fine for us, and 23.04 contains node 18 which is also fine. So I've updated our instructions to use the packaged nodejs, npm to install yarn, and that should also work on 23.04 too. This avoids the pipe-to-curl issue, and avoids any headaches related to node version managers.
* The remaining topic is image_optim, and I'd rather deal with that by silencing the warnings completely, see #4055

Let me know if I've missed anything. I think that after all the #4048 PRs are merged then the INSTALL.md will work fine with 23.04 too, they are very similar.

Issue #4040, opened 2023-05-21
Allow location to be searched with exiftool format


shdpl:
### Problem

Hi,
when i get location from the image using probably the most popular CLI tool which is the exiftool like that:
```
exiftool IMG_8109.JPG | grep 'GPS Position'
GPS Position                    : 55 deg 58' 24.28" N, 37 deg 24' 41.70" E
```
i am getting an output in fairly readable format, but when i search for it on https://www.openstreetmap.org/ nothing can be found.
I was thinking whether i should go straight for a github, but in the essence i think that's what open-source projects should be working like.

### Description

So my idea is to support the exiftool location format. I was also considering to put the request in my personal text file to come back straight with a pull request, but i got used to luxuries of having an issue request tracker so i decided to put it in open. I would appreciate any comments describing why is this not a good idea before i start to implement that myself.
Thanks!

tomhughes:
Well that is a completely whacky and obscure format that I've never seen anything else use so I would be reluctant to add a special case for one obscure piece of software - the DMS parsing is already insanely complex as it is.

pnorman:
exiftool can output in a format of your choice with the `-coordFormat` option. I recommend using that to get the format in a more sane format.

shdpl:
@tomhughes well, the way i made it work is only a modification of a single regular expression. Ultimately it's always up to the maintainer whether he wants to merge it or not. I think it would be useful for less technical users of those, otherwise, you can make *exiftool* output already recognized format with `exiftool -c "%.6f" my/file.png`.

tomhughes:
Yes I know it's only one regular expression - one already very complicated regular expression as demonstrated by the fact that there are something like 250 lines of code dedicated to testing that regular expression.

gravitystorm:
As per #4041, I'm going to close this issue for now.

PR #4041, opened 2023-05-22
Allow location to be searched with exiftool format #4040

shdpl:
Solves the issue described on https://github.com/openstreetmap/openstreetmap-website/issues/4040

### DIFF
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -205,7 +205,7 @@ class GeocoderController < ApplicationController
         params.merge!(ddm_to_decdeg(latlon)).delete(:query)
 
       elsif latlon = query.match(/^([NS])\s*(\d{1,3})°?\s*(\d{1,2})['′]?(?:\s*(\d{1,3}(\.\d*)?)?["″]?)?\W*([EW])\s*(\d{1,3})°?\s*(\d{1,2})['′]?(?:\s*(\d{1,3}(\.\d*)?)?["″]?)?$/).try(:captures) || # [NSEW] degrees, minutes, decimal seconds
-                     query.match(/^(\d{1,3})°?\s*(\d{1,2})['′]?(?:\s*(\d{1,3}(\.\d*)?)?["″]?)?\s*([NS])\W*(\d{1,3})°?\s*(\d{1,2})['′]?(?:\s*(\d{1,3}(\.\d*)?)?["″]?)?\s*([EW])$/).try(:captures)    # degrees, minutes, decimal seconds [NSEW]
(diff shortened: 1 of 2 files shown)


pnorman (changes requested):
This needs to modify the tests as well as the regular expression.

shdpl:
I hope that's what you ment.

To run the tests on docker I had to change PostgreSQL version from 11 to 12 and update *docker/postgres/openstreetmap-postgres-init.sh* with `createdb -E UTF-8 -O openstreetmap osm_test`, but I guess that you're doing something related to docker, so I'm not  creating any tickets.

gravitystorm:
@shdpl Thank you for your pull request.

I'm not a fan of making a super-complex regexp even more complicated, to support what I suspect might be quite a niche activity for possibly very small number of mappers!

In any case, this fails the rubocop test for line length (and the current limit of 248 is only a "TODO", not a reasonable length in itself) which is a good indication that a better solution is required.

My main concern is to make this code more maintainable, and that means more understandable. I had to investigate what `(?:` means (non-capturing group) and I certainly would struggle to explain what the whole regexp does if someone else asks me to explain it. So when someone (inevitably) comes along in the future and needs a change to or bugfix for that regexp, is it going to be easier or harder for them to do so? While only a small change, I think this makes it even harder.

I'd prefer to see that part of the code refactored, for improved readability. For example, I like the approach used by logstash with its grok patterns, which uses interpolation to build up long complex regexps from component parts (see [example](https://github.com/logstash-plugins/logstash-patterns-core/blob/main/patterns/legacy/grok-patterns)). It's possible to use interpolation within regexps in ruby too. This would allow creating some basic parts of the regexps (e.g. the commonly used `\d{1,3}(\.\d*)?` pattern), giving them clear names, and building the final regexps out of multiple named components. That would make it much easier to understand and add more supported formats.

I also suspect that refactoring that parsing code out of the controller and into a library might make it easier to test and maintain.

Overall, however, without a more compelling use-case I personally don't think any of this is worth the effort. I can see how loading images into an editor (e.g. iD, JOSM) can be super useful but I'm not sure how often mappers will be using exiftool and pasting the results into the search box.

So I'll leave it to you. This PR can't be merged as-is, due to the line length. If you want to add another format, then the code needs to be refactored first, probably in a separate PR, to be much easier to read, understand, and maintain (and easier to review PRs too! :smile: )

shdpl:
Maybe we should just close the PR with the related ticket and wait if there will be any more people interested in that?
I don't know if there's a point in making a refactoring for a sceptically received feature, especially since I'm not a Ruby developer.
I think it makes sense to just validate the usefulness by just waiting for someone to be interested in that as well.

gravitystorm:
> Maybe we should just close the PR with the related ticket and wait if there will be any more people interested in that?

OK, let's do that for now, we can always reopen again later if necessary.

PR #4042, opened 2023-05-23
Add OpenMapTiles vector map

zdila:
This pull request adds OpenMapTiles vector tiles among the featured layers. To ensure compatibility with the current stack based on Leaflet, it uses maplibre-gl-leaflet binding. 
The main advantage for users is support for labels in multiple languages: users can easily switch by changing language preferences in settings.

196365626-243c978f-256c-4a53-a6a7-aee2d63f6c74

An official email with a link to this PR was sent, together with an [edit of the “Featured tile layers” wiki page](https://wiki.openstreetmap.org/wiki/Featured_tile_layers).

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -1,3 +1,7 @@
+//= require maplibre-gl
+//= require @maplibre/maplibre-gl-leaflet
+//= require @maptiler/maplibre-gl-omt-language
+//= require i18n
 //= require qs/dist/qs
 
 L.extend(L.LatLngBounds.prototype, {
@@ -8,6 +12,48 @@ L.extend(L.LatLngBounds.prototype, {
 
   wrap: function () {
     return new L.LatLngBounds(this._southWest.wrap(), this._northEast.wrap());
+  }
+});
+
+L.MapTiler = L.MaplibreGL.extend({
+  options: {
+    maxZoom: 23,
+    attribution: "© <a href=\"https://www.openstreetmap.org/copyright\" target=\"_blank\">OpenStreetMap</a> contributors. Tiles courtesy of <a href=\"http://memomaps.de/\" target=\"_blank\">MeMoMaps</a>"
+  },
+  onAdd: function (map) {
+    function update3dMapUrl() {
+      var el = document.getElementById("terrain_3d_url");
+
+      if (el) {
+        var c = map.getCenter();
+        var zoom = map.getZoom() - 1;
+        var precision = Math.ceil(((zoom * Math.LN2) + Math.log(256 / 360)) / Math.LN10);
+
+        el.href = el.dataset.hrefTemplate
+          .replace("{lang}", I18n.locale.replace(/-.*/, ""))
+          .replace("{pos}", zoom + "/" + c.lat.toFixed(precision) + "/" + c.lng.toFixed(precision) + "/30.00/60.00");
+      }
+    }
+
+    map._update3dMapUrl = update3dMapUrl;
+    setTimeout(update3dMapUrl);
+    map.on("moveend", update3dMapUrl);
+    L.MaplibreGL.prototype.onAdd.call(this, map);
+    var m = this.getMaplibreMap();
+    m.on("load", function () {
+      m.setLanguage(I18n.locale.replace(/-.*/, ""));
+    });
+    L.MaplibreGL.prototype._update.call(this, map);
+  },
+  onRemove: function (map) {
+    L.MaplibreGL.prototype.onRemove.call(this, map);
+    map.off("moveend", map._update3dMapUrl);
+  }
+});
+
+L.OpenMapTiles = L.MapTiler.extend({
+  options: {
+    style: "https://api.maptiler.com/maps/openstreetmap/style.json?key=lmYA16sOOOz9r6DH7iA9"
   }
 });
 
@@ -65,6 +111,12 @@ L.OSM.Map = L.Map.extend({
     }).prop("outerHTML");
     var hotosm = I18n.t("javascripts.map.hotosm_credit", { hotosm_link: hotosm_link, osm_france_link: osm_france_link });
 
+    var openmaptiles_link = I18n.t("javascripts.map.openmaptiles", {
+      openmaptiles_url: "https://openmaptiles.org/",
+      maptiler_url: "https://www.maptiler.com/",
+      terrain_3d_url: "https://labs.maptiler.com/showcase/osm-3d-terrain/#style=openstreetmap&lang={lang}&mode=3d&position={pos}"
+    });
+
     this.baseLayers = [];
 
     this.baseLayers.push(new L.OSM.Mapnik({
@@ -72,6 +124,13 @@ L.OSM.Map = L.Map.extend({
       code: "M",
       keyid: "mapnik",
       name: I18n.t("javascripts.map.base.standard")
+    }));
+
+    this.baseLayers.push(new L.OpenMapTiles({
+      attribution: copyright + ". " + openmaptiles_link + ". " + terms,
+      code: "V",
+      keyid: "openmaptiles_osm",
+      name: I18n.t("javascripts.map.base.openmaptiles_osm")
     }));
 
     this.baseLayers.push(new L.OSM.CyclOSM({
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2953,6 +2953,7 @@ en:
         transport_map: Transport Map
         hot: Humanitarian
         opnvkarte: ÖPNVKarte
+        openmaptiles_osm: OpenMapTiles
       layers:
         header: Map Layers
         notes: Map Notes
@@ -2973,6 +2974,8 @@ en:
       memomaps: MeMoMaps
       hotosm_credit: "Tiles style by %{hotosm_link} hosted by %{osm_france_link}"
       hotosm_name: Humanitarian OpenStreetMap Team
+      openmaptiles: "Vector tiles from <a href='%{openmaptiles_url}' target='_blank'>OpenMapTiles</a> hosted by <a href='%{maptiler_url}' target='_blank'>MapTiler</a>. <a id='terrain_3d_url' data-href-template='%{terrain_3d_url}' target='_blank'>3D map ⛰️</a>"
+      maptiler: "Vector tiles from <a href='%{openmaptiles_url}' target='_blank'>OpenMapTiles</a> hosted by <a href='%{maptiler_url}' target='_blank'>MapTiler</a>. <a id='terrain_3d_url' data-href-template='%{terrain_3d_url}' target='_blank'>3D map ⛰️</a>"
     site:
       edit_tooltip: Edit the map
       edit_disabled_tooltip: Zoom in to edit the map
--- a/package.json
+++ b/package.json
@@ -2,10 +2,13 @@
   "name": "openstreetmap",
   "private": true,
   "dependencies": {
+    "@maplibre/maplibre-gl-leaflet": "^0.0.17",
+    "@maptiler/maplibre-gl-omt-language": "^0.0.1",
     "bs-custom-file-input": "^1.3.4",
     "jquery-simulate": "^1.0.2",
     "js-cookie": "^3.0.0",
     "leaflet": "^1.8.0",
+    "maplibre-gl": "^2.4.0",
     "leaflet.locatecontrol": "^0.79.0",
     "osm-community-index": "^5.2.0",
     "qs": "^6.9.4"
--- a/yarn.lock
+++ b/yarn.lock
@@ -53,6 +53,61 @@
   resolved "https://registry.yarnpkg.com/@humanwhocodes/object-schema/-/object-schema-1.2.1.tgz#b520529ec21d8e5945a1851dfd1c32e94e39ff45"
   integrity sha512-ZnQMnLV4e7hDlUvw8H+U8ASL02SS2Gn6+9Ac3wGGLIe7+je2AeAOxPY+izIPJDfFDb7eDjev0Us8MO1iFRN8hA==
 
+"@mapbox/geojson-rewind@^0.5.2":
+  version "0.5.2"
+  resolved "https://registry.yarnpkg.com/@mapbox/geojson-rewind/-/geojson-rewind-0.5.2.tgz#591a5d71a9cd1da1a0bf3420b3bea31b0fc7946a"
+  integrity sha512-tJaT+RbYGJYStt7wI3cq4Nl4SXxG8W7JDG5DMJu97V25RnbNg3QtQtf+KD+VLjNpWKYsRvXDNmNrBgEETr1ifA==
+  dependencies:
+    get-stream "^6.0.1"
+    minimist "^1.2.6"
+
+"@mapbox/jsonlint-lines-primitives@^2.0.2":
+  version "2.0.2"
+  resolved "https://registry.yarnpkg.com/@mapbox/jsonlint-lines-primitives/-/jsonlint-lines-primitives-2.0.2.tgz#ce56e539f83552b58d10d672ea4d6fc9adc7b234"
+  integrity sha512-rY0o9A5ECsTQRVhv7tL/OyDpGAoUB4tTvLiW1DSzQGq4bvTPhNw1VpSNjDJc5GFZ2XuyOtSWSVN05qOtcD71qQ==
+
+"@mapbox/mapbox-gl-supported@^2.0.1":
+  version "2.0.1"
+  resolved "https://registry.yarnpkg.com/@mapbox/mapbox-gl-supported/-/mapbox-gl-supported-2.0.1.tgz#c15367178d8bfe4765e6b47b542fe821ce259c7b"
+  integrity sha512-HP6XvfNIzfoMVfyGjBckjiAOQK9WfX0ywdLubuPMPv+Vqf5fj0uCbgBQYpiqcWZT6cbyyRnTSXDheT1ugvF6UQ==
+
+"@mapbox/point-geometry@0.1.0", "@mapbox/point-geometry@^0.1.0", "@mapbox/point-geometry@~0.1.0":
+  version "0.1.0"
+  resolved "https://registry.yarnpkg.com/@mapbox/point-geometry/-/point-geometry-0.1.0.tgz#8a83f9335c7860effa2eeeca254332aa0aeed8f2"
+  integrity sha512-6j56HdLTwWGO0fJPlrZtdU/B13q8Uwmo18Ck2GnGgN9PCFyKTZ3UbXeEdRFh18i9XQ92eH2VdtpJHpBD3aripQ==
+
+"@mapbox/tiny-sdf@^2.0.5":
+  version "2.0.6"
+  resolved "https://registry.yarnpkg.com/@mapbox/tiny-sdf/-/tiny-sdf-2.0.6.tgz#9a1d33e5018093e88f6a4df2343e886056287282"
+  integrity sha512-qMqa27TLw+ZQz5Jk+RcwZGH7BQf5G/TrutJhspsca/3SHwmgKQ1iq+d3Jxz5oysPVYTGP6aXxCo5Lk9Er6YBAA==
+
+"@mapbox/unitbezier@^0.0.1":
+  version "0.0.1"
+  resolved "https://registry.yarnpkg.com/@mapbox/unitbezier/-/unitbezier-0.0.1.tgz#d32deb66c7177e9e9dfc3bbd697083e2e657ff01"
+  integrity sha512-nMkuDXFv60aBr9soUG5q+GvZYL+2KZHVvsqFCzqnkGEf46U2fvmytHaEVc1/YZbiLn8X+eR3QzX1+dwDO1lxlw==
+
+"@mapbox/vector-tile@^1.3.1":
+  version "1.3.1"
+  resolved "https://registry.yarnpkg.com/@mapbox/vector-tile/-/vector-tile-1.3.1.tgz#d3a74c90402d06e89ec66de49ec817ff53409666"
+  integrity sha512-MCEddb8u44/xfQ3oD+Srl/tNcQoqTw3goGk2oLsrFxOTc3dUp+kAnby3PvAeeBYSMSjSPD1nd1AJA6W49WnoUw==
+  dependencies:
+    "@mapbox/point-geometry" "~0.1.0"
+
+"@mapbox/whoots-js@^3.1.0":
+  version "3.1.0"
+  resolved "https://registry.yarnpkg.com/@mapbox/whoots-js/-/whoots-js-3.1.0.tgz#497c67a1cef50d1a2459ba60f315e448d2ad87fe"
+  integrity sha512-Es6WcD0nO5l+2BOQS4uLfNPYQaNDfbot3X1XUoloz+x0mPDS3eeORZJl06HXjwBG1fOGwCRnzK88LMdxKRrd6Q==
+
+"@maplibre/maplibre-gl-leaflet@^0.0.17":
+  version "0.0.17"
+  resolved "https://registry.yarnpkg.com/@maplibre/maplibre-gl-leaflet/-/maplibre-gl-leaflet-0.0.17.tgz#1557f413efc53b43f3a9eac3960516d2a7ee714a"
+  integrity sha512-KEsGwX+PfpS2bPaZGf1jG6Awolg/Z4h/UuAM9qAll/2lZ4U+OdjC6x38sWj2bSW5PtHIWBmgPSIc2fFLooa3dA==
+
+"@maptiler/maplibre-gl-omt-language@^0.0.1":
+  version "0.0.1"
+  resolved "https://registry.yarnpkg.com/@maptiler/maplibre-gl-omt-language/-/maplibre-gl-omt-language-0.0.1.tgz#f698ba06b31e7f522a5dc02825742d8b8834c1b8"
+  integrity sha512-OSAJgWvGQxdt6ndUPfMo+bkn2UP2x38tY5/DFTnOrML6Cubdgo+S/5qhas5sd63fTTiouQv+E74Ubr4FDmAyGw==
+
 "@nodelib/fs.scandir@2.1.5":
   version "2.1.5"
   resolved "https://registry.yarnpkg.com/@nodelib/fs.scandir/-/fs.scandir-2.1.5.tgz#7619c2eb21b25483f6d167548b4cfd5a7488c3d5"
@@ -73,6 +128,30 @@
   dependencies:
     "@nodelib/fs.scandir" "2.1.5"
     fastq "^1.6.0"
+
+"@types/geojson@*", "@types/geojson@^7946.0.10":
+  version "7946.0.10"
+  resolved "https://registry.yarnpkg.com/@types/geojson/-/geojson-7946.0.10.tgz#6dfbf5ea17142f7f9a043809f1cd4c448cb68249"
+  integrity sha512-Nmh0K3iWQJzniTuPRcJn5hxXkfB1T1pgB89SBig5PlJQU5yocazeu4jATJlaA0GYFKWMqDdvYemoSnF2pXgLVA==
+
+"@types/mapbox__point-geometry@*", "@types/mapbox__point-geometry@^0.1.2":
+  version "0.1.2"
+  resolved "https://registry.yarnpkg.com/@types/mapbox__point-geometry/-/mapbox__point-geometry-0.1.2.tgz#488a9b76e8457d6792ea2504cdd4ecdd9860a27e"
+  integrity sha512-D0lgCq+3VWV85ey1MZVkE8ZveyuvW5VAfuahVTQRpXFQTxw03SuIf1/K4UQ87MMIXVKzpFjXFiFMZzLj2kU+iA==
+
+"@types/mapbox__vector-tile@^1.3.0":
+  version "1.3.0"
+  resolved "https://registry.yarnpkg.com/@types/mapbox__vector-tile/-/mapbox__vector-tile-1.3.0.tgz#8fa1379dbaead1e1b639b8d96cfd174404c379d6"
+  integrity sha512-kDwVreQO5V4c8yAxzZVQLE5tyWF+IPToAanloQaSnwfXmIcJ7cyOrv8z4Ft4y7PsLYmhWXmON8MBV8RX0Rgr8g==
+  dependencies:
+    "@types/geojson" "*"
+    "@types/mapbox__point-geometry" "*"
+    "@types/pbf" "*"
+
+"@types/pbf@*", "@types/pbf@^3.0.2":
+  version "3.0.2"
+  resolved "https://registry.yarnpkg.com/@types/pbf/-/pbf-3.0.2.tgz#8d291ad68b4b8c533e96c174a2e3e6399a59ed61"
+  integrity sha512-EDrLIPaPXOZqDjrkzxxbX7UlJSeQVgah3i0aA4pOSzmK9zq3BIh7/MZIQxED7slJByvKM4Gc6Hypyu2lJzh3SQ==
 
 acorn-jsx@^5.3.2:
   version "5.3.2"
@@ -176,6 +255,11 @@ cross-spawn@^7.0.2:
     shebang-command "^2.0.0"
     which "^2.0.1"
 
+csscolorparser@~1.0.3:
+  version "1.0.3"
+  resolved "https://registry.yarnpkg.com/csscolorparser/-/csscolorparser-1.0.3.tgz#b34f391eea4da8f3e98231e2ccd8df9c041f171b"
+  integrity sha512-umPSgYwZkdFoUrH5hIq5kf0wPSXiro51nPw0j2K/c83KflkPSTBGMz6NJvMB+07VlL0y7VPo6QJcDjcgKTTm3w==
+
 debug@^4.1.1, debug@^4.3.2:
   version "4.3.4"
   resolved "https://registry.yarnpkg.com/debug/-/debug-4.3.4.tgz#1319f6579357f2338d3337d2cdd4914bb5dcc865"
@@ -199,6 +283,11 @@ doctrine@^3.0.0:
   integrity sha512-yS+Q5i3hBf7GBkd4KG8a7eBNNWNGLTaEwwYWUijIYM7zrlYDM0BFXHjjPWlWZ1Rg7UaddZeIDmi9jF3HmqiQ2w==
   dependencies:
     esutils "^2.0.2"
+
+earcut@^2.2.4:
+  version "2.2.4"
+  resolved "https://registry.yarnpkg.com/earcut/-/earcut-2.2.4.tgz#6d02fd4d68160c114825d06890a92ecaae60343a"
+  integrity sha512-/pjZsA1b4RPHbeWZQn66SWS8nZZWLQQ23oE3Eam7aroEFGEvwKAsJfZ9ytiEMycfzXWpca4FA9QIOehf7PocBQ==
 
 escape-string-regexp@^4.0.0:
   version "4.0.0"
@@ -356,6 +445,11 @@ function-bind@^1.1.1:
   resolved "https://registry.yarnpkg.com/function-bind/-/function-bind-1.1.1.tgz#a56899d3ea3c9bab874bb9773b7c5ede92f4895d"
   integrity sha512-yIovAzMX49sF8Yl58fSCWJ5svSLuaibPxXQJFLmBObTuCr0Mf1KiPopGM9NiFjiYBCbfaa2Fh6breQ6ANVTI0A==
 
+geojson-vt@^3.2.1:
+  version "3.2.1"
+  resolved "https://registry.yarnpkg.com/geojson-vt/-/geojson-vt-3.2.1.tgz#f8adb614d2c1d3f6ee7c4265cad4bbf3ad60c8b7"
+  integrity sha512-EvGQQi/zPrDA6zr6BnJD/YhwAkBP8nnJ9emh3EnHQKVMfg/MRVtPbMYdgVy/IaEmn4UfagD2a6fafPDL5hbtwg==
+
 get-intrinsic@^1.0.2:
   version "1.1.3"
   resolved "https://registry.yarnpkg.com/get-intrinsic/-/get-intrinsic-1.1.3.tgz#063c84329ad93e83893c7f4f243ef63ffa351385"
@@ -364,6 +458,16 @@ get-intrinsic@^1.0.2:
     function-bind "^1.1.1"
     has "^1.0.3"
     has-symbols "^1.0.3"
+
+get-stream@^6.0.1:
+  version "6.0.1"
+  resolved "https://registry.yarnpkg.com/get-stream/-/get-stream-6.0.1.tgz#a262d8eef67aced57c2852ad6167526a43cbf7b7"
+  integrity sha512-ts6Wi+2j3jQjqi70w5AlN8DFnkSwC+MqmxEzdEALB2qXZYV3X/b1CTfgPLGJNMeAWxdPfU8FO1ms3NUfaHCPYg==
+
+gl-matrix@^3.4.3:
+  version "3.4.3"
+  resolved "https://registry.yarnpkg.com/gl-matrix/-/gl-matrix-3.4.3.tgz#fc1191e8320009fd4d20e9339595c6041ddc22c9"
+  integrity sha512-wcCp8vu8FT22BnvKVPjXa/ICBWRq/zjFfdofZy1WSpQZpphblv12/bOQLBC1rMM7SGOFS9ltVmKOHil5+Ml7gA==
 
 glob-parent@^6.0.2:
   version "6.0.2"
@@ -383,6 +487,15 @@ glob@^7.1.3:
     minimatch "^3.1.1"
     once "^1.3.0"
     path-is-absolute "^1.0.0"
+
+global-prefix@^3.0.0:
+  version "3.0.0"
+  resolved "https://registry.yarnpkg.com/global-prefix/-/global-prefix-3.0.0.tgz#fc85f73064df69f50421f47f883fe5b913ba9b97"
+  integrity sha512-awConJSVCHVGND6x3tmMaKcQvwXLhjdkmomy2W+Goaui8YPgYgXJZewhg3fWC+DlfqqQuWg8AwqjGTD2nAPVWg==
+  dependencies:
+    ini "^1.3.5"
+    kind-of "^6.0.2"
+    which "^1.3.1"
 
 globals@^13.19.0:
   version "13.19.0"
@@ -412,6 +525,11 @@ has@^1.0.3:
   integrity sha512-f2dvO0VU6Oej7RkWJGrehjbzMAjFp5/VKPp5tTpWIV4JHHZK1/BxbFRtf/siA2SWTe09caDmVtYYzWEIbBS4zw==
   dependencies:
     function-bind "^1.1.1"
+
+ieee754@^1.1.12:
+  version "1.2.1"
+  resolved "https://registry.yarnpkg.com/ieee754/-/ieee754-1.2.1.tgz#8eb7a10a63fff25d15a57b001586d177d1b0d352"
+  integrity sha512-dcyqhDvX1C46lXZcVqCpK+FtMRQVdIMN6/Df5js2zouUsqG7I6sFxitIC+7KYK29KdXOLHdu9zL4sFnoVQnqaA==
 
 ignore@^5.2.0:
   version "5.2.0"
@@ -443,6 +561,11 @@ inherits@2:
   version "2.0.4"
   resolved "https://registry.yarnpkg.com/inherits/-/inherits-2.0.4.tgz#0fa2c64f932917c3433a0ded55363aae37416b7c"
   integrity sha512-k/vGaX4/Yla3WzyMCvTQOXYeIHvqOKtnqBduzTHpzpQZzAskKMhZ2K+EnBiSM9zGSoIFeMpXKxa4dYeZIQqewQ==
+
+ini@^1.3.5:
+  version "1.3.8"
+  resolved "https://registry.yarnpkg.com/ini/-/ini-1.3.8.tgz#a29da425b48806f34767a4efce397269af28432c"
+  integrity sha512-JV/yugV2uzW5iMRSiZAyDtQd+nxtUnjeLt0acNdw98kKLrvuRVyB80tsREOE7yvGVgalhZ6RNXCmEHkUKBKxew==
 
 is-extglob@^2.1.1:
   version "2.1.1"
@@ -493,6 +616,16 @@ json-stable-stringify-without-jsonify@^1.0.1:
   resolved "https://registry.yarnpkg.com/json-stable-stringify-without-jsonify/-/json-stable-stringify-without-jsonify-1.0.1.tgz#9db7b59496ad3f3cfef30a75142d2d930ad72651"
   integrity sha512-Bdboy+l7tA3OGW6FjyFHWkP5LuByj1Tk33Ljyq0axyzdk9//JSi2u3fP1QSmd1KNwq6VOKYGlAu87CisVir6Pw==
 
+kdbush@^3.0.0:
+  version "3.0.0"
+  resolved "https://registry.yarnpkg.com/kdbush/-/kdbush-3.0.0.tgz#f8484794d47004cc2d85ed3a79353dbe0abc2bf0"
+  integrity sha512-hRkd6/XW4HTsA9vjVpY9tuXJYLSlelnkTmVFu4M9/7MIYQtFcHpbugAU7UbOfjOiVSVYl2fqgBuJ32JUmRo5Ew==
+
+kind-of@^6.0.2:
+  version "6.0.3"
+  resolved "https://registry.yarnpkg.com/kind-of/-/kind-of-6.0.3.tgz#07c05034a6c349fa06e24fa35aa76db4580ce4dd"
+  integrity sha512-dcS1ul+9tmeD95T+x28/ehLgd9mENa3LsvDTtzm3vyBEO7RPptvAD+t44WVXaUjTBRcrpFeFlC8WCruUR456hw==
+
 leaflet.locatecontrol@^0.79.0:
   version "0.79.0"
   resolved "https://registry.yarnpkg.com/leaflet.locatecontrol/-/leaflet.locatecontrol-0.79.0.tgz#0236b87c699a49f9ddb2f289941fbc0d3c3f8b62"
@@ -523,6 +656,36 @@ lodash.merge@^4.6.2:
   resolved "https://registry.yarnpkg.com/lodash.merge/-/lodash.merge-4.6.2.tgz#558aa53b43b661e1925a0afdfa36a9a1085fe57a"
   integrity sha512-0KpjqXRVvrYyCsX1swR/XTK0va6VQkQM6MNo7PqW77ByjAhoARA8EfrP1N4+KlKj8YS0ZUCtRT/YUuhyYDujIQ==
 
+maplibre-gl@^2.4.0:
+  version "2.4.0"
+  resolved "https://registry.yarnpkg.com/maplibre-gl/-/maplibre-gl-2.4.0.tgz#2b53dbf526626bf4ee92ad4f33f13ef09e5af182"
+  integrity sha512-csNFylzntPmHWidczfgCZpvbTSmhaWvLRj9e1ezUDBEPizGgshgm3ea1T5TCNEEBq0roauu7BPuRZjA3wO4KqA==
+  dependencies:
+    "@mapbox/geojson-rewind" "^0.5.2"
+    "@mapbox/jsonlint-lines-primitives" "^2.0.2"
+    "@mapbox/mapbox-gl-supported" "^2.0.1"
+    "@mapbox/point-geometry" "^0.1.0"
+    "@mapbox/tiny-sdf" "^2.0.5"
+    "@mapbox/unitbezier" "^0.0.1"
+    "@mapbox/vector-tile" "^1.3.1"
+    "@mapbox/whoots-js" "^3.1.0"
+    "@types/geojson" "^7946.0.10"
+    "@types/mapbox__point-geometry" "^0.1.2"
+    "@types/mapbox__vector-tile" "^1.3.0"
+    "@types/pbf" "^3.0.2"
+    csscolorparser "~1.0.3"
+    earcut "^2.2.4"
+    geojson-vt "^3.2.1"
+    gl-matrix "^3.4.3"
+    global-prefix "^3.0.0"
+    murmurhash-js "^1.0.0"
+    pbf "^3.2.1"
+    potpack "^1.0.2"
+    quickselect "^2.0.0"
+    supercluster "^7.1.5"
+    tinyqueue "^2.0.3"
+    vt-pbf "^3.1.3"
+
 minimatch@^3.0.5, minimatch@^3.1.1, minimatch@^3.1.2:
   version "3.1.2"
   resolved "https://registry.yarnpkg.com/minimatch/-/minimatch-3.1.2.tgz#19cd194bfd3e428f049a70817c038d89ab4be35b"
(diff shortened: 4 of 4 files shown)


tomhughes:
I thought we had agreed that adding support for vector tile layers should be separate from adding any specific layer - certainly at the least I'd like them to be separate commits.

The API key should probably be in the configuration file as well, as with the thunderforest layers, rather than hardcoded.

What are all the 3D terrain bits about?

zdila:
> I thought we had agreed that adding support for vector tile layers should be separate from adding any specific layer - certainly at the least I'd like them to be separate commits.

Support for vector tiles is actually just adding `maplibre-gl` and `@maplibre/maplibre-gl-leaflet` to `package.json`.

If you agree, I can prepare such PR.

> The API key should probably be in the configuration file as well, as with the thunderforest layers, rather than hardcoded.

I'll look at it.

> What are all the 3D terrain bits about?

It is a link to a demo site that shows full capability of vector maps (map rotation, animation, 3D terrain) which is not possible with Leaflet and maplibre-gl-leaflet library.

zdila:
This PR adds only one layer - vector map with carto-like style. Most of the code is actually about updating link URL in the attribution of this map so thet it would point to the same view of external 3D map demo.

zdila:
You can see it live at http://osm.openmaptiles.org/

tomhughes:
Ah yes it does work over HTTP but all the other links were HTTPS which doesn't work.

I think the 3D stuff should be separated out - most of the code is only actually there to do that I think and I think that's going to be a lot more contentious that the rest of it.

tomhughes:
Yes that's what I'm talking about - it wouldn't be acceptable as a layer on the site anyway as it doesn't appear to cover the whole world and I think that also makes linking to it questionable which is why I'd rather separate that decision/discussion from the question of adding the vector tile layer to the site.

zdila:
Can you please report an area that is missing terrain/map? Because I am checking it on my side and the coverage is global - even in the remote places.

tomhughes:
Oh if I hold on that button and move my mouse I can make it tilt and pan...

zdila:
> This is great! Do you have any guidance for us as to how it handles regional languages, e.g. `pt-PT` vs `pt-BR`? Or `sr` vs `sr-Latn`? And what happens if you pick a language that is not supported by GL-based renderings, like `my` (Burmese)?

Supported languages of OMT are documented at https://docs.maptiler.com/schema/planet/#languages

<details>
<summary>Comparison</summary>

|OSM|OMT|Language|
|---|---|--------|
|af||Afrikaans|
|aln||Gheg Albanian|
||am|Amharic|
|ar|ar|Arabic|
|arz||Egyptian Arabic|
|ast||Asturian|
|az|az|Azerbaijani|
|ba||Bashkir|
|be-Tarask||Belarusian (Taraškievica)|
|be|be|Belarusian|
|bg|bg|Bulgarian|
|bn||Bengali|
|br|br|Breton|
|bs|bs|Bosnian|
|ca|ca|Catalan|
|ce|co|Chechen, Corsican|
|cs|cs|Czech|
|cy|cy|Welsh|
|da|da|Danish|
|de|de|German|
|diq||Zazaki|
|dsb||Lower Sorbian|
|el|el|Greek|
|en-GB||English (UK)|
|en|en|English|
|eo|eo|Esperanto|
|es|es|Spanish|
|et|et|Estonian|
|eu|eu|Basque|
|fa||Persian|
|fi|fi|Finnish|
|fit||Tornedalen Finnish|
|fr|fr|French|
|fur||Friulian|
|fy|fy|West Frisian|
|ga|ga|Irish|
|gcf||Guadeloupean Creole French|
|gd|gd|Scottish Gaelic|
|gl||Galician|
|gsw||Swiss German|
|he|he|Hebrew|
|hi|hi|Hindi|
|hr|hr|Croatian|
|hsb||Upper Sorbian|
|hu|hu|Hungarian|
||hy|Armenian|
|ia||Interlingua|
|id|id|Indonesian|
|is|is|Icelandic|
|it|it|Italian|
|ja|ja|Japanese|
||ja-Hira|Japanese (Hiragana)|
||ja-Latn|Japanese (Latin)|
||ja_rm|Japanese (Romanization)|
||ja_kana|Japanese (Kana)|
|ka|ka|Georgian|
|kab||Kabyle|
|kk-cyrl|kk|Kazakh (Cyrillic)|
|km||Khmer|
|kn|kn|Kannada|
|ko|ko|Korean|
||ko-Latn|Korean (Latin)|
|ksh||Colognian|
|ku-Latn||Kurdish (Latin)|
||ku|Kurdish|
||la|Latin|
|lb|lb|Luxembourgish|
|lt|lt|Lithuanian|
|lv|lv|Latvian|
|mk|mk|Macedonian|
||ml|Malayalam|
|mo||Moldovan|
|mr||Marathi|
|ms||Malay|
||mt|Maltese|
|my||Burmese|
|nb||Norwegian Bokmål|
|nds||Low German|
|ne||Nepali|
|nl|nl|Dutch|
|nn||Norwegian Nynorsk|
||no|Norwegian|
|nqo||N’Ko|
|oc|oc|Occitan|
|pa||Punjabi|
|pl|pl|Polish|
|pnb||Western Punjabi|
|ps||Pashto|
|pt-BR||Portuguese (Brazil)|
|pt-PT||Portuguese (Portugal)|
|pt|pt|Portuguese|
||rm|Romansh|
|ro|ro|Romanian|
|ru|ru|Russian|
|sat||Santali|
|sc||Sardinian|
|scn||Sicilian|
|sco||Scots|
|sk|sk|Slovak|
|skr-arab||Saraiki (Arabic)|
|sl|sl|Slovenian|
|sq|sq|Albanian|
|sr-Latn|sr-Latn|Serbian (Latin)|
|sr|sr|Serbian|
|sv|sv|Swedish|
|ta|ta|Tamil|
|te|te|Telugu|
|th|th|Thai|
|tl||Tagalog|
|tr|tr|Turkish|
|tt||Tatar|
|uk|uk|Ukrainian|
||vi|Vietnamese|
|xmf||Mingrelian|
||yi|Yiddish|
||yo|Yoruba|
||zh|Chinese|
|zh-CN||Chinese (China)|
|zh-HK||Chinese (Hong Kong)|
|zh-TW||Chinese (Taiwan)|

</details>

Currently if the language is not supported then local language is used - same as on Carto map - value from the `name` key of the tag.

> I want to make sure that the translations work for all of the [languages that our user interface works in](https://github.com/openstreetmap/openstreetmap-website/tree/master/config/locales), hopefully so that there's no mismatch between what you can pick in the UI and what then happens with this layer.

I will discuss it with my team. In any case current implementation doesn't fallback to the other languages in the list but I will implement it into this PR. Also falling back to main locale if sub-locale is not available (`pt-BR` -> `pt`).

> I also noticed on the wiki that it says "map labels translations from Wikidata" - can you clarify this? How does this interact with the translated labels available in OSM? If a contributor spots a mistake on the map, can they edit OSM to fix it, or do they have to edit something on Wikidata?

We have decided to use names primarily from wikidata because there are more translations than in OSM. If the name is not in wikidata then one from OSM is used.

> > We added it there to allow visitors to see the OSM in 3D. We would love to have it directly on [osm.org](http://osm.org/). However, it is not technically possible because the site runs on Leaflet. Therefore, we added it at least as an external link.
> 
> I don't think the layer attribution is the right place for linking to external sites to view other renderings.
> 
> If we want to link to external sites to view other renderings, which haven't been through the new layer approval process, then let's discuss this separately. And if we want to adapt the site to support alternative (i.e. non-leaflet) tech, then let's discuss that separately too. But I don't think such links should be part of the attribution for another layer, and I definitely don't think that should be part of this PR.

We will adjust it in this PR.

pnorman:
> Currently if the language is not supported then local language is used - same as on Carto map - value from the `name` key of the tag.

What about when the language of the name tag is in a script Maplibre does not support?

zdila:
> What about when the language of the name tag is in a script Maplibre does not support?

Not sure if I understand you correctly, but for unsupported languages their local `name` will be displayed.

If you meant something else then please give me some example. You can try switching to (unsupported) language by setting it as primary in your browser (until I add a fallback to other languages your browser accepts - then try just with a single language configured).

zdila:
Meanwhile I have removed 3D map link in the attribution.

zdila:
> The API key should probably be in the configuration file as well, as with the thunderforest layers, rather than hardcoded

done

jachym:
> Maybe I just don't understand how to operate it then... I clicked the bottom button to switch back to 2D mode, panned to my home area, then tried to switch back to 3D mode by pressing the same button again and nothing happened.
> 
> Specifically I was looking at https://labs.maptiler.com/showcase/osm-3d-terrain/#style=openstreetmap&lang=en&mode=3d&position=15.46/51.76264/-0.00868 which is 2D and I can't find any way to switch to 3D there?

The map is 3D, but London is rather flat, sending better position: 
https://labs.maptiler.com/showcase/osm-3d-terrain/#style=openstreetmap&lang=en&mode=3d&position=12.51/55.4648/-3.3975/-4.80/85.00

As metioned, you hold Ctrl key and can change the view angle.

pnorman:
> Not sure if I understand you correctly, but for unsupported languages their local `name` will be displayed.

Maplibre doesn't support some scripts (e.g. Burmese, Lao, and other Brahmic scripts), so it can't display the local name.

ZeLonewolf:
> Maplibre doesn't support some scripts (e.g. Burmese, Lao, and other Brahmic scripts), so it can't display the local name.

That's not _exactly_ accurate. The issue is that MapLibre doesn't support ligatures and combining characters correctly, and instead displays them anywhere from slightly to terribly wrongly. However, OSM-Carto also renders some of these scripts wrongly as the examples below will show (and should probably be documented if they're not already).  Whether the "wrong" rendering makes the label unreadable or just slightly off will depend heavily on the script involved, and in my experience, it's the worst in Right-to-Left scripts.

This is what MapLibre displays for the United States label, in Lao:

[image](https://zelonewolf.github.io/openstreetmap-americana/#map=4/33.24/-96.49&language=lo)

However, it is supposed to look like this:
[image](https://lo.wikipedia.org/wiki/%E0%BA%AA%E0%BA%B0%E0%BA%AB%E0%BA%B0%E0%BA%A5%E0%BA%B1%E0%BA%94)

The issue is that the squiggly line above the second-to-last character is supposed to be directly over it, not off to one side.  This issue is pervasive across all ligature/combiner scripts. While a Lao reader might still understand the script above, it would definitely look wrong, and this problem in other scripts results in mangled text that's absolutely unreadable to native speakers.

This is something which has _always_ been broken in mapbox-gl and is a high priority for maplibre-gl to get right, and it's documented in maplibre/maplibre#193.

Examples of rendering the Lao capital of Vientiane, which should appear in local script as:

### ວຽງຈັນ

Google Maps:
[image](https://www.google.com/maps/@17.9637196,102.6238214,13z?entry=ttu)

OSM Americana (based on MapLibre):
[image](https://zelonewolf.github.io/openstreetmap-americana/#map=10.92/17.9901/102.6201)

OSM Carto:
image

However, looking at these examples, it seems that osm-carto isn't actually any different from what MapLibre is producing.

Let's take a look at Kathmandu, Nepal, which is supposed to appear as:

### काठमाडौँ

Google Maps:
[image](https://www.google.com/maps/@27.7089543,85.2849333,13z?entry=ttu)

OSM Americana
[image](https://zelonewolf.github.io/openstreetmap-americana/#map=11/27.6802/85.3489&language=en)

OSM Carto:
[[image](https://www.openstreetmap.org/#map=12/27.6858/85.2797)](https://www.openstreetmap.org/#map=11/17.9902/102.6817)


Mandalay, Burma, perhaps?

Should be:

### မန္တလေး

Google Maps:
[image](https://www.google.com/maps/@21.9579896,96.0893805,13z?entry=ttu)

OSM Americana:
[image](https://zelonewolf.github.io/openstreetmap-americana/#map=11/21.9869/96.12&language=en)

OSM Carto:
[image](https://www.openstreetmap.org/#map=11/21.9816/95.9621)

Burmese is where we really start to see the problem, with that infinity-sign looking diacritic that's supposed to be under the second character coming in rather mangled. However, OSM-Carto also gets it slightly wrong by jamming the ligature up into the character when there's supposed to be space between them.

LaoshuBaby:
图片

It seems that this new vector map does not handle the display of Chinese very well. It displays the Traditional and Simplified varient at the same time, because a DWG member ~~with an Asian cultural background~~ combined the two varients in `name:zh`, and usually The user will choose `zh-CN` or `zh-TW` two locales (for example, my browser uses `zh-CN`, so you can see other UI translations are displayed in my locale), but the label text is not correct from `zh-CN` match to `name:zh-Hans`.

SomeoneElseOSM:
Re "... because a DWG member ... combined the two variants in name:zh" that sounds like a tagging issue that's best discussed elsewhere - perhaps somewhere at https://community.openstreetmap.org/ ?

LaoshuBaby:
> Re "... because a DWG member ... combined the two variants in name:zh" that sounds like a tagging issue

Yes, use variants combined value is a tagging issue, we can discuss in other places.

My reply is about vector map's label text(`name:zh`) can't match browser's locale settings(`zh-CN`/`zh-TW`), There should be a fallback from `name:zh-Hans/t`->`name:zh`, I don’t know how the new vector map handles the ICU problem.

app/assets/javascripts/leaflet.map.js
```diff
@@ -11,6 +15,40 @@ L.extend(L.LatLngBounds.prototype, {
...
 });
 
+if (OSM.MAPTILER_KEY) {
+  L.OpenMapTiles = L.MaplibreGL.extend({
+    options: {
+      maxZoom: 23,
+      style: "https://api.maptiler.com/maps/openstreetmap/style.json?key=" + OSM.MAPTILER_KEY
+    },
+    onAdd: function (map) {
+      var supportedLanguages = "am,ar,az,be,bg,br,bs,ca,co,cs,cy,da,de,el,en,eo,es,et,eu,fi,fr,fy,ga,gd,he,hi,hr,hu,hy,id,is,it,ja,ja_kana,ja_rm,ja-Latn,ja-Hira,ka,kk,kn,ko,ko-Latn,ku,la,lb,lt,lv,mk,mt,ml,nl,no,oc,pl,pt,rm,ro,ru,sk,sl,sq,sr,sr-Latn,sv,ta,te,th,tr,uk,zh".split(",");
```
1ec5, line 25: `name:ja_kana` and `name:ja_rm` are both [documented on the wiki as being deprecated](https://wiki.openstreetmap.org/wiki/Multilingual_names#Japan), although they are still in use to some extent. Is it a good idea to support these nonstandard language codes as part of the website?

1ec5, line 25: This list doesn’t include the languages added to OpenMapTiles openmaptiles/openmaptiles#1477. I assume that’s because the MapTiler-hosted tiles are on OpenMapTiles v3.14, whereas this change is still waiting to be released. Is there a way to automate this list so that the website maintainers won’t have to remember to update it based on OpenMapTiles’ release schedule and MapTiler’s upgrade schedule?

tomhughes:
To be clear what languages are used in OSM tags are not directly relevant here - this is a mapping from web site locales to locales used by this layer.

How the layer maps from OSM tags to it's supported languages is a matter for the layer and the tools used to generate it's tiles from the OSM data.

1ec5:
Understood, but those OpenMapTiles properties are directly tied to the deprecated OSM keys, so it would essentially be a way to enable OSM users to enter nonstandard codes in OSM preferences to see those tags’ value. I guess it’s fine if it ends up being an obscure, undocumented edge case, but if it ends up getting used, then that would be problematic for other components that rely on this same preference for localization, such as iD.

Triomphe346:
> How the layer maps from OSM tags to it's supported languages is a matter for the layer and the tools used to generate it's tiles from the OSM data. 

I see on their own demo that they could correctly handle different Chinese variants (attach sample is Simplified Chinese only demo with 'zh-CN' browser setting). 

Demo of MapTiler in Simplified Chinese

It seems that they do not explicitly document this. If it is possible to support `zh-Hans`/`zh-Hant` variants disregarding the listed supported language code? (may need mapping `zh-CN` browser setting to `zh-Hans` in MapTiler & `zh-HK`/`zh-TW` browser setting to `zh-Hant` in MapTiler manually) @tomhughes

app/assets/javascripts/leaflet.map.js
```diff
@@ -11,6 +15,40 @@ L.extend(L.LatLngBounds.prototype, {
   }
 });
 
+if (OSM.MAPTILER_KEY) {
+  L.OpenMapTiles = L.MaplibreGL.extend({
```
1ec5, line 19: Arabic and Hebrew would be broken unless this code also [installs the right-to-left text plugin](https://maplibre.org/maplibre-gl-js-docs/example/mapbox-gl-rtl-text). (Note that it only supports these two languages, not other languages in the same scripts or other right-to-left scripts.)

tomhughes:
Whilst I agreed that the name collision is confusing the chosen name does need to be relatively succinct so "Default OpenMapTiles style" is probably not going to work.

ZeLonewolf:
To be consistent with the other titles, it would be something like Standard (Vector).  After all, it is simply re-creating the osm-carto style in vector technology.

matkoniecz:
It is not 1:1 recreation and it seems fine to include display of OpenMapTiles project name.

mnalis:
> It is not 1:1 recreation 

True, it would likely create confusion if two _"Standard"_ options would display different things.

> and it seems fine to include display of OpenMapTiles project name.

Also true, but neither does current _"Standard"_ include _"Carto"_ name, so it should not be a _requirement_. When clicking on the map, I assume the relevant details would still be included in lower right corner like in other maps? 

But I agree that (while not an absolute requirement) it would be preferable to have style name in there.

> Whilst I agreed that the name collision is confusing the chosen name does need to be relatively succinct

Absolutely agree. Maybe just _"Vector"_ / _"Vectors"_ (if that would not be on _"too short"_ side?)

zdila:
We prefer to call the style OpenMapTiles as we do on GitHub: https://github.com/openmaptiles/openmaptiles/tree/v3.14/style. The reason is that it becomes "The Style" of the OpenMapTiles project.

mnalis:
Soo, are there any blockers remaining for this to be merged?

tomhughes:
OWG needs to approve the layer.

wipfli:
Regarding rendering of text that requires shaping such as Devanagari or Khmer in MapLibre GL JS, I would like to add that this is indeed a priority to introduce as a new feature in the future.

At the moment, we cannot render text in the native language of 1.5 to 2 billion people so we should fix this...

ZeLonewolf:
I would like to better understand MapTiler/OpenMapTiles's plan for regular updates. My understanding is that MapTiler Cloud only updates when OpenMapTiles updates (on the order of many months), however, I understand that the expectation for featured layers is to update at least biweekly.  Will the OpenMapTiles map update on a regular basis, and if so how often?  Will it confuse end users if the OpenMapTiles version of the map is consistently out of date with respect to edits compared to the near-identical osm-carto raster edition?  Just trying to put myself in the shoes of end users that come to openstreetmap.org to learn about the project without necessarily the deep knowledge to appreciate the distinction between raster and vector stacks.

Discostu36:
> We have decided to use names primarily from wikidata because there are more translations than in OSM. If the name is not in wikidata then one from OSM is used.

In my opinion maps on openstreetmap.org should not show third-party data. If a local name is missing, visitors should see that and add them instead of not being able to tell that data is missing.

ZeLonewolf:
> In my opinion maps on openstreetmap.org should not show third-party data. If a local name is missing, visitors should see that and add them instead of not being able to tell that data is missing.

So to be clear, you also have a problem with using NaturalEarth data, which is also an "external data set"?

ZeLonewolf:
OpenMapTiles' use of wikidata is consistent with the featured tile layer requirements, which say:

> The proposed tile layer should not use non-OSM data where OSM data exists.

See: https://wiki.openstreetmap.org/wiki/Featured_tile_layers/Guidelines_for_new_tile_layers

Discostu36:
> So to be clear, you also have a problem with using NaturalEarth data, which is also an "external data set"?

I don't know how and where NaturalEarth data is used, but yes. The main page is the showcase for the OSM database and should only visualize data that is included and can be edited here. But this is only my opinion, this is for the OSMF to decide.

ZeLonewolf:
> I don't know how and where NaturalEarth data is used, but yes. The main page is the showcase for the OSM database and should only visualize data that is included and can be edited here. But this is only my opinion, this is for the OSMF to decide.

Natural Earth is used in the standard tile layer and all of its derivatives.

So basically you are saying that all of the current layers should be shut down.

Discostu36:
> > I don't know how and where NaturalEarth data is used, but yes. The main page is the showcase for the OSM database and should only visualize data that is included and can be edited here. But this is only my opinion, this is for the OSMF to decide.
> 
> Natural Earth is used in the standard tile layer and all of its derivatives.
> 
> So basically you are saying that all of the current layers should be shut down.

I'd prefer if you could not put words into my mouth. I don't know what data from NaturalEarth is used in OSM Carto and for what reason. And this if off-topic here.

ZeLonewolf:
> I don't think that's strictly accurate. According to what has been written above, then if wikidata has a translation, the existing OSM translation name tags are ignored. 
> 
> This is different from saying that if there is no translation in OSM, then wikidata is used to fill the gaps. In particular, it means that if the wikidata translation is wrong, there's nothing an OSM contributor can do about it directly - they will need to go to wikidata to fix the problem.
> 
> I don't think there would be as much concern about this if the names were primarily taken from OSM, and wikidata was only being used as a fallback to fill the gaps.

I don't believe that's a correct description. OSM is the primary and wikidata is a backup if a language tag is missing. If that's not happening it's a bug.

gravitystorm:
I'm taking my knowledge from this exchange, with my bold:

>>   I also noticed on the wiki that it says "map labels translations from Wikidata" - can you clarify this? How does this interact with the translated labels available in OSM? If a contributor spots a mistake on the map, can they edit OSM to fix it, or do they have to edit something on Wikidata?

> We have decided to use names primarily from wikidata because there are more translations than in OSM. **If the name is not in wikidata then one from OSM is used.**

ZeLonewolf:
> > > I don't know how and where NaturalEarth data is used, but yes. The main page is the showcase for the OSM database and should only visualize data that is included and can be edited here. But this is only my opinion, this is for the OSMF to decide.
> > 
> > Natural Earth is used in the standard tile layer and all of its derivatives.
> > 
> > So basically you are saying that all of the current layers should be shut down.
> 
> I'd prefer if you could not put words into my mouth. I don't know what data from NaturalEarth is used in OSM Carto and for what reason. And this if off-topic here.

It is very on topic. All of the current layers use data external to OSM, therefore that should not be a reason to disqualify the proposed OpenMapTiles layer.

Discostu36:
> It is very on topic. All of the current layers use data external to OSM, therefore that should not be a reason to disqualify the proposed OpenMapTiles layer.

No, because there is, in my opinion, no valid reason to use Wikidata data for the labels here, they can be easily added to the OSM data if missing. But there might be a valid reason for the use of NaturalEarth data at Carto that I don't know about. Can we now talk about this specific case and not about Carto?

ZeLonewolf:
> I'm taking my knowledge from this exchange, with my bold:
> 
> >>   I also noticed on the wiki that it says "map labels translations from Wikidata" - can you clarify this? How does this interact with the translated labels available in OSM? If a contributor spots a mistake on the map, can they edit OSM to fix it, or do they have to edit something on Wikidata?
> 
> > We have decided to use names primarily from wikidata because there are more translations than in OSM. **If the name is not in wikidata then one from OSM is used.**

Thanks for that. I agree that's a showstopper. We specifically ran into this in Americana and had to get it fixed upstream in planetiler. Unfortunately I can't find the planetiler ticket on it. At the time preferring wikidata was regarded as a bug so I'm surprised to hear OMT is doing that.

https://github.com/ZeLonewolf/openstreetmap-americana/issues/428

1ec5:
> We specifically ran into this in Americana and had to get it fixed upstream in planetiler. Unfortunately I can't find the planetiler ticket on it. At the time preferring wikidata was regarded as a bug so I'm surprised to hear OMT is doing that.

Planetiler inverted the logic to prefer OSM name tags over Wikidata labels in openmaptiles/planetiler-openmaptiles#55 for consistency with OpenMapTiles, which in turn tried to be consistent with Mapbox Streets: https://github.com/openmaptiles/openmaptiles/issues/251#issuecomment-309886301.[^mapbox]

I think the source of confusion may be that, if a feature has a `name` but no `name:la` in OSM, OpenMapTiles will set `name:la` to the Wikidata item’s Latin label instead of `name`. This is even the case if `name` happens to be a different Latin name, because OpenMapTiles [has no way of knowing](https://wiki.openstreetmap.org/wiki/Proposed_features/Language_information_for_name) it’s in Latin.

[^mapbox]: However, Mapbox actually does unconditionally prefer Wikidata labels over OSM name tags.

matkoniecz:
> All of the current layers use data external to OSM, therefore that should not be a reason to disqualify the proposed OpenMapTiles layer.

The big difference is that other renderers use data that is missing from OSM (terrain shape) or use matching OSM, for performance reasons (Carto).

While this would, according to description ("If the name is not in wikidata then one from OSM is used.") would force people to fix mistakes in Wikidata even where OSM data exists and is correct.

1ec5:
> While this would, according to description ("If the name is not in wikidata then one from OSM is used.") would force people to fix mistakes in Wikidata even where OSM data exists and is correct.

Fortunately, that turned out to be a description of Mapbox Streets rather than of OpenMapTiles. 😉

We don’t have to rely on vague descriptions because we can see the behavior for ourselves. For example, [this airport area in OSM](https://www.openstreetmap.org/way/648903263) has both the `name` and `name:en` keys set to `San José Mineta International Airport`, while [the corresponding Wikidata item](https://www.wikidata.org/wiki/Q1257392) has an English label of “San Jose International Airport”. (Both names are reasonable.) In MapTiler’s OpenMapTiles tileset, the `name` and `name:en` properties are both set to `San José Mineta International Airport`, so an English-speaking user will see the OSM area’s tag value, not the Wikidata item’s label.

By contrast, the OSM area lacks a `name:ja` tag, while the Wikidata item does have a Japanese label of “ノーマン・Y・ミネタ・サンノゼ国際空港”. A Japanese-speaking user will see Wikidata’s Japanese label rather than an OSM `name` tag in an undetermined language (which happens to be in English).

name: San José Mineta International Airport

(This screenshot is of a tile inspector behind a login wall that requires a MapTiler API token. However, you can see the same behavior in @ZeLonewolf’s Planetiler-powered tile server’s inspector by hovering over [this green dot](https://tile.ourmap.us/inspect.html#22/37.36291244/-121.92879588).)

Mapper feedback can go both ways. We could say that this fallback makes it more difficult for mappers to detect gaps in language coverage; if only the Japanese speaker could see the English name, they’d be tempted to add a `name:ja` tag on the spot. In a place as linguistically diverse as San José, no one would have a problem with that edit. But [much ink has been spilled](https://lists.openstreetmap.org/pipermail/talk/2015-June/thread.html#73165) about how translations and transliterations should not be indiscriminately tagged on mundane features in OSM in other places. The Wikidata fallback can serve to mitigate against edits that OSM doesn’t want.

The problem with such prominent placement on osm.org is that the mapper can’t visually determine whether the label comes from OSM or Wikidata, or that there’s even a discrepancy. OpenMapTiles doesn’t currently expose any metadata explicitly indicating the source of a particular name on a particular feature. However, the style could show both the localized and unlocalized names together whenever the two differ.

For example, in the proposed OpenMapTiles layer, a Latin speaker will see this city labeled as “Francofurtum” based on Wikidata instead of “Frankfort” based on OSM’s `name` tag:

name: Frankfort

([Planetiler equivalent](https://tile.ourmap.us/inspect.html#22/38.20090687/-84.87331472))

By contrast, [OSM Americana](https://zelonewolf.github.io/openstreetmap-americana/#map=11/38.2009/-84.8733&language=la) shows both the Wikidata fallback and the OSM `name` tag, which can give the mapper peace of mind:

[Francofurtum (Frankfort)](https://zelonewolf.github.io/openstreetmap-americana/#map=11/38.2009/-84.8733&language=la)

I realize this is not an appropriate place to provide design feedback on the proposed OpenMapTiles style, nor to toot Americana’s horn, but I think it’s important that we don’t view localized label sources as a black-and-white, zero-sum issue.

lonvia:
> so an English-speaking user will see the OSM area’s tag value, not the Wikidata item’s label.

This fallback doesn't work well for places that only have a `name` tag. In Germany, OMT seems to be using mainly names from wikidata when I look at the map with a German locale. The reason of course being that we don't tag `name:de` when there is only a `name` tag present and no translations. The issue is rather obvious with railway stations, e.g. https://osm.openmaptiles.org/#map=14/52.5229/13.4028&layers=V, even when using an English locale.

1ec5:
> > so an English-speaking user will see the OSM area’s tag value, not the Wikidata item’s label.
> 
> This fallback doesn't work well for places that only have a `name` tag. In Germany, OMT seems to be using mainly names from wikidata when I look at the map with a German locale. The reason of course being that we don't tag `name:de` when there is only a `name` tag present and no translations.

[This wiki page](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2559647#Repeating_name_with_language_specific_tag) discusses the issue in more detail. If the feature’s name is translated or transliterated into another language, then it is valid and expected to pair `name` with `name:de`. [This can come as a shock](https://lists.openstreetmap.org/pipermail/tagging/2022-March/thread.html#64124) to mappers who focus on monolingual regions and normally think of their language as the “obvious” one. To be sure, not everything named should have a `name:*`. An individual shop’s name typically does not have a reasonable translation, so `name:de` would serve very little purpose.

There is a gray area when it comes to features that do have translations but none of them is [locally attested or locally relevant](https://wiki.openstreetmap.org/wiki/Names#Language-specific_names). Data consumers face a catch-22 between, on the one hand, the principle that only locally attested names should be tagged in `name:*` and, on the other hand, the opinion that they should prefer `name` over any translations in Wikidata if there’s no `name:*`. Perhaps it isn’t a bad thing to expose mappers to this reality (optionally, as the Standard layer is still the default).

Again, an OpenMapTiles-based layer can address your feedback in at least two ways:

* Include both `name:de` and `name` in each label when the two differ (at the cost of some extra clutter and a stylistic departure from osm-carto)
* Disable the Wikidata fallback (which would require changes on MapTiler’s backend to avoid impacting their other customers)

> The issue is rather obvious with railway stations, e.g. https://osm.openmaptiles.org/#map=14/52.5229/13.4028&layers=V, even when using an English locale.

I assume you’re referring to the redundant “Station” suffix everywhere. This is the legacy of Wikidata bootstrapping its labels from Wikipedia (where the suffix is necessary for technical reasons that don’t apply to Wikidata). The issue can be comprehensively fixed in a matter of hours, as [I once did](https://community.openstreetmap.org/t/is-there-support-for-removing-all-name-tags-on-objects-with-a-wikidata-qid-present/8374/8) for place names in one language in many countries. Nevertheless, it highlights the fact that Wikidata is still a work in progress, just like OSM. If this project wants to hold a featured layer’s external data sources to a higher standard than OSM, then that is indeed this project’s prerogative.

matkoniecz:
> [This wiki page](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2559647#Repeating_name_with_language_specific_tag) discusses the issue in more detail. If the feature’s name is translated or transliterated into another language, then it is valid and expected to pair name with name:de.

Note 

> This fallback doesn't work well for places that only have a `name`  tag.

In such cases it is far less obvious is repeating `name` into `name:de` / `name:pl` is something that makes sense.

AndrewHain:
> I assume you’re referring to the redundant “Station” suffix everywhere. This is the legacy of Wikidata bootstrapping its labels from Wikipedia (where the suffix is necessary for technical reasons that don’t apply to Wikidata). The issue can be comprehensively fixed in a matter of hours, as [I once did](https://community.openstreetmap.org/t/is-there-support-for-removing-all-name-tags-on-objects-with-a-wikidata-qid-present/8374/8) for place names in one language in many countries. Nevertheless, it highlights the fact that Wikidata is still a work in progress, just like OSM. If this project wants to hold a featured layer’s external data sources to a higher standard than OSM, then that is indeed this project’s prerogative.

Names like “Leeuwarden station” are not intrinsically _wrong_, just unwanted and a hazard of using a project with different ends. A better solution than resorting to Wikidata could be committing to fix weaknesses in our own language names.

1ec5:
I’m not sure if this Wikidata discussion is really on-topic for this PR. MapTiler could take this issue off the table by disabling the Wikidata fallback on their end, but they’d have to determine whether they want to maintain a separate tileset for osm.org versus their other customers.

I also don’t think this will be the last style to be proposed that incorporates Wikidata in some fashion. In my opinion, there’s no strong need to prevent Wikidata labels from seeping into the website at all cost, as long as they only appear in a non-default style, and as long as they have some kind of special fallback styling or are combined with the OSM `name` as in https://github.com/openstreetmap/openstreetmap-website/pull/4042#issuecomment-1636888913.

> Names like “Leeuwarden station” are not intrinsically _wrong_, just unwanted and a hazard of using a project with different ends.

The English Wikipedia used to often omit “station” from railway station article titles, but they adopted this convention after [much discussion](https://en.wikipedia.org/wiki/Wikipedia_talk:Naming_conventions_%28US_stations%29#Previous_discussions) about considerations that have little bearing on Wikidata label conventions. I’d contend that the Wikidata community would _prefer_ to move “station” from each label to the corresponding description.

Regardless, you have a point that Wikidata labels are not necessarily designed to be map labels. Technically they’re more like the `name`s of convenience that mappers often set on route relations in OSM. A more clear-cut example is that “Washington, D.C.”, is a fine Wikidata label, but the map label really should be just “Washington”. I’ve opened openmaptiles/openmaptiles-tools#437 and onthegomap/planetiler#679 to have these tile generators prefer a specifically tagged “map label” statement over the more generic label.

> A better solution than resorting to Wikidata could be committing to fix weaknesses in our own language names.

Such as welcoming translations and transliterations that aren’t verifiable on the ground? Even more out of scope for this PR. 😉

matkoniecz:
> in my opinion, there’s no strong need to prevent Wikidata labels from seeping into the website at all cost

In my opinion any style that uses Wikidata names even if OSM name tags are provided should not be included (note, if `name:LANG` tag is not present then using that language label from Wikidata is not covered by that).

I think that we should not never need to tell anyone that fixing OSM data will not help and they need to fix something on third-party website to fix name display. Or encourage people to remove valid `wikipedia`/`wikidata` tags as some fields in Wikidata have wrong info.

(yes, I am aware about Natural Earth data resulting in some problems on low zoom levels of OSM Carto, but we should not make things significantly worse)
(also, note that Wikidata imported data from Wikipedia which in turn encouraged sourcing geographic data from Google Maps which is separate part of a quagmire)

1ec5:
> In my opinion any style that uses Wikidata names even if OSM name tags are provided should not be included (note, if name:LANG tag is not present then using that language label from Wikidata is not covered by that).

That’s the case we’re talking about here, if I’m not mistaken.

>  (also, note that Wikidata imported data from Wikipedia which in turn encouraged sourcing geographic data from Google Maps which is separate part of a quagmire)

Let’s not turn this PR into a therapy session about all of Wikidata’s faults. Google Maps uses Wikidata for labels, not the other way around.

zdila:
In the recent release of OpenMapTiles 3.15, the Wikidata translations have been switched off: https://maptiler.com/news/2024/04/openstreetmap-data-prepared-for-advanced-cartography/
 
Are there any other blockers for accepting this PR?

tordans:
Can we test the current state somewhere? 
https://osm.openmaptiles.org/ is down.
Maybe Tom can deploy this branch on one of the testing servers?

zdila:
@tordans https://osm.openmaptiles.org/ - I've just started it again.

jmaspons:
Can the language selector be added as a combobox at the map layers panel? Would be more discoverable and selectable for users without osm account.

tomhughes:
> Can the language selector be added as a combobox at the map layers panel? Would be more discoverable and selectable for users without osm account.

There are already separate issues for adding a language selector for non-logged in users.

LaoshuBaby:
> the Wikidata translations have been switched off

Is it possible to disable automatic transliteration of non-Latin names?

matkoniecz:
I think it is fine? In the same way as it would be fine to host nondefault layer showing names in Korean/Polish/Japanese?

jachym:
Any resolution regarding this PR? Can we have this PR merged and closed, I believed, we addressed the concerns and issues of the community.

ZeLonewolf:
New layers are approved by the Operations Working Group, which has not happened yet.

It was last discussed at their [May meeting](https://osmfoundation.org/wiki/Operations/Minutes/2024-05-16#OpenMapTiles_application_to_become_a_featured_layer). I do not see that it has been on an OWG meeting [agenda](https://osmfoundation.org/wiki/Working_Group_Minutes) since.

AntonKhorev:
Commits such as "fix lint issues" https://github.com/openstreetmap/openstreetmap-website/pull/4042/commits/888767144630f56340097450c362edc639a771d8 shouldn't exist, see [CONTRIBUTING.md#pull-requests](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests)

pnorman:
This is awaiting a response from MapTiler

AntonKhorev:
Still has "fix lint issues".

app/assets/javascripts/leaflet.map.js
zdila: List of supported languages has been moved to `@maptiler/maplibre-gl-omt-language` package. This can be also discussed in follow-up tickets if necessary.

zdila: List of supported languages has been moved to `@maptiler/maplibre-gl-omt-language` package. This can be also discussed in follow-up tickets if necessary.

zdila:
You can test on https://osm.openmaptiles.org/ with credentials - login:  `martin.zdila@maptiler.com`, password: `osm4ever`

app/assets/javascripts/leaflet.map.js
```diff
@@ -9,6 +14,43 @@ L.extend(L.LatLngBounds.prototype, {
   }
 });
 
+if (OSM.MAPTILER_KEY) {
+  maplibregl.setRTLTextPlugin(
+    "https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js",
```
1ec5, line 19: To avoid this external dependency on a Mapbox API, install the `@mapbox/mapbox-gl-rtl-text` package and have a build script copy node_modules/@mapbox/mapbox-gl-rtl-text/dist/mapbox-gl-rtl-text.js into app/assets/javascripts/. Then this becomes just `mapbox-gl-rtl-text.js`.

app/assets/javascripts/leaflet.map.js
```diff
@@ -9,6 +14,43 @@ L.extend(L.LatLngBounds.prototype, {
...
+      var m = this.getMaplibreMap();
+      var supportedLanguages = maplibregl.Map.prototype.supportedLanguages;
+      var language;
+      if (supportedLanguages.includes(I18n.locale)) {
+        language = I18n.locale;
+      } else {
+        var mainLocale = I18n.locale.slice(0, 2);
+        var localeIndex = supportedLanguages.findIndex(function (locale) {
+          return locale === mainLocale;
+        });
```
1ec5, line 39: This website and maplibre-gl-omt-language both support some locales that have three-character-long language codes, such as [Western Punjabi](https://github.com/openstreetmap/openstreetmap-website/blob/e206dd527ef7c6d1829f131533d722e536e4bfb5/config/locales/pnb.yml) (pnb). I don’t think this code will match such locales. If I’m not mistaken, the intention is to match on only the language code and ignore any script or country code. If so, this would be more robust:

```suggestion
        var mainLocale = I18n.locale.split('-')[0];
        var localeIndex = supportedLanguages.findIndex(function (locale) {
          return locale === mainLocale;
        });
```

Note that this is still quite a rudimentary locale matching heuristic. It won’t match `zh-Hans` or `zh-Hant`, which the plugin also supports.

app/assets/javascripts/leaflet.map.js
```diff
@@ -9,10 +14,58 @@ L.extend(L.LatLngBounds.prototype, {
...
+
+        el.href = el.dataset.hrefTemplate
+          .replace("{lang}", I18n.locale.replace(/-.*/, ""))
+          .replace("{pos}", zoom + "/" + c.lat.toFixed(precision) + "/" + c.lng.toFixed(precision) + "/30.00/60.00");
+      }
+    }
+
+    map._update3dMapUrl = update3dMapUrl;
+    setTimeout(update3dMapUrl);
+    map.on("moveend", update3dMapUrl);
```
AntonKhorev: You don't need any of this added in this commit if you remove it in the next one.

app/assets/javascripts/leaflet.map.js
zdila: Which build script do you mean? I added this to `package.json` but `docker compose build` fails because such path doesn't exist when building docker image.

```
  "scripts": {
    "postinstall": "cp node_modules/@mapbox/mapbox-gl-rtl-text/mapbox-gl-rtl-text.min.js app/assets/javascripts/mapbox-gl-rtl-text.min.js"
  },
```

app/assets/javascripts/leaflet.map.js
zdila: I have reworked language matching.

app/assets/javascripts/leaflet.map.js
```diff
@@ -9,6 +13,35 @@ L.extend(L.LatLngBounds.prototype, {
   }
 });
 
+if (OSM.MAPTILER_KEY) {
+  maplibregl.setRTLTextPlugin("https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js", true);
```
HolgerJeromin, line 17: I am pretty sure this need to be changed to be served from osm.org or probably bundled like the rest...

app/assets/javascripts/leaflet.map.js
jachym: @1ec5 and @HolgerJeromin any help/comments on that, so that we could more forward?

config/settings.yml
```diff
@@ -107,6 +107,8 @@ attachments_dir: ":rails_root/public/attachments"
 #logstash_path: ""
 # List of memcache servers to use for caching
 #memcache_servers: []
+# URL of Maptiler API for vector maps
+maptiler_url: "https://api.maptiler.com/"
```
tomhughes, line 111: Does this actually need to be a setting? The other layers are all hard coded and in fact this is hard coded in the javascript, just not when setting the security policy.

What we should have in here is a commented out setting for `maptiler_key` though, as we already do for the other map layers that need a key.

app/assets/javascripts/leaflet.map.js
tomhughes: To answer the question from that comment, you modify `app/assets/config/manifest.js` and add:

```js
//= link @mapbox/mapbox-gl-rtl-text/mapbox-gl-rtl-text.js
```

then add to `app/assets/javascripts/osm.js.erb` something like:

```js
RTL_TEXT_PLUGIN: <%= javascript_path("@mapbox/mapbox-gl-rtl-text/mapbox-gl-rtl-text.js").to_json %>,
```

and then you can use `OSM.RTL_TEXT_PLUGIN` here in place of the mapbox URL string.

app/assets/javascripts/leaflet.map.js
```diff
@@ -44,6 +77,32 @@ L.OSM.Map = L.Map.extend({
...
+        text: I18n.t("javascripts.map.website_and_api_terms")
+      }).prop("outerHTML");
+
+      this.baseLayers.push(new L.OpenMapTiles({
+        attribution: copyright + ". " + openmaptiles_link + ". " + terms,
+        code: "V",
+        keyid: "openmaptiles_osm",
+        name: I18n.t("javascripts.map.base.openmaptiles_osm")
+      }));
+    }
```
tomhughes, line 104: This should, if at all possible, be done in `config/layers.yml` rather than hand coding it - if we need to add new keys to that file and code here to handle them that's fine but all layers are supposed to be defined there now.

config/locales/en.yml
```diff
@@ -3316,7 +3318,7 @@ en:
       title: "Creating New Redaction"
     show:
       description: "Description:"
-      heading: "Showing Redaction \"%{title}\""
+      heading: 'Showing Redaction "%{title}"'
```
tomhughes, line 3321: Why has this changed?

tomhughes (changes requested):

tomhughes:
Also, the eslint warnings should be fixed.

app/assets/javascripts/leaflet.map.js
1ec5: See https://github.com/openstreetmap/openstreetmap-website/pull/4042#discussion_r1966816765, sorry for the delay.

app/assets/javascripts/leaflet.map.js
tomhughes: So I was able to use this definition:

```yaml
- leafletOsmId: "OpenMapTiles"
  code: "V"
  layerId: "openmaptiles_osm"
  nameId: "openmaptiles_osm"
  apiKeyId: "MAPTILER_KEY"
  credit:
    id: "openmaptiles"
    children:
      openmaptiles_link:
        id: "openmaptiles_url"
        href: "https://openmaptiles.org/"
      maptiler_link:
        id: "maptiler_url"
        href: "https://www.maptiler.com/"
```

with a few changes - move `L.OpenMapTiles` to `L.OSM.OpenMapTiles` and change it to set the style URL in a constructor using the `apikey` option, for example:

```js
    initialize: function (options) {
      L.Util.setOptions(this, {
        ...options,
        style: "https://api.maptiler.com/maps/openstreetmap/style.json?key=" + options.apikey
      });

      L.MaplibreGL.prototype.initialize.call(this);
    },
```

app/assets/javascripts/leaflet.map.js
```diff
@@ -44,6 +77,32 @@ L.OSM.Map = L.Map.extend({
...
+      });
+
+      var terms = $("<a>", {
+        href: "https://wiki.osmfoundation.org/wiki/Terms_of_Use",
+        text: I18n.t("javascripts.map.website_and_api_terms")
+      }).prop("outerHTML");
+
+      this.baseLayers.push(new L.OpenMapTiles({
+        attribution: copyright + ". " + openmaptiles_link + ". " + terms,
+        code: "V",
```
AntonKhorev, line 100: How do we allocate the single-letter codes? "V" is nowhere in the name of this layer.

"O" was formerly used for OPNVKarte, not sure if we can reuse it.

1ec5: I think it’s just based on the name of the layer, using whatever letter is available and unlikely to clash with the others. OpenHistoricalMap’s fork has also assigned [its own codes](https://github.com/OpenHistoricalMap/ohm-website/blob/ad4ea829a7f794a456ef5c1484e977653cd9f7c3/app/assets/javascripts/leaflet.map.js#L81), currently J, O, R, and W, but there hasn’t been a conflict because some of OSM’s featured layers have been removed. The only risk from reusing a previously assigned letter is that some old links to the old layer will now point to the new layer, but these links have been falling back to the Standard layer anyways.

zdila:
Just FYI I've left MapTiler and so I will no longer update this PR. Feel free to update it to your needs yourselves. Good luck and best regards.

jachym:
> Just FYI I've left MapTiler and so I will no longer update this PR. Feel free to update it to your needs yourselves. Good luck and best regards.

Yes, I confirm that. Many thanks @zdila to push this PR. 

We may take over it.

app/assets/javascripts/osm.js.erb
```diff
@@ -43,6 +43,10 @@ OSM = {
   OPEN_NOTE_MARKER: <%= image_path("open_note_marker.svg").to_json %>,
   CLOSED_NOTE_MARKER: <%= image_path("closed_note_marker.svg").to_json %>,
 
+<% if Settings.key?(:maptiler_key) %>
+  MAPTILER_KEY: <%= Settings.maptiler_key.to_json %>,
+<% end %>
```
hlfan, line 48: This should be with the other keys in the embedded Settings json

pnorman:
> Yes, I confirm that. Many thanks @zdila to push this PR.
> 
> We may take over it.

Are you planning on taking over this PR? I have some work that will build on it, but I want to figure out what order to do the work in.

petr-hajek:
> > Yes, I confirm that. Many thanks @zdila to push this PR.
> > We may take over it.
> 
> Are you planning on taking over this PR? I have some work that will build on it, but I want to figure out what order to do the work in.

Hey, I'd like to finish this, but I'm not sure when I will have time for it. I hope before the end of June.

JGreenlee:
Are the maintainers planning to make the final changes here, or should someone from the community do it?

petr-hajek:
> Are the maintainers planning to make the final changes here, or should someone from the community do it?

I am planning to do it in the next 2 weeks

tomhughes:
I think @pnorman has also been working on it.

pnorman:
My PR won't touch OpenMapTiles stuff.

petr-hajek:
@tomhughes I tried to address all the raised issues. Please bear with me as this is my first PR to this repo, I'm ready for any feedback. Could you authorize me to run the workflows to see if all the lints and test will pass?

In the scope of my work I also updated the newly added deps to their latest version.

@pnorman I think it should be compatible with your work in https://github.com/openstreetmap/openstreetmap-website/pull/6137 in case this gets merged first

app/assets/javascripts/leaflet.map.js
```diff
@@ -1,10 +1,43 @@
...
+      for (let i = 0; i < OSM.preferred_languages.length; i++) {
+        const preferredLanguage = OSM.preferred_languages[i].toLowerCase();
+        const matchedLanguage = supportedLanguages.find(function (supportedLanguage) {
+          return supportedLanguage.toLowerCase().replace("_", "-") === preferredLanguage;
+        });
+        if (matchedLanguage) {
+          maplibreMap.setLanguage(matchedLanguage);
+          break;
+        }
+      }
```
hlfan:
```suggestion
      for (const preferredLanguage of OSM.preferred_languages) {
        const preferred = preferredLanguage.toLowerCase().replace("-", "_");
        const match = supportedLanguages.find(supported =>
          supported.toLowerCase() === preferred
        );
        if (match) {
          maplibreMap.setLanguage(match);
          break;
        }
      }
```
Optionally put the `.find()` in one line or guard setLanguage with `continue`.

package.json
```diff
@@ -18,4 +22,4 @@
     "eslint-formatter-compact": "^8.40.0",
     "eslint-plugin-erb": "^2.1.0"
   }
-}
+}
```
hlfan: Why was this changed?

app/assets/javascripts/leaflet.map.js
```diff
@@ -1,10 +1,43 @@
...
   }
 });
 
+if (OSM.MAPTILER_KEY) {
+  maplibregl.setRTLTextPlugin(OSM.RTL_TEXT_PLUGIN, true);
+
+  L.OSM.OpenMapTiles = L.MaplibreGL.extend({
+    options: {
+      maxZoom: 23,
+      style: "https://api.maptiler.com/maps/openstreetmap/style.json?key=" + OSM.MAPTILER_KEY
```
hlfan:
```suggestion
      style: OSM.MAPTILER_OSM_STYLE_URL + "?key=" + OSM.MAPTILER_KEY
```

I know this kinda goes against https://github.com/openstreetmap/openstreetmap-website/pull/4042#discussion_r1966812213, but I'd prefer having the whole URL in the settings and basing the CSP on it like in #6137:
https://github.com/openstreetmap/openstreetmap-website/blob/7f98378630eaba699e152602046fa3a7eab924e1/app/controllers/application_controller.rb#L248-L249

app/assets/javascripts/leaflet.map.js
hlfan, line 1: This could instead be placed into a `leaflet.osm.vector.js` file and inserted at line 10 of `application.js`

app/assets/javascripts/leaflet.map.js
petr-hajek: @tomhughes WDYT?

vendor/assets/leaflet/leaflet.osm.vector.js
hlfan, line 1: This shouldn't be under vendor but next to the leaflet.map.js you pulled it out of.

config/layers.yml
```diff
@@ -79,3 +79,18 @@
       osm_france_link:
         id: "osm_france"
         href: "https://openstreetmap.fr/"
+
+- leafletOsmId: "OpenMapTiles"
+  code: "V"
+  layerId: "openmaptiles_osm"
+  nameId: "openmaptiles_osm"
+  apiKeyId: "maptiler_key"
```
hlfan, line 87: If you need the API key only in the style initialization the way you're currently doing it, then you don't need it here.
If you can set it by templating "{apikey}" then you don't need to forward it into osm.js

pnorman:
I *think* there was a question that still needed answering from MapTiler about the featured layer discussions. I'll have to review emails or check notes to be sure

tomhughes:
I think the only thing is that somebody suggested (and I can't remember where I saw it) that the layer included additional buildings that weren't from OSM data but I haven't had a chance to investigate that claim yet.

config/layers.yml
petr-hajek: using the apikey in the initialization and removed it from OSM

petr-hajek:
In the end I keep maplibre v4 with maplibre-gl-leaflet 0.0.22 since for some reason the custom attribution got broken after updating. Someone else can update this in some followup PR.

hlfan (approved):
Whether or not the style.json path definition changes, this now looks good to me. That could also be worth revisiting when there's more than one vector style.
Tests pass in my fork https://github.com/hlfan/openstreetmap-website/pull/18.
If you wanna do other reviewers a favor, go through the code comments and mark applicable ones as _resolved_.

petr-hajek:
> Whether or not the style.json path definition changes, this now looks good to me. That could also be worth revisiting when there's more than one vector style. Tests pass in my fork [hlfan#18](https://github.com/hlfan/openstreetmap-website/pull/18). If you wanna do other reviewers a favor, go through the code comments and mark applicable ones as _resolved_.

I'm sorry, unfortunately I cannot resolve them, I suppose since this is not originally my PR, so if you can resolve those raised by you, it'd be great, thanks.

jachym:
> I think the only thing is that somebody suggested (and I can't remember where I saw it) that the layer included additional buildings that weren't from OSM data but I haven't had a chance to investigate that claim yet.

This style is using OSM data only - adjusted according with the request of the community.

tomhughes:
Is there an api key I can use for testing? You can let me have it privately if you prefer...

petr-hajek:
> Is there an api key I can use for testing? You can let me have it privately if you prefer...

I believe it should work with the Free account so just by signing up here you can get one: https://cloud.maptiler.com/account/keys/

tomhughes:
Thanks @petr-hajek I had in fact already registered a key there I had just forgotten about it ;-)

I think this is good to merge now - what do you want to do about a key for us to use in production? Am I OK just to register a key for that or do you need to do something special?

hlfan:
@tomhughes given your comments at https://github.com/openstreetmap/openstreetmap-website/pull/6137#discussion_r2203290250, I think it would be better to split up the `leaflet.osm.vector.js` into a `leaflet.maplibregl.js` and a `leaflet.openmaptiles.js` file.

tomhughes:
Yes I deliberately didn't ask for that here as there's no point having two people do the same work - it's just a conflict I can resolve when merging. Indeed I've just done a test this afternoon of merging them both and resolving the conflicts.

petr-hajek:
> Thanks @petr-hajek I had in fact already registered a key there I had just forgotten about it ;-)
> 
> I think this is good to merge now - what do you want to do about a key for us to use in production? Am I OK just to register a key for that or do you need to do something special?

I'll double check with my teammates here in MapTiler and reach out to you in PM.

petr-hajek:
@tomhughes PM sent from https://www.openstreetmap.org/user/bananaboy on OSM - in case you want to use a different means of communication, please let me know there

tomhughes:
Thanks for all the work on this, which I think is good to merge now.

Firefishy:
Some initial PR: https://en.osm.town/@osm_tech/114902624740876388

Issue #4043, opened 2023-05-24
Coveralls builds show incorrect coverage


gravitystorm:
Something strange is going on with the Coverage CI tests. The individual jobs are failing, even though the overall coverage is fine:

![Screenshot from 2023-05-24 15-32-40](https://github.com/openstreetmap/openstreetmap-website/assets/360803/1d705911-e207-4d09-a33a-7bef13e99495)

If we dig into a build (e.g. https://coveralls.io/jobs/121929517), we can see the individual jobs have duplicate files, each with different counts for the coverage:

![Screenshot from 2023-05-24 15-51-32](https://github.com/openstreetmap/openstreetmap-website/assets/360803/23123a1b-6327-4c3a-817a-590204653cc9)

The most recent successful build was for https://github.com/openstreetmap/openstreetmap-website/commit/4a786a5e6e7f187a0f4740df53e3032caed688d3 (April 4th) https://coveralls.io/builds/58564512 and it doesn't show the duplicate files:

![Screenshot from 2023-05-24 15-48-17](https://github.com/openstreetmap/openstreetmap-website/assets/360803/c416919f-bac7-4db6-a0cd-271d4e4153e8)

gravitystorm:
It's hard to pinpoint exactly where coveralls started failing, since the first failure was unrelated (a translations issue) and there were a bunch of commits before that issue was fixed and the coveralls issue was revealed (coveralls doesn't run if the build itself fails for other reasons). 

But the most suspicious commit is ["Bump coverallsapp/github-action from 1.2.4 to 2.1.0"](https://github.com/openstreetmap/openstreetmap-website/commit/ce00b921cb6a65ad58767ef1dcf8efc6489bb5f2)

tomhughes:
I think that https://github.com/openstreetmap/openstreetmap-website/blob/c96b154b5f56ffbf40eb8a62ee528bef8adec69b/test/test_helper.rb#L19 is probably the root cause - we are generating both native simplecov HTML output and lcov formatted output.

As coveralls v2 has auto detection it's probably finding and reading both - adding an explicit format to choose one or the other will hopefully help.

PR #4044, opened 2023-05-24
Temporary revert of coveralls github action

gravitystorm:
Refs #4043

While we investigate what might be causing the problem, and how to fix it, let's revert the coveralls action to a known-good version in the meantime.


gravitystorm:
Let's see if this actually works first :-)

tomhughes:
Looks like https://github.com/openstreetmap/openstreetmap-website/commit/218b140912dd99d87c61d609da42b8507f89ccb8 has fixed the underlying problem so we don't need to do this.

PR #4045, opened 2023-05-27
 Rename id.rb to id.rbt

5212BABA:
Cmd @cflipse


tomhughes:
Why do you want to rename this? What are you trying to achieve?

tomhughes:
As you can see this causes tests to fail because `.rbt` is not a known ruby extension so this appears to be nonsensical and as there is no rationale for it I am closing.

Issue #4048, opened 2023-05-31
Finish upgrading to 22.04


gravitystorm:
In two recent commits (302a53d767a468a7bd4d716a1a0eba8ba38ad09b, dc9d52a7c844774d28a872f9e456eeed3c19bfee)  @tomhughes started the process of upgrading to Ubuntu 22.04 ("jammy"). This issue is to track the remaining tasks.

We have four places that are relevant to installation instructions, and they should all be kept in sync with one another.
* The written instructions in [INSTALL.md](https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md)
* The code in [Dockerfile](https://github.com/openstreetmap/openstreetmap-website/blob/master/Dockerfile)
* The code in [provision.sh](https://github.com/openstreetmap/openstreetmap-website/blob/master/script/vagrant/setup/provision.sh) (for vagrant)
* The code in [tests.yml](https://github.com/openstreetmap/openstreetmap-website/blob/master/.github/workflows/tests.yml) (github workflows)

The github workflows part is a bit of a curveball, since it uses "Github Flavoured Ubuntu" (see https://github.com/openstreetmap/openstreetmap-website/issues/3946#issuecomment-1445309595), so it doesn't use Ubuntu packages for node, geckodriver and other important dependencies, but rather does its own thing. But we should try, as much as possible, to keep these all in sync.

tomhughes:
I already fixed Dockerfile in dc9d52a7c844774d28a872f9e456eeed3c19bfee?

Incidentally the reason for the upgrade was that one of the gems we use dropped ruby 2.x support and 20.04 doesn't have ruby 3 in the main repos.

tomhughes:
We could fix INSTALL.md in line with Dockerfile but some of the steps (to get a working geckodriver mainly) are a bit nasty...

gravitystorm:
I think the above PRs have included everything, but if anyone spots anything that I've missed please let me know.

PR #4050, opened 2023-05-31
Update vagrant provisioning for 22.04

gravitystorm:
Refs #4048

* geckodriver is now inside the firefox snap package
* the yarnpkg package is completely broken, see https://bugs.launchpad.net/ubuntu/+source/node-yarnpkg/+bug/2003697


tomhughes:
Do we not need `snap install firefox` or is that already installed?

gravitystorm:
> Do we not need `snap install firefox` or is that already installed?

libvips-dev pulls in the firefox deb package (via a bunch of dependencies) and that triggers the snap install.

PR #4058, opened 2023-06-03
Clarify that symlinks refers to git not the operating system

simonpoole:
Just a small clarification.


AntonKhorev:
Symlinks refer to both git and OS. You have enable them in git and give yourself permissions to create them in Windows. That's why you have to do two things: git config and developer mode.

Issue #3182, opened 2021-04-23
Add pop-up or other obvious notification for getting a message or changeset comment


aawiseman:
I've seen that mappers don't often realize they have gotten a message or a changeset comment, which can be a problem in general, but especially if there are new mappers or people who are not following policies and may not realize it. Even if you tell new mappers that they might receive messages or changeset comments, it's not very clear or obvious that they have received one.

I think it would be good to have a pop-up window, blinking notification or some other obvious visual showing that they just received a message or changeset comment -- ideally as soon as it arrives if they are are looking at osm.org, and as soon as they log in if they aren't at osm.org. 

I think this could help with new mapper retention as well, since there have been studies like this that people who receive a comment or message are more likely to keep mapping https://core.ac.uk/download/pdf/159755161.pdf

tomhughes:
Well we send an email already, which will arrive even if they don't visit the site!

aawiseman:
Right, I just think an obvious visual on the website would help as well -- a pop-up they can't ignore, some kind of blinking or flashing thing, etc. They might not think to look for the email, it might go to spam, etc. 

I think in general the more ways people can see they are part of a community and are getting messages/comments, the better. Especially for new mapppers, an email saying "has commented on one of your changesets" is not super clear. 

Or maybe they only planned to map one time, so they might not even notice or care if they get an email, but if they get a notification while they're mapping, it might help to encourage them to keep mapping.

migurski:
Andrew, do you mean the same information as the unread-count next to My Messages in the account drop-down?

<img width="741" alt="Screen Shot 2021-04-23 at 4 07 03 PM" src="https://user-images.githubusercontent.com/58730/115937924-f41e5e00-a44d-11eb-9194-29fd5acfad51.png">

LinkedIn does a reasonable job of showing counts attached to top nav, we could imitate this.

<img width="531" alt="Screen Shot 2021-04-23 at 4 07 20 PM" src="https://user-images.githubusercontent.com/58730/115937941-fda7c600-a44d-11eb-80a5-9dd67b34100d.png">

aawiseman:
I think something more obvious than that -- for example I've seen new mappers not even notice that little number while looking over their shoulder and teaching them to map, and when I say "hey you got a message" it's not obvious where to look.

[...]

PR #4070, opened 2023-06-19
Change the way notifications about messages appear

Dimitar5555:
Ref: #3182

This PR removes the current light gray box for the message count and replaces it with a red dot. The element with messages number is also now red.

There is some jQuery code which might not work correctly, please test it (if it isn't covered by the automated test) and let me know if it works. The code is related to updating the number of unread messages on the navigation bar.

Screenshots:
Before:
Screenshot 2023-06-19 at 16-52-09 OpenStreetMap
After:
Screenshot 2023-06-19 at 16-52-00 OpenStreetMap

### DIFF
--- a/app/assets/javascripts/messages.js
+++ b/app/assets/javascripts/messages.js
@@ -1,7 +1,7 @@
 $(document).ready(function () {
   $(".inbox-mark-unread").on("ajax:success", function (event, data) {
     $("#inboxanchor").remove();
-    $("#user_profile_picture").after(data.inboxanchor);
+    $(".profile_photo").append(data.inboxanchor);
 
     $("#inbox-count").replaceWith(data.inbox_count);
 
@@ -10,7 +10,7 @@ $(document).ready(function () {
 
   $(".inbox-mark-read").on("ajax:success", function (event, data) {
     $("#inboxanchor").remove();
-    $("#user_profile_picture").after(data.inboxanchor);
+    $(".profile_photo").append(data.inboxanchor);
 
     $("#inbox-count").replaceWith(data.inbox_count);
 
@@ -19,7 +19,7 @@ $(document).ready(function () {
 
   $(".inbox-destroy").on("ajax:success", function (event, data) {
     $("#inboxanchor").remove();
-    $("#user_profile_picture").after(data.inboxanchor);
+    $(".profile_photo").append(data.inboxanchor);
 
     $("#inbox-count").replaceWith(data.inbox_count);
 
--- a/app/assets/javascripts/messages.js
+++ b/app/assets/javascripts/messages.js
@@ -1,7 +1,7 @@
 $(document).ready(function () {
   $(".inbox-mark-unread").on("ajax:success", function (event, data) {
     $("#inboxanchor").remove();
-    $(".user-button").before(data.inboxanchor);
+    $("#user_profile_picture").after(data.inboxanchor);
 
     $("#inbox-count").replaceWith(data.inbox_count);
 
@@ -10,7 +10,7 @@ $(document).ready(function () {
 
   $(".inbox-mark-read").on("ajax:success", function (event, data) {
     $("#inboxanchor").remove();
-    $(".user-button").before(data.inboxanchor);
+    $("#user_profile_picture").after(data.inboxanchor);
 
     $("#inbox-count").replaceWith(data.inbox_count);
 
@@ -19,7 +19,7 @@ $(document).ready(function () {
 
   $(".inbox-destroy").on("ajax:success", function (event, data) {
     $("#inboxanchor").remove();
-    $(".user-button").before(data.inboxanchor);
+    $("#user_profile_picture").after(data.inboxanchor);
 
     $("#inbox-count").replaceWith(data.inbox_count);
 
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -204,15 +204,6 @@ nav.secondary {
     }
   }
 
-  #inboxanchor {
-    display: inline-block;
-    height: 25px;
-    margin: 3px 0 3px 3px;
-    background-color: lighten($grey, 10%);
-    line-height: 20px;
-    border-radius: 3;
-  }
-
   .dropdown-menu {
     .count-number {
       font-size: 14px;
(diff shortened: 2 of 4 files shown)


tomhughes:
Why get rid of the message count? If you want to make it more prominent why not just change the colour as for the one in the menu?

gravitystorm:
@Dimitar5555 My concern with this PR is that I think it breaks our UI conventions. You've used the "danger" colour to highlight something that I wouldn't normally put in that category. We use "danger" for things like actions that cannot be reversed, or errors that can't be recovered from. We use warnings for things of lesser importance, like when the whole service is read-only or when there's (recoverable) errors in form submissions. 

But this makes me think that a "unread message count" status indicator wouldn't qualify even for a warning level, never mind our danger level.

I'm happy to hear your thoughts, but otherwise I'm minded to decline this proposal.

(As an aside, I'd be happy to see colour consistency between the number shown on the collapsed and expanded menus, and if this used a "information" level bootstrap class then so much the better!)

Dimitar5555:
I see from where you are coming from but I'm not sure if agree with you for this specific case.

It is generally agreed that red is used for notifications (see the [old facebook design](https://img.freepik.com/free-vector/facebook-notification-icons_23-2147675572.jpg), [iOS](https://i.stack.imgur.com/hx1fz.png) and [Android](https://i.stack.imgur.com/i3SS9.jpg)). `bg-info` could work but it seems like it may be less visible. See also the two comments from the referenced issue:

> LinkedIn does a reasonable job of showing counts attached to top nav, we could imitate this.

> Screen Shot 2021-04-23 at 4 07 20 PM

_Originally posted by @migurski in https://github.com/openstreetmap/openstreetmap-website/issues/3182#issuecomment-825986025_

> At bare minimum I think at least making the little "unread" notification on the upper right corner more obvious would be a big improvement, similar to migurski 's comment above with the Linkedin example. Making it red or blink or flash or something that signals "hey you, look here!" 

_Originally posted by @aawiseman in https://github.com/openstreetmap/openstreetmap-website/issues/3182#issuecomment-825990205_
            

Just for reference, here is how it would look using `bg-info`. In my opnion it looks a bit odd. It's also not "eye catching" like the red colour.
image

And near water features:
image


> As an aside, I'd be happy to see colour consistency between the number shown on the collapsed and expanded menus

It's the same colour. I assume that the different vertical sizes play some game with the perception of the red colour. You can check it with a colour picker.

Dimitar5555:
> > > As an aside, I'd be happy to see colour consistency between the number shown on the collapsed and expanded menus
> > 
> > It's the same colour. I assume that the different vertical sizes play some game with the perception of the red colour. You can check it with a colour picker.
> 
> Sorry, I mean the current situation should be changed - we have the indicator in grey, and then green in the dropdown, which is inconsistent.

That's also addressed.

AntonKhorev:
I think "danger" color should be displayed for blocked users.

pnorman:
> It is generally agreed that red is used for notifications (see the [old facebook design](https://img.freepik.com/free-vector/facebook-notification-icons_23-2147675572.jpg), [iOS](https://i.stack.imgur.com/hx1fz.png) and [Android](https://i.stack.imgur.com/i3SS9.jpg)). `bg-info` could work but it seems like it may be less visible.

Github, Twitter (I think), and Mastedon use blue.

image
image

Facebook uses red in places, blue in others.
image

Dimitar5555:
@gravitystorm suggested using `bg-info` which is more cyan than blue. `bg-primary` would work better since it's actually blue and should stand out more. I will add a screenshot with that variant tommorow.

For reference, these are the standard colours supported by Bootstrap 5:

cmoffroad:
Great idea, @Dimitar5555! 

When it comes to UX, using red for notification badges is a standard approach in popular design systems such as Android, iOS, and [Material.io](https://m3.material.io/components/badges/overview)

@gravitystorm It's important to note that red doesn't necessarily mean an error. Instead, it signifies the need for user attention and action, which is a perfect fit for notifications. While red has historically been associated with errors in forms and dialogs (alerts) and classified as a danger/error color in style sheets, many design systems continue to use it for notifications.

On the other hand, blue (specifically, the "bg-info" shade) typically denotes information that doesn't require action. Using blue for notifications would contradict the purpose of making this feature more prominent.

That being said, I would love to see a more modern design with rounded avatars and aligned badges, something like:

Screen Shot 2566-07-10 at 22 52 56

gravitystorm:
> When it comes to UX, using red for notification badges is a standard approach in popular design systems such as Android, iOS, and [Material.io](https://m3.material.io/components/badges/overview)

Sure, I understand that other design systems use red for notifications, but that doesn't mean that we have to. There's a balance to be struck between making the badge useful, without going down the "attention economy" UX (anti-)patterns found elsewhere. And as @pnorman points out, there's other examples of using a distinct-but-not-red notification colour too.

But in particular, the code in this PR is using `bg-danger`. It's a deliberate approach to use meaningful (i.e. not colour) names in our styles. This is particularly important when considering themes e.g. dark themes. So `danger` is an incorrect way to describe notification badges, even if they both happen to be red.

> @gravitystorm suggested using `bg-info` which is more cyan than blue.

If the bg-info colour doesn't really work for our colour scheme, then I'm happy to have suggestions for changing it. We haven't fully resolved our custom colours and the bootstrap defaults, so I know some combinations jar with each other. That might be better discussed separately though.

> `bg-primary` would work better since it's actually blue and should stand out more. I will add a screenshot with that variant tommorow.

Happy to see more suggestions. I still think that this is a "info" kind of thing rather than "primary", and if we use "info" and it looks wrong we should consider why that might be.

matkoniecz:
> Happy to see more suggestions. I still think that this is a "info" kind of thing rather than "primary", and if we use "info" and it looks wrong we should consider why that might be.

Is reassigning "info" colour feasible? Is it used elsewhere already?

Dimitar5555:
> gravitystorm suggested using `bg-info` which is more cyan than blue. `bg-primary` would work better since it's actually blue and should stand out more. I will add a screenshot with that variant tommorow.

I remember posting a screenshot but I guess I only thought to do it and never did, my bad.

I'm not sure if `bg-primary` is a good choice either but it's definitely better than `bg-info`. 

`bg-primary`:
image


Personally I feel like `#0073ff` would work better since it's a bit more vibrant.

image

Note to self: I've removed the `count-number` class from the elements which would break some JS code. I will have to restore it when we decide which colour is suitable.

gravitystorm:
Since this PR was made, the notification numbers have switched to using bootstrap badges, and are consistently coloured between the user menu overview and the menu item. There has been no changes to either the `primary` or `info` colours to make them more in harmony with our key colours (like the green currently used).

I'm still of the strong belief that we don't want to use either our `danger` colour, nor any other kind of eye-catching red, for message notifications. So I'm going to close this PR. Thanks @Dimitar5555 for your original work on this.

PR #4089, opened 2023-07-12
Replace page numbers with ID based selection for trace indexes

tomhughes:
This allows us to avoid horribly inefficient queries when people try and fetch ever large page numbers for the trace list forcing us to read ever large number of records.


app/views/traces/_trace_paging_nav.html.erb
```diff
@@ -1,26 +1,22 @@
...
+        <%= link_to t(".newer"), params.merge(:min_id => traces.first.id + 1, :max_id => nil), :class => "page-link" %>
       </li>
     <% else %>
       <li class="page-item disabled">
         <span class="page-link"><%= t(".newer") %></span>
       </li>
     <% end %>
 
-    <li class="page-item disabled">
-      <span class="page-link"><%= t(".showing_page", :page => page) %></span>
```
gravitystorm, line 14: Unused translation can be removed from `en.yml`

app/views/traces/index.html.erb
```diff
@@ -17,33 +17,33 @@
...
       <% end %>
     <% elsif current_user && current_user == @target_user %>
       <li class="nav-item">
-        <%= link_to t(".all_traces"), { :controller => "traces", :action => "index", :display_name => nil, :page => nil }, { :class => "nav-link" } %>
+        <%= link_to t(".all_traces"), { :controller => "traces", :action => "index", :display_name => nil, :max_id => nil }, { :class => "nav-link" } %>
       </li>
       <!-- my traces -->
       <li class="nav-item">
-        <%= link_to t(".my_traces"), { :action => "mine", :page => nil }, { :class => "nav-link active" } %>
+        <%= link_to t(".my_traces"), { :action => "mine", :max_id => nil }, { :class => "nav-link active" } %>
```
gravitystorm, line 33: I don't think any of these `:max_id => nil` are needed, since the link is being built from scratch (other links are built from merging with existing params elsewhere).

Or if I've misunderstood, and they are necessary, then I think a `:min_id => nil` would be needed too.

test/controllers/traces_controller_test.rb
```diff
@@ -8,51 +8,27 @@ def test_routes
       { :path => "/traces", :method => :get },
       { :controller => "traces", :action => "index" }
     )
-    assert_routing(
-      { :path => "/traces/page/1", :method => :get },
-      { :controller => "traces", :action => "index", :page => "1" }
-    )
```
gravitystorm, line 14: We should keep some some redirect tests to make sure the redirects aren't accidentally refactored away in future.

app/controllers/traces_controller.rb
```diff
@@ -54,14 +54,16 @@ def index
 
     @traces = @traces.tagged(params[:tag]) if params[:tag]
 
-    @params = params.permit(:display_name, :tag)
+    @params = params.permit(:display_name, :tag, :min_id, :max_id)
```
gravitystorm, line 57: Most implementations I've seen use "before" and "after" as the parameters. I'm not sure why, but maybe it makes it easier when reasoning about id sort order? (see other comment)

gravitystorm (changes requested):
Overall I'm happy with the idea of moving to cursor-based pagination. I'd like to do the same for all of our existing offset-based pagination, since it all has the same problems as here (degraded performance for increasing pagenumbers, highly unstable urls, etc).

In addition to the line-by-line comments, I wonder if rather than trying to get the logic and edge cases right (and making all the logic etc resuable across controllers) I wonder if we are better off finding a gem that does what we want. There's a few different ones that I've found, such as [rails_cursor_pagination](https://github.com/xing/rails_cursor_pagination), [pagy_cursor](https://github.com/Uysim/pagy-cursor), [etc](https://rubygems.org/search?query=name%3A+cursor++++updated%3A+%3E2021)

test/controllers/traces_controller_test.rb
tomhughes: I'd love to test the redirects but I couldn't find a way to do so... I think I've been here before and failed to find a way to do it :-(

app/controllers/traces_controller.rb
tomhughes: I based `max_id` on what we were already doing doing for changesets in the history view but I wasn't especially keen on it.

test/controllers/traces_controller_test.rb
gravitystorm: We use this approach elsewhere in controller tests:

```
visit foo
assert_redirected_to bar
```

Would that work for here?

tomhughes: Oh that might work sure, what I was looking for was away to do it with assert_routing/assert_recognizes rather than going through the whole stack.

gravitystorm: Yeah, I think since there's no controller involved in a redirect, it's not going to work with those assertions. https://stackoverflow.com/a/10844398/105451 has a reasonably well researched take on this, to do with how the routes are handled differently internally to rails when it's a redirect.

tomhughes:
I've pushed a new version that I think addresses all your review comments.

I did look at the possibility of using a gem but I'm not convinced there's anything that's useful for various reasons.

In many ways [pagy_cursor](https://github.com/Uysim/pagy-cursor) looks excellent, especially since it's an extension to [pagy](https://github.com/ddnexus/pagy) which provides nice view helpers and would be good to replace our existing paging code for places where we do real paging. The problem is that it doesn't really integrate with pagy in any meaningful way and you can't use the pagy navigation helpers as https://github.com/Uysim/pagy-cursor/issues/21 makes clear. It's not entirely their fault as pagy would need changes to allow it to support cursor based paging but there doesn't seem to be any effort to properly integrate them.

As to [rails_cursor_pagination](https://github.com/xing/rails_cursor_pagination) there is the issue of the rather ugly cursors but we could live with that - you could even argue it's an advantage. There's a killer bug at the moment though (https://github.com/xing/rails_cursor_pagination/issues/132) in that even if you disable totals it still does the query when working our whether next/previous links are needed!

PR #4099, opened 2023-07-19
Upgrade to bootstrap 5.2

gravitystorm:
I don't think there's anything in the migration notes that affects us. 

https://getbootstrap.com/docs/5.2/migration/

I'm way more interested in what bootstrap 5.3 has to offer (hint: colour modes!), so this is just a stepping stone to get us there.

(Previously blocked by #4088)


gravitystorm:
Eugh, something is still not right with precompiling! I'll return to this.

tomhughes:
What was the problem you ran into here? Only I just tried this locally and precompiling seemed to work and the site ran OK in both development and production modes?

gravitystorm:
` bundle exec rails assets:precompile` failed in the CI tests, and I could recreate that locally too. I think I had to clear my local cache to trigger it though.

The error message is the same as previously, namely:

```
SassC::SyntaxError: Error: Custom property values may not be empty.
        on line 1:69677 of stdin
>> t-size:1.25rem;--bs-navbar-toggler-icon-bg:url("data:image/svg+xml,%3csvg xm
```

This indicates that libsass is trying to parse it's own compressed output, which hits a bug in libsass. This is what #4088 was trying to fix - it avoided the reparsing for the main stylesheet by doing everything in one pass, and so it now works for development and testing (when running the tests without precompiling). 

I haven't figured out why the precompile step trips up, but I suspect it's something to do with merging even more stylesheets together (e.g. the leaflet or iD stylesheets or something like that) which triggers reparsing our main stylesheet.

https://github.com/gravitystorm/openstreetmap-website/issues/268 has some more details about all the different libsass vs sassc-rails vs dartsass etc - to help keep it straight in my own head!

tomhughes:
This really is very odd because I still can't reproduce it locally even with a clean checkout but I can in a CI run.

I wondered if it was because I'm using ruby 3.2 but I managed to get a CI run with that to work and that failed.

I guess we have to hope somebody fixes (or removes) the bootstrap dependency on sassc so we can use a newer sass compiler.

PR #4106, opened 2023-07-26
Replace page numbers with ID based selection for diary indexes

tomhughes:
Based on he model used for trace paging...


gravitystorm:
Great. I think we need to abstract some of this duplicated logic, but that can be done separately.

PR #4107, opened 2023-07-26
Allow administrators to see deleted diary entries

tomhughes:
Deleted diary entries are already visible in the index view but not when directly viewing them so this doesn't really change anything but does make the life of people processing reports easier.


gravitystorm (changes requested):
Looks good. I can't immediately think of a way to make the query building less clunky.

Only request from me is to add a test for the change in behaviour, to ensure the ability is retained in future.

tomhughes:
I've added tests.

While I was doing that I noticed that while moderators could hide diary entries and comments they couldn't unhide them which seemed odd - apart from anything else it means if they hit the hide link accidentally they can't undo it. So I've added a commit to extend unhide rights to them.

gravitystorm (approved):

PR #4111, opened 2023-07-28
Hide/show welcome banner with hidden attribute

AntonKhorev:
...instead of custom css hidden/visible classes.

I wanted it to be a simple example of a `hidden` attr use but unfortunately there are two additional conditions of welcome banner visibility: not in *small-nav* mode & in *overlay-sidebar* mode. Some `display: none/block` still have to be present in css.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -192,11 +192,11 @@ $(document).ready(function () {
   });
 
   if (Cookies.get("_osm_welcome") !== "hide") {
-    $(".welcome").addClass("visible");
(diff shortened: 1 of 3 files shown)


gravitystorm:
Thanks! 

Yes, it's a complex play between hiding/showing based on screen size, and then effectively overriding that based on the cookie. The bootstrap use of `[hidden] { display: none !important; }` ([documentation](https://getbootstrap.com/docs/5.1/content/reboot/#html5-hidden-attribute)) is key here, if I understand it correctly.

PR #4112, opened 2023-07-28
Remove unnecessary full-width float css from sidebar

AntonKhorev:
100% width + float is like stacking vertically. But this is the default behavior for block elements. And those divs are inside another floated element.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -422,9 +422,6 @@ body.small-nav {
 
     > div {
       position: relative;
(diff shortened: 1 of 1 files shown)


AntonKhorev:
`position: relative` is not necessary if there's nothing absolutely positioned inside in all possible situations, which I didn't test for yet.

gravitystorm:
OK, I'll merge this and we can return to the `position` stuff separately.

AntonKhorev:
Banner close buttons have position: absolute.

PR #4120, opened 2023-07-29
Remove deprecated attributes from share html

AntonKhorev:


tomhughes:
What's the impact of this? Does anything still pay attention to them? Is there anything we should be doing instead to achieve the same results or are the defaults good enough?

AntonKhorev:
The impact is shorter code to copy from a small textarea. Margins are zero by default, border is redefined in the style attribute.

gravitystorm:
Looks good to me. Since this only affects the code copy+pasted for new embeds, and doesn't affect any existing embeds, this is pretty low-risk.

Issue #2796, opened 2020-08-27
Changeset comments have individual links, but no button to grab them with


jidanni:
[Here](https://www.openstreetmap.org/changeset/20707962#c596364) I link directly to a changeset comment.

But how did I figure out how?
With Chromium Inspector (F12). That's because I'm a computer wiz.

But for normal mappers, they should be able to get the link from somewhere in the
`Comment from Nuerdensfields 24 minutes ago`
sentence. But due to some bug, they can't!

GurminderSingh7:
i want to help with this issue, could you assign this to me

AshBakchod:
Is this Issue still present or resolved?

jidanni:
Yes, it is still present.

pranshu-tomer:
I can't understand this Issue. Can you give me more clear idea about issue.

Dimitar5555:
> I can't understand this Issue. Can you give me more clear idea about issue.

When you hover over `commented X time ago` there is no link. On other platforms the same text is a link to that specific post/comment. For example in this issue, I can right click on `commented X minutes ago`, copy link and get `https://github.com/openstreetmap/openstreetmap-website/issues/2796#issuecomment-1880081307`.

gy-mate:
As far as I understand the  `.comment_by_html` translations should be changed in all ~90 languages from [`"Comment from %{user} %{time_ago}"`](https://github.com/openstreetmap/openstreetmap-website/blob/29020530a8fabe1d02e3bc9c41970c9030efc380/config/locales/en.yml#L476) to something like `"%{comment} from %{user} %{time_ago}"`¹ in order to be able to send the _Comment_ word as a preformatted link from [here](https://github.com/openstreetmap/openstreetmap-website/blob/29020530a8fabe1d02e3bc9c41970c9030efc380/app/views/changesets/show.html.erb#L35) somehow like this:

```ruby
<%= t comment.visible ? ".comment_by_html" : ".hidden_comment_by_html",
                  :comment => link_to(xxxxxx, "https://www.openstreetmap.org/changeset/#{changeset.id}##{comment.id}"),
                  :time_ago => friendly_date_ago(comment.created_at),
                  :user => link_to(comment.author.display_name, comment.author) %>
```

¹ It's difficult because the word _Comment_ should still be translated in each language, so somehow it should keep the translations _and_ add links to them.

All in all, this seems a bit more complicated to me than a _good first issue_.

kleenkanteen:
[...]

PR #4134, opened 2023-08-02
Change comment link

starsep:
SVG from https://icons.getbootstrap.com/icons/link/

image

### DIFF
--- a/app/views/browse/changeset.html.erb
+++ b/app/views/browse/changeset.html.erb
@@ -46,6 +46,12 @@
                     — <span class="action-button" data-comment-id="<%= comment.id %>" data-method="POST" data-url="<%= changeset_comment_hide_url(comment.id) %>"><%= t("javascripts.changesets.show.hide_comment") %></span>
                   <% end %>
                 </small>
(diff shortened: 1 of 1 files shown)


tomhughes:
Please, don't just write "change comment link" as it tells us nothing.

Instead explain how you are changing the link and why - what you are aiming to achieve and why that is a good thing that we should want to merge.

tomhughes:
It looks like you're actually adding a new link anyway, not changing one?

starsep:
Sorry for being clear.
I am adding button (link with icon) to changeset comments.
It allows to easily copy link and reference it in external communication. It's useful when discussions get long.
At moment there is no easy way to get this link as explained in https://github.com/openstreetmap/openstreetmap-website/issues/2796

AntonKhorev:
There's also a link icon in already existing sprites:
https://github.com/openstreetmap/openstreetmap-website/blob/77d495ccdc4c6d07cb7795177c73ba541afc44e0/app/assets/stylesheets/common.scss#L56C14

PR #4141, opened 2023-08-07
Avoid using partial rendering in /map.json

stillhart:
Using partial rendering too many times is slow in rails and a known issue. This is because the whole render context is given to the partial. This view is using this extensively and thousands of times. Just inlining the "osm object" should improve the performance. At the same time a lot of code can be de-duplicated. I kept a fallback in case of new objects. 

Related rails issue: https://github.com/rails/rails/issues/41452

That halfed rendering times in my case, and reduced allocations by 20%. 
```log
# With my patch:
web_1  | Completed 200 OK in 232ms (Views: 108.1ms | ActiveRecord: 31.3ms | Allocations: 221573)
web_1  | Completed 200 OK in 219ms (Views: 115.8ms | ActiveRecord: 21.0ms | Allocations: 218252)
web_1  | Completed 200 OK in 254ms (Views: 143.9ms | ActiveRecord: 22.9ms | Allocations: 218254)
web_1  | Completed 200 OK in 222ms (Views: 115.7ms | ActiveRecord: 21.9ms | Allocations: 218313)

# Without my patch:
web_1  | Completed 200 OK in 438ms (Views: 332.1ms | ActiveRecord: 22.8ms | Allocations: 253255)
web_1  | Completed 200 OK in 398ms (Views: 295.7ms | ActiveRecord: 21.8ms | Allocations: 250814)
web_1  | Completed 200 OK in 447ms (Views: 343.2ms | ActiveRecord: 22.5ms | Allocations: 250473)
web_1  | Completed 200 OK in 448ms (Views: 338.7ms | ActiveRecord: 25.2ms | Allocations: 250530)
```

### DIFF
--- a/app/views/api/map/index.json.jbuilder
+++ b/app/views/api/map/index.json.jbuilder
@@ -4,6 +4,42 @@ json.partial! "bounds"
 
 all = @nodes + @ways + @relations
 
-json.elements(all) do |obj|
-  json.partial! obj
+json.elements(all) do |object|
+  case object
+  when Relation, Node, Way
+    json.id object.id
+    json.timestamp object.timestamp.xmlschema
+    json.version object.version
+    json.changeset object.changeset_id
+    json.user object.changeset.user.display_name
+    json.uid object.changeset.user_id
+    json.tags object.tags unless object.tags.empty?
+    json.visible object.visible unless object.visible
+
+    case object
+    when Relation
+      relation = object
+      json.type "relation"
+      unless relation.relation_members.empty?
+        json.members(relation.relation_members) do |m|
+          json.type m.member_type.downcase
+          json.ref m.member_id
+          json.role m.member_role
+        end
+      end
+    when Node
+      node = object
+      json.type "node"
+      if node.visible
+        json.lat GeoRecord::Coord.new(node.lat)
+        json.lon GeoRecord::Coord.new(node.lon)
+      end
+    when Way
+      way = object
+      json.type "way"
+      json.nodes way.nodes.ids unless way.nodes.ids.empty?
+    end
+  else
+    json.partial! object
+  end
 end


stillhart:
Interesting. Just had a brief look into cgimap. Kinda sad that the openstreetmap mirror is not up-to-date, but at least here is a link to the actual project :-)

General question, is there a reason that cgimap is not included in the docker-compose of openstreetmap-website in some way? I would assume it's because a multi-repo setup is not fun to work with.

Second question: Now that this is usually served with cgimap, is there a reason not to accept my merge request? It still enhances the development experience. Also I would like to further improve the /map.json endpoint, since it seems to be implemented rather unoptimized. I can see tons of n+1 queries which probably could be fixed with some simple joins and includes.

tomhughes:
Well the docker container is for development work so performance is not the concern in the way it is in production.

Putting a precompiled cgimap in the container would be useless for anybody wanting to do development on cgimap and it's not clear to me how it helps people working on the ruby code.

No amount of optimisation of the ruby code for map is ever going to get anywhere near the performance we need for production - even if you can optimise the SQL the overhead of creating ruby objects for everything kills you. We've been there, done that, and moved on to a C++ implementation.

gravitystorm:
Thanks @stillhart for the PR. As @tomhughes says, this code isn't used in production on openstreetmap.org. However, I'm happy to receive pull requests to improve performance. Firstly, because these techniques can also apply to other parts of the codebase, and secondly, to help when developers are testing things locally (e.g. working on the data overlay UI, which uses the map call). So while there's limited benefits, there are still benefits to be had.

For this PR, it's effectively inlining the code from the three element partials. This means that it's duplicating code - the original `app/views/api/nodes/_node.json.jbuilder`partials still exists. Is there some way that we can avoid this duplication, so that there is only one part of code that generates the json for a way, instead of the same code appearing multiple times?

Secondly, I would like to keep the approach the same for JSON and XML. So if we do something for the json (e.g. inlining) then we should do the same for the XML, since (presumably?) the same problem applies there too.

You mention N+1 queries, and those are definitely worth solving, that might be worth doing in a separate PR :+1:

mmd-osm:
> You mention N+1 queries, and those are definitely worth solving, that might be worth doing in a separate PR

Changeset and User Loads seem to be one of the culprits, although they mostly end up being served by the cache. Preloading might be all it takes to get rid of them.

Nodes:

```
  CACHE Changeset Load (0.0ms)  SELECT "changesets".* FROM "changesets" WHERE "changesets"."id" = $1 LIMIT $2  [["id", 1874767], ["LIMIT", 1]]
  ↳ app/views/api/nodes/_node.xml.builder:7
  CACHE User Load (0.0ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ↳ app/views/api/nodes/_node.xml.builder:7
```


Ways:

```
  CACHE Changeset Load (0.0ms)  SELECT "changesets".* FROM "changesets" WHERE "changesets"."id" = $1 LIMIT $2  [["id", 1874767], ["LIMIT", 1]]
  ↳ app/views/api/ways/_way.xml.builder:7
  CACHE User Load (0.0ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ↳ app/views/api/ways/_way.xml.builder:7
  Node Load (1.2ms)  SELECT "current_nodes".* FROM "current_nodes" INNER JOIN "current_way_nodes" ON "current_nodes"."id" = "current_way_nodes"."node_id" WHERE "current_way_nodes"."way_id" = $1 ORDER BY "current_way_nodes"."sequence_id" ASC  [["way_id", 4000394980]]
  ↳ app/views/api/ways/_way.xml.builder:12
```

mmd-osm:
Some sites like https://scoutapm.com/blog/performance-impact-of-using-ruby-on-rails-view-partials seem to suggest that calling partials in a loop isn't the best idea. Alternative option might be to pass a collection to the partial instead (which happens to avoid the DRY issue mentioned earlier).

I can create a PR if this commit is of interest: https://github.com/mmd-osm/openstreetmap-website/commit/8c20bc1d98156af2501c9c4ec324473dc72a0994

mmd-osm:
Regarding N+1 queries, I tried a gem called 'bullet', which needs a bit of config in config/environments/development.rb:

```
  config.after_initialize do
    #Enable bullet in your application
    Bullet.enable = true
    Bullet.rails_logger = true
  end  
```

It would then complain about N*1 queries, or suggest to remove eager loading where not needed. 

```
GET /changeset/127673
USE eager loading detected
  OldRelation => [:redaction]
  Add to your query: .includes([:redaction])
Call stack
```

As an example, the suggested changes to "changeset_controller.rb - show" would be to add additional :includes for old_nodes/ways/relations:  `:include => [:old_tags, :redaction]`

```
    @node_pages, @nodes = paginate(:old_nodes, :include => [:old_tags, :redaction], :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "node_page")
    @way_pages, @ways = paginate(:old_ways, :include => [:old_tags, :redaction], :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "way_page")
    @relation_pages, @relations = paginate(:old_relations, :include => [:old_tags, :redaction], :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "relation_page")
 
```

AntonKhorev:
> As an example, the suggested changes to "changeset_controller.rb - show" would be to add additional :includes for old_nodes/ways/relations

See #4569

mmd-osm:
Rendering was improved in #4958, avoiding the DRY issues mentioned earlier.

PR #4144, opened 2023-08-08
Various css/bootstrap updates to diary pages

AntonKhorev:
- paddings and margins with bootstrap classes
- deleted post/comment background with bootstrap classes
- "Use map" in diary post form as a button
- `<article>` elements for posts


gravitystorm:
Thumbs up from me on this. But please note we try wherever possible to avoid setting margin-top, as per bootstrap guidelines:

https://getbootstrap.com/docs/5.1/content/reboot/#approach

PR #4148, opened 2023-08-09
Use correct extension and type for avatar images attached to mails

### LINKED ISSUE #4147: In the mail about someone solved an issue on the map the attached image has a wrong extension, png instead of jpg
### URL

_No response_

I got a mail about an issue has been solved by someone else. In the attachments there is an avatar of the person. The image has .png but the file format is .jpg.
There were no further problems with the mail program, this report is just for the records.

avatar
```
file avatar.png
avatar.png: JPEG image data, Exif standard: [TIFF image data, big-endian, direntries=13, height=3262, manufacturer=samsung, model=SM-G998B, orientation=upper-left, xresolution=206, yresolution=198, resolutionunit=2, software=G998BXXU5CVGB, datetime=2022:08:27 19:07:36, GPS-Data, width=2342], baseline, precision 8, 36x50, components 3
```

_No response_

tomhughes:
Fixes #4147

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -187,7 +187,25 @@ class UserMailer < ApplicationMailer
   end
 
   def attach_user_avatar(user)
-    attachments.inline["avatar.png"] = user_avatar_file(user)
+    @avatar = user_avatar_filename(user)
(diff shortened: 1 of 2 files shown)


gravitystorm:
Looks good to me.

PR #4149, opened 2023-08-10
Delete button in Home location section of user profiles

### LINKED ISSUE #3725: Mention how to remove profile location
On the edit profile page, if one has already set a location, there is no way to remove it apparently. One can only just pick a different location.

 But in fact the method to remove it is just to make the latitude and latitude longitude fields blank and then push Update Profile at the bottom. So perhaps mention this on that page. Because sometimes people might want to remove their location.

I can't find the exact file to put the words in to make a PR.

AntonKhorev:
home

- also undo delete button
- also show button to center map on currently entered coordinates
- also `has_home?` method in user models instead of doing home_lat && home_lon

Fixes #3725

### DIFF
--- a/app/assets/javascripts/user.js
+++ b/app/assets/javascripts/user.js
@@ -51,25 +51,41 @@ $(document).ready(function () {
       }
 
       map.on("click", function (e) {
-        if ($("#updatehome").is(":checked")) {
-          var zoom = map.getZoom(),
-              precision = OSM.zoomPrecision(zoom),
-              location = e.latlng.wrap();
+        if (!$("#updatehome").is(":checked")) return;
 
-          $("#home_lat").val(location.lat.toFixed(precision));
-          $("#home_lon").val(location.lng.toFixed(precision));
+        var zoom = map.getZoom(),
+            precision = OSM.zoomPrecision(zoom),
+            location = e.latlng.wrap();
 
-          respondToHomeUpdate();
-        }
+        $("#home_lat").val(location.lat.toFixed(precision));
+        $("#home_lon").val(location.lng.toFixed(precision));
+
+        deleted_lat = null;
+        deleted_lon = null;
+        respondToHomeUpdate();
       });
 
-      $("#home_lat, #home_lon").on("input", respondToHomeUpdate);
+      $("#home_lat, #home_lon").on("input", function () {
+        deleted_lat = null;
+        deleted_lon = null;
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/user.js
```diff
@@ -73,6 +127,41 @@ $(document).ready(function () {
...
+    } else {
+      marker.removeFrom(map);
+    }
+  }
+
+  function isCloseEnoughToMapCenter(location) {
+    var inputPt = map.latLngToContainerPoint(location),
+        centerPt = map.latLngToContainerPoint(map.getCenter());
+
+    return Math.abs(inputPt.x - centerPt.x) + Math.abs(inputPt.y - centerPt.y) < 10;
```
tomhughes, line 162: Would `distanceTo` work here to get the distance between two points? or just use `map.getBounds().contains(location)` to see if the location is already visible?

app/assets/javascripts/user.js
```diff
@@ -35,32 +38,83 @@ $(document).ready(function () {
...
+            precision = OSM.zoomPrecision(zoom),
+            location = e.latlng.wrap();
+
+        $("#home_lat").val(location.lat.toFixed(precision));
+        $("#home_lon").val(location.lng.toFixed(precision));
+
+        deleted_lat = null;
+        deleted_lon = null;
+        respondToHomeUpdate();
+      }).on("moveend", function () {
```
tomhughes, line 71: Why are we using a `moveend` event to detect typed changes in the input fields? At least I assume that is what this is trying to detect given we already have a click handler....

AntonKhorev: We are detecting if the map location is close enough to the inputs location. Then based on it we enable/disable the *show* button.

app/assets/javascripts/user.js
AntonKhorev: Yes, `distanceTo` would work. Updated.

gravitystorm:
The ruby style guide recommends against using auxiliary verbs in boolean methods:

https://rubystyle.guide/#bool-methods-prefix

So `User.has_home?` should avoid the `has_` prefix. Perhaps `User.home_location?` instead?

(This would have been flagged up by rubocop for any other model, but it's currently disabled for the User model due to the existing `has_role?` method which has yet to be refactored.)

AntonKhorev:
`has_role?` is the reason why I wrote `has_home?`

PR #4150, opened 2023-08-10
Use Bootstrap highlight class in note tables

AntonKhorev:
https://getbootstrap.com/docs/5.1/content/tables/#active-tables


AntonKhorev:
Let's try primary color instead.

gravitystorm:
Thanks @AntonKhorev

PR #4156, opened 2023-08-11
Don't do direction override from Bootstrap reboot of <code>

### LINKED ISSUE #4104: Code blocks in diary enteries appear reversed, with right-to-left locale
### Description

Code blocks in diary enteries appear reversed, with right-to-left locale

### URL

https://www.openstreetmap.org/user/iriman/diary/401796?locale=fa

1. Visit [https://www.openstreetmap.org/user/iriman/diary/401796](https://www.openstreetmap.org/user/iriman/diary/401796?locale=fa)**?locale=fa**
2. Find "bread:wholemeal=yes"
3. It's reversed!

image

`unicode-bidi: embed` maybe is better than "bidi-override"?

AntonKhorev:
Later versions of Bootstrap don't have it.
See https://github.com/twbs/bootstrap/pull/35230

Fixes #4104

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1233,6 +1233,8 @@ div.secondary-actions {
   code {
     background: $lightgrey;
     padding: 2px 3px;
+    direction: inherit; /* fix for Bootstrap < 5.2 */
+    unicode-bidi: unset; /* fix for Bootstrap < 5.2 */
   }
 
   pre {


tomhughes:
If I'm reading things right this is something that's going to happen anyway when we go to bootstrap 5.2 which will be hopefully be fairly soon if @gravitystorm can solve his problems?

AntonKhorev:
Upgrading to Bootstrap 5.2 is also going to solve this problem. This pull request is a fix in case you don't want to wait fo the upgrade.

PR #4158, opened 2023-08-12
Add order parameter to changeset query API call

AntonKhorev:
When you want to get an unknown number of changesets matching some criteria, you use `/api/0.6/changesets` with the corresponding parameters. If you receive a number of changesets that is equal to the limit, you know there might be more matching changesets. To get the next page of changesets you set the upper limit of the `time` parameter to the last received changeset.* The lower limit is usually set to some date before osm existed. Then you run the query again and receive the next page of results. That next page should consist of changesets created earlier than the ones received previously.

But what if you want a previous page? You can't set the upper time limit to the first received changeset because you'll get the same result again.* You can set the lower limit, but then you'll still need to decide on the upper limit. If you don't set the upper limit, you'll receive the first page with the latest changesets, not the previous one. You'll have to guess the upper time limit of the previous page, and guessing it correctly may take several requests.

If you could get changesets in ascending order, specifying lower limit would be enough to get the previous page.*

This pull request adds the `order` parameter to `/api/0.6/changesets`, similar to `/api/0.6/notes/search`. `order=oldest` returns changesets in ascending order.

\* some details about same-second and overlapping changesets omitted

### DIFF
--- a/app/controllers/api/changesets_controller.rb
+++ b/app/controllers/api/changesets_controller.rb
@@ -170,8 +170,15 @@ module Api
       changesets = conditions_closed(changesets, params["closed"])
       changesets = conditions_ids(changesets, params["changesets"])
 
-      # sort and limit the changesets
-      changesets = changesets.order("created_at DESC").limit(result_limit)
+      # sort the changesets
+      changesets = if params[:order] == "oldest"
+                     changesets.order("created_at ASC")
+                   else
+                     changesets.order("created_at DESC")
+                   end
+
+      # limit the result
+      changesets = changesets.limit(result_limit)
 
       # preload users, tags and comments, and render result
       @changesets = changesets.preload(:user, :changeset_tags, :comments)
--- a/test/controllers/api/changesets_controller_test.rb
+++ b/test/controllers/api/changesets_controller_test.rb
@@ -1968,6 +1968,41 @@ module Api
     end
 
     ##
+    # test the query functionality of changesets with the order parameter
+    def test_query_order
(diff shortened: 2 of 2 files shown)


mattfbacon:
> then you'll still need to decide on the upper limit.

Can't you just use the same approach as with the lower limit and set it to a value far far into the future?

AntonKhorev:
> Can't you just use the same approach as with the lower limit and set it to a value far far into the future?

Setting the upper limit far into the future is the same as not setting it at all. You'll get the "first page" of changesets, not the previous page.

mattfbacon:
Ah,I see, it always chooses the N most recent changes within the range.

AntonKhorev:
Looks like there are performance problems now. When the query is limited by a user, it runs ok. When it asks for every changeset in a specified time window, it's slow. Apparently the db doesn't like `WHERE (... and created_at <= $sometime) ORDER BY created_at ASC` with the where condition and the order going in different directions.

tomhughes:
Are you saying existing queries have got slower? or new ones that weren't possible before are slow?

AntonKhorev:
the new ones with `order=oldest`

AntonKhorev:
Example of slow request

`https://api.openstreetmap.org/api/0.6/changesets.json?limit=10&time=20200501T000000Z&order=oldest`

It probably goes through almost the entire created_at index because there's no condition created_at. The lower bound is checked only on closed_at. In the fix I added a condition that skips changesets that are definitely too old to be open at the lower bound (2 * max age).

PR #4159, opened 2023-08-12
Add JSON output to /api/0.6/capabilities

AntonKhorev:


test/controllers/api/capabilities_controller_test.rb
```diff
@@ -32,5 +36,28 @@ def test_capabilities
...
+      assert_equal Settings.tracepoints_per_page, js["api"]["tracepoints"]["per_page"]
+      assert_equal Changeset::MAX_ELEMENTS, js["api"]["changesets"]["maximum_elements"]
+      assert_equal Settings.default_changeset_query_limit, js["api"]["changesets"]["default_query_limit"]
+      assert_equal Settings.max_changeset_query_limit, js["api"]["changesets"]["maximum_query_limit"]
+      assert_equal Settings.max_number_of_relation_members, js["api"]["relationmembers"]["maximum"]
+      assert_equal "online", js["api"]["status"]["database"]
+      assert_equal "online", js["api"]["status"]["api"]
+      assert_equal "online", js["api"]["status"]["gpx"]
+      assert_equal Settings.imagery_blacklist.length, js["policy"]["imagery"]["blacklist"].length
+    end
```
mattfbacon, line 61: These tests generally look good, but they implicitly allow adding more keys to the JSON without changing the tests, which may not be desirable. If you constructed the object that we expect and then compared it with `js`, that might work better to test against added keys.

test/controllers/api/capabilities_controller_test.rb
```diff
@@ -9,6 +9,10 @@ def test_routes
         { :path => "/api/capabilities", :method => :get },
         { :controller => "api/capabilities", :action => "show" }
       )
+      assert_routing(
+        { :path => "/api/capabilities.json", :method => :get },
+        { :controller => "api/capabilities", :action => "show", :format => "json" }
+      )
       assert_recognizes(
         { :controller => "api/capabilities", :action => "show" },
         { :path => "/api/0.6/capabilities", :method => :get }
```
mattfbacon, line 18: Should there be another `assert_recognizes` here for the JSON route?

app/views/api/capabilities/show.json.jbuilder
```diff
@@ -0,0 +1,44 @@
...
+  end
+end
+
+json.policy do
+  json.imagery do
+    json.blacklist(Settings.imagery_blacklist) do |url_regex|
+      json.regex url_regex.to_s
+    end
+  end
+end
```
mattfbacon, line 44: This looks good as a verbatim copy of the XML builder (though I haven't looked at it in depth), but I wonder it it would be possible to consolidate this JSON generation and the XML generation into a single "data generation" step which would create a Ruby data structure, followed by a serialization step which could dynamically choose the format: JSON or XML.

mattfbacon (changes requested):

test/controllers/api/capabilities_controller_test.rb
AntonKhorev: The same is true for xml tests. Actually I've already added xml attributes without thinking about the test an changing anything in it.

test/controllers/api/capabilities_controller_test.rb
tomhughes: Yes it does need that so I've pushed a commit to add it...

PR #4164, opened 2023-08-13
Filter out changesets that are too old when sorting in ascending order

AntonKhorev:
Possible fix for https://github.com/openstreetmap/openstreetmap-website/pull/4158#issuecomment-1676405939

### DIFF
--- a/app/controllers/api/changesets_controller.rb
+++ b/app/controllers/api/changesets_controller.rb
@@ -165,7 +165,7 @@ module Api
       changesets = Changeset.all
       changesets = conditions_bbox(changesets, bbox)
       changesets = conditions_user(changesets, params["user"], params["display_name"])
-      changesets = conditions_time(changesets, params["time"])
+      changesets = conditions_time(changesets, params["time"], params[:order] == "oldest")
       changesets = conditions_open(changesets, params["open"])
       changesets = conditions_closed(changesets, params["closed"])
       changesets = conditions_ids(changesets, params["changesets"])
@@ -328,7 +328,7 @@ module Api
 
     ##
     # restrict changes to those closed during a particular time period
-    def conditions_time(changesets, time)
+    def conditions_time(changesets, time, lower_bound_created_condition)
       if time.nil?
         changesets
       elsif time.count(",") == 1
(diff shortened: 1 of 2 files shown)


tomhughes:
The real problem is that we're selecting on close time by sorting on start time - if the sort and condition were on the same time then things would be fine.

Start and close time are very closely correlated so there's seems to be no reason not to use the same time for both and without doing that I think that trying to page through the changesets may even miss some?

tomhughes:
I've changed it to sort by close time which should fix this for now but the whole thing seems a bit of a mess...

It would be better if everything used creation time because that would allow better optimisation when a user limit is included as we have an index on user+creation_time but not on user+close_time.

The time selection is also weird in that if you provide two bounds one is evaluated against close time and one against creation time?

AntonKhorev:
Time selection checks if a changeset was open at some time inside the range. I don't know why this check was implemented instead of checking if a changeset was created inside the range.

AntonKhorev:
If you sort by ascending close time, it's not necessarily an inverse of sorting by descending open time.

AntonKhorev:
Originally there was a simple `created_at` check added in https://github.com/openstreetmap/openstreetmap-website/commit/b875d14b7cd13432b0683ca548b341dcd8bddf80. Ten days later it was changed to open state time span check that uses both `created_at` and `closed_at` in https://github.com/openstreetmap/openstreetmap-website/commit/a90be5e69a478e2b49ae676b649a78589f85a50e.

tomhughes:
Are you actually a user of this API or was there some other motivation from this work? 

Ideally I'd like to find somebody that actually uses it to see if we can just make everything use the creation time which helps with a lot of things - it helps with optimisation and is also stable while the close time can change until a changeset is completed.

tomhughes:
Sorry I missed your reply yesterday...

I'm not opposed to changing the current time parameter (although I don't particularly like it) but what I 'm trying to understand is what uses of the API want - do they particularly want selection by one time or the other? Is stability (that repeatedly hitting the same search finds the same changesets) important?

Those are the kind of things that will help decide what the answer to all this is.

AntonKhorev:
Stability for changeset queries always existed because they were always sorted by creation time.

tomhughes:
That makes the ordering stable but not the selection - as the selection was (primarily) on close time the set of selected changesets for a given time value could change between calls.

I think that can only happen if you use a selection time that is after the current time though, because any changeset that has a close time before the current time is closed and can't have it's close time move to later - if you use a selection time in the future then a changeset that had a close time before that can change to have a close time after it and hence get added to the selection for a future call.

AntonKhorev:
> as the selection was (primarily) on close time

It wasn't *primarily*. It was equally on open and close time.

> the set of selected changesets for a given time value could change between calls
> I think that can only happen if you use a selection time that is after the current time though

If you use a selection time after the current time, you are able to change the selected set by opening a new changeset. No one would expect such set to be stable.

AntonKhorev:
Someone might want to sort changesets by close time if they want to show closing events independently. In [my changeset viewer](https://github.com/AntonKhorev/osm-changeset-viewer) there's an option to show them with output looking like this:

cset-viewer-closed-csets

Most of changesets on this image were made by StreetComplete. They can overlap in all kinds of ways. Sorting by open time and by close time may give significantly different results. Therefore any sudden change to sorting in the API may break already written code. For example, [this paging code](https://github.com/woodpeck/osm-revert-scripts/blob/54e99458bd029030b7e0d42141d392bc8f2ae4e2/UserChangesets.pm#L80) relies on changesets being sorted by `created_at`.

Here's a small example of changesets that change order depending on if they are sorted by open or close time: https://api.openstreetmap.org/api/0.6/changesets?user=550156&time=20220923T185808Z,20220923T195831Z

tomhughes:
Resolved by #4171.

PR #4168, opened 2023-08-14
Restore sorting by created_at for descending order changeset queries

AntonKhorev:
The fix https://github.com/openstreetmap/openstreetmap-website/commit/702c071e191e0c79537f2e68f1084aa42bcb3c7a changed all changeset sorting to closed_at. This altered the behavior existing before the `order` parameter and made descending-order queries slow.

### DIFF
--- a/app/controllers/api/changesets_controller.rb
+++ b/app/controllers/api/changesets_controller.rb
@@ -174,7 +174,7 @@ module Api
       changesets = if params[:order] == "oldest"
                      changesets.order(:closed_at => :asc)
                    else
-                     changesets.order(:closed_at => :desc)
+                     changesets.order(:created_at => :desc)
                    end
 
       # limit the result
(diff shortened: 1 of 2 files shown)


AntonKhorev:
> I think we need to stop making random changes here and sit back and think about what we're trying to achieve and how best to do it.

Currently the first thing I want is to restore the previous behavior when the order parameter is not used.

Try running a query *without* the order parameter:
https://api.openstreetmap.org/api/0.6/changesets.json?limit=10&time=20200101T000000Z,20200201T000000Z

> I don't think sorting by different fields when moving in different direction is going to work is it - if you start moving forward and then switch to moving backward you're going to find yourself getting different changesets?

This was a fix that didn't move in different directions:
https://github.com/openstreetmap/openstreetmap-website/pull/4164

tomhughes:
Yes by using a heuristic hack and not even a very good one - why two times max open when one would do? Even one is 24 hours though which is a lot of changesets to read and discard.

AntonKhorev:
One is not guaranteed to do because it could be off by a little. I assume there aren't any db constraints to make sure that open and close times differ no more than a day.

tomhughes:
Resolved by #4171.

PR #1096, opened 2015-11-30
Facebook emails are verified

Zverik:
E-mail addresses aquired bia Facebook Graph API (that is, OAuth) are always verified. If not, they are not returned at all. See [this stackoverflow explanation](http://stackoverflow.com/questions/14280535/is-it-possible-to-check-if-an-email-is-confirmed-on-facebook).

This PR enables confirm-less registration workflow for users registering via Facebook.

PR #4169, opened 2023-08-14
Use omniauth-microsoft_graph instead of omniauth-windowslive

milan-cvetkovic:
Omniauth-microsoft_graph correctly populates 'email' and 'name' fields used by OpenStreetMap. It also  uses updated endpoints for Microsoft identity provider.

Use email address returned by microsoft_graph provider as a verified address.

Upgrading exisiting users from windowslive to microsoft_graph:
 - upon next login exisitng windowslive users will have to authorize OpenStreetMap application to "Read Your Profile," required for proper reading of display name field.

The name of the identity provider in OSM is kept to 'windowslive':
 - the entries in users table with `provider == 'windowslive'` can be reused for microsoft_graph provider, since the uid field is preserved. Users will not need to repeat the sign up process.
 - OAuth2 callback is still `/auth/windowslive`, no updates to Microsoft Identity Provider portal App registration are necessary.


milan-cvetkovic:
This PR keeps the name of auth provider to "windowslive" to avoid database migration and re-configuring the OSM registration at  Azure  portal. It also reduces the number of affected files.

I can update the PR to rename the provider as well - this may make the code more maintainable in the long run.

@tomhughes - PR skips validation of emails for "Login with Microsoft.' since it does not appear to be possible to make an account without having a valid email - identity provider verifies it at the time of their account creation.

tomhughes:
But they're not always MS controlled emails? They can be from arbitrary third party providers, they just get verified at signup time?

tomhughes:
I don't know what @gravitystorm thinks but thinking about it I'm leaning to wards changing the provider name - there's only about 9200 records in the database that would need to be migrated and I've managed to find the app registration so I could add new callback URLs.

I'm thinking we should change it to `microsoft` rather than `microsoft_graph` though?

gravitystorm:
> upon next login **all users** will have to authorize OpenStreetMap application to "Read Your Profile," required for proper reading of display name field.

I assume this just refers to all users who currently have a windowlive omniauth auth provider, is that correct? Or have I misunderstood and it will apply to all osm.org users?

> I don't know what @gravitystorm thinks but thinking about it I'm leaning towards changing the provider name

I don't have any opinions on this. If it's straightforward to do, then I'm in favour of aligning the names, so that in future there's no confusion for any developers or site admins.

milan-cvetkovic:
@gravitystorm 
> I assume this just refers to all users who currently have a windowlive omniauth auth provider, is that correct? Or have I misunderstood and it will apply to all osm.org users?

Only users currently using `windowslive` id provider, would have to authorize OpenStreetMap for "Read Your Profile" access once they start using microsoft_graph provider. It would not apply to all osm.org users.

milan-cvetkovic:
@tomhughes
> But they're not always MS controlled emails? They can be from arbitrary third party providers, they just get verified at signup time?

This is correct. I assume the situation is similar with facebook login, but I cannot confirm since I am not their user... Maybe we can think of a separate strategy to detect and address stale emails in general case.

milan-cvetkovic:
Let me know what the preferred name is for the provider, so I can update the PR with that in mind.

I do have a commit ready if using "microsoft_graph" as the name. It touches more files, due to localized strings.

tomhughes:
I think we should just go with `microsoft` as it does appear to an extent in the UI and `microsoft_graph` is quite techy and not something I think users will recognise.

milan-cvetkovic:
@tomhughes - Do the end users ever see the name of the provider - it determines the oauth2 endpoiint and the names of the string and image resources?

milan-cvetkovic:
I have updated PR to rename the identity provider to "microsoft". I included the migration script to update auth_provider for existing "windowslive" users.

It will be necessary to update the OpenStreetMap App Registration on Azure portal - add https://openstreetmap.org/auth/microsoft/callback to the list of allowed return_url values.

tomhughes:
Thanks.

Our app is actually registered on https://apps.dev.microsoft.com/ rather than the azure portal. I know that's not allowing new registrations any more but I think I can still update the existing app - is that going to work OK with the new plugin or do we need to register a new app on the azure portal?

milan-cvetkovic:
@tomhughes 
> Our app is actually registered on https://apps.dev.microsoft.com/

I believe it should work, but I have no means of verifying. I tested this with my own app registration on portal, after setting up `microsoft_auth_id` and `microsoft_auth_secret` in `config/settings.yml`. If you have those values, you should be able to test it with local copy of OpenStreetMap server.

tomhughes:
Well except we already relaxed that when we allowed facebook supplied emails without validation in #1096 .

tomhughes:
I have merged and deployed this.

One note for future reference - the locales (other than `en.yml`) should not be modified as we only change those via TranslateWiki. I edited the PR to remove those changes before I merged it.

It turned out I did need to register a new app via the Azure portal as the old one didn't seem to work even with the extra URL added.

PR #4172, opened 2023-08-15
Fix predicate method names in the user model

tomhughes:
This renames the two `has_xxx?` predicate methods in the user model and enables the rubocop warning that catches them.


gravitystorm:
Thanks!

PR #4173, opened 2023-08-16
Small changes of diary comments

AntonKhorev:
includes margin-top removal per https://github.com/openstreetmap/openstreetmap-website/pull/4144#issuecomment-1677076131


gravitystorm:
Thanks!

PR #4176, opened 2023-08-17
Override bootstrap's box-shadow effect on search field & sidebar buttons

kimberlymarkley:
In an effort to address the issue regarding a search bar redesign opened a couple of years ago here:
https://github.com/openstreetmap/openstreetmap-website/issues/3123
and associated pull request here:
https://github.com/openstreetmap/openstreetmap-website/pull/3400

I have written a tiny bit of code to subdue the bootstrap CSS that adds a distracting blue box-shadow to the Search field and Go/Direction buttons. It is by no means a redesign, but by getting rid of the drop shadow, the Go Button no longer looks so misplaced while the user is typing.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -265,6 +265,11 @@ body.small-nav {
 
     .search_forms {
       display: block;
+      
+      /*override bootstrap's box-shadow */
+      .form-control:focus {
+        box-shadow: none!important;
+      }
     }
   }
 


tomhughes:
I'm not sure if this is a good idea or not - mostly I'm concerned about accessibility since the shadow help show the active component for people using keyboard navigation.

If it is a good idea then in keeping with our policy of minimising custom CSS the way to do it is to add the bootstrap `shadow-none` class to the components rather then using custom CSS with a `!important` flag.

AntonKhorev:
The Go button was changed a couple of times since I first commented. Does it still look misplaced?

gravitystorm:
@kimberlymarkley Thank you for your PR!

I agree with @tomhughes, that the shadow is part of the accessibility needs for keyboard navigation. And additionally, if we did decide to remove the shadow, we would do so using the class approach, rather than custom CSS.

So I'm going to decline this PR, but I would welcome other PRs to help improve the search experience.

PR #4179, opened 2023-08-17
Add color keys to note tables

AntonKhorev:
Some of the rows in a note table are highlighted, but it's not clear why.

A hint to this color coding is added here.

notes-key

Requires some modification of i18 strings. Before it's done everything looks like it did before.


gravitystorm:
@tomhughes I think this is going to have broken the translations, since the interpolations have changed with the key being kept the same.

tomhughes:
I checked that - it's actually OK because it is adding new interpolations so they will just be ignored until the string which uses them is updated.

Removing interpolations would break things but adding them is fine.

gravitystorm:
gotcha, thanks.

PR #4182, opened 2023-08-18
Add JSON output to /api/versions

AntonKhorev:


tomhughes:
Merged, with the addition of c58f4e4813bd80ba8f082b6925fc2f92aa1eea7d to ensure it defaults to XML output - the same fix had to be added to the capabilities change in f0590458484a5bb9a63e549303df7275cf58a69b after it was deployed when I realised it was defaulting to JSON and breaking iD.

Issue #1273, opened 2016-08-31
Request: Printable Directions With Highlighted Route


PorcelainMouse:
After using the directions or routing feature of OSM, directions are displayed over the map on the left side.  However, when printing such a web-page (in Firefox, at least), the directions are missing and only the map is printed.  Please add some way to make a hard copy of the directions and the map.  It doesn't need to look exactly as displayed, if that is more difficult.

mmd-osm:
See [Help OSM](https://help.openstreetmap.org/questions/51776/how-to-print-driving-directions) for full discussion.

CodeMouse92:
Agreed, this is the only thing preventing me from switching to OSM!

michaelblyons:
Seems like an odd oversight to affirmatively remove the directions list from the `print` stylesheet. 😕

tomhughes:
Do you have any evidence that is what is happening?

I suspect it's more likely that surrounding chrome (like the sidebar) was removed in order to include as much of the map as possible and that has the side effect of removing routing information.

That said, it is explicitly not the goal of openstreetmap.org to provide a full service google maps replacement for end users, so being able to print directions to take with you on a journey or whatever is not really in scope as a goal.

michaelblyons:
It's the whole sidebar, yes.
```css
header, #sidebar, #permalink, .leaflet-control {
  display: none;
}
```

landonjefftaylor:
I fully support this request. I drive frequently in rural areas without cell signal and it's useful to have printed driving directions. Screen snipping works okay, but it would be so nice to have a decent print feature like other map tools have.

gravitystorm:
This is mostly fixed by #4184. 

However, there are several features that are not implemented, or that could be improved, so it's not a complete solution. I hope that it is sufficient for most people's needs, since we continue to focus our efforts on other priorities.

PR #4184, opened 2023-08-18
Show sidebar in print media

AntonKhorev:
Prints sidebar contents before the map.
No special formatting except for hiding controls.
Should be better than nothing because you can always close the sidebar before printing if you don't like how it looks.
Shows directions as requested in #1273, but without direction arrows and wastes paper by having only one column.

### DIFF
--- a/app/assets/stylesheets/print.scss
+++ b/app/assets/stylesheets/print.scss
@@ -1,30 +1,58 @@
-header,
-#sidebar,
-#permalink,
-.leaflet-control {
-  display: none;
-}
-
 html {
   height: 100%;
 }
 
+body {
+  height: 100%;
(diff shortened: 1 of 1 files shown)


gravitystorm:
Thanks @AntonKhorev for your pull request!

I don't think that working on the printing here is particularly important to the project. I think this is more useful for non-mappers who are using the site as an end-user application, not for helping mappers achieve their tasks. So it's a pretty low priority and I don't want to attract more complaints from non-mappers pointing out the shortcomings (lack of route highlights, map positioning to cover start/end points, etc), since I don't think we should prioritise the work required to fix them.

Hopefully you are correct that it is "better than nothing"!

PR #4189, opened 2023-08-19
Change query limit settings format to nested mappings

AntonKhorev:
Instead of settings named like this `Settings.max_changeset_query_limit`
it will be easier to work with `Settings.changesets.max_query_limit`.

It matches the api capabilities response structure better.
It will be easier to create a generic [`result_limit` method](https://github.com/openstreetmap/openstreetmap-website/blob/406d90c5c8da5633324bb1d28d7a2103a61c95e4/app/controllers/api/changesets_controller.rb#L426).

Also added settings validation in `config/initializers/config.rb`.


tomhughes:
I'm not sure about this one I have to say - there are plenty of other settings that work as a group where we haven't done this.

AntonKhorev:
I don't want to write something like `Settings["max_#{type}_query_limit"]` in a possible superclass method.

AntonKhorev:
The idea is
- create a superclass for things searchable by the api; currently they are changesets and notes
- it's going to have `result_limit` method
- but `result_limit` reads type-specific settings
- then also add a method to read type-specific settings, rather than read them from the global `Settings` object; this is going to be either an abstract method to be defined in subclasses or derived from `controller_name` like in #3715

PR #4192, opened 2023-08-21
Make user lookup in changesets#index similar to lookup_user method

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/blob/7054cea48ee9728c155334b48777a5f25db89af8/app/controllers/application_controller.rb#L221
potential target for a refactor

Also add test for hidden edit history of inactive users.


tomhughes:
Why the focus on suspended users in the test rather than deleted users?

Suspended users are actually quite unusual - they basically only happen as a result of the automatic spam protections while deleted users will be the result of self-deletion or administrator action to close an account.

AntonKhorev:
The focus is on any status that is not active (or confirmed).

AntonKhorev:
Now tests both suspended and deleted users.

PR #4193, opened 2023-08-21
Lookup friend user before make/remove friend action

AntonKhorev:
custom lookup to use with `before_action`, similar to

https://github.com/openstreetmap/openstreetmap-website/blob/7054cea48ee9728c155334b48777a5f25db89af8/app/controllers/user_blocks_controller.rb#L114


gravitystorm:
Thanks! 

One day we should refactor that controller further, to use new/create/edit/destroy etc, and to remove the double-duty behaviour (where each action does two entirely different things, based on the method type).

PR #4197, opened 2023-08-21
Move user lookup/error methods from app controller to concerns

AntonKhorev:
For things that have username in their path like `https://www.openstreetmap.org/user/<username>/diary`. Later also for things with user id in their path.

The goal is to make `ApplicationController` smaller. <s>Code dealing with resolving user names/ids will go into the base class.</s> Adding anything to `ApplicationController` will cause Rubocop to complain.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -216,24 +216,6 @@ class ApplicationController < ActionController::Base
   end
 
   ##
-  # ensure that there is a "user" instance variable
-  def lookup_user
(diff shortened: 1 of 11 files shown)


AntonKhorev:
I can't even copy-paste from `ApplicationController` without Rubocop complaining.

tomhughes:
Is a base class the best solution here or would a concern that can be mixed in be better? See for example the current `user_concern` that is mixed in to the users and accounts controllers.

AntonKhorev:
Moved to user methods concern.

PR #4198, opened 2023-08-22
Add support for rate limiting signup requests

tomhughes:
This PR attempts to add support for rate limiting signups from a given IP address and/or email address.

Obviously we only allow one account per email address but this works on the basis of a canonical address where any `+` extension is removed, and dots are stripped from the local part for gmail.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -186,6 +186,9 @@ class UsersController < ApplicationController
         end
 
         if current_user.save
+          SIGNUP_IP_LIMITER&.update(request.remote_ip)
+          SIGNUP_EMAIL_LIMITER&.update(canonical_email(current_user.email))
+
           flash[:matomo_goal] = Settings.matomo["goals"]["signup"] if defined?(Settings.matomo)
 
           referer = welcome_path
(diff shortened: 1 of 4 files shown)


talllguy:
Thanks for the consideration Tom

Ignatych:
👍

geors:
👍

AntonKhorev:
> or the approach taken here where throttling is explicit in every controller.

how many controllers deal with user signups?

tomhughes:
I did look at a variety of gems but most wanted to use redis rather than memcache.

I didn't look at things which sit in the rack stack because I don't really see how they would work - they can rate limit loading of URLs but they can't really work on application state like signups for a give IP or email?

gravitystorm:
> how many controllers deal with user signups?

Sorry I wasn't clear - I expect that in the near future we will want to rate-limit a whole bunch of different things, and having one approach to rate limiting in multiple different contexts would be preferred.

AndrewHain:
Is there any good reason not to merge this as is now?

matkoniecz:
That will hit people organising workshops, where many people sign up at once from the same IP. But given scale of ongoing attack it makes sense to limit it before everyone goes insane.

And people can sign up from phones or in advance as workaround, I guess.

tomhughes:
Yes I'm aware that we can't be too strict on the IP limiting due to that kind of thing.

tomhughes:
I've merged this without waiting for a full review because we really need it live - happy to deal with any further review comments via followup PRs.

PR #4200, opened 2023-08-24
Reorganize blocks table from status/revoked to begin/end/status columns

AntonKhorev:
Currently [blocks tables](https://www.openstreetmap.org/user_blocks) contain too much repeated text like "Ends in almost 10 years and after the user has logged in." which could be made shorter. Partly such messages can be replaced by the column containing just dates.

The _Revoked by_ column is almost never used. Here I merge it with _Status_ column because revoked block is no longer active.

_Start date_ would be useful to see, especially if sorting by this date is added. This was earlier attempted in #1656.

blocktable


tomhughes:
Do you have a feel for how dependent on #4199 this is? I guess it tends to make the list wider, making that change more important?

AntonKhorev:
Table is going to be wider with longer usernames. When I made https://github.com/openstreetmap/openstreetmap-website/pull/4199 I was planning for one extra column , but I merged it into Status later.

AntonKhorev:
Before:
blocktable-before

After:
blocktable-after

Doesn't seem to be wider.

tomhughes:
I think this looks good so I'll go ahead and merge it, thanks!

Issue #4196, opened 2023-08-21
rate-limit changeset comments


matkoniecz:
### Problem

see also #2342 where rate limit for edits was proposed

https://resultmaps.neis-one.org/osm-discussion-comments?uid=20047765&commented - typically such rate of comments has no legitimate purpose

### Description

consider rate limiting changeset comments, especially for new accounts

maybe add option to delete changeset comments when blocking some account as vandal-only account?

### Screenshots

![screen06](https://github.com/openstreetmap/openstreetmap-website/assets/899988/7f9c0e39-b5a4-498c-9a2c-f57a2f0584dc)

all 2800+ comments were slurs and insults

tomhughes:
Some statistics to help guide possible limits...

There are roughly 20000 users (or ips) that have at some point posted over 10 comments an hour and roughly 400 that have exceeded 100 comments per hour.

There are about 1100 that have exceeded 100 per day and 23 that have exceeded 1000 per day

matkoniecz:
> That will of course go wrong the first time DWG do a mass revert and try to comment on all the reverted changesets.

I mentioned "especially for new accounts" and giving rate limit exception for accounts with block rights would be probably a good idea anyway.

The tricker cases are power users not in DWG.

matkoniecz:
Though limiting mass account creation would likely help with root cause here.

tomhughes:
Tuning things like this in one dimension is hard enough without adding in a second variable like how old the account is and people will just create submarine accounts and wait a while before using them.

Looking at the non-abusive cases with large rates they're not necessarily DWG but as you say "power users" who once posted a thousand in quick succession for some reason - either closing out a bunch of notes or doing some sort of import of hints from another database as notes.

matkoniecz:
Closing/opening notes is a separate thing.

Abusive changeset comments are especially annoying as you can easily harass any user you want (you cannot spam them by opening notes and it is trickier to do this by closing notes or commenting there, though I guess also possible)

tomhughes:
Well it's not a separate thing at the moment because due to the broken data model of notes there are comments associated with those events.

pnorman:
I've thought about this in the past but neglected to open an issue. I believe that *any* action that triggers an email should be rate limited similar to DMs. Catching someone from the DWG doing a revert that triggers thousands of emails because they left changeset comments on is a good thing, as if those emails go through they will cause problems on the email server or with spam prevention.

[...]

Issue #2342, opened 2019-08-06
Limit number of edits per user and day


woodpeck:
It regularly happens that people - usually, but not always new signups - upload hundreds of thousands of objects to OSM before someone notices and tells them to stop. Then we have to delete those hundreds of thousands of objects again. (Case in point from recent past, https://www.openstreetmap.org/user/maxiangying.) This is undesirable:
* it skews statistics ("wow, today was the day with the most edits since OSM started!!!" etc)
* it wastes processing time and bandwidth on the thousands of servers world wide that are configured to consume updates from OSM
* it wastes processing time and bandwidth on our own database server (where objects are first created, then deleted again, and a copy of the object is nonetheless held in history)
* it wastes the time of volunteers who have to remove the edits
* it embarrasses the person who uploaded the data (at least I hope it does)

While editors can, and should, inform their users about potential issues, I think it would also be worth contemplating to have some sort of rate limit on the API. It could be something that users can override but not accidentally - for example, you could be normally limited to X edits per day (exact numbers t.b.d.) and then you could click a button in your user preferences that says "I have read the data import and mechanical edit guidelines and I want to lift the limitation for one week" or so.

tomhughes:
The production API is in the cgimap repository, not here.

woodpeck:
Closed in favour of https://github.com/zerebubuth/openstreetmap-cgimap/issues/207

tomhughes:
From a practical point of view you need to define "edit" somehow - the obvious answer is changesets but that can be easily gamed so it will presumably need to be the number of objects added/changed/deleted.

Doing it per day is probably excessively expensive from an implementation standpoint because it requires resetting counters at some point, or keeping say hourly counts and expiring those over a day old. A token bucket type algorithm like the download limit probably makes more sense.

From a personal point of view I dread dealing with the resulting complaints...

simonpoole:
Actually I think the issue should remain here :-). 

At least the obvious, how stupid me would do it, implementation would involve automatically creating a block once the limit has been exceeded (because all the bits and pieces are already there and we don't want to create a slightly different implementation of the same thing).

I believe this would address @tomhughes  concerns to as the messaging can be clear what to do to be unblocked (OK I believe we still have the issue with iD borking when an account has been blocked but iD users are unlikely to run in to this anyway).

tomhughes:
[...]

PR #4202, opened 2023-08-25
Add rate limiting for changeset comments

tomhughes:
This inspired by but not identical to the algorithm proposed in #4196.

It ramps up to a specified maximum rate (default 60 per hour or 1 per minute) over the first two hundred changesets, subject to a configured minimum that defaults to 1 every 10 minutes with each unresolved report of the user halving the limit.

By my calculations this algorithm, with these defaults, will take just under 12 hours to reach the maximum limit if you constantly post as soon as you can - so that would be two hundred posts in 12 hours after which you can do 60 per hour and all assuming nobody has reported you.

Moderators have an arbitrary high limit of ten per second.

The down rating for reports might be a bit harsh but we can adjust it if that turns out the be the case.

### DIFF
--- a/app/controllers/api/changeset_comments_controller.rb
+++ b/app/controllers/api/changeset_comments_controller.rb
@@ -17,6 +17,7 @@ module Api
       # Check the arguments are sane
       raise OSM::APIBadUserInput, "No id was given" unless params[:id]
       raise OSM::APIBadUserInput, "No text was given" if params[:text].blank?
+      raise OSM::APIRateLimitExceeded if current_user.changeset_comments.where("created_at >= ?", Time.now.utc - 1.hour).count >= current_user.max_changeset_comments_per_hour
 
       # Extract the arguments
       id = params[:id].to_i
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -395,6 +395,19 @@ class User < ApplicationRecord
     max_friends.clamp(0, Settings.max_friends_per_hour)
   end
 
+  def max_changeset_comments_per_hour
+    if moderator?
+      36000
+    else
+      previous_comments = changeset_comments.limit(200).count
+      active_reports = issues.with_status(:open).sum(:reports_count)
+      max_comments = previous_comments / 200.0 * Settings.max_changeset_comments_per_hour
+      max_comments = max_comments.floor.clamp(Settings.min_changeset_comments_per_hour, Settings.max_changeset_comments_per_hour)
+      max_comments /= 2**active_reports
+      max_comments.floor.clamp(1, Settings.max_changeset_comments_per_hour)
+    end
+  end
+
   private
 
   def encrypt_password
--- a/test/controllers/api/changeset_comments_controller_test.rb
+++ b/test/controllers/api/changeset_comments_controller_test.rb
@@ -133,6 +133,27 @@ module Api
     end
 
     ##
+    # create comment rate limit
+    def test_create_comment_rate_limit
+      changeset = create(:changeset, :closed)
+      user = create(:user)
+
+      auth_header = basic_authorization_header user.email, "test"
+
+      assert_difference "ChangesetComment.count", Settings.min_changeset_comments_per_hour do
+        1.upto(Settings.min_changeset_comments_per_hour) do |count|
+          post changeset_comment_path(:id => changeset, :text => "Comment #{count}"), :headers => auth_header
+          assert_response :success
+        end
+      end
+
+      assert_no_difference "ChangesetComment.count" do
+        post changeset_comment_path(:id => changeset, :text => "One comment too many"), :headers => auth_header
+        assert_response :too_many_requests
+      end
+    end
+
+    ##
     # test hide comment fail
     def test_destroy_comment_fail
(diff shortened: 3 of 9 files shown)


app/controllers/api/changeset_comments_controller.rb
```diff
@@ -17,6 +17,7 @@ def create
       # Check the arguments are sane
       raise OSM::APIBadUserInput, "No id was given" unless params[:id]
       raise OSM::APIBadUserInput, "No text was given" if params[:text].blank?
+      raise OSM::APIRateLimitExceeded if current_user.changeset_comments.where("created_at >= ?", Time.now.utc - 1.hour).count >= current_user.max_changeset_comments_per_hour
```
gravitystorm, line 20: might be better slightly abstracted - e.g. `raise OSM::APIRateLimitExceeded if rate_limit_exceeded` and the details put elsewhere, just to make the method easier to read. 

I know we have the method complexity metrics effectively disabled in rubocop, otherwise I'm sure it would have flagged this.

(I'd also like to share limits between api vs website controllers in the long run, and so making reusable methods would be handy).

app/models/user.rb
```diff
@@ -395,6 +395,19 @@ def max_friends_per_hour
     max_friends.clamp(0, Settings.max_friends_per_hour)
   end
 
+  def max_changeset_comments_per_hour
+    if moderator?
+      36000
```
gravitystorm, line 400: Maybe better as "60 * 60 * 10" since I thought that was one per second when I read it at first.

app/models/user.rb
```diff
@@ -395,6 +395,19 @@ def max_friends_per_hour
...
+    else
+      previous_comments = changeset_comments.limit(200).count
+      active_reports = issues.with_status(:open).sum(:reports_count)
+      max_comments = previous_comments / 200.0 * Settings.max_changeset_comments_per_hour
+      max_comments = max_comments.floor.clamp(Settings.min_changeset_comments_per_hour, Settings.max_changeset_comments_per_hour)
+      max_comments /= 2**active_reports
+      max_comments.floor.clamp(1, Settings.max_changeset_comments_per_hour)
+    end
```
gravitystorm, line 408: That's an interesting algorithm! I do like relying on previous activity, rather than account age, so I think we could rework the existing limits along these lines too. And using reports as the main downward modifier is smart.

I suspect there might be non-moderator power users caught out by this, since a disagreement about reverts could lead to one person making ~10 reports and therefore severely rate-limiting the power user. Perhaps defending against that would be too much complexity for now.

Is there a naming issue here? The minimum number of comments per hour is actually 1, not `min_changeset_comments_per_hour`. I think an admin seeing the min and max settings might not realise that the min is more of a "default" or starter value, that can go both up and also go down lower.

I also think we need to move these rate limit calculations out of the user model, but I'm not yet sure where they should go. The model is already huge, and it almost feels like these are controller-level decisions rather than some property of the model in isolation.

test/controllers/api/changeset_comments_controller_test.rb
```diff
@@ -132,6 +132,27 @@ def test_create_comment_fail
...
+          assert_response :success
+        end
+      end
+
+      assert_no_difference "ChangesetComment.count" do
+        post changeset_comment_path(:id => changeset, :text => "One comment too many"), :headers => auth_header
+        assert_response :too_many_requests
+      end
+    end
+
```
gravitystorm, line 155: The tests only scratch the edge of the algorithm! Would be good to have tests covering

* accounts with lots of existing comments
* accounts with reports
* moderator accounts

gravitystorm (approved):
I'm happy to merge this as-is, these comments are just suggestions for improvements.

app/models/user.rb
tomhughes: The minimum value is the minimum that will happen in normal operation - it will only go below that when the user is reported. Open to suggestions for alternate names though - the basic idea was that the limit ramps up from min to max but can be restricted by reports.

Obviously there is a potential issue with non-moderator power users as you say, though to paraphrase @woodpeck do we really want those people commenting on thousands of changesets and generating email storms?

gravitystorm: > Obviously there is a potential issue with non-moderator power users as you say, though to paraphrase @woodpeck do we really want those people commenting on thousands of changesets and generating email storms?

:smile: Although if good stuff can be done outside of DWG, more power to them! But in this case, a power user can be prevented from making more than `1` changeset comment per hour, by only a few (bogus) reports.

For naming, my suggestion would be:
* `min_changeset_comments_per_hour` = 1
* `default_changeset_comments_per_hour` = 6
* `max_changeset_comments_per_hour` = 200
* `moderator_changeset_comments_per_hour` = 36_000

matkoniecz: > But in this case, a power user can be prevented from making more than 1 changeset comment per hour, by only a few (bogus) reports.

Maybe like proposed in https://github.com/openstreetmap/openstreetmap-website/issues/2342#issuecomment-1694650004 it would be possible to count how many users made reports against account, not simply how many reports were made? So simply making 100 bogus reports will not have so serious effects.

PR #4203, opened 2023-08-25
Use `where()` instead of `find()` to prevent RecordNotFound

ENT8R:
When using the Users API to find a set of specific users by their ids and including an id of a user that deleted their account in this set, the request will currently fail with a 404
For example requesting the details of the users with the ids `18597840` **and** `185978900` [fails with a 404](https://www.openstreetmap.org/api/0.6/users?users=18597840,185978900) whereas limiting the query to user `18597840` returns the [correct result](https://www.openstreetmap.org/api/0.6/users?users=18597840)
This happens because user `185978900` deleted their account and `find()` raises an `ActiveRecord::RecordNotFound` "If one or more records cannot be found for the requested ids". <sup><a href="https://api.rubyonrails.org/v7.0.7.2/classes/ActiveRecord/FinderMethods.html#method-i-find">1</a></sup>

Using `where()` to find the users by their ids does not result in an error being raised if (at least) one of the records could not be found.
I also added a fallback for rendering the XML in case no user was found at all (because `@users` is nil in that case which can not be converted to a String)

### DIFF
--- a/app/views/api/users/index.xml.builder
+++ b/app/views/api/users/index.xml.builder
@@ -1,4 +1,4 @@
 xml.instruct! :xml, :version => "1.0"
 xml.osm(OSM::API.new.xml_root_attributes) do |osm|
-  osm << render(@users)
+  osm << (render(@users) || "")
 end
--- a/test/controllers/api/users_controller_test.rb
+++ b/test/controllers/api/users_controller_test.rb
@@ -419,13 +419,15 @@ module Api
       check_json_details(js["users"][1], user3, false, false)
 
       get api_users_path, :params => { :users => create(:user, :suspended).id }
-      assert_response :not_found
+      assert_response :success
 
       get api_users_path, :params => { :users => create(:user, :deleted).id }
-      assert_response :not_found
+      assert_response :success
 
       get api_users_path, :params => { :users => 0 }
-      assert_response :not_found
+      assert_response :success
+      assert_equal "application/xml", response.media_type
+      assert_select "user", :count => 0
     end
 
     def test_index_oauth1
@@ -506,13 +508,15 @@ module Api
       check_json_details(js["users"][1], user3, false, false)
 
       signed_get api_users_path, :params => { :users => create(:user, :suspended).id }, :oauth => { :token => good_token }
-      assert_response :not_found
+      assert_response :success
 
       signed_get api_users_path, :params => { :users => create(:user, :deleted).id }, :oauth => { :token => good_token }
-      assert_response :not_found
+      assert_response :success
 
       signed_get api_users_path, :params => { :users => 0 }, :oauth => { :token => good_token }
-      assert_response :not_found
+      assert_response :success
+      assert_equal "application/xml", response.media_type
+      assert_select "user", :count => 0
     end
 
     def test_index_oauth2
@@ -593,13 +597,15 @@ module Api
       check_json_details(js["users"][1], user3, false, false)
 
       get api_users_path, :params => { :users => create(:user, :suspended).id }, :headers => bearer_authorization_header(good_token.token)
-      assert_response :not_found
+      assert_response :success
 
       get api_users_path, :params => { :users => create(:user, :deleted).id }, :headers => bearer_authorization_header(good_token.token)
-      assert_response :not_found
+      assert_response :success
 
       get api_users_path, :params => { :users => 0 }, :headers => bearer_authorization_header(good_token.token)
-      assert_response :not_found
+      assert_response :success
+      assert_equal "application/xml", response.media_type
+      assert_select "user", :count => 0
     end
 
     def test_gpx_files
--- a/test/controllers/api/users_controller_test.rb
+++ b/test/controllers/api/users_controller_test.rb
@@ -420,9 +420,13 @@ module Api
 
       get api_users_path, :params => { :users => create(:user, :suspended).id }
       assert_response :success
+      assert_equal "application/xml", response.media_type
+      assert_select "user", :count => 0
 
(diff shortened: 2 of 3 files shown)


ENT8R:
As I see there are multiple test cases covering this case and they are all expecting the method to raise an RecordNotFound in case one of the users was deleted. Is there a reason behind this? Because my use case is that I have an (possibly) outdated set of user ids and I obviously do not know whether certain users of this set decided to delete their account, but I would still be happy to receive a response from the server regarding the other users. If one of the users I initially requested does not show up in the response, I know that the account was deleted/suspended/...

tomhughes:
I don't see how a controller test can be testing for an exception - all a controller test can do is to observe the HTTP response code.

But if the tests are wrong then fix them.

tomhughes:
As I thought the test is not looking for an exception, it's looking for a 404 response - the fact that might be generated internally by an exception currently is merely an implementation detail.

The question is, when a method looks up a set of resources and none of them exist should it return 404 or should it return 200 with an empty response set? You can certainly make arguments either way though my inclination would be to go with an empty response set.

Either can be implemented - for a 404 just test for an empty set and return a not found response, and for 200 just leave it as it is and fix the test to expect that.

AntonKhorev:
I'd move to making multifetches work like filters. In case they don't match anything they should return empty 200 result.
That's what I did in #3707.

AntonKhorev:
By empty result I mean a valid xml document with an empty root element, not an empty response body.

ENT8R:
I decided to keep it as is (return 200 with empty response/root element) and modified the tests accordingly. I also had to sort the results by their ids because otherwise tests like the following which expect the result to be in a certain order were failing as `where()` does not sort the results by default:
https://github.com/openstreetmap/openstreetmap-website/blob/6759130cb3e266d63743fdf02978550f45eb7c4b/test/controllers/api/users_controller_test.rb#L411-L420

test/controllers/api/users_controller_test.rb
```diff
@@ -419,13 +419,15 @@ def test_index
       check_json_details(js["users"][1], user3, false, false)
 
       get api_users_path, :params => { :users => create(:user, :suspended).id }
-      assert_response :not_found
+      assert_response :success
```
tomhughes, line 422: You should add the same assertion here that you use later to check that no users are returned.

tomhughes, line 422: You should add the same assertion here that you use later to check that no users are returned.

test/controllers/api/users_controller_test.rb
```diff
@@ -419,13 +419,15 @@ def test_index
...
 
       get api_users_path, :params => { :users => create(:user, :deleted).id }
-      assert_response :not_found
+      assert_response :success
```
tomhughes, line 425: Same here - this should check no users are returned.

app/views/api/users/index.xml.builder
```diff
@@ -1,4 +1,4 @@
 xml.instruct! :xml, :version => "1.0"
 xml.osm(OSM::API.new.xml_root_attributes) do |osm|
-  osm << render(@users)
+  osm << (render(@users) || "")
```
tomhughes, line 3: Is this actually necessary? Won't an empty set render sensibly? If it is needed that what about the JSON version?

test/controllers/api/users_controller_test.rb
```diff
@@ -506,13 +508,15 @@ def test_index_oauth1
       check_json_details(js["users"][1], user3, false, false)
 
       signed_get api_users_path, :params => { :users => create(:user, :suspended).id }, :oauth => { :token => good_token }
-      assert_response :not_found
+      assert_response :success
```
tomhughes, line 511: This should check the set is empty.

test/controllers/api/users_controller_test.rb
```diff
@@ -506,13 +508,15 @@ def test_index_oauth1
...
 
       signed_get api_users_path, :params => { :users => create(:user, :deleted).id }, :oauth => { :token => good_token }
-      assert_response :not_found
+      assert_response :success
```
tomhughes, line 514: This should check the set is empty.

test/controllers/api/users_controller_test.rb
```diff
@@ -593,13 +597,15 @@ def test_index_oauth2
       check_json_details(js["users"][1], user3, false, false)
 
       get api_users_path, :params => { :users => create(:user, :suspended).id }, :headers => bearer_authorization_header(good_token.token)
-      assert_response :not_found
+      assert_response :success
```
tomhughes, line 600: This should check the set is empty.

test/controllers/api/users_controller_test.rb
```diff
@@ -593,13 +597,15 @@ def test_index_oauth2
...
 
       get api_users_path, :params => { :users => create(:user, :deleted).id }, :headers => bearer_authorization_header(good_token.token)
-      assert_response :not_found
+      assert_response :success
```
tomhughes, line 603: This should check the set is empty.

tomhughes (changes requested):

app/views/api/users/index.xml.builder
ENT8R: I think so, see also https://github.com/openstreetmap/openstreetmap-website/pull/2223#discussion_r300065590
If the set is empty, rendering fails with `ActionView::Template::Error: no implicit conversion of nil into String`, because `render()` returns nil in that case and trying to concatenate nil with an existing string will fail (but concatenating an empty string with another string works).
Regarding the JSON version, this is not an issue because no strings are concatenated there.

PR #4207, opened 2023-08-28
Restore original users index method; add search method for multi fetch

AntonKhorev:
Do this https://github.com/openstreetmap/openstreetmap-website/commit/b9c85c269726faad3c2613b37f01683c8b59f5c6#commitcomment-125595007 to avoid changing the existing api behavior.

Add a users search endpoint '/api/0.6/users/search.json?users=1,2,3' that doesn't responds with an error if some users don't exist.

Currently the `users` parameter is required, this is to be relaxed later.
Currently the results order is unspecified, this is to be specified later, probably by date.

### DIFF
--- a/app/abilities/api_ability.rb
+++ b/app/abilities/api_ability.rb
@@ -14,7 +14,7 @@ class ApiAbility
       can [:show, :download, :query], Changeset
       can [:index, :create, :comment, :feed, :show, :search], Note
       can :index, Tracepoint
-      can [:index, :show], User
+      can [:index, :search, :show], User
       can [:index, :show], Node
       can [:index, :show, :full, :ways_for_node], Way
       can [:index, :show, :full, :relations_for_node, :relations_for_way, :relations_for_relation], Relation
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -19,12 +19,28 @@ module Api
 
       raise OSM::APIBadUserInput, "No users were given to search for" if ids.empty?
 
-      @users = User.visible.where(:id => ids).in_order_of(:id, ids)
+      @users = User.visible.find(ids)
 
       # Render the result
       respond_to do |format|
         format.xml
         format.json
+      end
+    end
+
+    def search
+      raise OSM::APIBadUserInput, "The parameter users is required, and must be of the form users=id[,id[,id...]]" unless params["users"]
+
+      ids = params["users"].split(",").collect(&:to_i)
+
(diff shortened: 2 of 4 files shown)


tomhughes:
No, just no. We don't want duplicate methods and inconsistency between different endpoints just to keep one end user happy.

AntonKhorev:
So you don't want search method for users?

If you decided to keep users query ordered by ids, you can't get it to work at all when there are no ids, for example, if it's a date range query.

Or do you want to extend the index method for that and change the results order again?

AntonKhorev:
Didn't you write this:
https://github.com/openstreetmap/openstreetmap-website/commit/b4106383d99ccbf152d79b0f2c9deca95df9fb61#diff-7c3ebfc581188b2253bfe03a071fb6b02c99b3287533b623d69414e7d68d20f2R1186

the implementation of `index` together with tests? Now you're saying [it was an accident](https://github.com/openstreetmap/openstreetmap-website/commit/b9c85c269726faad3c2613b37f01683c8b59f5c6#commitcomment-125593592). It doesn't test for multiple users with one of them missing though, maybe the behavior in this case is an accident?

tomhughes:
I thought this was just a version of index that didn't 404 so that you could revert index to the silly 404 response?

AntonKhorev:
You didn't notice that I added `search` without the silly 404 response?

tomhughes:
Yes I know - that was my point that by adding search as a parallel to index (but without the 404) you can then revert index to the 404 behaviour, but I see no reason to want the old behaviour.

I hadn't looked into the history of it so maybe accident isn't the right word but it's clearly not a useful design so I have no idea why I did it like that.

My guess is that I based it on the single user fetch and didn't really think it through - before last week I wasn't even really aware that `find` could take multiple IDs as I would normally use `where` where selecting more than one object so maybe I literally copied the code and it worked and I didn't really notice the silliness.

AntonKhorev:
The reason to revert is what if there's more than one user who is going to complain about this change.

The method itself is 8 lines of code, not a big overhead to have in parallel.

Maybe you copied the code from the elements multifetch, in this case you got the behavior consistent with it. Another option is to make it similar to the changesets multifetch without silly 404, that's what I'm adding here. Both behaviors are consistent with some other api call.

AntonKhorev:
Element multifetch was done with `.find(ids)` since 2006: https://github.com/openstreetmap/openstreetmap-website/commit/f4743c08ac8a80fb7e20f5540fa405ce5efb7ff6

PR #4211, opened 2023-08-30
Address review comments for changeset comment limiting

tomhughes:
This aims to address the review comments from #4202 which was already merged.

It mostly makes the request changes - one different is that I went with `initial` rather than `default` as the new name for the starting limit as default seemed confusing when it's only the default for a new user.


gravitystorm:
Great, thanks! I agree `initial` is a better term too.

PR #4214, opened 2023-09-01
Add notes feed description for unspecified area

AntonKhorev:
You can get a feed of notes unrestricted by any bounding box: https://api.openstreetmap.org/api/0.6/notes/feed

It's not entirely clear if this is planned. The query runs rather slow, the [wiki](https://wiki.openstreetmap.org/wiki/API_v0.6#RSS_Feed:_GET_/api/0.6/notes/feed) shows `bbox` as required, but the requirement was dropped more than ten years ago.

If it is supposed to work, the description element in the feed is going to look strange. It refers to an area and has slots for bounding box lat/lon, but since the area is undefined, the slots are empty:

    <description>A list of notes, reported, commented on or closed in your area [(|) -- (|)]</description>

This PR adds another description string for feeds of all notes without area restrictions.

### DIFF
--- a/app/views/api/notes/feed.rss.builder
+++ b/app/views/api/notes/feed.rss.builder
@@ -6,7 +6,11 @@ xml.rss("version" => "2.0",
         "xmlns:georss" => "http://www.georss.org/georss") do
   xml.channel do
     xml.title t("api.notes.rss.title")
-    xml.description t("api.notes.rss.description_area", :min_lat => @min_lat, :min_lon => @min_lon, :max_lat => @max_lat, :max_lon => @max_lon)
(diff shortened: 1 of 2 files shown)


tomhughes:
I can't see any way it wasn't deliberate and I don't see any real reason why it should be especially slow - the core query it does to find the recent comments looks find and is using indexes and everything else is the same for all feeds.

I've added some more preloads in https://github.com/openstreetmap/openstreetmap-website/commit/704781ad0b108e80e4c6eef030c8c1c785e815e4 to minimise the database queries but it looks like most of the time is actually from rendering all the partials.

AntonKhorev:
I think it was slow because of repeated user db queries.

Issue #482, opened 2013-09-16
Permalinks break when user changes their display name


gravitystorm:
This problem is well know, and not new, but still needs to be solved.

All of the user-related urls, like diaries and even user profiles, use the "display name" in the url, which can be changed by the user, e.g.

`https://www.openstreetmap.org/user/Andy Allan`

This makes it impossible to create stable links to users or their diaries etc, and users can hamper communication and anti-vandalism activities by repeatedly changing their display name.

The urls need to be stable. One way to achieve this would be to use the user-id instead of the display name

`https://www.openstreetmap.org/user/4115`

The user id is already exposed in the API and used by third-party services, precisely because the display name is unstable. We could do the same on osm.org.

gravitystorm:
I've been thinking about this more (over the last 4 years, ahem) and so I'd like to propose the following:

* Usernames will no longer be allowed to start with numbers followed by a hyphen
* Any username (are there any?) that is currently in that format will need to be changed, e.g. contact users to let them choose, and then changing remaining hyphens to underscores.
* User links would be in the format `/user/4115-Andy Allan/` i.e. numeric ID followed by a hyphen followed by username.
* Old-style user links would be supported with redirects, e.g. `/user/Andy Allan` would 301 redirect to `/user/4115-Andy Allan/`.
* Correct usernames in links will be enforced e.g. `/user/4115-hahaha` would 301 redirect to `/user/4115-Andy Allan/`.
* Everything above also applies to subpages e.g. `/user/Andy Allan/edits` would 301 redirect etc
* The numeric-only usernames would still be allowed and still work, e.g. `/user/12345` would redirect to `/user/1111-12345`.

Any thoughts, or anything else that I should consider?

tomhughes:
Sounds reasonable - for the record there are 276 users affected.

tomhughes:
Only 74 of which have any edits.

simonpoole:
I see a bit of a problem with the scheme that it potentially permanently leaks the uid - displayname relationship to the outside. One could argue that it is unlikely that the user page in its current form will remain public in any case, but I would suggest to avoid everything which makes things even more complicated till we've decided on the scope of the GDPR rejig.

tomhughes:
I was literally about 30s away from having a go at implementing @gravitystorm's scheme, but I guess that's something else we can sacrifice on the altar of GDPR for now.

simonpoole:
[...]

PR #4215, opened 2023-09-01
Add user id redirect paths

AntonKhorev:
Add paths like `/uid/123` that redirect to user pages for users that are visible.

The idea is to extend them later for deleted users to make some of their pages available. For example, it is sometimes necessary to see the deleted user's edit history. The history is available through the api, but the website won't show anything because there's no valid url for the user.


tomhughes:
There's a whole ticket somewhere discussing solutions to this well known problem but I'd really rather not create a new top level namespace if we can avoid it.

mmd-osm:
That's issue #482

PR #4218, opened 2023-09-02
Disallow username changes to user_n if n isn't their id

AntonKhorev:
In case you don't want #4215, there's already a defacto url scheme with user ids: `/user/user_id`. That's what deleted accounts get reset to. But it has a couple of problems. "user_12345" is a valid username for anyone, not just the user with id=12345. Depending on how many usernames with this pattern already exist, we may want to stop allowing new ones.

I know one active user with this kind of name, but it matches their id.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -61,7 +61,7 @@ Metrics/BlockNesting:
 # Offense count: 26
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 299
+  Max: 305
 
 # Offense count: 59
 # Configuration parameters: AllowedMethods, AllowedPatterns.
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -99,7 +99,7 @@ class User < ApplicationRecord
   validates :display_name, :if => proc { |u| u.display_name_changed? },
                            :characters => { :url_safe => true },
                            :whitespace => { :leading => false, :trailing => false }
-  validate :display_name_cannot_be_user_id_with_other_id
+  validate :display_name_cannot_be_user_id_with_other_id, :if => proc { |u| u.display_name_changed? }
   validates :email, :presence => true, :confirmation => true, :characters => true
   validates :email, :if => proc { |u| u.email_changed? },
                     :uniqueness => { :case_sensitive => false }
@@ -126,7 +126,7 @@ class User < ApplicationRecord
   after_save :spam_check
 
   def display_name_cannot_be_user_id_with_other_id
-    display_name_changed? && display_name&.match(/^user_(\d+)$/i) do |m|
+    display_name&.match(/^user_(\d+)$/i) do |m|
       errors.add :display_name, I18n.t("activerecord.errors.messages.display_name_is_user_n") unless m[1].to_i == id
     end
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -99,6 +99,7 @@ class User < ApplicationRecord
   validates :display_name, :if => proc { |u| u.display_name_changed? },
                            :characters => { :url_safe => true },
                            :whitespace => { :leading => false, :trailing => false }
+  validate :display_name_cannot_be_user_id_with_other_id
   validates :email, :presence => true, :confirmation => true, :characters => true
   validates :email, :if => proc { |u| u.email_changed? },
                     :uniqueness => { :case_sensitive => false }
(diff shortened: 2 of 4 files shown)


pnorman:
👍 to the idea of requiring user_N to match their id. I can't review the code quality, but you might want to squish the two commits together.

gravitystorm:
@AntonKhorev thank you for the PR, that's a good catch. Without it, there's a risk of soft deletes failing.

I changed the code slightly - merging your two commits as @pnorman suggested, and also refactoring slightly to move the change detection to the `validate` statement, to align it with the other validations. I wanted to check that this actually worked (the documentation wasn't 100% clear to me) and by the time I'd checked I figured it was easier for me to commit the code than to ask you to just do the same work!

AntonKhorev:
> Without it, there's a risk of soft deletes failing.

Wasn't [this report](https://github.com/openstreetmap/openstreetmap-website/security/advisories/GHSA-2779-44wq-qcm8) seen by anyone again, like [previously](https://github.com/openstreetmap/openstreetmap-website/issues/4362)?
Looks like they are only visible to [osm org members](https://github.com/orgs/openstreetmap/people?query=role%3Aowner) and @gravitystorm is not one of them.

gravitystorm:
> Wasn't [this report](https://github.com/openstreetmap/openstreetmap-website/security/advisories/GHSA-2779-44wq-qcm8) seen by anyone again, like [previously](https://github.com/openstreetmap/openstreetmap-website/issues/4362)?

I think I read this at the time, but there's so many notifications from this repo that I can't say I remember the text of all of them. But when I was reviewing PRs I remembered that this one was an important one to deal with, which is why it I worked on it yesterday before the 100+ other open PRs.

I thought it was worth explaining clearly in my comment what the impact is, of course you understand it but I thought it would be worth clarifying for everyone else.

> Looks like they are only visible to [osm org members](https://github.com/orgs/openstreetmap/people?query=role%3Aowner) and @gravitystorm is not one of them.

For historical reasons I'm an owner for the openstreetmap organisation, so I can see the reports no problem.

However, they aren't super obvious though in the Github UI - there's no little numbers beside the `Security` tab, nor even the `Advisories` menu option. It's only when you click on that menu that you can see the list - and be reminded about any open ones!

I'll try to formally deal with the security advisory next week - I've accepted it, but now it's complaining about not having affected versions specified, and I need to figure out what is the most appropriate thing (git commit id? Date?) that I should put in there.

Screenshot from 2024-01-18 14-20-38

tomhughes:
> Wasn't [this report](https://github.com/openstreetmap/openstreetmap-website/security/advisories/GHSA-2779-44wq-qcm8) seen by anyone again, like [previously](https://github.com/openstreetmap/openstreetmap-website/issues/4362)?

I don't appear to have received any notification of it that I can see in my email archives.

tomhughes:
> I'll try to formally deal with the security advisory next week - I've accepted it, but now it's complaining about not having affected versions specified, and I need to figure out what is the most appropriate thing (git commit id? Date?) that I should put in there.

I normally ignore all that nonsense and just close it - if you actually fill that I in I fear that will allocate a CVE which is the last thing we want.

AntonKhorev:
> merging your two commits as @pnorman suggested

Previously pull requests were merged with a separate commit changing rubocop config: https://github.com/openstreetmap/openstreetmap-website/pull/4284 https://github.com/openstreetmap/openstreetmap-website/pull/4284/commits/393914d35b970e98c825dc650f679fcb12b6655e.

But why are those counting checks enabled at all, given that the solution to failing them is to increase the limit and disregard whatever Rubocop suggests https://github.com/openstreetmap/openstreetmap-website/pull/4387#discussion_r1416152114?

tomhughes:
The clue is in the `_todo` part of the name - they are checks that we haven't managed to enable yet.

I guess maybe we should just disable them fully - probably the idea of the auto-generated todo putting a numeric limit in instead was to stop people making them worse but obviously we don't do that.

I'm generally skeptical of most such metrics to be honest.

AntonKhorev:
> and I need to figure out what is the most appropriate thing (git commit id? Date?)

I think I tried entering commit numbers previously and it didn't work, it checks the format.

gravitystorm:
> I guess maybe we should just disable them fully - probably the idea of the auto-generated todo putting a numeric limit in instead was to stop people making them worse but obviously we don't do that.

Well maybe in future I should refuse any PRs that increase the number! But I think other people would be unhappy with that approach.

I don't want to turn it off, since it's making a point that needs to be made. `user.rb` is far too big to understand at a glance, and imho should be refactored, but like you say that's a longer-term TODO and I can't block a security fix by waiting for the refactoring.

I suspect what we should do is make it a per-file exemption rather than a global one, so that when we refactor `user.rb` we don't find that there are many other files that have grown in the meantime. 

> Previously pull requests were merged with a separate commit changing rubocop config: #4284 [393914d](https://github.com/openstreetmap/openstreetmap-website/commit/393914d35b970e98c825dc650f679fcb12b6655e).

:shrug: So I'm inconsistent, perhaps? But in general I like to make sure that tests pass for all commits, so that `git bisect` works smoothly. And it was easy for me to fix while I was making commits on this PR anyway. But perfect is the enemy of good so it's not a super strict rule or anything.

Issue #1311, opened 2016-10-06
Deleted user: history lost


woodpeck:
Deleted users used to be changed to something like "user_12345" (number chosen for illustration only), and you were still able to e.g. view the edit history of such a user. This is an important feature for the community to understand what someone has done (imagine e.g. a vandal making 20 edits then asking for his account to be deleted - you still want to be able to identify which edits they made).

However, for reasons I failed to identify in the code, deleted users now show up with the user name "deleted" (e.g. http://www.openstreetmap.org/changeset/37413625), and no hyperlink.

This is definitely a recent change but I cannot pinpoint where and when exactly it was introduced.

Requesting the raw XML for such an object still reveals the "user_12345" name, but then requesting that user's history through a link like http://www.openstreetmap.org/user/user_12345/history yields an error message as well. It would be desirable to be able to access the user's history in this way even though the account was deleted.

This might not be such a recent change but in March this year it was definitely still possible to access deleted users' histories in this way.

tomhughes:
Deleted users still are changed to that, they just aren't visible. As far as I know deleted users have never been visible, or at least not for a very long time.

tomhughes:
Oh the link may have been hidden in some specific contexts sure, but that was mostly because the link didn't work anyway, and just the name of it effectively revealed the identity because the uid to name mapping is known.

tomhughes:
So 80c46f54d4930deacb8664ff44703b8c3c5d088e introduced the current message but even before that they were shown as anonymous, along with very old users that have never agreed to make their edits public.

tomhughes:
So in fact it was the previous edit from the same merge (341f6cd9feed774f0abdb337f36272d5ecc92549) that made them anonymous. See #1154.

woodpeck:
Hmm, appears I was wrong with "in March this year it was definitely still possible to access deleted users' histories". But I'm adamant that until very recently, I was able to see whether two edits had been made by the same deleted user or by different deleted users...

tomhughes:
I don't see how - as far as I can the changeset list code has refused to give you a list for a non-active user for at least the last five years.

woodpeck:
Yes I concede the "give you a list" bit but I *was* told that one edit was by "user_1234" and the other by "user_2345", whereas now both edits are by "deleted". Or am I imagining that as well?

pnorman:
That was my recollection, but I haven't tested a great deal so can't be certain.

tomhughes:
[...]

PR #4224, opened 2023-09-03
Show edit history of deleted users

AntonKhorev:
Links to previous/next changesets from `/changeset/<id>`. Also links to `/user/<name>/history`, which becomes available for everyone. `<name>` is whatever the deletion procedure wrote to user's display name. This name is only used in urls, everywhere else the user is shown as "deleted".

Implements #1311


tomhughes:
I think whether we can/should do this depends on us reaching a conclusion on how we can/should handle deleted accounts in general which in turn means deciding what, from a data protection point of view, we are allowed to do.

AntonKhorev:
All that's made available here on the web interface is already available on the api.

tomhughes:
Yes why we need to figure out the correct policy and apply it consistently everywhere - that might this PR or it might mean making the API match the web interface.

AntonKhorev:
All changesets are available in dumps too.

AntonKhorev:
Your current policy:
https://github.com/openstreetmap/openstreetmap-website/commit/6c1d73a509a1ac2216a704f3dc5534d573e5d7ae#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7R245

All you can change is to make the edit history available only for registered users. But then that would likely apply to all history pages, not just ones of deleted users.

PR #4226, opened 2023-09-04
Add openid connect support using doorkeeper-openid_connect gem

milan-cvetkovic:
...as discussed in [Issue 507](https://github.com/openstreetmap/operations/issues/507) and described by @mmd-osm.

To activate, set the value of `doorkeeper_signing_key` to pem of the RSA private key.

Allows using openstreetmap as an identity provider.

Adds `openid` scope to OAuth2 authorizations, required to login to OSM.

Currently, the only claims returned are `sub` and `preferred_username`.


db/migrate/20230830115220_validate_create_doorkeeper_openid_connect_tables.rb
```diff
@@ -0,0 +1,5 @@
+class ValidateCreateDoorkeeperOpenidConnectTables < ActiveRecord::Migration[7.0]
```
tomhughes, line 1: Why is this a separate migration? Can we not just include it in the previous migration that adds the tables?

db/migrate/20230830115221_add_openid_scope.rb
```diff
@@ -0,0 +1,11 @@
+class AddOpenidScope < ActiveRecord::Migration[7.0]
+  def self.up
+    add_column :oauth_tokens, :allow_openid, :boolean, :null => false, :default => false
+    add_column :client_applications, :allow_openid, :boolean, :null => false, :default => false
```
tomhughes, line 4: Why are we adding openid connect support to OAuth 1? That's not even a thing is it?

config/locales/doorkeeper_openid_connect.en.yml
```diff
@@ -0,0 +1,23 @@
+en:
```
tomhughes, line 1: You can just add extra local files like this - if the gem doesn't make these available automatically then you'll have to add them to the main `en.yml` file.

Gemfile
```diff
@@ -78,6 +78,7 @@ gem "omniauth-rails_csrf_protection", "~> 1.0"
 
 # Doorkeeper for OAuth2
 gem "doorkeeper"
+gem "doorkeeper-openid_connect"
```
tomhughes, line 81: Pretty sure rubocop is going to reject this as not correctly sorted - it will need to be moved after `doorkeeper-i18n`.

tomhughes (commented):
This is just an initial review of obvious things as I haven't looked at any of the details of the gem and how it works.

milan-cvetkovic:
> This is just an initial review of obvious things as I haven't looked at any of the details of the gem and how it works.

Thanks for the quick review! I will look into addressing your comments.

tomhughes:
I reran the actions as they failed due to some github issue and the rubocop run has now picked up a bunch more things.

config/locales/doorkeeper_openid_connect.en.yml
milan-cvetkovic: I was confused with this file - it was generated by gem's installation.

Will move them to en.yml,

tomhughes:
I think from https://github.com/doorkeeper-gem/doorkeeper-openid_connect#internationalization-i18n that they will actually be found in the gem - the installer just added a copy of them so they could be customised.

That said it may be worth adding them so they can be translated - the main doorkeeper ones don't need to be as the `doorkeeper-i18n` gem does it.

db/migrate/20230830115220_validate_create_doorkeeper_openid_connect_tables.rb
milan-cvetkovic: Hm, db:migrate was failing due to possible deadlock when creating foreign keys and validating them in one shot, and suggested splitting to two migration scripts, which I did. I guess I could merge them into one.

Here is the error message from db:migrate

> === Dangerous operation detected #strong_migrations ===
>
> Adding a foreign key blocks writes on both tables. Instead,
> add the foreign key without validating existing rows,
> then validate them in a separate migration.

tomhughes: If there's a good reason then fine.

db/migrate/20230830115221_add_openid_scope.rb
milan-cvetkovic: Thanks for catching this. There is a number of classes used in both Oauth1 and Oauth2, for a moment I thought that these tables are used for both.

It was my (failed) attempt to fix the Oauth1 tests failing due to new `openid` scope in `lib/oauth.rb`. I will have to take into account that this scope is _only_ applicable to Oauth2,

db/migrate/20230830115220_validate_create_doorkeeper_openid_connect_tables.rb
milan-cvetkovic: Added clarifying comments in migration scripts.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,76 @@
...
+  issuer do |_resource_owner, _application|
+    "https://www.openstreetmap.org/oauth2"
+  end
+
+  signing_key Settings.doorkeeper_signing_key
+
+  subject_types_supported [:public]
+
+  resource_owner_from_access_token do |access_token|
+    # Example implementation:
```
tomhughes, line 13: This is no longer an example so we should remove this comment.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,76 @@
...
+    User.find_by(:id => access_token.resource_owner_id)
+  end
+
+  auth_time_from_resource_owner do |resource_owner|
```
tomhughes, line 17: If we're not going to provide this do we actually need the empty block or can we just remove it? The documentation says it's optional but recomended which suggests we can remote it?

The same goes for the reauthentication and account selections blocks, though in theory we could possibly implement those ones while we can't currently do this one as we don't record login times.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,76 @@
...
+  #   self.response_body = nil
+  #   @_response_body = nil
+  select_account_for_resource_owner do |resource_owner, return_to|
+    # Example implementation:
+    # store_location_for resource_owner, return_to
+    # redirect_to account_select_url
+  end
+
+  subject do |resource_owner, _application|
+    # Example implementation:
```
tomhughes, line 42: Another comment that should be removed, and I'd probably remove the second one with the alternate implementation as well.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,76 @@
...
+  # Protocol to use when generating URIs for the discovery endpoint,
+  # for example if you also use HTTPS in development
+  # protocol do
+  #   :https
+  # end
+
+  # Expiration time on or after which the ID Token MUST NOT be accepted for processing. (default 120 seconds).
+  # expiration 600
+
+  # Example claims:
```
tomhughes, line 58: I'd remove all this given we have a claims block below.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,76 @@
...
+    resource_owner.id
+
+    # or if you need pairwise subject identifier, implement like below:
+    # Digest::SHA256.hexdigest("#{resource_owner.id}#{URI.parse(application.redirect_uri).host}#{'your_secret_salt'}")
+  end
+
+  # Protocol to use when generating URIs for the discovery endpoint,
+  # for example if you also use HTTPS in development
+  # protocol do
+  #   :https
```
tomhughes, line 52: Not sure what to do here... I for one use https in dev but then again if it's only for discovery then that possibly doesn't matter.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  #   normal_claim :_bar_ do |resource_owner|
+  #     resource_owner.bar
+  #   end
+  # end
+
+  claims do
+    claim :preferred_username, :scope => :openid do |resource_owner, _scopes, _access_token|
+      # Pass the resource_owner's preferred_username if the application has
+      # `profile` scope access. Otherwise, provide a more generic alternative.
```
tomhughes, line 72: I'd drop this comment, especially as it's untrue because we're not requiring the profile scope like the original example code did.

Though actually if I understand the documentation right then the `profile` scope is required for this claim by default so I'm not sure why they were checking it again? If so then we need to add that scope...

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,76 @@
...
+      resource_owner.display_name
+    end
+  end
```
tomhughes, line 75: We should add `profile` and `email` claims - note that the email claim requires the email scope I think but we already have that as a privileged scope.

lib/oauth.rb
```diff
@@ -14,8 +15,10 @@ def description
     end
   end
 
-  def self.scopes(privileged: false)
-    scopes = SCOPES
+  def self.scopes(oauth2: false, privileged: false)
+    scopes = []
+    scopes += OAUTH2_SCOPES if oauth2
+    scopes += SCOPES
```
tomhughes, line 21: I'd put `SCOPES` back first as in the old code, and move `OAUTH2_SCOPES` to the end for consistency with ordering in other places.

tomhughes (changes requested):
Some more review comments based on my initial reading of the specification and the gem documentation.

I think we need to understand exactly the interaction between claims and scopes and which scopes we need to enable the different claims and I still need to read up on discovery though I'm not sure if that matters to us.

We also need to think about whether we want to make this generally available or whether it should be privileged and just to support internal use cases like the wiki.

config/initializers/doorkeeper_openid_connect.rb
tomhughes: Ah actually I see you scoped it to `openid` which overrides the default of `profile` so that's fine.

config/initializers/doorkeeper_openid_connect.rb
tomhughes: Actually we should just be able to to:

```ruby
protocol do
  Settings.server_protocol.to_sym
end
```

config/initializers/doorkeeper_openid_connect.rb
milan-cvetkovic: It looks like `auth_time_from_resource_owner` needs to exist even if empty, otherwise there is an error originating in [id_token.rb](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/5987683ccc22262beb6e44c76ca4b65288d6067a/lib/doorkeeper/openid_connect/id_token.rb#L68):

```
Doorkeeper::OpenidConnect::Errors::InvalidConfiguration - Failure due to Doorkeeper::OpenidConnect.configure.auth_time_from_resource_owner missing configuration.:
```

`reauthenticate_resource_owner` is used for `prompt=login` oidc parameter. If not present, or not functional (empty block is not functional) 500 error is triggered.
`select_account_for_resource_owner` is used for `prompt=select_account` oidc parameter. If not present, 500 error is triggered due to missing configuration (similar to the error as above). If empty body, it silently does not offer a selection list, but uses regular login flow.

I think it is safe to remove the latter two empty blocks.

config/initializers/doorkeeper_openid_connect.rb
milan-cvetkovic: `email` claim would normally go to `email` scope.

I was reluctant to offer email claim and scope due to it not being public information on OSM. If we add it, it appears that an application would be able to collect emails of OSM users, although with user's permission.

As for `profile`, I am guessing you had `profile` scope in mind, rather than claim. I wasn't really sure what claims to put into it.

`preferred_username` is sort of basic information, so I put it together with userid,

We could put other public information available in `/api/0.6/user/1.json` endpoint as part of `profile` scope. I didn't do it in this run, to simplify the PR. Also, I am not too sure how privileged accounts work (or should work) with Oauth

tomhughes: We should have `email` but attach it to the existing `email` scope which is a privileged scope that can only be enabled by administrators - we use it for community.osm.org and could do so for the wiki as well.

milan-cvetkovic: How would I test this locally - in other words how do I make my self an administrator in localhost setup?

tomhughes: You'll have to do it manually following the instructions at https://github.com/openstreetmap/openstreetmap-website/blob/master/CONFIGURE.md#giving-administratormoderator-permissions for the first administrator - after that an administrator can grant administrator and/or moderator to others by using the stars beside their name on the profile page.

milan-cvetkovic:
I added `email` claim to `read_email` privileged scope. id_token now returns `sub`, `preferred_name`, and `email` claims.

Please let me know if I should create separate `profile` claim, and what to put in it.

tomhughes:
The `profile` claim is probably not important for us, at least for the wiki. If we were to do it then I imagine `/user/display_name` would be the place to point it.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -27,12 +27,36 @@
   end
 
   claims do
-    claim :preferred_username, :scope => :openid do |resource_owner, _scopes, _access_token|
+    claim :preferred_username, :response => [:id_token, :user_info] do |resource_owner, _scopes, _access_token|
```
tomhughes, line 30: What was wrong with binding this to the `openid` scope?

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -27,12 +27,36 @@
...
       resource_owner.display_name
     end
 
     claim :email, :scope => :read_email, :response => [:id_token, :user_info] do |resource_owner, _scopes, _access_token|
       resource_owner.email
     end
+
+    claim :email_verified, :scope => :read_email, :response => [:id_token, :user_info] do |resource_owner, _scopes, _access_token|
```
tomhughes, line 38: I'm not sure we want to get into the business of claiming to know whether emails are valid, and that flag is I think always true for any active account anyway.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -27,12 +27,36 @@
...
+      resource_owner.email_valid
+    end
+
+    claim :profile do |resource_owner, _scopes, _access_token|
+      "https://www.openstreetmap.org/user/#{resource_owner.display_name}"
+    end
+
+    claim :description do |resource_owner, _scopes, _access_token|
```
tomhughes, line 46: I'm not sure how useful this, or any of the other custom claims you've added are given they're all OSM specific.

config/initializers/doorkeeper_openid_connect.rb
milan-cvetkovic: I guess it is not strictly required either way. https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims associates `preferred_username` with `profile` scope. However my tests with some providers show that it is returned when only `openid` scope is requested.

Contrary to the spec, they show up in id_token in addition to showing up in response to userinfo endpoint.

I was under the impression that you geared towards having `profile` scope with at least `profile` claim.

config/initializers/doorkeeper_openid_connect.rb
milan-cvetkovic: Ok, so maybe it is best to remove the extra claims, and leave at minimum required.

config/initializers/doorkeeper_openid_connect.rb
tomhughes: Initially I hadn't realised that you had linked it to the `openid` scope so I was thinking we needed to add `profile` for anybody to see it but then I realised my error. You might be right that we should use the default though.

milan-cvetkovic: I am a bit puzzled here too - some libraries ask for both "openid profile" scopes by default, and then the client app would add additional scopes to that. If we only have "openid" scope that would fail - the client app would have to do extra effort to remove "profile". I do not think this is mandated by the spec, but I guess most of the providers (if not all) offer both `openid` and `profile` scopes. They often put quite a bit of information from profile which is not required into id_token even though it is not required to be there (`name`, `email`, `preferred_name` often go to id_token...

milan-cvetkovic:
@tomhughes 

> I think we need to understand exactly the interaction between claims and scopes and which scopes we need to enable the different claims and I still need to read up on discovery though I'm not sure if that matters to us.
> 
> We also need to think about whether we want to make this generally available or whether it should be privileged and just to support internal use cases like the wiki.

The well-known endpoints (discovery) are required for openid connect to work, at least using doorkeeper-openid_connect plugin. In particular, the openid-configuration endpoint is queried for token_endpoint, jwksuri_endpoint and userinfo_endpont used in later phases.

Any 3rd party app using OSM as ID provider would require at least a persistent value correlated to userid, which would be returned as a `sub` field in `id_token`.

tomhughes:
Sure I realise we need to provide an ID but I didn't realise discovery was essential since it seemed to be a separate component but if we need it then fine.

One thing we do need to think about is how to test this...

milan-cvetkovic:
I tested most of the flow manually.

I did try to write automated tests, but was not successful. The most I got is to verify the existence of discovery endpoints.

I am not too familiar with the testing framework, but the problems started when initiating authorization, the test was not able to access the endpoints reported by openid-configuration, as if the routes were not present.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+Doorkeeper::OpenidConnect.configure do
+  issuer do |_resource_owner, _application|
+    "https://www.openstreetmap.org/oauth2"
```
mmd-osm, line 5: Do we really want a hardcoded server name here?
`"https://#{Settings.server_url}/oauth2"`

milan-cvetkovic: Is the value of `Settings.server_url` set to www.openstreetmap.org, and `Settings.server_protocol` to `https` in production? The template in Settings.yml shows different values.

Based on https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata, the value should be same as what webfinger returns, although this is somewhat conflicting in development environment - `http://localhost:3000/`.

Assuming Settings.server_url and server_protocol are set to https://www.openstreetmap.org, we should use
`"#{Settings.server_protocol}://#{Settings.url}"`

tomhughes: Yes, they are set by https://github.com/openstreetmap/chef/blob/eaf2ba261061613a6e23f6514afeb4f50817db66/cookbooks/web/resources/rails_port.rb#L349 so look like:

```yaml
server_protocol: https
server_url: www.openstreetmap.org
```

milan-cvetkovic: Updated to use `server_protocol` and `server_url` instead of hard-coded value

test/integration/oauth2_test.rb
```diff
@@ -65,13 +66,57 @@ def test_oauth2_pkce_s256
...
+    code = validate_redirect(client, state)
+
+    tokens = request_tokens(client, code, verifier)
+    id_token = tokens[:id_token]
+    access_token = tokens[:access_token]
+
+    assert_not_nil(id_token)
+
+    data, _headers = JWT.decode id_token, Doorkeeper::OpenidConnect.signing_key.keypair, true, {
+      :algorithm => [Doorkeeper::OpenidConnect.signing_algorithm.to_s],
```
tomhughes, line 87: We should probably get the key and algorithm the same way a real client will, by fetching `/oauth2/discovery/keys`.

In fact we should probably fetch `/.well-known/openid-configuration` and do some validation on it, then follow the `jwks_uri` entry to get the keys.

test/integration/oauth2_test.rb
```diff
@@ -116,6 +161,11 @@ def validate_redirect(client, state)
   end
 
   def request_token(client, code, verifier = nil)
+    tokens = request_tokens(client, code, verifier)
+    tokens[:access_token]
+  end
+
+  def request_tokens(client, code, verifier = nil)
```
tomhughes, line 168: I think we should leave this called `request_token` and get rid of the other routine, then have this just return `token` and have the callers pick out the bits they want.

tomhughes (changes requested):

tomhughes:
> Are there more changes to this PR that you had in mind?

I've done a new review which is mostly about the test.

Other than that it's on the agenda for this weeks operations call to discuss whether we want to make this globally available or keep it for internal use. I had hoped to have that discussion on the last call but we ran out of time.

test/integration/oauth2_test.rb
milan-cvetkovic: This was my original intention. The problem is that in test environment discovery endpoint returns a mix of test host names in the remaining endpoints:
- `issuer` is based on server_url and server_protocol, consistently in test and development
- the remaining endpoints have correct paths, but have different hosts in development and test:
   - host in development is `http://localhost:3000` for all remaining endpoints
   - host in testing is `http://www.example.com`

For example, `jwks_uri` is set to `http://www.example.com/oauth2/discovery/keys`.
I am not too sure where the example.com comes from, and how it disapears in development...

I added a separate test for discovery endpoint which verifies that the endpoints have correct path. More importantly it gets the key informrmation from jwks endpoint (using path only) and verifies that kid is in fact kid of the configured key

tomhughes (approved):

tomhughes:
I made a few minor textual cleanups and merged this.

As it involves a migration I'm going to wait until the database backup which is currently running completes (which should be tomorrow) before deploying it.

PR #4235, opened 2023-09-07
Don't use monospace font in single-line share inputs

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/commit/dc24f6c2195c6053efd43778d53ae68984e54107#commitcomment-126688182

If you don't want custom font sizes even for things that are guaranteed not to fit on screen, maybe you don't want custom fonts as well? The buttons are not in a monospace font, why the inputs are, if they are supposed to be consistent with the buttons? You can argue that if an input has just a single-line url, monospace font is not required. Look at your browser's location bar. Does it have a monospace font? Probably no.

Before:
image

After:
image

The characters are not as wide, mote text fits on screen.

### DIFF
--- a/app/assets/javascripts/leaflet.share.js
+++ b/app/assets/javascripts/leaflet.share.js
@@ -68,7 +68,7 @@ L.OSM.share = function (options) {
       .append($("<input>")
         .attr("id", "long_input")
         .attr("type", "text")
-        .attr("class", "form-control form-control-sm font-monospace")
+        .attr("class", "form-control form-control-sm")
         .attr("readonly", true)
         .on("click", select));
 
@@ -79,7 +79,7 @@ L.OSM.share = function (options) {
       .append($("<input>")
         .attr("id", "short_input")
         .attr("type", "text")
-        .attr("class", "form-control form-control-sm font-monospace")
+        .attr("class", "form-control form-control-sm")
         .attr("readonly", true)
         .on("click", select));
 


gravitystorm:
I don't understand the motivation for this change. [A few minutes ago you posted a screenshot of the github url-copy widget](https://github.com/openstreetmap/openstreetmap-website/commit/dc24f6c2195c6053efd43778d53ae68984e54107#commitcomment-126693406), which I thought you were using as a good example. But it uses monospace for - and only for - the bit that shows the URL. So if that's a good example, then why this PR?

> If you don't want custom font sizes even for things that are guaranteed not to fit on screen, maybe you don't want custom fonts as well?

There's a difference between removing custom CSS (i.e. font sizes that were previously defined in custom.css), and using bootstrap rules for fonts. So your statement is a non-sequitur, since this PR isn't removing custom fonts - it's not touching anything in custom.css.

I think it would be better to focus on why this change should be made, i.e. what improvements it brings, instead.

AntonKhorev:
> But it uses monospace for - and only for - the bit that shows the URL. So if that's a good example, then why this PR?

Because I don't expect to convince you to decrease the font size you've just increased.

> There's a difference between removing custom CSS (i.e. font sizes that were previously defined in custom.css), and using bootstrap rules for fonts. So your statement is a non-sequitur, since this PR isn't removing custom fonts - it's not touching anything in custom.css.

You removed custom css AND changed font size. You could have done only one of these things instead. You are explaining that now it's better because the inputs are consistent with the buttons. But they aren't because they use different fonts.

> I think it would be better to focus on why this change should be made, i.e. what improvements it brings, instead.

Because the characters are not as wide, that's another way to fit more text.

gravitystorm:
I've thought about this some more, and I'm going to decline this change. I don't think it's important to use a proportional font here just to squeeze in a few more characters. 

A alternative improvement would be to move closer to the github widget previously mentioned, i.e. have a click-to-copy button.

PR #4239, opened 2023-09-08
Before/after id pagination for user diary comments

AntonKhorev:
for `/user/<username>/diary/comments` pages

similar to diary entry and trace pagination


app/controllers/diary_entries_controller.rb
```diff
@@ -292,4 +276,24 @@ def set_map_location
...
+                 else
+                   items.order(:id => :desc)
+                 end
+
+    page_items = page_items.limit(20)
+    page_items = page_items.includes(includes)
+    page_items = page_items.sort.reverse
+
+    newer_items_id = page_items.count.positive? && items.exists?(["#{id_column} > ?", page_items.first.id]) && page_items.first.id
+    older_items_id = page_items.count.positive? && items.exists?(["#{id_column} < ?", page_items.last.id]) && page_items.last.id
```
tomhughes, line 295: I think it would be better to write these as guarded assignments - it took me quite a while to figure out what you were doing. I suggest doing:

```ruby
newer_items_id = page_items.first.id if page_items.count.positive? && items.exists?(["#{id_column} > ?", page_items.first.id])
```

AntonKhorev: I wrote it differently first, then Rubocop complained and told that those x = a && b are also an option.

AntonKhorev: Updated

Issue #1618, opened 2017-08-24
API for user blocks


willemarcel:
In OSMCha, we are adding a suspicion reason "User has multiple blocks" to changesets whose creator has more than 1 block. I believe it would be better if we consider the time distance from the last block and the number of changesets the user has. That way, we wouldn't flag changesets of users that were blocked many months ago, created a lot of changesets and probably learned the good practices.

For this we would need an API to user_blocks. Do the maintainers have something against having a endpoint to list general `user_blocks` and list `user_blocks` of a specific user? I can work on this if you consider we can have those endpoints.

simonpoole:
When I suggested adding this to India I was aware of the issue and thought s bit about how to resolve it.

The easy, low impact variant is to add a last_block attribute with the date the block expired. This can be done without introducing a new API call.

willemarcel:
@simonpoole Yes, it could solve as well.

woodpeck:
User blocks do not necessarily have anything to do with the user making bad edits. Someone could have made bad edits three weeks ago which need to be reverted but they won't get a block since the issue doesn't call for it. Another person could be making the same edits today and we block them in order to stop them. A third person could be making totally legitimate edits but due to a misconfiguration not see emails generated by changeset comments, and receive a block because that's the only way to get their attention.

I wouldn't like seeing blocks used to treat an user as "more suspicious" because this would, in turn, lead to calls for *removing* past block information from users who have "done nothing wrong"...

willemarcel:
@woodpeck I agree almost completely with you. We are adding information to the changesets to help other users to prioritize which edits should receive attention and be reviewed. We removed the word "suspicion" from the frontend in the hope that the users don't see it as an accusation.

We only flag users that have more than one block, so if someone had a misconfiguration in his email address and received only one block, it'll not affect him. The third case you mentioned shouldn't really be treated as more "suspicious", although I believe this block case in an exception.

Moreover, I think you'll agree that communication is an important aspect in a collaborative project and users must care about have a correct email configured and answer other contributors questions.

mmd-osm:
I think you need to at least define a cut off period of say 6-12 months, otherwise changesets like the following https://osmcha.mapbox.com/changesets/51453028 are getting flagged as "User has multiple flags" - that was like 5 years ago!

What if a users doesn't want osmcha (or any other app) to use that information for whatever purpose? Is there some opt-out mechanism? Do we need user consent to expose that information to external apps?

[...]

PR #4240, opened 2023-09-08
Add show user block api endpoint

AntonKhorev:
Going through the DWG Wishlist, there was a blocks api.

Currently scripted blocks in [osmtools](https://github.com/woodpeck/osm-revert-scripts) are done by emulating the browser, loading webpages and filling out forms. That requires passing the password around. It would be better to be able to access block functionality with a proper api and an oauth token.

But before doing block creation, there need to be block read api endpoints. Here the xml block show method is added at `/api/0.6/user_blocks/:id` with an output like:

```
<osm version="0.6" ...>
<user_block id="6" created_at="2023-08-01T15:55:58Z" updated_at="2023-08-05T00:46:36Z" ends_at="2023-08-05T00:46:36Z" needs_view="false">
  <user uid="1" user="fakeuser1"/>
  <creator uid="2" user="fakemod1"/>
  <revoker uid="2" user="fakemod1"/>
  <reason>longer block test</reason>
</user_block>
</osm>
```

All of this is already accessible through the web ui at `/user_blocks/:id`.

### DIFF
--- a/app/controllers/api/user_blocks_controller.rb
+++ b/app/controllers/api/user_blocks_controller.rb
@@ -1,0 +1,18 @@
+module Api
+  class UserBlocksController < ApiController
+    before_action :check_api_readable
+
+    authorize_resource
+
+    around_action :api_call_handle_error, :api_call_timeout
+    before_action :set_request_formats
+
+    def show
+      raise OSM::APIBadUserInput, "No id was given" unless params[:id]
+
+      @user_block = UserBlock.find(params[:id])
+    rescue ActiveRecord::RecordNotFound
+      raise OSM::APINotFoundError
+    end
+  end
+end
--- a/test/controllers/api/user_blocks_controller_test.rb
+++ b/test/controllers/api/user_blocks_controller_test.rb
@@ -7,6 +7,10 @@ module Api
         { :path => "/api/0.6/user_blocks/1", :method => :get },
         { :controller => "api/user_blocks", :action => "show", :id => "1" }
       )
+      assert_routing(
+        { :path => "/api/0.6/user_blocks/1.json", :method => :get },
+        { :controller => "api/user_blocks", :action => "show", :id => "1", :format => "json" }
+      )
     end
 
     def test_show
@@ -15,6 +19,12 @@ module Api
       get api_user_block_path(:id => block)
       assert_response :success
       assert_select "user_block[id='#{block.id}']", 1
+
+      get api_user_block_path(:id => block, :format => "json")
+      assert_response :success
+      js = ActiveSupport::JSON.decode(@response.body)
+      assert_not_nil js
+      assert_equal block.id, js["user_block"]["id"]
     end
 
     def test_show_not_found
--- a/test/controllers/api/user_blocks_controller_test.rb
+++ b/test/controllers/api/user_blocks_controller_test.rb
@@ -1,0 +1,26 @@
+require "test_helper"
+
+module Api
+  class UserBlocksControllerTest < ActionDispatch::IntegrationTest
+    def test_routes
+      assert_routing(
+        { :path => "/api/0.6/user_blocks/1", :method => :get },
+        { :controller => "api/user_blocks", :action => "show", :id => "1" }
+      )
+    end
+
+    def test_show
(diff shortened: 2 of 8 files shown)


mmd-osm:
#1618 seems to be related to this one

test/controllers/api/user_blocks_controller_test.rb
```diff
@@ -0,0 +1,30 @@
...
+  class UserBlocksControllerTest < ActionDispatch::IntegrationTest
+    def test_routes
+      assert_routing(
+        { :path => "/api/0.6/user_blocks/1", :method => :get },
+        { :controller => "api/user_blocks", :action => "show", :id => "1" }
+      )
+      assert_recognizes(
+        { :controller => "api/user_blocks", :action => "show", :id => "1", :format => "xml" },
+        { :path => "/api/0.6/user_blocks/1.xml", :method => :get }
+      )
```
tomhughes, line 13: We don't normally make the XML case explicit like that though I guess maybe we should have implicit and explicit cases for it but looking at other API controller tests at the moment only the JSON ones are explicit.

tomhughes (commented):
Should we be adding JSON output as well?

test/controllers/api/user_blocks_controller_test.rb
AntonKhorev: I copied it from some other place where it was.

AntonKhorev: This is where I copied it from:

https://github.com/openstreetmap/openstreetmap-website/blob/0e21afc56593eb9fce14d8cfe0984494e758c150/test/controllers/api/notes_controller_test.rb#L23

AntonKhorev:
> #1618 seems to be related to this one

This PR alone won't be enough for what was originally asked there. The next required thing is a search endpoint with the ability to search by usernames/ids.

> Should we be adding JSON output as well?

Yes, if the xml output format is ok. The biggest decision here was to make separate user/creator/revoker subelements.

test/controllers/api/user_blocks_controller_test.rb
AntonKhorev: Most of assert_recognizes tests were added for paths differing from the default ones
https://github.com/openstreetmap/openstreetmap-website/commit/90e46a58de583ac2bd7bb43077faa94186957a7a

I guess it's not required here, so I removed it. And added JSON.

tomhughes:
I assume there's nothing being exposed here that isn't already visible through the web interface? I think pretty much everything is visible there already?

mmd-osm:
As a normal user, I don't see any of the "uid" values, neither for the user which has been blocked, nor for the creator and/or potential block revoker.

By the way, does this also work for deleted users, like this one: https://www.openstreetmap.org/user_blocks/14719

app/controllers/api/user_blocks_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+  class UserBlocksController < ApiController
+    before_action :check_api_readable
+
+    authorize_resource
+
+    around_action :api_call_handle_error, :api_call_timeout
+    before_action :set_request_formats
+
+    def show
+      raise OSM::APIBadUserInput, "No id was given" unless params[:id]
```
mmd-osm, line 11: Maybe make it a bit clearer, which "id" we expect here, like "No user block id was given".

tomhughes:
> As a normal user, I don't see any of the "uid" values in the web interface, neither for the user which has been blocked, 
nor for the creator and potential block revoker.

Right, but in general the mapping between username and ids is not hidden - the user details call will get it for example.

AntonKhorev:
The web interface prefers to work with usernames while the api prefers user ids. For example you need a user id for [user details](https://wiki.openstreetmap.org/wiki/API_v0.6#Details_of_a_user:_GET_/api/0.6/user/#id). When you know a username and want to get the corresponding id through the api, you typically do a changeset search by the username and get the id from the search results. It's only difficult to do if the user has no contributions, but then, if it's a genuine user and not a pre-blocked vandal account, there has to be a problem with their contributions and changesets (or notes) searchable by the username exist.

PR #1642, opened 2017-09-29
Add list views of changeset comments

paulgillard:
This commit adds the following:
  * View of all changeset comments made on all changesets
  * View of all changeset comments made by a given user
  * View of all changeset comments made on a changeset created by a given user
  * View of all changeset comments made on a changeset subscribed to by logged in user

This commit builds on initial work by:
  Łukasz Gurdek <git@ukasiu.pl>
with additional work by:
  Éric Gillet <eric.gillet+gh@linuxw.info>
  Victor Grousset <victor@tuxayo.net>

PR #4248, opened 2023-09-14
List of user's changeset comments

AntonKhorev:
Add user's changeset comments next to user's diary comments.

image

Pagination is reused along with traces list.
Localization strings are mostly kept in diary comments section for now.

### DIFF
--- a/app/controllers/changeset_comments_controller.rb
+++ b/app/controllers/changeset_comments_controller.rb
@@ -1,0 +1,19 @@
+class ChangesetCommentsController < ApplicationController
+  include UserMethods
+
+  layout "site"
+
+  before_action :authorize_web
+  before_action :set_locale
+
+  authorize_resource
+
+  before_action :lookup_user
+  before_action -> { check_database_readable(:need_api => true) }
+  around_action :web_timeout
+
+  def index
+    @title = t ".title", :user => @user.display_name
+    @comments = []
+  end
+end
--- a/app/controllers/changeset_comments_controller.rb
+++ b/app/controllers/changeset_comments_controller.rb
@@ -1,5 +1,6 @@
 class ChangesetCommentsController < ApplicationController
   include UserMethods
+  include PaginationMethods
 
   layout "site"
 
@@ -14,6 +15,12 @@ class ChangesetCommentsController < ApplicationController
 
   def index
     @title = t ".title", :user => @user.display_name
-    @comments = []
+
+    comments = ChangesetComment.where(:author => @user)
+    comments = comments.visible unless current_user&.moderator?
+
+    @params = params.permit(:display_name, :before, :after)
+
+    @comments, @newer_comments_id, @older_comments_id = get_page_items(comments, :includes => [:author])
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -437,6 +437,8 @@ en:
         title_particular: "OpenStreetMap changeset #%{changeset_id} discussion"
       timeout:
         sorry: "Sorry, the list of changeset comments you requested took too long to retrieve."
+    index:
+      title: "Changeset Comments added by %{user}"
   changesets:
     changeset:
       no_edits: "(no edits)"
@@ -598,8 +600,9 @@ en:
   diary_comments:
     index:
       title: "Diary Comments added by %{user}"
-      heading: "%{user}'s Diary Comments"
-      subheading_html: "Diary Comments added by %{user}"
+      heading_html: "%{user}'s Comments"
+      diary_entries: "Diary entries"
+      changesets: "Changesets"
       no_comments: "No diary comments"
     page:
       post: Post
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -603,9 +603,10 @@ en:
       heading_html: "%{user}'s Comments"
       diary_entries: "Diary entries"
       changesets: "Changesets"
-      no_comments: "No diary comments"
+      no_comments: "No comments"
     page:
       post: Post
+      changeset: Changeset
       when: When
       comment: Comment
     new:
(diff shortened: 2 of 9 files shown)


config/locales/en.yml
```diff
@@ -548,10 +550,12 @@ en:
         description: "Recent diary entries from users of OpenStreetMap"
     comments:
       title: "Diary Comments added by %{user}"
-      heading: "%{user}'s Diary Comments"
-      subheading_html: "Diary Comments added by %{user}"
-      no_comments: "No diary comments"
+      heading_html: "%{user}'s Comments"
+      for_diary: "For diary"
+      for_history: "For map edits"
```
tomhughes, line 555: I don't think the "For" prefix reads very well here and there's also inconsistent pluralisation between the two tabs. I think I would just go with `Diaries` or `Diary entries` for the first one and `Changesets` for the second - we could do `Map edits` but we already using changesets in other places so I think we should be consistent.

config/routes.rb
```diff
@@ -112,11 +112,12 @@
   get "/relation/:id" => "browse#relation", :id => /\d+/, :as => :relation
   get "/relation/:id/history" => "browse#relation_history", :id => /\d+/, :as => :relation_history
   get "/changeset/:id" => "browse#changeset", :as => :changeset, :id => /\d+/
-  get "/changeset/:id/comments/feed" => "changeset_comments#index", :as => :changeset_comments_feed, :id => /\d*/, :defaults => { :format => "rss" }
+  get "/changeset/:id/comments/feed" => "changeset_comments#feed", :as => :changeset_comments_feed, :id => /\d*/, :defaults => { :format => "rss" }
   resources :notes, :path => "note", :only => [:show, :new]
 
   get "/user/:display_name/history" => "changesets#index"
   get "/user/:display_name/history/feed" => "changesets#feed", :defaults => { :format => :atom }
+  get "/user/:display_name/history/comments" => "changeset_comments#index", :as => :history_comments
```
tomhughes, line 120: I think we should use `changeset_comments` as the route name. I know that, for hysterical raisins, we use `history` in the URL but the controller is called `changeset_comments`.

tomhughes (commented):
I wonder if we should merge note comments under this interface eventually, and make the current user notes view just show notes they opened and not every comment - maybe as part of the planned restructuring of notes to make notes less linked to the initial comment.

config/routes.rb
gravitystorm: We should try to avoid the `:as` setting entirely. As we move to more resourceful routing, it becomes less and less necessary. Generally we want to use resourceful routing, and accept the auto generated route names as-is.

I need to think about the controller structure for this PR. I know that (ignoring the API) we've currently got a flat hierarchy of controllers, but this approach has disadvantages. For example, it's not clear that `app/controllers/changeset_comments_controller.rb#index` would be a user-scoped subset of changeset comments, and it would clash with any implementation of site-wide changeset_comments#index (e.g. for moderation). I want to consider the pros and cons of having a `user/changeset_comments` controller but I need to play around with that.

In any case, I would ask for resourceful routing here.

mmd-osm:
We seem to have some existing work on this topic: #1642
Anything which could be reused from there?

AntonKhorev:
@mmd-osm you have to look why https://github.com/openstreetmap/openstreetmap-website/pull/1642 failed. The reasons in its comments are:

- "offset/limit based pagination" - I waited for the maintainers to implement whatever they consider to be a correct pagination. Eventually this happened and now I'm copying it.
- "it's quite large" - I'm splitting this thing in steps and doing one functional step + one refactoring step. This particular pull request at this moment is in *we want more refactoring* state.

config/routes.rb
AntonKhorev: Renamed to `changeset_comments`. But there's also `changeset_comment` path.

config/locales/en.yml
AntonKhorev: Changed to "Diary entries" and "Changesets". But user profile menu has "Edits" (plural, changesets not mentioned) and "Diary" (singular, users don't have multiple diaries).

AntonKhorev:
> I wonder if we should merge note comments under this interface eventually, and make the current user notes view just show notes they opened and not every comment - maybe as part of the planned restructuring of notes to make notes less linked to the initial comment.

Adding note comments here is an obvious extension, however adding note actions is not as obvious. Note comments are currently tied to actions such as closing and reopening. This is something that [the planned restructuring](https://github.com/openstreetmap/openstreetmap-website/issues/3831) completely ignores. If I close a note without writing anything, is it a comment that should be shown here?

app/controllers/changeset_comments_controller.rb
```diff
@@ -1,15 +1,32 @@
...
+    comments = comments.visible unless current_user&.moderator?
+
+    @params = params.permit(:display_name, :before, :after)
+
+    @comments, @newer_comments_id, @older_comments_id = get_page_items(comments, [:author])
+  end
+
   ##
   # Get a feed of recent changeset comments
-  def index
```
gravitystorm, line 12: As per discussion at https://github.com/openstreetmap/openstreetmap-website/pull/4587 we should avoid creating new controller actions (or renaming existing ones) with non-resourceful names.

app/controllers/concerns/pagination_methods.rb
```diff
@@ -0,0 +1,27 @@
+module PaginationMethods
```
gravitystorm, line 1: I like extracting this to a concern :+1:

gravitystorm (changes requested):

app/controllers/concerns/pagination_methods.rb
AntonKhorev: merged in another pull request

app/controllers/changeset_comments_controller.rb
AntonKhorev: Rebased this pull request on the first commit of [unmerged pull request](https://github.com/openstreetmap/openstreetmap-website/pull/4587) because it moves feeds out of the way.

config/routes.rb
AntonKhorev: For "think about the controller structure ... would clash with any implementation of site-wide ... (e.g. for moderation)" you can create another controller when that clash happens, see for example #4969.

tomhughes:
Superseded by #5518.

PR #4250, opened 2023-09-16
Don't include unused bootstrap css

AntonKhorev:
Include only Bootstrap components that are in use as recommended [here](https://getbootstrap.com/docs/5.1/customize/sass/#importing)
by creating `_bootstrap-custom.scss` as described [here](https://github.com/twbs/bootstrap-rubygem#sass-individual-components).
Results in ~10% smaller css.

Actually I do this to modify Bootstrap's theme colors. It has to be done after importing variables but before importing components, and I can't do that if everything is imported in one `@import`.

### DIFF
--- a/app/assets/stylesheets/_bootstrap-custom.scss
+++ b/app/assets/stylesheets/_bootstrap-custom.scss
@@ -1,0 +1,53 @@
+/*!
+ * Bootstrap v5.1.3 (https://getbootstrap.com/)
+ * Copyright 2011-2021 The Bootstrap Authors
+ * Copyright 2011-2021 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/main/LICENSE)
+ */
+
+// scss-docs-start import-stack
+// Configuration
+@import "bootstrap/functions";
+@import "bootstrap/variables";
+@import "bootstrap/mixins";
+@import "bootstrap/utilities";
+
+// Layout & components
+@import "bootstrap/root";
+@import "bootstrap/reboot";
+@import "bootstrap/type";
+@import "bootstrap/images";
+@import "bootstrap/containers";
+@import "bootstrap/grid";
+@import "bootstrap/tables";
+@import "bootstrap/forms";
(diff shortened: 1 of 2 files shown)


gravitystorm:
In effect this is the opposite of https://github.com/openstreetmap/openstreetmap-website/pull/2560 where I included all of bootstrap (after solving various compatibility issues with our pre-bootstrap custom stylesheets).

At that time I wanted to include everything by default, in order to make it easier for other developers to use any component they see when reading the bootstrap documentation. So my hesitation here is that by customising the list of components to restrict it to only those that are in use today, it makes it slightly harder to use new components (or components that are added in future versions of bootstrap). For example, if someone has an idea of how to improve a page (e.g. the about page) by adding a component (e.g. breadcrumbs or carousel or something new in 5.4 etc), would they figure out why it's not working, or would they just give up first? What do you think?

> Actually I do this to modify Bootstrap's theme colors. It has to be done after importing variables

Can you expand on this please, perhaps with an example? We already change some of the theme colours in parameters.scss, which is included before bootstrap. [Bootstrap sets its variables with the `!default` flag to allow this](https://getbootstrap.com/docs/5.1/customize/sass/#variable-defaults). Are you trying to do something different or more advanced than this?

AntonKhorev:
> Can you expand on this please, perhaps with an example?

https://getbootstrap.com/docs/5.1/customize/sass/#add-to-map

You can't add to the theme map before it's defined. Then you may try to include Bootstrap first, but if you include everything, it's too late to modify the map.

Something like this would work, if there's an editable file with Bootstrap includes:

```
// Configuration
@import "bootstrap/functions";
@import "bootstrap/variables";
@import "bootstrap/mixins";
@import "bootstrap/utilities";

$theme-colors: map-merge($theme-colors, $custom-colors); // added custom themes

// Layout & components
@import "bootstrap/root";
@import "bootstrap/reboot";
@import "bootstrap/type";
@import "bootstrap/images";
...
```

Or do you prefer to redefine the entire theme map instead of adding to it?

PR #4251, opened 2023-09-17
Show hidden comments to moderators if requested

AntonKhorev:
Continuation of https://github.com/openstreetmap/openstreetmap-website/pull/4245.

With comment ids shown it's possible to hide specific comments. But the unhiding also requires comment ids. They are not known because the comments are hidden from the api calls.

Here a parameter is added to show hidden comments. It makes hidden comments visible to users who can unhide comments. The request looks like this:

```
/api/0.6/changeset/<id>?include_discussion=true&show_hidden_comments=true
```

I chose to add a parameter instead of showing the comments to anyone with permissions in case there are too many hidden comments. Maybe later some kind of limit will be added, making changeset with discussion requests to look like search requests. In this case `show_hidden_comments` would act similar to filters like `closed` for changesets/notes.

The `visible` attribute is added to each comment to indicate if it's hidden. See also https://github.com/zerebubuth/openstreetmap-cgimap/pull/289#issuecomment-1722363471.

See #3934 for why the parameter is named `show_hidden_...` I added `_comments` at the end in case there's something else to be hidden in changeset metadata. This allows for tags or bbox hiding to be added later if necessary.

### DIFF
--- a/app/controllers/api/changesets_controller.rb
+++ b/app/controllers/api/changesets_controller.rb
@@ -6,6 +6,7 @@ module Api
 
     before_action :check_api_writable, :only => [:create, :update, :upload, :subscribe, :unsubscribe]
     before_action :check_api_readable, :except => [:create, :update, :upload, :download, :query, :subscribe, :unsubscribe]
+    before_action :setup_user_auth, :only => [:show]
     before_action :authorize, :only => [:create, :update, :upload, :close, :subscribe, :unsubscribe]
 
     authorize_resource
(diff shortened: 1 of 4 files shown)


AntonKhorev:
can't write this because Rubocop

    assert_equal true, js["changeset"]["open"]

can't write this because updated Rubocop

    assert js["changeset"]["open"]

https://github.com/openstreetmap/openstreetmap-website/commit/e8af4b85bf001ade543d03a503c769055d42e6a2 no, thanks

    assert_operator js["changeset"], :[], "open"

have to write this instead

    assert_operator true, :==, js["changeset"]["open"]

tomhughes:
The rubocop thing was a temporary aberration that I think we've discussed elsewhere, and it's not actually insisting on going back to something sensible.

Other than that it looks good so I've applied the rubocop fixups and merged.

PR #4252, opened 2023-09-18
Remove fundraising banner, add State of the Map Europe 2023 banner

marjanvandekauter-tomtom:
This PR adds the banner for State of the Map Europe 2023, which links to https://stateofthemap.eu

Start date:

- in case of no time limit: September 25, 2023 (to be removed November 10, 2023)
- in case of a one-month time limit: October 5, 2023 (to be removed November 5, 2023)


tomhughes:
You realise there can be multiple banners at the same time? Did somebody ask you to remove the fundraising banner?

tomhughes:
Per https://operations.osmfoundation.org/policies/banner/ I think the limit here is one month.

marjanvandekauter-tomtom:
> You realise there can be multiple banners at the same time? Did somebody ask you to remove the fundraising banner?

@tomhughes I followed the guidelines [guidelines](https://github.com/openstreetmap/openstreetmap-website/blob/master/FAQ.md#how-do-i-create-a-banner-to-promote-my-openstreetmap-event) which state that old banners can (optionally) be removed and noticed that previous pull requests for banners also contained one banner only. Of course, if the SotM-EU banner is accepted and the plan is to use the fundraising banner again after the one-month time window, it makes more sense to not remove the previous banner. I'm not sure who decides on banner acceptance and rotation in this case--tagging @grischard who added the fundraising banner.

tomhughes:
Right but by "old" that meant ones which had passed their end date so were no longer being displayed - the fundraising one has an end date at the end of the year so is still live.

grischard:
Can we please have both?

marjanvandekauter-tomtom:
> Right but by "old" that meant ones which had passed their end date so were no longer being displayed - the fundraising one has an end date at the end of the year so is still live.

You're right, I added the fundraising banners again.

Don't if it's possible/esthetically acceptable to have both banners. I leave it up to the OWG and CWG to decide.

tomhughes:
It's fine - only one will display at a time but they will alternate.

marjanvandekauter-tomtom:
> It's fine - only one will display at a time but they will alternate.

Great. Sounds good, if @grischard is on board with it too.

config/banners.yml
```diff
@@ -8,3 +8,9 @@ donate_2023:
     - [banners/donate_2023_1@2x.jpg, 2x]
   dark: true
   enddate: 2023-dec-31
+sotmeu_2023:
+  id: sotmeu_2023
+  alt: State of the Map Europe 2023
+  link: https://stateofthemap.eu/
+  img: StateoftheMapEurope_2023.png
```
tomhughes, line 15: This needs to be `banners/StateoftheMapEurope_2023.png` for things to work - that's why the tests are failing.

tomhughes (changes requested):

config/banners.yml
marjanvandekauter-tomtom: fixed

tomhughes:
I added a commit (https://github.com/openstreetmap/openstreetmap-website/commit/8bd52cec3f95b4af1a98114c609b2762b0ba0aa9) to limit it to the requested one month period as per the policy.

PR #4254, opened 2023-09-20
Update Dockerfile to include image manipulation binaries.

erictheise:
A number of developers on the OpenHistoricalMap project use Docker. We've not synced with this, the parent project, for a while, and in doing so we found that libraries such as `advpng`, `gifsicle`, `jhead`, `pngcrush`, `svgo` and so on–mentioned in `INSTALL.md`–were not being installed, causing warnings when running a `docker compose up`.

This pull request adds those libraries to `Dockerfile`.

I'll note that there is still one warning
```
WARN: advpng none at /usr/bin/advpng (== none) is of unknown version
```
which is [said to be a result of bad packaging](https://bugs.launchpad.net/ubuntu/+source/advancecomp/+bug/1797662).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,3 +63,4 @@ RUN bundle install
 ADD package.json yarn.lock /app/
 ADD bin/yarn /app/bin/
 RUN bundle exec bin/yarn install
+RUN bundle exec bin/yarn global add svgo
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,14 @@ RUN apt-get update \
       unzip \
       nodejs \
       npm \
+      advancecomp \
+      gifsicle \
+      libjpeg-progs \
+      jhead \
+      jpegoptim \
+      optipng \
+      pngcrush \
+      pngquant \
  && npm install --global yarn \
  # We can't use snap packages for firefox inside a container, so we need to get firefox+geckodriver elsewhere
  && add-apt-repository -y ppa:mozillateam/ppa \
--- a/package.json
+++ b/package.json
@@ -8,8 +8,7 @@
     "leaflet": "^1.8.0",
     "leaflet.locatecontrol": "^0.79.0",
     "osm-community-index": "^5.2.0",
-    "qs": "^6.9.4",
-    "svgo": "^3.0.2"
+    "qs": "^6.9.4"
   },
   "devDependencies": {
     "eslint": "^8.0.0"
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "leaflet": "^1.8.0",
     "leaflet.locatecontrol": "^0.79.0",
     "osm-community-index": "^5.2.0",
-    "qs": "^6.9.4"
+    "qs": "^6.9.4",
+    "svgo": "^3.0.2"
   },
   "devDependencies": {
     "eslint": "^8.0.0"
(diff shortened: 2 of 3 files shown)


tomhughes:
Those binaries are all optional which is why we don't bother installing them - we use them in production but there's not real need to have them for development.

erictheise:
On a practical level, installing them into the Docker image dispenses with the irritation of one's display filling with warnings at multiple stages of `docker compose up` or when a Rails console is needed.

On a "keeping parity" level, they're presented as equally important with `libxml2` and `yarn` in the [MacOSX](https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md#macosx) section of the Installation instructions.

gravitystorm:
There should be no warnings emitted in either the development or test environments - see https://github.com/openstreetmap/openstreetmap-website/pull/4055 and https://github.com/openstreetmap/openstreetmap-website/pull/1795

I can see from a brief check at https://github.com/OpenHistoricalMap/ohm-website/tree/staging/config/image_optim that you are missing the config file from #4055, so my recommendation would be to merge that into your tree and you'll see the warnings disappear.

As @tomhughes says, we found that disabling them in test and development was the better approach, since it's a lot more packages to install for practically no (non-production) benefit - it's not worth the hassle for new developers. Particularly since there are often packaging issues too!

As for the MacOSX instructions, these should be updated to remove the unnecessary packages, but I would appreciate any assistance on that topic from someone who can test the instructions and make sure they are still accurate afterwards (i.e. removing a package doesn't unexpectedly remove a dependency that was actually necessary).

tomhughes:
Looks those were originally added to the MacOS instructions in cc101f582d14a2cfc74859a537df8b34b017cc8f though they've been moved around a bit since.

erictheise:
Thanks for the feedback and tip on the config file, @gravitystorm & @tomhughes. Keeping those binaries out of dev and test is the right way to go and I'll fast track that config file into our repo tomorrow.

OHM is some months behind you but I've been given the go ahead to get things synced up.

AntonKhorev:
Aren't system test using precompiled assets, and when you precompile you get all these warnings?

gravitystorm:
> Aren't system test using precompiled assets, and when you precompile you get all these warnings?

I've just tested this locally, and no, I don't get all these warnings when running system tests. I also checked that optipng is not installed, and the warning is shown when I edit `config/image_optim/test.yml`

I've also checked CI, [which precompiles the assets during testing](https://github.com/openstreetmap/openstreetmap-website/blob/926229e286520ed23fb5e8add94c6048aa592412/.github/workflows/tests.yml#L58-L59) and again there's no warnings from image_optim - which is to be expected, since we disable them for the test environment, and any precompilation will also happen in the same environment.

@AntonKhorev If you are seeing the warnings in either the development or test environments, please open an issue so that we can investigate further.

gravitystorm:
> Keeping those binaries out of dev and test is the right way to go and I'll fast track that config file into our repo tomorrow.

Great, thanks @erictheise for the feedback. Do let us know if we do anything here which makes life unintentionally difficult for other deployments like OHM - we don't have much insights into this.

Based on this discussion, I'm going to close this PR.

erictheise:
I can confirm that adding `config/image_optim/development.yml` solves the problem this pr set out to fix. @gravitystorm, I can be available to test or help rewrite the MacOSX instructions.

PR #4255, opened 2023-09-20
Add Tracetrack Topo as featured layer

pnorman:
image


pnorman:
> Can I assume this has been approved by OWG?

Yes.

> Is there any minutes (or similar) to show that it has been formally approved?

I don't think the minutes are published. @dorotheakazazi would know.

strongwillow:
How will the API key be set in `config/settings.yml`? When can it be merged?

tomhughes:
It will be configured from chef for our production environment as we do for other layers.

PR #4257, opened 2023-09-21
Docker: Use `docker compose`

tordans:
This PR change the `docker-compose` commands to the newer `docker compose`.

My understanding is, that `docker-compose` is deprecated and `docker compose` is more or less a drop in replacement.

https://docs.docker.com/compose/migrate/

> From July 2023 Compose V1 stopped receiving updates. It’s also no longer available in new releases of Docker Desktop.

I did a fresh setup on MacOS with this updated docs and it all worked out except for some failing tests which I suspect are unrelated.

This PR also extends the "Tests" section on how to prepare the tests db to make the copy-paste steps complete.

---

Open todos

- [x] Update installation section – https://github.com/openstreetmap/openstreetmap-website/pull/4257#issuecomment-1737513727
- [x] Update github actions to use `docker compose` – https://github.com/openstreetmap/openstreetmap-website/pull/4257#issuecomment-1737549187

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,17 +19,17 @@ jobs:
         touch config/settings.local.yml
     - name: Build Docker Image
       run: |
-        docker-compose build
-    - name: Start Docker-Compose
+        docker compose build
+    - name: Start Docker Compose
       run: |
-        docker-compose up -d
+        docker compose up -d
         sleep 15 # let the DB warm up a little
     - name: Prepare Database
       run: |
-        docker-compose run --rm web bundle exec rails db:migrate
-        docker-compose run --rm web bundle exec rails i18n:js:export
-        docker-compose run --rm web bundle exec rails assets:precompile
-        docker-compose run --rm web osmosis --rx docker/null-island.osm.xml --wd host=db database=openstreetmap user=openstreetmap password=openstreetmap validateSchemaVersion=no
+        docker compose run --rm web bundle exec rails db:migrate
+        docker compose run --rm web bundle exec rails i18n:js:export
+        docker compose run --rm web bundle exec rails assets:precompile
+        docker compose run --rm web osmosis --rx docker/null-island.osm.xml --wd host=db database=openstreetmap user=openstreetmap password=openstreetmap validateSchemaVersion=no
     - name: Test Basic Website
       run: |
         curl -siL http://127.0.0.1:3000 | egrep '^HTTP/1.1 200 OK'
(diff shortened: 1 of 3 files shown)


gravitystorm:
I'm happy to merge this. Before I do, do you have any comments on how we can improve the opening section of that document? A the moment it reads:

> You will need to install Docker and Docker Compose on your development machine:
>
> - [Install Docker](https://docs.docker.com/install/)
> - [Install Docker Compose](https://docs.docker.com/compose/install/)

But I think this is now confusing, in particular the second link is full of warnings to not use this any more.

If I understand the situation correctly, the Docker install guide (first link) will encourage developers to install Docker Desktop (on all platforms) and that comes with compose v2 included. However, if a developer installs the Docker CLI on linux, they have the option of installing the docker-compose plugin. Should we ignore this second option, and simplify the instructions? 

If you have a suggestion for updating this section that would be very welcome.

Firefishy:
I'd only recommend installing Docker Desktop for Mac and Windows users. I would not feel comfortable pushing Docker Desktop onto Linux users. 

docker compose is available in the `docker-compose-plugin` package (DEB/RPM) on linux only if you are using the [ Docker Engine install method](https://docs.docker.com/engine/install/). `docker compose` is generally not (yet) available in base linux distributions.

gravitystorm:
@Firefishy thanks! So should it then be:

* Docker Destop (windows and mac)
* Docker Engine + docker-compose-plugin (linux)

... and then "docker compose foo" (i.e. no hyphen) will then work on all platforms?

gravitystorm (changes requested):
We want to keep the manual instructions and the CI docker tests in sync. So that means .github/workflows/docker.yml needs the same change from docker-compose to docker compose

Firefishy:
> @Firefishy thanks! So should it then be:
> 
> * Docker Destop (windows and mac)
> * Docker Engine + docker-compose-plugin (linux)
> 
> ... and then "docker compose foo" (i.e. no hyphen) will then work on all platforms?

Correct.

Firefishy:
> We want to keep the manual instructions and the CI docker tests in sync. So that means .github/workflows/docker.yml needs the same change from docker-compose to docker compose

FYI: Both `docker-compose` (v1) and `docker compose` (v2) come pre-installed in GitHub actions. For consistency GitHub Actions should use `docker compose` (v2).

tordans:
@gravitystorm sorry, I think I just force-pushed on top of your changes. I started working on this PR a few hours ago when I checked the open tasks in the description but only now pushed my changes.

My update includes an updated installation instructions as well the update for the workflow file.

Feel free to force-push your changes on top of this again!

gravitystorm:
> @gravitystorm sorry, I think I just force-pushed on top of your changes. I started working on this PR a few hours ago when I checked the open tasks in the description but only now pushed my changes.

No worries! There's nothing else in my suggested changes that aren't included in your version, so I've merged it now.

PR #4260, opened 2023-09-27
Switch to dartsass-sprockets for stylesheets

gravitystorm:
[`sassc-rails`](https://github.com/sass/sassc-rails) and the underlying C-library [libsass](https://github.com/sass/libsass), are deprecated. `libsass` has problems when compiling newer versions of Bootstrap (5.2+), where it cannot re-parse its own output, and that causes problems during our asset precompilation. After extensive investigations I cannot find a workaround to keep using sassc-rails with newer versions of Bootstrap.

This PR switches our sass stylesheet compiler to [`dartsass-sprockets`](https://github.com/tablecheck/dartsass-sprockets), which is designed to be a drop-in replacement for `sassc-rails`, and which uses the currently maintained [dart implementation of sass](https://sass-lang.com/dart-sass/). 

Unfortunately the bootstrap gem had, [until recently](https://github.com/twbs/bootstrap-rubygem/commit/fde1edff90d9219bbef48ecc6a65f13835269bcd), a hard-coded dependency on `sassc-rails`, and therefore conflicted with `dartsass-sprockets` (the two cannot be used at the same time). Although this has been resolved in bootstrap-rubygems main branch, the fix has not been backported to the 5.1 and 5.2 branches. To work around this, I have backported the fix myself, and this PR uses the bootstrap 5.1 backported version.

I have also prepared two other branches, which I will open draft PRs for, namely:

* Upgrading to bootstrap 5.2, with the appropriate backport
* Upgrading to bootstrap 5.3, using bootstrap-rubygem main branch (backport no longer required)

I'm offering these as a series of incremental upgrades, with all the associated backporting hassle, just in case there are any problems with deployment that CI hasn't picked up, as well as to make them easier to review.

A note on `sass-embedded` pinning - in recent years dartsass has added a number of deprecation warnings relating to changes in the sass language. Bootstrap quickly receives fixes to deal with them, but these fixes are generally not backported to earlier releases. To avoid copious noise from these deprecations, it's best to use a version of dartsass that is contemporary with each bootstrap release.  Each PR in this series does so. When we update to 5.3+ this will be less of an issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get update \
       curl \
       default-jre-headless \
       file \
+      git-core \
       gpg-agent \
       libarchive-dev \
       libffi-dev \
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,9 @@ gem "json"
 gem "pg"
 
 # Use SCSS for stylesheets
-gem "sassc-rails"
+gem "dartsass-sprockets"
+# Pin the dependentent sass-embedded to avoid deprecation warnings in bootstrap
+gem "sass-embedded", "~> 1.55.0"
 
 # Use Terser as compressor for JavaScript assets
 gem "terser"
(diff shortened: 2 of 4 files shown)


tomhughes:
I don't believe git based gems work in our production environment. I can't remember what the problem is but I'm sure we tried it once and had to rollback and since then I've always published to rubygems when I do a private fork of something.

tomhughes:
Well whatever the problem was before it doesn't seem to be an issue any more and I have now deployed this successfully.

PR #4262, opened 2023-09-27
Upgrade to Bootstrap 5.3

gravitystorm:
This builds on #4260 and #4261 and upgrades to bootstrap 5.3. 

This allows us to drop the use of backported patches to the bootstrap gem, and revert to upstream - although the fix we need is not yet in a released version of the gem.


gravitystorm:
I haven't heard any problems with the 5.2 deployment last week, so I think we're good to go with the 5.3 upgrade.

tomhughes:
The only thing that makes me a bit nervous is installing a gem from a third party git branch - is there an urgent reason to do this rather than wait for upstream to release the 5.3.2 version that it looks like they started on last week?

gravitystorm:
There's no urgent reason for the upgrade. I started this chain of PRs before the efforts to release 5.3.2 was started, but it might take a while to happen due to dependency issues - see https://github.com/twbs/bootstrap-rubygem/pull/256#issuecomment-1739922285 

So we can wait for an official release, or we can use their main branch and start working on various opportunities that 5.3 opens up. I'm happy either way. If we wait, hopefully we won't have to wait for too long, but that's unclear.

As soon as any gem release is available, we'll be going back to just using releases in any case - using the unreleased version would is only temporary.

tomhughes:
I tweaked this to use the specific revision rather than the `main` branch so that it won't changed on future bundle updates while we wait for a new upstream release and then merged that.

gravitystorm:
Great thanks! There's finally some movement today on the underlying dartsass stuff, so hopefully the bootstrap-rubygem will get a release soon too.

PR #4269, opened 2023-10-02
Remove <picture> from logo in header

AntonKhorev:
A `picture` element was added to show svg logo in https://github.com/openstreetmap/openstreetmap-website/commit/ecc4b64a23d0d27d5a1a2eb472082123cdf985cd. Then a `srcset` attribute was added in case `picture` is not supported in https://github.com/openstreetmap/openstreetmap-website/commit/37a4fc6dd14dfedc0c44a8d7bf6453df0db9e1b5. But why do you need both nowadays? Basic `` is supported by anything, so if some browser doesn't understand `srcset`, there's a fallback.

(I know that picture+srcset is used similarly in several other places. I'm not removing picture everywhere in case I'm wrong.)

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,10 +1,7 @@
 <header class="closed clearfix">
   <h1>
     <a href="<%= root_path %>" class="geolink">
-      <picture>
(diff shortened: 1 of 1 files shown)


tomhughes:
Presumably the same goes for the ones in `app/views/layouts/_flash.html.erb` where `srcset` was also added to the image inside the picture element?

PR #4272, opened 2023-10-03
Bootstrap pagination in changeset pages for nodes/ways/relations

AntonKhorev:
for sidebars of `/changeset/<id>` pages

when narrow enough:
image

when wider:
image

when really wide, with decreased padding:
image


tomhughes:
Ideally I'd like to get rid of `classic_pagination` at some point but until then modernizing the output is definitely good!

PR #4279, opened 2023-10-08
Add a primary key to the changeset_tags table

tomhughes:
This seems to be the only table that doesn't actually have a primary key defined in the database for some reason, instead we just lie to rails about it's existence.

It also improves our monkey patched primary key support for migrations to allow concurrent builds of primary keys and fix removal of primary keys which it seems has never worked.

### DIFF
--- a/config/initializers/migrate.rb
+++ b/config/initializers/migrate.rb
@@ -1,31 +1,46 @@
 if defined?(ActiveRecord::ConnectionAdapters::AbstractAdapter)
   module OpenStreetMap
-    module ActiveRecord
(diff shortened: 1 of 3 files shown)


mmd-osm:
I've noticed that `changesets_subscribers` also doesn't have a primary key. Is that standard Rails behaviour for n:m relationship tables?

tomhughes:
Probably, yes - those tables don't actually exist as rails models for sure which is why I didn't notice it when I was looking at things where we were overriding the primary key in the rails model.

Issue #3129, opened 2021-03-13
Add block function for individual users


grischard:
Users who receive abusive messages currently have to report the users to the admins. A recent case took too long to handle.

The website should make it possible for anyone to painlessly block messages from another user, the way you can block a pest on many social media sites.

If I block someone, it should stop them from sending me messages and commenting on my changesets and on my diary entries.

The UI should make it possible to block someone from the messages and from the user page. Once I've blocked someone, it should still be possible to report them, and of course to unblock them.

The admins should be able to see (and clear?) blocks for or by one user without having to mess with sql commands.

Admins and moderators should be able to send messages even when blocked.

tomhughes:
Including an admin interface seems like an unnecessary distraction for an initial version at least - it's not clear when it would be used or even if it should be used. I can't imagine a situation where it would be appropriate for an admin to remove a block that a user had chosen to add.

I would say that the minimum necessary is a table listing blocks, a couple of appropriate links/buttons where users can block/unblock people and code in the message send function to respect blocks.

grischard:
This might also be an opportunity to deal with #1946

simonpoole:
...
> If I block someone, it should stop them from .... and commenting on my changesets ...
...
Isn't providing such a facility a total non-starter? 

Note that I -do- support fine grain control of access to the messaging system and potentially being able to block messages from individual users, though IMHO in general somebody sending any kind of unacceptable messages should simply be denied access.

rory:
On Sat, 13 Mar 2021 17:43 +01:00, Simon Poole ***@***.***> wrote:
> > If I block someone, it should stop them from .... and commenting on my changesets ...
> Isn't providing such a facility a total non-starter?

If one person (B)  is harassing another (A), surely allowing B to comment on A's changesets is just allowing the harassment to continue?

However, I can understand the concern, because changeset discussion comments are (supposed to be) how we communicate about map edits, and the map itself. Perhaps, since changeset comments are public, people might not say things on it that they say in private messages, so maybe this isn't needed....

dieterdreist:
sent from a phone

> On 13 Mar 2021, at 18:19, Rory ***@***.***> wrote:
> 
> However, I can understand the concern, because changeset discussion comments are (supposed to be) how we communicate about map edits, and the map itself. Perhaps, since changeset comments are public, people might not say things on it that they say in private messages, so maybe this isn't needed


[...]

PR #4284, opened 2023-10-11
Add UserMute to control message visibility

grekko:

Implements the UserMute-Feature as described in https://github.com/osmfoundation/ewg_bidding/tree/main/projects/ability-to-mute-other-users/proposals/gregory_igelmund-08-09-2023

Addresses #3129

## What changed?

- Refactors `app/views/messages/*.html.erb` templates (extracting shared code)
- Adds `UserMute`-Model w/ CRUD operations to mute/unmute users
- Extends `Message`-Model with a `muted`-flag
- Adjusts `UserMailer.message_notification`-calls to respect the `Message#muted?`-flag

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -47,13 +47,14 @@ class Ability
         can [:show], :dashboard
         can [:new, :create, :edit, :update, :comment, :subscribe, :unsubscribe], DiaryEntry
         can [:make_friend, :remove_friend], Friendship
-        can [:new, :create, :reply, :show, :inbox, :outbox, :mark, :destroy], Message
+        can [:new, :create, :reply, :show, :inbox, :outbox, :muted, :mark, :unmute, :destroy], Message
         can [:close, :reopen], Note
         can [:show, :edit, :update], :preference
         can [:edit, :update], :profile
         can [:new, :create], Report
         can [:mine, :new, :create, :edit, :update, :destroy], Trace
         can [:account, :go_public], User
+        can [:index, :create, :destroy], UserMute
 
         if user.moderator?
           can [:hide, :unhide, :hidecomment, :unhidecomment], DiaryEntry
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -47,7 +47,7 @@ class MessagesController < ApplicationController
       render :action => "new"
     elsif @message.save
       flash[:notice] = t ".message_sent"
-      UserMailer.message_notification(@message).deliver_later
+      UserMailer.message_notification(@message).deliver_later if @message.notify_recipient?
       redirect_to :action => :inbox
     else
       @title = t "messages.new.title"
@@ -107,6 +107,13 @@ class MessagesController < ApplicationController
     @title = t ".title"
   end
 
+  # Display the list of muted messages received by the user.
+  def muted
+    @title = t ".title"
+
+    redirect_to inbox_messages_path if current_user.muted_messages.none?
+  end
+
   # Set the message as being read or unread.
   def mark
     @message = Message.where(:recipient => current_user).or(Message.where(:sender => current_user)).find(params[:message_id])
@@ -125,6 +132,23 @@ class MessagesController < ApplicationController
   rescue ActiveRecord::RecordNotFound
     @title = t "messages.no_such_message.title"
     render :action => "no_such_message", :status => :not_found
+  end
+
+  # Moves message into Inbox by unsetting the muted-flag
+  def unmute
+    message = current_user.muted_messages.find(params[:message_id])
+
+    if message.unmute
+      flash[:notice] = t(".notice")
+    else
+      flash[:error] = t(".error")
+    end
+
+    if current_user.muted_messages.none?
+      redirect_to inbox_messages_path
+    else
+      redirect_to muted_messages_path
+    end
   end
 
   private
--- a/app/controllers/user_mutes_controller.rb
+++ b/app/controllers/user_mutes_controller.rb
@@ -1,0 +1,45 @@
+class UserMutesController < ApplicationController
+  include UserMethods
+
+  layout "site"
+
+  before_action :authorize_web
+  before_action :set_locale
+
+  authorize_resource
+
+  before_action :lookup_user, :only => [:create, :destroy]
+  before_action :check_database_readable
+  before_action :check_database_writable, :only => [:create, :destroy]
+
+  def index
+    @muted_users = current_user.muted_users
+    @title = t ".title"
+
+    redirect_to edit_account_path unless @muted_users.any?
+  end
+
+  def create
+    user_mute = current_user.mutes.build(:subject => @user)
+
+    if user_mute.save
+      flash[:notice] = t(".notice", :name => user_mute.subject.display_name)
+    else
+      flash[:error] = t(".error", :name => user_mute.subject.display_name, :full_message => user_mute.errors.full_messages.to_sentence.humanize)
+    end
+
+    redirect_back_or_to user_mutes_path(current_user)
+  end
+
+  def destroy
+    user_mute = current_user.mutes.find_by!(:subject => @user)
+
+    if user_mute.destroy
+      flash[:notice] = t(".notice", :name => user_mute.subject.display_name)
+    else
+      flash[:error] = t(".error")
+    end
+
+    redirect_back_or_to user_mutes_path(current_user)
+  end
+end
--- a/app/models/user_mute.rb
+++ b/app/models/user_mute.rb
@@ -1,0 +1,34 @@
+# == Schema Information
+#
+# Table name: user_mutes
+#
+#  id         :bigint(8)        not null, primary key
+#  owner_id   :bigint(8)        not null
+#  subject_id :bigint(8)        not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+# Indexes
+#
+#  index_user_mutes_on_owner_id_and_subject_id  (owner_id,subject_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (owner_id => users.id)
+#  fk_rails_...  (subject_id => users.id)
+#
+class UserMute < ApplicationRecord
+  belongs_to :owner, :class_name => "User"
+  belongs_to :subject, :class_name => "User"
+
+  validates :subject, :uniqueness => { :scope => :owner_id, :message => :is_already_muted }
+
+  def self.active?(owner:, subject:)
+    !subject.administrator? &&
+      !subject.moderator? &&
+      exists?(
+        :owner => owner,
+        :subject => subject
+      )
+  end
+end
--- a/app/views/application/_settings_menu.html.erb
+++ b/app/views/application/_settings_menu.html.erb
@@ -14,5 +14,10 @@
     <li class="nav-item">
       <%= link_to t(".oauth2_authorizations"), oauth_authorized_applications_path, :class => "nav-link #{'active' if controller_name == 'oauth2_authorized_applications'}" %>
     </li>
+    <% if current_user.muted_users.any? %>
+      <li class="nav-item">
+        <%= link_to t(".muted_users"), user_mutes_path, :class => "nav-link #{'active' if controller_name == 'user_mutes'}" %>
+      </li>
+    <% end %>
   </ul>
 <% end %>
--- a/app/views/messages/_messages_table.html.erb
+++ b/app/views/messages/_messages_table.html.erb
@@ -1,0 +1,13 @@
+<table class="table table-sm align-middle">
+  <thead>
+    <tr>
+      <% columns.each do |column| %>
+      <th><%= t ".#{column}" %></th>
+      <% end %>
+      <th class="d-flex justify-content-end"><%= t ".actions" %></th>
+    </tr>
+  </thead>
+  <tbody>
+    <%= render :partial => inner_partial, :collection => messages %>
+  </tbody>
+</table>
--- a/app/views/user_mutes/index.html.erb
+++ b/app/views/user_mutes/index.html.erb
@@ -1,0 +1,38 @@
+<% content_for :heading do %>
+  <h1><%= t ".my_muted_users" %></h1>
+<% end %>
+
+<%= render :partial => "settings_menu" %>
+
+<h4>
+  <%= t ".you_have_muted_n_users", :count => @muted_users.size %>
+</h4>
+<p>
+  <%= t ".user_mute_explainer" %>
+  <em><%= t ".user_mute_admins_and_moderators" %></em>
+</p>
+
+<% if @muted_users.any? %>
+  <table class="table">
+    <thead>
+      <tr>
+        <th><%= t ".table.thead.muted_user" %></th>
+        <th class="d-flex justify-content-end"><%= t ".table.thead.actions" %></th>
+      </tr>
+    </thead>
+    <tbody>
+      <% @muted_users.each do |user| %>
+        <tr>
+          <td>
+            <%= user_thumbnail_tiny user %>
+            <%= link_to user.display_name, user_path(user) %>
+          </td>
+          <td class="text-nowrap text-end">
+            <%= link_to t(".table.tbody.unmute"), user_mute_path(user), :method => :delete, :class => "btn btn-sm btn-primary" %>
+            <%= link_to t(".table.tbody.send_message"), new_message_path(user), :class => "btn btn-sm btn-secondary" %>
+          </td>
+        </tr>
+      <% end %>
+    </tbody>
+  </table>
+<% end %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1682,11 +1682,14 @@ en:
       old_messages:
         one: "%{count} old message"
         other: "%{count} old messages"
-      from: "From"
-      subject: "Subject"
-      date: "Date"
       no_messages_yet_html: "You have no messages yet. Why not get in touch with some of the %{people_mapping_nearby_link}?"
       people_mapping_nearby: "people mapping nearby"
+    messages_table:
+      from: "From"
+      to: "To"
+      subject: "Subject"
+      date: "Date"
+      actions: "Actions"
     message_summary:
       unread_button: "Mark as unread"
       read_button: "Mark as read"
@@ -1707,12 +1710,10 @@ en:
       title: "Outbox"
       my_inbox: "My Inbox"
       my_outbox: "My Outbox"
+      actions: "Actions"
       messages:
         one: "You have %{count} sent message"
         other: "You have %{count} sent messages"
-      to: "To"
-      subject: "Subject"
-      date: "Date"
       no_sent_messages_html: "You have no sent messages yet. Why not get in touch with some of the %{people_mapping_nearby_link}?"
       people_mapping_nearby: "people mapping nearby"
     reply:
@@ -1726,6 +1727,9 @@ en:
       wrong_user: "You are logged in as `%{user}' but the message you have asked to read was not sent by or to that user. Please login as the correct user in order to read it."
     sent_message_summary:
       destroy_button: "Delete"
+    heading:
+      my_inbox: "My Inbox"
+      my_outbox: "My Outbox"
     mark:
       as_read: "Message marked as read"
       as_unread: "Message marked as unread"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -40,6 +40,12 @@ en:
       messages:
         invalid_email_address: does not appear to be a valid e-mail address
         email_address_not_routable: is not routable
+      models:
+        user_mute:
+          attributes:
+            subject:
+              format: "%{message}"
+          is_already_muted: "is already muted"
     # Translates all the model names, which is used in error handling on the website
     models:
       acl: "Access Control List"
@@ -1673,8 +1679,6 @@ en:
   messages:
     inbox:
       title: "Inbox"
-      my_inbox: "My Inbox"
-      my_outbox: "My Outbox"
       messages: "You have %{new_messages} and %{old_messages}"
       new_messages:
         one: "%{count} new message"
@@ -1695,6 +1699,7 @@ en:
       read_button: "Mark as read"
       reply_button: "Reply"
       destroy_button: "Delete"
+      unmute_button: "Move to Inbox"
     new:
       title: "Send message"
       send_message_to_html: "Send a new message to %{name}"
@@ -1708,14 +1713,17 @@ en:
       body: "Sorry there is no message with that id."
     outbox:
       title: "Outbox"
-      my_inbox: "My Inbox"
-      my_outbox: "My Outbox"
       actions: "Actions"
       messages:
         one: "You have %{count} sent message"
         other: "You have %{count} sent messages"
       no_sent_messages_html: "You have no sent messages yet. Why not get in touch with some of the %{people_mapping_nearby_link}?"
       people_mapping_nearby: "people mapping nearby"
+    muted:
+      title: "Muted Messages"
+      messages:
+        one: "%{count} muted message"
+        other: "You have %{count} muted messages"
     reply:
       wrong_user: "You are logged in as `%{user}' but the message you have asked to reply to was not sent to that user. Please login as the correct user in order to reply."
     show:
@@ -1727,12 +1735,16 @@ en:
       wrong_user: "You are logged in as `%{user}' but the message you have asked to read was not sent by or to that user. Please login as the correct user in order to read it."
     sent_message_summary:
       destroy_button: "Delete"
-    heading:
+    heading: 
       my_inbox: "My Inbox"
       my_outbox: "My Outbox"
+      muted_messages: "Muted messages"
     mark:
       as_read: "Message marked as read"
       as_unread: "Message marked as unread"
+    unmute:
+      notice: "Message has been moved to Inbox"
+      error: "The message could not be moved to the Inbox."
     destroy:
       destroyed: "Message deleted"
   passwords:
@@ -2562,6 +2574,7 @@ en:
       oauth1_settings: OAuth 1 settings
       oauth2_applications: OAuth 2 applications
       oauth2_authorizations: OAuth 2 authorizations
+      muted_users: Muted Users
   oauth:
     authorize:
       title: "Authorize access to your account"
@@ -2749,6 +2762,8 @@ en:
       my_dashboard: My Dashboard
       blocks on me: Blocks on Me
       blocks by me: Blocks by Me
+      create_mute: Mute this User
+      destroy_mute: Unmute this User
       edit_profile: Edit Profile
       send message: Send Message
       diary: Diary
@@ -2939,6 +2954,29 @@ en:
       showing_page: "Page %{page}"
       next: "Next »"
       previous: "« Previous"
+  user_mutes:
+    index:
+      title: "Muted Users"
+      my_muted_users: "My muted users"
+      you_have_muted_n_users:
+        one: "You have muted %{count} User"
+        other: "You have muted %{count} users"
+      user_mute_explainer: "Messages of muted users are moved into a separate Inbox and you won't receive email notifications."
+      user_mute_admins_and_moderators: "You can mute Admins and Moderators but their messages will not be muted."
+      table:
+        thead:
+          muted_user: "Muted User"
+          actions: "Actions"
+        tbody:
+          unmute: "Unmute"
+          send_message: "Send message"
+
+    create:
+      notice: "You muted %{name}."
+      error: "%{name} could not be muted. %{full_message}."
+    destroy:
+      notice: "You unmuted %{name}."
+      error: "User could not be unmuted. Please try again."
   notes:
     index:
       title: "Notes submitted or commented on by %{user}"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -282,9 +282,12 @@ OpenStreetMap::Application.routes.draw do
   # messages
   resources :messages, :only => [:create, :show, :destroy] do
     post :mark
+    patch :unmute
+
     match :reply, :via => [:get, :post]
     collection do
       get :inbox
+      get :muted
       get :outbox
     end
   end
@@ -292,6 +295,12 @@ OpenStreetMap::Application.routes.draw do
   get "/user/:display_name/outbox", :to => redirect(:path => "/messages/outbox")
   get "/message/new/:display_name" => "messages#new", :as => "new_message"
   get "/message/read/:message_id", :to => redirect(:path => "/messages/%{message_id}")
+
+  # muting users
+  scope "/user/:display_name" do
+    resource :user_mute, :only => [:create, :destroy], :path => "mute"
+  end
+  resources :user_mutes, :only => [:index]
 
   # oauth admin pages (i.e: for setting up new clients, etc...)
   scope "/user/:display_name" do
--- a/db/migrate/20231010201451_create_user_mutes.rb
+++ b/db/migrate/20231010201451_create_user_mutes.rb
@@ -1,0 +1,15 @@
+class CreateUserMutes < ActiveRecord::Migration[7.0]
+  def change
+    create_table :user_mutes do |t|
+      t.references :owner, :null => false, :index => false
+      t.references :subject, :null => false, :index => false
+
+      t.timestamps
+
+      t.foreign_key :users, :column => :owner_id
+      t.foreign_key :users, :column => :subject_id
+
+      t.index [:owner_id, :subject_id], :unique => true
+    end
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -951,7 +951,8 @@ CREATE TABLE public.messages (
     to_user_id bigint NOT NULL,
     to_user_visible boolean DEFAULT true NOT NULL,
     from_user_visible boolean DEFAULT true NOT NULL,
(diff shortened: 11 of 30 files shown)


db/structure.sql
```diff
@@ -3510,3 +3597,5 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('7'),
 ('8'),
 ('9');
+
+
```
grekko, line 3601:
```suggestion
```

config/locales/en.yml
```diff
@@ -2900,6 +2912,23 @@ en:
...
       next: "Next »"
       previous: "« Previous"
+  user_mutes:
+    index:
+      you_have_muted_n_users:
+        zero: "You haven't muted any Users 🎉"
+        one: "You have muted %{count} User"
+        other: "You have muted %{count} Users"
+      user_mute_explainer: "Private messages of muted users are moved into a separate inbox and you won't receive email notifications."
+      user_mute_admins_and_moderators: "You can mute Admins and Moderators but their private messages cannot."
```
grekko, line 2922:
```suggestion
      user_mute_admins_and_moderators: "You can mute Admins and Moderators but their private messages cannot be muted."
```

config/locales/en.yml
```diff
@@ -2900,6 +2912,23 @@ en:
...
+      unmute: "Unmute"
+      send_message: "Send message"
+      title: "Muted Users"
+    create:
+      notice: "You muted %{name}."
+      error: "Something went wrong."
+    destroy:
+      notice: "You unmuted %{name}."
+      error: "Something went wrong."
```
grekko, line 2931: Either use controller action specific messages for the errors or reuse generic messages.

grekko:
Hey @gravitystorm 👋🏽

this is my take on implementing the `UserMute`-Feature. There are still things I currently consider to be missing and some things I would like to put up for discussion.

Before going into these I wanted to ask you to have a peek at the current shape of the code and let me know if you notice some issues with the fundamental structures though before I'd like into the details.

---

What I think is missing or I want to look more closely at:
- I am not super happy with the `before_save`-Callback in `Message` which calls out to `UserMute.active_for?` but haven't taken the time yet to think about a solid alternative that I'd prefer here.
- Add tests for new/changed controllers
- There are some system and integration tests. I might add an integration test since no JS is involved and the integration test would imo be appropriate here.

---

Open questions from the top of my head:

- I think the most natural place where users would expect the "UserMute"-action would be the `Messages#show`-screen, but I could not come up with a fitting UI-solution. Currently the "UserMute"-action lives just on the `Users#show`-screen. If it would live in `Messages#show`:
  - Where should this thing be placed? I'd expect it to be close to the Displayname/User image, not underneath the image, but I don't it to be rendered more prominent than necessary
  - Should it be a link "Mute this user" or a button?

- Users can move muted Messages from their `Muted messages`-Tab on `/messages/muted` back into the regular `My Inbox`-Tab. Should they be also able – at this point – to move Messages from the `My Inbox`-Tab to the `Muted messages`-Tag?

- Do we want to encourage users who have muted other users to remove the `UserMute`-again to prevent an accumulation of these mutes or should we wait how this feature plays out – and if it is even used? I am a fan of the latter.

app/models/user_mute.rb
```diff
@@ -0,0 +1,45 @@
...
+#  index_user_mutes_on_creator_id_and_appointee_id  (creator_id,appointee_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (appointee_id => users.id)
+#  fk_rails_...  (creator_id => users.id)
+#
+class UserMute < ApplicationRecord
+  belongs_to :creator, :class_name => "User"
+  belongs_to :appointee, :class_name => "User"
```
tomhughes, line 24: I think `appointee` is a pretty odd term to use for the muted user.

I mean I could argue for `mutor` and `mutee` for the creator and person being muted but that might be taking english a bit far.

Maybe `subject` for the person being muted?

On the other side `creator` isn't terrible but maybe `owner` is better? Not sure either really captures the concept well though.

app/models/user_mute.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    !sender.administrator? &&
+      !sender.moderator? &&
+      active_for?(
+        :current_user => message.recipient,
+        :other_user => sender
+      )
+  end
+
+  def self.active_for?(current_user:, other_user:)
```
tomhughes, line 39: Why use different names for the users in the argument list here? It's the same concept as the member names so why not use the same names?

Of course once you do that the question becomes is this method even needed when it's just forwarding the arguments to `exists?`?

config/locales/en.yml
tomhughes: That change is of course equivalent to saying "you can mute them but it has no effect" so should we just not allow them to be muted?

I guess if their status changes later then the mute would suddenly activate but then that raises more questions...

db/migrate/20231010201451_create_user_mutes.rb
```diff
@@ -0,0 +1,15 @@
+class CreateUserMutes < ActiveRecord::Migration[7.0]
+  def change
+    create_table :user_mutes do |t|
+      t.references :creator, :null => false
```
tomhughes, line 4: I think we can add `:index => false` here as we have a separate index on creator+appointee for uniqueness so also having one on creator is unnecessary.

tomhughes (commented):
Just a few initial comments, not a detailed review.

config/locales/en.yml
grekko: > That change is of course equivalent to saying "you can mute them but it has no effect" so should we just not allow them to be muted?

right. Users are able to mute Admins/Moderators but muting doesn't have an effect.

app/models/user_mute.rb
grekko: > Maybe `subject` for the person being muted?

I like `subject` much better!

I initially used `target` but didn't like it and went with `appointee` 😬 `subject` is better. I'll adjust it.

grekko: > … maybe `owner` is better?

I went with `creator` b/c `creator` is used for `UserBlock`-records but `owner` is imho also expressive and being used e.g. in the concept of `OAuth`-associated records.

I am not having feelings strong enough towards 1 or the other and am happy to go with `owner`. I'll adjust this.

app/models/user_mute.rb
grekko: > Of course once you do that the question becomes is this method even needed when it's just forwarding the arguments to exists??

good point. Initially there were more places from which `UserMute.active_for?` in which the the terms `current_user` and `other_user` are used to reference the instances. So I thought it was making the code a bit easier to read/grasp from that context.

Now there is just a single call, plus the call from `UserMute#for_message?` – I think it's ok to drop `UserMute#active_for?` and just call `UserMute#exists?` directly 👌🏽

config/locales/en.yml
grekko: > so should we just not allow them to be muted?

🤦🏽 I rushed over your text. You mean that we could just not display the `Mute this user`-action!

> I guess if their status changes later then the mute would suddenly activate but then that raises more questions...

Exactly this.

grekko:
@gravitystorm ping. Could you have a look at the general direction of the code in this PR? Do you have already some notes for me?

app/controllers/user_mutes_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      :subject => @user
+    )
+
+    if user_mute.save
+      flash[:notice] = t(".notice", :name => user_mute.subject.display_name)
+    else
+      flash[:error] = t(".error")
+    end
+
+    redirect_back :fallback_location => user_mutes_path(current_user)
```
tomhughes, line 32: It seems `redirect_back_or_to` is now preferred over `redirect_back` (that might be new in rails 7.1) though both work for now.

app/views/user_mutes/index.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    <tbody>
+      <% @muted_users.each do |user| %>
+        <tr>
+          <td>
+            <%= user_thumbnail_tiny user %>
+            <%= link_to user.display_name, user_path(user) %>
+          </td>
+          <td class="text-nowrap text-end">
+            <%= link_to t(".unmute"), unmute_user_path(user), :method => :delete, :class => "btn btn-sm btn-primary" %>
+            <%= link_to t(".send_message"), new_message_path(user), :class => "btn btn-sm btn-secondary" %>
```
tomhughes, line 30: An option to send a message to a muted user seems a bit strange when by definition any reply will wind up in the muted messages and won't be notified?

config/locales/en.yml
```diff
@@ -1724,9 +1729,16 @@ en:
       wrong_user: "You are logged in as `%{user}' but the message you have asked to read was not sent by or to that user. Please login as the correct user in order to read it."
     sent_message_summary:
       destroy_button: "Delete"
+    heading: 
+      my_inbox: "My Inbox"
+      my_outbox: "My Outbox"
+      muted_messages: "Muted messages"
```
tomhughes, line 1735: Should the capitalisation be consistent here given these are three labels that sit adjacent to each other?

config/locales/en.yml
```diff
@@ -2900,6 +2915,23 @@ en:
       showing_page: "Page %{page}"
       next: "Next »"
       previous: "« Previous"
+  user_mutes:
+    index:
+      you_have_muted_n_users:
+        zero: "You haven't muted any Users 🎉"
+        one: "You have muted %{count} User"
+        other: "You have muted %{count} Users"
```
tomhughes, line 2923: Why are we capitalising User/Users here?

tomhughes (changes requested):
In addition to the specific comments I've made, a few general observations/questions...

Firstly I think we should have a "mute user" button on the detail view (messages#show) for messages to make it more obvious that the option is available when a problem arises? Probably we should also have report user (or ideally report message) but that's not really in scope for this project.

Secondly should we have a mute/unmute button on the main message view - currently messages can only be unmuted from the index page and there's no way to mute a message at all if the user is not muted when it is received.

app/controllers/messages_controller.rb
```diff
@@ -127,6 +134,23 @@ def mark
     render :action => "no_such_message", :status => :not_found
   end
 
+  # Moves message into Inbox by unsetting the muted-flag
+  def unmute
+    message = Message.where(:recipient => current_user).find(params[:message_id])
```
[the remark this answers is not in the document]
gravitystorm: We prefer `find_by` instead of `where().first`, but `Model.find(id)` is still preferred to `Model.find_by(:id => id)`.

I think some kind of scoping is necessary to avoid unmuting other people's messages, although I'd normally try to find a `current_user.something.find()` to scope them rather than building a where query.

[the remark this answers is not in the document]
grekko: > This never actually changes the state of the message, so nothing happens when you click the Unmute button for a message in the user interface.

this is just the query for the `message`-record. 2 lines below `message.unmute` is called.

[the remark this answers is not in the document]
grekko: > I think some kind of scoping is necessary to avoid unmuting other people's messages

The query is already scoped to the `current_user`

```rb
Message.where(:recipient => current_user).find(params[:message_id])
```

But yes, I prefer a scope on the `current_user`, too. I'll change this too:

```rb
current_user.muted_messages.find(params[:message_id])
```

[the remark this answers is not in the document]
tomhughes: > this is just the query for the `message`-record. 2 lines below `message.unmute` is called.

Oh I missed that... In that case the question becomes why wasn't it working for me because the reason I stared looking closely at that code was that the unmute didn't seem to work.

config/locales/en.yml
grekko: Because all actions containing the keyword `User` on `Users#show` are currently capitalised.

SCR-20231023-lnud

grekko: ah. I understand that here in this context "Users" should not be capitalised since it's not referencing a language subject. I'll adjust the translation.

app/views/user_mutes/index.html.erb
grekko: > An option to send a message to a muted user seems a bit strange when by definition any reply will wind up in the muted messages and won't be notified?

I can imagine situations in which I muted another user just to prevent them from spamming my email inbox but still want to keep the conversation open.

config/locales/en.yml
grekko: To me the 3 label terms here are different. My reasoning:

- `Inbox` and `Outbox` are defined terms and therefore should be capitalised.
- `Muted messages` is not.

Bear in mind that I am not a native english speaker so I'm happy to adjust the label to `Muted Messages if you say so.

grekko:
Regarding the general feedback from @tomhughes 

> Firstly I think we should have a "mute user" button on the detail view (messages#show) for messages to make it more obvious that the option is available when a problem arises? […]

I think so too, but I haven't found the ideal place yet where I want to put that `mute user`-action.

This is the `Messages#show`-screen:

SCR-20231023-mdgi

The action should be imo close to the Username (at the top) and not at the bottom since all actions on the bottom of the message are message-related actions.

There is not much space at the top of the message. We could re-use the layout we already use on `Dashboards#show`.

SCR-20231023-njqe


The updated `Messages#show`-screen could look like this then:

SCR-20231023-nivc

grekko:
> Secondly should we have a mute/unmute button on the main message view - currently messages can only be unmuted from the index page and there's no way to mute a message at all if the user is not muted when it is received.

I understand your point. I imagine though that it would add unnecessary noise to the UI for many users when we add a `Mute` or `Move to muted Messages`-button next to each `Inbox`-message:

SCR-20231023-nnbv

vs.

SCR-20231023-nnyk

In my understanding only a fraction of users – lets say 5% for the sake of my argument – will have to use the `UserMute`-Feature on rare occasions.

Adding the `Mute`-action to all Inbox-messages for all users makes the `UserMute`-Feature a bit too prominent since 95% of the users never have to bother with the Feature.

What do you think? @tomhughes

app/controllers/messages_controller.rb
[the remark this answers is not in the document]
grekko: > In that case the question becomes why wasn't it working for me

not sure what went wrong there :)

gravitystorm (commented):
I'm sorry I haven't had the opportunity to do a full review - this is my initial thoughts from reading the code, without trying out the interfaces. Happy to discuss any of these points.

app/controllers/messages_controller.rb
```diff
@@ -47,7 +47,7 @@ def create
       render :action => "new"
     elsif @message.save
       flash[:notice] = t ".message_sent"
-      UserMailer.message_notification(@message).deliver_later
+      @message.notify_recipient_via_mail
```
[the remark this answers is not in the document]
grekko: > I think it should be the controller that decides what is the list of things that should be done.

I think that is still the case. The controller still makes the call, but now there is this extra logic which is imo a detail that the controller does not need to bother about.

I understand though that currently all direct calls to the `UserMailer` are made either from within the context of a Controller or Job and my change deviates from that pattern.

To me it is important that the logic wether a E-Mail Notification should be sent is kept in a single place and is not spread across the application code.

What do you think about this version?

```diff
diff --git a/app/controllers/messages_controller.rb b/app/controllers/messages_controller.rb
index 21d08e7af..2ca86fc02 100644
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -47,7 +47,7 @@ class MessagesController < ApplicationController
       render :action => "new"
     elsif @message.save
       flash[:notice] = t ".message_sent"
-      @message.notify_recipient_via_mail
+      UserMailer.message_notification(@message).deliver_later if @message.notify_recipient?
       redirect_to :action => :inbox
     else
       @title = t "messages.new.title"
diff --git a/app/models/message.rb b/app/models/message.rb
index 1be482cfb..fdeb89c4b 100644
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -74,10 +74,8 @@ class Message < ApplicationRecord
     md5.hexdigest
   end
 
-  def notify_recipient_via_mail
-    return false if muted?
-
-    UserMailer.message_notification(self).deliver_later
+  def notify_recipient?
+    !muted?
   end
 
   def unmute
diff --git a/script/deliver-message b/script/deliver-message
index a382bffa1..496773597 100755
--- a/script/deliver-message
+++ b/script/deliver-message
@@ -33,6 +33,6 @@ mail = Mail.new($stdin.read
 message = Message.from_mail(mail, from, to)
 message.save!
 
-message.notify_recipient_via_mail
+UserMailer.message_notification(message).deliver_later if message.notify_recipient?
 
 exit 0
```

app/controllers/messages_controller.rb
```diff
@@ -107,6 +107,13 @@ def outbox
     @title = t ".title"
   end
 
+  # Display the list of muted messages received by the user.
+  def muted
+    @title = t ".title"
+
+    redirect_to inbox_messages_path if current_user.muted_messages.none?
```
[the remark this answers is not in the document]
grekko: The redirect is meant to handle the situation in which a user moves the last muted message back into the Inbox, like so:



Once there are no muted Messages anymore, the `Muted`-Tab is hidden and the user is redirected back to the Inbox.

Rendering the template instead would create an inconsistent UI because no Tab would be active. Or we would need to display the `Muted`-Tab even though there are no messages and then hide it once the user clicks on any other link.

Why would you prefer rendering the template?

app/models/user_mute.rb
```diff
@@ -0,0 +1,36 @@
...
+#  fk_rails_...  (owner_id => users.id)
+#  fk_rails_...  (subject_id => users.id)
+#
+class UserMute < ApplicationRecord
+  belongs_to :owner, :class_name => "User"
+  belongs_to :subject, :class_name => "User"
+
+  validates :subject, :uniqueness => { :scope => :owner_id, :message => :is_already_muted }
+
+  def self.for_message?(message)
```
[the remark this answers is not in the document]
grekko: > […] the message object isn't being used for anything here, other than a sort of container for the two users.

To me having the method `.for_message?` expresses that the `UserMute` has this special relation to messages. When reading this I'd think that `UserMute` is designed to be working only in conjunction with messages. I'm aware that this is my subjective interpretation though and I not strongly for this specific implementation.

You were thinking about something like this?

```diff
diff --git a/app/models/message.rb b/app/models/message.rb
index 1be482cfb..bfb9c6a75 100644
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -87,6 +87,6 @@ class Message < ApplicationRecord
   private
 
   def set_muted
-    self.muted ||= UserMute.for_message?(self)
+    self.muted ||= UserMute.active?(owner: recipient, subject: sender)
   end
 end
diff --git a/app/models/user_mute.rb b/app/models/user_mute.rb
index e0b80d77a..311ca0bf7 100644
--- a/app/models/user_mute.rb
+++ b/app/models/user_mute.rb
@@ -23,13 +23,11 @@ class UserMute < ApplicationRecord
 
   validates :subject, :uniqueness => { :scope => :owner_id, :message => :is_already_muted }
 
-  def self.for_message?(message)
-    sender = message.sender
-
+  def self.active?(owner:, subject:)
     !sender.administrator? &&
       !sender.moderator? &&
       exists?(
-        :owner => message.recipient,
+        :owner => owner,
         :subject => sender
       )
   end
```

test/models/user_mute_test.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def test_messages_by_admins_or_moderators_are_never_muted
+    user = create(:user)
+
+    [create(:administrator_user), create(:moderator_user)].each do |admin_or_moderator|
+      create(:user_mute, :owner => user, :subject => admin_or_moderator)
+
+      message = build(:message, :sender => admin_or_moderator, :recipient => user)
+      message.save
```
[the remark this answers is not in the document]
grekko: 🤔 yeah, I would expect so. I am not sure the code calls `build` and `save` – I think this can and should just be `create` – thanks!

app/abilities/ability.rb
```diff
@@ -47,13 +47,14 @@ def initialize(user)
         can [:show], :dashboard
         can [:new, :create, :edit, :update, :comment, :subscribe, :unsubscribe], DiaryEntry
         can [:make_friend, :remove_friend], Friendship
-        can [:new, :create, :reply, :show, :inbox, :outbox, :mark, :destroy], Message
+        can [:new, :create, :reply, :show, :inbox, :outbox, :muted, :mark, :unmute, :destroy], Message
```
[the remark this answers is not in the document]
grekko: Thanks for bringing this up. I read the article and the described patterns and ideas sounds reasonable at first glance but I have many open questions on how I'd implement specific current non-CRUD operations in a CRUD-style.

Splitting up the `Messages#inbox`, `Messages#outbox` and `Messages#muted` actions into separate controllers, e.g. the `InboxController#index` you already mentioned seems straight forward to me.

Although I'd probably namespace those controllers e.g. like to `Messages::InboxController#index` just to make it very clear that this is about a Message Inbox and not any other Inbox.

Also I would say that `Messages#reply` could be `Messages::Replies#create` instead – in the spirit of keeping things CRUD-oriented – although I could be reading this as "This controller action creates a `Message::Reply`-record, although it actually creates just another `Message`-record 🤔 

---

The ideas of keeping things more uniform and simple resonate with me, but I'd need more time to fiddle around with the ideas and how to apply them without creating "just more code and files" without getting the benefit of more organised and easier to grasp code and files.

I must admit: Thats something I would not feel comfortable with experimenting in the scope of this PR :)

test/models/user_mute_test.rb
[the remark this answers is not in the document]
grekko: I updated the original commit and force pushed the changes. => resolving this conversation.

app/abilities/ability.rb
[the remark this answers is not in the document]
gravitystorm: > I must admit: Thats something I would not feel comfortable with experimenting in the scope of this PR :)

Sure, no worries! I just thought I would ask.

tordans:
@grekko Does this already handle the API part? We have this part here https://wiki.openstreetmap.org/wiki/API_v0.6#Details_of_the_logged-in_user:_GET_/api/0.6/user/details

I think, the easiest would be to change the queries for `<received count="1" unread="0"/>` to only include messages by unmuted users. However, I wonder if we should also add the full count just to give the API access to all the parts that the UI has as well. Something like `<received count="1" unread="0"/><received_mute count="1" unread="0"/>`

I would both not consider this a breaking change of the API. Changing the way the data is caculated is in line with how the new UI behaves. And adding something does not break (a well build api consumer).

(Disclaimer: I only did a quick scan of the files-list to check if this was solved already, sorry if it was…)

grekko:
> Does this already handle the API part? We have this part here

The code for the user serialisation is here https://github.com/openstreetmap/openstreetmap-website/blob/0844024c8e68ae0031e143d6094793f0aa32bc22/app/views/api/users/_user.json.jbuilder#L59-L63

The `muted` Messages are excluded here and not mentioned explicitly, so right know it behaves identical to the UI.

If API consumers want/need access to the `muted` Message count, one could add it the the serialised user object, but I didn't add it without having a use case here.

app/models/user_mute.rb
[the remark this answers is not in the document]
grekko: I'll apply the patch and will ask you for a re-review.

app/controllers/messages_controller.rb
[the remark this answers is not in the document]
grekko: I'll apply the patch and will ask you for a re-review.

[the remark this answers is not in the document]
grekko: The patch was applied => Closing the conversation (for now).

app/models/user_mute.rb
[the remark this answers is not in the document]
grekko: The patch was applied => Closing the conversation (for now).

grekko:
Hey @gravitystorm and/or @tomhughes 👋🏽 

Any chance to get a PR review with comments or Follow-Ups on the open PR conversations?
I would really like to get the PR in a mergeable shape.

app/views/messages/_messages_table.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <thead>
+    <tr>
+      <th><%= t ".from" %></th>
+      <th><%= t ".subject" %></th>
+      <th><%= t ".date" %></th>
+      <th class="d-flex justify-content-end"><%= t ".actions" %></th>
+    </tr>
+  </thead>
+  <tbody>
+    <%= render :partial => inner_partial, :collection => messages %>
```
gravitystorm, line 11: There is a level of indirection here with isn't used, since "inner_partial" is always equal to "message_summary". I guess what might have happened is that you started looking at the two different contents for this table - inbox and outbox being slightly different - but you haven't actually refactored the outbox, perhaps because the headings are different too.

So I think either more refactoring is required (so that the inner_partial is useful) or this level of indirection can be removed.

config/locales/en.yml
```diff
@@ -2906,6 +2927,30 @@ en:
       showing_page: "Page %{page}"
       next: "Next »"
       previous: "« Previous"
+  user_mutes:
+    index:
+      title: "Muted Users"
+      my_muted_users: "My muted users"
+      you_have_muted_no_users: "You have not muted any users 🎉"
```
gravitystorm, line 2934: No celebration emoji here please! I'm no big fan of emojis (they are an i18n nightmare), but this one should be used when something good happens to you, or you've taken a successful action, so it doesn't really fit the context of "nothing bad happened yet".

config/locales/en.yml
```diff
@@ -2906,6 +2927,30 @@ en:
...
+      you_have_muted_n_users:
+        one: "You have muted %{count} User"
+        other: "You have muted %{count} users"
+      user_mute_explainer: "Private messages of muted users are moved into a separate inbox and you won't receive email notifications."
```
gravitystorm, line 2938: We don't use the term "private messages" anywhere else in the UI, only "messages". So this could be confusing if someone is trying to work out if "private messages" are different from regular messages.

I think "private" can simply be removed in these two translations.

config/routes.rb
```diff
@@ -289,6 +292,11 @@
   get "/message/new/:display_name" => "messages#new", :as => "new_message"
   get "/message/read/:message_id", :to => redirect(:path => "/messages/%{message_id}")
 
+  # muting users
+  post "/user/:display_name/mute" => "user_mutes#create", :as => :mute_user
+  delete "/user/:display_name/mute" => "user_mutes#destroy", :as => :unmute_user
```
gravitystorm, line 297: If these can be added as resourceful routes then that would be better. Or at lest, specific actions on a resource or scoped resource. Something like (untested)

```ruby
scope "/user/:display_name" do
  resources :user_mutes, :only -> [:create, :destroy]
end
```

We're (slowly) getting rid of all the "full handwritten" routes so I'm keen to avoid adding more.

test/integration/user_muting_test.rb
```diff
@@ -0,0 +1,31 @@
+require "test_helper"
+
+class UserMutingTest < ActionDispatch::IntegrationTest
```
gravitystorm, line 3: This should perhaps be a system test rather than an integration test, where we can use more natural things like "click_button" rather than `assert_select` followed by directly posting to controllers.

test/integration/user_muting_test.rb
```diff
@@ -0,0 +1,31 @@
...
+  def test_muted_messages_are_moved_into_separate_inbox
+    user = create(:user)
+    other_user = create(:user)
+    user.mutes.create(:subject => other_user)
+
+    session_for(user)
+    get inbox_messages_path
+    assert_select "a[href='#{muted_messages_path}']", false
+
+    session_for(other_user)
```
gravitystorm, line 13: I think just `create(:message, ....)` rather than switching sessions here

test/models/user_mute_test.rb
```diff
@@ -0,0 +1,26 @@
+require "test_helper"
+
+class UserMuteTest < ActiveSupport::TestCase
+  def test_messages_by_muted_users_are_muted
+    user = create(:user)
+    muted_user = create(:user)
+    create(:user_mute, :owner => user, :subject => muted_user)
+
+    message = build(:message, :sender => muted_user, :recipient => user)
+    message.save
```
gravitystorm, line 10: `create` rather than `build` followed immediately by `save`?

app/views/application/_settings_menu.html.erb
```diff
@@ -14,5 +14,8 @@
     <li class="nav-item">
       <%= link_to t(".oauth2_authorizations"), oauth_authorized_applications_path, :class => "nav-link #{'active' if controller_name == 'oauth2_authorized_applications'}" %>
     </li>
+    <li class="nav-item">
+      <%= link_to t(".muted_users"), user_mutes_path, :class => "nav-link #{'active' if controller_name == 'user_mutes'}" %>
```
gravitystorm, line 18: We hide the "Muted messages" tab when there's no muted messages. Should we also hide the "Muted users" tab when you haven't muted anyone yet? There's not much value for this tab until you do (and the muting happens elsewhere), and most people are (hopefully) never going to need it.

gravitystorm (changes requested):
In general this looks good to me, just a few more notes on things I would like to see before it's ready to go.

app/views/application/_settings_menu.html.erb
grekko: yup. I changed `messages/outbox.html.erb` to also reuse `messages/_messages_table.html.erb` and therefore pass a custom `inner_partial` and a new configuration – `columns` to `messages/_messages_table.html.erb`.

I'd understand if the view logic seems a bit overcomplicated now, but I think its still not too hard to read and therefore the message table logic is kept together in a small set of files.

See c0cbf1c28da68739bc854eb8249f5298ce16e8a5

app/views/messages/_messages_table.html.erb
grekko: see c0cbf1c28da68739bc854eb8249f5298ce16e8a5

grekko:
@gravitystorm thanks for the review comments! I think I've addressed them all.

I just noticed and addressed 1 bug and 1 odd thing:
- The `Mute this User`-action on `app/views/users/show.html.erb` was displayed to unauthenticated users. I think this was confusion.
- The redirects in `UserMutes#destroy` after Unmuting the last muted user caused a redirect loop – not anymore.

test/integration/user_muting_test.rb
grekko: I've transformed the integration test into a system spec and split up the test cases so that no session switch/or sign_in/sign_out is necessary anymore to verify the behaviour.

app/controllers/messages_controller.rb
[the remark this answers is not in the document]
grekko: We also hide the `Muted users`-Tab on the Account-page, so I understand that we settled on this behaviour (for now) => resolving the conversation.

config/locales/en.yml
```diff
@@ -1726,9 +1735,16 @@ en:
...
       destroy_button: "Delete"
+    heading: 
+      my_inbox: "My Inbox"
+      my_outbox: "My Outbox"
+      muted_messages: "Muted messages"
     mark:
       as_read: "Message marked as read"
       as_unread: "Message marked as unread"
+    unmute:
+      notice: "Message has been moved to inbox"
```
grekko, line 1746:
```suggestion
      notice: "Message has been moved to Inbox"
```

config/locales/en.yml
```diff
@@ -1726,9 +1735,16 @@ en:
...
+      error: "The message could not be moved to the inbox."
```
grekko, line 1747:
```suggestion
      error: "The message could not be moved to the Inbox."
```

script/deliver-message
```diff
@@ -33,6 +33,6 @@ mail = Mail.new($stdin.read
 message = Message.from_mail(mail, from, to)
 message.save!
 
-UserMailer.message_notification(message).deliver
+UserMailer.message_notification(message).deliver_later if message.notify_recipient?
```
grekko, line 36:
```suggestion
UserMailer.message_notification(message).deliver if message.notify_recipient?
```

grekko:
I restored an unintended change (62fb859) and adjusted the locales so that "inbox" is written as "Inbox".

@gravitystorm Let me know if you spot anything else that you think needs adjustment.

gravitystorm:
> @gravitystorm Let me know if you spot anything else that you think needs adjustment.

The most obvious blocker at the moment is the git history, it would be great to clean that up and force push to this branch to update this PR.

I'm keen to get this merged, it looks pretty much ready to go now.

grekko:
> The most obvious blocker at the moment is the git history

Yup. I planned to invest the time to cleanup the git history once we feel comfortable to move forward. I will do this now then.

grekko:
@gravitystorm A fresh git history. I tried to cut it into meaningful parts. Let me know if you spot changes that you rather have merged into a single commit or split up.

gravitystorm (approved):
Thanks, I'm happy with this PR as-is, and I think it's ready to merge.

There were a couple of UI suggestions from @tomhughes that are not fully resolved in the discussion, and I would be happy to see them addressed as followups since I agree they are slightly tricky, and also I think they aren't blockers for the initial functionality.

grekko:
>  I'm happy with this PR as-is, and I think it's ready to merge.

Thanks a lot for the support and steering the process. 🙇🏽 

> There were a couple of UI suggestions…

I am happy to contribute changes in that direction. I see the changes are also rolled out and am currently testing them manually in production.

I am looking forward to see if the features get any traction and wether it adds the expected value for some users in the community.

PR #4286, opened 2023-10-12
.github/workflows/lint.yml: Check if annotate_models is up to date

grekko:

The model annotations are not so useful when they are outdated. Automatically checking the model annotations ensures they are up to date.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -95,4 +95,4 @@ jobs:
     - name: Run Annotate Models
       run: bundle exec rails annotate_models
     - name: Fail if model annotations are out of date
-      run: git diff --exit-code app/models/
+      run: git diff --exit-code
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -72,3 +72,18 @@ jobs:
         bundler-cache: true
     - name: Run brakeman
       run: bundle exec brakeman -q
+  annotate_models:
+    name: Rails Annotate Models
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v4
+    - name: Setup ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ env.ruby }}
+        bundler-cache: true
+    - name: Run Annotate Models
+      run: bundle exec rails annotate_models
+    - name: Fail if model annotations are out of date
+      run: git diff --exit-code > /dev/null
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -95,4 +95,4 @@ jobs:
     - name: Run Annotate Models
       run: bundle exec rails annotate_models
     - name: Fail if model annotations are out of date
-      run: git diff --exit-code > /dev/null
+      run: git diff --exit-code app/
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -73,6 +73,8 @@ jobs:
     - name: Run brakeman
       run: bundle exec brakeman -q
   annotate_models:
+    env:
+      RAILS_ENV: test
     name: Rails Annotate Models
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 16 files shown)


grekko:
> You'll need a prerequisites PR merged first that fixes all the annotations

I wanted to first see the new lint step fail – and see that it works – and then add a new commit to this branch that updates the annotations.

> The annotate gem is currently only included in the development group

true :)

> or create a ":developement, :test" group for things like this.

I'll add a `":developement, :test"`-group!

grekko:
@gravitystorm would you approve the WF run to see if the lint task works? 🙏

.github/workflows/lint.yml
```diff
@@ -72,3 +72,27 @@ jobs:
...
+      run: |
+        sudo systemctl start postgresql
+        sudo -u postgres createuser -s $(id -un)
+        createdb openstreetmap
+        cp config/github.database.yml config/database.yml
+        bundle exec rails db:test:load
+    - name: Run Annotate Models
+      run: bundle exec rails annotate_models
+    - name: Fail if model annotations are out of date
+      run: git diff --exit-code app/models/
```
gravitystorm, line 98: Is this restriction necessary? It means we need to change this in future if we annotate factories or routes etc.

If we can relax the restriction that's less work for the future.

gravitystorm (changes requested):
Looks good, just one inline question and two other requests:

* From CONTRIBUTING.md - "Avoid including "fixup" commits. If you have added a fixup commit (for example to fix a rubocop warning, or because you changed your own new code) please combine the fixup commit into the commit that introduced the problem. git rebase -i is very useful for this."
* Could you please have the changes to the docs as the first commit, and then the addition of the linting as a second commit, that way the CI would pass for either of them.

.github/workflows/lint.yml
grekko: I think it can be relaxed.

grekko:
I do not like rewriting the commit history in a PR with attached inline comments. It creates dangling comments and makes it harder to understand the PR history and structure retrospectively.

I'll create a new PR with the commits as you described.

---

On that note a question. I understand that PRs in this repo are committed with Merge Commits, which explains why it's preferred to have a clean commit history in each PR.

In the past years I've worked exclusively in projects squash-merging PRs. Why are Merge Commits the preferred way to merge and not squash-merging?

_I do not want to promote squash-merging here. Just want to understand why Merge Commits are preferred._

grekko:
Closing PR. Replaced by #4289

gravitystorm:
> I do not like rewriting the commit history in a PR with attached inline comments. It creates dangling comments and makes it harder to understand the PR history and structure retrospectively.

I think that it's fine to update an existing PR, despite the "dangling review comments" issue. But it's not a major deal. I think about 90% of the time it's better to update an existing PR.

> Why are Merge Commits the preferred way to merge and not squash-merging?

If we squash-merge then `git blame` only shows the whole PR, as if it was done in one large commit. This can also make it harder to run tools like `git bisect` to track down a problem, and it also means that individual component changes of a PR can't be explained in their corresponding commit messages.

So [we ask for contributors to](https://github.com/openstreetmap/openstreetmap-website/blob/3c706ab8470e3c638126abd765d3295ed543aea5/CONTRIBUTING.md#L104-L1050):

> * Split up large commits into smaller units of functionality.
> * Keep your commit messages relevant to the changes in each individual commit.

An example would be #4278 (Upgrade to rails 7.1) where there are a number of related-but-yet-distinct changes, and the commits explain each change in isolation. If we had just one squashed commit for that whole PR, it would be a nightmare in future to understand which code changes related to which part of the (consequently large, if you combined them all) commit message.

An alternative would be rebase-merging, where all the commits are kept but replayed/cherry-picked onto master. I'd be fine with that, if anyone things the merge commits themselves cause problems, but merge-commit-merges and rebase-merges are otherwise very similar.

tomhughes:
In addition to what @gravitystorm says when I'm merging I normally force a merge commit even if it would otherwise do a fast-forward (ie if it has already been derived from or rebased on current master) so that we can tell what PR bought a particular set of commits in from the record in the merge commit.

Issue #4246, opened 2023-09-13
Welcome page interrupts Oauth authorization flow for newly created OSM accounts


milan-cvetkovic:
### Problem

When a first-time user of a 3rd party app attempts to make a contribution to OSM, they have to create an OSM account and authorize the 3rd party app to act on their behalf - modify maps, read profile. Here is the typical scenario when user creates OSM account with username/password:

1. User initiates authorization of the app with OSM from the 3rd party app
2. User sees OSM login page, and choses to Sign-up instead, since they do not have OSM account.
3. User fills up email, password, display name, and submits the form.
4. User accepts the terms, and confirmation email is sent to them
5. After clicking the link in the email, they see "welcome" screen. The authorization of the 3rd party app has been interrupted.

To complete the authorization user needs to restart the process by initiating authorization of the app with OSM again, when
 OSM displays authorization screen, after accepting it, the authorization is completed. Typical web application would regain control of the flow at this time.

Similar flow happens when user uses "Login-with" instead of creating username/password OSM account. When email verification is not necessary, this becomes the only step where the flow is terminated:

1. User initiates authorization of the app with OSM from the 3rd party app
2. User sees OSM login page, and choses to use one of "Log-in with..." buttons.
3. User fills up display name, clicks "Sign-up"
4. User accepts the terms, and submits the form
5. If email verification is not performed, user would be redirected directly to "welcome" page. The authorization of the 3rd party app has been interrupted.

To complete the authorization user needs to restart the process by initiating authorization of the app with OSM.

I have observed this behaviour with several applications, JOSM, osmcha, Map builder, to name a few. I believe it is named "semi-automatic" in JOSM for this reason.

### Description

Prevent "welcome" screen to interrupt Oauth flow in the above scenarios.

Proposal, one of the following:

1. Skip "welcome" screen entirely if creating the user is initiated from Oauth flow:
    - If there is no email confirmation, after accepting the terms, user would be redirected to authorization page
    - If confirmation email is sent, the link in the email would redirect user directly to authorization page, after potentially having them sign-in if necessary
2. Add an additional "Continue with Authorization" button to Welcome screen that would be displayed only during Oauth flow. Again, even when user came to "welcome" screen from confirmation email, the Continue button would actually redirect them to authorization page.

I would prefer the former, simply due to having one click less to "the end of the tunnel."

Creating this issue for more detailed discussion of single topic mentioned in #2287.

gravitystorm:
[...]

PR #4288, opened 2023-10-13
Send welcome email instead of redirecting to welcome page when creating user from authorization

milan-cvetkovic:
When user creates OSM account during Oauth2 authorization initiated from a third-party app, send "Welcome" email instead of redirecting user to "Welcome" page.

- When creating OSM account from OSM website, behaviour is unchanged.
- When creating OSM account during OAuth2 authorization, send a welcome email
   and redirect user to authorization screen after completing the sign-up process.
   
  Related to #4246

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1594,6 +1594,90 @@ en:
       greeting: "Hi,"
       hopefully_you: "Someone (hopefully you) would like to change their email address over at %{server_url} to %{new_address}."
       click_the_link: "If this is you, please click the link below to confirm the change."
+    welcome_email:
+      subject: "[OpenStreetMap] Welcome!"
+      introduction: |
+        Welcome to OpenStreetMap, the free and editable map of the world. Now that you're signed up,
+        you're all set to get started mapping. Here's a quick guide with the most important
+        things you need to know.
+      whats_on_the_map:
+        title: What's on the Map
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -1594,6 +1594,90 @@ en:
       greeting: "Hi,"
       hopefully_you: "Someone (hopefully you) would like to change their email address over at %{server_url} to %{new_address}."
       click_the_link: "If this is you, please click the link below to confirm the change."
+    welcome_email:
+      subject: "[OpenStreetMap] Welcome!"
```
milan-cvetkovic, line 1598: I am not really sure how to reduce the duplication of messages for welcome email welcome screen.

It looks like the template files need to be stored in specific directories: `views/user_mailer/welcome_email.*.erb` and `views/site/welcome.html.erb` - so we cannot reuse the same template file.

I could potentially reuse the strings - use `layouts/site/welcome` strings instead of `layouts/user_mailer/welcome_email` in `views/user_mailer/welcome_email.*.erb`, but that didn't seem to be clean.

stalker314314:
Friendly ping:)

milan-cvetkovic:
Closing, since the alternate [PR #4329](https://github.com/openstreetmap/openstreetmap-website/pull/4329) has been merged.

PR #4289, opened 2023-10-14
.github/workflows/lint.yml: Check if annotate_models is up to date

grekko:

Replacement PR for #4286 

The model annotations are not so useful when they are outdated. Automatically checking the model annotations ensures they are up to date.


gravitystorm:
Thanks!

PR #4291, opened 2023-10-14
Allow localhost for OAuth2 redirect uri

Zverik:
I had to spend an hour debugging and digging deep into Doorkeeper to learn that the website was configured to accept only `127.0.0.1`, but not `localhost`. I guess this fixes that.


Zverik:
Oops. Duplicate of #4287.

Issue #2527, opened 2020-01-30
Fix dropdown widths for RTL languages


gravitystorm:
It looks like the width of the dropdowns isn't being correctly calculated for RTL display:

![Screen Shot 2020-01-30 at 13 57 09](https://user-images.githubusercontent.com/360803/73451713-8ed1df00-4368-11ea-96ac-c5c95d9dc958.png)

Eagle eyed viewers might also spot that the rounding on the button group is also incorrect.

Thanks to @mmd-osm for reporting.

iriman:
Also arrows of tooltips

![image](https://user-images.githubusercontent.com/8467903/73789374-e7afd600-47b3-11ea-9487-da4906b659cb.png)

hiddewie:
Also see https://github.com/Leaflet/Leaflet/issues/7201

gravitystorm:
I hope that both of these (menu rounding and dropdown widths) are fixed by upgrading to bootstrap 5, but we'll need to see if that's the case after the upgrade is complete. At that point, we can check the status of the other problems highlighted here too.

bgo-eiu:
@gravitystorm Do you know why the menu rounding might still be occurring? From what I understand, Bootstrap 5 has switched to using "logical" `border-radius` CSS statements which use options like `-start` and `-end`. I am not sure why upgrading has not changed this, as the `left` and `right` statements are still present. (I tried looking for where this could be happening, but I am not very familiar with SCSS.)

gravitystorm:
> Do you know why the menu rounding issue might still be occurring?

It's a fairly complex situation. Bootstrap 5 has moved to logical class names, but they still implement these with left/right css statements under the hood. You are then [expected to include the bootstrap-rtl variant](https://getbootstrap.com/docs/5.0/getting-started/rtl/) of their stylesheet, but [we use r2](https://github.com/openstreetmap/openstreetmap-website/issues/3433) instead.

I think they kept the left/right implementation for compatibility reasons in bootstrap 5, but I suspect they'll move to using the logical class names at some point in the future.

I would need to investigate further what is happening in this instance, which would then point the way forward. I'm not sure whether that will be changes to the CSS, switching r2 for the rtlcss version of bootstrap, or something else.

bgo-eiu:
Thanks for the explanation, I had seen that RTL documentation page but did not realize it expects a separate variant of their stylesheet

gravitystorm:
[...]

PR #4293, opened 2023-10-17
Fix tooltip left/right placement and arrows in rtl

AntonKhorev:
As mentioned in https://github.com/openstreetmap/openstreetmap-website/issues/2527#issuecomment-582127582 (but borders are still broken, however dropdown widths seem ok without changing anything).

image

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -175,8 +175,7 @@ $(document).ready(function () {
     }
   }
 
-  var placement = $("html").attr("dir") === "rtl" ? "right" : "left";
-  $(".leaflet-control .control-button").tooltip({ placement: placement, container: "body" });
+  $(".leaflet-control .control-button").tooltip({ placement: "left", container: "body" });
 
   var expiry = new Date();
   expiry.setYear(expiry.getFullYear() + 10);
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -2,6 +2,34 @@
 @import "bootstrap";
 @import "rails_bootstrap_forms";
 
+/* Bootstrap + r2 fixes */
+
+:root[dir=rtl] {
+  .bs-tooltip-auto[data-popper-placement^="right"] .tooltip-arrow {
+    /* no-r2 */
+    right: unset !important;
+    left: calc(-1 * var(--bs-tooltip-arrow-height)) !important;
+
+    &::before {
+      /* no-r2 */
+      left: unset !important;
+      right: -1px !important;
+    }
+  }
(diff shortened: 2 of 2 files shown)


tomhughes:
This is much better than before but it doesn't seem to be quite right yet - with both Firefox and Chromium I'm seeing a small gap between the tooltip and the arrow in RTL locales that isn't there in LTR ones:

Screenshot from 2023-10-17 18-42-06

tomhughes:
It seems to need two pixels on this machine, possibly because it's a hi-dpi screen.

What this is all saying to me though is that this is a hack rather than really fixing the underlying problem...

AntonKhorev:
Increased left/right positions to 2px, now I don't see the gaps even when zoomed in.

tomhughes:
This is what is actually needed:

```css
:root[dir=rtl] {
  .bs-tooltip-auto[data-popper-placement^="right"] .tooltip-arrow {
    /* no-r2 */
    right: unset !important;
    left: 0px !important;

    &::before {
      /* no-r2 */
      left: unset !important;
      right: -1px !important;
    }
  }

  .bs-tooltip-auto[data-popper-placement^="left"] .tooltip-arrow {
    /* no-r2 */
    left: unset !important;
    right: 0px !important;

    &::before {
      /* no-r2 */
      right: unset !important;
      left: -1px !important;
    }
  }
}
```

That reverses the changes that r2 is making to the bootstrap code.

HolgerJeromin:
> possibly because it's a hi-dpi screen.

CSS works on "CSS pixels" which are after that zoomed to the hi-dpi monitor. Otherwise all pixel definitions were wrong after switching to those monitors.

AntonKhorev:
> CSS works on "CSS pixels" which are after that zoomed to the hi-dpi monitor.

But fractional pixels are possible, which then get rounded to "real pixels", leading to those gaps appearing sometimes.

AntonKhorev:
Positioning the `.tooltip-arrow` is something you definitively need.

Things like the following are implementation details that are more likely to break after an update:

    &::before {
      /* no-r2 */
      left: unset !important;
      right: -1px !important;
    }

tomhughes:
Well obviously it's horrible and fragile whatever way you do it but I'd much rather match what bootstrap does then do something slightly different that tries to compensate and I don't see why one is more fragile than the other - changes to bootstrap are equally likely to break both but when breakage happens it's much easier to fix if you can look at what has changed in bootstrap and reflect that change in our overrides then if you have to try and adjust some compensatory change by trial and error.

AntonKhorev:
Updated for Bootstrap 5.3

Issue #3433, opened 2022-01-20
Consider dropping r2


gravitystorm:
I'm saying "consider" here since I haven't researched enough to check that it's fully feasible.

We use https://github.com/ded/R2 (technically https://github.com/mzsanford/R2rb ) for generating an RTL version of our CSS, with all the `left` swapped to `right` etc.

However, with modern approaches to CSS and grid layouts, we might be able to drop the need for this altogether. Rather than e.g. float-left for user icons, I've been refactoring to use grid layouts where the first item comes first, and that's either left or right depending on the text direction. That approach makes the layout happen correctly automatically, without needing two CSS files for the different text directions. Understanding our direction-dependent CSS stuff is additional overhead for new developers.

So we need to work through our CSS file, and remove any mentions of left and right (and any asymmetrical `padding` or `margin` combined declarations). See also #3432.

(We also need to remove any direction-dependent margin and padding bootstrap classes that have crept in to use, of which there are a couple, but these aren't affected by r2 so can be dealt with separately).

At that point, the only direction-dependent declarations will be things like icon-maps, which are surrounded by `no-r2` comments. If we make it to that point, then r2 can be dropped.

After this, anything needing left or right can be dealt with by using CSS or grid concepts of start and end, and these will work regardless of text direction.

AntonKhorev:
R2 was dropped in #4353

PR #4296, opened 2023-10-18
Use flex container instead of floats in header

### LINKED ISSUE #2536: 2 menus overlap from 840px to 910px width viewport
image
not always reproduced

AntonKhorev:
This is a step towards [Bootstrap navbar](https://getbootstrap.com/docs/5.2/components/navbar/), which also uses flex.

You don't want floats there because https://github.com/openstreetmap/openstreetmap-website/issues/3433#issuecomment-1632814133 too and they don't make sense anyway when flex is widely available in browsers.

Fixes #2536 if it was caused by line breaks when measuring widths.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -103,6 +103,8 @@ $(document).ready(function () {
 
     $("body").removeClass("compact-nav");
 
+    $("header").removeClass("text-nowrap");
+
     updateHeader();
 
     $(window).resize(updateHeader);
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -957,10 +957,9 @@ img.user_thumbnail {
 }
 
 img.user_thumbnail_tiny {
-  width: auto;
-  height: auto;
-  max-width: 25px;
-  max-height: 25px;
+  width: 25px;
+  height: 25px;
+  object-fit: contain;
 }
 
 /* General styles for action lists / subnavs */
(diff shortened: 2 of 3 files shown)


gravitystorm (changes requested):
Thanks @AntonKhorev , I definitely want to get rid of all of the floats.

However, in this PR the menus aren't working for me when I change the window size. For example, everything is fine at wide screen widths:

Screen Shot 2023-11-15 at 15 51 45

Then as the screen is narrowed there is a positioning error:

Screen Shot 2023-11-15 at 15 52 05

Then slightly narrower it looks OK again:

Screen Shot 2023-11-15 at 15 52 10

Then during the next transition there is also an error:

Screen Shot 2023-11-15 at 15 52 16

Then it looks OK again:

Screen Shot 2023-11-15 at 15 52 23

Other notes:
* The same errors are shown as the width increases (vs decreases)
* The same errors are shown when refreshing the page at the given error width
* UA: Firefox 119.0.1 (Ubuntu)
* If there's some way to avoid the absolute positioning (of the menu icon in small-nav mode) that would be great. Perhaps it can be positioned using flex too?

AntonKhorev:
Seems to be related to an avatar image.

AntonKhorev:
Changed user thumbnail to always be 25x25 px instead of width: auto, should help with width calculations.

AntonKhorev:
> If there's some way to avoid the absolute positioning (of the menu icon in small-nav mode) that would be great.

This is temporary if the end goal is [Bootstrap navbar](https://getbootstrap.com/docs/5.3/components/navbar/). It has its own menu button.

gravitystorm (approved):
Looks good to me, thanks!

PR #4297, opened 2023-10-18
Remove some tag table custom css

AntonKhorev:
I think @gravitystorm used tags table as an example of what's to be converted to Bootstrap, like in https://github.com/openstreetmap/openstreetmap-website/pull/3535#issuecomment-1125113979. I don't think think it's a good example because Bootstrap doesn't have:
- `table-layout: fixed`
- `white-space: pre-wrap`
- `border-collapse: separate`, which is required for rounded corners directly on the table
- without `border-collapse: separate` you need a wrapper with a border and rounded corners, then you can't use Bootstrap's bordered table class because it will have an outer border which you don't need.
- actually you could use `box-shadow` instead of a wrapper to draw a border, but Bootstrap doesn't have that too

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -736,45 +736,16 @@ tr.turn:hover {
   }
 
   .browse-tag-list {
-    background-color: $offwhite;
     table-layout: fixed;
-    border-collapse: separate;
-    border-spacing: 0;
-    width: 100%;
-    margin-bottom: $spacer;
-
-    th, td {
-      border-bottom: 1px solid $grey;
-    }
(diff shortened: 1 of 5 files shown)


tomhughes:
There seems to be quite a lot less padding in the table cells with this applied?

AntonKhorev:
Padding is smaller because I use `table-sm`. Do you want larger padding? I made it smaller because the table is usually in the sidebar.

AntonKhorev:
Increased horizontal padding. Vertical padding is slightly smaller because Bootstrap's predefined values are p-1 = 4px and p-2 = 8px and previously there was 6px.

PR #4299, opened 2023-10-18
Remove subscribe/unsubscribe button custom css

AntonKhorev:
The css was added before Bootstrap. Now together with Bootstrap it makes a small button slightly smaller. I don't see the point of it.

Button after this PR:
image


gravitystorm:
Thanks!

PR #4300, opened 2023-10-18
Simplify primary key for relation members tables

tomhughes:
The two relation members tables have, for some reason that I don't really understand, a primary key that includes all the fields rather than just those needed to identify the member uniquely.

We then have to lie to rails about the primary key makeup in order for things to work properly.

I can't see any purpose to this - there don't seem to be any relationships or queries that would be helped by the extra fields, and having noise between `version` and `sequence_id` means postgres won't be able to assume ordering during an index read of members for a relation and will have to do a sort.

### DIFF
--- a/app/models/old_relation_member.rb
+++ b/app/models/old_relation_member.rb
@@ -20,7 +20,6 @@
 
 class OldRelationMember < ApplicationRecord
   self.table_name = "relation_members"
-  self.primary_key = %w[relation_id version sequence_id]
 
   belongs_to :old_relation, :query_constraints => [:relation_id, :version], :inverse_of => :old_members
   # A bit messy, referring to the current tables, should do for the data browser for now
--- a/app/models/relation_member.rb
+++ b/app/models/relation_member.rb
@@ -19,7 +19,6 @@
 
 class RelationMember < ApplicationRecord
   self.table_name = "current_relation_members"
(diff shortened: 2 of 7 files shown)


gravitystorm:
I'm happy with the change. I'd like to suggest a different approach in the tests. Since the actual values of the `sequence_id` aren't important (they are never displayed anywhere, they just need to be larger and different than the previously used one) then we can use a Factorybot sequence for them:

```diff
 FactoryBot.define do
   factory :relation_member do
-    sequence_id { 1 }
+    sequence(:sequence_id)
```

```diff
 FactoryBot.define do
   factory :old_relation_member do
+    sequence(:sequence_id)
     member_role { "" }
```

Then we don't have to introduce all the explicit references to sequence_ids elsewhere in the tests. They'll be incremented automatically (albeit not starting from 1 for each relation, but who cares), and can still be overridden when necessary.

tomhughes:
I did wonder if there was a better way to do it with factorybot but didn't come up with a good solution that truly modelled how the sequences work so wound up basing it on how we were doing way_node instead.

We certainly could do it the way you say - it means they're not truly modelling how a real relation would work but that probably doesn't matter.

tomhughes:
I've changed the sequence number allocation now along the lines you suggested.

PR #4301, opened 2023-10-20
Add user block api call

AntonKhorev:
This lets block users by id, which helps in case they frequently change their username.

### DIFF
--- a/app/controllers/api/user_blocks_controller.rb
+++ b/app/controllers/api/user_blocks_controller.rb
@@ -1,5 +1,8 @@
 module Api
   class UserBlocksController < ApiController
+    before_action :check_api_writable, :only => :create
+    before_action :authorize, :only => :create
+
     authorize_resource
 
     before_action :set_request_formats
@@ -10,6 +13,34 @@ module Api
       @user_block = UserBlock.find(params[:id])
     rescue ActiveRecord::RecordNotFound
       raise OSM::APINotFoundError
+    end
+
+    def create
+      raise OSM::APIBadUserInput, "No user was given" unless params[:user]
+
+      user = User.visible.find_by(:id => params[:user])
+      raise OSM::APINotFoundError unless user
+      raise OSM::APIBadUserInput, "No reason was given" unless params[:reason]
+      raise OSM::APIBadUserInput, "No period was given" unless params[:period]
+
+      period = Integer(params[:period], :exception => false)
+      raise OSM::APIBadUserInput, "Period should be a number of hours" unless period
+
+      max_period = UserBlock::PERIODS.max
+      raise OSM::APIBadUserInput, "Period must be between 0 and #{max_period}" if period.negative? || period > max_period
+      raise OSM::APIBadUserInput, "Needs_view must be true if provided" unless params[:needs_view].nil? || params[:needs_view] == "true"
+
+      ends_at = Time.now.utc + period.hours
+      needs_view = params[:needs_view] == "true"
+      @user_block = UserBlock.create(
+        :user => user,
+        :creator => current_user,
+        :reason => params[:reason],
+        :ends_at => ends_at,
+        :deactivates_at => (ends_at unless needs_view),
+        :needs_view => needs_view
+      )
+      render :show
     end
   end
 end
--- a/lib/oauth.rb
+++ b/lib/oauth.rb
@@ -1,11 +1,11 @@
 module Oauth
   SCOPES = %w[
     read_prefs write_prefs write_diary
-    write_api write_changeset_comments read_gpx write_gpx write_notes write_redactions
+    write_api write_changeset_comments read_gpx write_gpx write_notes write_redactions write_blocks
     consume_messages send_messages openid
   ].freeze
   PRIVILEGED_SCOPES = %w[read_email skip_authorization].freeze
-  MODERATOR_SCOPES = %w[write_redactions].freeze
+  MODERATOR_SCOPES = %w[write_redactions write_blocks].freeze
 
   class Scope
     attr_reader :name
--- a/test/controllers/api/user_blocks_controller_test.rb
+++ b/test/controllers/api/user_blocks_controller_test.rb
@@ -4,6 +4,10 @@ module Api
   class UserBlocksControllerTest < ActionDispatch::IntegrationTest
     def test_routes
       assert_routing(
+        { :path => "/api/0.6/user_blocks", :method => :post },
+        { :controller => "api/user_blocks", :action => "create" }
+      )
+      assert_routing(
         { :path => "/api/0.6/user_blocks/1", :method => :get },
         { :controller => "api/user_blocks", :action => "show", :id => "1" }
       )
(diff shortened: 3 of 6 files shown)


lib/oauth.rb
```diff
@@ -1,7 +1,7 @@
 module Oauth
   SCOPES = %w[read_prefs write_prefs write_diary write_api read_gpx write_gpx write_notes].freeze
   PRIVILEGED_SCOPES = %w[read_email skip_authorization].freeze
-  OAUTH2_SCOPES = %w[openid].freeze
+  OAUTH2_SCOPES = %w[write_blocks openid].freeze
```
mmd-osm, line 4: I'm wondering if it would make sense to only allow moderators and admins to register a new app with "Create and revoke user blocks" permissions?

gravitystorm: Possibly. But on the other hand, moderator permissions come and go, and for example I could develop and register an app for use by moderators, despite not being a moderator (or admin) myself.

AntonKhorev: It wouldn't. For example, I need this feature for [osm-revert-scripts](https://github.com/woodpeck/osm-revert-scripts). You can run these scripts on OpenHistoricalMap too. I've registered an app there, but I don't have mod/admin permissions.

app/controllers/api/user_blocks_controller.rb
```diff
@@ -14,5 +16,30 @@ def show
...
+      raise OSM::APIBadUserInput, "No period was given" unless params[:period]
+
+      period = Integer(params[:period], :exception => false)
+      raise OSM::APIBadUserInput, "Period one is in a wrong format" unless period
+
+      max_period = UserBlock::PERIODS.max
+      raise OSM::APIBadUserInput, "Period must be between 0 and #{max_period}" if period.negative? || period > max_period
+      raise OSM::APIBadUserInput, "Needs_view must be true if provided" unless params[:needs_view].nil? || params[:needs_view] == "true"
+
+      @user_block = UserBlock.create(
```
mmd-osm, line 35: We should probably check if the user to be blocked is not a mod or admin and is also different from the current user id.

lib/oauth.rb
mmd-osm: I agree it's more convenient for a developer. On the other hand, mods need to be extra careful when authorizing apps, since there will be no control on who can request this permission for their app.

app/controllers/api/user_blocks_controller.rb
```diff
@@ -14,5 +16,30 @@ def show
     rescue ActiveRecord::RecordNotFound
       raise OSM::APINotFoundError
     end
+
+    def create
+      raise OSM::APIBadUserInput, "No user was given" unless params[:user]
+
+      user = User.visible.find_by(:id => params[:user])
+      raise OSM::APINotFoundError unless user
+      raise OSM::APIBadUserInput, "No reason was given" unless params[:reason]
```
mmd-osm, line 25: side note: "Reason" doesn't seem to have a max size check in the model. Maybe it would be good to introduce an upper limit, like 10'000 chars or something similar.

Regarding the reason_format, we're assuming the default "markdown", and don't want to give API users a choice here, right?

tomhughes: Neither of these comments has much to do with this PR which just introduces a new way of creating a block - both of those comments apply to the existing web interface as much as they do to an API call.

As far as format goes only markdown is supported for new data - the other formats are all legacy formats though I suspect they predate blocks in any case so all blocks are probably markdown.

app/controllers/api/user_blocks_controller.rb
```diff
@@ -14,5 +16,30 @@ def show
...
+      raise OSM::APIBadUserInput, "No period was given" unless params[:period]
+
+      period = Integer(params[:period], :exception => false)
+      raise OSM::APIBadUserInput, "Period one is in a wrong format" unless period
```
tomhughes, line 29: I'm not sure what `one` is referring to in this message? It should also be `the wrong format` but overall I'd probably replace the whole thing with something like `Period should be a number of hours`.

tomhughes (changes requested):

lib/oauth.rb
AntonKhorev: You can do redactions with just `write_api`, the permission any editor requires.

app/controllers/api/user_blocks_controller.rb
AntonKhorev: copy-paste accident, fixed

AntonKhorev: I was copying the structure of error messages from other controllers where you can find "Date #{params[:from]} is in a wrong format".

Changed to "Period should be a number of hours".

app/controllers/api/user_blocks_controller.rb
AntonKhorev: > What should happen if the user to be blocked is already currently blocked? Or if we're trying to POST the same user block twice, and introduce duplicates.

You can create multiple blocks from the web interface.

Currently we have two qualitatively different block types: one that requires reading the message and one that doesn't. In the future there might be more types. It might be useful to have multiple active blocks with different periods at the same time.

test/controllers/api/user_blocks_controller_test.rb
```diff
@@ -14,11 +14,21 @@ def test_routes
...
-      assert_select "user_block[id='#{block.id}']", 1
+      assert_select "osm>user_block", 1
+      assert_select "osm>user_block>@id", block.id.to_s
+      assert_select "osm>user_block>user", 1
+      assert_select "osm>user_block>user>@uid", blocked_user.id.to_s
+      assert_select "osm>user_block>creator", 1
+      assert_select "osm>user_block>creator>@uid", creator_user.id.to_s
+      assert_select "osm>user_block>revoker", 0
+      assert_select "osm>user_block>reason", 1
+      assert_select "osm>user_block>reason", "because running tests"
```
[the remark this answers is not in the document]
AntonKhorev: updated

test/controllers/api/user_blocks_controller_test.rb
```diff
@@ -42,5 +46,173 @@ def test_show_not_found
...
+      assert_select "osm>user_block>@id", block.id.to_s
+      assert_select "osm>user_block>@needs_view", "false"
+      assert_select "osm>user_block>user", 1
+      assert_select "osm>user_block>user>@uid", blocked_user.id.to_s
+      assert_select "osm>user_block>creator", 1
+      assert_select "osm>user_block>creator>@uid", creator_user.id.to_s
+      assert_select "osm>user_block>revoker", 0
+      assert_select "osm>user_block>reason", 1
+      assert_select "osm>user_block>reason", "because"
+    end
```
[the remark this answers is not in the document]
AntonKhorev: updated, had to use

```ruby
assert_select ">@id", block.id.to_s
```

test/controllers/api/user_blocks_controller_test.rb
```diff
@@ -42,5 +46,173 @@ def test_show_not_found
...
+
+    [
+      ["missing reason", "No reason was given", { :period => "10" }],
+      ["missing period", "No period was given", { :reason => "because" }],
+      ["non-numeric period", "Period one is in a wrong format", { :reason => "because", :period => "one" }],
+      ["negative period", "Period must be between 0 and #{UserBlock::PERIODS.max}", { :reason => "go away", :period => "-1" }],
+      ["excessive period", "Period must be between 0 and #{UserBlock::PERIODS.max}", { :reason => "go away", :period => "10000000" }],
+      ["unknown needs_view", "Needs_view must be true if provided", { :reason => "because", :period => "1", :needs_view => "maybe" }]
+    ].each do |name, message, params|
+      test "create invalid because #{name}" do
```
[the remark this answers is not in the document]
AntonKhorev: did private method with (message, params) arguments

lib/oauth.rb
AntonKhorev: Added a star next to `write_blocks` scope to make it more noticeable:
image

app/controllers/api/user_blocks_controller.rb
AntonKhorev: > We should probably check if the user to be blocked is not a mod or admin (and implicitly is also different from the current user id).

In case someone who is not supposed to have moderator permissions gets a token with `write_blocks`, blocking mods/admins is the last thing they'll want to do. This will be noticed very quickly. They'll be basically announcing themselves "I stole a mod token!"

mmd-osm: They might notice it quickly. However, an attacker might create hundreds or thousands of user blocks for mod accounts, so it would require an admin to revoke all user blocks and make the mod accounts usable again.

AntonKhorev: we can add a "revoke all blocks from this account" button to the web ui

mmd-osm: Yes, that would also work, and would be a topic for another PR.

lib/oauth.rb
AntonKhorev: addressed in #4387

app/controllers/api/user_blocks_controller.rb
AntonKhorev: "revoke all blocks" is added

AntonKhorev:
Updated to align with #5155.

app/controllers/api/user_blocks_controller.rb
```diff
@@ -12,5 +15,33 @@ def show
...
+    def create
+      raise OSM::APIBadUserInput, "No user was given" unless params[:user]
+
+      user = User.visible.find_by(:id => params[:user])
+      raise OSM::APINotFoundError unless user
+      raise OSM::APIBadUserInput, "No reason was given" unless params[:reason]
+      raise OSM::APIBadUserInput, "No period was given" unless params[:period]
+
+      period = Integer(params[:period], :exception => false)
+      raise OSM::APIBadUserInput, "Period should be a number of hours" unless period
```
mmd-osm, line 28: I'm wondering if the permitted values should be more in line with user_block_periods in settings.yml:

```
# Periods (in hours) which are allowed for user blocks
user_block_periods: [0, 1, 3, 6, 12, 24, 48, 96, 168, 336, 731, 4383, 8766, 87660]
```

Are values not explicitly mentioned there ok for the UI, or would they cause issues in dropdown lists?

AntonKhorev: Block durations are not stored directly. Dropdowns show the closes value to the remaining time: https://github.com/openstreetmap/openstreetmap-website/pull/5011

tomhughes (approved):

PR #4308, opened 2023-10-23
Add banner for SOTM Asia and Africa 2023

atomoil:
Here's a PR to add banners for the other 2 SOTMs happening in 2023 - Asia and Africa - to match the promotion for Europe.

~NOTE: Africa assets are still to come (hence this being a draft)~


tomhughes:
Well africa seems to be missing altogether? If we're waiting on it should we just do a separate PR for it?

I assume you're doing this on behalf of the respective organisers and that they're aware you're doing this?

pnorman:
Can you confirm SOTM Asia and Africa both have a [SOTM Quick Licence](https://osmfoundation.org/wiki/SOTM_Quick_Licence) as per the [banner policy](https://operations.osmfoundation.org/policies/banner/)?

atomoil:
> I assume you're doing this on behalf of the respective organisers and that they're aware you're doing this?

hi @tomhughes yes that's correct - I'm doing this on behalf of the Local Chapters and Communities Working Group, working with Arnalie who is also on the OSM Foundation Board.

tomhughes:
> hi @tomhughes yes that's correct - I'm doing this on behalf of the Local Chapters and Communities Working Group, working with Arnalie who is also on the OSM Foundation Board.

That doesn't answer the question of whether the organisers of those two conferences know about it.

atomoil:
> That doesn't answer the question of whether the organisers of those two conferences know about it.

Yes they do.

pnorman:
The SOTM Africa is not the right size.

https://github.com/openstreetmap/openstreetmap-website/blob/master/FAQ.md#how-do-i-create-a-banner-to-promote-my-openstreetmap-event

 > the image needs be exactly 350px wide and at most 350px tall

atomoil:
@pnorman image is resized - I also switched from png to jpg as the file size is much smaller, but then noticed that the doc you shared specifies PNG but the existing fundraising image is a jpg (which is why I tried a jpg export). Jpg is considerably smaller so seems better for serving on a website, but I can revert to png if you prefer.

arnalielsewhere:
> > hi @tomhughes yes that's correct - I'm doing this on behalf of the Local Chapters and Communities Working Group, working with Arnalie who is also on the OSM Foundation Board.
> 
> That doesn't answer the question of whether the organisers of those two conferences know about it.

Hi Tom, yes they did (I am part of SotM Asia organizing team). I shared about the banner policy and that they can directly send the request via github. They supplied the graphics but looks like none of their members had the time to submit the request.

atomoil:
@tomhughes / @pnorman, is there anything remaining for me to do on this PR before it can be considered for merging? e.g. is it a blocker that I've provided JPEGs rather than PNGs (because the JPEGs are considerably smaller)?

tomhughes:
I though we were still waiting on the size of one of the images being fixed but I see you did that.

Issue #4018, opened 2023-04-22
User account self-deletion allows bad actors to delete and recreate the same account name to "lose" changeset discussion and block history


SomeoneElseOSM:
### URL

https://github.com/openstreetmap/openstreetmap-website/pull/3398

### How to reproduce the issue?

There have been a number of recent examples of the following sort of activity, but lets give a specific example:

A user account was created with a name "francisdouglas88614" to "force through" some changes that they local community decided were not correct.  The UID was 19014237, and due to it being a problematic returning vandal they were reverted and blocked until they contacted the DWG with https://www.openstreetmap.org/user_blocks/7071 (see previous messages in that chain of blocks for the history).

They then delete that account, and create another one - francisdouglas88614 / 19021744, and try and force through their changes again.  They were reverted and blocked with https://www.openstreetmap.org/user_blocks/7074.

They then delete that account, and create another one - francisdouglas88614 / 19031302, and try and force through their changes again.  They were reverted and blocked with https://www.openstreetmap.org/user_blocks/7077.

This is just 1 username here; there are at least 9 or 10 others.  I don't believe that it is unfair to "name and shame" this account as this user has been widely discussed [elsewhere](https://community.openstreetmap.org/t/odd-edits-in-italy/97706).

Basically, this is pretty much as predicted by https://github.com/openstreetmap/openstreetmap-website/issues/1853#issuecomment-387682751 .

In additions to previous problems noted such as https://github.com/openstreetmap/openstreetmap-website/issues/3585 .

To be clear, the idea behind https://github.com/openstreetmap/openstreetmap-website/pull/3398 absolutely makes sense; but we should prevent it from being used by bad actors as it currently is.  There may of course be reasons why a long-term-blocked user should be able to delete their account, and they can always ask the admins to do that.  What I'm suggesting is that a user shouldn't be able to engage on vandalism, get caught, delete their account and repeat the process ad infinitum.

### Screenshot(s) or anything else?

n/a

fitojb:
For starters, used usernames should not be permitted to be reused

tomhughes:
I really don't think that's going to work - we remove the names because they often contain personal data so we don't really have any option.

Besides which it will just leads to loads of people asking why they can't have a name which appears to be unused and anything which gives me more support mails to answer is definitely not on.

HolgerJeromin:
> we remove the names because they often contain personal data so we don't really have any option.

We could save a hash instead.
[...]

PR #4313, opened 2023-10-24
Account deletion cool-down period

AntonKhorev:
As I posted in the dwg chat:

- the delay is 1 day to make sure that there are no open changesets when users are deleted
- there's a setting, delays are disabled when set to 0
- I assumed that no possible law/regulation can exist that would require us to act faster than within a day
- no checks if the user is blocked, only the last changeset opening time
- I'd check the closing time, but there's no index together with uid

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -419,6 +419,18 @@ class User < ApplicationRecord
     end
   end
 
+  def deletion_allowed_at
+    unless Settings.user_account_deletion_delay.nil?
+      last_changeset = changesets.reorder(:closed_at => :desc).first
+      return last_changeset.closed_at.utc + Settings.user_account_deletion_delay.hours if last_changeset
+    end
+    creation_time.utc
+  end
+
+  def deletion_allowed?
+    deletion_allowed_at <= Time.now.utc
+  end
+
   private
 
   def encrypt_password
--- a/app/views/account/deletions/show.html.erb
+++ b/app/views/account/deletions/show.html.erb
@@ -31,5 +31,13 @@
   <li><%= t ".retain_email" %></li>
 </ul>
 
-<%= link_to t(".delete_account"), account_path, { :method => :delete, :class => "btn btn-danger", :data => { :confirm => t(".confirm_delete") } } %>
+<% if current_user.deletion_allowed? %>
+  <%= link_to t(".delete_account"), account_path, { :method => :delete, :class => "btn btn-danger", :data => { :confirm => t(".confirm_delete") } } %>
+<% else %>
+  <div class="alert alert-warning">
+    <%= t ".recent_editing_html", :time => friendly_date(current_user.deletion_allowed_at) %>
+  </div>
+  <button class="btn btn-secondary" disabled><%= t(".delete_account") %></button>
+<% end %>
+
 <%= link_to t(".cancel"), edit_account_path, :class => "btn btn-link" %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -256,6 +256,7 @@ en:
         retain_notes: Your map notes and note comments, if any, will be retained but hidden from view.
         retain_changeset_discussions: Your changeset discussions, if any, will be retained.
         retain_email: Your email address will be retained.
+        recent_editing_html: "As you have edited recently your account cannot currently be deleted. Deletion will be possible in %{time}."
         confirm_delete: Are you sure?
         cancel: Cancel
   accounts:
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -53,6 +53,8 @@ api_timeout: 300
 web_timeout: 30
 # Periods (in hours) which are allowed for user blocks
 user_block_periods: [0, 1, 3, 6, 12, 24, 48, 96, 168, 336, 731, 4383, 8766, 87660]
+# Account deletion cooldown period (in hours) since last changeset close; null to disable, 0 to make sure there aren't any open changesets when the deletion happens
+user_account_deletion_delay: null
 # Rate limit for message sending
 max_messages_per_hour: 60
 # Rate limit for friending
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -61,7 +61,7 @@ Metrics/BlockNesting:
 # Offense count: 26
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 286
+  Max: 297
 
 # Offense count: 59
 # Configuration parameters: AllowedMethods, AllowedPatterns.
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -53,12 +53,16 @@ class AccountsController < ApplicationController
   end
 
   def destroy
-    current_user.soft_destroy!
+    if current_user.deletion_allowed?
+      current_user.soft_destroy!
 
-    session.delete(:user)
-    session_expires_automatically
+      session.delete(:user)
+      session_expires_automatically
 
-    flash[:notice] = t ".success"
-    redirect_to root_path
+      flash[:notice] = t ".success"
+      redirect_to root_path
+    else
+      head :bad_request
+    end
   end
 end
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -14,11 +14,12 @@
 #
 # Indexes
 #
-#  changesets_bbox_idx                (min_lat,max_lat,min_lon,max_lon) USING gist
-#  changesets_closed_at_idx           (closed_at)
-#  changesets_created_at_idx          (created_at)
-#  changesets_user_id_created_at_idx  (user_id,created_at)
-#  changesets_user_id_id_idx          (user_id,id)
+#  changesets_bbox_idx                        (min_lat,max_lat,min_lon,max_lon) USING gist
+#  changesets_closed_at_idx                   (closed_at)
+#  changesets_created_at_idx                  (created_at)
+#  changesets_user_id_created_at_idx          (user_id,created_at)
+#  changesets_user_id_id_idx                  (user_id,id)
+#  index_changesets_on_user_id_and_closed_at  (user_id,closed_at)
 #
 # Foreign Keys
 #
--- a/config/initializers/config.rb
+++ b/config/initializers/config.rb
@@ -79,6 +79,7 @@ Config.setup do |config|
     required(:max_number_of_relation_members).filled(:int?)
     required(:max_issues_count).filled(:int?)
     required(:api_timeout).filled(:int?)
+    required(:user_account_deletion_delay).maybe(:number?)
     required(:imagery_blacklist).maybe(:array?)
     required(:status).filled(:str?, :included_in? => ALLOWED_STATUS)
     required(:avatar_storage).filled(:str?)
--- a/db/migrate/20231117170422_add_closed_at_index_to_changesets.rb
+++ b/db/migrate/20231117170422_add_closed_at_index_to_changesets.rb
@@ -1,0 +1,7 @@
+class AddClosedAtIndexToChangesets < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :changesets, [:user_id, :closed_at], :algorithm => :concurrently
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2500,6 +2500,13 @@ CREATE INDEX index_changeset_comments_on_created_at ON public.changeset_comments
 
 
 --
+-- Name: index_changesets_on_user_id_and_closed_at; Type: INDEX; Schema: public; Owner: -
+--
+
+CREATE INDEX index_changesets_on_user_id_and_closed_at ON public.changesets USING btree (user_id, closed_at);
+
+
+--
 -- Name: index_changesets_subscribers_on_changeset_id; Type: INDEX; Schema: public; Owner: -
 --
 
@@ -3499,6 +3506,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('23'),
 ('22'),
 ('21'),
+('20231117170422'),
 ('20231101222146'),
 ('20231029151516'),
 ('20231010194809'),
--- a/test/controllers/accounts_controller_test.rb
+++ b/test/controllers/accounts_controller_test.rb
@@ -152,4 +152,23 @@ class AccountsControllerTest < ActionDispatch::IntegrationTest
     # Make sure we have a button to "go public"
     assert_select "form.button_to[action='/user/go_public']", true
   end
+
+  def test_destroy_allowed
+    user = create(:user)
+    session_for(user)
+
+    delete account_path
+    assert_response :redirect
+  end
+
+  def test_destroy_not_allowed
+    with_user_account_deletion_delay(24) do
+      user = create(:user)
+      create(:changeset, :user => user, :created_at => Time.now.utc)
+      session_for(user)
+
+      delete account_path
+      assert_response :bad_request
+    end
+  end
 end
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -282,4 +282,62 @@ class UserTest < ActiveSupport::TestCase
     oauth_access_token.reload
     assert_predicate oauth_access_token, :revoked?
   end
+
+  def test_deletion_allowed_when_no_changesets
+    with_user_account_deletion_delay(10000) do
+      user = create(:user)
+      assert_predicate user, :deletion_allowed?
+    end
+  end
+
+  def test_deletion_allowed_without_delay
+    with_user_account_deletion_delay(nil) do
+      user = create(:user)
+      create(:changeset, :user => user)
+      user.reload
+      assert_predicate user, :deletion_allowed?
+    end
+  end
+
+  def test_deletion_allowed_past_delay
+    with_user_account_deletion_delay(10) do
+      user = create(:user)
+      create(:changeset, :user => user, :created_at => Time.now.utc - 12.hours, :closed_at => Time.now.utc - 10.hours)
+      user.reload
+      assert_predicate user, :deletion_allowed?
+    end
+  end
+
+  def test_deletion_allowed_during_delay
+    with_user_account_deletion_delay(10) do
+      user = create(:user)
+      create(:changeset, :user => user, :created_at => Time.now.utc - 11.hours, :closed_at => Time.now.utc - 9.hours)
+      user.reload
+      assert_not_predicate user, :deletion_allowed?
+      assert_equal Time.now.utc + 1.hour, user.deletion_allowed_at
+    end
+  end
+
+  def test_deletion_allowed_past_zero_delay
+    with_user_account_deletion_delay(0) do
+      user = create(:user)
+      create(:changeset, :user => user, :created_at => Time.now.utc, :closed_at => Time.now.utc + 1.hour)
+      travel 90.minutes do
+        user.reload
+        assert_predicate user, :deletion_allowed?
+      end
+    end
+  end
+
+  def test_deletion_allowed_during_zero_delay
+    with_user_account_deletion_delay(0) do
+      user = create(:user)
+      create(:changeset, :user => user, :created_at => Time.now.utc, :closed_at => Time.now.utc + 1.hour)
+      travel 30.minutes do
+        user.reload
+        assert_not_predicate user, :deletion_allowed?
+        assert_equal Time.now.utc + 30.minutes, user.deletion_allowed_at
+      end
+    end
+  end
 end
--- a/test/system/account_deletion_test.rb
+++ b/test/system/account_deletion_test.rb
@@ -41,4 +41,59 @@ class AccountDeletionTest < ApplicationSystemTestCase
 
     assert_content "Account Deleted"
   end
+
+  test "can delete with any delay setting value if the user has no changesets" do
+    with_user_account_deletion_delay(10000) do
+      travel 1.hour do
+        visit edit_account_path
+
+        click_link "Delete Account..."
+
+        assert_no_content "cannot currently be deleted"
+      end
+    end
+  end
+
+  test "can delete with delay disabled" do
+    with_user_account_deletion_delay(nil) do
+      create(:changeset, :user => @user)
+
+      travel 1.hour do
+        visit edit_account_path
+
+        click_link "Delete Account..."
+
+        assert_no_content "cannot currently be deleted"
+      end
+    end
+  end
+
+  test "can delete when last changeset is old enough" do
+    with_user_account_deletion_delay(10) do
+      create(:changeset, :user => @user, :created_at => Time.now.utc, :closed_at => Time.now.utc + 1.hour)
+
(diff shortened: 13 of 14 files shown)


tomhughes:
I don't see any reason to change the user model and store a special time for this - it's just a single query against a changeset index to find the most recent change.

tomhughes:
Ah it looks like I misunderstood what you meant about adding a time to the model.

tomhughes:
Example of what a blocked deletion looks like:

image

AntonKhorev:
> I don't see any reason to change the user model and store a special time

Initially I didn't, but later I needed it in two different places: the controller that shows the screen from the comment above and the controller that actually deletes.

app/views/account/deletions/show.html.erb
```diff
@@ -31,5 +31,13 @@
   <li><%= t ".retain_email" %></li>
 </ul>
 
-<%= link_to t(".delete_account"), account_path, { :method => :delete, :class => "btn btn-danger", :data => { :confirm => t(".confirm_delete") } } %>
+<% if @allowed_at %>
+  <div class="alert alert-warning">
+    <%= t ".recent_editing_html", :time => friendly_date_time_element(@allowed_at) %>
+  </div>
+  <button class="btn btn-danger" disabled><%= t(".delete_account") %></button>
```
tomhughes, line 38: I wonder if we should just leave the button out when deletion is not possible? Even a disabled "danger" button still looks very clickable - the difference to a non-disabled one is quite subtle.

config/settings.yml
```diff
@@ -53,6 +53,8 @@ api_timeout: 300
 web_timeout: 30
 # Periods (in hours) which are allowed for user blocks
 user_block_periods: [0, 1, 3, 6, 12, 24, 48, 96, 168, 336, 731, 4383, 8766, 87660]
+# Account deletion cooldown period since last changeset close (in hours)
+user_account_deletion_delay: 24
```
tomhughes, line 57: Do we want to have a default for this? or should it default to off but be something we can enable operationally? In other words should we assume that other users of this code will want this.

config/locales/en.yml
```diff
@@ -256,6 +256,7 @@ en:
         retain_notes: Your map notes and note comments, if any, will be retained but hidden from view.
         retain_changeset_discussions: Your changeset discussions, if any, will be retained.
         retain_email: Your email address will be retained.
+        recent_editing_html: "Due to your recent editing activity you'll be able to delete your account in %{time}"
```
tomhughes, line 259: This doesn't read very well to me but I'm struggling to come up with a better phrasing, at least as a single sentence. The best I've come up with is:

> As you have edited recently your account cannot currently be deleted.
> Deletion will be possible in %{time}.

tomhughes:
> Initially I didn't, but later I needed it in two different places: the controller that shows the screen from the comment above and the controller that actually deletes.

Yes I has misunderstood what you meant by the commit comment and thought you were actually a field to the database - once I looked at the code I realised what you really meant.

app/views/account/deletions/show.html.erb
mmd-osm: On my mockup I used btn-secondary instead of btn-danger along with the disabled attribute.

AntonKhorev: changed to btn-secondary

config/locales/en.yml
mmd-osm: How about:

Based on your recent mapping activities, deletion of your account is only possible in x days.

config/settings.yml
AntonKhorev: Let's default to the previous behavior, `user_account_deletion_delay = 0`

config/locales/en.yml
AntonKhorev: Changed the message.

I started with the block expiry message to avoid introducing some grammar that might cause trouble during internationalization.

gravitystorm:
I'd like to know more about the purpose behind the proposed delay, since without that being clear it's hard to review the implementation. I don't have any access to DWG discussions (nor might other developers in future working on this).

There is mention here about open changesets - is it imperative that accounts remain open until the last changeset is closed? If so, why? My only guess would be some kind of moderation tool that is waiting for closed changesets, but I assume (perhaps incorrectly) that it's nothing to do with adding more data to an open changeset after the account is closed. Or is it some other reason?

If it needs to wait for the changeset to be closed, then why is there a configuration value? Why is it not based on `Changeset::MAX_TIME_OPEN?` Or if there is some moderator process that needs to wait until all changesets are closed, would it need to be `Changeset::MAX_TIME_OPEN + some_interval`?

I've read the recent discussions on user account self-close and reopen (#4018). If I understand correctly, all this PR would do is prevent a display_name from being quickly reused on another account - is this correct? Is this the only goal? The miscreants are presumably using fresh emails every time, since emails are currently still retained (#3488).

I have a list of suggestions on how to improve the PR, but it would be great for me to have the purpose and goals clear in my own head first, otherwise I don't want to make unhelpful suggestions based on an incorrect picture.

tomhughes:
The issue basically is DWG finding it hard to deal with vandalism when vandals delete their account (hence changing their name) immediately after making their edits.

No doubt some DWG people will be able to provide more details...

mmd-osm:
> If I understand correctly, all this PR would do is prevent a display_name from being quickly reused on another account - is this correct?

I think the scope is wider than that. Once a user has been deleted others will have a fairly difficult time to inspect these changes, etc. for potential vandalism. To me it seems that's a shortcoming of today's concept of deleting users, and I've even warned a few years ago that this is inviting vandalism.

By introducing a cool down period, we're adding a few weeks where a user won't be adding any new changes to the map.

gravitystorm:
@mmd-osm - thanks. Can I ask for more information on this bit, again for clarity:

> Once a user has been deleted others will have a fairly difficult time to inspect these changes, etc. for potential vandalism.

* What causes the difficulty when the account is deleted? Is it because deleted users are unlinked in the web interface? Or is it something else that causes difficulties?
* If the delay is e.g. 48 hours (as the tests suggest) doesn't that mean there will be the original problem after 49 hours?
* You mention "a few weeks" whereas the proposed tests suggest "48 hours" and the documentation in settings.yml only describes the settings of "24" (for changesets being closed) and "0" (to disable). It's all very unclear as to the logic of what the cool-down period is proposed to be, and how (if at all) it relates to changesets. Can you clarify?

> user won't be adding any new changes to the map and their previous changes can still be easily reviewed.

Presumably if the vandal can't close their first account, they will just abandon it and open another one with a slightly different email and use that instead. Based on the discussion in #4018 , these vandals aren't short of fresh email addresses to use on each new account.

So by my understanding, this cool-down period doesn't prevent any vandalism, it is just making reviewing existing vandalism easier. Is that right?

I'm sorry if this sounds really picky, but I want to make sure that I understand the root problem, and what we are trying to implement here, since it has implications for other parts of the user deletion process that we can implement in future.

mmd-osm:
> What causes the difficulty when the account is deleted? Is it because deleted users are unlinked in the web interface? Or is it something else that causes difficulties?

That's probably the biggest issue. It's near impossible to navigate through the list of all changesets created by a deleted users. Also third-party tools are struggling with deleted users, and fail to analyze such changesets: [Example OSMChag issue](https://github.com/mapbox/osmcha-frontend/issues/669).

> If the delay is e.g. 48 hours (as the tests suggest) doesn't that mean there will be the original problem after 49 hours?

In my original proposal I mentioned 28 days, which should leave plenty of time to inspect changesets, and give vandals less of an incentive for a hit-and-run kind of behavior. I find 48 hours a bit short for the purpose of reviewing changesets.

> You mention "a few weeks" whereas the proposed tests suggest "48 hours" and the documentation in settings.yml only describes the settings of "24" 

I think part of this complication is caused by the way how the time since the last edit is calculated. Since we're using the creaed_at timestamp, there might be a time period of up to 24 hours, where the changeset might potentially still be open, and further changes could occur. What we really want for this calculation is the timestamp when the last edit was uploaded, though it seems it's a bit difficult to easily access that information (due to lack of proper database index).

AntonKhorev:
48 hours is just a number used in tests, something longer than a day.

24 hours is a period after which all of the user's changesets are closed. Otherwise there are possible changeset close events for deleted users.

I don't mind the period being longer, that's why I don't base it on `Changeset::MAX_TIME_OPEN`. My original pull request had the period of 1 day because I thought it's easier to get a short period approved.

mmd-osm:
> 24 hours is a period after which all of the user's changesets are closed. Otherwise there are possible changeset close events for deleted users.

So the way this works is, whenever you create a changeset or upload some changes to the same changeset, the closed_at timestamp will be set to the current timestamp + 1 hour. In addition, the total time difference between created_at and closed_at can never exceed 24 hours.  -> Once you stop uploading data, all of your changesets will be implicitly closed latest in one hour.

SomeoneElseOSM:
@gravitystorm re "What causes the difficulty when the account is deleted?"  Some of the issues that we're seeing are described at https://github.com/openstreetmap/openstreetmap-website/issues/4018 .  I've added a few examples of that specific problem there; I could add more if it helps.

> Is it because deleted users are unlinked in the web interface?

This does make it difficult for mappers to report vandals, and (because userid isn't visible on e.g. https://www.openstreetmap.org/changeset/143079112 ) harder for everyone to see "who did what".  Everyone needs to click through to e.g. https://www.openstreetmap.org/api/0.6/changeset/143079112 and there's no way from there to click through to "were they blocked?".  To get to "were they reported" you have to copy and paste an XML value from that screen into the "issues" search .

https://github.com/openstreetmap/openstreetmap-website/issues/4018#issuecomment-1576811268 suggested a workload problem** ("Meanwhile I will get dragged into support complaints from people complaining that they can't delete their accounts.").  As I understand it, https://github.com/openstreetmap/openstreetmap-website/issues/4018#issuecomment-1773917139 and this pull request followed from that.

** and https://github.com/openstreetmap/openstreetmap-website/issues/4018#issuecomment-1776958530 later suggested an unwillingness to deal with the workload by spreading it more widely.

gravitystorm:
> Otherwise there are possible changeset close events for deleted users.

I still don't understand what actual problem would be caused by closing the account before the closed_at time of the last changeset. You can't upload any more changes after your account is closed, so what is the problem that you are trying to prevent? Is there a moderation tool that waits until closed_at times have past and then fails when a user has been deleted? Or something else?

(As @mmd-osm points out the *maximum* time a changeset can be open is 24 hours, but they can only be open for 1 hour from now without another upload to extend the time limit (by another hour). So there are never any changesets with a closed_at time more than 1 hour in the future.)

I sort of feel like this whole "24 to make sure there aren't any open changesets when the deletion happens" could be a misunderstanding or a red herring, or if it's genuininely solving a problem then we need to be very explicit about what that problem is.

gravitystorm:
> I've added a few examples of that specific problem there; I could add more if it helps.

Thanks @SomeoneElseOSM - it's not really the examples of the incidents, it's the details of the problems that are key. Like the "[5 Whys](https://en.wikipedia.org/wiki/Five_whys)" approach to problem solving, I'm trying to get onto the second or third "why".

> This does make it difficult for mappers to report vandals

This statement is absolutely gold, and exactly what I'm trying to get to! Thank you, I hadn't thought of that and I don't think this had been explicitly stated yet (or if it was, then I missed it).

> suggested a workload problem

We need to make sure that if we try to fix a problem we don't cause more problems. So for example, a long cool-down period might help mop up after vandals (it won't slow them down, as above), but might lead to more support requests for people who want to legitimately close their accounts and find the delay unreasonable.

> suggested an unwillingness to deal with the workload by spreading it more widely.

I think we can automate everything to do with account closures, so I would avoid making review queues / OTRS integration or anything like that. It's just adding more work for small groups of already busy people, and the fundamental goal of having software is to reduce the work! The wider topic of "what should be moderators/ what should be admins" and also "who can be moderators / who can be admins" is probably best left for a different discussion too!

SomeoneElseOSM:
@gravitystorm Re "adding more work for small groups of already busy people" are you aware of the amount of work that the DWG is having to do _precisely_ because the "vandals can delete their own account" loophole is open**? 

The approach taken by this PR isn't the one I would have taken (see https://github.com/openstreetmap/openstreetmap-website/issues/4018#issuecomment-1746990870 for my suggestions and offers to help there), but I can understand why it was submitted given the previous replies.

The barrier to entry to be able to deal with deleted users is basically "first, buy a large server and stick an 80GB ChangesetMD database on it".  Regular OSM users simply aren't able to do this and can't easily report problems beyond "there is some vandalism".

There are valid (legal) reasons why genuine users may want to delete their accounts, and valid (ease of use) reasons why that should be automatic if possible.  This does not apply to users signing up purely to abuse the site, the data on it, and its users.  We should not continue to make life easy for these people.

**  Would you like to read some of the abusive messages that we get from these vandals?  A request "can I delete my account" would not greatly add to that load.

gravitystorm:
> @gravitystorm Re "adding more work for small groups of already busy people" are you aware of the amount of work that the DWG is having to do _precisely_ because the "vandals can delete their own account" loophole is open

Yes, I am. That's exactly why I want to make sure we are reducing workload for DWG and not adding to it. 

Please, good faith and all that, and don't jump down my throat on this. I'm prioritising this work because I prioritise things that affect the moderators and administrators. I'm neither a moderator nor an admin, just a volunteer trying to make your lives easier.

AntonKhorev:
> I still don't understand what actual problem would be caused by closing the account before the closed_at time of the last changeset.

This is to avoid the situation when a user is "done", but their changesets are not "done".

> Is there a moderation tool that waits until closed_at times have past and then fails when a user has been deleted?

Kind of. You don't really wait for close_time because it's not observable until the changeset is closed. But you may want to wait for the changeset to close because otherwise you wouldn't know if you've downloaded it fully.

It's not the biggest problem with self-deletions. I only use it here as a guide to figuring out what's the shortest deletion cooldown that still makes sense to have.

gravitystorm:
> This is to avoid the situation when a user is "done", but their changesets are not "done".

Can you please give an example of a **specific problem** caused when a user is deleted before their final changeset closed time has passed. I don't think I have any examples of why this might be a problem. If there is one then I'm very keen to know more details.

I need to know this, since if something specifically breaks in this situation, then the user should never be deleted when there are changesets with a closed time that has not yet passed (even when the cool-down period is set to 0). Otherwise, if there's no specific problem, then we should remove all mention of open changesets from this PR, as it makes a difficult topic harder to understand.

SomeoneElseOSM:
> Can you please give an example of a specific problem caused when a user is deleted before their final changeset closed time has passed.

Obviously I'm not Anton, but when a user deletes their account when they have open changesets, the changeset feed has one extra user name for them.  For example, a "select distinct id,user_id,user_name from osm_changeset where user_id = 20495462 order by id;" on a ChangesetMD database populated from the changeset feed returns:

         id     | user_id  |   user_name
     -----------+----------+----------------
      143036765 | 20495462 | NIneligi Elity
      143036773 | 20495462 | NIneligi Elity
      143036781 | 20495462 | user_20495462
      143036796 | 20495462 | NIneligi Elity
      143036797 | 20495462 | user_20495462
      143036800 | 20495462 | NIneligi Elity
     
     (snip)
     
      143037002 | 20495462 | NIneligi Elity
      143037007 | 20495462 | NIneligi Elity
      143037008 | 20495462 | user_20495462
      143037010 | 20495462 | NIneligi Elity
     (44 rows)

This user was blocked while they were still active (https://www.openstreetmap.org/user_blocks/15241 ) and then deleted their account.

It's arguable how much _extra_ work this causes the DWG since vandals regularly change their names to evade detection before the final deletion (though not in this case) so the "final deleted name" is just one more.

gravitystorm:
> when a user deletes their account when they have open changesets, the changeset feed has one extra user name for them.

Thanks @SomeoneElseOSM , that's helpful. I had also had a think about how the changeset replication feed behaves, but I'm not super-familiar with that since I haven't used it for anything before. My conclusion was that if a user is deleted before their (final) changesets appear in the feed, then it will still work fine just with a different username - so very similar to what you have described above. And since users can rename themselves (and sometimes nothing to do with vandalism or account deletion) I think that there's no error condition there.

mmd-osm:
I think the changeset replication runs as a minutely job. I even found one example, where the changeset creation entry doesn't include the real user anymore. Presumably, the user was already deleted by the time the job was running.

```
543.osm:  <changeset id="143078978" created_at="2023-10-24T18:43:02Z" open="true" num_changes="0" user="user_20502046" uid="20502046" comments_count="0">
601.osm:  <changeset id="143078978" created_at="2023-10-24T18:43:02Z" closed_at="2023-10-24T19:43:02Z" open="false" num_changes="0" user="user_20502046" uid="20502046" comments_count="0">

```

test/controllers/account/deletions_controller_test.rb
```diff
@@ -0,0 +1,65 @@
+require "test_helper"
+
+class DeletionsControllerTest < ActionDispatch::IntegrationTest
```
gravitystorm, line 3: In general, if we are testing the user interface, we should do that with system tests. These are more readable and maintainable than doing xpath-style selects on the raw documents.

The controller tests should be kept for testing edge case behaviour, e.g. when sending incorrect information to the controller, that the UI wouldn't normally allow. So the accounts_controller_test part is good, since it's checking that people can't post deletions directly to the controller.

I'm well aware that our current test suite has an awful lot of UI testing in controller tests, but for new features we should use system tests. We also have system tests already for account deletion (and not controller tests) so let's stick with that.

app/controllers/account/deletions_controller.rb
```diff
@@ -7,6 +7,8 @@ class DeletionsController < ApplicationController
 
     authorize_resource :class => false
 
-    def show; end
+    def show
+      @allowed_at = current_user.deletion_allowed_at
```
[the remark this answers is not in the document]
AntonKhorev: removed @allowed_at

test/controllers/account/deletions_controller_test.rb
```diff
@@ -0,0 +1,65 @@
...
+      assert_select ".btn", "Delete Account" do
+        assert_select ">@disabled", 0
+      end
+    end
+  end
+
+  def test_past_delay
+    with_user_account_deletion_delay(48) do
+      user = create(:user)
+      (11..15).reverse_each do |n|
```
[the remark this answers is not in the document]
AntonKhorev: One changeset is not enough if we want to make sure that a correct changeset is used in the delay check, but ok, let's create only one changeset.

test/controllers/account/deletions_controller_test.rb
```diff
@@ -0,0 +1,65 @@
...
+      assert_select ".btn", "Delete Account" do
+        assert_select ">@disabled", 0
+      end
+    end
+  end
+
+  def test_delay
+    with_user_account_deletion_delay(48) do
+      user = create(:user)
+      (1..5).reverse_each do |n|
```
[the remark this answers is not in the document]
AntonKhorev: Maybe here it's possible to test correctly with only one changeset, if the `<time>` check is present.

app/models/user.rb
```diff
@@ -407,6 +407,14 @@ def max_changeset_comments_per_hour
     end
   end
 
+  def deletion_allowed_at
```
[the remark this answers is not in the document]
AntonKhorev: added `user.deletion_allowed?`

app/models/user.rb
```diff
@@ -407,6 +407,14 @@ def max_changeset_comments_per_hour
...
+    last_changeset = changesets.first
+    if last_changeset
+      earliest_possible_time = last_changeset.created_at + Settings.user_account_deletion_delay.hours
```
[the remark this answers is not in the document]
AntonKhorev: Added index to changesets.closed_at etc. Setting of `0` won't disable the delay with approach 2 as well.

config/settings.yml
```diff
@@ -53,6 +53,8 @@ api_timeout: 300
 web_timeout: 30
 # Periods (in hours) which are allowed for user blocks
 user_block_periods: [0, 1, 3, 6, 12, 24, 48, 96, 168, 336, 731, 4383, 8766, 87660]
+# Account deletion cooldown period (in hours) since last changeset close; 0 to disable, 24 to make sure there aren't any open changesets when the deletion happens
```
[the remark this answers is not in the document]
AntonKhorev: updated

test/controllers/account/deletions_controller_test.rb
AntonKhorev: moved tests from `test/controllers/account/deletions_controller_test.rb` to `test/system/account_deletion_test.rb`

gravitystorm (approved):

config/settings.yml
```diff
@@ -53,6 +53,8 @@ api_timeout: 300
 web_timeout: 30
 # Periods (in hours) which are allowed for user blocks
 user_block_periods: [0, 1, 3, 6, 12, 24, 48, 96, 168, 336, 731, 4383, 8766, 87660]
+# Account deletion cooldown period (in hours) since last changeset close; -24 to disable, 0 to make sure there aren't any open changesets when the deletion happens
+user_account_deletion_delay: -24
```
tomhughes, line 57: I'm not sure what the answer is but a negative cooldown period is clearly a nonsensical idea so we really need to find some better way of expressing this...

tomhughes (changes requested):

config/settings.yml
AntonKhorev: It's as nonsensical as a future close date.

It's possible to compare the current time with min(close time, current time), but then we throw away the check for open changesets.

AntonKhorev: I can add null as a 'disable' value to settings.yml.

gravitystorm: I'm working on a solution, I should have something to share shortly.

gravitystorm: https://github.com/gravitystorm/openstreetmap-website/commit/7802efc8c986bf0648dccd8f1f80cca631cf3420 - I think the clearest thing is to have two configuration options - one to turn cooldown on and off, and the other to configure the length of the cooldown.

What do you both think? I prefer the separate boolean config, rather than having one option with mixed integer/nil (where 0 and nil will have significantly different behaviours).

AntonKhorev: I already did null.

I don't care if it's null or extra boolean option or anything else.

gravitystorm: OK, let's go for that then.

gravitystorm:
@AntonKhorev thank you for all your work on this, and for making the changes coming from the reviews.

mmd-osm:
@gravitystorm : is this feature already in production? I'm wondering a bit, since https://www.openstreetmap.org/changeset/144808451 was created only 18 hours ago by a now deleted user (at the time of writing this comment).

mmd-osm:
@gravitystorm : good point. I don't seem to find "user_account_deletion_delay" in the Chef config, and this repo comes with a default value of "nil", which corresponds to the feature being inactive. As (another) outside observer, and given how quickly the user could delete themselves, I would assume that the setting hasn't made it into production yet.

SomeoneElseOSM:
My response to the concept was at https://github.com/openstreetmap/openstreetmap-website/pull/4313#issuecomment-1780987772 above.  On IRC, one reply to "what do we actually want to set the period to" was "one week maybe".  Given that we have to pick a number one week doesn't seem unreasonable to me.

Issue #4307, opened 2023-10-21
Prevent mechanical signups by malicious users


SomeoneElseOSM:
### URL

https://community.openstreetmap.org/t/how-about-limit-new-accounts/101656/237

### How to reproduce the issue?

Whilst the rate limiting introduced at https://github.com/openstreetmap/openstreetmap-website/pull/4198 is still working, we're still seeing "entirely mechanical" signups by malicious users.

It should not be possible for someone to script the OSM signup process in this way.  The effect is not directly related to, but the effect is compounded by, https://github.com/openstreetmap/openstreetmap-website/issues/4018 .  A user signing up to OSM must have demonstrate that they are a real human being actively managing the signup process, reading and understanding the contributor terms, etc.  The signups that we're seeing (often of consecutive userids) suggests that this is not the case.

I appreciate that there is pressure from elsewhere (see e.g. https://github.com/openstreetmap/openstreetmap-website/pull/4288 ) to _reduce_ signup interactivity.  This request doesn't necessarily contradict that, provided that the external sites that that pull request is designed to help can provide similar "this really is an interactive signup" information.

### Screenshot(s) or anything else?

A sample of problem userids include 20481051, 20481052, 20481053, 20483963, 20483964, 20483965, 20487767, 20487867, 20487869, 20490488, 20490489, 20490490, 20490858, 20490859, and 20490860 .

tomhughes:
What exactly are you suggesting? We don't have a magic wand...

Presumably you are asking for a captcha? Which will have all the usual problems and probably also be a problem for the ongoing issues around third party authentication.

maayanhaimdulberg:
Tel Aviv is gone from the map. Maybe things like this should be prevented somehow. https://www.openstreetmap.org/#map=14/32.0828/34.7637

tomhughes:
This ticket is about rate limiting signups - limiting edits is being discussed elsewhere though that only really applies to bulk edits as individual edits are almost impossible to prevent in advance.

matkoniecz:
We have #1083 "suppress creating accounts by bots/scripts"

Is this or previous spam wave some share characteristics of accounts that could allow to detect or target malicious users, beyond blanket suppressing of automated account creation?

matkoniecz:
#1083 has old (from 2015!) comment

> Yes, a captcha would be nice, but as far as I know there aren't any non-proprietary, effective captchas available.

has it changed?

tomhughes:
[...]

PR #4314, opened 2023-10-24
Add support for using an cloudflare turnstile widget during signup

tomhughes:
This is one option for implemented #4307 by using Cloudflare's turnstile widget on the signup page.


gravitystorm:
Overall looks good to me. Tests are obviously missing - we'll need some webmock to stub out the cloudflare responses, and we will want to avoid loading the JS from Cloudflare during testing too.

One bit of information from Cloudflare product pages that I noticed:

> "Managed mode: Cloudflare logic determines which users see widget"

I think that means that some users won't see the widget (the JS will just run automatically and put the result into the form). When that happens, the form label "Verify you're a human" will still show. (you can emulate this by putting a bogus `turnstile_site_key` in the config file, don't ask me how I know :smile:).

tomhughes:
I'm not entirely sure but I suspect that phrasing is a bit misleading. I don't think it ever goes into invisible mode rather what changes is whether you have to check the box manually or not.

gravitystorm:
https://developers.cloudflare.com/turnstile/reference/widget-types/ has more details about the available modes, one of which is "invisible", but I agree there's something unclear about the phrasing somewhere.

tomhughes:
Right but that is an advanced option and that same page describes managed mode as:

> Cloudflare will use information from the visitor to decide if an interactive challenge should be used. If we show an interaction, the user will be prompted to check a box (no images or text to decipher).

which is what led me to conclude that what they mean is that it will decide whether to show an _interactive_ widget or not.

1ec5:
Is this draft still relevant? As I understand it, we were moving away from Cloudflare in openstreetmap/chef#812 and the free Turnstile plan is fairly limited, so would we expect to ever deploy this option ourselves? Are there any alternatives we could investigate that serve the same purpose as Turnstile?

tomhughes:
It's not dependant on using cloudflare as a CDN in front of the site so it's not really blocked by us moving away from that. Indeed this PR was I believe created long before we ever used cloudflare in that way.

Obviously other options (recaptcha, hcaptcha, etc) are available but I picked this one because it's less of a pain for end users.

config/locales/en.yml
```diff
@@ -2906,21 +2753,53 @@ en:
...
+      privacy_policy_url: https://wiki.osmfoundation.org/wiki/Privacy_Policy
+      privacy_policy_title: OSMF privacy policy including section on email addresses
+      consider_pd_html: "I consider my contributions to be in the %{consider_pd_link}."
+      consider_pd: "public domain"
+      consider_pd_url: https://wiki.osmfoundation.org/wiki/Licence_and_Legal_FAQ/Why_would_I_want_my_contributions_to_be_public_domain
       or: "or"
       use external auth: "or sign up with a third party"
+      verify_human: Verify you're a human
+    create:
+      not_human: Human verification failed
```
pablobm:
```suggestion
      not_human: Unfortunately, we can't confirm that you are a human. This is required to create new accounts.
```

tomhughes:
I've rebased this and applied @pablobm's suggested change to the message in case we want to consider it.

Firefishy:
I've moved the [OSM wiki from using hcaptcha to Cloudflare Turnstile](https://github.com/openstreetmap/chef/commit/5c2020cfdb009a34333904fcc0773462e64a4ebc) to test out turnstile and to expose any potential issues with us using it.

Firefishy:
[Cloudflare Turnstile is not supported in Mainland China](https://developers.cloudflare.com/china-network/faq/#is-turnstile-available-in-mainland-china). We might want to allow sign-ups to OpenStreetMap from Mainland China without Turnstile, but this should not be a blocker for initially releasing of this feature.

I have not found evidence of Turnstile being blocked in any other regions.

Firefishy:
The wiki has been using Cloudflare Turnstile now for nearly 3 weeks, I haven't seen any complains and it seems to be working well. 

I'm keen for us to add a sign-up CAPTCHA to prevent scripted sign-up abuse, which allows attackers to trigger large volumes of confirmation emails to third-party addresses (email flooding / email bombing). We've had email delivery issues with Hotmail, Live.com and Outlook in the past because our systems were abused in this way.

Firefishy:
@hlfan @gravitystorm Any chance you could review this when you have a chance? It will potentially help with the operational issue scripted sign-ups used by spammers and other bad actors.

PR #4316, opened 2023-10-25
Set max length of block reason

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/4301#discussion_r1367692119


tomhughes:
Some basic statistics:

* The longest current reason is 7129 bytes
* The 99th percentile is 2668 bytes
* The 95th percentile is 1691 bytes

gravitystorm:
Merged, thanks!

PR #4318, opened 2023-10-28
Enable remaining rails 7.1 defaults

tomhughes:
This enables all the remaining rails 7.1 defaults in addition to the ones I accidentally pushed yesterday - apologies for that as I had meant them all to be in this pull request.

The only issue was caused by the new marshalling format which required a workaround due to https://github.com/rails/rails/issues/49826 which was in turn hidden until the final cut over by https://github.com/rails/rails/issues/49827.


gravitystorm:
LGTM thanks. I'd seen the HTML5 parser stuff in a Rails World talk, glad to get that merged.

PR #4319, opened 2023-10-29
Add the ability to rate limit edits

tomhughes:
This aims to implement #2342 by rate limiting edits.

The limits work in a similar way to the changeset comment limits and are configurable with the defaults being a ramp up from 1000 changes per hour to 100000 per hour over the first week after a user starts editing. Users that are reported lose some allowance subject to a minimum that defaults to 100 edits per hour.

The increase is non linear, increasing as the square of time elapsed by subject to the lower bound of the initial limit so that it starts to climb after about 18 hours as shown in this graph which shows hours on the X axis versus allowed edits per hour on the Y axis.

Screenshot from 2023-10-29 17-33-10

Moderators have a separately configurable rate limit and this also introduces a new `importer` role that has a higher rate limit and which can be granted to accounts doing large imports/bulk edits.

Obviously once this agreed cgimap will need work to respect the rate limits.


mmd-osm:
> Obviously once this agreed cgimap will need work to respect the rate limits.

It would probably be quicker to handle all non-moderator changeset uploads on the Rails port for a while and get some experience with the approach first. Working on both rails and cgimap is likely slowing things down too much.

zstadler:
Will it be possible to use the accumulated number of changes rather than the elapsed time as the measure by which the rate is increased?

I can see a malicious account creatingbg a benign changeset, waits a week doing nothing suspicious and then receives a decent rate. Providing an initial 10000 benign edits, working for a period of at least 10 hours,  in order to get a rate increase is much harder to achieve. On the other hand, I do not believe that would be unacceptable for new inexperienced users.

tomhughes:
I can see it would be better - my concern is the cost of computing that value as it means reading all the changesets for that user unless we do something to cache the value.

zstadler:
Yes, it would be reading the changesets, not the change files.
It is not fundementally different than what's done in `https://www.openstreetmap.org/user/`.

It would be useful to cache the open changesets. Still, _Perfect is the enemy of good_

tomhughes:
> It is not fundementally different than what's done in `https://www.openstreetmap.org/user/`.

I don't understand what you mean? Nothing on the user page shows the number of changes - the number of changesets yet but not the number of objects changed.

> It would be useful to cache the open changesets. Still, _Perfect is the enemy of good_

What does the number of open changesets have to do with anything?

tomhughes:
> It would probably be quicker to handle all non-moderator changeset uploads on the Rails port for a while and get some experience with the approach first. Working on both rails and cgimap is likely slowing things down too much.

There's no way for apache to know if the user is a moderator or not so we can't make redirects conditional like this and in any case the performance impact of doing that is likely to be substantial.

zstadler:
I'm deeply sorry I was not expressing myself clearly.

> > It is not fundementally different than what's done in `https://www.openstreetmap.org/user/`.
> 
> I don't understand what you mean? Nothing on the user page shows the number of changes - the number of changesets yet but not the number of objects changed.

I meant to say that [ GET /api/0.6/changesets](https://wiki.openstreetmap.org/wiki/API_v0.6#Query:_GET_/api/0.6/changesets) can provide the information needed to compute the total number of changes made by an account.

This is in contrast to an alternative approach of reading and analyzing the associated osmChange data using [GET /api/0.6/changeset/#id/download](https://wiki.openstreetmap.org/wiki/API_v0.6#Download:_GET_/api/0.6/changeset/#id/download) calls for each changeset.

I was assuming the "edit" link on the users page is using the [ GET /api/0.6/changesets](https://wiki.openstreetmap.org/wiki/API_v0.6#Query:_GET_/api/0.6/changesets) API point.

> > It would be useful to cache the open changesets. Still, _Perfect is the enemy of good_
> 
> What does the number of open changesets have to do with anything?

I meant to say that if there is a cache for the total number of changes made by each user, at any given moment, the number of useful lines in the the cache would be at the same order of magnitude as the number of open changesets in the system.

tomhughes:
Well sure the data is available (which is all the API being able to fetch it proved) but we put limits on that API call - we don't just allow people to fetch a user's entire edit history because of the cost of doing that.

Talking about the API is irrelevant anyway because the server code doesn't use the API or parse XML to do things like this, it just executes `current_user.changesets.sum(:num_changes)` but that will turn into something like:

```sql
SELECT SUM("changesets"."num_changes") FROM "changesets" WHERE "changesets"."user_id" = $1
```

which forces the database to read all the changeset records for the user.

I doubt the edit link uses that API call either - why would iD care about your previous changesets?

drolbr:
> Talking about the API is irrelevant anyway because the server code doesn't use the API or parse XML to do things like this, it would just execute `current_user.changesets.sum(:num_changes)` but that will turn into something like:
> 
> ```sql
> SELECT SUM("changesets"."num_changes") FROM "changesets" WHERE "changesets"."user_id" = $1
> ```

Can we take advantage of `changesets_user_id_created_at_idx` and look only into the last hour or last 24 hours or so?

tomhughes:
We'd certainly use that index yes (in fact `current_user.changesets` does but the sum causes the order to be lost) but limiting by date is probably not a good idea as it would penalise people with lots of infrequent edits.

Better is probably only to look at the last 100/200/1000 or whatever changesets.

mmd-osm:
> There's no way for apache to know if the user is a moderator or not so we can't make redirects conditional like this and in any case the performance impact of doing that is likely to be substantial.

So for our DWG members using _osmtools_ to revert changesets, there won't be any change, since they're using the single object upload on Rails anyway. I'm thinking about whitelisting some IP addresses used by specialized revert tools, such as Thanos. This of course assumes that the tool would be somehow restricted to mods only. (ping @Zaczero).

For sure there will be a performance impact for everyone else. As long as our revert tools can't really cope with the data volume, and we're still trying to work out a good strategy to identify excessive uploads, this seems acceptable to me as a temporary measure. Are you concerned that spike-0[678] can't cope with the additional load?

tomhughes:
Well I'm more concerned about slowing down upload calls than about literally overloading the servers.

AntonKhorev:
> since they're using the single object upload on Rails anyway

not necessarily

app/controllers/api/changesets_controller.rb
```diff
@@ -92,6 +92,10 @@ def upload
       diff_reader = DiffReader.new(request.raw_post, changeset)
       Changeset.transaction do
         result = diff_reader.commit
+        # the number of changes in this changeset has already bee
```
matkoniecz, line 95:
```suggestion
        # the number of changes in this changeset has already been
```

app/models/user.rb
```diff
@@ -407,6 +413,22 @@ def max_changeset_comments_per_hour
     end
   end
 
+  def max_changes_per_hour
+    if moderator?
+      Settings.moderator_changes_per_hour
+    elsif importer?
+      Settings.importer_changes_per_hour
+    else
+      first_change = changesets.reorder(:created_at => :asc).pick(:created_at) || Time.now.utc
```
matkoniecz, line 422: would it make sense to have time since first edit, after expiration of last block (if user was ever blocked)

That could help with accounts that mass-imported stuff, got block/zero hour block and then were importing things again on large scale shortly after block expired

(assuming that it is really easy to get latest block expiration and it has no noticeable performance impact)

test/helpers/user_roles_helper_test.rb
```diff
@@ -82,6 +137,24 @@ def test_role_icons_administrator
       <a data-confirm="Are you sure you want to revoke the role `moderator&#39; from the user `#{moderator_user.display_name}&#39;?" rel="nofollow" data-method="post" href="/user/#{ERB::Util.u(moderator_user.display_name)}/role/moderator/revoke">
       <img srcset="/images/roles/moderator.svg" border="0" alt="Revoke moderator access" title="Revoke moderator access" src="/images/roles/moderator.png" width="20" height="20" />
       </a>
+      <a data-confirm="Are you sure you want to grant the role `importer&#39; to the user `#{moderator_user.display_name}&#39;?" rel="nofollow" data-method="post" href="/user/#{ERB::Util.u(moderator_user.display_name)}/role/importer/grant">
+      <img srcset="/images/roles/blank_importer.svg" border="0" alt="Grant importer access" title="Grant importer access" src="/images/roles/blank_importer.png" width="20" height="20" />
+      </a>
+    HTML
+    assert_dom_equal expected, icons
+
+    importer_user = create(:importer_user)
```
matkoniecz, line 146: is `importer_user` intentional here given that later we have `Are you sure you want to grant the role `administrator&#39` ?

And `role/administrator/grant` ?

Zaczero:
I believe a 1000 changes per hour is too low. New users should be able to create at least 1 full changeset (10_000 changes) to avoid confusion. However, they should not be able to create 20_000 changes. Perhaps make it a daily limit instead of hourly, starting from max 20_000 changes per day.

tomhughes:
I think allowing 10000 is far too much given what we're trying to combat here - no real user that hasn't dived straight into an import is going to be generating changesets that big.

I have never done a changeset that big in 17 years of editing!

Zaczero:
I personally dislike the new importer role; everyone should be able to perform imports without much of an issue. In other words, the rate limit should be configured in a way that does not disrupt imports, assuming the account is in good condition.

Zaczero:
> I have never done a changeset that big in 16 years of editing!

It's not hard to imagine a newbie who performs 10_000 changes in a day (and does a single, bulk upload).

Zaczero:
One more argument: reverting 20,000 changes is not difficult, but when it's 2 million or more, it becomes challenging. Whether we save an additional 10,000 changes or not is insignificant and may potentially cause more harm than good.

zstadler:
I request that the requirements will continue to be discussed at the issue - https://github.com/openstreetmap/openstreetmap-website/issues/2342#issuecomment-1784142066

>    * Is it a limit on:
>       * The number of changesets?
>       *The changes_count of the changesets?
>        *The size in bytes of the change data (rate limit)?
>        * or something else?
>    * What are the initial limits?
>    * How should they be increased?

IMHO, the PR is a place to discuss the implementation of these requirements.

gravitystorm:
```
PG::ActiveSqlTransaction: ERROR:  ALTER TYPE ... ADD cannot run inside a transaction block
/app/config/initializers/migrate.rb:55:in `add_enumeration_value'
/app/db/migrate/20231029151516_add_importer_role.rb:3:in `up'
```

tomhughes:
Ah so that is what's going on with docker.... I couldn't see what the problem was but the error messages are weirdly in the middle of the output.

So it seems that restriction was relaxed in postgres 12 which is why the main tests work as the GitHub ubuntu images use postgres 14 but our docker setup still uses postgres 11...

woodpeck:
> So for our DWG members using _osmtools_ to revert changesets, there won't be any change, since they're using the single object upload on Rails anyway. 

While using single object upload is still an option with _osmtools_, the default is now using changeset/upload POST.

gravitystorm:
> So it seems that restriction was relaxed in postgres 12 which is why the main tests work as the GitHub ubuntu images use postgres 14 but our docker setup still uses postgres 11...

Great, thanks for looking into that. I'm happy to bump the requirement to postgres 12+ (since that's the lowest version available in the oldest Ubuntu LTS, 20.04) so it shouldn't affect too many developers/deployments (hopefully!). Should we use 12 in the docker tests to pick anything up that might affect focal environments? I'm happy either way.

Only remaining thing after that is to update INSTALL.md with the new minimum version.

I like that the algorithm is based on the one that we introduced for changeset comments recently, so we might be able to refactor that in future (perhaps initially moving all the user rate limits into a concern, to avoid giant `user.rb`). I'm also happy with the concept of the `importer` flag, in order to provide a pragmatic compromise for these accounts (and everyone already follows the Import Guidelines and uses a separate account, right? :smile: ).

I considered suggesting reworking the tests slightly (to use `Numeric::divmod` and `create_list`) but I don't think it's worth the additional refactoring at this point.

Otherwise I'm happy with this PR, and unless you have any more changes planned then I think it can be merged.

mmd-osm:
I'm not so happy that all the logic is now implemented in Rails, which means that we have almost zero reuse potential and need to re-implement exactly the same algorithm again on CGImap. 

I'd be happier to see either 
* have a REST endpoint which I could call from CGImap, or
* some postgresql database function I could call
both of which options could return some true/false kind of information, if the changeset is acceptable.

Then you could tweak both settings and algorithms as much as needed without ever having to touch CGImap.

tomhughes:
What would the REST endpoint return exactly? We don't really want to be having to make REST call from cgimap to rails on every changeset upload do we?

It would certainly be possible to put the limit check in a function in the database - how to configure it is an interesting question though as you would probably have to put the configuration values in a settings table or pass them all in on every call.

mmd-osm:
Both options have their pros/cons. A REST endpoint looks more like a proper interface to me. It could return some further details in case a changeset cannot be accepted, including a proper error message. Of course, it will introduce much more overhead compared to calling a function on the database. On the positive side, you could write all logic in RoR.

Regarding the database function, passing a few parameter values around wouldn't be much of an issue. If things start getting more complicated, that approach doesn't scale well. A settings table might be better in the long run.

lonvia:
If you define a Postgres function you can simply hard-code the configuration parameters into the function implementation using a bit of string replacement magic and have RoR recreate the function whenever you want to change the parameters.

tomhughes:
That's a possibility I guess though it means managing the function outside the normal database migration logic.

I've actually already written an plpgsql implementation with the configuration hard coded:

```sql
CREATE OR REPLACE FUNCTION check_rate_limit(user_id int8, new_changes int4)
  RETURNS boolean
  AS $$
DECLARE
  min_changes_per_hour int4;
  initial_changes_per_hour int4;
  max_changes_per_hour int4;
  importer_changes_per_hour int4;
  moderator_changes_per_hour int4;
  roles text[];
  last_block timestamp without time zone;
  first_change timestamp without time zone;
  active_reports int4;
  time_since_first_change double precision;
  max_changes double precision;
  recent_changes int4;
BEGIN
  min_changes_per_hour := 100;
  initial_changes_per_hour := 1000;
  max_changes_per_hour := 100000;
  importer_changes_per_hour := 1000000;
  moderator_changes_per_hour := 1000000;

  SELECT ARRAY_AGG(user_roles.role) INTO STRICT roles FROM user_roles WHERE user_roles.user_id = check_rate_limit.user_id;

  IF 'moderator' = ANY(roles) THEN
    max_changes := moderator_changes_per_hour;
  ELSIF 'importer' = ANY(roles) THEN
    max_changes := importer_changes_per_hour;
  ELSE
    SELECT user_blocks.created_at INTO STRICT last_block FROM user_blocks WHERE user_blocks.user_id = check_rate_limit.user_id ORDER BY user_blocks.created_at DESC LIMIT 1;

    IF FOUND THEN
      SELECT changesets.created_at INTO STRICT first_change FROM changesets WHERE changesets.user_id = check_rate_limit.user_id AND changesets.created_at > last_block ORDER BY changesets.created_at LIMIT 1;
    ELSE
      SELECT changesets.created_at INTO STRICT first_change FROM changesets WHERE changesets.user_id = check_rate_limit.user_id ORDER BY changesets.created_at LIMIT 1;
    END IF;

    SELECT COALESCE(SUM(issues.reports_count), 0) INTO STRICT active_reports FROM issues WHERE issues.reported_user_id = check_rate_limit.user_id AND issues.status = 'open';

    time_since_first_change := EXTRACT(EPOCH FROM NOW() - first_change);

    max_changes := max_changes_per_hour * POWER(time_since_first_change, 2) / POWER(7 * 24 * 60 * 60, 2);
    max_changes := GREATEST(initial_changes_per_hour, LEAST(max_changes_per_hour, FLOOR(max_changes)));
    max_changes := max_changes / POWER(2, active_reports);
    max_changes := GREATEST(min_changes_per_hour, LEAST(max_changes_per_hour, max_changes));
  END IF;

  SELECT COALESCE(SUM(changesets.num_changes), 0) INTO STRICT recent_changes FROM changesets WHERE changesets.user_id = check_rate_limit.user_id AND changesets.created_at >= now() - '1 hour'::interval;

  RETURN recent_changes + new_changes < max_changes;
END;
$$ LANGUAGE plpgsql;
```

DavidKarlas:
If plpgsql approach doesn't work out... or if future challenges that more sophisticated vandals will throw at OSM outgrow limits of plpgsql...

I wonder if rails needs anti-vandals mechanism at all since vandalism is probably only problem for production instance of osm.org, which will always need CGIMap to handle upload APIs... hence implementing anti-vandalism logic just on CGIMap is good enough...

Just an idea...

tomhughes:
Only diff uploads go to cgimap currently - single object changes go via rails.

mmd-osm:
Releasing and deploying a new CGImap version involves quite a number of manual steps. That’s not ideal when you’re expecting lots of smaller adjustments to happen. So both options are meant to get rid of this overhead. 

I also thought about the limits of what we could do in a db function. For the time being I would consider it a good option to make quick progress on this issue. It may not be the final solution, but that’s ok i think.

tomhughes:
Well that's the point of parametrising the algorithm so that we don't have to keep doing releases because any tuning is done by changing the configuration options.

That said I'm happy to go with the SQL function if I can come up with a good way of managing it.

mmd-osm:
I signed up as a new user,  tried to upload some data, and saw the following error message. Checking now...

select * from check_rate_limit(13,10000);
ERROR:  query returned no rows
CONTEXT:  PL/pgSQL function check_rate_limit(bigint,integer) line 29 at SQL statement

tomhughes:
What version of postgres are you running? It was fine for me in 15.

mmd-osm:
It's not the latest and greatest: ... server 12.12 (Ubuntu 12.12-1.pgdg20.04+1)

mmd-osm:
Once I comment out the user_blocks part, I'm no longer able to upload 10'000 changes as a new user:

image

mmd-osm:
Yup, this one works fine.

mmd-osm:
Could we also return the maximum number of changes from the db function that are currently allowed?  This would really to generate a meaningful error message.

tomhughes:
I've pushed an update to use an SQL based implementation - the SQL function is slightly different to the previous version as it is now called `api_rate_limit` and takes only a user ID as argument and returns the maximum number of changes that can currently be committed.

There's a migration to add the function and also a `db:update_functions` rake task that will recreate it if the settings are changed.

lib/database_functions.rb
```diff
@@ -0,0 +1,56 @@
...
+        SELECT user_blocks.created_at INTO last_block FROM user_blocks WHERE user_blocks.user_id = api_rate_limit.user_id ORDER BY user_blocks.created_at DESC LIMIT 1;
+
+        IF FOUND THEN
+          SELECT changesets.created_at INTO first_change FROM changesets WHERE changesets.user_id = api_rate_limit.user_id AND changesets.created_at > last_block ORDER BY changesets.created_at LIMIT 1;
+        ELSE
+          SELECT changesets.created_at INTO first_change FROM changesets WHERE changesets.user_id = api_rate_limit.user_id ORDER BY changesets.created_at LIMIT 1;
+        END IF;
+
+        IF NOT FOUND THEN
+          first_change := NOW();
```
mmd-osm, line 37: As mentioned in a previous comment, could you change all NOW() in this db function  to `now() at time zone 'utc'` to facilitate testing in non utc-0 environments.

db/structure.sql
```diff
@@ -107,6 +108,65 @@ CREATE TYPE public.user_status_enum AS ENUM (
...
+
+        time_since_first_change := EXTRACT(EPOCH FROM CURRENT_TIMESTAMP AT TIME ZONE 'UTC' - first_change);
+
+        max_changes := max_changes_per_hour * POWER(time_since_first_change, 2) / POWER(days_to_max_changes * 24 * 60 * 60, 2);
+        max_changes := GREATEST(initial_changes_per_hour, LEAST(max_changes_per_hour, FLOOR(max_changes)));
+        max_changes := max_changes / POWER(2, active_reports);
+        max_changes := GREATEST(min_changes_per_hour, LEAST(max_changes_per_hour, max_changes));
+      END IF;
+
+      SELECT COALESCE(SUM(changesets.num_changes), 0) INTO STRICT recent_changes FROM changesets WHERE changesets.user_id = api_rate_limit.user_id AND changesets.created_at >= now() - '1 hour'::interval;
```
mmd-osm, line 163: Another `now()` hiding in this line towards the end...

lib/database_functions.rb
pnorman: Is this because the DB columns are `timestamp` instead of `timestamptz` like `now()`?

mmd-osm: Probably yes. The database function wasn't working properly without "at time zone 'utc'".

tomhughes:
Hopefully all the time zone issues are fixed now and I've also refactored the report counting to match the changes made in the rails code for 4320.

DavidKarlas:
I wonder if there should be new table in database called "rate limited users/changesets" which would contain user id, # of attempted edits, # of max edits at time(returned by SQL function), and timestamp. Which would be visible to role "moderator" via website(maybe everyone?)? So they could grant users import role or block them. This would also give DWG insight and ability to decide to raise min limit from 1000 to1500 as an example, or go with 14 days ramp...

mmd-osm:
>  the SQL function is slightly different to the previous version as it is now called api_rate_limit and takes only a user ID as argument and returns the maximum number of changes that can currently be committed.

Good idea. Now it's even feasible to include additional details in the error message. As said, with JOSM's chunk-based upload, the max changes value can be easily determined anyway. On the other hand, additional details make it easier to help users who would be otherwise stuck.

Maybe we could also include this information in /user/details, to give editing apps a chance to be aware of these limits even before  trying to upload anything.



CGImap is already updated.

simonpoole:
> Now it's even feasible to include additional details in the error message.

@tomhughes & @mmd-osm have you finalized the error code and the message yet (because as you know we have to parse the text to extract the relevant parts for display)?

tomhughes:
As far as rails is concerned you just get a 429 error. I don't propose to include any further details.

mmd-osm:
@simonpoole : what would be your preferred option?

simonpoole:
> @simonpoole : what would be your preferred option?

No strong feelings, on the one hand withholding any useful information from vandals is good, on the other hand it means that somebody that inadvertently runs it to the limit (aka an importer) is left alone. I would suggest that even if we don't give details, the message includes a pointer to the DWG (for potentially getting the import vetted).

mmd-osm:
I'm a bit concerned about first time contributors working on HOT tasks in particular, that might not be prepared for this kind of message. So they will likely post a screenshot of the error message somewhere, asking on how to proceed. If the screenshot tells you something about 890 changes to be uploaded, while the limit would be 800, you could tell the person to simply try again in a few hours. In case they're trying to upload 10'000 changes as a beginner, sending them to DWG is of course the most reasonable thing to do.

Example: https://www.openstreetmap.org/changeset/143552081

simonpoole:
> You have exhausted your current upload allowance, please contact data@openstreetmap.org before trying again.

I've heard some grumbling from the depths of the DWG :-), so maybe I won't do exactly that. The problem with the HOT scenario is that there is not much that can be done (for example pointing to the import guidelines is just going to be worse) outside of asking HOT to prep people for this.

DavidKarlas:
Link to OSM Wiki page explaining things would also be very useful IMO.

mmd-osm:
Like _Visit wiki.osm.org/ratelimit for more details._ ?

DavidKarlas:
It's a start... :) wiki.osm.org/Error429

mmd-osm:
May I suggest that we create some Wiki page discussing the changeset upload rate limiting only. Overpass API (not turbo!) may also return an 429 error at times,but that's really distracting here and fairly off topic.

mmd-osm:
I've added a fair amount of ChatGPT generated content to explain the general concept of rate limiting, and also added some hints that we're offering "import" accounts.

mmd-osm:
Ok, this wiki thing wasn't really a good idea. I'm changing the text to `Upload has been blocked due to rate limiting. Please try again later.` now. Comments? -> https://github.com/zerebubuth/openstreetmap-cgimap/pull/297

app/controllers/api_controller.rb
```diff
@@ -192,4 +192,14 @@ def api_call_timeout(&block)
     ActiveRecord::Base.connection.raw_connection.cancel
     raise OSM::APITimeoutError
   end
+
+  ##
+  # check the api change rate limit
+  def check_rate_limit(new_changes = 1)
+    max_changes = ActiveRecord::Base.connection.select_value(
```
mmd-osm, line 199: It would be good if we could enable/disable this check based on some global setting. 

This way we could merge the code sooner and enable it on the dev instance for testing.

tomhughes:
I've deployed the branch to https://tomh.apis.dev.openstreetmap.org/ if you want to test it.

mmd-osm:
Thanks! I think tomh was one of the Rails only sites. Any chance we could include cgimap in the tests? It needs a new parameter to enable rate limiting.

tomhughes:
Hopefully that is done now...

mmd-osm:
Thanks for the quick config update. First test on tomh looks good here:

Chunk size = 10k: fails right away
image

Continue testing with chunk size 100: up to 900 changes can be uploaded:
https://tomh.apis.dev.openstreetmap.org/changeset/42

Then continue uploading in single object mode (->uses Rails implementation):

image

mmd-osm:
So the next test would be the moderator role, and of course the new importer role.
I've created this new user for this purpose: https://tomh.apis.dev.openstreetmap.org/user/mmd2mod
Could you please give this user moderation right?
I think that's all what is needed for me to toggle the importer role on "mmd2" later on.

tomhughes:
I've set moderator on that and importer on mmd2. I think currently you have to be admin to change either though we might want to change that for the importer role.

mmd-osm:
Ok, as an import user mmd2, I could successfully upload 999'999 changes now.

mmd-osm:
Same test with mmd2mod: 1 million changes successfully uploaded before hitting rate limit. I think the 999'999 for mmd2 was probably some mistake on my end when calculating the total number.

mmd-osm:
Next test with user mmd3, first time contributor with open community feedback. -> also ok.

mmd-osm:
@tomhughes : new build will be published soon: https://launchpad.net/~mmd-osm/+archive/ubuntu/cgimap-810-jammy/+packages

mmd-osm:
About two days after my last edit, I tried again to upload some changes, this time in iD. Some 6000+ changes went through before hitting rate limiting. Error message in iD also looks ok to me:

image

HarelM:
First and foremost, thanks for all the effort put in this PR, and in general to improve the situation.

I still believe using edits (number of changesets) instead of time would be aligned with what we want to achieve in the concept of allowing to edit more the more you edit, and not more as more time passes.
I think it can also simplify the code...?

As to @Zaczero comments, and this is mainly copying from my comment [here](https://github.com/openstreetmap/openstreetmap-website/issues/2342#issuecomment-1789728144)
> It's not hard to imagine a newbie who performs 10_000 changes in a day (and does a single, bulk upload).

It's really hard for me, I haven't seen any noob in our community does that, seems like a use case I can't understand, please enlighten me, honestly, no cynicism here. 

>One more argument: reverting 20,000 changes is not difficult, but when it's 2 million or more, it becomes challenging.
Whether we save an additional 10,000 changes or not is insignificant and may potentially cause more harm than good.

I can't stress this enough, but we are still working on reverts for all the vandalism that was intruded in the last month in Israel, super painful, super error prone and far-far from being easy.
Moreover, veteran editors, who are editing for years, have said that they don't feel confident continue editing while vandalism is so disruptive and there isn't the relevant security in place to prevent it.
So, if you ask me, yes, prevent from new users to upload these massive changeset and make sure we are encouraging veteran users to continue editing.

HarelM:
I couldn't find the error message text in this PR, but wouldn't it be better to say "See if you can split this edit to smaller chunks" or something similar instead of "try again later" to better guide the user on what to do in order to overcome the issue?

tomhughes:
Splitting into chunks isn't going to help if they then try and upload the same number of edits.

Look the point here is to add limits, not teach people how to evade them!

talllguy:
Importer role sounds like a good plan to me as a DWG member. Lots of importers are just oblivious to the wiki rules. Maybe they did a little Googling about OSM and set off to import some GIS data. Forgive me if this is covered elsewhere, but is there a process by which someone could be promoted to importer? Would they fill out a form showing the read the wiki and contacted the community, etc.? How would that work?

mmd-osm:
I'm not aware that this process has been defined. Currently, only administrators can grant importer rights. As mentioned above, it  is likely to change in the future to include moderators. In any case, the DWG could already start thinking about the kind of paperwork it wants to see from prospective importers. Maybe it's sufficient to add a section to https://wiki.openstreetmap.org/wiki/Import/Guidelines mentioning that importers need to get in touch with DWG and apply for an importer role.

Fizzie41:
> > It's not hard to imagine a newbie who performs 10_000 changes in a day (and does a single, bulk upload).
> 
> It's really hard for me, I haven't seen any noob in our community does that, seems like a use case I can't understand, please enlighten me, honestly, no cynicism here.

I had a DWG case a few months ago in Africa, where a group of people who had never touched it before were shown this is OSM, & this is how you add things, now go to it & start editing the map for this Government program!. Unfortunately, some members of the group were then apparently having a competition to see how quickly they could add 10000 random, untagged nodes to the map :-(

matkoniecz:
> then apparently having a competition to see how quickly they could add 10000 random, untagged nodes to the map :-(

that is supporting lower rate limits for new mappers, not going against it

Fizzie41:
> > then apparently having a competition to see how quickly they could add 10000 random, untagged nodes to the map :-(
> 
> that is supporting lower rate limits for new mappers, not going against it

Yep, that's what I would like to see!

zstadler:
Thank you @tomhughes for creating this PR and to everyone who helped in maturing and testing it!

When can the community expect it to be deployed to the production server?
Are there other PRs that need a deployment for this rate limit to be operational?

gravitystorm (approved):
Other than the enum suggestion (which can be left for later, it's not a blocker) then I'm happy with this. It's a shame that the logic has to go in as a database function, since that's harder to iterate/debug/lint etc, but I think it's a good solution.

I'll leave this for @tomhughes to merge when he is ready, due to the operational considerations.

AntonKhorev:
> It's a shame that the logic has to go in as a database function

Why not do it entirely in cgimap?

tomhughes:
> Why not do it entirely in cgimap?

Because we don't want to duplicate the logic in rails and cgimap.

simonpoole:
> > Why not do it entirely in cgimap?
> 
> Because we don't want to duplicate the logic in rails and cgimap.
To avoid more back and forth :-)

Because we need it in both rails and cgimap because the single object uploads are only present in the rails part of the API.

AntonKhorev:
> Because we need it in both rails and cgimap because the single object uploads are only present in the rails part of the API.

Why not add single object uploads to cgimap? It's [doable](https://github.com/zerebubuth/openstreetmap-cgimap/pull/300).

mmd-osm:
As mentioned in that pull request, I don't think it a good fit for cgimap, and I will not consider this option going forward. It's simply not worth the effort.

501Ghost:
Hi. I just want to say that I've managed to hit the rate limit with https://www.openstreetmap.org/changeset/144568191 and subsequent changesets. It kicked in after 100k changes. An hour later I was able to upload the remaining 102k changes of my huge changeset without hitting the rate limit. 

I'm happy that the rate limit exists, and the number is more than reasonable, but I was a bit surprised by it, and stressed out because I was afraid it would lead to strange editing issues/conflicts. Luckily that didn't happen and everything went smoothly.

Thank you for this nice implementation of the rate limit that didn't mess up my oversized edit.

anthaas:
Hi, in a mapathon in Brno, Czechia lots of attendees are new mappers creating their OSM accounts for this mapathon. They are mapping https://tasks.hotosm.org/projects/15495/tasks which includes round buildings. The limit is absurd since they are trying to upload 3k+ nodes (mapped in 10 minutes) and they never will be able to do that.
image (1)
image

tomhughes:
Any first time user that is mapping 3000 nodes in ten minutes cannot possibly be doing high quality work - that is 5 nodes a second, every second.

anthaas:
Building tools in JOSM creates 19 nodes for each bulding and it's only two click of the mouse. I can imagine very lively that result meets OSM standards.

woodpeck:
That would still mean 500 clicks of the mouse in 600 seconds - not counting any potential zoom operations. Frankly, this sounds like a recipe for problems, and HOT should be thankful for a rate limit that avoids cementing their bad rep for low-quality edits.

DwynDwe:
It is relatively easy to make this many edits in a short time. It's not necessarily poor quality work, especially when there are validators. Unfortunately, the current restriction discourages new users who want to contribute to openstreetmap.

HarelM:
Taking it a bit more easy and learning the system before making these huge edits makes a lot of sense, IMHO.
I would consider investing more time in teaching these new users how to properly map before making such huge amount of editing.

As I said before, if the lack of security (which was before this limit was introduced) discouraged veteran mappers from editing (which it did) I think there's no real question here which users are more valuable, at least for me it's clear...

matkoniecz:
> The limit is absurd since they are trying to upload 3k+ nodes (mapped in 10 minutes)

@anthaas Can you provide representative sample of edits they are trying to send? JOSM can save not uploaded edits, and it is possible to open them elsewhere (in another JOSM instance). It would allow to look into whether it was good idea to block such edits. Also, you can use it to upload edits from another account (if these edits were worth uploading)

I see some building mapped as round in this area, what maybe contributes to larger volume. But first time user that is mapping 3000 nodes in ten minutes? I am really curious about data quality. If that is typical problem, can you make and post screen capture of editing activity of such user? How they edit to end with that?

(I tried to look at edit quality in this area in edits done by HOT mappers but HOT tasking manager seems to not be making easy to review edits done with it, I ended creating https://github.com/hotosm/tasking-manager/issues/6147 https://github.com/hotosm/tasking-manager/issues/6148 https://hotosm.slack.com/archives/C4GLC45PY/p1701330196012119 before running out of time )

mmd-osm:
@matkoniecz : maybe try [osmcha.org](https://osmcha.org/changesets/144600179?filters=%7B%22comment%22%3A%5B%7B%22label%22%3A%2215495%22%2C%22value%22%3A%2215495%22%7D%5D%2C%22date__gte%22%3A%5B%7B%22label%22%3A%222023-11-01%22%2C%22value%22%3A%222023-11-01%22%7D%5D%7D) using changeset comment "#hotosm-project-15495". Some of the larger changesets have a fair amount of round buildings, like this one: https://overpass-api.de/achavi/?changeset=144010949&relations=true - it seems these buildings have been mostly created by copy & pasting an existing round building, which takes almost no time at all in JOSM.

zstadler:
It is very unlikely that a new user will do their edits with JOSM.
The [above example](https://www.openstreetmap.org/changeset/144010949) is the user's 50'th edit.
In his first hour of OSM editing he created about 700 elements. That was also for an hotosm project.

Like @matkoniecz,, I would like to see a real-world new user that hits the current rate limit.
Like @HarelM, I'd rather limit the rate of contribution for new users than have veteran mappers leave OSM. Not to mention the toll on the DWG members and the community members that fights vandalism.

Personally, I've spent more than 3 weeks of work in reverting the latest vandalism in Israel as well as recovering from well-meaning attempts to re-map the affected areas. I did that knowing this PR is in the works.

anthaas:
@matkoniecz  I don't like the idea of uploading changes for other users and taking their credits. That is not the idea behind mapathons. I believe you are familiar with the workflow of humanitarian mapping using the Tasking Manager. So when a user locks a task and tries to upload mapped buildings, the next step is marking the task as mapped and locking another one. This workflow is broken by this limit since the mapper shouldn't mark the task as mapped in the Tasking Manager because changes aren't uploaded yet.

Standard workflow of the mapping is to sweep the whole task, map buildings (or other features) and submit the work. In the JOSM training session, a buildings tools plugin and mapathoner plugin is introduced to ensure buildings are squared or rounded and tagged properly. And yes, we are having a JOSM training even for the new mappers. Training session usually looks like a 1 hour live demo by an experienced mapper on how to identify buildings and how to map followed by Q&A and individual mapping.

[mappingInOneMinute.osm.txt](https://github.com/openstreetmap/openstreetmap-website/files/13515978/mappingInOneMinute.osm.txt)
I’m attaching a file demonstrating a one minute test of the mapping.

SomeoneElseOSM:
To address some of the points above:

Where new users map their first changesets very quickly, and especially when they do so in JOSM, there is a clear correlation with poor-quality editing.  A significant subset of DWG tickets created about new users have simply been about poor editing quality, and our advice is often "slow down, and make smaller but better quality edits".  This has been more noticeable since the recent waves of vandalism where large number of objects were edited (or deleted) in a short period of time - we (the DWG) have been using various means to detect this, and in a few cases have found ourselves dealing with poor quality new editors who just needed to be encouraged to slow down a bit.

To address @zstadler's point above - these are "real-world new users that hit the current rate limit", but even though they're not mapping in bad faith, it still makes sense for them to be asked to slow down a bit.  As an example from the changeset linked above, https://www.openstreetmap.org/way/1223453025 has been mapped as "perfectly circular" with 30 nodes; the imagery suggests perhaps something closer to a hexagon.

mmd-osm:
I suppose buildings in the area are really circular, like shown on this example: https://upload.wikimedia.org/wikipedia/commons/4/43/House_in_Toteil_002.jpg ... we just don't have an established tagging as a node with building=african_hut and diameter=xyz, so people end up creating ways with lots of nodes to approximate the shape of the building.

matkoniecz:
https://community.openstreetmap.org/t/the-new-rate-limiting-prevents-participants-to-missing-maps-mapathons-from-saving-buildings/106610/ was created

Janabau:
> Taking it a bit more easy and learning the system before making these huge edits makes a lot of sense, IMHO. I would consider investing more time in teaching these new users how to properly map before making such huge amount of editing.
> 
> As I said before, if the lack of security (which was before this limit was introduced) discouraged veteran mappers from editing (which it did) I think there's no real question here **which users are more valuable, at least for me it's clear...**

This is not a question which users are more valuable @mmd-osm. It is not either, or. The open mapping community should enable to map both experienced users and new users. I think we should value people who join the community and want to contribute to humanitarian organizations on the ground, as in this case. Blocking the new ones if they are not doing anything wrong, just using the powerful plugins that experienced mappers train them in, is not particularly welcoming.

mmd-osm:
> This is not a question which users are more valuable @mmd-osm**

@Janabau : Uups, you're misquoting me here. I never said anything like that.

cquest:
As current vandalism is mainly tag related (changing tags on existing objects), or moving objects around (moving nodes), **not counting new nodes without tags** may be a way avoid this kind of problem (if it is possible and simple in the API code).

mmd-osm:
@cquest : I don't see how this would help with vandalism. Instead of moving nodes around, you could simply create lots of new nodes without any tags, and then replace existing way nodes by your new nodes. It's even more difficult to manage now, since you "only" changed 100 ways maaybe, but created havoc nonetheless.

fititnt:
> @mmd-osm 
> I suppose buildings in the area are really circular, like shown on this example: https://upload.wikimedia.org/wikipedia/commons/4/43/House_in_Toteil_002.jpg ... we just don't have an established tagging as a node with building=african_hut and diameter=xyz, so people end up creating ways with lots of nodes to approximate the shape of the building.

I believe your idea worth discussion with editors and map styles. It's a bit off topic for this thread, except that it would reduce need of a large amount of buildings everywhere that are truly round

LaoshuBaby:
> > Taking it a bit more easy and learning the system before making these huge edits makes a lot of sense, IMHO. I would consider investing more time in teaching these new users how to properly map before making such huge amount of editing.
> > As I said before, if the lack of security (which was before this limit was introduced) discouraged veteran mappers from editing (which it did) I think there's no real question here **which users are more valuable, at least for me it's clear...**
> 
> This is not a question which users are more valuable @mmd-osm. It is not either, or. The open mapping community should enable to map both experienced users and new users. I think we should value people who join the community and want to contribute to humanitarian organizations on the ground, as in this case. Blocking the new ones if they are not doing anything wrong, just using the powerful plugins that experienced mappers train them in, is not particularly welcoming.

This may be solved by relying on HOT event organizers to collect users' usernames before starting the mapathon and apply for a more relaxed rate limit for them, and must ensure that there are enough experienced mappers to review the corresponding editors. The current rate limit is definitely not perfect. , and will be more of a headache than the first 50 edits to Wikipedia, but it is difficult to find a simple and effective way to deal with the vandalism while also reduces the impact on new people who edit in friendly attitude.

In addition, 3000 points in 10 minutes, as a mapper with 5 years of experience, I can't do it even with the assistance of AI.

LaoshuBaby:
> Hey folks! I would really appreciate if you could stop attributing this quote from @HarelM to my user. Again, I didn't write anything like that.

Edited in https://github.com/openstreetmap/openstreetmap-website/pull/4319#issuecomment-1837448668

pnorman:
I'm not sure discussion in a merged PR is the best place to hash this out.

emakarov:
> ```
> PG::ActiveSqlTransaction: ERROR:  ALTER TYPE ... ADD cannot run inside a transaction block
> /app/config/initializers/migrate.rb:55:in `add_enumeration_value'
> /app/db/migrate/20231029151516_add_importer_role.rb:3:in `up'
> ```

@gravitystorm @tomhughes I just found this thread because face the same issue, but on Postgres 16 during launch via docker in k8s (performing migrations step). Do you have any recommendation on how to overcome this issue?

Woazboat:
Someone just ran into the rate limit with a new account and it scrambled the buildings they modified. Presumably some of the node movements got blocked by the server.

https://community.openstreetmap.org/t/help-why-did-these-changes-upload-as-a-scrambled-mess/134399

> [cyberutilia](https://community.openstreetmap.org/u/cyberutilia)
> Today I made a new osm account cause I moved back to my home country and wanted a fresh start, yada yada, I started mapping details of my neighborhood in josm (...) It was a 1000 and a hundred or so objects to upload, it said it will upload them in two requests. Seems like just the first request (first 1000) kinda worked and the 100 didn’t, some error showed up, something about rate limit, (...) I wanted to check out the already uploaded changes and I saw all this mess that I don’t know why it happened.

https://www.openstreetmap.org/changeset/170693282

https://imgur.com/a/QhMgOhL

image
image

https://overpass-api.de/achavi/?changeset=170693282

Woazboat:
The default josm upload chunk size is 1000 changes, which is exactly the initial upload limit. That means if someone makes more than 1000 changes with a new account, the first upload will go through while the rest of the changes are blocked, creating a broken partial upload.
Maybe the initial limit should be set to 999 instead? That way the first upload will already be blocked and no partial broken upload is made. That won't really fix the underlying issue, but it would prevent this specific failure case.

matkoniecz:
>  Maybe the initial limit should be set to 999 instead? That way the first upload will already be blocked and no partial broken upload is made. That won't really fix the underlying issue, but it would prevent this specific failure case.

it probably would be a good idea to make a new issue for that, if that change makes sense

comment in old merged PR are likely to be missed

see also https://github.com/openstreetmap/openstreetmap-website/pull/4319#issuecomment-1837528169

PR #4323, opened 2023-11-04
Discard jobs when then encounter a deserialization error

tomhughes:
This attempts to ensure that background jobs which fail to deserialise are discarded rather than retrying multiple times which is unlikely to achieve anything.

The main cause of such errors is database objects being deleted before the job runs - for example somebody adding a friend and then removing them again before the mail job runs causing a failure to find the friendship when the job runs.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
+  self.delivery_job = ActionMailer::MailDeliveryJob
 end
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
-  self.delivery_job = ActionMailer::MailDeliveryJob
+  self.delivery_job = MailDeliveryJob
 end
--- a/app/jobs/mail_delivery_job.rb
+++ b/app/jobs/mail_delivery_job.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-class ApplicationJob < ActiveJob::Base
+class MailDeliveryJob < ActionMailer::MailDeliveryJob
   discard_on ActiveJob::DeserializationError
 end
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -3,8 +3,6 @@
 class UserMailer < ApplicationMailer
   include ActionView::Helpers::AssetUrlHelper
 
-  self.delivery_job = ActionMailer::MailDeliveryJob
-
   default :from => Settings.email_from,
           :return_path => Settings.email_return_path,
           :auto_submitted => "auto-generated"
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -184,7 +184,7 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     assert_difference "User.count", 1 do
       assert_difference "ActionMailer::Base.deliveries.size", 1 do
         post users_path, :params => { :user => user.attributes, :referer => "/edit?editor=id#map=1/2/3" }
-        assert_enqueued_with :job => ActionMailer::MailDeliveryJob,
+        assert_enqueued_with :job => MailDeliveryJob,
                              :args => proc { |args| args[3][:params][:referer] == welcome_path(:editor => "id", :zoom => 1, :lat => 2, :lon => 3) }
         perform_enqueued_jobs
       end


app/mailers/application_mailer.rb
```diff
@@ -1,2 +1,3 @@
 class ApplicationMailer < ActionMailer::Base
+  self.delivery_job = ActionMailer::MailDeliveryJob
```
gravitystorm, line 2: Should this not be `self.delivery_job = MailDeliveryJob` i.e. referencing our custom job class that has the discard_on configuration? Otherwise I don't think anything has changed.

gravitystorm (changes requested):
Overall I'm a bit nervous about this. Having a quick read around, `ActiveJob::DeserializationError` can be triggered by all kinds of connection errors. For example, some developers in https://github.com/getsentry/sentry-ruby/issues/1071 were discussing when they received the following:

> ActiveJob::DeserializationError: Error while trying to deserialize arguments: could not obtain a connection from the pool within 5.000 seconds (waited 5.295 seconds); all pooled connections were in use

I'd hate to think that a db connection hiccup could cause thousands of legitimate jobs to be discarded! 

There's a suggestion in that thread that you can catch a different error, e.g.
`discard_on ActiveRecord::RecordNotFound` but that also leaves me worried, since that error might not be from the deserialisation, but instead from some other logic problem in the rest of the job code.

[The docs for `ActiveJob:DeserializationError`](https://edgeapi.rubyonrails.org/classes/ActiveJob/DeserializationError.html) suggest that the original error is available wrapped within that one, so perhaps it might be possible to discard only the jobs that raise a NotFound error during deserialization?

app/mailers/application_mailer.rb
tomhughes: Yes you're quite right.... I'll fix it.

tomhughes:
> [The docs for `ActiveJob:DeserializationError`](https://edgeapi.rubyonrails.org/classes/ActiveJob/DeserializationError.html) suggest that the original error is available wrapped within that one, so perhaps it might be possible to discard only the jobs that raise a NotFound error during deserialization?

I was starting to think that it wasn't possible because there's no way to attach a condition to `on_discard` to inspect the contents but then I looked at how it's implemented and it comes down to https://github.com/rails/rails/blob/6b93fff8af32ef5e91f4ec3cfffb081d0553faf0/activesupport/lib/active_support/rescuable.rb#L90 which, if it doesn't find a handle for `exception` will look for one for `exception.cause` instead which is just what we need!

tomhughes:
Ah except then we have the second problem of discarding jobs which fail to find some other record during execution :-(

gravitystorm:
I'm just coming back to this one, and perhaps there might be a way forward. The [API docs show](https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-discard_on) that `discard_on` can take a block, which allows us to access both the error and the job. So could we use that to only discard jobs if it is an ActiveJob::DeserializationError, but only when that's wrapping an ActiveRecord::RecordNotFound error (and not e.g. connection errors).

```ruby
class MailDeliveryJob < ActionMailer::MailDeliveryJob
  discard_on ActiveJob::DeserializationError do |job, error|
    unless error.is_a?(ActiveRecord::RecordNotFound)
      job.enqueue wait: 5.minutes
  end
end
```

We'd certainly need some tests to make sure this works properly!

I've also [asked on the rails mailing list](https://discuss.rubyonrails.org/t/is-discard-on-activejob-deserializationerror-a-footgun/88442) in case anyone else can shed some light on this. It feels like it could be something that applies more widely.

@tomhughes before I go any further with working on this, how often are you running into this in production? I can't tell how much effort would be worthwhile here.

gravitystorm:
Good news on this topic. One of the 37Signals developers came across the same issue of discarding legitimate jobs during networking blips, and found my post on the rails forum mentioned above. They've worked to get this fixed in rails, and https://github.com/rails/rails/pull/57742 has recently been merged so the fix will be in the next release.

At that point, we can use the following:

```diff
- discard_on ActiveJob::DeserializationError
+ discard_on ActiveJob::DeserializationError::RecordNotFound
```
... and that will work as intended, without any of the nasty side effects. 

In the meantime, I'll mark this as draft and awaiting upstream.

app/mailers/application_mailer.rb
tomhughes: Now I've actually fixed this.

pablobm:
I'm a bit rusty on ActiveJob, and perhaps too exposed to Sidekiq. With that in mind, isn't this a bit of an antipattern anyway? Providing a whole record to a job and leaving the deserialisation to ActiveJob. If the jobs were in charge of the deserialisation, this wouldn't be a problem, as they would know to do an early return and leave it there.

That being said, this appears to be the convention recommended by Rails these days (https://guides.rubyonrails.org/active_job_basics.html#globalid), and I understand that by delegating the deserialisation, it's one fewer duty for us to do manually.

So TIL, I guess.

PR #4324, opened 2023-11-04
Show unredacted element history to moderators

AntonKhorev:
Currently there's no obvious way to view redacted versions. The website won't show them. API-using apps will skip them because redacted versions are missing from element history request results unless a parameter that few people know about is added.

---

History page if logged in as a moderator:
image

After clicking the "View Unredacted History":
image

### DIFF
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -6,6 +6,7 @@ class BrowseController < ApplicationController
   before_action -> { check_database_readable(:need_api => true) }
   before_action :require_oauth
   before_action :update_totp, :only => [:query]
+  before_action :require_moderator_for_unredacted_history, :only => [:relation_history, :way_history, :node_history]
   around_action :web_timeout
   authorize_resource :class => false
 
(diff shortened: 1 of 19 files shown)


app/controllers/browse_controller.rb
```diff
@@ -77,4 +78,10 @@ def changeset
   end
 
   def query; end
+
+  private
+
+  def require_moderator_for_unredacted_history
+    deny_access(nil) if params[:show_redactions] && !current_user&.moderator?
```
tomhughes, line 85: I actually made `nil` the default for `deny_access` earlier today so if you rebase on master you can get rid of the argument here now.

tomhughes (commented):
All looks good really, just one very minor suggestion.

AntonKhorev:
> All looks good really

Actually not because it needs an update for element version pages. Currently it will show redacted versions to anyone who adds `show_redactions=true` to a version page url.

app/controllers/browse_controller.rb
AntonKhorev: This change is not merged yet
https://github.com/openstreetmap/openstreetmap-website/pull/4540/files#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07

tomhughes: Oh sorry about that.. I lost track of which things had and hadn't been merged over the weekend...

AntonKhorev:
Updated for element version pages.

image

PR #4332, opened 2023-11-08
Fix account update when using 3rd party identity provider

milan-cvetkovic:
If OSM account is created with external identity provider, `auth_uid` field in the form needs to be permitted.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -35,7 +35,7 @@ class AccountsController < ApplicationController
       :form_action => %w[accounts.google.com *.facebook.com login.live.com github.com meta.wikimedia.org]
     )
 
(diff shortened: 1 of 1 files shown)


tomhughes:
Can you explain which cases this fixes? Looking at the code I think it's when you have external auth enabled and you try and change something other than external auth in the settings?

tomhughes:
On further inspection I'm not sure I can see any path which needs this - it's certainly inconsistent that we allow provider but not uid but I'm not sure we need either?

If they're not changed then we don't need to update them, and if they are then we're going to take the other branch and redirect based on the raw parameters, and then the auth callback will fill them both in when it completes.

milan-cvetkovic:
This is scenario that fails today: User has created an account with 3rd party authorization. They go to their settings page and try to update anything. In my case I tried modifying "display name". Clicking on "Save Changes" fails the form validation, since the `auth_uid` field is present in the form (although unmodified and not visible), and is being submitted and not permitted.

I can no longer reproduce this failure, so I will close this PR.

PR #4333, opened 2023-11-08
Format right-to-left and bi-di text in OSM tag values

mapmeld:
The change fixes an issue where right-to-left language strings can be rendered incorrectly in the Tags table of Map Data or History views

## Sample bug in text rendering

I noticed this street label in Erbil with mismatched parentheses. This turned out to be an error in the name string:

Screen Shot 2023-11-08 at 11 01 08 AM

When reviewing the name in the Map Data (or now, [History view](https://www.openstreetmap.org/way/195240072/history)), it looks like the name is wrapped in parentheses. With a wider box or shorter string, the parentheses appear to wrap one word (مشکۆ) . What we're really seeing is the 0th character of the name string , ( , displayed on the left end of this line. 

Screen Shot 2023-11-08 at 11 06 06 AM
Screen Shot 2023-11-08 at 11 21 52 AM

## Corrected

In an updated version with `dir="auto"`, the string renders like it did in the tiles. This makes the misplaced and unpaired parenthesis visible in the web view.

Screen Shot 2023-11-08 at 11 22 16 AM

CSS has a direction property, but it can only be set to `ltr` or `rtl`, not `auto`

I've edited the name of the original way, but the issue can still be seen in the History tab.

### DIFF
--- a/app/views/browse/_tag.html.erb
+++ b/app/views/browse/_tag.html.erb
@@ -1,4 +1,4 @@
 <tr>
-  <th class='py-1 border-grey table-light fw-normal'><%= format_key(tag[0]) %></th>
-  <td class='py-1 border-grey border-start'><%= format_value(tag[0], tag[1]) %></td>
+  <th class='py-1 border-grey table-light fw-normal' dir='auto'><%= format_key(tag[0]) %></th>
+  <td class='py-1 border-grey border-start' dir='auto'><%= format_value(tag[0], tag[1]) %></td>
 </tr>


gravitystorm:
Hi @mapmeld thanks for your pull request.

This PR  addresses one small part of #3428 (and see also #3429 which mostly implements the rest, but is not yet merged).

Have you considered also setting dir=auto for keys? I know the convention in OSM is to use (mostly) English-language keys, but if we are establishing that all user-generated-content should be `dir="auto"` as part of the wider work, then perhaps we should consider setting it for keys here too. What do you think?

In any case, there has been other work on this file since your PR was created, so it needs to be rebased before we can merge it. Can you take care of that too please?

mapmeld:
@gravitystorm thanks - I rebased and have made both key and val have `dir="auto"`. I don't know of examples where the key is in an RTL language, but I agree it's alright to add it on.

PR #4339, opened 2023-11-13
Remove relative position from sidebar child divs other than banners

AntonKhorev:
Continuing https://github.com/openstreetmap/openstreetmap-website/pull/4112#issuecomment-1662212137


tomhughes:
I think I've looked at everything that uses the left sidebar and I can't see any issues with this, so it's looking good, thanks.

PR #4343, opened 2023-11-13
Remove custom css and large left margin for map noscript message

AntonKhorev:
Currently the message for disabled javascript has a large left margin. But what if the window is small? Also `z-index` doesn't do anything because it has static positioning.

Before:
image

After:
image


tomhughes:
That does look a lot better ;-) Thanks...

PR #4345, opened 2023-11-14
Replace padding: $spacer css with class="p-3"

AntonKhorev:
If you're negating it with a Bootstrap class (`mx-n3` on children, see https://github.com/openstreetmap/openstreetmap-website/commit/c8f0a81eb7043c297e050c1f3ff8a8cf7ff613bc), why not set it with a Bootstrap class?


tomhughes:
Looks good, thanks.

PR #4348, opened 2023-11-15
Switch to using native rails support for managing Postgres enumerations

tomhughes:
Followup as suggested during the #4319 review.


gravitystorm:
Thanks for the followup

PR #4353, opened 2023-11-18
Improve handling of RTL styling

### LINKED ISSUE #3910: RTL UI has some reversed features

tomhughes:
This PR reworks how we handle RTL support to try and fix some of the outstanding issues.

A key discovery was that although bootstrap has RTL support we weren't actually taking advantage of that but were instead getting our own variant of it that is kind of broken.

Upstream bootstrap uses [RTLCSS](https://rtlcss.com/) for it's RTL support and has a number of directives in the source to override various aspects of the automatic conversion and if you download bootstrap you will get both `bootstrap.css` and `bootstrap.rtl.css` which has been processed by RTLCSS to produce an RTL version.

The rubygems version of bootstrap doesn't use that though - it ships the unprocessed SCSS source for bootstrap and the only reason it has been working at all is that it gets included into our CSS and then processed by r2 but that of course ignores the RTLCSS directives in the bootstrap source.

So this PR switches our RTL support to use RTLCSS meaning that those directives are obeyed and various bugs go away and overrides in our source can be removed - it also gives us access to a much richer set of customisation directives for future use.

That still didn't fix #3910 though as there was nothing in any of the stylesheets to set the alignment of placeholder text and the browsers seem to default to putting it on the left even in an RTL context so I added an alignment directive which then gets reversed in RTL mode.

After all that I hit https://github.com/rubyjs/mini_racer/issues/289 which prevents `mini_racer` installing in the GitHub Actions ruby 3.0 or 3.1 environments - sadly upstream seem to have given up on it once they figured out 3.2 works but I managed to find a workaround after much digging.


gravitystorm:
> So this PR switches our RTL support to use RTLCSS meaning that those directives are obeyed and various bugs go away and overrides in our source can be removed - it also gives us access to a much richer set of customisation directives for future use.

Awesome, thanks for digging into this. 

> After all that I hit [rubyjs/mini_racer#289](https://github.com/rubyjs/mini_racer/issues/289) which prevents `mini_racer` installing in the GitHub Actions ruby 3.0 or 3.1 environments - sadly upstream seem to have given up on it once they figured out 3.2 works but I managed to find a workaround after much digging.

Not just github actions, because I ran into this with my own system too (I'm using ruby 3.0 for this project).

I'm going to open a separate issue to discuss this since it's not directly related to RTLCSS.

tomhughes:
> Not just github actions, because I ran into this with my own system too (I'm using ruby 3.0 for this project).

That's interesting because I couldn't reproduce it on the dev server or in a clean Ubuntu 22.04 container which both use 3.0...

PR #4358, opened 2023-11-21
Remove 'need to return bbox' comment on conditions_bbox

AntonKhorev:
This comment was added in https://github.com/openstreetmap/openstreetmap-website/commit/95d899786a1bbabacc0cd12ef1c4814118d9d0de when the `conditions_bbox` method actually returned bbox. Bbox returning was removed in https://github.com/openstreetmap/openstreetmap-website/commit/9c28a626cbefd289d17556747979bd73ee3a77cd yet the comment stayed and got copypasted later to the api controller.


tomhughes:
Another good catch, thanks.

Issue #631, opened 2013-12-06
Timestamps across the web site / make format user selectable


woodpeck:
This is a feature request for user-selectable date/time styles across the web site. A detailed explanation follows below.

For a long time we've been using vague timestamps like "about 5 years ago" in several places in the UI. - I see where the motivation comes from (less clutter, nicer on the eye, easier to translate, no time zone issues) but I've always hated it.

The reason is that more often than not, I was using the web site to investigate something. Was a certain edit _before_ or _after_ a certain other edit? Did they make that change _after_ I sent them the message? Was this account created before or after I blocked that other account? Was the block they got a reason for the edit, or was the edit the reason for the block? And so on. Precise time stamps make this kind of work much easier, especially if you're looking at a time ranger where everything is "about one year ago".

With the recent UI change, these imprecise time stamps are used in even more places, occasionally augmented by a hover facility that will show a precise time stamp. In this trac ticket https://trac.openstreetmap.org/ticket/5045 the issue has been discussed, and the complainant has made available a hack that would allow people to have precise time stamps at least where the hover code is enabled. 

So far, so good, and you'll say that my use case is niche and anyway I could build tools for myself to find out what I need to know. BUT, and I say this from my DWG background: Going forward, we will need more "ordinary users" taking stewardship if data, investigating strange edits and wrongdoings, reporting or flagging bad ones, even dealing with them themselves. The "niche" that DWG currently occupies will have to widen in the future, with more and more people looking at the finer details of what is going on in OSM. 

While "5 hours ago" is nice for the casually browsing user, it makes life harder for those who actually want to do some sleuthing, and we don't want to make it harder for them.

Yes it would be terrific if we could have a second web site that was specifically aimed at experts but we won't any time soon. So in the mean time it would be great if the needs of more analytically inclined users were kept in mind when changing things. It would be great if, in their profile settings, users could choose whether they want precise timestamps throughout, or the reader-friendly "5 hours ago" variant.

SomeoneElseOSM:
(re the comments above about tooltips) they don't appear everywhere.  See for example:

http://www.openstreetmap.org/user_blocks/493

HolgerJeromin:
Just another topic are mobile browsers where you have no way to get the exact Information.

polarbearing:
[...]

PR #4361, opened 2023-11-22
Add expansion of time info on touch and mouse activation

HolgerJeromin:
Add touch and mouse interaction of time display.

Initial no change:
image

After touch or mouse focus/selection we have:
image

ref #631
Fixes the "need the exact time info on mobile" in this ticket, but not the "be able to copy the exact time into clipboard" part.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -50,9 +50,9 @@ small, aside {
 time[title] {
   text-decoration: underline dotted;
 }
-time[title]:active::after,
-time[title]:focus::after {
-  content: " (" attr(title) ")";
+time[title][datetime]:active::after,
+time[title][datetime]:focus::after {
+  content: " (" attr(datetime) ")";
 }
 
 #container { position: relative; }
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -50,6 +50,10 @@ small, aside {
 time[title] {
   text-decoration: underline dotted;
 }
+time[title]:active::after,
+time[title]:focus::after {
+  content: " (" attr(title) ")";
+}
 
 #container { position: relative; }
 
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -53,6 +53,7 @@ time[title] {
 time[title][datetime]:active::after,
 time[title][datetime]:focus::after {
   content: " (" attr(datetime) ")";
+  display: inline-block;
 }
 
 #container { position: relative; }
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -26,11 +26,11 @@ module ApplicationHelper
   end
 
   def friendly_date(date)
-    tag.time(time_ago_in_words(date), :title => l(date, :format => :friendly), :datetime => date.xmlschema)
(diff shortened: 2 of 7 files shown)


tomhughes:
What's the purpose of setting the tabindex? Are there any side effects of that in terms of using keyboard navigation?

In that picture it looks like there is a spurious space between the open parenthesis and the expanded date - if that's true can we get rid of it?

HolgerJeromin:
> What's the purpose of setting the tabindex? Are there any side effects of that in terms of using keyboard navigation?

Without tabindex the action is not "sticky" so the expand is removed again with mouseup and touchend.

> In that picture it looks like there is a spurious space between the open parenthesis and the expanded date - if that's true can we get rid of it?

Yes, this is even encoded in the test (from #4349) if I understand the test:
`assert_match %r{^<time title=" *5 March 2014 at 18:58" datetime="2014-03-05T18:58:23Z">.*</time>$}, date`

It comes from `l(date, :format => :friendly)` where I have no idea which component implements this.
So this is IMO better tackled separately.

HolgerJeromin:
We could also move to the technical datetime, but this is IMO more ugly:
image

gravitystorm:
> It comes from `l(date, :format => :friendly)` where I have no idea which component implements this.

The `:friendly` formats are controlled by our locale files, for example in en.yml there is:

```
en:
  time:
    formats:
      friendly: "%e %B %Y at %H:%M"
```

The format codes are from ruby [strftime](https://docs.ruby-lang.org/en/3.2/strftime_formatting_rdoc.html), where `%e` means "Day of the month, in range (1..31), blank-padded:" and `%-d` would be the same but without the padding.

Note that each locale has its own formatting, e.g. japanese (ja) is `'%Y年%m月%e日 %H:%M'`

tomhughes:
I guess the question is, do we try and fix all the formats, or do we just call `strip` on the result of formatting the date...

app/assets/stylesheets/common.scss
```diff
@@ -50,6 +50,10 @@ small, aside {
 time[title] {
   text-decoration: underline dotted;
 }
+time[title]:active::after,
+time[title]:focus::after {
+  content: " (" attr(title) ")";
```
gravitystorm, line 55: This isn't particularly i18n friendly, because it hard-codes the characters used for parenthesis and adds a space character, which might not be wanted in some languages e.g. japanese etc (I don't know for sure, but that's why we leave this to our translators, who are the experts!).

If you want to show some text that's not a perfect match for the existing title attribute, then it would be better to add the new text as a different attribute, and use the i18n system to construct that new display text.

I'm also concerned about whether this works properly for RTL text, but I haven't assessed that.

Overall, this feels like a "neat hack", but using CSS interactions to inject content into the text doesn't feel to me like a particularly robust method.

HolgerJeromin: The Japanese translated title looks a bit chatty. (I marked my parenthesis.)

image
With datetime this would be:
image

HolgerJeromin:
I moved to datetime which is short enough (because this is only one timestamp) even for changesets
and visually works for `jp` and `ar` (ltr) locales.

I will not implement a more complex solution in a near timeframe. 
As this is interaction-only this is IMHO better that nothing to support touch interaction.
(Perfect is the enemy of good)

HolgerJeromin:
> In the RTL version that you show, it says "09T20:32:14Z-05-2017" which is garbled nonsense.

Oh, I missed that. You are right.
I fixed that problem. At least the solution is now not "very bad" anymore :-D

PR #4366, opened 2023-11-23
Remove unused css for lists inside primary nav on narrow screens

AntonKhorev:
That list inside the primary navigation doesn't exist anymore. The only list there is the edit button dropdown, but it's hidden in small-nav mode.


tomhughes:
Looks good, thanks.

PR #4367, opened 2023-11-24
Disable html share button for tracestracktopo layer

AntonKhorev:
fixes https://github.com/openstreetmap/openstreetmap-website/issues/4273

PR #4368, opened 2023-11-24
Read layers with available map key from config/key.yml

AntonKhorev:
There's a bunch of map layer configuration specified directly in javascript files, and I added more in https://github.com/openstreetmap/openstreetmap-website/pull/4367, the check for a layer name that can't be embedded. But it would be more convenient to have this configuration outside.

Here the map key button checks if the layer is one of `["mapnik", "cyclemap"]` to enable/disable itself, with these layer names directly in js. But we can look into `key.yml`, see which layers have keys and enable the key button based on that.

### DIFF
--- a/app/assets/javascripts/leaflet.key.js
+++ b/app/assets/javascripts/leaflet.key.js
@@ -24,7 +24,7 @@ L.OSM.key = function (options) {
     }
 
     function updateButton() {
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good, thanks.

PR #4370, opened 2023-11-25
Report changesets

AntonKhorev:
Restores changeset reporting. It was introduced in the [original issues implementation](https://github.com/openstreetmap/openstreetmap-website/pull/1576), but was [removed](https://github.com/openstreetmap/openstreetmap-website/pull/1576/commits/3006ce9713da119ecbe9e0e1f915df53ff1a898f) before it went live. Later it ended up on the 2021 DWG wishlist.

image

This is the original reporting categories. I asked about them a few months ago but nobody cam up with a better list.

image


tomhughes:
Looks great, and definitely something I've been wanting to try and avoid editing problems having to go through me and be redirected to DWG for action.

The only thing I wonder about is the introduction of multi-word report categories like `mechanical_edit` which wind up with things like this:

image

Maybe we should use the actual report category in the database to do a locale lookup to get a nicer string for the UI which would also then be translatable?

gravitystorm:
During the original implementation of reports and issues, this topic was raised by few different people. The main concern was that we should keep the report function focussed on topics where the general community cannot help themselves, e.g. diary spam. That's why changesets was removed from the original PR.

I suspect that adding changeset reporting will mean that a large amount of data problems will no longer be solved by individual community users, since that is usually more work than just reporting the changeset for someone else (i.e. DWG) to handle instead.

Is this something that DWG have discussed recently? Are they, as a whole group, happy for this feature to be added?

tomhughes:
The thing is people already do that - they just report the user instead of the changeset.

woodpeck:
I share gravitystorm's concerns (the DWG doesn't want to be the "please fix the map for me because I cannot be bothered" team) but TomH is also right, currently people will just report the "user" instead and nothing is gained.

So this pull request doesn't make anything worse in my opinion.

In the long run, if we want to keep the "report" function on the web site, we will have to find a way to let the community-at-large deal with certain types of reports. Admins/Moderators could maybe still triage everything that comes in, but there should be a way to say "this is something the community can handle" and then it goes into a queue where everyone, or maybe everyone with more than a thousand edits or something like that, can handle it.
 
(In case this isn't bleedlingly obvious, the reporting system currently doesn't have an easy way to interact with the complainant. There is way to no easy way to close an issue and fire off a friendly message to the user saying "please fix this yourself" or so, instead you have to manually write user-to-user messages and copy them, and any responses received, into the ticket if you want to keep a record. DWG solves this by copying nearly every issue into OTRS and tracking it from there, an option that would not be available to the community-at-large, so if the handling of issues is opened up to a wider group of people, that would also have to be tackled.)

gravitystorm:
> definitely something I've been wanting to try and avoid editing problems having to go through me and be redirected to DWG for action.

One option would be for us to change the default_assigned_role for user reports, so that they go to moderators first (which they do already for "vandal" reports, but not other categories. Would that help reduce the burden on the administrators?

https://github.com/openstreetmap/openstreetmap-website/blob/75042a04a1e2dec2c821de14237a6b4d5a95382a/app/controllers/reports_controller.rb#L57-L69

tomhughes:
Well sure, at the expense of the 90% of user reports that are about spam profiles being delayed and having to be processed by the moderators first.

Now we could leave `spam` as going to administrator though in my experience people seem to pick pretty randomly from the report reasons :-(

PR #4371, opened 2023-11-27
Add transportation map key for opnvkarte

AntonKhorev:
I decided to check how adding a map key would work by adding one for opnvkarte.

At first I wanted to copy the images from the key at https://publictransportmap.org/. They were base64 data uris, <s>so I added the ability to specify image sources directly in `key.yml`</s>, otherwise they are composed as `key/<layer name>/<image name>`. But I ended up making my own images <s>because the already existing i18n strings want to combine light rail and tram</s>.

I only did the transportation section. Now it's possible to see if a line is a bus or a tram. Other things on that map are more obvious.

image

### DIFF
--- a/config/key.yml
+++ b/config/key.yml
@@ -70,11 +70,20 @@ cyclemap:
   - { min_zoom: 14, max_zoom: 19, name: bicycle_parking, image: bicycle_parking.png }
   - { min_zoom: 16, max_zoom: 19, name: toilets, image: toilets.png }
 opnvkarte:
-  - { min_zoom: 6, name: rail, width: 52, height: 4, fill: "#868686" }
-  - { min_zoom: 6, name: train, width: 52, height: 4, fill: "#ffc366" }
+  - { min_zoom: 6, max_zoom: 7, name: rail, width: 52, height: 1, fill: "#868686" }
+  - { min_zoom: 8, max_zoom: 10, name: rail, width: 52, height: 2, fill: "#868686" }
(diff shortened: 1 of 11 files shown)


pnorman:
If the style doesn't combine light rail and tram to one appearance then the key shouldn't. Not all styles will group things the same way.

config/key.yml
```diff
@@ -70,3 +70,10 @@ cyclemap:
   - { min_zoom: 14, max_zoom: 19, name: bicycle_shop, image: bicycle_shop.png }
   - { min_zoom: 14, max_zoom: 19, name: bicycle_parking, image: bicycle_parking.png }
   - { min_zoom: 16, max_zoom: 19, name: toilets, image: toilets.png }
+opnvkarte:
+  - { min_zoom: 6, name: rail, source: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADQAAAAECAIAAABOYIsRAAAAF0lEQVQoz2Nsa2tjGKyAiWEQg1HHkQsA43UBmpvJuFsAAAAASUVORK5CYII=" }
```
HolgerJeromin, line 74: Perhaps we could use svgs here for easier changeing?

`"data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg"><rect width="300" height="100" style="fill:rgb(0,0,255)" /></svg>"`

AntonKhorev: I guess this will work for simple shapes.

Added svgs with various line widths for different zoom levels.

AntonKhorev:
Updated keys to width/height/fill and a few external svgs.

PR #4378, opened 2023-11-29
Use 5 meter radius for input locations in Valhalla

nilsnolde:
This will search for road segments within 5 meter radius of the input location and choose the best path from/to whichever segment gives the cheapest cost. It should be preferable in most scenarios compared to simply snapping to the closest segment.

I'll do the same PR for our web app at https://github.com/gis-ops/valhalla-app.

### DIFF
--- a/app/assets/javascripts/index/directions/fossgis_valhalla.js
+++ b/app/assets/javascripts/index/directions/fossgis_valhalla.js
@@ -53,7 +53,7 @@ function FOSSGISValhallaEngine(id, costing) {
         data: {
           json: JSON.stringify({
             locations: points.map(function (p) {
-              return { lat: p.lat, lon: p.lng };
+              return { lat: p.lat, lon: p.lng, radius: 5 };
             }),
             costing: costing,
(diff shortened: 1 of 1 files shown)


tomhughes:
That change doesn't seem to make any difference to the route from https://github.com/valhalla/valhalla/issues/4419 for me?

nilsnolde:
Yeah, you're right.. Sorry, didn't think that'd need testing, good that you did.. Not even 50 m makes a difference here. Urgh, that actually might need a bit of debugging, can't see why that wouldn't work. Only :rabbit: :hole: lately ;)

nilsnolde:
I mean the change is still valid generally, IMO it's still good to merge, even the trigger doesn't seem valid atm.

kevinkreiser:
> can't see why that wouldn't work

costing thing where sidewalks are favored more than shared access carriageways. basically the safety of the sidewalk is worth walking the extra distance. this would suggest a costing imbalance though, which is possible. or something else im not thinking of.

PR #4379, opened 2023-11-29
Add corporate members to the list of supporters

lonvia:
Ever since moving to commercial data centers, the membership fees of the OSMF corporate members have become substantial to ensuring funding for hosting. So lets give them a mention in the list of hosting partners.


tomhughes:
According to https://operations.osmfoundation.org/policies/hosting/ the top three donors are to be credited by name in that space...

gravitystorm:
Since the exact format of the credit is covered by an OSMF policy, I'm not comfortable changing it independently. Even if it's a PR from an OSMF Board member :smile:

If the policy is ever updated, then of course I'll be happy to merge the changes. But for now I'm going to close this PR.

pnorman:
The OWG made the decision we're going to change the policy and that three is no longer the ideal number for individual sponsors to list, as there's no longer the big gap between 3rd and 4th.

PR #4382, opened 2023-11-30
Add machine-readable category and time to report template

AntonKhorev:
Before I'm able to change [how report categories look](https://github.com/openstreetmap/openstreetmap-website/pull/4370#issuecomment-1826874270), I have to update [my browser extension](https://github.com/AntonKhorev/osm-dwg-helper). Currently categories are not marked in any machine-readable way, so my extension looks for substrings like "vandal" or "spam". If we start localizing categories, these substrings could be missing from reports html. To address that, I'm adding report categories as data attributes here. And I also wrap report dates in `<time>` with `datetime` attrs.

I've already released the new version of the extension. Now I'm waiting for it to propagate to users. Then I'll remove the draft status of this pull request.


tomhughes:
I have to say I'm not sure I like the idea of embedding data like this whose only purpose is to facilitate scraping of the page by third party components.

AntonKhorev:
> I have to say I'm not sure I like the idea of embedding data like this whose only purpose is to facilitate scraping of the page by third party components.

You already do this: https://github.com/openstreetmap/openstreetmap-website/pull/2266

AntonKhorev:
If you want an excuse for having categories in html, here's a report category filter:

image

Might be useful for issues with a lot of reports. Maybe I should also add filtering by reporting users.

PR #4387, opened 2023-12-03
Add oauth scope for redactions

AntonKhorev:
Currently redactions are allowed if the `write_api` is granted. But `write_api` is a very common scope requested by many apps for things like adding changeset comments. Those apps have no business being able to redact anything.

Here a separate scope for redactions is introduced. `write_api` still allows redactions, this is to be disabled in a later pull request, after [osmtools](https://github.com/woodpeck/osm-revert-scripts) are updated to request the new `write_redactions` scope. There's also redaction bot written in ruby, possibly also needs to be updated.

### DIFF
--- a/app/abilities/api_capability.rb
+++ b/app/abilities/api_capability.rb
@@ -32,9 +32,9 @@ class ApiCapability
           can [:destroy, :restore], ChangesetComment if scope?(token, :write_api)
           can :destroy, Note if scope?(token, :write_notes)
           if user&.terms_agreed?
-            can :redact, OldNode if scope?(token, :write_api)
-            can :redact, OldWay if scope?(token, :write_api)
-            can :redact, OldRelation if scope?(token, :write_api)
+            can :redact, OldNode if scope?(token, :write_api) || scope?(token, :write_redactions)
+            can :redact, OldWay if scope?(token, :write_api) || scope?(token, :write_redactions)
+            can :redact, OldRelation if scope?(token, :write_api) || scope?(token, :write_redactions)
           end
         end
       end
--- a/app/helpers/authorization_helper.rb
+++ b/app/helpers/authorization_helper.rb
@@ -1,0 +1,13 @@
+module AuthorizationHelper
+  include ActionView::Helpers::TranslationHelper
+
+  def authorization_scope(scope)
+    html = []
+    html << t("oauth.scopes.#{scope}")
+    if Oauth::MODERATOR_SCOPES.include? scope
+      html << " "
+      html << image_tag("roles/moderator.png", :srcset => image_path("roles/moderator.svg", :class => "align-text-bottom"), :size => "20x20")
+    end
+    safe_join(html)
+  end
+end
--- a/lib/oauth.rb
+++ b/lib/oauth.rb
@@ -1,6 +1,7 @@
 module Oauth
   SCOPES = %w[read_prefs write_prefs write_diary write_api read_gpx write_gpx write_notes].freeze
   PRIVILEGED_SCOPES = %w[read_email skip_authorization].freeze
+  MODERATOR_SCOPES = %w[].freeze
   OAUTH2_SCOPES = %w[openid].freeze
 
   class Scope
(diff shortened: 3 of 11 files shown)


app/helpers/authorization_helper.rb
```diff
@@ -0,0 +1,15 @@
+module AuthorizationHelper
+  include ActionView::Helpers::TranslationHelper
+
+  MODERATOR_SCOPES = %w[write_redactions].freeze
+
+  def authorization_scope(scope)
+    html = []
+    if MODERATOR_SCOPES.include? scope
+      html << image_tag("roles/moderator.png", :srcset => image_path("roles/moderator.svg", :class => "align-text-bottom"), :size => "20x20")
+      html << " "
```
tomhughes, line 10: Putting the marker before the scope name is going to mean the scopes don't lineup properly which doesn't seem right - maybe it should go after the name?

Should we also add the administrator icon to privileged scopes?

lib/oauth.rb
```diff
@@ -1,7 +1,7 @@
 module Oauth
   SCOPES = %w[read_prefs write_prefs write_diary write_api read_gpx write_gpx write_notes].freeze
   PRIVILEGED_SCOPES = %w[read_email skip_authorization].freeze
-  OAUTH2_SCOPES = %w[openid].freeze
+  OAUTH2_SCOPES = %w[write_redactions openid].freeze
```
tomhughes, line 4: Should we limit who can create an application that requests moderator permissions to moderators in the way we do for administrators and privileged scopes?

app/abilities/api_capability.rb
```diff
@@ -31,10 +31,10 @@ def initialize(token)
...
           can [:destroy, :restore], ChangesetComment if scope?(token, :write_api)
           can :destroy, Note if scope?(token, :write_notes)
-          if user&.terms_agreed?
-            can :redact, OldNode if scope?(token, :write_api)
-            can :redact, OldWay if scope?(token, :write_api)
-            can :redact, OldRelation if scope?(token, :write_api)
+          if user&.terms_agreed? && (scope?(token, :write_api) || scope?(token, :write_redactions))
+            can :redact, OldNode
+            can :redact, OldWay
+            can :redact, OldRelation
```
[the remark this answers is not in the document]
AntonKhorev: You have rubocop checking the number and complexity of conditions. What you'd prefer is going  to cause warnings.

app/helpers/authorization_helper.rb
AntonKhorev: > Should we also add the administrator icon to privileged scopes?

Are they supposed to be used by administrators?

I added these stars in response to https://github.com/openstreetmap/openstreetmap-website/pull/4301#discussion_r1367689627 - to avoid accidentally granting moderator-only permissions to some app that might misuse them. Users are supposed to notice the star in the authorization dialog and be extra careful. `skip_authorization` privileged scope skips that dialog entirely. Users won't have a chance to review the permissions in this case.

The difference is that only administrators can register an app with privileged scopes but anyone can use the app with these scopes(?). Apps with moderator scopes can be registered by anyone, but only moderators can use features granted by these scopes.

app/helpers/authorization_helper.rb
```diff
@@ -0,0 +1,15 @@
+module AuthorizationHelper
+  include ActionView::Helpers::TranslationHelper
+
+  MODERATOR_SCOPES = %w[write_redactions].freeze
```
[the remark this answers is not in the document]
AntonKhorev: Moved `MODERATOR_SCOPES` to `lib/oauth.rb`. It was in the helper because it's not used in any permission granting logic.

app/helpers/authorization_helper.rb
AntonKhorev: Basically do you want everyone who logged in at community.openstreetmap.org to see the admin star on https://www.openstreetmap.org/oauth2/authorized_applications next to "Read user email address"? Nobody gets to review that permission when they log in.

app/abilities/api_capability.rb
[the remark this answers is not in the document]
AntonKhorev: Did this and increased RuboCop complexity limits.

app/helpers/authorization_helper.rb
AntonKhorev: Moved the star marker after the name.

tomhughes: Thanks for moving the marker. You argument about the administrator case being differnet seems sensible so let's leave that.

tomhughes:
All merged and will be deploying soon, thanks.

PR #4388, opened 2023-12-03
Update color of mapnik map key for lakes

AntonKhorev:
The color was changed here:

https://github.com/gravitystorm/openstreetmap-carto/commit/5fd79b7bfad1d895c543e52838c65657fcd9fe1a#diff-7e644412d4287e20f0fe9e717feafdfeaf35670f5ebc2fb20a35ca067aece622L5

There are more color changes to be done. In many cases map key images are just solid rectangles. It's easier to set their size and color in `key.yml`, rather than to point to an external image, which nobody updates for years. Here I generate a svg data uri from width, height and fill properties in `key.yml`.


tomhughes:
I'm not sure what I think about this but I could be convinced... Am I right in thinking that if we took this then #4371 could be simplified to use this technology instead of embedding lots of SVGs in the key data?

AntonKhorev:
> Am I right in thinking that if we took this then #4371 could be simplified to use this technology instead of embedding lots of SVGs in the key data?

Yes

tomhughes:
Thanks. I think we'll go with this then...

PR #4389, opened 2023-12-04
Remove unused .current from primary nav

AntonKhorev:
There's this `current_page_class()` function that applies `current` css class to header menu items pointing to the current page. [This commit](https://github.com/openstreetmap/openstreetmap-website/commit/e5c33c119a9ddd4de149cd8913ed970601103a2b#diff-6d2ef58207c92d1533287af551a075122ba98240a861819047a2ae77b7674f0bR33-R34) broke it for primary nav buttons, but nobody noticed because the `current` class is not used for anything. Still the javascript applied it correctly on partial page reloads.

Here I remove `current` from primary navigation.

It's still used in secondary navigation, makes current links darker.

### DIFF
--- a/app/assets/javascripts/index/export.js
+++ b/app/assets/javascripts/index/export.js
@@ -58,7 +58,6 @@ OSM.Export = function (map) {
   }
 
   page.pushstate = page.popstate = function (path) {
-    $("#export_tab").addClass("current");
     OSM.loadSidebarContent(path, page.load);
   };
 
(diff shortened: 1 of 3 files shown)


AntonKhorev:
There used to be highlighting and there used to be normal map view:

image
(from https://wiki.openstreetmap.org/wiki/OpenStreetMap_Finder)

tomhughes:
I guess we should remove it as it's unused - we can always bring it back if we decide we want to highlight the active one in the future.

Issue #4298, opened 2023-10-18
Align structure.sql with generated output


gravitystorm:
There are a few small differences between our structure.sql file, and the output generated from rails schema dumping on my system. I'm not sure whether this is because of different rails versions or different postgresql versions used between the time that these bits of the file were generated, and today. But it's a bit annoying when trying to develop migrations, so if it's possible for us to resolve them it'll make our developers lives a little bit easier.

On my own system, I get four main differences:
* The `AS integer` phrase for sequences isn't included on my system, but we have it in our structure.sql file
* The `ar_internal_metadata` table has a slightly different definition, i.e.
  ```diff
  -    created_at timestamp(6) without time zone NOT NULL,
  -    updated_at timestamp(6) without time zone NOT NULL
  +    created_at timestamp without time zone NOT NULL,
  +    updated_at timestamp without time zone NOT NULL
  ```
* There is a comment added on the btree_gist extension
* The schema_migrations table has a wildly different order, but I'll ignore that for now

I'm interested to know, before I make a pull request, whether my experience is typical. You can test yourself by running `bundle exec rails db:schema:dump` followed by `git diff` to see what the differences are on your own system between the schema dump and what we have in structure.sql.

tomhughes:
I have loads of differences for things like that (which I think is probably down to postgres version) or field ordering (which is mostly down to me having applied and reverted migrations at various times and in various orders) and I generally figure that trying to "fix" them is as likely to break them for other people as it is to fix them for me... My current diff is:

```patch
diff --git a/db/structure.sql b/db/structure.sql
index 939799c0a..78eef205f 100644
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -9,6 +9,13 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
+--
+-- Name: public; Type: SCHEMA; Schema: -; Owner: -
+--
+
+-- *not* creating schema, since initdb creates it
+
+
 --
 -- Name: btree_gist; Type: EXTENSION; Schema: -; Owner: -
 --
@@ -16,6 +23,13 @@ SET row_security = off;
 CREATE EXTENSION IF NOT EXISTS btree_gist WITH SCHEMA public;
 
 
+--
+-- Name: EXTENSION btree_gist; Type: COMMENT; Schema: -; Owner: -
+--
+
+COMMENT ON EXTENSION btree_gist IS 'support for indexing common datatypes in GiST';
+
+
 --
[...]

PR #4390, opened 2023-12-06
Align structure with generated output

gravitystorm:
This PR makes a minor change to the structure.sql file to align it with the output from running migrations, along with some changes to the normalisation script. These normalisations can be removed since they aren't actually doing anything in CI.

* Adds the btree_gist comment. All developers will have this in their generated output from running the migrations so we should have it here too, to avoid noise in `git diff` (and therefore to make it easier to submit PRs with migrations).
* Remove the timestamp normalisation - see https://github.com/openstreetmap/openstreetmap-website/issues/4298#issuecomment-1832194246 on how to update really old pre-rails-6 databases, if you have one.
* Remove sequence type normalisation - this is actually important, and [differences shouldn't be suppressed](https://github.com/openstreetmap/openstreetmap-website/issues/4298#issuecomment-1832615439)
* Remove references to plpgsql extension, since it's been built-in since PostgreSQL 9.0
* Remove normalisation of immutable keyword, since we no longer have any functions that use it.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -17,6 +17,13 @@ CREATE EXTENSION IF NOT EXISTS btree_gist WITH SCHEMA public;
 
 
 --
+-- Name: EXTENSION btree_gist; Type: COMMENT; Schema: -; Owner: -
+--
+
+COMMENT ON EXTENSION btree_gist IS 'support for indexing common datatypes in GiST';
+
+
+--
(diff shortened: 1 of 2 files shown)


tomhughes:
I think the `btree_gist` comment thing may have been more that older versions didn't have the comment or had a different comment - we've always included the actual enablement of the extension.

Let's put it back in anyway and see if anybody complains...

PR #4392, opened 2023-12-06
Resourceful routing for password resets

gravitystorm:
This PR refactors the passwords controller to use resourceful routing. It thereby aligns the methods with those used by Devise, to smooth future migration.


tomhughes:
Took a while to bend my brain round it but I think I've convinced myself it's correct now, thanks!

PR #4398, opened 2023-12-11
Update mapnik map key rail features

AntonKhorev:
Light rail is visible at z8 and tram is visible at z12. I split their i18 string, adding `light_rail` and <s>replacing `tram_only` with `tram`</s>. Previously I had to add `tram_only` in #4371.

UPD: replacing doesn't quite work. Looks like it's better to keep `tram_only` for now and remove `tram`.


tomhughes:
Thanks for all the key updates. I'm not going to say I've verified every detail of them them but I'm sure they're a significant improvement.

Issue #4083, opened 2023-07-11
NoMethodError: undefined method `exists?' for File:Class


ttimhzv:
### How to reproduce the issue?

Hi, im getting next error when executing `bundle exec rails db:create` command, i followed all steps in tutorial, i created new user and set it in database.yml, i tried setting password too. Any idea why am i getting next error.

```
rails aborted!
NoMethodError: undefined method 'exists?' for File:Class
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/oauth-0.4.7/lib/oauth/consumer.rb:13:in 'block in <class:Consumer>'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/oauth-0.4.7/lib/oauth/consumer.rb:12:in 'each'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/oauth-0.4.7/lib/oauth/consumer.rb:12:in '<class:Consumer>'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/oauth-0.4.7/lib/oauth/consumer.rb:9:in '<module:OAuth>'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/oauth-0.4.7/lib/oauth/consumer.rb:8:in '<main>'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in 'require'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in 'require'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/oauth-0.4.7/lib/oauth/client/helper.rb:2:in '<main>'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in 'require'
/root/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in 'require'
[...]

PR #4399, opened 2023-12-11
Prevent the legacy oauth gem from crashing on Ruby >=3.2

til:
Hi 👋🏾 , openstreetmap newbie here, somewhat familiar with Rails. I wanted to try `openstreetmap-website`  locally and ran into the undefined method `File.exists?` error described [here](https://github.com/openstreetmap/openstreetmap-website/issues/4083) with Ruby 3.2.2.

The installation instructions do not explicitly rule out Ruby 3.2 so I found it a bit surprising. As said in #4083 the only reason this happens is the legacy oauth gem at 0.4.7, which is [going to be removed eventually](https://github.com/openstreetmap/operations/issues/867). Do you think the hack in this PR is an acceptable band-aid to ease the first-time developer experience until the oauth gem is removed? It's a bit ugly I admit, but couldn't think of a better way apart from forking the gem.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,6 +1,7 @@
 require_relative "boot"
 
 require "rails/all"
+require_relative "../lib/legacy_oauth_gem_support"
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
--- a/lib/legacy_oauth_gem_support.rb
+++ b/lib/legacy_oauth_gem_support.rb
@@ -1,0 +1,4 @@
+# This hack prevents the gem oauth 0.4.7 raising an exception on load
+# with Ruby >=3.2. Please remove this file once that version of the
+# oauth gem is not used anymore.
+File.singleton_class.alias_method(:exists?, :exist?) unless File.singleton_class.method_defined?(:exists?)


tomhughes:
I don't really like monkey patching in `application.rb` like that but the nature of this one is that it can't be done from an initialiser.

More likely if this becomes a critical issue I'll fork the gem and just change the Gemfile to take a git version.

til:
Ok, yes I agree it is unfortunate it cannot be fixed less obtrusively.

What about adding a sentence like this at the bottom of INSTALL.md instead?

```markdown
Note that with Ruby 3.2 and later the legacy oauth gem we currently use causes an error and needs to be patched, see [#4083](https://github.com/openstreetmap/openstreetmap-website/issues/4083).
```

Might save a new developer a minute. But maybe I worry about the onboarding experience too much – probably one would find the issue quickly after searching the repo anyway.

gravitystorm:
Does it work with using the https://github.com/Largo/file_exists gem?

tomhughes:
It should do but only if that gem is loaded first which may be hard to guarantee without adding a dependency to the oauth gem.

til:
Tried it: it works when it comes before `oauth-plugin` in the `Gemfile`. I assume this is deterministic. I'll open another PR.

tomhughes:
I've already got that done and I'm just running tests before pushing it ;-)

tomhughes:
Looks good: https://github.com/tomhughes/openstreetmap-website/actions/runs/7171013377

til:
Nice! Much better solution than patching from `application.rb`.

PR #4401, opened 2023-12-12
Add buried to list of secondary Wiki prefixes

KasperFranz:
This pull request adds the buried wikipedia/wikidata/wikicommon tag to the allowed secondary wiki prefixes


tomhughes:
This is a rather rarer tag than any of the existing ones I think - the rarest so far is `architect` by the looks of things which is about four times more common than `buried` which only has a few thousand uses worldwide according to taginfo.

gravitystorm:
@KasperFranz thanks for your first PR!

I agree with tom that we don't have a good way to decide what's worth putting onto this list, and where to draw the line for rare tags. However, I'll go with a bias towards action, so I'll merge this particular PR. If the list continues to expand we can reassess how to manage it.

KasperFranz:
@gravitystorm, @tomhughes 

Thank you for the feedback,

Yes the buried is currently not very widely used,

It might be good to get some kind of documentation on minimum usage, etc., before something like this can merge.

Issue #3546, opened 2022-05-08
The fields of the registration form must be validated immediately when typing


deevroman:
### Description

1. It is difficult to choose a free login. Most likely you won't be able to do it on the first try. And not even with the second one. To check this, you need to submit a form.

2. Any mistake in the form erases the password. And it needs to be entered at least twice! Nothing prevents you from checking all the fields automatically before sending, so that the user immediately knows that everything is filled out correctly.

### Screenshots

| <img width="610" alt="Снимок экрана 2022-05-08 в 18 34 24" src="https://user-images.githubusercontent.com/25708359/167303660-82afc619-c6e9-4067-b4c2-dfb435938948.png"> | <img width="604" alt="Снимок экрана 2022-05-08 в 18 34 31" src="https://user-images.githubusercontent.com/25708359/167303675-26a0a19a-b986-45a8-bad7-50dccad06661.png"> |
| ------------- | ------------- |

PR #4402, opened 2023-12-13
Add some javascript to check dynamically if the username is already taken

pietervdvn:
Hi,

I peeked over the shoulder of a friend who tried to sign up for openstreetmap. He wanted to take his given name as display name, which was already taken. When he tried some variations of his (very common) given name as well, those were taken too.

It would help if there was some more immediate feedback for this, so I've thrown in a bit of javascript into `new.html.erb` which checks 'on the go' if the name is taken or not:

[Screencast from 13-12-23 03:28:02.webm](https://github.com/openstreetmap/openstreetmap-website/assets/1466478/d3a7a718-3097-4c85-8033-540b5e86d74a)


tomhughes:
How is it actually checking? I can't actually see it constructing any sort of URL in that code?

Also that code looks like it has lots of quite modern javascript in it and I'm not sure what our policy is on what our maximum allowed syntax level is.

HolgerJeromin:
> How is it actually checking? I can't actually see it constructing any sort of URL in that code?

The code is running on the URL `https://www.openstreetmap.org/user/new` so fetching 
`fetch("./" + chosenUsername)` will result in checking `https://www.openstreetmap.org/user/<username>` which results in http 200 or 404

HolgerJeromin:
> Also that code looks like it has lots of quite modern javascript in it and I'm not sure what our policy is on what our maximum allowed syntax level is.

This is probably the main point:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility
Removing support for non-chrome-edge and other browsers which are older than 3.5 years.

But: Browsers without support will fail on parse of the `<script>` tag and fall back to behaviour which is live now "no direct user check".

tomhughes:
> `` will result in checking `https://www.openstreetmap.org/user/<username>` which results in http `200` or `404`

I'm not sure it's a good idea to abuse the web site URLs in that way in any case... I was thinking there was an API method that could be used but there isn't one that does the lookup by name.

app/views/users/new.html.erb
```diff
@@ -50,3 +50,55 @@
...
+    el.classList.remove("is-valid")
+    return
+  }
+  if (isValid) {
+    el.classList.add("is-valid")
+    el.classList.remove("is-invalid")
+  } else {
+    el.classList.add("is-invalid")
+    el.classList.remove("is-valid")
+  }
```
AntonKhorev, line 73:
```js
el.classList.toggle("is-valid", isValid)
el.classList.toggle("is-invalid", !isValid)
```

or `$("#user_display_name").toggleClass("is-valid", isValid)` etc if we're still using jquery

pietervdvn:
Hi all,

I'm very happy to see a lot of constructive criticism here, so I updated the code.

1. I moved all the code to 'users.js'
2. Used some JQuery, which significantly shortens the code
3. I also added in a light check for special characters
4. Added comments about some of the questions that were raised here, so that those issues are clear for future readers of the code.

About the syntax: if this improves matters for 99% of the new signups, that is a good step in the right direction. If the script doesn't work (e.g. for people using scriptblockers), they can still sign up and the display name will still be validated by the server. For those, the situation will stay as it is.

tomhughes:
It's not as simple as that - if there is syntax that the browser doesn't handle it may abort parsing and then everything else after that doesn't run.

app/assets/javascripts/user.js
```diff
@@ -223,4 +223,62 @@ $(document).ready(function () {
...
+    if(displayName === ""){
+      indicateInvalidDisplayName(false)
+      return;
+    }
+    if(displayName.match(/[\/;.,?%#]/)!== null) {
+      // We detected an invalid character
+      indicateInvalidDisplayName(false)
+      return
+    }
+    indicateInvalidDisplayName(!isTaken[displayName])
```
AntonKhorev, line 247: `!undefined` is going to be true, the username will be indicated as valid until the fetch completes

app/assets/javascripts/user.js
```diff
@@ -223,4 +223,62 @@ $(document).ready(function () {
   $("#read_tou").on("click", function () {
     $("#continue").prop("disabled", !($(this).prop("checked") && $("#read_ct").prop("checked")));
   });
+
+
+  const isTaken /* Record<string, string> */ = {}
+
+  function indicateInvalidDisplayName(isValid /* boolean | undefined*/) {
```
AntonKhorev, line 230: The function name is confusing. It indicates valid names too, and the argument is `isValid`.

app/assets/javascripts/user.js
```diff
@@ -223,4 +223,62 @@ $(document).ready(function () {
...
+    // isValid might also be 'undefined', hence we check explicitly for 'true' and 'false'
+    $("#user_display_name").toggleClass("is-valid", isValid === true)
+    $("#user_display_name").toggleClass("is-invalid", isValid === false)
+  }
+
+  function checkDisplayName(){
+    const displayName = $("#user_display_name").val()
+    if(displayName === ""){
```
AntonKhorev, line 238: Might check instead if the name is shorter than 3 characters.

tomhughes: We don't really want to be duplicating validations on the client as they will just get out of sync.

This is why it would probably better to have a specific backend method in the rails code that could be called with a proposed username and which would apply all the validations the user model will eventually apply.

app/assets/javascripts/user.js
```diff
@@ -223,4 +223,62 @@ $(document).ready(function () {
...
+    ).catch(reason => console.warn("Could not fetch info about", displayName))
+  }
+
+  $("#user_display_name").on("change", function () {
+    checkDisplayName()
+  })
+
+  $("#user_display_name").on("input", function () {
+    checkDisplayName()
+  })
```
AntonKhorev, line 281: Isn't only `input` handler enough?

app/assets/javascripts/user.js
AntonKhorev: Right now if I start typing the name, this code will say that the name is valid after the first typed character. But it's obviously invalid.

grischard:
> I was thinking there was an API method that could be used but there isn't one that does the lookup by name.

There is a single one. In [the civicrm username check extension](https://github.com/grischard/org.openstreetmap.username/blob/9d67583bd3e0342a9770cde48f23586f707012ee/username.php#L61) I check `https://api.openstreetmap.org/api/0.6/changesets?time=9999-01-01&display_name=TomH` or `https://api.openstreetmap.org/api/0.6/changesets?time=9999-01-01&display_name=GhostofTomH` for 404.

app/assets/javascripts/user.js
pietervdvn: > We don't really want to be duplicating validations on the client as they will just get out of sync.
> This is why it would probably better to have a specific backend method in the rails code that could be called with a proposed username and which would apply all the validations the user model will eventually apply.

Totally agree with this, but I have no idea how to do this. Even better would be to give a (translated) reason why it is not valid.

However, this is already a step in the right direction...

app/assets/javascripts/user.js
pietervdvn: In practice, yes. However, a piece of code (e.g. a password manager) might insert a value programmatically, which might not trigger 'input'

gravitystorm:
@pietervdvn Thanks for your PR. I agree that this sort of thing could be useful for people signing up, particularly since we're a large project with many existing registrations so users are increasingly likely to find their first-choice display_name is already taken.

However, there's a few things with this specific implementation that would need to be considered:
* display_name checking - this can't just be done against the user profile page. Firstly, it's not designed as a lightweight endpoint, there is a lot of stuff being rendered on that page. Although using HEAD will reduce the payload, I don't think it's a good idea to call HEAD on full pages for every keystroke - the latency is likely to be too high. A better approach would be a specific endpoint, e.g. a json endpoint that only does one thing. 
* display_name checking - this also can't be done against the user profile page, due to unconfirmed users, suspended users and deleted users. These are pages that will return a 404, but the display_name is already taken. 
* Lack of tests - this needs to have some decent tests, otherwise any breakage (e.g. due to refactoring) is not going to be noticed by developers - we rarely sign up for new accounts!
* This is "ad-hoc client-side validation" and in general that makes me hesitant. Particularly for things like character validation rules, it would be much better to drive the client-side validation from the existing model validations. Sure, this particular PR is simply for just one field on just one form, but if we write ad-hoc code we'll end up with a ton of javascript to maintain when we want to roll this out to other fields and/or other forms. So I would like to know more about what options there are in the rails ecosystem (e.g. https://github.com/judgegem/judge or others ) that can do this properly, with less custom code for us to maintain.
* Accessibility - I'm vaguely aware of there being [APIs in modern HTML to do client-side validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation), and this is likely to be more involved than just flipping a class on an element. But I'm not an expert on that. If we used a framework to do this, that project team would likely have more expertise on this topic than we have.
* Textual feedback - I would want to see the same text feedback on what the errors are when done both client-side and server-side. Just adding the class isn't enough to guide the users as to what is wrong and what they need to do to fix it.

app/assets/javascripts/user.js
HolgerJeromin: Did you found such a software?
There are keyboards (iPhone? IPhone landscape mode?) I saw which manipulates the string and pushes the value in one go. These did not triggered "change". But I did not found an interaction which does not trigger "input".

app/assets/javascripts/user.js
AntonKhorev: Last change of minimal display name length was [in 2006](https://github.com/openstreetmap/openstreetmap-website/commit/37bd2971b49da9364fe5dd25415e97e38b34abd7#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR10). Getting out of sync here is very unlikely. If you think that it is, you can just avoid validating short names by not doing requests and not adding the `is-valid` class.

AntonKhorev:
> A better approach would be a specific endpoint, e.g. a json endpoint that only does one thing.

But you need to determine two things here:
- is the name valid?
- is the name taken?

Or alternatively only check if the name is taken and indicate that and never claim that the name is valid.

AntonKhorev:
Requested in #3546 among other things.

Issue #4128, opened 2023-07-31
Proposal: Seamless Sign-On for OpenStreetMap users from 3rd party apps


stalker314314:
### Problem

Hi all,

Context: [Map builder](https://wiki.openstreetmap.org/wiki/Map_builder) (OSM map editor) team at Microsoft is trying to improve user experience during signup process. We’d like to propose couple of modifications to OSM code base for more streamlined single sign-in with 3rd party identity providers (Google, Facebook, Microsoft, Wikipedia, GitHub). The specific scenario of interest are the 3rd party applications acting on behalf of users, typically by submitting modifications to the map (map editor applications), one of which is our Map builder. Due to complex process involving multiple forms to fill and navigating up to 3 applications, it is anticipated that the number of users drops significantly before completing their first-time contribution.

We are ready to invest engineering resources in developing and implementing the proposal. Our hope is that this would benefit both signups coming from inside OSM website as well as from any other third-party signup process!

### Description

We are focusing here on:
- Reducing the information required from user to fill in when the information is available from IdP (Identity Provider),
- Removing the email verification step when it is not necessary,
- Reducing the number of steps required to complete the process.

We propose to do this gradually, together with inputs from you, in one or more PRs, and here is short breakdown of proposed changes:

- [ ] Pre-populate email field when using "Login with Microsoft" and "Login with GitHub"
- [ ] Improvements to Login and Sign-Up Screens
- [ ] Introduce "Themed layouts" for Sign-up and Login views.

We would like to hear your ideas, suggestions, and concerns in the comments before we dive in into implementation!

Here are all these changes in more tech details

---

## 1. Pre-populate email field

Currently, only "Login with Facebook" and "Login with Google" pre-populate email field in Sign-up form.
The `omniauth` plugin currently used for "Login With Microsoft" is `omniauth-windowslive` and it does not properly parse the information received from the identity provider. It leaves the email field empty, even though email is available. It appears that replacing `omniauth-windowslive` with alternative plugins providing similar functionality would properly parse the tokens.

- Use alternative plugin for "login With Microsoft" (modify `config/initializers/omniauth.rb`, update `Gemfile`).
- Verify migration path for the users with existing tokens obtained from `omniauth-windowslive`.
- If necessary, submit PRs for appropriate `omniauth` plugin strategies.
- Create a new plugin if updating strategy is necessary and existing plugins are no longer maintained.
- Investigate if GitHub and OpenMedia providers provide verified email addresses.
[...]

PR #4407, opened 2023-12-14
Implement allow list for account creation

milan-cvetkovic:
Allows mechanism to configure exceptions to rate limiting for account creation.

Uses ACL table with `allow_account_creation` key to allow bypassing account creation rate limiter from specific IP address/network range for users with specified email `domain` and email `mx` address.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -355,6 +355,8 @@ class UsersController < ApplicationController
                    domain_mx_servers(domain)
                  end
 
+    return true if Acl.allow_account_creation(request.remote_ip, :domain => domain, :mx => mx_servers)
+
     blocked = Acl.no_account_creation(request.remote_ip, :domain => domain, :mx => mx_servers)
 
     blocked ||= SIGNUP_IP_LIMITER && !SIGNUP_IP_LIMITER.allow?(request.remote_ip)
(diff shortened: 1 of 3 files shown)


app/controllers/users_controller.rb
```diff
@@ -355,6 +355,8 @@ def check_signup_allowed(email = nil)
                    domain_mx_servers(domain)
                  end
 
+    return true if Acl.allow_account_creation(request.remote_ip, :domain => domain, :mx => mx_servers)
```
stalker314314, line 358: I think blocking should have higher priorty (to be evaluated first) than allowing

tomhughes:
This is to allow us to whitelist Microsoft's account creation until #4128 is completed.

app/controllers/users_controller.rb
milan-cvetkovic: If the IP address is being blocked due to rate limiting, there would be no point allowing it afterwards.

milan-cvetkovic: Re-arranged to account for non-rate-based blocks first.

tomhughes (changes requested):
I think it would be better if an allow ACL took priority over a block ACL as it would provide an extra use for this in that I would be able to exempt specific IPs from a block which would be useful sometimes.

tomhughes (approved):

tomhughes:
Looks good now, thanks.

PR #4409, opened 2023-12-15
Remove png sprites

AntonKhorev:
Svg sprites were enabled in https://github.com/openstreetmap/openstreetmap-website/commit/0d0498d303f36749cd48d8e6eadb1abafe3f828c. Was it a correct fallback for png? Do we need a fallback?


tomhughes:
I guess https://caniuse.com/svg and https://caniuse.com/svg-css are the relevant statistics and I can't see anything obvious there that still needs a fallback?

AntonKhorev:
Bootstrap close buttons are also data url svgs. We are fully into svgs by now.

I don't know if that `sprite.png` was even intended to be a fallback and I don't think it was working like a fallback. You first defined `background-image` with a shortcut `background` property then redefined it with `background-image`, throwing away the old value.

tomhughes:
Sounds like a reasonable argument. Will merge it, thanks.

PR #4419, opened 2023-12-18
Optimize escaping of data uri svgs

AntonKhorev:
Now data uri svgs are used throughout the map key, and they affect the generated file size.

The default `url_encode` function %-encodes too many characters. There's a question which ones have to be encoded. `#` certainly is one of them, unencoded `#` will break the uri. If you strictly follow [the rfc from 1998](https://www.ietf.org/rfc/rfc2397.txt), spaces need to be encoded, yet nobody does that with svgs. For example [this library here](https://github.com/tigt/mini-svg-data-uri/blob/00dc78c8f77eb7b47299a9e3d564749105810c9c/index.js#L31) deliberately unencodes them. `<>` are commonly encoded by svg data uri libraries, yet the example in Wikipedia [contains them unencoded](https://en.wikipedia.org/wiki/Data_URI_scheme) and it seems to work fine.

I'm %-encoding [these chars](https://github.com/yoksel/url-encoder/blob/2c31ea619932c61f418bb8dfb945ddc95f9927fd/src/js/script.js#L15C44-L15C44) as used in [this tool](https://yoksel.github.io/url-encoder/).

Before:

    

After:

### DIFF
--- a/app/helpers/svg_helper.rb
+++ b/app/helpers/svg_helper.rb
@@ -1,0 +1,8 @@
+module SvgHelper
+  def solid_svg_image_tag(width, height, fill)
+    svg = "<svg xmlns='http://www.w3.org/2000/svg' width='#{width}' height='#{height}'>" \
+          "<rect width='100%' height='100%' fill='#{fill}' />" \
+          "</svg>"
+    image_tag "data:image/svg+xml,#{u(svg)}"
+  end
+end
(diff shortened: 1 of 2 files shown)


gravitystorm:
What's the comparison assuming gzip/brotli compression of the whole page output? It might be small enough that the extra code (particularly the regexp) isn't worth maintaining.

AntonKhorev:
Or the escaping of anything other than `#` could be removed. `<>` seems to be usually escaped because of [IE support](https://github.com/less/less.js/issues/2184#issuecomment-60499538).

AntonKhorev:
Actually why don't I inline the svgs completely, without ``?

PR #4421, opened 2023-12-19
Use inline svgs in map key

AntonKhorev:
Instead of [optimizing svg data uris](https://github.com/openstreetmap/openstreetmap-website/pull/4419) we can use inline svgs here.

Now this:

    

becomes this:

    <svg width="52" height="1" class="d-block mx-auto"><rect width="100%" height="100%" fill="#787878" /></svg>


tomhughes:
That's so much nicer than the data URIs ;-) Thanks!

PR #4425, opened 2023-12-20
Allow arrays of names in map key entries

### LINKED ISSUE #3816: Avoid using `to_sentence` to create map key entries
In https://github.com/openstreetmap/openstreetmap-website/pull/3731 it was shown that the map key isn't displayed correctly for certain languages.

The underlying cause of the problem is that we are using `.to_sentence` to join individually translated words, which is [unnecessary](https://github.com/openstreetmap/openstreetmap-website/pull/3731#issuecomment-1269667154) and causes this bug.

This issue is here since the original PR was closed, and an alternative PR hasn't yet been created, and we need to ensure that we don't forget about this problem in the meantime.

AntonKhorev:
Fix #3816 by not constructing sentences and not having combined i18n strings.

I don't think sentences are expected in map key, so instead of this:
image
we can connect terms by dots, like we do with *secondary actions* ("Download XML ·View History" on element pages etc):
image

Don't group locale strings in site.key.table.entry. Instead of this:

          cable:
            - Cable car
            - chair lift

have this:

          cable_car: "Cable car"
          chair_lift: "Chair lift"

Group them in `key.yml`, depending on layer. Instead of this:

    - { min_zoom: 12, name: cable, image: cable.png }

have this:

    - { min_zoom: 12, name: [cable_car, chair_lift], image: cable.png }

---

The advantage is that if features are grouped differently on various layers, their names can be translated independently. For example, mapnik has different colors for parks and allotments with two different entries in `key.yml`:

    - { min_zoom: 10, name: park, width: 26, height: 10, fill: "#c8facc" }
    - { min_zoom: 10, name: allotments, width: 26, height: 10, fill: "#d4e6cc" }

Opnvkarte renders parks and allotments the same, they can have one entry:

    - { min_zoom: 12, name: [park, allotments], width: 26, height: 10, fill: "#c2d89a" }

And there can be two independent locale strings:

    park: "Park"
    allotments: "Allotments"

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -50,6 +50,9 @@ class SiteController < ApplicationController
     expires_in 7.days, :public => true
     @key = YAML.load_file(Rails.root.join("config/key.yml"))
     @key.each_value do |layer_data|
+      layer_data.each do |entry|
+        entry["name"] = Array(entry["name"])
+      end
       layer_data.each_cons(2) do |entry, next_entry|
         entry["max_zoom"] = next_entry["min_zoom"] - 1 if entry["name"] == next_entry["name"] && !entry["max_zoom"] && next_entry["min_zoom"]
       end
--- a/app/views/site/key.html.erb
+++ b/app/views/site/key.html.erb
@@ -11,7 +11,7 @@
             <% end %>
           </td>
           <td>
-            <%= Array(t(".table.entry.#{entry['name']}")).to_sentence %>
+            <%= entry["name"].map { |feature_name| t ".table.entry.#{feature_name}" }.join(" · ") %>
           </td>
         <% end %>
       <% end %>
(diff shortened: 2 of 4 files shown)


tomhughes:
The basic idea seems sounds but I wonder if a mid-dot like this is the best separator?

In english I would probably just use `,` or `/` as a separator but I don't know how well that translates to other languages.

AntonKhorev:
Commas didn't translate well in #3731 that preceded https://github.com/openstreetmap/openstreetmap-website/issues/3816.

tomhughes:
Wasn't the issue there `to_sentence` which is more than just adding commas because it uses `and` for the last one which was the problem I think?

AntonKhorev:
"opposite direction commas" were also a problem. Middot is symmetric, so there's no "opposite direction middot". I fugured that's why it's used elsewhere.

tomhughes:
Let's go with this then - we can always change it if we don't like it later.

PR #4431, opened 2023-12-24
Add border/casing/line attributes to svg map keys

AntonKhorev:
This covers most of map key image types and makes it possible to remove many external images.

Before this there was only a `fill` attribute that created a `<rect ... fill="...">`.
Now:
- `border` adds `stroke` to `<rect>`
- `casing` adds two lines above and below with `stroke`
- `line` adds a horizontal line in the middle with `stroke`
- it's possible to add `-width`, `-dasharray` and `-dashoffset` suffixes to `border`/`casing`/`line` to get `stroke-width` etc


tomhughes:
I'd been meaning to ask about exactly this idea, so thanks for beating me to it!

PR #4433, opened 2023-12-24
Clean up sprites

AntonKhorev:
Originally `sprite.svg` had defiintions that are only useful whne editing in Inkscape. Later most of them were removed in https://github.com/openstreetmap/openstreetmap-website/commit/e592e12f1c87b60fe20fea40ecafeec799ea46ec. Still some useless things were left, for example auto-generated ids like `id="path3361-4"` that were not referenced anywhere.


tomhughes:
I guess it doesn't do any harm, though they'll just come back again if somebody edits it in inkscape in the future but then so will a lot of other things that the previous commit removed.

PR #4434, opened 2023-12-25
Add missing locale string for /oauth2/authorized_applications

AntonKhorev:
for this message when no no authorized apps:

> You have not yet authorized any [OAuth 2](https://oauth.net/2/) applications.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2683,6 +2683,7 @@ en:
       application: "Application"
       permissions: "Permissions"
       no_applications_html: "You have not yet authorized any %{oauth2} applications."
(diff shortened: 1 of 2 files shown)


gravitystorm:
Does this mean that the page isn't tested at all? I would have thought that the test suite would catch any completely missing translation keys.

AntonKhorev:
System tests fail on that page when there are no registered apps.

gravitystorm:
I therefore think we should add a test which a) fails without this fix and b) passes with this fix.

AntonKhorev:
Added a test.

tomhughes:
Looks good now, thanks.

PR #4436, opened 2023-12-25
Add system test with osm website oauth app registration

AntonKhorev:
Let's suppose you want to have a system test that checks if adding comments to notes work. This commenting is done via the api. For the api to work you have to [register the website oauth app](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONFIGURE.md#oauth-consumer-keys). Here I'm adding a test superclass that does this registration in `before_setup`.

### DIFF
--- a/test/system/oauth2_test.rb
+++ b/test/system/oauth2_test.rb
@@ -5,6 +5,12 @@ class Oauth2Test < ApplicationSystemTestCase
     sign_in_as(create(:user))
     visit oauth_authorized_applications_path
 
+    assert_text "OpenStreetMap Web Site"
+
+    accept_alert do
+      click_link "Revoke Access"
+    end
+
     assert_text "You have not yet authorized any OAuth 2 applications."
   end
 end
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -10,4 +10,15 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   driven_by :selenium, :using => :headless_firefox do |options|
     options.add_preference("intl.accept_languages", "en")
   end
+
+  def before_setup
+    super
+    osm_website_app = create(:oauth_application, :name => "OpenStreetMap Web Site", :scopes => "write_api write_notes")
+    Settings.oauth_application = osm_website_app.uid
+  end
+
+  def after_teardown
+    Settings.reload!
+    super
+  end
 end
--- a/test/system/note_comments_test.rb
+++ b/test/system/note_comments_test.rb
@@ -18,17 +18,25 @@ class NoteCommentsTest < ApplicationSystemTestCase
     assert_no_link "Log in to comment on this note"
   end
 
-  def test_action_text
+  def test_add_comment
     note = create(:note_with_comments)
-    sign_in_as(create(:user))
+    user = create(:user)
+    sign_in_as(user)
     visit note_path(note)
 
+    assert_no_content "Comment from #{user.display_name}"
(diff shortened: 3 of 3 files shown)


tomhughes:
Is there any reason not to just add this to `ApplicationSystemTestCase`? All system tests in this repository are testing the web site and a correctly setup instance of the web site has this OAuth application?

AntonKhorev:
It would be a bit strange that the website app registration is not automated for dev/production, but automated for tests.

Are we assuming that the correct osm-website setup always includes the app registration? Maybe someone wants to run it on a read-only db, then they don't need to register.

tomhughes:
Well "correct" from the point of view of our use of it, which is what we're retying to test.

If other people want to run it in a strange way then that's their problem.

AntonKhorev:
added to ApplicationSystemTestCase

tomhughes:
Thanks for making that change.

I wonder if rather than tearing this down every time and then setting it up again we should just get rid of the teardown and make the setup step conditional on `Settings.oauth_application.nil?` so that is just happens when the first test runs that needs it?

AntonKhorev:
Settings are persisted between tests unless reloaded, but the database state is rolled back. After that `Settings.oauth_application` will have an invalid value.

test/system/oauth2_test.rb
```diff
@@ -2,9 +2,13 @@
 
 class Oauth2Test < ApplicationSystemTestCase
   def test_authorized_applications
-    sign_in_as(create(:user))
+    sign_in_as(create(:user)) # successful login redirects to a page that authorizes the website app
     visit oauth_authorized_applications_path
 
+    accept_alert do
+      click_link "Revoke Access"
+    end
```
tomhughes, line 10: Rather than adding this (and the comment above) all of which took me a few minutes to understand, wouldn't it be simpler just to change the assert below to expect an application?

This test is just trying to check that the authorized applications page exists and displays rather than the specific contents...

An alternative which would be clearer would be to drop the comment and add an assertion above the revoke to check that the website application is listed and then revoke it and check you get the no apps message?

AntonKhorev: > Rather than adding this (and the comment above) all of which took me a few minutes to understand, wouldn't it be simpler just to change the assert below to expect an application?
> This test is just trying to check that the authorized applications page exists and displays rather than the specific contents...

I was asked to test _the specific contents_ of no registered apps: https://github.com/openstreetmap/openstreetmap-website/pull/4434#issuecomment-1869571084. I'm not testing for the website app to be listed here. I'm adapting the already existing test of the message that had a missing translation up until recently to work with the auto-registering website app, which you asked to register for every system test.

> An alternative which would be clearer would be to drop the comment and add an assertion above the revoke to check that the website application is listed and then revoke it and check you get the no apps message?

What is the alternative? I'm revoking the access and checking the message here. Do you want me to revoke it by manipulating models/db then reload the page to check the message?

tomhughes: The alternative I was suggesting was to remove the comment after the `create` and replace it with an assertion before the revoke that check there is a single app listed and that it's the web site app - it serves the same purpose as the comment in explaining what the revoke is doing but I think it's clearer and also extends testing of the page to test that it lists authorized apps correctly.

AntonKhorev: I replaced the comment with the assert, but this is really testing side effects of `sign_in_as()`. You could have another login method that avoids visiting the pages that auto-grant permissions to the website app.

PR #4437, opened 2023-12-26
Fix messages heading update when message is deleted

### LINKED ISSUE #4265: Messages counter is not reduced in "My Outbox" (/messages/outbox) when deleting a message
### URL

_No response_

** Starting state **

4 sent messages in the Outbox of the OSM DM-system.
"You have 4 sent messages"

** Process **

Delete one of those messages, using the "Delete" button on the right

** Expected result **

Expectation would be the reduction of the "You have X messages" counter, just like it happens in "My Inbox" (/messages/inbox).

** Observed result **

The counter remains the same until the page is reloaded.

Outbox, before delete:
grafik
Outbox, after delete:
grafik
Inbox, before delete:
grafik
Inbox, after delete:
grafik

AntonKhorev:
Fixes #4265 and similar bug for muted messages list.

### DIFF
--- a/app/assets/javascripts/messages.js
+++ b/app/assets/javascripts/messages.js
@@ -22,6 +22,8 @@ $(document).ready(function () {
     $(".user-button").before(data.inboxanchor);
 
     $("#inbox-count").replaceWith(data.inbox_count);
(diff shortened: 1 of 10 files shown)


tomhughes:
Looks good, thanks.

PR #4439, opened 2023-12-27
Set block ends_at to exactly created_at + duration

AntonKhorev:
If you block a user for two weeks, you'll see this:

image

> Duration: 1 week

The block end date is computed before `created_at` and its duration ends up being slightly shorter. Everything is rounded down and you get an incorrect value.


tomhughes:
Nice find...

PR #4440, opened 2023-12-28
Add revoke all blocks moderator action

AntonKhorev:
User pages have "Revoke all blocks" link when users have active blocks. These links are only visible to moderators.

image

image

See here for reasons: https://github.com/openstreetmap/openstreetmap-website/pull/4301#discussion_r1367510802

### DIFF
--- a/test/system/user_blocks_test.rb
+++ b/test/system/user_blocks_test.rb
@@ -27,12 +27,16 @@ class ReportNoteTest < ApplicationSystemTestCase
     assert_no_link "Revoke all blocks"
   end
 
-  test "revoke all link is present for moderators when viewed user has active blocks" do
+  test "revoke all link is present and working for moderators when viewed user has active blocks" do
     blocked_user = create(:user)
     create(:user_block, :user => blocked_user)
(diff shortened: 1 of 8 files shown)


test/system/user_blocks_test.rb
```diff
@@ -0,0 +1,71 @@
...
+  test "revoke all link is absent for regular users when viewed user has active blocks" do
+    blocked_user = create(:user)
+    create(:user_block, :user => blocked_user)
+    sign_in_as(create(:user))
+
+    visit user_path(blocked_user)
+    assert_no_link "Revoke all blocks"
+  end
+
+  test "revoke all link is present for moderators when viewed user has no active blocks" do
```
tomhughes, line 21: I think you meant `absent` in that description ;-)

tomhughes (changes requested):

test/system/user_blocks_test.rb
AntonKhorev: fixed

tomhughes (approved):
That's better, thanks.

PR #4441, opened 2023-12-28
Add markdown guidance for code blocks

altilunium:
Kramdown's code block markdown syntax is slightly different from the popular markdown dialect used on GitHub and GitLab, causing some confusion for users. 

image

We can solve this problem by adding code block markdown guidance to `_markdown_help.html.erb`.

### DIFF
--- a/app/views/shared/_markdown_help.html.erb
+++ b/app/views/shared/_markdown_help.html.erb
@@ -18,4 +18,7 @@
 
   <dt><%= t ".image" %></dt>
   <dd>![<%= t ".alt" %>](<%= t ".url" %>)</dd>
+
+  <dt><%= t ".codeblock%></dt>
+  <dd>~~~<br><%= t ".codeblock%><br>~~~</dd>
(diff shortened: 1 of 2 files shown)


app/views/shared/_markdown_help.html.erb
```diff
@@ -18,4 +18,7 @@
 
   <dt><%= t ".image" %></dt>
   <dd>![<%= t ".alt" %>](<%= t ".url" %>)</dd>
+
+  <dt><%= t ".codeblock%></dt>
+  <dd>~~~<br><%= t ".codeblock%><br>~~~</dd>
```
tomhughes, line 23: These two lines are missing a closing quote for the resource name - it should be `<%= t ".codeblock" %>`.

tomhughes (changes requested):

tomhughes:
That looks much better now, thanks.

tomhughes (approved):

PR #4443, opened 2023-12-29
test/system/messages_test.rb: Rename NoteCommentsTest -> MessagesTest

grekko:

The class names are expected to match the test file names.

Follow-Up for #4437

(btw. thanks @AntonKhorev for the fix!)


tomhughes:
Thanks for catching that - you're obviously more on the ball that me!

PR #4445, opened 2023-12-30
Truncate usernames in profile links

AntonKhorev:
Experimenting with truncation of very long usernames. In general you can't line-wrap them, see the note at https://getbootstrap.com/docs/5.3/utilities/text/#word-break

image

image

image

### DIFF
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -50,6 +50,16 @@ module UserHelper
     end
   end
 
+  def truncated_user_link(user, **options)
+    classes = %w[d-inline-block align-bottom text-truncate]
+    if options[:width]
+      style = "max-width: #{options[:width]}"
+    else
+      classes <<= "mw-100"
+    end
+    link_to user.display_name, user, :class => classes, :style => style
+  end
+
   # External authentication support
 
   def openid_logo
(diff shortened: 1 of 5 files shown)


tordans:
I think truncation is a valid solution in cases like this (where the username is not the center point of information). I also like to add a way to see the full name on mouse over like with a title attribute (eg ok the TH/TD) or a full tooltip so I don't need to leave the page to see the full name on the user's profile.

app/helpers/user_helper.rb
```diff
@@ -50,6 +50,16 @@ def user_image_url(user, options = {})
     end
   end
 
+  def truncated_user_link(user, **options)
+    classes = %w[d-inline-block align-bottom text-truncate]
+    if options[:width]
+      style = "max-width: #{options[:width]}"
```
gravitystorm, line 56: I'm fairly sure that inline styles aren't allowed by our Content Security Policy.

gravitystorm (changes requested):
I think this is a valid problem, so it's worth solving. But I'm surprised that the solution is only for a couple of places (changeset discussions, user messages and user blocks) and not everywhere else on the site too.

Also the tests fail, so those need fixing too.

app/helpers/user_helper.rb
tomhughes: They aren't...

AntonKhorev:
Superseded by #5372, #5371 etc

Issue #1087, opened 2015-11-17
Footway, path, combined description.


AllroadsNL:
I placed a issue in the wrong repository.
https://github.com/gravitystorm/openstreetmap-carto/issues/1970

Now that footway and path is combined, the combined legend name is wrong. I believe the default mapname now is footway for the red dotted line.
( Search footway https://github.com/openstreetmap/openstreetmap-website/search?utf8=%E2%9C%93&q=footway
Search legend https://github.com/openstreetmap/openstreetmap-website/search?utf8=%E2%9C%93&q=legend&type=Code
 ) Could not find where it is set, so made this issue on this main page.

Translations in other languages, like above example, is not a combined description. 
Footway, mostly mean only access by foot, path more vehicles allowed, also foot.
Path would be more correct. Footway is a subdivision of path. Cycleway also.
Maybe =  Path(foot)

tomhughes:
What is it you are actually saying is wrong?

tomhughes:
Presumably this is what you are talking about:

![screenshot from 2015-11-17 16-14-16](https://cloud.githubusercontent.com/assets/147741/11217055/610f98d2-8d46-11e5-9f01-b66468cafede.png)

Are you just saying that "Voetpad" is the wrong way to describe the dotted red line?

If you are then we aren't able to help - either you or somebody else will need to go to https://translatewiki.net/wiki/Translating:OpenStreetMap and update the Dutch translation to reflect that.

HolgerJeromin:
I have not really understood translatewiki. But it seems the "Main Road" thing is not translatable right now. Probably the strings are not updated.

tomhughes:
@HolgerJeromin Can we avoid confusing things by talking about things that aren't actually being complained about on this ticket... You know perfectly well that it will be translatable after the next TranslateWiki sync...

AllroadsNL:
Yes, that is mentioned, red dotted line is wrong.

The translation of footway to Dutch "Voetpad" is correct! no problem, the Dutch law use the a traffic_sign=NL:G7 and use the name "Voetpad".
http://wiki.openstreetmap.org/w/images/thumb/0/0a/Nederlands_verkeersbord_G7.svg/50px-Nederlands_verkeersbord_G7.svg.png
http://wiki.openstreetmap.org/wiki/NL:The_Netherlands_roads_tagging#Overige_wegen_en_paden
[...]

PR #4447, opened 2023-12-30
Split footway map key string

AntonKhorev:
Changes this mapnik map key
image
to this
image
and the same for cyclemap.

I'm not entirely sure that it should be done, but here's a complaint https://github.com/openstreetmap/openstreetmap-website/issues/1087 claiming that it's difficult to translate the combined *highway=footway / highway=path* string correctly.

The rendering is combined though on both [mapnik](https://github.com/gravitystorm/openstreetmap-carto/blob/258d354ae48d00e732ff389a08bc296d9d55eca9/style/roads.mss#L2301) and cyclemap layers.


tomhughes:
Does `highway=path` always get rendered as a footpath though? or does it depend on what tags it is combined with?

As I understand it (and it's not a tag I ever use) it's a generic tag that covers all sorts of paths and access tags or something are then used to provide the detail.

HolgerJeromin:
Tom is correct here:

https://www.openstreetmap.org/way/1112909185#map=18/52.51062/13.51854
Is rendered like a cycleway because of `bicycle=designated`

AntonKhorev:
> Does highway=path always get rendered as a footpath though?

in 95% of the cases:

    [feature = 'highway_path'][bicycle != 'designated'][horse != 'designated']

https://taginfo.openstreetmap.org/tags/highway%3Dpath#combinations

4.38%
[bicycle](https://taginfo.openstreetmap.org/keys/bicycle)=[designated](https://taginfo.openstreetmap.org/tags/bicycle=designated)

0.10%
[horse](https://taginfo.openstreetmap.org/keys/horse)=[designated](https://taginfo.openstreetmap.org/tags/horse=designated)

tomhughes:
But at the end of the day the text in the key is not supposed to be interpreted literally as a tag value, just generically as "this is a foot path".

It's a bit odd that we use `footway` in the key really as `footpath` would be more typical english.

Issue #2761, opened 2020-08-08
Explain to anonymous users they need to log in to comment on notes


boothym:
I was surveying and looking looking at a nearby note on my phone. I was not logged in so all the note displayed was the description, created by user/date, and the location.

Since anonymous users cannot leave a comment or resolve notes, it should be explained to them that they need to log in to do so. 

Something like "Log in to leave a comment on this note" could be added below - anonymous users may look at a note and have some local knowledge but it's not clear how to add a comment. 

Especially as anonymous users can create notes but then not reply to any new comments that others may post.

![Screenshot_2020-09-13 Note 1982902 OpenStreetMap](https://user-images.githubusercontent.com/16965772/93017439-af316f80-f5c0-11ea-8185-30de50b4077a.png)

mmd-osm:
I think everything on this topic has been discussed before in #1543 and #2303. Your requirement could probably be merged into #2303.

boothym:
Not sure about that. I'm not asking for anonymous users to be allowed to comment as per #1543 (it's a long thread though so perhaps I missed something). 

And maybe it's just not explained properly, but I didn't think #2303 covered my issue - anonymous users cannot even begin typing a comment to begin with as there's no comment box.

mmd-osm:
https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services suggests that you won’t be able to access an existing note unless you’re already logged in. So there’s probably no need to leave a message for anonymous users here that they can’t see anyway in the distant future.

boothym:
Didn't notice I was logged out until I looked at a note and saw the resolve/comment buttons were missing, which reminded me again of this issue.

If you view a changeset when logged out it says:

**Discussion
Log in to join the discussion**

So why can't it say "Log in to comment on this note"?

tomhughes:
Well of course it can say that - it can say anything we want it to.

That is just that the original implementer of the feature chose and nobody has ever opened a pull request to propose changing it.

boothym:
Looking at the changeset page this seems to be the code which adds the log in text: https://github.com/openstreetmap/openstreetmap-website/blob/d3388abe51a946ae0abc645d831a93b1b2cc6749/app/views/browse/changeset.html.erb#L69

` <% unless current_user %>
    <p class="notice">
      <%= link_to(t(".join_discussion"), login_path(:referer => request.fullpath)) %>
    </p>
  <% end %>`

Could the same be added to the note.html.erb file?

[...]

PR #4449, opened 2023-12-31
Add login notice for note discussions

AntonKhorev:
Add "Log in to comment on this note" link as requested in the [original post](https://github.com/openstreetmap/openstreetmap-website/issues/2761#issue-675527428) in https://github.com/openstreetmap/openstreetmap-website/issues/2761.

image

There also was [another request](https://github.com/openstreetmap/openstreetmap-website/issues/2761#issuecomment-1225816994) which I **didn't do**: adding "Log in to resolve/reactivate". Some users close and reopen notes several times for no apparent reason and previously we added texts to discourage that. I don't want to write anything that will encourage these actions.

### DIFF
--- a/app/views/notes/show.html.erb
+++ b/app/views/notes/show.html.erb
@@ -65,6 +65,10 @@
                                                    :url => comment_api_note_url(@note, "json") } %>
         </div>
       </form>
+    <% else -%>
+      <p>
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good, thanks.

PR #4450, opened 2023-12-31
Add cities to mapnik map key

AntonKhorev:
image


HolgerJeromin:
Can we rename "mapnik" to "osm-carto" in these files? In a separate PR.

tomhughes:
> Can we rename "mapnik" to "osm-carto" in these files? In a separate PR.

Well `standard` is the normal name used in the web site interface these days so it should probably use that name.

AntonKhorev:
It's also "Mapnik" in leaflet-osm
https://github.com/openstreetmap/leaflet-osm/blob/61737a7888bc83ad094077c34c4477087ae813a2/leaflet-osm.js#L15

Issue #2657, opened 2020-06-11
Default diary to user's language


boothym:
I clicked on User Diaries link to see what diaries had been posted recently, but the first 10 or so (this was yesterday) were not in a language I could read.

The only way currently to see the latest English language entries is to append "/en" to the url, or keep scrolling to find a post in English and then click on the "English (English)" link. 

Of course, this is also a problem when those who read other languages encounter lots of posts in English.

Would it be possible to add a link saying "| Show only English* language diary entries" next to the RSS | New Diary Entry | My Diary links? (* this would change to the preferred language of each user)

Even better would be if users could tick a box to in their settings, which would make the diary page default to their preferred language url ("diary/en"), and a "| Show all diary entries" link appear on the diary page.

MateoGreil:
I think you're right : all of this is not user-friendly.

> Would it be possible to add a link saying "| Show only English* language diary entries" next to the RSS | New Diary Entry | My Diary links? (* this would change to the preferred language of each user)

I can do it, I could even add a dropdown button to select among the preferred languages.

I think to select all of ours preferred languages could be better, but to do this, I need to edit the controller, and the routes, so I think I shall wait to get the confirmation of the main maintainers :
We could add a route `get "preferred_languages", :preferred_languages => true`.

Waiting for confirmations from the main maintainers.

tomhughes:
Why on earth would you need a route, never mind a global one.

There's no need to export the preferred languages to the browser to be able to return a filtered view.

MateoGreil:
Sorry, I misspoke, I wanted to say add a route `diary/preferred_languages`, of course not a global one.
Only accessible by authentified users and with preferred languages filled.

gravitystorm:
> this would change to the preferred language of each user

An important thing to note here is that we don't have "one preferred language" for each user. The preferred languages is a list of languages, so that needs to be taken into consideration.

Furthermore, I don't think we should hide large parts of the community from each other by default. I often see diary entries in languages that I can't read, but sometimes the pictures are enough (or mentions of something that I recognise, like 'JOSM') to make me want to translate them to find out what's going on. 

> Only accessible by authentified users and with preferred languages filled.

The preferred language setting should only be seen as an override to browser-specified defaults. So even for non-logged-in users we can detect the preferred languages. It's only that they don't have the opportunity to set a preference, not that we should ignore their preferred languages entirely.

MateoGreil:
[...]

PR #4451, opened 2024-01-02
Add links to diaries in preferred languages

AntonKhorev:
Adds a link requested in #2657 to https://www.openstreetmap.org/diary
Actually adds links for all preferred languages, using candidates logic https://github.com/openstreetmap/openstreetmap-website/blob/c95e2870f3310a68f8287bddcec6d0117b5f8961/lib/locale.rb#L44-L53

image

---

#2657 also requested:

> users could tick a box to in their settings, which would make the diary page default to their preferred language url

I don't think this is necessary. You can always bookmark diary pages in any specific language or subscribe to that language feed.

### DIFF
--- a/app/views/diary_entries/index.html.erb
+++ b/app/views/diary_entries/index.html.erb
@@ -14,15 +14,15 @@
         <ul class="clearfix">
           <% unless params[:friends] or params[:nearby] -%>
             <li><%= rss_link_to :action => "rss", :language => params[:language] %></li>
-            <% end -%>
+          <% end -%>
 
-            <% if @user && @user == current_user || !@user && current_user %>
-              <li><%= link_to image_tag("new.png", :class => "small_icon") + t(".new"), new_diary_entry_path, :title => t(".new_title") %></li>
-            <% end %>
+          <% if @user && @user == current_user || !@user && current_user %>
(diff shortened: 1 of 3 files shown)


app/views/diary_entries/index.html.erb
```diff
@@ -14,15 +14,19 @@
         <ul class="clearfix">
           <% unless params[:friends] or params[:nearby] -%>
             <li><%= rss_link_to :action => "rss", :language => params[:language] %></li>
-            <% end -%>
+          <% end -%>
 
-            <% if @user && @user == current_user || !@user && current_user %>
-              <li><%= link_to image_tag("new.png", :class => "small_icon") + t(".new"), new_diary_entry_path, :title => t(".new_title") %></li>
-            <% end %>
+          <% @languages&.each do |language| %>
```
tomhughes, line 19: I think these should probably go after "New Diary Entry" and before "My Diary" they're links to a subset of the diary in much the same way that "My Diary" is so they should probably be together.

tomhughes (changes requested):

app/views/diary_entries/index.html.erb
AntonKhorev: What I originally wanted to do is to move the "New Diary Entry" to the end, but I decided not to do it in the same pull request.

I moved it now:

image

tomhughes (approved):

tomhughes:
Looks great now, thanks.

PR #4452, opened 2024-01-02
Don't show native language name if it's the same as English name

AntonKhorev:
Languages have names in the format of "English name (native name)". These names are used in diary entry language links and inputs.

image

Sometimes *native name* is exactly the same as *English name*, so the same thing is displayed twice:

image

I think we don't need "(native name)" in this case:

image


tomhughes:
Merged, thanks.

PR #4454, opened 2024-01-03
Use assert_not_predicate in tests that have assert_predicate

AntonKhorev:
You have a test that looks like this:

https://github.com/openstreetmap/openstreetmap-website/blob/04c697b468057d1a3604694e7ead9d2c6efd1dfa/test/models/note_test.rb#L40-L42

Then Rubocop starts complaining that you don't use `assert_predicate` and you change the test:

https://github.com/openstreetmap/openstreetmap-website/blob/6892ccd0155b431cc567b4744b1e2759df3e8c68/test/models/note_test.rb#L40-L42

Yet you keep negative assertions as `assert_not` and don't change then to `assert_not_predicate`. Are there any good reasons for that other than Rubocop no complaining?

```ruby
assert_not_predicate create(:note, :status => "hidden"), :visible?
```

### DIFF
--- a/test/lib/bounding_box_test.rb
+++ b/test/lib/bounding_box_test.rb
@@ -222,7 +222,7 @@ class BoundingBoxTest < ActiveSupport::TestCase
   end
 
   def test_complete
-    assert_not @bbox_from_nils.complete?, "should contain a nil"
+    assert_not_predicate @bbox_from_nils, :complete?, "should contain a nil"
     assert_predicate @bbox_from_string, :complete?, "should not contain a nil"
   end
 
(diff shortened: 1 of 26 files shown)


tomhughes:
The [minitest style guide](https://minitest.rubystyle.guide/#refute-predicate) actually suggests using `refute_predicate` here and rubocop-minitest has a [cop](https://docs.rubocop.org/rubocop-minitest/cops_minitest.html#minitestrefutepredicate) to detect `refute` and suggest `refute_predicate` but it misses `assert_not`...

tomhughes:
The problem of course is that the rails cop effectively stops all the refute cops in rubcop-minitest working :cry:

PR #4455, opened 2024-01-03
Re-arrange login and signup screens as discussed in #4128

milan-cvetkovic:
Implements proposal discussed in [Issue 4128](https://github.com/openstreetmap/openstreetmap-website/issues/4128).

- Merge Signup screen and Terms screen during signup process.
- Use links to Contributor terms and Terms of use instead of in-place edit box.
- Terms screen used for accepting the new terms left unchanged.
- Re-arranged signup and login screens to feel like two tabs as in #4128.
- Add `preferred_auth_provider` parameter to signup and login screens, as well as the authorization request.
  This allows 3rd party apps to initiate authorization by adding `preferred_auth_provider`
  
  Screenshots in dev environment:

image
  
image

image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -726,6 +726,7 @@ tr.turn {
 
   &.new-user-main {
     background-image: image-url("sign-up-illustration.png");
+    background-position-x: 50px;
   }
 
   &.confirm-main {
@@ -734,17 +735,6 @@ tr.turn {
 
   &.new-user-terms {
     background-image: image-url("terms-illustration.png");
-  }
-
-  &.new-user-arm {
-    height: 110px;
-    width: 130px;
-    left: 280px;
-    top: 180px;
-    background-image: image-url("sign-up-illustration-arm.png");
-    position: absolute;
-    z-index: 100;
-    pointer-events: none;
   }
 }
 
@@ -1016,6 +1006,10 @@ div.secondary-actions {
       background-position: -45px -160px;
     }
   }
+}
+
+.auth-container {
+  max-width: 600px;
 }
 
 /* Rules for tabs inside secondary background sections */
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -15,6 +15,8 @@ class SessionsController < ApplicationController
     override_content_security_policy_directives(:form_action => []) if Settings.csp_enforce || Settings.key?(:csp_report_url)
 
     session[:referer] = safe_referer(params[:referer]) if params[:referer]
+
+    parse_oauth_referer session[:referer]
   end
 
   def create
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -78,10 +78,11 @@ class UsersController < ApplicationController
                                    :auth_provider => params[:auth_provider],
                                    :auth_uid => params[:auth_uid])
 
-      flash.now[:notice] = render_to_string :partial => "auth_association"
-
-      # validate before displaying to show email field if it conflicts
-      flash.now[:notice] = t ".duplicate_social_email" unless current_user.valid? && current_user.errors[:email].empty?
+      if current_user.valid? || current_user.errors[:email].empty?
+        flash.now[:notice] = render_to_string :partial => "auth_association"
+      else
+        flash.now[:warning] = t ".duplicate_social_email"
+      end
     else
       check_signup_allowed
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -60,6 +60,8 @@ class UsersController < ApplicationController
                  session[:referer]
                end
 
+    parse_oauth_referer @referer
+
     append_content_security_policy_directives(
       :form_action => %w[accounts.google.com *.facebook.com login.live.com github.com meta.wikimedia.org]
     )
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -70,7 +70,6 @@ class UsersController < ApplicationController
       redirect_to @referer || { :controller => "site", :action => "index" }
     elsif params.key?(:auth_provider) && params.key?(:auth_uid)
       self.current_user = User.new(:email => params[:email],
-                                   :email_confirmation => params[:email],
                                    :display_name => params[:nickname],
                                    :auth_provider => params[:auth_provider],
                                    :auth_uid => params[:auth_uid])
@@ -336,7 +335,7 @@ class UsersController < ApplicationController
   ##
   # return permitted user parameters
   def user_params
-    params.require(:user).permit(:email, :email_confirmation, :display_name,
+    params.require(:user).permit(:email, :display_name,
                                  :auth_provider, :auth_uid,
                                  :pass_crypt, :pass_crypt_confirmation,
                                  :consider_pd)
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -69,12 +69,17 @@ class UsersController < ApplicationController
       # page, instead send them to the home page
       redirect_to @referer || { :controller => "site", :action => "index" }
     elsif params.key?(:auth_provider) && params.key?(:auth_uid)
+      @verified_email = params[:verified_email]
+
       self.current_user = User.new(:email => params[:email],
                                    :display_name => params[:nickname],
                                    :auth_provider => params[:auth_provider],
                                    :auth_uid => params[:auth_uid])
 
       flash.now[:notice] = render_to_string :partial => "auth_association"
+
+      # validate before displaying to show email field if it conflicts
+      flash.now[:notice] = t ".duplicate_social_email" unless current_user.valid? && current_user.errors[:email].empty?
     else
       check_signup_allowed
 
@@ -90,7 +95,7 @@ class UsersController < ApplicationController
 
       Rails.logger.info "create: #{session[:referer]}"
 
-      if current_user.auth_provider.present? && current_user.pass_crypt.empty?
+      if current_user.auth_uid.present?
         # We are creating an account with external authentication and
         # no password was specified so create a random one
         current_user.pass_crypt = SecureRandom.base64(16)
@@ -100,14 +105,9 @@ class UsersController < ApplicationController
       if current_user.invalid?
         # Something is wrong with a new user, so rerender the form
         render :action => "new"
-      elsif current_user.auth_provider.present?
-        # Verify external authenticator before moving on
-        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation", "consider_pd")
-        redirect_to auth_url(current_user.auth_provider, current_user.auth_uid), :status => :temporary_redirect
       else
         # Save the user record
-        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation", "consider_pd")
-        save_new_user
+        save_new_user params[:verified_email]
       end
     end
   end
@@ -131,7 +131,7 @@ class UsersController < ApplicationController
       if current_user&.terms_agreed?
         # Already agreed to terms, so just show settings
         redirect_to edit_account_path
-      elsif current_user.nil? && session[:new_user].nil?
+      elsif current_user.nil?
         redirect_to login_path(:referer => request.fullpath)
       end
     end
@@ -220,12 +220,6 @@ class UsersController < ApplicationController
       session[:user_errors] = current_user.errors.as_json
 
       redirect_to edit_account_path
-    elsif session[:new_user]
-      session[:new_user]["auth_provider"] = provider
-      session[:new_user]["auth_uid"] = uid
-      session[:new_user]["verified_email"] = email if email_verified
-
-      save_new_user
     else
       user = User.find_by(:auth_provider => provider, :auth_uid => uid)
 
@@ -247,7 +241,8 @@ class UsersController < ApplicationController
           failed_login t("sessions.new.auth failure")
         end
       else
-        redirect_to :action => "new", :nickname => name, :email => email,
+        verified_email = UsersController.message_hmac(email) if email_verified && email
+        redirect_to :action => "new", :nickname => name, :email => email, :verified_email => verified_email,
                     :auth_provider => provider, :auth_uid => uid
       end
     end
@@ -263,14 +258,16 @@ class UsersController < ApplicationController
     redirect_to origin || login_url
   end
 
+  def self.message_hmac(text)
+    sha256 = Digest::SHA256.new
+    sha256 << Rails.application.key_generator.generate_key("openstreetmap/verified_email")
+    sha256 << text
+    Base64.urlsafe_encode64(sha256.digest)
+  end
+
   private
 
-  def save_new_user
-    new_user = session.delete(:new_user)
-    verified_email = new_user.delete("verified_email")
-
-    self.current_user = User.new(new_user)
-
+  def save_new_user(email_hmac)
     if check_signup_allowed(current_user.email)
       current_user.data_public = true
       current_user.description = "" if current_user.description.nil?
@@ -283,7 +280,7 @@ class UsersController < ApplicationController
       if current_user.auth_uid.blank?
         current_user.auth_provider = nil
         current_user.auth_uid = nil
-      elsif current_user.email == verified_email
+      elsif email_hmac && ActiveSupport::SecurityUtils.secure_compare(email_hmac, UsersController.message_hmac(current_user.email))
         current_user.activate
       end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -103,12 +103,12 @@ class UsersController < ApplicationController
         render :action => "new"
       elsif current_user.auth_provider.present?
         # Verify external authenticator before moving on
-        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation")
+        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation", "consider_pd")
         redirect_to auth_url(current_user.auth_provider, current_user.auth_uid), :status => :temporary_redirect
       else
         # Save the user record
-        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation")
-        redirect_to :action => :terms
+        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation", "consider_pd")
+        save_new_user
       end
     end
   end
@@ -168,48 +168,6 @@ class UsersController < ApplicationController
       referer = safe_referer(params[:referer]) if params[:referer]
 
       redirect_to referer || edit_account_path
-    else
-      new_user = session.delete(:new_user)
-      verified_email = new_user.delete("verified_email")
-
-      self.current_user = User.new(new_user)
-
-      if check_signup_allowed(current_user.email)
-        current_user.data_public = true
-        current_user.description = "" if current_user.description.nil?
-        current_user.creation_ip = request.remote_ip
-        current_user.languages = http_accept_language.user_preferred_languages
-        current_user.terms_agreed = Time.now.utc
-        current_user.tou_agreed = Time.now.utc
-        current_user.terms_seen = true
-
-        if current_user.auth_uid.blank?
-          current_user.auth_provider = nil
-          current_user.auth_uid = nil
-        elsif current_user.email == verified_email
-          current_user.activate
-        end
-
-        if current_user.save
-          SIGNUP_IP_LIMITER&.update(request.remote_ip)
-          SIGNUP_EMAIL_LIMITER&.update(canonical_email(current_user.email))
-
-          flash[:matomo_goal] = Settings.matomo["goals"]["signup"] if defined?(Settings.matomo)
-
-          referer = welcome_path(welcome_options)
-
-          if current_user.status == "active"
-            session[:referer] = referer
-            successful_login(current_user)
-          else
-            session[:pending_user] = current_user.id
-            UserMailer.signup_confirm(current_user, current_user.generate_token_for(:new_user), referer).deliver_later
-            redirect_to :controller => :confirmations, :action => :confirm, :display_name => current_user.display_name
-          end
-        else
-          render :action => "new", :referer => params[:referer]
-        end
-      end
     end
   end
 
@@ -268,7 +226,7 @@ class UsersController < ApplicationController
       session[:new_user]["auth_uid"] = uid
       session[:new_user]["verified_email"] = email if email_verified
 
-      redirect_to :action => "terms"
+      save_new_user
     else
       user = User.find_by(:auth_provider => provider, :auth_uid => uid)
 
@@ -308,6 +266,50 @@ class UsersController < ApplicationController
 
   private
 
+  def save_new_user
+    new_user = session.delete(:new_user)
+    verified_email = new_user.delete("verified_email")
+
+    self.current_user = User.new(new_user)
+
+    if check_signup_allowed(current_user.email)
+      current_user.data_public = true
+      current_user.description = "" if current_user.description.nil?
+      current_user.creation_ip = request.remote_ip
+      current_user.languages = http_accept_language.user_preferred_languages
+      current_user.terms_agreed = Time.now.utc
+      current_user.tou_agreed = Time.now.utc
+      current_user.terms_seen = true
+
+      if current_user.auth_uid.blank?
+        current_user.auth_provider = nil
+        current_user.auth_uid = nil
+      elsif current_user.email == verified_email
+        current_user.activate
+      end
+
+      if current_user.save
+        SIGNUP_IP_LIMITER&.update(request.remote_ip)
+        SIGNUP_EMAIL_LIMITER&.update(canonical_email(current_user.email))
+
+        flash[:matomo_goal] = Settings.matomo["goals"]["signup"] if defined?(Settings.matomo)
+
+        referer = welcome_path(welcome_options)
+
+        if current_user.status == "active"
+          session[:referer] = referer
+          successful_login(current_user)
+        else
+          session[:pending_user] = current_user.id
+          UserMailer.signup_confirm(current_user, current_user.generate_token_for(:new_user), referer).deliver_later
+          redirect_to :controller => :confirmations, :action => :confirm, :display_name => current_user.display_name
+        end
+      else
+        render :action => "new", :referer => params[:referer]
+      end
+    end
+  end
+
   def welcome_options
     uri = URI(session[:referer]) if session[:referer].present?
 
@@ -336,7 +338,8 @@ class UsersController < ApplicationController
   def user_params
     params.require(:user).permit(:email, :email_confirmation, :display_name,
                                  :auth_provider, :auth_uid,
-                                 :pass_crypt, :pass_crypt_confirmation)
+                                 :pass_crypt, :pass_crypt_confirmation,
+                                 :consider_pd)
   end
 
   ##
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -107,8 +107,13 @@ class UsersController < ApplicationController
       if current_user.invalid?
         # Something is wrong with a new user, so rerender the form
         render :action => "new"
+      elsif current_user.auth_provider.present?
+        # Verify external authenticator before moving on
+        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation")
+        redirect_to auth_url(current_user.auth_provider, current_user.auth_uid), :status => :temporary_redirect
       else
         # Save the user record
+        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation")
         save_new_user params[:email_hmac]
       end
     end
@@ -222,6 +227,12 @@ class UsersController < ApplicationController
       session[:user_errors] = current_user.errors.as_json
 
       redirect_to edit_account_path
+    elsif session[:new_user]
+      session[:new_user]["auth_provider"] = provider
+      session[:new_user]["auth_uid"] = uid
+
+      email_hmac = UsersController.message_hmac(email) if email_verified && email
+      save_new_user email_hmac
     else
       user = User.find_by(:auth_provider => provider, :auth_uid => uid)
 
@@ -270,6 +281,8 @@ class UsersController < ApplicationController
   private
 
   def save_new_user(email_hmac)
+    new_user = session.delete(:new_user)
+    self.current_user = User.new(new_user)
     if check_signup_allowed(current_user.email)
       current_user.data_public = true
       current_user.description = "" if current_user.description.nil?
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -71,7 +71,7 @@ class UsersController < ApplicationController
       # page, instead send them to the home page
       redirect_to @referer || { :controller => "site", :action => "index" }
     elsif params.key?(:auth_provider) && params.key?(:auth_uid)
-      @verified_email = params[:verified_email]
+      @email_hmac = params[:email_hmac]
 
       self.current_user = User.new(:email => params[:email],
                                    :display_name => params[:nickname],
@@ -109,7 +109,7 @@ class UsersController < ApplicationController
         render :action => "new"
       else
         # Save the user record
-        save_new_user params[:verified_email]
+        save_new_user params[:email_hmac]
       end
     end
   end
@@ -243,8 +243,8 @@ class UsersController < ApplicationController
           failed_login t("sessions.new.auth failure")
         end
       else
-        verified_email = UsersController.message_hmac(email) if email_verified && email
-        redirect_to :action => "new", :nickname => name, :email => email, :verified_email => verified_email,
+        email_hmac = UsersController.message_hmac(email) if email_verified && email
+        redirect_to :action => "new", :nickname => name, :email => email, :email_hmac => email_hmac,
                     :auth_provider => provider, :auth_uid => uid
       end
     end
@@ -262,7 +262,7 @@ class UsersController < ApplicationController
 
   def self.message_hmac(text)
     sha256 = Digest::SHA256.new
-    sha256 << Rails.application.key_generator.generate_key("openstreetmap/verified_email")
+    sha256 << Rails.application.key_generator.generate_key("openstreetmap/email_address")
     sha256 << text
     Base64.urlsafe_encode64(sha256.digest)
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -100,7 +100,7 @@ class User < ApplicationRecord
                            :whitespace => { :leading => false, :trailing => false },
                            :width => { :minimum => 3 }
   validate :display_name_cannot_be_user_id_with_other_id, :if => proc { |u| u.display_name_changed? }
-  validates :email, :presence => true, :confirmation => true, :characters => true
+  validates :email, :presence => true, :characters => true
   validates :email, :if => proc { |u| u.email_changed? },
                     :uniqueness => { :case_sensitive => false }
   validates :email, :if => proc { |u| u.email_changed? },
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -104,7 +104,7 @@
           <%= link_to t("layouts.logout"), logout_path(:referer => request.fullpath), :method => "post", :class => "geolink dropdown-item" %>
         </div>
       </div>
-    <% else %>
+    <% elsif (controller_name != "users" and controller_name != "sessions") || action_name != "new" %>
       <div class="d-inline-flex btn-group login-menu" role="">
         <%= link_to t("layouts.log_in"), login_path(:referer => request.fullpath), :class => "geolink btn btn-outline-secondary" %>
         <%= link_to t("layouts.sign_up"), user_new_path, :class => "btn btn-outline-secondary" %>
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -6,6 +6,10 @@
 <% content_for :heading_class, "p-0 mw-100" %>
 
 <% content_for :heading do %>
+  <% if @client_app_name %>
+    <p class="text-center text-muted fs-6 py-2 mb-0 bg-white"><%= t(".login_to_authorize_html", :client_app_name => @client_app_name) %></p>
+  <% end %>
+
   <div class="header-illustration new-user-main auth-container mx-auto">
     <ul class="nav nav-tabs position-absolute bottom-0 px-3 fs-6 w-100">
       <li class="nav-item">
@@ -19,7 +23,14 @@
 <% end %>
 
 <div id="login_login" class="auth-container mx-auto my-0">
-  <p class='text-muted'><%= t ".no account" %> <%= link_to t(".register now"), user_new_path(:referer => params[:referer]) %></p>
+  <% if @preferred_auth_provider %>
+    <%= render :partial => "auth_providers" %>
+    <div class="d-flex justify-content-center align-items-center">
+      <div class="border-bottom border-1 flex-grow-1"></div>
+      <div class="text-secondary mx-3"><%= t ".or" %></div>
+      <div class="border-bottom border-1 flex-grow-1"></div>
+    </div>
+  <% end %>
 
   <%= bootstrap_form_tag(:action => "login", :html => { :id => "login_form" }) do |f| %>
     <%= hidden_field_tag("referer", h(params[:referer]), :autocomplete => "off") %>
@@ -40,10 +51,17 @@
       <%= f.check_box :remember_me, { :label => t(".remember"), :tabindex => 3, :checked => (params[:remember_me] == "yes") }, "yes" %>
     <% end %>
 
-    <%= f.primary t(".login_button"), :tabindex => 4 %>
+    <div class="mb-3">
+      <%= f.primary t(".login_button"), :tabindex => 4 %>
+    </div>
   <% end %>
 
-  <hr>
-
-  <%= render :partial => "auth_providers" %>
+  <% unless @preferred_auth_provider %>
+    <div class="d-flex justify-content-center align-items-center">
+      <div class="border-bottom border-1 flex-grow-1"></div>
+      <div class="text-secondary mx-3"><%= t ".with external" %></div>
+      <div class="border-bottom border-1 flex-grow-1"></div>
+    </div>
+    <%= render :partial => "auth_providers" %>
+  <% end %>
 </div>
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -7,8 +7,15 @@
 
 <% content_for :heading do %>
   <div class="header-illustration new-user-main auth-container mx-auto">
-    <h1 class="pt-3"><%= t ".heading" %></h1>
-  <div>
+    <ul class="nav nav-tabs position-absolute bottom-0 px-3 fs-6 w-100">
+      <li class="nav-item">
+        <%= link_to t("sessions.new.tab_title"), "#", :class => "nav-link active" %>
+      </li>
+      <li class="nav-item">
+        <%= link_to t("users.new.tab_title"), url_for(:action => :new, :controller => :users), :class => "nav-link" %>
+      </li>
+    </ul>
+  </div>
 <% end %>
 
 <div id="login_login" class="auth-container mx-auto my-0">
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -1,5 +1,6 @@
 <% content_for :head do %>
   <%= javascript_include_tag "login" %>
+  <%= javascript_include_tag "auth_providers" %>
 <% end %>
 
 <% content_for :heading_class, "p-0 mw-100" %>
@@ -37,37 +38,5 @@
 
   <hr>
 
-  <div id="loginForm">
-    <div class="mb-3">
-      <label class="form-label"><%= t ".with external" %></label>
-
-      <ul class='list-inline' id="login_auth_buttons">
-        <li class="list-inline-item me-3">
-          <%= link_to image_tag("openid.png",
-                                :alt => t(".auth_providers.openid.title"),
-                                :size => "36"),
-                      "#",
-                      :id => "openid_open_url",
-                      :title => t(".auth_providers.openid.title") %>
-        </li>
-
-        <% %w[google facebook microsoft github wikipedia].each do |provider| %>
-          <% if Settings.key?("#{provider}_auth_id".to_sym) -%>
-            <li class="list-inline-item me-3"><%= auth_button provider, provider %></li>
-          <% end -%>
-        <% end -%>
-      </ul>
-
-      <%= form_tag(auth_path(:provider => "openid"), :id => "openid_login_form") do %>
-        <div id='login_openid_url' class="mb-3">
-          <label for='openid_url' class="form-label"><%= t ".openid_html", :logo => openid_logo %></label>
-          <%= hidden_field_tag("referer", params[:referer], :autocomplete => "off") %>
-          <%= text_field_tag("openid_url", "", :tabindex => 5, :autocomplete => "on", :class => "openid_url form-control") %>
-          <span class="form-text text-muted">(<a href="<%= t "accounts.edit.openid.link" %>" target="_new"><%= t "accounts.edit.openid.link text" %></a>)</span>
-        </div>
-
-        <%= submit_tag t(".login_button"), :tabindex => 6, :id => "login_openid_submit", :class => "btn btn-primary" %>
-      <% end %>
-    </div>
-  </div>
+  <%= render :partial => "auth_providers" %>
 </div>
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -17,7 +17,17 @@
     <%= hidden_field_tag("referer", h(params[:referer]), :autocomplete => "off") %>
 
     <%= f.text_field :username, :label => t(".email or username"), :tabindex => 1, :value => params[:username] %>
-    <%= f.password_field :password, :label => t(".password"), :tabindex => 2, :value => "", :help => link_to(t(".lost password link"), user_forgot_password_path) %>
+
+    <div class="row">
+      <div class="col">
+        <%= f.label :password, :class => "form-label" %>
+      </div>
+      <div class="col text-end">
+        <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
+      </div>
+    </div>
+    <input class="form-control mb-3" type="password" name="password" id="password" tabindex="2" value="" autocomplete="off" />
+
     <%= f.form_group do %>
       <%= f.check_box :remember_me, { :label => t(".remember"), :tabindex => 3, :checked => (params[:remember_me] == "yes") }, "yes" %>
     <% end %>
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -2,11 +2,15 @@
   <%= javascript_include_tag "login" %>
 <% end %>
 
+<% content_for :heading_class, "p-0 mw-100" %>
+
 <% content_for :heading do %>
-  <h1><%= t ".heading" %></h1>
+  <div class="header-illustration new-user-main auth-container mx-auto">
+    <h1 class="pt-3"><%= t ".heading" %></h1>
+  <div>
 <% end %>
 
-<div id="login_login">
+<div id="login_login" class="auth-container mx-auto my-0">
   <p class='text-muted'><%= t ".no account" %> <%= link_to t(".register now"), user_new_path(:referer => params[:referer]) %></p>
 
   <%= bootstrap_form_tag(:action => "login", :html => { :id => "login_form" }) do |f| %>
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -2,51 +2,48 @@
   <%= javascript_include_tag "user" %>
 <% end %>
 
-<% content_for :heading_class, "pb-0" %>
+<% content_for :heading_class, "p-0 mw-100" %>
+
 <% content_for :heading do %>
-  <div class='header-illustration new-user-main'>
-    <h1><%= t ".title" %></h1>
+  <div class="header-illustration new-user-main auth-container mx-auto">
+    <h1 class="pt-3"><%= t ".title" %></h1>
   </div>
-  <div class='header-illustration new-user-arm d-none d-md-block'></div>
 <% end %>
 
-<div class="row">
-  <div class='text-muted col-sm order-sm-2'>
-    <h4><%= t ".about.header" %></h4>
-    <p><%= t ".about.paragraph_1" %></p>
+<div class="auth-container mx-auto my-0">
+  <div class="text-muted fs-6">
+    <p><strong><%= t ".about.header" %></strong> <%= t ".about.paragraph_1" %></p>
     <p><%= t ".about.paragraph_2" %></p>
   </div>
 
-  <div class="col-sm">
-    <%= bootstrap_form_for current_user, :url => { :action => "create" } do |f| %>
-      <%= hidden_field_tag("referer", h(@referer)) unless @referer.nil? %>
+  <%= bootstrap_form_for current_user, :url => { :action => "create" } do |f| %>
+    <%= hidden_field_tag("referer", h(@referer)) unless @referer.nil? %>
 
-      <%= f.email_field :email, :tabindex => 1 %>
-      <%= f.email_field :email_confirmation, :help => t(".email_confirmation_help_html",
-                                                        :privacy_policy_link => link_to(t(".privacy_policy"),
-                                                                                        t(".privacy_policy_url"),
-                                                                                        :title => t(".privacy_policy_title"))),
-                                             :tabindex => 2 %>
+    <%= f.email_field :email, :tabindex => 1 %>
+    <%= f.email_field :email_confirmation, :help => t(".email_confirmation_help_html",
+                                                      :privacy_policy_link => link_to(t(".privacy_policy"),
+                                                                                      t(".privacy_policy_url"),
+                                                                                      :title => t(".privacy_policy_title"))),
+                                           :tabindex => 2 %>
 
-      <%= f.text_field :display_name, :help => t(".display name description"), :tabindex => 3 %>
+    <%= f.text_field :display_name, :help => t(".display name description"), :tabindex => 3 %>
 
-      <fieldset class="mb-3" id="auth_field">
-        <label for="user_auth_provider" class="form-label"><%= t(".external auth") %></label>
-        <div class="row">
-          <%= f.select(:auth_provider, Auth.providers, :default => "", :hide_label => true, :wrapper => { :class => "col-auto mb-0" }, :tabindex => 4) %>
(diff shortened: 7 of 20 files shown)


tomhughes:
Both @gravitystorm and I were very clear in #4128 that hiding the terms behind a link would not be acceptable unless LWG signed off on it. Has that happened?

What is the final screenshot? Is that what you get if use one of the third party signup buttons?

app/views/sessions/new.html.erb
```diff
@@ -1,51 +1,36 @@
...
-
-    <%= f.text_field :username, :label => t(".email or username"), :tabindex => 1, :value => params[:username] %>
-    <%= f.password_field :password, :label => t(".password"), :tabindex => 2, :value => "", :help => link_to(t(".lost password link"), user_forgot_password_path) %>
-    <%= f.form_group do %>
-      <%= f.check_box :remember_me, { :label => t(".remember"), :tabindex => 3, :checked => (params[:remember_me] == "yes") }, "yes" %>
-    <% end %>
+<div id="login_login" class="auth-container">
+  <div class="tabs">
+    <button class="tab-button active" disabled="true"><%= t ".tab_title" %></button>
+    <button class="tab-button" onclick="location.pathname='<%= url_for(:action => :new, :controller => :users) -%>'"><%= t "users.new.tab_title" %></button>
```
AntonKhorev, line 9: Why onclick? Why button? Why isn't it a link?

https://getbootstrap.com/docs/5.3/components/navs-tabs/#tabs

tomhughes: The `onlick` won't work anyway as our security policy doesn't allow inline javascript.

app/views/users/new.html.erb
```diff
@@ -1,52 +1,68 @@
...
+  <div class="tabs">
+    <button class="tab-button" onclick="location.pathname='<%= url_for(:action => :new, :controller => :sessions) -%>'"><%= t "sessions.new.tab_title" %></button>
+    <button class="tab-button active" disabled="true"><%= t ".tab_title" %></button>
   </div>
-  <div class='header-illustration new-user-arm d-none d-md-block'></div>
-<% end %>
 
-<div class="row">
-  <div class='text-muted col-sm order-sm-2'>
+  <div class="text-muted col-sm form-container">
```
AntonKhorev, line 12: `text-muted` on everything, even on input labels? That's not how they look on images in #4128.

milan-cvetkovic:
> Both @gravitystorm and I were very clear in #4128 that hiding the terms behind a link would not be acceptable unless LWG signed off on it. Has that happened?

@tomhughes " - we already asked LWG some time ago about it and @kathleenlu gave permission and recommendation (either links or checkboxes were fine, and we opted for links). We will forward you the mail and Kathleen can comment here as well.

milan-cvetkovic:
@tomhughes 

> What is the final screenshot? Is that what you get if use one of the third party signup buttons?

Yes, it is on the proposal labeled as "205".

It is simplified version of "Signup" screen for the user to enters minimal information after creating the OSM account linked to 3rd party provider - the email address is picked up from the 3rd party provider if available. If display name or email is not available, clicking "Signup" would highlight that field and add a note that the value is already in use. In case of email conflict, it would display it in edit box, forcing user to modify it, and go through the "verify email" scenario.

tomhughes:
> It is simplified version of "Signup" screen for the user to enters minimal information after creating the OSM account linked to 3rd party provider - the email address is picked up from the 3rd party provider if available. If display name or email is not available, clicking "Signup" would highlight that field and add a note that the value is already in use. In case of email conflict, it would display it in edit box, forcing user to modify it, and go through the "verify email" scenario.

So the email field would appear if the OAuth didn't return one?

But it does remove the ability for the user to change that email (if one was returned) or set a password as part of the signup.

app/views/sessions/new.html.erb
milan-cvetkovic: Fixed to use links instead.

app/views/users/new.html.erb
milan-cvetkovic: Good catch, thanks.

milan-cvetkovic:
> > It is simplified version of "Signup" screen for the user to enters minimal information after creating the OSM account linked to 3rd party provider - the email address is picked up from the 3rd party provider if available. If display name or email is not available, clicking "Signup" would highlight that field and add a note that the value is already in use. In case of email conflict, it would display it in edit box, forcing user to modify it, and go through the "verify email" scenario.
> 
> So the email field would appear if the OAuth didn't return one?

Yes, it should.

 
> But it does remove the ability for the user to change that email (if one was returned) or set a password as part of the signup.

Yes it does. That is the tradeoff for simplifying the signup process. User can modify those fields later in their settings.

app/controllers/sessions_controller.rb
```diff
@@ -23,6 +23,7 @@ def new
         ref_params = CGI.parse referer_query
         preferred = ref_params["preferred_auth_provider"].first
         @preferred_auth_provider = preferred if preferred && Settings.key?(:"#{preferred}_auth_id")
+        @client_app_name = Oauth2Application.find_by(:uid => ref_params["client_id"].first)&.name if ref_params["client_id"].first
```
tomhughes, line 26: This is probably better as:

```ruby
Oauth2Application.where(:uid => ref_params["client_id"].first).pluck(:name).first
```

app/controllers/users_controller.rb
```diff
@@ -68,6 +68,7 @@ def new
         ref_params = CGI.parse referer_query
         preferred = ref_params["preferred_auth_provider"].first
         @preferred_auth_provider = preferred if preferred && Settings.key?(:"#{preferred}_auth_id")
+        @client_app_name = Oauth2Application.find_by(:uid => ref_params["client_id"].first)&.name if ref_params["client_id"].first
```
tomhughes, line 71: Same here...

app/views/users/new.html.erb
```diff
@@ -4,6 +4,10 @@
 <% end %>
 
 <div class="auth-container">
+  <% if @client_app_name %>
+    <p class="association-note mt-4"><%= t(".login_to_authorize_html", :client_app_name => @client_app_name) %></p>
```
tomhughes, line 8: That translation name should probably be `signup_to_authorize_html` here? Though see later comment about the translations...

config/locales/en.yml
```diff
@@ -1808,6 +1808,7 @@ en:
     new:
       title: "Login"
       tab_title: "Login"
+      login_to_authorize_html: "You need an OpenStreeMap account to access %{client_app_name}."
```
tomhughes, line 1811: Should this be more like `Login to OpenStreetMap to access...` maybe? At the very least it needs to spell OpenStreetMap correctly ;-)

config/locales/en.yml
```diff
@@ -2695,6 +2696,7 @@ en:
     new:
       title: "Sign Up"
       tab_title: "Sign up"
+      login_to_authorize_html: "You need an OpenStreeMap account to access %{client_app_name}."
```
tomhughes, line 2699: Same here, maybe `Signup with OpenStreetMap to access...` or if we are going to have the same text in both places we should try and share one resource.

tomhughes (changes requested):
A first pass review, mostly concentrating on the server side.

I'm sure @gravitystorm will have more thoughts than me on the client side and bootstrap and whether there's any way to reduce the amount of custom CSS this is adding.

My main architectural question is whether we can avoid creating the two new actions, which creates a lot of duplication that is hard to avoid, and just use conditional sections in the views to change what is shown?

kathleenlu09:
@tomhughes Yes LWG has reviewed the latest screenshots, and we have no objections

app/controllers/users_controller.rb
```diff
@@ -76,39 +118,37 @@ def new
...
-      if current_user.invalid?
+      if current_user.invalid? || current_user.auth_provider.blank?
         # Something is wrong with a new user, so rerender the form
-        render :action => "new"
-      elsif current_user.auth_provider.present?
+        render :action => "new_association"
+      else
         # Verify external authenticator before moving on
         session[:new_user] = current_user.attributes.slice("email", "display_name", "pass_crypt", "consider_pd")
         redirect_to auth_url(current_user.auth_provider, current_user.auth_uid), :status => :temporary_redirect
```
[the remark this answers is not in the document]
milan-cvetkovic: I have re-arranged the code to avoid round-trip to the provider, and redirect to new-association instead. I also removed the use of `session[:new_user]`, since all the info from 3rd party provider is already present when rendering the page. In order to prevent users from tempering the email address received from provider, I pass around sha256 hmac instead of storing it in `session[:new_user]["verified_email"]`.

app/controllers/users_controller.rb
```diff
@@ -104,12 +104,12 @@ def create
...
       elsif current_user.auth_provider.present?
         # Verify external authenticator before moving on
-        session[:new_user] = current_user.attributes.slice("email", "display_name", "pass_crypt")
+        session[:new_user] = current_user.attributes.slice("email", "display_name", "pass_crypt", "consider_pd")
         redirect_to auth_url(current_user.auth_provider, current_user.auth_uid), :status => :temporary_redirect
       else
         # Save the user record
-        session[:new_user] = current_user.attributes.slice("email", "display_name", "pass_crypt")
-        redirect_to :action => :terms
+        session[:new_user] = current_user.attributes.slice("email", "display_name", "pass_crypt", "consider_pd")
```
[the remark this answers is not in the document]
milan-cvetkovic: Re-arranged the code not to use `session[:new_user]`

app/views/users/new_association.erb
```diff
@@ -0,0 +1,60 @@
...
+  <div class="text-muted col-sm form-container">
+
+    <h4><%= t ".welcome" %></h4>
+
+    <%= bootstrap_form_for current_user, :url => { :action => "create_association" } do |f| %>
+      <%= hidden_field_tag("referer", h(@referer)) unless @referer.nil? %>
+      <%= f.hidden_field :auth_provider %>
+      <%= f.hidden_field :auth_uid %>
+
+      <% if current_user.errors[:email].empty? %>
```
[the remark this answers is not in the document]
milan-cvetkovic: The reason for initially not displaying the email field is to simplify the screen and minimize the possibility of users being tempted to modify the value.

The field when there was no auth provider email available, or there already is an OSM user with the same email address.

app/models/user.rb
```diff
@@ -99,7 +99,7 @@ class User < ApplicationRecord
   validates :display_name, :if => proc { |u| u.display_name_changed? },
                            :characters => { :url_safe => true },
                            :whitespace => { :leading => false, :trailing => false }
-  validates :email, :presence => true, :confirmation => true, :characters => true
+  validates :email, :presence => true, :characters => true
```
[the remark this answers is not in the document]
milan-cvetkovic: It is arguable both ways. But if we add it here, we would have to add it to "new-association" screen as well, at least when user needs to enter the email.

app/controllers/users_controller.rb
```diff
@@ -65,6 +65,48 @@ def new
...
+        render :action => "new"
+      else
+        # Save the user record
+        session[:new_user] = current_user.attributes.slice("email", "display_name", "pass_crypt", "consider_pd")
+        save_new_user
+      end
+    end
+  end
+
+  def new_association
```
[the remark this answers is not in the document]
milan-cvetkovic: One way to do this would be to merge those two actions back into one, but that would make merged erb template quite complex, maybe based on the presence of @auth_uid hidden field.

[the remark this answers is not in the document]
milan-cvetkovic: Ok, so I merged back two actions into one. This results in less repetitions in controller actions and the string resources at the price of somewhat complex `users/new.html.erb`,

config/locales/en.yml
```diff
@@ -2701,20 +2703,36 @@ en:
...
+      tou: "terms of use"
+      contributor_terms_url: "https://wiki.osmfoundation.org/wiki/Licence/Contributor_Terms"
+      contributor_terms: "contributor terms"
+      continue: Sign Up
+      privacy_policy: privacy policy
+      privacy_policy_url: https://wiki.osmfoundation.org/wiki/Privacy_Policy
+      privacy_policy_title: OSMF privacy policy including section on email addresses
+      consider_pd_html: "I consider my contributions to be in the %{consider_pd_link}."
+      consider_pd: "public domain"
+      consider_pd_url: https://wiki.osmfoundation.org/wiki/Licence_and_Legal_FAQ/Why_would_I_want_my_contributions_to_be_public_domain
```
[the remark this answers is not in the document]
milan-cvetkovic: Resolved bu merging back `new-association` and `new` action in users controller.

app/views/users/new_association.erb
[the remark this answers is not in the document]
milan-cvetkovic: @tomhughes This is actually a bit tricky and probably rare scenario. User selects "login with X" and after logging in with X, X reports that user has an email address `user@X.com`. But that address is already used, likely by the same user. The implementation currently un-hides the email field and forces user to choose different email.

This is not really approprirate, since the user may not have another email address to offer. User would already have OSM username/password account. Perhaps the proper response would be to offer them to link their existing OSM account with auth_provider they chose.

Perhaps we could simply redirect them to "Login" page instead of offering them (and forcing them) to provide alternate email?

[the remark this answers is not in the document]
tordans: I like it a lot when a service just connects both accounts and logs me in. One can un-link them later, manually. But the user intend was to login and use the social provider, so doing both and connecting the account follows this intend. I don't thing the app need to request user input on this.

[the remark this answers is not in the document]
milan-cvetkovic: Hm, interesting. I thought that it may be a good idea to verify user's password as it is known to OSM. But maybe it is a valid assumption, at least when we trust 3rd party platform about the email address.

What do we do if we don't trust the email address (github, wikipedia, raw openid)? This would mean that user can create an account on github with someone elses email, then use that to login to OSM. Assuming github does not verify emails, this would allow hijacking OSM account...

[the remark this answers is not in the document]
tomhughes: Well if we don't trust the email address then we do a confirmation on it anyway before the account is activated.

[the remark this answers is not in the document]
milan-cvetkovic: I have updated the PR to modify the existing OSM account with `auth_uid` and `auth_provider` when the email reported by auth provider already has an OSM account. The existing `display_name` is not updated from auth_provider, the existing OSM display name is preserved.

milan-cvetkovic:
Attaching the screenshots for `preferred_provider` signup and login screens.

signup-preferred-provider
login-preferred-provider

simonpoole:
I would suggest removing the "I consider my contributions to be in the public domain" check box, it has always mostly meaningless and has been kept mainly for sentimental reasons. Redoing the login flow is a, just as good as any other, excuse to let it go at last.

simonpoole:
Further note "or sign up with a third party" has always been very very confusing wording. I would suggest "or sign up with a third party account".

tordans:
This looks great! A few quick notes on the screenshots:
- Some field labels have ":" colons, some don't. I think without colons looks cleaner
- The "or" has a line, right? That is nearly invisible in the screenshot. Should it have the same gray as the border of the box?
- Maybe move the "Lost your password" to the "Password"-Label line and align it to the right? It is pretty prominent for a secondary action.
- I think it would be great to have a custom primary navigation header for this flow. Except for the OSM Logo + Home-Link there is nothing there that I would expect a user to need in this case. Especially the sign up/…-buttons are just noise. But the others like "export" are not helpful in this use case either.
- A bit off topic here: I am used to see the ToS- and privacy policy-link as secondary links in a footer area. I just checked and the website does not have them ATM. One has to find them in the text in "[About](https://www.openstreetmap.org/about)". At least in Germany that would not be considered "easily enough accessible" AFAIK. Should we add them to the login page as small link below the box? – I don't want to open a general discussion here … just asking for that page and UI element.

@milan-cvetkovic could you mark which areas can of the sign up page can be customized by the requesting party? I assume it is those part…
> Free and editable
> Unlike other maps, OpenStreetMap is completely created by people like you, and it's free for anyone to fix, update, download and use.

and "Microsoft Map builder" in …
> Signup with OpenStreetMap to access Microsoft Map builder.

tomhughes:
> * A bit off topic here: I am used to see the ToS- and privacy policy-link as secondary links in a footer area. I just checked and the website does not have them ATM. One has to find them in the text in "[About](https://www.openstreetmap.org/about)". At least in Germany that would not be considered "easily enough accessible" AFAIK. Should we add them to the login page as small link below the box? – I don't want to open a general discussion here … just asking for that page and UI element.

This is completely off topic and not something that should be addressed as part of this.

> Free and editable
> Unlike other maps, OpenStreetMap is completely created by people like you, and it's free for anyone to fix, update, download and use.

I really hope that part can't be overridden - why do you think a third party should be able to change that?

milan-cvetkovic:
> This looks great! A few quick notes on the screenshots:

Thanks @tordans,
 
> 
> * Some field labels have ":" colons, some don't. I think without colons looks cleaner

I inherited the labels from the original login and signup pages, they were not consistent to start with... I will add colons to both of them.

> * The "or" has a line, right? That is nearly invisible in the screenshot. Should it have the same gray as the border of the box?

I had the horizontal line as $grey-100, will bump to $grey-300

> * Maybe move the "Lost your password" to the "Password"-Label line and align it to the right? It is pretty prominent for a secondary action.

I am not sure what you have in mind here - moving the link together with the label as "Password (<lost-password-link>):" seems straightforward, but I don't know what part of it would be aligned to the right. Again, I inherited this from the existing Login screen.

> * I think it would be great to have a custom primary navigation header for this flow. Except for the OSM Logo + Home-Link there is nothing there that I would expect a user to need in this case. Especially the sign up/…-buttons are just noise. But the others like "export" are not helpful in this use case either.

This is a regular header on top of the screen, I did not want to modify that at this point. Maybe this could be a part of a separate PR.

> * A bit off topic here: I am used to see the ToS- and privacy policy-link as secondary links in a footer area. I just checked and the website does not have them ATM. One has to find them in the text in "[About](https://www.openstreetmap.org/about)". At least in Germany that would not be considered "easily enough accessible" AFAIK. Should we add them to the login page as small link below the box? – I don't want to open a general discussion here … just asking for that page and UI element.

The regular header is on top of the screen, but because I resized the window when capturing, many menu items hid behind "More..." pull-down, including the "About" button.

> 
> @milan-cvetkovic could you mark which areas can of the sign up page can be customized by the requesting party? I assume it is those part…
> 
> > Free and editable
> > Unlike other maps, OpenStreetMap is completely created by people like you, and it's free for anyone to fix, update, download and use.
> 

This is fixed text, adopted from existing https://www.openstreetmap.org/user/new.

> and "Microsoft Map builder" in …
> 
> > Signup with OpenStreetMap to access Microsoft Map builder.

There are 2 things that can be customized by the 3rd party application authors:
 - "Microsoft Map builder" is the name of the application as registered by the application owner as OAuth2 application.
 - When the application triggers the authorization process, normally it would redirect user to `/www.openstreetmap.org/authorize` and pass a number of parameters required for authoriziation (client_id, resoinse_type,scope,...). If the client app sends and additional `preferred_auth_provider=...` parameter in this request, the resulting screen would bring the "sign-in" buttons to the top of the screen, and highlight the indicated auth-provider. The last screenshots were taken with `preferred_auth_provider=google` in addition to regular auth parameters. Switching between "Signup" and "Login" would preserve the authorization context (although it would not be displayed in the browser navigation bar, as captured in "/users/new" screenshot.

tomhughes:
> I inherited the labels from the original login and signup pages, they were not consistent to start with... I will add colons to both of them.

I think I agree with @tordans that dropping the colons would be better - it's also consistent with the forms on the settings and preferences pages for example.

> I am not sure what you have in mind here - moving the link together with the label as "Password ():" seems straightforward, but I don't know what part of it would be aligned to the right. Again, I inherited this from the existing Login screen.

He means having the label line with `Password` left aligned and `Lost your password?` right aligned - possibly that means making it two columns? Then the text entry is underneath. It sounds reasonable to me.

tordans:
> > > Free and editable
> > > Unlike other maps, OpenStreetMap is completely created by people like you, and it's free for anyone to fix, update, download and use.
> 
> This is fixed text, adopted from existing https://www.openstreetmap.org/user/new.

I see, thanks. On the previous design the text was some kind of additional information (on Desktop). The styling as headline + copy made more sense in this context IMO. To preserve this intention in the new layout, how about making the headlines less prominent?

> **Free and editable.** Unlike other maps, OpenStreetMap is completely created by people like you, and it's free for anyone to fix, update, download and use.

I also thing we should move the 
> Sign up to get started contributing.

to above the box where right now the "Signup with OpenStreetMap to access Microsoft Map builder." text is written. The sentences have the same meaning (in different flows) and are a good way to start into the page.

milan-cvetkovic:
> I also thing we should move the
> 
> > Sign up to get started contributing.
> 
> to above the box where right now the "Signup with OpenStreetMap to access Microsoft Map builder." text is written. The sentences have the same meaning (in different flows) and are a good way to start into the page.

Where would "Sign up to get started contributing" in case of regular signup screen (2nd screenshot at the top of PR)?

tordans:
> Where would "Sign up to get started contributing" in case of regular signup screen (2nd screenshot at the top of PR) go?

My thinking was to use the same place in both flows right above the box. In terms of content both have the same role to play but in a different context so it makes sense to me to use the same place in the UI.

milan-cvetkovic:
- Updated to include #4488 and #4489
- Re-arranged password field in login screen based on @tordans suggestions
- Moved "Sign up to get started contributing" to the top of "Sign in" screen

signup-regular
signup-preferred
login-preferred

milan-cvetkovic:
Merged with master to include #4492.

tordans:
@milan-cvetkovic I think that looks great. 

Is there a way to test it on one of the dev servers?

Two minor things I notices:


| - |
|--------|
| image | 
I was initially thinking to overwrite the first line with the second line in this specific flow. The second tells the story well enough on it's own. But having both works as well…
   

| - |
|--------|
| image | 
I wonder if we can get rid of the special font size on the intro paragraph and make it the same size as the "by signing up…" part below. Having viewer font sizes on one screen is generally a good thing. This might make the 2-line-paragraph take 3 lines but that is OK, I think.

tomhughes:
I've put up a test site at https://signup.apis.dev.openstreetmap.org/ for people to try.

It doesn't have any third party logins configured though which limits things a bit, but they wouldn't work without registering a whole bunch of apps with different callbacks on all the services.

milan-cvetkovic:
> I was initially thinking to overwrite the first line with the second line in this specific flow. The second tells the story well enough on it's own. But having both works as well…

This is probably possible (and simple to implement), but @tomhughes asked to keep that paragraph on the screen in one of the earlier comments. I guess we can keep them both for now.

milan-cvetkovic:
> One thing I noticed using the live site is that it's a bit odd the way the tab panel moves up and down when you switch between login and signup - maybe we should move the text above the panel inside it on the signup page? or just leave blank space there on the login screen so they line up?

Good catch. I think bringing it back to "signup" form would look better, since the portion above the signup/login tabs would stay unmodified when you ckick the tabs.

milan-cvetkovic:
@tordans 
> I wonder if we can get rid of the special font size on the intro paragraph "Free and editable…" and make it the same size as the "by signing up…" part below. Having fewer font sizes on one screen is generally a good thing. This might make the 2-line-paragraph take 3 lines but that is OK, I think.

Updated with same font size (16px).

Also, moved "Sign up to get started contributing" to inside signup form.

tordans:
Some testing on https://signup.apis.dev.openstreetmap.org/login

* It needs some space at image (it does have the space on the sign up view)
* FYI, this is the wording on the German UI ATM. Some text are longer, which is OK, just good to see that it is working image
* The new forgot-passwort-field looks non-ideal on mobile image (that can be fixed later, of course…) —changing from `col-sm` to just `col` solved it in the inspector: image



**Column spacing on mobile:**

I think the padding-x inside the box is a bit too wide on mobile (tested with iPhone 12 Pro Size in Chrome Devtools). I don't know a lot about Bootstrap anymore, but adding `px-4  px-sm-5` to the `col-sm` helps. It does make the sm+ case a bit bigger, but that is OK IMO. Someone who knows more about bootstrap might know how to overwrite just the mobile case…

| before | after |
|--------|--------|
| mobile-before | mobile-after |
| Bildschirmfoto 2024-01-27 um 07 48 40 | desktop-after |


---

Just double checking: We cannot trigger the UI that is used for the app-permissions-flow on that server, right? (Just showing it, I don't want to submit…)

milan-cvetkovic:
@tordans Thanks for the review.

Updated spacing on login screen above horizontal line.

> * The new forgot-passwort-field looks non-ideal on mobile

Fixed based on your advice.

> **Column spacing on mobile:**

Same here. As you noted, it results with somewhat more padding in regular screen (48px vs 40px)

milan-cvetkovic:
@tordans 
> Just double checking: We cannot trigger the UI that is used for the app-permissions-flow on that server, right? (Just showing it, I don't want to submit…)

You can see the final screen by passing `auth_uid`, `email` and `verified_mail` parameters:

[when auth_provider returns non-verified email](https://signup.apis.dev.openstreetmap.org/user/new?auth_provider=blah&auth_uid=0000&email=someone@somewhere.com&nickname=newusertest)

[when auth_provider returns no email](https://signup.apis.dev.openstreetmap.org/user/new?auth_provider=blah&auth_uid=0000&nickname=newusertest)

You can use other apps to trigger authorization process, for example JOSM allows you to specify URL address when creating Oauth2 token. This would only allow you to see the login and signup screens in this scenario.

milan-cvetkovic:
@tomhughes, are there any more modifications required for this PR?

app/assets/stylesheets/common.scss
```diff
@@ -1086,4 +1086,102 @@ div.secondary-actions {
   }
 }
 
+.auth-container {
```
gravitystorm, line 1089: I haven't had the chance for a full review, but these stylesheet changes stand out to me. Our general approach is to use the bootstrap framework, not custom CSS. In particular, custom max-widths, custom font sizes, custom border radii etc aren't the way we should be implementing changes.

Instead, bootstrap classes should be added to html elements, e.g. `rounded-2` should be used. These can of course be used sparingly, since html elements like `<small>` should do the right thing too.

I can see an implementation of tabs here too, and I would expect instead to see bootstrap tabs being used, as elsewhere on the site.

tomhughes:
I'll try and have a proper look again at the weekend but one thing I did just notice is that the login screen still has a trailing colon on the username label, and the (normally hidden) field for OpenID has the same thing on both screens.

app/views/users/new.html.erb
```diff
@@ -1,52 +1,100 @@
 <% content_for :head do %>
   <%= javascript_include_tag "user" %>
+  <%= javascript_include_tag "auth_providers" %>
 <% end %>
 
-<% content_for :heading_class, "pb-0" %>
-<% content_for :heading do %>
-  <div class='header-illustration new-user-main'>
```
gravitystorm, line 7: I'm surprised to see the header-illustration being removed from the signup pages, since it's one of the few pieces of genuinely original design in our project! But it's more surprising to see it removed from just this page, and not from the other pages that share the same design.

What's the design rationale, if any, behind removing it?

We don't have much in-house capacity for this kind of design, so it concerns me that whenever external companies come in, they often focus on these few pages (login signup etc) and keep redesigning over the top of whoever made the last external design changes. This leads to no redesign being anywhere near comprehensive, and each redesign not quite covering the same (small) ground as the previous one, and we end up with widespread inconsistencies.

gravitystorm (changes requested):
I've added a couple of design comments that stand out to me, but there's plenty more here for me to review later.

app/controllers/users_controller.rb
```diff
@@ -290,8 +241,23 @@ def auth_success
         else
           failed_login t("sessions.new.auth failure")
         end
+      elsif user.nil? && user = User.find_by(:email => email)
```
tomhughes, line 244: I don't think we need `user.nil?` here as we would have taken the previous branch if user was set.

app/controllers/users_controller.rb
```diff
@@ -290,8 +241,23 @@ def auth_success
...
+
+        user.deactivate! if user.status == "active" && !email_verified
+
+        if user.status == "active"
+          successful_login(user)
+        else
+          session[:token] = user.tokens.create.token
+          UserMailer.signup_confirm(user, user.tokens.create(:referer => session[:referer])).deliver_later
+          redirect_to :controller => :confirmations, :action => :confirm, :display_name => user.display_name
+        end
```
tomhughes, line 257: This is trying to deal with the case where somebody does a social signup and there is already an account with the email address the social provider returns and it looks problematic to me for a number of reasons:

* If the user already had a different social provider attached to their account this will overwrite it
* If the social provider satisfies our (deliberately weak) definition of providing a verified email then we allow them to attach to the existing account with no further verification but that is very different to allowing creation a new account based on a weak verification
* If the social provider email is not considered verified then we send a normal signup confirmation but the working of that will be a bit odd in this case I think plus we actual deactivate the account until that confirmation happens

There's a definite problem here, that I had a case of only this week, where somebody tries a social signup and they already have an account but I think we need to send a confirmation email in all cases and probably make it a different one saying that we need to confirm the social signup link and just use it to confirm the link rather than doing a deactivate first.

I'm not sure what to do where there is already a different provider attached though though arguably that's fine if we only change the auth provider after email confirmation and the email says we will be replacing the old one.

app/models/user.rb
```diff
@@ -181,11 +181,9 @@ def self.authenticate(options)
       transitions :from => :pending, :to => :active
     end
 
-    # Used in test suite, not something that we would normally need to do.
-    if Rails.env.test?
```
tomhughes, line 185: This is only being removed so that we can deactivate while revalidating an auth id association and as explained above I don't think we can do that so this can then be left as is.

tomhughes (changes requested):
Mostly looking good from a backend point of view with just a few minor issues but there is one big question around the new logic you've put in for dealing with linking a social signup to an existing account.

Also there's the inconsistent colons on the field labels that already mentioned and @gravitystorm has made some high level design comments in his review and will doubtless have more to say on the UI side before this is ready.

app/controllers/users_controller.rb
milan-cvetkovic: I agree with you that this is indeed a tricky scenario. I am not really a fan of hijacking the existing account once you signin successfully with social account with same email address.

> There's a definite problem here, that I had a case of only this week, where somebody tries a social signup and they already have an account but I think we need to send a confirmation email in all cases and probably make it a different one saying that we need to confirm the social signup link and just use it to confirm the link rather than doing a deactivate first.

I suspect user already had an OSM account and tried logging in with social account, with desire to add an alternate method of signing in. The functionality to do this already exists, by modifying in the "External Authentication" in settings.

> There's a definite problem here, that I had a case of only this week, where somebody tries a social signup and they already have an account but I think we need to send a confirmation email in all cases and probably make it a different one saying that we need to confirm the social signup link and just use it to confirm the link rather than doing a deactivate first.

This seems like a complex approach to me, since the confirmation email process currently practically only activates the account.

I suggest we step back and simply detect this scenario, and suggest the user to actually login with their email account/password, and modify their settings to include social sign in? This would also eliminate the scenarios where the user already uses a different social signup.

milan-cvetkovic:
> I'll try and have a proper look again at the weekend but one thing I did just notice is that the login screen still has a trailing colon on the username label, and the (normally hidden) field for OpenID has the same thing on both screens.

@tomhughes I fixed the OpenID label, but couldn't find the colon on username...

app/views/users/new.html.erb
milan-cvetkovic: Rationale for the removal was to make login and signup screens consistently looking and to spare some real estate on the screen. I guess we had not realized the importance of the art work.

Nevertheless, we asked our designer to update the design to incorporate the globe image and tabs similar to tabs found in GPX Traces in OSM.

app/assets/stylesheets/common.scss
milan-cvetkovic: I updated the PR to use minimal set of custom stylesheets, as suggested.

The only remaining pieces are the total width of the screen in signup and login screens and customization of bootstrap tabs for line style. Granted, the tab style is not same as currently used in OSM in GPX traces, it wouldn't work without darker background.

tomhughes:
> @tomhughes I fixed the OpenID label, but couldn't find the colon on username...

Don't worry about it, it's in the `en-GB` translatation not the main `en` file :-(

app/controllers/users_controller.rb
tomhughes: Yes that is certainly the simplest solution, and removes the problems with the current code.

milan-cvetkovic:
@gravitystorm Our designer adjusted login and signup screens to include globe image and OSM style tabs. Available in the same figma as original design mentioned in [Issue #4128](https://github.com/openstreetmap/openstreetmap-website/issues/4128#issuecomment-1675045559): https://www.figma.com/file/RqpiQmshkhjTew2Y1MvDn9/OSM-SSO?type=design&node-id=3%3A3490&mode=design&t=Dk7hr35VQWraaysh-1

under the heading "SSO Proposal V2"

app/controllers/users_controller.rb
milan-cvetkovic: I have reverted the "hijacking of the OSM account by social sign-in" feature. Instead, I only detect the condition and display a message as the flash notice. Currently the text says:

> If you already have OpenStreetMap account and wish to use 3rd party identity provider, please login using your password and modify settings of your account.

app/views/users/new.html.erb
milan-cvetkovic: Implemented alternative design with globe image on top of signup/login screens. 
Updated to use bootstrap tabs similar to the existing tabs in OSM website.

milan-cvetkovic:
@tomhughes test errors on Ubuntu 20.04 seem to be spurious, can the build be restarted?

app/views/sessions/new.html.erb
```diff
@@ -1,51 +1,71 @@
...
-<div id="login_login">
-  <p class='text-muted'><%= t ".no account" %> <%= link_to t(".register now"), user_new_path(:referer => params[:referer]) %></p>
+<% content_for :heading do %>
+  <% if @client_app_name %>
+    <p class="text-center text-muted fs-6 py-2 mb-0 bg-white"><%= t(".login_to_authorize_html", :client_app_name => @client_app_name) %></p>
+  <% end %>
 
-  <%= bootstrap_form_tag(:action => "login", :html => { :id => "login_form" }) do |f| %>
-    <%= hidden_field_tag("referer", h(params[:referer]), :autocomplete => "off") %>
+  <div class="row justify-content-center">
```
AntonKhorev, line 13: I'm getting a horizontal scrollbar on narrow screens here without `mx-0`.

app/views/users/new.html.erb
```diff
@@ -1,52 +1,113 @@
...
+<% content_for :heading_class, "p-0 mw-100" %>
+
 <% content_for :heading do %>
-  <div class='header-illustration new-user-main'>
-    <h1><%= t ".title" %></h1>
+  <% if @client_app_name %>
+    <p class="text-center text-muted fs-6 py-2 mb-0 bg-white"><%= t(".signup_to_authorize_html", :client_app_name => @client_app_name) %></p>
+  <% end %>
+
+  <div class="row justify-content-center">
```
AntonKhorev, line 13: Same horizontal scrollbar as [above](https://github.com/openstreetmap/openstreetmap-website/pull/4455/files#r1509949490).

app/assets/stylesheets/common.scss
```diff
@@ -1078,4 +1083,12 @@ div.secondary-actions {
...
+  max-width: 660px;
+}
+
+.nav-z-100 {
+  z-index: 100;
+}
```
AntonKhorev, line 1092: Is it necessary? It's used on a tab list inside the container with the globe image, but the image is background. There's nothing else that might cover the tabs.

app/views/users/new.html.erb
```diff
@@ -1,52 +1,113 @@
...
+                                                                 :target => :new),
+                                            :privacy_policy_link => link_to(t(".privacy_policy"),
+                                                                            t(".privacy_policy_url"),
+                                                                            :title => t(".privacy_policy_title"),
+                                                                            :target => :new),
+                                            :contributor_terms_link => link_to(t(".contributor_terms"),
+                                                                               t(".contributor_terms_url"),
+                                                                               :target => :new)) %></p>
+      <%= f.form_group do %>
+        <%= f.check_box :consider_pd,
```
AntonKhorev, line 90: I suspect that when this checkbox is placed here, people will tick it without reading what is says thinking that it's necessary to go forward.

app/views/sessions/new.html.erb
milan-cvetkovic: I am not able to reproduce this.

Can you post a screenshot and/or details on browser/os involved?

Are you suggesting that I should add "mx-0" to the styles of `<div class="row justify-content-center">` element?

AntonKhorev: Actually it's on any screen size.

image

image

Chromium Version 122.0.6261.94 (Official Build) snap (64-bit)
and
Firefox 123.0 on Ubuntu

AntonKhorev: Maybe replacing `row` with `d-flex` is better. With row you have negative margins, then padding inside.

AntonKhorev: Or maybe even

```
<div class="header-illustration new-user-main auth-container mx-auto">
```

instead of

```
<div class="row justify-content-center">
    <div class="header-illustration new-user-main auth-container">
```

milan-cvetkovic: Found it, actually horizontal scroll bar was always present. mx-0 fixes this

app/assets/stylesheets/common.scss
milan-cvetkovic: Not strictly necessary, but it depends on the length of text.

If the labels "Signup" and "Login" are long texts, the globe's arm would come behind the text, making the text of unselected tab somewhat obfuscated by the image behind it here is an example:

image.

AntonKhorev: Doesn't that happen with `.nav-z-100` too?

milan-cvetkovic: Actually, it looks like z-index does not make any difference here. Thanks!

simonpoole:
I haven't seen this issue mentioned yet, if it has, apologies. Testing on mobile or with a reduced width browser on the test instance I get if I re-click the  menu button. 



new_signup

To re-create: start on the map screen, then click the menu button, click Log In or Sign In, then click the menu button again,. Not a big drama, just slightly weird looking.

app/views/users/new.html.erb
AntonKhorev: > You might be referring to TOU and privacy policy

That's what users will think this checkbox is about, but it's not.

AntonKhorev:
> To re-create: start on the map screen, then click the menu button, click Log In or Sign In, then click the menu button again,. Not a big drama, just slightly weird looking.

Is it a suggestion to remove Login / Sign up buttons from the menu when on a Login / Sign up page?

app/views/sessions/new.html.erb
```diff
@@ -1,51 +1,69 @@
...
-  <p class='text-muted'><%= t ".no account" %> <%= link_to t(".register now"), user_new_path(:referer => params[:referer]) %></p>
+<% content_for :heading do %>
+  <% if @client_app_name %>
+    <p class="text-center text-muted fs-6 py-2 mb-0 bg-white"><%= t(".login_to_authorize_html", :client_app_name => @client_app_name) %></p>
+  <% end %>
 
-  <%= bootstrap_form_tag(:action => "login", :html => { :id => "login_form" }) do |f| %>
-    <%= hidden_field_tag("referer", h(params[:referer]), :autocomplete => "off") %>
+  <div class="header-illustration new-user-main auth-container mx-auto">
+    <ul class="nav nav-tabs position-absolute bottom-0 px-5 fs-6">
```
AntonKhorev, line 14: You can add `w-100` here to extend the border below the tabs.

simonpoole:
> Is it a suggestion to remove Login / Sign up buttons from the menu when on a Login / Sign up page?

It is already hidden on the initial display yes, IMHO that is probably best. We can't get completely rid of it, unless somebody comes up with a better way to be able to navigate to the login/signup page.

config/locales/en.yml
```diff
@@ -1849,43 +1848,17 @@ en:
   sessions:
     new:
       title: "Login"
```
simonpoole, line 1850: Login is inconsistently spelled "Log in"  or "Login" throughout this PR.

milan-cvetkovic: English is not my 1st language, so I need some guidance here: it is not clear to me when "log in" or "login" should be used.

OSM website was not consistent to start with: Here are the counts as reported by `grep -i "log in" config/locales/en.yml` and `grep -i "login" config/locales/en.yml` on master and this PR.

```
|        | master | this PR |
+--------+--------+---------+
| log in |    7   |    8    |
| login  |   29   |   30    |

Should I go ahead and replace all instances of "log in" with "login"?
```

tomhughes: Well wiktionary suggests that it is "login" when used as a noun, and "log in" when used as a verb. In fact it lists "login" as a verb as a mis-spelling of "log in" and refers to http://notaverb.com/login on the subject...

simonpoole: > English is not my 1st language

I don't think that would actually help as we are very very inconsistent in general (outside of OSM), there are at least login, log in, log-in, and log-on in use (not to mention that sign-in would work to, but that might get confused with sign-up). IMHO @gravitystorm  and @tomhughes should simply decide on one variant.

AntonKhorev: One of them is a verb, another is a noun. For buttons you'd use the "Log in" verb, but looks like we currently don't.

milan-cvetkovic: > One of them is a verb, another is a noun. For buttons you'd use the "Log in" verb, but looks like we currently don't.

Thanks, I will review my modifications having this suggestion in mind.

app/views/sessions/new.html.erb
milan-cvetkovic: Thanks, looks better, although I hadn't noticed it before.

milan-cvetkovic:
> > To re-create: start on the map screen, then click the menu button, click Log In or Sign In, then click the menu button again,. Not a big drama, just slightly weird looking.
> 
> Is it a suggestion to remove Login / Sign up buttons from the menu when on a Login / Sign up page?



> > Is it a suggestion to remove Login / Sign up buttons from the menu when on a Login / Sign up page?
> 
> It is already hidden on the initial display yes, IMHO that is probably best. We can't get completely rid of it, unless somebody comes up with a better way to be able to navigate to the login/signup page.

I removed Login/signup from top header when users or sessions controller is active.

simonpoole:
Further minor issue on mobile, margins (left and right) are too large, wit:

milan-cvetkovic:
> Further minor issue on mobile, margins (left and right) are too large/inconsistent, wit:
> 

fixed, thx

milan-cvetkovic:
@tomhughes @gravitystorm any more changes required for this PR?

app/assets/stylesheets/common.scss
```diff
@@ -750,6 +750,11 @@ tr.turn:hover {
 
   &.new-user-main {
     background-image: image-url("sign-up-illustration.png");
+    background-position-x: 100px;
```
tomhughes, line 753: I would probably make this more like 40-50 pixels - it stops the image being cropped on the right and also means more of it appears with a smallish phone screen.

app/assets/stylesheets/common.scss
```diff
@@ -750,6 +750,11 @@ tr.turn:hover {
...
+  }
+
+  &.new-user-blocked {
+    background-image: image-url("sign-up-illustration.png");
```
tomhughes, line 757: I started off asking if this also needed the position attribute but found that led me into the fact that the whole `users/blocked` view probably needs to be updated to match the new login/signup layout.

I've added an ACL to block the `blocked.example.com` domain on the test site so you can see what I mean by trying to signup with a `user@blocked.example.com` email.

tomhughes (commented):
I'll try and have a fuller look later in the week but this is just a couple of things I noticed.

app/assets/stylesheets/common.scss
milan-cvetkovic: I will reduce it to 50.

It is a double edged sword though, since if the tab labels are long words, the "Sign up" link colour would be difficult to read with blue hand of the globe.

app/assets/stylesheets/common.scss
milan-cvetkovic: I can definitely remove the arm from this page.

But other than that, the page does not look as finished.

tomhughes: The overlapping arm is actually an existing bug, as is the bizarrely massive font size. What I think is new is the relative width of that screen compared to the login screen which in turn means the header image is differently positioned.

gravitystorm:
> @tomhughes @gravitystorm any more changes required for this PR?

If you can find any way to move parts of this PR into separate PRs, then that would be great. For example, you fixed the blocked page (headings and paragraphs) in the latest commit, but that could be a completely separate PR that can be reviewed separately and merged first. Are there other things like that here?

Perhaps the "login" vs "log in" text changes? The logo change? Or other similar things that can be treated separately? Large pull requests are quadratically hard to deal with - not only is there more to review, but each time I come to re-review the PR, more things have changed, and I need to read and re-read both the changed and also the unchanged parts over and over. So it's really worthwhile to break down the PR into smaller parts, that will ideally only need one maintainer to review and merge.

app/views/layouts/_header.html.erb
```diff
@@ -104,7 +104,7 @@
           <%= link_to t("layouts.logout"), logout_path(:referer => request.fullpath), :method => "post", :class => "geolink dropdown-item" %>
         </div>
       </div>
-    <% else %>
+    <% elsif controller_name != "users" and controller_name != "sessions" %>
```
gravitystorm, line 107: This means that if you go to users#show (e.g. visit www.osm.org/user/foo while not logged in) then there is no button to log in or sign up.

If this problem only affects the login and signup pages it needs to be more targetted, either by specific controller methods or moving the signup code into a different controller (e.g. RegistrationsController, to align with Devise). 

Alternatively, perhaps there is a better design solution, since removing these buttons on the full-width layout looks weird and slightly inconsistent to me.

milan-cvetkovic: Good catch, thanks.

I will modify the condition to target only `new` action in sessions_controller and users_controller. Hopefully that covers it. Otherwise, I would revert the modifications to _hedaer.html.erb,  and address the "noisy" Login/Signup buttons from the header in a separate PR.

milan-cvetkovic:
> > @tomhughes @gravitystorm any more changes required for this PR?
> 
> If you can find any way to move parts of this PR into separate PRs, then that would be great. For example, you fixed the blocked page (headings and paragraphs) in the latest commit, but that could be a completely separate PR that can be reviewed separately and merged first. Are there other things like that here?
> 
> Perhaps the "login" vs "log in" text changes? The logo change? Or other similar things that can be treated separately? Large pull requests are quadratically hard to deal with - not only is there more to review, but each time I come to re-review the PR, more things have changed, and I need to read and re-read both the changed and also the unchanged parts over and over. So it's really worthwhile to break down the PR into smaller parts, that will ideally only need one maintainer to review and merge.

I have created 3 separate PRs as you suggested:
- #4455
- #4578
- #4580

I understand the frustration with large PRs like this. They are not pleasant neither for reviewers nor the contributor.

The moving target for contributor (master branch being updated resulting in merge conflicts) triggers me to rebase the PR branch, breaking the correlation between comments and code, and that makes reviewing the PR even more unpleasant, usually resulting in full re-review.

In case of this PR there is a really big modification that cannot be reasonably split into smaller PRs, namely "joining sign-up and terms screens into one" which touches large portions of `users_controller.rb` and related views.

However, it seems that the small PRs I created above will not reduce the size of this PR significantly, since this PR would modify or move the same pieces of code. I will have to wait until they are merged before rebasing this PR once again...

stalker314314:
Hi @tomhughes , @gravitystorm ,
We are close to 3 months on this PR. I understand that PR is huge and I understand that this is just a small bit of work that you are doing, but please try to expedite this, as @milan-cvetkovic basically did everything he was asked to. Yesterday Milan even rebased this PR and split it into 9 logical commits that should also help with reviewing. This work is part of our ongoing effort to enable SSO in Map builder editor and this is starting to be bottleneck for our work. Please try to find time to review this (and I really hope there will be no blocking comments as this PR is now 2.5 months out:)

tomhughes:
I realise that we may not be able to progress at things at the rate you are used to in a corporate environment but this is certainly something we want to see and have been keen to progress and we have done a number of reviews and provided regular feedback along the way.

I will have another look in due course but as far as I am aware the basic technology aspects of this are all fine now and I'm happy with the UI appearance and the only thing I was really waiting on was for @gravitystorm to be happy with the UI implementation at which point I would have been to do a final review and in all probability go ahead and merge it.

app/views/layouts/_header.html.erb
AntonKhorev: I'd show/hide the buttons only in small layout:
https://github.com/openstreetmap/openstreetmap-website/commit/09877a78e901698e10e8172ed868395653767ae6

milan-cvetkovic: @AntonKhorev Interesting. It looks good to me, if we desire to keep the header buttons and tabs at the same time.

@gravitystorm
> Alternatively, perhaps there is a better design solution, since removing these buttons on the full-width layout looks weird and slightly inconsistent to me.

There is a small inconsistency if we keep Login/Signup on the pages along with the tabs under the globe: The login buttons have different referrer:
- login button in adds the referrer to the current page
- login button in the tab keeps the referrer as originally set by initial "login" button that invoked it.

milan-cvetkovic:
@gravitystorm Not trying to be pushy, but do you have an idea when you could give another look?

gravitystorm:
> @gravitystorm Not trying to be pushy, but do you have an idea when you could give another look?

For the last few weeks (and the next two) I've been particularly busy with some private matters. I have this on my list of things to review, but I'm not yet sure if I will have time in the next 2 weeks to look at this, particularly since it's a large PR it's harder for me to review when I have only short periods of availability at a time.

There's also (at the time of writing) 125 other pull requests, and I'm sure everyone else wants to be at the front of the queue too!

app/controllers/users_controller.rb
```diff
@@ -101,14 +107,9 @@ def create
       if current_user.invalid?
         # Something is wrong with a new user, so rerender the form
         render :action => "new"
-      elsif current_user.auth_provider.present?
-        # Verify external authenticator before moving on
-        session[:new_user] = current_user.slice("email", "display_name", "pass_crypt", "pass_crypt_confirmation")
```
AntonKhorev, line 106: `session[:new_user]` filtered out other parameters such as `auth_provider` which was added later in `auth_success`. A new users was created from `session[:new_user]`.

With `session[:new_user]` gone its possible to create users with fake values for `auth_provider` by opening the signup form like this: `/user/new?auth_provider=test1&auth_uid=test1`

app/controllers/users_controller.rb
```diff
@@ -290,7 +243,8 @@ def auth_success
           failed_login t("sessions.new.auth failure")
         end
       else
-        redirect_to :action => "new", :nickname => name, :email => email,
+        verified_email = UsersController.message_hmac(email) if email_verified && email
```
AntonKhorev, line 246: `verified_email` is no longer a email address and it's named `email_hmac` elsewhere. Should it be renamed here too?

app/controllers/users_controller.rb
milan-cvetkovic: I believe that user would not be created immediately after `/user/new`.

They would still have to click on "Sign up" on the screen that offers them to confirm their display name. From here, they would have to verify their email as well, unless they also correctly guess the value of `verified_email` value. Finally, they would probably have to reset their password as well, having no other way to login, and password being set up to a random value.

If this is a real concern, I can bring back the extra round trip to the external provider before creating user record.

milan-cvetkovic: > > They would still have to click on "Sign up"
> 
> Yes, but right after that the user is written to the db with arbitrary auth_provider. That happens before any further confirmation.

Correct. This is pretty much same as if user is created by regular "signup" screen, before they confirm the email address. The only difference is that there is arbitrary, not usable, value of `auth_provider`, and the user does not know the password assigned to them.

I removed the extra round trip after @tomhuges suggestion here: https://github.com/openstreetmap/openstreetmap-website/pull/4455#discussion_r1443987257.

Having a user with bogus value of `auth_provider` does not hurt OSM web site. While the user cannot use the bogus value that they manually entered, they can use the account by username/password after resetting the password. Since there is a manual action required to actually create the user record, there is no danger of bulk creation of user accounts this way, more than it is possible today.

milan-cvetkovic: @tomhughes I am planning to re-introduce the verification round trip, due to security concerns raised by @AntonKhorev

milan-cvetkovic:
@AntonKhorev

> > I would suggest removing the "I consider my contributions to be in the public domain" check box, it has always been mostly meaningless and has been kept mainly for sentimental reasons. Redoing the login flow is a, just as good as any other, excuse to let it go at last.
> 
> That's probably the best idea, because its [placement is confusing](https://github.com/openstreetmap/openstreetmap-website/pull/4455#discussion_r1509952021). When I was testing this PR I ticked it without noticing at least once.
> 
> If we still need to let users to declare their edits PD, we can move the checkbox to their account settings.

Let's do the removal of the check box in a separate PR

app/controllers/users_controller.rb
tomhughes: What "security concern" is that? Why do we care if somebody creates a user with a "fake" authentication provider? What exactly is the claimed attack vector here?

tomhughes: That user would never be able to login with the provider in question because it's not a provider we know anything about and if we do decide we care we can just validate that parameter against the expected values - no need to do a revalidation that I can see.

milan-cvetkovic: @tomhughes - sorry, I didn't see your comment before pushing the extra round-trip. It is a separate commit so it would be easy to remove.

I don't really see the attack vector here, other than users creating accounts for which they would have to reset the password.

milan-cvetkovic: @tomhughes , @AntonKhorev Can we reach consensus here?

I don't mind it either way, but I prefer the no-verification round trip purely due to being more responsive to user.

AntonKhorev: User A gives user B a registration url with planted auth provider.

tomhughes: Then what? What happens to a user with an auth provider of `fred` recorded in the database that causes a problem? Other than that they can't login until they do a password reset if they didn't chose a password when creating the account.

tomhughes (approved):

tomhughes (commented):
I've done a new review, including pushing a minor fix for a grammar issue, and just have a couple of comments, both about the preferred provider stuff.

The first is with the highlighting of the preferred authentication provider:

image

as you can see in that image the text alongside the preferred provider is not well aligned vertically relative to the icon which in turn means the icon is visibly closer to the bottom border than the top one.

The second is the positioning of the client specific prompt to action above the main header bar:

image

I don't think it's good to put it above the header like that - nothing else on the site does that or causes the header to move down like that.

I would suggest moving it and having it replace the generic "Sign up to get started contributing." text with the client specific call to action. That is obviously a less prominent position (though putting it above the header actually also makes it hard to notice or did for me) so maybe it should be bolded or emphasised in some other way.

milan-cvetkovic:
@tomhughes

> as you can see in that image the text alongside the preferred provider is not well aligned vertically relative to the icon which in turn means the icon is visibly closer to the bottom border than the top one.

This should be fixed now, thanks.

> I don't think it's good to put it above the header like that - nothing else on the site does that or causes the header to move down like that.

The [design document](https://github.com/openstreetmap/openstreetmap-website/issues/4128#issuecomment-1675045559)  actually has a light grey divider line above the div with the text "Sign up with OpenStreetMap to access..." text. I missed that, so it all became white.

Here is the reasoning why the design is with the text on top:
- the instructions to sign up or log in are presented before the choice of actions (Sign Up and Log In tabs)
- the user comes to this page while trying to do an action on a 3rd party app - it might be surprising "why am I here?" The first thing they should see is the explanation- which "Sign up with OSM to continue with your app...." or "Login to OSM to continue with your app..." is trying to do
- The text stripe is above the globe image because the tabs are embedded into the Globe image.

> I would suggest moving it and having it replace the generic "Sign up to get started contributing." text with the client specific call to action. That is obviously a less prominent position (though putting it above the header actually also makes it hard to notice or did for me) so maybe it should be bolded or emphasised in some other way.

I also have the feeling that the text is almost unnoticable, but I am not sure where to move it. If we keep it where on top of the globe image, we could potentially make it more noticable, perhaps by different background colour. I doubt that an additional grey line as originally suggested would do the trick.

Here are the original design and proposed design with more prominent background, similar to how `flash` notice looks (both designs have incomplete top navigation bar): 

image

Proposed design: 
image

tomhughes:
Well the design document isn't really comparable because it doesn't have the header bar at all...

What I can say is that when I tested this I was expecting to see that message and initially I couldn't find it because I wasn't looking above the header so if somebody that's actually seeking it can't see it is a user going to notice it? Then again maybe somebody less familiar with the site design would be more inclined to notice it there.

tomhughes:
A new idea... How about if we put this message as a flash notice? That would move it immediately below the header but place it on a green background. It would also be consistent with how other informational notices are presented on the site.

milan-cvetkovic:
> Well the design document isn't really comparable because it doesn't have the header bar at all...

It actually does have the header design - in the left menu there is a list of "Pages" with "SSO v2" with globe image header. Here is the direct link to the second page: https://www.figma.com/file/RqpiQmshkhjTew2Y1MvDn9/OSM-SSO?type=design&node-id=614-151&mode=design.

> A new idea... How about if we put this message as a flash notice? That would move it immediately below the header but place it on a green background. It would also be consistent with how other informational notices are presented on the site.

I replaced the top bar with flash notice, but it shows up above the header, and it is left-aligned.

image

I am not sure if placing the message bellow the header which includes the tabs would look good: here is an image from dev env:
image

tomhughes:
Yes that definitely isn't going to work... Not one of my better ideas. I've backed out that change but added a small fix for the flash messages when signing up with a social account which I noticed when looking at your change.

I'm still not really keen on having the message above the heading bar but I'm going to go ahead and merge it and we can always improve things later if @gravitystorm has better ideas or other things he's like changing once he has time to look at things again.

tomhughes:
That is merged and live now. Thanks for all your work on this and sorry it's taken so long!

gravitystorm:
I've created an issue with some UI-focussed review notes at #4773 .

I just want to also touch on reviewing this PR in general. I think in future I would push back harder on doing so many things at once, all in the same PR. I wasn't joking back in August 2023 when I said

> Great! Definitely more than one PR involved here 😄

and in March:

> If you can find any way to move parts of this PR into separate PRs, then that would be great

I appreciate that some small parts were separated off, but there was still a large amount of both functionality changes happening at the same time as significant user interface redesign going on. There was no need, for example, to have the `auth_button_preferred` helper in the same PR as removing the contributor terms, in the same PR as fixing the 3rd-party signup workflow, in the same PR as discussing auth round trip security, etc etc. 

In particular, although there were several conversations about how hard this was making it for us to review the PR, the conversation then moved on to pressuring Tom and I to review and merge the PR regardless [[1]](https://github.com/openstreetmap/openstreetmap-website/pull/4455#issuecomment-2007142676), [[2]](https://github.com/openstreetmap/openstreetmap-website/pull/4455#issuecomment-2027017070). I have to be honest that it left a bad taste in my mouth, and I'm keen to avoid that happening again in future.

So please, smaller, more focussed PRs next time! :smile: It will make things easier for everyone.

PR #4459, opened 2024-01-05
Add :closed trait to note factory

AntonKhorev:
Currently you create a closed note in tests like this:

```ruby
create(:note, :status => "closed", :closed_at => Time.now.utc)
```

You have to remember to set `:closed_at`. There's code that assumes that closed notes have this time set. If omitted, it's going to be unset.

Here I add the `:closed` trait to the note factory that will set `:closed_at` and also add a closing comment.


tomhughes:
Looks good, thanks.

Issue #2581, opened 2020-04-05
Add language dropdown menu


AntMadeira:
Hi there. 
Recently I realized there is no way to know which languages are available on the Profile settings, which can cause problems to new users and newbies (as it happened to me before).
Would it be possible to add a dropdown menu with all the languages to facilitate this task?

Regards.

HolgerJeromin:
Thank for filing the issue.
This was discussed already in #1115

kjonosm:
I read this issue being different from #1115 as it seems to not ask for a language switcher on the homepage also for non-logged in users but (only) for a dropdown menu within the preferences panel instead of a free text field to help choosing the preferred language. Just like it's already implemented for choosing the preferred editor.

<img width="964" alt="Bildschirmfoto 2022-09-07 um 10 12 31" src="https://user-images.githubusercontent.com/51236777/188829207-63e7ae7f-5419-43d6-88e6-324bafb3d68f.png">

I support this request because the chosen language is not only used for the website but for the iD editor, as well.

tomhughes:
But that field is a simple selection, it's a list of preferred languages, so it needs something much more sophisticated than a drop down.

AntMadeira:
If this is not technically easy, why not implement a simpler solution, like a info tooltip or something just informative, like I suggest [here](https://github.com/openstreetmap/openstreetmap-website/issues/1115#issuecomment-609461120)?

PR #4461, opened 2024-01-06
Add language dropdown menu to My Preferences

AntonKhorev:
Implements https://github.com/openstreetmap/openstreetmap-website/issues/2581 in https://www.openstreetmap.org/preferences/edit

image

image


tomhughes:
There is some overlap here with #3618 which attempted to add a locale selector to the front page, only in as much as that if we did both we'd want to use the same selector in both places.

I've just rebased that on the current code (there's a live version at https://locale.apis.dev.openstreetmap.org/) but the styling of the selector is broken - think it's some conflict between bootstrap and the wikimedia styling of the language selector :disappointed:

tomhughes (commented):
Is adding new languages at the start the best choice? Won't most users add languages starting with their first choice and then moving to other languages they understand?

AntonKhorev:
If a language is added at the end, a user may choose the language, press the Add button and expect the ui to switch to that language. But it's probably not going to happen because another language that was in the preferences before is also supported and still has higher priority.

AntonKhorev:
> Won't most users add languages starting with their first choice and then moving to other languages they understand?

Maybe, but I expect most users to pick only one language. If they want to pick several, they are probably used to editing strings with multiple language codes and they are not going to use the dropdown at all.

AntonKhorev:
Maybe change the button label from "Add" to "Prepend"?

AntonKhorev:
Superseded by #6024

PR #4463, opened 2024-01-08
Changeset page fixes

AntonKhorev:
- convert changeset comment related tests to system tests because https://github.com/openstreetmap/openstreetmap-website/pull/4313#discussion_r1395452804 and similar tests for notes are also system tests
- shorten changeset page view code, remove unnecessary attributes (`class="notice"`, `class="action-button`, `data-changeset-id`)
- convert comment hide/unhide `<span>`s to buttons, now they can get keyboard focus
- display error message when subscribe fails, similar to https://github.com/openstreetmap/openstreetmap-website/pull/4462


tomhughes:
I'm not at all sure I agree with @gravitystorm about the tests... I mean in theory he's right but system tests have a huge cost disadvantage over controller tests and the environment for running them is held together with sticky tape and can fall apart at a moments notice leaving us with degraded tests until we can figure out how to make it work again.

Even if it is the right thing to do in general asserting that the page contains a given piece of text somewhere is in no way equivalent to properly checking the HTML structure as the controller test was doing.

AntonKhorev:
But do you want test like this that check that there's a form with `action='#'` inside some unspecified `div`?

```ruby
    assert_select "div#content" do
      assert_select "div#sidebar" do
        assert_select "div#sidebar_content" do
          assert_select "div" do
            assert_select "form[action='#']" do
              assert_select "textarea[name=text]"
            end
          end
        end
      end
    end
```

The exact structure is not important and I change it here anyway. The important thing is that it is inside the sidebar.

How about tests like this?

```ruby
    within_sidebar do
      assert_no_button "Comment"
      assert_text "Changeset still open"
    end
```

AntonKhorev:
and you can't test any interactions on changeset pages with controller tests because javascript

tomhughes:
I think this all looks good, so lets get it merged...

PR #4464, opened 2024-01-08
Remove hardcoded rate limit value. Fixes #4427

### LINKED ISSUE #4427: Hardcoded rate limit value leads to slow tests
In the calculation for the changeset_comments rate limit, there is a hardcoded value of "200" for deciding if a user is experienced or not.

https://github.com/openstreetmap/openstreetmap-website/blob/b09fc8373d31751a3d677cc8c3cae714a2280ecd/app/models/user.rb#L414

This has to be matched by the same hardcoded value in the tests

https://github.com/openstreetmap/openstreetmap-website/blob/b09fc8373d31751a3d677cc8c3cae714a2280ecd/test/controllers/api/changeset_comments_controller_test.rb#L161

However, creating 200 dummy changeset comments can take a few seconds. It would be a nice-to-have to allow this value to be overridden, and a lower (faster) value to be used in the test suite. Reducing the value from 200 to 20 reduces that test from 5.6 to 1.6 seconds on my machine.

It's also a good idea to avoid magic numbers in the code anyway. It took me a while to figure that the number in the test suite was a specific number, and not just "some arbitrary large number".

GobinathAL:


config/settings.yml
```diff
@@ -63,6 +63,7 @@ max_friends_per_hour: 60
 min_changeset_comments_per_hour: 1
 initial_changeset_comments_per_hour: 6
 max_changeset_comments_per_hour: 60
+min_changeset_comments_for_max_rate_limit: 200
```
tomhughes, line 66: That's a very long name and also inaccurate because it's the maximum number of changesets we consider, not the minimum.

I'm not sure what the best name is though - maybe something like `changeset_comments_for_rate_limit` would do? It should also be moved above the other settings I think rather then intermingling it with the `per_hour` settings.

tomhughes (changes requested):

config/settings.yml
GobinathAL: Named it  min_changeset_comments_for_max_rate_limit because it's the minimum number of changeset comments a user should have to get the max rate limit of 60 per hour. But i agree it's long.

mmd-osm: @GobinathAL : can you update the pull request as suggested, please?

GobinathAL: @mmd-osm I've updated it.

AntonKhorev: You don't need another commit to rename a thing you added in a previous commit. One commit is enough here.

tomhughes (approved):
Apologies for the long delay in looking at this again.

Having now done so it seems I probably misunderstood what was going on before and my suggestion to move the setting was completely wrong and my suggested named wasn't ideal either.

Rather than bother you to fix it again I've done it myself, moving the name back to where it was and switching to a name that parallels the existing name in the edit rate limit code that uses the same algorithm.

Thanks for doing all the hard work of finding what needed changing, and for persisting with this. I'll merge this now.

PR #4469, opened 2024-01-10
Remove OpenLayers build config file

AntonKhorev:
All of the OpenLayers files were removed in https://github.com/openstreetmap/openstreetmap-website/pull/2729. They are still served at https://www.openstreetmap.org/openlayers/OpenLayers.js https://www.openstreetmap.org/openlayers/img/marker.png etc.

But you certainly don't need the build config file here. You're not doing any new builds and it's probably outdated.


AntonKhorev:
I had to dig deep into openlayers commit history to see any `.cfg` files in use:

https://github.com/openlayers/openlayers/tree/613b90c7f3ea1ac0be01dd175754c636527c89f5/build

PR #4473, opened 2024-01-12
Use test() instead of define_method() to add locale-specific tests

AntonKhorev:


tomhughes:
That's a bit nicer ;-)

PR #4475, opened 2024-01-12
Remove temporary "_only" suffixes from map key entries

AntonKhorev:
They were required to avoid failing tests when both string-valued and array-valued versions of entries existed.


tomhughes:
Looks good, and thanks for double checking so thoroughly.

PR #4478, opened 2024-01-13
Remove element web abilities that were copied to api

AntonKhorev:
When this disentanglement https://github.com/openstreetmap/openstreetmap-website/commit/7b057545c0b2030aad9981bd93699f9e33ad7d5f happened, some abilities we left in `ability.rb` despite being api-related and copied to `api_ability.rb`.

Element abilities are handled by `browse_controller.rb` on the web side.


tomhughes:
Nice find!

Issue #2093, opened 2018-12-31
Geolocation marker color now black


HolgerJeromin:
I am not sure, but i am pretty sure that the central circle of the marker was not black.

![rps20181231_071829_168](https://user-images.githubusercontent.com/2410353/50555412-62aedf00-0ccc-11e9-8509-6efb82b3baf2.jpg)

tomhughes:
Well as far as I know nothing has changed recently...

tomhughes:
As far as I can see that dot has no styling at all so it is just inheriting the default body colour - maybe there has been a recent change in browsers to make them inherit colours from HTML into an SVG element?

tomhughes:
Ah no this is an upstream change in `Leaflet.locate` which was merged in to our code a couple of months ago (https://github.com/openstreetmap/openstreetmap-website/commit/fde98f7ead8b85a6a2b3ed748ca6f5a199a8e09f#diff-553fba9048416e7d29b3a167a39d676fL104).

tomhughes:
Actually I think the bigger problem is the CSP blocking inline styling...

PR #4479, opened 2024-01-14
Preview colour tag value using svg

AntonKhorev:
These previews originally accepted colors as inline styles https://github.com/openstreetmap/openstreetmap-website/pull/1779#issuecomment-374135049. That was later changed to data attributes + javascript https://github.com/openstreetmap/openstreetmap-website/pull/1779#issuecomment-374449624. We can avoid javascript by using inline svgs. We already do this in the map key.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -346,10 +346,6 @@ $(document).ready(function () {
           });
         }
       });
-
-      $(".colour-preview-box").each(function () {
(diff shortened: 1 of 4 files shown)


AntonKhorev:
By the way style attributes are not currently disallowed by CSP.

tomhughes:
Yes we had to allow it on pages which display the slippy map as leaflet uses inline styles - that was #2093.

PR #720, opened 2014-03-14
Add view pages for specific versions

UrszulaPapajak:
Fixes #662

PR #4480, opened 2024-01-16
Element version pages

AntonKhorev:
Basic support for https://github.com/openstreetmap/openstreetmap-website/issues/662. Doesn't try to show the old versions on the map.

The paths scheme is `/node/123/history/4` like it was in https://github.com/openstreetmap/openstreetmap-website/pull/720. This is because you'll probably want resourceful routes, with index pointed later to `/node/123/history`.

image

image

Not necessarily a replacement of https://github.com/openstreetmap/openstreetmap-website/pull/4468. You may still want links to specific versions within full history pages, especially if they'll get more version comparison functionality, for example like it was attempted in https://github.com/openstreetmap/openstreetmap-website/pull/3535.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -381,6 +381,7 @@ $(document).ready(function () {
     "/node/:id(/history)": OSM.Browse(map, "node"),
     "/node/:id/history/:version": OSM.OldBrowse(),
     "/way/:id(/history)": OSM.Browse(map, "way"),
+    "/way/:id/history/:version": OSM.OldBrowse(),
     "/relation/:id(/history)": OSM.Browse(map, "relation"),
     "/changeset/:id": OSM.Changeset(map),
     "/query": OSM.Query(map)
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -383,6 +383,7 @@ $(document).ready(function () {
     "/way/:id(/history)": OSM.Browse(map, "way"),
     "/way/:id/history/:version": OSM.OldBrowse(),
     "/relation/:id(/history)": OSM.Browse(map, "relation"),
+    "/relation/:id/history/:version": OSM.OldBrowse(),
     "/changeset/:id": OSM.Changeset(map),
     "/query": OSM.Query(map)
   });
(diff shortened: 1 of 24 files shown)


app/controllers/old_nodes_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  before_action -> { check_database_readable(:need_api => true) }
+  before_action :require_oauth
+
+  authorize_resource
+
+  around_action :web_timeout
+
+  def show
+    @type = "node"
+    @feature = OldNode.preload(:old_tags, :changeset => [:changeset_tags, :user]).find([params[:id], params[:version]])
```
tomhughes, line 15: In the `browse#node` action for the current view we preload the containing relations here - is that not possible for the historical objects? We still seem to be rendering them?

app/assets/javascripts/index.js
```diff
@@ -355,6 +355,17 @@ $(document).ready(function () {
     return page;
   };
 
+  OSM.OldBrowse = function () {
+    var page = {};
+
+    page.pushstate = page.popstate = function (path) {
+      OSM.loadSidebarContent(path, function () {
```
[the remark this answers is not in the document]
AntonKhorev: Dropped.

It used to set up color previews and will be needed later to draw the elements on the map.

app/assets/javascripts/index.js
```diff
@@ -369,8 +380,11 @@ $(document).ready(function () {
     "/user/:display_name/history": history,
     "/note/:id": OSM.Note(map),
     "/node/:id(/history)": OSM.Browse(map, "node"),
+    "/node/:id/history/:version": OSM.OldBrowse(map, "node"),
```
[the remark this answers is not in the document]
AntonKhorev: Removed.

It doesn't take any arguments for now, while I'm not attempting to draw anything.

app/controllers/old_nodes_controller.rb
AntonKhorev: No, we are not rendering them for historic elements.

https://github.com/openstreetmap/openstreetmap-website/blob/6163b01f146d97201e0c2d9efff83dd5dbadebf7/app/models/old_node.rb#L98-L101

tomhughes: Ah I had missing that the method didn't actually return anything!

PR #4485, opened 2024-01-17
Drop lowercase index on display names

tomhughes:
It's no longer needed as we now use the new canonical index...


gravitystorm:
Great, thanks.

PR #4488, opened 2024-01-19
Remove ":" from login form input labels

AntonKhorev:
On the login page redesign PR there's a discussion about colons in input labels starting at https://github.com/openstreetmap/openstreetmap-website/pull/4455#issuecomment-1899984034. The colons are in translation strings, thus they can be dealt with outside of that PR.

Those colons are not needed because they are mostly not used in other places. One place that has a colon is an input on the "forgot password" page, let's remove it from there too.

Another place is the share panel. I'm not so sure about colons there and I'm not removing them in this PR.

image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1753,7 +1753,7 @@ en:
     new:
       title: "Lost password"
       heading: "Forgotten Password?"
-      email address: "Email Address:"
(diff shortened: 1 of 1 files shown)


gravitystorm:
Sounds good, thanks. I agree that the share panel needs some more thought.

PR #4490, opened 2024-01-20
Add previous/next page arrows to newer/older pagination

AntonKhorev:
For [traces](https://www.openstreetmap.org/traces), [diaries](https://www.openstreetmap.org/diary) and comments pages.

image

image

image

Arrows are not displayed on extra small screen sizes because they can make things worse.

image


tomhughes:
Looks good, thanks.

PR #4492, opened 2024-01-23
Set OpenID tabindex higher than username/password form tabindexes

AntonKhorev:
If you press tab while in the OpenID input, you'll get to the login button _above it_, not below. Same thing happens in #4455. `:tabindex => 3` was there since at least 2010 while things around it changed.

image


tomhughes:
I suspect approximately nobody uses OpenID but nice catch!

PR #4496, opened 2024-01-25
Return an error when a disabled authentication mechanism is used

tomhughes:
This extends the previous work to allow authentication basic auth and OAuth 1 to be disabled to return an error message so that users have some chance of discovering what is going on.

I'll make this a draft for now as we may want to add more detail to the messages...

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -107,7 +107,7 @@ class ApiController < ApplicationController
       self.current_user = User.find(doorkeeper_token.resource_owner_id)
     elsif Authenticator.new(self, [:token]).allow?
       # self.current_user setup by OAuth
-    elsif Settings.basic_auth_support
+    else
       username, passwd = auth_data # parse from headers
       # authenticate per-scheme
       self.current_user = if username.nil?
@@ -117,8 +117,14 @@ class ApiController < ApplicationController
                           else
                             User.authenticate(:username => username, :password => passwd) # basic auth
                           end
-      # log if we have authenticated using basic auth
-      logger.info "Authenticated as user #{current_user.id} using basic authentication" if current_user
+      if username && current_user
+        if Settings.basic_auth_support
+          # log if we have authenticated using basic auth
+          logger.info "Authenticated as user #{current_user.id} using basic authentication"
+        else
+          report_error t("application.basic_auth_disabled", :link => t("application.auth_disabled_link")), :forbidden
+        end
+      end
     end
 
     # have we identified the user?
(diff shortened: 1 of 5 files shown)


mmd-osm:
Could this be implemented as Apache rewrite rule, to also cover CGImap without much extra effort?

mmd-osm:
Assuming we set `basic_auth_support: false`, an API user would see one of these two error messages:

* HTTP 403 HTTP Basic Authentication is disabled 
* HTTP 403 Couldn't authenticate you

I'm wondering if this makes much of a difference for a user, when they can't use Basic auth anyway and need to switch to OAuth 2. Tbh, I'm struggling a bit to see the benefit of checking whether the credentials are valid.

tomhughes:
To minimise the difference from normal behaviour, and hence the risk of breaking things in unexpected ways?

tomhughes:
I've added a link to a stub wiki page in the error message so we have a way to provide extra information and cleared the draft flag as this is now ready for review.

Operationally to deal with the cgimap issues we could force anything with a basic auth header into rails instead of cgimap during the brownouts so they get the better error but that's hard for OAuth 1 as there are multiple ways of providing the signature I think and some may not be visible to apache, notably if they are in POST parameters.

mmd-osm:
> Operationally to deal with the cgimap issues we could force anything with a basic auth header into rails instead of cgimap during the brownouts so they get the better error

I'm not sure this is really needed. Let's take a look at the different scenarios.

Out of the four endpoints to do changes on the database, only the changeset upload is handled by CGImap. As soon as there are no open changesets left anymore, they're stuck with the changeset create on Rails. The error message on Rails would suffice in this case.

The user based rate limiting for various fetch element endpoints is implemented on CGImap only. I suppose if we're diverting these calls to Rails, it would not trigger an error there.

If we want to make users aware that their rate limiting isn't working anymore, this would have to be done on Apache or CGImap then?

tomhughes:
Can I give this a bump, given we're supposed to be doing brownouts like now, and disabling the old auth methods at the end of the month?

gravitystorm:
OK, this looks fine to me. 

I had to do some reading as to the differences between OAuth 1.0 vs 1.0a, and also (re-)learning about how the rails filter chains work since it wasn't initially obvious to me that rendering in the filter was enough to stop the rest of the action from running too.

It's a little complex to understand how the `Settings.oauth_10_support` and `Settings.oauth_10a_support` interact, since the second also affects oauth 1.0 tokens, but hopefully we can clean this up soon. In reality, OAuth 1.0 is already turned off anyway.

With regards to Basic Auth, we'll need to do some work to remove usage of basic_auth in our tests in the near future.

PR #4499, opened 2024-01-26
Improve styling of help page

tomhughes:
Drop the redundant links in the footer and turn the title into a header and extend it to cover the whole card.


gravitystorm:
Nice, definitely better.

PR #4500, opened 2024-01-28
Allow registration of OAuth 1.0 applications to be disabled

tomhughes:


app/controllers/oauth_clients_controller.rb
```diff
@@ -19,7 +19,11 @@ def show
   end
 
   def new
-    @client_application = ClientApplication.new
+    if Settings.oauth_10_registration
+      @client_application = ClientApplication.new
+    else
+      redirect_to :action => "index"
```
gravitystorm, line 25: I think this needs a flash message. If we're doing more than just removing the link from the index page, then I guess we're trying to handle external links to `:new` e.g. from a tutorial somewhere. A flash message would explain why the link to `:new` no longer works.

gravitystorm: Does `:create` need a similar redirect? Or are we happy that if the form is removed, then anyone going to the lengths of making a direct `:post` with the correct CSRF token is only causing problems for themselves?

config/settings.yml
```diff
@@ -97,6 +97,7 @@ attachments_dir: ":rails_root/public/attachments"
 basic_auth_support: true
 # Enable legacy OAuth 1.0 support
 oauth_10_support: true
+oauth_10_registration: true
```
gravitystorm, line 100: I think just a simple test for the `:new -> :index` redirect would be needed and is also worthwhile.

gravitystorm (changes requested):
Other than the inline comments, looks good to me.

app/controllers/oauth_clients_controller.rb
tomhughes: I've added the flash message - as you say I didn't bother with `:create` because you'd have to try really hard and this is only going to be short lived.

config/settings.yml
tomhughes: I've added a test and then spent ages wondering why it broke everything only to eventually figure out that yesterday's bundle update gave us a buggy bootsnap!

gravitystorm: > yesterday's bundle update gave us a buggy bootsnap!

Sorry I found that a couple of hours ago, but I was still deep-diving on something else so hadn't made a PR for bootsnap yet!

gravitystorm:
Thanks

Issue #4471, opened 2024-01-11
Trim username/e-mail whitespace in login page


zstadler:
### Problem

OSM users are redirected to OSM login for authentication. We continuously receive complaints about users unable to access their accounts.

95% of the user issues are due to extra space added to the OSM username/email that is added by many mobile keyboards with auto-completion. Users do not notice that extra space and their login fails

### Description

The requested solution to the problem is to trim leading and trailing whitespace. It is a safe change since leading and trailing whitespace characters are not permitted as OSM usernames or an e-mail address

### Screenshots

The error produced when a space character is appended to the username, as done by many mobile keyboards with auto-completion

![image](https://github.com/openstreetmap/openstreetmap-website/assets/1304610/66999f36-6b57-4e02-a286-cd4e213856cf)

tomhughes:
We already have validations on the user model to prevent leading and trailing whitespace on these fields?

zstadler:
Apparently, we do. 

For leading whitespace
![image](https://github.com/openstreetmap/openstreetmap-website/assets/1304610/106de83b-0b02-41c5-bdfd-11ecf3bd139c)

For trailing whitespace:
![image](https://github.com/openstreetmap/openstreetmap-website/assets/1304610/869f534e-1e6d-4c06-9152-b647e3ca60dc)

birante:
Hi everyone @tomhughes @zstadler,

I'm Birante, a newcomer. I've been working on this bug and have created a [PR](https://github.com/openstreetmap/openstreetmap-website/pull/4501) to fix it. I would appreciate your feedback on my proposal. Thank a lot.

divyanshuagarwal-23:
Hi everyone @zstadler I am working on this issue
1. Please assign this issue to me
2. is there any slack channel I can join, as I am new to this project

PR #4501, opened 2024-01-29
Remove whitespace on signup fields #4471

birante:


gravitystorm (changes requested):
Hi @birante , thank you for this pull request.

First, I think there is a misunderstanding about the problem described in #4501. The issue described is for the login page (not the signup page), and it is for the situation where someone accidentally includes leading/trailing whitespace when trying to sign in.

Secondly, a PR to fix this issue will need tests, to prove that it works, and to prevent any future breakage if we ever refactor the code.

Finally, I would be interested to see if there is a rubygem that is already written to tackle this problem, so that we can easily apply the same solution to other forms on the website. For example, we could use something like https://github.com/holli/auto_strip_attributes

tomhughes:
I don't think a gem that auto strips the model attributes is going to help here because the login code is not actually building a model at all - if it was the model would fail validation because we already have validations for this.

The problem basically is https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/sessions_controller.rb#L50 which creates (or rather looks up) the model using the provided username and password.

gravitystorm:
> I don't think a gem that auto strips the model attributes is going to help here because the login code is not actually building a model at all

Good point.

birante:
hello @tomhughes  @gravitystorm  thanks a lot for the feedback and sorry for the misunderstanding about the issue.

I will fix it write the test and push back.

Best.

PR #4502, opened 2024-01-29
Show last authorization time for OAuth 2 applications

### LINKED ISSUE #4494: Display authenticatoin-related timestamps on Oauth2 tab
### Problem

When logged in, on `Oauth1 settings` tab `https://www.openstreetmap.org/user/MYUSERNAME/oauth_clients` I can see the  application name and **when it was issued**.

However, on `OAuth 2 authorizations` tab at https://www.openstreetmap.org/oauth2/authorized_applications there is no timestamp column, only `Application` and `Permissions` .

### Description

I find timestamp information very valuable for my authorizations. I'd like to suggest adding two additional columns to `OAuth 2 authorizations` tab:

- timestamp where it was issued (i.e. just like Oauth1 tab shows)
- timestamp when that authentication token was last used (that helps a lot determining if it should be revoked).

### Screenshots

_No response_

tomhughes:
Fixes #4494

### DIFF
--- a/app/views/oauth2_authorized_applications/_application.html.erb
+++ b/app/views/oauth2_authorized_applications/_application.html.erb
@@ -9,6 +9,9 @@
       <% end -%>
     </ul>
   </td>
+  <td class="align-middle">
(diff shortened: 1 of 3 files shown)


gravitystorm:
LGTM, thanks

PR #4505, opened 2024-02-03
Update en.yml to mention how to unsubscribe

jidanni:
Email was missing unsubscribe link.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1569,8 +1569,8 @@ en:
       hi: "Hi %{to_user},"
       header: "%{from_user} has commented on the OpenStreetMap diary entry with the subject %{subject}:"
       header_html: "%{from_user} has commented on the OpenStreetMap diary entry with the subject %{subject}:"
-      footer: "You can also read the comment at %{readurl} and you can comment at %{commenturl} or send a message to the author at %{replyurl}"
-      footer_html: "You can also read the comment at %{readurl} and you can comment at %{commenturl} or send a message to the author at %{replyurl}"
+      footer: "You can also read or unsubscribe to the comment at %{readurl} and you can comment at %{commenturl} or send a message to the author at %{replyurl}"
+      footer_html: "You can also read or unsubscribe to the comment at %{readurl} and you can comment at %{commenturl} or send a message to the author at %{replyurl}"
     message_notification:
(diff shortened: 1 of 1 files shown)


tomhughes:
The idea is fine but the actual wording is not - it should be "unubscribe from" but in any case you're unsubscribing from the diary entry not the specific comment.

It really needs to be a separate clause rather than being wedged in with replying like that, and then it could have it's own link directly to the unsubscribe action (though that might need some server work to support GET from a mail client).

tomhughes:
I'm kind of surprised Google thought they could show the email like that from a data protection point of view - certainly we can't though we could link to sending an OSM message.

The risk is that we wind up with developers being spammed by dozens of users demanding to know what's going on so I can't say I'm keen on it.

gravitystorm:
> I'm kind of surprised Google thought they could show the email like that from a data protection point of view - certainly we can't though we could link to sending an OSM message.
> 
> The risk is that we wind up with developers being spammed by dozens of users demanding to know what's going on so I can't say I'm keen on it.

I think the above comment was intended for a different issue - https://github.com/openstreetmap/operations/issues/867#issuecomment-1925203709

tomhughes:
Well the GET would redirect to the login page if necessary and then back to the original action like any other request?

It probably would need to then ask for confirmation before unsubscribing as we don't want accidental unsubscribes from things trying to "check" the URL.

AntonKhorev:
I was going to say that with "Unsubscribe" [moved](https://github.com/openstreetmap/openstreetmap-website/pull/4518) the `#comments` link can serve as the unsubscribe link. But then I thought what if the user is not logged in? Then there would be no button.

AntonKhorev:
I guess it should be a page like this one

image

when `/user/*/diary/*/unsubscribe` is opened with GET.

AntonKhorev:
Is it done now with https://github.com/openstreetmap/openstreetmap-website/pull/4523 and https://github.com/openstreetmap/openstreetmap-website/commit/f0764d3ecae3aa9251eadff96990f4afe530924d for changesets?

mmd-osm:
I think #4523 supersedes this pull request, and I don't see any follow up comments from OP except for some "eyes" emoji.
Closing here. Thanks!

PR #4511, opened 2024-02-10
Adjust button order on welcome banner

RicoElectrico:
Make more emphasis on edit button by changing "learn more" button style to outlined and also moving it to the right (following the reading order).

This is in line with UX best practices, hopefully could increase conversions slightly.

The proposed layout would be like this:

obraz


tomhughes:
What is your definition of reading order? Surely that is left to right and the whole point is to have "learn more" before "start mapping" to try and encourage people to read before they dive in?

hlfan (approved):

Issue #4506, opened 2024-02-03
History navigation inconsistent  


wireguy:
### URL

https://www.openstreetmap.org/relation/1941310

### How to reproduce the issue?

going to: [https://www.openstreetmap.org/relation/1941310](https://www.openstreetmap.org/relation/1941310)
shows the latest version of the relation, 5, but the history navigation has the previous version as 1, and the next version as 5:

<img width="237" alt="image" src="https://github.com/openstreetmap/openstreetmap-website/assets/47227325/6ec52a9d-0812-4d0d-b8c8-fcbf673a3a1b">

going to [https://www.openstreetmap.org/relation/1941310/history/5](https://www.openstreetmap.org/relation/1941310/history/5)
also shows version 5 of the relation, but the history navigation has the previous version as 4, and no next version

<img width="232" alt="image" src="https://github.com/openstreetmap/openstreetmap-website/assets/47227325/81f5b3e9-3f9e-4640-93ee-bea7b6d24e0f">

The second way seems correct, from version 5, the previous should be version 4.

### Screenshot(s) or anything else?

issue seems to be in:

[app/views/browse/feature.html.erb](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/views/browse/feature.html.erb)

tomhughes:
That is the intended behaviour - the "current" version view doesn't have next/previous version buttons because that would make no sense. Instead it has first/last buttons which take you into the history view and there you have next/previous buttons.

It's not ideal that the icons are the same but I don't have any better ideas either...

wireguy:
ideas:

<img width="156" alt="image" src="https://github.com/openstreetmap/openstreetmap-website/assets/47227325/fdfbbafe-7e7c-43fc-ac2d-c2dd953a469e">

or just simple text 

<<   <   5  >  >>
|<    <  5  >   >|

<< or  |< go to the first
< go to the previous

AntonKhorev:
I plan to have < > icons for previous/next versions and << >> for first/last. That's why I started doing svg icons that are currently on diaries/traces pages.

tomhughes:
I believe this has now been resolved by #4513.

PR #4513, opened 2024-02-10
SVG icons in element version and changeset links

AntonKhorev:
Links to first/last versions on the current version page:
image

Now they look differently to previous/next version links, to avoid confusion #4506.

Links to previous/next versions on the old version page:
image

And also on user's changeset pages, that's where the initial "<<" / ">>" link style came from:
image

### DIFF
--- a/app/helpers/svg_helper.rb
+++ b/app/helpers/svg_helper.rb
@@ -44,7 +44,7 @@ module SvgHelper
     height = 15
     pad = 2
     segment = (0.5 * height) - pad
-    width = segment + (2 * pad)
+    width = (segment + (2 * pad)).ceil
     path_data = "M#{side * (pad - (0.5 * width))},#{pad} l#{side * segment},#{segment} l#{-side * segment},#{segment}"
     path_tag = tag.path :d => path_data, :fill => "none", :stroke => "currentColor", :"stroke-width" => 1.5
     tag.svg path_tag, :width => width, :height => height, :viewBox => "-#{0.5 * width} 0 #{width} #{height}", :class => options[:class]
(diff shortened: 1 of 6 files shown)


Dimitar5555:
Are SVGs really necessary? There are symbols with a similar look (https://en.wikipedia.org/wiki/Guillemet) that would also work.

tomhughes:
Can you explain why you think that's better? If we went that route is there a risk that an unusual character might not render on some browsers if they're using a font that doesn't have it?

grischard:
Guillemets are standard for quotes in so many languages that I'm more concerned about this appearing like a quote to some than a font not having them.

Dimitar5555:
> Can you explain why you think that's better?

It's a single a single character that requires only a single `if`and no calculations.

> If we went that route is there a risk that an unusual character might not render on some browsers if they're using a font that doesn't have it?

> Guillemets are standard for quotes in so many languages that I'm more concerned about this appearing like a quote to some than a font not having them.

Looking at [character font support table](https://www.fileformat.info/info/unicode/char/00AB/fontsupport.htm), almost all (if not all) major fonts support it and have a similar rendering. I don't see any font that renders them like standard quotation marks,

Morsealfabetet, LastResort, Ekushey have different rendering but it's not quotes. The first two are quite specific, the last one is a Bengali font which may or may not be widely used.

HolgerJeromin:
There are many [quote style](https://en.m.wikipedia.org/wiki/Quotation_mark#Summary_table) so the problem described was

> Guillemets are standard for quotes

and not

> some Guillemets are looking like quotes

tomhughes:
Looks good to me, and it's consistent with what we're already doing in other places so I don't see any reason to take a different approach here.

PR #4515, opened 2024-02-11
SessionsController: strip username

### LINKED ISSUE #4471: Trim username/e-mail whitespace in login page

simon04:
Fixes #4471. Closes #4501.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -20,7 +20,7 @@ class SessionsController < ApplicationController
   def create
     session[:remember_me] ||= params[:remember_me]
     session[:referer] = safe_referer(params[:referer]) if params[:referer]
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good, thanks.

zstadler:
Thank you!

PR #4516, opened 2024-02-11
Update package.json versions from yarn.lock

simon04:
As a developer, I'd like to see the versions in use from the package.json file. This matches the behaviour when running `yarn upgrade-interactive --latest` to update versions manually.

Re-configure dependabot's versioning-strategy accordingly, see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "npm"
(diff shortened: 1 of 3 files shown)


gravitystorm:
I'm going to close this one now, but feel free to reopen if you want to discuss it further.

PR #4517, opened 2024-02-11
Move previous/next user's changeset links to bottom

AntonKhorev:
Links in the sidebar on /changeset/_id_ pages before:
image

After:
image

The idea is that "Changeset XML" and "osmChange XML" are related to the current changeset and should be closer to the changeset details than links for navigation between changesets.

A similar order is already in use on element version pages:

image


tomhughes:
Certainly seems reasonable that we should be consistent in the ordering at least.

Issue #4491, opened 2024-01-23
Comment vs. Subscribe on diary posts


jidanni:
### Problem

1. Fill out a comment on somebody's diary post.
2. At the bottom you will find two buttons:
- Comment, and
- Subscribe

The user thinks s/he can only pick one.

Yes, if one presses "comment" one does also get subscribed, but the interface makes it look like one must make a choice.

gravitystorm:
I agree, the design here isn't great. I've thought about it before without coming up with an obvious good solution.

divyanshuagarwal-23:
Hi everyone @jidanni  I am working on this issue

Please assign this issue to me
is there any slack channel I can join, as I am new to this project

jidanni:
@divyanshuagarwal-23 I would love to,  but I'm not in charge here.

AntonKhorev:
Something like this, with the _Subscribe_ button above comments, similar to changeset discussion?

![image](https://github.com/openstreetmap/openstreetmap-website/assets/4158490/0b977401-3c74-400c-93f7-8c146e19b941)

AntonKhorev:
After #4518 the Subscribe button is no longer at the bottom.

PR #4518, opened 2024-02-11
Add diary discussion heading and move subscribe button next to it

AntonKhorev:
Places (un)subscribe button on diary entry pages like it's on changeset pages. That requires adding a heading above the comments to make it clear to what you're subscribing.

image

Previously the button was right below the new comment box which is somewhat confusing, see #4491. You don't want to accidentally press it after typing in a comment.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -827,14 +827,11 @@ tr.turn:hover {
     height: 400px;
     display: none;
   }
-  .comments {
-    max-width: 740px;
+  .diary-comment .col-auto {
+    width: 62px;
   }
-  .diary-comment {
-    border-top: 1px dashed $grey;
-    &:first-child {
-      border-top: 1px solid $grey;
-    }
+  .diary-comment .col {
+    max-width: 690px;
   }
 }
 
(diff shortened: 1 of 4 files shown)


tomhughes:
I see that there's already a dotted line between comments which doesn't do that though... I was looking at an entry with only one comment originally so I didn't see that. Personally I think they should all be full width?

AntonKhorev:
I wanted to remove dotted lines later because diary comments is the only place where they are used.

tomhughes:
I can see you could maybe remove them between the comments but I think you'd still want one at the top otherwise the Subscribe button will wind up floating in the middle of the page vertically - the horizontal line here serves to "anchor" it to the rest of the layout which is why I thought the line should go right over with the intention that the button would also be pushed over.

AntonKhorev:
- made all lines full width while keeping max comment width
- changed dashed lines to lighter solid lines
- made avatar column constant width (otherwise it depends on image width)
- decreased gaps between avatars and comments

with comments:
image

with no comments:
image

tomhughes:
That looks much better, thanks. I've added one more tweak in 168087eba0556d427bd21fd0272bc76afeb6df00 to stop the lines overflowing into the negative margins that are part of how the bootstrap grid works.

PR #4522, opened 2024-02-14
Remove html type from changeset feed entry titles

AntonKhorev:
Look at https://github.com/openstreetmap/openstreetmap-website/commit/fc9768f34c803f9a9746d1705af097655d55f3e2. It added `type="html"` to changeset feed entries for the reasons stated in the commit message. But is it actually correct? Let's check.

I created a couple of changesets. This is how their comments are shown on the `/history` page. I expect to see the same in a feed reader.

image

This is how their titles look in `/history/feed` output. They are always escaped once.

    <title type="html">Changeset 58 - is it going to &lt;em&gt;emphasize&lt;/em&gt;?</title>
    ...
    <title type="html">Changeset 57 - this is unescaped (&lt;); this is escaped (&amp;lt;)</title>

This is how they look in [Liferea](https://en.wikipedia.org/wiki/Liferea). Not what's expected.

image

This is how they look after `type="html"` is removed.

image

### DIFF
--- a/app/views/changesets/index.atom.builder
+++ b/app/views/changesets/index.atom.builder
@@ -25,7 +25,7 @@ atom_feed(:language => I18n.locale, :schema_date => 2009,
                  :type => "application/osmChange+xml"
 
       if !changeset.tags.empty? && changeset.tags.key?("comment")
-        entry.title t("browse.changeset.feed.title_comment", :id => changeset.id, :comment => changeset.tags["comment"]), :type => "html"
+        entry.title t("browse.changeset.feed.title_comment", :id => changeset.id, :comment => changeset.tags["comment"])
       else
         entry.title t("browse.changeset.feed.title", :id => changeset.id)
       end
--- a/test/controllers/changesets_controller_test.rb
+++ b/test/controllers/changesets_controller_test.rb
@@ -243,6 +243,7 @@ class ChangesetsControllerTest < ActionDispatch::IntegrationTest
     create(:changeset_tag, :changeset => changeset)
     create(:changeset_tag, :changeset => changeset, :k => "website", :v => "http://example.com/")
     closed_changeset = create(:changeset, :closed, :num_changes => 1)
+    create(:changeset_tag, :changeset => closed_changeset, :k => "website", :v => "https://osm.org/")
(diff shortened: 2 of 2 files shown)


AntonKhorev:
This is what https://validator.w3.org/feed/ says about changesets with comments like "<<<<<<<<"

image

tomhughes:
There has of course been a change to that code since the type attribute was added, namely removing the use of the `h` function to escape things, which might explain things.

tomhughes:
Fundamentally I think it was always wrong because changeset comment tags are text and don't permit HTML so while it may have appeared to fix whatever the apparent problem was it was the wrong fix because any escaping was XML escaping of text and not HTML at all.

tomhughes (commented):
I think it would probably be a good idea to have some tests here?

AntonKhorev:
> namely removing the use of the h function to escape things

If the comment was escaped twice, it couldn't contain `&lt;`.

I'm not sure what to test here aside from presence/absence of `type="html"`.

tomhughes:
I had in mind checking that the text content was as expected, with the right amount of escaping?

AntonKhorev:
But I'm not changing the text contents here.

Actually I discovered this while writing a test for another thing. I wanted to move locale strings from "browse" section to "changesets" because feeds are not handled by browse controller. The test was supposed to fail on missing locale strings. Maybe I'll add it to this pull request.

tomhughes:
Yes I know you're not changing the contents, but you are changing what you are asserting the contents is and I thought it would be useful to check the contents really were what we were saying.

We don't know whether in the past it really was double escaped and it was (sort of) correct to claim it was HTML or if the previous edit was just always wrong but either way a test might have revealed either that the previous edit was bogus or else when it stopped being correct because the escaping changed.

AntonKhorev:
Added tests for changeset entry titles

tomhughes:
Thanks. I merged that and tweaked the test slightly in 81857a744613f3a9f3aa4fb1848e9a9768e174c7 to include some special character to test the escaping is correct.

AntonKhorev:
Looks like I lost that commit where I put "<" in a changeset comment.

PR #4523, opened 2024-02-17
Add unsubscribe link to diary comment notification email

AntonKhorev:
as I propose in #4505:

- add diary subscribe/unsubscribe pages for GET requests that contain a button to make a POST request
  image
- these pages redirect to login if required
- add an unsubscribe page link to comment notification emails


tomhughes:
That all looks great, thanks.

Issue #4507, opened 2024-02-05
Add Home Location privacy notice


jidanni:
### Problem

Users don't realize their Home Location setting will be revealed to others, and not e.g., just to help them navigate quickly to their home location.

### Description

On https://www.openstreetmap.org/profile/edit > Home Location

Be sure to add a privacy disclosure:
- This location will be revealed to other users (on https://www.openstreetmap.org/dashboard and elsewhere.)

I was going to add it somewhere on config/locales/en.yml but I would probably mess
it up.

PR #4524, opened 2024-02-17
fixed tooltip

### LINKED ISSUE #4446: Rename the "Add a note to the map button" to "Report an error on map"
### Problem

The current title does not capture the essence and encourages misuse

### Description

Снимок экрана 2023-12-30 в 12 36 00

1. The name does not imply that you can use it to report errors
2. The current name causes people to leave personal notes on the map.

What similar buttons look like on other mapping sites:

<details>
<summary>Yandex Maps</summary>

Снимок экрана 2023-12-30 в 12 46 58

</details>

<details>
<summary>Mapbox</summary>

Снимок экрана 2023-12-30 в 12 47 56

</details>

### Screenshots

_No response_

SharanRP:
closes #4446

### DIFF
--- a/app/views/profiles/edit.html.erb
+++ b/app/views/profiles/edit.html.erb
@@ -43,6 +43,9 @@
   <fieldset>
     <legend><%= t ".home location" -%></legend>
     <p id="home_message" class="text-muted m-0<% if current_user.home_location? %> invisible<% end %>"><%= t ".no home location" %></p>
+    <p class="text-muted m-0<% if current_user.home_location? %> invisible<% end %>">
+       <b><%= t(".home_location_warning") %></b>
+    </p>
     <div class="row">
       <%= f.text_field :home_lat, :wrapper_class => "col-sm-4 d-flex flex-column", :class => "mt-auto", :id => "home_lat" %>
       <%= f.text_field :home_lon, :wrapper_class => "col-sm-4 d-flex flex-column", :class => "mt-auto", :id => "home_lon" %>
--- a/config/locales/en-GB.yml
+++ b/config/locales/en-GB.yml
@@ -1367,6 +1367,7 @@ en-GB:
       image size hint: (square images at least 100x100 work best)
       home location: Home Location
       no home location: You have not entered your home location.
+      home location warning : This location will be revealed to other users.
       update home location on click: Update home location when I click on the map?
   sessions:
     new:
--- a/config/locales/en-GB.yml
+++ b/config/locales/en-GB.yml
@@ -2098,7 +2098,7 @@ en-GB:
     site:
       edit_tooltip: Edit the map
       edit_disabled_tooltip: Zoom in to edit the map
-      createnote_tooltip: Add a note to the map
+      createnote_tooltip: Move the marker and write a note to fix map errors.
       createnote_disabled_tooltip: Zoom in to add a note to the map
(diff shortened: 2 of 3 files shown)


SharanRP:
also closes #4446

config/locales/en-GB.yml
```diff
@@ -1367,6 +1367,7 @@ en-GB:
       image size hint: (square images at least 100x100 work best)
       home location: Home Location
       no home location: You have not entered your home location.
+      home location warning : This location will be revealed to other users.
```
AntonKhorev, line 1370: You only need to edit config/locales/en.yml.

AntonKhorev:
You wanted to say that it also closes #4507. But why didn't you make two pull requests instead? Those are different issues that don't depend on each other.

config/locales/en.yml
```diff
@@ -3095,7 +3095,7 @@ en:
     site:
       edit_tooltip: Edit the map
       edit_disabled_tooltip: Zoom in to edit the map
-      createnote_tooltip: Add a note to the map
+      createnote_tooltip: Move the marker and write a note to fix map errors.
```
AntonKhorev, line 3098: You are writing about the marker in a tooltip that is shown before the marker is visible.

tomhughes:
As @AntonKhorev says these should be two separate PRs so please remove the second commit and submit it separately.

config/locales/en.yml
SharanRP: what should be written then?

SharanRP (commented):
I have made the required changes

config/locales/en-GB.yml
```diff
@@ -2098,7 +2098,7 @@ en-GB:
     site:
       edit_tooltip: Edit the map
       edit_disabled_tooltip: Zoom in to edit the map
-      createnote_tooltip: Add a note to the map
+      createnote_tooltip: Add a note to the map.
```
tomhughes, line 2101: Please don't make changes to `en-GB.yml` or any other locale file other than `en.yml` as translations need to be done on Translatewiki not here.

config/locales/en.yml
```diff
@@ -3095,7 +3096,7 @@ en:
     site:
       edit_tooltip: Edit the map
       edit_disabled_tooltip: Zoom in to edit the map
-      createnote_tooltip: Add a note to the map
+      createnote_tooltip: Add a note to the map.
```
tomhughes, line 3099: This was correct as it was - none of the other tooltips have a full stop.

app/views/profiles/edit.html.erb
```diff
@@ -43,6 +43,9 @@
   <fieldset>
     <legend><%= t ".home location" -%></legend>
     <p id="home_message" class="text-muted m-0<% if current_user.home_location? %> invisible<% end %>"><%= t ".no home location" %></p>
+    <p class="text-muted m-0<% if current_user.home_location? %> invisible<% end %>">
+       <b><%= t(".home_location_warning") %></b>
```
tomhughes, line 47: This change will no longer work as you've removed the `home_location_warning` string from the locale file - you were asked to move this to a separate PR which would have required moving both this and the string but instead you just removed the string from this PR and left this.

tomhughes (changes requested):

SharanRP:
ok I will open a new PR

PR #4529, opened 2024-02-19
Link to current and old element versions from changeset pages

AntonKhorev:
Changeset pages have links to changed elements. Previously elements only had current version and full history pages, so links were to current versions. Now there are pages for each element version too, and links on changeset pages have version number displayed. It makes sense to link that v-number part to that exact version.

image

image


tomhughes:
Looks good to me, thanks.

PR #4530, opened 2024-02-20
Move changeset feed locale strings from browse section to changesets

AntonKhorev:
Changeset feeds are not handled by browse controller, and I want to remove more stuff from browse controller later, possibly everything.


gravitystorm:
> I want to remove more stuff from browse controller later, possibly everything. 

Just as an aside, I would be quite happy to see that. I started this process last year by moving browse#note into notes#show, and so I think things like browse#relations -> relations#show etc would be a good idea too. I think the main thing would be putting the shared partials in a logical place (notes was easier since it didn't share any code with anything else in browse.

PR #4532, opened 2024-02-23
Convert note pagination to newer/older note pages

AntonKhorev:
image

A bit more work is required here because the order is unstable. It could be changed by any incoming note comment, causing before/after links to behave unexpectedly. I need to add dates to those links and use dates as fallbacks if that note was updated.


tomhughes:
There's a number of uncontroversial changes here which aren't directly related to this PR and could be handled separately but having now spent about half an hour staring at this the core of it concerns me - it seems to be introducing an extraordinary level of complexity to me and I don't claim to understand it all yet at all.

I wonder if we're missing a trick here - although this view is a list of notes what it's really doing is listing notes in order of their more recent comment so is there a way we can use the comment ID values, which are nice and stable, as the cursor?

AntonKhorev:
> is there a way we can use the comment ID values, which are nice and stable, as the cursor?

Something like this?

1. don't do #4481, otherwise not all notes have a comment
2. mark the last note comment / unmark previously last comment when updating the note
3. get marked comments sorted by ids

Not sure about table joins if we want to replicate the existing behavior where user's notes are all notes acted upon by the user. We get a comment then join it with a note, then join it again with comments to see if one of them is the user's.

PR #4535, opened 2024-02-24
Use rails generated tokens for emails

tomhughes:
This replaces our `user_tokens` table with a modern system based on rails builtin token generation capabilities which generates signed tokens using a key derived from the installations base secret avoiding the need to persist the tokens on disk.

It also scopes the tokens so that each token can only be used for the intended purpose.

Currently this still allows old tokens to be used - once this has been live for a few weeks a second PR will remove that support and drop the old table.


gravitystorm:
This is great.

PR #4536, opened 2024-02-24
Add a limit on the number of points in a GPS trace

tomhughes:
Currently I've set the default limit at 10 million points - we have exactly one trace larger than that in the database and this is from about seven years ago. In fact there are only two with over 5 million points.


tomhughes:
So after some discussion with @gravitystorm I ran a test import of the largest ever trace which contained about 15 million points and it took a bit over 7 hours to import on my desktop.

Looking at the database a total of 433 traces exist (out of around 10 million) with over one million points and the number per year is frequently in the single digits:

```
 extract | count 
---------+-------
    2008 |     4
    2009 |    10
    2010 |    10
    2011 |     8
    2012 |     7
    2013 |    11
    2014 |    30
    2015 |     4
    2016 |     5
    2017 |     7
    2018 |   230
    2019 |    12
    2020 |    70
    2021 |     8
    2022 |     6
    2023 |    11
(16 rows)
```

As a result the default has now been reduced to one million points though it remains configurable.

gravitystorm:
Thanks!

PR #4537, opened 2024-02-24
Use a reduced priority for batch trace uploads

tomhughes:
This sets the priority of the trace import job based on the number of pending traces the user has so the more they up load at once the lower the priority, thus allowing other users who only upload a few to jump in front of them.

### DIFF
--- a/app/controllers/api/traces_controller.rb
+++ b/app/controllers/api/traces_controller.rb
@@ -35,7 +35,7 @@ module Api
         trace = do_create(params[:file], tags, description, visibility)
 
         if trace.id
-          TraceImporterJob.perform_later(trace)
+          trace.schedule_import
           render :plain => trace.id.to_s
         elsif trace.valid?
           head :internal_server_error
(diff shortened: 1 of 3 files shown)


gravitystorm (commented):
+1 for making the queue more fair.

I think the risk with this approach is that if someone adds 1000 traces, and then keeps adding more batches 50 at a time before the original 1000 are complete, then the final traces from the original batch (e.g. with priorities 990-999) might sit in the queue for a long long time, while subsequent traces are processed first (since they go in the queue with e.g. 400-450).

I'm not sure if I'm overthinking this though, maybe it's not a real problem.

https://www.salsify.com/blog/engineering/queue-fairness touches on this same topic (this PR implements "the first solution" and it doesn't seem straightforward to implement a better algorithm at the point of dequeuing (rather than setting the priority at the point of enqueuing).

I'm surprised there's no gems to do this, it seems quite a common topic to discuss fair dequeuing for multi-user queues

gravitystorm:
> I'm not sure if I'm overthinking this though, maybe it's not a real problem.

I think a more realistic problem is that if the queue is never fully drained, the lowest-priority jobs might never be processed (e.g. if there's always at least `num_workers` "priority-0" jobs in the queue, "priority-1000" jobs will never be processed). But looking at the production graphs with @tomhughes the queues are drained completely on a regular basis, so this is not (currently) a real concern.

PR #4540, opened 2024-02-25
Allow users to delete their own diary entries

### LINKED ISSUE #4175: Suggestion: One should be able to delete their own diary entries
### Problem

Currently there's no way to delete a post/entry.

### Description

While I understand that what's have been published on the internet can never be truly deleted but that's not I'm suggesting here. Just delete the entry from the official OSM database and not care who archives/mirrors it anywhere.

### Screenshots

_No response_

tomhughes:
Fixes #4175

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -43,7 +43,7 @@ class Ability
         can :read, :dashboard
         can [:read, :update], [:preferences, :profile]
         can [:create, :subscribe, :unsubscribe], DiaryEntry
-        can :update, DiaryEntry, :user => user
+        can [:update, :hide, :unhide], DiaryEntry, :user => user
         can [:create], DiaryComment
         can [:show, :create, :destroy], Follow
         can [:read, :create, :destroy], Message
--- a/app/views/diary_entries/_diary_entry.html.erb
+++ b/app/views/diary_entries/_diary_entry.html.erb
@@ -43,7 +43,7 @@
         </li>
       <% end %>
 
-      <% if can? :hide, DiaryEntry %>
+      <% if can? :hide, diary_entry %>
         <li>
           <% if diary_entry.visible %>
             <%= link_to t(".hide_link"), hide_diary_entry_path(diary_entry.user, diary_entry), :method => :post, :data => { :confirm => t(".confirm") } %>
--- a/test/controllers/diary_entries_controller_test.rb
+++ b/test/controllers/diary_entries_controller_test.rb
@@ -784,24 +784,34 @@ class DiaryEntriesControllerTest < ActionDispatch::IntegrationTest
   end
 
   def test_hide
-    user = create(:user)
-    diary_entry = create(:diary_entry, :user => user)
+    author = create(:user)
+    diary_entry = create(:diary_entry, :user => author)
 
     # Try without logging in
-    post hide_diary_entry_path(user, diary_entry)
+    post hide_diary_entry_path(author, diary_entry)
     assert_response :forbidden
     assert DiaryEntry.find(diary_entry.id).visible
 
     # Now try as a normal user
-    session_for(user)
-    post hide_diary_entry_path(user, diary_entry)
+    session_for(create(:user))
+    post hide_diary_entry_path(author, diary_entry)
     assert_redirected_to :controller => :errors, :action => :forbidden
     assert DiaryEntry.find(diary_entry.id).visible
 
+    # Now try as the author
+    session_for(author)
+    post hide_diary_entry_path(:display_name => author.display_name, :id => diary_entry)
+    assert_response :redirect
+    assert_redirected_to :action => :index, :display_name => author.display_name
+    assert_not DiaryEntry.find(diary_entry.id).visible
+
+    # Reset
+    diary_entry.reload.update(:visible => true)
+
     # Now try as a moderator
     session_for(create(:moderator_user))
-    post hide_diary_entry_path(user, diary_entry)
-    assert_redirected_to :action => :index, :display_name => user.display_name
+    post hide_diary_entry_path(author, diary_entry)
+    assert_redirected_to :action => :index, :display_name => author.display_name
     assert_not DiaryEntry.find(diary_entry.id).visible
 
     # Reset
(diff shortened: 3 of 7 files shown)


app/abilities/ability.rb
```diff
@@ -43,7 +43,7 @@ def initialize(user)
         can [:new, :show, :create, :destroy], :oauth2_authorization
         can [:edit, :update, :destroy], :account
         can [:show], :dashboard
-        can [:new, :create, :edit, :update, :comment, :subscribe, :unsubscribe], DiaryEntry
+        can [:new, :create, :edit, :update, :comment, :subscribe, :unsubscribe, :hide, :unhide], DiaryEntry
```
AntonKhorev, line 46: Is there a way to check the user here like this?

    can [:hide, :unhide], DiaryEntry, :user => user

AntonKhorev: For example, you can put this in `hide` method:

    authorize! :hide, entry

In theory there's [`load_and_authorize_resource`](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/controller_helpers.md#authorize_resource-load_resource-load_and_authorize_resource) helper.

tomhughes: Sure there is a whole discussion at https://github.com/CanCanCommunity/cancancan/blob/develop/docs/fetching_records.md which suggests it can work but I haven't managed to make any sense of it yet and we're not using it anywhere else so I think we should be consistent here and look at that separately?

gravitystorm: I agree with both of you - there's features of CanCanCan that we can use to simplify this, but it's also something that we can tackle separately. Anton's example is correct, we can build permissions within the ability file like

```
if user
  can [:hide], DiaryEntry, :user => user # can only hide their own entries
 ....
if user.moderator?
  can [:hide], DiaryEntry # can hide all entries
```

We can then go back to using the `can :hide, diary_entry` instead of replacing that with inline permission checks like `if current_user && (current_user == diary_entry.user || current_user.moderator?)` which is logic that should be in the abilities system.

I'm happy to have this as a followup task though, particularly since merging this first will make the followup easier due to the additional tests.

config/locales/en.yml
```diff
@@ -554,8 +554,8 @@ en:
         other: "%{count} comments"
       no_comments: No comments
       edit_link: Edit this entry
-      hide_link: Hide this entry
-      unhide_link: Unhide this entry
+      delete_link: Delete this entry
+      restore_link: Restore this entry
```
gravitystorm, line 558: Is it worth sticking with the original terminology? I think this is the first case where normal users will see a "hide" button (they already see plenty of delete buttons, e.g. traces, messages), but in general the hide/unhide pair is used for reversible actions and delete for actions that can't be undone (even if the delete is really only a soft delete).

I don't want moderators to start hesitating that this might not be the same as the rest of the soft deletes.

gravitystorm (changes requested):
Happy in principle, just a few minor things to consider here.

app/views/diary_entries/_diary_entry.html.erb
```diff
@@ -33,12 +33,12 @@
         </li>
       <% end %>
 
-      <% if can? :hide, DiaryEntry %>
+      <% if current_user && (current_user == diary_entry.user || current_user.moderator?) %>
```
gravitystorm, line 36: perhaps also `|| current_user.administrator?` here?

tomhughes: Sure, added now.

app/abilities/ability.rb
tomhughes: I've rebased this and applied this change to let CanCanCan control which entries can be hidden.

AntonKhorev:
The ability to delete diary posts is mentioned in the privacy policy https://osmfoundation.org/wiki/Privacy_Policy#How_can_you_control_the_processing_of_your_data_and_reduce_privacy_related_issues , as if you can already do that. It's somewhat true because you can edit the posts, and therefore delete all the text.

If we do true hiding, the question what supposed to happen with comments to hidden posts remains. Maybe we can do *hide completely if there are no comments by other users that are not hidden, otherwise show only comments*.

config/locales/en.yml
hlfan: And having a "delete" link link to a "hide" path is less confusing?

app/abilities/ability.rb
```diff
@@ -43,7 +43,7 @@ def initialize(user)
         can :read, :dashboard
         can [:read, :update], [:preferences, :profile]
         can [:create, :subscribe, :unsubscribe], DiaryEntry
-        can :update, DiaryEntry, :user => user
+        can [:update, :hide, :unhide], DiaryEntry, :user => user
```
pablobm, line 46: I'm not familiar with CanCanCan and this declaration is confusing me.

I played with it a bit. Do I understand correctly that the `:user => user` part is only used when doing manual checks (`can?(:hide, entry)`) but not automatically with `authorize_resource`?

config/locales/en.yml
pablobm: I agree that "hide" and "delete" should have clearly defined meanings.

Perhaps there's another word we can use here, like "unpublish". Doesn't sound as final as "delete". Thoughts?

test/controllers/diary_entries_controller_test.rb
```diff
@@ -793,11 +793,21 @@ def test_hide
...
 
     # Now try as a normal user
-    session_for(user)
+    session_for(create(:user))
     post hide_diary_entry_path(user, diary_entry)
     assert_redirected_to :controller => :errors, :action => :forbidden
     assert DiaryEntry.find(diary_entry.id).visible
 
+    # Now try as the author
+    session_for(user)
```
pablobm: I think it's worth renaming this variable as `author` to reduce confusion. Same under `unhide`.

pablobm (commented):
> If we do true hiding, the question what supposed to happen with comments to hidden posts remains. Maybe we can do hide completely if there are no comments by other users that are not hidden, otherwise show only comments.

On this question: is there any reason to not hide the comments? The current implementation, where entry+comments are hidden as a unit, seems good to me.

hlfan:
What about we leave the changes to the partial and the translation to another PR? The rest doesn't seem blocking to me.

app/abilities/ability.rb
tomhughes: No I think that's always applied, otherwise anybody would be able able to update other people's diary entries at the moment?

test/controllers/diary_entries_controller_test.rb
tomhughes: I've made that change now.

app/abilities/ability.rb
tomhughes: Ah no we do a manual check in the controller methods so yes I think you're right, because it needs the diary entry object  to get the user from for the check.

tomhughes:
I've dropped the change from Hide/Unhide to Delete/Restore now.

hlfan (approved):
This looks good to me now, thank you!

PR #4550, opened 2024-02-28
Drop user tokens table

tomhughes:
This drops the user tokens table - draft for now until all the tokens have aged out and we are ready to drop the table


tomhughes:
The last tokens have now expired so this is now ready for review.

gravitystorm:
I'm happy with this (and one bit in particular :wink:) so it's good to merge.

However, there's one little wrinkle for deploying to production, and I'm not sure whether we need to be concerned about that. In most of our PRs we have migrations and code that depends on the new db structure (and that would fail on a missing new table or attribute etc) in the same commit. So the implication is that deployments should (in general) run the migrations first, then deploy the new code afterwards, to avoid errors. But with this PR, you would need to deploy the new code first, then run the migrations second.

If we were using point releases, then maybe we would have the code changes in one release, and the dropping of the unused tables in a subsequent release. But we don't do release versions, and I think that you @tomhughes run the migrations and code deploys manually, so perhaps the exact sequencing isn't super important. It's certainly not yet documented anywhere.

If anyone else has thoughts on this for future, please let me know!

PR #4552, opened 2024-02-29
Move changeset show action to changesets controller

AntonKhorev:
Moves all changeset-related code out of browse controller.

This is already true on the javascript side with [history.js](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/javascripts/index/history.js) and [changeset.js](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/javascripts/index/changeset.js) separate from [browse.js](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/javascripts/index/browse.js).


test/controllers/site_controller_test.rb
```diff
@@ -147,7 +147,7 @@ def test_permalink
 
     get permalink_path(:code => "wBz3--", :changeset => 4)
     assert_response :redirect
-    assert_redirected_to :controller => :browse, :action => :changeset, :id => 4, :anchor => "map=3/4.8779296875/3.955078125"
+    assert_redirected_to :controller => :changesets, :action => :show, :id => 4, :anchor => "map=3/4.8779296875/3.955078125"
```
tomhughes, line 150: This can use `changeset_path` now I think?

test/controllers/changesets_controller_test.rb
```diff
@@ -121,8 +125,9 @@ def test_index_bbox
   # Checks the display of the user changesets listing
   def test_index_user
     user = create(:user)
-    create(:changeset, :user => user)
-    create(:changeset, :closed, :user => user)
+    create(:changeset, :user => user, :num_changes => 1)
+    create(:changeset, :closed, :user => user, :num_changes => 1)
```
tomhughes, line 129: Why does this need changes when neither the `index` action or the rest of this test has changed?

test/controllers/changesets_controller_test.rb
```diff
@@ -121,8 +125,9 @@ def test_index_bbox
...
+    user.reload
```
tomhughes, line 130: Why do we need to reload the user? Is that to cause the cached changeset count on the user to update or something?

tomhughes (commented):
Nice work, just a few small queries...

test/controllers/changesets_controller_test.rb
AntonKhorev: This test was comparing empty list of changesets (because changesets without changes are filtered out) with another empty list of changesets (because `user.changesets` weren't updated after `create(:changeset, ...)`). So it wasn't testing anything and it wasn't testing my changes.

test/controllers/changesets_controller_test.rb
AntonKhorev: see below

test/controllers/site_controller_test.rb
AntonKhorev: Updated. This change could have been done before too. Same for some other `assert_redirected_to`.

PR #4554, opened 2024-03-02
Remove assert_response when followed by assert_redirected_to

AntonKhorev:
`assert_redirected_to` does `assert_response` anyway:

https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/actionpack/lib/action_dispatch/testing/assertions/response.rb#L61


tomhughes:
Ha yes, I've always thought is was a bit silly doing the `assert_response` when were already checking the redirect.

PR #4562, opened 2024-03-06
Add Turbo to replace custom JS

grekko:
## Context

This PR is part of a conversation with @gravitystorm and @tordans on the recent OSM Hackweekend in Karlsruhe and following conversations. The goal is to show how to integrate Rails Trubo to the project and what a migration path would look like.

Navigating the OSM website should be smooth. [Rails UJS](https://github.com/rails/rails/tree/main/actionview/app/assets/javascripts) and custom JavaScript enables users to manage their Messages in a SPA-like fashion without adding too much complexity to the codebase.

[Turbo](https://turbo.hotwired.dev) is the successor of Rails UJS. It provides SPA-like behaviour to Rails applications similar to Rails UJS but works a bit differently and allows to write even less JavaScript.

With the latest changes to [Turbo](https://turbo.hotwired.dev) (see [v8.0.0](https://github.com/hotwired/turbo/releases/tag/v8.0.0) page refreshes with morphing) it has gotten even easier to manage front-end state without custom JavaScript.

## What has changed?

- Adds the [`turbo-rails`](https://github.com/hotwired/turbo-rails)-gem which includes the javascript library and ruby helper code to integrate Turbo Drive, Frames and Streams
- Replaces the JSON payloads + Rails UJS solution with Turbo Drive + DOM Morphing to manage partial front end updates for `Messages#{mark, destroy}` controller actions

<details>
  <summary>
  The UI/UX is mostly identical with 2 exceptions:
<ul>
<li>Flash Messages are rendered.</li>
<li>The Messages front end state not just partially but fully updated.</li>
</ul>
</summary>

### UI / UX Changes

**TBD**: The flash messages are now displayed (they were not displayed using UJS + custom JS). While the flash message is a bit disruptive, I am not sure if I want to suppress/hide them or keep them.

The main difference between the Rails UJS and the Turbo version is that not just dedicated sections of the front-end state are managed, but the full page is managed.

See e.g. what happens when all messages are deleted.

**Before**

delete-last-before

**After**

delete-last-after

### Demo

Besides the now-appearing flash messages the UX for managing messages should be identical.

- The Message Counters are all updated
- Deleted messages are faded out
- The scroll position is preserved

Deleting messages

</details>

## Dev Notes

- Turbo and Rails UJS can co-exist
  - Turbo (Drive) is therefore setup so that its behaviour must be enabled via opt-in data-attributes
- Turbo relies on the concept of `HTML over the Wire` (see <https://hotwired.dev>). If there is still the need to send/receive JSON payloads, Rails UJS might be the right tool

## Next steps

I'd like to get feedback on the state of the PR and would like to know if other developers are happy with the outcome or if they foresee any issues with gradually replacing Rails UJS with Turbo.

Gradually replacing Rails UJS and integrating Turbo into other places of the OSM website application could be a good next EWG project.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -69,6 +69,10 @@ window.updateLinks = function (loc, zoom, layers, object) {
 };
 
 $(document).ready(function () {
+  // NB: Turns Turbo Drive off by default. Turbo Drive must be opt-in on a per-link and per-form basis
+  // See https://turbo.hotwired.dev/reference/drive#turbo.session.drive
+  Turbo.session.drive = false;
+
   var headerWidth = 0,
       compactWidth = 0;
 
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -112,6 +112,7 @@ $(document).ready(function () {
     updateHeader();
 
     $(window).resize(updateHeader);
+    $(document).on("turbo:render", updateHeader);
   }, 0);
 
   $("#menu-icon").on("click", function (e) {
--- a/app/assets/javascripts/messages.js
+++ b/app/assets/javascripts/messages.js
@@ -1,36 +1,16 @@
 $(document).ready(function () {
-  $(".inbox-mark-unread").on("ajax:success", function (event, data) {
-    updateHtml(data);
-    updateReadState(this, false);
+  $(".messages-table .destroy-message").on("turbo:submit-end", function (event) {
+    if (event.detail.success) {
+      event.target.dataset.isDestroyed = true;
+    }
   });
 
-  $(".inbox-mark-read").on("ajax:success", function (event, data) {
-    updateHtml(data);
-    updateReadState(this, true);
+  $(".messages-table .message-summary").on("turbo:before-morph-element", function (event) {
+    if ($(event.target).find("[data-is-destroyed]").length > 0) {
+      event.preventDefault(); // NB: prevent Turbo from morhping/removing this element
+      $(event.target).fadeOut(800, "linear", function () {
+        $(this).remove();
+      });
+    }
   });
-
-  $(".inbox-destroy").on("ajax:success", function (event, data) {
-    updateHtml(data);
-
-    $(this).closest("tr").fadeOut(800, "linear", function () {
-      $(this).remove();
-    });
-  });
-
-  function updateHtml(data) {
-    $("#inboxanchor").remove();
-    $(".user-button").before(data.inboxanchor);
-
-    $("#inbox-count").replaceWith(data.inbox_count);
-    $("#outbox-count").replaceWith(data.outbox_count);
-    $("#muted-count").replaceWith(data.muted_count);
-  }
-
-  function updateReadState(target, isRead) {
-    $(target).closest("tr")
-      .toggleClass("inbox-row", isRead)
-      .toggleClass("inbox-row-unread", !isRead)
-      .find(".inbox-mark-unread").prop("hidden", !isRead).end()
-      .find(".inbox-mark-read").prop("hidden", isRead);
-  }
 });
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -2,6 +2,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge" />
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <%= javascript_include_tag "es6" unless browser.es6? %>
+  <%= javascript_include_tag "turbo", :type => "module" %>
   <%= javascript_include_tag "application" %>
   <%= javascript_include_tag "i18n/#{I18n.locale}" %>
   <%= stylesheet_link_tag "screen-#{dir}", :media => "screen" %>
--- a/app/views/messages/_message_summary.html.erb
+++ b/app/views/messages/_message_summary.html.erb
@@ -1,13 +1,13 @@
-<tr id="inbox-<%= message.id %>" class="inbox-row<%= "-unread" unless message.message_read? %>">
-  <td><%= link_to message.sender.display_name, message.sender %></td>
-  <td><%= link_to message.title, message %></td>
+<tr id="inbox-<%= message.id %>" class="message-summary inbox-row<%= "-unread" unless message.message_read? %>">
+  <td><%= link_to message.sender.display_name, user_path(message.sender) %></td>
+  <td><%= link_to message.title, message_path(message) %></td>
   <td class="text-nowrap"><%= l message.sent_on, :format => :friendly %></td>
   <td class="text-nowrap d-flex justify-content-end gap-1">
-    <%= button_to t(".unread_button"), message_mark_path(message, :mark => "unread"), :remote => true, :class => "btn btn-sm btn-primary", :form => { :class => "inbox-mark-unread", :hidden => !message.message_read? } %>
-    <%= button_to t(".read_button"), message_mark_path(message, :mark => "read"), :remote => true, :class => "btn btn-sm btn-primary", :form => { :class => "inbox-mark-read", :hidden => message.message_read? } %>
-    <%= button_to t(".destroy_button"), message_path(message, :referer => request.fullpath), :method => :delete, :remote => true, :class => "btn btn-sm btn-danger", :form_class => "inbox-destroy" %>
+    <%= button_to t(".unread_button"), message_mark_path(message, :mark => "unread"), :class => "btn btn-sm btn-primary", :form => { :"data-turbo" => true, :class => "inbox-mark-unread", :hidden => !message.message_read? } %>
+    <%= button_to t(".read_button"), message_mark_path(message, :mark => "read"), :class => "btn btn-sm btn-primary", :form => { :"data-turbo" => true, :class => "inbox-mark-read", :hidden => message.message_read? } %>
+    <%= button_to t(".destroy_button"), message_path(message, :referer => request.fullpath), :method => :delete, :class => "btn btn-sm btn-danger", :form => { :"data-turbo" => true, :class => "destroy-message" } %>
     <% if message.muted? %>
-      <%= button_to t(".unmute_button"), message_unmute_path(message), :method => :patch, :class => "btn btn-sm btn-secondary" %>
+      <%= button_to t(".unmute_button"), message_unmute_path(message), :method => :patch, :class => "btn btn-sm btn-secondary", :form => { :"data-turbo" => true } %>
     <% end %>
   </td>
 </tr>
--- a/app/views/messages/_message_summary.html.erb
+++ b/app/views/messages/_message_summary.html.erb
@@ -3,11 +3,11 @@
   <td><%= link_to message.title, message_path(message) %></td>
   <td class="text-nowrap"><%= l message.sent_on, :format => :friendly %></td>
   <td class="text-nowrap d-flex justify-content-end gap-1">
-    <%= button_to t(".unread_button"), message_mark_path(message, :mark => "unread"), :class => "btn btn-sm btn-primary", :form => { :"data-turbo" => true, :class => "inbox-mark-unread", :hidden => !message.message_read? } %>
-    <%= button_to t(".read_button"), message_mark_path(message, :mark => "read"), :class => "btn btn-sm btn-primary", :form => { :"data-turbo" => true, :class => "inbox-mark-read", :hidden => message.message_read? } %>
-    <%= button_to t(".destroy_button"), message_path(message, :referer => request.fullpath), :method => :delete, :class => "btn btn-sm btn-danger", :form => { :"data-turbo" => true, :class => "destroy-message" } %>
+    <%= button_to t(".unread_button"), message_mark_path(message, :mark => "unread"), :class => "btn btn-sm btn-primary", :form => { :data => { :turbo => true }, :class => "inbox-mark-unread", :hidden => !message.message_read? } %>
+    <%= button_to t(".read_button"), message_mark_path(message, :mark => "read"), :class => "btn btn-sm btn-primary", :form => { :data => { :turbo => true }, :class => "inbox-mark-read", :hidden => message.message_read? } %>
+    <%= button_to t(".destroy_button"), message_path(message, :referer => request.fullpath), :method => :delete, :class => "btn btn-sm btn-danger", :form => { :data => { :turbo => true }, :class => "destroy-message" } %>
     <% if message.muted? %>
-      <%= button_to t(".unmute_button"), message_unmute_path(message), :method => :patch, :class => "btn btn-sm btn-secondary", :form => { :"data-turbo" => true } %>
+      <%= button_to t(".unmute_button"), message_unmute_path(message), :method => :patch, :class => "btn btn-sm btn-secondary", :form => { :data => { :turbo => true } } %>
     <% end %>
   </td>
 </tr>
--- a/app/views/messages/_messages_table.html.erb
+++ b/app/views/messages/_messages_table.html.erb
@@ -1,4 +1,4 @@
-<table class="table table-sm align-middle">
+<table class="table table-sm align-middle messages-table">
   <thead>
     <tr>
       <% columns.each do |column| %>
--- a/config/eslint.json
+++ b/config/eslint.json
@@ -13,7 +13,8 @@
     "OSM": "writable",
     "Matomo": "readonly",
     "Qs": "readonly",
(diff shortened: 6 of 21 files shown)


app/assets/javascripts/messages.js
```diff
@@ -1,36 +1,17 @@
 $(document).ready(function () {
-  $(".inbox-mark-unread").on("ajax:success", function (event, data) {
-    updateHtml(data);
-    updateReadState(this, false);
+  $(".messages-table .inbox-destroy").on("turbo:submit-end", function (event) {
```
grekko, line 2: There is no need for jQuery here. I just kept it for consistency but would love to remove it here.

app/assets/javascripts/messages.js
```diff
@@ -1,36 +1,17 @@
...
+    if (event.detail.success) {
+      // NB: Mark element to be removed (faded out) via animation
+      event.target.dataset.withDeleteAnimation = true;
+    }
   });
 
-  $(".inbox-mark-read").on("ajax:success", function (event, data) {
-    updateHtml(data);
-    updateReadState(this, true);
+  $(".messages-table .message-summary").on("turbo:before-morph-element", function (event) {
```
grekko, line 9: AFAIU the `turbo:before-morph-element`-event is only fired for DOM elements with an `id`. Therefore I had to set the `id`-attribute in `app/views/messages/_sent_message_summary.html.erb` which is non-obvious here. 

I tried to dig into the responsible underlying Idiomorph-lib [code](https://github.com/bigskysoftware/idiomorph/blob/8e40c42cc573609eb6863e72fa3403574974dd7d/src/idiomorph.js#L122-L155) which is used by Turbo (see https://github.com/hotwired/turbo/blob/8c668612b3715cdae0b291abd2ce61baf5813c71/src/core/drive/morph_renderer.js#L28-L41) to figure out if there is a more obvious configuration e.g. a `data`-attribute to subscribe to such events, but could not find anything useful yet.


It is a bit unfortunate that the `id` is set just to subscribe for the `turbo:before-morph-element`-event. We could rename the `id` so reflect this since we do not care about the actual value here, e.g. to `id="turbo-morph-element-#{message_summary.id}"`.

app/views/messages/_messages_table.html.erb
```diff
@@ -1,4 +1,4 @@
-<table class="table table-sm align-middle">
+<table class="table table-sm align-middle messages-table">
```
grekko, line 1: Personally I'd prefer to use a data-attribute or `js-`-prefixed CSS selector when defining JS-hooks, but I wanted to keep this consistent with other code I read.

app/assets/javascripts/messages.js
tomhughes: See I have a visceral hatred of doing stuff with raw DOM methods and much prefer jQuery so I suspect that is a separate discussion the project is going to need to have...

tomhughes:
So I'm not familiar with turbo at all beyond the quick look I've just had over lunch but in principle it looks nice and I don't have any objections - definitely not if it eventually lets us get rid of the client side router that nobody understands.

I do think you probably chose a bad example to get started with... I can see why you chose it, because it's somewhere that is already manually doing partial updates but in order to preserve that behaviour you've wound up with a lot of really ugly stuff to handle the transitions :disappointed:

Possibly eventually that's something [view transitions](https://turbo.hotwired.dev/handbook/drive#view-transitions) could handle better? They're not supported in firefox yet though...

I suspect it would have been compelling to demonstrate what's possible with something like the diary entries or trace lists where I think you could pretty much just wrap the content in a turbo frame and get instant partial updates with no mess.

One thing I'm unclear on is why we're installing turbo with npm when the `turbo-rails` gem already includes a copy which seems to be the [recommended](https://turbo.hotwired.dev/handbook/installing) way to get it when using rails?

gravitystorm:
> I do think you probably chose a bad example to get started with...

That's my fault then - I suggested this, along with DiaryComments, as the two places I would investigate first. I hadn't considered anything in traces.

tomhughes:
> That's my fault then - I suggested this, along with DiaryComments, as the two places I would investigate first. I hadn't considered anything in traces.

Yes it's probably not obvious until you start working on it then the transition for deletion will complicate matters so I can understand the logic.

grekko:
@tomhughes thanks for the quick and helpful feedback.

> a lot of really ugly stuff to handle the transitions 😞

yes, true. I'd like to keep this extra stuff tightly packed together, so we can at least have it not become a remnant at some point in the future. I'll take tomorrow some time to tinker with this a bit.

> Possibly eventually that's something [view transitions](https://turbo.hotwired.dev/handbook/drive#view-transitions) could handle better?

👍🏽

> One thing I'm unclear on is why we're installing turbo with npm when the turbo-rails gem already includes a copy which seems to be the [recommended](https://turbo.hotwired.dev/handbook/installing) way to get it when using rails?

yes. good point. I started initially with just the npm-package because the turbo-rails gem brings a lot of helper code that was not even used in the first draft of the PR and everything works great without the gem. This PR just uses the `meta`-tag helper code from the gem.

Then I figured it would be better to have the gem included since it provides access to even more helper-methods e.g. for Turbo Frames and just forgot to drop the npm-package.

config/eslint.json
```diff
@@ -13,7 +13,8 @@
     "OSM": "writable",
     "Matomo": "readonly",
     "Qs": "readonly",
-    "updateLinks": "readonly"
+    "updateLinks": "readonly",
+    "Turbo": "writable"
```
AntonKhorev, line 17: "readonly" works fine here, you're not going to rewrite `Turbo` anywhere in js

tomhughes: He is writing to `Turbo.session.drive` to make turbo opt-in rather than opt-out.

AntonKhorev: He's not rewriting `window.Turbo`. That's what eslint checks.

grekko: > That's what eslint checks.

I'll look at this and will adjust the config if `readonly` suffices 👍🏽

grekko: I think it does. Locally `eslint` did not complain with `readonly`.

app/assets/javascripts/messages.js
grekko: Maybe it's ok to rely on the `id` being present and not having a stronger indication what the purpose of the `id` is. I'll close this conversation.

gravitystorm:
@grekko Sorry, there's now a merge conflict on this PR. Could you update it please?

Otherwise, I'm happy to merge this.

gravitystorm (approved):

app/assets/javascripts/application.js
```diff
@@ -108,6 +112,7 @@ $(document).ready(function () {
     updateHeader();
 
     $(window).resize(updateHeader);
+    document.addEventListener("turbo:render", updateHeader);
```
tomhughes, line 115: So obviously this comes down to the whether-to-use-jquery question which was briefly mentioned before, but obviously this is using a raw API for something that we have generally done with jquery in the past.

app/assets/javascripts/messages.js
```diff
@@ -1,36 +1,16 @@
...
+    if (event.detail.success) {
+      event.target.dataset.isDestroyed = true;
+    }
   });
 
-  $(".inbox-mark-read").on("ajax:success", function (event, data) {
-    updateHtml(data);
-    updateReadState(this, true);
+  $(".messages-table .message-summary").on("turbo:before-morph-element", function (event) {
+    if (event.target.querySelector("[data-is-destroyed]")) {
```
tomhughes, line 9: Again this is a raw API where confusingly two lines later it uses jquery instead...

app/views/layouts/_head.html.erb
```diff
@@ -2,6 +2,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge" />
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <%= javascript_include_tag "es6" unless browser.es6? %>
+  <%= javascript_include_tag "turbo", :type => "module" %>
```
tomhughes, line 5: What was the thinking behind making this a separate include rather than adding it to the application bundle? or is it necessary in order to include it as a module? I confess to not understanding javascript modules and how they interact with other sorts of javascript and rails assets very well...

app/views/messages/_message_summary.html.erb
```diff
@@ -1,13 +1,13 @@
...
   <td><%= link_to message_summary.title, message_path(message_summary) %></td>
   <td class="text-nowrap"><%= l message_summary.sent_on, :format => :friendly %></td>
   <td class="text-nowrap d-flex justify-content-end gap-1">
-    <%= button_to t(".unread_button"), message_mark_path(message_summary, :mark => "unread"), :remote => true, :class => "btn btn-sm btn-primary", :form => { :class => "inbox-mark-unread", :hidden => !message_summary.message_read? } %>
-    <%= button_to t(".read_button"), message_mark_path(message_summary, :mark => "read"), :remote => true, :class => "btn btn-sm btn-primary", :form => { :class => "inbox-mark-read", :hidden => message_summary.message_read? } %>
-    <%= button_to t(".destroy_button"), message_path(message_summary, :referer => request.fullpath), :method => :delete, :remote => true, :class => "btn btn-sm btn-danger", :form_class => "inbox-destroy" %>
+    <%= button_to t(".unread_button"), message_mark_path(message_summary, :mark => "unread"), :class => "btn btn-sm btn-primary", :form => { :"data-turbo" => true, :class => "inbox-mark-unread", :hidden => !message_summary.message_read? } %>
```
tomhughes, line 6: I'm not a fan of `:"..."` symbols I have to say... Can this be done as `:data => { :turbo => true }` which I think is how we've normally done data, or does that not work in this context?

grekko: 🤔 I think in this context it should work as long as the `:data => { :turbo => true }`-structure is passed into the `:form => …`-key.

This would be:

```
:form => { :data => { :turbo => true }, :class => "inbox-mark-unread", …
```

I deliberately wanted to avoid more nesting in the hash structure, but I'd prefer to keep it consistent with the pre-existing style in the project. I'll have a look for other places in the partials and will report back.

grekko: see 2a2a317

grekko: In other templates similar attributes are nested like `:data => { :foo => "bar" }`, e.g. here:

- https://github.com/openstreetmap/openstreetmap-website/blob/8da80ff471f7d04984ff5abab0c5fe32920330ba/app/views/layouts/_search.html.erb#L24
- https://github.com/openstreetmap/openstreetmap-website/blob/8da80ff471f7d04984ff5abab0c5fe32920330ba/app/views/traces/show.html.erb#L66
… and more places.

So for consistency reasons I agree to nest the data-attributes here, too. See 2a2a317

app/assets/javascripts/application.js
grekko: For consistency I dropped jQuery in there.

app/assets/javascripts/messages.js
grekko: For consistency I dropped jQuery in there.

app/views/layouts/_head.html.erb
grekko: This is the only way I was able to import Turbo (which comes bundled in ESM format in the `turbo-rails`-gem).

Trying to import it inside the `application.js` with Sprockets raises errors:

```diff
diff --git a/app/assets/javascripts/application.js b/app/assets/javascripts/application.js
index c71c0652d..25ca33d34 100644
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -18,6 +18,7 @@
 //= require qs/dist/qs
 //= require bs-custom-file-input
 //= require bs-custom-file-input-init
+//= require turbo
 
 /*
  * Called as the user scrolls/zooms around to manipulate hrefs of the
diff --git a/app/views/layouts/_head.html.erb b/app/views/layouts/_head.html.erb
index 3c691612a..0ee5bf9c0 100644
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -2,7 +2,6 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge" />
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <%= javascript_include_tag "es6" unless browser.es6? %>
-  <%= javascript_include_tag "turbo", :type => "module" %>
   <%= javascript_include_tag "application" %>
   <%= javascript_include_tag "i18n/#{I18n.locale}" %>
   <%= stylesheet_link_tag "screen-#{dir}", :media => "screen" %>
```

SCR-20240320-rxmw

AntonKhorev: Since modules are deferred, you had to defer the init code:

https://github.com/openstreetmap/openstreetmap-website/pull/4562/files#diff-0e145ba000e96b1a45d230275e350e4e8514efb4db7dbcfa228fe1003896c9b4R72-R74

Is there a way for Sprockets to translate a module into non-module?

tomhughes: I've been looking and I can't see any obvious solution so it's probably not worth spending any more time on it when we have something that works - the sprockets asset pipeline is kind of deprecated now anyway.

There is a separate problem that I have discovered though, which is that turbo is trying to inject CSS for the progress bar and failing as our CSP rules prohibit inline CSS rules. It has code to work with rails builting security policy support but we're still using secure_headers so https://github.com/openstreetmap/openstreetmap-website/blob/8da80ff471f7d04984ff5abab0c5fe32920330ba/app/views/layouts/_head.html.erb#L14 needs to be removed and replaced with this:

```
<meta name="csp-nonce" content="<%= content_security_policy_style_nonce %>" />
```

which sets the CSP nonce for turbo using secure_headers instead.

grekko: > […] which is that turbo is trying to inject CSS for the progress bar and failing as our CSP rules prohibit inline CSS styles

I haven't noticed that. Did this come up as a warning/error in the JS console or how did you notice?

tomhughes: Yes you should get a warning in the console.

tomhughes: Ah you'll need to have `csp_enforce:true` in `config/settings.local.yml` as well.

grekko: ah! Now I can see it.

SCR-20240321-hvcz

some questions :)

I see that the file is tracked in git. Is that `settings.yml` loaded in Production and `enforce_csp: false`?

---

I'll have a closer look at the error. My CSP knowledge is still limited. I'll have to do a bit of homework first.

tomhughes: Sorry yes that file is for local overrides of the defaults. There's a live version with my [fix](https://github.com/tomhughes/openstreetmap-website/commit/1b9d82d66aba56453ed75070637945967a30e4d5) running at https://tomh.apis.dev.openstreetmap.org/.

grekko: > which sets the CSP nonce for turbo using secure_headers instead.

ok. I had to read up a bit on the `secure_headers`-gem CSP implementation and CSP generally.

I understand the the `csp_meta_tag` currently does nothing since its provided by Rails, which builds on Rails CSP configuration (currently nothing is configured in `config/initializers/content_security_policy.rb`).

`content_security_policy_style_nonce` is provided by `secure_headers` (here https://github.com/github/secure_headers/blob/7a23cb6b350b024a786e163e81c902552b9c484f/lib/secure_headers/view_helper.rb#L83-L85) and therefore the correct way to safelist certain inline styles.

I'll adjust the PR to reflect this. Thanks @tomhughes for pointing this out and please correct anything or add anything you think one should know.

As I've mentioned my understanding of CSP is still limited and I have not practical experiences yet.

grekko: Actually 🤔 thinking about it for another second: The change of replacing `csp_meta_tag` with `<meta name="csp-nonce" content="<%= content_security_policy_style_nonce %>" />` is only indirectly related to this PR, no?

So to keep the PR and its changeset smaller I'd like to move this change into a separate PR. Does that sounds right? @tomhughes

tomhughes: Well we haven't used nonces before so it hasn't been an issue before now but sure that can be a separate pull request. I think I have a branch somewhere where I started on moving to the rails builtin policy stuff so I should probably resurrect that at some point...

grekko: https://github.com/openstreetmap/openstreetmap-website/pull/4606 addressed the CSP violation

tomhughes:
Thanks for making those changes. I think this is all good now.

Issue #4561, opened 2024-03-05
Changeset comment text is lost when switching entity pages


dreua:
### URL

https://www.openstreetmap.org/changeset/148222445?way_page=1

### How to reproduce the issue?

1. You need a changeset with multiple pages of changed enitties
2. Write some text in the comment box
3. Click on one of the numbers in order to see more of the changed entities
4. Your comment is gone forever

### Screenshot(s) or anything else?

This is *very* bad UX. I am very grateful to everyone who built openstreetmap.org as well as iD but I'll be even more grateful when this bug is fixed and I won't loose a comment ever again.

HolgerJeromin:
Solution: Add conditionally a `beforeunload` handler. But beware of this:

> Important: the beforeunload event should only be used to alert the user of unsaved changes. Once those changes are saved, the event should be removed. It should never be added unconditionally to the page, as doing so can hurt performance in some cases. 

https://developer.chrome.com/docs/web-platform/page-lifecycle-api#the_beforeunload_event

dreua:
This was Firefox on Linux btw, haven't tried other browsers.

dreua:
Ideally I'd say there isn't a reload at all, just load the data via JS and leave the text field alone? But a warning would be definitely better than nothing

danieldegroot2:
Related #2088(, #4028, #3928, #3303, #933, #1089(, #647, #4039), #2973, #1212, #1333 (not reproduced));

_[How you can help with reviewing pull requests](#3815)_

AntonKhorev:
How is https://github.com/openstreetmap/openstreetmap-website/issues/933 related to this? Unless you think that anything about the sidebar is related.

danieldegroot2:
@AntonKhorev Reduced to relevant issues only.

PR #4563, opened 2024-03-06
Reload only element type sections when switching between changeset element pages

AntonKhorev:
Fixes #4561.

image
image

PR #4565, opened 2024-03-07
Reload only element type sections with Turbo frames

AntonKhorev:
Fixes https://github.com/openstreetmap/openstreetmap-website/issues/4561. Updated version without doing #4563 first. I don't do the entire sidebar with Turbo. The only change to javascript here is to suppress our router.js when a pagination turbo link is clicked. I have to mark those links explicitly for now with `data-turbo="true"`.

image
image


tomhughes (approved):

dreua:
Thank you :heart:

PR #4566, opened 2024-03-07
Enforce Not Null Constraints on Redaction Title and Description

IshmeetSingh06:
Fixes: https://github.com/openstreetmap/openstreetmap-website/issues/2033

### DIFF
--- a/app/models/redaction.rb
+++ b/app/models/redaction.rb
@@ -3,8 +3,8 @@
 # Table name: redactions
 #
 #  id                 :integer          not null, primary key
-#  title              :string
-#  description        :text
+#  title              :string           not null
+#  description        :text             not null
 #  created_at         :datetime
 #  updated_at         :datetime
 #  user_id            :bigint(8)        not null
--- a/db/migrate/20240307180830_add_check_constraint_to_redaction_title_and_description.rb
+++ b/db/migrate/20240307180830_add_check_constraint_to_redaction_title_and_description.rb
@@ -1,0 +1,24 @@
+class AddCheckConstraintToRedactionTitleAndDescription < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def up
+    Redaction.where(:title => nil).find_in_batches(:batch_size => 1000) do |redactions|
+      redactions.each do |r|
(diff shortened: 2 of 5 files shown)


tomhughes:
Just enforcing a database constraint is not very user friendly - there should be validations in the model as well.

IshmeetSingh06:
> there should be validations in the model as well.



> Just enforcing a database constraint is not very user friendly - there should be validations in the mod

Hi @tomhughes validations are already present on the model

tomhughes:
So how can #2033 happen then? How are the validations being bypassed?

IshmeetSingh06:
> So how can #2033 happen then? How are the validations being bypassed?

Validations were added on November 6th 2018, while the issue was raised on October, I've checked and confirmed that even without the NOT NULL constraint we wont be able to create a redaction with empty title or description unless someone tries to create one through console, Since this is an very old issue (approx 5yrs) we can avoid adding the NOT NULL constraints (because Rails will handle that for us) and just update the test cases only.

tomhughes:
Well if there are legacy records with null values then we may not be able to add constraints (without changing those records) but I will have to check that.

tomhughes:
I take it all the complicated adding of check constraints and then adding null constraints and dropping the check constraints is something strong migrations insisted on?

Also you need to add the `structure.sql` changes when you add a migration as the test failure is showing.

IshmeetSingh06:
> I take it all the complicated adding of check constraints and then adding null constraints and dropping the check constraints is something strong migrations insisted on?
Yes
> Also you need to add the `structure.sql` changes when you add a migration as the test failure is showing.
Sure will do, should I remove the block's which were backfilling the data on missing title's and descriptions?, as you've confirmed there are no such entries in the DB

tomhughes:
> Sure will do, should I remove the block's which were backfilling the data on missing title's and descriptions?, as you've confirmed there are no such entries in the DB

I doubt that any other users of this code have redactions but it probably doesn't hurt to leave that in.

IshmeetSingh06:
PR is ready @tomhughes

tomhughes:
Thanks. I squashed together some of your commits to get a clean history but other than that it all looks good.

PR #4567, opened 2024-03-08
Display changeset element pagination vertically

AntonKhorev:
When using #4563 without full sidebar reload, pagination links moving around becomes more annoying. It happens because sometimes they fit to the right of the "Nodes (...)" heading and sometimes they don't. We have more vertical space available and can "rotate" the links to distribute them along the right edge of the sidebar. In this case they'll always fit to the right of the heading and the element list.

image

Includes first commits of #4563 with refactoring.

### DIFF
--- a/app/helpers/browse_helper.rb
+++ b/app/helpers/browse_helper.rb
@@ -83,24 +83,26 @@ module BrowseHelper
   end
 
   def sidebar_classic_pagination(pages, page_param)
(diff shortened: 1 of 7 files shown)


AntonKhorev:
Probably looks too weird and may not work well with something like #5355 where I don't want to spend extra horizontal space.

PR #4569, opened 2024-03-09
Preload element tags when showing changesets

AntonKhorev:
Changeset show action generates a sql query for each element to get its tags:

```
  OldWayTag Load (0.5ms)  SELECT "way_tags".* FROM "way_tags" WHERE "way_tags"."way_id" = $1 AND "way_tags"."version" = $2  [["way_id", 52], ["version", 1]]
  ↳ app/models/old_way.rb:82:in `to_h'
  OldWayTag Load (0.4ms)  SELECT "way_tags".* FROM "way_tags" WHERE "way_tags"."way_id" = $1 AND "way_tags"."version" = $2  [["way_id", 53], ["version", 1]]
  ↳ app/models/old_way.rb:82:in `to_h'
  OldWayTag Load (0.3ms)  SELECT "way_tags".* FROM "way_tags" WHERE "way_tags"."way_id" = $1 AND "way_tags"."version" = $2  [["way_id", 54], ["version", 1]]
  ↳ app/models/old_way.rb:82:in `to_h'
  OldWayTag Load (0.3ms)  SELECT "way_tags".* FROM "way_tags" WHERE "way_tags"."way_id" = $1 AND "way_tags"."version" = $2  [["way_id", 55], ["version", 1]]
  ↳ app/models/old_way.rb:82:in `to_h'
  OldWayTag Load (0.4ms)  SELECT "way_tags".* FROM "way_tags" WHERE "way_tags"."way_id" = $1 AND "way_tags"."version" = $2  [["way_id", 56], ["version", 1]]
  ↳ app/models/old_way.rb:82:in `to_h'
  OldWayTag Load (0.3ms)  SELECT "way_tags".* FROM "way_tags" WHERE "way_tags"."way_id" = $1 AND "way_tags"."version" = $2  [["way_id", 57], ["version", 1]]
  ↳ app/models/old_way.rb:82:in `to_h'
  OldWayTag Load (0.3ms)  SELECT "way_tags".* FROM "way_tags" WHERE "way_tags"."way_id" = $1 AND "way_tags"."version" = $2  [["way_id", 58], ["version", 1]]
  ↳ app/models/old_way.rb:82:in `to_h'
```

It's possible to roll them into one query. Classic pagination has a parameter for that.

```
  OldWayTag Load (0.3ms)  SELECT "way_tags".* FROM "way_tags" WHERE "way_tags"."way_id" IN ($1, $2, $3, $4, $5, $6, $7) AND "way_tags"."version" = $8  [["way_id", 52], ["way_id", 53], ["way_id", 54], ["way_id", 55], ["way_id", 56], ["way_id", 57], ["way_id", 58], ["version", 1]]
  ↳ app/views/changesets/show.html.erb:82
```

### DIFF
--- a/app/controllers/changesets_controller.rb
+++ b/app/controllers/changesets_controller.rb
@@ -84,9 +84,9 @@ class ChangesetsController < ApplicationController
                 else
                   @changeset.comments.includes(:author)
                 end
-    @node_pages, @nodes = paginate(:old_nodes, :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "node_page")
-    @way_pages, @ways = paginate(:old_ways, :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "way_page")
-    @relation_pages, @relations = paginate(:old_relations, :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "relation_page")
+    @node_pages, @nodes = paginate(:old_nodes, :include => :old_tags, :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "node_page")
+    @way_pages, @ways = paginate(:old_ways, :include => :old_tags, :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "way_page")
+    @relation_pages, @relations = paginate(:old_relations, :include => :old_tags, :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "relation_page")
     if @changeset.user.active? && @changeset.user.data_public?
       @next_by_user = @changeset.user.changesets.where("id > ?", @changeset.id).reorder(:id => :asc).first
       @prev_by_user = @changeset.user.changesets.where("id < ?", @changeset.id).reorder(:id => :desc).first


AntonKhorev:
Although I'm not sure if rails' `.includes` works correctly with composite keys.

```
irb(main):020> OldNode.where("changeset_id = 81").includes(:old_tags).explain
  OldNode Load (0.7ms)  SELECT "nodes".* FROM "nodes" WHERE (changeset_id = 81)
  OldNodeTag Load (0.6ms)  SELECT "node_tags".* FROM "node_tags" WHERE "node_tags"."node_id" IN ($1, $2, $3) AND "node_tags"."version" IN ($4, $5, $6)  [["node_id", 1314], ["node_id", 1312], ["node_id", 1313], ["version", 1], ["version", 3], ["version", 2]]
=> 
EXPLAIN SELECT "nodes".* FROM "nodes" WHERE (changeset_id = 81)
                                     QUERY PLAN
-------------------------------------------------------------------------------------
 Index Scan using nodes_changeset_id_idx on nodes  (cost=0.28..8.29 rows=1 width=53)
   Index Cond: (changeset_id = 81)
(2 rows)

EXPLAIN SELECT "node_tags".* FROM "node_tags" WHERE "node_tags"."node_id" IN ($1, $2, $3) AND "node_tags"."version" IN ($4, $5, $6) [["node_id", 1314], ["node_id", 1312], ["node_id", 1313], ["version", 1], ["version", 3], ["version", 2]]
                                  QUERY PLAN
------------------------------------------------------------------------------
 Bitmap Heap Scan on node_tags  (cost=12.86..19.98 rows=3 width=29)
   Recheck Cond: (node_id = ANY ('{1314,1312,1313}'::bigint[]))
   Filter: (version = ANY ('{1,3,2}'::bigint[]))
   ->  Bitmap Index Scan on node_tags_pkey  (cost=0.00..12.86 rows=3 width=0)
         Index Cond: (node_id = ANY ('{1314,1312,1313}'::bigint[]))
(5 rows)
```

Conditions like `WHERE "node_tags"."node_id" IN ($1, $2, $3) AND "node_tags"."version" IN ($4, $5, $6)` select unnecessary tag versions.

pnorman:
Rather than getting the way from the DB and then sending a query with the node list, could we get all the info on the first place? This might need two queries if running on the read replicas.

Fwiw, I've never really liked how the browse page shows information that is not part of the full API.

tomhughes:
@AntonKhorev that is definitely a bug so we can't use this as it is.

@pnorman are you talking about the fact it makes an API call to get the geometry? that doesn't seem to be anything to do with this PR though?

tomhughes:
So although that query does potentially select more records than it should it looks like rails discards the extra ones during the process of associating the tags with the parent records... An alternative is to use `eager_load` which does the preload using a join and avoids the over selection.

PR #4570, opened 2024-03-11
Fix whitespace in changeset element lists

AntonKhorev:
Whitespace at the start of a list item becomes significant when element icons are added using `::before` pseudoelement. You can see the extra space when elements with/without icons are next to each other. I'm removing unnecessary whitespace inside `<li>` here.

Before:
image

After:
image


tomhughes:

PR #4571, opened 2024-03-13
Don't assume order when testing links to changeset elements

AntonKhorev:
I noticed that one test fails sometimes:

```
Failure:
ChangesetsControllerTest#test_show_paginated_element_links [test/controllers/changesets_controller_test.rb:312]:
Expected exactly 1 element matching "a[href='/way/109']", found 0.
Expected: 1
  Actual: 0

bin/rails test test/controllers/changesets_controller_test.rb:301
```

That's probably because I assumed that on the changeset page the elements will be shown in the same order as they were created. But actually they are not sorted, so sometimes the order may be different.

Here I remove the ordering assumption and check the next page for an element that was missing from the first page.

### DIFF
--- a/test/controllers/changesets_controller_test.rb
+++ b/test/controllers/changesets_controller_test.rb
@@ -306,14 +306,60 @@ class ChangesetsControllerTest < ActionDispatch::IntegrationTest
     relations = create_list(:relation, page_size + 1, :with_history, :changeset => changeset)
 
     sidebar_browse_check :changeset_path, changeset.id, "changesets/show"
-    page_size.times do |i|
-      assert_dom "a[href='#{node_path nodes[i]}']", :count => 1
-      assert_dom "a[href='#{old_node_path nodes[i], 1}']", :count => 1
-      assert_dom "a[href='#{way_path ways[i]}']", :count => 1
-      assert_dom "a[href='#{old_way_path ways[i], 1}']", :count => 1
-      assert_dom "a[href='#{relation_path relations[i]}']", :count => 1
-      assert_dom "a[href='#{old_relation_path relations[i], 1}']", :count => 1
+
+    next_page_node = nil
+    nodes.each do |node|
+      node_skip = true
+      assert_dom "a[href='#{node_path node}']", :minimum => 0, :maximum => 1 do
+        node_skip = false
+      end
+      assert_dom "a[href='#{old_node_path node, 1}']", :count => (node_skip ? 0 : 1)
+      if node_skip
+        assert_nil next_page_node
+        next_page_node = node
+      end
     end
+    assert_not_nil next_page_node
+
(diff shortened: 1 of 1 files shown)


gravitystorm:
> But actually they are not sorted, so sometimes the order may be different.

Good catch.

Would it be a better idea for us to sort the elements instead? I've had a brief look at the website, and it would seam reasonable to sort the elements by id and version. That way the order is stable, and also might be helpful for mappers too (e.g I think if I saw nodes 1, 2, 5, ..., 20 on the first page, I'd be surprised to find node 4 mentioned on page 7).

I also think the test could be improved. At the moment the test creates 21 items for each type of element, and then checks that the first 20 are visible. But it doesn't check that the 21st item is *not* visible, and it doesn't check that there are page links to view the next page. I can see what you are doing with the changes that you propose, but I think that leads to a very complex test, in order to handle a situation (lack of sort order) that I don't think is intentional.

AntonKhorev:
> But it doesn't check that the 21st item is not visible

It does:
https://github.com/openstreetmap/openstreetmap-website/pull/4571/files#diff-6cdb8b8fa44e422834c4815943a8e4933d6d09e4687f338d007daeff1a90daa9R322

AntonKhorev:
> it would seam reasonable to sort the elements by id and version

Another reasonable order is by modification time, which is what usually happens already.

gravitystorm:
> It does:
> https://github.com/openstreetmap/openstreetmap-website/pull/4571/files#diff-6cdb8b8fa44e422834c4815943a8e4933d6d09e4687f338d007daeff1a90daa9R322

I'm sorry I wasn't clear - I was commenting on the original test. I meant that even if we establish a sort order, the test could still benefit from improvements.

AntonKhorev:
Changeset download sorts elements by (timestamp, version):
https://github.com/openstreetmap/openstreetmap-website/commit/8a57904398175d9143cec737282f5066793bdbc3

AntonKhorev:
Meanwhile the test keeps failing sometimes https://github.com/openstreetmap/openstreetmap-website/actions/runs/8284476948/job/22677523145?pr=4578

tomhughes:
I'm with @gravitystorm that this is the wrong fix - aside from the horrendous complexity involved in this fix it's fundamentally wrong to try and do pagination without a fixed order.

Without a sort you have no guarantee that the database will use the same order for each page so you may wind up seeing some things twice and not seeing others at all as you move through the pages.

AntonKhorev:
15 years of wrong pagination https://github.com/openstreetmap/openstreetmap-website/commit/2ed84e26cdeaaa20b8f607624d65b7891fcf1fed

AntonKhorev:
Not needed if #5209 is merged.

PR #4572, opened 2024-03-13
Update hosting partners as per OWG policy

lonvia:
OWG has changed its [Hosting Provider Credit Policy](https://operations.osmfoundation.org/policies/hosting/). This implements the new policy.

I had to rename the translation string because removing an interpolation from the string caused all kinds of test failures with the yet-to-be-updated translations. Let me know if you prefer a different solution here.


Firefishy:
Looks good to me

gravitystorm:
Thanks!

PR #4574, opened 2024-03-13
Add more wiki prefixes

urbalazs:
Both cases are documented on OSM wiki. See:

- https://wiki.openstreetmap.org/wiki/Key:manufacturer:wikidata
- https://wiki.openstreetmap.org/wiki/Key:manufacturer:wikipedia
- https://wiki.openstreetmap.org/wiki/Key:model:wikidata
- https://wiki.openstreetmap.org/wiki/Key:model:wikipedia


gravitystorm:
Thanks @urbalazs . I don't want to see this list expand too much (there's over 400 values according to taginfo) but since these are on the summary pages on the wiki (as linked in the code) I'm happy to merge these two.

PR #4575, opened 2024-03-13
Collect node/way/relation CanCanCan subjects into arrays

AntonKhorev:
Different element types usually have the same actions with three very similar lines in ability files. It's possible to turn them into one line, also decreasing the complexity Rubocop complains about. I still have PRs where I need to increase complexity limits like #4301. After this change I won't have to do it.


tomhughes:
Nice (well other than that horrid line end condition but that's rubocop's fault not yours).

PR #4576, opened 2024-03-14
Move element history actions to old element controllers

AntonKhorev:
Now they are in `index` actions to match `show` actions from #4480.

Views and path helpers are not changed in this PR.


tomhughes:
Looks good, thanks.

PR #4577, opened 2024-03-14
Replace "Login" labels with "Log in" when used as a verb

milan-cvetkovic:
Initiated by the following thread https://github.com/openstreetmap/openstreetmap-website/pull/4455#discussion_r1512725362

Split a separate PR from PR #4455.


tomhughes:
Thanks for splitting it out, looks good to me.

PR #4578, opened 2024-03-14
Fix "blocked" user signup page layout

milan-cvetkovic:
Split away from #4455, result of discussion https://github.com/openstreetmap/openstreetmap-website/pull/4455#discussion_r1519599220:

- Remove hanging arm
- Remove `<h1>` and `<h2>` tags for main text.
- Emphasize1st paragraph with `<strong>`

### DIFF
--- a/app/views/users/blocked.html.erb
+++ b/app/views/users/blocked.html.erb
@@ -3,10 +3,9 @@
   <div class='header-illustration new-user-main'>
     <h1><%= t "users.new.title" %></h1>
   </div>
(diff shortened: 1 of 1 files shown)


milan-cvetkovic:
@tomhughes Any idea why the test on Ubuntu 20.04, Ruby 3.1 fails - the failure does not seem to be related to this PR?

tomhughes:
I think that's #4571.

tomhughes:
That looks quite a lot better, thanks.

PR #4579, opened 2024-03-15
Move current element actions to their own controllers

AntonKhorev:
Continuing #4576, now moving `show` actions from Browse controller to Nodes/Ways/RelationsController.


tomhughes:
Another good step in the right direction, thanks.

PR #4580, opened 2024-03-15
Use official "Log in to Microsoft" icon

milan-cvetkovic:
Downloaded from https://learn.microsoft.com/en-us/entra/identity-platform/howto-add-branding-in-apps
on 2024-03-15.


app/helpers/user_helper.rb
```diff
@@ -58,7 +58,11 @@ def openid_logo
 
   def auth_button(name, provider, options = {})
     link_to(
-      image_tag("#{name}.svg", :alt => t("sessions.new.auth_providers.#{name}.alt"), :class => "rounded-3"),
+      image_tag("#{name}.svg",
+                :alt => t("sessions.new.auth_providers.#{name}.alt"),
+                :class => "rounded-3",
+                :width => "36px",
+                :height => "36px"),
```
AntonKhorev, line 65: `px` shouldn't be in width/height attributes: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#height

You can write `:size => "36x36"` instead: https://api.rubyonrails.org/classes/ActionView/Helpers/AssetTagHelper.html#method-i-image_tag-label-Options

tomhughes:
Looks good - the test failure is unrelated.

PR #4581, opened 2024-03-16
Test links to containing relations from element pages

AntonKhorev:
Added tests that would have failed because https://github.com/openstreetmap/openstreetmap-website/pull/4579#issuecomment-2000202320


tomhughes:
Great, thanks for doing that.

PR #4582, opened 2024-03-16
Don't pass paths to element_single_current_link helper

AntonKhorev:
`link_to` can often be simplified by passing to it just an object instead of a path or a controller name. This may help with refactoring. The simplification can be done in many other places in addition to what's in this PR, but just in case I'll do it in small steps, at least until #4581 is merged.


tomhughes:
Looks good. Suspect I introduced some other cases that can be simplified like that with my commit last night...

PR #4583, opened 2024-03-16
Move changeset subscribe/unsubscribe to resourceful routes

AntonKhorev:
Since changesets are already declared as resources, I moved subscribe/unsubscribe routes under their resources definition. The only problem is that rails named their path helpers differently, `subscribe_changeset_path` etc instead of previous `changeset_subscribe_path`.


tomhughes:
Looks good to me, thanks.

PR #4598, opened 2024-03-20
Fix trace feed image urls and local variable names

### LINKED ISSUE #4596: Refactor traces georss description partial
While reviewing #4595 I spotted a few things that could be improved with the [`traces/_description`](https://github.com/openstreetmap/openstreetmap-website/blob/3b1fb1c73bf1c82e29ecc8bf33e64f7985c4a8c0/app/views/traces/_description.html.erb) partial:

* It uses a relative path for the icon image, but this partial is used for rss feeds. It should probably be a fully qualified url
* the trace object is passed in, but without an alias. This means that we have e.g. `description.user` which is confusing, since `description` is a Trace object here. We should do something like `render(:partial => "description", :object => trace, :as => "trace")` or similar, or with passing a locals array
* I think we can avoid the trailing `.id` on the object passed to the path method

Overall, `trace_icon_path(description.user, description.id)` could end up something like `trace_icon_url(trace.user, trace)`

AntonKhorev:
fixes #4596

### DIFF
--- a/app/views/traces/_description.html.erb
+++ b/app/views/traces/_description.html.erb
@@ -1,4 +1,4 @@
-<%= image_tag trace_icon_path(description.user, description.id) %>
+<%= image_tag trace_icon_url(description.user, description.id) %>
 <% if description.size -%>
 <%= t ".description_with_count", :count => description.size, :user => description.user.display_name %>
 <% else -%>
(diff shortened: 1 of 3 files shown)


gravitystorm:
Thanks!

PR #4599, opened 2024-03-20
Rename message local in message table partials

AntonKhorev:
see https://github.com/openstreetmap/openstreetmap-website/issues/4596#issuecomment-2009693014


gravitystorm:
Great, thanks.

There's probably other places where using `:object` means that it's confusing, and some places where it's probably not confusing at all. I'm happy to treat each on a case-by-case basis, and I'm happy with both this approach and also the alternative `:locals` approach.

PR #4602, opened 2024-03-20
Avoid lego translation in trace time/user/tags string

### LINKED ISSUE #4597: Avoid lego translation in trace descriptions
https://github.com/openstreetmap/openstreetmap-website/blob/3b1fb1c73bf1c82e29ecc8bf33e64f7985c4a8c0/app/views/traces/_trace.html.erb#L31-L38

This is a lego translation (parts of a sentence stuck together) and so is hard/impossible to translate.

See also:
* #4012 (which is similar, but for emails)
* #3740 (the second example actually refers to this specific problem)

AntonKhorev:
Fixes #4597, but tags are still joined by `,` which might be a problem for some languages.

### DIFF
--- a/app/views/traces/_trace.html.erb
+++ b/app/views/traces/_trace.html.erb
@@ -29,11 +29,13 @@
       </li>
     </ul>
     <p class="text-muted mb-0">
-      <%= friendly_date_ago(trace.timestamp) %>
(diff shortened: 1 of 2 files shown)


gravitystorm:
Thanks for sorting this too.

PR #4603, opened 2024-03-21
Remove unused "traces_from" locale string

AntonKhorev:
Not in use since https://github.com/openstreetmap/openstreetmap-website/commit/62349d18564b6d0ac9a4aac61b298df913f5c08d

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2557,7 +2557,6 @@ en:
       upload_trace: "Upload a trace"
       all_traces: "All Traces"
       my_traces: "My Traces"
-      traces_from: "Public Traces from %{user}"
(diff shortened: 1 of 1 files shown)


gravitystorm:
It looks like I made a typo in the original commit. We have two very similar translations:

* public_traces_from: "Public GPS Traces from %{user}"
* traces_from: "Public Traces from %{user}"

The first is used in the controller for titles, and all of these titles contain GPS. For tabs, none of the tab labels contained "GPS", but I've made the mistake and used `public_traces_from` instead of `traces_from`.

So I think the fix here should be to fix the typo in the view, rather than remove the translation. This will make the tab labels consistent.

PR #4604, opened 2024-03-21
Make factory-created revoked blocks inactive

AntonKhorev:
While testing you can create a revoked block with `create(:user_block, :revoked)`. However it's going to be active because its expiration date is not set. This can't happen when you normally use the osm website because block model's `revoke!` method expires the block.

Here I make all factory-created revoked blocks also expired. And just in case I also set the creation date earlier than the expiration date.


gravitystorm:
Thanks!

Issue #4509, opened 2024-02-07
Proposal for Messages API


milan-cvetkovic:
### Problem

There are already several issues related to OSM Messaging and the visibility of messages to the user base:

- [PR #4070](https://github.com/openstreetmap/openstreetmap-website/issues/4070) - cosmetic modification on how number of new messages is presented
- [Issue #3129](https://github.com/openstreetmap/openstreetmap-website/issues/3129) - (closed) Block function to individual users
- [Forum discussion](https://community.openstreetmap.org/t/about-making-osms-messaging-features-more-efficient/100050)

As a main form of communicating to users OSM uses email. In some scenarios, OSM website displays a message notification indicator. In other scenarios, i.e. changeset comments, notes and diary entries activities, OSM website only sends emails to subscribed users. In some cases, there are RSS feeds as well.

If user uses alternate applications to contribute to OSM, there is limited API support for querying information related to messages users would be interested in:

 - Querying user details shows the summary counters for direct messages. Some applications use this info to display a notification and redirect users to OSM web site to access their messages.
 - There is no support for reading, replying or modifying messages.
 - Extracting the changeset comments user is subscribed to has limited support. There are active PRs improving the support in this area: [PR #4359](https://github.com/openstreetmap/openstreetmap-website/pull/4359). As of now, there is indirect support through RSS feed. This is not in scope for this github issue.

## Proposal - Extend API v0.6 to include Messages API

This is the proposal for OSM Messages API addressing direct messages only.

### Requirements

  - applications need to ask the user's permission to access the messages on OSM on their behalf. The existing OAuth2 authorization framework would support the authorization, however the messaging is not covered by existing scopes. We propose    to add 2 additional scopes, one for reading messages, and an additional scope for creating new messages and sending them.
  - Messagesing API requires the client to be properly authorized by authenticated user.
    Only messages related to the authorizing user are accessible - the user must be either the recipient or the originator of the message.
  - To prevent bulk messaging, sending a message must be throttled, similar to how changeset submission is throttled.
  
This proposal assumes the existing data model for messages in OSM. A message is shared by the originator and recipient. The messages can be marked as read/unread or deleted by the user viewing them. OSM website displays the messages separately in "inbox" and "outbox".

In addition, the API provisions for paginating the list of messages. Message id is used for ordering since for any other fields, for example ordering by message timestamp, would require an additional index.

### Throttling Considerations

[...]

PR #4605, opened 2024-03-21
Add Messages API

milan-cvetkovic:
...as discussed in [Proposal for Messages API - Issue #4509](https://github.com/openstreetmap/openstreetmap-website/issues/4509) and documented in [Messaging API reference](https://wiki.openstreetmap.org/w/index.php?title=Messaging_API_proposal)

### DIFF
--- a/app/controllers/api/messages_controller.rb
+++ b/app/controllers/api/messages_controller.rb
@@ -1,0 +1,149 @@
+# The MessagesController is the RESTful interface to Message objects
+
+module Api
+  class MessagesController < ApiController
+    before_action :authorize
+
+    before_action :check_api_writable, :only => [:create, :update, :destroy]
+    before_action :check_api_readable, :except => [:create, :update, :destroy]
+
+    authorize_resource
+
+    around_action :api_call_handle_error, :api_call_timeout
+
+    before_action :set_request_formats
+
+    def inbox
+      @skip_body = true
+      @messages = Message.includes(:sender, :recipient).where(:to_user_id => current_user.id)
+
+      show_messages
+    end
+
+    def outbox
+      @skip_body = true
+      @messages = Message.includes(:sender, :recipient).where(:from_user_id => current_user.id)
+
+      show_messages
+    end
+
+    # Dump the details on a message given in params[:id]
+    def show
+      @message = Message.includes(:sender, :recipient).find(params[:id])
+
+      raise OSM::APIAccessDenied if current_user.id != @message.from_user_id && current_user.id != @message.to_user_id
+
+      # Render the result
+      respond_to do |format|
+        format.xml
+        format.json
+      end
+    end
+
+    # Create a new message from current user
+    def create
+      # Check the arguments are sane
+      raise OSM::APIBadUserInput, "No title was given" if params[:title].blank?
+      raise OSM::APIBadUserInput, "No body was given" if params[:body].blank?
+
+      # Extract the arguments
+      if params[:recipient_id]
+        recipient_id = params[:recipient_id].to_i
+        recipient = User.find(recipient_id)
+      elsif params[:recipient]
+        recipient_display_name = params[:recipient]
+        recipient = User.find_by(:display_name => recipient_display_name)
+      else
+        raise OSM::APIBadUserInput, "No recipient was given"
+      end
+
+      raise OSM::APIRateLimitExceeded if current_user.sent_messages.where(:sent_on => Time.now.utc - 1.hour..).count >= current_user.max_messages_per_hour
+
+      @message = Message.new(:sender => current_user,
+                             :recipient => recipient,
+                             :sent_on => Time.now.utc,
+                             :title => params[:title],
+                             :body => params[:body],
+                             :body_format => "markdown")
+      @message.save!
+
+      UserMailer.message_notification(@message).deliver_later if @message.notify_recipient?
+
+      # Return a copy of the new message
+      respond_to do |format|
+        format.xml { render :action => :show }
+        format.json { render :action => :show }
+      end
+    end
+
+    # Update read status of a message
+    def update
+      @message = Message.find(params[:id])
+      read_status_idx = %w[true false].index params[:read_status]
+
+      raise OSM::APIBadUserInput, "Invalid value of `read_status` was given" if read_status_idx.nil?
+      raise OSM::APIAccessDenied unless current_user.id == @message.to_user_id
+
+      @message.message_read = read_status_idx.zero?
+      @message.save!
+
+      # Return a copy of the message
+      respond_to do |format|
+        format.xml { render :action => :show }
+        format.json { render :action => :show }
+      end
+    end
+
+    # Delete message by marking it as not visible for the current user
+    def destroy
+      @message = Message.find(params[:id])
+      if current_user.id == @message.from_user_id
+        @message.from_user_visible = false
+      elsif current_user.id == @message.to_user_id
+        @message.to_user_visible = false
+      else
+        raise OSM::APIAccessDenied
+      end
+
+      @message.save!
+
+      # Return a copy of the message
+      respond_to do |format|
+        format.xml { render :action => :show }
+        format.json { render :action => :show }
+      end
+    end
+
+    private
+
+    def show_messages
+      @messages = @messages.where(:muted => false)
+      if params[:order].nil? || params[:order] == "newest"
+        @messages = @messages.where(:id => ..params[:from_id]) unless params[:from_id].nil?
+        @messages = @messages.order(:id => :desc)
+      elsif params[:order] == "oldest"
+        @messages = @messages.where(:id => params[:from_id]..) unless params[:from_id].nil?
+        @messages = @messages.order(:id => :asc)
+      else
+        raise OSM::APIBadUserInput, "Invalid order specified"
+      end
+
+      limit = params[:limit]
+      if !limit
+        limit = Settings.default_message_query_limit
+      elsif !limit.to_i.positive? || limit.to_i > Settings.max_message_query_limit
+        raise OSM::APIBadUserInput, "Messages limit must be between 1 and #{Settings.max_message_query_limit}"
+      else
+        limit = limit.to_i
+      end
+
+      @messages = @messages.limit(limit)
+
+      # Render the result
+      respond_to do |format|
+        format.xml
+        format.json
+      end
+    end
+  end
+end
--- a/app/views/api/messages/_message.json.jbuilder
+++ b/app/views/api/messages/_message.json.jbuilder
@@ -1,0 +1,17 @@
+json.id message.id
+json.from_user_id message.from_user_id
+json.from_display_name message.sender.display_name
+json.to_user_id message.to_user_id
+json.to_display_name message.recipient.display_name
+json.title message.title
+json.sent_on message.sent_on.xmlschema
+
+if current_user.id == message.from_user_id
+  json.deleted !message.from_user_visible
+elsif current_user.id == message.to_user_id
+  json.message_read message.message_read
+  json.deleted !message.to_user_visible
+end
+
+json.body_format message.body_format
+json.body message.body unless @skip_body
--- a/app/views/api/messages/_message.xml.builder
+++ b/app/views/api/messages/_message.xml.builder
@@ -1,0 +1,21 @@
+attrs = {
+  "id" => message.id,
+  "from_user_id" => message.from_user_id,
+  "from_display_name" => message.sender.display_name,
+  "to_user_id" => message.to_user_id,
+  "to_display_name" => message.recipient.display_name,
+  "sent_on" => message.sent_on.xmlschema,
+  "body_format" => message.body_format
+}
+
+if current_user.id == message.from_user_id
+  attrs["deleted"] = !message.from_user_visible
+elsif current_user.id == message.to_user_id
+  attrs["message_read"] = message.message_read
+  attrs["deleted"] = !message.to_user_visible
+end
+
+xml.message(attrs) do |nd|
+  nd.title(message.title)
+  nd.body(message.body) unless @skip_body
+end
--- a/app/views/api/messages/outbox.json.jbuilder
+++ b/app/views/api/messages/outbox.json.jbuilder
@@ -1,0 +1,5 @@
+json.partial! "api/root_attributes"
+
+json.messages(@messages) do |message|
+  json.partial! message
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -78,6 +78,15 @@ OpenStreetMap::Application.routes.draw do
       end
     end
 
+    resources :messages, :path => "user/messages", :constraints => { :id => /\d+/ }, :only => [:create, :show, :destroy], :controller => "messages", :as => :api_messages do
+      collection do
+        get "inbox"
+        get "outbox"
+      end
+    end
+
+    post "/user/messages/:id" => "messages#update", :as => :api_message_update
+
     post "gpx/create" => "traces#create"
     get "gpx/:id" => "traces#show", :as => :api_trace, :id => /\d+/
     put "gpx/:id" => "traces#update", :id => /\d+/
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -59,6 +59,10 @@ user_block_periods: [0, 1, 3, 6, 12, 24, 48, 96, 168, 336, 731, 4383, 8766, 8766
 user_account_deletion_delay: null
 # Rate limit for message sending
 max_messages_per_hour: 60
+# Default limit on the number of messages returned by inbox and outbox message api
+default_message_query_limit: 100
+# Maximum number of messages returned by inbox and outbox message api
+max_message_query_limit: 100
 # Rate limit for friending
 max_friends_per_hour: 60
 # Rate limit for changeset comments
--- a/lib/oauth.rb
+++ b/lib/oauth.rb
@@ -2,7 +2,7 @@ module Oauth
   SCOPES = %w[read_prefs write_prefs write_diary write_api read_gpx write_gpx write_notes].freeze
   PRIVILEGED_SCOPES = %w[read_email skip_authorization].freeze
   MODERATOR_SCOPES = %w[write_redactions].freeze
-  OAUTH2_SCOPES = %w[write_redactions openid].freeze
+  OAUTH2_SCOPES = %w[write_redactions consume_messages send_messages openid].freeze
 
   class Scope
     attr_reader :name
--- a/test/controllers/api/messages_controller_test.rb
+++ b/test/controllers/api/messages_controller_test.rb
@@ -1,0 +1,611 @@
+require "test_helper"
+
+module Api
+  class MessagesControllerTest < ActionDispatch::IntegrationTest
+    ##
+    # test all routes which lead to this controller
+    def test_routes
+      assert_routing(
+        { :path => "/api/0.6/user/messages/inbox", :method => :get },
+        { :controller => "api/messages", :action => "inbox" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/inbox.xml", :method => :get },
+        { :controller => "api/messages", :action => "inbox", :format => "xml" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/inbox.json", :method => :get },
+        { :controller => "api/messages", :action => "inbox", :format => "json" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/outbox", :method => :get },
+        { :controller => "api/messages", :action => "outbox" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/outbox.xml", :method => :get },
+        { :controller => "api/messages", :action => "outbox", :format => "xml" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/outbox.json", :method => :get },
+        { :controller => "api/messages", :action => "outbox", :format => "json" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/1", :method => :get },
+        { :controller => "api/messages", :action => "show", :id => "1" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/1.xml", :method => :get },
+        { :controller => "api/messages", :action => "show", :id => "1", :format => "xml" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/1.json", :method => :get },
+        { :controller => "api/messages", :action => "show", :id => "1", :format => "json" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages", :method => :post },
+        { :controller => "api/messages", :action => "create" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/1", :method => :post },
+        { :controller => "api/messages", :action => "update", :id => "1" }
+      )
+      assert_routing(
+        { :path => "/api/0.6/user/messages/1", :method => :delete },
+        { :controller => "api/messages", :action => "destroy", :id => "1" }
+      )
+    end
+
+    def test_create_success
+      recipient = create(:user)
+      sender = create(:user)
+
+      sender_token = create(:oauth_access_token,
+                            :resource_owner_id => sender.id,
+                            :scopes => %w[send_messages consume_messages])
+      sender_auth = bearer_authorization_header(sender_token.token)
+
+      msg = build(:message)
+
+      assert_difference "Message.count", 1 do
+        assert_difference "ActionMailer::Base.deliveries.size", 1 do
+          perform_enqueued_jobs do
+            post api_messages_path,
+                 :params => { :title => msg.title,
+                              :recipient_id => recipient.id,
+                              :body => msg.body,
+                              :format => "json" },
+                 :headers => sender_auth
+            assert_response :success
+          end
+        end
+      end
+
+      assert_equal "application/json", response.media_type
+      js = ActiveSupport::JSON.decode(@response.body)
+      jsm = js["message"]
+      assert_not_nil jsm
+      assert_not_nil jsm["id"]
+      assert_equal sender.id, jsm["from_user_id"]
+      assert_equal sender.display_name, jsm["from_display_name"]
+      assert_equal recipient.id, jsm["to_user_id"]
+      assert_equal recipient.display_name, jsm["to_display_name"]
+      assert_equal msg.title, jsm["title"]
+      assert_not_nil jsm["sent_on"]
+      assert_equal !msg.from_user_visible, jsm["deleted"]
+      assert_not jsm.key?("message_read")
+      assert_equal "markdown", jsm["body_format"]
+      assert_equal msg.body, jsm["body"]
+    end
+
+    def test_create_fail
+      recipient = create(:user)
+
+      sender = create(:user)
+      sender_token = create(:oauth_access_token,
+                            :resource_owner_id => sender.id,
+                            :scopes => %w[send_messages consume_messages])
+      sender_auth = bearer_authorization_header(sender_token.token)
+
+      assert_no_difference "Message.count" do
+        assert_no_difference "ActionMailer::Base.deliveries.size" do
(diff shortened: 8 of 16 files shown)


lib/oauth.rb
```diff
@@ -2,7 +2,7 @@ module Oauth
   SCOPES = %w[read_prefs write_prefs write_diary write_api read_gpx write_gpx write_notes].freeze
   PRIVILEGED_SCOPES = %w[read_email skip_authorization].freeze
   MODERATOR_SCOPES = %w[write_redactions].freeze
-  OAUTH2_SCOPES = %w[write_redactions openid].freeze
+  OAUTH2_SCOPES = %w[write_redactions message_send message_consume openid].freeze
```
tordans, line 5: Would it be more consistent to have something like 'write_messages', 'read_messages'?

tordans (commented):
I don't know enough about he code but it looks like this likely duplicates some logic. Can this be shared in a concern or something similar?

lib/oauth.rb
milan-cvetkovic: To clarify, I don't mind the names and/or having different set of scopes.
- read message without modifying anything (peek messages)
- mark messages read or unread
- mark messages as deleted
- send messages to other users

It seemed as if the reading messages makes sense together with the ability to mark them as read and deleted, and I named the scope for those ops `message_consume`

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+      respond_to do |format|
+        format.xml { render :action => :show }
+        format.json { render :action => :show }
+      end
+    end
+
+    private
+
+    def show_messages
+      if params[:min_id]
```
AntonKhorev, line 126: Elsewhere (#4532) I wanted to standardize on `before`/`after` for exclusive bounds (already in use in pagination) and `from`/`to` fro inclusive bounds (already in use in the api, although just for dates).

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+        @messages = @messages.where("id >= ?", params[:min_id]) unless params[:min_id] == "0"
+        @messages = @messages.order(:id => :asc)
+      else
+        @messages = @messages.where("id <= ?", params[:max_id]) unless params[:max_id].nil? || params[:max_id] == "0"
+        @messages = @messages.order(:id => :desc)
+      end
+
+      limit = params[:count]
```
AntonKhorev, line 134: This parameter is called `limit` everywhere else.

app/views/api/messages/_message.xml.builder
```diff
@@ -0,0 +1,12 @@
...
+  xml.from_user_id message.from_user_id
+  xml.to_user_id message.to_user_id
+  xml.title message.title
+  xml.sent_on message.sent_on
+  xml.message_read message.message_read
+  xml.from_user_visible message.from_user_visible
+  xml.to_user_visible message.to_user_visible
+  xml.body_format message.body_format
+  xml.body message.body unless @skip_body
+end
```
AntonKhorev, line 12: This is "notes api style" (ids as elements rather than attributes etc) which is an exception among osm apis.

app/controllers/api/messages_controller.rb
milan-cvetkovic: Ok, will modify to `limit`.

app/views/api/messages/_message.xml.builder
milan-cvetkovic: True, I modeled after notes api, without knowing that it is an exception. Will modify the spec to use attributes, with exception of `<body>` and `<title>`.

app/controllers/api/messages_controller.rb
milan-cvetkovic: Using timestamp in this case would require us to add an additional index. There is also an edge case of multiple messages with same timestamp, potentially resulting in result set longer than "limit".

I wonder if it is possible for the list to be mis-ordered by timestamp, when it is ordered by `id`? I guess it depends on how autoincrement is implemented.

I am not sure if adding an index is worth it...

AntonKhorev: I'm not saying you should sort by timestamps. I'm saying that instead of `min` I'd call this parameter `from`, because that's how it's already called in other places. But since in other places it's for dates and here it's for ids, I guess it should be `from_id` or maybe `from[id]`.

app/views/api/messages/_message.xml.builder
```diff
@@ -0,0 +1,11 @@
+xml.tag! "message", :id => message.id,
+                    :from_user_id => message.from_user_id,
+                    :to_user_id => message.to_user_id,
+                    :sent_on => message.sent_on,
```
AntonKhorev, line 4: Another thing where notes api differ from everything else is date formats. In other places dates look like this: [`2014-04-02T17:09:44Z`](https://www.openstreetmap.org/api/0.6/way/1000). In the notes api they look like this: [`2013-05-01 08:10:42 UTC`](https://www.openstreetmap.org/api/0.6/notes/1000).

Dates here are written like in the notes api.

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    end
+
+    private
+
+    def show_messages
+      if params[:min_id]
+        @messages = @messages.where("id >= ?", params[:min_id]) unless params[:min_id] == "0"
+        @messages = @messages.order(:id => :asc)
+      else
+        @messages = @messages.where("id <= ?", params[:max_id]) unless params[:max_id].nil? || params[:max_id] == "0"
```
AntonKhorev, line 130: Are you sure you want "0" as a special value for max id? Because if treated normally, it means "select messages with id <= 0", that is "select no messages at all".

app/controllers/api/messages_controller.rb
milan-cvetkovic: I am not sure about this - how would the request for latest messages look like with `to` or `to_id`?

get 10 messages in descending order, then get the following 10 messages in desc order (assuming the smallest id from 1st request was 1000):
```
GET /api/0.6/user/messages/inbox?to=0&limit=10
GET /api/0.6/user/messages/inbox?to=1000&limit=10
```

instead of:
```
GET /api/0.6/user/messages/inbox?max_id=0&limit=10
GET /api/0.6/user/messages/inbox?max_id=1000&limit=10
```

Here `max_id` would imply descending, it wouldn't be so obvious if the name of parameter is named `to` or `to_id`.

app/views/api/messages/_message.xml.builder
milan-cvetkovic: Thanks, I wasn't aware of this. The Z format is more appropriate for API, since it is easier to machine parse. Will modify.

app/controllers/api/messages_controller.rb
AntonKhorev: That's because you change the order based on these parameters. Elsewhere in the api, the order is either always descending or there's a parameter to choose between descending/ascending.

app/controllers/api/messages_controller.rb
milan-cvetkovic: I am open to suggestions.

Maybe it would be enough to simply use missing `max_id` as `get the latest in descending order`

app/controllers/api/messages_controller.rb
AntonKhorev: Maybe you can argue that `min` and `max` look better than `from` and `to`. In this case we should add them as aliases for `from`/`to` for other endpoints.

lib/oauth.rb
milan-cvetkovic: @tordans Would `cosume_messages` and `send_messages` be better?

I've just noticed that all other scopes are all in _verb_noun_ form.

app/controllers/api/messages_controller.rb
milan-cvetkovic: `from_id` and `order=newest` or `order=oldest` would be more in line with existing OSM api?

app/controllers/api/messages_controller.rb
milan-cvetkovic: Modified to use `from_id` instead of `min_id` and `max_id`, and use absence of `from_id` to retrieve the first batch.

milan-cvetkovic:
@tomhughes Friendly ping, any chance of reviewing the PR?

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,154 @@
+# The MessagesController is the RESTful interface to Message objects
+
+module Api
+  class MessagesController < ApiController
+    before_action :authorize
+
+    before_action :check_api_writable, :only => [:create, :update, :destroy]
+    before_action :check_api_readable, :except => [:create, :update, :destroy]
```
tomhughes, line 8: This should be `inbox`, `outbox` and `show` surely?

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+    def inbox
+      @skip_body = true
+      @messages = Message.where(:to_user_id => current_user.id)
+
+      show_messages
+    end
+
+    def outbox
+      @skip_body = true
+      @messages = Message.where :from_user_id => current_user.id
```
tomhughes, line 27: Can we please add parentheses for consistency with the `inbox` action.

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+        recipient = User.find(recipient_id)
+      elsif params[:recipient]
+        recipient_display_name = params[:recipient]
+        recipient = User.find_by(:display_name => recipient_display_name)
+      else
+        raise OSM::APIBadUserInput, "No recipient was given"
+      end
+      title = params[:title]
+      body = params[:body]
+      body_format = params[:body_format]
```
tomhughes, line 64: Is there any particular point to copying these into local variables which are only going to be used once?

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+        format.json
+      end
+    end
+
+    # Create a new message from current user
+    def create
+      # Check the arguments are sane
+      raise OSM::APIBadUserInput, "No title was given" if params[:title].blank?
+      raise OSM::APIBadUserInput, "No body was given" if params[:body].blank?
+      raise OSM::APIBadUserInput, "No body format was given" if params[:body_format].blank?
```
tomhughes, line 50: Do we want to allow a choice of body format? There's no choice with the existing web interface and the field mostly exists to support legacy messages from before we switched to markdown and all new messages are markdown.

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+      # Return a copy of the message
+      respond_to do |format|
+        format.xml { render :action => :show }
+        format.json { render :action => :show }
+      end
+    end
+
+    private
+
+    def show_messages
```
tomhughes, line 125: This should probably exclude deleted messages and maybe muted ones as well?

app/views/api/messages/_message.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.id message.id
+json.from_user_id message.from_user_id
+json.to_user_id message.to_user_id
+json.title message.title
+json.sent_on message.sent_on.xmlschema
+json.message_read message.message_read
+json.from_user_visible message.from_user_visible
+json.to_user_visible message.to_user_visible
```
tomhughes, line 8: This means user can see whether the other person has deleted the message which isn't something that can see at the moment - do we want that?

app/views/api/messages/_message.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.id message.id
+json.from_user_id message.from_user_id
+json.to_user_id message.to_user_id
+json.title message.title
+json.sent_on message.sent_on.xmlschema
+json.message_read message.message_read
```
tomhughes, line 6: This allows a sender to see if the recipient has read a message which is not possible at the moment - do we want that?

app/views/api/messages/_message.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.id message.id
+json.from_user_id message.from_user_id
+json.to_user_id message.to_user_id
```
tomhughes, line 3: Do we want to include the user names as well?

app/views/api/messages/_message.xml.builder
```diff
@@ -0,0 +1,11 @@
+xml.tag! "message", :id => message.id,
+                    :from_user_id => message.from_user_id,
+                    :to_user_id => message.to_user_id,
+                    :sent_on => message.sent_on.xmlschema,
+                    :message_read => message.message_read,
+                    :from_user_visible => message.from_user_visible,
+                    :to_user_visible => message.to_user_visible,
+                    :body_format => message.body_format do
```
tomhughes, line 8: All comments about the JSON version apply here as well.

app/views/api/messages/outbox.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.partial! "api/root_attributes"
+
+json.messages(@messages) do |message|
+  xml.tag! "messages" do
```
tomhughes, line 4: This should be in the XML version surely?

config/routes.rb
```diff
@@ -78,6 +78,15 @@
...
     end
 
+    resources :messages, :path => "user/messages", :constraints => { :id => /\d+/ }, :only => [:create, :show, :destroy], :controller => "messages", :as => :api_messages do
+      collection do
+        get "inbox"
+        get "outbox"
+      end
+    end
+
+    post "/user/messages/:id" => "messages#update", :as => :api_message_update
```
tomhughes, line 88: Why can't this be done by allowing `:update` on the resource block above?

config/routes.rb
```diff
@@ -78,6 +78,15 @@
       end
     end
 
+    resources :messages, :path => "user/messages", :constraints => { :id => /\d+/ }, :only => [:create, :show, :destroy], :controller => "messages", :as => :api_messages do
```
tomhughes, line 81: Do we want to nest messages under user? The IDs are global so there's no need for it unless we think there's some reason it makes more sense that way.

test/controllers/api/messages_controller_test.rb
```diff
@@ -0,0 +1,570 @@
...
+      # Stub nominatim response for note locations
+      stub_request(:get, %r{^https://nominatim\.openstreetmap\.org/reverse\?})
+        .to_return(:status => 404)
+    end
+
+    ##
+    # test all routes which lead to this controller
+    def test_routes
+      assert_routing(
+        { :path => "/api/0.6/user/messages/inbox", :method => :get },
```
tomhughes, line 16: This (and outbox) should probably have XML and JSON versions as for show?

test/controllers/api/messages_controller_test.rb
```diff
@@ -0,0 +1,570 @@
+require "test_helper"
+
+module Api
+  class MessagesControllerTest < ActionDispatch::IntegrationTest
+    def setup
```
tomhughes, line 5: Has this been copied from notes and not removed?

tomhughes (changes requested):
I've had a first pass over the code and an initial peek at the tests and added my comments. I'll definitely want @gravitystorm to have a look as well before we merge.

milan-cvetkovic:
> I've had a first pass over the code and an initial peek at the tests and added my comments. I'll definitely want @gravitystorm to have a look as well before we merge.

Thanks for the review, will look into your comments shortly

app/controllers/api/messages_controller.rb
milan-cvetkovic: I added the text since it appears that messages generated from emails have `body_format` set to `text` https://github.com/openstreetmap/openstreetmap-website/blob/bd3507291c2d2d86796f99ede11560554d7235c5/app/models/message.rb#L59C1-L59C29.

Will remove the choice and ue markdown without any other choices.

app/controllers/api/messages_controller.rb
milan-cvetkovic: I believe we need to display deleted messages in order to allow API to "undelete" them.

app/views/api/messages/_message.json.jbuilder
milan-cvetkovic: Good catch, thanks

app/views/api/messages/_message.json.jbuilder
milan-cvetkovic: Good catch, thanks.

config/routes.rb
milan-cvetkovic: Adding `:update` to resources above would add PUT and PATCH routes. PUT implies that the whole message would be modified, and I didn't see any PATCH ops in [API_v0.6 spec](https://wiki.openstreetmap.org/wiki/API_v0.6)

The API proposal suggested POST with single parameter: `read_status` with possible values of `true` and `false`, similar to how  "POST /api/0.6/changeset/comment/#comment_id/hide" works.

app/controllers/api/messages_controller.rb
milan-cvetkovic: @tomhughes What are you suggesting here - `inbox`, `outbox` and `show` are covered by `:check_api_readable`, did you mean to exclude them as well?

tomhughes: Oh sorry I hadn't noticed that was one `:except` and was reading them both as `:only`.

config/routes.rb
gravitystorm: In general I would not recommend any nesting. I would only add the nesting if either:
* There's a need for the id of the parent object (e.g. /node/:id/relations), or
* There's a need to disambiguate between two same-named controllers (e.g. the classic `/admin/users/`)

However, in this particular case, I support @milan-cvetkovic , since it actually matches our existing API. As far as I can tell, almost every existing call where the responses are specific to the current user (and don't refer to public objects) are nested under `/user`:
* /api/0.6/user/details
* /api/0.6/user/gpx_files
* /api/0.6/user/preferences

However some are not:
* /api/0.6/permissions

(There are other calls where it's a user-specific action on an otherwise public object, like subscribing to a changeset, but I treat these as a different situation than the above).

So I think we should add the nesting for messages too, purely for consistency, even though I don't like it!

app/views/api/messages/_message.json.jbuilder
milan-cvetkovic: User names are not available in message itself, I would have to either modify the model and include those two, or query for user names in order to return them.

They would be nice to have, but they would cost some database lookups

app/controllers/api/messages_controller.rb
milan-cvetkovic: Now excluding muted messages, both in inbox and outbox.

I guess this implementation does not allow viewing and manipulation of muted messages.

milan-cvetkovic:
Rebased and updated based on #4859.

milan-cvetkovic:
@tomhughes Please advise if any modifications are required for this PR, thanks.

app/controllers/api/messages_controller.rb
tomhughes: Sounds good to me. We can always add a way to request muted messages later if somebody has a use for that but by default they should be excluded.

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+      if params[:order].nil? || params[:order] == "newest"
+        @messages = @messages.where(:id => ..params[:from_id]) unless params[:from_id].nil?
+        @messages = @messages.where(:muted => false).order(:id => :desc)
```
tomhughes, line 124: Can we lift the `.where(:muted => false)` up above the if block rather than repeating it please?

app/views/api/messages/_message.json.jbuilder
```diff
@@ -0,0 +1,15 @@
...
+json.title message.title
+json.sent_on message.sent_on.xmlschema
+
+if current_user.id == message.from_user_id
+  json.from_user_visible message.from_user_visible
```
tomhughes, line 8: Should this just be `visible` in the API response? or maybe even invert it and call it `deleted` instead?

app/views/api/messages/_message.json.jbuilder
```diff
@@ -0,0 +1,15 @@
...
+elsif current_user.id == message.to_user_id
+  json.message_read message.message_read
+  json.to_user_visible message.to_user_visible
```
tomhughes, line 11: See above re `from_user_visible` value.

app/views/api/messages/_message.json.jbuilder
tomhughes: The question is how likely is it that the user of the API will just come back and make user details calls to get the names anyway? If they're going to do that then it would be better to do it here where we can use joins to get the names more efficiently...

app/views/api/messages/_message.xml.builder
```diff
@@ -0,0 +1,19 @@
...
+  "sent_on" => message.sent_on.xmlschema,
+  "body_format" => message.body_format
+}
+
+if current_user.id == message.from_user_id
+  attrs["from_user_visible"] = message.from_user_visible
+elsif current_user.id == message.to_user_id
+  attrs["message_read"] = message.message_read
+  attrs["to_user_visible"] = message.to_user_visible
+end
```
tomhughes, line 14: As before comments about the JSON version apply here as well.

config/routes.rb
tomhughes: If @gravitystorm thinks it is, on balance, the right choice then I think we can leave this as it is.

config/settings.yml
```diff
@@ -59,6 +59,10 @@ user_block_periods: [0, 1, 3, 6, 12, 24, 48, 96, 168, 336, 731, 4383, 8766, 8766
 user_account_deletion_delay: null
 # Rate limit for message sending
 max_messages_per_hour: 60
+# Default limit on the number of messages returned by inbox and outbox message api
+default_message_query_limit: 100
+#Maximum number of messages returned by inbox and outbox message api
```
tomhughes, line 64: Space after the `#` for consistency please...

test/controllers/api/messages_controller_test.rb
tomhughes: This still seems to be present? What was the resolution?

tomhughes (changes requested):
A few more minor comments...

config/routes.rb
tomhughes: Technically I guess `PATCH` would be the correct thing but as you say it's probably not consistent with the API to date and I haven't managed to find a way to move that route inside the resource block while keeping it as POST unless @gravitystorm has some ideas?

simonpoole:
A comment on the generated output of the new API
```
"version": "0.6",
...
```
I don't believe we would want to make a statement on the copyright of the messages in the first place (because they are essentially private and in any scenario you would need to agree with whoever you had communicated with on publishing them), but if it would be anything, the copyright would be held by the individual author of the respective e-mail.

milan-cvetkovic:
> A comment on the generated output of the new API
> 
> ```
> "version": "0.6",
>     "generator": "OpenStreetMap server",
>     "copyright": "OpenStreetMap and contributors",
>     "attribution": "http://www.openstreetmap.org/copyright",
>     "license": "http://opendatacommons.org/licenses/odbl/1-0/",
> ...
> ```
> 
> I don't believe we would want to make a statement on the copyright of the messages in the first place (because they are essentially private and in any scenario you would need to agree with whoever you had communicated with on publishing them), but if it would be anything, the copyright would be held by the individual author of the respective e-mail.

This comes from including `root_attributes` in json and xml templates. I probably have it as a result of copying from other templates. However, it is not clear to me if we should include those headers in messages - your comment seems reasonable.

Side note - looking at notes (where I probably copied at least some templates from) I noticed that XML output does include the copyright notice, while json does not.

app/views/api/messages/_message.json.jbuilder
milan-cvetkovic: Updated to `deleted`

app/views/api/messages/_message.json.jbuilder
milan-cvetkovic: Updated to `deleted`

test/controllers/api/messages_controller_test.rb
milan-cvetkovic: Sorry, I guess there were 2 mentions of notes, and I only removed one.

milan-cvetkovic:
@tomhughes I updated the code according to your suggestions from the last review.

Please advise if more modifications are necessary.

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,151 @@
...
+    before_action :check_api_writable, :only => [:create, :update, :destroy]
+    before_action :check_api_readable, :except => [:create, :update, :destroy]
+
+    authorize_resource
+
+    around_action :api_call_handle_error, :api_call_timeout
+
+    before_action :set_request_formats
+
+    skip_authorization_check :only => [:inbox, :outbox]
```
tomhughes, line 16: Why are we skipping the authorization check for these methods?

app/controllers/messages_controller.rb
```diff
@@ -9,8 +9,7 @@ class MessagesController < ApplicationController
   authorize_resource
 
   before_action :lookup_user, :only => [:new, :create]
-  before_action :check_database_readable
-  before_action :check_database_writable, :only => [:new, :create, :reply, :mark, :destroy]
+  before_action :check_api_writable, :only => [:new, :create, :reply, :mark, :destroy]
```
tomhughes, line 12: Why are we changing the non-API access to check the API status instead of the database status? and why are we getting rid of the readable check for methods like `inbox` and `show`?

app/controllers/api/messages_controller.rb
```diff
@@ -0,0 +1,151 @@
...
+    def outbox
+      @skip_body = true
+      @messages = Message.includes(:sender, :recipient).where(:from_user_id => current_user.id)
+
+      show_messages
+    end
+
+    # Dump the details on a message given in params[:id]
+    def show
+      @message = Message.find(params[:id])
```
tomhughes, line 34: It doesn't matter as much here as it does for `inbox` and `outbox` but we might as well add the `includes(:sender, :recipient)` here as well.

test/controllers/api/messages_controller_test.rb
```diff
@@ -0,0 +1,611 @@
...
+      assert_equal recipient.display_name, jsm["to_display_name"]
+      assert_equal msg.title, jsm["title"]
+      assert_not_nil jsm["sent_on"]
+      assert_equal !msg.from_user_visible, jsm["deleted"]
+      assert_not jsm.key?("message_read")
+      assert_equal "markdown", jsm["body_format"]
+      assert_equal msg.body, jsm["body"]
+    end
+
+    def test_read_status
```
tomhughes, line 262: It would probably be better to call this `test_update_status` as the current name can be read two ways and I initially read it as testing that the status can be read rather than that the read status can be changed.

test/controllers/api/messages_controller_test.rb
```diff
@@ -0,0 +1,611 @@
...
+      # user | inbox | outbox
+      #   1  |   0   |   3
+      #   2  |   2   |   1
+      #   3  |   2   |   0
+      create(:message, :unread, :sender => user1, :recipient => user2)
+      create(:message, :unread, :sender => user1, :recipient => user2)
+      create(:message, :unread, :sender => user1, :recipient => user3)
+      create(:message, :unread, :sender => user2, :recipient => user3)
+
+      # only auithorized users
```
tomhughes, line 437: Spelling error...

tomhughes (changes requested):

app/controllers/messages_controller.rb
milan-cvetkovic: Thanks for noticing this. I am not sure how it sneaked in.

app/controllers/api/messages_controller.rb
milan-cvetkovic: Fixed.

milan-cvetkovic:
@tomhughes code updated for comments from July 28.

tomhughes (approved):
Thanks. I think this all looks good now!

PR #4606, opened 2024-03-21
Replace csp_meta_tag with `secure_headers` nonce

grekko:

Unblocks the PR #4562

The `csp_meta_tag` does not generate a `csp-nonce`-tag since currently the `secure_headers`-gem is responsible to manage the `Content-Security-Policy` directives.

Once `secure_headers`-logic is moved/delegated back to Rails the usage of `csp_meta_tag` is useful again. Otherwise it is a bit confusing.

Once this PR is merged the CSP violation in #4562 will be solved. See more here https://github.com/openstreetmap/openstreetmap-website/pull/4562#discussion_r1532380754

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -11,6 +11,6 @@
   <%= yield :head %>
   <%= yield :auto_discovery_link_tag %>
   <%= csrf_meta_tag %>
-  <%= csp_meta_tag %>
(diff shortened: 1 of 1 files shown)


tomhughes:
Can we drop the filename from the commit comment please? We can tell what files a commit affects without repeating it in the comment - the will also avoid the need to abbreviate `with`.

grekko:
@tomhughes I've updated the commit message

tomhughes:
Looks good, thanks.

PR #4608, opened 2024-03-21
Simplify links on diary pages

AntonKhorev:
Similar to `link_to`/`*_path` simplifications I did before.


tomhughes:
Looks good, thanks.

PR #4609, opened 2024-03-21
Simplify remaining link_to ... user_path in views

AntonKhorev:


tomhughes:
All looks good, thanks.

Issue #4588, opened 2024-03-18
Are /api/0.6/notes/addPOIexec, closePOIexec etc api entry point still necessary?


AntonKhorev:
### Problem

There's a number of api entry points for notes
https://github.com/openstreetmap/openstreetmap-website/blob/e9666aa673e9c69fd9391765ac3390853eeca344/config/routes.rb#L102-L106
that are duplicates of documented entry points.
They existed since openstreetbugs times
https://github.com/openstreetmap/openstreetmap-website/commit/e3b8b8933044104bb515d34c3d36851caba78ac7#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5

Are they still in use now? Can they be removed?

tomhughes:
They were added for compatibility with old clients but I have no idea if anything still uses them - checking will need a full list so I can check the logs so here it is:

* POST to api/0.6/notes/addPOIexec
* POST to api/0.6/notes/closePOIexec
* POST to api/0.6/notes/editPOIexec
* GET from api/0.6/notes/getGPX
* GET from api/0.6/notes/getRSSfeed

tomhughes:
None of those appears to have been used in the last month that we have logs for.

gravitystorm:
Closed by #4610

PR #4610, opened 2024-03-21
Remove old note api endpoints

AntonKhorev:
Remove endpoints listed in #4588.

- they are undocumented
- they don't do anything that documented endpoints can't
- [nobody seems to be using them](https://github.com/openstreetmap/openstreetmap-website/issues/4588#issuecomment-2004302883)
- you can't easily move their routes inside resources because they'll overwrite path helpers of documented endpoints

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -99,12 +99,6 @@ OpenStreetMap::Application.routes.draw do
       end
     end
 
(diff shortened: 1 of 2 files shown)


gravitystorm:
Merged, thanks!

PR #4612, opened 2024-03-21
Simplify and test downloaded changeset element sorting

AntonKhorev:
To find out which element order makes sense for #4571, I looked at how elements are sorted in changeset downloads. They are sorted by timestamp, then by version. Then I decided to simplify the code by replacing `sort`+ifs with `sort_by`, and I noticed that there are no tests for sort order.

Cgimap actually sorts further by type and id too: https://github.com/zerebubuth/openstreetmap-cgimap/blob/ea3e59543116fa97fd68e6d97813b1ee46e54290/src/osmchange_responder.cpp#L32. With `sort_by` it's easy to add this to osm-website if required.


tomhughes:
Looks good to me, thanks.

Issue #981, opened 2015-06-04
Clearer icon for Map Key


dkniffin:
I think we should rethink the icon used for the map key. The current one (lowercase i) gives the idea of "information", which is misleading. What kind of information does it display? Honestly, I didn't even know there was a key available, until I just hovered over it.

Something like this might be better:
![image](https://cloud.githubusercontent.com/assets/1680080/7988446/b96728d4-0aa0-11e5-85fe-7f8f9750edc3.png)
I also think we should avoid an actual key icon, as it implies authentication, etc.

matkoniecz:
> What kind of information does it display?

How map presents information.

> as it implies authentication

And it would be clear only for limited audience (in many languages name for map key is not related in any way to word key).

> Something like this might be better:

Would it work using only two colours like current icon?

dkniffin:
Sorry I never responded to this a year ago. I still think changing this icon is a good idea. I do think it works as a two-color icon, such as:

![image](https://cloud.githubusercontent.com/assets/1680080/19315777/0ad8745e-906d-11e6-93a3-e6a942f66722.png)

danieldegroot2:
@gravitystorm Requesting to label this issue at least as `ui` to make it easier to find.

gravitystorm:
@danieldegroot2 I've added the labels that you requested. If you are finding the labels useful, then I can also give you "triage" permissions to add labels yourself - just let me know and I'll do so. We've already got two other people with "triage" permissions who add labels from time to time.

hlfan:
> gives the idea of "information", which is misleading

And the map key isn't information about how data is presented on the map?

PR #4616, opened 2024-03-23
Rename browse controller to queries controller

AntonKhorev:
The only thing left in browse controller was the query sidebar.


tomhughes:
I fine with this in principle, I just wonder what the right name is and I'm not sure that `QueriesController` is it.

I started off thinking it should be singular as `QueryController` which I realise doesn't fit with resourceful naming from using a resource route but then maybe that's the point, that it's not managing a set of resources so shouldn't be a resourceful route?

Then I thought maybe it should be `FeaturesController` because features are the resource that is being queried? At least that's how the UI describes it...

AntonKhorev:
What's wrong with `QueriesController`? When you see this controller name, you know it handles either `/queries/:id` or `/query` paths. `FeaturesController` instead you'll have to investigate because there's no "features" route.

If we change the path however... I'm thinking about combining *query features* with *map key*. There's a proposal to change [the map key icon](https://github.com/openstreetmap/openstreetmap-website/issues/981), but I don't think users will find it significantly more clear. The query button might open the sidebar right away and show *map key* there. And also sort out #3819, #3833 and #3709.

AntonKhorev:
Superseded by #5751

Issue #2399, opened 2019-10-16
web_timeout needs to handle layout=false


gravitystorm:
We surround certain web requests with a timeout handler. (API requests are handled separately). However, it returns a full web page with layout, which can lead to unexpected results when an html fragment was expected.

 For example, `app/controllers/changesets_controller.rb#index` normally does a 
```ruby
render :action => :index, :layout => false
```

But when it times out you get a mess like this:

![Screenshot from 2019-10-16 12-13-34](https://user-images.githubusercontent.com/360803/66922716-e5b8e000-f027-11e9-9311-cd79c0595af7.png)

It might be possible to solve with refactoring in the controllers, so e.g. using 'layout false' at the  class level instead of on individual `render` calls.

danieldegroot2:
More details at https://github.com/openstreetmap/openstreetmap-website/issues/3883#issuecomment-1380073000

PR #4617, opened 2024-03-23
Fix timeout messages for former browse pages

AntonKhorev:
Probably was broken for notes for a long time and then I broke some more while dismantling browse controller.


tomhughes:
Is using a negative timeout going to work reliably I wonder?

I did try and write some tests for timeouts myself recently but struggled to get reliable tests though I was specifically trying to get a timeout to happen during a database query which adds extra complication.

mmd-osm:
Would any database query do? Maybe something like `select pg_sleep(5);` could be used to have a bit more deterministic query runtime?

tomhughes:
Yes that was exactly what I experimented with at the time ;-)

tomhughes:
Sure what concerned me is whether that timeout is guaranteed to fire no matter how quick the action is - the timeout works by having a background thread signal the current thread when the time expires so there's no absolute guarantee of it expiring immediately.

AntonKhorev:
If it doesn't fire for some reason, things like `select pg_sleep(5);` are still not going to be helpful. I'd need to inject this query somehow, but why won't I simply inject a timeout exception?

AntonKhorev:
Which status code should timeouts return? This one https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504 ?

app/controllers/application_controller.rb
```diff
@@ -206,13 +206,13 @@ def web_timeout(&block)
     if e.is_a?(Timeout::Error) ||
        (e.is_a?(ActiveRecord::StatementInvalid) && e.message.include?("execution expired"))
       ActiveRecord::Base.connection.raw_connection.cancel
-      render :action => "timeout"
+      render "browse/timeout", :layout => map_layout
```
tomhughes: I know you change it again later but I don't think it's reasonable to have this as an intermediate step as it leaves the site broken - the while point of only setting an action was to render the timeout action for the current controller because timeouts can occur in different controllers.

I think this needs more though about how to do this change and the following one to ensure we keep things roughly working through each step.

app/views/changesets/timeout.atom.builder
```diff
@@ -1,6 +1,6 @@
 atom_feed(:language => I18n.locale, :schema_date => 2009,
-          :id => url_for(params.merge(:only_path => false)),
-          :root_url => url_for(params.merge(:only_path => false, :format => nil)),
+          :id => url_for(params.permit(:format, :display_name, :bbox, :friends, :nearby).merge(:only_path => false)),
+          :root_url => url_for(params.permit(:format, :display_name, :bbox, :friends, :nearby).merge(:action => :index, :format => nil, :only_path => false)),
```
tomhughes, line 3: We should probably factor out the `params.permit` into a variable to avoid duplicating it?

AntonKhorev:
related #2399

app/controllers/application_controller.rb
AntonKhorev: The site is broken before the first step. How can it work through each step unless you squash all of the steps?

app/controllers/notes_controller.rb
```diff
@@ -11,7 +11,7 @@ class NotesController < ApplicationController
 
   before_action :lookup_user, :only => [:index]
   before_action :set_locale
-  around_action :web_timeout
+  around_action :web_timeout, :only => :show
```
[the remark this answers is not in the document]
AntonKhorev: There was no code for timeouts on non-map layout pages. https://github.com/openstreetmap/openstreetmap-website/commit/a7f0de41e65f90c8a20e1a219a5b7ecf498628a7 enabled timeouts here without defining any timeout templates.

AntonKhorev:
> We've always used [408 Request Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) for timeouts as far as I know.

Isn't this code for clients not using the connection for too long? That's not what happening here.

app/controllers/application_controller.rb
AntonKhorev: Removed temporary `render "browse/timeout"`.

app/controllers/notes_controller.rb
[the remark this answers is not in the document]
AntonKhorev: Moved notes-related changes to #4647.

app/views/changesets/timeout.atom.builder
AntonKhorev: Added a variable.

I didn't try to write clean `params.permit` code because some dubious code already exist. I was going to fix it later.

app/controllers/application_controller.rb
tomhughes: There are two controllers (changesets and changeset_comments) which have a timeout view already and with this change they will now render the browse timeout instead of their own one if they timeout? That was what I meant by breaking things that currently worked.

app/controllers/notes_controller.rb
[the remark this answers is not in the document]
tomhughes: Just because there wasn't a template defined doesn't mean the timeout didn't work and protect the site - it just means the user got a bad response when the timeout happened!

PR #4619, opened 2024-03-24
Replace collection partial with single item in changeset comments feed

AntonKhorev:


tomhughes:
All looks good, thanks.

PR #4621, opened 2024-03-25
Replace lookup_user_by_id with load_resource

AntonKhorev:
`.find(params[:id])` is exactly what CanCanCan's `load_resource` does:
https://github.com/CanCanCommunity/cancancan/blob/develop/docs/controller_helpers.md#authorize_resource-load_resource-load_and_authorize_resource


tomhughes:
Nice find!

PR #4625, opened 2024-03-26
Use newer/older pagination on admin users page

AntonKhorev:
image

This gets rid of classic pagination on this page. There's going to be only two pages with classic pagination after that: user blocks and changeset elements.

Differences to the previous version:

- Users are listed in the opposite order.
- Of course now there's no page numbers and it's impossible to switch to the nth page, but I don't know if this was ever used.
- Normally with before/after pagination there's no total count. Classic pagination had it, which might be useful when combined with some filter like ip address. I added the count here because there won't be any large traffic to this page.

### DIFF
--- a/app/controllers/concerns/pagination_methods.rb
+++ b/app/controllers/concerns/pagination_methods.rb
@@ -1,0 +1,27 @@
+module PaginationMethods
+  extend ActiveSupport::Concern
+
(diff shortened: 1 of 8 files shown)


tomhughes:
Thanks - this will actually be really useful for me!

PR #4627, opened 2024-03-26
Switch to using rails builtin content security policy support

tomhughes:
This replaces `secure_headers` with the rails builtin support for content security policy.

All other headers that `secure_headers` was setting are already set to the same values by rails with the exception of two which used to be the same but have now been changed:

* `X-XSS-Protection` was changed by https://github.com/rails/rails/pull/41769
* `X-Download-Options` is no longer set since https://github.com/rails/rails/pull/43968

The only slight annoyance is that rails doesn't make appending (as against overriding) very easy and doesn't have a good way to make changes dynamically within an action - that only really affects `map_layout`.


app/controllers/export_controller.rb
```diff
@@ -4,6 +4,10 @@ class ExportController < ApplicationController
   before_action :update_totp, :only => [:finish]
   authorize_resource :class => false
 
+  content_security_policy do |policy|
```
gravitystorm, line 7: Should this be `content_security_policy(:only => :embed) do |policy|`?

gravitystorm (changes requested):
Looks good to me, just one small question inline. 

Also needs rebasing.

app/controllers/export_controller.rb
tomhughes: Yes indeed, now fixed.

tomhughes:
Rebased and comment addressed.

gravitystorm:
Great, thank you.

PR #4628, opened 2024-03-27
Use "visible" scope when finding traces

AntonKhorev:
Allows to eliminate some :not_found branches.

### DIFF
--- a/app/controllers/traces_controller.rb
+++ b/app/controllers/traces_controller.rb
@@ -68,10 +68,9 @@ class TracesController < ApplicationController
   end
 
   def show
-    @trace = Trace.find(params[:id])
+    @trace = Trace.visible.find(params[:id])
 
-    if @trace&.visible? &&
-       (@trace&.public? || @trace&.user == current_user)
+    if @trace.public? || @trace.user == current_user
       @title = t ".title", :name => @trace.name
     else
       flash[:error] = t ".trace_not_found"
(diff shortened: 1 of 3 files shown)


app/controllers/traces_controller.rb
```diff
@@ -68,10 +68,9 @@ def index
   end
 
   def show
-    @trace = Trace.find(params[:id])
+    @trace = Trace.visible.find(params[:id])
 
-    if @trace&.visible? &&
-       (@trace&.public? || @trace&.user == current_user)
+    if @trace&.public? || @trace&.user == current_user
```
tomhughes, line 73: I realise it's not a change from this PR but I don't think we need the conditional dereference here do we? The `find` will have raised an exception if nothing was found...

tomhughes (commented):
Also this could be extended to the icon and picture actions, or rather to the show actions in the new controllers. Those need to check `inserted` as well which could either be done with a where or with a new scope on trace.

app/controllers/traces_controller.rb
AntonKhorev: replaced `&.` with `.` here

AntonKhorev:
Added `Trace.visible.find` to icon/picture controllers. Later I want to move this visibility check to CanCanCan ability conditions.

tomhughes:
Looks good now, thanks.

PR #4632, opened 2024-03-28
Remove unused includes of ConsistencyValidations

AntonKhorev:
Why is ConsistencyValidations a model concern? It raises API exceptions.

### DIFF
--- a/app/controllers/changesets_controller.rb
+++ b/app/controllers/changesets_controller.rb
@@ -16,9 +16,6 @@ class ChangesetsController < ApplicationController
 
   around_action :web_timeout
 
-  # Helper methods for checking consistency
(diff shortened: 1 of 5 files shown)


tomhughes:
It's only `check_changeset_consistency` that is used in controllers - the other two methods are used in models so maybe it should be split? Though I'm not sure you could use the same name for both given that controllers can apparently find model validations, which rather surprised me.

Then again `check_changeset_consistency` is only used in one controller so is there any point in splitting it out?

This PR looks good anyway.

AntonKhorev:
> `check_changeset_consistency` is only used in one controller

Actually it's used in all other consistency checks. It's copy-pasted there with slight modifications. See #4640.

PR #4633, opened 2024-03-28
Trace image cleanups

tomhughes:
This further simplifies the trace image controllers by adding an `imported` scope to the trace model which eliminates a duplicate "not found" branch.

It also drops some redundant code for handling legacy traces that should have been removed in #3491.


gravitystorm:
LGTM, thanks

PR #4635, opened 2024-03-29
Remove unused user parameter for subscribe/unsubscribe headings

AntonKhorev:
I was experimenting with subscribe/unsubscribe to diary entry page headings. I decided not to put username there but forgot to remove it from translation arguments.


tomhughes:
Looks good, thanks.

PR #4636, opened 2024-03-29
Redirect from old diary comments pagination

AntonKhorev:
Like was done in https://github.com/openstreetmap/openstreetmap-website/commit/c909b29c35b32be6f918f4393692033b40c303ea for trace pagination.

`/user/username/diary/comments/123` -> `/user/username/diary/comments` because page numbers no longer exist


tomhughes:
Looks good, thanks.

PR #4637, opened 2024-03-29
Use before/after pagination on user blocks page

AntonKhorev:
image

This changes the order of blocks in the table, but:

- The current order (by end date desc) is not very useful, because the beginning of the list is dominated by long-term blocks. You'll have to page through a lot of "Ends in almost 10 years" blocks for throwaway vandalism accounts before you reach "real" blocks for people to pay attention to, those are not 10 years right away.
- There are plans to let the table be reordered, see #4200.

### DIFF
--- a/app/controllers/user_blocks_controller.rb
+++ b/app/controllers/user_blocks_controller.rb
@@ -1,5 +1,6 @@
 class UserBlocksController < ApplicationController
   include UserMethods
+  include PaginationMethods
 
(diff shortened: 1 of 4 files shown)


tomhughes:
So the new order is just by the block ID but sorting in descending order which effectively puts the most recent blocks first which is probably as useful as anything any certainly more useful than the current order as you say.

PR #4638, opened 2024-03-30
Rename api changeset show template

AntonKhorev:
Currently there are very similarly named template files for api changeset views:
- changeset.xml.builder
- _changeset.xml.builder
- changesets.xml.builder

Usually there's a show template for other controllers. Here I'm renaming `changeset.xml.builder` to `show.xml.builder` and the same for json.


tomhughes:
Looks like a reasonable change to me, thanks.

PR #4640, opened 2024-03-30
Refactor element consistency checks

AntonKhorev:
There are two sets of checks:
- old-new element consistency
- changeset-user consistency

Either one of them or both of them are performed in different situations.

### DIFF
--- a/test/models/node_test.rb
+++ b/test/models/node_test.rb
@@ -435,4 +435,13 @@ class NodeTest < ActiveSupport::TestCase
       node.update_from(new_node, user)
     end
   end
+
+  test "raises id precondition exception when deleting" do
+    user = create(:user)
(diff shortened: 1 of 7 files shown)


test/models/node_test.rb
```diff
@@ -362,4 +362,86 @@ def test_containing_relations
     assert_equal relation_member2.relation.id, cr.second.id
     assert_equal relation_member3.relation.id, cr.third.id
   end
+
+  test "raises missing changeset exception when creating" do
```
tomhughes, line 366: We should really be consistent and stick to using `do test_` or else change all the existing test definitions but I see some using this style have already snuck in without being noticed so I guess a few more won't hurt :-(

tomhughes:
Looks like a nice improvement, thanks.

PR #4641, opened 2024-04-02
Simplify api element paths in tests

AntonKhorev:
Like #4591 etc.

One goal was to get rid of explicit `id` parameters because if I make old elements subresources of current elements, rails would want `node_id`/`way_id`/`relation_id` instead of `id`. Old element models already have these fields.


tomhughes:
Looks good to me, thanks.

PR #4642, opened 2024-04-02
Rename api OldController to OldElementsController

AntonKhorev:
#3715 adds `ElementsController`. When I originally did it, I didn't notice that a similar superclass already exists for old elements because it's not named after elements.

The name "OldController" is not specific enough because there are other "old" things besides old elements like old tags or old relation members.


tomhughes:
Looks good to me, thanks.

PR #4643, opened 2024-04-02
Move edit/revoke block buttons to the bottom of the block page

AntonKhorev:
Currently it's difficult to go between all blocks, blocks on, blocks by pages. For example, it's not obvious where's the link to the *all blocks* page. I want to make some kind of navigation for it. Something like tabs similar to how they are used on gpx traces pages will probably do:
image

But first I need to make room for tabs. Block pages have *Edit* and *Revoke* links in the heading where the tabs are supposed to be:
image

But why these links are there? If they are supposed to work on a currently shown block, they should be below the block, again similar to how *Edit*/*Delete* buttons are located on traces pages.

This PR moves the links and styles them similarly to traces buttons:
image

Later I'll turn *View all blocks* into a tab and add other tabs.

### DIFF
--- a/app/views/user_blocks/revoke.html.erb
+++ b/app/views/user_blocks/revoke.html.erb
@@ -21,7 +21,7 @@
       </div>
     </div>
 
-    <%= f.primary t(".revoke") %>
+    <%= f.submit t(".revoke"), :class => "btn btn-danger" %>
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good to me, thanks.

PR #4644, opened 2024-04-03
Add navigation tabs to block list/show pages

AntonKhorev:
image
image

There's a potential problem though. Checks if a user has created any blocks are going to be made more often with this navigation, but there's no `creator_id` index on the blocks table. The options are:
- Do nothing because blocks pages are not the most visited and the blocks table is not the most populated.
- Remove `.blocks_created.exists?` checks. That would show more unnecessary and unexpected tabs: blocks by regular users but they can't make any blocks.
- Check if a user can create blocks instead of checking if they have created any. That would remove tabs for blocks by former moderators.
- Add a `creator_id` index on the blocks table.
- Add a `(creator_id, id)` index on the blocks table that will also help with sorting.

### DIFF
--- a/app/views/user_blocks/_navigation.html.erb
+++ b/app/views/user_blocks/_navigation.html.erb
@@ -26,4 +26,12 @@
                   :class => ["nav-link", { :active => action_name == "blocks_by" && current_user == @user }] %>
     </li>
   <% end %>
+  <% by_user = @user || @user_block&.creator %>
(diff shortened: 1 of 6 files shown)


tomhughes:
Oh that's because there is no active tab then... It's just a bit confusing.

AntonKhorev:
I haven't decided what to call it. "Block #n" maybe because putting two usernames "x blocked by y" is too long?

PR #4645, opened 2024-04-03
Take rss/upload buttons out of trace nav tabs list

AntonKhorev:
*** this works, but I want to try another thing first ***

---

Here's the trace navigation:
image

What if we make the window smaller? It won't look as good because the active tab is detached from the page contents:
image

After the fix in this PR:
image

---

image

After:
image

https://github.com/openstreetmap/openstreetmap-website/commit/81ad7154747244abfa2e38dc529e62e6bd8998bd


tomhughes:
Thanks - that looks much better!

PR #4646, opened 2024-04-03
Use turbo for pagination

tomhughes:
This extends turbo usage to everything that uses the shared pagination helpers, namely:

* Diaries
* User diary comments view
* Traces
* Administrator user listing
* User blocks

### DIFF
--- a/app/views/diary_entries/comments.html.erb
+++ b/app/views/diary_entries/comments.html.erb
@@ -7,28 +7,30 @@
   <h4><%= t ".no_comments" %></h4>
 
 <% else %>
-  <table class="table table-striped" width="100%">
-    <thead>
+  <turbo-frame id="pagination" target="_top">
+    <table class="table table-striped" width="100%">
+      <thead>
+        <tr>
+          <th width="25%"><%= t ".post" %></th>
+          <th width="25%"><%= t ".when" %></th>
+          <th width="50%"><%= t ".comment" %></th>
+        </tr>
+      </thead>
+      <% @comments.each do |comment| -%>
       <tr>
-        <th width="25%"><%= t ".post" %></th>
-        <th width="25%"><%= t ".when" %></th>
-        <th width="50%"><%= t ".comment" %></th>
+        <td width="25%" class="<%= "text-muted" unless comment.visible? %>"><%= link_to comment.diary_entry.title, diary_entry_path(comment.diary_entry.user, comment.diary_entry) %></td>
+        <td width="25%" class="<%= "text-muted" unless comment.visible? %>">
+          <%= friendly_date_ago(comment.created_at) %>
+        </td>
+        <td width="50%" class="richtext text-break<%= " text-muted" unless comment.visible? %>"><%= comment.body.to_html %></td>
       </tr>
-    </thead>
-    <% @comments.each do |comment| -%>
-    <tr>
-      <td width="25%" class="<%= "text-muted" unless comment.visible? %>"><%= link_to comment.diary_entry.title, diary_entry_path(comment.diary_entry.user, comment.diary_entry) %></td>
-      <td width="25%" class="<%= "text-muted" unless comment.visible? %>">
-        <%= friendly_date_ago(comment.created_at) %>
-      </td>
-      <td width="50%" class="richtext text-break<%= " text-muted" unless comment.visible? %>"><%= comment.body.to_html %></td>
-    </tr>
-    <% end -%>
-  </table>
+      <% end -%>
+    </table>
 
-  <%= render "shared/pagination",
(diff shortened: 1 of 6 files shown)


nertc:
For some reason, `:data => { :turbo => { :frame => "pagination"` doesn't work. Turbo frame still behaves according to `target="_top"` and refreshes whole page instead of just updating `turbo-frame` element. I tried adding `"data-turbo-frame" => "pagination"` to the links and it started working properly. Maybe Turbo can't read frame parameter from `data-turbo="{"frame":"pagination","action":"advance"}"` attribute or it has lower priority index than `target` attribute.

nertc:
With the current solution, back-end side still renders a whole page including header and other elements, that won't be updated (see screenshot). For example, generating a header requires ~1s extra time and when user uses pagination, in some cases, ignoring everything that won't be rendered in the `turbo-frame` may give us 2-10x faster calls. In addition to this, people who use application from mobile phones, spend more data than it is needed. Turbo Streams are great tool for solving this, or we can use API endpoints that render only partial which should be rendered in the `turbo-frame` element (but both ways may need a small amount of JS code). I have the same problem in https://github.com/openstreetmap/openstreetmap-website/pull/5057. I think it will be great if we have one standard about using `turbo-frames`.

Screenshot:

image

AntonKhorev:
You don't need extra js code to render/update only the target frame. If you want to optimize server-side rendering, you can check which frame was requested like this:

https://github.com/openstreetmap/openstreetmap-website/blob/5add2d7e1d1d365e80d1d4ef417ff249ca7f35c0/app/controllers/changesets_controller.rb#L79

tomhughes:
Yes I know this works by rendering the whole page and sending the changed part - it's a starting point that does the easy things and which can be improved on in the future.

tomhughes:
I've updated this to use `turbo_frame_request_id` to limit what is rendered on a turbo navigation.

app/views/diary_comments/_page.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    </tr>
+    <% end -%>
+  </table>
+
+  <%= render "shared/pagination",
+             :newer_key => "diary_comments.page.newer_comments",
+             :older_key => "diary_comments.page.older_comments",
+             :newer_id => @newer_comments_id,
+             :older_id => @older_comments_id %>
+</frame>
```
nertc, line 26: `turbo-frame` tag was opened, but `frame` tag was closed. Instead of `</frame>` should be `</turbo-frame>`

app/views/traces/_page.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <%= render @traces %>
+    </tbody>
+  </table>
+
+  <%= render "shared/pagination",
+             :newer_key => "traces.page.newer",
+             :older_key => "traces.page.older",
+             :newer_id => @newer_traces_id,
+             :older_id => @older_traces_id %>
+</turbo-frame id="pagination">
```
nertc, line 19: Closing tag doesn't need `id`. Instead of `</turbo-frame id="pagination">` should be `</turbo-frame>`

nertc (changes requested):
The change is functional, only the needed parts are rendered, and pagination works properly with history advancing and optimizations. After these two lines will be corrected, I think, this PR will be ready for merging.

app/views/traces/_page.html.erb
tomhughes: Thanks for spotting that - it should be fixed now.

nertc (approved):
Thank you. Everything is fixed. This PR is ready to be merged.

AntonKhorev:
Needs to be rebased after #5095 remerging.

tomhughes:
Rebased...

AntonKhorev:
Do we merge this or wait for @gravitystorm to review?

mmd-osm:
I'm testing different user supplied values for before and after. Both `/diary?before=0` and `/user/{user}/diary?after=0` trigger an HTTP 400 Bad request. Is this expected, or should the result be an empty list, of a list of diaries starting with the very first one?

tomhughes:
That comes from #4680 because before that people were passing all sorts of silly values which were causing exceptions and triggering error rate alerts.

mmd-osm:
Ok, sounds good. I'm not expecting normal users to tweak these values, so this is really stopping people using silly values.

AntonKhorev:
`after=0` is useful to get the earliest entries

gravitystorm:
This looks great, and I'm surprised how well it works with so little custom code.

Merged, thanks! 

And thanks to @nertc @AntonKhorev and @mmd-osm for helping review this too.

PR #4647, opened 2024-04-04
Fix timeout on show note pages

AntonKhorev:
Also disables timeout on index pages that show user's notes because:
- note timeouts never worked correctly
- no working non-map layout timeouts were implemented on the website

I might add index timeout later, but that wouldn't be fixing things, that would be implementing something new. 

Taken out of #4617 to keep that PR about fixing things that used to work.


tomhughes:
> I might add index timeout later, but that wouldn't be fixing things, that would be implementing something new.

I don't follow the logic here... You're adding `:only` so previous there would have been a timeout around the index action surely? It might not have rendered a nice error when it triggered but it would have protected the server from locking up a daemon for a long time while it processed the action surely which is the most important thing.

AntonKhorev:
The logic is so https://github.com/openstreetmap/openstreetmap-website/pull/4617#discussion_r1550144840 is no longer a reason to reject #4617. Notes timeouts could be fixed later.

PR #4652, opened 2024-04-06
Adapt striped table override to dark mode

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/issues/2332#issuecomment-1728044333 for how striped tables currently look in dark mode.

This change keeps the current "striped" color in light mode, while in dark mode it looks like this:
image

### DIFF
--- a/app/assets/stylesheets/parameters.scss
+++ b/app/assets/stylesheets/parameters.scss
@@ -21,6 +21,6 @@ $link-hover-color: #24d;
 $link-decoration: none;
 $link-hover-decoration: underline;
 
-$table-striped-bg: $offwhite;
+$table-striped-bg: rgba(#77f, 0.05);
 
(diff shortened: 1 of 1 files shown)


tomhughes:
Can you explain _why_ this works? It's changing from a [pale grey](https://color-hex.org/color/f8f8ff) to a [light purple](https://color-hex.org/color/7777ff) but also using an alpha channel to make it mostly transparent?

Out of interest I looked at the default colours - that would give us this:

image

instead of this:

image

which I personally think is fine except possible that it makes the striping very close to the header colour - probably it's the secondary colour so would in fact be the same as the heading if we switch that to secondary in #4651.

AntonKhorev:
One thing I didn't like about the default stripe color is that it's darker than bg-body-tertiary I tried to use for the heading in #4651.

PR #4653, opened 2024-04-06
Enable Bootstrap dark theme if user prefers dark color scheme

AntonKhorev:
Adapted from https://getbootstrap.com/docs/5.3/customize/color-modes/#javascript. I don't add any mode selector, the theme is set to system color mode.

Currently doesn't work with a lot of pages that set custom colors.


AntonKhorev:
superseded by #4761

PR #4654, opened 2024-04-06
Replace border-grey css class with border-secondary-subtle

AntonKhorev:
No noticeable difference (#cccccc vs #cfcfcf) and no problems with dark mode.


tomhughes:
Looks good to me.

PR #4658, opened 2024-04-07
Update heading, stripes and tab border colors

AntonKhorev:
Have to make these changes at the same time to see if they work together:
- `bg-body-secondary` in heading sections instead of #4651
- darker tab borders when they are inside `bg-body-secondary`
- default striped tables instead of #4652

image

This is how it looks in dark mode:
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1063,4 +1063,11 @@ div.secondary-actions {
   }
 }
 
+/* Rules for tabs inside secondary background sections */
+
+.bg-body-secondary .nav-tabs {
+  --bs-border-color: var(--bs-secondary-border-subtle);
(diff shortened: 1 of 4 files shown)


tomhughes:
This mostly looks good to me.

I think the only issue is that there's a visible line under the tabs - it's a 1px border on the bottom of the tabs control but it looks odd because it only covers the middle section and not the margins.

It was actually there before, just much harder to see with the old colours....

tomhughes:
Thanks - that looks really good now and I actually quite like having that subtle border under the heading.

AntonKhorev:
Here's the easiest change - add border to every heading section + negative margin to tabs:

image

I can change negative margin to `calc(-1 * var(--bs-border-width))` to avoid pixel widths in css.

Other options are:
- same but with top border for every heading too
- bottom border only for headings with tabs

PR #4659, opened 2024-04-08
Add block tab to show block pages

AntonKhorev:
To have an active tab on a block page - see https://github.com/openstreetmap/openstreetmap-website/pull/4644#issuecomment-2041438566

image


tomhughes:
Looks good to me, thanks.

PR #4661, opened 2024-04-08
Use secondary text color in header menu

AntonKhorev:
Uses `text-secondary` or `text-secondary-emphasis` if it's a current page.
Replaces: 
- `$darkgrey`: `text-secondary` matches it
- `darken($darkgrey, 25%)`: in dark mode it shouldn't be darkened

Before:
image

After:
image

I also removed the current page color change in the "more" dropdown because it worked in the opposite direction, making the text lighter. Also there's no similar color change in the user menu dropdown.

Before, with "Communities" as the current page:
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -150,11 +150,6 @@ nav.primary {
 nav.secondary {
   .nav-link {
     padding: 0.2rem;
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me... I struggled to even see the colour variation in the drop down menu!

PR #4663, opened 2024-04-08
Invert routing direction icons in dark mode

AntonKhorev:
image

image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -575,6 +575,12 @@ div.direction {
 div.direction.i#{$i} { background-position: #{($i)*-20}px 0px; }
 }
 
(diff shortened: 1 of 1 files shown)


AntonKhorev:
Either merge in #4653, then enabling dark mode in Firefox is going to work.

Or manually add `data-bs-theme="dark"` to the root `<html>` element.

tordans:
Should we add a user profile setting that overwrites the system preferences and allows testing? This would also be a feature flag for testing. We could deploy that disabled so more users can test live over time...

AntonKhorev:
```js
document.documentElement.dataset.bsTheme='dark'
```

tomhughes:
Looks good to me, thanks.

PR #4667, opened 2024-04-08
Support dark mode in header

AntonKhorev:
image
image

image
image

image
image


tordans:
A minor thing here is that there user icon in dark mode has less visible space to the user name. Maybe we should add space and make the background (more) visible in light mode as well?

tomhughes:
Looks good to me, thanks.

PR #4668, opened 2024-04-08
Adapt About page backgrounds to dark mode

AntonKhorev:
I made the "OpenStreetMap provides map data" section lighter in light mode and darker in dark mode. The previous color, `bg-dark`, is exactly `bg-body` in dark mode. If kept, there would be no visible separation with the next section.

image
image


tomhughes:
Looks OK to me, thanks.

PR #4669, opened 2024-04-08
Adapt tags table to dark mode

AntonKhorev:
image
image


tomhughes:
This looks like it is changing from the way you're supposed to customise tables in bootstrap to a bodging things with a background class which I don't really like.

I see https://getbootstrap.com/docs/5.3/content/tables/ says that tables probably won't get colour mode adaptive styling until v6 which I assume is the problem? That dark mode is a colour mod adaptive styling?

tomhughes:
Superceded by #4700.

PR #4671, opened 2024-04-09
Use svg default avatar image on webpages

AntonKhorev:
The svg file has a transparent background, which should help in dark mode by not being white and with a perception that [the space between the image and other things became smaller](https://github.com/openstreetmap/openstreetmap-website/pull/4667#issuecomment-2044173704).

image

with header PR merged:
image

`avatar_small.png` and `avatar_large.png` are still used in:
- emails
- Gravatar default image. Their docs don't say if svgs are supported.


AntonKhorev:
You already can upload an avatar image with transparency and see the grey background:

image

I don't know if it's wanted or unwanted.

Do you want to force background to `bg-body` under every avatar image?

image

tomhughes:
I think we probably should yes - better to change the behaviour in the rare case of a user avatar with a transparent background which was never deliberate that to change the entirely deliberate missing avatar case.

AntonKhorev:
Updated to use `bg-body` background:
image
image

PR #4672, opened 2024-04-09
Adapt notes table to dark mode

AntonKhorev:
image

### DIFF
--- a/app/views/notes/index.html.erb
+++ b/app/views/notes/index.html.erb
@@ -1,9 +1,11 @@
+<% submitted_class = "bg-primary bg-opacity-25" %>
+
 <% content_for :heading do %>
   <h1><%= t ".heading", :user => @user.display_name %></h1>
   <p><%= t ".subheading_html",
            :user => link_to(@user.display_name, @user),
-           :submitted => tag.span(t(".subheading_submitted"), :class => "px-2 py-1 bg-primary bg-opacity-25"),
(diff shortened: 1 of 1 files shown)


app/views/notes/index.html.erb
```diff
@@ -24,19 +26,20 @@
       </tr>
     </thead>
   <% @notes.each do |note| -%>
-    <tr<% if note.author == @user %> class="table-primary"<% end %>>
-      <td>
+    <% cell_class = submitted_class if note.author == @user -%>
+    <tr>
```
tordans, line 30: In https://getbootstrap.com/docs/4.0/content/tables/#contextual-classes it looks like it should be possible to keep this on the `tr` which means all the added complexity with `submitted_class` could go away again in favor of inline classes. 
Would something like … be possible? (The code is probably not right, just to illustrate)

```suggestion
    <%= tag.tr :class => ("bg-primary bg-opacity-25" if note.author == @user") do %>
       // td|s
    <% end %>
```

AntonKhorev: Their contextual classes don't change with light/dark theme.

Open the docs for the current version, there's a theme switch in the upper right corner
https://getbootstrap.com/docs/5.3/content/tables/#variants

tordans: I see, but what is it about our utility classes that don't work on `tr`? I just tried it and they don't, but I did not see why …

tomhughes:
Superseded by #4700.

PR #4674, opened 2024-04-10
Use Bootstrap toggle buttons for layers menu

AntonKhorev:
Rewrite of #3674 - layer buttons as [Bootstrap radio toggle buttons](https://getbootstrap.com/docs/5.3/forms/checks-radios/#radio-toggle-buttons)

Enables keyboard navigation for layer buttons which is missing now.

The only custom css is to:
- set button heights
- override borders to have hover effects because Bootstrap toggle buttons don't have any hover effect (they'd need another color to tell hover states apart from selected states, they solved it by removing hover states)

image
image


tomhughes:
Looks good to me, thanks.

PR #4675, opened 2024-04-10
Fix right sidebar background

AntonKhorev:
Fixes this:
image

I thought it wouldn't happen because sidebars are become full-width only in small layouts. But the right sidebar has its own logic when to go full width. It's possible to have an intermediate screen width which is both wide enough to have a floating search panel and narrow enough to have full-width right sidebar.

image


tomhughes:
Looks good - thanks for the quick fix.

PR #4676, opened 2024-04-10
Fix header background

AntonKhorev:
Not sure when the header background got broken:
image

It was working when I made #4667 with screenshots.

Testing again:
image
image


tomhughes:
Looks good - thanks for the quick fix.

PR #4679, opened 2024-04-11
Invert trace images in dark mode

AntonKhorev:
Do 85% invert because the page background is not 100% black.

image
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1036,4 +1036,12 @@ div.secondary-actions {
   margin-bottom: -1px;
 }
 
+/* Rules for traces */
+
+@include color-mode(dark) {
+  img.trace_image {
+    filter: invert(.85);
+  }
+}
+
 @import 'browse';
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/common.scss
```diff
@@ -1036,4 +1036,12 @@ div.secondary-actions {
   margin-bottom: -1px;
 }
 
+/* Rules for traces */
+
+@include color-mode(dark) {
+  img.trace_image {
```
tordans, line 1042: Should this maybe be a custom utility class? The utility class name would tell more about what it does. Something like 'dark-invert-image'.

HolgerJeromin: this name would be wrong is non-dark style.
Or do I miss something?

tordans: My point is not the suggested name but changing the name from a specific element name to a utility class name.

tomhughes:
Merged #4681 in preference to this.

PR #4680, opened 2024-04-11
Add parameter validation to pagination

tomhughes:
This adds validation to page/before/after/max_id parameters to various paginated requests so that we get a clean error instead of exception from the database when some nonsense value is fed into a query.


gravitystorm:
Thanks!

PR #4681, opened 2024-04-11
Invert trace images in dark mode + blend with background

AntonKhorev:
An alternative to #4679 which also changes light mode.

Trace images have white background which they probably shouldn't have, it should be transparent. Besides dark mode, this white background shows up in striped tables:

image

But we can make it behave as transparent by using blend modes:

image
image


tomhughes:
I think this is definitely the better option - fixes dark mode and improves the existing light more presentation!

PR #4682, opened 2024-04-11
Remove custom right sidebar section class

AntonKhorev:
Uses `border-secondary-subtle` if there's a border between section, like in #4654.

Dark mode before, the border is too bright:
image

Dark mode after:
image


tomhughes:
Looks good to me, thanks.

PR #4683, opened 2024-04-11
Use section elements instead of div.section on about page

AntonKhorev:
`section` class is not in use since https://github.com/openstreetmap/openstreetmap-website/commit/304deb3ca7adb17b412b0fda4fad10d73d0b90b6#diff-e86a39e6d0b163f152db8843fc7d647a3f5d648de4a74f1732bef4474b6e5bc8L2693-L2699

But those are actually sections and can be replaced with `<section>` elements.


tomhughes:
Looks good to me, thanks.

Issue #4665, opened 2024-04-08
Input type=number for export lat/lon inputs?


AntonKhorev:
### URL

https://www.openstreetmap.org/export

### How to reproduce the issue?

Switch to a rtl language and look at negative coordinates. That's probably not how they write negative numbers. Can be fixed by using `type=number` on inputs.

### Screenshot(s) or anything else?

I changed the left inputs to `type=number` for comparison:

![image](https://github.com/openstreetmap/openstreetmap-website/assets/4158490/e830bbb9-cec5-4375-8ee0-4d9ddd10bfa7)

In Persian the characters are different. Should we use them?
![image](https://github.com/openstreetmap/openstreetmap-website/assets/4158490/8a8eda43-0229-4638-a6c6-f5ffc67aa506)

HolgerJeromin:
I am not sure how relevant iOS mobile devices are at this page but at Mobile Safari is in the range of unusable with `type=number`.
Depending on the device locale we have a dot or comma (not changeable from the page with `lang=en` or so).
And (as a bonus) the onscreen keyboard has no possibility to enter minus!
Disclaimer: my iOS tests are two years old.

AntonKhorev:
Another way to fix this issue is to use `dir` attributes while keeping input type text. That's going to keep decimal dots and put minus signs on the left.

HolgerJeromin:
> Another way to fix this issue is to use dir attributes

That would be even more correct. I like that.

PR #4684, opened 2024-04-11
Force ltr direction in export inputs

AntonKhorev:
See #4665.

This PR keeps 0123456789 digits with decimal dot and makes `-` always appear on the left side.

Before:
image

After:
image


HolgerJeromin (approved):

tomhughes:
I guess that makes sense. Presumably even RTL languages put the minus before when using arabic numerals.

PR #4686, opened 2024-04-12
Use Bootstrap utilities to set z-indexes of map ui and fix some bugs

AntonKhorev:
## bug 1

Open the right sidebar, then open a dropdown menu ("more" or user menu):

image

Seems to work fine. Now make the screen smaller. If you get the intermediate width like in #4675, the dropdown menu will be covered by the sidebar:

image

Bootstrap sets z-index to 1000 on dropdowns which should be high enough, but we use even higher z-indexes on map ui elements. We do this to keep them above Leaflet's range of z-indexes. What we could do instead and what this PR does is to isolate Leaflet in its own [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context). Here I set the entire map at z-0 with a Bootstrap utility class. The right sidebar gets z-2 to stay above the floating left sidebar, which gets z-1.

image

## bug 2

Open the right sidebar on a small screen then try to open the menu:

image

The menu is a part of the header, with z-index below the right sidebar. In addition to the fixes above I set the header at z-3:

image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -90,7 +90,6 @@ time[title] {
 header {
   height: $headerHeight;
   position: relative;
-  z-index: 1001;
   font-size: 14px;
 
   h1, nav, nav > ul, nav > ul > li {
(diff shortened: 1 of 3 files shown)


tomhughes:
That looks much better! Thanks for the fix.

PR #4692, opened 2024-04-13
Remove unused $offwhite color

AntonKhorev:
It had no effect inside .inbox-row selector because table cells override the background.


tomhughes:
Looks good to me, thanks.

PR #4693, opened 2024-04-13
Adapt notes table to dark mode by overriding Bootstrap colors

AntonKhorev:
Alternative version of #4672 where we pretend that Bootstrap's `table-primary` supports dark mode.

`table-primary` is kept in the generated html. For notes table we only have to redefine a couple of colors in dark mode. In general this is not enough because there are variations for stripes, active colors etc but we're not using any of those here.

image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -799,6 +799,15 @@ tr.turn:hover {
   }
 }
 
+/* Rules for notes pages */
+
+@include color-mode(dark) {
+  .notes .table-primary td {
+    --bs-table-bg: var(--bs-primary-bg-subtle);
+    --bs-table-color: var(--bs-emphasis-color);
+  }
+}
+
 /* Rules for messages pages */
 
 .messages {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/common.scss
```diff
@@ -799,6 +799,15 @@ tr.turn:hover {
   }
 }
 
+/* Rules for notes pages */
+
+@include color-mode(dark) {
+  .notes .table-primary td {
```
tordans, line 805:
```suggestion
  .notes .table-primary-subtle td {
```

And in https://github.com/openstreetmap/openstreetmap-website/pull/4693/files#diff-4fe647349351c7492508c788230b91b7f4445d4b821f6781ac7fcfad5e879f21R27 also `table-primary-subtle`.

tordans (commented):
Visually, I like https://github.com/openstreetmap/openstreetmap-website/pull/4672 more. I also like more, that the changes are contained to the one view template and not split up into css+view. But then again, I am a big fan of utility class css like Tailwind…

If we go with this PR, I suggest to use the `sublte` term consistently and not overwrite "primary" with "primary-subtle".

app/assets/stylesheets/common.scss
AntonKhorev: There's no standard `table-primary-subtle` class. The background of `table-primary` happens to match `bg-primary-subtle` in light mode.

tomhughes:
Superseded by #4700.

PR #4694, opened 2024-04-13
Set maximum size limits for user provided content

mmd-osm:
I believe we should continue adding some upper size limits for various input fields, like we already do for [note comments](https://github.com/openstreetmap/openstreetmap-website/commit/d9a48d66f977afbe9d5e196e9e22d882b16b3566). 

Limits in this pull request are a best guess only, and I'm open to make them larger if needed. However, they shouldn't be unlimited as they are today. 

I'm aware that there are existing limits imposed by the web server, but I find them too large for input fields.

### DIFF
--- a/app/models/changeset_comment.rb
+++ b/app/models/changeset_comment.rb
@@ -30,7 +30,7 @@ class ChangesetComment < ApplicationRecord
   validates :changeset, :associated => true
   validates :author, :associated => true
   validates :visible, :inclusion => [true, false]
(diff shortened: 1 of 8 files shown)


AntonKhorev:
Error message when trying to add a comment that is too long:
image

mmd-osm:
For some reason, the three limits for comments don't seem to work yet. I don't get the error message above, instead  the comment is simply discarded. 

This should really behave like on the profile page:

image

PR #4695, opened 2024-04-13
Adapt notes table to dark mode by overriding Bootstrap colors

AntonKhorev:
Version 3 of #4672 and #4693 which slightly changes light mode colors but probably nobody is going to notice it.

`table-*` classes [happen to use "subtle" background colors](https://github.com/openstreetmap/openstreetmap-website/pull/4693#discussion_r1564038196) that look fine in light mode. They [don't adapt to dark mode](https://github.com/openstreetmap/openstreetmap-website/pull/4672#discussion_r1557742890) so we have to override them if we [still want to use them](https://github.com/openstreetmap/openstreetmap-website/pull/4669#issuecomment-2045718454). An obvious override is to set them to "subtle" colors in dark mode, that's what I tried in #4672 and it looks too dark. In this PR I'm doing the same thing I did in the legend abode the table and use the "non-subtle" color with 25% opacity.

image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -799,6 +799,13 @@ tr.turn:hover {
   }
 }
 
+/* Rules for notes pages */
+
+.notes .table-primary td {
+  --bs-table-bg: rgb(var(--bs-primary-rgb), .25);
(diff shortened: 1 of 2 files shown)


tomhughes:
Superseded by #4700.

PR #4696, opened 2024-04-13
Remove the arm pointing to the new user registration form

AntonKhorev:
There's a arm pointing to the new user form.
image

Unfortunately it has a shadow fading into white. This is going to stand out in dark mode.
image

The Microsoft redesign https://github.com/openstreetmap/openstreetmap-website/pull/4455 doesn't have the arm.
image

Instead of trying to fix the shadow, we might remove the arm even before switching to https://github.com/openstreetmap/openstreetmap-website/pull/4455.

image


tomhughes:
There's overlap with #4455 here, though I think that actually already removes the down pointing arm?

tordans:
> There's overlap with #4455 here, though I think that actually already removes the down pointing arm?

That is my understanding as well and I thing it would be better to focus on reviewing and merging #4455 and not add possible merge conflicts my cherry-picking changes.

AntonKhorev:
> There's overlap with https://github.com/openstreetmap/openstreetmap-website/pull/4455 here

Of course there is, that's why I'm making this pull request. The third image in the PR description is from https://github.com/openstreetmap/openstreetmap-website/pull/4455 in dark mode.

> it would be better to focus on reviewing and merging https://github.com/openstreetmap/openstreetmap-website/pull/4455 and not add possible merge conflicts

Or you can look at this PR as a [thing that can be treated separately](https://github.com/openstreetmap/openstreetmap-website/pull/4455#issuecomment-1994918240), a smaller part of https://github.com/openstreetmap/openstreetmap-website/pull/4455.

AntonKhorev:
This is still needed to remove the arm png.

AntonKhorev:
There's one svg file with both the entire globe and the arm layer. Editing that svg is tricky because it requires to have a specific font for 1's and 0's to look the same.

`common.scss` shouldn't have references to the arm image now.

tomhughes:
Ah yes my grep was a bit too generic.... Looks good then, thanks.

PR #4697, opened 2024-04-13
Remove highway=byway icon

AntonKhorev:
[Deprecated](https://community.openstreetmap.org/t/tracks-everywhere-or-how-to-draw-byways-bridleways-and-footpaths/51474/11), [not in use](https://taginfo.openstreetmap.org/tags/highway=byway#overview).


tomhughes:

PR #4698, opened 2024-04-13
Crop white background from browse highway icons

AntonKhorev:
Highway element icons on changeset pages sometimes have white background parts. They become visible in dark mode. This PR crops icons to remove those parts.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #4699, opened 2024-04-13
Darken + add transparency to counter badges

AntonKhorev:
We have a custom color for counter badges. But it's too bright for dark mode. It's too bright even for light mode because it's brightness is too close to the header background.

Light mode before:
image

Light mode after:
image

Dark mode before:
image

Dark mode after:
image


tomhughes:
Looks good to me, thanks.

PR #4700, opened 2024-04-14
Support dark mode in notes and tags tables

AntonKhorev:
Previous attempts:
- #4669
- #4672
- #4693
- #4695

Assumptions:
- We're committed to style tables with `table-*` classes as if Bootstrap supports dark mode.
- Because Bootstrap actually doesn't support dark mode with `table-*` we want some isolated fix that's going to be removed once the support is there.  This fix is in the changes to `common.scss`.
- We don't want classes like `table-light` because we don't want table cells to stay light in dark mode. But this class was used in tags table for tag keys, so it has to be replaced with some other `table-*` class. It's replaced here with `table-secondary` although this makes tag key cells darker in light mode.

---

Tags table in light mode before the changes:
image

Tags table in light mode after the changes:
image

Tags table in dark mode after the changes:
image

---

Notes table doesn't change in light mode.
Notes table in dark mode after the changes:
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -66,6 +66,21 @@ time[title] {
   color: $blue;
 }
 
+/* Bootstrap contextual table classes overrides in dark mode */
+
+@include color-mode(dark) {
+  .table-primary {
+    --bs-table-bg: rgb(var(--bs-primary-rgb), .25);
+  }
+  .table-secondary {
+    --bs-table-bg: rgb(var(--bs-secondary-rgb), .25);
+  }
+  .table-primary, .table-secondary {
+    --bs-table-color: initial;
+    border-color: inherit;
+  }
+}
+
 /* Rules for the header */
 
 #menu-icon {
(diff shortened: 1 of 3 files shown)


tomhughes:
I think this is a good solution - simple and easy to remove when bootstrap gets dark mode support for tables.

We don't actually know what `table-light` will wind up doing in dark mode but I think moving to `table-secondary` is fine as at least on the screen I'm looking at this on now it was hard to see the difference between the two columns and a bit more contrast is actually quite a good thing.

maro-21:
A [blue font color on a dark blue background](https://private-user-images.githubusercontent.com/4158490/322282583-f7c3c843-d7a3-46bb-96f9-318eb7d94e7c.png) is not a reasonable choice.
Please think about people with imperfect eyesight. There should be more contrast. Could you restore [the previous color](https://private-user-images.githubusercontent.com/4158490/322282661-0b2948db-acc0-4865-ad4c-cbe2fcd43aea.png)? What was wrong with it?

https://coolors.co/contrast-checker/2748dd-e7e7e7

AntonKhorev:
@maro-21 your images are 404

mmd-osm:
@maro-21 : none of the links you have posted are working here. Can you please try again? The contrast checker link wants to comapare "#FFFFFF" and "#FFFFFF", which seems a bit pointless.

PR #4701, opened 2024-04-14
Add white shadow to browse icons

AntonKhorev:
It's difficult to see some of the browse icons in dark mode. There are various ways to fix it. The simplest one I came up with is to add white drop shadows.

Before:
image

After:
image

Some icons look worse with it, for example "bar", but "place of worship" improves.

### DIFF
--- a/app/assets/stylesheets/browse.scss
+++ b/app/assets/stylesheets/browse.scss
@@ -8,6 +8,7 @@
     display: inline-block;
     width: 25px;
     margin-left: -25px;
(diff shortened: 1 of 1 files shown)


AntonKhorev:
Another option: increase brightness in dark mode

```
@include color-mode(dark) {
  .browse-section {
    .node::before,
    .way::before,
    .relation::before {
      mix-blend-mode: lighten;
    }
  }
}
```

image

PR #4702, opened 2024-04-15
Fix message table action cell css display property

AntonKhorev:
Fixes https://github.com/AntonKhorev/openstreetmap-website/commit/8f946f13607c06df101d26b2b47087b342a29a45#commitcomment-140768884

Before:
image

After:
image

PR #4703, opened 2024-04-15
Adapt messages table to dark mode

AntonKhorev:
- I remove the custom *unread message* background and replace it with `table-success`. The custom background was applied directly to table cells which we presumably don't want. This changes the green color used in the table.
- I extend dark mode overrides to cover `table-success`.
- I remove the `inbox-row` class. It was also used in the outbox.

Light mode before:
image

Light mode after:
image

Dark mode after:
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -810,14 +810,6 @@ tr.turn {
   }
 }
 
-/* Rules for messages pages */
-
-.messages {
-  .inbox-row-unread td {
-    background: #CBEEA7;
-  }
-}
-
 /* Rules for user images */
 
 img.user_image {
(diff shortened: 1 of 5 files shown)


tordans:
Side node @ Tom: When reviewing this, you can probably review https://github.com/openstreetmap/openstreetmap-website/pull/4702 as the same time, which changes the same page AFAIK.

tordans:
Follow up: I think from an accessibility point of view we should consider encoding the information that we (only) present via background colors with text/html parts. Some ideas…
- add hidden columns for screen reader only that have text describing the "read/unread" state
- use some aria property on the row to provide this read/unread status https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes

tomhughes:
I'm not sure about this one - the green is a bit pale for my tastes but I could probably live with that but what I really find a bit odd is the way the highlighted rows get a bolder lower border because `table-success` also changes the border colour.

AntonKhorev:
Lowered [`$table-border-factor`](https://getbootstrap.com/docs/5.3/content/tables/#sass-variables):
image

tomhughes:

PR #4704, opened 2024-04-15
Use "emphasis" link color in changeset lists instead of dark text

AntonKhorev:
Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #4705, opened 2024-04-15
Use Bootstrap hoverable rows in directions tables

AntonKhorev:
There was a hover effect on rows in directions tables. It got broken probably when Bootstrap started setting background directly on table cells instead of table rows.

Here I'm using their standard hoverable rows. The color is different. We used to have a yellow highlight, but that one would have to be adapted to dark mode while the standard highlight works fine.

Before:
image

After:
image
image


tomhughes:
Looks good to me, thanks.

PR #4707, opened 2024-04-16
Move next page button of shared pagination to the right side

AntonKhorev:
image
image
image
image


tordans:
I don't consider this an improvement. It is better to have those element in one group. That is why the design is rounded on one side and not the other.

tomhughes:
I think it's probably OK especially combined with #4710 which otherwise looks like quite a lot of buttons altogether.

I agree that we might want to round the other side of the buttons though?

AntonKhorev:
Now there are oldest/newest buttons at these places.

PR #4708, opened 2024-04-16
Reduce test run time portion of build process (8:31 reduced to 3:51)

josh-works:
update: This is now ready to merge!

Doing a quick audit of the test harness, to see where time is being spent, and to try side-stepping some of it. 

I'm doing lots of this work locally, but defer to using Github Actions timing data as the gold standard for if I'm successful or not. 

[I have some live, detailed, time-stamped stream-of-consciousness notes here](https://gist.github.com/josh-works/552c7db716372431d5b181f19ee36804), might be helpful to future me or others.

I am primarily investigating 'just' the tests, rather than tweaking the build pipeline at all. Faster/easier tests are so low-risk, high-reward. 

Here's the test suite on `main`, right now, as I open this PR:

2024-04-16 at 10 24 AM

here's the detailed results on tests as being run on another PR: https://github.com/openstreetmap/openstreetmap-website/actions/runs/8708315212/job/23885377955?pr=4706

You can drill into the details, and see the tests take times like:

- 7m 0s
- 8m 0s
- 7m 26s
- 8m 13s
- 8m 7s
- 8m 31s

## the finished results

the current PR, with logging turned off, stubbing out the password hash process in the user factory returns, in the same order, much improved results:

- 3:50 (improvement of 4:10, 58% faster)
- 3:42 (improvement of 4:20 , 73% faster)
- 3:49 (improvement of 3:20 64% faster)
- 4:12  (improvement of 4m, 52% faster)
- 3:58 (improvement of 4:10, 56% faster)
- 3:55 (improvement of 4:30, 62% faster)

## change 1, turn off logging, minor

I expect 'turning off logging' to lead to a 6% decrease in test time. That would be each of these figures going down by about 30 seconds. (this is something [Vladimir Dementyev](https://github.com/palkan) said on a [codewithjason podcast episode](https://www.codewithjason.com/podcast/12948312-185-test-suite-performance-with-vladimir-dementyev/)

## change 2, eliminate 4000 calls to `PasswordHash.create`, meaningful

the next iteration of the PR, in addition to turning off logging, eliminates the `PasswordHash.create` call, on each the ~1000 users set up across each thread. 

There's 4 threads, usually, 1000 users each, or 4000 total users. Doesn't seem like calling a single method 4000 times would be super expensive, but it seems to account for *three or four total minutes of the 8 minute test run time.* (Makes sense, I suppose, as that function is intended to be arbitrarily expensive). I just hard-coded the value to a valid string.

I ended up with this fix after noticing a LOT of time being spent in the `user` factory. [detailed notes here](https://gist.github.com/josh-works/552c7db716372431d5b181f19ee36804#let_it_be-as-a-strategy)

Current test run time is: 4:52 total (sometimes lower), down from 9:30, _on my machine_. 

Questions? Any reasons to not merge this? 

I cannot see the exact credit 'cost' of each build, I'd love to see that figure, if it's available, to also know how this change lowers the 'credits per build'.

### DIFF
--- a/test/factories/user.rb
+++ b/test/factories/user.rb
@@ -2,7 +2,7 @@ FactoryBot.define do
   factory :user do
     sequence(:email) { |n| "user#{n}@example.com" }
     sequence(:display_name) { |n| "User #{n}" }
-    pass_crypt { PasswordHash.create("test").first }
+    pass_crypt { "$argon2id$v=19$m=65536,t=3,p=4$M+oP6KQOdFLDeiYa/gbzAg$pULNwXgKt2sEmTxMXxt298skEqc8MjnPwGsk075jLrk" }
 
     # These attributes are not the defaults, but in most tests we want
     # a 'normal' user who can log in without being redirected etc.
--- a/test/factories/user.rb
+++ b/test/factories/user.rb
@@ -1,8 +1,10 @@
+pass_crypt = PasswordHash.create("test").first 
+
 FactoryBot.define do
   factory :user do
     sequence(:email) { |n| "user#{n}@example.com" }
     sequence(:display_name) { |n| "User #{n}" }
-    pass_crypt { "$argon2id$v=19$m=65536,t=3,p=4$M+oP6KQOdFLDeiYa/gbzAg$pULNwXgKt2sEmTxMXxt298skEqc8MjnPwGsk075jLrk" }
+    pass_crypt { pass_crypt }
 
     # These attributes are not the defaults, but in most tests we want
     # a 'normal' user who can log in without being redirected etc.
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -11,6 +11,9 @@ Rails.application.configure do
   # While tests run files are not watched, reloading is not necessary.
   config.enable_reloading = false
 
+  # disable logging in tests, for speed increases. Set to :info to bring back logging
+  config.log_level = :warn
+
   # Eager loading loads your entire application. When running a single test locally,
   # this is usually not necessary, and can slow down your test suite. However, it's
   # recommended that you enable it in continuous integration systems to ensure eager


josh-works:
Sorta stream-of-consiousness style, I made another change, after following some idea-threads for a few minutes, possibly got a good win, the tests now look down by three minutes, from something like 7:50 to 4:50.

I'm interested in seeing it in the CI pipeline, of course. 👀

gravitystorm:
> am eager for another admin approval

Great work. Just as a heads-up, I don't get notifications when you edit comments, so I missed that you had pushed new changes and wanted CI to be re-run. Simply add a new comment (as you later did) and everything will be notified properly.

tomhughes:
Password hash creation is deliberately slow, to make brute forcing hard - it's a design goal of any modern password hashing algorithm.

josh-works:
I'd sort of like to get this PR merged as is - I've got some other ideas that I'm exploring, but these two changes seem appropriately low-risk and high-reward. The test improvement speed is good enough. The total CI build time has now changed, too, but I'm not quite sure in exactly what ways. 

For what its worth, I wonder if at some times of day the tests run faster than at other times of day. I found one PR where the tests 'ran fast' on first blush, with a `successful in 5 m` message visible - I misplaced it and am looking again, but it also was in an errored state, so I'm not sure it's a valid data point anyway. 

Generally, the `docker/docker (pull request)` task has been posting as 'successful' in 12-14 minutes, with a bunch of tests taking 9 minutes.

When my tests are taking 5 minutes, that same task shows completing in 10 minutes, so this might represent a total CI run time savings of 20-40% (12->10 or 14-> 10)

Not bad at all.

test/factories/user.rb
```diff
@@ -2,7 +2,7 @@
   factory :user do
     sequence(:email) { |n| "user#{n}@example.com" }
     sequence(:display_name) { |n| "User #{n}" }
-    pass_crypt { PasswordHash.create("test").first }
+    pass_crypt { "$argon2id$v=19$m=65536,t=3,p=4$M+oP6KQOdFLDeiYa/gbzAg$pULNwXgKt2sEmTxMXxt298skEqc8MjnPwGsk075jLrk" }
```
tomhughes, line 5: Is there some way we could run `PasswordHash.create` once and then reuse the result, rather than actually hardcoding a single value?

josh-works: @tomhughes perhaps we were already on the exact same page. The hardcoded value I got was the result of running `PasswordHash.create` - I printed all the used values across 4000 users to the terminal, was seeing things like:

```
"$argon2id$v=19$m=65536,t=3,p=4$M+oP6KQOdFLDeiYa/gbzAg$pULNwXgKt2sEmTxMXxt298skEqc8MjnPwGsk075jLrk"
"$argon2id$v=19$m=65536,t=3,p=4$MWk3F1jq8OGs8m1KADSKzQ$97iChQ0pI7ZhdKWvK2zuALdl6vr1cko9YneceBZ0SRY"
"$argon2id$v=19$m=65536,t=3,p=4$I6joRKGjBBDFUkDVTGl3Hw$Mm7HBWpQiA8zw0y9DlGScJsAZWLEpEYP/vwyYOCQgXo"
"$argon2id$v=19$m=65536,t=3,p=4$3W1VtIzJn00BkJjIbi7L9Q$jSrBc7co66BsfTLINrDjPMwHfmdH2WtXRYvC8qw8b5A"
"$argon2id$v=19$m=65536,t=3,p=4$1QCD2+t4UuM1awxraePirQ$iQp7Rf1NskAbIAWzTv5NBdtiSR+JzuGwk2sJUQWbLc0"
"$argon2id$v=19$m=65536,t=3,p=4$Sm2M0daauln/R67Dc2NNCQ$zo5gMtLEGSUZy1SPsGc2rnlPsUPtW11w4SZHNuKloH4"
"$argon2id$v=19$m=65536,t=3,p=4$gHbAIYV/n1rLO+ICu61fAQ$n++ouqaI35kxerp6woJMjb3xJS+7ALUa6h7zM5eNNeM"
```

so, the given hardcoded value _is_ the result of this function call, I'm sort of still inclined to leave it as is. 

It feels a smidge safer/safe enough to put in the User factory a magic string, rather than hiding that there is the same password in use across all of the temporary users. 

Perhaps there's a threat vector I'm not perceiving, happy to hear more, if something still feels off.

josh-works: yoooo I just found a clean way of doing exactly what you suggested:

https://github.com/josh-works/mastodon/blob/babbde5d6f6d3b3724ae71550ae7814d47f16ed0/spec/fabricators/account_fabricator.rb#L3

I'll implement this here. It looks better than what I wrote, and better than what I was imagining based on what _you_ said.

josh-works: OK, pushed the refactor. Hilariously I'd not seen this pattern in factories before, so it sort of slipped my mind as an option available to me. Thanks for the nudge.

josh-works:
This is ready for re-review/workflows to run, I perceive. Thanks @tomhughes.

tomhughes:
Times on my dev box, before:

> Finished in 96.928949s, 18.0441 runs/s, 10375.3730 assertions/s.

and after:

> Finished in 41.567882s, 42.0758 runs/s, 24193.5346 assertions/s.

tomhughes:
Looks great to me, thanks a lot!

PR #4711, opened 2024-04-17
Use Bootstrap actionable items in query results

AntonKhorev:
Similar to #4705, but for a [list-group](https://getbootstrap.com/docs/5.3/components/list-group/#links-and-buttons).

image


tomhughes:
I'm fine with this in principle - the only odd thing is that the list hover highlight colour is quite a lot paler than the table one. The list one is just the tertiary colour while the table one applies an alpha mask to the emphasis colour.

AntonKhorev:
Changed the list hover background to be the same as the table one:
image
image

tomhughes:
That looks better, thanks.

PR #4712, opened 2024-04-17
Dim map tiles in dark mode

AntonKhorev:
Map tiles are going to be too bright in dark mode. The question is then what to do with them. There are two options: dim them or invert them, and the existing dark themes like [this one](https://userstyles.world/style/15596/openstreetmap-dark-theme) may offer a choice.

Here I'm making a conservative choice of 80% dimming.

image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -474,6 +474,13 @@ body.small-nav {
   border-radius: 4px !important;
 }
 
+@include color-mode(dark) {
+  .leaflet-tile-container,
+  .mapkey-table-entry td:first-child > * {
+    filter: brightness(.8);
+  }
+}
+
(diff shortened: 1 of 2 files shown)


tomhughes:
This looks reasonable to me... I'm no expert on dark mode but the tiles ar every bright without this and much more reasonable with it though I did have to turn my laptop brightness up a bit but I suspect that's just something I would have to do if I used dark mode as I normally keep it quite low.

mxdanger:
> There are two options: dim them or invert them

You left out the third option of not changing anything! Just leaving my two cents here to say this should be reverted. Thanks.

fincentxyz:
Why dim the entire map? This makes things very difficult to read... Maybe a toggle to dim the map would be more appropriate. (As well as a toggle for dark mode in general)

PR #4713, opened 2024-04-17
Force word wrap in tag table

### LINKED ISSUE #4709: Long links are overriding the tag name
### URL

https://www.openstreetmap.org/node/2570217182

Open the page and look at the tags, this is what I see (Hebrew UI):
Images links are overriding the tags name:
image

I'm using chrome, and I'm guessing this is a simple css fix of wordbreak somewhere.

AntonKhorev:
Fixes #4709, kind of.

Aside from differences in browser behavior, there's also a decision by the Bootstrap developers. [They disable `.text-break` in rtl mode](https://getbootstrap.com/docs/5.3/utilities/text/#word-break). They say that rtl is most likely Arabic and word breaks in Arabic are wrong therefore disable word breaks for any rtl. But in our case any long words in tags table are likely not Arabic but rather urls or something similar.

Here I'm putting the css properties that are normally in Bootstrap's `.text-break` to our custom css. This will restore word breaks in rtl tag tables, occasionally with "wrong" Arabic word breaks.

Screenshots in Chrome (Firefox does word wraps anyway for whatever reason):

Before:
image

After:
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -616,6 +616,8 @@ tr.turn {
   .browse-tag-list {
     table-layout: fixed;
     white-space: pre-wrap;
+    word-wrap: break-word;
+    word-break: break-word;
 
     tr:last-child th, tr:last-child td {
       border-bottom: 0;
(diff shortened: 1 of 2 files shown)


tomhughes:
You would think that if arabic doesn't do word breaks then the browser would fail to find breaks in genuine arabic text and that the whole thing wouldn't be an issue...

This sounds like a reasonable fix in any case.

PR #4716, opened 2024-04-24
Use emphasis color in "Where this is?" hover state

AntonKhorev:
Before in dark mode:
image

After in dark mode:
image

After in light mode:
image


tomhughes:
Looks good to me, thanks.

PR #4720, opened 2024-04-24
Use Bootstrap actionable items in changeset lists

AntonKhorev:
Like #4711 but for changeset history lists.

image
image


tomhughes:
Looks good to me, thanks.

PR #4722, opened 2024-04-25
Use bindPopup options to set min width of user popup

AntonKhorev:
Leaflet may get confused if you use css `min-width` property on popup contents.

Examples in Chromium where scrollbars always take width.

Here I made the contents longer and limited the popup height while keeping css `min-width`:
image

This is with `minWidth` `.bindPopup` option, no horizontal scrollbars appear:
image


tomhughes:
I can't easily test this, but it seems reasonable and correct.

PR #4723, opened 2024-04-25
Center the nav links vertically

### LINKED ISSUE #3372: Navbar items not vertically centered
image

The navbar items' text is not vertically centered like the other items.

lectrician1:
To close #3372

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -111,7 +111,8 @@ header {
   font-size: 14px;
 
   h1, nav, nav > ul, nav > ul > li {
-    display: inline-block;
+    display: flex;
+    align-items: center;
   }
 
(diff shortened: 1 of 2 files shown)


tomhughes (commented):
The vertical alignment looks much better which is great but at least in Firefox the horizontal spacing between the items is reduced - the cause seems to be that previously the DOM put a whitespace item between each list item and that is gone with the flex layout.

Adding a one pixel margin on the list item seems to get it back to roughly where it was before but maybe there's a better solution or maybe people think the reduced spacing is ok?

AntonKhorev:
This shouldn't be required with #4828 merged.

tomhughes:
That's very true. Thanks for the work @lectrician1 but I think this has been resolved by #4828 now.

PR #4725, opened 2024-04-27
Remove simulating clicks on <a> when <li> is clicked

AntonKhorev:
The simulation is not required because the link is stretched.

I stretched links in https://github.com/openstreetmap/openstreetmap-website/commit/62ba143c3c0b7608dcd41f9604fe35ca22d126b4 but I was focused on removing the mouse pointer css and didn't change javascript.


tomhughes:
Nice to get rid of that lot! Thanks.

PR #4727, opened 2024-04-27
Remove custom small font size

AntonKhorev:
For `<small>` Bootstrap sets the font size to almost the same, 12.25px vs our custom 12px:
https://github.com/twbs/bootstrap/blob/d7b22b77dbe7d1bd8d5cb03cec5d89b480500afb/scss/_reboot.scss#L211

Looks like we don't use `<aside>` anywhere.


tomhughes:
Looks good to me, thanks.

PR #4728, opened 2024-04-27
Stretch query result links

AntonKhorev:
#4725 for query results


tomhughes:
Well that's quite a nice improvement to the code with no change in functionality! Thanks.

PR #4710, opened 2024-04-16
Add oldest/newest links to shared pagination

AntonKhorev:
#4707 was intended for this PR and possibly for putting something between </> buttons. We already do it on changeset history pages.

image
image
image

PR #4729, opened 2024-04-27
Add oldest/newest links to shared pagination - with gap between "newer" and "older"

AntonKhorev:
#4710 + #4707 + all borders rounded. With this rounding the buttons won't be able to collapse into one block, there will always be a gap between "newer" and "older".

image
image
image

### DIFF
--- a/app/controllers/concerns/pagination_methods.rb
+++ b/app/controllers/concerns/pagination_methods.rb
@@ -6,8 +6,8 @@ module PaginationMethods
   ##
   # limit selected items to one page, get ids of first item before/after the page
   def get_page_items(items, includes: [], limit: 20)
-    param! :before, Integer, :min => 1
-    param! :after, Integer, :min => 1
+    param! :before, Integer, :min => 0
+    param! :after, Integer, :min => 0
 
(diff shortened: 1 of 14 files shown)


tordans:
From a UX point of view I suggest to only show one of the action (next or last) as full text in the full width UI. We should decide which is more important and make the other option a icon+tooltip only UI. This will declutter the UI a lot and give users a clear focus on the primary action.

It will also reduce the need to split the actions in two groups. There is a reason most UI library show them as a group. It reduces clutter in the UI, makes it easy to understand the actions as relation and reduces the distance the mouse has to move to click back and forth.

---

I am assuming the screenshots are taken from a page in the middle of the list. On the first and last page the buttons should be inactive or even hidden (inactive is the better choice most of the time).

AntonKhorev:
> It will also reduce the need to split the actions in two groups.

No, it won't because I did it for a different reason. I did it to avoid having four small buttons next to each other on small screens. That would invite tapping a wrong button. There's another way to split them though.

AntonKhorev:
Superseded by https://github.com/openstreetmap/openstreetmap-website/pull/4734

PR #4730, opened 2024-04-27
Use Bootstrap classes for diary navigation icons

AntonKhorev:
Removes custom classes `rsssmall` and `small_icon`.


tomhughes:
Nice, and the icon is better aligned as well! Thanks.

PR #4732, opened 2024-04-27
Remove custom css applied to images inside paragraphs

AntonKhorev:
- Many images are not inside `<p>`, this css doesn't apply to them.
- `width: auto` is a default width
- `max-width: 100%` shrinks images that are too wide to fit on the screen. Images inside rich text (diaries, comments) already have this set elsewhere.
- It affects image on the [copyright](https://www.openstreetmap.org/copyright) page, but that image is narrow enough. If there's any wide image that requires shrinking, [`.img-fluid`](https://getbootstrap.com/docs/5.3/content/images/#responsive-images) can be used instead.
  image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -10,11 +10,6 @@ body {
   font-size: $typeheight;
 }
 
-p > img {
-  width: auto;
-  max-width: 100%;
-}
-
 time[title] {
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good. I'll probably add `img-fluid` to that image anyway even thought he page has to be ludicrously narrow to trigger it there.

PR #4735, opened 2024-04-29
Invert dark browse icons in dark mode

AntonKhorev:
Some icons in the browse sidebar are too dark for dark mode. #4701 was a quick fix that could be applied to all icons. Here's another fix where only the selected dark icons are inverted.

In order to achieve that I did the following:
- icon styles are set with a Sass `@each` loop that goes over the icon data
- icon data is a nested Sass map where each icon has a filename and possibly an invert flag
- css properties are moved to `common.scss` to avoid Sass processing errors because they now use Bootstrap mixins and Bootstrap is included in `common.scss`

Node icons in dark mode:
image image image image image image

Way icons are mostly unchanged.

Other required fixes:
- [x] Tree icon needs to have white background removed. Actually a couple of other icons too, but they are inverted and white background is not noticeable. Fixed in #4736.


tomhughes:
Very nice - all looks good to me, thanks.

PR #4736, opened 2024-04-29
Remove white background from tree, gate, level crossing icons

AntonKhorev:
image

Before in dark mode:
image

After in dark mode:
image

Foreground brightness of some icons is changed in #4735.


tomhughes:
Looks good to me, thanks.

PR #4737, opened 2024-04-29
Adapt Leaflet context menus to dark mode

AntonKhorev:
Before in light and dark mode:
image

After in light mode:
image

After in dark mode:
image


tomhughes:
Looks good to me, thanks.

PR #4738, opened 2024-04-29
Remove color override for user dropdown message count

AntonKhorev:
There's a message counter in the upper right corner (`#inboxanchor`). It was gray for a long time (https://github.com/openstreetmap/openstreetmap-website/commit/ff752b072f44aad142314a57eed79271928d181a) but why? Other counters are green, including the same counter in the dropdown menu.

Before in light mode:
image

Before in dark mode:
image

After in light mode:
image

After in dark mode:
image


tomhughes:
Seems reasonable to me - as you say it was somewhere inconsistent!

PR #4740, opened 2024-04-30
Don't change display property of messages table header cell

AntonKhorev:
I accidentally changed display properties of table cells by using `.d-flex`, then I fixed it in #4702. However by that time `.d-flex` spread to more table cells in https://github.com/openstreetmap/openstreetmap-website/commit/3e69a4512f2d1b39f48719882a0d112d042da73a. Here I'm removing it from message actions header cell.


tomhughes:
Looks good to me, thanks.

PR #4742, opened 2024-04-30
Delete role icon pngs, use only svgs

AntonKhorev:
Instead of doing #4741 or something similar, let's start with removing png role icons. Those are not used by browsers because svgs are also provided.

The reason I'm doing this is that I want to change some colors (white "blank" color, too dark blue moderator color) and I don't want to change pngs too.


tomhughes:
Why instead of? Is there a problem with #4741 or something?

PR #4744, opened 2024-04-30
Set trace thumbnail link display property to inline-block, fixing focus outline

AntonKhorev:
Try to move keyboard focus to a trace thumbnail while on a trace list page. This is what you'll see in Firefox:
image

After the fix:
image

`` is an inline element by default, `<a>` is also an inline element, so a link with a thumbnail inside is as tall as a line of text. And it's shorter than the actual image size. You can see this in browser dev tools, both in Firefox and in Chrome. Chrome still draws the outline around the image but Firefox doesn't.


tomhughes:
Looks good to me, thanks.

PR #4746, opened 2024-05-01
Restore constraints on note id parameter, simplify tests

AntonKhorev:
Elements, changesets, notes had constraints in their id parameter in routes. https://github.com/openstreetmap/openstreetmap-website/commit/9748ce301c9488b9cd32439448770c3d4a38ebd6 removed this constraint for notes for unknown reason. Without this constraint notes controller needs different tests. The part of "browse_check" that checks negative ids had to be commented out.

This PR restores the constraint which allows to use the same "browse_check" as for every other entity formerly handled by browse controller.


tomhughes:
Not sure what was going on with that previous change, but this looks good to me. Thanks.

Issue #3592, opened 2022-07-08
Viewing user edit history does not show empty changesets


deevroman:
### URL

https://www.openstreetmap.org/user/TrickyFoxy/history

### Screenshot(s) or anything else?
OSMCha:
<img width="349" alt="Снимок экрана 2022-07-08 в 15 19 58" src="https://user-images.githubusercontent.com/25708359/177991435-a89badc6-62dd-46e4-9499-2292c7427a13.png">
OSM website:
<img width="335" alt="Снимок экрана 2022-07-08 в 15 20 12" src="https://user-images.githubusercontent.com/25708359/177991475-40e485e1-0846-45ba-9810-931055be0bf8.png">

tomhughes:
That's probably deliberate because an empty changeset is not really very interesting.

tomhughes:
Indeed the starting condition for any changeset list `conditions_nonempty` from https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/changesets_controller.rb#L45,

deevroman:
This is a strange solution: empty changeset is a sign of an error in the editor

woodpeck:
No, it is possible that the user wanted to upload a change (and the editor duly created a changeset) but then it turned out that the change could not be uploaded due to a conflict (someone else had modified the object in the mean time) and the user decides not to proceed. Other similar situations are possible and there is no way to "cancel" a changeset once opened.

tomhughes:
Actually judging by the comment the main purpose was to eliminate in process changesets that hadn't had any uploads yet.

mmd-osm:
If there's a need to see these empty changesets, you can still use the OSM API: https://api.openstreetmap.org/api/0.6/changesets?display_name=TrickyFoxy

mmd-osm:
As per #4747 and earlier comments this seems like „works as designed“.

PR #4747, opened 2024-05-01
Link to previous/next nonempty user's changeset on changeset pages

### LINKED ISSUE #450: Inconsistency in displaying/ignoring empty changesets
When you view all changesets for a user e.g.

http://www.openstreetmap.org/user/tecpenleal/edits

those changesets that have zero edits are not shown. However, when you then click one one of the user's changesets you will get a set of forward/backward navigation buttons for previous/later changesets of that user:

```
« 17108201    tecpenleal      17108276 »
```

In this navigation, the empty changesets are not ignored. I think it would be better to treat empty changesets consistently, either display them everywhere or nowhere.

AntonKhorev:
Fixes #450, hiding previous/next empty changesets.

There's also an opposite request #3592.


tomhughes:
Looks good to me, thanks.

PR #4752, opened 2024-05-05
Redirect to inbox after marking a message as read/unread

### LINKED ISSUE #4750: "Mark as unread" doesn't do anything inside messages
### URL

https://www.openstreetmap.org/messages/{message_id}

1. Open a message
2. (Decide that you will reply to it later)
3. Press "Mark as unread"
4. "Message has been marked as unread"
5. Go to the inbox
6. Notice that the message you just marked as unread is actually read

Maybe at step 3 the user should be sent back to the inbox, so that he doesn't "read" the message again?

AntonKhorev:
Fixes #4750 by redirecting when "Mark as unread" is clicked on a message page. For example Github notifications work like that: if you mark a notification as unread it redirects you to the notifications list.

I rename the `unread_message` variable in tests because the message doesn't stay unread.

I reuse the xhr test lines for referer test because xhr is not in use since the Turbo update.

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -127,7 +127,7 @@ class MessagesController < ApplicationController
     @message.message_read = message_read
     if @message.save
       flash[:notice] = notice
-      redirect_back_or_to inbox_messages_path, :status => :see_other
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #4753, opened 2024-05-05
Remove tests for message paths without ids

AntonKhorev:
Remove tests with "... with no ID should fail ... assert_response :success" and like. They don't do what their comments say and never test missing ids.

Rails is being [smart with path helpers](https://thoughtbot.com/blog/avoid-making-path-helpers-guess) here. It sees that an id is missing and it uses the id from the previous request.


tomhughes:
No idea what I thought I was doing in 03cde71a5ee0fcbf4aa07c94bc973f5ab2a90292 when I introduced that nonsense! Thanks for catching it...

PR #4758, opened 2024-05-06
Stop using the session to persist the referer during login

tomhughes:
Using the session to persist the referer is not necessary and can be problematic in that if a login/signup is aborted partway through then the referer information can persist in the session and be reused during a later attempt.


gravitystorm:
Looks good to me. Happy to see this simplification.

PR #4759, opened 2024-05-06
Use inline svg for notice icons

AntonKhorev:
Allows to set fill to `currentColor` which adapts to dark mode.

Before:
image
image

After:
image
image


tomhughes:
Looks good to me, thanks.

PR #4760, opened 2024-05-06
Have two columns for layer buttons between small and medium screen widths

AntonKhorev:
[Originally](https://github.com/openstreetmap/openstreetmap-website/pull/3620) I wanted smaller buttons but that's not easy to do if I'm not allowed to check their widths. Then let's at least not have overly wide buttons when the screen is about 700px wide.

Before:
image

After:
image

### DIFF
--- a/app/assets/javascripts/leaflet.layers.js
+++ b/app/assets/javascripts/leaflet.layers.js
@@ -5,14 +5,21 @@ L.OSM.layers = function (options) {
     var layers = options.layers;
 
     var baseSection = $("<div>")
-      .attr("class", "base-layers d-grid gap-3 p-3 border-bottom border-secondary-subtle")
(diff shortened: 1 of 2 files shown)


tomhughes:
Looking at this in responsive view and gradually reducing the size I don't really like the way it switches from one column to two and then back to one.

That's not how real users will use it though - they'll most just have a device and see whatever view that device's size gives them. Well at least unless they use a desktop browser and make the window small enough to not get the large view which would be kind of wierd.

So my next question was, what sort of devices exactly get this view, and the answer initially seems to be "not many" as tablets are too large to get it and even large phones are too small to get it.

What does trigger it of course is a phone in landscape mode, and then it kind of makes sense because you have lots of width for the two columns, but not much height if you want to avoid scrolling.

Was that the use case you had in mind for this?

PR #4761, opened 2024-05-06
Enable dark theme if user prefers dark color scheme - no javascript version

AntonKhorev:
We probably can enable dark mode now.

This version uses system settings with media queries and doesn't need javascript. It is enabled by the Bootstrap option [`$color-mode-type: media-query`](https://getbootstrap.com/docs/5.3/customize/color-modes/#building-with-sass). Previous version #4653 used javascript to track system settings and change `data-bs-theme` attributes accordingly.

The problem with this no-javascript approach is that `data-bs-theme` attributes are ignored. But we're already using them for inverting close buttons on dark banners. When in light mode, dark banners have their close buttons inverted because banners themselves have a dark theme attribute. Aside from not working with `$color-mode-type: media-query`, this approach has another drawback: you can't invert buttons on light banners in dark mode. Once the entire webpage gets its dark mode attribute, you can't go back to light mode.

The solution in this PR is:
- use `$color-mode-type: media-query` that will take care of most of light/dark mode styling
- use `data-bs-theme` on banners
- add css to enable/disable close button inversion inside elements with `data-bs-theme`


AntonKhorev:
conflicts with scale control color fixes

gravitystorm:
Looks good to me! I think it's worth starting with the automatic mode, since that should just work for most people (and will automatically support day/night switching for devices that do that).

DaveF63:
I like dark modes in general, but can someone (other than TH) explain the reason for the lack of choice & the dimmed tile window?

tomhughes:
> I like dark modes in general, but can someone (other than TH) explain the reason for the lack of choice & the dimmed tile window?

Is my opinion not welcome then? What exactly have I done to upset you?

Vectorial1024:
@DaveF63 

Random OSM user passing by, but my guess is that, per https://github.com/gravitystorm/openstreetmap-carto/issues/5044, there are actually no "official" dark tiles from osm-carto yet, so the dimming is a stop-gap measure to get dark mode published. I would expect the tile-dimming problem to be dealt with sometime later by someone else.

PR #4762, opened 2024-05-07
Use inline svgs for user role icons

AntonKhorev:
An alternative to #4741. I want to change a couple of colors in "star" role icons. I'm not yet changing them here, but I'm making them easier to change by removing a bunch of files that would have to be edited.

The difference to #4741 is that I don't use a single svg file. That file would have been more complicated than a ruby method generating all role icons.


tomhughes:
Looks good to me. I mean the other one looked fine as well but I assume you prefer this one as that is still draft.

PR #4764, opened 2024-05-07
Adapt Leaflet scale and attribution controls to dark mode

AntonKhorev:
image


tomhughes:
I'm not sure if there's much that can be done about it but the bars on the scale control are pretty hard to see with this when in dark mode?

AntonKhorev:
I couldn't decide if I should make them lighter or darker. This probably happens because the map is dimmed.

AntonKhorev:
Made the bars lighter because if tiles are adjusted further they'll likely become darker:

image

tomhughes:

Issue #4765, opened 2024-05-07
Tags diff in object history


altilunium:
### Problem

There is no diff visualization in the [current OSM object's history section](https://www.openstreetmap.org/node/1/history).

### Description

![image](https://github.com/openstreetmap/openstreetmap-website/assets/70379302/1166e3ad-fd30-4028-ab38-53594db0c5d6)

Embed this CSS and JS to the "/history" page

CSS : 
~~~css
    .history-diff-new-tag {
      background: rgba(17,238,9,0.6) !important;
    }
    .history-diff-modified-tag {
      background: rgba(223,238,9,0.6) !important;
    }
    .history-diff-deleted-tag {
      background: rgba(238,51,9,0.6) !important;
    }
    .non-modified-tag .empty-version {
        
    }
    .hidden-non-modified-tag, .hidden-empty-version {
        display: none;
    }
    .hidden-version, .hidden-h4 {
        display: none;
    }
    
    #sidebar_content h2:not(.changeset-header){
        font-size: 1rem;
    }
    
    h4 {
        font-size: 1rem;
    }
    
    .find-deleted-user-btn {
        cursor: pointer !important;
    }
    
    .copied {
      background-color: red !important;
      transition:all 0.3s;
    }
    .was-copied {
      background-color: none !important;
      transition:all 0.3s;
    }
~~~

JS : 
~~~javascript
function addHistoryLink() {
    if (!location.pathname.includes("/node")
        && !location.pathname.includes("/way")
        && !location.pathname.includes("/relation")
        || location.pathname.includes("/history")
    ) return;
    if (document.querySelector('.history_button_class')) return true;
    let versionInSidebar = document.querySelector("#sidebar_content h4 a")
    if (!versionInSidebar) {
        return
    }
    let a = document.createElement("a")
    let curHref = document.querySelector("#sidebar_content h4 a").href.match(/(.*)\/(\d+)$/)
    a.href = curHref[1]
    a.textContent = "🕒"
    a.classList.add("history_button_class")
    if (curHref[2] !== "1") {
        versionInSidebar.after(a)
        versionInSidebar.after(document.createTextNode("\xA0"))
    }
}

function makeHistoryCompact() {
    // todo -> toogleAttribute
    if (document.querySelector(".compact-toggle-btn").textContent === "><") {
        document.querySelectorAll(".non-modified-tag").forEach((el) => {
            el.classList.replace("non-modified-tag", "hidden-non-modified-tag")
        })
[...]

PR #4766, opened 2024-05-07
Tags diff in object history

altilunium:
Implementation of https://github.com/openstreetmap/openstreetmap-website/issues/4765


altilunium:
image
image

deevroman:
You should not copy code that is intended to work as an extension. It is written carelessly, does a lot of extra work, and changes styles a lot. Simple copy-paste won't help.

altilunium:
@deevroman 

Ok. Sorry.

PR #4767, opened 2024-05-07
Tags diff in object history (2nd try)

altilunium:
Implementation of https://github.com/openstreetmap/openstreetmap-website/issues/4765

Fixing ERB lint failing checks : https://github.com/openstreetmap/openstreetmap-website/pull/4766


tomhughes:
You don't have to close a PR if you need to change it - just push a new version of your branch and the PR will update.

altilunium:
Thank you for the information.

It turns out that the error grew even more when I tried to fix one of the lint checks. I am not very familiar with Ruby, and I think I need to learn about it more before submitting another pull request. For now, I will close this. :)

gravitystorm:
> It turns out that the error grew even more when I tried to fix one of the lint checks. I am not very familiar with Ruby, and I think I need to learn about it more before submitting another pull request. For now, I will close this. :)

If you have any questions about the lint checks, or Ruby/Rails etc, please feel free to ask and we're happy to answer your questions. Hopefully the lint messages given (e.g. "avoid using inline script elements") are self-explanatory but we can elaborate on them too.

tomhughes:
It's not that it can't, it's that it is configured not to because it is configured to enforce our coding standards.

mmd-osm:
By the way, does it make sense to implement this feature in Javascript, or would we rather use server side rendering?

AntonKhorev:
I planned to do this server-side. The maintainers here probably want it done with Turbo. I already used Turbo in the sidebar in #4565.

altilunium:
image

... My submitted code didn't even touch or use the 'browser' gem in the [Gemfile](https://github.com/openstreetmap/openstreetmap-website/blob/master/Gemfile)

gravitystorm:
> I planned to do this server-side. The maintainers here probably want it done with Turbo.

Thanks @mmd-osm and @AntonKhorev for pointing this out - yes, server-side is preferred for arranging data and generating html, and as little javascript as is required for interactivity, generally using the Turbo/HotWire approach.

gravitystorm:
> My submitted code didn't even touch or use the 'browser' gem in the [Gemfile](https://github.com/openstreetmap/openstreetmap-website/blob/master/Gemfile).

Yes, unfortunately that was a problem with a recent update to the code base, but we’ve fixed that now.

PR #4768, opened 2024-05-08
Pin browser to versions that support ruby 3.0

gravitystorm:
The latest release of the `browser` gem, version 6.0.0, dropped support for ruby 3.0 and ruby 3.1. Ruby 3.0 is our minimum supported version, at least until we drop support for Ubuntu 20.04.

This PR constrains `browser` to "< 6" in order to keep supporting ruby 3.0

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -131,7 +131,7 @@ gem "gd2-ffij", ">= 0.4.0"
 gem "marcel"
 
 # Used for browser detection
-gem "browser"
+gem "browser", "< 6" # for ruby 3.0 support
 
 # Used for S3 object storage
 gem "aws-sdk-s3"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -136,7 +136,7 @@ GEM
     brakeman (6.1.2)
       racc
     brotli (0.5.0)
-    browser (6.0.0)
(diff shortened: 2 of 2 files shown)


Firefishy:
I guess add the ruby version to the Gemfile wouldn't work, as that wouldn't allow ruby 3.1.x
```
ruby ~> 3.0.0
```

gravitystorm:
> I guess adding the supported ruby version to the Gemfile wouldn't work

I struggled to find the specifications for the ruby keyword in Gemfiles, and also what bundler might then do with it. For example, [heroku says it needs three digits specifiers](https://devcenter.heroku.com/articles/ruby-versions#ruby-version-specifiers) which won't allow what we want. 

And even if we add it, it's not clear to me if bundler uses that information (as opposed to only using the currently running ruby version) when figuring out the dependency graph.

tomhughes:
Ah sorry I misunderstood what was going on when the tests started failing last night and somehow decided actions was playing up an didn't investigate further.

gravitystorm:
> Ah sorry I misunderstood what was going on when the tests started failing last night and somehow decided actions was playing up an didn't investigate further.

No worries!

PR #4770, opened 2024-05-08
Lighten moderator icon color

AntonKhorev:
Before:

image
image

You can see that the moderator star is significantly darker than the others and it is too dark for dark mode.

After:

image
image


tomhughes:

PR #4772, opened 2024-05-08
Remove bs-custom-file-input javascript plugin

AntonKhorev:
That was a [Bootstrap 4 plugin](https://getbootstrap.com/docs/4.6/components/forms/#file-browser). Bootstrap 5 docs [no longer recommend it](https://getbootstrap.com/docs/5.3/forms/form-control/#file-input). It doesn't do anything because there are no inputs matching [its selector](https://github.com/Johann-S/bs-custom-file-input/blob/a4cb893188f06e94c811bac1e5c3682bb4b0e459/src/selector.js#L2).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,8 +16,6 @@
 //= require matomo
 //= require richtext
 //= require qs/dist/qs
-//= require bs-custom-file-input
-//= require bs-custom-file-input-init
 
 /*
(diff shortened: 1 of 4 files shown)


tomhughes:
So in fact it's never done anything by the looks of it... It came in with 8d2271699bfe8199ebeac2d2bf491e2133cc02d8 on the basis of the recommendation in the documentation but we don't seem to have realised be needed to mark inputs with wanted it to apply to with a special class...

PR #4776, opened 2024-05-09
Remove border attribute from <img> on Copyright page

AntonKhorev:
This is a long-deprecated attribute. I already removed some in #4762. The one on the copyright page is the only `border` attribute outside emails.


tomhughes:
Looks good to me, thanks.

Issue #4769, opened 2024-05-08
Cartography for dark mode


gravitystorm:
This is a followup discussion for https://github.com/openstreetmap/openstreetmap-website/pull/4712

Dark mode cartography is a tricky subject. Unlike the rest of the site, the colours used on the maps:

a) are not under our control
b) are under the control of different groups of cartographers on a per-layer basis
c) are generally extremely carefully chosen
d) are often very close to one another, with significant meaning often between similar colours

"Dark Modes" in general are also tricky, because they are more complex than they might first seem. For example, the text and background colours are generally inverted (black text on white background becomes white text on black background) and not just "darkened" as the mode name might suggest (which would be very black text on a grey background). But that's not the end of the story either, since the inversion is not symmetric - white (or very light grey) text on a black background is hard to read, so the inversion needs to be adjusted for human perception reasons. Here's the values chosen by Bootstrap:

| | Light Mode | Dark Mode|
|---|---|---|
|text|gray-900 | gray-300 |
|background| white | gray-900 |

But even then, not all colours are even inverted in dark mode. You can see with the example dropdowns on the Bootstrap documentation that the primary colour highlighting is exactly the same - even the brightness is unchanged: https://getbootstrap.com/docs/5.3/customize/color-modes/#example

The final aspect is on human colour perception. There are certain colours (e.g. yellow) that appear brighter for the same brightness value in RGB, and there are ranges of the colour spectrum where humans can distinguish colours more easily (e.g. 10% brightness differences are hard to distinguish in very dark colours, and bright colours are easier to separate from each other than dark ones). This means that mechanical changes to brightness or naive inversions could make the cartography illegible. The work that went into CartoCSS to support HSLUV colour definitions and perceptual colour functions can provide illustrative background information.

---

So what should we do about the maps? I can see 3 main options:

1) Leave it to the cartographers
2) Use a simple darken
3) Use inversion+rotation

### 1) Leave it to the cartographers

[...]

PR #4777, opened 2024-05-09
Manage map tiles dark mode with leaflet-osm plugin

AntonKhorev:
I guess #4769 meant "how about we use dark tile variants when available". If [Transport Dark](https://www.thunderforest.com/maps/transport-dark/) is approved to be used on the osm website, we can switch to it in dark mode when "Transport Map" is selected.

Here's how it can work together with prefers-color-scheme switching from #4761:

image
image
image
image

Layers without alternative dark tiles get dimmed, layers with switch to their alternative tiles.

I modified leaflet-osm in this repository directly. Is https://github.com/openstreetmap/leaflet-osm maintained? Its [npm package](https://www.npmjs.com/package/leaflet-osm) was last updated years ago.


AntonKhorev:
Added context menu options similar to "Map Brightness" dropdown in https://userstyles.world/style/15596/openstreetmap-dark-theme

image image

AntonKhorev:
- [ ] There's a bug when switching to a layer that comes earlier in the list because I assumed that "add" event happens after "remove" event.
- [ ] Context menu is not translated.

pnorman:
> I modified leaflet-osm in this repository directly. Is https://github.com/openstreetmap/leaflet-osm maintained

Yes. It's had updates in the last 8 months, when we added or removed featured layers.

tomhughes:
Specifically I have maintainer access there ;-)

AntonKhorev:
> Yes. It's had updates in the last 8 months, when we added or removed featured layers.

Supposedly not when layers are removed because `L.OSM.OPNVKarte` is still there. And it would be odd for this layer to disappear for anyone who is using leaflet-osm independently from the osm website.

hlfan:
Great work! I almost feel bad for missing this.

> Context menu selection is not remembered

What is currently planned for this? Just persistence via `localStorage` or similar? Something more intricate? User syncing like mentioned in #5324 (also how would this work for non logged-in users?)?

PR #4778, opened 2024-05-10
Fix CSP failures for Microsoft social sign-in

milan-cvetkovic:
Add login.microsoftonline.com to CSP allow list for `/account/new`, `/account/edit` and `/users/new`

To reproduce:
`/account/edit`
- Login as existing user
- go to "My Settings"
- change "External Authentication" to Microsoft
- click "Save Changes"
- Page fails to load due to CSP violation

`/account/update` ?

`/users/new`
- Login to https://login.live.com with your existing MS account
- navigate to https://www.openstreetmap.com/login
- click on Microsoft Icon to use social account
- on `/user/new` page click "Sign up"
- Page fails to load due to CSP violation

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -17,7 +17,7 @@ class AccountsController < ApplicationController
     @tokens = current_user.oauth_tokens.authorized
 
     append_content_security_policy_directives(
-      :form_action => %w[accounts.google.com *.facebook.com login.live.com github.com meta.wikimedia.org]
+      :form_action => %w[accounts.google.com *.facebook.com login.live.com login.microsoftonline.com github.com meta.wikimedia.org]
     )
 
     if errors = session.delete(:user_errors)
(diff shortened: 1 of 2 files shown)


tomhughes:
Has there been some change at the Microsoft end to the URLs that are used? Do we still need the live.com URL as well as this?

tomhughes:
It looks like this dates to 7428da74c24c69b4a51e8a0fed57b71806cd7e6c when we changed from the windowslive omniauth plugin to the microsoft_graph plugin. I'm just astonished it has been broken for nearly a year with nobody reporting it!

As far as I can see the live.com URLs can be removed as the new plugin doesn't reference those at all and the old one does.

milan-cvetkovic:
I have not been able to hit /account/update route to try it out.

> As far as I can see the live.com URLs can be removed as the new plugin doesn't reference those at all and the old one does.

It appears so, in the 2 test cases I was able to try - I am not sure how to trigger /account/update route.

Interestingly, if we get rid of the second round trip as it was suggested in #4455, but later re-introduced in [Re-introduce additional round trip for verifying auth_provider](https://github.com/openstreetmap/openstreetmap-website/pull/4455/commits/4965c19b7a8c96ab87b543af1fc36a1ad7514c09), we could remove the CSP setup in [`users_controller.rb`](https://github.com/openstreetmap/openstreetmap-website/blob/ef00f9a4674d2a0719635333cef8ac6bdc21d940/app/controllers/users_controller.rb#L65C5-L67C6) entirely:

tomhughes:
I've amended it to remove live.com and will merge it shortly. Thanks for the catch.

milan-cvetkovic:
> I've amended it to remove live.com and will merge it shortly. Thanks for the catch.

Hm, I still see live.com...

tomhughes:
Argh.. Sorry about that... I missed the `-a` switch when amending the commit :-( I've pushed a fix now.

Issue #4773, opened 2024-05-08
Post-merge UI review for #4455


gravitystorm:
This issue is a post-merge UI review for #4455 . cc @milan-cvetkovic 

(I haven't done a post-merge review before, so I'm not sure if there's a better way to do it than to create a new issue - adding a review to a closed PR doesn't seem right either :shrug: )

# Globe positioning

The globe has been repositioned using an arbitrary offset. This appears to be to keep it clear of the tabs, but it doesn't take into account translations.

```diff
   &.new-user-main {
     background-image: image-url("sign-up-illustration.png");
+    background-position-x: 50px;
   }
  ```

![Screenshot from 2024-05-08 16-58-29](https://github.com/openstreetmap/openstreetmap-website/assets/360803/5c2dbf0a-254d-4756-ba92-a1451de846fc)

The offset should be removed, or designed to adapted to different sized tab labels.

# Page width

```diff
+ .auth-container {
+  max-width: 600px;
+ }
```

This is a random width, not based on anything from the bootstrap grid. It makes the login and signup pages inconsistent with the rest of the site.

# Tab alignment

The edge of the tabs should be aligned with the edge of the content. See the traces page for a working example.
![Screenshot from 2024-05-08 17-13-35](https://github.com/openstreetmap/openstreetmap-website/assets/360803/43df826d-d8b8-4e0c-a696-d8d952d6e4f7)

# Top menu buttons

The login/signup buttons disappear from the top menu in certain circumstances, and it leads to jarring inconsistencies. It means the rest of the menu moves around, and it also still doesn't work as intended (e.g. if you make a mistake while signing up, the signup page is shown with errors and the buttons reappear).

```diff
- <% else %>
+ <% elsif (controller_name != "users" and controller_name != "sessions") || action_name != "new" %>
```
This should be reverted. If there's still a desire to remove these buttons (e.g. from the narrow screen layout) then let's discuss that separately.

# auth button sizes

The auth button sizes were reduced from 36px to 24px, but I didn't see any explanation why making them smaller is better?

# sizing for auth_button vs auth_button_preferred

In app/helpers/user_helper we no have two different ways to describe the size of the icon, these should be consistent:

```
                :size => "24"),
```
```
                :width => "24px",
                :height => "24px")
```

# bg-white

The "Log into OpenstreetMap to access..." panel uses `bg-white`, but text on a forced-white background doesn't work in dark mode
[...]

PR #4779, opened 2024-05-10
Use consistent sizing for social signup buttons

milan-cvetkovic:
As suggested in #4773


tomhughes:
Looks good to me, thanks.

PR #4781, opened 2024-05-12
Use bg-body in "log in / sign up to access" messages

AntonKhorev:
Fixes "bg-white" from #4773.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #4782, opened 2024-05-12
Remove unused map sprite image

AntonKhorev:
Not in use since https://github.com/openstreetmap/openstreetmap-website/commit/6619b777e353cbdfdcc6f7ac0e77f8bca9f78656


tomhughes:
Nice bit of ancient history there...

PR #4789, opened 2024-05-14
Add permalinks to changeset comments

### LINKED ISSUE #2796: Changeset comments have individual links, but no button to grab them with

gy-mate:
Fixes #2796

### DIFF
--- a/app/views/changesets/show.html.erb
+++ b/app/views/changesets/show.html.erb
@@ -52,6 +52,12 @@
                              :data => { :method => comment.visible ? "DELETE" : "POST",
                                         :url => api_changeset_comment_visibility_path(comment) } %>
             <% end %>
+            <a href="#c<%= comment.id %>">
+              <svg width="16" height="16" fill="currentColor">
+                <path d="M4.715 6.542 3.343 7.914a3 3 0 1 0 4.243 4.243l1.828-1.829A3 3 0 0 0 8.586 5.5L8 6.086a1 1 0 0 0-.154.199 2 2 0 0 1 .861 3.337L6.88 11.45a2 2 0 1 1-2.83-2.83l.793-.792a4 4 0 0 1-.128-1.287z" />
+                <path d="M6.586 4.672A3 3 0 0 0 7.414 9.5l.775-.776a2 2 0 0 1-.896-3.346L9.12 3.55a2 2 0 1 1 2.83 2.83l-.793.792c.112.42.155.855.128 1.287l1.372-1.372a3 3 0 1 0-4.243-4.243z" />
+              </svg>
+            </a>
           </small>
           <div class="mx-2">
             <%= comment.body.to_html %>
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -380,6 +380,11 @@ $(function () {
       return;
     }
 
+    // Open local anchor links as normal.
+    if ($(this).attr("href")?.startsWith("#")) {
+      return;
+    }
+
     // Ignore cross-protocol and cross-origin links.
     if (location.protocol !== this.protocol || location.host !== this.host) {
       return;


app/views/changesets/show.html.erb
```diff
@@ -33,7 +33,7 @@
         <li id="c<%= comment.id %>">
           <small class='text-muted'>
             <%= t comment.visible ? ".comment_by_html" : ".hidden_comment_by_html",
-                  :time_ago => friendly_date_ago(comment.created_at),
+                  :time_ago => link_to(friendly_date_ago(comment.created_at), "https://www.openstreetmap.org/changeset/#{@changeset.id}##{comment.id}"),
```
HolgerJeromin, line 36:
```suggestion
                  :time_ago => link_to(friendly_date_ago(comment.created_at), "/changeset/#{@changeset.id}##{comment.id}"),
```
But there are probably more rail-ish ways to create that url....

gravitystorm: You can generally get the path to an object using `*_path` helpers, e.g. `changeset_path`. So changeset_path(changeset) will link to the changeset. If you need an anchor (i.e. a `#something` in the url) then have a look through the views for other places that use the word "anchor", like for diary entry comments.

You will also find that in some circumstances you can simplify the `link_to("foo", changeset_path(changeset)) further, since (most of the time) rails knows how to build a path for an object, e.g. `link_to("foo", changeset)`. You can check other places in the existing codebase to learn what works.

AntonKhorev:
There was a previous attempt https://github.com/openstreetmap/openstreetmap-website/pull/4134. It had the advantage of NOT adding link to the time, in case clicking the time is supposed to do something else like https://github.com/openstreetmap/openstreetmap-website/pull/4361.

app/views/changesets/show.html.erb
AntonKhorev: Url fragments are incorrect. They are `#c#{comment.id}`, not `##{comment.id}`.

```suggestion
                  :time_ago => link_to(friendly_date_ago(comment.created_at), :anchor => "c#{comment.id}"),
```

gy-mate:
I'm sorry for the merge commits—I'll pay attention to this next time! 🤞

gy-mate:
> There was a previous attempt #4134. It had the advantage of NOT adding link to the time, in case clicking the time is supposed to do something else like #4361.

@AntonKhorev You're right—I've tried to go this way. Does it seem okay to you?

HolgerJeromin (approved):
But I did not tested this.

tomhughes:
Can you rebase this on current master please, to get rid of the merge commits, and also squash any fixup commits into the commit they are fixing.

tomhughes:
Given that it winds up with one tiny change, yes, that is perfect.

Do we actually use that icon anywhere else at the moment? It seems very bold to me, and poorly aligned:

image

gy-mate:
> Do we actually use that icon anywhere else at the moment?

The regex search `\Wicon link` on the whole codebase yields only this result. So I don't think so.

> It seems very bold to me

@tomhughes Shall I change https://github.com/gy-mate/openstreetmap-website/blob/2db5b718047d8dfb494fdc8fdd6012bf73f3f0ab/app/assets/images/sprite.svg#L30 to https://github.com/twbs/icons/blob/31673a7249bf9043f065aa5d263fc454946de3d8/icons/link.svg#L2-L3? I've looked it up and it seems that the MIT license (which they use) is compatible with GPLv2 (that this repo uses).

hlfan:
An icon with three chain links (where the middle one is just a line) could be simpler to fine-tune the thickness of.

hlfan:
How about this?
```svg
<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20">
  <path fill="none" stroke="#00F" stroke-dasharray="17.571 6.505 25" stroke-linecap="round" stroke-width="2" d="M12 4a1 1 0 014 4l-3 3A1 1 0 019 7ZM8 16a1 1 0 01-4-4l3-3a1 1 0 014 4Z"/>
</svg>
```

HolgerJeromin:
@hlfan 
Would be nice if we could reuse the color of the text or link. 
Easiest would be to use [`currentColor`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword):
```html
<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20">
  <path fill="none" stroke="currentColor" stroke-dasharray="17.571 6.505 25" stroke-linecap="round" stroke-width="2" d="M12 4a1 1 0 014 4l-3 3A1 1 0 019 7ZM8 16a1 1 0 01-4-4l3-3a1 1 0 014 4Z" ></path>
</svg>

**Comment by hlfan** [comment]
But if the link icon isn't used from the sprite (when directly used) then it should be removed from there. The dark blue from the sprite only makes sense for light mode anyway.

**Comment by gy-mate** [comment]
> That certainly looks better I think.

@tomhughes Done!

**Comment by gy-mate** [comment]
> A lot of sites (eg github markdown rendering) also only show it when your mouse is over the item in question.

If someone can do this, sure! But until then, I think that the addition of a link icon in this form (if I've done it correctly) would still be useful.

File: app/views/changesets/show.html.erb [hunk 2]
```diff
@@ -53,6 +53,9 @@
                                         :url => api_changeset_comment_visibility_path(comment) } %>
             <% end %>
           </small>
+          <a href="#c<%= comment.id %>">
+            <div class="icon link" style="transform: rotate(-45deg);"></div>
```
tomhughes, line 57: Inline styles aren't allowed by our security policy in production - this will either need to be in the stylesheet or we could just change the icon given nothing else is using it and even if something else did we'd want the icons to be consistent?

tomhughes (changes requested):

gy-mate:
> But if the link icon isn't used from the sprite (when directly used) then it should be removed from there.

I think it's better for performance if we keep it in the sprite.

app/views/changesets/show.html.erb
gy-mate: I've moved the `rotate()` function inside the `<g id="link">` tag in `sprite.svg`.

tomhughes:
Thanks - can you just squash that into the previous commit? There's no point in making two separate changes to the sprite.

gy-mate:
@tomhughes Sure, done!

app/assets/images/sprite.svg
```diff
@@ -26,8 +26,9 @@
   <g id="close">
     <path d="M206,5 L205,6 L205,7 L208,10 L205,13 L205,14 L206,15 L207,15 L210,12 L213,15 L214,15 L215,14 L215,13 L212,10 L215,7 L215,6 L214,5 L213,5 L210,8 L207,5 L206,5 z" fill="#CCCCCC"/>
   </g>
-  <g id="link">
-    <path d="M185,6 L183,8 L183,12 L185,14 L187,14 L185,12 L185,8 L190,8 L190,11 L192,11 L192,8 L190,6 L187,6 L185,6 z M192,6 L194,8 L194,12 L189,12 L189,9 L187,9 L187,12 L189,14 L194,14 L196,12 L196,8 L194,6 L192,6 z" fill="#0000FF"/>
+  <g id="link" transform="rotate(-45,103.57601,-214.31998)">
```
AntonKhorev, line 29: You don't need to rotate it manually because Bootstrap Icons already has the rotated version
https://icons.getbootstrap.com/icons/link-45deg/

app/views/changesets/show.html.erb
```diff
@@ -53,6 +53,9 @@
                                         :url => api_changeset_comment_visibility_path(comment) } %>
             <% end %>
           </small>
+          <a href="#c<%= comment.id %>">
+            <div class="icon link"></div>
+          </a>
```
AntonKhorev, line 58: Here's what I'd do:

```suggestion
            <a href="#c<%= comment.id %>">
              <svg width="16" height="16" fill="currentColor">
                <path d="M4.715 6.542 3.343 7.914a3 3 0 1 0 4.243 4.243l1.828-1.829A3 3 0 0 0 8.586 5.5L8 6.086a1 1 0 0 0-.154.199 2 2 0 0 1 .861 3.337L6.88 11.45a2 2 0 1 1-2.83-2.83l.793-.792a4 4 0 0 1-.128-1.287z"/>
                <path d="M6.586 4.672A3 3 0 0 0 7.414 9.5l.775-.776a2 2 0 0 1-.896-3.346L9.12 3.55a2 2 0 1 1 2.83 2.83l-.793.792c.112.42.155.855.128 1.287l1.372-1.372a3 3 0 1 0-4.243-4.243z"/>
              </svg>
            </a>
          </small>
```

This is [the rotated icon](https://icons.getbootstrap.com/icons/link-45deg/). Yes it's going to be repeated as many times as there are comments, but that shouldn't be a big problem with compression. The advantage to inlining is that you can use `currentColor` instead of specifying the color.

AntonKhorev:
What should happen when the link icon is clicked?

app/assets/images/sprite.svg
gy-mate: > actually I wouldn't put it in sprite.svg.

Then the whole sprite would need a restructuring, I think. If you'd like to do so, I'll accept your code suggestion right away!

AntonKhorev:
> A lot of sites (eg github markdown rendering) also only show it when your mouse is over the item in question.

Bootstrap has this effect in their [docs](https://getbootstrap.com/docs/5.3/), but not in the library itself. It takes a few lines of css, they added it in https://github.com/twbs/bootstrap/pull/32953.

app/views/changesets/show.html.erb
gy-mate: Thanks! :)

@tomhughes Would this be okay for you?

gy-mate: Done!

@AntonKhorev Thanks a lot for the suggestion! :)

AntonKhorev (approved):

gy-mate:
Great, thanks! :)

HolgerJeromin (approved):

tomhughes (approved):

PR #4791, opened 2024-05-14
Add warnings when creating notes anonymously

### LINKED ISSUE #244: Remind anonymous users that they can login to get notified of changes to their note
I've seen several anonymous notes that I think came from OSM users or people at least willing to log in; it would be great to provide a link to the login page and remind them that if they want to receive updates for their note they should complete the login or registration process.

AntonKhorev:
Fixes #244.

The goal is to:
- tell users who are not registered and add notes why they might want to register
- tell users who forgot to log in that they might want to log in

image

### DIFF
--- a/app/views/notes/new.html.erb
+++ b/app/views/notes/new.html.erb
@@ -4,6 +4,11 @@
 
 <div class="note">
   <p class="alert alert-info"><%= t(".intro") %></p>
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good to me, thanks.

PR #4775, opened 2024-05-09
Use SVGs instead of PNGs for interface icons where possible

hiddewie:
Part of https://github.com/openstreetmap/openstreetmap-website/issues/850

This pull request replaces PNG icons with their minimized SVG variants, to ensure screens with high pixel densities can display the OpenStreetMap website with beautiful graphics.  

Split off pull requests:
- Terms illustration (http://127.0.0.1:3000/user/terms) https://github.com/openstreetmap/openstreetmap-website/pull/4793
- Sign up illustrations (http://127.0.0.1:3000/user/new) https://github.com/openstreetmap/openstreetmap-website/pull/4793
- Welcome sprites (http://127.0.0.1:3000/welcome) https://github.com/openstreetmap/openstreetmap-website/pull/4796
- Menu "hamburger" icon https://github.com/openstreetmap/openstreetmap-website/pull/4798
- CC BY SA copyright logo in https://github.com/openstreetmap/openstreetmap-website/pull/4795
- ~Avatar default icons~ (does not work with Gravatar)
- OpenID connect icon (http://127.0.0.1:3000/login) https://github.com/openstreetmap/openstreetmap-website/pull/4795
- Routing sprites (http://127.0.0.1:3000/directions?engine=fossgis_osrm_car&route=52.0991%2C4.2970%3B51.9244%2C4.4778) https://github.com/openstreetmap/openstreetmap-website/pull/4794
- OSM icon (except in emails) (http://127.0.0.1:3000) https://github.com/openstreetmap/openstreetmap-website/pull/4799
- Note marker icons https://github.com/openstreetmap/openstreetmap-website/pull/4797

For each of the icons, I replaced the PNG with the corresponding SVG. The PNG variant has been removed. Also, I simplified and minimized the SVGs for improved compression and less data transfer. This also removes privacy sensitive information from the SVG files.

In all cases, I can create separate pull requests if that makes the review process easier.

Let me know if I can add more changes to this pull requests for the remaining images, or if I should create separate pull requests.

### Browser support

I could not find documentation about which browsers are supposed to be supported by the OpenStreetMap website. 

Internet Explorer 11 is the only browser that might have problems displaying SVG graphics related to scaling. This browser is no longer supported.

See https://caniuse.com/svg

In particular I removed the `srcset` for replacing a PNG icon with an SVG icon, because browsers can display SVG icons as source in ``.

Issue #850, opened 2014-12-03
retina compatible icons


fuddl:
on the way to retina support (#104), one thing that has to be done is adding high res icons.

I mean these:
![osm icons](https://cloud.githubusercontent.com/assets/842548/5280619/ed069450-7af3-11e4-8694-64a7f76d24a7.png)

are these available as svg somewhere? or as a font?

Zverik:
Is this issue still relevant?

bcherny:
yes - icons still look bad on retina:

<img width="62" alt="screen shot 2015-09-10 at 11 11 01 am" src="https://cloud.githubusercontent.com/assets/1761758/9796799/d1050610-57ac-11e5-82ce-e355c239d758.png">

can we just serve the SVG instead of PNG? svg is:

- [supported by all modern browsers](http://caniuse.com/#feat=svg)
- future compatible with higher dpi devices that may appear in the near future
- smaller (just 1.1kb gzipped after running through svgo) compared to PNG (3.3kb gzipped)

HolgerJeromin:
I am pretty sure osm.org should be usable with Internet Explorer < 9. 
Thats why i [suggested](https://github.com/openstreetmap/openstreetmap-website/issues/850#issuecomment-65446147) a compatible version.

bcherny:
can you make a pr?

HolgerJeromin:
Made two PR for the two changes and thrown both SVGs into SVGO (nice tool btw :-)
The directions button has no SVG file in the repo. Perhaps @zerebubuth can help here.

The marker (for routing and other stuff) were not changed. These are interacting with the map via drag and drop and i was not brave enough for changing them, too.

HolgerJeromin:
could we close this or should we wait for 
https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/images/directions.png

tomhughes:
Well this is unfortunately the kind of ticket that is not sufficiently concrete to be easily closeable. There are many image that are still not retina compatible, and even some of the ones we have added patches for are not really sorted.

In particular I discovered after I merged your `picture` based fix for the logo that one of the two main browsers where retina support is needed (iOS Safari) doesn't actually support `picture` elements at all, so we really need to revisit that.

HolgerJeromin:
[...]

PR #4793, opened 2024-05-14
Replace sign up, terms and confirmation illustrations with SVG

hiddewie:
Part of https://github.com/openstreetmap/openstreetmap-website/issues/850

Split off from https://github.com/openstreetmap/openstreetmap-website/pull/4775

This pull request replaces PNG images for the signup and terms illustrations with their minimized SVG variants, to ensure screens with high pixel densities can display the OpenStreetMap website with beautiful graphics.  

~TODO: the confirm illustration (from https://github.com/openstreetmap/openstreetmap-website/pull/4775#issuecomment-2106310923)~

The illustrations have all been resized to the same size and aligned on one corner.

Sign up page (http://127.0.0.1:3000/user/new):

image

Confirmation page (http://127.0.0.1:3000/user/iertn/confirm):

image

Terms page (http://127.0.0.1:3000/user/terms)

image

PR #4794, opened 2024-05-14
Routing sprites to SVG

hiddewie:
Part of https://github.com/openstreetmap/openstreetmap-website/issues/850

Split off from https://github.com/openstreetmap/openstreetmap-website/pull/4775

This pull request replaces PNG icons for the routing to their minimized SVG variants, to ensure screens with high pixel densities can display the OpenStreetMap website with beautiful graphics.  

Tested on http://127.0.0.1:3000/directions?engine=fossgis_osrm_car&route=52.0991%2C4.2970%3B51.9244%2C4.4778#map=11/52.0130/4.3863&layers=N

image


tomhughes:
Is there a good reason to minimize the source when we run them all through `svgo` before serving them in production?

hiddewie:
> Is there a good reason to minimize the source when we run them all through svgo before serving them in production?

2 main reasons for me:
- For example https://www.openstreetmap.org/assets/sprite-e144d5cba10d95411535e0f63a538f760b299585fa9726c2ba8a37e3a6164300.svg and https://www.openstreetmap.org/assets/osm_logo-4afddaae0230a5a46687fdc751ed256dfdccde144118cb02a7d7960f207a4b92.svg (2 SVGs fetched from the frontpage): they contain unneeded attributes, whitespace and even some personal information like filenames. Minimizing them with a tool, and then by hand, makes the files smaller than can be achieved automatically. Maybe the tool can be configured better to remove unneeded attributes, but this also works.
- Reviewing changes is easier when the SVG content is much smaller. Although I must admit that having newlines instead of a single huge long line is easier to read. I might add in newlines again in all these SVGs.

tomhughes:
Being able to review changes is something that definitely counts against full minimisation. I don't mind it being reduced to a nicely formatted handwritten SVG standard but full machine minimisation will make review of future changes hard.

hiddewie:
Yes I agree. In the entire SVG series I have re-added all newline/indent whitespace to ensure the SVG itself looks somewhat nice to humans.

A simple diff should be reviewable code-wise, although the illustrations in e.g. https://github.com/openstreetmap/openstreetmap-website/pull/4793/files#diff-eac95d76558af100602bc5274b9b685530cba96ce5abb5fddeb5f3b3cfa5c215 will remain big blobs of SVG code.

tomhughes:

hiddewie:
Thanks for all the reviews!

PR #4799, opened 2024-05-14
OSM icons for website to SVG

hiddewie:
Part of https://github.com/openstreetmap/openstreetmap-website/issues/850

Split off from https://github.com/openstreetmap/openstreetmap-website/pull/4775

This pull request replaces PNG images for the OSM logo with their minimized SVG variants, to ensure screens with high pixel densities can display the OpenStreetMap website with beautiful graphics.  

Email has not been changed because SVG support in e-mails is difficult (ref https://github.com/openstreetmap/openstreetmap-website/pull/4775#issuecomment-2102560613).

image


tomhughes:
Looks good to me. How sure are we that all OpenGraph consumers are going to cope with SVG icons?

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/4793#issuecomment-2112619979

What would you rather edit
image or image ?

Some people want to edit the logo: https://community.openstreetmap.org/t/binary-numbers-on-osm-logo-could-be-changed-to-have-a-real-meaning/100498

hiddewie:
> copy+paste error in the issue description

Oops, fixed. I was indeed copy-pasting all of the PR descriptions as I split them off from the initial PR.

> Some people want to edit the logo: https://community.openstreetmap.org/t/binary-numbers-on-osm-logo-could-be-changed-to-have-a-real-meaning/100498

Yes indeed, I checked the history of the logo:

image

image

image

So in more than 10 years, the logo has been created and modified once. The other 2 commits are renames.

---

On the topic of using a tool like `svgo` for minifying / optimizing such SVGs. I tried this locally, and I found some problems with this approach, taking the source design files and using them directly to produce (minified) output for the web.

Problems:
- Clips sometimes do not work as expected
- Text is not rendered correctly if the font is not present on the system, or not at all
- Images such as in the OSM logo are still persisted (outside the viewable area) into the output SVG.

I ran the `svgo` tool like 
```
svgo --input ... --output ...
```
so with the default plugins enabled. I could not find any plugins in the list of plugins that would help us solve the above problems.

Some example output of the terms SVG: terms

My conclusion is that the `svgo` tool cannot solve the SVG minimization problem right from the source illustration files. So if the design illustrations should be kept intact, then they should be kept in a different directory (where `assets/` in the project root?).

Then the process to convert such a designer SVG file into a file ready for minimization, would be: 
- Convert all text to paths
- Remove clips by intersecting the paths in the clip with the clip mask
- Remove any content outside the viewbox
- Remove any image content

Then, I put the files in https://vecta.io/nano to minimize them, but they could also be processed by `svgo` (as long as all plugins are enabled, some are disabled at the moment in https://github.com/openstreetmap/openstreetmap-website/blob/master/config/image_optim.yml#L5).

After that I auto-formatted (Jetbrains IDE) the minimized SVGs to make them nicer to read for humans with newlines and indent.

hiddewie:
It seems that Facebook does not like SVGs as opengraph images, so I will revert that one as well.

tomhughes:
I assume the `cleanupIDs` svgo plugin is disabled because we have things that use ids to pick out parts of sprite images for display and `removeUnknownsAndDefaults` is disabled because of https://github.com/openstreetmap/iD/issues/4105.

gravitystorm:
> [Merge branch 'master' into retina-svg-osm-logo](https://github.com/openstreetmap/openstreetmap-website/pull/4799/commits/2fae97966469d764a8ae27bcaa606c02a29892e9)

Please avoid merge commits in PRs - you can rebase your changes onto the current master branch instead.

https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests

> Avoid including "merge" commits. If your PR can no longer be merged cleanly (for example, an unrelated change to Gemfile.lock on master now conflicts with your PR) then please rebase your PR onto the latest master. This allows you to fix the conflicts, while keeping the PR a straightforward list of commits. If there are no conflicts, then there is no need to rebase anything.

AntonKhorev:
merged, thanks

hiddewie:
Thanks!

PR #4802, opened 2024-05-15
Routing sprite cleanup

AntonKhorev:
Remove some things that are only there because of the svg editor like transforms with fractional pixel values applied to paths with fractional pixel values.


tomhughes:
Is getting rid of the groups a good idea? They make it clear where each individual sprite begins and ends and allows them to be selected in an editor?

AntonKhorev:
First, second and third `<path>` elements colored:
image

tomhughes:
Fine, in that case I guess there's no reason to keep them.

PR #4808, opened 2024-05-17
Fix white circular elements of welcome sprite

AntonKhorev:
Before / after:
image image


tomhughes:
I noticed this looked wrong in eog when looking at it the other day though oddly the inline viewer here doesn't seem to show the same problems and neither did inkscape... Thanks for the fix anyway.

PR #4810, opened 2024-05-18
Welcome page html and css cleanup

AntonKhorev:
Mostly I removed custom css and unnecessary markup, but I also changed the image placement from inlined/floated to their own column.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

Issue #4806, opened 2024-05-17
Improve the text of a mail about comments to a changeset


deevroman:
### Problem

Inspired by [comment](https://t.me/ruosm/816094) in the community chat: 

> It took me 2 months and a block before I found how to reply. I see the comments on the mail, I take note, but how to reply is unclear.

<img width="784" alt="Снимок экрана 2024-05-17 в 10 35 11" src="https://github.com/openstreetmap/openstreetmap-website/assets/25708359/89493797-fc44-4562-bf3e-cd924114cef8">

What if the text "More details about the changeset can be found at %{url}." is replaced with: "Reply or learn more about the changeset at %{url}"?

HolgerJeromin:
The text is defined here:

https://github.com/openstreetmap/openstreetmap-website/blob/41ed1f8e528406c47b6a0afe3b6deef2de34f411/config/locales/en.yml#L1692

AntonKhorev:
Currently it's similar to note notifications that also don't say how to reply.

kleenkanteen:
@HolgerJeromin How does one test sending an email locally?

AntonKhorev:
> Why is the text duplicated on lines 1692-1693 in the file you quoted?

Emails contain both html and plaintext.

AntonKhorev:
> How does one test sending an email locally?

If you want to test the mailer output, see for example https://github.com/openstreetmap/openstreetmap-website/blob/master/test/mailers/user_mailer_test.rb

AntonKhorev:
> Also is there a script that allows one make edits to a text, say for this sentence for example, in all languages using a translation API?

https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#i18n

sjvudp:
Sorry, if this is not exactly on-topic, but:
I frequently get messages about responses to an OSM note (hint). As it lacks the original note, I can only guess from the mail message what it might refer to. To find out, I'll have to follow the link.
Working on a desktop, but reading mail on the mobile (while on the road), I cannot easily reply to the comments I still would have to log in to OSM from the mobile phone (but I don't have the credentials there, because I use a random passwords only my password manager on the desktop can remember.
So it would be nice if the link in the mail message had some magic access token embedded that does not require login, just to respond to a comment.
[...]

PR #4812, opened 2024-05-18
chore: clarify email reply text

kleenkanteen:
For #4806

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1673,8 +1673,8 @@ en:
         your_note_html: "%{commenter} has reactivated one of your map notes near %{place}."
         commented_note: "%{commenter} has reactivated a map note you have commented on. The note is near %{place}."
         commented_note_html: "%{commenter} has reactivated a map note you have commented on. The note is near %{place}."
-      details: "Reply or learn more about the changeset at %{url}."
-      details_html: "Reply or learn more about the changeset at %{url}."
+      details: "Reply or learn more about the note at %{url}."
+      details_html: "Reply or learn more about the note at %{url}."
     changeset_comment_notification:
       description: "OpenStreetMap Changeset #%{id}"
       hi: "Hi %{to_user},"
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -1673,8 +1673,8 @@ en:
         your_note_html: "%{commenter} has reactivated one of your map notes near %{place}."
         commented_note: "%{commenter} has reactivated a map note you have commented on. The note is near %{place}."
         commented_note_html: "%{commenter} has reactivated a map note you have commented on. The note is near %{place}."
-      details: "More details about the note can be found at %{url}."
-      details_html: "More details about the note can be found at %{url}."
+      details: "Reply or learn more about the changeset at %{url}."
```
gravitystorm, line 1676: This looks like a copy+paste error - you’ve changed note to change set here (twice)

gravitystorm (changes requested):

config/locales/en.yml
kleenkanteen: woops, fixed

tomhughes:

kleenkanteen:
@tomhughes by the way, shouldn't PR's by default be squashed when merged? there's 2 commits on the master branch when there should be just one imo for a cleaner history:

image

There's a github setting that allows only squash merging:

image

Also why is there no development branch? That way, each commit to the master branch would be major changes, while each commit to to the development branch would be minor changes like my PR here.

tomhughes:
No, you should ideally have squashed them together as explained at https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests.

I didn't notice you hadn't done that or I would have asked you do so. I don't actually merge on github so the github settings are not relevant.

gravitystorm:
> Also any thoughts on my developement branch question in my above post?

We merge all accepted PRs to master immediately, so that they can be immediately deployed, and the community immediately benefits from them. It doesn't matter whether the change is big or small, there's no reason to delay merging into master or to create extra complexity by having multiple long-running branches.

Issue #2869, opened 2020-10-01
Changeset comments show italics for non-Latin scripts


fqdb:
1. There are not supposed to be italic Chinese characters, and probably the same goes for the majority of non-Latin scripts
2. It's extremely difficult to read, especially because it breaks anti-aliasing in some browsers
3. At least in my experience even the Latin italics are less legible than normal text, not sure why this was changed at all.

![image](https://user-images.githubusercontent.com/2826336/94788541-006c9c00-03d4-11eb-85dd-fbf7afcc5950.png)

tomhughes:
Well you will find the history in #580 and #2838.

schuetzm:
Maybe the `font-synthesis` property [1] would help here? But it seems it's still a proposal only supported by FF and Safari, and could have the side-effect of disabling italics with some Latin fonts, too.

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/font-synthesis

HolgerJeromin:
I really like font-synthesis. 
Here we have the relevent issue in chromium for missing support in Chrome, Edge, Vivaldi, Opera
https://bugs.chromium.org/p/chromium/issues/detail?id=509989

Dimitar5555:
As a partial workaround, you can set a custom font for specific alphabets via Firefox's preferences and force websites to use it. You can't fine tune it to a specific website, but it's better than nothing and makes the text a lot more readable.

![image](https://github.com/openstreetmap/openstreetmap-website/assets/19364673/427a9cec-b260-40e3-9d29-0b8bed060114)

AntonKhorev:
Italics replaced with larger font in #4813.

PR #4813, opened 2024-05-18
Use font size instead of italic for changeset/trace descriptions

AntonKhorev:
The problem and links to previous solutions are in #2869. Changeset descriptions were made to use italics "to make them look more like quotes" but that doesn't work well with Chinese and maybe other writing systems and large blocks of italics don't look good. How else can the descriptions be rendered? I tried a few options:

- Bold font: looks too bold. Intermediate font weights: not clear if they are going to be available with every font and writing system that might be in use.
- Decorations like underline: can't use them because descriptions sometimes are links, sometimes are not links, sometimes they contain links inside.
- Colors: also don't look right. Primary color is blue, that looks like a link. Secondary is gray we use for deemphasising. Other colors we mostly don't use, they stick out too much.

How about font size then? Let's make the font slightly bigger.

However that may make another problem worse. The text is now more likely not to fit horizontally. You can't always force words to wrap, but then horizontal scrolling should work fine just for the overflowing descriptions. Except it won't work fine when stretched links are used, and they are used in changeset lists. In this case let's truncate the text while still letting it to wrap if possible.

Screenshots in Chromium, because it shows scrollbars:
image
image
image

### DIFF
--- a/app/views/browse/_common_details.html.erb
+++ b/app/views/browse/_common_details.html.erb
@@ -7,7 +7,7 @@
   #<%= link_to_unless_current common_details.version, :controller => "old_#{@type.pluralize}", :action => :show, :version => common_details.version %>
 </h4>
 
-<p class="fs-6">
+<p class="fs-6 overflow-x-auto">
   <% if common_details.changeset.tags["comment"].present? %>
     <%= linkify(common_details.changeset.tags["comment"]) %>
   <% else %>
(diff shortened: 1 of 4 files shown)


gravitystorm:
Great, I think this is an improvement on what I originally implemented.

maro-21:
Finally no italics!! Thanks a lot!

PR #4816, opened 2024-05-19
Make the login and signup screens focus on the first input

### LINKED ISSUE #4814: Cursor not in first input field of login form
### URL

https://www.openstreetmap.org/login?referer=%2F#map=6/51.330/10.453

Originally described in German: https://community.openstreetmap.org/t/cursor-im-login-formular-in-das-erste-eingabefeld-positionieren-was-spricht-dagegen/112560

When opening the login form in Firefox, no cursor is visible; specifically it's not visible in the first input field.
So typing goes to "somewhere", but not to the first (username) input field. When using KeePass auto-type, the password might be pasted into the username field.

Happens in Edge also, BTW.

Examining the login form, the first input field is a hidden one. Maybe the simplest solution is to place the input field after the "real" ones. Also it's debatable whether a hidden input field should get the focus (and invisible cursor) if it's the first one.

grafik

tomhughes:
Fixes #4814

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -35,7 +35,7 @@
   <%= bootstrap_form_tag(:action => "login", :html => { :id => "login_form" }) do |f| %>
     <%= hidden_field_tag("referer", h(params[:referer]), :autocomplete => "off") %>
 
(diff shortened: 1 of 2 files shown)


gravitystorm:
Thank you.

PR #4822, opened 2024-05-21
Remove inbox template and anchor id

AntonKhorev:
This partial template used to be included from different places, but now it's included only from header. I want to make changes to it that are specific to the account button. It would make no sense for them to be in a different file.

Also `id="inboxanchor"` is not necessary after Turbo updates.


tomhughes:
Looks good to me, thanks.

PR #4823, opened 2024-05-21
Use .icon-link to align logo

AntonKhorev:
Removes pixel sizes specific to the logo.

`.icon-link` also makes the gap between the logo and "OpenStreetMap" slightly bigger:
image
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -107,14 +107,8 @@ header {
     padding: $lineheight * 0.5;
   }
 
-  img.logo {
-    margin-top: -2px;
-  }
-
   h1 {
     font-size: 18px;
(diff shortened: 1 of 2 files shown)


tomhughes (changes requested):
This causes the text part of the logo to get a link underline when hovered, which it didn't have before... The text also moves down slightly relative to the icon which I'm not sure about - the alignment looks slightly worse to me?

AntonKhorev:
Does it really move down?

image image

image

AntonKhorev:
Restored the original horizontal gap:
image
image

tomhughes (approved):
There definitely is a slight movement vertically - if you have the old and new versions open in two tabs and flip back and forth you can see it. I don't think it's enough to worry about though.

PR #4824, opened 2024-05-21
Remove unnecessary user menu wrapper

AntonKhorev:
Was used to replace the inbox counter before Turbo:
https://github.com/openstreetmap/openstreetmap-website/commit/1a7b7f5d18931f60f4171b7ac410ea4434828dd1#diff-305c268f66d28c4cd669b9a69449100d2de05f2b61da1e58914295c1eea45e16L22

`<span class='username'>` is also unnecessary but I'm going to use it later to limit the username width.


gravitystorm:
Great, thanks.

PR #4825, opened 2024-05-22
Sticky close button

### LINKED ISSUE #2174: Allow closing side panels even when not scrolled all the way back to the top
1. Scroll the left side panel on [example](https://www.openstreetmap.org/way/28461079#map=10/14.6882/121.3110).
1. Attempt to close the left side panel.
2. One cannot find the X to close it, unless one scrolls all the way back up to the top.

AntonKhorev:
Allows close buttons to stay visible when sidebar contents is scrolled.

image image

image image

Fixes #2174.

Also related to [this](https://github.com/openstreetmap/openstreetmap-website/pull/4647#issuecomment-2123124562) because the button is not inside `sidebar_content` and is not being resent with page templates.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -401,7 +401,7 @@ $(document).ready(function () {
     }
   });
 
-  $(document).on("click", "#sidebar_content .btn-close", function () {
+  $(document).on("click", "#sidebar_close button", function () {
     OSM.router.route("/" + OSM.formatHash(map));
   });
 });
--- a/app/assets/javascripts/leaflet.sidebar-pane.js
+++ b/app/assets/javascripts/leaflet.sidebar-pane.js
@@ -20,14 +20,9 @@ L.OSM.sidebarPane = function (options, uiClass, buttonTitle, paneTitle) {
     var $ui = $("<div>")
       .attr("class", uiClass + "-ui");
 
-    $("<div class='d-flex p-3 pb-0'>")
-      .appendTo($ui)
-      .append($("<h2 class='flex-grow-1 text-break'>")
-        .text(I18n.t(paneTitle)))
-      .append($("<div>")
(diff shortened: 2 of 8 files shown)


tordans:
I think we should update the design of the circle to make it more clear when scrolled. Maybe a light gray background (for non hover) with a white border for some separation to the background?

AntonKhorev:
The circle doesn't have a hover state, it's not clickable. If I give it a gray background or a border it will look like clickable.

If you actually want the circle to be clickable, you want a larger close button. Do you want it? I didn't change the button size in this pull request.

tordans:
The transparent background suggest that it is clickable as a whole element already. Changing it to behave like that would be an UX improvement to me, yes. And would also make the scrolled-state more clear.

AntonKhorev:
> The transparent background suggest that it is clickable as a whole element already.

My goal was for the background was to suggest the opposite.

tordans:
> You can open another issues about that.

This PR makes the button sticky. I agree that the sticky button needs a visual separation to the content below. I don't think that it is good UX to make this area non-interactive (hover, click) – it's not what I would expect the interface to behave.

Unfortunately Bootstrap does not follow those UX decisions for their close button https://getbootstrap.com/docs/5.0/components/close-button/ (which only shows the non sticky version, but still…). However they do use a larger surface area for their burger menu https://getbootstrap.com/docs/5.0/components/navbar/#external-content which follows the same principle, IMO.

AntonKhorev:
> Unfortunately Bootstrap does not follow those UX decisions for their close button

If anyone wants larger close buttons, it's not difficult to do.
image

> However they do use a larger surface area for their burger menu

Their burger button is similar to other outline buttons. You can remove border and padding if you want.
image

tomhughes:
Superseded by #5778.

PR #4826, opened 2024-05-23
Fixed "Tab alignment" issue described in #4773

nenad-vujicic:
This PR addresses "Tab alignment" issue mentioned in the #4773

I removed px-3 class from ul tag defining area in which are "Log in" and "Sign up" tabs.

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -11,7 +11,7 @@
   <% end %>
 
   <div class="header-illustration new-user-main auth-container mx-auto">
-    <ul class="nav nav-tabs position-absolute bottom-0 px-3 fs-6 w-100">
+    <ul class="nav nav-tabs position-absolute bottom-0 fs-6 w-100">
       <li class="nav-item">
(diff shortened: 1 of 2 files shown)


AntonKhorev:
`px-3` also on the `/user/new` page.

tomhughes (changes requested):
As @AntonKhorev pointed out this needs to be done for the signup page as well or the tabs will move as you switch between the two.

nenad-vujicic:
Dear @AntonKhorev, @tomhughes ,

Thanks for reviewing! I apologize for non-complete testing first time. Now, both tabs will be aligned.

Please, let me know if you find another place which could be improved.

Thanks,
Nenad.

tomhughes (changes requested):
That's fine now thanks, but there's no need for a separate commit. If you can just squash that into the first commit then I'll merge it.

nenad-vujicic:
Dear @tomhughes,

They are squashed now into the first commit. Thank you very much!

Thanks,
Nenad.

tomhughes:
Thanks - that is merged now.

PR #4828, opened 2024-05-24
Adjust vertical alignment of secondary header navigation

AntonKhorev:
Like #4723 + adjustments inside the user button + increased horizontal spacing between menu items (in #4723 the spacing decreased because with `display: flex` whitespace between items gets removed).

I get rid of `h1, nav, nav > ul, nav > ul > li` selector and do other changes in order to switch later to Bootstrap navbar and decrease the number of menu/sidebar visibility modes.

Before / after when not logged in:
image
image

Before / after with issues counter:
image
image

Before / after with message counter:
image
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -150,7 +150,7 @@ nav.primary {
 
 nav.secondary {
   .nav-link {
-    padding: 0.2rem;
+    padding: 0.3rem;
   }
 }
 
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -98,10 +98,6 @@ header {
   position: relative;
   font-size: 14px;
 
-  h1, nav, nav > ul, nav > ul > li {
-    display: inline-block;
-  }
-
   > * {
     height: 100%;
     padding: $lineheight * 0.5;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/common.scss
```diff
@@ -191,7 +187,7 @@ body.small-nav {
     min-height: $headerHeight;
 
     &.closed nav {
```
tordans, line 189: I wonder: do we use this .closed in a lot of places? Otherwise we could use a utility class on nav to hide it, right?

AntonKhorev: It's used only here. It controls the visibility in response to the menu button. You wouldn't need any class for that, just show/hide from javascript. It was done in #4365.

tordans: Sounds good. I like that #4365 make it that we stay in one world to toggle the visibility and not scatter logic between JS and CSS (and HTML).

AntonKhorev: I'm not pursuing https://github.com/openstreetmap/openstreetmap-website/pull/4365 right now because maybe I'll go straight to [Bootstrap navbar](https://getbootstrap.com/docs/5.3/components/navbar/) which has its own javascript.

tomhughes:
Looks like a nice improvement to me. Thanks.

PR #4829, opened 2024-05-24
Keep header h1 height equal to default header height

AntonKhorev:
In small mode logo+h1 move up slightly. It seems like they shouldn't because there's `> * { height: 100%; ...}` css, but it doesn't do anything. What actually kept them from moving was `padding-top: 15px;` which I removed in #4823. I'm not going to put this padding back in. Instead I just set their height directly.

Before / after:
image image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -103,11 +103,11 @@ header {
   }
 
   > * {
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #4832, opened 2024-05-25
Remove custom css from "Enable overlays" text

AntonKhorev:
I use `.small` font size from Bootstrap which is slightly smaller than the current custom 13px size.

Before / after:
image image


tomhughes:
Looks good to me, thanks.

PR #4833, opened 2024-05-25
Remove custom css from note descriptions

AntonKhorev:
- left margin is similar to what's being used on note comments
- bottom margin - I don't think it's required here because paragraphs inside descriptions already have bottom margins

Before / after:
image image


tomhughes:
Looks good to me, thanks.

PR #4835, opened 2024-05-26
Remove custom css from "Load Data" button

AntonKhorev:
image

PR #4836, opened 2024-05-26
Remove unused .browse_status css class

AntonKhorev:
#4835 removes `#browse_status` but there's also `.browse_status`. I can't find any use of this class at any time.


tomhughes:
Looks good to me, thanks.

PR #4837, opened 2024-05-26
Replace .btn-wrapper with Bootstrap gutters/gaps

AntonKhorev:
See the deleted css comment for why it was used. You can achieve the same effect with [gutters](https://getbootstrap.com/docs/5.3/layout/gutters/) when using Bootstrap rows or with `.d-flex` + [gaps](https://getbootstrap.com/docs/5.3/utilities/spacing/#gap).

Account settings in a very narrow window:
image

Note edit buttons:
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -637,16 +637,6 @@ tr.turn {
   }
 }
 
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me, thanks.

PR #4838, opened 2024-05-26
Remove unused #container css

AntonKhorev:
Like #4836 I can't find it ever being used. Maybe it was meant to be `<div id="content">`?


tomhughes:
Looks good to me, thanks.

Issue #2233, opened 2019-05-25
add "edit" button to mobile version


ralfhauser:
on https://www.openstreetmap.org/node/846169263, I would like to add wheelchair=yes tag when on my mobile.
The responsive design unfortunately hides the "edit" button - so I cannot.
(I am logged in on my smart-phone)

HolgerJeromin:
My default on android is: install the editor app Vespucci, after that use the "geo uri" in the share part of the website.

ralfhauser:
I see the benefit of an app for power users, but for infrequent/novice users a minimal web interface would be good too ?

tomhughes:
Enabling the button wouldn't really be very useful because iD doesn't really work in a useful way on a screen that small. Here's what it would look like on an iPhone 8+ for example:

![Screenshot_2019-05-25 OpenStreetMap](https://user-images.githubusercontent.com/147741/58365707-0b4e1d00-7ec0-11e9-8402-149001ac74cf.png)

mmmatthew:
Here is a comment about adding mobile support to iD: https://github.com/openstreetmap/iD/issues/5890#issuecomment-464382886

mmd-osm:
You could try the master preview page: http://preview.ideditor.com/master

BjornRasmussen:
@tomhughes just press the inspect button on the top and iD becomes a lot better.  I've been able to do a lot with iD lately, including adding service roads and buildings, using the validator to upgrade tags, and adding POIs, all from my smartphone.
The biggest barrier is getting to iD from the osm website.

ralfhauser:
In my experience, 90% of novices are lost by not downloading an app from the app-store.

Questions:
1) is there any https://en.wikipedia.org/wiki/Progressive_web_applications as an interim step if someone wants to add "wheelchair"
2) if I could organize the financing for enabilng the pure web interface to allow to edit such tags, would that help ?

Adamant36:
"In my experience, 90% of novices are lost by not downloading an app from the app-store." 

I think most people edit OSM at this point probably also have OsmAnd installed on their cell phones. You can edited POI's pretty easily through that by holding down on the POI, clicking actions, and pushing the "modify POI" button. 

[...]

PR #4839, opened 2024-05-26
Show edit and export on small screens

tordans:
I looked into making the edit button visible on small screen https://github.com/openstreetmap/openstreetmap-website/issues/2233 and ran into a few issues that are making this less straight forward than I was hoping it to be.

1. Bootstrap does not provide responsive classes to change the button group to vertical. 
    - There is an issue on this at https://github.com/twbs/bootstrap/issues/39213
    - We could change the layout of course. 
    - We could also change the classes in JavaScript, which I did for this draft.
3. Bootstrap has a bug with vertical button list when the first button is a dropdown. 
   - I reported this in https://github.com/twbs/bootstrap/issues/40487 and my PR https://github.com/twbs/bootstrap/pull/40488 might fix it. 
   - I added this fix as a workaround to our css for this draft.
5. Bootstrap does not support vertical button groups with split buttons ATM. 
   - See https://getbootstrap.com/docs/5.2/components/button-group/#vertical-variation _"Split button dropdowns are not supported here."_. 
   - We can work around this by adding a few utility classes, which I did for this draft. 
   - But since the way we do responsive design on the website is a bit special, there are slight changes to what we have right now in certain edge case screen sizes. We could use more JS to fix this if needed…
7. Bootstrap does not handle hidden items in a group.
   - We are hiding the "Export" Page on the button group via CSS ATM and Bootstrap does not handle the case of hidden elements, so the button group borders are wrong. 
   - For this solution I made all menu items visible. Following the logic that the reason why we have an export page so prominently on the page at all is to showcase the story of OSM, which is just as true on mobile. Hiding part of the story is worse than showing it so users can learn about it an realize they need a different device to use this part of the website. – We could continue hiding the page, of course, but need to add more custom css or JS to make bootstrap work.

**This is what it looks like:**

**Logged out:**

image

image

**Logged in:**

Bildschirmfoto 2024-05-26 um 07 17 19

Bildschirmfoto 2024-05-26 um 07 17 27

Bildschirmfoto 2024-05-26 um 07 17 37

---

**Where do go from here?**

I consider hiding the Edit button in 2024 a bug which we should fix. And this PR is a way to do this.

In the near future we will need to reconsider this part of the menu, especially the screen usually visible on mobile to incorporate editors in the menu (https://github.com/openstreetmap/operations/issues/877#issuecomment-2130858478).

---

**Open issues:**

- [ ] When changing the screen width without reload, the class change from "button group" to "vertical button group" is not applied. 
    image
    I can fix since depending on the feedback on this PR.

---

Closes https://github.com/openstreetmap/openstreetmap-website/issues/2233

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -79,8 +79,11 @@ $(document).ready(function () {
 
     if (windowWidth < compactWidth) {
       $("body").removeClass("compact-nav").addClass("small-nav");
+      // Toggling the classes is needed until https://github.com/twbs/bootstrap/issues/39213 is added
+      $("nav.primary > .btn-group").removeClass("btn-group").addClass("btn-group-vertical");
     } else if (windowWidth < headerWidth) {
       $("body").addClass("compact-nav").removeClass("small-nav");
+      $("nav.primary > .btn-group").removeClass("btn-group-vertical").addClass("btn-group");
     } else {
       $("body").removeClass("compact-nav").removeClass("small-nav");
     }
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -199,9 +199,7 @@ body.small-nav {
     }
   }
 
-  #sidebar .search_forms,
-  #edit_tab,
-  #export_tab {
+  #sidebar .search_forms {
     display: none;
   }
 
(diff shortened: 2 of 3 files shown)


AntonKhorev:
Why do you wan a vertical button group? Doesn't it take too much space?

AntonKhorev:
"Export" probably shouldn't be in this button group at all. With Export removed buttons should fit on a screen as a horizontal group.

AntonKhorev:
I guess we don't need this PR now?

PR #4840, opened 2024-05-26
allow data URIs for images in iD

tyrasd:
Fixes problems like https://github.com/openstreetmap/iD/issues/10259.

This is a slight regression in #4627, as the code previously _appended_[^1] the additional CSP rules and now they are overwritten. I assume this was an oversight in the migration process, or was the omission of the existing `img-src` rules there on purpose? In that case, the `:data` type should be added as `policy.img_src("*", :blob, :data)`.

[^1]: see https://github.com/openstreetmap/openstreetmap-website/pull/4627/files#diff-c7e7144de4dcf44dbb148d4acc6aa4a9d8581c6a581966a0bbfb598e79730f14L139

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -18,7 +18,7 @@ class SiteController < ApplicationController
 
   content_security_policy(:only => :id) do |policy|
     policy.connect_src("*")
-    policy.img_src("*", :blob)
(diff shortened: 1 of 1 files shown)


tomhughes:
Apologies for that, and thanks for the catch. I think I probably thought that as`*` matched anything I didn't need to worry about inheriting the defaults and forgot that would lose `data:` from the rules.

PR #4841, opened 2024-05-26
update script-src CSP rules for iD

tyrasd:
* by now, all used Bing services are queried directly via REST APIs, so `dev.virtualearth.net` should not be required anymore
* `unsafe-eval` was added with 527ec293c2cd84e777e8f05b4bdcf2b3b611a5e0, but as far as I can see, neither the current mapillary SDK nor any other parts of iD perform any dirty tricks with `eval` & co.

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -19,7 +19,6 @@ class SiteController < ApplicationController
   content_security_policy(:only => :id) do |policy|
     policy.connect_src("*")
     policy.img_src("*", :blob)
-    policy.script_src(*policy.script_src, "dev.virtualearth.net", :unsafe_eval)
     policy.style_src(*policy.style_src, :unsafe_inline)
   end
 


tyrasd:
PS: I noticed that as of #4627 the content security policy has a `nonce` set up[^1] for the `style-src` attribute . This makes specifying `unsafe-inline` for iD essentially obsolete, as modern browsers will ignore it in favor of nonce'd inline styles. 

Inline styles were allowed in iD's CSP since c5d3335a6c5d4a1b74d3f9d509bbc6c7dba65fb2, but I don't think we still need them today[^2]. But maybe I'm overlooking something? :thinking: 

[^1]: See https://github.com/openstreetmap/openstreetmap-website/pull/4627/files#diff-c1c619ffb7b249550067cb696b8e7d6c29d1efe2ed4cf5b7a8bb6bed47b409d1R41
[^2]: Apart from some unnecessary inline-styles in with some of the bundled icons, which I just transformed into proper svg attributes with https://github.com/openstreetmap/iD/commit/3de2840ac3b5849bfdecd5f0b9b9a83be262e7e2

tomhughes:
Looks good to me, thanks.

tyrasd:
> but as far as I can see, neither the current mapillary SDK nor any other parts of iD perform any dirty tricks with eval & co.

Actually, this was wrong. Mapillary-js does still do this in one occasion:
https://github.com/mapillary/mapillary-js/blob/v4.1.2/src/graph/FilterCreator.ts#L45

#4856 fixes it

PR #4842, opened 2024-05-26
Truncate user menu username

AntonKhorev:
Follows #4828. If the username is long, the user menu button won't be oversized and won't trigger *small mode* for the header.

image


tomhughes:
Looks good to me, thanks.

PR #4843, opened 2024-05-27
Inline welcome sprite

AntonKhorev:
In #4820 I sorted all sprites into groups. This pull request is the next step. Each sprite is a trivial svg used only in one place, therefore it's possible to inline it. That makes the sprite map in css unnecessary.


tordans:
What would be your argument vs having separate files for them? 
Pro inline would be performance I guess. But not really with http2. 
Pro files would be: easier to maintain wen used twice. Easier to preview from the file system during development. The file name is a signal ob what this svg is. ...

tomhughes:
Looks good to me (though I can't find a way to do inline SVGs in emacs ;-)).

PR #4844, opened 2024-05-27
Replace png directions icon with inline svg

AntonKhorev:
Before/after:
image
image

2x zoom before/after:
image
image


tomhughes:
Looks good to me, thanks.

PR #4846, opened 2024-05-27
Add proper referrer for authorization scenario

milan-cvetkovic:
Fixes the following:
- `users_controller#new` loses referer in authorization scenario, when it was invoked after social signup succeded
- the second invocation of `auth_success`, triggered by re-authorization initiated from `users_controller#create` does not have referrer field set
- as a result, the final welcome screen does not offer final authorization, and drops into ID instead

Introduced by #4758.


tomhughes:
Oops, sorry about that, and thanks for the fix.

gravitystorm:
Ideally bug fixes like this - particularly for something that was broken due to a refactor - need a test added too. This will prevent the same bug from reappearing during any future refactoring.

PR #4847, opened 2024-05-27
Social sign-in: avoid re-authorization in `users_controller#create`

milan-cvetkovic:
It does not add any additional guards against malicious users:

Malicious user may attempt to invoke `POST /users/new` with bogus values for `auth_provider` and `auth_uid` resulting with a new account to which user would have a way to login, other than sending a password reset request.

In some cases, re-authorization would introduce additional "Please login to your social account", or "Are you sure you want to be logged in" popup triggered by identity provider.

This PR removes the re-authorization request from `POST /users/new` in authorization flow.

### DIFF
--- a/test/integration/user_creation_test.rb
+++ b/test/integration/user_creation_test.rb
@@ -45,47 +45,6 @@ class UserCreationTest < ActionDispatch::IntegrationTest
     assert_select "form > div > input.is-invalid#user_email"
   end
 
-  def test_user_create_association_bad_auth_provider
-    assert_difference("User.count", 0) do
-      assert_no_difference("ActionMailer::Base.deliveries.size") do
-        perform_enqueued_jobs do
-          post "/user/new",
-               :params => { :user => { :email => "test@example.com",
-                                       :display_name => "new_tester",
-                                       :pass_crypt => "testtest",
-                                       :pass_crypt_confirmation => "testtest",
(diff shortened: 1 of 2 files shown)


test/integration/user_creation_test.rb
```diff
@@ -363,7 +307,7 @@ def test_user_create_openid_failure
   end
 
   def test_user_create_openid_redirect
-    auth_uid = "http://localhost:1123/new.tester"
+    auth_uid = "12345654321"
```
tomhughes, line 310: I'm not sure why you're changing this to something that doesn't look like an OpenID id? You didn't change it in the `success` test?

milan-cvetkovic: The reason is to have different values for `auth_uid` and `openid_url`, to be able to verify exactly which one goes where bellow in `assert_redirected_to` and `post "/user/new"`.

I think I should probably update `success` test as well.

milan-cvetkovic: Updated `success` test case as well. Please let me know if it should be reverted to assume same values for `openid_url` and `auth_uid`

tomhughes: I would have stuck with URLs and just made them different from each other but it probably doesn't really matter.

tomhughes:
Thanks. That all looks good now and it certainly simplifies things a lot!

Issue #4130, opened 2023-08-02
Features should be enabled based on screen width, not menu width


gravitystorm:
When the screen is too narrow to display the 'compact-nav' list of menus (namely: logo+primary-nav+compact-secondary-nav+(user_menu or login_menu)), we switch to the 'small-nav' setup, with just the logo and the "burger" menu button.

However, this breakpoint depends on the calculated width of the menu items, not the width of the screen. It means that some features, like the edit button and the export button, are disabled based on the length of certain strings, rather than the screen width. Even in English, this means it depends on the length of the user's display name:

![Screenshot from 2023-08-02 11-43-24](https://github.com/openstreetmap/openstreetmap-website/assets/360803/0f73ef45-668f-4001-b4d3-c5ec805c2320)
![Screenshot from 2023-08-02 11-44-15](https://github.com/openstreetmap/openstreetmap-website/assets/360803/95b9a745-c585-4943-9a05-5733cedf444c)
![Screenshot from 2023-08-02 11-44-27](https://github.com/openstreetmap/openstreetmap-website/assets/360803/4446cd51-e72a-4793-aa9f-a3ac8d8e84b5)

I think it's reasonable to show the edit button, even in the 'small-nav' menu, for a screen width of over 1100 pixels!

Even without the silly name, the point still stands, particularly for languages where the four labels (edit, history, export, more) might be significantly different widths from in English. And it will also impact English, if we ever add or remove buttons from the primary menu.

We should decide what minimum width is appropriate for each tool, and hide (or disable) the buttons based on the screen width, instead of the menu width.

AntonKhorev:
The obvious choice for small-to-compact breakpoint is 768px. Sidebars switch at this breakpoint. However the header ui (logo + edit + history + export + more + login + signup) still doesn't fit into this size with some localizations. This doesn't depend on the username because it happens when logged out.

I think the Export button should be moved to the secondary nav where it can collapse into More. #351 already did something like this long ago. The only reason for Export to stay as a button is that it opens a map and other secondary nav items don't. But that's their current implementation, I can imagine GPS Traces opening a map.

Issue #3040, opened 2021-01-07
Move export tab to right hand tool pane


lectrician1:
The `Export` button to the right of the `Edit` and `History` buttons is rarely-used.

The majority usage population of the OSM website are mappers like me. 
I never have used the `Export` feature and probably never will because other file export options such as from the Overpass API and JOSM's import are much more convenient. 
Also, the pane does not even explain that it's exporting in the `.osm` format, offer to export in other formats say through a dropdown menu, and only has links to other sites.

It does not make sense to have it as a top-level heading link.

**I think it's pane contents should be either:**
- Moved to the `Share` button on the options pane right of the map and the hover text be renamed to `Share/Export`.
![ink](https://user-images.githubusercontent.com/22477298/103845057-e5f42f80-5060-11eb-9904-8dd977f1a566.png)
- Be added as a new dedicated Export button on the options pane right of the map.

**Here is what the site would look like without it:**
![image](https://user-images.githubusercontent.com/22477298/103846404-9ebb6e00-5063-11eb-91a8-e071f9a4df5a.png)

HolgerJeromin:
I also never use the export. But this option is one of the mayor difference to other mapping data sources.
You can download the raw data...

dieterdreist:
I find the export button a bit confusing, because the API is reserved for editing purposes. Maybe it should be renamed to "get the data" or something like this, with links to the planet and extracts, it is already there and shows when you select an area that is too big: "This area is too large to be exported as OpenStreetMap XML Data. Please zoom in or select a smaller area, or use one of the sources listed below for bulk data downloads."

lectrician1:
Just for clarification _I am not asking to remove the button_, just move it's view to another button that is less-prominent.

tomhughes:
Yes we're well aware that this is really just a trojan horse because you didn't think you'd be able to persuade us to add a Wiki tab unless you first removed something else - it would probably have helped your case here if you had been up front about that when opening this.

Personally I would be quite happy to remove the export button because I can't see that it gets any significant use because it's not really very useful. It's one of those things that seems reasonable on the surface but because it relies on an API that is only designed to deliver up small areas it's hard to see that it has much practical use, as the fact that we've had to add lots of hints on how to get larger areas to the panel!

lectrician1:
> you didn't think you'd be able to persuade us to add a Wiki tab unless you first removed something else

[...]

PR #4854, opened 2024-05-29
Move Export button to secondary nav

AntonKhorev:
*Export* is no longer a green outline button but rather a gray secondary navigation item.

The main reason for this is to allow *Export* to collapse into *More*. This lets the header to be narrower in *compact mode*. Hopefully then *compact mode* will be available above the Bootstrap medium breakpoint of 768px, which then can be made a fixed breakpoint for going between *small* and *compact* modes. (If not, maybe try moving the *History* button next.) See #4130.

I also don't hide the *Edit* button because it's easier not to hide it. If I kept it hidden exactly like it was before, the *History* button in the small mode menu would have had rounded borders only on one side. So coincidentally I also do here what #4839 tried to do except I don't need a vertical button group for edit-history-export. There's no *Export* in the group and *Edit*+*History* should fit horizontally.

image
image
image

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -412,6 +412,9 @@ $(document).ready(function () {
 
     if (OSM.router.route(this.pathname + this.search + this.hash)) {
       e.preventDefault();
+      if (this.pathname !== "/directions") {
+        $("header").addClass("closed");
+      }
     }
   });
 
--- a/app/assets/javascripts/index/search.js
+++ b/app/assets/javascripts/index/search.js
@@ -32,6 +32,7 @@ OSM.Search = function (map) {
 
   $(".describe_location").on("click", function (e) {
     e.preventDefault();
+    $("header").addClass("closed");
     var center = map.getCenter().wrap(),
         precision = OSM.zoomPrecision(map.getZoom()),
         lat = center.lat.toFixed(precision),
--- a/app/assets/javascripts/router.js
+++ b/app/assets/javascripts/router.js
@@ -101,6 +101,16 @@ OSM.Router = function (map, rts) {
 
   var router = {};
 
+  function updateSecondaryNav() {
+    $("header nav.secondary > ul > li > a").each(function () {
+      var active = $(this).attr("href") === window.location.pathname;
+
+      $(this)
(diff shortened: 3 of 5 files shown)


tordans:
This solution is a lot cleaner than my take at https://github.com/openstreetmap/openstreetmap-website/pull/4839

And in general I am all for re-evaluating the primary navigation, there is a lot of potential for some UX design and information architecture work here.

I am wondering if there is a way to keep this PR a bit closer to what we have right now:
In the desktop screen size, there is not issue with the export being part of the green button group.
Can we maybe keep the export part of the button group for the highest breakpoint. And only move it in the "more" navigation one breakpoint below?

On the dropdown menu screen size: Keeping the horizontal button group certainly is the cleaner solution given Bootstraps limitations. We will very likely need to revisit this once we add more mobile editors in the future. But we can look into that once we know the details.

AntonKhorev:
I wanted to get rid of *compact mode* and move the secondary menu items in/out of *More* depending on how much space is available. It was easier to do it together with the fixed breakpoint for small mode. See #4862.

image
image
image

This is how it looks if you get down to the 768px breakpoint:
image

Seems to work fine, but that's in English. Here's Bulgarian, and now things don't fit even with fully collapsed secondary menu items:
image

Without the green *Export* button everything seems to fit fine:
image

---

One of the alternatives to moving *Export* from buttons to the secondary menu is to move *Export* dynamically to *More* if there's not enough space. *Traces*, *Diaries* etc will move between the secondary menu and *More*, *Export* will move between buttons and *More*. That'll be a little more javascript.

But does *Export* deserve to be a button? Is it used often enough to justify this spot and then this moving behavior?

One advantage could be that if *Export* is moved to *More* and there's still not enough space, it's possible to similarly move *History*. But I don't know if there's a locale that requires this.

AntonKhorev:
#2420 does the opposite thing and adds one more button.

simonpoole:
There is not really a reason to keep the history button where it is: Assuming no revolutionary changes (aka multiple bounding box per changeset), the history function is not particularly useful and could easily be moved to create space for other things.

AntonKhorev:
#3040 proposes to move export to the map toolbar instead

PR #4855, opened 2024-05-29
Url-encode "?" characters in wikipedia tag values

### LINKED ISSUE #4851: Wikipedia links with question mark
### URL

https://www.openstreetmap.org/node/945504911

1. Go to this page in the browser:
https://www.openstreetmap.org/node/945504911

2. Click on Wikipedia link: en:Are Years What? (for Marianne Moore)
3. Get the error.

Correct link:
https://en.wikipedia.org/wiki/en:Are%20Years%20What%3F%20(for%20Marianne%20Moore)?uselang=en

Current link:
https://en.wikipedia.org/wiki/en:Are%20Years%20What?%20(for%20Marianne%20Moore)?uselang=en

The same bug was fixed in the iD editor earlier:
https://github.com/openstreetmap/iD/issues/10065

image
image

AntonKhorev:
Fixes #4851.

I only escape `?`s. Maybe there are also other characters that need escaping, but we don't escape everything that's possible in the existing test cases.

### DIFF
--- a/app/helpers/browse_tags_helper.rb
+++ b/app/helpers/browse_tags_helper.rb
@@ -98,8 +98,10 @@ module BrowseTagsHelper
       encoded_section = ""
     end
 
+    encoded_value = value.gsub("?") { |c| CGI.escape(c) }
+
     {
-      :url => "https://#{lang}.wikipedia.org/wiki/#{value}?uselang=#{I18n.locale}#{encoded_section}",
+      :url => "https://#{lang}.wikipedia.org/wiki/#{encoded_value}?uselang=#{I18n.locale}#{encoded_section}",
       :title => value + section
     }
   end
--- a/test/helpers/browse_tags_helper_test.rb
+++ b/test/helpers/browse_tags_helper_test.rb
@@ -205,6 +205,10 @@ class BrowseTagsHelperTest < ActionView::TestCase
     assert_equal "https://de.wikipedia.org/wiki/de:Liste der Baudenkmäler in Eichstätt?uselang=en#Br%C3%BCckenstra%C3%9Fe_1%2C_Ehemaliges_Bauernhaus", link[:url]
     assert_equal "de:Liste der Baudenkmäler in Eichstätt#Brückenstraße 1, Ehemaliges Bauernhaus", link[:title]
 
+    link = wikipedia_link("wikipedia", "en:Are Years What? (for Marianne Moore)")
+    assert_equal "https://en.wikipedia.org/wiki/en:Are Years What%3F (for Marianne Moore)?uselang=en", link[:url]
+    assert_equal "en:Are Years What? (for Marianne Moore)", link[:title]
+
     I18n.with_locale "pt-BR" do
       link = wikipedia_link("wikipedia", "zh-classical:Test#Section")
       assert_equal "https://zh-classical.wikipedia.org/wiki/zh-classical:Test?uselang=pt-BR#Section", link[:url]


gravitystorm:
I think this isn't quite the right approach, since as you say there are probably different characters that also need escaping (for example, a wikipedia link with a forward slash in the title, if that's possible).

I think we should look for something that encodes url components, rather than making our own function. I found this interesting article:

https://bibwild.wordpress.com/2023/02/14/escaping-encoding-uri-components-in-ruby-3-2/

I think it's reasonable for us to use the `ERB:Utils` approach, since we are not yet on ruby 3.2+

AntonKhorev:
> (for example, a wikipedia link with a forward slash in the title, if that's possible)

Forward slashes don't need to be escaped. `(`, `)`, `:` also don't need to be escaped.

> I think we should look for something that encodes url components

Are you also against replacing ` ` with `_`? That's something a standard url encoder is not going to do.

This is what you're going to get with `ERB:Utils`:
```
https://de.wikipedia.org/wiki/de:Englischer Garten (München)?uselang=en#Japanisches_Teehaus
https://de.wikipedia.org/wiki/de%3AEnglischer%20Garten%20%28M%C3%BCnchen%29?uselang=en#Japanisches_Teehaus
```

https://datatracker.ietf.org/doc/html/rfc3986#section-3.3

AntonKhorev:
https://url.spec.whatwg.org/#path-state - browsers probably do this. This should include not just URIs but also IRIs. You don't have to encode non-latin letters, browser will do it before making a http request.

AntonKhorev:
Actually what I said about forward slashes is not entirely true. What's true: 
- They are valid characters in paths.
- Wikipedia even internally doesn't escape them. Example: https://www.openstreetmap.org/node/2634817058 -> look at a [list of sculptures often with `/` in their names](https://en.wikipedia.org/wiki/List_of_sculptures_by_Tony_Smith). In the page source you'll see `<a href="/wiki/Moondog_(3/3)" title="Moondog (3/3)"><i>Moondog</i> (3/3)</a>`.
- But that wouldn't work if the name contains `/../`. Browsers will go up one path segment and remove part of the name. I don't know if such names are allowed on Wikipedia. Maybe they are not, then they don't have to escape `/`.

talllguy:
Wikipedia [has a section](https://en.wikipedia.org/wiki/Wikipedia:Naming_conventions_(technical_restrictions)#Question_marks_and_plus_signs) dedicated to the <kbd>?</kbd> and <kbd>+</kbd> in article titles and does recommend escaping them as the PR does. @AntonKhorev would it be worth adding the plus sign to the change?

### Plus Sign

Although the links led with [a + work (see +1 Records for example)](https://www.openstreetmap.org/node/11949307016), Wikipedia recomends encoding the URL like so https://en.wikipedia.org/wiki/%2B1_Records

### Forward slash

Per the namespace article:

> Page names consisting of exactly one or two periods (full stops), or beginning with ./ or ../, or containing /./ or /../, or ending with /. or /.., are not allowed.

Wikipedia _does_ escape them _only if_ the article begins with a <kbd>/</kbd>, e.g., https://en.wikipedia.org/wiki//dev/null

AntonKhorev:
> Wikipedia does escape them only if the article begins with a /, e.g., https://en.wikipedia.org/wiki//dev/null

Does it really?

image

If it did, they wouldn't need that rule about "Page names consisting of exactly one or two periods ..."

gravitystorm:
Closed in favour of #4887 (merged)

PR #4856, opened 2024-05-29
reintroduce unsafe-eval CSP rule for iD (Mapillary layer)

tyrasd:
Fixes a regression introduced by me in #4841. Fixes https://github.com/openstreetmap/iD/issues/10265


gravitystorm:
Merged, thanks!

tyrasd:
Thanks for the quick merge.  @gravitystorm would it be possible to roll this change out on osm.org, such as mappers can use mapillary from within iD again? :blush:

tomhughes:
@gravitystorm has no ability to roll anything out.

PR #4857, opened 2024-05-29
Fixed "or divider" issue described in #4773

nenad-vujicic:
This PR addresses "or divider" issue mentioned in the #4773

In app\views\sessions\new.html.erb and app\views\users\new.html.erb added mb-2 after "or" divider.

Before:
image
After:
image

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -25,7 +25,7 @@
 <div id="login_login" class="auth-container mx-auto my-0">
   <% if @preferred_auth_provider %>
     <%= render :partial => "auth_providers" %>
-    <div class="d-flex justify-content-center align-items-center">
+    <div class="d-flex justify-content-center align-items-center mb-2">
       <div class="border-bottom border-1 flex-grow-1"></div>
       <div class="text-secondary mx-3"><%= t ".or" %></div>
       <div class="border-bottom border-1 flex-grow-1"></div>
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -31,7 +31,7 @@
 
     <% unless @preferred_auth_provider.nil? %>
       <%= render :partial => "auth_providers" %>
-      <div class="d-flex justify-content-center align-items-center">
+      <div class="d-flex justify-content-center align-items-center mb-2">
         <div class="border-bottom border-1 flex-grow-1"></div>
         <div class="text-secondary mx-3"><%= t ".or" %></div>
         <div class="border-bottom border-1 flex-grow-1"></div>


app/views/sessions/new.html.erb
```diff
@@ -27,7 +27,7 @@
     <%= render :partial => "auth_providers" %>
     <div class="d-flex justify-content-center align-items-center">
       <div class="border-bottom border-1 flex-grow-1"></div>
-      <div class="text-secondary mx-3"><%= t ".or" %></div>
+      <div class="text-secondary mx-3 mb-2"><%= t ".or" %></div>
```
tomhughes, line 30: Was there a reason for choosing the `mb-2` rather than `mb-3` that @gravitystorm suggested?

tomhughes (changes requested):
This moves the line relative to the `or` text so the line is more or less aligned with the bottom of the text but I think it's better when the line is aligned with the middle of the text? I don't think the old version quite has them perfectly middle aligned but it's certainly a lot closer than this.

nenad-vujicic:
Dear @tomhughes ,

Regarding using mb-2 vs mb-3 .. When looking at distances between "Log in with Microsoft" and dividing line and dividing line and "Email Address and Username", they look more similar when using mb-2 than when using mb-3. Please, take a look at below screenshots (I can quickly switch to mb-3 if you prefer?):

mb-2:
image
mb-3:
image

Regarding problems with aligning lines with the middle of the "or" text .. I moved mb-2 to upper div, and this returned us to previous alignment. The problem with making this alignment perfect is browser's layout manager uses bounding box of uppercase of the "or" label, not the lowercase (doesn't respect kerning properties). I can fix it by setting a margin, but it will create a problem again if user selects a language with big glyphs (?locale=kn, ..) or switch to uppercase "OR", but it wouldn't look so nice again .. Please, let me know if you have any other idea?

Btw, I'm constantly getting for one- or two-lines changes failing checks (yesterday, Docker, today, Tests / Ubuntu ... Ruby 3.0). Do you have ideas why do they happen or how to mitigate them?

Thanks,
Nenad.

gravitystorm:
> Btw, I'm constantly getting for one- or two-lines changes failing checks (yesterday, Docker, today, Tests / Ubuntu ... Ruby 3.0). Do you have ideas why do they happen or how to mitigate them?

In this case it just had one job as "cancelled" and I don't know why that happened. I've re-run the failed jobs now (it's not clear to me how many people have permission to re-run failed tests).

Let us know if it happens in future. Usually jobs fail with a more obvious error message.

nenad-vujicic:
> > Btw, I'm constantly getting for one- or two-lines changes failing checks (yesterday, Docker, today, Tests / Ubuntu ... Ruby 3.0). Do you have ideas why do they happen or how to mitigate them?
> 
> In this case it just had one job as "cancelled" and I don't know why that happened. I've re-run the failed jobs now (it's not clear to me how many people have permission to re-run failed tests).
> 
> Let us know if it happens in future. Usually jobs fail with a more obvious error message.

Great, thanks! Now it passed all checks!

Thanks,
Nenad.

tomhughes (approved):
Thanks, that looks much better now. I realise exactly centring the text and line is hard - as it happens I was reading https://tonsky.me/blog/centering/ last night which has more detail than you could ever hope for...

PR #4860, opened 2024-05-29
Login form fixes

gravitystorm:
Followup for #4455 

This fixes two regressions in the login form:
 * re-enable translations for the password input label - thanks @Pikse for the report
 * removes autocomplete=off which is mostly ignored by browsers but shouldn't be used anyway.


tomhughes:
Looks good to me, thanks.

PR #4864, opened 2024-06-01
Don't include language prefix in wikipedia links

AntonKhorev:
In links like [https://de.wikipedia.org/wiki/de:Test?uselang=en](https://de.wikipedia.org/wiki/de:Test?uselang=en) what do you need `de:` for? `de` is already in the domain name `de.wikipedia.org`. iD and Overpass Turbo don't include this prefix.

If the prefix in the path differs from domain name, as was in one of the test cases [https://fr.wikipedia.org/wiki/de:Test?uselang=en](https://fr.wikipedia.org/wiki/de:Test?uselang=en), you get a redirect, in this case to [https://de.wikipedia.org/wiki/Test?uselang=en](https://de.wikipedia.org/wiki/Test?uselang=en). Just generate this url instead.

### DIFF
--- a/app/helpers/browse_tags_helper.rb
+++ b/app/helpers/browse_tags_helper.rb
@@ -69,38 +69,35 @@ module BrowseTagsHelper
 
     case key
     when "wikipedia", /^(#{SECONDARY_WIKI_PREFIXES}):wikipedia/o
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #4866, opened 2024-06-01
Larger font size in search form, svg icon for "Go" button

### LINKED ISSUE #1414: Search results are too wide on safari/iPhone
Steps to reproduce:

1. Open openstreetmap.org on iPhone iOS 10.2.
2. Click on the menu icon top right.
3. Search someting.
4. The result list ist too wide, so the x for closing the layer is not visible.

img_2271

AntonKhorev:
Should fix #1414 according to https://github.com/openstreetmap/openstreetmap-website/issues/1414#issuecomment-2130065798. I can't test it without an iPhone.

Changed small form controls to default ones. This makes font size to be 16px.

But there's another obstacle: the *Go* button. "Go" is short enough in English, but how about other languages. With a larger font it may get too large. To avoid that I replaced "Go" with a magnifying glass icon.

Before/after in English:
image
image

Before/after in Bulgarian (wider labels):
image
image

Before/after in Polish (already trying to replace "Go" with a symbol):
image
image

Before/after in Arabic:
image
image

200% zoom:
image

### DIFF
--- a/app/views/layouts/_search.html.erb
+++ b/app/views/layouts/_search.html.erb
@@ -3,12 +3,12 @@
     <div class="row gx-2 mx-0">
       <div class="col">
         <div class="input-group flex-nowrap">
-          <%= text_field_tag "query", params[:query], :placeholder => t("site.search.search"), :autofocus => autofocus, :autocomplete => "on", :class => "form-control form-control-sm z-0", :dir => "auto" %>
+          <%= text_field_tag "query", params[:query], :placeholder => t("site.search.search"), :autofocus => autofocus, :autocomplete => "on", :class => "form-control z-0 py-1 px-2", :dir => "auto" %>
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me although obviously the proof of whether it fixes #1414 will have to wait until it's deployed.

I guess we should be doing the same to the routing fields as well as it looks they are using the same text size that search was?

systemed:
This does indeed fix #1414 - massive thanks.

Yes, the issue remains present in the routing From/To fields.

deevroman:
It seems that in the final version the icon was not centered and stands out from the rest


Снимок экрана 2024-06-02 в 14 46 46

AntonKhorev:
> It seems that in the final version the icon was not centered

This is to get a smaller distance between the magnifying glass handle and the edge of the button in both ltr and rtl modes.

PR #4867, opened 2024-06-01
Write closed note controls using submit_tag

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/commit/c276625e7f35a1f9c31fe95325e150879f0b9a73 replaced html + interpolated attributes with `submit_tag`, but only for open note buttons. This PR does the same for closed note buttons.

image


tomhughes:
Looks good to me, thanks.

PR #4868, opened 2024-06-01
Inline svg sprites on about page

AntonKhorev:
Like  #4843 but for the *About* page.

Before/after on 200% zoom:
image image


tomhughes:
Looks good to me, thanks.

PR #4870, opened 2024-06-03
Increase font size in routing form

AntonKhorev:
Don't use `*-sm` Bootstrap controls to have 16px fonts and avoid auto-zoom on iPhone, see https://github.com/openstreetmap/openstreetmap-website/pull/4866#issuecomment-2143803410.

Before/after:
image image


tomhughes:
Looks good to me, thanks.

Issue #1353, opened 2016-10-31
The text does not fit in the button


zlant:
The text does not fit in the button
![image](https://cloud.githubusercontent.com/assets/8606466/19855201/541b1d80-9f84-11e6-8d11-f84bf26b6f16.png)

Solution:
`max-width: 180px;`
replace
`max-width: 100%;`
in openstreetmap-website/app/assets/stylesheets/common.scss

tomhughes:
Well except that still leaves the buttons over two lines presumably, when they are supposed to be together on one line I think?

Is there no way to make the text shorter?

tomhughes:
Incidentally you need to provide some context - what exactly is that sidebar? I assume it's a note? or maybe changeset history? Which line exactly in common.scss are you suggesting we change?

zetx16:
The forum discussed a long time shorter replacement. None came to nothing.

This is note. Line 2164.

d1g:
@tomhughes 
> Is there no way to make the text shorter?

We did several attempts on this. We have no better _and_ shorter words for this. "комментировать" is the least confusing translation AFAIK.

Shorter text labels "Обработать" are confusing (users open/close them repeatedly just to understand what button does)
Same about "Разрешить".

mmd-osm:
I would assume this was fixed in https://github.com/openstreetmap/openstreetmap-website/pull/4871, as mentioned there by @AntonKhorev

PR #4871, opened 2024-06-03
Use button elements that allow word wrap for note controls

AntonKhorev:
To make sure that #1353 is not a problem even on very narrow screens. `<button>` allows word wrap inside; if you need several words for example for "comment and close", they'll still fit on screen as long as every individual word fits.

Changesets already use `<button>` in discussion forms.

Before:
image

After:
image

### DIFF
--- a/app/assets/javascripts/index/note.js
+++ b/app/assets/javascripts/index/note.js
@@ -36,12 +36,12 @@ OSM.Note = function (map) {
   };
 
   function initialize(path, id, callback) {
-    content.find("input[type=submit]").on("click", function (e) {
+    content.find("button[type=submit]").on("click", function (e) {
       e.preventDefault();
       var data = $(e.target).data();
       var form = e.target.form;
 
-      $(form).find("input[type=submit]").prop("disabled", true);
+      $(form).find("button[type=submit]").prop("disabled", true);
 
       $.ajax({
         url: data.url,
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

tomhughes:
Oddly although this passes in CI it is causing test failures on my machine, for example:

```
Failure:
NoteCommentsTest#test_add_comment [test/system/note_comments_test.rb:44]:
expected to find text "Comment from User 59" in "Unresolved note #2707\nDescription\nThis is note comment 2\nCreated by anonymous less than 1 minute ago\nLocation: 1.0000000, 1.0000000\nThis note includes comments from anonymous users which should be independently verified.\nResolve\nComment\nIf this note contains sensitive information that needs to be removed, you can report this note. For all other problems with the note, please resolve it yourself with a comment."


bin/rails test test/system/note_comments_test.rb:25
```

tordans:
Looking into this a bit…

In https://github.com/openstreetmap/openstreetmap-website/pull/4871/files#diff-8f18485b763264500054435af432745e2a51b4402140728dc4df89024768b71bL48 we take the ajax type from the event. But it looks like we expect a specific type, so maybe we should hard code it here?

The other thing would be to make it explicit that the button is a submit button with "type=submit" which is what the Bootstrap docs use in their form example: https://getbootstrap.com/docs/5.3/forms/overview/#overview

However, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button?retiredLocale=de#type does not go into details on this. So maybe https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button?retiredLocale=de#formmethod is more what we need…

AntonKhorev:
@tomhughes

> Oddly although this passes in CI it is causing test failures on my machine

Did you rebuild the javascript?

> It's failing because changing to a button is making it do a GET not a POST

It's going to fail if you run old js with new html.

AntonKhorev:
> The other thing would be to make it explicit that the button is a submit button

It could be helpful to do the opposite thing, so they wouldn't submit anything if their click handler is not installed.

tomhughes:
I thought test mode would build assets on the fly but you're right that running a precompile fixes it.

PR #4872, opened 2024-06-03
Show changeset element counts using badges

AntonKhorev:
We use counter badges in other places. We can use them on changeset pages in element headings instead of numbers in ().

Before/after:
image image

### DIFF
--- a/app/views/changesets/_paging_nav.html.erb
+++ b/app/views/changesets/_paging_nav.html.erb
@@ -1,12 +1,18 @@
 <% if pages.page_count == 1 %>
   <h4 class="fs-5">
-    <%= t ".#{type.pluralize}", :count => pages.item_count %>
+    <%= t ".#{type.pluralize}_title" %>
+    <span class="badge count-number">
+      <%= pages.item_count %>
+    </span>
   </h4>
 <% elsif pages.page_count > 1 %>
   <h4 class="fs-5">
-    <%= t ".#{type.pluralize}_paginated", :x => pages.current_page.first_item,
-                                          :y => pages.current_page.last_item,
-                                          :count => pages.item_count %>
+    <%= t ".#{type.pluralize}_title" %>
+    <span class="badge count-number">
+      <%= t ".range", :x => pages.current_page.first_item,
+                      :y => pages.current_page.last_item,
+                      :count => pages.item_count %>
+    </span>
   </h4>
 
   <%= sidebar_classic_pagination(pages, "#{type}_page") do |page|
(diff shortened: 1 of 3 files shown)


nenad-vujicic (changes requested):
This looks and works great, thanks for it! Requires updating with the latest changes of _paging_nav.html.erb / en.yml.

AntonKhorev:
rebased

nenad-vujicic (approved):
It looks and works perfectly now, thank you very much!

config/locales/en.yml
```diff
@@ -523,12 +523,13 @@ en:
...
-      nodes: "Nodes (%{count})"
+      nodes_title: "Nodes"
       nodes_paginated: "Nodes (%{x}-%{y} of %{count})"
-      ways: "Ways (%{count})"
+      ways_title: "Ways"
       ways_paginated: "Ways (%{x}-%{y} of %{count})"
-      relations: "Relations (%{count})"
+      relations_title: "Relations"
       relations_paginated: "Relations (%{x}-%{y} of %{count})"
+      range: "%{x}-%{y} of %{count}"
```
hlfan, line 532: Why can't the `*_paginated` translations be removed?

app/views/changesets/_paging_nav.html.erb
```diff
@@ -1,8 +1,25 @@
...
+<% elsif pages.page_count > 1 %>
+  <h4 class="fs-5">
+    <%= t ".#{type.pluralize}_title" %>
+    <span class="badge count-number">
+      <%= t ".range", :x => pages.current_page.first_item,
+                      :y => pages.current_page.last_item,
+                      :count => pages.item_count %>
+    </span>
+  </h4>
+
```
hlfan, line 17: I'd find a duplicated if easier to read than a duplicated h4.
```suggestion
  <h4 class="fs-5">
    <%= t ".#{type.pluralize}_title" %>
    <span class="badge count-number">
      <%= if pages.page_count == 1
            pages.item_count
          elsif pages.page_count > 1
            t(".range", x: pages.current_page.first_item,
                        y: pages.current_page.last_item,
                        count: pages.item_count)
          end %>
    </span>
  </h4>
<% if pages.page_count > 1 %>
```

config/locales/en.yml
AntonKhorev: Search for `_paginated` to see where they are used.

app/views/changesets/_paging_nav.html.erb
AntonKhorev: This is reproducing what I removed from `browse_helper.rb`, except as a code block inside the template.

tomhughes:
Looks good to me, thanks.

PR #4873, opened 2024-06-03
Remove .btn-sm from changeset comment button

AntonKhorev:
Note comment buttons are not small:
image

But changeset comment buttons are. This makes them larger.

Before/after:
image image


tomhughes:
Looks good to me, thanks.

PR #4875, opened 2024-06-04
Add note and comment counts to the user profile

nenad-vujicic:
This PR addresses "Add note and comment counts too the user profile" issue mentioned in the #1643 

In app\views\users\show.html.erb added displaying notes and diary comments counts for both cases when displaying user's own profile page and when displaying user's profile page to the public. Also, added using built-in cached counts for displayed counts.

Before:
image

After:
image

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -14,7 +14,8 @@
               <span class='badge count-number'><%= number_with_delimiter(current_user.changesets.size) %></span>
             </li>
             <li>
-              <%= link_to t(".my notes"), user_notes_path(@user) %>
+              <%= link_to t(".my notes"), user_notes_path(current_user) %>
+              <span class='badge count-number'><%= number_with_delimiter(current_user.note_comments.size) %></span>
             </li>
             <li>
               <%= link_to t(".my traces"), :controller => "traces", :action => "mine" %>
@@ -26,6 +27,7 @@
             </li>
             <li>
               <%= link_to t(".my comments"), diary_comments_path(current_user) %>
+              <span class='badge count-number'><%= number_with_delimiter(current_user.diary_comments.size) %></span>
             </li>
             <li>
               <%= link_to t(".my settings"), edit_account_path %>
@@ -59,6 +61,7 @@
             </li>
             <li>
               <%= link_to t(".notes"), user_notes_path(@user) %>
+              <span class='badge count-number'><%= number_with_delimiter(@user.note_comments.size) %></span>
             </li>
             <li>
               <%= link_to t(".traces"), :controller => "traces", :action => "index", :display_name => @user.display_name %>
@@ -76,6 +79,7 @@
             </li>
             <li>
               <%= link_to t(".comments"), diary_comments_path(@user) %>
+              <span class='badge count-number'><%= number_with_delimiter(@user.diary_comments.size) %></span>
             </li>
             <li>
               <% if current_user and current_user.friends_with?(@user) %>
--- a/db/migrate/20240605134916_add_notes_and_diary_comments_counter_caches.rb
+++ b/db/migrate/20240605134916_add_notes_and_diary_comments_counter_caches.rb
@@ -1,0 +1,30 @@
+class AddNotesAndDiaryCommentsCounterCaches < ActiveRecord::Migration[7.1]
+  class DiaryComment < ApplicationRecord
+  end
+
+  class NoteComment < ApplicationRecord
+  end
+
+  class User < ApplicationRecord
+  end
+
+  def self.up
+    add_column :users, :diary_comments_count, :integer, :default => 0
+    add_column :users, :note_comments_count, :integer, :default => 0
+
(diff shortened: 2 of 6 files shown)


app/views/users/show.html.erb
```diff
@@ -15,6 +15,7 @@
             </li>
             <li>
               <%= link_to t(".my notes"), user_notes_path(@user) %>
+              <span class='badge count-number'><%= number_with_delimiter(current_user.notes.size) %></span>
```
tordans, line 18: This should be '@user', should it?

nenad-vujicic: Dear @tordans ,

Thank you for reviewing. In this case, I think they are same (2 proxies to the same object). Maybe I'm wrong :-) ?

Thanks,
Nenad.

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/1774#issuecomment-369141894.

AntonKhorev:
For note counts it's not obvious which notes to count, just those created by the user or all modified by the user. The *Notes* link shows the latter.

For comments count, again it's not obvious which comments these are. Currently the link shows only diary comments, but many also [want changeset comments](https://wiki.openstreetmap.org/wiki/Top_Ten_Tasks#List_of_changeset_discussions_for_a_user). My attempt at implementing it is #4248, which first requires #4587.

app/views/users/show.html.erb
gravitystorm: They are not generally the same. `current_user` is the currently logged in user, and `@user` is the user whose profile page you are showing.

Now in this specific situation, they are the same user, since they come after an equality check (`if current_user and @user.id == current_user.id`), so they can be used interchangeable - but elsewhere in this file they cannot.

I think the problem here is that although this block is mostly using `current_user` (e.g. for the edits count, 6 lines above), the line immediately above is using `@user`, and so seeing it switch between two lines suggests an error. My initial reaction is that we should use `current_user` throughout the `<!-- Displaying user's own profile page -->` section, for clarity.

nenad-vujicic:
Dear @AntonKhorev , @gravitystorm , @tordans ,

Thanks for reviewing and commenting. Do you agree to close this PR and try to solve these issues in #4248 and #4587?

Thanks,
Nenad.

tomhughes:
You need to run `bundle exec rails annotate_models` to update the model annotations for the schema change.

nenad-vujicic:
> You need to run `bundle exec rails annotate_models` to update the model annotations for the schema change.

Dear @tomhughes ,

Thank you very much for above. Now it passes all tests. Please, let me know if I can improve it somehow further.

Thanks,
Nenad.

db/migrate/20240605134916_add_notes_and_diary_comments_counter_caches.rb
```diff
@@ -0,0 +1,6 @@
+class AddNotesAndDiaryCommentsCounterCaches < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :diary_comments_count, :integer, :default => 0
+    add_column :users, :note_comments_count, :integer, :default => 0
```
tomhughes, line 4: These need to call `reset_counters` for any user that already has comments or the counts won't be correct - see `db/migrate/20121005195010_add_diary_entry_counter_caches.rb` for a way to to it reasonably efficiently.

There are less than 10k users with diary comments so that is probably fine but there are nearly half a million with not comments which may prove to be a problem in terms of time take to run the migration and populate the counters to we might have to do it with a rake task rather than in the migration.

app/views/users/show.html.erb
```diff
@@ -11,21 +11,23 @@
           <ul class='clearfix'>
             <li>
               <%= link_to t(".my edits"), :controller => "changesets", :action => "index", :display_name => current_user.display_name %>
-              <span class='badge count-number'><%= number_with_delimiter(current_user.changesets.size) %></span>
+              <span class='badge count-number'><%= number_with_delimiter(current_user.changesets_count) %></span>
```
tomhughes, line 14: Doesn't `changesets.size` already delegate to the counter? If it doesn't then this isn't working at the moment, which isn't good, and if it does why the change - what is the canonical rails way to do this?

tomhughes (changes requested):
I don't think we need to delay this - the counts should reflect what the links currently do and if we change what they do then we can consider how to deal with the counts then.

Other than the specific comments I have added please adjust the commit message as per my comments on #4881.

AntonKhorev:
> I don't think we need to delay this - the counts should reflect what the links currently do

https://github.com/openstreetmap/openstreetmap-website/pull/1774#discussion_r171697859

AntonKhorev:
We're almost at the state #1774 was left in.

nenad-vujicic:
Dear @tomhughes ,

Thank you very much for your review and comments. I returned to ".size" in app/views/users/show.html.erb and improved migration script (to update counters only for users with > 0 comments).

Please, let me know if I can improve it somehow further.

Thanks,
Nenad.

tomhughes (approved):
That looks good now, thanks.

Issue #1361, opened 2016-11-08
Diary posts should populate Open Graph tags for social sharing


bhousel:
It makes me sad when I try to share a diary post on a social site like Facebook or Twitter, and the site can't automatically fetch the title and highlight image because we don't set the necessary Open Graph tags.

Open Graph debug tool here:
https://developers.facebook.com/tools/debug/sharing/

For example, if I try to share this post:
https://www.openstreetmap.org/user/PlaneMad/diary/39846

I will end up with a generic share item with no image, and the description just talks about OpenStreetMap, not a description of the post that I am trying to share:

<img width="600" alt="screenshot 2016-11-08 09 35 28" src="https://cloud.githubusercontent.com/assets/38784/20103128/b7dfc646-a596-11e6-84cd-565d334a5dcd.png">

tomhughes:
Is this not a duplicate of #739?

tomhughes:
We seem to be passing the diary entry title, and I'm not sure how we could pass an image - even if we could find the images (which would be hard) how would we choose which one if there was more than one?

bhousel:
I'm considering #739 as sharing for map links and #1361 as sharing for diary links. 

> We seem to be passing the diary entry title, and I'm not sure how we could pass an image - even if we could find the images (which would be hard) how would we choose which one if there was more than one?

Maybe just the first image?  It's better than nothing.   

Yes, I agree this is a hard problem.  Honestly I don't really think we should have diaries in OpenStreetMap.  We should instead focus our limited developer attention on making the best map possible.   But if we _are_ going to embed a barebones blogging engine, we need to at least make the content socially shareable.

How would you feel about splitting the diary code out into a separate repository?

tomhughes:
I agree that we probably shouldn't have added the diaries in the first place, but it's hard to go back on that now...

I'm not quite sure how you thinking splitting them out would work/help? Do you mean we would run them as a separate service? That they would somehow be pulled back into the existing code? Bearing in mind we have outstanding code to pull them into a moderation system which would likely increase the level of integration...

simonpoole:
@bhousel while naturally every extra feature is a bit of a pain with our development model, the user diaries are for all practical purposes the only way an OSM contributor can reach a larger part of the community in a reasonably light weight way. 

[...]

PR #4876, opened 2024-06-04
Remove Open Graph og:image:secure_url property

AntonKhorev:
On the actual osm website `og:image:secure_url` is the same as `og:image`, therefore it's not [an alternate url](https://ogp.me/#structured).

~~~
<meta property="og:image" content="https://www.openstreetmap.org/assets/osm_logo_256-ed028f90468224a272961c380ecee0cfb73b8048b34f4b4b204b7f0d1097875d.png">
<meta property="og:image:secure_url" content="https://www.openstreetmap.org/assets/osm_logo_256-ed028f90468224a272961c380ecee0cfb73b8048b34f4b4b204b7f0d1097875d.png">
~~~

Same for OpenHistoricalMap and probably every publicly available openstreetmap-website instance.

Removing `og:image:secure_url` would make it easier to replace the image if necessary, like #1361.

### DIFF
--- a/app/helpers/open_graph_helper.rb
+++ b/app/helpers/open_graph_helper.rb
@@ -4,8 +4,7 @@ module OpenGraphHelper
       "og:site_name" => t("layouts.project_name.title"),
       "og:title" => [title, t("layouts.project_name.title")].compact.join(" | "),
       "og:type" => "website",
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #4877, opened 2024-06-04
Fix setting "From" in directions form based on search query

AntonKhorev:
When you type a search query and click directions button, the *From* field is supposed to be filled based on your query.

image
image

This has been broken since at least https://github.com/openstreetmap/openstreetmap-website/commit/f7b8b114a63f58015a8f0645cda6e0578cbe9f42 when the button css classes changed.

### DIFF
--- a/app/assets/javascripts/index/search.js
+++ b/app/assets/javascripts/index/search.js
@@ -9,9 +9,9 @@ OSM.Search = function (map) {
     }
   });
 
-  $(".search_form a.button.switch_link").on("click", function (e) {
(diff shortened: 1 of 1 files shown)


tordans:
We might want to start using explicit data attributes to connect JS to HTML. This kind of bugs are just way to easy with those deep nested selectors based on styling classes.

tomhughes:
Ha every time I do a search and then convert to routing I think that it used to copy the value over and then decide I must be misremembering but no, we just broke it. Thanks for the fix!

PR #4878, opened 2024-06-04
Remove javascripts.changesets locale strings

AntonKhorev:
They are used not in javascript but rather in .html.erb templates.


tomhughes:
Looks good to me, thanks.

Issue #3437, opened 2022-01-26
Add last edited date to user profile


boothym:
Mockup of the user profile page below, which would show the date a user last edited alongside the date they joined osm.org.

I feel this would be useful as it basically indicates when the user was last active, as sometimes you go to a profile to check their edits, before realising they haven't been active for a long time.

![lastedited](https://user-images.githubusercontent.com/16965772/151213826-63160540-1200-4862-9e62-401ecd1e929b.png)

mmd-osm:
Since this would be shown on a user profile page, "Last edited" might be somewhat misleading, as it could also refer to the last time the user profile page has been updated.

boothym:
I suppose "Last changeset" might be better then.

mmd-osm:
Changeset sounds a bit too technical for people unfamiliar with OSM. Maybe call it "Last map edit" or so.

matkoniecz:
> I feel this would be useful as it basically indicates when the user was last active, as sometimes you go to a profile to check their edits, before realising they haven't been active for a long time.

I do it very very often and such change would be really helpful.

PR #4881, opened 2024-06-06
Fixed "Add last edited date to user profile" issue described in #3437

nenad-vujicic:
This PR addresses "Add last edited date to user profile" issue mentioned in the #3437 

In app/views/users/show.html.erb added new definition for "Last map edit" as date of the user's last changeset modification or "No activity yet" label if user has not edited map yet (previously mentioned '-' could be misunderstood as website error). Also, added labels to translation files.

Before:
image
After:
image
image


tomhughes:
Thanks - the code all looks good so I'll merge this.

You might notice I edited the commit message - for future reference it's better to keep the subject line as a simple explanation of the change on not waste the limited space there on explaining the background which is better left for the body of the message. If you want to say it fixes a bug then a simple `Fixes #nnn` at the end of the body is enough and will also cause github to close the bug when it is merged.

PR #4883, opened 2024-06-07
Don't add site name to og:title when page title is provided

AntonKhorev:
It makes sense to add "OpenStreetMap" to `<title>` of every page, but for Open Graph there's already `og:site_name`. I looked at various sites like github, osm forum, they don't add site name to `og:title`.


tordans:
Agreed

tomhughes:
Looks good to me, thanks.

PR #4884, opened 2024-06-07
Add title to copyright page

AntonKhorev:
I'm moving the title string from `.legal_babble.title_html` to `.title` because it's not in that section and not html.

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -110,6 +110,7 @@ class SiteController < ApplicationController
   end
 
   def copyright
+    @title = t ".title"
     @locale = params[:copyright_locale] || I18n.locale
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/site_controller.rb
```diff
@@ -110,6 +110,7 @@ def edit
   end
 
   def copyright
+    @title = t ".title"
```
tomhughes, line 113: Should this use the page locale rather than the user's locale? In other words move it after `@locale` is set and add `, :locale => @locale` to it?

gravitystorm: Yeah, this could be decided either way. I'm happy with having the title in the same language as the rest of the chrome, so I'm happy to merge as-is.

gravitystorm:
Thanks @AntonKhorev

PR #4886, opened 2024-06-07
Don't set width of richtext preview

AntonKhorev:
Richtext preview div has its width set to the edit textarea, but why?

Make the browser window wide enough and start editing some text:
image

Click *Preview* and start shrinking the window. You may get something like this because the preview wouldn't shrink:
image

---

Preview and Help won't overlap if the width setting code is removed:
image


tomhughes:
I think I did it to ensure the preview was the same width as the editor (though setting `max-width` might have been better) but as the editor (now at least) has a width of 100% it doesn't really achieve anything now so this looks good to me.

Issue #967, opened 2015-05-13
Clicking wikipedia-link with a section does not always work 


daganzdaanda:
Hi, I have noticed something when looking at the tags of an object: 
Wikipedia links that point to sections (with # ) don't always work.

http://www.openstreetmap.org/way/130469169
has the tag
wikipedia=de:Englischer Garten (München)#Japanisches Teehaus

It's a clickable link to
`http://de.wikipedia.org/wiki/de:Englischer%20Garten%20(München)?uselang=de#Japanisches Teehaus`
This brings up the page, but does not jump to the section.

The section is reached by this link instead:
<http://de.wikipedia.org/wiki/Englischer_Garten_(München)#Japanisches_Teehaus>
Interestingly, both JOSM and OverpassTurbo translate the tag into the working link. 

It seems that it has to do with the spaces and also non-ascii characters: 
Simple one-word-anchors without "Umlauts" seem to be fine: <http://www.openstreetmap.org/node/1594364342>, but with an Umlaut, it does not work: <http://www.openstreetmap.org/node/1383144922>

Maybe loosely related is #712, that seems to be the last issue about wikipedia-links.

tomhughes:
Well those are two completely different links - you have removed the `uselang` parameter!

tomhughes:
Also, the link on the web site works find for me.

tomhughes:
So this is arguably a bug in mediawiki in that it apparently translates space to underscore in the URL but not in the anchor.

tomhughes:
To summarise, and ignoring all the irrelevant confusion about umlauts, and the various accidental modifications of unrelated bits of the URL in the initial report, the problem is that we generate this link:

```
http://de.wikipedia.org/wiki/de:Englischer%20Garten%20(M%C3%BCnchen)?uselang=en-GB#Japanisches%20Teehaus
```

when we need to generate:

```
http://de.wikipedia.org/wiki/de:Englischer%20Garten%20(M%C3%BCnchen)?uselang=en-GB#Japanisches_Teehaus
```

with an underscore in the anchor component, because mediawiki requires spaces to be translated to underscore for the anchor name when multiple words are used in a section heading.

The question is, is this really a bug in the web site core? or is the data in the database wrong? It seems that `de:Englischer Garten (München)#Japanisches Teehaus` is not really the right reference and it should be `de:Englischer Garten (München)#Japanisches_Teehaus`.

daganzdaanda:
[...]

PR #4887, opened 2024-06-08
Fully encode wikipedia tag values

AntonKhorev:
Fix https://github.com/openstreetmap/openstreetmap-website/issues/4851 by using the opposite approach to #4855 and what's currently deployed: urlencode everything unless it needs to be unencoded. Things that are not urlencoded:
- `:` in language prefix
- `#` before article section name

Urls will get uglier, especially non-English ones.

### DIFF
--- a/app/helpers/browse_tags_helper.rb
+++ b/app/helpers/browse_tags_helper.rb
@@ -85,20 +85,15 @@ module BrowseTagsHelper
       title_section = value
     end
 
-    if title_section =~ /^([^#]*)#(.*)/
-      # Contains a reference to a section of the wikipedia article
-      # Must break it up to correctly build the url
-      title = Regexp.last_match(1)
-      encoded_section = "##{CGI.escape(Regexp.last_match(2).gsub(/ +/, '_'))}"
-    else
-      title = title_section
-      encoded_section = ""
-    end
+    title, section = title_section.split("#", 2)
+    url = "https://#{lang}.wikipedia.org/wiki/#{wiki_encode(title)}?uselang=#{I18n.locale}"
+    url += "##{wiki_encode(section)}" if section
 
-    {
-      :url => "https://#{lang}.wikipedia.org/wiki/#{title}?uselang=#{I18n.locale}#{encoded_section}",
-      :title => value
-    }
+    { :url => url, :title => value }
+  end
+
+  def wiki_encode(s)
+    u s.tr(" ", "_")
   end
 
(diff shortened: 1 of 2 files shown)


AntonKhorev:
> What I would prefer, is rather than u s.tr(" ", "_") is to avoid using the (mediawiki-specific) underscores

How are you going to avoid underscores in mediawiki-specific ids?

> url_encode(section)

https://en.wikipedia.org/wiki/London#Leisure_and_entertainment

vs

https://en.wikipedia.org/wiki/London#Leisure%20and%20entertainment

gravitystorm:
Sorry, perhaps I wasn't clear. My suggestion is to avoid converting spaces to underscores, but of course any underscores already used in the tag would be preserved.

```
irb(main):001> require 'erb'
=> true
irb(main):002> ERB::Util.url_encode('Leisure_and_entertainment')
=> "Leisure_and_entertainment"
```

So if a mapper added spaces in the ids, that's their mistake, because there's no spaces in the ids. If they used underscores, those would be preserved.

What I think we should avoid is *converting* spaces into underscores, because it's unnecessary.

AntonKhorev:
> So if a mapper added spaces in the ids, that's their mistake

Not according to the tests:

https://github.com/openstreetmap/openstreetmap-website/blame/ebaae09797d245ad05f1c00c2f8359c8868fe14d/test/helpers/browse_tags_helper_test.rb#L200

Spaces after `#` seem to be more prevalent than underscores in osm data.

gravitystorm:
> Spaces after `#` seem to be more prevalent than underscores in osm data.

Ah, interesting, I wasn't aware that we were doing that for the fragments deliberately. Thanks for pointing that out.

It feels to me like it's the wrong thing to do, since the spaces are completely invalid in mediawiki fragments, and therefore they are invalid data when used in an OSM tag too. Tracing it back, it seems like in [May 2015](https://github.com/openstreetmap/openstreetmap-website/issues/967) other data consumers were already silently fixing the broken fragment data, and so the same automatic fixup was also applied to this codebase too. Unfortunately it means that every other consumer of OSM data will need to do the same too! Oh well.

So we would need to keep the `tr` for the fragments, but for the page name itself, it's still unnecessary. But overall I still stand by my earlier comment that's it's not the most important thing for us to deal with.

AntonKhorev:
I guess many osm contributors don't try to enter a valid fragment after `#`. They are entering a section title from which the fragment is generated.

PR #4889, opened 2024-06-10
Move diary comments index action to comments controller

AntonKhorev:
There was a diary comment model but no controller. Everything related to comments was in the diary entries controller.

Here I'm moving `diary_entries#comments` to `diary_comments#index`. There are at least three other comment actions, I'll move them later.


tomhughes:
Looks good to me, thanks.

PR #4891, opened 2024-06-10
Parse og:image urls using Addressable::URI

### LINKED ISSUE #4890: HTTP 500 when trying to open a diary entry via a link
### URL

https://www.openstreetmap.org/user/SomeoneElse/diary/404191

1. Open link

Снимок экрана 2024-06-10 в 16 16 17

p.s. the entry is visible in the list of user entries https://www.openstreetmap.org/user/SomeoneElse/diary

AntonKhorev:
Fixes #4890.

I used an URI parser to construct absolute URIs, but that thing is stricter than browsers.

<s>Now I only parse urls that don't start with http[s]: and catch invalid URI exceptions.</s>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,6 +72,7 @@ gem "validates_email_format_of", ">= 1.5.1"
 gem "quad_tile", "~> 1.0.1"
 
 # Sanitise URIs
+gem "addressable", "~> 2.8"
 gem "rack-uri_sanitizer"
 
 # Omniauth for authentication
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -602,6 +602,7 @@ DEPENDENCIES
   actionpack-page_caching (>= 1.2.0)
   active_record_union
   activerecord-import
+  addressable (~> 2.8)
   annotate
   argon2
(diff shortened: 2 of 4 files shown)


tomhughes:
Is there actually a valid reason to exclude http URLs though?

As far as I can see the URL with the spaces in was invalid and the fact that browsers choose to accept it anyway doesn't change that - by excluding http URLs from validation all sorts of other things which browsers wouldn't accept can also get through.

AntonKhorev:
There is a possibility that open graph consumers won't like urls with spaces. Open graph docs don't specify what standard urls should conform to. Ideally I'd parse all urls but I couldn't find a library that parses them like a browser.

AntonKhorev:
Better fix:
- join urls with Addressable::URI - this one doesn't fail on urls with spaces
- normalize urls after that - replaces spaces with %20 in case open graph consumers wouldn't like them
- catch Addressable::URI::InvalidURIError and return the default image

tomhughes:
The `og_image_url` method is a bit ugly but I haven't come up with a better way so I guess this is good to go.

PR #4892, opened 2024-06-12
Add missing graphhopper icons

AntonKhorev:
GraphHopper's routing icon map is missing some entries that are [possible in the results](https://docs.graphhopper.com/#operation/getRoute!c=200&path=paths/instructions/sign&t=response). They are represented as a straight arrow because this icon is at offset zero in the routing sprite svg.
image

[Keep left/right](https://www.openstreetmap.org/directions?engine=graphhopper_car&route=53.2390%2C14.0022%3B53.3071%2C13.9935) before/after:

image image

[U-turn](https://www.openstreetmap.org/directions?engine=graphhopper_car&route=59.92761%2C30.29539%3B59.92773%2C30.29552) before/after:

image image


tomhughes:
Looks good to me, thanks.

PR #4894, opened 2024-06-14
Added indicator message preview is empty.

### LINKED ISSUE #3748: Indicate if the message preview is empty
The message preview should indicate to the user it is empty, similar to how GitHub says "Nothing to preview". It should either
- Display such text on the preview; or
- grey out the preview button. (may have hover text or similar to indicate the reason)

**Current behaviour** (large empty area)
image

nenad-vujicic:
This PR addresses "Indicate if the message preview is empty" issue mentioned in the #3748

Added check in Preview button click event handler if body (message, diary entry, diary comment) is empty and if positive, performs reporting validity (displays tooltip). Fixes #3748.

image

### DIFF
--- a/app/assets/javascripts/richtext.js
+++ b/app/assets/javascripts/richtext.js
@@ -43,6 +43,12 @@ $(document).ready(function () {
     var preview = $(this).parents(".richtext_container").find(".richtext_preview");
     var minHeight = editor.outerHeight() - preview.outerHeight() + preview.height();
 
+    if (editor.val().length === 0) {
+      event.preventDefault();
+      editor[0].reportValidity();
+      return;
+    }
+
     if (preview.contents().length === 0) {
       preview.oneTime(500, "loading", function () {
         preview.addClass("loading");


AntonKhorev:
Github's "Nothing to preview" is not ideal because it looks exactly as if you wrote a message "Nothing to preview".

tomhughes:
Maybe we should just disable the preview button when there is nothing to preview?

AntonKhorev:
I was thinking about [alert](https://getbootstrap.com/docs/5.3/components/alerts/) to make "Nothing to preview" look different. With a disabled button users have to figure out why it's disabled.

AntonKhorev:
I have a branch with tabs by the way:

image

image

nenad-vujicic:
Dear @AntonKhorev , @tomhughes ,

I updated the sources to trigger validation and error reporting like when pushing Send / Publish (like on above screenshot). Please, let me know if it looks fine now or if I can improve it somehow.

Thanks,
Nenad.

AntonKhorev:
Is it necessary to have an "Add subject" placeholder right under a "Subject" label? Maybe you can argue that the "Body" label is not clear enough, but then why is the placeholder "Add your message" even for diary entries? And what does all of this have to do with the issue #3748?

AntonKhorev:
> I updated the sources to trigger validation

It also validates the subject and will tell to fill out the subject if its empty while not saying anything about the body. Does that make sense when the user asks for a preview of the body?

app/assets/javascripts/richtext.js
```diff
@@ -34,15 +34,54 @@ $(document).ready(function () {
...
+   * diary_entry_title (Subject) element, .. when publish / update
+   * button is pressed.
+   */
+  $(".btn-primary[id=\"publish_or_update_diary_entry_button_id\"]").click(function () {
+    var diaryEntryTitleElement = document.getElementById("diary_entry_title");
+    if (diaryEntryTitleElement) {
+      diaryEntryTitleElement.setAttribute("required", "required");
+    }
+  });
+
```
AntonKhorev, line 59: None of this is required for what you're trying to do.

app/assets/javascripts/richtext.js
```diff
@@ -34,15 +34,54 @@ $(document).ready(function () {
...
+      }
+
+      var diaryEntryTitleElement = document.getElementById("diary_entry_title");
+      if (diaryEntryTitleElement) {
+        diaryEntryTitleElement.removeAttribute("required");
+      }
+
+      return;
+    }
+
```
AntonKhorev, line 84: I don't necessarily agree that triggering the validation is a good idea, but if you really want to do it, you can just call `reportValidity` on an input:

```suggestion
    event.preventDefault();

    if (!editor[0].reportValidity()) {
      return;
    }
```

nenad-vujicic:
Dear @AntonKhorev ,

Thank you very much for your review & suggestion. When I was trying reportValidity(), I forgot to put preventDefault() before :-).

I used validation because checking Subject / Body when pushing Send / Publish / Update used validation. What is the ideal solution here, using validation or bootstrap alerts?

Thanks,
Nenad.

nenad-vujicic:
Added alternative (#5023) using bootstrap alerts

AntonKhorev:
Closed in favor of #5023

PR #4895, opened 2024-06-14
Added link to nominatim results in searching results

### LINKED ISSUE #3205: Link to nominatim including search query
Search results include "Results from OpenStreetMap Nominatim" which links to https://nominatim.openstreetmap.org/. To help mappers debug search results to determine if unexpected search results are OSM data issues or not, it can be handy to inspect the full search results output available at nominatim.openstreetmap.org.

To make this easier for mappers we should directly link including the search query, via `https://nominatim.openstreetmap.org/ui/search.html?q=`.

Either this URL could replace the "Results..." link or we could add another link "View search on Nominatim" at the bottom which includes the full query included link.

nenad-vujicic:
This PR addresses "Link to nominatim including search query" issue mentioned in the #3205 

Fixes #3205. Added caching of nominatim URL query parameters in GeocoderController#search for both direct and reverse geocoding. In app/views/geocoder/search.html.erb added displaying cached URL as forwarding link when clicked on "OpenStreetMap Nominatim" label.

image

### DIFF
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -13,10 +13,10 @@ class GeocoderController < ApplicationController
     @sources = []
 
     if @params[:lat] && @params[:lon]
-      @sources.push "latlon"
-      @sources.push "osm_nominatim_reverse"
+      @sources.push({ :name => "latlon", :parameters => "" })
+      @sources.push({ :name => "osm_nominatim_reverse", :parameters => "reverse?format=html&#{nominatim_reverse_url_parameters}" })
     elsif @params[:query]
-      @sources.push "osm_nominatim"
+      @sources.push({ :name => "osm_nominatim", :parameters => "search?format=html&#{nominatim_url_parameters}" })
     end
 
     if @sources.empty?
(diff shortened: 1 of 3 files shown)


app/controllers/geocoder_controller.rb
```diff
@@ -15,8 +15,32 @@ def search
...
       @sources.push "latlon"
       @sources.push "osm_nominatim_reverse"
+
+      # get query parameters
+      lat = params[:lat]
+      lon = params[:lon]
+      zoom = params[:zoom]
+
+      # create URL for nominatim reverse search
+      @cached_osm_nominatim_parameters = "reverse?format=html&lat=#{lat}&lon=#{lon}&zoom=#{zoom}&accept-language=#{http_accept_language.user_preferred_languages.join(',')}"
```
tomhughes, line 25: This, and the equivalent code for forward searches, is just duplicating the work that the search methods do.

It would be better to move that code into private `nominatim_reverse_url` and `nominatim_url` methods that can then be used in both places.

app/views/geocoder/search.html.erb
```diff
@@ -3,10 +3,17 @@
 <%= render "sidebar_header", :title => t("site.sidebar.search_results") %>
 
 <% @sources.each do |source| %>
-  <h4>
-    <%= t(".title.results_from_html", :results_link => link_to(t(".title.#{source}"),
-                                                               t(".title.#{source}_url"))) %>
-  </h4>
+  <% if source.include?("osm_nominatim") %>
```
tomhughes, line 6: Rather than having this look at the name of the source it would be better to make `@sources` an array of hashes, each with a `:name` and an optional `:url` key.

tomhughes (changes requested):

nenad-vujicic:
Dear @tomhughes , @gravitystorm ,

Can you please, re-start checks on the latest commit of this PR? I'm getting some strange error, while all tests pass on my side locally.

Thanks,
Nenad.

app/views/geocoder/search.html.erb
nenad-vujicic: Dear @tomhughes ,

Thank you very much for your review and comments. I've just made above updates. Please, let me know if you have idea how we could improve it further.

Thanks,
Nenad.

tomhughes (approved):
Thanks. I was actually thinking of the whole URL rather than just the parameters in the hash but I can see why you didn't do that so I think this is good to go now.

tomhughes:
I've opened #4922 as a followup with some additional cleanup.

PR #4896, opened 2024-06-14
Remove global "awaiting geocode" routing state

AntonKhorev:
`/directions` javascript has an `awaitingGeocode` state variable, which is true when one of endpoints awaits geocoding results. But each endpoint has its own `awaitingGeocode` state variable too. Endpoint methods manipulate both global and their own state. Why do they need the global state? Only to decide if they need to run `getRoute()`. But `getRoute()` checks `awaitingGeocode` on endpoints anyway. There's no need for the global state.


tomhughes:
Looks good to me, thanks.

PR #4897, opened 2024-06-14
Use .is-invalid to highlight routing inputs with failed geocode lookups

AntonKhorev:
Currently `.error` is added to these inputs, but it doesn't make them look any different.

image


tomhughes:
Looks good to me, thanks.

PR #4900, opened 2024-06-15
Write directions sidebar header using .erb template

AntonKhorev:
Other sidebar pages have at least a heading and some placeholder in their `.html.erb` template, for example [geocoder search](https://github.com/openstreetmap/openstreetmap-website/blob/84aa7f455ae58b4b051368e35dbc43f780b3aa0f/app/views/geocoder/search.html.erb). Directions search is the exception, it's template is [essentially empty](https://github.com/openstreetmap/openstreetmap-website/blob/84aa7f455ae58b4b051368e35dbc43f780b3aa0f/app/views/directions/search.html.erb). Instead everything is written by `directions.js`.

Here I'm moving the directions heading to `.html.erb`. Actually I'm doing this to add more things to `.html.erb` later. Preview: #4901.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -133,7 +133,7 @@ OSM.Directions = function (map) {
     // copy loading item to sidebar and display it. we copy it, rather than
     // just using it in-place and replacing it in case it has to be used
     // again.
-    $("#sidebar_content").html($(".directions_form .loader_copy").html());
+    $("#directions_content").html($(".directions_form .loader_copy").html());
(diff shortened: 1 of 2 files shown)


AntonKhorev:
I have `endpoint.remove()` in #5064. I can rename it to `disable`, add `enable`, enable the listeners from enable, remove them in `disable`.

AntonKhorev:
I did the thing above but there's still blinking caused by `OSM.loadSidebarContent` messing with `map.setSidebarOverlaid`.

AntonKhorev:
Superseded by #5754

PR #4902, opened 2024-06-15
Use callbacks instead of direct getRoute() calls from route endpoints

AntonKhorev:
Directions page javascript controller has two route endpoint objects. They manage from/to inputs and markers. When inputs and markers are updated, endpoint objects call `getRoute` method of the controller to redraw the route. But why should endpoints know anything about the entire route? They only need to report that an input or a marker was updated and let the outside code decide what to do in response.


tomhughes:
Looks good to me, thanks.

PR #4904, opened 2024-06-17
Fixed displacing points specified from context menu

### LINKED ISSUE #2982: "Directions from here" often places the point hundreds of meters elsewhere
I can always move the red and green makers to any arbitrary point anywhere I want along the road with the mouse, and it works great.

I just don't understand why the right click menu's "Directions from here" and "Directions to here" often place the point hundreds of meters elsewhere, whereupon I simply have to grab it and put it back where I intended it to be.

nenad-vujicic:
This PR addresses "Directions from here often places the point hundreds of meters elsewhere" issue mentioned in #2982 

Fixes #2982 by updating directions.js so it doesn't perform unnecessary geocoding anymore for previously selected point (either source or destination), when entering second point. Also, improved setting markers by positioning it to selected position (not to result of geocoding).

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -5,6 +5,7 @@
 OSM.Directions = function (map) {
   var awaitingRoute; // true if we've asked the engine for a route and are waiting to hear back
   var chosenEngine;
+  var coordinatesRegularExpression = "([+-]?\\d+(\\.\\d*)?)(?:\\s+|\\s*[/,]\\s*)([+-]?\\d+(\\.\\d*)?)";
 
   var popup = L.popup({ autoPanPadding: [100, 100] });
 
@@ -20,18 +21,13 @@ OSM.Directions = function (map) {
     weight: 12
   });
 
-  var endpointDragCallback = function (dragging) {
-    if (map.hasLayer(polyline)) {
-      getRoute(false, !dragging);
-    }
-  };
-  var endpointGeocodeCallback = function () {
-    getRoute(true, true);
+  var endpointUpdatedCallback = function (fitBounds) {
+    getRoute(fitBounds, true);
   };
 
   var endpoints = [
-    Endpoint($("input[name='route_from']"), OSM.MARKER_GREEN, endpointDragCallback, endpointGeocodeCallback),
-    Endpoint($("input[name='route_to']"), OSM.MARKER_RED, endpointDragCallback, endpointGeocodeCallback)
+    Endpoint($("input[name='route_from']"), OSM.MARKER_GREEN, endpointUpdatedCallback),
+    Endpoint($("input[name='route_to']"), OSM.MARKER_RED, endpointUpdatedCallback)
   ];
 
   var expiry = new Date();
@@ -51,7 +47,7 @@ OSM.Directions = function (map) {
     select.append("<option value='" + i + "'>" + I18n.t("javascripts.directions.engines." + engine.id) + "</option>");
   });
 
-  function Endpoint(input, iconUrl, dragCallback, geocodeCallback) {
+  function Endpoint(input, iconUrl, endpointCoordinatesUpdatedCallback) {
     var endpoint = {};
 
     endpoint.marker = L.marker([0, 0], {
@@ -71,8 +67,8 @@ OSM.Directions = function (map) {
       var dragging = (e.type === "drag");
       if (dragging && !chosenEngine.draggable) return;
       if (dragging && awaitingRoute) return;
-      endpoint.setLatLng(e.target.getLatLng());
-      dragCallback(dragging);
+
+      endpoint.setLatLng(e.target.getLatLng(), true, !dragging);
     });
 
     input.on("keydown", function () {
@@ -82,17 +78,31 @@ OSM.Directions = function (map) {
     input.on("change", function (e) {
       // make text the same in both text boxes
       var value = e.target.value;
-      endpoint.setValue(value);
+      endpoint.setValue(value, false, null, true);
     });
 
-    endpoint.setValue = function (value, latlng) {
+    endpoint.setValue = function (value, force_reverse_geocoding, latlng, fitMap) {
+      if (!value && !latlng) {
+        endpoint.value = value;
+        delete endpoint.latlng;
+        input.val(value);
+
+        return;
+      }
+
+      if (value === endpoint.value) {
+        endpoint.setLatLng(endpoint.latlng, false, fitMap);
+        return;
+      }
+
       endpoint.value = value;
       delete endpoint.latlng;
-      input.removeClass("is-invalid");
       input.val(value);
 
       if (latlng) {
-        endpoint.setLatLng(latlng);
+        endpoint.setLatLng(latlng, true, fitMap);
+      } else if (!force_reverse_geocoding && endpoint.value.match(new RegExp("^" + coordinatesRegularExpression + "$"))) {
+        endpoint.setLatLng(L.latLng(endpoint.value.split(/[/,]/)), false, fitMap);
       } else {
         endpoint.getGeocode();
       }
@@ -109,31 +119,64 @@ OSM.Directions = function (map) {
 
       var viewbox = map.getBounds().toBBoxString(); // <sw lon>,<sw lat>,<ne lon>,<ne lat>
 
-      $.getJSON(OSM.NOMINATIM_URL + "search?q=" + encodeURIComponent(endpoint.value) + "&format=json&viewbox=" + viewbox, function (json) {
-        endpoint.awaitingGeocode = false;
-        endpoint.hasGeocode = true;
-        if (json.length === 0) {
-          input.addClass("is-invalid");
-          alert(I18n.t("javascripts.directions.errors.no_place", { place: endpoint.value }));
-          return;
-        }
+      if (endpoint.value.match(new RegExp("^" + coordinatesRegularExpression + "$"))) {
+        var latlng_array = endpoint.value.split(/[/,]/);
+        $.getJSON(OSM.NOMINATIM_URL + "reverse?lat=" + latlng_array[0] + "&lon=" + latlng_array[1] + "&format=json&viewbox=" + viewbox, function (json) {
+          endpoint.awaitingGeocode = false;
+          endpoint.hasGeocode = true;
 
-        endpoint.setLatLng(L.latLng(json[0]));
+          endpoint.setLatLng(L.latLng(latlng_array), false, true);
 
-        input.val(json[0].display_name);
+          if (!json || !json.display_name) {
+            input.addClass("is-invalid");
+            alert(I18n.t("javascripts.directions.errors.no_place", { place: endpoint.value }));
+            return;
+          }
 
-        geocodeCallback();
-      });
+          if (endpoint.latlng.distanceTo(L.latLng(json.lat, json.lon)) > 200.0) {
+            input.val(endpoint.value);
+          } else {
+            endpoint.value = json.display_name;
+            input.val(json.display_name);
+          }
+        });
+      } else {
+        $.getJSON(OSM.NOMINATIM_URL + "search?q=" + encodeURIComponent(endpoint.value) + "&format=json&viewbox=" + viewbox, function (json) {
+          endpoint.awaitingGeocode = false;
+          endpoint.hasGeocode = true;
+          if (!json || json.length === 0) {
+            input.addClass("is-invalid");
+            alert(I18n.t("javascripts.directions.errors.no_place", { place: endpoint.value }));
+            return;
+          }
+
+          endpoint.setLatLng(L.latLng(json[0]), false, true);
+
+          endpoint.value = json[0].display_name;
+          input.val(json[0].display_name);
+        });
+      }
     };
 
-    endpoint.setLatLng = function (ll) {
-      var precision = OSM.zoomPrecision(map.getZoom());
-      input.val(ll.lat.toFixed(precision) + ", " + ll.lng.toFixed(precision));
+    endpoint.setLatLng = function (ll, populate_input, fitMap) {
+      if (!ll) {
+        return;
+      }
+
+      input.removeClass("is-invalid");
+
+      if (populate_input) {
+        var precision = OSM.zoomPrecision(map.getZoom());
+        endpoint.value = ll.lat.toFixed(precision) + ", " + ll.lng.toFixed(precision);
+        input.val(endpoint.value);
+      }
       endpoint.hasGeocode = true;
       endpoint.latlng = ll;
       endpoint.marker
         .setLatLng(ll)
         .addTo(map);
+
+      endpointCoordinatesUpdatedCallback(fitMap);
     };
 
     return endpoint;
@@ -200,7 +243,7 @@ OSM.Directions = function (map) {
     select.val(index);
   }
 
-  function getRoute(fitRoute, reportErrors) {
+  function getRoute(fitBounds, reportErrors) {
     // Cancel any route that is already in progress
     if (awaitingRoute) awaitingRoute.abort();
 
@@ -246,6 +289,14 @@ OSM.Directions = function (map) {
           $("#sidebar_content").html("<div class=\"alert alert-danger\">" + I18n.t("javascripts.directions.errors.no_route") + "</div>");
         }
 
+        if (fitBounds) {
+          var partiallyCombinedBounds = L.latLngBounds();
+          partiallyCombinedBounds.extend(L.latLngBounds(o, o));
+          partiallyCombinedBounds.extend(L.latLngBounds(d, d));
+
+          map.fitBounds(partiallyCombinedBounds.pad(0.05));
+        }
+
         return;
       }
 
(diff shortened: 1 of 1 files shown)


nenad-vujicic:
Dear @tomhughes ,

Thank you for your review and comment. I updated PR and commit messages. Please, let me know if I can improve it further.

Thanks,
Nenad.

AntonKhorev:
> so it doesn't perform unnecessary geocoding anymore for the previously selected point (either source or destination), when entering second point

Doesn't it still do unnecessary geocoding for the first point? And then its marker may get shifted as a result of this geocoding.

What an endpoint should do is to detect that it has coordinates as a value and don't try to query nominatim if it does. Search has server-side regexps to detect coordinates:
https://github.com/openstreetmap/openstreetmap-website/blob/7cd900eca0af0f2eaeb54360ec8d1b9153434455/app/controllers/geocoder_controller.rb#L196
Here something simpler might suffice like *number, number* format without all those degree signs, N/S/E/W letters etc.

AntonKhorev:
related #1754

nenad-vujicic:
Dear @AntonKhorev ,

I've just updated PR with the latest, simplified and corrected sources. I made short video displaying new functionality (I added debugging alert("GEOCODING") call to directions.js just before $.getJSON(OSM.NOMINATIM_URL .. line to catch geocoding requests).

https://github.com/openstreetmap/openstreetmap-website/assets/25298521/c4247a43-3f3f-4100-a04b-7debab859945

Now, geocoding is called only once when user selects "Directions from here" or "Directions to there" or when new address is selected. Another way is to call geocoding only when address is typed, but, then, address will be lost when context menu options are selected.

Please, let me know which one option is ideal or if it could be further improved?

Thanks,
Nenad.

app/assets/javascripts/index/directions.js
```diff
@@ -127,8 +146,6 @@ OSM.Directions = function (map) {
     };
 
     endpoint.setLatLng = function (ll) {
-      var precision = OSM.zoomPrecision(map.getZoom());
-      input.val(ll.lat.toFixed(precision) + ", " + ll.lng.toFixed(precision));
```
AntonKhorev, line 131: This code was used by `$("#map").on("drop" ...` to populate an input when the marker next to the input is dragged and dropped on the map.

What happens now:
[Screencast from 04.07.2024 06:35:02.webm](https://github.com/openstreetmap/openstreetmap-website/assets/4158490/9d85463c-aeb6-4007-87f9-32396d413cf7)

What used to happen:
[Screencast from 04.07.2024 06:38:26.webm](https://github.com/openstreetmap/openstreetmap-website/assets/4158490/16a35160-cc5b-46f3-acad-ccfcb9c8c0b0)

nenad-vujicic:
Dear @AntonKhorev ,

Good catch, thank you very much! I've just fixed it and now it works fine! Please, let me know if I could further improve it.

Thanks,
Nenad.

AntonKhorev:
[Going to Nominatim with coordinates](https://github.com/openstreetmap/openstreetmap-website/pull/4904#issuecomment-2186609047) still happens. Try entering "0,0" into from/to inputs.

Before the changes the code always went to Nominatim and searched for these numbers, found them somewhere in Taiwan (or elsewhere depending on the current map view) and placed the marker there.

After the changes the same happens except the marker is at 0, 0.

We should
- either decide that we're not searching for input values when they match the coordinate format,
- or never populate inputs with coordinates (and maybe populate them with reverse geocoding results).

nenad-vujicic:
Dear @AntonKhorev ,

Thanks for reviewing and comments. Now, we are contacting nominatim only if: point is defined using context menu (right click + Directions from/to here) or something is entered that doesn't look like coordinates (Number "," or "/" Number). Also, in separate commit, I improved focusing of map view, so it focuses on union of markers and path (if exists), previously it focused only on path.

Please, let me know if we could further improve it.

Thanks,
Nenad.

AntonKhorev:
Another bug:

https://github.com/user-attachments/assets/9b7a2e50-c768-4c5b-8c4b-c31c1e38f96a

nenad-vujicic:
Dear @AntonKhorev ,

Thank you for catching this! I've just fixed it. Please, let me know if I could further improve it.

Also, can you please restart PR checking workflows? I get very strange error about setting up database.

Thanks,
Nenad.

app/assets/javascripts/index/directions.js
```diff
@@ -82,17 +84,28 @@ OSM.Directions = function (map) {
     input.on("change", function (e) {
       // make text the same in both text boxes
       var value = e.target.value;
-      endpoint.setValue(value);
+      endpoint.setValue(value, false);
     });
 
-    endpoint.setValue = function (value, latlng) {
+    endpoint.setValue = function (value, force_reverse_geocoding, latlng) {
```
AntonKhorev, line 90: What is `force_reverse_geocoding`? Is there any code here that actually does reverse geocoding?

app/assets/javascripts/index/directions.js
```diff
@@ -402,8 +440,10 @@ OSM.Directions = function (map) {
       }
     }
 
-    endpoints[0].setValue(params.from || "", from);
-    endpoints[1].setValue(params.to || "", to);
+    var ll_from_search_form = decodeURIComponent(location.search).match(/^\?from=([+-]?\d+(\.\d*)?)(?:\s+|\s*[/,]\s*)([+-]?\d+(\.\d*)?)$/);
```
AntonKhorev, line 443: Does this regexp ever match anything?

"Directions from here" generate `location.search` like `?from=52.54124,13.56290&to=` which doesn't match. Do you want it to match?

app/assets/javascripts/index/directions.js
nenad-vujicic: Yes, there is. Reverse geocoding is done when user right-clicks on map and selects "Directions from here" or "Directions to here".

Variable force_reverse_geocoding is introduced to pass information that user did this (right clicked and selected from context menu) and to perform reverse geocoding.

app/assets/javascripts/index/directions.js
nenad-vujicic: Yes, it does. It captures the situation when user inputs some coordinates into search field and then pushes "Find directions between two points" button. Earlier, this performed reverse geocoding, now it doesn't do it anymore.

Our idea behind the current solution is "if user enters the coordinates or drags pointers, leave the coordinates. If user right-clicks and selects from context menu, perform reverse geocoding".

app/assets/javascripts/index/directions.js
AntonKhorev: There's no forcing of reverse geocoding. There's no code calling `reverse` like [here](https://github.com/openstreetmap/openstreetmap-website/blob/3debc151a86ed0016364d2d1348ecd82f5fa75fc/app/controllers/geocoder_controller.rb#L184). Nominatim will usually do reverse geocoding when "Directions from/to here" are selected, but that's because it decides on its own that the input query looks like coordinates. It won't make this decision correctly every time.

https://github.com/user-attachments/assets/4a12304d-e834-46f3-ae8f-4a6e147e5144

nenad-vujicic: Thanks for catching this! I've just updated with adding code for calling "reverse" like in above link.

nenad-vujicic: > Thanks for catching this! I've just updated with adding code for calling "reverse" like in above link.

We have just gotten another idea how to improve this part. I'll make another push soon.

nenad-vujicic: @AntonKhorev I've just pushed the latest fixes.

We improved logic of what is displayed in input boxes after right-clicking and selecting option from context menu. Now, if nominatim doesn't return anything or it returns something that is more than 200m away from clicked position, we will display coordinates (Number, Number), otherwise we will display what nominatim returned. The logic behind is "in input boxes should be displayed user-friendly text iff user entered it manually or user clicked on / or near the object recognized by nominatim, otherwise, coordinates will be displayed".

For calculating distance between position where user clicked and position of what nominatim returned, we used distanceTo method provided by Leaflet's LatLng which reduces it to calculating distance on great circle. After testing on Longyearbyen and Longyearbyen, it seems this approximation works fine.

Please, let me know if we can further improve it.

Thanks,
Nenad.

nertc:
When user tries to add a marker out of the borders of countries, alert is thrown "Sorry - couldn't locate", but marker is still added and if user drags it to the valid position, route is not updated (see Video 1).

Video 1:

https://github.com/user-attachments/assets/0c503351-5a3f-4050-bbc8-9f5ed605a0f2

Contrary to this, if user adds marker in the valid location and then drags it out of the borders, route is still calculated and shown without any alerts (see Video 2).

Video 2:

https://github.com/user-attachments/assets/b10d91c4-5c47-4b1e-8975-69f1c97bf088

Maybe, in the first case, we should still show coordinates and calculate route even though Nominatim location was not found (as it is done in the second case).

nertc:
The same logic is with "Couldn't find a route between those two places" error. If it is thrown, dragging location pin doesn't update route (see Video).

Video:

https://github.com/user-attachments/assets/4d45b9d1-be6a-4120-9704-519f71845b13

nertc:
I know that we are using "var" everywhere, but I think, it would be great if we stick to "let" and "const", where using "var" is not absolute necessity. "let" and "const" are less vague about scopes and much bug-proof. There are some examples [Stackoverflow link](https://stackoverflow.com/questions/762011/what-is-the-difference-between-let-and-var). These are the lines you used "var" on:

Line 129: https://github.com/openstreetmap/openstreetmap-website/pull/4904/files#diff-fdb9558c2ba79ddf9fbcb92594a8a8fa54abb1b3e25ce264de37951b428a958bR129

Line 179: https://github.com/openstreetmap/openstreetmap-website/pull/4904/files#diff-fdb9558c2ba79ddf9fbcb92594a8a8fa54abb1b3e25ce264de37951b428a958bR179

Line 301: https://github.com/openstreetmap/openstreetmap-website/pull/4904/files#diff-fdb9558c2ba79ddf9fbcb92594a8a8fa54abb1b3e25ce264de37951b428a958bR301

Line 316: https://github.com/openstreetmap/openstreetmap-website/pull/4904/files#diff-fdb9558c2ba79ddf9fbcb92594a8a8fa54abb1b3e25ce264de37951b428a958bR316

Line 456: https://github.com/openstreetmap/openstreetmap-website/pull/4904/files#diff-fdb9558c2ba79ddf9fbcb92594a8a8fa54abb1b3e25ce264de37951b428a958bR456

Line 469: https://github.com/openstreetmap/openstreetmap-website/pull/4904/files#diff-fdb9558c2ba79ddf9fbcb92594a8a8fa54abb1b3e25ce264de37951b428a958bR469

app/assets/javascripts/index/directions.js
```diff
@@ -109,26 +125,61 @@ OSM.Directions = function (map) {
...
 
-      $.getJSON(OSM.NOMINATIM_URL + "search?q=" + encodeURIComponent(endpoint.value) + "&format=json&viewbox=" + viewbox, function (json) {
-        endpoint.awaitingGeocode = false;
-        endpoint.hasGeocode = true;
-        if (json.length === 0) {
-          input.addClass("is-invalid");
-          alert(I18n.t("javascripts.directions.errors.no_place", { place: endpoint.value }));
-          return;
-        }
+      if (endpoint.value.match(/^([+-]?\d+(\.\d*)?)(?:\s+|\s*[/,]\s*)([+-]?\d+(\.\d*)?)$/)) {
```
nertc, line 128: As we reuse `/^([+-]?\d+(\.\d*)?)(?:\s+|\s*[/,]\s*)([+-]?\d+(\.\d*)?)$/` regex, it would be more understandable if we create separate constant variable for it and then use it with the variable name.

app/assets/javascripts/index/directions.js
```diff
@@ -109,26 +125,61 @@ OSM.Directions = function (map) {
...
+            input.val(endpoint.value);
+          } else {
+            endpoint.value = json.display_name;
+            input.val(json.display_name);
+          }
+
+          geocodeCallback();
+        });
+      } else {
+        $.getJSON(OSM.NOMINATIM_URL + "search?q=" + encodeURIComponent(endpoint.value) + "&format=json&viewbox=" + viewbox, function (json) {
```
nertc, line 152: When concatenating many strings, I think, using [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) are more suitable and easier to read way.

app/assets/javascripts/index/directions.js
```diff
@@ -109,26 +125,61 @@ OSM.Directions = function (map) {
...
+          endpoint.awaitingGeocode = false;
+          endpoint.hasGeocode = true;
+          if (!json || json.length === 0) {
```
nertc, line 155: This can be shortened as `!json?.length`

AntonKhorev: `let/const`, template literals, `?.` etc - we should decide on these in a separate issue, not in this PR.

nertc: If it needs a separate issue, we can create it, however I didn't mean refactoring of the old code (which definitely would be a separate issue), but to change them in the new code. Also, if somewhere we have coding style guidelines, please, link it and I'll create an issue linked to it.

nertc (commented):
Everything I tested, worked perfectly. This is just recommendation for less code to be written.

nertc (approved):

nenad-vujicic:
All above is fixed and PR is again ready to be reviewed. Thanks!

app/assets/javascripts/index/directions.js
```diff
@@ -51,7 +43,7 @@ OSM.Directions = function (map) {
     select.append("<option value='" + i + "'>" + I18n.t("javascripts.directions.engines." + engine.id) + "</option>");
   });
 
-  function Endpoint(input, iconUrl, dragCallback, geocodeCallback) {
+  function Endpoint(input, iconUrl) {
```
AntonKhorev, line 46: You're undoing #4902. I was going to move endpoint code into a separate module because it's difficult to test this one.

nenad-vujicic: I apologize for the undoing, I thought it's just simplifying .. I returned using callback for calling getRoute, so, Endpoint can be pulled out now to separate module.

Thanks,
Nenad.

app/assets/javascripts/index/directions.js
```diff
@@ -5,6 +5,7 @@
 OSM.Directions = function (map) {
   var awaitingRoute; // true if we've asked the engine for a route and are waiting to hear back
   var chosenEngine;
+  var coordinatesRegularExpression = "([+-]?\\d+(\\.\\d*)?)(?:\\s+|\\s*[/,]\\s*)([+-]?\\d+(\\.\\d*)?)";
```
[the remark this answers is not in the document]
AntonKhorev: Why does this expression have capture groups if they are not used? Instead there's `endpoint.value.split(/[/,]/)`.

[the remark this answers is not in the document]
nenad-vujicic: Yes, I agree, I could have used better solutions on many places. My intention was to fix the bug and to make it simple, short and readable as much as I can. The source code is already a little bit complex and requires refactoring. We can optimize it in some other PR.

app/assets/javascripts/index/directions.js
```diff
@@ -82,17 +78,31 @@ OSM.Directions = function (map) {
     input.on("change", function (e) {
       // make text the same in both text boxes
       var value = e.target.value;
-      endpoint.setValue(value);
+      endpoint.setValue(value, false, null, true);
     });
 
-    endpoint.setValue = function (value, latlng) {
+    endpoint.setValue = function (value, force_reverse_geocoding, latlng, fitMap) {
```
AntonKhorev, line 84: Why would a single endpoint control fitting?

Then below you have:

    endpoints[0].setValue(params.from || "", !ll_from_search_form, from, false);
    endpoints[1].setValue(params.to || "", !ll_from_search_form, to, true);

And not it's not clear why the first endpoint gets `fitMap = false` but the second one gets `fitMap = true`.

nenad-vujicic: > Why would a single endpoint control fitting?
> 

Because the route can change on single endpoint update, so, it can go out of viewing area and we need to fit bounds.

> Then below you have:
> 
> ```
> endpoints[0].setValue(params.from || "", !ll_from_search_form, from, false);
> endpoints[1].setValue(params.to || "", !ll_from_search_form, to, true);
> ```
> 
> And not it's not clear why the first endpoint gets `fitMap = false` but the second one gets `fitMap = true`.

Because in some situations there is a flickering if we are doing fitting both times.

AntonKhorev: >> Why would a single endpoint control fitting?

> Because the route can change on single endpoint update

It's for the code outside of endpoints to determine what to do with the route.

app/assets/javascripts/index/directions.js
```diff
@@ -109,31 +119,64 @@ OSM.Directions = function (map) {
...
+
+          endpoint.setLatLng(L.latLng(latlng_array), false, true);
+
+          if (!json || !json.display_name) {
+            input.addClass("is-invalid");
+            alert(I18n.t("javascripts.directions.errors.no_place", { place: endpoint.value }));
+            return;
+          }
+
+          if (endpoint.latlng.distanceTo(L.latLng(json.lat, json.lon)) > 200.0) {
```
AntonKhorev, line 136: What was this distance check about? Was it to filter out useless reverse geocoding results that are too far away from the endpoint?

nenad-vujicic: Yes, exactly! If it is found something near (e.g. 200m) the clicked coordinates, accept it, otherwise, display coordinates.

AntonKhorev: I don't do this in #5064 because it's a different problem, but it probably should be solved somewhere.

What's "near" probably depends on the current zoom level if the coordinates came from a click on the map.

nenad-vujicic:
Closed in favor of #5064

app/assets/javascripts/index/directions.js
AntonKhorev: Looked at what is already in use by included libraries:

- `let/const` - used by Bootstrap and Turbo
- `?.` - used by Turbo
- template literals - used by Bootstrap and Turbo

PR #4905, opened 2024-06-17
Move diary comments hide/unhide actions to comments controller

AntonKhorev:
Continuation of #4889


tomhughes:
Looks good to me, and some nice bug fixes where `unhide` was missed out in various places!

PR #4906, opened 2024-06-17
SOTMEU 2024 banner

matkoniecz:
Consulted with @wzbartczak @Archibaldero (who made banner)

SOTMEU 2024 qualifies for banner and got trademark approval

Follows https://github.com/openstreetmap/openstreetmap-website/blob/master/FAQ.md#how-do-i-create-a-banner-to-promote-my-openstreetmap-event - though in unlikely case of being allowed to start display banner earlier it would be nice


config/banners.yml
```diff
@@ -8,13 +8,13 @@ donate_2023:
...
-  id: sotmeu_2023
-  alt: State of the Map Europe 2023
-  link: https://stateofthemap.eu/
-  img: banners/StateoftheMapEurope_2023.png
-  startdate: 2023-oct-05
-  enddate: 2023-nov-05
+sotmeu_2024:
+  id: sotmeu_2024
+  alt: State of the Map Europe 2024
+  link: https://tobilet.pl/state-of-the-map-europe-2024.html
```
pnorman, line 14: The links have generally gone to the conference page, not the individual ticket page. That would be https://stateofthemap.eu/ in this case.

matkoniecz: Yes, that was intentional request from organizers. @wzbartczak

wzbartczak: Yes, we agreed that the banner should be a direct link to purchase tickets.

matkoniecz: And while making it I assume that such divergence from tradition would be fine, but I have no great experience with this banners.

Though looking at https://github.com/openstreetmap/openstreetmap-website/blob/master/FAQ.md#how-do-i-create-a-banner-to-promote-my-openstreetmap-event again

> link - URL for your event page (e.g. https://supporting.openstreetmap.org/)

I guess it could be interpreted narrowly and as requesting to link the general event page only

tomhughes: I think making it a link to the ticket page is a mistake, especially given the language barrier of that page being mostly in Polish.

gravitystorm: Thanks for pointing this out @pnorman, definitely worth double-checking. 

Since it's been confirmed by the conference organisers, and I think it's really their decision to make, I'm therefore happy to merge this as-is.

gravitystorm:
Thanks @matkoniecz !

matkoniecz:
@gravitystorm Is it maybe possible to make another PR to start display earlier (now)? Or is it intentional delay to allow people review banner in case of some problematic things being missed?

(https://github.com/openstreetmap/openstreetmap-website/blob/master/FAQ.md#how-do-i-create-a-banner-to-promote-my-openstreetmap-event is ordering week delay for initial PR - maybe to avoid people demanding immediate review of PR)

tomhughes:
Well you put the start date in... If you wanted it to start immediately why did you set that? Sure it can be changed, subject to the rules in the policy about how long it can run.

tomhughes:
I believe per https://operations.osmfoundation.org/policies/banner/ you're allowed a month so yes a new PR to remove the start date should be fine.

matkoniecz:
> Well you put the start date in... If you wanted it to start immediately why did you set that? 

mandatory per https://github.com/openstreetmap/openstreetmap-website/blob/master/FAQ.md#how-do-i-create-a-banner-to-promote-my-openstreetmap-event

> The pull request must indicate when the banner should start being shown, which must be at least 7 days after the pull request was created.

PR #4908, opened 2024-06-19
Add the ability to limit changeset size

tomhughes:
This aims to allow the bounding box of changesets to be limited. The basic technology is the same as #4319 with the maximum size ramping up over a period of time for new users.

The defaults allow new users to start with a limit equivalent to a square of 1.5 degrees on each side, and ramp up to the whole world over 28 days with it taking just over three days to start increasing and penalties being applied if the user is reported.

The limit is on the linear dimensions (width + length) not the area of the bounding box to avoid people doing X and Y movements in separate changesets which would then always have an area of zero.

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -130,6 +130,8 @@ class Changeset < ApplicationRecord
   def update_bbox!(bbox_update)
     bbox.expand!(bbox_update)
 
+    raise OSM::APISizeLimitExceeded if bbox.linear_size > size_limit
+
     # update active record. rails 2.1's dirty handling should take care of
     # whether this object needs saving or not.
     self.min_lon, self.min_lat, self.max_lon, self.max_lat = @bbox.to_a.collect(&:round) if bbox.complete?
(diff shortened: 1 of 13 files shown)


gravitystorm:
All looks good to me, thanks!

mmd-osm:
The discussion in #4910 highlights an important point: since the db function only returns the size, the bounding box size calculation needs to be implemented twice now. In case there are some changes (like the discussion in #4910 hints towards), we would have to change it in Rails and CGImap.

Initially, i thought it might be a good idea to simply pass some combination of user_id, min_lat, min_lon, max_lat, max_lon bbox values as a parameter to the database function, and let it figure out the rest.

Any chance we could adjust the db function in this way?

notevenaperson:
I don't know if this is the right place to bring this up. How about changing the openstreetmap-site to use a **heatmap** instead a **box**? The kind of spread out changesets are only annoying because of how the main site _shows_ recent changes.

This change feels like a instance of [tagging for the renderer](https://wiki.openstreetmap.org/wiki/Tagging_for_the_renderer), but here you're changing policy. Don't you see what's happening?

tomhughes:
That's not the reason for this change - the reason for this is to make vandalism harder.

notevenaperson:
> That's not the reason for this change - the reason for this is to make vandalism harder.

I often see experienced editors rioting in the comments of a changeset made by a beginner for having a box that is spanning continents. Rather than being a problem regarding what editors are allowed to do, it's a problem with the UI of the website. I suppose this change is justifiable to prevent a type of vandalism that is hard to undo, but it wouldn't be justifiable if it was for the purpose I initially thought it was for.

PR #4915, opened 2024-06-20
Url-encode wikimedia_commons tag values

### LINKED ISSUE #4914: Wikimedia Commons links with question mark
### URL

https://www.openstreetmap.org/node/3623700647

1. Go to this page in the browser:
https://www.openstreetmap.org/node/3623700647

2. Click on wikimedia_commons link: Category:What If? (Bonn)

3. Get the error.

Similar to https://github.com/openstreetmap/openstreetmap-website/issues/4851

image
image

AntonKhorev:
Fixes #4914.

Standard urlencoding without underscores for spaces.

### DIFF
--- a/app/helpers/browse_tags_helper.rb
+++ b/app/helpers/browse_tags_helper.rb
@@ -116,7 +116,7 @@ module BrowseTagsHelper
   end
 
   def wikimedia_commons_link(key, value)
-    if key == "wikimedia_commons" && value =~ /^(file|category):(.+)$/i
+    if key == "wikimedia_commons" && value =~ /^(file|category):([^#]+)/i
       namespace = Regexp.last_match(1)
       title = Regexp.last_match(2)
       return {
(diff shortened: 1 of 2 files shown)


AntonKhorev:
Sometimes this tag also has fragments:
https://www.openstreetmap.org/way/258522856

That's not going to work not just because of encoding, but also because `?uselang`.
Those fragments shouldn't be there anyway. They're trying to open the image viewer, probably unintentionally as a result of copy-pasting the url when the viewer was open.

tomhughes:
Right but currently the URLs with fragments "work" in that they go to the right page (but the uselang is lost as it becomes part of the fragment) and with this the # will get encoded and it will go to the wrong page?

Maybe we should just add `(?:#.*)?` to the end of the match to discard any fragment?

AntonKhorev:
Looks like urls "work" even with encoded `#` but I'm not sure how reliable this is. Added discarding.

tomhughes:
Looks good to me now, thanks.

PR #4916, opened 2024-06-21
Style auth buttons using .btn & .btn-light classes

kcne:
This PR addresses #4773  discussion regarding auth_button_preferred css classes section. Auth anchor tags now use button styles as base style which will enforce consistency with other buttons.

The login screen maintains the same ui structure with minor styling enchancments
Screenshot 2024-06-21 at 14 31 26

### DIFF
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -77,7 +77,7 @@ module UserHelper
                 :size => "24") + t("application.auth_providers.#{name}.title"),
       auth_path(options.merge(:provider => provider)),
       :method => :post,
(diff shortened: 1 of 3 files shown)


tomhughes (changes requested):
That discussion is about the preferred button on the left (the one with the button style outline) which this PR doesn't actually seem to change? I don't have a problem with also doing the ones on the right but the preferred provider one needs to be done as well.

kcne:
Preferred auth button has been updated as well. Could you please check if everything looks okay?
@tomhughes

kcne:
Screenshot 2024-06-27 at 16 39 43
Screenshot 2024-06-27 at 16 39 58

tomhughes (approved):

PR #4917, opened 2024-06-21
Arrange button sizes to default size of 36px

kcne:
As discussed in #4773, in auth button sizes section it was proposed that 36px should remain as the default size for the auth buttons. This PR addresses that comment.

### DIFF
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -53,7 +53,7 @@ module UserHelper
   # External authentication support
 
   def openid_logo
-    image_tag "openid.svg", :size => "24", :alt => t("application.auth_providers.openid_logo_alt"), :class => "align-text-bottom"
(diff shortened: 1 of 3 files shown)


AntonKhorev:
> it was proposed that 36px should remain as the default size for the auth buttons

Actually https://github.com/openstreetmap/openstreetmap-website/issues/4773 doesn't propose changing the size back to 36px but rather asks why was is changed.

Maybe it was done to accommodate that wider "Log in with ..." button.

kcne:
Regarding this issue, I have a few remarks. I tested the 36px button size on smaller screens, and it displayed correctly without any issues. Additionally, Additionally, we discussed this with our designer, and he agreed that the larger button size looks better and is more user-friendly in this context. @AntonKhorev

tomhughes:
It looks fine to be and I think it looks better so I've merged it, thanks.

I did also add 1c179621d24fbe8a3de71760ecd95e7d431a4405 to switch to the SVG version of the OpenID logo as it became more obviously fuzzy on hidpi screens with the larger size.

Issue #4912, opened 2024-06-19
Automate some code review comments


gravitystorm:
There's a few common code review comments that could be automated (or in the meantime, community members are welcome to jump in and help out the maintainers with these reviews!):

* Any changes to locale files, other than `en.yml`, should be flagged and the PR amended
* Any PRs that have a merge commit should be flagged and the PR amended

In both cases, the response is more-or-less the same content each time, usually with a link to the right place in our documentation. 

We could automate this, which would bring a few benefits, namely
* Faster feedback cycle. Contributors would get the message straight away, while they might still be working on their PR, and not several hours/days later.
* Less work for maintainers - we're overstretched already, so it's one less thing to take care of.
* More polished responses - sometimes we're a bit pressed for time, or forget the link to the documentation, or similar.
* More support for maintainers - like linters and tests, it could highlight things we might miss.

I've seen other projects using Danger for this, since it's able to access the state of the PR in git, i.e. it can reason about the changes, not just the end state (otherwise we'd use linters etc for that).

https://danger.systems/ruby/

Anyone fancy implementing these rules? Or any reasons not to?

rkoeze:
This seems reasonable. One alternative could be to use a [Github PR template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) with a checklist of requirements. With that said I think there are benefits to automatically enforcing certain things.

(I also wonder if using a template to ensure consistent PR descriptions would make life easier for the maintainers. I’m happy to put together a draft PR if that would be helpful, but I don't want to steer this conversation off-topic!)

nenad-vujicic:
> This seems reasonable. One alternative could be to use a [Github PR template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) with a checklist of requirements. With that said I think there are benefits to automatically enforcing certain things.

Or perhaps a combination of these two? Using templates for enforcing better (standardized) descriptions and e.g. Danger for automatized labeling of PRs?

gravitystorm:
I've merged both these PRs, so this issue can be closed. 

If anyone has ideas for how to automate more aspects of reviewing pull requests, please open an issue! The more that we can do automatically, the quicker that contributors will get feedback, and the more likely they are to update their PR.

PR #4919, opened 2024-06-22
Add pull request template

rkoeze:
Adds a PR template to encourage standard descriptions as discussed in https://github.com/openstreetmap/openstreetmap-website/issues/4912#issuecomment-2182690072.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,13 @@
+<!--
+Please read the contributing guidelines before making a PR:
+  https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md
+
+Pay particular attention to the section on how to present PRs:
+  https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests
+-->
+
+### Description
+<!--Describe your changes in detail. If you have made changes to the UI, include screenshots. If your PR addresses a Github issue, please link to it.-->
+
+### How has this been tested?
+<!--Explain the steps you took to test your code.-->


tomhughes:
Does that template really address the things @gravitystorm was talking about? It doesn't seem like it does to me but then I'm not sure a PR template can which is why I was a bit confused by the suggestion - the automation he was suggesting seems more likely to work.

I'm certainly not convinced that asking people to assess the risk of a change is going to be useful - risk of what exactly? If somebody truly thought a change was risky then they shouldn't really be proposing it!

rkoeze:
Hi @tomhughes! I think there's probably a version of a PR template that does address what @gravitystorm was talking about (most likely in the form of a checklist). This template does not try to do that. It's aim is to enforce a consistent PR description. If that doesn't make sense for this repo, I'm happy to close this PR. After all, this change is meant to make the lives of reviewers easier!

I can definitely remove the risk section. I've found that in other projects asking for a risk assessment can be a useful exercise for the contributor. It can also give the reviewer helpful context when testing. A worst case idea of "what could go wrong here" is what I mean by risk.

Thank you for your feedback.

.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+<!--Please read the contributing guidelines before making a PR: https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md-->
```
tordans, line 1:
```suggestion
<!-- 
Please read the contributing guidelines before making a PR: 
  https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md

Especially the section on how to present PRs:
  https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests
-->
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
...
+
+### Description
+<!--Describe your changes in detail. If you have made changes to the UI, include screenshots. If your PR address a Github issue, please link to it.-->
+
+### How has this been tested?
+<!--Explain the steps you took to test your code.-->
+
+### Risks
+<!--Is this a low, medium, or high risk change? Explain why.-->
```
tordans, line 10:
```suggestion
```

tordans (commented):
I think this is a good template that will make thinks better for some PRs. I certainly does not harm the process, because one can always remove all this very easily when creating a PR that does not fit this schema.

I would also remove the risk section. And make the intro even more explicit on the PR section which is where all the "do squash, do rebase, no merge commits" part is written.

rkoeze:
Thanks @tordans! I removed the risk section and added the link to the PR section of the contributing guidelines as you suggested.

nenad-vujicic:
Dear @rkoeze ,

I really like idea of using PR templates. Thank you very much for it and for your PR. But, here is how it looks when I add it to my forked repository (maybe I made mistake while adding to repository, please, let me know):

image

Wouldn't be better if it would look like when reporting issue, i.e. something like this:

image

So, perhaps we could create new folder (.github/PULL_REQUEST_TEMPLATE) and sets of templates (perhaps in YML format, like for issues?), like in case of reporting issue, where we will fill the form rather than edit template?

Thanks,
Nenad.

rkoeze:
Hi @nenad-vujicic! Thank you for the feedback. If you click preview, the markdown should correctly render the "Description"  and "How has this been tested?" sections. The rest of the template should not render as it's only intended to be viewed by the person writing the PR.

As far as I'm aware it's not possible to use YML for PR templates (although I'm happy to be corrected on that point). One advantage of using markdown in this fashion is that while it provides a schema for the PR description, the individual making the PR can still alter it if it does not suit their needs.

nenad-vujicic (approved):
> ...
> As far as I'm aware it's not possible to use YML for PR templates (although I'm happy to be corrected on that point). One advantage of using markdown in this fashion is that while it provides a schema for the PR description, the individual making the PR can still alter it if it does not suit their needs.

You are right, thanks, we cannot use YML for PR templates. We can embed external forms, but, still through markdown, which wouldn't look so nice.

I anyway, still very like this PR and think it will make the workflow better!

Thanks,
Nenad.

gravitystorm:
Thanks for the suggestion @rkoeze and for making changes based on the feedback. I've merged this now, let's see how it goes!

Thanks also to @tordans and @nenad-vujicic for reviewing this PR, it's appreciated.

PR #4922, opened 2024-06-25
Simplify handling of geocoder URLs

tomhughes:
This avoids having to build them in multiple places and also ensures we link to what was actually searched rather than some random string from the locale file.


gravitystorm:
Looks good to me, thanks.

PR #4923, opened 2024-06-25
Add entries for cowshed and sty to en.yml

### LINKED ISSUE #3449: Add building=cowshed and building=sty to en.yml
### Description

Add [building=cowshed](https://wiki.openstreetmap.org/wiki/Tag:building%3Dcowshed) and [building=sty](https://wiki.openstreetmap.org/wiki/Tag:building%3Dsty)  to en.yml so that they can be translated. ([building=stable](https://wiki.openstreetmap.org/wiki/Tag:building%3Dstable) is already there).

### Screenshots

_No response_

Hufkratzer:
This adds entries for [building=cowshed](https://wiki.openstreetmap.org/wiki/Tag:building%3Dcowshed) and [building=sty](https://wiki.openstreetmap.org/wiki/Tag:building%3Dsty) to en.yml so that they can be translated. 

Resolves #3449


tomhughes:
Each of those has about 25000 uses according to taginfo.

Does nominatim actually have any search results for them though? I guess if there are any which a name then it might be able to find them?

tomhughes:
Well the strings you wanted to add are for displaying search results aren't they? They're under `geocoder.search_osm_nominatim`...

mmd-osm:
Oh, it seems I pointed the OP to the wrong location in that file…

Hufkratzer:
The reason I opened https://github.com/openstreetmap/openstreetmap-website/issues/3449 two years ago was that I wanted to be able to translate "cowshed" and "sty" to German in this list:
stall

I can not find these strings in Translatewiki and think adding them in en.yml will solve that.

tomhughes:
Ah yes actually the query tool does also use those strings, despite the name, so this is good to merge I think.

Issue #670, opened 2014-01-08
Allow home location marker on main map to be hidden


pnorman:
Out of #669, we don't currently provide a way to hide the home location marker after using "Go to Home Location". Regardless of that issue, it'd be nice to have a way to hide it, short of reloading.

AntonKhorev:
Was fixed in #4926. The marker is shown on `/account/home` page and removed when visiting any other page. Like https://github.com/openstreetmap/openstreetmap-website/issues/670#issuecomment-31867047, but there's no sidebar, and therefore no close button, but you can for example click the osm logo to visit `/` and remove the marker.

PR #4926, opened 2024-06-27
Home Location page

### LINKED ISSUE #669: Clicking on logo does not remove home location marker
Clicking on "View" used to remove the marker from "Go to Home Location" but clicking on the logo does not do so.

### LINKED ISSUE #1810: Go to Your Home Location gone sometimes
Logged in, visit
https://www.openstreetmap.org/

Click your head in the upper right.
Notice Go to Your Home Location.

Now click https://www.openstreetmap.org/traces and
click your head in the upper right.
Notice Go to Your Home Location is gone.

AntonKhorev:
This adds a home location page with its own URL similar to what's suggested in https://github.com/openstreetmap/openstreetmap-website/issues/670#issuecomment-31867047, without sidebar and extra controls. Instead of `/:username/home` I use the path `/account/home` because it's easier to implement, no need to check `:username`.

A link to this page is available on every other page with the user dropdown menu. This fixes #1810.

image

It's possible to leave this page for another map page. In this case the javascript controller removes the home location marker. This fixes #669 and #670.

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -29,7 +29,7 @@ class Ability
 
     if user&.active?
       can :welcome, :site
-      can :read, [:deletion, :account_terms, :account_pd_declaration]
+      can :read, [:deletion, :account_terms, :account_pd_declaration, :account_home]
 
       if Settings.status != "database_offline"
         can [:read, :create, :destroy], :changeset_subscription
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -21,6 +21,7 @@
 //= require index/directions
 //= require index/changeset
 //= require index/query
+//= require index/home
 //= require router
 
 $(document).ready(function () {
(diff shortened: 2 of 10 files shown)


nertc:
Described functionality works perfectly. Redirection is flawless in all situations (browser back/forward button navigation, page refresh, browser link navigation, app menu navigation). Home icon appears in /account/home and disappears as soon as a user is redirected to another page. The only question I have, wouldn't it be better if we used I18n translate function `I18n.t` instead of hard coded text while using `click_on` and `assert_text` in tests. In that case, even if, in the future, we decide to change texts, it won't break tests.

AntonKhorev:
> The only question I have, wouldn't it be better if we used I18n translate function I18n.t instead of hard coded text while using click_on and assert_text in tests. In that case, even if, in the future, we decide to change texts, it won't break tests.

Changing names of i18n strings will break tests then.

AntonKhorev:
Updated for let/const.

tomhughes:
Looks good to me, thanks.

PR #4929, opened 2024-06-28
Rename map layer javascript files

AntonKhorev:
We have a bunch of things named "browse". On the client side we have the `OSM.Browse` controller for single element pages: https://github.com/openstreetmap/openstreetmap-website/blob/ecd091c976cc8f2a3cf96fe08dc3211aad689ae2/app/assets/javascripts/index.js#L327

But there's also [browse.js](https://github.com/openstreetmap/openstreetmap-website/blob/ecd091c976cc8f2a3cf96fe08dc3211aad689ae2/app/assets/javascripts/index/browse.js). It initializes the data layer:
image

This is confusing because most of controllers, for example `OSM.Directions`, are in their own javascript files. You'd expect to find `OSM.Browse` in `browse.js` but no, `browse.js` is something completely different.

This pull request renames everything "browse" about `browse.js` and moves the file into the `layers` directory. And it also moves notes layer there.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -12,9 +12,9 @@
 //= require leaflet.contextmenu
 //= require index/contextmenu
 //= require index/search
-//= require index/browse
+//= require index/layers/data
 //= require index/export
-//= require index/notes
+//= require index/layers/notes
 //= require index/history
 //= require index/note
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good to me, thanks.

PR #4930, opened 2024-06-28
Browse node version coordinates

AntonKhorev:
It's now possible to see node locations on node version pages for visible versions:

image
image

Also no extra api requests to fetch node coordinates for map markers.
Before (XHR) / after (no XHR):
image image

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -327,17 +327,17 @@ $(document).ready(function () {
   OSM.Browse = function (map, type) {
     var page = {};
 
-    page.pushstate = page.popstate = function (path, id) {
+    page.pushstate = page.popstate = function (path, id, version) {
       OSM.loadSidebarContent(path, function () {
-        addObject(type, id);
+        addObject(type, id, version);
       });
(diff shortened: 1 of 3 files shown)


tomhughes:
Adding the coordinates to the history view is obviously no problem, but it's also only a tiny part of this change.

Most of this change is I think to implement the second part - so in order to save the API call we're adding a whole load of complication and causing nodes to handled completely differently to other geometry types. Is that really a good idea?

AntonKhorev:
There's no single API call to get geometry old versions of elements other than nodes. If you don't want to save the API call, you'll have to handle nodes differently, namely only show old nodes but not elements of other types on the map. This PR saves the call for nodes in order to do later the same for other types.

jidanni:
Hmm, Wikipedia also allows one to link to any exact version of anything.

mmd-osm:
> There's no single API call to get geometry [of] old versions of elements other than nodes.

This doesn’t mean we couldn’t introduce such an API endpoint. Ways should be ok, relations could be a bit expensive.

PR #4933, opened 2024-06-30
Update en.yml to new world view

jidanni:
Don't say "other" or "all other". Say instead "some" or "many".

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2747,7 +2747,7 @@ en:
       support: support
       about:
         header: Free and editable.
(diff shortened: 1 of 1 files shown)


tomhughes:
Who or what exactly are you quoting as giving that advice? Whose "world view" is this updating to?

mmd-osm:
My interpretation of what the OP wants to say is that OSM is not the only map that "_is completely created by people like you, and it's free for anyone to fix, update, download and use_". (think of OpenHistoricalMap or OpenGeofiction, maybe?).

If there's really nothing left to work on, we could maybe discuss this. For now, i would just close it as not planned.

gravitystorm:
> If there's really nothing left to work on, we could maybe discuss this. For now, i would just close it as not planned.

Agreed.

PR #4934, opened 2024-06-30
Set i18n locale before writing "report a problem" on embed page

AntonKhorev:
`/export/embed` page has the "Report a problem" link which is supposed to be localized depending on the browser's accept-language header. But that doesn't happen because javascript waits for `window.onload` before setting the language. By that time `L.Control.OSMReportAProblem` is already created with a default locale.

image


tomhughes:
Looks good to me, thanks.

PR #4935, opened 2024-07-01
Refactor login form to use form helpers for password field

kcne:
This PR addresses issue #4773: "Raw HTML input used for password field." The login form previously used raw HTML for the password input field, which bypassed the automatic HTML attribute generation and site-wide overrides provided by form helpers.

- Replaced the raw HTML input for the password field with the f.password_field helper.
- Ensured the alignment of the password label and the "forgot password" link using Bootstrap classes.

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -33,15 +33,12 @@
 
     <%= f.text_field :username, :label => t(".email or username"), :autofocus => true, :tabindex => 1, :value => params[:username] %>
 
-    <div class="row">
-      <div class="col">
-        <%= f.label :password, t(".password"), :class => "form-label" %>
-      </div>
-      <div class="col text-end">
-        <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
-      </div>
+    <div class="d-flex flex-wrap column-gap-3 justify-content-between align-items-baseline">
+      <%= f.label :password, t(".password"), :class => "my-2" %>
+      <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
     </div>
-    <input class="form-control mb-3" type="password" name="password" id="password" tabindex="2" value="" autocomplete="on" />
+
+    <%= f.password_field :password, :autocomplete => "on", :tabindex => 2, :value => "", :skip_label => true %>
 
     <%= f.form_group do %>
       <%= f.check_box :remember_me, { :label => t(".remember"), :tabindex => 3, :checked => (params[:remember_me] == "yes") }, "yes" %>
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -33,8 +33,8 @@
 
     <%= f.text_field :username, :label => t(".email or username"), :autofocus => true, :tabindex => 1, :value => params[:username] %>
 
-    <div class="d-flex flex-wrap column-gap-3 justify-content-between align-items-baseline">
-      <%= f.label :password, t(".password"), :class => "my-2" %>
+    <div class="d-flex flex-wrap column-gap-3 justify-content-between align-items-baseline mb-2">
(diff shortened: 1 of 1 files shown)


app/views/sessions/new.html.erb
```diff
@@ -33,15 +33,12 @@
...
 
-    <div class="row">
-      <div class="col">
-        <%= f.label :password, t(".password"), :class => "form-label" %>
-      </div>
-      <div class="col text-end">
-        <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
-      </div>
+    <div class="col d-flex justify-content-between align-items-center">
+      <%= f.label :password, t(".password"), :class => "my-1" %>
```
tomhughes, line 37: I think that class wants to be `my-2` to make the whole row the same height as before, and as the label for the username - as it is the label for password is closer to the field than the label for username is.

tomhughes (changes requested):

app/views/sessions/new.html.erb
kcne: Spot on, resolved.

app/views/sessions/new.html.erb
```diff
@@ -33,15 +33,12 @@
...
-      <div class="col text-end">
-        <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
-      </div>
+    <div class="col d-flex justify-content-between align-items-center">
+      <%= f.label :password, t(".password"), :class => "my-2" %>
+      <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
     </div>
-    <input class="form-control mb-3" type="password" name="password" id="password" tabindex="2" value="" autocomplete="on" />
+
+    <%= f.password_field :password, :autocomplete => true, :tabindex => 2, :value => "", :skip_label => true %>
```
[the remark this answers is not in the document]
kcne: My bad, missed that one. Resolved now.

app/views/sessions/new.html.erb
```diff
@@ -33,15 +33,12 @@
...
-    <div class="row">
-      <div class="col">
-        <%= f.label :password, t(".password"), :class => "form-label" %>
-      </div>
-      <div class="col text-end">
-        <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
-      </div>
+    <div class="col d-flex justify-content-between align-items-center">
+      <%= f.label :password, t(".password"), :class => "my-2" %>
+      <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
```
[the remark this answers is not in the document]
kcne: No need for col, got confused there a bit. 
Resolved with a new commit.
Thank you for review

app/views/sessions/new.html.erb
```diff
@@ -33,15 +33,12 @@
...
 
-    <div class="row">
-      <div class="col">
-        <%= f.label :password, t(".password"), :class => "form-label" %>
-      </div>
-      <div class="col text-end">
-        <small><%= link_to(t(".lost password link"), user_forgot_password_path) %></small>
-      </div>
+    <div class="d-flex flex-wrap column-gap-3 justify-content-between align-items-baseline">
+      <%= f.label :password, t(".password"), :class => "my-2" %>
```
AntonKhorev, line 37: https://github.com/openstreetmap/openstreetmap-website/pull/4935#discussion_r1661380070 talks about the height of the whole row, yet the margin is applied just to the label. Why not:
```suggestion
    <div class="d-flex flex-wrap column-gap-3 justify-content-between align-items-baseline mb-2">
      <%= f.label :password, t(".password") %>
```

Before/after:
image image

Also we usually add bottom margin, not both top and bottom.

AntonKhorev (changes requested):

AntonKhorev:
merged, thanks

PR #4936, opened 2024-07-01
Separate translatable messages for privacy policy link in signup form

kcne:
This PR addresses the need for distinct translatable messages for the privacy policy link in different contexts within the signup form mentioned in #4773.

In the signup form, the privacy_policy_link and its respective label were reused in two different sentences (in email help HTML and by signing up HTML messages). Different contexts may require distinct translations due to language-specific rules, such as declension.
##### Changes Made
- Added separate translation keys for the privacy policy link in the email_help and by_signing_up sections.
- Updated the form to use these new translation keys, ensuring the correct link label is used in each context.

This PR aims to solve the issue of translation nuances by providing separate translatable messages for the privacy policy link in different contexts within the signup form. This approach should enhance localization and ensure accurate translations.

Intuitively this makes sense, but any suggestions for improvements are very welcome.

### DIFF
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -44,10 +44,10 @@
     <%= f.hidden_field :auth_uid unless current_user.auth_uid.nil? %>
 
     <% if current_user.auth_uid.nil? or @email_hmac.nil? or not current_user.errors[:email].empty? %>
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #4941, opened 2024-07-03
Change auth provider logo alt texts to indicate that these are logos

AntonKhorev:
Auth provider images have `alt` attributes that instead of describing the images say what's going to happen when they are clicked. This is wrong and redundant because their buttons already have titles that say "Log in with ...":

image

This PR changes the alt texts to "... logo", similar to the alt text of OSM logo:

image


HolgerJeromin (approved):

tomhughes:
Looks good to me, thanks.

PR #4942, opened 2024-07-03
Remove unused gradients from note markers

AntonKhorev:
You can easily see that they are not used because they don't have ids.


tomhughes:
Looks good to me, thanks.

PR #4943, opened 2024-07-03
Move source images to assets directory

AntonKhorev:
Now we have the `assets` directory that is outside of `app`. It is for images that are not directly served, likely unoptimized svgs which you may want to open with an image editor. Pull requests like #4793 move source svgs that correspond to optimized svgs into that directory. But there's also a couple of images that don't have directly served equivalents. This PR moves them into `assets`.


tomhughes:
Looks good to me, thanks.

Issue #4613, opened 2024-03-22
Order of comments in changeset comments feeds


AntonKhorev:
### Problem

Does it make sense for comments in a feed to go in ascending order, oldest first?

Comment feeds for all comments are sorted in descending order:
https://master.apis.dev.openstreetmap.org/history/comments/feed

But feeds for individual changesets are sorted in ascending order:
https://master.apis.dev.openstreetmap.org/changeset/307182/comments/feed

If you add a limit, and there's more comments than the limit, you'll never get new comments:
https://master.apis.dev.openstreetmap.org/changeset/307182/comments/feed?limit=1

milan-cvetkovic:
We probably want descending order, with newest comments being fist on the list.

I think it would also be useful to order them by `id` instead of `created_at`, since in theory ordering by timestamp and id does not have to match.

PR #4944, opened 2024-07-03
Order of comments in changeset comments feeds

nertc:
This PR addresses "Order of comments in changeset comments feeds" issue mentioned in the [#4613](https://github.com/openstreetmap/openstreetmap-website/issues/4613)
 
Reverse the output of the changeset comments feed from "created_at" ascending order to descending. Fixes [#4613](https://github.com/openstreetmap/openstreetmap-website/issues/4613)

Screenshot 2024-07-03 133736

### DIFF
--- a/test/controllers/changeset_comments_controller_test.rb
+++ b/test/controllers/changeset_comments_controller_test.rb
@@ -46,6 +46,23 @@ class ChangesetCommentsControllerTest < ActionDispatch::IntegrationTest
         assert_select "item", :count => 3
       end
     end
+    # Rails::Dom::Testing.html_document_fragment.parse(icons)
+    # Gets comment Ids from HTML and checks that they are in descending order
+
+    last_comment_id = -1
+    assert_select "rss", :count => 1 do
+      assert_select "description", :count => 3 do |descriptions|
+        descriptions.children.each do |description|
+          changeset_dom = Rails::Dom::Testing.html_document_fragment.parse(description.content)
+          comment = changeset_dom.at_css(".changeset-comment-text")
+          next unless comment
+
+          id = comment.content.split[-1].to_i
+          assert_operator id, "<", last_comment_id if last_comment_id != -1
+          last_comment_id = id
+        end
+      end
+    end
   end
 
   ##
--- a/app/controllers/changeset_comments_controller.rb
+++ b/app/controllers/changeset_comments_controller.rb
@@ -18,7 +18,7 @@ class ChangesetCommentsController < ApplicationController
       changeset = Changeset.find(id)
 
       # Return comments for this changeset only
-      @comments = changeset.comments.includes(:author, :changeset).limit(comments_limit)
+      @comments = changeset.comments.includes(:author, :changeset).reverse_order.limit(comments_limit)
     else
       # Return comments
       @comments = ChangesetComment.includes(:author, :changeset).where(:visible => true).order("created_at DESC").limit(comments_limit).preload(:changeset)


AntonKhorev:
Unfortunately the tests don't check which comments are in feeds, only their counts.

nertc:
@AntonKhorev Thank you for the review. I updated tests, so they check order of the feed.

test/controllers/changeset_comments_controller_test.rb
```diff
@@ -46,6 +46,14 @@ def test_feed
         assert_select "item", :count => 3
       end
     end
+
+    # Gets comment Ids from HTML and checks that they are in descending order
+    last_comment_id = -1
+    @response.body.split("Changeset comment ")[1..].each_with_index do |data, index|
+      id = Integer data.split("<")[0]
```
tomhughes, line 53: The normal way to convert a string to an integer is to use `.to_i` rather than calling the constructor like this.

test/controllers/changeset_comments_controller_test.rb
```diff
@@ -46,6 +46,14 @@ def test_feed
         assert_select "item", :count => 3
       end
     end
+
+    # Gets comment Ids from HTML and checks that they are in descending order
+    last_comment_id = -1
+    @response.body.split("Changeset comment ")[1..].each_with_index do |data, index|
```
tomhughes, line 52: It would be better to test against the parsed response rather than trying to use string operations to inspect HTML embedded in feed XML from the raw response.

What I'm not sure about is how easy that is when you're looking at the HTML content.

I think we should certainly be able to use he parsed response to find the feed entries though even if we then extract the ID from the HTML using string operations.

tomhughes (changes requested):

test/controllers/changeset_comments_controller_test.rb
nertc: Thanks for comments. I agree with you, but when I parsed XML, the only id info was still in the CDATA containing HTML (fields exposed in the XML were identical for every comment e.g. comment creation time, because they are created with no delay). So even if I used parsed data, I would still use HTML raw string manipulation. Certainly, it's doable (and at first I did it that way), but it makes the test more complicated (so I rewrote it this way). If it's okay, test to be more complicated, I'll do it with XML parser, but I think this solution is already more complicated than the regular test should be.

test/controllers/changeset_comments_controller_test.rb
nertc: Thanks for comments. I agree with you, but when I parsed XML, the only id info was still in the CDATA containing HTML (fields exposed in the XML were identical for every comment e.g. comment creation time, because they are created with no delay). So even if I used parsed data, I would still use HTML raw string manipulation. Certainly, it's doable (and at first I did it that way), but it makes the test more complicated (so I rewrote it this way). If it's okay, test to be more complicated, I'll do it with XML parser, but I think this solution is already more complicated than the regular test should be.

AntonKhorev: I think there is a test somewhere that parses embedded html.

nertc: I changed the test to use parsed XML and HTML parser.

test/controllers/changeset_comments_controller_test.rb
nertc: I rewrote the test in a different way

test/controllers/changeset_comments_controller_test.rb
AntonKhorev: This is the test I meant
https://github.com/openstreetmap/openstreetmap-website/blob/e3c919275818d46a515c183cf373d7a53bc6942a/test/controllers/traces_controller_test.rb#L729

tomhughes:
Thanks. I think that looks good now.

PR #4945, opened 2024-07-03
Use inline svg for new diary entry icon

AntonKhorev:
with `fill=currentColor`, should look better in dark mode

Before:
image image

After:
image image


tomhughes:
Looks good to me, thanks.

PR #4947, opened 2024-07-03
Move continue auth/start mapping button to appear fixed to the bottom of the welcome page     

kcne:
#### Objective  
To enhance the user experience by pinning the existing "Continue Auth/Start Mapping" button at the bottom of the page. This adjustment is aimed at improving user navigation and simplifying the OAuth process.  
  
Screenshot 2024-07-03 at 17 16 28

#### Key Points  
- **Enhanced UX:** By fixing the button at the bottom, we ensure it’s always visible, making it more prominent for users during signup flow.
- **Simplified OAuth Process:** This change makes the OAuth process more intuitive and seamless, reducing potential user confusion and promoting a smoother transition.

### DIFF
--- a/app/views/site/welcome.html.erb
+++ b/app/views/site/welcome.html.erb
@@ -77,17 +77,17 @@
 
 <%= render "any_questions" %>
 
-<div class='text-center mb-3'>
-<% if params[:oauth_return_url] %>
-  <a class="btn btn-primary" href="<%= params[:oauth_return_url] %>"><%= t ".continue_authorization" %></a>
-<% else %>
-  <a class="btn btn-primary start-mapping" href="<%= edit_path %>"><%= t ".start_mapping" %></a>
-<% end %>
-</div>
-
-<div class='alert alert-primary'>
+<div class='alert alert-primary mb-5'>
   <h2><%= t ".add_a_note.title" %></h2>
   <p><%= t ".add_a_note.para_1" %></p>
   <p><%= t ".add_a_note.para_2_html", :map_link => link_to(t(".add_a_note.the_map"), root_path),
                                       :note_icon => tag.span(:class => "icon note bg-dark rounded-1") %></p>
+</div>
+
+<div class='d-flex justify-content-center align-items-center mw-100 fixed-bottom bg-white py-2 border-top'>
+  <% if params[:oauth_return_url] %>
+    <a class="btn btn-primary d-block d-sm-inline-block" href="<%= params[:oauth_return_url] %>"><%= t ".continue_authorization" %></a>
+  <% else %>
+    <a class="btn btn-primary start-mapping d-block d-sm-inline-block" href="<%= edit_path %>"><%= t ".start_mapping" %></a>
+  <% end %>
 </div>


tomhughes:
The obvious risk here is that it encourages people to hit the button and get going without reading the page first.

stalker314314:
Just to give more context to the public - Map builder started experimenting with SSO flow and we notice that "only" 35% of users finishes sign-up procedure (@pietervdvn can comment too, if you have numbers for MapComplete). This PR tries to improve sign-up rate for both users coming to osm.org and coming from third-party apps.

There is not much to say that isn't already said in https://github.com/openstreetmap/openstreetmap-website/issues/4246. In that solution, we thought it would be good to skip "welcome" page (or shorten it) as it presents partially unnecessary information for users coming from other apps and move part of its content for async consumption (e-mail) and move education bits to editors. If we don't want to go with that solution, it would be at least good to have this button pinned. We tried to do pinning for both buttons ("Start mapping" and "Continue auth") as per @tomhughes comment in https://github.com/openstreetmap/openstreetmap-website/issues/4246#issuecomment-1770532334. 

> The obvious risk here is that it encourages people to hit the button and get going without reading the page first.

I guess if people do not want to read page, they will just scroll to the bottom even today (if they figure out there is something at the bottom!). Those who want to read it - can read it, this change doesn't block that. Most important parts IMHO ("What's on the map" and about not copying from other maps) are prominent and visible even without scrolling.

**At the end of day - this is just a standard visual clue to signify to users that there is something to click/proceed.**

P.S. If you think it is mandatory for users to read it, we can disable button until users scroll to the bottom (as some legal text require user to do), but I hope we are not going there (and whole solution would look like "design by committee" to me😄).

P.P.S. I can go to philosophical rent about ad tech/social network companies and web trends and how they used "engagement" metrics and stole our attention span, so that now every wall of text is insurmountable mountain to pass, and how it affects osm.org, but it is more for diary:)

tordans:
I agree that this improves the UX and Flow with IMO minimal negative side effects.

I also still think a separate flow for this use case as discussed in the linked issue (that removes this page but preserves the information as part of a follow up email) would be better. However, given the constrains to not do this, this is a good idea.

stalker314314:
@tomhughes With given explanation and @tordans comment, can you reassess this? Otherwise, can you propose alternative approach (which is not status quo) on this? We proposed disabling button until page is scrolled to the bottom, what are your thoughts to that? Or please close PR if you don't want to change anything, it will help us focus on other stuff

tomhughes:
I was kind of leaving this to see if any of the other maintainers had an opinion before considering it again... Any thoughts @gravitystorm or @AntonKhorev?

AntonKhorev:
If we want something like this we can add a footer to the layout similar to `.content-heading`:

Screen Shot 2024-07-18 at 16 52 30

image

https://github.com/AntonKhorev/openstreetmap-website/tree/x/welcome-sticky-footer

kcne:
@AntonKhorev I agree, would you like me to fetch your commits and add them to this PR, or do you prefer to open a new PR for this change?

AntonKhorev:
@kcne If you want to finish this PR you can use my commits here.

tomhughes:
Something has gone badly wrong - you've just pushed over a hundred commits that look like they been cherry picked from master or something. I think you've managed to make your history diverge somehow.

kcne:
Didn't rebase from master branch. Can you check if everything is okay now?

tomhughes:
Yes it's fixed now.

AntonKhorev:
The commits here should probably be squashed because they partly undo previous commits. And we have [conventions](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#committing) for commit messages like starting with a capital letter.

kcne:
Squashed commits and renamed according to conventions. If you have any additional ideas how to improve this further please let me know! Thank you.

app/views/site/welcome.html.erb
```diff
@@ -77,17 +77,20 @@
...
   <p><%= t ".add_a_note.para_1" %></p>
   <p><%= t ".add_a_note.para_2_html", :map_link => link_to(t(".add_a_note.the_map"), root_path),
                                       :note_icon => tag.span(:class => "icon note bg-dark rounded-1") %></p>
 </div>
+
+<% content_for :footer_outer_class, "sticky-bottom" %>
+<% content_for :footer do %>
+  <div class='d-flex justify-content-center align-items-center'>
+    <% if params[:oauth_return_url] %>
+      <a class="btn btn-primary d-block d-sm-inline-block" href="<%= params[:oauth_return_url] %>"><%= t ".continue_authorization" %></a>
```
AntonKhorev, line 91: What are `d-block d-sm-inline-block` classes supposed to do on a flex item?

kcne: To be honest, this PR was from a while ago, and I don't remember exactly why I added it. I must have encountered some layout issues at the time. However, I've now removed it locally, and everything looks fine on both small and large screens, so I'll just go ahead and remove it since it seems unnecessary.

kcne: Updated the code, thank you.

gravitystorm:
@kcne are you able to rebase this PR please?

kcne:
Rebased and tested manually in the browser. Should be ok now

gravitystorm:
Thanks @kcne for your patience and for the rebase! Now merged.

Issue #4543, opened 2024-02-26
Password Visibility Toggle in Login Page.


Ri-Sharma:
### Problem

In the current login page implementation, users might face difficulty while entering their password due to its hidden nature. This lack of visibility can lead to errors during input. 

### Description

To enhance the user experience and facilitate accurate password entry, I propose the implementation of a password reveal functionality. This feature, commonly denoted by an eye icon, allows users to toggle between obscured and visible password states, providing user with the flexibility to verify their input. This improvement aims to contribute to a smoother login process.

### Screenshots

![image](https://github.com/openstreetmap/openstreetmap-website/assets/80570105/7e2f6790-4a88-4eba-93c0-9c76ef675be8)

IshmeetSingh06:
I can pick this up if it's an actual requirement @gravitystorm

sjvudp:
For me it's not an issue (on the desktop at least), because I use a password manager to either
* paste the password from the clipboard, or
* use "auto-type" to let the password manager fill in the (username and) password field

On the mobile phone with a touchscreen keyboard it might be an issue if typing the password manually, however.

PR #4951, opened 2024-07-05
Password fields with show/hide functionality

kcne:
### Implementation Details

This PR introduces functionality that allows users to toggle the visibility of password fields in signup and login forms. This enhancement improves both usability and accessibility.

### Changes Made

1. **Add `password_field_with_toggle` partial**
   - Integrated a reusable Ruby component that can be utilized within both login and signup forms.

2. **Add Show/Hide JavaScript Helpers**
   - Implemented JavaScript functionality to toggle the visibility of password fields using the `addTogglePasswordEventListener` function.

### Screenshots

Screenshot 2024-07-04 at 23 41 05

Signup Page

Screenshot 2024-07-05 at 17 02 56

Signup Page with validation errors

Screenshot 2024-07-05 at 17 04 23

Login Page


AntonKhorev:
Right click > Reveal password in Firefox

HolgerJeromin:
ref #4543

AntonKhorev:
Anyone who wants this functionality probably wants it not just for openstreetmap website. They are using a browser that has it or install an extension.

mmd-osm:
Right, Edge doesn't need this code, the browser would automatically add it, unless you disable it using some CSS code: https://learn.microsoft.com/en-us/microsoft-edge/web-platform/password-reveal

Since I didn't find any special CSS for edge, I was wondering, if adding this button would interfere with what Edge is doing.

kcne:
I'll try and test this with Edge. I can try to conditionally render this buttons according to the browser being used.

HolgerJeromin:
> according to the browser being used

Browser sniffing is and was always bad. Please try to avoid it.

kcne:
According to [docs](https://learn.microsoft.com/en-us/microsoft-edge/web-platform/password-reveal) it can be disabled using CSS only. Amended the changes to the commit.

AntonKhorev:
> According to [docs](https://learn.microsoft.com/en-us/microsoft-edge/web-platform/password-reveal) it can be disabled using CSS only

According to docs you shouldn't do this.

stalker314314:
I agree that it can be said that we are fiddling with what could be work of "user agent" (in original sense). Put that way - this PR is not needed. However, I think that #4543 should be closed too (or at least rephrased), just that it doesn't give impression this is something it needs to  be implemented

PR #4952, opened 2024-07-05
Use alt text in OpenID image alt attribute

AntonKhorev:
Continuing #4941 for OpenID button that had its image alt attribute set to its title text.


tomhughes:
Looks good to me, thanks.

PR #4953, opened 2024-07-05
Remove padding around tabs on blocked signup page

AntonKhorev:
#4826 didn't remove `px-3` from all pages. One more page with login/signup tabs is the blocked signup page.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #4954, opened 2024-07-05
Set "login/signup" page width equal to "about" page width

AntonKhorev:
Currently login pages with preferred auth provider look like [this](https://www.openstreetmap.org/login?referer=https%3A%2F%2Fexample.com%2F%3Fpreferred_auth_provider%3Dgoogle):

image

The buttons don't fit in one row even on max width. This happens because the page width is still 600px, the logo images were increased to 36px and they are allowed to take only half of the page width.

Now we can remember that 600px is an arbitrary page width (#4773, Page width). It was chosen likely because the standard width is too high for login pages. The other width we use is 760px on [the about page](https://www.openstreetmap.org/about). We can use it for login pages too in exactly the same manner, without special `.auth-container` classes:

image

Now the buttons fit.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -677,7 +677,13 @@ tr.turn {
   padding: $lineheight;
 }
 
-/* Overrides for pages that use new layout conventions */
+/* Rules for login and signup pages */
+
+.sessions-new, .users-new, .users-create {
+  #content .content-inner {
+    max-width: 760px;
+  }
+}
 
 .header-illustration {
   background-position: 0 0;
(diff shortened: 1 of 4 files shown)


tomhughes:
I'm certainly sympathetic to this - if I had noticed it when merging the increase in icon size then I would probably have questioned it but I don't have a full set of authentication providers in my dev instance so didn't see it.

I think my only question before merging it would be to ask if @stalker314314 has any comments on how and why the 600px width was chosen?

stalker314314:
We mostly did this based on design from our designer, this is where 600px comes from. To be fair, this is showing in popup screens (on sometimes already small displays in laptops), so we should be conservative with widths, but this change is making design better, for sure!

tomhughes:
Great - this looks good then and having less different widths seems like a win.

PR #4956, opened 2024-07-05
Improves displaying lat, lon values close to 0.0

### LINKED ISSUE #4950: Searches for latitudes or longitudes between 1 and -1 result in invalid E-notation numbers
### URL

https://www.openstreetmap.org/search?query=0.0000123%2C0.0000456

Search for a coordinate within 1 degree of the equator or prime meridian, such as very near Greenwich or Ecuador. Note that the resulting coordinates are displayed in computery syntax like "1.23e-05" instead of "0.0000123". This may cause confusion or errors (for example if the values are copy-pasted into a program expecting plain decimal coordinates, or if average users are unfamiliar with E notation.)

This may in fact be an upstream issue with Nominatim, etc, in which case hopefully we can fix it there as well. This issue was discovered when a user attempted to import or access a bookmark with E notation coordinates inside of Organic Maps: it's unclear which software generated the offending geo: URI, but my understanding of the RFC and coordinate systems is that such a rendering should never be presented to end users and is generally unsupported in parsers. Since we often validate odd input on mobile against the OSM website and other OSM ecosystem tools, it makes sense to try and squash this bug wherever it's found.

Example queries:

https://www.openstreetmap.org/search?query=45.0%2C0.0000456

https://www.openstreetmap.org/search?query=0.0000123%2C-75.456

_No response_

nenad-vujicic:
This PR addresses issue about using exponential instead of standard notation for latitudes and longitudes close to 0.0 in search results described in #4950

Fixes #4950. Improved result_to_html routine by detecting latitude and longitude values close to 0.0 and replacing their representation (exponential) with standard, by calculating number of required decimals (by looking at the position of decimal point if available) and generating standard representation.

Before:
image
After:
image

### DIFF
--- a/app/helpers/geocoder_helper.rb
+++ b/app/helpers/geocoder_helper.rb
@@ -2,6 +2,37 @@ module GeocoderHelper
   def result_to_html(result)
     html_options = { :class => "set_position stretched-link", :data => {} }
 
+    result.each do |key, value|
+      if key.to_s == "name" && value.match(%r{^([+-]?\d+(\.\d*)?)(([eE][-+]?\d+)?)(?:\s+|\s*[,/]\s*)([+-]?\d+(\.\d*)?)(([eE][-+]?\d+)?)$})
+        lat, lon = value.split(%r{[\/,]})
+
+        if lat.index(".").positive?
+          number_of_decimals = lat.length - lat.index(".") - 1
+          lat = format("%0.#{number_of_decimals}f", lat.to_f)
+        end
+
+        if lon.index(".").positive?
+          number_of_decimals = lon.length - lon.index(".") - 1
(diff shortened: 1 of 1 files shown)


tomhughes:
Thanks for the work but I'm not sure this is the best approach - it's better where we can to simply avoid the conversion to floating point and then back to string and to just preserve the original string.

I've opened #4961 with my preferred way of doing things.

mmd-osm:
@nenad-vujicic Are you happy to close this one, since we have an alternative implementation now?

nenad-vujicic:
> @nenad-vujicic Are you happy to close this one, since we have an alternative implementation now?

Yes, sure! @tomhughes , thank you very much for great solution :-)

Thanks,
Nenad.

PR #4957, opened 2024-07-05
Set a custom user agent when making http requests

tomhughes:


lib/nominatim.rb
```diff
@@ -11,14 +11,16 @@ def self.describe_location(lat, lon, zoom = nil, language = nil)
       url = "#{Settings.nominatim_url}reverse?lat=#{lat}&lon=#{lon}&zoom=#{zoom}&accept-language=#{language}"
 
       begin
-        response = Timeout.timeout(4) do
-          REXML::Document.new(Net::HTTP.get(URI.parse(url)))
+        response = OSM.http_client.get(URI.parse(url)) do |request|
+          request.timeout = 4
```
AntonKhorev, line 15: > undefined method `timeout=' for #<struct Faraday::Request ...

```suggestion
          request.options.timeout = 4
```

tomhughes: Thanks for the catch - now fixed.

AntonKhorev (approved):

PR #4958, opened 2024-07-06
Avoid calling json.partial! in loop, render collections instead

mmd-osm:
See discussion in #4141


tomhughes:
Looks good to me, thanks.

PR #4959, opened 2024-07-06
Remove "whereami" search parameter, use "lat" and "lon" instead

AntonKhorev:
`whereami` was introduced in #1968. It's only used to suppress another parameter `latlon_digits` which indicates that we're not sure if the query string is "lat,lon" or "lon,lat". `whereami=1` says that we know it's "lat,lon". But why don't we use `lat` and `lon` parameters directly?

### DIFF
--- a/app/assets/javascripts/index/contextmenu.js
+++ b/app/assets/javascripts/index/contextmenu.js
@@ -51,7 +51,7 @@ OSM.initializeContextMenu = function (map) {
           lat = latlng.lat.toFixed(precision),
           lng = latlng.lng.toFixed(precision);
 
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good to me, and a lot more sensible!

tomhughes:
This has the side affect that "whereami" no longer adds the coordinates to the search box as that box reflects the query parameter.

AntonKhorev:
"whereami" didn't fully work. Search input value partly relied on input autocompletion by the browser, partly on javascript, which is partly buggy, but you never see that because the buggy parts never run.

PR #4960, opened 2024-07-06
Don't call check_signup_allowed inside save_new_user

AntonKhorev:
Users controller has a `save_new_user` method. Its entire body is wrapped by `if check_signup_allowed(current_user.email)`.

There's only one place where `save_new_user` is called. It's also wrapped by `if check_signup_allowed(current_user.email)`:

https://github.com/openstreetmap/openstreetmap-website/blob/e3c919275818d46a515c183cf373d7a53bc6942a/app/controllers/users_controller.rb#L90-L105

Why do we need to call `check_signup_allowed` both inside and outside of `save_new_user`?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -262,40 +262,38 @@ class UsersController < ApplicationController
   private
 
   def save_new_user(email_hmac, referer = nil)
-    if check_signup_allowed(current_user.email)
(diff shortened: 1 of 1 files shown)


tomhughes:
I'm guessing there's some history which means it probably made sense originally but as you say it seems pointless now.

PR #4961, opened 2024-07-07
Avoid use of scientific notation in geocoder results

### LINKED ISSUE #4950: Searches for latitudes or longitudes between 1 and -1 result in invalid E-notation numbers

tomhughes:
This fixes #4950 by modifying the geocoder to avoid printing results in scientific notation.

The main fix is simply to avoid converting the user entered string to numeric format and back but for DMS style inputs we need to convert them to numeric form so we can do calculations so we use bigdecimal rather than float as it can force fixed precision format when converting to a string.

### DIFF
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -226,18 +226,18 @@ class GeocoderController < ApplicationController
 
   def to_decdeg(captures)
     ns = captures.fetch("ns").casecmp("s").zero? ? -1 : 1
-    nsd = captures.fetch("nsd", "0").to_f
-    nsm = captures.fetch("nsm", "0").to_f
-    nss = captures.fetch("nss", "0").to_f
+    nsd = BigDecimal(captures.fetch("nsd", "0"))
+    nsm = BigDecimal(captures.fetch("nsm", "0"))
+    nss = BigDecimal(captures.fetch("nss", "0"))
 
     ew = captures.fetch("ew").casecmp("w").zero? ? -1 : 1
-    ewd = captures.fetch("ewd", "0").to_f
-    ewm = captures.fetch("ewm", "0").to_f
-    ews = captures.fetch("ews", "0").to_f
+    ewd = BigDecimal(captures.fetch("ewd", "0"))
+    ewm = BigDecimal(captures.fetch("ewm", "0"))
+    ews = BigDecimal(captures.fetch("ews", "0"))
 
     lat = ns * (nsd + (nsm / 60) + (nss / 3600))
     lon = ew * (ewd + (ewm / 60) + (ews / 3600))
 
-    { :lat => lat, :lon => lon }
+    { :lat => lat.round(6).to_s("F"), :lon => lon.round(6).to_s("F") }
   end
 end
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -206,20 +206,16 @@ class GeocoderController < ApplicationController
     if query = params[:query]
       query.strip!
 
-      if latlon = query.match(/^([NS])\s*(\d{1,3}(\.\d*)?)\W*([EW])\s*(\d{1,3}(\.\d*)?)$/).try(:captures) || # [NSEW] decimal degrees
-                  query.match(/^(\d{1,3}(\.\d*)?)\s*([NS])\W*(\d{1,3}(\.\d*)?)\s*([EW])$/).try(:captures)    # decimal degrees [NSEW]
-        params.merge!(nsew_to_decdeg(latlon)).delete(:query)
+      if latlon = query.match(/^(?<ns>[NS])\s*(?<nsd>\d{1,3}(?:\.\d+)?)°?\W*(?<ew>[EW])\s*(?<ewd>\d{1,3}(?:\.\d+)?)°?$/) ||                                                                                                     # [NSEW] decimal degrees
+                  query.match(/^(?<nsd>\d{1,3}(?:\.\d+)?)°?\s*(?<ns>[NS])\W*(?<ewd>\d{1,3}(?:\.\d+)?)°?\s*(?<ew>[EW])$/) ||                                                                                                     # decimal degrees [NSEW]
+                  query.match(/^(?<ns>[NS])\s*(?<nsd>\d{1,3})°?(?:\s*(?<nsm>\d{1,3}(?:\.\d+)?)['′]?)\W*(?<ew>[EW])\s*(?<ewd>\d{1,3})°?(?:\s*(?<ewm>\d{1,3}(?:\.\d+)?)['′]?)$/) ||                                               # [NSEW] degrees, decimal minutes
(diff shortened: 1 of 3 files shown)


test/controllers/geocoder_controller_test.rb
```diff
@@ -368,17 +368,17 @@ def latlon_check(query, lat, lon)
     assert_template :layout => "map"
     assert_equal %w[latlon osm_nominatim_reverse], assigns(:sources).pluck(:name)
     assert_nil @controller.params[:query]
-    assert_in_delta lat, @controller.params[:lat]
-    assert_in_delta lon, @controller.params[:lon]
+    assert_in_delta lat, @controller.params[:lat].to_f
+    assert_in_delta lon, @controller.params[:lon].to_f
```
AntonKhorev, line 372:
```suggestion
    assert_match(/^[+-]?\d+(\.\d*)?$/, @controller.params[:lat])
    assert_match(/^[+-]?\d+(\.\d*)?$/, @controller.params[:lon])
    assert_in_delta lat, @controller.params[:lat].to_f
    assert_in_delta lon, @controller.params[:lon].to_f
```

test/controllers/geocoder_controller_test.rb
```diff
@@ -368,17 +368,17 @@ def latlon_check(query, lat, lon)
...
     get search_path(:query => query), :xhr => true
     assert_response :success
     assert_template :search
     assert_template :layout => "xhr"
     assert_equal %w[latlon osm_nominatim_reverse], assigns(:sources).pluck(:name)
     assert_nil @controller.params[:query]
-    assert_in_delta lat, @controller.params[:lat]
-    assert_in_delta lon, @controller.params[:lon]
+    assert_in_delta lat, @controller.params[:lat].to_f
+    assert_in_delta lon, @controller.params[:lon].to_f
```
AntonKhorev, line 381:
```suggestion
    assert_match(/^[+-]?\d+(\.\d*)?$/, @controller.params[:lat])
    assert_match(/^[+-]?\d+(\.\d*)?$/, @controller.params[:lon])
    assert_in_delta lat, @controller.params[:lat].to_f
    assert_in_delta lon, @controller.params[:lon].to_f
```

test/controllers/geocoder_controller_test.rb
AntonKhorev, line 1: Shouldn't we test near-zero latlon values?

```ruby
  ##
  # Test identification of lat/lon pairs with values close to zero
  def test_identify_latlon_close_to_zero
    [
      "0.00001 -0.00002",
      "+0.00001 -0.00002"
    ].each do |code|
      latlon_check code, 0.00001, -0.00002
    end
  end
```

test/controllers/geocoder_controller_test.rb
tomhughes: Nice idea. I've added that and also added a test case using small values to check it doesn't trigger scientific formatting.

app/controllers/geocoder_controller.rb
```diff
@@ -228,39 +224,20 @@ def normalize_params
...
-      lon = captures[9].casecmp("w").zero? ? -(captures[5].to_f + ((captures[6].to_f + (captures[7].to_f / 60)) / 60)) : captures[5].to_f + ((captures[6].to_f + (captures[7].to_f / 60)) / 60)
-    rescue StandardError
-      lat = captures[0].casecmp("s").zero? ? -(captures[1].to_f + ((captures[2].to_f + (captures[3].to_f / 60)) / 60)) : captures[1].to_f + ((captures[2].to_f + (captures[3].to_f / 60)) / 60)
-      lon = captures[5].casecmp("w").zero? ? -(captures[6].to_f + ((captures[7].to_f + (captures[8].to_f / 60)) / 60)) : captures[6].to_f + ((captures[7].to_f + (captures[8].to_f / 60)) / 60)
-    end
-    { :lat => lat, :lon => lon }
+    lat = ns * (nsd + (nsm / 60) + (nss / 3600))
+    lon = ew * (ewd + (ewm / 60) + (ews / 3600))
+
+    { :lat => lat.to_s("F"), :lon => lon.to_s("F") }
```
AntonKhorev, line 241: Need to `.truncate()` to some reasonable precision before `.to_s` or else the length blows up:

image

AntonKhorev (changes requested):

app/controllers/geocoder_controller.rb
tomhughes: Fixed... 
image

AntonKhorev (approved):

test/controllers/geocoder_controller_test.rb
```diff
@@ -223,6 +223,18 @@ def test_identify_latlon_sw_dms
...
+  def test_identify_latlon_close_to_zero
+    [
+      "0.0000123 -0.0000456",
+      "+0.0000123 -0.0000456",
+      "N 0° 0' 0.4428\", W 0° 0' 1.6416\""
+    ].each do |code|
+      latlon_check code, 0.0000123, -0.0000456
+    end
+  end
+
```
AntonKhorev, line 237: This one is going to fail with `TypeError: can't convert nil into BigDecimal`

```suggestion

  ##
  # Test identification of integer lat/lon pairs using N/E with degrees
  def test_identify_latlon_ne_d_int_deg
    [
      "N50° E14°"
    ].each do |code|
      latlon_check code, 50, 14
    end
  end
```

[Used to work correctly](https://www.openstreetmap.org/search?query=N50%C2%B0%20E14%C2%B0)

AntonKhorev (changes requested):

test/controllers/geocoder_controller_test.rb
tomhughes: Ha that's effectively an old lurking bug - the first two regexps that are supposed to match the degrees only case but didn't allow the optional degree symbol while the rest of the regexps wrongly made everything after degrees optional so would over match.

All fixed now hopefuly...

test/controllers/geocoder_controller_test.rb
```diff
@@ -60,6 +60,19 @@ def test_identify_latlon_ne_d
...
+    [
+      "N50 E14",
+      "N50° E14°",
+      "50N 14E",
+      "50°N 14°E"
+    ].each do |code|
+      latlon_check code, 50, 14
+    end
+  end
+
```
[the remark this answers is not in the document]
AntonKhorev: If we don't want to identify them, we can change `\.\d*` to `\.\d+` in regexps.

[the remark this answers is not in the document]
tomhughes: Fixed that now...

AntonKhorev:
hopefully those were all the bugs, merged

AntonKhorev:
and I was wrong, there's more

PR #4962, opened 2024-07-08
Add gap between rows of auth buttons

AntonKhorev:
Currently auth buttons have horizontal margins that create horizontal gaps and nothing to create vertical gaps. It's easier to set the gap class on their flex container and get gaps in both directions.

Before/after:
image image


tomhughes:
Looks good to me, thanks.

PR #4963, opened 2024-07-08
Remove unnecessary classes from preferred auth button

AntonKhorev:
.rounded is the default rounding for this button type.
.fs-6 is the default font size for this button type.

image


tomhughes:
Looks good to me, thanks.

PR #4968, opened 2024-07-09
Remove former .auth-container wrapper divs

AntonKhorev:
They are unnecessary after #4954.

One of them had unused id `login_login`.


tomhughes:
Looks good to me, thanks.

PR #4969, opened 2024-07-09
Move users list to administrator module

AntonKhorev:
I wanted to rename `user_new_path`, it's standard name should be `new_user_path`. I tried using resourceful routes but we have `create` action mapped to `POST` `/user/new` which is also nonstandard. Then I'd have to do both `new` and `create` actions with resourceful routes and this will want to rewrite `users_path`. `users_path` is currently mapped to admin users list. I have to rename it first which is easier to do if I move it to a different module.


AntonKhorev:
After back-and-forth in #4587 a different module scheme was chosen, so this probably needs updating.

AntonKhorev:
Superseded by #5391

PR #4971, opened 2024-07-10
Update OpenID input label

AntonKhorev:
- Get rid of html locale string `openid_html: "%{logo} OpenID"` just to prepend the logo image.
- Make the label read "OpenID URL", which includes removing the logo alt text inside the label. In this case the logo is just a visual cue, you don't want screen readers to read its alt text and then read the actual label. I added "URL" because it's present on [the livejournal login page](https://www.livejournal.com/identity/login.bml), where OpenID originated.

image

### DIFF
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -52,10 +52,6 @@ module UserHelper
 
   # External authentication support
 
-  def openid_logo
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me, thanks.

PR #4972, opened 2024-07-10
Change OpenID provider link to button

AntonKhorev:
image

It's not a link. It doesn't have a proper href, that's why `#` was used.


tomhughes:
Looks good to me, thanks.

Issue #1886, opened 2018-05-30
When downloading image by "Share", allow other map styles with static image APIs


pnorman:
@gravitystorm's styles have a [static maps API](https://www.thunderforest.com/docs/static-maps-api/). Currently we disable export as image for layers other than "Standard", but instead we should use their APIs if they are available.

PR #4973, opened 2024-07-10
When downloading image by "Share", allow other map styles with static image APIs

nertc:
This PR addresses "When downloading image by 'Share', allow other map styles with static image APIs" issue mentioned in the https://github.com/openstreetmap/openstreetmap-website/issues/1886

Added API for downloading Transport and Cycle Map images, as they are supported by the [ThunderForest](https://www.thunderforest.com/docs/static-maps-api/). Also changed unsupported map text to show that not only Standard, but Transport and Cycle Map layers also have downloading capability. Fixes https://github.com/openstreetmap/openstreetmap-website/issues/1886

Transport layer:
Screenshot 2024-07-10 140949

Cycle Map layer:
Screenshot 2024-07-10 141009

Unsupported Humanitarian layer:
Screenshot 2024-07-10 141024

### DIFF
--- a/app/assets/javascripts/leaflet.share.js
+++ b/app/assets/javascripts/leaflet.share.js
@@ -160,6 +160,7 @@ L.OSM.share = function (options) {
     $("<div>")
       .appendTo($form)
       .attr("class", "row mb-3")
+      .attr("id", "mapnik_scale_row")
       .append($("<label>")
         .attr("for", "mapnik_scale")
         .attr("class", "col-auto col-form-label")
@@ -195,7 +196,7 @@ L.OSM.share = function (options) {
             .attr("class", "form-check-input")
             .bind("change", toggleFilter))));
 
-    ["minlon", "minlat", "maxlon", "maxlat"].forEach(function (name) {
+    ["minlon", "minlat", "maxlon", "maxlat", "lat", "lon"].forEach(function (name) {
       $("<input>")
         .attr("id", "mapnik_" + name)
         .attr("name", name)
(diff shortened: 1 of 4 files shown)


tomhughes:
Have we confirmed with @gravitystorm that he's happy for us to do this? We have to take pretty extreme measures to protect our infrastructure from abuse of this facility so load may be a concern but I don't know exactly how his static map endpoint works.

nertc:
Hello @gravitystorm, I am interested what do you think about this issue. When user will download an image with Cycle Map or Transport layer, it will call `https://tile.thunderforest.com/static/{style}/{lon},{lat},{zoom}/{width}x{height}{scale}.{format}?apikey={apikey}` as indicated on [Thunderfores - Static Maps API](https://www.thunderforest.com/docs/static-maps-api/). Image generation requires one api call. ApiKey is the same as in the tiles request. In terms of security, API key is shown as query parameter both in tiles and image generation requests.

pnorman:
We obviously should get Andy's okay, but a vector tile based static map endpoint is at least a couple of orders of magnitude more efficient than ours which does DB queries on demand.

gravitystorm:
> Hello @gravitystorm, I am interested what do you think about this issue.

I'm happy on behalf of Thunderforest for this to be implemented. As @pnorman says, it's much less resource intensive on our infrastructure than the approach used by OSMF for the standard map layer. And if there's any unforeseen problems, then I'm happy to revisit this later.

(As a maintainer here, I haven't reviewed the PR!)

tomhughes (changes requested):
There's a message (from the `javascripts.share.image_dimensions` key) that appears before the download button and which currently says the image will be of the standard layer which is no longer true.

I think that dates from before we started hiding image export on other layers and the reference to the layer can probably be removed now.

nertc:
@tomhughes Thanks for the review. I changed the key so it will show `Image will show %{layer} layer at %{width} x %{height}` appropriate layer.

Transport Map layer example:
Image will show Transport Map layer at 1522 x 1464
Screenshot 2024-07-15 192120

tomhughes (approved):
Thanks - that looks good now.

PR #4974, opened 2024-07-10
Test Nominatim.describe_location

AntonKhorev:
Tests for #4957

### DIFF
--- a/test/lib/nominatim_test.rb
+++ b/test/lib/nominatim_test.rb
@@ -1,0 +1,31 @@
+require "test_helper"
+
+class NominatimTest < ActiveSupport::TestCase
+  def test_describe_location
+    stub_request(:get, %r{^https://nominatim\.example\.com/reverse\?})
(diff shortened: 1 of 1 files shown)


AntonKhorev:
`with_settings(:server_url ...)` seems to be ignored by Tests / Ubuntu but not by Docker

tomhughes:
Thanks - looks good and I verified that it catches the mistake I made in my PR the other day ;-)

AntonKhorev:
But it doesn't work correctly on github. It sends 'User-Agent'=>'test.host' instead of what I'm trying to substitute.

tomhughes:
Sorry I misunderstood what you were saying... I thought you had worked around it.

The problem is that `OSM.http_client` only creates the client once, the first time it is called, so what user agent is used depends on what order the tests run in.

I think we can just make the test check that the user agent is `test.host` instead of trying to change it.

tomhughes:
Made that change in 63f0b9257d62818d435ef6510340b9681efe547b.

PR #4849, opened 2024-05-28
Fixed "auth_button_preferred alignment" issue described in #4773

nenad-vujicic:
This PR addresses "auth_button_prefered alignment" issue mentioned in the #4773

In app/views/application/_auth_providers_html.erb, updated div's styles and updated layout (parent div + 2 (or 1) child div area). Here are the screenshots before and after the changes:

Before:
image
image

After:
image
image

PR #4976, opened 2024-07-11
Stack preferred auth button over other auth buttons on small screens

AntonKhorev:
There's an attempt to wrap auth buttons below the preferred button with `flex-wrap` here:
https://github.com/openstreetmap/openstreetmap-website/blob/bc8ce5729da0a7df4306ae5f785f68980a9bf47b/app/views/application/_auth_providers.html.erb#L2

That's not going to work after the children of this flex container got 50% width in #4849. They always fit in one row, no matter how narrow 50% of that row is.

Here I'm using Bootstrap's grid system with two-column row above the small width breakpoint if there's a preferred auth button.

Before/after:
image image


tomhughes:
Looks good to me, thanks.

PR #4977, opened 2024-07-12
Show/hide entire OpenID form

AntonKhorev:
When the login page is loaded, this form gets hidden by javascript:
image

Later if the OpenID button is clicked, it gets shown.

Actually not quite. The form is always visible, what's getting shown/hidden are the div containing the url input and the *Continue* button. But there's not reason to show/hide them separately when you can show/hide the entire form.


tomhughes:
Looks good to me, thanks.

PR #4978, opened 2024-07-12
Move diary comment action to comments controller

AntonKhorev:
Continues #4905.

Supersedes #4634.


tomhughes:
Looks good to me, thanks.

PR #4982, opened 2024-07-13
Use casecmp?() instead of casecmp().zero?

AntonKhorev:


tomhughes:
Looks good to me, thanks.

PR #4983, opened 2024-07-13
Check that fractional part is present after . in geocoder latlon tests

AntonKhorev:
Since we decided that numbers can't end with `.` and changed the geocoder regexps, it makes sense to change the test regexps too.


tomhughes:
Looks good to me, thanks.

PR #4984, opened 2024-07-13
Parse lat and lon independently when using dms notation

AntonKhorev:
Current parsing tries to have as many numbers in longitude as there are in longitude, sometimes splitting "12" (12 deg) into (1 deg 2 min).

Before/after:
image image


tomhughes:
Looks good to me, thanks.

Issue #1391, opened 2016-12-14
When a user posts a story in their diary, give them the option to share with social media


openbrian:
Give them the option to share with Twitter, Facebook, Instagram, Jabber (XMPP) etc.

hd719:
Hello can I work on this?

pnorman:
It's open source, so you can work on any issues you want to. This one is probably a reasonably good starter issue because it's a small front-end change, and it's unlikely anyone else is working on it.

tomhughes:
I wouldn't say it was small myself. I mean it won't need that much code but it will presumably mean dragging in some third party component(s) and hence making a choice about which is best to use. There will also be significant privacy issues to deal with.

By all means do have a go at it - nobody else is as far as I know - but I wouldn't say it's trivial.

pnorman:
I don't see significant privacy issues, maybe we're imagining different implementations? For the twitter case it'd be a link like `https://twitter.com/intent/tweet?original_referer=http%3A%2F%2Fpaulnorman.ca%2Fblog%2F2016%2F11%2Fserving-vector-tiles%2F&ref_src=twsrc%5Etfw&text=Serving%20Vector%20Tiles%20-%20Paul%E2%80%99s%20Blog&tw_p=tweetbutton&url=http%3A%2F%2Fwww.paulnorman.ca%2Fblog%2F2016%2F11%2Fserving-vector-tiles%2F`

e.g. 
![image](https://cloud.githubusercontent.com/assets/1190866/25608237/fe561dca-2ece-11e7-989c-c6f891d567a1.png)

There probably is some third-party component that URL formation for common social media networks

tomhughes:
Well that's the ultra low tech approach for twitter sure, but I think a more normal approach would be to use one of the gems I'm sure exists that adds some sort of sharing panel for a range of sites - the important thing would be to ensure that it didn't load anything from the sites until you chose to use it.

hd719:
Hey @tomhughes and @pnorman I tried running `bundle exec rake db:migrate` and I ran into this error. How would I fix this? 

```
Hamels-MacBook-Air:openstreetmap-website HD$ bundle exec rake db:migrate
svgo worker: `svgo` not found; please provide proper binary or disable this worker (--no-svg
o argument or `:svgo => false` through options)
```

-Thank you.

hd719:
[...]

Issue #739, opened 2014-05-07
Facebook share image


talllguy:
I was discussing this with @mvexel [the other day on twitter](https://twitter.com/talllguy/status/462703000096489472). Sharing a static map URL looks like this on Facebook:
![screenshot 2014-05-03 17 16 43](https://cloud.githubusercontent.com/assets/1827881/2906142/6a31f834-d60d-11e3-833a-37266e8594ea.png)
For a richer experience, it would be great if that included a map preview, which can be accomplished by adding an `og:image` tag (see http://ogp.me/) in the html. Facebook caches the `og:image` so a truly dynamic preview isn't realistic. However, we could designate a particular well mapped area as the preview image, e.g. Berlin, or the full extent.

jidanni:
Facebook still displays generic OSM logo istead of the real map.
Debug with https://developers.facebook.com/tools/debug/

tomhughes:
When did we ever claim otherwise?

1ec5:
As a stopgap until a rendered tile is possible, how about reusing the SVG images from https://github.com/gravitystorm/openstreetmap-carto/tree/master/symbols? Ideally the images from https://github.com/openstreetmap/openstreetmap-website/tree/master/app/assets/images/browse would be presented instead, but they’re still in PNG format and won’t scale well beyond 16×16 pixels. `og:image:width` and `og:image:height` could be used to scale the icon down in case a life-size [hamburger](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/fast_food.16.svg) is too much.

stefanb:
Couldn't the  opengraph `og:image` tag URL point to the export API, eg http://render.openstreetmap.org/cgi-bin/export?bbox=14.414062499999998,46.00280468514235,14.558258056640623,46.106922264575914&scale=48033&format=png

I do realize this could be a performance issue, but nothing that some caching couldn't resolve. (Facebook likely caches the images themselves)

tomhughes:
Given that the export API frequently doesn't work for large parts of the day that seems like an extremely bad idea.

stefanb:
Oh, that I wasn't aware of. Thanks for sharing.

openbrian:
+1

pnorman:
> However, we could designate a particular well mapped area as the preview image, e.g. Berlin, or the full extent

I see three options for this issue. 

1. generate some static image for the `og:image` tag on pages where the map is being browsed;
[...]

PR #4985, opened 2024-07-13
Add social sharing functionality

kcne:
### Add social network sharing to diary entry pages

**Description:**
This PR addresses the issue #1391 by introducing social media sharing functionality for diary entries, allowing users to share posts on platforms like Twitter, Facebook, LinkedIn, Mastodon, Telegram, and Email. It includes required JavaScript to handle opening share URLs in pop-up windows and Ruby helper methods to filter and validate allowed social media sites. The changes also incorporate error logging for invalid sites and missing icons, ensuring robustness.

**Implementation details:**
- `lib/social_share_button_helper.rb` - config, validation & helper functions
- `render_social_share_buttons` function inside `application_helper.rb` responsible for rendering the buttons with helper utilization
- `social_share_button.js` - javascript utility functions for opening the popups, generating urls for sharing & handling onclick events on buttons

**Screenshots:**
Screenshot 2024-07-17 at 16 05 18
Screenshot 2024-07-17 at 16 07 12

Reviews welcome,
Thank you.


mmd-osm:
Thank you. Some of the links seem to be a bit dated, e.g. Twitter is now known as X, and links/logos would have to be updated, I suppose.

kcne:
Updated the x icon and link as suggested by @mmd-osm. Also if you think we should include some other socials too, suggestions are welcome. Any chance @tomhughes @gravitystorm @AntonKhorev some of you guys can review this soon?
Thank you.

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,45 @@ def render_flash(flash)
...
+        :via => opts[:via]
+      }
+    ) do
+      social_buttons = valid_sites.filter_map do |site|
+        if SocialShareButtonHelper.icon_path(site).empty?
+          Rails.logger.error("Invalid site: #{site}")
+          next
+        end
+
+        link_to "", {
```
AntonKhorev, line 107: Why do links to social media have to be generated by javascript instead of being available here?

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,45 @@ def render_flash(flash)
...
+
+    valid_sites, invalid_sites = SocialShareButtonHelper.filter_allowed_sites(sites)
+
+    # Log invalid providers
+    invalid_sites.each do |invalid_site|
+      Rails.logger.error("Invalid site or icon not configured: #{invalid_site}")
+    end
+
+    content_tag(
+      :div,
```
AntonKhorev, line 92:
```suggestion
    tag.div(
```

app/views/diary_entries/_diary_entry.html.erb
```diff
@@ -1,3 +1,7 @@
+<% content_for :head do %>
+  <%= javascript_include_tag "social_share_button" %>
```
AntonKhorev, line 2: You'll have as many `<script>` tags as there are articles and a bunch of js errors about redeclarations.

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,45 @@ def render_flash(flash)
...
+        if SocialShareButtonHelper.icon_path(site).empty?
+          Rails.logger.error("Invalid site: #{site}")
+          next
+        end
+
+        link_to "", {
+          :rel => ["nofollow", opts[:rel]].compact,
+          :data => { :site => site, :title => opts[:title], :desc => opts[:desc], :url => opts[:url] },
+          :class => "social-share-button ssb-icon",
+          :onclick => "return SocialShareButton.share(this);"
```
AntonKhorev, line 111: `onclick` is probably some leftover because it refers to `SocialShareButton` which doesn't exist

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,45 @@ def render_flash(flash)
...
+      :div,
+      :class => "social-share-button d-flex gap-1 align-items-end flex-wrap",
+      :data => {
+        :title => opts[:title],
+        :url => opts[:url],
+        :desc => opts[:desc],
+        :via => opts[:via]
+      }
+    ) do
+      social_buttons = valid_sites.filter_map do |site|
```
AntonKhorev, line 101: Why assigning to `social_buttons` if you return this value right away?

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,45 @@ def render_flash(flash)
...
+          Rails.logger.error("Invalid site: #{site}")
+          next
+        end
+
+        link_to "", {
+          :rel => ["nofollow", opts[:rel]].compact,
+          :data => { :site => site, :title => opts[:title], :desc => opts[:desc], :url => opts[:url] },
+          :class => "social-share-button ssb-icon",
+          :onclick => "return SocialShareButton.share(this);"
+        } do
```
AntonKhorev, line 112: There probably should also be a `title` attribute with "Share on (platform)" value.

app/assets/javascripts/social_share_button.js
```diff
@@ -0,0 +1,59 @@
...
+  const height = Math.max(100, Math.min(screen.height, initialHeight));
+
+  const left = (screen.width / 2) - (width / 2);
+  const top = (screen.height * 0.3) - (height / 2);
+  const opts = `width=${width},height=${height},left=${left},top=${top},menubar=no,status=no,location=no`;
+
+  const newWindow = window.open(url, "popup", opts);
+
+  if (!newWindow || newWindow.closed || typeof newWindow.closed == "undefined") {
+    console.error("Popup blocked. Please allow popups for this website.");
```
AntonKhorev, line 18: Who is supposed to be able to read this message? Regular users don't look at their browser console.

app/helpers/application_helper.rb
kcne: Good point, just refactored according to this

app/helpers/application_helper.rb
kcne: ✔️

app/helpers/application_helper.rb
kcne: True that. Just removed it completely.

app/assets/javascripts/social_share_button.js
kcne: Removed this.

app/helpers/application_helper.rb
kcne: Just a habit thing, now it's gone.

app/helpers/application_helper.rb
kcne: Makes sense, so I added it.

talllguy:
Thanks for the hard work on this! Since we're on the topic of social sharing I thought I'd resurface an idea from 10 years ago that I still think would add value to users of the site. Perhaps the technology hurdles back then are less likely to be hurdles now. https://github.com/openstreetmap/openstreetmap-website/issues/739

app/views/diary_entries/_diary_entry.html.erb
kcne: Added this into `_head.html.erb`

kcne:
Refactored this according to @AntonKhorev review. Please let me know If you think I can further improve this. Thank you!

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,29 @@ def render_flash(flash)
...
+    end
+
+    tag.div(
+      :class => "social-share-button d-flex gap-1 align-items-end flex-wrap"
+    ) do
+      valid_sites.map do |site|
+        link_to SocialShareButtonHelper.generate_share_url(site, opts), {
+          :rel => ["nofollow", opts[:rel]].compact,
+          :class => "ssb-icon",
+          :title => "Share on #{site.capitalize}"
```
AntonKhorev, line 97: This needs to be translated.

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,29 @@ def render_flash(flash)
...
+        } do
+          image_tag(SocialShareButtonHelper.icon_path(site), :alt => site.capitalize, :size => 28)
```
AntonKhorev, line 99: `:alt => site.capitalize`: "email" is not a proper site name, you'll have to translate it instead of just capitalizing. 

You'll have something like:
https://github.com/openstreetmap/openstreetmap-website/blob/1b288d40c7fff3d3be8d82d4e96c9008778f4dba/config/locales/en.yml#L2597-L2614

app/views/diary_entries/_diary_entry.html.erb
```diff
@@ -44,4 +44,10 @@
       <% end %>
     </ul>
   </nav>
+
```
AntonKhorev, line 47: The buttons ate too close to the actions list above. Probably need to add margins to it.
image

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,29 @@ def render_flash(flash)
...
+      Rails.logger.error("Invalid site or icon not configured: #{invalid_site}")
+    end
+
+    tag.div(
+      :class => "social-share-button d-flex gap-1 align-items-end flex-wrap"
+    ) do
+      valid_sites.map do |site|
+        link_to SocialShareButtonHelper.generate_share_url(site, opts), {
+          :rel => ["nofollow", opts[:rel]].compact,
+          :class => "ssb-icon",
```
AntonKhorev, line 96: If all icons are going to be circular, you can add the `rounded-circle` class to get circular focus outlines:
image

lib/social_share_button_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def self.valid_site?(site)
+    SOCIAL_SHARE_CONFIG.key?(site.to_sym)
+  end
+
+  def self.generate_share_url(site, params)
+    site = site.to_sym
+    case site
+    when :email
+      "mailto:?subject=#{URI.encode_www_form_component(params[:title])}&body=#{URI.encode_www_form_component(params[:url])}"
```
AntonKhorev, line 31: `URI.encode_www_form_component` is a bit different from javascript's `encodeURIComponent`. It encodes spaces as `+`. But [mailto rfc](https://datatracker.ietf.org/doc/html/rfc6068#section-5) says spaces should be encoded as `%20`.

tomhughes (changes requested):
Added a few comments inline... Other things to consider are whether to offer some other sites - bluesky and threads being the obvious candidates.

app/assets/javascripts/social_share_button.js
```diff
@@ -0,0 +1,27 @@
+// Opening pop-ups with share URL
+function openShareUrl(url, initialWidth = 640, initialHeight = 480) {
+  if (typeof url !== "string" || !url.startsWith("http")) {
+    console.error("Invalid URL"); // Consider removing this line if console warnings should be avoided.
```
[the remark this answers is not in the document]
kcne: In current state it is not really possible, since all the urls are hardcoded.

lib/social_share_button_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+    when :x
+      via_str = params[:via] ? "&via=#{URI.encode_www_form_component(params[:via])}" : ""
+      hashtags_str = params[:hashtags] ? "&hashtags=#{URI.encode_www_form_component(params[:hashtags].join(','))}" : ""
+      "https://x.com/intent/tweet?url=#{URI.encode_www_form_component(params[:url])}&text=#{URI.encode_www_form_component(params[:title])}#{hashtags_str}#{via_str}"
+    when :linkedin
+      "https://www.linkedin.com/shareArticle?mini=true&url=#{URI.encode_www_form_component(params[:url])}"
+    when :facebook
+      "https://www.facebook.com/sharer/sharer.php?url=#{URI.encode_www_form_component(params[:url])}"
+    when :mastodon
+      "https://mastodon.social/share?text=#{URI.encode_www_form_component(params[:title])}&url=#{URI.encode_www_form_component(params[:url])}"
```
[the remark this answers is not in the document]
kcne: Good catch but I'm a little stuck here, there is n numbers of mastodon instances. Do you have some recommendation on how to approach this?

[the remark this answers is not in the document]
kcne: Maybe use [mapstodon.space](mapstodon.space)?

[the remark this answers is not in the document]
mmd-osm: A lot of folks hang out on en.osm.town. Maybe check out https://github.com/Juerd/tootpick?tab=readme-ov-file which offers a popup to enter the mastodon instance (might be include a few better known ones as preselect?).

kcne:
Thank you for the review!
I've resolved the requested changes, added the translations as requested, and the mailto links should now be opening correctly. Once we round this up and resolve the mastodon thing with @tomhughes  I can add bluesky and threads support as a separate commit if you agree.

lib/social_share_button_helper.rb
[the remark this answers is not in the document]
kcne: Found out there is [mastodonshare.com](https://mastodonshare.com) that takes text and url query parameters and let's you choose the mastodon instance you want to share on. Updated the code accordingly. Can you please test this and let me know if this solution is okay. @mmd-osm @tomhughes

kcne:
> Found out there is [mastodonshare.com](https://mastodonshare.com) that takes text and url query parameters and let's you choose the mastodon instance you want to share on. Updated the code accordingly. Can you please test this and let me know if this solution is okay. @mmd-osm @tomhughes

Updated mastodon link, everything should be working fine now. Can you please check and let me know if anything else can be improved. Thank you.

app/assets/javascripts/social_share_button.js
```diff
@@ -0,0 +1,27 @@
...
+
+  const left = (screen.width / 2) - (width / 2);
+  const top = (screen.height * 0.3) - (height / 2);
+  const opts = `width=${width},height=${height},left=${left},top=${top},menubar=no,status=no,location=no`;
+
+  window.open(url, "popup", opts);
+}
+
+$(document).ready(function () {
+  $(".ssb-icon").on("click", function (e) {
```
AntonKhorev, line 19: This is not going to work with Turbo paging because `$(document).ready` won't happen when you click *Newer/Older Entries*. This can be fixed by delegating the event to something that stays on the page or the entire document, but the other question is: Should there be share buttons on multiple entry pages at all? We're also trying to truncate the entries on such pages, with #5121 or something like. You made your own version in #5125 and I don't see share buttons in it. Maybe the buttons should only be present on single entry pages?

app/assets/javascripts/social_share_button.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  const width = Math.max(100, Math.min(screen.width, initialWidth));
+  const height = Math.max(100, Math.min(screen.height, initialHeight));
+
+  const left = (screen.width / 2) - (width / 2);
+  const top = (screen.height * 0.3) - (height / 2);
+  const opts = `width=${width},height=${height},left=${left},top=${top},menubar=no,status=no,location=no`;
+
+  window.open(url, "popup", opts);
```
AntonKhorev, line 15: Email link only works as expected for me when it's *not* opened with `window.open`.

It works when middle-clicked, or when the event handler is broken because of Turbo + `$(document).ready`.

app/assets/javascripts/social_share_button.js
kcne: I agree on this one. While working on this, there was not too much information indicating someone is going to do such thing. If we are going to truncate the text, then this definitely should only be displayed on single entry pages. Although I'm not sure should I do it right away or wait for #5121 to be merged. Thank you.

AntonKhorev: It's already not working correctly on index pages so you'll have to do something. Removing the buttons is easier than fixing them, right?

kcne:
As @AntonKhorev suggested, we're now displaying these icons only when users open the full diary entry, as outlined in #5121. Since this is currently used in one place, turbo pagination issues can be handled in a future PR if a use case arises for combining it with turbo pagination. Thank you.

app/assets/javascripts/social_share_button.js
AntonKhorev: Still opens a window for `mailto:` links. What do you expect to appear in that window?

kcne: Expected behaviour is to open mailto link in new tab where you can select your mail app/provider, which should be working for the timebeing. Do you suggest we should handle mail sharing differently?

AntonKhorev: But you're also opening a new window that's not going to contain anything and will have to be closed manually. Or is it working differently for you?

kcne: Actually for me it works like this - I'm using firefox, hadn't check for the other browsers.

When I click share it opens share link in pop up window to select the email provider:
Screenshot 2024-10-22 at 10 07 59
After email provider selection, i get the automatically formatted email message ready to be sent:
Screenshot 2024-10-22 at 10 08 23

AntonKhorev: Here's how it looks for me:
image

I get a blank window in background and a new tab.

kcne: In this case the email dialog gets opened right away. Fixed now, thanks.

AntonKhorev (approved):
It's fine if we're ok with mastodonshare.com for now, but we can replace it later.

app/views/diary_entries/show.html.erb
```diff
@@ -1,3 +1,5 @@
+ <%= javascript_include_tag "social_share_button" %>
```
AntonKhorev, line 1:
```suggestion
<% content_for :head do %>
  <%= javascript_include_tag "social_share_button" %>
<% end %>
```

That's how we do it, for example, in https://github.com/openstreetmap/openstreetmap-website/blob/fe81ac334cfef04751a959297dca24e472cb025f/app/views/dashboards/show.html.erb#L13-L15

app/views/diary_entries/_diary_entry.html.erb
```diff
@@ -44,4 +44,5 @@
...
```
AntonKhorev, line 47: We probably don't need to touch this file?

AntonKhorev (changes requested):

app/views/diary_entries/show.html.erb
kcne: Thanks for the tip. Updated now.

app/views/diary_entries/_diary_entry.html.erb
kcne: You're right. I don't know how I missed this. I've restored the changes to this file and updated the commit. Thank you.

app/views/diary_entries/show.html.erb
AntonKhorev: But it's not updated.

kcne: Updated now, forgot to push, sorry.

AntonKhorev:
Merged, thanks.

Although we're effectively outsourcing Mastodon host option to mastodonshare.com, it works for anonymous users too. Currently our preferences page is not available for anonymous users but we probably should make it available and save preferences to cookies. We have other preferences like #5201 and color mode settings and we shouldn't add a button for each one.

app/helpers/application_helper.rb
```diff
@@ -75,4 +76,32 @@ def render_flash(flash)
   rescue StandardError
     flash.inspect if Rails.env.development?
   end
+
+  # Generates a set of social share buttons based on the specified options.
+  def render_social_share_buttons(opts = {})
+    sites = opts.fetch(:allow_sites, [])
```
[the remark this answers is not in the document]
AntonKhorev: We can probably use it if sharing on some site requires some setting, to filter out that site if it's not set, like if we add Mastodon host preference.

gravitystorm:
> anonymous users

I think you mean users who are not logged in - "anonymous users" is the term we use for accounts with `user.data_public == false`. They are quite rare now since [they haven't been permitted for a long time](https://wiki.openstreetmap.org/wiki/Anonymous_edits).

As for preferences, I'm happy keeping most preferences available for only logged in users, since account creation is free and we aim to serve our mappers (who by definition have accounts) more than just casual visitors. Storing preferences in cookies does have drawbacks (like cross-device consistency, automatic cookie expiry etc) so it's not my first choice in general.

AntonKhorev:
> I think you mean users who are not logged in

yes

> As for preferences, I'm happy keeping most preferences available for only logged in users, since account creation is free and we aim to serve our mappers (who by definition have accounts) more than just casual visitors. Storing preferences in cookies does have drawbacks (like cross-device consistency, automatic cookie expiry etc) so it's not my first choice in general.

Are you against the entire idea of #1115 or just against the cookies and want preferences stored somewhere else like session? I don't think anyone who is not logged in expects cross-device consistency etc.

PR #4988, opened 2024-07-15
Added workflow for PR labeling using Danger

nenad-vujicic:
This PR proposes possible solution for labeling pull requests using Danger described in #4912

Proposed solution introduces new workflow, executed whenever new PR is created or updated. Workflow checks if PR is too big (>500 changed lines) or changes one of translation files (YML, except en.yml) or contains merge-commits and generates appropriate labels using Danger and danger-auto_label plugin. If none of these is found, green "Good PR" is generated.

image

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -16,16 +16,16 @@ end
 
 # Report if some translation file (except en.yml) is modified
 if modified_yml_files.empty?
-  auto_label.remove("Compromised Translations")
+  auto_label.remove("Inappropriate Translations")
 else
   modified_files_str = modified_yml_files.map { |file| "`#{file}`" }.join(", ")
   warn("The following YAML files other than `en.yml` have been modified: #{modified_files_str}. Only `en.yml` is allowed to be changed. Translations are updated via Translatewiki, see CONTRIBUTING.md.")
-  auto_label.set(pr_number, "Compromised Translations", "B60205")
+  auto_label.set(pr_number, "Inappropriate Translations", "B60205")
 end
 
 # Report if there are merge-commits in PR
 if git.commits.any? { |c| c.parents.count > 1 }
-  warn("Merge commits are found in PR. Please rebase to get rid of the merge commits in this PR and read CONTRIBUTE.md.")
+  warn("Merge commits are found in PR. Please rebase to get rid of the merge commits in this PR, see CONTRIBUTING.md.")
   auto_label.set(pr_number, "Merge Commits", "D93F0B")
 else
   auto_label.remove("Merge Commits")
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,0 +1,32 @@
+# Get PR number
+pr_number = github.pr_json["number"]
+
+# Report if number of changed lines is > 500
+if git.lines_of_code > 500
+  warn("Number of updated lines of code is too large to be in one PR. Perhaps it should be separated into two or more?")
+  auto_label.set(pr_number, "Big PR", "FBCA04")
+else
+  auto_label.remove("Big PR")
+end
+
+# Get list of translation files (except en.yml) which are modified
+modified_yml_files = git.modified_files.select do |file|
+  file.start_with?("config/locales") && File.extname(file) == ".yml" && File.basename(file) != "en.yml"
+end
+
+# Report if some translation file (except en.yml) is modified
+if modified_yml_files.empty?
+  auto_label.remove("Compromised Translations")
+else
+  modified_files_str = modified_yml_files.map { |file| "`#{file}`" }.join(", ")
+  warn("The following YAML files other than `en.yml` have been modified: #{modified_files_str}. Only `en.yml` is allowed to be changed. Translations are updated via Translatewiki, see CONTRIBUTING.md.")
+  auto_label.set(pr_number, "Compromised Translations", "B60205")
+end
+
+# Report if there are merge-commits in PR
+if git.commits.any? { |c| c.parents.count > 1 }
+  warn("Merge commits are found in PR. Please rebase to get rid of the merge commits in this PR and read CONTRIBUTE.md.")
+  auto_label.set(pr_number, "Merge Commits", "D93F0B")
+else
+  auto_label.remove("Merge Commits")
+end
--- a/.github/workflows/labeling.yml
+++ b/.github/workflows/labeling.yml
@@ -1,0 +1,25 @@
+name: Danger
+
+run-name: Updating labels for current PR
+
+on:
+  pull_request_target:
+    types: [opened, synchronize]
+
+jobs:
+  danger:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Setup ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.1
+          rubygems: 3.4.10
+          bundler-cache: true
+      - name: Danger
+        env:
+          DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          bundle exec danger --verbose
--- a/Gemfile
+++ b/Gemfile
@@ -145,6 +145,8 @@ gem "unicode-display_width"
 group :development do
   gem "better_errors"
   gem "binding_of_caller"
+  gem "danger"
+  gem "danger-auto_label"
   gem "debug_inspector"
   gem "i18n-tasks"
   gem "listen"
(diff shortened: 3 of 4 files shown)


Dangerfile
```diff
@@ -0,0 +1,32 @@
...
+end
+
+# Report if there are merge-commits in PR
+are_merge_commits_available = git.commits.any? { |c| c.parents.count > 1 }
+warn("Merge commits found in this pull request. Please, read CONTRIBUTE.md!") if are_merge_commits_available
+auto_label.set(pr_number, "Merge Commits", "ffaec9") if are_merge_commits_available
+
+# Report "Everything is fine!" if no warnings were generated
+message("Everything is fine!") if !is_big_pr && modified_yml_files.empty? && !are_merge_commits_available
+auto_label.set(pr_number, "Good PR", "00ff00") if !is_big_pr && modified_yml_files.empty? && !are_merge_commits_available
```
tomhughes, line 32: Again this could be a block if and I suspect you could use `unless violation_reports.empty?` here rather than retesting all the previous conditions, which will quickly get out of control.

Dangerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Report if some translation file (except en.yml) is modified
+unless modified_yml_files.empty?
+  modified_files_str = modified_yml_files.map { |file| "`#{file}`" }.join(", ")
+  warn("The following YAML files other than `en.yml` have been modified: #{modified_files_str}. Only `en.yml` is allowed to be changed.")
+  auto_label.set(pr_number, "Compromised Translations", "ff0000")
+end
+
+# Report if there are merge-commits in PR
+are_merge_commits_available = git.commits.any? { |c| c.parents.count > 1 }
```
tomhughes, line 26: Again it would probably be better to make this a block if, or maybe we should use the `linear_history` plugin? That would give a more generic message though.

Dangerfile
```diff
@@ -0,0 +1,32 @@
+# Remove all previously added labels
+auto_label.remove("Big PR")
+auto_label.remove("Compromised Translations")
+auto_label.remove("Merge Commits")
+auto_label.remove("Good PR")
+
+# Report if number of changed lines is > 500
+pr_number = github.pr_json["number"]
+is_big_pr = git.lines_of_code > 500
```
tomhughes, line 9: It would probably be better to make this a block if rather than doing the test and then guarding each line with the same guard.

Dangerfile
```diff
@@ -0,0 +1,32 @@
+# Remove all previously added labels
+auto_label.remove("Big PR")
+auto_label.remove("Compromised Translations")
+auto_label.remove("Merge Commits")
+auto_label.remove("Good PR")
+
+# Report if number of changed lines is > 500
+pr_number = github.pr_json["number"]
```
tomhughes, line 8: This is a common variable, not something related to this check, so it should be lifted up above it.

tomhughes (changes requested):
In addition to the inline comments, should we update `.rubocop.yml` to add an include for `Dangerfile` so it gets linted/

nenad-vujicic:
Dear @tomhughes ,

Thanks for reviewing and comments. I've just updated with above fixes. However, I'm still getting strange error during checking about setting up database. Can you please restart these checks?

Regarding linear_history plugin, its implementation is exactly the same we are using in Dangerfile and it reports "Please rebase to get rid of the merge commits in this PR", so, is it better to install new gem for perhaps simpler code or to use current solution? I would be happy to replace with using linear_history plugin if required :-)

Please, let me know if I could improve these further.

Thanks,
Nenad.

tomhughes (approved):
If it's already being linted then that fine. I thought rubocop would have caught some of the things I found which is why I thought it might not be but maybe I was just over optimistic about it.

This looks good to me now but I'll leave it for a bit in case @gravitystorm has any comments before I merge it.

nenad-vujicic:
Rebased and updated Gemfile.lock to work with the latest sources.

kcne (commented):
It would be beneficial to incorporate a check for our [commit conventions](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#committing) in this PR. Ensuring adherence to these conventions can help maintain consistency across our codebase.

I would also suggest adding pre-commit hooks to the repository. This addition could ease the maintenance workload and guide contributions and commits in the right direction before they are even made. Although I'm not fully versed in how pre-commit hooks integrate with forks, I'm willing to explore this further if you would be open to merging such a change.

tomhughes:
What do you mean by adding pre-commit hooks? If you mean that will run on people's checkouts then there is no way to add those to a repository in a way that means they will automatically be enabled in checkouts. We can certainly provide examples but people would have to enable them by hand.

kcne:
By adding pre-commit hooks, I mean providing a setup that contributors can manually enable to streamline the process of following our contributing guidelines before committing. While hooks can't be automatically enabled, we can provide contributors with scripts and instructions to simplify their setup. Tools like [Overcommit](https://github.com/sds/overcommit) for Ruby projects can be a good start for this.

nenad-vujicic:
Hi @gravitystorm, did you maybe have a chance to take a look at this PR?

Thanks!

Dangerfile
```diff
@@ -0,0 +1,44 @@
+# Get PR number and initialize is_good_pr variable
+pr_number = github.pr_json["number"]
+is_good_pr = true
+
+# Report if number of changed lines is > 500
+if git.lines_of_code > 500
+  warn("Number of updated lines of code is too large to be in one PR. Perhaps it should be separated into two or more?")
+  auto_label.set(pr_number, "Big PR", "ffff00")
```
gravitystorm, line 8: These colours are a bit eye-catching for labels. Here's some colour values from the github default colour set, which will harmonise better with our existing labels:

#B60205 - red
#D93F0B - orange
#FBCA04 - yellow

However, I've also checked the source code for the danger-auto_label plugin, and the colour values are ignored if the label already exists. This means that we can change them in future without them being overridden by this code, so maybe it doesn't matter what we set them to here.

Dangerfile
```diff
@@ -0,0 +1,44 @@
...
+  auto_label.remove("Merge Commits")
+end
+
+# Report "Everything is fine!" if no warnings were generated
+if is_good_pr
+  message("Everything is fine!")
+  auto_label.set(pr_number, "Good PR", "00ff00")
+else
+  auto_label.remove("Good PR")
+end
```
gravitystorm, line 44: I don't think we need labels on good prs, but I'm open to feedback on this. If we keep the label, then the name of the label needs to be changed to be more accurate - it's only "Good" as far as Danger is concerned, and it will be very confusing to have "Good PR" added to a PR that fails CI, which is a much more thorough assessment of the PR.

Similarly, the success message (if any) should be something different, which reflects that some aspects of the PR have been checked, but without implying that "everything" is OK.

gravitystorm (changes requested):
Overall I'm very in favour of this PR, and I only have a few minor requests to improve it.

Dangerfile
HolgerJeromin: Markdown'ified:
`#B60205` - red
`#D93F0B` - orange
`#FBCA04` - yellow

nenad-vujicic:
Updated colors (see below screenshots) and message when there are no warnings (to "PR passed all labeling tests. It is small enough, doesn't contain merge-commits and doesn't change non-EN translations!").

image

Here is how they look in dark mode:

image

Btw, failed CI test triggered all other tests to be canceled. When clicked on details, it displayed "Internal server error", so, it should pass after restart.

Please, let me know if we can further improve it.

Thanks.

Dangerfile
```diff
@@ -0,0 +1,44 @@
...
+modified_yml_files = git.modified_files.select do |file|
+  file.start_with?("config/locales") && File.extname(file) == ".yml" && File.basename(file) != "en.yml"
+end
+
+# Report if some translation file (except en.yml) is modified
+if modified_yml_files.empty?
+  auto_label.remove("Compromised Translations")
+else
+  modified_files_str = modified_yml_files.map { |file| "`#{file}`" }.join(", ")
+  warn("The following YAML files other than `en.yml` have been modified: #{modified_files_str}. Only `en.yml` is allowed to be changed.")
```
mmd-osm, line 24: We could add a hint how to contribute translations. Something like: 
 "Translations are updated via Translatewiki, see CONTRIBUTING.md."

nenad-vujicic: Added, thanks!

gravitystorm:
Please rebase, to resolve conflicts in Gemfile.lock

nenad-vujicic:
> Please rebase, to resolve conflicts in Gemfile.lock

Rebased, thanks!

nenad-vujicic:
Removed "Good" label and rebased again. Thanks!

gravitystorm:
Merged, thanks! In particular, thanks for your patience and for the multiple rebases.

Issue #2273, opened 2019-06-22
Show "reporting user" on "issues" screen.


SomeoneElseOSM:
Currently we see "Status", ""Reports", "Reported Item", "Reported User" and "Last Updated".

It'd be great to see "Reporting User" there too.  Often a person accused of problematical edits will also accuse their accuser, and sometimes one person will report a whole bunch of related items which are best handled as a group.

gravitystorm:
Since each issue can have multiple reports, we might have multiple reporters for a given issue. I would propose showing up to three reporters and then "and X more" if there are more than that.

samrjenkins:
Does anybody know if `reporting_user` is currently stored against `Issue`? Or will this change require migrations?

tomhughes:
Well it will be on the report, not the issue. It can't be on the issue because an issue can have multiple reports.

tomhughes:
So to be clear you want `issue.reports[n].user` but the real problem is that `n` because this is a one-many relationship so there is no single reporting user that can be shown.

Issue #3105, opened 2021-02-18
Message lists not paged


simonpoole:
This is obviously a rare use case, but from a correctness pov should probably be fixed.

The inbox/outbox message lists are not paged. In normal use this is likely not a real issue, however in the case of the SOSM account that we use for sending welcome messages in Switzerland the outbox now contains nearly 10'000 messages which are all shown on the same page. At some time in the future this will likely simply blow up the browser.

woodpeck:
A paginated message history would make searching ("find messages I have sent to/received from user X") much more difficult (each page would have to be searched separately). Is this an acceptable downside?

simonpoole:
> 
> 
> A paginated message history would make searching ("find messages I have sent to/received from user X") much more difficult (each page would have to be searched separately). Is this an acceptable downside?

Well it depends on when you paginate say at 1'000 messages the downside should be bearable.

PR #4990, opened 2024-07-15
Show "reporting user" on "issues" screen

nertc:
This PR addresses "Show 'reporting user' on 'issues' screen" issue mentioned in the https://github.com/openstreetmap/openstreetmap-website/issues/2273

On the issues page, there was added a new column called "Reporter Users". It will show unique reporters of the item. If there are more than reporters for the given item than the size of the block can handle, ellipsis will be shown at the end of the line. Each of user display name will have a link to the user. Fixes https://github.com/openstreetmap/openstreetmap-website/issues/2273

Update:
Added fixed max-width for reporters' column. If the text can't be fitted, ellipsis will be shown. On hover, box shows all the reporters.

Screenshot:
 
Screenshot 2024-07-18 194055

### DIFF
--- a/app/controllers/issues_controller.rb
+++ b/app/controllers/issues_controller.rb
@@ -44,6 +44,15 @@ class IssuesController < ApplicationController
     end
 
     @issues, @newer_issues_id, @older_issues_id = get_page_items(@issues, :limit => @params[:limit])
+
+    @unique_reporters = @issues.each_with_object({}) do |issue, reporters|
+      user_ids = issue.reports.order(:created_at => :desc).map(&:user_id).uniq
+      reporters[issue.id] = {
+        :count => user_ids.size,
+        :users => User.in_order_of(:id, user_ids.first(3))
+      }
+    end
+
     render :partial => "page" if turbo_frame_request_id == "pagination"
   end
 
--- a/test/system/issues_test.rb
+++ b/test/system/issues_test.rb
@@ -204,4 +204,48 @@ class IssuesTest < ApplicationSystemTestCase
       assert_no_content(/extra_#{n}[^\d]/i)
     end
   end
+
+  def test_single_issue_reporters
+    sign_in_as(create(:moderator_user))
+    issue = create(:issue, :assigned_role => "moderator")
+    issue.reports << create(:report, :user => create(:user, :display_name => "Test Name"))
+
+    visit issues_path
+    assert_content issue.reported_user.display_name
+    assert_content issue.reports.first.user.display_name
+  end
+
+  def test_multiple_issue_reporters
+    sign_in_as(create(:moderator_user))
+    issue = create(:issue, :assigned_role => "moderator")
+
+    create_list(:report, 5, :issue => issue)
+
+    visit issues_path
+    0.upto(1).each do |n|
+      assert_no_content issue.reports[n].user.display_name
+    end
+    2.upto(4).each do |n|
+      assert_content issue.reports[n].user.display_name
+    end
+  end
+
+  def test_ordering_issue_reporters
+    sign_in_as(create(:moderator_user))
+    issue = create(:issue, :assigned_role => "moderator")
+
+    create_list(:report, 5, :issue => issue)
+
+    4.downto(0).each do |n|
+      issue.reports << create(:report, :user => issue.reports[n].user)
+    end
+
+    visit issues_path
+    0.upto(2).each do |n|
+      assert_content issue.reports[n].user.display_name
+    end
+    3.upto(4).each do |n|
+      assert_no_content issue.reports[n].user.display_name
+    end
+  end
 end
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -816,6 +816,14 @@ tr.turn {
   }
 }
 
+/* Rules for the issues page */
+
+.issues.issues-index {
+  td.reporter_users {
+    max-width: 5rem;
+  }
+}
+
 /* Rules for the account confirmation page */
 
 .accounts-terms-show {
--- a/app/views/issues/_page.html.erb
+++ b/app/views/issues/_page.html.erb
@@ -13,6 +13,7 @@
           <th><%= t ".reports" %></th>
           <th><%= t ".reported_item" %></th>
           <th><%= t ".reported_user" %></th>
+          <th class="reporter_users"><%= t ".reporter_users" %></th>
           <th><%= t ".last_updated" %></th>
         </tr>
       </thead>
@@ -23,6 +24,14 @@
             <td class="text-nowrap"><%= link_to t(".reports_count", :count => issue.reports_count), issue %></td>
             <td><%= link_to reportable_title(issue.reportable), reportable_url(issue.reportable) %></td>
             <td><%= link_to issue.reported_user.display_name, issue.reported_user if issue.reported_user %></td>
+            <td class="reporter_users text-truncate">
+              <% @unique_reporters[issue.id][:users].each do |reporter| %>
+                <%= link_to reporter.display_name, reporter, :class => "d-block text-truncate", :title => reporter.display_name %>
+              <% end %>
+              <% if @unique_reporters[issue.id][:count] > 3 %>
+                <p class="m-0"><%= t ".more_reporters", :count => @unique_reporters[issue.id][:count] - 3 %></p>
+              <% end %>
+            </td>
             <td>
               <% if issue.user_updated %>
                 <%= t ".last_updated_time_ago_user_html", :user => link_to(issue.user_updated.display_name, issue.user_updated),
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1509,9 +1509,11 @@ en:
       reports: Reports
       last_updated: Last Updated
       last_updated_time_ago_user_html: "%{time_ago} by %{user}"
+      reporter_users: Reporter Users
       reports_count:
         one: "%{count} Report"
         other: "%{count} Reports"
+      more_reporters: "and %{count} more"
       reported_item: Reported Item
       states:
         ignored: Ignored


app/views/issues/index.html.erb
```diff
@@ -60,6 +61,13 @@
           <td class="text-nowrap"><%= link_to t(".reports_count", :count => issue.reports_count), issue %></td>
           <td><%= link_to reportable_title(issue.reportable), reportable_url(issue.reportable) %></td>
           <td><%= link_to issue.reported_user.display_name, issue.reported_user if issue.reported_user %></td>
+          <td>
+            <% unique_reporters = issue.reports.uniq { |report| report.user.id } %>
```
tomhughes, line 65: This could probably be better done as `issue.reports.preload(:user).map(&:user).uniq` which would also get a list of users instead of a list of reports.

app/views/issues/index.html.erb
```diff
@@ -60,6 +61,13 @@
...
+            <% unique_reporters[..2].each_with_index do |report, index| %>
```
tomhughes, line 66: This could be done more simply with something like:

```
safe_join(unique_reporters[..2].map { |u| link_to u.display_name, u }, ",")
```

That's assuming the previous edit to make the reporter list a list of users.

app/views/issues/index.html.erb
```diff
@@ -60,6 +61,13 @@
...
+              <%= link_to(report.user.display_name, report.user) %><%= "," unless index == 2 || index == unique_reporters.size - 1 %>
+            <% end %>
+            <%= "..." if unique_reporters.size > 3 %>
```
tomhughes, line 69: Maybe instead of truncating the list here we should put the whole thing in but use CSS to limit the width and add an ellipsis? With a tooltip to show the follow value even?

tomhughes (changes requested):
Although this is described as showing the "last three unique reporters" there isn't actually any ordering that I can see, and ordering on time is hard to combine with uniqification...

AntonKhorev:
Currently the ignored issues page takes 10 seconds to load and the resolved issues won't load at all for me. If the queries are to be any heavier, it's going to get even worse.

nertc:
@tomhughes Thanks for the review. I've updated according to your feedback. Tooltip is a good idea for giving more information, but I have a question about, if users get the information correctly and won't be confused, because when a user hovers on a profile link, tooltip will show all the reporters (current solution) and he might be confused if the link redirects him to the specific user or some list of users. If we add a tooltip only to the places with no link, user will have to hover either semicolons or ellipsis, which are hard to target because of their small size.

nertc:
@AntonKhorev If the issues page is already slow, we can temporarily archive this PR and reopen it after optimization.

nertc:
@AntonKhorev About the issues page loading time, I am interested, what is the average number of issues displayed, when the page is loaded. If the loading problem arises because of the great number of issues, we can add infinite scroll (load-on-scroll) functionality or pagination to resolve it. If either solution is okay, I'll create a separate PR for it.

AntonKhorev:
> what is the average number of issues displayed, when the page is loaded.

There's an acceptable number of open issues because they are being dealt with, but once that happens they become either resolved or ignored, those numbers keep growing. Total number of issues is currently at ~30000, that means 2000..20000 issues on ignored/resolved pages for moderator/administrator queues.

>If the loading problem arises because of the great number of issues, we can add infinite scroll (load-on-scroll) functionality or pagination to resolve it.

We can, if we want to trade it for another problem of making searches more difficult. See #3105 for a similar issue.

nertc:
@AntonKhorev What do you mean with making searches more difficult? From the user experience perspective, infinite scroll doesn't add any difficulties or additional clicks to see issues. If you mean searching with a browser built-in search functionality (because it needs all the issues to be rendered to work) and admins/moderators are using it, then we have to change or enhance our search so that browser built-in search wouldn't be needed. This way we will both optimize the page and make the user experience more website oriented.

nertc:
As https://github.com/openstreetmap/openstreetmap-website/pull/5057 is merged, this PR was updated to match pagination changes of the issues page.

In terms of performance, on the local machine I have 1000 users and 1000 issues, and pagination time was increased only by 150-200ms (it should be similar increase in the real environment, as before for the https://github.com/openstreetmap/openstreetmap-website/pull/5057 I tried to match up timing of the local and real environments).

All the Ruby variable initializations and calculations (from the previous solution) were moved from `issues/index.html.erb` file to `issues_controller.rb` to make html files less filled with calculation code.

Also, as we are using Bootstrap Tooltips in other places, `turbo_tooltip.js` file was created. [Bootstrap Tooltips Documentation](https://getbootstrap.com/docs/5.0/components/tooltips/) mentions:

> ... you must initialize them yourself

> Tooltips must be hidden before their corresponding elements have been removed from the DOM.

Because of these limitations, it has some bugs when working with Turbo (tooltips are not closing, tooltips got frozen on the top-left side of the screen ...). `turbo_tooltip.js` initializes Bootstrap Tooltips, solves the problem that caused tooltips bugs and also is reusable to be used in the future everywhere, where Turbo will be used and tooltips will be needed (Diary Entries page, Diary Entry Comments, User Blocks page ...).

app/views/issues/index.html.erb
nertc: `safe_join` was used to join names list.

app/views/issues/index.html.erb
nertc: To avoid Ruby on Rails logic code in the HTML, `preload` logic was moved in the controller.

AntonKhorev:
I don't think tooltips were a good idea. You can't interact with the usernames they show. And truncation works fine for individual names, but then you'd have to do it inside the tooltip too. Do you want to make html tooltips? And then there's Turbo cleanup.

image

nertc:
> And if there's more than say 3, write something like
> 
> > UserA
> > UserBBBBBBBBB...
> > UserC
> > and 2 more users

@AntonKhorev do you mean to make this kind of truncation in the tooltip?

AntonKhorev:
I mean let's not have tooltips at all. This is what a table cell contents may look like:

> UserA
> UserBBBBBBBBB...
> UserC
> and 2 more users

"and 2 more users" initially can be just a passive text, later it can bechanged to some js expansion or turbo streams rewrite.

nertc:
@AntonKhorev It will make `Reporter Users` cells too high which will cause additional whitespaces in other cells.

AntonKhorev:
> It will make Reporter Users cells too high which will cause additional whitespaces in other cells.

Is that additional whitespace going to be a problem?

nertc:
> > It will make Reporter Users cells too high which will cause additional whitespaces in other cells.
> 
> Is that additional whitespace going to be a problem?

Sorry for the late answer. I think it will be a problem in terms of UI and UX:
UI - visually, additional whitespaces in the table make cells feel disconnected from each other and also create an impression that something is missing in the cell (something went wrong)
UX - as there will be additional whitespaces, they'll make page longer and, therefore, harder to scroll down. In addition to this, big whitespaces in cells generally make information harder to read (instead of regular left-to-right or right-to-left reading, user has to read every block of information top-to-bottom)

AntonKhorev:
> instead of regular left-to-right or right-to-left reading, user has to read every block of information top-to-bottom

Which blocks are "every"? Reporter user lists? Aren't lists regularly read top-to-bottom?

nertc:
@AntonKhorev I'll update this PR to have vertical list of `Reporting users`. Do you also suggest to completely remove tooltips? I think tooltips are quite handy even if user can't click on the profile links (to just take a glance at reporting users' names), but it introduces additional complexities (which I don't know if are worth it, though, I think, as these complexities are created as reusable code, it will benefit website code in future too)

AntonKhorev:
> I think tooltips are quite handy even if user can't click on the profile links (to just take a glance at reporting users' names)

Can't you don this already by looking at the link location that the browse shows?

nertc:
> > I think tooltips are quite handy even if user can't click on the profile links (to just take a glance at reporting users' names)
> 
> Can't you don this already by looking at the link location that the browse shows?

If there is no tooltip on the issues, which has more than 3 reporters, users won't be able to take a look at reporters full list without opening the issue page.

AntonKhorev:
> If there is no tooltip on the issues, which has more than 3 reporters, users won't be able to take a look at reporters full list without opening the issue page.

For now they won't. But then we'll make it clickable to show the rest of the users. This should work better than a tooltip because it's usable with touch interfaces too.

nertc:
PR was updated. Bootstrap tooltips were removed. Instead of the horizontal view of reporters list, now it is vertical and after 3 users it shows "and X more". I tried aligning data of the table vertically middle, but top one was visually more appealing. This is screenshot of how it looks now:
image

AntonKhorev:
Are reporters sorted? Judging by https://github.com/openstreetmap/openstreetmap-website/pull/4990#pullrequestreview-2180762354 there was an attempt to show last three unique reporters, but now looks like it's not happening.

If there's an issue that was resolved, it has `read_reports`. Then the same item may get reported again and get new `unread_reports`. If you look at the issues list, reporters from `read_reports` are not as important and maybe shouldn't be shown at all. Or maybe they should, but only if the status filter is set to *resolved* (otherwise there's no available reports because they all are read).

But all of those considerations could be skipped (at least initially) if we sort reporters by their last reports. Unread reports always come chronologically after read reports, therefore last reporters will come from unread reports first.

nertc:
Sorting reporters by their last reports is a good idea. I'll do it and update PR accordingly. I think, after this PR is merged then, if needed, we can enhance "reporters" functionality by adding some filters to it. As `reports` and `issues` are different tables, adding filters to `reports` may cause some additional performance and UX difficulties. I think, it deserves a separate PR.

nertc:
PR was updated. Reporters are now sorted according to their last report creation datetime in descending order. Tests were also added to test sorting functionality.

app/controllers/issues_controller.rb
```diff
@@ -44,6 +44,11 @@ def index
     end
 
     @issues, @newer_issues_id, @older_issues_id = get_page_items(@issues, :limit => @params[:limit])
+
+    @unique_reporters = @issues.each_with_object({}) do |issue, reporters|
+      reporters[issue.id] = issue.reports.order(:created_at => :desc).preload(:user).map(&:user).uniq
```
AntonKhorev, line 49: Is it possible to select unique user ids, then limit them to 3 per issue, and only after that load users?

nertc: Technically it's possible. I'll try to do it in the cleanest possible way.

nertc: PR was updated. Now "loading users" only affects first 3 users that should be shown.

app/controllers/issues_controller.rb
```diff
@@ -44,6 +44,15 @@ def index
     end
 
     @issues, @newer_issues_id, @older_issues_id = get_page_items(@issues, :limit => @params[:limit])
+
+    @unique_reporters = @issues.each_with_object({}) do |issue, reporters|
+      user_ids = issue.reports.order(:created_at => :desc).map(&:user_id).uniq
+      reporters[issue.id] = {
+        :count => user_ids.size,
+        :users => User.where(:id => user_ids.first(3)).index_by(&:id).values_at(*user_ids.first(3))
```
AntonKhorev, line 52:
```suggestion
        :users => User.in_order_of(:id, user_ids.first(3))
```

test/system/issues_test.rb
```diff
@@ -204,4 +204,53 @@ def test_issues_pagination
...
+    assert_content issue.reports.first.user.display_name
+  end
+
+  def test_multiple_issue_reporters
+    sign_in_as(create(:moderator_user))
+    issue = create(:issue, :assigned_role => "moderator")
+
+    1.upto(5).each do |n|
+      issue.reports << create(:report, :user => create(:user, :display_name => "Test Name #{n}"))
+    end
```
AntonKhorev, line 224:
```suggestion
create_list(:report, 5, :issue => issue)
```
because you're not checking if usernames match `"Test Name #{n}"`.

And probably the same in the next test.

app/controllers/issues_controller.rb
nertc: Thanks for the suggestion. PR was updated. User selection was changed with much smaller and cleaner code you suggested.

test/system/issues_test.rb
nertc: Thanks for the suggestion. Both tests were changed to match smaller list creation you suggested.

AntonKhorev:
Although it can still be optimized, let's start with this.

tomhughes:
The phrase "reporting user" as in the title of this PR makes more sense than "reporter user" that is actually used...

PR #4991, opened 2024-07-15
Check comment ids instead of factory sequence numbers in changeset feed tests

AntonKhorev:
I'm not sure what [that](https://github.com/openstreetmap/openstreetmap-website/pull/4944#discussion_r1665980153) was about because comment ids are not inside CDATA or HTML or `<description>`, but rather inside `<link>` or `<guid>`:

image

The test picked numbers from
https://github.com/openstreetmap/openstreetmap-website/blob/a3441739ca4b70ee634e4e3c29415efe738df52e/test/factories/changeset_comments.rb#L3
which happens to also work because if they are descending so do the ids. But it's more complicated because of extra html parsing and more confusing because these numbers are called id yet they are not.


tomhughes:
Thanks - that looks a lot nicer!

PR #4993, opened 2024-07-15
Set OpenID URL input value in html instead of javascript

AntonKhorev:
That javascript comes from [ages ago](https://github.com/openstreetmap/openstreetmap-website/commit/ebd109c946d71b3ccc616ea3966e56ae019638d7) when each auth provider button was setting its own value.


tomhughes:
Looks good to me, thanks.

PR #4995, opened 2024-07-16
Remove name argument from auth button helpers

AntonKhorev:
`auth_button` and `auth_button_preferred` helpers have the following first two arguments:
- `name`, which is an auth provider name
- `provider`, which is also an auth provider name, exactly the same

They used to be different [at some point](https://github.com/openstreetmap/openstreetmap-website/commit/e434cb154c5ccecd7fc0ce672e4de79bd9e7e06f), but not anymore. Now the calls look like `auth_button provider, provider`.

We can remove the `name` argument.

### DIFF
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -52,29 +52,29 @@ module UserHelper
 
   # External authentication support
 
-  def auth_button(name, provider, options = {})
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good to me, thanks.

PR #4757, opened 2024-05-06
Simplify message paths in tests

AntonKhorev:
See #4641 and previous similar pull requests.

PR #4996, opened 2024-07-16
Simplify subscribe and hide path helper calls

AntonKhorev:
Like #4757 and previous PRs.


tomhughes:
Looks good to me, thanks.

PR #4998, opened 2024-07-17
Set search input value to 'lat, lon' params if query param is missing

### LINKED ISSUE #4997: 'Show address' no longer adds coords to search box
### URL

https://www.openstreetmap.org/search?lat=-32.05040&lon=115.77587#map=19/-32.05040/115.77587

1. Right click anywhere on the map and select 'Show address'
2. See the coordinates show up under 'Results from Internal'
3. Note that the coordinates _do not_ show up in the search box as they used to.

OpenStreetMap

AntonKhorev:
Fixes https://github.com/openstreetmap/openstreetmap-website/pull/4959#issuecomment-2232620015

Fixes #4997


tomhughes:
Looks good to me, thanks.

PR #4999, opened 2024-07-17
Remove unnecessary loop when rendering preferred auth provider button

AntonKhorev:
If prefered_auth_button_available == true, we've already found @preferred_auth_provider among all available providers, no need to look for it again.


tomhughes:
Looks good to me, thanks.

PR #5001, opened 2024-07-18
Set search input value when searching for coordinates

AntonKhorev:
Fixes setting search input value after full page reloads that only seems to work because of [autocompletion](https://github.com/openstreetmap/openstreetmap-website/pull/4959#issuecomment-2233238733).

`_search.html.erb` fills the input with `params[:query]` which is not going to exist after geocoder's `normalize_params`.


tomhughes:
Looks good to me, thanks.

PR #5002, opened 2024-07-18
Use Auth.providers in auth providers template

AntonKhorev:
`_auth_providers.html.erb` iterates over `%w[google facebook microsoft github wikipedia]` and checks which provider is available. But the `Auth` module already does that. `Auth` is only used on account settings pages and has a "provider" "None" specific to these pages. If "None" is removed, `Auth` can be reused in `_auth_providers.html.erb`.


tomhughes:
Looks good to me, thanks.

PR #5003, opened 2024-07-18
Fix marking muted messages as read/unread

AntonKhorev:
Otherwise you get this error:

image

image


tomhughes:
Looks good to me, thanks.

PR #5007, opened 2024-07-19
Move auth provider icons to auth_providers subdirectory

AntonKhorev:
It's a little easier to work with them when they are not in one directory with every other image.


tomhughes:
Looks good to me, thanks.

Issue #4931, opened 2024-06-28
Map Data checkbox: perhaps use toggle slider instead


jidanni:
### Problem

Here we are in the right side panel Layers menu,
looking at the Map Data checkbox:
 
![Screenshot 2024-06-28 21 03 37](https://github.com/openstreetmap/openstreetmap-website/assets/1230959/91f7dc3f-e1f9-42bf-b9f0-c0a191ea3315)

In a desert or jungle, checking and unchecking the checkbox works fine. Any data items will appear on the map in blue.

But in big cities with plenty of streets and shops, checking the Map Data checkbox might entail a massive data flow to one's computer. So much so that it becomes unresponsive. Whereupon we wonder if we really did check the checkbox in the first place. So we think we didn't and then check it again, only compounding the problem, etc.

Therefore perhaps if a slider was used, like often seen on cellphones, that way we could see if we really slid it to the on or off position or not.
Hopefully that would work differently than the checkbox.

Also first update the indicator (checkbox or slider) (to some intermediate state), then do the task, that might help here too.

And also display a prominent "Loading..." while the data is still flowing in. And maybe even have an "abort" button... even if pressing ESC or X in the browser also works.

When testing, remember that some users have very slow data connections too.

mmd-osm:
Similar to #2800 and #3243. Maybe we can merge all 4 current issues into one, and check the performance part first: https://github.com/openstreetmap/openstreetmap-website/issues?q=%22map+data%22+label%3Amapview+

nertc:
This issue is not closed and needs to be reopened. Sorry, it seems I wrongly used `Fixes` keyword. I meant that https://github.com/openstreetmap/leaflet-osm/pull/38 was part of the fix, not the whole fix.

PR #5009, opened 2024-07-20
Map Data checkbox: perhaps use toggle slider instead

nertc:
This PR addresses "Map Data checkbox: perhaps use toggle slider instead" issue mentioned in the https://github.com/openstreetmap/openstreetmap-website/issues/4931

Several changes were made to the data loading functionality:

1. Clicking "Map Data" checkbox will add loading spinner to the label. It will be shown until back end collects and returns all the information (see "Loading Video"). After all the information is returned, spinner will be removed.
2. After getting information and agreeing to show it on the map, rendering process will be asynchronous. Therefore, there will be no more browser or page freezes and it will have animation like visual (see "Render Video").
3. If user unchecks "Map Data" checkbox, removing process will be also asynchronous and there won't be any freezes. It will also have animation like visual (see "Remove Video").

- Checkbox state will be updated as soon as it is clicked, so changing it to the toggle one has no more benefits.
- If the user decides to uncheck "Map Data" while back end collects information, loading spinner will disappear.
- If the user decides to uncheck "Map Data" while objects is rendered, it won't trigger change until the rendering process is done. After the process, application will check and if "Map Data" is unchecked, it will start removing rendered objects.

Connected to the PR https://github.com/openstreetmap/leaflet-osm/pull/38
Fixes https://github.com/openstreetmap/openstreetmap-website/issues/4931

### Videos:

Loading Video

https://github.com/user-attachments/assets/4b8a9245-4847-4a2f-8c40-ad16ce933e05

Render Video

https://github.com/user-attachments/assets/8e11fc5f-1d9c-4458-a5a8-96479627d5ff

Remove Video

https://github.com/user-attachments/assets/0743916d-353e-4717-8600-c74937b6f2fe

### DIFF
--- a/app/assets/javascripts/index/layers/data.js
+++ b/app/assets/javascripts/index/layers/data.js
@@ -89,6 +89,15 @@ OSM.initializeDataLayer = function (map) {
 
     if (dataLoader) dataLoader.abort();
 
+    $("#layers-data-loading").remove();
+
+    const spanLoading = $("<span>")
+      .attr("id", "layers-data-loading")
+      .attr("class", "spinner-border spinner-border-sm ms-1")
+      .attr("role", "status")
+      .html("<span class='visually-hidden'>" + I18n.t("browse.start_rjs.loading") + "</span>")
+      .appendTo($("#label-layers-data"));
+
     dataLoader = new AbortController();
     fetch(url, { signal: dataLoader.signal })
       .then(response => {
@@ -131,7 +140,10 @@ OSM.initializeDataLayer = function (map) {
           $("#browse_status").empty();
         });
       })
-      .finally(() => dataLoader = null);
+      .finally(() => {
+        dataLoader = null;
+        spanLoading.remove();
+      });
   }
 
   function onSelect(layer) {
--- a/app/assets/javascripts/leaflet.layers.js
+++ b/app/assets/javascripts/leaflet.layers.js
@@ -100,6 +100,7 @@ L.OSM.layers = function (options) {
 
         const label = $("<label>")
           .attr("class", "form-check-label")
+          .attr("id", `label-layers-${name}`)
           .appendTo(item);
 
         let checked = map.hasLayer(layer);
@@ -114,10 +115,15 @@ L.OSM.layers = function (options) {
 
         input.on("change", function () {
           checked = input.is(":checked");
+          if (layer.cancelLoading) {
+            layer.cancelLoading();
+          }
+
           if (checked) {
             map.addLayer(layer);
           } else {
             map.removeLayer(layer);
+            $(`#layers-${name}-loading`).remove();
           }
         });
 
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -47,7 +47,7 @@ L.OSM.Map = L.Map.extend({
     this.noteLayer = new L.FeatureGroup();
     this.noteLayer.options = { code: "N" };
 
-    this.dataLayer = new L.OSM.DataLayer(null);
+    this.dataLayer = new L.OSM.DataLayer(null, { asynchronous: true });
     this.dataLayer.options.code = "D";
 
     this.gpsLayer = new L.OSM.GPS({
@@ -306,7 +306,8 @@ L.OSM.Map = L.Map.extend({
               way: objectStyle,
               area: objectStyle,
               changeset: changesetStyle
-            }
+            },
+            asynchronous: true
           });
 
           map._objectLayer.interestingNode = function (node, wayNodes, relationNodes) {
--- a/vendor/assets/leaflet/leaflet.osm.js
+++ b/vendor/assets/leaflet/leaflet.osm.js
@@ -104,6 +104,13 @@ L.OSM.DataLayer = L.FeatureGroup.extend({
     }
   },
 
+  loadingLayers: [],
+
+  cancelLoading: function () {
+    this.loadingLayers.forEach(layer => clearTimeout(layer));
+    this.loadingLayers = [];
+  },
+
   addData: function (features) {
     if (!(features instanceof Array)) {
       features = this.buildFeatures(features);
@@ -220,9 +227,11 @@ L.OSM.DataLayer = L.FeatureGroup.extend({
   eachLayer: function (method, context, asynchronous = false) {
     for (let i in this._layers) {
       if (asynchronous) {
-        setTimeout(() => {
-          method.call(context, this._layers[i]);
-        });
+        this.loadingLayers.push(
+          setTimeout(() => {
+            method.call(context, this._layers[i]);
+          })
+        );
       } else {
         method.call(context, this._layers[i]);
       }


AntonKhorev:
Are you planning to get the changes to `vendor/assets/leaflet/leaflet.osm.js` merged in https://github.com/openstreetmap/leaflet-osm ?

nertc:
@AntonKhorev Yes. Should I create a separate PR in the https://github.com/openstreetmap/leaflet-osm for it?

nertc:
I created a PR for the [leaflet-osm](https://github.com/openstreetmap/leaflet-osm) (https://github.com/openstreetmap/leaflet-osm/pull/38) that needs to be merged before this PR will be, as asynchronous loading part depends on it.

app/assets/javascripts/index/layers/data.js
```diff
@@ -86,6 +86,16 @@ OSM.initializeDataLayer = function (map) {
...
     if (dataLoader) dataLoader.abort();
 
+    let spanLoading = $("#layers-data-loading");
+
+    if (!spanLoading.length) {
+      spanLoading = $("<span>")
+        .attr("id", "layers-data-loading")
+        .appendTo($("#label-layers-data"));
+
+      spanLoading.append(" " + I18n.t("javascripts.map.layers.loading"));
```
HolgerJeromin, line 96: I am not sure if this "append" is safe in laguages which are written from right to left.

AntonKhorev: This is used in other places for loading indicators:
https://getbootstrap.com/docs/5.3/components/spinners/

nertc: @AntonKhorev Thanks for the recommendation. I replaced a text with a bootstrap spinner.

HolgerJeromin: With rtl the spinner should be at the left.

https://developer.mozilla.org/en-US/docs/Web/CSS/:dir
Could be a solution with ::after or ::before elements.

nertc: @HolgerJeromin Now it works this way when Arabic is chosen as a default language, spinner is shown on the left side of the text.

https://github.com/user-attachments/assets/21d7c1ef-8bec-4664-9236-7ecc5aef5b81

HolgerJeromin: I have no idea how it works, but the video looks nice :-)

nertc:
This PR is linked to the https://github.com/openstreetmap/leaflet-osm/pull/38. If leaflet-osm PR is not merged, this PR can't be merged.

nertc:
PR was updated to include changes made in the https://github.com/openstreetmap/leaflet-osm/pull/38. As leaflet-osm PR is now merged, this PR can be reviewed and merged.

AntonKhorev:
Now you can check *Map Data*, wait for the rendering to start then quickly click *Map Data* 2*n times and watch how data is added then removed then added then removed... n times. While this is happening and *Map Data* is checked, pan to a different place to get a stream of javascript errors.

nertc:
@AntonKhorev I couldn't reproduce JS errors in the console. But I changed logic and now when user clicks "Map Data" checkbox, previous action is skipped immediately. Therefore, there will be no 2*n loads and errors. When we will be okay with the solution, I'll create a new PR in the [leaflet-osm](https://github.com/openstreetmap/leaflet-osm).

app/assets/javascripts/index/layers/data.js
nertc: `Loading` text was changed with the loader spinner.

nertc:
If there are no other change requests for this PR, I'll create a new PR for the [leaflet-osm](https://github.com/openstreetmap/leaflet-osm) to merge changes done in the `vendor/assets/leaflet/leaflet.osm.js`.

nertc:
PR was updated. Branch was synced with master and conflicts were resolved. Also, instead of handling canceling layer loading in the `setTimeout` callback, now it is handled by clearing `setTimeouts`.

Optimizations for layer rendering were done in https://github.com/openstreetmap/leaflet-osm/pull/43, but still this PR is relevant, because it adds loading spinner for those cases when data is processed by the back end, and also for slower CPUs (for example, when I set `6x slowdown` to the CPU) it will provide smooth asynchronous loading without any freezes.

nertc:
PR was updated. `master` was rebased. Conflicts were resolved. Failed tests were not impacted by this PR (for more information see https://github.com/openstreetmap/openstreetmap-website/pull/5439#issuecomment-2711306484).

nertc:
This PR is linked to the https://github.com/openstreetmap/leaflet-osm/pull/47.

tomhughes (changes requested):
This mostly works great - the only issue I can see is that if you enable the data layer and then reload the page then it doesn't enable the spinner while it does the initial page load though it does appear on subsequent map drags.

nertc:
PR was updated. `master` was rebased.

@tomhughes Thank you for the review. Spinner issue on page refresh was fixed.

tomhughes:
Thanks for the fix. I think this looks good now.

deevroman:
This PR brought two problems:

1. When moving the map, a white flash occurs and all objects on the map are redrawn:

https://github.com/user-attachments/assets/3c87c769-174c-4513-a5be-115eb9517bb8

2. Rendering slowed down. What used to occupy ~400ms now occupies ~1.3s. Hiding a data layer was previously performed on ~100ms, now ~1.5s (!)

Both problems arise due to the fact that before that the redrawing of the map was performed only once, after the completion of the removal/adding of objects to the SVG-element of the map.

I can agree that the first appearance of the data layer looks beautiful. But the work with the map worsened.

nertc:
@deevroman Thank you for your feedback. #5799 disabled asynchronous layer loading and, therefore, there shouldn't be any flashes or slow rendering process anymore.

deevroman:
@nertc no, https://github.com/openstreetmap/openstreetmap-website/pull/5799 disabled asynchrony not for the Map Data layer, but for the layer that shows the active object.

nertc:
You are right, I thought, it disabled both asynchronous functionalities.
I pulled `master` and tested turning asynchronous loading off.

1. About flash: it is because data is loaded asynchronously and as everything is rendered one by one (and not as one operation that may cause freezing of the website), on fast machines it may be perceived as a flash.
2. About performance: these are several rendering test results from my local machine:
    ```
    3260 features | 178ms -> 328ms (~1.8x)
    3163 features | 169ms -> 264ms (~1.5x)
    2319 features | 114ms -> 196ms (~1.7x)
    ```
    Therefore, on average previously it was 1.7x faster (which is logical, because now it is asynchronous and is not done in one operation).

But in terms of redrawing all of the features, when map is moved, there was no change that would impact that behavior. Even after turning asynchronous functionality off, all of the features were still redrawn.

Turning asynchronous functionality off will solve flash problem, but redrawing all the features will stay the same. Though, for those who have weaker CPUs (for example when I run performance test on `CPU 20x slowdown`), rendering data asynchronously helps user not to have frozen web page and browser. I think, if you suggest turning this functionality off or some optimization about rendering data, it will be better to be discusses in a separate issue.

deevroman:
What do you use to test CPU slowdowns? Test rendering >4k features.

Can you give us an example of when the user has time to notice the browser hanging, while making the user guess that they are waiting for rendering, and not downloading data?

nertc:
For `CPU slowdown` I use browser built-in functionality.

Difference between getting data from server and rendering them is that while getting data, there are no CPU intense tasks and loading icon is shown next to the `data layer` checkbox. When rendering is started, it uses CPU to render data and may freeze application or browser.

In the video you can see when rendering data can take long time (full video can't be uploaded because of the limitations of GitHub, in reality it took about 30 seconds to render all elements). In such case, there would be 30 seconds freeze of browser if there was synchronous process. 

https://github.com/user-attachments/assets/266a42bc-f150-42e0-ae69-709781c7bda0

deevroman:
> For CPU slowdown I use browser built-in functionality.

Ah, chromium, got it. 20x slowdown is unrealistic. With this value, I can't move the map normally, even without the data layer. But even with 20x slowdown, the layer with 11k features is drawn in 4 seconds when rendered synchronously on my PC.

> In such case, there would be 30 seconds freeze of browser if there was synchronous process.

This is not true because synchronous rendering is faster than asynchronous rendering. 

In general, I will most likely suggest a PR that disables asynchrony.

PR #5011, opened 2024-07-22
Select closest value to remaining duration when editing a block

AntonKhorev:
Let's suppose you create a block and you set its duration to 12 hours. Then you realize you need to edit this block. You go to `/user_blocks/:id/edit` and you see among other things a duration select set to 12 hours as expected:
image

But what if you come an hour later? Now the remaining duration is 11 hours which is not one of the available select options. The select will look like this, defaulting to 0 hours, probably not the best choice:
image

Here I set the select to the closest available value.


tomhughes:
Looks good to me, thanks.

PR #5012, opened 2024-07-23
Check diary entry edit/update ability using CanCanCan

AntonKhorev:
Does [this](https://github.com/openstreetmap/openstreetmap-website/pull/4540#discussion_r1610270567) but for `:update` ability (which includes `:edit` ability).


tomhughes:
Looks good to me.

For anybody that isn't aware of it @AntonKhorev is taking advantage of aliases defined by https://github.com/CanCanCommunity/cancancan/blob/develop/docs/define_check_abilities.md#can-actions which means that `:update` implies `:edit` as well. I can see there are a number of other places where we could take advantage of those aliases...

PR #5013, opened 2024-07-23
Change diary comment post paths

AntonKhorev:
This is how path are going to look if comments are turned into nested resources of diary entries:

- `newcomment` => POST on `comments`
- `hidecomment/:comment` => `comments/:comment/hide`
- `unhidecomment/:comment` => `comments/:comment/unhide`

All changed paths are POST paths, changing them shouldn't break anything.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -237,13 +237,13 @@ OpenStreetMap::Application.routes.draw do
   scope "/user/:display_name" do
     resources :diary_entries, :path => "diary", :only => [:edit, :update, :show], :id => /\d+/
   end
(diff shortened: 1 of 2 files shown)


tomhughes:
Isn't the canonical naming rule in rails to use the singular form for individual objects? So `/comments` for the index but `/comment/:id` for a specific comment?

AntonKhorev:
Normally paths are in [plural](https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions). Singular is for [singular resources](https://guides.rubyonrails.org/routing.html#singular-resources), if, for example, a diary entry could have only one comment.

tomhughes:
Guess I looked at the wrong examples in our existing routes then... So this looks good now, thanks.

PR #5014, opened 2024-07-24
Allow to edit inactive blocks

AntonKhorev:
Expired/revoked blocks have no edit links. Editing them results in "The block has already expired and cannot be edited."

This PR removes the `active?` check on update actions which allows editing inactive blocks.

## Why?

Some block are issued automatically and there are false positives among them. Any moderator can revoke erroneous blocks but it's also desirable to edit the block messages. This may be impossible for two reasons:
1. Revoked block is already expired.
2. Revoked block was created by a different moderator.

This PR removes reason 1.


tomhughes:
Looks good to me, thanks.

PR #5016, opened 2024-07-25
Allow to edit revoked blocks

AntonKhorev:
Part two of #5014. Those who revoke a block will be able to edit it.

### DIFF
--- a/test/controllers/user_blocks_controller_test.rb
+++ b/test/controllers/user_blocks_controller_test.rb
@@ -212,7 +212,7 @@ class UserBlocksControllerTest < ActionDispatch::IntegrationTest
     check_block_buttons block, :edit => 1
 
     session_for(revoker_user)
-    check_block_buttons block
+    check_block_buttons block, :edit => 1
   end
 
   private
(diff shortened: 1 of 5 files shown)


test/controllers/user_blocks_controller_test.rb
```diff
@@ -483,13 +483,44 @@ def test_update_expired
...
+    session_for(creator_user)
+    check_block_updates(block)
+
+    session_for(revoker_user)
+    check_block_updates(block)
+  end
+
+  private
+
+  def check_block_updates(block)
```
tomhughes, line 516: Can we move this to the end with the other private methods rather than switching back and forth between public and private?

tomhughes (changes requested):

test/controllers/user_blocks_controller_test.rb
AntonKhorev: Do you want to move only `check_block_updates` or every other such method too?

tomhughes: Oh I hadn't realised you had done the same thing in that other PR that I merged yesterday... Personally I would prefer to move that as well.

AntonKhorev: I moved the methods here. But I didn't do a similar private section just yesterday, there should be more of them elsewhere.

tomhughes (approved):
Looks good now, thanks.

PR #5017, opened 2024-07-25
Use assert_not_predicate instead of assert_not next to assert_predicate

AntonKhorev:
Rubocop would complain about something like `assert issue.open?` and it gets changed to `assert_predicate`, but `assert_not issue.open?` is left alone for some reason.

I already did this change in #4454. I guess I missed some.


tomhughes:
I think I did explain on #4454 why it doesn't  find them. Looks good to me anyway.

PR #5018, opened 2024-07-25
Simplify api changeset path helper calls

AntonKhorev:
Like #4996


tomhughes:
Looks good to me, thanks.

PR #5019, opened 2024-07-25
Add expiry for OAuth 2 grants and tokens

tomhughes:
This extends the existing cleanup script that runs every night in production to deal with OAuth 2 grants and tokens.


gravitystorm:
Merged, thanks. 

At some point we could convert this script into a rake task, to make it more discoverable (and add a description etc). Or even better, a simple rake task that calls a method on another class, which makes the logic easier to test. But for now that's all overkill.

PR #5020, opened 2024-07-25
Convert tests using OAuth 1 tokens to use OAuth 2 tokens

tomhughes:
This converts some tests which are testing OAuth 1 tokens and which don't have an OAuth 2 equivalent to use OAuth 2 instead.


AntonKhorev:
Thanks, merged.

Do we now drop tests like https://github.com/openstreetmap/openstreetmap-website/blob/a932d7cd7939e3d74fbe8009ea518a5e3361d060/test/controllers/api/users_controller_test.rb#L78 ?

tomhughes:
I've got another PR coming up to remove OAuth 1 which will remove those.

PR #5021, opened 2024-07-26
Move notes and blocks resources to api namespace

AntonKhorev:
If we want to [avoid](https://github.com/openstreetmap/openstreetmap-website/pull/4248#discussion_r1326279791) `:as` for api path helpers we can move resources to a namespace that will add the `api` prefix automatically.

This PR shouldn't rename any path helpers.


tomhughes:
Looks good to me, thanks.

PR #5023, opened 2024-07-26
Added bootstrap alert when body preview is empty

### LINKED ISSUE #3748: Indicate if the message preview is empty

nenad-vujicic:
PR is alternative to #4894 for displaying indicator, using bootstrap alerts, if (message, diary entry, diary entry comment) body is empty and user selects Preview button.

Fixes #3748. Added check to SiteController#preview to detect if (message, diary entry, diary entry comment) body is empty and if positive, returns rendered bootstrap alert "Nothing to preview" to be displayed in richtext_field.

After:
image

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -138,7 +138,12 @@ class SiteController < ApplicationController
   end
 
   def preview
-    render :html => RichText.new(params[:type], params[:text]).to_html
+    if params[:text].blank?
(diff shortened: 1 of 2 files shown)


AntonKhorev:
Thanks, merged.

PR #5024, opened 2024-07-26
Queries and keys controllers in features namespace

AntonKhorev:
Version 2 of #4616.

`BrowseController` only did feature queries. #4616 renamed it to `QueriesController` but that wasn't entirely ok because it didn't tell what was queried.

Here I'm renaming it to `Features::QueriesController`.


tomhughes:
Are we every likely to have anything else in `Features` though? Maybe it should just be `FeatureQueries` instead?

Also should it be singular `Query` rather than `Queries`?

AntonKhorev:
> Are we every likely to have anything else in Features though?

Moved map key there too.

I thought about combining them into one toolbar button. When you click it, it shows map key. If you click on the map after that, it does a query. I haven't done this because it's possible to make queries work without an extra click, and then it's not clear how to get to the map key.

AntonKhorev:
Superseded by https://github.com/openstreetmap/openstreetmap-website/pull/5751 and #5830.

PR #5029, opened 2024-07-29
Copyedit English localization as American English

### LINKED ISSUE #4975: Missing American English localization
As far as I can tell, config/locales/en.yml is _de facto_ written in Commonwealth English, with Britishisms such as “centre”, “colour”, “licence”, and “shop” generally prevailing over Americanisms such as “center”, “color”, “license”, and “store”. There’s also an en-GB.yml for even more classically British turns of phrase, but no en-US.yml for even basic quirks of American English. A user who knows American English will most likely configure their browser and account to prefer the en-US locale code, but they’ll see Commonwealth English instead.

For the most part, this is no problem: “colour” and “licence” merely look quirky, maybe slightly erroneous to someone unfamiliar with the dialect. But this codebase also maintains translations for a number of feature tags that would be much less intuitive to an American English speaker.

Not a single American knows what these are, even though “license” is spelled the American way:

https://github.com/openstreetmap/openstreetmap-website/blob/63f0b9257d62818d435ef6510340b9681efe547b/config/locales/en.yml#L1275

You’d have to be an OpenHistoricalMap contributor like me to know what these are:

https://github.com/openstreetmap/openstreetmap-website/blob/63f0b9257d62818d435ef6510340b9681efe547b/config/locales/en.yml#L749

99.9 out of 100 Americans would tell you this is scientific laboratory. iD calls it a “drugstore”, which is apparently closer to the intended use of this tag:

https://github.com/openstreetmap/openstreetmap-website/blob/63f0b9257d62818d435ef6510340b9681efe547b/config/locales/en.yml#L1297

For sure, Americans know what these are, but `amenity=toilets` isn’t supposed to refer to a lavatory fixture specifically, and it’s a bit impolite:

https://github.com/openstreetmap/openstreetmap-website/blob/63f0b9257d62818d435ef6510340b9681efe547b/config/locales/en.yml#L2387

Americans mostly use “shop” and “store” interchangeably, and so does this repository:

https://github.com/openstreetmap/openstreetmap-website/blob/63f0b9257d62818d435ef6510340b9681efe547b/config/locales/en.yml#L1301-L1307 https://github.com/openstreetmap/openstreetmap-website/blob/63f0b9257d62818d435ef6510340b9681efe547b/config/locales/en.yml#L1384-L1390

One way or another, we should provide more regionally appropriate terms to American English speakers, to reduce confusion about what these tags mean. I don’t have an opinion on whether we should split out a separate en-US.yml or have the Americanisms take over en.yml, since either way we’d be more likely to load the expected strings when the user says they prefer en-US in their account settings.

1ec5:
I’ve copyedited the entire English localization, fixing a number of spelling, capitalization, and grammar issues. The English localization now uses American English vocabulary and spelling throughout.

## Rationale

Ideally, the `en` localization would be some flavour of “international” English, presumably an amalgam of non-American dialects. However, most user agents follow [the industry standard](https://cldr.unicode.org/translation/translation-guide-general/default-content) of treating `en` as American English and `en-001` as international English. I don’t think we can expect Translatewiki.net to supply a separate `en-US` localization anytime soon, since [MediaWiki has essentially declined](https://phabricator.wikimedia.org/T33874) to distinguish between `en` and `en-US`. The `en-GB` localization will continue to be written in British English, but Australians, Canadians, Indians, New Zealanders, South Africans, etc. will see Americanisms until Translatewiki.net adds separate localizations for their dialects, which fortunately would be more feasible.

## Scope

These changes affect user interface strings across the entire project. Most of the changes relate to vocabulary, but I took the opportunity to fix some glaring issues, such as tag names that got mixed up.

The date format and quotation marks follow the American standard. I refrained from addressing other typographical issues, such as curly quotes, or using (or avoiding) the Oxford comma more consistently; those things can happen in a followup PR. Because many strings can refer to features anywhere in the world, I’ve also refrained from unduly applying U.S. legal norms to these strings; in fact, I’ve globalized a number of strings that had previously made regional assumptions.

## Process

I went through the file line by line, pausing every so often to avoid [semantic satiation](https://en.wikipedia.org/wiki/Semantic_satiation). Then I searched for some Britishisms like -our and -ise.

Many of the strings are applied to Nominatim search results or the map key. Mappers will undoubtedly care that these strings retain some resemblance to the terms used elsewhere in other OSM software. For the most part, I followed the preset names in [the iD tagging schema’s (American) English localization](https://github.com/openstreetmap/id-tagging-schema/blob/455d733d87a6619cd7639ac66fcec167b922c93a/interim/source_strings.yaml), figuring that a user should see something similar on the homepage and the Edit tab.

## Caveats

Some terms simply don’t have a single household word across the English speaking world – or even across the U.S. – while others have no real word in the real world because of artificial OSM tagging distinctions. At some point, id-tagging-schema started compromising by tacking on alternative names with slashes, but a more general term would be more appropriate for these strings. Since this project only identifies existing features rather than helping users choose tags for a new feature, I took some liberties in order to avoid verbose, ambivalent strings.

I’m aware that many of these changes may break longtime users’ intuition and muscle memory. Many users who don’t speak English as a first language probably still use the English localization anyways, and some may be familiar with the raw OSM tagging vocabulary that previously was reflected in these strings somewhat haphazardly. It may be necessary to scale back these changes somewhat, but I’ve started with a more maximal set of changes in order to document the tradeoffs we end up making.

~~On that note, this is a draft until I’ve gotten a chance to annotate some tricky decisions that I’d like feedback on.~~

Fixes #4975.


tomhughes:
I've only glimpsed at this but it's immediately obvious that you've gone far beyond just fixing british/american spelling variation to make editorial changes around how things should be presented and I'm not sure it's wise to do all that in one PR as it will be very hard to review.

1ec5:
Agreed, I’m writing up some comments that annotate things that don’t quite fit the bill, so we can figure out what should get split out.

config/locales/en.yml
```diff
@@ -907,122 +907,122 @@ en:
...
           speed_camera: "Speed Camera"
-          steps: "Steps"
+          steps: "Stairway"
           stop: "Stop Sign"
           street_lamp: "Street Lamp"
           tertiary: "Tertiary Road"
-          tertiary_link: "Tertiary Road"
-          track: "Track"
+          tertiary_link: "Tertiary Link Road"
+          track: "Land Management Road or Vehicular Trail"
```
1ec5, line 963: A “track”, unqualified, is something a deer forges in the woods, a trail of shoeprints in the mud, or [maybe](https://ell.stackexchange.com/a/343841) the imprint of a farm tractor cutting through the fields. Qualified, it could be a `railway=track` (train track), `leisure=track` (running track), or `highway=raceway` (racetrack), but definitely nothing to do with `highway=track`.

`highway=track` is the only tag for which I had to equivocate on the description. This tag is essentially used for two different purposes:

1. Special-purpose roads built for a productive land-based activity (forestry roads, mining roads, levee roads, fire roads, dirt roads on a farm)
2. In some regions, quasi-roads used by offroading vehicles for recreation

There isn’t a concise, obvious word for the first purpose, let alone both at the same time. id-tagging-schema uses “Track / Land-Access Road” as a grand compromise, but this is a word salad that only makes sense to someone who managed to read all of openstreetmap/id-tagging-schema#288 and the preceding mailing list threads.

I considered a number of [terms of art](https://en.wiktionary.org/wiki/term_of_art#English) for this tag. The two that stood out to me are “land management road” and “resource road”. These are established albeit obscure terms for use case (1) in the U.S. and Canada, respectively.

An American user probably won’t have encountered the phrase “land management road” per se, but they’ll hopefully recall the Bureau of Land Management and the various activities allowed on BLM lands. But they might misunderstand “management” to imply that the road is actively maintained. Appending “vehicular trail” would counter that assumption.

The current description is very long. I’m inclined to switch to the Canadian “resource road”, which is shorter. It’s also more enigmatic, so the user is less likely to assume anything in particular about the road.

config/locales/en.yml
```diff
@@ -748,14 +748,14 @@ en:
           fire_station: "Fire Station"
           food_court: "Food Court"
           fountain: "Fountain"
-          fuel: "Filling Station"
-          gambling: "Gambling"
-          grave_yard: "Grave Yard"
+          fuel: "Gas Station"
```
1ec5, line 751: Some time ago, I was showing someone OSM; they were unfamiliar with this archaic term and misread it as “filing station”. It was during tax season, so they thought it the feature was miscategorized as a government office. More recently, I was chatting with someone who snarked that it has something to do with a cafeteria.

config/locales/en.yml
```diff
@@ -1232,187 +1232,187 @@ en:
...
-          coffee: "Coffee Shop"
-          computer: "Computer Shop"
-          confectionery: "Confectionery Shop"
+          car: "Car Dealership"
+          car_parts: "Car Parts Store"
+          car_repair: "Car Repair Shop"
+          carpet: "Carpet Store"
+          charity: "Charity Store"
+          cheese: "Cheese Store"
+          chemist: "Drugstore"
```
1ec5, line 1299: This is a notorious string among the U.S. mapping community. It’s quite difficult to fathom why a pharmacist would set up shop inside a chemistry lab. On the other hand, I hope Commonwealth English speakers who wind up seeing the replacement won’t think it refers to a vendor of illicit drugs. Personally, I haven’t been a fan of equating drugstores with `shop=chemist` (openstreetmap/iD#3213), but that’s a done deal now.

config/locales/en.yml
```diff
@@ -2266,7 +2266,7 @@ en:
       welcomemat:
         url: https://welcome.openstreetmap.org/
         title: For Organizations
-        description: With an organization making plans for OpenStreetMap? Find what you need to know in the Welcome Mat.
+        description: Do you belong to an organization that wants to work with OpenStreetMap? Find what you need to know in the Welcome Mat.
```
1ec5, line 2269: The old wording might read better in some other dialect of English, but it never felt grammatical to me because it starts with a preposition phrase and lacks a subject. The new wording more closely matches the masthead on the Welcome Mat site itself.

config/locales/en.yml
```diff
@@ -2450,17 +2450,17 @@ en:
       local_chapters:
         title: Local Chapters
         about_text: |
-          Local Chapters are country-level or region-level groups that have taken the formal step of
-          establishing not-for-profit legal entities. They represent the area's map and mappers when
+          Local chapters are national or regional groups that have taken the formal step of
+          establishing nonprofit legal entities. They represent the area's map and mappers when
```
1ec5, line 2454: At least one of the local chapters, OpenStreetMap U.S., is a nonprofit organization. It is _not_ a not-for-profit organization. The two terms are [legally quite distinct](https://www.indeed.com/career-advice/career-development/non-profit-vs-not-for-profit-vs-for-profit), and mixing them up could dissuade some people from joining or donating to OSMUS.

1ec5 (commented):
As you can see, spelling is the least of our concerns. I could split out a scaled-back PR that only replaces British spellings with American spellings, but I would still advocate for a number of other changes to resolve key misunderstandings. Alternatively, we could split up this PR section by section, isolating the Nominatim and map key integrations that will probably be more contentious.

config/locales/en.yml
```diff
@@ -907,122 +907,122 @@ en:
...
+          brownfield: "Brownfield"
           cemetery: "Cemetery"
           commercial: "Commercial Area"
           conservation: "Conservation Area"
           construction: "Construction Area"
           farmland: "Farmland"
           farmyard: "Farmyard"
-          forest: "Forest"
-          garages: "Garages"
+          forest: "Managed Forest"
```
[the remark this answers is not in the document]
zekefarwell: Given that most data consumers treat `natural=wood` and `landuse=forest` as synonyms with both meaning any type of woodland, I'd suggest leaving this one alone.

config/locales/en.yml
```diff
@@ -764,61 +764,61 @@ en:
...
           pub: "Pub"
           public_bath: "Public Bath"
           public_bookcase: "Public Bookcase"
           public_building: "Public Building"
           ranger_station: "Ranger Station"
-          recycling: "Recycling Point"
+          recycling: "Recycling"
           restaurant: "Restaurant"
-          sanitary_dump_station: "Sanitary Dump Station"
+          sanitary_dump_station: "RV Toilet Disposal"
```
zekefarwell, line 789: I'm not too familiar with RVs, but it seems like this is may still called a dump station in American English, just not "sanitary" dump station. "RV dump station" seems commonly used.

config/locales/en.yml
```diff
@@ -907,122 +907,122 @@ en:
...
           road: "Road"
           secondary: "Secondary Road"
-          secondary_link: "Secondary Road"
+          secondary_link: "Secondary Link Road"
           service: "Service Road"
-          services: "Motorway Services"
+          services: "Service Area"
           speed_camera: "Speed Camera"
-          steps: "Steps"
+          steps: "Stairway"
```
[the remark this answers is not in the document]
zekefarwell: Hmm.  I'm not sure about this.  I don't find "steps" to be an unusual term, just a more general one than "stairs" or "stairway".  In my experience "stairs" or "stairway" generally refers to a set of steps with greater length than width and ascending/descending at least one floor's worth of vertical.  In cases where there are only a few steps, the steps are very wide, or are low angle I generally here these just called "steps", not "stairs" or "stairway".

config/locales/en.yml
```diff
@@ -1232,187 +1232,187 @@ en:
...
-          cheese: "Cheese Shop"
-          chemist: "Chemist"
-          chocolate: "Chocolate"
-          clothes: "Clothes Shop"
-          coffee: "Coffee Shop"
-          computer: "Computer Shop"
-          confectionery: "Confectionery Shop"
+          car: "Car Dealership"
+          car_parts: "Car Parts Store"
+          car_repair: "Car Repair Shop"
```
zekefarwell, line 1295: Maybe this is regional, but it seems to me that "Auto" or "Automotive" is far more common than "Car" in the context of parts and repair.  i.e. auto parts store, auto mechanic, automotive repair.  On the other hand, "Car Dealership" definitely sounds right to me.  I don't commonly hear or read about auto dealerships or automotive dealerships.

config/locales/en.yml
```diff
@@ -1232,187 +1232,187 @@ en:
...
           herbalist: "Herbalist"
-          hifi: "Hi-Fi Shop"
-          houseware: "Houseware Shop"
-          ice_cream: "Ice Cream Shop"
-          interior_decoration: "Interior Decoration"
-          jewelry: "Jewelry Shop"
-          kiosk: "Kiosk Shop"
-          kitchen: "Kitchen Shop"
+          hifi: "Hifi Store"
+          houseware: "Houseware store"
```
zekefarwell, line 1341:
```suggestion
          houseware: "Houseware Store"
```

config/locales/en.yml
```diff
@@ -1232,187 +1232,187 @@ en:
...
           hotel: "Hotel"
           information: "Information"
           motel: "Motel"
           museum: "Museum"
           picnic_site: "Picnic Site"
-          theme_park: "Theme Park"
-          viewpoint: "Viewpoint"
-          wilderness_hut: "Wilderness Hut"
+          theme_park: "Amusement Park"
+          viewpoint: "Vista Point"
```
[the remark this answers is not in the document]
zekefarwell: Is it more common?  I'm pretty familiar with the term "viewpoint" ([example of use by the US National Park Service](https://www.nps.gov/yose/planyourvisit/viewpoints.htm)).  "Vista" or "vista point" seem somewhat familiar, but I wouldn't say more common.

config/locales/en.yml
1ec5: There are a bunch of possibilities here – even just “Mechanic”. The main thing is that `shop=car_parts` is a “store” while `shop=car_repair` is a “shop” (as in “workshop”).

config/locales/en.yml
1ec5: Thanks, I’m not very familiar with these facilities, so I just went with id-tagging-schema on this one. I’ll change it to “RV Dump Station”.

config/locales/en.yml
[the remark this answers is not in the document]
1ec5: Hm, maybe I’m just more familiar with the highway signs that say “vista point” (hardly ever “viewpoint”). I’ll revert it back to “Viewpoint” since at least it isn’t too confusing.

config/locales/en.yml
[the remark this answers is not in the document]
1ec5: I’ve been told off for documenting `highway=steps` for just a few steps, but if it can cover that, then sure, I can revert it back to “Steps”.

systemed:
I know this is well-intentioned and addresses a genuine technical issue, but I'm a little sad about this. It's essentially squashing one of the appealing aspects of OpenStreetMap's individuality in favour of unintentionally reinforcing some low-level cultural imperialism - i.e. the notion that the lingua franca of open source software is always American English. Having the project default to British English was always one of the little markers of "OpenStreetMap is different", a bit like Ruby's social norms being essentially Japanese.

The real issue appears to be a TranslateWiki/Mediawiki bug as linked in #4975, and I wonder whether some work around the osm-website/TranslateWiki interface (about which I know precisely nothing) could bypass that.

Of course, I haven't contributed to openstreetmap-website since approximately 1776[^1] so my viewpoint means ~diddly squat~ zip here. But it's a shame nonetheless.

That aside, I would suggest splitting the PR into "pure UI changes" and "tagging-related changes", because doing things like redefining `highway=track` has mapping implications of which, as we say in [insular Latin](https://en.wikipedia.org/wiki/Insular_Latin), hic svnt dracones.

[^1]: I had to look that one up. I could tell you when the Synod of Whitby was though. Or the first Book of Common Prayer.

tomhughes:
To be fair to @1ec5 he is following through on @gravitystorm saying that en should be US and the en-GB translation should be the British version.

That said I think this is pretty much unreviewable as it goes so far beyond simple spelling and there is so much here that I would take issue with that I doubt we could ever converge on an acceptable version in a reasonable time frame. What I'm not sure of is what a reviewable version would look like.

config/locales/en.yml
woodpeck: Your change seems to claim that all local chapters are nonprofit legal entities. Even if this were true at the moment - which I'm not sure about -, it is not required that a new local chapter must be nonprofit. The [LC template agreement](https://osmfoundation.org/wiki/Local_Chapters/Template_agreement) only says they must be not-for-profit. Your change is therefore misleading. It may well be that OSM US are exceeding the requirements for a local chapter in this regard but I think that the OSM US website would be the correct place to mention that.

tomhughes: Well it already says that unless you're going to argue that `not-for-profit` and `nonprofit` somehow have different meanings?

config/locales/en.yml
```diff
@@ -1031,7 +1031,7 @@ en:
           orchard: "Orchard"
           plant_nursery: "Plant Nursery"
           quarry: "Quarry"
-          railway: "Railway"
+          railway: "Railroad Right of Way"
```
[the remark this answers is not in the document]
woodpeck: Wiki says "an area of land dedicated to train operation or support, around tracks, yards, sidings, station complexes, and ancillary man-made objects along the tracks", and I've frequently seen this used in the latter meanings - land owned by the railway but not necessarily with tracks on it. Could be a locomotive shed too. It seems to be that your "right of way" term is not covering that.

config/locales/en.yml
```diff
@@ -1187,39 +1187,39 @@ en:
...
-          accountant: "Accountant"
-          administrative: "Administration"
+          accountant: "Accountant Office"
+          administrative: "Administrative Office"
           advertising_agency: "Advertising Agency"
-          architect: "Architect"
-          association: "Association"
-          company: "Company"
+          architect: "Architect Office"
+          association: "Nonprofit Organization Office"
```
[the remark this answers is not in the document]
woodpeck: Elsewhere you linked to a document that explicitly mentioned associations as examples for "not-for-profit", rather than "nonprofit", organisations. So the qualifier "nonprofit" seems too limited for me here.

lonvia:
I'd just like to comment on the classification terms, if I may, having contributed many of them. Feel free to ignore me.

Keeping the English version quite close to the OSM tag name has always been somewhat intentional.
Quite a few of these classification terms are very technical. They are certainly not part of a standard English curriculum. Speaking for myself, I've learned many of the English terms by contributing to OSM. In fact, the key and value names have become the basis of the technical language we use to communicate among each other. I can casually mention that I've been on a tracktype-grade-3 yesterday and everybody here will immediately understand. Using translations into another English dialect on the main osm.org website is bound to cause confusion for everybody who is used to the OSM terminology which just happens to be British English. osm.org is not a end-user site, it is for mappers. So I'd expect to find "mapping speech" where it exists.

zekefarwell:
> To be fair to @1ec5 he is following through on @gravitystorm saying that en should be US and the en-GB translation should be the British version.

Link to https://github.com/openstreetmap/openstreetmap-website/issues/4975#issuecomment-2222363119 where this was stated for convenience.

> That said I think this is pretty much unreviewable as it goes so far beyond simple spelling and there is so much here that I would take issue with that I doubt we could ever converge on an acceptable version in a reasonable time frame. 

Even though I'd be very happy to see many of these changes merged, I tend to agree that this PR should get split out into more manageable chunks.  Some of these changes would be very beneficial to new US mappers as they learn OSM tagging terminology.  When a particular term has a different meaning or is very rare in US English it can lead to misunderstandings and incorrect data entry.  Many of the other changes are more subjective, changing a word that is less common in US English to one that is more common.  I'd hate to see the consequential changes delayed or not happen at all due to disagreement over less important ones.

A consequential example:  The en_GB term "camp site" translates to en_US as "campground" and the en_GB term "camp pitch" translates to en_US as "campsite".  This would be very helpful to clarify in an en_US localization of the site!

A not so consequential example:  "waste basket" vs "trash can".  Trash can is probably the more common en-US term, but waste basket is also used.  There might be a slight semantic difference and "waste basket" might be a bit out of fashion, but I don't think picking one or the other would have much impact on a new mapper's understanding.

> I've learned many of the English terms by contributing to OSM. In fact, the key and value names have become the basis of the technical language we use to communicate among each other. ... osm.org is not a end-user site, it is for mappers. So I'd expect to find "mapping speech" where it exists.

I can relate to this.  I've learned many new terms through OSM that I wouldn't have otherwise in my native US English.  A common "mapping speech" (sometimes I call this the en_OSM dialect of English 😀) is certainly crucial for communication.  However, doesn't the fact that these terms are localized into many other languages already run a bit counter to this idea that mapping speech is expected on osm.org?  I could see an argument that the mapping speech term should be shown along side the localized term but that's not currently what happens when viewing the site in a language other than English.

config/locales/en.yml
```diff
@@ -1040,109 +1040,109 @@ en:
...
           marina: "Marina"
           miniature_golf: "Miniature Golf"
-          nature_reserve: "Nature Reserve"
-          outdoor_seating: "Outdoor Seating"
+          nature_reserve: "Nature Preserve"
+          outdoor_seating: "Outdoor Seating Area"
           park: "Park"
           picnic_table: "Picnic Table"
-          pitch: "Sports Pitch"
+          pitch: "Playing Field"
```
zekefarwell, line 1069: I'm not sure if this works.  Though `leisure=pitch` is used for many "sports fields" or "playing fields", it is also used for tennis courts, basketball courts, hockey rinks, and perhaps other sports areas that aren't fields.  It seems that in  British English the term "pitch" covers all these area types but we don't have an equivalent umbrella term in US English.

zekefarwell:
> It's essentially squashing one of the appealing aspects of OpenStreetMap's individuality in favour of unintentionally reinforcing some low-level cultural imperialism - i.e. the notion that the lingua franca of open source software is always American English.

I can empathize with this.  I don't enjoy feeling like my culture is getting pushed onto others.  I'd be perfectly happy to have an en_US localization that isn't the default if the maintainers want and are able to set things up this way.

config/locales/en.yml
1ec5: In standard British English, can “pitch” actually refer to a badminton court or horseback riding arena? I thought that was “mapping speak” anyways. We can make it “Sports Field or Court” if that would be more understandable. I was trying to avoid equivocation in these strings, but this is a relatively good case for conjoining two terms.

config/locales/en.yml
```diff
@@ -907,122 +907,122 @@ en:
...
           traffic_signals: "Traffic Signals"
           trailhead: "Trailhead"
           trunk: "Trunk Road"
-          trunk_link: "Trunk Road"
-          turning_circle: "Turning Circle"
-          turning_loop: "Turning Loop"
-          unclassified: "Unclassified Road"
+          trunk_link: "Trunk Link Road"
+          turning_circle: "Cul-de-Sac"
+          turning_loop: "Cul-de-Sac"
```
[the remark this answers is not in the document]
1ec5: “Cul-de-sac” can refer to either the whole street or just its bulbous end. “Cul-de-sac street” emphasizes the broader meaning. If the end of the street has a turning hammer, we’d tag it as a `highway=turning_circle` and I think most people would handwave about it being a cul-de-sac. But if it ends abruptly without a widening, it would be a “dead end”, not a “cul-de-sac”.

config/locales/en.yml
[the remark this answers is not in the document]
1ec5: On second thought, we can probably get away with calling it “Railroad Corridor”, since the broader meaning (say, everything within 10&nbsp;miles of the tracks) would be unmappable anyways.

config/locales/en.yml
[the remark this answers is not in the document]
1ec5: Good point, “nonprofit” would be a better match for the less common `office=charity` tag. This should probably be changed in id-tagging-schema too. The problem with “association” unqualified is that it usually implies something more like `club=*` or `amenity=social_centre` in this dialect, but it’s rather difficult to give any name to these presets without making regional legal assumptions.

How about “Association Office”? At least the “office” would emphasize that this isn’t likely to be a community meeting hall of some sort.

config/locales/en.yml
1ec5: As I understand the U.S. situation, nonprofit and not-for-profit are both types of tax-exempt; nonprofit isn’t not-for-profit-deluxe. But I guess we don’t want to detract from the explanation that follows – the user will care more about the service provided than the legal status in this context. Should we avoid either term and just say it’s a legal entity, unqualified? If we need to emphasize that these entities are noncommercial, we could use that term.

1ec5:
> I know this is well-intentioned and addresses a genuine technical issue, but I'm a little sad about this. It's essentially squashing one of the appealing aspects of OpenStreetMap's individuality in favour of unintentionally reinforcing some low-level cultural imperialism - i.e. the notion that the lingua franca of open source software is always American English. Having the project default to British English was always one of the little markers of "OpenStreetMap is different", a bit like Ruby's social norms being essentially Japanese.

I completely understand. Personally, I like the occasional quirky Britishism, if it’s just a difference of spelling or diction. And I’m a _huge_ fan of the Oxford comma (sometimes odiously branded the Harvard comma). But there are many problematic tag translations that we should have some solution for. I’m here because, a few weeks ago, [we had an impromptu mapping campaign](https://en.osm.town/@watmildon/112764519783533761) around that most American of institutions, the drive-through liquor store, and a bunch of us saw “Off License” and thought it was a bug. Maybe a misplaced license agreement string?

I honestly have no problem with relegating the Americanisms to en-US.yml if we can ensure that works with our translation management system. The CLDR default content locales cause problems for other languages too, like [Portuguese](https://community.openstreetmap.org/t/utilizacao-de-etiqueta-name-pt-br/100139/10). But Translatewiki.net has taken a hard stance that they want to follow Standards. One workaround would be to maintain the American English localization manually outside of Translatewiki.net, but the maintainers would have to review updates manually too.

> Keeping the English version quite close to the OSM tag name has always been somewhat intentional.
> Quite a few of these classification terms are very technical. They are certainly not part of a standard English curriculum. Speaking for myself, I've learned many of the English terms by contributing to OSM. In fact, the key and value names have become the basis of the technical language we use to communicate among each other. I can casually mention that I've been on a tracktype-grade-3 yesterday and everybody here will immediately understand. Using translations into another English dialect on the main osm.org website is bound to cause confusion for everybody who is used to the OSM terminology which just happens to be British English. osm.org is not a end-user site, it is for mappers. So I'd expect to find "mapping speech" where it exists.

If that’s what we want, then this diff will get even bigger. For better or worse, this localization already taking liberties with some strings to sound more British and less Mapper, such as `shop=alcohol` as “Off License”, `amenity=biergarten` as “Beer Garden”, and `highway=bus_guideway` as “Guided Bus Lane”. If anyone is depending on the search results to learn the raw tags, they’re going to be in for a surprise with `railway=halt` as “Train Stop” and `railway=switch` as “Railway Points” (which I changed to “Switch”).

Maybe we could add a tooltip to each row of the search results and map key that indicates the raw tag, just like when you hover over a linked element in the element inspector. That could have the educational effect you’re expecting – even if the user speaks Dutch.

landuse=meadow

> That said I think this is pretty much unreviewable as it goes so far beyond simple spelling and there is so much here that I would take issue with that I doubt we could ever converge on an acceptable version in a reasonable time frame. What I'm not sure of is what a reviewable version would look like.

Yes, I really appreciate all of you taking the time to look at this draft so far, but I really didn’t intend this to be a final product to ram through. This maximalist changeset gives native speakers an opportunity to discuss individual changes line by line, and then we can figure out how to stage the changes in more digestible chunks for the maintainers to review. Many of the changes are inessential, as @zekefarwell points out: if we can’t identify any strong reason to change a string, we can leave it alone. We could decide that “shop” gives the site a more lighthearted feel compared to “store”, though a few “dealerships” and “parlors” might still be appropriate.

How about we start with some things that are clearly bugs, regardless of dialect, such as `man_made=dolphin` being mislabeled as “Mooring Post”? Then maybe some things like `tourism=camp_site` and `highway=unclassified` that currently hew to OSM English but are likely contributing to mistagging. To me, the -ours and -ises and single quotation marks are the lowest priority of all.

config/locales/en.yml
```diff
@@ -764,61 +764,61 @@ en:
...
           social_facility: "Social Facility"
           studio: "Studio"
           swimming_pool: "Swimming Pool"
-          taxi: "Taxi"
+          taxi: "Taxi Stand"
           telephone: "Public Telephone"
-          theatre: "Theatre"
-          toilets: "Toilets"
+          theatre: "Theater"
+          toilets: "Restroom"
```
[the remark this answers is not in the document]
1ec5: On the other hand, if we want to lean into British quirkiness, how about “Water Closet”? 😉

config/locales/en.yml
zekefarwell: The [Wikipedia page on pitches](https://en.wikipedia.org/wiki/Pitch_(sports_field)) includes badminton and various other court sports.  Of course the article "needs additional citations" so who knows.  The [Cambridge ](https://dictionary.cambridge.org/dictionary/english/pitch)& [Collins ](https://www.collinsdictionary.com/dictionary/english/pitch)dictionaries both define "pitch" as an area with painted lines or markings for playing a sport.  This sounds to me like a court would fit this description since a specific surface type is not stated.  They both include hockey (played on an ice rink not a field) as an example.  I'm not sure exactly what the term "pitch" covers in British English, but it does seem to have broader meaning than "playing field" which typically indicates a grass or turf surface.  I don't think I've ever heard a basketball or tennis court referred to as a playing field, but maybe it is technically still correct when discussing the rules of the sport and the "field of play".

config/locales/en.yml
```diff
@@ -1232,187 +1232,187 @@ en:
...
-          postcode: "Postcode"
+          postcode: "Postal Code"
           quarter: "Quarter"
           region: "Region"
           sea: "Sea"
           square: "Square"
           state: "State"
           subdivision: "Subdivision"
-          suburb: "Suburb"
+          suburb: "Urban District"
```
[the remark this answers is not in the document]
jacobwhall: I can confirm this is confusing for an American English speaker: I didn't know the difference until reading this comment.

config/locales/en.yml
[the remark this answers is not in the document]
jacobwhall: Maybe `Organization Office`?

config/locales/en.yml
```diff
@@ -1187,37 +1187,37 @@ en:
...
           newspaper: "Newspaper Office"
           ngo: "NGO Office"
-          notary: "Notary"
+          notary: "Notary Office"
           religion: "Religious Office"
           research: "Research Office"
-          tax_advisor: "Tax Advisor"
-          telecommunication: "Telecommunication Office"
+          tax_advisor: "Tax Advisor Office"
+          telecommunication: "Telecom Office"
```
jacobwhall, line 1216: I think that in this case `Telecommunication Office` is just as clear, and better matches the tag

config/locales/en.yml
```diff
@@ -1232,187 +1232,187 @@ en:
...
+          appliance: "Appliances Store"
+          art: "Art Store"
+          baby_goods: "Baby Goods Store"
+          bag: "Bag Store"
           bakery: "Bakery"
-          bathroom_furnishing: "Bathroom Furnishing"
-          beauty: "Beauty Shop"
-          bed: "Bedding Products"
-          beverages: "Beverages Shop"
+          bathroom_furnishing: "Bathroom Furnishing Store"
```
jacobwhall, line 1284: I recommend `Bathroom Design Store` to match `shop=kitchen` below.

config/locales/en.yml
```diff
@@ -685,46 +685,46 @@ en:
...
-          animal_boarding: "Animal Boarding"
+          animal_boarding: "Animal Boarding Facility"
           animal_shelter: "Animal Shelter"
-          arts_centre: "Arts Centre"
+          arts_centre: "Arts Center"
           atm: "ATM"
           bank: "Bank"
           bar: "Bar"
-          bbq: "BBQ"
+          bbq: "Barbecue"
```
jacobwhall, line 714: I think `BBQ` is just as clear, and it matches the tag.

config/locales/en.yml
```diff
@@ -829,72 +829,72 @@ en:
...
           duplex: "Duplex House"
-          farm: "Farm House"
-          farm_auxiliary: "Auxiliary Farm House"
+          farm: "Farmhouse"
+          farm_auxiliary: "Farm Building"
           garage: "Garage"
           garages: "Garages"
           greenhouse: "Greenhouse"
-          hangar: "Hangar"
+          hangar: "Hangar Building"
```
jacobwhall, line 851: I usually do not hear fellow American English speakers say "building" after "hangar."

config/locales/en.yml
```diff
@@ -829,72 +829,72 @@ en:
...
           hospital: "Hospital Building"
           hotel: "Hotel Building"
           house: "House"
           houseboat: "Houseboat"
           hut: "Hut"
           industrial: "Industrial Building"
           kindergarten: "Kindergarten Building"
-          manufacture: "Manufacturing Building"
+          manufacture: "Industrial Production Building"
```
jacobwhall, line 859: I recommend `Factory Building`

config/locales/en.yml
```diff
@@ -829,72 +829,72 @@ en:
...
           dressmaker: "Dressmaker"
           electrician: "Electrician"
-          electronics_repair: "Electronics Repair"
+          electronics_repair: "Electronics Repair Shop"
           gardener: "Gardener"
           glaziery: "Glaziery"
-          handicraft: "Handicraft"
-          hvac: "HVAC Craft"
-          metal_construction: "Metal Constructioner"
+          handicraft: "Handicraft Workspace"
```
jacobwhall, line 895: [This Wikipedia page](https://en.wikipedia.org/wiki/Handicraft) offers some alternative names. `Handcrafting Workspace` or `Artisan` would sound more natural to my ear.

config/locales/en.yml
```diff
@@ -907,122 +907,122 @@ en:
...
           abandoned: "Abandoned Highway"
-          bridleway: "Bridleway"
-          bus_guideway: "Guided Bus Lane"
+          bridleway: "Bridle Path"
+          bus_guideway: "Guided Busway"
           bus_stop: "Bus Stop"
-          construction: "Highway under Construction"
-          corridor: "Corridor"
+          construction: "Highway Under Construction"
+          corridor: "Indoor Corridor"
```
jacobwhall, line 928: I recommend `Hallway`

config/locales/en.yml
```diff
@@ -907,122 +907,122 @@ en:
...
-          motorway_junction: "Motorway Junction"
-          motorway_link: "Motorway Road"
+          motorway: "Controlled-Access Highway"
+          motorway_junction: "Highway Exit"
+          motorway_link: "Highway Ramp"
           passing_place: "Passing Place"
           path: "Path"
-          pedestrian: "Pedestrian Way"
-          platform: "Platform"
+          pedestrian: "Pedestrian Mall"
```
jacobwhall, line 944: I've never heard "Pedestrian Mall" in the U.S. Perhaps `Pedestrian Road` would fit better.

config/locales/en.yml
```diff
@@ -1040,109 +1040,109 @@ en:
...
           vineyard: "Vineyard"
-          "yes": "Landuse"
+          "yes": "Land Use Area"
         leisure:
-          adult_gaming_centre: "Adult Gaming Centre"
+          adult_gaming_centre: "Adult Gaming Center"
           amusement_arcade: "Amusement Arcade"
           bandstand: "Bandstand"
           beach_resort: "Beach Resort"
           bird_hide: "Bird Hide"
```
jacobwhall, line 1049: This should be `Bird Blind` in en-US.

jacobwhall, line 1049: This should be `Bird Blind` in en-US.

config/locales/en.yml
jacobwhall: "Sports Pitch" is a confusing term to me as an American English speaker, even if there are nuanced differences I think `Playing Field` is clearer overall.

config/locales/en.yml
```diff
@@ -1232,187 +1232,187 @@ en:
...
           hearing_aids: "Hearing Aids"
           herbalist: "Herbalist"
-          hifi: "Hi-Fi Shop"
-          houseware: "Houseware Shop"
-          ice_cream: "Ice Cream Shop"
-          interior_decoration: "Interior Decoration"
-          jewelry: "Jewelry Shop"
-          kiosk: "Kiosk Shop"
-          kitchen: "Kitchen Shop"
+          hifi: "Hifi Store"
```
jacobwhall, line 1340: I recommend either [Hi-Fi or HiFi](https://en.wikipedia.org/wiki/High_fidelity)

config/locales/en.yml
1ec5: Yeah, this came from id-tagging-schema, which needs to distinguish between the presets for `aeroway=hangar` and `building=hangar`. But that distinction is probably less important in the context of these search results.

config/locales/en.yml
1ec5: “Pedestrian mall” has historically been the preferred term in American English, as seen in [Google Books Ngram Viewer](https://books.google.com/ngrams/graph?content=pedestrian+mall%2Cpedestrian+precinct%2Cpedestrian+road%2Cpedestrian+zone&year_start=1920&year_end=2022&corpus=en-US&smoothing=3). However, I appear to be dating myself: “pedestrian street” is more popular in the last several years, and “pedestrian zone” is also becoming common. That’s the term [Wikipedia](https://en.wikipedia.org/wiki/Pedestrian_zone) uses too, to avoid favoring American/Australian terms over the British term “pedestrian precinct”. “Pedestrian road” appears to be very rare.

A screenshot of Google Books Ngram Viewer showing “pedestrian mall” as by far the preferred term in American English from 1940, peaking in the 1970s, until “pedestrian street” overtook it in 2016. “Pedestrian zone” also ties with it as of 2021. “Pedestrian precinct” and “pedestrian road” are virtually unknown in this dialect.

I think “pedestrian street” would be intuitive and pretty close to your suggestion. “Pedestrian zone” would be problematic because it too strongly suggests an area, whereas this tag can validly appear on a way.

config/locales/en.yml
1ec5: The [`building=manufacture`](https://wiki.openstreetmap.org/wiki/Tag:building%3Dmanufacture) documentation says this is narrower than `building=industrial` but broader than [`building=factory`](https://wiki.openstreetmap.org/wiki/Tag:building%3Dfactory), encompassing storage buildings. I’m not sure why id-tagging-schema felt the need to call it something longwinded; “Manufacturing Building” seems OK given the other overlapping terms.

config/locales/en.yml
1ec5: I was under the impression that “BBQ” only refers to a cookout or the dish, whereas the equipment would only be called a “barbecue” or a “grill”. But maybe this is too pedantic? Ironically, for the dish, `cuisine=bbq` has been deprecated in favor of `cuisine=barbecue`.

config/locales/en.yml
jacobwhall: Yes, now that you say it I think "Pedestrian Street" is even better. Good idea re: ngram!

config/locales/en.yml
jacobwhall: Oh wow I've never considered that distinction before but you may be right. "Barbecue" may indeed be more appropriate for the grill itself. I do not have a strong preference here.

config/locales/en.yml
```diff
@@ -1435,19 +1435,19 @@ en:
...
+        level2: "International Border"
+        level3: "Regional Administrative Boundary"
+        level4: "Regional Administrative Boundary"
+        level5: "Regional Administrative Boundary"
+        level6: "Local Administrative Boundary"
+        level7: "Local Administrative Boundary"
+        level8: "Local Administrative Boundary"
+        level9: "Sublocal Administrative Boundary"
+        level10: "Sublocal Administrative Boundary"
+        level11: "Sublocal Administrative Boundary"
```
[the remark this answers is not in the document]
1ec5: This is superseded by #5493.

config/locales/en.yml
```diff
@@ -764,61 +764,61 @@ en:
...
+          waste_dump_site: "Dump"
           watering_place: "Watering Place"
-          water_point: "Water Point"
-          weighbridge: "Weighbridge"
+          water_point: "RV Drinking Water"
+          weighbridge: "Truck Scale"
           "yes": "Amenity"
         boundary:
-          aboriginal_lands: "Aboriginal Lands"
+          aboriginal_lands: "Indigenous Boundary"
```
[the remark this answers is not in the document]
andrewharvey: Which is a bit ironic considering we don't use the tag here https://wiki.openstreetmap.org/wiki/Australian_Tagging_Guidelines/Australia%27s_First_Peoples#Aboriginal_Reservations and claim it's better suited for use in the US.

andrewharvey:
Just a note that if this is merged we'll need to do `en_AU` as currently most of the `en` ones work okay for Australian English, but that won't be the case if this is merged.

matkoniecz:
@1ec5 given that it is stuck since 2024...

maybe making PR with subset of changes, for example starting with ones changing more UK-specific ones and confusing otherwise, if any, may be helpful?

config/locales/en.yml
```diff
@@ -243,31 +237,58 @@ en:
...
         full: Full note
+  account:
+    deletions:
+      show:
+        title: Delete My Account
+        warning: Warning! The account deletion process is final and cannot be reversed.
+        delete_account: Delete Account
+        delete_introduction: "You can delete your OpenStreetMap account using the button below. Please note the following details:"
+        delete_profile: Your profile information, including your avatar, description and home location, will be removed.
+        delete_display_name: Your display name will be removed and can be reused by other accounts.
```
matkoniecz, line 257: is it really US/UK change?

maybe it should be dropped and if needed done via separate PR?

config/locales/en.yml
```diff
@@ -388,12 +358,20 @@ en:
...
+      type:
+        node: node
+        way: way
+        relation: relation
+        changeset: changeset
+        note: note
     timeout:
       title: Timeout Error
-      sorry: "Sorry, the data for the %{type} with the id %{id} took too long to retrieve."
+      sorry: "Sorry, the data for the %{type} with the ID %{id} took too long to retrieve."
```
matkoniecz, line 374: this and rss->RSS are not really US-English specific (probably), so are out of scope of this PR, right?

config/locales/en.yml
```diff
@@ -732,46 +685,46 @@ en:
...
         aeroway:
-          aerodrome: "Aerodrome"
+          aerodrome: "Airport"
           airstrip: "Airstrip"
           apron: "Airport Apron"
           gate: "Airport Gate"
           hangar: "Hangar"
           helipad: "Helipad"
-          holding_position: "Holding Position"
+          holding_position: "Aircraft Holding Position"
```
matkoniecz, line 699: this is also probably not really UK vs US ? 

if helpful can be done as a separate PR, right?

config/locales/en.yml
matkoniecz: this looks like a case where UK term may more confusing for general international audience

config/locales/en.yml
```diff
@@ -732,46 +685,46 @@ en:
...
           bank: "Bank"
           bar: "Bar"
-          bbq: "BBQ"
+          bbq: "Barbecue"
           bench: "Bench"
-          bicycle_parking: "Cycle Parking"
-          bicycle_rental: "Cycle Rental"
-          bicycle_repair_station: "Bicycle Repair Station"
-          biergarten: "Beer Garden"
+          bicycle_parking: "Bicycle Parking"
```
matkoniecz, line 716: this also looks like a case where UK term may more confusing for general international audience than US one

PR #5030, opened 2024-07-29
Define fixed size for trace images

mmd-osm:
Since [loading trace images](https://www.openstreetmap.org/traces) from AWS S3 might take a bit, users are currently experiencing a re-rendering of the page. I found this a bit annoying, and added a fixed 50x50 size for the trace png now. With the change in place, the re-rendering of the page no longer happens. The trace png is shown initially as a blank spot.

#### First rendering (w/o images)
image

#### Second rendering (w/ images)
image

#### Blank spot with 50x50 shown as long as image hasn't been loaded yet. No re-rendering of the page
image

### DIFF
--- a/app/helpers/trace_helper.rb
+++ b/app/helpers/trace_helper.rb
@@ -2,4 +2,12 @@ module TraceHelper
   def link_to_tag(tag)
     link_to(tag, :tag => tag, :page => nil)
   end
+
+  def trace_icon(trace, options = {})
+    options[:class] ||= "trace_image"
+    options[:alt] ||= ""
+
+    image_tag trace_icon_path(trace.user, trace),
+              options.merge(:size => 50)
+  end
 end
--- a/app/views/traces/_trace.html.erb
+++ b/app/views/traces/_trace.html.erb
@@ -2,7 +2,7 @@
   <td>
     <% if Settings.status != "gpx_offline" %>
       <% if trace.inserted %>
-        <%= link_to image_tag(trace_icon_path(trace.user, trace), :alt => "", :class => "trace_image"),
+        <%= link_to trace_icon(trace),
                     show_trace_path(trace.user, trace),
                     :class => "d-inline-block" %>
       <% else %>


app/helpers/trace_helper.rb
```diff
@@ -2,4 +2,14 @@ module TraceHelper
...
+
+  def trace_image(trace, options = {})
+    options[:class] ||= "trace_image"
+    options[:alt] ||= ""
+
+    link_to(image_tag(trace_icon_path(trace.user, trace),
+                      options.merge(:width => 50, :height => 50)),
+            show_trace_path(trace.user, trace),
+            :class => "d-inline-block")
+  end
```
AntonKhorev, line 14: Do we really need a helper for this? If it's supposed to be reusable, why `d-inline-block` is baked into the link?

Why the name `trace_image`? There are two trace images, the name doesn't tell which one this is. Why "image" if it returns a link?

mmd-osm: I renamed trace_image to trace_preview_image. Not sure if there's a better name for "preview". Then I moved everything which doesn't belong there back to _trace.html.erb. 

The reusable is following the pattern in user_helper.rb. I started adding options[:class], options[:alt], etc in _trace.html.erb, but this looked a bit ugly and cluttered the html code. That's why I moved it to a helper.

mmd-osm: `trace_icon` sounds fine... changed

app/helpers/trace_helper.rb
```diff
@@ -2,4 +2,12 @@ module TraceHelper
   def link_to_tag(tag)
     link_to(tag, :tag => tag, :page => nil)
   end
+
+  def trace_icon(trace, options = {})
+    options[:class] ||= "trace_image"
+    options[:alt] ||= ""
+
+    image_tag trace_icon_path(trace.user, trace),
+              options.merge(:width => 50, :height => 50)
```
AntonKhorev, line 11:
```suggestion
              options.merge(:size => 50)
```

`user_helper.rb` is partially using `:size`, I think we were moving towards it from `:width`/`:height` for square images.

mmd-osm: Yes, :size is ok, the resulting css will be the same as before. Changed.

AntonKhorev:
Thanks, merged.

Do you think the same should be done on trace show pages with bigger images? They all are 250x250, right?

PR #5031, opened 2024-07-29
GPX Import email in text format

### LINKED ISSUE #5025: 'GPX Import success' e-mail contains only HTML and not a text part
### URL

n/a

The '[OpenStreetMap] GPX Import success' e-mail contains only HTML and not a text part.

The message consists of:

Screenshot 2024-07-26 at 17 24 32

_No response_

mmd-osm:
Fixes #5025

### DIFF
--- a/app/views/user_mailer/_gpx_description.text.erb
+++ b/app/views/user_mailer/_gpx_description.text.erb
@@ -1,0 +1,8 @@
+<% trace_name = @trace_name %>
+<% trace_description = @trace_description %>
+<% if @trace_tags.length > 0 %>
+  <% tags = @trace_tags.map { |trace_tag| trace_tag.tag }.join(", ") %>
+  <%= t ".description_with_tags", :trace_name => trace_name, :trace_description => trace_description, :tags => tags %>
+<% else %>
+  <%= t ".description_with_no_tags", :trace_name => trace_name, :trace_description => trace_description %>
+<% end %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1626,11 +1626,14 @@ en:
       befriend_them: "You can also add them as a friend at %{befriendurl}."
       befriend_them_html: "You can also add them as a friend at %{befriendurl}."
     gpx_description:
+      description_with_tags: "It looks like your GPX file %{trace_name} with the description %{trace_description} and the following tags: %{tags}"
       description_with_tags_html: "It looks like your GPX file %{trace_name} with the description %{trace_description} and the following tags: %{tags}"
+      description_with_no_tags: "It looks like your GPX file %{trace_name} with the description %{trace_description} and no tags"
       description_with_no_tags_html: "It looks like your GPX file %{trace_name} with the description %{trace_description} and no tags"
     gpx_failure:
       hi: "Hi %{to_user},"
       failed_to_import: "failed to import. Here is the error:"
+      more_info: "More information about GPX import failures and how to avoid them can be found at %{url}."
       more_info_html: "More information about GPX import failures and how to avoid them can be found at %{url}."
       import_failures_url: "https://wiki.openstreetmap.org/wiki/GPX_Import_Failures"
       subject: "[OpenStreetMap] GPX Import failure"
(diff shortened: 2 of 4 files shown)


app/views/user_mailer/_gpx_description.text.erb
```diff
@@ -0,0 +1,8 @@
+<% trace_name = @trace_name %>
+<% trace_description = @trace_description %>
+<% if @trace_tags.length > 0 %>
+  <% tags = safe_join @trace_tags.map { |trace_tag| trace_tag.tag }, ", " %>
+  <%= t ".description_with_tags_html", :trace_name => trace_name, :trace_description => trace_description, :tags => tags %>
```
AntonKhorev, line 5: Tags are html-encoded: `&` to `&amp;`, `<` to `&lt;` etc.

mmd-osm: This should be fixed now. I also added some non _html translation strings, just in case. This is similar to the other user_mailer texts.

config/locales/en.yml
```diff
@@ -1626,11 +1626,14 @@ en:
       befriend_them: "You can also add them as a friend at %{befriendurl}."
       befriend_them_html: "You can also add them as a friend at %{befriendurl}."
     gpx_description:
+      description_with_tags: "It looks like your GPX file %{trace_name} with the description %{trace_description} and the following tags: %{tags}"
       description_with_tags_html: "It looks like your GPX file %{trace_name} with the description %{trace_description} and the following tags: %{tags}"
```
AntonKhorev, line 1630: In the html version the most important link is the link to the trace page (`/user/:user/traces/:id`), which is inserted at `%{trace_name}`. Plaintext version doesn't have this link, `%{trace_name}` is just the filename. Maybe we can rewrite `description_with_tags` and `description_with_no_tags` to include the url? Then they won't be just copies without the `_html` suffixes.

mmd-osm: Yes, good point, I have added `trace_url` now. The html version somehow assumes this URL could be null, which is not clear to me.

AntonKhorev: This can happen on failure when `@trace_url` is not set. You'll get a plaintext message like "... It looks like your GPX file Trace 1.gpx at  with the description ...". I guess we need two plaintext locale strings, with an url and without it.

mmd-osm: I added some new text "The trace is available for download at %{trace_url}" that is only shown in case of success, and removed the %{trace_url} again from description_with(_no)?_tags. I think this is a bit cleaner, and also easier to read as an email.

AntonKhorev: The logic of this email, I assume, is not to tell the user how to download a file, but to tell which file was processed. The filename is likely generated by whatever track writing software and may not be easily recognizable, that's why we add an url. The user doesn't need to download the file because they've just uploaded it.

But if you think that your changes are easier to read, shouldn't the html version be also modified in this manner?

mmd-osm: The download URL that is shown in the email is controlled by the Rails port, not the user uploading the trace. Here's an example:

```
Content-Type: text/plain;
 charset=UTF-8
Content-Transfer-Encoding: 7bit

Hi mmd2,

  It looks like your GPX file Track_2014_06_13_184543.gpx with the description dsfasf and the following tags: d

loaded successfully with 3427 out of a possible 3427 points.

The trace is available for download at http://openstreetmap.example.com/user/mmd2/traces/64

All your successfully uploaded GPX traces can be found at http://openstreetmap.example.com/traces/mine

```

mmd-osm: Updated text email:

```
Content-Type: text/plain;
 charset=UTF-8
Content-Transfer-Encoding: 7bit

Hi mmd2,

  It looks like your GPX file Track_2014_06_13_184543.gpx with the description dsfsdf and the following tags: sdfsdf

loaded successfully with 3427 out of a possible 3427 points.

Your trace is available at http://openstreetmap.example.com/user/mmd2/traces/65

All your successfully uploaded GPX traces can be found at http://openstreetmap.example.com/traces/mine

```

AntonKhorev:
Thanks, merged.

PR #5033, opened 2024-07-30
Infinite scroll - Issues page

nertc:
This PR addresses heavy calls and long load time of the Issues page. Problem was mentioned in https://github.com/openstreetmap/openstreetmap-website/pull/4990 by @AntonKhorev.

Added infinite scroll functionality. On the first load, there will be only loader and after that, every time loader will be shown on the screen, new 50 issues will be called and displayed. In terms of performance, locally I have 1000 users and 1000 issues. First load time was decreased from ~8s to ~1s, and after that every call for the next 50 issues takes ~150ms. In real environment, it will have much better improvement, as instead of searching for 30000 issues, it will take only 50 (in theory, the difference should be 30x better in the real environment, then it is on my local machine).

Video:

https://github.com/user-attachments/assets/33254baf-f553-4a27-8b3e-9ce651a08c53

Issues not found:

image

User not found:

image

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -57,7 +57,7 @@ class Ability
 
         if user.moderator?
           can [:hide, :unhide], [DiaryEntry, DiaryComment]
-          can [:index, :show, :resolve, :ignore, :reopen], Issue
+          can [:index, :show, :resolve, :ignore, :reopen, :page], Issue
           can :create, IssueComment
           can [:new, :create, :edit, :update, :destroy], Redaction
           can [:new, :edit, :create, :update, :revoke, :revoke_all], UserBlock
@@ -65,7 +65,7 @@ class Ability
 
         if user.administrator?
           can [:hide, :unhide], [DiaryEntry, DiaryComment]
-          can [:index, :show, :resolve, :ignore, :reopen], Issue
+          can [:index, :show, :resolve, :ignore, :reopen, :page], Issue
           can :create, IssueComment
           can [:set_status, :destroy, :index], User
           can [:grant, :revoke], UserRole
--- a/app/assets/javascripts/issues.js
+++ b/app/assets/javascripts/issues.js
@@ -1,0 +1,48 @@
+$(document).ready(function () {
+  let beforeInput;
+
+  function fetchIssues() {
+    const data = {
+      limit: 50
+    };
+
+    if (beforeInput) {
+      data.before = beforeInput.value;
+      beforeInput.remove();
+    }
+
+    $.ajax({
+      oauth: true,
+      url:
+        window.location.origin +
+        window.location.pathname +
+        "/page" +
+        window.location.search,
+      data,
+      cache: false
+    }).done(function (result) {
+      $("#reports_table").append($(result));
+      beforeInput = $("input[name='before']")[0];
+
+      if (!beforeInput) {
+        $("#reports_loading").remove();
+      }
+    });
+  }
+
+  let options = {
+    root: null,
+    threshold: 0.1
+  };
+
+  function handleIntersect(entries) {
+    entries.forEach((entry) => {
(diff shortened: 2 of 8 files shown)


AntonKhorev:
You could have addressed it by using the existing pagination code. But again see https://github.com/openstreetmap/openstreetmap-website/pull/4990#issuecomment-2248072093. Yes, it may involve a browser built-in search.

nertc:
> Yes, it may involve a browser built-in search.

There shouldn't be any need of built-in browser search on the issues page. If there is any, we can add needed search fields to make website responsible for filtering in all the user wanted way.

tomhughes:
Do we really want to do infinite scroll for this when everything else uses pagination, and infinite scroll needs client side code which we prefer to avoid.

nertc:
@tomhughes 
- Pagination adds extra clicks to navigate through issues. Regarding lists, for me, infinite scroll is a much smoother experience (+ it doesn't change current user behavior, they just need to scroll, as it was before, without adding additional complexity).
- Also, I think, it will be good, if we have at least one page, which is done using infinite scroll, and if in the future, we decide that any page needs to implement infinite scroll, there will be a working example (therefore, it will reduce future development time).
- In addition, rendering search fields needs extra query and filter for users, using infinite scroll and rendering only the list part reduces time and server load.
- This implementation uses not complex JS, and provides the same benefits that single-page applications have, while everything is still rendered in the back end side.

AntonKhorev:
> Pagination adds extra clicks to navigate through issues

Which are faster to do than scrolling.

AntonKhorev:
> rendering search fields needs extra query and filter for users, using infinite scroll and rendering only the list part reduces time and server load

Do you think this can't be done with pagination?

nertc:
> > Pagination adds extra clicks to navigate through issues
> 
> Which are faster to do than scrolling.

Do you mean development time, or the user interaction time?

AntonKhorev:
> > > Pagination adds extra clicks to navigate through issues
> > 
> > Which are faster to do than scrolling.
> 
> Do you mean development time, or the user interaction time?

User interaction time, but actually both.

nertc:
> > > > Pagination adds extra clicks to navigate through issues
> > > 
> > > 
> > > Which are faster to do than scrolling.
> > 
> > 
> > Do you mean development time, or the user interaction time?
> 
> User interaction time, but actually both.

Development time is definitely less with the pagination, as we have working examples and template for it. But in terms of user interaction time, only scrolling introduces less complexity, than scrolling and then clicking (in this case, clicking part is added to the scrolling part).

The benefit I see for the pagination is that if a user opens report page of the issue X and then decides to open report page of the issue X+1, he will have to scroll all X issues once more, while with pagination it will be redirected to the list page where X and X+1 issues are shown. I am not neither moderator nor administrator, so I don't know how often there will be cases of such behavior. If it is often, I think, pagination will be much better solution.

AntonKhorev:
> than scrolling and then clicking

You won't need to scroll if the buttons are at the top of the page.

nertc:
> You won't need to scroll if the buttons are at the top of the page.

It's a good idea, if a user tries to find a specific issue/report in the list, but if he wants to take a glance at the list or only open issues with the high report count, then he still needs "scroll + click" while pagination is used. Overall, both strategies have own benefits and drawbacks. If pagination is our common approach, users are more used to it and they have no problem with it, I'll change this PR to implement pagination instead of the infinite scroll.

nertc:
As we decided to implement pagination instead of the infinite scroll, I am closing this PR. Separate PR https://github.com/openstreetmap/openstreetmap-website/pull/5057 was created about pagination.

nertc:
As we decided to implement pagination instead of the infinite scroll, I am closing this PR. Separate PR https://github.com/openstreetmap/openstreetmap-website/pull/5057 was created about pagination.

PR #5034, opened 2024-07-30
Add planet link and explanation to robots.txt

grischard:
Recent scraping incidents have made think that a friendly pointer to planet.osm.org probably wouldn't hurt.


nenad-vujicic:
Perhaps you could squash these two commits into one, please, before approving?

Thanks,
Nenad.

gravitystorm:
Merged, thanks!

PR #5035, opened 2024-07-31
Trace picture with fixed size

mmd-osm:
Follow up for https://github.com/openstreetmap/openstreetmap-website/pull/5030#issuecomment-2260651720


AntonKhorev:
Merged, thanks.

PR #5038, opened 2024-08-01
Trace import: remove activerecord-import gem

### LINKED ISSUE #4994: Retire activerecord-import gem?
Back in the days we've added this gem to speed up trace points imports. https://blog.saeloun.com/2022/07/26/rails-6-insert-all/ says this approach is outdated, because Rails 6 adds  `insert_all` support out of the box.

mmd-osm:
This PR removes the activerecord-import dependency and changes the tracepoint import to use `import_all!` instead.

I hope I didn't miss anything here. In particular `Tracepoint.insert_all!(tracepoints.map(&:attributes))` would need a review if that's the proper way to import Tracepoint instances. 

Fixes #4994

### DIFF
--- a/app/models/trace.rb
+++ b/app/models/trace.rb
@@ -230,16 +230,17 @@ class Trace < ApplicationRecord
           tp.timestamp = point.timestamp
           tp.gpx_id = id
           tp.trackid = point.segment
+          tp.validate!
           tracepoints << tp
         end
 
-        # Run the before_save and before_create callbacks, and then import them in bulk with activerecord-import
+        # Run the before_save and before_create callbacks, and then do a bulk insert
         tracepoints.each do |tp|
           tp.run_callbacks(:save) { false }
           tp.run_callbacks(:create) { false }
         end
 
-        Tracepoint.import!(tracepoints)
+        Tracepoint.insert_all!(tracepoints.map(&:attributes)) # rubocop:disable Rails::SkipsModelValidations
       end
 
       if gpx.actual_points.positive?
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,6 @@ gem "file_exists"
 
 # Load rails plugins
 gem "actionpack-page_caching", ">= 1.2.0"
-gem "activerecord-import"
 gem "active_record_union"
 gem "bootstrap", "~> 5.3.2"
 gem "bootstrap_form", "~> 5.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -80,8 +80,6 @@ GEM
       activemodel (= 7.2.1.2)
       activesupport (= 7.2.1.2)
       timeout (>= 0.4.0)
-    activerecord-import (1.8.1)
-      activerecord (>= 4.2)
     activestorage (7.2.1.2)
       actionpack (= 7.2.1.2)
       activejob (= 7.2.1.2)
@@ -667,7 +665,6 @@ DEPENDENCIES
   aasm
   actionpack-page_caching (>= 1.2.0)
   active_record_union
-  activerecord-import
   addressable (~> 2.8)
   annotate
   argon2


.rubocop.yml
```diff
@@ -71,6 +71,7 @@ Rails/SkipsModelValidations:
   Exclude:
     - 'db/migrate/*.rb'
     - 'app/controllers/users_controller.rb'
+    - 'app/models/trace.rb'
```
tomhughes, line 74: Why do we need to exclude this?

app/models/trace.rb
```diff
@@ -233,13 +233,13 @@ def import
...
 
-        # Run the before_save and before_create callbacks, and then import them in bulk with activerecord-import
+        # Run the before_save and before_create callbacks, and then do a bulk insert
         tracepoints.each do |tp|
           tp.run_callbacks(:save) { false }
           tp.run_callbacks(:create) { false }
         end
 
-        Tracepoint.import!(tracepoints)
+        Tracepoint.insert_all!(tracepoints.map(&:attributes))
```
tomhughes, line 242: I've done some testing and I think this looks right - my main concern is the timestamp column as `attributes` returns it in string form.

It would probably be a good idea to add a test to the trace model tests that imports `a.gpx` and checks that the one point it imports has the expected values for each attribute as we don't seem to have one that checks the imported data at the moment.

tomhughes (changes requested):
Overall the big difference between `import!` and `insert_all!` is that the former runs validations and the latter does not so maybe if we're going to change we should run `validate!` on each point after it is created?

Really the ideal way to use `insert_all!` would be not to build the models at all and just build a list of hashes but as it stands we need to run the save callbacks to fill in the tile number...

.rubocop.yml
mmd-osm: Rubocop was complaining about the use of insert_all! because it's bypassing validations.

app/models/trace.rb
mmd-osm: Good idea, I've added some unit tests for the API endpoint. It's a bit unfortunate though, that lat/lon values are identical in  `a.gpx`.

I've also added a call to `validate!`. I'm not sure if this whole validation is working as expected. Like before, I'm getting import failures in case the timestamp is missing, which is good. However, missing lat/lon values are not reported neither in the current code nor in the updated version in this PR.

tomhughes: There are no validation on lat/lon at the moment so that's not surprising...

mmd-osm: Ok, good. I believe all open points should be addressed by now.

tomhughes:
The main reason I haven't merged this is that I can't make up my mind if I like it... I understand that in theory moving to the built-in version is good but then it turned out they're not equivalent really so we have to jump through hoops and disable cops and it just starts to smell a little.

mmd-osm:
Well, we could disable rubocop's skip model validation check for that single line only, in case this would be an improvement.

```
# rubocop:disable Rails::SkipsModelValidations

 Tracepoint.insert_all!(tracepoints.map(&:attributes))

# rubocop:enable Rails::SkipsModelValidations
```

AntonKhorev:
Do we even need to build `Tracepoint` models? The only nontrivial thing there is tile calculation.

mmd-osm:
The question is what happens with validations. Do we want to maintain them in two different places (Tracepoint model and trace.rb)?

AntonKhorev:
Tracepoint model doesn't need validations because nothing is written using that model.

mmd-osm:
Tracepoint.import! calls model validations internally, like we discussed above. Try to upload data without timestamp and see what happens.

AntonKhorev:
Don't insert data without timestamps.

`delete_all` is already present in `trace.rb`. It doesn't call any hooks yet nobody is bothered by it. Might as well have `insert_all` and do all validations in `trace.rb`.

mmd-osm:
My goal was to remove the activerecord-import gem by a built in equivalent method. Since the validations are not automatically executed anymore, I added the call to the validate method. Rubocop is not clever enough to figure out that we're validating the model contents earlier on, so we need an additional annotation. I really don't have an issue with that.

Now we could of course further change the implementation and get rid of the model instances. However, that's a second step, and not needed to replace activerecord-import. That's not to say that this isn't a valid discussion to have, it just doesn't fit well here.

AntonKhorev:
This PR wasn't merged because

> we have to jump through hoops and disable cops

and I'm saying that we shouldn't worry here about model validations and what Rubocop thinks about them.

mmd-osm:
This comment still refers to the previous version where the whole file was added to .rubocop.yml.

@tomhughes hasn’t commented on the inline annotation variant yet, and I would suggest to wait for feedback before continuing.

Like I said, I don’t really have an issue with this annotation.

AntonKhorev:
> Rubocop is reporting a false positive here and we mark it as such.

Any use of `insert_all` etc that you want to keep is a false positive. You'll have to disable Rubocop for it one way or another.

mmd-osm:
A few years back, I've introduced both the activerecord-import gem and `import!` method call to speed up Tracepoint imports. Alternative options weren't available then. Now with the recent Rails version, it's a good time to move to the core Rails code.

At least today, we have exactly one location where `insert_all` will be used. Nobody else was doing anything with that gem over the past few years. The only other attempt to use activerecord-import was a performance improved changeset upload implementation which never got merged. So it's somewhat unlikely that we will see this issue in many other places in the future.

PR #5041, opened 2024-08-02
Detect <img> images in markdown rich text

AntonKhorev:
It's possible to add images to markdown rich text without the `` syntax, by using html ``. Such images are not detected by #4882 because they have a different type in the kramdown parse tree. This PR enables detection of ``.


tomhughes:
Looks good to me, thanks.

PR #5042, opened 2024-08-03
Replace richtext edit/preview buttons with tabs

AntonKhorev:
Why are the *Edit* and *Preview* buttons inside the help card? The *Preview* button there makes it inconvenient to preview on small screens because you'll have to scroll to the end of the help pane to press the button and then scroll back to see the preview.

Before:
image

After:
image

Uses https://getbootstrap.com/docs/5.3/components/navs-tabs/#javascript-behavior to switch between tab panes.

Compatible with #5023.

### DIFF
--- a/app/assets/javascripts/richtext.js
+++ b/app/assets/javascripts/richtext.js
@@ -14,9 +14,6 @@ $(document).ready(function () {
   /* Disable all the edit buttons */
   $(".richtext_doedit").prop("disabled", true);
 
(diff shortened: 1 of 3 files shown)


tomhughes:
I'm fine with the general idea here, and anything that gets rid of a load of my javascript is great so I'm going to merge this.

I'm not a huge fan of the white on white tabs though so hopefully somebody can come up with a way of improving that if I don't get there first.

PR #5043, opened 2024-08-04
Show richtext tabs over secondary body background

AntonKhorev:
Addresses https://github.com/openstreetmap/openstreetmap-website/pull/5042#issuecomment-2267514649 by using the same background as in content heading.

image
image
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -903,10 +903,12 @@ div.secondary-actions {
 
 /* Rules for tabs inside secondary background sections */
 
-.bg-body-secondary .nav-tabs {
-  --bs-border-color: var(--bs-secondary-border-subtle);
-  --bs-secondary-bg: var(--bs-secondary-border-subtle);
-  margin-bottom: -1px;
+.bg-body-secondary {
+  &.nav-tabs, .nav-tabs {
+    --bs-border-color: var(--bs-secondary-border-subtle);
+    --bs-secondary-bg: var(--bs-secondary-border-subtle);
+    margin-bottom: -1px;
+  }
 }
 
 /* Rules for traces */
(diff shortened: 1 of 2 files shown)


nertc:
In terms of comment resolution, your commit does a great job. But as we are changing color of the tabs, maybe something like this (see attached images) would be better for visual. Gap between tabs and textarea gives me an impression that they are separate functionalities, but for the user they should be interconnected. What do you think about it?

Hard borders:
Screenshot 2024-08-05 194318

Soft borders:
Screenshot 2024-08-05 194432

mmd-osm:
One other issue I've noticed is when you switch back from Preview to Edit mode, the textarea should be focused again. Ideally, the cursor would be set to the same location as before clicking on the preview tab. This is how Github preview works at least.

AntonKhorev:
@nertc 
> Gap between tabs and textarea gives me an impression that they are separate functionalities

Is this impression because of the background change? Maybe we shouldn't change it. Github has white-on-white tabs in other places:
image

AntonKhorev:
> when you switch back from Preview to Edit mode, the textarea should be focused again.

With Github it happens on click (including pressing space/enter), not on switch. You can switch between tabs using left/right keys, focus shouldn't jump away if you do that.

> Ideally, the cursor would be set to the same location as before clicking on the preview tab.

That's the default browser behavior, focusing on a textarea already works that way.

nertc:
> Is this impression because of the background change?

Yes. When it is filled, it has the need that area, which is changed by the tabs, should be sticked to it. But when it's white-on-white, user assumes everything that is below the tabs is the area that will be changed by the tabs (like Github).

> Maybe we shouldn't change it.

Mainly, I like designs with less borders and more space, but OSM heavily depends on borders and separations with the background color. I think, if we change background of tabs, it will be more relevant to the website design.

PR #5045, opened 2024-08-04
Changeset: avoid XML rendering in loop

mmd-osm:
XML rendering is much more consistent already. I only found one location where we are still calling `render` inside a loop.

### DIFF
--- a/app/views/api/changesets/index.xml.builder
+++ b/app/views/api/changesets/index.xml.builder
@@ -1,7 +1,4 @@
 xml.instruct! :xml, :version => "1.0"
-
 xml.osm(OSM::API.new.xml_root_attributes) do |osm|
-  @changesets.each do |changeset|
-    osm << render(changeset)
-  end
+  osm << (render(@changesets) || "")
 end


nenad-vujicic (approved):
This looks and works perfectly on my side. Thanks!

app/views/api/changesets/index.xml.builder
```diff
@@ -1,7 +1,4 @@
 xml.instruct! :xml, :version => "1.0"
-
 xml.osm(OSM::API.new.xml_root_attributes) do |osm|
-  @changesets.each do |changeset|
-    osm << render(changeset)
-  end
+  osm << (render(@changesets) || "")
```
AntonKhorev, line 3: Do we need this `|| ""`? 

`render(@changesets)` returns `nil` on empty set, then `nil` gets appended to a stream which should work fine:

https://github.com/rails/builder/blob/789896b306e36a5fe166388413b16e9323260601/lib/builder/xmlbase.rb#L116
https://github.com/rails/rails/blob/9ba208c16835f4a174ae9fd385ebc18972d758a4/actionview/lib/action_view/buffers.rb#L97

AntonKhorev:
More render inside a loop:

https://github.com/openstreetmap/openstreetmap-website/blob/7e22d42897697623ebd0e511a527803a73749ff5/app/views/api/changesets/download.xml.builder

app/views/api/changesets/index.xml.builder
AntonKhorev: Looks like it was needed at some point in a similar situation: https://github.com/openstreetmap/openstreetmap-website/commit/925bd812a2c9000c4901384b70f20e65b0f53cd4

AntonKhorev:
Anyway I merged this. `|| ""` can be looked into later.

app/views/api/changesets/index.xml.builder
mmd-osm: It’s for consistency reasons with the other 10 or so other places where we use the same pattern.

AntonKhorev: But is it required in any of those places?

mmd-osm: I haven't checked  the other places if it's really needed there. It could be that it was required some 10 years ago, and never revisited if that's still the case.

mmd-osm:
> [download.xml.builder](https://github.com/openstreetmap/openstreetmap-website/blob/7e22d42897697623ebd0e511a527803a73749ff5/app/views/api/changesets/download.xml.builder)

No real good idea how to get this working. Passing the collection alone does not work, since each element would have to be embedded into some `<create>` ... `</create>` tags.

PR #3945, opened 2023-02-25
Set og:image:alt OpenGraph header to “OpenStreetMap logo”

amandasaurus:
This might prevent sites from interpreting, or guessing, at what an alt text for this image is.

The current [Facebook Debugger for osm.org](https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fwww.openstreetmap.org%2F) shows all the standard Open Graph tags that we use, but included is a Facebook derived value for `og:image:alt` of `Sexoffenderrecord Map Near Me Free`.

I have no idea where it's getting that. I hope by manually specifying this, we might be able to override it.

PR #5046, opened 2024-08-05
Set og:image:alt OpenGraph property

AntonKhorev:
`og:image:alt` is set to
- "OpenStreetMap logo" when the default osm logo image is used as `og:image`, like attempted in #3945
- alt text from the first image of a diary entry when it's use as `og:image`
- nothing with the property skipped if the first image has no `alt` attribute, which is unlikely because you can't do it with markdown syntax

### DIFF
--- a/app/controllers/diary_entries_controller.rb
+++ b/app/controllers/diary_entries_controller.rb
@@ -69,6 +69,7 @@ class DiaryEntriesController < ApplicationController
     if @entry
       @title = t ".title", :user => params[:display_name], :title => @entry.title
       @og_image = @entry.body.image
+      @og_image_alt = @entry.body.image_alt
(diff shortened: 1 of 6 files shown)


gravitystorm:
> which is unlikely because you can't do it with markdown syntax

Just as a side note, it's quite easy (and common) to skip the alt text with the markdown syntax, for example:

```
![](http://tile.osm.org/0/0/0.png)
```

Ref: https://spec.commonmark.org/0.31.2/#example-581

I'll leave it to you to decide if this is important enough to add more tests to cover this edge case.

AntonKhorev:
`` is an empty alt, not skipped alt. You can see that the alt attribute is present in the output html.

PR #5050, opened 2024-08-06
Don't use XML::Parser when testing changeset upload diffResults

AntonKhorev:
`@response.body` is already parsed for use with `assert_dom`, why parse it again with `XML::Parser.string`?


tomhughes:
Looks good to me, thanks.

tomhughes:
There's a fair few more of these in the api controller tests by the look of it... They seem to be quite ancient history.

PR #5053, opened 2024-08-06
Lock note during status update to avoid race condition

### LINKED ISSUE #4528: Validation for note transitions
### URL

https://api.openstreetmap.org/api/0.6/notes/52751

I am analyzing the notes dump from Planet and found this invalid transition: reopen twice.

image

This is impacting the validations I am doing on my code.

I would like to know if this is normal (no validation). Or what is the issue behind this?

_No response_

mmd-osm:
As discussed in #4528, I'm proposing to add some note id locking as described here: https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/Locking/Pessimistic.html (first two examples).

The change covers destroy, comment, close and reopen methods, since all of them  are evaluating a note's visible and/or closed flag.

The important change is the "FOR UPDATE" addition when reading a note id initially. This way, two parallel processes cannot read the same note id at the same time and then both decide that e.g. reopening is permitted.

```
  TRANSACTION (0.1ms)  BEGIN
  ↳ app/controllers/api/notes_controller.rb:208:in `block in reopen'
  Note Load (14.7ms)  SELECT "notes".* FROM "notes" WHERE "notes"."id" = $1 LIMIT $2 FOR UPDATE  [["id", 3], ["LIMIT", 1]]
  [...]
  Note Update (0.3ms)  UPDATE "notes" SET "updated_at" = $1, "status" = $2, "closed_at" = $3 WHERE "notes"."id" = $4  [["updated_at", "2024-08-06 17:46:42.692378"], ["status", "open"], ["closed_at", nil], ["id", 3]]
  ↳ app/models/note.rb:58:in `reopen'
   [...]
  TRANSACTION (1.6ms)  COMMIT
```

Fixes #4528

### DIFF
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -115,12 +115,12 @@ module Api
       comment = params[:text]
 
       # Find the note and check it is valid
-      @note = Note.find(id)
-      raise OSM::APINotFoundError unless @note
-      raise OSM::APIAlreadyDeletedError.new("note", @note.id) unless @note.visible?
-
-      # Mark the note as hidden
       Note.transaction do
+        @note = Note.lock.find(id)
+        raise OSM::APINotFoundError unless @note
+        raise OSM::APIAlreadyDeletedError.new("note", @note.id) unless @note.visible?
+
+        # Mark the note as hidden
         @note.status = "hidden"
         @note.save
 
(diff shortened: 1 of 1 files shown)


gravitystorm (approved):
This looks good to me, but I defer to @tomhughes since he has more database expertise. He mentioned in the original issue that a transaction should be enough, but I don't think it's the case and I agree with @mmd-osm that a lock is necessary. Otherwise two transactions could run in parallel, both read the original status, be happy, and each add a comment without causing either transaction to fail.

I've checked the rails documentation and code, to make sure that the transaction will be rolled back (and therefore the lock released) if we raise errors ourselves.

mmd-osm:
> to make sure that the transaction will be rolled back 

Trying to reopen an open note (which is triggering an exception) immediately results a rollback. So this should work as expected:

```
  TRANSACTION (0.1ms)  BEGIN
  ↳ app/controllers/api/notes_controller.rb:208:in `block in reopen'
  Note Load (0.9ms)  SELECT "notes".* FROM "notes" WHERE "notes"."id" = $1 LIMIT $2 FOR UPDATE  [["id", 5], ["LIMIT", 1]]
  ↳ app/controllers/api/notes_controller.rb:208:in `block in reopen'
  TRANSACTION (0.2ms)  ROLLBACK
  ↳ app/controllers/api/notes_controller.rb:207:in `reopen'
```

tomhughes:
Locking is actually one place where I'm not very good because most of my work with databases is only on reading not on making updates so I don't normally have to worry about them.

I think the locks are necessary because the default isolation level in Postgres is `read committed` and you would have to run at `repeatable read` for write conflicts to be detected automatically. Even then it would only work for writes to the note objects and wouldn't prevent duplicate comments as that involves the creation of separate objects that wouldn't clash.

So yes, this looks reasonable to me.

PR #5054, opened 2024-08-07
Redirect to newly created diary comment

AntonKhorev:
Fixes #5048

PR #5055, opened 2024-08-07
Don't order by id when getting last records

AntonKhorev:
Ordering by id [happens by default](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-last).

Some more `order(:id)` are removed in #5054.


nenad-vujicic (approved):

tomhughes:
Looks good to me, thanks.

PR #5056, opened 2024-08-07
Set og:description meta tag to diary entry description

### LINKED ISSUE #1361: Diary posts should populate Open Graph tags for social sharing

AntonKhorev:
Uses the first paragraph with text truncated to 500 characters as an OpenGraph description on diary entry pages. Uses the default text on other pages or if no paragraph with text is found.

Fixes #1361

### DIFF
--- a/app/helpers/open_graph_helper.rb
+++ b/app/helpers/open_graph_helper.rb
@@ -1,7 +1,7 @@
 module OpenGraphHelper
   require "addressable/uri"
 
-  def opengraph_tags(title = nil, og_image = nil, og_image_alt = nil)
+  def opengraph_tags(title, og_image, og_image_alt)
     tags = {
       "og:site_name" => t("layouts.project_name.title"),
       "og:title" => title || t("layouts.project_name.title"),
--- a/app/helpers/open_graph_helper.rb
+++ b/app/helpers/open_graph_helper.rb
@@ -7,7 +7,7 @@ module OpenGraphHelper
       "og:title" => title || t("layouts.project_name.title"),
       "og:type" => "website",
       "og:url" => url_for(:only_path => false),
-      "og:description" => t("layouts.intro_text")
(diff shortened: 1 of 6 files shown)


gravitystorm (changes requested):
Just a small tweak here, otherwise looks good to me.

app/helpers/open_graph_helper.rb
```diff
@@ -1,15 +1,15 @@
 module OpenGraphHelper
   require "addressable/uri"
 
-  def opengraph_tags(title = nil, og_image = nil, og_image_alt = nil)
+  def opengraph_tags(title = nil, properties = {})
```
gravitystorm, line 4: This helper is only called in one place in the codebase, and it is called with two parameters. So the defaults never apply. They can be omitted entirely. 

Even if you call it as `opengraph_tags(nil, nil)` (or things that evaluate to nil, like an undefined instance variable) then `properties` will be `nil`, not `{}`.

AntonKhorev: removed default values

kcne (approved):
This is very useful and a great piece of work. The code seems fine as far as I can tell. This update also aligns well with PR addressing social sharing functionality #4985.

gravitystorm (approved):

gravitystorm:
Merged, thanks!

PR #5057, opened 2024-08-07
Add pagination to Issues page

nertc:
This PR addresses heavy calls and long load time of the Issues page. Problem was mentioned in https://github.com/openstreetmap/openstreetmap-website/pull/4990 by @AntonKhorev.

Added pagination functionality. When pagination buttons will be clicked, next or previous 50 issues will be called and displayed without a page refresh (Turbo will make a call and render returned Turbo Frame). In terms of performance, on the local environment I have 1000 users and 1000 issues. Initial loading time was decreased from ~8s to ~1s and after that every pagination needs ~150ms. In the real environment, it will have much better improvement, as instead of searching for 30000 issues, it will take only 50 (in theory, the difference should be 30x better in the real environment, then it is on my local machine). Because of the pagination functionality, now issues are sorted according to IDs (it should be the same sort as creation time sort).

To address @AntonKhorev 's [comment](https://github.com/openstreetmap/openstreetmap-website/pull/4990#issuecomment-2248072093) about browser built-in search, user now can change `?limit=50` query parameter in the url to any number. Therefore, if any user wants to use browser built-in search, he can just change it, for example, to `?limit=10000` and 10000 issues will be rendered.

To address @nertc 's [comment](https://github.com/openstreetmap/openstreetmap-website/pull/5033#issuecomment-2260783731) about clicking `go back` button of the browser, now every page is added to the browser history. Therefore, `go back` button will work fine, if user enters any report and then decides to go back to the list.

Video:

https://github.com/user-attachments/assets/b16f08f8-1e2a-472e-9a38-fb77e84bffac

Issues not found:

image

User not found:

Screenshot 2024-08-19 220613

### DIFF
--- a/app/controllers/issues_controller.rb
+++ b/app/controllers/issues_controller.rb
@@ -1,4 +1,6 @@
 class IssuesController < ApplicationController
+  include PaginationMethods
+
   layout "site"
 
   before_action :authorize_web
@@ -11,6 +13,8 @@ class IssuesController < ApplicationController
   before_action :check_database_writable, :only => [:resolve, :ignore, :reopen]
 
   def index
+    @params = params.permit(:before, :after, :limit, :status, :search_by_user, :issue_type, :last_updated_by)
+    @params[:limit] ||= 50
     @title = t ".title"
 
     @issue_types = []
@@ -18,7 +22,7 @@ class IssuesController < ApplicationController
     @issue_types.push("DiaryEntry", "DiaryComment", "User") if current_user.administrator?
 
     @users = User.joins(:roles).where(:user_roles => { :role => current_user.roles.map(&:role) }).distinct
-    @issues = Issue.visible_to(current_user).order(:updated_at => :desc)
+    @issues = Issue.visible_to(current_user)
 
     # If search
     if params[:search_by_user].present?
@@ -39,6 +43,8 @@ class IssuesController < ApplicationController
       last_updated_by = params[:last_updated_by].to_s == "nil" ? nil : params[:last_updated_by].to_i
       @issues = @issues.where(:updated_by => last_updated_by)
     end
+
+    @issues, @newer_issues_id, @older_issues_id = get_page_items(@issues, :limit => @params[:limit])
   end
 
   def show
--- a/app/controllers/issues_controller.rb
+++ b/app/controllers/issues_controller.rb
@@ -45,6 +45,7 @@ class IssuesController < ApplicationController
     end
 
     @issues, @newer_issues_id, @older_issues_id = get_page_items(@issues, :limit => @params[:limit])
+    render :partial => "page" if turbo_frame_request_id == "pagination"
   end
 
   def show
--- a/app/views/issues/_page.html.erb
+++ b/app/views/issues/_page.html.erb
@@ -1,34 +1,36 @@
-<table class="table table-sm">
-  <thead>
-    <tr>
-      <th><%= t ".status" %></th>
-      <th><%= t ".reports" %></th>
-      <th><%= t ".reported_item" %></th>
-      <th><%= t ".reported_user" %></th>
-      <th><%= t ".last_updated" %></th>
-    </tr>
-  </thead>
-  <tbody>
-    <% @issues.each do |issue| %>
+<turbo-frame id="pagination" target="_top">
+  <table class="table table-sm">
+    <thead>
       <tr>
-        <td><%= t ".states.#{issue.status}" %></td>
-        <td class="text-nowrap"><%= link_to t(".reports_count", :count => issue.reports_count), issue %></td>
-        <td><%= link_to reportable_title(issue.reportable), reportable_url(issue.reportable) %></td>
-        <td><%= link_to issue.reported_user.display_name, issue.reported_user if issue.reported_user %></td>
-        <td>
-          <% if issue.user_updated %>
-            <%= t ".last_updated_time_ago_user_html", :user => link_to(issue.user_updated.display_name, issue.user_updated),
-                                                      :time_ago => friendly_date_ago(issue.updated_at) %>
-          <% else %>
-            <%= friendly_date_ago(issue.updated_at) %>
-          <% end %>
-        </td>
+        <th><%= t ".status" %></th>
+        <th><%= t ".reports" %></th>
+        <th><%= t ".reported_item" %></th>
+        <th><%= t ".reported_user" %></th>
+        <th><%= t ".last_updated" %></th>
       </tr>
-    <% end %>
-  </tbody>
-</table>
-<%= render "shared/pagination",
-           :newer_key => "issues.page.newer_issues",
-           :older_key => "issues.page.older_issues",
-           :newer_id => @newer_issues_id,
-           :older_id => @older_issues_id %>
+    </thead>
+    <tbody>
+      <% @issues.each do |issue| %>
+        <tr>
+          <td><%= t ".states.#{issue.status}" %></td>
+          <td class="text-nowrap"><%= link_to t(".reports_count", :count => issue.reports_count), issue %></td>
+          <td><%= link_to reportable_title(issue.reportable), reportable_url(issue.reportable) %></td>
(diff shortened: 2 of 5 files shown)


nertc:
`Tests / Ubuntu 20.04, Ruby 3.0 (pull_request)` has failed while checking `test/controllers/changesets_controller_test.rb:324`. I haven't changed anything in the "changesets controller". Therefore, I don't know if fixing it is in the scope of this PR. If it's okay to fix it in this PR, I'll add `:wait => 1` to the end of the assertions. I think, it's some kind of Timing Issues.

tomhughes:
No please don't add random waits!

tomhughes:
Also why on earth does this need a whole load of javascript when all we're doing is adding next/previous page buttons? IF you want to do a partial load instead of reloading the whole page then you should be using turbo not doing it yourself.

gravitystorm:
I agree with Tom here, the SPA-like behaviour should be implemented using the Turbo framework, not with custom javascript. 

Otherwise, I'm happy to see pagination used for these pages. I would suggest just implementing the pagination first, then returning to the concept of using turbo drive separately, and ideally for all the paginated pages at the same time. 

Using Turbo will also avoid having to define an extra `page` route, which doesn't fit into our resourceful-routing approach.

tomhughes:
Note that we already have #4646 to add turbo to everything that uses the server side pagination helpers.

AntonKhorev:
and #4565 for changeset element pages

nertc:
Thank you for review, I'll update this PR accordingly to use Turbo instead of the JS code.

nertc:
I updated this PR to handle navigation using Turbo. It depends on Turbo Frames functionality. On my local machine all calls (first load and navigation) now need ~1.5s instead of ~8s. I tried to decrease loading time of navigation (like it was with the last JS solution ~200ms), but both, using Turbo Streams and custom partial rendering endpoint (like it was in the last solution), required extra JS code for good navigation history, so I sticked to the current simpler solution.

Current solution solves @gravitystorm 's comment:
> avoid having to define an extra page route

Thank you for linking your PRs, they were good examples.

nertc:
@AntonKhorev suggested a perfect solution in the [comment](https://github.com/openstreetmap/openstreetmap-website/pull/4646#issuecomment-2284931759) of the PR https://github.com/openstreetmap/openstreetmap-website/pull/4646. I updated code according to the suggestion and without any sacrifice of performance, extra JS code or history management, pagination time was again decreased to ~150ms (previous solution needed ~1.5s).

tomhughes:
Right but ideally we don't want to be doing this to each user of pagination separately, we want to do it across all users which is what my PR is all about. I will be reviewing @AntonKhorev's suggestion in the context of that PR when I get time.

app/views/issues/_page.html.erb
```diff
@@ -0,0 +1,40 @@
+<turbo-frame id="pagination" data-turbo-action="advance">
+  <% if @issues.length == 0 %>
+    <p><%= t(@user_not_found ? ".user_not_found" : ".issues_not_found") %></p>
+  <% else %>
```
tomhughes, line 4: I think this first branch should be outside the pagination frame because it's not actually paginated - only the else clause should be inside the frame as that is the part that can be navigated to different pages.

app/views/issues/_page.html.erb
```diff
@@ -0,0 +1,40 @@
+<turbo-frame id="pagination" data-turbo-action="advance">
```
tomhughes, line 1: This shouldn't have `data-turbo-action` but should have `target="_top"` and the advance action (and `pagination` as a target) should be on the next/previous buttons (which #4646 takes care of) so that other links inside the pagination section will break out to a full page load.

app/controllers/issues_controller.rb
```diff
@@ -39,6 +44,10 @@ def index
       last_updated_by = params[:last_updated_by].to_s == "nil" ? nil : params[:last_updated_by].to_i
       @issues = @issues.where(:updated_by => last_updated_by)
     end
+
+    @issues, @newer_issues_id, @older_issues_id = get_page_items(@issues, :limit => @params[:limit])
+
+    render :partial => "page" if turbo_frame_request_id == "pagination" || turbo_frame_request_id == "search"
```
tomhughes, line 50: I think the second part of this condition is wrong (in fact it's not clear to be the `search` frame is needed at all) and instead the search form should have a `data-turbo-frame="pagination"` attribute to target loading of the pagination frame.

app/views/issues/index.html.erb
```diff
@@ -4,72 +4,42 @@
...
 
-<%= form_tag(issues_path, :method => :get) do %>
-  <div class="row gx-1">
-    <div class="mb-3 col-md-auto">
-      <%= select_tag :status,
-                     options_for_select(Issue.aasm.states.map(&:name).map { |state| [t(".states.#{state}"), state] }, params[:status]),
-                     :include_blank => t(".select_status"),
-                     :data => { :behavior => "category_dropdown" },
-                     :class => "form-select" %>
+<turbo-frame id="search">
```
tomhughes, line 7: Do we actually need this frame? When does it get updated?

tomhughes (changes requested):
I think it might be better to split this into two commits - one that switches issues to use the common pagination helpers and a second to introduce turbo support.

app/views/issues/_page.html.erb
nertc: I agree with you. I wrote it this way for it to be both functional and not to conflict with https://github.com/openstreetmap/openstreetmap-website/pull/4646 after merging, but I'll change it back.

app/controllers/issues_controller.rb
nertc: It was written to update only the list part of the page when the search button was clicked. But it may be out of scope of this PR (as it introduces new action other than pagination). I'll remove and it may be a separate PR when we will finish pagination functionality.

tomhughes: Right, but you can do that by telling the search form to target the `pagination` frame, which is all that is actually being updated given that is what you render.

If you thought the search form needed to be updated as well then the above should be rendering a larger section for `search` but I don't think that does need to be updated does, only the results (which are in the `pagination` frame) need to be updated?

nertc:
@tomhughes Thank you for review. I updated PR according to your comments.

tomhughes:
As #4646 is merged now, can you rebase this on current master?

test/system/issues_test.rb
```diff
@@ -158,7 +158,35 @@ def test_issue_index_with_multiple_roles
...
+    # First Page
+    assert_no_content I18n.t("issues.index.user_not_found")
+    assert_no_content I18n.t("issues.index.issues_not_found")
+    assert_css "tr", :count => 51
+
+    # Second Page
+    click_on I18n.t("issues.page.older_issues")
+    assert_no_content I18n.t("issues.index.user_not_found")
+    assert_no_content I18n.t("issues.index.issues_not_found")
+    assert_css "tr", :count => 31, :wait => 1.5
```
tomhughes, line 184: I assume that `:wait` is a timeout? Why do we need that - what happens if we don't have it? Is there a default timeout that is too short or does it wait forever?

tomhughes: That doesn't really make any sense - a shorter timeout shouldn't fix it. I suspect it was actually just chance that you set a timeout and then that run happened to work within the timeout.

All of which is why I hat tests which rely on timeouts, because they're not deterministic and can fail arbitrarily on different machines or even on the same machine under different conditions.

It seems it's unavoidable with browser based testing like this though, so I suspect we should just increase the default to 5 or 10 or something.

nertc: I agree with you, it seems strange. I thought, maybe default time of my local machine was different with the one of the versions of GitHub workflow tests. I'll remove it and push without `:wait`-s. Let's see if it still fails.

nertc: All tests have passed. It's strange because in the first version all the tests were failing. Maybe changing a solution helped timing.

tomhughes (approved):
Thanks for that change. I think this is good to go now...

PR #5059, opened 2024-08-08
Move richtext help to a tab on small screens

AntonKhorev:
As I said in #5042 you have to scroll to the end of the help card in order to reach the edit/preview buttons. But you also have to scroll there in order to reach the submit button which is also annoying. What if we don't display the help card on small screens unless the user requests it?

When the screen is wide enough the UI looks like it did before:
image

When the screen is narrow:
image
image


tomhughes:
Looks good to me, thanks.

PR #5062, opened 2024-08-10
Move directions endpoints code to its own module

AntonKhorev:
Continues #4902. Apparently it's too tempting to put the routing logic back into endpoints code (https://github.com/openstreetmap/openstreetmap-website/pull/4904#pullrequestreview-2228123839, https://github.com/openstreetmap/openstreetmap-website/pull/4904#pullrequestreview-2230576231). Moving the code to another file will hopefully dissuade that.

### DIFF
--- a/app/assets/javascripts/index/directions-endpoint.js
+++ b/app/assets/javascripts/index/directions-endpoint.js
@@ -1,0 +1,84 @@
+OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, geocodeCallback) {
+  var endpoint = {};
+
+  endpoint.marker = L.marker([0, 0], {
+    icon: L.icon({
+      iconUrl: iconUrl,
+      iconSize: [25, 41],
+      iconAnchor: [12, 41],
+      popupAnchor: [1, -34],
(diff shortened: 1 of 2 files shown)


nenad-vujicic (approved):
Works great on my side! I think this is huge step forward in simplifying directions.js, fixing #2982 issue and closing PR #4904.

Thanks!

AntonKhorev:
I'm not fixing #2982 here yet. Fixing it would require answering https://github.com/openstreetmap/openstreetmap-website/pull/4904#issuecomment-2211735073. Do we actually need reverse geocoding? As I understand, #4904 tries to keep doing reverse geocoding in the same situations as before. Picked "Directions from here" from context menu? Do reverse geocoding. Dragged an endpoint marker? Don't do reverse geocoding. But why like that? Why not the other way around or never or always? The answers that make sense are:
- Never do reverse geocoding. But that may look like we're removing a feature that some people use.
- Do it if the user requests. This will require adding some button near or inside the from/to inputs.
- Always do it. This actually makes most sense. The endpoint module can handle this internally. It doesn't have to report if this operation succeeded to the outside code.

tomhughes:
Looks good to me, thanks.

Issue #1910, opened 2018-06-21
Include POI names in routing URL


Ekleog:
Hello,

When I search for a Paris → Lyon path, I come upon this page: https://www.openstreetmap.org/directions?engine=graphhopper_car&route=48.857%2C2.351%3B45.758%2C4.832 (or similar with OSRM, hence my supposition it's an issue with the OSM website, but if you think that's an upstream issue, I can take it to OSRM and/or GraphHopper :))

Unfortunately, if you open it, you won't be able to know that I searched for “Paris” and “Lyon”, and you will only have coordinates available.

I recently wanted to show off OSM to friends by sending them a path to a POI I had recently added that isn't in google map, but the absence of this feature makes it harder for me to do this “elegantly”.

Anyway, thank you for the work you make for OSM, that's a wonderful project!

tomhughes:
I thought this was a duplicate but I can't find it right now.

It's kind of deliberate though, to stop the URLs getting too long. We do try to convert back to a POI name from the coordinates but it doesn't always convert back to the original name.

Ekleog:
Hmm… I can confirm that the converting-back-to-POI-name works on project-osrm.org (eg. [there](http://map.project-osrm.org/?z=10&center=45.645368%2C4.747982&loc=48.856610%2C2.351499&loc=45.757814%2C4.832011&hl=en&alt=0)), but it doesn't work on the main openstreetmap.org website. It would already be half-solving this issue, indeed :)

Also, maybe having an option to have a longer link with the full information could help? Actually, honestly I personally don't care much about the length of addresses in most situations (ie. except when pasting URLs on IRC, where they could be broken). This is thus from my point of view something that is not at feature-parity with google maps (which turns out to have… quite lengthy URLs).

That said, having a choice would likely be best, indeed :) but in most cases when I share a URL I expect people to have the same information as I do, even though I know this is less and less true in today's web.

tomhughes:
We explicitly do not aim for feature parity with google maps - we are not trying to build a full service end user mapping portal.

The web site exists to support mappers rather than end users and routing is primarily there as a tool for detecting broken connections rather than as a way to publish routes to the world.

Ekleog:
Hmm… To me, it sounds like a debatable opinion from a marketing point of view (as the openstreetmap.org website is the first point of access for someone willing to try OSM), but if the decision is conscious, then so be it :)

[...]

PR #5064, opened 2024-08-11
Always do reverse geocoding when routing

### LINKED ISSUE #1874: Couldn't locate "..." with Directions and automatic reverse geocoding
Someone reported the following issue on the [German forum](https://forum.openstreetmap.org/viewtopic.php?id=62401) about some strange interaction between Directions and Automatic Reverse Geocoding. This could be some side effect of #1800.

How to reproduce:

1. Move to example location Lyon, Perrache, France
2. Context menu: set "from" location
3. Context menu: set different "to" location
4. Meanwhile from and to get automatically reverse geocoded.
5. Error message is shown about unknown locations ("Couldn't locate ...")

Pretty strange that the reverse geocoded location returned by Nominatim isn't know to Nominatim anymore at a later point in time.

I added a short video, hope that's ok to demonstrate the issue.

peek 2018-05-20 09-45

### LINKED ISSUE #1911: Directions interface does not use exact location when selected by right-clicking
#### Issue description
The directions interface does not use the location that I click. Instead it somehow snaps to a nearest 'node' or grid of some kind.

#### Steps to reproduce the issue
1. On the main map page, right-click on a location and select 'Directions from here'.
2. Right-click somewhere else and select 'Directions to here'.

#### What's the expected result?
I should get directions from the exact location I clicked first, to the exact location I clicked second.

#### What's the actual result?
I get directions from some point near the location I clicked first, to some point near the location I clicked second.

#### Additional details
- Directions between arbitrary points are not impossible for the engine, as is evidenced by manually clicking and dragging the start and end nodes of the generated directions to arbitrary locations (they do not snap back then).
- Possibly the cause of this behaviour is [here](https://github.com/openstreetmap/openstreetmap-website/blob/999b9c69e1af252610a2ad338d33d16273fcc22f/app/assets/javascripts/index/contextmenu.js#L22)? (I wanted to try it out but I couldn't get the dev env running)

AntonKhorev:
There was a number of bugs related to route endpoint locations caused by various reasons. Some of there reasons were already addressed. The one left is Nominatim deciding to do reverse geocoding when receiving something that looks like coordinates in its freeform query. The query itself is *not* a reverse geocoding query (`/reverse?...`), yet Nominatim will do reverse geocoding if it consists of two floating point numbers.

The problem is that we have inputs that can contain either addresses or coordinates. Then the seemingly obvious thing to get coordinates is to pass this input contents to Nominatim. And this is how you get bugs. It the input already contained coordinates, Nominatim would reverse-geocode them to some place and return the coordinates of that place that are different from the coordinates input already had. And then the endpoint marker would jump somewhere.

If we want to prevent this, we have to check ourselves if we already have two floating point numbers. If we do, we shouldn't make Nominatim's freeform geocoding query. Since we're going to parse coordinates inside the endpoint module to make that check, `endpoint.setValue` doesn't need its `latlng` argument and it is removed.

Since users probably expect reverse geocoding to be happening even if we had coordinates, `endpoint.setValue` is going to make a reverse-geocoding query explicitly, without overwriting the original coordinates. With changes here, `endpoint.setValue` works like this:
- The `value` parameter is checked if it's coordinates.
- If not, geocoding is performed. 
- If yes, reverse geocoding is done.

Reverse geocoding is also done on marker drag end.

Another thing that's done is caching of reverse geocoding results. The reason is that there are common scenarios when you get the same coordinates that were reverse-geocoded. For example, you set the starting point using the context menu, then you add the ending point. The context menu will open a directions url containing the same starting point coordinates, they were looked up already.

Another common scenario is swapping the route start/end. To avoid reverse-geocoding both ends again we can also swap their reverse-geocoding results. That's what `endpoint.swapCachedReverseGeocodes` does.

---

This should fix #1874 and #2982.

Also addresses #1910 to some extent. But if we want to get back "Paris" from the coordinates and not some address inside Paris, we'd need to pass along a zoom level. Nominatim [supposedly can use it](https://nominatim.org/release-docs/develop/api/Reverse/#result-restriction). Or pass search queries along. Presumably that wasn't done to [keep urls short](https://github.com/openstreetmap/openstreetmap-website/issues/1910#issuecomment-399255834), but we can pass only short queries for example.

Also should fix #1911 if the problem is markers moving from clicked locations to geocoding locations. In this case it's the step 3 of https://github.com/openstreetmap/openstreetmap-website/issues/2982#issuecomment-734512269.

May close #1754 because everything was rewritten.

### DIFF
--- a/app/assets/javascripts/index/directions-endpoint.js
+++ b/app/assets/javascripts/index/directions-endpoint.js
@@ -55,7 +55,7 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
     endpoint.setValue(value);
   }
 
-  endpoint.setValue = function (value, latlng) {
+  endpoint.setValue = function (value) {
     endpoint.value = value;
     removeLatLng();
     input.removeClass("is-invalid");
@@ -63,6 +63,9 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
 
     if (endpoint.geocodeRequest) endpoint.geocodeRequest.abort();
     delete endpoint.geocodeRequest;
+
+    var coordinatesMatch = value.match(/^\s*([+-]?\d+(?:\.\d*)?)(?:\s+|\s*[/,]\s*)([+-]?\d+(?:\.\d*)?)\s*$/);
+    var latlng = coordinatesMatch && L.latLng(coordinatesMatch[1], coordinatesMatch[2]);
 
     if (latlng) {
       setLatLng(latlng);
--- a/app/assets/javascripts/index/directions-endpoint.js
+++ b/app/assets/javascripts/index/directions-endpoint.js
@@ -86,6 +86,15 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
     }
   };
 
+  endpoint.swapCachedReverseGeocodes = function (otherEndpoint) {
+    var g0 = endpoint.cachedReverseGeocode;
+    var g1 = otherEndpoint.cachedReverseGeocode;
+    delete endpoint.cachedReverseGeocode;
+    delete otherEndpoint.cachedReverseGeocode;
+    if (g0) otherEndpoint.cachedReverseGeocode = g0;
+    if (g1) endpoint.cachedReverseGeocode = g1;
+  };
+
   function getGeocode() {
     var viewbox = map.getBounds().toBBoxString(); // <sw lon>,<sw lat>,<ne lon>,<ne lat>
     var geocodeUrl = OSM.NOMINATIM_URL + "search?q=" + encodeURIComponent(endpoint.value) + "&format=json&viewbox=" + viewbox;
--- a/app/assets/javascripts/index/directions-endpoint.js
+++ b/app/assets/javascripts/index/directions-endpoint.js
@@ -66,7 +66,12 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
 
     if (latlng && endpoint.cachedReverseGeocode && endpoint.cachedReverseGeocode.latlng.equals(latlng)) {
       setLatLng(latlng);
-      endpoint.value = endpoint.cachedReverseGeocode.value;
+      if (endpoint.cachedReverseGeocode.notFound) {
+        endpoint.value = value;
+        input.addClass("is-invalid");
+      } else {
+        endpoint.value = endpoint.cachedReverseGeocode.value;
+      }
       input.val(endpoint.value);
       changeCallback();
       return;
@@ -123,6 +128,7 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
     endpoint.geocodeRequest = $.getJSON(reverseGeocodeUrl, function (json) {
       delete endpoint.geocodeRequest;
       if (!json || !json.display_name) {
+        endpoint.cachedReverseGeocode = { latlng: latlng, notFound: true };
         return;
       }
 
--- a/app/assets/javascripts/index/directions-endpoint.js
+++ b/app/assets/javascripts/index/directions-endpoint.js
@@ -57,16 +57,24 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
   }
 
   endpoint.setValue = function (value) {
-    endpoint.value = value;
-    removeLatLng();
-    input.removeClass("is-invalid");
-    input.val(value);
-
     if (endpoint.geocodeRequest) endpoint.geocodeRequest.abort();
     delete endpoint.geocodeRequest;
+    input.removeClass("is-invalid");
 
     var coordinatesMatch = value.match(/^\s*([+-]?\d+(?:\.\d*)?)(?:\s+|\s*[/,]\s*)([+-]?\d+(?:\.\d*)?)\s*$/);
     var latlng = coordinatesMatch && L.latLng(coordinatesMatch[1], coordinatesMatch[2]);
+
+    if (latlng && endpoint.cachedReverseGeocode && endpoint.cachedReverseGeocode.latlng.equals(latlng)) {
+      setLatLng(latlng);
+      endpoint.value = endpoint.cachedReverseGeocode.value;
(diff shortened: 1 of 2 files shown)


nenad-vujicic:
Thanks for great PR! The sources are now much more readable and it seems a lot of bugs are fixed.

Just a few remarks:

1. Perhaps it would be good to fit map's bounds (directions.js Ln 159) on union of markers and route polyline (instead of route polyline only). 2 cases when this is useful: 
a) at least one (or both) of markers are too far from retrieved route and they are not visible in final display and
b) routing engine was unable to return the route (e.g. place markers at (0,0) and (1,1)) and only markers are displayed (but, if you type them, map didn't focus on them).

2. It seems we still have duplicated calls to nominatim when we select from context menu (other cases are OK). Please, take a look at below video:

https://github.com/user-attachments/assets/2f7a4ed8-1e82-4eb4-becf-fd28a7c3e05c

Anyway, even with above remarks, I think this is great PR. Thanks for making it!

nenad-vujicic (approved):

AntonKhorev:
> It seems is-invalid is not removed from inputs after successful contacting nominatim

They aren't removed in the original code either. The difference is it doesn't call nominatim, but it should have removed is-invalid as soon as the coordinates updated.

AntonKhorev:
"2." form https://github.com/openstreetmap/openstreetmap-website/pull/5064#issuecomment-2285981589 - I only see one request to Nominatim happening when "Directions from/to here" is selected.

nenad-vujicic:
> "2." form [#5064 (comment)](https://github.com/openstreetmap/openstreetmap-website/pull/5064#issuecomment-2285981589) - I only see one request to Nominatim happening when "Directions from/to here" is selected. I fixed something like this in [2d1cf4c](https://github.com/openstreetmap/openstreetmap-website/commit/2d1cf4c380442595d7a4629f3a4c79124277f12b) but wasn't it written before the comment?


https://github.com/user-attachments/assets/e5b3980e-4779-4104-8cc0-4b581a631ea4

If one point is valid and you change the second one, yes, there is only one request from nominatim. But, if one point is invalid and you change the second one, it will request from nominatim 2 times (once for each point).

I used the latest sources at the moment on which I applied this PR.

Thanks!

AntonKhorev:
@nenad-vujicic how about now? I added caching of not found results from Nominatim.

nenad-vujicic:
> @nenad-vujicic how about now? I added caching of not found results from Nominatim.

It's perfect now. Thank you very much!

There is one another issue, but probably not part of this PR. In some situations (take a look at attached video), on invalid search scene map fits to some other boundary box and route is displayed on wrong position. After pushing OK, it retrieves good route, displays it on good position and fits to good boundary box.

https://github.com/user-attachments/assets/05ddd792-be25-4990-bb54-1f1936d4f8c9

AntonKhorev:
> In some situations (take a look at attached video), on invalid search scene map fits to some other boundary box and route is displayed on wrong position

Could be caused by `alert()`. I'll later remove them.

AntonKhorev:
I guess this can be merged now. I don't think there's anything else worth taking out and merging separately.

app/assets/javascripts/index/directions-endpoint.js
```diff
@@ -79,12 +111,32 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
...
+      endpoint.value = json[0].display_name;
       input.val(json[0].display_name);
 
       changeCallback();
     });
   }
 
+  function getReverseGeocode() {
+    var latlng = endpoint.latlng.clone();
+    var reverseGeocodeUrl = OSM.NOMINATIM_URL + "reverse?lat=" + latlng.lat + "&lon=" + latlng.lng + "&format=json";
```
tomhughes, line 123: Do we need to control the match precision here? Might this cause a location to "snap" to something a long way away?

tomhughes (commented):
I'm not sure I agree with either part of the title of this PR as I don't really see any simplification - quite the opposite if anyway. The "always do reverse geocoding" is also misleading because it implies extending an existing use when in fact this introduces reverse geocoding for the first time?

app/assets/javascripts/index/directions-endpoint.js
```diff
@@ -79,12 +111,32 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
...
+
+    endpoint.geocodeRequest = $.getJSON(reverseGeocodeUrl, function (json) {
+      delete endpoint.geocodeRequest;
+      if (!json || !json.display_name) {
+        input.addClass("is-invalid");
+        alert(I18n.t("javascripts.directions.errors.no_place", { place: endpoint.value }));
```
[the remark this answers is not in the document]
AntonKhorev: It shouldn't be an error but it currently is, that's why I kept it. That's not the only thing wrong about error handling here.

Removed `addClass` and `alert` now.

app/assets/javascripts/index/directions-endpoint.js
AntonKhorev: The location wouldn't snap in the same way is does on the currently deployed version. The pointer is not going to move, just the address would be too imprecise.

Ultimately we should control the precision but I didn't want to do it in this pull request, see https://github.com/openstreetmap/openstreetmap-website/pull/4904#discussion_r1723165682

AntonKhorev:
> in fact this introduces reverse geocoding for the first time?

Reverse geocoding happens in the deployed version when Nominatim decides to do it. I don't know if it's intentional or accidental.

- context menu, directions from/to here - reverse geocoding
- drag/drop a marker - no reverse geocoding
- type in coordinates - reverse geocoding if you typed two numbers with decimal points (or what's the logic for detecting coordinates in Nominatim), no reverse geocoding otherwise

AntonKhorev:
> I'm not sure I agree with either part of the title of this PR as I don't really see any simplification - quite the opposite if anything.

A large part of this pull request is already merged with other pull request. The simplification here is that you don't have to pass `latlng` to `endpoint.setValue`.

tomhughes:
> A large part of this pull request is already merged with other pull request

Can we rebase it then so we can see what's left?

AntonKhorev:
Updated pull request description.

tomhughes (approved):
I've had another look through this, and played with a bit in my dev environment and I think it's probably good to merge now. Thanks for all the work on it.

PR #5069, opened 2024-08-13
Add deactivates_at date to user blocks

AntonKhorev:
Block deactivation dates that take needs_view-block views into account were derived using updated_at. This was possible because inactive blocks couldn't be edited and their updated_at date wouldn't change. The logic is explained here:

https://github.com/openstreetmap/openstreetmap-website/blob/e0e6c7f1c26fcddff883dad38a6207475150f323/app/helpers/user_blocks_helper.rb#L20-L24

With editing of inactive blocks enabled deactivation date needs to be saved explicitly. It is defined for blocks that don't have `needs_view` and will time out even if the user doesn't look at them. Block with `needs_view` don't have a defined deactivation date. It's going to be set as soon as the user views the block and clears `needs_view`. Its value is going to be either the block end date if it's in the future or the current date.

---

After creating `deactivates_at` we no longer need `needs_view`. A block with `needs_view` is just a block without `deactivates_at`. But I can't remove `needs_view` right away because strong_migrations won't let me to drop the column without extra steps.

### DIFF
--- a/app/controllers/user_blocks_controller.rb
+++ b/app/controllers/user_blocks_controller.rb
@@ -26,6 +26,7 @@ class UserBlocksController < ApplicationController
   def show
     if current_user && current_user == @user_block.user
       @user_block.needs_view = false
+      @user_block.deactivates_at = [@user_block.ends_at, Time.now.utc].max
       @user_block.save!
     end
   end
@@ -49,6 +50,7 @@ class UserBlocksController < ApplicationController
         :ends_at => now + @block_period.hours,
         :needs_view => params[:user_block][:needs_view]
       )
+      @user_block.deactivates_at = @user_block.ends_at unless @user_block.needs_view
 
       if @user_block.save
         flash[:notice] = t(".flash", :name => @user.display_name)
@@ -72,11 +74,16 @@ class UserBlocksController < ApplicationController
         @user_block.reason = params[:user_block][:reason]
         @user_block.needs_view = params[:user_block][:needs_view]
         @user_block.ends_at = Time.now.utc + @block_period.hours
+        @user_block.deactivates_at = (@user_block.ends_at unless @user_block.needs_view)
         if !user_block_was_active && @user_block.active?
           flash.now[:error] = t(".inactive_block_cannot_be_reactivated")
           render :action => "edit"
         else
-          @user_block.ends_at = @user_block.ends_at_was unless user_block_was_active
+          unless user_block_was_active
+            @user_block.ends_at = @user_block.ends_at_was
+            @user_block.deactivates_at = @user_block.deactivates_at_was
+            @user_block.deactivates_at = [@user_block.ends_at, @user_block.updated_at].max unless @user_block.deactivates_at # take updated_at into account before deactivates_at is backfilled
+          end
           if @user_block.save
             flash[:notice] = t(".success")
             redirect_to @user_block
--- a/app/helpers/user_blocks_helper.rb
+++ b/app/helpers/user_blocks_helper.rb
@@ -20,7 +20,7 @@ module UserBlocksHelper
       # the max of the last update time or the ends_at time is when this block finished
       # either because the user viewed the block (updated_at) or it expired or was
       # revoked (ends_at)
-      last_time = [block.ends_at, block.updated_at].max
+      last_time = block.deactivates_at || [block.ends_at, block.updated_at].max
       t("user_blocks.helper.time_past_html", :time => friendly_date_ago(last_time))
     end
   end
--- a/app/models/user_block.rb
+++ b/app/models/user_block.rb
@@ -2,16 +2,17 @@
 #
 # Table name: user_blocks
 #
-#  id            :integer          not null, primary key
-#  user_id       :bigint(8)        not null
-#  creator_id    :bigint(8)        not null
-#  reason        :text             not null
-#  ends_at       :datetime         not null
-#  needs_view    :boolean          default(FALSE), not null
-#  revoker_id    :bigint(8)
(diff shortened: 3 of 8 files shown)


mmd-osm:
Please don't remove needs_view, CGImap has a dependency on this field: https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/src/backend/apidb/readonly_pgsql_selection.cpp#L786-L791

I haven't checked the new logic in detail, in case this has some impact on readonly_pgsql_selection::is_user_blocked, we need to coordinate this change.

app/controllers/user_blocks_controller.rb
```diff
@@ -67,15 +69,24 @@ def update
...
-      )
-        flash[:notice] = t(".success")
-        redirect_to(@user_block)
       else
-        render :action => "edit"
+        user_block_was_active = @user_block.active?
+        @user_block.reason = params[:user_block][:reason]
+        @user_block.needs_view = params[:user_block][:needs_view]
+        @user_block.ends_at = Time.now.utc + @block_period.hours
+        @user_block.deactivates_at = (@user_block.ends_at unless @user_block.needs_view)
```
tomhughes, line 77: I don't think this needs parentheses does it?

app/controllers/user_blocks_controller.rb
```diff
@@ -67,15 +69,24 @@ def update
...
+        if !user_block_was_active && !@user_block.active?
+          @user_block.ends_at = @user_block.ends_at_was
+          @user_block.deactivates_at = @user_block.deactivates_at_was
+        end
+        @user_block.deactivates_at = [@user_block.ends_at, @user_block.updated_at].max if !@user_block.deactivates_at && !@user_block.needs_view
+
+        if @user_block.save
+          flash[:notice] = t(".success")
+          redirect_to(@user_block)
```
tomhughes, line 86: Does this needs parentheses?

app/controllers/user_blocks_controller.rb
```diff
@@ -67,15 +69,24 @@ def update
...
+        user_block_was_active = @user_block.active?
+        @user_block.reason = params[:user_block][:reason]
+        @user_block.needs_view = params[:user_block][:needs_view]
+        @user_block.ends_at = Time.now.utc + @block_period.hours
+        @user_block.deactivates_at = (@user_block.ends_at unless @user_block.needs_view)
+        if !user_block_was_active && !@user_block.active?
+          @user_block.ends_at = @user_block.ends_at_was
+          @user_block.deactivates_at = @user_block.deactivates_at_was
+        end
+        @user_block.deactivates_at = [@user_block.ends_at, @user_block.updated_at].max if !@user_block.deactivates_at && !@user_block.needs_view
```
tomhughes, line 82: I don't think this needs both conditions does it? The only way `deactivates_at` can be unset is it `needs_view` is false?

The whole logic here is really hard to grasp and reason about to be honest and I wonder if there's a better way to do it or if we should even be allowing all the cases this is allowing - should we be allowed an expired block to be reactivated?

app/helpers/user_blocks_helper.rb
```diff
@@ -20,7 +20,7 @@ def block_status(block)
       # the max of the last update time or the ends_at time is when this block finished
       # either because the user viewed the block (updated_at) or it expired or was
       # revoked (ends_at)
-      last_time = [block.ends_at, block.updated_at].max
+      last_time = block.deactivates_at || [block.ends_at, block.updated_at].max
```
tomhughes, line 23: Do we need the second half of this? It's just for old blocks because we're not backfilling `deactivates_at` for them?

app/controllers/user_blocks_controller.rb
AntonKhorev: Yes it does because I want the assignment to always happen. That expression evaluates to nil if @user_block.needs_view.

AntonKhorev:
> Please don't remove needs_view, CGImap has a dependency on this field

I was going to replace it with https://www.postgresql.org/docs/12/ddl-generated-columns.html

app/helpers/user_blocks_helper.rb
AntonKhorev: Yes, this is until backfilling is done. After that the logic will be simpler everywhere. `last_time` is just `block.deactivates_at`, `block.active?` is comparing the current time with `block.deactivates_at`.

app/controllers/user_blocks_controller.rb
AntonKhorev: No, but I didn't change this place
https://github.com/openstreetmap/openstreetmap-website/blob/7cd7db571824a583229be914d56652517540452c/app/controllers/user_blocks_controller.rb#L75-L76

app/models/user_block.rb
```diff
@@ -28,6 +29,8 @@
 class UserBlock < ApplicationRecord
   validate :moderator_permissions
   validates :reason, :characters => true
+  validates :deactivates_at, :comparison => { :greater_than_or_equal_to => :ends_at }, :if => -> { !needs_view }
```
tomhughes, line 32: I think this could be `:unless => -> { needs_view }`?

app/controllers/user_blocks_controller.rb
AntonKhorev: Now with reactivations disabled the `@user_block.deactivates_at = [ ... ].max` condition should be simpler.

app/models/user_block.rb
AntonKhorev: updated to `:unless`

tomhughes:
I think this is good to go now, thanks.

Issue #5010, opened 2024-07-22
NodeJS versions are out of sync again


gravitystorm:
In #4992 the nodejs version was updated to node18, but only for Docker. The version wasn't updated for either the installation instructions (INSTALL.md), nor for the vagrant installation. These three alternatives should be kept in sync at all times.

The reason that nodejs had to be upgraded to version 18 was because eslint 9 requires it. The previous version of eslint, version 8, also ran on node 12, which is what is shipped on Ubuntu 22.04. https://github.com/openstreetmap/openstreetmap-website/commit/938037d6a26baecd334944957882c0108d2176e6 was the upgrade. This wasn't picked up as a breaking change in CI, since the "github flavoured ubuntu" used in CI doesn't use the Ubuntu node packages that INSTALL.md and vagrant use.

We need to do one of the following:
* Downgrade to eslint 8, and remove the external node18 installation from Docker, or
* Add the external node18 installation to both INSTALL.md and the vagrant configuration, or
* Set 24.04 as our minimum development version, update INSTALL.md, vagrant, Docker, other CI too

We also need to consider what extra things we can do to avoid this from happening again. For example,
* Is it possible to put maintainer-facing notes (or Dangerfile etc) to avoid these becoming out-of-sync?
* Is it possible to change github CI to use a fixed version of node, that we can match to the version that's shipped in Ubuntu?

mmd-osm:
Are you trying to run this on arm64? https://github.com/rubyjs/mini_racer has some comment in the readme about platforms: 

make sure to have PLATFORMS set correctly in Gemfile.lock via [bundle lock --add-platform](https://bundler.io/v2.4/man/bundle-lock.1.html#SUPPORTING-OTHER-PLATFORMS)

kcne:
> Are you trying to run this on arm64? https://github.com/rubyjs/mini_racer has some comment in the readme about platforms: 
> 
> 
> 
> make sure to have PLATFORMS set correctly in Gemfile.lock via [bundle lock --add-platform](https://bundler.io/v2.4/man/bundle-lock.1.html#SUPPORTING-OTHER-PLATFORMS)

Tried running on amd64 linux machine also but I got error there too. I don't think that is the problem here but I will take a look at the docs just in case.

nenad-vujicic:
> We also need to consider what extra things we can do to avoid this from happening again. For example,
> 
[...]

PR #5070, opened 2024-08-13
Update to rails 7.2.0

tomhughes:


tomhughes:
Looks like this is a no-go until we can move to ruby 3.1 then...

mmd-osm:
I did a smoke test on Ubuntu 22.04 + ruby 3.3.2 (2024-05-30 revision e5a195edf6), which I installed manually at one point. Tests are passing without issue:

```
Finished in 66.145466s, 26.2301 runs/s, 15520.4439 assertions/s.
1735 runs, 1026607 assertions, 0 failures, 0 errors, 0 skips
```

A few tests are reporting some warning "Test is missing assertions":

- test_3rd_party_token: test/integration/client_applications_test.rb:73
- test_body_class: test/helpers/application_helper_test.rb:92
- test_header_nav_link_class: test/helpers/application_helper_test.rb:94
- test_lost_password_recovery_success: test/integration/user_creation_test.rb:174
- test_show_no_friends: test/controllers/dashboards_controller_test.rb:13

tomhughes:
Not sure why you're getting those messages because I don't see them.

I mean I can see what the problem is - people have written stubs for tests that don't do anything yet. It's just a bit odd that you get warned about them and I don't.

tomhughes:
In some cases they're not even stubs - a test of an action with no assertions still checks that a given action will run without throwing any exceptions.

mmd-osm:
According to the docs, https://api.rubyonrails.org/v7.1.3.4/classes/ActiveSupport/Testing/Assertions.html#method-i-assert_nothing_raised should handle this case.

tomhughes:
Yes I figured it out... I had `RUBYOPT` set to suppress warnings in my environment. I think at some point there were some annoying warnings we couldn't get rid of and rails doesn't normally use actual ruby warnings like that when it wants to warn about something.

gravitystorm:
I think it's a shame that we have to drop 22.04 support, since I expect some developers won't have yet upgraded their Ubuntu to 24.04. (I haven't, but I used rbenv - we don't want to rely on that for new developers). 

It seems like an unfortunate combination of release dates and support periods:
* ruby 3.0 was released in Dec 2020
* ruby 3.1 was released in Dec 2021, which was in time for Ubuntu 22.04 but wasn't the chosen version
* ruby 3.0 support ceased in Apr 2024
* rails drops 3.0 support (and therefore Ubuntu 22.04 support) in Aug 2024

So we end up with a period of only 4 months where we have two versions of Ubuntu available. We'll face a similar problem in two year's time, because Ubuntu 24.04 shipped with ruby 3.2, which will be EOL in Mar 2026 i.e. just before Ubuntu 26.04 LTS. I wish either ruby would extend their support a little, or Ubuntu would use the newest ruby, or rails would be more tolerant of slightly EOL rubies, all the make this developer experience a bit smoother.

Nevertheless, I'm happy to move forward. I would say that we should:

* Drop 22.04 from the testing matrix. If we can't run on real Ubuntu 22.04, there's little point in testing on Github Flavoured Ubuntu
* Adjust the Vagrantfile to use `ubuntu/noble64` or `generic/ubuntu2404` as appropriate
* There's a failure in building the docker image in CI

tomhughes:
There's an extra complication which is that we can't (yet) use the GitHub 24.04 image to run tests as it's still beta and is missing firefox (planned to be fixed before it leaves beta) so for now we probably have to stick with 22.04 for GHA.

tomhughes:
Also we're waiting on https://github.com/jumph4x/canonical-rails/pull/92 before we can go to 7.2.1 which is already out.

tomhughes:
I've updated this as follows:

* Reverted changes to ubuntu versions in testing matrix for now pending resolution of https://github.com/actions/runner-images/issues/10374
* Updated vagrantfile
* Fixed "no assertion" warnings in tests

The docker issue is https://github.com/rubyjs/mini_racer/issues/305 / https://github.com/rubyjs/mini_racer/issues/305 which stops mini_racer installing on 24.04 currently.

tomhughes:
Oh just to add that all the production servers, and the dev server, are now on Debian 12 and hence ruby 3.1.

mmd-osm:
https://github.com/openstreetmap/openstreetmap-website/issues/5010#issuecomment-2261537069 proposes some changes to use Debian 12 for Docker. Maybe we should try to resolve the remaining issue there, to keep everything a bit more in sync.

gravitystorm:
Looks good to me, thanks!

tomhughes:
Oh I was going to wait until we could do 7.2.x updates in case there security issues that couldn't wait...

tomhughes:
It looks like 7.2.1 is OK at least as there don't look to be any security fixes.

gravitystorm:
> Oh I was going to wait until we could do 7.2.x updates in case there security issues that couldn't wait...

Ah, I didn't realise it was blocked for any reason. I saw that the canonical-rails PR had been merged already, but I guess the problem is that it hasn't been included in a release.

I can make a PR that moves to canonical-rails from git master, if that helps? That way we can get point releases for rails 7.2.x

tomhughes:
I don't think we need to worry for now but it's a way out if we do need to do a rails update urgently.

gravitystorm:
Ah, I created https://github.com/openstreetmap/openstreetmap-website/pull/5246 before I saw your comment, so the work is done if you want it. I've no idea whether rails 7.2.2 or canonical-rails will be the first to be released.

PR #5075, opened 2024-08-14
Switch to edit pane if rich text field validation fails

AntonKhorev:
Before you'd click the submit button and nothing would happen if the text is invalid:
image

With this PR the *Edit* pane shows which lets you see the validation message:
image


tomhughes:
Looks good to me, thanks.

tomhughes:
Looks OK to me, thanks.

PR #5076, opened 2024-08-14
Add debug gem

gravitystorm:
This is included in new rails 7+ apps, so let's include it too.

This also means it's available on ruby 3.0, and is automatically updated on newer versions of ruby too.

I've copied this directly from [the upstream `rails new` Gemfile template](https://github.com/rails/rails/blob/eaa74eedba945e1bdcedd7fa6ac38c6493edd735/railties/lib/rails/generators/rails/app/templates/Gemfile.tt#L45-L46).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -180,4 +180,7 @@ end
 
 group :development, :test do
   gem "annotate"
+
+  # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
+  gem "debug", :require => "debug/prelude"
 end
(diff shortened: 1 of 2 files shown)


tomhughes:
What is the actual effect of this? The web page helpfully says it "enters a debugging session" when it hits a break point but doesn't really explain how you connect to that session and interact with it?

gravitystorm:
I was just using it to try to figure out something in the tests. If I put `binding.break` in the middle of the test (e.g. deep in some `assert_dom` stuff) and then run the test, it turns into the rough equivalent of an irb session. I can then type commands, inspect variables, and then exit out to allow the test to complete.

tomhughes:
Ah ok in a test makes more sense because you have a console then though doesn't the parallelisation cause problems? or did you just run a single test?

PR #5078, opened 2024-08-14
Add article:published_time meta tag to diary entries

AntonKhorev:
Mastodon seems to be using it:
https://en.osm.town/@openstreetmap/112670776752214880


tomhughes:
Looks good to me, thanks.

PR #4901, opened 2024-06-15
Use svg symbols for routing icons

AntonKhorev:

PR #5080, opened 2024-08-15
Browse icons as svg

mmd-osm:
The browse icons haven't been updated for quite some time and are still served as PNG icons. As the icons should reflect the current carto style, I copied a fair amount to this PR, and added some coloring similar to [amenity-points.mss](https://github.com/gravitystorm/openstreetmap-carto/blob/master/style/amenity-points.mss).

One of the design goals for the conversion was to use osm-carto svg files as-is, and at the same time use svg files as < img  >, rather than inlining them. This has some implications on coloring: color values need to be translated to a filter, for which i used https://blog.union.io/code/2017/08/10/img-svg-fill/ + https://stackoverflow.com/questions/24933430/img-src-svg-changing-the-styles-with-css/52041765#52041765

Before converting more icons in browse.scss + common.scss, I wanted to get some feedback if the overall approach is sound. As an example, for the fill parameter I tried "#666666" as parameter value, as well as the original color descriptions from osm-cargo (e.g. "amenity-brown"). I find the latter a bit easier to understand.

Please take a look, and let me know how we should proceed.

image

image


AntonKhorev:
> As the icons should reflect the current carto style

Should they?

AntonKhorev:
> use svg files as < img >, rather than inlining them

Why not inline `<symbol>`/`<use>` with colors defined as css custom properties, which you can redefine in dark mode?

mmd-osm:
This seems like a good idea. Would you use some Gem for it, like https://github.com/jamesmartin/inline_svg and change the browse view accordingly? Do we already have something similar in the code base today, which we could reuse?

mmd-osm:
@AntonKhorev : Just for clarification: is this the approach you had in mind? https://stackoverflow.com/a/68569598/3501889

I feel like the inline_svg gem is loading the file when rendering the page and injecting the contents into the HTML document, which is not great for caching.

AntonKhorev:
I wanted inline svg with a bunch of `<symbol>`s added once to every browse page and inline svgs with `<use>`s referring to these symbols. I wanted to try this approach with routing icons first because there are fewer of them. I partially did this in https://github.com/openstreetmap/openstreetmap-website/pull/4901: https://github.com/openstreetmap/openstreetmap-website/blob/59571f8a8eda2723dcf6546e7a94ae1d73d4aba1/app/views/directions/search.html.erb#L3-L78

mmd-osm:
Thank you for clarifying! I think it would be a bit challenging to always include all symbol svg files in the browse page due to their total size (around 200K).

I thought about adding some configuration table to move most of the logic from browse.scss to a yml file (at least for svg). For this purpose I have introduced a new config/browse_image.yml, and the corresponding initializer in config/initializers/browse_image.rb.

In some cases the same image is used for different tags, so we would need some sort of configuration anyway to map tags to svg images and define how they should be filled.

```
irb(main):001> BROWSE_IMAGE["shop_pet".to_sym]
=> {:image=>"shop_pet.svg", :fill=>"shop"}
```

In the next step BROWSE_IMAGE can then be used in helpers/browse_helper.rb to find out if an SVG exists for a "{key}_{value}", and then use the external_svg approach I mentioned in my previous post.

mmd-osm:
I believe the code should be good for another review. The screenshot below shows mixed svg/png rendering (1 png, rest is svg). The exact symbol location probably need some further tweaking in CSS. Also, the code in browser_helper.rb might need some further improvements. Focus right now should be that we're ok with the approach.

image

AntonKhorev:
> I think it would be a bit challenging to always include all symbol svg files in the browse page due to their total size (around 200K).

If they specify path coordinates to a millionth fraction of a pixel, it's not surprising.

mmd-osm:
Oh well, a lot of effort was already spent to optimize svg files, see https://github.com/gravitystorm/openstreetmap-carto/pull/4457 ... 

Since I'm loading svg files on demand only and cache them in the browser, that's not really a concern.

AntonKhorev:
Do all of the svg icons consist of a single path to be displayed with fill and no stroke?

mmd-osm:
Most of them have a single path, some of the tourism_* ones have 2, and amenity_bureau_de_change.svg has 4. Very few like shop_bookmaker.svg and natural_spring.svg have a stroke param in there.

I'm wondering how this is relevant, since I'm using the original carto svgs and don't really looked inside of them.

mmd-osm:
Ah, good catch. I think element_single_current_link in browser_helper.rb needs a similar logic as we have in element_list_item already.

AntonKhorev:
> Very few like shop_bookmaker.svg and natural_spring.svg have a stroke param in there.
> I'm wondering how this is relevant, since I'm using the original carto svgs and didn't really look inside them

If they actually use strokes, how would you know that their color is correct? natural_spring.svg has a white outline which you won't be able to see over a white background. shop_bookmaker.svg actually doesn't use stroke.

mmd-osm:
Ok, I have updated element_single_current_link now. There's probably a better way to do the block lambda expression below, but at least the icons are showing up now.

```ruby
      block = lambda {
        printable_element_name object
      }
```

image


image

AntonKhorev:
> Most of them have a single path

If all of them had a single path, you'd read those paths from the helper and generate `<svg><path d=#{path_data} /></svg>`.

mmd-osm:
I don't use natural_spring.svg at the moment. Also, I tested a fair amount of icons locally, to see if they look ok.

mmd-osm:
> If all of them had a single path, you'd read those paths from the helper and generate <svg><path d=#{path_data} /></svg>.

Right. Then I need to start parsing svg contents on the server and render them on the backend. Now I simply refer to some external svg file.

mmd-osm:
@AntonKhorev : I think I'm done now with my changes. Anything else we need to check before merging?

AntonKhorev:
Some icons before/after in dark mode:

image image ; image image ; image image

- some icons are too dark, I inverted and hue-rotated them; here it's not done I suppose; see place_of_worship icon for an extreme example
- gate icon has white background
- do we need a smaller icon for waste baskets here just because it's smaller on the map render?

app/assets/stylesheets/common.scss
```diff
@@ -945,6 +945,76 @@ img.trace_image {
...
     margin-left: 25px;
   }
+
+  .svg_icon {
+    overflow: hidden;
+    display: inline-block;
+    margin-left: -25px;
+    width: 25px;
+    height: 18px;
+    /*rtl:ignore*/ transform: translate(2px, 0px);
```
AntonKhorev, line 955: What's this `translate` for?

Old css background image above, new svg image below:
image

AntonKhorev:
I still think it's [easier](https://github.com/openstreetmap/openstreetmap-website/pull/5080#issuecomment-2298987792) to switch to ``s first, then start replacing them with svgs, assuming all of the images have the same size. Hopefully we can avoid random pixel offsets in css.

mmd-osm:
Thank you for reviewing this PR another time.

> some icons are too dark, I inverted and hue-rotated them; here it's not done I suppose; see place_of_worship icon for an extreme example

Agree, that's also something I've noticed. I'm pushing another commit now which should improve the situation. As a general remark, I would leave that sort of fine tuning to a follow up pull request.

> do we need a smaller icon for waste baskets here just because it's smaller on the map render?

Yes, it matches exactly the size you see on the map.

> alignment wasn't great before, here it's even worse, see traffic lights for example

I've checked all 219 SVGs, and found that less than 10 of them might need some fine tuning. The vast majority of new SVGs is just fine. Hence, I think that's a good topic for follow up pull requests as well. 

> I still think it's https://github.com/openstreetmap/openstreetmap-website/pull/5080#issuecomment-2298987792 to switch to s first, then start replacing them with svgs, assuming all of the images have the same size. Hopefully we can avoid random pixel offsets in css.

I'm not entirely clear what you're suggesting here. Assuming we need convert SVGs to PNGs and maintain new entries in browse.scss for them, that would be a very clear no-go for me. Speaking of experience in this PR, it's creating a huge effort with very limited options for later reuse. 

I would suggest to focus on any remaining import issues with SVGs instead in this PR, rather than spending time on PNGs which we want to phase out anyway.

app/assets/stylesheets/common.scss
mmd-osm: The purpose of the translate is to improve the positioning of SVGs. Like I've written below, that's good for the majority of new SVGs.

The old css PNG images will be replaced by SVGs in later PRs, by which time this issue will become obsolete. Until then, some fine tuning could still be done in follow up PRs.

AntonKhorev:
>> do we need a smaller icon for waste baskets here just because it's smaller on the map render?

> Yes, it matches exactly the size you see on the map.

No, we don't need it to match *exactly the size you see on the (one particular rendering of the) map*. We need to decide what size the icons should be and make them that size.

AntonKhorev:
> Assuming we need to convert SVGs to PNGs

I'm not suggesting that.

mmd-osm:
I've added dedicated css formatting for 9 SVG symbols now, to improve positioning and/or size of the respective symbols. Maybe you could take another look now. If necessary, please suggest exactly which symbol should be changed and how.

The approach should be flexible enough to allow for further fine-tuning in the future.

app/assets/stylesheets/common.scss
mmd-osm: highway=turning_circle has been replaced by an SVG symbol.

mmd-osm:
> > Assuming we need to convert SVGs to PNGs
> 
> I'm not suggesting that.

Ok, I think you would need to further elaborate on your proposal, because it’s still not clear to me how this looks in detail. Please provide some working code snippets which addresses both position and coloring of svg. Also some comment why it is better or with issues it solves. We can then look into this in more detail and work out the best approach.

mmd-osm:
Here's another comparison at 250% zoom. The old rendering looked a bit off, in particular for the _(part of ways...)_ section.

Old | New
-- | --
old1  | new1


_As mentioned earlier, the new icons + colors are based on https://wiki.openstreetmap.org/wiki/OpenStreetMap_Carto/Symbols. It is expected that some of them differ from the old ones._

AntonKhorev:
> Ok, I think you would need to further elaborate on your proposal

#5385 as a first step in replacing css backgrounds with images. See how I only use one pixel value of 20px / 1.25rem.

> The old rendering looked a bit off, in particular for the (part of ways...) section.

*Part of ways* I'd do them completely differently, see #5317.

mmd-osm:
Closing since there were too many breaking upstream changes in the meantime.

PR #5081, opened 2024-08-15
Rename geocodeCallback to changeCallback, call it from latlon setValue

AntonKhorev:
Another piece of #5064 which changes how callbacks work. The code outside endpoints module shouldn't care if geocoding was performed or not. Previously there was a special callback for completed geocoding, you waited for it and then you built the route. But if you passed coordinates to endpoint.setValue you knew that geocoding may not happen therefore you had to try to build the route right away. Now you just always wait for a callback because it happens even without geocoding.


nenad-vujicic (approved):
It works fine on my side. Thanks!

tomhughes:
Looks good to me, thanks.

PR #5082, opened 2024-08-15
Prevent reactivating inactive blocks

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/5069/files#r1715714691

Let's disable block reactivations first.

image


tomhughes:
Looks good to me, thanks.

PR #5083, opened 2024-08-16
Don't call getGeocode from getRoute

AntonKhorev:
Part of #5064.

`getRoute` shouldn't know anything about geocoding, doesn't need to wait for every kind of geocoding (namely for reverse one), therefore shouldn't command the endpoints to do it. All necessary geocoding is triggered by `endpoint.setValue` and input event listeners.

### DIFF
--- a/app/assets/javascripts/index/directions-endpoint.js
+++ b/app/assets/javascripts/index/directions-endpoint.js
@@ -43,18 +43,12 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
       setLatLng(latlng);
       setInputValueFromLatLng(latlng);
       changeCallback();
-    } else {
+    } else if (endpoint.value) {
       endpoint.getGeocode();
     }
   };
 
   endpoint.getGeocode = function () {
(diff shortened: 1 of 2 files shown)


tomhughes:
Is there some reason for pulling about #5064 into fragments? I know I skipped reviewing that one - that was because it was a bit larger so I was going to leave it to the weekend...

AntonKhorev:
If you apply `Remove latlng parameter from endpoint.setValue`, you'll have to apply the entire https://github.com/openstreetmap/openstreetmap-website/pull/5064, otherwise reverse geocoding will stop working. But then you may run into precision problems if you don't do also #5084.

tomhughes:
What is `Remove latlng parameter from endpoint.setValue` though? I don't see any PR called that?

AntonKhorev:
First commit in https://github.com/openstreetmap/openstreetmap-website/pull/5064: https://github.com/openstreetmap/openstreetmap-website/pull/5064/commits/b9f8057b40f9da3874921c161033986c406ed47c

tomhughes:
Looks good to me, thanks.

PR #5084, opened 2024-08-16
Change zoom precision function

AntonKhorev:
This changes the number of fractional digits to log10(pixels / degrees) with pixels = 2**(8 + zoom) and degrees = 180.

---

If you apply the entire #5064, zoom in to the max zoom level, place an endpoint marker and try to drag it around, you'll notice it moves by jumping several pixels. This is because I limit its coordinates to zoom precision and zoom precision is not high enough. Its current version was introduced in https://github.com/openstreetmap/openstreetmap-website/commit/b28511faca45d8d8abc0d3c2fde5a501bbe94062. You can notice something strange about it: the zoom value went from being an argument of `Math.pow` to an argument of `Math.log`. `zoom` is already on a log scale, why do you need to take its log again?

| zoom | old zoomPrecision | new zoomPrecision |
| --- | --- | --- |
| 0 | 0 | 1 |
| 1 | 0 | 1 |
| 2 | 1 | 1 |
| 3 | 2 | 2 |
| 4 | 2 | 2 |
| 5 | 3 | 2 |
| 6 | 3 | 2 |
| 7 | 3 | 3 |
| 8 | 3 | 3 |
| 9 | 4 | 3 |
| 10 | 4 | 4 |
| 11 | 4 | 4 |
| 12 | 4 | 4 |
| 13 | 4 | 5 |
| 14 | 4 | 5 |
| 15 | 4 | 5 |
| 16 | 4 | 5 |
| 17 | 5 | 6 |
| 18 | 5 | 6 |
| 19 | 5 | 6 |
| 20 | 5 | 7 |
| 21 | 5 | 7 |

Some layers go up to zoom 21, precision limits become very noticeable there. Even on level 0 you need at least one digit after `.` to translate -90..+90 to 256 tile pixels.

### DIFF
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -204,7 +204,9 @@ OSM = {
   },
 
   zoomPrecision: function(zoom) {
-    return Math.max(0, Math.ceil(Math.log(zoom) / Math.LN2));
+    var pixels = Math.pow(2, 8 + zoom);
+    var degrees = 180;
+    return Math.ceil(Math.log10(pixels / degrees));
   },
 
   locationCookie: function(map) {


tomhughes:
I'm not disputing there may be a problem here but it would be good to understand how you have derived the new formula you're using - the fact that the output repeats some values 3 times and some 4 times seems a little to me and suggests that the formula is slightly off somewhere causing rounding issues?

AntonKhorev:
`ceil( log10(2**(8+zoom) / 180) )`

`2**(8+zoom)` = world size in pixels
`180` = span of latitude values
`2**(8+zoom) / 180` = number of pixels in one degree. If there are 100 pixels in one degree, you need log10(100) = 2 fractional digits to represent a degree value for a given pixel offset

AntonKhorev:
> repeats some values 3 times and some 4

Because numbers are base-ten not base-some-power-of-2 and zoom levels magnify by powers of 2.

AntonKhorev:
If you don't mind raising something to a variable power* and taking a logarithm of the result, here's a more readable version.

\* could have used `pixels = 1 << (8 + zoom)`, not sure if it's more clear

tomhughes:
Thanks. I think that makes it a lot clearer what is going on for future maintainer.

PR #5085, opened 2024-08-16
Remove duration and needs_view inputs from inactive block edit form

AntonKhorev:
Since #5082 removed the ability to reactivate blocks, we can remove the inputs for doing that too. Inactive blocks will have only editable reason input.

image


tomhughes:
Looks good to me, thanks.

PR #5086, opened 2024-08-16
Add navigation tabs to edit block pages

AntonKhorev:
Before:
image

After:
image

The last tab leads to the block show page and is selected even if we're on edit page. I don't think it's worth it to add another tab just for the edit page.


tomhughes:
Looks good to me, thanks.

PR #5087, opened 2024-08-16
Truncate long usernames in blocks tables

AntonKhorev:
Before:
image

After:
image

### DIFF
--- a/app/views/user_blocks/_block.html.erb
+++ b/app/views/user_blocks/_block.html.erb
@@ -1,9 +1,9 @@
 <tr>
   <% if show_user_name %>
-  <td><%= link_to block.user.display_name, block.user %></td>
+  <td><%= link_to block.user.display_name, block.user, :class => "username d-inline-block text-truncate text-wrap" %></td>
   <% end %>
   <% if show_creator_name %>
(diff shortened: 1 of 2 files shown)


app/views/user_blocks/_block.html.erb
```diff
@@ -1,9 +1,9 @@
 <tr>
   <% if show_user_name %>
-  <td><%= link_to block.user.display_name, block.user %></td>
+  <td><%= link_to block.user.display_name, block.user, :class => "username d-inline-block text-truncate text-wrap" %></td>
   <% end %>
   <% if show_creator_name %>
-  <td><%= link_to block.creator.display_name, block.creator %></td>
+  <td><%= link_to block.creator.display_name, block.creator, :class => "username d-inline-block text-truncate text-wrap" %></td>
```
tordans, line 6: I always like it when there is a title attribute on truncated stings so I have a quick way go see the whole thing. In this case maybe on the TD because that gives a bigger hover area.

AntonKhorev: This is a link. Your browser will show it on mouseover somewhere.

tomhughes:
Looks good to me, thanks.

PR #5088, opened 2024-08-16
Copy "From" endpoint value to search input when closing directions

AntonKhorev:
When you close the directions form there's this code that tries to use the "From" value for searching.
image

It does so by opening the url `/?query=...`. The url is wrong, it should be `/search?query=...`, thus nothing happens.

I could have fixed it by using `/search?query=...`, but I don't like that it runs the search right away, when the close button is clicked. Instead I set the search input value.

image


tomhughes:
Looks good to me, thanks.

PR #5089, opened 2024-08-17
Don't update ends_at when editing inactive blocks

AntonKhorev:
If you edit an inactive block, you can change its text but you can't reactivate it. You are prevented from reactivating it by keeping the block duration at 0. But even setting the duration to 0 changes the block end time, because it's set to now + duration. This in turn affects the duration as reported on the block show page, that shows duration = end_at - created_at. You wanted to just update the text, yet you altered the duration too. That's not something you expect to happen.

This PR keeps old ends_at time when updating inactive blocks.

Example with 0-hour block edited months later:

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5090, opened 2024-08-17
Add navigation tabs to new block pages

AntonKhorev:
Similar to #5086

Before:
image
...
image

After:
image
...
image

Related to #3267 where back buttons get outlines. This PR removes one of these buttons.

### DIFF
--- a/app/views/user_blocks/_navigation.html.erb
+++ b/app/views/user_blocks/_navigation.html.erb
@@ -34,10 +34,16 @@
                   :class => ["nav-link", { :active => action_name == "blocks_by" }] %>
     </li>
   <% end %>
-  <% if @user_block %>
+  <% if @user_block&.persisted? %>
     <li class="nav-item">
       <%= link_to t(".block", :id => @user_block.id),
                   user_block_path(@user_block),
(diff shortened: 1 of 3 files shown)


app/views/user_blocks/_navigation.html.erb
```diff
@@ -34,11 +34,18 @@
...
   <% end %>
-  <% if @user_block %>
+  <% if @user_block&.persisted? %>
     <li class="nav-item">
       <%= link_to t(".block", :id => @user_block.id),
                   user_block_path(@user_block),
                   :class => "nav-link active" %>
     </li>
   <% end %>
+  <% if @user_block&.new_record? %>
```
tomhughes, line 44: This could just be `elsif` instead of starting a new if?

tomhughes (changes requested):

app/views/user_blocks/_navigation.html.erb
AntonKhorev: changed to `elsif`

tomhughes (approved):
Looks good to me now, thanks.

PR #5091, opened 2024-08-17
Remove /user_blocks/new route

AntonKhorev:
`/user_blocks/new` doesn't work because there's no username


tomhughes:
Looks good to me, but I wonder if we should in fact replace the non-standard `/blocks/new/:display_name` with `/user_blocks/new?display_name=`?

PR #5092, opened 2024-08-17
Define diary entry hide/unhide routes using post on members

AntonKhorev:
should be easier to read


tomhughes:
Looks good to me, thanks.

PR #5093, opened 2024-08-17
Use shallow routes for diary comments

AntonKhorev:
Changes routes from 

`/user/:display_name/diary/:id/comments/:comment/hide`

to

`/diary_comments/:comment/hide`

\+ the same for unhide.

This is what you'll get if you define comments as a shallow nested resource like this:

```ruby
  resources :users, :path => "user", :param => :display_name do
    resources :diary_entries, :path => "diary" do
      resources :diary_comments, :shallow => true do
        ...
      end
    end
  end
```

Paths with display_name and diary entry id are too long and these names/ids are not used for anything in the controller.

### DIFF
--- a/app/views/diary_entries/_diary_comment.html.erb
+++ b/app/views/diary_entries/_diary_comment.html.erb
@@ -13,9 +13,9 @@
     <% if can? :hide, DiaryComment %>
       <span>
         <% if diary_comment.visible? %>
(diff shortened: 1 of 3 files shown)


nenad-vujicic (approved):

kcne:
This significantly simplifies the routes for hiding and unhiding diary comments by removing redundant path parameters, making the code cleaner and easier to maintain. Thank you.

kcne (approved):

gravitystorm:
Merged, thanks! I agree that shallow routes are a good idea when there's otherwise a lot of unnecessary url nesting.

PR #5094, opened 2024-08-19
Record revokers when deactivating blocks by editing

AntonKhorev:
You can deactivate a block using the revoke action. But you can also deactivate a block by editing it and setting duration=0 + needs_view=false. The difference is that you're not being recorded as a revoker if you do that. This PR records you as a revoker in this case too.

This kind of editing is only available to you if you're the block creator. The point of this PR is to relax this check later and let any moderator revoke blocks by editing, then remove the revoke action. Instead of first going to the revoke confirmation page with "Are you sure you wish to revoke this block?" and then possibly going to the edit page, moderators will be able to edit the block reason and revoke the block from the same page.


tomhughes:
Looks good to me, thanks.

PR #5095, opened 2024-08-19
Add enable/disable endpoint methods

AntonKhorev:
Part of #5064. I'm pulling it out because it's also required for #4900.

The noticeable difference after this PR should be geocoding request cancelling when leaving the directions page. Previously you could pick "Directions from/to here" then quickly close the directions form or go to any other map layout page. The geocoding request could terminate after that, placing an endpoint marker on the map despite not being on the directions page.

### DIFF
--- a/app/assets/javascripts/index/directions-endpoint.js
+++ b/app/assets/javascripts/index/directions-endpoint.js
@@ -25,6 +25,8 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, iconUrl, dragCallback, ch
     input.off("keydown", inputKeydownListener);
     input.off("change", inputChangeListener);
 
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

HolgerJeromin:
Merged as https://github.com/openstreetmap/openstreetmap-website/commit/99a7d21a9bc50cdb38559dd0a8b60bc60072b5a1

tomhughes:
Well actually it was the earlier version which was merged which is why github didn't detect the merge... Now I need to figure out the difference...

tomhughes:
Remerged as c5aa012dc252414127f4c3d2f724c7f8e384390a which I think just adds a missing abort call - something that I nearly asked about before the first merge!

PR #5096, opened 2024-08-19
Show/hide "Where is this?" when setting search value from directions

AntonKhorev:
Fix for #5088.

If the "From" value is long enough [like this](https://www.openstreetmap.org/directions?from=7%2C%20Bell%20Yard%2C%20Gray%27s%20Inn%2C%20Holborn%2C%20London%2C%20Greater%20London%2C%20England%2C%20WC2A%202JR%2C%20United%20Kingdom) and you close the directions form, you'll get the search value overlapping with "Where is this?".

image
image

Also sometimes it's possible to delete the value, close the directions and get a blank search input without "Where is this?".


tomhughes:
Looks good to me, thanks.

PR #5099, opened 2024-08-20
Always use coordinates when using directions from context menu

AntonKhorev:
Part of the problem #5064 tries to solve are the result of the context menu using addresses from endpoint inputs.

You right click somewhere, select "Directions from here" and it goes to and url like this: `/directions?from=51.420436%2C-0.174805&to=`. Then you right click somewhere else, select "Directions to here" and the url becomes something like this: `/directions?from=Park%20Road%2C%20Collier%27s%20Wood%2C%20London%20Borough%20of%20Merton%2C%20London%2C%20Greater%20London%2C%20England%2C%20SW19%202HT%2C%20United%20Kingdom&to=51.420563%2C-0.171007`. The coordinates in `from` got replaced by an address. It's impossible to get the original coordinates from an address.

The solution here is to save the coordinates to `data-lat` and `data-lon` attributes of inputs. The context menu then can generate directions urls using the values stored there.

This doesn't solve the entire problem because there's also uncontrolled reverse geocoding decided by Nominatim when using its freeform query.

### DIFF
--- a/app/assets/javascripts/index/contextmenu.js
+++ b/app/assets/javascripts/index/contextmenu.js
@@ -11,7 +11,7 @@ OSM.initializeContextMenu = function (map) {
 
       OSM.router.route("/directions?" + Qs.stringify({
         from: lat + "," + lng,
-        to: $("#route_to").val()
+        to: getDirectionsEndpointCoordinatesFromInput($("#route_to"))
       }));
     }
   });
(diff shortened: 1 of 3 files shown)


nenad-vujicic (approved):

tomhughes:
Looks good to me, thanks.

PR #5100, opened 2024-08-20
Add overcommit to manage git hooks

tomhughes:
As suggested in https://github.com/openstreetmap/openstreetmap-website/pull/4988#issuecomment-2285568484 this adds support overcommit to install git gooks to run checks.

With this change `bundle exec overcommit --install` will install hooks to run a variety of checks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .DS_Store
 .bundle/config
 .idea
+.local-overcommit.yml
 .ruby-gemset
 .ruby-version
 .vagrant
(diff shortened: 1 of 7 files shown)


mmd-osm:
We could add a few words in CONTRIBUTING.md on how to enable these checks, otherwise it will be difficult for new folks to discover this feature.

tomhughes:
Added...

gravitystorm:
I've been trying this out today, and it works much better than I thought it was going to! In particular, I thought the precommit hooks would be too slow to be worthwhile, but clearly it's been optimised to make rubocop target only changed files, or similar.

I'm happy to merge this, thanks!

tomhughes:
Oh yes it only targets changed files in the pre-commit checks - the only slow thing is the brakeman check in the pre-push handler which I was a bit less sure about.

gravitystorm:
I'd be happy to disable that if it's slow - it's not something that most people will run into (compared with the other checks) and it's picked up by CI anyway.

Issue #3887, opened 2023-01-10
Provide a summary view of recent diaries


yvecai:
The page showing the latest diary entries (https://www.openstreetmap.org/diary) is quite long to browse as more than often some of them look like full-fledged academic papers. 
Sometimes an entry generates discussion in the comments for a week or more, and it's not easy to get back to it browsing this page.
It would be nice IMO to provide only a partial view (10 lines or so?) for each entry.
Could be collapsible parts in the page or a truncated version with a link to the single diary.

tomhughes:
I thought there was already an issue for this though I can't find it right now.

I have to say that I don't think we ever really expected people to write such long essays in their diary posts...

It's tricky because personally I find it annoying to have to keep clicking back and forth between the index and the posts yet I also find it annoying when there is suddenly a massive post in the index.

yvecai:
I did search and found nothing. 

Another way around is maybe a simple <detail> html tag for each entry that would be already nice to quickly collapse entries.
Not sure of the implementation details and if the diary entries would be compatible, though.

mmd-osm:
Probably a duplicate of #3721, although that one also talks about RSS feeds.

yvecai:
I obtained the mockup below from a local copy of today's page.
Just added \<**details**/> tags around each diary post, and \<**summary**/> around the post header with _Find&Replace_. Any of the 20 posts from today expands in the page flawlessly ! 

![Capture d’écran du 2023-01-11 21-15-38](https://user-images.githubusercontent.com/506919/211909750-235f1453-e843-4f4c-8568-7dc9c70083f2.png)

I could certainly be happy with this ! 

Though I admit the navigation is not straightforward: clicking on the post title will open the diary link, while clicking everywhere else on the post header (avatar , text or blank space) will expand the content.

Issue #3721, opened 2022-09-25
user diary pages (html and rss) contain all entries in their entirety


mfrasca:
When you open a personal diary page you will get the whole text of all entries. Also if you open the RSS feed, the "description" entity contains the whole text of the post.

This, in particular if it's by a prolific author, can be quite a long page. My own 'mariotomo' page is around 140kbytes, just the html. 

These complete user's diary pages are not easy to navigate. Some entries are short, some longer, some are extensive articles, possibly holding quite a few images. Finding the beginning of the next entry amounts to hitting "search in page" and type the text "Posted by".

I would like to see something like a "normal" rss feed, with the description cut at some 200-500 chars, possibly with the option to reduce the entries to the title, or include the (shortened) description.

tomhughes:
No, no, no.

Full text vs truncated text for RSS is an old, old argument and truncated text is almost always used only by sites that are monetising content either by requiring an account to read the full content or that just want you to see the adverts.

I absolutely hate it!

I want to see the full text in my feed reader so that I don't have to actually open the site.

mfrasca:
> I want to see the full text in my feed reader so that I don't have to actually open the site.

that's a funny point of view, let's say I can understand it, not agree, that not, but let's accept it as a legitimate point of view.

what about the html page?

AntonKhorev:
Have two different feeds, full and abridged?

My diary page is ~200kb compessed. I have a table of contents on my profile page to find my diary entries.

PR #5103, opened 2024-08-21
Provide a summary view of recent diaries

nertc:
This PR addresses "Provide a summary view of recent diaries" issue mentioned in the https://github.com/openstreetmap/openstreetmap-website/issues/3887

[Bootstrap Accordion](https://getbootstrap.com/docs/5.2/components/accordion/) was added to the diary entries list pages. Now the content of the diary entry will be collapsed until toggle button is clicked. View of the diary entry details (show) page remains the same. Fixes https://github.com/openstreetmap/openstreetmap-website/issues/3887

Screenshots:
Screenshot 2024-08-26 122446
Screenshot 2024-08-26 122459
Screenshot 2024-08-26 122517

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -46,6 +46,14 @@ time[title] {
 .icon.note.grey   { /*rtl:ignore*/ background-position: -240px -20px; }
 .icon.query       { /*rtl:ignore*/ background-position: -260px 0; }
 
+.mh-10-collapsed {
+  transition: none;
+
+  &.collapse:not(.show) {
+    max-height: 10rem;
+  }
+}
+
 /* Utility for de-emphasizing content */
 
 .text-body-secondary a {
--- a/app/controllers/diary_entries_controller.rb
+++ b/app/controllers/diary_entries_controller.rb
@@ -15,6 +15,10 @@ class DiaryEntriesController < ApplicationController
 
   allow_thirdparty_images :only => [:new, :create, :edit, :update, :index, :show]
 
+  content_security_policy(:only => :index) do |policy|
+    policy.img_src(*policy.img_src, "data:")
+  end
+
   def index
     if params[:display_name]
       @user = User.active.find_by(:display_name => params[:display_name])
--- a/app/views/diary_entries/_diary_entry.html.erb
+++ b/app/views/diary_entries/_diary_entry.html.erb
@@ -1,7 +1,6 @@
 <article class='diary_post border-top border-secondary-subtle py-3<%= " text-body-secondary px-3 bg-danger bg-opacity-10" unless diary_entry.visible %> user_<%= diary_entry.user.id %>'>
-  <%= render :partial => "diary_entry_heading", :object => diary_entry, :as => "diary_entry" %>
-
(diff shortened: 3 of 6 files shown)


nertc:
Failed test is out of scope of this PR, as there was no change or impact related to the test that was failed.

nertc:
This PR also partially refers to https://github.com/openstreetmap/openstreetmap-website/issues/3721 (changes HTML view of the diary entries list, but doesn't change RSS)

mmd-osm:
I haven't tested this code yet, though it seems that toggling a diary entry would either show the whole text, or only the title. I believe #3887 talked about showing the first few _n_ lines of the body, with an option to further expand the body of the diary. Would this be possible to implement?

tomhughes:
Yes this starts with all entries collapsed to just their headings.

It also fails if you have CSP enforcement enabled because bootstrap is using `data:` urls for the arrow images so they don't appear unless you allow that in the controller with:

```ruby
content_security_policy(:only => :index) do |policy|
  policy.img_src(*policy.img_src, "data:")
end
```

nertc:
@mmd-osm Yes, it is possible. I made this change according to the last screen mentioned in the Issue https://github.com/openstreetmap/openstreetmap-website/issues/3887. I'll change this PR to show part of the content even when it is collapsed.

@tomhughes I'll take a look if there is any way to make it work without custom code.

nertc:
@mmd-osm I updated PR so that now `10 rem` (about 7.5 paragraph lines) of content is shown when the entry is collapsed. If entry is shorter, it doesn't leave any redundant gaps. I needed to add 2 lines of custom CSS as Bootstrap Accordion doesn't support behavior of showing any content when collapsed (it sets `display: none`). The only downside of the current solution is that when entry is less than 7 lines, accordion button is still shown but clicking it doesn't change anything. The only fix (based on limited functionality of Bootstrap Accordion) can be made by adding custom JS code. From my experience, we try to avoid adding additional JS code. If it is okay to add one for this functionality, I can add it too.

@tomhughes I implemented solution suggested by you. I found some issues opened about Bootstrap CSP errors (https://github.com/twbs/bootstrap/issues/25394), but even [Bootstrap 5 mentions](https://getbootstrap.com/docs/5.0/customize/overview/#csps-and-embedded-svgs) that some of their functionality uses `data:` in image urls. As mentioned in [Stackexchange](https://security.stackexchange.com/questions/94993/is-including-the-data-scheme-in-your-content-security-policy-safe/167244#167244) and [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_as_an_Image#restrictions) adding `data:` as `background-image` or other image source doesn't raise any risks. Therefore, if we don't have extremely strict CSP policy, we can use solution suggested by you.

Thank you for reviewing

tomhughes:
I guess the question is do we really want/need the collapse and expand functionality or is it better just to show a truncated post and let people click through if they want to see the whole thing, which I think is a more typical way to handle this sort of thing.

nertc:
I created a separate PR for the truncated version of the entry https://github.com/openstreetmap/openstreetmap-website/pull/5121. If we decide to truncate diary entries instead of showing summary of them, I'll close this PR.

nertc:
Closing this PR in favor of #5121 PR

PR #5104, opened 2024-08-21
Edit unrevoked block

AntonKhorev:
Continues #5094.

Moderators who didn't create an active block can open its edit page and the form for them will look like this:
image

They can update the reason and click revoke. They'll be recorded as revokers and the block will be deactivated.

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -60,7 +60,10 @@ class Ability
           can [:index, :show, :resolve, :ignore, :reopen], Issue
           can :create, IssueComment
           can [:new, :create, :edit, :update, :destroy], Redaction
-          can [:new, :edit, :create, :update, :revoke, :revoke_all], UserBlock
+          can [:new, :create, :revoke, :revoke_all], UserBlock
+          can :update, UserBlock, :creator => user
+          can :update, UserBlock, :revoker => user
+          can :update, UserBlock, :active? => true
         end
 
         if user.administrator?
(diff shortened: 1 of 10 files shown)


app/abilities/ability.rb
```diff
@@ -60,7 +60,10 @@ def initialize(user)
           can [:index, :show, :resolve, :ignore, :reopen], Issue
           can :create, IssueComment
           can [:new, :create, :edit, :update, :destroy], Redaction
-          can [:new, :edit, :create, :update, :revoke, :revoke_all], UserBlock
+          can [:new, :create, :revoke, :revoke_all], UserBlock
+          can :update, UserBlock, :creator => user
+          can :update, UserBlock, :revoker => user
+          can :update, UserBlock, :active? => true
```
tomhughes, line 66: Does this act as an AND or an OR of the conditions? Because neither seems right? Surely we want active AND (creator OR revoker) as the test?

tomhughes (commented):
Why exactly do we want to stop other moderators editing unless they also revoke?

app/abilities/ability.rb
AntonKhorev: It's [OR](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/combine_abilities.md) and that's what we want. Before this PR you had to be either a creator or a revoker to edit. Now additionally you can become a revoker by editing. If the block is active, it has no revoker yet. Previously you had to first revoke it, that made you a revoker, then you could edit its message. Now you can edit its message and revoke it at the same time.

tomhughes: Thanks for the explanation. I had missed that this was inside a moderator test and so only applied to moderators...

tomhughes:
Looks good to me, thanks.

PR #5106, opened 2024-08-22
Backfill block deactivates_at

AntonKhorev:
Continues #5069.

I wrote it [like strong_migrations wants](https://github.com/ankane/strong_migrations?tab=readme-ov-file#backfilling-data).

After the backfilling is done this should be empty:

    SELECT * FROM user_blocks WHERE needs_view = (deactivates_at IS NOT NULL);

A block either needs view or has a deactivation date.


tomhughes:
Was the `sleep` something that strong migrations actually insisted on or did you just take that manually from their "best practices" nonsense?

AntonKhorev:
I didn't check if it complains without sleep because with the amount of blocks we have it shouldn't make much difference.

tomhughes:
Well yes this case would be fine with a simple `update_all` but strong migrations doesn't allow us to make human judgements... We've never done the sleep thing before though and it looks horribly cargo cultish to me especially a sleep so short that it probably takes more time than that for ruby to setup the next update anyway!

mmd-osm:
Found some discussion on the topic in https://github.com/ankane/strong_migrations/issues/163

TL;DR: they want to avoid CPU spikes on the DB server by flooding the server with tons of requests in a short amount of time.

tomhughes:
Interesting because I assumed it was about creating a window between queries to let other queries that might be waiting get in... I can't imagine a sleep that short will have any meaningful effect on CPU load.

gravitystorm:
Adding `sleep` feels a bit cargo-culty to me too. The original [documentation in ActiveRecord](https://github.com/rails/rails/commit/d13623ca46c82120c398f4634e206422fc3ad7ea0) has been there since 2009, and the world of databases has changed a lot since then (e.g. PCIe, SSD), and even the example hinted more at a non-realistic scenario (the example uses a 50s wait between batches, in the context of people entering a club).

I'm also not bothered by "CPU spikes". If the db has available CPU capacity then lets use it. So the only situation would be if the queries from the migration completes saturates the CPU (or disk IO, or networking, or similar). In this case it certainly won't, but perhaps it would happen if we e.g. did `Node.in_batches`. But in that case, I think `0.01` is a number plucked from thin air, and if the database can't handle the normal queries plus the updates, then I think backpressure from the db (i.e. these queries taking longer to complete) will regulate the migration better than arbitrary sleeps. The whole point of using batches is to give the db repeated opportunities to run other queries, it doesn't need a gap to choose the next in its queue.

So my recommendation is to remove the sleep, so that the next time a contributor is looking to do something similar, they are less likely to include it in their migration too.

mmd-osm:
I think we need to look at the big picture. During heavy revert activity due to vandalism we have seen last year that our database replicas lag behind several hours: https://prometheus.osm.org/d/ST-7bi5Gz/api-database?orgId=1&from=1699142400000&to=1699315199000&viewPanel=8

To avoid a similar situation due to backfilling, some sleep delay might be useful. 0.01 seems a bit meaningless to me, though good values are difficult to guess. That’s where additional testing could come in handy.

tomhughes:
What on earth does that have to do with anything? We're talking about load on the master not the size of the WAL logs which is largely what causes replication lag.

mmd-osm:
Yes, I know. For the sake of managing the load on the master, adding some extra sleep is almost certainly not required given the available hardware.

The point I wanted to make is that other parts (like replication) may have constraints that would benefit from additional delays.

tomhughes:
I've dropped the sleep and will merge this now - thanks for working on it.

PR #5107, opened 2024-08-22
Convert dashboard test into a system test

AntonKhorev:
Addresses https://github.com/openstreetmap/openstreetmap-website/pull/5070#issuecomment-2304501335 about `dashboards_controller_test.rb`. I converted the test into a system test because this way it's easier to test for the displayed content. I'm checking for "You have not added any friends" where previously was no assertion. Also I don't rely on a html element id that was used only for testing.

### DIFF
--- a/test/system/dashboard_test.rb
+++ b/test/system/dashboard_test.rb
@@ -1,0 +1,27 @@
+require "application_system_test_case"
+
+class DashboardSystemTest < ApplicationSystemTestCase
+  test "show no users if have no friends" do
+    user = create(:user)
+    sign_in_as(user)
+
+    visit dashboard_path
+    assert_text "You have not added any friends yet."
+  end
+
+  test "show users if have friends" do
+    user = create(:user, :home_lon => 1.1, :home_lat => 1.1)
+    friend_user = create(:user, :home_lon => 1.2, :home_lat => 1.2)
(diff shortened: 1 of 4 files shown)


test/system/dashboard_test.rb
```diff
@@ -0,0 +1,27 @@
+require "application_system_test_case"
+
+class DashboardSystemTest < ApplicationSystemTestCase
```
tomhughes, line 3: Shouldn't this just be `DashboardTest` to match the filename?

AntonKhorev: I had problems with some system test previously that had the same name as some other test. I did the same thing as https://github.com/openstreetmap/openstreetmap-website/commit/45e6ebcafc5741750357b7fd6e11ec96c2cbfb24.

AntonKhorev: We have test named after:
- a thing in singular
- a thing in plural
- an action as a noun
- an action as a verb

In the [guides](https://guides.rubyonrails.org/v7.1/testing.html#system-testing) they have system tests named after things in plural and model tests after things in singular. If we want to do the same, we'll have to rename, for example, `ConfirmationResendSystemTest` to `ConfirmationsTest`.

tomhughes:
Let's merge this now and we can worry about standardising test names later.

PR #5110, opened 2024-08-23
Remove revoke block action

AntonKhorev:
After #5104 you can edit any block you can revoke, therefore you don't need a dedicated revoke page/action. Revoke pages are worse edit pages that don't let you edit the block message.


tomhughes:
Looks good to me, thanks.

PR #5111, opened 2024-08-23
Disable turbo on all links inside pagination frames except older/newer

AntonKhorev:
Something had to break after #4646.

Turbo handles all links inside its frames unless told not to. If any link leads to a page that has some javascript running on load, it's going to be broken. The most common javascript in our case is rich rext controls code. If you go to a diaries page and click "comment", turbo takes you to a diary entry page with broken richtext javascript. And it's not as easy to fix as adding the page onload code to an equivalent turbo event because now you also have to clean up when leaving the page (using a back button for example), something that wasn't possible before (except on our map layout pages with our special router).

I did here the same as in #4565: set `data-turbo="false"` on the entire frame and set it to true just on previous/next page links.

### DIFF
--- a/app/views/diary_comments/_page.html.erb
+++ b/app/views/diary_comments/_page.html.erb
@@ -1,4 +1,4 @@
-<turbo-frame id="pagination" target="_top">
+<turbo-frame id="pagination" target="_top" data-turbo="false">
   <table class="table table-striped" width="100%">
     <thead>
       <tr>
--- a/app/views/diary_entries/_page.html.erb
+++ b/app/views/diary_entries/_page.html.erb
@@ -1,4 +1,4 @@
-<turbo-frame id="pagination" target="_top">
+<turbo-frame id="pagination" target="_top" data-turbo="false">
   <h4><%= t ".recent_entries" %></h4>
 
   <%= render @entries %>
--- a/app/views/shared/_pagination.html.erb
+++ b/app/views/shared/_pagination.html.erb
@@ -7,7 +7,7 @@
     <% end %>
     <% if newer_id -%>
       <li class="page-item d-flex">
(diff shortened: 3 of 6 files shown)


tomhughes:
That sounds like a suboptimal solution - there must be a way to make this work with turbo and to be honest I'm surprised that it doesn't work out of the box on a full page switch with turbo.

AntonKhorev:
Are you surprised we're doing this? https://github.com/openstreetmap/openstreetmap-website/blob/30f975c609bfd05a654ad9facaf13135ca0691cd/app/assets/javascripts/application.js#L72

tomhughes:
No, but I was playing with turning it off the other day, to see what would be required to change that ;-)

AntonKhorev:
Probably rewriting the entire client-side map pages router among other things.

AntonKhorev:
We have links to map pages from traces lists. Click "View map" there, click "back", click "forward", see what happens.

tomhughes:
Yes I got as far as discovering that but did get as far as testing if `turbo_page_requires_reload` on those pages fixed it though it should do.

I guess we need to merge this for now anyway, before the turbo stuff gets deployed.

tomhughes:
It looks like the richtext issue can be fixed with:

```patch
diff --git a/app/assets/javascripts/richtext.js b/app/assets/javascripts/richtext.js
index e069f6f88..d51b8ff96 100644
--- a/app/assets/javascripts/richtext.js
+++ b/app/assets/javascripts/richtext.js
@@ -1,4 +1,4 @@
-$(document).ready(function () {
+$(document).on("turbo:load", function () {
   /*
    * When the text in an edit pane is changed, clear the contents of
    * the associated preview pne so that it will be regenerated when
```

but there's probably a lot more `document.ready` calls that will need to be reviewed as well.

AntonKhorev:
> It looks like the richtext issue can be fixed with

No, it can't be fixed like that, see my first post here. Event listeners on `window` are not going to uninstall themselves.

PR #5115, opened 2024-08-24
Compacted repeated "Results" into one sentence

### LINKED ISSUE #5065: Search Results Results from → Search Results from
### Problem

Perhaps the repeated word "Results" could be compacted into just one instance.

Screenshot 2024-08-11 19 05 38

Yes, even when there are more than one source too.

nenad-vujicic:
PR addresses repeated "Results" in sidebar described in #5065 

Fixes #5065. Replaced "Search Results" with "Search".

Before
image

After
image

Before
image

After
image

### DIFF
--- a/app/views/geocoder/search.html.erb
+++ b/app/views/geocoder/search.html.erb
@@ -4,7 +4,7 @@
 
 <% @sources.each do |source| %>
   <h4>
-    <%= t(".title.results_from_html", :results_link => link_to(t(".title.#{source[:name]}"), source[:url].to_s)) %>
+    <%= link_to t(".title.#{source[:name]}"), source[:url].to_s %>
   </h4>
   <div class="search_results_entry mx-n3" data-href="<%= url_for @params.merge(:action => "search_#{source[:name]}") %>">
     <div class="text-center loader">
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -674,7 +674,6 @@ en:
   geocoder:
     search:
       title:
-        results_from_html: 'Results from %{results_link}'
         latlon: Internal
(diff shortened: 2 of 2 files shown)


mmd-osm:
(sorry, posted in the wrong issue, posting again...)

@jidanni can you maybe take a look at the screenshots above if this looks as expected?

jidanni:
I think it looks great!

AntonKhorev:
But the sidebar is only for search results, not for the entire search functionality. For example, the search form is above the sidebar heading.

For directions the equivalent heading doesn't have the word "result" in it and just reads "Directions", but directions are already a result of routing.

AntonKhorev:
> > But the sidebar is only for search results, not for the entire search functionality. For example, the search form is above the sidebar heading.
> 
> I fully agree with this, but we can do this (displaying the results instead of having entire functionality) in next PR.

I didn't say to do anything about it.

AntonKhorev:
> Version 2: (without Search results)

Let's not remove the heading. There's no other sidebar page with a skipped top heading level.

nenad-vujicic:
Sure! Here is the latest version:

image
image

AntonKhorev (approved):
"Results from ..." were added in https://github.com/openstreetmap/openstreetmap-website/commit/06b2d278ea6c273c7e4e01c6f28bb46505efd6d1. Maybe having this phrase made more sense back then because [results.html.erb](https://github.com/openstreetmap/openstreetmap-website/blob/06b2d278ea6c273c7e4e01c6f28bb46505efd6d1/app/views/geocoder/results.html.erb) didn't have the overall header.

AntonKhorev:
Merged, thanks.

Issue #2644, opened 2020-06-06
[interal message system] request: add potiential forward (function)/button


sun-geo:
Use case: i'm using occasionally (gladly) the OSM.org internal message system for private pn's. Sometimes i get no reply on a message which i did sent somebody for any request or a question regarding a map change. It's known that there are situations that you will not get an answer from every receivers on first request. So i would forward my own message to same person let say 1 or 2 weeks later. For this action i would like to have a forward (function)/button.
It seems ther are 2 options for meaningful implementation positions:

a) on the view of the outbox maybe left side of the "delete" button
b) when open the sent message on the top, somewhere next to the subjet text

Thanks a lot

mmd-osm:
You didn’t say why you can’t just open up the old message in your outbox, select and copy relevant parts of your text and then paste that text into a new message. Seems like the first thing I would try to do.

sun-geo:
@mmd-osm 
Thanks for the response. That's true (_"...didn’t say ..."_). I do this exactly the way how you describe it above.
Additionally I manually add the "> " line prefixes syntax, so that the old text has a small offset on the left side. Furthermore I manually add the date when I did send the old/initial message. I using the approximately same style how the system is doing this automatically when you reply an answer message, like this:
_"On 2020-03-31 10:37:25 UTC USERNAME wrote:"_

These additionally manual steps are the explanation, why I would like to have this automatic forward function/button also for the outbox to not spend time for these manually steps.

mmd-osm:
Thanks. I asked that question, because you've already stated a solution "i would like to have a forward (function)/button" in your post, without fully describing the "as-is" situation. Maybe we will come up with an entirely different solution, which is only possible, if we understand your requirements.

sun-geo:
Is okay, i'm happy you asked for more details, so that a potential implementation would/will better suit. By the way, if you have any other ideas for a maybe different solution (as written in (a) and/or (b)) for changing the osm.org code to get the similar result (reducing or even removing the mentioned manually steps) i'm happy to hear. I'm also open to change the subject title of this GitHub issue, if desired, any suggestions?

PR #5120, opened 2024-08-26
Allows sending follow-ups for sent messages

### LINKED ISSUE #3497: Add a reply button to a send email
### Description

UseCase: I send an email and want to send a follow up as a reply to that email. In other workflows, I would open the previous mail, click reply to follow up on the email.

However, in OSM message such a button is missing. I need to write an independent message, loosing the context (or I have to manually re-create the context).

Possible solution: Adopt the reply button from incoming mails to outgoing mails and pre-fill a message to the recipient with the previous mail as comment (just with incoming mail-replies).

This is the page I am referring to.
Route example: at https://www.openstreetmap.org/messages/1016723
image

### Screenshots

_No response_

nenad-vujicic:
PR adds Reply button when displaying sent messages, so users can send follow-ups
 
Fixes #3497. Adds Reply button when displaying sent messages (app/views/messages/show.html.erb) and allows replying to messages for which sender is current user (app/controllers/messages_controller.rb).

image

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -89,6 +89,16 @@ class MessagesController < ApplicationController
       @title = @message.title
 
       render :action => "new"
+    elsif message.sender == current_user
+      @message = Message.new(
+        :recipient => message.recipient,
+        :title => "Re: #{message.title.sub(/^Re:\s*/, '')}",
+        :body => "On #{message.sent_on} #{message.sender.display_name} wrote:\n\n#{message.body.gsub(/^/, '> ')}"
+      )
+
+      @title = @message.title
+
+      render :action => "new"
     else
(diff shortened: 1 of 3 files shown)


nertc:
Functionality works perfectly. I think, it also fixes https://github.com/openstreetmap/openstreetmap-website/issues/2644. Despite of the fact, that it implements a different solution (reply instead of forward), there won't be a problem stated in the issue. 
Also, shouldn't there be a test for checking redirection and rendering of the reply button in outbox items?

tomhughes (changes requested):
This looks good to me if we could just add some tests as @nertc suggests?

nenad-vujicic:
Improved tests, thanks!

tomhughes (approved):
That looks good now, thanks.

PR #5121, opened 2024-08-27
Provide a truncated view of recent diaries

nertc:
This PR is alternative to the https://github.com/openstreetmap/openstreetmap-website/pull/5103.

This PR addresses "Provide a summary view of recent diaries" issue mentioned in the https://github.com/openstreetmap/openstreetmap-website/issues/3887

Diary entries will be truncated after 1000 characters on the list pages. To avoid long diaries with pictures on the list page, every picture is handled as 500 characters. View of the diary entry details (show) page remains the same. "See full entry" button was added to the entry to explicitly mention that full entry is no more visible in the list and enhance UX. Fixes https://github.com/openstreetmap/openstreetmap-website/issues/3887

Screenshots:

image
image
Screenshot 2024-08-27 113359
Screenshot 2024-08-27 113411

### DIFF
--- a/app/views/diary_entries/_diary_entry.html.erb
+++ b/app/views/diary_entries/_diary_entry.html.erb
@@ -2,7 +2,15 @@
   <%= render :partial => "diary_entry_heading", :object => diary_entry, :as => "diary_entry" %>
 
   <div class="richtext text-break" xml:lang="<%= diary_entry.language_code %>" lang="<%= diary_entry.language_code %>">
-    <%= diary_entry.body.to_html %>
+    <% if truncated %>
+      <% truncated_entry = diary_entry.body.truncate_html(2000) %>
+      <%= truncated_entry[:html] %>
+      <% if truncated_entry[:truncated] %>
+        <p>&hellip; <%= link_to t(".full_entry"), diary_entry_path(diary_entry.user, diary_entry) %></p>
+      <% end %>
+    <% else %>
+       <%= diary_entry.body.to_html %>
+    <% end %>
   </div>
 
   <% if diary_entry.latitude and diary_entry.longitude %>
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -63,6 +63,48 @@ module RichText
       nil
     end
 
+    def truncate_html(max_length = nil, img_length = 1000)
+      html_doc = to_html
+      return html_doc if max_length.nil?
+
+      doc = Nokogiri::HTML::DocumentFragment.parse(html_doc)
+      keep_or_discards = %w[p h1 h2 h3 h4 h5 h6 pre a table ul ol dl]
+      accumulated_length = 0
+      exceeded_node_parent = nil
+      truncated = false
+
+      doc.traverse do |node|
+        if accumulated_length >= max_length
+          if node == exceeded_node_parent
+            exceeded_node_parent = node.parent
+            node.remove if keep_or_discards.include?(node.name)
+          else
+            node.remove
+          end
+          next
+        end
+
+        next unless node.children.empty?
+
+        if node.text?
+          accumulated_length += node.text.length
+        elsif node.name == "img"
+          accumulated_length += img_length
+        end
+
+        if accumulated_length >= max_length
+          truncated = true
+          exceeded_node_parent = node.parent
+          node.remove
+        end
+      end
+
+      {
+        :truncated => truncated,
+        :html => doc.to_html.html_safe
+      }
+    end
+
     protected
 
     def simple_format(text)
--- a/app/views/diary_entries/_page.html.erb
+++ b/app/views/diary_entries/_page.html.erb
@@ -1,7 +1,7 @@
 <turbo-frame id="pagination" target="_top" data-turbo="false">
   <h4><%= t ".recent_entries" %></h4>
 
-  <%= render @entries %>
+  <%= render @entries, :truncated => true %>
 
   <%= render "shared/pagination",
              :newer_id => @newer_entries_id,
--- a/app/views/diary_entries/show.html.erb
+++ b/app/views/diary_entries/show.html.erb
@@ -14,7 +14,7 @@
   </div>
 <% end %>
 
-<%= render @entry %>
+<%= render @entry, :truncated => false %>
 <%= social_share_buttons(:title => @entry.title, :url => diary_entry_url(@entry.user, @entry)) %>
 
 <div id="comments" class="comments mb-3 overflow-hidden">
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -608,6 +608,7 @@ en:
     diary_entry:
       posted_by_html: "Posted by %{link_user} on %{created} in %{language_link}."
       updated_at_html: "Last updated on %{updated}."
+      full_entry: See full entry
       comment_link: Comment on this entry
       reply_link: Send a message to the author
       comment_count:
--- a/test/system/diary_entry_test.rb
+++ b/test/system/diary_entry_test.rb
@@ -74,4 +74,54 @@ class DiaryEntrySystemTest < ApplicationSystemTestCase
     assert_link "Diary Entries in Portuguese", :href => "/diary/pt"
     assert_no_link "Diary Entries in Russian"
   end
+
+  test "should not be hidden on the list page" do
+    body = SecureRandom.alphanumeric(1998)
+    create(:diary_entry, :body => body)
+
+    visit diary_entries_path
+
+    assert_content body
+    assert_no_content I18n.t("diary_entries.diary_entry.full_entry")
+  end
+
+  test "should be hidden on the list page" do
+    body = SecureRandom.alphanumeric(2000)
+    create(:diary_entry, :body => body)
+
+    visit diary_entries_path
+
+    assert_no_content body
+    assert_content I18n.t("diary_entries.diary_entry.full_entry")
+  end
+
+  test "should be partially hidden on the list page" do
+    text1 = "a" * 500
+    text2 = "b" * 500
+    text3 = "c" * 999
+    text4 = "dd"
+    text5 = "ff"
+
+    body = "<p>#{text1}</p><div><p>#{text2}</p><p>#{text3}<a href='#'>#{text4}</a></p></div><p>#{text5}</p>"
+    create(:diary_entry, :body => body)
+
+    visit diary_entries_path
+
+    assert_content text1
+    assert_content text2
+    assert_no_content text3
+    assert_no_content text4
+    assert_no_content text5
+    assert_content I18n.t("diary_entries.diary_entry.full_entry")
+  end
+
+  test "should not be hidden on the show page" do
+    body = SecureRandom.alphanumeric(2001)
+    diary_entry = create(:diary_entry, :body => body)
+
+    visit diary_entry_path(diary_entry.user, diary_entry)
+
+    assert_content body
+    assert_no_content I18n.t("diary_entries.diary_entry.full_entry")
+  end
 end


kcne:
I agree with @tomhughes' comment on #5103 that this approach is more standard for addressing #3887. Since we're using Markdown, we should account for edge cases where the entry starts with images, headings, code blocks, or tables, which may look unreadable when truncated. Instead of using max-height, alternative solution would be to extract the first paragraph or a fixed number of characters and append ellipses (...). This would ensure the preview remains readable and coherent without cutting off important elements.

nertc:
@kcne Thank you for the review. I thought about truncating with a fixed number of characters, but it adds complexity in the code (it needs JS to calculate and trim number of characters) and introduces complexities about consistent height of the shown post (for example, if there is a post that starts with 5 big picture, if all of the pictures are shown (because of the small character number), shown post will be much bigger, than other posts' bodies, which include only characters and words). Same goes for showing the first paragraph, which in some cases will be huge and, in other cases, tiny. Truncation with a number of characters also needs additional custom parsing to avoid cut in the middle of an image link or any character sequence that is written for formatting. Of course, all of these are possible to make, but, I think, complexity they introduce is not worth, as there will be no such big quality of difference in visuals.

AntonKhorev:
One problem with this version is that the entire diary entry text is still present on the page. You can search with Ctrl-F and get results that are invisible.

nertc:
@AntonKhorev It's a good point, I'll try to fix it and, as we will need more complex solution than just adding `max-height`, I'll also take @kcne's recommendations into account.

AntonKhorev:
When I thought how to do truncated diary entries I first considered doing it server-side. I already go through markdown parse trees in the richtext library to get images and descriptions for Open Graph tags. I can probably extend that to generate truncated entries.

Then I thought that ultimately I want to limit entry heights. Isn't it easier to ask the browser what height each element is instead of trying to estimate it server-side? And if there's an image, it usually doesn't have its height specified in the source markdown code, therefore I can't know it server-side but client-side I can.

But it's not that simple. Images can load any time, including never. I'd have to listen to load events to recalculate the heights. Now it doesn't look like a good idea. Let's suppose there's a large image and some text below. Before the image is loaded that text fits in the truncated view. Am I going to remove the text after the image loads, possibly when the user is reading it? I guess it's better to estimate heights and truncate server-side.

nertc:
I almost finished the code. I think it will be ready tomorrow. The idea is to truncate entries server-side and, for example, if we show only 1000 characters, estimate that image will take space of 500, therefore, it will truncate text and also won't allow to add many photos that will overload diary entries list page.

nertc:
This PR was updated. Now truncation is handled in the back end. After 1000 characters, entry is truncated and `...` are added to it. Every image is handled as 500 characters to avoid entries with a big number of images.
@kcne @AntonKhorev Thank you for reviews and recommendations

app/views/diary_entries/_diary_entry.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= render :partial => "diary_entry_heading", :object => diary_entry, :as => "diary_entry" %>
 
   <div class="richtext text-break" xml:lang="<%= diary_entry.language_code %>" lang="<%= diary_entry.language_code %>">
-    <%= diary_entry.body.to_html %>
+    <%= truncated ? diary_entry.truncated_body(1000).to_html : diary_entry.body.to_html %>
```
AntonKhorev, line 5: First markdown from the RichText object is converted to html. Then html is processed by `truncated_body` and a new RichText object is constructed. Then html is again extracted from that object. And if the html is nontrivial it may get mangled in the process, turning, for example, this

```
There's a table below:

| thing | 42 | an important thing |
| thing | 43 | an important thing |
```

image

into this

image

app/models/concerns/truncatable.rb
```diff
@@ -0,0 +1,30 @@
...
+      if accumulated_length >= max_length
+        node.remove unless truncated_node.ancestors.include?(node)
+        next
+      end
+
+      next unless node.children.empty?
+
+      content_length = node.text? ? node.text.length : empty_tag_length
+      if accumulated_length + content_length >= max_length
+        node.content = node.text.truncate(max_length - accumulated_length) if node.text?
```
AntonKhorev, line 21: Do we really need to truncate node contents rather than keeping/discarding it as a whole?

nertc: If we discard it as a whole, diary entry of someone who created only one huge paragraph, won't be shown at all.
If we keep it fully, diary entry of someone who created only one huge paragraph, will overload diary entries list.
That's why I think truncation is the best method in this case.

AntonKhorev: If you write one huge paragraph, yes it's going to be discarded. Don't write one huge paragraph.

The limit of 1000 characters could be increased though.

nertc: What do you think is a good number for the limit?

nertc:
PR was updated. Paragraph truncation was removed. Max length of the diary entry in the list was changed to 2000. Images now take size of 1000 character. Additional parsing to the RichText was removed and now HTML is directly sanitized and marked as safe.

app/views/diary_entries/_diary_entry.html.erb
nertc: Additional parsing was removed and now HTML is directly sanitized and converted using `RichText::SimpleFormat.new.sanitize`

app/models/concerns/truncatable.rb
nertc: Truncation was removed and character limit was increased to 2000. Thanks for reviewing.

tomhughes (commented):
I wonder about the overall approach of doing this in a concern - would it be better to do it in `RichText::Base` perhaps?

What I envisage is a new method there to do truncation and then altering the `to_html` methods in the derived classes to take an optional truncation length (defaulting to `nil`) and apply the truncation method to their result if it's set - then instead of a boolean parameter the views can pass in a truncation length.

Just a thought, but it seems with the current implementation things get messy when trying to deal with the order things are being done and what is and isn't sanitised etc.

app/models/concerns/truncatable.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      accumulated_length += node.text? ? node.text.length : empty_tag_length
+      if accumulated_length < max_length
+        last_child = node
+      else
+        node.remove
+      end
+    end
+
+    RichText::SimpleFormat.new.sanitize(doc)
```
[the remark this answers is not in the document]
nertc: Sorry for the late answer. As HTML was truncated and, therefore, changed, Ruby required additional sanitization to ensure that everything was still okay. I'll move truncation method in the `RichText::Base`. I think, it will solve most of problems and additional complexities introduced by adding this method to the concern.

nertc:
PR was updated. Truncation logic was moved to `RichText::Base` and now `.to_html` can take truncation length as an argument (if not provided, full content will be displayed).

app/models/concerns/truncatable.rb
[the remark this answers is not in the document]
nertc: Thank you for the review. PR was updated. From now, sanitization doesn't need to be run more than one time.

lib/rich_text.rb
```diff
@@ -65,6 +65,32 @@ def description
...
+
+      doc.traverse do |node|
+        if accumulated_length >= max_length
+          node.remove unless !last_child.nil? && last_child.ancestors.include?(node)
+          next
+        end
+
+        next unless node.children.empty?
+
+        accumulated_length += node.text? ? node.text.length : empty_tag_length
```
AntonKhorev, line 83: > Images now take size of 1000 character.

Any empty tag costs 1000 characters. For example, empty paragraph costs 1000 characters. Something like this becomes empty:

```
<p></p>
<p></p>

hello?
```

Some tags are naturally empty, for example `<br>`. In this example "3" is truncated:

```
1  
2  
3
```
(note spaces in markdown after "1" and "2")

lib/rich_text.rb
```diff
@@ -65,6 +65,32 @@ def description
...
     protected
 
+    def truncate_html(html_doc, max_length = nil, empty_tag_length = 1000)
+      return html_doc if max_length.nil?
+
+      doc = Nokogiri::HTML::DocumentFragment.parse(html_doc)
+      accumulated_length = 0
+      last_child = nil
+
+      doc.traverse do |node|
```
AntonKhorev, line 75: My idea was that there are some types of elements that you either completely keep or completely discard. A paragraph would be one of them. Here instead a paragraph may get truncated after after any of elements inside it. Let's say you have a link, any if that link happens to go over the truncation threshold, the rest of the paragraph gets thrown out. That can make it unreadable.

This is even more true for tables. Truncation as implemented here can stop in the middle of some cell. It won't try to complete a row, although completing a row wouldn't work in general. Tables probably also need to be either entirely in or entirely out. Something like `<div>` or `<section>`, that can be included partially.

lib/rich_text.rb
nertc: Maybe changing checking if the tag is empty to checking if the tag is `` will be better. Therefore, tags like `<br>` or `<p></p>` will take 1 symbol space, while `` will take for example 1000 symbol space. If we want any other tags to be treated in a specific way, feel free to suggest.

lib/rich_text.rb
nertc: >Let's say you have a link, any if that link happens to go over the truncation threshold, the rest of the paragraph gets thrown out. That can make it unreadable.

What do you suggest in this situation instead of removing whole paragraph, should it be done like it was before (like this `This is long paragraph` -> `This is lon...`)?

lib/rich_text.rb
nertc: Changed additional length to be added only when the tag name is ``.

lib/rich_text.rb
nertc: Added `keep_or_discards` array which contains tag names which will be either completely kept or completely discarded.

nertc:
@AntonKhorev Thank you for the review. PR was updated accordingly.

tomhughes:
I think this needs some sort of indication when an entry has been truncated - an ellipsis at the end of the truncated text or something.

lib/rich_text.rb
```diff
@@ -65,6 +65,42 @@ def description
 
     protected
 
+    def truncate_html(html_doc, max_length = nil, img_length = 1000)
+      return html_doc if max_length.nil?
+
+      doc = Nokogiri::HTML::DocumentFragment.parse(html_doc)
+      keep_or_discards = %w[p h1 h2 h3 h4 h5 h6 pre a table ul ol dl]
+      accumulated_length = 0
+      parent = nil
```
AntonKhorev, line 74: If you had to describe the possible values of `parent`, what would you say? It's not a parent of every traversed node.

AntonKhorev:
> I think this needs some sort of indication when an entry has been truncated - an ellipsis at the end of the truncated text or something.

Do we want another method instead of building truncation into `to_html`? We need it to return the html and whether it was truncated, so we can display "See full entry" only if truncation happened.

AntonKhorev:
If the truncation was the last step, I'd move to another module/class, which I'd use only when I need to truncate. But currently html conversion is followed by sanitize and linkify (why not first linkify then sanitize?). Can we truncate after sanitize/linkify?

lib/rich_text.rb
nertc: It is the parent of the node which exceeds the limit length. Do you suggest changing the variable name?

nertc:
> If the truncation was the last step, I'd move to another module/class, which I'd use only when I need to truncate. But currently html conversion is followed by sanitize and linkify (why not first linkify then sanitize?). Can we truncate after sanitize/linkify?

Unfortunately, from what I've tested, we can't truncate after sanitize/linkify, because truncation changes HTML code and Ruby on Rails doesn't render it if it is not sanitized/linkified once more.

lib/rich_text.rb
nertc: Name was changed to `exceeded_node_parent`.

nertc:
PR was updated. `master` was merged and merge conflicts were fixed.

PR was changed according to reviews. @tomhughes to indicate if an entry has been truncated, as @AntonKhorev suggested, `See full entry` will not be shown on entries, which are not truncated.

AntonKhorev:
> @AntonKhorev Unfortunately, from what I've tested, we can't truncate after sanitize/linkify, because truncation changes HTML code and Ruby on Rails doesn't render it if it is not sanitized/linkified once more.

It's rendered if you add `html_safe`, like `sanitize` does:

```ruby
    def to_truncated_html(truncation_length = nil)
      truncate_html(to_html, truncation_length).to_s.html_safe
    end
```

tomhughes:
> ... to indicate if an entry has been truncated

What should this look like because I don't seem to be seeing this? I also can't see anything in the code that looks like it would do it?

AntonKhorev:
> > ... to indicate if an entry has been truncated
> 
> What should this look like because I don't seem to be seeing this? I also can't see anything in the code that looks like it would do it?

The link to the full entry is displayed alongside the rest like *add a comment* separated by |. I thought if it's noticeable enough and I guess this comment means that it isn't and should be moved above.

The check that detects the truncation is somewhere in the template, which is also dubious.

nertc:
PR was updated.
@AntonKhorev `truncate_html` was moved to be done after `linkify` and `sanitize`.
@tomhughes visual of `See full entry` link was changed. Now it is separate from other links and easier to see (see screenshots).
Logic of checking if HTML was truncated, was changed and now it is saved in the object instead of doing additional truncation.

Screenshots:
image
image

app/views/diary_entries/_diary_entry.html.erb
```diff
@@ -2,7 +2,10 @@
   <%= render :partial => "diary_entry_heading", :object => diary_entry, :as => "diary_entry" %>
 
   <div class="richtext text-break" xml:lang="<%= diary_entry.language_code %>" lang="<%= diary_entry.language_code %>">
-    <%= diary_entry.body.to_html %>
+    <%= truncated ? diary_entry.body.to_html(2000) : diary_entry.body.to_html %>
+    <% if truncated && diary_entry.body.was_truncated?(2000) %>
```
tomhughes, line 6: This whole business of having the rich text object cache which lengths it has done truncations on is a pretty horrible interface.

Really `to_html` should indicate directly if it's done a truncation though it would probably be quite painful - it seems like one option would be to have it return a pair of values (the HTML and a truncation flag) and the other would be for it to return a single object with `html` and `truncated?` methods.

nertc: `truncate_html` method was modified to be public and return object containing `html` and `truncated` information. To avoid calling method twice and caching (e.g. Memoization), truncated object is saved as `truncated_entry` variable in the `.html.erb` file.

nertc:
Thanks for the review @tomhughes. PR was updated according to the suggestion.

tomhughes:
Thanks - that looks like a much nicer solution than what I was imagining! I think this is good to merge now.

PR #5124, opened 2024-08-27
Use turbo to update issue search results

tomhughes:
This extends #5057 to use turbo when changing search parameters on the issue list.

The user not found message when searching for an unknown user has been moved from the flash message into the results pane to ensure it can be updated when a search is done.


gravitystorm:
LGTM, thanks!

Issue #5127, opened 2024-08-28
Test failures on issues pagination


gravitystorm:
We're getting test failures related to the issue pagination, e.g.

```
Failure:
IssuesTest#test_issues_pagination [test/system/issues_test.rb:184]:
expected to find visible css "tr" 31 times, found 51 matches: 
"Status Reports Reported Item Reported User Last Updated",
"Open 0 Reports User 83 User 83 less than 1 minute ago", 
"Open 0 Reports User 82 User 82 less than 1 minute ago", 
"Open 0 Reports User 81 User 81 less than 1 minute ago", 
"Open 0 Reports User 80 User 80 less than 1 minute ago", 
...
```

Here's an example from [a recent unrelated PR](https://github.com/openstreetmap/openstreetmap-website/actions/runs/10601741224/job/29382202917)

Refs #5057 cc @nertc

tomhughes:
Yes this is basically the timing issue that we discussed on that PR I think, that sometimes turbo doesn't complete changing the page in 2s and the assert fails.

nertc:
Yes, I agree with @tomhughes. If it slows the process down, we can either delete tests temporarily, add custom `:wait`-s or change default waiting time.

gravitystorm:
Capybara has "intelligent waiting" by default, where it doesn't fail assertions immediately after a click, it retries until it detects the content on the page and then runs the assertions. So I wonder if it's not that the next page is taking 2+ seconds to render (and that would be a problem in itself) but rather that the turbo has just started to run, but capybara thinks the content is already present and so immediately tests it, before turbo is finished (which might be only a few ms later).

So it might not be related to timeouts (and there's no timeout information in the error) but rather that capybara's running too quickly. In which case extending the (maximum) wait time might not help.

It's perhaps notable that this particular assertion isn't checking that a specific thing will appear (e.g. specific text, or html with a specific id, from the second page of results) that capybara would normally intelligently wait for. Instead it's checking for html that's also on the original page (just in a different count). Could that be interfering with the intelligent waiting?

https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends

tomhughes:
Turbo adds the `aria-busy` attribute (https://turbo.hotwired.dev/reference/attributes#automatically-added-attributes) to the frame while a navigation is in progress so maybe we can wait until that is removed before re-evaluating?

nertc:
[...]

PR #5129, opened 2024-08-29
Test failures on issues pagination

nertc:
This PR addresses "Test failures on issues pagination" issue mentioned in the https://github.com/openstreetmap/openstreetmap-website/issues/5127

Test was modified. Now instead of checking the number of `tr`, test tries to search for the specific issues. According to the [Capybara Documentation](https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends), this approach should fix occasional failing of issues page tests.

### DIFF
--- a/test/system/issues_test.rb
+++ b/test/system/issues_test.rb
@@ -163,8 +163,8 @@ class IssuesTest < ApplicationSystemTestCase
   end
 
   def test_issues_pagination
-    1.upto(80).each do
-      user = create(:user)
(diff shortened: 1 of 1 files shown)


AntonKhorev:
#5127 didn't happen for me, so I don't know if this PR fixes it. But I merged it because I'm changing this test in my other PR.

Also I decreased the number of issues tested from 80 to 8 and page size from 50 to 5 using `:limit`. Capybara worked much slower for me with 80 issues.

Issue #3532, opened 2022-04-14
Update 404 page


berrely:
### The problem
The current 404 page that users get when typing in an invalid URL is uninformative, clunky and ugly.
* Instead of hierarchically displaying the most important information, it opens the prompt with “File not found”, only mentioning the specific error at the end of the second line.
* It has a lot of redundancy (“OpenStreetMap server”), and doesn't give good advice on what to do.
* The second line says that if you have found a broken link or bug, you should ask the OSM community, even though the correct forums for bugs would be repositories like this one.
* It also looks incredibly ugly, with loads of whitespace and the default browser font.
### The solution
* The 404 page should be redesigned(!)
* * The text should be rewritten to be more helpful
* * The header should be more clear on the specific error
* * Links should be provided to get back to the homepage (currently hidden behind the giant logo)
* * The bug text should be removed or refactored to be more appropriate

See also a possible mockup attached.

### Screenshots

**Current 404 page:**
![](https://i.imgur.com/Bk5b1lB.png)
**Mockup of possible 404 page:**
![image](https://user-images.githubusercontent.com/20139168/163435807-5a2602e0-1adb-4fd2-946d-90980c3ad49d.png)

tomhughes:
So the most important thing, to be rendered in the largest font, is the number 404 which will mean nothing to most non-technical users?

Whatever you may think of the old version, it puts technical details like that where they belong, in parentheses after the human readable explanation.

This is (in most cases) not the correct place to contact anyway - the vast majority of 404 errors will probably not be because of any error in the software, they will simply be stale search results for things which have been deleted.

Can I ask what led you to this? Did you encounter a 404 somewhere?

mmd-osm:
(previous discussion on the topic in #1550)

berrely:
> Can I ask what led you to this? Did you encounter a 404 somewhere?

It was an old about page, though I don't have the link to hand.

> This is (in most cases) not the correct place to contact anyway - the vast majority of 404 errors will probably not be because of any error in the software, they will simply be stale search results for things which have been deleted.

I agree, I don't think the 404 page should suggest reporting a bug in any case.

> Whatever you may think of the old version, it puts technical details like that where they belong, in parentheses after the human readable explanation.

At this point the number “404” has become synonymous with a file not existing on the target server, so it effectively acts as a human-readable explanation, though I agree for non-technical users it may not make sense (it is just a mockup, the 404 could stay at the end of the text).

[...]

PR #5130, opened 2024-08-29
Refactor error layout with small styling enchancments

kcne:
This PR addresses parts of [#3532](https://github.com/openstreetmap/openstreetmap-website/issues/3532) by making the error layout more consistent with the rest of the site's UI.

**Changes:**

- Encapsulated meta tags into the `layouts/head` partial for better organization.
- Applied Bootstrap's `container` class to the `<body>` for a consistent layout.
- Added a `main` element with `mt-4` to improve spacing.
- Introduced a "Home" button with `btn btn-primary` to facilitate easier navigation.

These updates enhance the overall design and user experience of the error pages.

#### Before
Screenshot 2024-08-29 at 20 34 27

#### After
Screenshot 2024-08-29 at 20 35 04
Screenshot 2024-08-29 at 20 35 19

### DIFF
--- a/app/views/layouts/error.html.erb
+++ b/app/views/layouts/error.html.erb
@@ -4,14 +4,19 @@
     <meta charset="utf-8">
     <title>OpenStreetMap</title>
     <%= stylesheet_link_tag "errors", :media => "screen" %>
-    <%= render :partial => "layouts/meta" %>
+    <%= render :partial => "layouts/head" %>
   </head>
-  <body>
-    <a href="<%= root_path %>">
-      <%= image_tag "osm_logo.svg", :alt => t("layouts.logo.alt_text"), :class => "logo" %>
-    </a>
-    <div class="details">
-      <%= yield %>
-    </div>
+   <body class="container">
+    <main class="mt-4">
+      <a href="<%= root_path %>">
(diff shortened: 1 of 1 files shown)


app/views/layouts/error.html.erb
```diff
@@ -4,14 +4,19 @@
     <meta charset="utf-8">
     <title>OpenStreetMap</title>
     <%= stylesheet_link_tag "errors", :media => "screen" %>
-    <%= render :partial => "layouts/meta" %>
+    <%= render :partial => "layouts/head" %>
```
AntonKhorev, line 7: `<head>` inside `<head>`?

You're proposing to load everything on 404 pages: jquery, leaflet, bootstrap, turbo etc. And you're doing this to set max width and add one link styled as a button, right?

kcne: Well since you opened a new PR that "does this right" I will close this anyway.

AntonKhorev: Looking [here](https://github.com/openstreetmap/openstreetmap-website/pull/1550#issuecomment-304481362) I can guess that the page was supposed to work when the asset pipeline is broken, although currently it includes a logo asset.

Issue #3876, opened 2023-01-06
Render errors rather than redirecting?


gravitystorm:
We currently have a few "error" pages, including from the errors_controller but also other pages like site_controller/offline. To show these pages, users typically go through this flow:

* Attempt to view page A
* Something in A triggers an error
* User receives a redirect to the error page B
* User views the error on page B

It's not clear to me what advantage, if any, there is being redirected. The alternative would be that when an error (or similar) is reached in the controller, it renders the error page from a shared template. That avoids having any redirects, and it avoids changing the URL. For example, if you find a page that triggers a 500 internal server error, your can copy the link into the bug report more easily :smile:

I haven't come across any other sites that follow this redirect-to-error-page approach. I'm curious as to whether this is deliberate or whether it's just happened this way over for historical reasons.

PR #5131, opened 2024-08-30
Update error pages

AntonKhorev:
You can debate what changes are needed on error pages.  #5130 does some, referring to #3532, but that issue proposes some things I wouldn't agree with.

For comparison here's what I'd change.

image
image
image

### DIFF
--- a/app/views/layouts/_meta.html.erb
+++ b/app/views/layouts/_meta.html.erb
@@ -1,3 +1,5 @@
+<meta http-equiv="X-UA-Compatible" content="IE=edge" />
+<meta name="viewport" content="width=device-width, initial-scale=1">
 <% [57, 60, 72, 76, 114, 120, 144, 152, 180].each do |size| -%>
 <%= favicon_link_tag "apple-touch-icon-#{size}x#{size}.png", :rel => "apple-touch-icon", :sizes => "#{size}x#{size}", :type => "image/png" %>
 <% end -%>
--- a/app/assets/stylesheets/errors.scss
+++ b/app/assets/stylesheets/errors.scss
@@ -1,3 +1,7 @@
+body {
+  font-family: system-ui;
+}
+
 .logo {
   float: left;
   margin: 10px;
--- a/app/assets/stylesheets/errors.scss
+++ b/app/assets/stylesheets/errors.scss
@@ -1,18 +1,43 @@
 body {
+  margin: 1rem;
+  margin-top: 2rem;
   font-family: system-ui;
 }
 
-.logo {
-  float: left;
-  margin: 10px;
+main {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 1rem 2rem;
+  max-width: 960px;
 
-  img {
-    display: block;
-    max-width: 100%;
-    height: auto;
+  .logo {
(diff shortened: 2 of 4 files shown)


kcne:
Maybe also add a back to homepage link since it's not really visually clear and intuitive that clicking on OSM logo navigates to homepage.

AntonKhorev:
I looked at previous discussions and I'm a little surprised about this "back to homepage link". Why is is necessarily "back"? If I came to osm through a broken link and landed on the 404 page, I'm not going "back" to any homepage.

kcne:
I appreciate your point about the term "back" possibly implying a prior visit to OSM, which might not always be the case.

However, the core issue here is user experience and intuitiveness. While a user might not be "going back" in all scenarios, the primary intent of a "back to homepage" link is to provide a clear and intuitive way to navigate to a familiar starting point—in this case OSM.

Clicking on the logo to return to the homepage is a common web convention, but it’s not universally understood by all users. Therefore, having a clearly labeled link offers an additional, explicit option for a broader range of users, including those who may not immediately recognize the logo as a navigational element.

It doesn't have to be named "back to homepage," but I still believe having a link that clearly navigates to openstreetmap.org is a good way to give users the option to continue to OSM, even if they landed on an error page.

AntonKhorev:
> having a clearly labeled link offers an additional, explicit option for a broader range of users

There's a different opinion in https://github.com/openstreetmap/openstreetmap-website/pull/1550#issuecomment-525766131, but it doesn't really matter for this PR because I'm not changing the text of the page. If you think an extra homepage link is necessary on the 404 page or on every error page, you can add it without doing anything I do here, unless you want this link styled as a Bootstrap button.

nenad-vujicic:
Honestly, adding URL for which an error occurred (unless #3876 is resolved) or Page URL + actions (user actions, not stack / call trace) after which error occurred sound much better option than home / root URL or leaving nothing? But, I agree this will take much more time and is not part of #5130  #5131 . I apologize if this is too off-topic.

nenad-vujicic:
> Honestly, adding URL for which an error occurred (unless #3876 is resolved) or Page URL + actions (user actions, not stack / call trace) after which error occurred sound much better option than home / root URL or leaving nothing? But, I agree this will take much more time and is not part of #5130 #5131 . I apologize if this is too off-topic.

Also, perhaps we should move toward reporting errors through flash[] mechanism (i.e. using bootstrap alerts) and leave these super-simple error pages for fatal errors only? Again, I apologize for going off-topic, I would just like to read your thoughts.

app/views/layouts/_meta.html.erb
```diff
@@ -1,3 +1,5 @@
+<meta http-equiv="X-UA-Compatible" content="IE=edge" />
```
nertc, line 1: I know that these two lines were copied from the `_head.html.erb`, but for the consistency of this file, I think, it will be better if we use `<%= tag.meta ... %>` instead.

tomhughes:
Most of the discussion here seems to be about whether this goes far enough but I don't see any significant complaints about the things this does do, and it looks good to me so I'm going to merge it.

If people want to propose additional changes in other PRs then please feel free to do so.

PR #5133, opened 2024-08-30
Test oauth2 application list with moderator permission icon

AntonKhorev:
Will fail if #5058 is applied because https://github.com/openstreetmap/openstreetmap-website/pull/5058#pullrequestreview-2272966401.

### DIFF
--- a/test/controllers/oauth2_applications_controller_test.rb
+++ b/test/controllers/oauth2_applications_controller_test.rb
@@ -53,6 +53,16 @@ class Oauth2ApplicationsControllerTest < ActionDispatch::IntegrationTest
     assert_select "tbody tr", 2
   end
 
+  def test_index_with_moderator_app
+    user = create(:user)
+    create(:oauth_application, :owner => user, :scopes => "write_redactions")
+
+    session_for(user)
+
+    get oauth_applications_path
+    assert_response :success
+  end
+
   def test_new
     user = create(:user)
 


tomhughes:
Looks good to me, thanks.

AntonKhorev:
> bundle exec i18n-tasks missing

Most of this list is not missing translations.

mmd-osm:
Yes, that’s what I meant the list is a bit noisy.

AntonKhorev:
> a bit noisy

It's almost entirely noisy. Aside from `oauth.for_roles.moderator` only `traces.update.title` is legitimate.

> Everything with a dot or underscore at the end is probably part of some string concatenation. These would have to be excluded from the analysis.

`geocoder.search_osm_nominatim.admin_levels.level` doesn't end with a dot or underscore.

mmd-osm:
I don’t think 8 false positives out of several hundred translation usages (I haven’t counted the exact number) is too bad when running a tool for the first time. These could be easily addressed by adding some annotations.

mmd-osm:
Adding this config to config/i18n-tasks.yml will silence the existing Javascript issue:

```
ignore_missing:
  - 'geocoder.search_osm_nominatim.admin_levels.level'
  - 'javascripts.directions.engines.:'
  - 'javascripts.directions.instructions.exit_counts.:'
  - 'javascripts.map.base.gps'
  - 'javascripts.map.layers.:'
  - 'javascripts.map.locate.:'
  - 'javascripts.query.:'
  - 'javascripts.site.map_'
```

i18n-tasks has a related check to find out unused translations.  i18n-tasks.yml already has some osm website repo specific entries underneath `ignore_unused:`. It seems that list is slightly out of date.

```
bundle exec i18n-tasks unused
```

AntonKhorev:
> Adding this config to config/i18n-tasks.yml will silence the existing Javascript issues

Then you'll have unused entries in the ignore list if you don't maintain it too.

`javascripts.map.base.gps` - this is not just *missing* but rather *missing because unused and doesn't make sense*. "gps" is not a base layer.

PR #5135, opened 2024-08-31
Fix page title for failed trace updates

AntonKhorev:
Missing locale string, see https://github.com/openstreetmap/openstreetmap-website/pull/5133#issuecomment-2322940489


tomhughes:
Looks good to me, thanks.

PR #5136, opened 2024-08-31
Remove unused reference to missing gps layer name

AntonKhorev:
`name: I18n.t("javascripts.map.base.gps")` was added in https://github.com/openstreetmap/openstreetmap-website/commit/df17b997b33fc9d0c60360de7f27aadc76b930a0 without a translation provided. Went unnoticed because the name is not actually used anywhere, see https://github.com/openstreetmap/openstreetmap-website/pull/5133#issuecomment-2323053456.


tomhughes:
Looks good to me, thanks.

PR #5138, opened 2024-08-31
Remove unused fail locale strings from user role grant/revoke

AntonKhorev:
They were replaced by more specific messages in https://github.com/openstreetmap/openstreetmap-website/commit/1e3b3c1f10213c48a40083e8bbca1429d9f819d6.


tomhughes:
Looks good to me, thanks.

PR #5142, opened 2024-09-01
Drop unused oauth_10a_disabled translation

tomhughes:
This got left in when dropping OAuth 1 support...


AntonKhorev:
Merged, thanks

PR #5145, opened 2024-09-01
I18N: messages: avoid dynamic translation keys

mmd-osm:
By moving the translation outside the loop and calling t(...) with explicitly stated keys, `bundle exec i18n-tasks unused` no longer reports the keys as unused.

Should we rename `i18n_key` to something different as well?


AntonKhorev:
> Should we rename i18n_key to something different as well?

Obviously if you do this change, it's not a key anymore.

AntonKhorev:
Thanks, merged.

PR #5147, opened 2024-09-02
Allow more formats for internal latlon search

### LINKED ISSUE #2366: Search by latitude & longitude succeeds inconsistently
I'll frequently search for a location by latitude and longitude, in either degrees/minutes/seconds, or decimal degrees.

Even using the same format, some queries succeed and display an appropriate map, others fail to find a location.

Search by latitude and longitude should accept reasonably configured inputs, and show the corresponding location on the map.

Examples of input which should be accepted (others may also apply):

```
28.59° N, 81.60° W
28.59 N, 81.60 W
28.59, -81.60
28.59°10"20' N, 81.60°10"20' W
28.59°10"20' -81.60°10"20'
```

AntonKhorev:
Fixes #2366 by handling all mentioned formats. The new ones are:
- `28°10'20" -81°10'20"` - deg/min/sec markers without NSEW
- `28° 34' 31,098" N 16° 11' 14,592" W` - decimal comma with NSEW

### DIFF
--- a/app/controllers/geocoder_controller.rb
+++ b/app/controllers/geocoder_controller.rb
@@ -213,7 +213,11 @@ class GeocoderController < ApplicationController
                       :lon => dms_to_decdeg("ew", "ew", captures)
         params.delete(:query)
 
-      elsif match = query.match(%r{^(?<ns>[+-]?)\s*#{dms_regexp('ns')}(?:\s+|\s*[,/]\s*)(?<ew>[+-]?)\s*#{dms_regexp('ew')}$})
+      elsif match = query.match(%r{^
+                      (?<ns>[+-]?)\s*#{dms_regexp('ns', :comma => false)}
+                      (?:\s+|\s*[,/]\s*)
+                      (?<ew>[+-]?)\s*#{dms_regexp('ew', :comma => false)}
+                    $}x)
         captures = match.named_captures.compact
         params.merge! :lat => dms_to_decdeg("ns", "+-", captures),
                       :lon => dms_to_decdeg("ew", "+-", captures)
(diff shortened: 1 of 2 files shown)


AntonKhorev:
> rounding problems

What are these problems? 0.001 seconds is less that 1 pixel on max zoom.

nenad-vujicic:
> > rounding problems
> 
> What are these problems? 0.001 seconds is less that 1 pixel on max zoom.

Yes, I meant it just to follow other cases (look at screenshot). As I wrote above, I am super-satisfied with this PR's solution and only proposed improvement in future.

image

Thanks.

AntonKhorev:
Values are left intact as strings if there are no minutes/seconds. Do you want mins/secs in search results instead of decimal degrees if mins/secs were in the query?

nenad-vujicic:
> Values are left intact as strings if there are no minutes/seconds. Do you want mins/secs in search results instead of decimal degrees if mins/secs were in the query?

If others agree, yes. Whatever is written in search box (coordinates string) should be simply copied to search results (again coordinates string), without converting to single / double precision between. So, if user enters 44°49′04″N 20°27′25″E, we should display 44°49′04″N 20°27′25″E, not 44.81778,20.45694. Again, iff others agree.

test/controllers/geocoder_controller_test.rb
```diff
@@ -57,6 +88,7 @@ def test_identify_latlon_ne_d
       "50.06773N, 14.37742E"
     ].each do |code|
       latlon_check code, 50.06773, 14.37742
+      assert_nil @controller.params[:latlon_digits]
```
tomhughes, line 91: Why do we want to assert this? The only thing this test should be concerned about is what results it gets from the controller method - how the controller does that internally is none of it's concern?

app/controllers/geocoder_controller.rb
```diff
@@ -206,42 +206,51 @@ def normalize_params
...
-    lat = ns * (nsd + (nsm / 60) + (nss / 3600))
-    lon = ew * (ewd + (ewm / 60) + (ews / 3600))
-
-    { :lat => lat.round(6).to_s("F"), :lon => lon.round(6).to_s("F") }
+  def dms_to_decdeg(prefix, directions, captures)
+    extract_number = ->(suffix) { captures.fetch("#{prefix}#{suffix}", "0").sub(",", ".") }
+
+    positive_direction, negative_direction = directions.chars
+    sign = captures.fetch(prefix, positive_direction).casecmp?(negative_direction) ? "-" : ""
+    deg = if captures["#{prefix}m"] || captures["#{prefix}s"]
```
tomhughes, line 246: Why do we need to special case the case when there are only degrees? What was wrong with the old way of doing it?

test/controllers/geocoder_controller_test.rb
```diff
@@ -44,6 +44,37 @@ def test_identify_latlon_basic
...
+      "+25.79°, -80.27°"
+    ].each do |code|
+      latlon_check code, 25.79, -80.27
+      assert @controller.params[:latlon_digits]
+    end
+  end
+
+  ##
+  # Test identification of basic lat/lon pairs with degrees/mins
+  def test_identify_latlon_basic_dm
```
tomhughes, line 69: Should we not have a dms version as well?

app/controllers/geocoder_controller.rb
```diff
@@ -206,42 +206,51 @@ def normalize_params
...
-
+      if match = query.match(/^(?<ns>[NS])\s*#{dms_regexp('ns')}\W*(?<ew>[EW])\s*#{dms_regexp('ew')}$/) ||
+                 query.match(/^#{dms_regexp('ns')}\s*(?<ns>[NS])\W*#{dms_regexp('ew')}\s*(?<ew>[EW])$/)
+        captures = match.named_captures.compact
+        params.merge! :lat => dms_to_decdeg("ns", "ns", captures),
+                      :lon => dms_to_decdeg("ew", "ew", captures)
+        params.delete(:query)
+
+      elsif match = query.match(%r{^
+                      (?<ns>[+-]?)\s*#{dms_regexp('ns', :comma => false)}
```
tomhughes, line 217: Why not allow commas in this case? Is it because it conflicts with the comma separating lat and lon? Having to exclude it in this case makes things a lot more ugly :-(

app/controllers/geocoder_controller.rb
```diff
@@ -206,42 +206,51 @@ def normalize_params
...
-    ewd = BigDecimal(captures.fetch("ewd", "0"))
-    ewm = BigDecimal(captures.fetch("ewm", "0"))
-    ews = BigDecimal(captures.fetch("ews", "0"))
-
-    lat = ns * (nsd + (nsm / 60) + (nss / 3600))
-    lon = ew * (ewd + (ewm / 60) + (ews / 3600))
-
-    { :lat => lat.round(6).to_s("F"), :lon => lon.round(6).to_s("F") }
+  def dms_to_decdeg(prefix, directions, captures)
+    extract_number = ->(suffix) { captures.fetch("#{prefix}#{suffix}", "0").sub(",", ".") }
```
tomhughes, line 242: Is this really better than just adding the `.sub` to the four fetches? I understand it's removing duplication but it does make the code much harder to read and understand.

test/controllers/geocoder_controller_test.rb
AntonKhorev: This entire file tests `@controller.params` all over the place. It has been doing it for a long time (https://github.com/openstreetmap/openstreetmap-website/commit/aaeca5b5346b01d7ef26a301492804d36049eff8) and you didn't mind adding more recently (https://github.com/openstreetmap/openstreetmap-website/commit/6d217da3b65be4f2973c536fd8fb0765e2a6e450).

app/controllers/geocoder_controller.rb
AntonKhorev: Because we don't need to do min/sec math with divisions and results that have to be rounded to some precision. This special case was previously here: https://github.com/openstreetmap/openstreetmap-website/blob/1fce0c00f12ac6bbb56ad7c05c8c9f53c0d6edb3/app/controllers/geocoder_controller.rb#L213-L214 except it was triggered not by a lack of min/sec but by the entire query being in a simplest format possible. So you had [1.1111111111 2.2222222222](https://www.openstreetmap.org/search?query=1.1111111111%202.2222222222) keeping lat/lon intact but [1.1111111111 N 2.2222222222 E](https://www.openstreetmap.org/search?query=1.1111111111%20N%202.2222222222%20E) going into math mode with rounding.

AntonKhorev:
The correct solution to various problems here is to move the parsing code to a library, maybe even make a gem. Then we won't have to test changes to parameters.

PR #5149, opened 2024-09-02
Fixed bug causing bad error message when requested missing changeset

### LINKED ISSUE #5148: Strange error message for not found changeset ID
### URL

https://www.openstreetmap.org/changeset/1234567890

Just try some changeset ID which doesn't exist. Error messages about not found nodes, ways, relations are fine, the problem is only with changesets.

image

nenad-vujicic:
PR fixes bug causing bad error message when requested displaying missing changeset

Fixes #5148. Added forgotten @type = "changeset" in ChangesetsController#show.

Before:
image
After:
image

### DIFF
--- a/app/controllers/changesets_controller.rb
+++ b/app/controllers/changesets_controller.rb
@@ -75,6 +75,7 @@ class ChangesetsController < ApplicationController
   end
 
   def show
+    @type = "changeset"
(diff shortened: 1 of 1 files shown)


AntonKhorev:
Merged, thanks.

PR #5150, opened 2024-09-02
Bootstrap notes pagination

AntonKhorev:
Smaller PR compared to #4532. I'm just changing how the paging buttons look on `/user/:name/notes` pages.

Before:
image

After:
image


tomhughes:
It's smaller, but is it actually a useful step towards moving to cursor based pagination? or is it a dead end?

AntonKhorev:
I'm also moving locale strings from changeset_paging_nav because they are not used for changeset paging. That doesn't depend on having or not having cursor based pagination.

tomhughes:
This is a definite improvement so we might as well merge it, thanks.

PR #5152, opened 2024-09-03
Remove OAuth 1 ApiCapability user branch

AntonKhorev:
I looked at Coveralls and noticed that `token.user` is never accessed:

image

I guess this branch is not required after #5058.

### DIFF
--- a/app/abilities/api_capability.rb
+++ b/app/abilities/api_capability.rb
@@ -5,7 +5,7 @@ class ApiCapability
 
   def initialize(token)
     if Settings.status != "database_offline"
-      user = (User.find(token.resource_owner_id) if token.respond_to?(:resource_owner_id))
+      user = User.find(token.resource_owner_id)
 
       if user&.active?
         can [:create, :comment, :close, :reopen], Note if scope?(token, :write_notes)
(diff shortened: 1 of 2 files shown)


app/abilities/api_capability.rb
```diff
@@ -5,11 +5,7 @@ class ApiCapability
 
   def initialize(token)
     if Settings.status != "database_offline"
-      user = if token.respond_to?(:resource_owner_id)
-               User.find(token.resource_owner_id)
-             elsif token.respond_to?(:user)
-               token.user
-             end
+      user = (User.find(token.resource_owner_id) if token.respond_to?(:resource_owner_id))
```
tomhughes, line 8: We don't actually need the condition at all - we should always have a token and it should always be an OAuth 2 token now.

There is a test that will need fixing but it's a bogus test that deliberately passes `nil` as the token which is not valid. The real code never does that.

tomhughes (changes requested):

app/abilities/api_capability.rb
AntonKhorev: Removed that part of the test because you don't have any abilities by default and it only tested `cannot?`.

tomhughes (approved):
Thanks - that looks good to me now.

PR #3038, opened 2021-01-05
Remove both Potlatch versions

tomhughes:
Fixes #2622

PR #5153, opened 2024-09-03
Remove node.tags_as_hash method

AntonKhorev:
Not used, was last used for Potlatch undelete that was removed in  #3038.

`node.tags` is a hash anyway.


tomhughes:
Looks good to me, thanks.

PR #5154, opened 2024-09-03
Change bearer_authorization_header(token.token) to (token)

mmd-osm:
Follow up for https://github.com/openstreetmap/openstreetmap-website/pull/5141#issuecomment-2325069540

PR #5155, opened 2024-09-03
Simplify creation of bearer authorization headers

tomhughes:
Uses the new features added to the bearer_authorization_header while dropping basic authentication support to simplify the creation of bearer authorization headers for API tests.

This is an alternative to #5154 that goes further in simplifying things.


AntonKhorev:
Merged, thanks.

Issue #3289, opened 2021-08-07
Add search box to diaries page


jidanni:
On https://www.openstreetmap.org/diary
add a box to search, for e.g.,
- diaries by a certain user
- diaries on a certain subject
- even: diaries containing a certain string.

mmd-osm:
> diaries by a certain user
* Use https://www.openstreetmap.org/user/jidanni/diary

mvexel:
I just [wrote about this on my diary](https://www.openstreetmap.org/user/mvexel/diary/399950)! This would be a lovely feature to have.

mvexel:
There was an interesting suggestion in the comments section of my diary entry. It seems that search engines are pretty capable to handle a diary-specific search given the right syntax. I am guessing most users coming across the User Diaries page do not know how to compose a search command like this. So I suggested (copying from the diary comment):

> Perhaps there is a way to create a search box that parses whatever the user enters and redirects (new tab) to search engine results. So a user would enter, for example, `openstreetcam` and this would open a tab with `https://duckduckgo.com/?q=site%3Aopenstreetmap.org%2Fuser+diary+%22openstreetcam%22`.

As an added bonus, it would expose more users to a privacy-centered search engine.

Would that be a feasible implementation of this?

mvexel:
Someone suggested documenting search strategies using existing external search engine syntax, so I took the [User Diaries OSM wiki page](https://wiki.openstreetmap.org/w/index.php?title=User_diaries) (which was a redirect) and added some hopefully relevant content there.

kcne:
### Progress on Implementing Search for Diaries

I've explored several approaches to add search functionality to the diaries page. Here’s a summary of what has been done so far, along with some considerations and questions moving forward:

1. **PostgreSQL `LIKE` Text Searching:**
   - **Method:** Implemented a basic text search using PostgreSQL’s `LIKE` operator.
   - **Results:** This method is fast and responsive, but the relevance of the results may not be ideal since `LIKE` searches do not inherently provide any ranking mechanism.
   - **Code Example:**
     ```ruby
     entries = entries.where("title LIKE :query OR body LIKE :query", query: search_query)
     ```

2. **Using the `pg_search` Gem:**
   - **Method:** Integrated the `pg_search` gem to enable full-text search capabilities.
   - **Results:** This allows for more sophisticated search capabilities, including relevance ranking and partial word matching. However, with around 600,000 entries in the database, this approach has shown to be quite slow.
[...]

PR #5156, opened 2024-09-03
Implement Balanced Full-Text Search for Diary Entries

kcne:
This PR adds search functionality to the diary entries page, addressing the need for users to find entries by title and body #3289 

### Key Changes

- **Model Updates**:
  - Integrated the `pg_search` gem to enable full-text search on diary entries with relevance ranking.
  - Added a `searchable` column using a `tsvector` for efficient searching and created an index for performance optimization.

- **Controller Enhancements**:
  - Modified the `index` action in `DiaryEntriesController` to support search queries, allowing users to filter results by keywords in titles and bodies, with optional language filtering.

- **Testing**:
  - Added tests to verify the search functionality, including keyword search, language filtering, and basic edge cases.

### Context

Previous methods explored for implementing search functionality, such as the PostgreSQL LIKE operator and pg_trgm search, were either too fast but lacked relevance ranking or too slow to be practical, especially with large datasets like the diary entries (600,000+ records). For instance, LIKE provided speed but no relevance, while pg_trgm could take over 40 seconds to run on a dataset of this size, making it unfeasible. The pg_search gem, utilizing tsearch, offers a balance between relevance and performance, though it required additional database migrations and optimizations. Despite the need for these migrations, this approach serves as a "golden middle," providing a reasonable trade-off between speed and search result quality IMHO.

### Commit Summary

- **Add pg_search gem**: Introduced `pg_search` for full-text search capabilities.
- **Add searchable column to diary entries migration**: Created a `tsvector` column for storing precomputed search data.
- **Add index to searchable diary entries migration**: Indexed the `searchable` column for performance improvements.
- **Add search query logic to diary entry model and controller**: Integrated search functionality into the model and controller.
- **Add search functionality tests to DiaryEntriesController**: Added basic tests to verify the new functionality.

### Work in Progress

This is a draft PR. I want to confirm if I'm heading in the right direction. If the approach is approved, I'll add more tests for better coverage and handling of edge cases. Any comments and recommendations welcome.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,9 @@ gem "json"
 # Use postgres as the database
 gem "pg"
 
+# Use postgres search
+gem "pg_search"
+
 # Use SCSS for stylesheets
 gem "dartsass-sprockets"
 # Pin the dependentent sass-embedded to avoid deprecation warnings in bootstrap
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -414,6 +414,9 @@ GEM
       ast (~> 2.4.1)
       racc
     pg (1.5.7)
+    pg_search (2.3.7)
+      activerecord (>= 6.1)
+      activesupport (>= 6.1)
     popper_js (2.11.8)
     progress (3.6.0)
     psych (5.1.2)
@@ -680,6 +683,7 @@ DEPENDENCIES
   openstreetmap-deadlock_retry (>= 1.3.1)
   overcommit
   pg
+  pg_search
   puma (~> 5.6)
   quad_tile (~> 1.0.1)
   rack-cors
--- a/app/controllers/diary_entries_controller.rb
+++ b/app/controllers/diary_entries_controller.rb
@@ -244,3 +244,47 @@ class DiaryEntriesController < ApplicationController
     end
   end
 end
+
+def test_index_search
+  user = create(:user)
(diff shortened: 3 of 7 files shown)


kcne:
Any guidance on this would be highly appreciated @tomhughes @gravitystorm @AntonKhorev

tomhughes:
Well waiting more than 18 hours before chasing would be my first advice...

The first question really, is whether there is need for this. Other than that one request has there been any other interest in this that you're aware of?

kcne:
Thanks for the feedback. I commented on the issue earlier and thought I'd get a response there, but I realize now I should have tagged you. Apologies for pinging without waiting.

Beyond the issue and comments, I haven’t noticed any additional demand. However, I think this could improve the user experience, especially for those looking for specific content in diaries.

If it’s not a priority right now, I’m happy to pause or adjust as needed.

Also, out of context, but from a contributor perspective, maintaining tags like "help needed" or priority tags on issues would greatly help in understanding which issues are a priority and which are still open for discussion.

tomhughes:
I mean if we're going to do it then this is probably the right approach though I'll need to brush up on freetext searching in postgres before I can comment on the details.

We do already have support for freetext searching on notes as a model though it may not be a good one.

kcne:
Closing as not planned.

PR #5157, opened 2024-09-04
Switch from ubuntu to debian and refactor Dockerfile

### LINKED ISSUE #5010: NodeJS versions are out of sync again

kcne:
Fixes #5010

A few notes regarding this issue and PR:

- Since @tomhughes mentioned in this [comment](https://github.com/openstreetmap/openstreetmap-website/pull/5070#issuecomment-2329719170) that all the production and development servers are now on Debian 12, I thought this would be a good time to sync the Docker image and revisit the previous work.
- The issues noted in this [comment](https://github.com/openstreetmap/openstreetmap-website/issues/5010#issuecomment-2261537069) have been resolved. These were likely caused by geckodriver v0.34 being incompatible with the newer version of Firefox, which was reflected in the test logs afterwards.
- Node.js and osmosis are now installed as a system packages, and ESLint is working correctly out of the box again.
- I encountered an issue when running tests for the first time inside the Docker container, where some assets were missing. Running `bundle exec rake assets:precompile` resolved these errors.
- Everything seems to be working fine now, but it would be helpful if a few of us could test this locally before merging.

Let me know if you have any feedback or suggestions!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,51 +1,39 @@
-FROM ubuntu:22.04
+FROM debian:bookworm
 
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Install system packages then clean up to minimize image size
 RUN apt-get update \
- && apt-get install --no-install-recommends -y \
-      build-essential \
-      curl \
-      default-jre-headless \
-      file \
-      git-core \
-      gpg-agent \
-      libarchive-dev \
-      libffi-dev \
-      libgd-dev \
-      libpq-dev \
-      libsasl2-dev \
-      libvips-dev \
-      libxml2-dev \
-      libxslt1-dev \
-      libyaml-dev \
-      locales \
-      postgresql-client \
-      ruby \
-      ruby-dev \
-      ruby-bundler \
-      software-properties-common \
-      tzdata \
-      unzip
-
-# Install Node.js 18 and npm
-RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
- && apt-get install -y nodejs
+  && apt-get install --no-install-recommends -y \
+  build-essential \
+  curl \
+  default-jre-headless \
+  file \
+  git-core \
+  gpg-agent \
+  libarchive-dev \
+  libffi-dev \
+  libgd-dev \
+  libpq-dev \
+  libsasl2-dev \
+  libvips-dev \
+  libxml2-dev \
+  libxslt1-dev \
+  libyaml-dev \
+  locales \
+  postgresql-client \
+  ruby-dev \
+  ruby-bundler \
+  tzdata \
+  unzip \
+  nodejs \
+  npm \
+  osmosis \
+  ca-certificates \
+  firefox-esr
 
 # Install yarn globally
-RUN npm install --global yarn \
- # We can't use snap packages for firefox inside a container, so we need to get firefox+geckodriver elsewhere
- && add-apt-repository -y ppa:mozillateam/ppa \
- && echo "Package: *\nPin: release o=LP-PPA-mozillateam\nPin-Priority: 1001" > /etc/apt/preferences.d/mozilla-firefox \
- && apt-get install --no-install-recommends -y \
-      firefox-geckodriver \
- && apt-get clean \
- && rm -rf /var/lib/apt/lists/*
-
-# Install compatible Osmosis to help users import sample data in a new instance
-RUN curl -OL https://github.com/openstreetmap/osmosis/releases/download/0.47.2/osmosis-0.47.2.tgz \
- && tar -C /usr/local -xzf osmosis-0.47.2.tgz
+RUN npm install --global yarn
 
 ENV DEBIAN_FRONTEND=dialog
 
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -75,11 +75,15 @@ Run the Rails database migrations:
 
 Prepare the test database:
 
-     docker compose run --rm web bundle exec rails db:test:prepare
+    docker compose run --rm web bundle exec rails db:test:prepare
 
 Run the test suite:
 
     docker compose run --rm web bundle exec rails test:all
+
+If you encounter errors about missing assets, precompile the assets:
+
+    docker compose run --rm web bundle exec rake assets:precompile
 
 ### Loading an OSM extract
 


mmd-osm:
> I encountered an issue when running tests for the first time inside the Docker container, where some assets were missing. Running bundle exec rake assets:precompile resolved these errors.

I'd suggest  to include this info in our readme file: https://github.com/openstreetmap/openstreetmap-website/blob/master/DOCKER.md

> Everything seems to be working fine now, but it would be helpful if a few of us could test this locally before merging.

Same here. I tested on a fresh install. Everything is looking good here.

One thing I don't particularly like about the Dockerfile is that the web container runs everything as root user. We should revisit this topic in another issue. The current Dockerfile has the same issue, and it's unrelated to the switch to Debian.

mmd-osm (approved):

nenad-vujicic (approved):
It works great on my side. Thanks!

Dockerfile
```diff
@@ -1,51 +1,45 @@
...
-
-# Install compatible Osmosis to help users import sample data in a new instance
-RUN curl -OL https://github.com/openstreetmap/osmosis/releases/download/0.47.2/osmosis-0.47.2.tgz \
- && tar -C /usr/local -xzf osmosis-0.47.2.tgz
+RUN npm install --global yarn
+
+# Install geckodriver and firefox
+RUN apt-get update                             \
+  && apt-get install -y --no-install-recommends \
+  ca-certificates curl firefox-esr           \
```
tomhughes, line 41: These installs can be done with all the others now - they were only separate because we needed to install `software-properties-common` first to get `add-apt-repository` which was then used here.

Dockerfile
```diff
@@ -1,51 +1,45 @@
...
+  libvips-dev \
+  libxml2-dev \
+  libxslt1-dev \
+  libyaml-dev \
+  locales \
+  postgresql-client \
+  ruby \
+  ruby-dev \
+  ruby-bundler \
+  software-properties-common \
```
tomhughes, line 28: I think this was only needed for `add-apt-repository` so it can probably be dropped now.

tomhughes (changes requested):

Dockerfile
kcne: `ruby` and `software-properties-common` can be dropped you are right
- When `ruby-bundler` is removed I get ` bundle: command not found` error
- When `ruby-dev` is removed errors are encountered while running `bundle install`

tomhughes: It was only `software-properties-common` that I was referring to.

Dockerfile
kcne: Great catch, moved that so they are installed with others

Dockerfile
```diff
@@ -1,51 +1,45 @@
...
-# Install compatible Osmosis to help users import sample data in a new instance
-RUN curl -OL https://github.com/openstreetmap/osmosis/releases/download/0.47.2/osmosis-0.47.2.tgz \
- && tar -C /usr/local -xzf osmosis-0.47.2.tgz
+RUN npm install --global yarn
+
+# Install geckodriver and firefox
+RUN apt-get update                             \
+  && apt-get install -y --no-install-recommends \
+  ca-certificates curl firefox-esr           \
+  && curl -L https://github.com/mozilla/geckodriver/releases/download/v0.35.0/geckodriver-v0.35.0-linux64.tar.gz | tar xz -C /usr/local/bin
```
tomhughes, line 42: I think we might be able to drop this - it seems that the ruby webdrivers gem is able to download this itself as needed now.

kcne:
> I'd suggest to include this info in our readme file: https://github.com/openstreetmap/openstreetmap-website/blob/master/DOCKER.md

Definitely agree, I'm just not sure if we should do it here on in separate PR.


> One thing I don't particularly like about the Dockerfile is that the web container runs everything as root user. We should revisit this topic in another issue. The current Dockerfile has the same issue, and it's unrelated to the switch to Debian.

I'm not sure how managing users works in Docker and what are the pros/cons of using approach you mentioned.

mmd-osm:
A dedicated user is a security feature and recommended to prevent privilege escalation attacks, e.g. https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-2-set-a-user

We'll address this topic in another issue...

Dockerfile
mmd-osm: Tests are passing without downloading "curl -L https://github.com/mozilla/geckodriver/releases/..." first. So yes, installing geckodriver could be dropped as well.

nertc (approved):
Tested this on both existing and new clone of the application. Tests and linters work as intended. Application starts and works perfectly. There are no errors while building.

Dockerfile
kcne: You're right. Tests are passing without it. I removed it accordingly. Thank you @tomhughes @mmd-osm

Dockerfile
kcne: Alright, resolved. Sorry for the confusion.

kcne:
Amended the last commit with last required changes and marked this as ready for review. Thank you guys for reviews and taking time to test this locally.

tomhughes (approved):

PR #5163, opened 2024-09-06
Remove Oauth::OAUTH2_SCOPES

AntonKhorev:
After OAuth1 was removed, all scopes are OAuth2 scopes. Former OAuth2-only scopes now can be combined with the rest.


tomhughes:
Yes I noticed that the other day and was meaning to have a look at it so that looks good to me, thanks.

PR #5168, opened 2024-09-08
Disable redactions with write_api scope

AntonKhorev:
Continues #4387.

Requires `write_redactions` scope to redact. Previously it was possible to redact with either `write_redactions` or `write_api`.


tomhughes:
Are DWG ready for this? All their tooling is now using `write_redactions` instead of `write_api`?

AntonKhorev:
If anyone is still using `write_api`, it could only be because they have some old version of scripts somewhere which they forgot to update.

tomhughes:
Fine, then this is good to go I think. Thanks.

PR #5171, opened 2024-09-08
Remove unused messages.outbox.actions string

AntonKhorev:
Was added in https://github.com/openstreetmap/openstreetmap-website/commit/3e69a4512f2d1b39f48719882a0d112d042da73a together with `messages.messages_table.actions`, but only the latter is in use.


tomhughes:
Looks good to me, thanks.

PR #5173, opened 2024-09-08
Move ?: operator outside of t() for "Only ... can edit" block messages

AntonKhorev:
This helps `i18n-tasks` to see that these strings are used and stops `i18n-tasks unused` from reporting them.


tomhughes:
Looks good to me, thanks.

PR #4734, opened 2024-04-28
Add oldest/newest links to shared pagination - with gaps around "newer" and "older" v2

AntonKhorev:
#4733 without "oldest"/"newest" button text on any screen width and with `title` tooltips.

image
image

Partly addresses #2387 ("No way to jump to the last page").

PR #4733, opened 2024-04-28
Add oldest/newest links to shared pagination - with gaps around "newer" and "older"

AntonKhorev:
Alternative split, see https://github.com/openstreetmap/openstreetmap-website/pull/4729#issuecomment-2081255352

image
image
image

PR #5174, opened 2024-09-08
Shared pagination: label instead of i18n key as parameter

mmd-osm:
Pass translated label to shared pagination instead of i18n key. 

This reduces the number of reported issues when running `bundle exec i18n-tasks unused`

Apparently, there are lots of pending PR out there that would conflict with this change: #4734, #4733, #4729, #4710
What's the status here? Which one of these alternatives will make it in the end?

### DIFF
--- a/app/views/diary_comments/_page.html.erb
+++ b/app/views/diary_comments/_page.html.erb
@@ -19,8 +19,8 @@
   </table>
 
   <%= render "shared/pagination",
-             :newer_key => "diary_comments.page.newer_comments",
-             :older_key => "diary_comments.page.older_comments",
+             :newer_label => t("diary_comments.page.newer_comments"),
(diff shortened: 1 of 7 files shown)


AntonKhorev:
In #4734 etc I need four different labels, so instead I pass the scope and append `.newest` / `.newer` / `.older` / `.oldest`. And what if I need even more strings, for example, a different tooltip string for each button? Do you propose to pass all of them separately?

mmd-osm:
4 strings might be still ok, for more strings it would be good to have an alternative approach.

AntonKhorev:
If the goal is to make `i18n-tasks unused` stop complaining we can create a scope just for pagination strings and tell i18n-tasks to ignore it.

PR #5177, opened 2024-09-09
Move notification code out of note update transactions

AntonKhorev:
We don't need to notify anyone until the transaction is completed. The notification code was inside the transaction since the very beginning in https://github.com/openstreetmap/openstreetmap-website/commit/b3d62bb85a03a0b68068d3d2e7801ec2cd2e3e7b and I don't know why.

### DIFF
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -82,7 +82,7 @@ module Api
       # Extract the arguments
       lon = OSM.parse_float(params[:lon], OSM::APIBadUserInput, "lon was not a number")
       lat = OSM.parse_float(params[:lat], OSM::APIBadUserInput, "lat was not a number")
-      comment = params[:text]
+      text = params[:text]
 
       # Include in a transaction to ensure that there is always a note_comment for every note
       Note.transaction do
@@ -94,7 +94,7 @@ module Api
         @note.save!
 
         # Add a comment to the note
-        add_comment(@note, comment, "opened")
+        add_comment(@note, text, "opened")
       end
 
       # Return a copy of the new note
@@ -112,7 +112,7 @@ module Api
 
       # Extract the arguments
       id = params[:id].to_i
-      comment = params[:text]
+      text = params[:text]
 
       # Find the note and check it is valid
       Note.transaction do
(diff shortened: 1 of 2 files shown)


mmd-osm:
Was the notification previously sent immediately or only at the time of committing the transaction? Which issue is this change trying to solve?

AntonKhorev:
It wasn't sent, it was scheduled to be sent but was it still correct? I actually want to do one more thing inside the transaction after adding a comment. How am I supposed to do that if I keep the notifications where they are now, inside the comment method?

mmd-osm:
Well, I was commenting on this question:

> The notification code was inside the transaction since the very beginning in ... and I don't know why.

If you keep everything in a single transaction, it's either all or nothing. If you split it up in multiple transactions, creating a comment may succeed, but later on, the notification may still fail independently. Usually, you try to avoid multiple transactions for a single request, because it can cause some inconsistent results.

AntonKhorev:
There's one transaction. If it fails it doesn't roll back notifications.

mmd-osm:
Based on what you wrote earlier on ("it was scheduled to be sent"), I assumed that notifications are first written to the database, then picked up by another job later on which is doing the real work. This should only happen after the transaction was committed to the database. I haven't checked how the scheduling part is implemented exactly. If writing comments and scheduling notifications are not part of the same transaction today, then this argument does not hold anymore.

AntonKhorev:
I tried throwing an exception from within the transaction. Database changes got rolled back but notifications were still scheduled.

tomhughes:
In our production configuration they are written to the database for later processing but in development they aren't which may explain the confusion...

AntonKhorev:
Are `UserMailer...deliver_later` intentionally inside the transaction here? If yes, why aren't they in other places?

https://github.com/openstreetmap/openstreetmap-website/blob/a39886ac4c3deef56e0bb47a883322b635fe0e1f/app/controllers/diary_comments_controller.rb#L36-L44

https://github.com/openstreetmap/openstreetmap-website/blob/a39886ac4c3deef56e0bb47a883322b635fe0e1f/app/controllers/friendships_controller.rb#L24-L26

https://github.com/openstreetmap/openstreetmap-website/blob/a39886ac4c3deef56e0bb47a883322b635fe0e1f/app/controllers/messages_controller.rb#L49-L51

etc

AntonKhorev:
It doesn't matter as long as nobody adds anything that should be inside the transaction and after the comments are saved.

PR #5178, opened 2024-09-09
Fix changeset comments feed timeout

AntonKhorev:
Was part of #4617 which is outdated now.

Comments feed timeout had a html template despite not being an html page and an atom template despite not being an atom feed.


tomhughes:
Looks good to me, thanks.

PR #5179, opened 2024-09-10
Translation scope for shared pagination

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/5174#issuecomment-2338001493

- "Older/Newer" strings are moved to the `shared.pagination` scope. 
- `shared.pagination` is ignored by `i18n-tasks`.
- Instead of receiving (multiple, currently two but possibly more in the future) translation keys, the `shared/pagination` template receives one translation scope.
- Actually it's not necessary to pass this scope in most of the cases because it can be guessed from the controller name.

### DIFF
--- a/app/views/diary_comments/_page.html.erb
+++ b/app/views/diary_comments/_page.html.erb
@@ -19,7 +19,6 @@
   </table>
 
   <%= render "shared/pagination",
-             :translation_scope => "shared.pagination.diary_comments",
              :newer_id => @newer_comments_id,
              :older_id => @older_comments_id %>
 </turbo-frame>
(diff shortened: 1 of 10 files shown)


tomhughes:
Tests always run in english - more specifically they will be using `I18n.default_locale` as their locale unless they deliberately change to something else and that defaults to `:en` and is never changed by us.

tomhughes:
Actually as this is a system test the story is slightly different but for a user that is no logged in the locale will be derived from the user agent's `Accept-Language` header and we explicitly set that to english (https://github.com/openstreetmap/openstreetmap-website/blob/master/test/application_system_test_case.rb#L11)  and if a logged in user has no language set (as here) then the first request to try and set a locale will set it from the browser (https://github.com/openstreetmap/openstreetmap-website/blob/a39886ac4c3deef56e0bb47a883322b635fe0e1f/app/controllers/application_controller.rb#L205).

tomhughes:
Looks good to me, thanks.

PR #5181, opened 2024-09-10
Limit the number of users counted for users#index

tomhughes:
This view uses cursor based pagination but then undermines that by counting the total number of users so limit the effect of that but counting a maximum of ten pages of users.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -38,7 +38,9 @@ class UsersController < ApplicationController
       users = users.where(:status => @params[:status]) if @params[:status]
       users = users.where(:creation_ip => @params[:ip]) if @params[:ip]
 
-      @users_count = users.count
+      @users_count = users.limit(501).count
+      @users_count = I18n.t("count.at_least_pattern", :count => 500) if @users_count > 500
+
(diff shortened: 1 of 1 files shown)


app/controllers/users_controller.rb
```diff
@@ -38,7 +38,9 @@ def index
       users = users.where(:status => @params[:status]) if @params[:status]
       users = users.where(:creation_ip => @params[:ip]) if @params[:ip]
 
-      @users_count = users.count
+      @users_count = users.limit(501).count
+      @users_count = "500+" if @users_count > 500
```
AntonKhorev, line 42: I was told [this](https://github.com/openstreetmap/openstreetmap-website/pull/4201#discussion_r1310176384) about "number+".

tomhughes: Thanks - I've fixed that now.

AntonKhorev:
@gravitystorm also had a [different opinion](https://github.com/openstreetmap/openstreetmap-website/pull/4201#discussion_r1310163561) about *499, 500, 500+* vs *499, 500+*. But I guess it's not as important as enabling some kind of a limit.

PR #5182, opened 2024-09-10
Add creation_address column to users

tomhughes:
This is a properly typed and indexed column that is intended to replace creation_ip to allow fast queries.

A future PR will migrate existing data and update the site to use the new field after which the old one can be dropped.


AntonKhorev:
Merged, thanks.

PR #5184, opened 2024-09-11
Rename _blocks partial template to _page

AntonKhorev:
Similar templates for users, traces, comments etc are named "page", only the blocks one is named differently. That's because it predates the others, being added in https://github.com/openstreetmap/openstreetmap-website/commit/daa2496024cf0bdc65e58eb6e2802784f411f68b.


tomhughes:
I did consider this when I was doing the turbo work and decided not to bother but I guess it doesn't hurt.

PR #5187, opened 2024-09-11
Pulled out ?: operator from tr in changesets/index

nenad-vujicic:
PR pulls out ?: operator from tr in changesets/index.html.erb, so, changesets.index.{empty, empty_area, empty_user, no_more, no_more_area, no_more_user} translation keys are not more reported as unused.


AntonKhorev:
Merged, thanks.

PR #5188, opened 2024-09-11
Removed unused translations in user_role namespace

nenad-vujicic:
PR removes user_role.{grant,revoke}.{title, heading, confirm} translation keys from en.yml started / ended using in 30d5e78 and 8a7040e / 5f33656.


tomhughes:
Looks good to me, thanks.

PR #5189, opened 2024-09-11
Replace 'remove this redaction' confirmation and request with Turbo

AntonKhorev:
The existing `:method => "delete" ...` don't entirely work for some reason. The browser gets stuck on redirects after the request is done.

image


nenad-vujicic (approved):
Works great on my side, thank you!

tomhughes:
Looks good to me, thanks.

PR #5077, opened 2024-08-14
Add locale selector

nertc:
This PR addresses "Language switcher on the homepage" issue mentioned in https://github.com/openstreetmap/openstreetmap-website/issues/1115

The code is based on @tomhughes 's PR https://github.com/openstreetmap/openstreetmap-website/pull/3618, which, as Tom mentioned in the [comment](https://github.com/openstreetmap/openstreetmap-website/pull/4461#issuecomment-1879660419), after rebasing had some styling issues. Most of the functional part of this PR was written by @tomhughes.

Language selector button, to be consistent, now is not hidden when the user is logged in. Instead, it redirects to the "My Preferences" page, where user can change language preferences. When the user is logged out, it still works like in https://github.com/openstreetmap/openstreetmap-website/pull/3618 _ shows "jquery uls" language selector popup / dropdown.

The reason for logged in user redirection instead of any complex functionality of preferred language change while clicking language selector, was compatibility with the PR https://github.com/openstreetmap/openstreetmap-website/pull/4461. After both PRs will be merged, there will be one smooth simple UX flow for logged in users.

In addition to these, this PR also adds:
- Tests for checking if redirection works correctly
- Both LTR and RTL language supported icon
- Language selector icon with the same default and hover colors as other menu items

Default view of the language selector:

image

This PR has many different visual impacts, so to make description shorter instead of directly adding images, I'll add links to them.

LTR screenshots:
- [Opened language selector](https://github.com/user-attachments/assets/0ee30d22-734d-491a-8355-acc7c0f51dba)
- [Medium resolution menu](https://github.com/user-attachments/assets/772b883a-7911-4f34-a81f-90a0c46a5f68)
- [Medium resolution menu hover](https://github.com/user-attachments/assets/cede63b4-447c-430e-b23a-28904dbdc3a2)
- [Small resolution menu](https://github.com/user-attachments/assets/18daace0-c617-4483-b7b0-9137906603bd)

RTL screenshots:
- [Default view](https://github.com/user-attachments/assets/6cc3b52d-09a6-4795-ab6c-6c56d2f361c2)
- [Medium resolution menu](https://github.com/user-attachments/assets/cb5306c5-96fa-418e-836e-71ef965f7fc5)
- [Small resolution menu](https://github.com/user-attachments/assets/7a0e913f-e59d-47a2-85c3-8db9e0e1c3e3)

PR #5201, opened 2024-09-12
Add locale selector

nertc:
This PR addresses "Language switcher on the homepage" issue mentioned in https://github.com/openstreetmap/openstreetmap-website/issues/1115

This PR is mainly based on https://github.com/openstreetmap/openstreetmap-website/pull/5077, https://github.com/openstreetmap/openstreetmap-website/pull/3618 and https://github.com/openstreetmap/openstreetmap-website/pull/4461. As solution using Wikimedia selector added external files, many lines of CSS and JS code, this version takes inspiration from https://github.com/openstreetmap/openstreetmap-website/pull/4461 and creates a reusable language selector using `<select>` tag.

Language selector button, to be consistent, is not hidden when the user is logged in. Instead, it redirects to the "My Preferences" page, where user can change language preferences. When the user is logged out, clicking language selector opens language selection menu.

The reason for logged in user redirection instead of any complex functionality of preferred language change while clicking language selector, was compatibility with the PR https://github.com/openstreetmap/openstreetmap-website/pull/4461. After both PRs will be merged, there will be one smooth simple UX flow for logged in users. Plus, it makes finding language selection page much easier for new users.

Default view of the language selector:

357859639-0ca3007a-ff80-480b-8b37-1816193dc377

Opened language selector:

image

Mobile view of the selector:

Screenshot_20240912-144716

This PR has many different visual impacts, so to make description shorter instead of directly adding images, I'll add links to them.

LTR screenshots:

[Medium resolution menu](https://github.com/user-attachments/assets/772b883a-7911-4f34-a81f-90a0c46a5f68)
[Medium resolution menu hover](https://github.com/user-attachments/assets/cede63b4-447c-430e-b23a-28904dbdc3a2)
[Small resolution menu](https://github.com/user-attachments/assets/18daace0-c617-4483-b7b0-9137906603bd)

RTL screenshots:

[Default view](https://github.com/user-attachments/assets/6cc3b52d-09a6-4795-ab6c-6c56d2f361c2)
[Medium resolution menu](https://github.com/user-attachments/assets/cb5306c5-96fa-418e-836e-71ef965f7fc5)
[Small resolution menu](https://github.com/user-attachments/assets/7a0e913f-e59d-47a2-85c3-8db9e0e1c3e3)


app/assets/javascripts/language_selector.js
```diff
@@ -0,0 +1,6 @@
+$(document).ready(function () {
+  $(".language-change-trigger").on("change", function () {
```
tomhughes, line 2: In the interests of ensuring turbo compatibility it's better to write this as:

```
$(document).on("change", ".language-change-trigger", function () {
```

it probably doesn't natter in this case as it's in the header but there's no reason not to be safe in case it ever gets moved.

tomhughes (changes requested):

AntonKhorev:
What's the advantage of adding the language icon to the sprite map (three times) as opposed to rendering an `<svg>` with `fill="currentColor"`?

nertc:
@AntonKhorev Currently, as I found in the project, for icons we use `.icon` class, which uses sprite with `background-image` and `background-position` properties. Adding icon to the sprite was solely for the accordance with the current state and practices of the project.

nertc:
Thank you for the review

@tomhughes I've updated accordingly
@AntonKhorev If it's okay to directly use SVG instead of the `.icon` class practice of the project, I am all for it. I think it's even better solution and produces cleaner, more reusable code.

AntonKhorev:
> Adding icon to the sprite was solely for the accordance with the current state and practices of the project.

Currently we use both `<svg>` and css sprite maps. One problem with sprite maps is that it's more difficult to control the colors of the icons.

app/assets/images/sprite.svg is mainly used for buttons inside Leaflet map views, where we don't entirely control html.

nertc:
@AntonKhorev Okay, thanks. I'll update it to `<svg>`. It will be more comfortable to reuse and will produce cleaner code.

nertc:
PR was updated to include SVG in the HTML. I tried several different methods to move SVG from sprite, including adding it to the SvgHelper module and adding separate file for it, but current solution turned out to be the most optimal one (SvgHelper had conflicts with the linter because of long lines and separate file was not sufficiently customizable with Bootstrap classes).

app/views/layouts/_header.html.erb
```diff
@@ -76,6 +85,15 @@
           <li><%= link_to t("layouts.copyright"), copyright_path, :class => "dropdown-item" %></li>
           <li><%= link_to t("layouts.help"), help_path, :class => "dropdown-item" %></li>
           <li><%= link_to t("layouts.about"), about_path, :class => "dropdown-item" %></li>
+          <li>
+            <% if current_user && current_user.id %>
+              <%= link_to(preferences_path) do %>
+                <%= render "shared/language_selector", :hoverable => true, :black => true, :classes => "dropdown-item", :disabled => true %>
```
AntonKhorev, line 91: `dropdown-item` here doesn't entirely work, and possibly other Bootstrap classes either don't work as expected or are missing. Some related odd behavior of language link/label:

- not logged in, no compact mode: no focus outline
- not logged in, compact mode: same + keyboard up/down controls stop working correctly
- logged in, no compact mode: focus outline is present but non-standard because no `nav-link`, but if you add it you'll need more css fixes
- logged in, compact mode: keyboard controls also not working correctly, in a slightly different manner

nertc: I'll try to solve the focusing issue without any custom classes. In terms of collapsibility of the translation icon, purpose was to take as little space as possible when the screen is smaller. Without collapsing, HTML code will be shorter and with less "if-else"-es.

AntonKhorev: Do you still want this icon collapsing into *More*?

nertc: I'll try to move it out of the "More" and test with different resolutions. If it doesn't break any resolution, I'll keep it out of the "More". It will need less HTML code.

nertc: I've updated this PR. Tried several solutions and current one needed the smallest and cleanest code. Also, one CSS class was added `.text-transparent` to make text and background of the `<select>` tag visually hidden but still focusable.

Language selector will no more be moved to the compact menu. Visually it doesn't take much space and contextually it's more relevant to be as an icon than as a compacted menu button.

app/views/layouts/_header.html.erb
```diff
@@ -78,6 +78,15 @@
           <li><%= link_to t("layouts.about"), about_path, :class => "dropdown-item" %></li>
         </ul>
       </li>
+      <li class="nav-item">
+        <% if current_user && current_user.id %>
+          <%= link_to(preferences_path, :class => "nav-link text-secondary ms-1") do %>
```
AntonKhorev, line 83: You have `ms-1` for disabled selector icons but no margin for enabled icons.

app/views/shared/_language_selector.html.erb
```diff
@@ -0,0 +1,12 @@
+<label class="position-relative d-flex <%= classes %>" role="button">
+  <svg width="20" height="20" fill="currentColor">
+    <path d="M20 18h-1.44a.61.61 0 0 1-.4-.12.81.81 0 0 1-.23-.31L17 15h-5l-1 2.54a.77.77 0 0 1-.22.3.59.59 0 0 1-.4.14H9l4.55-11.47h1.89zm-3.53-4.31L14.89 9.5a11.62 11.62 0 0 1-.39-1.24q-.09.37-.19.69l-.19.56-1.58 4.19zm-6.3-1.58a13.43 13.43 0 0 1-2.91-1.41 11.46 11.46 0 0 0 2.81-5.37H12V4H7.31a4 4 0 0 0-.2-.56C6.87 2.79 6.6 2 6.6 2l-1.47.5s.4.89.6 1.5H0v1.33h2.15A11.23 11.23 0 0 0 5 10.7a17.19 17.19 0 0 1-5 2.1q.56.82.87 1.38a23.28 23.28 0 0 0 5.22-2.51 15.64 15.64 0 0 0 3.56 1.77zM3.63 5.33h4.91a8.11 8.11 0 0 1-2.45 4.45 9.11 9.11 0 0 1-2.46-4.45z" />
+  </svg>
+  <% unless disabled %>
+    <select role="button" class="p-0 position-absolute top-0 start-0 w-100 h-100 language-change-trigger text-transparent bg-transparent <%= classes %>">
+      <% Language.order(:english_name).each do |language| %>
+        <option class="form-select" value="<%= language.code %>" <%= "selected" if request.cookies["_osm_locale"] == language.code %>><%= language.name %></option>
```
AntonKhorev, line 8: Shouldn't the comparison here be with `I18n.locale.to_s` instead of `request.cookies["_osm_locale"]`? What if the cookie is not yet set, which option is going to be selected?

app/views/layouts/_header.html.erb
nertc: I updated it to be consistent for both situations.

app/views/shared/_language_selector.html.erb
nertc: If there was not a cookie, none of the options would be preselected. But for the better UX this code was updated to use `I18n.locale.to_s` and show preselected language even if it is set not by the cookie but browser.

AntonKhorev:
Looks like the languages you get from `Language` don't quite match the ones expected by `Locale.list`. For example, we have `pt-BR` and `pt` in `Language` but `pt` and `pt-PT` in translations. As a result you can't select European Portuguese.

app/views/shared/_language_selector.html.erb
```diff
@@ -0,0 +1,16 @@
+<label class="position-relative d-flex <%= classes %>" role="button">
+  <svg width="20" height="20" fill="currentColor">
+    <path d="M20 18h-1.44a.61.61 0 0 1-.4-.12.81.81 0 0 1-.23-.31L17 15h-5l-1 2.54a.77.77 0 0 1-.22.3.59.59 0 0 1-.4.14H9l4.55-11.47h1.89zm-3.53-4.31L14.89 9.5a11.62 11.62 0 0 1-.39-1.24q-.09.37-.19.69l-.19.56-1.58 4.19zm-6.3-1.58a13.43 13.43 0 0 1-2.91-1.41 11.46 11.46 0 0 0 2.81-5.37H12V4H7.31a4 4 0 0 0-.2-.56C6.87 2.79 6.6 2 6.6 2l-1.47.5s.4.89.6 1.5H0v1.33h2.15A11.23 11.23 0 0 0 5 10.7a17.19 17.19 0 0 1-5 2.1q.56.82.87 1.38a23.28 23.28 0 0 0 5.22-2.51 15.64 15.64 0 0 0 3.56 1.77zM3.63 5.33h4.91a8.11 8.11 0 0 1-2.45 4.45 9.11 9.11 0 0 1-2.46-4.45z" />
+  </svg>
+  <% unless disabled %>
+    <select role="button" class="p-0 position-absolute top-0 start-0 w-100 h-100 language-change-trigger text-transparent bg-transparent <%= classes %>">
+      <% Locale.available
+               .map { |locale| Language.find_by(:code => locale.to_s) }
```
AntonKhorev, line 8: You'll get a ton of db queries:

```
  Language Load (0.2ms)  SELECT "languages".* FROM "languages" WHERE "languages"."code" = $1 LIMIT $2  [["code", "af"], ["LIMIT", 1]]
  ↳ app/views/shared/_language_selector.html.erb:8
  Language Load (0.3ms)  SELECT "languages".* FROM "languages" WHERE "languages"."code" = $1 LIMIT $2  [["code", "ar"], ["LIMIT", 1]]
  ↳ app/views/shared/_language_selector.html.erb:8
  Language Load (0.3ms)  SELECT "languages".* FROM "languages" WHERE "languages"."code" = $1 LIMIT $2  [["code", "az"], ["LIMIT", 1]]
  ↳ app/views/shared/_language_selector.html.erb:8
  Language Load (0.3ms)  SELECT "languages".* FROM "languages" WHERE "languages"."code" = $1 LIMIT $2  [["code", "be"], ["LIMIT", 1]]
  ↳ app/views/shared/_language_selector.html.erb:8
  Language Load (0.3ms)  SELECT "languages".* FROM "languages" WHERE "languages"."code" = $1 LIMIT $2  [["code", "bg"], ["LIMIT", 1]]
  ↳ app/views/shared/_language_selector.html.erb:8
...
```

app/views/shared/_language_selector.html.erb
```diff
@@ -0,0 +1,16 @@
...
+               .select { |locale| locale }
+               .sort_by { |locale| locale[:english_name] }
+               .each do |language| %>
+        <option class="form-select" value="<%= language.code %>" <%= "selected" if I18n.locale.to_s == language.code %>><%= language.name %></option>
```
AntonKhorev, line 12: Using Language table is probably a bad idea. I tried this:

```suggestion
      <% Locale.available
               .select { |locale| I18n.exists? "shared.language_selector.language", :locale => locale, :fallback => false }
               .sort_by { |locale| locale.to_s == "en" ? "English" : t(".language", :locale => locale) }
               .each do |locale| %>
        <option class="form-select" value="<%= locale.to_s %>" <%= "selected" if I18n.locale.to_s == locale.to_s %>><%= locale.to_s == "en" ? "English" : t(".language", :locale => locale) %></option>
```

with this in locale file:

```
  share:
    language_selector:
      language: THIS LANGUAGE NAME
```

Now translators have to add whatever is the appropriate representation of their language there, and then it will appear in the selector.

nertc: Pushed code can be shortened as `Language.where(:code => Locale.available.map {|locale| locale.to_s})`. This approach will also cause only one DB query. If there is no other reason for language not to be used, I'll push the code with this approach. Translating language names using i18n and not having one source for them, may cause duplicate translations or inconsistency (because of multiple functionalities using language names).

AntonKhorev: > Translating language names using i18n and not having one source for them

i18n is going to be their *one source*

> may cause duplicate translations or inconsistency (because of multiple functionalities using language names).

We already have two different functionalities using two different sets of languages: i18n and languages of diary entries.

nertc:
This PR was updated according to the suggestion of @AntonKhorev.

As described in the `CONTRIBUTING.md` only updated language config was `en`:
> ## i18n
>
> If you make a change that involve the locale files (in `config/locales`) then please only submit changes to the `en.yml` file. The other files are updated via [Translatewiki](https://translatewiki.net/wiki/Translating:OpenStreetMap) and should not be included in your pull request.

Because of this, currently only English is shown in the list of language selector. At first, I thought to make a list of all language names in the config file, but when translations will be made, there will be a lot of duplication. The only downside of the current solution is that there will be only those languages shown, which have `shared.language_selector.language` key.

app/assets/javascripts/language_selector.js
nertc: Code was updated according to the recommendation.

AntonKhorev:
> The only downside of the current solution is that there will be only those languages shown, which have shared.language_selector.language key.

Those that have translations that are maintained, so not much of a downside.

I also had a more defensive version of that key with `language: THIS LANGUAGE NAME` but maybe it's not required. It's possible to document strings on translatewiki like this
https://translatewiki.net/wiki/Osm:Site.copyright.native.native_link/qqq

nertc:
Is there anything that should be fixed in this PR?

AntonKhorev:
Depends on how much @gravitystorm is against settings in cookies https://github.com/openstreetmap/openstreetmap-website/pull/4985#issuecomment-2551576536.

nertc:
PR was updated. `master` was rebased as a base for this PR. Conflicts were resolved. There were some changes related to the header in the `master`, so logic and HTML of this PR was changed accordingly to fix problems related to locale selector, which were raised after merging. Also, `.not-collapsible` class was added to the locale selector and, in the future, if any other fixed not-collapsible items will be added to the header menu, adding that class to them will automatically make them not-collapsible.

app/assets/stylesheets/common.scss
```diff
@@ -52,6 +52,13 @@ time[title] {
   color: $blue;
 }
 
+/* Utility for transparent color */
+
+.text-transparent{
```
tomhughes, line 57: Space before the brace please.

app/views/shared/_language_selector.html.erb
```diff
@@ -0,0 +1,15 @@
+<label class="position-relative d-flex <%= classes %>" role="button">
+  <svg width="20" height="20" fill="currentColor">
+    <path d="M20 18h-1.44a.61.61 0 0 1-.4-.12.81.81 0 0 1-.23-.31L17 15h-5l-1 2.54a.77.77 0 0 1-.22.3.59.59 0 0 1-.4.14H9l4.55-11.47h1.89zm-3.53-4.31L14.89 9.5a11.62 11.62 0 0 1-.39-1.24q-.09.37-.19.69l-.19.56-1.58 4.19zm-6.3-1.58a13.43 13.43 0 0 1-2.91-1.41 11.46 11.46 0 0 0 2.81-5.37H12V4H7.31a4 4 0 0 0-.2-.56C6.87 2.79 6.6 2 6.6 2l-1.47.5s.4.89.6 1.5H0v1.33h2.15A11.23 11.23 0 0 0 5 10.7a17.19 17.19 0 0 1-5 2.1q.56.82.87 1.38a23.28 23.28 0 0 0 5.22-2.51 15.64 15.64 0 0 0 3.56 1.77zM3.63 5.33h4.91a8.11 8.11 0 0 1-2.45 4.45 9.11 9.11 0 0 1-2.46-4.45z" />
+  </svg>
+  <% unless disabled %>
+    <select role="button" class="p-0 position-absolute top-0 start-0 w-100 h-100 language-change-trigger text-transparent bg-transparent <%= classes %>">
```
tomhughes, line 6: I don't really like the whole thing about overlaying an SVG on a transparent select to be honest but I don't have a good alternative either.

config/locales/en.yml
```diff
@@ -1984,6 +1984,8 @@ en:
       users:
         older: Older Users
         newer: Newer Users
+    language_selector:
+      language: English
```
tomhughes, line 1988: I think this is very problematic... We're going to have translators translating this to the localised version of "English" rather than the name of their own language.

Also until it is translated languages just don't appear, and we start with a menu that only has english on it and hence looks really silly.

Maybe this should have one key for each language (so `languages.en`, `languages.fr` etc) with the english version and then people can translate them all even if many of them won't immediately be used and we can either show both versions with one in parentheses in the menu, or we can show the localised version with fallback to english.

tomhughes (changes requested):

config/locales/en.yml
AntonKhorev: I proposed making it obvious that "English" shouldn't be translated by not including "English" in translations:
https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r1804852184

nertc:
PR was updated. `master` was rebased as a base for this PR.
@tomhughes Code was changed according to comments.

app/assets/stylesheets/common.scss
nertc: Space was added.

config/locales/en.yml
nertc: Language codes were added in `en.yml`. Now even before translation of these new keys, all languages will appear in selector.

app/views/shared/_language_selector.html.erb
nertc: For now, I haven't changed anything related to this comment. If you prefer any of these or other solutions, I'll update PR accordingly.

nertc:
Thanks for rerunning tests.
PR was updated. `master` was merged to avoid previous test errors.

app/views/shared/_language_selector.html.erb
```diff
@@ -0,0 +1,15 @@
...
+      <% Locale.available
+               .select { |locale| I18n.exists? "shared.language_selector.#{locale}" }
```
AntonKhorev, line 8: This filters out some languages because of case sensitivity:
- be-tarask
- kk-Cyrl
- skr-Arab

app/views/shared/_language_selector.html.erb
```diff
@@ -0,0 +1,15 @@
...
+               .sort_by { |locale| t(".#{locale}") }
+               .each do |locale| %>
+        <option class="form-select" value="<%= locale.to_s %>" <%= "selected" if I18n.locale.to_s == locale.to_s %>><%= t(".#{locale}") %></option>
```
AntonKhorev, line 11: `t(".#{locale}")` - this is writing language names in the current locale. If anyone switches to some writing system you don't know, which is a likely scenario in #1115, good luck finding your language.

Here's how it looks if I translate some lanuages to Russian, with English highlighted:
image

config/locales/en.yml
```diff
@@ -2032,6 +2032,114 @@ en:
...
+      hsb: Sorbian, Upper
+      hu: Hungarian
+      ia: Interlingua
+      id: Indonesian
+      is: Icelandic
+      it: Italian
+      ja: Japanese
+      ka: Georgian
+      kab: Kabyle
+      kk-cyrl: Kazakh (Cyrillic)
```
AntonKhorev, line 2088: If we are to use this approach (ask translators to do n x n language name translations), the keys here have to be whatever `I18n::Locale::Tag::Rfc4646` normalizes them to, see https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2084730324.

This is the change for Kazakh:
```suggestion
      kk-Cyrl: Kazakh (Cyrillic)
```

app/views/shared/_language_selector.html.erb
```diff
@@ -0,0 +1,15 @@
...
+      <% end %>
+    </select>
```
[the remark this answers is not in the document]
AntonKhorev: You'll get a dozen of exceptions where a language name x is not available in the language x.

Meanwhile we have all the names in the first lines of `config/locales/*.yml`, except for `en`.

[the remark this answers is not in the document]
1ec5: [`Intl.DisplayNames.prototype.of()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/of) returns undefined if the language code is unrecognized, so we could fall back to the raw code. It only throws an exception on a malformed language code. But if we already have a string that we can access at build time, that would be good too. In that case, we wouldn’t need all those added lines in the YAML file, one for each language…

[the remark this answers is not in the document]
1ec5: https://translatewiki.net/w/i.php?title=Special:Translations&message=Osm%3ASite.copyright.native.native+link shows that most localizations are correctly translating into their own language, though a few have `THIS_LANGUAGE_NAME_HERE` untranslated.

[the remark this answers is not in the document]
AntonKhorev: After #6024 this can be replaced with 

```suggestion
    <%= select_tag "language",
                   options_for_select(AVAILABLE_LANGUAGES.map { |e| [e[:native_name], e[:code]] },
                                      I18n.locale),
                   :role => "button",
                   :class => "p-0 position-absolute top-0 start-0 w-100 h-100 language-change-trigger text-transparent bg-transparent #{classes}" %>
```

https://github.com/AntonKhorev/openstreetmap-website/tree/a/pull/5201

app/views/shared/_language_selector.html.erb
AntonKhorev: can be resolved with https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2121255541

app/views/shared/_language_selector.html.erb
AntonKhorev: can be resolved with https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2121255541

config/locales/en.yml
AntonKhorev: can be resolved with https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2121255541 by not going with this approach

nertc:
PR was updated. `master` was merged.

@AntonKhorev Thank you for your PR. `Locale selector` was changed accordingly.

app/views/shared/_language_selector.html.erb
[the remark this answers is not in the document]
nertc: `Locale selector` was changed according to https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2121255541

config/locales/en.yml
nertc: Locale selector was changed according to https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2121255541

app/views/shared/_language_selector.html.erb
nertc: Locale selector was changed according to https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2121255541

app/views/shared/_language_selector.html.erb
nertc: Locale selector was changed according to https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2121255541

tomhughes:
I think this is fine as a starting point and we should merge it - we can always look to improve things further later.

Thanks for all the work on it.

PR #5202, opened 2024-09-12
Add subway, tram, bus stops to mapnik map key

AntonKhorev:
image


tomhughes:
Looks good to me, thanks.

PR #5203, opened 2024-09-12
Displays links for friend options on user profile only if user is signed in

### LINKED ISSUE #5199: buttons on userpages visible after logging out
On a userpage
https://www.openstreetmap.org/user/test?locale=en
we have buttons like
`Send Message` | `Add Friend` | `Report this User` | `Mute this User`
and on the same userpage after logging out from above ones we can't see
`Report this User` and `Mute this User` anymore
but we still can see `Send Message` and `Add Friend`.

I think these two actions shouldn't be visible either, because they require logging in.

What do you think?

nenad-vujicic:
PR proposes displaying "Send Message", "Add Friend" and "Unfriend" links (buttons) on user profile page only if user is signed in. Previously, these links (buttons) were always displayed.

Fixes #5199 

Signed In:
image

Signed Out:
image

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -81,15 +81,15 @@
               <%= link_to t(".comments"), diary_comments_path(@user) %>
               <span class='badge count-number'><%= number_with_delimiter(@user.diary_comments.size) %></span>
             </li>
-            <li>
-              <% if current_user and current_user.friends_with?(@user) %>
-                <%= link_to t(".remove as friend"), remove_friend_path(:display_name => @user.display_name), :method => :post %>
-              <% elsif current_user %>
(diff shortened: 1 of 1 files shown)


tomhughes (changes requested):
I think we should definitely remove add/remove friend as they can't be correctly rendered without knowing the current friendship state but I think we should probably leave send message as it's reasonable to offer to everybody, even those that would have to signup to send a message.

kcne (approved):

nenad-vujicic:
> I think we should definitely remove add/remove friend as they can't be correctly rendered without knowing the current friendship state but I think we should probably leave send message as it's reasonable to offer to everybody, even those that would have to signup to send a message.

Thanks, updated. Here are the latest screenshots:

Logged-in:
image

Logged-out:
image

tomhughes (approved):
Thanks - that looks good now.

PR #5204, opened 2024-09-12
Fix friendships limit flash message

AntonKhorev:
If you try to add too many friends, you're supposed to see this message:
image

But you won't see it because of `flash.now` followed by a redirect.


tomhughes:
Looks good to me, thanks.

PR #5205, opened 2024-09-12
Remove classic pagination

AntonKhorev:
This removes classic pagination library:
- necessary parts of its `Paginator` class moved to controller's `ElementsPaginator`
- necessary parts of `PaginationHelper` module moved to `sidebar_classic_pagination` helper


tomhughes:
Shouldn't we just be converting changesets to use the same pagination logic as everything else, rather than copying parts of classic pagination into the controller just so we can pretend we got rid of it?

AntonKhorev:
> Shouldn't we just be converting changesets to use the same pagination logic as everything else

That depends on how significant are the differences between changeset elements and everything else. The differences are:

- Changesets have an upper bound on the number of elements. There can be a really large and growing number of traces etc. If you wanted to use numbered pages and go to page N, that N could be huge and you'll have to skip a huge number of record with `offset`. But for changeset elements N can't get higher than 10000 / page size.
- Changesets don't change once closed. For a list of traces it doesn't make mach sense to be on page 1, 2, 5 etc because new traces are added and old ones are pushed off the page. Changeset elements are not going to be pushed of their page.

gravitystorm:
> Shouldn't we just be converting changesets to use the same pagination logic as everything else, rather than copying parts of classic pagination into the controller just so we can pretend we got rid of it?

I agree.

> * Changesets have an upper bound on the number of elements. There can be a really large and growing number of traces etc. If you wanted to use numbered pages and go to page N, that N could be huge and you'll have to skip a huge number of record with `offset`. But for changeset elements N can't get higher than 10000 / page size.
> 
> * Changesets don't change once closed. For a list of traces it doesn't make mach sense to be on page 1, 2, 5 etc because new traces are added and old ones are pushed off the page. Changeset elements are not going to be pushed of their page.

These are good reasons why the paginator **doesn't have to be** cursor-based, but they aren't reasons why it **can't** be cursor-based. So I think the simplification of only having one type of pagination is worthwhile.

> * Traces etc have a well-defined order represented by a single id, changeset elements don't.

Since we need a well-defined order for either approach to work, I think it's best to get the order defined and working with the cursor-based pagination.

AntonKhorev:
> These are good reasons why the paginator doesn't have to be cursor-based, but they aren't reasons why it can't be cursor-based. So I think the simplification of only having one type of pagination is worthwhile.

Do you propose to remove the ability to go to page N?

gravitystorm:
> Do you propose to remove the ability to go to page N?

Sure, I don't see a strong need for that. Particularly if the order hasn't been defined before.

AntonKhorev:
Do you also propose to remove the ability to see how many elements of each type were changed?

gravitystorm:
No.

AntonKhorev:
Superseded by #6299.

PR #5206, opened 2024-09-12
Make select all checkbox on users list work with turbo

tomhughes:
This reworks the event handling to allow the "select all" checkbox to work after changing page in the user list.


AntonKhorev:
Merged, thanks.

PR #5207, opened 2024-09-12
Backfill the creation_address column from creation_ip

tomhughes:


db/migrate/20240912181413_backfill_user_creation_address.rb
```diff
@@ -0,0 +1,11 @@
+class BackfillUserCreationAddress < ActiveRecord::Migration[7.1]
+  def up
+    User
```
AntonKhorev, line 3: If you refer to a model from within a migration, there's a possibility that at some point in the future this model won't exist but you still want the migration to run. You can declare that model inside the migration, then the migration will work. Looks like we've been doing this for a while:

https://github.com/openstreetmap/openstreetmap-website/blob/b81d645f0c6a78b944708005f188ba1a5ab6f4e7/db/migrate/20180204153242_tile_users.rb#L2-L3

tomhughes: I've made that change now.

AntonKhorev:
Merged, thanks.

PR #5208, opened 2024-09-12
Removed unused translations from activerecord.errors and updated i18n-tasks.yml

nenad-vujicic:
PR removes unused translations activerecord.errors.messages.{invalid_email_address, email_address_not_routable} started / ended using in 63862d0 / 90a7efc and updated i18n-tasks.yml to ignore reporting auth.providers.* as unused although they are used in app/views/application/_auth_providers.html.erb, ..


tomhughes:
Looks good to me, thanks.

Issue #4660, opened 2024-04-08
Global element versioning


AntonKhorev:
### Problem

Sometimes the code tries to reproduce the sequence of element modifications but the data to do this is not directly available.

Changeset downloads rely primarily on timestamps, and then additionally sort by version, type etc and it's done differently in [osm-website](https://github.com/openstreetmap/openstreetmap-website/blob/0c4cbda662502f05646a4e82f5d8f639b9183059/app/controllers/api/changesets_controller.rb#L158) and [cgimap](https://github.com/zerebubuth/openstreetmap-cgimap/blob/2fa2d702017ddd136555e3562f574ac91870e8bc/src/osmchange_responder.cpp#L32). This additional ordering may not be correct. For example, it makes sense to reverse ordering by type for deletions. Relying on the additional ordering becomes necessary if timestamps are truncated, and that may happen if the data is restored from a dump.

Changeset webpages show lists elements. Those also reproduce the modification sequence in most of the cases but that happens as a side-effect of them not being sorted. This lack of sorting causes problems elsewhere and so does assuming that the elements are not sorted, see https://github.com/openstreetmap/openstreetmap-website/pull/4571.

### Description

Is it feasible to do this thing?
https://github.com/openstreetmap/openstreetmap-website/blob/0c4cbda662502f05646a4e82f5d8f639b9183059/app/controllers/api/changesets_controller.rb#L154-L157

1. Create one sequence to serve three old element tables.
2. Start it at a large enough number.
3. Add a column to each old element table with a default value taken from this sequence. This should work with db clients that don't yet know about this column but insert elements.
4. Backfill the column according to [this order](https://github.com/zerebubuth/openstreetmap-cgimap/blob/2fa2d702017ddd136555e3562f574ac91870e8bc/src/osmchange_responder.cpp#L32). It's not necessarily correct, maybe update it later.
5. Use this global version column for ordering (old) elements inside changesets on osm website and in other places.

mmd-osm:
> Backfill the column according to [this order](https://github.com/zerebubuth/openstreetmap-cgimap/blob/2fa2d702017ddd136555e3562f574ac91870e8bc/src/osmchange_responder.cpp#L32). It's not necessarily correct, maybe update it later.

I'm not sure how feasible this is from a total runtime point of view. In any case, mass updating old element tables is probably bad news for our minutely diff replication, since it would end up generating billions of entries in our replication log files.

[...]

PR #5209, opened 2024-09-12
Order changeset elements for consistent pagination

tomhughes:
Changeset elements in the history view are being paginated but no order is being specified so stepping through the pages may not give consistent results and may skip elements or show them twice - this is currently visible in the form of occasional test failures in `test_show_paginated_element_links` when it doesn't show elements in the order `create_list` created them.

### DIFF
--- a/app/controllers/changesets_controller.rb
+++ b/app/controllers/changesets_controller.rb
@@ -79,13 +79,13 @@ class ChangesetsController < ApplicationController
     @changeset = Changeset.find(params[:id])
     case turbo_frame_request_id
     when "changeset_nodes"
-      @node_pages, @nodes = paginate(:old_nodes, :conditions => { :changeset_id => @changeset.id }, :per_page => 20, :parameter => "node_page")
+      @node_pages, @nodes = paginate(:old_nodes, :conditions => { :changeset_id => @changeset.id }, :order => [:node_id, :version], :per_page => 20, :parameter => "node_page")
       render :partial => "elements", :locals => { :type => "node", :elements => @nodes, :pages => @node_pages }
     when "changeset_ways"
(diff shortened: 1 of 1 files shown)


kcne (approved):

AntonKhorev:
This was suggested in https://github.com/openstreetmap/openstreetmap-website/pull/4571#issuecomment-1994856482 but I haven't done it because we don't have appropriate indexes. This requires reading all changeset node/ways/relations and sorting them. We also have counts of changeset elements but at least counting doesn't require sorting.

Also if you want the pagination here to be closer to like it is for traces, diary entries etc (https://github.com/openstreetmap/openstreetmap-website/pull/5205#issuecomment-2346879691), that's done with sorting by some id. We can't sort changeset elements just by id because ids are not unique, you can have several versions of the same element. That's why this PR sorts by (id, version) but it's not quite the same as in other places. What would have made it nearly the same is something like #4660.

tomhughes:
Well what I mostly want in the short term is for the tests to be reliable, but the reason they're not reliable is that the web site is not reliable - next and previous may not step through the objects reliably.

As we're paginating this data with a full count of pages we're already reading it all anyway so having to sort it doesn't change much.

I don't care much what the order is so long as it's stable which I'm not sure timestamp+version is (though adding id as a third element would guarantee stability).

mmd-osm:
Timestamp + version is not stable. cgimap does mass inserts/updates, where elements have the exact same timestamp down to the µs.

AntonKhorev:
> we're already reading it all anyway so having to sort it doesn't change much

I can't check this because I don't know what's the distribution of requested changesets and their timings. Let's merge this and see. Maybe everything is going to work fast enough, then we don't need to bother with extra indexes.

PR #5211, opened 2024-09-13
Fix element pagination width calculation

### LINKED ISSUE #5105: sidebar_classic_pagination - width calculation not clear
I've noticed that in some cases, pagination cells are very narrow, in particular when we're showing two disabled cells ("..."), like in example 1 below.

To calculate a threshold value, we're currently applying the following logic (introduced in  cfa1ce8617430cdec47409a455ea2aead713334e):

```ruby
      width += body.length
```

I'm wondering why we're not taking the actual length of the strings we're showing on the screen into account, rather than the number of array elements (that includes some technical details). So instead of the previous calculation, that would be:

```ruby
      width += body.first.length
```

### Example 1

image

Debug output:

```
====
["1", 1] --- 2
["...", "disabled"] --- 2
["5", 5] --- 2
["6", 6] --- 2
["7", "active"] --- 2
["8", 8] --- 2
["9", 9] --- 2
["...", "disabled"] --- 2
["11", 11] --- 2
====
36
====
```

### Example 2

image

Debug output:

```
====
["1", 1] --- 2
["2", 2] --- 2
["3", "active"] --- 2
["4", 4] --- 2
["5", 5] --- 2
["...", "disabled"] --- 2
["11", 11] --- 2
====
28
====
```

Example in production: https://www.openstreetmap.org/changeset/155575172?node_page=5

### Debug code

```ruby
  def sidebar_classic_pagination(pages, page_param)
    max_width_for_default_padding = 35

    puts "===="
    width = 0
    pagination_items(pages, {}).each do |body|
      width += 2 # padding width
      width += body.length
      puts "#{body} --- #{body.length}"
    end
    puts "===="
    puts width
    puts "===="
    link_classes = ["page-link", { "px-1" => width > max_width_for_default_padding }]
```

AntonKhorev:
Fixes #5105, see https://github.com/openstreetmap/openstreetmap-website/issues/5105#issuecomment-2349260230.

`body` in `|body|` gets assigned the entire array, while `body` in `|(body)|` gets the first element, which is the intended effect.

### DIFF
--- a/app/helpers/browse_helper.rb
+++ b/app/helpers/browse_helper.rb
@@ -86,7 +86,7 @@ module BrowseHelper
     max_width_for_default_padding = 35
 
     width = 0
-    pagination_items(pages, {}).each do |body|
+    pagination_items(pages, {}).each do |(body)|
       width += 2 # padding width
       width += body.length
     end


app/helpers/browse_helper.rb
```diff
@@ -86,7 +86,7 @@ def sidebar_classic_pagination(pages, page_param)
     max_width_for_default_padding = 35
 
     width = 0
-    pagination_items(pages, {}).each do |body|
+    pagination_items(pages, {}).each do |(body)|
```
tomhughes, line 89: Just use `|body, _|` instead if you want to destructure it without an unused variable warning.

tomhughes (changes requested):

app/helpers/browse_helper.rb
AntonKhorev: Rubocop is going to complain then that I'm probably iterating over a hash and that I should use each_key instead.

mmd-osm: Body is an array, like `["3", "active"]`, so applying `each_key` here might be a bit confusing (why is the first array element a key?). On the other hand, I'm not clear how `(body)` works.

In my proposal I left this part as is, and opted to access the array by element index instead:
`width += body.first.length` or `width += body[0].length`

By the way, I was wondering a bit why pagination_items returns an array of arrays, instead an array of hashes, where the purpose of each attribute would be a bit clearer. That's maybe something for another issue.

tomhughes: Using `each_key` won't work anyway as it's only defined on hashes. Dropping the `each` and making `pagination_items` yield the items directly instead of building a list and returning it does then allow the syntax I wanted without warnings but as you say maybe these should be hashes if that allows us to get rid of the `page_or_class` duality thing.

tomhughes:
I'm going to merge this for now to fix things but as discussed I think there is scope for future improvements here to get rid of this weird syntax.

PR #5212, opened 2024-09-13
Drop creation_ip column

tomhughes:
This is the next step now that `creation_address` has been backfilled...


app/models/user.rb
```diff
@@ -50,6 +49,8 @@ class User < ApplicationRecord
   require "digest"
   include AASM
 
+  self.ignored_columns += ["creation_ip"]
```
AntonKhorev, line 52: What `strong_migrations` wants you to do is

1. stop using the column and start ignoring it like in #2432
2. remove the column like in #2439
3. stop ignoring the column like in #2461 because it doesn't exist

Here instead you removing the column and then you start ignoring it. What is ignoring supposed to achieve if the column already doesn't exist?

(And if you don't care about `strong_migrations` you also wouldn't ignore the column.)

tomhughes: Well I don't really care about `strong_migrations` but I can't ignore it so I need to shut it up... Doing this project in the officially sanctioned way would have needed something like six PRs :-( My plan was just to make sure this was deployed to the web servers before running the migrations (they always run manually) but I can split it up if you prefer.

tomhughes:
I've opened #5218 for the first phase and rebased this on that and switched it to draft for now.

tomhughes:
This is now ready for review again, as #5218 has been merged and deployed.

PR #5213, opened 2024-09-14
Always show changeset element page links below headings

AntonKhorev:
What I tried to achieve previously was to save space. If the page links fit to the right of the heading, let them be there.

But this works fine only with full sidebar reloads. With turbo frame reloads it's more noticeable when the links become wider, don't fit and move below the heading. Here I'm making them always be below.

Before:

image
image

After:

image
image


kcne:
In my opinion this looks more coherent anyway. Thank you.

kcne (approved):

tomhughes:
Looks good to me, thanks.

PR #5214, opened 2024-09-14
Fix browse elements timeouts

AntonKhorev:
Like #5178, was part of https://github.com/openstreetmap/openstreetmap-website/pull/4617.

Shows messages like this on element, element version and element history page timeouts:
image

Currently those are failing with "Missing template .../timeout".


tomhughes:
Looks good to me, thanks.

PR #5218, opened 2024-09-15
Replace creation_ip with creation_address

tomhughes:
This replaces `creation_ip` with `creation_address` and ignores `creation_ip` in preparation for removing it.


AntonKhorev:
Merged, thanks.

PR #5220, opened 2024-09-17
Remove unnecessary comma from several translatable messages

amire80:


amire80:
Yes, but it's still unnecessary :)

"data" or "history" is the subject (подлежащее), "took" is the verb (сказуемое), and there's no need for a comma between a subject and a verb.

The old message with the unnecessary comma was probably copied to the recently-added ones, and it should be removed from all of them.

tomhughes:
I agree, that comma doesn't look right to me. Thanks for the fix.

PR #5221, opened 2024-09-17
Added greeting to friendship notification

nenad-vujicic:
PR adds greeting 'Hi {user display name},' to friendship notification messages (both HTML and TEXT). As a side effect, user_mailer.friendship_notification.hi is not anymore reported as unused.


tomhughes:
Looks good to me, thanks.

PR #5223, opened 2024-09-18
Removed unused translations in changesets namespace

nenad-vujicic:
PR makes following changes:

1. Updates i18n-tasks.yml to ignore unused translations for keys from changesets.paging_nav.* (because they are used in pagination mechanism using string interpolation),
2. Pulls out ?: operator from t() in changesets/show.html.erb so changesets.show.{comment_by_html, hidden_comment_by_html} translation keys are no longer reported as unused and
3. Removes unused translation keys changesets.{changeset.anonymous, no_such_entry.title} from en.yml because .anonymous was added in fddda4e and was stopped being used in 3e305fb and .title was added in f0764d3 but never used.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -441,7 +441,6 @@ en:
         sorry: "Sorry, the list of changeset comments you requested took too long to retrieve."
   changesets:
     changeset:
-      anonymous: "Anonymous"
       no_edits: "(no edits)"
       view_changeset_details: "View changeset details"
     index:
@@ -473,7 +472,6 @@ en:
       title: "Changeset %{id}"
       created_by_html: "Created by %{link_user} on %{created}."
     no_such_entry:
-      title: "No such changeset"
       heading: "No entry with the id: %{id}"
       body: "Sorry, there is no changeset with the id %{id}. Please check your spelling, or maybe the link you clicked is wrong."
     show:


app/views/changesets/show.html.erb
```diff
@@ -40,9 +40,10 @@
         <% next unless comment.visible || current_user&.moderator? %>
         <li id="c<%= comment.id %>">
           <small class='text-body-secondary'>
-            <%= t comment.visible ? ".comment_by_html" : ".hidden_comment_by_html",
+            <% comment_info = comment.visible ? "comment_by_html" : "hidden_comment_by_html" %>
+            <%= t(".#{comment_info}",
```
AntonKhorev, line 44: Looks like once you do `t ".#{comment_info}"` i18n-tasks stops reporting `changesets.show.`-anything as missing.

You could have inserted this line with the same effect:

    <% t ".#{'shut up'}" %>

config/i18n-tasks.yml
```diff
@@ -132,6 +132,7 @@ ignore_unused:
   - 'reports.new.categories.*' # double interpolation in reports_helper
   - 'shared.pagination.*'
   - 'auth.providers.*'
+  - 'changesets.paging_nav.*' # used by string interpolation in elements pagination
```
AntonKhorev, line 135: Won't be required if #4872 is merged, although it uses the same effect as in https://github.com/openstreetmap/openstreetmap-website/pull/5223/files#r1766713701 to some extent.

nenad-vujicic:
Thanks, I've just made following updates:

1. Removed 1 in favor of #5225 
2. Removed 2 in favor of #4872 
3. Rebased

PR #5225, opened 2024-09-19
Don't use ?: in "Comment from user" changeset discussion strings

AntonKhorev:
Does what https://github.com/openstreetmap/openstreetmap-website/pull/5223/commits/4564073106d5d48f3781a207693c39a13c130b66 did but without triggering extra ignores by i18n-tasks.


tomhughes:
Looks good to me, thanks.

PR #5231, opened 2024-09-21
OCI: ignore invalid replacement token

mmd-osm:
In #5210, updating the OSM community index failed due to invalid replacement tokens. This issue needs to be resolved upstream. However, we can add a bit of error handling and still apply the update. EN is used for erroneous translations.


tomhughes:
Looks good to me, thanks.

PR #5232, opened 2024-09-21
Trace import: validate background job results

mmd-osm:
While the discussion in #5038 is still ongoing, I have moved the unit test to a new PR.
Validating the results of trace import background job is useful for both scenarios


tomhughes:
Looks good to me, thanks.

PR #5235, opened 2024-09-24
Removed unused translation keys from diary_entries

nenad-vujicic:
PR removes diary_entries.location.{edit,view} unused translation keys. They were added at 0b913ef and stopped being used at 53fc3e8 / aa121fa.


tomhughes:
Looks good to me, thanks.

PR #5236, opened 2024-09-24
Improved unused translation keys in users

nenad-vujicic:
PR removes users.new.tou (added 9649b19 and has never been used, used layouts.tou instead), replaces users.terms.decline with users.terms.cancel and together with users.terms.continue starts being used at /users/terms.html.erb (where both were added at 9a9e13a and removed at 1be8d59, probably forgotten to be re-added under t()).


tomhughes:
Looks good to me, thanks.

PR #5241, opened 2024-09-30
Removed unused translations in user_blocks

nenad-vujicic:
PR removes user_blocks.show.{confirm, needs_view, show} translation keys added at daa2496. confirm and show has never been used, while needs_view ended being used at ca06b3c.


tomhughes:
Looks good to me, thanks.

PR #5242, opened 2024-09-30
Masks off unused translation keys in activerecord

nenad-vujicic:
PR replaces .is_already_muted with I18n.t("activerecord.errors.models.user_mute.is_already_muted") in user_mute.rb and adds activerecord.errors.models.user_mute.attributes.subject.format under ignore_unused (in i18n-tasks.yml) since used for formatting above validation error message, so they are not reported as unused anymore.

### DIFF
--- a/config/i18n-tasks.yml
+++ b/config/i18n-tasks.yml
@@ -134,6 +134,8 @@ ignore_unused:
   - 'auth.providers.*'
   - 'time.formats.blog' # used for formatting blog dates
   - 'time.formats.friendly' # used for formatting dates / times in a friendly way
+  - 'activerecord.errors.models.user_mute.attributes.subject.format' # used for formatting error message during validation in user_mute.rb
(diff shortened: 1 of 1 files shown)


tomhughes:
Is this really an improvement though? We're hard coding something that rails does for us automatically in order to keep i18n-tasks happy - is that a good thing?

nenad-vujicic:
I agree, thanks. Honestly, I couldn't decide which one I didn't like more, hardcoding translating or adding to ignore list in i18n-tasks.yml (the latest commit). Is there a better way to solve this, or should this translation remain reported as unused?

mmd-osm:
I was also wondering about this. IMHO the right place to address this issue would be a proper parser in https://github.com/glebm/i18n-tasks, which is able to handle model validation error messages. I didn't find anything in there, maybe we could raise an issue?

tomhughes:

mmd-osm:
As a follow up, this use case is not (yet) supported by i18n-tasks. The referenced GitHub issue has some comments on how this could be implemented.

PR #5243, opened 2024-09-30
Masked off reporting as unused various keys

nenad-vujicic:
This PR:

- adds time.formats.{blog, friendly} translation keys under ignore_unused in i18n-tasks.yml since used on numerous places for formatting dates / times,
- removes site.sidebar.close because started being used at 3a51160 and ended being used at 6adcce4 and
- removes sessions.new.title because started being used at f18baae and has been used only in test/system/issues_test.rb (from which we removed it because tests are always in EN).

Mentioned translation keys are no longer reported as unused.


tomhughes:
Looks good to me, thanks.

PR #5245, opened 2024-10-02
Move api error handling and timeouts to parent class

### LINKED ISSUE #4861: Refactor api_call_* filters
Most controllers have the same set of `around_action` filters, e.g.

```ruby
around_action :api_call_handle_error, :api_call_timeout
```

However, some don't:
* api/users_controller: `around_action :api_call_handle_error` - I think the lack of timeout is an oversight?
* api/user_preferences_controller: `around_action :api_call_handle_error` - again, I think this is an oversight?
* api/changesets_controller: `around_action :api_call_handle_error;  around_action :api_call_timeout, :except => [:upload]` - I can see a logic for removing the timeout for the upload, since it could be quite big (n.b. rails not used in production for this)
* api/traces_controller: `around_action :api_call_handle_error` - I think adding the timeout handling here would be fine, but maybe exempt `:create` to align with changesets#upload ?

I would like to refactor the controllers, similar to #4859 so that these handlers are applied by default, and exceptions are used (e.g. `skip_around_action` only where necessary.

gravitystorm:
Fixes #4861

Since the around_action is defined before authorize_resource is called, the handler needs to pass on the CanCan::AccessDenied exception.

I've added the timeouts where I think they were missing (e.g. UserPreferencesController) but I've kept the exception for changeset#upload and traces#create

### DIFF
--- a/app/controllers/api/capabilities_controller.rb
+++ b/app/controllers/api/capabilities_controller.rb
@@ -5,7 +5,6 @@ module Api
     authorize_resource :class => false
 
     before_action :set_request_formats
-    around_action :api_call_handle_error, :api_call_timeout
 
     # External apps that use the api are able to query the api to find out some
     # parameters of the API. It currently returns:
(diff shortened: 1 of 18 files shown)


tomhughes:
Looks good to me, thanks.

PR #5246, opened 2024-10-02
Use unreleased version of canonical_rails to allow upgrades to rails 7.2.1+

gravitystorm:
canonical-rails compatibility was fixed in https://github.com/jumph4x/canonical-rails/pull/92 but this is not yet released.

This PR allows us to update to rails point releases, by using the version of canonical-rails from their master branch that includes the compatibility fix. We can go back to using released versions when canonical-rails gets its next release.


tomhughes:
...and now there are security alerts on rails 7.2 so I will merge this.

PR #5247, opened 2024-10-02
Switch to rails 7.2 defaults

tomhughes:
There doesn't seem to be anything in the new defaults that is likely to cause us a problem.

* `enqueue_after_transaction_commit` probably gets disabled by delayed_job in production anyway
* `web_image_content_types` our production libvips has webp support
* `validate_migration_timestamps` shouldn't affect us as we don't have future dated migrations
* `postgresql_adapter_decode_dates` shouldn't affect us as we don't use raw queries like that
* `yjit` doesn't affect production as we're not using 3.3 but doesn't seem to cause any problems in my 3.3 environment


gravitystorm:
LGTM too.

PR #5239, opened 2024-09-26
Allow filtering of notes by status (open/closed) on user profile

kcne:
This PR implements a basic filtering functionality for the notes displayed on a user's profile page, addressing part of the long-standing request from issue #832.

#### Key Changes:
- Introduced tabs for filtering notes by status (Open/Closed) on the user profile page.
- When no status is specified in the URL, the **Open Notes** tab is selected and displayed by default.
- The URL is updated with a `status` parameter to reflect the current filter state, ensuring clarity for users and enabling easy sharing/bookmarking of filtered views.
- Maintains existing pagination and ordering logic, ensuring consistency with the current note display behavior.

#### Future Steps:
This is the first step toward more comprehensive sorting and filtering functionalities, such as sorting by creation date, solved date, and more, as requested in #832. By implementing this, we improve the visibility of open (unresolved) notes while preserving the ability to view both open and closed notes for users accustomed to that behavior.

#### Screenshots:
Screenshot 2024-09-26 at 04 46 21

Screenshot 2024-09-26 at 04 46 39

Screenshot 2024-09-26 at 04 46 55

PR #5255, opened 2024-10-09
Add filtering functionality to user notes page

kcne:
This PR addresses  #832 by adding sorting and filtering functionality to user note pages. This PR is also an alternative, or better put expanding #5239 so that filtering and sorting functionality can be expanded further on if needed.

#### Key Changes:
- **Filter by Status**: Users can filter notes by their status (open, closed, or all).
- **Filter by Note Type**: View notes the user submitted or commented on (including those with anonymous comments).
- **Date Range Filtering**: Users can filter notes based on a specified `from` and `to` date range.
- **Sorting Options**: Sort notes by `created_at` or `updated_at` in ascending or descending order.

Tests for added methods and scopes have been added to ensure that these features work as expected across different scenarios, including handling of date ranges and note types.

### Screenshots:
Screenshot 2024-10-08 at 16 47 57
Screenshot 2024-10-08 at 16 48 22
Screenshot 2024-10-08 at 16 48 34


app/views/notes/index.html.erb
```diff
@@ -8,8 +9,37 @@
...
   <h4><%= t ".no_notes" %></h4>
-
 <% else %>
+  <%= form_tag(url_for("controller" => "notes", "action" => "index"), "method" => :get, "data" => { "turbo" => true, "turbo-frame" => "pagination", "turbo-action" => "advance" }) do %>
+    <div class="row gx-2">
+      <div class="col-md-auto mb-3">
+        <%= label_tag :status, t(".status") %>
+        <%= select_tag  :status,
+                        options_for_select([[t(".all"), "all"], [t(".open"), "open"], [t(".closed"), "closed"]], params[:status]),
+                        :include_blank => t(".select_status"),
```
AntonKhorev, line 19: Why do you need `:include_blank`s? You have labels that say what these selects are and the filter acts as if the first options are selected, but you don't show them as selected.

app/views/notes/index.html.erb
```diff
@@ -8,8 +9,37 @@
...
+                        :class => "form-select" %>
+      </div>
+      <div class="col-md-auto mb-3">
+        <%= label_tag :note_type, t(".note_type") %>
```
AntonKhorev, line 23: I wouldn't call this a "note type". It's an interaction type between the user and the note.

And then there's a question if this kind of filtering can be done efficiently because here you have `INNER JOIN "note_comments"` twice.

app/views/notes/index.html.erb
kcne: Thank you for pointing that out. I've remove them and added preselected all fields accordingly.

app/views/notes/index.html.erb
kcne: You're right. I've renamed that into interaction type which is a better naming for this.

kcne:
> About the date pickers:
> 
> I wanted to add them to our standard pagination because we already have date indexes. Then we wouldn't need them here, specifically in this notes filer. But this plan haven't worked so far because:
> 
>     * I wanted to first add a simpler thing to the standard pagination, namely links to first/last pages. I made several versions of this feature - [Add oldest/newest links to shared pagination #4710](https://github.com/openstreetmap/openstreetmap-website/pull/4710) [Add oldest/newest links to shared pagination - with gap between "newer" and "older" #4729](https://github.com/openstreetmap/openstreetmap-website/pull/4729) [Add oldest/newest links to shared pagination - with gaps around "newer" and "older" #4733](https://github.com/openstreetmap/openstreetmap-website/pull/4733) [Add oldest/newest links to shared pagination - with gaps around "newer" and "older" v2 #4734](https://github.com/openstreetmap/openstreetmap-website/pull/4734) - but none of them got much attention from the maintainers. I can still choose to add date pickers, but this will give me one more pull request that's going to be in conflict with the others which I'll have to maintain.
> 
>     * As you can see this notes list doesn't even use the standard pagination. I tried doing that too [Convert note pagination to newer/older note pages #4532](https://github.com/openstreetmap/openstreetmap-website/pull/4532). That wasn't convincing enough because it was too complicated. The complications were there because I used note ids as a cursor but I couldn't sort by them, this wouldn't get notes in the last updated order. The suggested solution by one of the maintainers is to use last comment ids instead, but it's currently being undermined by another maintainer who wants to remove opening comments.

Thank you for the detailed background and context, Anton. I agree that using date pickers makes sense if we’re filtering notes by dates, especially since the current setup isn’t relying on standard pagination. 

As for the next steps, I’m not sure whether it’s better to get this PR merged as is and refactor later when the notes move to standard pagination, or if we should address the pagination issue first and then refactor this PR accordingly.

On a side note, I’ve managed to implement sortable columns for `created_at` and `updated_at` locally, but I’ll hold off on submitting that until this PR is merged to keep things cleaner and easier to review.

kcne:
Thank you for the detailed background and context, Anton. I agree that using date pickers makes sense if we’re filtering notes by dates, especially since the current setup isn’t relying on standard pagination. 

As for the next steps, I’m not sure whether it’s better to get this PR merged as is and refactor later when the notes move to standard pagination, or if we should address the pagination issue first and then refactor this PR accordingly.

On a side note, I’ve managed to implement sortable columns for `created_at` and `updated_at` locally, but I’ll hold off on submitting that until this PR is merged to keep things cleaner and easier to review.

AntonKhorev:
> As for the next steps, I’m not sure whether it’s better to get this PR merged as is and refactor later when the notes move to standard pagination, or if we should address the pagination issue on user note pages first and then refactor this PR accordingly.

You can make a pull request that just adds the status filter. That doesn't depend on pagination and shouldn't tank the performance because:
- status is checked anyway to see if the note is not hidden
- we're doing a similar thing in the api

app/views/notes/index.html.erb
```diff
@@ -6,41 +7,69 @@
...
-<% if @notes.empty? %>
-  <h4><%= t ".no_notes" %></h4>
+<%= form_tag(url_for("controller" => "notes", "action" => "index"), "method" => :get, "data" => { "turbo" => true, "turbo-frame" => "pagination", "turbo-action" => "advance" }) do %>
+  <div class="row gx-2">
+    <div class="col-md-auto mb-3">
+      <%= label_tag :status, t(".status") %>
+      <%= select_tag  :status,
+                      options_for_select([[t(".all"), "all"], [t(".open"), "open"], [t(".closed"), "closed"]], params[:status] || "all"),
+                      :class => "form-select",
+                      :onchange => "this.form.requestSubmit()" %>
```
AntonKhorev, line 17: Not going to work with the `csp_enforce: true` setting.

AntonKhorev:
> On a side note, I’ve managed to implement sortable columns on a table UI end for created_at and updated_at locally

I looked again at why similar things were rejected in the past like #1656 and there doesn't seem to be any clear reason.

AntonKhorev:
Besides the [double-join db query](https://github.com/openstreetmap/openstreetmap-website/pull/5255#discussion_r1801064970), there's couple more issues with the interaction type: ["blue on blue" colors](https://github.com/openstreetmap/openstreetmap-website/pull/4700#issuecomment-2221682594) and whether the submitted/commented classification makes sense at all (what if I closed a note without a comment?). I wanted to try opened/closed/commented instead.

kcne:
> Besides the [double-join db query](https://github.com/openstreetmap/openstreetmap-website/pull/5255#discussion_r1801064970), there's couple more issues with the interaction type: ["blue on blue" colors](https://github.com/openstreetmap/openstreetmap-website/pull/4700#issuecomment-2221682594) and whether the submitted/commented classification makes sense at all (what if I closed a note without a comment?). I wanted to try opened/closed/commented instead.

First of all, thank you for taking time and effort reviewing this. Few things I would like to reassess:
 - Related to the commented/submitted criteria we could use `EXISTS` subqueries which removes the need for a double joins, potentially making the query simpler and more performant. I could try testing this in terms of performance with large number of mocked entries so we can better get the idea of how this would affect the performance.

```ruby
  scope :filter_by_note_type, lambda { |note_type, user_id|
    case note_type
    when "commented"
      where("EXISTS (SELECT 1 FROM note_comments WHERE note_comments.note_id = notes.id AND (note_comments.author_id != ? OR note_comments.author_id IS NULL))", user_id)
    when "submitted"
      where("EXISTS (SELECT 1 FROM note_comments WHERE note_comments.note_id = notes.id AND note_comments.author_id = ? AND note_comments.id = (SELECT MIN(nc.id) FROM note_comments nc WHERE nc.note_id = notes.id))", user_id)
    else
      all
    end
  }
```

- Regarding the open/closed/commented differentiation instead of commented/submitted - It makes sense to me too but I would also like to hear the @tomhughes @gravitystorm opinions on this too, so we could all sync on right approach for this.

kcne:
As for splitting this into multiple smaller PRs, I would prefer we merge this as a complete feature/functionality. Adding filters one by one would significantly increase the time required and might leave the feature feeling incomplete. This PR covers all the intended functionality cohesively, and it seems better to refine and adjust post-merge if needed rather than delay further by segmenting the changes.

AntonKhorev:
> 3. When you mentioned "blue on blue" colors I'm not sure if this is something I've introduced in this PR

Obviously you didn't introduce it. That comment was made in July.

> since when I toggle dark mode on my end everything is looking fine.

Check it when the note is colored as "submitted".

AntonKhorev:
> Adding filters one by one would significantly increase the time required

Time for what? For PR review?

> and might leave the feature feeling incomplete.

Are you sure? Is it complete with exactly the filters you added and not some other filters?

kcne:
> Time for what? For PR review?

Exactly, since each PR would need to be reviewed sequentially, it would prolong the overall process.

> Are you sure? Is it complete with exactly the filters you added and not some other filters?

Yes, the goal was to implement the filters that were most commonly requested and improve the feature based on that feedback. However, if you have any suggestions for additional filters that could enhance the usability for mappers, I’m definitely open to discussing and considering those ideas too.

AntonKhorev:
> Yes, the goal was to implement the filters that were most commonly requested and improve the feature based on that feedback.

Where are these requests? For example #832 wants status filter but doesn't say anything about date ranges. Why would a status filter be incomplete without a date range filter?

kcne:
I opened new PR #5297 according to suggestion here. Closing this as not planed for now.

> You can make a pull request that just adds the status filter. That doesn't depend on pagination and shouldn't tank the performance because:
> 
>     - status is checked anyway to see if the note is not hidden
> 
>     - we're doing a similar thing in the api

PR #5257, opened 2024-10-11
Update en-GB.yml: don't say "import" when you mean "upload"

jidanni:
https://wiki.openstreetmap.org/wiki/Import says                                              
> Unlike editing the map or **uploading GPX traces**, importing combines                     
> an existing dataset with the OSM dataset and thus typically includes a                     
> complex merging process.                                                                   
                                                                                             
Therefore the mail one gets, 
> Subject: [OpenStreetMap] GPX Import success                                                   

shouldn't say "Import", it should just say "Upload".                                         
                                                                                             
I'll just fix these two. I appears elsewhere too.

### DIFF
--- a/config/locales/en-GB.yml
+++ b/config/locales/en-GB.yml
@@ -1145,9 +1145,9 @@ en-GB:
       see_their_profile: You can see their profile at %{userurl}.
       befriend_them: You can also add them as a friend at %{befriendurl}.
     gpx_failure:
-      subject: '[OpenStreetMap] GPX Import failure'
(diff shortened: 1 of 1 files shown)


mmd-osm:
As per https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#i18n, I'm closing this PR:

_If you make a change that involve the locale files (in config/locales) then please **only submit changes to the en.yml file**. The other files are updated via [Translatewiki](https://translatewiki.net/wiki/Translating:OpenStreetMap) and should not be included in your pull request._

Issue #3310, opened 2021-09-03
Modify the way "Friends" are added


Fizzie41:
I recently received a message saying that "(A person that I've never heard of) has added you as a friend on OpenStreetMap", but when I clicked on the link to their user page, it said that that User does not exist.

Following advice, I raised the matter with the DWG as I was concerned about spam, & they advised that that user would have already been deleted, probably after somebody else reported them for the same reason.

I made some suggestions to prevent this from happening again, but they said I should raise them here. 

"Making friends" shouldn't only be controlled by one side of the proposed new arrangement. If I decide that I want to be friends with you, then instead of you just getting a message that Fizzie41 is now your friend, you should get a message saying that "Fizzie41 (link to user page) would like to become your friend - do you agree? - Yes / No". If you say Yes, then I get a response saying that you have agreed; if No, I either get nothing back at all, or I get a polite "Sorry, your Friend request has been declined" message.

Further to that, is there any way for a User to see who has listed them as a friend?

Thanks,

Graeme

tomhughes:
Well there are two approaches to "making friends" in use - the facebook model that requires agreement and the twitter model which doesn't. Though I think even facebook has one way follows as well now.

We chose to go with the second model.

Friending on OSM doesn't really do anything anyway - it's entirely harmless to the other party (you in this case) because other than sending you a notification it doesn't change anything for you.

All it really does is "bookmark" somebody so that they are included in a list of people that you can narrow down the trace and changeset views to just those people.

Probably the real answer is that we just shouldn't send notifications, and should maybe rename the feature as "follow" rather than "friend" to better reflect what it really does.

Fizzie41:
Thanks Tom!

In that case, no, it's probably nothing to be concerned about then.

Although "follow", rather than "friend" is a good idea!

matkoniecz:
> Probably the real answer is that we just shouldn't send notifications

That would be a very good idea as scam accounts started to use this functionality (other people reporting, I also just got and reported such spam).

Dimitar5555:
Since #5261 is merged, this issue should be marked as resolved.

Fizzie41:
> All it really does is "bookmark" somebody so that they are included in a list of people that you can narrow down the trace and changeset views to just those people.

Following up on this ...

I am now a member of the DWG.

During this last week, I got a message that <user>has added me as a friend, but on checking to see who they are, it turns out to be somebody who's account I have recently blocked for vandalism! 

[...]

PR #5261, opened 2024-10-15
Modify the way "Friends" are added

nertc:
This PR addresses "Modify the way 'Friends' are added" issue mentioned in https://github.com/openstreetmap/openstreetmap-website/issues/3310

This PR replaces all occurrences of "friends" to "followers", including texts, tests and URLs.

This PR was not broken down to smaller pieces to maintain consistency and avoid different names of the same functionality.


config/locales/en.yml
```diff
@@ -91,7 +91,7 @@ en:
         support_url: Support URL
         allow_read_prefs:  read their user preferences
         allow_write_prefs: modify their user preferences
-        allow_write_diary: create diary entries, comments and make friends
+        allow_write_diary: create diary entries, comments and follow users
```
AntonKhorev, line 94: This permission is not used now but if it was, would it make sense to bundle follow/unfollow together with diary editing/commenting? Maybe just remove following from here?

tomhughes: Agreed as that is never used and there's no good reason to include that Id just drop it from the comment and if we ever want a permission for managing follows we can add that separately.

config/locales/en.yml
```diff
@@ -1603,12 +1603,12 @@ en:
       footer_html: "You can also read the message at %{readurl} and you can send a message to the author at %{replyurl}"
     friendship_notification:
       hi: "Hi %{to_user},"
-      subject: "[OpenStreetMap] %{user} added you as a friend"
-      had_added_you: "%{user} has added you as a friend on OpenStreetMap."
+      subject: "[OpenStreetMap] %{user} followed you"
+      had_followed_you: "%{user} has followed you on OpenStreetMap."
```
tomhughes, line 1607: I'm not sure why this key has a `had_` prefix but it was always incorrect so I'd suggest just dropping it and using `followed_you`.

mmd-osm: I would probably expect a slightly different wording here, like "%{user} is following you on OpenStreetMap now.". 

"has followed you" sounds to me like this is a thing of the past, and not necessarily the case anymore.

config/routes.rb
```diff
@@ -280,8 +280,8 @@
   resource :profile, :only => [:edit, :update]
 
   # friendships
-  match "/user/:display_name/make_friend" => "friendships#make_friend", :via => [:get, :post], :as => "make_friend"
-  match "/user/:display_name/remove_friend" => "friendships#remove_friend", :via => [:get, :post], :as => "remove_friend"
+  match "/user/:display_name/follow_user" => "friendships#follow_user", :via => [:get, :post], :as => "follow_user"
+  match "/user/:display_name/unfollow_user" => "friendships#unfollow_user", :via => [:get, :post], :as => "unfollow_user"
```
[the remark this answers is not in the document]
gravitystorm: From a quick look at some gems that implement this functionality ([1](https://github.com/cmer/socialization), [2](https://github.com/tcocca/acts_as_follower)) the convention is to use "follow" as a noun, so table is `follows`, controller is `follows_controller`, model is `Follow` etc. But I think we should rename the controllers/models separately.

What it does suggest though is that we should use resourceful routing with this in mind, so for example "/user/:display_name/follow" with create/destroy actions (and new/edit for the GET pages). See `user_mutes` for an example in the routes.rb file that's a bit like this.

test/system/dashboard_test.rb
```diff
@@ -17,9 +17,9 @@ class DashboardSystemTest < ApplicationSystemTestCase
     sign_in_as(user)
 
     visit dashboard_path
-    assert_no_text "You have not added any friends yet."
+    assert_no_text "You have not followed any user yet."
 
-    friends_heading = find :element, "h2", :text => "My friends"
+    friends_heading = find :element, "h2", :text => "Followed users"
```
kcne, line 22: Maybe use  "Following" instead of "Followed users" for better clarity. It aligns with the common terminology users are familiar with, like "Following" and "Followers," which may feel more intuitive.

config/locales/en.yml
```diff
@@ -652,18 +652,18 @@ en:
...
+      button: "Follow User"
+      success: "You are now following %{name}!"
+      failed: "Sorry, failed to follow %{name}."
+      already_followed: "You already follow %{name}."
+      limit_exceeded: "You have followed a lot of users recently. Please wait a while before trying to follow any more."
+    unfollow_user:
+      heading: "Do you want to unfollow %{user}?"
+      button: "Unfollow"
+      success: "%{name} was unfollowed by you."
+      not_followed: "%{name} is not followed by you."
```
kcne, line 666: A bit more active voicing alternative would be "You are not following %{name}".

config/locales/en.yml
```diff
@@ -652,18 +652,18 @@ en:
...
+      heading: "Do you want to follow %{user}?"
+      button: "Follow User"
+      success: "You are now following %{name}!"
+      failed: "Sorry, failed to follow %{name}."
+      already_followed: "You already follow %{name}."
+      limit_exceeded: "You have followed a lot of users recently. Please wait a while before trying to follow any more."
+    unfollow_user:
+      heading: "Do you want to unfollow %{user}?"
+      button: "Unfollow"
+      success: "%{name} was unfollowed by you."
```
kcne, line 665: This sound a bit passive to me, altough I'm not sure. Alternative could be something like "You successfully unfollowed %{name}."

config/locales/en.yml
```diff
@@ -652,18 +652,18 @@ en:
...
-    remove_friend:
-      heading: "Unfriend %{user}?"
-      button: "Unfriend"
-      success: "%{name} was removed from your friends."
-      not_a_friend: "%{name} is not one of your friends."
+    follow_user:
+      heading: "Do you want to follow %{user}?"
+      button: "Follow User"
+      success: "You are now following %{name}!"
+      failed: "Sorry, failed to follow %{name}."
```
kcne, line 659: Suggesting one more alternative: "Sorry, your request for following %{name} has failed." Wording sounds a bit unnatural here imho

app/views/dashboards/show.html.erb
```diff
@@ -27,19 +27,19 @@
     </div>
 
     <div class="col-md">
-      <h2><%= t ".my friends" %></h2>
+      <h2><%= t ".followed users" %></h2>
```
kcne, line 30: Maybe use "Following" instead of "Followed users" for better clarity. It aligns with the common terminology users are familiar with, like "Following" and "Followers," which may feel more intuitive. Just a thought.

config/routes.rb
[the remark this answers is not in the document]
nertc: When I started implementing same architecture as `user_mutes`, it caused to change some more files and add some complexities. This PR is already `+104 -104` line change. I think, it will be better to create a separate PR about changing routing architecture of the `follow` functionality after this PR will be merged to avoid making it even bigger and more complex.

nertc:
Thank you for the review. This PR was updated according to recommendations.

config/locales/en.yml
nertc: `follow users` part was removed

app/views/dashboards/show.html.erb
```diff
@@ -27,19 +27,19 @@
...
           <ul class='clearfix'>
-            <li><%= link_to t(".friends_changesets"), friend_changesets_path %></li>
-            <li><%= link_to t(".friends_diaries"), friends_diary_entries_path %></li>
+            <li><%= link_to t(".followed_changesets"), friend_changesets_path %></li>
+            <li><%= link_to t(".followed_diaries"), friends_diary_entries_path %></li>
           </ul>
         </nav>
         <div>
-          <%= render :partial => "contact", :collection => friends, :locals => { :type => "friend" } %>
+          <%= render :partial => "contact", :collection => friends, :locals => { :type => "followed user" } %>
```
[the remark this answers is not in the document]
nertc: `followed user` was shortened as `following` (according to the common practice mentioned in the https://github.com/openstreetmap/openstreetmap-website/pull/5261#discussion_r1803345542)

config/locales/en.yml
nertc: `had` was removed and now key is `followed_you`. Also, wording was changed according to suggestions.

config/locales/en.yml
nertc: Wording was changed according to the suggestion.

config/locales/en.yml
nertc: Wording was changed according to the suggestion.

app/views/dashboards/show.html.erb
nertc: I agree with you. Wording was changed from `followed users` to `followings`. I was also thinking about it. `followed users` is more understandable, but `followings` keeps sufficient understandability and makes terminology shorter and more visually appealing.

config/locales/en.yml
nertc: Wording was changed according to the suggestion.

config/routes.rb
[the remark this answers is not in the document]
nertc: @AntonKhorev `app/views/user_mailer/friendship_notification.html.erb` and `app/views/user_mailer/friendship_notification.text.erb` changes are about redirections from mail notifications. For the consistent visual output this PR is focused on changing every occurrence of "friend" on user side and changing them to "following". Main focus is what will users see (including both texts and URLs). I agree that it will be great if friendship functionality is refactored (currently it is just renamed), but those changes are out of scope of this PR and will be done as a separate PR as soon as this PR will be merged (adding them to this PR will complicate things as there will be not only user side changes, but logical changes in `routes.rb`, `ability.rb` and etc. which will need much deeper testing than just text and URL comparisons).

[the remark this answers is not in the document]
AntonKhorev: > `app/views/user_mailer/friendship_notification.html.erb` and `app/views/user_mailer/friendship_notification.text.erb` changes are about redirections from mail notifications.

I don't see any redirects from old GET paths to new GET paths (but I see changes in tests that checked the paths).

> For the consistent visual output this PR is focused on changing every occurrence of "friend" on user side and changing them to "following". Main focus is what will users see (including both texts and URLs).

Maybe it shouldn't focus on *both*?

nertc:
PR was updated. Now friendship controller and model are refactored to be named as "follow" and also implement standard CRUD architecture using `create` and `destroy`. Old controllers, models, views and tests were removed, and new ones were added with the new names and architecture.

config/routes.rb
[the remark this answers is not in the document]
nertc: PR was updated accordingly to include refactor and renaming of the `friendship` functionality.

config/routes.rb
```diff
@@ -280,8 +280,10 @@
   resource :profile, :only => [:edit, :update]
 
   # friendships
-  match "/user/:display_name/make_friend" => "friendships#make_friend", :via => [:get, :post], :as => "make_friend"
-  match "/user/:display_name/remove_friend" => "friendships#remove_friend", :via => [:get, :post], :as => "remove_friend"
+  scope "/user/:display_name" do
+    resource :follow, :only => [:create, :destroy], :path => "follow"
+  end
+  get "/user/:display_name/follow" => "follows#index"
```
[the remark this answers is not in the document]
nertc: It's a page that either shows unfollow functionality or follow functionality based on if user already follows `:display_name`. As it is default `GET` method for this functionality (while other ones are `POST` and `DELETE`), I decided to name it `follows#index`. If you have an idea of a better name, I am open to it.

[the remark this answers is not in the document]
AntonKhorev: Why do you need this page instead of using the standard `:new` and `:edit` pages?
https://guides.rubyonrails.org/v7.2/routing.html#singular-resources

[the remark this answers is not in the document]
nertc: I agree with you. Changed `GET /follow` to `GET /follow/edit`. It's much intuitive and consistent.

config/routes.rb
```diff
@@ -280,8 +280,9 @@
   resource :profile, :only => [:edit, :update]
 
   # friendships
-  match "/user/:display_name/make_friend" => "friendships#make_friend", :via => [:get, :post], :as => "make_friend"
-  match "/user/:display_name/remove_friend" => "friendships#remove_friend", :via => [:get, :post], :as => "remove_friend"
+  scope "/user/:display_name" do
+    resource :follow, :only => [:create, :destroy, :edit], :path => "follow"
+  end
```
AntonKhorev, line 285: Still no redirects. Old `make_friend` urls in emails are going to be broken.

app/controllers/follows_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  layout "site"
+
+  before_action :authorize_web
+  before_action :set_locale
+  before_action :check_database_readable
+
+  authorize_resource
+
+  before_action :check_database_writable, :only => [:create, :destroy, :edit]
+  before_action :lookup_friend, :only => [:create, :destroy, :edit]
```
AntonKhorev, line 13: Do you need `:only` here? It lists all of the actions.

app/models/follow.rb
```diff
@@ -18,9 +18,9 @@
...
 #
 
-class Friendship < ApplicationRecord
+class Follow < ApplicationRecord
   self.table_name = "friends"
 
-  belongs_to :befriender, :class_name => "User", :foreign_key => :user_id
-  belongs_to :befriendee, :class_name => "User", :foreign_key => :friend_user_id
+  belongs_to :befriender, :class_name => "User", :foreign_key => :user_id, :inverse_of => :follows
+  belongs_to :befriendee, :class_name => "User", :foreign_key => :friend_user_id, :inverse_of => :follows
```
AntonKhorev, line 25: We probably want to keep the existing db column names like `friend_user` for now, but what about `:befriender` and `:befriendee`? Currently you have code like this, mixing "follow" and "friend" terminology:

```ruby
Follow.find_by(:befriender => user, :befriendee => follow)
```

nertc: I tried to change as few things in DB as possible, but I agree with you. It will be changed in the next commit.

AntonKhorev: `:befriender` and `:befriendee` are not in db, `:friend_user_id` is. I'm not proposing to change `:friend_user_id` right now.

nertc: `:befriender: was changed to `:follower` and `:befriendee` to `:following`

app/controllers/follows_controller.rb
nertc: `:only` was removed

config/routes.rb
nertc: Redirections were added from `/make_friend` and `/remove_friend` to `/follow/edit`

app/views/users/show.html.erb
```diff
@@ -84,9 +84,9 @@
             <% if current_user %>
               <li>
                 <% if current_user.friends_with?(@user) %>
-                  <%= link_to t(".remove as friend"), remove_friend_path(:display_name => @user.display_name), :method => :post %>
+                  <%= link_to t(".unfollow"), follow_path(:display_name => @user.display_name), :method => :delete %>
                 <% else %>
-                  <%= link_to t(".add as friend"), make_friend_path(:display_name => @user.display_name), :method => :post %>
+                  <%= link_to t(".follow"), follow_path(:display_name => @user.display_name), :method => :post %>
```
AntonKhorev, line 89: This doesn't quite work. There's a link that points to `/user/.../follow`. I can try to open it in a new tab and it won't work because the page doesn't exist. The page is at `.../edit` but you can't change the link to `.../edit` because you can't post/delete it. Looks like that out of options in https://github.com/openstreetmap/openstreetmap-website/pull/5261#discussion_r1874500481 `:show` would have worked the best, especially if you agree with https://github.com/openstreetmap/openstreetmap-website/pull/5449#issuecomment-2578415862.

nertc: My thought was to use `:edit` as the functionality was focused on editing but not showing information about `follows`. But as https://github.com/openstreetmap/openstreetmap-website/pull/5449#issuecomment-2578415862 suggests, if we use `:show` also as edit page where it is possible, I will change it accordingly. I thought your comment meant to use standard `Singular Resource` routes in general, not `:show` in particular.

config/locales/en.yml
```diff
@@ -686,19 +686,22 @@ en:
...
-      success: "%{name} was removed from your friends."
-      not_a_friend: "%{name} is not one of your friends."
+  follows:
+    edit:
+      follow:
+        heading: "Do you want to follow %{user}?"
+        button: "Follow User"
+      unfollow:
+        heading: "Do you want to unfollow %{user}?"
+        button: "Unfollow"
```
AntonKhorev, line 696: Why one of the buttons is "Follow User" but another one is just "Unfollow"?

nertc: I tried to stick to the formulations from the `friendship` and as they were `button: "Add as friend"` and `button: "Unfriend"`, I changed them to `button: "Follow User"` and `button: "Unfollow"` accordingly.

app/models/user.rb
```diff
@@ -58,8 +58,8 @@ class User < ApplicationRecord
   has_many :new_messages, -> { where(:to_user_visible => true, :muted => false, :message_read => false).order(:sent_on => :desc) }, :class_name => "Message", :foreign_key => :to_user_id
   has_many :sent_messages, -> { where(:from_user_visible => true).order(:sent_on => :desc).preload(:sender, :recipient) }, :class_name => "Message", :foreign_key => :from_user_id
   has_many :muted_messages, -> { where(:to_user_visible => true, :muted => true).order(:sent_on => :desc).preload(:sender, :recipient) }, :class_name => "Message", :foreign_key => :to_user_id
-  has_many :friendships, -> { joins(:befriendee).where(:users => { :status => %w[active confirmed] }) }
-  has_many :friends, :through => :friendships, :source => :befriendee
+  has_many :follows, -> { joins(:following).where(:users => { :status => %w[active confirmed] }) }
+  has_many :followings, :through => :follows, :source => :following
```
AntonKhorev, line 62: Why `:following` and not `:followee`?

Currently `:follows` are right next to `:followings` and you have to know the difference between these.

nertc: > Maybe use "Following" instead of "Followed users" for better clarity. It aligns with the common terminology users are familiar with, like "Following" and "Followers," which may feel more intuitive. Just a thought.

As mentioned in previous comments (https://github.com/openstreetmap/openstreetmap-website/pull/5261#discussion_r1803345542) common terminology is to use `followers` (user who follows) and `followings` (user who is followed). Therefore, I used `:followings`.

About `:follows`, it was change for `:friendships`. Another version I thought about was to name it `:followerships` but it needed additional `:class_name => ` attributes to work and for that, I chose to have standard `:follows` (as produced from the model's name `Follow`)

app/views/users/show.html.erb
AntonKhorev: I wanted to say that singular resources don't have `:index`.

app/models/user.rb
AntonKhorev: OK, let's keep `followings`.

config/locales/en.yml
AntonKhorev: Of course you wouldn't write "Unfriend as friend" (as opposed to "Add as friend"), but you can write "Unfollow User".

app/views/users/show.html.erb
nertc: `:edit` was changed with `:show` and now user can open `follow` and `unfollow` links in the new tab too.

config/locales/en.yml
nertc: `Unfollow` was changed to `Unfollow User`

AntonKhorev:
Merged, thanks.

nertc:
@tomhughes After this PR was merged, on 21 Jan there was a case when there was an old version of the subject and a new version of the body in the mail. After I saw this screenshot, I tried to find out the reason, but mail subject and body were changed in the same PR, I couldn't reproduce it neither on production nor on local environment and after this case there were no PRs merged that were related to the friendship functionality. My thought was that it may be connected to deployment (maybe some job that affects subject was not yet finished). If you have any ideas or information what could cause difference between subject and body, feel free to share.

Screenshot:
follow

gravitystorm:
If I had to guess, my guess would be that your locale on the website is "en-GB". This would allow for the email in the screenshot, because:

* In en.yml the `friendship_notification.subject` text was changed, but the i18n key remained the same
* In en.yml the email content was changed, and the i18n key was also changed from `had_added_you` to `followed_you`
* Then the deployment happened, but en-GB.yml was not yet updated at this point
* So en-GB.yml had the old `subject` text and still said "friend"
* and en-GB.yml did not have the new `followed_you` key
* Then the email was created, with en-GB as the locale. The subject in en-GB was therefore the old subject. The `followed_you` key was not yet available, and so that was looked up in `en.yml` to find the text, which is the new one.

This is just guesswork though, maybe something else happened, but it at least fits what you observed.

verhovsky:
"Followings" for "list of people you're following" sounds grammatically incorrect to me. In my experience websites use just "Following" for this as in "[people you are] following".

AntonKhorev:
> "Followings" for "list of people you're following" here

See https://github.com/openstreetmap/openstreetmap-website/pull/5261#discussion_r1910425185, apparently some people think it's correct.

verhovsky:
But it's not. Following that comment thread, the only one who said they think that it's correct is @nertc, the original suggestion was "Following". As soon as I saw "Followings" on the website I thought "this was contributed by someone whose native language is not English" and came searching for this PR.

AntonKhorev:
https://en.wiktionary.org/wiki/following#Noun, 4

verhovsky:
The first quote for that definition comes from [this paper](https://link.springer.com/chapter/10.1007/978-3-642-54212-1_1), and this is the first paragraph of its abstract

> Online social network is becoming more and more prevalent currently. [...] Weibo increases rapidly and draws 100 million users within a year and a half [...]

further

> There are three reasons made us do research on Sina Weibo: (1) Weibo grows two
times faster than Twitter; 

The first quote used to "prove" that "followings" can be used in this way is from a paper written by people who are clearly not native English speakers.

I don't want to argue about this further, if you don't believe me, don't believe me.

SomeoneElseOSM:
> https://en.wiktionary.org/wiki/following#Noun, 4

(pedant alert)

Someone on social media will have "a following", which is a group of people that follow them.  The plural of that is ("followings") would apply to "more than one person who has more than one group following them", and wouldn't really apply here.

The words "You are following" would work in the English version of https://www.openstreetmap.org/dashboard .  Would that not fit because of space or translation reasons (i.e. is this word used elsewhere and must it be a noun)?

nertc:
Thank you @gravitystorm. I asked Aleksandar, and it seems, you are right. He has `en-GB` as `preferred language`.

1ec5:
I just noticed this renaming as I went to update translations over on Translatewiki.net. Like others here, I was also struck by the oddity of describing an individual as “a following” and a set of individuals as “followings”. It’s basically a made-up word at this point. Worse, as https://github.com/openstreetmap/openstreetmap-website/pull/5261#issuecomment-2613978828 points out, it’s completely backwards: a following is the people who are following _you_, not the people whom _you_ are following.

As it’s used in social media, “follow” is a figure of speech, a [skeuomorph](https://en.wikipedia.org/wiki/Skeuomorph), for the act of chasing after someone. The opposite of a _follower_ is a _leader_, and I think that underscores how poorly the social media model fits this website. (And, let’s be honest, the opposite of a _stalker_ is a _target_.)

In Vietnamese, I’ve already been translating “subscribe (to a changeset)” as “follow”, so I’d probably do that for this as well, but “a follow” unfortunately has no translation. The closest I can come up with would be “a followed user”, which might as well be what we say in English. Do we even need a user-facing word for this concept, apart from the internal data model? The dashboard has plenty of space to say “Users you are following” (above “Other nearby users”), and the notifications can say “[User name] is now following you”.

As far as I know, all this function currently does is list a user on your dashboard: https://github.com/openstreetmap/openstreetmap-website/issues/3310#issuecomment-912283262. Is the intention to eventually make this functionality behave more like a social media feed? If not, there must be some word that better describes it than either “friend” or “follow”. Maybe a better mental model would be “pinning a pinned user”.

PR #5262, opened 2024-10-15
Add older/newer links in front of diary, comment, issue, block pages

AntonKhorev:
image


nenad-vujicic (approved):
This works great on my side, thank you very much!

Wild idea: since we already have decision procedure, which answers the question "do we have multiple pages?", why don't we use it for hiding pagination buttons when they are not required / usable?

nertc (approved):
Tested and works perfectly. Every occurrence of "shared/pagination" has its own pair on top of the page.

tomhughes:
Sorry it's taken so long, but this looks good to me, thanks.

PR #5266, opened 2024-10-16
Avoid linking to redirects on the osmfoundation website

gravitystorm:
wiki.osmfoundation.org redirects to osmfoundation.org, so we can update our links to point directly at the correct page.


tomhughes:
Looks good to me, thanks.

PR #5276, opened 2024-10-21
Don't show older/newer buttons if all items fit on one page

AntonKhorev:
Do the suggestion https://github.com/openstreetmap/openstreetmap-website/pull/5262#pullrequestreview-2369829924. It's somewhat related to that pull request because pagination buttons will take more space. Maybe we don't want that if the page contents is small.

With this PR pagination buttons are shown as usual if there's more than one page:
image

But if there's only one page, they are removed:
image


nenad-vujicic (approved):
Works great! Thank you very much!!!

tomhughes:
Looks good to me, thanks.

PR #5277, opened 2024-10-21
Use system font on error pages

AntonKhorev:
I opened #5131 and put a number of changes to error pages to see which of them are acceptable. The discussion there however has been about making more changes. I guess that means nobody objects to the ones I made and we can have them merged.

This PR changes the font of error pages to a default system ui font. Bootstrap does the same thing except also adds fallback fonts for older browsers that don't understand `font-family: system-ui`. We don't need a serif font for a text this short because we don't use serif fonts even for longer texts on the website.

Addresses this from #3532:
> It also looks incredibly ugly, with loads of whitespace and the default browser font.

image

### DIFF
--- a/app/assets/stylesheets/errors.scss
+++ b/app/assets/stylesheets/errors.scss
@@ -1,3 +1,7 @@
+body {
+  font-family: system-ui;
+}
+
 .logo {
   float: left;
   margin: 10px;


kcne (approved):

tomhughes:
My PR hasn't been merged in two months so it must be good is an interesting take.

That said I'm not totally sure why I didn't merge it - possibly I was concerned about adding so much custom CSS code which we are generally trying to avoid.

Is there a reason not to use bootstrap here?

AntonKhorev:
> My PR hasn't been merged in two months so it must be good is an interesting take.

How else can I get changes from that PR to be discussed instead of being asked to make more changes like:

> Is there a reason not to use bootstrap here?

Yes https://github.com/openstreetmap/openstreetmap-website/pull/5130#discussion_r1738684207

AntonKhorev:
Merged in #5131.

Issue #4946, opened 2024-07-03
Better handling of version links for an unversioned element view


HolgerJeromin:
### Problem

Navigating older versions of an element is a bit cumbersome

### Description

Visit for example:
https://www.openstreetmap.org/way/238241022
This element has 110 versions and 110 is also the version I see right now.
We have on the bottom these links:
![image](https://github.com/openstreetmap/openstreetmap-website/assets/2410353/8d6d72ea-3bd7-44e6-b460-1fcd69a5140e)
Visiting the first version (`#1`) is "never" (debatable) useful.
Visiting the last version (`#110`) is never useful because it is exactly the same information which I am just seeing.

Clicking last version (`#110`) I now get 
![image](https://github.com/openstreetmap/openstreetmap-website/assets/2410353/67eb736a-22ab-484b-960d-651789c4dd53)

Perhaps this should be the view I should get in the first place...

tomhughes:
I think it's wrong to say the first version is never useful. I can certainly imagine wanting to use it.

I believe the "current" view and the history view of the current version are also not always exactly the same though they're certainly very similar.

PR #5278, opened 2024-10-21
Display old node versions on map view

AntonKhorev:
Shows node versions like in #4930 but without complications https://github.com/openstreetmap/openstreetmap-website/pull/4930#issuecomment-2198493197 and without saving api calls.

image

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -327,18 +327,18 @@ $(document).ready(function () {
   OSM.Browse = function (map, type) {
     var page = {};
 
-    page.pushstate = page.popstate = function (path, id) {
+    page.pushstate = page.popstate = function (path, id, version) {
       OSM.loadSidebarContent(path, function () {
-        addObject(type, id);
+        addObject(type, id, version);
       });
     };
 
-    page.load = function (path, id) {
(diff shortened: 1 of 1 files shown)


nenad-vujicic (approved):

I believe similar changes to remaining OSM.OldBrowse() occurrences will enable displaying old ways / relations versions? :-)

Q: I found one strange behavior with nodes / ways / relations pagination controls. They initially display "Version 1" and "last version", although they should perhaps display only "previous version" link initially. Is this expected behavior? I apologize for bad place for posting this, but didn't want to create new issue if behavior is expected.

image

AntonKhorev:
> I believe similar changes to remaining OSM.OldBrowse() occurrences will enable displaying old ways / relations versions? :-)

As soon as https://github.com/openstreetmap/openstreetmap-website/pull/4930#issuecomment-2198793487 is done.

> Q: I found one strange behavior with nodes / ways / relations pagination controls. They initially display "Version 1" and "last version", although they should perhaps display only "previous version" link initially.

But before that thing is done, *current version without a number* and *version with the last number* are really different pages, although the UI is probably not the best here. See also #4946.

tomhughes:
Looks good to me, thanks.

PR #5283, opened 2024-10-24
Note subscriptions

AntonKhorev:
Adds a subscription functionality to notes, similar to changesets and diary entry subscriptions.

Without this it's impossible to unsubscribe from notes you may not care about. If you did any mass note manipulation in some area, chances are someone else will start altering these notes later. In this case you'll receive waves of emails and there's no way to stop them.

See also https://osmfoundation.org/wiki/Operations/Minutes/2024-05-02#Discussion_on_spam_reports_via_ISP. "Encourage Microsoft to work on maintaining separate subscriptions for notes, as there was some work already done by them on refactoring the way notes are presented and they are interested in improving the notes." - I'm not sure what work on notes was done by Microsoft at that time.

---

Needs to be merged in at least two parts because backfilling. Backfilling is done in a migration here which may not be the best idea. We already did data migrations like that but they were simpler.

image

### DIFF
--- a/app/abilities/api_capability.rb
+++ b/app/abilities/api_capability.rb
@@ -9,6 +9,7 @@ class ApiCapability
 
       if user&.active?
         can [:create, :comment, :close, :reopen], Note if scope?(token, :write_notes)
+        can :create, NoteSubscription if scope?(token, :write_notes)
         can [:show, :data], Trace if scope?(token, :read_gpx)
         can [:create, :update, :destroy], Trace if scope?(token, :write_gpx)
         can [:details], User if scope?(token, :read_prefs)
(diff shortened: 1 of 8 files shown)


matkoniecz:
> If you did any mass note manipulation in some area, chances are someone else will start altering these notes later. In this case you'll receive waves of emails and there's no way to stop them.

relatively few mass note manipulation should be done then abandon discussion, though I expect that DWG is often involved in such actions which are entirely valid (undoing mass closing of notes for example)

also, very useful to disengage from very heated discussion where you are unable to participate in an useful way

AntonKhorev:
> DWG is often involved in such actions which are entirely valid (undoing mass closing of notes for example)

Yes. I got a batch of emails about reopened notes today and I got one yesterday.

AntonKhorev:
Implemented in https://github.com/openstreetmap/openstreetmap-website/pull/5284, https://github.com/openstreetmap/openstreetmap-website/pull/5304, https://github.com/openstreetmap/openstreetmap-website/pull/5310, https://github.com/openstreetmap/openstreetmap-website/pull/5314, https://github.com/openstreetmap/openstreetmap-website/pull/5346.

PR #5284, opened 2024-10-24
Note subscriptions db table

AntonKhorev:
Part one of #5283. Creates the note subscription table and starts adding subscriptions for anyone who comments a note. These subscriptions are not yet used for anything, email notifications still go to all note commenters.

### DIFF
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -401,6 +401,8 @@ module Api
       note.comments.map(&:author).uniq.each do |user|
         UserMailer.note_comment_notification(comment, user).deliver_later if notify && user && user != current_user && user.visible?
       end
+
+      NoteSubscription.find_or_create_by(:note => note, :user => current_user) if current_user
     end
   end
 end
--- a/test/controllers/api/notes_controller_test.rb
+++ b/test/controllers/api/notes_controller_test.rb
@@ -101,10 +101,12 @@ module Api
       )
     end
 
-    def test_create_success
+    def test_create_anonymous_success
       assert_difference "Note.count", 1 do
         assert_difference "NoteComment.count", 1 do
-          post api_notes_path(:lat => -1.0, :lon => -1.0, :text => "This is a comment", :format => "json")
+          assert_no_difference "NoteSubscription.count" do
+            post api_notes_path(:lat => -1.0, :lon => -1.0, :text => "This is a comment", :format => "json")
+          end
         end
       end
       assert_response :success
@@ -135,7 +137,7 @@ module Api
       assert_nil js["properties"]["comments"].last["user"]
     end
 
-    def test_create_fail
+    def test_create_anonymous_fail
       assert_no_difference "Note.count" do
         assert_no_difference "NoteComment.count" do
           post api_notes_path(:lon => -1.0, :text => "This is a comment")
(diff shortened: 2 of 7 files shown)


app/controllers/api/notes_controller.rb
```diff
@@ -401,6 +401,8 @@ def add_comment(note, text, event, notify: true)
       note.comments.map(&:author).uniq.each do |user|
         UserMailer.note_comment_notification(comment, user).deliver_later if notify && user && user != current_user && user.visible?
       end
+
+      NoteSubscription.find_or_create_by(:note => note, :user => current_user) if current_user
```
tomhughes, line 405: I think this could be simplified using the association?

```suggestion
      current_user&.note__subscriptions.find_or_create_by(:note => note)
```

test/controllers/api/notes_controller_test.rb
```diff
@@ -200,14 +202,44 @@ def test_create_fail
...
+      assert_equal "Point", js["geometry"]["type"]
+      assert_equal [-1.0, -1.0], js["geometry"]["coordinates"]
+      assert_equal "open", js["properties"]["status"]
+      assert_equal 1, js["properties"]["comments"].count
+      assert_equal "opened", js["properties"]["comments"].last["action"]
+      assert_equal "This is a comment", js["properties"]["comments"].last["text"]
+      assert_equal user.display_name, js["properties"]["comments"].last["user"]
+
+      note = Note.last
+      subscription = NoteSubscription.last
```
tomhughes, line 228: This (and all the other similar versions in other tests) is relying on nothing else having created notes or users in a way that overlaps this test and I'm not sure how safe that is when tests are running in parallel?

Maybe it would be better to get the node ID from the JSON response and then do `assert Note.exists?(id)` and a similar exists assertion on the subscription record?

gravitystorm: > I'm not sure how safe that is when tests are running in parallel?

Parallel tests are run in separate databases (e.g. `openstreetmap_test-0` etc)

AntonKhorev: > This ... is relying on nothing else having created notes

There is a check `assert_difference "Note.count", 1`

> (and all the other similar versions in other tests)

Other tests in this file or other tests elsewhere? You can find similar uses of `.last` from years ago.

app/controllers/api/notes_controller.rb
AntonKhorev: That would require one more `&.` after `note__subscriptions`, so it depends on whether you think that chained `&.`s are simplifications.

tomhughes: I don't believe it does - if `current_user` is defined then the `note_subscriptions` association will always exist. It might evaluate to an empty list but the association object will exist and can have methods invoked on it.

AntonKhorev: Try running Rubocop.

tomhughes: Hmm you're quite right, and rubocop is quite right, which I find quite odd but apparently ruby does continue evaluating the line after the first `&.` fails.

Personally I still think I prefer it but it's not a huge issue.

tomhughes:
It looks like there are no significant issues here, so I'll merge this. Thanks for the work.

PR #5288, opened 2024-10-25
Decrease padding between close button and inputs in directions form

AntonKhorev:
That space on the top of direction forms is too wasteful.

Before/after:
image image


tomhughes:
Looks good to me, thanks.

PR #5289, opened 2024-10-25
Remove callback from initialize() in note js controller

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/commit/a796c41881c26e6da8828a71dbfb533613e5a83f added a callback to `function initialize()` in `note.js`. It's called at the very end of the function. Not scheduled in some event, not skipped by early termination, nothing runs after it. Passing a callback to `initialize()` is the same as running `initialize(); callback();`


tomhughes:
Looks good to me, thanks.

PR #5290, opened 2024-10-26
Improvements to danger workflow

tomhughes:
This gets the danger workflow working by switching to my fixed version (current open upstream as https://github.com/danger/danger/pull/1501) and also tweaks a few other things:

* Renames the workflow to reflect that this is not just about labelling
* Drops the `run-name` so that he runs will be names after the PR instead of all having the same name
* Adjusts the label names to be consistent with our other labels.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,7 +1,5 @@
 name: Danger
 
-run-name: Updating labels for current PR
-
 on:
   pull_request_target:
     types: [opened, synchronize]
--- a/.github/workflows/labeling.yml
+++ b/.github/workflows/labeling.yml
@@ -28,4 +28,4 @@ jobs:
         env:
           DANGER_GITHUB_BEARER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          bundle exec danger --verbose --base=${{ github.event.pull_request.base.sha }} --head=${{ github.event.pull_request.head.sha }}
+          bundle exec danger --verbose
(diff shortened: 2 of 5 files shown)


mmd-osm:
I seems you need to update the commit hash in Gemfile.lock. This works for me:

```
diff --git a/Gemfile.lock b/Gemfile.lock
index 346af2baa..d1bf08281 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ GIT
 
 GIT
   remote: https://github.com/tomhughes/danger.git
-  revision: 5794971ade2fb4239a346a4902700e0c6d4c04c7
+  revision: a265cf74d2f464a25796b48d95697f5eed553454
   ref: pull-request-target
   specs:
```

tomhughes:
Ah sorry... I cherry picked and forgot I had updated the danger branch since. Now fixed.

mmd-osm:
Looks good now. The danger failure is expected, you can't test any changes to Danger with a cross repo PR.

gravitystorm:
Thanks for all your work getting to the bottom of this problem!

mmd-osm:
I think there's still an issue... I did a rebase on master + force push. Danger is still failing: https://github.com/openstreetmap/openstreetmap-website/pull/5080

mmd-osm:
I tried another pull request #5270, and here it seems to be working fine. The only obvious difference I can spot is that #5080 has a lot more commits.

PR #5291, opened 2024-10-26
Consistent use of quotes

amire80:
Most strings used single quotes, and a few used
opening backticks. Now it's single quotes everywhere.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1076,7 +1076,7 @@ en:
         man_made:
           adit: "Adit"
           advertising: "Advertising"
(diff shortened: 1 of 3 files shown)


mmd-osm:
Quick comment: you would also need to change a few unit tests with the same issue:

- test/controllers/user_roles_controller_test.rb:74
- test/controllers/user_roles_controller_test.rb:134
- test/controllers/messages_controller_test.rb:248
- test/controllers/messages_controller_test.rb:305

Can you take a loot at those as well?

tomhughes:
Looks good to me, thanks.

PR #5293, opened 2024-10-28
Use resourceful routes for granting/revoking user roles

AntonKhorev:
Replaces:
- `post grant_role_path` with `put user_role_path`
- `post revoke_role_path` with `delete user_role_path`


app/controllers/user_roles_controller.rb
```diff
@@ -9,15 +9,15 @@ class UserRolesController < ApplicationController
...
   before_action :lookup_user
   before_action :require_valid_role
-  before_action :not_in_role, :only => [:grant]
-  before_action :in_role, :only => [:revoke]
+  before_action :not_in_role, :only => :update
+  before_action :in_role, :only => :destroy
 
-  def grant
+  def update
     @user.roles.create(:role => @role, :granter => current_user)
```
gravitystorm, line 16: I would suggest that `create` is a better method name, because:

* `create` and `destroy` are more natural opposites of each other than `update` and `destroy`
* Since the next line is also `@user.roles.create(...)` then having both the controller-method-name and relation-method  being the same (`create`) will make it easier for developers to understand the code.

AntonKhorev: The routing system requires some convincing to achieve that. The less ugly version I could come up with is singular resources for roles, despite there are multiple possible roles for each user:

    resource :role, :controller => "user_roles", :path => "roles/:role", :only => [:create, :destroy]

PR #5295, opened 2024-10-29
Create base and head branches before running danger

tomhughes:
Let's have another go at getting danger working shall we...

This avoids the need to path danger, and resolves the problems with not fetching enough history for PRs with many commits, by prefetching the data and creating local `danger_base` and `danger_head` branches which are the default names that danger looks for.


mmd-osm:
Gemfile.lock update seems to be missing in the PR...

tomhughes:
That should be fixed now...

gravitystorm:
Merged, thanks!

PR #5297, opened 2024-10-30
Add status filter to user's note page

kcne:
### Description
This PR partly addresses #832 and serves as a smaller alternative to #5255 as suggested in comments. It aims to enhance user experience on user note pages by providing filtering functionality for notes by status. The `with_status` scope is added to the `Note` model, and the `NotesController` is updated accordingly. Default behavior remains unchanged since the status preselected value is set to `All` by default.

### How has this been tested?
Test for the new scope is added to note model test to ensure functionality is working as expected.

### Screenshots
Screenshot 2024-10-30 at 22 07 56
Screenshot 2024-10-30 at 22 07 46

### DIFF
--- a/app/views/notes/index.html.erb
+++ b/app/views/notes/index.html.erb
@@ -6,6 +6,20 @@
            :commented => tag.span(t(".subheading_commented"), :class => "px-2 py-1 bg-body") %></p>
 <% end %>
 
+<%= form_with :url => user_notes_path(@user), :method => :get, :data => { :turbo => true } do %>
+  <div class="row gx-2 align-items-end">
+    <div class="col-sm-auto mb-3">
+      <%= label_tag :status, t(".status") %>
+      <%= select_tag :status,
+                     options_for_select([[t(".all"), "all"], [t(".open"), "open"], [t(".closed"), "closed"]], params[:status] || "all"),
+                     :class => "form-select" %>
+    </div>
+    <div class="col-sm-auto mb-3">
+      <%= submit_tag t(".apply"), :name => nil, :class => "btn btn-primary" %>
+    </div>
+  </div>
+<% end %>
+
 <% if @notes.empty? %>
   <h4><%= t ".no_notes" %></h4>
 
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -18,12 +18,13 @@ class NotesController < ApplicationController
   def index
     param! :page, Integer, :min => 1
 
-    @params = params.permit(:display_name)
+    @params = params.permit(:display_name, :status)
     @title = t ".title", :user => @user.display_name
     @page = (params[:page] || 1).to_i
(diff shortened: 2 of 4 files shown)


tomhughes (changes requested):
In addition to the inline comment this could do with tests at the controller or system level rather than just testing the model method.

app/models/note.rb
```diff
@@ -34,6 +34,16 @@ class Note < ApplicationRecord
...
+  scope :with_status, lambda { |status|
+    case status
+    when "open"
+      where(:status => "open")
+    when "closed"
+      where(:status => "closed")
+    else
+      all
+    end
+  }
```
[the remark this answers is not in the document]
AntonKhorev: Do we even need the scope? A similar scope for issues is not used by the issues controller:

https://github.com/openstreetmap/openstreetmap-website/blob/e15a92a30213d87e10e000a240a7dcab66c208b4/app/controllers/issues_controller.rb#L37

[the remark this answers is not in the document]
kcne: Fixed and updated the PR. Since this is the only place using the scope, I removed it to simplify the code and maintain consistency. If a need arises, we can reintroduce it later.

Also, I’ve added a test in the controller to cover the basic status filtering functionality. Let me know if further tests are needed.

app/views/notes/index.html.erb
```diff
@@ -6,6 +6,20 @@
            :commented => tag.span(t(".subheading_commented"), :class => "px-2 py-1 bg-body") %></p>
 <% end %>
 
+<%= form_with :url => url_for(:controller => "notes", :action => "index"), :method => :get, :data => { :turbo => true } do %>
```
AntonKhorev, line 9: Why `:url => url_for(:controller => ...)` and not `:url => user_notes_path(@user)`?

app/views/notes/index.html.erb
```diff
@@ -6,6 +6,20 @@
...
+  <div class="row gx-2 align-items-end">
+    <div class="col-sm-auto mb-3">
+      <%= label_tag :status, t(".status") %>
+      <%= select_tag :status,
+                     options_for_select([[t(".all"), "all"], [t(".open"), "open"], [t(".closed"), "closed"]], params[:status] || "all"),
+                     :class => "form-select" %>
+    </div>
+    <div class="col-sm-auto mb-3">
+      <%= submit_tag t(".apply"), :class => "btn btn-primary" %>
```
AntonKhorev, line 18: We have this in issues search form to avoid adding `&commit=...` to urls:

```suggestion
      <%= submit_tag t(".apply"), :name => nil, :class => "btn btn-primary" %>
```

app/views/notes/index.html.erb
kcne: Missed this one. Refactored. Thank you.

AntonKhorev:
Merged, thanks.

PR #5301, opened 2024-11-04
Added new Danger rule for catching outdated Gemfile.lock

nenad-vujicic:
### Description

Updated the Dangerfile by adding a new rule that checks if the Gemfile is updated while the Gemfile.lock is not, and if so, generates the "gemfile-lock-outdated" PR label.

image

Created upon the [suggestion](https://github.com/openstreetmap/openstreetmap-website/pull/5295#issuecomment-2447849335).

### How has this been tested?

Tested by running RuboCop locally and by committing changes to my personal fork before creating a problematic PR.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -30,3 +30,13 @@ if git.commits.any? { |c| c.parents.count > 1 }
 else
   auto_label.remove("merge-commits")
 end
+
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #5302, opened 2024-11-05
Add profile location

nertc:
PR adds location name info on the user profile page. Location name can be changed from "Edit Profile" page either by manual typing or auto-filling according to the home location.

One column was added to the user table to save user's location.
JS logics are responsible for handling:
1) Delete-Undelete buttons interactions
2) Nominatim calls to locate home locations country name
3) Warning logic of the location name

image
image
image

### DIFF
--- a/app/assets/javascripts/home_location_name-endpoint.js
+++ b/app/assets/javascripts/home_location_name-endpoint.js
@@ -1,0 +1,65 @@
+OSM.HomeLocationNameGeocoder = function Endpoint(latInput, lonInput, locationNameInput) {
+  const endpoint = {
+    autofill: true,
+    countryName: locationNameInput.val().trim()
+  };
+
+  let requestController = null;
+
+  endpoint.updateHomeLocationName = function (
+    updateInput = true,
+    lat = latInput.val().trim(),
+    lon = lonInput.val().trim(),
+    successFn
+  ) {
+    if (!lat || !lon || !endpoint.autofill) {
+      return;
+    }
+
+    const geocodeUrl = "/search/nominatim_reverse_query",
+          csrf_param = $("meta[name=csrf-param]").attr("content"),
+          csrf_token = $("meta[name=csrf-token]").attr("content"),
+          params = new URLSearchParams({
+            lat,
+            lon,
+            zoom: 3
+          });
+    params.set(csrf_param, csrf_token);
+
+    if (requestController) {
+      requestController.abort();
+    }
+    const currentRequestController = new AbortController();
+    requestController = currentRequestController;
+
+    fetch(geocodeUrl, {
+      method: "POST",
+      body: params,
+      signal: requestController.signal,
+      headers: { accept: "application/json" }
+    })
+      .then((response) => response.json())
+      .then((data) => {
+        const country = data.length ? data[0].name : "";
+
+        if (updateInput) {
+          $("#home_location_name").val(country);
+        } else if (endpoint.countryName !== country) {
+          endpoint.autofill = false;
+        }
+        endpoint.countryName = country;
+        requestController = null;
+
+        if (successFn) {
+          successFn();
+        }
+      })
+      .catch(() => {
+        if (currentRequestController === requestController) {
+          requestController = null;
+        }
+      });
+  };
+
+  return endpoint;
+};
--- a/app/assets/javascripts/user.js
+++ b/app/assets/javascripts/user.js
@@ -1,4 +1,5 @@
 //= require leaflet.locate
+//= require ./home_location_name-endpoint
 
 (function () {
   $(document).on("change", "#user_all", function () {
@@ -8,13 +9,17 @@
 
 $(function () {
   const defaultHomeZoom = 12;
-  let map, marker, deleted_lat, deleted_lon;
+  let map, marker, deleted_lat, deleted_lon, deleted_home_name, homeLocationNameGeocoder, savedLat, savedLon;
 
   if ($("#map").length) {
     map = L.map("map", {
       attributionControl: false,
       zoomControl: false
     }).addLayer(new L.OSM.Mapnik());
+
+    savedLat = $("#home_lat").val();
+    savedLon = $("#home_lon").val();
+    homeLocationNameGeocoder = OSM.HomeLocationNameGeocoder($("#home_lat"), $("#home_lon"), $("#home_location_name"));
 
     const position = $("html").attr("dir") === "rtl" ? "topleft" : "topright";
 
@@ -48,9 +53,8 @@ $(function () {
         $("#home_lat").val(lat);
         $("#home_lon").val(lon);
 
-        deleted_lat = null;
-        deleted_lon = null;
-        respondToHomeUpdate();
+        clearDeletedText();
+        respondToHomeLatLonUpdate();
       }).on("moveend", function () {
         const lat = $("#home_lat").val().trim(),
               lon = $("#home_lon").val().trim();
@@ -68,9 +72,15 @@ $(function () {
       });
 
       $("#home_lat, #home_lon").on("input", function () {
-        deleted_lat = null;
-        deleted_lon = null;
-        respondToHomeUpdate();
+        clearDeletedText();
+        respondToHomeLatLonUpdate();
+      });
+
+      $("#home_location_name").on("input", function () {
+        homeLocationNameGeocoder.autofill = false;
+        clearDeletedText();
+
+        respondToHomeLatLonUpdate(false);
       });
 
       $("#home_show").click(function () {
@@ -82,21 +92,25 @@ $(function () {
 
       $("#home_delete").click(function () {
         const lat = $("#home_lat").val(),
-              lon = $("#home_lon").val();
+              lon = $("#home_lon").val(),
+              locationName = $("#home_location_name").val();
 
-        $("#home_lat, #home_lon").val("");
+        $("#home_lat, #home_lon, #home_location_name").val("");
         deleted_lat = lat;
         deleted_lon = lon;
-        respondToHomeUpdate();
+        deleted_home_name = locationName;
+
+        respondToHomeLatLonUpdate(false);
         $("#home_undelete").trigger("focus");
       });
 
       $("#home_undelete").click(function () {
         $("#home_lat").val(deleted_lat);
         $("#home_lon").val(deleted_lon);
-        deleted_lat = null;
-        deleted_lon = null;
-        respondToHomeUpdate();
+        $("#home_location_name").val(deleted_home_name);
+        clearDeletedText();
+
+        respondToHomeLatLonUpdate(false);
         $("#home_delete").trigger("focus");
       });
     } else {
@@ -110,14 +124,26 @@ $(function () {
     }
   }
 
-  function respondToHomeUpdate() {
+  function respondToHomeLatLonUpdate(updateLocationName = true) {
     const lat = $("#home_lat").val().trim(),
-          lon = $("#home_lon").val().trim();
+          lon = $("#home_lon").val().trim(),
+          locationName = $("#home_location_name").val().trim();
     let location;
 
     try {
       if (lat && lon) {
         location = L.latLng(lat, lon);
+        if (updateLocationName) {
+          if (savedLat && savedLon && $("#home_location_name").val().trim()) {
+            homeLocationNameGeocoder.updateHomeLocationName(false, savedLat, savedLon, () => {
+              savedLat = savedLon = null;
+              homeLocationNameGeocoder.updateHomeLocationName();
+            });
+          } else {
+            savedLat = savedLon = null;
+            homeLocationNameGeocoder.updateHomeLocationName();
+          }
+        }
       }
       $("#home_lat, #home_lon").removeClass("is-invalid");
     } catch (error) {
@@ -127,8 +153,11 @@ $(function () {
 
     $("#home_message").toggleClass("invisible", Boolean(location));
     $("#home_show").prop("hidden", !location);
-    $("#home_delete").prop("hidden", !location);
-    $("#home_undelete").prop("hidden", !(!location && deleted_lat && deleted_lon));
+    $("#home_delete").prop("hidden", !location && !locationName);
+    $("#home_undelete").prop("hidden", !(
+      (!location || !locationName) &&
+      ((deleted_lat && deleted_lon) || deleted_home_name)
+    ));
     if (location) {
       marker.setLatLng([lat, lon]);
       marker.addTo(map);
@@ -153,6 +182,12 @@ $(function () {
     } else {
       $("input#user_auth_uid").hide().prop("disabled", true);
     }
+  }
+
+  function clearDeletedText() {
+    deleted_lat = null;
+    deleted_lon = null;
+    deleted_home_name = null;
   }
 
   updateAuthUID();
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,6 +12,7 @@
 #  home_lat             :float
 #  home_lon             :float
 #  home_zoom            :integer          default(3)
+#  home_location_name   :string
 #  pass_salt            :string
 #  email_valid          :boolean          default(FALSE), not null
 #  new_email            :string
--- a/app/views/profiles/edit.html.erb
+++ b/app/views/profiles/edit.html.erb
@@ -52,6 +52,7 @@
         <button type="button" id="home_undelete" class="btn btn-outline-primary" hidden><%= t ".undelete" %></button>
       </div>
     </div>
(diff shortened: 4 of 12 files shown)


tomhughes:
This seems like an awful lot of complication...

It seems to me that we have the option of either allowing the user to input freetext, or of generating something from the location with Nominatim but this tries to do both at once which just makes things very complicated and probably quite confusing.

How often does nomination actually manage to match? I don't see any indication of zoom on the reverse geocode, and you're just taking the last name, so how likely is it that will match what the user entered even if they are trying to be accurate? Are we just going to wind up telling everybody their location doesn't match?

nertc:
@tomhughes Thank you for the comment. I agree with you about the complexity of the code, but there are several reasons for both methods.

Changing location name manually solves cases like:
1. Disputed areas - someone living in the area A may not want to show that area is part of country B and not C.
2. Several countries - someone who has citizenship of several countries may want to set his location name to something like "Georgia & Germany"
3. Choosing privacy - someone may want to explicitly mention "Tbilisi, Georgia", but someone may want to just show the country and keep other details private

Meanwhile, changing location with Nominatim autofill introduces:
1. Better UX - if user is okay with showing only home country, autofill avoids extra typing from user
2. Standardization - default value introduces default standard for user to easily perceive what kind of and what detalization of information is recommended for the field
3. Motivation to fill - when user sees a warning about the field that it doesn't match home location, it subconsciously motivates him to click the autofill button (just one easy click) and increases use of the feature

By choosing only one solution we sacrifice either functionality or better UX.

Currently, autofill suggests name of the country and warning is only shown if manually typed location doesn't contain country name. Therefore, "Germany & Georgia", "Tbilisi, Georgia", "Georgia" won't show any autofill warning if user has home location set in Georgia.

test/system/user_location_change_test.rb
```diff
@@ -0,0 +1,22 @@
...
+class UserLocationChangeTest < ApplicationSystemTestCase
+  def setup
+    stub_request(:get, /.*gravatar.com.*d=404/).to_return(:status => 404)
+  end
+
+  test "User can change their location" do
+    user = create(:user)
+    sign_in_as(user)
+
+    visit user_path(user.display_name)
```
AntonKhorev, line 12: You write it differently below
```suggestion
    visit user_path(user)
```

app/assets/javascripts/user.js
```diff
@@ -178,6 +220,42 @@ $(document).ready(function () {
...
+    const lat = $("#home_lat").val().trim();
+    const lon = $("#home_lon").val().trim();
+    if (!lat || !lon) {
+      return;
+    }
+
+    const geocodeUrl = `${OSM.NOMINATIM_URL}reverse?format=json&lat=${lat}&lon=${lon}`;
+
+    if (locationInput.request) locationInput.request.abort();
+    locationInput.request = $.getJSON(geocodeUrl, function (data) {
```
AntonKhorev, line 233: blocked by content security policy

app/views/profiles/edit.html.erb
```diff
@@ -52,6 +52,13 @@
         <button type="button" id="home_undelete" class="btn btn-outline-primary" hidden><%= t ".undelete" %></button>
       </div>
     </div>
+    <div class="my-2">
+      <%= f.text_field :location_name, :wrapper_class => "col-sm-4 d-flex flex-column pe-3", :class => "mt-auto", :id => "location_name" %>
+      <div id="location_name_warning" class="row align-items-center d-none">
+        <p class="m-0 w-auto fs-6 pe-1"><%= t ".location_name_warning" %></p>
+        <button id="location_default_name" class="btn btn-link p-0 w-auto text-sm-start" type="button"></button>
+      </div>
```
AntonKhorev, line 60: This looks like a sentence and it's not obvious that it ends with a button I'm supposed to click to update the value:

image

app/assets/javascripts/user.js
```diff
@@ -178,6 +220,42 @@ $(document).ready(function () {
...
   }
 
+  function updateHomeLocation() {
+    const lat = $("#home_lat").val().trim();
+    const lon = $("#home_lon").val().trim();
+    if (!lat || !lon) {
+      return;
+    }
+
+    const geocodeUrl = `${OSM.NOMINATIM_URL}reverse?format=json&lat=${lat}&lon=${lon}`;
```
AntonKhorev, line 230: If you only want a country why don't you ask for a country
```suggestion
    const geocodeUrl = `${OSM.NOMINATIM_URL}reverse?format=json&lat=${lat}&lon=${lon}&zoom=3`;
```
instead of getting a full address of things you don't care about and then doing `.split(",")`?

test/system/user_location_change_test.rb
nertc: Was updated and removed to remain consistent in the test file.

app/assets/javascripts/user.js
nertc: Was updated to get information from the local call instead of directly accessing Nominatim.

app/views/profiles/edit.html.erb
nertc: Was changed with the button named "Autofill %{country}".

app/assets/javascripts/user.js
nertc: Thank you for the suggested code. Was updated accordingly.

nertc:
Thank you for the review. PR was updated according to the comments.

app/assets/javascripts/user.js
```diff
@@ -178,6 +220,48 @@ $(document).ready(function () {
...
+    });
+  }
+
+  function updateLocationWarning() {
+    const locationName = $("#location_name").val();
+    if (!locationInput.dirty || locationName.includes(locationInput.countryName)) {
+      $("#location_name_warning").addClass("d-none");
+    } else {
+      $("#location_name_warning").removeClass("d-none");
+    }
```
AntonKhorev, line 262: Wouldn't using jQuery's `show` / `hide` / `toggle` be simpler than manipulating classes?

app/views/profiles/edit.html.erb
AntonKhorev: The language of %{country} is currently based on the browser settings / Accept-Language header. Is it what we want?

app/views/users/show.html.erb
```diff
@@ -139,6 +139,14 @@
       <div class='text-body-secondary'>
         <small>
           <dl class="list-inline">
+            <% if @user.location_name && @user.location_name.strip.present? %>
+              <dt class="list-inline-item m-0">
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-geo-alt-fill" viewBox="0 0 16 16">
+                  <path d="M8 16s6-5.686 6-10A6 6 0 0 0 2 6c0 4.314 6 10 6 10m0-7a3 3 0 1 1 0-6 3 3 0 0 1 0 6" />
+                </svg>
```
AntonKhorev, line 146: We probably want `<title>` in this svg.

app/views/profiles/edit.html.erb
AntonKhorev: Maybe we want something like this?
image

"Autofill %{country}" is probably not going to look good because it's likely going to be composed of two different languages.

AntonKhorev: and the button can become too large
image

nertc: Thanks for suggestions. I agree with you, it seems current solution has some UI problems, but adding additional disabled input doesn't align with the UI practices we are currently using. Maybe making it more minimalistic will be better:
We can remove button and have only following logic:
```
If user has empty `Location Name` field when entering `Profile Edit`:
    Until user changes `Location Name` field manually:
        Autofill will automatically fill `Location Name` field on pin change
Otherwise (if user has already saved `Location Name` and, therefore, field is filled, when entering `Profile Edit`):
    If `Location Name` field value exactly matches with the country name of the pin:
        Until user changes `Location Name` field manually:
            Autofill will automatically fill `Location Name` field on pin change
    Otherwise:
        Autofill will be deactivated
```

AntonKhorev: > If `Location Name` field value exactly matches with the country name of the pin

That requires fetching the country from Nominatim before the location changes, possibly on page load, right?

nertc: Yes, either on page load, or if we want to avoid an additional call on the page load, we can call Nominatim twice on the first change of the pin (one for the previous value and one for the new value, to compare them)

app/views/users/show.html.erb
nertc: `<title><%= t ".home location" %></title>` was added to the svg.

nertc:
Merge conflict was fixed and `master` was rebased.

app/controllers/geocoder_controller.rb
```diff
@@ -148,7 +148,11 @@ def search_osm_nominatim_reverse
                     :type => object_type, :id => object_id)
     end
 
-    render :action => "results"
+    if params[:format] == "json"
+      render :html => @results.pluck(:name).join(", ")
```
AntonKhorev, line 152: It doesn't output json when `params[:format] == "json"`, does it? Did you actually want plaintext?

nertc: PR was updated. Instead of HTML now `format=json` returns JSON array. I thought about plaintext, but I think, JSON object is much more versatile. Plus, it contains all the data that is contained in the html version (without `format=json`)

app/controllers/geocoder_controller.rb
```diff
@@ -148,7 +148,11 @@ def search_osm_nominatim_reverse
                     :type => object_type, :id => object_id)
     end
 
-    render :action => "results"
+    if params[:format] == "json"
+      render :json => @results
+    else
+      render :action => "results"
+    end
```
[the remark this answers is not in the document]
nertc: Code was changed to use `respond_to`.

nertc:
@tomhughes Thanks for the review. PR was updated accordingly.

app/assets/javascripts/user.js
```diff
@@ -131,14 +150,22 @@ $(document).ready(function () {
...
     try {
       if (lat && lon) {
         location = L.latLng(lat, lon);
+        if (updateLocationName) {
+          if (locationInput.checkLocation) {
+            updateHomeLocation(false, savedLat, savedLon, updateHomeLocation);
+          } else {
+            updateHomeLocation();
+          }
+        }
```
AntonKhorev, line 168: I put `catch` here because `L.latLng()` can throw on invalid coordinates. What is `updateHomeLocation` supposed to be able to throw?

app/assets/javascripts/user.js
```diff
@@ -176,6 +206,64 @@ $(document).ready(function () {
     }
   }
 
+  function updateHomeLocation(updateInput = true, lat = $("#home_lat").val().trim(), lon = $("#home_lon").val().trim(), successFn) {
```
AntonKhorev, line 209: The naming here... This function is called from `respondToHomeUpdate`. `respondToHomeUpdate` means that "Home" was updated and we're doing something in response to that update. And a possible part of the response is `updateHomeLocation`. Wasn't "Home" updated already?

app/models/user.rb
```diff
@@ -12,6 +12,7 @@
 #  home_lat             :float
 #  home_lon             :float
 #  home_zoom            :integer          default(3)
+#  location_name        :string
```
AntonKhorev, line 15: Why is it not `home_location_name`? You're sometimes referring to it as simply "home location", right?

app/assets/javascripts/user.js
AntonKhorev: Given https://github.com/openstreetmap/openstreetmap-website/pull/5302#discussion_r1989618279, I guess this is "updateHomeLocationName".

app/assets/javascripts/user.js
```diff
@@ -8,9 +8,19 @@
...
   if ($("#map").length) {
+    savedLat = $("#home_lat").val();
+    savedLon = $("#home_lon").val();
+    locationInput = {
+      dirty: false,
+      requestController: null,
+      countryName: $("#location_name").val().trim(),
+      deletedText: null,
+      checkLocation: savedLat && savedLon && $("#location_name").val().trim()
+    };
```
AntonKhorev, line 22: When a similar thing appeared in `directions.js` - endpoints with their own data structure and their own fetches - I moved them into their own module.

app/assets/javascripts/user.js
nertc: It shouldn't throw anything, but it should only be called if `L.latLng` doesn't fail. I can move it out from the `try` block, but it will add additional checks of `isNaN(lat)` and `isNaN(lon)`. Do you think it should be moved out?

app/assets/javascripts/user.js
nertc: `respondToHomeUpdate` was changed with `respondToHomeLatLonUpdate` and changing home location name now has easier to understand name `homeLocationNameEndpoint.updateHomeLocationName`.

app/assets/javascripts/user.js
nertc: I created a separate module named `home_location_name-endpoint` according to `directions-endpoint`. Functionalities directly related to the endpoint were moved accordingly.

nertc:
@AntonKhorev Thanks for the review. PR was updated accordingly.

app/views/users/show.html.erb
```diff
@@ -139,6 +139,16 @@
       <div class='text-body-secondary'>
         <small>
           <dl class="list-inline">
+            <% if @user.home_location_name && @user.home_location_name.strip.present? %>
```
tomhughes, line 142: This could just be `@user.home_location_name&.strip.present?`?

app/assets/javascripts/home_location_name-endpoint.js
```diff
@@ -0,0 +1,65 @@
+OSM.HomeLocationNameEndpoint = function Endpoint(latInput, lonInput, locationNameInput) {
```
tomhughes, line 1: I don't really get what `Endpoint` means in this name? Would somethinglike `HomeLocationGeocoder` be a better name maybe?

app/assets/javascripts/home_location_name-endpoint.js
```diff
@@ -0,0 +1,65 @@
...
+  const endpoint = {
+    dirty: false,
```
tomhughes, line 3: What does `dirty` mean here? What is the flag mean to do? This implementing the autofill logic described in https://github.com/openstreetmap/openstreetmap-website/pull/5302#discussion_r1897207842 I think? So maybe `autofillName` or something would be a better name?

app/views/users/show.html.erb
nertc: Was changed accordingly.

app/assets/javascripts/home_location_name-endpoint.js
nertc: `HomeLocationNameEndpoint` was chosen to be similar to `OSM.DirectionsEndpoint` from `app\assets\javascripts\index\directions-endpoint.js`.

Was renamed to `HomeLocationGeocoder`.

app/assets/javascripts/home_location_name-endpoint.js
nertc: Property name `dirty` was inspired from Angular. It means that input was changed manually. But, in this case, this property has more specific role and, I agree with you, name like `autofill` is more relevant.

`dirty` was changed to `autofill`.

nertc:
@tomhughes Thanks for the review. PR was updated accordingly.

nertc:
PR was updated. Merge conflicts were resolved.

app/controllers/geocoder_controller.rb
```diff
@@ -157,7 +157,11 @@ def search_osm_nominatim_reverse
                     :type => object_type, :id => object_id)
     end
 
-    render :action => "results"
+    respond_to do |format|
+      format.html { render :action => "results" }
+      format.json { render :json => @results }
+      format.any { render :action => "results" }
```
tomhughes, line 163: Presumably we don't really need the first line here if the render is just the same as the any case?

app/views/users/show.html.erb
```diff
@@ -143,6 +143,16 @@
       <div class='text-body-secondary'>
         <small>
           <dl class="list-inline">
+            <% if @user.home_location_name&.strip.present? %>
+              <dt class="list-inline-item m-0">
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-geo-alt-fill" viewBox="0 0 16 16">
```
tomhughes, line 148: I think I'd be inclined to add `align-bottom` to the classes here - it seems to give slightly better vertical centring of the icon relative to the text for me?

tomhughes:
I think this is looking good now - thanks for all the work on it!

PR #5303, opened 2024-11-05
Make the "remember me" option work as intended

tomhughes:
The "remember me" checkbox has not been working properly, or rather it has been treated as checked even when it isn't.

The reason is the "Gotcha" details in the [check_box documentation](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html#method-i-check_box) which means that the parameter has the value `"0"` when unticked.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -20,7 +20,7 @@ class SessionsController < ApplicationController
   end
 
   def create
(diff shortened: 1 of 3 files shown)


gravitystorm:
There's the integration tests that are now failing but it's not clear to me what's going wrong with them.

gravitystorm:
Merged, thanks!

PR #5304, opened 2024-11-06
Backfill note subscriptions

AntonKhorev:
Part two of  #5283.

Creates a subscription for every note commented by a non-anonymous user. Does this by:
- iterating through batches of note comments
- getting distinct non-null (user, note) pairs
- inserting these pairs using `upsert_all`, this skips already existing pairs


db/migrate/20241023004427_backfill_note_subscriptions.rb
```diff
@@ -0,0 +1,14 @@
+class BackfillNoteSubscriptions < ActiveRecord::Migration[7.2]
+  class NoteComment < ApplicationRecord; end
+  class NoteSubscription < ApplicationRecord; end
+
+  disable_ddl_transaction!
+
+  def up
+    NoteComment.in_batches(:of => 1000) do |note_comments|
+      rows = note_comments.distinct.where.not(:author_id => nil).pluck(:author_id, :note_id)
+      attrs = %w[user_id note_id]
```
tomhughes, line 10: There's no need to initialise this for every batch, it could go before the loop?

AntonKhorev: Moved `attrs = ...` outside, but I don't expect any difference about this compared to db reads/writes.

tomhughes:
This looks good now, thanks.

PR #5307, opened 2024-11-07
Replace query marker fade animation with css

### LINKED ISSUE #5305: Query highlight flickers when existing query is still loading
### URL

_No response_

Minor visual bug

1. Use Query features button at any location.
2. Click somewhere, then quickly click somewhere else while loading search results for the first query.
3. First circular area highlight loads, then fades away. Second circular area highlight however flickers and quickly fades away.

Chrome 130.0.6723.117, Edge 130.0.2849.68, Firefox 130.0 on Windows 10

_No response_

AntonKhorev:
Fixes #5305.

Currently query markers are animated using javascript which requires managing timed events. We don't do it correctly, we don't cancel them when markers are removed. That's why #5305 happens.

This PR replaces the javascript animation with the css which doesn't require any event handlers. I could have added animation end events to remove the marker layer but once faded out it doesn't affect anything and it's cleaned up in `page.unload`.

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -289,18 +289,10 @@ OSM.Query = function (map) {
       .hide();
 
     if (marker) map.removeLayer(marker);
-    marker = L.circle(latlng, radius, featureStyle).addTo(map);
-
-    $(document).everyTime(75, "fadeQueryMarker", function (i) {
-      if (i === 10) {
-        map.removeLayer(marker);
-      } else {
-        marker.setStyle({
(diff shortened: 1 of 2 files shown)


tomhughes (commented):
This is great, but despite the time being set the same it definitely seems to run faster to me? It's hard to actually time it but it seems like it might be close to twice as fast maybe because the javascript introduces extra delays running the function?

I think we should probably increase the time in the CSS to somewhere nearer 1500 to get closer to the same behaviour.

tomhughes:
This looks good now, thanks.

PR #1167, opened 2016-02-28
Allow the hiding of changeset tags

woodpeck:
This pull request establishes a simple mechanism for moderators to hide/unhide changset tags. All tags of a changeset are hidden, or none are hidden; there is no per-tag granularity. Tags are only hidden from the web interface, and not from the API. 
This is meant to be used in cases where people upload data with offensive changeset comments, or where advertising is placed in the changeset comment.

PR #5309, opened 2024-11-07
Delete changeset tags

### LINKED ISSUE #5219: Add the ability to redact changeset comments
### Problem

A user recently uploaded edits suggesting the objects are related to their female family members (details censored) in the title. DWG replied that they are unable to act on it.  

### Description

Upload comments, and changeset discussions, should be made redactable. This closes a privacy and safety loophole. 

There's an imaginable risk that users may be doxxed in replies when controversies fall to the worst from vandals. It should be removable. 

### Screenshots

_No response_

AntonKhorev:
Lets moderators delete changeset tags. A typical use case is somebody creating a changeset with a comment tag that says "my house" or "<person's name>'s house" which DWG is later asked to remove.

Fixes #5219.

Moderators will see the "Manage changeset tags" link at the bottom of the changeset page:
image

It leads to a page like this:
image

---

There was another attempt at tag hiding with a different approach #1167. It required database schema changes and marked the entire changeset as having its tags hidden. The tags were still kept in the database.

This PR doesn't change the schema and actually deletes the tags. Such deletions are not unusual and happen during regular changeset updates: https://github.com/openstreetmap/openstreetmap-website/blob/3a1b4cbe3396092efa2ea2c60cdbfb2a6ed6de0d/app/models/changeset.rb#L176


AntonKhorev:
> The big question for me is how updates are propagated to the changeset replication feed

This is one of the excuses why hiding/deleting changeset tags wasn't done before. There are others such as we don't have changeset versions to store previous set of tags.

If the feed is generated by [this](https://github.com/openstreetmap/chef/blob/c8b4fa19916eb02e6b3861216cfeb3b4f7a4f5b5/cookbooks/planet/files/default/replication-bin/replicate-changesets), tag deletions probably won't be propagated. But we can automatically add a changeset comment that some tag was deleted, this should get picked up by the feed.

gravitystorm:
> This is one of the excuses why hiding/deleting changeset tags wasn't done before.

I'd call it a 'good reason', not an 'excuse'. Having a user interface to remove or edit tags is straightforward, but making sure the updates are properly replicated is required too.

AntonKhorev:
I've decided against automatically adding a comment to trigger the replication. Surely moderators can add a comment after deleting tags. Or they can choose not to add a comment if they don't want to replicate the changes, although that's less likely.

AntonKhorev:
If you delete a tag and comment the changeset, that replicates the changeset and the deleted tag is not included.

AntonKhorev:
Changed to a singular resource in order to have `destroy` action, was using nonstandard `delete` action on a collection before.

woodpeck:
I know that we should ideally separate the software from operational requirements. The software should not be monkey-patched just because its users have a pressing need. However (:smile:), the DWG is facing a rising number of requests where people want their original changeset comment modified or hidden and this currently requires someone with raw database access rights to fiddle with the SQL - an action that will leave no trace and certainly not replicate to downstream users. In some cases the data protection case is strong enough that the mappers would doubtless succeed in court against the OSMF - we've just been lucky until now that nobody was bellicose enough. 

(Yes, it is unfortunate that mappers are thoughtless enough to enter stuff they later want hidden, but the law is on their side and probably for good reason.)

Anton's implementation leaves it up to the moderator who hides a changeset tag; the moderator can write a changeset comment saying "I have hidden something here for XY reason", which will lead to the whole changeset metadata re-entering the replication feed and being properly replicated downstream - or the moderator can decide not to do that in which case nothing will be replicated. I think this, while perhaps not perfect, is good enough to go and its deployment would relieve a lot of pressure on the moderators and database admins.

Thank you for considering it.

1ec5:
> the moderator can write a changeset comment saying "I have hidden something here for XY reason", which will lead to the whole changeset metadata re-entering the replication feed and being properly replicated downstream - or the moderator can decide not to do that in which case nothing will be replicated.

If there’s concern that the moderator would forget to post the comment, a message confirming that the deletion was successful could include a button to post a standard comment to the changeset. Or the site could do it automatically, similar to what some third-party tools do when reverting a changeset.

PR #5310, opened 2024-11-08
Send notifications to note subscribers instead of commenters

AntonKhorev:
Part 3 of #5283. Start using the subscriptions table to send notifications.


tomhughes:
Looks good to me, thanks.

PR #5312, opened 2024-11-11
Don't update deactivates_at if block was already viewed

AntonKhorev:
When a blocked user visits their block page, its deactivation time is recorded. Problem is this only needs to happen if there's no deactivation time set. If it's already set, it should be kept at what it is.

The possible bug is that if the user visits their block page after the block is lifted, that visit time gets into `deactivates_at` and is reported as the block end time despite the block ending earlier.

### DIFF
--- a/test/controllers/user_blocks_controller_test.rb
+++ b/test/controllers/user_blocks_controller_test.rb
@@ -147,14 +147,72 @@ class UserBlocksControllerTest < ActionDispatch::IntegrationTest
     assert_select "h1 a[href='#{user_path active_block.user}']", :text => active_block.user.display_name
     assert_select "h1 a[href='#{user_path active_block.creator}']", :text => active_block.creator.display_name
     assert UserBlock.find(active_block.id).needs_view
+  end
 
-    # Login as the blocked user
-    session_for(active_block.user)
+  ##
+  # test clearing needs_view by showing a zero-hour block to the blocked user
+  def test_show_sets_deactivates_at_for_zero_hour_block
+    user = create(:user)
+    session_for(user)
 
-    # Now viewing it should mark it as seen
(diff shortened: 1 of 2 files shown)


test/controllers/user_blocks_controller_test.rb
```diff
@@ -147,14 +147,72 @@ def test_show
...
+  end
+
+  ##
+  # test clearing needs_view by showing a timed block to the blocked user
+  def test_show_sets_deactivates_at_for_timed_block
+    user = create(:user)
+    session_for(user)
+
+    freeze_time do
+      block = create(:user_block, :needs_view, :created_at => Time.now.utc, :ends_at => Time.now.utc + 1.day, :user => user)
```
tomhughes, line 188: I recently got rid of all the `.day` uses in this file because they cause confusion when running the tests just before a DST change when a day may not be 24 hours so this should probably be `24.hours` instead.

test/controllers/user_blocks_controller_test.rb
```diff
@@ -147,14 +147,72 @@ def test_show
...
+
+      travel 1.hour
+
+      get user_block_path(block)
+      assert_response :success
+      block.reload
+      assert_not block.needs_view
+      assert_equal Time.now.utc + 22.hours, block.deactivates_at
+
+      travel 1.day
```
tomhughes, line 208: Same here...

tomhughes (changes requested):

test/controllers/user_blocks_controller_test.rb
AntonKhorev: updated, but if it's actually an issue, `test/factories/user_blocks.rb` still uses days

tomhughes (approved):
Looks good now, thanks.

PR #5314, opened 2024-11-12
Note subscriptions API

AntonKhorev:
Part 4 of #5283

Adds API endpoints to subscribe/unsubscribe from a note.

This is like the changeset subscribe/unsubscribe
https://wiki.openstreetmap.org/wiki/API_v0.6#Subscribe:_POST_/api/0.6/changeset/#id/subscribe
but since we like resourceful routes, instead of `POST .../subscribe` and `POST .../unsubscribe`, this PR adds:

- `POST /api/0.6/notes/:id/subscription` to add a subscription
- `DELETE /api/0.6/notes/:id/subscription` to remove a subscription

### DIFF
--- a/app/controllers/api/note_subscriptions_controller.rb
+++ b/app/controllers/api/note_subscriptions_controller.rb
@@ -1,0 +1,18 @@
+module Api
+  class NoteSubscriptionsController < ApiController
+    before_action :check_api_writable
+    before_action :authorize
+
+    authorize_resource
+
+    def create
+      note_id = params[:note_id].to_i
+      note = Note.find(note_id)
+      note.subscribers << current_user
+    rescue ActiveRecord::RecordNotFound
+      report_error "Note #{note_id} not found.", :not_found
+    rescue ActiveRecord::RecordNotUnique
+      report_error "You are already subscribed to note #{note_id}.", :conflict
+    end
+  end
+end
--- a/app/controllers/api/note_subscriptions_controller.rb
+++ b/app/controllers/api/note_subscriptions_controller.rb
@@ -14,5 +14,14 @@ module Api
     rescue ActiveRecord::RecordNotUnique
       report_error "You are already subscribed to note #{note_id}.", :conflict
     end
+
+    def destroy
+      note_id = params[:note_id].to_i
+      note = Note.find(note_id)
+      count = note.subscriptions.where(:user => current_user).delete_all
+      report_error "You are not subscribed to note #{note_id}.", :not_found if count.zero?
+    rescue ActiveRecord::RecordNotFound
+      report_error "Note #{note_id} not found.", :not_found
+    end
   end
 end
--- a/test/controllers/api/note_subscriptions_controller_test.rb
+++ b/test/controllers/api/note_subscriptions_controller_test.rb
@@ -1,0 +1,77 @@
+require "test_helper"
+
+module Api
+  class NoteSubscriptionsControllerTest < ActionDispatch::IntegrationTest
+    def test_routes
+      assert_routing(
+        { :path => "/api/0.6/notes/1/subscription", :method => :post },
+        { :controller => "api/note_subscriptions", :action => "create", :note_id => "1" }
+      )
+    end
+
+    def test_create
+      user = create(:user)
+      auth_header = bearer_authorization_header user
+      note = create(:note_with_comments)
+      assert_empty note.subscribers
+
+      assert_difference "NoteSubscription.count", 1 do
+        assert_difference "note.subscribers.count", 1 do
+          post api_note_subscription_path(note), :headers => auth_header
+          assert_response :success
+        end
+      end
+      assert_equal user, note.subscribers.last
+    end
+
+    def test_create_fail_anonymous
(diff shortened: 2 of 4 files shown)


app/controllers/api/note_subscriptions_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    before_action :authorize
+
+    authorize_resource
+
+    def create
+      note_id = params[:note_id].to_i
+      note = Note.find(note_id)
+      note.subscribers << current_user
+    rescue ActiveRecord::RecordNotUnique
+      head :conflict
```
tomhughes, line 13: Is `conflict` right here? This happens if you try and subscribe to a note you're already subscribed to right?

test/controllers/api/note_subscriptions_controller_test.rb
```diff
@@ -0,0 +1,132 @@
...
+      create(:note_subscription, :user => user, :note => note)
+      assert_equal [user], note.subscribers
+
+      assert_difference "NoteSubscription.count", -1 do
+        assert_difference "note.subscribers.count", -1 do
+          delete api_note_subscription_path(note), :headers => auth_header
+          assert_response :success
+        end
+      end
+    end
```
tomhughes, line 92: Maybe assert that `note.subscribers` is empty here? To mirror what the creation test does?

app/controllers/api/note_subscriptions_controller.rb
AntonKhorev: `conflict` is what changeset subscribe does in this case, I'm doing the same here

https://wiki.openstreetmap.org/wiki/API_v0.6#Subscribe:_POST_/api/0.6/changeset/#id/subscribe

test/controllers/api/note_subscriptions_controller_test.rb
AntonKhorev: added

app/controllers/api/note_subscriptions_controller.rb
tomhughes: OK but that does raise a custom error that includes a nice message explaining the problem - should we do that here? and for the not found case on unsubscription.

I'm not really sure why changesets do it by throwing an exception - a custom message could be just as easily returned inline.

AntonKhorev: added error messages

gravitystorm:
Looks good to me. Only two small things:
* PR contains a fixup commit (i.e. a commit that changes code that itself is added by this PR)
* I'm not sure if the error messages should come from the translation system. Other usage of `report_error` is inconsistent whether error strings are translated or not.

AntonKhorev:
> PR contains a fixup commit

Let me guess what's a *fixup* here. The fact that looking up the note is not required in `destroy` if there aren't two different error messages (*note not found* or *not subscribed*)? Now I changed it to always look up the note. Is there still a fixup?

gravitystorm:
No, that's not what I meant. A "fixup commit" is a commit that's added to a pull request that fixes problems with code introduced in the same pull request.

From CONTRIBUTING.md (emphasis added)

> Avoid including "fixup" commits. If you have added a fixup commit (for example to fix a rubocop warning, or **because you changed your own new code**) please combine the fixup commit into the commit that introduced the problem. `git rebase -i` is very useful for this.

In this PR, the first commit includes:

```diff
+    rescue ActiveRecord::RecordNotUnique
+     head :conflict
```

The third commit includes

```diff
   rescue ActiveRecord::RecordNotUnique
-      head :conflict
+      report_error "You are already subscribed to note #{note_id}.", :conflict
```

Since the `head :conflict` is added and then removed, the PR should be rebased so that that code isn't mentioned in the history (for example, when using `git blame`.

AntonKhorev:
There weren't any *problems* with the code. Not having an error message is not a *problem*. You could as well have said that not adding `destroy` together with `create` is a problem because it requires changing a.k.a. fixing routes and abilities.

AntonKhorev:
Changed to have error messages right away.

The original version had response bodies intentionally undefined, and they are still intentionally undefined for ok responses.

gravitystorm:
Merged, thanks for the updated version.

I'm interested to hear any thoughts on the translations:

> "I'm not sure if the error messages should come from the translation system. Other usage of report_error is inconsistent whether error strings are translated or not."

We can always refactor this separately, since it's not just this code that would be affected.

AntonKhorev:
Most of error messages are not translated. `report_error` receives many from exceptions defined in `lib/osm.rb`, no translations there.

nenad-vujicic:
> ... A "fixup commit" is a commit that's added to a pull request that fixes problems with code introduced in the same pull request.
> 
> From CONTRIBUTING.md (emphasis added)
> 
> > Avoid including "fixup" commits ...

Is catching "fixup" commits (a line added (removed) in one commit and then removed (added) in a subsequent commit within the same PR) using a `danger workflow` something that could help maintainers? It’s a bit tricky to define what constitutes a "fixup" because, if a commit represents a logical unit, a line could have meaning in one commit but not necessarily in another, thus justifying its removal.

PR #5311, opened 2024-11-08
Restore loading spinner in richtext previews

AntonKhorev:
Richtext preview loading animation was added in https://github.com/openstreetmap/openstreetmap-website/commit/eb789dbf757e760e8348e0aa75343eb53996b7e3 but later probably got broken. Its `loading` css class was removed in https://github.com/openstreetmap/openstreetmap-website/commit/df1ec6b6809b533829fc2c171699540bbb92a85b but javascript still adds/removes it to no effect.

This PR adds a Bootstrap spinner to the preview pane.

image

PR #5315, opened 2024-11-13
Use CSS animations to show delayed loading spinner

AntonKhorev:
Similarly to #5307, we can avoid scheduling/unscheduling js events to show spinners in the richtext preview #5311 and the sidebar loader:

image

There's a delay to avoid blinking on fast loads, that delay can be implemented using a css animation.

And we also can stop using `vendor/assets/jquery/jquery.timers.js` which does some ill-advised things like [listening to `unload`](https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event).

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -25,8 +25,6 @@
 //= require qs/dist/qs
 
 $(document).ready(function () {
-  var loaderTimeout;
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good to me, thanks.

PR #5317, opened 2024-11-13
Show "node also part of ways" as nested lists on way pages

### LINKED ISSUE #1147: Website element display doesn't know about closed ways ...
Super serious, major major, must be fixed tomorrow issue :-).

On the other hand it has likely been aournd since we have the rails port so I suppose we could just continue to ignore it.

See for example http://www.openstreetmap.org/way/189128109 node #3984033658 is the end and start node of the adjacent polygon. 

Don't ask why I noticed this in the first place. If I'm not too lazy I might submit a PR:

AntonKhorev:
Way pages have lists of nodes. Every way node can be shared with other ways, and if it is, then all of those ways are also listed. Currently it's done as "(part of ways ...)" phrases with ways listed using `to_sentence`. Those are cumbersome to deal with. You can never be sure about grammar in translations. Also it's more difficult to style because each way may or may not have an icon, there's a space reserved for that icon, that icon has to be positioned into that space probably with a bunch of pixel values. That's what #5080 does and that's one thing I don't like about it.

In this PR I convert those phrases with inline icons into nested lists. After that icons are going to be located at the beginning of `<li>`s both for nested lists of ways and outer lists of nodes. This will make icons easier to position.

Also fixes #1147. Inner lists of ways contain only unique ways to avoid https://github.com/openstreetmap/openstreetmap-website/issues/1147#issuecomment-2189648596. The outer list of all nodes has details collapsed for any node that was already written (see node 2071 on screenshots below).

Before:
image

After:
image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -340,8 +340,8 @@ en:
         one: "%{count} node"
         other: "%{count} nodes"
       also_part_of_ways:
-        one: "also part of way:"
-        other: "also part of ways:"
+        one: "also part of %{count} other way"
+        other: "also part of %{count} other ways"
     relation:
       title_html: "Relation: %{name}"
       history_title_html: "Relation History: %{name}"
--- a/app/views/browse/_way.html.erb
+++ b/app/views/browse/_way.html.erb
@@ -25,6 +25,7 @@
       <details <%= "open" if way.way_nodes.count < 10 %>>
         <summary><%= t ".nodes_count", :count => way.way_nodes.count %></summary>
         <ul class="list-unstyled">
+          <% visited_way_nodes = {} %>
           <% way.way_nodes.each do |wn| %>
             <li>
               <% related_ways = wn.node.ways.uniq.sort.reject { |w| w.id == wn.way_id } %>
@@ -34,8 +35,8 @@
                 <div>
                   <%= element_single_current_link "node", wn.node %>
                 </div>
-                <div class="ms-4">
-                  <%= t ".also_part_of_ways", :count => related_ways.size %>
+                <%= tag.details :class => "ms-4", :open => !visited_way_nodes[wn.node.id] do %>
+                  <summary><%= t ".also_part_of_ways", :count => related_ways.size %></summary>
                   <ul class="list-unstyled">
                     <% related_ways.each do |related_way| %>
                       <li>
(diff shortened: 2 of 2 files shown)


config/locales/en.yml
```diff
@@ -338,9 +338,9 @@ en:
       nodes_count:
         one: "%{count} node"
         other: "%{count} nodes"
-      also_part_of_html:
-        one: "part of way %{related_ways}"
-        other: "part of ways %{related_ways}"
+      also_part_of_ways:
+        one: "also part of %{count} way"
+        other: "also part of %{count} ways"
```
tomhughes, line 343: Should these be `... other way(s)` here? It's not really grammatically correct at the moment given the count exclude one way that the node is a member of.

tomhughes (commented):
I think my main question here is whether we should start these (mostly) expanded or whether they should all be collapsed? Starting them expanded does mean that you're using more vertical space and reducing what top level information can be seen initially?

config/locales/en.yml
AntonKhorev: I had it in one of the versions, but then I decided that maybe the messages get too long. Put it back.

AntonKhorev:
About starting (not) expanded and vertical space:

The first obvious thing that I came up this is to use roughly the same logic other `<detail>` elements are open/closed by default. Way nodes has this line `<details <%= "open" if way.way_nodes.count < 10 %>>` which expands way nodes if there are fewer than 10 of them. So I thought I'd expand node ways if there's only one other way for a node. Collapsing just one way won't save much space.

But then I checked some actual ways in osm and looks like their nodes in most of the cases are members of one or two ways. Other cases are relatively rare. If I do what I described in the previous paragraph, I wouldn't have to collapse almost any `also part of` list. So collapsing 1-way lists doesn't save much vertical space, and collapsing many-way lists also doesn't save much vertical space. What does save it then?

I expect most of space wasted in this case:
- there's a long way with many nodes
- nodes mostly don't have any tags
- parts of the way share nodes with other ways, forming sequences of nodes with the same type (where type is "nothing special", just show the generic node icon), name (actually no name, just show the id) and set of shared ways

Here's an example that has two such sequences with one more special node in between:

image

The idea is then to collect every such sequence into one list element, and that's what #5355 does.

PR #5319, opened 2024-11-13
Minor changes to docs

gravitystorm:
This PR combines a few minor changes to the docs that I noticed while working on #5318 

* Add missing top-level headers to two files
* Standardise on fenced code blocks in markdown, for consistency and to allow language declarations
* Use language declarations for yaml code blocks

I also considered adding language information for code intended to be typed at the console, but then that turns out to be more complex with a choice of declarations (e.g. `bash` vs `console`) and whether to add `$ ` as a guide that it should be typed, and what to do with things that start with a console command but is then mainly `irb` from that point on, so I'll leave all that for some other time.

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -76,7 +76,7 @@ For iD, do the following:
 
 An example excerpt from settings.local.yml:
 
-```
+```yaml
 # Default editor
 default_editor: "id"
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me, thanks.

PR #5316, opened 2024-11-13
Remove unused timer from new note js controller

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/commit/51dcf86f40200bf4728097868f45aa2fe12968f3 added the "removenote" timer. Its use was removed in https://github.com/openstreetmap/openstreetmap-website/commit/d380ce79aabcd98ade3eabe8f982e516311f6aee but the `stopTime` cleanup stuck.

PR #5320, opened 2024-11-14
Remove jQuery.timers plugin

AntonKhorev:
After merging #5315 and #5316 this jQuery plugin is not used anywhere.

As I say in #5315 it listens to unload event. The version we're using is old, there are several newer forks but they all also listen to unload.


tomhughes:
Looks good to me, thanks.

PR #5330, opened 2024-11-15
Apply dimming filter to tiles directly

AntonKhorev:
In dark mode tiles are actually darker than I intended them to be:
image vs image

See https://github.com/openstreetmap/openstreetmap-website/pull/5325#issuecomment-2478943023 and https://github.com/openstreetmap/openstreetmap-website/pull/5325#issuecomment-2479026722 for explanations.


gravitystorm:
Merged, thanks!

Issue #5329, opened 2024-11-15
Dark Mode bugs/improvements for the UI


gravitystorm:
I want to have an issue here for people to report bugs with the dark mode as it affects the user interface.

For this issue, any discussion of toggles, user preferences, or the maps are off-topic.

For example, here's one thing I've spotted and we can discuss - the buttons for third-party sign-ins look a bit off, since they maintain the same light grey background as light mode:

![Screenshot from 2024-11-15 12-24-17](https://github.com/user-attachments/assets/902c837c-159b-466a-91eb-a992c5f51988)

If you spot other things like this with the dark mode, please comment below (or even better, make a PR!)

deevroman:
https://www.openstreetmap.org/traces

You might want to change the text color on a bright green background. For the example I changed the color to black and it seems to read better

<img width="976" alt="Снимок экрана 2024-11-15 в 16 06 38" src="https://github.com/user-attachments/assets/6ae37ad7-5c54-451a-849d-e88be9f3d871">

Nekzuris:
Readability is bad due to poor contrast with blue.
![image](https://github.com/user-attachments/assets/e3e25e08-2733-4e78-ab47-1beea9ab753d)

Dimitar5555:
The numbers on profile pages are hard to read.

![image](https://github.com/user-attachments/assets/f2dd05ff-047d-4188-9e5b-1fe3f7acbc95)

---

The popovers (or whatever these balloons are called) should have dark background, right?
![image](https://github.com/user-attachments/assets/e45759a7-b55d-4fb0-9dd1-726136f5a926)

---

Some items in the map legend are hard or almost impossible to see, others have distorted colours (major roads for example). The screenshot is from zoom level 17.
![image](https://github.com/user-attachments/assets/be6cdeed-16a8-4529-9f19-dc87c14636d5)

---

The buttons on each message in the inbox and outbox should probably use outline buttons. 
![image](https://github.com/user-attachments/assets/6a0d79ec-5404-4fbc-aba7-6b12b3c4205e)

Nekzuris:
Contrast is a bit too low in the topbar, even in light mode. Chrome shows green at 4.5 but I think it should be even higher since this text is important.
![image](https://github.com/user-attachments/assets/909efbe8-e8ce-402b-9296-1c7bc08101b4)
![image](https://github.com/user-attachments/assets/cd30b217-83b8-4d89-95ce-9a5ecf729d54)

[...]

PR #5331, opened 2024-11-15
Set colors for trace visibility badges using .text-bg-*

AntonKhorev:
As recommended in https://getbootstrap.com/docs/5.3/components/badge/#background-colors, this sets both background and text colors in one declaration, with text depending on background.

See https://github.com/openstreetmap/openstreetmap-website/issues/5329#issuecomment-2478795244

image

image


gravitystorm:
Merged, thanks! I hadn't seen this approach before.

PR #5332, opened 2024-11-15
Set background for layer buttons to secondary body color

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/issues/5329#issuecomment-2478786176

Before:
image image

After:
image image


gravitystorm:
Merged, thanks!

PR #5333, opened 2024-11-15
Style small auth buttons similarly to preferred auth buttons

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/issues/5329#issue-2661831842

Before:
image
image

After:
image
image


gravitystorm:
Nice fix, thanks!

PR #5336, opened 2024-11-16
Dark mode menu

AntonKhorev:
Dark mode filter menu from #4777 but without dark transport map tiles.

image


app/assets/javascripts/index/contextmenu.js
```diff
@@ -74,6 +74,66 @@ OSM.initializeContextMenu = function (map) {
...
+      name: "brightness80",
+      filter: "brightness(.8)"
+    },
+    {
+      name: "brightness60",
+      filter: "brightness(.6)"
+    },
+    {
+      name: "invert",
+      filter: "invert(.8) hue-rotate(180deg)"
```
scy, line 100: I think .8 has pretty weak contrast and would suggest .95 instead. Or provide several options, like .8, .9, .95 and 1.

mxdanger:
Will 100% brightness be the default? From what I understand to be the sole criticism (mine included) of the initial introduction dark mode was the dimmed map.

tomhughes:
Seriously if we have to do this then we should just give up - it's not so much user friendly as user hostile to be giving space to something like this that is going to be completely incomprehensible to most users.

PR #5337, opened 2024-11-16
Color mode preference

AntonKhorev:
A quick implementation of auto/light scheme switch for those who want to stay on the light scheme even when their browser says they want the dark one.

image
image
image

Its on the `/preferences` page but it's not actually a preference stored in the `User` model. It's stored in a cookie. The question is do we want to store it in a cookie? Maybe we do because what browser reports as a preferred scheme is per browser, not per user, then perhaps its override should also be per browser. And it will be easier to extend the availability of this option to anonymous users. It is kind of available already, if they know how to edit cookies.

### DIFF
--- a/app/assets/stylesheets/parameters.scss
+++ b/app/assets/stylesheets/parameters.scss
@@ -20,4 +20,3 @@ $table-border-factor: .1;
 $list-group-hover-bg: rgba(var(--bs-emphasis-color-rgb), .075);
 
 $enable-negative-margins: true;
-$color-mode-type: media-query;
--- a/app/assets/stylesheets/screen-auto-ltr.scss
+++ b/app/assets/stylesheets/screen-auto-ltr.scss
@@ -1,0 +1,3 @@
+@use "common" with (
+  $color-mode-type: media-query
+);
--- a/app/assets/stylesheets/screen-auto-rtl.rtlcss.scss
+++ b/app/assets/stylesheets/screen-auto-rtl.rtlcss.scss
@@ -1,0 +1,3 @@
+@use "common" with (
+  $color-mode-type: media-query
+);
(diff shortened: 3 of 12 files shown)


AntonKhorev:
But should it be in user preferences?

AntonKhorev:
And if stored in the database, should it be in `User` or `UserPreference`? This one depends on how do we want to make it available to other apps.

hlfan:
I don't think this preference mashup between local and database storage is a great idea.

But if logged-in users have the preference synchronized and non-logged-in users have the color and language (editor choice seems pointless for non-logged-in users) stored in the cookies it could be interesting.

Or just decuple the local settings from the synced ones.

AntonKhorev:
By making the setting available to other apps I meant iD of course. Its color scheme preferably should match the entire site, that means it has to know somehow whether to use `prefers-color-scheme` or ignore it. Let's suppose we made this setting somehow readable through the api. iD will look at it and adjust its color scheme accordingly.

Now let's suppose we don't use iD embedded in the osm-website. We'll run it from https://ideditor-release.netlify.app/, It can still read the setting in the api, but the reason to apply it is gone because iD is no longer embedded. Then maybe it's better to pass it as a startup parameter through
https://github.com/openstreetmap/openstreetmap-website/blob/22942b1ca144444bdece1636451fd3ad86530b3e/app/views/site/_id.html.erb#L4
In this case we wouldn't need to worry how exactly we store the setting.

hlfan:
How would not logged-in users change this setting?

gravitystorm:
I would prefer #5339

gravitystorm:
Closing this, since it was an alternative to #5339 / #5362 which is now merged.

PR #5338, opened 2024-11-17
Enable dark mode attribution font color

### LINKED ISSUE #5329: Dark Mode bugs/improvements for the UI

hlfan:
### Description
This PR increases the specificity of the dark mode attribution selector. The style was first added in #4764, but didn't overwrite Leaflet's default link selector. That increased specificity fixes #5329's bug report about this by @Nekzuris

| before | after |
|-|-|
| image | image |

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -507,7 +507,7 @@ body.small-nav {
     filter: brightness(.8);
   }
 
-  .leaflet-control-attribution a {
+  .leaflet-container .leaflet-control-attribution a {
     color: var(--bs-link-color);
   }
 


AntonKhorev:
This was supposed to be working already #4764 but something broke it.

AntonKhorev:
The specificity is the same as in Leaflet css, then it depends on the order but I don't see where we changed the order. It would have been in `app/views/layouts/_head.html.erb`.

AntonKhorev:
It's probably Sass's `@include color-mode(dark)` generating an extra selector in `$color-mode-type: data` and not generating it in `$color-mode-type: media-query`, resulting in lower specificity.

Issue #5324, opened 2024-11-14
Add light/dark mode preference


elevenchars:
### Problem

Browsers do not let you set a per-site dark mode preference. Some users do not want to use the new dark mode.

### Description

A toggle similar to the Wikipedia appearance (see below). It should probably be accessible for users who are not logged in, as well.

### Screenshots

![image of Wikipedia's "appearance" menu](https://github.com/user-attachments/assets/9621932c-375e-4178-aec8-409e1e7b84c7)

DaveF63:
The tile screen didn't need dimming; it makes it difficult to read.

TurnrDev:
I agree, there needs to be a config option.

mxdanger:
There should be absolutely no image or map dimming. It just reduces readability.

Just leaving this comment here so an option is **NOT** created for it. The dimming shouldn’t be a thing.

nickvet419:
It darkens the map background, which also dims the map. and no option to disable dark mode on the page.

kartomonkey:
Having a dark menu bar is nice, but simply dimming the map is just not acceptable. Maybe as a first step towards a truly dark mode, but please, please, please make it easy for the user to switch back to light mode.
The toggle is a must!

gravitystorm:
Please keep discussions here related to having an option to enable or disable dark mode. The exact implementation of dark mode for the maps is off-topic for this issue.

hlfan:
> related to having an option to enable or disable dark mode

Just one option? Why not more?
![www openstreetmap org](https://github.com/user-attachments/assets/864749a6-a3b9-4e96-bd44-01b9aecfe030)

pnorman:
> Some users do not want to use the new dark mode

This would only apply to users who want to use dark mode in general, but not on OSM. Is this a large body of users? Could it be reduced with improvements to dark mode?

TurnrDev:
> Could it be reduced with improvements to dark mode?

For myself, I value the ability to turn off dark mode for individual sites entirely. When working on maps, I often cross reference images, notes and other information, most of which don't have a "dark mode" so the switching back and forth is jarring.

pnorman:
> For myself, I value the ability to turn off dark mode for individual sites entirely. When working on maps, I often cross reference images, notes and other information, most of which don't have a "dark mode" so the switching back and forth is jarring

I would have thought this would result in you turning off dark mode in general. What I'm interested in is why you want dark mode for everything else, but not for OSM.

[...]

PR #5339, opened 2024-11-17
Color mode preference stored in db

AntonKhorev:
The same as #5337 but stores the preferred color scheme value in user preferences db table instead of cookies.

### DIFF
--- a/app/assets/stylesheets/parameters.scss
+++ b/app/assets/stylesheets/parameters.scss
@@ -20,4 +20,3 @@ $table-border-factor: .1;
 $list-group-hover-bg: rgba(var(--bs-emphasis-color-rgb), .075);
 
 $enable-negative-margins: true;
-$color-mode-type: media-query;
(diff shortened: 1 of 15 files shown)


gravitystorm:
I think the preference should have three options - auto (default), light and dark. In auto mode, either the browser color-mode should be used, or if that's not available, we should use light mode.

See https://github.com/openstreetmap/openstreetmap-website/issues/5324#issuecomment-2504102012 for more details.

AntonKhorev:
I also think that it should have three options. The question is do you want a larger pull request that adds three options or do you want a smaller one (you usually say you want small pull requests) that lets you disable the auto dark mode?

gravitystorm:
I think if we both agree to add a new preference with 3 options, there's no need to split that into different pull requests for each option! :smile:

AntonKhorev:
Updated with dark option

PR #5327, opened 2024-11-15
Remove brightness filter for map tiles in dark mode styles

sammyhawkrad:
### Description
It's nice to finally have a dark mode on the website but reducing the brightness of the default map tiles is not ideal and negatively affects the user experience. This PR removes the filter on the map tiles to keep the default view of the map.

Sets the dark mode view to this:
image

from this:
image

I also tested the brightness at .95 which is quite bearable if the dimming is necessary.
image

PR #5340, opened 2024-11-18
Remove brightness filter for map tiles in dark mode

Nekzuris:
Closes #5327 
See #5328 and https://github.com/openstreetmap/openstreetmap-website/issues/5329#issuecomment-2482070226


AndrewKvalheim:
Is this not an exact duplicate of #5327?

hlfan:
Isn't this also superseded by #5507?

tomhughes:
I believe so, yes.

PR #5323, opened 2024-11-14
Adding map note tags - part 1 - added migration script and model files

nenad-vujicic:
### Description

Added migration for creating `note_tags` DB table, created model file `NoteTag.rb`, updated associations with `Note` class and added `note_tag` factory and unit tests `NoteTagTests` for testing basic functionalities (key/value lengths validity, key/value lengths invalidity, orphaned tags invalidity and note-tags uniqueness).

This PR is first step of adding support for map note tags described in #5294 and which fully implementation can be found [here](https://github.com/nenad-vujicic/openstreetmap-website/tree/issue_5294_poc) (decomposed to smaller because of simpler review).

### How has this been tested?

Tested by running linters from `lint.yml` workflow, unit tests and by manually inserting note-tags manipulations and as part of [complete implementation](https://github.com/nenad-vujicic/openstreetmap-website/tree/issue_5294_poc) locally.

PR #5344, opened 2024-11-19
Adds non-mutable note tags support

### LINKED ISSUE #5294: Proposal for adding support for map notes tags
### Problem

Do you have some key-value meta-data you would like to store with map notes? Would you like to do it like with tags in nodes / ways / relations / changesets / .. instead of writing in description / comments? Note tags have already been mentioned / requested several times in #385 , #801 , #3932 and similar. Here we expose our plan about how we could add them.

### Description

Here is a list of PRs (steps) we would like to work on with rough description of what they will do. We would highly appreciate your feedback and suggestions on how we could improve the plan:

1. Create migration script for creating note-tags table:
   - create new table **note_tags** with note_id (bigint(8)), k (string, default “”, not-null) and v (string, default(“”), not-null) attributes
   - **(note_id, k)** will be set as private key and **note_id** will be foreign key referring id attribute from **notes** table
2. Add loading and displaying note tags using **browse/tag_details** partial
   - Add displaying of note tags using **browse/tag_details** partial between Description (currently "special first comment") and Comments (remaining comments)
   - Add / improve new / existing unit tests for testing new functionalities
3. Update Notes API to include retrieving note tags
   - Update **(j)builder** / **API::NotesController** / **Note model** files to include note tags in generated files
      It will insert **<tag k="" v=""/>** pairs between **ID** and **URL** like in [this example](https://wiki.openstreetmap.org/wiki/API_v0.6#:~:text=api/0.6/notes-,%3C%3Fxml%20version%3D%221.0%22%20encoding%3D%22UTF%2D8%22%3F%3E%0A%3Cosm%20version%3D%220.6,html%3E%0A%09%09%09%3C/comment%3E%0A%09%09%09...%0A%09%09%3C/comments%3E%0A%09%3C/note%3E%0A%09...%0A%3C/osm%3E,-Response%20JSON)
   - Add / improve new / existing unit tests for testing new functionalities
4. Update Notes API to include creating Note instances with tags
   - Update **API::NoteController** to create Note instances with tags
   - Add / improve new / existing unit tests for testing new functionalities
5. Improve existing and add new unit tests for testing new features
   - Add more stressful unit tests for testing new features 
6. Update Wiki documentation about OSM API v0.6 section Map Notes (this will be more of a side task than a PR)

### Screenshots

image

nenad-vujicic:
### Description

PR adds support for **non-mutable** note tags as described in #5294 Following changes are made:

1. Created `note_tags` table, connected it with note table (using foreign key and associations), created `NoteTag` model file and `NoteTagTest` class with basic tests (PR #5323 does this)
2. Added preparing and rendering of note tags in left sidebar using `browse/tag_details` partial, also, created test with multiple tags added rendering and checking if rendered HTML contains tags
3. Updated (j)builder files to write note tags in XML, JSON, GPX, RSS and feed files and updated test to check existence of tags in each of above files (creation of tag is performed using ActiveLayer)
4. Added support for creating tags at note creation time using POST HTTP request and improved tests to check existence of tags in each of generated files (creation of tag is performed using POST HTTP request). Also, added creating tag `created_by, OpenStreetMaps-Website` for notes created using OSM website.

Displayed resolved note:
image

Displayed opened note:
image

XML output for "opened note":
image

Displayed note without tags (all current):
image 

### How has this been tested?

Run tests from workflows, tested creating and displaying notes manually as administrator / moderator / regular user, tested displaying already existing notes (they will be displayed as earlier - without tags)

### DIFF
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -50,7 +50,10 @@ OSM.NewNote = function (map) {
       data: {
         lat: location.lat,
         lon: location.lng,
-        text: $(form.text).val()
+        text: $(form.text).val(),
+        tags: JSON.stringify({
+          created_by: "OpenStreetMap-Website"
+        })
       },
       success: function (feature) {
         noteCreated(feature, marker);
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -83,11 +83,25 @@ module Api
       lat = OSM.parse_float(params[:lat], OSM::APIBadUserInput, "lat was not a number")
       comment = params[:text]
 
+      # Extract the tags, if present
+      tags = []
+      if params[:tags].present?
+        parsed_tags = JSON.parse(params[:tags])
+        parsed_tags.each do |key, value|
+          tags << { :k => key.presence || "", :v => value.presence || "" }
+        end
+      end
+
       # Include in a transaction to ensure that there is always a note_comment for every note
       Note.transaction do
         # Create the note
         @note = Note.create(:lat => lat, :lon => lon)
         raise OSM::APIBadUserInput, "The note is outside this world" unless @note.in_world?
+
+        # Create a NoteTag for each tag in the tags array
+        tags.each do |tag|
+          @note.note_tags.create(:k => tag[:k], :v => tag[:v])
+        end
 
         # Save the note
         @note.save!
--- a/app/models/note_tag.rb
+++ b/app/models/note_tag.rb
@@ -1,0 +1,20 @@
+# == Schema Information
+#
+# Table name: note_tags
+#
+#  note_id :bigint(8)        not null, primary key
+#  k       :string           default(""), not null, primary key
+#  v       :string           default(""), not null
+#
+# Foreign Keys
+#
+#  note_tags_id_fkey  (note_id => notes.id)
+#
+
+class NoteTag < ApplicationRecord
+  belongs_to :note
+
+  validates :note, :associated => true
+  validates :k, :v, :allow_blank => true, :length => { :maximum => 255 }, :characters => true
+  validates :k, :uniqueness => { :scope => :note_id }
+end
--- a/app/views/api/notes/_note.xml.builder
+++ b/app/views/api/notes/_note.xml.builder
@@ -14,6 +14,10 @@ xml.note("lon" => note.lon, "lat" => note.lat) do
 
   xml.date_closed note.closed_at if note.closed?
 
+  note.tags.each do |k, v|
+    xml.tag(:k => k, :v => v)
+  end
+
   xml.comments do
     note.comments.each do |comment|
       xml.comment do
--- a/db/migrate/20241030122707_create_note_tags.rb
+++ b/db/migrate/20241030122707_create_note_tags.rb
@@ -1,0 +1,12 @@
+class CreateNoteTags < ActiveRecord::Migration[7.2]
+  def change
+    # Create a table, primary and foreign keys
+    create_table :note_tags, :primary_key => [:note_id, :k] do |t|
+      t.column "note_id", :bigint, :null => false
+      t.column "k",  :string, :default => "", :null => false
+      t.column "v",  :string, :default => "", :null => false
+
+      t.foreign_key :notes, :column => :note_id, :name => "note_tags_id_fkey"
+    end
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1062,6 +1062,17 @@ CREATE TABLE public.note_subscriptions (
 
 
 --
+-- Name: note_tags; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.note_tags (
+    note_id bigint NOT NULL,
+    k character varying DEFAULT ''::character varying NOT NULL,
+    v character varying DEFAULT ''::character varying NOT NULL
+);
+
+
+--
 -- Name: notes; Type: TABLE; Schema: public; Owner: -
 --
 
@@ -2026,6 +2037,14 @@ ALTER TABLE ONLY public.note_comments
 
 ALTER TABLE ONLY public.note_subscriptions
     ADD CONSTRAINT note_subscriptions_pkey PRIMARY KEY (user_id, note_id);
+
+
+--
+-- Name: note_tags note_tags_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.note_tags
+    ADD CONSTRAINT note_tags_pkey PRIMARY KEY (note_id, k);
 
 
 --
@@ -3211,6 +3230,14 @@ ALTER TABLE ONLY public.note_comments
 
 
 --
+-- Name: note_tags note_tags_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.note_tags
+    ADD CONSTRAINT note_tags_id_fkey FOREIGN KEY (note_id) REFERENCES public.notes(id);
+
+
+--
 -- Name: redactions redactions_user_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: -
 --
 
@@ -3397,6 +3424,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('23'),
 ('22'),
 ('21'),
+('20241030122707'),
 ('20241023004427'),
 ('20241022141247'),
 ('20240913171951'),
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -94,6 +94,16 @@ class Note < ApplicationRecord
     comments.first.author_ip
   end
 
+  def tags
+    unless @tags
+      @tags = {}
+      note_tags.each do |tag|
+        @tags[tag.k] = tag.v
+      end
+    end
+    @tags
+  end
+
   private
 
   # Fill in default values for new notes
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -26,6 +26,8 @@ class Note < ApplicationRecord
   has_many :subscriptions, :class_name => "NoteSubscription"
   has_many :subscribers, :through => :subscriptions, :source => :user
 
+  has_many :note_tags
+
   validates :id, :uniqueness => true, :presence => { :on => :update },
                  :numericality => { :on => :update, :only_integer => true }
   validates :latitude, :longitude, :numericality => { :only_integer => true }
--- a/app/views/api/notes/_note.gpx.builder
+++ b/app/views/api/notes/_note.gpx.builder
@@ -23,5 +23,9 @@ xml.wpt("lon" => note.lon, "lat" => note.lat) do
     xml.status note.status
 
     xml.date_closed note.closed_at if note.closed?
+
+    note.tags.each do |k, v|
+      xml.tag(:k => k, :v => v)
+    end
   end
 end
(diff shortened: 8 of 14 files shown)


nenad-vujicic:
Here are a few more benefits of adding note tags to OSM, as implemented by this PR:

1. Improves **notes categorization** (by e.g. adding descriptive tags for better organization like "Missing place" or similar):

image

2. Improves **collaboration & communication** (by e.g. marking resolving some note requires outdoor survey):

image

3. Now we are ready for **advanced filtering** by specific tags (needs development)

4. Improves **public outreach** (by e.g. adding tags about mapathons)

image

5. Improves **multi-language support** (by e.g. adding language in which note description is written)

image

@tyrasd @westnordost @Zverik @ToeBee @angoca @talllguy what do you think about this PR?

Thanks,
Nenad.

westnordost:
Many of the use cases covered by this issue/PR are currently solved with hash tags (e.g. #surveyme) and tools are currently made to take these into account. Tools would need to be adapted to support tags as well.

The advantage of having tags implemented like this properly would be that there could be an efficient API with which to filter notes. (Right now, this is the _only_ advantage I can think of over hash tags and stuff.) For example:
- search for notes created by StreetComplete for certain quest types, to find common issues (misunderstandings, missing answer options) with quests
- search for notes created by certain versions of StreetComplete, e.g. to find notes that where perhaps created in error due to the above
- search specifically for notes that need e.g. survey. (However, this use case would not be useful for StreetComplete, because StreetComplete downloads all notes, anyway)

db/structure.sql
```diff
@@ -9,6 +9,13 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
+--
+-- Name: public; Type: SCHEMA; Schema: -; Owner: -
+--
+
+-- *not* creating schema, since initdb creates it
+
+
```
tomhughes, line 18: This block is noise created by the version of `pg_dump` you're using and has nothing to do with the actual changes in this PR so should be removed.

db/migrate/20241030122707_create_note_tags.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    # Create a table and primary key
+    create_table :note_tags, :primary_key => [:note_id, :k] do |t|
+      t.column "note_id", :bigint, :null => false
+      t.column "k",  :string, :default => "", :null => false
+      t.column "v",  :string, :default => "", :null => false
+    end
+
+    # Add foreign key without validation
+    add_foreign_key :note_tags, :notes, :column => :note_id, :name => "note_tags_id_fkey", :validate => false
```
tomhughes, line 11: Would moving this inside the `create_table` as `t.foreign_key` avoid the need to create it with validation disabled? Not sure if the strong migrations is clever enough to know that is safe?

If not then you can just add a second migration to enable validation as it's safe to do on an empty table.

app/assets/javascripts/index/new_note.js
```diff
@@ -50,7 +50,8 @@ OSM.NewNote = function (map) {
       data: {
         lat: location.lat,
         lon: location.lng,
-        text: $(form.text).val()
+        text: $(form.text).val(),
+        tags: "created_by:OpenStreetMaps-Website"
```
tomhughes, line 54: There is no `s` on the end of OpenStreetMap ;-) I'm not sure what is the best value for the tag here but it definitely shouldn't have the trailing s.

app/controllers/api/notes_controller.rb
```diff
@@ -83,12 +85,30 @@ def create
...
+      if params[:tags].present?
+        # Split by commas to get individual key-value pairs
+        pairs = params[:tags].split(",")
+
+        # For each pair in parameters, store it in tags variable
+        pairs.each do |pair|
+          key, value = pair.split(":", 2)
+          tags << { :k => sanitize(key), :v => sanitize(value) } if key && value
+        end
+      end
```
tomhughes, line 99: Encoding all the tags in a string like this is nasty and creates all sorts of problems if people use colons or commas in tag names or values - colons in particular are historically common in tag names for other object types.

I'm not sure what the solution is as this API is unusual in using form parameters rather than an XML or JSON body to create an object.

tomhughes (changes requested):

app/controllers/api/notes_controller.rb
AntonKhorev: JSON just for tags?

nenad-vujicic:
OK, here is another attempt. I made the above corrections and updated it to pass tag definitions as a nested object serialized into JSON, supporting various characters in keys and values.

app/views/api/notes/_note.xml.builder
```diff
@@ -14,6 +14,10 @@ xml.note("lon" => note.lon, "lat" => note.lat) do
 
   xml.date_closed note.closed_at if note.closed?
 
+  note.tags.each do |k, v|
```
kcne, line 17: This code is duplicated multiple times, consider creating a shared helper method to reuse instead.

app/models/note_tag.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+#  note_tags_id_fkey  (note_id => notes.id)
+#
+
+class NoteTag < ApplicationRecord
+  belongs_to :note
+
+  validates :note, :associated => true
+  validates :k, :v, :allow_blank => true, :length => { :maximum => 255 }, :characters => true
+  validates :k, :uniqueness => { :scope => :note_id }
```
kcne, line 19: To ensure stronger data integrity, we can enforce that k and v are always present, within length limits, and unique per note. Additionally, using `presence: true` for note simplifies and improves performance compared to `associated: true`

```ruby
validates :note, presence: true
validates :k, presence: true, length: { maximum: 255 }, uniqueness: { scope: :note_id }
validates :v, presence: true, length: { maximum: 255 }
```

db/migrate/20241030122707_create_note_tags.rb
```diff
@@ -0,0 +1,12 @@
...
+    # Create a table and primary key
+    create_table :note_tags, :primary_key => [:note_id, :k] do |t|
+      t.column "note_id", :bigint, :null => false
+      t.column "k",  :string, :default => "", :null => false
+      t.column "v",  :string, :default => "", :null => false
+
+      t.foreign_key :notes, :column => :note_id, :name => "note_tags_id_fkey"
+    end
+  end
+end
```
kcne, line 12: This migration works well for ensuring uniqueness with a composite primary key. However, if querying by tags (e.g., `{k: "created_by", v: "openstreetmap"}`) is expected, a surrogate primary key (`id`) with additional indexes might be more efficient. For example:

```ruby
create_table :note_tags do |t|
  t.bigint :note_id, null: false
  t.string :k, null: false, default: ""
  t.string :v, null: false, default: ""

  t.timestamps
end

add_index :note_tags, [:note_id, :k], unique: true       # Enforce uniqueness
add_index :note_tags, [:k, :v]                          # Optimize tag-based queries
```

This approach aligns with Rails conventions, simplifies queries, and allows adding indexes for tag lookups (`k`/`v`).

AntonKhorev: How is a surrogate key going to help with queries?

kcne: I misunderstood the use case, you are right it doesn't. Adding the index on k and v would do.. I will edit the top comment for more clarity

nenad-vujicic:
Here is the latest update with removed validation (which allows k and v to be empty strings, like they can be in e.g. [this node](https://www.openstreetmap.org/node/9842077514) (@kcne thanks for helping in catching this). I haven't added other stuffs (composite index - this will be part of separate PR, pulling out shared routine for writing tags - somehow this doesn't affect readability to me, but we could pull it out for all elements in a separate PR), but can add them if others request. 

@tomhughes @gravitystorm @AntonKhorev can you, please, share your thoughts about this PR and a way it solves the problem? If we want to do it in a proper way, we will need to pass XML / JSON as a payload, but the problem with this approach is we'll either break backward compatibility or would need to create additional route which will work in a new way (this looks a little bit dirty to me and more like a workaround than a proper way). What is the preferred way for you at the moment (the current one in this PR or passing XML/JSON as a payload using separate route) or can you suggest some other better solution?

Thanks,
Nenad.

app/views/api/notes/_note.rss.builder
```diff
@@ -19,4 +19,8 @@ xml.item do
   xml.geo :lat, note.lat
   xml.geo :long, note.lon
   xml.georss :point, "#{note.lat} #{note.lon}"
+
+  note.tags.each do |k, v|
+    xml.tag(:k => k, :v => v)
+  end
```
AntonKhorev, line 25: Is it going to be understood by any rss readers?

nenad-vujicic: Thanks for catching this. Honestly, I'm not sure about what could be the best solution here. I believe we have multiple options:

1. Adding Atom + XHTML namespaces and displaying tags like in [user's history feed](https://www.openstreetmap.org/user/Nenad%20Vujicic/history/feed) (I'm a little bit afraid to add new namespaces because of client support)
2. Using e.g. <dc:subject> for simplified, flattened representation of tags (the easiest and the most portable, but not so nice visual representation of tags)
3. Creating custom or some other namespace or some other solution?

Can you suggest what could be the best approach here?

nenad-vujicic: Or perhaps using one dc:subject element per note-tag pair (supported by the DCMI standard)? This was used in the most recent push.

AntonKhorev: Or not writing the tags in rss for now. Currently nobody is looking for them there.

nenad-vujicic: Thanks, removed!

app/models/note_tag.rb
```diff
@@ -0,0 +1,20 @@
...
+# Foreign Keys
+#
+#  note_tags_id_fkey  (note_id => notes.id)
+#
+
+class NoteTag < ApplicationRecord
+  belongs_to :note
+
+  validates :note, :associated => true
+  validates :k, :v, :allow_blank => true, :length => { :maximum => 255 }, :characters => true
```
tomhughes, line 18: Why are we allowing blank values? Possibly there's an argument for a blank value but certainly not for a key?

db/migrate/20241030122707_create_note_tags.rb
```diff
@@ -0,0 +1,12 @@
+class CreateNoteTags < ActiveRecord::Migration[7.2]
+  def change
+    # Create a table, primary and foreign keys
+    create_table :note_tags, :primary_key => [:note_id, :k] do |t|
+      t.column "note_id", :bigint, :null => false
+      t.column "k",  :string, :default => "", :null => false
+      t.column "v",  :string, :default => "", :null => false
```
tomhughes, line 7: Do these really need a default value?

app/models/note_tag.rb
AntonKhorev: Blank keys are allowed in other tags. Do we have many of them?

tomhughes: 125 changesets, 58 relations, 1300 ways (in current versions) but in any case just because we allowed silly things in the past doesn't mean we should continue to allow them in new models.

mmd-osm: I’m rejecting empty keys in cgimap since many years now. I guess they originate from single object rails code.

db/migrate/20241030122707_create_note_tags.rb
nenad-vujicic: Thanks, I'll fix this in incoming PR(s)

PR #5346, opened 2024-11-21
Add subscribe/unsubscribe buttons to note pages

AntonKhorev:
Part 5 of https://github.com/openstreetmap/openstreetmap-website/pull/5283 that adds the button to subscribe/unsubscribe from a note.

image

### DIFF
--- a/app/assets/javascripts/index/note.js
+++ b/app/assets/javascripts/index/note.js
@@ -37,31 +37,35 @@ OSM.Note = function (map) {
   };
 
   function initialize(path, id) {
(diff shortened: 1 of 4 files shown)


AntonKhorev:
Enabled showing errors for (un)subscribe buttons.

I didn't want to do this because the most likely error is that you're blocked and your access to the api is removed. But we shouldn't block people from unsubscribing. I thought that maybe I'll allow that first.

tomhughes (approved):
Thanks - that looks good now.

PR #5347, opened 2024-11-21
Custom layer filters

AntonKhorev:
Applies different filters in dark mode: invert+rotate for layers without hillshading, dim for layers with.

Doesn't implement any controls to changing the filter and, consequently, doesn't store selected values. For storing I'd start with deciding if #5339 is ok first.

image

image


nenad-vujicic (approved):

gravitystorm:
I don't approve of this PR, because I believe the cartographers, not the developers, should choose what their "dark mode maps" look like. See https://github.com/openstreetmap/openstreetmap-website/issues/5328#issuecomment-2488361637

PR #5348, opened 2024-11-21
Collect map layer credits into a data structure

AntonKhorev:
Can we have a proper map layer configuration that can be edited without going into javascript and altering bits across different files? I started doing this, here's a subtask of making a configuration for layer credits. It removes a bunch of interdependent variables like `var cyclosm_link`, `var osm_france_link`, `var cyclosm` etc.


tomhughes:
Looks good to me, thanks.

PR #5349, opened 2024-11-21
Move map layer ids/codes/keys to layer definition data

AntonKhorev:
Next step after #5348. Removes all layer-specific code from `leaflet.map.js` and moves layer definitions to `config/layers.yml`.


tomhughes:
Looks good to me, thanks.

Issue #3836, opened 2022-12-09
Website from mobile should have option to edit by opening vespucci


angoca:
### Problem

When solving notes by clicking a link on a mobile phone, there is not a way to do a simple change on the map from the website.

When doing the same from desktop, one can open iD or Josm (remote control). But there is not any integrated option to do a simple map change from the website. This will be useful to solve notes.

### Description

Having an option to redirect to osm editors like vespucci or other mobile apps will be great to improve the way notes are being solved.

tomhughes:
Well if the mobile editors want to provide a way for us to to that then it's no problem for us to add, but they really need to get together and agree a URL scheme or something and then register to handle it, then we can add it to the edit menu.

HolgerJeromin:
> When solving notes by clicking a link on a mobile phone, there is not a way to do a simple change on the map from the website.

You can click on the "Geo-URI" link in the share menü on the right. This will open vespucci (and other apps) at this location.

angoca:
I see that many applications open with this kind of URL (StreetComplete, Vespucci). Now, I see that the web interface could be modified because I find the word "Share" is not very. I agree that it could be "share" for some cases, but not all of them. Also, another icon could be more intuitive.

tomhughes:
Well ideally it would be on the edit menu, but I don't really want to use a geo uri there as that is claimed by viewers as well as editors and I'd rather have something that was specific to mobile editors.

simonpoole:
See https://github.com/openstreetmap/openstreetmap-website/pull/1478 This would have to be revisited as lots of things have changed wrt URL handling in Android. As has been noted geo: URLs have  been supported for a long time.

westnordost:
FYI https://en.osm.town/@foxy/115912841548838147

StreetComplete can be started via a link with
- https://streetcomplete.app/s (starts app if installed, leads to webpage with download buttons if not)
- or streetcomplete://s (starts app if installed)

PR #5351, opened 2024-11-22
Add josm scheme link to edit menu

AntonKhorev:
image

On Android, Vespucci is registered for `josm:` links: https://vespucci.io/tutorials/vespucci_intents/#supported-intents

This provides an option to run Vespucci as requested by #3836.


tomhughes:
Why on earth is it `josm:` though...

AntonKhorev:
Probably because it supports some of the commands that remote control also supports, and remote control is mostly a JOSM thing.

matkoniecz:
> it requires detecting the OS, something we'd rather not do

is it problematic for some reason? (is there really no reliable way to do it?)

AntonKhorev:
I tried registering this script as a `josm:` protocol handler on a desktop with JOSM remote control running, seems to work:

```perl
#!/usr/bin/perl

use LWP::Simple;
my $url = $ARGV[0];
$url =~ s{^josm:}{http://127.0.0.1:8111};
get $url;
```

PR #5352, opened 2024-11-23
Put api keys directly in js layer definitions

AntonKhorev:
Currently map layer API keys have one extra level of indirection:
- keys are defined in the settings
- `osm.js.erb` reads them from the settings and writes them to `OSM.THUNDERFOREST_KEY` and `OSM.TRACESTRACK_KEY`
- `layers.yml` contain names of `OSM` properties to check for keys

Instead of this `osm.js.erb` could write keys directly to layer definitions, skipping `OSM.THUNDERFOREST_KEY` and `OSM.TRACESTRACK_KEY`. `layers.yml` could contain the corresponding settings names. The advantage is that Thunderforest and Tracestrack keys are no longer special. A new key can be added just by editing `settings.yml` and `layers.yml`. Also it's possible to skip layer definitions if they require a key and that key is missing.

### DIFF
--- a/config/layers.yml
+++ b/config/layers.yml
@@ -28,7 +28,7 @@
   code: "C"
   layerId: "cyclemap"
   nameId: "cycle_map"
-  apiKeyId: "THUNDERFOREST_KEY"
+  apiKeyId: "thunderforest_key"
   canEmbed: true
   canDownloadImage: true
   credit:
@@ -42,7 +42,7 @@
   code: "T"
   layerId: "transportmap"
   nameId: "transport_map"
-  apiKeyId: "THUNDERFOREST_KEY"
+  apiKeyId: "thunderforest_key"
   canEmbed: true
   canDownloadImage: true
   credit:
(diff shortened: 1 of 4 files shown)


AntonKhorev:
I moved that ruby code to a module, but now there's no dependency on it. If it's changed, `osm.js` won't be recompiled.

config/layers.yml
```diff
@@ -42,7 +42,7 @@
   code: "T"
   layerId: "transportmap"
   nameId: "transport_map"
-  apiKeyId: "THUNDERFOREST_KEY"
+  apiKeyId: "thunderforest_key"
```
hlfan, line 45: Now that this is used in embed.js.erb as well, that file would need to be accommodated too.

hlfan (changes requested):

tomhughes (changes requested):
As @hlfan says this will need modifying to support `embed.js.erb` now - possibly the block of ruby code there should become an extra method in the `MapLayers` library that this introduces?

hlfan:
Maybe something along the lines of this?
```ruby
def self.embedDefinitions(layers_filename)
  self.definitions(layers_filename)
    .select { |entry| entry["canEmbed"] }
    .each_with_object({}) do |entry, obj|
      obj[entry["layerId"]] = {
        layer: entry["leafletOsmId"],
        darkLayer: entry["leafletOsmDarkId"],
        apikey: entry["apikey"]
      }.compact
    end
end
```
But I'm no ruby expert

PR #5353, opened 2024-11-24
Add default browse icons for nodes, ways, relations

AntonKhorev:
Add icons for elements that don't have any icon defined by their tags.

image image

Those are the standard node/way/relation icons as used in the wiki, taginfo etc with small modifications. The most visible modification is that I made the borders lighter. I did this to make them stand out less when displayed along with tag icons:

image

Actually I'm doing this to try out some vertical space saving changes to #5317 that are going to look better if every list element item has something to act as a bullet point. That would be an icon, so it's better to always have one.


tomhughes:
Looks good to me, thanks.

PR #5357, opened 2024-11-27
Resolve erb_lint deprecations

gravitystorm:
They have standardized on underscores. See https://github.com/Shopify/erb_lint/pull/360


tomhughes:
Looks good to me, thanks.

PR #5358, opened 2024-11-27
Move html_safe declaration for user_mailer from layout to helper

gravitystorm:
This allows us to enable output safety checks for all code within erb files.

`rubocop --auto-gen-config` ignores code within the erb files, so it is easier to maintain an exclusion on the helper than inside the layout.


tomhughes:
Looks good to me, thanks.

Issue #5258, opened 2024-10-12
use highway=busway instead of emergency=yes when determining feature type in query results


tguen:
### Problem

In query results, features with highway=busway and emergency=yes are listed only as "yes" in query results. 

For example, [this one](https://www.openstreetmap.org/way/334029996) shown below.

### Screenshots

![busway](https://github.com/user-attachments/assets/ac85fa11-2155-4ce8-8360-6f63c72b75ae)

mmd-osm:
New entry for highway=busway to be added here: https://github.com/openstreetmap/openstreetmap-website/blob/master/config/locales/en.yml#L920

mmd-osm:
New entry for highway=busway to be added here: https://github.com/openstreetmap/openstreetmap-website/blob/master/config/locales/en.yml#L920

![image](https://github.com/user-attachments/assets/a1663f9d-1739-42e8-a839-480e8fb56ddf)

mmd-osm:
According to https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#nominatim-prefixes, "I18n keys under the geocoder.search_osm_nominatim" are managed by Nominatim maintainers.

I'm not exactly clear how this was done in the past. Maybe @lonvia can help here.

lonvia:
Feel free to add new entries. And feel free to make a PR to get of this section in CONTRIBUTING.md.

prasathsarath:
@tguen / @gravitystorm / @lonvia  : I will work on this issue, please assign this to me

HolgerJeromin:
Not all projects work with assignments. Especially to new contributors. This can result in [cookie licking](https://www.redhat.com/en/blog/dont-lick-cookie).
Go ahead, get familiar to the code base and work on this if you like. 
If you solve this. Hurray!

psy-duck1:
Hey there, i'm interested in this issue will you please assign this to me.

hlfan:
You don't need to be assigned to work on this, as already pointed out:

> Not all projects work with assignments. Especially to new contributors. This can result in [cookie licking](https://www.redhat.com/en/blog/dont-lick-cookie).
Go ahead, get familiar to the code base and work on this if you like.
If you solve this. Hurray!

And you can even take a look at #5321 and #5361.

tomhughes:
Fixed by #5655.

PR #5361, opened 2024-11-30
Yatharths11/issue5258

Yatharths11:
### Description

Updated en.yml to add the necessary change to show the text 
used highway=busway instead of emergency=yes

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -89,12 +89,12 @@ en:
         url: Main Application URL (Required)
         callback_url: Callback URL
         support_url: Support URL
-        allow_read_prefs: read their user preferences
+        allow_read_prefs:  read their user preferences
         allow_write_prefs: modify their user preferences
         allow_write_diary: create diary entries, comments and make friends
-        allow_write_api: modify the map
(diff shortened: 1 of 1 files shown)


tomhughes:
Why would `highway=highway` be a bus way? It's a nonsense tag that has [no uses](https://taginfo.openstreetmap.org/tags/highway=highway#overview).

config/locales/en.yml
```diff
@@ -917,6 +917,7 @@ en:
         highway:
           abandoned: "Abandoned Highway"
           bridleway: "Bridleway"
+          highway: "Busway"
```
hlfan, line 920: If you mean to resolve #5258, the key here should be busway.

hlfan (changes requested):

gravitystorm:
Thank you @Yatharths11 for making your pull request to this project!

Unfortunately, as @hlfan has pointed out, it doesn't quite resolve the problem. You need to change your code, as he describes.

Please feel free to either reopen this PR when you have made the fix, or you could open a new PR instead.

PR #5362, opened 2024-12-01
Map color mode preference

AntonKhorev:
Includes #5339. Implements step 5 from https://github.com/openstreetmap/openstreetmap-website/issues/5324#issuecomment-2504102012.

image
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -502,17 +502,29 @@ body.small-nav {
 }
 
 @include color-mode(dark) {
-  .leaflet-tile-container .leaflet-tile,
-  .mapkey-table-entry td:first-child > * {
-    filter: brightness(.8);
-  }
-
   .leaflet-container .leaflet-control-attribution a {
     color: var(--bs-link-color);
   }
 
   .leaflet-control-scale-line {
     border-color: rgba(var(--bs-light-rgb), .75) !important;
+  }
+}
+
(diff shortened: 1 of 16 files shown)


app/assets/stylesheets/common.scss
```diff
@@ -516,6 +511,23 @@ body.small-nav {
   }
 }
 
+@mixin dark-map-color-scheme {
+  .leaflet-tile-container .leaflet-tile,
+  .mapkey-table-entry td:first-child > * {
+    filter: brightness(.8);
```
hlfan, line 517: Doesn't the mentioned step 5 include the need for the filters to be set individually to give the cartographers a choice?
I don't see that as a necessity for this PR, but already setting the filter through a custom property defined on the layer could ease the transition to that being implemented and also act as default and fallback:
```css
here {
    filter: var(--dark-map-filter);
}
somewhere else {
    .leaflet-layer {
        --dark-map-filter: brightness(.8);
    }
}
```

AntonKhorev: Step 5 has an impossible requirement of "make it dark but don't pick any filter". The closest possible thing is "don't touch existing filters".

omcnoe:
This will only be configurable for logged in users, meaning that non-logged in will still be stuck on the dark UI + dimmed tiles.

gravitystorm:
Thanks @AntonKhorev , I think this works well.

PR #5363, opened 2024-12-02
Dynamic "more" dropdown

AntonKhorev:
image
image
image
image

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -71,12 +71,12 @@ $(document).ready(function () {
   Turbo.session.drive = false;
 
   var headerWidth = 0,
-      compactWidth = 0;
+      breakpointWidth = 768;
 
   function updateHeader() {
     var windowWidth = $(window).width();
 
-    if (windowWidth < compactWidth) {
+    if (windowWidth < breakpointWidth) {
       $("body").addClass("small-nav");
       expandSecondaryMenu();
     } else if (windowWidth < headerWidth) {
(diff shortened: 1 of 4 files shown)


nenad-vujicic:
Somehow it moves all remaining buttons too early on my side. Here is a video:

https://github.com/user-attachments/assets/88d8c735-84e2-4d3b-bbc8-bca56b052df8

AntonKhorev:
> it moves all remaining buttons too early

Do you mean the width breakpoint for "small mode" with burger menu is too high? It has to be one of [the Bootstrap breakpoints](https://getbootstrap.com/docs/5.3/layout/breakpoints/) if I'm to use their navbar later. The only smaller breakpoint is 576px, but even with both *History* and *Export* moved to the secondary navigation things are not going to fit in some locales:

image

If we really want the 576px breakpoint, a possible extra trick is to hide "OpenStreetMap" next to the logo between 576px and 768px.

nenad-vujicic:
> > it moves all remaining buttons too early
> 
> Do you mean the width breakpoint for "small mode" with burger menu is too high?

No, I think "next change" should be moving `User Diaries` to `More`, not switching to burger menu. Then, on next width decrease, moving `GPS Traces` to `More` .. i.e. it should behave like wrapping (flex-wrap) but instead of automatic moving to next line (wrapping) it should move buttons to `More`.

Somehow, I prefer this PR (initial commits with mentioned improvements) over #5370 but let's hear what others have to say.

gravitystorm:
This looks good to me, and in particular I'm grateful for the move towards resolving #4130.

Merged, thanks.

PR #5364, opened 2024-12-02
Set dir=auto on username in header

AntonKhorev:
There's a problem with truncated usernames if their directionality doesn't match the ui. This PR fixes the current username in the header.

Before:
image

After:
image

There's also a bigger issue #3428 and some stale pull requests trying to fix it.

PR #5365, opened 2024-12-02
Set dir=auto on usernames in block lists

AntonKhorev:
Same as #5364 but for block lists.

Before:
image
image

After:
image
image


HolgerJeromin (approved):

tomhughes:
Looks good to me, thanks.

PR #5370, opened 2024-12-03
No compact nav 2 hide logo

AntonKhorev:
A version of #5363 that uses a smaller breakpoint by hiding "OpenStreetMap" as described in https://github.com/openstreetmap/openstreetmap-website/pull/5363#issuecomment-2513598424.

image
image

image
image


HolgerJeromin:
This would result in having the projects name "never" visible for small screens... not sure this is good

PR #5372, opened 2024-12-03
Truncate if necessary and set dir=auto for message usernames

AntonKhorev:
Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5374, opened 2024-12-04
Use relation icon instead of word in containing relations

### LINKED ISSUE #5368: Replace the text with the icon in the parent relation list
### Problem

After #5353 this already works for the ways:

https://www.openstreetmap.org/node/7925741250

But not for relations:

Снимок экрана 2024-12-03 в 00 58 53

https://www.openstreetmap.org/node/3062459558

It will probably be enough to add the `relation` html-class and update the string by removing the Relation at the beginning.  

### Description

_No response_

### Screenshots

_No response_

AntonKhorev:
Fixes #5368.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5375, opened 2024-12-04
Use fetch() instead of iframe to make remote control requests

AntonKhorev:
An alternative to #3760.

Uses no-cors mode https://github.com/openstreetmap/openstreetmap-website/pull/3760#issuecomment-1290363544 which should work with Potlatch. Requires a browser that is not too old (probably 5 years old max).


tomhughes:
Looks good to me, thanks.

PR #5371, opened 2024-12-03
Truncate if necessary and set dir=auto for changeset usernames

AntonKhorev:
Before:
image
image

After:
image
image

PR #5377, opened 2024-12-04
Truncate if necessary and set dir=auto for note usernames

### LINKED ISSUE #3964: broken order on note creation dates with some RTL usernames
screen01

day counter should be after username, also RTL one, at least in English

may be subset of #3428

definitely subset of #3427

AntonKhorev:
Similar to #5371 but for notes.

Fixes #3964.

Before:
image
note "43 (part of username) minutes ago", this is issue #3964

After:
image


tomhughes:
Looks good to me, thanks.

PR #5378, opened 2024-12-04
Include data: when using allow_thirdparty_images CSP

AntonKhorev:
We have `allow_thirdparty_images` policy on some pages that sets `img-src` policy to `*`. But `*` does not include data uris. That causes some of the ui elements to disappear.

For example, on `/account/edit` there's no dropdown icon:
image

After this fix:
image


tomhughes:
I suspect this got broken when we moved to the builtin CSP support. It looks good to me anyway,

PR #5380, opened 2024-12-05
Use translated model names in issue type select

AntonKhorev:
in the search form on `/issues`

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5381, opened 2024-12-05
Don't show types twice in issue type select

AntonKhorev:
If you have both moderator and admin flags, the issue search form is going to have "User" twice.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5382, opened 2024-12-06
Don't interpolate type in issue title

AntonKhorev:
The title of `/issues/:id` pages is constructed by substituting the status into the locale string, making it difficult to translate.

image

This PR makes it closer to what we have for note titles:
https://github.com/openstreetmap/openstreetmap-website/blob/fe81ac334cfef04751a959297dca24e472cb025f/config/locales/en.yml#L2977-L2979


tomhughes:
Looks good to me, thanks.

PR #5383, opened 2024-12-06
Let header username expand to screen width in small nav mode

AntonKhorev:
Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5384, opened 2024-12-06
Remove logger.debug from relation update action

AntonKhorev:
Was added in https://github.com/openstreetmap/openstreetmap-website/commit/2b7a40069f7af2e72c81f9488411135f286515f9. A similar debug statement also existed for nodes, added in https://github.com/openstreetmap/openstreetmap-website/commit/328d47e506972fededfa1080967224928c36a4cf, later removed.


tomhughes:
Looks good to me, thanks.

PR #5385, opened 2024-12-06
Write browse element icons as <img>

AntonKhorev:


app/helpers/browse_helper.rb
```diff
@@ -1,15 +1,43 @@
...
+    unless object.redacted?
+      target_tags = object.tags.find_all { |k, _v| BROWSE_ICONS.key? k.to_sym }.sort
+      title = target_tags.map { |k, v| "#{k}=#{v}" }.to_sentence unless target_tags.empty?
+
+      target_tags.each do |k, v|
+        k = k.to_sym
+        v = v.to_sym
+        if v != :* && BROWSE_ICONS[k].key?(v)
+          icon_data = BROWSE_ICONS[k][v]
+        elsif BROWSE_ICONS[k].key?(:*)
```
tomhughes, line 14: I'd suggest not allowing a wildcard key only match to override a more specific key=value match here:

```suggestion
        elsif icon_data.nil? && BROWSE_ICONS[k].key?(:*)
```

app/views/browse/_way.html.erb
```diff
@@ -24,17 +24,18 @@
...
       <details <%= "open" if way.way_nodes.count < 10 %>>
         <summary><%= t ".nodes_count", :count => way.way_nodes.count %></summary>
-        <ul class="list-unstyled">
+        <ul class="list-unstyled browse-element-list">
           <% way.way_nodes.each do |wn| %>
-            <li>
+            <%= element_list_item "node", wn.node do %>
               <%= element_single_current_link "node", wn.node %>
               <% related_ways = wn.node.ways.reject { |w| w.id == wn.way_id } %>
+              <% icon_connector = " " %>
```
tomhughes, line 32: Why use a variable for this when it doesn't ever seem to change?

tomhughes (changes requested):
So to summarise on the question of what happens when there are multiple icons the old logic is that priority is defined by the order of the rules  in the CSS which is mostly alphabetical except that the generic building rule is with the way rules which come after the node rules although the node/way split is really nonsense.

The new logic handles everything alphabetically, except that now the last rule wins rather than the first, and building is now ordered with other things so loses to shop but beats amenity.

My principle suggestion is that a key/value rule should always win over a key only rule as it's more specific.

If we want total backwards compatibility then the first matching rule should win - an way to do that would be to walk `target_tags` in reverse order when looking for a match.

I'm not sure that really matters though - if there are multiple key/value matches then there's no real right or wrong answer as to which one to choose.

app/views/browse/_way.html.erb
AntonKhorev: That was the easiest way to stop linters from complaining. Otherwise they tell to use interpolation. But as soon as I use interpolation, html gets escaped.

app/helpers/browse_helper.rb
AntonKhorev: Now I have default priorities of 1 for type icons, 10 for `k=*` and 100 for `k=v`; and it's possible to set a different priority in `browse_icons.yml`. I did it to control which icon is shown in case of `building=yes` + `shop=unknown_shop_type`.

app/views/browse/_way.html.erb
AntonKhorev: And it won't exist if we do #5317. Not having to deal with inline icons is one of the goals of that PR.

tomhughes (approved):
Hmm that style warning is a bit annoying, because string interpolation would be really nasty there, so I guess we live with the dummy variable for now.

Other than that is looks good now.

PR #5386, opened 2024-12-07
Allow specifying extra layer options in layers.yml

AntonKhorev:
Now you can (re)define other options accepted by Leaflet tile layers, for example `url` or `maxZoom`. Also you define new tile layers that are not present in `leaflet.osm.js` because `leafletOsmId` is no longer mandatory.

This lets you add, for example, dark version of  Transport Map by adding to `config/layers.yml`:

```yml
- url: 'https://{s}.tile.thunderforest.com/transport-dark/{z}/{x}/{y}{r}.png?apikey={apikey}'
  maxZoom: 21
  code: "U"
  keyId: "transportmapdark"
  nameId: "transport_map"
  apiKeyId: "THUNDERFOREST_KEY"
  credit:
    id: "thunderforest_credit"
    children:
      thunderforest_link:
        id: "andy_allan"
        href: "https://www.thunderforest.com/"
```

### DIFF
--- a/app/assets/javascripts/leaflet.layers.js
+++ b/app/assets/javascripts/leaflet.layers.js
@@ -29,7 +29,7 @@ L.OSM.layers = function (options) {
 
       map.whenReady(function () {
         var miniMap = L.map(mapContainer[0], { attributionControl: false, zoomControl: false, keyboard: false })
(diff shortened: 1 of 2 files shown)


tomhughes:
Nice - making it easier for people reusing the software for other sites to change the map layers has been a long standing problem.

PR #5387, opened 2024-12-09
Disable filter inheritance on tiles, apply filter to tile container

AntonKhorev:
If a filter is applied to every tile separately, it's possible that the gaps between tiles become visible when zooming. I've only seen this happening in Chrome, both on desktop and mobile.

It could have been the reason why I applied the filter to `.leaflet-tile-container` in the first place, which I had to undo in #5330. Now I'm applying the filter to the container again, but I disable the unexpected filer inheritance (which is disabled in the current unreleased Leaflet code).

Before:

https://github.com/user-attachments/assets/cfa6b459-b1b1-4a7e-8523-5c656e62d3ca

After:

https://github.com/user-attachments/assets/92d9c6b4-bb08-4f16-ac01-edfdecc0fc76

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -506,9 +506,13 @@ body.small-nav {
 }
 
 @include color-mode(dark) {
-  .leaflet-tile-container .leaflet-tile,
+  .leaflet-tile-container,
   .mapkey-table-entry td:first-child > * {
     filter: var(--dark-mode-map-filter);
+  }
+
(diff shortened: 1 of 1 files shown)


AntonKhorev:
Looks like the amount of this grid lines effect depends on the filter used. With invert+rotate filters it almost disappears on desktop, but is still significantly noticeable on mobile. But that could be just because the overall brightness of tiles.

tomhughes:
Looks good to me, thanks.

PR #5388, opened 2024-12-09
Add 'canEmbed' layer option

AntonKhorev:
Check the added `canEmbed` layer option instead of `map.getMapBaseLayerId() !== "tracestracktopo"` when deciding whether *Share > HTML* should be available.


gravitystorm:
This seems reasonable to me, but it's marked as a draft. Is there something else that needs working on first, or is it just the merge conflicts now?

AntonKhorev:
There was something else, updated now.

PR #5389, opened 2024-12-09
Rename 'keyid' layer option to 'layerId'

AntonKhorev:
Each base map layer has a `keyid`. It was introduced in https://github.com/openstreetmap/openstreetmap-website/commit/79df86023c5db1f1c47c80808565733c1200656a to hide/show map key entries for different layers. But since then it was also used for various other things and the function to get it is `getMapBaseLayerId`. So it's a `layerId` actually.


tomhughes:
Looks good to me, thanks.

PR #5391, opened 2024-12-10
Move admin users list to its own controller

AntonKhorev:
Previous attempt: #4969

What I'm trying to achieve:
- free `users_path` helper for POST requests that create users, as explained in #4969
- make `app/controllers/users_controller.rb` smaller


tomhughes:
Looks good to me, thanks.

PR #5394, opened 2024-12-11
iD: listen to URL hash changes

tyrasd:
When the `map` hash parameter is changed (and was not not triggered by map interaction from inside iD itself): pan to the new location.

This mirrors the behaviour when manually editing the `map` hash parameter on osm.org outside of iD.

This also fixes https://github.com/openstreetmap/iD/issues/10592 (error in js console when iD is opened outside of an iframe by directly navigating to osm.org/id).

### DIFF
--- a/app/assets/javascripts/id.js
+++ b/app/assets/javascripts/id.js
@@ -28,6 +28,12 @@ document.addEventListener("DOMContentLoaded", function () {
       .containerNode(container)
       .init();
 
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, though I would not that using openstreetmap.org/id directly is not really supported - it's an internal implementation detail that we certainly don't guarantee.

PR #5396, opened 2024-12-12
Use map tiles dark mode with leaflet-osm plugin

hlfan:
This lean and updated variant of #4777 follows the rules laid out in #5328 that were implemented in #5362.
Since the editing of the user preference happens on a different view/site, there is no need to listen to color scheme changes on the fly and the whole logic can happen in the initialization of `L.OSM.TileLayer`.
|Before|After|
|-|-|
|image|image|

Additionally, a "dark" CSS class is introduced to allow better style targeting if the user wants to.
for example like this:
image

### DIFF
--- a/vendor/assets/leaflet/leaflet.osm.js
+++ b/vendor/assets/leaflet/leaflet.osm.js
@@ -12,7 +12,7 @@ L.OSM.TileLayer = L.TileLayer.extend({
     options = L.Util.setOptions(this, options);
     url = isDarkMap ? options.darkUrl : options.lightUrl;
     if (url) this.schemeClass = isDarkMap ? 'dark' : 'light';
-    L.TileLayer.prototype.initialize.call(this, options.url);
+    L.TileLayer.prototype.initialize.call(this, url || options.url);
     this.on('add', function () {
       if (this._container && this.schemeClass) this._container.classList.add(this.schemeClass);
       if (this._container && this.options.filter) this._container.style.setProperty('--dark-mode-map-filter', this.options.filter);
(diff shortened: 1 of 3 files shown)


tomhughes:
Changes to anything under `vendor` need to be done in the appropriate upstream repository, not here.

vendor/assets/leaflet/leaflet.osm.js
```diff
@@ -7,8 +7,20 @@ L.OSM.TileLayer = L.TileLayer.extend({
   },
 
   initialize: function (options) {
+    isDarkMap = OSM.isDarkMap();
+    options.filter = (isDarkMap ? options.darkFilter : options.lightFilter) || options.filter || 'none';
     options = L.Util.setOptions(this, options);
-    L.TileLayer.prototype.initialize.call(this, options.url);
+    url = isDarkMap ? options.darkUrl : options.lightUrl;
```
tomhughes, line 13: There is no `lightUrl` anywhere?

tomhughes:
> > inline styles are not allowed by our security policy in production
> 
> The `style-src 'self' 'unsafe-inline'` in the CSP header tells me a different story.

OK so it is allowed for views with a map. I'm note sure why that is offhand but our goal is to not have it so we don't add new inline styles ourselves but we sometimes have to allow it for third party components.

vendor/assets/leaflet/leaflet.osm.js
hlfan: Well, if the provided layers can be expected always to be a light-mode-first choice, the logic can be simplified to this more unreadable version:
```js
    options.filter = isDarkMap && options.darkFilter || 'none';
    options = L.Util.setOptions(this, options);
    url = isDarkMap && options.darkUrl;
    this.schemeClass = url && 'dark';
```
But I thought an implementation agnostic to that would be easier to think through with more options for cartographers' choices.

hlfan:
> leaflet-osm should be getting involved in automatic transforms - it should have the ability to select a dark tile version where there is one

Do you mean only the extension of `L.OSM.TransportMap` should be added to leaflet-osm?

PR #5397, opened 2024-12-13
Use map tiles dark mode without leaflet-osm plugin

hlfan:
Like #5396 but without involving leaflet-osm.
Again, since embeds carry no color scheme preferences, I skipped the partially redundant implementation there.

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -35,7 +35,7 @@ L.OSM.Map = L.Map.extend({
         } else if (property === "darkUrl") {
           if (OSM.isDarkMap()) {
             layerOptions.url = value;
-            layerOptions.schemeClass = 'dark';
+            layerOptions.schemeClass = "dark";
           }
         } else {
           layerOptions[property] = value;
@@ -57,17 +57,19 @@ L.OSM.Map = L.Map.extend({
       code: "G"
     });
 
-    this.on("layeradd", function ({layer}) {
+    this.on("layeradd", function ({ layer }) {
       if (this.baseLayers.indexOf(layer) >= 0) {
         this.setMaxZoom(layer.options.maxZoom);
         const container = layer.getContainer();
         if (!container) return;
         if (layer.options.schemeClass) container.classList.add(layer.options.schemeClass);
         const filterRecievers = [container];
-        if (document.getElementById('map').contains(container))
-          filterRecievers.push(document.querySelector('.key-ui'));
-        for (let el of filterRecievers)
-          el.style.setProperty('--dark-mode-map-filter', layer.options.filter || 'none');
+        if (document.getElementById("map").contains(container)) {
+          filterRecievers.push(document.querySelector(".key-ui"));
+        }
+        for (let el of filterRecievers) {
+          el.style.setProperty("--dark-mode-map-filter", layer.options.filter || "none");
+        }
       }
     });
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/leaflet.map.js
```diff
@@ -52,9 +57,19 @@ L.OSM.Map = L.Map.extend({
...
-    this.on("layeradd", function (event) {
-      if (this.baseLayers.indexOf(event.layer) >= 0) {
-        this.setMaxZoom(event.layer.options.maxZoom);
+    this.on("layeradd", function ({ layer }) {
+      if (this.baseLayers.indexOf(layer) >= 0) {
+        this.setMaxZoom(layer.options.maxZoom);
+        const container = layer.getContainer();
+        if (!container) return;
+        if (layer.options.schemeClass) container.classList.add(layer.options.schemeClass);
+        const filterRecievers = [container];
```
gravitystorm, line 66: minor spelling error - `filterReceivers`

gravitystorm (changes requested):
I'm very happy to see this work, thanks @hlfan!

The code looks fine to me, but I'm not a javascript expert so maybe there are better ways to write parts.

Two other changes requested:

* Please rework your commits (e.g. using `git rebase -i`) to avoid having the fixup commits (e.g. combine the linting fixes into the original commits)
* There's still a darken filter applied to the map previews - you can see that the colour of the green spaces on the transport dark map are darkened compared to the main map

Screenshot from 2024-12-18 11-31-42

tomhughes (commented):
This has gone rather further than I intended from my comments on #5396 by moving everything back into the web site code.

My thinking was the leaflet-osm would still have the `darkUrl` attribute on the layer, and probably an option to the constructor that would cause it to be used if available, and only if that wasn't possible would the web site code activate a filter instead.

The only thing I'm not sure about is how to signal back to this code whether dark more was possible.

hlfan:
@tomhughes This is also further than I anticipated it would go.

But I found it too hacky to overwrite the class's properties from the instance each time, and I especially wanted to avoid using `__proto__`. Yet I couldn't find a way to do that.

But then I stumbled upon #5386 and its new layer definition handler that allowed me to eliminate those hacks since this would specify the scheme on the instance creation.

Also not involving leaflet-osm at all would increase the flexibility with things like SVG filters, where I don't see an easy option to have that sourced from JS nicely.

tomhughes:
> But I found it too hacky to overwrite the class's properties from the instance each time, and I especially wanted to avoid using `__proto__`. Yet I couldn't find a way to do that.

I don't see any reason why what I was suggesting would need any of that but I've actually come up with a better solution now.

What really bothered me was having the URL sometimes in leaflet-osm and sometimes here, so what I've done is to add a new `TransportDarkMap` later to leaflet-osm in https://github.com/openstreetmap/leaflet-osm/commit/ebfaf22426c265aaf3f401d70ffd02ca920623c9.

So now we can have a `leafletOsmDarkId: "TransportDarkMap"` key in `layers.yml` and if that key exists we construct that layer instead of the normal one and then don't do any filtering.

hlfan:
Given my limited familiarity with this repo, I decided to place cornerstones in the space of possibilities I saw rather than try for a perfect solution the first few times without much feedback.

But now that I see a more modular solution I'll put this PR on hold too, just fixing the missing filter propagation, and move on to attempt 3 to avoid cluttering up the history.

PR #5401, opened 2024-12-16
Update nominatim terms

lonvia:
### Description

I'm planning to introduce three new categories for Nominatim classes: lock, water and information. These are the main terms for it.

### How has this been tested?

Only made sure the CI runs successfully.


tomhughes:
Looks good to me, thanks.

Issue #5373, opened 2024-12-03
Calendar heatmap of user contributions on user profile


kcne:
### Problem

The current user profile page provides a detailed summary of contributions but lacks a visual representation of activity trends. A calendar heatmap, similar to GitHub's user contribution calendar, would make it easier to understand user engagement over time at a glance.

### Description

I have implemented a working calendar heatmap locally for changesets using the [cal-heatmap-rails library](https://rubygems.org/gems/cal-heatmap-rails/versions/3.6.2?locale=en) for display, querying changesets directly from the database. The heatmap visually displays daily contributions, with color intensity representing the number of changes for the changset on the given date.

**Additional Ideas:**
- Not sure if we should count changsets only in initial version or add diary entries and note creations/comments as well.

- Since we have dark mode on OSM plan is for that to be reflected on the heatmap as well.

**Next Steps:**  
I’ll open a PR after refining the implementation. In the meantime, I’d love to hear your thoughts and suggestions on this feature and the proposed contribution types.

### Screenshots

<img width="631" alt="Screenshot 2024-12-03 at 11 55 46" src="https://github.com/user-attachments/assets/dcb68c4f-f82f-4faa-b87b-d42ee29e3389">

AntonKhorev:
[cal-heatmap-rails](https://rubygems.org/gems/cal-heatmap-rails/versions/3.6.2?locale=en)

> Version Released:
> on Oct 2, 2018

Are you sure about this gem? Looks like all it does is repackaging of (outdated version of) https://github.com/wa0x6e/cal-heatmap.

maro-21:
The calendar heatmap is already on http://hdyc.neis-one.org.

tordans:
> > Ping https://github.com/openstreetmap/openstreetmap-website/issues/5356 which looks like the ambrella issue for this.
> Thanks for suggestion but in the end i ended up just mocking data in the database locally. Since the table is indexed it should not present issues performance wise.

FYI my ping was meant to link the different issues on the topic of adding features to the profile page together. (Not a comment on the content of this issue.) I think it is good to have separate issues per feature idea but one meta issue on the topic in general that glues those idea into one concept.

LivInTheLookingGlass:
> The calendar heatmap is already on http://hdyc.neis-one.org.

Which is closed source and out of the way

1ec5:
[...]

PR #5402, opened 2024-12-16
Add user profile heatmap visualization for contributions

kcne:
### Description
This PR addresses #5373 by adding a heatmap visualization on user profile pages to display daily contribution activity over the past year. 

**Key changes include:**
- Added a heatmap container in the user profile view (`users/show.html.erb`).
- Implemented JavaScript logic (`heatmap.js`) using CalHeatmap to render the heatmap dynamically based on contribution data.
- Updated localization (`config/locales/en.yml`) to include tooltips and month/weekday labels for the heatmap.
- Added tests in `users_controller_test.rb` to validate heatmap data handling.

Screenshots:
Screenshot 2024-12-16 at 16 35 58
Screenshot 2024-12-16 at 16 37 14
Screenshot 2024-12-16 at 16 36 25

### How has this been tested?
- Verified heatmap rendering with mock data for different users and activity levels.
- Tested responsiveness and theme support (light/dark mode).
- Checked correct localization of tooltips, month names, and weekday labels.
- Ran automated tests to ensure proper heatmap data handling in the controller.

Let me know if you need further tweaks or enhancements!

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -20,3 +20,4 @@
 //= link leaflet/dist/images/marker-icon.png
 //= link leaflet/dist/images/marker-icon-2x.png
 //= link leaflet/dist/images/marker-shadow.png
+//= link cal-heatmap/dist/cal-heatmap.css
--- a/app/assets/javascripts/heatmap.js
+++ b/app/assets/javascripts/heatmap.js
@@ -1,0 +1,101 @@
+//= require d3/dist/d3
+//= require cal-heatmap/dist/cal-heatmap
+//= require popper
+//= require cal-heatmap/dist/plugins/Tooltip
+
+/* global CalHeatmap, Tooltip */
+document.addEventListener("DOMContentLoaded", () => {
+  const heatmapElement = document.querySelector("#cal-heatmap");
+
+  if (!heatmapElement) {
+    return;
+  }
+
+  const heatmapData = heatmapElement.dataset.heatmap ? JSON.parse(heatmapElement.dataset.heatmap) : [];
+  const colorScheme = document.documentElement.getAttribute("data-bs-theme") ?? "auto";
+  const rangeColors = ["#14432a", "#166b34", "#37a446", "#4dd05a"];
+  const startDate = new Date(Date.now() - (365 * 24 * 60 * 60 * 1000));
+  const monthNames = I18n.t("date.abbr_month_names");
+
+  const mediaQuery = window.matchMedia("(prefers-color-scheme: dark)");
+
+  let cal = new CalHeatmap();
+  let currentTheme = getTheme();
+
+  function renderHeatmap() {
+    cal.destroy();
+    cal = new CalHeatmap();
+
+    cal.paint({
+      itemSelector: "#cal-heatmap",
+      theme: currentTheme,
+      domain: {
+        type: "month",
+        gutter: 4,
+        label: {
+          text: (timestamp) => monthNames[new Date(timestamp).getMonth() + 1],
+          position: "top",
+          textAlign: "middle"
+        },
+        dynamicDimension: true
+      },
+      subDomain: {
+        type: "ghDay",
+        radius: 2,
+        width: 11,
+        height: 11,
+        gutter: 4
+      },
+      date: {
+        start: startDate
+      },
+      range: 13,
+      data: {
+        source: heatmapData,
+        type: "json",
+        x: "date",
+        y: "total_changes"
+      },
+      scale: {
+        color: {
+          type: "threshold",
+          range: currentTheme === "dark" ? rangeColors : Array.from(rangeColors).reverse(),
+          domain: [10, 20, 30, 40]
+        }
+      }
+    }, [
+      [Tooltip, {
+        text: (date, value) => getTooltipText(date, value)
+      }]
+    ]);
+  }
+
+  function getTooltipText(date, value) {
+    const localizedDate = I18n.l("date.formats.long", date);
+
+    if (value > 0) {
+      return I18n.t("javascripts.heatmap.tooltip.contributions", { count: value, date: localizedDate });
+    }
+
+    return I18n.t("javascripts.heatmap.tooltip.no_contributions", { date: localizedDate });
+  }
+
+  function getTheme() {
+    if (colorScheme === "auto") {
+      return mediaQuery.matches ? "dark" : "light";
+    }
+
+    return colorScheme;
+  }
+
+  if (colorScheme === "auto") {
+    mediaQuery.addEventListener("change", (e) => {
+      currentTheme = e.matches ? "dark" : "light";
+      renderHeatmap();
+    });
+  }
+
+  renderHeatmap();
+});
+
+
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1007,3 +1007,7 @@ img.trace_image {
     }
   }
 }
+
+.heatmap-wrapper {
+  height: 130px;
+}
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -21,9 +21,27 @@ class UsersController < ApplicationController
   def show
     @user = User.find_by(:display_name => params[:display_name])
 
-    if @user &&
-       (@user.visible? || current_user&.administrator?)
+    if @user && (@user.visible? || current_user&.administrator?)
       @title = @user.display_name
+
+      @heatmap_data = Rails.cache.fetch("heatmap_data_user_#{@user.id}", :expires_in => 1.day) do
+        one_year_ago = 1.year.ago.beginning_of_day
+        today = Time.zone.now.end_of_day
+
+        Changeset
+          .where(:user_id => @user.id)
+          .where(:created_at => one_year_ago..today)
+          .where(:num_changes => 1..)
+          .group("date_trunc('day', created_at)")
+          .select("date_trunc('day', created_at) AS date, SUM(num_changes) AS total_changes")
+          .order("date")
+          .map do |changeset|
+            {
+              :date => changeset.date.to_date.to_s,
+              :total_changes => changeset.total_changes.to_i
+            }
+          end
+      end
     else
       render_unknown_user params[:display_name]
     end
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -1,3 +1,7 @@
+<% content_for :head do %>
+  <%= stylesheet_link_tag "cal-heatmap/dist/cal-heatmap" %>
+  <%= javascript_include_tag "heatmap" %>
+<% end %>
 <% content_for :heading do %>
   <div class="row">
     <div class="col-sm-auto">
@@ -232,6 +236,29 @@
 <% end %>
 
 <div class="richtext text-break clearfix"><%= @user.description.to_html %></div>
+
+<% if @heatmap_data.present? %>
+  <div class="row">
+    <div class="col overflow-auto">
+      <div class="heatmap-wrapper d-flex align-items-start">
+        <!-- Labels -->
+        <ul class="list-unstyled d-flex flex-column justify-content-between ch-domain-text mb-0 mt-4">
+          <li>&nbsp;</li>
+          <li><%= t("date.abbr_day_names")[1] %></li>
+          <li>&nbsp;</li>
+          <li><%= t("date.abbr_day_names")[3] %></li>
+          <li>&nbsp;</li>
+          <li><%= t("date.abbr_day_names")[5] %></li>
+          <li>&nbsp;</li>
+        </ul>
+        <!-- Heatmap -->
+        <div id="cal-heatmap" class="ms-2"
+          data-heatmap="<%= @heatmap_data.to_json %>">
+        </div>
+      </div>
+    </div>
+  </div>
+<% end %>
 
 <% if current_user and @user.id == current_user.id %>
   <div class="my-3">
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3329,6 +3329,12 @@ en:
     home:
       marker_title: My home location
       not_set: Home location is not set for your account
+    heatmap:
+      tooltip:
+        no_contributions: "No contributions on %{date}"
+        contributions:
+          one: "%{count} contribution on %{date}"
+          other: "%{count} contributions on %{date}"
   redactions:
     edit:
       heading: "Edit Redaction"
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -329,4 +329,97 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     get auth_failure_path, :params => { :origin => "http://www.google.com" }
     assert_redirected_to login_path
   end
+
+  def test_show_heatmap_data
+    user = create(:user)
+    # Create two changesets
+    create(:changeset, :user => user, :created_at => 6.months.ago, :num_changes => 10)
+    create(:changeset, :user => user, :created_at => 3.months.ago, :num_changes => 20)
+
+    get user_path(user.display_name)
+    assert_response :success
+    # The data should not be empty
+    assert_not_nil assigns(:heatmap_data)
+
+    heatmap_data = assigns(:heatmap_data)
+    # The data should be in the right format
+    assert(heatmap_data.all? { |entry| entry[:date] && entry[:total_changes] }, "Heatmap data should have :date and :total_changes keys")
+  end
+
+  def test_show_heatmap_data_caching
+    # Enable caching to be able to test
+    Rails.cache.clear
+    @original_cache_store = Rails.cache
+    Rails.cache = ActiveSupport::Cache::MemoryStore.new
+
+    user = create(:user)
+
+    # Create an initial changeset
+    create(:changeset, :user => user, :created_at => 6.months.ago, :num_changes => 15)
+
+    # First request to populate the cache
+    get user_path(user.display_name)
+    first_response_data = assigns(:heatmap_data)
+    assert_not_nil first_response_data, "Expected heatmap data to be assigned on the first request"
+    assert_equal 1, first_response_data.size, "Expected one entry in the heatmap data"
+
+    # Inspect cache after the first request
+    cached_data = Rails.cache.read("heatmap_data_user_#{user.id}")
+    assert_equal first_response_data, cached_data, "Expected the cache to contain the first response data"
+
+    # Add a new changeset to the database
+    create(:changeset, :user => user, :created_at => 3.months.ago, :num_changes => 20)
+
+    # Second request
+    get user_path(user.display_name)
+    second_response_data = assigns(:heatmap_data)
+
+    # Confirm that the cache is still being used
+    assert_equal first_response_data, second_response_data, "Expected cached data to be returned on the second request"
+
+    # Clear the cache and make a third request to confirm new data is retrieved
+    Rails.cache.clear
+    get user_path(user.display_name)
+    third_response_data = assigns(:heatmap_data)
+
+    # Ensure the new entry is now included
+    assert_equal 2, third_response_data.size, "Expected two entries in the heatmap data after clearing the cache"
+
+    # Reset caching config to defaults
+    Rails.cache.clear
+    Rails.cache = @original_cache_store
+  end
+
+  def test_show_heatmap_data_no_changesets
+    user = create(:user)
+
+    get user_path(user.display_name)
+    assert_response :success
+    # There should be no entries in heatmap data
+    assert_empty assigns(:heatmap_data)
+  end
+
+  def test_heatmap_rendering
+    # Test user with no changesets
+    user_without_changesets = create(:user)
+    get user_path(user_without_changesets)
+    assert_response :success
+    assert_select "div#cal-heatmap", 0
+
+    # Test user with changesets
+    user_with_changesets = create(:user)
+    create(:changeset, :user => user_with_changesets, :created_at => 3.months.ago.beginning_of_day, :num_changes => 42)
+    create(:changeset, :user => user_with_changesets, :created_at => 4.months.ago.beginning_of_day, :num_changes => 39)
+    get user_path(user_with_changesets)
+    assert_response :success
+    assert_select "div#cal-heatmap[data-heatmap]" do |elements|
+      # Check the data-heatmap attribute is present and contains expected JSON
+      heatmap_data = JSON.parse(elements.first["data-heatmap"])
+      expected_data = [
+        { "date" => 4.months.ago.to_date.to_s, "total_changes" => 39 },
+        { "date" => 3.months.ago.to_date.to_s, "total_changes" => 42 }
+      ]
+      assert_equal expected_data, heatmap_data
+    end
+  end
 end
--- a/yarn.lock
+++ b/yarn.lock
@@ -103,6 +103,20 @@
   resolved "https://registry.yarnpkg.com/@humanwhocodes/retry/-/retry-0.4.1.tgz#9a96ce501bc62df46c4031fbd970e3cc6b10f07b"
   integrity sha512-c7hNEllBlenFTHBky65mhq8WD2kbN9Q6gk0bTk8lSBvc554jpXSkST1iePudpt7+A/AQvuHs9EMqjHDXMY1lrA==
 
+"@observablehq/plot@^0.6.0":
+  version "0.6.17"
+  resolved "https://registry.yarnpkg.com/@observablehq/plot/-/plot-0.6.17.tgz#9a43f785e89e6d4413701e2b24d2ec8cc2700e02"
+  integrity sha512-/qaXP/7mc4MUS0s4cPPFASDRjtsWp85/TbfsciqDgU1HwYixbSbbytNuInD8AcTYC3xaxACgVX06agdfQy9W+g==
+  dependencies:
+    d3 "^7.9.0"
+    interval-tree-1d "^1.0.0"
+    isoformat "^0.2.0"
+
+"@popperjs/core@^2.11.6":
+  version "2.11.8"
+  resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.11.8.tgz#6b79032e760a0899cd4204710beede972a3a185f"
+  integrity sha512-P1st0aksCrn9sGZhp8GMYwBnQsbvAWsZAX44oXNNvLHGqAOcoVxmjZiohstwQ7SqKnbR47akdNi+uleWD8+g6A==
+
 "@stylistic/eslint-plugin-js@^4.0.0":
   version "4.0.1"
   resolved "https://registry.yarnpkg.com/@stylistic/eslint-plugin-js/-/eslint-plugin-js-4.0.1.tgz#c55e59ad2f00914f9eb103ba9e7be00aff7715dd"
@@ -182,6 +196,11 @@ balanced-match@^1.0.0:
   resolved "https://registry.yarnpkg.com/balanced-match/-/balanced-match-1.0.2.tgz#e83e3a7e3f300b34cb9d87f615fa0cbf357690ee"
   integrity sha512-3oSeUO0TMV67hN1AmbXsK4yaqU7tjiHlbxRDZOpH0KW9+CeX4bRAaX0Anxt0tx2MrpRpWwQaPwIlISEJhYU5Pw==
 
+binary-search-bounds@^2.0.0:
+  version "2.0.5"
+  resolved "https://registry.yarnpkg.com/binary-search-bounds/-/binary-search-bounds-2.0.5.tgz#125e5bd399882f71e6660d4bf1186384e989fba7"
+  integrity sha512-H0ea4Fd3lS1+sTEB2TgcLoK21lLhwEJzlQv3IN47pJS976Gx4zoWe0ak3q+uYh60ppQxg9F16Ri4tS1sfD4+jA==
+
 brace-expansion@^1.1.7:
   version "1.1.11"
   resolved "https://registry.yarnpkg.com/brace-expansion/-/brace-expansion-1.1.11.tgz#3c7fcbf529d87226f3d2f52b966ff5271eb441dd"
@@ -189,6 +208,21 @@ brace-expansion@^1.1.7:
   dependencies:
     balanced-match "^1.0.0"
     concat-map "0.0.1"
+
+cal-heatmap@^4.2.4:
+  version "4.2.4"
+  resolved "https://registry.yarnpkg.com/cal-heatmap/-/cal-heatmap-4.2.4.tgz#0c35616be0dc09df64e1163a804231d8d3f9feaa"
+  integrity sha512-TTNoQTRxHXrttOEbkraKv9vy2VpfQIwVLQJTlAfcBusQK9qrBL/UBO+WloAxv2yrR+P8URA2cuXEdc5iztER9g==
+  dependencies:
+    "@observablehq/plot" "^0.6.0"
+    "@popperjs/core" "^2.11.6"
+    d3-color "^3.1.0"
+    d3-fetch "^3.0.1"
+    d3-selection "^3.0.0"
+    d3-transition "^3.0.1"
+    dayjs "^1.11.7"
+    eventemitter3 "^5.0.0"
+    lodash-es "^4.17.21"
 
 callsites@^3.0.0:
   version "3.1.0"
@@ -215,6 +249,11 @@ color-name@~1.1.4:
   resolved "https://registry.yarnpkg.com/color-name/-/color-name-1.1.4.tgz#c2a09a87acbde69543de6f63fa3995c826c536a2"
   integrity sha512-dOy+3AuW3a2wNbZHIuMZpTcgjGuLU/uBL/ubcZF9OXbDo8ff4O8yVp5Bf0efS8uEoYo5q4Fx7dY9OgQGXgAsQA==
 
+commander@7:
+  version "7.2.0"
+  resolved "https://registry.yarnpkg.com/commander/-/commander-7.2.0.tgz#a36cb57d0b501ce108e4d20559a150a391d97ab7"
+  integrity sha512-QrWXB+ZQSVPmIWIhtEO9H+gwHaMGYiF5ChvoJ+K9ZGHG/sVsa6yiesAD1GC/x46sET00Xlwo1u49RVVVzvcSkw==
+
 concat-map@0.0.1:
   version "0.0.1"
   resolved "https://registry.yarnpkg.com/concat-map/-/concat-map-0.0.1.tgz#d8a96bd77fd68df7793a73036a3ba0d5405d477b"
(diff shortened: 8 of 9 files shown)


app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,116 @@
...
+        padding: [23, 10, 0, 0]
+      }],
+      [Tooltip, {
+        text: (date, value) => getTooltipText(date, value, locale)
+      }]
+    ]);
+  } catch (error) {
+    console.error("Error during Cal-Heatmap initialization:", error);
+  }
+});
```
HolgerJeromin, line 76:
```suggestion
}, {once: true});
```

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,116 @@
...
+
+
+  const heatmapData = heatmapElement.dataset.heatmap ? JSON.parse(heatmapElement.dataset.heatmap) : [];
+
+  try {
+    // eslint-disable-line no-undef
+    const cal = new CalHeatmap();
+    const applicationData = $("head").data();
+    const locale = applicationData.locale;
+    const theme = getTheme();
```
HolgerJeromin, line 18: We can watch for color theme with 
https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/change_event

app/assets/javascripts/heatmap.js
kcne: Fixed. Thanks for the suggestion

app/assets/javascripts/heatmap.js
kcne: Refactored the code accordingly so it takes into account if the theme of the browser is changed. Thank you for pointing this out

HolgerJeromin:
Oh, I forgot about this. 
But the user can now set his "Preferred Website Color Scheme" which should be the basis of the theme not the browser setting.

gravitystorm:
What's the source(s) of all the javascript in `vendor/`? I would expect to see `Vendorfile` updated, or preferably if these are sourced from e.g. node modules, then they should be included in package.json and added using the assets pipeline. If any of these are source files, i.e. written just for this project, then they should be in app/.

Also, how does the inclusion of popper interact with the popper already included? We end up with `//= require popper` twice in application.js.

gravitystorm:
> What's the source(s) of all the javascript in `vendor/`?

OK, I found the answer for the cal-heatmap part of the code, which is available as an npm module. In which case, using package.json will allow us to upgrade to new versions automatically, and would be the preferred option.

tordans:
UX thoughts:
- Do we want users to be able to disable this feature in their profile? – I am not a fan of this kind of configurations, but we do give a lot more visibility to once activity with this that we had before and the page is public (AKA no login required).
- How do we want to handle the "never contributed" case? Showing the heatmap in this case would suggest "this user might have contributed changes before that time…". – I think it would be best to not show it or show a placeholder instead.

Code comments:
- Do we want to add those JS files to the application js or create a separate contribution_heatmap js for this? https://github.com/openstreetmap/openstreetmap-website/pull/5402/files#diff-0e145ba000e96b1a45d230275e350e4e8514efb4db7dbcfa228fe1003896c9b4R18-R22

tomhughes:
I've tested the performance of the underlying query and using the user with the most changesets in the last year (just over 50 thousand) it took a fraction over one second which is probably acceptable.

That result (which was a full 365 days of data) serialised to just under 10Kb which is a fair chunk to be caching but hopefully most entries will be smaller than that.

kcne:
> Oh, I forgot about this. But the user can now set his "Preferred Website Color Scheme" which should be the basis of the theme not the browser setting.

Thank you for the suggestion. I’ll refactor accordingly to ensure alignment with the preferred color scheme setting.



> It is a very minor nitpick, but is it possible to dynamically determine what's the first weekday based on the user's locale and adjust the heatmap accordingly? [Intl.Locale.prototype.getWeekInfo()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo) could be used. It isn't currently available in Firefox but all other browsers have it.
> 
> P.S. Added [bug 1937867](https://bugzilla.mozilla.org/show_bug.cgi?id=1937867) to Firefox's bug tracker.

I tried dynamically adjusting the start and end weekdays using [ghDay](https://cal-heatmap.com/docs/options/subDomain#ghday) as the subdomain, but it rounds to the first and last weeks of the month, making implementation tricky. Accommodating this would require significant effort, which may not be worth it for such a minor adjustment.



> What's the source(s) of all the javascript in `vendor/`? I would expect to see `Vendorfile` updated, or preferably if these are sourced from e.g. node modules, then they should be included in package.json and added using the assets pipeline. If any of these are source files, i.e. written just for this project, then they should be in app/.

To be honest I wasn’t  too familiar with how our assets pipeline works,so I just added the `cal-heatmap` library and its dependencies in `vendor/` and added required code in `manifest.js` and `application.js` to make it load correctly. If using `package.json` is preferred, I can update the PR accordingly. I'm not sure if there is some additional config needed for this to be accomplished or would updating the `package.json` file would be enough.

Thank you for pointing out the existing `popper` inclusion. I’ll remove the duplicate entry from `application.js` and the `vendor` folder.

AntonKhorev:
> Separate JS Files: There's already a heatmap.js file containing heatmap-related functionality, included in application.js for precompilation.

How is it included in `application.js` if you're including `heatmap.js` [here](https://github.com/openstreetmap/openstreetmap-website/pull/5402/files#diff-78ff736409d758722403cce31873ba803b3fb526d37398ca21caa557b54dfd95R3)? But `d3` and `cal-heatmap` are added to `application.js`, bringing its size up to ~400kb from ~250kb.

AntonKhorev:
CSP violations, days of week won't be visible if not resolved:

> Content-Security-Policy: (Report-Only policy) The page’s settings would block an inline style (style-src-attr) from being applied

image

AntonKhorev:
> I've updated the code to account for the "Preferred Website Color Scheme." Theme settings should now be working correctly.

It's not going to work correctly because our *Auto* option follows `prefers-color-scheme` which can change any time. If it does, you'll get for example this:

image

Looks like cal-heatmap doesn't have a proper auto scheme ["as not all websites support dark/light mode"](https://cal-heatmap.com/docs/options/theme).

You'd have to then setup a listener like [this](https://github.com/openstreetmap/openstreetmap-website/pull/4777/commits/e35141b2d1c1733d9a15a01c08b796e4edda6b4e#diff-4fe14e869cf17b068e1bea009a55b094fed13f0cad85337cf4769f32b40a01b5R11-R13) but without using Leaflet of course and redraw the heatmap.

Same is true for #5426 by the way.

hlfan:
> You'd have to then setup a listener like [this](https://github.com/openstreetmap/openstreetmap-website/pull/4777/commits/e35141b2d1c1733d9a15a01c08b796e4edda6b4e#diff-4fe14e869cf17b068e1bea009a55b094fed13f0cad85337cf4769f32b40a01b5R11-R13) but without using Leaflet of course and redraw the heatmap.
> 
> Same is true for #5426 by the way.

Any other change to the theme of the page without reloading will break most of the CSS theming anyway since the stylesheets were split in #5362 by the way.

It would be great to have a global leaflet-independent watcher where clients can add themselves to a callback list.

AntonKhorev:
> Any other change to the theme of the page without reloading will break most of the CSS theming anyway since the stylesheets were split in #5362 by the way.

You don't need to change the theme of the page, neither here, nor in https://github.com/openstreetmap/openstreetmap-website/pull/5426.

> It would be great to have a global leaflet-independent watcher where clients can add themselves to a callback list.

Leaflet code is just a wrapper for `addEventListener`. You can do the same with jQuery.

hlfan:
> You don't need to change the theme of the page

I did that for testing, but since listening increased the complexity unnecessarily, I left that for a second pass.

But suppose the site needs a theme watcher anyway because the cal-heatmap doesn't want to let CSS do the color mixing theme-agnostically. In that case, instead of having two listeners, a common solution that includes MutationObservers outside of jQuery would be the one with the least overhead.

AntonKhorev:
Why MutationObservers?

AntonKhorev:
Also the conditions under which watching `prefers-color-scheme` are different for this PR and https://github.com/openstreetmap/openstreetmap-website/pull/5426. Here you need to watch if the website theme is auto, in https://github.com/openstreetmap/openstreetmap-website/pull/5426 you need to watch when the website and the map themes are both auto.

hlfan:
The MutationObserver thought came from my tinkering with the site locally wanting to reduce the reloads, forgetting that a toggle if added would already have onchange events.

> So maybe you don't want one global watcher.

Yeah, I kinda missed that... Still, having a similar style on such similar functions and having them moved to some unified location would be a nice thing to have, currently, they could be much more different. But that's more of a refactor thing if both are merged.

kcne:
> > Separate JS Files: There's already a heatmap.js file containing heatmap-related functionality, included in application.js for precompilation.
> 
> How is it included in `application.js` if you're including `heatmap.js` [here](https://github.com/openstreetmap/openstreetmap-website/pull/5402/files#diff-78ff736409d758722403cce31873ba803b3fb526d37398ca21caa557b54dfd95R3)? But `d3` and `cal-heatmap` are added to `application.js`, bringing its size up to ~400kb from ~250kb.

I've moved the import to `heatmap.js`. Should be resolved now.



> > I've updated the code to account for the "Preferred Website Color Scheme." Theme settings should now be working correctly.
> 
> It's not going to work correctly because our _Auto_ option follows `prefers-color-scheme` which can change any time. If it does, you'll get for example this:
> 
> image
> 
> Looks like cal-heatmap doesn't have a proper auto scheme ["as not all websites support dark/light mode"](https://cal-heatmap.com/docs/options/theme).
> 
> You'd have to then setup a listener like [this](https://github.com/openstreetmap/openstreetmap-website/pull/4777/commits/e35141b2d1c1733d9a15a01c08b796e4edda6b4e#diff-4fe14e869cf17b068e1bea009a55b094fed13f0cad85337cf4769f32b40a01b5R11-R13) but without using Leaflet of course and redraw the heatmap.
> 
> Same is true for #5426 by the way.

Added the logic to handle the auto prefered-color-changes. Should be resolved now as well.





> CSP violations, days of week won't be visible if not resolved:
> 
> > Content-Security-Policy: (Report-Only policy) The page’s settings would block an inline style (style-src-attr) from being applied
> 
> image

I'm working on resolving this CSP violations. They are from `CalendarLabel` plugin. Right now I'm trying to generate nonce for the .js file and use that to validate and transpass the CSP violation. I'm not sure if it is the right approach for this since I never had experience dealing with similar stuff before. If there is a more elegant and compact solution, please feel free to point it out.

Thank you.

hlfan:
> Added the logic to handle the auto prefered-color-changes. Should be resolved now as well.

For the wrong one though. Another copy of the html's bs-theme attribute isn't necessarily helpful, and the media query is still not being listened to.

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,125 @@
...
+    }]
+  ]);
+}, { once: true });
+
+function getThemeFromColorScheme(colorScheme) {
+  if (colorScheme === "auto") {
+    return window.matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light";
+  }
+  return colorScheme; // Return "light" or "dark" directly if specified
+}
```
kcne, line 89: I appreciate your feedback @hlfan. From my understanding, the primary issue was that when the “Auto” option is enabled, changes in the system’s prefers-color-scheme weren’t being reflected. With the new matchMedia logic, however, we’re directly listening for those changes, which triggers a re-draw of the heatmap whenever the user’s color scheme switches.

If you’re still seeing a scenario where the theme isn’t updating correctly, please let me know so I can investigate further. Otherwise, the current solution should correctly handle automatic theme changes.

HolgerJeromin: `getThemeFromColorScheme` returns the correct theme. Yes. 
But it is only called on load (`DOMContentLoaded`) and not when the user changes operating system theme.

Missing something like:
`window.matchMedia("(prefers-color-scheme: dark)").addEventListener("change", xy);`

hlfan:
@HolgerJeromin I'd recommend `window.matchMedia("...").onchange = xy` due to the [depreciation](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener) of `matchMedia.addEventListener`.

kcne:
I need to test this once again and get back to you. I guess you are right since there is some strange behaviour right now that I'm currently looking into.

HolgerJeromin:
@hlfan  This deprecates `addListener` (old IE invention) not `addEventListener`

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,134 @@
...
+
+  let cal = new CalHeatmap();
+  renderHeatmap(theme);
+
+  if (colorScheme === "auto") {
+    mediaQuery.addEventListener("change", (e) => {
+      const newTheme = e.matches ? "dark" : "light";
+      renderHeatmap(newTheme);
+    });
+  }
```
kcne, line 35: @hlfan Sorry about confusion here. 
In the process of rebasing the commit for latest changes I forgot to pop stashed changes so the file version didn't reflect last changes. Thanks for catching that and pointing it out. 

This is the updated version. It's working as expected in my firefox browser locally now. Please let me know if there is something else I should modify here. 

Thank you.

simonpoole:
Nice ....... but, sorry to be a spoil sport here, this changes the information available on the user page significantly, the likely solution would be to make it available only to logged in users. 

Both the LWG and EWG should be informed and https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services#User_page_(done) updated to reflect the additional feature.

PS: I'm not quite sure why the user page item is marked as done, when it obviously isn't, but that is OT.

AntonKhorev:
@simonpoole This is not something you should worry about because it will take one `if` statement to remove the entire thing unlike let's say feeds.

app/assets/javascripts/application.js
```diff
@@ -15,6 +15,11 @@
 //= require matomo
 //= require richtext
 //= require qs/dist/qs
+//= require d3
+//= require cal-heatmap
+//= require calendar-label
+//= require popper
+//= require tooltip
```
AntonKhorev: You don't need to put these scripts here because you remove them later.

app/controllers/users_controller.rb
```diff
@@ -53,6 +53,23 @@ def show
     if @user &&
        (@user.visible? || current_user&.administrator?)
       @title = @user.display_name
+
+      one_year_ago = 1.year.ago
```
AntonKhorev, line 57: Is this done to freeze the starting point because we're not sure when the block below runs? Then why is the same not done with the ending point? And shouldn't it be truncated to a day?

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,134 @@
...
+      data: {
+        source: heatmapData,
+        type: "json",
+        x: "date",
+        y: "total_changes"
+      },
+      scale: {
+        color: {
+          type: "threshold",
+          range: currentTheme === "dark" ? rangeColors : rangeColors.reverse(),
```
AntonKhorev, line 73: `rangeColors.reverse()` reverses the array in place, probably not what you want.

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,134 @@
...
+  const translations = I18n.translations[locale] || I18n.translations.en;
+  const dateTranslations = translations && translations.date;
+  const monthNames = dateTranslations && dateTranslations.month_names;
+
+  const months = monthNames || [];
+  const monthName = months[jsDate.getMonth() + 1] || `${jsDate.getMonth + 1}.`;
+  const day = jsDate.getDate();
+  const year = jsDate.getFullYear();
+
+  const localizedDate = `${day}. ${monthName} ${year}.`;
```
AntonKhorev, line 130:
```suggestion
  const localizedDate = I18n.l("date.formats.long", date);
```

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,134 @@
...
+  const heatmapElement = document.querySelector("#cal-heatmap");
+
+  if (!heatmapElement) {
+    console.warn("Heatmap element not found in the DOM.");
+    return;
+  }
+
+  const heatmapData = heatmapElement.dataset.heatmap ? JSON.parse(heatmapElement.dataset.heatmap) : [];
+  const colorScheme = heatmapElement.dataset.siteColorScheme || "auto";
+  const locale = $("head").data().locale;
```
AntonKhorev, line 19: If `application.js` is initialized, the locale is available in `I18n.locale`.

app/assets/javascripts/heatmap.js
HolgerJeromin:
```suggestion
          range: currentTheme === "dark" ? rangeColors : Array.from(rangeColors).reverse(),
```
would solve that.

simonpoole:
> @simonpoole This is not something you should worry about because it will take one `if` statement to remove the entire thing unlike let's say feeds.

Oh, I'm not concerned at all. I was just pointing this out for good form, aka I saw something that I know is wrong, and I'm giving the people here a fighting chance to do the right thing.

app/assets/javascripts/heatmap.js
AntonKhorev: I think the entire heatmap can be delayed by putting it in turbo frame or maybe some collapsible section (maybe for performance reasons). You need `I18n` initialized before anyway, I don't know how you're going to translate dates correctly otherwise. Although popup dates seem to work even if `I18n` is not initialized before the heatmap because they are constructed later.

app/assets/javascripts/application.js
kcne: This should be removed. Thanks.

app/assets/javascripts/heatmap.js
kcne: Resolved now. Thanks.

app/assets/javascripts/heatmap.js
kcne: This simplifies code my code a lot. Thank you very much sir

app/controllers/users_controller.rb
kcne: This wasn’t about freezing; it should run each time the user profile page is accessed. I guess I left it like that by accident. The idea was to fetch the changeset data from one year ago up until the current date. The thought was that starting one year ago would naturally include changesets up to this moment, effectively capturing the data for the past year.

However, I’ve now refactored the code to explicitly set both the starting and ending points. This makes the date range more declarative and ensures better clarity and consistency, while also eliminating potential ambiguities.

kcne: Just double check if it looks alright before resolving comment. Thank you.

app/assets/javascripts/application.js
AntonKhorev: This should be removed but this is not yet removed because this commit is still in the branch.

kcne: You were right, I was overriding changes in the later commit, instead of just updating this commit. It's fixed now. Thank you.

app/assets/javascripts/heatmap.js
kcne: I noticed you opened #5538. Are the changes there related to the initialization concerns mentioned here?  

Regarding delaying the heatmap or using a turbo frame/collapsible section for performance, I’m not sure about the best metrics to determine what's acceptable in terms of load time/size or resource usage. Here's the network tab from Chrome when testing locally with the cache disabled: 

Screenshot 2025-01-24 at 17 20 16  

If there are specific metrics or scenarios you'd recommend testing to ensure this is performant and I18n is properly initialized, I’d be happy to try those out.

Thank you.

kcne:
Updated the code again, resolved some issues raised here. Few remarks:

> CSP violations, days of week won't be visible if not resolved:
> 
> > Content-Security-Policy: (Report-Only policy) The page’s settings would block an inline style (style-src-attr) from being applied
> 
> image

The issue was caused by the `CalendarLabel` plugin. Given the simplicity of our use case, I opted to convert the labels to use plain HTML/CSS instead of introducing custom CSP policies for controller/view/JS files. Labels should now display correctly, and the CSP violations should be resolved. Please let me know if there’s anything I might have missed.

> Is it the intended effect that nothing gets rendered in case the user made no edits in the span of one year?

The heatmap now renders regardless of whether the user has made edits in the past year. If there’s interest in hiding the heatmap for users with no contributions, I can refactor it to include a fallback or placeholder message instead. Let me know your preference.

tomhughes:
> I'm working on resolving this CSP violations. Right now I'm trying to generate nonce for the .js file and use that to validate and transpass the CSP violation. I'm not sure if it is the right approach for this since I never had experience dealing with similar stuff before. If there is a more elegant and compact solution, please feel free to point it out.

Using nonces should be an absolute last resort though it is possible if we need to - it is enabled for styles and the nonce is available in the `csp-nonce` meta tag in the document head.

Much better to use a proper stylesheet and just add and remove classes if you can though.

kcne:
> > I'm working on resolving this CSP violations. Right now I'm trying to generate nonce for the .js file and use that to validate and transpass the CSP violation. I'm not sure if it is the right approach for this since I never had experience dealing with similar stuff before. If there is a more elegant and compact solution, please feel free to point it out.
> 
> Using nonces should be an absolute last resort though it is possible if we need to - it is enabled for styles and the nonce is available in the `csp-nonce` meta tag in the document head.
> 
> Much better to use a proper stylesheet and just add and remove classes if you can though.

I agree, Tom. Since it was an external dependency, I removed it and implemented the labels using plain HTML and CSS instead.

app/assets/config/manifest.js
```diff
@@ -7,13 +7,17 @@
...
 //= link_tree ../../../vendor/assets/iD/iD/img
 //= link_directory ../../../vendor/assets/iD/iD/data .json
 //= link_directory ../../../vendor/assets/iD/iD/locales .json
 //= link_directory ../../../vendor/assets/iD/iD/mapillary-js .css
 //= link_directory ../../../vendor/assets/iD/iD/mapillary-js .js
 //= link_directory ../../../vendor/assets/iD/iD/pannellum .js
 //= link_directory ../../../vendor/assets/iD/iD/pannellum .css
+//= link_directory ../../../vendor/assets/d3 .js
+//= link_directory ../../../vendor/assets/cal-heatmap .js
+//= link_directory ../../../vendor/assets/cal-heatmap .css
```
AntonKhorev, line 21: To address https://github.com/openstreetmap/openstreetmap-website/pull/5402#issuecomment-2549612467 you'd do something like https://github.com/AntonKhorev/openstreetmap-website/commit/7ac349e61d57bf568b55c13a44f0fd0b16261ae3. Except of course you'd start with installing it using `yarn add cal-heatmap` and you wouldn't copy dependencies to `vendor` manually.

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,104 @@
...
+      [Tooltip, {
+        text: (date, value) => getTooltipText(date, value, locale)
+      }]
+    ]);
+  }
+});
+
+function getThemeFromColorScheme(colorScheme) {
+  if (colorScheme === "auto") {
+    return window.matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light";
```
AntonKhorev, line 85: You already have `mediaQuery` defined. It can be used here:
```suggestion
    return mediaQuery.matches ? "dark" : "light";
```

For this to work, declarations have to be moved around. Maybe this function doesn't need to be declared. You already have `if (colorScheme === "auto") ...` above that is not inside a function.

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,104 @@
...
+      scale: {
+        color: {
+          type: "threshold",
+          range: currentTheme === "dark" ? rangeColors : Array.from(rangeColors).reverse(),
+          domain: [10, 20, 30, 40]
+        }
+      }
+    }, [
+      [Tooltip, {
+        text: (date, value) => getTooltipText(date, value, locale)
```
AntonKhorev, line 77: `getTooltipText` doesn't have the `locale` parameter. And you probably don't need to pass `locale` around to other places, given that you can read it from `I18n.locale`.

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,104 @@
...
+  return colorScheme;
+}
+
+function getMonthNameFromTranslations(locale, monthIndex) {
+  const translations = I18n.translations[locale] || I18n.translations.en;
+  const date = translations && translations.date;
+  const abbrMonthNames = date && date.abbr_month_names;
+
+  const months = abbrMonthNames || [];
+  return months[monthIndex + 1] || "";
```
AntonKhorev, line 96:
```suggestion
  return I18n.strftime(new Date(2000, monthIndex, 15), "%b");
```

app/assets/stylesheets/common.scss
```diff
@@ -1003,3 +1003,23 @@ img.trace_image {
...
+  padding-top: 21px;
+  z-index: 1;
+  overflow-y: hidden;
+}
+
+.heatmap {
+  margin-left: 50px;
+  height: 100%;
+  z-index: 0;
+}
```
AntonKhorev, line 1025: Why is it here and not in `heatmap.css`?

app/assets/stylesheets/common.scss
```diff
@@ -1003,3 +1003,23 @@ img.trace_image {
...
+.heatmap-wrapper {
+  position: relative;
+  height: 130px;
+}
+
+.heatmap-labels {
+position: absolute;
+  top: 2px;
+  left: 17px;
+  padding-top: 21px;
```
AntonKhorev, line 1016: Why do you need both `top` and `padding-top`?

app/assets/stylesheets/common.scss
```diff
@@ -1003,3 +1003,23 @@ img.trace_image {
...
+  z-index: 1;
+  overflow-y: hidden;
+}
+
+.heatmap {
+  margin-left: 50px;
+  height: 100%;
+  z-index: 0;
```
AntonKhorev, line 1024: Do these two properties do anything?

app/assets/config/manifest.js
kcne: Alright, fixed now and rebased the commits accordingly. Thank you for putting effort into demonstrating how to properly do this!

app/assets/javascripts/heatmap.js
kcne: You were completely right. I tried removing the method, but instead of doing two things in one method or in if statement i just refactored the method and moved it around a bit. Thank you.

app/assets/javascripts/heatmap.js
kcne: Fixed. Removed locale altogether since as you pointed out I can use I18n directly. Thank you.

app/assets/javascripts/heatmap.js
kcne: This is great, makes code much cleaner and shorter. Removed method altogether and just used the code you provided in the text callback. Thank you.

app/assets/stylesheets/common.scss
kcne: I had issues while making this work on UI with positioning. Fixed now, removed padding top and relied on top property completely instead. Everything works properly. Marking as resolved.

app/assets/stylesheets/common.scss
kcne: I thought they did as precaution but they don't. Removed now and everything works ok. Thank you.

app/assets/stylesheets/common.scss
kcne: I decided to remove `heatmap.css` completely since we opted to using npm package and linking stylesheet inside `show.html.erb` directly. Since it's only 18 lines of code I think we can leave it in `common.scss` for now. If you still think I should move it to separate file instead, just let me know.

app/views/users/show.html.erb
```diff
@@ -1,3 +1,7 @@
+<% content_for :head do %>
+  <%= stylesheet_link_tag "heatmap" %>
+  <%= javascript_include_tag "heatmap" %>
```
AntonKhorev: The linked files don't exist yet. And the stylesheet won't exist later too.

app/views/users/show.html.erb
```diff
@@ -233,6 +237,12 @@
 
 <div class="richtext text-break clearfix"><%= @user.description.to_html %></div>
 
+<% if @heatmap_data.present? %>
+  <div class="overflow-auto" style="max-height: 500px;">
+    <%= tag.div(:id => "cal-heatmap", :data => { :heatmap => @heatmap_data.to_json }) %>
+  </div>
+<% end %>
```
AntonKhorev: You overwrite this later almost entirely. I'd keep `if` because we will not output the heatmap [under some conditions](https://github.com/openstreetmap/openstreetmap-website/pull/5402#issuecomment-2602166082).

`style="max-height: 500px;"` shouldn't be here.

app/views/users/show.html.erb
```diff
@@ -237,11 +237,29 @@
 
 <div class="richtext text-break clearfix"><%= @user.description.to_html %></div>
 
-<% if @heatmap_data.present? %>
-  <div class="overflow-auto" style="max-height: 500px;">
-    <%= tag.div(:id => "cal-heatmap", :data => { :heatmap => @heatmap_data.to_json }) %>
+<div class="small">
```
AntonKhorev, line 240: What does `class="small"` do?

app/views/users/show.html.erb
```diff
@@ -237,11 +237,29 @@
...
+  <div class="row">
+    <div class="col overflow-auto">
+      <div id="cal-heatmap-wrapper" class="heatmap-wrapper">
```
AntonKhorev, line 243: Why wouldn't you use a single flex container that puts weekday labels next to the heatmap? No position absolute/relative, probably only one pixel value of top padding for the labels to align them.

app/views/users/show.html.erb
kcne: Removed from this commit and added later when the files are added. Thank you.

app/views/users/show.html.erb
kcne: Okay, kept the if statement as it is now and also removed styling mentioned and `console.warn` statment from the javascript code. Thank you.

kcne: When we decide the conditions for renedering we can refactor the if statement as necessary. Thank you.

app/views/users/show.html.erb
kcne: Removed that div entirely, thank you. I thought  it would help make it responsive on small screens in bootstrap grid, but apparently it doesn't do anything in this case.

test/controllers/users_controller_test.rb
```diff
@@ -333,4 +333,97 @@ def test_auth_failure_callback
...
+    # There should be no entries in heatmap data
+    assert_empty assigns(:heatmap_data)
+  end
+
+  def test_heatmap_rendering
+    # Test user with no changesets
+    user_without_changesets = create(:user)
+    get user_path(user_without_changesets)
+    assert_response :success
+    assert_select "div#cal-heatmap", 1
```
AntonKhorev, line 411: This fails because `@heatmap_data.present?` is false when `@heatmap_data` is an empty array.

app/views/users/show.html.erb
```diff
@@ -233,6 +237,30 @@
...
+          <li>&nbsp;</li>
+          <li><%= t("date.abbr_day_names")[3] %></li>
+          <li>&nbsp;</li>
+          <li><%= t("date.abbr_day_names")[5] %></li>
+          <li>&nbsp;</li>
+        </ul>
+        <!-- Heatmap -->
+        <div id="cal-heatmap" class="heatmap"
+          data-heatmap="<%= @heatmap_data.to_json %>"
+          data-site_color_scheme="<%= preferred_color_scheme(:site) %>">
```
AntonKhorev, line 257: The color scheme is already present in `data-bs-theme` on `<html>` ("auto" if that attribute is missing).

test/controllers/users_controller_test.rb
kcne: Updated the test to reflect latest changes. Thank you.

app/views/users/show.html.erb
kcne: Refactored the code to use flexbox instead. It is indeed much simpler, as I could get rid of some custom css classes. Thanks once again.

config/locales/en.yml
```diff
@@ -3300,6 +3300,10 @@ en:
       show_address: Show address
       query_features: Query features
       centre_map: Centre map here
+    heatmap:
+      tooltip:
+        contributions: "%{count} contributions on %{date}"
```
AntonKhorev, line 3305: You need `one` and `other` versions with `%{count}`.

kcne: Updated keys and js code respectively. Now we got `no_contributions`, `contributions` with `%{count}` and `one_contribution`.

config/locales/en.yml
```diff
@@ -3302,6 +3302,11 @@ en:
       show_address: Show address
       query_features: Query features
       centre_map: Centre map here
+    heatmap:
+      tooltip:
+        one_contribution: "1 contribution on %{date}"
+        contributions: "%{count} contributions on %{date}"
```
AntonKhorev, line 3308:
```suggestion        
        contributions:
          one: "%{count} contribution on %{date}"
          other: "%{count} contributions on %{date}"
```

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,99 @@
...
+  renderHeatmap();
+});
+
+function getTooltipText(date, value) {
+  const localizedDate = I18n.l("date.formats.long", date);
+
+  if (value > 0) {
+    return value === 1 ?
+      I18n.t("javascripts.heatmap.tooltip.one_contribution", { date: localizedDate }) :
+      I18n.t("javascripts.heatmap.tooltip.contributions", { count: value, date: localizedDate });
```
AntonKhorev, line 95:
```suggestion
    return I18n.t("javascripts.heatmap.tooltip.contributions", { count: value, date: localizedDate });
```

kcne: Resolved.

config/locales/en.yml
kcne: Resolved.

kcne:
The test failing here doesn't have nothing to do with the changes introduced in this PR. I tried force pushing again to rerun checks but again got failing test on different version of Ruby this time, which is weird. Is there some kind of bug in CI/CD right now or should I push again to rerun checks?

yarn.lock
```diff
@@ -4,24 +4,24 @@
 
 "@aashutoshrathi/word-wrap@^1.2.3":
   version "1.2.6"
-  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
+  resolved "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz"
```
AntonKhorev, line 7: Now there's a ton of `registry.yarnpkg.com` -> `registry.npmjs.org` changes.

kcne: Must have messed this up during conflict. I will try to restore the file and do rebase again. Thank you.

AntonKhorev: Now there are no updates to `yarn.lock`.

kcne: Should be resolved now.

kcne:
I've resolved the conflicts in `yarn.lock` again. Since changes keep getting merged into master, I keep running into conflicts. Are there any remaining issues to address, or is this good to merge now?

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,100 @@
+//= require d3/dist/d3
+//= require cal-heatmap/dist/cal-heatmap
+//= require popper
```
tomhughes, line 3: I don't understand how this works? Shouldn't it be `@popperjs/core` or something here? How is it finding `popper` and is it actually finding the version we think it is?

kcne: This import is the same as the one inside `application.js` and was not introduced in this PR. It's provided by the [popper_js gem](https://github.com/glebm/popper_js-rubygem), which looking at the repo automatically supplies popper.js assets to rails.

AntonKhorev:
With #5654 it's possible to translate the tooltips with language-specific pluralizations:

```
    heatmap:
      tooltip:
        contributions:
          one: "%{count} правка %{date}"
          few: "%{count} правки %{date}"
          many: "%{count} правок %{date}"
          other: ""
        no_contributions: "Нет правок %{date}"
```

image image image

Otherwise only `one` for `count==1` and `other` rules are supported.

Do we merge #5654 first?

AntonKhorev (approved):

kcne:
Now that #5654 was merged, I think current implementation is already setup to support this. Please let me know if i need to add anything else to support it. Thank you.

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,100 @@
...
+    cal = new CalHeatmap();
+
+    cal.paint({
+      itemSelector: "#cal-heatmap",
+      theme: currentTheme,
+      domain: {
+        type: "month",
+        gutter: 4,
+        label: {
+          text: (timestamp) => I18n.strftime(new Date(2000, new Date(timestamp).getMonth(), 15), "%b"),
```
tomhughes, line 35: If we've constructed a date from the timestamp why do you then extract the month from that and build a new date? If the format only uses the month the rest of the date doesn't matter?

It might a bit faster to just lookup `I18n.t("date.abbr_month_names")` once before the paint call and then index that array with the month number rather than calling strftime repeatedly - just beware of 0/1 based indexing!

tomhughes (changes requested):

app/assets/javascripts/heatmap.js
kcne: You're right. Updated the code. Thank you.

app/assets/javascripts/heatmap.js
```diff
@@ -0,0 +1,102 @@
+//= link cal-heatmap/dist/cal-heatmap.css
```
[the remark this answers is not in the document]
tomhughes: I think the best solution is probably to create `app/assets/heatmap.scss` which can require that, and also include the heap rule you added to `common.scss` and then have the view load `heatmap` rather than `cal-heatmap/dist/cal-heatmap`.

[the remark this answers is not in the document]
kcne: Thanks for the feedback, Tom. Since the code inside `common.scss` was unnecessary, I removed it completely, linked the `.css` file to `manifest.js`, and included it in the view directly. Let me know if any further modifications are needed.

tomhughes (approved):
I might still have added a local stylesheet with a require just so there was consistency between the javascript and the styles but it doesn't really matter so I think this is good to go now - thanks for your work on it!

verhovsky:
You need to hide days that haven't happened, like GitHub. Every time you open it, it looks like the user stopped contributing a few weeks ago.

HolgerJeromin:
> You need to hide days that haven't happened, like GitHub. Every time you open it, it looks like the user stopped contributing a few weeks ago.

Would be nice. 
But I also did not found a way to do that with the used library...

PR #5403, opened 2024-12-16
Add date and url settings for offline message

AntonKhorev:
The current message doesn't say much about how long the api/db is expected to stay down.

on map layout:
image

on site layout:
image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -259,12 +259,7 @@ class ApplicationController < ActionController::Base
 
     request.content_security_policy = policy
 
-    case Settings.status
-    when "database_offline", "api_offline"
-      flash.now[:warning] = t("layouts.osm_offline")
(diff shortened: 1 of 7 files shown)


tomhughes (commented):
I had been thinking we should make the message configurable because in this case it wasn't really database maintenance but that is problematic from a translation point of view.

Maybe we should just remove the references to database maintenance in favour of this solution with a URL to an explanatory notice?

HolgerJeromin:
> Maybe we should just remove the references to database maintenance in favour of this solution with a URL to an explanatory notice?

Perhaps this is generic enough:
```
The OpenStreetMap database is currently in read-only mode while essential maintenance work is carried out.

AntonKhorev:
Removed "database" from "maintenance":
image image

nenad-vujicic (approved):
This works great on my side, thanks.

tomhughes:

PR #5408, opened 2024-12-17
Show 'api offline' message on new note page

### LINKED ISSUE #5404: Notes UI is still presented, despite being "read-only"
### URL

https://www.openstreetmap.org/note/new#map=17/51.060265/3.721265&layers=N

0. Get OSM.org to be in "read-only" mode. (For example, sabotage the upstream router)
1. Go to OSM.org
2. Click the "new note button"
3. The new-note UI will still happily popup, as if nothing is wrong
4. Write a comment and click "Create new note"
5. Nothing happens

It would be nice if no form popped up, but a small "sorry, we cannot add notes right now" would be shown instead.

_No response_

AntonKhorev:
Show this on `/note/new` if the api is not online:
image

This is shown in read-only mode, offline mode already has a different message "Database offline for maintenance".

Fixes #5404.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -44,5 +44,7 @@ class NotesController < ApplicationController
     render :template => "browse/not_found", :status => :not_found
   end
 
-  def new; end
+  def new
+    render :action => :new_readonly if api_status != "online"
+  end
 end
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me, thanks.

PR #5410, opened 2024-12-17
Add standard patch/put api message update action

AntonKhorev:
A standard HTTP verb for update action in Rails is PATCH/PUT, but for messages API [it was made POST for some reason](https://wiki.openstreetmap.org/wiki/Messaging_API_proposal#Update_read_status_of_a_message). Here I'm adding PATCH/PUT. I'm keeping POST outside of the resources, but we should deprecate it.

Also I'm moving the resources to the api namespace. This allows to remove `:as`.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -77,17 +77,17 @@ OpenStreetMap::Application.routes.draw do
         put "" => "user_preferences#update_all", :as => ""
       end
     end
+  end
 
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

nickawyse-spec:
John W. Hinman

PR #5412, opened 2024-12-18
Pass correct URL to Facebook sharing

tomhughes:
It was trying to pass the literal text `params[:url]` as the URL due to incorrect quoting.


gravitystorm:
Merged thanks. Feels to me like some tests are missing from the original PR, which might have caught this.

PR #5413, opened 2024-12-18
Convert social share helper library into a real helper

tomhughes:
This converts the confusingly named `SocialShareButtonHelper` which is really a library, but tested by a helper test, into a real helper and moves the actual `render_social_share_buttons` helper there from the application helper.

The other methods are then private as they're only used by `render_social_share_buttons` so are no longer tested separately as they're not visible to the test but will be tested implicitly by the calls to `render_social_share_buttons` in the other tests so coverage should be the same.


gravitystorm:
Merged, thanks.

PR #5416, opened 2024-12-18
Add 'canDownloadImage' layer option

AntonKhorev:
Follows #5388. Removes hardcoded layer id checks in *Share > Image*, except for one special mapnik check.

If a layer that can't be downloaded is selected, the message shows a list of available layers:

Before:
image

After:
image

### DIFF
--- a/app/assets/javascripts/leaflet.share.js
+++ b/app/assets/javascripts/leaflet.share.js
@@ -338,7 +338,7 @@ L.OSM.share = function (options) {
 
     function update() {
       const layer = map.getMapBaseLayer();
-      var canEmbed = layer && layer.options.canEmbed;
+      var canEmbed = Boolean(layer && layer.options.canEmbed);
       var bounds = map.getBounds();
 
       $("#link_marker")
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/leaflet.share.js
```diff
@@ -411,34 +416,20 @@ L.OSM.share = function (options) {
       $("#mapnik_image_width").text(mapWidth);
       $("#mapnik_image_height").text(mapHeight);
 
-      const layerId = map.getMapBaseLayerId();
-      const layerKeys = new Map([
-        ["mapnik", "standard"],
-        ["cyclemap", "cycle_map"],
-        ["transportmap", "transport_map"]
-      ]);
+      const canDownloadImage = layer && layer.options.canDownloadImage;
```
tomhughes, line 419: This doesn't guarantee a boolean valued result - specifically if the `canDownloadImage` key is missing then the result is `undefined` not `false` and at least in Firefox calling `toggle(undefined)` toggles the current visibility instead of always hiding the element though I don't think the behaviour is defined at all by jQuery.

tomhughes (changes requested):

app/assets/javascripts/leaflet.share.js
HolgerJeromin:
```suggestion
      const canDownloadImage = !!(layer && layer.options.canDownloadImage);
```
or
```suggestion
      const canDownloadImage = Boolean(layer && layer.options.canDownloadImage);
```
[Modern js](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) would be (available across browsers since September 2020):
```suggestion
      const canDownloadImage = layer?.options.canDownloadImage ?? false;
```

AntonKhorev: Updated to `Boolean( ... )`, eslint doesn't like `!!` too.

tomhughes (approved):
That looks much better now, thanks.

Issue #5414, opened 2024-12-18
Don't use a popup window for social sharing


gravitystorm:
I think that using a popup window for the social sharing functionality isn't a great user experience, but I would like to have a more detailed discussion about the pros and cons.

In my mind, either one of these would be preferable:
* Links open in a new tab, or
* Use a [bootstrap modal dialog](https://getbootstrap.com/docs/5.3/components/modal/)

What do other people think? Why were popups originally chosen?

cc @kcne

kcne:
Popups were chosen initially, inspired by [rails-social-share-button](https://github.com/Bunlong/rails-social-share-button), which is now deprecated. I didn't thought too much about alternatives at the time. 

To be honest,  I’m not a fan of popups either. Refactoring to open links in new tabs would probably simplify the code and make maintenance easier.

For the other alternatives, using Bootstrap modals with iframes likely won’t work due to CSP restrictions on most websites. While the [Web Share API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API#browser_compatibility) is promising, it’s not yet widely supported.

Opening links in new tabs seems like the cleanest and most maintainable solution in my opinion. I would be eager to do a refactor if there is agreement on the approach.

AntonKhorev:
Anyone who wants new tabs can already open the links in new tabs by middle-clicking.

PR #5418, opened 2024-12-19
Center share popup inside the browser window

AntonKhorev:
What the existing share popup window placement does is it places the popup horizontally in the center of the screen. I don't know if the popup can get placed literally in the corner (https://github.com/openstreetmap/openstreetmap-website/issues/5414#issuecomment-2551615977), maybe with a multi-monitor setup? But it still could be a surprising location if the monitor is large enough. Here the logic is changed to place the popup in the center of the browser window. Is it less surprising?

The only alternative to a popup window is a new tab, which is already available with middle click or context menu on share links.

### DIFF
--- a/app/assets/javascripts/social_share_button.js
+++ b/app/assets/javascripts/social_share_button.js
@@ -2,8 +2,8 @@ function openShareUrl(url, initialWidth = 640, initialHeight = 480) {
   const width = Math.max(100, Math.min(screen.width, initialWidth));
   const height = Math.max(100, Math.min(screen.height, initialHeight));
 
-  const left = (screen.width / 2) - (width / 2);
-  const top = (screen.height * 0.3) - (height / 2);
+  const left = screenLeft + ((outerWidth - width) / 2);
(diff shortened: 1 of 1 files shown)


tomhughes:
It's definitely not centred for me:
image
Most likely either Firefox on linux or my Gnome/Wayland setup is ignoring the position.

tomhughes:
It works in Chromium so it must be a Firefox thing.

This is definitely better when it works anyway, so thanks.

tomhughes:
Firefox is reporting the screen size as 1920x1080 which is half the real size of my display so it's only centring in the top left quadrant :-( I assume it's some hidpi scaling thing but I'm struggling to find any mention of it anywhere on the web.

tomhughes:
Though the screen height and width don't actually affect anything if they're larger than 640x480 so actually I don't understand this code at all.

AntonKhorev:
Screen size should be reported in css pixels according to MDN docs.

Issue #5359, opened 2024-11-29
Bogus comment is added when a note is reopened


Dimitar5555:
### URL

https://www.openstreetmap.org/note/4537701

### How to reproduce the issue?

1. Press "Reopen" on a closed note

The bogus is content is likely dynamically generated, since it "leaks" the user's UI language and the current map view.

### Screenshot(s) or anything else?

Also see 
https://www.openstreetmap.org/note/3958106
https://www.openstreetmap.org/note/4482755
https://www.openstreetmap.org/note/4537279

tomhughes:
If you want to open test notes please use https://master.apis.dev.openstreetmap.org/ not the live site!

PR #5419, opened 2024-12-19
Improve note testing and text handling

AntonKhorev:
- Add test that would have failed in #5359.
- Change the code that selects the text input from `$("textarea")` to `content.find("textarea")`. This should stop the html share code from getting posted to notes, although I couldn't reproduce that happening like @SomeoneElseOSM reported it yesterday. When you press *Hide* the textarea inside the left sidebar exists and is the first one on the page, I don't know how that other textarea got selected instead.


tomhughes:
Looks good to me, thanks.

PR #5185, opened 2024-09-11
Remove :new abilities implied by :create abilities

AntonKhorev:
Similar to #5164

PR #5164, opened 2024-09-06
Remove :edit abilities implied by :update abilities

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/5012#issuecomment-2245708279. This leaves only one odd `:edit` ability on `:site`.

PR #5422, opened 2024-12-20
Replace show and index abilities with read ability

AntonKhorev:
Use [CanCanCan's `:read` alias](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/define_check_abilities.md#can-actions) instead of `:index` and `:show` abilities. Similar to https://github.com/openstreetmap/openstreetmap-website/pull/5164 and https://github.com/openstreetmap/openstreetmap-website/pull/5185.

I made a couple of exceptions:
- `Note` has `:show` and `:index` abilities in different branches. I kept them. It's probably because one is on map layout, another one is on site layout and we try to avoid redirecting to the 403 page from map layout.
- I kept `:index` for `:site`. `:index` here means an entirely different thing.

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -5,8 +5,7 @@ class Ability
 
   def initialize(user)
     can :query, :browse
-    can :show, [Node, Way, Relation]
-    can [:index, :show], [OldNode, OldWay, OldRelation]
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good to me, thanks.

PR #5423, opened 2024-12-20
Move user preferences routes to api namespace

AntonKhorev:
Done to eventually move all api routes to api namespace.

Gives `api_` prefix to path helpers.


tomhughes:
Looks good to me, thanks.

PR #5424, opened 2024-12-20
Simplify calls to api_message_path helper

AntonKhorev:
Mostly changes
```
api_message_path(:id => msg.id)
```
to
```
api_message_path(msg)
```
in tests.

PR #5425, opened 2024-12-20
Simplify calls to user block path helpers

AntonKhorev:
Similar to #5424, for `*_user_block_path` helpers.


tomhughes:
Looks good to me, thanks.

PR #5428, opened 2024-12-20
Don't map multiple paths in a single route

tomhughes:
This is deprecated in rails 8.0 and 8.1 will remove it.


AntonKhorev:
Merged, thanks.

PR #5430, opened 2024-12-20
Update to rails 8

tomhughes:
First draft of updating to rails 8 which currently needs to pull a couple of dependent gems from git as they are awaiting new releases for rails 8 compatibility.


tomhughes:
So this also requires ruby 3.2 which is going to be painful for us in production, so I've had to drop 3.1 support.

It also needs a git version of `annotate` though https://github.com/ctran/annotate_models/issues/1032 suggests that it may be dead and `annotaterb` may be taking over.

tomhughes:
I think this is probably ready for review - the only real annoyance is having to use an unreleased version of `active_record_union` but I don't think there's much point waiting any longer on that.

gravitystorm:
All looks good to me, just need to clarify what's happening with canonical_rails. This PR has the config completely commented out. Is the config no longer necessary, or is the gem not working?

tomhughes:
Oh think I forgot about that... Will have to investigate.

tomhughes:
Looks like they've done a release so I'll update that.

tomhughes:
Ah I had actually re-enabled the gem but forgot to uncomment the config again. Now fixed.

gravitystorm:
Thanks! Merged now.

PR #5433, opened 2024-12-21
Resourceful routes for users api

AntonKhorev:
Moves `user/details` and `user/gpx_files` to their own controllers.


tomhughes (commented):
This mostly looks fine. I think the only thing I'm not sure about is moving `user#details` into a separate controller - it's basically an alias for `user#show` so does it make any sense to separate it?

That said @gravitystorm has a better grasp than me of what is the canonical rails to do these things so I'd be interest to here what he thinks.

AntonKhorev:
Let's keep `user/details` non-resourceful for now, I'll try something else later.

gravitystorm:
Looks good to me. Merged, thanks!

PR #5434, opened 2024-12-22
Use resourceful routes for api messages in/outbox

AntonKhorev:


tomhughes:
Looks good to me, and the difference between inbox and outbox is clearly a bug that needs fixing anyway. Thanks.

PR #5435, opened 2024-12-23
Remove check_api_readable from api messages controller

AntonKhorev:
This check was moved to `ApiController` superclass in #4859, but `MessagesController` was written before that and merged after that in #4605.


tomhughes:
Looks good to me, thanks.

PR #5437, opened 2024-12-23
Use resourceful route for user status

AntonKhorev:
Creates `:user_status` resource with its own controller. You can `:update` this resource instead of doing `:set_status` on `User`.

"Soft destroy" `:destroy` action is also moved to this resource because it comes together with status update. It does a similar thing (changes status), has the same permissions (admin), is linked from the same place in the UI.

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -66,7 +66,8 @@ class Ability
           can [:hide, :unhide], [DiaryEntry, DiaryComment]
           can [:read, :resolve, :ignore, :reopen], Issue
           can :create, IssueComment
-          can [:set_status, :destroy], User
+
+          can [:update, :destroy], :user_status
           can [:read, :update], :users_list
           can [:create, :destroy], UserRole
         end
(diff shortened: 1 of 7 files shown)


nertc (approved):
Tested and code works properly. I like this change, this kind of separation of functionalities make code more readable and easier to work with.

gravitystorm:
I think you can `update` the `status`, but you can't `destroy` the `status`. You can `destroy` the `user`.

So I think refactoring the `set_status` is good but I think the destroy method should stay in the same place.

AntonKhorev:
> You can destroy the user.

You can't, the user is deactivated and gets some of the properties wiped. It's a status change with extras.

If you don't like `:destroy` on status, what if we do `@user.soft_destroy!` inside `update` too?

gravitystorm:
> what if we do `@user.soft_destroy!` inside `update` too?

OK, that sounds like a good compromise to me!

AntonKhorev:
Moved user soft destroy to `put user_status_path(user, :event => "soft_destroy")`.

gravitystorm:
Merged, thanks!

Issue #2245, opened 2019-06-04
Please make contact information fields, how to contact the user, email address, telegrams, various social networks https://www.openstreetmap.org


Sowa1980:
Please make contact information fields, how to contact the user, email address, telegrams, various social networks https://www.openstreetmap.org

Сделайте пожалуйста, поля для контактной информации, как связаться с пользователем, электронный адрес почты, телеграмм, различные соц сети https://www.openstreetmap.org

tomhughes:
There are already tickets requesting this.

tomhughes:
Well I thought there was anyway though I can't find them - they might be so old they're in trac rather than here.

The closest is #1274 but that's really about the social signin, and we don't expose those identities anyway.

We would need to be extremely careful (if only for GDPR compliance) that people were aware that the information was going to be public.

simonpoole:
While conceivably allowing the users to add this kind of information as part of a more general way of allowing the user to be contacted, aka not just e-mail would be a good thing, we wouldn't be able to make it _available_  to other users without introducing some kind of opt in mechanism (which in itself would have all kind of complications). 

See see https://github.com/openstreetmap/openstreetmap-website/issues/2102 https://github.com/openstreetmap/openstreetmap-website/pull/1431 https://github.com/openstreetmap/openstreetmap-website/issues/908 for related discussion.

mmd-osm:
Social networks is already there:

http://neis-one.org/2016/05/verified-osm-contributors/
https://www.openstreetmap.org/user/malenki/diary/38381

Nakaner:
Doesn't adding fields for contact information (especially social networks) weaken the personal message feature of openstreetmap.org? At the moment, we expect any mapper to be reachable via personal messages and changeset comments, don't we?

simonpoole:
@Nakaner at least previous discussion always resolved around -notifying- users or new communications via different means than just e-mail and not providing these as alternatives to the existing facilities.

Sowa1980:
And why not let the users themselves decide that this information is available or not available at his request, I personally do not mind it.

А почему не дать пользователям сами решать что бы данная информация была доступна или не доступна по его желанию, я вот лично не против этого.

nertc:
[...]

PR #5439, opened 2024-12-23
Add social profile links

nertc:
This PR addresses "Please make contact information fields, how to contact the user, email address, telegrams, various social networks https://www.openstreetmap.org" issue mentioned in the https://github.com/openstreetmap/openstreetmap-website/issues/2245

### Description
This PR adds:
- DB: Social links table, which contains social links added by the user
- Model: Parser to differentiate different kind of websites
- Controller: Logic for CRUD operations for social links
- View: Functionality to show, edit, add and delete social links
- JS: Was only used for adding and removing fields on the user side when they interact to the social links fields

Currently parser parses these websites and adds their icons which are taken from Bootstrap-Icons:
- Discord
- Facebook
- Github
- Gitlab
- Instagram
- Linkedin
- Line
- Mastodon
- Medium
- Quora
- Reddit
- Skype
- Slack
- Snapchat
- Stack Overflow
- Strava
- Substack
- Telegram
- Threads
- Tik-Tok
- Twitch
- X (Twitter)
- Vimeo
- WhatsApp
- YouTube

Discord, Line, Skype and Slack doesn't show username in the URL. To avoid showing some Id numbers on the profile page, for these cases instead of the Id only the name of the application is shown.
Applications other than those mentioned in the list show their URL on the profile page and have general web icon (globe icon).
If there is an idea that some of them should be removed or others should be added, feel free to share recommendations.

Parser was done in the Ruby to avoid adding more JS client-side logic to the website. If it is preferable, it can be moved to JS.

### How has this been tested?
There are validation and functional tests written to test the functionality. In addition to this, different kind of manual testing was done to ensure that all the icons, errors, fields and etc. were displayed correctly.

### Screenshots

Logged out:
image
image

Logged in:
image

Edit Profile page:
image
image


AntonKhorev:
Do we need another set of icons in addition to `app/assets/images/social_icons`?

app/models/social_link.rb
```diff
@@ -0,0 +1,67 @@
...
+    :vimeo => %r{vimeo.com/([a-zA-Z0-9_]+)},
+    :whatsapp => %r{wa\.me/(\d+)},
+    :youtube => %r{youtube\.com/@([a-zA-Z0-9_-]+)}
+  }.freeze
+
+  NO_USERNAME_PLATFORMS = %w[discord line skype slack].freeze
+
+  def parsed
+    URL_PATTERNS.each do |platform, pattern|
+      username = url.match(pattern)
```
AntonKhorev, line 57: It's not a username, is it?

app/views/profiles/edit.html.erb
```diff
@@ -40,6 +40,21 @@
...
+      <%= f.fields_for :social_links do |social_link_form| %>
+        <div class="social-link-fields row mb-3">
+          <%= social_link_form.text_field :url, :hide_label => true, :wrapper_class => "col-sm-8" %>
+          <%= social_link_form.check_box :_destroy, :wrapper_class => "d-none social_link_destroy" %>
+          <%= social_link_form.label :_destroy, t(".social_links.remove"), :class => "btn btn-outline-primary col-sm-1 align-self-start" %>
+        </div>
+      <% end %>
+    </div>
+
+    <%= link_to "+", "#", :id => "add-social-link", :class => "btn btn-outline-primary" %>
```
AntonKhorev, line 55: Why is it a link and not a button?

Why adding is "+" but removing is "Remove"?

app/assets/javascripts/user.js
```diff
@@ -225,4 +225,27 @@ $(document).ready(function () {
...
+      <div class="social-link-added-fields row mb-3">
+        <div class="col-sm-8">
+          <input class="form-control" type="text" name="user[social_links_attributes][${newIndex}][url]" id="user_social_links_attributes_${newIndex}_url">      
+        </div>
+        <button type="button" class="btn btn-outline-primary col-sm-1 align-self-start">${I18n.t("javascripts.profiles.edit.social_links.remove")}</button>
+      </div>
+    `);
+
+    socialLinkForm.find("button").click(function () {
+      $(this).parent().remove();
```
AntonKhorev, line 242: I press this button, this removes the link, and I'd think it's saved. But actually not, failing the validation will resurrect the link.

app/models/social_link.rb
```diff
@@ -0,0 +1,67 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
+
+class SocialLink < ApplicationRecord
+  belongs_to :user
+
+  validates :url, :presence => true, :format => { :with => URI::DEFAULT_PARSER.make_regexp(%w[http https]), :message => I18n.t("profiles.edit.social_links.http_parse_error") }
```
AntonKhorev, line 23: This doesn't check if `url` starts with `http://`.

And if you also disable CSP:
image

AntonKhorev:
When I'm making a site for myself and there's this kind of UI with *Add* buttons adding an input and *Remove* buttons next to added inputs, I don't bother doing it. I just put a `<textarea>`, it's easier to edit. In this case I'd have a textarea with each nonempty line corresponding to a link. Why don't people do this? Is it too geeky?

app/models/social_link.rb
nertc: It's a user's name shown in the URL of some websites' profile page (for example, in the `https://github.com/nertc` `nertc` is a username)

app/views/profiles/edit.html.erb
nertc: I had a different logic that needed a link, as I changed logic, I agree with you, I'll change it to button.

My idea was that `+` sign doesn't require any additional icons and is intuitive, while `Remove` needs `trash icon` to be intuitive (`-` sign doesn't always mean removal and sometimes it may be ambiguous). But I agree with you, maybe it's better to be consistent in this too. Which one do you think is better, to change `+` sign to `Add` (which will be much easier and less-code solution) or to change `Remove` to `trash icon` (which may be nicer in terms of UI, but will require trash icon and, therefore, more code)?

app/models/social_link.rb
AntonKhorev: `MatchData` object is not a username.

nertc:
> When I'm making a site for myself and there's this kind of UI with _Add_ buttons adding an input and _Remove_ buttons next to added inputs, I don't bother doing it. I just put a `<textarea>`, it's easier to edit. In this case I'd have a textarea with each nonempty line corresponding to a link. Why don't people do this? Is it too geeky?

Thanks for the review. I agree with you, if I did a website only for me, it would be a quicker solution to add a textarea and it would do the job. But for most users current solution will be more intuitive and appealing in terms of UI/UX. It's more common and therefore understandable approach (for example GitHub has the same approach on the profile page).

app/models/social_link.rb
nertc: To use more intuitive name and not to confuse "user's name" to username, name of the variable `username` was changed to `names`

app/views/profiles/edit.html.erb
nertc: Changed `+` to `Add social link`

app/models/social_link.rb
nertc: Changed `:with` to regex pattern that will check the string to start with `http://` or `https://`

nertc:
Merge conflict was fixed.

app/views/profiles/edit.html.erb
```diff
@@ -40,6 +40,21 @@
...
   </fieldset>
 
+  <fieldset class="mb-3">
+    <legend><%= t ".social_links.legend" %></legend>
+    <div id="social_links">
+      <%= f.fields_for :social_links do |social_link_form| %>
+        <div class="social-link-fields row mb-3">
+          <%= social_link_form.text_field :url, :hide_label => true, :wrapper_class => "col-sm-8" %>
+          <%= social_link_form.check_box :_destroy, :wrapper_class => "d-none social_link_destroy" %>
+          <%= social_link_form.label :_destroy, t(".social_links.remove"), :class => "btn btn-outline-primary col-sm-1 align-self-start" %>
```
AntonKhorev, line 50: `col-sm-1` is too narrow

app/models/social_link.rb
AntonKhorev: Is it more intuitive? Now it suggests there are multiple names in this variable.

app/models/social_link.rb
```diff
@@ -0,0 +1,67 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
+
+class SocialLink < ApplicationRecord
+  belongs_to :user
+
+  validates :url, :format => { :with => %r{\Ahttps?://.+\z}, :message => I18n.t("profiles.edit.social_links.http_parse_error") }
```
AntonKhorev, line 23: model is referring to controller's i18n key here

app/views/profiles/edit.html.erb
```diff
@@ -40,6 +40,21 @@
     </div>
   </fieldset>
 
+  <fieldset class="mb-3">
```
AntonKhorev, line 43: Why is the link editor placed between *avatar* and *home location*? If you don't change the order of other things on this page, this place makes the least sense to me.

app/views/social_links/_show.html.erb
```diff
@@ -0,0 +1,8 @@
+<% social_links.each do |social_link| %>
+  <p class="text-body-secondary">
+    <%= image_tag "/assets/social_link_icons/#{social_link.parsed[:platform].nil? ? 'other' : social_link.parsed[:platform]}.svg",
+                  :alt => social_link.parsed[:platform].nil? ? "other" : social_link.parsed[:platform],
+                  :class => "me-2" %>
+    <%= link_to social_link.parsed[:name], social_link.url, :class => "text-body-secondary", :rel => "nofollow me" %>
+  </p>
```
AntonKhorev, line 7: You can get line wraps like this, where the text goes below the icon:

image

app/models/social_link.rb
```diff
@@ -0,0 +1,67 @@
...
+    :snapchat => %r{snapchat\.com/add/([a-zA-Z0-9_-]+)},
+    :stackoverflow => %r{stackoverflow\.com/users/(\d+/[a-zA-Z0-9_-]+)},
+    :strava => %r{strava\.com/athletes/([a-zA-Z0-9_-]+)},
+    :substack => %r{substack\.com/@([a-zA-Z0-9_-]+)},
+    :telegram => %r{t\.me/([a-zA-Z0-9_]+)},
+    :threads => %r{threads\.net/@([a-zA-Z0-9_]+)},
+    :tiktok => %r{tiktok\.com/@([a-zA-Z0-9_]+)},
+    :twitch => %r{twitch\.tv/([a-zA-Z0-9_]+)},
+    :twitter_x => %r{(?:twitter|x)\.com/([a-zA-Z0-9_]+)},
+    :vimeo => %r{vimeo.com/([a-zA-Z0-9_]+)},
```
AntonKhorev, line 48: unescaped `.`

app/models/social_link.rb
```diff
@@ -0,0 +1,67 @@
...
+    :whatsapp => %r{wa\.me/(\d+)},
+    :youtube => %r{youtube\.com/@([a-zA-Z0-9_-]+)}
+  }.freeze
+
+  NO_USERNAME_PLATFORMS = %w[discord line skype slack].freeze
+
+  def parsed
+    URL_PATTERNS.each do |platform, pattern|
+      names = url.match(pattern)
```
AntonKhorev, line 57: matches urls like `https://evil.example.com/reddit.com/user/NotRedditor`

app/views/profiles/edit.html.erb
nertc: What do you think is a good place for it? Should it be between the description and avatar, before the description or the last one?

app/views/users/show.html.erb
```diff
@@ -231,10 +231,15 @@
...
-    <%= link_to t(".edit_profile"), edit_profile_path, :class => "btn btn-outline-primary" %>
+<div class="row">
+  <div class="col-sm-3">
+    <%= render "social_links/show",
+               :social_links => @user.social_links %>
+    <% if current_user and @user.id == current_user.id %>
+      <div class="my-3">
+        <%= link_to t(".edit_profile"), edit_profile_path, :class => "btn btn-outline-primary" %>
+      </div>
+    <% end %>
```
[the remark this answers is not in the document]
nertc: I also thought about a separate page or in-place editing for social links, but for consistency, preferred to keep it in the way that everything is done on the profile page. I agree with you about changing general behavior of editing profile page, but, I think, it will be better if we do it as a separate issue and PR and it will change not some part of `profile/edit` but whole behavior to in-place edit. For now, to avoid any confusion from users about the function of the button, I'll move it to where it was before (under the description block).

LivInTheLookingGlass:
> but it will make `social profile links` too colorful and too much attention grabbing.

Is it possible to use CSS monochrome for these? It might not work for all, but it could reduce asset duplication, which would help a bit with caching

app/models/social_link.rb
nertc: From now it will check whether regex of the websites directly follows to `http://www.`, `https://www.`, `http://` or `https://`.

app/models/social_link.rb
nertc: Was fixed.

app/views/social_links/_show.html.erb
nertc: From now, URL won't wrap and will show ellipses (`...`) if it doesn't fit.

app/views/profiles/edit.html.erb
nertc: It was changed to `col-sm-2`. Increasing button's width further will make it too wide.

nertc:
Thanks for the review

app/models/social_link.rb
nertc: It is named `names`, because it will either be `null` or array that will contain one element (user's name). Naming it `name` would cause to interpret it as a string or other single object variable, while naming it `names` emphasizes that it is an array that contains a name of the user.

app/assets/javascripts/user.js
```diff
@@ -10,6 +10,35 @@ $(document).ready(function () {
   var defaultHomeZoom = 12;
   var map, marker, deleted_lat, deleted_lon;
 
+  if ($("#social_links").length) {
+    $("#add-social-link").click(function (event) {
+      event.preventDefault();
+      const newIndex = -(new Date().getTime());
```
HolgerJeromin, line 16:
```suggestion
      const newIndex = -Date.now();
```

nertc: Thanks for the suggestion. PR was updated accordingly.

db/migrate/20250106113207_create_social_links.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSocialLinks < ActiveRecord::Migration[7.2]
+  def change
+    create_table :social_links do |t|
+      t.references :user, :null => false, :foreign_key => true
+      t.string :url
```
tomhughes, line 5: This should have `:null => false` as well I think - there's no good reason to have a social link with no actual link.

app/views/social_links/_show.html.erb
```diff
@@ -0,0 +1,8 @@
+<% social_links.each do |social_link| %>
+  <div class="text-body-secondary d-flex mb-3">
+    <%= image_tag "/assets/social_link_icons/#{social_link.parsed[:platform].nil? ? 'other' : social_link.parsed[:platform]}.svg",
```
tomhughes, line 3: I'm pretty sure this shouldn't have `/assets/` at the start of the link, and that doing so may interfere with using proper hashed asset paths.

app/models/social_link.rb
```diff
@@ -0,0 +1,67 @@
...
+#  fk_rails_...  (user_id => users.id)
+#
+
+class SocialLink < ApplicationRecord
+  belongs_to :user
+
+  validates :url, :format => { :with => %r{\Ahttps?://.+\z}, :message => :http_parse_error }
+
+  URL_PATTERNS = {
+    :discord => %r{\Ahttps?://(?:www\.)?discord\.com/users/(\d+)},
```
tomhughes, line 26: No bluesky?

app/models/social_link.rb
```diff
@@ -0,0 +1,67 @@
...
+    :facebook => %r{\Ahttps?://(?:www\.)?facebook\.com/([a-zA-Z0-9.]+)},
+    :github => %r{\Ahttps?://(?:www\.)?github\.com/([a-zA-Z0-9_-]+)},
+    :gitlab => %r{\Ahttps?://(?:www\.)?gitlab\.com/([a-zA-Z0-9_-]+)},
+    :instagram => %r{\Ahttps?://(?:www\.)?instagram\.com/([a-zA-Z0-9._]+)},
+    :linkedin => %r{\Ahttps?://(?:www\.)?linkedin\.com/in/([a-zA-Z0-9_-]+)},
+    :line => %r{\Ahttps?://(?:www\.)?line\.me/ti/p/([a-zA-Z0-9_-]+)},
+    :mastodon => %r{\Ahttps?://(?:www\.)?mastodon\.social/@([a-zA-Z0-9_]+)},
```
tomhughes, line 33: Once again mastodon is a problem because mastodon.social is just one option, but I'm not sure what the solution is... At the very least en.osm.town should probably be recognised though.

app/assets/javascripts/user.js
```diff
@@ -10,6 +10,35 @@ $(document).ready(function () {
...
+  if ($("#social_links").length) {
+    $("#add-social-link").click(function (event) {
+      event.preventDefault();
+      const newIndex = -Date.now();
+      const socialLinkForm = $(`
+        <div class="social-link-added-fields row mb-3">
+          <div class="col-sm-8">
+            <input class="form-control" type="text" name="user[social_links_attributes][${newIndex}][url]" id="user_social_links_attributes_${newIndex}_url">      
+          </div>
+          <button type="button" class="btn btn-outline-primary col-sm-2 align-self-start">${I18n.t("javascripts.profiles.edit.social_links.remove")}</button>
```
tomhughes, line 22: That I18n resource doesn't exist - the resource is `profiles.edit.social_links.remove` and is not currently exported to javascript.

tomhughes (changes requested):

app/assets/javascripts/user.js
nertc: https://github.com/openstreetmap/openstreetmap-website/blob/698c29697dcf468035de2d270c8f0359e30f4609/config/locales/en.yml#L3304-L3307

There is a key for `javascripts.profiles.edit.social_links.remove`. Adding `profiles.edit.social_links.remove` directly into the JS causes it to be missing for the JS code:
Screenshot 2025-02-17 163153

app/models/social_link.rb
nertc: `en.osm.town` was added. It will have the same icon as `mastodon.social`.

app/models/social_link.rb
```diff
@@ -0,0 +1,68 @@
...
+    :linkedin => %r{\Ahttps?://(?:www\.)?linkedin\.com/in/([a-zA-Z0-9_-]+)},
+    :line => %r{\Ahttps?://(?:www\.)?line\.me/ti/p/([a-zA-Z0-9_-]+)},
+    :mastodon => %r{\Ahttps?://(?:www\.)?(?:mastodon\.social|en\.osm\.town)/@([a-zA-Z0-9_]+)},
+    :medium => %r{\Ahttps?://(?:www\.)?medium\.com/@([a-zA-Z0-9_]+)},
+    :quora => %r{\Ahttps?://(?:www\.)?quora\.com/profile/([a-zA-Z0-9_-]+)},
+    :reddit => %r{\Ahttps?://(?:www\.)?reddit\.com/user/([a-zA-Z0-9_-]+)},
+    :skype => %r{\Ahttps?://(?:www\.)?join\.skype\.com/invite/([a-zA-Z0-9_-]+)},
+    :slack => %r{\Ahttps?://(?:www\.)?join\.slack\.com/shareDM/([a-zA-Z0-9_~-]+)},
+    :snapchat => %r{\Ahttps?://(?:www\.)?snapchat\.com/add/([a-zA-Z0-9_-]+)},
+    :stackoverflow => %r{\Ahttps?://(?:www\.)?stackoverflow\.com/users/(\d+/[a-zA-Z0-9_-]+)},
```
AntonKhorev, line 40: This captures the numeric id too, and it's going to be shown along with the username:

image

app/models/social_link.rb
nertc: Social networks were added according to icons that were provided from Bootstrap Icons. Now I added Bluesky, but Bootstrap doesn't have any icons for it, therefore, Font Awesome icon was used.

app/views/social_links/_show.html.erb
nertc: Was fixed accordingly, `/assets/` was removed.

db/migrate/20250106113207_create_social_links.rb
nertc: I agree with you, was fixed accordingly.

app/models/social_link.rb
```diff
@@ -0,0 +1,68 @@
...
+    :github => %r{\Ahttps?://(?:www\.)?github\.com/([a-zA-Z0-9_-]+)},
+    :gitlab => %r{\Ahttps?://(?:www\.)?gitlab\.com/([a-zA-Z0-9_-]+)},
+    :instagram => %r{\Ahttps?://(?:www\.)?instagram\.com/([a-zA-Z0-9._]+)},
+    :linkedin => %r{\Ahttps?://(?:www\.)?linkedin\.com/in/([a-zA-Z0-9_-]+)},
+    :line => %r{\Ahttps?://(?:www\.)?line\.me/ti/p/([a-zA-Z0-9_-]+)},
+    :mastodon => %r{\Ahttps?://(?:www\.)?(?:mastodon\.social|en\.osm\.town)/@([a-zA-Z0-9_]+)},
+    :medium => %r{\Ahttps?://(?:www\.)?medium\.com/@([a-zA-Z0-9_]+)},
+    :quora => %r{\Ahttps?://(?:www\.)?quora\.com/profile/([a-zA-Z0-9_-]+)},
+    :reddit => %r{\Ahttps?://(?:www\.)?reddit\.com/user/([a-zA-Z0-9_-]+)},
+    :skype => %r{\Ahttps?://(?:www\.)?join\.skype\.com/invite/([a-zA-Z0-9_-]+)},
```
AntonKhorev, line 37: `www.join.skype.com` doesn't exist. Maybe we don't need `www` everywhere.

app/models/social_link.rb
nertc: Should we omit numeric id? As Stack Overflow shows it for every profile link, I thought, it may have some meaning for users.

app/models/social_link.rb
AntonKhorev: Now it's added but it's at the end of the list for some reason.

app/views/social_links/_show.html.erb
```diff
@@ -0,0 +1,8 @@
+<% social_links.each do |social_link| %>
+  <div class="text-body-secondary d-flex mb-3">
+    <%= image_tag "social_link_icons/#{social_link.parsed[:platform].nil? ? 'other' : social_link.parsed[:platform]}.svg",
+                  :alt => social_link.parsed[:platform].nil? ? "other" : social_link.parsed[:platform],
+                  :class => "me-2" %>
+    <%= link_to social_link.parsed[:name], social_link.url, :class => "text-body-secondary d-block text-truncate", :rel => "nofollow me" %>
```
AntonKhorev, line 6: Do you need `d-block` on flex items?

app/views/social_links/_show.html.erb
```diff
@@ -0,0 +1,8 @@
+<% social_links.each do |social_link| %>
+  <div class="text-body-secondary d-flex mb-3">
+    <%= image_tag "social_link_icons/#{social_link.parsed[:platform].nil? ? 'other' : social_link.parsed[:platform]}.svg",
```
AntonKhorev, line 3: If all images are of the same size, you can add `:size => "16x21"` to avoid links jumping horizontally when the page loads.

app/views/social_links/_show.html.erb
```diff
@@ -0,0 +1,8 @@
...
+                  :alt => social_link.parsed[:platform].nil? ? "other" : social_link.parsed[:platform],
+                  :class => "me-2" %>
+    <%= link_to social_link.parsed[:name], social_link.url, :class => "text-body-secondary d-block text-truncate", :rel => "nofollow me" %>
+  </div>
```
AntonKhorev, line 7: What do you think about this?

```suggestion
  <%= link_to social_link.url, :class => "icon-link mw-100 text-body-secondary mb-3", :rel => "nofollow me" do %>
    <%= image_tag "social_link_icons/#{social_link.parsed[:platform].nil? ? 'other' : social_link.parsed[:platform]}.svg",
                  :alt => social_link.parsed[:platform].nil? ? "other" : social_link.parsed[:platform],
                  :size => "16" %>
    <span class="text-truncate">
      <%= social_link.parsed[:name] %>
    </span>
  <% end %>
```

AntonKhorev: They are in a list on github for example
image

app/assets/javascripts/user.js
tomhughes: You can control which strings are exported to javascript with `config/i18n-js.yml` but I'm confused now because I was getting missing translations with the current code. Maybe I forgot to re-export the translations.

app/models/social_link.rb
```diff
@@ -0,0 +1,68 @@
...
+    URL_PATTERNS.each do |platform, pattern|
+      names = url.match(pattern)
+      if names
+        return {
+          :platform => platform.to_s,
+          :name => NO_USERNAME_PLATFORMS.include?(platform.to_s) ? platform.to_s.capitalize : names[1]
+        }
+      end
+    end
+    { :platform => nil, :name => url }
```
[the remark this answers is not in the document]
nertc: `http(s)://` is now removed in the list. I was also thinking about removing `www.`. What do you think about it?

app/views/social_links/_show.html.erb
nertc: Thanks for the suggestion. It was applied and `<div>`s were changed with `<ul><li>` list structure.

app/views/social_links/_show.html.erb
nertc: `:size => "16"` was added to the `` tag to avoid text jumps. Thanks for the suggestion.

app/views/social_links/_show.html.erb
```diff
@@ -0,0 +1,8 @@
+<% social_links.each do |social_link| %>
+  <div class="text-body-secondary d-flex mb-3">
+    <%= image_tag "social_link_icons/#{social_link.parsed[:platform].nil? ? 'other' : social_link.parsed[:platform]}.svg",
+                  :alt => social_link.parsed[:platform].nil? ? "other" : social_link.parsed[:platform],
```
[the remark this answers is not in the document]
nertc: Translation was added.

app/models/social_link.rb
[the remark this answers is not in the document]
HolgerJeromin: I always hate if browsers liying about the real url, but here this is probably a good idea.

app/models/social_link.rb
nertc: `www.` was removed from Skype, Slack and WhatsApp invitation links (all three of them were broken if they were used with `www.`). As I checked, `en.osm.town` was also broken with `www.` and, therefore, Mastodon regex was changed to handle that situation accordingly.

app/models/social_link.rb
nertc: Regex was changed to hide Id number of the user from the list.

app/assets/javascripts/user.js
nertc: Key was changed to `profiles.edit.social_links.remove` and `config/i18n-js.yml` was changed accordingly. Approach you suggested is more consistent than what this PR had before. Thanks for the suggestion.

app/models/social_link.rb
nertc: Bluesky position was changed to be on the top of the list.

tomhughes:
That seems to have been some glitch with the tests. I forced a rerun and they passed fine.

nertc:
> That seems to have been some glitch with the tests. I forced a rerun and they passed fine.

I rebased `master` to https://github.com/openstreetmap/openstreetmap-website/pull/5009 and it has the same exact errors.

nertc:
PR was updated. `master` was rebased. Conflicts were resolved.

app/models/social_link.rb
```diff
@@ -0,0 +1,68 @@
...
+#
+#  fk_rails_...  (user_id => users.id)
+#
+
+class SocialLink < ApplicationRecord
+  belongs_to :user
+
+  validates :url, :format => { :with => %r{\Ahttps?://.+\z}, :message => :http_parse_error }
+
+  URL_PATTERNS = {
```
1ec5, line 25: Some suggestions for additional sites that users might want to display on their profiles in the same manner, in case the list isn’t strictly limited to social media accounts:

* OSM Community Forum (though it should be automatic)
* OpenGeofiction and OpenHistoricalMap
* Wikipedia, Wikidata, Wikivoyage, and Wikimedia Commons

app/models/social_link.rb
```diff
@@ -0,0 +1,68 @@
...
+    :bluesky => %r{\Ahttps?://(?:www\.)?bsky\.app/profile/([a-zA-Z0-9\._-]+)},
+    :discord => %r{\Ahttps?://(?:www\.)?discord\.com/users/(\d+)},
+    :facebook => %r{\Ahttps?://(?:www\.)?facebook\.com/([a-zA-Z0-9.]+)},
+    :github => %r{\Ahttps?://(?:www\.)?github\.com/([a-zA-Z0-9_-]+)},
+    :gitlab => %r{\Ahttps?://(?:www\.)?gitlab\.com/([a-zA-Z0-9_-]+)},
+    :instagram => %r{\Ahttps?://(?:www\.)?instagram\.com/([a-zA-Z0-9._]+)},
+    :linkedin => %r{\Ahttps?://(?:www\.)?linkedin\.com/in/([a-zA-Z0-9_-]+)},
+    :line => %r{\Ahttps?://(?:www\.)?line\.me/ti/p/([a-zA-Z0-9_-]+)},
+    :mastodon => %r{\Ahttps?://(?:(?:www\.)?mastodon\.social|en\.osm\.town)/@([a-zA-Z0-9_]+)},
```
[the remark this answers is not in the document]
tomhughes: I mean we could just assume that `@user@domain` and `https://domain/@user` were mastodon - does anything else must use a leading `@` like that on the user?

We should certainly include the domain in what we show though.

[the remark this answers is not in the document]
1ec5: Bluesky also uses a leading at sign, but only Mastodon uses two at signs as far as I know.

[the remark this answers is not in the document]
tomhughes: True, and we only recognise Bluesky via a URL anyway which doesn't include an @ at all.

tomhughes (changes requested):
I think this is almost there so I'd like to get it merged.

Can you rebase it please and also give some consideration to the suggestions @1ec5 made about additional sites and also about a more general recognition of mastodon handles?

app/models/social_link.rb
1ec5: By the way, I think these can be tail work. It could be a good opportunity for someone less involved with the codebase to contribute something small.

app/models/social_link.rb
```diff
@@ -0,0 +1,68 @@
...
+class SocialLink < ApplicationRecord
+  belongs_to :user
+
+  validates :url, :format => { :with => %r{\Ahttps?://.+\z}, :message => :http_parse_error }
+
+  URL_PATTERNS = {
+    :bluesky => %r{\Ahttps?://(?:www\.)?bsky\.app/profile/([a-zA-Z0-9\._-]+)},
+    :discord => %r{\Ahttps?://(?:www\.)?discord\.com/users/(\d+)},
+    :facebook => %r{\Ahttps?://(?:www\.)?facebook\.com/([a-zA-Z0-9.]+)},
+    :github => %r{\Ahttps?://(?:www\.)?github\.com/([a-zA-Z0-9_-]+)},
```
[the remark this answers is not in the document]
1ec5: This isn’t critical. Maybe I’ll open a separate issue for it later on, depending how this overall feature is received.

app/models/social_link.rb
tomhughes: Agreed - the one I'm most interested in addressing before we merge is widening mastodon support. The rest can be left for followup work I think.

nertc:
@tomhughes sorry for the delay on this PR. Yes, I am currently focused on this one and will push as soon as possible.

nertc:
PR was updated. `master` was merged. All conflicts were resolved.
PR was changed according to the suggestions.
Only RuboCop error is about the length of the `User` class (this PR added only 2 necessary lines to it). I can change `Metrics/ClassLength` to 623 from 621, but I think it either should be changed to some meaningful value in the separate PR (it shouldn't be changed every time `User` class is modified) or removed completely.
Feel free to suggest further changes related both to the code and RuboCop.

app/models/social_link.rb
nertc: Mentioned website URLs were added to the parser.
Neither `Bootstrap Icons` nor `FontAwesome Icons` had icons related to the mentioned websites (other than Wikipedia). For `OpenGeofiction`, `Wikipedia`, `Wikidata` and `Wikivoyage`, I modified their logos SVG files to have one color (for consistent design and visual simplicity).
For `OSM Wiki` and `OpenHistoricalMap`, there were no simple SVG logos to modify them to have one color, so I added filter to match similar color palette.
For `OSM Community Forum`, there was no SVG logo containing "message" icons that is used on the website. I used OSM SVG logo and added filter to it in the same way as to `OSM Wiki`.

Screenshot 1 (`OpenHistoricalMap`, `OSM Community Forum`, `OSM Wiki`):
image

Screenshot 2:
image

tomhughes:
You'll need to bump that value for now - we can't take this with failing lint. See https://github.com/openstreetmap/openstreetmap-website/pull/5988#issuecomment-2859279812 and following comments for recent discussion around this.

app/models/social_link.rb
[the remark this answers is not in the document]
nertc: Parser was modified. `mastodon.social` and `en.osm.town` URLs will be parsed to be shown as `@nixCraft@mastodon.social` and `@compass@en.osm.town`.
Also, parser now supports URLs in the format `@something@somewebsite.com`. They are automatically marked as `mastodon` URLs and formatted to work as a link to `https://somewebsite.com/@something`.

Screnshots:
image
image

nertc:
Last commits from `master` were merged and `Metrics/ClassLength` was changed to 623.

app/models/social_link.rb
```diff
@@ -0,0 +1,90 @@
...
+    :bluesky => %r{\Ahttps?://(?:www\.)?bsky\.app/profile/([a-zA-Z0-9\._-]+)},
+    :discord => %r{\Ahttps?://(?:www\.)?discord\.com/users/(\d+)},
+    :facebook => %r{\Ahttps?://(?:www\.)?facebook\.com/([a-zA-Z0-9.]+)},
+    :flickr => %r{\Ahttps?://(?:www\.)?flickr\.com/people/([a-zA-Z0-9@._-]+)},
+    :github => %r{\Ahttps?://(?:www\.)?github\.com/([a-zA-Z0-9_-]+)},
+    :gitlab => %r{\Ahttps?://(?:www\.)?gitlab\.com/([a-zA-Z0-9_-]+)},
+    :instagram => %r{\Ahttps?://(?:www\.)?instagram\.com/([a-zA-Z0-9._]+)},
+    :linkedin => %r{\Ahttps?://(?:www\.)?linkedin\.com/in/([a-zA-Z0-9_-]+)},
+    :line => %r{\Ahttps?://(?:www\.)?line\.me/ti/p/([a-zA-Z0-9_-]+)},
+    :mastodon => %r{\Ahttps?://(?:(?:www\.)?(mastodon\.social|en\.osm\.town))/([@a-zA-Z0-9_]+)},
```
tomhughes, line 35: The `@` is only valid, and is required, as the first character of the handle so should be before the character class I think?

app/models/social_link.rb
1ec5: Thank you, this looks pretty good. Yeah, OHM definitely needs a more adaptable logo for situations like this, but that’s outside of the PR’s scope. 😉

app/models/social_link.rb
1ec5: Yes, a leading `@` is required. Technically the `@` can still appear inside the handle in order to load a profile remotely from a different server, for example, https://mapstodon.space/@tomh@en.osm.town. But I think it’s perfectly fine if we don’t handle that case, since the user should provide a direct URL anyways.

nertc:
PR was updated. `master` was merged and conflicts were resolved. Code was changed according to the reviews.

app/models/social_link.rb
nertc: `@` sign was moved out of the `square brackets`

tomhughes (approved):
I pushed a small change to remove `config/i18-js.yml` which had crept back in here during a rebase and to cleanup trailing whitespace in a few files and with that I think this is good to merge - thanks for working on it.

AntonKhorev:
This got too clever with *label of hidden checkbox styled as a button* with consequences for keyboard navigation.

app/models/social_link.rb
hlfan: > For `OSM Community Forum`, there was no SVG logo containing "message" icons

Fixed in https://osm.wiki/File:OSM_Community_Forum_Logo.svg

Koreller:
@nertc and @AntonKhorev It is possible to add "How Did You Contribute" (HDYC) and "Humanitarian OpenStreetMap Team" (HOT) links ?
For example with my profil : https://www.hdyc.neis-one.org/?Koreller and https://tasks.hotosm.org/users/Koreller
* Logo HDYC : https://commons.wikimedia.org/wiki/File:HDYC_logo.svg and
* Logo HOT : https://wiki.openstreetmap.org/wiki/File:Hot_logo.svg
Thanks a lot for your help !

hlfan:
Possible? Definitely, see #6122 as an example.

Will that get merged? That's the point I'm more worried about.

PR #5440, opened 2024-12-24
Resourceful routes for received/issued blocks

AntonKhorev:
This creates `:issued_blocks` and `:received_blocks` resources for users. `:revoke_all` action is changed to deleting received blocks.


tomhughes:
Using the `edit` action as the revoke all view is a bit weird but I guess if we ever want to more general edit we can change the title of that page and the link to it and then just have revoke as one thing you can do from the edit page.

Other than that this looks good to me, thanks.

PR #5442, opened 2024-12-24
Wrap diary entry location coordinates

### LINKED ISSUE #1191: endless diary location longitude
When selecting a coordinate for a diary entry, one can scroll the map (when fully zoomed out) round and round all the time. So far this isn't to bad – but when one decides to finally place a marker the longitude can show a ridiculous value which isn't accepted by the diary interface.
The problem isn't really the endless scrolling which a lot of users won't do - but if zoomed fully out - which of the Americas will have the exact coordinate (for my use case)?
2016-03-24_211323_scr
2016-03-24_211601_scr

AntonKhorev:
Fixes #1191.

### DIFF
--- a/app/assets/javascripts/diary_entry.js
+++ b/app/assets/javascripts/diary_entry.js
@@ -2,8 +2,10 @@ $(document).ready(function () {
   var marker, map;
 
   function setLocation(e) {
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #5443, opened 2024-12-25
New note page on low zooms

### LINKED ISSUE #1291: low zoom URLs for adding a note handled badly
Starting from a URL like this: http://www.openstreetmap.org/note/new#map=2/44.7/25.5

The site is in a slightly buggy state where you're too zoomed out to add a note, but there's no message to explain that, and actually even when you zoom in, you don't get a marker to drag.

You can also get into this state by following a link to this URL: http://www.openstreetmap.org/note/new when your last viewed map position is zoomed out (and that's not an unreasonable URL to link to. Indeed our own site links to it here: https://www.openstreetmap.org/fixthemap )

AntonKhorev:
If
- you open `/note/new` while your zoom is too low as described in #1291 
- you zoom out far enough while on `/note/new`,

show a warning message and disable the submit button.

image

Fixes #1291 and another bug: note marker was disappearing together with notes layer when visible area became too large.

### DIFF
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -96,9 +96,7 @@ OSM.NewNote = function (map) {
     newNoteMarker.addTo(map);
     addHalo(newNoteMarker.getLatLng());
 
-    newNoteMarker.on("remove", function () {
-      addNoteButton.removeClass("active");
-    }).on("dragend", function () {
+    newNoteMarker.on("dragend", function () {
       content.find("textarea").focus();
     });
   }
@@ -158,7 +156,6 @@ OSM.NewNote = function (map) {
       createNote(location, text, (feature) => {
         content.find("textarea").val("");
         addCreatedNoteMarker(feature);
-        addNoteButton.removeClass("active");
         OSM.router.route("/note/" + feature.properties.id);
       });
     });
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -107,6 +107,14 @@ OSM.NewNote = function (map) {
     newNoteMarker = null;
   }
 
+  function updateControls() {
+    const zoomedOut = addNoteButton.hasClass("disabled");
+    const withoutText = content.find("textarea").val() === "";
+
+    content.find("#new-note-zoom-warning").prop("hidden", !zoomedOut);
+    content.find("input[type=submit]").prop("disabled", zoomedOut || withoutText);
+  }
+
   page.pushstate = page.popstate = function (path) {
     OSM.loadSidebarContent(path, function () {
       page.load(path);
@@ -114,9 +122,6 @@ OSM.NewNote = function (map) {
   };
 
   page.load = function (path) {
-    if (addNoteButton.hasClass("disabled")) return;
-    if (addNoteButton.hasClass("active")) return;
-
     addNoteButton.addClass("active");
 
     map.addLayer(noteLayer);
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/index/new_note.js
```diff
@@ -35,11 +35,13 @@ OSM.NewNote = function (map) {
     OSM.router.route("/note/new");
   });
 
-  function createNote(marker, form, url) {
-    var location = marker.getLatLng().wrap();
+  function createNote(form, url) {
+    if (!newNoteMarker) return;
```
tomhughes, line 39: Why do we now need to check this? As far as I can see nothing has changed and if can be null now then it always could be?

app/assets/javascripts/index/new_note.js
```diff
@@ -77,23 +77,20 @@ OSM.NewNote = function (map) {
...
-      });
+  function addHalo(latlng) {
+    if (halo) map.removeLayer(halo);
+    halo = L.circleMarker(latlng, {
+      weight: 2.5,
+      radius: 20,
+      fillOpacity: 0.5,
+      color: "#FF6200"
+    });
+    map.addLayer(halo);
```
tomhughes, line 85: The vertical whitespace was there to make this more readable...

app/assets/javascripts/index/new_note.js
```diff
@@ -130,11 +127,14 @@ OSM.NewNote = function (map) {
     });
 
     newNoteMarker.on("dragstart dragend", function (a) {
-      newHalo(newNoteMarker.getLatLng(), a.type);
+      removeHalo();
+      if (a.type !== "dragstart") {
```
tomhughes: Why not check for the type being `dragend` which would be much easier to reason about than having to find the set of captured events and then remove `dragstart` from that to work out when this is going to trigger?

test/system/create_note_test.rb
```diff
@@ -24,4 +24,25 @@ class CreateNoteTest < ApplicationSystemTestCase
       end
     end
   end
+
+  test "can open new note page when zoomed out" do
```
tomhughes, line 28: Wouldn't it make more sense to start zoomed in and then test that things disable when you zoom out? It's more realistic than starting from a zoomed out URL which is something the site would never offer as the control is still disabled at low zoom.

I mean there's nothing wrong with testing this, but we should test that zooming out triggers disablement as well.

app/assets/javascripts/index/new_note.js
AntonKhorev: Made some changes to be sure that:
- `addNewNoteMarker` is called from `page.load`
- `removeNewNoteMarker` is called only from `page.unload`
- `createNote` doesn't access the marker

app/assets/javascripts/index/new_note.js
AntonKhorev: restored whitespace

app/assets/javascripts/index/new_note.js
AntonKhorev: halo is supposed to reappear on drag end, so I guess it makes more sense to check for `dragend`. Updated.

test/system/create_note_test.rb
AntonKhorev: Starting zoomed out was a scenario from #1291 that I was going to fix. You can get there by opening `/note/new` from `/fixthemap`.

Added starting zoomed in, then zooming out, then zooming in again.

tomhughes (approved):
Looks good to me now, thanks.

PR #5445, opened 2024-12-27
Accessibility issue fixed #5431

### LINKED ISSUE #5431: Accessibility issue: links "click here"
Copy from [Translatewiki](https://translatewiki.net/w/i.php?title=Translating_talk:OpenStreetMap&oldid=12791558#Accessibility_issue:_Osm:Confirmations.confirm.click_here/en,_Osm:Confirmations.confirm.resend_html/en):

The current message text is:

* **Confirmations.confirm.resend html**: <code>If you need us to resend the confirmation email, %{reconfirm_link}.</code>
* **Confirmations.confirm.click here**: <code>click here</code>

For accessibility, “click here” should never be a link text, because the link text should always describe the link. A more accessible version of these messages would be:

* **Confirmations.confirm.resend html**: <code>If needed, we can %{reconfirm_link} to you.</code>
* **Confirmations.confirm.click here**: <code>resend the confirmation email</code>

[Rationale from the W3C](https://www.w3.org/TR/WCAG20-TECHS/F84.html): _Many blind people who use screen readers call up a dialog box that has a list of links from the page. They use this list of links to decide where they will go. But if many of the links in that list simply say "click here" or "more" they will be unable to use this feature in their screen reader, which is a core navigation strategy. That's why it's a failure of 2.4.9 to not provide any way of allowing them to know the destination from the link text alone. It is also true for people who tab through links. If all they hear as they tab through the document is "click here, click here, click here etc." they will become confused._

peach280:
I have tried to fix the accessibility issue.
Instead of click here, resend the confirmation email message has been used

### DIFF
--- a/app/views/confirmations/confirm.html.erb
+++ b/app/views/confirmations/confirm.html.erb
@@ -29,6 +29,6 @@
 
   <p class='text-body-secondary'>
     <%= t ".resend_html",
-          :reconfirm_link => link_to(t(".click_here"), url_for(:action => "confirm_resend")) %>
+          :reconfirm_link => link_to(t("resend the confirmation email"), url_for(:action => "confirm_resend")) %>
   </p>
 <% end %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,3 +1,4 @@
+confirmations
 en:
   html:
     dir: ltr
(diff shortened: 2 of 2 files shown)


maro-21:
fixes #5431

maro-21:
But you changed only one part of the sentece. Please see the issue again for the proposed change.

peach280:
I think i have changed the other part now. Please review.

hlfan:
The page you changed the translations of has been redesigned in #5447, which has already been merged. What would you like a review for?

AntonKhorev:
Superseded by #5447.

PR #5446, opened 2024-12-29
Move code that runs after user save from save_new_user to create method

AntonKhorev:
Currently `save_new_user` does other things like redirects after the actual save. That obscures which pages are reachable from actions that call `save_new_user`, see https://github.com/openstreetmap/openstreetmap-website/issues/5431#issuecomment-2564561794.

Also `save_new_user` accesses referer in two different ways: as a function argument and as `params[:referer]`.

---

Added test for clicking *Sign Up*, filling out email-login-password, clicking *Sign Up*, opening confirmation link from email.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -77,7 +77,24 @@ class UsersController < ApplicationController
         render :action => "new"
       else
         # Save the user record
-        save_new_user params[:email_hmac], params[:referer]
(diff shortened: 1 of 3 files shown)


tomhughes:
I mean it's not really clear why `save_new_user` exists when it's only called once but I guess this fine.

AntonKhorev:
`save_new_user` was called from two different places when if was first introduced. The other was I think save after terms page.

PR #5447, opened 2024-12-29
Replace resent confirmation link with button

### LINKED ISSUE #5431: Accessibility issue: links "click here"

AntonKhorev:
Includes #5446

There's an issue that the *resend confirmation email* link reads "click here" #5431. I think there shouldn't be any link at all, this should be a button because it does something, not just loads a page. And the page shouldn't be accessible by GET, the button should trigger POST. When you accept the confirmation link, there's also a button which posts, but its hidden and autoclicked by javascript. The current resend link+GET existed since https://github.com/openstreetmap/openstreetmap-website/commit/bddf524c888c7364119a673e9b6c0518b0c6f93d.

Fixes #5431 by not having a link, and not writing "click here" on a button.

image


tomhughes:
This depends on #5446 being merged first.

tomhughes:
Looks good to me, thanks.

PR #5452, opened 2025-01-01
API active blocks list

AntonKhorev:
See https://community.openstreetmap.org/t/is-this-the-same-person/123461/30

Problem: how an app accessing OSM API can determine that the current user is blocked, and tell them that they're blocked? Here's a method I came up with: https://github.com/streetcomplete/StreetComplete/issues/6062#issuecomment-2565450817. It has two hacky steps:
1. accessing `/api/0.6/user/#id` without auth header, because `/api/0.6/user/details` is inaccessible while blocked
2. generating a login link with referrer pointing to the block list

Some people access an endpoint that is known to generate 403 errors only when blocked instead of **1**, but there's no guarantee that things will stay that way. `/api/0.6/user/#id` may also become inaccessible without authorization in the future. This PR does **1** in an unambiguous way by providing an endpoint that lists active blocks for the current user. It it accessible even if the user is blocked.

`GET user/blocks/active` outputs a list of `<user_block>` elements like #4240. Currently I'm skipping block reason texts because I'm making this similar to *Messages API* which also skips message texts in lists. While this list is short and adding reasons wouldn't make the results very long, we may add a similar list of all blocks / by all users and it will get long.

I probably should add a limit but I can't imagine anyone collecting a lot of *active* blocks. More than one is already unusual.

`user/blocks/active` requires `read_prefs` scope, which may look strange, but that's because this scope is also responsible for user identity. Similarly `user/details` requires `read_prefs` despite not reading preferences.

### DIFF
--- a/app/controllers/api/user_blocks/active_lists_controller.rb
+++ b/app/controllers/api/user_blocks/active_lists_controller.rb
@@ -1,0 +1,13 @@
+module Api
+  module UserBlocks
+    class ActiveListsController < ApiController
+      before_action :authorize
+
+      authorize_resource :class => :active_user_blocks_list
+
+      before_action :set_request_formats
+
+      def show; end
+    end
+  end
+end
--- a/app/controllers/api/user_blocks/active_lists_controller.rb
+++ b/app/controllers/api/user_blocks/active_lists_controller.rb
@@ -1,7 +1,7 @@
 module Api
   module UserBlocks
     class ActiveListsController < ApiController
-      before_action :authorize
+      before_action -> { authorize(:skip_blocks => true) }
 
       authorize_resource :class => :active_user_blocks_list
 
--- a/app/controllers/api/user_blocks/active_lists_controller.rb
+++ b/app/controllers/api/user_blocks/active_lists_controller.rb
@@ -7,7 +7,10 @@ module Api
 
       before_action :set_request_formats
 
-      def show; end
+      def show
+        @user_blocks = current_user.blocks.active.order(:id => :desc)
+        @skip_reason = true
+      end
     end
   end
 end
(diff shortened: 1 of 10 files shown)


tomhughes:
I don't understand why we don't just make the user details call work even when you're blocked - in fact all read calls I would have thought should work.

That's not to say we might not want this as well as the details call only gives a count of blocks and not the details though I guess we could add that as an alternative to this.

AntonKhorev:
If I remember correctly, at some point DWG discussed making user/details accessible for blocked users and concluded that sometimes we want it blocked. It's similar to changeset comments: in most of the cases we want blocked users being able to post comments but sometimes we don't.

We'd need different types of blocks. I'm thinking about blocks that would have scopes like auth tokens. Abilities would be checked against (scopes granted by the token) - (scopes removed by active blocks).

app/controllers/api/user_blocks/active_lists_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Api
+  module UserBlocks
+    class ActiveListsController < ApiController
+      before_action :disable_blocks_check
```
tomhughes, line 4: Why use flash to smuggle the skip blocks flag to `setup_user_auth`? Could we not just pass an argument to the `authorize` callback and have that pass it on to `setup_user_auth`?

AntonKhorev: I made it similar to [disable_terms_redirect](https://github.com/openstreetmap/openstreetmap-website/blob/b44399fb7c1747a811ea613e7782f77849a6d2e1/app/controllers/api/users_controller.rb#L52-L57)

AntonKhorev: Updated to match #5658

tomhughes (approved):

PR #5453, opened 2025-01-01
Use turbo for the settings page to stop the URL changing

tomhughes:
This changes the user settings page to use turbo to fix the problem identified in https://github.com/openstreetmap/openstreetmap-website/pull/5449#issuecomment-2565252266 where the URL changes when submitting a settings page with errors.

It also fixes an issue where errors weren't moved from the `email` to the `new_email` attribute properly, and simplifies checking if errors are present.


gravitystorm:
I think I would prefer something similar to #5449 (reasoning at https://github.com/openstreetmap/openstreetmap-website/pull/5449#issuecomment-2578415862 ).

tomhughes:
Superseded by #5689 and #5696.

PR #5454, opened 2025-01-02
Remove "expired unread" short block status

AntonKhorev:
### Description
Changes *Status* column of `/user_blocks` to no longer say "expired unread".
It was supposed to mean that the block is past its end date but still wasn't read by the user. That turned out to be confusing because although it said "expired", such blocks were active.

### How has this been tested?
Added tests for `block_short_status` helper.


app/helpers/user_blocks_helper.rb
```diff
@@ -27,14 +27,10 @@ def block_status(block)
...
   def block_short_status(block)
     if block.active?
-      if block.needs_view?
-        if block.ends_at > Time.now.utc
-          t("user_blocks.helper.short.active_unread")
-        else
-          t("user_blocks.helper.short.expired_unread")
-        end
-      else
+      if block.ends_at > Time.now.utc || block.deactivates_at
```
tomhughes, line 30: Is the `deactivates_at` test needed?

I don't see how that will ever be set for a block which is active but which has passed the `ends_at` test, given that a block is active if it hasn't passed the end time or needs a view, and viewing a block that needs a view clears the needs view flag.

AntonKhorev: I was thinking what to output in case we're between `ends_at` and `deactivates_at` because it's allowed by validations. But since it's currently impossible to get such `deactivates_at`, let's remove it.

tomhughes (approved):
Looks good now, thanks.

PR #5455, opened 2025-01-02
Remove commit parameter from diary controller test

AntonKhorev:
`commit` is the name of the submit button, but it's never checked by the controller and its value is not "save" in the actual forms.


tomhughes:
I suspect that's some ancient history... Look good to me anyway, thanks.

PR #5456, opened 2025-01-02
Remove test_logout_removes_session_token

AntonKhorev:
This test was about tokens that don't exist since #4535. That PR modified the test by removing the assert that checked the number of tokens in https://github.com/openstreetmap/openstreetmap-website/commit/4dff06a6293971c3e17f8508859a1d80717a23f6. Yet the test was kept although it doesn't do what it was supposed to do.

It also has `post user_save_path` that used to happen after `post user_new_path` when the terms page was shown during signups. This is not happening after #4455, and `post user_save_path` was removed from other tests in https://github.com/openstreetmap/openstreetmap-website/commit/1276fb944a8024884f3121d21577ef892353dfd1.


tomhughes:
Looks good to me, thanks.

PR #5457, opened 2025-01-02
Use resourceful route for new message

AntonKhorev:
This changes the path from
`/message/new/:display_name`
to
`/messages/new/:display_name`


tomhughes:
Looks good to me, thanks.

PR #5458, opened 2025-01-02
Use resourceful route for message reply

AntonKhorev:
Creates a nested `reply` resource for messages.

Changes reply path from `/messages/:id/reply` to `/messages/:id/reply/new` because it's a new reply.

Removes unused POST route. It was previously removed in https://github.com/openstreetmap/openstreetmap-website/commit/40cab845fbabb3f0aa8131c16e48cd3d58b6dcb3, then restored because it was still in use. I don't think it's in use now.

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -42,7 +42,7 @@ class Ability
         can :update, DiaryEntry, :user => user
         can [:create], DiaryComment
         can [:make_friend, :remove_friend], Friendship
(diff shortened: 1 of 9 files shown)


tomhughes (changes requested):
This mostly looks good but it turns out it fails when applied on top of #5457 as the reply route no longer works properly when combined with the change to new.

AntonKhorev:
Updated for compatibility with #5457.

tomhughes (approved):
Looks good now, thanks.

PR #5459, opened 2025-01-03
Use resourceful routes for user new/create actions

AntonKhorev:
- `user_new_path` helper renamed to `new_user_path`
- user create action moved from `POST /user/new` to `POST /user`, which corresponds to `users_path` helper


tomhughes:
Looks good to me, thanks.

PR #5460, opened 2025-01-03
Convert Auth.providers into array

AntonKhorev:
There's `Auth` module that lists currently enabled auth providers. It's used in two different places: auth providers template for login/signup pages and account edit page.

What the module actually does is it prepares options for `<select>` on the account edit page. That's why it includes localization strings as keys of a hash. The template on login/signup pages needs different strings and ignores the keys. Why do we need then to include the localization strings in `Auth` at all?


tomhughes (approved):
Looks good to me, thanks.

PR #5461, opened 2025-01-03
Rename Account module to Accounts

AntonKhorev:
Our other modules are named in plural, except for `Api`.

Removes `account` next to `accounts` in views and i18n strings.


tomhughes (approved):
Looks good to me, thanks.

PR #5462, opened 2025-01-04
Use resourceful routes for terms view/accept/decline

AntonKhorev:
- make user terms a nested resource inside the account resource
- `GET /user/terms` changed to `GET /account/terms`. It's inconvenient to have it at `/user/terms` because that depends on the declaration order. If user resources are declared first, they capture `terms` as a username. I still add a redirect from `/user/terms` to `/account/terms` which has to come before `resources :users`, but with redirects it's obvious that they are temporary.
- `POST /user/save` changed to `PUT /account/terms`. With `user#save` it's not clear that this is a target of the terms form.


test/integration/user_terms_seen_test.rb
```diff
@@ -25,12 +25,12 @@ def test_terms_presented_at_login
     assert_template "sessions/new"
     post "/login", :params => { :username => user.email, :password => "test", :referer => "/diary/new" }
     # but now we need to look at the terms
-    assert_redirected_to :controller => :users, :action => :terms, :referer => "/diary/new"
+    assert_redirected_to "/account/terms?#{{ :referer => '/diary/new' }.to_query}"
```
gravitystorm, line 28:
```ruby
assert_redirected_to account_terms_path(:referer => "/diary/new")
```
... is a bit easier to read

gravitystorm (changes requested):
Looks good to me, one small change in the tests.

I also appreciate the tests are slightly refactored, and this is an improvement.

test/integration/user_terms_seen_test.rb
AntonKhorev: The entire test didn't use path helpers and has path expressions like `"/api/#{Settings.api_version}/user/preferences"`. I wrote `"/account/terms?#{{ :referer => '/diary/new' }.to_query}"` in case that was on purpose.

Updated to `account_terms_path(:referer => "/diary/new")`.

gravitystorm:
Merged, thanks!

PR #5463, opened 2025-01-04
Remove old browse redirects

AntonKhorev:
Remove browse redirects introduced in https://github.com/openstreetmap/openstreetmap-website/commit/3cd5f45e08d977d04a778ab8802f71df85edc314. They are over 10 years old and probably not used by anyone. Depending on if they are actually not used according to the logs of course.


tomhughes:
I'm afraid to say there were still about 26000 hits on `/browse` yesterday :disappointed:

AntonKhorev:
Are those bots?

Are there hits on `/create-account.html` or `/forgot-password.html`?

AntonKhorev:
https://wiki.openstreetmap.org/wiki/OpenStreetMap_account

tomhughes:
That had links to `/forgot-password.html` which I have no fixed but I don't see any to `/create-account.html`?

AntonKhorev:
I'll close this since browse paths are still in use.

PR #5464, opened 2025-01-04
Add OldElementsController base class

AntonKhorev:
We already have [OldElementsController](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/api/old_elements_controller.rb) on the API side. This PR adds a similar class on the web side. Removes some copypasted code.


tomhughes:
Looks good to me, thanks.

PR #5465, opened 2025-01-04
Show outbox after sending new message

AntonKhorev:
Currently if you send a new message, you are redirected to your inbox. But you probably want to see the result of the message being sent, and that is in the outbox.

Redirect to inbox existed long ago https://github.com/openstreetmap/openstreetmap-website/commit/e6af088dda8e85f53b11ffbaddb11edf08e4a046, before there was outbox, which was added in https://github.com/openstreetmap/openstreetmap-website/commit/d736a158bee0ff17bdba30496b87e1bfe20e2910.


tomhughes:

Issue #5467, opened 2025-01-04
bug with selenium while running the tests


etiennejourdier:
Hi,
I am preparing my very first PR, but I can't run the tests requested in [CONTRIBUTING.md](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md)

`bundle exec teaspoon` gives me this error:
```
/home/<username>/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/response.rb:62:in 'add_cause': binary is not a Firefox executable (Selenium::WebDriver::Error::InvalidArgumentError)
```

and `bundle exec rails test:all` gives me 260 times the same error:
```
Error:
UserMutingTest#<test_name>
Selenium::WebDriver::Error::InvalidArgumentError: binary is not a Firefox executable
```

I noticed that selenium is used since the recent PR #5216 by @kcne, could you help me ?

I read that this bug can occur depending on the path to firefox or geckoriver. On my system:
- firefox is in `/usr/bin/firefox`
- geckodriver is in `/snap/bin/geckodriver`

By the way, I also noticed a bug with `bundle exec erblint` which returns `no files found or given, specify files or config...`. I had to modify the command to `bundle exec erblint *` but then it checks non erb files like GemFile and LICENSE. What is the correct command ?

Thanks !

tomhughes:
A snap based firefox is unlikely to work.

AntonKhorev:
> a bug with `bundle exec erblint`

Don't the docs tell you to run it as `bundle exec erblint .`? See `CONTRIBUTING.md`.

AntonKhorev:
When I run system tests, I replace 

```ruby
  driven_by :selenium, :using => :headless_firefox do |options|
    options.add_preference("intl.accept_languages", "en")
  end
```

with

```ruby
  driven_by :selenium, :using => :firefox do |options|
    options.add_preference("intl.accept_languages", "en")
    options.binary = "/path/to/firefox/binary/not/from/snap"
  end
```

in `test/application_system_test_case.rb` because I don't run all of the tests and I want to see what the browser does on the tests I run.

etiennejourdier:
> Don't the docs tell you to run it as `bundle exec erblint .`?

oops, sorry, my mistake, I didn't see the `.` ...

etiennejourdier:
> A snap based firefox is unlikely to work.

[...]

PR #5421, opened 2024-12-19
Bump allowed ECMAScript version to 2021 (ES12)

tomhughes:
We've been on 2015 for a long time as we no longer care about IE there's probably no good reason not to bump it.

I've had a look at the support matrices and up to 2022 just about everything seems to be supported in all current browsers. Detailed results can be found:

* [ECMAScript 2016](https://caniuse.com/?feats=array-includes,mdn-javascript_operators_exponentiation)
* [ECMAScript 2017](https://caniuse.com/?feats=async-functions,object-values,object-entries,mdn-javascript_builtins_object_getownpropertydescriptors,pad-start-end,mdn-javascript_grammar_trailing_commas_trailing_commas_in_functions)
* [ECMAScript 2018](https://caniuse.com/?feats=mdn-javascript_builtins_regexp_dotall,js-regexp-lookbehind,mdn-javascript_builtins_symbol_asynciterator,mdn-javascript_functions_method_definitions_async_generator_methods,mdn-javascript_grammar_template_literals_template_literal_revision,mdn-javascript_operators_destructuring_rest_in_objects,mdn-javascript_operators_destructuring_rest_in_arrays,promise-finally)
* [ECMAScript 2019](https://caniuse.com/?feats=array-flat,mdn-javascript_builtins_object_fromentries,mdn-javascript_builtins_string_trimstart,mdn-javascript_builtins_symbol_description,mdn-javascript_statements_try_catch_optional_catch_binding,mdn-javascript_builtins_json_stringify_well_formed_stringify,mdn-javascript_builtins_array_sort_stable_sorting,mdn-javascript_builtins_function_tostring_tostring_revision)
* [ECMAScript 2020](https://caniuse.com/?feats=mdn-javascript_operators_optional_chaining,mdn-javascript_operators_nullish_coalescing,mdn-javascript_builtins_globalthis,es6-module-dynamic-import,bigint,mdn-javascript_builtins_promise_allsettled,mdn-javascript_builtins_string_matchall,mdn-javascript_statements_export_namespace,mdn-javascript_operators_import_meta)
[...]

PR #5468, opened 2025-01-04
display an encouragement to contribute after many anonymous notes

etiennejourdier:
### Motivation
In the French community, we've had a problem for over a year with someone creating a huge number of anonymous notes, which hampers the processing of other relevant notes in the same area. See [this discussion](https://forum.openstreetmap.fr/t/notes-anonymes-a-gogo/18362) on the French community forum.

I thought it would be a good idea to count the number of anonymous notes created in the same browser and to display a warning when this number becomes too high, to encourage visitors to become users and contributors.

### Description

This PR creates a new warning in the UI of the new note form if an anonymous visitor has already created too many anonymous notes (threshold set at 20 for now)
image

**Changes**:
- in template views/note/new.html.erb
  - add a new div element displaying the warning (hidden at first)
- in locales/en.yml
  - add text and links for the warning, with ‘(N)’ being replaced by the number of anonymous notes
- in new_note.js
  - when an anonymous note is created: add +1 to the counter in localStorage
  - when the form is displayed: check the counter in localStorage. If more that 20 notes have already been created: display the warning, and replace (N) in the text warning with the number of anonymous notes already created.
 
**Possible additional change (not included yet)**
After an even larger number of anonymous notes (50?), I thought we could hide the textarea, and trigger its display by clicking on a button _sure you want to post an anonymous note again?_, which would force the visitor to perform an additional action. This is not included in this PR yet, but tell me if you want me to add it.

### How has this been tested?
The code has been tested on my personal computer using rails built-in webserver. I verified that the new note form is still working fine both when logged in and in anonymous mode.

However, as described in issue #5467 I was not able to make all the tests requested in [CONTRIBUTING.md](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md).

As this is my very first PR, I'm not sure I've done everything right, so I welcome any relevant comments!

### DIFF
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -162,6 +162,10 @@ OSM.NewNote = function (map) {
       newNoteMarker.dragging.disable();
 
       createNote(location, text, (feature) => {
+        if (typeof OSM.user === "undefined") {
+          var anonymousNotesCount = Number(Cookies.get("_osm_anonymous_notes_count")) || 0;
+          Cookies.set("_osm_anonymous_notes_count", anonymousNotesCount + 1, { secure: true, expires: 30, path: "/", samesite: "lax" });
+        }
         content.find("textarea").val("");
         addCreatedNoteMarker(feature);
         OSM.router.route("/note/" + feature.properties.id);
--- a/app/controllers/concerns/session_methods.rb
+++ b/app/controllers/concerns/session_methods.rb
@@ -39,6 +39,8 @@ module SessionMethods
     session[:fingerprint] = user.fingerprint
     session_expires_after 28.days if session[:remember_me]
 
+    cookies.delete :_osm_anonymous_notes_count
+
     target = referer || url_for(:controller => :site, :action => :index)
 
     # The user is logged in, so decide where to send them:
--- a/app/controllers/confirmations_controller.rb
+++ b/app/controllers/confirmations_controller.rb
@@ -30,6 +30,7 @@ class ConfirmationsController < ApplicationController
         user.email_valid = true
         flash[:notice] = gravatar_status_message(user) if gravatar_enable(user)
         user.save!
+        cookies.delete :_osm_anonymous_notes_count
         referer = safe_referer(params[:referer]) if params[:referer]
 
         pending_user = session.delete(:pending_user)
--- a/app/views/notes/new.html.erb
+++ b/app/views/notes/new.html.erb
@@ -5,9 +5,17 @@
 <div class="note">
   <p class="alert alert-info"><%= t(".intro") %></p>
   <% if !current_user %>
-    <p class="alert alert-warning"><%= t ".anonymous_warning_html",
-                                         :log_in => link_to(t(".anonymous_warning_log_in"), login_path(:referer => new_note_path)),
-                                         :sign_up => link_to(t(".anonymous_warning_sign_up"), new_user_path) %></p>
+    <div class="alert alert-warning pb-0">
+      <p><%= t ".anonymous_warning_html",
+               :log_in => link_to(t(".anonymous_warning_log_in"), login_path(:referer => new_note_path)),
+               :sign_up => link_to(t(".anonymous_warning_sign_up"), new_user_path) %></p>
+    <% if @anonymous_notes_count >= 10 %>
+      <p><%= t ".counter_warning_html",
+               :x_anonymous_notes => t(".x_anonymous_notes", :count => @anonymous_notes_count),
+               :contribute_by_yourself => link_to(t(".counter_warning_guide_link.text"), t(".counter_warning_guide_link.url")),
+               :community_can_help => link_to(t(".counter_warning_forum_link.text"), t(".counter_warning_forum_link.url")) %></p>
+    <% end %>
+    </div>
   <% end %>
   <p class="alert alert-warning" id="new-note-zoom-warning" hidden><%= t "javascripts.site.createnote_disabled_tooltip" %></p>
   <form action="#">
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -45,6 +45,7 @@ class NotesController < ApplicationController
   end
 
   def new
+    @anonymous_notes_count = request.cookies["_osm_anonymous_notes_count"].to_i || 0
     render :action => :new_readonly if api_status != "online"
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3087,9 +3087,19 @@ en:
     new:
       title: "New Note"
       intro: "Spotted a mistake or something missing? Let other mappers know so we can fix it. Move the marker to the correct position and type a note to explain the problem."
-      anonymous_warning_html: "You are not logged in. Please %{log_in} or %{sign_up} if you want to receive updates for your note."
+      anonymous_warning_html: "You are not logged in. Please %{log_in} or %{sign_up} to receive updates for your note and help mappers resolve it."
       anonymous_warning_log_in: "log in"
       anonymous_warning_sign_up: "sign up"
+      counter_warning_html: "You have already posted at least %{x_anonymous_notes}, that's great for the community, thank you! Now we encourage you to %{contribute_by_yourself}, it is not that complicated, and %{community_can_help}."
+      x_anonymous_notes:
+        one: "%{count} anonymous note"
+        other: "%{count} anonymous notes"
+      counter_warning_guide_link:
+        text: "contribute by yourself"
+        url: https://wiki.openstreetmap.org/wiki/Beginners%27_guide
+      counter_warning_forum_link:
+        text: "the community can help you"
+        url: https://community.openstreetmap.org/
       advice: "Your note is public and may be used to update the map, so don't enter personal information, or information from copyrighted maps or directory listings."
       add: Add Note
     new_readonly:


tomhughes:
If they're valid notes I don't really see the problem? Are these genuine notes or is this some sort of spam attack?

app/assets/javascripts/index/new_note.js
```diff
@@ -152,6 +152,16 @@ OSM.NewNote = function (map) {
       .on("input", updateControls)
       .focus();
 
+    var anonymousNotesCount = Number(localStorage.getItem("anonymousNotesCount"));
+
+    if (typeof OSM.user === "undefined" && anonymousNotesCount >= 20) {
+      var counterWarning = content.find("#new-note-counter-warning");
+      if (typeof counterWarning.html() !== "undefined") {
+        counterWarning.html(counterWarning.html().replace("(N)", anonymousNotesCount));
```
AntonKhorev, line 160: That's not how translatable numerals work. The number could also affect other parts of the sentence in various languages ("24 заметки", "25 заметок").

AntonKhorev:
I'd store the counter in a cookie, set an expiration date for that cookie to something like a week and replace the existing "You are not logged in" alert with this more motivating one with a note counter server-side.

app/assets/javascripts/index/new_note.js
```diff
@@ -162,6 +172,9 @@ OSM.NewNote = function (map) {
       newNoteMarker.dragging.disable();
 
       createNote(location, text, (feature) => {
+        if (typeof OSM.user === "undefined") {
+          localStorage.setItem("anonymousNotesCount", anonymousNotesCount + 1);
```
HolgerJeromin, line 176: If we keep `localstorage` (see @AntonKhorev comment):
`localStorage.setItem` can throw an exception.

app/assets/javascripts/index/new_note.js
```diff
@@ -152,6 +152,16 @@ OSM.NewNote = function (map) {
       .on("input", updateControls)
       .focus();
 
+    var anonymousNotesCount = Number(localStorage.getItem("anonymousNotesCount"));
```
HolgerJeromin, line 155:
```suggestion
    var anonymousNotesCount = Number(localStorage.getItem("anonymousNotesCount") ?? 0);
```
after #5421 is merged

gendy54:
> If they're valid notes I don't really see the problem? Are these genuine notes or is this some sort of spam attack?

This contributor's notes are potentially problematic and pollute the other notes.
image

cquest:
> If they're valid notes I don't really see the problem? Are these genuine notes or is this some sort of spam attack?

A large part of them are not valid or useless.

This PR should not be seen as a way to solve a one guy problem, but a global issue with anonymous notes. The lack of communication channel makes a lot of them difficult to deal with, so at least when someone is creating a lot of notes, we should push to create that communication channel, and it also a way to avoid spam.

app/assets/javascripts/index/new_note.js
etiennejourdier: ah yes you're right, I hadn't thought of that, could you point me to a place in the code and/or translatewiki so that I can see how to manage it?

etiennejourdier:
> I'd store the counter in a cookie, set an expiration date

I am not a specialist of cookies VS localStorage, the only advantage of cookies is the possibility to set an expiration date ?

> replace the existing "You are not logged in" alert with this more motivating one

good idea !

> with a note counter server-side.

I'm not sure I understood this part. I imagined this would require storing an ‘anonymous visitor ID’ in the cookie and creating a table on the server to associate this ID with the notes created by the visitor? but maybe I've misunderstood?

app/assets/javascripts/index/new_note.js
etiennejourdier: Is the nullish coalescence operator `??` you propose below sufficient to handle the case of an exception? Or does it only work for `null` and `undefined`?

HolgerJeromin: `??` replaces `null` and `undefined` values only.

etiennejourdier: OK I will add a `try ... catch`

app/assets/javascripts/index/new_note.js
etiennejourdier: OK I understood how pluralization works with i18n, but it is server-side ... If the value of the counter is in the browser (whether in a cookie or in localStorage), it has to be inserted by jquery, so after the translation, and it's too late to manage the different plurals, right ? Is there another option to manage plurals with jquery ? Or is it why you propose the counter to be server-side ?

AntonKhorev:
> I am not a specialist of cookies VS localStorage, the only advantage of cookies is the possibility to set an expiration date ?

The browser sends cookies to the server. It doesn't send localStorage. If the server knows the counter value, it can display another message based on that value, that's what I mean by:

> replace the existing "You are not logged in" alert with this more motivating one with a note counter server-side.

AntonKhorev:
> I imagined this would require storing an ‘anonymous visitor ID’ in the cookie

This thing already exists as session id in `_osm_session` cookie. And that's another option. Instead of managing the counter and its expiration date in javascript, you can make a request at some special endpoint after a note is successfully created here https://github.com/openstreetmap/openstreetmap-website/blob/97b14ce22a0c7d17bb0b303b3795297721bc31d4/app/assets/javascripts/index/new_note.js#L48
That request will increase the counter stored in `session` if necessary. After that `app/views/notes/new.html.erb` or its controller can check the counter value and display the message based on that.

app/assets/javascripts/index/new_note.js
AntonKhorev: There is i18n library in javascript too but it's better to avoid it if possible. You'd have to declare strings in a different section of `en.yml`, then all of those strings are sent over to the client.

Example of javascript i18n with count:
https://github.com/openstreetmap/openstreetmap-website/blob/97b14ce22a0c7d17bb0b303b3795297721bc31d4/app/assets/javascripts/leaflet.locate.js#L8

etiennejourdier:
Thanks for your time, comments, and advice. I just pushed a new version with :

- the counter stored in a cookie, with a duration of 30 days, as anonymous visitors don't necessarily post notes every day (I'm not sure about other cookie options, please review)
- the cookie read in the controller
- a message in the view template, included in the already existing warning
- with pluralization managment in en.yml

I have slightly modified the standard warning in this way, to insist that discussion will help to resolve the note:
image

And I added the encouragement to contribute yourself in the same box, after at least 20 notes, with a link to the beginners' guide, and to the forum.
image

I know that finding the perfect sentence is always complicated, and English is not my mother tongue, so tell me if you prefer another sentence.

AntonKhorev (changes requested):
You're adding things like `var anonymousNotesCount = Number(localStorage.getItem("anonymousNotesCount"));`
in your first commit, then you completely remove them in the second. In this case you can just squash these commits together.

etiennejourdier:
> you can just squash these commits together.

Good idea, I'll do that, it'll be cleaner. Sorry, I'm not really used to PR so I hadn't thought of it myself.

app/views/notes/new.html.erb
```diff
@@ -5,9 +5,17 @@
...
   <p class="alert alert-info"><%= t(".intro") %></p>
   <% if !current_user %>
-    <p class="alert alert-warning"><%= t ".anonymous_warning_html",
-                                         :log_in => link_to(t(".anonymous_warning_log_in"), login_path(:referer => new_note_path)),
-                                         :sign_up => link_to(t(".anonymous_warning_sign_up"), new_user_path) %></p>
+    <p class="alert alert-warning">
+      <%= t ".anonymous_warning_html",
+            :log_in => link_to(t(".anonymous_warning_log_in"), login_path(:referer => new_note_path)),
+            :sign_up => link_to(t(".anonymous_warning_sign_up"), new_user_path) %>
+    <% if @anonymous_notes_count >= 10 %>
```
AntonKhorev, line 12: You have a `>=` condition here, but the message reads `You have already posted more than %{x_anonymous_notes}`. If I create exactly 10 notes, it will tell me that I "posted more than 10".

app/views/notes/new.html.erb
```diff
@@ -5,9 +5,17 @@
...
+      <br><%= t ".counter_warning_html",
+                :x_anonymous_notes => t(".x_anonymous_notes", :count => @anonymous_notes_count),
+                :contribute_by_yourself => link_to(t(".counter_warning_guide_link.text"), t(".counter_warning_guide_link.url")),
+                :community_can_help => link_to(t(".counter_warning_forum_link.text"), t(".counter_warning_forum_link.url")) %>
+    <% end %>
+    </p>
```
AntonKhorev, line 18: How about something like this to add a gap between the messages?

```erb
    <div class="alert alert-warning pb-0">
      <p>...</p>
    <% if @anonymous_notes_count >= 10 %>
      <p>...</p>
    <% end %>
    </div>
```

image

AntonKhorev:
What shall we do with the counter
- if the user logs in?
- if the user creates an account?

etiennejourdier:
> What shall we do with the counter
> 
> * if the user logs in?
> * if the user creates an account?

I thought about it, we could delete the cookie after logging in, but it seemed pointless as the cookie will expire 30 days after the last anonymous note anyway. Do you agree?

AntonKhorev:
> but it seemed pointless as the cookie will expire 30 days after the last anonymous note anyway

The point could be that if I log in my session may expire faster than that and I'll se the message again. But maybe it's not worth doing anything about it.

app/views/notes/new.html.erb
etiennejourdier: good idea, I've modified it

app/controllers/concerns/session_methods.rb
```diff
@@ -39,6 +39,8 @@ def successful_login(user, referer = nil)
     session[:fingerprint] = user.fingerprint
     session_expires_after 28.days if session[:remember_me]
 
+    cookies.delete :_osm_anonymous_notes_count
```
etiennejourdier, line 42: > The point could be that if I log in my session may expire faster than that and I'll se the message again. But maybe it's not worth doing anything about it.

I added the deletion of the cookie after successful login

etiennejourdier:
Hi,
Thanks again for the positive and pertinent comments, I pushed a new version with:
- only 1 squashed commit
- updated threshold
- modified warning with a gap
- deletion of the cookie after successful login

app/views/notes/new.html.erb
```diff
@@ -5,9 +5,17 @@
...
-    <p class="alert alert-warning"><%= t ".anonymous_warning_html",
-                                         :log_in => link_to(t(".anonymous_warning_log_in"), login_path(:referer => new_note_path)),
-                                         :sign_up => link_to(t(".anonymous_warning_sign_up"), new_user_path) %></p>
+    <div class="alert alert-warning pb-0">
+      <p><%= t ".anonymous_warning_html",
+            :log_in => link_to(t(".anonymous_warning_log_in"), login_path(:referer => new_note_path)),
+        :sign_up => link_to(t(".anonymous_warning_sign_up"), new_user_path) %></p>
+    <% if @anonymous_notes_count > 10 %>
+      <p><%= t ".counter_warning_html",
+                :x_anonymous_notes => t(".x_anonymous_notes", :count => @anonymous_notes_count),
```
AntonKhorev, line 14: The message is still wrong, see the second paragraph of https://github.com/openstreetmap/openstreetmap-website/pull/5468#discussion_r1908778180.

```suggestion
                :x_anonymous_notes => t(".x_anonymous_notes", :count => @anonymous_notes_count - 1),
```

AntonKhorev:
> deletion of the cookie after successful login

If you do this, you probably also want to delete the cookie on signup. I'm not sure if it's better to do before or after confirming the email. If before, you'd delete the cookie somewhere here:

https://github.com/openstreetmap/openstreetmap-website/blob/a6c6f26acbc9d5f62c4eddfb73bbbd5c5e03621a/app/controllers/users_controller.rb#L80

app/views/notes/new.html.erb
etiennejourdier: I'm really sorry, I'd totally missed the second paragraph of your comment... Finally I changed the message to "at least x anonymous notes" because the counter may miss notes if the cookie has been deleted in the meantime.

app/controllers/confirmations_controller.rb
```diff
@@ -30,6 +30,7 @@ def confirm
         user.email_valid = true
         flash[:notice] = gravatar_status_message(user) if gravatar_enable(user)
         user.save!
+        cookies.delete :_osm_anonymous_notes_count
```
etiennejourdier, line 33: I thought it would be more logical to put the deletion of the cookie on signup here, after email validation. But as I don't have a mail server, I wasn't able to test it. Could that be a problem?

AntonKhorev: If you run osm-website in a terminal, emails are printed there. You can find the plaintext version of the confirmation email and copy the link from it (or more likely the part of it that starts with `/user` because the domain name is set to some nonexistent value).

AntonKhorev (approved):

AntonKhorev:
Merged, thanks! I pushed [some whitespace changes](https://github.com/openstreetmap/openstreetmap-website/compare/b49cdc187e50fcbb5fda72719847e9215c5b81f4..0f2df0b9efcfff1d6bb0f7b2d16fa23f85d8708d) to satisfy erblint test. That test didn't run because it needed approval, so you couldn't see it here.

PR #5469, opened 2025-01-05
Fix allow_thirdparty_images options

AntonKhorev:
When you write something like `allow_thirdparty_images :only => :index` you'd expect the CSP to be altered only on `index` action. But actually `:only => ...` was ignored and `allow_thirdparty_images` ran on all actions because `content_security_policy` didn't receive `options` correctly.

Other `allow_` methods from `app/controllers/application_controller.rb` should be similarly fixed, except I haven't figured out whether they are required at all. For example `allow_all_form_action` in `app/controllers/oauth2_authorizations_controller.rb` came from https://github.com/openstreetmap/openstreetmap-website/commit/b96f3867e61dad3d7f14a0d8da01ea0cab1c83ec, but are redirects done using form actions? Forms in `app/views/oauth2_authorizations/new.html.erb` have actions pointing to the osm website.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,7 +20,7 @@ class ApplicationController < ActionController::Base
   helper_method :oauth_token
 
   def self.allow_thirdparty_images(**options)
-    content_security_policy(options) do |policy|
+    content_security_policy(**options) do |policy|
       policy.img_src("*", :data)
     end
   end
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me, and the others should definitely be fixed as well.

I believe the `allow_all_form_action` one is related to a difference in how browsers enforce CSP for form submissions - specifically some (eg Firefox) only require that the initial URL submitted to meet the form action policy while others (eg Chrome) require any subsequent redirects to meet it as well.

Certainly the `session#new` one can lead to an eventual redirect after a successful authentication and I guess the OAuth one can as well.

PR #5470, opened 2025-01-05
Remove zoomend listener on history unload

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/commit/6b194ed6278d4b05a04a20d6907203ed329bef77 added a map zoom listener to history pages because rendered changeset rectangles depend on the map scale. But the listener is never removed on unload.


tomhughes:
Looks good to me, thanks.

PR #5471, opened 2025-01-05
Pass bbox to changeset page via data attribute

### LINKED ISSUE #4081: Javascript error when opening a page of a changeset without a bounding box
### URL

https://www.openstreetmap.org/changeset/138022970

See the browser console. 

- `Uncaught TypeError: this._northEast is undefined` in Firefox
- `Uncaught TypeError: Cannot read properties of undefined (reading 'lat')` in Chrome

_No response_

AntonKhorev:
Fixes #4081 on `/changeset/:id` pages.

Previously changeset bbox was loaded using `L.OSM.DataLayer`. That requied additional http request. Also `L.OSM.DataLayer` assumed that all changesets have bounding boxes which is not true.

### DIFF
--- a/app/assets/javascripts/index/changeset.js
+++ b/app/assets/javascripts/index/changeset.js
@@ -1,29 +1,26 @@
 OSM.Changeset = function (map) {
   var page = {},
-      content = $("#sidebar_content"),
-      currentChangesetId;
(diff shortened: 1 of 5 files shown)


tomhughes:
Looks good to me, thanks.

PR #5473, opened 2025-01-06
Support unwrapped bbox values in changeset history queries

### LINKED ISSUE #3423: Not all changesets shown in history view on low zoom levels
### URL

https://www.openstreetmap.org/history#map=1/0/0

To reproduce issue:
1. Got to OSM website
2. From the menu in top-left corner, select "History" (Ajalugu in attached screenshot)
3. Zoom out or adjust width of browser window until all continents are visible at least twice
4. (Optional): To make effects of the bug more clear, make sure that longitude of both edges of the map are approximately same

Issue: Only changesets from that very narrow gap are shown.
Expected behaviour: All changesets around the world are shown.

The bug is caused by Leaflet's method `.wrap()` on [line 62 of history.js](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/javascripts/index/history.js#L62). The easiest fix is to simply remove said method because it's used for offsetting user's bbox of current viewport into valid latitude/longitude. However, function that processes the passed parameter in [changesets_controller.rb](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/changesets_controller.rb#L90-L92) as of now uses simple number comparison to filter changesets and don't mind unusual bbox parameters. Bit more sophisticated fix is to add client-side bbox validation where too large bboxes would be clipped from -180/180 meridian. But bigger question is that since this issue was introduced as [single commit](https://github.com/openstreetmap/openstreetmap-website/commit/8fe5dbb92cec25467f1f1c23f65113c5e956a7ff) 8 years ago and not fixed since, is this a bug or an intentional feature to reduce server load?

I came across this bug while working on PR #3417 and had actually already fixed it there using the lazy method, but wanted to clarify if wrapping was intentional or it was added by accident.

Screenshot with sample of the issue (loaded bit more changesets than the default 20 to show the trend). Note the map boundaries.

pilt

AntonKhorev:
Fixes #3423.

> Issue: Only changesets from that very narrow gap are shown.

How narrow this area is depends on zoom and map view width. It's an area between wrapped left and (independently) wrapped right longitude bounds because:

> The bug is caused by Leaflet's method .wrap()

It's not Leaflet's method. Maybe it existed previously in Leaflet and was removed, but we added it again in https://github.com/openstreetmap/openstreetmap-website/commit/f612f573f92c524bbcf4b913e71c9f288383a845. Implementing wrap on `L.LatLngBounds` by wrapping its corners independently is not going to give correct results in a general case.

> Bit more sophisticated fix is to add client-side bbox validation where too large bboxes would be clipped from -180/180 meridian.

It's not difficult to write the logic without clipping. Unfortunately `BoundingBox.from_bbox_params` insists on clipping, that's why I'm not using `BoundingBox`.

> is this a bug or an intentional feature to reduce server load?

I don't think that the current behavior reduces server load. The more changesets you filter out, the longer you have to scan the table. If the current implementation filters out changesets that shouldn't have been filtered out, it slows down the server.

---

This PR fixes only how changesets are selected by the `/history` page based on the current view area, which gets translated into the `bbox` parameter. It doesn't change how bboxes are displayed, which needs further fixing. When you zoom out, you can see the same place on the map several times. If there's a changeset, it also needs to be rendered several times. That usually doesn't matter much because you can see one of its renderings. But that doesn't work if you zoom in close to the date line. Only one side is going to be visible on the map. On the other hand the list in the sidebar should be correct.

image
image

### DIFF
--- a/app/controllers/changesets_controller.rb
+++ b/app/controllers/changesets_controller.rb
@@ -54,7 +54,10 @@ class ChangesetsController < ApplicationController
                        changesets.where("false")
                      end
       elsif @params[:bbox]
-        changesets = conditions_bbox(changesets, BoundingBox.from_bbox_params(params))
+        bbox_array = @params[:bbox].split(",").map(&:to_f)
+        raise OSM::APIBadUserInput, "The parameter bbox must be of the form min_lon,min_lat,max_lon,max_lat" unless bbox_array.count == 4
+
+        changesets = conditions_bbox(changesets, *bbox_array)
       elsif @params[:friends] && current_user
         changesets = changesets.where(:user => current_user.followings.identifiable)
       elsif @params[:nearby] && current_user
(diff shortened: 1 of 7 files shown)


AntonKhorev:
Do you want the longitude span check on the javascript side? Something like this before sending the bbox:

```
if (max_lon - min_lon >= 360) {
  min_lon = -180;
  max_lon = 180;
}
```

This will eliminate [one out of three cases here](https://github.com/openstreetmap/openstreetmap-website/pull/5473/files#diff-cf3fe9dc357f802fd9155582ffc8c6f71185065267991a2135323ad2d044b4f3R157-R163). But how would you modify `wrap` to produce a valid bbox in cases like `min_lon = 170; max_lon = 190`?

app/controllers/changesets_controller.rb
```diff
@@ -113,21 +116,28 @@ def show
...
       changesets
+    elsif db_min_lon > db_max_lon
+      changesets.where("min_lon < ? or max_lon > ?", db_max_lon, db_min_lon)
+    else
+      changesets.where("min_lon < ? and max_lon > ?", db_max_lon, db_min_lon)
     end
   end
 
+  def wrap_lon(lon)
+    ((lon + 180) % 360) - 180
```
tomhughes, line 138: I don't think this works? If we consider a range from -270 to 270 and apply this to both then we wind up with 90 to -90 which is clearly nonsense?

Even considering a simple globe spanning -180 to 180 wraps both to -180 leaving us with an empty box...

tomhughes (changes requested):

app/controllers/changesets_controller.rb
AntonKhorev: > I don't think this works?

There are branches for "nonsense" cases.

> If we consider a range from -270 to 270 and apply this to both then we wind up with 90 to -90 which is clearly nonsense?

```
elsif db_min_lon > db_max_lon
```

> Even considering a simple globe spanning -180 to 180 wraps both to -180 leaving us with an empty box...

```
if max_lon - min_lon >= 360
```

AntonKhorev: Added tests for -270 to 270 and -180 to 180.

Also added `:bbox` argument to changesets factory to shorten `create(:changeset, ...)` in tests.

AntonKhorev:
Added comments to `conditions_bbox` because it might not be obvious why it works across the antimeridian.

tomhughes (commented):
Can you squash the final commit into the previous one please, so that the comments come in along with the code changes given that they're kind of important in explaining what the code is doing?

AntonKhorev:
Squashed comments into the previous commit.

tomhughes (approved):

PR #5474, opened 2025-01-06
Use baselayerchange/overlayadd/overlayremove instead of layeradd/layerremove

### LINKED ISSUE #5466: "layeradd layerremove" event handlers dramatically slow down Map data layer rendering
### URL

_No response_

I was trying to figure out why the Map Data layer is slow when displaying a lot of elements. In Firefox Profiler, I saw that a lot of time was spent on handlers for adding new objects to the map.

Снимок экрана 2025-01-04 в 21 18 56

https://share.firefox.dev/4a3ZXM9

Specifically:
https://github.com/openstreetmap/openstreetmap-website/blob/95b919c5d1e33688b0841c7d38602372c2e4c4a3/app/assets/javascripts/index.js#L176-L184

I don’t understand why these handlers need to update data about the current coordinates, and to prove their uselessness I did the following:

1. Updated the page with Map Data enabled
2. Waited for the data to download
3. Allowed display of a large amount of data

Rendering took a few seconds. 

After that, I did the same thing, but deleted the handlers with `map.off("layeradd layerremove")`. After that, the rendering is understated for less than a second.

https://github.com/user-attachments/assets/f8232635-7e97-4bd1-87ab-1d46b7b79e70

Further movement of the map is also noticeably faster and allows you to display tens of thousands of objects.

Note: `map` variable is not available by default, it is provided by my better-osm-org extension. This extension doesn't do anything else, and rendering performance doesn't depend on how it works.

You can also analyze the performance profile collected in the video: https://share.firefox.dev/4a3ZXM9

_No response_

deevroman:
### Description
This is fix #5466 After merging https://github.com/openstreetmap/leaflet-osm/pull/43 flame graph looks like this:

https://share.firefox.dev/4a2Wc9Z

Most of the time is spent on two handlers that update information about the current location and elements in the right sidebar. 

Unfortunately, replacing `layeradd layerremove` with `baselayerchange overlaylayerchange` (https://github.com/openstreetmap/openstreetmap-website/issues/5466#issuecomment-2571570472) is difficult, since in addition to the Map Data layer, other elements may be displayed on the map. For example, the currently selected object on the map. It is not clear how to separate changes to Map Data objects and changes to the selected element. It is necessary to track changes in the active element, otherwise at least the Edit button will break.

~~Therefore, I implemented a little ugly, but safer in terms of breaking business logic. When adding the Map Data layer, I set the pause flag in the slowdown handlers. While the Map Data is rendering, the user cannot interact with the interface, so it seems safe to pause the event handlers.~~

Total 15s vs 0.5s when rendering 5000 elements.

https://share.firefox.dev/4h4IK7C

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -40,6 +40,9 @@ L.OSM.Map = L.Map.extend({
       }
 
       const layer = new layerConstructor(layerOptions);
+      layer.on("add", () => {
+        this.fire("baselayerchange", { layer: layer });
+      });
       this.baseLayers.push(layer);
     }
 
@@ -53,8 +56,14 @@ L.OSM.Map = L.Map.extend({
       pane: "overlayPane",
       code: "G"
     });
+    this.gpsLayer.on("add", () => {
+      this.fire("overlayadd", { layer: this.gpsLayer });
+    }).on("remove", () => {
+      this.fire("overlayremove", { layer: this.gpsLayer });
+    });
 
-    this.on("layeradd", function (event) {
+
+    this.on("baselayerchange", function (event) {
       if (this.baseLayers.indexOf(event.layer) >= 0) {
         this.setMaxZoom(event.layer.options.maxZoom);
       }
@@ -109,14 +118,18 @@ L.OSM.Map = L.Map.extend({
     var layers = layerParam || "M",
         layersAdded = "";
 
-    for (var i = this.baseLayers.length - 1; i >= 0; i--) {
+    for (let i = this.baseLayers.length - 1; i >= 0; i--) {
+      if (layers.indexOf(this.baseLayers[i].options.code) === -1) {
+        this.removeLayer(this.baseLayers[i]);
+      }
+    }
+
+    for (let i = this.baseLayers.length - 1; i >= 0; i--) {
       if (layers.indexOf(this.baseLayers[i].options.code) >= 0) {
         this.addLayer(this.baseLayers[i]);
         layersAdded = layersAdded + this.baseLayers[i].options.code;
       } else if (i === 0 && layersAdded === "") {
         this.addLayer(this.baseLayers[i]);
-      } else {
-        this.removeLayer(this.baseLayers[i]);
       }
     }
   },
@@ -297,6 +310,7 @@ L.OSM.Map = L.Map.extend({
       }
 
       if (callback) callback(this._objectLayer.getBounds());
+      this.fire("overlayadd", { layer: this._objectLayer });
     } else { // element handled by L.OSM.DataLayer
       var map = this;
       this._objectLoader = $.ajax({
@@ -328,6 +342,7 @@ L.OSM.Map = L.Map.extend({
           map._objectLayer.addTo(map);
 
           if (callback) callback(map._objectLayer.getBounds());
+          map.fire("overlayadd", { layer: map._objectLayer });
         }
       });
     }
@@ -337,6 +352,7 @@ L.OSM.Map = L.Map.extend({
     this._object = null;
     if (this._objectLoader) this._objectLoader.abort();
     if (this._objectLayer) this.removeLayer(this._objectLayer);
+    this.fire("overlayremove", { layer: this._objectLayer });
   },
 
   getState: function () {
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -173,7 +173,7 @@ $(document).ready(function () {
   var expiry = new Date();
   expiry.setYear(expiry.getFullYear() + 10);
 
-  map.on("moveend layeradd layerremove", function () {
+  map.on("moveend baselayerchange overlayadd overlayremove", function () {
     updateLinks(
       map.getCenter().wrap(),
       map.getZoom(),
@@ -205,7 +205,7 @@ $(document).ready(function () {
   });
 
   if (OSM.MATOMO) {
-    map.on("layeradd", function (e) {
+    map.on("baselayerchange overlayadd", function (e) {
       if (e.layer.options) {
         var goal = OSM.MATOMO.goals[e.layer.options.layerId];
 
--- a/app/assets/javascripts/index/layers/data.js
+++ b/app/assets/javascripts/index/layers/data.js
@@ -25,18 +25,16 @@ OSM.initializeDataLayer = function (map) {
     onSelect(e.layer);
   });
 
-  map.on("layeradd", function (e) {
-    if (e.layer === dataLayer) {
-      map.on("moveend", updateData);
-      updateData();
-    }
+  dataLayer.on("add", function () {
+    map.fire("overlayadd", { layer: this });
+    map.on("moveend", updateData);
+    updateData();
   });
 
-  map.on("layerremove", function (e) {
-    if (e.layer === dataLayer) {
-      map.off("moveend", updateData);
-      $("#browse_status").empty();
-    }
+  dataLayer.on("remove", function () {
+    map.off("moveend", updateData);
+    $("#browse_status").empty();
+    map.fire("overlayremove", { layer: this });
   });
 
   function updateData() {
--- a/app/assets/javascripts/index/layers/notes.js
+++ b/app/assets/javascripts/index/layers/notes.js
@@ -23,10 +23,12 @@ OSM.initializeNotesLayer = function (map) {
   noteLayer.on("add", () => {
     loadNotes();
     map.on("moveend", loadNotes);
+    map.fire("overlayadd", { layer: noteLayer });
   }).on("remove", () => {
     map.off("moveend", loadNotes);
     noteLayer.clearLayers();
     notes = {};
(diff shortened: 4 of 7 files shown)


AntonKhorev:
> For example, the currently selected object on the map. It is not clear how to separate changes to Map Data objects and changes to the selected element. It is necessary to track changes in the active element, otherwise at least the Edit button will break.

`addObject`/`removeObject` in `leaflet.map.js` could fire some event for active object change. The listener that calls `updateLinks` could listen to that event.

deevroman:
I also saw a problem in PR: when you click the Map Data checkbox, the cookie with the current location is no longer updated

deevroman:
Okay, firing events from `addObject` / `removeObject` turned out to be the best idea. And I changed this PR to use `baselayerchange`/`overlaylayerchange`

I also replaced layeradd for `map.on("layeradd",...)`, which tracks the layer's on/off state with `dataLayer.on("add", ...)`. Using `overlaylayerchange` in this case doesn't work, because this event isn't thrown when opening a new page with `...&layers=D` Plus `add ' / 'remove' looks cleaner in my opinion.

mmd-osm:
With the latest changes I was able to render Lake Huron locally for the first time (455k nodes). That is definitely a significant speed-up.

image

HolgerJeromin:
Jquery uses `JSON.parse` for parsing of `ajax()` result.

In a next step switching to `fetch" would allow using
https://developer.mozilla.org/en-US/docs/Web/API/Response/json
Where the parsing is done without blocking the main thread.

AntonKhorev:
We probably need to fire `baselayerchange` either from `map.setState` or from `map.updateLayers`. Otherwise if you edit `&layers=X` in the location bar, share urls are updated to its previous value.

deevroman:
> Otherwise if you edit &layers=X in the location bar, share urls are updated to its previous value.

Do I understand correctly that you are talking about the current behavior of the website, and not after PR? At least in Firefox, changing `&layers=X ' does not change the layers on the map, only if you open in a new tab

mmd-osm:
@AntonKhorev 's comment wasn't exactly clear to me either. I think it's the following. Note that once you change the layer in the URL, the share link "layer" attribute is updated accordingly:

Peek 2025-01-14 22-58

deevroman:
Ah, I understand. I only tested changing the letters that represent the overlay (DNG). Changing the letters that respond to tile layers works now. PR broke it. I'll try to fix it

AntonKhorev:
Leaflet ships with a layers control which lets you switch between layers. That thing is responsible for [firing `baselayerchange` events](https://github.com/Leaflet/Leaflet/blob/c511e66c20dce06ac05c43cf287b3ecdba00d924/src/control/Control.Layers.js#L319). We don't use this control because we have our own layers menu. Therefore we're responsible for firing that event, which we do: https://github.com/openstreetmap/openstreetmap-website/blob/914fff9d4dec3d502d524f5b0f60797f66f8af65/app/assets/javascripts/leaflet.layers.js#L70

The problem is that our layers menu is not the only thing that can switch layers. All other places where layer change happens need also to fire `baselayerchange`. I think that `map.updateLayers` is the only such place. Lack of `baselayerchange` there becomes more noticeable with this PR because it uses this event.

Similar story with `overlaylayerchange`, although it's not a standard Leaflet event. Leaflet layers controluses `overlayadd` and `overlayremove`.

deevroman:
Okay, I added the `baselayerchange` firing. So, when changing D/N/G, the update does not work now, so firing `overlaylayerchange` is better done in a separate PR.

app/assets/javascripts/leaflet.map.js
```diff
@@ -116,6 +116,7 @@ L.OSM.Map = L.Map.extend({
       } else {
         this.removeLayer(this.baseLayers[i]);
       }
+      this.fire("baselayerchange", { layer: this.baseLayers[i] });
```
AntonKhorev, line 119: This fires the event for every possible base layer. It's supposed to be fired only for the layer that became visible, that's how it works in `leaflet.layers.js`.

AntonKhorev (changes requested):

app/assets/javascripts/leaflet.map.js
deevroman: Implemented a check that there is already a layer on the map. But the code doesn’t look so attractive anymore, so it might be better to add add/remove handlers for all base layers.

app/assets/javascripts/leaflet.map.js
```diff
@@ -108,14 +108,21 @@ L.OSM.Map = L.Map.extend({
...
+      var baselayerChanged = false;
       if (layers.indexOf(this.baseLayers[i].options.code) >= 0) {
+        baselayerChanged = !this.hasLayer(this.baseLayers[i]);
         this.addLayer(this.baseLayers[i]);
         layersAdded = layersAdded + this.baseLayers[i].options.code;
       } else if (i === 0 && layersAdded === "") {
+        baselayerChanged = !this.hasLayer(this.baseLayers[i]);
         this.addLayer(this.baseLayers[i]);
       } else {
+        baselayerChanged = this.hasLayer(this.baseLayers[i]);
```
AntonKhorev, line 120: `baselayerchange` is supposed to be fired *once* for the layer that became visible. It's not supposed to be fired for removed layers. See that `null` in Leafet code?
https://github.com/Leaflet/Leaflet/blob/c511e66c20dce06ac05c43cf287b3ecdba00d924/src/control/Control.Layers.js#L317-L319

"This event would be triggered after the other layer events, and it would include a layer property set to the resulting base layer." https://github.com/Leaflet/Leaflet/pull/1064

AntonKhorev (changes requested):

app/assets/javascripts/leaflet.map.js
AntonKhorev: You could put `this.fire ...` next to `this.addLayer ...`, that's one extra line of code.

AntonKhorev:
Do we fix `baselayerchange` everywhere before doing this PR that changes listeners for this event?

app/assets/javascripts/leaflet.map.js
deevroman: So the next handler also violates this expectation?

https://github.com/openstreetmap/openstreetmap-website/blob/b45eb03a6c026ec0f19415100c7bb14c4c53e6cb/app/assets/javascripts/leaflet.layers.js#L62-L71

AntonKhorev: The answer to the above question was *yes* but it probably didn't affect anything. However if we start using `baselayerchange` in more places, it might. For example, this code, if rewritten for `baselayerchange` wouldn't work correctly, if it expects this event to happen like Leaflet would fire it:

https://github.com/openstreetmap/openstreetmap-website/blob/b45eb03a6c026ec0f19415100c7bb14c4c53e6cb/app/assets/javascripts/index.js#L208-L216

deevroman: It seems that `"baselayerchange overlaylayerchange"` will work in this place

deevroman:
@mmd-osm 
> Besides, there may be some timing issue in the current code, as the behavior is not 100% reproducible for me, even today. Sometimes the map layer changes along with the share link, in other cases, the old layer is still shown.

I think I found a way to reproduce this (If you meant it):

1. Open https://www.openstreetmap.org/#map=19/60.008946/30.217791&layers=Y
2. Change Y -> H
3. Change H -> Y (the layer has not changed)


https://github.com/user-attachments/assets/31a836d2-405e-4c17-9877-c7544d6d879d

deevroman:
I remade PR. I abandoned `overlaylayerchange` because this event is not in Leaflet and it is not suitable in our situation.

In the case of `baselyerchange`, we can expect that our last base layer is always removed and a new base layer is added. Therefore, the removal event is not necessary to track the layer, `baslaverchange` is enough. But in the case of overlays, this is not so, overlays can just be removed from the map, without adding a new one.

However, after rejecting `layerremove`, it becomes important to maintain the invariant that there can't be multiple base layers on the map. I.e. first remove all previous layers and only then add a new one (this is important for example for a field with an iframe, sometimes you can see two base layers at the same time)

So I use `overlayadd`/`overlayremove` which are in Leaflet https://leafletjs.com/reference.html#map-event
I fire these events when the `add`/` remove` is triggered, as well as for active objects that should be shown on the map (current node, paths, note, ...)

---

What was tested:

1: Basic layer switching functionality

https://github.com/user-attachments/assets/1d7bce4e-3ab6-4366-874f-6a2bb16afb00

2. Switching layers in the URL

https://github.com/user-attachments/assets/6a201eab-65cd-44a9-804e-1efddcc3e256

3. Update links to the active object in the Edit button

https://github.com/user-attachments/assets/539fd547-09af-4d0f-8299-0b1699581a2b

AntonKhorev:
Merged, thanks.

mmd-osm:
I will create the PR to switch to JSON format shortly: https://github.com/mmd-osm/openstreetmap-website/tree/patch/datalayerjson

PR #5475, opened 2025-01-06
Make browse not found pages consistent

AntonKhorev:
*Not found* pages for various types became slightly different. This PR makes them similar to each other.

The types are (Old)(Node/Way/Relation), Note and Changeset.

- The main template is kept at `browse/not_found`. 
- The type-specific part is the message displayed on that page, it's kept in `not_found_message` templates.
- The string with word substitution "Sorry, %{type} #%{id} could not be found." is replaced by type-specific strings like "Sorry, changeset #%{id} could not be found." where only id (and version numbers in some cases) are substituted. Those should be easier to translate.


tomhughes:
Looks good to me, thanks.

PR #5478, opened 2025-01-07
Report deactivates_at as block read date

AntonKhorev:
I changed `updated_at` to `deactivates_at` in #5476 and forgot to make the same change on the next line. This can cause block update dates to be reported as block read dates. Those dates are different for blocks that were edited by moderators after they were read. The reason text can still be updated for such blocks.


tomhughes:
Looks good to me, thanks.

PR #5479, opened 2025-01-07
Add ElementsController base class

AntonKhorev:
Like #5464 but for current elements.

Also like `Api::ElementsController` in #3715 if that gets merged.


tomhughes:
Looks good to me, thanks.

PR #5483, opened 2025-01-08
Restrict version of dry-schema for ruby 3.1.2 compatibility

### LINKED ISSUE #5482: dry-schema update exposes bug in debian/bookworm version of ruby 3.1
In https://github.com/openstreetmap/openstreetmap-website/commit/045f0ac1bec2797b1e8f743d88d1a58687af24c7 we updated dry-schema from 1.13.4 to 1.14.0. We're now getting build failures with the message:

```
SyntaxError: /var/lib/gems/3.1.0/gems/dry-schema-1.14.0/lib/dry/schema/dsl.rb:185: no anonymous block parameter
```

I've checked that using an older version of dry-schema (and dry-validation, which depends on this) works fine:

```
gem "dry-schema", "= 1.13.0"
gem "dry-validation", "= 1.10.0"
```

I suspect this is due to https://bugs.ruby-lang.org/issues/18673 which is fixed in ruby 3.1.3, but in debian/bookworm we currently have 3.1.2-7 (and it's not clear to me how to find out which patches debian have backported from 3.1.3 to their version of 3.1.2).

https://packages.debian.org/bookworm/ruby3.1

I think options include pinning dry-schema to 1.13.4, or figuring out if debian need to update their ruby, or something else?

gravitystorm:
https://bugs.ruby-lang.org/issues/18673 affects some methods using anonymous blocks, and was fixed in ruby 3.1.3. However, debian/bookworm ruby 3.1 package is based on 3.1.2 and does not yet contain this fix.

Since bookworm is used in our Docker build, and also used in production by OSMF, then we can restrict dry-schema to a version that does not use anonymous blocks.

Fixes #5482

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ gem "bootstrap_form", "~> 5.0"
 gem "cancancan"
 gem "config"
 gem "delayed_job_active_record"
+gem "dry-schema", "< 1.14.0" # see https://github.com/openstreetmap/openstreetmap-website/issues/5482
 gem "dry-validation"
 gem "frozen_record"
 gem "http_accept_language", "~> 2.1.1"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -230,13 +230,13 @@ GEM
       concurrent-ruby (~> 1.0)
       dry-core (~> 1.1)
       zeitwerk (~> 2.6)
-    dry-schema (1.14.0)
+    dry-schema (1.13.4)
       concurrent-ruby (~> 1.0)
       dry-configurable (~> 1.0, >= 1.0.1)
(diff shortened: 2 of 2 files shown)


nenad-vujicic (approved):
This works great on my side! Thanks!

tomhughes:
Much as it pains me I guess we have to take this.

tomhughes:
Why has this downgraded us all the way to `1.13.0` when we were on `1.13.14` before I updated yesterday?

gravitystorm:
> Why has this downgraded us all the way to `1.13.0` when we were on `1.13.4` before I updated yesterday?

Sorry, that's my mistake, related to the order I was testing things locally. I've pushed an updated version.

PR #5484, opened 2025-01-08
Show diary entry after creation

gravitystorm:
This allows users to review their post and use the social share buttons more easily.

Refs #1391

### DIFF
--- a/test/controllers/diary_entries_controller_test.rb
+++ b/test/controllers/diary_entries_controller_test.rb
@@ -159,8 +159,8 @@ class DiaryEntriesControllerTest < ActionDispatch::IntegrationTest
       post diary_entries_path(:diary_entry => { :title => "New Title", :body => "This is a new body for the diary entry", :latitude => "1.1",
                                                 :longitude => "2.2", :language_code => "en" })
     end
-    assert_redirected_to :action => :index, :display_name => user.display_name
     entry = DiaryEntry.last
+    assert_redirected_to diary_entry_path(user, entry)
     assert_equal user.id, entry.user_id
     assert_equal "New Title", entry.title
     assert_equal "This is a new body for the diary entry", entry.body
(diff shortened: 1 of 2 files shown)


test/controllers/diary_entries_controller_test.rb
```diff
@@ -159,7 +159,7 @@ def test_create
       post diary_entries_path(:diary_entry => { :title => "New Title", :body => "This is a new body for the diary entry", :latitude => "1.1",
                                                 :longitude => "2.2", :language_code => "en" })
     end
-    assert_redirected_to :action => :index, :display_name => user.display_name
+    assert_redirected_to diary_entry_path(user, user.diary_entries.last)
     entry = DiaryEntry.last
```
AntonKhorev, line 163: It's a bit odd that the same diary entry has two different references next to each other: `user.diary_entries.last` and `DiaryEntry.last`.

gravitystorm: Yes, I thought that too. I could refactor that as part of this PR, or separately. Which would you prefer? (and do you have any preferences for referring to the entry e.g. `user.diary_entries.last` or `DiaryEntry.last` or something else?)

AntonKhorev: > separately

Won't that be a fixup commit which we don't like here?

> do you have any preferences for referring to the entry

Maybe if we use `assert_difference "DiaryEntry.count"`, we should also use `DiaryEntry.last` (or change `assert_difference`).

gravitystorm: > Maybe if we use `assert_difference "DiaryEntry.count"`, we should also use `DiaryEntry.last` (or change `assert_difference`).

I've gone for DiaryEntry.last, but rearranged slightly so that we use the `entry` variable for the redirection assertion too. Let me know what you think.

AntonKhorev:
Merged, thanks

PR #5485, opened 2025-01-08
Moves the note's description, author ID, and author IP from the first comment to the note itself

nenad-vujicic:
### Description
PR improves using of note's description / author_id / author_ip by:

1. Adding new fields to notes table for storing these 3 values
2. Copying values from first note's comment to notes table
3. Replacing using note's description / author / author_id from first note's comment with from note's table directly

A few comments:
- We left adding description to body when creating new note - it looked dirty to add blank string ("") or similar, also, left comments body intact during data-migration for improving performances,
- We kept backward compatibility of generated XML / JSON / .. by inserting note's description as a first comment's text,
- We added more robust version of data-migration script for retrying migration of failed notes.

### How has this been tested?
By running automated tests and by semi-manual testing (semi-manual generating notes, doing migration and applying changes, manually checking rendered notes and new memory variables content).

### DIFF
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -81,19 +81,22 @@ module Api
       # Extract the arguments
       lon = OSM.parse_float(params[:lon], OSM::APIBadUserInput, "lon was not a number")
       lat = OSM.parse_float(params[:lat], OSM::APIBadUserInput, "lat was not a number")
-      comment = params[:text]
+      description = params[:text]
+
+      # Get note's author info (for logged in users - user_id, for logged out users - IP address)
+      note_author_info = author_info
 
       # Include in a transaction to ensure that there is always a note_comment for every note
       Note.transaction do
         # Create the note
-        @note = Note.create(:lat => lat, :lon => lon)
+        @note = Note.create(:lat => lat, :lon => lon, :description => description, :user_id => note_author_info[:user_id], :user_ip => note_author_info[:user_ip])
         raise OSM::APIBadUserInput, "The note is outside this world" unless @note.in_world?
 
         # Save the note
         @note.save!
 
-        # Add a comment to the note
-        add_comment(@note, comment, "opened")
+        # Add opening comment to the note
+        add_comment(@note, description, "opened")
       end
 
       # Return a copy of the new note
@@ -380,10 +383,8 @@ module Api
     end
 
     ##
-    # Add a comment to a note
-    def add_comment(note, text, event, notify: true)
-      attributes = { :visible => true, :event => event, :body => text }
-
+    # Get author's information (for logged in users - user_id, for logged out users - IP address)
+    def author_info
       if doorkeeper_token
         author = current_user if scope_enabled?(:write_notes)
       else
@@ -391,9 +392,23 @@ module Api
       end
 
       if author
-        attributes[:author_id] = author.id
+        { :user_id => author.id }
       else
-        attributes[:author_ip] = request.remote_ip
+        { :user_ip => request.remote_ip }
+      end
+    end
+
+    ##
+    # Add a comment to a note
+    def add_comment(note, text, event, notify: true)
+      attributes = { :visible => true, :event => event, :body => text }
+
+      note_comment_author_info = author_info
+
+      if author_info[:user_id]
+        attributes[:author_id] = note_comment_author_info[:user_id]
+      else
+        attributes[:author_ip] = note_comment_author_info[:user_ip]
       end
 
       comment = note.comments.create!(attributes)
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -2,20 +2,27 @@
 #
 # Table name: notes
 #
-#  id         :bigint(8)        not null, primary key
-#  latitude   :integer          not null
-#  longitude  :integer          not null
-#  tile       :bigint(8)        not null
-#  updated_at :datetime         not null
-#  created_at :datetime         not null
-#  status     :enum             not null
-#  closed_at  :datetime
+#  id          :bigint(8)        not null, primary key
+#  latitude    :integer          not null
+#  longitude   :integer          not null
+#  tile        :bigint(8)        not null
+#  updated_at  :datetime         not null
+#  created_at  :datetime         not null
+#  status      :enum             not null
+#  closed_at   :datetime
+#  description :text             default(""), not null
+#  user_id     :bigint(8)
+#  user_ip     :inet
 #
 # Indexes
 #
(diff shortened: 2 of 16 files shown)


tomhughes:
How does this relate to the notes restructuring plan from #3831? Is this a step on the way to that or is it going in a different direction?

kcne:
Yes, this does align with the goals of #3831. This is also alternative to #4481 in a way. In fact, the initial motivation for this change came from a [comment](https://github.com/openstreetmap/openstreetmap-website/issues/5294#issuecomment-2541604426) in #5294, which suggested that note versioning could be a right approach on adding tags. To implement note versioning properly, this PR serves as a first step towards that. It might be worth opening a new issue specifically about note versioning, linking back to that comment for context.

AntonKhorev:
> How does this relate to the notes restructuring plan from https://github.com/openstreetmap/openstreetmap-website/issues/3831?

What I was going to say in comments for #5294 is that the plan from #3831 tries to solve [*(a)* by ignoring *(b)*](https://github.com/openstreetmap/openstreetmap-website/issues/5294#issuecomment-2541527509). You can read the original post of #3831 and see that it doesn't mention not actions at all. And those actions, they are stored in the comments table. You'll still have another *leaky abstraction* as that post says. Note comments won't be really comments. You'll close a note without a comment, that will still generate a "comment" entry in `NoteComments`.

> Is this a step on the way to that or is it going in a different direction?

The direction is hopefully the same for things stated explicitly in #3831. It's different for things omitted.

nenad-vujicic:
This PR should be first step toward adding editable note-tags and solving #3831. In next PRs we planned to:

1. Add notes versioning (add version column to `notes` table and add new table `old_notes` variant),
2. Add note tags (similar to #5344),
3. Add displaying history of notes (API, website, pagination, ... - this will be further decomposed),
4. Add user-interface for creating / updating / deleting note tags (this will be further decomposed),
5. Remove event from `NoteComment` and special note-comments,
6. ...

We thought to do above as a sequence of as small as possible PRs (like this one), but if you would prefer one big, we can do it? Every comment / suggestion is very welcome :-)

tomhughes:
I hadn't previously noticed that #5294 had proposed adding versioning - that's a whole other level of additional complication.

I can understand why it might be needed for tags, and I guess it also allows for editing the initial description if we want once that it part of the note itself rather than being a special comment. I do wonder how it interacts with comments though - do we duplicate those for every version like we do with members of other objects or do we always show all comments whatever the version or maybe track add the current version to the comment when it's created and then show comments up to the note's version when showing an old version of the note?

tomhughes (changes requested):
This is going to need to be broken down in to multiple PRs as we can't take all three migrations in one go, and we can't take code changes that rely on new fields until after those fields have been created.

The first two PRs should be one to create new methods on `Note` that correspond to the new fields but just look (for now) at the first comment, and one to add the new fields to `Note`.

After that a PR to validate the new key and one to adjust the new methods in `Note` to take the description from the note if the first comment has been migrated.

Only after that can we have a PR to actually do the migration.

nenad-vujicic:
> ... I do wonder how it interacts with comments though - do we duplicate those for every version like we do with members of other objects or do we always show all comments whatever the version or maybe track add the current version to the comment when it's created and then show comments up to the note's version when showing an old version of the note?

Our current plan is to `always show all comments whatever the version`. However, if others prefer a different approach, we are open to making updates.

AntonKhorev:
> I do wonder how it interacts with comments though - do we duplicate those for every version like we do with members of other objects or do we always show all comments whatever the version or maybe track add the current version to the comment when it's created and then show comments up to the note's version when showing an old version of the note?

I wouldn't say "duplicate (comments) for every version like we do with members of other objects" - do we even have comments for versioned objects elsewhere?

The idea is to take the current comments, let's call them *api 0.6 comments*, and split them into actual comments and versions. Versions will contain states (open/closed/hidden) that were previously in api 0.6 comments. After that it should be possible to reassemble actual comments and versions into api 0.6 comments for use in api calls. This reassembling wouldn't require associating comments with versions if I'm correct.

grekko:
Hi @nenad-vujicic thanks for putting in the effort to address https://github.com/openstreetmap/openstreetmap-website/issues/3831 

You might have noticed that I've started some time ago working on https://github.com/openstreetmap/openstreetmap-website/pull/4481 to address the same issue. Since your first PR in this direction (https://github.com/openstreetmap/openstreetmap-website/pull/5494) has been merged & approved I noticed the merge conflicts in my PR when I tried to merge the latest changes to `master` into my branch and wonder if I should close my PR and let you take over or if there is a way that I can contribute anything to your work in order to move things forward a bit faster.

Since your efforts have picked up more traction than mine, let me know how you would prefer to proceed at this point.

nenad-vujicic:
> Hi @nenad-vujicic thanks for putting in the effort to address #3831
> 
> You might have noticed that I've started some time ago working on #4481 to address the same issue. Since your first PR in this direction (#5494) has been merged I noticed the merge conflicts when I tried to merge the latest changes into my branch and wonder if I should close my PR and let you take over or if there is a way that I can contribute anything to your work in order to move things forward a bit faster.
> 
> Since your efforts have picked up more traction than mine, let me know how you would prefer to proceed at this point.

Hi @grekko, thanks for asking this! We hope we'll merge transition from using first comment to using notes records, adding notes versions, adding editable note tags and removing of special first comment in next few months. There are a lot of stuffs [in reworking proposal](https://github.com/openstreetmap/openstreetmap-website/issues/3831#issuecomment-1341562044) and [refactor notes document](https://github.com/osmfoundation/ewg_bidding/pull/9/files) which are not directly referenced by above plan, like improving creation process, ... Also, any comment on #5511 and others are highly appreciated :-)

grekko:
> Hi @grekko, thanks for asking this! We hope we'll merge transition from using first comment to using notes records, adding notes versions, adding editable note tags and removing of special first comment in next few months. There are a lot of stuffs [in reworking proposal](https://github.com/openstreetmap/openstreetmap-website/issues/3831#issuecomment-1341562044) and [refactor notes document](https://github.com/osmfoundation/ewg_bidding/pull/9/files) which are not directly referenced by above plan, like improving creation process, ... Also, any comment on #5511 and others are highly appreciated :-)

My question was more specifically how you think I should handle the overlapping goals of your PRs and the changes in #4481 :)

I decided to close my PR since it has been dangling for a year now while your work is getting traction and being merged.

If you feel that you want to split up some work and want to invest the time to coordinate, let me know but I'll let you keep the lead here 💪🏽

nenad-vujicic:
Closed in favor of PRs #5568, #5579 and PR #5609.

PR #5487, opened 2025-01-09
Move trace descriptions out of success/failure sentences in emails

### LINKED ISSUE #4012: Further rework lego messages for trace notification emails
In #2268 an effort was made to reduce lego messages. However it appears trace notification emails still include lego messages.

Here sentence beginning and ending are glued together without allowing different ordering in translations:

```
<p>
  <%= render :partial => "gpx_description" %>
  <%= t(".loaded", :trace_points => @trace_points, :count => @possible_points) %>
</p>
```

```
<p>
  <%= render :partial => "gpx_description" %>
  <%= t ".failed_to_import" %>
</p>
```

AntonKhorev:
Fixes #4012. As I say in https://github.com/openstreetmap/openstreetmap-website/issues/4012#issuecomment-2529365698 the current success/failure sentences are too long because we try to put in trace name/description/tags into them. I don't think anyone wants to read sentences that long. Shorter sentences won't have to be combined from parts, which will improve translatability.

## Before

HTML success email:
image

## After

HTML success email:
image

HTML failure email:
image

Plaintext success email:
```
Hi fakeuser1,

It looks like your file loaded successfully with 295 out of a possible 295 points.

Your file details:

Filename
	11716462.gpx
URL
	http://openstreetmap.example.com/user/fakeuser1/traces/100
Description
	new trace description here
Tags
	new, trace, tags, here


All your successfully uploaded GPX traces can be found at http://openstreetmap.example.com/traces/mine
```

Plaintext failure email:
```
Hi fakeuser1,

It looks like your file failed to be imported as a GPS trace.

Your file details:

Filename
	Screenshot_from_2025_01_06_12_50_06.png
Description
	new trace description here
Tags
	new, trace, tags, here


Please verify that your file is a valid GPX file or an archive containing GPX file(s) in the supported format (.tar.gz, .tar.bz2, .tar, .zip, .gpx.gz, .gpx.bz2). Could there be a format or syntax issue with your file? Here is the importing error:

==
Fatal error: Start tag expected, '<' not found at :1.
==

More information about GPX import failures and how to avoid them can be found at https://wiki.openstreetmap.org/wiki/GPX_Import_Failures
```

### DIFF
--- a/app/views/user_mailer/_gpx_description.html.erb
+++ b/app/views/user_mailer/_gpx_description.html.erb
@@ -1,9 +1,16 @@
-<% trace_name = tag.strong(@trace_name) %>
-<% trace_name = link_to(trace_name, @trace_url) if @trace_url %>
-<% trace_description = tag.em(@trace_description) %>
-<% if @trace_tags.length > 0 %>
-  <% tags = safe_join @trace_tags.map { |trace_tag| tag.em trace_tag.tag }, ", " %>
-  <%= t ".description_with_tags_html", :trace_name => trace_name, :trace_description => trace_description, :tags => tags %>
-<% else %>
-  <%= t ".description_with_no_tags_html", :trace_name => trace_name, :trace_description => trace_description %>
-<% end %>
+<p><%= t ".details" %></p>
+
+<dl>
+  <dt><%= t ".filename" %>
+  <% if @trace_url %>
+    <dd><%= link_to(@trace_name, @trace_url) %>
+  <% else %>
+    <dd><%= @trace_name %>
+  <% end %>
+  <dt><%= t ".description" %>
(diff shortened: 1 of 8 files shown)


nenad-vujicic (approved):
This works and looks great on my side. Thank you very much!

I would just suggest putting the property and value in the same row, as shown in

image

However, this is a bit subjective and can be addressed in another PR if others prefer.

AntonKhorev:
> I would just suggest putting the property and value in the same row

In the plaintext version? I didn't do that because I didn't want to count characters in localization strings.

tomhughes:
I don't think we need to worry too much about the details of the text version as most people are never going to see it.

This looks pretty good to me I think, thanks.

PR #5493, opened 2025-01-12
Label administrative boundaries by real-world designation

### LINKED ISSUE #5450: Change naming of boundaries based on admin_level=*
### Problem

As discussed in [this forum thread](https://community.openstreetmap.org/t/admin-level-boundary-naming-on-osm-org-and-translation/123358/19) the current naming scheme for boundaries is a peculiar mishmash of terminology from different English speaking countries that actually applies to none at all. This not only results in weird names for boundaries in English, but inconsistency in translation of the website, as some communities try to rectify this on their own, or have conflicting translations for English terminology.

### Description

 I propose that these are reverted to generic terminology, perhaps based on pure admin_level number, until a more appropriate solution on a per country basis can be implemented. Minh Nguyen has proposed [one such solution](https://github.com/openstreetmap/openstreetmap-website/blob/bdcd9ce35c71c0cf1b485dfb0a11ea54a01f057b/config/locales/en.yml#L1440-L1450), albeit as part of his larger proposal to convert the site to American English by default, with which I disagree.

### Screenshots

_No response_

1ec5:
Search results for Nominatim and Overpass queries now label administrative boundaries according to the real-world designation according to the [`border_type=*`](https://wiki.openstreetmap.org/wiki/Key:border_type#Administrative_boundaries) tag. If the value of this tag is unrecognized, the label falls back to a more generic “Administrative Boundary (Level _n_)”, where _n_ is the `admin_level=*` value. The English localization no longer attempts to translate these administrative levels to a fictional country’s administrative hierarchy.

<table>
<thead>
<tr>
<th>Before</th><th>After</th>
</tr>
</thead>
<tbody>
<tr></tr>
<tr>
<td>

[County Boundary, City Boundary, Municipality Boundary, County Boundary, Country Boundary, City Boundary, State Boundary](https://www.openstreetmap.org/query?lat=39.252971&lon=-84.281251#map=18/39.253109/-84.280527)

</td>
<td>

[County Boundary, City Boundary, Township Boundary, County Boundary, International Boundary, City Boundary, State Boundary](http://0.0.0.0:3000/query?lat=39.252971&lon=-84.281251#map=18/39.253109/-84.280527)

</td>

</tr>
<tr></tr>
<tr>
<td>

[County Boundary, State Boundary, Country Boundary, Municipality Boundary, Country Boundary, City Boundary, County Boundary, Municipality Boundary, Region Boundary](https://www.openstreetmap.org/query?lat=43.068068&lon=-79.019396#map=18/43.068339/-79.018591)

</td>
<td>

[County Boundary, State Boundary, Provincial Boundary, International Boundary, Administrative Bounary (Level 8), Administrative Boundary (Level 6), Town Boundary, Administrative Boundary (Level 5)](http://0.0.0.0:3000/query?lat=43.068068&lon=-79.019396#map=18/43.068339/-79.018591)

</td>
</tr>
<tr></tr>
<tr>
<td>

[Country Boundary, State Boundary, Country Boundary, County Boundary, Region Boundary, Country Boundary, State Boundary, County Boundary, State Boundary, County Boundary, Country Boundary, Administrative Boundary, Municipality Boundary, City Boundary](https://www.openstreetmap.org/query?lat=48.96610&lon=8.23288#map=19/48.966040/8.232842)

</td>
<td>

[International Boundary, Administrative Boundary (Level 4), International Boundary, Departmental Boundary, Administrative Boundary (Level 5), International Boundary, State Boundary, County Boundary, State Boundary, County Boundary, International Boundary, Administrative Boundary, Administrative Boundary (Level 7)](http://0.0.0.0:3000/query?lat=48.96610&lon=8.23288#map=19/48.966040/8.232842)

</td>

</tr>
<tr></tr>
<tr>
<td>

[County, City, Hamlet, City, County, City, County, City, County](https://www.openstreetmap.org/search?query=Hamilton#map=11/39.1673/-84.5384)

</td>
<td>

[County, City, Hamlet, City, County, City, County, City, County](http://0.0.0.0:3000/search?query=Hamilton#map=11/39.1673/-84.5384)

</td>
</tr>
</tbody>
</table>

There are [many tagging schemes](https://wiki.openstreetmap.org/wiki/Key:border_type#Alternatives) for the real-world designation, each with their advantages and disadvantages, but I focused on `border_type=*` for now because it’s the most common scheme globally.

Fixes #5450.


config/locales/en.yml
```diff
@@ -1489,16 +1489,43 @@ en:
...
+        national: "International Boundary"
+        neighbourhood: "Neighborhood Boundary"
+        parish: "Parish Boundary"
+        province: "Provincial Boundary"
+        região: "Regional Boundary"
+        region: "Regional Boundary"
+        state: "State Boundary"
+        town: "Town Boundary"
+        township: "Township Boundary"
+        village: "Village Boundary"
```
1ec5, line 1528: For starters, I added a translatable string for each of the 30 or so `border_type=*` values that occur most frequently with `boundary=administrative`. Only one translatable string is provided per `border_type=*` value globally. This will likely present a problem for some languages that would translate English words like “district”, “county”, or “town” differently depending on the country. I think we’ll need to add some more specific strings qualified by country code, as explained in https://github.com/openstreetmap/openstreetmap-website/issues/5450#issuecomment-2566049746, but it’ll require a bit more research to find out which values are important in which countries.

app/controllers/geocoder_controller.rb
```diff
@@ -101,9 +101,13 @@ def search_osm_nominatim
       if klass == "boundary" && type == "administrative"
         rank = (place.attributes["address_rank"].to_i + 1) / 2
         prefix_name = t "geocoder.search_osm_nominatim.admin_levels.level#{rank}", :default => prefix_name
+        border_type = false
+        place_type = false
```
tomhughes, line 105: I think `nil` would be a better initial value here given they are going to be strings if any match is found.

tomhughes (commented):
I'm not totally convinced showing the numeric level is useful but generally this looks good apart from the specific point I've mentioned in a comment.

config/locales/en.yml
```diff
@@ -1489,16 +1489,43 @@ en:
...
+        level2: "International Boundary"
+        level3: "Administrative Boundary (Level 3)"
+        level4: "Administrative Boundary (Level 4)"
+        level5: "Administrative Boundary (Level 5)"
+        level6: "Administrative Boundary (Level 6)"
+        level7: "Administrative Boundary (Level 7)"
+        level8: "Administrative Boundary (Level 8)"
+        level9: "Administrative Boundary (Level 9)"
+        level10: "Administrative Boundary (Level 10)"
+        level11: "Administrative Boundary (Level 11)"
```
AntonKhorev, line 1501: Those are longer prefixes, results often don't fit on one line and have line breaks in inconvenient places:

image

1ec5: I noticed that other boundary types don’t say “Boundary”. For example, `boundary=electoral` is just “Electoral”. We could shorten “Administrative Boundary (Level 4)” to just “Administrative (Level 4)”, but I’m ambivalent about similarly truncating the `border_type=*`-based labels, such as “City ~~Boundary~~”. That would easily get confused with the `place=*`-based labels, which follow different criteria.

Another option would be to drop the level suffix, since @tomhughes doesn’t seem enamored with including this information anyways. To the extent that this introduces ambiguity in the search results, the solution would be to tag boundaries with `border_type=*`, which is a good idea anyways.

app/controllers/geocoder_controller.rb
```diff
@@ -101,9 +101,13 @@ def search_osm_nominatim
...
         place.elements["extratags"].elements.each("tag") do |extratag|
-          prefix_name = t "geocoder.search_osm_nominatim.prefix.place.#{extratag.attributes['value']}", :default => prefix_name if extratag.attributes["key"] == "linked_place" || extratag.attributes["key"] == "place"
+          border_type = t "geocoder.search_osm_nominatim.border_types.#{extratag.attributes['value']}", :default => prefix_name if extratag.attributes["key"] == "border_type"
+          place_type = t "geocoder.search_osm_nominatim.prefix.place.#{extratag.attributes['value']}", :default => prefix_name if extratag.attributes["key"] == "linked_place" || extratag.attributes["key"] == "place"
```
AntonKhorev, line 108: Shouldn't `:default` be `border_type` and `place_type` correspondingly, given that later the values are combined with `place_type || border_type || prefix_name`?

1ec5: Yes, good catch. I originally had two loops that set `prefix_name` directly. Now we probably don’t need a `:default` at all.

config/locales/en.yml
AntonKhorev: > I noticed that other boundary types don’t say “Boundary”. For example, boundary=electoral is just “Electoral”. 

Those are constructed here:
https://github.com/openstreetmap/openstreetmap-website/blob/5d76ec051e2c429b6647401674e13688c1251956/app/assets/javascripts/index/query.js#L91-L100

AntonKhorev: I thought about using icons in search/query result lists, like on changeset/element pages. For administrative boundaries it could be something like a square with a border similar to the boundary line rendered on the map, maybe with a number inside for the level.

1ec5: Not a bad idea overall, though the numbers probably aren’t as intuitive for many mappers as designations coming from either `border_type=*` or `place=*`. I had forgotten about the code to transform an unrecognized tag into a label. As long as we have labels, we could fall back to a format string like “_Border Type_ Boundary” if `border_type=*` is present but set to an unrecognized value.

AntonKhorev: Numbers tell how relatively big/important the area is, something that may not be clear otherwise.

1ec5: Would the number be clearer than the locally common word for the type of boundary? Mappers who specialize in boundaries would know the local `admin_level=*` scale by heart, but others would find words more intuitive and would already know which are larger than others. Also, the Overpass results are ranked by size.

tomhughes:
Are you able to address the review comments so we can think about getting this merged?

app/controllers/geocoder_controller.rb
1ec5: Fixed.

1ec5:
Thanks for your patience. I’ve addressed the code-level feedback. https://github.com/openstreetmap/openstreetmap-website/pull/5493#discussion_r1912562048 is unresolved; there are a couple options for making it briefer (in English at least), but not without losing some detail.

tomhughes (approved):
Other than the question of the length of the new `admin_level` based levels this looks fine to me, and in practice I have yet to see them appear in testing as `border_type` takes priority.

I think this is good to go and we can revisit those labels later if they turn out to be a problem.

PR #5494, opened 2025-01-13
Adds description and author_id methods to Note

nenad-vujicic:
### Description
PR adds `description` and `author_id` (`author_ip` is already there) methods to Note model file. Methods use first comment for retrieving requested information. Also, adds unit tests for testing functionality of added methods.

This PR is 1st from set of PRs described [here](https://github.com/openstreetmap/openstreetmap-website/pull/5485#pullrequestreview-2545467592).

### How has this been tested?
Automated unit tests and manual testing

### DIFF
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -82,12 +82,22 @@ class Note < ApplicationRecord
     closed_at + DEFAULT_FRESHLY_CLOSED_LIMIT
   end
 
-  # Return the author object, derived from the first comment
+  # Return the note's description, derived from the first comment
+  def description
+    comments.first.body
+  end
+
+  # Return the note's author object, derived from the first comment
   def author
     comments.first.author
   end
 
-  # Return the author IP address, derived from the first comment
+  # Return the note's author ID, derived from the first comment
+  def author_id
+    comments.first.author_id
+  end
+
+  # Return the note's author IP address, derived from the first comment
   def author_ip
     comments.first.author_ip
   end
--- a/app/views/notes/index.html.erb
+++ b/app/views/notes/index.html.erb
@@ -48,7 +48,7 @@
       </td>
       <td><%= link_to note.id, note %></td>
       <td><%= note_author(note.author) %></td>
-      <td><%= note.comments.first.body.to_html %></td>
+      <td><%= note.description.to_html %></td>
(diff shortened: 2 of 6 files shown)


nenad-vujicic:
I also added using of these new Note methods (in added commits). Also, eliminated logic from `.(j)builder` / `comments partial` mentioned [here](https://github.com/openstreetmap/openstreetmap-website/pull/5485#discussion_r1912484985). I wasn't sure about new version of:

`@notes = @notes.joins(:comments).where("to_tsvector('english', note_comments.body) @@ plainto_tsquery('english', ?)", params[:q]) if params[:q]`

in `app/controllers/api/notes_controller.rb`, so I thought to wait, with rewriting this, for creating new records in database (in PR2 & PR3).

app/models/note_comment.rb
```diff
@@ -36,6 +36,26 @@ class NoteComment < ApplicationRecord
   validates :event, :inclusion => %w[opened closed reopened commented hidden]
   validates :body, :length => { :maximum => 2000 }, :characters => true
 
+  # Return the note's description for first special comment (event is "opened")
+  # otherwise return the comment text
+  def tmp_body
```
tomhughes, line 41: What is the point of these two `tmp` methods? They just wind up circling back to the same comment so there's no need for them at this point.

tomhughes (changes requested):

app/models/note_comment.rb
nenad-vujicic: I was trying to improve 'direct' access to comment's body / author (where first comment is accessed), so when we do data-migration it selects note's description / author (for first special comment) if migration is performed, otherwise, if migration is not done, selects comment's body / author.

Another reason is to hide logic (and don't do unwrapping of if-statements) from .(j)builder / _comment.html.erb / .. files where first comments are also accessed (mentioned [here](https://github.com/openstreetmap/openstreetmap-website/pull/5485#discussion_r1912484985)).

Perhaps we can find some better name or remove if really not necessary?

nenad-vujicic: Removed problematic routines and their usages. Failing test is probably transient, it passes on my side locally.

tomhughes (approved):

PR #5495, opened 2025-01-13
Test anonymous notes counter

AntonKhorev:
Tests I used for #5468. They create 10 anonymous notes and check that the encouragement message appears, then log in or sign up and check that the message disappears.


nenad-vujicic (approved):

tomhughes:
Looks good to me, thanks.

PR #5503, opened 2025-01-16
Removed unused translations in preferences.show

nenad-vujicic:
After merging 967e6d4 preferences.show.{preferred_editor, preferred_languages} are not used anymore. PR removes these translation keys.


AntonKhorev:
Thanks, these translations are in activerecord attributes.

PR #5504, opened 2025-01-16
Fixes translation key in TermsController

nenad-vujicic:
PR updates translation key which holds URL for redirecting in TermsController. Probably forgotten to update in 606b5c1 where translation keys were changed.

### DIFF
--- a/app/controllers/accounts/terms_controller.rb
+++ b/app/controllers/accounts/terms_controller.rb
@@ -42,7 +42,7 @@ module Accounts
 
           redirect_to referer || edit_account_path
         elsif params[:decline]
-          redirect_to t("users.terms.declined"), :allow_other_host => true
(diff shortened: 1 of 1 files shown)


AntonKhorev:
Is this branch reachable now? We don't show the terms page to unregistered users after the signup redesign, therefore `current_user` should be set.

nenad-vujicic:
> Is this branch reachable now? We don't show the terms page to unregistered users after the signup redesign, therefore `current_user` should be set.

Well, I was wondering about this too, but left untouched because didn't fully understand it. Do we need to simply remove this branch (and translation key) or to refactor entire `TermsController#update` (I don't see it's called from anywhere)?

nenad-vujicic:
Closed in favor of #5506

PR #5505, opened 2025-01-16
Use class based map tiles dark mode

hlfan:
A cleaner, class-based variant of #5426, enabling moving more logic out of `layeradd` into `initialize`.

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -21,7 +21,7 @@ L.OSM.Map = L.Map.extend({
       if (layerDefinition.apiKeyId && !OSM[layerDefinition.apiKeyId]) continue;
 
       let layerConstructor = L.OSM.TileLayer;
-      const layerOptions = {};
+      const layerOptions = { className: "" };
 
       for (const [property, value] of Object.entries(layerDefinition)) {
         if (property === "credit") {
@@ -32,6 +32,11 @@ L.OSM.Map = L.Map.extend({
           layerOptions.apikey = OSM[value];
         } else if (property === "leafletOsmId") {
           layerConstructor = L.OSM[value];
+        } else if (property === "leafletOsmDarkId" && OSM.isDarkMap() && L.OSM[value]) {
+          layerConstructor = L.OSM[value];
+          layerOptions.className += " dark";
+        } else if (property === "filterClass" && OSM.isDarkMap()) {
+          layerOptions.className += " " + value;
         } else {
           layerOptions[property] = value;
         }
@@ -52,10 +57,12 @@ L.OSM.Map = L.Map.extend({
       code: "G"
     });
 
-    this.on("layeradd", function (event) {
-      if (this.baseLayers.indexOf(event.layer) >= 0) {
-        this.setMaxZoom(event.layer.options.maxZoom);
-      }
+    this.on("layeradd", function ({ layer }) {
+      if (this.baseLayers.indexOf(layer) < 0) return;
+      this.setMaxZoom(layer.options.maxZoom);
+      const key = document.querySelector(".key-ui");
+      if (!key) return;
+      key.className = "key-ui " + layer.options.className;
     });
 
     function makeAttribution(credit) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/leaflet.map.js
```diff
@@ -32,6 +32,11 @@ L.OSM.Map = L.Map.extend({
           layerOptions.apikey = OSM[value];
         } else if (property === "leafletOsmId") {
           layerConstructor = L.OSM[value];
+        } else if (property === "leafletOsmDarkId" && OSM.isDarkMap() && L.OSM[value]) {
+          layerConstructor = L.OSM[value];
+          layerOptions.className += " dark";
+        } else if (property === "filterClass" && OSM.isDarkMap()) {
```
tomhughes, line 38: I think something like `darkFilterClass` would be better as the property name, so that it's clear what the filtering is for.

app/assets/javascripts/leaflet.map.js
```diff
@@ -52,10 +57,12 @@ L.OSM.Map = L.Map.extend({
...
     });
 
-    this.on("layeradd", function (event) {
-      if (this.baseLayers.indexOf(event.layer) >= 0) {
-        this.setMaxZoom(event.layer.options.maxZoom);
-      }
+    this.on("layeradd", function ({ layer }) {
+      if (this.baseLayers.indexOf(layer) < 0) return;
+      this.setMaxZoom(layer.options.maxZoom);
+      const key = document.querySelector(".key-ui");
```
tomhughes, line 63: We normally use jQuery rather than native methods like this.

app/assets/javascripts/leaflet.map.js
```diff
@@ -52,10 +57,12 @@ L.OSM.Map = L.Map.extend({
...
+      if (!key) return;
+      key.className = "key-ui " + layer.options.className;
```
tomhughes, line 65: ...and if jQuery is used then this can use `addClass` and not have to worry about preserving pre-existing classes.

config/layers.yml
```diff
@@ -31,6 +31,7 @@
   apiKeyId: "THUNDERFOREST_KEY"
   canEmbed: true
   canDownloadImage: true
+  filterClass: "no-filter"
```
tomhughes, line 34: I assumed that the filter class was going to be for turning on filtering, not for turning it off? My understanding from #5426 is that enabling filtering for a layer will only happen on the basis of agreement with the designer of that layer.

app/assets/stylesheets/common.scss
```diff
@@ -516,13 +515,13 @@ body.small-nav {
...
+  .leaflet-layer,
   .mapkey-table-entry td:first-child > * {
-    filter: var(--dark-mode-map-filter);
+    filter: var(--dark-mode-map-filter, none);
   }
 
-  .leaflet-tile-container .leaflet-tile {
-    filter: none;
+  .no-filter, .dark {
+    --dark-mode-map-filter: none;
```
tomhughes, line 524: This all seems very confused - you've removed the default value for the filter variable, and set a default of `none` when using it, so why do we need to explicitly set it to none here?

app/assets/stylesheets/common.scss
```diff
@@ -516,13 +515,13 @@ body.small-nav {
 }
 
 @mixin dark-map-color-scheme {
-  .leaflet-tile-container,
+  .leaflet-layer,
```
tomhughes, line 518: Why the change in class here?

tomhughes (changes requested):

app/assets/javascripts/leaflet.map.js
hlfan: I tried that multiple ways, but jQuery-based variants always errored out, so I just left it this way.

app/assets/javascripts/leaflet.map.js
hlfan: I also have to remove the classes from the previously selected layer. And I'd rather do that in a single line without needing a separate function on layerremove.

config/layers.yml
hlfan: I appeal that as long as there isn't documentation on which layers should get which treatment in dark mode, explicitly specifying at least in one file that this layer already has had a review of how it should be displayed in dark mode is critical for maintainability and avoiding unnecessary re-reviews.

app/assets/javascripts/leaflet.map.js
hlfan: I've looked further into it, this is on me for having minified leaflet overwrite the `$` function.
This works:
```js
jQuery(".key-ui").attr("class", "key-ui " + layer.options.className);
```

tomhughes:
Superseded by #5507.

PR #5506, opened 2025-01-16
Remove account terms code unreachable for unregistered users

AntonKhorev:
After the terms page was taken out of the signup process, its controller still had the code for this kind of use by unregistered users. Later I moved the page to `/account/terms` which implies that its only for active users, but I still haven't removed the code.

Here I'm actually removing the unused code. Things like `redirect_to login_path(:referer => request.fullpath) if current_user.nil?` are handled by abilities, only active users are able to access the terms.


tomhughes:
Looks good to me, thanks.

PR #5507, opened 2025-01-16
Use dark map tiles classlessly

hlfan:
Barebones classless filterless version of #5505.

### DIFF
--- a/config/layers.yml
+++ b/config/layers.yml
@@ -39,6 +39,7 @@
         href: "https://www.thunderforest.com/"
 
 - leafletOsmId: "TransportMap"
+  leafletOsmDarkId: "TransportDarkMap"
   code: "T"
   layerId: "transportmap"
   nameId: "transport_map"


app/assets/stylesheets/common.scss
```diff
@@ -9,7 +9,7 @@
 
 body {
   font-size: $typeheight;
-  --dark-mode-map-filter: brightness(.8);
+  --dark-mode-map-filter: none;
```
HolgerJeromin, line 12: Is this needed?
```suggestion
```

hlfan: Having the custom property defined with a default value of `none` increases compatibility, since it doesn't even alter transitions into and out of undefined anymore.
It also increases the probability of no filter being applied for users using custom styles that may conflict with the TransportDarkMap.

tomhughes:
I've tweaked this a bit to remove the unused code for adding a class and reworded the commit messages but other than that I think this looks good now. Thanks for all your work on the different versions.

hlfan:
> Thanks for all your work on the different versions.

Thank _you_ for being the necessary devil's advocate and pushing me to ditch the futureproofing attempts.

PR #5509, opened 2025-01-17
Use form with normal-sized button on follows show pages

AntonKhorev:
Originally the make/remove friend pages contained a form:
https://github.com/openstreetmap/openstreetmap-website/blob/17323257685bf5134d9286f5baefaddcae65478c/app/views/friendships/make_friend.html.erb#L5-L10

#5261 replaces these two pages with one `/user/:user/follow` page. That have doesn't have a form. Instead it has a link styled as a button. This is probably because there were multiple changes to paths during the development of that PR. But I don't know why it shouldn't be a true button with the paths we eventually settled on. Also the link has a small button css class on it. I don't know why the button that is the only thing on the page should be small. Probably it was a copypasted code from subscribe/unsubscribe buttons or something similar.

Before:
image

After:
image

### DIFF
--- a/app/views/follows/show.html.erb
+++ b/app/views/follows/show.html.erb
@@ -2,7 +2,9 @@
   <h1><%= t(@already_follows ? ".unfollow.heading" : ".follow.heading", :user => @friend.display_name) %></h1>
 <% end %>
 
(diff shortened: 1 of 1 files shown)


gravitystorm:
Merged, thanks!

PR #5511, opened 2025-01-17
Adds optional use of notes records

nenad-vujicic:
### Description

PR adds use of `description`, `user_id`, `user_ip` from `notes` table if data-migration is done, otherwise use `body`, `author_id`, `author_ip` from note's first comment. Decision procedure "data-migration is done" is defined with "is user_ip or user_id not nil?".

This PR is 4th from set of PRs described [here](https://github.com/openstreetmap/openstreetmap-website/pull/5485#pullrequestreview-2545467592).

### How has this been tested?

Automated unit tests and manual testing.

### DIFF
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -28,6 +28,8 @@
 class Note < ApplicationRecord
   include GeoRecord
 
+  belongs_to :author, :class_name => "User", :foreign_key => "user_id", :optional => true
+
   has_many :comments, -> { left_joins(:author).where(:visible => true, :users => { :status => [nil, "active", "confirmed"] }).order(:created_at) }, :class_name => "NoteComment", :foreign_key => :note_id
   has_many :all_comments, -> { left_joins(:author).order(:created_at) }, :class_name => "NoteComment", :foreign_key => :note_id, :inverse_of => :note
   has_many :subscriptions, :class_name => "NoteSubscription"
@@ -89,14 +91,28 @@ class Note < ApplicationRecord
     closed_at + DEFAULT_FRESHLY_CLOSED_LIMIT
   end
 
-  # Return the note's description, derived from the first comment
-  def description
-    comments.first.body
+  def author_deleted?
+    !author.nil? && author.status == "deleted"
   end
 
-  # Return the note's author object, derived from the first comment
+  # Return the note's description, unless record is unavailable and
+  # it will be derived from the first comment
+  def description
+    if user_ip.nil? && user_id.nil?
+      all_comments.first.body
+    else
+      RichText.new("text", super)
+    end
+  end
+
+  # Return the note's author object, unless record is unavailable and
+  # it will be derived from the first comment
   def author
-    comments.first.author
+    if user_ip.nil? && user_id.nil?
+      all_comments.first.author
+    else
+      super
+    end
   end
 
   private
--- a/app/assets/javascripts/index/layers/notes.js
+++ b/app/assets/javascripts/index/layers/notes.js
@@ -44,10 +44,13 @@ OSM.initializeNotesLayer = function (map) {
       marker.setIcon(noteIcons[feature.properties.status]);
     } else {
       let title;
-      const description = feature.properties.comments[0];
 
-      if (description?.action === "opened") {
-        title = description.text;
+      if (feature.properties.comments.length > 0) {
+        const description = feature.properties.comments[0];
+
+        if (description?.action === "opened") {
+          title = description.text;
+        }
       }
 
       marker = L.marker(feature.geometry.coordinates.reverse(), {
--- a/app/helpers/note_helper.rb
+++ b/app/helpers/note_helper.rb
@@ -1,6 +1,14 @@
 module NoteHelper
   include ActionView::Helpers::TranslationHelper
 
+  def note_description(author, description)
+    if !author.nil? && author.status == "deleted"
+      RichText.new("text", t("notes.show.description_when_author_is_deleted"))
+    else
+      description
+    end
+  end
+
   def note_event(event, at, by)
     if by.nil?
       t("notes.show.event_#{event}_by_anonymous_html",
--- a/app/views/api/notes/_note.rss.builder
+++ b/app/views/api/notes/_note.rss.builder
@@ -13,7 +13,7 @@ xml.item do
   xml.guid api_note_url(note)
   xml.description render(:partial => "description", :object => note, :formats => [:html])
 
-  xml.dc :creator, note.author.display_name if note.author
+  xml.dc :creator, note.author.display_name unless note.author.nil? || note.author_deleted?
 
   xml.pubDate note.created_at.to_fs(:rfc822)
   xml.geo :lat, note.lat
--- a/app/views/notes/index.html.erb
+++ b/app/views/notes/index.html.erb
@@ -48,7 +48,7 @@
       </td>
       <td><%= link_to note.id, note %></td>
       <td><%= note_author(note.author) %></td>
-      <td><%= note.description.to_html %></td>
+      <td><%= note_description(note.author, note.description).to_html %></td>
       <td><%= friendly_date_ago(note.created_at) %></td>
       <td><%= friendly_date_ago(note.updated_at) %></td>
     </tr>
--- a/app/views/notes/show.html.erb
+++ b/app/views/notes/show.html.erb
@@ -52,10 +52,10 @@
     <% end %>
   </div>
 
-  <% if @note_comments.length > 1 %>
+  <% if @note_comments.length > (@note.author_deleted? ? 0 : 1) %>
     <div class='note-comments'>
       <ul class="list-unstyled">
-        <% @note_comments.drop(1).each do |comment| %>
+        <% @note_comments.drop(@note.author_deleted? ? 0 : 1).each do |comment| %>
           <li id="c<%= comment.id %>">
             <small class='text-body-secondary'><%= note_event(comment.event, comment.created_at, comment.author) %></small>
             <div class="mx-2">
--- a/app/views/notes/show.html.erb
+++ b/app/views/notes/show.html.erb
@@ -5,7 +5,7 @@
 <div>
   <h4><%= t(".description") %></h4>
   <div class="overflow-hidden ms-2">
-    <%= h(@note.description.to_html) %>
+    <%= h(note_description(@note.author, @note.description).to_html) %>
   </div>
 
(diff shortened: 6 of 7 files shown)


tomhughes (commented):
Why does this need to use `[:user_id]` etc rather than the more typical `.user_id` form?

nenad-vujicic:
> Why does this need to use `[:user_id]` etc rather than the more typical `.user_id` form?

I replaced using `[:user_id]` with `user_id` (same with user_ip), but had to leave self[:description] because of possible recursion. I couldn't use self.{user_id, user_ip, description} because of Rubocop warnings.

app/models/note.rb
```diff
@@ -89,24 +89,39 @@ def freshly_closed_until
...
+  # Return the note's description, unless record is unavailable and
+  # it will be derived from the first comment
   def description
-    comments.first.body
+    if user_ip.nil? && user_id.nil?
+      comments.first.body
+    else
+      RichText.new("text", self[:description])
+    end
   end
```
AntonKhorev, line 100: I assume that `self[:description]` is going to be the true note description, the text note was created with. `comments.first.body` is not necessarily that text because `comments.first` is not necessarily the first api 0.6 comment. `comments` only contain api 0.6 comments by active users and doesn't contain api 0.6 comments of deleted users.

app/models/note.rb
```diff
@@ -89,24 +89,39 @@ def freshly_closed_until
...
+  # Return the note's author ID, unless record is unavailable and
+  # it will be derived from the first comment
   def author_id
-    comments.first.author_id
+    if user_ip.nil? && user_id.nil?
+      comments.first.author_id
+    else
+      user_id
+    end
   end
```
AntonKhorev, line 115: If user_id is a true id of the user who created the note, it won't necessarily match `author.id`, see my comment for `description`. Attempts to fix that for `author` were made in the past but were rejected by the maintainers, see for example #3607.

app/models/note.rb
tomhughes: We could make it consistent by using `all_comments` instead but that obviously changes the behaviour from what we have now, but then that's going to be the case anyway when this transition is completed?

AntonKhorev: Weren't you insisting on not showing the true description until LWG agrees?

tomhughes: I don't think I've said that as such.

What I have said is that we need a proper documented policy about how to handle deleted users and what we should and shouldn't show. I don't think I specifically asked for that to come from LWG though obviously that may be needed to answer questions about what we can and can't do.

app/models/note.rb
nenad-vujicic: The latest push should fix this. Thanks!

app/models/note.rb
nenad-vujicic: Have we reached a decision on this? We planned to copy the body of the first note's comment (i.e. one with event set to "opened") to note's description ([data migration-script](https://github.com/nenad-vujicic/openstreetmap-website/commit/0cc2d59311479c56e009e0e63b4bdb54113aeb61)), but can easily switch to another logic?

Btw, before data-migration PR we planned 2 more steps ([updating creating](https://github.com/nenad-vujicic/openstreetmap-website/commit/8df2f38e21c0703dfb49267d2b6a3f9bc808a63b) and [updating searching](https://github.com/nenad-vujicic/openstreetmap-website/commit/75212422a99134aa0b67fac4f7c1533f9ffba15d)).

AntonKhorev: Copying is not the problem, the problem is you won't be able to the description if the user who created it is deleted. You'll have to remember that whenever you access `note.description`.

There was a draft policy by DWG about deleted users that was supposed to be sent to LWG but wasn't. It says that DWG recommends note comments of deleted users to be visible. If that's impossible, "a comment is not shown because ..." is to be displayed instead. Descriptions are not mentioned but the same should apply to them.

The default solution if no decision was made was to replace description with "deleted" when the note is displayed:

https://github.com/openstreetmap/openstreetmap-website/pull/4481/files#diff-74d80704129ce910bd38a56b250122e716028ac0085b3627ea8875b9216b65ceR25-R29

Since you already started using `note.description`, one of the following things should happen:
- you remove deleted descriptions here in the model
- you remove them on use
- we actually do a *proper documented policy* (in other words it's easier to remove deleted descriptions for now)

app/models/note.rb
AntonKhorev: Now you are revealing the true author as soon as migration happens, but not before.

app/models/note.rb
```diff
@@ -89,24 +91,44 @@ def freshly_closed_until
...
 
-  # Return the note's author IP address, derived from the first comment
+  # Return the note's author IP address, unless record is unavailable and
+  # it will be derived from the first comment
   def author_ip
```
AntonKhorev, line 126: Do we need this method at all?

nenad-vujicic: Thanks! Removed both `author_ip` and `author_id` methods since they are not used anywhere.

app/models/note.rb
nenad-vujicic: Thanks! Started using `all_comments` instead of `comments`.

app/models/note.rb
nenad-vujicic: Thanks! Started displaying "deleted" if note's author is deleted.

app/models/note.rb
```diff
@@ -89,24 +91,27 @@ def freshly_closed_until
...
-
-  # Return the note's author IP address, derived from the first comment
-  def author_ip
-    comments.first.author_ip
+    if user_ip.nil? && user_id.nil?
+      all_comments.first.author
+    else
+      self[:author]
+    end
   end
```
AntonKhorev, line 115: If `author` returns the true author, `description` should also return the true description. There's `note_author` helper which hides the author when necessary. Similarly you can make a helper that hides the description, and you won't need `I18n.t("notes.show.description_when_author_is_deleted")` inside the model. And you'll use the helper in these places: https://github.com/openstreetmap/openstreetmap-website/pull/5494/commits/2c19c2143ec51e507d19b41fb6c89454b3d83dfc.

In #3607 I also [removed the deleted author]( https://github.com/openstreetmap/openstreetmap-website/pull/3607/files#diff-9b8b3e6d46a2a808a27e4c2f8d88c5470d39ee77146f8bb530124c0cacc35dbeR16) from rss.

nenad-vujicic: I've just pushed the latest changes without solving [above](https://github.com/openstreetmap/openstreetmap-website/pull/5511#discussion_r1927277396).

nenad-vujicic: Added 2 commits solving [above two problems](https://github.com/openstreetmap/openstreetmap-website/pull/5511#discussion_r1927277396).

nenad-vujicic:
@AntonKhorev @tomhughes  do you have any further comment / request on this PR?

Thanks!

app/models/note.rb
```diff
@@ -89,24 +91,18 @@ def freshly_closed_until
...
-    comments.first.author_id
-  end
-
-  # Return the note's author IP address, derived from the first comment
-  def author_ip
-    comments.first.author_ip
+    if user_ip.nil? && user_id.nil?
+      all_comments.first.author
+    else
+      association(:author).reader
```
AntonKhorev, line 104: Is this going to work?

```suggestion
      super
```

`association` is [not documented](https://github.com/rails/rails/blob/cf6ff17e9a3c6c1139040b519a341f55f0be16cf/activerecord/lib/active_record/associations.rb#L51).

app/models/note.rb
```diff
@@ -95,11 +95,6 @@ def author_deleted?
     !author.nil? && author.status == "deleted"
   end
 
-  # Return the note's description, derived from the first comment
-  def description
-    comments.first.body
-  end
```
AntonKhorev: Don't you use this method in the helper before this commit and the helper assumes it returns the migrated description?

test/models/note_test.rb
```diff
@@ -47,15 +47,6 @@ def test_closed?
...
   end
 
-  def test_description
-    comment = create(:note_comment)
-    assert_equal comment.body, comment.note.description
-
-    user = create(:user)
-    comment = create(:note_comment, :author => user)
-    assert_equal comment.body, comment.note.description
-  end
```
AntonKhorev, line 57: You dropped `note.description` a few commits before. Why did you keep the tests until this commit?

+ same for `note.author_id` and `author_ip`

app/views/api/notes/_note.json.jbuilder
```diff
@@ -7,6 +7,7 @@ end
 
 json.properties do
   json.id note.id
+  json.description note_description(note)
```
AntonKhorev, line 10: Why are we changing the API? Did anyone suggested changing the API?

AntonKhorev:
The commits here are doing things in a strange order. One way to untangle them is to make another pull request that drops author_id/_ip without doing other things.

app/views/api/notes/_note.json.jbuilder
AntonKhorev: Let's suppose you have a normal note. No deleted users touched it. From the API point of view its description is in the first comment. If you add a description like here, you'll have it twice in the API output. You can't remove it from the first comment because it's not backwards compatible. Does it make sense to have it twice? And that's not the only way to fix the tooltip problem. You could instead add some attribute to an API comment that indicates if it's a description or not. In other words, you shouldn't do this here because it's a completely different problem.

nenad-vujicic:
> The commits here are doing things in a strange order. One way to untangle them is to make another pull request that drops author_id/_ip without doing other things.

Added #5568 

The problem with the current solution is JS is breaking (displaying only a subset of notes from bounding box) when trying to display note without visible comments (e.g. we create new user, login as that user, create new note without further commenting / making other actions, delete user, navigate to area where previously created note is located). This bug is solved by #3617 but it's still not merged. Shall we wait #3617 to be merged or add its solution to this PR or deal with these problematic notes in some other way?

app/models/note.rb
nenad-vujicic: Thanks, fixed!

app/models/note.rb
nenad-vujicic: Thanks, fixed.

app/views/api/notes/_note.json.jbuilder
nenad-vujicic: Thanks, postponed to some future PR.

nenad-vujicic:
Closed in favor of #5568, #5579 and #5609.

PR #5513, opened 2025-01-18
Remove "Public editing" line from settings page

AntonKhorev:
There's the *Public editing* line on the `/account/edit` page between *External Authentication* and *Contributor Terms*. I don't think it's useful anymore. All of the accounts created in the last 15 years have public editing enabled. All new users are invited to start editing on the welcome page, so they know they can edit. All users who have been around for a while know they can edit. They don't need to be told that they can edit on the *Settings* page. This PR removes the *Public editing* line.

Before:
image

After:
image

---

For a few users that still have an account without public editing enabled, this line is useless anyway because there's an entire section at the bottom of the page about public editing.

Before when not gone public:
image

After when not gone public:
image

### DIFF
--- a/app/views/accounts/edit.html.erb
+++ b/app/views/accounts/edit.html.erb
@@ -30,19 +30,6 @@
   </fieldset>
 
   <div class="mb-3">
-    <label class="form-label"><%= t ".public editing.heading" %></label>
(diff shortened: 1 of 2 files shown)


tomhughes:

PR #5514, opened 2025-01-18
Remove public domain checkbox from signup and terms pages

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/4455#issuecomment-1899125291, https://github.com/openstreetmap/openstreetmap-website/pull/4455#discussion_r1509952021, https://github.com/openstreetmap/openstreetmap-website/pull/4455#issuecomment-2069941509 for reasons to remove the checkbox from the signup page.

The checkbox is confusingly placed right above the *Sign Up* button:
image
Users might think they have to check it in order to proceed, but they actually don't have to do it. If they check it accidentally and register an account, there's no way to undo it.

I also removed a similar checkbox from the terms page. That page is currently not accessible for new accounts because terms are accepted by signing up. But that's probably going to change because the terms page could be reused for accepting updated terms. Supposedly terms of use were updated and many users need to accept the new terms if GDPR changes happen. If there's a  public domain checkbox on that page, users also might also accidentally check it.

If anyone actually want to declare their contributions in public domain, I added a special page for that. There's a link to it from *Contributor Terms* on the *My Settings* page. I didn't make the link very noticeable because I don't expect many people to visit that page:
image

The page is at `/account/pd_declaration` and is just a checkbox and a confirm button:
image

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -38,6 +38,7 @@ class Ability
         can [:read, :create, :destroy], :oauth2_authorization
         can [:update, :destroy], :account
         can :update, :account_terms
+        can :create, :account_pd_declaration
         can :read, :dashboard
         can [:create, :subscribe, :unsubscribe], DiaryEntry
         can :update, DiaryEntry, :user => user
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -29,7 +29,7 @@ class Ability
 
     if user&.active?
       can :welcome, :site
-      can :read, [:deletion, :account_terms]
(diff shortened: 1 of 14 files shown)


simonpoole:
You shouldn't be expending any effort on retaining the PD checkbox imho. It has no practical meaning and should have been removed years ago.

gravitystorm:
Merged, thanks @AntonKhorev 

I also agree with @simonpoole, and I would be happy to remove the check box entirely and drop the database column too. However, I feel that this would be a topic that would benefit from wider feedback (e.g. from OSMF) so for now we can keep it just in case.

PR #5515, opened 2025-01-18
Replace lookup_friend with lookup_user in FollowsController

AntonKhorev:
More *friend* cleanups. Removes `lookup_friend` method that sets `@friend` variable, replacing it with `lookup_user` that sets `@user`. `@user` will allow to treat `follow` as a nested resource of a user.


tomhughes:
Looks good to me, thanks.

PR #5516, opened 2025-01-18
Use resourceful routes for api tracepoints

AntonKhorev:
Doesn't do much, mostly changes `trackpoints_path` to `api_tracepoints_path`, with `api_` prefix like other api paths.


tomhughes:
Looks good to me, thanks.

PR #5517, opened 2025-01-18
Use resourceful routes for traces feeds

AntonKhorev:
Moves `georss` action from traces controller into its own feeds controller.


tomhughes:
Looks good to me, thanks.

PR #5518, opened 2025-01-19
List of user's changeset comments

AntonKhorev:
A new version of #4248.

The main difference is that I'm creating another set of controllers inside the `Users` module for diary, changeset and maybe later note comments. The already existing [diary comments controller](https://github.com/openstreetmap/openstreetmap-website/blob/2074e9dce23e1e6380c6c995a79a76671323c905/app/controllers/diary_comments_controller.rb) has handles both comments belonging to a diary entry and comments belonging to a user. The `index` action is for users' comments, the rest is for comments of diary entries. This is not very convenient. I'm moving `index` to a new controller in the `Users` module and I'm creating a similar controller for changeset comments.

image


tomhughes:
This all looks great to me, thanks.

We probably want to think about indexes because changeset comments don't have an index on author+id (neither do diary comments but some people have 80k changeset comments vs a few hundred diary comments) but I don't think that's a blocker based on some quick tests.

PR #5519, opened 2025-01-19
Use dark map tiles in embed

hlfan:
This PR
- Adds layer type switching based on a **new theme parameter** and prefers-color-scheme
- Rewrites argument extraction and application of layer and bbox
- Lints this file apparently for the first time

Doesn't touch the CSS though.


tomhughes (commented):
Please try to avoid including stylistic changes like replacing single quotes with double quotes in the same commit as functional changes as it makes it very hard to tell what the actual functional changes are when reviewing...

hlfan:
Just doing the same linting as usual, why is this not wanted now @tomhughes?

tomhughes:
Firstly I didn't say it wasn't wanted, I said it is better to keep it in a separate commit to make reviewing easier.

Secondly what "same linting as usual" is this? As far as I know it was already passing our eslint rules so is this some additional linting of your own?

hlfan:
So, when I tested the changes locally, I edited the evaluated JS to make iteration faster for me. Then the linter got mad. But eslint alone doesn't check files that have embedded ruby, which I'll criticise in an issue later on.
As I don't like touching and "un-touching" code, I guess I'll keep making PRs.

hlfan:
See also #5523

Issue #4755, opened 2024-05-05
Centering lost after search on Android


BertMule:
### How to reproduce the issue?

If one searches on Android in the website,  for like a town, one gets the town centred in view.
There is a large overlay, by which one can even see its boundaries.

But when one closes that overlay by the cross, the centring is lost by something like 10 kilometres to the south-west!
I don't even understand what and why this happens.
Anyway, this is completely annoying , as one has to zoom out, and manually look for the town one already had in view!

So please recentre to the found item after closing the overlay.

By the way, showing search history is more useful to the average user than mapping history, which is present by a large button right under the search entry. Confusing too.

tomhughes:
There are many existing complaints about the map moving when the sidebar is opened/closed so this is probably a duplicate but the code is basically incomprehensible and nobody dares touch it I'm afraid :-(

BertMule:
Not surprising, as people should have noticed already.

But this basic (malfunctioning) functionality should be fixed.

tomhughes:
Patches welcome.

Issue #991, opened 2015-06-12
Map shifts position incorrectly on mobile view


erjiang:
Steps to reproduce:

1. Open openstreetmap.org on iPhone.
2. Zoom in on a particular point or feature.
3. Tap add-note button.
4. Map and marker center on a new location to the right (east) of the original location.

I guess there's some code that offsets the map to deal with the left-hand pane, but the pane is on the top in the mobile view, not the left.

Aury88:
I confirm the wrong behavior on smartphone (on pc the note sidepanel doesn't move the map, it "only" put the note locator in the center of the narrower map, more at east, that is still a wrong behavior in my pov )
imho the map shouldn't move when the panel is opened ... if someone want to put a note probably he want to put it in a place near the centre of the visualized map so the map should remain centered in that place when the side/top panel is opened.

dieterdreist:
I can confirm that this is still the case. It is not limited to mobile and it happens when you close the left sidebar (map jumps to the left). It doesn't happen at widths of 1200px or more (desktop / scale=1), but it happens when the screen is narrower than that.

On small screens (sidebar from above rather than left), the map jumps to the left + down so that the top border of the visible map remains at the same point, but the left border and right border is shifted to the right, effectively panning the map to the left. While there are probably good intentions behind this, it leads to a very confusing result if you were focussing on a part that is now off screen.
On closing of the sidebar it does the same movement in the opposite direction.
Maybe the vertical movement is happening automatically (as a result of reducing the map height when the sidebar pops up from above).

Here some screenshots to illustrate the problem:
![Screenshot 2019-11-18 at 22 57 05](https://user-images.githubusercontent.com/581288/69097516-fe694b00-0a56-11ea-9b56-bc9edf603c34.png)
![Screenshot 2019-11-18 at 22 57 25](https://user-images.githubusercontent.com/581288/69097518-ff01e180-0a56-11ea-9491-252e34094856.png)
![Screenshot 2019-11-18 at 22 57 38](https://user-images.githubusercontent.com/581288/69097519-ff01e180-0a56-11ea-8c7c-8d2e8460d199.png)
![Screenshot 2019-11-18 at 22 57 51](https://user-images.githubusercontent.com/581288/69097521-ff9a7800-0a56-11ea-94de-934cebc06eed.png)

SomeoneElseOSM:
> I can confirm that this is still the case. 

[...]

PR #5521, opened 2025-01-19
Fix left/right sidebar resizing on mobile devices

deevroman:
This PR fixes multiple issues about the map resetting its position when opening various menus on mobile devices.

https://github.com/openstreetmap/openstreetmap-website/issues/4755
https://github.com/openstreetmap/openstreetmap-website/issues/991 (partly because the sidebar can still cover the selected object in the Map data layer)
...

1. Get rid of the hardcoded left sidebar width.
2. On mobile devices, use adjust the height, not the width of the map

### How has this been tested?

<table>

<tr> 

<th>

Before

https://github.com/user-attachments/assets/9c283266-b43a-48d2-9f25-ffe73752c837

<tr> 

<th>

After

https://github.com/user-attachments/assets/9cf73154-c043-430a-a143-f8d47ad0f080

<th>

After with RTL

https://github.com/user-attachments/assets/ed973523-98d2-46a8-be2b-ca8619e1085f

</tr> 

</table>

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -347,15 +347,22 @@ L.OSM.Map = L.Map.extend({
   },
 
   setSidebarOverlaid: function (overlaid) {
-    var sidebarWidth = 350;
+    var mediumDeviceWidth = window.getComputedStyle(document.documentElement).getPropertyValue("--bs-breakpoint-md");
+    var isMediumDevice = window.matchMedia(`(max-width: ${mediumDeviceWidth})`).matches;
+    var sidebarWidth = $("#sidebar").width();
+    var sidebarHeight = $("#sidebar").height();
     if (overlaid && !$("#content").hasClass("overlay-sidebar")) {
       $("#content").addClass("overlay-sidebar");
       this.invalidateSize({ pan: false });
-      if ($("html").attr("dir") !== "rtl") {
+      if (isMediumDevice) {
+        this.panBy([0, -sidebarHeight], { animate: false });
+      } else if ($("html").attr("dir") !== "rtl") {
         this.panBy([-sidebarWidth, 0], { animate: false });
       }
     } else if (!overlaid && $("#content").hasClass("overlay-sidebar")) {
-      if ($("html").attr("dir") !== "rtl") {
+      if (isMediumDevice) {
+        this.panBy([0, $("#map").height() / 2], { animate: false });
+      } else if ($("html").attr("dir") !== "rtl") {
         this.panBy([sidebarWidth, 0], { animate: false });
       }
       $("#content").removeClass("overlay-sidebar");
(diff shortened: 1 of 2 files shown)


deevroman:
I didn't take into account that sometimes centering is useful and necessary:

https://github.com/user-attachments/assets/2c222546-6754-4605-9591-d42df41e9ffb

tomhughes:
This PR is marked as draft - if it's ready for review can you clear the draft flag please?

deevroman:
Okay, it seems that this PR at least does not make it worse.

The only regression I noticed was a noticeable shift of the map with the note marker when opening the left and right sidebars at the same time. But I think this is a rare use case:


https://github.com/user-attachments/assets/b758e136-bfa6-4689-ba69-bfdd96412094

It is also worth noting that sometimes you can notice a map shift by one pixel due to not accurately calculating the map position using`. height() / 2` Maybe I'm being too careful because I use a Retina display.

app/assets/javascripts/leaflet.map.js
```diff
@@ -347,15 +347,22 @@ L.OSM.Map = L.Map.extend({
   },
 
   setSidebarOverlaid: function (overlaid) {
-    var sidebarWidth = 350;
+    var mobileDeviceWidth = window.getComputedStyle(document.documentElement).getPropertyValue("--bs-breakpoint-md");
+    var isMobileDevice = window.matchMedia(`(max-width: ${mobileDeviceWidth})`).matches;
```
tomhughes, line 351: I wonder if `mobileDevice` is the right way to describe this - what we're really looking at is  which side of bootstrap's medium breakpoint we are which in turn controls whether "sidebars" go at the side or the top. Maybe @gravitystorm has a thought?

Also I think technically to match https://getbootstrap.com/docs/5.0/layout/breakpoints/#max-width we should subtract `.02` from max-width in the match?

tomhughes (commented):
This mostly looks fine for me and seems to work well, or at least much better than what we have now! I've just added one inline comment.

app/assets/javascripts/leaflet.map.js
gravitystorm: > Maybe @gravitystorm has a thought?

In general we must avoid referring to narrow screens as "mobile devices" - that causes endless confusion among our developers. Mobile devices can have large screen widths, desktop windows can become narrow, etc.

With bootstrap the `md` means "medium" not "mobile device" - perhaps this is the source of the confusion here?

https://getbootstrap.com/docs/5.3/layout/breakpoints/#available-breakpoints

tomhughes (approved):
Thanks for the fix - this looks good to me now.

AntonKhorev:
Left sidebar used to cover the entire screen in the small layout. Later a fix which I can't find was applied to make it cover only a half. This PR probably fixes that fix.

tomhughes:
It still seems to be restricted to half the screen for me:

image

PR #5524, opened 2025-01-20
Add a page to view active blocks

AntonKhorev:
This PR adds `/account/blocks` page that does the following:
- if there are needs_view blocks, redirects to the earliest one of them
- if there are any active blocks, redirects to the earliest one of them
- if there are no active blocks, displays a message that there aren't any blocks, possibly with a link to `/user/:username/blocks` for reviewing past blocks

image

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -38,6 +38,7 @@ class Ability
         can [:read, :create, :destroy], :oauth2_authorization
         can [:update, :destroy], :account
         can :update, :account_terms
+        can :read, :account_block
         can :read, :dashboard
         can [:create, :subscribe, :unsubscribe], DiaryEntry
         can :update, DiaryEntry, :user => user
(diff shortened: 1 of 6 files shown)


tomhughes:
I think this looks fine from a code point of view - my only question is how we use this without effectively turning this URL into API that is baked into clients.

Is you plan that we include this URL in the error the API returns, with the intention that clients will display that error to the user in a way that makes the link clickable?

AntonKhorev:
I'm thinking if it's important to handle the situation where you're blocked while using a token from one account but logged into osm-website with another account. If it is important, sending blocked users to `/account/blocks` from apps won't quite work because that page wouldn't know about their token account.

tomhughes:
So are you saying this doesn't work as is then? If it's not capable of doing those things? Sounds like we would need it to include the user so we could force a reauthentication if necessary?

PR #5527, opened 2025-01-20
Enable I18n fallback for the embed page

### LINKED ISSUE #5526: Embedded map - missing cs-CZ translation
### URL

https://www.openstreetmap.org/export/embed.html

Visit https://www.openstreetmap.org/export/embed.html on mobile. Only mobile browsers are affected.

Instead of _"Nahlásit problém"_ there is _"[missing "cs-CZ.javascripts.embed.report_problem" translation]"_.

Replicated with
Firefox 135; Chrome 131
Android 15

Image

tomhughes:
Enable I18n fallback for the embed page to fix #5526.

### DIFF
--- a/app/assets/javascripts/embed.js.erb
+++ b/app/assets/javascripts/embed.js.erb
@@ -11,6 +11,9 @@ if (navigator.languages) {
   I18n.locale = navigator.language;
 }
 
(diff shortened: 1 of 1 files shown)


AntonKhorev:
Merged, thanks.

Issue #3971, opened 2023-03-14
Show notes more often


Cj-Malone:
### Problem

I often have to re enable the notes layer, it seems to reset to be disabled quite often. And because I often forget to re enable it I solve less notes.

### Description

I think notes should be a more sticky setting, or potentially **enabled by default**.

tomhughes:
It's already included in the layers cookie so I don't know how we could make it more sticky.

Personally I have the opposite problem to you, that I often come back to find it's on for some reason and have to turn it off.

Certainly we're not going to be making it the default, both because it's pretty annoying and because of the load it would likely put on the server.

Unless you can identify some specific case where it gets disabled when you don't think it should I don't think there's much we can do - as it stands this report is too vague to be actionable.

Cj-Malone:
> It's already included in the layers cookie

I don't have an issue staying signed in. Do the different cookies have different TTL so something?

> so I don't know how we could make it more sticky

A user pref would be more sticky than a cookie, but that obviously has issues too.

As always, thanks for the discussion.

tomhughes:
The cookie (the `_osm_location` one) has a ten year lifetime while logins are only valid for as long as the browser is open.

Cj-Malone:
I just restarted Firefox twice, I stayed signed in, and notes stayed enabled so there must be a difference between the docs and the reality of that "as long as the browser is open".

So that `_osm_location` contains both the location and the open layers? And that gets overridden when I navigate to a new URL with a `#`? I think I've found the issue.

I often use OSM Smart Menu to switch between OSM, Mapillary, FHDOT and Survey Me. That sets the location in the url, but not any layers that were active.

tomhughes:
Yes although cookies without a timeout are session cookies that traditionally vanish when you quit the browser many browsers these days have optional session preservation for at least some types of restart which confuses things.

The location cookie encodes both the location and the active layers yes - basically the state of your map view - so any time you visit the site it will be updated and if you visit a view that has the note layer disabled then that will update the location cookie to reflect that.

Cj-Malone:
Also it seems it's not just links with `#`, enable notes then visit https://www.openstreetmap.org/?mlat=53.32607&mlon=-1.49033 and it will also disable the notes layer.

For links that I control, is there a way for me to link to a specific location without enabling or disabling a users notes preference?

mmd-osm:
[...]

PR #5529, opened 2025-01-21
Consistent usage of layer information in _osm_location cookie

mmd-osm:
As reported in #3971, some URLs like https://www.openstreetmap.org/?mlat=53.32607&mlon=-1.49033 reset the currently selected layers. In order to restore the previous setting, a user needs to select both map and overlay layers after clicking on the link. 

It seems a bit premature to simply override the current layers with default values, in case the link does not include any layer information.

Instead, this PR unconditionally extracts the current layers from the _osm_location cookie, so it can be applied consistently across different url parsing alternatives.

As before, if the link includes layers in the hash value, that information takes precedence.

This change mainly impacts line 144 below:

```
mapParams.layers = hash.layers || (loc && loc[3]) || '';
```

loc[3] will now always contain  the selected layers, as long as the cookie is available.

### DIFF
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -80,7 +80,7 @@ OSM = {
   },
 
   mapParams: function (search) {
-    var params = OSM.params(search), mapParams = {}, loc, match;
+    var params = OSM.params(search), mapParams = {}, match;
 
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #5530, opened 2025-01-21
Preload parent objects for user comments

tomhughes:
I noticed that the comment views were loading objects one by one and this should make it more efficient.


AntonKhorev:
Merged, thanks.

PR #5531, opened 2025-01-21
Add indexes on user+id to changeset and diary comments

tomhughes:
This adds indexes on `user+id` to diary and changeset comments to make comment views more efficient.

Once this is merged we can switch the `diary_comments` association on users to sort by id instead of creation date and then drop the creation date indexes.

### DIFF
--- a/app/models/changeset_comment.rb
+++ b/app/models/changeset_comment.rb
@@ -12,6 +12,7 @@
 # Indexes
 #
 #  index_changeset_comments_on_author_id_and_created_at     (author_id,created_at)
+#  index_changeset_comments_on_author_id_and_id             (author_id,id)
 #  index_changeset_comments_on_changeset_id_and_created_at  (changeset_id,created_at)
 #  index_changeset_comments_on_created_at                   (created_at)
 #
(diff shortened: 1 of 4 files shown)


AntonKhorev:
Creation date indexes might still be useful for combining pagination with date pickers https://github.com/openstreetmap/openstreetmap-website/pull/5255#issuecomment-2413857393

tomhughes:
Well then you'd ideally want an index on author+date+id though we have to think about how much space we want o use on indexes at some point - is here really a need for filtering by date? How often do you know that you need to look at comments from a particular date?

tomhughes:
Those indexes are all pretty tiny currently so I guess keeping them doesn't hurt.

AntonKhorev:
> filtering by date

Not filtering really, just get the starting id by date and then use it as `before` or `after`.

AntonKhorev:
> an index on author+date+id

`+id` is probably not necessary. If a user managed to create multiple objects during the same microsecond we can afford picking a wrong one. Or select min/max(id) where created_at is between target_date and target_date+/-epsilon.

PR #5532, opened 2025-01-21
Set needs_view by default for new blocks

AntonKhorev:
If a block is created without `needs_view`, it's less likely to get noticed by the receiving user. Nobody could come up with any reason why blocks shouldn't be `needs_view` by default.


tomhughes:
Looks good to me, thanks.

PR #5533, opened 2025-01-22
Use a custom cop to check controller action names

gravitystorm:
SpecificActionNames cop taken from https://github.com/rubocop/rubocop-rails/pull/827

We're most of the way through renaming our controller actions to follow the resourceful routing conventions. At this point, I think it would be helpful to provide automated feedback to other developers, if they try writing new code that doesn't follow this convention. By using a custom rubocop configuration, we can catch these situations during CI, but also this helps developers who have rubocop integrated into their IDE to receive early feedback while still writing the code.

The todo list of excluded files is currently provided explicitly in `.rubocop.yml`. This is because if you have more than 15 files failing a check, rubocop will instead disable the cop entirely when (re-)generating `.rubocop_todo.yml`, which would defeat the point of adding it.


tomhughes:
I don't really like bundling third party code like this, but the upstream PR doesn't seem to be going anywhere :disappointed:

PR #5534, opened 2025-01-22
Use flash.now for messages to be shown during this request

gravitystorm:
I'm not sure why the other files were in rubocop_todo, perhaps they had been refactored since the file was last generated.


tomhughes:
Looks good to me, thanks.

PR #5535, opened 2025-01-22
Create changeset subscription resource

AntonKhorev:
This is for pages accessed by following unsubscribe links from comment notification emails, not for subscribe/unsubscribe buttons on changeset pages.

Changes `/changeset/:id/subscribe` and `/changeset/:id/unsubscribe` to `/changeset/:id/subscription`, similar to #5261. Now there's a subscription resource, you can get/post/delete at its path.

I'm not sure why we need these pages since there's a subscribe/unsubscribe button on the changeset page. Probably to have an uncluttered unsubscribe page because they we added in #4534. If so, same should be done for note subscriptions.


tomhughes:
They were added so that we could fill in the `List-Unsubscribe` header in emails as Google had made that a requirement for gmail.

tomhughes:

PR #5536, opened 2025-01-23
Message read_mark and mute resources

AntonKhorev:
Converts remaining actions of messages controller that RuboCop doesn't like into resources:
- `read_mark` which can be created and destroyed instead of `mark` action
- `mute` which can be destroyed instead of `unmute` action

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -43,7 +43,7 @@ class Ability
         can :update, DiaryEntry, :user => user
         can [:create], DiaryComment
         can [:show, :create, :destroy], Follow
-        can [:read, :create, :mark, :unmute, :destroy], Message
+        can [:read, :create, :unmute, :destroy], Message
         can [:close, :reopen], Note
         can [:read, :update], :preference
         can :update, :profile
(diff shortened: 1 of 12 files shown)


app/abilities/ability.rb
```diff
@@ -43,7 +43,7 @@ def initialize(user)
         can :update, DiaryEntry, :user => user
         can [:create], DiaryComment
         can [:show, :create, :destroy], Follow
-        can [:read, :create, :mark, :unmute, :destroy], Message
+        can :manage, Message
```
tomhughes, line 46: This shouldn't be removing `unmute` in the first commit as it doesn't move to it's own controller until the third commit.

tomhughes (changes requested):

app/abilities/ability.rb
AntonKhorev: And it didn't because `:manage` grants all abilities.

Rewrote it to not use `:manage`. Marks and mutes now authorize messages as parent resources, which means they check `:show` (included in `:read`) on messages.

tomhughes (approved):
Thanks - this all looks good now.

PR #5537, opened 2025-01-23
Trace data resource

AntonKhorev:
Adds a resource for downloading the trace at `/traces/:id/data` with redirect from `/trace/:id/data`.

There's already a similar resource on the api side.


tomhughes:
Looks good to me, thanks.

PR #5538, opened 2025-01-23
Initialize I18n and OSM as soon as application.js loads

AntonKhorev:
That's what was happening before https://github.com/openstreetmap/openstreetmap-website/commit/d7b4f88d5e342b5c2b4f2435d0394e4286524678, but after that the init code was moved inside `$(document).ready( ... )`. That means it may not run before other code wants to use it, like https://github.com/openstreetmap/openstreetmap-website/pull/5402#discussion_r1923067165. `<head>` already exists by the time application.js runs, it should be possible to read html attributes from there.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,6 +16,29 @@
 //= require richtext
 //= require qs/dist/qs
 
+{
+  const application_data = $("head").data();
+
+  I18n.default_locale = OSM.DEFAULT_LOCALE;
+  I18n.locale = application_data.locale;
+  I18n.fallbacks = true;
+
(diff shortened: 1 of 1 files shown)


tomhughes:
I generally try to avoid doing anything before `ready` fires as it's hard to know what's safe but I guess this OK as `head` must have loaded in order for the javascript to load. I do wonder what #5402 is doing that it needs this before the pafe is ready though...

HolgerJeromin:
>  but I guess this OK as head must have loaded in order for the javascript to load.

Yes. 
FYI:
`document.body` is not ready sync (if your `<script>` element has no `defer` attribute)

tomhughes:
I thought I had merged this one the other day but apparently not...

PR #5541, opened 2025-01-23
Add user transient attribute to oauth factories

AntonKhorev:
It's annoying to type
```
create(:oauth_access_token, :resource_owner_id => user.id, ...
```

This PR shortens it to
```
create(:oauth_access_grant, :user => user, ...
```

Doorkeper classes don't have actual associations with our models and looks like this line in `oauth_access_grant` factory never worked
```
resource_owner_id :factory => :user
```
And I also have to tell RuboCop not to enforce these associations.

### DIFF
--- a/test/abilities/api_abilities_test.rb
+++ b/test/abilities/api_abilities_test.rb
@@ -36,7 +36,7 @@ end
 
 class ModeratorApiAbilityTest < ApiAbilityTest
   test "Note permissions" do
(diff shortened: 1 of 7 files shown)


tomhughes:
I think I did it the way I did to make it consistent with the non-test code, where there is no user/resource_owner association on the access token model that doorkeeper provides, but this does make the tests a lot nicer.

PR #5543, opened 2025-01-24
Move scope_enabled? to ApiController

AntonKhorev:
`scope_enabled?` is only called from api controllers/views and it references `doorkeeper_token` available only during api calls.


tomhughes:
Looks good to me, thanks.

PR #5544, opened 2025-01-24
Simplify write_notes scope check in api notes controller

AntonKhorev:
`scope_enabled?` returns nil if there's no `doorkeeper_token`, then `author` stays undefined. The else branch removed in this PR would have done the same because there's no `current_user` in this case too.


tomhughes:
Yes it's probably a hangover from when we had other ways to authenticate to the API that would set `current_user` but it's redundant now as you say.

Issue #5523, opened 2025-01-19
Add linting for .js.erb files


hlfan:
### Problem

The JS with embedded Ruby isn't held to the same standards as all the other JS code.

### Description

Please look into enabling linting files with embedded Ruby in ESLint.
The [eslint-plugin-erb](//github.com/Splines/eslint-plugin-erb) could help here.

### Screenshots

![osm.js.erb](https://github.com/user-attachments/assets/d032ec80-5f32-4e4a-83af-3027e6e66663)
![embed.js.erb](https://github.com/user-attachments/assets/a6e11800-77f7-415f-bc68-702798f59bdd)
![id.js.erb](https://github.com/user-attachments/assets/a09b71c5-15ec-4315-953d-fa19b9722a7d)

gravitystorm:
Thanks for the suggestion @hlfan. I've looked into this today, but I've run into a problem with the plugin that you found - we're writing our `.js.erb` files in a way that it isn't happy with.

I've logged an issue for now upstream https://github.com/Splines/eslint-plugin-erb/issues/15

You can see my work so far at https://github.com/gravitystorm/openstreetmap-website/commit/4671f59862f3ba8271f7fbe6214771e534924b7d if anyone else wants to try a different approach.

hlfan:
Is this the shortest time the _awaiting upstream_ tag was necessary?

gravitystorm:
> Is this the shortest time the _awaiting upstream_ tag was necessary?

I think so!

HolgerJeromin:
We should discuss if we want to keep the current eslint config.
Most stylistic rules are now removed from eslint because most use a formatter like prettier.io

hlfan:
That would indeed be an interesting issue to be discussed.

mmd-osm:
Maybe create a new issue for this topic. https://github.com/prettier/plugin-ruby could be starting point.

Splines:
> Most stylistic rules are now removed from eslint because most use a formatter like prettier.io

[...]

PR #5548, opened 2025-01-24
Fix some eslint errors (quotes spacing and parseInt radix) on .js.erb files

gravitystorm:
These have been flagged by the eslint-erb-plugin, and need to be fixed before that can be added.

Refs #5523


tomhughes:
Looks good to me, thanks.

PR #5549, opened 2025-01-24
Add eslint-plugin-erb to provide linting of our .js.erb files

### LINKED ISSUE #5523: Add linting for .js.erb files

gravitystorm:
Fixes #5523 

This is a draft, since our .js.erb files don't currently pass the linting. I've fixed the straightforward ones in #5548 but there are a few more to complete before this can be merged.

I've opened this pull request to make it easier for other developers (who have more JS skills) to help. My recommendation is to checkout this pull request locally, fix the errors, and submit those fixes separately as PRs (e.g. cherry-pick your fixes onto our master branch).

### DIFF
--- a/app/assets/javascripts/edit/id.js.erb
+++ b/app/assets/javascripts/edit/id.js.erb
@@ -10,15 +10,15 @@ $(document).ready(function () {
     var params = {};
 
     if (mapParams.object) {
-      params.id = mapParams.object.type + '/' + mapParams.object.id;
+      params.id = mapParams.object.type + "/" + mapParams.object.id;
       mapParams = OSM.parseHash(location.hash);
       if (mapParams.center) {
-        params.map = mapParams.zoom + '/' + mapParams.center.lat + '/' + mapParams.center.lng;
+        params.map = mapParams.zoom + "/" + mapParams.center.lat + "/" + mapParams.center.lng;
       }
     } else if (id.data("lat") && id.data("lon")) {
       params.map = "16/" + id.data("lat") + "/" + id.data("lon");
     } else {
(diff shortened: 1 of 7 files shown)


mmd-osm:
_✖ 40 problems (31 errors, 9 warnings)
  26 errors and 0 warnings potentially fixable with the `--fix` option._

Is `--fix` of any help here, have you tried this out already?

gravitystorm:
> Is `--fix` of any help here, have you tried this out already?

We run eslint via `bundle exec rails eslint`, and that doesn't support the `--fix` option. Our configuration doesn't support running eslint directly, since we [use ruby code in the configuration](https://github.com/openstreetmap/openstreetmap-website/blob/master/config/initializers/eslint.rb).

I'd like to get it running directly at some point, to improve integration between eslint and IDEs, which are more likely to support eslint directly than via rails tasks. Maybe there's also a way to get the `--fix` working via rails.

mmd-osm:
According to https://github.com/openstreetmap/openstreetmap-website/blob/master/lib/tasks/eslint.rake#L23 we would need to use  `bundle exec rails eslint:fix` instead. This seems to work for me at least.

hlfan:
`64:5` `error` `Assignment to function parameter 'search'` `no-param-reassign` is really unnecessary, the entire function should be refactored to use the URLSearchParams Web API.

PR #5550, opened 2025-01-24
Refactor embed.js.erb

hlfan:
Refactors from #5519 and #5522 without adding dark mode (yet).

### DIFF
--- a/app/assets/javascripts/embed.js.erb
+++ b/app/assets/javascripts/embed.js.erb
@@ -15,14 +15,7 @@ I18n.default_locale = <%= I18n.default_locale.to_json %>;
 I18n.fallbacks = true;
 
 window.onload = function () {
-  var query = (window.location.search || "?").slice(1),
-      args = {};
-
-  var pairs = query.split("&");
-  for (var i = 0; i < pairs.length; i++) {
-    var parts = pairs[i].split("=");
-    args[parts[0]] = decodeURIComponent(parts[1] || "");
-  }
+  var args = Object.fromEntries(new URLSearchParams(window.location.search).entries());
 
   var mapnikOptions = {
 <% if Settings.key?(:tile_cdn_url) %>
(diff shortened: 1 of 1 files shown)


HolgerJeromin (approved):

app/assets/javascripts/embed.js.erb
```diff
@@ -15,14 +15,7 @@ I18n.default_locale = <%= I18n.default_locale.to_json %>;
...
 window.onload = function () {
-  var query = (window.location.search || "?").slice(1),
-      args = {};
-
-  var pairs = query.split("&");
-  for (var i = 0; i < pairs.length; i++) {
-    var parts = pairs[i].split("=");
-    args[parts[0]] = decodeURIComponent(parts[1] || "");
-  }
+  var args = Object.fromEntries(new URLSearchParams(window.location.search).entries());
```
tomhughes, line 18: Do you actually need `.entries()` here? I believe the `URLSearchParams` object itself should be iterable?

app/assets/javascripts/embed.js.erb
```diff
@@ -40,17 +33,16 @@ window.onload = function () {
...
-    new L.OSM.Mapnik(mapnikOptions).addTo(map);
-  }
+  var baseLayers = {
+    cyclosm: ["CyclOSM"],
+    cyclemap: ["CycleMap", thunderforestOptions],
+    transportmap: ["TransportMap", thunderforestOptions],
+    hot: ["HOT"],
+    mapnik: ["Mapnik", mapnikOptions]
+  };
+  baseLayers["cycle map"] = baseLayers.cyclemap;
```
tomhughes, line 43: I think I'd just include this in the object definition after the `cyclemap` line. I realise it means duplicating the right hand side but I'm not sure that's any worse than this and I think it makes it a bit easier to see what's going on - also `baseLayers` could then be `const` I think?

tomhughes (commented):
This looks pretty good - just a couple of minor suggestions that I've made inline.

app/assets/javascripts/embed.js.erb
HolgerJeromin: `const` defines only the variable. I does not prevent changing the properties of the object.
But yes, changing all `var` into `let`/`const` would be good because they work with much less surprise.

app/assets/javascripts/embed.js.erb
```diff
@@ -40,17 +33,16 @@ window.onload = function () {
...
+  var [layer, options] = baseLayers[args.layer] || baseLayers.mapnik;
+  new L.OSM[layer](options || {}).addTo(map);
```
AntonKhorev, line 45: Don't we have [layers.yml](https://github.com/openstreetmap/openstreetmap-website/blob/master/config/layers.yml) with all layers config? Only `mapnik` needs a special case, the rest can be generated from there. Maybe `"cycle map"` is also a special case, or we can remove spaces before looking up the layer.

app/assets/javascripts/embed.js.erb
HolgerJeromin: which was done in #5558

app/assets/javascripts/embed.js.erb
hlfan: I also think big potential here was hidden in the else-if chain. I definitely can see the next version here be [layers.yml](//github.com/openstreetmap/openstreetmap-website/blob/master/config/layers.yml)-based, but it should be filtered to what's necessary and include toggling to `leafletOsmDarkId`.
And I haven't figured out how to do that compactly so this is more of a stop-gap that I can throw a ternary onto to add dark mode later.

PR #5556, opened 2025-01-26
Drop some legacy browser support code

tomhughes:
This drops the ES6 polyfill which is a bit pointless since #5421 allows ES2021 code now.

It also drops an IE specific legacy favicon tag, which allows us to drop the dependency on the browser gem.


HolgerJeromin (approved):

kcne (approved):

AntonKhorev:
Merged, thanks.

Issue #5554, opened 2025-01-25
Modernize eslint config


HolgerJeromin:
We could discuss if we want to ~~keep~~ change the current [eslint config](https://github.com/openstreetmap/openstreetmap-website/blob/master/config/eslint.js).
Most stylistic rules are now removed from eslint because most use a formatter like prettier.io

But the style rules are moved to [another repo](https://eslint.style/) so we could keep them.

Or we could remove them and move to prettier.
Having both prettier and eslint is perhaps making new contributors unhappy. At least when there is CI rejecting PR which is not formatted by prettier.

#5523 has some (off topic) discussions.

tomhughes:
Can you explain what this has to do with keeping the current config? Are you saying that the current config is invoking some style rules that are no longer having any effect? We'd still want to keep the other rules though? We wouldn't want to just dump the whole thing...

HolgerJeromin:
Sorry, I am not an native speaker. I dod not want to remove the config, but change its content.

HolgerJeromin:
As a first step I migrated the deprecated rules.
After that I want to check which rules are already included in `js.configs.recommended`

PR #5557, opened 2025-01-26
Move deprecated eslint rules to stylistic plugin

HolgerJeromin:
### Description
I moved all deprecated eslint to the https://eslint.style/ plugin.

~~To have easy editor support I added a workspace file for visual studio code. Right now this only has a link to eslint config file.
This seems to be a smaller change than moving the config file to default filename in root folder.~~
I removed that commit to wait for decision in #5572

### How has this been tested?
Running eslint in editor.

ref #5554


tomhughes:
Shouldn't this update `yarn.lock` as well as `package.json`?

kcne:
Up until now, whenever I was writing JavaScript, I struggled to get ESLint to work as seamlessly as Prettier—for example, automatically linting files and displaying errors in real-time while writing code. I noticed you added a VS Code workspace in this PR, and I’m hopeful this will help improve the experience. Thank you for adding this!

HolgerJeromin:
> Shouldn't this update yarn.lock as well as package.json?

added.

Splines:
> Up until now, whenever I was writing JavaScript, I struggled to get ESLint to work as seamlessly as Prettier

Feel free to use a configuration like the one [I use here](https://github.com/Manim-Notebook/manim-notebook/blob/68874a7e30489d8774ffc53ddfd96cbc7d885ba6/.vscode/settings.json#L1-L25) for VSCode. This works perfectly fine with the ESLint VSCode extension. It shows problems in real time in the code and whenever I save the file, it autofixes it, e.g. replaces `'` by `"` (dependent on your configuration).

tomhughes:
Does that vscode config file work automatically? It seems like an unlikely location for it to be found? I thought `.vscode/settings.json` was the location for workspace setting to be picked up automatically...

config/eslint.js
```diff
@@ -68,9 +95,6 @@ module.exports = [
       "no-lone-blocks": "error",
       "no-lonely-if": "error",
       "no-loop-func": "error",
-      "no-mixed-operators": "error",
-      "no-multiple-empty-lines": "error",
```
hlfan, line 72: Why has [this rule](//eslint.style/rules/default/no-multiple-empty-lines) not been migrated to stylistic?

hlfan (changes requested):

config/eslint.js
HolgerJeromin: Good catch!
I readded it again and checked that I did not missed any other.

hlfan (approved):

config/eslint.js
hlfan: Now the JSON diff (replacing `(\w+)(?= ?:)` with `"$1"`) shows all rules have been migrated!

package.json
```diff
@@ -10,6 +10,7 @@
   },
   "devDependencies": {
     "eslint": "^9.0.0",
+    "@stylistic/eslint-plugin": "^3.0.0",
```
AntonKhorev, line 13: This includes typescript. Javascript-only version is `@stylistic/eslint-plugin-js`.

HolgerJeromin: I have no idea how my old experiments with that did not worked. But now it does without ts dependency.

AntonKhorev (approved):

AntonKhorev:
Merged, thanks.

PR #5558, opened 2025-01-26
Replace var in javascript with let/const

tomhughes:
This updates the eslint config to prohibit use of  `var` and to prefer `const` over `let` and then fixes all uses of `var` to respect that.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -25,7 +25,7 @@
 //= require router
 
 $(document).ready(function () {
-  var map = new L.OSM.Map("map", {
+  const map = new L.OSM.Map("map", {
     zoomControl: false,
     layerControl: false,
     contextmenu: true,
@@ -33,7 +33,7 @@ $(document).ready(function () {
   });
 
   OSM.loadSidebarContent = function (path, callback) {
-    var content_path = path;
+    let content_path = path;
 
     map.setSidebarOverlaid(false);
 
@@ -78,7 +78,7 @@ $(document).ready(function () {
       });
   };
 
-  var params = OSM.mapParams();
+  const params = OSM.mapParams();
 
   map.attributionControl.setPrefix("");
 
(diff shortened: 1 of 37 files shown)


AntonKhorev:
> and then fixes all uses of var to respect that

Not in unmerged pull requests.

hlfan (approved):

AntonKhorev:
If this is merged, I'll have to rewrite #5064 again. That's a bugfix PR. People are still [complaining about bugs it fixes](https://community.openstreetmap.org/t/osm-website-routing-start-endpoints-jump-from-original-location/124529).

HolgerJeromin (approved):

app/assets/javascripts/index.js
```diff
@@ -42,16 +40,16 @@ $(document).ready(function () {
       .empty();
 
     $.ajax({
-      url: content_path,
+      url: path,
```
hlfan, line 43: If you're already cleaning up after me, you can define the function as `function (url, callback)` for even more conciseness.

hlfan: Or maybe not given #5663

nertc (approved):
All occurrences of `var` are changed with `let/const`. Checked all changed variables and `let` and `const` are set accordingly (every variable that is not changed after declaration, is `const`). Also spent some time on website to test it generally, if any page threw error in the console. Everything I checked works as expected, there are no errors.

HolgerJeromin:
Can we merge this please?
At least I hesitate to do small things which only creates  conflicts with this PR

gravitystorm (approved):
Needs rebase, but otherwise looks good to me. I agree it would be good to merge this as soon as we can.

tomhughes:
I've rebased this now.

HolgerJeromin (approved):

gravitystorm:
Merged, thanks!

PR #5551, opened 2025-01-24
Added error handling for data layer

mmd-osm:
As discussed in https://github.com/openstreetmap/openstreetmap-website/pull/5474#issuecomment-2585349128, this PR adds error handling for failed API calls when displaying the data layer.

Sample screenshot:

image
image

Oh well, and this one is a topic for another PR:

image

PR #5561, opened 2025-01-27
Close button for loading error alert

deevroman:
#5551 added a window with an error if the user is trying to download too much data. But this window cannot be closed, only turn off the Map Data layer. But what if you have already turned it off?

https://github.com/user-attachments/assets/0b40a9ed-c3b7-47ee-90fb-67b264880ff9

Now it looks like this (similar to the warning about loading a large amount of data):


tomhughes:
Looks good to me, thanks.

PR #5562, opened 2025-01-27
Increase the limit when a warning about a large number of objects is displayed

deevroman:
After #5474, data rendering is no longer a bottle of performance. Modern browsers do a great job with moving the map, which displays many objects (moreover, it can also be accelerated)

The only exception is mobile browsers. Therefore, I stopped at 4000 objects.


tomhughes:
Looks good to me, thanks.

PR #5565, opened 2025-01-28
Remove IE conditional compilation that limited max features

AntonKhorev:
This targets Internet Explorer 8, a browser that wouldn't work for many other reasons.


tomhughes:
Looks good to me, thanks.

PR #5567, opened 2025-01-28
Remove jquery-simulate

hlfan:
It [isn't under active development](/jquery/jquery-simulate) anymore.


tomhughes:
Is that replacement something new that didn't exist when `simulate` was created? or does it have limitations compared to `simulate` that we were trying to avoid?

hlfan:
Never mind, what I thought was this version working wasn't and failed middle-click and ctrl+click tests.

PR #5569, opened 2025-01-28
Replace info message with clearer version

mjdominus:
### Description

When a user asks for an "Issues" report, and they have no unsaved edits, the message formerly said:

       You have no edits yet

This changes it to say

       You have no unsaved edits

(The terminology seems to be inconsistent; sometimes we say "unsaved edits" and sometimes "unsaved changes".)

### How has this been tested?

It's just a string change in the `locales/*.json` files, so I didn't test.


tomhughes:
Changes to iD need to be filed at https://github.com/openstreetmap/iD.

PR #5570, opened 2025-01-29
Fix map blink when active Mapnik + some overlay

deevroman:
#5474 added an unpleasant blink when an overlay is enabled and Mapnik is base layer, and you try to close the sidebar:

https://github.com/user-attachments/assets/4d3063b9-dd39-4553-844b-e2bd99c11241

This is because `var layers = layerParam || "M";` does not add `M` when overlay is enabled (`layerParam` is not empty). I added a check that if Mapnik is the current layer, then we add its code, so that`.removeLayer()` was not called for it.

Checking `this.getMapbaslaver() &&` is important because when opening the tab it can return `undefined`.

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -116,6 +116,9 @@ L.OSM.Map = L.Map.extend({
 
   updateLayers: function (layerParam) {
     var layers = layerParam || "M";
+    if (this.getMapBaseLayer() && this.getMapBaseLayer().options.code === "M" && !layers.includes("M")) {
+      layers += "M";
+    }
 
     for (let i = this.baseLayers.length - 1; i >= 0; i--) {
       if (layers.indexOf(this.baseLayers[i].options.code) === -1) {


AntonKhorev:
When the layers parameter `layers=` is empty (or doesn't exist), it could mean one of two things:
- keep currently enabled layers
- reset to default layers (standard base + no overlays)

osm-website's code is undecided about this:
- `map.setState` behaves as if it wants to keep the currents layers because it won't call `updateLayers` when `state.layers` are empty. `state.layers` are populated with the layers param elsewhere
- `OSM.formatHash` behaves as if empty value means standard layer because it removes `M`. Removing `M` doesn't always produce empty layers param because of overlay codes, and that's why this PR was opened.

app/assets/javascripts/leaflet.map.js
```diff
@@ -116,6 +116,9 @@ L.OSM.Map = L.Map.extend({
 
   updateLayers: function (layerParam) {
     var layers = layerParam || "M";
+    if (this.getMapBaseLayer() && this.getMapBaseLayer().options.code === "M" && !layers.includes("M")) {
+      layers += "M";
+    }
```
AntonKhorev, line 121: I prefer a different approach to adding more `M`'s #5573

deevroman: No problems. If PR is no longer relevant, you can close it.

AntonKhorev: Can you check if https://github.com/openstreetmap/openstreetmap-website/pull/5573 actually fixes the same problem and close your pr if it does?

PR #5571, opened 2025-01-29
Refactor embed.js.erb with layers.yml

hlfan:
Like #5550 but with reading the layers.yml file instead of duplicating information


app/assets/javascripts/embed.js.erb
```diff
@@ -15,42 +15,44 @@ I18n.default_locale = <%= I18n.default_locale.to_json %>;
...
-    new L.OSM.CycleMap(thunderforestOptions).addTo(map);
-  } else if (args.layer === "transportmap") {
-    new L.OSM.TransportMap(thunderforestOptions).addTo(map);
-  } else if (args.layer === "hot") {
-    new L.OSM.HOT().addTo(map);
-  } else {
-    new L.OSM.Mapnik(mapnikOptions).addTo(map);
-  }
+  const layers = <%=
+(YAML.load_file(Rails.root.join("config/layers.yml"))
```
mmd-osm, line 39: You could probably use an initializer here to avoid parsing the file all the time: https://github.com/openstreetmap/openstreetmap-website/blob/master/config/initializers/browse_icons.rb

hlfan: When this was introduced in osm.js.erb in #5349 it was also not deemed necessary.

mmd-osm: Ok, let's ignore this, this file part of compiled assets, so my comment doesn't apply.

app/assets/javascripts/embed.js.erb
```diff
@@ -15,42 +16,44 @@ I18n.default_locale = <%= I18n.default_locale.to_json %>;
...
```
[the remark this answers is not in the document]
hlfan: Sure, I tested it directly with `puts`, which didn't like the dangling to_json, but this isn't the case here.

hlfan:
Ok I don't think I could've failed harder in squashing the previous commits 🫠

tomhughes (changes requested):
The code changes here look great now but the commit history is rather messy - can we get rid of the merges please and rebase it instead? and merge any fixup commits into the thing they're fixing - having two `Refactor embed.js.erb with layers.yml` commits in one PR can't be right.

hlfan:
> The code changes here look great now but the commit history is rather messy - can we get rid of the merges please and rebase it instead? and merge any fixup commits into the thing they're fixing - having two `Refactor embed.js.erb with layers.yml` commits in one PR can't be right.

Well, I tried to and failed miserably, as you may have seen. I fixed it in another branch, but as PRs can't have the source branch changed, I opened #5580.

tomhughes:
You don't need to "change the source branch" because the source branch will still be master.

What you need to do is to rebase, and then force push.

hlfan:
But I don't want to pull from the master anyway but from my dev branch, the master should stay the target branch (why would I want to change that now).
> What you need to do is to rebase, and then force push.

What do you think have I been trying to do? Unnecessarily messing up the commit history? This is what I get for trying to follow CONTRIBUTING.md

AntonKhorev:
git rebase master your-branch-name

tomhughes:
In this case `git rebase -i master` is probably better so you can then choose to fold some of the fixup commits into the thing they're fixing, and then `git push -f` to force push the result.

But it you've done a separate PR now then I can review that instead so probably no point in redoing this one now, just something to bear in mind for the future.

PR #5573, opened 2025-01-29
Don't remove-readd standard layer with visible overlays

AntonKhorev:
This is a first commit of another fix which I didn't finish yet because of https://github.com/openstreetmap/openstreetmap-website/pull/5570#issuecomment-2621121469 and some other things[^1] but it should fix the same problem #5570 fixes. Unlike #5570 this PR removes references to `"M"` being the default layer. #5570 add a bunch more. I want to remove `"M"`s because the layer config is in `config/layers.yml` and you just read the code from there.

`map.updateLayers` already has an assumption that the first layer is the default one in addition to the assumption that it's  `"M"`. Potentially these assumptions may diverge, so it's better to remove one, and I'd rather remove `"M"` because it's also avoids the situation of what if there's no `"M"` layer. And that's not impossible, M used to stand for "Mapnik", but we don't call this layer "Mapnik" anymore.

[^1]: like `OSM.STATUS` checks

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -115,19 +115,18 @@ L.OSM.Map = L.Map.extend({
   },
 
   updateLayers: function (layerParam) {
-    var layers = layerParam || "M";
+    const oldBaseLayer = this.getMapBaseLayer();
+    let newBaseLayer;
 
-    for (let i = this.baseLayers.length - 1; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


deevroman:
Yes, it also corrects blinking and works expected:

https://github.com/user-attachments/assets/7f394758-06a4-4f6c-b782-68f31d069e9e

The change in the layer through the URL also works.

tomhughes:
I can't reproduce the blink in Firefox, but the code looks good to me, thanks.

PR #5574, opened 2025-01-29
Remove unnecessary wrapper divs from data layer error message

AntonKhorev:
The message was added in #5551.

There is a slight difference, the flex wrapper shrinks the alert paragraph. We don't do that in `displayFeatureWarning`.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5576, opened 2025-01-29
Show only description as a marker tooltip in note layer

AntonKhorev:
Currently note markers in note layers don't necessarily display note descriptions. There's no check if the first "comment" returned by the api is a description or is it a regular comment because the description is hidden:

https://github.com/user-attachments/assets/50af9642-0067-4b9a-941e-64826c0271d0

I wrote this comment https://github.com/openstreetmap/openstreetmap-website/pull/5511#discussion_r1931817739 where I say that it can be fixed by adding attributes to comments that indicate whether they are descriptions or actually comments. But that is not necessary because there's a status attribute, and if its value is `opened`, the "comment" is a description. This PR adds this check to note layer.

### DIFF
--- a/app/assets/javascripts/index/layers/notes.js
+++ b/app/assets/javascripts/index/layers/notes.js
@@ -40,9 +40,16 @@ OSM.initializeNotesLayer = function (map) {
     if (marker) {
       marker.setIcon(noteIcons[feature.properties.status]);
     } else {
+      let title;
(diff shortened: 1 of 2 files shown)


tomhughes:
Hopefully this will stop being a problem once the notes schema migration is complete but this looks fine as a temporary fix.

PR #5577, opened 2025-01-29
Abort loading when note layer is removed

AntonKhorev:
Same as #5575 but for notes.


tomhughes:
Looks good to me, thanks.

PR #5578, opened 2025-01-29
Select user ids from reports using pluck

AntonKhorev:
#4990 added the *Reporter Users* column. It uses reports to sort the users, but actually all it needs to get from the db are user ids. The existing version with `map` fetches the entire report by doing `SELECT "reports".*`. Here I'm replacing it with `pluck` that does `SELECT "reports"."user_id"`.


tomhughes:
It should also mean that the `uniq` becomes part of the database query I think. Looks good to me in any case!

PR #5579, opened 2025-01-30
Adds optional use of note's author and description v2

nenad-vujicic:
### Description

It seems to me that #5511 didn't go in satisfying direction, so, created new PR based on [original suggestion](https://github.com/openstreetmap/openstreetmap-website/pull/5485#pullrequestreview-2545467592).

PR adds optional use of author and description from notes (in case data-migration is done) and author and description from first visible note comment (in case data-migration is not done). Also, adds author association to Note model.

### How has this been tested?

Automated unit tests and manual testing. Also tested after applying [migration script](https://github.com/openstreetmap/openstreetmap-website/pull/5485/commits/3981e9e02daec8441340bda6572111c795df82cb).

### DIFF
--- a/app/helpers/note_helper.rb
+++ b/app/helpers/note_helper.rb
@@ -1,6 +1,14 @@
 module NoteHelper
   include ActionView::Helpers::TranslationHelper
 
+  def note_description(author, description)
+    if !author.nil? && author.status == "deleted"
+      RichText.new("text", t("notes.show.description_when_author_is_deleted"))
+    else
+      description
+    end
+  end
+
   def note_event(event, at, by)
     if by.nil?
       t("notes.show.event_#{event}_by_anonymous_html",
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -28,6 +28,8 @@
 class Note < ApplicationRecord
   include GeoRecord
 
+  belongs_to :author, :class_name => "User", :foreign_key => "user_id", :optional => true
+
   has_many :comments, -> { left_joins(:author).where(:visible => true, :users => { :status => [nil, "active", "confirmed"] }).order(:created_at) }, :class_name => "NoteComment", :foreign_key => :note_id
   has_many :all_comments, -> { left_joins(:author).order(:created_at) }, :class_name => "NoteComment", :foreign_key => :note_id, :inverse_of => :note
   has_many :subscriptions, :class_name => "NoteSubscription"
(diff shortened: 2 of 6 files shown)


app/models/note.rb
```diff
@@ -91,12 +93,20 @@ def freshly_closed_until
 
   # Return the note's description, derived from the first comment
   def description
-    comments.first.body
+    if user_ip.nil? && user_id.nil?
+      comments.first.body
+    else
+      RichText.new("text", super)
+    end
   end
```
AntonKhorev, line 101: How is this PR going to go in a different direction?

You're again first revealing the description (if the migration is completed) then maybe later you'll hide it with a helper. That brings you back to https://github.com/openstreetmap/openstreetmap-website/pull/5511#discussion_r1921570888. Why wouldn't you start with a helper?

nenad-vujicic: > How is this PR going to go in a different direction?

Well, honestly, I don't know .. We started with one idea (simple wrappers around `comments`) and converged to something different (`all_comments` + other stuffs) in #5511, so I thought there is perhaps another way. Plus, I'm not sure why #5511 is evolving so slow (I believe I messed up something, but not sure where). It's not a problem to wait some time (how much we need) to digest these changes and be sure they will work fine, but is that the case here (we already have some plan + it's ~50 lines of code)?

If you are sure #5511 is proper direction, I'll be very happy to close this PR :-)

app/helpers/note_helper.rb
```diff
@@ -1,6 +1,14 @@
 module NoteHelper
   include ActionView::Helpers::TranslationHelper
 
+  def note_description(author, description, user_id, user_ip)
+    if !(user_id.nil? && user_ip.nil?) && !author.nil? && author.status == "deleted"
+      RichText.new("text", t("notes.show.description_when_author_is_deleted"))
+    else
+      description
+    end
+  end
```
AntonKhorev, line 10: By *Why wouldn't you start with a helper?* I meant *Why wouldn't you make it the first commit of the PR*.

As long as you're taking `description` from exactly the same place as `author`, you only need to check if `author` is deleted. You don't need to check `user_id` and `user_ip`. You already have a commit doing exactly that in another PR: https://github.com/openstreetmap/openstreetmap-website/pull/5511/commits/9e70276e1997d22e7e4cf496b87157749bb9e468.

nenad-vujicic: Thanks for catching this! Fixed.

Also, reordered commits. Thanks!

PR #5580, opened 2025-01-31
Refactor embed.js.erb and inlcude layers from layers.yml

hlfan:
Like #5571 but without fixup commits.
Closes #5550 and closes #5571


tomhughes:
Looks good to me, thanks.

PR #5581, opened 2025-01-31
Add coordinate precision function

hlfan:
The amount of `latlng.lat/lng.toFixed(precision)` calls has been bugging me.
I've put a function next to OSM.zoomPrecision that outputs a latlng array that can be destructured, spread and `join`ed.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -116,18 +116,14 @@ OSM.Directions = function (map) {
     // Cancel any route that is already in progress
     if (routeRequest) routeRequest.abort();
 
-    var o = endpoints[0].latlng,
-        d = endpoints[1].latlng;
+    const points = endpoints.map(p => p.latlng);
 
-    if (!o || !d) return;
+    if (!points[0] || !points[1]) return;
     $("header").addClass("closed");
-
-    var precision = OSM.zoomPrecision(map.getZoom());
 
     OSM.router.replace("/directions?" + Qs.stringify({
       engine: chosenEngine.id,
-      route: o.lat.toFixed(precision) + "," + o.lng.toFixed(precision) + ";" +
-             d.lat.toFixed(precision) + "," + d.lng.toFixed(precision)
+      route: points.map(p => OSM.cropLocation(p, map.getZoom()).join()).join(";")
     }));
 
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/index/directions.js
```diff
@@ -116,18 +116,14 @@ OSM.Directions = function (map) {
     // Cancel any route that is already in progress
     if (routeRequest) routeRequest.abort();
 
-    var o = endpoints[0].latlng,
-        d = endpoints[1].latlng;
+    const points = endpoints.map(p => p.latlng);
 
-    if (!o || !d) return;
+    if (!points[0] || !points[1]) return;
```
hlfan, line 121: Which option would you prefer?
```js
if (!points[0] || !points[1]) return;
if (!(points[0] && points[1])) return;
if (!points.every(p => p)) return;
if (points.some(p => !p)) return;
```

HolgerJeromin: I find the current PR version the easiest to read.

app/assets/javascripts/osm.js.erb
```diff
@@ -211,11 +207,16 @@ OSM = {
     return Math.ceil(Math.log10(pixels / degrees));
   },
 
+  cropLocation: function (latLng, zoom = map.getZoom()) {
```
tomhughes, line 210: As the test failures showed the default value here doesn't work (because `map` is not defined) so we should get rid of it.

tomhughes (changes requested):
Other than the one change I've requested can you squash the two fixup commits down into the first commit please.

app/assets/javascripts/osm.js.erb
hlfan: I'm still trying to work out the problem in #5571, but force-pushing from another branch helped.

tomhughes: It might help to read https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History but in this case all it needed was `git rebase -i master` which would open an editor with a list of commits on the branch - then you can just change `pick` to `fixup` (or `f`) for the two later ones to merge them into the first one and save and exit the editor to finish.

tomhughes (approved):
Thanks for the changes - this looks good now.

PR #5582, opened 2025-01-31
Extend user block periods

kmpoppe:
Extend user block periods to include 15 and 20 years

### Description
There are currently around 10 user blocks that were created with a duration of 9 years and will end on 2026-01-01. Earlier blocks from 2015 and before that were decided as "indefinite" have ended on 2025-01-01.

To prevent 10 year bans from expiring even though the user was supposed to be banned indef. and to renew bans that have expired, this PR adds block periods of 15 and 20 years.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -54,7 +54,7 @@ api_timeout: 300
 # Timeout for web pages in seconds
 web_timeout: 30
 # Periods (in hours) which are allowed for user blocks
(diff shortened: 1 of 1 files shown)


tomhughes:
If the DWG wants indefinite then we should just add indefinite rather than adding another five years every so often, especially since that won't actually change any existing blocks.

kmpoppe:
Should be discussed, agreed. Closing.

PR #5584, opened 2025-02-01
Remove unused visible_nodes hashes from api element controllers

AntonKhorev:
There was a method called `to_xml_node` defined on way models. It had a cache parameter, and `visible_nodes` was passed as this parameter in https://github.com/openstreetmap/openstreetmap-website/commit/6300fa2a4fa717ef60f715a5303916b85d951c47. Later in https://github.com/openstreetmap/openstreetmap-website/commit/363155a2a86796d27ee9161f1ee9b74d6fa307e5 xml output was rewritten and `to_xml_node` was no longer used. The method was removed in https://github.com/openstreetmap/openstreetmap-website/commit/2b1bac1279e268c8e5fcca34c950f30d2c06ec1c. Same for relations.

However `visible_nodes` stayed. It's being written to but is never read.

### DIFF
--- a/app/controllers/api/relations_controller.rb
+++ b/app/controllers/api/relations_controller.rb
@@ -102,14 +102,11 @@ module Api
         node_ids += way_node_ids.flatten
         nodes = Node.where(:id => node_ids.uniq).includes(:node_tags)
 
-        visible_nodes = {}
-
         @nodes = []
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #5585, opened 2025-02-01
Remove ids collected in parallel to model instances in map api

AntonKhorev:
Another instance of `visible_nodes` like in #5584, but this time it's being read. However it's largely useless because nodes are already collected into `@nodes`. Same for `way_ids`.


tomhughes:
Looks good to me, thanks.

PR #5586, opened 2025-02-01
Use flat_map to collect way node ids in maps controller

AntonKhorev:
...instead of calling `flatten!` on the result


tomhughes:
Looks good to me, thanks.

PR #5588, opened 2025-02-01
Use URI.encode_uri_component when generating share urls

### LINKED ISSUE #5587: generate_share_url: wrong encoding for email title
### URL

_No response_

1. Pick random diary entry with spaces in title, e.g. https://www.openstreetmap.org/user/watmildon/diary/406063
2. Click on "Share via Email" button
3. Note that the subject in your email client reads as "`OSM+US+Board+Candidate+Statement+2025`"

Issue: `title = URI.encode_www_form_component(title) `  is wrong encoding for email subjects, it should probably be `URI.encode_uri_component`, which replaces space characters by %20

See https://stackoverflow.com/questions/68835994/how-to-have-a-subject-line-with-spaces-in-an-email-link

_No response_

AntonKhorev:
Fixes #5587.

I think it's safer to always use `URI.encode_uri_component` instead of `URI.encode_www_form_component`. Space is going to be encoded as `%20`. Space as `+` sometimes doesn't work.

### DIFF
--- a/app/helpers/social_share_button_helper.rb
+++ b/app/helpers/social_share_button_helper.rb
@@ -35,8 +35,8 @@ module SocialShareButtonHelper
 
   def generate_share_url(site, title, url)
     site = site.to_sym
-    title = URI.encode_www_form_component(title)
-    url = URI.encode_www_form_component(url)
+    title = URI.encode_uri_component(title)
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks,

PR #5589, opened 2025-02-02
Filter out invisible nodes and ways during initial selection

tomhughes:
This takes #5585 a bit further and drops the separate visibility filtering in favour of ensuring that we always filter for that as part of the initial database query.

Technically this creates a slight change in behaviour in that previous nodes which were visible but outside the bounding box would be returned if they were part of an invisible way that had nodes inside the bounding box but that seems like it was more of a bug than a feature.


AntonKhorev:
Looks like cgimap only checks visibility when selecting nodes in bbox. It's easier not to check it after that because you can get ids from connector tables (`current_way_nodes` etc) that don't store visibility.

mmd-osm:
> [...] if they were part of an invisible way that had nodes inside the bounding box  [...]

Tbh, I don't really understand this use case, and I didn't find a corresponding unit test for this specific scenario.

As soon as we delete a way (which sets the `current_way` table's `visible` flag to false), we also delete all entries in the `current_way_nodes` table for this way. An invisible way doesn't have any nodes at all, in particular not inside the bounding box.

tomhughes:
I mean maybe it wasn't actually possible, but that was the logical effect of the code.

PR #5590, opened 2025-02-02
API element and changeset resources

AntonKhorev:
Removes all weird actions like `relations_for_way` from api element and changeset controllers.

Previous attempt to update changeset routes: #4589.

Too large, I'll split it into smaller PRs.

- [ ] after upload resource is done, address #5758


jeanozahn:
Great!

PR #5591, opened 2025-02-02
API element resources - index paths

AntonKhorev:
Part of #5590.

Makes element index paths resourceful:
- `/api/0.6/nodes`
- `/api/0.6/ways`
- `/api/0.6/relations`

Aside from the `index` action, enables `create` by posting at these paths.

However, `create` already exists in the api at `PUT /api/0.6/node/create` etc. This is not a standard way to define `create` actions. `PUT` is supposed to be [idempotent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT), but obviously `create` create is not. You do `PUT /api/0.6/node/create`, you get one new node. You do it twice, you get two.[^1] I'm keeping the existing put-creates by mapping them to the same actions as proper post-creates.

[^1]: That also means that `update` actions which are implemented with `PUT`s are wrong because they create as many new versions as there are `PUT`s. We should be creating new versions by `POST`s instead. I'm not fixing this yet here.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -39,7 +39,6 @@ OpenStreetMap::Application.routes.draw do
     get "node/:id" => "nodes#show", :as => :api_node, :id => /\d+/
     put "node/:id" => "nodes#update", :id => /\d+/
     delete "node/:id" => "nodes#delete", :id => /\d+/
-    get "nodes" => "nodes#index"
 
     put "way/create" => "ways#create"
     get "way/:id/history" => "old_ways#history", :as => :api_way_history, :id => /\d+/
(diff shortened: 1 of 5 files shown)


gravitystorm:
Looks good to me. Merged, thanks.

AntonKhorev:
> Why plural?

Because Rails. They like plurals.

westnordost:
The other calls (get, put, delete) are in singular, though.

AntonKhorev:
They should also be in plural.

westnordost:
What? But they aren't. (??)

PR #5592, opened 2025-02-02
Replace qs with URLSearchParams

tomhughes:
As javascript now has a builtin way to parse URL search parameters we don't really need the qs module any more, so replace it with the builtin support.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -14,7 +14,6 @@
 //= require oauth
 //= require matomo
 //= require richtext
-//= require qs/dist/qs
 
 {
   const application_data = $("head").data();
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/application.js
```diff
@@ -45,34 +44,34 @@
  */
 window.updateLinks = function (loc, zoom, layers, object) {
   $(".geolink").each(function (index, link) {
-    var href = link.href.split(/[?#]/)[0],
-        args = Qs.parse(link.search.substring(1)),
-        editlink = $(link).hasClass("editlink");
+    let href = link.href.split(/[?#]/)[0];
```
HolgerJeromin, line 47: Not sure if this is a good and correct change:
```suggestion
    let href = link.protocol + '//' + link.hostname + link.pathname;
```

tomhughes: I'm not actually changing that line though? Other than replacing `var` with `let`?

HolgerJeromin: I thought the general goal was to replace complex code with modern APIs.
If this is an external library or own code is for me not important. 
Do what you want in this PR.

app/assets/javascripts/index/contextmenu.js
```diff
@@ -39,7 +37,7 @@ OSM.initializeContextMenu = function (map) {
     callback: function describeLocation(e) {
       const [lat, lon] = OSM.cropLocation(e.latlng, map.getZoom()).map(encodeURIComponent);
```
AntonKhorev, line 38: `encodeURIComponent` here probably doesn't do anything

app/assets/javascripts/index/contextmenu.js
```diff
@@ -64,6 +62,10 @@ OSM.initializeContextMenu = function (map) {
     else map.contextmenu.enable();
   });
 
+  function stringifyArgs(args) {
+    return new URLSearchParams(args).toString();
+  }
+
```
AntonKhorev, line 68: Is this function required? You can replace 
```js
OSM.router.route("/directions?" + stringifyArgs({
...
```
with
```js
OSM.router.route("/directions?" + new URLSearchParams({
...
```
and this will work because string coercion runs `toString`.

tomhughes: Thanks for point that out - hopefully I've fixed all those now.

app/assets/javascripts/index/contextmenu.js
tomhughes: Maybe, but it's not something new in this pull request?

PR #5593, opened 2025-02-02
Don't check ids of new notes in create note system tests

AntonKhorev:
If I'm making purely ui tests here I can't possibly know id of new notes, therefore I shouldn't check them.

But the actual reason is that it looks like `assert_content "Unresolved note ##{Note.last.id}"` line is sometimes reached before db writes are complete. `Note.last.id` is not going to have an expected value in these cases.


tomhughes:
Looks good to me, thanks.

PR #5594, opened 2025-02-02
Change "reporter users" to "reporting users"

tomhughes:
Reporter users isn't really grammatically correct so change it to reporting users instead.


AntonKhorev:
Merged, thanks.

PR #5595, opened 2025-02-03
Don't apply extra encodeURIComponent to lat/lon search params

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/5592#discussion_r1938530806

In practice this should make no difference because lat/lon don't contain any characters that need escaping. But there's still no reason for attempting to escape them twice.


HolgerJeromin (approved):

hlfan (approved):
Thanks for catching that!

tomhughes:
Looks good to me, thanks.

PR #5597, opened 2025-02-03
Simplify iD's initialisation

hlfan:
Over the years this script has grown an increasing amount of technical debt. I tried to remove some.

### DIFF
--- a/app/assets/javascripts/edit/id.js.erb
+++ b/app/assets/javascripts/edit/id.js.erb
@@ -1,53 +1,36 @@
 $(document).ready(function () {
-  var id = $("#id-embed");
+  const id = $("#id-embed");
 
-  if (id.data("configured") === true) {
-    var hash = location.hash.substring(1);
-    var hashParams = hash ? OSM.params(hash) : {};
-    var mapParams = OSM.mapParams();
-    var params = new URLSearchParams();
-
-    if (mapParams.object) {
-      params.set("id", mapParams.object.type + "/" + mapParams.object.id);
-      mapParams = OSM.parseHash(location.hash);
-      if (mapParams.center) {
-        params.set("map", mapParams.zoom + "/" + mapParams.center.lat + "/" + mapParams.center.lng);
-      }
-    } else if (id.data("lat") && id.data("lon")) {
-      params.set("map", "16/" + id.data("lat") + "/" + id.data("lon"));
-    } else {
-      params.set("map", (mapParams.zoom || 17) + "/" + mapParams.lat + "/" + mapParams.lon);
-    }
-
-    if (hashParams.background) params.set("background", hashParams.background);
-    if (hashParams.comment) params.set("comment", hashParams.comment);
-    if (hashParams.disable_features) params.set("disable_features", hashParams.disable_features);
-    if (hashParams.hashtags) params.set("hashtags", hashParams.hashtags);
-    if (hashParams.locale) params.set("locale", hashParams.locale);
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/edit/id.js.erb
```diff
@@ -1,50 +1,30 @@
...
-      }
-    } else if (id.data("lat") && id.data("lon")) {
-      params.set("map", "16/" + id.data("lat") + "/" + id.data("lon"));
-    } else {
-      params.set("map", (mapParams.zoom || 17) + "/" + mapParams.lat + "/" + mapParams.lon);
-    }
+  if (id.data("configured") === false) {
+    alert(I18n.t("site.edit.id_not_configured"));
+    return;
+  }
```
tomhughes, line 7: I'm not sure I really understand why making this an early return is an improvement? In general I'm not keen on early/multiple returns though I've mostly been holding back on that. If we're going to have it then I'd suggest a blank line afterwards at least.

app/assets/javascripts/edit/id.js.erb
```diff
@@ -1,50 +1,30 @@
...
-    if (id.data("gpx")) {
-      params.set("gpx", id.data("gpx"));
-    } else if (hashParams.gpx) {
-      params.set("gpx", hashParams.gpx);
-    }
+  params.set("map", [
+    (mapParams.object && hashArgs.center && hashArgs.zoom) || (id.data("lat") && id.data("lon") && 16) || mapParams.zoom || 17,
+    (mapParams.object && hashArgs.center?.lat) || (id.data("lon") && id.data("lat")) || mapParams.lat,
+    (mapParams.object && hashArgs.center?.lng) || (id.data("lat") && id.data("lon")) || mapParams.lon
+  ].join("/"));
```
tomhughes, line 20: I'm sure this is all very clever, but it is better than what we had before? Just trying to work out if it's actually doing the same thing is making my head hurt and it seems to me at least to be much harder to reason about and figure out what the supported parameter combinations are and how they map to a location to be passed to iD.

app/assets/javascripts/edit/id.js.erb
hlfan: I'm not that eager to have that early return, but in cases where the `else` is waiting for 40 lines to surprise me, making me look for which `if` it refers to, I'd rather get it out of the way first. But I can drop it if that's not wanted.

app/assets/javascripts/edit/id.js.erb
hlfan: It decouples the parameters' conditions, opening them up to simplification.
``` js
[
  mapParams.zoom || 17,  // dropping hashArgs, this is already used in mapParams if it exists,
                         // and the secondary default value of 16
  hashArgs.center?.lat || id.data("lat") || mapParams.lat,
  hashArgs.center?.lng || id.data("lon") || mapParams.lon
]
```
And I'd argue this is less hard to reason about than a version written with `if`-`else if`.

app/assets/javascripts/edit/id.js.erb
HolgerJeromin: Early return results often in less indention which is IMO good.

And while reading the code you do not need to find the `else` branch.
Having the branch with fewer code lines on top can help without the downsides of multiple returns.

tomhughes: I guess I was just taught that multiple exit functions were bad but I know it's not something there is universal agreement on.

PR #5600, opened 2025-02-03
Refactor api changeset discussion tests

AntonKhorev:
Replaces lines like this
```ruby
assert_select "osm>changeset>discussion>comment:nth-child(1)>@id", comment1.id.to_s
```
with nested `assert_dom`s.


tomhughes:
Looks good to me, thanks.

PR #5604, opened 2025-02-04
Reevaluate iteration methods

hlfan:
Converts more loops to for..of loops due to increased readability and change .forEach() in combination with .push() to .map().

### DIFF
--- a/app/assets/javascripts/leaflet.layers.js
+++ b/app/assets/javascripts/leaflet.layers.js
@@ -60,11 +60,11 @@ L.OSM.layers = function (options) {
       });
 
       input.on("click", function () {
-        layers.forEach(function (other) {
+        for (const other of layers) {
           if (other !== layer) {
             map.removeLayer(other);
           }
-        });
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/leaflet.layers.js
```diff
@@ -60,11 +60,9 @@ L.OSM.layers = function (options) {
...
 
       input.on("click", function () {
-        layers.forEach(function (other) {
-          if (other !== layer) {
-            map.removeLayer(other);
-          }
-        });
+        layers
+          .filter(other => other !== layer)
+          .forEach(map.removeLayer);
```
AntonKhorev, line 65: Have you tried running it?

hlfan: Always in the places where you least expect it...

HolgerJeromin (approved):

tomhughes:
Looks good to me now, thanks.

PR #5605, opened 2025-02-04
Don't html-escape note description

AntonKhorev:
The description is already html-safe after to_html.

This is a leftover from https://github.com/openstreetmap/openstreetmap-website/commit/01aa27031526280d8b1d219902a7b4c97617d19f, we don't explicitly escape other text->richtext->html like note comments.


tomhughes:
Looks good to me, thanks.

PR #5607, opened 2025-02-04
Add copyright/attribution message to downloaded map images

### LINKED ISSUE #551: Downloaded map images do not have a copyright/attribution message
Map images that are downloaded via "Share" - "Download" do not have a copyright message. This is mainly an issue when the download gets printed without further processing (assuming that somebody that imbeds an image in a website should be capable of adding correct attribtution). 

Best solution is likely to have this as a user selectable option.

### LINKED ISSUE #735: Not all redactions shown
On http://www.openstreetmap.org/redactions only 10 redactions are shown, but there 20.

mmd-osm:
This PR covers the UI facing parts for https://github.com/openstreetmap/chef/pull/735 and adds WebP as additional output format, as well as a checkbox to toggle copyright/attribution in the map export.

image

Fixes #551 
Depends on https://github.com/openstreetmap/chef/pull/735 (=make sure the export script has been deployed and works properly before merging)

### DIFF
--- a/app/assets/javascripts/leaflet.share.js
+++ b/app/assets/javascripts/leaflet.share.js
@@ -159,6 +159,7 @@ L.OSM.share = function (options) {
           .attr("class", "form-select w-auto")
           .append($("<option>").val("png").text("PNG").prop("selected", true))
           .append($("<option>").val("jpeg").text("JPEG"))
+          .append($("<option>").val("webp").text("WEBP"))
           .append($("<option>").val("svg").text("SVG"))
           .append($("<option>").val("pdf").text("PDF"))));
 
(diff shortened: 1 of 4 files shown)


tomhughes:
I raised this at https://github.com/openstreetmap/chef/pull/735 but it might be best discussed here - should the attribution be optional? or should we always add it?

matkoniecz:
I would drop that optional attribution, and just always include it.

If someone knows what they are doing, they can produce image in a different way, edit it (say produce larger and crop) and so on.

But attribution directly on image is less likely to be lost, more clear to users and we avoid extra complexity. Also, based on my experience: attribution next to image is likely to be lost when people share image further. Happened with one of quite widely shared OSM-based map that I made.

Optional attribution field may be also confusing for users and cause them to think that it is actually entirely optional to attribute OSM. And in general osm.org suffers from to much options and configurations rather than from having not enough of them.

pnorman:
I was originally in favour of optional attribution, but @matkoniecz arguments persuade me to think it should always be there.

mmd-osm:
It seems there's no need to support optional attribution. I'm going to close this PR and create another one for this webp one liner:

```
diff --git a/app/assets/javascripts/leaflet.share.js b/app/assets/javascripts/leaflet.share.js
index e89608afd..fdad69498 100644
--- a/app/assets/javascripts/leaflet.share.js
+++ b/app/assets/javascripts/leaflet.share.js
@@ -159,6 +159,7 @@ L.OSM.share = function (options) {
           .attr("class", "form-select w-auto")
           .append($("<option>").val("png").text("PNG").prop("selected", true))
           .append($("<option>").val("jpeg").text("JPEG"))
+          .append($("<option>").val("webp").text("WEBP"))
           .append($("<option>").val("svg").text("SVG"))
           .append($("<option>").val("pdf").text("PDF"))));
```

PR #5609, opened 2025-02-05
Fixes bug when deleted user's note lacks comments

nenad-vujicic:
### Description

PR fixes bug where a note of deleted user has no comments, but the code tries to access the first comment by switching to using `.all_comments` (instead of `.comments`) and removing dropping first comment (if author is deleted) when displaying notes comments in sidebar.

Suggested [here](https://github.com/openstreetmap/openstreetmap-website/pull/5579#issuecomment-2634185487).

### How has this been tested?

Automated tests, manual testing before and after migration.


app/views/notes/show.html.erb
```diff
@@ -52,10 +52,10 @@
     <% end %>
   </div>
 
-  <% if @note_comments.length > 1 %>
+  <% if @note_comments.length > (!@note.author.nil? && @note.author.status == "deleted" ? 0 : 1) %>
```
AntonKhorev, line 55: I think it's better to make `@note_comments` contain true comments in the controller instead of having this condition checked twice here.

`@note_comments` were used in three places of this view: description, anonymous warning and discussion.
- Description is handled differently now.
- Discussion is going to work with all `@note_comments` without deciding which to drop.
- Anonymous warning will need to check if either note author is nil or any comment's author is nil. Maybe this check is also better done in the controller and saved to some variable.

nenad-vujicic: Fixed in the last push. Thanks!

Q: Do we need to replace current anonymous warning to something like "This note **is created by anonymous user or** includes comments from anonymous users **so** should be independently verified." or current message is good enough?

AntonKhorev: Maybe but nobody is complaining about this message. It mostly means that the note was created by an anonymous user because anonymous comments were disabled years ago. So the message probably should say just that the note was created by an anonymous user unless there are also anonymous comments.

AntonKhorev:
Merged, thanks.

tomhughes:
I understand the desire to fix this but we did have a plan to do that by refactoring notes, which is now well under way, so I'm not quite sure why we had to jump in with this interim solution which has no tests and which I'm not sure is even correct.

AntonKhorev:
> but we did have a plan to do that by refactoring notes

Which part of that plan included reproducing buggy behavior throughout migrations?

PR #5612, opened 2025-02-05
Replace deprecated eslint rules

### LINKED ISSUE #5554: Modernize eslint config

HolgerJeromin:
Replace deprecated eslint rules

no-new-object => no-object-constructor
https://eslint.org/docs/latest/rules/no-new-object
no-process-env => will be catched by no-undef

closes #5554


tomhughes:
So as https://eslint.org/docs/latest/rules/no-process-env explains there is a replacement (in a plugin) for `no-process-env` but as it's a Node.js specific rule it doesn't really concern us anyway and as `process.env` doesn't exist in our setup it will alert in other ways as @HolgerJeromin says.

So this looks good to me, thanks.

PR #5613, opened 2025-02-05
Fix all no-implicit-globals eslint errors

HolgerJeromin:
### Description
I also changed the eslint rule to an error in a separate commit.

### DIFF
--- a/app/assets/javascripts/social_share_button.js
+++ b/app/assets/javascripts/social_share_button.js
@@ -1,15 +1,15 @@
-function openShareUrl(url, initialWidth = 640, initialHeight = 480) {
-  const width = Math.max(100, Math.min(screen.width, initialWidth));
-  const height = Math.max(100, Math.min(screen.height, initialHeight));
-
-  const left = screenLeft + ((outerWidth - width) / 2);
-  const top = screenTop + ((outerHeight - height) / 2);
-  const opts = `width=${width},height=${height},left=${left},top=${top},menubar=no,status=no,location=no`;
-
-  window.open(url, "popup", opts);
-}
-
 $(document).ready(function () {
+  function openShareUrl(url, initialWidth = 640, initialHeight = 480) {
+    const width = Math.max(100, Math.min(screen.width, initialWidth));
+    const height = Math.max(100, Math.min(screen.height, initialHeight));
+
+    const left = screenLeft + ((outerWidth - width) / 2);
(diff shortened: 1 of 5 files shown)


hlfan (commented):
You don't need to touch every single line if you write the globals as
```js
const DemoEngine = function (id, costing) {
  ...
};
```

app/assets/javascripts/social_share_button.js
```diff
@@ -1,21 +1,22 @@
-function openShareUrl(url, initialWidth = 640, initialHeight = 480) {
-  const width = Math.max(100, Math.min(screen.width, initialWidth));
-  const height = Math.max(100, Math.min(screen.height, initialHeight));
+(function () {
+  function openShareUrl(url, initialWidth = 640, initialHeight = 480) {
```
tomhughes, line 2: Can't this one be fixed by just moving the function inside the ready call, rather than adding another layer using the horrid IIFE syntax?

HolgerJeromin:
> You don't need to touch every single line via indenting if you write the globals as
> const DemoEngine = function (...) {   ... };

Then we would have `DemoEngine` poluting the global scope...

app/assets/javascripts/social_share_button.js
HolgerJeromin: Perhaps we can use the following pattern. But we have to skip `no-lone-blocks` rule.

```js
// eslint ignore no-lone-blocks
{
  function FOSSGISOSRMEngine(id, vehicleType) { ... }
  FOSSGISOSRMEngine(...)
}

tomhughes: Well for that specific one we can just move the function inside the ready callback surely?

HolgerJeromin: oh, you meant the ready call of jQuery. Would be probably another idea. Yes.

AntonKhorev:
> > You don't need to touch every single line via indenting if you write the globals as
> > `const DemoEngine = function (...) {   ... };`
> 
> Then we would have `DemoEngine` poluting the global scope...

You can put it in `OSM` scope. `OSM.RoutingEngines.DemoEngine = ...`

hlfan:
If it's put in `OSM` scope, maybe put the factories in the neighborhood of `OSM.Directions.engines` where the engines end up.

tomhughes (approved):
Looks good to me now, thanks.

PR #5614, opened 2025-02-05
Mark valid usecases for alert and change eslint rule to error

HolgerJeromin:
Mark valid usecases for alert

I think with marking this valid use cases it is ok to change eslint rule to error.


tomhughes (changes requested):
Looks like you missed one?

HolgerJeromin:
Oh, yes. Will fix the pr tomorrow

tomhughes (approved):
Looks good to me now, thanks.

PR #5617, opened 2025-02-05
API element resources - show/update/delete paths

AntonKhorev:
Follows #5591.

- Renames `delete` actions to standard `destroy`.
- Adds tests for `Last-Modified` headers when showing elements.
- Adds more tests for the contents of element `show` and `full` responses.


gravitystorm:
Merged, thanks!

PR #5620, opened 2025-02-05
API element resources - full paths

AntonKhorev:
`full` is not a standard action name, but it's also not really a nested resource. Here I'm mapping them to `show` action with a parameter. A similar thing is done with changeset discussions where the discussion/comment elements are extra things added to the usual `show` output if a parameter is set.

Also there were `user/details` in #5433 which we didn't turn into a nested resource.


tomhughes:
Looks good to me, thanks.

PR #5622, opened 2025-02-06
Updates notes filtering to search description too

nenad-vujicic:
### Description

PR updates notes filtering (querying) to search for query text in both note comments and note's description.

### How has this been tested?

Using automated tests before and after migration plus manual tests before and after migration.

### DIFF
--- a/db/migrate/20250206202905_add_text_index_to_notes.rb
+++ b/db/migrate/20250206202905_add_text_index_to_notes.rb
@@ -1,0 +1,7 @@
+class AddTextIndexToNotes < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_index :notes, "to_tsvector('english', description)", :using => "GIN", :name => "index_notes_on_description", :algorithm => :concurrently
+  end
+end
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -263,7 +263,9 @@ module Api
       end
 
       # Add any text filter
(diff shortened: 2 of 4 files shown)


tomhughes (changes requested):
We'll need to add a text index on `notes.description` like the one we have on `note_comments.body` or this will have to read every note.

nenad-vujicic:
> We'll need to add a text index on `notes.description` like the one we have on `note_comments.body` or this will have to read every note.

Pushed, thanks! Is it safe to have it in same PR or we have to create new PR with migration for adding index?

db/migrate/20250206202905_add_text_index_to_notes.rb
```diff
@@ -0,0 +1,11 @@
+class AddTextIndexToNotes < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def up
+    add_index :notes, "to_tsvector('english', description)", :using => "GIN", :name => "index_notes_on_description", :algorithm => :concurrently
+  end
+
+  def down
```
tomhughes, line 8: There's no need for a separate `down` here - you get get rid of it and rename `up` as `change` and rails will figure out how to reverse things if necessary.

tomhughes (changes requested):

db/migrate/20250206202905_add_text_index_to_notes.rb
nenad-vujicic: Fixed (and pushed), thanks!

tomhughes (approved):
Thanks - this looks good now I think.

PR #5625, opened 2025-02-06
Remove spaces before colons in en.yml

AntonKhorev:
Otherwise it's getting fixed in some unrelated commit, like #4042.


tomhughes:
Looks good to me, thanks.

PR #5626, opened 2025-02-06
API element resources - nested elements

AntonKhorev:
Creates nested resources instead of the following actions: `ways_for_node`, `relations_for_node`, `relations_for_way`, `relations_for_relation`. After that there's no nonstandard actions in api element controllers for current versions of elements.

The current structure is a bit surprising too. Here's a route to get ways that include a given node for example:
```
get "node/:id/ways" => "ways#ways_for_node"
```
As you can see the action is not defined in the nodes controller, rather it's in the ways controller. The ways controller receives an id param that's not an id of a way. In this PR I'm adding another controller for ways that include a node, and three more controllers for relations that contain a node/way/relation.


tomhughes:
Looks good to me, thanks.

PR #5627, opened 2025-02-06
Improvements for notes with deleted authors

tomhughes:
This is a follow on to #5609 that does a couple of things.

Firstly it adds a test for a note whose original author has been deleted.

Secondly it corrects the test for whether to drop the first comment to align with what the note model does in the `comments` association so that the first comment is only dropped if the author was active because in other cases the model will already have excluded it. This primarily affects users in state `suspended` I think.


AntonKhorev:
If we want to be exact about suspended users, should we treat them always like deleted? There were explicit checks for deleted in helper that let suspended through.

PR #5631, opened 2025-02-08
Drop eslint initialiser

tomhughes:
This was presumably required at some point but the i18n files seem to pass eslint fine now and even if they didn't it would probably be better to ignore them in the eslint config.


HolgerJeromin (approved):

AntonKhorev:
Actually i18n files don't pass, they're excluded elsewhere: https://github.com/openstreetmap/openstreetmap-website/blob/4027d0c30abb5c50ac61785a57a8699de000ddd7/lib/tasks/eslint.rake#L13

That was added in https://github.com/openstreetmap/openstreetmap-website/commit/d67b0578511e585fabebeb4800cb1706e63b23fb, after the initializer was added in https://github.com/openstreetmap/openstreetmap-website/commit/9256397e46250352d02d0b7bd814b217b96df45a.

PR #5632, opened 2025-02-08
Set iD map param with array

hlfan:
Splitting parameter population and assignment, enabling a singular later default zoom value definition.
A new take on what was dropped in #5597.

### DIFF
--- a/app/assets/javascripts/edit/id.js.erb
+++ b/app/assets/javascripts/edit/id.js.erb
@@ -1,7 +1,8 @@
 $(document).ready(function () {
-  const id = $("#id-embed");
+  const id = $("#id-embed"),
+        idData = id.data();
 
-  if (id.data("configured") === false) {
+  if (!idData.configured) {
     // eslint-disable-next-line no-alert
     alert(I18n.t("site.edit.id_not_configured"));
     return;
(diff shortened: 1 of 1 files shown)


HolgerJeromin (approved):

app/assets/javascripts/edit/id.js.erb
```diff
@@ -12,17 +12,18 @@ $(document).ready(function () {
...
 
   if (mapParams.object) {
     params.set("id", mapParams.object.type + "/" + mapParams.object.id);
-    if (hashArgs.center) {
-      params.set("map", hashArgs.zoom + "/" + hashArgs.center.lat + "/" + hashArgs.center.lng);
-    }
+    map = [hashArgs.zoom, hashArgs.center?.lat, hashArgs.center?.lng];
   } else if (id.data("lat") && id.data("lon")) {
-    params.set("map", "16/" + id.data("lat") + "/" + id.data("lon"));
+    map = [null, id.data("lat"), id.data("lon")];
```
tomhughes, line 21: Did you intend to change the zoom from 16 to 17 here?

hlfan: Yes, the secondary default value that was introduced in https://github.com/openstreetmap/openstreetmap-website/commit/6494747882358cc1506ae442c89eed3db1de94a4 seemed like an arbitrary deviation IMO.

AntonKhorev:
A stylistic change without clear improvements, not even shorter. Adds:
- uninitialized variable
- null value
- magic array index number

app/assets/javascripts/edit/id.js.erb
HolgerJeromin: We have again both 16 and 17 default zoom level in this PR.

tomhughes (approved):
Thanks for addressing all the review comment - this looks good to me now.

The only thing that slightly confuses me is the need for parentheses around the destructuring on lines 20 and 22 as it seems to work without that when I try it in the node console but eslint does fail to parse it unless you include them!

hlfan:
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) says:
> The parentheses around the assignment statement are required when using object literal destructuring assignment without a declaration.

As for why it works in the console?
¯\\\_(ツ)_/¯

tyrasd:
FYI, this added a small bug, see https://github.com/openstreetmap/iD/issues/10779?reload=1

I really appreciate the cleanup efforts of the iD integration code. :+1:  Nevertheless, it would have been nice to tag me in this (and the other couple of related PRs). That way, I could have given look as well. :blush:

hlfan:
This bug was introduced with [this fix](https://github.com/openstreetmap/openstreetmap-website/compare/f359a73d5b64e046ab7f783aa37b685ebc65c052..5310c5b9dd67f7d416ee778a36f43c21126a1575) to uninitialized variables. Making the remaining `else if` an `if` is the only thing required to fix this.

tyrasd:
> fix to uninitialized variables

yeah, but the previous version of the code used the _uninitialized_ state explicitly to encode the case for when the `#map` parameter should explicitly be absent in the url of the iD iframe. :wink:

hlfan:
The map parameter would always be set. But that was even before the _"stylistic change without clear improvements"_.

PR #5635, opened 2025-02-09
Add button for directions from location

hlfan:
### Description
Adds a button into the directions search form:
| before | after |
|--------|--------|
| before | after |

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -73,6 +73,31 @@ OSM.Directions = function (map) {
     }));
   });
 
+  $(".directions_form .locate").on("click", function () {
+    if ($(".directions_form .locate").hasClass("disabled")) return;
+    if (!navigator.geolocation) {
+      $(".directions_form .locate").addClass("disabled");
+      return;
+    }
+    navigator.geolocation.getCurrentPosition(geoSuccess);
+
+    function geoSuccess(position) {
+      const location = L.latLng(position.coords.latitude, position.coords.longitude),
+            [coordFrom, coordTo] = endpoints.map(ll => ll.latlng);
+      let routeTo;
+      if (coordTo) {
+        routeTo = coordTo.lat + "," + coordTo.lng;
+      } else if (coordFrom) {
+        routeTo = coordFrom.lat + "," + coordFrom.lng;
+        endpoints[0].swapCachedReverseGeocodes(endpoints[1]);
+      }
+      const routeFrom = location.lat + "," + location.lng;
+      OSM.router.route("/directions?" + new URLSearchParams({
+        route: routeFrom + ";" + routeTo
+      }));
+    }
+  });
+
   $(".directions_form .btn-close").on("click", function (e) {
     e.preventDefault();
     $(".describe_location").toggle(!endpoints[1].value);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/index/search.js
```diff
@@ -11,7 +11,7 @@ OSM.Search = function (map) {
     e.preventDefault();
     var query = $(this).closest("form").find("input[name=query]").val();
     if (query) {
-      OSM.router.route("/directions?from=" + encodeURIComponent(query) + OSM.formatHash(map));
+      OSM.router.route("/directions?to=" + encodeURIComponent(query) + OSM.formatHash(map));
```
tomhughes, line 14: This seems like quite a drastic change, and one that's unrelated to the main goal of this pull request?

AntonKhorev:
How is adding another *show location* button better than integrating directions with the existing button?

app/assets/javascripts/index/search.js
hlfan: If I phrase the main goal of this pull request as "I want to get more easily from where I am to where I searched," then this change makes the flow clearer, even though it is not needed when the handler function is implemented the way I propose.
This is also the way it works for every non-OSM map app I tested.

hlfan:
How would you integrate directions with the existing button?
I based this on my local transport authorities' apps' (which use OSM too) design, which I find uses the available space well.

app/assets/stylesheets/common.scss
```diff
@@ -581,6 +581,15 @@ header .search_forms,
   }
 }
 
+.search_forms {
+  .geolocate {
+    transition: filter 0.15s ease-in-out;
```
gravitystorm, line 586: We don't use transitions on any other button content, so I don't think it's necessary here

app/assets/stylesheets/common.scss
```diff
@@ -581,6 +581,15 @@ header .search_forms,
...
+  }
+  .btn:not(:hover) .geolocate {
+    filter: brightness(53%);
```
gravitystorm, line 589: I can see what this is doing, but I don't think it's the right approach. We should either:

* Add icons to the spritesheet, with the colours that we want (for light mode, and dark mode if necessary)
* Or we should take the approach of the "swap directions" button, which has inline SVG and automatic colouring.

app/assets/javascripts/index/search.js
```diff
@@ -11,7 +11,7 @@ OSM.Search = function (map) {
     e.preventDefault();
     const query = $(this).closest("form").find("input[name=query]").val();
     let search = "";
-    if (query) search = "?" + new URLSearchParams({ from: query });
+    if (query) search = "?" + new URLSearchParams({ to: query });
```
gravitystorm, line 14: This suggested behaviour is very common. I know that because personally it usually frustrates me - when I'm using other apps and planning to go from A to B I search for A first, then the site offers direction *to* A, and then I need to swap start/end points. So that's why I know that `search query` -> `routing destination` is much more common than `search query` -> `routing origin`.

Overall I'd be in favour of this change, even if it frustrates me, because it's much more likely that if you search for one place and use the geolocation button, the searched place is going to be the destination.

hlfan: See #5710

PR #5636, opened 2025-02-09
Use subqueries to find nested members more efficiently

tomhughes:
Use subqueries to do the selection in one go in the database which also avoids to deduplicate in rails.

Also pushes all visibility filtering into the query.


AntonKhorev:
Merged, thanks.

PR #5637, opened 2025-02-09
Fix grammar and replace abbreviation in CONTRIBUTING md

SergeCroise:
### Description
- a change that involv**es** ...
- (**for example** using ...

### How has this been tested?
Render the markdown file 
see also:
https://www.merriam-webster.com/dictionary/e.g.#examples

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,7 +87,7 @@ and why it should be the way it is.
 
 ## i18n
 
-If you make a change that involve the locale files (in `config/locales`) then please
+If you make a change that involves the locale files (in `config/locales`) then please
 only submit changes to the `en.yml` file. The other files are updated via
 [Translatewiki](https://translatewiki.net/wiki/Translating:OpenStreetMap) and should
 not be included in your pull request.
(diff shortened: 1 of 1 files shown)


tomhughes:
The `involves` is fine, and correct, but there's really no need for a comma after `e.g.` any more than there would be if you spelled it in full - the dictionary example you pointed at is rather different because it's between two commas as an aside.

What would probably be good is simply to expand `e.g.` as `for example` which is likely to be more readable for people that don't have english as a first language.

CONTRIBUTING.md
```diff
@@ -143,7 +143,7 @@ with making the reviews as straightforward as possible:
   large to review in one sitting, or if changes are requested, then the
   maintainer needs to repeatedly re-read code that has already been considered.
 * The commit history is important. This is a large codebase, developed over many
-  years by many developers. We frequently need to read the commit history (e.g.
+  years by many developers. We frequently need to read the commit history (e.g.,
```
SergeCroise, line 146:
```suggestion
  years by many developers. We frequently need to read the commit history (for example
```

tomhughes: Yes that's fine, can you fold that into the existing commit rather than having me commit it as a fixup? Probably change the title of the commit as well to remove the punctuation reference.

PR #5638, opened 2025-02-09
Convert API element history actions to versions resources

AntonKhorev:
Declares resources on the following paths: `node/:id/history`, `way/:id/history`, `relation/:id/history`.

`history` action becomes a standard `index`.

They were named "history" and I keep this name in paths, but later I'll have to refer to each item in "history", and how do you name such item? Probably "version". Then a collection of these items is "versions", that's why I'm calling the resources "versions".


tomhughes:
Looks good to me, thanks.

PR #5640, opened 2025-02-09
Move eslint ignore rules to the configuration file

tomhughes:
Following on from @AntonKhorev's observation in #5631 this moves eslint ignore rules to the configuration.

Instead of having the rake task select which files to check, we just check the whole tree and have eslint ignore third party files and generated files that we don't control.


AntonKhorev:
Merged, thanks.

PR #5641, opened 2025-02-09
Make API old element show paths resourceful

AntonKhorev:
Declares `versions` resources with `show` action on these API paths: `node/:id/:version`, `way/:id/:version`, `relation/:id/:version`.


tomhughes:
Looks good to me, thanks.

PR #5642, opened 2025-02-10
Get directions using fetch

hlfan:
Start to move from $.ajax to fetch with the thing I recently restructured.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -3,7 +3,7 @@
 //= require_tree ./directions
 
 OSM.Directions = function (map) {
-  var routeRequest = null; // jqXHR object of an ongoing route request or null
+  let controller = null; // the AbortController for the current route request if a route request is in progress
   var chosenEngine;
 
   var popup = L.popup({ autoPanPadding: [100, 100] });
@@ -23,7 +23,7 @@ OSM.Directions = function (map) {
   var endpointDragCallback = function (dragging) {
     if (!map.hasLayer(polyline)) return;
     if (dragging && !chosenEngine.draggable) return;
-    if (dragging && routeRequest) return;
+    if (dragging && controller) return;
 
     getRoute(false, !dragging);
   };
@@ -109,7 +109,7 @@ OSM.Directions = function (map) {
 
   function getRoute(fitRoute, reportErrors) {
     // Cancel any route that is already in progress
-    if (routeRequest) routeRequest.abort();
+    if (controller) controller.abort();
 
     const points = endpoints.map(p => p.latlng);
 
@@ -126,20 +126,8 @@ OSM.Directions = function (map) {
     // again.
     $("#sidebar_content").html($(".directions_form .loader_copy").html());
     map.setSidebarOverlaid(false);
-
-    routeRequest = chosenEngine.getRoute(points, function (err, route) {
-      routeRequest = null;
-
-      if (err) {
-        map.removeLayer(polyline);
-
-        if (reportErrors) {
-          $("#sidebar_content").html("<div class=\"alert alert-danger\">" + I18n.t("javascripts.directions.errors.no_route") + "</div>");
-        }
-
-        return;
-      }
-
+    controller = new AbortController();
+    chosenEngine.getRoute(points, controller.signal).then(function (route) {
       polyline
         .setLatLngs(route.line)
         .addTo(map);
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/index/directions/fossgis_valhalla.js
```diff
@@ -99,22 +99,18 @@
...
-          success: function ({ trip }) {
-            if (trip.status === 0) {
-              callback(false, _processDirections(trip.legs));
-            } else {
-              callback(true);
+        });
+        return fetch(OSM.FOSSGIS_VALHALLA_URL + "?" + query)
+          .then(response => response.json())
+          .then(({ trip }) => {
+            if (trip.status !== 0) {
```
HolgerJeromin, line 106: The callback result can not be used.
```ts
            if (trip.status !== 0) {
              return callback(true);
            }
            return callback(false, _processDirections(trip.legs));
```
or
```ts
            if (trip.status !== 0) {
              callback(true);
              return;
            }
            callback(false, _processDirections(trip.legs));
```

HolgerJeromin:
This pushes json parsing off the js main thread. Hurray

app/assets/javascripts/index/directions/fossgis_valhalla.js
hlfan: Made this change to all error callbacks.

HolgerJeromin (approved):

hlfan:
Actually needs to be abortable...

app/assets/javascripts/index/directions.js
```diff
@@ -126,20 +127,9 @@ OSM.Directions = function (map) {
...
-          $("#sidebar_content").html("<div class=\"alert alert-danger\">" + I18n.t("javascripts.directions.errors.no_route") + "</div>");
-        }
-
-        return;
-      }
-
+    controller = new AbortController();
+    awaitingRoute = true;
+    chosenEngine.getRoute(points, controller.signal).then(function (route) {
       polyline
```
HolgerJeromin, line 133: Add `controller = undefined;` here. Same in `catch`

app/assets/javascripts/index/directions.js
```diff
@@ -212,6 +202,13 @@ OSM.Directions = function (map) {
         map.setSidebarOverlaid(true);
         // TODO: collapse width of sidebar back to previous
       });
+    }).catch(function () {
+      map.removeLayer(polyline);
+      if (reportErrors) {
+        $("#sidebar_content").html("<div class=\"alert alert-danger\">" + I18n.t("javascripts.directions.errors.no_route") + "</div>");
+      }
+    }).finally(function () {
+      awaitingRoute = false;
```
HolgerJeromin, line 211: Add `controller = undefined;` here.

app/assets/javascripts/index/directions.js
```diff
@@ -3,7 +3,8 @@
 //= require_tree ./directions
 
 OSM.Directions = function (map) {
-  var routeRequest = null; // jqXHR object of an ongoing route request or null
+  let awaitingRoute = false; // true if a route request is in progress
```
HolgerJeromin, line 6: Do we really need both? The controller and the boolean?

hlfan:
I still don't know what I need to change in the [testing stub](https://github.com/hlfan/openstreetmap-website/blob/fetch-directions/test/system/directions_test.rb#L56)

HolgerJeromin:
> I still don't know what I need to change in the testing stub

I would guess:

```ruby
  def stub_straight_routing(start_instruction: "Start here", finish_instruction: "Finish there")
    stub_routing <<~CALLBACK
      const distance = points[0].distanceTo(points[1]);
      const time = distance * 30;
      return Promise.resolve({
        line: points,
        steps: [
          [points[0],  8, "<b>1.</b> #{start_instruction}", distance, points],
          [points[1], 14, "<b>2.</b> #{finish_instruction}", 0, [points[1]]]
        ],
        distance,
        time
      });
    CALLBACK
  end

  def stub_routing(callback_code)
    execute_script <<~SCRIPT
      $(() => {
        for (const engine of OSM.Directions.engines) {
          engine.getRoute = (points, signal) => {
              #{callback_code}
          };
        }
      });
    SCRIPT
  end
end

File: app/assets/javascripts/index/directions/fossgis_osrm.js [hunk 5]
```diff
@@ -154,38 +154,31 @@
...
-          url: OSM.FOSSGIS_OSRM_URL + req_path,
-          data,
-          dataType: "json",
-          success: function (response) {
+        return fetch(OSM.FOSSGIS_OSRM_URL + req_path + "?" + query, { signal })
+          .then(response => response.json())
+          .then(response => {
             if (response.code !== "Ok") {
-              return callback(true);
+              throw new TypeError("Response " + response.code);
```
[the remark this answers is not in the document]
hlfan: I just want to return from the function and reject the promise in one go. What I throw matters more to eslint than to me. But I thought I'd be nice for the good of the codebase.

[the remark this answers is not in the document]
tomhughes: Sure, but this doesn't really fit the description of `TypeError` at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError does it? So maybe just use `Error` instead?

tomhughes (approved):
Thanks - this looks good now I think.

PR #5643, opened 2025-02-10
Enable more linting rules

hlfan:
Found some that improve legibility.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -113,13 +113,13 @@ $(document).ready(function () {
    */
   setTimeout(function () {
     $("header").children(":visible").each(function (i, e) {
-      headerWidth = headerWidth + $(e).outerWidth();
+      headerWidth += $(e).outerWidth();
     });
 
     $("body").addClass("compact-nav");
 
     $("header").children(":visible").each(function (i, e) {
-      compactWidth = compactWidth + $(e).outerWidth();
+      compactWidth += $(e).outerWidth();
     });
 
     $("body").removeClass("compact-nav");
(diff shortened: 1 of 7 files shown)


nenad-vujicic (approved):
This looks and works great on my side. It's much more readable to me. Thanks!

app/assets/javascripts/index/note.js
```diff
@@ -55,7 +55,8 @@ OSM.Note = function (map) {
           content.find("#comment-error")
             .text(xhr.responseText)
             .prop("hidden", false)
-            .get(0).scrollIntoView({ block: "nearest" });
+            .get(0)
+            .scrollIntoView({ block: "nearest" });
```
AntonKhorev, line 59: This is a misleading change. `scrollIntoView` is not called on the same object as `text` and `prop`. You'd have to assign the result of `get` to some variable.

hlfan: Removed the `newline-per-chained-call` rule.

tomhughes (commented):
Why is the title of the second commit "Use less linting rules" which it's actually adding more linting rules?

hlfan:
In https://github.com/openstreetmap/openstreetmap-website/commit/2602d1888b7c3dfaba91804b133a4b3b6d28543f, I added six rules. However, with https://github.com/openstreetmap/openstreetmap-website/commit/63362cce7bac4e4e60f1c2c3ee888d311d029c3d, I only added five, so I'm adding fewer linting rules than I previously did.

tomhughes:
Sure, but we want the commit description to describe what is happening relative to the code base it's being applied to, not relative to some previous version of this pull request.

hlfan:
Does this work for you?

tomhughes:
Yes that looks better, thanks.

PR #5645, opened 2025-02-10
Create integration tests for element versions

AntonKhorev:
The only non-resourceful thing in the elements API is `redact` action. However before I can do anything with it I need to take care of tests.

You'd expect controller action tests to first set up the state of the elements, then perform the request and finally check the response and the new state. Unfortunately tests for redacting/unredacting element versions don't do that. Instead of checking the state after the action they do new show/index requests and check responses to those. This is double-checking show/index actions for redacted/unredacted states which is already done. But what's worse, if I'm to move (un)redacting to a different controller, the test would check actions of different controllers. They'll become integration tests.

But wait, some would-be integration tests already exist for different actions. You can see "FIXME: Move this test to being an integration test" on `test_show` in `old_nodes_controller_test.rb`. And you can see `with_controller` inside the tests. Also there's a similar situation in old ways controller.

This PR moves `test_show` tests to integration tests and replaces them with tests that just check the show action.

### DIFF
--- a/test/controllers/api/old_nodes_controller_test.rb
+++ b/test/controllers/api/old_nodes_controller_test.rb
@@ -70,7 +70,35 @@ module Api
                     "redacted node #{node_v1.node_id} version #{node_v1.version} shouldn't be present in the history, even when logged in."
     end
 
-    # TODO: test_show
+    def test_show
+      node = create(:node, :version => 2)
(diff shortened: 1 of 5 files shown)


tomhughes:
Looks good to me, thanks.

PR #5647, opened 2025-02-10
Fix remaining key ordering issues in eslint configuration

tomhughes:
Fixes remaining key ordering issues in eslint configuration and converts the warning to an error.

### DIFF
--- a/config/eslint.js
+++ b/config/eslint.js
@@ -17,23 +17,20 @@ module.exports = [
     ]
   },
   {
-    plugins: {
-      "@stylistic": stylisticJs
-    },
     languageOptions: {
       ecmaVersion: 2021,
-      sourceType: "script",
       globals: {
         ...globals.browser,
         ...globals.jquery,
         Cookies: "readonly",
         I18n: "readonly",
         L: "readonly",
-        OSM: "writable",
         Matomo: "readonly",
+        OSM: "writable",
         Turbo: "readonly",
         updateLinks: "readonly"
-      }
+      },
+      sourceType: "script"
     },
     linterOptions: {
       // The "unused disable directive" is set to "warn" by default.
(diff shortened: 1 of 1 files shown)


hlfan (approved):

config/eslint.js
```diff
@@ -166,10 +167,10 @@ module.exports = [
     files: ["config/eslint.js"],
     languageOptions: {
       ecmaVersion: 2019,
-      sourceType: "commonjs",
       globals: {
         ...globals.commonjs
-      }
+      },
+      sourceType: "commonjs"
     }
```
HolgerJeromin, line 174: If you are willing to sort everything (not sure if this is good to read) this should be done here:
```
rules: {eslint sort-keys: ["error"]}

Comment (HolgerJeromin) [reply config/eslint.js:174]: Ah, I see. We can [ignore some rules in some lines](https://eslint.org/docs/latest/use/configure/rules#disabling-rules) with eslint. But not activate a rule within a range.
So yes, current master warns, so resorting properties is probably the best thing we can do, if we want to force the rules to be sorted...

Comment (tomhughes) [reply config/eslint.js:174]: Yes I didn't particularly want to sort them but I didn't want seven new warnings either...

Comment (HolgerJeromin) [reply config/eslint.js:174]: Yeah, activate for this file in config and disable in most other parts is a bit cumbersome:
```
/* eslint-disable sort-keys */
... start of the json config
/* eslint-enable sort-keys */
...rules sorted. yeah
/* eslint-disable sort-keys */
... rest of the config

hlfan: I just stumbled over something I didn't see when introducing this rule!

With `eslint sort-keys: ["error", "asc", {minKeys: 40}]` only the long rule list will get flagged for being unsorted, and the current 10 - 97 key count difference is also quite forgiving.

tomhughes:
Superseded by #5651.

PR #5649, opened 2025-02-11
Refactor API old element redaction tests

AntonKhorev:
This PR changes redact/unredact tests for the reasons stated in #5645.

Tests will have this structure:
```ruby
      node = create(:node, :with_history, :version => 2)
      old_node = node.old_nodes.find_by(:version => 1)
      redaction = create(:redaction)
      auth_header = ...

      post node_version_redact_path(*old_node.id), :params => { :redaction => redaction.id }, :headers => auth_header

      assert_response ...
      assert ... old_node.reload.redaction
```

Element redaction property is checked after every response. Previously it wasn't checked after failed responses.


tomhughes (commented):
Why does the last commit only do nodes and ways and not relations?

PR #5650, opened 2025-02-11
Adds storing note's properties to notes

nenad-vujicic:
### Description

Adds storing note's description, author_id and author_ip to notes table too when creating new note.

### How has this been tested?

Automatic tests, manual testing before and after data migration.

### DIFF
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -81,19 +81,22 @@ module Api
       # Extract the arguments
       lon = OSM.parse_float(params[:lon], OSM::APIBadUserInput, "lon was not a number")
       lat = OSM.parse_float(params[:lat], OSM::APIBadUserInput, "lat was not a number")
-      comment = params[:text]
+      description = params[:text]
+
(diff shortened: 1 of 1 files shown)


app/controllers/api/notes_controller.rb
```diff
@@ -379,17 +382,30 @@ def bbox_condition(notes)
       end
     end
 
+    ##
+    # Get author's information (for logged in users - user_id, for logged out users - IP address)
+    def author_info
+      author = current_user if scope_enabled?(:write_notes)
```
tomhughes, line 388: I realise this is just copying the old code but I don't think we need to do this in this way now - if the scope is enabled then there must be a current user so we can just make the `if` below do the scope test and then use `current_user.id` to set the user id?

tomhughes (changes requested):

app/controllers/api/notes_controller.rb
nenad-vujicic: Fixed (and pushed)! Thanks!

tomhughes (approved):

PR #5651, opened 2025-02-11
Force only big objects in eslint configuration to be sorted

hlfan:
A less intrusive variant of #5647


tomhughes:
Looks good to me, thanks.

PR #5652, opened 2025-02-11
Split directions engine select into modes and providers

### LINKED ISSUE #5629: Split the combined directions engines translations into routers and modes
### Problem

Only 46 of the 78 YAML files with engine translations have all nine engines translated. Thus, terms will be mixed-matched for 32 languages.
Take [Afrikaans](/openstreetmap/openstreetmap-website/blob/master/config/locales/af.yml#L1975), where this messes up the sorting as well:
Image

This also makes the translations more prone to [errors](/openstreetmap/openstreetmap-website/blob/master/config/locales/da.yml#L3308) and [inconsistencies](/openstreetmap/openstreetmap-website/blob/master/config/locales/eu.yml#L2801):
Image
Image

### Description

Split up the nine combined translations into seven separate translations,
* three for the modes,
* three for the routers (usually don't need to be set) and
* one template (only needs to be overwritten in special cases, defaults to "%{modes} (%{routers})").

While most translations could be automatically transferred, I don't know how this could be done with translate-wiki.

### Screenshots

_No response_

hlfan:
Attempt to fix #5629.
Splits the directions engine select into a radio button group for modes and a smaller select for routing providers.
| before | after |
|-|-|
| before | after |

Includes tooltips for the mode radio button group:
image

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -41,17 +41,8 @@ OSM.Directions = function (map) {
   const expiry = new Date();
   expiry.setYear(expiry.getFullYear() + 10);
 
-  const engines = OSM.Directions.engines;
-
-  engines.sort(function (a, b) {
-    return a.localeId.localeCompare(b.localeId);
-  });
-
+  const modeGroup = $(".routing_modes");
   const select = $("select.routing_engines");
-
-  engines.forEach(function (engine, i) {
-    select.append("<option value='" + i + "'>" + engine.localeId + "</option>");
-  });
 
   $(".directions_form .reverse_directions").on("click", function () {
     const coordFrom = endpoints[0].latlng,
@@ -97,10 +88,32 @@ OSM.Directions = function (map) {
   }
 
   function setEngine(id) {
-    const index = engines.findIndex(engine => engine.id === id);
-    if (index < 0) return;
-    chosenEngine = engines[index];
-    select.val(index);
+    const engines = OSM.Directions.engines;
+    const desired = engines.find(engine => engine.id === id);
+    if (!desired || (chosenEngine && chosenEngine.id === id)) return;
+    chosenEngine = desired;
+
+    const modes = engines
+      .filter(engine => engine.provider === chosenEngine.provider)
+      .map(engine => engine.mode);
+    modeGroup
+      .find("input[id]")
+      .prop("disabled", function () {
+        return !modes.includes(this.id);
+      })
+      .prop("checked", function () {
+        return this.id === chosenEngine.mode;
+      });
+
+    const providers = engines
+      .filter(engine => engine.mode === chosenEngine.mode)
+      .map(engine => engine.provider);
+    select
+      .find("option[value]")
+      .prop("disabled", function () {
+        return !providers.includes(this.value);
+      });
+    select.val(chosenEngine.provider);
   }
 
   function getRoute(fitRoute, reportErrors) {
@@ -227,8 +240,14 @@ OSM.Directions = function (map) {
   setEngine("fossgis_osrm_car");
   setEngine(Cookies.get("_osm_directions_engine"));
 
+  modeGroup.on("change", "input[name='modes']", function (e) {
+    setEngine(chosenEngine.provider + "_" + e.target.id);
+    Cookies.set("_osm_directions_engine", chosenEngine.id, { secure: true, expires: expiry, path: "/", samesite: "lax" });
+    getRoute(true, true);
+  });
+
   select.on("change", function (e) {
-    chosenEngine = engines[e.target.selectedIndex];
+    setEngine(e.target.value + "_" + chosenEngine.mode);
     Cookies.set("_osm_directions_engine", chosenEngine.id, { secure: true, expires: expiry, path: "/", samesite: "lax" });
     getRoute(true, true);
   });
@@ -310,11 +329,6 @@ OSM.Directions.engines = [];
 OSM.Directions.addEngine = function (engine, supportsHTTPS) {
   if (document.location.protocol === "http:" || supportsHTTPS) {
     engine.id = engine.provider + "_" + engine.mode;
-    engine.localeId = `${
-      I18n.t("site.search.modes." + engine.mode)
-    } (${
-      I18n.t("site.search.providers." + engine.provider)
-    })`;
     OSM.Directions.engines.push(engine);
   }
 };
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -310,7 +310,11 @@ OSM.Directions.engines = [];
 OSM.Directions.addEngine = function (engine, supportsHTTPS) {
   if (document.location.protocol === "http:" || supportsHTTPS) {
     engine.id = engine.provider + "_" + engine.mode;
-    engine.localeId = I18n.t("javascripts.directions.engines." + engine.id);
+    engine.localeId = `${
+      I18n.t("site.search.modes." + engine.mode)
+    } (${
+      I18n.t("site.search.providers." + engine.provider)
+    })`;
     OSM.Directions.engines.push(engine);
   }
 };
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -44,15 +44,13 @@ OSM.Directions = function (map) {
   const engines = OSM.Directions.engines;
 
   engines.sort(function (a, b) {
-    const localised_a = I18n.t("javascripts.directions.engines." + a.id),
-          localised_b = I18n.t("javascripts.directions.engines." + b.id);
-    return localised_a.localeCompare(localised_b);
+    return a.localeId.localeCompare(b.localeId);
   });
 
   const select = $("select.routing_engines");
 
   engines.forEach(function (engine, i) {
-    select.append("<option value='" + i + "'>" + I18n.t("javascripts.directions.engines." + engine.id) + "</option>");
+    select.append("<option value='" + i + "'>" + engine.localeId + "</option>");
   });
 
   $(".directions_form .reverse_directions").on("click", function () {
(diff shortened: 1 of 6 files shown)


tomhughes:
You should never edit any translation here (other than `en.yml` obviously) so you need to remove all those changes and let Translatewiki handle them.

hlfan:
Doesn't this mean all the translations will have to start from scratch?
Then it could take years until some translations will be back.

config/locales/en.yml
```diff
@@ -3207,16 +3207,15 @@ en:
...
-        fossgis_osrm_bike: "Bicycle (OSRM)"
-        fossgis_osrm_car: "Car (OSRM)"
-        fossgis_osrm_foot: "Foot (OSRM)"
-        graphhopper_bicycle: "Bicycle (GraphHopper)"
-        graphhopper_car: "Car (GraphHopper)"
-        graphhopper_foot: "Foot (GraphHopper)"
-        fossgis_valhalla_bicycle: "Bicycle (Valhalla)"
-        fossgis_valhalla_car: "Car (Valhalla)"
-        fossgis_valhalla_foot: "Foot (Valhalla)"
+        fossgis_osrm: "OSRM"
```
HolgerJeromin, line 3215: does these engine names need a translation key?

hlfan: If the only possible solution is to discard every related translation, I would not let the engines be translated.
However, I understand why these were transliterated for non-Latin scripts; my goal here was to keep as many translations as possible, including the engine names.

tomhughes:
If you want this to be considered for merging then you should remove the draft flag.

That said I'm not sure I really see the point of it - is this really just to reduce the number of translations from 9 to 6 so saving a total of 3 translations? All while introducing new "lego" to the translations though admittedly one that is probably relatively harmless.

hlfan:
The issues laid out in #5629 aren't page-breaking, but IMO changing to a system that removes the possibility of inconsistencies entirely is worth the lego.
Usually, only 3 translations for the modes are needed, only the non-Latin scripts need the graceful fallback.
And it's a step in the direction of #3123 if you wanna see it that way.

AntonKhorev:
Why would you do lego instead of two selects for mode and provider?

hlfan:
I wanted to split the translations first and move to an engine select and transport mode radio toggle buttons later, as restructuring the engines in an intermediate step seemed the best way to go about it.
Doing it as three smaller PRs instead of a big PR because I don't think it's necessary.

app/assets/javascripts/index/directions.js
```diff
@@ -96,15 +91,36 @@ OSM.Directions = function (map) {
...
+  function setEngine(id) {
+    const engines = OSM.Directions.engines;
+    const desired = engines.find(engine => engine.id() === id);
+    if (!desired || (chosenEngine && chosenEngine.id() === id)) return;
+    chosenEngine = desired;
+
+    const modes = engines
+      .filter(engine => engine.provider === chosenEngine.provider)
+      .map(engine => engine.mode)
+      .sort((a, b) => I18n.t("javascripts.directions.modes." + a).localeCompare(I18n.t("javascripts.directions.modes." + b)));
```
AntonKhorev, line 103: What's the point of this sort?

app/assets/javascripts/index/directions.js
```diff
@@ -96,15 +91,36 @@ OSM.Directions = function (map) {
...
+    modeGroup.html("");
+    for (const mode of new Set(modes)) {
+      modeGroup.append(`<input type="radio" class="btn-check" name="modes" id="${mode}" autocomplete="off">`);
+      modeGroup.append(`<label class="btn btn-outline-secondary px-2" for="${mode}" title="${
+        I18n.t("javascripts.directions.modes." + mode)
+      }"><svg class="d-block" width="16" height="16" fill="currentColor"><path d="${modeIconPaths[mode]}"></path></svg></label>`);
+    }
```
AntonKhorev, line 110: Are you sure you want to redraw the buttons instead of adding them once in the html template? Even if you plan for routing engines that don't support every mode, would you remove the mode button rather than disable it?

app/assets/javascripts/index/directions.js
```diff
@@ -96,15 +91,36 @@ OSM.Directions = function (map) {
...
+    $(".routing_modes input#" + chosenEngine.mode).prop("checked", true);
+
+    const providers = engines
+      .filter(engine => engine.mode === chosenEngine.mode)
+      .map(engine => engine.provider)
+      .sort((a, b) => I18n.t("javascripts.directions.providers." + a).localeCompare(I18n.t("javascripts.directions.providers." + b)));
+    select.html("");
+    for (const provider of new Set(providers)) {
+      select.append(`<option value="${provider}">${
+        I18n.t("javascripts.directions.providers." + provider)
```
AntonKhorev, line 120: I know this *let's interpolate unescaped html from external source* was in the original code, but you don't have to replicate it.

app/assets/javascripts/index/directions.js
hlfan: The options of the select currently get sorted so there's a different order depending on the language.
In english it's 🚲-🚙-🚶, in german  🚙-🚲-🚶.
Would you like to specify a fixed order for all languages?

HolgerJeromin: Many maps are specific for one transport mode. 
So having a fixed sort order can be problematic (osm data is also good for foot stuff).
That being said, alphabetically is probably the most the most neutral one...

app/assets/javascripts/index/directions.js
hlfan: Disabling the buttons and the selects options seems better, disabling all in the template and enabling each in js when supported.

However, I don't want to move too much to be hardcoded like it was already criticized in https://github.com/openstreetmap/openstreetmap-website/pull/3400#issuecomment-1057225749

app/views/layouts/_search.html.erb
```diff
@@ -66,7 +66,31 @@
...
+               { :id => "bicycle", :icon => "M4 4.5a.5.5 0 0 1 .5-.5H6a.5.5 0 0 1 0 1v.5h4.14l.386-1.158A.5.5 0 0 1 11 4h1a.5.5 0 0 1 0 1h-.64l-.311.935.807 1.29a3 3 0 1 1-.848.53l-.508-.812-2.076 3.322A.5.5 0 0 1 8 10.5H5.959a3 3 0 1 1-1.815-3.274L5 5.856V5h-.5a.5.5 0 0 1-.5-.5m1.5 2.443-.508.814c.5.444.85 1.054.967 1.743h1.139zM8 9.057 9.598 6.5H6.402zM4.937 9.5a2 2 0 0 0-.487-.877l-.548.877zM3.603 8.092A2 2 0 1 0 4.937 10.5H3a.5.5 0 0 1-.424-.765zm7.947.53a2 2 0 1 0 .848-.53l1.026 1.643a.5.5 0 1 1-.848.53z" },
+               { :id => "foot", :icon => "M9.5 1.5a1.5 1.5 0 11-3 0 1.5 1.5 0 013 0M6.44 3.752A.75.75 0 017 3.5h1.445c.742 0 1.32.643 1.243 1.38l-.43 4.083a1.8 1.8 0 01-.088.395l-.318.906.213.242a.8.8 0 01.114.175l2 4.25a.75.75 0 11-1.357.638l-1.956-4.154-1.68-1.921A.75.75 0 016 8.96l.138-2.613-.435.489-.464 2.786a.75.75 0 11-1.48-.246l.5-3a.75.75 0 01.18-.375l2-2.25zm-.19 7.993v-1.418l1.204 1.375.261.524a.8.8 0 01-.12.231l-2.5 3.25a.75.75 0 11-1.19-.914zm4.22-4.215-.494-.494.205-1.843.006-.067 1.124 1.124h1.44a.75.75 0 010 1.5H11a.75.75 0 01-.531-.22Z" }
```
hlfan, line 73: It would be nice if these paths can be moved into the SVG defs of #4901 later

app/assets/javascripts/index/directions.js
tomhughes: I'm not sure why the current code sorts them but it seems wrong to me - there's no logical reason why the order in which they are shown should depend on the language.

app/views/layouts/_search.html.erb
```diff
@@ -65,8 +65,36 @@
...
+               .sort_by { |mode| mode[:title] }
+               .each do |mode| %>
+            <input type="radio" class="btn-check" name="modes" id="<%= mode[:id] %>" autocomplete="off" disabled>
+            <label class="btn btn-outline-secondary px-2" for="<%= mode[:id] %>" title="<%= mode[:title] %>">
+              <svg class="d-block" width="16" height="16" fill="currentColor">
+                <use href="#icon-<%= mode[:id] %>"></use>
+              </svg>
+            </label>
+          <% end %>
+        </div>
```
tomhughes, line 89: This goes to the previous discussion about sorting, but the above is far more complicated than it really needs to be, and all so we can put the icons in a different order for different languages which seems quite mad to me.

I would just make this a fixed order then we don't need to predefined the icons in a hidden block so that we can sort them.

app/views/layouts/_search.html.erb
```diff
@@ -65,8 +65,36 @@
...
+        <select class="routing_engines form-select py-1 px-2" name="routing_engines">
+          <% %w[fossgis_osrm graphhopper fossgis_valhalla]
+               .map { |id| { :id => id, :title => t("site.search.providers.#{id}") } }
+               .sort_by { |engine| engine[:title] }
```
tomhughes, line 93: I guess this does need some sort of order to be applied because the order in which the engines are registered is probably not well defined but I'd be inclined to sort by the provider ID rather than the translated name so that the order is consistent across languages.

app/assets/javascripts/index/directions.js
hlfan: Well then which order should they be shown in?

tomhughes:
Thanks for the ordering change - can we just squash those last two commits together? There doesn't seem to be any point doing it one way and then fixing it to do it a different way.

hlfan:
Squashed. Just wanted to have https://github.com/openstreetmap/openstreetmap-website/commit/0a0835dbfa38bef1c1e19d852e349e13a6753e7c not yet mixed in in case of a revision.

tomhughes:
Thanks. I think this looks good now.

hlfan:
I just found a couple things to remove, hold on

tomhughes:
I had already merged this before that last change - can you open a new PR for it?

hlfan:
See #5729

config/locales/en.yml
hlfan: Well, I've looked at Translatewiki; sadly, it didn't sort itself out.

PR #5654, opened 2025-02-11
Add pluralizers to I18n

AntonKhorev:
Required to correctly translate tooltips in #5402.

Actually it's already required for "Show My Location" tooltips.

Adds pluralizers from [make-plural](https://github.com/eemeli/make-plural/) as recommended by [i18n-js docs](https://fnando.github.io/i18n/v4.5.1/classes/Pluralization.html). Not exactly like on that page because we're still using an older version of i18n-js which is [not compatible with the new one](https://github.com/fnando/i18n-js/blob/main/MIGRATING_FROM_V3_TO_V4.md).

I'm not sure if pluralization in i18n-js v3 is documented anywhere. [This comment](https://github.com/fnando/i18n-js/issues/133#issuecomment-41999391) has an outdated link to the source code, but I think the entire repository was overwritten since.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -11,15 +11,23 @@
 //= require leaflet.zoom
 //= require leaflet.locationfilter
 //= require i18n
+//= require make-plural/cardinals
 //= require matomo
 //= require richtext
 
 {
   const application_data = $("head").data();
+  const locale = application_data.locale;
 
   I18n.default_locale = OSM.DEFAULT_LOCALE;
-  I18n.locale = application_data.locale;
+  I18n.locale = locale;
   I18n.fallbacks = true;
+
+  // '-' are replaced with '_' in https://github.com/eemeli/make-plural/tree/main/packages/plurals
+  const pluralizer = plurals[locale.replace(/\W+/g, "_")] || plurals[locale.split("-")[0]];
+  if (pluralizer) {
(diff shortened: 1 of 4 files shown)


app/views/layouts/_head.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= tag.head :data => application_data do %>
   <%= render :partial => "layouts/meta" %>
   <%= javascript_include_tag "turbo", :type => "module" %>
+  <%= javascript_include_tag "make-plural/cardinals" %>
```
tomhughes, line 4: What's the reason for loading this as a separate file rather than requiring it in `application.js`?

AntonKhorev: Name collisions. They have `const a`, `const b` etc defined in the global scope. When I added `make-plural/cardinals` as `//= require ...` in `application.js` it was picking wrong pluralization rules.

tomhughes: Obviously that global namespace pollution is poor but I don't see why it would break anything, and having tried it myself I can't reproduce any problem once the assets are properly rebuilt.

AntonKhorev: When I started doing this, we were still using vars. It could be var redeclarations, which are no longer happening, because now we're using let/const.

AntonKhorev: Changed to `//= require`

tomhughes:
Looks good to me now, thanks.

PR #5655, opened 2025-02-11
Added prefix for busways in en.yml

WillBerg510:
### Description
Fixed issue #5258 by adding `busway: "Busway"` into the highway section of en.yml. When querying features, busways no longer show the prefix "Yes".

image


tomhughes:
Looks good to me, thanks.

PR #5656, opened 2025-02-11
API element resources - redaction paths

AntonKhorev:
The last non-resourceful thing in the elements api are redactions. They are mapped to `redact` actions and `redact` is not a standard name. One option would be to map them to `destroy` actions of element versions, but they can also remove redactions and removing redactions with `destroy` is going to look strange. Instead this PR adds `redaction` resources with `create` and `destroy` actions. `create` adds a redaction to an element version, `destroy` removes a redaction. For backwards compatibility `create` also removes redactions when none is supplied in parameters because the old `node/:node_id/:version/redact`, `way/:way_id/:version/redact` and `relation/:relation_id/:version/redact` are also mapped to `create` action on `redaction` resources.

All nonstandard action names should be gone from the elements api after this PR.


tomhughes:
The code looks fine to me - the only real questions are around the API change and the slightly weird things needed to keep the legacy endpoint working so I'll give @gravitystorm a chance to comment before I merge it.

gravitystorm:
I think my main comment here is about the resource naming. It doesn't feel quite right to say that we are creating a "redaction" since we need to pass a redaction_id while doing so. It feels to me like the resource name is not quite right, but I can't find a better name. I've done some research to see if there is an existing term for "individual parts of a redaction" but with no success. Possible terms would include things like RedactionEntry, RedactionMember, etc, or if it makes more sense to treat each individual thing as a "Redaction", then we could rename redactions to "RedactionCollection" or similar.

But this is all too much work I think, and not worth it. So perhaps we should stick with what we have here, even despite the names being not quite right.

Using the code for a different action (i.e. calling the `destroy` action from within the `create` action) feels weird, but I guess it works, given the lack of templates etc. An alternative would be to have a private method that is called from both actions, which might be more conventional?

AntonKhorev:
Added a parameter on old routes.

tomhughes:
Thanks for the update - this looks good to me now.

PR #5658, opened 2025-02-12
Stop using session flash to communicate with callbacks

tomhughes:
Using for this is silly, and has weird effects as observed in the fact that I had to fix a test which had accidentally relied on fetching user details from the API causing a session to spring into existence because it wrote to flash.


gravitystorm:
Merged, thanks!

PR #5662, opened 2025-02-12
Use fetch to make overpass requests for the query tool

tomhughes:
Switches query tool to use `fetch` for it's overpass requests.

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -153,20 +153,24 @@ OSM.Query = function (map) {
     $ul.empty();
     $section.show();
 
-    if ($section.data("ajax")) {
-      $section.data("ajax").abort();
+    if ($section.data("controller")) {
+      $section.data("controller").abort();
     }
 
-    $section.data("ajax", $.ajax({
-      url: url,
+    $section.data("controller", new AbortController());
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/index/query.js
```diff
@@ -221,16 +225,15 @@ OSM.Query = function (map) {
...
         $section.find(".loader").hide();
 
         $("<li>")
           .addClass("list-group-item")
-          .text(I18n.t("javascripts.query." + status, { server: url, error: error }))
+          .text(I18n.t("javascripts.query.error", { server: url, error: error.message }))
           .appendTo($ul);
-      }
-    }));
+      });
```
HolgerJeromin, line 236: Missing controller cleanup?

tomhughes: Added, though I don't believe it really matters much - it just means on the next query we'll send an abort on a controller that has nothing listening to it's signal.

HolgerJeromin: Thanks. 
Will now result in a bit to feed the garbage collector (potentially the whole fetch tree with data).

tomhughes:
Superseded by #5669.

PR #5666, opened 2025-02-13
Comment changesets scope

### LINKED ISSUE #2631: Create new oauth permission to post changeset comments
Currently we require the "modify the map" permission" to post comments to changesets through the API. It would be better if we could separate it in a new permission. That way, tools like OSMCha that allow users to post comments to changesets can ask the user only for that permission.

Related to #2629 .

AntonKhorev:
This PR introduces `write_changeset_comments` OAuth scope that allows creating changeset comments (and hiding/unhiding them for moderators). This is a subset of what `write_api` allows.

`write_api` still works as before looking from outside. Internally it's translated to `write_changeset_comments` and `write_map`. `write_map` is not exposed anywhere, it's only known to `ApiController` and `ApiAbility`. Doorkeeper doesn't know about it. Later it can be added as a selectable scope, but that would require updating cgimap.

Fixes #2631.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -87,12 +87,13 @@ en:
         url: Main Application URL (Required)
         callback_url: Callback URL
         support_url: Support URL
-        allow_read_prefs:  read their user preferences
+        allow_read_prefs: read their user preferences
         allow_write_prefs: modify their user preferences
         allow_write_diary: create diary entries and comments
-        allow_write_api:   modify the map
-        allow_read_gpx:    read their private GPS traces
-        allow_write_gpx:   upload GPS traces
+        allow_write_api: modify the map
+        allow_write_changeset_comments: comment on changesets
+        allow_read_gpx: read their private GPS traces
+        allow_write_gpx: upload GPS traces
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -87,12 +87,13 @@ en:
...
         support_url: Support URL
-        allow_read_prefs:  read their user preferences
+        allow_read_prefs: read their user preferences
         allow_write_prefs: modify their user preferences
         allow_write_diary: create diary entries and comments
-        allow_write_api:   modify the map
-        allow_read_gpx:    read their private GPS traces
-        allow_write_gpx:   upload GPS traces
+        allow_write_api: modify the map
+        allow_write_changeset_comments: comment changesets
```
tomhughes, line 94: Can we make this `comment on changesets` please?

tomhughes (changes requested):

config/locales/en.yml
AntonKhorev: changed to "comment on changesets"

tomhughes (approved):
Thanks - this looks good now.

Issue #5619, opened 2025-02-05
Rename database tables


gravitystorm:
For historical reasons, we've ended up with quite a few database tables names that aren't aligned with the model names. I think in every case we prefer the model names (since these are easier to change, they tend to have been changed already without renaming the underlying table).

```bash
andy@denali:~/src/openstreetmap-website$ grep -rn 'self\.table_name' app/models/
app/models/old_way.rb:24:  self.table_name = "ways"
app/models/old_way_tag.rb:16:  self.table_name = "way_tags"
app/models/way_tag.rb:15:  self.table_name = "current_way_tags"
app/models/tracepoint.rb:26:  self.table_name = "gps_points"
app/models/trace.rb:31:  self.table_name = "gpx_files"
app/models/way_node.rb:20:  self.table_name = "current_way_nodes"
app/models/relation_member.rb:21:  self.table_name = "current_relation_members"
app/models/node.rb:31:  self.table_name = "current_nodes"
app/models/way.rb:26:  self.table_name = "current_ways"
app/models/old_way_node.rb:20:  self.table_name = "way_nodes"
app/models/old_relation_tag.rb:16:  self.table_name = "relation_tags"
app/models/old_relation.rb:24:  self.table_name = "relations"
app/models/relation.rb:26:  self.table_name = "current_relations"
app/models/old_relation_member.rb:22:  self.table_name = "relation_members"
app/models/node_tag.rb:15:  self.table_name = "current_node_tags"
app/models/old_node.rb:30:  self.table_name = "nodes"
app/models/relation_tag.rb:15:  self.table_name = "current_relation_tags"
app/models/follow.rb:22:  self.table_name = "friends"
app/models/tracetag.rb:20:  self.table_name = "gpx_file_tags"
app/models/old_node_tag.rb:16:  self.table_name = "node_tags"
```

We might also want to rename certain tables that don't currently have a model, such as `changeset_subscribers` (to e.g. `changeset_subscriptions`) but I haven't made a list of these.

Having these mismatched table/model names doesn't provide any benefits to anyone, as far as I can tell. It also makes it harder for new developers to get started, since they might look at the database structure (e.g. #5308) and then get confused when they can't find the corresponding models. In particular, I found it very confusing when I first started, since we have a table called "nodes" and a "Node" model, but these do not correspond with each other!

So my proposal is to rename the tables to match the model names.

[...]

PR #5667, opened 2025-02-13
Copies notes details from first comments to notes table

nenad-vujicic:
### Description

PR adds migration script for copying notes descriptions, user_ids and user_ips from notes first comments to notes table records. In 2 commits we propose 2 variants of migration script (not sure which is more appropriate for the situation because of large number of notes):

1. basic version - for every `batch of notes`, dynamically generates SQL query for copying values and executes it (simpler to understand, ..).
2. enhanced version - same as previous but keeps track about notes for which migration failed (for some reason?) and retries again later (more robust, ..)

### How has this been tested?

Automated and manual tests before and after data migrations (both basic and enhanced one)


tomhughes:
I think this will do the job, obviously with batches of note IDs in the CTE expression:

```sql
WITH first_comment AS(
  SELECT DISTINCT ON (note_id) *
  FROM note_comments
  WHERE note_id BETWEEN 1000 AND 2000
  ORDER BY note_id, id
)
UPDATE notes
SET description = first_comment.body,
    user_id = first_comment.author_id,
    user_ip = first_comment.author_ip
FROM first_comment
WHERE first_comment.note_id = notes.id
  AND first_comment.event = 'opened';
```

nenad-vujicic:
> I think this will do the job, obviously with batches of note IDs in the CTE expression:
> 

Thanks! I think I fixed it (with the latest push). Here is the code:
```ruby
Note.in_batches(:of => 1000) do |notes|
  note_ids = notes.pluck(:id)

  sql_query = <<-SQL.squish
    WITH first_comment AS(
      SELECT DISTINCT ON (note_id) *
      FROM note_comments
      WHERE note_id BETWEEN #{note_ids.min} AND #{note_ids.max}
      ORDER BY note_id, id
    )
    UPDATE notes
    SET description = first_comment.body,
        user_id = first_comment.author_id,
        user_ip = first_comment.author_ip
    FROM first_comment
    WHERE first_comment.note_id = notes.id
      AND first_comment.event = 'opened';
  SQL

  ActiveRecord::Base.connection.execute(sql_query)
end
```

gravitystorm:
In my own projects, we use a different approach for data migrations.

* First, we avoid using rails built-in database migrations. These are great for structure changes, but not for data changes, since they can't be tested. And there's often some edge cases that we want to have confidence about before we run the data migration and tests are the foundation for confidence.
* We create a job to migrate the data, so that it runs in a background process. We have a rake task that simply queues the job and nothing else. This is because rake tasks are convenient for us to run, but again are very hard to test. Jobs are easy to test.
* We have a separate queue for slow jobs, with a separate set of workers, so that if it takes e.g. 10min to run we are still sending out confirmation emails in the meantime.
* For the task itself, we write it in a way that it can be retried multiple times. So for example we only select the objects that have not yet been migrated (e.g. `Note.where(description: nil).in_batches`) so that if the job is retried, it starts from where it left off, not from the start again.

We do this because in so many cases either we forgot about an edge case but picked it up in testing, or there turns out to be some unexpected data that messes things up (an example would be a note with no comments, even if there's no way that could have happened, in theory, somehow it always does). It also avoid any worries about locks, long-running transactions, transactions with large amounts of uncommitted changes, timeouts, network interruptions, db process restarts and so on.

I say all this to provide an alternative approach that I know works. But in this case, since @tomhughes will be running the actual migration process, whatever approach he prefers is what we should do!

db/migrate/20250212160355_backfill_note_descriptions.rb
```diff
@@ -0,0 +1,30 @@
...
+          WHERE note_id BETWEEN #{note_ids.min} AND #{note_ids.max}
+          ORDER BY note_id, id
+        )
+        UPDATE notes
+        SET description = first_comment.body,
+            user_id = first_comment.author_id,
+            user_ip = first_comment.author_ip
+        FROM first_comment
+        WHERE first_comment.note_id = notes.id
+          AND first_comment.event = 'opened';
```
gravitystorm, line 24: Why is this restricted to only comments with event = 'opened'? That's not the same logic as is currently used for creating note descriptions using comments - it uses `all_comments.first` without any similar restriction.

gravitystorm (commented):
I would also interested in what the next step will be. How do you propose removing the first comments, when they are no longer used? That migration will need to work in an "online" manner, i.e. there needs to be code that runs before the migration, during the migration and after the migration. How will that code know if the first comment has been already deleted for a given note? It can't just be `all_comments.drop(1)` for example, since that won't work after the next migration has completed.

tomhughes:
> How will that code know if the first comment has been already deleted for a given note? It can't just be `all_comments.drop(1)` for example, since that won't work after the next migration has completed.

I was imagining deleting the first comment only if it was an `opened` comment and on top of that we could also require author and/or text to be the same.

nenad-vujicic:
> I would also interested in what the next step will be.

At the moment, the plan is following:

1. Perform data migration
2. If we succeeded to migrate all notes, remove optional using records from first special comment, otherwise we'll have to leave some fallback mechanism
3. Create DB table `old_notes` which will store note versions and update Note model with appropriate associations
  - This will be decomposed into several PRs
  - Perhaps this is good place to discuss how to apply #5619 to notes tables (perhaps use `notes_versions` instead of `old_notes`?)
4. Add creating versions on notes actions (create, reopen, close, ..)
5. Populate `note_versions` from `note_comments`
6. Create helper routine which will generate list of comments (i.e. discussion, used in left side-bar, rss feed, xml / json files, ..) from `note_versions` table and from `note_comments` table (but using only note comments with non-blank bodies
7. Replace displaying / using discussions with helper routine
8. Remove column `event` from `note_comments` table and comments with blank bodies
9. .. continue with adding note tags ..

I'm aware this plan is full of flaws and there are multiple points which should be further discussed / defined, so everyone's comment is very welcome!

db/migrate/20250212160355_backfill_note_descriptions.rb
nenad-vujicic: We have proposal for supporting notes with a missing opening comment (and generally all comments) in #5712. Do we have other special cases we should support before this?

nenad-vujicic:
@tomhughes do you think we can continue with current version of this PR or we should try with more robust data-migration proposed [here](https://github.com/openstreetmap/openstreetmap-website/pull/5667#issuecomment-2659898440)?

tomhughes:
Well we don't currently have the infrastructure in place for that method, and I'm not sure this change is large enough to really need it.

The big problem with that approach is that it doesn't force the migration to happen, so you have to document that the job needs to be run when upgrading past a certain point and then hope that everybody using the code base reads that document and runs the job at the right time.

tomhughes:
I think this is fine in this case, so I'm going to go ahead and merge it. Thanks for all the work.

PR #5669, opened 2025-02-13
Transition more requests to fetch

hlfan:
### Description
Moves all AJAX requests that don't need OAuth to fetch. (except Matomo as this fetches a script)

### How has this been tested?
Logpoints and quite a bit of erratic map movement to check for aborts

### DIFF
--- a/app/assets/javascripts/index/layers/data.js
+++ b/app/assets/javascripts/index/layers/data.js
@@ -79,20 +79,27 @@ OSM.initializeDataLayer = function (map) {
 
   function getData() {
     var bounds = map.getBounds();
-    var url = "/api/" + OSM.API_VERSION + "/map?bbox=" + bounds.toBBoxString();
+    var url = "/api/" + OSM.API_VERSION + "/map.json?bbox=" + bounds.toBBoxString();
 
     /*
      * Modern browsers are quite happy showing far more than 100 features in
-     * the data browser, so increase the limit to 4000 by default.
+     * the data browser, so increase the limit to 4000.
      */
     const maxFeatures = 4000;
 
     if (dataLoader) dataLoader.abort();
 
-    dataLoader = $.ajax({
-      url: url,
-      dataType: "json",
-      success: function (data) {
+    dataLoader = new AbortController();
+    fetch(url, { signal: dataLoader.signal })
+      .then(response => {
+        if (response.ok) return response.json();
+        const status = response.statusText || response.status;
+        if (response.status !== 400) throw new Error(status);
+        return response.text().then(text => {
+          throw new Error(text || status);
+        });
+      })
+      .then(function (data) {
         dataLayer.clearLayers();
 
         var features = dataLayer.buildFeatures(data);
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/index/layers/data.js
```diff
@@ -116,26 +124,14 @@ OSM.initializeDataLayer = function (map) {
...
-          displayLoadError(XMLHttpRequest.responseText, closeError);
-        } else if (XMLHttpRequest.statusText) {
-          displayLoadError(XMLHttpRequest.statusText, closeError);
-        } else {
-          displayLoadError(String(XMLHttpRequest.status), closeError);
-        }
-      }
-    });
+        });
+      }).finally(() => dataLoader = null);
```
tomhughes, line 134: Can we put the `.finally` on a new line for consistency with other places?

app/assets/javascripts/index/layers/notes.js
```diff
@@ -71,30 +71,28 @@ OSM.initializeNotesLayer = function (map) {
     var size = bounds.getSize();
 
     if (size <= OSM.MAX_NOTE_REQUEST_AREA) {
-      var url = "/api/" + OSM.API_VERSION + "/notes.json?bbox=" + bounds.toBBoxString();
+      var url = "/api/" + OSM.API_VERSION + "/notes?bbox=" + bounds.toBBoxString();
```
tomhughes, line 74: Was there a reason for preferring sending an accept header over using the `.json` extension?

app/assets/javascripts/index/layers/notes.js
```diff
@@ -71,30 +71,28 @@ OSM.initializeNotesLayer = function (map) {
...
+            delete oldNotes[feature.properties.id];
+            notes[feature.properties.id] = updateMarker(marker, feature);
+          }
 
-      noteLoader = null;
+          for (var id in oldNotes) {
+            noteLayer.removeLayer(oldNotes[id]);
+          }
+        })
+        .catch(() => {})
```
tomhughes, line 94: Do we actually need this `.catch` if it doesn't do anything?

app/assets/javascripts/index/query.js
```diff
@@ -157,16 +157,18 @@ OSM.Query = function (map) {
       $section.data("ajax").abort();
     }
 
-    $section.data("ajax", $.ajax({
-      url: url,
+    $section.data("ajax", new AbortController());
+    fetch(url, {
       method: "POST",
-      data: {
+      headers: { "Content-Type": "application/x-www-form-urlencoded" },
```
tomhughes, line 163: Is this necessary? I didn't actually check what it was sending when I did my version but it seemed to work and I assumed having `URLSearchParams` as the body was enough of a hint to make it send that type?

app/assets/javascripts/index/search.js
```diff
@@ -47,21 +47,22 @@ OSM.Search = function (map) {
...
         csrf_param = $("meta[name=csrf-param]").attr("content"),
         csrf_token = $("meta[name=csrf-token]").attr("content"),
-        params = {};
+        params = new URLSearchParams();
 
     $(this).hide();
     div.find(".loader").show();
 
-    params[csrf_param] = csrf_token;
+    params.append(csrf_param, csrf_token);
```
tomhughes, line 55: Was there a reason for using `.append` here rather than `.set` like other places?

app/assets/javascripts/index/layers/notes.js
tomhughes: So this is what I wasn't sure about - what happens if a fetch is rejected and nothing catches the rejection.

If that leads to a message in the console then we should probably have a no-op catch on all of them? Unless we want network errors to get reported in the console but in that case the ones that exist to catch aborts should rethrow other errors?

hlfan: > have a no-op catch on all of them?

Currently, I only put catches on fetches with AbortSignals, cause these are the errors I can catch.
Network errors will still end up in the console.

app/assets/javascripts/index/layers/data.js
hlfan: Done.

app/assets/javascripts/index/layers/notes.js
hlfan: Swapped.

app/assets/javascripts/index/query.js
hlfan: Removed all three.

app/assets/javascripts/index/search.js
hlfan: Nope, fixed.

HolgerJeromin (approved):

tomhughes (approved):
Thanks for the changes. I think this all looks good now.

PR #5671, opened 2025-02-13
Set directions starting point to "Show My Location"

AntonKhorev:
Answers the question https://github.com/openstreetmap/openstreetmap-website/pull/5635#issuecomment-2646420820.

If the location was already acquired with *Show My Location*, it is used as a starting location in *Directions* if
- the starting location is empty AND
- no `route` param was provided (this is to support *Reverse Directions*; points are reversed using the `route` parameter; we want to suppress substituting the location when empty value is swapped into the starting point)

If *Show My Location* wasn't activated before opening the *Directions* page, the activation will use the acquired location as soon as it's available, again if the starting location is empty.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -4,6 +4,8 @@
 
 OSM.Directions = function (map) {
   let controller = null; // the AbortController for the current route request if a route request is in progress
+  let lastFoundLocation = null;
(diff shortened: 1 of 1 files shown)


hlfan:
This will set the starting point marker regardless of whether the user is on the directions page.

image

hlfan (approved):

tomhughes:
Thanks for this, but I think I prefer @hlfan's version in #5727.

Issue #5488, opened 2025-01-09
Add documentation for running system specs on Ubuntu


gravitystorm:
Our system specs (and our teaspoon specs) don't work on Ubuntu, due to incompatibilities with firefox snap. We should document that they don't work, or we should document how to make them work, or (ideally) figure out if there's a way to get them to work automatically.

The root of the problem is that on Ubuntu, firefox is distributed as a snap package, and so the path returned by `which firefox` is not actually a firefox binary, it's just a bit of snap-related code. 

My personal not-quite-working-workarounds are, in `test/application_system_test_case.rb`:

```diff
 class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   driven_by :selenium, :using => :headless_firefox do |options|
     options.add_preference("intl.accept_languages", "en")
+    options.binary = "/snap/firefox/current/usr/lib/firefox/firefox"
   end
```

I've still found problems with some javascript stuff that I haven't yet resolved.

For `/test/teaspoon_env.rb`:

```diff
   config.driver_options = {
     :client_driver => :firefox,
     :selenium_options => {
-      :options => Selenium::WebDriver::Firefox::Options.new(:args => ["-headless"])
+      :options => Selenium::WebDriver::Firefox::Options.new(:args => ["-headless"], :binary => "/snap/firefox/current/usr/lib/firefox/firefox")
     }
   }
```

I'm not sure whether documenting that they don't work is easier than documenting the workarounds!

AntonKhorev:
Fixed in #6111

> On Ubuntu 24.04, if Firefox is installed via snap, you may need to override the Firefox binary path ...

PR #5672, opened 2025-02-14
Add option to override Firefox binary used in system tests

AntonKhorev:
Addresses #5488 by providing a setting instead of workarounds such as modifying `test/application_system_test_case.rb`.

I'm also using these workarounds and it's annoying because you can accidentally commit them. But if there's a setting, I can put it in some file that's ignored by git, for example in `config/settings/test.local.yml` (assuming tests run in test environment; teaspoon tests by default don't but you can run them with `RAILS_ENV=test bundle exec teaspoon`).

### DIFF
--- a/config/settings/test.yml
+++ b/config/settings/test.yml
@@ -53,3 +53,5 @@ doorkeeper_signing_key: |
   cK1+/2V+OkM/0nXjxPwPj7LiOediUyZNUn48r29uGOL1S83PSUdyST207CP6mZjc
   K8aJmnGsVEAcWPzbpNh14q/c
   -----END PRIVATE KEY-----
+# Override Firefox binary used in system tests
+#system_test_firefox_binary:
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -182,3 +182,5 @@ smtp_password: null
 #  -----BEGIN PRIVATE KEY-----
 #  ...
 #  -----END PRIVATE KEY-----
+# Override Firefox binary used in system tests
+#system_test_firefox_binary:
```
tomhughes, line 186: Can we put this in `config/settings/test.yml` as it's only for testing?

AntonKhorev: Is teaspoon supposed to be run in test environment?

tomhughes: That's actually an excellent question - my answer would be yes, and it doesn't work in my dev environment unless I force it to, but currently there is nothing to make that happen by default.

tomhughes: Actually the github workflow does run it with `RAILS_ENV=test` set.

AntonKhorev: Moved to `config/settings/test.yml`.

Then we have to update `CONTRIBUTING.md` to explain how to run teaspoon.

tomhughes:
Thanks, this looks good to me now.

PR #5673, opened 2025-02-14
Remove current_user branch from ApiController#deny_access

AntonKhorev:
Unreachable because if there's no `doorkeeper_token`, there's also no `current_user`.

This is the only place in the api where `current_user` is set:
https://github.com/openstreetmap/openstreetmap-website/blob/f5af8befa9ffe0c95f4a3c58d2fbb63a2e971ab0/app/controllers/api_controller.rb#L105


tomhughes:
Looks good to me, thanks.

PR #5676, opened 2025-02-15
Include spam reports on unresolved issues in the spam score

tomhughes:
I think this should be reasonably safe - it will need three reports to trigger a suspension or possibly less if there are already links in the content adding to a score.

Given the negative influence of edits on the score it's unlikely that people reporting users for map spam will manage to trigger this, leaving the normal DWG management and edit rate limits to deal with those.


gravitystorm:
Merged thanks. As discussed, a followup could be for the report creation process to also trigger the `user.spam_check` process

PR #5677, opened 2025-02-15
User list enhancements

tomhughes:
This enhances the user list view for administrators to include a search parameter bar at the top, to allow searching by IP subnet, and to properly preserve all search fields in links.


gravitystorm:
Could do with some tests, and there's no handling of error states if you (ahem) make a mistake when entering IP subnets.

But otherwise looks good to me.

PR #5175, opened 2024-09-09
Send <head> along with pagination frames

AntonKhorev:
Things broken by Turbo pagination part 2, the previous part is #5111.

Previous/next page visits are promoted to full page visits with Turbo rewriting most of `<head>` tags. If new tags aren't sent together with the updated frame, old ones will be simply deleted. That includes deleting a page title and csrf token tags, which is going to break some functionality, particularly logouts.

Open any page with Turbo pagination, click *Older/Newer* link, notice that the page title is gone. Then click *Logout* and it's not going to work. https://github.com/openstreetmap/openstreetmap-website/pull/4646#issuecomment-2284931759 worked correctly for changeset elements pagination because it's not promoted to page visits.

PR #5678, opened 2025-02-15
Prevent caching a partial XHR response as a full page URL

### LINKED ISSUE #5663: After restoring a closed tab, the page's CSS styles are not loaded

0. Open https://openstreetmap.org/history
1. Select any changeset
2. Close the tab
3. Wait a few seconds
4. Restore the tab
5. Styles are not loaded

https://github.com/user-attachments/assets/f7da2062-92a2-424a-b494-5535a8efc0f4

Firefox 136.0b4. I've also seen reports of a similar issue in Discord and the local community's Telegram chat, but I'm not sure if the steps are exactly the same.

It seems that the problem appeared today/yesterday.

tomhughes:
Long term we should fix the sidebar to use turbo but for now this fixes #5663.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -38,6 +38,14 @@ $(document).ready(function () {
 
     $("#sidebar_loader").show().addClass("delayed-fade-in");
 
+    // Prevent caching the XHR response as a full-page URL
+    // https://github.com/openstreetmap/openstreetmap-website/issues/5663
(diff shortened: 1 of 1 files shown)


AntonKhorev:
We already have the same problem with Turbo where we do frame updates promoted to page visits. We only do this in newer/older pagination. See #5175.

AntonKhorev:
Merged, thanks.

Not for anyone checking this update: it doesn't necessarily invalidate already incorrectly cached pages.

Issue #865, opened 2015-01-06
changeset-page doesn't pan to changeset location when opened from links in rss-feed


tyrasd:
Links to changesets from the changeset discussions RSS feed contain a hash (apparently linking to specific comments, example: https://www.openstreetmap.org/changeset/27962007#c3893) which somehow confuses the URL-hash parser and prevents the map to be panned to the changeset's location.

(probably related: #555)

mmd-osm:
To reproduce:

```bash
curl https://www.openstreetmap.org/changeset/27962007/comments/feed
```


```xml
<rss version="2.0" xmlns:dc="http://purl.org/dc/elements/1.1/">
  <channel>
    <title>OpenStreetMap changeset discussion</title>
    <link>https://www.openstreetmap.org/</link>
<item>
  <title>New comment on changeset #27962007 by tunnelbauer</title>
  <link>https://www.openstreetmap.org/changeset/27962007#c3893</link>
  <guid>https://www.openstreetmap.org/changeset/27962007#c3893</guid>
  <description>
    <![CDATA[<h2>New comment on changeset #27962007 by tunnelbauer</h2>
<div class="changeset-comment" style="margin-top: 5px">
  <div class="changeset-comment-description" style="font-size: smaller; color: #999999">Updated <time title=" 6 January 2015 at 18:26" datetime="2015-01-06T18:26:01Z">over 9 years ago</time> by tunnelbauer</div>
  <div class="changeset-comment-text">Hallo wirthba,

ich habe den Punkt schon das letzte Mal wieder losgelöst von dem Autobahnzubringer - diese Straßen sind miteinander nicht verbunden. Zum einen steht auf der rechten Seite des Autbahnzubringer eine rd. 3m hohe Lärmschutzwand und zum anderen ist die Straße mit &quot;access=no&quot; rund 4m unter dem Niveau des Zubringers - somit können die nicht einfach miteinander verbunden werden.</div>
</div>
]]>
  </description>
  <dc:creator>tunnelbauer</dc:creator>
  <pubDate>Tue, 06 Jan 2015 18:26:01 +0000</pubDate>
</item>
  </channel>
</rss>
```

gravitystorm:
This is also a problem with the notes comment feed, which contains links with fragments (targetting the comment id).

`curl https://www.openstreetmap.org/api/0.6/notes/feed?bbox=8.1706155,51.8144663,8.5269845,51.9737924`

```
[...]

PR #5679, opened 2025-02-15
fix map panning to objects when url hash is not empty but has no map location

tyrasd:
Fix for #865. Unfortunately, the code is slightly different when handling OSM objects (nodes/ways/relations), notes or changesets. I tried to make it a little bit more coherent by always using the `map.addObject({}, callback)` approach to zoom to the location only if the `hash` parameters had no location information.


tomhughes:
Looks good to me, thanks.

PR #5680, opened 2025-02-15
Use title case consistently for nav tab labels

gravitystorm:
In almost all tab navs we already use [title case](https://en.wikipedia.org/wiki/Title_case), and this PR changes the remaining few to match.


tomhughes:
Looks good to me, thanks.

PR #5684, opened 2025-02-16
Add option to use headless Firefox in system tests

AntonKhorev:
Continues #5672.

System test normally run in a headless browser, which is what you want if you run the entire test suite. But when writing and running a single test, it's useful to see what's happening. This PR adds a setting to choose whether tests should run in a headless Firefox.


tomhughes:
I usually just hack to code temporarily to do that, but it looks good to me, thanks.

PR #5686, opened 2025-02-16
don't re-pan the map when selecting another note that is already on screen

### LINKED ISSUE #648: Don't zoom in when commenting/closing a note.
Complement to: https://github.com/openstreetmap/openstreetmap-website/issues/643
(and thanks for resolving it).
Please also do not zoom in when commenting or closing a note.

### LINKED ISSUE #5683: Selecting a note centres the map on it and zooms out
### Problem

Recently a behaviour was introduced where selecting a note centres the map on the note, and zooms out to a certain level if currently zoomed in further than that level. This is a problem:

1. When browsing notes, we may be zoomed in (like in my screenshot), and we don't want to suddenly be moved and zoomed out of the area we were focusing on.
2. This is especially noticeable if the user wants to browse several notes, because the map will keep moving as we select each one, which is distracting and we have to keep re-orientating ourselves.
3. The map flashes each time I select a note.
4. Without a node selected, after hard refreshing (CTRL+F5) the page, this issue does not occur for the first note I click on. It happens on all subsequent occasions.

The previous behaviour where the map does not move or zoom when selecting a note should be restored.

### Description

_No response_

### Screenshots

Image

tyrasd:
Fixes #5683 (a regression in #5679): the problem was that when clicking on a new note on the map, the URL hash with the `map` parameter is cleared somewhere before  `pushstate` gets called, resulting in the map to be panned in `map.addObject` during the `initialize` step. It's now solved by passing an additional parameter to `initialize` and only do the map centering there.

This also allowed to fix #648 via 793757a6f.

### DIFF
--- a/app/assets/javascripts/index/note.js
+++ b/app/assets/javascripts/index/note.js
@@ -44,7 +44,7 @@ OSM.Note = function (map) {
         oauth: true,
         success: () => {
           OSM.loadSidebarContent(path, () => {
-            initialize(path, id);
+            initialize(path, id, false);
           });
         },
         error: (xhr) => {
--- a/app/assets/javascripts/index/note.js
+++ b/app/assets/javascripts/index/note.js
@@ -22,16 +22,10 @@ OSM.Note = function (map) {
 
   page.pushstate = page.popstate = function (path, id) {
     OSM.loadSidebarContent(path, function () {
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #5688, opened 2025-02-16
Make reports trigger a spam check on the reported user

tomhughes:
As suggested in #5676 this triggers a spam check on a reported user after the report is saved, as their score may have changed.


gravitystorm (changes requested):
Needs tests. Mostly just to demonstrate to other developers what it's supposed to achieve, but also so that if we add auto-close for issues (or similar) in future, we don't accidentally break this part during refactoring.

tomhughes:
Added a test, and also fixed the scoring to prevent one user making multiple reports to increase the score.

PR #5689, opened 2025-02-16
User settings fixes

tomhughes:
This is the cleanups from #5453 without the main change to switch to turbo.


gravitystorm:
Merged, thanks

PR #5692, opened 2025-02-16
Use ubuntu-latest for all workflows

tomhughes:
There's generally no point running on multiple versions as the important things like ruby are not the OS versions in any case and production runs on Debian now not Ubuntu.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   danger:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


gravitystorm:
Yes, I agree. I used to think this was useful, but see all my previous mentions of "Github flavoured Ubuntu" that shows it is of limited utility.

HolgerJeromin:
That is good:

> Ubuntu 20.04 Actions runner image will begin deprecation on 2025-02-01 and will be fully unsupported by 2025-04-01

https://github.com/actions/runner-images/issues/11101

PR #5693, opened 2025-02-16
Test map panning when visiting changeset and note comment links

AntonKhorev:
Tests for bugs fixed in #5679.


tomhughes:
Looks good to me, thanks.

PR #5694, opened 2025-02-16
Enable exporting direction paths

### LINKED ISSUE #5224: Allow exporting direction path tracks
### Problem

Directions tracks are great,
> https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=24.18978%2C120.86377%3B24.18252%2C120.86482

But the website needs to provide a free and open export method, like
- [ ] Export track to GPX
- [ ] Export track to KML

else the direction tracks are merely one more proprietary solution, requiring the user to reverse-engineer the internal communications with the browser, if the user wishes to export the path:

Internal communication interception (via DevTools):
> Request URL:
> https://routing.openstreetmap.de/routed-car/route/v1/driving/120.8648163,24.1825192;120.863771,24.189783?overview=false&geometries=polyline&steps=true&hints=t8PogbGSCYcCAAAAUAAAAAAAAAAAAAAAAppJQMyWr0IAAAAAAAAAAAIAAABQAAAAAAAAAAAAAAAiQQAAGzw0B34bcQEbPDQHVxtxAQAA7xBkH6V2%3...

So the user must employ a [polyline decoder](https://www.google.com/search?q=polyline+decoder), or SVG analyzer...
https://stackoverflow.com/questions/32476218/how-to-decode-encoded-polylines-from-osrm-and-plotting-route-geometry

Well all I got were sections of the equator.  So I'll just give up trying to reverse engineer it.

Probably end up using Google,
https://gis.stackexchange.com/questions/152571/export-google-maps-route-to-kml-gpx

hlfan:
### Description
Closes #5224 by adding a link below the instructions table to download the route as a geoJSON:

image

### How has this been tested?
The functionality? Locally.
The translation? Changed on Thursday.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -36,6 +36,8 @@ OSM.Directions = function (map) {
     OSM.DirectionsEndpoint(map, $("input[name='route_to']"), OSM.MARKER_RED, endpointDragCallback, endpointChangeCallback)
   ];
 
+  let downloadURL = null;
+
   const expiry = new Date();
   expiry.setYear(expiry.getFullYear() + 10);
 
@@ -188,6 +190,16 @@ OSM.Directions = function (map) {
 
         turnByTurnTable.append(row);
       });
+
+      const blob = new Blob([JSON.stringify(polyline.toGeoJSON())], { type: "application/json" });
+      URL.revokeObjectURL(downloadURL);
+      downloadURL = URL.createObjectURL(blob);
+
+      $("#sidebar_content").append(`<p class="text-center"><a href="${downloadURL}" download="${
+        I18n.t("javascripts.directions.filename")
+      }">${
+        I18n.t("javascripts.directions.download")
+      }</a></p>`);
 
       $("#sidebar_content").append("<p class=\"text-center\">" +
         I18n.t("javascripts.directions.instructions.courtesy", { link: chosenEngine.creditline }) +
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3312,6 +3312,8 @@ en:
           ninth: "9th"
           tenth: "10th"
       time: "Time"
+      download: "Download track as GeoJSON"
+      filename: "route"
     query:
       node: Node
       way: Way
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3312,7 +3312,7 @@ en:
           ninth: "9th"
           tenth: "10th"
       time: "Time"
-      download: "Download track as GeoJSON"
+      download: "Download route as GeoJSON"
       filename: "route"
     query:
       node: Node


app/assets/javascripts/index/directions.js
```diff
@@ -189,6 +192,20 @@ OSM.Directions = function (map) {
...
 
+      const download = exportGroup.toGeoJSON();
+      for (const i in endpoints) {
+        download.features[i].properties["marker-color"] = endpoints[i].marker.getElement().src.match(/marker-(\w+)/)[1];
+      }
+      const blob = new Blob([JSON.stringify(download)], { type: "application/json" });
+      URL.revokeObjectURL(downloadURL);
+      downloadURL = URL.createObjectURL(blob);
+
+      $("#sidebar_content").append(`<p class="text-center"><a href="${downloadURL}" download="${
```
HolgerJeromin, line 203: You clean up the old objectUrl but do not clean up sidebar_content.
By reading the code this could perhaps result in multiple `<p>`

hlfan: It already gets cleared in line 158?

app/assets/javascripts/index/directions.js
```diff
@@ -36,6 +36,9 @@ OSM.Directions = function (map) {
     OSM.DirectionsEndpoint(map, $("input[name='route_to']"), OSM.MARKER_RED, endpointDragCallback, endpointChangeCallback)
   ];
 
+  const exportGroup = L.layerGroup([...endpoints.map(m => m.marker), polyline]);
+  let downloadURL = null;
+
```
tomhughes, line 41: Does this code need to be here, or can we move it down to be with the other code?

My concern is that the second code block makes assumptions about how the order of features in the layer group relates to the endpoints array that is established by this code and I can't see a good way to avoid that so I'd at least like to keep all the code together if we can.

hlfan: Well, this doesn't need to be done for every route displayed, so I want it to be outside of a function. For the download URL, I want to be able to revoke the last blob URL when creating the new one.
For now, I could drop the marker-color assignment or just export the polyline, as the pages linked in https://github.com/openstreetmap/openstreetmap-website/issues/5224#issuecomment-2362118967 do.

app/assets/javascripts/index/directions.js
HolgerJeromin: oh. yes. I did not checked the whole file 🤦

HolgerJeromin (approved):

app/assets/javascripts/index/directions.js
tomhughes: No that's fine.

config/locales/en.yml
```diff
@@ -3312,6 +3312,8 @@ en:
           ninth: "9th"
           tenth: "10th"
       time: "Time"
+      download: "Download track as GeoJSON"
```
tomhughes, line 3315: Can we make this `Download route as` please.

tomhughes (changes requested):

config/locales/en.yml
hlfan: Done.

PR #5695, opened 2025-02-16
Remove some unnecessary calls to clear the mail queue

gravitystorm:
These were not needed at the end of most tests for quite some time, since rails clears them by default for controller and integration tests, and also no longer required in system tests after #5691

There's still some calls that could be removed, for example by also automatically clearing `job` tests and by slightly reworking one of the mega tests in the `passwords_controller_test`.


tomhughes:
Looks good to me, thanks.

PR #5696, opened 2025-02-17
Move edit account action to show

AntonKhorev:
Does this https://github.com/openstreetmap/openstreetmap-website/pull/5453#issuecomment-2661441402, replacing `/account/edit` paths with `/account`.


config/routes.rb
```diff
@@ -297,14 +297,15 @@
...
-  resource :account, :only => [:edit, :update, :destroy] do
+  resource :account, :only => [:show, :update, :destroy] do
     scope :module => :accounts do
       resource :terms, :only => [:show, :update]
       resource :pd_declaration, :only => [:show, :create]
       resource :deletion, :only => :show
       resource :home, :only => :show
     end
   end
+  get "/account/edit", :to => redirect(:path => "/account"), :as => nil
```
tomhughes, line 308: Do we actually need this redirect? Are we expecting sites outside openstreetmap.org to have direct links to the account edit page?

tomhughes (commented):
I'm not sure there's a good reason for the step by step change of `edit_account_path` to `account_path` here... I'd be inclined to squash all that down into the main commit (which already changes some uses) and just have the two preparatory commits and the one main commit.

config/routes.rb
AntonKhorev: This one is even less likely to be linked:
https://github.com/openstreetmap/openstreetmap-website/blob/483f4c1341ee2b0eb438dc7704e4ba272d1e74fe/config/routes.rb#L300

Do we need it?

AntonKhorev:
Squashed everything after the main commit.

config/routes.rb
tomhughes: There seem to be a handful of hits on that but they may all be nonsense.

tomhughes:
Thanks. I don't know what the answer is on the redirects but I'll merge this - we can always remove those later if we can confirm they're not being used.

gravitystorm:
> Thanks. I don't know what the answer is on the redirects but I'll merge this - we can always remove those later if we can confirm they're not being used.

I've always assumed that any "GET" page might be linked from elsewhere. For the account page, I can imagine user documentation (e.g. on the wiki, learnosm, or internal instructions at other organisations) might point to it, but without logs from the OSMF it's just guesswork.

PR #5697, opened 2025-02-17
Always check changes/comments counts in single changeset tests

AntonKhorev:
Make `assert_single_changeset` check attributes that are always present in changeset responses.


tomhughes:
Looks good to me, thanks.

PR #5698, opened 2025-02-17
don't set map coords if mapParams contains object, but no "map" in hash

tyrasd:
fixes https://github.com/openstreetmap/iD/issues/10779 (a regression introduced in #5632)

### DIFF
--- a/app/assets/javascripts/edit/id.js.erb
+++ b/app/assets/javascripts/edit/id.js.erb
@@ -22,7 +22,7 @@ $(document).ready(function () {
       ({ zoom, center: { lat, lng: lon } } = hashArgs);
     } else {
       setMapView = false;
-    };
+    }
   } else if (idData.lat && idData.lon) {
     ({ zoom, lat, lon } = { zoom: 16, ...idData });
   }
--- a/app/assets/javascripts/edit/id.js.erb
+++ b/app/assets/javascripts/edit/id.js.erb
@@ -14,14 +14,22 @@ $(document).ready(function () {
   const mapParams = OSM.mapParams();
   const params = new URLSearchParams();
   let { zoom, lat, lon } = mapParams;
+  let setMapView = true;
 
   if (mapParams.object) {
     params.set("id", mapParams.object.type + "/" + mapParams.object.id);
-    if (hashArgs.center) ({ zoom, center: { lat, lng: lon } } = hashArgs);
+    if (hashArgs.center) {
+      ({ zoom, center: { lat, lng: lon } } = hashArgs);
+    } else {
+      setMapView = false;
+    };
   } else if (idData.lat && idData.lon) {
     ({ zoom, lat, lon } = { zoom: 16, ...idData });
   }
-  params.set("map", [zoom || 17, lat, lon].join("/"));
+
+  if (setMapView) {
+    params.set("map", [zoom || 17, lat, lon].join("/"));
+  }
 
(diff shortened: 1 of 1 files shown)


hlfan:
idData already has the coordinates, so why can't this be used here?

tyrasd:
ehrm… 

Given that:
* `mapParams.center/.zoom` get the location via different means (either via the URL's hash, `mlat/mlon` query parameters or the previously last visited location that's stored in a cookie)
* iD will not auto-center to a the supplied osm features if a `#map=…` parameter is present

The expected behavior is that:
* if iD is loaded via a link that only has a map object (e.g. `osm.org/edit?node=1`, or `osm.org/edit?note=1234`), the editor should auto-center to the respective object during startup
* if iD is loaded via a link that has both a map object and a map hash (e.g. `osm.org/edit?node=1#map=0/0/0`), it should set the map to the given lat/lon/zoom values
* if iD is loaded via a bare link (`osm.org/edit`), it should use the map view that is stored in the cookie

So… since `mapParams` should already have the information from the `hashArgs`, I guess it should be able to reduce the inner `if` to something like the following, is that what you mean by _Removing the `else`_??

```js
    if (!hashArgs.center) {
      setMapView = false;
    };
```

I just tried to keep the fix here to the bare minimum and didn't want to touch the rest of the logic…

PS: if I see this correctly, `idData.lat/.lon` are only set when the editor is loaded using a `…?note=xxx` link, see 6494747882358cc1506ae442c89eed3db1de94a4. This is AFAICT not necessary anymore because notes are by now also using the `mapParams.object` path. But let's handle that in another PR, shall we?

tyrasd:
> idData already has the coordinates

This one is also a bit confusing. Normally, `idData.lat/.lon` are not set. If I see this correctly, `idData.lat/.lon` are only set when the editor is loaded using a `…?note=xxx` link, see 6494747882358cc1506ae442c89eed3db1de94a4. This is AFAICT not necessary anymore because notes are by now also using the `mapParams.object` path. But let's handle that in another PR, shall we?

hlfan:
> in this case, params.set("map", …) must not be called!

Why would you not want to use the coordinates for nodes, ways and notes if these are already loaded in the [site_controller](https://github.com/openstreetmap/openstreetmap-website/blob/483f4c1341ee2b0eb438dc7704e4ba272d1e74fe/app/controllers/site_controller.rb#L86)?
In my experience this ommission adds a noticeable delay to seeing the object, as iD first loads the fully zoomed out view and only later can zoom in.
If this delay can be reduced, I'm happy to remove the server side preset coordinates.

tyrasd:
> The map parameter would always be set.

not in the iframe's `src` parameter before 5310c5b :shrug: 
image

> Why would you not want to use the coordinates for nodes, ways and notes if these are already loaded in the [site_controller](https://github.com/openstreetmap/openstreetmap-website/blob/483f4c1341ee2b0eb438dc7704e4ba272d1e74fe/app/controllers/site_controller.rb#L86)?


Ok, wait. Of course it would be good to load the editor right away at the correct coordinates. For nodes and notes this is trivial (we can use the `lat`/`lon` directly), not sure how doable it would be in the ruby code for ways and relations, as there would be some zoom-level calculation to do (which depends on the size of the feature as well as the size of the browser window and the applied browser-zoom %).

What I was trying to do here is to merely fix the regression from #5632 by restoring the previous behavior, and nothing more. If you can come up with a better fix, please go ahead, I'm happy to close this one in favor of a better solution.

tyrasd:
closing in favour of #5700

PR #5699, opened 2025-02-17
Add jquery editor support

HolgerJeromin:
At least vscode (derivates) provide first class intellisense support out of the box with that.

before
image
after
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,8 @@
     "osm-community-index": "^5.2.0"
   },
   "devDependencies": {
+    "@types/jquery": "^3.5.32",
+    "@types/leaflet": "^1.9.16",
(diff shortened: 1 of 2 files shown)


tomhughes:
So just installing those packages (even though we don't use them) somehow changes vscode's behaviour? Not really sure I like forcing everybody to install unused packages :-(

HolgerJeromin:
> So just installing those packages (even though we don't use them) somehow changes vscode's behaviour?

Yes. 
The editor searches known files in `node_modules/@types`.
In the past you needed include comments, but this is (at least for jQuery) not needed anymore.

tomhughes:
I guess there's no harm from having a few extra modules...

PR #5700, opened 2025-02-17
Pre-center objects with embed-data

hlfan:
Fixes https://github.com/openstreetmap/iD/issues/10779, closes #5698
Before the regression from #5632, the location source table was:

typ |           |          | hash-loc | !hash-loc
:-: |  :------: |  :-----: | :------: | :------:
n/w |  data-loc |  map-obj | hash-loc | nil
rel | !data-loc |  map-obj | hash-loc | nil
gpx |  data-loc | !map-obj | data-loc | data-loc
nil | !data-loc | !map-obj | map-loc  | map-loc

I changed it to this:

typ |           |          | hash-loc | !hash-loc
:-: |  :------: |  :-----: | :------: | :------:
n/w |  data-loc |  map-obj | hash-loc | data-loc
rel | !data-loc |  map-obj | hash-loc | nil
gpx |  data-loc | !map-obj | data-loc | data-loc
nil | !data-loc | !map-obj | map-loc  | map-loc

And I took the opportunity to restructure the conditions in the spirit of #5632.

### DIFF
--- a/app/assets/javascripts/edit/id.js.erb
+++ b/app/assets/javascripts/edit/id.js.erb
@@ -13,15 +13,18 @@ $(document).ready(function () {
   const hashArgs = OSM.parseHash(location.hash);
   const mapParams = OSM.mapParams();
   const params = new URLSearchParams();
-  let { zoom, lat, lon } = mapParams;
+  let zoom, lat, lon;
 
+  if (idData.lat && idData.lon) {
+    ({ zoom, lat, lon } = { zoom: 16, ...idData });
+  } else if (!mapParams.object) {
+    ({ zoom, lat, lon } = mapParams);
+  }
   if (mapParams.object) {
     params.set("id", mapParams.object.type + "/" + mapParams.object.id);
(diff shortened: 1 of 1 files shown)


tyrasd:
Yeah, this seems bood to me. Thanks @hlfan

tomhughes:
Looks good to me, thanks.

davidpnewton:
This doesn't entirely fix the problem.

The centre of the screen is now in the correct place. However the zoom level is utterly inappropriate for the features in question. For a single node it goes to zoom level 16 by default whereas zoom level 20 or 21 would be far better for a single node.

hlfan:
I already wanted to increase the zoom in general in https://github.com/openstreetmap/openstreetmap-website/pull/5632#discussion_r1948124352 but if we already have a zoom set in the [site_controller](https://github.com/openstreetmap/openstreetmap-website/blob/483f4c1341ee2b0eb438dc7704e4ba272d1e74fe/app/controllers/site_controller.rb#L86) we could use it here too

PR #5702, opened 2025-02-18
Justify secondary nav

hlfan:
Making the secondary nav flushed right while expanded and centered while in the small nav


AntonKhorev:
The menu is left-aligned to avoid jumping around horizontally when the page is loaded:

[Screencast from 19.02.2025 01:28:08.webm](https://github.com/user-attachments/assets/9e958aca-dfc6-454c-9632-ff83f8de507a)

tomhughes:
Looks good to me now, thanks.

It would be nice if we could get it to balance the lines better when there are multiple lines but I don't think there's an easy way to do that.

PR #5703, opened 2025-02-18
Test changeset tags in api changeset show responses

AntonKhorev:
Tags were tested in json responses but together with comments, however tests for comments already exist with various conditions (hidden comments, hidden comments shown to moderators).

Tags in xml responses weren't tested.


tomhughes:
Looks good to me, thanks.

PR #5705, opened 2025-02-19
Remove current user check from dashboard view

AntonKhorev:
Dashboard view only outputs its contents if this condition is true:

```ruby
if current_user and @user.id == current_user.id
```

But it's always true. The page is only accessible when authorized, therefore `current_user` is set. `current_user` is assigned to `@user`, so the `id`s are also always equal. That makes `@user` unnecessary and the entire check unnecessary.


nenad-vujicic (approved):
This works great on my side. Thanks!

tomhughes:
Looks good to me, thanks.

PR #5706, opened 2025-02-19
Allow to agree to new terms if already agreed to old terms

AntonKhorev:
In case *Contributor Terms* or *Terms of Use* were updated since last agreeing to them, which is supposedly the case for many people, it's impossible to accept the new terms.

This PR makes it possible by
- telling when the terms were accepted in *My Account*
- linking to the *Terms* page
  image
- disabling the checks if terms were previously accepted on the *Terms* page, which made it do nothing or redirect back to *My Account*

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -258,16 +258,21 @@ en:
       my_account: My Account
       current email address: "Current Email Address"
       external auth: "External Authentication"
-      contributor terms:
+      contributor_terms:
         heading: "Contributor Terms"
-        agreed: "You have agreed to the new Contributor Terms."
-        not yet agreed: "You have not yet agreed to the new Contributor Terms."
-        review link text: "Please follow this link at your convenience to review and accept the new Contributor Terms."
+        agreed: "You agreed to the Contributor Terms on %{date}."
+        not_agreed: "You have not yet agreed to the Contributor Terms."
         agreed_with_pd: "You have also declared that you consider your edits to be in the Public Domain."
-        link: "https://osmfoundation.org/wiki/Licence/Contributor_Terms"
-        link text: "what is this?"
         not_agreed_with_pd: "You haven't declared that you consider your edits to be in the Public Domain."
-        pd_link_text: "declare"
+        review: "Review the Terms"
+        review_and_accept: "Review and accept the Terms"
+        consider_pd: "Consider Public Domain"
+      terms_of_use:
+        heading: "Terms of Use"
+        agreed: "You agreed to the Terms of Use on %{date}."
+        not_agreed: "You have not yet agreed to the Terms of Use."
+        review: "Review the Terms"
+        review_and_accept: "Review and accept the Terms"
       save changes button: Save Changes
       delete_account: Delete Account...
     go_public:
--- a/test/controllers/accounts/terms_controller_test.rb
+++ b/test/controllers/accounts/terms_controller_test.rb
@@ -48,32 +48,123 @@ module Accounts
       assert_response :success
     end
 
-    def test_update_not_seen_without_referer
-      user = create(:user, :terms_seen => false, :terms_agreed => nil)
-      session_for(user)
+    def test_update_decline_by_not_checking_the_boxes
+      freeze_time do
+        user = create(:user, :terms_seen => false, :terms_agreed => nil, :tou_agreed => nil)
+        session_for(user)
 
-      put account_terms_path, :params => { :read_ct => 1, :read_tou => 1 }
-      assert_redirected_to account_path
-      assert_equal "Thanks for accepting the new contributor terms!", flash[:notice]
+        put account_terms_path, :params => { :continue => "Continue" }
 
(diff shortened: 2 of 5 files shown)


pnorman:
> In case _Contributor Terms_ or _Terms of Use_ were updated since last agreeing to them

The Contributor Terms and Terms of use are different. The Contributor Terms are what grants the OSMF the right to distribute your data and they've never been updated. The [Terms of Use](https://osmfoundation.org/wiki/Terms_of_Use) cover accessing OSMF services. These have had minor updates, but they're not integrated into the website as they apply to all OSMF services, regardless of if the user has had an account.

Handling new CTs would be nice to have, but it's not a feature that would be used right now.

AntonKhorev:
> but they're not integrated into the website

https://github.com/openstreetmap/openstreetmap-website/pull/2028

AntonKhorev:
Since *terms* aren't part of the form, I moved them below the form. I also added *Terms of Use* label/section, which I didn't do previously because there's only one *terms* page.

AntonKhorev:
> There's no need to "accept" changes to the terms though, or rather the terms says that you're expected to check them when you want and that [continuing to use the service constitutes acceptance of any changes](https://osmfoundation.org/wiki/Terms_of_Use#XI._Changes_to_Services_or_Terms).

This interpretation contradicts some of changes in #2028. It added a db column named `tou_agreed`, which you don't need because it's impossible not to agree. Also it contradicts https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services. Particularly the table at the beginning of that page has a *not accepted/unknown* column, which is impossible if the terms are accepted automatically.

AntonKhorev:
Do we just change the language from "accept" the ToU to "confirm that you read", or did we go from *mandatory* acceptance to *automatic* acceptance since https://github.com/openstreetmap/openstreetmap-website/pull/2028#issuecomment-439124390? With automatic acceptance we don't need to make any GDPR-related changes because whenever you use the data you "accept" the ToU and get all of the associated "rights". But obviously that's not what those who want GDPR-changes think. EWG, for example, asked me recently if I plan to "make accepting the terms of use a requirement". They think the ToU need to be "accepted". Now it may be the case that what EWG thinks is not the same what the current LWG thinks about all this.

Alternatively we can ignore all of this, forget about the ToU and use the current Privacy policy as a guide, see https://github.com/openstreetmap/openstreetmap-website/issues/5804#issuecomment-2726403892 minus the github argument. Show some things only to logged in users and ignore the `tou_agreed` attribute. Then this pull request is not very useful and so is #2028 from 2019, aside from adding links to the terms pages.

AntonKhorev:
LWG says that there is a need to "accept" the terms and that users need to be able to accept the Terms of Use by visiting the terms page. It's what this PR lets them to do.

config/locales/en.yml
```diff
@@ -256,16 +256,21 @@ en:
...
-        link: "https://osmfoundation.org/wiki/Licence/Contributor_Terms"
-        link text: "what is this?"
         not_agreed_with_pd: "You haven't declared that you consider your edits to be in the Public Domain."
-        pd_link_text: "declare"
+        review: "Review the Terms"
+        review_and_accept: "Review and accept the Terms"
+        consider_pd: "Consider Public Domain"
+      terms_of_use:
+        heading: "Terms of Use"
+        agreed: "You have agreed to the Terms of Use on %{date}."
```
1ec5: Minor grammatical nit (also above):

```suggestion
        agreed: "You agreed to the Terms of Use on %{date}."
```

AntonKhorev: updated

AntonKhorev:
> Just to confirm, does this PR result in actively presenting pre-2019 users with some sort of notice about the new terms, without having to visit the account settings first?

No it doesn't. There's no point in *actively presenting* something we can't agree to have for ~5 months. If/when we finally allow all users to visit the terms page, we can think about *actively presenting* it.

test/controllers/accounts/terms_controller_test.rb
```diff
@@ -48,32 +48,72 @@ def test_show_not_seen_with_referer
...
+        assert_equal "Thanks for accepting the new contributor terms!", flash[:notice]
+
+        user.reload
+        assert user.terms_seen
+        assert_equal Time.now.utc, user.terms_agreed
+        assert_equal Time.now.utc, user.tou_agreed
+      end
+    end
+
+    def test_update_previously_accepted
```
tomhughes: We should probably have `test_decline_previously_accepted` to check that the dates don't change if you decline when revisiting the terms.

AntonKhorev: Added more decline tests: (previously not accepted, previously accepted) x (by clicking *Cancel*, by submitting without checked boxes[^1])

[^1]: normally not possible if javascript is working correctly

tomhughes (approved):
Thanks for the extra tests. I think this looks good now.

PR #5707, opened 2025-02-19
Migrate eslint config to esmodules

HolgerJeromin:
And upgrade `@stylistic/eslint-plugin-js` to `4.0.x`

closes #5708

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -22,7 +22,7 @@ PreCommit:
     enabled: true
   EsLint:
     enabled: true
-    command: ["bin/yarn", "eslint", "-c", "config/eslint.js"]
+    command: ["bin/yarn", "eslint", "-c", "config/eslint.config.mjs"]
     exclude:
       - vendor/**/*.js
   RailsSchemaUpToDate:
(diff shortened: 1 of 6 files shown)


HolgerJeromin:
I could have kept the name in the end, but having the standard `.mjs` module extension feels not too bad.

hlfan (approved):

tomhughes:
Is this actually working?

According to https://eslint.style/guide/migration#approach-2-migrate-to-1-to-1-plugins if you only import the `@stylistic/eslint-plugin-js` module then you need to use `@stylistic/js/...` for the rules but we aren't? 

If you go with https://eslint.style/guide/migration#approach-1-migrate-to-single-plugin and import `@stylistic/eslint-plugin` then you can drop the `/js/` from the rule names?

HolgerJeromin:
> Is this actually working?

yes:
image

> If you go with https://eslint.style/guide/migration#approach-1-migrate-to-single-plugin and import @stylistic/eslint-plugin then you can drop the /js/ from the rule names?

You did not wanted to pull in `typescript` dependency, so we added the js-only package.

I register the js rules without `js` suffic (`plugins: {"@stylistic": stylisticJs }`) thats why I do not need it in the rule name.
But I am not sure if this is smart or stupid. I only understood this logic after your comment.

tomhughes:
Ah I missed that the prefix was defined by the import.

tomhughes:
Looks good to me, thanks.

PR #5711, opened 2025-02-19
Set zoom in id iframe data

hlfan:
Follow-up to #5700 to include the zoom as well


tomhughes:
Looks good to me, thanks.

PR #5712, opened 2025-02-19
Displays "unknown" description for notes with missing opening comment

nenad-vujicic:
### Description

It seems there are some notes in database with a missing opening comment ([here](https://github.com/openstreetmap/openstreetmap-website/pull/5667#discussion_r1957131929)). This PR adds displaying "unknown" as description for such notes and remaining comments (if available) as part of discussion.

### How has this been tested?

Automatic and manual testing before and after [data migration](https://github.com/openstreetmap/openstreetmap-website/pull/5667#issuecomment-2659190584).


app/controllers/notes_controller.rb
```diff
@@ -43,7 +43,7 @@ def show
 
     @note_includes_anonymous = @note.author.nil? || @note_comments.find { |comment| comment.author.nil? }
 
-    @note_comments = @note_comments.drop(1) if @note.author.nil? || @note.author.active?
+    @note_comments = @note_comments.drop(1) if (@note.author.nil? || @note.author.active?) && @note.description.present?
```
tomhughes, line 46: Shouldn't this condition just be that we drop the first comment if it's of type `opened` rather than trying to guess that from other things?

nenad-vujicic: Thanks. Just updated to use check if first comment is of type `opened` (in entire PR).

It drops first comment (because it's used as description) if its author is not deleted (otherwise it would display "deleted" as description) and if first comment is opening (otherwise it would display "unknown" as description).

tomhughes: Nothing in the helper is looking at `@note_comments` though? All the tests in there are either on the note (which might defer to `all_comments` for an old note) or on `all_comments` directly.

nenad-vujicic: Thank you very much. I think I got it .. Now it drops if first comment (of `@note_comments`) is `opened` and uses same check in helper routine!

tomhughes (approved):
Thanks. I think this all looks good now.

PR #5713, opened 2025-02-20
API changeset resources - index path

AntonKhorev:
Like #5591 but for changesets, with the same modification for `create` action.


tomhughes:
Looks good to me, thanks.

PR #5715, opened 2025-02-21
Update minimum ruby version to 3.2

tomhughes:
We are now running ruby 3.4 in production so there's no longer any need to keep 3.1 support and dropping it enables us to stop holding back several gems and opens the path to a rails 8 upgrade.


mmd-osm:
By the way, are we already using YJIT or planning to do so in the future?

tomhughes:
It's enabled by default in rails 7.2 and later if you're using a ruby that has it (so 3.3 or later).

AntonKhorev:
Merged, thanks.

PR #5716, opened 2025-02-21
Switch to using annotaterb for annotations

tomhughes:
The annotate gem we were using hasn't had any updates for several years, doesn't support rails 8, and generally seems to be abandoned.


AntonKhorev:
Is the github workflow working correctly? I see a bunch of messages like

```
Unable to process app/models/changeset.rb: Couldn't find Active Storage configuration in /home/runner/work/openstreetmap-website/openstreetmap-website/config/storage.yml
```

I broke one annotation and *Rails Annotate Models* job still passed:
https://github.com/AntonKhorev/openstreetmap-website/commit/05e1951fcc22ba9ada39664e5310e15dd788678c

tomhughes:
Yes looks like `annotaterb` must do something differently that means we need to setup storage.

I've fixed that nopw, and tested a broken annotation in https://github.com/tomhughes/openstreetmap-website/actions/runs/13593200376/job/38004040104.

AntonKhorev:
Merged, thanks.

PR #5717, opened 2025-02-21
API changeset resources - show path

AntonKhorev:
Continues #5713 by declaring changesets resources with `show` and `update` actions.


tomhughes:
Looks good to me, thanks.

PR #5719, opened 2025-02-22
Extend user list to allow searching by name or email

tomhughes:
Extends the user list interface used by administrators to allow searching by name or email which will be one less thing I have to in database directly.


AntonKhorev:
Merged, thanks.

PR #5720, opened 2025-02-22
Append one table per route leg

hlfan:
### Description
I modified the directions engine interface to return the steps grouped into legs and create a turn instruction table for each leg. This change isn't visible to the user, as there isn't a third endpoint yet.
### How has this been tested?
So much browser-based that I initially forgot to update the testing stub

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -155,22 +155,25 @@ OSM.Directions = function (map) {
           I18n.t("javascripts.directions.descend") + ": " + formatHeight(route.descend) + ".");
       }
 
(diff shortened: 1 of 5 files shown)


tomhughes:
So the goal here is to eventually support via points?

hlfan:
The goal is to display the routing engines' responses more accurately. Since all of these support multiple legs, the sidebar does too, now.
And while this is a step to support via points, I wouldn't call that the goal.

tomhughes:
Right but as you yourself said there's no actual change to what is displayed at the moment, unless I've misunderstood something? So that's because the engines only return multiple legs with via points?

PR #5721, opened 2025-02-22
API changeset download resource

AntonKhorev:
Continues #5717.

Creates `download` resources on changesets instead of `download` actions.


tomhughes:
Looks good to me, thanks.

PR #5724, opened 2025-02-23
Add show_redactions param to changeset downloads

AntonKhorev:
This PR adds a missing feature in osm-website as compared to cgimap.

There's a `show_redactions` parameter which lets moderators see redacted element versions. In osm-website it works with element versions and histories. In cgimap it also works with changeset downloads. Without this parameter it's impossible to know which elements were changed in a given changeset. If you do a redaction and don't record all elements+versions somewhere, you won't be able to check whether everything was redacted correctly. You'll know that something in a changeset was redacted because the changes count won't match the number of elements in the download. But since the elements are missing, you won't be able to check their next versions, and you often need to redact the versions after the target changeset too.

### DIFF
--- a/app/controllers/api/changesets/downloads_controller.rb
+++ b/app/controllers/api/changesets/downloads_controller.rb
@@ -1,6 +1,8 @@
 module Api
   module Changesets
     class DownloadsController < ApiController
+      before_action :setup_user_auth
+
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #5725, opened 2025-02-23
Make changeset comment create action resourceful

AntonKhorev:
Paths are not changed because there was no `PUT .../create` for creating changeset comments.

The only thing that's going to look a little strange is the path helper. It's named `api_changeset_changeset_comments_path` with "changeset" twice. That's how rails normally names it because we have `Changeset` parent model/controller and `ChangesetComment` child model/controller, and `Changeset`+`ChangesetComment` gives `changeset_changeset_comments`. To avoid that it's possible to rename the child resources to just `comments` and get `api_changeset_comments_path` helper, however this name is already occupied by changeset comments that are not nested in changesets: #4359.

### DIFF
--- a/app/controllers/api/changeset_comments_controller.rb
+++ b/app/controllers/api/changeset_comments_controller.rb
@@ -24,16 +24,16 @@ module Api
     # Add a comment to a changeset
     def create
       # Check the arguments are sane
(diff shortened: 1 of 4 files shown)


tomhughes:
The name of that helper is unfortunate as you say but there's not a lot we can do by the sounds of it, so I think this looks good, thanks.

PR #5726, opened 2025-02-23
API changeset comment visibility resource

AntonKhorev:
Previously there were `hide` and `unhide` actions on changeset comments and an attempt to change them into standard actions which didn't quite work. It went like this: hiding a comment is like deleting, let's map `hide` path to `destroy` action; unhiding is like... there's no standard undelete/undestroy, let's invent another nonstandard `restore` name for now. `create` couldn't be used as an opposite action because it creates a new comment.

However `destroy` and `create` could be used if another resource is introduced. That's what this PR does. Since these operations change `visible` attributes, I guess the resource name should be `visibility`. Then we'll have something similar to message `read_mark` from #5536 that can be created and deleted and also changes an attribute.

That requires a different path. Since the collection path was already introduced in #4359 as `/api/0.6/changeset_comments`, the visibility resource path should be`/api/0.6/changeset_comments/:id/visibility`. `DELETE`/`POST` at this path hide/unhide the comment. For backwards compatibility `POST /api/0.6/changeset/comment/:id/hide` is mapped to `destroy` and `POST /api/0.6/changeset/comment/:id/unhide` is mapped to `create` actions on visibility.

This removes all nonstandard actions from changeset comments controller.


tomhughes:
Looks good to me, thanks.

PR #5727, opened 2025-02-26
Set directions departure to location

hlfan:
Closes #5635, closes #5671

An event-based variant of #5671, vectorizing the coordinates for more compact fallback


tomhughes:
This is throwing an exception for me if I activate it with routing already open:

image

tomhughes:
Because the `latlng` that arrives in the `startinglocation` event handler is an object with the `0` and `1` keys from the array but also a bunch of other keys.

tomhughes:
Thanks. This look good to me now.

PR #5730, opened 2025-02-27
Replace jquery's getJSON with fetch

hlfan:
Moving even more requests away from XHR and jquery.


tomhughes:
Looks good to me, thanks.

PR #5732, opened 2025-02-28
Improve the web manifest

hlfan:
Looping through icon resolutions and adding a few parameters.


hlfan:
So how can I not flash-bang users without `preferred_color_scheme`?

tomhughes:
Looks good to me, thanks.

To make `preferred_color_scheme` work we'd need to generate two versions of `manifest.json` and then have `app/views/layouts/_meta.html.erb` decide which one to link to in the meta headers, as `app/views/layouts/_head.html.erb` already does for style sheets.

PR #5735, opened 2025-02-28
Fix webmanifest icon list

hlfan:
I broke the webmanifest icon list by not removing gsub.
But I'm making up for it by adding an SVG icon.


tomhughes:
Looks good to me, thanks.

Issue #1123, opened 2016-01-01
Accept geo: URI in search and/or other parameter


mmn:
It appears I cannot provide geo: URIs in the search field on the website, openstreetmap.org (try with something like "geo:13.4125,103.8667" as suggested in RFC5870 https://tools.ietf.org/html/rfc5870 ).

Such a feature would be very useful for automated linking from client software, websites etc. There is of course the possibility of directly linking to /#map=$z/$lat/$lon - but that's not standardized anywhere. Accepting a geo: URI would be great for wider compatibility.

For example, I could then setup a default handler in Mozilla Firefox for example so when clicking a geo: URI it could offer openstreetmap.org as a handler (like how the web-based IRC client Mibbit is offered for irc: links)

I wrote on the dev mailing list for OSM about this and hope that once geo: URIs can be linked, we can give a patch to Mozilla and hopefully see implementation pretty soon for clickable geo: links.

tomhughes:
Why not make the handler strip the geo prefix before doing the search? That will already work fine.

Searching for a URI seems like a fairly weird thing to do...

mmn:
On 2016-01-01 14:06, Tom Hughes wrote:
> Why not make the handler strip the geo prefix before doing the search? That will already work fine.

Because it is a URI in its entirety and thus much more distinguishable
from just a generic search term :)

Also for the reason specified further down in my original email as it
can easily be used as a protocol handler. When sending the URI to a
third party service (Mibbit, OSM.org...) the 'geo:' prefix is a good,
easy and generic way to relay important information.

So as Mozilla Firefox would like to send the full geo: URI OSM.org
should reasonably accept it, assuming this interoperability is desired.

> Searching for a URI seems like a fairly weird thing to do...

It could just as well be its own parameter (parsed after URL decoded
since it can contain '?', '&' etc.) or just used in place of the
parameters after '#' or something else. My proposal is just to make it
easy to provide a geo: URI to the web interface via a parameter so it
can be explicitly linked from an external resource.

I don't know which solution would be the smoothest to implement as I
don't know the openstreetmap.org codebase. I'd be willing to write a
patch if I get nudged in the right direction though, so I don't have to
start from scratch in understanding it :)

-- 
Mikael Nordfeldth
https://blog.mmn-o.se/
XMPP/mail: mmn@hethane.se
OpenPGP Fingerprint: AE68 9813 0B7C FCE3 B2FA 727B C7CE 635B B52E

[...]

PR #5736, opened 2025-03-01
Add PWA geo protocol handler

hlfan:
### Description
This PR adds a geo protocol handler in the web manifest and the functionality for parsing in mapParams.
We can't directly parse it as it arrives wrapped in `encodeURIComponent`,
`geo:48.410,12.060?z=9` would be handled as https://www.openstreetmap.org/?geouri=geo%3A48.410%2C12.060%3Fz%3D9
~~This also includes a heuristic zoom fallback based on the precision.~~
### How has this been tested?
With `navigator.registerProtocolHandler`, which can be included later to settle #1123 (I don't know where to put it).

### DIFF
--- a/app/assets/favicons/manifest.json.erb
+++ b/app/assets/favicons/manifest.json.erb
@@ -14,5 +14,11 @@
 	"start_url": "/",
 	"theme_color": "#7ebc6f",
 	"background_color": "#fff",
-	"display": "minimal-ui"
+	"display": "minimal-ui",
+	"protocol_handlers": [
+		{
+			"protocol": "geo",
+			"url": "/?geouri=%s"
+		}
+	]
 }
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -121,6 +121,29 @@ OSM = {
     return mapParams;
   },
 
+  parseGeoURI: function (geoURI) {
+    let url;
+    try {
+      url = new URL(geoURI);
+    } catch (e) { return; }
+    if (!url?.pathname || url.protocol !== "geo:") return;
+    const [path, ...params] = url.pathname.split(";");
+    let coords;
+    try {
+      coords = L.latLng(path.split(","));
+    } catch (e) { return; }
+    if (!coords) return;
+    const searchParams = new URLSearchParams(params.join("&").toLowerCase());
+    const crs = searchParams.get("crs");
+    if (crs && crs !== "wgs84") return;
+    const uncertainty = parseFloat(searchParams.get("u"));
+    const zoom = parseFloat(url.searchParams.get("z"));
+    const out = { coords };
+    if (uncertainty >= 0) out.uncertainty = uncertainty;
+    if (zoom >= 0) out.zoom = zoom;
+    return out;
+  },
+
   parseHash: function (hash = location.hash) {
     const args = {};
 
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -62,6 +62,7 @@ OSM = {
       mapParams.marker = true;
       mapParams.mlon = parseFloat(params.get("mlon"));
       mapParams.mlat = parseFloat(params.get("mlat"));
+      mapParams.mrad = parseFloat(params.get("mrad"));
     }
 
     // Old-style object parameters; still in use for edit links e.g. /edit?way=1234
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -56,6 +56,7 @@ OSM = {
 
   mapParams: function (search) {
     const params = new URLSearchParams(search || location.search),
+          geoURI = OSM.parseGeoURI(params.get("geouri")),
           mapParams = {};
 
     if (params.has("mlon") && params.has("mlat")) {
@@ -63,6 +64,12 @@ OSM = {
       mapParams.mlon = parseFloat(params.get("mlon"));
       mapParams.mlat = parseFloat(params.get("mlat"));
       mapParams.mrad = parseFloat(params.get("mrad"));
+    }
+    if (geoURI) {
+      mapParams.marker = true;
+      mapParams.mlon = geoURI.coords.lng;
+      mapParams.mlat = geoURI.coords.lat;
+      mapParams.mrad = geoURI.uncertainty;
     }
 
     // Old-style object parameters; still in use for edit links e.g. /edit?way=1234
@@ -94,6 +101,12 @@ OSM = {
       mapParams.lon = params.get("mlon");
       mapParams.lat = params.get("mlat");
       mapParams.zoom = params.get("zoom") || 12;
+    } else if (geoURI?.uncertainty > 0 && !("zoom" in geoURI)) {
+      mapParams.bounds = geoURI.coords.toBounds(geoURI.uncertainty * 4);
+    } else if (geoURI) {
+      mapParams.lon = geoURI.coords.lng;
+      mapParams.lat = geoURI.coords.lat;
+      mapParams.zoom = geoURI.zoom || 12;
     } else if (loc) {
       [mapParams.lon, mapParams.lat, mapParams.zoom] = loc;
     } else if (OSM.home) {
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -211,7 +211,9 @@ $(function () {
     map.setView([params.lat, params.lon], params.zoom);
   }
 
-  if (params.marker) {
+  if (params.marker && params.mrad) {
+    L.circle([params.mlat, params.mlon], { radius: params.mrad }).addTo(map);
+  } else if (params.marker) {
     L.marker([params.mlat, params.mlon]).addTo(map);
   }
 
--- a/test/javascripts/osm_test.js
+++ b/test/javascripts/osm_test.js
@@ -73,6 +73,33 @@ describe("OSM", function () {
       expect(params).to.have.property("zoom", 16);
     });
 
+    it("parses geoURIs", function () {
+      const params = OSM.mapParams("?geouri=geo%3A57.6247%2C-3.6845");
+      expect(params).to.have.property("lat", 57.6247);
+      expect(params).to.have.property("lon", -3.6845);
+      expect(params).to.have.property("mlat", 57.6247);
+      expect(params).to.have.property("mlon", -3.6845);
+      expect(params).to.have.property("zoom", 12);
+    });
+
+    it("parses zoom in geoURIs", function () {
+      const params = OSM.mapParams("?geouri=geo%3A57.6247%2C-3.6845%3Fz%3D16");
+      expect(params).to.have.property("lat", 57.6247);
+      expect(params).to.have.property("lon", -3.6845);
+      expect(params).to.have.property("mlat", 57.6247);
+      expect(params).to.have.property("mlon", -3.6845);
+      expect(params).to.have.property("zoom", 16);
+    });
+
+    it("parses uncertainty in geoURIs", function () {
+      const params = OSM.mapParams("?geouri=geo%3A57.6247%2C-3.6845%3Bu%3D100");
+      const expected = L.latLngBounds([57.62290336944585, -3.6878552857327764], [57.62649663055414, -3.6811447142672233]);
+      expect(params).to.have.property("mlat", 57.6247);
+      expect(params).to.have.property("mlon", -3.6845);
+      expect(params).to.have.property("mrad", 100);
+      expect(params).to.have.property("bounds").deep.equal(expected);
+    });
+
     it("parses lat/lon/zoom from the hash", function () {
       location.hash = "#map=16/57.6247/-3.6845";
       const params = OSM.mapParams("?");
--- a/test/javascripts/osm_test.js
+++ b/test/javascripts/osm_test.js
@@ -130,6 +130,99 @@ describe("OSM", function () {
     });
   });
 
+  describe(".parseGeoURI", function () {
+    it("parses basic geoURIs", function () {
+      let params = OSM.parseGeoURI("geo:57.6247,-3.6845");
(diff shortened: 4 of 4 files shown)


AntonKhorev:
`test/javascripts/osm_test.js` has tests for `mapParams`. If new parameters are added, you should test them.

app/assets/favicons/manifest.json.erb
```diff
@@ -10,5 +10,11 @@
 	"start_url": "/",
 	"theme_color": "#7ebc6f",
 	"background_color": "#fff",
-	"display": "minimal-ui"
+	"display": "minimal-ui",
+	"protocol_handlers": [
+		{
+			"protocol": "geo",
+			"url": "<%= Settings.server_protocol + "://" + Settings.server_url %>?geouri=%s"
```
AntonKhorev, line 17: Is `<%= Settings.server_protocol + "://" + Settings.server_url %>` required? I don't see url protocols/hosts in examples [1](https://developer.chrome.com/docs/web-platform/best-practices/url-protocol-handler) [2](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/protocol_handlers)

app/assets/javascripts/osm.js.erb
```diff
@@ -80,6 +80,11 @@ OSM = {
 
     const hash = OSM.parseHash(location.hash);
 
+    const geoURI = Object.fromEntries(new URLSearchParams(
+      params.geouri?.replaceAll(/[?;]/g, "&")?.replace(":", "=")
```
AntonKhorev, line 84: What you want to do requires zero regexes and replaces. Start with `new URL(params.geouri)`.

hlfan: How does that help me with parsing semicolon-separated query parameters?

hlfan: Also, the URL constructor throws a TypeError when no URI is given, instead of returning empty, which is what the current variant does without needing extra conditions

AntonKhorev: How many semicolon-separated parameters do you need to parse?

AntonKhorev: like `:`, `&` is possible inside geouri parameters https://datatracker.ietf.org/doc/html/rfc5870#section-3.3

hlfan: > Are you going to look at anything past the first semicolon?

In RFC 5870 uncp comes after crsp, so to parse `geo:57.6247,-3.6845;crs=wgs84;u=100`, I need to.
> like `:`, `&` is possible inside geouri parameters

Well then `new URL()` isn't going to help me either.

AntonKhorev: `new URL()` is going to help you to get `z` and to get everything between `geo:` and `?` without writing things like `slice(4)`.

Do you plan to show the uncertainty? Because if not, maybe you don't need `u`. And if you need `u`, surely there's a simpler regexp just for `u` than `/;([\w-]+)=([\w-.!~*'()[\]:&+$%]+)/g`.

hlfan: So do you want me to follow RFC 5870 or not?
https://github.com/openstreetmap/openstreetmap-website/pull/5736#discussion_r1979882902 indicates yes, https://github.com/openstreetmap/openstreetmap-website/pull/5736#discussion_r1976402137 indicates no

AntonKhorev: If you don't need `u`, you don't need any regexps.

If you need `u`, that parameter can't have `&` inside and you can have a simpler regexp. But do you need `u`? Are you going to show the uncertainty?

If you want to parse *every* parameter, you'll need a complicated regexp. What do you need other parameters for?

`z` is not a RFC 5870 parameter.

app/assets/javascripts/osm.js.erb
```diff
@@ -100,6 +106,9 @@ OSM = {
       mapParams.lon = params.mlon;
       mapParams.lat = params.mlat;
       mapParams.zoom = params.zoom || 12;
+    } else if (geoURI.pathname) {
+      [mapParams.lat, mapParams.lon] = geoURI.pathname.split(",");
+      mapParams.zoom = geoURI.searchParams?.get("z") || Math.round(geoURI.pathname.replaceAll(/,?[-\d]+\./g, "").length / 0.6) - 2;
```
AntonKhorev, line 111: Is `Math.round(...)` the inverse of `OSM.zoomPrecision()`?

app/assets/javascripts/osm.js.erb
```diff
@@ -100,6 +106,9 @@ OSM = {
       mapParams.lon = params.mlon;
       mapParams.lat = params.mlat;
       mapParams.zoom = params.zoom || 12;
+    } else if (geoURI.pathname) {
+      [mapParams.lat, mapParams.lon] = geoURI.pathname.split(",");
```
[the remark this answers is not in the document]
hlfan: I thought we wouldn't parse these parameters with the move to `new URL` and parseFloat takes care of these things anyway.

app/assets/javascripts/osm.js.erb
```diff
@@ -63,13 +63,19 @@ OSM = {
...
 
     if (params.mlon && params.mlat) {
       mapParams.marker = true;
       mapParams.mlon = parseFloat(params.mlon);
       mapParams.mlat = parseFloat(params.mlat);
     }
+    if (geoURI.lon && geoURI.lat) {
+      mapParams.marker = true;
+      mapParams.mlon = parseFloat(geoURI.lon);
+      mapParams.mlat = parseFloat(geoURI.lat);
```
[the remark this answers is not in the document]
hlfan: I'd say show a marker so later when uncertainty support is implemented in an approved way,
changing L.marker to L.circle when a radius is specified in https://github.com/openstreetmap/openstreetmap-website/blob/a28f78c69a19c9c41dc96dcaad9f87cf5b530f96/app/assets/javascripts/index.js#L225 doesn't alter the UX too much. Otherwise it would be more confusing.

app/assets/javascripts/osm.js.erb
hlfan: Not entirely,
- replaceAll takes the decimal places
- and this / 0.6 - 2 is close enough to the inverse of `OSM.zoomPrecision()`

I removed `Math.round(...)` to make this more clear.

app/assets/javascripts/osm.js.erb
```diff
@@ -63,12 +63,28 @@ OSM = {
 
   mapParams: function (search) {
     const params = OSM.params(search),
+          geoURI = URL.canParse(params.geouri) ? new URL(params.geouri) : { pathname: "", searchParams: [] },
           mapParams = {};
 
+    [geoURI.coordstring, ...geoURI.params] = geoURI.pathname.split(";");
+    geoURI.params = Object.fromEntries([...geoURI.searchParams, ...geoURI.params.map(p => p.split("=").map(decodeURIComponent))]);
```
AntonKhorev, line 70: Why would you go through this extra effort of merging url and geo parameters? How about urls like `geo:59.95121,30.37522;z=ignore?z=15`?

app/assets/javascripts/osm.js.erb
```diff
@@ -63,12 +63,28 @@ OSM = {
...
+    geoURI.coords = L.latLng(geoURI.coordstring.split(","));
+    geoURI.epsg = geoURI.params.crs?.toUpperCase()?.replace(":", "");
+    if (geoURI.epsg?.startsWith("EPSG")) {
+      geoURI.coords = L.CRS.EPSG3857.unproject(L.CRS[geoURI.epsg]?.project(geoURI.coords));
+    }
```
AntonKhorev, line 75: Is it just an excuse to parse all geo parameters or something that's actually supposed to work?

app/assets/javascripts/osm.js.erb
```diff
@@ -80,6 +80,10 @@ OSM = {
 
     const hash = OSM.parseHash(location.hash);
 
+    const geoURI = Object.fromEntries(new URLSearchParams(
+      params.geouri?.replaceAll(/[?;]/g, "&")?.replaceAll(":", "=")
```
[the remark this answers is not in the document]
AntonKhorev: Looks like the example I copied from Wikipedia is wrong because `crs` values can't contain `:`.

```
crsp          = ";crs=" crslabel
crslabel      = "wgs84" / labeltext
labeltext     = 1*( alphanum / "-" )
alphanum      = ALPHA / DIGIT
```

Nonetheless other parameters can contain `:`.

app/assets/javascripts/osm.js.erb
AntonKhorev: `crs` can't even be `EPSG:something` according to the current rfc https://github.com/openstreetmap/openstreetmap-website/pull/5736#discussion_r2001173709

hlfan:
I'm now choosing the zoom parameter differently so even if a parameter value ending with `&z` gets parsed wrong (it couldn't be in the parameter name anyway and `=` isn't allowed in values) that won't affect the ability to use the last valid value (to prefer `?z` over `;z`).

app/assets/javascripts/osm.js.erb
hlfan: I'm now using `URLSearchParams.getAll` to get the last zoom parameter that can be parsed as an integer.

app/assets/javascripts/osm.js.erb
```diff
@@ -93,6 +100,11 @@ OSM = {
       mapParams.lon = params.get("mlon");
       mapParams.lat = params.get("mlat");
       mapParams.zoom = params.get("zoom") || 12;
+    } else if (geoURI.has("geo") && geoURI.has("u") && !geoURI.has("z")) {
+      mapParams.bounds = L.latLng(geoURI.get("geo").split(",")).toBounds(geoURI.get("u") * 4);
+    } else if (geoURI.has("geo")) {
+      [mapParams.lat, mapParams.lon] = geoURI.get("geo").split(",");
+      mapParams.zoom = geoURI.getAll("z").reduce((out, cur) => parseInt(cur, 10) || out, 12);
```
AntonKhorev, line 107: Because you won't give up on merging different parameters, you are:
- reading zoom off the geouri parameters, despite it's not defined there and you can't know what it means if you find it
- treating 0 as an invalid zoom value
- reading `u` from url parameters, again not knowing what it means; try `geo:60,30?u=no`

hlfan: I think you've added enough constraints that I have to switch to an overly explicit implementation. Have fun pinpointing the tests you don't like.

AntonKhorev: See https://github.com/openstreetmap/openstreetmap-website/pull/5736#discussion_r1979882902. I can come up tests that fail bacause of `getAll` (or you might say because of trying to parse geo params using `URLSearchParams`, but that might have worked because there are extra requirements for the order of geo parameters). `getAll` was introduced as a workaround to allow parameter merging. After merging was removed, more `getAll`s got added.

`geo:60,30;u=1000;v=&u`

app/assets/javascripts/osm.js.erb
```diff
@@ -121,6 +135,27 @@ OSM = {
...
+    const [path, ...params] = url.pathname.split(";");
+    let coords;
+    try {
+      coords = L.latLng(path.split(","));
+    } catch { return; }
+    if (!coords) return;
+    const searchParams = new URLSearchParams(params.join("&"));
+    const crs = searchParams.get("crs");
+    if (crs && crs !== "wgs84") return;
+    const uncertainty = parseFloat(searchParams.get("u"));
```
AntonKhorev, line 151: This still won't work correctly because geo params are case insensitive. But the fix should be easy, just convert the entire geo uri to lowercase.

hlfan: As long as there aren't case sensitive parameter values that need to be handled differently. But I don't see that happening soon

app/assets/javascripts/osm.js.erb
```diff
@@ -121,6 +135,30 @@ OSM = {
...
+      url = new URL(geoURI.toLowerCase());
+    } catch (e) { return; }
+    if (url.protocol !== "geo:" || !url.pathname) return;
+    const [path, ...params] = url.pathname.split(";");
+    let coords;
+    try {
+      coords = L.latLng(path.split(","));
+    } catch (e) { return; }
+    if (!coords) return;
+    const searchParams = new URLSearchParams(params.join("&"));
```
AntonKhorev, line 151: I checked Google Maps and looks like `z` is case sensitive, so it's better to move `toLowerCase` here:

```suggestion
    const searchParams = new URLSearchParams(params.join("&").toLowerCase());
```

`url.protocol !== "geo:"` should still work because the `URL` class converts it to lowecase on its own.

Zverik: They do `toLowerCase()` in line 142.

hlfan: And if we're gonna copy every implementation detail, then #1807 would like to have a word too.

AntonKhorev: @Zverik I said "to move ... here", obviously from line 142.

If anyone knows about any implementation that treats `z` as case-insensitive, please tell. There's no formal specification for it as far as I know.

AntonKhorev:
I'm merging this but I couldn't get it fully working, as well as pwas with protocol handler.

It should work in Chromium on Linux, the browser sees that the pwa has the protocol handler:
image

Then I think it should create a `.desktop` file, which happens for me, but the desktop file doesn't contain the `MimeType` entry to associate with the protocol. I go it working by manually creating a `.desktop` file with `MimeType` and `Exec=/snap/bin/chromium --profile-directory=Default --app=http://127.0.0.1:3000/?geouri=%u`.

hlfan:
> I couldn't get it fully working

probably because of the [still very low browser compatibility](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/protocol_handlers#browser_compatibility).
Given it works out of the box with Edge, I see this as a soft launch.

PR #5737, opened 2025-03-01
Register startinglocation listener as a named function

AntonKhorev:
The listener is not going to be run once if it's constructed inside the `locateactivate` listener. It's going to be run however many times `locateactivate` happens.

Actually neither this custom event nor `locateactivate` is required, but that's for the next pull request.


tomhughes:
Looks good to me, thanks.

hlfan (approved):

PR #5739, opened 2025-03-01
Use location instead of window.location and document.location

AntonKhorev:
Currently we have in use:
- `location`
- `window.location`
- `document.location`

The only one that's missing is `window.document.location`. These are [mostly the same](https://stackoverflow.com/questions/2430936/whats-the-difference-between-window-location-and-document-location).

We can just use `location` everywhere because it's shorter.


tomhughes:
Looks good to me, thanks.

PR #5740, opened 2025-03-01
Set the environment variable in teaspoon example

AntonKhorev:
See #5672.

It didn't take me long to forget how to run teaspoon, go to the docs, copy the command and see it fail.


tomhughes:
Looks good to me, thanks, though I really want to figure out how to make it happen automatically but so far my attempts have failed.

PR #5741, opened 2025-03-01
Added WebP as new output format for map download

mmd-osm:
We're supporting WebP format for map downloads now. The change is a left over from https://github.com/openstreetmap/openstreetmap-website/pull/5607#issuecomment-2660979843


tomhughes:
Looks good to me, thanks.

PR #5744, opened 2025-03-03
Removes optional using first comments records

nenad-vujicic:
### Description

PR removes optional using of note's first comments records for retrieving note's description and author in Note model file. 

Also, adds note's initializing when created using `note_comments` factory by initializing note's `description`, `user_ip` and `user_id` from note's comment (this is used only when notes are created by `create(:note_comment, ...)`) in unit tests.

### How has this been tested?

Automated unit tests and manual testing on previously created set of notes (with special cases).

### DIFF
--- a/test/models/note_test.rb
+++ b/test/models/note_test.rb
@@ -48,20 +48,29 @@ class NoteTest < ActiveSupport::TestCase
   end
 
   def test_description
-    comment = create(:note_comment)
-    assert_equal comment.body, comment.note.description
+    note = create(:note)
+    assert_equal "Default note's description", note.description
 
-    user = create(:user)
-    comment = create(:note_comment, :author => user)
-    assert_equal comment.body, comment.note.description
+    note = create(:note, :description => "Test description #1")
+    assert_equal "Test description #1", note.description
+
+    comment = create(:note_comment)
+    assert_equal "Default note's description", comment.note.description
+
+    comment = create(:note_comment, :note => build(:note, :description => "Test description #2"))
+    assert_equal "Test description #2", comment.note.description
   end
 
   def test_author
+    user = create(:user)
+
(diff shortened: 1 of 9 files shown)


tomhughes:
I don't think this is the right approach to the tests - a factory that modifies the record's parent seems wrong to me.

Probably the `note_comment` factory shouldn't be defaulting to creating an `opened` comment now, as in the future there won't be any such thing. It should probably default to `commented` and the `note` factory should generate food defaults for the initial open and tests should be adapted to all that.

test/models/note_test.rb
```diff
@@ -48,21 +48,31 @@ def test_closed?
...
+    assert_equal "Default note's description", comment.note.description
+
+    comment = create(:note_comment, :note => build(:note, :description => "Test description #2"))
+    assert_equal "Test description #2", comment.note.description
 
     user = create(:user)
+
     comment = create(:note_comment, :author => user)
-    assert_equal user, comment.note.author
+    assert_equal "Default note's description", comment.note.description
```
tomhughes, line 66: Why move this from `test_author` to `test_description` exactly?

I mean obviously it is now testing the description, but why? It just seems to be duplicating the testing that's already done (unless you expect setting the author to change the behaviour of description) while not actually testing `author` at all anymore?

nenad-vujicic: Thanks, removed!

tomhughes: I wasn't really talking about just that one line, but everything from that point on and the general change of merging the two tests and dropping the testing of author?

tomhughes:
I've made one more small change to the note model `test_author` method to actually the author method on notes and I think this is good now.

tomhughes:
Found another problem in `test/models/issue_test.rb` that needed tweaking so I added that.

PR #5745, opened 2025-03-03
I18n fallback fixes

tomhughes:
A few fixes and cleanups for I18n fallback:

* Disable client side fallback in javascript that isn't doing anything as we only load one locale - note that in the embed view the story is different as we load all locales there.
* Drop an old monkey patch that no longer seems to be needed.
* Fix the Norwegian fallback to work - currently it is being overwritten after it is added.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -21,7 +21,6 @@
 
   I18n.default_locale = OSM.DEFAULT_LOCALE;
   I18n.locale = locale;
-  I18n.fallbacks = true;
 
   // '-' are replaced with '_' in https://github.com/eemeli/make-plural/tree/main/packages/plurals
   const pluralizer = plurals[locale.replace(/\W+/g, "_")] || plurals[locale.split("-")[0]];
--- a/config/initializers/i18n.rb
+++ b/config/initializers/i18n.rb
@@ -29,8 +29,6 @@ I18n::Backend::Simple.prepend(OpenStreetMap::I18n::NormaliseLocales)
 I18n::Backend::Simple.include(I18n::Backend::PluralizationFallback)
 I18n::Backend::Simple.include(I18n::Backend::Fallbacks)
 
-I18n.fallbacks.map("no" => "nb")
-
 I18n.enforce_available_locales = false
 
 if Rails.env.test?
--- a/config/initializers/i18n.rb
+++ b/config/initializers/i18n.rb
@@ -21,19 +21,10 @@ module OpenStreetMap
         super
       end
     end
-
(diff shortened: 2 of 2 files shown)


hlfan:
Shouldn't this affect the change put in place in #5527 too?

app/assets/javascripts/application.js
```diff
@@ -21,7 +21,6 @@
 
   I18n.default_locale = OSM.DEFAULT_LOCALE;
   I18n.locale = locale;
-  I18n.fallbacks = true;
```
hlfan, line 24: Why only remove it here and not in the embed?
https://github.com/openstreetmap/openstreetmap-website/blob/998ef40d19762fc879c9607941890b058c9014b2/app/assets/javascripts/embed.js.erb#L16

tomhughes: As I explained when I opened this PR the embed case is different - there we load `embed.js` which contains only a single string but does contain all languages - we do it that way because language selection has to happen client side.

That said I just realised that although it contains all languages it has already applied fallback to them, so there is no opportunity to do client fallback.

tomhughes: I've removed fallback for the embed case now.

tomhughes: Actually there is still a good reason to have it - to provide fallback when the client side determined language code is not one we support at all. So I have put it back.

AntonKhorev:
How does the Norwegian fallback work? I can see it's working if I replace the locale selection logic with `I18n.locale = 'no'` in `set_locale`

https://github.com/openstreetmap/openstreetmap-website/blob/95b8297955268874bbc4dc657e96f5bc1f0fabb8/app/controllers/application_controller.rb#L204-L214

But is `Locale.available.preferred(preferred_languages)` ever going to return `'no'`?

tomhughes:
There are two written forms of Norwegian - Bokmål (code `nb`) and Nynorsk (code `nn`) but older clients would sometimes send `no` as a preferred language (see aa90ace7b4d1250e2bf4986e9510ef127cebc4ae, d5561dd27a3affe7f395febf4880110f66e2af74 and ece52c818feb85e0a28787f753b17bf95ff68526 for some history where we transitioned from `no` to `nb` on the server) so we had that fallback but at some point it stopped working and we didn't notice.

It does looks like it's true that there's no way to actually select that language now anyway so I guess given that it hasn't been working for some time and nobody has complained we can probably just drop it now.

tomhughes:
I've dropped that fallback now.

AntonKhorev:
Merged, thanks.

Issue #5298, opened 2024-10-31
Contribution activity on user page


kcne:
### Problem

Currently, the OpenStreetMap user profile page lacks a comprehensive contribution activity feed similar to those seen on platforms like GitHub. This means users cannot easily track or showcase their activity across changeset edits, note interactions, diary entries, and GPS trace uploads. This gap limits the visibility of a user's contributions and engagement within the community.

### Description

This is proposal for adding a user activity feed to the OpenStreetMap profile page to provide a detailed view of contributions. This would enhance user interaction by showcasing a summary of recent actions in a structured and engaging format.

**Technical Proposal**
The implementation will be split into multiple Pull Requests (PRs) for incremental development:

**PR-1-1: Database Schema Setup**
- **Add `user_activities` table**:
  - Define columns: `id`, `user_id` (indexed), `activity_type`, `object_type`, `object_id`, `metadata`, `created_at` (indexed).
- **Tests**:
  - Write tests to ensure the schema setup functions correctly.
- **Data Retention Policy**:
  - Establish an archival policy for data older than a specified period (e.g., one year).

**PR-1-2: Model Layer - UserActivity Model and Basic Tracking Logic**
- **Create `UserActivity` Model**:
  - Define core associations (`belongs_to :user`) and basic validations.
- **Model Callbacks for Logging**:
  - Implement `after_create` and `after_update` callbacks in relevant models (`Changeset`, `Note`, `DiaryEntry`, `GPSTrace`) for logging actions.
- **Shared Logic**:
  - Create an `ActivityTrackable` concern for models needing activity tracking logic.
- **Tests**:
  - Ensure the concern and model validations are covered by comprehensive tests.

**PR-2: Controller Layer - User Activity Controller**
- **Create `UserActivityController`**:
  - Implement an `index` action to fetch user activities with support for pagination, filtering, and sorting.
- **Querying**:
  - Utilize scopes in the `UserActivity` model for efficient filtering and preloading of data.
- **Tests**:
  - Write controller tests to validate that scoped queries return correct results.

**PR-3: Frontend Layer - Activity Feed Display**
- **Implement Activity Feed Partial**:
  - Create a partial to display user activities in the profile section with a standardized format.
- **Integration Tests**:
  - Conduct view tests to confirm the feed works as expected.
- **Pagination**:
  - Implement infinite scroll or use cursor-based pagination to maintain consistency across the section.

### Screenshots

[...]

PR #5747, opened 2025-03-04
Add date indexes on gpx_files and notes tables

kcne:
### Description

This PR adds database indexes to improve query performance for the user activity feature as part of #5298. Specifically, it adds concurrent indexes on:

- `gpx_files` table: `[user_id, timestamp]` index to optimize retrieval of user's GPS trace uploads
- `notes` table: `[user_id, created_at]` index to optimize retrieval of user's note activities

These indexes will support efficient querying of user activities without table scans, improving performance for the activity feed implementation.

### How has this been tested?

Testing was performed by:
1. Running the migrations locally to verify successful index creation
2.  Checking the database schema to confirm indexes are present

### DIFF
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -16,10 +16,11 @@
 #
 # Indexes
 #
-#  index_notes_on_description  (to_tsvector('english'::regconfig, description)) USING gin
-#  notes_created_at_idx        (created_at)
-#  notes_tile_status_idx       (tile,status)
-#  notes_updated_at_idx        (updated_at)
+#  index_notes_on_description             (to_tsvector('english'::regconfig, description)) USING gin
+#  index_notes_on_user_id_and_created_at  (user_id,created_at) WHERE (user_id IS NOT NULL)
+#  notes_created_at_idx                   (created_at)
+#  notes_tile_status_idx                  (tile,status)
+#  notes_updated_at_idx                   (updated_at)
 #
 # Foreign Keys
 #
--- a/app/models/trace.rb
+++ b/app/models/trace.rb
@@ -16,9 +16,10 @@
 #
 # Indexes
 #
-#  gpx_files_timestamp_idx           (timestamp)
-#  gpx_files_user_id_idx             (user_id)
-#  gpx_files_visible_visibility_idx  (visible,visibility)
+#  gpx_files_timestamp_idx            (timestamp)
+#  gpx_files_user_id_idx              (user_id)
+#  gpx_files_visible_visibility_idx   (visible,visibility)
+#  index_gpx_files_on_user_id_and_id  (user_id,id)
 #
 # Foreign Keys
 #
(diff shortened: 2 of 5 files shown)


db/migrate/20250304172701_add_notes_user_id_created_at_index.rb
```diff
@@ -0,0 +1,7 @@
+class AddNotesUserIdCreatedAtIndex < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_index :notes, [:user_id, :created_at], :algorithm => :concurrently
```
tomhughes, line 5: I'd suggest adding `:where => "user_id IS NOT NULL"` here so that we don't index anonymous notes.

db/migrate/20250304172700_add_gpx_files_user_id_timestamp_index.rb
```diff
@@ -0,0 +1,7 @@
+class AddGpxFilesUserIdTimestampIndex < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_index :gpx_files, [:user_id, :timestamp], :algorithm => :concurrently
```
tomhughes, line 5: I think it would be to index on `:user_id, :id` here as IDs are a proxy for timestamp. Such an index would already be useful for pagination of a user's traces and can be used for efficient lookup recent traces even if you want to limit by time as this test shows:

```
apis_master=# explain select * from gpx_files where user_id = 1 and timestamp > '2024-01-01' order by id desc limit 10;
                                             QUERY PLAN                                              
-----------------------------------------------------------------------------------------------------
 Limit  (cost=2.51..2.51 rows=1 width=106)
   ->  Sort  (cost=2.51..2.51 rows=1 width=106)
         Sort Key: id DESC
         ->  Index Scan using gpx_files_user_id_idx on gpx_files  (cost=0.28..2.50 rows=1 width=106)
               Index Cond: (user_id = 1)
               Filter: ("timestamp" > '2024-01-01 00:00:00'::timestamp without time zone)
(6 rows)

apis_master=# create index on gpx_files(user_id, id);
CREATE INDEX
apis_master=# explain select * from gpx_files where user_id = 1 and timestamp > '2024-01-01' order by id desc limit 10;
                                                QUERY PLAN                                                 
-----------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..2.50 rows=1 width=106)
   ->  Index Scan Backward using gpx_files_user_id_id_idx on gpx_files  (cost=0.28..2.50 rows=1 width=106)
         Index Cond: (user_id = 1)
         Filter: ("timestamp" > '2024-01-01 00:00:00'::timestamp without time zone)
(4 rows)
```

Adding the index switches to a descending index scan on ID which will also be more or less timestamp order and so long as you have some limit on the number of changesets you will consider it should be reasonably efficient?

db/migrate/20250304172701_add_notes_user_id_created_at_index.rb
kcne: I agree. Fixed.

db/migrate/20250304172700_add_gpx_files_user_id_timestamp_index.rb
kcne: Very smart work around to be honest, would never thought of this myself. Fixed, thank you.

tomhughes:

PR #5749, opened 2025-03-05
Add change counts to history and update icons

AntonKhorev:
It's convenient to see change counts in changeset history lists and we have enough space to add them.

Also our current comment icon is actually a reused *add note* icon, that's why it has a tip in the center and a plus sign that makes no sense.

image

I replaced it with the Bootstrap *chat* icon.

---

With this PR:
image

dark:
image

rtl:
image


tomhughes:
Looks good to me, thanks.

PR #5751, opened 2025-03-05
Rename browse controller to feature queries controller

AntonKhorev:
A simpler version of #4616 and #5024 where I don't try to combine feature queries with map key.

I just rename the former browse controller to `FeatureQueries` as suggested in https://github.com/openstreetmap/openstreetmap-website/pull/5024#issuecomment-2253146205.


tomhughes:
Looks good to me, thanks.

Issue #5670, opened 2025-02-13
Don't classify Coveralls upload server errors as failures


hlfan:
### URL

https://github.com/openstreetmap/openstreetmap-website/actions/workflows/tests.yml

### How to reproduce the issue?

Erm...

1. Make sure the Coveralls API is in maintenance mode or something like that
2. Have a CI task try to upload coverage data to Coveralls

### Screenshot(s) or anything else?

https://status.coveralls.io/incidents/0nz347snpb16

https://github.com/openstreetmap/openstreetmap-website/issues/5668#issuecomment-2657106910

matkoniecz:
Reporting failed job as success would not be really better in this case.

gravitystorm:
> Reporting failed job as success would not be really better in this case.

If I understand correctly, this is not failed jobs but jobs with errors. So if the job runs, but for example the overall coverage is too low, then the job will fail and CI will show a failure. But this is different from the job having an error, for example not being able to connect to coveralls.io.

In short, the coveralls job can return a) success b) failure c) error, and CI can only be 1) success or 2) failure, and the setting is to choose whether  c == 1 or c == 2

So the setting is to decide whether to handle a coveralls error as a failure (as now) or as a success. The risk is that the job might error for reasons under our control (e.g. gem update breaks simplecov) and if it's not causing failures in CI, we might not notice for a long time.

So my proposal is to keep the setting as it is now (coveralls error == CI fail) and we can always change this temporarily if there is a problem with coveralls that we need to work around.

matkoniecz:
> So my proposal is to keep the setting as it is now (coveralls error == CI fail) and we can always change this temporarily if there is a problem with coveralls that we need to work around.

+1, as author comment linked in initial report

I think that `Allow coveralls CI steps to "fail on error"` is not solving or improving anything and I would suggest to close this issue and to not reopen it again in future

matkoniecz:
Can we close this one? It is based on misunderstanding of my comment

gravitystorm:
I've seen the thumbs up from both @matkoniecz and @hlfan for my suggestion to close this, so closing now.

tomhughes:
I really think we need to revisit this - having every PR show as erroneous because coveralls is down for maintenance for long periods is a massive pain for maintainers trying to work out what can be reviewed.

tomhughes:
This is the current coveralls status, and the end time has already been extended at least once:

[...]

Issue #5616, opened 2025-02-05
Move html from javascript to views


gravitystorm:
We have an amount of html building being created using JavaScript code. For example, in `leaflet.share.js` there is a substantial amount of html construction:

https://github.com/openstreetmap/openstreetmap-website/blob/7cf9bf0e553a141acab2afea53842f97eddd0a40/app/assets/javascripts/leaflet.share.js#L37-L53

This makes it harder to find, and much harder to maintain, than having the html in a view like the rest of the site. It's also mostly unnecessary, since html like this isn't dynamic, it's the same for every user.

We should move all of this html into views, and use the javascript only for interactivity like changing content (e.g. taking results of an ajax query and putting them into the html) or to show/hide certain elements after UI interactions.

Satya900:
Please assign me this issue i would like to work on it

HolgerJeromin:
@Satya900  We do [not work with assigns](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#assigning-issues). Just go ahead! :)

AntonKhorev:
> For example, in leaflet.share.js there is a substantial amount of html construction: ...

Was done in https://github.com/openstreetmap/openstreetmap-website/pull/4335/files#diff-2b34a8a6e260ed3c05d38b8250e07e05617d32985d2250826373e7c2bce40d8f.

Was ignored by everyone.

gravitystorm:
> Was ignored by everyone.

That PR is marked as a draft, so I haven't reviewed it.

AntonKhorev:
That PR was marked as draft after it was ignored for months and I gave up on maintaining it.

matkoniecz:
Would  #4335 be reviewed if it would be updated so it works again?

PR #5752, opened 2025-03-05
Remove controls sprite

hlfan:
Put the sprite icons into HTML with JS or Ruby.


hlfan:
Coveralls is 504-ing again... #5670

app/views/site/fixthemap.html.erb
```diff
@@ -22,7 +22,15 @@
...
     <h3 class='fs-5'><%= t "site.welcome.add_a_note.title" %></h3>
     <p><%= t "site.welcome.add_a_note.para_1" %></p>
-    <p><%= t ".how_to_help.add_a_note.instructions_1_html", :note_icon => tag.a(:class => "icon note bg-dark rounded-1") %></p>
+    <p><%= t(
+             ".how_to_help.add_a_note.instructions_1_html",
+             :note_icon => link_to(
+               tag.svg(:width => "20", :height => "20", :fill => "currentColor", :class => "align-top") do
+                 tag.path(:d => "M17 12H15v2H13v2h2v2h2V16h2V14H17V12ZM3 12H6l2 3 2-3h3l1-1V4L13 3H3L2 4v7Z")
+               end,
+               "/note/new"
```
tomhughes, line 31: This should use `new_note_path` rather than hardcoding the link.

app/assets/javascripts/fixthemap.js
```diff
@@ -1,8 +1,8 @@
 $(function () {
   const params = OSM.params();
 
-  let url = "/note/new";
   if (!params.zoom) params.zoom = 17;
-  if (params.lat && params.lon) url += OSM.formatHash(params);
-  $(".icon.note").attr("href", url);
+  if (params.lat && params.lon) {
+    $("a[href*='note']")[0].href += OSM.formatHash(params);
```
tomhughes, line 6: This is rather nasty as it means this code is having to make assumptions about the URL that has been filled in by the server, plus we're hoping that no other links contain `note` anywhere... Can we not mark the element with a class that can be used to pick it out precisely?

app/assets/stylesheets/common.scss
```diff
@@ -309,10 +277,15 @@ body.small-nav {
...
-  .icon {
-    margin: 10px;
-  }
+.leaflet-control.active .control-button {
+  background-color: $vibrant-green;
+}
+
+.geolocate {
+	background: currentColor;
+	clip-path: polygon(20% 50%, 20% 40%, 70% 20%, 80% 20%, 80% 30%, 60% 80%, 50% 80%, 50% 50%);
```
tomhughes, line 288: Indentation is wrong here - think it's using a tab or something?

tomhughes (changes requested):

app/assets/javascripts/leaflet.query.js
```diff
@@ -8,9 +8,23 @@ L.OSM.query = function (options) {
...
+          "x": "12.5",
+          "y": "18.512",
+          "font-family": "Helvetica, sans-serif",
+          "font-weight": "Bold",
+          "font-size": "12"
+        })
+      )
+      .attr("fill", "currentColor")
+      .attr("class", "h-100 w-100")
+      .appendTo(link);
```
HolgerJeromin, line 26: Is this easier to read?
```suggestion
    $(`<svg>
    <path d="M3 1s9 7 9 7c-1.056.587-2.112 1.173-3.168 1.76l3.031 6.615c.346.753.015 1.643-.738 1.989-.753.345-1.643.014-1.989-.739L6.201 11.221C5.134 11.814 4.067 12.407 3 13V1z" />
    <text "x": "12.5" "y": "18.512" 
    "font-family"="Helvetica, sans-serif" "font-weight"="Bold" "font-size"="12"
    "class"="h-100 w-100"
    "fill"="currentColor" 
    >?</text></svg>`)
      .appendTo(link);
```
Probably easier to process by the browser...

hlfan: Probably best to define somewhere else collectively in the html and just reuse, to get the clip path out of the css too.
But where (if not in a sprite)?

AntonKhorev:
The sprite existed mainly because of Leaflet controls. They are created with javascript. You'll start putting more stuff in javascript to avoid using the sprite, and then @gravitystorm will tell you this: #5616.

My plan was to put Leaflet control icons into map layout and refer to them with `<use>` from javascript.

hlfan:
So where should I put the `defs` then?

app/assets/javascripts/leaflet.query.js
HolgerJeromin: Oh, you are right. 
DOM APIs like `innerHTML` and `insertAdjacentHTML` creates valid svg elements. jQuery does not with this api.

AntonKhorev: I'd put the svgs into `app/views/layouts/map.html.erb` but some non-map layout pages like *Dashboard* also use them. So I guess it should be a template that is included by the map layout and those other pages.

hlfan: Doesn't the dashboard use the map layout via the site controller?

AntonKhorev: Dashboard has its own controller `app/controllers/dashboards_controller.rb`.

AntonKhorev:
Is this an intentional change to how active "Show my location" looks?

image -> image

I'm asking because there are other changes that probably weren't intentional.

hlfan:
Matching the active button behaviour in 7a9871c1b3d6317b361af5110eadae97ee3b5725 was intentional, that's why I split it into a separate commit.

But changing the foreground color instead of the background color for locating, querying and notating can be done by changing this line alone:
https://github.com/openstreetmap/openstreetmap-website/blob/1f182c673bc66404934fce5692561befa17c19b4/app/assets/stylesheets/common.scss#L283

AntonKhorev:
> Matching the active button behaviour in [bc50867](https://github.com/openstreetmap/openstreetmap-website/commit/bc508673e788ce8c00cf2f030261f733184c2177) was intentional, that's why I split it into a separate commit.

I asked if you wanted to change how this button looks when active. Did you want to change how this button looks?

> But changing the foreground color instead of the background color for locating, querying and notating can be done by changing this line alone:
> 
> https://github.com/openstreetmap/openstreetmap-website/blob/5fbbc4801b6c1e0ed258526f55043bfcc6955d74/app/assets/stylesheets/common.scss#L283

You changed their active background color. Did you want to do it?

image

You removed the green background from some buttons and added it to another button.

hlfan:
Ah, now I see what you mean.
Fixed that now:
image

AntonKhorev:
If you put the icons in a erb template and include that template from map layout and other pages that use the icons, you'll have all the icons in one place without paths in javascript. Something like this: https://github.com/AntonKhorev/openstreetmap-website/commit/56aa9aff03c8c8608aecd29ff6e70b95c1b433df

hlfan:
Most aren't needed anywhere other than the map layout, so only the zoom/locate icons need to be there.

tomhughes (commented):
I don't see any obvious issues here but I'll leave @AntonKhorev to look at it again as he's been dealing with it.

app/views/layouts/_defs_note_icon.html.erb
```diff
@@ -0,0 +1,3 @@
+<defs>
+  <path id="icon-note" fill="currentColor" d="M17 12H15v2H13v2h2v2h2V16h2V14H17V12ZM3 12H6l2 3 2-3h3l1-1V4L13 3H3L2 4v7Z" />
+</defs>
```
AntonKhorev, line 3: Is it worth doing small templates like this one, just because there's one page that uses this icon but not the rest?

hlfan: I didn't particularly like having the same path written two different ways (with and without tag helpers) in three different files either. I think it makes `welcome` and `fixthemap` a bit easier to read but I can remove the template if wanted.

hlfan: Or maybe make `:note_icon => render :partial`, having the path in two templates but it's linked cleaner.

hlfan: I do think that readability wise this is difficult improve on:
https://github.com/openstreetmap/openstreetmap-website/blob/88d59de7a36b668c6f0ace10800bf4a398e3b1bc/app/views/site/welcome.html.erb#L92
https://github.com/openstreetmap/openstreetmap-website/blob/88d59de7a36b668c6f0ace10800bf4a398e3b1bc/app/views/site/fixthemap.html.erb#L23

hlfan:
Does the icon's `clip-path` CSS property work when looked at in the style inspector?
If the `clip-path` URL doesn't work, we'd need to fall back to specifying the polygon as I previously had.
That works fine for now, but it makes icons with holes in it impossible (like circular target icons often used for user location toggles), and the CSS isn't as nice.

AntonKhorev:
> it makes icons with holes in it impossible

Are you sure about this?

app/views/layouts/_defs_note_icon.html.erb
hlfan: As we don't have a way to load an SVG into a template yet, I'm gonna resolve this until there's progress on #5880

app/views/site/fixthemap.html.erb
```diff
@@ -22,8 +18,11 @@
   <div class='col-sm'>
     <h3 class='fs-5'><%= t "site.welcome.add_a_note.title" %></h3>
     <p><%= t "site.welcome.add_a_note.para_1" %></p>
-    <p><%= t ".how_to_help.add_a_note.instructions_1_html", :note_icon => tag.a(:class => "icon note bg-dark rounded-1",
-                                                                                :title => t("javascripts.site.createnote_tooltip")) %></p>
+    <p><%= map_hash = params[:lat].to_f && params[:lon].to_f ? "#map=#{params[:zoom].to_i || 17}/#{params[:lat].to_f}/#{params[:lon].to_f}" : ""
```
AntonKhorev, line 21: Do you have to change the link in this pull request, if you're not adding this link to the welcome page?

`params[:lat].to_f && params[:lon].to_f` evaluates to `0.0` which is true in Ruby.

hlfan: I can't leave it unchanged as the JS uses the icon class; in an earlier iteration, I did cd1a05af94e88053c24f56404bd03418d2568bc0 instead. Would this work better?

app/views/site/fixthemap.html.erb
```diff
@@ -22,8 +18,11 @@
...
+           t ".how_to_help.add_a_note.instructions_1_html",
+             :note_icon => link_to(render(:partial => "layouts/note_icon"),
+                                   "#{new_note_path}#{map_hash}",
+                                   :title => t("javascripts.site.createnote_tooltip")) %></p>
```
AntonKhorev, line 25: > as the JS uses the icon class

And what's the problem with the icon class?

```suggestion
    <p><%= t ".how_to_help.add_a_note.instructions_1_html",
             :note_icon => tag.a(render(:partial => "layouts/note_icon"),
                                 :class => "icon note d-inline-flex align-bottom bg-dark rounded-1 text-white",
                                 :title => t("javascripts.site.createnote_tooltip")) %></p>
```

hlfan: actually why not?

hlfan: Ok, I initially misunderstood what you meant, but it turned out we were thinking the same thing.

test/system/index_test.rb
```diff
@@ -5,16 +5,16 @@ class IndexTest < ApplicationSystemTestCase
     node = create(:node)
     visit node_path(node)
     assert_selector "#content.overlay-right-sidebar"
-    find(".icon.share").click
+    find(".control-share a").click
```
AntonKhorev, line 8: If we want tests not failing on intermediate commits, [Click links instead of icons in tests](https://github.com/openstreetmap/openstreetmap-website/pull/5752/commits/43c3a7cb8b8e5ce96cbcb15b96f679fbb5b7506c) should be squashed into [Unify active control button behaviour](https://github.com/openstreetmap/openstreetmap-website/pull/5752/commits/8b59047a1cd593e0b077dc503b0ea3a6461ef472).

app/views/site/fixthemap.html.erb
```diff
@@ -22,7 +22,8 @@
   <div class='col-sm'>
     <h3 class='fs-5'><%= t "site.welcome.add_a_note.title" %></h3>
     <p><%= t "site.welcome.add_a_note.para_1" %></p>
-    <p><%= t ".how_to_help.add_a_note.instructions_1_html", :note_icon => tag.a(:class => "icon note bg-dark rounded-1",
+    <p><%= t ".how_to_help.add_a_note.instructions_1_html", :note_icon => tag.a(render(:partial => "layouts/note_icon"),
+                                                                                :class => "icon note",
```
AntonKhorev, line 26: Continuing https://github.com/openstreetmap/openstreetmap-website/pull/5752#discussion_r2021990094

Do you still think you need map control icons in three different places as opposed to one, which you could use like this:

```suggestion
    <% note_icon = capture do %>
      <%= tag.a :class => "icon note", :title => t("javascripts.site.createnote_tooltip") do %>
        <svg width="20" height="20"><use href="#icon-note" /></svg>
      <% end %>
    <% end %>
    <p><%= t ".how_to_help.add_a_note.instructions_1_html", :note_icon => note_icon %></p>
```

(this is ignoring that you changed how this link looks)

hlfan: Well, I do think the icons should only be defined once. In an SVG each, so only the necessary icons are put into the views.
But while #5880 is stalling, would it be ok to define the icons like this?
https://github.com/openstreetmap/openstreetmap-website/blob/cb0e3d3575da000e9dbbde31bc85efc7a7f66a5d/app/views/layouts/_defs_markers.html.erb#L15-L24

test/system/index_test.rb
hlfan: Switching the two worked for me

app/views/site/fixthemap.html.erb
AntonKhorev: > But while https://github.com/openstreetmap/openstreetmap-website/issues/5880 is stalling, would it be ok to define the icons like this?

How is it related to this PR? You're not changing note markers here.

hlfan: I meant a central definition template that is rendered with which icons are needed to not have to load more (like in #5854). Ideally I could inline just the note icon svg without an extra definition (#5880), which would eliminate the need for this template.

app/views/site/fixthemap.html.erb
hlfan: Aaaaand the JS is gone

app/views/site/_add_a_note.html.erb
```diff
@@ -1,6 +1,8 @@
 <p><%= t ".para_1" %></p>
 <p><%= t ".para_2_html",
-         :note_icon => tag.a(:href => new_note_path(:anchor => map_hash(params)),
-                             :class => "icon note bg-dark rounded-1",
-                             :title => t("javascripts.site.createnote_tooltip")),
+         :note_icon => (link_to new_note_path(:anchor => map_hash(params)), :title => t("javascripts.site.createnote_tooltip") do
+                          tag.svg :width => 20, :height => 20, :class => "align-top" do
```
AntonKhorev, line 4: This changes how the icon looks:
image image

To restore the previous look do this:
```suggestion
         :note_icon => (link_to new_note_path(:anchor => map_hash(params)), :title => t("javascripts.site.createnote_tooltip"),
                                                                            :class => "d-inline-block rounded-1 bg-dark text-white align-top" do
                          tag.svg :width => 20, :height => 20, :class => "d-block" do
```

hlfan: The box was only necessary because the icon from the sprite wasn't able to use `currentColor`, so the background had to be changed. Using the native link color makes it clear that it's a link.

AntonKhorev: If the link doesn't have to look like a button, it doesn't need to be an icon.

hlfan: It already doesn't look like a button in dark mode
image

AntonKhorev: It's still roughly the same color at least.

AntonKhorev: We can change the background to `bg-black` for example:
image image

And maybe add `bg-opacity-75`; the actual buttons are black with 0.6 opacity, this should be close enough:
image image

AntonKhorev: This probably looks better.

```suggestion
         :note_icon => (link_to new_note_path(:anchor => map_hash(params)), :title => t("javascripts.site.createnote_tooltip"),
                                                                            :class => "d-inline-block rounded-1 text-white bg-black bg-opacity-75 border border-2 my-n1 align-text-bottom" do
                          tag.svg :width => 20, :height => 20, :class => "d-block" do
```

image image

image image

AntonKhorev:
Merged, thanks.

PR #5753, opened 2025-03-06
Use svg symbols for routing icons

AntonKhorev:
#4901 v2

PR #5754, opened 2025-03-06
Write directions sidebar header using .erb template

AntonKhorev:
Other sidebar pages have at least a heading and some placeholder in their `.html.erb` template, for example [geocoder search](https://github.com/openstreetmap/openstreetmap-website/blob/84aa7f455ae58b4b051368e35dbc43f780b3aa0f/app/views/geocoder/search.html.erb). Directions search is the exception, it's template is [essentially empty](https://github.com/openstreetmap/openstreetmap-website/blob/84aa7f455ae58b4b051368e35dbc43f780b3aa0f/app/views/directions/search.html.erb). Instead everything is written by `directions.js`.

Here I'm moving the directions heading to `.html.erb`.

This is a prerequisite for:
- Also putting routing icons into the erb template. This allows to remove the routing sprite and its css where the icons are referenced by numbers instead of names, and remove the dark mode css invert for the icons because inline icons use current color. See #5753.
- Sort out #4825. The heading also contains the close button. Every other page has its close button in its erb template along with the heading.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -37,7 +37,7 @@ $(function () {
 
     map.setSidebarOverlaid(false);
 
-    $("#sidebar_loader").show().addClass("delayed-fade-in");
+    $("#sidebar_loader").prop("hidden", false).addClass("delayed-fade-in");
 
     // Prevent caching the XHR response as a full-page URL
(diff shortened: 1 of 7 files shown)


tomhughes:
Looks good to me, thanks.

PR #5755, opened 2025-03-06
Use browser default color for visited changeset links in history

### LINKED ISSUE #604: Changeset history - there's no indication to a user which changesets have already been viewed
Take this

http://www.openstreetmap.org/user/SomeoneElse/history

as an example.

Typically, when looking down a changeset list I'd read down the list and open some in a new tab to look at them in more detail.

Normally, a web link changes colour when it has been previously opened.  It is not possible to see which changesets have been opened because no such colour change occurs.

AntonKhorev:
Typically we don't color visited links differently, but I think it would make sense for *history* pages. Most of the changeset links you'll see there are not going to be visited, but you'd like to be able to see those few that were visited.

Looks like there's no Bootstrap class for visited links. I'm reverting to the browser default color for visited changeset links in css.

image
image

Fixes #604.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -635,6 +635,10 @@ tr.turn {
       position: relative;
       z-index: 2; /* needs to be higher than Bootstrap's stretched link ::after z-index */
     }
+
+    a.changeset_id:visited {
+      color: revert !important;
+    }
   }
 
   .changeset_more .loader {


nenad-vujicic (approved):
This works great on my side! Although most modern websites do not implement this feature, between usability (mostly for moderators) and keeping stable visual harmony I vote for usability. Thanks!

gravitystorm:
I'm going to decline this PR. I can understand that it can be useful to see visited links. However, I don't think it's worthwhile diverging from the bootstrap defaults, and in particular having some links behave differently from others.

Thanks @AntonKhorev for the suggestion.

PR #5756, opened 2025-03-06
Increase contrast of comments/changes counters in history

AntonKhorev:
Addresses https://github.com/openstreetmap/openstreetmap-website/issues/5329#issuecomment-2704353857, although it's not a dark mode issue and not a recent change issue. Contrast was low all along.

Before:
image
image

After:
image
image


tomhughes:
Looks good to me, thanks.

PR #5761, opened 2025-03-07
User activity queries

kcne:
### Description

Adds a new UserActivities module that provides structured access to a user's activity history as part of #5298. The module fetches and organizes activities (changesets, diary entries, comments, GPX uploads) by day and type using SQL queries.

Key features:
- Groups activities hierarchically by day and type
- Supports pagination
- Handles visibility flags

This provides a foundation for displaying user activity history in a structured timeline view.

### How has this been tested?

Added test suite in `test/lib/user_activities_test.rb`
   
For the context this is the UI which integrates this changes later and which is planned as part of the next PR:
   
Screenshot 2025-03-07 at 08 11 22

### DIFF
--- a/lib/user_activities.rb
+++ b/lib/user_activities.rb
@@ -1,0 +1,194 @@
+# rubocop:disable Metrics/ModuleLength
+module UserActivities
+  def self.for_user(user_id, limit: 5, offset: 0)
+    ActiveRecord::Base.connection.execute(activity_sql(user_id, :limit => limit, :offset => offset)).to_a
+  end
+
+  def self.count_activities(user_id)
+    ActiveRecord::Base.connection.execute(count_sql(user_id)).first["count"]
+  end
+
+  private_class_method def self.count_sql(user_id)
+    <<~SQL.squish
+      WITH activities AS (
+        #{base_activities_sql(user_id)}
+      )
+      SELECT COUNT(DISTINCT DATE_TRUNC('day', timestamp)) as count
+      FROM activities
+    SQL
+  end
+
+  private_class_method def self.activity_sql(user_id, limit:, offset:)
+    <<~SQL.squish
+      WITH activities AS (
+        #{base_activities_sql(user_id)}
+      ),
+      activity_days AS (
+        SELECT DISTINCT DATE_TRUNC('day', timestamp) as day
+        FROM activities
+        ORDER BY day DESC
+        LIMIT #{limit}
+        OFFSET #{offset}
+      ),
+      activities_by_day AS (
+        SELECT
+          DATE_TRUNC('day', a.timestamp) AS activity_date,
+          a.category,
+          a.activity_type,
+          COUNT(*) as count,
+          JSONB_AGG(
+            JSONB_BUILD_OBJECT(
+              'id', a.activity_id,
+              'reference_id', a.reference_id,
+              'additional_reference_id', a.additional_reference_id,
+              'description', a.description,
+              'source_type', a.source_type,
+              'user_display_name', a.user_display_name,
+              'timestamp', a.timestamp
+            )
+            ORDER BY a.timestamp DESC
+          ) as items
+        FROM activities a
+        INNER JOIN activity_days d ON DATE_TRUNC('day', a.timestamp) = d.day
+        GROUP BY activity_date, a.category, a.activity_type
+      ),
+      grouped_activities AS (
+        SELECT
+          activity_date,
+          JSONB_OBJECT_AGG(
+            CONCAT(category, ':', activity_type),
+            JSONB_BUILD_OBJECT(
+              'count', count,
+              'items', items
+            )
+          ) as activity_groups
+        FROM activities_by_day
+        GROUP BY activity_date
+        ORDER BY activity_date DESC
+      )
+      SELECT * FROM grouped_activities
+    SQL
+  end
+
+  # Extract common activity types
+  private_class_method def self.base_activities_sql(user_id)
+    quoted_user_id = ActiveRecord::Base.connection.quote(user_id)
+    [
+      changeset_sql(quoted_user_id),
+      diary_entry_sql(quoted_user_id),
+      changeset_comment_sql(quoted_user_id),
+      note_comment_sql(quoted_user_id),
+      diary_comment_sql(quoted_user_id),
+      gpx_file_sql(quoted_user_id)
+    ].join(" UNION ALL ")
+  end
+
+  private_class_method def self.changeset_sql(quoted_user_id)
+    <<~SQL.squish
+      SELECT
+        c.created_at AS timestamp,
+        CAST('changeset' AS text) AS category,
+        CAST('opened' AS text) AS activity_type,
+        c.id AS activity_id,
+        c.id AS reference_id,
+        NULL AS additional_reference_id,
+        NULL AS description,
+        'changeset' AS source_type,
+        NULL AS user_display_name
+      FROM changesets c
+      WHERE c.user_id = #{quoted_user_id}
+    SQL
+  end
+
+  private_class_method def self.diary_entry_sql(quoted_user_id)
+    <<~SQL.squish
+      SELECT
+        d.created_at AS timestamp,
+        CAST('diary' AS text) AS category,
+        CAST('diary_entry' AS text) AS activity_type,
+        d.id AS activity_id,
+        d.id AS reference_id,
+        d.title AS additional_reference_id,
+        d.body AS description,
+        'diary' AS source_type,
+        u.display_name AS user_display_name
+      FROM diary_entries d
+      JOIN users u ON u.id = #{quoted_user_id}
+      WHERE d.user_id = #{quoted_user_id}
+      AND d.visible = true
+    SQL
+  end
+
+  private_class_method def self.changeset_comment_sql(quoted_user_id)
+    <<~SQL.squish
+      SELECT
+        cc.created_at,
+        CAST('comment' AS text) AS category,
+        CAST('comment' AS text) AS activity_type,
+        cc.id,
+        cc.changeset_id AS reference_id,
+        NULL AS additional_reference_id,
+        cc.body,
+        'changeset' AS source_type,
+        NULL AS user_display_name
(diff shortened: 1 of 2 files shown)


tomhughes:
Did you try doing this via normal ActiveModel queries before resorting to raw SQL queries?

I certainly don't like the idea of interpolating arguments into the queries - the fact that you called the argument `quoted_user_id` immediately hints at the risks of doing so. Is it not possible to use bound parameters with these queries?

What analysis have you done of the execution plan(s) and the likely performance? Is everything using indexes or are there things which are doing table scans?

kcne:
> Did you try doing this via normal ActiveModel queries before resorting to raw SQL queries?
> 
> I certainly don't like the idea of interpolating arguments into the queries - the fact that you called the argument `quoted_user_id` immediately hints at the risks of doing so. Is it not possible to use bound parameters with these queries?
> 
> What analysis have you done of the execution plan(s) and the likely performance? Is everything using indexes or are there things which are doing table scans?


Thank you for the feedback Tom. I've completely refactored the code to address your concerns:

1. **ActiveRecord Usage**: Now using ActiveRecord for all activity fetching. Only kept raw SQL for the `activity_days` method since the UNION ALL across multiple tables is complex to express in pure ActiveRecord.

2. **SQL Injection Protection**: Eliminated all string interpolation and replaced `quoted_user_id` with proper parameter binding:
   ```ruby
   ActiveRecord::Base.sanitize_sql([sql, { :user_id => user_id, :limit => limit, :offset => offset }])
   ```

3. **Performance Analysis**: Conducted thorough testing with multiple users:
   - Execution time: 9-14ms for users with ~25 activity days
   - Query planning: ~0.15-0.20ms, execution: ~0.15-0.19ms
   - All tables have appropriate indexes except `gpx_files` which has `[user_id,id]` index as agreed in #5747 

This was just the initial implementation. Let me know if further adjustments are needed. Thank you.

lib/user_activities.rb
```diff
@@ -0,0 +1,202 @@
...
+      UNION ALL
+      SELECT DATE_TRUNC('day', changeset_comments.created_at) AS day FROM changeset_comments WHERE author_id = :user_id
+      UNION ALL
+      SELECT DATE_TRUNC('day', note_comments.created_at) AS day FROM note_comments WHERE author_id = :user_id
+      UNION ALL
+      SELECT DATE_TRUNC('day', diary_comments.created_at) AS day FROM diary_comments WHERE user_id = :user_id AND visible = true
+      UNION ALL
+      SELECT DATE_TRUNC('day', gpx_files.timestamp) AS day FROM gpx_files WHERE user_id = :user_id
+    SQL
+  end
```
tomhughes, line 49: The problem with doing the union first, and only applying the order and limit to the result is that we can't use any indexes and will wind up fetching all the user's activity for all time and the sorting the combined result before applying the limit as this plan shows:

```
                                                                       QUERY PLAN                                                                       
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=2981.15..2981.16 rows=5 width=8)
   ->  Sort  (cost=2981.15..2995.09 rows=5575 width=8)
         Sort Key: (date_trunc('day'::text, changesets.created_at)) DESC
         ->  Append  (cost=0.57..2888.55 rows=5575 width=8)
               ->  Index Only Scan using changesets_user_id_created_at_idx on changesets  (cost=0.57..1406.24 rows=4373 width=8)
                     Index Cond: (user_id = 3980)
               ->  Index Scan using diary_entry_user_id_created_at_index on diary_entries  (cost=0.42..5.02 rows=4 width=8)
                     Index Cond: (user_id = 3980)
                     Filter: visible
               ->  Index Only Scan using index_changeset_comments_on_author_id_and_created_at on changeset_comments  (cost=0.43..19.49 rows=96 width=8)
                     Index Cond: (author_id = 3980)
               ->  Index Only Scan using index_note_comments_on_author_id_and_created_at on note_comments  (cost=0.43..151.67 rows=418 width=8)
                     Index Cond: (author_id = 3980)
               ->  Index Scan using index_diary_comments_on_user_id_and_id on diary_comments  (cost=0.41..263.85 rows=268 width=8)
                     Index Cond: (user_id = 3980)
                     Filter: visible
               ->  Index Scan using gpx_files_user_id_idx on gpx_files  (cost=0.43..206.02 rows=416 width=8)
                     Index Cond: (user_id = 3980)
(18 rows)
```

if we use subqueries to sort and limit each branch before the union like this:

```sql
(SELECT DATE_TRUNC('day', changesets.created_at) AS day FROM changesets WHERE user_id = :user_id ORDER BY created_at DESC LIMIT :limit)
UNION ALL
(SELECT DATE_TRUNC('day', diary_entries.created_at) AS day FROM diary_entries WHERE user_id = :user_id AND visible = true ORDER BY created_at DESC LIMIT :limit)
UNION ALL
(SELECT DATE_TRUNC('day', changeset_comments.created_at) AS day FROM changeset_comments WHERE author_id = :user_id ORDER BY created_at DESC LIMIT :limit)
UNION ALL
(SELECT DATE_TRUNC('day', note_comments.created_at) AS day FROM note_comments WHERE author_id = :user_id ORDER BY created_at DESC LIMIT :limit)
UNION ALL
(SELECT DATE_TRUNC('day', diary_comments.created_at) AS day FROM diary_comments WHERE user_id = :user_id AND visible = true ORDER BY created_at DESC LIMIT :limit)
UNION ALL
(SELECT DATE_TRUNC('day', gpx_files.timestamp) AS day FROM gpx_files WHERE user_id = :user_id ORDER BY id DESC LIMIT :limit)
```

then we get a much better plan where we only read N items from each activity class before sorting the result and picking the top N overall items:

```
                                                                                  QUERY PLAN                                                                                  
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=25.51..25.53 rows=5 width=8)
   ->  Sort  (cost=25.51..25.59 rows=29 width=8)
         Sort Key: "*SELECT* 1_1".day DESC
         ->  Append  (cost=0.57..25.03 rows=29 width=8)
               ->  Subquery Scan on "*SELECT* 1_1"  (cost=0.57..2.68 rows=5 width=8)
                     ->  Limit  (cost=0.57..2.18 rows=5 width=16)
                           ->  Index Only Scan Backward using changesets_user_id_created_at_idx on changesets  (cost=0.57..1406.24 rows=4373 width=16)
                                 Index Cond: (user_id = 3980)
               ->  Subquery Scan on "*SELECT* 2"  (cost=0.42..5.42 rows=4 width=8)
                     ->  Limit  (cost=0.42..5.02 rows=4 width=16)
                           ->  Index Scan Backward using diary_entry_user_id_created_at_index on diary_entries  (cost=0.42..5.02 rows=4 width=16)
                                 Index Cond: (user_id = 3980)
                                 Filter: visible
               ->  Subquery Scan on "*SELECT* 3"  (cost=0.43..1.92 rows=5 width=8)
                     ->  Limit  (cost=0.43..1.42 rows=5 width=16)
                           ->  Index Only Scan Backward using index_changeset_comments_on_author_id_and_created_at on changeset_comments  (cost=0.43..19.49 rows=96 width=16)
                                 Index Cond: (author_id = 3980)
               ->  Subquery Scan on "*SELECT* 4"  (cost=0.43..2.74 rows=5 width=8)
                     ->  Limit  (cost=0.43..2.24 rows=5 width=16)
                           ->  Index Only Scan Backward using index_note_comments_on_author_id_and_created_at on note_comments  (cost=0.43..151.67 rows=418 width=16)
                                 Index Cond: (author_id = 3980)
               ->  Subquery Scan on "*SELECT* 5"  (cost=0.41..5.83 rows=5 width=8)
                     ->  Limit  (cost=0.41..5.33 rows=5 width=16)
                           ->  Index Scan Backward using diary_comment_user_id_created_at_index on diary_comments  (cost=0.41..263.85 rows=268 width=16)
                                 Index Cond: (user_id = 3980)
                                 Filter: visible
               ->  Subquery Scan on "*SELECT* 6"  (cost=0.43..4.99 rows=5 width=8)
                     ->  Limit  (cost=0.43..4.49 rows=5 width=16)
                           ->  Index Scan Backward using index_gpx_files_on_user_id_and_id on gpx_files  (cost=0.43..337.88 rows=416 width=16)
                                 Index Cond: (user_id = 3980)
(30 rows)
```

The only real problem is that when you do that you can't support offset as you don't know what offset to start at in each table, but do we actually need offset? We're only interested in the most recent activity? If we do need it than the subqueries would each need to use offset+limit as their limit which is still better until offset gets too large at least.

Note that CTEs are an alternative to inline subqueries here - the result should be the same.

lib/user_activities.rb
```diff
@@ -0,0 +1,202 @@
...
+  # Fetch activities for specific days
+  private_class_method def self.fetch_activities(user_id, days)
+    [
+      fetch_changesets(user_id, days),
+      fetch_diary_entries(user_id, days),
+      fetch_changeset_comments(user_id, days),
+      fetch_note_comments(user_id, days),
+      fetch_diary_comments(user_id, days),
+      fetch_gpx_files(user_id, days)
+    ].flatten
```
tomhughes, line 60: I'm not really sure why joining the relations returned by these functions with `union` or `union_all` is so hard? It would mean one database round trip instead of six?

tomhughes (changes requested):
I'm not sure why you say union is hard? We have https://github.com/brianhempel/active_record_union installed already so you can just use the `union` or `union_all` method on a relation to merge it with another one?

lib/user_activities.rb
kcne: Thank you for the time spent on this Tom. I refactored to use the limits as suggested. Also opted for cursor based pagination to support `load more` functionality later. It should also follow the pattern and use the indexes efficitently.

lib/user_activities.rb
kcne: Fixed.

kcne:
Refactored the code according to suggestions. Please let me know if you have any further suggestions. Thank you.

kcne:
@tomhughes, @AntonKhorev  when you find time to look at this again, just let me know if i need to adjust this further. Thank you.

Issue #3211, opened 2021-06-04
Some things you do with issues do not actually update the updated_* fields


woodpeck:
The "issue" object has an "updated_at" field and an "updated_by" field. These should ideally be changed at the same time, whenever an issue is modified in any way, including when it attracts new reports or new comments.

Instead:

* when an issue is ignored https://github.com/openstreetmap/openstreetmap-website/blob/70d7d8d850148f714b70a3297c02a8203214dec6/app/controllers/issues_controller.rb#L62 or reopened https://github.com/openstreetmap/openstreetmap-website/blob/70d7d8d850148f714b70a3297c02a8203214dec6/app/controllers/issues_controller.rb#L72 this works correctly.
* when an issue is resolved https://github.com/openstreetmap/openstreetmap-website/blob/70d7d8d850148f714b70a3297c02a8203214dec6/app/controllers/issues_controller.rb#L51 then the updated_at changes, but the updated_by does not.
* when a comment is added to it https://github.com/openstreetmap/openstreetmap-website/blob/70d7d8d850148f714b70a3297c02a8203214dec6/app/controllers/issue_comments_controller.rb#L9-L17 then neither the updated_at changes nor the updated_by.
* when a new report comes in for an existing issue, then I *think* the updated_at gets changed but not the updated_by (not sure).

gravitystorm:
>     * when an issue is resolved

I've fixed this in #3217, it's a clear oversight compared to the others actions that already do this.

However, the two other situations need a bit more discussion, namely adding comments and adding new reports. In general, I don't think that we should consider the issue to be updated when a new comment is added or a new report created. I think this is similar to e.g. Diary Entries or Notes or Changesets - I would be surprised if a diary entry was shown as "updated at: now" if I happened to add a comment to a diary entry. So I think we should continue that principle, and mark the updated_at (and updated_by) for issues only if attributes of the Issue are changed.

Would you agree?

In saying that, each of those two situation **can** involve changing the attributes of the issue - due to side effects. Namely, if you add a comment and reassign the issue, that does change the one of the attributes (who it's assigned to i.e. `assigned_role`). 

[...]

PR #5763, opened 2025-03-07
Create last managing user column in issues table

AntonKhorev:
Issues have `updated_at` and `user_updated` attributes. They are updated separately, sometimes without new `user_updated` being written. This makes "Last Updated / (date) by (user)" statements as they are shown in issue tables possibly incorrect:

image

Various solutions were discussed, some in #3211, but it's still not obvious which one is the best. The simplest one however is to stop claiming that the user from `user_updated` did their update at `updated_at`. This can be done by displaying the user in another column:

image

The column title suggested by @SomeoneElseOSM was "last managed by" but I don't want to stretch the table too much horizontally. Or we might need a wider table.

### DIFF
--- a/app/views/issues/_page.html.erb
+++ b/app/views/issues/_page.html.erb
@@ -15,6 +15,7 @@
           <th><%= t ".reported_user" %></th>
           <th class="reporting_users"><%= t ".reporting_users" %></th>
           <th><%= t ".last_updated" %></th>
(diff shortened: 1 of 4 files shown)


tomhughes:
I'm not completely sure about "managed" but I don't have any better ideas for how to describe it, so this looks good to me.

PR #5764, opened 2025-03-07
Convert more markers to SVG

hlfan:
### Description

- Makes all[^1] markers SVGs with a similar structure. (addresses https://github.com/openstreetmap/openstreetmap-website/issues/850#issuecomment-2323133938)
- Changes the note markers' ✔️, ❌ & ➕ to be drawn with a reused stroke instead of areas to reduce file size and to enable an easier switch to SVG `marker` elements for more dynamic pins in the future.
- Continues to use marker-shadow.png from leaflet

[^1]: not the yellow marker as I don't have the colours for that one


AntonKhorev:
Is it more clear from the routing icons, which one is start/finish? Maybe the start icon doesn't need to be changed?

AntonKhorev:
By the way https://antonkhorev.github.io/msvgen/

I'm trying to parametrize the markers by currentColor for situations where several markers are to be displayed at the same time, like search results. Still haven't done gradients.

hlfan:
If you mean you need to create the gradients' stops from currentColor to `use` these like for the directions icons (meaning without JS) that would need color-mix or the relative color functions which have only been available in css for less than a year.
But given that parametrication building these from templates with Ruby could be a big later addition.

hlfan:
> Maybe the start icon doesn't need to be changed?

I think changing just one is more confusing, the directions apps I've seen that have different routing markers do eithet use numbers, letters or symbols like these:
here
(here)
tomtom
(tomtom)

And because osm is international I didn't want to use letters or numbers.

hlfan:
> Still haven't done gradients.

I've also fiddled around a bit, trying to make a marker-based solution [here](https://codepen.io/hlfan/pen/OPJOGxg).
Still trying to get rid of the mask as a gradient workaround.

hlfan:
I've decided to reduce this PR to just the conversion of the SVG pins. While modularizing and parametrizing the markers is a commendable effort, it shouldn't be part of this PR, but the next step instead.
I've removed from this PR:
- the marker renaming
- the two new direction markers
- the resizing to 25x40
- the new marker outline paths since they have a noticeable jump in curvature
- the gradient domain extension

hlfan:
I've experimented with joining the arc and the line with a curve so the curvature matches.
While this makes insetting the paths more difficult, this can be replaced with a `clip-path`.
Instead of a direct join
```
M12.5 41
L1.2664 17.9825
a12.5 12.5 0 1122.4671 0
z
```
it's a bit longer:
```
M12.5 41
l9.56-19.12
C23.53 18.94 25 16 25 12.5
a1 1 0 00-25 0
c0 3.5 1.47 6.44 2.94 9.38
z
```
But it looks much better:
path join detail
And also better than the current path:
image

PR #5765, opened 2025-03-07
Revise search forms

hlfan:
### Description
- Changes search form to explicit display-flex, removing the need for col divs.
- Moves the directions form's routing markers inside the input field to be more compact.

before:
openstreetmap org_directions
after:
openstreetmap org_directions

and in rtl:
openstreetmap org_directions


tomhughes:
Looks good to me, thanks.

PR #5766, opened 2025-03-07
Move a couple of i18n strings to dashboards/contact

AntonKhorev:
dashboards/contact template is the only place where these string are used.


gravitystorm:
Merged, thanks!

PR #5767, opened 2025-03-07
Load more reporting users when clicking "and n more" in issues table

AntonKhorev:
Implemented with Turbo frames.

image image


tomhughes:
Looks good to me, thanks.

PR #5774, opened 2025-03-08
Fix comments/changes counts wrapping on history pages

AntonKhorev:
Long usernames without spaces were causing flex container wraps, pushing comment/edit counts below.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5776, opened 2025-03-08
Remove SVG symbol tags

hlfan:
The routing icons currently use the `symbol` element but use none of the structure and semantics that would justify these.
Also none of the SVG `symbol` specific tags - `width`, `height`, `x`, `y`, `refX`, `refY`, `preserveAspectRatio`, or `viewBox` - are used.
This PR removes this unnecessary extra layer introduced in #5753 and uses basic groups if needed.


AntonKhorev:
> I also run the path data through svgo.

Why? Paths here are hand-written, you're not going to save much but you'll make them less readable.

PR #5768, opened 2025-03-07
Add right u-turn for GraphHopper and Valhalla

AntonKhorev:
image [[1]](https://www.openstreetmap.org/directions?engine=graphhopper_car&route=51.444388%2C-0.234680%3B51.444275%2C-0.234637)

PR #5777, opened 2025-03-08
Add latent OSRM support for left uturn

hlfan:
This adds #5768 for OSRM, which will be apparent when fossgis make up their mind about supporting this.

### DIFF
--- a/app/assets/javascripts/index_modules/directions/engines/fossgis_osrm.js
+++ b/app/assets/javascripts/index_modules/directions/engines/fossgis_osrm.js
@@ -93,31 +93,31 @@
     }
 
     function _processDirections(leg) {
-      function getManeuverId({ maneuver, mode, intersections }) {
+      function getManeuverId({ maneuver: { type, modifier }, mode, intersections }) {
         // special case handling
         if (mode === "ferry") return "ferry";
         if (intersections.some(i => i.classes?.includes("ferry"))) return "ferry";
 
-        switch (maneuver.type) {
+        switch (type) {
           case "on ramp":
           case "off ramp":
           case "merge":
           case "end of road":
           case "fork":
-            return maneuver.type + " " + (maneuver.modifier.indexOf("left") >= 0 ? "left" : "right");
+            return type + " " + (modifier.includes("left") ? "left" : "right");
           case "depart":
           case "arrive":
           case "roundabout":
           case "rotary":
           case "exit roundabout":
           case "exit rotary":
-            return maneuver.type;
+            return type;
           case "roundabout turn":
           case "turn":
-            return "turn " + maneuver.modifier;
+            return "turn " + modifier;
             // for unknown types the fallback is turn
           default:
-            return "turn " + maneuver.modifier;
+            return "turn " + modifier;
         }
       }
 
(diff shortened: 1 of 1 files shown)


tomhughes:
Does this really work? The difference between the bearings tells you how close to parallel the two lanes are, not which way you turned?

hlfan:
The bearing after is on the crossing road. Otherwise it wouldn't work.

AntonKhorev:
https://github.com/Project-OSRM/osrm-backend/issues/3077#issuecomment-348926341 suggests looking at `driving_side`, but looks like it's set to `right` even in the UK.

AntonKhorev:
Comment about `driving_side` for reference: https://github.com/Project-OSRM/osrm-backend/issues/3077#issuecomment-348926341

hlfan:
So should I replace the working bearing-based implementation with a non-working one based on `driving_side` just because it's the intended way?

AntonKhorev:
> working bearing-based implementation

image
[[place]](https://www.openstreetmap.org/#map=19/51.195720/4.501449)

It's not uncommon for the difference between the bearings to be exactly 180. How are you going to decide on the u-turn direction in this case?

hlfan:
Making a wrong decision [_sometimes_](https://www.youtube.com/watch?v=GiYO1TObNz8) is still better than systematically ignoring a choice and being wrong in left-driving countries the whole time.

hlfan:
It still fails in places like [here](https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=53.411809%2C-2.965858%3B53.412402%2C-2.966225), where that step has:
```json
{
  "maneuver": {
    "bearing_after": 261,
    "bearing_before": 265,
    "modifier": "uturn",
  },
  "driving_side": "right"
}
```
and the `driving_side` is wrong more often.

AntonKhorev:
What was [this change](https://github.com/openstreetmap/openstreetmap-website/pull/5777/files/e7db3420dc6fa8a1d34e45a99c12c6f9255e6a89..2db8892dc5f1b3fd2248265c9a06e472b825db23) supposed to achieve?

How about making no decision because you can't make a correct one? If OSRM builds a route going to a dead end with u-turn and then going back, there's no way guess the u-turn direction even if you analyze the entire route geometry.

The best way to get u-turn directions is probably doing what OSRM is supposed to do: checking if u-turns inside a left/right-hand driving country.

1ec5:
> and the `driving_side` is wrong more often.

As noted in https://github.com/Project-OSRM/osrm-backend/issues/2269#issuecomment-357154269, a GeoJSON of [left-hand-drive regions](https://github.com/Project-OSRM/osrm-backend/blob/4ee9968e3b585ee8ef28aec4a7fe3cfdea1f9ea2/data/driving_side.geojson) needs to be passed into `osrm-extract`. Otherwise, `driving_side` will always fall back to the default value of `right`. This seems to be the current behavior of the FOSSGIS OSRM server.

1ec5:
I contacted the FOSSGIS routing server’s e-mail address to ask them to look into the missing parameter when running `osrm-extract`.

1ec5:
The repository responsible for running `osrm-extract` doesn’t accept bug reports, so I filed one in fossgis-routing-server/cbf-routing-profiles#24 instead.

tomhughes:
So as I understand it as this PR now stands it's effectively a no-op in that the server always returns `right` which means we will continue selecting the same icon we do now.

I think therefore that this is good to merge - nothing will change but if the server is ever fixed then we will start doing the right thing.

PR #5778, opened 2025-03-08
Sticky close button v2

AntonKhorev:
Previous version: #4825

The changes are:
- https://github.com/openstreetmap/openstreetmap-website/pull/4825#issuecomment-2123914398, light gray background when scrolled, but with a shadow instead of a border for some separation.
- https://github.com/openstreetmap/openstreetmap-website/pull/4825#issuecomment-2131233781, the transparent area around the button was supposedly suggesting that it's clickable, therefore I removed it. Now I'm using a smaller non-transparent area that is clickable.

So I went from a larger transparent area to a smaller area with a shadow to indicate that the button is floating above the scrolled sidebar content.

Looks as usual when the sidebar contents is not scrolled:
image

Gets a slightly gray background and a shadow when scrolled:
image

Focus:
image

Dark mode:
image

Right sidebar:
image

Fixes https://github.com/openstreetmap/openstreetmap-website/issues/2174.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -398,7 +398,7 @@ $(function () {
     }
   });
 
-  $(document).on("click", "#sidebar_content .btn-close", function () {
+  $(document).on("click", "#sidebar .sidebar-close-controls button", function () {
     OSM.router.route("/" + OSM.formatHash(map));
   });
 });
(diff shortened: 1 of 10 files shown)


tomhughes:
Looks good to me, thanks.

Issue #3888, opened 2023-01-10
Allow translation of distances in directions output


maro-21:
Where: https://www.openstreetmap.org/directions

There are no spaces between numbers and units.
Ideally it should be a non-breaking space.
https://english.stackexchange.com/a/2911

Example:
https://osm.org/directions?engine=graphhopper_car&route=40.7182%2C-73.8983%3B40.7522%2C-74.1812

![spaces](https://user-images.githubusercontent.com/57260466/211626313-668c4da0-a89b-46dc-9a7d-245c73f21634.png)

tomhughes:
Why do you think there should be a space? It wouldn't be normal in English.

tomhughes:
I mean a space isn't wrong. but neither is not having one in my experience. Maybe @systemed has some copy editor thoughts on the subject...

systemed:
It's more usual (though not universal) in British English to omit the space. See, for example, https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#measurements.

maro-21:
So allow to add spaces in other languages, because in Polish it is incorrect.

maro-21:
I think in English there should be a space too. Because meters are SI units. I don't know how it is with feet or how is it in Asian languages.
https://english.stackexchange.com/questions/15953/how-should-i-write-units
https://en.wikipedia.org/wiki/International_System_of_Units#Unit_symbols_and_the_values_of_quantities

PR #5781, opened 2025-03-09
Translate step distance units

AntonKhorev:
There was an issue #3888 that asked to make directions distances translatable. It was done, but not entirely. Step distances were still untranslatable after the issue was closed.

Before in Russian:

image
Notice that the total distance units are "км", but the first step has untranslated "km".

After:

image

---

I undid some of the changes from #5633 that made it difficult to find where strings like `distance_km` are used. To do that I had to disable `no-else-return` eslint rule introduced in #5643. This is not one of the [recommended rules](https://eslint.org/docs/latest/rules/).

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -80,6 +80,14 @@ OSM.Directions = function (map) {
     }
   }
 
+  function getDistText(dist) {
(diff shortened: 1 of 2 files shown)


tomhughes:
I was in two minds about those changes when I reviewed #5633 so this looks good to me, thanks.

PR #5783, opened 2025-03-09
Allow user list to be filtered for the presence of edits

tomhughes:
This allows the user list that administrators use to be filtered by whether a user does, or does not, have edits as requested by @Firefishy.

### DIFF
--- a/app/views/users/lists/show.html.erb
+++ b/app/views/users/lists/show.html.erb
@@ -32,6 +32,13 @@
                          :class => "form-control" %>
     </div>
     <div class="mb-3 col-md-auto">
+      <%= select_tag :edits,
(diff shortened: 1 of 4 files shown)


app/views/users/lists/show.html.erb
```diff
@@ -31,6 +31,13 @@
                          :autocomplete => "on",
                          :class => "form-control" %>
     </div>
+    <div class="mb-3 col-md-auto">
+      <%= select_tag :edits,
+                     options_for_select([["Has Edits", "yes"], ["No Edits", "no"]], params[:edits]),
```
AntonKhorev, line 36: There's no `t("...")` for "Has Edits", "No Edits", although there is `t(".edits")` on the next line.

tomhughes: Fixed.

AntonKhorev:
Merged, thanks.

PR #5786, opened 2025-03-10
Order issue reports by id

AntonKhorev:
Reports on `/issues/:id` weren't sorted in any particular order. This is usually unnoticeable because typically they come in the creation order from the db. But that doesn't always happen.

image

This PR sorts the reports by id explicitly.

I don't know if we need an index on (issue_id, report_id). If there aren't many reports per issue, maybe not.

Same thing is probably required for issue comments.


tomhughes:
Looks good to me, thanks. As you say, we probably don't need to worry about an index for now given the small number of records involved.

PR #5787, opened 2025-03-10
Show dates of reported items on issue pages

AntonKhorev:
Allows to see if the reported item wasn't created/changed recently. For reported users shows only the account creation date.

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5789, opened 2025-03-11
Update Gemfile

Artyom10-Pt:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,81 +1,31 @@
 source "https://rubygems.org"
 
-# Require rails
+# Core dependencies
 gem "rails", "~> 7.2.0"
 gem "turbo-rails"
+gem "pg"  # PostgreSQL database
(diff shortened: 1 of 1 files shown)


tomhughes:
Thanks, but this is something we do routinely and we don't really need people filing PRs for it.

tomhughes:
Oh sorry I thought you mean updating the gems in the lock, but even so this looks like you're just reorganising it in line with your personal preferences.

If you want to do something like that you need to provide more of a rationale and not just say "updating gemfile" with no explanation of what you're trying to achieve.

gravitystorm:
Hi @Artyom10-Pt thanks for your PR. It looks like you've removed lots of gems that we rely on, like bootstrap. You can see more information in the output of the CI checks.

If you'd like to tell us a bit more about what you are trying to do, we can help you with this PR.

PR #5790, opened 2025-03-11
Fix overlap between left sidebar heading and close button

AntonKhorev:
Sticky close buttons have a layer that covers their shadows when sidebars are not scrolled. That layer is slightly larger than it needs to be. It overlaps with the heading and can block out letters in the heading. I'm making the cover layer smaller in this PR.

Before:
image
image

After:
image
image

Actually with this PR the cover layer is slightly smaller than it needs to be to fully block the shadow, but hopefully it's unnoticeable. If it becomes a problem, I'll have to set sizes directly in css, currently they are set in Bootstrap units. There's also a tradeoff between not having this overlap and the heading width, given that the heading content is variable.

### DIFF
--- a/app/views/layouts/_sidebar_close.html.erb
+++ b/app/views/layouts/_sidebar_close.html.erb
@@ -9,6 +9,6 @@
   </div>
 </div>
 <div class="sidebar-close-controls position-relative">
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #5791, opened 2025-03-11
Convert reportable_dates helper into heading helper

AntonKhorev:
In #5787 I appended `created on %{datetime_created} ...` to the reported item in the heading. That can't be translated correctly because I'm constructing a sentence and "created"/"updated" need to agree with the subject in grammatical gender etc. The entire subheading needs to be translated together.

I'm also removing the double "Note":

Before:
image

After:
image


tomhughes:
Looks good to me, thanks.

PR #5793, opened 2025-03-11
User list fixes

tomhughes:
Some fixes for the user list view:

* Fix a server error if the user hits confirm or hide without selecting any users
* Preserve the user and edit filters across a confirm or hide action
* Drop redundant parameters from a form URL that duplicate hidden fields


AntonKhorev:
What does "X" mean? d715f5b7bbc9d27951006322462bab0b05a2daa4

tomhughes:
> What does "X" mean? [d715f5b](https://github.com/openstreetmap/openstreetmap-website/commit/d715f5b7bbc9d27951006322462bab0b05a2daa4)

Sorry, that was meant to be squashed into the previous commit but it seems I forgot to change the action when I rebased it... Fixed now.

PR #5796, opened 2025-03-12
Monkey patch Selenium to allocate unique bidi ports

tomhughes:
Firefox 136 has started binding an extra port at startup when running via selenium but unfortunately it always uses port 9222 which causes problems running tests in parallel as discussed in https://github.com/mozilla/geckodriver/issues/2218.

This monkey patches things to try and allocate unique ports.

I believe this is the cause of the sporadic test failures we've been seeing the last few days - on my machine it happens basically every time since 136 came out but I probably have much high parallelism than Actions.

### DIFF
--- a/config/initializers/selenium.rb
+++ b/config/initializers/selenium.rb
@@ -1,0 +1,35 @@
+if Rails.env.test?
+  require "active_support/testing/parallelization"
+
+  module OpenStreetMap
+    module Selenium
+      module BidiPort
+        module ClassMethods
+          attr_accessor :websocket_port
+        end
+
+        def self.prepended(base)
+          class << base
+            prepend ClassMethods
+          end
+
+          base.websocket_port = 10000
+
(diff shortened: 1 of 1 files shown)


tomhughes:
Well that doesn't seem to have worked very well... Will switch this to draft for now.

hlfan:
Could Firefox be fixed to v135 in the meantime?

tomhughes:
I don't even know for sure that GHA has 136 yet. I'm just assuming that is the problem given when it started happening but in any case as far as I know there's no way to control the firefox version in the GHA images.

tomhughes:
So it seems that the process ID based solution doesn't work very well in actions so I've updated to a more robust solution based on the worker number and also incrementing to handle cases where a worker process starts more than one instance.

tomhughes:
I don't normally self-merge things but nobody has spotted any problems with this and it's getting annoying seeing a sea of red in the PR queue so I'm going to merge and if anybody spots a problem we can fix it later.

PR #5797, opened 2025-03-13
Changeset subscription model

### LINKED ISSUE #5540: Create a ChangesetSubscription model
We have various subscription models, such as NoteSubscription and DiaryEntrySubscription. We also have a table "changeset_subscribers" which does the same thing (pairs of user_id/object_id), but we don't have a corresponding model. Instead we declare it as a join association in Changesets, and use custom methods on the changeset model (e.g. `changeset.subscribe(user)`) to indirectly add and remove the rows in that table.

I think it would be worth refactoring this code, to make subscriptions behave the same way across all three models. So we should

* Create a ChangesetSubscription model, although it will need `self.table_name = "changeset_subscribers"`
* Refactor `changeset_subscriptions_controller` to use the ChangesetSubscription model, likely through changeset.subscriptions.new
* Remove the subscribe/unsubscribe methods from the Changeset model, since these aren't acting on the Changeset model itself.

AntonKhorev:
Fixes #5540.

### DIFF
--- a/app/controllers/api/changesets_controller.rb
+++ b/app/controllers/api/changesets_controller.rb
@@ -82,7 +82,7 @@ module Api
       cs.save_with_tags!
 
       # Subscribe user to changeset comments
-      cs.subscribe(current_user)
+      cs.subscribers << current_user
 
       render :plain => cs.id.to_s
     end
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -63,7 +63,7 @@ class User < ApplicationRecord
   has_many :preferences, :class_name => "UserPreference"
   has_many :changesets, -> { order(:created_at => :desc) }, :inverse_of => :user
   has_many :changeset_comments, :foreign_key => :author_id, :inverse_of => :author
-  has_and_belongs_to_many :changeset_subscriptions, :class_name => "Changeset", :join_table => "changesets_subscribers", :foreign_key => "subscriber_id"
+  has_many :changeset_subscriptions, :foreign_key => :subscriber_id
```
tomhughes, line 66: Technically this changes the association to return subscription records rather than changeset record - based on the name that is likely correct but presumably it means nothing is using it...

tomhughes: That does indeed seem to be the case so I think this is probably good as it is.

gravitystorm:
Thanks both!

PR #5799, opened 2025-03-14
Disable asynchronous object layer loading

### LINKED ISSUE #5798: opening https://www.openstreetmap.org/way/12432594/ link does not center map where way is
### URL

https://www.openstreetmap.org/way/12432594/

1) open https://www.openstreetmap.org/way/12432594/
2) note that map is not moving to way location

it used to work very recently

confirmed on IRC it is not just me

AntonKhorev:
Fixes #5798.

I see #5009 and its commit mentioning only "Map Data checkbox". The object layer is not related to that checkbox and the code that uses the object layer doesn't expect async loading.


tomhughes:
Looks good to me, thanks.

PR #5800, opened 2025-03-14
Align bottoms of date-username-changeset_id and counters in history

AntonKhorev:
Moves "Closed (date) by (user)" lines down slightly if necessary to align with edit counters.

Before / after for different users:
image image

Before / after for a single user:
image image


tomhughes:
Looks good to me, thanks.

PR #5801, opened 2025-03-14
API changeset subscription resource

AntonKhorev:
Similar to web changeset subscriptions https://github.com/openstreetmap/openstreetmap-website/pull/5535, but for the api.

Adds `/api/0.6/changeset/:changeset_id/subscription` path at which you can POST/DELETE to subscribe/unsubscribe.


tomhughes:
Looks good to me, thanks.

PR #5803, opened 2025-03-14
Fix wget command for Windows installation guide

linhle15-ll:
### Description
In Window Shell, I directed to `cd openstreetmap-website` > `ls` but cannot see data file for District of Columbia although the download StatusCode : 200 and StatusDescription : OK.

Added a note in DOCKER.md for Windows users regarding potential issues when downloading OSM extracts using wget. Some Windows configurations do not properly save the file unless -OutFile is explicitly specified.

### How has this been tested?
Tested on Windows 11 PowerShell using wget with and without -OutFile.
Verified that using -OutFile ensures the file is properly saved and can be found in the directory.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -116,7 +116,11 @@ For example, let's download the District of Columbia from Geofabrik or [any othe
 ```
 wget https://download.geofabrik.de/north-america/us/district-of-columbia-latest.osm.pbf
 ```
+**Windows users:** On some Windows configurations, the downloaded file may not be saved properly. If you encounter issues, try using an explicit output filename with `-OutFile` in the wget command:
 
+```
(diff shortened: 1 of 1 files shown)


linhle15-ll:
Hope this help Window users (I use Window 11).

tomhughes:
I'm not sure we can turn the documentation into a tutorial for the quirks of each individual operating system... If wget isn't downloading to the current directory by default then where is is downloading to? Maybe that's a bug that should be raised with wget rather than trying to work around it here.

If we're going to make a change here then I suggest just adding a `-o` switch (you don't need to to spell it out fully) to the wget command rather than adding extra instructions for Windows.

linhle15-ll:
Thank you for getting back to me!

By the way, I am considering applying for Open Street Map in the Google
Summer of Code. Do you have any suggestions on things I should do or write
about in my proposal? I am very excited to make contributions to OSM via
this program (and out of it).

Best regards,
Linh Ngoc Le

PR #5806, opened 2025-03-15
Fix heatmap month titles in Western Hemisphere

### LINKED ISSUE #5802: Contribution heatmap is off by a month
The horizontal axis on the user contribution heatmap is off by a month. Each label across the top appears to be centered over the following month. For example, if you’ll excuse me showing off my own profile:

The tooltip says March 04, 2025, but the label above it says February.

Not sure if this is a bug in cal-heatmap upstream or in this repository, since [the template](https://github.com/openstreetmap/openstreetmap-website/blob/037be97bf1fd758dfacd7861276f13a629261a59/app/views/users/show.html.erb#L245-L253) only contains an explicit list for the day labels along the side.

AntonKhorev:
Fixes #5802

Before when simulating some place in the US (Chrome dev tools > Sensors > Location):
image

After:
image

### DIFF
--- a/app/assets/javascripts/heatmap.js
+++ b/app/assets/javascripts/heatmap.js
@@ -33,7 +33,7 @@ document.addEventListener("DOMContentLoaded", () => {
         type: "month",
         gutter: 4,
         label: {
-          text: (timestamp) => monthNames[new Date(timestamp).getMonth() + 1],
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #5811, opened 2025-03-16
Update to i18n-js 4.x

tomhughes:
This updates i18n-js to the 4.x branch which requires quite a bit of re-engineering due to upstream API changes.

Although this looks quite large the vast majority of it is just replacement of `I18n.t` with `OSM.i18n.t` as you now need to instantiate an instance of I18n instead of just using global functions.


hlfan:
How about instanciating i18n or even just aliasing `OSM.i18n = { t: I18n.t }` in a separate commit first so the main version upgrade can be a smaller commit?

tomhughes:
I've split this as suggested so that the mass renaming can be done first and then the functional changes.

app/assets/javascripts/application.js
```diff
@@ -19,13 +19,13 @@
...
-  I18n.default_locale = OSM.DEFAULT_LOCALE;
-  I18n.locale = locale;
+  OSM.i18n.defaultLocale = OSM.DEFAULT_LOCALE;
+  OSM.i18n.locale = application_data.locale;
 
   // '-' are replaced with '_' in https://github.com/eemeli/make-plural/tree/main/packages/plurals
   const pluralizer = plurals[locale.replace(/\W+/g, "_")] || plurals[locale.split("-")[0]];
   if (pluralizer) {
-    I18n.pluralization[locale] = (count) => [pluralizer(count), "other"];
+    OSM.i18n.pluralization.register(locale, I18n.useMakePlural({ pluralizer: locale }));
```
AntonKhorev, line 28: If you want to write this according to i18n-js docs, it should be:

```js
    OSM.i18n.pluralization.register(locale, I18n.useMakePlural({ pluralizer }));
```

However that's not the best you can do.

English has `one` / `other` forms. Some languages have `one` / `few` / `many` forms. Let's suppose that there's a string with `count` that's not yet translated into such language. First, `count` goes to the pluralizer and the pluralizer says *it has to be `many` form*. The string with `many` key is looked up and it's missing. Then a fallback to English happens, the string is looked up and it's missing again because English has no `many` form. The user gets to see an error message instead of any fallback:

image

That's why I wrote the pluralizer wrapper like this for i18n-js v3:

```js
    I18n.pluralization[locale] = (count) => [pluralizer(count), "other"];
```

It has a built-in fallback to `other` which is going to be present in English. It's not entirely correct because rules for `one` form can also be different and you sometimes get fallbacks to English `one` instead of `other`, but usually it's bearable.

The equivalent for i18n-js v4 looks like this:

```js
    OSM.i18n.pluralization.register(locale, (_, count) => [pluralizer(count), "other"]);
```

tomhughes: The actual example upstream is:

```js
register("ru", useMakePlural({ pluralizer: ru })
```

which I misread as I didn't notice the difference between the constant string `ru` and the variable `ru` so I'll fix that now.

I had been struggling to understand your fallback simply because upstream doesn't seem to document the return value of the pluralizer so while I thought it was some sort of fallback but I've found the code in the source now and understand it so I'll fix that as well.

tomhughes: All updated.

PR #5812, opened 2025-03-16
Add 'Load more' button that loads newer changesets to history pages

AntonKhorev:
If you get to the user's history page from the heatmap, it will start on a selected date. You can scroll down, press *Load more* and and load changesets that were created before the initially visible set. But what if you want to see changesets created after the date?

This PR adds a *Load more* button in frond of the changeset list, if there are any changesets that were opened after the ones in the list.

image image

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -47,12 +47,24 @@ OSM.History = function (map) {
     $("#sidebar_content .changesets").html(html);
   }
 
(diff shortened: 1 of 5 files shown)


DavidKarlas:
Should existing button be renamed to "Load older" and new button to "Load newer"? Just idea...

tomhughes:
Looks good to me, thanks, and we can always change the labels later if we want.

PR #5805, opened 2025-03-15
Add links to days in user history from heatmap

AntonKhorev:
Because navigating through user history is difficult, I wanted this: https://github.com/openstreetmap/openstreetmap-website/issues/5804#issuecomment-2726693156

This PR makes it that clicking on a nonempty calendar cell takes you to the corresponding day in the user's changeset history.

image

PR #5814, opened 2025-03-17
Create heatmap links using <a> elements

AntonKhorev:
#5805 added links to heatmaps, but those weren't real links. Those were click listeners that rewrite page locations. The browser won't show their urls and they can't be opened in new tabs.

This PR creates real links by wrapping svg rectangles in `<a>` elements.

image


tomhughes:
Looks good to me, thanks.

PR #5816, opened 2025-03-17
Add intellisense for calHeatmap

HolgerJeromin:
Add intellisense for calHeatmap

The type annotations of the cal-heatmap file is not very good, but still useful:

image


AntonKhorev:
I can see these without any changes:

image
image

PR #5817, opened 2025-03-17
Move directions form buttons closer to edges

AntonKhorev:
Before / after:
image image

- Adds more space between Close and Reverse buttons. See "The close button is a bit too close the reverse directions button" in https://github.com/openstreetmap/openstreetmap-website/issues/3123#issuecomment-2726356151

- Places Close button exactly above Sidebar Close button. 
image


tomhughes:
Looks good to me, thanks.

PR #5668, opened 2025-02-13
rerun script/misc/update-wiki-pages

matkoniecz:
### Description
standard rerun of wiki pages script, like most of my PRs in this repo

### How has this been tested?
not at all, I just looked at diff for suspicious parts

PR #5818, opened 2025-03-18
update-wiki-pages: process ChangesetKey template

k-yle:
### Description
This perl script was last run in #5668, which caused the links for some keys like [`host=*`](https://osm.wiki/Key:host) to disappear, because these pages were migrated from [`Template:KeyDescription`](https://osm.wiki/Template:KeyDescription) to [`Template:ChangesetKeyDescription`](https://osm.wiki/Template:ChangesetKeyDescription)[^1][^2].

This PR updates the script to support [`Template:ChangesetKeyDescription`](https://osm.wiki/Template:ChangesetKeyDescription), which is used by pages under the `Key:` namespace, and/or the `ChangesetKey:` namespace.

### How has this been tested?
Ran the perl script and commited the relevant changes.

[^1]: [`Key:changesets_count` diff#2685844](https://osm.wiki/w/index.php?title=Key:changesets_count&diff=2685844&oldid=2685832)
[^2]: [`Key:host` diff#2612462](https://osm.wiki/w/index.php?title=Key:host&diff=2612462&oldid=2612460)

### DIFF
--- a/script/misc/update-wiki-pages
+++ b/script/misc/update-wiki-pages
@@ -71,9 +71,15 @@ for my $lang ('', map { "${_}:" } qw[ Pt Fi De It HU Cz Fr RU Pl ]) {
     $cnt = stick_content_in_hash("tag", "Template:${lang}ValueDescription", \%feature);
     ok(1, "    Got $cnt value pages");
     $count{value} += $cnt;
+
+    # ChangesetKey pages
+    ok(1, "    Getting changeset key pages");
+    $cnt = stick_content_in_hash("key", "Template:${lang}ChangesetKeyDescription", \%feature, "ChangesetKey");
+    ok(1, "    Got $cnt changeset key pages");
+    $count{changesetkey} += $cnt;
 }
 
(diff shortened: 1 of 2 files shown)


script/misc/update-wiki-pages
```diff
@@ -8,6 +8,7 @@ use Pod::Usage;
 use MediaWiki::API;
 use Test::More qw(no_plan);
 use YAML::XS qw(Dump);
+# NOTE: on macos, LWP::Protocol::https also needs to be installed
```
tomhughes, line 11: There's nothing macos specific about that - the https protocol handler is in a separate CPAN module and is in a separate package for most if not all linux distributions as well.

But because the protocol handlers are loaded dynamically they don't need to be explicitly required here and won't be a dependency of LWP and hence of anything that requires LWP.

k-yle: thanks for the explaination, I've removed this comment

tomhughes (approved):

PR #5820, opened 2025-03-19
Switch to dynamic allocation for the Firefox bidi port

tomhughes:
As pointed out in https://github.com/mozilla/geckodriver/issues/2218#issuecomment-2735625538 you can use port zero to ask Firefox to dynamically allocate a port which should make this more robust as well as much simpler.

### DIFF
--- a/config/initializers/selenium.rb
+++ b/config/initializers/selenium.rb
@@ -4,28 +4,10 @@ if Rails.env.test?
   module OpenStreetMap
     module Selenium
       module BidiPort
-        module ClassMethods
-          attr_accessor :websocket_port
-        end
-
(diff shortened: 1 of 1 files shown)


gravitystorm:
Merged, thanks

Openstreetmapler:
I think this pull request completely broke route planning in Firefox, as there’s now a missing `currentLocale` function:
https://github.com/openstreetmap/openstreetmap-website/issues/5821

tomhughes:
It also hasn't been deployed yet!

Openstreetmapler:
My issue is now fixed … it was caused by the recent i18n update: #5811

PR #5823, opened 2025-03-19
Add reload on iframe redirect

### LINKED ISSUE #5807: Duplicate header displayed after resuming browser session
### URL

https://www.openstreetmap.org/edit?editor=id&relation=3121668#map=8/50.469/5.884

Minor visual bug.
If my PC has been in sleep mode for long enough it has to revive the active applications.
The browser is opened and all pages are resumed (reloaded).
Upon going to the openstreetmap.org/edit tab I am logged out (working as intended)
However, the header is duplicated - one displaying the logged in state, one below showing the logged out state.
(the first/second header is separate content, the first one does not have the vertical scroll bar on the right)

If I now open the edit link from knooppuntnet which I used in a different tab, or any /edit link, it will work fine.
If I i.e. go to my profile in the existing tab instead, and use the back browser button, (though to be expected) it will still be shown incorrectly.

Chrome 134.0.6998.89 on Windows 10.

Image

hlfan:
In some cases, an expired session only causes the window in the iframe to reload on the `/edit` page.
This fixes #5807 (the issue for that bug)

### DIFF
--- a/app/assets/javascripts/edit/id.js.erb
+++ b/app/assets/javascripts/edit/id.js.erb
@@ -36,4 +36,10 @@ $(function () {
   if (idData.gpx) params.set("gpx", idData.gpx);
 
   id.attr("src", idData.url + "#" + params);
+
+  id.ready(function () {
+    if (!this.contentWindow.location.href.startsWith(idData.url)) {
+      location.reload();
+    }
+  });
 });


tomhughes:
It's a bit nasty but I don't see a better solution - the ideal would be that the server forces a full page reload when it needs to redirect the id frame to the login but that isn't possible for obvious reasons because of the security boundary between the main page and the frame.

hlfan:
I also considered replacing the already-loaded content, but that would probably have become more nasty.

app/assets/javascripts/edit/id.js.erb
```diff
@@ -36,4 +36,10 @@ $(function () {
   if (idData.gpx) params.set("gpx", idData.gpx);
 
   id.attr("src", idData.url + "#" + params);
+
+  id.ready(function () {
+    if (!this.contentWindow.location.href.startsWith(idData.url)) {
```
tyrasd, line 41: I get a `Uncaught TypeError: this.contentWindow is undefined` error every time I regularly start an iD editing session. Is that expected? Would it not be better to skip the check if everything is loaded correctly, maybe like this:

```js
if (this.contentWindow && !this.contentWindow.location.href.startsWith(idData.url)) {
```

AntonKhorev: I'm not sure if this fix ever worked. I've seen double header from #5807 with this PR merged.

PR #5824, opened 2025-03-19
Make URLs more relative

hlfan:
Removes hard-coding of URLs, e.g. not forcing a protocol onto the user.


tomhughes:
Looks good to me, thanks.

AntonKhorev:
Why do we have to replace substrings in urls?

tomhughes:
I'm not really sure why it's doing that... There an aspiration to separate API calls onto a separate domain which I guess is the reason but we've never got anywhere near making it a reality.

PR #5825, opened 2025-03-19
Remove OSM.params in favor of URLSearchParams

hlfan:
Removes `OSM.params()` in favor of `new URLSearchParams(location.search)`.
Adds `location.hash` as a default parameter for `OSM.parseHash`.


AntonKhorev:
How about we remove `OSM.params` instead?

tomhughes:
Looks good to me, thanks.

Issue #568, opened 2013-11-30
No longer any way to view a user's changeset feed a page at a time


SomeoneElseOSM:
The previous paged mode was extremely useful - it allowed you to see if a particular user had been editing in a particular area a page at a time.

For example, someone could view http://www.openstreetmap.org/user/SomeoneElse/ a page at a time and tell from that that between dates X and Y, the only edits were in a particular geographic area.  

It's no longer possible ot do this.  Instead of being able to see the next page there's a "load more" button at the bottom of the screen that shows all the changesets from the current date, and the map display at the right contains boxes for all of the changesets that you've already looked at.

Do that 3 or 4 times for me and even if you zoom in the next time that you press "load more" the map will attempt to show boxes in the UK and 2 in Indonesia, making it impossible to see where in the UK the newly-displayed edits were made.

If I want to see "where user X edited in March last year" I can't.

PR #5826, opened 2025-03-19
Rewrite history page url when scrolling

### LINKED ISSUE #647: Viewing a changeset loses context of previous changeset list

### LINKED ISSUE #933: "Load more" on user's changesets doesn't alter the url (ie doesn't keep state)

AntonKhorev:
The goal here is when viewing history to be able to go back to roughly the same place after visiting a changeset page. It works like this:
- when the changeset list is scrolled down and some item leaves the viewport, the current page location gets changed to include a `before=:changeset_id` parameter, `:changeset_id` being the id of the changeset that left the viewport
- when the changeset list is scrolled up, same thing happens with `after=:changeset_id`
- when the history page is visited and its url includes a `before` parameter, the sidebar is scrolled so the first item is at the top of the viewport
- when the history page is visited and its url includes an `after` parameter, the sidebar is scrolled so the last item is at the bottom of the viewport

With these changes you can open a history page, scroll and load more if necessary, click on a changeset, then click back and arrive at a view that includes roughly the same changesets.

https://github.com/user-attachments/assets/5b745d80-748e-4c4a-a724-c874e15943f5

Fixes #647. Also fixes the main issue from #1089 which is the same as #647.

Kind of fixes #933. Clicking *Load more* won't rewrite the url, but scrolling will. *Load more* is actually a link but it doesn't do the expected thing when used as a link, that's to be fixed later.

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -4,7 +4,7 @@ OSM.History = function (map) {
   const page = {};
 
   $("#sidebar_content")
-    .on("click", ".changeset_more a", loadMore)
+    .on("click", ".changeset_more a", loadMoreChangesets)
     .on("mouseover", "[data-changeset]", function () {
       highlightChangeset($(this).data("changeset").id);
     })
@@ -132,7 +132,7 @@ OSM.History = function (map) {
     }
   }
 
-  function update() {
+  function loadFirstChangesets() {
     const data = new URLSearchParams();
     const params = new URLSearchParams(location.search);
 
@@ -173,7 +173,7 @@ OSM.History = function (map) {
       });
   }
 
-  function loadMore(e) {
+  function loadMoreChangesets(e) {
     e.preventDefault();
     e.stopPropagation();
 
@@ -250,17 +250,17 @@ OSM.History = function (map) {
     map.addLayer(group);
 
     if (location.pathname === "/history") {
-      map.on("moveend", update);
+      map.on("moveend", loadFirstChangesets);
     }
 
     map.on("zoomend", updateBounds);
 
-    update();
+    loadFirstChangesets();
   };
 
   page.unload = function () {
     map.removeLayer(group);
-    map.off("moveend", update);
+    map.off("moveend", loadFirstChangesets);
     map.off("zoomend", updateBounds);
     disableChangesetIntersectionObserver();
   };
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -27,6 +27,65 @@ OSM.History = function (map) {
     return layer.id;
   };
 
+  let changesetIntersectionObserver;
+
+  function disableChangesetIntersectionObserver() {
+    if (changesetIntersectionObserver) {
+      changesetIntersectionObserver.disconnect();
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/index/history.js
```diff
@@ -27,6 +27,65 @@ OSM.History = function (map) {
...
+        if (id) {
+          OSM.router.replace(location.pathname + "?" + new URLSearchParams({ before: id }) + location.hash);
+        }
+      } else if (closestTargetToBottom) {
+        const id = $(closestTargetToBottom).data("changeset")?.id;
+        if (id) {
+          OSM.router.replace(location.pathname + "?" + new URLSearchParams({ after: id }) + location.hash);
+        }
+      }
+    }, { root: $("#sidebar")[0] });
```
hlfan, line 82: Does the callback need to be redefined every time?

AntonKhorev: Every time as opposed to what? Are you proposing to reuse one callback when creating new observers?

hlfan: I only meant moving the definition out of enableChangesetIntersectionObserver
```js
changesetIntersectionObserver = new IntersectionObserver(observerCallback, { root: $("#sidebar")[0] });
```

AntonKhorev: That's not enough because you'd also have to remove `ignoreIntersectionEvents`.

hlfan: You can also do that in the callback?

hlfan:
This might also close #568 too.

app/assets/javascripts/index/history.js
AntonKhorev: I can't do that in the callback, it's a state variable that is outside the callback.

tomhughes (changes requested):
Looks good to me - just one small cosmetic comment.

app/assets/javascripts/index/history.js
```diff
@@ -97,11 +158,22 @@ OSM.History = function (map) {
...
       .then(function (html) {
         displayFirstChangesets(html);
+        enableChangesetIntersectionObserver();
+
+        if (params.has("before")) {
+          const firstItem = $("#sidebar_content .changesets ol").children().first()[0];
+          firstItem?.scrollIntoView();
+        }
+        if (params.has("after")) {
+          const firstItem = $("#sidebar_content .changesets ol").children().last()[0];
```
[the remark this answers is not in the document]
AntonKhorev: renamed

tomhughes:

PR #5827, opened 2025-03-20
Update straight routing in tests to use named directions icons

AntonKhorev:
Routing icons were updated to be referenced by names instead of numbers, but there was test code that still used numbers. This didn't affect the functioning of the tests because the icons weren't checked.

This PR updates the *straight routing stub* to use named icons:

image


tomhughes:
Looks good to me, thanks.

PR #5829, opened 2025-03-20
Show heatmap only to logged in users

AntonKhorev:
I'd like to add a year selector to the heatmap and maybe more things. Since that will reveal even more information about users we might want to hide the heatmap for those who didn't log in. A proper check would be the terms of use acceptance, but that's not possible until #5706 is merged.

### DIFF
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -399,19 +399,23 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     assert_empty assigns(:heatmap_data)
   end
 
-  def test_heatmap_rendering
-    # Test user with no changesets
-    user_without_changesets = create(:user)
-    get user_path(user_without_changesets)
+  def test_show_heatmap_rendering_of_user_with_no_changesets
+    user = create(:user)
+
+    get user_path(user)
+
     assert_response :success
     assert_select "div#cal-heatmap", 0
+  end
 
-    # Test user with changesets
-    user_with_changesets = create(:user)
-    changeset39 = create(:changeset, :user => user_with_changesets, :created_at => 4.months.ago.beginning_of_day, :num_changes => 39)
-    _changeset5 = create(:changeset, :user => user_with_changesets, :created_at => 3.months.ago.beginning_of_day, :num_changes => 5)
-    changeset11 = create(:changeset, :user => user_with_changesets, :created_at => 3.months.ago.beginning_of_day, :num_changes => 11)
-    get user_path(user_with_changesets)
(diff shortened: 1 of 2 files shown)


Firefishy:
In addition to this, some users may still prefer to hide their heatmap.

Here is what the feature looks like on github:

Screenshot 2025-03-20 at 23 34 17

AntonKhorev:
We don't have:
- private profiles
- private contributions

Firefishy:
> We don't have:
> 
> * private profiles
> * private contributions

The "Make [...] and **hide activity**" is the relevant portion.

AntonKhorev:
- [ ] hide activity and *make my edits more difficult to review*, edits that are not in some private repository and weren't reviewed before being merged with the rest of the data

tomhughes:
Well we can allow moderators to see the heatmap even where the user has asked to hide it, and I was assuming that we would, which helps with the review problem.

kcne:
I've already started working on adding opt-out settings for users and will be making a PR in the next few days.

AntonKhorev:
If users can remove the heatmap, it can't be used for navigation.

> Well we can allow moderators to see the heatmap even where the user has asked to hide it

This implies that it's moderators' job to review all changesets.

tomhughes:
Look I don't want to remove it but it legally we have no choice but to offer that option then so be it - it would be good to get an actual statement from the current LWG rather than comments from a former member of course.

My comment about moderators was simply that I thought part of what you were trying to do was make it easier for DWG to use the heatmap to explore a user's edits during investigations and I was making the point that we could preserve that ability.

test/controllers/users_controller_test.rb
```diff
@@ -330,20 +330,31 @@ def test_auth_failure_callback
...
     get user_path(user)
+
     assert_response :success
     # The data should not be empty
     heatmap_data = assigns(:heatmap_data)
     assert_not_nil heatmap_data
     assert_predicate heatmap_data[:data], :any?
     # The data should be in the right format
-    assert(heatmap_data[:data].values.all? { |entry| entry[:date] && entry[:total_changes] }, "Heatmap data should have :date and :total_changes keys")
+    assert(heatmap_data[:data].values.all? { |entry| entry[:date] && entry[:total_changes] && entry[:max_id] }, "Heatmap data should have :date and :total_changes keys")
```
hlfan, line 357: Do wee need to keep that comment or is "incorrect heatmap data structure" enough?

AntonKhorev: Rewrote it without naming the kys in the message.

hlfan (approved):

PR #5830, opened 2025-03-20
Map key resource

AntonKhorev:
The remainder of #5024 done similarly to #5751, without `Features` namespace.

Moves map key from `site` to its own resource/controller.


tomhughes:
Looks good to me, thanks.

PR #5833, opened 2025-03-21
Change heatmap color to sqrt from 0 to max changes

### LINKED ISSUE #5831: Increase dynamic range of contribution heatmap
The new display is really neat, but for heavier users like me it doesn't really represent [my usage](https://www.openstreetmap.org/user/JuanElDemografo) well. The color for a day with ~30 contributions isn't significantly different from one with over 5000 contributions.

Image

HolgerJeromin:
### Description

Changed heatmap color range interpolation.

Image

That changes the heatmap to be not comparable from one user to another, but better shows the day to day changes of one user.

### How has this been tested?
Runtime changes in devTools

fix #5831

### DIFF
--- a/app/assets/javascripts/heatmap.js
+++ b/app/assets/javascripts/heatmap.js
@@ -11,10 +11,12 @@ document.addEventListener("DOMContentLoaded", () => {
     return;
   }
 
+  /** @type {{date: string; max_id: number; total_changes: number}[]} */
   const heatmapData = heatmapElement.dataset.heatmap ? JSON.parse(heatmapElement.dataset.heatmap) : [];
   const displayName = heatmapElement.dataset.displayName;
   const colorScheme = document.documentElement.getAttribute("data-bs-theme") ?? "auto";
-  const rangeColors = ["#14432a", "#166b34", "#37a446", "#4dd05a"];
+  const rangeColorsDark = ["#14432a", "#4dd05a"];
+  const rangeColorsLight = ["#4dd05a", "#14432a"];
   const startDate = new Date(Date.now() - (365 * 24 * 60 * 60 * 1000));
   const monthNames = OSM.i18n.t("date.abbr_month_names");
 
@@ -59,9 +61,9 @@ document.addEventListener("DOMContentLoaded", () => {
       },
       scale: {
         color: {
-          type: "threshold",
-          range: currentTheme === "dark" ? rangeColors : Array.from(rangeColors).reverse(),
-          domain: [10, 20, 30, 40]
+          type: "sqrt",
+          range: currentTheme === "dark" ? rangeColorsDark : rangeColorsLight,
+          domain: [0, Math.max(0, ...heatmapData.map(d => d.total_changes))]
         }
       }
     }, [


app/assets/javascripts/heatmap.js
```diff
@@ -59,9 +60,12 @@ document.addEventListener("DOMContentLoaded", () => {
...
       scale: {
         color: {
-          type: "threshold",
+          type: "linear",
+          // reverse mutates the array. But with theme changes we must work on the same data.
           range: currentTheme === "dark" ? rangeColors : Array.from(rangeColors).reverse(),
-          domain: [10, 20, 30, 40]
+          domain: [0, heatmapData.reduce((max, item) => {
+            return item.total_changes > max ? item.total_changes : max;
+          }, 0)]
```
hlfan, line 68: I think the 20 % performance loss is worth going for the more concise option:
```suggestion
          domain: [0, Math.max(...heatmapData.map(d => d.total_changes)), 0)]
```

app/assets/javascripts/heatmap.js
```diff
@@ -59,9 +60,10 @@ document.addEventListener("DOMContentLoaded", () => {
       },
       scale: {
         color: {
-          type: "threshold",
-          range: currentTheme === "dark" ? rangeColors : Array.from(rangeColors).reverse(),
-          domain: [10, 20, 30, 40]
+          type: "linear",
+          // reverse mutates the array. But with theme changes we must work on the original data.
+          range: currentTheme === "dark" ? rangeColors : Array.from(rangeColors).toReversed(),
```
hlfan, line 65: Array.toReversed() has only been available in browsers since mid-2023 and wouldn't be needed in combination with Array.from().
I think the previous version was fine enough, `[...rangeColors].reverse()` would also work.

app/assets/javascripts/heatmap.js
HolgerJeromin: I was not was not awake enough to fiddle myself something and trusted copilot to find something smart. :-(
Thanks for your input!
Changed to something similar like your suggestion.

hlfan: Well, it's faster, but we're not exactly doing massive-scale data science here.

app/assets/javascripts/heatmap.js
HolgerJeromin: I experimented with `toReversed` (from es2023) and failed to revert as I found that eslint had es2021 as min version.

app/assets/javascripts/heatmap.js
```diff
@@ -59,9 +60,10 @@ document.addEventListener("DOMContentLoaded", () => {
       },
       scale: {
         color: {
-          type: "threshold",
+          type: "linear",
```
hlfan, line 63: It would be nice to have changes in fewer contributions me more visible than the same change in more contributions.
While I don't think the type `log` (or `symlog` for better zero mapping) is necessary, `sqrt` would still do that a bit.

HolgerJeromin: Changed to `sqrt` as this got good votes in referenced item.

app/assets/javascripts/heatmap.js
```diff
@@ -59,9 +60,10 @@ document.addEventListener("DOMContentLoaded", () => {
       },
       scale: {
         color: {
-          type: "threshold",
+          type: "sqrt",
+          // reverse mutates the array. But with theme changes we must work on the original data.
```
AntonKhorev, line 64: What does this comment say?

If we need to stop contributors from trying `rangeColors.reverse()` [again](https://github.com/openstreetmap/openstreetmap-website/pull/5402#discussion_r1923048595) and again, we can have two different `rangeColors` arrays defined earlier. Maybe the colors in them also need to be different, not just reversed.

app/assets/javascripts/heatmap.js
```diff
@@ -11,10 +11,11 @@ document.addEventListener("DOMContentLoaded", () => {
     return;
   }
 
+  /** @type {{date: string; max_id:number; total_changes: number}[]} */
```
AntonKhorev, line 14: If you want spaces after `:`, the one after `max_id:` is missing.

AntonKhorev:
Merged, thanks.

PR #5834, opened 2025-03-21
Use Date().toLocaleString() to construct month titles

AntonKhorev:
In some languages month names when used inside dates may differ from month names when used alone as titles. `OSM.i18n.t("date.abbr_month_names")` gives you month names for dates. That doesn't work with calendar month titles.

Example in Russian:
"May" = "май"
"21 May, 2024" = "21 мая 2024"
You don't want "мая" as a title, you want "май".

Before:
image

After:
image


HolgerJeromin (approved):

tomhughes:
It's a bit annoying having to use mix our own translation with the browser's like this bit it seems to work OK so this looks good to me.

Issue #3432, opened 2022-01-20
Avoid using text-align on body element


gravitystorm:
As part of #3427 and following up from the last part of #3428 we have a problem where all text, even with dir=auto, is forced into a pre-determined alignment. This is because we set the global text-alignment explicitly in our CSS:

https://github.com/openstreetmap/openstreetmap-website/blob/3a534ea895fd60d2c5cccc66251e5472868c0f4e/app/assets/stylesheets/common.scss#L29-L31

This property then cascades down to every element. Setting dir="auto" on a child element is not enough to override this global text-align. But if we omit the text alignment altogether, everything works fine by default. Having dir=rtl on the body makes everything right-aligned by default, and where we specify dir="auto", then the alignment will automatically flip based on content (since there is no overriding text-align set).

Unfortunately bootstrap 4 has [`text-align: left` hardcoded](https://github.com/twbs/bootstrap/blob/349a373ff62bf530135ad95d7d1d3f1be6abbf22/scss/_reboot.scss#L53) and expects RTL-only sites to override this. But in bootstrap 5, which has additional focus on RTL, that was [removed](https://github.com/twbs/bootstrap/pull/28989), and the [default `text-align` is now null](https://github.com/twbs/bootstrap/blob/520cc8de92ca48d7ac60102d992d7afe9788b239/scss/_variables.scss#L401) allowing dir="auto" to work as intended.

So fixing this is split into three parts:
* Upgrade to bootstrap 5 (#3230).
* Remove the global `text-align` from our own CSS.
* Check the other `text-align` declarations in our CSS, and see which can be removed.

PR #5835, opened 2025-03-21
add dir="auto" to paragraphs in rich_text.rb

NeatNit:
(the edit history of this comment contains some questions that are no longer relevant)

### Description
This PR adds `dir="auto"` to user-generated content that uses the classes in `lib/rich_text.rb`, such as changeset comments, map notes, and the discussions under them. It also applies a necessary CSS rule for text alignment on those discussions.

This would solve a significant portion of #3428. It *might* have some overlap with #3429, ~~but at a glance it doesn't seem like it~~. It seems some of the changes in that PR would have affected the same content, however they would have done so with less finesse.

I never used Ruby before (and technically, I still haven't... see "How has this been tested?"), so this change is entirely based on the documentation found [here](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format). I hope I've understood it correctly. Further changes were made as a response to the discussion below.

### How has this been tested?
**This has not been tested at all. For all I know, it might even contain syntax errors.** I don't currently have the means to clone and run this repository. I know this is bad form to put it mildly, but with a change this small and focused I am hoping you can forgive and review it nonetheless, and run any tests in your already-set-up environment.

If this is totally unacceptable, let me know.

### Things a reviewer should check or know the answer to
- Which parts of the website would be affected by this change?
- Does it actually work? (As said, I can't test it)
- Does this exclusively affect user-generated content, or is it sometimes applied to other kinds of content? If so, does it ever cause undesirable results? (I believe it wouldn't)
- Are there any other places where text alignment doesn't change based on text direction? (This was only an issue on discussion comments because they are contained in `<li>` elements, see discussion below)

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -68,7 +68,7 @@ module RichText
     protected
 
     def simple_format(text)
-      SimpleFormat.new.simple_format(text)
+      SimpleFormat.new.simple_format(text, :dir => "auto")
     end
 
     def sanitize(text)


NeatNit:
Trying to read the code more thoroughly (keep in mind it's still a bit alien to me), I'm fairly sure I'm adding that parameter in the wrong place... But uhh, it's the thought that counts, right? 😅

I really don't care if my PR gets merged or not, so if you want to just take the idea and implement it correctly in a separate PR, that's fine by me. It will probably be the most efficient way for everyone involved. Otherwise, I might *eventually* have the means to do this properly, but my timeline for that is really not good (optimistically I *might* be able to start trying it over a month from now).

lib/rich_text.rb
```diff
@@ -86,7 +86,7 @@ def linkify(text, mode = :urls)
 
   class HTML < Base
     def to_html
-      linkify(sanitize(simple_format(self)))
+      linkify(sanitize(simple_format(self, dir: "auto")))
```
AntonKhorev, line 89: [This helper](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format) is not called directly. Look at `def simple_format(text)` above in this file, it doesn't have extra arguments.

NeatNit:
Related to #3432, it seems that in the case of changeset comments, they are arranged in a list, and `text-align` is being set by the browser stylesheet (in Firefox at least) to `<li>` elements:

```css
li {
  display: list-item;
  text-align: match-parent;
}
```

This makes sense for lists in their usual setting but not so much in this case. So one possible solution is:

```css
li p {
	text-align: start;
}
```

So that paragraphs won't inherit the alignment from their parent too. But again, I'm not a web developer, so more professional discretion might be needed.

lib/rich_text.rb
NeatNit: Yup, that's what I thought. How about this? (commit I just pushed)

AntonKhorev:
If you want to modify css, it makes sense to add a class on the entire comment and define properties in `common.scss`, similar to

https://github.com/openstreetmap/openstreetmap-website/blob/6af214b2285d3292f214320b7221f7c9a1d76914/app/assets/stylesheets/common.scss#L874

NeatNit:
> If you do this, you'd also set `dir=auto` on the entire comment.

This would not be useful. Each paragraph needs to have `dir="auto"` separately. Different paragraphs in the same comment can have a different direction. This is why it needs to be an argument to `simple_format`.

Thanks for the tip though, I will look into how to do that... Might not have to be in the same PR though. I don't want this PR to grow to the size of #3429 handling each case separately. Perhaps your suggestion should be added there instead. As I said, even without `text-align`, just having `dir="auto"` already offers the most important benefit by correcting the placement of punctuation and mixed directionality text.

AntonKhorev:
> This would not be useful. Each paragraph needs to have dir="auto" separately.

If you want a single comment with different language paragraphs, then looks like yes, you need `dir` on paragraphs.

> Thanks for the tip though, I will look into how to do that...

Something like this:
```erb
          <div class="mx-2 comment_text">
            <%= comment.body.to_html %>
          </div>
```

```css
.comment_text {
  p {
    text-align: start;
  }
}
```

NeatNit:
> Something like this:

How about almost exactly that :)

I applied it to note comments as well. I'm not sure if any other places need the same treatment.

AntonKhorev:
`text-align: start;` is the default value, unless you're in a `<li>`. Sidebar comments are in html lists, but should they? We don't output diary comments as a list.

If you view comments on [this page](https://www.openstreetmap.org/user/NeatNit/changeset_comments), you'll notice that adding `dir=auto` on paragraphs is enough for them to align properly.

I guess browsers change the alignment inside lists to keep the text next to the bullets, and even if we hide bullets, that alignment sticks.

NeatNit:
> Sidebar comments are in html lists, but should they?

This is above my pay grade 😉 and definitely outside the scope of this PR

NeatNit:
For kramdown, this approach should work: https://codeberg.org/NeatNit/kramdown-bidi/src/branch/main/code.rb (the code works for sure but I haven't tested the results in a browser)

I've stayed up too late again... I'll add it to this PR tomorrow :)

Thanks to the Ruby Discord for helping me with syntax.

AntonKhorev:
With #5837 merged it should be enough to just set `dir=auto` on paragraphs:

image

You can remove css changes and squash the commits.

NeatNit:
Awesome, I'll do that ASAP :)

tomhughes:
Can you also squash the fixup commits into the commit they're fixing please.

tomhughes:
In fact this is going to just wind up as a single commit I think so squash everything together unless you have a good reason not to.

NeatNit:
> In fact this is going to just wind up as a single commit I think so squash everything together unless you have a good reason not to.

That's what I intend to do. Just need to rebase with the new master. Will get behind a computer in a few minutes to do it. I contemplated doing it from termux, but that sounds too painful.

NeatNit:
If you have the option to squash and merge through GitHub, that would mean I made a successful PR without ever cloning the repository to my machine. Can you?

tomhughes:
I've done a rebase and will merge this now - thanks for all the work.

NeatNit:
Yay! How long do you think until it goes live on osm.org?

NeatNit:
Looks like I missed something important (main changeset comment). I'll lump that in with my next PR, which handles Markdown.

Issue #5785, opened 2025-03-10
Set dir="auto" on changeset comment bodies


NeatNit:
### URL

https://www.openstreetmap.org/changeset/163401189

### How to reproduce the issue?

This is a sub-issue of #3428.

Some keywords for search: RTL bidi bidirectional LTR

Comments inherit the directionality of the entire website instead of the directionality of their content. Example screenshot:

![Image](https://github.com/user-attachments/assets/db2127c5-cb13-436b-969c-1a9e095bc1a3)

Each comment should have `dir="auto"` in its HTML. Poking around the source (which I have never worked with before, so this may be wrong), I think I've found the relevant code:

https://github.com/openstreetmap/openstreetmap-website/blob/a0768a4d04d3eb5afa9e8864b8fc73a64286bf2c/app/views/changesets/show.html.erb#L62-L64

So there is a "band-aid" solution and a more involved "full" solution that would produce better results.

Band-aid solution takes two steps:
1. Add `dir="auto"` to the above snippet:
    ```erb
          <div class="mx-2" dir="auto">
            <%= comment.body.to_html %>
          </div>
    ```
2. Add `text-align: start;` to the `mx-2` class CSS.

Result:

![Image](https://github.com/user-attachments/assets/4cf22bf6-d99a-473a-b8b3-e907fe35fdc8)

Perfect.

The full solution would take into account the fact that different paragraphs should have different directionality. For example, if I write:

```
This line is English and should be LTR.

השורה הזאת בעברית וצריכה להיות מימין לשמאל.

השורה הזאת מערבבת LTR ו-RTL אבל מתחילה ב-RTL ולכן צריכה להיות מימין לשמאל.
```

Github would render it correctly:

> This line is English and should be LTR.
> 
> השורה הזאת בעברית וצריכה להיות מימין לשמאל.
> 
> השורה הזאת מערבבת LTR ו-RTL אבל מתחילה ב-RTL ולכן צריכה להיות מימין לשמאל.

Each of these lines is `<p dir="auto">` so behaves correctly.

~~However it seems that `comment.body.to_html` *does not* split the text into different `<p>` elements, it's instead a single `<p>` with a single text block containing `<br />` for line splits. **This is not usable for expected formatting.** You would have to replace this function with one that produces separate elements (preferably `<p>`), each with `dir="auto"`. Couple with appropriate `text-align` CSS property, and the result would be ideal.~~

[...]

PR #5837, opened 2025-03-22
Put changeset and note comments in <article> elements

AntonKhorev:
Our `/changeset/:id` and `/note/:id` pages have comments in the sidebar. Those comments are marked up as list items: 
```html
<ul>
  <li id="c1">...</li>
  <li id="c2">...</li>
  ...
</ul>
```

Turn out there's a drawback to that: list items redefine text alignment inside them. Normally you wouldn't notice that everything is either left-aligned or right-aligned in case of rtl. But if you want mixed ltr and rtl text inside, you'll have to override `text-align`. See https://github.com/openstreetmap/openstreetmap-website/pull/5835#issuecomment-2745310604, see "Add text-align: start; to the mx-2 class CSS." in #5785.

Another way to mark up comments is to use `<article>` elements. There are examples of that on [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) and in the [html spec](https://html.spec.whatwg.org/multipage/sections.html#the-article-element). If we use `<article>`, #5835 will just need to add `dir="auto"` to paragraphs without modifying css.

```html
<article id="c1">
<article id="c2">
...
```

### DIFF
--- a/app/views/changesets/show.html.erb
+++ b/app/views/changesets/show.html.erb
@@ -36,35 +36,31 @@
     <% end %>
   </div>
 
-  <% if @comments.length > 0 %>
-    <ul class="list-unstyled">
-      <% @comments.each do |comment| %>
-        <% next unless comment.visible || current_user&.moderator? %>
-        <li id="c<%= comment.id %>">
-          <small class='text-body-secondary'>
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me, thanks.

I was going to say we should do the same for diary entries, but I see that already use `article` so there's nothing to do there.

PR #5839, opened 2025-03-23
Remove unused options argument from user_image_url

AntonKhorev:
Not in use since https://github.com/openstreetmap/openstreetmap-website/commit/f70edc02f086d97727453e738e9f0083c68ec706. Whatever it did would only apply to Gravatar images and not to other kind of images.


tomhughes:
Looks good to me, thanks.

PR #5840, opened 2025-03-23
Add dir="auto" to Markdown content for bidi

NeatNit:
### Description
1. Derive a kramdown converter subclass that adds `dir="auto"` to HTML elements where appropriate.
2. Call this new converter instead of the default HTML converter
3. Add `dir="auto"` to changeset comment tag (unrelated, should have been part of #5835)

This solves the last (to my knowledge) of the most prominent user-generated content bidi issues of #3428, that being user diaries. Other less prominent issues are handled in #3429, which best I can tell does not overlap with this PR.

### How has this been tested?
The kramdown subclass was tested here: https://codeberg.org/NeatNit/kramdown-bidi/src/branch/main/code.rb

Its integration into the OpenStreetMap website **has not been tested.**

I don't expect you'd like the amount of comments in that converter, or that github issue comment link in the source code. In my opinion it's needed to understand the choices I've made. If you prefer, I can remove some of the comments, just tell me which ones.

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -1,5 +1,45 @@
 # frozen_string_literal: true
 
+module Kramdown
+  module Converter
+    class HtmlBidi < Html
+      # Add dir="auto" for certain elements.
+
+      # Elements marked "container" or "child" are not viable in modern browsers unless and until the behaviour of dir="auto" changes:
+      # https://github.com/whatwg/html/issues/10097#issuecomment-2746324449
+      # instead, without dir, they'll inherit their parent's direction (for containers, likely the website's root direction)
+      [
+        'p', 'header',
+
+        'codespan', 'codeblock', # inner text will usually be LTR and should render accordingly
+
+        #'blockquote', # container
+
+        'ul', 'ol', # match Discourse; sometimes container, but this is probably the least bad solution
+        #'li', # child; sometimes container
+
+        'table', # similar to Discourse, where the <div> parent of <table> gets dir="auto" (no parent div in kramdown)
+        #'td', # child
+        #'th', # child; not actually a thing in kramdown but hypothetically might be in the future
+
+        'dl',
+        #'dd', 'dt', # child - since it's similar to a list, it has the same behavior as a list
+
+        'a', # <a> might need more sophisticated bidi handling
+        'math', # don't know how this ends up, but dir="auto" is probably correct
+      ].each do |name|
+        define_method :"convert_#{name}"  do |el, indent|
+          attr_bak = el.attr.dup # can't avoid mutating the attr hash, so make a backup
+          el.attr['dir'] = 'auto' unless el.attr.key?('dir') # if by some miracle dir is already defined, don't override it
+          ret = super(el, indent)
+          el.attr.replace(attr_bak) # restore backup
+          ret
+        end
+      end
+    end
+  end
+end
+
 module RichText
   SPAMMY_PHRASES = [
     "Business Description:", "Additional Keywords:"
@@ -96,7 +136,7 @@ module RichText
 
   class Markdown < Base
     def to_html
-      linkify(sanitize(document.to_html), :all)
+      linkify(sanitize(document.to_html_bidi), :all)
     end
 
     def to_text
--- a/app/views/changesets/show.html.erb
+++ b/app/views/changesets/show.html.erb
@@ -3,7 +3,7 @@
 <%= render "sidebar_header", :title => t(".title", :id => @changeset.id) %>
 
 <div class="browse-section">
-  <p class="fs-6 overflow-x-auto">
+  <p class="fs-6 overflow-x-auto" dir="auto">
     <%= linkify(@changeset.tags["comment"].to_s.presence || t("browse.no_comment")) %>
   </p>
   <%= tag.p :class => "details", :data => { :changeset => changeset_data(@changeset) } do %>


tomhughes:
I thought the previous PR was handling all rich text... If it doesn't handle markdown then it's of limited use - any HTML content is limited to legacy material from many years ago so it's really only things that are limited to plain text that it will affect.

lib/rich_text.rb
```diff
@@ -1,5 +1,45 @@
 # frozen_string_literal: true
 
+module Kramdown
```
tomhughes, line 3: Injecting code into the `Kramdown` namespace is not ideal as we don't own that.

lib/rich_text.rb
```diff
@@ -96,7 +136,7 @@ def to_text
 
   class Markdown < Base
     def to_html
-      linkify(sanitize(document.to_html), :all)
+      linkify(sanitize(document.to_html_bidi), :all)
```
tomhughes, line 139: How does this work? I don't see anything defining a `to_html_bidi` method?

NeatNit:
> I thought the previous PR was handling all rich text...

I'm sorry I didn't make that clear enough. Changeset comments and notes were the main (only?) thing affected by the previous PR. Legacy or not, they are extremely prominent and important. Are there any plans to move them to Markdown?

I did mention in the comments of the previous PR that I'll postpone Markdown for a separate PR, because at the time it seemed out of reach for me. This is that. I got across that gap faster than I thought. Again, sorry it wasn't clear enough.

tomhughes:
It's probably my fault as much as anything for not paying close enough attention to what the previous PR was doing.

I wonder if this would this be better done with a sanitize transformer that would work on the HTML regardless of how it was generated?

lib/rich_text.rb
NeatNit: https://kramdown.gettalong.org/rdoc/Kramdown/Parser/Base.html

> Implementing a new parser is rather easy: just derive a new class from this class and put it in the [Kramdown::Parser](https://kramdown.gettalong.org/rdoc/Kramdown/Parser.html) module – the latter is needed so that the auto-detection of the new parser works correctly.

lib/rich_text.rb
NeatNit: https://kramdown.gettalong.org/rdoc/Kramdown/Parser/Base.html

> Implementing a new parser is rather easy: just derive a new class from this class and put it in the [Kramdown::Parser](https://kramdown.gettalong.org/rdoc/Kramdown/Parser.html) module – the latter is needed so that the auto-detection of the new parser works correctly.

lib/rich_text.rb
NeatNit: Oops, wrong link. This is the one: https://kramdown.gettalong.org/rdoc/Kramdown/Converter/Base.html

> Implementing a new converter is rather easy: just derive a new class from this class and put it in the [Kramdown::Converter](https://kramdown.gettalong.org/rdoc/Kramdown/Converter.html) module (the latter is only needed if auto-detection should work properly).

Basically the same thing.

kramdown converts "html_bidi" to "HtmlBidi" and picks the class using that.

NeatNit:
> I wonder if this would this be better done with a sanitize transformer that would work on the HTML regardless of how it was generated?

I don't know. Is there anything wrong with this approach? It seems to be the cleanest way to do it within kramdown. In what way would sanitize be better?

NeatNit:
I just realized I should make that smarter handling of `<a>` tags before merging this. Marked as draft in the meantime.

AntonKhorev:
> I thought the previous PR was handling all rich text... If it doesn't handle markdown then it's of limited use - any HTML content is limited to legacy material from many years ago so it's really only things that are limited to plain text that it will affect.

Previous PR was for plaintext.

NeatNit:
I guess I should have mentioned that I did improve handling of `<a>` tags. Ready for your scathing reviews :P

NeatNit:
I'm considering rewriting this from scratch, with the following changes which would result in nicer code and better output:

1. Instead of subclassing the HTML converter, I will modify the document upon its creation. This would mean no longer touching the Kramdown namespace.
2. As I explain [here](https://github.com/whatwg/html/issues/10097#issuecomment-2746324449), `dir="auto"` is just not powerful enough to produce optimal results. I would like to implement server-side the algorithm I consider optimal, and set explicit `dir="ltr"` or `dir="rtl"` on each element for which it is relevant.

This would undoubtedly create the best bidi markdown implementation the world has ever seen!

tomhughes:
I really don't think we want to get into writing and maintaining our own complicated algorithms for determining direction, especially given the huge number of elements we would need to apply it to, and a especially before there is any consensus from the wider world about the correct behaviour - we're not trying to be a bleeding edge academic research project.

NeatNit:
In that case, I stand by the current version of this PR.

lib/rich_text.rb
```diff
@@ -1,5 +1,53 @@
...
+        "dl",
+        # "dd", "dt", # child - since it's similar to a list, it has the same behavior as a list
+
+        "math" # don't know how this ends up, but dir="auto" is probably correct
+      ].each do |name|
+        define_method :"convert_#{name}" do |el, indent|
+          attr_bak = el.attr.dup # can't avoid mutating the attr hash, so make a backup
+          el.attr["dir"] = "auto" unless el.attr.key?("dir") # if by some miracle dir is already defined, don't override it
+          ret = super(el, indent)
+          el.attr.replace(attr_bak) # restore backup
```
tomhughes, line 34: What happens if we don't restore the original attributes? Does the element get reused in a way that breaks things?

I can't say for sure as I couldn't find anything that broke when that restore was removed but would just doing `el = el.dup` as the first thing to duplicate the element be a workable alternative?

lib/rich_text.rb
```diff
@@ -1,5 +1,53 @@
...
+          ret
+        end
+      end
+
+      # only add dir="auto" to bare links
+      def convert_a(el, indent)
```
AntonKhorev, line 40: Do we need to set `dir` on inline elements too?

AntonKhorev:
This is a version that modifies the document, consisting of replaced `document` method, assuming we don't need to add `dir` on inline elements:

```ruby
    def document
      return @document if @document

      @document = Kramdown::Document.new(self)

      add_dir = lambda do |element|
        target_types = %w[p header pre codeblock ul ol table dl math]
        element.attr["dir"] ||= "auto" if target_types.include?(element.type == :html_element ? element.value : element.type.to_s)
        element.children.each(&add_dir)
      end
      add_dir.call(@document.root)

      @document
    end
```

lib/rich_text.rb
NeatNit: Yes.

טקסט `func(arg, 1235, "א")` טקסט

the codespand should render as: `func(arg, 1235, "א")`

lib/rich_text.rb
```diff
@@ -1,5 +1,53 @@
...
+        attr_bak = el.attr.dup
+        el.attr["dir"] = "auto" if !el.attr.key?("dir") && el.attr["href"] == inner(el, indent)
```
AntonKhorev, line 42: And this check `el.attr["href"] == inner(el, indent)`, is it required? The comment says "only add dir="auto" to bare links", which ones are bare? We add some links with autolinker later.

NeatNit: You're right, the autolinker needs to apply `dir="auto"` or `dir="ltr"` too. My code will pretty much only affect links made with the syntax `<https://example.com/>` which is pretty obscure IMHO.

Links with custom text (`[text](url)`) *should not* have `dir`. Links made with linkify need `dir="auto"` to handle cases like these:

מילים https://example.com/משהו/כלשהו

מילים https://example.com/משהו/כלשהו/index.html מילים

מילים https://example.com/ מילים

all of those links should render as isolated LTR. Note that GitHub falls here. Whereas: 

מספרים [one וגם two](https://example.com/numbers/) ועוד

the link direction should not be isolated - the word "one" should appear to the right of "two". Result on GitHub is correct.

NeatNit:
> This is a version that modifies the document, consisting of replaced `document` method, assuming we don't need to add `dir` on inline elements:
> 
> ```ruby
>     def document
>       return @document if @document
> 
>       @document = Kramdown::Document.new(self)
> 
>       add_dir = lambda do |element|
>         target_types = %w[p header pre codeblock ul ol table dl math]
>         element.attr["dir"] ||= "auto" if target_types.include?(element.type == :html_element ? element.value : element.type.to_s)
>         element.children.each(&add_dir)
>       end
>       add_dir.call(@document.root)
> 
>       @document
>     end
> ```

Thank you, I want to use this method instead of mine. Just a couple of questions:

Is it good practice to use a lambda like that? Wouldn't it be better to define a function normally and call that? (I'm a Ruby newbie, but this seems a bit strange to me)

And, is `element.type == :html_element` supposed to catch elements written as HTML within the Markdown source? In which case I think we should not add `dir="auto"` - if the user crafts their own HTML for whatever reason, we shouldn't change their work.

----

I will not be able to make modifications to the code for the next few days (until Tuesday or Wednesday, most likely), but I'd like to do this:

- [ ] Modify document upon creation instead of deriving a converter (i.e. what I replied to just now) 
- [ ] Make `linkify` add `dir="ltr"` to links - this would also affect changeset comments and the like
- [ ] Made triple sure that links made with `<URL>` markdown syntax actually get `dir="auto"` - I did not test this enough and it likely fails in some cases. Namely when the URL contains Unicode characters and they get replaced by `%D7%90` or whatever. 

I'll make these changes as soon as I can, next week. If any more feedbacks comes up I'll handle it too.

NeatNit:
Ready for review.

NeatNit:
@AntonKhorev Can I ping you to review this? I don't expect to need more changes.

NeatNit:
@gravitystorm or @tomhughes Perhaps you could review?

tomhughes:
This has been going for a while now and the latest version looks reasonable to me so I'm going to merge it - if anybody does think there are further improvements possible then they can be addressed in followup PRs.

PR #5842, opened 2025-03-25
Fix "undefined" comment when hiding resolved notes

AntonKhorev:
There's no comment textarea on resolved note pages, but pressing *Hide* if you're a moderator still tries to read a value and submit it as a comment.

image

With this PR the value is read only if the textarea is present.

### DIFF
--- a/app/assets/javascripts/index/note.js
+++ b/app/assets/javascripts/index/note.js
@@ -41,8 +41,11 @@ OSM.Note = function (map) {
             data = new URLSearchParams();
       content.find("button[name]").prop("disabled", true);
 
-      if (name !== "subscribe" && name !== "unsubscribe" && name !== "reopen") {
-        data.set("text", content.find("textarea").val());
+      if (name !== "subscribe" && name !== "unsubscribe") {
+        const textarea = content.find("textarea");
+        if (textarea.length) {
+          data.set("text", textarea.val());
+        }
       }
 
       fetch(url, {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/index/note.js
```diff
@@ -41,8 +41,11 @@ OSM.Note = function (map) {
             data = new URLSearchParams();
       content.find("button[name]").prop("disabled", true);
 
-      if (name !== "subscribe" && name !== "unsubscribe" && name !== "reopen") {
-        data.set("text", content.find("textarea").val());
+      if (name !== "subscribe" && name !== "unsubscribe") {
+        const textarea = content.find("textarea");
+        if (textarea.length) {
+          data.set("text", textarea.val());
+        }
```
hlfan, line 48: The API returns the same 400 error if data="" or data.text="".
Therefore the change can be simplified to this:
```suggestion
        data.set("text", content.find("textarea").val() || "");
```

tomhughes: Which API call are you talking about? Surely we want to avoid any errors?

AntonKhorev: I guess @hlfan wants to say that the OSM API reacts the same to empty `text` parameters and missing `text` parameters. Internally it's not true because note comment body is nullable. On the other hand we are unlikely to make a distinction between "" and null when we'll be deciding if what's currently stored as a comment is a real comment or not. And the following line is also ignoring this distinction because the comment model creates a `RichText` object even for null-comments:

https://github.com/openstreetmap/openstreetmap-website/blob/f1686b59cd92d2af077d49ef41440bc581016ac8/app/views/api/notes/_note.xml.builder#L30

This `if` is always true.

tomhughes:
I think this is probably fine as it is, thanks.

Issue #5780, opened 2025-03-08
pretty formatting of OSM URLS in notes, changeset comment, and comments


andrewharvey:
### Problem

Full URLs pasted within changeset comments and notes makes them harder to read, especially since they only have limited screen real estate.

### Description

As suggested by @1ec5 at https://github.com/openstreetmap/openstreetmap-website/issues/844#issuecomment-2708470716 we could take inspiration from GitHub and pretty format pasted URLs.

I'm thinking within any

1. changeset comment
2. comment on a changeset
3. note
4. note comment

Then format the content as follows

| Pasted URL | Formatted Display |
|---|---|
| https://www.openstreetmap.org/note/4655490 | [note/4655490](https://www.openstreetmap.org/note/4655490) |
| https://www.openstreetmap.org/changeset/163353772 | [changeset/163353772](https://www.openstreetmap.org/changeset/163353772) |
| https://www.openstreetmap.org/way/1249366504 | [way/1249366504](https://www.openstreetmap.org/way/1249366504) |
| https://www.openstreetmap.org/way/1249366504/history | [way/1249366504/history](https://www.openstreetmap.org/way/1249366504/history) |
| https://www.openstreetmap.org/way/1249366504/history/2 | [way/1249366504 (v2)](https://www.openstreetmap.org/way/1249366504/history/2) |
| https://www.openstreetmap.org/node/12639964186 | [node/12639964186](https://www.openstreetmap.org/node/12639964186) |
| https://www.openstreetmap.org/relation/7876483 | [relation/7876483](https://www.openstreetmap.org/relation/7876483) |
| https://www.openstreetmap.org/user/aharvey | [@aharvey](https://www.openstreetmap.org/user/aharvey) |
| https://wiki.openstreetmap.org/wiki/Key:boundary | [`boundary=*`](https://wiki.openstreetmap.org/wiki/Key:boundary) |
| https://wiki.openstreetmap.org/wiki/Tag:boundary=place | [`boundary=place`](https://wiki.openstreetmap.org/wiki/Tag:boundary=place) |
| `boundary=*` | [`boundary=place`](https://wiki.openstreetmap.org/wiki/Key:boundary) |
| `boundary=place` | [`boundary=place`](https://wiki.openstreetmap.org/wiki/Tag:boundary=place) |

hlfan:
Also n12639964186, w1249366504 and r7876483 could have links added as well; or be converted to the {object}/{id} format.

[...]

Issue #844, opened 2014-11-24
allow markdown in changeset comments


fuddl:
would be nice to link to something like another edit or a wiki page. or posting pictures

planemad:
Would love to see this. Having images (even thumbnails) adds great context to conversation and makes the thread more readable.

pnorman:
Do we do Markdown on anything API accessible?

andrewharvey:
would be nice to use selected wikitext templates like `{{tag|informal|yes}}` or alternatively just parse out a pasted URL like `https://wiki.openstreetmap.org/wiki/Tag:informal%3Dyes` and display it as [`informal`](https://wiki.openstreetmap.org/wiki/Key:informal)=[`yes`](https://wiki.openstreetmap.org/wiki/Tag:informal%3Dyes).

Suggesting tags to use within a changeset comment means either
1. enter the tag as `key=value` without any links (but then especially new users don't know what that means or where to read up about it)
2. enter the full URL but this makes the comment long and hard to read the exactly key=value from the url.
3. do both of the above, but the comment is still too long and harder to read

AntonKhorev:
> would be nice to use selected wikitext templates like {{tag|informal|yes}} or alternatively just parse out a pasted URL like https://wiki.openstreetmap.org/wiki/Tag:informal%3Dyes

But will those writing comments be bothered to paste the correct urls/templates?

If we just parse `key=value` (and maybe also `key=*`) as tag references, how many false positives/negatives we'll get?

andrewharvey:
> But will those writing comments be bothered to paste the correct urls/templates?

Good point, then perhaps as you suggest parsing key=value would be okay and/or reducing a pasted full wiki link.

Although the more we do this automatically the more important a preview/edit button is.

1ec5:
Many websites that accept Markdown end up pretty-printing raw URLs. For example, GitHub turns `https://github.com/openstreetmap/openstreetmap-website/issues/844#issuecomment-2705479356` into https://github.com/openstreetmap/openstreetmap-website/issues/844#issuecomment-2705479356. It isn’t the most discoverable syntax: you’d probably see people occasionally trying to replicate the presentation by hand as ```[`informal=yes`](https://wiki.openstreetmap.org/wiki/Tag:informal%3Dyes)```, but I don’t think that would be a huge problem.

andrewharvey:
[...]

PR #5844, opened 2025-03-25
Shorten matching urls in linkify

AntonKhorev:
After looking at #844 and #5780 I think that first we can do something with full osm urls. Before trying to match patterns like `key=value` or `changeset 12345` or `n12345` we need to check them for false positives/negatives among existing comments. And the first to do with urls that makes sense is to shorten them like [better-osm-org](https://github.com/deevroman/better-osm-org) does: replace `https://www.openstreetmap.org/` (and other osm domains) with `osm.org`. The only possible confusion here could be in comments like "When I tried opening https://www.osm.org/ ... but when I tried https://www.openstreetmap.org/ ..." that talk about different osm domains. Such comments are unlikely in plaintext changeset and note comments, and other editable texts support markdown syntax `[...](...)` where you can directly specify the visible text.

image

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -76,11 +76,7 @@ module RichText
     end
 
     def linkify(text, mode = :urls)
-      if text.html_safe?
-        Rinku.auto_link(text, mode, tag_builder.tag_options(:rel => "nofollow noopener noreferrer")).html_safe
-      else
-        Rinku.auto_link(text, mode, tag_builder.tag_options(:rel => "nofollow noopener noreferrer"))
-      end
+      Rinku.auto_link(ERB::Util.html_escape(text), mode, tag_builder.tag_options(:rel => "nofollow noopener noreferrer")).html_safe
     end
   end
 
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -76,7 +76,13 @@ module RichText
     end
 
     def linkify(text, mode = :urls)
-      Rinku.auto_link(ERB::Util.html_escape(text), mode, tag_builder.tag_options(:rel => "nofollow noopener noreferrer")).html_safe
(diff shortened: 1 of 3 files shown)


lib/rich_text.rb
```diff
@@ -76,11 +76,13 @@ def sanitize(text)
...
 
     def linkify(text, mode = :urls)
-      if text.html_safe?
-        Rinku.auto_link(text, mode, tag_builder.tag_options(:rel => "nofollow noopener noreferrer")).html_safe
-      else
-        Rinku.auto_link(text, mode, tag_builder.tag_options(:rel => "nofollow noopener noreferrer"))
-      end
+      link_attr = tag_builder.tag_options(:rel => "nofollow noopener noreferrer")
+      Rinku.auto_link(ERB::Util.html_escape(text), mode, link_attr) do |url|
+        %r{^https?://([^/]*)(.*)$}.match(url) do |m|
```
hlfan, line 81: Maybe put the www subdomain check into the regex:
```suggestion
        %r{^https?://(?:www\.)?([^/]*)(.*)$}.match(url) do |m|
```
Otherwise linkify_hosts / linkify_domains could get unwieldy long when including all the domains listed in [osm.wiki/wiki/Domain names](//osm.wiki/wiki/Domain_names):
```yaml
["openstreetmap.com", "openstreetmap.net", "openstreetmaps.org", "openmaps.org", "osm.org"]
```

AntonKhorev: It can only get 2x as long. And you copy the domain names from the chef repo. If the names can be managed there, they can be managed here.

tomhughes: The settings file can always use the multiline array syntax if the list is too unwieldy.

tomhughes:

PR #5846, opened 2025-03-26
Update strong_migrations to be compatible with rails 8

### LINKED ISSUE #5845: Migration run failed saying To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
### URL

_No response_

I'm trying to host on my local server, all was going well, but while running migration, got this error, To use retry middleware with Faraday v2.0+, install `faraday-retry` gem.
what to do? or how to fix this?
I am using docker and following the official documentation.

Image

gravitystorm:
We previously pinned to an old version for ruby 3.1 support, but that's no longer required since rails needs 3.2 anyway.

Fixes #5845

I'll follow up with the issue about why CI didn't catch this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,7 +66,7 @@ gem "rack-cors"
 gem "rails-i18n", "~> 8.0.0"
 gem "rails_param"
 gem "rinku", ">= 2.0.6", :require => "rails_rinku"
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #5848, opened 2025-03-26
Test migrations

### LINKED ISSUE #5847: Migrations don't run in CI with rails 8
We have code to run the migrations in CI, mostly to make sure that when they run, they produce the same results as structure.sql

When we upgraded to Rails 8, however, these migrations no longer run. I believe it's because, if rails spots an empty database, it now ignores the migrations and just runs the structure - but I'm not yet certain on that.

Here's two CI runs to compare, [before](https://github.com/openstreetmap/openstreetmap-website/actions/runs/13957644201/job/39072542124) and [after](https://github.com/openstreetmap/openstreetmap-website/actions/runs/14069719853/job/39400815755) the rails 8 upgrade

Image

Image

We need to figure out how to make rails 8 actually run the migrations.

(This new behaviour also masked #5845 / #5846 from being detected by CI)

tomhughes:
Fixes #5847 by forcing migrations to run in test, and fixing some noise that had crept into the structure as a result of them not running.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,6 +45,7 @@ jobs:
     - name: Populate database
       run: |
         sed -f script/normalise-structure db/structure.sql > db/structure.expected
+        rm -f db/structure.sql
         bundle exec rails db:migrate
         sed -f script/normalise-structure db/structure.sql > db/structure.actual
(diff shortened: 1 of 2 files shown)


gravitystorm:
Merged, thanks for figuring this out!

PR #5850, opened 2025-03-27
Remove TagHelper include from RichText module

AntonKhorev:
We use TagHelper module to convert `:rel => "nofollow noopener noreferrer"` into `'rel="nofollow noopener noreferrer"'`, probably because [Rinku](https://github.com/vmg/rinku)'s readme says thet we can generate an attribute string from a hash. And to do that we have to call what is now a private method of TagHelper; maybe it wasn't private when the readme was last updated.

Or we could just write a string.

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -25,8 +25,6 @@ module RichText
   end
 
   class Base < String
-    include ActionView::Helpers::TagHelper
-
     def spam_score
       link_count = 0
       link_size = 0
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

NeatNit:
I noticed that linkified links only have `rel="nofollow"` instead of the full `rel="nofollow noopener noreferrer"`. I don't think this is my browser or extensions doing funky stuff because I see it in the raw response data in the Network tab of Firefox's dev tools. Is this supposed to happen?

AntonKhorev:
We got another linkify in ApplicationHelper. That one only adds `rel="nofollow"`.

NeatNit:
I see... So when is the Rich Text one used? Seems weird that even rich text seems to get the ApplicationHelper one.

PR #5851, opened 2025-03-27
Support interactive debugger with docker-compose

rkoeze:
### Description

Configures the `web` service in `docker-compose.yml` file to allow for interactive debugging. `tty` allocates a pseudo terminal while `stdin_open` runs the container with stdin. Running `docker attach WEB_CONTAINER_ID` connects to the running container where the debugger can be interacted with.

One alternative is to run `docker-compose run --rm --service-ports -it web bash` and then `bundle exec rails s -p 3000 -b '0.0.0.0'` inside the container. However, I think leaning into `docker-compose.yml` makes sense here — we get consistent terminal behavior as a default when running `docker compose up` and `docker attach`. 

### How has this been tested?
I ran `docker-compose up` and then in a separate terminal ran `docker attach WEB_CONTAINER_ID`.  When I hit a `debugger` I can then step through the code.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,8 @@ services:
       /tmp/pids/
     ports:
       - "3000:3000"
+    stdin_open: true
+    tty: true
     command: bundle exec rails s -p 3000 -b '0.0.0.0'
     depends_on:
       - db


nenad-vujicic (approved):
This works great on my side and makes debugging faster. Thanks!

rkoeze:
@nenad-vujicic thanks for the review! It occurs to me that it might be useful to add this to the [Docker documentation](https://github.com/openstreetmap/openstreetmap-website/blob/master/DOCKER.md).  I will do that in a separate PR.

danieldegroot2:
For features used only for debugging, it might be useful to add a comment. (see also https://github.com/openstreetmap/openstreetmap-website/pull/1290#discussion_r79106626 )

tomhughes:
As I don't use docker and I've never used the interactive debugger it's hard for me to comment on this, and would need a lot of work for me to even test it so I think I'm going to rely on @nenad-vujicic having tested it and the fact that there doesn't seem to be anything here that's likely to cause problems in any case.

Thanks for working on it, and sorry for the delay in reviewing and merging it.

PR #5852, opened 2025-03-27
Remove extra sanitize after simple_format RichText::HTML

AntonKhorev:
`simple_format` already calls `sanitize` internally. We don't call `sanitize` again from a similar place of `RichText::Text`.


tomhughes:
Looks food to me, thanks.

HolgerJeromin:
> Looks food to me, thanks.

you can use "saved replies" here at github...

Issue #5804, opened 2025-03-15
Make contribution heat map opt in


Discostu36:
### Problem

There have been several critical comments about the new contribution heat map feature. Some users see the publication of their contribution patterns as a violation of their privacy or even a breach of GDPR.

Some of the discussions can be found here:
- https://community.openstreetmap.org/t/umfrage-aktivitatskalender-auf-jeder-osm-user-seite/127294
- https://en.osm.town/@osm_tech/114160805593628586

### Description

Make this feature opt-in. Users need to activate it in their preferences for the heat map to be displayed.

simonpoole:
The OSMF privacy policy guarantees that contribution meta data will only be shown to logged in users https://osmfoundation.org/wiki/Privacy_Policy#Who_has_access_to_the_data (yes, the OSMF violates this all over the place, but that doesn't give anybody a free get out of jail card to add to those violations). 

So either the privacy policy needs to be changed, iffy because this changes the basic tenants it is built on and not a question that is of a technical nature, or the feature needs to only available to logged in users as I pointed out in the relevant PR back in January.

Making the feature configurable is a good idea in any case, heck even Microsoft (github) does that.

deevroman:
How about hiding pages `osm.org/user/username` from not registered users by default? And after that, you can add a setting of the visibility of the profile data. Yes there is more osm.org/user/username/*, but at least it will be consistent with the API, which does not show profile descriptions without authorization.

As I understand it, this should still be done to comply with the privacy policy, so why limit yourself to configuring the heat map.

p.s. despite my position on the situation, here I write without sarcasm and exaggeration

simonpoole:
See https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services iirc at the time it was assumed that the the the user page would only contain content that the user had added with the intent of making it public, see https://osmfoundation.org/wiki/Privacy_Policy#Data_you_may_voluntarily_add_to_your_profile_or_diary_posts

tomhughes:
I would argue that by signing up to the project it's your intent to make your edits public - building a public data set is the entire point of the project after all but what do I know.

I certainly don't see any point in showing things to logged in users that isn't shown to logged out users given there is no barrier at all to signing up.

simonpoole:
> I would argue that by signing up to the project it's your intent to make your edits public - building a public data set is the entire point of the project after all but what do I know.

[...]

PR #5853, opened 2025-03-27
Add hide/show contribution heatmap selection on edit profile section

kcne:
### Description  
This PR introduces a user preference to toggle the visibility of the contribution heatmap on their profile page. Users can now enable or disable this feature through the edit profile section. Referencing #5804 

**Changes include:**  
- Added `show_contribution_heatmap` boolean field to the `users` table (default: `true`)  
- Added a checkbox in the profile edit form to manage heatmap visibility  
- Updated the `users` controller to respect this preference when rendering profile data  
- Added a test to ensure the feature works as expected  

### How has this been tested?  
The changes have been verified through:  
1. Automated tests  
2. Manual testing by creating a new user and toggling the setting in the edit profile form  

### Screenshots:
Screenshot 2025-03-27 at 09 02 05


kcne:
@AntonKhorev @tomhughes Just a friendly ping..
When you get a chance to take a look, I'd appreciate any feedback—just let me know if this is heading in the right direction or if you'd like to see any changes. Thank you.

tomhughes:
I haven't really reviewed the PRs relating to hiding the heatmap because I was waiting to see how the policy questions played out so I would know what we needed to do.

hlfan:
Has there been any policy decision yet?

1ec5:
> Has there been any policy decision yet?

By the way, I think this is in reference to the question in https://github.com/openstreetmap/openstreetmap-website/pull/5706#issuecomment-2744397402 about whether the terms of service apply to everyone automatically or only those who’ve affirmatively checked a box on the website. We’re still waiting for the outcome of that discussion. In other tickets, there was also a discussion about whether hiding the heatmap would disadvantage other users who need to inspect contribution history to weed out misuse.

Regardless, do we anticipate that the website will ultimately have an option to show or hide the heatmap? We could always fine-tune the behavior tied to this option later based on either discussion. (For example: “Display a heatmap of your contributions on your profile page to eligible logged-in users.”)

tomhughes:
I don't see any great need for an option personally, but it people really want one then fine. I'm not sure what happens if we start adding more activity related information to the profile though - do we have a long list of options to turn things on and off? or just one to hide everything?

1ec5:
> By the way, I think this is in reference to the question in https://github.com/openstreetmap/openstreetmap-website/pull/5706#issuecomment-2744397402 about whether the terms of service apply to everyone automatically or only those who’ve affirmatively checked a box on the website. We’re still waiting for the outcome of that discussion.

As noted in https://github.com/openstreetmap/openstreetmap-website/pull/5706#issuecomment-2990218356, per the LWG, users need to acknowledge the terms of use explicitly. So a decision on that is no longer a blocker for landing this change.

> I'm not sure what happens if we start adding more activity related information to the profile though - do we have a long list of options to turn things on and off? or just one to hide everything?

Under the hood, it’s probably a good idea to track heatmap display independently of other activity-related features, in case we need more nuance down the line. In terms of the UI, it’s only one option for now, which we could always generalize in the future if needed. #6084 split the profile options into multiple pages, so we’ll need to figure out which page should list the heatmap option. The existing pattern seems to be optimized for having a specific page to serve as the target of each button on the profile page. If we follow this pattern, we’ll need to add a button near the heatmap and create a new page for that option, at the risk of extending the row of tabs farther to the right.

db/migrate/20250317000001_add_show_contribution_heatmap_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddShowContributionHeatmapToUsers < ActiveRecord::Migration[7.2]
+  def change
+    add_column :users, :show_contribution_heatmap, :boolean, :default => true, :null => false
```
pablobm, line 3: Two questions on this one:
- How about using the existing user preferences instead of a new column?
- If we proceed like this: it ok to create a column with a default value in a large table? My recollection is that the column may take some time to fill out, perhaps even locking the table. I can't be sure though. Perhaps @Firefishy can refresh my memory?

pablobm (commented):
Looks like we can finally move this forward! 🚀 @kcne - Would you be able to rebase? There have been significant changes to the user profile page since this PR was created, so it will need some rework. Also I have left a comment about the data model.

db/migrate/20250317000001_add_show_contribution_heatmap_to_users.rb
pablobm: In case anyone reads this in the future: it appears that this used to be an issue, as Postgres would lock the table to fill in all the default values. This is not the case any more.

tomhughes: Specifically that changed with [Postgres 11](https://www.postgresql.org/docs/release/11.0/):

> Many other useful performance improvements, including the ability to avoid a table rewrite for ALTER TABLE ... ADD COLUMN with a non-null column default

gravitystorm: > * it ok to create a column with a default value in a large table?

We use [strong migrations](https://github.com/ankane/strong_migrations) to check that the migrations are safe to run in production. This means that contributors don't need to learn all the edge cases, and maintainers don't need to know them all either and can merge PRs with confidence if CI is green.

Previously strong migrations would flag this situation for PG < 11, but it now supports only PG 12+ and so it [dropped the check last year](https://github.com/ankane/strong_migrations/commit/1e428665978185bc3b21a36db723dd91c10b5e12#diff-2c318d74d04a82865bea4780a048a37f4871393db99d20b798eb19516c188104R44-L46)

kcne:
Closing in favour of #6606

1ec5:
Thank you for your work on this PR, @kcne!

Issue #2787, opened 2020-08-21
Adjust start/finish icons used in directions feature


boothym:
Currently the only difference between the start and finish directions icons in the directions feature is the green and red colours.

This is obviously not good for colourblind users - there should be something other colour to differentiate between them.

For example, cycle.travel uses a chequered flag for the finish icon: ![red marker with chequered flag](https://cycle.travel/site_images/route_end.png)

1ec5:
Even as someone who can distinguish the two colors, I always get them mixed up when a site uses these green and red pins. For some reason, my head says red = not yet started, green = complete.

Issue #872, opened 2015-01-15
can't activate "internal" search results


richlv:
search for some coordinates (like "N57 00.549 E024 07.196"), see the internal result being returned.
click on it, notice zoom happening and marker appearing, try to zoom in on it... as soon as mouse cursor is moved away, marker disappears.

tested with opera 12.16 on linux

tomhughes:
I think this is deliberate - search results are highlighted as you hover over them. Otherwise there would be no obvious time when the marker could be removed again.

richlv:
but for object results (node, way etc) sidebar is opened and marker does not disappear.
with the coord results, zooming/panning is very inconvenient as the marker disappears as soon as user gets to the map area.
maybe similar sidebar could be used on the page to keep the coords "activated" ?

kerosin:
agree
If clicking on Nominatim results you are getting a sidebar and a outline of your request on your mapview. Clicking on a result from "Internal" or "Geonames" does nothing (expect the hover feature).
This is inconsistent. User interactions like clicking shouldn't be ignored. I would expect a static marker for the clicked result when clicking.

stephan75:
The author of https://trac.openstreetmap.org/ticket/5283 asked to have a cross-link to this github issue.

harry-wood:
I noticed Robert Orzanna tweeted about this issue today, complete with animated gif:
https://twitter.com/orschiro/status/713026546353700864 

Maybe in the case of a coordinates result, when you click, it could just reload the page with a marker permalink URL (one like this I mean: http://www.openstreetmap.org/?mlat=51.50774&mlon=-0.12794 )

orschiro:
Dear @harry-wood and others, 

Indeed I tweeted about this problem yesterday.

I searched for exact coordinates and on click of the search result, I expected the marker to stay activated so that I can zoom into the map. However, the marker does not remain persistent.

Please let me know if I can provide any more details or help solve the problem.

Thankfully,

~Robert

NoSuck:
I also expected that activating an “internal” search-results coordinate would add its marker to the map.  This is a frustrating issue.

fgnievinski:
searching by coordinates is a common use case. currently when hovering on a link in the list of matches, it displays a temporary marker on the map. additionally, clicking on a match causes the map to be centered at the marker (panning without zooming). 

[...]

PR #5854, opened 2025-03-27
Switch to inline colourable markers

hlfan:
### Description
This is an attempt at parametrifying marker colors with `currentColor` since it has come up in #5764.
Together with optimizing the path geometry, this required rewriting the marker layout to be compatible with partially transparent gradients.
While an even simpler path has been proposed and demonstrated in https://antonkhorev.github.io/msvgen, I was searching for a solution with G² curvature continuity, which I found by constraining a Bezier curve to a specified slope, 1.92 in this case.[^1] (See [CodePen](//codepen.io/hlfan/pen/OPJOGxg))

Where I had the source colors, I tried to fit the colors to the sources, e.g. for the note markers see this before/after:
image

For now, this is only a barebones switch to the new markers, but moving in the direction of https://github.com/openstreetmap/openstreetmap-website/issues/872#issuecomment-2589899767, #2787 or #507

Closes #5764.

### How has this been tested?
Server side and client/browser side separately, no integration testing though.

### Considerations
For the raster icons where I didn't have the colors, I took them from the note markers, so
- either I need better sources than I could find, or
- the dot markers will differ in color a bit more:
image

[^1]: Leaflet's marker.svg actually has two slopes for the tip, 1.9205 on the left and 1.9296 on the right


hlfan:
So apparently `DivIcon` does not support shadows. (https://github.com/Leaflet/Leaflet/issues/5083#issuecomment-307381003)

app/views/notes/index.html.erb
```diff
@@ -26,6 +26,10 @@
 <% else %>
   <%= render :partial => "notes_paging_nav" %>
 
+  <svg width="0" height="0" class="end-100 position-absolute">
```
tomhughes, line 29: As all three places that include the partial wrap it in the same `svg` element why not include that in the partial?

app/views/notes/index.html.erb
```diff
@@ -26,6 +26,10 @@
...
+    <%= render :partial => "layouts/defs_markers", :locals => { :types => %w[cross tick] } %>
```
tomhughes, line 30: What's the logic behind `defs_markers` as the name? I assume defs is definitions or something? I suggest just using `markers` as the name I think.

hlfan: That and `<defs>` being the subtree root. Not needed when it includes the SVG tag.

hlfan:
@tomhughes do you think it's better to match the current marker colors more closely or to reduce variations in color (only have one green, one red...)?

tomhughes:
I couldn't see that much of a difference to be honest - there was a bit of one with the routing markers but nothing terrible.

I know this overlaps with work @AntonKhorev has been doing though which is why I've asked him to review it.

tomhughes:
The only other thing that struck me was the various hex colour constants scattered around but I'm not sure there's any good way to centralise them.

hlfan:
Maybe a central dict in getMarker? That would also account for unifying the colors.

hlfan:
Do we need to involve more overhead for storing the colors or is putting it in an initializer good enough?

hlfan:
Hey @AntonKhorev, when you get a chance, would love your eyes on this one if it’s not getting in the way of your daily PR

hlfan:
Should I split up the "Update and inline markers" commit into three parts and temporarily make `OSM.getMarker` use a dict so I can update the options separately?
Would that make this PR easier to digest?

app/views/layouts/_search.html.erb
```diff
@@ -68,13 +68,17 @@
         <div class="d-flex flex-column gap-1 flex-grow-1">
           <div class="d-flex align-items-center">
             <div class="routing_marker_column position-absolute">
-              <%= image_tag "marker-green.png", :class => "img-fluid", :data => { :type => "from" }, :draggable => "true" %>
+              <svg viewBox="0 0 25 40" class="img-fluid" data-type="from" draggable="true">
+                <use href="#pin-dot" color="<%= MARKER_COLORS[:green] %>" />
+              </svg>
```
AntonKhorev: Changes I had to make to keep drag events working (1/3):
```suggestion
              <span draggable="true" data-type="from">
                <svg viewBox="0 0 25 40" class="img-fluid" draggable="true">
                  <use href="#pin-dot" color="<%= MARKER_COLORS[:green] %>" />
                </svg>
              </span>
```

app/views/layouts/_search.html.erb
```diff
@@ -68,13 +68,17 @@
...
             </div>
             <%= text_field_tag "route_from", params[:from], :placeholder => t("site.search.from"), :autocomplete => "on", :class => "form-control py-1 px-2 ps-4", :dir => "auto" %>
           </div>
           <div class="d-flex align-items-center">
             <div class="routing_marker_column position-absolute">
-              <%= image_tag "marker-red.png", :class => "img-fluid", :data => { :type => "to" }, :draggable => "true" %>
+              <svg viewBox="0 0 25 40" class="img-fluid" data-type="to" draggable="true">
+                <use href="#pin-dot" color="<%= MARKER_COLORS[:red] %>" />
+              </svg>
```
AntonKhorev: Changes I had to make to keep drag events working (2/3):
```suggestion
              <span draggable="true" data-type="to">
                <svg viewBox="0 0 25 40" class="img-fluid" draggable="true">
                  <use href="#pin-dot" color="<%= MARKER_COLORS[:red] %>" />
                </svg>
              </span>
```

app/assets/javascripts/index/directions.js
```diff
@@ -156,14 +156,18 @@ OSM.Directions = function (map) {
...
     if (dt.setDragImage) {
-      const img = $("<img>").attr("src", $(e.originalEvent.target).attr("src"));
+      const img = jqthis.clone()
+        .toggleClass("img-fluid position-absolute bottom-100 end-100")
+        .attr({ width: "25", height: "40" })
+        .appendTo(document.body);
       dt.setDragImage(img.get(0), 12, 21);
+      setTimeout(() => img.remove(), 0);
     }
   });
```
AntonKhorev: Changes I had to make to keep drag events working (3/3):
```js
  $(".routing_marker_column > span").on("dragstart", function (e) {
    const dt = e.originalEvent.dataTransfer;
    dt.effectAllowed = "move";
    const jqthis = $(this);
    dt.setData("text", JSON.stringify(jqthis.data()));
    if (dt.setDragImage) {
      const img = jqthis.clone()
        .appendTo(document.body);
      img.find("svg")
        .toggleClass("img-fluid position-absolute bottom-100 end-100")
        .attr({ width: "25", height: "40" });
      dt.setDragImage(img.get(0), 12, 21);
      setTimeout(() => img.remove(), 0);
    }
  });
```

app/views/dashboards/show.html.erb
```diff
@@ -3,6 +3,7 @@
 <% end %>
 
 <svg class="d-none"><%= render :partial => "layouts/control_icons", :locals => { :icons => %w[zoomin zoomout geolocate] } %></svg>
+<%= render :partial => "layouts/markers", :locals => { :types => %w[dot] } %>
```
AntonKhorev, line 6: The same thing is needed in:
- `app/views/profiles/locations/show.html.erb`
- `app/views/diary_entries/new.html.erb`

config/initializers/marker_colors.rb
```diff
@@ -0,0 +1,5 @@
+MARKER_COLORS = {
+  :red => "#f6110a",
+  :green => "#9cef11",
+  :blue => "#0b8ef1"
+}.freeze
```
AntonKhorev: Why not css properties? That's where I'd expect to find color definitions.

You can put something like this in `common.scss`:
```css
:root {
  --marker-red: #f6110a;
  --marker-green: #9cef11;
  --marker-blue: #0b8ef1;
}
```

Then you can reference them like this: `color="var(--marker-green)"`.

You won't need an initializer, and you probably won't need embed anything into js like `<%= MARKER_COLORS.to_json %>`.

AntonKhorev: Although there is a gotcha with print css. If markers are supposed to be colored in print style, these colors need to be included there too.

app/views/layouts/_markers.html.erb
```diff
@@ -0,0 +1,44 @@
+<svg width="0" height="0" class="end-100 position-absolute">
```
AntonKhorev, line 1: I know that `d-none` here won't entirely work, but this is also not entirely working. You'll get this gap in print
image
which doesn't look like much, except on map layouts you'll get two pages instead of one.

hlfan: Added a copy of #5999 for `position-absolute`.

hlfan:
You bet I've looked into using `marker` elements to make (a) the result smaller and (b) remove the need for erb-ing the SVG file.
And even then, it's all wrapped in a `defs` element, so you wouldn't really be able to see any preview.

PR #5855, opened 2025-03-27
Shorten matching wiki urls in linkify

AntonKhorev:
Like #5844, but for wiki links that are also common in comments.

Before:
image

After:
image


deevroman:
Do need to save /wiki? Wiki understands paths without /wiki and makes a redirect.

AntonKhorev:
> Do need to save /wiki? Wiki understands paths without /wiki and makes a redirect.

Maybe later. This won't work for some links. For example `/wiki/w` and `/w` are going to show different pages.

tomhughes:
Looks good to me, thanks.

AntonKhorev:
> Wiki understands

Actually Apache understands, using a [config](https://github.com/openstreetmap/chef/blob/0cad06b01638b433858a5be37f87c236084cb743/cookbooks/mediawiki/templates/default/apache.erb#L72) with a bunch of `RewriteCond`/`RewriteRule`.

Issue #5856, opened 2025-03-27
The word "followings" should probably be replaced in English strings


amire80:
"Friend" was replaced with "following" in 2b7a48029bff06c190544f91c5969b09aeae552b. I understand the general idea of it, but this created several instances of the word "followings", which sounds weird in this context as a plural English noun.

It should probably be refined to something like "people you follow", or whatever works best in the context.

AntonKhorev:
Fixed in #5857

PR #5857, opened 2025-03-27
Rephrase strings with the word "followings"

amire80:
See issue #5856.

I updated these strings to the best of my understanding, but I'm open to better suggestion to rephrase those strings.

Thank you!

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -578,7 +578,7 @@ en:
       use_map_link: Use Map
     index:
       title: "Users' Diaries"
-      title_followed: "Your Followed Users' diaries"
+      title_followed: "Followed Users' diaries"
       title_nearby: "Nearby Users' Diaries"
       user_title: "%{user}'s Diary"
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -578,7 +578,7 @@ en:
       use_map_link: Use Map
     index:
       title: "Users' Diaries"
-      title_followed: "Followings' Diaries"
+      title_followed: "Your Followed Users' diaries"
```
tomhughes, line 581: I think this could just be `Followed Users' diaries` which aligns well with the nearby users phrasing.

tomhughes (changes requested):
Other than the one minor suggestion I made this looks great to me.

config/locales/en.yml
amire80: Done in another commit. Thank you!

hlfan (approved):

tomhughes (approved):
Thanks for the fix. I think this looks good now.

PR #5860, opened 2025-03-28
Consolidate leaflet marker creation in OSM.getMarker

hlfan:
OSM.getUserIcon() was used for much more than just the icons on the user page.
I've made it work with the icon paths already stored in the OSM object and made it compatible with the note icons to eliminate the `noteIcons` objects.

I previously included a variant of this in #5854, but I think it makes more sense as a separate PR.

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -391,13 +391,17 @@ OSM.isDarkMap = function () {
   return window.matchMedia("(prefers-color-scheme: dark)").matches;
 };
 
-OSM.getUserIcon = function (url) {
-  return L.icon({
-    iconUrl: url || OSM.MARKER_RED,
-    iconSize: [25, 41],
-    iconAnchor: [12, 41],
-    popupAnchor: [1, -34],
-    shadowUrl: OSM.MARKER_SHADOW,
-    shadowSize: [41, 41]
-  });
+OSM.getMarker = function ({ icon = "MARKER_RED", shadow = true, height = 41 }) {
+  const options = {
+    iconUrl: OSM[icon.toUpperCase()] || OSM.MARKER_RED,
+    iconSize: [25, height],
+    iconAnchor: [12, height],
+    popupAnchor: [1, -34]
+  };
+  if (shadow) {
+    options.shadowUrl = OSM.MARKER_SHADOW;
+    options.shadowSize = [41, 41];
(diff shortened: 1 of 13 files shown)


app/assets/javascripts/leaflet.map.js
```diff
@@ -391,13 +391,17 @@ OSM.isDarkMap = function () {
...
-  return L.icon({
-    iconUrl: url || OSM.MARKER_RED,
-    iconSize: [25, 41],
-    iconAnchor: [12, 41],
-    popupAnchor: [1, -34],
-    shadowUrl: OSM.MARKER_SHADOW,
-    shadowSize: [41, 41]
-  });
+OSM.getMarker = function ({ icon = "MARKER_RED", shadow = true }) {
+  const height = icon.includes("NOTE") ? 40 : 41;
```
tomhughes, line 395: Should `height` be a parameter rather than trying to guess it from the name? It's a bit odd that we do it for height but make the user pass a flag for shadow?

app/assets/javascripts/leaflet.map.js
```diff
@@ -391,13 +391,17 @@ OSM.isDarkMap = function () {
...
+  const options = {
+    iconUrl: OSM[icon.toUpperCase()] || OSM.MARKER_RED,
+    iconSize: [25, height],
+    iconAnchor: [12, height],
+    popupAnchor: [1, -34]
+  };
+  if (shadow) {
+    options.shadowUrl = OSM.MARKER_SHADOW;
+    options.shadowSize = [41, 41];
```
tomhughes, line 404: Should this use `height` as the second value? I don't think it actually makes any difference right now as we only switch height for notes which always pass disable shadows but if we're trying to make it generic then maybe it should?

hlfan: The size of the PNG doesn't change just because the marker does. [`shadowAnchor`](https://leafletjs.com/reference.html#icon-shadowanchor) shouldn't change either, so it needs to be specified separately.

app/assets/javascripts/leaflet.map.js
hlfan: Probably including height in the parameters is a better solution. #5854 which I wrote this im mind with would remove that change anyway so I didn't think about that.

tomhughes (approved):
Thanks for the change - this looks good to me now.

PR #5861, opened 2025-03-28
Fix copy-pasting breaking links in shorten_host

hlfan:
Removes the check for removed protocol that fails when copy-pasting links. See https://github.com/openstreetmap/openstreetmap-website/issues/5780#issuecomment-2760445765 for details.

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -75,7 +75,9 @@ module RichText
 
     def linkify(text, mode = :urls)
       link_attr = 'rel="nofollow noopener noreferrer"'
-      Rinku.auto_link(ERB::Util.html_escape(text), mode, link_attr) do |url|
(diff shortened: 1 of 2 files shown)


HolgerJeromin (approved):

AntonKhorev:
> breaking links

Which links are being broken?

hlfan:
Copying and pasting shortened osm.org links makes the pasted links not getting linkified, as the current regex requires a  protocol that is missing in the copied shortened links.

AntonKhorev:
Can you write a test case that fails without your fix and passes with it?

hlfan:
Sure, I'll see what I can do

AntonKhorev:
You'll see that your fix doesn't fix anything because that regexp is not used to detect links.

PR #5862, opened 2025-03-28
Simpliy linkify normalisation to use a single array of rules

tomhughes:
Rather than adding an ever larger list of configuration options for link normalisation this uses a single option that provides an array of rules to apply.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -136,16 +136,13 @@ overpass_credentials: false
 graphhopper_url: "https://graphhopper.com/api/1/route"
 fossgis_osrm_url: "https://routing.openstreetmap.de/"
 fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
-# Main website hosts to match in linkify
-linkify_hosts: ["www.openstreetmap.org", "www.osm.org", "www.openstreetmap.com", "openstreetmap.org", "osm.org", "openstreetmap.com"]
-# Shorter host to replace main hosts
-linkify_hosts_replacement: "osm.org"
-# Wiki website hosts to match in linkify
-linkify_wiki_hosts: ["wiki.openstreetmap.org", "wiki.osm.org", "wiki.openstreetmap.com", "wiki.openstreetmaps.org", "osm.wiki", "www.osm.wiki", "wiki.osm.wiki"]
-# Shorter host to replace wiki hosts
-linkify_wiki_hosts_replacement: "osm.wiki"
-# Regexp for wiki prefix that can be removed
-linkify_wiki_optional_path_prefix: "^/wiki(?=/[A-Z])"
+# Normalisation rules for linkify
+linkify_rules:
+  - hosts: ["www.openstreetmap.org", "www.osm.org", "www.openstreetmap.com", "openstreetmap.org", "osm.org", "openstreetmap.com"]
+    host_replacement: "osm.org"
+  - hosts : ["wiki.openstreetmap.org", "wiki.osm.org", "wiki.openstreetmap.com", "wiki.openstreetmaps.org", "osm.wiki", "www.osm.wiki", "wiki.osm.wiki"]
+    host_replacement: "osm.wiki"
+    path_prefix: "^/wiki(?=/[A-Z])"
 # External authentication credentials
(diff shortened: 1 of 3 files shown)


AntonKhorev:
I tried to do a similar thing back in #4189 and it wasn't merged.

config/settings.yml
```diff
@@ -136,16 +136,13 @@ overpass_credentials: false
...
-# Wiki website hosts to match in linkify
-linkify_wiki_hosts: ["wiki.openstreetmap.org", "wiki.osm.org", "wiki.openstreetmap.com", "wiki.openstreetmaps.org", "osm.wiki", "www.osm.wiki", "wiki.osm.wiki"]
-# Shorter host to replace wiki hosts
-linkify_wiki_hosts_replacement: "osm.wiki"
-# Regexp for wiki prefix that can be removed
-linkify_wiki_optional_path_prefix: "^/wiki(?=/[A-Z])"
+# Normalisation rules for linkify
+linkify_rules:
+  - hosts: ["www.openstreetmap.org", "www.osm.org", "www.openstreetmap.com", "openstreetmap.org", "osm.org", "openstreetmap.com"]
+    host_replacement: "osm.org"
```
hlfan, line 142: idk if anybody still uses these, but I guess `path_prefix: "^/browse(?=/\w+)"` would fit.

config/settings.yml
```diff
@@ -136,16 +136,13 @@ overpass_credentials: false
...
+  - hosts : ["wiki.openstreetmap.org", "wiki.osm.org", "wiki.openstreetmap.com", "wiki.openstreetmaps.org", "osm.wiki", "www.osm.wiki", "wiki.osm.wiki"]
+    host_replacement: "osm.wiki"
+    path_prefix: "^/wiki(?=/[A-Z])"
```
AntonKhorev, line 145: With this name you'll have to guess what this prefix means. The first guess will likely be that it has to match if any url transformation is to happen, and it's not the case.

hlfan:
Would you be able to rebase this?

tomhughes:
Done, though maybe `linkify_rules` should be moved to `linkify.normalisation_rules` in the config?

hlfan:
Yeah, and I'd say `linkify_rules[].path_prefix` to `linkify.normalisation_rules[].optional_path_prefix` too.

tomhughes:
Updated with those changes now...

lib/rich_text.rb
```diff
@@ -133,9 +131,12 @@ def auto_link(text, mode)
     end
 
     def format_link_text(url)
-      url = shorten_host(url, Settings.linkify_hosts, Settings.linkify_hosts_replacement)
-      url = shorten_host(url, Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement) do |path|
-        path.sub(Regexp.new(Settings.linkify_wiki_optional_path_prefix || ""), "")
+      url = Array
```
hlfan, line 126: Splitting the host and path related stuff made the expansion easier to read:

https://github.com/openstreetmap/openstreetmap-website/blob/c1dc08b5308b47ba5579fde79123f5588e4f2427/lib/rich_text.rb#L92-L93

Doing that here too instead of reassigning the URL could be an option.

lib/rich_text.rb
```diff
@@ -133,9 +131,12 @@ def auto_link(text, mode)
...
+            .wrap(Settings.linkify&.normalisation_rules)
+            .reduce(url) do |normalised_url, rule|
+        shorten_host(normalised_url, rule.hosts, rule.host_replacement) do |path|
+          path.sub(Regexp.new(rule.optional_path_prefix || ""), "")
+        end
```
hlfan:
```suggestion
            .reduce(url) { |url, rule| shorten_host(url, rule) }
```

```ruby
    def shorten_host(url, rule)
      %r{^(https?://([^/]*))(.*)$}.match(url) do |m|
        scheme_host, host, path = m.captures
        if rule.hosts&.include?(host)
          path = path.sub(Regexp.new(rule.optional_path_prefix || ""), "")
          if rule.host_replacement
            "#{rule.host_replacement}#{path}"
```

PR #5863, opened 2025-03-29
Add viewbox param to Nominatim link in search results

### LINKED ISSUE #5859: Link to Nominatim is missing viewbox parameter
### Problem

Follow-up to #3205. The new direct links to Nominatim from #4895 are awesome but it looks like the `viewbox=` parameter doesn't work as intended. It never seems to be included in the URL. You can already see in the screen shot of PR #4895 that it is missing.

### Description

_No response_

### Screenshots

_No response_

AntonKhorev:
Fixes #5859.

image

Also includes zoom value in reverse geocoding Nominatim link.

### DIFF
--- a/app/assets/javascripts/index/search.js
+++ b/app/assets/javascripts/index/search.js
@@ -18,9 +18,15 @@ OSM.Search = function (map) {
   $(".search_form").on("submit", function (e) {
     e.preventDefault();
     $("header").addClass("closed");
-    const query = $(this).find("input[name=query]").val();
-    let search = "/";
-    if (query) search = "/search?" + new URLSearchParams({ query });
+    const params = new URLSearchParams({
+      query: this.elements.query.value,
+      zoom: map.getZoom(),
+      minlon: map.getBounds().getWest(),
+      minlat: map.getBounds().getSouth(),
+      maxlon: map.getBounds().getEast(),
+      maxlat: map.getBounds().getNorth()
+    });
+    const search = params.get("query") ? `/search?${params}` : "/";
     OSM.router.route(search + OSM.formatHash(map));
   });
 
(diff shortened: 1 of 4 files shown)


app/controllers/geocoder_controller.rb
```diff
@@ -9,14 +9,17 @@ class GeocoderController < ApplicationController
   authorize_resource :class => false
 
   def search
-    @params = normalize_params
+    normalize_params
```
tomhughes, line 12: This could be a `before_action` now if we don't need to use the result?

app/assets/javascripts/index/search.js
```diff
@@ -18,9 +21,14 @@ OSM.Search = function (map) {
   $(".search_form").on("submit", function (e) {
     e.preventDefault();
     $("header").addClass("closed");
-    const query = $(this).find("input[name=query]").val();
-    let search = "/";
-    if (query) search = "/search?" + new URLSearchParams({ query });
+    const params = new URLSearchParams;
+    for (const paramName of ["query", "zoom", "minlon", "minlat", "maxlon", "maxlat"]) {
```
tomhughes, line 25: If we're going to build the URL in the submit handler why not just add the parameters from the bounding box here instead of adding hidden input fields and an event handler to keep them up to date?

app/controllers/geocoder_controller.rb
AntonKhorev: moved `normalize_params` to `before_action`

app/assets/javascripts/index/search.js
AntonKhorev: That was my attempt to remove `submit` listener. I thought about letting Turbo handling the entire thing. But since I didn't complete that, I'll get the zoom and bbox here.

tomhughes (approved):
Looks good to me now, thanks.

PR #5865, opened 2025-03-30
Propagate iD title to parent

### LINKED ISSUE #5864: Say "iD | OpenStreetMap" in browser tab when using iD
Here we see the right tab name shows that we are browsing a Node: 7..., via its '<TITLE>',
Image

So perhaps the left tab could say
- iD | OpenStreetMap

hlfan:
### Description
Fixes #5864

image
### How has this been tested?
With the dev tools

### DIFF
--- a/app/assets/javascripts/id.js
+++ b/app/assets/javascripts/id.js
@@ -76,5 +76,10 @@ document.addEventListener("DOMContentLoaded", function () {
       const data = parent.OSM.mapParams();
       goToLocation(data);
     });
+
+    const projectTitle = parent.document.title;
+    new MutationObserver(() =>
+      parent.document.title = [document.title, projectTitle].filter(t => t).join(" | ")
+    ).observe(document.querySelector("title"), { childList: true, subtree: true, characterData: true });
   }
 });
--- a/app/views/site/id.html.erb
+++ b/app/views/site/id.html.erb
@@ -6,6 +6,7 @@
   <!--[if !IE || gte IE 9]><!-->
   <%= javascript_include_tag "id" %>
   <!-- <![endif]-->
+  <title></title>
 </head>
 <body>
 <% data = {}


app/assets/javascripts/id.js
```diff
@@ -76,5 +76,13 @@ document.addEventListener("DOMContentLoaded", function () {
       const data = parent.OSM.mapParams();
       goToLocation(data);
     });
+
+    const firstParentTitle = parent.document.title;
+    new MutationObserver(function (mutations) {
+      if (!mutations.some(mutation => mutation.target.tagName === "TITLE")) return;
+      const newTitle = `${document.title} | ${firstParentTitle}`;
+      if (parent.document.title === newTitle) return;
+      parent.document.title = newTitle;
```
AntonKhorev, line 85:
```suggestion
      parent.document.title = `${document.title} | ${firstParentTitle}`;
```

app/assets/javascripts/id.js
```diff
@@ -76,5 +76,13 @@ document.addEventListener("DOMContentLoaded", function () {
...
+    }).observe(document.head, { childList: true, subtree: true });
```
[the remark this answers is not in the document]
AntonKhorev: You have to observe `document.head` because there's no title element initially. However if you `<title>` to `app/views/site/id.html.erb`, you'll be able to observe it right away. Even an empty title should work.

app/assets/javascripts/id.js
```diff
@@ -76,5 +76,10 @@ document.addEventListener("DOMContentLoaded", function () {
       const data = parent.OSM.mapParams();
       goToLocation(data);
     });
+
+    const projectTitle = parent.document.title;
+    new MutationObserver(() =>
+      parent.document.title = [document.title, projectTitle].filter(t => t).join(" | ")
+    ).observe(document.querySelector("title"), { childList: true, characterData: true });
```
AntonKhorev, line 83: I don't know why you removed `subtree` and how `characterData` is going to work without it.

app/views/site/id.html.erb
```diff
@@ -3,9 +3,8 @@
 <head>
   <meta charset='utf-8'>
   <%= stylesheet_link_tag "id" %>
-  <!--[if !IE || gte IE 9]><!-->
   <%= javascript_include_tag "id" %>
-  <!-- <![endif]-->
```
AntonKhorev, line 8: You don't need to make unrelated changes in this pull request.

app/assets/javascripts/id.js
hlfan: Took out `subtree` while moving to watch the title directly and forgot to put it back in.
_"Sometimes, my [confusion] is... It's almost frightening."_

AntonKhorev:
Merged, thanks.

PR #5866, opened 2025-03-30
Replace differently sized direction icon arrows with a marker

hlfan:
### Description
The way the directions icons are currently drawn, the diagonal arrows are a fair bit smaller than the arrows in cardinal directions. The size stays consistent when the arrows are instead drawn with a marker.
Compare the icons in this chart, where the old icon is red, the new one green, and the overlap yellow:
routing sprite icons
This requires the SVG element to be visible, so d-none was swapped.

### How has this been tested?
in-browser

### DIFF
--- a/app/views/directions/search.html.erb
+++ b/app/views/directions/search.html.erb
@@ -1,27 +1,30 @@
 <% content_for(:content_class) { "overlay-sidebar" } %>
 
-<svg width="20" height="20" class="d-none">
+<svg width="0" height="0" class="position-absolute">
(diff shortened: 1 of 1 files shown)


tomhughes:
Sorry @hlfan I merged this accidentally and have now reverted it but there doesn't seem to be any way to reopen the PR so if you want to pursue it can you reopen but you might want to read @AntonKhorev's comments in #5867 first.

AntonKhorev:
You want horizontal and vertical shape boundaries to go exactly over pixel boundaries, otherwise shapes may look more blurry and smaller.

PR #5867, opened 2025-03-30
Revert "Replace differently sized arrows with a marker"

AntonKhorev:
Reverts #5866. Before replacing "differently sized direction icon arrows" it would have been preferable to figure out why they are differently sized.

Arrows after #5866 on 1x pixel density display:
image

The original arrows this PR restores:
image


tomhughes:
Sorry yes I just realised I merged this by accident - it wasn't what I intended to merge.

PR #5868, opened 2025-03-31
Search and latlon/nominatim query resources

AntonKhorev:
Includes #5863.

Removes nonstandard action names from geocoder controller. Renames this controller to *search* because path is *search*, path helper is *search_path*, javascript controller is *search.js*.

Takes three search actions out of this controller and makes them nested *_query* resources:
- latlon_query
- <s>osm_nominatim_query</s> nominatim_query
- <s>osm_nominatim_reverse_query</s> nominatim_reverse_query


app/assets/javascripts/index/search.js
```diff
@@ -18,9 +21,14 @@ OSM.Search = function (map) {
...
-    let search = "/";
-    if (query) search = "/search?" + new URLSearchParams({ query });
+    const params = new URLSearchParams;
+    for (const paramName of ["query", "zoom", "minlon", "minlat", "maxlon", "maxlat"]) {
+      const paramValue = this.elements[paramName].value;
+      if (paramValue) {
+        params.set(paramName, paramValue);
+      }
+    }
+    const search = params.get("query") ? `/search?${params}` : "/";
```
hlfan, line 31: Do you really want to expose all these parameters in the URL?
The bbox params don't even update the map hash that's also there

tomhughes:
I wonder, do we need the `osm` prefix on the nomination searches? I can't remember what the history is but it probably dates to when we had a number of other search methods and even then I'm not sure it was really a good idea.

app/assets/javascripts/index/search.js
AntonKhorev: > Do you really want to expose all these parameters in the URL?

Yes. Search results depend on them. If you don't agree, why put anything in the URL? Why should `query` be in the URL?

AntonKhorev:
I removed the `osm_` prefix, but we're still calling it *OpenStreetMap Nominatim* in the ui:
image

Should we say just *Nominatim*?

AntonKhorev:
Removed `osm_` and *OpenStreetMap* from results headings too:
image

tomhughes:

PR #5869, opened 2025-04-01
Fix fixthemap note link

AntonKhorev:
Embed pages have links to the *fixthemap* page that have lat/lon/zoom params:
image

[Example link](https://www.openstreetmap.org/fixthemap?lat=53.25688279713156&lon=-1.4329004287719729&zoom=13)

*fixthemap* has an *Add a Note* section with a new note icon that is supposed to be a link:
image

But it's not a link because javascript writing it is broken. `zoom` is used as a string with `+` operator, concatenation happens instead of addition, and link generation fails.

This PR forces zoom to be a number inside `OSM.formatHash` which is used to generate a link.
image

Actually we don't need javascript to write new note links because lat/lon/zoom are available server-side, but I'm not changing that yet.

### DIFF
--- a/test/system/fixthemap_test.rb
+++ b/test/system/fixthemap_test.rb
@@ -1,0 +1,11 @@
+require "application_system_test_case"
+
+class FixthemapTest < ApplicationSystemTestCase
+  test "should have 'create a note' link with correct map hash" do
+    visit fixthemap_path(:lat => 60, :lon => 30, :zoom => 10)
+
+    within_content_body do
+      assert_link "Add a note to the map", :href => %r{/note/new#map=10/60(\.\d+)?/30(\.\d+)?}
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


hlfan:
> Actually we don't need javascript to write new note links because lat/lon/zoom are available server-side, but I'm not changing that yet.

I added that to the second commit of #5752 because I was altering the link building anyway.

AntonKhorev:
Removed the coordinate precision test in case they are added differently, like from rails code.

test/system/fixthemap_test.rb
```diff
@@ -0,0 +1,11 @@
+require "application_system_test_case"
+
+class FixthemapTest < ApplicationSystemTestCase
+  test "should have 'create a note' link with correct map hash" do
+    visit fixthemap_path(:lat => 60, :lon => 30, :zoom => 10)
+
+    within_content_body do
+      assert_link "Add a note to the map", :href => %r{/note/new#map=10/60\.\d+/30\.\d+}
```
hlfan, line 8:
```suggestion
      assert_link "Add a note to the map", :href => %r{/note/new#map=10/60(\.\d+)?/30(\.\d+)?}
```
There could also be no decimal places.

hlfan (approved):

test/system/fixthemap_test.rb
AntonKhorev: updated

PR #5870, opened 2025-04-01
Deduplicate note adding instructions

hlfan:
We have two versions of the guide to notes: one with the link as an icon and one with the link as plaintext in a barely different color. I dropped the second version.

### DIFF
--- a/app/views/site/welcome.html.erb
+++ b/app/views/site/welcome.html.erb
@@ -86,8 +86,6 @@
 </div>
 
 <div class='alert alert-secondary'>
-  <h2><%= t ".add_a_note.title" %></h2>
-  <p><%= t ".add_a_note.para_1" %></p>
-  <p><%= t ".add_a_note.para_2_html", :map_link => link_to(t(".add_a_note.the_map"), root_path),
-                                      :note_icon => tag.span(:class => "icon note bg-dark rounded-1") %></p>
+  <h2><%= t "site.add_a_note.title" %></h2>
+  <%= render "add_a_note" %>
 </div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2368,11 +2368,6 @@ en:
           explanation_html: |
             If you have noticed a problem with our map data, for example a road is missing or your address, the best way to
             proceed is to join the OpenStreetMap community and add or repair the data yourself.
-        add_a_note:
-          instructions_1_html: |
-            Just click %{note_icon} or the same icon on the map display.
-            This will add a marker to the map, which you can move
-            by dragging. Add your message, then click save, and other mappers will investigate.
       other_concerns:
         title: Other concerns
         concerns_html: |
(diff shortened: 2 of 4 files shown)


app/views/site/welcome.html.erb
```diff
@@ -86,8 +86,7 @@
 </div>
 
 <div class='alert alert-primary'>
-  <h2><%= t ".add_a_note.title" %></h2>
-  <p><%= t ".add_a_note.para_1" %></p>
-  <p><%= t ".add_a_note.para_2_html", :map_link => link_to(t(".add_a_note.the_map"), root_path),
-                                      :note_icon => tag.span(:class => "icon note bg-dark rounded-1") %></p>
+  <h2><%= t "notes.guide.title" %></h2>
+  <p><%= t "notes.guide.para_1" %></p>
+  <p><%= t "notes.guide.para_2", :note_icon => tag.a(:href => new_note_path, :class => "icon note bg-dark rounded-1") %></p>
```
nenad-vujicic, line 91: Here is how this renders on my side:

image

hlfan: Ok, I added back the `_html` suffix, but where did the red outline come from?

HolgerJeromin: >  but where did the red outline come from?

I would guess from the screenshot tool of nenad? :-)

nenad-vujicic: > Ok, I added back the `_html` suffix, but where did the red outline come from?

Thanks, now it works great on my side! Yes, I added red outline when prepared screenshot, sorry for confusion.

AntonKhorev:
There's already one deduplicated section `<%= render "any_questions" %>`. You could have done the same with *Add a Note!*, except maybe without including `<h2>`/`<h3>` in the template because they're slightly different. In this case you wouldn't need to come up with i18n scope names and use the template defaults.

app/views/site/welcome.html.erb
```diff
@@ -86,8 +86,6 @@
 </div>
 
 <div class='alert alert-primary'>
-  <h2><%= t ".add_a_note.title" %></h2>
-  <p><%= t ".add_a_note.para_1" %></p>
-  <p><%= t ".add_a_note.para_2_html", :map_link => link_to(t(".add_a_note.the_map"), root_path),
-                                      :note_icon => tag.span(:class => "icon note bg-dark rounded-1") %></p>
+  <h2><%= t "site.add_a_note.title" %></h2>
+  <% render "add_a_note" %>
```
AntonKhorev, line 90:
```suggestion
  <%= render "add_a_note" %>
```

config/locales/en.yml
```diff
@@ -2481,16 +2476,15 @@ en:
...
-        the_map: the map
+    add_a_note:
+      title: No Time To Edit? Add a Note!
+      para_1: |
+        If you just want something small fixed and don't have the time to sign up and learn how to edit, it's
+        easy to add a note.
+      para_2_html: |
+        Just click %{note_icon} or the same icon on the map display.
+        This will add a marker to the map, which you can move
+        by dragging. Add your message, then click save, and other mappers will investigate.
```
AntonKhorev, line 2487: Maybe we still need the map link too. If I don't know anything about osm, arrive at the welcome page and read this, I might have a question: where's the map display?

```suggestion
      para_2_html: |
        Just click %{note_icon} or the same icon on %{map_link}.
        This will add a marker to the map, which you can move
        by dragging. Add your message, then click save, and other mappers will investigate.
      the_map_display: the map display
```
\+ add `:map_link` from the template

Adding a note used not to work at all on low zooms, that's probably why that link existed only on fixthemap page. But it still may be more intuitive to open the map first, find the place, and after that click *add a note*, so the map link is still useful.

hlfan: I don't think an extra link would help if it's as barely distinguishable as the current link to the_map. Do you have any plan to improve this?
In welcome, _the map_ is mentioned five times before this section, why should it be linked that time? That's why I didn't choose that translation.

AntonKhorev: > barely distinguishable

only in dark mode

> Do you have any plan to improve this?

Change alert color to secondary:

image image

> In welcome, the map is mentioned five times before this section, why should it be linked that time? 

Because here it is a "call to action" while previously it's not.

hlfan: Then it might not be a good idea to merge the translations after all.

tomhughes:

PR #5872, opened 2025-04-01
Use L.OSM.locate control in user.js

AntonKhorev:
There's a claim in #5752 that "Due to the nature of the leaflet locate plugin, that icon is done with ..." and then various approaches are used such as clip path and polygon in css instead of a regular svg. While true, this is easily fixable inside our `L.OSM.locate` control. But turns out that we don't use our own `L.OSM.locate` in `user.js`. That's because I wasn't paying attention to this file when doing https://github.com/openstreetmap/openstreetmap-website/commit/12a7c9d023bcca01657e357b2ed28cc6bcd20e06.

This PR makes `user.js` use `L.OSM.locate`.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -1,7 +1,6 @@
 //= require_self
 //= require leaflet.sidebar
 //= require leaflet.sidebar-pane
-//= require leaflet.locatecontrol/dist/L.Control.Locate.umd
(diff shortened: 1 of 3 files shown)


hlfan (approved):

hlfan (approved):

tomhughes:
Looks good to me, thanks.

PR #5876, opened 2025-04-02
History page links

AntonKhorev:
Changes *Load more* buttons into links similar to those used for pagination.
Changes their text to *Newer/Older Changesets*, see https://github.com/openstreetmap/openstreetmap-website/pull/5812#issuecomment-2735280265.

Now these links can be copied and opened like any other links.

image

For example, you can middle-click a link to open the next page in a new tab.

image


gravitystorm:
Looks good! Thanks.

PR #5563, opened 2025-01-27
DataLayer to consume JSON format responses

mmd-osm:
https://github.com/openstreetmap/leaflet-osm/pull/44 introduced OSM JSON support in the underlying leaflet osm package. This PR enables the new feature by requesting application/json format (via Accept header).

Issue #2800, opened 2020-08-28
Map Data checkmark is unresponsive


martinstingl:
When I enable "Map Data", the checkmark is NOT set until the map data has loaded, which takes some time (depending on how much data is fetched).

Please allow the checkmark to be immediately set and unset by clicking with the mouse, independent of the data load state. 
The loading process of the data should then just "look" after the checkmark state.

tomhughes:
What browser are you using? Only the check mark appears immediately for me and the data then loads in the background.

If your browser isn't doing that then it may not be something we can fix, as the check mark is set and unset by the browser and we just watch for the change event to trigger loading the data so if your browser chooses not to set the check mark until after the event handler has been fired that might not be fixable.

dieterdreist:
I was also experiencing the unresponsive checkmark in mobile Safari but it
is now fixed for me (just checked, iOS 13.6.1)

martinstingl:
I am using Firefox 80.0 with Linux Mint 20 Xfce.

tomhughes:
Well I was also using Firefox 80 on linux and wasn't seeing the problem.

martinstingl:
Even if there is a kind of event handler working, something prevents my browser from checking the checkmark first and then fetching the data. 
Currently my browser marks not until loading and rendering the map data is finished.

The same behaviour is there when deselecting the box: only after redrawing the map, the box is unchecked.

jidanni:
The map data check mark is extremely dangerous. Sure you might enable it while you're looking at a forest but the moment you slide it into some data-dense area your cellphone browser is probably going to be out of business and there's no way you're going to get back into that menu to turn off.

AntonKhorev:
Currently it takes a noticeable time (like a second) for the checkbox to update if there's a lot of elements to display (over 10000). It doesn't wait for loading over the network, the wait is for Leaflet to add the layer.

AntonKhorev:
There is a pull request that addresses specifically adding layers: #5009. However it schedules layers with `setTimeout` without any possibility to cancel the process.

nertc:
Last update of the PR #5009 adds functionality for canceling layer loading/unloading. Currently code in that PR works like this:
```
When checkmark is checked:
    If unloading is in progress, cancel unloading
    Run asynchronous layer loading
When checkmark is removed:
    If loading is in progress. cancel loading
    Run asynchronous layer unloading
```

[...]

PR #5877, opened 2025-04-02
Disabling asynchronous Map data layer rendering

deevroman:
Full discussion here: https://github.com/openstreetmap/openstreetmap-website/pull/5009#issuecomment-2722980833 But I'll duplicate the comment here.

https://github.com/openstreetmap/openstreetmap-website/pull/5009 brought two problems:

1. When moving the map, a white flash occurs and all objects on the map are redrawn:

https://github.com/user-attachments/assets/3c87c769-174c-4513-a5be-115eb9517bb8

2. Rendering slowed down. What used to occupy ~400ms now occupies ~1.3s. Hiding a data layer was previously performed on ~100ms, now ~1.5s (!)

Both problems arise due to the fact that before that the redrawing of the map was performed only once, after the completion of the removal/adding of objects to the SVG-element of the map.

--

This PR only removes asynchronous rendering. The loader that shows downloading data near the checkbox remains.

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -43,7 +43,7 @@ L.OSM.Map = L.Map.extend({
     this.noteLayer = new L.FeatureGroup();
     this.noteLayer.options = { code: "N" };
 
-    this.dataLayer = new L.OSM.DataLayer(null, { asynchronous: true });
+    this.dataLayer = new L.OSM.DataLayer(null);
     this.dataLayer.options.code = "D";
 
     this.gpsLayer = new L.OSM.GPS({


nenad-vujicic:
I also succeeded to reproduce flickering by switching between e.g. point [A](https://www.openstreetmap.org/#map=18/59.924360/30.304869&layers=ND) and point [B](https://www.openstreetmap.org/#map=18/59.925492/30.305593&layers=ND), although I have high-performance machine (by e.g. pressing Back and Forward).

Wouldn't be better approach to replace [clearLayers](https://github.com/openstreetmap/openstreetmap-website/blob/2075eb8329e5e011e8acaa90ade8229f6c9ea88b/app/assets/javascripts/index/layers/data.js#L112) + [rebuilding data layer](https://github.com/openstreetmap/openstreetmap-website/blob/2075eb8329e5e011e8acaa90ade8229f6c9ea88b/app/assets/javascripts/index/layers/data.js#L80) (which is actually causing the flickering, if I understood it well) with `remove out-of-viewport features and add new-features` logic?

deevroman:
the flicker causes: clearLayers + event loop steps + rebuilding data layer. Perhaps you could first add new objects to the map and then delete the old ones. However, it would probably still be noticeable due to the transparency of the lines.

But in general, I don’t understand why all these complications are there, when everything worked fine before the appearance of asynchrony.

tomhughes:
Well https://github.com/openstreetmap/leaflet-osm/pull/38 explains the reasons for asynchronous operation - are you saying that the reasons there are somehow incorrect?

deevroman:
> are you saying that the reasons there are somehow incorrect?

Yes, adding asynchrony was a treatment for the symptoms. Hangups were resolved in the following PRs:

https://github.com/openstreetmap/leaflet-osm/pull/43 (freezing before a warning about 2000+ elements)
https://github.com/openstreetmap/openstreetmap-website/pull/5563 (XML -> JSON)
https://github.com/openstreetmap/openstreetmap-website/pull/5474 (freezing after consent to rendering)

After that, work resumed on the https://github.com/openstreetmap/openstreetmap-website/pull/5009 and with the addition of the loader, asynchrony was enabled.

AntonKhorev:
Flickering:

https://github.com/openstreetmap/openstreetmap-website/blob/5441fa79dba55f6d694649c69757fbd56aa2c566/app/assets/javascripts/index/layers/data.js#L112

I'm not sure how to remove flickering if we're first clearing everything (ignoring async option by the way), then scheduling async layer additions.

> Freezes were resolved in the following PRs: ...

This one wasn't: https://github.com/openstreetmap/openstreetmap-website/issues/2800#issuecomment-2621712294

deevroman:
> This one wasn't: https://github.com/openstreetmap/openstreetmap-website/issues/2800#issuecomment-2621712294

Don't forget that for 4000+ elements, a warning is shown before rendering. Only after you click on it will the browser be blocked.

tomhughes:
I'm a bit worried that we're going round in circles here but I'll merge this for now... Hopefully somebody can work on better ways to do updates in the future.

PR #5879, opened 2025-04-02
Align globe to the end of the container.

gravitystorm:
This removes the arbitrary pixel-based alignment of the globe, which did not take tab label translations into account. Instead, we give as much space as possible to the tabs, by aligning to the end of the container (right-hand side on LTR, left hand side on RTL).

Refs #4773

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -727,7 +727,7 @@ tr.turn {
 
   &.new-user-main {
     background-image: image-url("sign-up-illustration.svg");
-    background-position-x: 70px;
+    background-position-x: 100%;
   }
 
   &.confirm-main {


app/assets/stylesheets/common.scss
```diff
@@ -727,7 +727,7 @@ tr.turn {
 
   &.new-user-main {
     background-image: image-url("sign-up-illustration.svg");
-    background-position-x: 70px;
+    background-position-x: 100%;
```
hlfan, line 730:
```suggestion
```
`background-position` is already set to `right` in `.header-illustration`.

hlfan (changes requested):

app/assets/stylesheets/common.scss
hlfan: But since we flip the x-axis it needs an `/*rtl:ignore*/`

AntonKhorev:
If it was that easy...

Here's how it's going to look on small screens:
image

hlfan:
```css
background-position: bottom right;
```
And crop away the empty space on the left.
Then:
ezgif-2d8920a03b79b6

gravitystorm:
Closing in favour of #5894

PR #5881, opened 2025-04-03
Don't set bbox history param when the entire world is visible

AntonKhorev:
The regular `/history` page shows changesets intersecting with the map view. To get them, the current map view bounds are sent as the `bbox` parameter. But the bound can be outside of [-180..180] longitude range, therefore also a `.wrap()` fix is applied. The fix itself is broken, see #5473 which removes it and implements support for longitudes outside of [-180..180].

Still, we don't need to send the `bbox` parameter with huge longitude values when the entire planet is visible.[^1] If there's no `bbox`, no filtering by `bbox` happens and all changesets are visible. This is what this PR does.

This is a partial fix for #3423 when the entire planet is visible. #5473 is still required for map views with the antimeridian that don't include the entire planet. But even without #5473 you'll be able to fully zoom out, click *History* and see all the latest changesets.

image

[^1]: Strictly speaking, that never happens because poles are infinitely away, but we can consider them visible if ~[-85..85] latitude range is visible.

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -213,7 +213,15 @@ OSM.History = function (map) {
   }
 
   function setBboxFetchData(data) {
-    data.set("bbox", map.getBounds().wrap().toBBoxString());
+    const crs = map.options.crs;
+    const sw = map.getBounds().getSouthWest();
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #5882, opened 2025-04-03
Enable native share if available

AntonKhorev:
I wanted to check if [navigator.share](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share) works. This PR adds a hidden button for native share and unhides it if native share is available. On desktops it may not be available but it worked for me in Chrome on an Android phone.[^1]

I'm not sure if it's very useful because you can just share the current page url from the browser menu.

image

[^1]: The server needs to run either on https or localhost, so you'll likely need to enable port forwarding on chrome://inspect/#devices to test this locally.


HolgerJeromin (approved):

tomhughes:
Can confirm it works nicely in Firefox on Android but not on desktop - there is an option to enabled it on desktop but (at least on linux) nothing actually happens when you click it and the console shows that the promise is resolving an "Operation not supported" exception.

1ec5:
> I'm not sure if it's very useful because you can just share the current page url from the browser menu.

In Safari on macOS, a share sheet appears with a variety of options, including shortcuts to people I’ve chatted with recently and actions provided by third-party extensions. Ultimately it’s based on a URL, but it can save users some steps.

kcne (approved):

tomhughes:
> Firefox doesn’t officially support this API yet; you have to toggle `dom.webshare.enabled` in about:config first.

Yes I know - that's what I meant about enabling it. But even when enabled it doesn't actually do anything, at least on linux.

I don't think that really matters though.

tomhughes:
MDN claims that Chrome supports it from 129 but it doesn't seem to on linux at least - possibly Windows only or something. Mobile works fine though which is the main use case anyway.

tomhughes:
I don't see any issues with this, so I think it's good to merge, thanks.

PR #5884, opened 2025-04-03
Simplify changeset icon definitions

hlfan:
Makes the changeset icon definitions a bit more compact and doesn't redefine them in every single history entry.


nenad-vujicic (changes requested):
This works and looks great, but wouldn't be better practice to put definitions of `icon-comment` and `icon-pencil` at the beginning of `_changeset` (before `<%= tag.li ...`) so we don't have to redefine if use `_changeset` on some other place?

nenad-vujicic (approved):
Thanks, now it looks great!

tomhughes:
I've mostly been ignoring the SVG PRs as I don't really know much about SVG and there was a lot of debate between you, @AntonKhorev and @gravitystorm about what was the best approach.

hlfan:
Since the merge of @AntonKhorev's #6155 the information about how often the icon was used before has been disjointed enough that defining the paths just once becomes unfeasible. However, the definitions can still be simplified.

tomhughes:
Looks good to me, thanks.

PR #5885, opened 2025-04-04
Remove manual edit after 4-second timeout on welcome page

AntonKhorev:
Go to `/welcome` and click *start mapping*:

image

What's going to happen is that if you didn't give the geolocation permission yet:
1. the permission dialog pops up
2. if you fail to react to it in four seconds, you get redirected to the map view

Four seconds is probably too short. And why do you need any timeout here? Supposedly this was a workaround for [a Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=675533) but it was fixed years ago.

This PR removes the timeout.

### DIFF
--- a/app/assets/javascripts/welcome.js
+++ b/app/assets/javascripts/welcome.js
@@ -15,10 +15,6 @@ $(function () {
       $(".start-mapping").addClass("loading");
 
       if (navigator.geolocation) {
-        // handle firefox's weird implementation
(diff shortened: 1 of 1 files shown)


HolgerJeromin (approved):

hlfan (approved):

tomhughes:
Looks good to me, thanks.

PR #5887, opened 2025-04-04
Clean OSM.formatHash calls

hlfan:
There is not just a `formatHash` call in `fixthemap` but also one in `welcome` that can be written in the template.
Looked through all `formatHash` calls and cleaned up another call.
Builds on the commit from #5752 that fixes gravitystorm#257.
See also #5869.

### DIFF
--- a/app/helpers/geocoder_helper.rb
+++ b/app/helpers/geocoder_helper.rb
@@ -7,7 +7,7 @@ module GeocoderHelper
           elsif result[:min_lon] && result[:min_lat] && result[:max_lon] && result[:max_lat]
             "/?bbox=#{result[:min_lon]},#{result[:min_lat]},#{result[:max_lon]},#{result[:max_lat]}"
           else
-            "/#map=#{result[:zoom]}/#{result[:lat]}/#{result[:lon]}"
+            "/##{map_hash(result)}"
           end
 
     result.each do |key, value|
@@ -21,6 +21,12 @@ module GeocoderHelper
     html << " " if result[:suffix] && result[:name]
     html << result[:suffix] if result[:suffix]
     safe_join(html)
+  end
+
+  def map_hash(params)
+    return nil unless params[:lat].present? && params[:lon].present?
+
+    "map=#{params[:zoom] || 17}/#{params[:lat]}/#{params[:lon]}"
   end
 
   def describe_location(lat, lon, zoom = nil, language = nil)
--- a/app/views/site/welcome.html.erb
+++ b/app/views/site/welcome.html.erb
@@ -81,7 +81,7 @@
 <% if params[:oauth_return_url] %>
   <a class="btn btn-primary" href="<%= params[:oauth_return_url] %>"><%= t ".continue_authorization" %></a>
 <% else %>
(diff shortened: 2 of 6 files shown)


app/helpers/geocoder_helper.rb
```diff
@@ -23,6 +23,12 @@ def result_to_html(result)
     safe_join(html)
   end
 
+  def map_hash(params)
+    return "" unless params[:lat].present? && params[:lon].present?
+
+    "#map=#{params[:zoom].to_i || 17}/#{params[:lat].to_f}/#{params[:lon].to_f}"
```
tomhughes, line 29: Why do we need to convert lat and lon to floating point and then immediately back to string again?

app/views/site/welcome.html.erb
```diff
@@ -81,7 +81,8 @@
 <% if params[:oauth_return_url] %>
   <a class="btn btn-primary" href="<%= params[:oauth_return_url] %>"><%= t ".continue_authorization" %></a>
 <% else %>
-  <a class="btn btn-primary start-mapping" href="<%= edit_path %>"><%= t ".start_mapping" %></a>
+  <%= editor_query = "?editor=#{params[:editor].to_s.split(/\W/)[0]}" if params[:editor].present?
```
tomhughes, line 84: Why do we need to split the editor into words and then just take the first one? The javascript implementation that this is replacing doesn't seem to do that?

app/helpers/geocoder_helper.rb
hlfan: Brakeman complained

app/views/site/welcome.html.erb
hlfan: Brakeman complained, again

app/helpers/geocoder_helper.rb
tomhughes: I just tried removing them and I couldn't get brakeman to complain... What exactly did it say?

app/views/site/welcome.html.erb
tomhughes: Same here, I couldn't reproduce any complaint.

app/helpers/geocoder_helper.rb
hlfan: Interesting - I can't either now... but I remember some error before I made this PR. But I'm not complaining.

app/helpers/geocoder_helper.rb
```diff
@@ -23,6 +23,12 @@ def result_to_html(result)
     safe_join(html)
   end
 
+  def map_hash(params)
+    return "" unless params[:lat].present? && params[:lon].present?
+
+    "#map=#{params[:zoom] || 17}/#{params[:lat]}/#{params[:lon]}"
```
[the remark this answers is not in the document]
hlfan: It needs the `map=` to work, though?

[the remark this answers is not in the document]
tomhughes: Sorry yes you're right so maybe string interpolation is easier for that...

tomhughes (approved):

PR #5888, opened 2025-04-04
Remove .clearfix from places without floats

AntonKhorev:
We don't need [Bootstrap's clearfix class](https://getbootstrap.com/docs/5.3/helpers/clearfix/) because we're not using css floats anywhere.[^1] That includes user-provided richtext because we're removing `style` and `class` attributes.

[^1]: except on the *Export* page, see the next PR.


tomhughes:
Looks good to me, thanks.

PR #5889, opened 2025-04-04
Use flex+justify instead of floats on export page

AntonKhorev:
In the export box on the `/export` page we're positioning inputs using floats. Then we have to guard them with `/*rtl:ignore*/` in css. But actually we don't need any custom css here because we set `dir='ltr'` on the entire box. If we don't use any styles that mention left/right/start/end, `dir` will work as expected. Instead of left/right floats, I'm using flex with justify-content-between to push the inputs to the edges of the box.

image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -674,12 +674,6 @@ tr.turn {
     > * {
         margin: -1px;
     }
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #5892, opened 2025-04-05
Remove invalid setStyle from data layer

AntonKhorev:
Look at data layer. Notice that everything is blue:
image

Now look at the `setStyle` call that I remove in this PR. Notice that the colors are supposed to be black, red, green. It's broken since https://github.com/openstreetmap/openstreetmap-website/commit/5e9ab5bc5edb63b7ab7719fa989a18f52e3637a2, but I'm not fixing it because those colors don't look very well over the tiles and everyone probably got used to blue.

The argument to `setStyle` is invalid; and when used on feature group layer, `setStyle` should be called after adding the geometry.
https://leafletjs.com/reference.html#featuregroup-setstyle

### DIFF
--- a/app/assets/javascripts/index/layers/data.js
+++ b/app/assets/javascripts/index/layers/data.js
@@ -2,21 +2,6 @@ OSM.initializeDataLayer = function (map) {
   let dataLoader, loadedBounds;
   const dataLayer = map.dataLayer;
 
-  dataLayer.setStyle({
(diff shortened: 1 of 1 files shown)


tomhughes:
I don't actually remember the data layer being anything other than blue but it does look like the previous, presumably working, styling existed before #328 broke it.

Looks good to me anyway.

PR #5894, opened 2025-04-05
Optimize user creation illustrations by reusing elements

hlfan:
Reducing the size of the previously _minified_ illustrations by 88.733% (123.59kB to 13.93kB) by reusing elements in the SVGs and in between the SVGs.


AntonKhorev:
1\. you're not saving much
sign-up-illustration.svg
Transferred: 5.55 kB

2\. you already know about this issue from #5879
image

hlfan:
> you're not saving much

5195 + 5865 + 4015 bytes vs 755 + 1941 + 1270 + 238 + 915
so 5119 B instead 15075 B gzipped, so still less than a third.

> you already know about this issue

yes and my ideas escalated from there.

AntonKhorev:
> 5195 + 5865 + 4015 bytes

There's no way anyone is going to see all three images.

hlfan:
Has the sign-up flow changed so users don't see all images when creating a new account?

hlfan:
@AntonKhorev should I only remove the file splitting or more?

hlfan:
@gravitystorm does this close #5879 for you?

hlfan:
Incredibly, #6797 only touched another `header-illustration new-user-...` class definition.

gravitystorm:
> @gravitystorm does this close #5879 for you?

Yes, looks like a better approach.

hlfan:
> I think the same change needs to be made for the other locations.

I adjusted `users/new`. The ones I haven't changed only include a heading and don't seem to cause any overlap.

gravitystorm (approved):

Issue #5539, opened 2025-01-23
Refactoring Controllers to use Resourceful Actions


gravitystorm:
For a few years we've been slowly refactoring our controllers to use "Resourceful Actions". I think it is worth having an issue here to explain the process, so that other people know what's going on, and how they can help.

In Rails you can name your controller actions whatever you like. For example, we have `go_public` in UsersController and `subscribe` in ChangesetsController. Resourceful actions (sometimes known as "restful" actions, there's no strict terminology here) are where a controller only has actions from the following list:

* index
* show
* new
* create
* edit
* update
* destroy

This helps keep our code organised, and therefore easier to maintain. We can be clear about which actions need database write access (create, update, destroy), we know which actions will need a view template and which will iterate over a list. This refactoring allows us to use more [resourceful routing](https://guides.rubyonrails.org/routing.html#resource-routing-the-rails-default) with fewer customisations.

When a controller has additional methods, that usually indicates that we need (at least) one more controller. The first step is to consider what kind of action you are looking at:

* Is it showing a list of items? It's probably an `index` action
* Is it showing just one item? It's probably a `show` action
* Is it creating something new? It's probably a `create` action
* etc

The tricky part is trying to find the right name for the new controller. Here's my guidance, perhaps other people can help clarify further.

First, controllers should usually be named after nouns. The actions are verbs, and those verbs should apply to a noun. So for example, with ChangesetsController#subscribe, then instead of using the verb "subscribe" you could use the verb "create" on the noun "subscription". (An alternative might be to "create" a "subscriber", which is also a noun (and changesets do have "subscribers"), but we aren't creating the user here so "creating" a "subscriber" isn't quite accurate).

Secondly, you will likely then up with a chain of multiple nouns. So in our example, we have a "subscription" noun and also the "changeset" noun, since we are creating a subscription to the changeset. In general, if one of the nouns is the same as a model, it will be the name of the module. So in `app/controllers/changesets/subscriptions_controller.rb` we have 

```
module Changesets
   class SubscriptionsController < ApplicationController
```

(The alternative here would be to have a subscriptions module with various controllers named after our models inside that, which could work, but is harder to navigate for developers).

[...]

PR #5895, opened 2025-04-05
Draft: Refactor confirmations controller

rkoeze:
### Description
Draft PR for a restful refactor of the confirmations controller (issue https://github.com/openstreetmap/openstreetmap-website/issues/5539). The original intention here was for a purely structural set of changes that otherwise preserved existing functionality. However, moving toward a more restful structure prompted reusing the `#show` endpoint for both confirmation scenarios, which in turn introduced a few changes in behavior. 

I am looking for some initial feedback that this is going in an acceptable direction. I am of course happy to break this out into smaller PRs. 

Changes:
- Introduce restful routes for confirmations controller
- Deprecate `new_user` and `new_email` tokens in favor of single `account_confirmation` token
- Some UI changes around confirmation and error messages

Remaining work:
- [ ] Fix failing tests
- [ ] Add more test coverage

### How has this been tested?
I've gone through both of the new account and email change confirmation flows locally. I've also tested out the scenarios where confirmation emails are sent out with the currently deployed token generation logic but are handled by the code in this PR.


app/controllers/confirmations_controller.rb
```diff
@@ -10,113 +10,74 @@ class ConfirmationsController < ApplicationController
...
+  before_action :require_cookies, :only => [:show]
 
-  def confirm
-    if request.post?
-      user = User.find_by_token_for(:new_user, params[:confirm_string])
+  def show
+    if params[:confirm_string]
+      user = User.lookup_by_confirmation_token(params[:confirm_string])
 
       if !user
```
rkoeze, line 20: This was one of the bigger changes in terms of UX. Now if a user can't be found or their account has already been confirmed, they are redirected to the root path in both the new account and email change cases.

app/models/user.rb
```diff
@@ -445,6 +451,55 @@ def deletion_allowed?
     deletion_allowed_at <= Time.now.utc
   end
 
+  # Gracefully fallback to looking up users by deprecated new_user and new_email tokens
+  def self.lookup_by_confirmation_token(account_confirmation_token)
```
rkoeze, line 455: The idea here is that we will support the `new_user` and `new_email` tokens issued to users who have signed up or changed their emails for some period of time. At that point this code and the older routes can be removed.

rkoeze:
Closing this as it's a fairly large PR and I think at least some of these changes can be broken out into smaller PRs.

PR #5896, opened 2025-04-05
Show valid username in header when renaming

AntonKhorev:
When trying to rename a user account to an invalid name, the invalid name is shown in the upper right corner as if the rename already happened:
image

This PR shows the name before the invalid change:
image

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -75,7 +75,7 @@
             <span class="badge count-number position-static m-1"><%= current_user.new_messages.size %></span>
           <% end %>
           <span class='username align-middle text-truncate' dir='auto'>
-            <%= current_user.display_name %>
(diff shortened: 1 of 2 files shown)


rkoeze:
Nice find!  I reproduced the error on main and then confirmed the fix running this branch.

I wonder if it might make sense to rename the `display_name` field in the test to  `Valid Display Name`. That way if the test fails to find the valid username you get an error that more clearly suggests the element that could not be found. So instead of:
```
expected to find button "Valid User" that is not disabled but there were no matches
```
You get:

```
expected to find button "Valid Display Name" that is not disabled but there were no matches
```
Feel free to ignore, just a suggestion.

tomhughes:
If the test fails then the first thing you're going to do is look at the test and I think it's pretty obvious there what's going on so I think this is probably fine.

PR #5897, opened 2025-04-06
Prefer COPY to ADD in Dockerfile

rkoeze:
### Description
Updates the Dockerfile to use `COPY` instead of `ADD`. From [Docker's best practices doc](https://docs.docker.com/build/building/best-practices/#add-or-copy):

> You'll mostly want to use COPY for copying files from one stage to another in a multi-stage build.

and:

> The ADD instruction is best for when you need to download a remote artifact as part of your build.

Since we're just copying files, let's prefer `COPY`.

### How has this been tested?
I rebuilt my docker container with these new changes and was able to successfully navigate the app locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,10 +42,10 @@ RUN mkdir -p /app
 WORKDIR /app
 
 # Install Ruby packages
-ADD Gemfile Gemfile.lock /app/
+COPY Gemfile Gemfile.lock /app/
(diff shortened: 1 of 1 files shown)


hlfan (approved):

tomhughes:
Looks good to me, thanks. The test failures are related to coveralls and can be ignored.

PR #5898, opened 2025-04-06
Move directions step numbering out of engine-specific code

AntonKhorev:
Writing step numbers is currently the responsibility of each routing engine:

image

But that shouldn't be the case. Writing the numbers outside would later help to stop constructing html by concatenation.


tomhughes:
Looks good to me, thanks.

hlfan:
Did you intentionally remove the instruction indicies from the popup?

PR #5900, opened 2025-04-06
Avoid virtual keyboard when visiting new note page

AntonKhorev:
Another autofocus-related change in addition to #5899.

Normally on map view pages we have the search input autofocused. But on the *Add a note* page it makes sense to focus on the note description input. Since it may be loaded in the sidebar after the regular page load, we're focusing with javascript instead of a html attribute.

Now try adding a note on a phone with a small screen. What's normally a sidebar now takes up the upper half of the screen. You get the lower half to see the map and to place the note marker. Except not really because the virtual keyboard shows up as soon as the description input gains focus, that is right away:

Screenshot_20250406-144921

This PR adds [a workaround](https://gist.github.com/supertanuki/1f73f9aa33d5010bf07205804596e042) to suppress the initial focus from triggering the virtual keyboard. The proper API for that [is still experimental](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API).

### DIFF
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -130,7 +130,9 @@ OSM.NewNote = function (map) {
 
     content.find("textarea")
       .on("input", updateControls)
-      .trigger("focus");
+      .attr("readonly", "readonly") // avoid virtual keyboard popping up on focus
+      .trigger("focus")
+      .removeAttr("readonly");
 
     content.find("input[type=submit]").on("click", function (e) {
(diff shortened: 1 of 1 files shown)


tomhughes:
This doesn't seem to work in Firefox on Android but it does in Chrome, and presumably in Safari on iOS, so it looks good to me.

AntonKhorev:
Even if I delay removing `readonly`, Firefox waits for it and as soon as I remove this attribute, virtual keyboard pops up.

PR #5901, opened 2025-04-06
More directions refactoring

hlfan:
### Description
I took another look at the directions engines after #5898. I may have found too much, but I kept the commit size low this time.

### How has this been tested?
in chrome and firefox

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -184,7 +184,7 @@ OSM.Directions = function (map) {
         );
 
       // Add each row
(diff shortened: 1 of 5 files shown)


tomhughes:
I think the only thing I might question is the removal of multi-leg support but I guess without waypoints the engines are only going to return one leg and in general this looks good to me, thanks.

hlfan:
GraphHopper never had multi-leg support, and it's not like we're on the verge of adding it.

AntonKhorev:
I'm going to undo things like https://github.com/openstreetmap/openstreetmap-website/pull/5901/commits/2af72b2e6f4e4c1c0775f16f102f50a59bdeb0c3.

PR #5902, opened 2025-04-07
Add more files to `.dockerignore`

rkoeze:
### Description
Adds some more files to the `.dockerignore`. In general this brings up closer to the [Rails defaults](https://github.com/rails/rails/blob/1dc50194047b6a371dd78b486a247babfaa9f0e0/railties/lib/rails/generators/rails/app/templates/dockerignore.tt).

### How has this been tested?
I added `COPY .  .` to the Dockerfile and then ran `docker build -t web . && docker run --rm -it -p 3000:3000 web sh` with and without these changes to make sure that the correct files were ignored when the image was built.


tomhughes (changes requested):
The third commit should be squashed into the first one rather than keeping it as a fixup commit.

rkoeze:
@tomhughes sorry about that, and thanks for the reminder

tomhughes (approved):
Thanks for the fix - this looks good now.

PR #5903, opened 2025-04-07
Render changesets on both sides of the antimeridian

AntonKhorev:
Continues #5473 by fixing the rendering.

When anything is added to the map, its longitude could be as well (the original longitude) + n * 360. n is an integer and usually it's 0 but that doesn't work when looking at the map near the antimeridian. Here I'm picking such n for each changeset bbox that it's going to be at the closest location to the map view center.

image


tomhughes:

Issue #5905, opened 2025-04-08
Hide future days in user heatmap


HolgerJeromin:
### Problem

![Image](https://github.com/user-attachments/assets/baaee4f2-cf21-4c55-9fe8-5a7456f4dbe5)
Probably solvable:
https://github.com/wa0x6e/cal-heatmap/issues/513

PR #5906, opened 2025-04-08
Refactor Heatmap: Manual Labels & Yearly Template

kcne:
### Description
Closing #5905 . This PR improves the user profile heatmap by adding a `yearly` template to display exact one year starting from todays date respectively and `CalendarLabel` plugin for better month labeling. The changes include:

- Adding a yearly template to show exactly one year of data
- Integrating the CalendarLabel plugin for proper month label display since there is no way to do it automatically when creating new template according to official [docs](https://cal-heatmap.com/).
- Handling the special case where months start after the 15th by rotating them to the end

### How has this been tested?
- Tested the heatmap display with different start dates to verify month label rotation
- Verified tooltips appear correctly and follow the mouse cursor
- Checked dark/light theme compatibility
- Verified the heatmap shows exactly one year of data

### Screenshots

Before:
Screenshot 2025-04-08 at 09 57 02

After:
Screenshot 2025-04-08 at 09 58 06

### DIFF
--- a/app/assets/javascripts/heatmap.js
+++ b/app/assets/javascripts/heatmap.js
@@ -4,54 +4,363 @@
 //= require cal-heatmap/dist/plugins/Tooltip
 
 /* global CalHeatmap, Tooltip */
-document.addEventListener("DOMContentLoaded", () => {
-  const heatmapElement = document.querySelector("#cal-heatmap");
 
-  if (!heatmapElement) {
-    return;
+document.addEventListener("DOMContentLoaded", () => {
+  const config = {
+    rowsCount: 7,
+    allowedDomainType: ["ghDay"],
+    cellSize: 11,
+    cellGutter: 4,
+    // Heuristic: If heatmap starts >15 days into the month, shift month labels for better alignment.
+    daysInMonthThreshold: 15,
+    monthLabelHeight: 20
+  };
+
+  let heatmapStartDate, heatmapEndDate;
+
+  const calculateWeekLabels = (DateHelper, i18nData) => {
+    if (!DateHelper || !i18nData?.abbr_day_names || !Array.isArray(i18nData.abbr_day_names) || i18nData.abbr_day_names.length !== 7) {
+      return Array(config.rowsCount).fill("");
+    }
+
+    const weekStart = DateHelper.date().startOf("week");
+    const weekStartsOnMonday = weekStart.day() === 1;
+    const { abbr_day_names: abbrDayNames } = i18nData;
+    const labels = Array(config.rowsCount).fill("");
+    const labeledRowIndices = [1, 3, 5];
+
+    labeledRowIndices.forEach(rowIndex => {
+      let dayIndex;
+
+      if (weekStartsOnMonday) {
+        dayIndex = (rowIndex + 1) % 7;
+      } else {
+        dayIndex = rowIndex;
+      }
+
+      labels[rowIndex] = abbrDayNames[dayIndex];
+    });
+
+    return labels;
+  };
+
+  const calculateMonthLabels = (startDate, i18nData) => {
+    // Assumes abbr_month_names is 1-indexed [null, Jan, Feb, ... Dec]
+    if (!startDate || !i18nData?.abbr_month_names || !Array.isArray(i18nData.abbr_month_names) || i18nData.abbr_month_names.length !== 13) {
+      return [];
+    }
+
+    const { abbr_month_names: abbrMonthNames } = i18nData;
+    const date = new Date(startDate);
+    const startMonthIndex = date.getUTCMonth();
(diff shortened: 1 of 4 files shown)


hlfan:
> If we prefer that behavior for OSM, I’m happy to refactor it — just let me know!

Could be great iff the `weekInfo` of `Intl.Locale` would be respected to not have the week start on sundays for everyone

kcne:
I was going to look into that next.. I'll let you know if i find a proper solution. Thanks for the recommendation.

tomhughes (commented):
This is adding a whole load of names to the global namespace and I'm not sure any of it is necessary...

Is there are reason all the things declared at global scope before the event listener can't be declared inside it as they were before?

HolgerJeromin:
We could move to es modules. 
There every file has its own scope.
But I have no idea how ruby works with that :)

tomhughes:
Well as far as I know it doesn't if you're using sprockets so it means moving to a whole new asset toolchain.

kcne:
Got this working using the `CalendarLabel` plugin, though I'm seeing some CSP warnings in the console. I'm pretty sure the CSP warnings here are related to [CalendarLabel](https://cal-heatmap.com/docs/plugins/calendarLabel) plugin.

- The first week is fully rendered.
- The heatmap always ends on today's date.
- Starting day of the week is localized.
- Refactored so the code stays inside `DOMContentLoaded` event listener.

#### Screenshots  
Sunday as first day locale:
Screenshot 2025-04-13 at 18 13 20

Monday as first day locale:
Screenshot 2025-04-13 at 18 13 39

CSP warnings:
Screenshot 2025-04-13 at 18 14 03

### Notes

- One challenge I ran into was rendering day labels in the view. I couldn’t find a way to determine whether the week starts on Sunday or Monday from within the view itself, so I opted to use `CalendarLabel` here as well.
- For rendering the month labels, I opted for the `CalendarLabel` API—it’s much cleaner this way, as computing and positioning the labels manually would be unnecessarily complex.

I'm unsure whether we should adjust the config to resolve the CSP warnings, or consider refactoring to manually render the labels instead. Would appreciate any thoughts or suggestions here..

kcne:
I updated the code to manually render the labels in order to eliminate CSP warnings and remove the need for extra plugins. Please let me know if you prefer this manual approach or would rather stick with the plugin. 

The main advantage of manual rendering is that it resolves the CSP issues, but it does add maintenance overhead and introduces new code that needs careful review and testing. 

Using the label plugin is a lot cleaner and less code/maintenance burden, not being sure if there’s a way to address the CSP warnings, or if ignoring them would be acceptable in some cases. I’m open to feedback on direction and awaiting for reviews in either way. Thank you.

tomhughes:
Just to add of course CSP warnings can be resolved but we're not going to be allowing inline script so I'd need to know where that is coming from exactly to know what me might do abut it.

It also looks like it was trying to load something from jsdelivr but we'd really want to host whatever that was ourselves.

tomhughes:
I honestly don't know where to start with this but it seems to me that it takes us from using a third party component with some reasonably simple configuration to writing a vast amount of code to take over much of the work which that component was doing.

I don't even understand how it works... Previously we had a domain of months broken down into days, which seems reasonable, but now we have a domain of days broken down into years? Except clearly we don't if look at the chart it produces... Is the new custom `yearly` domain badly named or something?

The question I keep asking myself every time I look at this and try and understand it is, is all this worth it just to make the heatmap look a bit closer to the github one? At least I assume that's the goal...

It would certainly be nice if we could get it to stop at the current date instead of the end of the current month, which seems to be the main noticeable effect of this to me when I try it, but is it worth this amount of work to get that?

HolgerJeromin:
I also do not like the amount of code. 
But having 0-30 "wrong" day icons is ugly, too.
Perhaps this could be added upstream as an option...
On the other hand: the calendar lib has no updates since a long time... :-(

kcne:
Thanks everyone for the valuable feedback.

I completely understand the concerns around code complexity. My initial intent was to address the visual inconsistency of showing "empty" days beyond today’s date and to ensure proper localization of week starts — something users from different locales will appreciate. I agree that this resulted in more manual code than ideal, especially after trying to avoid CSP issues introduced by the plugin.

Regarding the domain naming: it's a bit misleading, I agree. I’m could rename it to something more accurate, like daysYearlyDomain, to better reflect its structure and avoid confusion.

I also noticed the alternative PR (#5998), which, to be honest, is a much simpler approach that achieves the same goal. I’m open to closing this in favor of merging the better solution, even though I’ve spent quite a bit of time working through this.

tomhughes:
Thanks for all the work on this but I think there's general agreement that #5998 is the better option so I've merged that now.

PR #5908, opened 2025-04-08
Update FAQ.md

NovenkoOlya:


tomhughes:
What exactly is the goal here? How is the new wording better than the old wording?

In general when opening a PR you need to explain what it is you're trying to do and why your PR improves things, not just say "Update <file>" and expect us to guess why you think that is needed.

lonvia:
That's a misguided attempt to fulfil GSOC requirements. Feel free to just close.

PR #5910, opened 2025-04-09
Add zoom to describe location query

### LINKED ISSUE #5909: Error contacting nominatim.openstreetmap.org: 400
At osm.org in the search box there is a "Where is this" link.

Clicking that seems to always give "Error contacting nominatim.openstreetmap.org: 400".  Is this temporary capacity issue, something local to me, or something else?

This currently happens in Firefox on a Linux laptop and on a phone (different internet connections).

hlfan:
Little patch that attempts to close #5909


hlfan:
Closing in favor of #5911.

PR #5913, opened 2025-04-09
Directions route output module

AntonKhorev:
Previously I moved endpoints code out of `directions.js`, but there's still too much code.

This PR creates a module for valid route output which includes:
- total time/distance
- directions table
- download link
- link to the routing engine
- route geometry drawn on the map with hovers/popups

### DIFF
--- a/app/assets/javascripts/index/directions-route-output.js
+++ b/app/assets/javascripts/index/directions-route-output.js
@@ -95,13 +95,15 @@ OSM.DirectionsRouteOutput = function (map) {
           .openOn(map);
       });
 
-      row.hover(function () {
-        highlight
-          .setLatLngs(lineseg)
-          .addTo(map);
-      }, function () {
-        map.removeLayer(highlight);
-      });
+      row
+        .on("mouseenter", function () {
+          highlight
+            .setLatLngs(lineseg)
+            .addTo(map);
+        })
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/index/directions-route-output.js
```diff
@@ -0,0 +1,140 @@
...
+        .on("mouseleave", function () {
+          map.removeLayer(highlight);
+        });
+    }
+
+    const blob = new Blob([JSON.stringify(polyline.toGeoJSON())], { type: "application/json" });
+    URL.revokeObjectURL(downloadURL);
+    downloadURL = URL.createObjectURL(blob);
+
+    $("#directions_content").append(`<p class="text-center"><a href="${downloadURL}" download="${
```
tomhughes, line 113: Shouldn't this, and the one that follows, be appending to the `content` parameter rather than looking it up in the DOM again?

AntonKhorev: Yes, it should. Updated.

tomhughes (approved):

hlfan (changes requested):
I think as an "output", that module should get initialized with `$("#directions_content")` so we don't need to specify the location of the output every time.

AntonKhorev:
`#directions_content` is not present all of the time.

PR #5914, opened 2025-04-09
Attach routing credit to route

hlfan:
Much like in the API responses from GraphHopper or Overpass Turbo, this PR adds the credit information to the returned routes, removing the need to pass the `chosenEngine` to the `DirectionsRouteOutput`.


nenad-vujicic (approved):

hlfan:
Yep, meant `OSM.FOSSGIS_OSRM_URL + "about.html"` but that's not what I wrote, is it?

PR #5915, opened 2025-04-10
Add kilometers/miles switch to directions

AntonKhorev:
image

Previous attempt was in #1747. This PR is a complete rewrite.

The main difference is that #1747 tried to store in locale strings what kind of "imperial" units are in use, along with conversion factors. I don't think that was a good idea and I don't do that in this PR.

I also moved most of html writing from javascript to the erb template because it's easier to deal with flex containers and button groups this way. This is probably what @gravitystorm wants for all javascript page controllers.

### DIFF
--- a/app/assets/javascripts/index/directions-route-output.js
+++ b/app/assets/javascripts/index/directions-route-output.js
@@ -13,10 +13,15 @@ OSM.DirectionsRouteOutput = function (map) {
     weight: 12
   });
 
+  let distanceUnits = "km";
   let downloadURL = null;
 
   function translateDistanceUnits(m) {
-    return [m, "m", m / 1000, "km"];
+    if (distanceUnits === "km") {
+      return [m, "m", m / 1000, "km"];
+    } else {
+      return [m / 0.3048, "ft", m / 1609.344, "mi"];
+    }
   }
 
   function formatTotalDistance(minorValue, minorName, majorValue, majorName) {
@@ -117,6 +122,17 @@ OSM.DirectionsRouteOutput = function (map) {
     writeSummary(route);
     writeSteps(route);
 
+    $("#directions_distance_units_km").off().on("change", () => {
+      distanceUnits = "km";
+      writeSummary(route);
+      writeSteps(route);
+    });
+    $("#directions_distance_units_mi").off().on("change", () => {
+      distanceUnits = "mi";
+      writeSummary(route);
+      writeSteps(route);
+    });
+
     const blob = new Blob([JSON.stringify(polyline.toGeoJSON())], { type: "application/json" });
     URL.revokeObjectURL(downloadURL);
     downloadURL = URL.createObjectURL(blob);
@@ -139,6 +155,9 @@ OSM.DirectionsRouteOutput = function (map) {
     map
       .removeLayer(popup)
       .removeLayer(polyline);
+
+    $("#directions_distance_units_km").off();
+    $("#directions_distance_units_mi").off();
 
     $("#directions_route_steps").empty();
 
--- a/app/assets/javascripts/index/directions-route-output.js
+++ b/app/assets/javascripts/index/directions-route-output.js
@@ -16,31 +16,37 @@ OSM.DirectionsRouteOutput = function (map) {
   let downloadURL = null;
 
   function formatTotalDistance(m) {
+    const scope = "javascripts.directions.distance_in_units";
+
     if (m < 1000) {
-      return OSM.i18n.t("javascripts.directions.distance_m", { distance: Math.round(m) });
+      return OSM.i18n.t("m", { scope, distance: Math.round(m) });
     } else if (m < 10000) {
-      return OSM.i18n.t("javascripts.directions.distance_km", { distance: (m / 1000.0).toFixed(1) });
+      return OSM.i18n.t("km", { scope, distance: (m / 1000.0).toFixed(1) });
     } else {
-      return OSM.i18n.t("javascripts.directions.distance_km", { distance: Math.round(m / 1000) });
+      return OSM.i18n.t("km", { scope, distance: Math.round(m / 1000) });
     }
   }
 
   function formatStepDistance(m) {
+    const scope = "javascripts.directions.distance_in_units";
+
     if (m < 5) {
       return "";
     } else if (m < 200) {
-      return OSM.i18n.t("javascripts.directions.distance_m", { distance: String(Math.round(m / 10) * 10) });
+      return OSM.i18n.t("m", { scope, distance: String(Math.round(m / 10) * 10) });
     } else if (m < 1500) {
-      return OSM.i18n.t("javascripts.directions.distance_m", { distance: String(Math.round(m / 100) * 100) });
+      return OSM.i18n.t("m", { scope, distance: String(Math.round(m / 100) * 100) });
     } else if (m < 5000) {
-      return OSM.i18n.t("javascripts.directions.distance_km", { distance: String(Math.round(m / 100) / 10) });
+      return OSM.i18n.t("km", { scope, distance: String(Math.round(m / 100) / 10) });
     } else {
-      return OSM.i18n.t("javascripts.directions.distance_km", { distance: String(Math.round(m / 1000)) });
+      return OSM.i18n.t("km", { scope, distance: String(Math.round(m / 1000)) });
     }
   }
 
   function formatHeight(m) {
-    return OSM.i18n.t("javascripts.directions.distance_m", { distance: Math.round(m) });
+    const scope = "javascripts.directions.distance_in_units";
+
+    return OSM.i18n.t("m", { scope, distance: Math.round(m) });
   }
 
   function formatTime(s) {
--- a/app/assets/javascripts/index/directions-route-output.js
+++ b/app/assets/javascripts/index/directions-route-output.js
@@ -33,13 +33,13 @@ OSM.DirectionsRouteOutput = function (map) {
     if (m < 5) {
       return "";
     } else if (m < 200) {
-      return OSM.i18n.t("m", { scope, distance: String(Math.round(m / 10) * 10) });
+      return OSM.i18n.t("m", { scope, distance: Math.round(m / 10) * 10 });
     } else if (m < 1500) {
-      return OSM.i18n.t("m", { scope, distance: String(Math.round(m / 100) * 100) });
+      return OSM.i18n.t("m", { scope, distance: Math.round(m / 100) * 100 });
     } else if (m < 5000) {
-      return OSM.i18n.t("km", { scope, distance: String(Math.round(m / 100) / 10) });
+      return OSM.i18n.t("km", { scope, distance: Math.round(m / 100) / 10 });
     } else {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/index/directions-route-output.js
```diff
@@ -13,15 +13,32 @@ OSM.DirectionsRouteOutput = function (map) {
...
+      } else if (km < 10) {
+        return OSM.i18n.t("javascripts.directions.distance_km", { distance: km.toFixed(1) });
+      } else {
+        return OSM.i18n.t("javascripts.directions.distance_km", { distance: Math.round(km) });
+      }
     } else {
-      return OSM.i18n.t("javascripts.directions.distance_km", { distance: Math.round(m / 1000) });
+      const ft = m / ftSize;
+      const mi = m / miSize;
+      if (ft < 1000) {
```
tomhughes: Is 1000 still the right boundary to use in feet/miles? It's very different to 1000m in terms of distance... Something like 3280 would be the equivalent, or alternatively 5280 if we wanted to stick to using minor units up to the first major unit.

Then again at least in the UK we're more likely to use yards than feet as the minor unit...

AntonKhorev: It was at 0.2 miles in https://github.com/openstreetmap/openstreetmap-website/pull/1747 which is about 1000 feet:

https://github.com/openstreetmap/openstreetmap-website/blob/ff73d379a723060870f591451f4092343a3b4077/app/assets/javascripts/index/directions.js#L151

It depends on what is the goal. If the goal is not to have too large number in minor units, the boundary can be the same number.

1ec5: For what it’s worth, the [MapLibre Navigation SDK](https://github.com/maplibre/maplibre-navigation-ios/blob/f91faffa1aa03be16fc588dc5076b7811d1c5166/MapboxCoreNavigation/DistanceFormatter.swift#L146-L160) switches at 1,000&nbsp;meters = 1&nbsp;kilometer, 176&nbsp;yards = 1&nbsp;mile for the UK, or 528&nbsp;feet = 1&nbsp;mile for the U.S. (ignoring the rounding increments that are applied for readability at a glance). [Ferrostar](https://github.com/stadiamaps/ferrostar/blob/f10b01c03bec9554a7cf827327795d253ad4c1f0/apple/Sources/FerrostarCarPlayUI/Measurement/MKDistanceFormatterUnits.swift#L36-L54) switches at 1,000&nbsp;meters = 1&nbsp;kilometer, 300&nbsp;yards = 0.17&nbsp;miles for the UK, or 289&nbsp;feet = 0.16&nbsp;miles for the U.S. These navigation SDKs don’t aim for equivalent thresholds between measurement systems, since most users aren’t switching back and forth between them.

app/assets/javascripts/index/directions-route-output.js
```diff
@@ -105,22 +126,37 @@ OSM.DirectionsRouteOutput = function (map) {
...
+    $("#directions_distance_units_mi").off().on("change", () => {
+      distanceUnits = "mi";
+      writeSummary(route);
+      writeSteps(route);
+    });
 
     const blob = new Blob([JSON.stringify(polyline.toGeoJSON())], { type: "application/json" });
     URL.revokeObjectURL(downloadURL);
     downloadURL = URL.createObjectURL(blob);
+    $("#directions_route_download").prop("href", downloadURL);
```
hlfan, line 155: If you have the DOM anyway you can drop `downloadURL`

```suggestion
    URL.revokeObjectURL(downloadLink.href);
    downloadLink.href = URL.createObjectURL(blob);
```

app/assets/javascripts/index/directions-route-output.js
AntonKhorev: > 528 feet = 1 mile

You mean at 0.1 miles? That's moving the boundary in a different direction.

AntonKhorev: 0.1 miles is where Google Maps are switching. The logic is probably the lowest pretty number in major units that doesn't let minor units go into four digits.

0.2 miles from https://github.com/openstreetmap/openstreetmap-website/pull/1747 was a close candidate that allowed a small range of 4 digit numbers.

app/assets/javascripts/index/directions-route-output.js
```diff
@@ -13,34 +13,65 @@ OSM.DirectionsRouteOutput = function (map) {
...
-    return OSM.i18n.t("javascripts.directions.distance_m", { distance: Math.round(m) });
+    const scope = "javascripts.directions.distance_in_units";
+
+    if (distanceUnits === "km") {
+      return OSM.i18n.t("m", { scope, distance: Math.round(m) });
+    } else {
+      const ft = m / ftSize;
+      return OSM.i18n.t("ft", { scope, distance: Math.round(ft) });
+    }
   }
```
hlfan, line 75: I think separating the unit-agnostic logic from the unit-specific stuff could make the functions more straightforward:

```js
const common = {
  scope: "javascripts.directions.distance_in_units",
  _formatTotal(minorValue, minorName, majorValue, majorName) {
    const scope = this.scope;
    if (minorValue < 1000) return OSM.i18n.t(minorName, { scope, distance: Math.round(minorValue) });
    if (majorValue < 10) return OSM.i18n.t(majorName, { scope, distance: majorValue.toFixed(1) });
    return OSM.i18n.t(majorName, { scope, distance: Math.round(majorValue) });
  },
  _formatStep(minorValue, minorName, majorValue, majorName) {
    const scope = this.scope;
    if (minorValue < 200) return OSM.i18n.t(minorName, { scope, distance: Math.round(minorValue / 10) * 10 });
    if (minorValue < 1500) return OSM.i18n.t(minorName, { scope, distance: Math.round(minorValue / 100) * 100 });
    if (majorValue < 5) return OSM.i18n.t(majorName, { scope, distance: majorValue.toFixed(1) });
    return OSM.i18n.t(majorName, { scope, distance: Math.round(majorValue) });
  },
  _formatHeight(value, name) {
    return OSM.i18n.t(name, { scope: this.scope, distance: Math.round(value) });
  },
  formatTime(s) {
    const m = Math.round(s / 60);
    const h = Math.floor(m / 60);
    return h + ":" + `${m - (h * 60)}`.padStart(2, 0);
  }
};
const formatters = {
  km: {
    ...common,
    formatTotal(m) { return this._formatTotal(m, "m", m / 1000, "km"); },
    formatStep(m) { return this._formatStep(m, "m", m / 1000, "km"); },
    formatHeight(m) { return this._formatHeight(m, "m"); }
  },
  mi: {
    ...common,
    ftSize: 0.3048,
    miSize: 1609.344,
    formatTotal(m) { return this._formatTotal(m / this.ftSize, "ft", m / this.miSize, "mi"); },
    formatStep(m) { return this._formatStep(m / this.ftSize, "ft", m / this.miSize, "mi"); },
    formatHeight(m) { return this._formatHeight(m / this.ftSize, "ft"); }
  }
};
```

Then, set `currentFormatter = formatters.mi` in the `#directions_distance_units_mi.change` handler and use the `formatStep`, `formatTotal`, etc.. directly from `currentFormatter`.

app/assets/javascripts/index/directions-route-output.js
AntonKhorev: We have two different m-km boundaries:

1. The first added was for step distances at 1500m in https://github.com/openstreetmap/openstreetmap-website/commit/c6e8775dbab81daf9515fe94fa0ed4c51cab5f3b
2. The second one was for the summary at 1000m in https://github.com/openstreetmap/openstreetmap-website/commit/3ef8024b1d33f0de6227a26967ba78ef85672a94

Why do we need two? I don't know. I'd expect the step one to be lower to have smaller numbers inside the steps table, but no.

app/assets/javascripts/index/directions-route-output.js
hlfan: And it can do the job of `downloadURL` too since you preserve the DOM starting with 9e908a00b0f7d7c4ffbbf93b8dae80a3d2c3b05f

AntonKhorev: I'm saying that it's not a string variable that preserves the value you put in it.

app/assets/javascripts/index/directions-route-output.js
AntonKhorev: I'm not doing 0.1 miles because it's going to one significant digit too soon. I tried different boundaries and I chose:
- `minorValue < 1000 || majorValue < 0.5` for total distance
- `minorValue < 1500 || majorValue < 0.5` for step distance

1ec5: > You mean at 0.1 miles?

Yep, that was a typo, fixed. The difference between metric and non-metric in these SDKs is due to a large quantity like 999&nbsp;meters still being comprehensible compared to 5,280&nbsp;feet. Fewer digits helps too.

AntonKhorev: > They tend to start seeing miles at either 0.1 miles or ¼ mile.

Do you propose changing `majorValue < 0.5` to `majorValue < 0.25`?

1ec5: That would seem more reasonable to me. 👍

AntonKhorev: Changed to 0.25.

app/assets/javascripts/index/directions-route-output.js
```diff
@@ -13,34 +13,49 @@ OSM.DirectionsRouteOutput = function (map) {
...
     } else {
-      return OSM.i18n.t("javascripts.directions.distance_km", { distance: Math.round(m / 1000) });
+      return [m / 0.3048, "ft", m / 1609.344, "mi"];
     }
   }
 
-  function formatStepDistance(m) {
-    if (m < 5) {
+  function formatTotalDistance(minorValue, minorName, majorValue, majorName) {
+    const scope = "javascripts.directions.distance_in_units";
```
hlfan, line 28:
```suggestion
  const formatters = {
    km: { units: m => [m, "m", m / 1000, "km"] },
    mi: { units: m => [m / 0.3048, "ft", m / 1609.344, "mi"] }
  };
  let currentFormatter = formatters.km;
  let downloadURL = null;

  function formatTotalDistance(m) {
    const [minorValue, minorName, majorValue, majorName] = currentFormatter.units(m);
    const scope = "javascripts.directions.distance_in_units";
```
so the `...translateDistanceUnits(...)` can be dropped

AntonKhorev: If we ignore that what's called a *formatter* here doesn't format anything, I'm not going to do functions/objects as values of km/mi preference until I decide where I'm storing that preference. Let's suppose the preference is persisted in a cookie. Certainly I'm not going to store a function or an object there.

hlfan: Why would you want to store a function in a cookie?
`formatterUnits[Cookies.get("distanceUnits")]` or `formatterUnits[localStorage.distanceUnits]` works just fine.

tomhughes (approved):
Thanks for the work on this... I think it looks good to merge now.

PR #5917, opened 2025-04-10
Update history.js - Correct function name for updating changeset loca…

danieldegroot2:
Fixes regression from https://github.com/openstreetmap/openstreetmap-website/commit/f11b38709b6cdf6b837d822f4c1e04f06d65ca71 where changeset highlights for `/user/history` disappear after automatically zooming in, do not load at all or are incorrectly scaled when manually zooming.

Corrects invalid function name (`Uncaught TypeError: changesetsLayer.updateChangesetsPositions is not a function`)

cc: @AntonKhorev


tomhughes:
Looks good to me, thanks.

PR #5918, opened 2025-04-10
Put map key presence in js layer definitions

hlfan:
Links _key.yml_ to _layers.yml_ in the `MapLayers` module by putting a `hasKey` flag into the `LAYER_DEFINITIONS`.
This removes the need for `LAYERS_WITH_MAP_KEY` as the key button can now directly check `layer.options.hasKey`.

### DIFF
--- a/lib/map_layers.rb
+++ b/lib/map_layers.rb
@@ -1,5 +1,6 @@
 module MapLayers
-  def self.full_definitions(layers_filename)
+  def self.full_definitions(layers_filename, legends: nil)
+    legended_layers = YAML.load_file(Rails.root.join(legends)).keys if legends
     YAML.load_file(Rails.root.join(layers_filename))
         .reject { |layer| layer["apiKeyId"] && !Settings[layer["apiKeyId"]] }
         .map do |layer|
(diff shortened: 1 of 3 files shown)


lib/map_layers.rb
```diff
@@ -1,12 +1,14 @@
...
+  def self.full_definitions(layers_filename, map_keys: nil)
+    keys = YAML.load_file(Rails.root.join(map_keys)).keys if map_keys
     YAML.load_file(Rails.root.join(layers_filename))
         .reject { |layer| layer["apiKeyId"] && !Settings[layer["apiKeyId"]] }
         .map do |layer|
           if layer["apiKeyId"]
             layer["apikey"] = Settings[layer["apiKeyId"]]
             layer.delete "apiKeyId"
           end
+          layer["hasKey"] = true if keys&.include?(layer["layerId"])
```
AntonKhorev, line 11: The word "key" is overused. "hasKey" - which key? Here it's right next to another entirely different "key".

hlfan: I appended my changes, but the different terminologies still bug me. That, however, can be changed in another PR.

AntonKhorev:
Merged, thanks.

PR #5920, opened 2025-04-11
Test if 'Show address' includes zoom parameter

AntonKhorev:
Test for #5916.


tomhughes:
Looks good to me, thanks.

PR #5923, opened 2025-04-11
Update to rack 3

tomhughes:
This updates to rack 3 which was blocked by omniauth being locked to 2.0 so this updates omniauth to 2.1 which unlocks the rack 3 update which in turn requires a puma update.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,7 +12,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from RailsParam::InvalidParameterError, :with => :invalid_parameter
 
-  before_action :fetch_body
+  after_action :close_body
 
   attr_accessor :current_user, :oauth_token
 
@@ -244,10 +244,10 @@ class ApplicationController < ActionController::Base
   #
   #   https://issues.apache.org/bugzilla/show_bug.cgi?id=44782
   #
-  # To work round this we call rewind on the body here, which is added
-  # as a filter, to force it to be fetched from Apache into a file.
-  def fetch_body
-    request.body.rewind
+  # To work round this we call close on the body here, which is added
+  # as a filter, to let Apache know we are done with it.
+  def close_body
+    request.body&.close
   end
 
   def map_layout
--- a/Gemfile
+++ b/Gemfile
@@ -78,7 +78,7 @@ gem "rack-uri_sanitizer"
 
 # Omniauth for authentication
 gem "multi_json"
-gem "omniauth", "~> 2.0.2"
+gem "omniauth", "~> 2.1.3"
 gem "omniauth-facebook"
 gem "omniauth-github"
 gem "omniauth-google-oauth2", ">= 0.6.0"
@@ -162,7 +162,7 @@ group :test do
   gem "jwt"
   gem "minitest", "~> 5.1"
   gem "minitest-focus", :require => false
-  gem "puma", "~> 5.6"
+  gem "puma", "~> 6.6"
   gem "rails-controller-testing"
   gem "rubocop"
   gem "rubocop-capybara"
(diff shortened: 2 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -247,7 +247,7 @@ def respond_to_timeout
   # To work round this we call rewind on the body here, which is added
   # as a filter, to force it to be fetched from Apache into a file.
   def fetch_body
-    request.body.rewind
+    request.body&.rewind
```
rkoeze, line 250: For my own edification was this linked to one of the version bumps?

rkoeze (approved):
Perused the changelogs for omniauth, rack and puma.  Didn't see anything worrisome in the first two. The Puma changelist was pretty big given that we're bumping a major version. But running locally everything worked well!

app/controllers/application_controller.rb
mmd-osm: Try running unit tests to find out why this is needed.

rkoeze: Thanks @mmd-osm for the response. It's clear to me why the safe navigation operator is needed to get the tests to pass. I was trying to understand which upstream change prompted this.

mmd-osm: I'm assuming this might be something unit test specific. A non-existing request body used to be represented by an empty string, and is now nil.

When running the code in rails server, request body will be #<Puma::NullIO:0x0000xxx> in this case. This wouldn't trigger the undefined method 'rewind' for nil error anymore.

AntonKhorev:
Does this work now, with lowercase header keys?
https://github.com/openstreetmap/openstreetmap-website/blob/fd7b428007905463e2359ed0823920954c699375/config/initializers/cors.rb#L10

app/controllers/application_controller.rb
tomhughes: It seems rack 3 has a change in it's mock requests that tests use so that `body` is now `nil` for requests like GET that don't have any body - hence what I had to add that to avoid issues.

tomhughes:
> Does this work now, with lowercase header keys?

Am I missing some context here? What "lowercase header keys" are you referring to?

AntonKhorev:
Is `if headers["Access-Control-Allow-Origin"]` ever going to be true in [config/initializers/cors.rb](https://github.com/openstreetmap/openstreetmap-website/blob/fd7b428007905463e2359ed0823920954c699375/config/initializers/cors.rb#L10)?

tomhughes:
Why would it not be if a request arrives with that header? Are you suggesting that rack 3 has changed how it present request headers or something?

tomhughes:
I assume you're thinking about https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#response-headers-must-be-lower-case which is technically about response headers though the same is essentially true for request headers.

I think the reason they don't mention request headers is that the `headers` object rack passes you is always a `Rack::Headers` object which does case insensitive lookups, so the case here shouldn't matter.

AntonKhorev:
Previously I got `Cache-Control: no-cache` in responses to for example `/api/capabilities`. Now I get `cache-control: max-age=0, private, must-revalidate`.

AntonKhorev:
I'm talking about response headers.

tomhughes:
Yes it downcases everything now but so long as you're using `Rack::Headers` instead of a plain hash any lookups are also downcased.

AntonKhorev:
`headers` here
https://github.com/openstreetmap/openstreetmap-website/blob/fd7b428007905463e2359ed0823920954c699375/config/initializers/cors.rb#L9

were `Rack::Utils::HeaderHash` before and became `Hash` after this PR.

tomhughes:
Nothing there specifies a class? The class is what ever rack passes up the stack, and from my debugging that is `Rack::Headers`.

AntonKhorev:
If you run this test https://github.com/openstreetmap/openstreetmap-website/blob/fd7b428007905463e2359ed0823920954c699375/test/integration/cors_test.rb#L4

do you get to this line
https://github.com/cyu/rack-cors/blob/a3b0a90d7bd7f8cc32df6bcf993b14673f466283/lib/rack/cors/resource.rb#L48

where `headers` is starting to get constructed as a plain hash?

PR #5924, opened 2025-04-12
Change colors of changeset bboxes when they enter/exit the viewport

AntonKhorev:
On history pages, if you load many changesets and look at the map, it's difficult to find the corresponding changesets in the sidebar. The idea here is to color changesets differently, depending on whether they are above/in/below the sidebar viewport.

- The orange color that was previously used for all changesets is now used only for the ones in the viewport.
- The bluish ("colder") color is for older changesets below the viewport. If you see this color on the map, you know you have to scroll down to find the changeset in the sidebar.
- The reddish ("hotter") is for newer changesets. Actually the color became more brownish as I desaturated it because it has to beless saturated than the usual orange.

https://github.com/user-attachments/assets/0f6eb2d8-3b66-4890-ae23-9f160aadb50b

### DIFF
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -64,12 +64,8 @@ OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
     }
   },
 
-  highlightChangeset: function (id) {
-    this.getLayer(id)?.setStyle({ fillOpacity: 0.3, color: "#FF6600", weight: 3 });
-  },
-
-  unHighlightChangeset: function (id) {
-    this.getLayer(id)?.setStyle({ fillOpacity: 0, color: "#FF9500", weight: 2 });
+  toggleChangesetHighlight: function (id, state) {
+    this.getLayer(id)?.setStyle(state ? { fillOpacity: 0.3, color: "#FF6600", weight: 3 } : { fillOpacity: 0, color: "#FF9500", weight: 2 });
   },
 
   getLayerId: function (layer) {
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -1,15 +1,15 @@
 OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
-  _changesets: [],
+  _changesets: new Map,
 
   updateChangesets: function (map, changesets) {
-    this._changesets = changesets;
+    this._changesets = new Map(changesets.map(changeset => [changeset.id, changeset]));
     this.updateChangesetShapes(map);
   },
 
   updateChangesetShapes: function (map) {
     this.clearLayers();
 
-    for (const changeset of this._changesets) {
+    for (const changeset of this._changesets.values()) {
       const bottomLeft = map.project(L.latLng(changeset.bbox.minlat, changeset.bbox.minlon)),
             topRight = map.project(L.latLng(changeset.bbox.maxlat, changeset.bbox.maxlon)),
             width = topRight.x - bottomLeft.x,
@@ -30,13 +30,15 @@ OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
                                         map.unproject(topRight));
     }
 
-    this._changesets.sort(function (a, b) {
+    const changesetEntries = [...this._changesets];
+    changesetEntries.sort(([, a], [, b]) => {
       return b.bounds.getSize() - a.bounds.getSize();
     });
+    this._changesets = new Map(changesetEntries);
 
     this.updateChangesetLocations(map);
 
-    for (const changeset of this._changesets) {
+    for (const changeset of this._changesets.values()) {
       const rect = L.rectangle(changeset.bounds,
                                { weight: 2, color: "#FF9500", opacity: 1, fillColor: "#FFFFAF", fillOpacity: 0 });
       rect.id = changeset.id;
@@ -47,7 +49,7 @@ OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
   updateChangesetLocations: function (map) {
     const mapCenterLng = map.getCenter().lng;
 
-    for (const changeset of this._changesets) {
+    for (const changeset of this._changesets.values()) {
       const changesetSouthWest = changeset.bounds.getSouthWest();
       const changesetNorthEast = changeset.bounds.getNorthEast();
       const changesetCenterLng = (changesetSouthWest.lng + changesetNorthEast.lng) / 2;
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -1,9 +1,16 @@
 OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
   _changesets: new Map,
 
-  _getChangesetStyle: function ({ isHighlighted }) {
-    let className = "changeset-in-sidebar-viewport";
+  _getChangesetStyle: function ({ isHighlighted, sidebarRelativePosition }) {
+    let className;
 
+    if (sidebarRelativePosition > 0) {
+      className = "changeset-above-sidebar-viewport";
+    } else if (sidebarRelativePosition < 0) {
+      className = "changeset-below-sidebar-viewport";
+    } else {
+      className = "changeset-in-sidebar-viewport";
+    }
     if (isHighlighted) {
       className += " changeset-highlighted";
     }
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/index/history-changesets-layer.js
```diff
@@ -1,15 +1,40 @@
 OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
-  _changesets: [],
+  newerColor: "#AA8866",
+  olderColor: "#8888AA",
```
tomhughes, line 3: If we're going to use named constants for these colours shouldn't we do it for the others as well?

tomhughes (commented):
Technically this looks so I think the main questions here will be whether (a) people think it's a good idea and (b) whether these are the best colour choices.

app/assets/javascripts/index/history-changesets-layer.js
AntonKhorev: Includes workaround for https://github.com/leaflet/leaflet/issues/2662

tomhughes:
Thanks. This looks good to me now, and nobody's had anything to say about the colours so I see no reason not to merge this.

mxdanger:
This should be reverted. It's really trying to solve a problem but going about it in the wrong way. This PR just visualizes the annoyance of not being able to see that the changeset you're hovering over is.

Here's my proposal:

1. Revert this PR as it won't be needed anymore.
2. Use `scrollIntoView()` to bring the selected list-group-item into view in the sidebar so you can always see what you're hovering over in the map.
3. If we're feeling generous, use `scroll-behavior: smooth;`
4. Say goodby to the days of trying to find the corresponding changesets in the sidebar.

AntonKhorev:
> selected list-group-item

Selected how? There's no selection mechanism as of right now.

limes11:
I strongly urge you to rethink this approach. Whilst I do not oppose using different colours depending on the age of changesets, the fact that older changesets are no longer clickable, makes scanning an area which does not show high activity, really painful: 
Large changesets covering several countries or continents - which have always been a nuisance - cause relevant changesets to drop down in the list making it necessary to perform a well-adjusted scrolling manoeuvre constantly keeping an eye on the bounding box color before one is finally able to click on the desired bounding box. 
This feels idiotic. I understand the problem you are trying to solve but this solution is not the way forward.

Besides, for people with issues of color-blindness, orange and red are too similar. I can't tell them apart without looking very carefully.

mxdanger:
> > selected list-group-item
> 
> 
> 
> Selected how? There's no selection mechanism as of right now.

You might be confused. When you hover over a changeset bbox on the map it marks the respective changeset in the side bar as selected.

AntonKhorev:
> > Selected how? There's no selection mechanism as of right now.
> 
> You might be confused. When you hover over a changeset bbox on the map it marks the respective changeset in the side bar as selected.

@mxdanger The changeset is not "selected" on hover (maybe you looked at css class names and saw *selected* but it's not really selected). If what you propose is sidebar scrolling frantically as you move your mouse over the map (and not scrolling anywhere when you're using a device without a mouse), I expect about 10x more posts starting with "This should be reverted".

AntonKhorev:
> I strongly urge you to rethink this approach. Whilst I do not oppose using different colours depending on the age of changesets, the fact that older changesets are no longer clickable, makes scanning an area which does not show high activity, really painful: Large changesets covering several countries or continents - which have always been a nuisance - cause relevant changesets to drop down in the list making it necessary to perform a well-adjusted scrolling manoeuvre constantly keeping an eye on the bounding box color before one is finally able to click on the desired bounding box.

Older changesets are clickable unless covered by those large nuisance chagesets when they are inside the sidebar viewport. I'm not sure if changesets that cover the entire map viewport should be hoverable/clickable, but on the other hand there has to be some indication that they are intersecting the map view.

There's also an opinion that [changesets outside of the sidebar viewport shouldn't be shown at all](https://community.openstreetmap.org/t/better-osm-org-a-script-that-adds-useful-little-things-to-osm-org/121670/105). In this case you wouldn't see any older changesets.

> Besides, for people with issues of color-blindness, orange and red are too similar. I can't tell them apart without looking very carefully.

Orange and orange are even more similar. And again there's an opinion about "color noise", that is colors being too different.

mxdanger:
> If what you propose is sidebar scrolling frantically as you move your mouse over the map (and not scrolling anywhere when you're using a device without a mouse), I expect about 10x more posts starting with "This should be reverted".

Considering that your implementation doesn't really solve the issue of "it's difficult to find the corresponding changesets in the sidebar" I don't think you should be dismissive of the scroll into view idea, which would make it incredibly easy and intuitive to see what changeset you're hovering over is without having to click on it.

Give it a try and see if it works as a viable option rather than outright dismissing it, surely it shouldn't be too hard to implement `scrollIntoView()` [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) into the function that adds the selected css class, and get some feedback from the community forums. Sure it might jump around if you move your mouse all over the map but it makes intuitive sense.

Playing a how/cold game to hunt down a changeset does not make sense, however I guess on mobile it's fine.

AntonKhorev:
> Give it a try and see if it works as a viable option rather than outright dismissing it

You give it a try.

deevroman:
> Use scrollIntoView() to bring the selected list-group-item into view in the sidebar so you can always see what you're hovering over in the map.

This is a tricky function. I use it in my script for osm.org to navigate between objects in a changesets, and sometimes it works unpredictably and moves the scroll to the wrong place (or rather, sometimes it doesn't move the scroll at all). Perhaps this is a problem with the current layout, but in any case, I would be careful in expectations of it.

> Sure it might jump around if you move your mouse all over the map but it makes intuitive sense.

What is the point of doing this, then? Imagine, you are scrolling through changesets, and finally an interesting cluster of edits appears in the area you need, you move the cursor to them and the sidebar starts jumping. Then it turns out that you need to load more changesets, and you move the cursor to the sidebar again, enjoying its jumping.

Also, this is not the only solution. You can scroll with a delay. You can scroll by alt + click by bbox. You can display by hover title with meta information. You can show a popup. PR welcome :)

> This should be reverted.

Now, viewing changesets on osm.org is a clean slate, so almost any change doesn't make worse.

AntonKhorev:
> You can scroll with a delay.

Then you'll have to decide on the length of the delay. Too short and it's useless. Too long and it's annoying because you have to wait for at least this long, and usually it's annoying for delays >= 300ms.

limes11:
> > I strongly urge you to rethink this approach. Whilst I do not oppose using different colours depending on the age of changesets, the fact that older changesets are no longer clickable, makes scanning an area which does not show high activity, really painful: Large changesets covering several countries or continents - which have always been a nuisance - cause relevant changesets to drop down in the list making it necessary to perform a well-adjusted scrolling manoeuvre constantly keeping an eye on the bounding box color before one is finally able to click on the desired bounding box.
> 
> Older changesets are clickable unless covered by those large nuisance chagesets when they are inside the sidebar viewport. I'm not sure if changesets that cover the entire map viewport should be hoverable/clickable, but on the other hand there has to be some indication that they are intersecting the map view.

In principle, omitting changesets covering the entire map viewport sounds like an interesting idea. These changesets usually require to be looked at separately anyways. When verifying recent activities, one is in any case forced to look at the map using all sorts of zoom levels. Eventually, larger changesets are going to be included.

> 
> There's also an opinion that [changesets outside of the sidebar viewport shouldn't be shown at all](https://community.openstreetmap.org/t/better-osm-org-a-script-that-adds-useful-little-things-to-osm-org/121670/105). In this case you wouldn't see any older changesets.

This suggestion disconnects entirely from requirements of average contributors. And judging from the smiley, I doubt this was meant seriously.

> 
> > Besides, for people with issues of color-blindness, orange and red are too similar. I can't tell them apart without looking very carefully.
> 
> Orange and orange are even more similar. And again there's an opinion about "color noise", that is colors being too different.

Admittedly, the *same* color would have been a worse choice, but this cannot be an argument for keeping a poor choice of colors. Removing some red in the orange rendering the color more yellowish would probably be enough help for color-blind people (like myself). I don't understand the "color noise" issue - this recent update fundamentally relies on three sufficiently different colors in order to maintain useability of the history functionality altogether. It's not about map rendering, it's about detecting map changes.

deevroman:
> In principle, omitting changesets covering the entire map viewport sounds like an interesting idea.

In practice, you will encounter a problem: what is the threshold value to take? What if a big changeset did change something in your region? For example, I have a cyclist in a city of five million people who really likes to make changesets after driving around the entire city :)

> This suggestion disconnects entirely from requirements of average contributors. 

Be careful with generalizations if you haven't conducted any survey or experiment. 

> And judging from the smiley, I doubt this was meant seriously.

The smile emphasized the suddenness of the idea. And in fact, this can already be achieved if you reload the page (or open the changeset and go back), because /history remembers the scroll position.

DesertBear:
Sorry, but this change is terrible.

I check my local area (35×35 miles) daily for changes.  This involves panning the area and then clicking on new changeset bounding boxes.  In a populated area, this can be quite a few changesets.  If there's a particular changeset that want to check further, I then view the changeset number in achavi.  

With this latest change to OSM History, only five changesets maximum are clickable (although realistically, only two or three are clickable because some of those are going to be large area edits).  So now we have scroll down the left hand column until the changeset bounding box turns orange, then move over to the map and click on the changeset to view.  I have to repeatedly keep doing this over and over again now.

Scroll, scroll, scroll, click.  Scroll, scroll, scroll, click.  Scroll, scroll, scroll, click. 

I tried it for a week, but I just can't use it like this – it's beyond tedious.

PR #5925, opened 2025-04-12
Add OSRM ferry support

### LINKED ISSUE #1000: Ferry routing not in text in routing directions
The text field should really say it is by ferry, the first 200 meters in step 10 in [this route](http://www.openstreetmap.org/directions?engine=osrm_car&route=-19.9861%2C23.4224%3B-18.8167%2C22.4209#map=8/-19.368/22.606) are by ferry but that is not indicated in the text field itself.

Perhaps it should be split up so "10. 200 meters by ferry" would be displayed. According to OSRM the value of "travel by ferry" or similar is returned, but the display side is not using it.

hlfan:
### Description
Closes #1000 again

### How has this been tested?
Without the added translation

### DIFF
--- a/app/assets/javascripts/index/directions/fossgis_osrm.js
+++ b/app/assets/javascripts/index/directions/fossgis_osrm.js
@@ -30,6 +30,7 @@
         "rotary": "roundabout",
         "exit roundabout": "exit_roundabout",
         "exit rotary": "exit_roundabout",
+        "ferry": "ferry",
         "depart": "start",
         "arrive": "destination"
       };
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/index/directions/fossgis_osrm.js
```diff
@@ -30,6 +30,7 @@
         "rotary": "roundabout",
         "exit roundabout": "exit_roundabout",
         "exit rotary": "exit_roundabout",
+        "ferry": "continue",
```
tomhughes, line 33: Shouldn't this be:

```suggestion
        "ferry": "ferry",
```

As it stands I get the ferry icon but the text is still just "Continue on..." rather than the new ferry message.

tomhughes (changes requested):

app/assets/javascripts/index/directions/fossgis_osrm.js
hlfan: I shot myself in the foot by not testing that, didn't I?

tomhughes (approved):

hlfan:
Did you just merge the old version in 283e034b562ed4786d9f9fc6020e2c61fd1dca02?

tomhughes:
Apparently I did, which is really weird because I retested it before I pushed and it seemed to be correct!

PR #1004, opened 2015-07-11
Adds key to cyclemap

emersion:
Related to #982.

Adds key to cyclemap.

WIP: not finished yet, I just want to know if I do it the right way :-P

PR #5926, opened 2025-04-12
CyclOSM map key for (inter)national/regional/local bike routes

AntonKhorev:
image
image

I also changed the label for these routes from "cycleway" to "bike route". "Bike route" is used in [the CyclOSM legend](https://www.cyclosm.org/legend.html). "Cycleway" comes from #1004. The problem is that these are not necessarily cycleways, they are routes that may or may not go over cycleways. And they are labeled along with actual cycleways.


tomhughes:
Looks good to me, thanks.

PR #5921, opened 2025-04-11
Test if history shows changesets at expected locations

AntonKhorev:
Test for #5917.

PR #5927, opened 2025-04-13
Wait for changeset link before clicking the map

AntonKhorev:
The test from #5921 might have a timing issue when the map is clicked before the chagesets are loaded. This PR fixes it by asserting that there's a link to the changeset in the sidebar, which should wait for things to load and the link to appear.


tomhughes:
Looks good to me, thanks.

PR #5930, opened 2025-04-13
Workaround a bug in file 5.46

tomhughes:
This refactors our trace file format detection so that we only do it in one and then adjusts that code to work around https://bugs.astron.com/view.php?id=643 which otherwise means that we fail to identify zip files on systems using version 5.46 of the file command.

### DIFF
--- a/app/models/trace.rb
+++ b/app/models/trace.rb
@@ -167,13 +167,12 @@ class Trace < ApplicationRecord
   end
 
   def xml_file
-    file.open do |tracefile|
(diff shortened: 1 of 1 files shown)


AntonKhorev:
Merged, thanks.

Although I didn't get osm-website fully working on Fedora 42 because I gave up on trying to compile native gems with gcc 15.

tomhughes:
Oh yes I hacked mkmf to allow some of the extensions to build by forcing `-std=c11` or something.

PR #5932, opened 2025-04-15
Add frozen_string_literal comments to ruby files

tomhughes:
Ruby 4 intends to make string literals frozen by default and to aid transition it is possible to opt into that behaviour on a per-file basis by adding a special comment to the top, as this PR does.

Note that ruby 3.4 already warns if you try and modify a non-frozen string literal and you have deprecation warnings enabled (which our testing framework does by default) so this will help us avoid introducing such warnings by effectively making them a hard error instead.


gravitystorm:
From https://bugs.ruby-lang.org/issues/20205

> And any application for which all dependencies have been made fully frozen string literal compatible can set RUBYOPT="--enable=frozen_string_literal" and start immediately removing magic comment from their codebase.

Instead of adding the comment to every file (and then removing them all again in ruby 4) could we add the RUBYOPT to CI? That way CI would fail if any string mutation is added in future PRs.

tomhughes:
The problem with that is that it will also affect any gems we use so if any of them try to modify a constant string (for example https://github.com/chef/ffi-libarchive/pull/90) then what is currently a deprecation warning will become a test failure.

tomhughes:
To explain what made me suggest this upgrading to Fedora 42 and hence ruby 3.4 led to quite a lot of deprecation warnings almost all of which (apart from the ffi-libarchive ones) come down to controller tests which make a POST request with a constant string as the body - the eventual analysis of what's going on is https://github.com/rack/rack/pull/2322#issuecomment-2802767710 and the easiest workaround pending solution in Rack and/or StringIO was to freeze strings ourselves.

larouxn:
I second this change. It's been best practice to add these `frozen_string_literal` comments for a little while now as a way of preparing for Ruby's eventual enforced string freezing. Even supposing we can remove the comments in Ruby 4, at least we'll know the codebase is ready. Plus, we'll probably able to auto-remove them via RuboCop ( same as how we're adding them here, nice :clap: ) or a clever global search and replace since they're always the first line followed by a newline.

:wave: @tomhughes, feel free to let me know if there's anything I can do to help push this forward. :rocket:

gravitystorm:
Thanks @tomhughes , merged.

I've been trying to avoid making changes to 800+ files since the whole frozen-string thing was first mentioned in the ruby2.x era, hoping that we could find a different approach. However, I can't see a better alternative now that ruby 3.4 is generating copious warnings. If all our deps were safe (which they aren't) and if there was a straightfoward way to pass the `--enable=frozen_string_literal` option to just affected versions of ruby (i.e. 3.4+) for every possible way a dev might try to run our code (which there isn't), then I would have chosen that instead.

I'm already looking forward to removing these pragmas at some point in due course! Here's hoping that upstream ruby don't make the transition drag out for too much longer.

PR #5936, opened 2025-04-16
Change highlighted bbox border width to 4px on history pages

AntonKhorev:
This is https://github.com/openstreetmap/openstreetmap-website/pull/5866#issuecomment-2764695781 for borders. You want horizontal/vertical line boundaries to coincide with pixel boundaries. If the line is drawn at integer coordinates, it means that its width has to be an even number. The line shape boundaries are width/2 to one side and width/2 to another and you want the original coords +- width/2 to still be integers.

Before, notice the blurry changeset bbox:
image

This will be noticeable on 1x pixel density devices, and technically on 3x or other odd-numbered density, although physical pixels probably get too small to notice.

After:
image

### DIFF
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -16,7 +16,7 @@ OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
     }
 
     return {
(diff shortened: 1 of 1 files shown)


nenad-vujicic (approved):

tomhughes:
Looks fine to me, thanks.

PR #5937, opened 2025-04-16
Switch to rails 8.0 defaults

tomhughes:
There doesn't seem to be anything in the new defaults that is likely to cause us a problem.

* `to_time_preserves_timezone` should have no effect as we always work with UTC times, and the production servers run in UTC time
* `strict_freshness` at worse could mean sending fewer 304 responses but in practice I don't think we have anything much served by rails where conditional gets do anything
* hopefully we don't have any regexps that are slow enough to be an issue


rkoeze (approved):
Looks good!

mmd-osm:
Although I couldn't figure out how "to_time_preserves_timezone = :zone" works, or what "receiver" the documentation talks about, I didn't encounter any obvious issues on a non UTC-time machine. In particular, test cases were working fine.

tomhughes:
The receiver is the object you're calling the method on.

AntonKhorev:
Merged, thanks.

I don't think we use `to_time` directly.

PR #5938, opened 2025-04-17
Inline SVGs with inline_svg gem

### LINKED ISSUE #5880: Can we avoid inlining svgs into erb templates?
I've seen a lot of PRs recently where we are inlining SVGs into the output html, rather than having them as sprites or standalone images. I can see the benefit of doing so, particularly for the ability to set `stroke="currentColor"`.

However, there are some downsides of putting them in the erb templates. It makes the icons hard to edit, for example with tools like Inkscape (without a lot of copy+pasting back and forward). For some erb templates it can also become a large amount of effectively "binary" data on show - it's not really possible to read the `path` and figure out what it shows or how to edit it, and so it's not much different than having inline PNG data.

Is there another approach that we can use, that allows the SVGs to be inlined in the output, but kept as separate SVG files in the source code? Perhaps some `inline_svg` helpers that can read the files, output either the desired paths or the whole file, and that makes any changes required to convert the input file into what we want to output (e.g. changing the stroke colour) automatically?

@hlfan @AntonKhorev you both have the most experience with these, is there something that I haven't considered?

hlfan:
### Description
Using the inline_svg gem, this PR resolves #5880.
### How has this been tested?
Verifying that the local Rails output matches


AntonKhorev:
Do you want both optimizing svg by micromanaging their attributes and externalizing them *as effectively "binary" data*? How is that supposed to work at the same time?

hlfan:
Neither.
You already optimized the SVGs when you added them into the templates[^1]. The additional `xmlns` is only for increased compatibility during development.
Pulling the SVGs out increases the visualizability (file thumbnails you wouldn't get when they're packed into templates) and supports the _separation of concerns_ raised in the original issue.

Just because _you_ understand both SVG internals and layout doesn't mean every contributor should have to.
By keeping them separated, layout edits don’t require SVG expertise, and SVG tweaks won’t accidentally break the layout.

[^1]: see, e.g., the original [new.svg](https://github.com/openstreetmap/openstreetmap-website/blob/f9db802c7b2f8ee9582cb2c9f7196d6abafdfced/app/assets/images/new.svg)

hlfan:
Is the intent clearer with the partial-like style?

tomhughes:
Looks good to me, thanks.

PR #5940, opened 2025-04-19
Changeset highlights in a layer on top of normal bboxes

AntonKhorev:
A fix for ["Also, sometimes the hover fails and the bbox ends up not on top of other bboxes"](https://community.openstreetmap.org/t/better-osm-org-a-script-that-adds-useful-little-things-to-osm-org/121670/105), look at the animation of the lower bbox in the second image.

Sometimes there's a bunch of changesets on the same place. When one of them is highlighted, the orange bbox is visible as intended if it happens to be rendered above others:
image

If it's below, the highlight is less visible:
image

This PR moves the highlight to another layer rendered on top of normal bboxes. The highlight is non-interactve to avoid interfering with hover events on bboxes below.
image
image

### DIFF
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -1,6 +1,4 @@
 OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
-  _changesets: new Map,
-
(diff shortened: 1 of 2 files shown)


rkoeze (approved):
Looks good to me 👍

PR #5942, opened 2025-04-21
Make teaspoon UI usable

hlfan:
The teaspoon UI at localhost:3000/teaspoon currently gets broken because the entire DOM is discarded in tests.
This alters the tests to make the UI usable (although a leftover _osm_location cookie can cause tests to fail in there).

### DIFF
--- a/test/javascripts/osm_test.js
+++ b/test/javascripts/osm_test.js
@@ -220,14 +220,15 @@ describe("OSM", function () {
 
   describe(".locationCookie", function () {
     it("creates a location cookie value", function () {
-      $("body").html($("<div id='map'>"));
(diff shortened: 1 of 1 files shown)


tomhughes:
I didn't even know teaspoon had a UI... How do you use it?

hlfan:
idk, I only stumbled upon that in the rails error during db migration.
Also maybe calling it a UI is a bit much, but the dashboard should work out of the box (apart from it being replaced by the tests)

tomhughes:
Managed to figure it out now - you have to start a server with `RAILS_ENV=test bundle exec rails server` and then visit `/teaspoon` to access it.

This does indeed seem to make things work, so it looks good to me.

Issue #5646, opened 2025-02-10
Add support for json response for `/api/0.6/user/gpx_files` endpoint


HarelM:
### Problem

Client side javascript parsing for xml isn't great (although possible).

### Description

It would be great to have json support for the `/api/0.6/user/gpx_files` endpoint like other supported json responses.

### Screenshots

N/A.

Satya900:
i have generated a pr for this issue   [LINK for PR](https://github.com/openstreetmap/openstreetmap-website/pull/5779)

HarelM:
Thanks for looking into this @Satya900!
You can link the issue from the PR as well to automatically close this issue when the PR is merged (by adding a comment on the PR main message with "resolves #xxx"

Satya900:
> Thanks for looking into this [@Satya900](https://github.com/Satya900)! You can link the issue from the PR as well to automatically close this issue when the PR is merged (by adding a comment on the PR main message with "resolves #xxx"
Done as per your instructions

Issue #3107, opened 2021-02-18
[💡] Documenting API using Swagger / rswag 


mmd-osm:
Documenting APIs on the WIki (https://wiki.openstreetmap.org/wiki/API_v0.6) has a few issues: it isn't maintained consistently, it's not considered authoritative by any means, and not easily consumable.

I looked at [`rswag`](https://github.com/rswag/rswag) as a way to include API documentation right inside this repo, along with a nice UI to test endpoints and a much easier way for API consumers to generate their own code based on the OAS3 spec that comes along with it.

I'm just posting a few screenshots how this could look like. It's mostly auto-generated content that can be enhanced by additional details/descriptions/error status code lists/examples, etc. Maybe this is even good enough to get rid of the Wiki page altogether and direct developers to the swagger docs instead.

#### A list of API 0.6 endpoints, as served via  http://localhost:3000/api-docs/index.html

![sw1](https://user-images.githubusercontent.com/5842757/108423896-bf690d00-7238-11eb-9156-fd72064b5cb8.png)

#### Testing the /api/capabilities endpoints in the browser

![sw2](https://user-images.githubusercontent.com/5842757/108423981-da3b8180-7238-11eb-8873-7044cff993a0.png)

simonpoole:
Being the victim of absolute hopeless swagger generated documentation before (for example maproulette), it is important to note that it is just a system to semi-automatically generate good-looking documentation, not good documentation. 

The later still requires significant effort regardless of the system. Given that 

- the rails port has essentially no code documentation at all
- moving the API documentation from the wiki to the rails port would put the onus on the current maintainers (but see above)
- and significantly raise the bar for contributing to it

I suspect that however well intentioned this is, it basically boils down to not documenting actual current behaviour of the API at all.

The other aspect is that, while a matter of contention among the maintainers and original rails-port devs, you could argue that the API should be specific implementation independent, and as a consequence as a matter of principle the documentation should be independent of any such implementation.

tomhughes:
Yes I'm not convinced about the wisdom of documenting the API in this repository when most of it (at least as used in production) isn't actually implemented here.

mmd-osm:
> moving the API documentation from the wiki to the rails port would put the onus on the current maintainers (but see above)

[...]

PR #5943, opened 2025-04-21
Add support for json traces API

HarelM:
### Description

This adds support for json for traces API (both for `index` and `show`).
This PR is to replace #5779.

To copy from the PR initial post there:
This PR adds JSON format support to the user's GPX files API endpoint, enhancing the API's flexibility and consistency with other endpoints that already support JSON responses. The implementation:

Adds JSON format handling to the traces controller (index, show)
Creates a JSON view template matching the XML structure
Ensures proper content type headers are set
This enhancement improves client-side parsing efficiency while maintaining backward compatibility with the existing XML format.
Resolves https://github.com/openstreetmap/openstreetmap-website/issues/5646

### How has this been tested?
Unfortunately, I was not able to test this.

Is there a more comprehensive guide as to how to setup a development environment? I've looked at what the CI does, mainly for the docker build and was able to boot up and instance, but I wasn't sure how I can create users there and if the mail stuff will work and how, so I was not able to test it properly.


app/controllers/api/traces_controller.rb
```diff
@@ -12,7 +12,14 @@ class TracesController < ApiController
     def show
       @trace = Trace.visible.find(params[:id])
 
-      head :forbidden unless @trace.public? || @trace.user == current_user
+      if @trace.public? || @trace.user == current_user
+        respond_to do |format|
+          format.xml { render :content_type => "application/xml" }
+          format.json { render :content_type => "application/json" }
+        end
```
tomhughes, line 19: I don't think you need the explicit content type here. I think this will suffice:
```suggestion
        respond_to do |format|
          format.xml
          format.json
        end
```

app/controllers/api/users/traces_controller.rb
```diff
@@ -7,7 +7,10 @@ class TracesController < ApiController
 
       def index
         @traces = current_user.traces.reload
-        render :content_type => "application/xml"
+        respond_to do |format|
+          format.xml { render :content_type => "application/xml" }
+          format.json { render :content_type => "application/json" }
+        end
```
tomhughes, line 13: Same goes here.

tomhughes (changes requested):
This will definitely need some tests - basically cloning the existing XML tests and adjusting them for JSON would be a good start.

Does CONTRIBUTING.md not explain what you need to know?

HarelM:
Thanks for the review, I'll look into addressing the issues.

app/controllers/api/users/traces_controller.rb
HarelM: Done in: 42e841c31dc9b006b0fe1a6ba7eaa4413b9da5f3

HarelM:
I've added some tests according to what I saw in other controllers' tests.
Let me know if more tests are needed.

As to contributing, I think it's missing some explanation of default users that are automatically created and how to login to those (like the nobody user). And how to create users in case those are needed.

How to boot the app and see that everything is working as expected might need to be a bit more explicitly written I think.

I can open a PR against the contributing.md for what I think might help future developers if that might be of interest.

app/controllers/api/users/traces_controller.rb
mmd-osm: Currently, test cases seem to be failing with `Expected response to be a <2XX: success>, but was a <406: Not Acceptable>`

I think we would need something like  `before_action :set_request_formats` to set XML as default format, similar to what we're doing for element endpoints (node/way/relation).

tomhughes:
There aren't any default users created - there's some information in CONFIGURE.md on setting up at initial user.

tomhughes:
I think you need to change the two controllers to add something like this at the top:

```ruby
before_action :set_request_formats, :only => [:index]
```

That should stop those test failures on the XML version.

Please also squash the fixup commits into the one they are fixing (so commits 2-4 into commit 1 and commit 6 into commit 5).

HarelM:
Ah, I wasn't expecting this file to contain this information, cool, thanks, looks like this is what I was looking for.

HarelM:
Thanks for the info regarding the `before_action`. I'll see how I can improve the git history once everything is passing.
BTW, why not simply use squash merge?

tomhughes:
I prefer to use merge commits so we record what PR changes have come from - feel free to squash everything together if you prefer but having the tests in a separate commit is also fine we just prefer not to have fixups.

HarelM:
Sure, I'll try and squash everything to a single commit, hopefully easier...

HarelM:
Ok. Everything should be in a single commit now.

app/views/api/traces/show.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+json.osm do
+  json.partial! "api/traces/trace", trace => @trace
```
tomhughes, line 2:
```suggestion
  json.partial! "api/traces/trace", :trace => @trace
```

app/views/api/users/traces/index.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+json.gpx_files @traces do |trace|
+  json.partial! "api/traces/trace", trace => trace
```
tomhughes, line 2:
```suggestion
  json.partial! "api/traces/trace", :trace => trace
```

test/controllers/api/traces_controller_test.rb
```diff
@@ -44,11 +48,21 @@ def test_show
...
+
+      # We should be able to do it with the owner of the trace with json format
+      auth_header = bearer_authorization_header public_trace_file.user
+      get api_trace_path(public_trace_file, :format => "json"), :headers => auth_header
+      assert_response :success
+      assert_equal "application/json", response.media_type
+      js = ActiveSupport::JSON.decode(@response.body)
+      assert_not_nil js
+      assert_equal public_trace_file.id, js.id
+      assert_equal public_trace_file.user.id, js.uid
```
tomhughes, line 65:
```suggestion
      assert_equal public_trace_file.id, js["osm"]["id"]
      assert_equal public_trace_file.user.id, js["osm"]["uid"]
```

HarelM:
These cycles are too long, I'll work on trying to run this locally and make sure the tests pass...

test/controllers/api/users/traces_controller_test.rb
```diff
@@ -34,6 +38,16 @@ def test_index
...
+
+        # check that we get a response when logged in with json
+        auth_header = bearer_authorization_header user, :scopes => %w[read_gpx]
+        get api_user_traces_path(:format => "json"), :headers => auth_header
+        assert_response :success
+        assert_equal "application/json", response.media_type
+        js = ActiveSupport::JSON.decode(@response.body)
+        assert_not_nil js
+        assert_equal trace1.id, js[0]["id"]
+        assert_equal trace2.id, js[1]["id"]
```
tomhughes, line 50:
```suggestion
        assert_equal trace1.id, js["gpx_files"][0]["id"]
        assert_equal trace2.id, js["gpx_files"][1]["id"]
```

tomhughes (changes requested):
The changes I've suggested fix the failing tests, but they also reveal a problem as they show that the structure of the JSON returned by the two methods is not consistent - one has a top level hash with a `osm` member pointing at the details and the other has a top level hash with a `gpx_files` member pointing at an array of details.

Looking at other API methods the responses should probably look like:

```json
{
  "version": "0.6",
  "trace": {
    ...
  }
}
```
and:

```json
{
  "version": "0.6",
  "traces": [
    {
      "trace": {
        ...
      }
    },
    ...
  ]
}
```

tomhughes:
The top level boilerplate can be added with:

```ruby
json.partial! "api/root_attributes"
```
which you'll find in other json.builder views.

HarelM:
Thanks for the info! I'll look into aligning those so that it will return in a similar way.

HarelM:
I've changed the structure according to what suggested.
Sorry for being a noob :-/
Unfortunately, when running the docker CI code on my mac not all the tests pass, so I'm having a hard time running the tests locally before I commit.
I've enabled the GitHub actions on my fork, so I'll be able to see the failing tests there.

HarelM:
I kept the traces as and array of objects, each one of them a trace, but without the `{ trace: { ... } }`, was your comment about including the `trace` property for every trace in the array intentional?

Regardless, this is now ready for review, the tests are passing in my actions in the fork and I squashed everything to a single commit.

Let me know what else is missing, or if more changes are needed.
Thanks!

app/controllers/api/traces_controller.rb
```diff
@@ -12,7 +13,14 @@ class TracesController < ApiController
...
 
-      head :forbidden unless @trace.public? || @trace.user == current_user
+      if @trace.public? || @trace.user == current_user
+        respond_to do |format|
+          format.xml
+          format.json
+        end
+      else
+        head :forbidden
+      end
```
hlfan, line 23:
```suggestion
      return head :forbidden unless @trace.public? || @trace.user == current_user

      respond_to do |format|
        format.xml
        format.json
      end
```
I'd keep the `head :forbidden unless` guard clause from before

test/controllers/api/users/traces_controller_test.rb
```diff
@@ -34,6 +38,16 @@ def test_index
...
+
+        # check that we get a response when logged in with json
+        auth_header = bearer_authorization_header user, :scopes => %w[read_gpx]
+        get api_user_traces_path(:format => "json"), :headers => auth_header
+        assert_response :success
+        assert_equal "application/json", response.media_type
+        js = ActiveSupport::JSON.decode(@response.body)
+        assert_not_nil js
+        assert_equal trace1.id, js["traces"][0]["id"]
+        assert_equal trace2.id, js["traces"][1]["id"]
```
hlfan, line 50: I'd like to see the tags being checked like for the XML here too

hlfan (approved):
Looks good, only noted some minor ideas

test/controllers/api/users/traces_controller_test.rb
HarelM: Added checks. 
Let me know if anything is missing.

hlfan (approved):

tomhughes (approved):
Thanks for all this work - this looks good to me now.

I did push a small change to tidy up the commit message and I'll merge it now.

mmd-osm:
@HarelM : It would be nice if you could document your API change on the API 0.6 wiki page ( https://wiki.openstreetmap.org/wiki/API_v0.6 ) to help others developers.

You can find a bit of change history in the header section. Furthermore, you could write a few words in the respective endpoint section, maybe including sample payloads for the new JSON format.

HarelM:
I wanted to see that it is deployed first, but sure, I've updated the wiki page.
As a side note, wouldn't it be easier to simply generate an openAPI/swagger page that is generated from the code itself to avoid mismatches?

mmd-osm:
Thanks! By the way, I've proposed swagger / openapi a few years ago in #3107...

PR #5946, opened 2025-04-22
Prefer new style hash syntax in config files

rkoeze:
### Description
Ran `bin/rails app:update` to get an idea of what our configuration files looked like relative to the 8.0 defaults. There were a couple places where the diff boiled down to the old versus new hash syntax. Figured I'd update those so the next time anyone runs `bin/rails app:update` they don't get picked up in the diff. Or in the case of `test.rb`, the `Settings.server_url` value is the only difference.


rkoeze:
And now I see that we have a cop for preferring hash rocket syntax! Closing this out.

PR #5951, opened 2025-04-25
Remove unneeded and broken bootstrap close button theme overrides

### LINKED ISSUE #5947: Invisible "X" button when using dark mode preference
### URL

https://www.openstreetmap.org/node/8043886647 (or any place where an "X" button is visible)

1. Set  "Preferred Website Color Scheme" to "Auto" in https://www.openstreetmap.org/preferences and, make sure that your OS/browser is set to prefer dark theme (or use the option in the dev tools)
2. Open https://www.openstreetmap.org/node/8043886647 (notice that the X button in the side panel is light grey)
3. Go to OSM's preferences and set  "Preferred Website Color Scheme" to "Dark"
4. Open https://www.openstreetmap.org/node/8043886647 (notice that the X button in the side panel is completely black)

|Auto and prefer dark | Dark and prefer dark |
|--|--|
| Image | Image |

hlfan:
### Description
Fixes #5947 by removing the broken reference to `--bs-btn-close-white-filter` and also the `filter: none` because both are already taken care of by Bootstrap's definition of `--bs-btn-close-filter`.

### How has this been tested?
dev tools

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -82,16 +82,6 @@ time[title] {
   100% { opacity: 1 }
 }
 
-/* Bootstrap close button overrides for nested light/dark themes */
-
-[data-bs-theme="dark"] .btn-close {
-  filter: var(--bs-btn-close-white-filter);
-}
-
-[data-bs-theme="light"] .btn-close {
-  filter: none;
-}
-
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #5952, opened 2025-04-26
Update fix for close button inside themed elements

AntonKhorev:
Update the close button light/dark theme fix from #4761 after Bootstrap v5.3.4 broke it. #5951 removed the fix as "unneeded".

Currently the fix is needed for *auto site theme in osm preferences* + *no system dark theme preference* + *dark banner theme*.

Black close button that should be white without the fix:
image

White close button with the fix:
image

Possible Bootstrap issue: https://github.com/twbs/bootstrap/issues/39481

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -84,12 +84,14 @@ time[title] {
 
 /* Bootstrap close button overrides for nested light/dark themes */
 
-[data-bs-theme="dark"] .btn-close {
-  filter: var(--bs-btn-close-white-filter);
+[data-bs-theme="dark"],
+[data-bs-theme] [data-bs-theme="dark"] {
+  --bs-btn-close-filter: invert(1) grayscale(100%) brightness(200%);
 }
 
-[data-bs-theme="light"] .btn-close {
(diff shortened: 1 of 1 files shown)


hlfan (changes requested):
Ok this implementation makes it way more clear where this is needed.
But setting `--bs-btn-close-filter` means you don't have to target `.btn-close` anymore.

app/assets/stylesheets/common.scss
```diff
@@ -82,6 +82,18 @@ time[title] {
   100% { opacity: 1 }
 }
 
+/* Bootstrap close button overrides for nested light/dark themes */
```
hlfan, line 85:
```suggestion
/* Bootstrap overrides for nested light/dark themes */
```

hlfan (approved):

app/assets/stylesheets/common.scss
AntonKhorev: But the overrides are for close button. If they affect anything else, that's not intentional.

tomhughes:
Looks good to me, thanks.

PR #5953, opened 2025-04-26
Fix svg icon path in manifest.json

AntonKhorev:
I don't think the svg icon added in #5735 ever worked.

Before:
image
See that the path is wrong.

After:
image
image


hlfan (approved):

tomhughes:
Looks good to me, thanks.

PR #5956, opened 2025-04-26
Add routing provider tips

hlfan:
Addresses a concern from #3123 about the routing engine providers not being labeled.
| image | image | image |
|-|-|-|

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -57,6 +57,8 @@ OSM.Directions = function (map) {
     OSM.router.route("/" + OSM.formatHash(map));
   });
 
+  $(".directions_form select + .input-group-text > span").tooltip();
+
   function setEngine(id) {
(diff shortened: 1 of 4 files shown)


HolgerJeromin:
This looks like a different info for each engine.

hlfan:
@HolgerJeromin I put the info circle (this time really a bootstrap icon svg) next to the select:
image
Is this better?

AntonKhorev:
Weren't we trying to remove unnecessary controls like the *Go* button?

hlfan:
But you still want a tooltip?

AntonKhorev:
Let's suppose I don't know anything about osm. I open the site, I see the button with two arrows, there's a tooltip on it that says "Find directions...". I click it and I see on the top a car, a bicycle and a person walking, they have toolips but I'll be able to guess what they mean without tooltips too. Then there's a dropdown that says "OSRM". What is OSRM?

PR #5957, opened 2025-04-26
Show warning when visiting login page while already logged in

### LINKED ISSUE #5441: Login page does not check if a user is logged in
### URL

https://www.openstreetmap.org/login?referer=%2Fedit%3Frelation%3D15090094#map=19/33.411025/-114.923289

1.  Open osm.org in a container that is not normally used for osm.org (i.e. you are not logged in there)
2. Press "Edit" (you are redirected to the login page, some referrer value is added to the URL)
3. Change the container of the tab (right click tab -> "Open in new container tab" -> select the container where you are logged into OSM)
4. The login page is displayed even though you are logged in
 
The expected behavior would be to check if the user is logged in already and if he is, redirect him to the `referer` or to the main page. 

The bug appears in Firefox when using [multi-account-containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers/). I don't know if it's reproducible in other browsers.

image

AntonKhorev:
Fixes #5441 by showing a warning as described in https://github.com/openstreetmap/openstreetmap-website/issues/5441#issuecomment-2563121432.

This is what you'll see if you visit the login page while already logged in:
image

Additionally there's going to be a link (styled as button) if there's a referer parameter and it's a "safe" referer (according to the `safe_referer` method of `ApplicationController`; the check looks a bit silly because it has to satisfy both RuboCop and Brakeman):
image

In the scenario described in https://github.com/openstreetmap/openstreetmap-website/issues/5441#issue-2758188137 that link should open the editor.

### DIFF
--- a/app/assets/javascripts/login.js
+++ b/app/assets/javascripts/login.js
@@ -2,5 +2,6 @@ $(function () {
   // Preserve location hash in referer
   if (location.hash) {
     $("#referer").val($("#referer").val() + location.hash);
+    $("#referer_link").prop("hash", location.hash);
   }
 });
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -16,6 +16,9 @@ class SessionsController < ApplicationController
   def new
     referer = safe_referer(params[:referer]) if params[:referer]
 
+    @safe_referer = referer
+    @safe_referer = nil if referer != params[:referer]
+
     parse_oauth_referer referer
   end
 
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -37,7 +37,7 @@
         <%= t ".access_another_page" %>
       </p>
       <p class="text-center">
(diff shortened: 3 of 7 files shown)


app/assets/javascripts/login.js
```diff
@@ -2,5 +2,6 @@ $(function () {
   // Preserve location hash in referer
   if (location.hash) {
     $("#referer").val($("#referer").val() + location.hash);
+    $("#referer_link").prop("href", $("#referer_link").prop("href") + location.hash);
```
hlfan, line 5: The insistence on using jQuery reduces the legibility:
``` suggestion
    $("#referer")[0].value += location.hash;
    $("#referer_link")[0].hash = location.hash;
```

AntonKhorev: `$("#referer_link")[0].hash = location.hash;` requires the link to exist and we're not checking that.

Changed to `$("#referer_link").prop("hash", location.hash);`.

tomhughes:
Looks great to me - thanks!

PR #5960, opened 2025-04-27
Remove routing submit button

hlfan:
Reorganises the routing form according to proposals D/E of https://github.com/openstreetmap/openstreetmap-website/issues/3123#issuecomment-803143179.

localhost_3000_

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -114,7 +114,8 @@ OSM.Directions = function (map) {
       if (fitRoute) {
         routeOutput.fit();
       }
-    }).catch(async function () {
+    }).catch(async function (error) {
+      if (error.name === "AbortError") return;
       await sidebarLoaded();
       routeOutput.remove();
       if (reportErrors) {
--- a/app/views/layouts/_search.html.erb
+++ b/app/views/layouts/_search.html.erb
@@ -38,28 +38,6 @@
     </div>
 
     <div class="d-flex flex-column gap-2">
-      <div class="d-flex gap-1 align-items-center">
-        <div class="d-flex flex-column gap-1 flex-grow-1">
-          <div class="d-flex gap-2 align-items-center">
-            <div class="routing_marker_column position-absolute">
-              <%= image_tag "marker-green.png", :class => "img-fluid", :data => { :type => "from" }, :draggable => "true" %>
-            </div>
-            <%= text_field_tag "route_from", params[:from], :placeholder => t("site.search.from"), :autocomplete => "on", :class => "form-control py-1 px-2 ps-4", :dir => "auto" %>
-          </div>
-          <div class="d-flex gap-2 align-items-center">
-            <div class="routing_marker_column position-absolute">
-              <%= image_tag "marker-red.png", :class => "img-fluid", :data => { :type => "to" }, :draggable => "true" %>
-            </div>
-            <%= text_field_tag "route_to", params[:to], :placeholder => t("site.search.to"), :autocomplete => "on", :class => "form-control py-1 px-2 ps-4", :dir => "auto" %>
-          </div>
-        </div>
-        <button type="button" class="reverse_directions btn btn-outline-secondary border-0 p-2 me-n1" title="<%= t("site.search.reverse_directions_text") %>">
-          <svg class="d-block" width="22" height="22" viewBox="-11 -11 22 22" fill="none" stroke="currentColor" stroke-width="2">
-            <path d="m-4 -2 0 10 m-4 -4 4 4 4 -4" />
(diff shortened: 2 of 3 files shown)


tomhughes (changes requested):
There's some sort of race condition in the new test here that causes it to sometimes pass and sometimes fail.

hlfan:
Is there a function like `click_on` that I can call by class name instead of by text content?

hlfan:
Is there a function like `click_on` that I can call by class name instead of by text content?

app/assets/stylesheets/common.scss
```diff
@@ -582,6 +582,13 @@ header .search_forms,
   }
 }
 
+.directions_form {
+  div {
+    align-items: center;
+    grid-template-columns: 1fr auto;
+  }
+}
+
```
AntonKhorev, line 591: Imagine @gravitystorm appearing here and telling all of what he told in #3620 about *custom grid*.

hlfan: I've tried to build this grid layout with display flex (in the `d-flex` and `row`/`col` flavours).
In neither can I get the close button to be in just the right place, aligned with the horizontal center of the mode and provider selectors, _and_ the vertical center of the reverse directions button and the submit button when focused.

While I've gotten it working in an HTML table (Bootstrap `.table` messed up too much), that feels like the wrong direction to move markup-wise. And the markup is quite clunky:
```html
<table class="w-100">
	<tr>
		<td class="pb-1 pe-1">
			<div class="align-items-center d-flex gap-2">
				<div class="btn-group routing_modes" role="group">...</div>
				<select class="..." name="routing_engines">...</select>
			</div>
		</td>
		<td class="pb-1">
			<div class="align-items-center d-flex flex-column gap-2">
				<button type="button" class="btn-close p-2 rounded-5" aria-label="Close"/>
			</div>
		</td>
	</tr>
	<tr>
		<td class="pe-1">
			<div class="d-flex flex-column gap-1">
				<div class="align-items-center d-flex">...</div>
				<div class="align-items-center d-flex">...</div>
			</div>
		</td>
		<td class="col-1">
			<div class="align-items-center d-flex flex-column gap-2">
				<button type="button" class="..." title="Reverse Directions">...</button>
				<input type="submit" name="commit" value="Go" class="...">
			</div>
		</td>
	</tr>
</table>
```

AntonKhorev:
Why do we need the *Go* button at all? One scenario is when the user closes the directions panel by clicking this close button: 
image

Why would the user do that? Maybe they're on a slow connection and don't want extra route requests while they're moving the markers around and changing the routing options.[^1] When they're done with markers and options, they can click *Go* and make the directions panel reappear. So *Go* = *Show Directions (again)*. In this case it makes sense for this button to be where the panel is supposed to be, that is below the form. And if the form was closed in the scenario described in this paragraph, the button should be visible because nobody would figure out they need to tab to it.

But we wanted to hide the button, right? We can hide it if one of these is true:
- one or both endpoints are not ready
- the directions panel is visible

Then normally users won't see the *Go* button. And when they'll see it, it's because the directions are not shown; in this case it doesn't take away the screen space from the direcions.

Showing *Go* below the form avoids the problems of having to squeeze it into the side and resizing/moving other inputs/buttons to make space for it. We don't even know how wide *Go* is because of i18n.

[^1]: Currently the panel reappears when the options are changed, but it probably shouldn't.

tomhughes:
I certainly don't understand the point of having the button present but invisible so that it can only be navigated to with tab - what is supposed to be the use case for that?

hlfan:
> I certainly don't understand the point of having the button present but invisible so that it can only be navigated to with tab - what is supposed to be the use case for that?

I thought it might be better for accessibility to have a dedicated submit button — where there's no defined "inputleave event" triggering the callback (see the test I needed to rewrite) — but for now, I'm leaving the `d-none` instead of the `visually-hidden` class there as it _shouldn't_ be a problem even for that case.

tomhughes:
I mean I was really thinking why does there need to be a submit element at all?

Then again I don't really like what you've had to do to the test... I mean testing that the reverse direction regenerates the route is fine but we should also test that a route is generated when you enter both endpoints?

hlfan:
If capybara really doesn't trigger the onchange event after filling in the input, then some other interaction is needed for the routing to run.
You already saw that even with re-selecting the first input,
> There's some sort of race condition in the new test here that causes it to sometimes pass and sometimes fail.

This should have been fixed years ago by what I could find, maybe there's a simple solution that has evaded me...

I'm on board with more tests when that fill_in thing is sorted out.

hlfan:
Are the tests OK now @tomhughes?

tomhughes (approved):

PR #5962, opened 2025-04-29
Limit number of directions endpoint geocoding results to 1

AntonKhorev:
Geocoding for route endpoints requests the default number of results, which is 10, but only uses the first one.


nenad-vujicic (approved):
This looks and works great on my side! Good catch, thanks!

tomhughes:
Looks good to me, thanks.

PR #5963, opened 2025-04-29
Split interactive history layer into areas and borders above areas

AntonKhorev:
The goal is to have the following z-order:
1. areas of changeset bboxes outside the sidebar viewport
2. areas of changeset bboxes inside the sidebar viewport
3. borders of changeset bboxes outside the sidebar viewport
4. borders of changeset bboxes inside the sidebar viewport

This allows changesets inside the viewport to be selectable from a big pile of bboxes in one place and not completely prevent selecting smaller outside bboxes.

Here's a larger in-viewport bbox on top of the smaller out-of-viewport bboxes, even with the pointer inside one of these smaller bboxes:
image

But when the pointer is above a border of a smaller bbox, that bbox is highlighted:
image

The principle is that you should be able to select the things you see before anything else, and you actually see only the borders, not the areas inside them. There could be a larger area, possibly covering the entire screen, an if it gets sorted on top, you won't be able to select anything below it. That's why bboxes were sorted by size. But the size-sorting alone can bury the in-viewport bboxes under others. This PR tries to give priority to both smaller bboxes and in-viewport bboxes, possibly addressing complaints like https://github.com/openstreetmap/openstreetmap-website/pull/5924#issuecomment-2815496727. I may still try to remove bboxes covering the entire screen completely from the map later, if I'm able to come up how to show these changesets in some other manner.

### DIFF
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -9,11 +9,19 @@ OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
     }
   },
 
-  _getInteractiveStyle: function (changeset) {
+  _getAreaStyle: function (changeset) {
+    return {
+      weight: 0,
+      fillOpacity: 0,
+      className: this._getSidebarRelativeClassName(changeset)
+    };
+  },
+
+  _getBorderStyle: function (changeset) {
     return {
       weight: 2,
       color: "var(--changeset-border-color)",
-      fillOpacity: 0,
+      fill: false,
       className: this._getSidebarRelativeClassName(changeset)
     };
   },
@@ -73,7 +81,8 @@ OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
         changesetSouthWest.lng -= shiftInWorldCircumferences * 360;
         changesetNorthEast.lng -= shiftInWorldCircumferences * 360;
 
-        this._interactiveLayer.getLayer(changeset.id)?.setBounds(changeset.bounds);
+        this._areaLayer.getLayer(changeset.id)?.setBounds(changeset.bounds);
+        this._borderLayer.getLayer(changeset.id)?.setBounds(changeset.bounds);
         this._highlightLayer.getLayer(changeset.id)?.setBounds(changeset.bounds);
       }
     }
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/index/history-changesets-layer.js
```diff
@@ -89,13 +98,20 @@ OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
...
 
     for (const changeset of this._changesets.values()) {
-      const rect = L.rectangle(changeset.bounds, this._getInteractiveStyle(changeset));
+      const rect = L.rectangle(changeset.bounds, this._getAreaStyle(changeset));
+      rect.id = changeset.id;
+      rect.addTo(this._areaLayer);
+    }
+
+    for (const changeset of this._changesets.values()) {
+      const rect = L.rectangle(changeset.bounds, this._getBorderStyle(changeset));
```
tomhughes, line 112: Can we not add the changeset to both layer groups in the same loop rather than doing two passes?

AntonKhorev: We can't because the order of `<path>` elements inside Leaflet's svg layer depends on the order of `addTo` calls. Even if the rectangles are added to different feture groups, their `<path>`s end up in the same `<svg>` element.

The alternative is to put feature grous in different [panes](https://leafletjs.com/reference.html#map-pane), but that becomes more complicated, and I'll need even more panes [later](https://github.com/openstreetmap/openstreetmap-website/pull/5964).

hlfan: But at that stage you don't need the changeset reference anymore.

```javascript
const changesetAreas = [];
const changesetBorders = [];
for (const changeset of this._changesets.values()) {
  const arect = L.rectangle(changeset.bounds, { className: this._getSidebarRelativeClassName(changeset)) });
  const brect = L.rectangle(changeset.bounds, { className: this._getSidebarRelativeClassName(changeset)) });
  arect.id = brect.id = changeset.id;
  changesetAreas.push(arect);
  changesetBorders.push(arect);
}
changesetAreas.forEach(a => this._areaLayer.addLayer(a));
changesetBorders.forEach(b => this._borderLayer.addLayer(b));
this._areaLayer.setStyle({
  weight: 0,
  fillOpacity: 0
});
this._borderLayer.setStyle({
  weight: 2,
  color: "var(--changeset-border-color)",
  fill: false
});
```

No get...Style wrappers around _getSidebarRelativeClassName needed.

hlfan (changes requested):

app/assets/javascripts/index/history-changesets-layer.js
tomhughes: > We can't because the order of `<path>` elements inside Leaflet's svg layer depends on the order of `addTo` calls. Even if the rectangles are added to different feture groups, their `<path>`s end up in the same `<svg>` element.

Oh right... I assumed that was the point of having separate layer groups so that it would order the groups in the order they were added. What's the point of splitting it into two groups then?

AntonKhorev: To get rectangles by changeset ids. There's a different rectangle for the same id in each group.

AntonKhorev:
@hlfan see #5964

app/assets/javascripts/index/history-changesets-layer.js
tomhughes: Ah right yes... That makes sense ;-)

tomhughes:

PR #5964, opened 2025-04-29
Add outlines for in-viewport changeset bboxes

AntonKhorev:
This PR adds outlines on history pages for:
- bboxes of changesets inside the sidebar viewport
- highlighted bboxes

This attempts to address [the *hard to find the edit* / *color noise* problem](https://community.openstreetmap.org/t/better-osm-org-a-script-that-adds-useful-little-things-to-osm-org/121670/106).

I tried different colors for outlines. For example, I tried [black](https://community.openstreetmap.org/t/better-osm-org-a-script-that-adds-useful-little-things-to-osm-org/121670/108) for hover highlights. That makes them more noticeable, but the color itself looks too out of place. So temporarily I'm keeping all of them white.

## In-viewport bboxes

When there's a lot of bboxes shown on the map, the in-viewport bboxes don't stand out enough from the rest:
image

With outlines they are more visible:
image

## Highlighted bboxes

This is how a group of out-of-viewport bboxes look without any of them highlighted:
image

They don't look much different when one of them is highlighted, and you can't easily say see which bbox it is:
image

But with an outline it's easier to see the highlighted bbox:
image

### DIFF
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -5,16 +5,8 @@ OSM.HistoryChangesetBboxLayer = L.FeatureGroup.extend({
 
   updateChangesetLayerBounds: function (changeset) {
     this.getLayer(changeset.id)?.setBounds(changeset.bounds);
-  }
-});
+  },
 
-OSM.HistoryChangesetBboxAreaLayer = OSM.HistoryChangesetBboxLayer.extend({});
-
-OSM.HistoryChangesetBboxBorderLayer = OSM.HistoryChangesetBboxLayer.extend({});
-
-OSM.HistoryChangesetBboxHighlightLayer = OSM.HistoryChangesetBboxLayer.extend({});
-
-OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
   _getSidebarRelativeClassName: function ({ sidebarRelativePosition }) {
     if (sidebarRelativePosition > 0) {
       return "changeset-above-sidebar-viewport";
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/common.scss
hlfan, line 1: Have you thought about a general changeset class so you don't need to define the outline color thrice and the fill color twice with the same color?

hlfan: Yeah, but I'm talking about the current state - when you change the colors later this outdated comment will be collapsed anyway.
Until then I'm still behind this suggestion, another nesting layer would then still make the selectors shorter: `&.in-sidebar-viewport`

app/assets/javascripts/index/history-changesets-layer.js
```diff
@@ -119,14 +119,12 @@ OSM.HistoryChangesetsLayer = L.FeatureGroup.extend({
...
     if (!changeset) return;
 
-    let highlightRect = this._highlightLayer.getLayer(id);
-    if (!state && highlightRect) {
-      this._highlightLayer.removeLayer(highlightRect);
-    }
-    if (state && !highlightRect) {
-      highlightRect = L.rectangle(changeset.bounds, this._getHighlightStyle(changeset));
+    if (state) {
+      const highlightRect = L.rectangle(changeset.bounds, this._getHighlightStyle(changeset));
```
tomhughes: So we're now relying on leaflet discard any pre-existing layer with the same ID when we add this one? Is that documented anywhere?

Or is the assumption just that we won't toggle a changeset on that is already on?

AntonKhorev: We are relying on:
- if the highlight layer is already added, it has correct geometry, style etc
- Leaflet [won't add a layer with the same id to the group](https://github.com/Leaflet/Leaflet/blob/c4a1e362bfd7f7709efdaff94f154e100a706129/src/layer/FeatureGroup.js#L29-L31), but looks like it's not documented unless [this counts as documented](https://github.com/Leaflet/Leaflet/pull/798)

AntonKhorev: Overriding `getLayerId` to check if the layer is added by id is also not documented, but @jfirebaugh made it work both here a56d1036d58d45777d6adeffd5a950519bc70318 and in Leaflet https://github.com/Leaflet/Leaflet/pull/1581.

tomhughes: I guess that's probably good enough.

tomhughes:
I think looks good, thanks.

PR #5965, opened 2025-04-30
API changeset close resource

AntonKhorev:
Continues #5801.

Creates changeset close resource that responds to `PUT /api/0.6/changeset/:changeset_id/close`, which doesn't change the API.

I think `PUT` is wrong here because it should be idempotent. If executed twice, the result should be the same. And it's the same, changeset is closed, except for the response status. The second response is `Conflict`, but shouldn't it be `OK`? I'm not changing this yet, if this needs fixing, the backwards-compatible fix is to make the route respond to `POST`.

### DIFF
--- a/app/abilities/api_ability.rb
+++ b/app/abilities/api_ability.rb
@@ -34,7 +34,7 @@ class ApiAbility
         can :read, :active_user_blocks_list if scopes.include?("read_prefs")
 
         if user.terms_agreed?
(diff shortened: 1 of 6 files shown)


mmd-osm:
In case you’re changing something regarding this endpoint, can you also create an issue for cgimap as well? I can’t commit this will be picked up in the near future, but at least it should be documented.

tomhughes:
Looks good to me and as you say any change to the semantics is a larger issue that would need to be discussed.

Issue #5959, opened 2025-04-27
Race condition when writing directions content


hlfan:
### URL

afaict only in development environments

### How to reproduce the issue?

1. Set your location with the dev tools if wanted/needed
2. Search in OSM for a place near the location
3. Activate the "show location" feature
4. Click the directions button to "Find directions between two points"

The output gets written immediately, since the endpoints are close.
Only later (since it's not cached) does Rails respond with the contents of `directions/search.html.erb`.
So, since the `#directions_content` div wasn't there when the instructions were meant to be written, it remains blank.

AntonKhorev:
We are not cancelling the requests correctly, see #5893.

hlfan:
Nowhere in here is a request aborted.

AntonKhorev:
Great, we have more bugs then. This one, buggy route request cancelling, and we don't even try to set up an abort controller for sidebar loading.

PR #5967, opened 2025-04-30
Unify directions page handler and parallelize requests

hlfan:
#5961's fix to #5959 was to delay starting the route request until the sidebar loaded, which isn't needed.

With this fix to #5959, I'm merging the page's pushstate/popstate with load and pulling out the sidebar requirement into a helper function also used in getRoute after the request, enabling the requests to be run in parallel.

This should remove half a second of extra latency for people with geostationary satellite internet.


hlfan:
I'm not in charge of deciding if that has to wait, but I don't see any big conflicts between the two PRs.

tomhughes:

PR #5968, opened 2025-04-30
Add geo URI handler registration on link interaction

hlfan:
For the geoURI handling to be available more broadly, on the first hover over the geoURI link of a session, try to register the handler. If not accepted, try again only in the next session.


app/assets/javascripts/leaflet.share.js
```diff
@@ -227,6 +229,16 @@ L.OSM.share = function (options) {
       const precision = 5 * Math.pow(10, Math.floor(Math.LOG10E * Math.log(scale)) - 2);
       return precision * Math.ceil(scale / precision);
     }
+
+    function askToHandleGeoURI() {
+      if (sessionStorage.getItem("geoHandlerAsked")) return;
+      try {
+        navigator.registerProtocolHandler("geo", "/?geouri=%s");
+        sessionStorage.setItem("geoHandlerAsked", "registered");
```
AntonKhorev: "registered" gets recorded even if the registration prompt is closed without registering anything.

app/assets/javascripts/leaflet.share.js
```diff
@@ -29,6 +29,8 @@ L.OSM.share = function (options) {
 
     $ui.find(".share-tab [id]").on("click", select);
 
+    $("#geo_uri").on("mouseover click focus", askToHandleGeoURI);
```
AntonKhorev, line 24: Is `mouseover` a good idea? I move my mouse pointer over the link and then this thing pops up on top, shifting everything down (Firefox):
image

I wouldn't want links moving away from my pointer.

app/assets/javascripts/leaflet.share.js
AntonKhorev: Actually it's recorded before the user even reacts to the prompt.

app/assets/javascripts/leaflet.share.js
AntonKhorev: `click` also has odd behavior. If `mouseover` is removed, the standard click handler wouldn't run after the *Add application* prompt appears and shifts everything down. On the other hand, getting both *Add application* and *Choose application* would also look strange:

image

app/assets/javascripts/leaflet.share.js
hlfan: Setting "asked" or "blocked" instead is ok?

AntonKhorev: If you have the key named `geoHandlerAsked` then I'd expect its value to be `yes` (or `no` but you probably don't need `no`).

app/assets/javascripts/leaflet.share.js
AntonKhorev: What do you think about adding a button to trigger the registration? I think that automatic registration prompt would get annoying no matter how it's implemented.

hlfan: I wanted to make it as non-intrusive as possible while not cluttering up the pane more. Also, afaik, a button couldn't trigger the registration for the webmanifest-variant. But the layout shift from Firefox makes this unacceptable...

A by-default hidden button would be nice, but there isn't a way to figure out if the site already has the handler registered.

app/assets/javascripts/leaflet.share.js
hlfan: Then what about `wasAbleToAskForHandler` with `yes`, `no`, and no value?

app/assets/javascripts/leaflet.share.js
```diff
@@ -226,6 +228,16 @@ L.OSM.share = function (options) {
...
       return precision * Math.ceil(scale / precision);
     }
+
+    function askToHandleGeoURI() {
+      if (sessionStorage.getItem("wasAbleToAskForHandler")) return;
+      try {
+        navigator.registerProtocolHandler("geo", "/?geouri=%s");
+        sessionStorage.setItem("wasAbleToAskForHandler", "yes");
+      } catch (e) {
+        sessionStorage.setItem("wasAbleToAskForHandler", "no");
```
pablobm: Why yes/no? I would have expected a simple boolean like `askedForHandler=true|false`. Are we expecting to use the "no" value at some point?

Also: should the key have a more specific name? Could there be other "handlers" in the future? I was thinking something like `askedForGeoProtocolHandler`.

hlfan: > Are we expecting to use the "no" value at some point?

Just to remember to only ask again in the next session.

app/assets/javascripts/leaflet.share.js
```diff
@@ -226,6 +228,16 @@ L.OSM.share = function (options) {
...
+      }
```
pablobm: How about this instead?
- It avoids handling the success and error cases separately.
- Uses a boolean instead of a "yes"/"no" string, which I think is more natural.
- Uses a more specific name, just in case we want to ask for other handlers in the future.

```suggestion
      if (sessionStorage.getItem("askedForGeoProtocolHandler")) return;
      try {
        navigator.registerProtocolHandler("geo", "/?geouri=%s");
      } finally {
        sessionStorage.setItem("askedForGeoProtocolHandler", true);
      }
```

app/assets/javascripts/leaflet.share.js
pablobm: Sorry, I mean why use yes/no instead of a boolean. Hang on, I'm going to do an "add a suggestion" instead to illustrate... (Closing this thread and picking up in a new one).

hlfan:
Since now over ⅞ of desktop users can register the protocol handler by installing the PWA, I'm closing this PR since making the UX unintrusive proved too hard due to the prompt behavior in Firefox and isn't worth the effort anymore.

PR #5969, opened 2025-04-30
Reorganize parseHash to avoid needing to destructure parseHash.center

hlfan:
The destructuring of parseHash.center makes using parseHash more unwieldy. Adding these parameters separately allows for easier integration elsewhere.


tomhughes:
Looks fine to me, thanks.

PR #5972, opened 2025-05-01
Move dashboard-specific javascript to dashboard.js

### LINKED ISSUE #5971: HomeLocationNameGeocoder gets called with empty inputs on dashboard
### URL

https://www.openstreetmap.org/dashboard

With a location set for the current user, load the page

Uncaught TypeError: Cannot read properties of undefined (reading 'trim')
    at Object.Endpoint [as HomeLocationNameGeocoder]
https://github.com/openstreetmap/openstreetmap-website/blob/4f52ba1bd324d26d34dcf132d39391d3f1744742/app/assets/javascripts/home_location_name-endpoint.js#L4
    at HTMLDocument.\<anonymous\>
https://github.com/openstreetmap/openstreetmap-website/blob/4f52ba1bd324d26d34dcf132d39391d3f1744742/app/assets/javascripts/user.js#L22

AntonKhorev:
Fixes #5971.

`user.js` does too many different things. It's better to take it apart and have smaller scripts included from all of the pages that currently include it.

This PR takes out dashboard-specific code from `user.js` and puth it into `dashboard.js`. This code doesn't include `OSM.HomeLocationNameGeocoder` which would fail on dashboard pages breaking the map:

image

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -1,0 +1,33 @@
+//= require leaflet.locate
+
+$(function () {
+  const defaultHomeZoom = 12;
+  let map;
+
+  if ($("#map").length) {
+    map = L.map("map", {
+      attributionControl: false,
+      zoomControl: false
+    }).addLayer(new L.OSM.Mapnik());
+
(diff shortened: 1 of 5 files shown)


tomhughes:
Looks good to me, thanks.

PR #5973, opened 2025-05-01
Implement JSON support for the changeset download API

### LINKED ISSUE #5970: JSON support for /changeset/:id/download
### Problem

- **As a** developer of web-based tools, 
- **I want** to avoid bundling a massive xml-parsing library just for a single API endpoint (since almost all others support JSON now)
- **so that** the page loads faster, we avoid unnecessary library, avoid boilerplate code, etc. etc.

### Description

Over the years, JSON support has gradually been added to most APIs.

Uploading & downloading changesets are one of the few APIs that are still XML-only. 

- https://github.com/zerebubuth/openstreetmap-cgimap/pull/407 tracks uploading changesets as JSON
- In this issue tracks downloading changesets in JSON
 

I imagine we could just create `/changeset/:id/download.json` to match `/changeset/:id/download`. The JSON format would be the same as what https://github.com/zerebubuth/openstreetmap-cgimap/pull/407 is proposing[^1]

[^1]: specially, a single array with `"action": "create" | "modify" | "delete"` instead of the 3 arrays for `create`, `modify`, `delete` which is more similar to the current XML format.

### Screenshots

_No response_

k-yle:
Closes #5970

### Description
Created a new endpoint `/changeset/:id/download.json` which returns the `osmChange` in a JSON format in addition to the existing XML endpoint.

For example:

```jsonc
{
  "version": "0.6",
  "osmChange": {
    "create": [
      { "type": "node", "id": 7, "lat": -36.8359797, "lon": 174.7683977, "timestamp": "2025-05-01T09:00:52Z", "version": 1, "changeset": 2, "user": "kylenz_testing", "uid": 1 }
    ],
    "modify": [
      { "type": "way", "id": 2, "timestamp": "2025-05-01T09:00:52Z", "version": 2, "changeset": 2, "user": "kylenz_testing", "uid": 1, "nodes": [5, 6, 7, 5], "tags": { "landuse": "industrial" } }
    ],
    "delete": [
      { "type": "node", "id": 4, "timestamp": "2025-05-01T09:00:52Z", "version": 2, "changeset": 2, "user": "kylenz_testing", "uid": 1, "visible": false }
    ]
  }
}

```

**Important:** This proposed format is different to the format proposed in https://github.com/zerebubuth/openstreetmap-cgimap/pull/407 They key different is that we keep it simple like the current XML format, instead of adding a new `action` property to every node/way/relation. [Comparison](https://github.com/mmd-osm/openstreetmap-cgimap/blob/89abe65e5231959ae5ad4fa16aa479dd44d48505/test/test_apidb_backend_changeset_uploads.cpp#L2596-L2622)

In my opinion, this format is better because:
* it's consistent with the current XML format
* It doesn't use a top-level array. top-level arrays are annoying because they prevents other properties from being added in the future. Some apps already add custom properties to the XML file<sup>[[1]](https://forum.osm.org/t/108670)</sup>
* It's more organised and understandable since it doesn't add any new properties to the `Node`/`Way`/`Relation` objects/classes.

### How has this been tested?
- Tested locally with a fresh database, using changesets that create+modify+delete multiple nodes+ways+relations
- Unit tests


app/views/api/changesets/downloads/show.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.partial! "api/root_attributes"
+
+json.osmChange do
+  json.create @created.map { |elt| JSON.parse(render(partial: "api/#{elt.to_partial_path}", object: elt)) }
+  json.modify @modified.map { |elt| JSON.parse(render(partial: "api/#{elt.to_partial_path}", object: elt)) }
+  json.delete @deleted.map { |elt| JSON.parse(render(partial: "api/#{elt.to_partial_path}", object: elt)) }
```
hlfan, line 6: I don't think having the json serialized, parsed and serialized again is the best way to do this.

k-yle: any suggestions how to do it more efficiently? I experimented with jbuilder's `.merge` but couldn't find a solution

mmd-osm:
> In my opinion, this format is better because: …

Now for the upload, you might have multiple create/modify/delete blocks in random order, including repeating blocks for the same action. How would this work with your proposed format?

k-yle:
> Now for the upload, you might have multiple create/modify/delete blocks in random order, including repeating blocks for the same action. How would this work with your proposed format?

If I understand you correctly, you're referring to the case where someone uploads an XML file like this:

```xml
<osmChange>
  <create> <node id="1" … />  </create>
  <update> <node id="2" … />  </update>
  <create> <way id="3" … />  </create> <!-- create again -->
  <create> <node id="4" … />  </create> <!-- and again -->
</osmChange>
``` 

Is there a reason why uploaders can't just store 3 arrays (`create`, `update`, `delete`) and append each element to the correct array before uploading it?

mmd-osm:
They could in theory, but the xml format today doesn't have any assumptions in that direction. Processing happens from top to bottom in the exact sequence defined in xml, and you're free to do pretty much whatever you like in terms of repeated actions.

k-yle:
got it, I wonder if we actually need to support this behaviour in the new JSON upload endpoint. Since this is brand new, we have a (somewhat rare) opportunity to define the format more strictly. 

I think it's important that upload and download used the same format (and this time we can document it on the wiki properly and perhaps even create a JSON schema file, either of which would be a big improvement over [the current docs](https://osm.wiki/OsmChange))

lonvia:
Changefiles today are usually sorted by (OSM type, OSM id, version). This ensures that most(*) dependent changes are already available, when reading an object. Consumers may rely on that order.

(*) The obvious exception are nested relations, but they are a beast to handle anyway.

mmd-osm:
I think we can keep the sorting of objects (per action) as is, regardless of the format.

Obviously we would have to define in which sequence create/modified/delete actions need to be processed, b/c JSON properties don't have a defined sequence (as opposed to json array elements).

I agree that the action attribute doesn't add much value, the information is anyway available in version and visible attributes. I guess there are historical reasons which I'm not familiar with in detail.

However, for the upload there's an additional 'if-unused' attribute, that wouldn't fit in a pure play osm format. Some sort of extension seems unavoidable in this case.

mmd-osm:
I tried a few variations to have `if-unused` on the same level as it is in the XML version. This turned out to be tricky. It would require us to introduce another `elements` array that would have to be added to create and modify actions as well for the sake of consistency.

<details>
<summary>(expand me)</summary>

```JSON
    "delete": {
      "if-unused": true,
      "elements": [
        {
          "type": "relation",
          "id": 18320662,
          "version": 1,
          "changeset": 159556552
        },
        {
          "type": "way",
          "id": 1338869356,
          "version": 1,
          "changeset": 159556552
        },
```
</details>

All in all, this seems rather ugly and I wouldn't recommend it. Also, it doesn't address the issue that some elements could have "if-unused": true, while other could have "if-unused": false. 

I see two options here: we could add it as an  "options" property on global level. Besides, "if-used" could also be added to each delete element, like I've mentioned in my previous post. I'm a bit undecided if we should support both, and allow the property on the delete element level to override the global option. Maybe that's over-engineering, and having only a global property would be just fine.

I'm demo'ing both variants in the example below:

```JSON
{
  "version": "0.6",
  "generator": "demo",
  "options": {
    "close_changeset_after_uploading": true,
    "if-unused": true
  },
  "osmChange": {
    "create": [
      {
        "type": "node",
        "id": -1,
        "lat": 81.0714,
        "lon": -139.0956,
        "changeset": 159556552,
        "tags": {
          "man_made": "mast",
          "name": "Principal"
        }
      }
    ],
    "modify": [
      {
        "type": "node",
        "id": 12394725038,
        "version": 3,
        "lat": -50.2701,
        "lon": 131.176,
        "changeset": 159556552,
        "tags": {
          "man_made": "mast",
          "name": "Future"
        }
      }
    ],
    "delete": [
      {
        "type": "relation",
        "id": 18320662,
        "version": 1,
        "changeset": 159556552,
        "if-unused": true
      },
      {
        "type": "way",
        "id": 1338869356,
        "version": 1,
        "changeset": 159556552,
        "if-unused": false
      },
      {
        "type": "node",
        "id": 12394725046,
        "version": 1,
        "changeset": 159556552,
        "if-used": true
      }
    ]
  }
}

```

An (optional) `options` property would also be a good place to store additional instructions, like "close the changeset after uploading". The options property isn't relevant for the download and can be omitted. The idea is to have one format that's fit for both upload and download scenarios.

mmd-osm:
> > Changefiles today are usually sorted by (OSM type, OSM id, version). This ensures that most(*) dependent changes are already available, when reading an object. Consumers may rely on that order.

>  I think we can keep the sorting of objects (per action) as is, regardless of the output format (ie. for the changeset download case).

Ok, I guess my comment was a bit premature. This happens when commenting on a mobile without checking the actual implementation...

Sorting today is based on timestamp, object version, type (node/way/rel), and object id ([souce](https://github.com/zerebubuth/openstreetmap-cgimap/blob/v2.0.1/src/osmchange_responder.cpp#L32-L62)). The action itself doesn't matter. It will be determined for each single element ([source](https://github.com/zerebubuth/openstreetmap-cgimap/blob/v2.0.1/src/osmchange_responder.cpp#L188-L207)).

As can be seen in the following examples, the current sorting wouldn't work anymore and re-grouping by action would be mandatory for the proposed JSON format in this PR. API clients would definitely see a fairly different sorting in this case.

* https://www.openstreetmap.org/api/0.6/changeset/52067542/download
* https://www.openstreetmap.org/api/0.6/changeset/52067546/download
* https://www.openstreetmap.org/api/0.6/changeset/52067549/download

I think, my original proposal in https://github.com/zerebubuth/openstreetmap-cgimap/pull/407 would avoid this issue. Iit  uses the exact same sorting as today without a need to regroup the data.

k-yle:
for completeness, another option that hasn't been mentioned yet is:

```jsonc
{
  "osmChange": [
    { "type": "create", "elements": [a, b, c] },
    { "type": "delete", "elements": [d, e] },
    { "type": "modify", "elements": [f, g] },
    { "type": "delete", "elements": [h, i], "ifUnused": true },
  ],
  // extra properties for 'options' or 'changesetTags' or whatever could go here
}
```

this would address all the limitations from the other proposals, _and_ avoids adding properties to the `Node`/`Way`/`Relation` objects[^1]

[^1]: in my opinion, we should avoid adding temporary properties to these objects, because this could become messy and could be weird to implement in a type-safe way in some languages

mmd-osm:
Indeed, this looks like a good option. It's probably a matter of taste to have "type": "create" or "action": "create", either way should be ok. I can see the benefit of keeping the element objects clean.

NB: I plan to use an event-based parser to avoid keeping the entire parsed document in memory. The proposed structure has some implications on how the parser needs to work for if-unused as well as the additional properties/options.

On the other hand, it's a good reminder that we can't make assumptions about the order of properties (i.e. we can't assume insertion order). I think this should still be doable.

mmd-osm:
So we could probably deploy the cgimap implementation on the dev server (as dedicated entry in https://apis.dev.openstreetmap.org/) for further testing, and get some feedback on the proposed format.

The implementation covers changeset download and upload in JSON format. I'm using "action": "create" now, since we're already using type for "node"/"way"/"relation".

jake-low:
> Edit: I've just noticed that https://github.com/OSMCha/osmchange-parser has already implemented the re-grouping. Maybe @jake-low has some insights to share on the JSON format.

Just dropping in to respond to this ping, sorry it took a while.

I extracted osmchange-parser from the backend of OSMCha last year while doing some refactoring. I published it as its own package on the off chance it would be useful somewhere else, but to my knowledge nobody else is currently using it. The JSON structure emitted by that parser isn't something I gave a lot of thought to; it mirrors the input XML structure pretty closely simply because that was easy to accomplish with a SAX-style parser.

I didn't see anyone respond above to @joto's point, but I'd like to reiterate it: I don't think osmChange files need to categorize edits into `create`/`modify`/`delete` groups explicitly. These groups are implied by other properties that are also included in the file (the element's `version` number and `visible` boolean). Implementations that deal with osmChange files need to handle cases where the stated type of the edit doesn't match its semantics ([example](https://github.com/bdon/OSMExpress/blob/main/python/examples/augmented_diff.py#L126-L134)), which means they might as well just ignore the stated type all the time, and instead infer the type from the `version` and `visible` attributes.

Given this, I think a candidate JSON schema that's worth considering is one like this:

```js
{
  "version": "0.6",
  "elements": [
    // this is an array of elements in normal OSM order
    // (sorted by type, then ID, then version number).
    {
      "type": "node",
      "id": 349018659,
      "lat": 47.4624576,
      "lon": -121.4782337,
      "timestamp": "2024-08-19T15:53:56Z",
      "version": 7,
      "changeset": 155467928,
      "user": "Mateusz Konieczny - bot account",
      "uid": 3199858,
      "tags": {
        "ele": "1907.7",
        "gnis:feature_id": "1521553",
        "name": "Kaleetan Peak",
        "natural": "peak",
        "source": "USGS",
        "wikidata": "Q49040648",
        "wikipedia": "en:Kaleetan Peak"
      }
    },
    // ... followed by more elements (omitted in this example) ...
  ]
}
```

This has the advantage that the response is the _exact_ same schema as if you fetched an element individually from the API (e.g. from `/node/:id` or `/node/:id/history`). For comparison:

```
λ curl -s 'https://www.openstreetmap.org/api/0.6/node/349018659' -H 'Accept: application/json' | jq
{
  "version": "0.6",
  "generator": "openstreetmap-cgimap 2.0.1 (764192 spike-08.openstreetmap.org)",
  "elements": [
    {
      "type": "node",
      "id": 349018659,
      "lat": 47.4624576,
      "lon": -121.4782337,
      "timestamp": "2024-08-19T15:53:56Z",
      "version": 7,
      "changeset": 155467928,
      "user": "Mateusz Konieczny - bot account",
      "uid": 3199858,
      "tags": {
        "ele": "1907.7",
        "gnis:feature_id": "1521553",
        "name": "Kaleetan Peak",
        "natural": "peak",
        "source": "USGS",
        "wikidata": "Q49040648",
        "wikipedia": "en:Kaleetan Peak"
      }
    }
  ]
}
```

This is nice especially in strongly typed languages: if you have a type definition for an `Element` as returned by the OSM API, you can reuse that definition when fetching data from any of these endpoints. (I'm ignoring the `if-unused` attribute used for uploading in these examples for simplicity)

```typescript
// this type definition works for responses from several different endpoints
// (/node/:id, /node/:id/history, /changeset/:id/download, etc)
type OSMResponse = {
  version: string;
  generator: string;
  copyright: string;
  attribution: string;
  license: string;
  elements: Element[];
}

type Element = Node | Way | Relation;

type Node = {
  type: 'node';
  id: number;
  lat: number;
  lon: number;
  timestamp: string;
  version: number;
  changeset: number;
  user: string;
  uid: number;
  tags: Record<string, string>;
}

type Way = ...
```

mmd-osm:
Thank you for your feedback, in particular for the interesting example about the create action.

> I didn't see anyone respond above to @joto's point

I think I sort of addressed this one:  _I agree that the action attribute doesn't add much value, the information is anyway available in version and visible attributes. I guess there are historical reasons which I'm not familiar with in detail._

When designing the format from scratch again, there's a high probability that we might simply skip it. It seems there were good reasons to introduce it for the XML format back in the days... zerebubuth might know?

By the way, I'm assuming that even in case we wouldn't use an action attribute, the data is still processed top to bottom, meaning that you can't create a new way, and refer to new nodes that are appearing further down in the file.

> This is nice especially in strongly typed languages: if you have a type definition for an Element as returned by the OSM API, you can reuse that definition when fetching data from any of these endpoints.

Right, it seems we all agree that  a common definition of an element has some benefits, and that we don't want to taint it with unrelated properties, such as "if-unused".

> I'm ignoring the if-unused attribute used for uploading in these examples for simplicity

The question now is, how would you handle such additional attributes in case of an upload? My goal was to keep upload and download somehow similar enough that we don't end up with entirely different structures.

The other goal I had in mind was to keep XML and JSON format similar enough, that both can cover the same use cases, and they don't look totally different (which could result in a major maintenance burden in the future).

joto:
> It seems there were good reasons to introduce it for the XML format back in the days.

I don't think too much thought went into that decision at that time. It probably simply seemed a reasonable way to do things and it totally does make sense if you are thinking in terms of a "diff" file and not in terms of versioning. I am not sure about the timeline there, it might well be that we did not ship history files back then, so the idea of having a "visible" flag instead had not come up. I never thought about the file format until I had to implement it and found it more complex than it would need to be due to that extra level of hierarchy in the XML. And interestingly enough, we still only have change files in XML and not in PBF, because in PBF there would be no other way to format them than with that "visible" flag. (Side note: The [underlying software](https://github.com/openstreetmap/osmdbt) has supported creating PBF diffs for years now, they are just not used.)

k-yle:
If we decide to get rid of the `create`/`update`/`delete` blocks, then there are two options (broadly speaking):

1. allow attributes to be specified some some elements only. Requires a nested array:

```jsonc
{
  "osmChange": [
    { "elements": [a, b, c, d] },
    { "elements": [e, f], "ifUnused": true },
  ],
  // extra properties for 'options' or 'changesetTags' or whatever could go here
}
```

2. require all attributes to specified per-changeset:

```jsonc
{
  "osmChange": [a, b, c, d, e, f],
  "ifUnused": true,
  // extra properties for 'options' or 'changesetTags' or whatever could go here
}
```

---

I know that iD specifies `if-unused` for *all* elements, and I'm struggling to think of a use-case where you would want `if-unused` to only apply to some items

Issue #5810, opened 2025-03-16
Add space and headline to user profile heatmap


tordans:
I am very happy to see https://github.com/openstreetmap/openstreetmap-website/pull/5402 merged and live. Congratulations!

I want to suggest to add a bit of space, a headline and maybe a subheadline above the new element to give the profile page more structure. Right now, the contribution chart has little context and merges with other sections of the page depending on the content users added.

Currently: 

> <img width="974" alt="Image" src="https://github.com/user-attachments/assets/471e7925-8eb9-45b8-8619-0ed76d45667e" />

Proposal:

><img width="944" alt="Image" src="https://github.com/user-attachments/assets/16a86003-fb5b-43a8-89ef-235438bf5992" />

**What to show…**

- We could keep it very simple with a headline like that and some spacing.
- We could make the headline dynamic like on https://hdyc.neis-one.org/?tordans
   > 152 mapping days with 510 changesets in the last year
- We might want to add a sub headline which explains what this is
   > **Mapping activity**
   > Show mapping activity per day for the last 365 days.

PR #5976, opened 2025-05-03
Add headline to heatmap

rkoeze:
### Description
Adds dynamic headline to heatmap. Addresses https://github.com/openstreetmap/openstreetmap-website/issues/5810.
Screenshot 2025-05-04 at 6 33 05 AM

### How has this been tested?
On my local machine I've verified that the header works as expected in the case that a user has 0, 1, and >1 contributions.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -248,6 +248,9 @@
 <div class="richtext text-break"><%= @user.description.to_html %></div>
 
 <% if @heatmap_data.present? %>
+  <h2 class="text-body-secondary fs-5">
+    <%= t(".contributions", :count => @changes_count) %>
+  </h2>
   <div class="row">
     <div class="col overflow-auto">
       <div class="heatmap-wrapper d-flex align-items-start">
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -43,6 +43,8 @@ class UsersController < ApplicationController
             }
           end
       end
+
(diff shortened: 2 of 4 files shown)


app/views/users/show.html.erb
```diff
@@ -238,6 +238,11 @@
 <div class="richtext text-break"><%= @user.description.to_html %></div>
 
 <% if @heatmap_data.present? %>
+  <div class="row">
+    <div class="col text-body-secondary fs-5">
+      <%= t(".contributions.#{@changes_count == 1 ? 'one' : 'other'}", :changes_count => @changes_count) %>
```
tomhughes, line 243: If you call the variable in the translation `count` then you won't need to work out the subkeys yourself as rails i18n will do it for you...

tomhughes (changes requested):
Please squash at least the first three commits together - the tests can be a separate commit if you prefer or you can just squash everything into one.

app/views/users/show.html.erb
```diff
@@ -248,6 +248,11 @@
 <div class="richtext text-break"><%= @user.description.to_html %></div>
 
 <% if @heatmap_data.present? %>
+  <div class="row">
+    <div class="col text-body-secondary fs-5">
+      <%= t(".contributions", :count => @changes_count) %>
+    </div>
+  </div>
```
AntonKhorev, line 255: You don't need `row`/`col` here, you need `<h2>`.

(heatmap probably also doesn't need `row`/`col`)

tomhughes:

PR #5977, opened 2025-05-04
API changeset upload resource

AntonKhorev:
Folows #5965. Finishes API changeset resources refactor.

Converts [diff upload](https://wiki.openstreetmap.org/wiki/API_v0.6#Diff_upload:_POST_/api/0.6/changeset/#id/upload) action of api changeset controller into its own recource/controller.

The PR consists mostly of moving and rewriting tests.


tomhughes:

PR #5979, opened 2025-05-04
API versions, capabilities and permissions resources

AntonKhorev:
Converts to resources everything in `routes.rb` that comes before changesets, except for the deprecated `/api/capabilities`.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,7 +10,7 @@ OpenStreetMap::Application.routes.draw do
   # API
   namespace :api do
     get "capabilities" => "capabilities#show", :as => nil # Deprecated, remove when 0.6 support is removed
-    get "versions" => "versions#show"
+    resource :versions, :only => :show
   end
 
   namespace :api, :path => "api/0.6" do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,15 +9,12 @@ OpenStreetMap::Application.routes.draw do
 
   # API
   namespace :api do
-    get "capabilities" => "capabilities#show" # Deprecated, remove when 0.6 support is removed
+    get "capabilities" => "capabilities#show", :as => nil # Deprecated, remove when 0.6 support is removed
     get "versions" => "versions#show"
   end
 
-  scope "api/0.6", :module => :api do
-    get "capabilities" => "capabilities#show"
-  end
-
   namespace :api, :path => "api/0.6" do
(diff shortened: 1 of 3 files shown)


mmd-osm:
I'm a bit surprised about "deprecated /api/capabilities". Was there any previous discussion on this topic? How would an api client find out about the current API version?

tomhughes:
By using `/api/versions`?

mmd-osm:
This seems like a breaking change. As an example, JOSM needs to call /api/versions + /api/0.6/capabilities, instead of /api/capabilities as it is today.

tomhughes:
Looks good to me, thanks.

gravitystorm:
> JOSM needs to call /api/versions + /api/0.6/capabilities, instead of /api/capabilities as it is today.

Only if it supports multiple API versions. From the original commit message, "If a client only supports one version
they can jump straight to the versioned capabilities call."

> The deprecation happened 6 years ago already, and so far one of the biggest API consumer hasn't implemented it yet...

It currently makes no difference, so I'm not surprised that nobody has changed anything yet!

gravitystorm:
> I think, as long as API clients don't implement the new way, you can't make any progress on the deprecation

My original plan, fwiw, was to keep the /api/capabilities call as returning the same response as /api/0.6/capabilities, until API 0.6 is removed. That way, all clients will have (hopefully!) changed to using the versioned capabilities as part of supporting the new API. But, as you well know, there's no concrete progress there.

1ec5:
I [updated the documentation](https://wiki.openstreetmap.org/wiki/Special:Diff/2852642) to reflect the removal.

gravitystorm:
> I [updated the documentation](https://wiki.openstreetmap.org/wiki/Special:Diff/2852642) to reflect the removal.

It hasn't been removed - see https://www.openstreetmap.org/api/capabilities . This PR was just some refactoring:

> Converts to resources everything in routes.rb that comes before changesets, except for the deprecated /api/capabilities.

This just meant that the deprecated call was "not converted to resources", not that it was removed. It's still in the routes file:

https://github.com/openstreetmap/openstreetmap-website/blob/52f7ad9828ed91a37a99c6fdee1e3013d947d8b1/config/routes.rb#L12

1ec5:
Oh! My bad. I [reverted the change](https://wiki.openstreetmap.org/wiki/Special:Diff/2852676).

PR #5980, opened 2025-05-05
Adjust colors of out-of-viewport bboxes on history pages

AntonKhorev:
- Make newer changeset colors redder. The blue color of older changesets was [more noticeable](https://community.openstreetmap.org/t/blaue-bounding-boxes/128904) than the red of newer. It wasn't very red because I didn't want it to grab too much attention. But looks like it's better to make it slightly more red to increase the distance between it and the in-viewport orange.
- Make fill colors lighter. The highlight fill colors for out-of-viewport changesets are significantly darker than the color for in-viewport changesets. That's because I left there a temporary gray color before I decided on older/newer colors. Now after changing the newer color I can adjust the fill colors to be both lighter and closer to the newer/older colors.

Newer changeset before:
image

Newer changeset after:
image

Older changeset before:
image

Older changeset after:
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -605,7 +605,7 @@ tr.turn {
 
 .changeset-above-sidebar-viewport {
   --changeset-border-color: #CC6655;
-  --changeset-fill-color: #888888;
+  --changeset-fill-color: #DDBBBB;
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks OK to me, thanks.

PR #5985, opened 2025-05-06
Suppress strokes on bbox area layers using stroke: false

AntonKhorev:
Previously strokes were suppressed using weight: 0 which made them have zero width. That produced some unnecessary svg attributes for strokes.


tomhughes:
Looks good to me, thanks.

PR #5987, opened 2025-05-06
Company Name on Profile Page

### LINKED ISSUE #5986: Company Name on Profile Page
### Problem

Currently profile page doesn't provide functionality to add company or work name on the profile (other than adding it manually in the description, which makes visual representation of it inconsistent).

### Description

Company name functionality may encourage users to share more about themselves, especially if they are working on the project closely related to mapping or at the company that actively contributes to the community. In addition to this, it will increase visual consistency. 

### Screenshots

Image

Image

nertc:
This PR addresses "Company Name on Profile Page" issue mentioned in #5986

PR adds Company Name functionality on Profile page. If user modifies company field, value will be shown on the profile page. It can be changed from `Edit Profile`. `company` field can be empty, in which case, company information will be hidden on the profile page.
Ruby - `ProfilesController` has one more field to save
DB - nullable `company` column added to the user table
JS - no change

Tested: manually, controller tests (checking controller saving functionality), system tests (checking translations and flow)

Fixes #5986

Screenshots:
image
image

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -153,6 +153,14 @@
               </dt>
               <dd class="list-inline-item"><%= @user.home_location_name %></dd>
             <% end %>
+            <% if @user.company&.strip.present? %>
+              <dt class="list-inline-item m-0 align-text-bottom">
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-suitcase-lg-fill" viewBox="0 0 16 16">
+                  <path d="M7 0a2 2 0 0 0-2 2H1.5A1.5 1.5 0 0 0 0 3.5v9A1.5 1.5 0 0 0 1.5 14H2a.5.5 0 0 0 1 0h10a.5.5 0 0 0 1 0h.5a1.5 1.5 0 0 0 1.5-1.5v-9A1.5 1.5 0 0 0 14.5 2H11a2 2 0 0 0-2-2zM6 2a1 1 0 0 1 1-1h2a1 1 0 0 1 1 1zM3 13V3h1v10zm9 0V3h1v10z" />
+                </svg>
+              </dt>
+              <dd class="list-inline-item"><%= @user.company %></dd>
+            <% end %>
             <dt class="list-inline-item m-0"><%= t ".mapper since" %></dt>
             <dd class="list-inline-item"><%= l @user.created_at.to_date, :format => :long %></dd>
             <dt class="list-inline-item m-0"><%= t ".last map edit" %></dt>
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -29,6 +29,8 @@ class ProfilesController < ApplicationController
       current_user.image_use_gravatar = true
     end
 
+    current_user.company = params[:user][:company]
+
     current_user.home_lat = params[:user][:home_lat]
     current_user.home_lon = params[:user][:home_lon]
     current_user.home_location_name = params[:user][:home_location_name]
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -34,6 +34,7 @@
 #  diary_comments_count :integer          default(0)
 #  note_comments_count  :integer          default(0)
 #  creation_address     :inet
+#  company              :string
 #
 # Indexes
 #
--- a/app/views/profiles/edit.html.erb
+++ b/app/views/profiles/edit.html.erb
@@ -40,6 +40,10 @@
     </div>
   </fieldset>
 
+  <fieldset class="mb-3">
+    <%= f.text_field :company, :wrapper_class => "col-sm-8", :id => "company" %>
+  </fieldset>
+
   <fieldset>
     <legend><%= t ".home location" -%></legend>
     <p id="home_message" class="text-body-secondary m-0<% if current_user.home_location? %> invisible<% end %>"><%= t ".no home location" %></p>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -143,6 +143,7 @@ en:
         active: "Active"
         display_name: "Display Name"
         description: Profile Description
+        company: Company
         home_lat: Latitude
         home_lon: Longitude
         languages: Preferred Languages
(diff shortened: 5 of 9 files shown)


tomhughes:
What's the logic behind the suitcase icon for company? Is that something that's commonly used for this purpose?

nertc:
@tomhughes Yes, in my practice, `suitcase icon` is commonly used to refer to a job or company. If it's not intuitive, it can be changed to the `building icon` mentioned in the issue. My thought was that the `building icon` may be slightly less intuitive on OSM, as `building icon` may be misinterpreted not as a company but as some other building or mapping related information.

mmd-osm:
It would be good to have a maximum field length, both for home_location_name as well as company.

tomhughes:
One possible concern with this is that people could start claiming an association with companies that they are not in fact associated with...

1ec5:
There’s already a fairly strong disincentive to claim corporate affiliation in at least some parts of the OSM community. 😅 Maybe optimistically linking to a matching wiki page could help by giving the user a way to verify for themselves. Do we have a way to link to only certain categories of wiki pages like [Organised Editing Teams](https://wiki.openstreetmap.org/wiki/Category:Organised_Editing_Teams)? If spam is a concern, I can probably rig something to alert wiki administrators when a page gets added to that category (or a more specific category of your choosing), so we can add it to our watchlists.

nertc:
@tomhughes generally, users are responsible for the correctness of their information on the web and validation of the information of others. Even now, anyone can claim anything in the description field. If someone finds any incorrectness, it can be reported too. If we are talking about specific teams or roles like the Organized Editing Team, some API calls may be added in the separate PR (which will focus on the specific teams and companies) to show the badge or name of them on the profile page.

uw-amy-bordenave:
(Note that above mention of "@[mmd-osm](https://github.com/mmd-osm) mentioned this pull request: Intermittent CORS issues with editing API #6028" is just due to a link typo which they've fixed, so disregard that)

---

I support this change! I chose to make it very clear in [my profile's description](https://www.openstreetmap.org/user/UW%20Amy%20Bordenave) that my account is company/organization-related, but affiliation is a pretty common optional field for many sites now.

As far as concerns regarding incorrect claims of association - no more of an issue than it is now if someone were to falsely claim something in their profile description.

If verification is important, you could allow any entry and also add a little check-mark or similar icon if the account is linked in an Organized Editing/Activities subpage.

For example, OSM users [UW Amy Bordenave](https://www.openstreetmap.org/user/UW%20Amy%20Bordenave), [UW Ryan Liu](https://www.openstreetmap.org/user/UW%20Ryan%20Liu), and [UW Maddie](https://www.openstreetmap.org/user/UW%20Maddie) would get the verification check-mark because they are all linked on [Organised Editing/Activities/OpenSidewalks](https://wiki.openstreetmap.org/wiki/Organised_Editing/Activities/OpenSidewalks). Anyone else could claim they're part of the org and just wouldn't get the check-mark.

That's not perfect - a user can actually be part of multiple Organised Editing Activities and anyone can edit the Wiki, so maybe more brainstorming needed!

tomhughes:
We're not going to have the wiki driving checkmarks, for many reasons, of which it not proving anything at all is just the cherry on the cake.

app/views/users/show.html.erb
```diff
@@ -153,6 +153,14 @@
               </dt>
               <dd class="list-inline-item"><%= @user.home_location_name %></dd>
             <% end %>
+            <% if @user.company&.strip.present? %>
+              <dt class="list-inline-item m-0 align-text-bottom">
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-suitcase-lg-fill" viewBox="0 0 16 16">
+                  <path d="M7 0a2 2 0 0 0-2 2H1.5A1.5 1.5 0 0 0 0 3.5v9A1.5 1.5 0 0 0 1.5 14H2a.5.5 0 0 0 1 0h10a.5.5 0 0 0 1 0h.5a1.5 1.5 0 0 0 1.5-1.5v-9A1.5 1.5 0 0 0 14.5 2H11a2 2 0 0 0-2-2zM6 2a1 1 0 0 1 1-1h2a1 1 0 0 1 1 1zM3 13V3h1v10zm9 0V3h1v10z" />
+                </svg>
+              </dt>
+              <dd class="list-inline-item"><%= @user.company %></dd>
```
1ec5, line 162: > Maybe optimistically linking to a matching wiki page could help by giving the user a way to verify for themselves.

For what it’s worth, I think a verification system (whether based on the wiki or backlinks or whatever) can be separate from this PR. The existing reporting system can probably keep things from getting out of hand in the meantime. If we’re concerned about impersonation, we could add a tooltip to this box indicating that the company affiliation is self-reported.

app/views/users/show.html.erb
```diff
@@ -153,6 +153,14 @@
               </dt>
               <dd class="list-inline-item"><%= @user.home_location_name %></dd>
             <% end %>
+            <% if @user.company&.strip.present? %>
+              <dt class="list-inline-item m-0 align-text-bottom">
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-suitcase-lg-fill" viewBox="0 0 16 16">
+                  <path d="M7 0a2 2 0 0 0-2 2H1.5A1.5 1.5 0 0 0 0 3.5v9A1.5 1.5 0 0 0 1.5 14H2a.5.5 0 0 0 1 0h10a.5.5 0 0 0 1 0h.5a1.5 1.5 0 0 0 1.5-1.5v-9A1.5 1.5 0 0 0 14.5 2H11a2 2 0 0 0-2-2zM6 2a1 1 0 0 1 1-1h2a1 1 0 0 1 1 1zM3 13V3h1v10zm9 0V3h1v10z" />
```
1ec5, line 159: Similar to the location item above, the path should have a `title` containing alternative text for accessibility.

uw-amy-bordenave:
> We're not going to have the wiki driving checkmarks, for many reasons, of which it not proving anything at all is just the cherry on the cake.

Makes sense - that's probably for the best. Regardless, I support this addition 😃

nertc:
Thank you for the reviews. PR was updated according to the comments.

@tomhughes icon was changed from the `suitcase` to the `building`.
@mmd-osm length of the `company field` was changed (I'll create a separate small PR for `home_location field`)
As @1ec5 suggested, `company` now has tooltip that it `is self-reported and not verified`

app/views/users/show.html.erb
nertc: `title` was added

nertc:
PR was updated. `master` was merged. Merge conflicts were resolved.

PR #5988, opened 2025-05-07
Move `gravatar_enable` method to User model

rkoeze:
### Description
The `gravatar_enable` method configures a gravatar for the user. Since this is business logic let's put it on the model instead of in the controller. This also opens the door to testing this method outside of the request/response lifecycle.

### How has this been tested?
The existing test coverage in `ConfirmationsController` continues to pass.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -66,7 +66,7 @@ Metrics/BlockNesting:
 # Offense count: 26
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 309
+  Max: 321
 
 # Offense count: 58
 # Configuration parameters: AllowedMethods, AllowedPatterns.
--- a/app/controllers/confirmations_controller.rb
+++ b/app/controllers/confirmations_controller.rb
@@ -28,7 +28,7 @@ class ConfirmationsController < ApplicationController
       else
         user.activate
         user.email_valid = true
-        flash[:notice] = gravatar_status_message(user) if gravatar_enable(user)
+        flash[:notice] = gravatar_status_message(user) if user.gravatar_enable
         user.save!
         cookies.delete :_osm_anonymous_notes_count
         referer = safe_referer(params[:referer]) if params[:referer]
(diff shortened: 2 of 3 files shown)


tomhughes:
You can just bump that limit - anything in `.rubocop_todo.yml` is just there to allow things that we haven't got around to doing anything yet and for those metrics ones at least it's not at all clear what if anything we even want to do.

gravitystorm:
> You can just bump that limit - anything in `.rubocop_todo.yml` is just there to allow things that we haven't got around to doing anything yet and for those metrics ones at least it's not at all clear what if anything we even want to do.

Well.... they are there in order to discourage the situation from getting even worse. In many circumstances it's pragmatic to bump them up a bit (e.g. adding one more line to a 250-line class) but if we keep hitting these limits or if we blow past them in one PR, it's an indication that we should be refactoring.

For the metrics, I tend to find that the rubocop defaults are too small, our current values are way too large, and we could agree on a custom limit somewhere between the two (with a bit of refactoring for the worst offending files).

rkoeze:
As a fly-by-night contributor I'll happily defer to whatever course of action you prefer. A couple of thoughts:
- I put together [a quick spike](https://github.com/rkoeze/openstreetmap-website/pull/2) showing how we might use concerns to reorganize the code in some of our larger classes. If you think it would be worthwhile I can create an issue where we can discuss this at more length.
- This feels like a tradeoff where neither outcome, domain logic in the controller versus a larger User model, is perfect. I feel like the latter is a slight improvement on the former. But then again, I authored the PR so of course I would say that.
- Provided you agree the premise of this PR makes sense (because if it doesn't, I can just go ahead and close it),  would you like me to refrain from doing any more work on this until the User model is cleaned up? I would prefer to push forward but then again I just like shipping code. Don't trust me!

tomhughes:
Please just update rubocop_todo for the purposes of this PR and if you want to try and address those metrics warnings then we can do so in separate PRs.

I'm somewhat dubious about some of the class length metrics personally - sure you can "solve" them by moving methods to a concern but a concern that is only used by one class seems a but pointless and to my mind just makes things harder as you have to go hunting around to find out where something is when you want to work on it. I understand that others may disagree however ;-)

tomhughes:
Looks good to me, thanks.

PR #5992, opened 2025-05-08
Refactor directions output rendering and unit handling

hlfan:
### Description
- Group unit translations by sending the scope with `translateDistanceUnits`
- Regroup `writeSummary` and `writeSteps` into `writeContent` and `writeStep`
- Unify unit toggling into `writeContent` with `data-unit` to simplify event binding.

### How has this been tested?
locally

### DIFF
--- a/app/assets/javascripts/index/directions-route-output.js
+++ b/app/assets/javascripts/index/directions-route-output.js
@@ -15,17 +15,16 @@ OSM.DirectionsRouteOutput = function (map) {
 
   let distanceUnits = "km";
   let downloadURL = null;
+  let route = {};
 
   function translateDistanceUnits(m) {
-    if (distanceUnits === "km") {
-      return [m, "m", m / 1000, "km"];
-    } else {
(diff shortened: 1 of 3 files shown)


tomhughes:
Looks good to me, thanks.

AntonKhorev:
What was an improvement here?

Especially given this [let's move unit conversions out of formatters](https://github.com/openstreetmap/openstreetmap-website/pull/5915#discussion_r2039306783), which I did, followed by [let's move unit conversions in](https://github.com/openstreetmap/openstreetmap-website/pull/5915#discussion_r2040287390), which I ignored.

Why shouldn't I move unit conversions out again?

AntonKhorev:
The only thing that could be significantly improved here were km/mi buttons code. I didn't do this because I haven't finished with them. There were supposed to be tooltips and I also wanted to try miles-yards, in which case the buttons would took different because they couldn't be as small as "mi" anymore. I put off miles-yards because when I was comparing miles outputs with Google Maps, I wasn't able to get it to show yards even in Britain with en-GB locale.

PR #5993, opened 2025-05-08
Remove reddish color and add shadows to history changeset highlights

AntonKhorev:
Currently highlighted in-viewport bboxes change their color to more reddish. Changing their color made sense previously when only one orange color was in use. However now we have different colors, with red being for newer changesets. The problem is that if the highlighted changeset is located on a pile of newer changesets, it's made less noticeable by this color change. I think it's better to do highlights by other means without color changes. In this PR I'm making highlighted outlines bolder fo in-viewport changesets.

Before / after:
image image

This should especially helpful for valious cases of color blindness.

Before / after for completely colorblind users:
image image

But I had to do one more thing. Changing the color still made highlighted changesets stand out more among other in-viewport changesets, compared with same color + bolder outline. To compensate for keeping the same color I added shadows to highlighted changesets. This should make sense because they are displayed "above" the rest. Initially I wanted a light shadow, like I do for sticky close buttons, and that's what the PR still has for out-of-viewport changesets. However for in-viewport changesets that wasn't enough and I gave them heavier shadows.

Before / after:
image image

The reason why light shadows are not enough for in-viewport highlights is because they can be highlighted without direcly hovering over them on the map. You need to be able to notice highlights while also looking at the sidebar and hovering there:
image

That all makes borders + outlines quite thick, but I think that it can only be a noticeable problem on small screens, such as phones. And on phones you'll rarely see hover highlights.

### DIFF
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -76,7 +76,7 @@ OSM.HistoryChangesetBboxHighlightOutlineLayer = OSM.HistoryChangesetBboxLayer.ex
       weight: changeset.sidebarRelativePosition === 0 ? 8 : 6,
       color: "var(--changeset-outline-color)",
       fill: false,
-      className: this._getSidebarRelativeClassName(changeset)
+      className: this._getSidebarRelativeClassName(changeset) + " changeset-highlight-outline"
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #5994, opened 2025-05-08
Improve statistucs numbers

vjik:
### Description

Improve numbers format on [statistics page](https://planet.openstreetmap.org/statistics/data_stats.html).

Before:

image

After:

image

### How has this been tested?

Manually


nenad-vujicic:
Hi,

Can you please squash your commits into one and remove empty parentheses when calling `format()`?

Thanks!

script/statistics
```diff
@@ -23,11 +29,11 @@ begin
...
-    puts "<tr><td>Number of users</td><td>#{user_count}</td></tr>"
-    puts "<tr><td>Number of uploaded GPS points</td><td>#{tracepoint_count}</td></tr>"
-    puts "<tr><td>Number of nodes</td><td>#{node_count}</td></tr>"
-    puts "<tr><td>Number of ways</td><td>#{way_count}</td></tr>"
-    puts "<tr><td>Number of relations</td><td>#{relation_count}</td></tr>"
+    puts "<tr><td>Number of users</td><td>#{user_count.format()}</td></tr>"
+    puts "<tr><td>Number of uploaded GPS points</td><td>#{tracepoint_count.format()}</td></tr>"
+    puts "<tr><td>Number of nodes</td><td>#{node_count.format()}</td></tr>"
+    puts "<tr><td>Number of ways</td><td>#{way_count.format()}</td></tr>"
+    puts "<tr><td>Number of relations</td><td>#{relation_count.format()}</td></tr>"
```
1ec5, line 36: If you’re open to it, an alternative approach would be to add a small amount of JavaScript that [formats the number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) on the fly.

1ec5 (commented):
This will make the stats much more readable at a glance.

Do we know of anything that scrapes this page and possibly expects these numbers to remain unformatted? Does this also affect [the statistics page on the planet site](https://planet.openstreetmap.org/statistics/data_stats.html) or only on the main domain? I’m no fan of scraping pages for stuff like this, but we don’t have an equivalent API endpoint, so I might need to spread the word about this change.

script/statistics
tomhughes: I think that would be a very bad idea...

tomhughes:
This is the script that generates the page on planet.osm.org yes - that's the only place this output will appear these days though there is probably a redirect from the main site for historical reasons.

It's basically a piece of historical legacy from the very early days that I'd really like to get rid of to be honest - there are probably far better sources of statistics these days.

It certainly used to be the case that [Andy Robinson](https://www.openstreetmap.org/user/blackadder) collected that page every day and generated graphs showing the change over time but I have no idea if he still does that.

vjik:
> Can you please squash your commits into one and remove empty parentheses when calling `format()`?

done

PR #5996, opened 2025-05-08
Use webp tile format for tracestrack topo

quinncnl:
### Description

Tracestrack Maps recently started providing webp tiles. webp is widely supported and has better compression. It loads faster without noticeable quality degradation.

### How has this been tested?

The new tiles can be tested in Map Explorer: https://console.tracestrack.com/explorer#basemap=topo


tomhughes:
Code in the vendor directory needs to be changed upstream before it can be updated here.

quinncnl:
Hi @tomhughes I updated leaflet-osm. Please take a look there.

PR #5997, opened 2025-05-09
Use require_relative in scripts

AntonKhorev:
I tried running the statistics script because of #5994:

```
ruby script/statistics
```

and it didn't work until I made this change. `config/environment.rb` already uses `require_relative`.


mmd-osm:
`ruby ./script/statistics` works with the current code.

AntonKhorev:
Updated `script/deliver-message` and `script/update-spam-blocks`.

tomhughes:
Looks good to me now, thanks.

PR #5998, opened 2025-05-10
Prerender heatmap in partial

### LINKED ISSUE #5810: Add space and headline to user profile heatmap

### LINKED ISSUE #5905: Hide future days in user heatmap

hlfan:
Ditching the ever-increasing amount of code required to accommodate CalHeatmap like in #5906 and going for a custom responsive solution.
Fixes #5905 and closes #5906. Also fixes #5810.
This also makes the history links of the heatmap tab-navigatable, even before hovering over each link.

### DIFF
--- a/app/assets/javascripts/heatmap.js
+++ b/app/assets/javascripts/heatmap.js
@@ -1,91 +1,67 @@
-//= require d3/dist/d3
-//= require cal-heatmap/dist/cal-heatmap
-//= require popper
-//= require cal-heatmap/dist/plugins/Tooltip
+$(function () {
+  const heatmap = $(".heatmap").removeClass("d-none").addClass("d-grid");
+  const weekInfo = getWeekInfo();
+  const maxPerDay = heatmap.data("max-per-day");
+  const weekdayLabels = heatmap.find("[data-weekday]");
+  let weekColumn = 1;
+  let previousMonth = null;
 
-/* global CalHeatmap, Tooltip */
-document.addEventListener("DOMContentLoaded", () => {
-  const heatmapElement = document.querySelector("#cal-heatmap");
-
-  if (!heatmapElement) {
-    return;
+  for (const day of weekdayLabels) {
+    const $day = $(day);
+    const weekday = $day.data("weekday");
+    if (weekday < weekInfo.firstDay % 7) {
+      $day.insertAfter(weekdayLabels.last());
+    }
+    const weekdayRow = getWeekdayRow(weekday);
+    if (weekdayRow % 2 === 0) $day.remove();
+    $day.css("grid-area", weekdayRow + " / 1");
   }
 
-  /** @type {{date: string; max_id: number; total_changes: number}[]} */
-  const heatmapData = heatmapElement.dataset.heatmap ? JSON.parse(heatmapElement.dataset.heatmap) : [];
-  const displayName = heatmapElement.dataset.displayName;
-  const colorScheme = document.documentElement.getAttribute("data-bs-theme") ?? "auto";
-  const rangeColorsDark = ["#14432a", "#4dd05a"];
-  const rangeColorsLight = ["#4dd05a", "#14432a"];
-  const startDate = new Date(Date.now() - (365 * 24 * 60 * 60 * 1000));
-
-  const mediaQuery = window.matchMedia("(prefers-color-scheme: dark)");
-
-  let cal = new CalHeatmap();
-  let currentTheme = getTheme();
-
-  function renderHeatmap() {
-    cal.destroy();
-    cal = new CalHeatmap();
-
-    cal.paint({
-      itemSelector: "#cal-heatmap",
-      theme: currentTheme,
-      domain: {
-        type: "month",
-        gutter: 4,
-        label: {
-          text: (timestamp) => new Date(timestamp).toLocaleString(OSM.i18n.locale, { timeZone: "UTC", month: "short" }),
-          position: "top",
-          textAlign: "middle"
-        },
-        dynamicDimension: true
-      },
-      subDomain: {
-        type: "ghDay",
-        radius: 2,
-        width: 11,
-        height: 11,
-        gutter: 4
-      },
-      date: {
-        start: startDate
-      },
-      range: 13,
-      data: {
-        source: heatmapData,
-        type: "json",
-        x: "date",
-        y: "total_changes"
-      },
-      scale: {
-        color: {
-          type: "sqrt",
-          range: currentTheme === "dark" ? rangeColorsDark : rangeColorsLight,
-          domain: [0, Math.max(0, ...heatmapData.map(d => d.total_changes))]
-        }
+  for (const day of heatmap.find("[data-date]")) {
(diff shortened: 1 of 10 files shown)


HolgerJeromin:
Nice. 🥳
Can you add a screenshot when ready?

hlfan:
default (ISO 8601):
localhost_3000_user_hltest

"ar-AF" locale:
localhost_3000_user_hltest (1)

"dv-MV" locale:
localhost_3000_user_hltest (2)

kcne:
Even though I spent quite a bit of time trying to make our custom configuration work alongside cal-heatmap, I have to say - this approach makes a lot more sense. Modifying the library ended up requiring more code than writing the config itself.

I really like that this removes the third-party dependency and results in a much cleaner, more maintainable, and self-contained solution.

Just a quick question: what border-radius did you end up using for the squares? It seems to vary slightly across the three screenshots. Also, slightly increasing the spacing between columns and rows could help make the heatmap look a bit more balanced and visually appealing.

Thank you.

tomhughes:
With this approach is the javascript actually doing anything that couldn't be done in the rails rendering on the server?

hlfan:
For now, I've stopped just short of CLDR-dependent stuff using `Intl.locale`. But the [TwitterCldr](https://github.com/twitter/twitter-cldr-rb) gem seems to have the necessary information to move everything to Ruby.

app/assets/javascripts/heatmap.js
```diff
@@ -1,91 +1,62 @@
...
+      $day.addClass("d-none");
+      continue;
+    }
+    const count = $day.data("count") ?? 0;
+    const tooltipText = getTooltipText(date, count);
+    $day
+      .css("grid-area", getWeekdayRow(date.getUTCDay()) + " / " + weekColumn)
+      .attr("aria-label", tooltipText)
+      .tooltip({
+        title: tooltipText,
```
tomhughes, line 43: Currently the tooltips are prone to line wrapping which doesn't look very good so I'd suggest adding a custom class here like:

```js
customClass: "heatmap-tooltip"
```

and then a CSS rule to allow wider tooltips:

```css
.heatmap-tooltip {
  --bs-tooltip-max-width: 300px;
}
```

hlfan:
I've now moved the bootstrap styling to the CSS, saving about a dozen+ kB in the page response size.
Also `data-month` now counts beyond 12 so I don't have to deal with jQuery selecting both labels of the same month.
And I've fixed the month labels when the calendar reaches into a month by adding more padding for the month label generation.

kcne:
I've noticed that the border radius of the day cells changes between desktop and mobile — it looks slightly rounded on desktop but becomes almost pill-shaped on mobile. If this isn’t intentional, we might want to set a fixed border-radius to ensure visual consistency across breakpoints.

Desktop:
Screenshot 2025-05-15 at 10 09 28

Mobile:
Screenshot 2025-05-15 at 10 09 35


On macOS using firefox, when scrolling horizontally, the scrollbar overlaps with the heatmap area. It might be a good idea to add some padding at the bottom of the container to prevent this overlap and improve the scrolling experience. 

Screenshot 2025-05-15 at 10 05 17

It might also be a good idea to apply light/dark variants for tooltips based on the current theme to keep visual consistency.

hlfan:
> we might want to set a fixed border-radius

This is what is currently being applied with Bootstrap's `rounded-1`, but I agree a relative border-radius looks more consistent.

> the scrollbar overlaps with the heatmap area

I've added `mb-1` as a compromise between thick overlaid and classic scrollbars, as I can't get `scrollbar-gutter` to produce the desired effect.

> apply light/dark variants for tooltips

That's what Bootstrap already does, although to prevent the tooltip "merging" with the background and for better accessibility, dark tooltips are used in light mode and light ones in dark mode.

app/assets/stylesheets/common.scss
```diff
@@ -960,6 +960,33 @@ img.trace_image {
...
+      @extend .bg-body-secondary, .bg-opacity-75;
+    }
+    &:hover {
+      border: solid 1px #8884;
+    }
+  }
+}
+
+.tooltip.wide {
+	--bs-tooltip-max-width: none;
```
tomhughes, line 991: You've got a tab here as the indent.

hlfan: Not again me not checking the style from the chrome dev tools...

hlfan: Ok I've made myself a pre-commit hook that checks that this doesn't happen again.

tomhughes (approved):
Thanks. I think this is good now.

verhovsky:
I don't like the green.

hlfan:
Could this be a new edition of #5802 but for days? And caching is probably also involved.

Which green would you prefer (and for which color scheme)?

In general it would be better to open a new issue for that though.

verhovsky:
> Which green

The same one as github

Screenshot 2025-05-16 at 15 09 05

and I think their implementation of just 5 color buckets is easier to read than a continuous gradient.

Also, it would be cool if there was a "See more" button that would load the entire contribution graph since account creation, like github lets you click through the years

Screenshot 2025-05-16 at 15 11 07

hlfan:
The yearly history has already been mentioned in #5829, which should IMO be a precondition for that.

And anything involving different light and dark color palletes requires watching the color scheme media query again, which I really want to avoid.

git-tgo:
nenad-vujicic > ... looks now (Edge, Chrome ...

I can partly agree.

  | logged in | logged out
-- | -- | --
FF | ok | ok
Edge | not ok | ok
Chrome | not ok | ok

tomhughes:
Looks OK to me in both Firefox and Chromium on linux whether or not I am logged in.

hlfan:
I think I've found the culprit:
https://github.com/openstreetmap/openstreetmap-website/blob/32f23461fc37f93ba7fcddd901ed156d544c575e/app/assets/javascripts/heatmap.js#L23
getUTCDay is [0,6], weekInfo.firstDay is [1,7]

verhovsky:
> just 5 color buckets is easier to read than a continuous gradient

for example I can't see any of my contributions because in the last few months I've had days with a lot more contributions

Screenshot 2025-05-17 at 11 14 13

PR #5999, opened 2025-05-11
Add Bootstrap .d-none class to print styles to hide svg defs

AntonKhorev:
I merged #5752 which started to use svg `<defs>` more widely. They are in `<svg>` elements that still take space on the page unless hidden. We hide these images with `d-none` class. The problem is this is a Bootstrap class and we don't load Bootstrap for print styles. This is how the map looks in print view:

image

Notice the empty space above the map. I also did that hiding before with routing icons but it wasn't as noticeable because I set width/height on the svg defs image to small enought values.

This PR adds `d-none` class to the print stylesheet:

image

Another option could have been using `hidden` attribute, but it works slightly differently (looks like it prevents margin folding, so the svg still takes some space, even if used together with width=0 + height=0). And I'm not sure if it's standardized yet (https://github.com/w3c/svgwg/issues/508).

### DIFF
--- a/app/assets/stylesheets/print.scss
+++ b/app/assets/stylesheets/print.scss
@@ -52,6 +52,7 @@ header,
 /* Rules for attribution text under the main map shown on printouts */
 
 #attribution {
+  display: block;
   page-break-inside: avoid;
   height: 40px;
   font-size: 12px;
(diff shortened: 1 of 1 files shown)


hlfan (approved):

tomhughes:
Looks good to me, thanks.

PR #6000, opened 2025-05-11
Remove unused Leaflet popup paragraph margins css

AntonKhorev:
Was added in https://github.com/openstreetmap/openstreetmap-website/commit/0d35a10a503f466f5cb7fe41fcb9e0c5167d9ee1, the commit message doesn't say why. There's a number of other Leaflet popup paragraph margin overrides. This one doesn't seem to be active in dashboard/directions/diary popups.


tomhughes:
Looks good to me, thanks.

PR #6001, opened 2025-05-11
Remove min height from dashboard map popups

AntonKhorev:
This min-height only affects dashboard popups of users with small avatars.

Before:
image

After:
image

It doesn't seem to make look anything better. It could have been useful in the past if floated images were used, but I don't thinnk they were when this css rule was introduced. It was first added in https://github.com/openstreetmap/openstreetmap-website/commit/a36f3558dd43dd5a598e36dd21fd5f7d2b4a94f5 and then changed to 50px in https://github.com/openstreetmap/openstreetmap-website/commit/40012189cb311510e545202d688294b9a8490ab1.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -744,7 +744,6 @@ tr.turn {
 
 .content_map .leaflet-popup-content {
   margin: $spacer;
-  min-height: 50px;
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #6003, opened 2025-05-11
Remove unused dropdown caret css

AntonKhorev:
Works fine without this css rule:
image

because `.caret` is not used anymore. It was used explicitly before #2524, after that maybe it was in Bootstrap, but currently it isn't.


tomhughes:
Looks good to me, thanks.

PR #6007, opened 2025-05-12
Use I18n.exists? to filter available locales

AntonKhorev:
The method that checks if a locale exists currently does it by accessing the translation and checking if exception is (not) thrown. It could use `I18n.exists?` instead.


nenad-vujicic (approved):
Works great on my side, thanks!

tomhughes:
Looks good to me, thanks.

PR #6010, opened 2025-05-13
Add latest diaries partial to profile page

### LINKED ISSUE #6009: Add recent diaries section to user profile
### Problem

Currently, the OpenStreetMap (OSM) user profile page only shows the user information and contribution heatmap. While it's possible to navigate to a user's diary entries by visiting the /user/diary page, this content remains somewhat hidden and underutilized.

### Description

It would enhance user engagement and profile richness if we showcased a selection of recent diary entries directly on the user's profile page. Specifically:
 - Show the latest 4 diary entries.
- Optionally make this feature opt-out.
- Alternatively, sort them by most commented instead of latest if feasible.

This section could be placed before the contribution heatmap in the profile grid layout.

### Screenshots

Image

kcne:
### Description

Closes #6009.
Adds recent diary entries partial and displays it on the user profiles pages by implementing a responsive card layout. It takes 4 most recent diary entries by default. Each diary entry is now shown in a Bootstrap card that displays the title, truncated content, comment count, and creation date. The layout is responsive - showing as a single column on mobile and two columns on larger screens

### How has this been tested?
Both testing manually to make sure it works along different screen sizes and adding tests to `users_controller_test.rb`.

### Screenshots
Screenshot 2025-05-13 at 09 02 26

### DIFF
--- a/app/views/diary_entries/_profile_diaries.html.erb
+++ b/app/views/diary_entries/_profile_diaries.html.erb
@@ -1,0 +1,31 @@
+<% if diary_entries.present? %>
+  <h2 class="text-body-secondary fs-5 mt-4"><%= t(".latest_diaries") %></h2>
+  <div class="row row-cols-1 row-cols-md-2 g-4">
+    <% diary_entries.each do |entry| %>
+      <div class="col profile-diary-card">
+        <div class="card h-100">
+          <div class="card-body d-flex flex-column">
+            <p class="card-title">
+              <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" aria-label="Diary Entry"><path d="M2 6h4" /><path d="M2 10h4" /><path d="M2 14h4" /><path d="M2 18h4" /><rect width="16" height="20" x="4" y="2" rx="2" /><path d="M9.5 8h5" /><path d="M9.5 12H16" /><path d="M9.5 16H14" /></svg>
+              <%= link_to entry.title, diary_entry_path(@user, entry) %>
+            </p>
+            <p class="card-text flex-grow-1"><%= truncate(strip_tags(entry.body.to_html), :length => 150) %></p>
+
+            <div class="card-text d-flex justify-content-between align-items-center mt-auto">
+              <small class="text-body-secondary d-flex align-items-center">
+                <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="me-1" aria-label="Comments"><path d="M7.9 20A9 9 0 1 0 4 16.1L2 22Z" /></svg>
+                <%= link_to diary_entry_path(@user, entry, :anchor => "comments"), :class => "text-body-secondary" do %>
+                  <span><%= t(".comments", :count => entry.comments.size) %></span>
+                <% end %>
+              </small>
+              <small class="text-body-secondary d-flex align-items-center">
+                <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="me-1" aria-label="Date"><path d="M8 2v4" /><path d="M16 2v4" /><rect width="18" height="18" x="3" y="4" rx="2" /><path d="M3 10h18" /></svg>
+                <span><%= l(entry.created_at.to_date, :format => :long) %></span>
+              </small>
+            </div>
+          </div>
+        </div>
+      </div>
+    <% end %>
+  </div>
+<% end %>
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -251,6 +251,8 @@
   <%= render :partial => "heatmap", :locals => @heatmap_data %>
 <% end %>
 
+<%= render :partial => "diary_entries/profile_diaries", :locals => { :diary_entries => @user.diary_entries.visible.order(:created_at => :desc).limit(4) } %>
+
 <% if current_user and @user.id == current_user.id %>
   <div class="my-3">
     <%= link_to t(".edit_profile"), edit_profile_path, :class => "btn btn-outline-primary" %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -650,6 +650,11 @@ en:
       my_diary: My Diary
       new: New Diary Entry
       new_title: Compose a new entry in my user diary
+    profile_diaries:
+      latest_diaries: "Latest Diaries"
+      comments:
+        one: "%{count} comment"
+        other: "%{count} comments"
   diary_comments:
     new:
       heading: Add a comment to the following diary entry discussion?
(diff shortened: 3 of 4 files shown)


tomhughes:
If we're going to call it "Recent Diaries" should we apply some sort of date limit to the selection? Otherwise maybe "Latest Diaries" would be a better title?

app/views/diary_entries/_profile_diaries.html.erb
```diff
@@ -0,0 +1,27 @@
+<% if diary_entries.present? %>
+  <h5 class="mt-4 mb-3"><%= t(".recent_diaries") %></h5>
```
rkoeze, line 2: Would it make sense to style the "Recent Diaries" string so that it's consistent with the "x contributions in the last year" string?

app/views/diary_entries/_profile_diaries.html.erb
```diff
@@ -0,0 +1,27 @@
...
+        <div class="card h-100">
+          <div class="card-body d-flex flex-column">
+            <p class="card-title">
+              <%= image_tag "note.svg", :class => "card-icon", :alt => "Note" %>
+              <%= link_to entry.title, diary_entry_path(@user, entry) %>
+            </p>
+            <p class="card-text flex-grow-1"><%= truncate(strip_tags(entry.body.to_html), :length => 150) %></p>
+
+            <div class="card-text d-flex justify-content-between align-items-center mt-auto">
+              <small class="text-body-secondary">
```
rkoeze, line 15: On the `/diary` page we make the `x comments` string a clickable link that takes the user to the comment section of the post. We could also do that here.

rkoeze: Although on second thought we don't show very much of the message so that would very much mean we are enabling people to comment before reading!

app/views/diary_entries/_profile_diaries.html.erb
```diff
@@ -0,0 +1,27 @@
...
+                <%= image_tag "comment.svg", :class => "card-icon", :alt => "Comments" %> <%= entry.comments.size %> <%= "comment".pluralize(entry.comments.size) %>
+              </small>
+              <small class="text-body-secondary">
+                <%= image_tag "calendar.svg", :class => "card-icon", :alt => "Date" %> <%= l(entry.created_at.to_date, :format => :long) %>
```
rkoeze, line 19: The date string and icon don't look like they're aligned vertically. Same thing for the comments. Maybe they're not intended to be. If so ignore!

rkoeze (commented):
Looks great. Left a couple of comments.

I wonder if there's an opportunity to use some of these design elements on the `/diaries` page.

app/views/diary_entries/_profile_diaries.html.erb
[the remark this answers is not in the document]
kcne: Refactored. Thank you.

app/views/diary_entries/_profile_diaries.html.erb
kcne: Certainly does. Applied. Thank you.

app/views/diary_entries/_profile_diaries.html.erb
kcne: I agree that should be link to comments section. I left it that way for now.

app/views/diary_entries/_profile_diaries.html.erb
kcne: Now using flexbox here to align vertically. Thank you.

kcne:
> If we're going to call it "Recent Diaries" should we apply some sort of date limit to the selection? Otherwise maybe "Latest Diaries" would be a better title?

Changed the naming here since latest diaries makes a bit more sense in this regard. Thank you.

> Looks great. Left a couple of comments.
> 
> I wonder if there's an opportunity to use some of these design elements on the `/diaries` page.

I'm not sure. Only thing that comes to my mind here is maybe having the switch to display as grid/block.

kcne:
> I wonder if this should go underneath the heatmap rather than above it? It pretty much pushes the heatmap below the fold on a laptop...

Repositioned as suggested and resolved conflicts. Let me know if some additional changes are needed. Thank you.

app/views/diary_entries/_profile_diaries.html.erb
```diff
@@ -0,0 +1,31 @@
+<% if diary_entries.present? %>
+  <h2 class="text-body-secondary fs-5 mt-4"><%= t(".latest_diaries") %></h2>
+  <div class="row row-cols-1 row-cols-md-2 g-4">
+    <% diary_entries.each do |entry| %>
+      <div class="col profile-diary-card">
+        <div class="card h-100">
+          <div class="card-body d-flex flex-column">
+            <p class="card-title">
+              <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" aria-label="Note"><path d="M2 6h4" /><path d="M2 10h4" /><path d="M2 14h4" /><path d="M2 18h4" /><rect width="16" height="20" x="4" y="2" rx="2" /><path d="M9.5 8h5" /><path d="M9.5 12H16" /><path d="M9.5 16H14" /></svg>
```
hlfan, line 9: "Note" doesn't seem like the right label here, maybe _Diary Entry_?

app/views/diary_entries/_profile_diaries.html.erb
```diff
@@ -0,0 +1,31 @@
...
+              <%= link_to entry.title, diary_entry_path(@user, entry) %>
+            </p>
+            <p class="card-text flex-grow-1"><%= truncate(strip_tags(entry.body.to_html), :length => 150) %></p>
+
+            <div class="card-text d-flex justify-content-between align-items-center mt-auto">
+              <small class="text-body-secondary d-flex align-items-center">
+                <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="me-1" aria-label="Comments"><path d="M7.9 20A9 9 0 1 0 4 16.1L2 22Z" /></svg>
+                <%= link_to "#{diary_entry_path(@user, entry)}#comments", :class => "text-body-secondary" do %>
```
hlfan, line 17: The anchor could be passed to the path.
```suggestion
                <%= link_to diary_entry_path(@user, entry, :anchor => "comments"), :class => "text-body-secondary" do %>
```

app/views/diary_entries/_profile_diaries.html.erb
```diff
@@ -0,0 +1,31 @@
...
+                  <span><%= entry.comments.size %> <%= "comment".pluralize(entry.comments.size) %></span>
```
hlfan, line 18: Shouldn't this be translated?

hlfan (changes requested):
Thanks for the update, it's coming together! I still found a few nitpicks.

app/views/diary_entries/_profile_diaries.html.erb
kcne: Sure. Resolved.

app/views/diary_entries/_profile_diaries.html.erb
kcne: Resolved. Thanks.

app/views/diary_entries/_profile_diaries.html.erb
kcne: Yes. Thanks for the catch.

kcne:
> Thanks for the update, it's coming together! I still found a few nitpicks.

Resolved the issues. Thank you for review.

hlfan (approved):
Thanks for the changes!

tomhughes:
I'm slightly worried that stripping all the HTML tags from the body text might be problematic but we'll probably only find out if people start seeing cases where it causes problems so I think this is probably good to merge - thanks for working on it.

PR #6012, opened 2025-05-14
Allow new diary entry language to be specified in params

AntonKhorev:
It's possible to open the new diary entry link with parameters to initialize the form values like this:  
https://www.openstreetmap.org/diary/new?diary_entry[title]=Subject&diary_entry[body]=texttexttext

The only value that is not supported is entry language because it's overwritten by user preferences. But the prefences are supposed to set only the default language value judging by `diary.default_language` name, not to override explicitly specified values.

This PR allows to use the `diary_entry[language_code]` parameter to specify the language of a new diary entry.


tomhughes:
Sure it's possible, but it's an implementation detail not a public API that we want people to take advantage of, so what exactly is the use for making this work? Is there some internal use where we need it?

tomhughes:
I still don't really understand but I guess this is harmless.

PR #6013, opened 2025-05-15
Add new diary entry links to user's own diary entry pages

AntonKhorev:
You open your diary, look at the top of the page and see something like this:
image

You click *New Diary Entry*, write an entry and publish it. Now you look at the top of the page again and see:
image

Where's the *New Diary Entry* link? The answer is, you're on a different page. Previously you were on your *index* page with several entries, but right after publishing you ended up on the *show* page of your new entry. And for whatever reason we don't show *New Diary Entry* on *show* pages. But I don't think there's a good reason, after all the RSS link for your entire diary is still there.

This PR adds *New Diary Entry* links on *show* pages of your own diary entries.
- includes *Rename entry to diary_entry* commit which turned out to be unnecessary for implementing this, but I think is a good idea anyway

### DIFF
--- a/app/controllers/diary_comments_controller.rb
+++ b/app/controllers/diary_comments_controller.rb
@@ -12,21 +12,21 @@ class DiaryCommentsController < ApplicationController
   allow_thirdparty_images :only => :create
 
   def create
-    @entry = DiaryEntry.find(params[:id])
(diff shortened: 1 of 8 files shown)


tomhughes:
Looks good to me, thanks.

PR #6015, opened 2025-05-15
Use entire width of sidebar for directions table

AntonKhorev:
Our search results and changeset lists in the sidebar look like this:
image

They take full width with borders between entries going from one side to another.

However out routing directions look a little different:
image

They have margins which are especially noticeable when enties are highlighted. Margins don't get highlighted. This PR removes the margins:
image

### DIFF
--- a/app/views/directions/show.html.erb
+++ b/app/views/directions/show.html.erb
@@ -111,9 +111,11 @@
     </div>
   </div>
 
-  <table class='table table-hover table-sm mb-3'>
-    <tbody id="directions_route_steps"></tbody>
-  </table>
+  <div class="mx-n3">
+    <table class='table table-hover table-sm mb-3'>
+      <tbody id="directions_route_steps"></tbody>
+    </table>
+  </div>
 
   <p class="text-center">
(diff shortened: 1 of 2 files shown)


app/views/directions/show.html.erb
```diff
@@ -111,9 +111,11 @@
...
   </div>
 
-  <table class='table table-hover table-sm mb-3'>
-    <tbody id="directions_route_steps"></tbody>
-  </table>
+  <div class="mx-n3">
+    <table class='table table-hover table-sm mb-3'>
+      <tbody id="directions_route_steps"></tbody>
+    </table>
+  </div>
```
hlfan, line 118: You don't have to wrap the table in a div
```suggestion
  <table class='table table-hover table-sm w-auto mx-n3 mb-3'>
    <tbody id="directions_route_steps"></tbody>
  </table>
```

AntonKhorev: Small layout with `w-auto`:
image

Small layout with the wrapper:
image

hlfan (approved):

tomhughes:
Looks good to me, thanks.

PR #6017, opened 2025-05-16
Improve dates in heatmap tooltips

tomhughes:
This uses a custom date format to drop the year from the heatmap tooltips, which doesn't really seem necessary when it only covers a single year - certainly GitHub doesn't show a year.

It also suppresses the leading zero from the day number, so overall "04 January, 2025" becomes "January 4" which seems nicer to me.


AntonKhorev:
Anyway, looks like we're doing it for other things already and TranslateWiki has no issues with it.

PR #6018, opened 2025-05-16
Extract heatmap data assembly into helper

### LINKED ISSUE #6021: Profile heatmap doesn't work in Chrome 136
### URL

https://www.openstreetmap.org/user/TrickyFoxy

Open any active user profile

Version 136.0.7103.114 (Official Build) (arm64)
macOS 15

At least one of the users has everything running on Chrome 132

Image

Firefox:

Image

hlfan:
Make the heatmap data generation more modular for yearly contribution calendars and address some points from the post merge discussion of #5998.
Also fixes #6021.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -25,12 +25,12 @@ class UsersController < ApplicationController
       @title = @user.display_name
 
       @heatmap_data = Rails.cache.fetch("heatmap_data_of_user_#{@user.id}", :expires_at => Time.zone.now.end_of_day) do
-        one_year_ago = 1.year.ago.beginning_of_day
-        today = Time.zone.now.end_of_day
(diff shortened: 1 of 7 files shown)


app/controllers/users_controller.rb
```diff
@@ -2,6 +2,7 @@ class UsersController < ApplicationController
   include EmailMethods
   include SessionMethods
   include UserMethods
+  include HeatmapHelper
```
tomhughes, line 5: This should probably be a `HeatmapMethods` concern rather than a helper - helpers are primarily about adding methods to view, where they are made available automatically, while concerns are about extending models and controllers.

hlfan: I don't think it should have been in the controller at all.

tomhughes (approved):

PR #6024, opened 2025-05-17
Advanced preferences, language dropdown in basic preferences

### LINKED ISSUE #2581: Add language dropdown menu

AntonKhorev:
The goal was to get #5201 to a mergeable state, but its language dropdown is in suboptimal state. So I decided to take a second attempt at #4461 which also has a language selector. If the attempt is successful, the selector can be reused in #5201.

The problems in #5201 with the language selector were:
- Forcing translators to translate every language name into every language;
- Throwing away some languages because of case-sensitive matching;
- Showing languages in the current locale which you may not be able to read. This should be a problem for *a computer in India* scenario from #1115, where they have not just different languages but different scripts too. Anyone selects a *wrong* script, and then nobody is able to read anything, language names included.

The list in this PR is generated using this idea https://github.com/openstreetmap/openstreetmap-website/pull/5201#discussion_r2086953628, by reading native language names from comments in yml files. If we're not sure whether these comments are going to stay (the files are updated by TranslateWiki), we can dump them to some other file and read from there (UPD: done). Note that the UI language set differs from the diary language set in `config/languages.yml`. The resulting language dropdown shows native language names in case/diactric-insensitive order.

image

image

image

Fixes #2581 by showing all available UI languages in a dropdown. You might still want to specify other languages for displaying name tags, that's what the language input on the *Advanced Preferences* page is for. I tried putting everyting on one page, but that complicates the user interface as you can see in #4461. A one-page solution would require either showing two language inputs at the same time or adding some button to switch between them, requiring javascript. The solution here doesn't require extra javascript.


config/initializers/i18n.rb
```diff
@@ -45,3 +45,35 @@ def store_translations(locale, data, options = {})
...
+
+AVAILABLE_LANGUAGES.sort_by! do |entry|
+  # https://stackoverflow.com/a/74029319
+  diactrics = [*0x1DC0..0x1DFF, *0x0300..0x036F, *0xFE20..0xFE2F].pack("U*")
+  entry[:native_name]
+    .downcase
+    .unicode_normalize(:nfd)
+    .tr(diactrics, "")
+    .unicode_normalize(:nfc)
+end
```
tomhughes, line 79: The more correct way to do this is to use a gem like `ffi-icu` or `twitter_cldr` that can do proper unicode sorting, for example with `ffi-icu` you would do something like:

```ruby
collator = ICU::Collation::Collator.new("root")
collator.strength = :primary
AVAILABLE_LANGUAGES.sort! do |a, b|
  collator.compare(a, b)
end
```
Reducing strength from the default of `:tertiary` to `:primary` makes the sort case and accent insensitive though even at tertiary it still sorts by the base character first and only uses accent (at level secondary) and then case (at level tertiary) to break ties which probably makes little practical difference for the data set here.

AntonKhorev: Isn't it then better to dump all languages to some yaml file in a predetermined order, rather than adding another gem?

AntonKhorev: dumped to `config/ui_languages.yml`

PR #6026, opened 2025-05-19
Remove .display_name from user path helper calls

AntonKhorev:
Replaces
```
user_path(user.display_name)
```
and likes with
```
user_path(user)
```

They mostly were replaced already but heatmap and friends-follows brought some back.


tomhughes:
Looks good to me, thanks.

PR #6031, opened 2025-05-20
input_marker checked by default

### LINKED ISSUE #6025: link by gps coordinates does not show a pin
### URL

https://www.openstreetmap.org/?#map=15/55.70787/13.21166

## Action

Send or receive a link with map parameter with latitude and longitude <https://www.openstreetmap.org/?#map=15/55.70787/13.21166>

## Actual behavior

The map is shown without a pin.

Image

## Expected behavior

The map is shown with a pin at the exact given coordinates.

Image

_No response_

monperrus:
fix #6025


monperrus:
yes, the default can only match the intuitive expectation of the majority. hard to measure though.

I believe that lots of people expects markers because this is how it's done in lots of Maps apps, incl. Google Maps.

PR #6032, opened 2025-05-20
Don't draw compass needle for user's location

### LINKED ISSUE #6030: Don't draw compass needle for user's location
When I click the show my location button on the homepage, it also has a compass needle that points north. I'm not looking north, I dont think this data is even available in the browser. It should just be a blue circle without the needle

image

verhovsky:
Closes #6030

I didn't test this.

### DIFF
--- a/app/assets/javascripts/leaflet.locate.js
+++ b/app/assets/javascripts/leaflet.locate.js
@@ -8,6 +8,7 @@ L.OSM.locate = function (options) {
         return OSM.i18n.t("javascripts.map.locate." + options.unit + "Popup", { count: options.distance });
       }
     },
+    showCompass: false,
     ...options
   });
 


HolgerJeromin:
This checks internally for support of 
https://developer.mozilla.org/en-US/docs/Web/API/DeviceOrientationEvent/alpha
or 
https://developer.apple.com/documentation/webkitjs/deviceorientationevent/1804777-webkitcompassheading
When this is no available, the compass hidden (for example on my desktop vivaldi/chrome)

HolgerJeromin (changes requested):
should be closed

hlfan (changes requested):
File a bug report at https://github.com/domoritz/leaflet-locatecontrol instead of disabling the needle for everyone

tomhughes:
Probably https://github.com/domoritz/leaflet-locatecontrol/issues/367 is the culprit which is fixed in 0.84.2 but it looks like dependabot hasn't noticed that for some reason.

tomhughes:
Oh no sorry I was misreading `yarn.lock` and we are on 0.84.2 now.

hlfan:
That fix has landed here, and this only only regards the transform attribute I misplaced. This doesn't have to do with the location marker selection.

PR #6037, opened 2025-05-22
Use same logic for feature names

### LINKED ISSUE #3447: Render addr:housename where name not set
### Description

OSM uses the `name` of an object in changesets, however some people use `addr:housename` instead and these are not rendered in the same way.

Here's an example: https://www.openstreetmap.org/changeset/116790763 where "967626192, v6" is Agate House.

Could the site be changed so that `if name is null and addr:housename exists` display that for the object name?

### Screenshots

_No response_

hlfan:
JS and Ruby side feature naming has been a bit different. Using a similar structure and logic, these parts can become nearly identical to read.
I also moved the standard address to i18n as some languages have it the other way round.
Closes #3447

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -109,23 +109,14 @@ OSM.Query = function (map) {
 
   function featureName(feature) {
     const tags = feature.tags,
-          locales = OSM.preferred_languages;
+          localeKeys = OSM.preferred_languages.map(locale => `name:${locale}`);
 
-    for (const locale of locales) {
-      if (tags["name:" + locale]) {
-        return tags["name:" + locale];
-      }
+    for (const key of [...localeKeys, "name", "ref", "addr:housename"]) {
+      if (tags[key]) return tags[key];
     }
+    // TODO: Localize format to country of address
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/index/query.js
```diff
@@ -108,23 +108,25 @@ OSM.Query = function (map) {
...
-      }
+      if (tags[key]) return tags[key];
     }
 
     if (tags["addr:housenumber"] && tags["addr:street"]) {
-      return tags["addr:housenumber"] + " " + tags["addr:street"];
+      return OSM.i18n.t(
+        "javascripts.query.address",
+        { housenumber: tags["addr:housenumber"], street: tags["addr:street"] }
+      );
```
AntonKhorev, line 129: Should an addres at an arbitrary location be formatted according to a rule for the current ui language?

hlfan: Fair point. It has to be made pretty complicated if it's done at all.
Let's stick to the half wrong way it's currently done in the meantime.

AntonKhorev:
Merged, thanks.

PR #6038, opened 2025-05-22
Remove legacy OpenID login

### LINKED ISSUE #6020: Removing legacy OpenID login option
Before starting a new PR, I wanted to get some feedback, if this login option is working, and if anyone is still using it? My impression is that it is in some broken state for quite a while. I would therefore suggest to remove it. 

[OmniAuth::OpenID](https://github.com/omniauth/omniauth-openid) also mentions that it  _"Provides strategies for authenticating to providers using the OpenID v2 standard, which is deprecated, and very old."_

Endpoints like `https://www.google.com/accounts/o8/` that are being checked in users_controller.rb are out of service for nearly 10 years now.

I also tried with a real OpenID provider (`https://login.ubuntu.com/`). Although the login procedure would take me to Ubuntu to confirm the request, I'm getting an "Authentication failed" failure later on on osm.org. 

For clarity, this is what I'm talking about:

Image

mmd-osm:
### Description
Fixes #6020

### How has this been tested?

Scenarios tested on local instance:

1. GitHub as single third party provider
- Log in / Sign up
- Change external authentication on "My Account" from "Github" to "None", and back to "Github", check log in process

2. No third party providers configured
- Check that UI doesn't mention "or log in with a third party" along with an empty list of third party providers

Open issues:

- app/assets/javascripts/auth_providers.js seems to have a timing issue on /user/new page. Previously, the auth provider pointed to /auth/github?referer=undefined. I added a fallback to replace undefined by an empty string. It would probably make sense to call this code from app/assets/javascripts/user.js in some different way. Please comment with suggestions for improvement.

- I couldn't find out how to test `preferred_auth_provider`. Is this parameter documented somewhere?


mmd-osm:
Regarding the `preferred_auth_provider`, I was looking for the exact URL to trigger the different layout. It seems I need to include this parameter along with the respective auth provider in the referer parameter:

* http://localhost:3000/user/new?referer=%2F%3Fpreferred_auth_provider%3Dgithub
* http://localhost:3000/login?referer=%2Fuser%2Fnew%3Fpreferred_auth_provider%3Dgithub

I couldn't figure out from the lengthy discussion in #4455, if that's really  the way this is supposed to work. Alternatives like https://www.openstreetmap.org/login?preferred_auth_provider%3Dgithub don't appear to be right.

A comment in session_methods.rb seems to indicate that this parameter should be piggybacked into the referrer of the oauth2 authorisation request. Oh well... 

For both URLs I've mentioned, Sign up and Login page layout looks ok to me. I think that's good enough for now.

image

image

app/views/application/_auth_providers.html.erb
```diff
@@ -12,34 +12,10 @@
...
   <% end %>
-
-  <%# :tabindex starts high to allow rendering at the bottom of the template %>
-  <%= form_tag auth_path(:provider => "openid"), :id => "openid_login_form", :class => "collapse" do %>
-    <div class="mb-3">
-      <label for="openid_url" class="form-label">
-        <%= image_tag "auth_providers/openid.svg", :size => "36", :alt => "", :class => "align-text-bottom" %>
-        <%= t ".openid_url" %>
-      </label>
-      <%= hidden_field_tag("referer", params[:referer], :autocomplete => "off") %>
```
mmd-osm, line 38: Ok, I think I know what's going on. We used to have the same "referer" field twice on the /user/new page, once on the openid form (this line), and another time as field  in the "new_user" form:   `<%= hidden_field_tag("referer", h(@referer)) unless @referer.nil? %>` (new.html.erb).

Since the field in the openid form was always present, auth_providers.js never returned "undefined". Now that we're removing this form, this assumption no longer holds true.

So the change in auth_providers.js to have an empty string as fallback should be backwards compatible. Previously, the "auth buttons" always had a referer URL parameter, even if the value was an empty string.

tomhughes: I think that's an existing bug to be honest but preserving the status quo is fine for this pull request.

app/views/accounts/show.html.erb
```diff
@@ -24,9 +24,8 @@
                    :include_blank => t("auth.providers.none"),
                    :hide_label => true,
                    :wrapper => { :class => "col-auto mb-0" } %>
-      <%= f.text_field(:auth_uid, :hide_label => true, :wrapper => { :class => "col mb-0" }) %>
+      <%= f.hidden_field(:auth_uid, :hide_label => true, :wrapper => { :class => "col mb-0" }) %>
```
tomhughes, line 27: I suspect we don't actually need this field any more - it was only there to allow the UID to be edited when OpenID was in use and for other methods users shouldn't be changing the UID so I think just leaving it out will be fine.

We would need to change the `update` method in `AccountsController` to stop comparing the uid for equality and then including it in the redirect but we probably want to do the second part anyway as it's only there for openid.

mmd-osm:
I also haven't addressed this bit yet:

https://github.com/openstreetmap/openstreetmap-website/blob/dcd886e3e37af9b5f0003aec0563eef5a7b02c1e/app/controllers/users_controller.rb#L142-L146

Should we remove it? It seems to be some sort of upgrade from legacy openid in case of a google auth provider.  A few more test cases are also depending on it, and could be removed, too.

tomhughes:
Thanks. I think this looks good now.

PR #6039, opened 2025-05-22
Drop skype from social links

tomhughes:
Drop skype from social links for it has ceased to be and is an ex-social network.


AntonKhorev:
Merged, thanks.

PR #6040, opened 2025-05-22
Focus on newly added social link fields

tomhughes:
When a new social link is added presumably the user is about to fill it in, so focus on it.


AntonKhorev:
Merged, thanks.

PR #6042, opened 2025-05-22
Focus first endpoint when routing is activated

tomhughes:
When routing is activated we are presumably about to look for a route, and anything already in the search box will have been transferred to the route end field, so focus on the route start field ready to accept input.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -216,6 +216,8 @@ OSM.Directions = function (map) {
 
     endpoints[0].enableListeners();
     endpoints[1].enableListeners();
+
+    endpoints[0].focusInput();
   }
 
   const page = {};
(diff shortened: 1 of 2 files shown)


hlfan (approved):

AntonKhorev:
Do you want more virual keyboard issues on phones?

tomhughes:
Does the focusing of the search box on page load causes issues on phones then? or do we only do that on desktop?

AntonKhorev:
Autofocus attribute on page load doesn't cause the keyboard to pop up, javascript focus does.

app/assets/javascripts/index/directions.js
```diff
@@ -216,6 +216,8 @@ OSM.Directions = function (map) {
 
     endpoints[0].enableListeners();
     endpoints[1].enableListeners();
+
+    endpoints[0].focusInput();
```
hlfan, line 220: Well then how about focusing only when reasonable?
```suggestion
    if (!endpoints[0].value) endpoints[0].focusInput();
```

PR #6044, opened 2025-05-23
Add additional common SPAMMY_PHRASES

Firefishy:
### Description
Add additional spam phrases to catch additional spam profiles.

### How has this been tested?
TBC


tomhughes:
I've no idea what I was thinking when I added that blacklist in 12d3f94e3c but I really don't want to be going down the route of playing whackamole in this repository.

gravitystorm:
@tomhughes I agree that we don't want to go down this route, so I'm minded to decline this PR. Do we also want to remove the existing list of phrases, to simplify the code?

tomhughes:
Apparently I added that in 12d3f94e3c8da1ac4a000d52cc3066b5652f20a4 but I have no memory of it... No doubt it was targeting some specific SEO operation at that time so is likely not doing much now.

pablobm:
While I agree that this route is not desirable in the long term, I have to ask: do we have a better solution at the moment?

@Firefishy: would this patch at least alleviate the current workload from spam moderation? Are these phrases still reflective of the latest waves of spam?

Firefishy:
> @Firefishy: would this patch at least alleviate the current workload from spam moderation? Are these phrases still reflective of the latest waves of spam?

Yes, this will help to reduce the 100s of spam accounts that are created each day. This isn't the ultimate _solution_, but an intermediate measure. It should catch around 15% of the spam accounts.

tomhughes:
It doesn't actually stop anything being created, they just get automatically suspended.

tomhughes:
At least very least I don't think the list of phrases should be baked into the source, we should just have a list that can be configured in settings which can be updated by chef in the production deployment.

pablobm:
Keeping them in DB might be useful. For example if we want to use them to train a filter.

Firefishy:
> It doesn't actually stop anything being created, they just get automatically suspended.

Sure. Any measure we take will not stop the spam, but this will make it a bit more difficult to successfully post spam.

pablobm:
> we should just have a list that can be configured in settings

Agree. That's much better.

pablobm:
I propose that we close this. I'm going to create a new PR (maybe not today, definitely by next week) that does the same thing but in DB, removing the `SPAMMY_PHRASES` constant in the process and with a little bit of refactoring.

I've started a conversation with the DWG about spam management (in all its forms). I plan on doing some work on this in the following months. This piece here offers a first, small step towards active work on the topic. Rest assured that I won't do anything substantial without consulting maintainers and DWG.

Issue #6022, opened 2025-05-17
Fix UserLocationChangeTest


hlfan:
Specifically test_User_can_change_their_location, which has caused many failed runs for completely unrelated changes, even on the master branch.

tomhughes:
Well I'd love to fix it but I can't work out what's going wrong!

hlfan:
Which is exactly why I opened this issue

tomhughes:
I did have a go at reproducing it in non-headless mode to try and see what was happening but that seemed to stop it happening for me which I guess goes with the obvious fact that it seems to be in some way timing related.

mmd-osm:
We never see these screenshots images taken when something is failing: `[Screenshot Image]: /home/runner/work/openstreetmap-website/openstreetmap-website/tmp/screenshots/failures_test_translated_links.png`

https://dev.to/ugifractal/rails-system-test-save-failed-screenshots-1c0m suggests to upload these images to Github Actions, in order to inspect them later on.

AntonKhorev:
It fails while still on *Edit Profile* page. You can see that from the text output.

AntonKhorev:
My guess is that the user show action is too slow because it needs to compute the heatmap.

tomhughes:
Isn't it supposed to wait for the page transition to complete though? I guess there is some timeout on how long it will wait but it would help if it said the timeout expired instead of just looking for the element in the old page...

hlfan:
For now this seems to have been fixed in #6089

hlfan:
Well the [`UserSocialLinksTest`](https://github.com/hlfan/openstreetmap-website/actions/runs/15981857333/job/45077819688#step:13:48) just made a return

PR #6046, opened 2025-05-23
Move heatmap to Turbo frame

AntonKhorev:
Might help with #6022 if I'm correct about https://github.com/openstreetmap/openstreetmap-website/issues/6022#issuecomment-2889043683. System tests wait for asserted elements to appear but only for two seconds by default I think. This PR lets everything to appear on the user profile page before the heatmap loads. Or it might not help because we keep adding more, like latest diary entries, but then we can do the same with them.


hlfan:
> if it would be possible to reserve a fixed space up front for the frame to fill

The old SVG version had a fixed height, but now with it being responsive it's more complicated, but with min-height, max-height and an em-based height it's certainly possible.

AntonKhorev:
> The cells resize to fit into the container without scrolling

You get both scrolling and small cells on a phone.

> the current min width of 10 px can easily be increased

To what? Increase it to 14px and no *responsive* resizing is going to happen. That's going to be an improvement even on wide screens because the current max cell size is slightly smaller than 14px which makes cells to have different sizes because of rounding.

PR #6050, opened 2025-05-24
Rename "social share buttons"/"social icons" to "share buttons"/"share button icons"

AntonKhorev:
We have two folders with images: `social_icons` and `social_link_icons`. What's the difference? One is for share buttons on diary pages, another is for links on profile pages.

`social_icons` is for what's currently called "social share buttons". However not all of them are "social". They include email which is not necessarily social and native share functionality which could share to anything.

This PR renames "social share buttons" modules, methods etc to just "share buttons", and `social_icons` directory to `share_button_icons`.


tomhughes:
Looks good to me, thanks.

PR #6052, opened 2025-05-25
Invert icons in dark mode with common class

hlfan:
Instead of adding a rule for each invertible icon/image, bundle these together into a utility class.


AntonKhorev:
The answer is going to be something like https://github.com/openstreetmap/openstreetmap-website/pull/5964#discussion_r2068774443. For example, `#menu-icon` is not going to exist once it's replaced by the [Bootstrap navbar](https://getbootstrap.com/docs/5.3/components/navbar/).

PR #6053, opened 2025-05-26
Drop selenium monkey patch in favour of upstream fix

tomhughes:
Upstream now has a workaround in place, first attempted in https://github.com/SeleniumHQ/selenium/pull/15458 and then fixed by me in https://github.com/SeleniumHQ/selenium/issues/15707 .


AntonKhorev:
Merged, thanks.

Issue #6057, opened 2025-05-27
Slow response in rendering phase while calling /map endpoint


mahmoudhanafy:
Hi Team, Thanks for the great work you are doing. I'm completely new to ruby and to this project.

I'm currently running a local version of osm-website, and I'm calling this endpoint:
```
https://myLocalWebsite/api/0.6/map?bbox=46.699791,24.6334425,46.7233944,24.6471732
```
but it's taking too long in the rendering phase, I see the following line in logs:
```
Completed  200 OK  in 25244ms (Views: 19820.2ms | ActiveRecord: 4122.8ms (26457 queries, 24031 cached) | GC: 1631.1ms)
```

I see a lot of queries like this:
```
SELECT "changesets".* FROM "changesets" WHERE "changesets"."id" = $1 LIMIT $2
```

So I have updated maps_controller to include changeset with nodes [here](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/api/maps_controller.rb#L31), instead of loading it while rendering. it improved the performance but it's still taking around 11-12 seconds. 

When I'm trying the same request on the public website, it's taking less than 1 second (even without my change)
```
https://api.openstreetmap.org/api/0.6/map?bbox=46.699791,24.6334425,46.7233944,24.6471732
```
So I'm wondering if there is any type of caching or multi-threading happening on the public website but not happening on my version.

Thanks in advance,

HolgerJeromin:
> So I'm wondering if there is any type of caching or multi-threading happening on the public website but not happening on my version.

The main performance boost is that the public website does not run this code at all :-)
It uses an API compatible C++ implementation from:
https://github.com/zerebubuth/openstreetmap-cgimap

mahmoudhanafy:
Thanks for your response @HolgerJeromin 
I didn't know about cgimap before, I'm using it now and it's working much faster.
I'm going to close this issue, as it should be fixed by the PR I merged.

PR #6058, opened 2025-05-27
Include changeset with node instead of loading it while rendering

mahmoudhanafy:
Related to this issue: https://github.com/openstreetmap/openstreetmap-website/issues/6057

The exact details in the issue above. 

As changeset is required while rendering [here](https://github.com/openstreetmap/openstreetmap-website/blob/812557147870c02d5dadc22e42b1cf805006292f/app/views/api/nodes/_node.xml.builder#L7)

### DIFF
--- a/app/controllers/api/maps_controller.rb
+++ b/app/controllers/api/maps_controller.rb
@@ -44,7 +44,7 @@ module Api
       else
         way_nodes = WayNode.where(:node_id => node_ids)
         way_ids = way_nodes.collect { |way_node| way_node.id[0] }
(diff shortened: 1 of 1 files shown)


mahmoudhanafy:
Hi @tomhughes
Thanks for the feedback, I have updated the PR.
Please have a look.

mmd-osm:
How much faster is this now compared to the original Rails  version?

mmd-osm:
@1ec5: this change is probably only relevant for developers running their local Rails instance. The code is not used in production.

PR #6059, opened 2025-05-27
"Scroll to changeset" context menu item

AntonKhorev:
When a changeset is visible on the map but not in the sidebar, you know in which direction to scroll but you don't know how long to scroll. If it's too long, you probably don't want to scroll the sidebar manually.

Now you'll be able to scroll with two clicks using a context menu.

image

image

image

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -116,7 +116,10 @@ OSM.History = function (map) {
 
   function toggleChangesetHighlight(id, state) {
     changesetsLayer.toggleChangesetHighlight(id, state);
-    $("#changeset_" + id).toggleClass("selected", state);
+    $("#sidebar_content .changesets ol li").removeClass("selected");
+    if (state) {
+      $("#changeset_" + id).addClass("selected");
+    }
   }
 
   function clickChangeset(id, e) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/index/history.js
```diff
@@ -112,7 +116,10 @@ OSM.History = function (map) {
 
   function toggleChangesetHighlight(id, state) {
     changesetsLayer.toggleChangesetHighlight(id, state);
-    $("#changeset_" + id).toggleClass("selected", state);
+    $("#sidebar_content .changesets ol li").removeClass("selected");
+    if (state) {
+      $("#changeset_" + id).addClass("selected");
+    }
```
tomhughes, line 122: Do we need a similar change inside the `toggleChangesetHighlight` call on the layer or is there something that stops that getting "stuck" in the same way?

AntonKhorev: Highlights don't get stuck in the layer because everything is redrawn when changesets enter/leave the viewport during scrolling. But I changed it anyway to match the highlight logic in the sidebar. Now only one changeset can be highlighted. Previously it was possible to highlight several, but this wasn't used for anything.

tomhughes (approved):
Thanks. I think this is good now.

PR #6061, opened 2025-05-28
Rename confirmations controller test class

AntonKhorev:
Change the class name to ConfirmationsControllerTest to match the filename and the tested controller.


tomhughes:
Looks good to me, thanks.

PR #6062, opened 2025-05-28
Always use button elements for Remove social link buttons

AntonKhorev:
Fixes https://github.com/openstreetmap/openstreetmap-website/pull/5439#issuecomment-2904227124

Here's how tabbing through social link inputs looks like:

https://github.com/user-attachments/assets/0bb01428-9b4c-4a27-9093-6d563c912d45

Notice that only the newly added *Remove* buttons are reachable. That's because *Remove* buttons for already existing links are not actually buttons. They are labels for hidden checkboxes styled as buttons.

This PR makes *Remove* buttons actual `<button>` elements. It also adds placeholders for link inputs, mainly because it's easier to test:
image

Also removes interpolation of strings from Translatewiki into html like this:
```html
<button type="button" class="btn btn-outline-primary col-sm-2 align-self-start">${OSM.i18n.t("javascripts.social_links.remove")}</button>
```
by writing html in the erb template.

### DIFF
--- a/app/assets/javascripts/user.js
+++ b/app/assets/javascripts/user.js
@@ -12,8 +12,7 @@ $(function () {
   let map, marker, deleted_lat, deleted_lon, deleted_home_name, homeLocationNameGeocoder, savedLat, savedLon;
 
   if ($("#social_links").length) {
-    $("#add-social-link").click(function (event) {
-      event.preventDefault();
+    $("#add-social-link").click(function () {
       const newIndex = -Date.now();
       const socialLinkForm = $(`
         <div class="social-link-added-fields row mb-3">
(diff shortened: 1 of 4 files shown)


AntonKhorev:
> The only real use I could think of was accessibility but the actual links aren't numbered once they're filled in so it doesn't seem like it helps with relating them to the remove buttons.

Placeholders are announced by screen readers (at least the on that comes with Ubuntu) even when not visible. But `<label>` is announced too and may take precedence. Now I removed `<label>`s from `.html.erb` and I'm labelling the inputs with `placeholder` and `aria-label` attributes, like GitHub does in this form:
image

AntonKhorev:
> Is the `javascripts.social_links.remove` translation still used somewhere else? Otherwise it should be removed since it's already a duplicate.

It's not used. That's why it was already removed.

tomhughes:

PR #6063, opened 2025-05-29
Move layer definitions to pane controllers

### LINKED ISSUE #5616: Move html from javascript to views

hlfan:
Closes #5616 as a partial spiritual successor of #4335

### DIFF
--- a/app/controllers/layers_panes_controller.rb
+++ b/app/controllers/layers_panes_controller.rb
@@ -4,6 +4,10 @@ class LayersPanesController < ApplicationController
   authorize_resource :class => false
 
   def show
+    @base_layers = MapLayers.full_definitions("config/layers.yml")
+    @overlay_layers = [{ :layer_id => "noteLayer", :name => "notes", :max_area => Settings.max_note_request_area },
+                       { :layer_id => "dataLayer", :name => "data", :max_area => Settings.max_request_area },
+                       { :layer_id => "gpsLayer", :name => "gps" }]
     render :layout => false
   end
 end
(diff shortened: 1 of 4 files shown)


AntonKhorev:
Is it possible to split this pull request per each sidebar pane? For example, to make a PR that moves only the map key first?

hlfan:
Sure, the map key sidebar pane content loading update would need to be merged before the other panes' PRs as they would depend on that.

tomhughes:
I mean it is already split into separate commits I think and I did have a quick look at it reviewing each commit but I've been meaning to come back and have a better look... As I recall my main question was whether the URLs should all be at the top level or should be under some common prefix.

hlfan:
> under some common prefix

Good point, what do you think of "/panes/{pane}"?

hlfan:
A little bit more cleaning up the HTML by putting the layer definitions in the pane controller like how also the legend is prepared. Then I think the structure is clean enough to also close the underlying issue.

tomhughes:
Looks good to me, thanks.

PR #6064, opened 2025-05-29
Social link icon filter

AntonKhorev:
This is how some social link icons look on profile pages in Firefox:
image

For comparison in Chrome they look like this:
image

The icons are supposed to be grayscale, which is achieved by applying `<feColorMatrix>` filter to the entire `<svg>`. Firefox doesn't like that, maybe because `filter="url(#grayscale)"` references something inside the svg. Anyway there's a simpler way to define this filter without using `url`.

If icons are kept with colors, the dark mode filter would have to be changed because currently the icons look like this:
image

This PR replaces filter attributes with `filter="grayscale(100%) brightness(65%)"` and removes `<feColorMatrix>`. For OpenGeoFiction icon it sets colors directly.

After in Firefox:
image image

### DIFF
--- a/app/assets/images/social_link_icons/ogf.svg
+++ b/app/assets/images/social_link_icons/ogf.svg
@@ -1,20 +1,12 @@
 <!-- https://en.namu.wiki/w/%EC%98%A4%ED%94%88%EC%A7%80%EC%98%A4%ED%94%BD%EC%85%98 -->
-<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 107.43229 100.52796"  filter="url(#grayscale)">
-  <defs>
-    <filter id="grayscale">
-      <feColorMatrix type="matrix" values="0.13 0.13 0.13 0 0
-                                           0.15 0.15 0.15 0 0
-                                           0.16 0.16 0.16 0 0
-                                           0 0 0 1 0"/>
-    </filter>
-  </defs>
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 107.43229 100.52796">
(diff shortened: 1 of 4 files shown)


hlfan:
The OGF icon doesn't even need a filter:

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="41.59 94.407 107.43229 100.52796" fill="none" stroke="#212529" stroke-linecap="round" stroke-linejoin="round">
  <path d="M44.9612 115.8098a27.592 69.359-70 0074.5893 27.3579M57.771 101.4969a46.869 69.359-70 0060.6758 21.8713m-72.8155 13.4881a71.626 95.628-70 0061.5908 21.8426M93.1428 98.583a132.103 47.814-70 00-21.8354 63.7494"/>
  <ellipse cx="-121.37007" cy="94.174011" rx="20.410715" ry="33.072914" transform="scale(-1 1) rotate(-20)"/>
  <g fill="#212529" fill-opacity=".13333333" stroke-width="3" transform="scale(-1 1)">
    <ellipse cx="-121.74805" cy="94.363007" rx="39.6875" ry="33.639881" transform="rotate(-20)"/>
    <circle cx="-114.63764" cy="171.88988" r="6.8035707"/>
    <circle cx="-130.51263" cy="182.47322" r="4.9136896"/>
    <circle cx="-144.11978" cy="190.03275" r="3.4017854"/>
  </g>
</svg>
```

hlfan:
> If icons are kept with colors, the dark mode filter would have to be changed because currently the icons look like this

Or just inline the icons so they can use currentColor, then the osm logo derivates (ohm, forum, wiki) can be in color, no inversion necessary.

PR #6065, opened 2025-05-29
Make "W" in osm wiki social icon more readable

AntonKhorev:
The icon currently has "Wiki" inside the magnifying glass. I don't think anyone can read "iki" there without zooming in really far. "iki" also takes away space that can be used to make "W" more readable.

This PR removes "iki" and makes "W" larger.

Before:

image

After:

image


hlfan:
Shouldn't alterations to official logos involve the LWG?

tomhughes:
Where did that icon come from? It doesn't seem to be the favicon for the wiki?

tomhughes:
The "unreadable" part actually has an aria label on it so was it an attempt at accessibility? Though the label is just "iki" and the W has a label of "W" which seems quite silly so may something added that automatically?

1ec5:
> It is the logo in the wiki, idk why it isn't the favicon.

Tracking in openstreetmap/operations#1245.

> @1ec5 was involved in the discussion about it not showing up in the vector 2022 skin.

In [that discussion](https://wiki.openstreetmap.org/wiki/Talk:Wiki#Wordmark_in_Vector_New_version), I think we were going for a logo plus a wordmark, so there wasn’t as much need to deface the logo more legibly.

For the social media bar, I wonder if we even need to repeat the OSM logo motif on OSM-internal sites. Maybe we could just use a generic 📝 icon or something for the wiki, 💬 for the forum, etc., in the same stroked style as the 🌐 for unrecognized websites.

> Though the label is just "iki" and the W has a label of "W" which seems quite silly so may something added that automatically?

Huh, that’s cute. Inkscape automatically adds a `<title>` when you convert text to paths, but some other tool might add `aria-title`. I’m not confident that any screen reader will actually do anything with this label, since it’s embedded inside an `` tag (with `osm_wiki` as the alt text).

hlfan:
> I wonder if we even need to repeat the OSM logo motif on OSM-internal sites

Posing the risk of OHM looking more internal than others, unless we give it the same treatment.

1ec5:
And with my wiki hat on, the difference is very minor with the low contrast. I don’t expect this small deviation from the wiki’s current full-size logo to be a problem. For those who squint, it’s just a little more legible, so it can’t hurt.

AntonKhorev:
How about this:

- no "iki" in microscopic font size (#6072 still wants it)
- two icons for osm for osm subdomains, like this: (forum icon).(osm logo)
- similarly two icons for ohm and ogf subdomains

hlfan:
> no "iki" in microscopic font size

which makes the icon barely different from the Wikipedia icon, sans serif W vs. serif W

> two icons for osm for osm subdomains

Not everything has to be indicated by an additional icon, which also takes up space horizontally, making the user description smaller.

Another idea is to have affiliate links (c.osm.org, osm.wiki, ohm, ogf) in a different color (or background) for icon and text.
Otherwise OGF would be "left out" because they already don't use the osm logo.

Putting the affiliate links first, maybe separated by a hr or in a possibly colored box could also work.

AntonKhorev:
> > no "iki" in microscopic font size
> 
> which makes the icon barely different from the Wikipedia icon, sans serif W vs. serif W

That's why two icons for osm subdomains.

> > two icons for osm for osm subdomains
> 
> Not everything has to be indicated by an additional icon,

That's why no two icons for everything.

> which also takes up space horizontally, making the user description smaller.

This problem already exists. GitHub solves it by having a wider layout.

> Another idea is to have affiliate links (c.osm.org, osm.wiki, ohm, ogf) in a different color (or background) for icon and text.

The color currently wouldn't tell anyone anything.

> Otherwise OGF would be "left out"

How?

> because they already don't use the osm logo.

Why would it matter?

> Putting the affiliate links first, maybe separated by a hr or in a possibly colored box could also work.

That's a change to how link ordering currently works.

hlfan:
Should the osm.town Mastodon also get osm-related treatment?

AntonKhorev:
> Should the osm.town Mastodon also get osm-related treatment?

Maybe.

hlfan:
> That's a change to how link ordering currently works.

Separating them into a section would fit quite well to #6089.

But that aside, what do you think of badges like in #6072?

AntonKhorev:
> Separating them into a section would fit quite well to #6089.

Which ones should be separated into a section? Osm (sub)domains, osm + ohm + ogf, osm + osm Mastodon?

hlfan:
Is something speaking against all of the above?

AntonKhorev:
> Is something speaking against all of the above?

So you want a section for links to:

- osm
- osm wiki
- osm forum
- osm mastodon
- maybe even ohm, depending on your answer elsewhere

Do these links also get their own edit page?

hlfan:
I'm thinking of wikidata style grouping. Depending on the type of claim, it's placed either in the statements or the identifiers section, regardless where it was entered. And the ordering in the edit page wouldn't change. 

Then, if an OSM graphic needs to be involved, the osm section could get a osm background, removing the need to scale down any icon and related readability issues.

AntonKhorev:
> I'm thinking of wikidata style grouping. Depending on the type of claim, it's placed either in the statements or the identifiers section, regardless where it was entered. And the ordering in the edit page wouldn't change.

Which section should get an edit button? Both of them with buttons leading to the same place?

> Then, if an OSM graphic needs to be involved, the osm section could get a osm background, removing the need to scale down any icon and related readability issues.

Again, this depends on what do you consider to be belonging to the osm section. Should ogf be in the osm section?

hlfan:
> Which section should get an edit button?

I'd say make one edit button for the whole sidebar.
Having any more just clutteres up the space considerably.

AntonKhorev:
> > Which section should get an edit button?
> 
> I'd say make one edit button for the whole sidebar. Having any more just clutteres up the space considerably.

This is one step away from having one edit button for the whole profile. Which brings us back to https://github.com/openstreetmap/openstreetmap-website/pull/6084#issuecomment-2936439795.

hlfan:
If we go ahead with regarding the osm-related definition as too fuzzy for a section, even with a two icon approach you still need to make a decision about when to have two icons and when not.
Just an "H" as a logo could be enough in the osm ecosystem to be recognized as ohm and not say Hostinger or Hacker rank, but particularly straightforward and non-confusing it isn't.

PR #6066, opened 2025-05-29
Save any screenshots when tests fail in CI

tomhughes:
This implements the suggestion from #6022 to upload any screenshots from failing system tests as an artefact.


hlfan:
Can [origin/save-screenshots](https://github.com/openstreetmap/openstreetmap-website/tree/save-screenshots) be deleted, or is it still needed?

tomhughes:
Err.... That must have been an accidental push to the wrong repo. I'll delete it.

PR #6067, opened 2025-05-30
Use Object.values in overpass query element merge

AntonKhorev:
Replace `Object.keys().map()` with `Object.values()` which wasn't widely supported when the code was written.


tomhughes:
Looks good to me, thanks.

PR #6069, opened 2025-05-31
Fix sorting of enclosing features with bounds including antimeridian

AntonKhorev:
When using *Query features*, enclosing features are sorted by their bounding box size. But the sorting is broken if the bounding boxes cross the antimeridian. This is noticeable in Russia. The entire Russia should be larger than some place inside Russia, therefore Russia should be at the end of the list. Instead Russia is at the start, because the bbox area is calculated as negative.

This is what Overpass returns as bounds for Russia. `maxlon` is smaller than `minlon`, going up to [this place](https://www.openstreetmap.org/#map=10/65.7679/-168.9787) and wrapping to a negative value:
```json
  "bounds": {
    "minlat": 41.1850968,
    "minlon": 19.4041722,
    "maxlat": 82.0586232,
    "maxlon": -168.9769440
  },
```

Before / after:
image image

Same for Fiji:

Before:
image

After:
image

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -224,15 +224,12 @@ OSM.Query = function (map) {
       });
   }
 
-  function compareSize(feature1, feature2) {
-    const width1 = feature1.bounds.maxlon - feature1.bounds.minlon,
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #6070, opened 2025-05-31
Use map control labels in system tests

AntonKhorev:


tomhughes:
Looks good to me, thanks.

PR #6071, opened 2025-05-31
Truncate rich text description on word boundaries if possible

AntonKhorev:
#6010 added truncated diary entries to profile pages. The truncation procedure is different from from others already in use. It gets html, strips tags and does the standard rails truncation to 150 chars:

```ruby
truncate(strip_tags(entry.body.to_html), :length => 150)
```

So it's truncate to plaintext. This is similar to richtext `.description` for `og:description` from #5056. I experimented using `entry.body.description` instead of `truncate(...)`. One thing that I wanted to change in `.description` is a truncation separator. This is what this PR does.

Before:
image

After:
image

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -239,7 +239,7 @@ module RichText
 
       return nil if text.blank?
 
-      text_truncated_to_word_break = text.truncate(DESCRIPTION_MAX_LENGTH, :separator => " ")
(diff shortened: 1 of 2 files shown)


AntonKhorev:
Splitting on `\b` requires agreeing with the regexp rules of `\b` on what a word is. For example "-" is not a word character. "abc-def" is two words with two boundaries inside: "abc" + "-def" and "abc-" + "def".

tomhughes:

PR #6072, opened 2025-06-01
Use only differences to OSM logo in internal social link icons

hlfan:
Following the discussion in #6065, this PR drops the OSM logo parts of social link icons for OSM related projects to make the icons more easily distinguishable.


AntonKhorev:
Let's start with a change that shouldn't be in this PR. "Use bs light mode bg secondary color consistently". Are you taking line width into account? (no)

hlfan:
>Let's start with a change that shouldn't be in this PR

Removed since taken care of in #6074

hlfan:
I've introduced badges for osm related links:
light2xdark

AntonKhorev:
The advantage of badges is that the icons take the "same" space, which helps with their alignment. That comes at a cost of adding smaller icons.

If you want to shink the osm logo, you'll have to keep its colors, otherwise it will become unrecognizable. But then you're deviating from https://github.com/openstreetmap/openstreetmap-website/pull/5439#issuecomment-2577314212 and https://github.com/openstreetmap/openstreetmap-website/pull/5439#discussion_r2090571628.

Why is it the osm logo that's getting shrunk and not the wiki/forum? Why is the osm logo combined with ohm and ogf?

app/views/social_links/_show.html.erb
```diff
@@ -5,6 +5,11 @@
         <%= image_tag "social_link_icons/#{social_link.parsed[:platform].nil? ? 'other' : social_link.parsed[:platform]}.svg",
                       :alt => social_link.parsed[:platform].nil? ? t(".other") : social_link.parsed[:platform],
                       :size => "16" %>
+        <% if social_link.parsed[:is_osm_related] %>
+          <div class="position-absolute">
+            <%= image_tag "osm_logo.svg", :class => "count-number ms-2 mt-3 rounded-1", :width => "12", :height => "12", :alt => t("layouts.logo.alt_text") %>
```
AntonKhorev: Whad does this badge have to do with `count-number`? There's no number in it.

hlfan:
> Why is it the osm logo that's getting shrunk and not the wiki/forum?

Because it's not the important part, you see the e.g. the W icon which narrows the platforms down to two options, and the indicator just says if it's osm affiliated or not, osm wiki or standard wikipedia.

If instead the osm icon is the large one, that rather reiterates you're on osm (duh) and more importantly it doesn't resolve the readability issues mentioned in #6065.
> Why is the osm logo combined with ohm and ogf?

For ogf it's reasonable to drop in the first place, but the H icon of ohm is quite ambiguous. And when that osm-related platform gets an indicator, that should be the case for every. It doesn't have to be a badge though.

1ec5:
What if we reduce the OHM icon to the outline of the map with the H centered inside of it, similar to [the Public Domain Map logo](https://commons.wikimedia.org/wiki/File:Public_Domain_Map_logo.svg)? This would just be an icon anyways, not necessarily the project’s official logo (which obviously needs work). The wiki icon could be something similar with a W, not sure about the others.

AntonKhorev:
> > Why is it the osm logo that's getting shrunk and not the wiki/forum?
> 
> Because it's not the important part, you see the e.g. the W icon which narrows the platforms down to two options, and the indicator just says if it's osm affiliated or not, osm wiki or standard wikipedia.

What are those two options? Osm wiki and Wikipedia? But there's also the ogf wiki.

Doesn't the osm logo also narrow the platforms?

If I add a link to my other osm account, which icon should it get, a large osm logo or a small one over what exactly?

> If instead the osm icon is the large one, that rather reiterates you're on osm (duh)

It's on a link, so it doesn't reiterate anything, unless you look up where the link leads. But then you can similarly argue against any icon.

> and more importantly it doesn't resolve the readability issues mentioned in #6065.

What does resolve the readability issues? 16x16px "W" as opposed to 12x12px "W" on a badge? (But yes, I'd rather have 16x16 "W" because I'd rather not have badges).

> > Why is the osm logo combined with ohm and ogf?
> 
> For ogf it's reasonable to drop in the first place,

Ogf support is incomplete, only the wiki links are detected but not the main site. By the same logic as currently used in this PR, the ogf logo should become a badge.

> but the H icon of ohm is quite ambiguous. And when that osm-related platform gets an indicator, that should be the case for every. It doesn't have to be a badge though.

Are you saying that ohm should have an osm indicator?

hlfan:
> But there's also the ogf wiki.

The ogf wiki's logo is considerably different from the icon of other wikis as it resembles more 💭 than a W.
> If I add a link to my other osm account, which icon should it get, a large osm logo or a small one over what exactly?

For accounts, reusing `avatar.svg` is an option. This would be marked as osm-related.
> Are you saying that ohm should have an osm indicator?

More or less only because it's icon is an OSM logo derivative, but yes.
But decide for yourself in which case the OHM logo is more easily recognisable:
| a | b |
|-|-|

gravitystorm:
I can appreciate the intent here, but I'm not a fan of the overall result. I've been testing this locally today:

eight

I don't think it's good to have some icons bigger than others, and sticking out to the side like this. I think it looks messy and inconsistent, compared to the current status of everything being the same size and in a vertical line.

Additionally, I'm not sure if it's intentional that the osm_related_indicator (34x34) is bigger than the default height of the list items (at least here on my machine, at 29px), which makes the indicators overlap with each other. Here's a picture showing the dimensions of the list item.

Screenshot from 2026-03-18 16-28-25

Overall, I appreciate that the tiny icons with even tinier differences are hard to tell apart. But I think a better solution would be, if easily distinguishable icons are desired, to design distinctive 16x16 icons for these platforms.

PR #6074, opened 2025-06-01
Redraw OpenGeofiction social icon

AntonKhorev:
#6072 tried to change the color of the ogf icon to match the colors of other social icons. Except that wouldn't work as intended because the lines in the ogf icon are narrower than 1px. It already looks bleak, and with that color change it's going to be bleaker. There's not much to be done about it other than redrawing the entire icon. That's what this PR does.

Here's the ogf icon changes compared to the generic website icon.

Before (100% / 200%):
image image

After (100% / 200%):
image image


tomhughes:
Looks good to me, thanks.

PR #6076, opened 2025-06-01
Reuse bounds size functions

hlfan:
Reuse more utilities instead of redoing from scratch.

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -223,11 +223,8 @@ OSM.Query = function (map) {
   }
 
   function featureArea({ bounds }) {
-    const height = bounds.maxlat - bounds.minlat;
-    let width = bounds.maxlon - bounds.minlon;
-
-    if (width < 0) width += 360;
-    return width * height;
+    if (bounds.maxlon < bounds.minlon) bounds.maxlon += 360;
+    return OSM.boundsArea([[bounds.minlat, bounds.minlon], [bounds.maxlat, bounds.maxlon]]);
   }
 
   /*
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -1,7 +1,6 @@
 L.extend(L.LatLngBounds.prototype, {
-  getSize: function () {
-    return (this._northEast.lat - this._southWest.lat) *
-           (this._northEast.lng - this._southWest.lng);
+  toArray: function () {
+    return [[this.getSouth(), this.getWest()], [this.getNorth(), this.getEast()]];
   },
 
   wrap: function () {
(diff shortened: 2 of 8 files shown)


AntonKhorev:
What do you think `.wrap()` on `L.LatLngBounds` does? See https://github.com/openstreetmap/openstreetmap-website/pull/5473#issue-2770179737.

hlfan:
Updated `.getSize()` so `.wrap()` isn't necessary anymore.

AntonKhorev:
Why do you want to add a Leaflet dependency to the code that doesn't use Leaflet?

hlfan:
> Why do you want to add a Leaflet dependency to the code that doesn't use Leaflet?

Good question, I removed it!

app/assets/javascripts/index/query.js
```diff
@@ -252,7 +247,7 @@ OSM.Query = function (map) {
    */
   function queryOverpass(lat, lng) {
     const latlng = L.latLng(lat, lng).wrap(),
-          bounds = map.getBounds().wrap(),
+          bounds = map.wrapLatLngBounds(map.getBounds()),
```
AntonKhorev: What does this wrap do, given that later each bbox corner point goes through `OSM.cropLocation` which calls `wrap` again?

hlfan: Removed, previously didn't question 2d22ab10f759828e27876f138b419ff4c0913b87.

test/javascripts/osm_test.js
```diff
@@ -55,10 +55,10 @@ describe("OSM", function () {
     it("parses bbox params", function () {
       const expected = L.latLngBounds([57.6247, -3.6845], [57.7247, -3.7845]);
       let params = OSM.mapParams("?bbox=-3.6845,57.6247,-3.7845,57.7247");
-      expect(params).to.have.property("bounds").deep.equal(expected);
+      expect(L.latLngBounds(params.bounds)).to.deep.equal(expected);
```
AntonKhorev: The test claims to test `.mapParams`. Why does it need to convert both the expected and the actual value to `L.latLngBounds` before comparing them?

app/assets/javascripts/leaflet.map.js
```diff
@@ -369,6 +368,14 @@ L.extend(L.Icon.Default.prototype, {
   }
 });
 
+OSM.boundsArea = function ([[minLat, minLon], [maxLat, maxLon]]) {
+  const height = maxLat - minLat;
+  let width = maxLon - minLon;
+
+  if (width < 0) width += 360;
```
AntonKhorev: This fix is specific to what Overpass returns. You want to put it into a generic area calculation. Do you think it's going to work for all possible bounding boxes?

PR #6078, opened 2025-06-02
Test sorting multiple enclosing features across antimeridian

AntonKhorev:
#6076 as it currently is will fail this test.


tomhughes:
Looks good to me, thanks.

PR #6079, opened 2025-06-02
Add lang attribute to Nominatim results from CJK languages

mapmeld:
### Description

In Unicode, some CJK characters such as 化 have one codepoint but will appear differently in Simplified Chinese (<span lang="zh-Hans">化</span>), Traditional Chinese (<span lang="zh-Hant">化</span>), and Japanese (<span lang="ja">化</span>). On the frontend, we can display names correctly using an HTML attribute such as `lang="zh-Hant"` This issue is known as [Han unification](https://en.wikipedia.org/wiki/Han_unification) and it has appeared over the years [in many software projects](https://issues.chromium.org/issues/41315603)

This was addressed in iD https://github.com/openstreetmap/iD/pull/10716 and is a long-running discussion in openstreetmap-carto.

If we add `&addressdetails=1` to Nominatim queries, we can read the country_code and display the best label for mainland China, Hong Kong, Japan, or Taiwan.

### How has this been tested?

This can be tricky to test, as **many names do not change**, and the display_name will be in your browser's language if it's available

- Search results will have a lang tag, such as `lang="zh-HK"` or `lang="ja"`, regardless of language of display_name
- In Taiwan, a search result for <span lang="zh-Hant">彰化</span> should show a horizontal bar in <span lang="zh-Hant">化</span>
- In mainland China, a search result for <span lang="zh-Hans">玉门 expressway</span> should return a split frame <span lang="zh-Hans">门</span>  in the second character, not the 门 with a +

### Notes

As an alternative to adding `&addressdetails=1` to queries, we could possibly parse display_name (varies with the browser language) or use geo bounding boxes?

This matching of languages is imperfect, but without a language tag we are always using your browser's default for any CJK character. It would be difficult to make exceptions (for example, Japanese restaurants in these countries) without a name regex, a language tag, or access to other tags

This does not affect CJK search results from other countries

I have heard that there are some variations for Cyrillic in [Bulgaria](https://en.wikipedia.org/wiki/Bulgarian_alphabet) and [Serbia](https://en.wikipedia.org/wiki/Serbian_Cyrillic_alphabet#Differences_from_other_Cyrillic_alphabets), particularly in italics? But I don't know how universal it is. [Additional info](https://commons.wikimedia.org/wiki/File:Special_Cyrillics_BGDPT.svg)

### DIFF
--- a/app/controllers/searches/nominatim_reverse_queries_controller.rb
+++ b/app/controllers/searches/nominatim_reverse_queries_controller.rb
@@ -2,6 +2,8 @@ module Searches
   class NominatimReverseQueriesController < QueriesController
     include NominatimMethods
 
+    LANGUAGE_CODES = { "cn" => "zh-Hans", "hk" => "zh-HK", "jp" => "ja", "tw" => "zh-Hant" }.freeze
+
     def create
       # get query parameters
       zoom = params[:zoom]
@@ -12,6 +14,14 @@ module Searches
       # ask nominatim
       response = fetch_xml(nominatim_reverse_query_url(:format => "xml"))
 
+      # add lang attribute for frontend in certain regions
+      addressparts = response.elements["reversegeocode/addressparts"]
+      lang = nil
+      if addressparts
+        region_code = addressparts.elements["ISO3166-2-lvl3"]&.text == "CN-HK" ? "hk" : addressparts.elements["country_code"]&.text
+        lang = region_code ? LANGUAGE_CODES[region_code] : nil
+      end
+
       # parse the response
       response.elements.each("reversegeocode/result") do |result|
         lat = result.attributes["lat"]
@@ -21,6 +31,7 @@ module Searches
         description = result.text
 
         @results.push(:lat => lat, :lon => lon,
+                      :lang => lang,
                       :zoom => zoom,
                       :name => description,
                       :type => object_type, :id => object_id)
--- a/app/controllers/concerns/nominatim_methods.rb
+++ b/app/controllers/concerns/nominatim_methods.rb
@@ -18,7 +18,7 @@ module NominatimMethods
     exclude = "&exclude_place_ids=#{params[:exclude]}" if params[:exclude]
 
     # build url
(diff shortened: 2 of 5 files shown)


app/controllers/searches/nominatim_reverse_queries_controller.rb
```diff
@@ -12,6 +14,14 @@ def create
       # ask nominatim
       response = fetch_xml(nominatim_reverse_query_url(:format => "xml"))
 
+      # add lang attribute for frontend in certain regions
+      addressparts = response.elements["reversegeocode/addressparts"]
+      lang = nil
+      if addressparts
+        region_code = addressparts.elements["ISO3166-2-lvl3"]&.text == "CN-HK" ? "hk" : addressparts.elements["country_code"]&.text
+        lang = region_code ? LANGUAGE_CODES[region_code] : nil
```
1ec5, line 22: Does this assume the displayed name is always in Chinese if the result is from China? That could result in suboptimal font selection and unexpected screen reader behavior if the name is in a different language due to the interface language. Based on `extratags`, could we determine whether the display name is based on one of the preferred interface languages or falling back to `name=*`?

lonvia: You cannot make such assumptions. Nominatim decides for each part of the display name separately which translation fits best. Consequently, this whole guessing here will go wrong if a user with Japanese language settings searches something in China and gets a result which is mixed Chinese/Japanese. It's also not clear to me, if a Japanese reader would rather expect Japanese han even in the Chinese-language parts or if that would only be the case when they are able to read the different han versions.

Sounds to me like it is worth tackling language-tagging of the display name in Nominatim instead of making educated guesses here. We currently have a GSoC project running for transliteration of display names. Language-tagging will be a part of that. Should be feasible to return that information.

tomhughes:
Based on what @lonvia says I think we have to refuse this as it stands but thanks for trying to address the issue and I'd encourage you to work with @lonvia and the Nominatim side to get the necessary information coming from Nominatim and them to come back here with a version based on that.

PR #6080, opened 2025-06-02
Remove "preferences updated" flash fallback to old message

AntonKhorev:
It was necessary for tests relying on old French message before getting an update from Translatewiki.


hlfan:
Do we really need these 20 chars in a separate file?
Couldn't this be done using `render :text => t ".message"` instead of `render :partial => ...`?

tomhughes:
Regardless of whether it's worth having it all this looks fine to me as it stands, thanks.

PR #6081, opened 2025-06-03
Fix template rendered on profile update failure

AntonKhorev:
Go to `/profile`, add an empty social link, click *Update Profile*. You should see this:

image

But actually you'll get error 500.

When doing #6060 I forgot to update the template that's rendered when `current_user.save` fails from `:edit` to `:show`. Turned out we didn't have a test for profile update failures.


tomhughes:
Looks good to me, thanks.

PR #6083, opened 2025-06-03
make sure iD's iframe is focused on startup

tyrasd:
Focus iD's iframe element to allow iD to receive keyboard shortcuts directly after loading. This is useful when people want to start adding new elements right away using the shortcuts (`1`, `2`, `3`), to pan/zoom the map using the arrow keys, to open the help dialog `?`, to directly perform operations on a pre-selected map feature, or to use standard browser keyboard navigation (e.g. `tab`) in the editor.

This fixes https://github.com/openstreetmap/iD/issues/9828

This PR also contains a fix the annoying false-positive exception on iD's startup, see https://github.com/openstreetmap/openstreetmap-website/pull/5823#discussion_r2035464188


HolgerJeromin (approved):

app/assets/javascripts/edit/id.js.erb
```diff
@@ -37,7 +37,8 @@ $(function () {
   id.attr("src", idData.url + "#" + params);
 
   id.ready(function () {
-    if (!this.contentWindow.location.href.startsWith(idData.url)) {
+    id.focus();
+    if (this.contentWindow && !this.contentWindow.location.href.startsWith(idData.url)) {
```
hlfan, line 40: When splitting it out, I'd check for valid chaining until the href instead of just the content window.
```suggestion
    const href = this?.contentWindow?.location?.href;
    if (href && !href.startsWith(idData.url)) {
```

Otherwise this looks good to me.

tomhughes:
Does this run the fix of causing keyboards to popup on mobile devices? or are we safe because it's not an input field being focused?

AntonKhorev:
Have you tried adding [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) to the `<iframe>`?

tyrasd:
> Does this run the fix of causing keyboards to popup on mobile devices? 

I tested it now, and it does not cause the keyboard from opening on a mobile browser. Is there a known bug where this is the case? :thinking:

tyrasd:
> Have you tried adding [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) to the `<iframe>`?

I haven't, but given that the attribute is still marked as (kind of) _newly available_, I'd say the manual `.focus()`ing might be to prefer for now.

app/assets/javascripts/edit/id.js.erb
tyrasd: IMHO not: If there is a `this.contentWindow`, but it (for whatever reason) is not a real `document` object (or a `document` object that does not have a `location` property, etc.), I'd rather see a real exception on this line. :shrug:

AntonKhorev:
So what if it's newly available? The worst thing that would happen is that autofocus won't work. Update your browser then. Many of browsers that had `autofocus` added recently are mobile versions, where it's mostly irrelevant.

tyrasd:
Ok, I changed it to use the `autofocus` attribute now.

The most relevant browser with only "recent" implementation of the attribute is Firefox with its full implementation in early 2023. I guess that's acceptable.

app/views/site/_id.html.erb
```diff
@@ -7,5 +7,5 @@
      data[:zoom] = @zoom if @zoom
      data[:gpx] = trace_data_url(params[:gpx], :format => :xml) if params[:gpx]
      data[:url] = id_url(:locale => params[:locale]) %>
-  <%= tag.iframe "", :frameBorder => 0, :id => "id-embed", :class => "w-100 h-100", :allowfullscreen => "", :data => data %>
+  <%= tag.iframe "", :frameBorder => 0, :id => "id-embed", :class => "w-100 h-100", :allowfullscreen => true, :autofocus => true, :data => data %>
```
tyrasd, line 10: PS: I've also changed the boolean attribute `allowfullscreen` to `true` as [recommended](https://api.rubyonrails.org/v8.0.2/classes/ActionView/Helpers/TagHelper.html) in the docs.

app/assets/javascripts/edit/id.js.erb
hlfan: Wouldn't you also get the same real exception with the code as it is without touching it, now that you're not focusing iD with js?

app/views/site/_id.html.erb
tomhughes: Thanks - can you squash the commits together please and then I'll merge this.

tyrasd: done

tomhughes:
Thanks... This look good to me now.

PR #6084, opened 2025-06-03
Profile edit tabs

AntonKhorev:
Splits profile editing into multiple pages. Currently the page is too long with too many different controls.

image
image

Adds edit buttons for each page:
image
image

See also https://github.com/openstreetmap/openstreetmap-website/pull/5439#discussion_r1937023088 and https://github.com/openstreetmap/openstreetmap-website/pull/5439#discussion_r1937045183.

This might help with system tests that fail on profile update. Tests could be failing because scrolling to the *Update Profile* button takes time that's counted towards timeouts. On smaller pages there won't be a need to scroll. So far none of the test failures happened because the profile page didn't load on time.


tomhughes:
This looks pretty good to me.

The only thing that concerns me is that setting up a new account becomes quite long winded - you start editing the profile and fill in one tab and hit update and it sends you back to the profile and you have to scroll down and find edit again to the second tab and so on.

I wonder if we should change where we send people after an update somehow to make it easier to move on to the next tab?

AntonKhorev:
Replaced one "Edit Profile" button with multiple buttons, one for each tab. One of them is the avatar image, that's how it's done on GitHub.

tomhughes:
So this is what a user's own profile looks like with the latest proposal:

image

I'm not certain I like it so I'd be interested to hear what other people think.

I think the avatar will definitely need a tooltip to indicate that you can click on it to change it, as github do.

hlfan:
I don't think even a tooltip would make it discoverable enough, I think it needs a label, either "Edit" again or at least the icon currently used for `changeset.num_changes`.

AntonKhorev:
We can do #6089 and have all edit buttons in the content body instead of trying to squeeze them in the heading, like the current version of this PR does:

image

A *Change Avatar* button could be placed on the top of the sidebar, that's not too far away from the avatar image.

AntonKhorev:
Updated with unfoldable edit buttons in sidebar as shown in https://github.com/openstreetmap/openstreetmap-website/pull/6089#issuecomment-2973621776.

Profile edit tests still haven't failed even once after I split the profile edit page.

tomhughes:
Very nice - looks good to me now, thanks!

PR #6089, opened 2025-06-05
Move profile location and company to sidebar

AntonKhorev:
This will allow placing edit buttons next to location/company in the sidebar too. The buttons won't look as out of place as they do in the heading if #6084 is implemented.

image

### DIFF
--- a/app/views/social_links/_show.html.erb
+++ b/app/views/social_links/_show.html.erb
@@ -1,7 +1,7 @@
-<ul class="list-unstyled social_links">
+<ul class="list-unstyled social_links mb-0">
   <% social_links.each do |social_link| %>
     <li>
-      <%= link_to social_link.parsed[:url], :class => "icon-link mw-100 text-body-secondary mb-3", :rel => "nofollow me" do %>
+      <%= link_to social_link.parsed[:url], :class => "icon-link mw-100 text-body-secondary mb-2", :rel => "nofollow me" do %>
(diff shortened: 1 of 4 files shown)


hlfan:
Having both the home location and the company in one section would be cleaner imo.
Then it will be a distinct "section without links" above the "section with links".

tomhughes:
I think the general idea here is nice, and should help with the edit buttons so I guess we just need to figure out the best presentation.

AntonKhorev:
> Then it will be a distinct "section without links" above the "section with links".

Maybe we'll add links from company to organized mapping wiki page.

AntonKhorev:
This is how it's going to look with edit buttons added:

image

AntonKhorev:
A version with secondary-color buttons:

image

AntonKhorev:
...or wider buttons with subtler borders:

image

AntonKhorev:
Edit buttons in sidebar:
- One button is bad because which edit page added in #6084 should it open?
- Many buttons bad because clutter and they make the sidebar look significantly different to the owner, if compared to how it looks to other users.

Possible solution: do both of the above.

https://github.com/user-attachments/assets/89913d2e-7d69-4ecb-bbf3-0114bb9fe652

Uses https://getbootstrap.com/docs/5.3/components/collapse/ , no custom javascript required.

tomhughes:
Regardless of how we handle editing there doesn't seem to be any disagreement with the idea of moving stuff to the sidebar.

Ordinarily I might have suggested getting rid of the line under the social links so the lines act as separators, but I can see that might help when the buttons are present.

PR #6101, opened 2025-06-15
Load map key only once

hlfan:
Pulling this first part out of #6063.


tomhughes (approved):
Thanks for the fix - this looks good now.

PR #6102, opened 2025-06-15
Clean up query related code

hlfan:
Building on the changes from #6076 that didn't involve ripping out leaflet for consistency and instead prepare the bounds explicitly.
And another few things that caught my eye there.

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -251,7 +251,7 @@ OSM.Query = function (map) {
    */
   function queryOverpass(lat, lng) {
     const latlng = L.latLng(lat, lng).wrap(),
-          bounds = map.getBounds().wrap(),
+          bounds = map.getBounds(),
           zoom = map.getZoom(),
           bbox = [bounds.getSouthWest(), bounds.getNorthEast()]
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/index/query.js
```diff
@@ -250,34 +242,31 @@ OSM.Query = function (map) {
    * In both cases we then ask to retrieve tags and the geometry
    * for each object.
    */
-  function queryOverpass(lat, lng) {
-    const latlng = L.latLng(lat, lng).wrap(),
-          bounds = map.getBounds().wrap(),
```
tomhughes, line 255: You've dropped the `wrap` here because it's "unnecessary" but why is it no longer necessary? Presumably you think overpass handles out of range bounds?

hlfan: `bounds` goes through `OSM.cropLocation` which calls `wrap` again - https://github.com/openstreetmap/openstreetmap-website/pull/6076#discussion_r2132857315

tomhughes: Ah right, that makes sense.

tomhughes:
I think this all looks good, thanks.

PR #6103, opened 2025-06-15
Use a bootstrap dropdown for language selection

tomhughes:
This updates the new locale menu to use a bootstrap dropdown instead of a simple HTML select so that it's appearance better matches the rest of the site.

To do this I had to move the selector out of the secondary nav bar which probably isn't ideal but without that something in bootstrap was grabbing all the list items for the dropdown and moving them to the outer list for the nav bar... That move also causes the icon to move to the right which I think is probably fine but could likely be changed if I knew what I was doing.

I'm not sure the effect in small screen mode is ideal either... I did consider moving the language selector to the top to always be visible in that case which I think is fine but possibly a bit weird?

Other changes are improving the tests a bit and adding an aria label to the selector button to explain what it is for screen readers (and make it easy to click in tests).

New design, large screen, with menu open:

image

and small screen:

image

### DIFF
--- a/app/assets/javascripts/language_selector.js
+++ b/app/assets/javascripts/language_selector.js
@@ -1,4 +1,4 @@
-$(document).on("change", ".language-change-trigger", function () {
-  Cookies.set("_osm_locale", this.value, { secure: true, path: "/", samesite: "lax" });
+$(document).on("click", "#language-selector .dropdown-menu button", function () {
+  Cookies.set("_osm_locale", $(this).data("language"), { secure: true, path: "/", samesite: "lax" });
   document.location.reload();
 });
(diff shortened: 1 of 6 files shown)


AntonKhorev:
I tried using Bootstrap dropdown for this months ago. Results are significantly worse.

Your small screen is not small enough, and browser dev tools make it look larger by not showing parts of the screen space taken by various ui elements on an actual phone.

This is how a native `<select>` looks like:
Screenshot_20250616-112502

This is Bootstrap, with the dropdown not fitting on the screen:
Screenshot_20250616-112646

tomhughes:
What size screen is that? I was using a Pixel 2 at 411x731 when testing...

AntonKhorev:
Those single-column dropdowns are probably not the best idea. I made a version with [Bootstrap Modal](https://getbootstrap.com/docs/5.3/components/modal/) and up to four columns:

image

Gets down to two columns on reasonably small screens (360px):

image

tomhughes:
Superseded by #6119.

PR #6106, opened 2025-06-15
Move share sidebar to view

hlfan:
Half of #6063 regarding the share sidebar

### DIFF
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -4,7 +4,7 @@ class Ability
   include CanCan::Ability
 
   def initialize(user)
-    can :read, [:feature_query, :legend_pane]
+    can :read, [:feature_query, :legend_pane, :share_pane]
     can :read, [Node, Way, Relation, OldNode, OldWay, OldRelation]
     can [:show, :create], Note
     can :read, :directions
--- a/app/assets/javascripts/leaflet.share.js
+++ b/app/assets/javascripts/leaflet.share.js
@@ -6,251 +6,37 @@ L.OSM.share = function (options) {
           adjustButton: false
         });
 
-  control.onAddPane = function (map, button, $ui) {
+  function init(map, $ui) {
     // Link / Embed
-    $("#content").addClass("overlay-right-sidebar");
 
-    const $linkSection = $("<div>")
-      .attr("class", "share-link p-3 border-bottom border-secondary-subtle")
-      .appendTo($ui);
+    $ui.find("#link_marker").on("change", toggleMarker);
 
-    $("<h4>")
-      .text(OSM.i18n.t("javascripts.share.link"))
-      .appendTo($linkSection);
-
-    let $form = $("<form>")
-      .appendTo($linkSection);
-
-    $("<div>")
-      .attr("class", "form-check mb-3")
-      .appendTo($form)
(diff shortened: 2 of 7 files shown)


app/controllers/shares_controller.rb
```diff
@@ -0,0 +1,10 @@
+class SharesController < ApplicationController
+  before_action :authorize_web
+  before_action :set_locale
+  authorize_resource :class => false
+
+  def show
+    expires_in 7.days, :public => true
```
AntonKhorev: This caching is what I wanted to remove from map key, once it's loaded on demand. There was an issue somewhere about map key being outdated, but I can't find it.

I don't see a good reason why this caching should be replicated instead.

app/controllers/shares_controller.rb
```diff
@@ -0,0 +1,10 @@
+class SharesController < ApplicationController
```
AntonKhorev: This controller doesn't control "shares". It controls the share sidebar pane. The name could be `SharePanesController`.

app/controllers/shares_controller.rb
hlfan: Can the sidebar panes (map key too) be cached without expiration until the underlying yaml file changes? Then the controllers could again be similar.

app/controllers/shares_controller.rb
hlfan: And the MapKeysController doesn't control the API keys for the maps. Perhaps moving the LegendPanesController to /panes/legend should come first to ensure we proceed consistently.

app/controllers/shares_controller.rb
AntonKhorev: > the underlying yaml file

Which file is it?

app/controllers/shares_controller.rb
AntonKhorev: After eight months of trying to move map keys out of site controller https://github.com/openstreetmap/openstreetmap-website/pull/5024 I wasn't very picky about its name. You don't need to retell me my argument from https://github.com/openstreetmap/openstreetmap-website/pull/5918#discussion_r2040628062.

app/controllers/shares_controller.rb
hlfan: The share pane should update with config/layers.yml, legend pane with config/key.yml.

AntonKhorev: Are you sure you're not forgetting about `config/locales/*.yml`? Because what's going to happen is: you move i18n strings, this invalidates all localized stings for some time, somebody opens the pane and gets it not localized, next day maybe translatewiki pushes the update with strings moved to their new place, but that person won't see this for another week because the pane is cached.

AntonKhorev: I was going to say that if you want the browser to recheck whether the page has changed and receive `304
Not Modified` if it hasn't, this is what normally happens without extra caching instructions. But now I'm not entirely sure after looking at new note pages in the left sidebar, could be that `?xhr=1` unfinished buisiness.

app/controllers/shares_controller.rb
tomhughes: The legend has been moved to `/panes/legend` now...

PR #6107, opened 2025-06-15
Move layers sidebar to view

hlfan:
Half of #6063 regarding the layers sidebar


tomhughes (changes requested):
Presumably this should now be under panes like the legend?

It should also have a basic controller test that checks the route and that the action runs without error.

PR #6109, opened 2025-06-16
Fix antimeridian bounds size

hlfan:
Keeps the width of the antimeridian at 0.


AntonKhorev:
Merged, thanks. Although I don't know if this makes any difference because area calculation is done only on enclosing features.

hlfan:
It doesn't make a difference to the area calculation, but it's more accurate to what it says it does.

PR #6113, opened 2025-06-17
Rename map key to legend

hlfan:
This PR renames all references to "map key" to "legend" in the codebase, as the term "legend" has less ambiguity and improves clarity.
As an example, `(?<!f|foreign|foreign[_ ]|p|primary|primary[_ ]|Settings\.|Object\.|shift|meta|ctrl|alt|api|send_)key(?!board|down|frames)` was the regex used to check occurrences of "key", because there have been instances where only "key" was used instead of "map key" or "mapkey".

The only exception is the translation source together with the SiteTest and IndexTest, which still use "Map Key". This ensures translatewiki moves the existing translations.
The excessive caching in the controller is also removed, as it has become unnecessary with 938212c429a159a1ceeec13b91c4ae966ef6663e.
The _panes_ context has been added to the resource path and the controller name, following discussions in https://github.com/openstreetmap/openstreetmap-website/pull/6063#issuecomment-2973551702 and https://github.com/openstreetmap/openstreetmap-website/pull/6106#discussion_r2149530329 respectively.


app/controllers/legend_panes_controller.rb
```diff
@@ -1,12 +1,11 @@
-class MapKeysController < ApplicationController
+class LegendPanesController < ApplicationController
   before_action :authorize_web
   before_action :set_locale
   authorize_resource :class => false
 
   def show
-    expires_in 7.days, :public => true
```
tomhughes, line 7: I don't think this was particularly because of re-requests from a single session, it was just that it was something that shouldn't be changing much so there was no reason not to cache it. It came in from 817434a2e6f1510bf805e35f0c879b005ebe11dd originally.

hlfan: I've dropped it for now, it can be further discussed in https://github.com/openstreetmap/openstreetmap-website/pull/6106#discussion_r2149515805 or #6112

AntonKhorev: There are reasons not to cache it, but yes, it can be removed in another PR.

AntonKhorev (approved):

tomhughes (approved):
Looks good to me now, thanks.

PR #6115, opened 2025-06-18
Make flash alerts dismissable

AntonKhorev:
image

## Why

I was redoing the language selector. I made it always update the language, even when the user is logged in, which means updating the preferences when a new language is picked. But updating the preferences generates a flash message which may overlap with map controls on small screens.

image

If the alert can be closed, the overlap is not a big problem.

### DIFF
--- a/app/views/layouts/_flash.html.erb
+++ b/app/views/layouts/_flash.html.erb
@@ -1,13 +1,16 @@
 <% [[:error, :danger], [:warning, :warning], [:notice, :success]].each do |flash_type, bootstrap_type| %>
   <% if flash[flash_type] %>
-    <%= tag.div :class => "alert alert-#{bootstrap_type} row mx-0 mb-0 p-3 rounded-0 align-items-center",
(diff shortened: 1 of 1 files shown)


app/views/layouts/_flash.html.erb
```diff
@@ -1,13 +1,17 @@
 <% [[:error, :danger], [:warning, :warning], [:notice, :success]].each do |flash_type, bootstrap_type| %>
   <% if flash[flash_type] %>
-    <%= tag.div :class => "alert alert-#{bootstrap_type} row mx-0 mb-0 p-3 rounded-0 align-items-center",
+    <%= tag.div :class => "alert alert-dismissible alert-#{bootstrap_type} d-flex mb-0 p-3 gap-3 rounded-0 align-items-center",
+                "aria-label" => t("javascripts.close"),
```
hlfan: Do you really want the aria label here or on the added button? Because it has another "Close".

tomhughes:
Looks good to me, thanks.

PR #6116, opened 2025-06-18
Move user list javascript to its own file

AntonKhorev:
Continues #5972 by moving things out of `user.js`.


tomhughes:
Looks good to me, thanks.

PR #6117, opened 2025-06-19
Newer/Older Version buttons in element history

AntonKhorev:
Paginates element history like changesets history, with *Older/Newer* buttons and 20 versions per page. Should help with history pages of elements with hundreds of versions.

Uses Turbo streams, no javascript required.

image


tomhughes:
Looks good to me, and a really nice solution with turbo streams, thanks!

PR #6119, opened 2025-06-20
Language select modal dialog

AntonKhorev:
Implements the dialog from https://github.com/openstreetmap/openstreetmap-website/pull/6103#issuecomment-2985190722. The list inside the dialog has up to four columns to save on vertical scrolling.

image

Changes the button to match existing buttons, as #6103 did:
image
image

In small layout, shows the button next to the menu button:
image

If the user is logged in, the dialog still opens, instead of redirecting to preferences. When you press the language select button you probably expect it to open the dialog instead of doing something else. This was the reason for #6115.
image

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -235,6 +235,10 @@ button.d-md-none[data-bs-target="#select_language_dialog"] {
 
 .select_language_list {
   column-width: 160px;
+
+  small {
+    font-size: 10px;
+  }
 }
 
 /* Utility for styling notification numbers */
--- a/app/views/layouts/_select_language_list.html.erb
+++ b/app/views/layouts/_select_language_list.html.erb
@@ -7,6 +7,9 @@
                                :language_code => language[:code] } }) do %>
         <% if language[:short_native_name] %>
           <%= language[:short_native_name] %>
+          <% if language[:short_native_note] %>
+            <small>(<%= language[:short_native_note] %>)</small>
+          <% end %>
         <% else %>
           <%= language[:native_name] %>
         <% end %>
--- a/config/ui_languages.yml
+++ b/config/ui_languages.yml
@@ -68,6 +68,8 @@
   :native_name: kréyòl Gwadloup
 - :code: ku-Latn
   :native_name: kurdî (latînî)
+  :short_native_name: kurdî
+  :short_native_note: latînî
 - :code: lv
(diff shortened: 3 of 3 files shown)


tomhughes (commented):
Looks very good, and definitely a better effort than mine.

Just one thing... I think the commit message on dc53dc65ea5434e98b23d668334d6d4cacd0e9a1 is wrong? I'm guessing you meant to say `preferred_editor` not `language`?

AntonKhorev:
> Just one thing... I think the commit message on [dc53dc6](https://github.com/openstreetmap/openstreetmap-website/commit/dc53dc65ea5434e98b23d668334d6d4cacd0e9a1) is wrong? I'm guessing you meant to say `preferred_editor` not `language`?

Probably. I was going to add language skipping too, but the line `if params[:language] != I18n.locale.to_s` was already there and I decided that it was enough, and after that I either forgot to update or overwrote the commit message.

Changed to "preferred editor".

AntonKhorev:
Tests were passing when I marked this as ready for review
https://github.com/openstreetmap/openstreetmap-website/actions/runs/15779219028

See #6125, maybe they'll pass after merging that.

It doesn't look like stuck language settings from previous tests.

tomhughes:
They'll probably pass now if rerun - the system tests just aren't reliable and as we add more of them thing just get worse.

AntonKhorev:
They are reliably failing at `ReportDiaryEntryTest#test_missing_report_params` after a number of reruns.

tomhughes:
Passes for me locally though, but then before I merged I did a run and got two system failures then ran again and it passed. The system test framework just doesn't seem to be able to reliably wait for transitions to complete before checking assertions :-(

Anyway this looks good now, thanks.

AntonKhorev:
> The system test framework just doesn't seem to be able to reliably wait for transitions to complete before checking assertions :-(

Not in this case. It loads the page which should have the flash alert rendereded on rails side, but for some reason when assert runs the alert is gone.

We still have other failures where transitions from profile edit pages don't happen on time. They should be fixed by #6084, which requires #6089.

woodpeck:
After this pull request has been merged, an attempt to print a map from the web site now generates a four-page printout of which the first three pages are a language selector:

lang

I have had a user reporting this with Firefox on Windows 11, and can confirm it happens with Firefox 135.0.1 on Linux.

PR #6120, opened 2025-06-21
Rework element versions navigation

### LINKED ISSUE #4946: Better handling of version links for an unversioned element view

### LINKED ISSUE #5618: Wrong details after selecting version
Visit https://www.openstreetmap.org/way/576529969

Okay now on the bottom select version 1. 

Okay now select view details. 

Hey we're looking at the details of version 2. I thought you were going to give us the details of version 1!

AntonKhorev:
When I added element version pages in https://github.com/openstreetmap/openstreetmap-website/pull/4480 I made their navigation similar to changeset navigation. For a changeset you have (< user's previous changeset) . (user's history) . (user's next changeset >). So I copied this principle and got (< previous version) . (version history) . (next version >), which after https://github.com/openstreetmap/openstreetmap-website/pull/4513 looks like this:

image

But this navigation doesn't work very well when you're on the current version page, causing inconveniences and confusion like https://github.com/openstreetmap/openstreetmap-website/issues/4946 and #5618. Users might expect to see the link to the previous version, but it's not there.

image

---

Here's hopefully a better versions navigation.

image

image
image
image
image
image
image

It has a scrollable list of ~100 versions around the visited version and the always visible links to the first/last versions. If the first/last version is the one that's visited, the version next to it is also always visible. For the last version it happens both when the version is opened as the current version (`/node/123`) and as the version with the last number (`/node/123/history/45` if we're looking at the node 123 with 45 versions).

Fixes #5618 by replacing legacy "Details" links with "Node" / "Way" / "Relation" top-level breadcrumbs links. "Details" made sense back when there were only two pages per element: "Details" and "History".

Fixes #4946 by making links to previous version always available, for previous versions of current versions and of versions picked from history.

### DIFF
--- a/app/assets/javascripts/index/element.js
+++ b/app/assets/javascripts/index/element.js
@@ -7,25 +7,58 @@
 
   OSM.Element = function (map, type) {
     const page = {};
+    let scrollStartObserver, scrollEndObserver;
 
     page.pushstate = page.popstate = function (path, id, version) {
       OSM.loadSidebarContent(path, function () {
-        scrollToActiveVersion();
+        initVersionsNavigation();
         page._addObject(type, id, version);
       });
     };
 
     page.load = function (path, id, version) {
-      scrollToActiveVersion();
+      initVersionsNavigation();
       page._addObject(type, id, version, true);
     };
 
     page.unload = function () {
       page._removeObject();
+      scrollStartObserver?.disconnect();
+      scrollStartObserver = null;
+      scrollEndObserver?.disconnect();
+      scrollEndObserver = null;
     };
 
     page._addObject = function () {};
     page._removeObject = function () {};
+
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, and a nice improvement. Thanks!

PR #6122, opened 2025-06-21
Add Mapillary social link

hlfan:
There's even https://osm.wiki/Key:mapillary for pictures, and now you can link yourself too, as https://osm.org/user/Minh%20Nguyen already does.


tomhughes:
Looks good to me, thanks.

PR #6125, opened 2025-06-23
Replace redirect with bad request error for new report with missing params

AntonKhorev:
If the user somehow manages to open `/reports/new` page without `reportable_id` and `reportable_type` parameters, send a bad request error instead of redirection. 

The error happens when you try to open https://www.openstreetmap.org/reports/new. This page shouldn't be accessible without `reportable_id` / `reportable_type`. Links without these parameters shouldn't exist. The redirect also displays a message "Cannot create a new report", but the `new` action is not creating anything, it just displays a form; creating is done by the `create` action which you can't access by opening a broken url.

The redirect with a message was added in https://github.com/openstreetmap/openstreetmap-website/commit/2fc70be734fb6439b1eef340d396af62965817df.

---

The test that I delete in pull request started to fail for unknown reasons when trying to merge #6119. But that pull request doesn't change anything about the redirec and the displayed flash message. What did change recently is that flash messages were made dismissable in #6115, but the tests didn't start to fail because of that.

image

Screenshots of failed tests show that there's no message. Whatever is the cause of that, it doesn't make sense to fix this message because I don't think it should be displayed.

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -14,7 +14,7 @@ class ReportsController < ApplicationController
       @report = Report.new
       @report.issue = Issue.find_or_initialize_by(create_new_report_params)
     else
-      redirect_to root_path, :notice => t(".missing_params")
+      head :bad_request
     end
   end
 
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me, thanks.

PR #6126, opened 2025-06-23
Change incomplete report flash style from success to warning

AntonKhorev:
If you're reporting something and you managed to submit the form without filling out the required fields[^1], there's a message "Please provide the required details". The problem is that it's styled as a *success* message, as if the report was created, while actually it wasn't.

Before:
image

After:
image

[^1]: This shouldn't normally happen, you'll get a javascript error, as long as `required` html attributes are there.


tomhughes:
Looks good to me, thanks.

PR #6127, opened 2025-06-23
Integrate human-readable wikidata browser plugin

hlfan:
### Description
See original plugin [announcement](https://community.openstreetmap.org/t/announcing-human-readable-wikidata-browser-plugins-for-openstreetmap-org/108180/9) and [codebase](https://github.com/ZeLonewolf/osm-wikidata-greasemonkey) from @ZeLonewolf.
Updates the CSP and sets the requested paths from the settings.
Changes to the plugin are requesting more of the user's languages, getting the images more directly and replacing flexbox with float to accommodate longer descriptions better.

### How has this been tested?
DevTools to test on real data.

### DIFF
--- a/app/assets/javascripts/index/element.js
+++ b/app/assets/javascripts/index/element.js
@@ -1,9 +1,17 @@
 (function () {
+  let abortController = null;
+  const languagesToRequest = [...new Set([...OSM.preferred_languages.map(l => l.toLowerCase()), "mul", "en"])];
+  const wikisToRequest = [...new Set(languagesToRequest.filter(l => l !== "mul").map(l => l.split("-")[0] + "wiki"))];
+  const localeName = new Intl.DisplayNames(OSM.preferred_languages, { type: "language" });
+  const isOfExpectedLanguage = ({ language }) => languagesToRequest[0].startsWith(language) || language === "mul";
+
   $(document).on("click", "a[href='#versions-navigation-active-page-item']", function (e) {
     scrollToActiveVersion();
     $("#versions-navigation-active-page-item a.page-link").trigger("focus");
     e.preventDefault();
   });
+
+  $(document).on("click", "button.wdt-preview", e => previewWikidataValue($(e.currentTarget)));
 
   OSM.Element = function (map, type) {
     const page = {};
@@ -13,12 +21,14 @@
       OSM.loadSidebarContent(path, function () {
         initVersionsNavigation();
         page._addObject(type, id, version);
+        abortController = new AbortController();
       });
     };
 
     page.load = function (path, id, version) {
       initVersionsNavigation();
       page._addObject(type, id, version, true);
+      abortController = new AbortController();
     };
 
     page.unload = function () {
@@ -27,6 +37,7 @@
       scrollStartObserver = null;
       scrollEndObserver?.disconnect();
       scrollEndObserver = null;
+      abortController?.abort();
     };
 
     page._addObject = function () {};
@@ -100,5 +111,109 @@
     } else {
       scrollableList.scrollLeft = scrollableList.scrollWidth - scrollableList.offsetWidth;
     }
+  }
+
+  function previewWikidataValue($btn) {
+    if (!OSM.WIKIDATA_API_URL) return;
+    const items = $btn.data("qids");
+    if (!items?.length) return;
+    $btn.prop("disabled", true);
+    fetch(OSM.WIKIDATA_API_URL + "?" + new URLSearchParams({
+      action: "wbgetentities",
+      format: "json",
+      origin: "*",
+      ids: items.join("|"),
+      props: "labels|sitelinks/urls|claims|descriptions",
+      languages: languagesToRequest.join("|"),
+      sitefilter: wikisToRequest.join("|")
+    }), {
+      headers: { "Api-User-Agent": "OSM-TagPreview (https://github.com/openstreetmap/openstreetmap-website)" },
+      signal: abortController?.signal
+    })
+      .then(response => response.ok ? response.json() : Promise.reject(response))
+      .then(({ entities }) => {
+        if (!entities) return Promise.reject(entities);
+        $btn
+          .closest("tr")
+          .after(
+            items
+              .filter(qid => entities[qid])
+              .map(qid => getLocalizedResponse(entities[qid]))
+              .filter(data => data.label || data.icon || data.description || data.article)
+              .map(data => renderWikidataResponse(data, $btn.siblings(`a[href*="wikidata.org/entity/${data.qid}"]`)))
+          );
+      })
+      .catch(() => $btn.prop("disabled", false));
+  }
+
+  function getLocalizedResponse(entity) {
+    const rank = ({ rank }) => ({ preferred: 1, normal: 0, deprecated: -1 })[rank] ?? 0;
+    const toBestClaim = (out, claim) => (rank(claim) > rank(out)) ? claim : out;
+    const toFirstOf = (property) => (out, localization) => out ?? entity[property][localization];
+    const data = {
+      qid: entity.id,
+      label: languagesToRequest.reduce(toFirstOf("labels"), null),
+      icon: [
+        "P8972", // small logo or icon
+        "P154", // logo image
+        "P14" // traffic sign
+      ].reduce((out, prop) => out ?? entity.claims[prop]?.reduce(toBestClaim)?.mainsnak?.datavalue?.value, null),
+      description: languagesToRequest.reduce(toFirstOf("descriptions"), null),
+      article: wikisToRequest.reduce(toFirstOf("sitelinks"), null)
+    };
+    if (data.article) data.article.language = data.article.site.replace("wiki", "");
+    return data;
+  }
+
+  function renderWikidataResponse({ icon, label, article, description }, $link) {
+    const cell = $("<td>")
+      .attr("colspan", 2)
+      .addClass("bg-body-tertiary");
+
+    if (icon && OSM.WIKIMEDIA_COMMONS_URL) {
+      let src = OSM.WIKIMEDIA_COMMONS_URL + "Special:Redirect/file/" + encodeURIComponent(icon);
+      if (!icon.endsWith(".svg")) src += "?width=128";
+      $("<a>")
+        .attr("href", OSM.WIKIMEDIA_COMMONS_URL + "File:" + encodeURIComponent(icon) + `?uselang=${OSM.i18n.locale}`)
+        .append($("<img>").attr({ src, height: "32" }))
+        .addClass("float-end mb-1 ms-2")
+        .appendTo(cell);
+    }
+    if (label) {
+      const link = $link.clone()
+        .text(label.value)
+        .attr("dir", "auto")
+        .appendTo(cell);
+      if (!isOfExpectedLanguage(label)) {
+        link.attr("lang", label.language);
+        link.after($("<sup>").text(" " + localeName.of(label.language)));
+      }
+    }
+    if (article) {
+      const link = $("<a>")
+        .attr("href", article.url + `?uselang=${OSM.i18n.locale}`)
+        .text(label ? OSM.i18n.t("javascripts.element.wikipedia") : article.title)
+        .attr("dir", "auto")
+        .appendTo(cell);
+      if (label) {
+        link.before(" (");
+        link.after(")");
+      }
+      if (!isOfExpectedLanguage(article)) {
+        link.attr("lang", article.language);
+        link.after($("<sup>").text(" " + localeName.of(article.language)));
+      }
+    }
+    if (description) {
+      const text = $("<div>")
+        .text(description.value)
+        .addClass("small")
+        .attr("dir", "auto")
+        .appendTo(cell);
+      if (!isOfExpectedLanguage(description)) {
+        text.attr("lang", description.language);
+      }
+    }
+    return $("<tr>").append(cell);
   }
 }());
--- a/app/helpers/browse_tags_helper.rb
+++ b/app/helpers/browse_tags_helper.rb
@@ -17,10 +17,16 @@ module BrowseTagsHelper
     elsif wdt = wikidata_links(key, value)
       # IMPORTANT: Note that wikidata_links() returns an array of hashes, unlike for example wikipedia_link(),
       # which just returns one such hash.
+      svg = button_tag :type => "button", :role => "button", :class => "btn btn-link float-end d-flex m-1 mt-0 me-n1 border-0 p-0 wdt-preview", :data => { :qids => wdt.map { |w| w[:title] } } do
+        tag.svg :width => 27, :height => 16 do
+          concat tag.title t("browse.tag_details.wikidata_preview", :count => wdt.length)
+          concat tag.path :fill => "currentColor", :d => "M0 16h1V0h-1Zm2 0h3V0h-3Zm4 0h3V0h-3Zm4 0h1V0h-1Zm2 0h1V0h-1Zm2 0h3V0h-3Zm4 0h1V0h-1Zm2 0h3V0h-3Zm4 0h1V0h-1Zm2 0h1V0h-1Z"
+        end
+      end
       wdt = wdt.map do |w|
         link_to(w[:title], w[:url], :title => t("browse.tag_details.wikidata_link", :page => w[:title].strip))
       end
-      safe_join(wdt, ";")
+      svg + safe_join(wdt, ";")
     elsif wmc = wikimedia_commons_link(key, value)
       link_to h(wmc[:title]), wmc[:url], :title => t("browse.tag_details.wikimedia_commons_link", :page => wmc[:title])
     elsif url = wiki_link("tag", "#{key}=#{value}")
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -136,6 +136,9 @@ overpass_credentials: false
 graphhopper_url: "https://graphhopper.com/api/1/route"
 fossgis_osrm_url: "https://routing.openstreetmap.de/"
 fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
+# Endpoints for Wikimedia integration
+wikidata_api_url: "https://www.wikidata.org/w/api.php"
+wikimedia_commons_url: "https://commons.wikimedia.org/wiki/"
 # Main website hosts to match in linkify
 linkify_hosts: ["www.openstreetmap.org", "www.osm.org", "www.openstreetmap.com", "openstreetmap.org", "osm.org", "openstreetmap.com"]
 # Shorter host to replace main hosts
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -19,6 +19,8 @@ OSM = {
       graphhopper_url
       fossgis_osrm_url
       fossgis_valhalla_url
+      wikidata_api_url
+      wikimedia_commons_url
     ]
       .each_with_object({}) do |key, hash|
         hash[key.to_s.upcase] = Settings.send(key) if Settings.respond_to?(key)
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -633,11 +633,14 @@ tr.turn {
 #sidebar_content {
   .browse-tag-list {
     table-layout: fixed;
-    white-space: pre-wrap;
-    word-wrap: break-word;
-    word-break: break-word;
 
-    tr:last-child th, tr:last-child td {
+    tr > *:not([colspan]) {
+      white-space: pre-wrap;
+      word-wrap: break-word;
+      word-break: break-word;
+    }
+
+    tr:last-child > * {
       border-bottom: 0;
     }
   }
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -246,8 +246,9 @@ class ApplicationController < ActionController::Base
   def map_layout
     policy = request.content_security_policy.clone
 
-    policy.connect_src(*policy.connect_src, "http://127.0.0.1:8111", Settings.nominatim_url, Settings.overpass_url, Settings.fossgis_osrm_url, Settings.graphhopper_url, Settings.fossgis_valhalla_url)
+    policy.connect_src(*policy.connect_src, "http://127.0.0.1:8111", Settings.nominatim_url, Settings.overpass_url, Settings.fossgis_osrm_url, Settings.graphhopper_url, Settings.fossgis_valhalla_url, Settings.wikidata_api_url)
     policy.form_action(*policy.form_action, "render.openstreetmap.org")
+    policy.img_src(*policy.img_src, Settings.wikimedia_commons_url, "upload.wikimedia.org")
     policy.style_src(*policy.style_src, :unsafe_inline)
 
     request.content_security_policy = policy
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -421,6 +421,9 @@ en:
         key: "The wiki description page for the %{key} tag"
         tag: "The wiki description page for the %{key}=%{value} tag"
       wikidata_link: "The %{page} item on Wikidata"
+      wikidata_preview:
+        one: "Wikidata item preview"
+        other: "Wikidata items preview"
       wikipedia_link: "The %{page} article on Wikipedia"
       wikimedia_commons_link: "The %{page} item on Wikimedia Commons"
       telephone_link: "Call %{phone_number}"
@@ -3526,6 +3529,8 @@ en:
       nothing_found: No features found
       error: "Error contacting %{server}: %{error}"
       timeout: "Timeout contacting %{server}"
+    element:
+      wikipedia: "Wikipedia"
     context:
       directions_from: Directions from here
       directions_to: Directions to here
--- a/test/helpers/browse_tags_helper_test.rb
+++ b/test/helpers/browse_tags_helper_test.rb
@@ -41,14 +41,20 @@ class BrowseTagsHelperTest < ActionView::TestCase
     assert_dom_equal "<a title=\"The Test article on Wikipedia\" href=\"https://en.wikipedia.org/wiki/Test?uselang=en\">Test</a>", html
 
     html = format_value("wikidata", "Q42")
-    assert_dom_equal "<a title=\"The Q42 item on Wikidata\" href=\"//www.wikidata.org/entity/Q42?uselang=en\">Q42</a>", html
+    dom = Rails::Dom::Testing.html_document_fragment.parse html
+    assert_select dom, "a[title='The Q42 item on Wikidata'][href$='www.wikidata.org/entity/Q42?uselang=en']", :text => "Q42"
+    assert_select dom, "button.wdt-preview>svg>path[fill]", 1
 
     html = format_value("operator:wikidata", "Q12;Q98")
-    assert_dom_equal "<a title=\"The Q12 item on Wikidata\" href=\"//www.wikidata.org/entity/Q12?uselang=en\">Q12</a>;" \
-                     "<a title=\"The Q98 item on Wikidata\" href=\"//www.wikidata.org/entity/Q98?uselang=en\">Q98</a>", html
+    dom = Rails::Dom::Testing.html_document_fragment.parse html
+    assert_select dom, "a[title='The Q12 item on Wikidata'][href$='www.wikidata.org/entity/Q12?uselang=en']", :text => "Q12"
+    assert_select dom, "a[title='The Q98 item on Wikidata'][href$='www.wikidata.org/entity/Q98?uselang=en']", :text => "Q98"
+    assert_select dom, "button.wdt-preview>svg>path[fill]", 1
 
     html = format_value("name:etymology:wikidata", "Q123")
-    assert_dom_equal "<a title=\"The Q123 item on Wikidata\" href=\"//www.wikidata.org/entity/Q123?uselang=en\">Q123</a>", html
+    dom = Rails::Dom::Testing.html_document_fragment.parse html
+    assert_select dom, "a[title='The Q123 item on Wikidata'][href$='www.wikidata.org/entity/Q123?uselang=en']", :text => "Q123"
+    assert_select dom, "button.wdt-preview>svg>path[fill]", 1
 
     html = format_value("wikimedia_commons", "File:Test.jpg")
     assert_dom_equal "<a title=\"The File:Test.jpg item on Wikimedia Commons\" href=\"//commons.wikimedia.org/wiki/File:Test.jpg?uselang=en\">File:Test.jpg</a>", html


tomhughes:
I don't understand how this relates to the code in that repository? Have you just copy and pasted it? Only it doesn't seem to match either file there as far as I can see but I also don't see how it refers to it... If you have copy and pasted it then did you use the chrome or firefox version? Why are there different versions?

Copy and pasting certainly isn't the way we want to be interacting with an upstream repository...

I'm also not sure we want to go down this path, both because it opens the door to lots of sites wanting to do the same thing and makes it hard to say no to others if we say yes here and secondly because depending on third party sites has risks - have we tested how this behaves if wikidata is down?

hlfan:
> did you use the chrome or firefox version? Why are there different versions?

CSP workaround for chrome that uses a service worker, rest of the files are identical. But both variations make direct stylings from js, I rewrote that to use bootstrap and jquery.

Brian wrote previously he doesn't see the necessary time investment worth it, so he wrote it as an extension instead.

> how this behaves if wikidata is down?

I that case or when the CSP isn't changed the script aborts. That's why I don't like the idea of integrating this further into the server side.

> both because it opens the door to lots of sites wanting to do the same thing and makes it hard to say no to others

Obviously a strong point, but I'd say right now Wikimedia has more OSM integration as the other way around. And I guess that could act as a criterion for accepting/declining.

1ec5:
> Copy and pasting certainly isn't the way we want to be interacting with an upstream repository...

If this becomes a built-in feature, I assume it would obviate the Chrome and Firefox plugins, right @ZeLonewolf? So probably no need to keep them in sync.

> Obviously a strong point, but I'd say right now Wikimedia has more OSM integration as the other way around.

Another way of looking at it is that editors like iD have long had this same functionality. Obviously a different set of consequences for any downtime, but some mappers would probably appreciate seeing the same kind of preview that they do while editing, as with the color swatches for `colour=*` and such.

ZeLonewolf:
Thanks for mentioning me. To be clear, @hlfan has taken on this PR without any input or involvement from me, which I welcome. I made those plugins CC0 so that someone motivated could work on integrating this functionality into the website without license restrictions.

> If this becomes a built-in feature, I assume it would obviate the Chrome and Firefox plugins, right @ZeLonewolf? So probably no need to keep them in sync.

Yes. If this functionality becomes integrated into osm.org, I would take down my Chrome and Firefox plugins and archive the source code repository.

As noted, there are separate Chrome and Firefox plugins because of differences in their plugin API models at the time that I developed the plugins. The actual functionality of doing the lookups up in wikidata and modifying the DOM are browser-independent.

hlfan:
> without any input or involvement from me, which I welcome

If you want to count https://c.osm.org/t/131530/55 as input you may.

mmd-osm:
According to https://www.mediawiki.org/wiki/Wikibase/API/en the endpoint supports multiple ids in one request, separated by a pipe character. Since we anyway know all ids upfront, we could do with a single page update (=less flickering in case of multiple * wikidata * tags).

AntonKhorev:
How about:
- not requesting anything without the user asking for it (try using this on element history pages)
- not rewriting the tag value
- not using blue background for something containing links

hlfan:
> Yes. If this functionality becomes integrated into osm.org, I would take down my Chrome and Firefox plugins and archive the source code repository.

Or, at least for this version, it just becomes a plugin that clicks each `.explainable` link on any mutation to circumvent the interaction restriction.

tomhughes:
> Another way of looking at it is that editors like iD have long had this same functionality.

I don't understand this comment because I don't see anything like this in iD?

hlfan:
> > Another way of looking at it is that editors like iD have long had this same functionality.
> 
> I don't understand this comment because I don't see anything like this in iD?

image
Not in the tags but in the fields, but the browse view doesn't have that.

AntonKhorev:
Text is aligned to three different left edges.

image

Maybe float the image to the right side?

app/assets/javascripts/index/element.js
```diff
@@ -23,6 +23,88 @@
...
+        $("<img>")
+          .attr({ src: OSM.WIKIMEDIA_COMMONS_URL + "Special:FilePath/" + data.icon, height: "32" })
+          .addClass("float-start mb-2 me-2")
+          .appendTo(cell);
+      }
+      if (data.label) {
+        labelLink.clone()
+          .text(data.label)
+          .removeAttr("title")
+          .addClass("ms-1")
```
AntonKhorev: What is this margin for? If the goal is to make it stand out more compared to the text below, maybe `fw-medium` will work better?

hlfan: That was originally on the article link to separate it, so that margin should be on the other side. Idk how I managed to change it halfway.

hlfan:
> Maybe float the image to the right side?

Done.

image

Also decreased the bottom margin to reduce unnecessarily text wrapping.

app/assets/javascripts/index/element.js
```diff
@@ -23,6 +23,88 @@
...
+      const wikis = [...new Set(langs.map(l => l.split("-")[0] + "wiki"))];
+      fetch(OSM.WIKIDATA_URL + "?" + new URLSearchParams({
+        action: "wbgetentities",
+        format: "json",
+        origin: "*",
+        ids: qid,
+        props: "labels|sitelinks|claims|descriptions",
+        languages: langs.join("|"),
+        sitefilter: wikis.join("|")
+      }))
```
AntonKhorev: I think we need:

1. Create an abort controller in `page.load`/`pushstate`/`popstate`.
2. Pass its signal to this fetch.
3. Abort in `page.unload`, this will stop all currently running fetches.

app/assets/javascripts/index/element.js
```diff
@@ -23,6 +23,90 @@
...
+        .addClass("bg-body-tertiary");
+      labelLink
+        .closest("tr")
+        .after($("<tr>").append(cell));
+
+      if (data.icon) {
+        $("<img>")
+          .attr({ src: OSM.WIKIMEDIA_COMMONS_URL + "Special:FilePath/" + data.icon, height: "32" })
+          .addClass("float-end mb-1 ms-2")
+          .appendTo(cell);
```
AntonKhorev: In case anyone asks for an image attribution, we can have it linked to its Commons page.

```suggestion
        $("<a>")
          .attr({ href: OSM.WIKIMEDIA_COMMONS_URL + "File:" + data.icon })
          .append(
            $("<img>")
              .attr({ src: OSM.WIKIMEDIA_COMMONS_URL + "Special:FilePath/" + data.icon, height: "32" }))
          .addClass("float-end mb-1 ms-2")
          .appendTo(cell);
```

app/assets/javascripts/index/element.js
```diff
@@ -23,6 +23,90 @@
   };
 
   OSM.MappedElement = function (map, type) {
+    let abortController = null;
+
+    function markWikidataLinkAsExplainable(i, link) {
```
AntonKhorev: Why is all of this wikidata code inside `OSM.MappedElement`? The idea behind `OSM.Element` vs `OSM.MappedElement` is that `OSM.MappedElement` has enough data to be displayed on the map. For wikidata you need access to tags, both `OSM.Element` and `OSM.MappedElement` have it.

A consequence of putting wikidata code inside `OSM.MappedElement` is that it won't run if you open an old version of a way or a relation, those don't fetch the data necessary for map rendering. But it will run for an old node.

config/settings.yml
```diff
@@ -136,6 +136,9 @@ overpass_credentials: false
 graphhopper_url: "https://graphhopper.com/api/1/route"
 fossgis_osrm_url: "https://routing.openstreetmap.de/"
 fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
+# Endpoints for Wikimedia integration
+wikidata_url: "https://www.wikidata.org/w/api.php"
```
AntonKhorev: A better name is `wikidata_api_url` because what if you want to generate a link to the wikidata web page, it won't have `w/api.php` in it. Or maybe add `w/api.php` in `explainWikidataLink` when building an url.

app/assets/javascripts/index/element.js
```diff
@@ -8,17 +8,116 @@
...
+      }
+      if (data.description) {
+        $("<div>")
+          .text(data.description)
+          .addClass("small")
+          .appendTo(cell);
+      }
+    }
+
+    function initWikidataLinks() {
```
AntonKhorev: I'd check here if wikidata settings are defined and quit if they are not (empty string in `wikidata_url` maybe? or null, but then you'd also have to update `application_controller.rb` to skip nils).

app/assets/javascripts/index/element.js
```diff
@@ -8,17 +8,117 @@
...
+      const markWikidataLinksAsExplainable = () => $("a[href*='wikidata.org/entity/Q']:not([class*='explain'])").each(markWikidataLinkAsExplainable);
+      abortController = new AbortController();
+      markWikidataLinksAsExplainable();
+      $("#sidebar-content").on("turbo:before-stream-render", event => {
+        const defaultRender = event.detail.render;
+        event.detail.render = function (streamElement) {
+          defaultRender(streamElement);
+          markWikidataLinksAsExplainable();
+        };
+      });
```
AntonKhorev: I was going to suggest another modification later after this PR is merged. The Q-links work differently depending on their state, they either open a wikidata page or add a row to the tags table. The state is invisible to the user, and this is not the most intuitive behaviour. So I was going to suggest to keep the Q-links working as before and have a button, likely with the wikidata logo, that adds a wikidata row to tags. The button could be placed where we have a color icon:

image

But looking at the code here, maybe it's easier to do this right away. We can have wikidata buttons rendered server-side with delegated click listeners. Then listening to turbo events that look for new Q-links won't be necessary.

AntonKhorev:
> I'm also not sure we want to go down this path, both because it opens the door to lots of sites wanting to do the same thing and makes it hard to say no to others if we say yes here

We can say that we won't add other data sources until we come up with some inclusion criteria. Wikidata would likely pass any reasonable criteria.

> and secondly because depending on third party sites has risks - have we tested how this behaves if wikidata is down?

Not a big problem if we don't load the data by default.

app/assets/javascripts/index/element.js
hlfan: I was going to ask about implementing the behaviour this way, but with your comment about `not rewriting the tag value` I decided to not bring it up.

If I select `$("q-link-icon")` to attach the listeners, that selection wont be touched by turbo, right?
So watch `#element_versions_list` instead?

AntonKhorev: I wouldn't say it rewrites the tag value. I think it's obvious what is a tag value and what is an icon on the side of the table cell, because tag values are text and not images. Previously you changed the text.

If you install the listener like this
https://github.com/openstreetmap/openstreetmap-website/blob/3aebcb18fdf437b6994866fc66120f48e37f0cd2/app/assets/javascripts/index/element.js#L2
the selection for whatever you write after `"click"` will happen inside the listener on `document`. Turbo stream is not rewriting the entire document killing off that listener. Although maybe `#sidebar` would have been better than `document` in this case, turbo stream doesn't rewrite it either.

app/helpers/browse_tags_helper.rb
```diff
@@ -17,10 +17,16 @@ def format_value(key, value)
     elsif wdt = wikidata_links(key, value)
       # IMPORTANT: Note that wikidata_links() returns an array of hashes, unlike for example wikipedia_link(),
       # which just returns one such hash.
+      svg = tag.a :href => "", :class => "wdt-preview", :data => { :qids => wdt.map { |w| w[:title] } } do
+        tag.svg :width => 27, :height => 14, :class => "float-end m-1" do
```
AntonKhorev: Can you float the entire link? Otherwise its focus outline looks like this:
image

```suggestion
      svg = tag.a :href => "", :class => "wdt-preview float-end m-1 d-flex", :data => { :qids => wdt.map { |w| w[:title] } } do
        tag.svg :width => 27, :height => 14 do
```

AntonKhorev: Actually why is it a link and not a button? It doesn't have `href`, right?

app/assets/javascripts/index/element.js
```diff
@@ -60,6 +66,92 @@
       }, { threshold });
     }
 
+    function initWikidataLinks() {
+      if (!OSM.WIKIDATA_API_URL) return;
+      abortController = new AbortController();
+      $("#sidebar_content").on("click", "a.wdt-preview", e => {
+        e.preventDefault();
+        previewWikidataValue($(e.currentTarget));
+      });
```
AntonKhorev: 1. Open any element page with wikidata.
2. Close the sidebar.
3. Press *back* to reopen the element page.
4. Click the *wikidata* button.

You'll see two wikidata rows added. That's because the click listener is installed twice, and you can repeat the procedure for more installs. This is why I install `.on("click", "a[href='#versions-navigation-active-page-item']", ...)` outside of `OSM.Element`, `.on()` runs only once there. `abortController` will also need to be moved out in this case, which should work fine because we have at most only one sidebar content running at the same time.

app/helpers/browse_tags_helper.rb
hlfan: I already sliced the pixel perfect wikidata logo to not bloat up the table row, a button requires some padding eating up more space.

Also which browser is this? When I distributed the classes the outline was in the right place for me.

AntonKhorev: Firefox on Ubuntu.

Buttons don't require padding, they have some by default which you can remove.

app/helpers/browse_tags_helper.rb
```diff
@@ -17,10 +17,16 @@ def format_value(key, value)
     elsif wdt = wikidata_links(key, value)
       # IMPORTANT: Note that wikidata_links() returns an array of hashes, unlike for example wikipedia_link(),
       # which just returns one such hash.
+      svg = tag.a :href => "", :class => "float-end m-1 d-flex wdt-preview", :data => { :qids => wdt.map { |w| w[:title] } } do
```
AntonKhorev: In Bootstrap if `<a>` is replaced with `<button type="button">`, these extra classes will make it look like a link:
`btn btn-link border-0 p-0`.

AntonKhorev: If you keep it as `<a>`, it's actually a link to the same page. You can middle-click it to open the page. Does that make sense?

hlfan: Yes, I just needed to re-read the first comment a few times to find out the n of In was missing.

AntonKhorev: For example, a screen reader is going to announce `<a>` as "Wikidata item preview visited link", which it isn't and if you visit it as a link, you're not getting wikidata item preview. `<button>` is announced as "Wikidata item preview push button".

hlfan: Your suggested solution changes the focused state to this: image
I haven't found a `disabled` and `outline` compatible button solution yet.

Firefishy:
Are WikiData ok with us doing this? Will they rate limit us like they do on wiki.osm.org with WikiCommons?

AntonKhorev:
> Are WikiData ok with us doing this? Will they rate limit us like they do on wiki.osm.org with WikiCommons?

Are they rate-limiting users who click the currently available Qnnnnn wikidata links?

AntonKhorev:
We're using two Wikimedia services her, Wikidata and Commons.

Wikidata usage should be ok except for [user agent header](https://www.mediawiki.org/wiki/API:Etiquette#The_User-Agent_header):

> It is best practice to set a descriptive User Agent header.
> [...]
> If you are calling the API from browser-based JavaScript, you may not be able to influence the User-Agent header, depending on the browser. To work around this, use the Api-User-Agent header.

So I guess we should add an `Api-User-Agent` header to the fetch in this PR. They have [examples](https://foundation.wikimedia.org/wiki/Policy:Wikimedia_Foundation_User-Agent_Policy#JavaScript).

AntonKhorev:
Hotlinking to images on Commons:

I'm not sure if it's allowed according to [their policy](https://commons.wikimedia.org/w/index.php?title=Commons:HOTLINK), but we're already doing it in iD. What iD does is closer to what they say is not allowed:

> Wikimedia generally does not allow 'hot spider' services, where each time someone performs a search on their site, the query is redirected to our site.

iD shows hotlinked images in response to typing in *Search feature type*,. Some of the images come from Facebook, some from Commons.

hlfan:
> we should add an Api-User-Agent header to the fetch in this PR

How does `OSM TagPreview (https://github.com/openstreetmap/openstreetmap-website)` look to you? Any other details I'm missing?

AntonKhorev:
Human-readable wikidata addon also limits image sizes (`?width=128`) which may result in transfers for raster images (https://www.openstreetmap.org/node/689097130 Starbucks logo: 19.40 kB in the addon vs 54.21 kB in this PR), but for svgs limiting widths may be counterproductive. If we decide to go with images we probably want smaller images.

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +110,83 @@
...
+    if (data.label) {
+      $btn
+        .siblings(`a[href*="wikidata.org/entity/${data.qid}"]`)
+        .clone()
+        .text(data.label)
+        .addClass("me-1")
+        .appendTo(cell);
+    }
+    if (data.article) {
+      $(`<${data.label ? "sup" : "div"}>`)
```
1ec5: The <sup>superscript</sup> suffix style probably comes from Wikidata’s former practice of glossing any fallback label with its fallback language, for example, “openstreetmap-website <sup>English</sup>” when you view [Q115259353](https://www.wikidata.org/wiki/Q115259353) in a language other than English. Otherwise, a user might not easily intuit that a suffix like “enwiki” means the title came from the English Wikipedia.

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +110,83 @@
...
+        .append($("<a>")
+          .attr("href", `https://${data.article.site.slice(0, -4)}.wikipedia.org/wiki/` + encodeURIComponent(data.article.title))
```
1ec5: The Wikidata API can return the item’s label in the relevant language. I think the label would be more relevant to OSM than the Wikipedia article, which often contains disambiguators and substitute characters for technical reasons. The label is designed to pair well with the description, which we’re using below. (Ideally, we’d fall back to the `mul` code to improve language coverage: ZeLonewolf/osm-wikidata-greasemonkey#9.)

If the idea is that the Wikidata item is already linked above, maybe we could link the label to the Wikidata item redundantly, then have a separate Wikipedia link right after it, reusing app/assets/images/social_link_icons/wikipedia.svg.

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +110,83 @@
...
+      })
+      .catch(() => $btn.removeClass("disabled").addClass("wdt-preview"));
+  }
+
+  function getLocalizedResponse(entity) {
+    const localizedProperty = (property, langs) => langs.reduce((out, lang) => out ?? entity[property][lang], null);
+    const data = {
+      qid: entity.id,
+      label: localizedProperty("labels", langs)?.value,
+      icon: ["P8972", "P154", "P14"].reduce((out, prop) => out ?? entity.claims[prop]?.[0]?.mainsnak?.datavalue?.value, null),
```
1ec5: For reference:

* [small logo or icon <small>(P8972)</small>](https://www.wikidata.org/wiki/Property:P8972)
* [logo image <small>(P154)</small>](https://www.wikidata.org/wiki/Property:P154)
* [traffic sign <small>(P14)</small>](https://www.wikidata.org/wiki/Property:P14)

P14 surprised me at first, but then I remembered that this would probably result in displaying a route marker alongside a `wikidata=*` tag when viewing a route relation. That does make some sense to me.

AntonKhorev:
> How does OSM TagPreview (https://github.com/openstreetmap/openstreetmap-website) look to you?

I think there shouldn't be a space in `OSM TagPreview`.

https://httpwg.org/specs/rfc9110.html#field.user-agent

AntonKhorev:
@Firefishy Did Wikimedia ever reacted to how we use their images in iD, or was it just in wiki? Or are they already rate-limiting us in iD too but it's not very noticeabe because images are small?

app/assets/javascripts/index/element.js
1ec5: In case the item has multiple statements associated with one of these properties, the `claims[prop]` array will contain multiple items, but the first one isn’t guaranteed to be the best. For example, [Deutsche Post <small>(Q157645)</small>](https://www.wikidata.org/wiki/Q157645#P154) has four different P154 statements. The last is the most current, so the `mainsnak` has a `rank` of `preferred` rather than `normal`.

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +110,83 @@
...
+  }
+
+  function renderWikidataResponse(data, $btn) {
+    const cell = $("<td>")
+      .attr("colspan", 2)
+      .addClass("bg-body-tertiary");
+
+    if (data.icon && OSM.WIKIMEDIA_COMMONS_URL) {
+      $("<a>")
+        .attr("href", OSM.WIKIMEDIA_COMMONS_URL + "File:" + data.icon)
```
1ec5: Do we need to escape the `icon` in any way?

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +110,83 @@
...
+        .append($("<img>").attr({ src: OSM.WIKIMEDIA_COMMONS_URL + "Special:FilePath/" + data.icon, height: "32" }))
```
1ec5: Special:FilePath is now an alias for Special:Redirect:

```suggestion
        .append($("<img>").attr({ src: OSM.WIKIMEDIA_COMMONS_URL + "Special:Redirect/file/" + data.icon, height: "32" }))
```

1ec5: > I'm not sure if it's allowed according to [their policy](https://commons.wikimedia.org/w/index.php?title=Commons:HOTLINK), but we're already doing it in iD.

As far as I know, Wikimedia would prefer that we use the Wikimedia Commons API to get the image thumbnail. For example, [this API call](https://commons.wikimedia.org/wiki/Special:ApiSandbox#action=query&format=json&prop=imageinfo&titles=File%3ADHL%20Group%2006.2023.svg&formatversion=2&iiprop=url%7Csize&iiurlwidth=32) gets a thumbnail of [this image](https://commons.wikimedia.org/wiki/File:DHL_Group_06.2023.svg) scaled to 32&nbsp;pixels wide, along with responsive images at 1.5× and 2×. The API allows batching up to 50 images at a time. That said, Special:Redirect is unlikely to be a problem if we’re only showing an individual image on demand. If we were showing a big gallery of images, that would be a different story.

1ec5: > Did Wikimedia ever reacted to how we use their images in iD, or was it just in wiki? Or are they already rate-limiting us in iD too but it's not very noticeabe because images are small?

Only on the wiki, because the requests to Wikimedia Commons are centralized through the wiki’s server. This client-side usage is different in that each individual user will request only the images they’re interested in.

app/assets/javascripts/index/element.js
hlfan: How about linking the label to the article to
1. not have redundant links and
2. not have to think about how to tell the user about the article?

Also if we use wikipedia.svg it should be inlined to be able to use currentColor.

app/assets/javascripts/index/element.js
hlfan: Is using `Intl.DisplayNames` better?

app/assets/javascripts/index/element.js
```diff
@@ -1,10 +1,16 @@
 (function () {
+  let abortController = null;
+  const langs = [...new Set([...OSM.preferred_languages.map(l => l.toLowerCase()), "mul", "en"])];
+  const wikis = [...new Set(langs.map(l => l.split("-")[0] + "wiki"))];
```
1ec5: Wikimedia wiki codes don’t quite conform to ISO&nbsp;639. For example, the response will include `be_x_oldwiki` for what this repository calls `be-Tarask`. But this will affect only a handful of locales that we have localizations for, so it probably isn’t critical.

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +110,97 @@
...
+    if (!OSM.WIKIDATA_API_URL) return;
+    const items = $btn.data("qids");
+    if (!items?.length) return;
+    $btn.addClass("disabled").removeClass("wdt-preview");
+    fetch(OSM.WIKIDATA_API_URL + "?" + new URLSearchParams({
+      action: "wbgetentities",
+      format: "json",
+      origin: "*",
+      ids: items.join("|"),
+      props: "labels|sitelinks|claims|descriptions",
```
1ec5: We can request `sitelinks/urls` here, so that each object in `sitelinks` will contain a `url` property set to the absolute URL to the article.

app/assets/javascripts/index/element.js
1ec5: Oh, I think I misread this part of the diff. I didn’t realize we’re always showing the label if available, followed by the superscript link to Wikipedia. I thought it was all one link based on an earlier screenshot.

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +110,97 @@
...
+    if (data.label) {
+      $btn
+        .siblings(`a[href*="wikidata.org/entity/${data.qid}"]`)
+        .clone()
+        .text(data.label)
+        .addClass("me-1")
+        .appendTo(cell);
+    }
+    if (data.article) {
+      const lang = data.article.site.slice(0, -4);
```
1ec5: Is this trimming “wiki” from the end of the sitename?

app/assets/javascripts/index/element.js
1ec5: This feature hooks into something labeled `wikidata` with a Wikidata icon, so there should probably be an explicit reference to Wikipedia to counter the assumption that the link goes to Wikidata. The reference to Wikipedia would be more relevant than the language name, since most of the time it’s going to be the user’s preferred language anyways. If we don’t want to bother with a Wikipedia icon, “(Wikipedia)” in regular text would communicate the link target pretty well.

app/assets/javascripts/index/element.js
hlfan: Yeah, is there a way to get the language label also from the API to not have to do that?

app/assets/javascripts/index/element.js
hlfan: If you'd like to have two links (Wikidata + Wikipedia), how should they be separated?
Putting the WP link at the end of the cell is also an option, but that still has to work if no description is displayed.

app/assets/javascripts/index/element.js
1ec5: I think this is fine, since lots of Wikimedia-related software makes the same assumption. It just needs a comment explaining the magic number −4. Or you could more explicitly replace “wiki” with the empty string.

app/assets/javascripts/index/element.js
1ec5: I guess I was thinking of something like this (for English), based on how Wikidata formats labels on its own site:

* [Bench](https://www.wikidata.org/wiki/Q816714) ([Wikipedia](https://en.wikipedia.org/wiki/Bench_(British_clothing_brand)))
* [Shell](https://www.wikidata.org/wiki/Q110716465)
* [Q111384397](https://www.wikidata.org/wiki/Q111384397)

The QID would be redundant to what’s in the normal tag view, though…

And for, say, Spanish with a fallback to English, something like this would be possible with what you’ve written:

* [Bench](https://www.wikidata.org/wiki/Q816714) <sup>English</sup>

hlfan: So Spanish falling back to French should look like this?
[Bench](https://www.wikidata.org/entity/Q816714?uselang=es)<sup> francés</sup> ([Wikipedia](https://fr.wikipedia.org/wiki/Bench_(entreprise)?uselang=es)<sup> francés</sup>)

1ec5: I guess that works, though I’m not sure it’s necessary to qualify the Wikipedia link with the language, since the URL already hints at the language. The main reason Wikidata appends the language name to labels is to avoid confusion when different languages have false cognates and such, but this isn’t a risk for a link with a fixed label or icon.

hlfan: If we want to qualify the label language, then imo the Wikipedia link should get the same treatment, because if they were added in a separate tag, the language would be indicated by a prefix like fr:...
Formatting it nicer and removing the prefix because "the URL already hints at the language" seems like a bad idea to me.

In the Wikipedia you can find links like `redlink [es]` with a interwikilink to the other language. This and the reason Wikidata is language agnostic is why I put the enwiki site code here in the first place.

1ec5: OK, that’s fair. I do see a lot of second- and third-party Wikidata tools labeling links like “Wikipedia [es]” as well, with the language code wrapped in an `<abbr>` tag that provides a tooltip for the human-readable language name. We don’t have any precedent for that in this codebase as far as I know, but personally I find it self-explanatory.

On that note, we should probably surround any foreign-language text with an element that specifies a `lang=` attribute for proper font selection and screen reader support.

hlfan: I've added `lang=` and `dir=auto` for language fallbacks. I put `dir=auto` instead of `<bdi>` to avoid additional nesting, even though I haven't seen the bda fail on this yet.

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +112,108 @@
       scrollableList.scrollLeft = scrollableList.scrollWidth - scrollableList.offsetWidth;
     }
   }
+
+  function previewWikidataValue($btn) {
+    if (!OSM.WIKIDATA_API_URL) return;
+    const items = $btn.data("qids");
+    if (!items?.length) return;
+    $btn.addClass("disabled").removeClass("wdt-preview");
```
AntonKhorev: Was [this](https://github.com/openstreetmap/openstreetmap-website/pull/6127#discussion_r2180631919) about that a disabled button still gets a focus outline? Then this:

```suggestion
    $btn.prop("disabled", true);
```

and later
```js
      .catch(() => $btn.prop("disabled", false));
```

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +112,108 @@
...
+    if (data.label) {
+      const label = $btn
+        .siblings(`a[href*="wikidata.org/entity/${data.qid}"]`)
+        .clone()
+        .text(data.label.value)
+        .attr("dir", "auto");
+      cell.append($("<span>").append(label));
+      if (isUnexpectedLanguage(data.label.language)) {
+        label.attr("lang", data.label.language);
+        label.after($("<sup>").text(" " + localeName.of(data.label.language)));
```
AntonKhorev: I had to figure out what "Multiple languages" means on the third item:
image

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +112,108 @@
...
+      }
+    }
+    if (data.article) {
+      const label = $("<a>")
```
AntonKhorev: Is it "label" or "article"?

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +112,108 @@
...
+        .addClass("float-end mb-1 ms-2")
+        .appendTo(cell);
+    }
+    if (data.label) {
+      const label = $btn
+        .siblings(`a[href*="wikidata.org/entity/${data.qid}"]`)
+        .clone()
+        .text(data.label.value)
+        .attr("dir", "auto");
+      cell.append($("<span>").append(label));
```
AntonKhorev: What is this `<span>` for?

app/assets/javascripts/index/element.js
hlfan: I'd like the button when focused to have an outline as that's more visible than the slight blue background.

The improvement to the state management here is great.

app/assets/javascripts/index/element.js
hlfan: Oh, yeah, names and such using the mul language code shouldn't get a label.

app/assets/javascripts/index/element.js
hlfan: Well, both should be "link", huh?

app/assets/javascripts/index/element.js
hlfan: To group label elements and article elements. Not neeeded, removed.

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +112,108 @@
...
+    if (!items?.length) return;
+    $btn.prop("disabled", true);
+    fetch(OSM.WIKIDATA_API_URL + "?" + new URLSearchParams({
+      action: "wbgetentities",
+      format: "json",
+      origin: "*",
+      ids: items.join("|"),
+      props: "labels|sitelinks/urls|claims|descriptions",
+      languages: langs.join("|"),
+      sitefilter: wikis.join("|")
```
AntonKhorev: This parameter is going to be something like `sitefilter=enwiki|mulwiki`.

What is `mulwiki`? I don't see it among [the possible values](https://www.wikidata.org/w/api.php?action=help&modules=wbgetentities).

app/assets/javascripts/index/element.js
```diff
@@ -1,10 +1,18 @@
 (function () {
+  let abortController = null;
+  const langs = [...new Set([...OSM.preferred_languages.map(l => l.toLowerCase()), "mul", "en"])];
```
AntonKhorev: Can we give this a more specific name than `langs` to avoid guessing what it actually means? Is it *preferred languages from mediawiki point of view*?

app/assets/javascripts/index/element.js
```diff
@@ -101,4 +112,108 @@
...
+              .map(qid => getLocalizedResponse(entities[qid]))
+              .filter(data => data.label || data.icon || data.description || data.article)
+              .map(data => renderWikidataResponse(data, $btn.siblings(`a[href*="wikidata.org/entity/${data.qid}"]`)))
+          );
+      })
+      .catch(() => $btn.prop("disabled", false));
+  }
+
+  function getLocalizedResponse(entity) {
+    const localizedProperty = (property, langs) => langs.reduce((out, lang) => out ?? entity[property][lang], null);
```
AntonKhorev: See my previous comment about `langs`.

`langs` here are either `langs` there or `wikis` there. Are `wikis` `langs`?

app/assets/javascripts/index/element.js
hlfan: Filtered out.

AntonKhorev:
Merged, thanks.

For admins:
if Wikimedia wouldn't like hotlinking to their images, set `wikimedia_commons_url` to an empty string. But we're doing this and more from iD already.

app/assets/javascripts/index/element.js
hlfan: That worked out well, didn't it? (#6202)

app/assets/javascripts/index/element.js
```diff
@@ -1,10 +1,18 @@
 (function () {
+  let abortController = null;
+  const languagesToRequest = [...new Set([...OSM.preferred_languages.map(l => l.toLowerCase()), "mul", "en"])];
```
1ec5, line 3: @lucaswerkmeister [suggested](https://wikis.world/@LucasWerkmeister/114876146945706671) passing in `languagefallback=1` instead of manually requesting a list of languages. I believe it would cause the response to use MediaWiki’s internal [language fallback chain](https://meta.wikimedia.org/wiki/Wikidata/Notes/Language_fallback) rather than the OSM setting or browser setting. @hlfan would it be a better match for the intended behavior here?

PR #6128, opened 2025-06-24
Move element / element history sidebar code to its own file

AntonKhorev:
Part 1 of #6120.

Creates `element.js` and moves `OSM.Browse` and `OSM.OldBrowse` there.

Everything in `element.js` is wrapped in iife which is currently not required, but #6120 later adds functions there; this is to avoid reindenting the code later.


tomhughes:
Looks good to me, thanks.

PR #6129, opened 2025-06-24
Element version nav  refactor tests

AntonKhorev:
Part 2 of #6120.


hlfan:
Could the 3 tests be split up over node/way/relation to have 9 files instead of three over 300 lines long ones?

AntonKhorev:
The tests are similar for different types and likely could be refactored, it just wasn't a priority. "over 300 lines" is short when compared to other test files.

test/controllers/nodes_controller_test.rb
```diff
@@ -13,82 +13,6 @@ def test_routes
...
-  def test_show_deleted
-    node = create(:node, :visible => false)
-    sidebar_browse_check :node_path, node.id, "elements/show"
-    assert_select "h4", /^Version/ do
-      assert_select "a[href='#{old_node_path node, 1}']", :text => "1", :count => 1
-    end
-    assert_select "a[href='#{api_node_path node}']", :count => 0
-  end
-
-  def test_show_redacted
```
tomhughes, line 49: There doesn't seem to be any replacement for this test in the new system tests?

AntonKhorev: Because you can't view a redacted version as a current version. Current version can't be redacted. When looking at a current version, you wouldn't know which one, if any, of past versions was redacted.

tomhughes: Yes you're right - it's quite misnamed because what it's really testing is that we don't show a location for a deleted node. The fact that the node has a prior redacted version isn't really relevant.

tomhughes:
So I think is this good to merge now, thanks.

PR #6134, opened 2025-06-25
Element version navigation steps

AntonKhorev:
Next improvement after #6120, in case there's a lot of skipped versions.

Changes version lists like [1, ..., 234, 235, 236, 237] to [1, ..., 50, ..., 100, ..., 150, ..., 200, 234, 235, 236, 237].


tomhughes:
Looks good to me, thanks.

PR #6136, opened 2025-06-26
Follow redirects to suspended page in spammy tests

AntonKhorev:
You don't need to write comments about following the redirect if you actually `follow_redirect!`.


tomhughes:
Looks good to me, thanks.

PR #6137, opened 2025-06-27
Add Shortbread-based vector tile layer

pnorman:
Based on #4042 and https://github.com/openstreetmap/openstreetmap-website/pull/4042/files#r1966844335

This PR adds a Leaflet.Maplibregl-based layer using Versatiles Colorful.

The issue I'm having right now is relative URLs. To use relative URLs for sprites, tiles, and glyphs, you need to use [`TransformStyleFunction`](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/TransformStyleFunction/). This can only be supplied to [`Map.setStyle()`](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#setstyle).

In the demo page I did this [fairly easily](https://github.com/openstreetmap/vectortile-website/blob/6d459cec4940e8488d0f7695940256f5f5dc8feb/demo/shortbread/index.html#L62-L69), but I wasn't dealing with Leaflet.Maplibregl.

I'm not sure what the best approach here is. Some options I considered are
1. making colorful.json use absolute paths. This has a few secondary effects
   a. I'd still need to use `TransformStyleFunction` on the demo page since vector.openstreetmap.org should load tiles/sprites/glyphs via the CDN and dribble.openstreetmap.org should load tiles/sprites/glyphs off of dribble
   b. I'd want to put CORS restrictions on `https://vector.openstreetmap.org/demo/shortbread/colorful.json` restricting it to OSM domains and testing domains (e.g. localhost). Users of the tile layer should be hosting the style on their website, not depending on what we're hosting
2. Figuring out how to use `TransformStyleFunction` in Leaflet. I'm not sure if this is possible or where to start.

I did consider building the colorful.json style in this repo but decided against it. Some aspects of the style (glyphs, sprites) will have to be built externally no matter what, and it's best to keep the style JSON build with those other parts.

What I need is some input on what approach to take with the relative/absolute URLs.

### DIFF
--- a/app/assets/javascripts/leaflet.shortbread.js
+++ b/app/assets/javascripts/leaflet.shortbread.js
@@ -1,0 +1,30 @@
+//= require leaflet.maplibre
+
+L.OSM.Shortbread = L.MaplibreGL.extend({
+  onAdd: function (map) {
+    L.MaplibreGL.prototype.onAdd.call(this, map);
+    const styleURL = "https://vector.openstreetmap.org/demo/shortbread/colorful.json";
+    this.getMaplibreMap().setStyle(styleURL, {
+      transformStyle: (previousStyle, nextStyle) => ({
+        ...nextStyle,
+        sprite: [...nextStyle.sprite.map(s => {
+          return {
+            ...s,
+            url: new URL(s.url, styleURL).href
+          };
+        })],
+        // URL will % encode the {} in glyph and source URL so assemble them manually
+        glyphs: (new URL(styleURL)).origin + nextStyle.glyphs,
+        sources: {
+          "versatiles-shortbread": {
+            ...nextStyle.sources["versatiles-shortbread"],
+            tiles: [(new URL(styleURL)).origin + nextStyle.sources["versatiles-shortbread"].tiles[0]]
+          }
+        }
+      })
+    });
+  },
+  onRemove: function (map) {
+    L.MaplibreGL.prototype.onRemove.call(this, map);
+  }
+});
--- a/config/layers.yml
+++ b/config/layers.yml
@@ -79,3 +79,12 @@
       osm_france_link:
         id: "osm_france"
         href: "https://openstreetmap.fr/"
+
+- leafletOsmId: "Shortbread"
+  code: "S"
+  layerId: "shortbread"
+  nameId: "shortbread"
+  credit:
+    id: "make_a_donation"
+    href: "https://supporting.openstreetmap.org"
+    donate: true
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -20,3 +20,4 @@
 //= link leaflet/dist/images/marker-icon.png
 //= link leaflet/dist/images/marker-icon-2x.png
 //= link leaflet/dist/images/marker-shadow.png
+//= link @mapbox/mapbox-gl-rtl-text/dist/mapbox-gl-rtl-text.js
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -7,6 +7,7 @@
 //= require osm
 //= require leaflet/dist/leaflet-src
 //= require leaflet.osm
+//= require leaflet.shortbread
 //= require leaflet.map
 //= require leaflet.zoom
 //= require leaflet.locationfilter
--- a/app/assets/javascripts/leaflet.maplibre.js
+++ b/app/assets/javascripts/leaflet.maplibre.js
@@ -1,0 +1,4 @@
+//= require maplibre-gl
+//= require @maplibre/maplibre-gl-leaflet
+
+maplibregl.setRTLTextPlugin(OSM.RTL_TEXT_PLUGIN, true);
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -43,6 +43,8 @@ OSM = {
   OPEN_NOTE_MARKER: <%= image_path("open_note_marker.svg").to_json %>,
   CLOSED_NOTE_MARKER: <%= image_path("closed_note_marker.svg").to_json %>,
 
+  RTL_TEXT_PLUGIN: <%= javascript_path("@mapbox/mapbox-gl-rtl-text/dist/mapbox-gl-rtl-text.js").to_json %>,
+
   apiUrl: function (object) {
     const apiType = object.type === "note" ? "notes" : object.type;
     let url = "/api/" + OSM.API_VERSION + "/" + apiType + "/" + object.id;
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -245,8 +245,8 @@ class ApplicationController < ActionController::Base
 
   def map_layout
     policy = request.content_security_policy.clone
-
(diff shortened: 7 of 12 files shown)


tomhughes:
I think the options for dealing with the transformation are:

* Don't set the style in the options so the maplibre map object is created with no style and then change your `onAdd` method to do `this.getMaplibreMap().setStyle(...)` after calling the super class `onAdd`
* Open a PR to maplibre-gl-js to add extend the `Map` constructor to take the transform option and pass it on to `setStyle`

pnorman:
I've got it working now. When this is ready for merging I'll rebase and squish before taking it out of draft.

How should I handle updating yarn.lock? Should I just accept all the changes?

tomhughes:
If the `yarn.lock` changes are to add the two extra modules and any dependencies they have then sure.

pnorman:
I'm not sure what the CI failure is from.
```
Sprockets::FileNotFound: couldn't find file '@mapbox/mapbox-gl-rtl-text/mapbox-gl-rtl-text.js' (Sprockets::FileNotFound)
```

I don't get this in my local environment or see it in #4042

tomhughes:
Shouldn't the path be `@mapbox/mapbox-gl-rtl-text/dist/mapbox-gl-rtl-text.js`? I don't see the version without `dist`

hlfan:
> When this is ready for merging I'll rebase and squish before taking it out of draft.

Moving the layer definition out of [leaflet.map.js](https://github.com/openstreetmap/openstreetmap-website/pull/6137/files#diff-f060eef487077b042f45b3a12e7a2cbd54a446f8c744cdb98405a1295430e8b9) into [leaflet.osm.vector.js](https://github.com/openstreetmap/openstreetmap-website/pull/4042/files#diff-d3683fffa71b6e90481fbdb28ba0d272249bc9f6fd6d20ae2812e04c5a5c7542) before that would be great.

app/assets/javascripts/leaflet.map.js
hlfan, line 1: See https://github.com/openstreetmap/openstreetmap-website/pull/6137#issuecomment-3062857646, no other layers are defined in this file

tomhughes: I think I would suggest actually creating two new files - a `leaflet.maplibregl.js` that does the maplibre requires and global configuration (basically just configuring the RTL plugin) and ` leaflet.shortbread.js` or something that requires `leaflet.maplibregl.js` and then defines the layer class.

That file should then be included in `application.js` alongside `leaflet.osm` that defines the other layers.

app/assets/javascripts/leaflet.map.js
```diff
@@ -1,10 +1,45 @@
...
+//= require @maplibre/maplibre-gl-leaflet
+
 L.extend(L.LatLngBounds.prototype, {
   getSize: function () {
     return (this._northEast.lat - this._southWest.lat) *
            (this._northEast.lng - this._southWest.lng);
   }
 });
 
+if (OSM.SHORTBREAD_STYLE_URL) {
```
tomhughes: The problem with making this conditional is that if there is no URL configured the class won't be defined but `layers.yml` will still reference the layer and hence we will try and use the class.

We either need to say the URL is mandatory, or we extend the layer definitions to allow the layer to be skipped.

hlfan: That should be doable by adding this single line to `map_layers.rb`:
```ruby
.reject { |layer| layer["layerId"] == "shortbread" && !Settings["shortbread_style_url"] }
```

tomhughes: Sure. I wonder why the URL needs to be configurable though? None of the other layers is?

hlfan: Well, there's already quite a section in the `embed.js` just to allow overriding the standard layer's URL with `Settings.tile_cdn_url`. I think for the OSMF-provided services, a tighter integration is better.

tomhughes: Yes I was looking at that earlier but as far as I can tell we don't actually use that... Plus as you say it's only used in embed...

Looking further it was used for a few months in late 2020 when we were moving the CDN to fastly to point embed at the fastly CDN instead of our old CDN for testing I think.

app/assets/javascripts/leaflet.shortbread.js
```diff
@@ -0,0 +1,30 @@
+//= require leaflet.maplibre
+
+const styleURL = "https://vector.openstreetmap.org/demo/shortbread/colorful.json";
```
hlfan: Why is this a global constant?

mmd-osm:
Did you plan to integrate the language selector as well, to show the `lang_en` text version? I think this would be a nice addition, and frequently requested since about 1872.

image

pnorman:
> Did you plan to integrate the language selector as well

No. I don't plan to look at any language stuff until that stuff has been defined in the shortbread spec

app/assets/javascripts/leaflet.map.js
hlfan: What do you think of swapping the hierarchy of the new files, with `leaflet.shortbread.js` required in `leaflet.maplibregl.js`?

The current way would need to require `leaflet.maplibregl.js` twice,
once in `leaflet.shortbread.js` and once in a possible `leaflet.openmaptiles.js`.

config/layers.yml
```diff
@@ -79,3 +79,12 @@
       osm_france_link:
         id: "osm_france"
         href: "https://openstreetmap.fr/"
+
+- leafletOsmId: "Shortbread"
+  code: "S"
+  layerId: "shortbread"
+  nameId: "shortbread"
```
AntonKhorev, line 86: You can add `maxZoom: 22` here because various things don't work on higher zooms.

For example, you'll get gigantic scale bars, and Leaflet doesn't care about their `maxWidth` option below 1 meter.
image

tomhughes: The maptiler PR went with 23 as a maximum, and encoded it in an overriden constructor to make it the default for the layer.

tomhughes:
Other than the zoom question, which is a minor issue that we can fix in a followon PR if necessary I think this is good to merge. Are you happy that this is ready to go for production on the server side @pnorman?

pnorman:
I'm happy on the server side

AntonKhorev:
> Other than the zoom question, which is a minor issue

Have you tried it on low zooms?

image

You can get gigantic scale bars on any zoom level.

On low zooms you get jumping after zoom to constrain the map to the window size. The actual scale on zoom 0 depends entirely on the window size.

config/layers.yml
AntonKhorev: Maptiler went with 23 for `L.OSM.OpenMapTiles`, not for `L.OSM.Shortbread`.

hlfan:
> You can get gigantic scale bars on any zoom level.

So why limit it then? You can get gigantic scale bars on any current layer too.
image

tomhughes:
Well no I didn't try it on low zooms because you were talking about the maxZoom limit... I can't reproduce it on low zooms even with a very small window size and in any case I don't see why this layer would be different to any others other than that it allows larger zooms.

Thanks for your work on this @pnorman and I still think this is OK to merge in it's current state.

Issue #6142, opened 2025-06-28
Element version navigation: Non-functional scrollbar on some versions, divider removed on node and history pages


danieldegroot2:
### How to reproduce the issue?

Ref #6120

I was looking at an object and noticed the new element navigation.
- Since the header 'version 3' and the version list are not directly next to each other it was not immediately clear this navigation was for the version list. (the old navigation had 'version' written next to both previous and next link)
- Since it looks similar to node/way/relation list pages on changesets, it was weird to see a pagination list with no list of items below it, instead there is some amount of empty white space.
- Actually, it seems that on object's 'Node' and 'History' pages, the footer divider above the navigation and xml download button (horizontal grey bar) is removed. (this is not the case on the 'Version' pages)

I moved to looking at another object with 7 versions. Below the version list are previous/next arrows which do nothing when clicked. These are for the scrollbar which appears on objects with many versions. (there is no scrollbar)

Browser: Chrome
OS: Windows 10

cc: @AntonKhorev

### Screenshot(s) or anything else?

Object with 'Version 3' https://www.openstreetmap.org/node/11371833435
![Image](https://github.com/user-attachments/assets/e65d0bf4-466d-4843-af0a-6443832d0200)

Compare with 'Version' page (same for version 1 and 2) https://www.openstreetmap.org/node/11371833435/history/3
![Image](https://github.com/user-attachments/assets/630d33f5-754b-490a-9fde-de69e6b339aa)

Object with seven versions; example https://www.openstreetmap.org/node/8274525075
![Image](https://github.com/user-attachments/assets/a8fb62fd-ef36-4c6e-a27c-c652970a5267)

AntonKhorev:
This is four different issues, some are debatable. If you want to discuss the following, please reopen separate issues for them:
- "it was weird to see a pagination list with no list of items below it" - maybe will get moved to the top but it was at the bottm previously, some people may find moving it surprising
- "it was not immediately clear this navigation was for the version list" - maybe "Versions:" will get added above the list after a moderator-only link is moved, but titles saying "Version #n" already exist

danieldegroot2:
@AntonKhorev Sorry for not making them separate issues and not properly separating reproduction steps and issues.

It is understandable you went with this design.

See #6146 for second part of this issue.

danieldegroot2:
From [ceirios](https://www.openstreetmap.org/user/ceirios) on OSM World Discord (partially via private message)

[...]

PR #6143, opened 2025-06-28
Remove .browse-section and other css classes previously used for testing

AntonKhorev:
`.browse-section` has some custom css associated with it but it's now outdated, including one issue mentioned in #6142, "the footer divider".

`.browse-node`, `.browse-way` and `.browse-relation` are not necessary after #6129 because they were only used in tests. `.browse-redacted` is already removed in #6129.


tomhughes:
Looks good to me, thanks.

PR #6144, opened 2025-06-28
Make element version scrollbar appear only when necessary

AntonKhorev:
Mentioned in #6142, "Non-functional scrollbar".

This is for browsers that display scrollbars not only during scrolling. On Linux this is currently Chrome, but not Firefox.

If the version list is long enough to have the potentially scrollable part (currently >= 5 versions) but it fits in the sidebar entirely without scrolling, the scrollbar may still be visible.

image

This PR sets the horizontal overflow to `auto` instead of `scroll`, which hides the scrollbar when there's nothing to scroll.

image


HolgerJeromin (approved):

hlfan (approved):

tomhughes:
Looks good to me, thanks.

PR #6148, opened 2025-06-29
Add versions label to breadcrumbs

### LINKED ISSUE #6146: Element history: Add header to versions list
### Problem

Was https://github.com/openstreetmap/openstreetmap-website/issues/6142#issuecomment-3015699689

From [ceirios](https://www.openstreetmap.org/user/ceirios) on OSM World Discord (partially via private message)

> I quite like the new browser, it's growing on me [...] its easier [to use]
> im not sure how i feel about the design [...] I do agree that it needs a "Versions:" header

On the element node/history/version pages, the element version list appears on a new line with barely to no text indicating it lists element version numbers, not something else.
(initially this is unclear. though after a moment, users will figure out it obviously can't list anything else)

Example https://www.openstreetmap.org/node/8274525075

If/where it fits, add an explicit heading 'Versions(:)' above the versions list, to clearly indicate it lists element versions.

The version list itself can stay at the bottom of the page.

---

Only on individual version pages there is text directly above the list beside Node/History which mentions current version (yes, this is also the heading of the page itself, but it is nowhere near the version list)

sidenote: Anton mentioned change was currently blocked by removing moderator-only link.

### Description

_No response_

### Screenshots

_No response_

AntonKhorev:
Adds a *Versions:* label above the versions navigeation list on element pages.

Fixes #6146.

Before for most of the users:
image

After:
image

Before for moderators:
image

After:
image

### DIFF
--- a/app/views/elements/show.html.erb
+++ b/app/views/elements/show.html.erb
@@ -7,6 +7,10 @@
 <% if @feature.visible? %>
   <div class='secondary-actions mb-3'>
     <%= link_to(t("browse.download_xml"), :controller => "api/#{@type.pluralize}", :action => :show) %>
+    <% if current_user&.moderator? %>
+      &middot;
+      <%= link_to t("browse.view_unredacted_history"), :controller => "old_#{@type.pluralize}", :params => { :show_redactions => true } %>
+    <% end %>
   </div>
 <% end %>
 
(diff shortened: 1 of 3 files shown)


tomhughes:
I assume it's deliberate that the new label is styled as body text even when it's a link?

AntonKhorev:
Yes because it's not a "real" link, it's handled by javascript, like the one on an element version page:

image

A real link would write its anchor to the page url, something I don't want to happen.

tomhughes:
Great. I think this looks good then.

PR #6149, opened 2025-06-30
Introduce scoped placeholder `type:` for element routes

hlfan:
### Description
Undoes 223d5fd13245308eaf0ee92e8f946aab078f4850 while introducing a Ruby word array inspired enum placeholder.

### How has this been tested?
Local installation, DevTools & CI


app/assets/javascripts/index.js
```diff
@@ -300,12 +300,8 @@ $(function () {
...
     "/user/:display_name/history": history,
     "/note/:id": OSM.Note(map),
-    "/node/:id(/history)": OSM.MappedElement(map, "node"),
-    "/node/:id/history/:version": OSM.MappedElement(map, "node"),
-    "/way/:id(/history)": OSM.MappedElement(map, "way"),
-    "/way/:id/history/:version": OSM.Element(map, "way"),
-    "/relation/:id(/history)": OSM.MappedElement(map, "relation"),
-    "/relation/:id/history/:version": OSM.Element(map, "relation"),
+    "/type:node way relation/:id(/history)": OSM.MappedElement(map),
+    "/type:node way relation/:id/history/:version": OSM.Element(map),
```
AntonKhorev: Did you notice that there are four `MappedElement`s and two `Element`s?

hlfan: Now I did...
Yet another point that the syntax as it is obscures this a bit

app/assets/javascripts/router.js
```diff
@@ -7,8 +7,10 @@
   of the above pages support.
 
   The router is initialized with a set of routes: a mapping of URL path templates
-  to route controller objects. Path templates can contain placeholders
-  (`/note/:id`) and optional segments (`/:type/:id(/history)`).
+  to route controller objects. Path templates can contain
+    placeholders (`/note/:id`),
+    scoped placeholders (`/type:node way relation/:id`) and
```
AntonKhorev, line 12: The current syntax is similar to Rails routes:
- https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Base.html#method-i-match
- https://guides.rubyonrails.org/routing.html#non-resourceful-routes

If we want to stick to that, we need something like [segment constraints](https://guides.rubyonrails.org/routing.html#segment-constraints).

PR #6151, opened 2025-06-30
Stop GitHub from cancelling related workflows early

hlfan:
Most workflow cancellations occur just seconds before they would finish anyway, so failing fast doesn't safe many resources, it just makes workflow reattempts consume more.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,7 @@ jobs:
     strategy:
       matrix:
         ruby: ['3.2', '3.3', '3.4']
+      fail-fast: false
     runs-on: ubuntu-latest
     env:
       RAILS_ENV: test


AntonKhorev:
Or we can make the tests deterministic by merging #6089 and #6084. But instead we decided to have a discussion about how many sections should there be in the sidebar.

hlfan:
This doesn't have to be an exclusive or, does it?

hlfan:
Well merging #6089 didn't fix all the tests, so that's no alternative.

AntonKhorev:
If workflows fail only on real bugs, it makes sense to terminate early. Or if they fail because some external service like Coveralls is down.

hlfan:
> If workflows fail only on real bugs, it makes sense to terminate early.

That seems like an argument for #6154 to me.

I'll get back to you on this once real bug caused failures like in this [Docker workflow](https://github.com/openstreetmap/openstreetmap-website/actions/runs/16006135781/job/45153148119#step:8) make its way to the Test workflow.

hlfan:
Why should we cancel successfully run steps like [this one](https://github.com/hlfan/openstreetmap-website/actions/runs/16058832401/job/45319779765#step:13:16)?

AntonKhorev:
The step was either successful or it was cancelled.

The entire run was cancelled to not waste time running tests that need to be rerun after fixing bugs that are already found by other tests (assuming https://github.com/openstreetmap/openstreetmap-website/pull/6151#issuecomment-3025978203, in which case you are unlikely to have a successful run).

hlfan:
Let's take a look at the latest test run that didn't fail on "real bugs", [this test run](https://github.com/openstreetmap/openstreetmap-website/actions/runs/16122953585) for #6170:

The three executions of `rails test:all` finished in 297.341805s, 281.573575s, and 298.088535s.
Once this step finishes, the remainder of the workflow usually takes less than 15 seconds.

Now, one Ruby run fails:
- Either the other runs can complete, costing about 20 seconds of compute time plus the re-run of one job.
- Or this run cancels the other runs, costing the re-run of two jobs.

Unless you can make the jobs take less than 20 seconds, the current variant takes more time in this example.

---

Due to the usually parallel execution of the test runs, unless the error happens within the first two minutes of the workflow during the setup, `rails test:all` will only report an error after about seven minutes into the run, when the other runs are also almost complete.
Thus, the "wasted [compute] time running tests that need to be rerun" will be bigger **with** `fail-fast`, **unless** test-caused (not bug-caused) failures occur more often than once in every 21 runs — based on the fact that cancelling other jobs saves only about 20 seconds per job, while letting them finish typically costs around 7 minutes total.

And currently, I don't see test failures being even close to the necessary 95.2% bug-caused failure rate.
Therefore, `fail-fast` currently **increases** total compute time, because most jobs are cancelled _just seconds before completion_, and have to be re-run anyway. So we end up paying twice.

AntonKhorev:
> And currently, I don't see test failures being even close to the necessary 95.2% bug-caused failure rate.

The premise of [your question](https://github.com/openstreetmap/openstreetmap-website/pull/6151#issuecomment-3019543128) is that there are only "real bugs" failures.

Why is it important to run the test to its completion if you're going to run it again anyway?

https://github.com/openstreetmap/openstreetmap-website/actions/runs/16122953585 was supposedly fixed by https://github.com/openstreetmap/openstreetmap-website/commit/33f74d5725bd9825c014b2394c8f20ad0ac2cfc4 (or another similar issue was fixed, I think there was a couple of those). You had to go back in history for five days to find this failed run.

hlfan:
> Why is it important to run the test to its completion if you're going to run it again anyway?

image

There's an option not to.

I'm not saying we shouldn't improve our tests (#6022), but there are cases where the fault isn't in this codebase.
Luckily, the timing worked out not resulting in a run cancellation in https://github.com/openstreetmap/openstreetmap-website/actions/runs/15982286778/job/45079192697#step:13

gravitystorm:
I agree with @hlfan on this. It sort of vaguely makes sense, in principle, to cancel the other jobs when one fails, except for two key things that happen in reality:

* Our jobs only fail at the end. We run the entire test suite to completion, rather than failing early if a single test fails (i.e. `bundle exec rails test` not `bundle exec rails test --fail-fast`. So, as @hlfan discusses, it's likely that if one job fails, the other jobs are very nearly close to finishing anyway.
* Our jobs usually fail for "transient bug in test suite" reasons, much more than "this PR has a coding mistake" reasons. I use the "re-run failed tests" options all the time, both on PRs and on our master branch, when I see that there was a CI failure. So when one spurious job failure cancels the others, we're re-running jobs that were definitely nearly finished and probably were going to pass. (One caveat is that I don't know how often the coveralls failures would affect all three jobs completing close together, maybe the failures will be correlated on that).

So thanks @hlfan , merged.

PR #6154, opened 2025-06-30
Add reasonable compile time limit

hlfan:
Another way of freeing up recources: a timeout.
For now, only where I've seen multiple workflows hang:
- https://github.com/openstreetmap/openstreetmap-website/actions/runs/15960712372
- https://github.com/openstreetmap/openstreetmap-website/actions/runs/15918518252
- https://github.com/openstreetmap/openstreetmap-website/actions/runs/15472149750


tomhughes:
Those are all on a branch aren't they, so it's just a bug in the PR that should be fixed.

Not that I'm sure why a missing file error doesn't just cause an immediate abort? But finding out why and fixing that seems like a better solution than adding an arbitrary timeout.

Firefishy:
I think adding some timeout safeguards to our GitHub Actions would be a good idea. We don't have unlimited runtime and the default timeout is an extremely long 6 hours.

mmd-osm:
I’ve frequently seen hanging test runs in cgimap, which turned out to be related to a used up quota. The job was being silently killed, and eventually hit the timeout, although it’s been dead for some minutes already. You should be able to see more detailed stats as GitHub org manager or owner.

Firefishy:
Screenshot 2025-07-01 at 20 46 09

kcne:
I also agree that adding reasonable timeouts is a good idea. It would help free up resources and improve reliability -especially in cases where jobs silently hang or get stuck without failing fast.

tomhughes:
Would it make more sense to set the timeout at the job level rather than just on one step?

For the `docker` and `tests` workflow it would probably need to be 15 minutes or so on the main job while `lint` probably just needs a minute or two on each job.

hlfan:
Maybe both?

.github/workflows/danger.yml
```diff
@@ -11,6 +11,7 @@ permissions:
 jobs:
   danger:
     runs-on: ubuntu-latest
+    timeout-minutes: 1
```
hlfan: Well that has to be revised: https://github.com/openstreetmap/openstreetmap-website/actions/runs/16033899766/job/45240902000?pr=6154
That's why I initially only wanted to introduce one timeout.

tomhughes:
Looks good to me, thanks.

PR #6155, opened 2025-07-01
Reorganize changeset history info into two lines, reuse it in element details

### LINKED ISSUE #3621: Mark discussed changesets in the history
### Description

In the history of an object, mark the changeset if it has a discussion, e.g. like this:
discussed

### Screenshots

_No response_

### LINKED ISSUE #5757: Omit changeset "<comment count>💬" from changeset list where there are no comments
### Problem

A common use case of mine is to scan through a list of changesets to see what has comments. However perhaps due to recent changes? I find it hard to spot changesets with comments vs those without.

Perhaps even making the 💬 sand out more with more colour would be good too.

### Description

I suggest omitting the "0 💬" if no comments exist and only showing it where comments do exist.

### Screenshots

Image

AntonKhorev:
Currently if you open `/history` you'll see something like this:
image

In most of the cases each changeset entry has this visual structure:
~~~
(comment tag)

Closed (time) by (username) ·   (n comments)
#(id)                           (n changes)
~~~

Sometimes in case of a longer username the structure is:
~~~
(comment tag)

Closed (time) by                (n comments)
(username) · #(id)              (n changes)
~~~

The username is moved to the next line, id is moved to the right.

It would be easier to examine the history if things were moving around less often. That's what this PR tries to do, and hopefully it solves a couple of other problems.

For usernames to be less often carried over to next lines, we can make their lines longer. We can do it by not having comments/changes counts on the side, moving them below that line instead and arranging them horizontally, on the same line with ids:
~~~
(comment tag)

Closed (time) by (username)
#(id)              (n changes) (n comments)
~~~

image

Long usernames will get carried over, but they won't shift ids to the right:
image

Another advantage of this arrangement is that you get two columns for changes and comments counters. If you're interested in discussed changesets, you can visually scan the column with comments counters. Nonzero counts should be even easier to spot because there aren't any edit counts right below them. This should help with https://github.com/openstreetmap/openstreetmap-website/issues/5757#issuecomment-3017341686.

Additionally we can take the id-counters line and reuse it in element details, as requested in #3621. This should be useful on element history pages when looking for controversial changes to a single element. Edit counts are irrelevant for this goal and are not shown.

image

---

Fixes #3621.

Fixes #5757 further.

### DIFF
--- a/app/views/browse/_common_details.html.erb
+++ b/app/views/browse/_common_details.html.erb
@@ -21,10 +21,11 @@
           :time_ago => friendly_date_ago(common_details.timestamp),
           :user => changeset_user_link(common_details.changeset) %>
   </div>
-  <div>
+  <%= render "changesets/changeset_line", :num_changes => nil,
+                                          :num_comments => common_details.changeset.comments.length do %>
     <%= t "browse.in_changeset" %>
     #<%= link_to common_details.changeset_id, common_details.changeset %>
-  </div>
+  <% end %>
 
   <% if @type == "node" and common_details.visible? %>
     <div>


hlfan (approved):

tomhughes:
Looks good to me, thanks.

PR #6161, opened 2025-07-01
Add test for set_locale fix in notes controller

AntonKhorev:
Tests the fix in c266f3c90a154dcb43ab0ff2b2e4337c510707d8.


tomhughes:
Looks good to me although really there are dozens of other pages you could add the same test to - this just happened to be the one where it was broken.

PR #6162, opened 2025-07-02
Refactor share pane buttons to use Bootstrap tabs

hlfan:
Trade extra JS for extra HTML since that's now a thing


tomhughes:
Looks good to me, thanks.

PR #6164, opened 2025-07-02
Enable arrow-spacing ESLint rule

hlfan:
The arrow syntax should be equally as checked as the regular function syntax.

### DIFF
--- a/config/eslint.config.mjs
+++ b/config/eslint.config.mjs
@@ -46,6 +46,7 @@ export default [
     rules: {
       "@stylistic/array-bracket-newline": ["error", "consistent"],
       "@stylistic/array-bracket-spacing": "error",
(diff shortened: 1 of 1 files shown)


HolgerJeromin (approved):
I do not like stylistic eslint rules (but prettier is harder to add here) but it is consistent.

tomhughes:
Looks good to me, thanks.

hlfan:
Unfortunate merge timing Mr. Hughes:
https://github.com/openstreetmap/openstreetmap-website/blob/ef2f248e0075a4cd7d77eefc77665f1730d6bd4b/app/assets/javascripts/leaflet.share.js#L15

HolgerJeromin:
> Unfortunate merge timing Mr. Hughes:

For the record:
Fixed with af0c68e

PR #6166, opened 2025-07-03
Fix message reuse for Terms of Use link in signup form

Pikse:
Currently `layouts.tou` message (link label) is reused in two different sentences, i.e. in messages `accounts.terms.show.tou explain html` and `users.new.by_signing_up.html`. In some languages however the link label needs different declension in these sentences.

Here `users.new.by_signing_up.tou` message is added for use in the latter sentence in signup form. In English it's identical to previous `layouts.tou` message.

Similarly in #4936 two other link labels in the same sentence were already adjusted like this.

It also appears that at some point the terms of use link label already existed for this particular sentence, as still documented in translatewiki, but no one remembered to actually use it, and then it got removed as an unused message in b0ad96e.

### DIFF
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -68,7 +68,7 @@
   <% end %>
 
   <p class="mb-3 text-body-secondary fs-6"><%= t(".by_signing_up.html",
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #6168, opened 2025-07-04
Move changeset bbox test to integration tests

AntonKhorev:
As requested in FIXME:
```
# FIXME: This should really be moded to a integration test due to the with_controller
```


tomhughes:
Looks good to me, thanks.

PR #6170, opened 2025-07-05
Enhanced changeset statistics

AntonKhorev:
Rails part of https://github.com/zerebubuth/openstreetmap-cgimap/pull/510


mmd-osm:
>  cgimap update is supposedly ready now

Right, PR https://github.com/zerebubuth/openstreetmap-cgimap/pull/510 is ready to be merged. 

My proposal to move ahead is as follows:

- ~Merge cgimap PR 510. This automatically triggers a deployment to the dev instance~ -> done
- ~Merge Rails part of the solution, which includes database migration (dev + production)~ -> done
- ~Enable new config parameter on dev instance via Chef~ -> done
- ~Allow for some tests on dev for a week or two, maybe?~ -> done
- ~Meanwhile prepare new cgimap release~ -> done
- ~Deploy new cgimap release to production~ -> done
- ~Enable new config parameter on production~ -> done
- Possibly plan for some backfilling (this could already be tested on the dev instance)

I've added a new configuration parameter to enable enhanced changeset stats: (default is "false" at the moment).

```
  --changeset-enhanced-stats arg  enable enhanced changeset stats
```

This should translate to `CGIMAP_CHANGESET_ENHANCED_STATS=true` as environment variable.

Please feel free to comment if I missed anything.

tomhughes:
This looks good to me, thanks, and it sounds like we're good to merge and deploy it per @mmd-osm's plan.

deevroman:
> Actually the stats are displayed only if the new counts match the old num_changes;

And in what cases do these counts not match? For example, why are there no statistics for 169196100? https://openstreetmap.org/user/TrickyFoxy/history?after=169150701

mmd-osm:
You can check https://master.apis.dev.openstreetmap.org/user/mmd2mod/history for how it will look like once cgimap 2.1.0 is deployed in production.

PR #6171, opened 2025-07-07
Move integration tests out of api relation controller tests

AntonKhorev:
I wanted to do create/modify/delete stats per changeset in #6170 because it's already done in cgimap. That requires updating api tests, and relation tests still contain tests that are actually integration tests.

This PR moves relation - old relation - changeset integration tests out of api relation controller tests.


tomhughes:
Looks good to me, thanks.

PR #6184, opened 2025-07-12
Lock describe_location button to query placeholder

hlfan:
Use the input state to toggle the describe_location button's visibility with CSS instead of manually updating.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -556,6 +556,10 @@ header .search_forms,
 .search_form {
   .describe_location {
     font-size: 10px;
+
+    input:not(:placeholder-shown) + .input-group-text & {
+      display: none;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/common.scss
```diff
@@ -556,6 +556,10 @@ header .search_forms,
 .search_form {
   .describe_location {
     font-size: 10px;
+
+    input:not(:placeholder-shown) + .input-group-text & {
```
tomhughes: This rule doesn't actually work because it compiles to:

```css
input:not(:placeholder-shown) + .input-group-text .search_form .describe_location
```

when what you want is:

```css
.search_form input:not(:placeholder-shown) + .input-group-text .describe_location
```

tomhughes (changes requested):
In addition to my inline comment there's a separate issue on Firefox which I've opened #6185 for.

hlfan:
Dependent on #6185

PR #6187, opened 2025-07-12
Don't do partial renders for turbo page visits

### LINKED ISSUE #6172: Paging interferes with logout flow
### URL

_No response_

* We're starting on a page that supports paging (so either diaries or traces) and navigate to an older entry. The resulting URL in the browser window is then something like:
- https://www.openstreetmap.org/diary?before=407009
- https://www.openstreetmap.org/traces?before=12015435
* Press logout
* Notice an HTTP 422 Unprocessable Entity error code, browser points to https://www.openstreetmap.org/logout?referer=%2Ftraces

_No response_

tomhughes:
Setting `turbo-action=advance` as our pagination helper does causes a frame navigation to be [promoted to a page visit](https://turbo.hotwired.dev/handbook/frames#promoting-a-frame-navigation-to-a-page-visit) which means turbo will try and [merge the head elements](https://turbo.hotwired.dev/handbook/drive#page-navigation-basics) and remove meta and link elements which are missing from the turbo response.

In order to avoid breaking things we need to ensure we send the full head element in such cases or things like the CSRF token will be lost.

Fixes #6172

### DIFF
--- a/app/controllers/accounts/deletions_controller.rb
+++ b/app/controllers/accounts/deletions_controller.rb
@@ -1,6 +1,6 @@
 module Accounts
   class DeletionsController < ApplicationController
-    layout "site"
+    layout :site_layout
 
     before_action :authorize_web
     before_action :set_locale
--- a/app/controllers/accounts/pd_declarations_controller.rb
+++ b/app/controllers/accounts/pd_declarations_controller.rb
@@ -1,6 +1,6 @@
 module Accounts
   class PdDeclarationsController < ApplicationController
-    layout "site"
+    layout :site_layout
 
     before_action :authorize_web
     before_action :set_locale
--- a/app/controllers/accounts/terms_controller.rb
+++ b/app/controllers/accounts/terms_controller.rb
@@ -2,7 +2,7 @@ module Accounts
   class TermsController < ApplicationController
     include SessionMethods
 
-    layout "site"
+    layout :site_layout
 
     before_action -> { authorize_web(:skip_terms => true) }
     before_action :set_locale
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -2,7 +2,7 @@ class AccountsController < ApplicationController
   include SessionMethods
   include UserMethods
 
-  layout "site"
+  layout :site_layout
 
   before_action :authorize_web
   before_action :set_locale
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -39,6 +39,10 @@ class ApplicationController < ActionController::Base
 
   private
 
+  def site_layout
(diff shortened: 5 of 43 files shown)


AntonKhorev:
But you don't have to resend the entire page. #5175 fixed this by sending `<head>` + partial without the rest of the page. However that wasn't the only issue, another one was #6112 and at the time I didn't know how/if it was related, that's why I kept that PR as a draft. Now I know it was because of `Vary` header.

tomhughes:
No you're right that you don't have to send the whole page - you can do `render "_page"` to render the template but not as a partial but you still get the whole of the layout so you don't really save much.

Yes I guess your solution from #5175 works as well though it seems quite hacky and I don't know if it's really guaranteed to work given what you send isn't really valid HTML but then neither is just sending the partial.

I think what we're trying to do is really push things beyond what's intended though? Most examples seem to assume you're just sending the full page as normal and leaving the client to figure out what bits are needed...

AntonKhorev:
This is the intended layout:
https://github.com/hotwired/turbo-rails/blob/main/app/views/layouts/turbo_rails/frame.html.erb

Ours need to be updated according to:
https://github.com/hotwired/turbo-rails?tab=readme-ov-file#a-note-on-custom-layouts

tomhughes:
I've added that in now although I don't think it's a perfect solution - it restore the CSRF meta tags which is the immediate problem here but there's lots of other stuff that gets lost from the `head` element.

In principle the `yield :head` lets you add stuff in but that requires you to have been able to call `content_for :head` without having to add it to every individual page but that would have to be done from the layout which we're not in control of.

If it proves to be a problem then we could replace their special layout with one of our own that includes a fuller `head` element without all the rest of the layout.

You do of course still get the rest of the page outside the layout in any case.

AntonKhorev:
I updated my #5175. It only changes controllers that use turbo pagination.

Their `frame.html.erb` isn't entirely working for us because we have a lot of stuff inside `<head>` that's not added using `content_for :head`. I'm including the entire `layouts/head` which should be almost everything except for maybe rss links in some cases.

tomhughes:
Well yes I considered doing that as I said and it would be easy enough if people prefer to go that way.

I don't see any point in limiting it to specific requests though - surely the mechanism here that does it automatically for all frame requests is better and more future proof?

tomhughes:
I've updated this to send the full head by using our own layout...

AntonKhorev:
Not using Turbo is even more future proof, if we're not interested in making responses faster.

AntonKhorev:
The difference between payloads here and in #5175 is not that large after all.

If we need to save on the output or on rendering of what's not going to make it to the output, we can add `if turbo_frame_request_id` to `content_for :heading` for example.

PR #6189, opened 2025-07-12
Store routing icons separately

hlfan:
Test driving the inline_svg gem.

Continues #5753 and closes #5776 following the discussion in #5880.

The color replacement is still broken, and I don't know what else to try.


mmd-osm:
I've tested your code (aa65a878f7f610554b8ad67f3731c49d7e07f8d3), and for me color replacement is working ok. I also put in some debug "hello world" code in both transform methods, and they're being called when loading the page.

image

Dark mode:

image

hlfan:
Apparently, some change to the initializer config didn't get loaded in correctly. I simply didn't assume my code wasn't at fault, but I just needed to restart...

tomhughes:
Yes if you change initialisers you need to restart and they're only run once at startup.

config/initializers/inline_svg.rb
```diff
@@ -0,0 +1,7 @@
+require Rails.root.join("lib/svg_to_symbol_transform.rb")
+require Rails.root.join("lib/replace_magenta_svg_color.rb")
```
tomhughes: I'd probably just put the custom transform classes inline here rather than putting them in `lib` especially as autoloading doesn't even work here - at least I assume that's why you had to require them explicitly.

Other initialisers that do similar things just define the code they want in place. I do usually put it in `module OpenStreetMap` though, to make it clear it's a customisation if it appears in a stack trace.

config/initializers/inline_svg.rb
```diff
@@ -0,0 +1,23 @@
...
+        attr.value = "currentColor" if attr.value.strip.downcase == value
+      end
+      doc
+    end
+  end
+
+  class SvgToSymbolTransform < InlineSvg::CustomTransformation
+    def transform(doc)
+      with_svg(doc) do |svg|
+        svg.name = "symbol"
```
tomhughes, line 5: I don't think this transform is working - functionally everything does work but if you look at the source the outer `svg` contains a bunch of `svg` elements not `symbol` elements.

hlfan: Yes, I've noticed. Currently writing a test for the sprite rendering correctly in `directions_controller_test.rb` because I don't have a better idea of where to put it.

tomhughes: Ah yes I can see it's working in the latest version.

hlfan: Given there isn't a "lib" to test against anymore, a simple DOM check seems the most straightforward to me.
Or should testing that be done somewhere else?

config/initializers/inline_svg.rb
```diff
@@ -0,0 +1,23 @@
+module OpenStreetMap
+  class SvgCurrentColorTransform < InlineSvg::CustomTransformation
+    def transform(doc)
+      doc.xpath("//@fill|//@stroke").each do |attr|
+        attr.value = "currentColor" if attr.value.strip.downcase == value
+      end
+      doc
+    end
+  end
```
AntonKhorev: Are you sure you need this magenta transform instead of just using `currentColor` in svg files?

tomhughes: I assumed there was some reason we couldn't do that but if we can then great.

AntonKhorev: It's not going to be the actual current color when opened as an external file, but the color seems to be defaulting to black[^1] in everything I opened it with. And I'd rather look at black than at magenta.

[^1]: ["Depends on user agent"](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/color) via `color` attribute.

app/views/directions/show.html.erb
```diff
@@ -1,79 +1,9 @@
...
-  <symbol id="routing-sprite-ferry" fill="none" stroke="currentColor" stroke-width="1">
-    <path d="M10.5 8 l-6 2 l2.5 2 v1.5 a2.828 2.828 0 0 1 1.5 1 a2.828 2.828 0 0 1 4 0 a2.828 2.828 0 0 1 1.5 -1 v-1.5 l2.5 -2 z" fill="currentColor" />
-    <path d="M6.5 9.5 v-5 h8 v5 m-5.5 -6 h3" />
-    <path d="M5.5 16.5 a1.414 2.828 0 0 1 2 0 a1.414 2.828 0 0 0 2 0 a1.414 2.828 0 0 1 2 0 a1.414 2.828 0 0 0 2 0 a1.414 2.828 0 0 1 2 0" />
-  </symbol>
+<svg class="d-none">
+  <% Rails.root.glob("app/assets/images/routing/*.svg").each do |path| %>
+    <%= inline_svg_tag "routing/#{File.basename(path)}", :id => "routing-sprite-#{File.basename(path, '.svg')}", :to_symbol => true %>
+  <% end %>
 </svg>
```
AntonKhorev, line 14: Why the mirroring had to be removed?

config/initializers/inline_svg.rb
hlfan: Some icons in this repo (like [ohm.svg](https://github.com/openstreetmap/openstreetmap-website/blame/master/app/assets/images/social_link_icons/ohm.svg)) say they were exported from Illustrator.
Illustrator has currently [no support](https://illustrator.uservoice.com/forums/601447-illustrator-desktop-bugs/suggestions/42330448-please-allow-fill-currentcolor-to-be-set-for-svg) for currentColor.

But I'm up for removing the transform if this will be checked in the PRs.
I don't really like the magenta icons either, but it helps to catch icons set explicitly black. This would otherwise only be quickly obvious when reviewing in dark mode.

app/views/directions/show.html.erb
hlfan: It was easier to get working without, will add that back after I figured out which translation (20 or 21 px) should be applied. I'm not gonna mirror each icon separately.

AntonKhorev: Did you make any changes that prevent existing `<use>`s like the following from working?

```svg
  <symbol id="routing-sprite-slight-left">
    <use href="#routing-sprite-slight-right" transform="matrix(-1 0 0 1 20 0)" />
  </symbol>
```

Because if not, you don't need any transforms for mirroring.

hlfan: No, that's exactly my plan.
I've decided to move the right and sharp right turn icons 1px toward the center so that these start at the same point as the slight right turn icon.
That means I can use the `transform="matrix(-1 0 0 1 20 0)"` for every icon.

image

tomhughes:
Thanks for working on this... I think it looks good now.

PR #6191, opened 2025-07-13
Configure builtin apps script

AntonKhorev:
I had to reinstall openstreetmap-website and decided to automate the [OAuth app registration step](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONFIGURE.md#oauth-consumer-keys).


tomhughes:
This should probably be a rake task in `lib/tasks` I think... I keep meaning to convert the current random tools in `scripts` to be proper tasks.

mmd-osm:
I'm using something similar that creates a few sample users + also registers an app for JOSM: https://gist.github.com/mmd-osm/a3a0b5a8799a333e7fae578038aa7a93

I'm also generating a doorkeeper_signing_key in that script and add it to settings.local.yml, which is a prerequisite for OIDC scenarios, and also JOSM.

Feel free to pick whatever makes sense.

AntonKhorev:
Updated as a rake task.

CONFIGURE.md
```diff
@@ -57,6 +57,19 @@ There are two built-in applications which communicate via the API, and therefore
 * iD
 * The website itself (for the Notes functionality)
 
+You can register them by running a script and copying its output to `config/settings.local.yml`:
+
+```
+./script/register-oauth-apps "My New User Name"
```
tomhughes: This should be `bundle exec rails db:register_oauth_apps` now.

AntonKhorev: Updated.

Also renamed the task to `oauth:register_apps` because I'll probably add another OAuth task for doorkeeper_signing_key.

Also I'm appending the settings to `config/settings.local.yml` instead of telling the user to copy-paste because I've already checked that these settings didn't exist.

PR #6192, opened 2025-07-13
Image download w/ submit_tag

mmd-osm:
Once a user clicks on "Download" Image, there's currently no visual feedback that anything happens. Since the image generation may take a few seconds, a user may be tempted to hit the download button multiple times.

This PR disables the button as soon as a hits the download button, similar to the Export Map function. This way, we're improving UI consistency across different export functions.

I'm still looking for ideas on how / when to re-enable the button...

image


tomhughes:
I think it definitely needs more work on the re-enablement but I'm not sure there's a good way to do it without changing the way the download works.

I'm not actually sure that you even need that code as I think rails does it already but it does so as soon as the initial form post returns so doesn't include the time take to do the real work from the redirect.

mmd-osm:
I think what you describe is documented here: https://turbo.hotwired.dev/handbook/drive#form-submissions

_During a submission, Turbo Drive will set the “submitter” element’s [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) attribute when the submission begins, then remove the attribute after the submission ends._

Somehow this last bit never worked when testing the new code.  If I don't manually remove the "disabled" attribute, it will remain disabled until I reload the page. I'm not exactly sure why is this. Maybe the POST + 302 Redirect + GET is treated differently? Otherwise, I'd agree that a properly working "turbo:submit-end" would be a bit pointless from a UX point of view.

I've also tried a `fetch` based approach where I have more control over the download. This turned out to be a bit trickier than expected, which is why I started with the current approach.

tomhughes:
Like I say it won't help anyway with the current code but I turned on event logging to see if there was an event we could use and I say an event being fired from rails form submission code and it looked like it had code to re-enable the button but maybe it was conditional.

mmd-osm:
The click.rails event seems to be handled by jquery-ujs (https://github.com/rails/jquery-ujs/blob/master/src/rails.js#L524-L551). There's also some code in place to re-enable form elements, but my debugger isn't stopping there. It seems the turbo drive link I've posted earlier on isn't relevant?

```
    $document.on('ajax:complete.rails', rails.formSubmitSelector, function(event) {
      if (this === event.target) rails.enableFormElements($(this));
    });
```

mmd-osm:
So the button is set to disabled in the setTimeout function below.

```javascript
    $document.on('submit.rails', rails.formSubmitSelector, function(e) {
      var form = $(this),
        remote = rails.isRemote(form),
        blankRequiredInputs,
        nonBlankFileInputs;

[...]

      } else {
        // Slight timeout so that the submit button gets properly serialized
        setTimeout(function(){ rails.disableFormElements(form); }, 13);                // <<<<<
      }
    });
```
    
Re-enabling the button would work, if we were to use ajax:
    
```javascript
    $document.on('ajax:complete.rails', rails.formSubmitSelector, function(event) {
      if (this === event.target) rails.enableFormElements($(this));
    });    
```

However, in this scenario, it's a plain HTTP POST, followed by a 302 response and another HTTP GET on render.openstreetmap.org.

This event simply never fires, because it's not a "remote" call. The source code mentions a "data-remote" attribute to send the request as AJAX.

However, CSP doesn't allow render.openstreetmap.org, the render site isn't properly handling CORS preflight requests, and the required totp cookie would be lost as well.

Long story short, I don't see much alternatives at the moment to implement this.

AntonKhorev:
> CSP doesn't allow render.openstreetmap.org

Easily fixed by allowing them. They are already allowed for `form_action`, just add them to `connect_src` too.

> the required totp cookie would be lost

Would it? Is it even required by the exporter?

> the render site isn't properly handling CORS preflight requests

This sort of works too, `access-control-allow-origin: *` is in the response. However the status is 400 instead of 204, that means *CORS request did not succeed*.

AntonKhorev:
> Somehow this last bit never worked when testing the new code. If I don't manually remove the "disabled" attribute, it will remain disabled until I reload the page.

Add `data-turbo="true"` to the form:
```html
<form id="export-image" action="/export/finish" method="post" data-turbo="true">
```

And this on the button, with proper i18n:
```html
<input type="submit" class="btn btn-primary" value="<%= t ".download" %>" data-turbo-submits-with="Downloading..." />
```

CORS still wouldn't work though, until render.openstreetmap.org supports preflight requests.

mmd-osm:
> Would it? Is it even required by the exporter?

Yes, it's explicitly checked here: https://github.com/openstreetmap/chef/blob/master/cookbooks/tile/templates/default/export.erb#L179-L183 + https://github.com/openstreetmap/chef/blob/master/cookbooks/tile/templates/default/export.erb#L190-L192

tomhughes:
> CORS still wouldn't work though, until render.openstreetmap.org supports preflight requests.

That shouldn't be hard to fix if we need to - is it preflighting because this will switch to POST or because of the cookie or something?

AntonKhorev:
During preflight the export script shouldn't look at cookies because it won't receive any cookies. So it should quit with 204 [here](https://github.com/openstreetmap/chef/blob/849abef3edeac8e563ad793f55dd4bdea4b9b232/cookbooks/tile/templates/default/export.erb#L161) or earlier.

AntonKhorev:
> > Would it? Is it even required by the exporter?
> 
> Yes, it's explicitly checked here [...]

I didn't notice that the cookie is being sent even when I'm using my own osm-website instance, and the cookie is going to be valid if I exported on osm.org recently.

mmd-osm:
>  the cookie is going to be valid if I exported on osm.org recently

I think the TOTP cookie is valid for an hour or so. I'm also testing locally by running an export on osm.org to get a valid token, then continue with local tests.

Maybe a cookie isn't the best choice for this scenario.

mmd-osm:
> During preflight the export script shouldn't look at cookies because it won't receive any cookies. 

Agreed, the script isn't handling preflight requests properly at the moment. Maybe Apache could also return something for OPTIONS requests, and only trigger the script for GET.

mmd-osm:
>  is it preflighting because this will switch to POST or because of the cookie or something

In the AJAX case, it's the second (GET) request on render.osm.org that triggers the preflight request.

tomhughes:
Hopefully https://github.com/openstreetmap/chef/commit/61da17940d24cfa899ff3086b5f53ffe9feb18ab will do the necessary to allow the preflight check.

mmd-osm:
https://render.openstreetmap.org/cgi-bin/export seems to return multiple / deviating Access-Control-Allow-Origin headers. Maybe Apache adds another header on top of what the script already does?

image


image

tomhughes:
Yes I'm just pushing a fix for that...

mmd-osm:
I'm using the code below, which adds `Access-Control-Request-Headers: x-csrf-token,x-turbo-request-id` in the preflight request.

This fails because we're not allowing these headers:

_Reason: header ‘x-csrf-token’ is not allowed according to header ‘Access-Control-Allow-Headers’ from CORS preflight response_




```erb
diff --git a/app/views/share_panes/show.html.erb b/app/views/share_panes/show.html.erb
index 23f46dc4d..f4ae5ac95 100644
--- a/app/views/share_panes/show.html.erb
+++ b/app/views/share_panes/show.html.erb
@@ -40,7 +40,7 @@
       <% end %>
     </ul>
   </div>
-  <form id="export-image" action="/export/finish" method="post">
+  <form id="export-image" action="/export/finish" method="post" data-turbo="true">
     <div class="row mb-3">
       <label for="mapnik_format" class="col-auto col-form-label"><%= t ".format" %></label>
       <div class="col-auto">

diff --git a/app/controllers/application_controller.rb b/app/controllers/application_controller.rb
index 03ea6bedb..33ca6ccb6 100644
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -246,7 +246,7 @@ class ApplicationController < ActionController::Base
   def map_layout
     policy = request.content_security_policy.clone
 
-    policy.connect_src(*policy.connect_src, "http://127.0.0.1:8111", Settings.nominatim_url, Settings.overpass_url, Settings.fossgis_osrm_url, Settings.graphhopper_url, Settings.fossgis_valhalla_url)
+    policy.connect_src(*policy.connect_src, "http://127.0.0.1:8111", "render.openstreetmap.org", Settings.nominatim_url, Settings.overpass_url, Settings.fossgis_osrm_url, Settings.graphhopper_url, Settings.fossgis_valhalla_url)
     policy.form_action(*policy.form_action, "render.openstreetmap.org")
     policy.style_src(*policy.style_src, :unsafe_inline)
```

tomhughes:
Those should be allowed now...

mmd-osm:
Now I'm getting _Error - Missing or invalid token_, which makes sense because we're not sending any cookies across.

I'm not even sure this works. fetch is using credentials option "same-origin" by default, and even "include: always include credentials, even cross-origin", would not work due to the SameSite: Lax setting for the _osm_totp_token

mmd-osm:
The other issue I've noticed is that Thunderforest maps (and all others) need to be added to policy.connect_src as well.

Although the download is triggered, the resulting PNG file disappears into nowhere.

tomhughes:
Maybe we should allow the token to be sent as a parameter instead of a cookie? Would it be possible to send it that way?

The thunderforest layers are the only other ones which allow export so they're the only ones that will need changing - are there any CORS issues there?

mmd-osm:
I think we could skip the cookie altogether:

From app/controllers/application_controller.rb -> update_totp -> generates new TOTP cookie today.

`ROTP::TOTP.new(Settings.totp_key, :interval => 3600).now` needs to be added to app/controllers/export_controller.rb and passed as URL parameter to the render.osm.org URL.

Render script needs to be enhanced to check new URL parameter next to the cookie.

Thunderforest layers already work without showing signs of CORS issues, although the request gets lots somewhere in turbo:before-fetch-response / turbo:submit-end events. At one point it tries to parse the PNG image as HTML, which fails, then silently discards the image. I guess some additional event handling would be needed here.

tomhughes:
I've pushed a change to use the `token` parameter from the URL as the token if there is no cookie.

mmd-osm:
The preflight request is failing with HTTP 500. There might be something wrong with `form.getfirst("token")`. Elsewhere we're using `getvalue` to fetch the parameter value, and check if a value is available:

```
elif "token" in form:
  token = form.getvalue("token")
```

tomhughes:
The python cgi library is weird and the form object it returns for the options request doesn't work at all so we now ignore it.

We need to find a replacement anyway as the latest python release has dropped it :-(

mmd-osm:
So the image download seems to work now. I still want to hardcode PNG format for thunderforst (that's the only format available for download), and check how the code handles bad request due to timeouts during image generation.

I'm seeing occiasional CSRF issues that were also showing up in other Turbo scenarios. I will not look into them for now.

mmd-osm:
Instead of hardcoding formats, I have added a new attribute "formats" to layers.yml, and disable non-supported formats in the select options:

image

Then there's a bit of logic to select a non-disabled option, after options have been updated. This is relevant when switching from Mapnik to Thunderforest, while having a non-PNG format still selected.

app/assets/javascripts/leaflet.share.js
```diff
@@ -7,8 +7,33 @@ L.OSM.share = function (options) {
         });
 
   function init(map, $ui) {
-    // Link / Embed
+    $("#export-image").on("turbo:submit-end", function (event) {
```
hlfan: Could this code block be moved into the `// Image` section just below?
I'd suggest just below the csrf input.

app/assets/javascripts/leaflet.share.js
```diff
@@ -7,8 +7,33 @@ L.OSM.share = function (options) {
...
+          const a = document.createElement("a");
+          a.href = url;
+          const contentType = blob.type;
+          let extension = "bin";
+          if (contentType === "image/png") extension = "png";
+          else if (contentType === "image/jpeg") extension = "jpg";
+          else if (contentType === "image/svg+xml") extension = "svg";
+          else if (contentType === "image/webp") extension = "webp";
+          else if (contentType === "application/pdf") extension = "pdf";
+          a.download = `map.${extension}`;
```
hlfan: For the directions route download, I made the filename translatable.
This would be nice to have here, too.

hlfan: `blob.type` is normally used for automatically determining the file extension. Why is this done manually here?

mmd-osm: Maybe I'm missing something here. blob.type contains a mime type at this point which I'm mapping to a file suffix. Also, the url doesn't have any suffix. How would an automatic determination work here?

Image

app/assets/javascripts/leaflet.share.js
mmd-osm: good point

app/assets/javascripts/leaflet.share.js
hlfan: Without a value in the download attribute, the browser will decide filename and extension from the blob type for object URLs ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#download)), but with `download="map"` set the extension should still be determined from the `blob.type`.

But I'm not sure if this gets correctly forwarded from `response.blob()`.

app/assets/javascripts/leaflet.share.js
```diff
@@ -8,7 +8,6 @@ L.OSM.share = function (options) {
 
   function init(map, $ui) {
     // Link / Embed
-
```
hlfan: What has this empty line done to you? /j

mmd-osm: oops, fixed.

app/assets/javascripts/leaflet.share.js
mmd-osm: Right, a filename without suffix works... I have removed all the manual stuff now. Thanks!

mmd-osm: done

mmd-osm:
Currently the export script returns errors as text/html, which Turbo happily picks up to replace the osm.org page. In this case, the user has to reload the page. Browser back button would take them to a random place, which is quite annoying.

Maybe it's better for the export script to return error messages as plain text. Turbo ignores this response, we can display the alert popup and nothing else happens.

As an alternative, https://discuss.hotwired.dev/t/finally-understanding-turbo-stream/1964/8 talks about some funny stuff we could do with Turbo Streams, though it seems a bit over-engineered for this use case.

leaflet.share.js:

```js
      } else {
        let detailMessage;
        try {
          detailMessage = event?.detail?.error?.message;
          if (!detailMessage) {
            detailMessage = await event.detail.fetchResponse.responseText;
          }
        } catch {
          detailMessage = "(unknown)";
        }
        // eslint-disable-next-line no-alert
        alert(OSM.i18n.t("javascripts.share.export_failed", { reason: detailMessage }));
      }
```
export.erb:

```python
# Routine to report an error
def output_error(message, status = "400 Bad Request"):
  print("Status: %s" % status)
  output_headers("text/plain")
  print(message)
```

Result:

image

tomhughes:
Sounds reasonable to me... I've changed the export script to use plain text now.

mmd-osm:
Mmh, Thunderforst is still sending us HTML error messages. Can we change this somehow?

tomhughes:
Well only by asking Andy nicely!

mmd-osm:
I'm not sure I have enough karma points to propose such a change.

Anyway, one alternative I found is to handle the Turbo before-fetch-response event. In case the endpoint is sending us failures in html format, we tell Turbo to no longer follow up on this particular event. A bit later, the event handler then no longer tries to replace the current page by the HTML response, which is what we're trying to avoid here.

```javascript
    $("#export-image").on("turbo:before-fetch-response", function (event) {
      const res = event.detail.fetchResponse.response;
      if (!res.ok && res.headers.get("content-type")?.includes("text/html")) {
        event.preventDefault();
        event.stopPropagation();
      }
    });
```

It seems the turbo:submit-end is nevertheless triggered as an event, and as a result we're getting this nice error popup, and nothing else: 

image

Does this seem reasonable?

Relevant code on Turbo side:

```javascript
  async receive(response) {
    const fetchResponse = new FetchResponse(response);
    const event = dispatch("turbo:before-fetch-response", {
      cancelable: true,
      detail: {
        fetchResponse: fetchResponse
      },
      target: this.target
    });
    if (event.defaultPrevented) {
      this.delegate.requestPreventedHandlingResponse(this, fetchResponse);
    } else if (fetchResponse.succeeded) {
      this.delegate.requestSucceededWithResponse(this, fetchResponse);
    } else {
      this.delegate.requestFailedWithResponse(this, fetchResponse);
    }
    return fetchResponse;
  }
```

tomhughes:
Seems reasonable to me... Should I maybe revert the change to our export to match?

mmd-osm:
I gave it a quick try and reverted to the previous version. I'd say, let's keep the text version, to avoid showing HTML tags in the popup:

image

Maybe Andy should really change the Content-Type header... it doesn't seem to be HTML, after all.

mmd-osm:
Last update for today. I did a bit of refactoring and split the larger functions into smaller parts.

hlfan:
How about trying to parse the HTML?
```js
detailMessage = $(detailMessage).text()
```

mmd-osm:
Well, the response really has to be HTML. In case of Thunderforest where we're getting some plain text, jquery will fail with an exception "Error: Syntax error, unrecognized expression: Requested image is too large."

For the former version of the export.erb script, it would work though:

```<html>\n<head>\n<title>Error</title>\n</head>\n<body>\n<h1>Error</h1>\n<p>The server is too busy at the moment. Please wait a few minutes before trying again.</p>\n</body>\n</html>\n```

image

What do you think about the following?

```javascript
        if (!detailMessage) {
          detailMessage = await event.detail.fetchResponse.responseText;
          const div = document.createElement("div");
          div.innerHTML = detailMessage;
          detailMessage = div.textContent || div.innerText || "(unknown)";
        }
```

hlfan:
That includes the title in the HTML head, too; you'd probably only want to output the body's text.

mmd-osm:
I think I need a DOMParser for this task. Something like:

```javascript
let parser = new DOMParser();
let doc = parser.parseFromString(detailMessage, 'text/html');
detailMessage  = doc.body ? doc.body.textContent.trim() : "(unknown)";
```

This manages to parse both text and html.

app/assets/javascripts/leaflet.share.js
```diff
@@ -28,6 +28,65 @@ L.OSM.share = function (options) {
...
+      alert(OSM.i18n.t("javascripts.share.export_failed", { reason: detailMessage }));
+    }
+
+    $("#export-image").on("turbo:submit-end", async function (event) {
+      if (event.detail.success) {
+        await handleExportSuccess(event.detail.fetchResponse);
+      } else {
+        await handleExportError(event);
+      }
+    });
```
hlfan: I found having the logic in a Promise pipeline rather than split over three async functions easier to follow:
```js
    function extractTextFromHTML(htmlString) {
      const parser = new DOMParser();
      const doc = parser.parseFromString(htmlString, "text/html");
      return doc.body ? doc.body.textContent.trim() : "(unknown)";
    }

    function notifyExportFailure(reason) {
      // eslint-disable-next-line no-alert
      alert(OSM.i18n.t("javascripts.share.export_failed", { reason }));
    }

    $("#export-image").on("turbo:submit-end", function ({ detail }) {
      if (detail.success) {
        detail.fetchResponse.response.blob()
          .then(downloadBlob)
          .catch(() => { notifyExportFailure("(blob error)"); });
      } else {
        Promise.resolve()
          .then(() => detail?.error?.message || detail.fetchResponse.responseText.then(extractTextFromHTML))
          .then(notifyExportFailure)
          .catch(() => { notifyExportFailure("(unknown)"); });
      }
    });
```
This requires `OSM.i18n.t("javascripts.share.filename")` to be inside `downloadBlob`.

mmd-osm: I have to say that I find async/await much more readable than Promise pipelines. Also, your example is mixing up error path and success path in one event handler, which I  tried to avoid by splitting it up in different functions.

So while your code is a bit shorter, I'd prefer to stick to the current version in this case.

config/layers.yml
```diff
@@ -4,6 +4,7 @@
   nameId: "standard"
   canEmbed: true
   canDownloadImage: true
+  formats: [png, jpeg, webp, svg, pdf]
```
hlfan: You add this for JS, but the select options are still hard-coded:
```erb
        <%= all_formats = @downloadable_layers.flat_map { |layer| layer["formats"] }.uniq
            select_tag("mapnik_format",
                       options_for_select(all_formats.map { |f| [f.upcase, f] }),
                       { :class => "form-select w-auto", :required => true }) %>
```

mmd-osm: Originally, having the list of formats was a workaround b/c handling Thunderforest error responses didn't work properly. Now that this is fixed, I should remove updateSelectOptions and the list of supported formats. It's really a topic for another PR.

A user sees this message now:

Image


Current code can be picked up later from here: https://gist.github.com/mmd-osm/33d454c19432116195ba97bfcddb051e

mmd-osm:
I think all comments should be addressed now...

app/assets/javascripts/leaflet.share.js
hlfan: I don't think you'd need to `await` in here, then the function can also be synchronous.

mmd-osm: I tested this version, and it seems to work. However, both handle* functions would still return a Promise which is resolved later. In the meantime, Turbo submit cleanup code continues to run. That's not so bad here as it resets UI element statuses mainly, and it's the very last event in this whole process.

One thing I found a bit strange is that I see jquery code as Event dispatcher, which happens to also show up in jquery-simulate. That's function `dispatch` in the screenshot below, node_modules/jquery-simulate/external/jquery-1.7.2/jquery.js:3331 (and also in jquery-rails-4.3.3/vendor/assets/javascripts/jquery3.js)

image

I'm not so sure why I'm seeing this code outside of unit tests. It would be nicer to test Turbo without this intermediate jquery code.

jquery-simulate describes itself as "Simulate events to help unit test user interactions.". Maybe it's used in a different way in this project?

-> app/assets/javascripts/index/history.js -> clickChangeset ?

Caveat: I haven't fully analyzed this jquery bit, and I may be misrepresenting things here...

hlfan: jquery-simulate is used for forwarding changeset bbox clicks to the corresponding link, so modifier keys work as expected.
Not sure why it should interfere here.

If I understand correctly the awaits here are more for de-promise-ifying the already resolved promises than actual awaiting at the time of the submit-end event, so I'm not expecting a generally noticeable timing difference.

mmd-osm: I know this sounds a bit strange... with jquery-simulate in place, I have seen out of sequence calls to `this.delegate.formSubmissionFinished(this);` (part of Turbo) before our own `turbo:before-fetch-response` event handler has finished.

Now, with jquery-simulate uninstalled and the respective code in history.js removed, I don't see the out of sequence calls anymore. That's also what I would expect here.

I tested each case on Firefox and Chrome with consistent behavior.

mmd-osm: I'm switching to native event handler now, `document.getElementById("export-image").addEventListener`, because I don't want to investigate weird jquery issues any further at this point.

mmd-osm: I’m wondering if we could retire this plugin (the code is fairly old), and use something more modern like https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click

Maybe a topic for another pull request.

tomhughes: We already use the native `click` method in a number of places - the only place that uses simulate is the history code.

Note that `click` originally only worked on buttons and input fields and not links (which is what the history code uses) but that changed in Chome 19 / Firefox 5!

tomhughes:
Looks good to me - thanks for all the work to get it sorted.

PR #6196, opened 2025-07-14
Clear floats in diary entries

### LINKED ISSUE #6186: Rendering of diary discussions
Sometimes diary discussions are not rendered underneath the diary post, as shown on the screenshot below. 

Image

Sample page: https://master.apis.dev.openstreetmap.org/user/mmd2/diary/40

I think in this case, `align="text"` is causing the issue. Once I remove it, the diary along with discussion looks as expected: https://master.apis.dev.openstreetmap.org/user/mmd2/diary/41

AntonKhorev:
If a diary entry has floated contents, its controls and discussion may be displayed to the side of floated elements.

This PR clears floats in diary entry texts.

Example entry text to trigger the bug:
```
<img width="100" height="1000" align="left">
```

Before:
image

After:
image

Fixes #6186.


tomhughes:
Looks good to me, thanks.

PR #6197, opened 2025-07-15
use JOSM/tag2link to linkify tag values

### LINKED ISSUE #986: Display a link for mapillary keys

### LINKED ISSUE #2405: Link to openplaques

### LINKED ISSUE #3920: link ref:IE:smr value
I hope this is the right place to ask this: There are now more than 1,500 cases of `ref:IE:smr` (Irish Sites and Monuments Record no.). Would it be possible to link the number as described on the [wiki page](https://wiki.openstreetmap.org/wiki/Key:ref:IE:smr), i.e. to https://maps.archaeology.ie/historicenvironment/?SMRS=VALUE?

Thanks!
#heritage #ireland

### LINKED ISSUE #5200: Support Panoramax in feature links (browse tags helper)
### Problem

_No response_

### Description

In this example: https://www.openstreetmap.org/way/84864029, for `operator:wikidata`, if the value is `Q110319893`, a link goes to https://www.wikidata.org/entity/Q110319893?uselang=fr

I'd like to see the same thing with the key `panoramax`. I'd like to see a link that points to `https://api.panoramax.xyz/#focus=pic&pic=$1`. $1 is the value of the key panoramax. For the example above, it would link to `https://api.panoramax.xyz/#focus=pic&pic=6023ce64-8442-4787-8c05-68cb8690393f`.

https://wiki.openstreetmap.org/wiki/Item:Q22269

### Screenshots

_No response_

k-yle:
Closes #986, Closes #2405, Closes #3920, Closes #5200, Partially closes openstreetmap/iD#11120, Partially closes openstreetmap/iD#9510

This supersedes #2621. cc @francois2metz and @pyrog from the old PR

### Description

OSM keys like [`mapillary`](https://osm.wiki/Key:mapillary) are now rendered as a clickable link:

This uses the data aggregated by [JOSM/tag2link](https://github.com/JOSM/tag2link), as [suggested](https://github.com/openstreetmap/openstreetmap-website/pull/2621#issuecomment-2660925603) by Tom, so the logic in this repo is significantly simpler than #2621. It works with semicolons and full URLs in the tag value.

### How has this been tested?
Unit tests and locally with docker


config/initializers/tag2link.rb
```diff
@@ -0,0 +1,24 @@
+# A map of each OSM key to its formatter URL. For example:
+# { "ref:vatin" => "https://example.com/$1" }
+TAG2LINK = lambda {
+  # the JSON data is an array with duplicate entries, which is not efficient for lookups.
+  # So, convert it to a hash and only keep the item with the best rank.
+  array = JSON.parse(Rails.root.join("node_modules/tag2link/index.json").read)
+
+  ranks = %w[deprecated normal preferred].freeze
```
tomhughes: Do we want to include `deprecated` at all?

config/initializers/tag2link.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  output = {}
+
+  all_keys = array.map { |item| item["key"] }.uniq
+
+  all_keys.each do |key|
+    # for each key, find the item with the best rank
+    best_definition = array
+                      .select { |item| item["key"] == key }
+                      .max_by { |item| ranks.index(item["rank"]) }
```
tomhughes: Unfortunately this is not a stable algorithm as there are keys with more than one entry at the same rank - even more than one `preferred` in some cases like `Key:seamark:radio_station:mmsi` so we might get different results for different runs unless we can break the tie somehow.

k-yle: <details>
<summary>This problem seems to only affect 17 tags, most of which are quite obscure (click to expand)</summary>

```py
de:amtlicher_gemeindeschluessel = 7
fhrs:id = 2
hashtags = 42
icao = 8
ref:EU:bwid = 4
ref:FR:MemorialGenWeb = 2
ref:FR:museofile = 2
ref:FR:SIREN = 5
ref:ibnr = 11
ref:INEP = 2
ref:INSEE = 4
ref:ruian = 3
ref:vatin = 4
ref:WDPA = 2
*:mmsi = 2
uic_ref = 6
woeid = 2
```

</details>

Maybe we should skip these ones if there are multiple matches with equal probability? If someone complains, then perhaps they should fix the problem upstream

config/initializers/tag2link.rb
k-yle: good point, fixed

config/initializers/tag2link.rb
mmd-osm: What would be required to mark some of the links as "primary" or "leading one"? 

`de:amtlicher_gemeindeschluessel` as an example defines nationwide municipality keys. However, due to the federal nature of the services, not every endpoint mentioned in the JSON file serves all numbers.

- https://www.statistikportal.de/de/gemeindeverzeichnis/12067137 (ok)
- https://www.statistik-bw.de/Service/Gemeindeverzeichnis/Gem.jsp?G=12067137 (not working)
- https://www.statistik.bayern.de/mam/produkte/statistik_kommunal/2018/12067137.pdf (not working)

tomhughes: Well that's what I thought `preferred` was, as against `normal`, but in seems in some cases there is more than one that is `preferred`.

config/initializers/tag2link.rb
```diff
@@ -0,0 +1,25 @@
+# A map of each OSM key to its formatter URL. For example:
+# { "ref:vatin" => "https://example.com/$1" }
+TAG2LINK = lambda {
+  # the JSON data is an array with duplicate entries, which is not efficient for lookups.
```
tomhughes: Why do we need an immediately invoked lambda here? Just to avoid namespace pollution? You could actually write it all with only one variable other than the end result:

```ruby
TAG2LINK = JSON.parse(Rails.root.join("node_modules/tag2link/index.json").read)
               .reject { |item| item["rank"] == "deprecated" }
               .group_by { |item| item["key"] }
               .transform_keys { |key| key.sub(/^Key:/, "") }
               .transform_values { |items| items.max_by { |item| ranks.index(item["rank"]) }.fetch("url") }
```

which just needs the array of ranks...

config/initializers/tag2link.rb
k-yle: since we're only talking about 17 examples here, I'd suggest we do not show the link if there is >1 URL with the best rank. 

As the `de:amtlicher_gemeindeschluessel` example shows, in these cases there may not be a link that works for all values[^1]

[^1]: although.... shouldn't the federal https://statistikportal.de be `preferred` if it includes all values?

config/initializers/tag2link.rb
k-yle: yeah I just didn't want to pollute the global namespace.

this is much more elegant, thanks. i'm not a ruby developer so any suggestions about code style are most welcome

AntonKhorev:
Can we start with rejecting `"source": "wikidata:P3303"` *third-party formatter URL*? Because why should `icao=*` be linked to uk.flightaware.com and not anything else?

1ec5:
Or at least preferring `osmwiki:P8` and `wikidata:P1630` as sources before falling back to `wikidata:P3303`. But I assume the `preferred` rank would take care of most of that anyways.

AntonKhorev:
> Or at least preferring `osmwiki:P8` and `wikidata:P1630` as sources before falling back to `wikidata:P3303`. But I assume the `preferred` rank would take care of most of that anyways.

In my example with `icao=*` preferring `osmwiki:P8` and `wikidata:P1630` OR preferring `preferred` wouldn't change anything because such entries don't exist. If there's an official formatter, it's going to be preferred over unofficial ones. The problem is when there isn't one, because you get linked to some random site. This is not what is being asked for in the issues this PR attempts to close.

k-yle:
i think excluding all [`wikidata:P3303`](https://wikidata.org/wiki/Property:P3303) is reasonable.  There are only 4 keys where every URL is unofficial (`icao`, `ref:admiralty`, `ref:UAI`, `uic_ref`). So these 4 would have no link.

<details><summary>source</summary>

```js
const { default: list } = await import('https://esm.sh/tag2link');
Object.values(
  Object.groupBy(
    list.filter(item => item.rank !== 'deprecated'),
    item => item.key
  )
).filter(sublist => sublist.every(item => item.source === 'wikidata:P3303'));
```

</details>

config/initializers/tag2link.rb
k-yle: > [...] in some cases there is more than one that is `preferred`.

Wikidata considers this an error:

image


After exluding [`wikidata:P3303`](https://wikidata.org/wiki/Property:P3303) and cleaning up a few cases where the only difference was `www.` or `http[s]`, there are only 4 left which are ambiguous:

`de:amtlicher_gemeindeschluessel`, `ref:INEP`, `ref:ruian`, `woeid`

So... I think we just exclude any keys where there isn't a single-best URL.

<details>
<summary>source</summary>

```js
const { default: uniqBy } = await import("https://esm.sh/lodash.uniqby");
const { default: list } = await import("https://esm.sh/tag2link");

const RANKS = ["normal", "preferred"];

Object.fromEntries(
  Object.values(
    Object.groupBy(
      list.filter(
        (item) => item.rank !== "deprecated" && item.source !== "wikidata:P3303"
      ),
      (item) => item.key
    )
  )
    .map((sublist) =>
      // remove items with the same URL and sort by rank
      uniqBy(
        sublist.sort((a, b) => RANKS.indexOf(b.rank) - RANKS.indexOf(a.rank)),
        (item) => item.url
      )
    )
    .filter((sublist) => {
      // keep only those where the best & second-best have the same rank
      const [best, secondBest] = sublist;
      return best && secondBest && best.rank == secondBest.rank;
    })
    .map((sublist) => [
      // print the ones with the best rank
      sublist[0].key,
      sublist.filter((item) => item.rank === sublist[0].rank),
    ])
);
```


</details>

mmd-osm:
I checked the results returned by `JSON.parse(Rails.root.join("node_modules/tag2link/index.json").read)`, and noticed that none of the entries have `rank=deprecated`. 

Google+ as an example is marked as deprecated on the OSM Wiki / taginfo (https://taginfo.openstreetmap.org/keys/contact:google_plus), while wikidata points to an web archive link to unearth long gone pages.

```
 {"key" => "Key:contact:google_plus", "url" => "https://web.archive.org/web/20190401000000/https://plus.google.com/$1", "source" => "wikidata:P1630", "rank" => "normal"},
```

So it seems that wikidata takes precedence over the wiki?

What would be really helpful would be some kind of documentation for users to correct incorrect or missing links.
It should describe in a few words which data sources are considered in which order and which rules are applied. Anything that helps this to be more of a self-service would be really good.

How often is the NPM updated and who should we contact if the updates somehow get stuck?

Otherwise this looks good to me :+1:

k-yle:
> [...] none of the entries have `rank=deprecated`

Tag2link [exclud](https://github.com/JOSM/tag2link/blob/eacf0ec9b5a39611f83515f2b3d7abed226187cb/tag2link.dataitem.sparql#L6)[es deprecated ranks](https://github.com/JOSM/tag2link/blob/eacf0ec9b5a39611f83515f2b3d7abed226187cb/tag2link.wikidata.sparql#L10) itself (as of https://github.com/JOSM/tag2link/issues/3), but we also explicitly exclude deprecated ranks in case the behaviour changes upstream.


> Google+ as an example is marked as deprecated on the OSM Wiki / taginfo [...]

The word "deprecated" is used for two different purposes on the wiki. A tag can have "[status](https://osm.wiki/Tag_status)=deprecated" in the OSM wiki's infobox. This is different to "[rank](https://wikidata.org/wiki/Help:Ranking)=deprecated" in wikibase. The tag [status](https://osm.wiki/Tag_status) is irrelevant here.

> So it seems that wikidata takes precedence over the wiki?

We don't prioritise based on source, only [based on rank](https://github.com/openstreetmap/openstreetmap-website/blob/981d9750bab9f9dd548103d3ab0f783cca68adf0/config/initializers/tag2link.rb#L11). For Google+: 

* On the OSM wiki, [Key:contact:google_plus <sub>(Q1184)</sub>](https://osm.wiki/Item:Q1184) has no [formatter URL <sub>(P8)</sub>](https://osm.wiki/Property:P8)
* On wikidata, [Google+ ID<sub>(P2847)</sub>](https://wikidata.org/wiki/Property:P2847) does have a [formatter URL<sub>(P1630)</sub>](https://www.wikidata.org/wiki/Property:P1630), which is what tag2link [is using](https://github.com/JOSM/tag2link/blob/eacf0ec9b5a39611f83515f2b3d7abed226187cb/index.json#L176-L181).

> What would be really helpful would be some kind of documentation [...]

There's [a little bit of documentation here](https://github.com/JOSM/tag2link#osm-wiki-data), I could create an OSM Wiki page with more info (?) or were you thinking of more documentation on GitHub?

> How often is the NPM updated

seems to be [automatically updated on the 21st day](https://github.com/JOSM/tag2link/blob/eacf0ec9b5a39611f83515f2b3d7abed226187cb/.github/workflows/scheduled_update.yaml#L4) of the month.

> [...] and who should we contact if the updates somehow get stuck?

@stoecker I assume you're the current maintainer?

tomhughes (approved):
This looks good to me now. Thanks for working on it.

gabortim:
> How often is the NPM updated and who should we contact if the updates somehow get stuck?

The repo should be updated monthly. Contact JOSM devs on https://github.com/JOSM/tag2link repo or on the good old Trac.

PR #6198, opened 2025-07-15
Keep search input until sidebar close

hlfan:
Addressing a part of #3123 to keep the search input longer, now that #6184 simplified the state management.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -347,6 +347,7 @@ $(function () {
   });
 
   $(document).on("click", "#sidebar .sidebar-close-controls button", function () {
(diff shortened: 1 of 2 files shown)


tomhughes:
What exactly is this supposed to change? Because I can't manage to find anything that's behaving differently with it merged...

hlfan:
From the search, when selecting a result, the search input stays. It will persist when navigating to history or export, and only be removed once the sidebar is closed via the close button.

tomhughes:
That's what I thought it was supposed to do but for some reason it didn't seem to be doing that when I tried it.

Today it does seem to be working though so I must have been doing something wrong!

PR #6199, opened 2025-07-16
Add user preferences support to iD editor via server-side rendering

Deeptanshu-sankhwar:
### Summary
This PR implements server-side rendering of user preferences for the iD editor, making them accessible in the client-side JavaScript context.

### Technical Changes

**Server-side (`app/views/site/id.html.erb`):**
- Added user preference collection logic that fetches all preferences for the current user
- Converts preferences from ActiveRecord objects to a hash structure (`{key => value}`)
- Serializes preferences as JSON and includes them in the data attributes passed to the iD container

**Client-side (`app/assets/javascripts/id.js`):**
- Added preference parsing logic that extracts user preferences from the container's dataset
- Parses JSON-encoded preferences and makes them available globally via `window.OSM.userPreferences`
- Provides fallback to empty object for unauthenticated users or users without preferences

### Implementation Details
- Follows existing pattern used for other iD editor data (token, locale, asset_map)
- User preferences are fetched via `current_user.preferences.each_with_object({})` for efficient hash construction
- Data flow: Database → Ruby hash → JSON string → HTML data attribute → JavaScript object
- Preferences are only included when `current_user` exists (authenticated users)
- Client-side code gracefully handles missing or malformed preference data

### Use Case
This enables the iD editor to access user-specific preferences (such as display settings, editor configurations, etc.) that were previously only available through separate API calls, improving performance and reducing client-server round trips during editor initialization.

### Testing
- Verified with authenticated users having preferences
- Verified with authenticated users having no preferences  
- Verified with unauthenticated users
- Confirmed preferences are accessible via `window.OSM.userPreferences` in browser console

### DIFF
--- a/app/views/site/id.html.erb
+++ b/app/views/site/id.html.erb
@@ -10,20 +10,17 @@
 </head>
 <body>
 <% data = {}
-  if Settings.key?(:id_application)
-    token = current_user.oauth_token(Settings.id_application)
-    data[:token] = token.token
-  end
-  data[:locale] = ID::LOCALES.preferred(preferred_languages).to_s
-  data[:asset_map] = assets("iD").to_json
+   if Settings.key?(:id_application)
+     token = current_user.oauth_token(Settings.id_application)
+     data[:token] = token.token
+   end
+   data[:locale] = ID::LOCALES.preferred(preferred_languages).to_s
+   data[:asset_map] = assets("iD").to_json
 
-  user_preferences = current_user.preferences.each_with_object({}) do |pref, hash|
-    if pref.k.start_with?("iD-")
-      hash[pref.k] = pref.v
-    end
-  end
-  data[:user_preferences] = user_preferences.to_json
-%>
+   user_preferences = current_user.preferences.each_with_object({}) do |pref, hash|
+     hash[pref.k] = pref.v if pref.k.start_with?("iD-")
+   end
+   data[:user_preferences] = user_preferences.to_json %>
 <%= tag.div "", :id => "id-container", :data => data %>
 </body>
 </html>


app/assets/javascripts/id.js
```diff
@@ -14,6 +14,15 @@ document.addEventListener("DOMContentLoaded", function () {
...
     idContext.connection().apiConnections([]);
     const url = location.protocol + "//" + location.host;
+
+    // Access user preferences from server-side rendered data
+    const userPreferences = container.dataset.userPreferences ?
+      JSON.parse(container.dataset.userPreferences) : {};
+
+    // Make user preferences available globally in iD context
+    window.OSM = window.OSM || {};
+    window.OSM.userPreferences = userPreferences;
```
tyrasd: This could probably better handled in method of the `idContext` object, instead of as a new global variable.

tyrasd:
> enables the iD editor to access user-specific preferences

Just for clarity: this is not quite yet implemented in the current version of iD, but @Deeptanshu-sankhwar is working on it right now. This means that the code will only be needed to be merged once we're deploying the next release of iD which comes with that code (that PR than probably be replacing this one). But we thought it might be nice to put this out anyway for you to give it a look already.

app/views/site/id.html.erb
```diff
@@ -15,7 +15,15 @@
      data[:token] = token.token
    end
    data[:locale] = ID::LOCALES.preferred(preferred_languages).to_s
-   data[:asset_map] = assets("iD").to_json %>
+   data[:asset_map] = assets("iD").to_json
+
+  # Include user preferences for iD editor
+   if current_user
```
tyrasd: it's probably not required to check for `current_user` as this page requires to be logged in to load, and the variable is already used on line 14.

app/views/site/id.html.erb
```diff
@@ -15,7 +15,15 @@
...
+     user_preferences = current_user.preferences.each_with_object({}) do |pref, hash|
```
mmd-osm: Do we really need all user preferences for iD? This could be a fair amount of data, such as lots of overpass turbo queries. How about only adding preferences that are relevant for iD?

app/assets/javascripts/id.js
Deeptanshu-sankhwar: I have now embedded a function in the `idContext` object  to store and then consume user preferences

app/views/site/id.html.erb
Deeptanshu-sankhwar: Thanks @tyrasd , I have removed the redundant check now

pnorman:
> This enables the iD editor to access user-specific preferences (such as display settings, editor configurations, etc.) that were previously only available through separate API calls, improving performance and reducing client-server round trips during editor initialization.

It seems that all of these would come from one call of `/api/0.6/user/preferences`. As it's only one call, it doesn't seem like it would have a large performance impact. Or is the concern that a user might have other large preferences (e.g. overpass turbo queries) and the `/preferences` result will be too large?

app/views/site/id.html.erb
```diff
@@ -10,12 +10,20 @@
...
-     token = current_user.oauth_token(Settings.id_application)
-     data[:token] = token.token
-   end
-   data[:locale] = ID::LOCALES.preferred(preferred_languages).to_s
-   data[:asset_map] = assets("iD").to_json %>
+  if Settings.key?(:id_application)
+    token = current_user.oauth_token(Settings.id_application)
+    data[:token] = token.token
+  end
+  data[:locale] = ID::LOCALES.preferred(preferred_languages).to_s
```
hlfan: Could you run [`bundle exec erb_lint . --autocorrect`](https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#coding-style)?

Deeptanshu-sankhwar: Sure @hlfan , thanks for pointing out. I've pushed after running this lint fix

hlfan: And all the changes should probably be squashed into one commit.

PR #6200, opened 2025-07-16
Use svg symbols for map controls icons

hlfan:
Cleaning up the map controls sprite creation after #6189 added the way to do that.


tomhughes:
Looks good to me, thanks.

PR #6202, opened 2025-07-17
Stop Commons from redirecting outside CSP

hlfan:
For mobile devices, the request to `OSM.WIKIMEDIA_COMMONS_URL` from #6127 will be redirected to `commons.m.wikimedia.org`, which is subsequently blocked by the CSP.

Directly requesting the desktop version removes the extra redirect.


app/assets/javascripts/index/element.js
```diff
@@ -171,8 +171,8 @@
       .addClass("bg-body-tertiary");
 
     if (icon && OSM.WIKIMEDIA_COMMONS_URL) {
-      let src = OSM.WIKIMEDIA_COMMONS_URL + "Special:Redirect/file/" + encodeURIComponent(icon);
-      if (!icon.endsWith(".svg")) src += "?width=128";
+      let src = OSM.WIKIMEDIA_COMMONS_URL + "Special:Redirect/file/" + encodeURIComponent(icon) + "?mobileaction=toggle_view_desktop";
```
1ec5, line 174: In cursory testing, this doesn’t cause subsequent user-initiated page loads on Commons to switch back to desktop view, which is good. Still a bit counterintuitive in this code, so a comment explaining the parameter would be a good idea.

1ec5 (commented):
This is another subtle difference between Special:Redirect and the MediaWiki API, which is why the initial request for the item details doesn’t run into this problem.

app/assets/javascripts/index/element.js
AntonKhorev: A longer comment can be added to the commit message, to avoid making whoever finds this change looking up its pull request.

AntonKhorev:
Merged, thanks.

Apparently GitHub doesn't generate notifications if the only thing that changes are commit messages. You might want to leave a reply in this case.

PR #6204, opened 2025-07-18
Reorder paginated items

AntonKhorev:
1. Go to the diary of someone who has a lot of entries written, for example https://www.openstreetmap.org/user/SomeoneElse/diary
2. Press the [*>>* button](https://www.openstreetmap.org/user/SomeoneElse/diary?after=0) to get to the earliest entries
3. Notice that it doesn't work

The actual bug existed before *>>* was added. *>>* here is just the easiest way to notice it.

The reason is that the `diary_entries` association of the user model has an order defined as `order(:created_at => :desc)`. When the pagination method sorts user diary entries by id, that order gets added after the existing order. The entries are still going to be sorted by `:created_at => :desc` first and by id next. The added order by id doesn't do anything, therefore ascending order by id doesn't work, therefore pages using the `after` parameter don't work correctly.

### DIFF
--- a/app/controllers/concerns/pagination_methods.rb
+++ b/app/controllers/concerns/pagination_methods.rb
@@ -11,11 +11,11 @@ module PaginationMethods
 
     qualified_cursor_column = "#{items.table_name}.#{cursor_column}"
     page_items = if params[:before]
-                   items.where("#{qualified_cursor_column} < ?", params[:before]).order(cursor_column => :desc)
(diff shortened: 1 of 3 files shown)


hlfan (approved):
Personally, I'd define `check_pagination_of_50_entries` before `test_index_paged`, but this is mostly for git blame tidiness.

AntonKhorev:
It's a private method. We have private methods at the end of the file.

tomhughes:
Looks good to me, thanks.

PR #6205, opened 2025-07-18
Remove inkscape/sodipodi tags/attributes from osm forum icon

AntonKhorev:
...because I looked at the file in #6159.


tomhughes:
Looks good to me, thanks.

PR #6206, opened 2025-07-18
Change edit link to profile location in dashboard

AntonKhorev:
When you don't have your home location set, this link in your dashboard leads to *Profile Description*:

image

It makes more sense to link to *Profile Location* instead.


tomhughes:
Looks good to me, thanks.

PR #6210, opened 2025-07-20
Restore about section partial

hlfan:
`_about_section.html.erb` was removed in #4868 due to the incompatibility with manually inlined icons.
With the icons inlined with the new gem, we can re-add the layout partial.
This slims #5938. See also the changed files in [this diff](https://github.com/openstreetmap/openstreetmap-website/compare/c36f09b28db7244066df53b9e52c7ab32a1693da..4f163a434f5b15c3ce4d602d87ed6119e3b3b221#diff-20b1b4068a14bbb692bf60b143f2cd17d495893dbd397e41eadeb732056341d7).


tomhughes:
Looks good to me, thanks.

PR #6211, opened 2025-07-20
Make Valhalla use fork icons

hlfan:
Valhalla seems to use "stay" for forks, and "slight" icons are already separate.
I was unable to find more concrete documentation, though.


tomhughes:
Looks good to me, thanks.

PR #6212, opened 2025-07-20
Use more Rails' form tag helpers in directions form

hlfan:
### Description
Using a few more form tag helpers makes the search layout quite a bit more readable.
Less repetition and better integration of the inline_svg tags following #5938.
### How has this been tested?
Local installation.


tomhughes:
Looks good to me, thanks.

PR #6213, opened 2025-07-20
Add static vector style dark mode switching

hlfan:
Nothing dynamic yet, as that's also not yet implemented for the Thunderforest raster tiles, but at least basic dark mode compatibility for another layer.


tomhughes:
Are you happy with this @pnorman?

hlfan:
So, now I'm happy. Gotta use the layers infrastructure that's already built.

pnorman:
> Are you happy with this @pnorman?

The eclipse style isn't great, but  it's probably no worse than the client-side color changes made to OpenStreetMap Carto so I'm okay with it.

hlfan:
Anything else needed @tomhughes?

tomhughes:
Looks good to me, thanks.

PR #4771, opened 2024-05-08
Make blank role icons transparent

AntonKhorev:
Before:
image
image

After:
image
image

PR #6215, opened 2025-07-21
Replace custom role icon helpers with inline_svg_tag

AntonKhorev:
Inserts these icons using `inline_svg_tag`
image
image

I'm doing this also to check where it's best to store color definitions, see https://github.com/openstreetmap/openstreetmap-website/pull/5854#discussion_r2216313135. `parameters.scss` seems like a possible place because we already have some colors in it.


hlfan:
Given your preference for fills and strokes lining up on the pixel grid, as you've stated in https://github.com/openstreetmap/openstreetmap-website/pull/5866#issuecomment-2764695781, the stroke width of 1.5 for blank role icons that you introduced in #4771 seems like something to take a closer look at:

| stroke width 1.5px | stroke width 2px |
|-|-|
| image | image |

You may notice the icon in this example is offset on the x-axis by a subpixel, which is why the pixels are not quite symmetrical. Still, with a 2px stroke, the main horizontal bar is fully opaque, the 1.5px bar isn't.

I suggest reverting the stroke-width change from #4771, then the only other change needed can be overwritten by the `SvgFillOverrideTransform` from #5884. Then there's no need the store essentially the same icon twice.

AntonKhorev:
I remove `SvgFillOverrideTransform` in another pull request.

It doesn't matter much how unfilled role icons look because only admins are going to see them.

hlfan:
Does it matter that the colors you define (wherever) are not the colors that end up on the screen due to the resulting lower opacity?

app/assets/stylesheets/parameters.scss
```diff
@@ -20,3 +20,13 @@ $table-border-factor: .1;
...
+
+.role-icon-administrator {
+  color: #f69e42;
+}
+.role-icon-moderator {
+  color: #447eff;
+}
+.role-icon-importer {
+  color: #38e13a;
+}
```
tomhughes: These aren't really parameters. are they?

hlfan: How about renaming `common.scss` to `screen.scss` and putting these screen + print definitions in a new `common.scss`?

AntonKhorev:
Here's a version with one svg where fill is altered. No inline_svg_tag transforms are required for that because you can change svg properties using css.

app/assets/stylesheets/parameters.scss
AntonKhorev: Put them in another file, `colors.scss`.

I hope to use `common.scss` for print style later, Bootstrap has [some support for that](https://getbootstrap.com/docs/5.3/utilities/display/#display-in-print).

hlfan:
The `svg.role-icon.inactive` rule could be generalized into bootstrap-style `.fill-none`. Probably better than much custom stuff.

AntonKhorev:
There's no `fill="none"` there. I've restored how "blank" icons look to what they were years ago:

https://github.com/openstreetmap/openstreetmap-website/blob/12d797b3513bba475a8939eda9fa202f28bc5306/app/assets/images/roles/blank_administrator.svg

https://github.com/openstreetmap/openstreetmap-website/blob/12d797b3513bba475a8939eda9fa202f28bc5306/app/assets/images/roles/blank_moderator.svg

https://github.com/openstreetmap/openstreetmap-website/blob/12d797b3513bba475a8939eda9fa202f28bc5306/app/assets/images/roles/blank_importer.svg

tomhughes:
Looks good to me, thanks.

PR #6216, opened 2025-07-21
Update "Latest diaries" icons

AntonKhorev:
What I wanted is to use the same comments icon with the same filled-unfilled logic for latest diary cards as for changesets. But once I did that, I had to update other icons in the cards to match the style of comment icons. I took them all from Bootstrap Icons. And I updated title icons and text next to them to better support multiple lines.

Before:
image

After:
image


app/assets/images/icons/calendar.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" viewBox="0 0 16 16">
```
hlfan, line 1: Since you overwrite the image size anyway, is the width and height in this file still needed?

hlfan (approved):
Ah yes, the bootstrap icons, where `A7.5 6.5 0 0 1 .5 8 7.5 6.5 0 1 1 8 14.5a7.5 6.5 0 0 1-2.4-.34` is for some reason written as the combination of `A9 9 0 0 0 8 15c4.418 0 8-3.134 8-7s-3.582-7-8-7-8 3.134-8 7c0 1.76.743 3.37 1.97 4.6` and `A8 8 0 0 0 8 14c3.996 0 7-2.807 7-6s-3.004-6-7-6-7 2.808-7 6c0 1.468.617 2.83 1.678 3.894`.

And SVGO also can't compress this down as far, but since content size has not been a point of contention so far I don't think this matters.

app/assets/images/icons/calendar.svg
AntonKhorev: `xmlns` is also not needed, but you want it for some reason.

I overwrite the size here, but maybe I won't overwrite it elsewhere, if I use this image elsewhere.

hlfan: I see you also removed the "bi" classes that Bootstrap uses to set the icon size to the font size that would set the size for you.

`xmlns` isn't needed when part of an HTML file, but for standalone files, as they are stored, ["an SVG namespace declaration must be provided so that all SVG elements are identified as belonging to the SVG namespace."](https://www.w3.org/TR/SVG/struct.html#Namespace)

AntonKhorev: If you care about standalone files, why don't you want them to have sizes defined? ("bi" wouldn't set sizes of standalone files)

hlfan: If I open the file in the browser standalone, I'd rather have it fill the viewport than be a speck of dust at the corner of my screen.
Github's files changed preview of the icon would also be larger, so differences can be spotted more easily.

tomhughes:
This looks fine to me, and the way things are sized appears to be no different to the other icon images we've been adding so whatever the rights or wrongs of that discussion I don't think it needs to hold this up as we can always changes things like that later.

PR #6217, opened 2025-07-21
Remove temporary debug code

hlfan:
~~Making RuboCop happier~~

### DIFF
--- a/test/controllers/api/maps_controller_test.rb
+++ b/test/controllers/api/maps_controller_test.rb
@@ -125,10 +125,6 @@ module Api
       maxlat = node.lat + 0.1
       bbox = "#{minlon},#{minlat},#{maxlon},#{maxlat}"
       get api_map_path(:bbox => bbox)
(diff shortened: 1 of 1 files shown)


tomhughes:
I'm not really sure that code should be there at all - it looks like temporary debug code that somebody has made permanent but putting it behind `$VERBOSE` but when I write such code I normally take it out again when I'm done.

How do you even cause `$VERBOSE` to be set?

tomhughes:
One was introduced in f4e29905263efbb2eafdedad6a5c050ba559efd8 and the second in 75a49786f8bcb76715cf4009ba807a844baae282.

hlfan:
Not quite, f4e29905263efbb2eafdedad6a5c050ba559efd8 just moved it, it has been there since API 0.5 in commit 78b440ffc1e0448f8a6d73c8821dd58634ffb475.

tomhughes:
Looks good to me, thanks.

PR #6228, opened 2025-07-23
Bundle vector layer maxZoom definitions in layers.yml

### LINKED ISSUE #6222: Shortbread vector tiles >=z24 causes strange panning behavior
### URL

https://www.openstreetmap.org/#map=24/36.73505736/-80.40928239&layers=S

1. Select Shortbread layer
2. Enable "Map Data" toggle
3. Zoom to z24 or greater
4. Observe that the vector tiles and map data are not aligned when panning

This doesn't happen with the MapTiler OMT tiles because the maxZoom is 23.
https://github.com/openstreetmap/openstreetmap-website/blob/298b14de9447e5d3e5e7b0258c1d174764ec486a/app/assets/javascripts/leaflet.maptiler.js#L7

https://github.com/user-attachments/assets/a99b74df-e674-4bf1-b5b7-aeecd5c066fa

hlfan:
In contrast to raster tile layers, the maxZoom for vector tile layers isn't determined by the served tiles (up to zoom level 14).
Hence, the maxZoom for vector tile layers is more of an integration decision than an inherent property of the tile layer itself.
Therefore, I argue, the maxZoom for vector tile layers should be set in the layers.yml instead.

Also resolves #6222.

### DIFF
--- a/app/assets/javascripts/leaflet.maptiler.js
+++ b/app/assets/javascripts/leaflet.maptiler.js
@@ -4,7 +4,6 @@
 L.OSM.OpenMapTiles = L.MaplibreGL.extend({
   initialize: function (options) {
     L.MaplibreGL.prototype.initialize.call(this, {
-      maxZoom: 23,
       style:
         "https://api.maptiler.com/maps/openstreetmap/style.json?key=" +
         options.apikey,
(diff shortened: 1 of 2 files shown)


tomhughes:
Yes I was struggling with whether this should be done in the layers config or in the constructors to mirror how raster layers work but your argument makes a lot of sense, thanks.

PR #6235, opened 2025-07-23
Make search show multiple markers

hlfan:
This PR improves the visualization and interactivity of search results in the sidebar and map.

- Search result entries now have permanent map markers (instead of hover-only).
- Each marker is assigned a unique color based on the golden angle[^1], improving distinction even with many results.
- List entries visually match their corresponding markers via a colored border (`--marker-color`).
- Hovering over a list entry highlights the marker, and vice versa.
- Clicking a marker forwards that to the associated list entry.
- Refactored fetch logic into a reusable `fetchReplace()` function.

This touches a multitude of issues, see the list at https://github.com/openstreetmap/openstreetmap-website/issues/872#issuecomment-2588598803.

[^1]: Why this way? This approach scales better for arbitrary result counts. The golden angle ensures that hues are distributed evenly and only begin to visually converge after over 20 entries.

https://github.com/user-attachments/assets/b9af00e5-2968-4d0e-8bc7-4f075853edf7

### DIFF
--- a/app/assets/javascripts/index/search.js
+++ b/app/assets/javascripts/index/search.js
@@ -33,11 +33,10 @@ OSM.Search = function (map) {
 
   $("#sidebar_content")
     .on("click", ".search_more a", clickSearchMore)
-    .on("click", ".search_results_entry a.set_position", clickSearchResult)
-    .on("mouseover", "li.search_results_entry:has(a.set_position)", showSearchResult)
-    .on("mouseout", "li.search_results_entry:has(a.set_position)", hideSearchResult);
+    .on("click", ".search_results_entry a.set_position", clickSearchResult);
 
   const markers = L.layerGroup().addTo(map);
+  let processedResults = 0;
 
   function clickSearchMore(e) {
     e.preventDefault();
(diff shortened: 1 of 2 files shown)


tomhughes:
How is this handling "load more" results? Does it recompute colours over the whole result set or does it compute a separate set of colours for each load?

Controversial idea - should be zoom the map to cover all the results? As it stands many of the markers may not be visible which seems a bit weird?

hlfan:
As long as the newly loaded results are after the previous list items, the index is correct.
If, e.g., the internal search results load after nominatim-reverse, then these get the same color.
Maybe incrementing an internal counter is a better idea, which could make adjacent search results be colored a bit too similarly.

The first search result includes one in New York, so maybe not all results, but only "close" ones?
And how to define "close"? Maybe let the geocoder set the view?

app/assets/stylesheets/common.scss
```diff
@@ -567,6 +567,17 @@ header .search_forms,
...
 
+/* Rules for search results */
+
+.search_results_entry li.list-group-item {
+  border-right: 1em solid var(--marker-color);
+}
+
+.leaflet-marker-icon.active > svg {
+	transform: scale(1.5);
+	transform-origin: bottom;
```
hlfan, line 578: Yup, I forgot to re-enable my pre-commit hook that guards against tabs.

app/assets/javascripts/index/search.js
```diff
@@ -48,34 +47,41 @@ OSM.Search = function (map) {
...
-      const data = $(this).find("a.set_position").data();
-
-      marker = L.marker([data.lat, data.lon], { icon: OSM.getMarker({}) });
-
-      $(this).data("marker", marker);
-    }
-
+    const index = processedResults++;
+    const listItem = $(this);
+    const color = `hwb(${(index * (Math.sqrt(5) - 1) * 180) % 360}deg 5% 5%)`;
```
tomhughes: This actually appears to be the reciprocal of the golden angle (~222.5 degrees) not the golden angle (~137.5 degrees) at least according to https://en.wikipedia.org/wiki/Golden_angle?

The golden angle would be:

```suggestion
    const color = `hwb(${(index * (3 - Math.sqrt(5)) * 180) % 360}deg 5% 5%)`;
```

Possibly that calculation should be pulled out to a `goldenAngle` variable that is then used in the template, as a way of documenting what we're doing?

hlfan: I found that the first colors, when rotating the other way, line up more with the other marker colors:
|+angle|-angle|
|-|-|
|`#0df250`|`#0d50f2`|
|`#930df2`|`#93f20d`|

Yeah, pulling out that variable is a good idea.

tomhughes:
Maybe the markers should get larger when you mouse over them, to match what happens when you mouse over the list entries? It would help make it obvious which marker you were over when several are close together?

tomhughes:
I fear we might get some complaints that this is a bit garish but lets give it a go and see what happens... Thanks for all the work on it.

nebulon42:
I just wanted to mention that I think this is a great UX enhancement for search. Yes, you can discuss about the colors, maybe toning them down a bit would be nice, but overall great stuff. Many thanks for this and to all working on improving the website.

jidanni:
There needs to be a mouseover, else if there is only one marker visible,
users won't figure it out.
https://community.openstreetmap.org/t/whats-that-stripe-next-to-search-results/138432

PR #6236, opened 2025-07-23
Fix vector tile attribution issues

tomhughes:
This hopefully fixes the vector tile attribution issues.

The second part is somewhat gross but hard to avoid until we can get leaflet-maplibre-gl fixed because it registers a load callback on the underling GL map that will call `getAttribution` and add the result to the attribution control and then never remove it again.

To avoid that we temporarily replace `getAttribution` with a dummy version that it will capture for it's callback...


hlfan:
I think `L.OSM.MaplibreMap` would be a misnomer, since `L.OSM.Map` has the whole custom layer & object & integration logic.
`L.OSM` defines itself a `L.OSM.TileLayer`, so `L.OSM.MaplibreGL` would follow that naming logic.

tomhughes:
All review comments have been addressed.

petr-hajek (approved):
tested and works well, thanks!

petr-hajek:
@tomhughes can we merge this?

tomhughes:
@petr-hajek well as far as I'm concerned it's ready but as I wrote it I prefer not to merge it myself.

I've just simplified it in fact now my upstream fix for sticky attributions is deployed.

petr-hajek:
@AntonKhorev maybe you could have a look and merge this? Many thanks!

AntonKhorev:
Merged, thanks.

PR #6239, opened 2025-07-24
Clean up some svg-related markup

hlfan:
Two things:

- A thing I missed in #6200 that was made possible by not needing to sometimes render the definition.
- A thing where the 13 minutes between the merges of #6215 and #5854 weren't enough to adapt the code to the test-drive.


tomhughes:
Looks good to me, thanks.

PR #6248, opened 2025-07-25
feat: add automatic 3rd party OAuth provider redirect

matisiekpl:
### Description

Following Pull Request proposes adding new query parameter to login page that allows to specify OAuth provider that user
would use for logging in.

### Usage
Navigating user to `https://openstreetmap.com/login?&autologin_provider=google` would automatically trigger redirect to Google Sign In.
Parameter accepts values such as `google`, `facebook`, `microsoft`, `github`, `wikipedia`.

### Motivation
I am founder of https://aedmapa.pl/ mobile app that allows users to browse nearby Automated External Defibrillators (AEDs).
My app allows to contribute to OSM database by adding new AEDs. I'm trying to improve contribution experience.

I get a lot of emails from users that asks for support adding new AEDs. In my app there is a flow for adding new AED. It looks like that:
1. Login (show OpenStreetMap OAuth screen)
2. Authenticate (user successfully signed in using OSM account)
3. Select desired location
4. Insert AED

Untitled report (1)

**Looking at conversation metrics, there is a big 93% drop-off during login process.**
I was trying to investigate this drop-off and I've few hypothesis (mostly gathered from email / phone conversations with users).
- Most of my users are not OpenStreetMap contributors - my app is a first contact for them with OSM data. They don't have OpenStreetMap account.
- Users says they don't want to create new account
- Google Sign In button is simply hidden until the first scroll (please see screenshot).

IMG_0593

In Polish Language version, the buttons are completely invisible.

IMG_0594

My proposed solution to the problem is to automatically trigger Google Sign In flow. Following PR implements this on OSM website side.

### How has this been tested?

This code has been tested on my local development machine with multiple values passed as query parameter.

### DIFF
--- a/app/assets/javascripts/auth_providers.js
+++ b/app/assets/javascripts/auth_providers.js
@@ -9,7 +9,6 @@ $(function () {
   // Auto-click authentication button if autologin_provider query parameter is present
   const urlParams = new URLSearchParams(window.location.search);
   const autologinProvider = urlParams.get('autologin_provider');
-  console.log(window.location.search);
   if (autologinProvider) {
     const providers = ['google', 'facebook', 'microsoft', 'github', 'wikipedia'];
     if (providers.includes(autologinProvider)) {


tomhughes:
We've already got a preferred provider option - presumably you're aware of that?

Given that was as far as we were prepared to go for Microsoft I'm not sure how keen we're going to be to revisit that now and go further.

matisiekpl:
Thanks for mentioning that param @tomhughes . I will try with that. Closing

PR #6250, opened 2025-07-25
Overwrite marker shadow interactability

hlfan:
Since Leaflet's divIcon doesn't support putting elements in the shadow pane, all `OSM.getMarker` shadows are currently interactive.
This PR overwrites this behaviour.


tomhughes:
Looks good to me, thanks.

PR #6251, opened 2025-07-25
Use SVG marker in embed

hlfan:
Since #6249 mentioned retina displays on the embed page, I changed my decision on whether to put the SVG marker there too.
This also already overwrites the interactivity for the marker, which is tackled by #6250 for the main page.


tomhughes:
Looks good to me, thanks.

PR #6252, opened 2025-07-25
Retire jquery-simulate

mmd-osm:
Remove jquery-simulate, replace it by https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click

Follow up for https://github.com/openstreetmap/openstreetmap-website/pull/6192#discussion_r2230801035

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -1,4 +1,3 @@
-//= require jquery-simulate/jquery.simulate
 //= require ./history-changesets-layer
 
 OSM.History = function (map) {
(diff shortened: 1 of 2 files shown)


tomhughes:
The one thing it looses, and the reason the old code passed the original event down, is the ability to have a shift click on the map be treated as a shift click on the list entry.

hlfan:
Ctrl and Shift click opened the changeset in a new tab resp. window for me. Where does it loose that functionality?

mmd-osm:
https://github.com/openstreetmap/openstreetmap-website/pull/6254 looks like the better option... closing here.

tomhughes:
I just tried it again and neither shift-click or ctrl-click seems to work for my in firefox or chrome on linux.

I agree that #6254 is a better solution anyway so I don't think we need to investigate further.

mmd-osm:
So this PR doesn't have the issues we're seeing with mouse move events, as per #6265....

PR #6254, opened 2025-07-26
Add native links to HistoryChangesetsLayer

hlfan:
Alternative to #6252 that closes #6252.
UX difference is that this makes the URL show in the browser's status bar when hovering over a changeset bbox:
image


tomhughes:
Looks good to me, thanks.

PR #6255, opened 2025-07-26
Map XML download with Turbo

mmd-osm:
Map XML download currently has a few issues, such as a button staying in disabled state after a download, or error messages that are shown on a blank page.

Applying the same concepts as in #6192, this PR aims at making the download a bit more robust and user friendly.

We have a bit of code duplication here, and it would probably make sense to move downloadBlob, handleExportError and  turbo:submit-end handler to a common file. Any ideas?

### DIFF
--- a/app/assets/javascripts/download_util.js
+++ b/app/assets/javascripts/download_util.js
@@ -1,0 +1,53 @@
+(function () {
+  OSM.downloadBlob = function (blob, filename) {
+    const url = URL.createObjectURL(blob);
+    const a = document.createElement("a");
+    a.href = url;
+    a.download = filename;
+    document.body.appendChild(a);
+    a.click();
+    document.body.removeChild(a);
+    URL.revokeObjectURL(url);
+  };
+
+  class DownloadUtil {
+    static async handleExportSuccess(fetchResponse, filename) {
+      try {
+        const blob = await fetchResponse.response.blob();
+        OSM.downloadBlob(blob, filename);
+      } catch (err) {
+        // eslint-disable-next-line no-alert
+        alert(OSM.i18n.t("javascripts.share.export_failed", { reason: "(blob error)" }));
+      }
+    }
+
+    static async handleExportError(event) {
+      let detailMessage;
+      try {
+        detailMessage = event?.detail?.error?.message;
+        if (!detailMessage) {
+          const responseText = await event.detail.fetchResponse.responseText;
+          const parser = new DOMParser();
+          const doc = parser.parseFromString(responseText, "text/html");
+          detailMessage = doc.body ? doc.body.textContent.trim() : "(unknown)";
+        }
+      } catch (err) {
+        detailMessage = "(unknown)";
+      }
+      // eslint-disable-next-line no-alert
+      alert(OSM.i18n.t("javascripts.share.export_failed", { reason: detailMessage }));
+    }
+
+    static getTurboBlobHandler(filename) {
+      return function (event) {
+        if (event.detail.success) {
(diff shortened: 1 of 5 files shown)


hlfan:
I could see the whole thing, including handleExportSuccess and handleExportError, made into a higher-order function that's called like:
```js
document.querySelector(".export_form").addEventListener("turbo:submit-end", getTurboDownload("map.osm"));
```
Or a more modular version with a callback:
```js
document.querySelector(".export_form").addEventListener("turbo:submit-end", getTurboBlobHandler(blob => downloadBlob(blob, "map.osm")));
```

But in any case, I'd like to have `downloadBlob` separately to reuse it for the directions geojson download.

As for where to store them, that's probably a question of whether there are any things to download outside of the index/map_layout pages; otherwise, I'd put it there and in the OSM namespace.

mmd-osm:
I did a bit of refactoring based on the callback idea. Still a first draft for now, we can move things around as needed. I wasn't exactly sure if downloadBlob is sufficiently separated from the rest.

hlfan:
I'd replace the `javascript_include_tag`s with `//= require download_util` to not bind it to the views.

app/assets/javascripts/leaflet.share.js
```diff
@@ -28,52 +30,13 @@ L.OSM.share = function (options) {
...
-    async function handleExportSuccess(fetchResponse) {
-      try {
-        const blob = await fetchResponse.response.blob();
-        const filename = OSM.i18n.t("javascripts.share.filename");
-        downloadBlob(blob, filename);
-      } catch (err) {
-        // eslint-disable-next-line no-alert
-        alert(OSM.i18n.t("javascripts.share.export_failed", { reason: "(blob error)" }));
-      }
-    }
```
hlfan, line 33: ‌

mmd-osm: Somehow the comment here got lost...

app/assets/javascripts/download_util.js
```diff
@@ -0,0 +1,53 @@
...
+          detailMessage = doc.body ? doc.body.textContent.trim() : "(unknown)";
+        }
+      } catch (err) {
+        detailMessage = "(unknown)";
+      }
+      // eslint-disable-next-line no-alert
+      alert(OSM.i18n.t("javascripts.share.export_failed", { reason: detailMessage }));
+    }
+
+    static getTurboBlobHandler(downloadCallback) {
```
tomhughes: This is setup to return a function that captures the callback but in fact both users are using very similar callbacks - in particular they are both just forwarding to `OSM.downloadBlob`.

So wouldn't make more sense for this just to take the filename (which is the only thing that changes and needs to be captured) and put the `downloadBlob` call in here?

mmd-osm: Well in this case, we could also replace the callback in handleExportSuccess by the filename, and simply call OSM.downloadBlob right there.

tomhughes: Indeed, and once all that is done `downloadBlob` doesn't really need to be exported in `OSM` and could just be a function in the `DownloadUtil` class?

mmd-osm: @hlfan mentioned earlier on _I'd like to have downloadBlob separately to reuse it for the directions geojson download._

So that's why I kept it outside the class for now.

app/assets/javascripts/leaflet.share.js
hlfan: Oh sorry, I was wanting to ask if we need two blank lines here. The previous didn't have two either, and this version is even shorter.

tomhughes:
Thanks... I think this looks good then.

PR #6258, opened 2025-07-27
Lower max allowed blank line count

hlfan:
Only very few affected gaps, none of which looked like it was needed for contextual separation.


tomhughes:
Looks good to me, thanks.

PR #6260, opened 2025-07-27
Make handler able to handle SVG links

hlfan:
SVG anchors don't have the URL properties, which forces a page reload. This can be avoided by getting the properties from the href.


tomhughes:
Looks good to me, thanks.

PR #6261, opened 2025-07-27
Always show marker shadows

hlfan:
For the affected note makers, this makes them stand out more against the background and each other.


tomhughes:
Yes I'm not sure there was any good reason why some markers has shadows and some didn't, so this looks good to me, thanks.

PR #6262, opened 2025-07-27
Remove hard-coded stuff in header

hlfan:
Aligning the buttons with flexbox means they move up a pixel or two.
The locale selector button adapts its height to the available space, now 35px high instead of 36.5px.
Setting the menu item stroke and width with the Bootstrap colors (after inlining it) removes the need to invert it.
Closes #6052.


tomhughes:
Looks good to me, thanks.

PR #6263, opened 2025-07-27
Bootstrap modal instead of alert()

mmd-osm:
For better UX consistency, this PR replaces alert() by a centered Bootstrap modal window. It's currently used for map export messages only.

Before:

image

After:

image

### DIFF
--- a/app/assets/javascripts/download_util.js
+++ b/app/assets/javascripts/download_util.js
@@ -10,14 +10,20 @@
     URL.revokeObjectURL(url);
   };
 
+  OSM.showAlert = function (message) {
+    const modalBody = document.getElementById("osm_alert_message");
+    modalBody.textContent = message;
+    const alertModal = new bootstrap.Modal(document.getElementById("osm_alert_modal"));
+    alertModal.show();
+  };
+
   class DownloadUtil {
     static async handleExportSuccess(fetchResponse, filename) {
       try {
         const blob = await fetchResponse.response.blob();
         OSM.downloadBlob(blob, filename);
       } catch (err) {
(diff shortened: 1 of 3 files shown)


tomhughes:
I was considering this myself but I was looking at toasts for this...

app/assets/javascripts/download_util.js
```diff
@@ -10,14 +10,21 @@
...
+    alertModal.show();
+  };
+
   class DownloadUtil {
     static async handleExportSuccess(fetchResponse, filename) {
       try {
         const blob = await fetchResponse.response.blob();
         OSM.downloadBlob(blob, filename);
       } catch (err) {
         // eslint-disable-next-line no-alert
```
hlfan: These eslint commands are no longer needed then

mmd-osm: done

mmd-osm:
I find toast messages helpful in case of success messages (eg. "Message successfully sent"). Usually, they're disappearing after a few seconds without any user interaction required. They're not distracting the user flow in any way.

Now with error messages, users probably want to take a closer look, or even take a screenshot if they can't figure out what the error is about. The default `autohide: true` is less than ideal here.

You can dismiss the bootstrap modal error message by clicking anywhere on the screen. So there isn't a huge difference to a toast, considering that the user needs to dismiss the toast message anyway (for error messages with autohide: false).

tomhughes:
Oh sure I wasn't going to make them autohide but you're right there's probably not much difference between a non-autohide toast and a modal.

This looks good to me I think.

PR #6265, opened 2025-07-28
Remove link targets while dragging the map

### LINKED ISSUE #6264: History: Dragging selects changeset
### URL

https://www.openstreetmap.org/history#map=11/53.0895/-6.3309

1. Go to https://www.openstreetmap.org/history#map=11/53.0895/-6.3309
2. Hover over https://www.openstreetmap.org/changeset/169475227
3. Click and hold left mouse button, and drag in any direction (but do not go outside the map, ie. over sidebar)
4. Drag is registered as click (or similar) and changeset is selected instead.

This happens consistently.

~~Maybe it is regression from removing onclick handler/simulation in [this commit](https://github.com/openstreetmap/openstreetmap-website/commit/fe604257b8831c0ada2f4e914558b841ee3ece32#diff-25cab16df4a58e0c8e53d0a0e70d9bbf28ef3beb9e6bf1208234fdeed7515392L29)~~
Edit: could be Leaflet issue, though it popped up only now.

Workaround: Zoom out to z=9 so it unloads or drag cursor horizontally until changeset is ie. under left sidebar out of view.

Instance openstreetmap.org
Map Standard (OSM Carto); any map style
Language on osm.org Nederlands (dutch, nl-nl), English (default british english, en-gb); any language
Logged in/out; both
Browser Chrome 138.0.7204.169, Firefox 141.0, Edge 138.0.3351.109
OS Windows 10

cc: @hlfan 

---

See also https://github.com/openstreetmap/iD/issues/10485 (similar issue)

---

Edit: based on comment browser console warnings are unrelated.

Chrome console browser console warning

```
application-d72c1f95d8d6a3166d6b7512d08ca116305a422d96d9ec8b2e1b982cd590980f.js:31705 Image " " could not be loaded. Please make sure you have added the image with map.addImage() or a "sprite" property in your style. You can provide missing images by listening for the "styleimagemissing" map event.
Z | @ | application-d72c1f95…2cd590980f.js:31705
-- | -- | --
  | _getImagesForIds | @ | application-d72c1f95…2cd590980f.js:31709
  | (anonymous) | @ | application-d72c1f95…2cd590980f.js:31709
  | getImages | @ | application-d72c1f95…2cd590980f.js:31709
  | (anonymous) | @ | application-d72c1f95…2cd590980f.js:31709
  | (anonymous) | @ | application-d72c1f95…2cd590980f.js:31705
  | e | @ | application-d72c1f95…2cd590980f.js:31705
  | getImages | @ | application-d72c1f95…2cd590980f.js:31709
  | (anonymous) | @ | application-d72c1f95…2cd590980f.js:31709
  | (anonymous) | @ | application-d72c1f95…2cd590980f.js:31705
  | (anonymous) | @ | application-d72c1f95…2cd590980f.js:31705
  | e | @ | application-d72c1f95…2cd590980f.js:31705
  | processTask | @ | application-d72c1f95…2cd590980f.js:31705
  | receive | @ | application-d72c1f95…2cd590980f.js:31705
  | (anonymous) | @ | application-d72c1f95…2cd590980f.js:31705
```

Firefox console warnings (first already visible before issue happens)

```
Content-Security-Policy: ‘'unsafe-inline'’ binnen style-src wordt genegeerd: nonce-source of hash-source opgegeven [www.openstreetmap.org](https://www.openstreetmap.org/)
MouseEvent.mozPressure wordt niet meer ondersteund. Gebruik in plaats daarvan PointerEvent.pressure. [application-d72c1f95d8d6a3166d6b7512d08ca116305a422d96d9ec8b2e1b982cd590980f.js:16801:6](https://www.openstreetmap.org/assets/application-d72c1f95d8d6a3166d6b7512d08ca116305a422d96d9ec8b2e1b982cd590980f.js)
MouseEvent.mozInputSource wordt niet meer ondersteund. Gebruik in plaats daarvan PointerEvent.pointerType. [application-d72c1f95d8d6a3166d6b7512d08ca116305a422d96d9ec8b2e1b982cd590980f.js:16801:6](https://www.openstreetmap.org/assets/application-d72c1f95d8d6a3166d6b7512d08ca116305a422d96d9ec8b2e1b982cd590980f.js)
```

Edge: Issue happens, but no browser console errors
Sidenote: It seemed to have issues with loading map(Carto) icons, but those errors were lost on logging in.

hlfan:
Fixes #6264 by removing the link target when moving the map.
Since the click event happens before moveend, and the click events aren't recognized as drag events, I don't think there's a less bodgy solution.

### DIFF
--- a/app/assets/javascripts/index/history-changesets-layer.js
+++ b/app/assets/javascripts/index/history-changesets-layer.js
@@ -5,7 +5,8 @@ OSM.HistoryChangesetBboxLayer = L.FeatureGroup.extend({
 
   addChangesetLayer: function (changeset) {
     const style = this._getChangesetStyle(changeset);
-    const anchor = $(L.SVG.create("a")).attr("href", $(`#changeset_${changeset.id} a.changeset_id`).attr("href"));
+    const link = $(`#changeset_${changeset.id} a.changeset_id`).attr("href");
+    const anchor = $(L.SVG.create("a")).attr("href", link);
     const rectangle = L.rectangle(changeset.bounds, {
       ...style,
       contextmenu: true,
@@ -21,6 +22,8 @@ OSM.HistoryChangesetBboxLayer = L.FeatureGroup.extend({
       }]
     });
     rectangle.id = changeset.id;
+    this._map?.on("movestart", () => anchor.removeAttr("href"));
+    this._map?.on("moveend", () => anchor.attr("href", link));
     rectangle.on("add", function () {
       $(this.getElement()).replaceWith(anchor).appendTo(anchor);
     });
--- a/app/assets/javascripts/index/search.js
+++ b/app/assets/javascripts/index/search.js
@@ -66,12 +66,16 @@ OSM.Search = function (map) {
   function showSearchResult() {
     const index = processedResults++;
     const listItem = $(this);
+    const data = listItem.find("a.set_position").data();
+    if (!data) return;
     const inverseGoldenAngle = (Math.sqrt(5) - 1) * 180;
     const color = `hwb(${(index * inverseGoldenAngle) % 360}deg 5% 5%)`;
     listItem.css("--marker-color", color);
-    const anchor = $("<a>").attr("href", listItem.find("a.set_position").attr("href"));
-    const data = listItem.find("a.set_position").data();
+    const link = listItem.find("a.set_position").attr("href");
+    const anchor = $("<a>").attr("href", link);
     const marker = L.marker([data.lat, data.lon], { icon: OSM.getMarker({ color, className: "activatable" }) });
+    map.on("movestart", () => anchor.removeAttr("href"));
+    map.on("moveend", () => anchor.attr("href", link));
     marker.on("mouseover", () => listItem.addClass("bg-body-secondary"));
     marker.on("mouseout", () => listItem.removeClass("bg-body-secondary"));
     marker.on("add", function () {


mmd-osm:
Somehow this doesn't work for me in Firefox. After a few mousemove and mouseup events, I'm getting click events in Leaflet, Turbo, Bootstrap and jQuery, and eventually, the single changeset page is loaded.

image

hlfan:
I'm afraid I'm out of ideas that don't include `a.click()` in some shape or another.

mmd-osm:
Maybe there's some clever way to tame all these event handlers. For the time being, I would be ok with click(), and leave the shift-click/ctrl-click topic for later. Once we move away from Leaflet, this discussion might be no longer relevant anyway.

mmd-osm:
I believe the click is handled by this "a" link handler in index.js. Once I do a preventDefault / stopPropagation on the event, I'm no longer redirected to the changeset page.

```
  $(document).on("click", "a", function (e) {
    if (e.isDefaultPrevented() || e.isPropagationStopped() || $(e.target).data("turbo")) {
      return;
    }
```

hlfan:
We don't want to stop all clicks from happening, only the ones while dragging.
The "drag" event that ends up there is indistinguishable from an intended click.
If we want links to end up there without intervention (`a.click()` or similar), we need to prevent them when dragging.
And the least convoluted way to do that is to remove the `href`.

Hence this PR.

But if the detection of "when dragging" is faulty, to me, there don't seem to be any criteria left to consider.

hlfan:
Maybe put the re-enabling in a timeout? But then that also has to be fine-tuned.

mmd-osm:
Well, i thought we need to make the code in index.js aware of mouse move / dragging events, and then discard events in the click handler, if needed. Leaflet already does that job. Unfortunately, the index.js event handler has no idea about the Leaflet state.

hlfan:
Have you looked at the types of events that end up there? The event type is _always_ `click`. That's the problem with that idea.

mmd-osm:
I think you're right, I couldn't make much progress with this approach. In the meantime, I have created an alternative for click() with key modifiers included: #6267
This is seems to work ok here...

hlfan:
Closing in favor of #6267 due to #6276.

PR #6267, opened 2025-07-29
Fix map dragging issue in history view

mmd-osm:
Try 2, now including key modifiers to open link in new tab/window.

Previous attempt in #6252

Partially reverts #6254


tomhughes:
The title is a bit misleading really because jquery-simulate has already been retired - this is really an alternate way to replace it to avoid the issues with dragging the map reported in #6264.

tomhughes:
The downside of the approach here is that it has to make assumptions about what modifiers do which may not be true in all browsers...

mmd-osm:
PR and commit message have been renamed, it makes sense to drop the reference to jquery-simulate.

I tested on Firefox, Chrome, Opera, Palemoon on Linux, +Edge on Win11. It should work on MacOS, but I haven't tested it.

mmd-osm:
Let's merge #6281 instead.

PR #6268, opened 2025-07-30
Replace javascript alerts with bootstrap modals

tomhughes:
This builds on the work from #6263 by replacing all other uses of `alert` with bootstrap modals.

### DIFF
--- a/app/assets/javascripts/download_util.js
+++ b/app/assets/javascripts/download_util.js
@@ -10,13 +10,6 @@
     URL.revokeObjectURL(url);
   };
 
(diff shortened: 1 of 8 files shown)


AntonKhorev:
Merged, thanks, although I don't think all of these messages need to be in any kind of modal. Remote editing fail probably does, others not.

tomhughes:
The unconfigured iD one might as well be modal, or just a message embedded in the page, given we can't launch it.

The failed geocoding one could probably be done better in some other way I agree.

PR #6269, opened 2025-07-30
Fix message reuse on account terms page

Pikse:
After having replaced another use of `layouts.tou` earlier (#6166) I now notice it still has two distinct uses. It is used as a standalone header and as a link text in subsequent sentence. Some languages need different capitalization  and/or declension for these uses.

Here new message is added for use in the subsequent sentence. In English it's identical to previous `layouts.tou` message.


tomhughes:
Looks good to me, thanks.

PR #6270, opened 2025-07-31
Restrict marker scaling to search results

hlfan:
It doesn't really make sense to have every marker scale in response to being hovered over. This PR restricts this behaviour from #6235 only to the search markers.

### DIFF
--- a/app/assets/javascripts/index/search.js
+++ b/app/assets/javascripts/index/search.js
@@ -71,7 +71,7 @@ OSM.Search = function (map) {
     listItem.css("--marker-color", color);
     const anchor = $("<a>").attr("href", listItem.find("a.set_position").attr("href"));
     const data = listItem.find("a.set_position").data();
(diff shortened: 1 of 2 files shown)


danieldegroot2:
Scaling is also useful for overlapping notes. Is this still included?
It might not be useful for home location marker, unless used in combination with search/notes.
Where a marker is the only marker possible it can be removed.

hlfan:
When a note is selected, another marker is drawn for the note. Then, hovering over the note marker can give you this result:

image

To fix that, while keeping the scaling, requires quite a bit more code.

tomhughes:
Looks good to me, thanks.

lumikeiju:
Personally, I like the scale-up-on-hover behavior! It makes the page feel more responsive and less outdated.

tomhughes:
It's still there for the search results, which is the only place we intended to have it.

My original thought when it happening elsewhere was raised yesterday was that it probably does make sense for notes as well but @hlfan has explained why that needs more work and why the current behaviour causes weird effects for notes.

PR #6274, opened 2025-07-31
Add display:block to the editor iframe

jleedev:
### Description

Pressing the down arrow key will sometimes scroll the parent element of the iframe, since it generates an inline box with some whitespace below. This parent element having overflow: hidden meant that it was impossible to scroll deliberately and I would be stuck with a white gutter on the bottom of the viewport for the remainder of my editing session.

Instead of overflow: hidden, give the iframe display: block to avoid generating an inline block.

### How has this been tested?

Manually in Firefox and Chrome.

### DIFF
--- a/app/views/site/_id.html.erb
+++ b/app/views/site/_id.html.erb
@@ -1,11 +1,11 @@
 <%= javascript_include_tag "edit/id" %>
 
-<div id="map" class="h-100 overflow-hidden">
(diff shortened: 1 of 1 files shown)


tomhughes:
I don't know how to reproduce the original problem so I'll have to take your word for it that this is a fix - certainly it doesn't cause any issues I can see.

PR #6279, opened 2025-08-01
Fix typo in importer role icon tooltip.

OrichalcumCosmonaut:
### Description

I noticed that the tooltip for the importer role icon used “a” where it should be “an”, so I made a pull request to fix this.

### How has this been tested?

It hasn’t, but I assume a single-character change to a locale string wouldn’t need to be.


tomhughes:
Looks good to me, thanks.

PR #6280, opened 2025-08-01
Use more form helpers

hlfan:
Somewhere (I didn't find it on the fly), that was said to be preferred.


tomhughes:
Looks good to me, thanks.

PR #6284, opened 2025-08-03
Handle html error message for map xml export

mmd-osm:
This PR moves the HTML error response handler to download_util.js, and applies it to the Map XML export as well.

Before:

image

After:

image


app/assets/javascripts/download_util.js
```diff
@@ -52,7 +52,20 @@
         }
       };
     }
+
+    static getHtmlResponseHandler(event) {
```
hlfan: That function doesn't return a function, so I'd remove the "get" in the name.
It's also turbo specific, so that should probably be in the name.

app/assets/javascripts/download_util.js
```diff
@@ -52,7 +52,20 @@
...
+      const response = event.detail.fetchResponse.response;
+      const contentType = response.headers.get("content-type");
+
+      if (!response.ok && contentType?.includes("text/html")) {
+        // Prevent Turbo from replacing the current page with an error HTML response
+        // from the image export endpoint
```
hlfan:
```suggestion
        // from the export endpoint
```

app/assets/javascripts/download_util.js
mmd-osm: Ok, how about turboHtmlResponseHandler?

app/assets/javascripts/download_util.js
```diff
@@ -52,7 +52,20 @@
...
+        event.preventDefault();
+        event.stopPropagation();
+      }
+    };
   }
 
   OSM.getTurboBlobHandler = DownloadUtil.getTurboBlobHandler;
+  OSM.getHtmlResponseHandler = DownloadUtil.getHtmlResponseHandler;
```
hlfan: Does it have to be defined in the class, can't it be just written here?
Or even outside the IIFE since it's in the OSM namespace.

mmd-osm: I wanted to have a similar code structure for different turbo handlers. That's pretty much the only reason.

tomhughes: I was thinking much the same things when I looked at this, namely why are we defining things in a private class and them immediately exporting them to the global namespace, and why are we using an IIFE for things in the global namespace.

I guess if the class remains private the `getTurboBlobHandler` at least would have to remain in the IIFE even if was moved outside the class as it needs to reference it.

I don't think any of this is a show stopper in any case.

mmd-osm: @hlfan volunteers to do the cleanup in #6285. I’m closing this one now.

hlfan: I'd mark it as a draft in the meantime, since these PRs are targeting different things.

mmd-osm: Not an issue, we can easily apply this change later on once the dust has settled a bit. I don’t want the three of us working on the same code in three different pull requests.

PR #6285, opened 2025-08-03
Flatten DownloadUtil

hlfan:
Cleaning up a relatively unnecessary class.
See also https://github.com/openstreetmap/openstreetmap-website/pull/6284#discussion_r2249999654.

The last few commits may be debatable.

### DIFF
--- a/app/assets/javascripts/download_util.js
+++ b/app/assets/javascripts/download_util.js
@@ -17,25 +17,19 @@ OSM.showAlert = function (message) {
 };
 
 OSM.getTurboBlobHandler = function (filename) {
-  async function handleExportSuccess({ fetchResponse }) {
-    try {
-      const blob = await fetchResponse.response.blob();
-      OSM.downloadBlob(blob, filename);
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

mmd-osm:
I have to say that I'm a bit surprised to see the Promise pipelines sneaking in again, after we've been discussing the point in #6192. This feels a bit like going in circles and wasn't really necessary to flatten DownloadUtils. Next time someone modernizes the code, we'll be back at async/await anyway :)

hlfan:
> The last few commits may be debatable.

Modernize away then! There are lots of fetches to take on.

tomhughes:
> I have to say that I'm a bit surprised to see the Promise pipelines sneaking in again, after we've been discussing the point in #6192.

I probably never even read that discussion between you and @hlfan so it's not surprising I didn't consider it... Personally I have no problems with promise pipelines and they're generally considered a modern and safer way to do parallel computation.

PR #6291, opened 2025-08-05
Instantiate js controllers only in router

hlfan:
A step towards lazily instantiating these controllers.

A few event handlers (like for query and search) need to be moved to allow the controllers to really be instantiated lazily.


tomhughes:
Looks good to me, thanks.

PR #6293, opened 2025-08-05
Use <h1> as Bootstrap modal title

AntonKhorev:
As recommended in https://getbootstrap.com/docs/5.3/components/modal/#modal-components

*Select Language* modal uses `<h1 class="modal-title fs-5" ...>`. Why should the alert modal be different?


tomhughes:
Looks good to me, thanks.

PR #6294, opened 2025-08-05
Move Bootstrap modal ok string out of javascripts

AntonKhorev:
This string doesn't need to be in javascripts because we're not placing it in javascript. We don't need to send it again along with other js i18n strings.


hlfan (approved):

tomhughes:
Looks good to me, thanks.

PR #6295, opened 2025-08-05
Instantiate js controllers only once needed

hlfan:
Second part of #6291.

### DIFF
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -1,3 +1,13 @@
+OSM.initializations.push(function () {
+  $(".control-note .control-button").on("click", function (e) {
+    e.preventDefault();
+    e.stopPropagation();
+
+    if ($(this).hasClass("disabled")) return;
+
+    OSM.router.route("/note/new");
+  });
+});
 OSM.NewNote = function (map) {
   const noteLayer = map.noteLayer,
         content = $("#sidebar_content"),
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/index/new_note.js
```diff
@@ -7,15 +7,6 @@ OSM.NewNote = function (map) {
...
       halo;
 
-  addNoteButton.on("click", function (e) {
-    e.preventDefault();
-    e.stopPropagation();
-
-    if ($(this).hasClass("disabled")) return;
-
-    OSM.router.route("/note/new");
-  });
```
AntonKhorev, line 17: Do the event handlers need to to be moved to another file? Why wouldn't adding this on the top of `new_note.js` work?

```js
$(function () {
  $(".control-note .control-button").on("click", function (e) {
    e.preventDefault();
    e.stopPropagation();

    if ($(this).hasClass("disabled")) return;

    OSM.router.route("/note/new");
  });
});
```

hlfan: Every other affected event handler interacts with the `map`, which wouldn't be available in an IIFE.
I'd like to do the same for all of these, even if it's not required for this specific example.

pablobm:
@hlfan - Would you be able to rebase this? I was having a look today, perhaps I can provide a review :slightly_smiling_face:

hlfan:
sure

app/assets/javascripts/index/new_note.js
pablobm: In the interest of separation of concerns, would it be possible to move these back into their modules, but as functions that can are called from `initializeMapButtons`?

So for example, this code here could live as `OSM.NewNote.initialize()`, called from `OSM.initializeMapButtons`.

It's not trivial because then there's the question of what to do with the routes, which are functions... I played a bit with the idea at https://github.com/hlfan/openstreetmap-website/compare/lazy-load-controllers...pablobm:lazy-load-controllers. What do you think?

pablobm (approved):
I approve of this solution 👨‍⚖️

tomhughes:
Looks good to me, thanks.

Issue #2799, opened 2020-08-28
Support Sign in with Apple


hikemaniac:
The OpenStreetMap website currently allows signing in with several third party providers including Google, Microsoft, Wikipedia, GitHub, Yahoo, etc... Apple recently launched "[Sign in with Apple](https://developer.apple.com/documentation/sign_in_with_apple)" as their login framework, so it might be worth some consideration to add it as well, especially since it's seemingly gaining popularity quickly - I've already seen it adopted by many different services.

Docs:
- [Sign in with Apple JS](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js)
- [Sign in with Apple REST API](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api)

As it's apparently necessary to be enrolled in the Apple Developer Program to use the software :confused:, I'm not even sure if we can use their framework. I just wanted to put this forward as a heads-up so feel free to dismiss this if it's not worth the implementation effort or we can't comply with their conditions.

tomhughes:
Before anybody thinks about working on this LWG need to decide if the terms are acceptable - we got bitten by that with twitter where it was implemented and then couldn't be deployed.

tomhughes:
Good news is that here is an [omniauth plugin](https://github.com/nhosoya/omniauth-apple), which is what actually matters rather than technical documentation.

migurski:
Rambly thought about 3rd-party auth: people often have accounts with several of the offered providers and it’s not always easy to remember which one you used to link an account. I have been confusedly locked-out or forgotten by sites like Stackoverflow in the past because I wasn't sure which provider I’d originally chosen, and I have heard from others that this is a common problem. There’s not often a good way to remember which unrelated 3rd party’s logo you clicked the last time you logged in to something.

Suggestion: can we set a long-lasting cookie that remembers the chosen provider? Here’s a basic picture of what that might look like:

![Artboard](https://user-images.githubusercontent.com/58730/91605904-13e49d00-e926-11ea-9bdc-9ca0d780c2f5.png)

danieldegroot2:
> Suggestion: can we set a long-lasting cookie that remembers the chosen provider?

@migurski This is off-topic for this issue. Open a new issue, in order for it to be tracked.

HarelM:
Hi, any updates on this?
I'm using OSM in order to allow users to login to my app that allows some basic editing.
[...]

PR #6298, opened 2025-08-06
Add support for Apple authentication

tomhughes:
This adds support for Sign in with Apple as requested in #2799.

The main commit is a straightforward addition of a new authentication provider along the lines of the existing ones.

The second commit is a workaround for the problem described in https://github.com/openstreetmap/openstreetmap-website/issues/2799#issuecomment-3156472926 and works by monkey patching the omniauth oauth2 provider to redirect any callback that comes in via POST to GET so that we get the session cookie, allowing us to associate the response with the ongoing login attempt.


AntonKhorev:
I can't really test it because I don't have the keys. I guess there are reasons why this line doesn't need to be changed:
https://github.com/openstreetmap/openstreetmap-website/blob/d8fff3c499653ac492b7ad82dd56fb5ba6dae359/app/controllers/application_controller.rb#L30

Another thing is if there's an existing account, and you open `/account` then change *External Authentication* to *Apple*, does it fail correctly? It redirects me to `/login` with an "Authentication failed" flash message, but I'm already logged in.

tomhughes:
So actually we do need to add `appleid.apple.com` to that list but it's only needed for the signup page and the account edit page and lot for the login page - it's also only needed by chrome and not by firefox.

What I don't understand is why signup needs it and login doesn't as neither uses a form post for the button - account edit does redirect from a form post so is expected.

Needless to say although I tested with both firefox and chrome I did so from the login page and just let it create a new account that way...

tomhughes:
> Another thing is if there's an existing account, and you open `/account` then change _External Authentication_ to _Apple_, does it fail correctly?

It fails as well as it does for any other authentication provider, which is to say not very well... You get sent back to the form with no provider selected and no error shown.

The reason is that the `has already been taken` error has been recorded against the `auth_uid` field but that isn't shown so the error is never rendered.

PR #6299, opened 2025-08-06
Scrollable pagination for changeset elements

AntonKhorev:
We have scrollable pagination for element versions:
image

This PR makes changeset elements navigation also scrollable. This makes it much faster to get to page # anything larger than 3.

Before:
image

After:
image

The last step is to remove the classic pagination library that is not used anywhere after other changes in this PR.


tomhughes:
Looks good to me, thanks.

PR #6271, opened 2025-07-31
Move more styling from CSS to Bootstrap

hlfan:
Continuing the effort to slim down common.scss

PR #6302, opened 2025-08-07
Fix erroneously collapsed CSS rule

### LINKED ISSUE #6301: osm unusable on mobile because of banner
### URL

https://www.openstreetmap.org

got to the openstreetmap website on a phone

there's no way to make these  messages go away. I can get glimpses of the map when I use search, but there's no way to use the full designated area. there are always the "advertisements" or whatever in the way and no way to close them.

(Fennec 141)

Image

hlfan:
Closes #6301 by restoring the behaviour prior to #6271.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -187,7 +187,7 @@ body.small-nav {
   .overlay-sidebar #sidebar {
     .welcome,
     #banner {
-      display: block;
+      display: none;
     }
   }
 }


AntonKhorev:
You're fixing a regression from #6271. Banners weren't shown in small layouts prior to that.

I don't see anywhere in the description of that pull request that we changed our mind about showing banners. Was it intended? Are there any other undeclared changes?

AntonKhorev:
So I guess it wasn't intended to start showing banners in small layouts. Then the next questions are why aren't we showing them and should we start? Maybe we're not showing them because a banner in the corner of the screen vs a banner that takes up most of the screen are different levels of annoyance. Banners on small screens will more likely get dismissed without being read. Maybe we need a different format for small banners.

hlfan:
So, when I was collapsing the `.welcome` and `#banner` rules I missed that the declaration of the new copy-pasted rule was different. Should've been caught in review, hence note to self to make commits even smaller.

As for mobile banner formats, maybe only something text-based, notice-like could be appropriate.

tomhughes:
Speaking as the review of #6271 the declared purpose of that PR was moving styling from CSS to bootstrap classes rather than any optimisation of the CSS itself.

It was also essentially impossible to do a detailed code review of it, or it would have taken an unreasonable amount of time for me to go and read up on what each bootstrap class that was added did and correlate that to removed CSS and vice versa.

So review was largely based on a visual inspection of the site with the PR merged to check that there were no obvious issues in the areas of the site affected by the changes, so mostly I was concerned with the header and the search bar and so on. I probably only looked at a normal browser and not a mobile size though. Also I had probably already cancelled any banners on my browser so wouldn't have seen them again.

hlfan:
> rather than any optimisation of the CSS itself

"You're also moving CSS around, would you mind splitting that into a separate commit" would have been perfectly acceptable too.

PR #6304, opened 2025-08-08
Scrollable pagination javascript module

AntonKhorev:
Part 1 of #6299.

The idea is to move scrollable pagination out of element page javascript and make it later applicable to any numbered pagination, particularly changeset elements pagination.

This PR pulls out pagination code from element js and puts it into its own file. There should be no visible differences.

image


app/assets/javascripts/numbered_pagination.js
```diff
@@ -0,0 +1,69 @@
...
+    shadowEffect?.disable();
+    shadowEffect = null;
+  });
+
+  $(document).on("numbered_pagination:center", function () {
+    const [scrollableList] = $("#versions-navigation-list-middle");
+
+    if (!scrollableList) return;
+
+    const [activeStartItem] = $("#versions-navigation-list-start #versions-navigation-active-page-item");
```
HolgerJeromin, line 58: As you are working on a DOM element (and not jquery collections) I would go with this

```suggestion
    const activeStartItem = document.querySelector("#versions-navigation-list-start #versions-navigation-active-page-item");
```

AntonKhorev: And I would go with what is in this pull request because it's an intermediate step consisting mostly of moving code around with minimal changes.

HolgerJeromin: I did not saw the source of your move (reviewed on mobile) so I thought this code was new.

app/assets/javascripts/index/element.js
```diff
@@ -5,7 +5,7 @@
   const isOfExpectedLanguage = ({ language }) => languagesToRequest[0].startsWith(language) || language === "mul";
 
   $(document).on("click", "a[href='#versions-navigation-active-page-item']", function (e) {
-    scrollToActiveVersion();
+    $(document).trigger("numbered_pagination:focus");
```
tomhughes: Should that be `:center` as there doesn't seem to be a `:focus` listener?

AntonKhorev: Yes. I renamed it, but this one got lost in an intermediate version.

Updated.

tomhughes:

Issue #5758, opened 2025-03-07
Show new/modified/deleted count in place of "<edits> 🖊" in changeset card


andrewharvey:
### Problem

The current "<edits> 🖊" count in the changesets cards makes it hard to know exactly what type of edits were made.

### Description

In OSMCha it shows added/modified/deleted feature counts as ![Image](https://github.com/user-attachments/assets/53b9caaf-3a0e-4db0-9012-9b0791304042) which is very clear, it makes it easy to scan through a list of changesets for example to spot only new features, vs deletions or modifications. I suggest to change  "<edits> 🖊" count to something like OSMCha with <new> | <modified> | <deleted>

### Screenshots

![Image](https://github.com/user-attachments/assets/7c5d1700-7dba-421a-a2fe-a48970fe0d09)

tomhughes:
I think it would be very hard to fit in that much information in the space available - using colour would probably make what is currently a very subtle indication into something much more garish that wouldn't really fit with the surrounding design.

mmd-osm:
I suppose this will be rather slow, because we don’t store this information on changeset level.

AntonKhorev:
We can start storing these counts for new changesets.

mmd-osm:
I’m sure you’d need to backfill this information for all changesets, if you don’t like users complaining about inconsistent numbers in the future.

AntonKhorev:
We can backfill it as slow as we need and tell the complaining users to use osmcha during that time.

maro-21:
> In OSMCha it shows added/modified/deleted feature counts as ![Image](https://github.com/user-attachments/assets/53b9caaf-3a0e-4db0-9012-9b0791304042)

If OSMCha has it, why copy the same thing to osm.org? If someone needs accurate data, that's what OSMCha is for.

andrewharvey:
> If OSMCha has it, why copy the same thing to osm.org? If someone needs accurate data, that's what OSMCha is for.

Better to have this by default on OSM.org in my opinion. Fragmenting into many different tools makes the learning curve harder and many people won't discover those tools. Then you had the add challenge of needing to copy paste changeset IDs all the time because all you have is a link to osm.org.

Currently OSMCha is very painful to use, 15 seconds load time for each changeset to review, 10 seconds to load the list of changesets, if they have a useful feature then that's more the reason to add it back into OSM.org.

mmd-osm:
[...]

PR #6305, opened 2025-08-08
Changeset num_changed_(element type) methods

AntonKhorev:
Optional part of https://github.com/openstreetmap/openstreetmap-website/pull/6299.

That PR deals with pages of elements changed in a given changeset. We need to know the total number of elements and derive the total number of pages. Currently we're just counting them all with sql `COUNT(*)`. But we don't need to do it for new changesets because we have per element statistics. See https://github.com/openstreetmap/openstreetmap-website/issues/5758#issuecomment-2781054236.

This PR adds some changeset model methods for easier access to number of changed nodes/ways/relations.

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -241,6 +241,10 @@ class Changeset < ApplicationRecord
   def num_modified_elements = num_modified_nodes + num_modified_ways + num_modified_relations
   def num_deleted_elements = num_deleted_nodes + num_deleted_ways + num_deleted_relations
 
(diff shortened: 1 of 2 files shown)


tomhughes:
I was going to say that we should just get on with backfilling the statistics rather than adding new and potentially slow methods that fallback to counting the records but I see what you're saying that this is just for cases where we're already doing that count anyway and this just makes it more efficient when we can.

So I think this is good, thanks. Obviously we should still get on with the backfilling then we can get rid of the fallback!

PR #6312, opened 2025-08-10
Fix "Log in with Apple" title

AntonKhorev:
https://github.com/openstreetmap/openstreetmap-website/commit/fc5f0f2ec064710caba8aa2aa693e0a515f97690#r163725641


tomhughes:
Thanks... That the danger of reviving a three year old PR, that something subtle has been change since it was created that you don't notice ;-)

PR #6314, opened 2025-08-10
Generic scrollable numbered pagination

AntonKhorev:
Part 2 of https://github.com/openstreetmap/openstreetmap-website/pull/6299.

This PR makes the numbered pagination module generic by removing removing all references to element versions. It also allows to have multiple scrollable pagination controls on a single webpage. This is what we need to reuse this module for changeset elements in the next part.


tomhughes:
Looks good to me, thanks.

PR #6316, opened 2025-08-12
Enable scrollable pagination for changeset elements

AntonKhorev:
Part 3 of https://github.com/openstreetmap/openstreetmap-website/pull/6299.

### DIFF
--- a/app/controllers/changesets_controller.rb
+++ b/app/controllers/changesets_controller.rb
@@ -157,7 +157,7 @@ class ChangesetsController < ApplicationController
   end
 
   def load_nodes
-    @nodes_count = @changeset.old_nodes.count
+    @nodes_count = @changeset.actual_num_changed_nodes
     @current_node_page = params[:node_page].to_i.clamp(1, element_pages_count(@nodes_count))
     @nodes = @changeset.old_nodes
                        .order(:node_id, :version)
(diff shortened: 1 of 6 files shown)


AntonKhorev:
Coverage decrease is because classic pagination is not in use anymore. I didn't remove it in this PR to avoid complaints from Danger, but instead I got complaints from Coveralls.

hlfan:
Since the classic pagination removal is the only difference, I'd merge #6299 instead.

app/controllers/changesets_controller.rb
```diff
@@ -153,4 +155,41 @@ def wrap_lon(lon)
...
+  def load_relations
+    @relations_count = @changeset.actual_num_changed_relations
+    @current_relation_page = params[:relation_page].to_i.clamp(1, element_pages_count(@relations_count))
+    @relations = @changeset.old_relations
+                           .order(:relation_id, :version)
+                           .offset(ELEMENTS_PER_PAGE * (@current_relation_page - 1))
+                           .limit(ELEMENTS_PER_PAGE)
+  end
+
+  helper_method def element_pages_count(elements_count)
```
tomhughes, line 186: Did you see the way of using `helper_method` somewhere? I can guess why it works, but it's not how the rails documentation shows it being used or how we use it elsewhere... We normally declare the method as normal and then add `helper_method :method_name` to make it a helper.

AntonKhorev: `def` returns a symbol. Rails' documentation of `helper_method` [predates](https://github.com/rails/rails/commit/e976c489e6416cdc4714721df78dd43dd6d13d99) [that](https://stackoverflow.com/a/26216164).

Do you want an extra line of code for `helper_method` separate from the method itself?

tomhughes: Oh that's interesting about `def` and something I didn't know! I guess there's no reason not to do it this way.

tomhughes:
Looks good to me, thanks, and I've checked the coveralls results and as you say it's only redundant code that is no longer being tested.

PR #6318, opened 2025-08-12
Update rails-dom-testing to 2.3.0

tomhughes:
I've been holding back updates to `rails-dom-testing` for a while now because it was breaking some tests so this takes the updates and fixes the test problems.


test/helpers/changesets_helper_test.rb
```diff
@@ -32,7 +32,7 @@ def test_changeset_details
     changeset_details_dom = Rails::Dom::Testing.html_document_fragment.parse "<div>#{changeset_details(changeset)}</div>"
     assert_dom changeset_details_dom, ":root", :text => /^Closed .* by #{changeset.user.display_name}$/ do
       assert_dom "> time", :count => 1 do
-        assert_dom "> @title", "Created: Mon, 01 Jan 2007 00:00:00 +0000\nClosed: Tue, 02 Jan 2007 00:00:00 +0000"
+        assert_dom "> @title", "Created: Mon, 01 Jan 2007 00:00:00 +0000 Closed: Tue, 02 Jan 2007 00:00:00 +0000"
```
AntonKhorev: Don't we want to test for newline here?

```suggestion
        assert_dom "> @title", :html => "Created: Mon, 01 Jan 2007 00:00:00 +0000\nClosed: Tue, 02 Jan 2007 00:00:00 +0000"
```

This works in both versions of rails-dom-testing.

test/controllers/user_blocks_controller_test.rb
```diff
@@ -421,7 +421,7 @@ def test_create
     post user_blocks_path
     assert_response :not_found
     assert_template "users/no_such_user"
-    assert_select "h1", "The user  does not exist"
+    assert_select "h1", "The user does not exist"
```
AntonKhorev: This breaks because the message itself doesn't make sense, if we take i18n into account. There's an empty username with spaces around it. Temporarily I'd do `:html` fix like in [changesets_helper_test](https://github.com/openstreetmap/openstreetmap-website/pull/6318/files#r2273910522) because it allows commits to be in a different order.

AntonKhorev:
I'd first apply the fixes and after that update rails-dom-testing. That way there won't be any test errors right after the rails-dom-testing commit, in case we'll need to bisect later.

test/helpers/changesets_helper_test.rb
tomhughes: Change applied.

test/controllers/user_blocks_controller_test.rb
tomhughes: Done.

tomhughes:
I've reordered the commits as suggested.

AntonKhorev:
Merged, thanks.

PR #6320, opened 2025-08-13
Antimeridian map data display

mmd-osm:
This PR enables map data display across the antimeridian. It splits up the /map request into two requests if needed.

Depending on the current map center,  cloned nodes and ways with +/- 360 offset are added to the list of features. This is described in more detail here: https://stackoverflow.com/a/54709437/3501889 + https://stackoverflow.com/a/38824652 by one of the main Leaflet contributors.

Before:

image

After:

image

image


tomhughes:
Looks good to me, thanks.

PR #6326, opened 2025-08-15
Lazy loading relation members

mmd-osm:
This PR introduces async loading of relation members, which is the main reason for large relations to take quite some time to load and display.

Current process:
1. Navigate to relation -> Rails renders tags, member list, returns final result to browser. Until then, a blank screen is shown.
2. Browser fetches full relation details, renders map

New process:
1. Navigate to relation -> render tags only. This step is expected to return results much quicker to the browser
2. Browser requests relation member list in separate query. Turbo updates DOM once finished.
3. Concurrent to step 2, browser requests full relation detail and starts rendering the map.

For large member lists, we're hiding the list by default anyway, so that the user doesn't notice the DOM update.  If relation member loading takes longer, a spinner is shown when expanding the member list.


app/abilities/ability.rb
```diff
@@ -6,6 +6,8 @@ class Ability
   def initialize(user)
     can :read, [:feature_query, :layers_pane, :legend_pane, :share_pane]
     can :read, [Node, Way, Relation, OldNode, OldWay, OldRelation]
+    can :read, RelationMember
+    can :read, OldRelationMember
```
AntonKhorev: Are those actually "can read relation members if there's more than ten members"?

mmd-osm: Not necessarily. The two new controllers can handle any number of members. It’s only up to app/views/browse/_relation.html.erb decide when to use them. This could be changed without touching the controllers.

mmd-osm: By the way, one of the two new controllers was missing previously in the commit. Now added + some unit tests.

config/routes.rb
```diff
@@ -142,9 +142,13 @@
   get "/node/:id" => "nodes#show", :id => /\d+/, :as => :node
   get "/node/:id/history" => "old_nodes#index", :id => /\d+/, :as => :node_history
   resources :old_nodes, :path => "/node/:id/history", :id => /\d+/, :version => /\d+/, :param => :version, :only => :show
+
   get "/relation/:id" => "relations#show", :id => /\d+/, :as => :relation
+  get "/relation/:id/member" => "relation_members#show", :id => /\d+/, :as => :relation_member
```
tomhughes: Should this (and the other route) be `/members` given that they load all members not just on?

mmd-osm: make sense, fixed.

mmd-osm:
+ added redaction handling

app/views/browse/_relation_member_frame.html.erb
```diff
@@ -0,0 +1,13 @@
+<turbo-frame id="<%= frame_id %>">
+  <% if relation.redacted? && !params[:show_redactions] %>
```
tomhughes: Can this ever happen? If it's redacted and we're not showing redactions then we won't have generated the frame will we?

I guess there's an issue that somebody could request the frame manually to try and see the redactions but the controller could just reject requests for a frame for a redacted relation unless the show redactions flag is set?

mmd-osm: The error message was intended for someone trying to directly access the new endpoint. It cannot happen in our UI, because, as you said, the relation version won't show any details except for this message: _Version x of this relation cannot be shown as it has been redacted. Please see Redaction y for details._

So in the OldRelationMembersController, I would have to add something like:

```
return deny_access(nil) if @feature.redacted? && !params[:show_redactions]
```

mmd-osm:
Originally, I used the `loading="lazy"` attribute to fetch relation members on demand only, when a user expands the list (that's where the lazy loading in the commit message comes from).

With the current implementation, all relation members are prefetched. Maybe this needs to be tuned for some use cases like the relation history, to limit the number of requests. Or it could be used by default, once the relation member tag prefetch is in place (see below discussion).

```erb
          <turbo-frame id="member_relation_<%= relation.id %>"
            <%# Render relation history lazily %>
            <% if local_assigns[:relation_counter] %>
              loading="lazy"
            <% end %>
```

mmd-osm:
Furthermore, prefetching member tags could be useful to avoid lots of small queries that are currently slowing down the process. The following could be called from the two new controllers. I've added that now as separate commit, that could  be moved to another PR if needed.

```ruby
  def prefetch_member_tags(feature)
    members = feature.relation_members.map(&:member).compact
    grouped_members = members.group_by(&:class)

    {
      Node => :node_tags,
      Way => :way_tags,
      Relation => :relation_tags
    }.each do |klass, tag_assoc|
      records = grouped_members[klass] || []
      ActiveRecord::Associations::Preloader.new(
        :records => records,
        :associations => tag_assoc
      ).call
    end
  end
```

tomhughes:
I was already planning to look at prefetching the tags once this is merged but I don't know what it would need anything that complicated - will a normal rails preload not work?

tomhughes:
Presumably the issue is the polymorphic relationship - does rails not handle chasing preloads across that?

mmd-osm:
Agree, it looks a bit ugly. It's more of a first attempt to see if prefetching has some benefits. The challenge seems to be that tag associations are all named differently across the three models.

mmd-osm:
An alias_association might be an option, but that feature hasn't landed yet: https://discuss.rubyonrails.org/t/feature-proposal-alias-association/83439/9 and https://github.com/rails/rails/pull/49801. The former `alias_attribute` seems to have been deprecated/removed.

tomhughes:
So the issue is that the different models have different names for the tag association and if they were all the same then the normal methods would work?

In which case I'd suggest we should normalise them - that is a little complicated as `tags` is already in use as a separate method on the models...

mmd-osm:
Ok, this sounds like a bigger task which won't fit in this PR. I have removed the prefetch commit now and kept a copy of it over at: https://github.com/mmd-osm/openstreetmap-website/tree/prefetch_relation_member

mmd-osm:
> So the issue is that the different models have different names for the tag association and if they were all the same then the normal methods would work?

I'm a bit lost how the normalisation is supposed to work - without changing anything on the db schema. I tried to use   `has_many :element_tags, :as => :relation_tags`, but PostgreSQL was complaining about some unknown columns/types.

In the meantime, the previous code is now a bit more generic, and derives associations from class names. I think that's not too bad. ([link](https://github.com/mmd-osm/openstreetmap-website/tree/prefetch_relation_member))

```ruby
  def prefetch_member_tags(feature)
    members = feature.relation_members.filter_map(&:member)

    members.group_by(&:class).each do |klass, records|
      association = :"#{klass.name.downcase}_tags"

      ActiveRecord::Associations::Preloader.new(
        :records => records,
        :associations => association
      ).call
    end
  end
```

tomhughes:
So I tried adding associations to node/way/relation along these lines:

```ruby
has_many :element_tags, :class_name => "NodeTag"
```

and with all three in place I could then do:

```ruby
Relation.preload(:relation_tags, :relation_members => { :member => :element_tags }).find(2)
```

and it would preload everything.

So as I suspected we just need to rename the associations to have a common name and it should all work.

Note that the old models already have a common name (`old_tags`) so this should already work there.

mmd-osm:
That's interesting. At one point I tried class_name along with `:as => ... `which didn't work. So ":as" was really causing some major issues.

To make tests pass again, I had to change about 32 lines across 18 files - that's less than I expected: https://github.com/mmd-osm/openstreetmap-website/commit/7fb9a23db1819a3fc775703b71592a032705acd1

tomhughes:
I think this is good to merge as it is - we can work on the tag thing separately given it's something we don't do now. Thanks for all the work on it.

mmd-osm:
~I've moved the prefetching to a new PR #6337.~ --> superseded by #6338.

Issue #6325, opened 2025-08-15
impossible to log in with new created user


xronoc:
### How to reproduce the issue?

I've created user in web interface, activated with user = User.find_by(:display_name => "bernys"), granted it administrator and moderator roles.
login page http://localhost:3000 gives
NoMethodError in SiteController#index

### Screenshot(s) or anything else?

![Image](https://github.com/user-attachments/assets/157ac08e-d25a-465e-a9ed-868adc19b54f)

tomhughes:
What efforts have you made to debug thus? Did you discover anything?

tomhughes:
You haven't configured the OAuth keys, have you?

See https://github.com/openstreetmap/openstreetmap-website/blob/master/CONFIGURE.md#oauth-setup for what you need to do, but read the rest of that file as well!

xronoc:
> You haven't configured the OAuth keys, have you?
> 
> See https://github.com/openstreetmap/openstreetmap-website/blob/master/CONFIGURE.md#oauth-setup for what you need to do, but read the rest of that file as well!

yes, i haven't! But the instruction solves problem for first user only. What to do to add second one?

tomhughes:
That isn't a per-user thing, it's adding one set of applications for the whole installation.

If you're getting that error then you don't have `oauth_application`/`oauth_key` set correctly in your settings.

xronoc:
i figure it. now it works. thanks!

PR #6327, opened 2025-08-16
Add a note that you only need to register iD and website once

AntonKhorev:
To clear up this: https://github.com/openstreetmap/openstreetmap-website/issues/6325#issuecomment-3192253190

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -100,6 +100,8 @@ There are two built-in applications which communicate via the API, and therefore
 * **iD** - the web-based editor
 * **The website itself** - for the Notes functionality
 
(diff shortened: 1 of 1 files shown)


CONFIGURE.md
```diff
@@ -100,6 +100,8 @@ There are two built-in applications which communicate via the API, and therefore
 * **iD** - the web-based editor
 * **The website itself** - for the Notes functionality
 
+You need to register these applications with *one* of the users you created. After that iD and Notes functionality become available to every user of the website.
```
tomhughes, line 103:
```suggestion
You need to register these applications with *one* of the users you created. After that iD and Notes functionality becomes available to every user of the website.
```

tomhughes:
Merged with the minor edit I proposed added.

Issue #5264, opened 2024-10-16
Add devcontainer configuration


gravitystorm:
We added support for Docker-based development in [2021](https://github.com/openstreetmap/openstreetmap-website/pull/2409) (based on a PR started in 2019). At this time there were no clear conventions for setting up docker-based development environments, so we did our own thing.

Nowadays, a number of tools support the "[Development Container](https://containers.dev)" specification. This provides a few additional benefits, for example tighter integration between code editors and the containers (e.g. running the rubocop process within the container, but seeing the syntax highlighting in your editor UI). Dev containers are [supported](https://containers.dev/supporting) by vscode and intellij code editors, and are also a part of Github Codespaces. Codespaces allow developers to make code changes, run tests and make PRs, all without installing anything locally. So it would be nice to fully support that too.

It's also notable that rails 7.2 added built-in support for devcontainers configuration, so running `bundle exec rails devcontainer` gets us started on the setup and generating the configuration files.

I therefore propose:
* We set up the devcontainer configuration and make sure it works.
* We move the docker part of CI to use the devcontainer configuration.
* We remove the original Docker configuration and update the documentation.
* We also consider removing the vagrant configuration.

mmd-osm:
Thanks, this looks like a great suggestion. I gave it a try, created a new clone of the repo, then added a new devcontainer using `bundle exec rails container.`

Initially I was a bit struggling with (Debian, Rails, npm) packages on generated devcontainer. I believe we can probably reuse a lot of the existing Dockerfile and copy it over to .devcontainer/Dockerfile, thereby hopefully addressing these issues. There's a bit of overlap between the "postCreateCommand" in devcontainer.json and the existing Dockerfile, which would need to be resolved.

Then I added memcached and mailpit to compose.yaml and adjusted settings.local.yml accordingly. Both were fairly trivial to get up and running. They allow me to run more production-like scenarios without much hassle. Also, unit tests are passing successfully.

Overall I see a much improved DX compared to the old Dockerfile.

<details>
<summary>Sample screenshot</summary>

![image](https://github.com/user-attachments/assets/d5769107-8b75-4d51-b355-f095b531ad50)
</details>

I've uploaded the config files supporting this little experiment to: https://gist.github.com/mmd-osm/7f3c871a17d496475e7c10e054ba921c

---

(_the following has been addressed to some extent in other PRs already_)

In addition, I highly recommend to include a set up script for the dev container, which:
* Creates admin + testuser
[...]

PR #6333, opened 2025-08-18
Add and unignore `.ruby-version` file to ensure consistent dev Ruby

larouxn:
### Description

Adds and unignores `.ruby-version` file to ensure a consistent development Ruby for all contributors. Doing so also helps push contributors onto newer rubies which can sometimes be quite helpful, such as with the latest [Ruby 3.4.5](https://www.ruby-lang.org/en/news/2025/07/15/ruby-3-4-5-released/) which addressed a ton of GCC 15 gem compilation error issues.

Additionally, having a `.ruby-version` file has become a somewhat standard practice in Ruby development such that the [chruby CLI tool](https://github.com/postmodern/chruby) and the [ruby-lsp VSCode extension](https://marketplace.visualstudio.com/items?itemName=Shopify.ruby-lsp) expect it to exist. There's even a proposal to spec the `.ruby-version` file format over in rubygems: https://github.com/orgs/rubygems/discussions/7074.

**Why is `.ruby-version` currently ignored?**

Looks like it may have been arbitrarily ignored as a side change in https://github.com/openstreetmap/openstreetmap-website/commit/14ac1babc2517320a2c90fa9b4ac36a5a6e68018 over 10 years ago.

### How has this been tested?

Yes it has been tested. See the screencasts below.

**Before** (`master` branch)

https://github.com/user-attachments/assets/55d7605f-c4ce-4914-92ea-8c9ac0319375

**After** (this branch)

https://github.com/user-attachments/assets/cefcc1e8-4156-4d1a-bde8-f427646346f4

Can also see that the later shell (above screencast i.e. this branch) properly uses Ruby 3.4.5 immediately by default thanks to the `.ruby-version`, which is highly preferable.


tomhughes:
I mean it only "ensures a consistent development Ruby" for contributors that pay attention to it in some way... I'm not opposed to it but we probably need to consider what value it should contain.

larouxn:
That's fair. If a contributor doesn't use `chruby`, `rbenv`, `rvm`, or `ruby-lsp` (may be others) this won't do anything for them. That said, `chruby` + `ruby-install`, `rbenv`, and `rvm` are most of the most popular Ruby development setups so I think a decent number of contributors may benefit from having a checked-in `.ruby-version`.

> After rbenv injects itself into your PATH at installation time, any invocation of `ruby`, `gem`, `bundler`, or other Ruby-related executable will first activate rbenv. Then, rbenv scans the current project directory for a file named `.ruby-version`. If found, that file determines the version of Ruby that should be used within that directory. Finally, rbenv looks up that Ruby version among those installed under `~/.rbenv/versions/`.

Source: [rbenv's README](https://github.com/rbenv/rbenv/blob/d1a19a398f78ec8e94a5ad251aaf8d241053fdbe/README.md)

> Additionally you might want to list your preferred ruby version in a `.ruby-version` file stored in your project folder. This would cause automatic switch to selected ruby whenever you enter the folder.

Source: [rvm's README](https://github.com/rvm/rvm/blob/551ef582307ae524ec9c2999fc0abf820c286dde/README.md)

pablobm:
The Dockerfile currently pulls `ruby:3.2-bookworm`, and `INSTALL.md` lists "Ruby 3.2+". If a version is chosen, it should be consistent across all instances.

I'm generally in favour of standardising on one version. Subtle changes to Ruby can cause different behaviours across different setups, same as with versions of gems. I don't think it necessarily has to be the latest available (3.4.5 in this case), but instead I'd lean towards whatever is used on production, precisely for the sake of consistency.

larouxn:
> I'd lean towards whatever is used on production, precisely for the sake of consistency.

That makes sense to me. Apologies, I hadn't yet checked what version is in production. Seemingly it's not Ruby 3.4.5, I hope it's [3.2.9](https://www.ruby-lang.org/en/news/2025/07/24/ruby-3-2-9-released/) if on 3.2.x, though maybe we can upgrade to 3.3.x and then 3.4.5 some time soon. :)

tomhughes:
I've never used any of those tools so I don't really have any useful comments/experience to add there.

Production is actually running 3.4.5 currently, which is actually slightly ahead of the 3.4.4 that I'm using for development.

larouxn:
> Production is actually running 3.4.5 currently, which is actually slightly ahead of the 3.4.4 that I'm using for development.

I'm super glad to hear that. Ruby 3.4's been great for both development and production performance.

I hear that as I think we're both on Fedora where Ruby 3.4.4 is the current bundled version. Hoping they bump that soon, was looking into such today.

_Anddd_ I suppose then that the version in the `.ruby-version` in this PR is accurate to production then. :)

AntonKhorev:
> Adds and unignores .ruby-version file to ensure a consistent development Ruby for all contributors.

Why do we need a consistent Ruby if we're supporting a range of versions?

> Doing so also helps push contributors onto newer rubies which can sometimes be quite helpful, such as with the latest [Ruby 3.4.5](https://www.ruby-lang.org/en/news/2025/07/15/ruby-3-4-5-released/) which addressed a ton of GCC 15 gem compilation error issues.

Why do I need to get pushed onto this version when I'm writing 3.2-compatible code?

larouxn:
> Why do we need a consistent Ruby if we're supporting a range of versions?

For sure, it makes sense we can still support Ruby 3.3 and 3.2 as they're still supported. At the same time I feel it makes most sense to have all contributors and production be on the same version of Ruby for maximum consistency and predictability between development and production.

> Why do I need to get pushed onto this version when I'm writing 3.2-compatible code?

Shouldn't we prefer people write code that adheres to the latest version of Ruby both to prevent sooner need for maintenance given production runs latest Ruby? Ruby 3.2 is EOL in 7 months and I suppose soon after it's EOL it will no longer be supported in this repo. Meanwhile 3.4 has +2.5 years left. ([source](https://www.ruby-lang.org/en/downloads/branches/))

A bit contrived _but_ for example, when Ruby 4 comes out and strings are frozen and production is upgraded to Ruby 4, assuming we don't merge https://github.com/openstreetmap/openstreetmap-website/pull/5932 then writing valid Ruby 3.3/3.4 code that modifies strings would break CI and production.

---

Glad to close this PR if it's not wanted. All good. :bowing_man:

pablobm:
Thank you all for the interesting conversation. Am I correct in understanding the following?
- The main OpenStreetMap web app runs Ruby 3.4.5.
- This is an FOSS app that anyone can install and run on their own, and for this reason OSM wants to support a certain infrastructure flexibility, in this case with Ruby versions.

If the above is correct, then having people developing under different rubies can be useful. ~CI runs several versions, but test coverage is not great so things can slip.~ (Correction after double-checking and seeing it's pretty good). CI running several versions also helps.

Having said that, I do feel a bit weird about the misalignment between development and production. Although Ruby is relatively stable, incompatibilities do happen. Having `.ruby-version` will not affect your development if you don't use any of these version management tools. However, if you do use them and you prefer developing with a different Ruby version, then perhaps there's a problem: the tool may get in the way and push you to use the given version (with some it can be configured away, but it's still a friction).

As for the choice of minimum Ruby version, am I correct that the supported versions of Ruby depend on the Rails version? Rails 8.0 supports Ruby 3.2+, so I'm guessing the support grid will change when upgrading to a new Rails in the future.

tomhughes:
Until a few months ago production was running with the system ruby, which for Debian 12 was 3.1 which was becoming increasingly problematic as it was approaching EOL and gems were dropping support for it.

So we switched to using https://github.com/fullstaq-ruby/server-edition and once we did that there was little reason not to use the latest version.

While the openstreetmap.org deployment is our primary use case and tends to take priority over other things we try not to deliberately break third party usage unnecessarily so we still test on 3.2 and 3.3 as well for now though I suspect we may have to drop 3.2 in the not too distant future.

larouxn:
> So we switched to using https://github.com/fullstaq-ruby/server-edition and once we did that there was little reason not to use the latest version.

TIL this is a thing. Neat! :+1: 

Am I understanding correctly that production is running Ruby 3.4.5 but development for which Docker is encouraged is using Ruby 3.2 re: below?

https://github.com/openstreetmap/openstreetmap-website/blob/d85d1309c1ed1f14b5ee300ef0b182c725940d35/Dockerfile#L1

I understand supporting non-EOL Rubies (3.2, 3.3, 3.4) but should this be bumped to [3.4-bookerworm](https://hub.docker.com/layers/library/ruby/3.4-bookworm/images/sha256-1390a133776f76f63d3e99d9c45b5f3353b1af41044024ff8098a5cd993c97cb)? That would actually better accomplish what this PR was trying to do since it synchronizes development and production Ruby versions.

Firefishy:
Just as a note, [fullstaq-ruby/server-edition](https://github.com/fullstaq-ruby/server-edition) is not yet available for ARM64. I am slowly behind the scenes working on a PR to add ARM64 support there.

tomhughes:
I can't really speak to what's best for the docker stuff as I don't really use it...

Firefishy:
My humble recommendation for the container would be to stick to the [Official Ruby upstream containers](https://hub.docker.com/_/ruby). 

Either automatic upgrades of minor version:
```
FROM ruby:3.4-bookworm
```

Or exact version pinned:
```
FROM ruby:3.4.5-bookworm
```

UPDATE:[ Docker Hub has annoying usage limits](https://docs.docker.com/docker-hub/usage/pulls/), there are official images at GHCR, but they are not as widely used:
eg:
```
FROM ghcr.io/ruby/ruby:3.4
FROM ghcr.io/ruby/ruby:3.4.5
```

larouxn:
> FROM ghcr.io/ruby/ruby:3.4.5

My vote would be to update the `Dockerfile` to source from the above GitHub source Ruby 3.4.5 image. :+1: 

While we're talking about Docker and dev env and all, would people be open to adding a `.devcontainer` setup, like [rails/rails](https://github.com/rails/rails/tree/41f4e981ded1c0d8c2b3bc5bff5fb2af38da04ed/.devcontainer) has, to this repo? Doing so would likely make dev env setup for new contributors much much simpler, quicker, and less (human) error prone than the current somewhat manual process.

mmd-osm:
>be open to adding a .devcontainer setup

We have an  issue on that topic already: #5264 ... maybe you could follow up there?

larouxn:
:wave: Hello again. Considering development for `openstreetmap-website` is primarily done with either Docker or Vagrant, potentially via `devcontainer` someday (https://github.com/openstreetmap/openstreetmap-website/issues/5264), and in any of those three a `.ruby-version` is not necessary, while nice to have I'm opting to close this PR.

PR #6334, opened 2025-08-19
Remove `file_exists` gem

rkoeze:
### Description
The file_exists gem [was originally added](https://github.com/til/openstreetmap-website/commit/c68fa8f8f5a89213aa22920edc72170093c5abbe) to support `File.exists?`, which was deprecated in Ruby 3.2 but still used by version 0.4.7 of the oauth gem. Since `File.exists?` was removed in version [0.5.0](https://github.com/oauth-xx/oauth-ruby/releases/tag/v0.5.0) of the oauth gem (see commit [here](https://github.com/ruby-oauth/oauth/pull/106/files)) and we are on 1.1.0, the dependency is no longer needed and can now be removed.

### How has this been tested?
The tests pass. I was also able to navigate the app after this change.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,9 +45,6 @@ gem "argon2"
 # Support brotli compression for assets
 gem "sprockets-exporters_pack"
 
-# Restore File.exists? for oauth gem
-gem "file_exists"
-
 # Load rails plugins
 gem "actionpack-page_caching", ">= 1.2.0"
 gem "activerecord-import"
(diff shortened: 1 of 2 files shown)


rkoeze:
Looks like I got bit by a Coveralls 503

tomhughes:
To be honest I didn't think we were using `oauth` at all now, but I see it's will pulled in by `omniauth-mediawiki` so we are. Looks good to me anyway, thanks.

PR #6336, opened 2025-08-19
Fix Apple logo aspect ratio and dark mode appearance on login page

### LINKED ISSUE #6328: Apple Logo on login page is not square
### URL

https://www.openstreetmap.org/login

Probably needs to have the width and height removed from the svg.

Image

MAHanupriSAR:
Fixed Apple logo aspect ratio by using viewbox in SVG.
Additionally, updated the logo styling so it displays correctly in dark mode, improving readability and visual consistency.  

Fixes #6328  

### Before

**Light mode:**  
before

**Dark mode:**  
beforeDark

### After

**Light mode:**  
AfterLight

**Dark mode:**  
afterDark


tomhughes:
Why all the whitespace around it and the curved corners? The other icons don't have that...

MAHanupriSAR:
> Why all the whitespace around it and the curved corners? The other icons don't have that...

Should it be transparent with no background?

tomhughes:
If you look at the other logos in that directory the image goes right up to the edges and there is nothing surrounding them?

AntonKhorev:
I'd make the images larger, some of them are already button-shaped:
image

MAHanupriSAR:
> If you look at the other logos in that directory the image goes right up to the edges and there is nothing surrounding them?

I've made the requested changes, please review again.

image

image

Dimitar5555:
Black on dark background doesn't look good. I assume that it should be white/gray in that case, but I haven't checked Apple's logo guidelines.

Firefishy:
> Black on dark background doesn't look good. I assume that it should be white/gray in that case, but I haven't checked Apple's logo guidelines.

Not exact, but [here are some guidelines](https://www.apple.com/legal/sales-support/certification/docs/logo_guidelines.pdf): 

"_Apple channel signatures can be displayed only in all-black or all-white. When placing a signature on black or color backgrounds, use an all-white signature reversed out of the background color. Never place an Apple channel signature on a visually cluttered or patterned background._"

Firefishy:
These appear to be the [specific guide for Apple ID sign in](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple).

AntonKhorev:
@MAHanupriSAR can you remove the merge commit and squash the remaining commits into one?

MAHanupriSAR:
> @MAHanupriSAR can you remove the merge commit and squash the remaining commits into one?

Did it, could you please check ?

tomhughes:
This is fine in light mode, but in dark mode it needs to be white on black as @Firefishy says but currently we get this:

image

MAHanupriSAR:
> This is fine in light mode, but in dark mode it needs to be white on black as

How the logo should appear in dark mode??
White apple logo with no background
or
Keep the logo black while adding a white background to improve visibility in dark mode?

tomhughes:
See https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple#Black for details - it should be a white logo on the existing black background.

AntonKhorev:
> See https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple#Black for details - it should be a white logo on the existing black background.

Only in light mode. In dark mode it shouldn't.

tomhughes:
Well black on black is clearly insane. I don't understand what you mean or why you've merged this but I can't deploy it as it is!

AntonKhorev:
Black on black is already deployed, from your earlier PR:
image

As it is right now, this PR fixes #6328. Which logos should be inverted (we're inverting GitHub but not others) and whether they should have extra padding is a different question.

tomhughes:
Ah I didn't realise that was already broken.

louwers:
Nice one

PR #6337, opened 2025-08-20
Enable prefetching of relation member tags

mmd-osm:
This PR enables fetching of relation member tags. Follow up for https://github.com/openstreetmap/openstreetmap-website/pull/6326#issuecomment-3202276027

As part of this change, existing associations node_tags, way_tags and relation_tags are all renamed to element_tags, so that Rails can easily prefetch all tags across different element types.


mmd-osm:
Closing in favor of #6338

tomhughes:
Oh sorry... I was going to look at merging this and then rebase mine but if you're happy to close this then that works as well.

mmd-osm:
No worries, that's totally fine. Rebasing is not worth the effort in this case.

PR #6338, opened 2025-08-20
Do more preloading of tags and members

tomhughes:
This adds some additional preloading of object tags and members for improved performance.


app/controllers/relation_members_controller.rb
```diff
@@ -1,7 +1,7 @@
 class RelationMembersController < ElementsController
   def show
     @type = "relation"
-    @feature = Relation.preload(:relation_members => :member).find(params[:id])
+    @feature = Relation.preload(:relation_members => { :member => :element_tag }).find(params[:id])
```
mmd-osm: I think there's a typo in this line. It should be element_tags with "s" at the end. Right now, prefetching isn't really working.

app/controllers/old_relation_members_controller.rb
```diff
@@ -2,7 +2,7 @@ class OldRelationMembersController < OldElementsController
   def show
     @type = "relation"
     @current_feature = Relation.find(params[:id])
-    @feature = OldRelation.preload(:old_members => :member).find([params[:id], params[:version]])
+    @feature = OldRelation.preload(:old_members => { :member => :old_tags }).find([params[:id], params[:version]])
```
mmd-osm: Prefetching isn't working here as well. In my branch, I used `element_tags` instead of `old_tags`. Members of a relation point to the current nodes/ways/relations, even for OldRelations.

app/controllers/relation_members_controller.rb
tomhughes: That should be fixed now.

app/controllers/old_relation_members_controller.rb
tomhughes: That should be fixed now.

app/controllers/api/nodes_controller.rb
```diff
@@ -30,7 +30,7 @@ def index
 
     # Dump the details on a node given in params[:id]
     def show
-      @node = Node.find(params[:id])
+      @node = Node.includes(:element_tags).find(params[:id])
```
AntonKhorev, line 34: Does this make any difference besides running the db query earlier? Or do we want to run it earlier?

tomhughes: Yes strictly speaking that one probably doesn't make any real difference. I really just did it for consistency with the other method.

AntonKhorev:
Merged, thanks.

PR #6340, opened 2025-08-21
Fix Postgres.app PATH on macOS

siyulai21:
### Description
In INSTALL.md, within the Manual Installation Guide section:
```
- export PATH=/Applications/Postgres.app/Contents/MacOS/bin:$PATH
+ export PATH=/Applications/Postgres.app/Contents/Versions/latest/bin:$PATH
```
…/Contents/MacOS/bin was the path before Postgres.app 9.2. The CLI tools now live under Contents/Versions/[version]/bin with a stable Versions/latest/bin symlink. The current path in INSTALL.md can lead to psql not being found.

### How has this been tested?
Docs change only. I verified locally on macOS by trying both paths and confirming psql only resolves from …/Versions/latest/bin.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -98,7 +98,7 @@ nano ~/.profile
 and adding:
 
 ```bash
-export PATH=/Applications/Postgres.app/Contents/MacOS/bin:$PATH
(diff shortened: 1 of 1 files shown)


tomhughes:
As I don't use or have access to MacOS I'll have to take your word for it! Thanks for the update.

Issue #6341, opened 2025-08-21
Broken grammatical gender syntax in translated interface strings


1ec5:
I just went through Translatewiki.net fixing a number of [syntax errors](https://github.com/openstreetmap/openstreetmap-website/blob/4eaf4404ad5629f0f5341433720cabd63518035a/config/locales/ja.yml#L2485-L2486) from [missing braces](https://github.com/openstreetmap/openstreetmap-website/blob/4eaf4404ad5629f0f5341433720cabd63518035a/config/locales/fi.yml#L3029-L3030) in the `{{PLURAL|…}}` syntax that were probably introduced by translators accepting machine translation suggestions without reading them closely.

Along the way, I noticed that some translators have gotten the idea that OSM messages there support full MediaWiki wikitext syntax, since [the `{{PLURAL|…}}` syntax for CLDR-based projects](https://translatewiki.net/wiki/Plural#Plural_in_CLDR) looks a lot like the `{{PLURAL:…}}` magic word in wikitext. In particular, they’re adding [`{{GENDER:…}}`](https://translatewiki.net/wiki/Gender) to switch between grammatical genders where the language apparently makes a distinction.

https://github.com/openstreetmap/openstreetmap-website/blob/4eaf4404ad5629f0f5341433720cabd63518035a/config/locales/hsb.yml#L1488-L1489 https://github.com/openstreetmap/openstreetmap-website/blob/4eaf4404ad5629f0f5341433720cabd63518035a/config/locales/sq.yml#L1142

The Greek localization also used this keyword in e-mail subject lines until b45eb03a6c026ec0f19415100c7bb14c4c53e6cb.

In theory, we could support varying interface messages by grammatical gender by adding a user preference. That’s how MediaWiki supports the `{{GENDER:…}}` magic word throughout its interface. But storing this information would raise privacy concerns, especially if a user can get e-mail notifications that address another user by their choice of grammatical gender. A preference seems like overkill. It seems like we have a relatively small number of potentially gender-specific messages, though it’s hard to tell if some languages are just working around the problem.

Unless anyone feels strongly otherwise, I’m thinking we could just replace the `{{GENDER:…}}` syntax with a more literal “foo/bar” format, similar to the workaround that the Greek translators are using.

tomhughes:
Unless you know of some i18n plugin to do that it sounds like "in theory" is doing a lot of heavy lifting here...

tomhughes:
[...]

PR #6342, opened 2025-08-21
Add test for mediawiki magic words in translations

tomhughes:
Add test for GENDER macro in translations and fix broken translations as discussed in #6341.

### DIFF
--- a/test/lib/i18n_test.rb
+++ b/test/lib/i18n_test.rb
@@ -57,10 +57,17 @@ class I18nTest < ActiveSupport::TestCase
 
   Rails.root.glob("config/locales/*.yml").each do |filename|
     code = File.basename(filename, ".yml")
+    yml = YAML.load_file(filename)
+
     test "#{code} for raw html" do
-      yml = YAML.load_file(filename)
       assert_nothing_raised do
         check_values_for_raw_html(yml)
+      end
+    end
+
+    test "#{code} for mediawiki magic" do
+      assert_nothing_raised do
+        check_values_for_mediawiki_magic(yml)
       end
     end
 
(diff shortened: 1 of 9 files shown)


test/lib/i18n_test.rb
```diff
@@ -135,6 +142,16 @@ def check_values_for_raw_html(hash)
     end
   end
 
+  def check_values_for_gender_macro(hash)
+    hash.each_pair do |k, v|
+      if v.is_a? Hash
+        check_values_for_gender_macro(v)
+      else
+        raise "Avoid using GENDER macro in '#{k}: #{v}'" if v.include? "{{GENDER:"
```
1ec5: This would be a good way to detect the broken `{{PLURAL` syntax I spotted too, since the correct syntax on Translatewiki.net populates different entries in our YAML file instead of leaving that token around.

1ec5 (approved):

test/lib/i18n_test.rb
tomhughes: I've updated it to handle `PLURAL` and `GRAMMAR` as well now.

AntonKhorev:
Shouldn't the commits be in the opposite order, for the same reasons as https://github.com/openstreetmap/openstreetmap-website/pull/6318#issuecomment-3184530884?

tomhughes:
I guess... I wrote them that way round because I wrote the test and checked that it caught the problem and then fixed the problem.

I've reversed them now.

AntonKhorev:
Merged, thanks.

tomhughes:
@translatewiki just a heads up, this deletes some invalid translations so might need manual intervention on your side.

PR #6344, opened 2025-08-22
Cache expanded preferred_languages

mmd-osm:
`preferred_languages.expand` in browse_helper.rb is a bit expensive, due to the Locale handling that's being triggered for every single element (see stacktrace below).

However, preferred_languages are not expected to change during a request. They seem like a good candidate for caching. preferred_languages is already cached in the application_controller, but the expanded version is not.

~Caching should work on a request basis. After the request is done, the instance variable should be empty. Unit tests somehow don't have a concept of a request, that's why I had to manually set the value back to nil.~

I've seen about 50% performance improvement on top of #6338 by caching preferred_languages this way.

locale_keys

### DIFF
--- a/app/helpers/browse_helper.rb
+++ b/app/helpers/browse_helper.rb
@@ -73,6 +73,8 @@ module BrowseHelper
   private
 
   def feature_name(tags)
+    return nil if tags.empty?
+
     locale_keys = preferred_languages.expand.map { |locale| "name:#{locale}" }
 
     (locale_keys + %w[name ref addr:housename]).each do |key|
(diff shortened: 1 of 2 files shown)


tomhughes:
So rubocop has thrown a fit on https://docs.rubocop.org/rubocop-rails/cops_rails.html#railshelperinstancevariable which is a bit of a problem.

I wonder if we could get most of the benefit by making `Locale::List#expand` and/or `Locale#candidates` memoize their result - the locale to tag name mapping would still have to be done each time but I suspect the locale expansion is the expensive part?

mmd-osm:
> Locale::List#expand and/or Locale#candidates memoize their result

I haven't really considered this point. Somehow I thought this is some third party library code, and not something we could change in our repo.

> the locale to tag name mapping would still have to be done each time but I suspect the locale expansion is the expensive part?

Yes, locale expand is the expensive part.

tomhughes:

PR #6347, opened 2025-08-23
Backfill enhanced changeset stats

mmd-osm:
One topic that was left in #6170 is backfilling of old changesets. I have created a new rake task for the job. By default, if updates all changesets where the enhanced changeset stats differ from num_changes in chunks of 10'000.
Start/end changeset id, as well as chunk size can be adjusted using environment variables.

It's probably a good idea to try this out first on https://master.apis.dev.openstreetmap.org.

### DIFF
--- a/lib/tasks/backfill_changeset_stats.rake
+++ b/lib/tasks/backfill_changeset_stats.rake
@@ -1,0 +1,111 @@
+# rubocop:disable Metrics/BlockLength
+namespace :db do
+  desc "Backfill enhanced changeset stats"
+  task :changeset_stats => :environment do
+    chunk_size = ENV["CHUNK_SIZE"]&.to_i || 10_000
+
+    min_id = ENV["MIN_CHANGESET"]&.to_i || Changeset.minimum(:id)
+    max_id = ENV["MAX_CHANGESET"]&.to_i || Changeset.maximum(:id)
+
+    puts "Backfilling enhanced changeset stats from id #{min_id} to #{max_id} in chunks of #{chunk_size}"
+
+    Changeset.where(:id => min_id..max_id)
+             .where('(num_created_nodes + num_modified_nodes + num_deleted_nodes +
+                      num_created_ways + num_modified_ways + num_deleted_ways +
+                      num_created_relations + num_modified_relations + num_deleted_relations)
+                     != num_changes')
+             .in_batches(:of => chunk_size) do |batch|
+      ids = batch.ids
+      print "Processing changesets #{ids.first} to #{ids.last} ..."
+
+      id_list = "{#{ids.join(',')}}"
+
+      # Backfill enhanced changeset stats
+      sql = <<~SQL.squish
+        WITH changes AS (
+            SELECT
+              nodes.changeset_id,
+              CASE WHEN nodes.version = 1 THEN 1 ELSE 0 END AS num_created_nodes,
+              CASE WHEN nodes.version > 1 AND nodes.visible THEN 1 ELSE 0 END AS num_modified_nodes,
+              CASE WHEN nodes.version > 1 AND NOT nodes.visible THEN 1 ELSE 0 END AS num_deleted_nodes,
+              0 AS num_created_ways,
+              0 AS num_modified_ways,
+              0 AS num_deleted_ways,
+              0 AS num_created_relations,
+              0 AS num_modified_relations,
+              0 AS num_deleted_relations
+            FROM nodes
+            WHERE nodes.changeset_id = ANY($1::bigint[])
+          UNION ALL
+            SELECT
+              ways.changeset_id,
+              0 AS num_created_nodes,
+              0 AS num_modified_nodes,
+              0 AS num_deleted_nodes,
+              CASE WHEN ways.version = 1 THEN 1 ELSE 0 END AS num_created_ways,
+              CASE WHEN ways.version > 1 AND ways.visible THEN 1 ELSE 0 END AS num_modified_ways,
+              CASE WHEN ways.version > 1 AND NOT ways.visible THEN 1 ELSE 0 END AS num_deleted_ways,
+              0 AS num_created_relations,
+              0 AS num_modified_relations,
+              0 AS num_deleted_relations
+            FROM ways
+            WHERE ways.changeset_id = ANY($1::bigint[])
+          UNION ALL
+            SELECT
+              relations.changeset_id,
+              0 AS num_created_nodes,
+              0 AS num_modified_nodes,
+              0 AS num_deleted_nodes,
+              0 AS num_created_ways,
+              0 AS num_modified_ways,
+              0 AS num_deleted_ways,
+              CASE WHEN relations.version = 1 THEN 1 ELSE 0 END AS num_created_relations,
+              CASE WHEN relations.version > 1 AND relations.visible THEN 1 ELSE 0 END AS num_modified_relations,
+              CASE WHEN relations.version > 1 AND NOT relations.visible THEN 1 ELSE 0 END AS num_deleted_relations
+            FROM relations
+            WHERE relations.changeset_id = ANY($1::bigint[])
+        ),
+        total AS (
+          SELECT
+            changes.changeset_id,
+            SUM(changes.num_created_nodes) AS num_created_nodes,
+            SUM(changes.num_modified_nodes) AS num_modified_nodes,
+            SUM(changes.num_deleted_nodes) AS num_deleted_nodes,
+            SUM(changes.num_created_ways) AS num_created_ways,
+            SUM(changes.num_modified_ways) AS num_modified_ways,
+            SUM(changes.num_deleted_ways) AS num_deleted_ways,
+            SUM(changes.num_created_relations) AS num_created_relations,
+            SUM(changes.num_modified_relations) AS num_modified_relations,
+            SUM(changes.num_deleted_relations) AS num_deleted_relations
+          FROM changes
+          GROUP BY changes.changeset_id
+        )
+        UPDATE changesets
+        SET num_created_nodes      = total.num_created_nodes,
+            num_modified_nodes     = total.num_modified_nodes,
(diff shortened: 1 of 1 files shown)


lib/tasks/backfill_changeset_stats.rake
```diff
@@ -0,0 +1,126 @@
...
+            COALESCE(num_created_nodes, 0) +
+            COALESCE(num_modified_nodes, 0) +
+            COALESCE(num_deleted_nodes, 0) +
+            COALESCE(num_created_ways, 0) +
+            COALESCE(num_modified_ways, 0) +
+            COALESCE(num_deleted_ways, 0) +
+            COALESCE(num_created_relations, 0) +
+            COALESCE(num_modified_relations, 0) +
+            COALESCE(num_deleted_relations, 0)
+          ) IS DISTINCT FROM num_changes
```
tomhughes: None of those columns are nullable so we don't really need the `COALESECE` calls and we can use normal comparison rather than `IS DISTINCT FROM`?

mmd-osm: fixed

mmd-osm:
I tried to use in_batches now, and moved the former preselection query to the "where" clause. Also, instead of CTE, I'm using a prepared statement and pass the list of ids as a parameter.

As a side note (it's not causing any obvious issues): `batch.ids` creates some wild SQL behind the scenes that includes the preselection check as well. https://github.com/rails/rails/issues/47462#issuecomment-1729471451 suggested the following, which I found too much of a hack in the end.

```
      ids = batch.where_values_hash[batch.klass.primary_key]
```

tomhughes:
I've managed to avoid the need for multiple nested aggregations, which tests suggest is quite a bit more efficient, by doing this:

```sql
WITH total AS (
  SELECT
    changesets.id AS changeset_id,
    SUM(CASE WHEN nodes.version = 1 THEN 1 ELSE 0 END) AS num_created_nodes,
    SUM(CASE WHEN nodes.version > 1 AND nodes.visible THEN 1 ELSE 0 END) AS num_modified_nodes,
    SUM(CASE WHEN nodes.version > 1 AND NOT nodes.visible THEN 1 ELSE 0 END) AS num_deleted_nodes,
    SUM(CASE WHEN ways.version = 1 THEN 1 ELSE 0 END) AS num_created_ways,
    SUM(CASE WHEN ways.version > 1 AND ways.visible THEN 1 ELSE 0 END) AS num_modified_ways,
    SUM(CASE WHEN ways.version > 1 AND NOT ways.visible THEN 1 ELSE 0 END) AS num_deleted_ways,
    SUM(CASE WHEN relations.version = 1 THEN 1 ELSE 0 END) AS num_created_relations,
    SUM(CASE WHEN relations.version > 1 AND relations.visible THEN 1 ELSE 0 END) AS num_modified_relations,
    SUM(CASE WHEN relations.version > 1 AND NOT relations.visible THEN 1 ELSE 0 END) AS num_deleted_relations
  FROM changesets
  LEFT OUTER JOIN nodes ON nodes.changeset_id = changesets.id
  LEFT OUTER JOIN ways ON ways.changeset_id = changesets.id
  LEFT OUTER JOIN relations ON relations.changeset_id = changesets.id
  WHERE changesets.id = ANY($1::bigint[])
  GROUP BY changesets.id
)
UPDATE changesets
SET num_created_nodes      = total.num_created_nodes,
    num_modified_nodes     = total.num_modified_nodes,
    num_deleted_nodes      = total.num_deleted_nodes,
    num_created_ways       = total.num_created_ways,
    num_modified_ways      = total.num_modified_ways,
    num_deleted_ways       = total.num_deleted_ways,
    num_created_relations  = total.num_created_relations,
    num_modified_relations = total.num_modified_relations,
    num_deleted_relations  = total.num_deleted_relations
FROM total
WHERE changesets.id = total.changeset_id;
```

I also got rid of the renaming of the CTE in the update but it didn't seem to serve any purpose.

In principle the CTE shouldn't need to access `changesets` but when I tried to replace that with an `unnest` of the array postgres decided to table scan `relations` instead of using the index which was very bad.

mmd-osm:
Thanks a lot for testing the query with a bigger data volume. In my tests, I used a smaller extract, which isn't ideal to anticipate performance issues in a production environment. I have replaced the previous code with your suggestion and pushed it again.

tomhughes:
It would have been better if I'd tested that the results were correct though, which they aren't :-(

You can get rid of the nested grouping but you do still need the union, like this:

```sql
WITH changes AS (
    SELECT
      nodes.changeset_id,
      CASE WHEN nodes.version = 1 THEN 1 ELSE 0 END AS num_created_nodes,
      CASE WHEN nodes.version > 1 AND nodes.visible THEN 1 ELSE 0 END AS num_modified_nodes,
      CASE WHEN nodes.version > 1 AND NOT nodes.visible THEN 1 ELSE 0 END AS num_deleted_nodes,
      0 AS num_created_ways,
      0 AS num_modified_ways,
      0 AS num_deleted_ways,
      0 AS num_created_relations,
      0 AS num_modified_relations,
      0 AS num_deleted_relations
    FROM nodes
    WHERE nodes.changeset_id = ANY($1::bigint[])
  UNION ALL
    SELECT
      ways.changeset_id,
      0 AS num_created_nodes,
      0 AS num_modified_nodes,
      0 AS num_deleted_nodes,
      CASE WHEN ways.version = 1 THEN 1 ELSE 0 END AS num_created_ways,
      CASE WHEN ways.version > 1 AND ways.visible THEN 1 ELSE 0 END AS num_modified_ways,
      CASE WHEN ways.version > 1 AND NOT ways.visible THEN 1 ELSE 0 END AS num_deleted_ways,
      0 AS num_created_relations,
      0 AS num_modified_relations,
      0 AS num_deleted_relations
    FROM ways
    WHERE ways.changeset_id = ANY($1::bigint[])
  UNION ALL
    SELECT
      relations.changeset_id,
      0 AS num_created_nodes,
      0 AS num_modified_nodes,
      0 AS num_deleted_nodes,
      0 AS num_created_ways,
      0 AS num_modified_ways,
      0 AS num_deleted_ways,
      CASE WHEN relations.version = 1 THEN 1 ELSE 0 END AS num_created_relations,
      CASE WHEN relations.version > 1 AND relations.visible THEN 1 ELSE 0 END AS num_modified_relations,
      CASE WHEN relations.version > 1 AND NOT relations.visible THEN 1 ELSE 0 END AS num_deleted_relations
    FROM relations
    WHERE relations.changeset_id = ANY($1::bigint[])
),
total AS (
  SELECT
    changes.changeset_id,
    SUM(changes.num_created_nodes) AS num_created_nodes,
    SUM(changes.num_modified_nodes) AS num_modified_nodes,
    SUM(changes.num_deleted_nodes) AS num_deleted_nodes,
    SUM(changes.num_created_ways) AS num_created_ways,
    SUM(changes.num_modified_ways) AS num_modified_ways,
    SUM(changes.num_deleted_ways) AS num_deleted_ways,
    SUM(changes.num_created_relations) AS num_created_relations,
    SUM(changes.num_modified_relations) AS num_modified_relations,
    SUM(changes.num_deleted_relations) AS num_deleted_relations
  FROM changes
  GROUP BY changes.changeset_id
)
UPDATE changesets
SET num_created_nodes      = total.num_created_nodes,
    num_modified_nodes     = total.num_modified_nodes,
    num_deleted_nodes      = total.num_deleted_nodes,
    num_created_ways       = total.num_created_ways,
    num_modified_ways      = total.num_modified_ways,
    num_deleted_ways       = total.num_deleted_ways,
    num_created_relations  = total.num_created_relations,
    num_modified_relations = total.num_modified_relations,
    num_deleted_relations  = total.num_deleted_relations
FROM total
WHERE changesets.id = total.changeset_id
 ```

It turns out though that there doesn't seem to be much point as a naive version using subqueries is just as fast:

```sql
UPDATE changesets
   SET num_created_nodes = (SELECT COUNT(*) FROM nodes WHERE changeset_id = changesets.id AND version = 1),
       num_modified_nodes = (SELECT COUNT(*) FROM nodes WHERE changeset_id = changesets.id AND version > 1 AND visible),
       num_deleted_nodes = (SELECT COUNT(*) FROM nodes WHERE changeset_id = changesets.id AND version > 1 AND NOT visible),
       num_created_ways = (SELECT COUNT(*) FROM ways WHERE changeset_id = changesets.id AND version = 1),
       num_modified_ways = (SELECT COUNT(*) FROM ways WHERE changeset_id = changesets.id AND version > 1 AND visible),
       num_deleted_ways = (SELECT COUNT(*) FROM ways WHERE changeset_id = changesets.id AND version > 1 AND NOT visible),
       num_created_relations = (SELECT COUNT(*) FROM relations WHERE changeset_id = changesets.id AND version = 1),
       num_modified_relations = (SELECT COUNT(*) FROM relations WHERE changeset_id = changesets.id AND version > 1 AND visible),
       num_deleted_relations = (SELECT COUNT(*) FROM relations WHERE changeset_id = changesets.id AND version > 1 AND NOT visible)
WHERE id = ANY($1::bigint[])
```

Both take about 5s to do the first 10000 changesets when the cache is hot for the indexes.

mmd-osm:
What a pity. I only did a quick performance check, but haven't validated results :(

I have now pushed the first (union all) version. In my local test, it happens to be about 10 times faster than the naive version. I suppose the values I'm seeing in `EXPLAIN (verbose, analyze, buffers)` are in a similar way heavily skewed by the somewhat low amount of data in my db. At least they're consistent with the actual runtimes I'm seeing.

One thing you could maybe try out is the CHUNK_SIZE parameter. I've been testing values of 100'000 and 1'000'000. The default value of 10'000 didn't turn out to be the most efficient one.

tomhughes:
I was running the SQL manually with a modified where clause that just did `id < 10000` for testing with the production database.

tomhughes:
Thanks. Sorry for all the messing about about I think this is probably good to merge now.

mmd-osm:
According to prometheus, backfilling took about 9.5 hours, which is ok for a one time activity.

PR #6351, opened 2025-08-25
Inline auth provider logos and make Apple logo invertible

AntonKhorev:


AntonKhorev:
Tests fail because Translatewiki.

tomhughes:
I mean it looks a bit shit being out of proportion to the others but if that's what Apple wants...

I'm not sure increasing the size really helps anyway because the problem is not so much the absolute size as the relative comparison with the other icons.

PR #6352, opened 2025-08-25
Fix test failures from #6351

tomhughes:
The translation noise in the #6351 failures was hiding a real failure, which this fixes.

### DIFF
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -1,4 +1,6 @@
 module UserHelper
+  include InlineSvg::ActionView::Helpers
+
   # User images
 
   def user_image(user, options = {})
--- a/test/helpers/user_helper_test.rb
+++ b/test/helpers/user_helper_test.rb
@@ -111,8 +111,7 @@ class UserHelperTest < ActionView::TestCase
 
   def test_auth_button
     button = auth_button("google")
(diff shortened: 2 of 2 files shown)


rkoeze (approved):
LGTM, thanks

AntonKhorev:
`test/helpers/user_helper_test.rb` also fails on `test_user_image_sizes_avatar` for me, depending on the order of execution.

AntonKhorev:
If `test_user_image_sizes_gravatar` runs before `test_user_image_sizes_avatar`, `test_user_image_sizes_avatar` passes;
if `test_user_image_sizes_avatar` runs first, it fails.

AntonKhorev:
It can't be related to my changes, but I don't even see how it is related to anything. A test fails/passes depending on previous `Kernel.respond_to?` call. Tried on ruby 3.2.6, then 3.2.9.

```ruby
  def test_what
    # fails without this line:
    Kernel.instance_method(:respond_to?).bind_call(self, :does_not_matter_what)

    user = create(:user)
    user.avatar.attach(:io => File.open("test/gpx/fixtures/a.gif"), :filename => "a.gif")
    variant = user.avatar.variant(:resize_to_limit => [100, 100])
    image_tag variant
  end
```

PR #6353, opened 2025-08-25
Added error handling for object display on map

mmd-osm:
Now that we're able to display large relations without hitting gateway timeout errors, users will try this out on fairly large objects. At some point they're likely to encounter rate limiting. Previously, any error messages were silently discarded, and an empty map is shown instead, which is a bit confusing.

I'm reusing the good old error handling code from app/assets/javascripts/index/layers/data.js in this PR to display a proper error message instead:

error1

### DIFF
--- a/app/assets/javascripts/download_util.js
+++ b/app/assets/javascripts/download_util.js
@@ -54,3 +54,17 @@ OSM.turboHtmlResponseHandler = function (event) {
     event.stopPropagation();
   }
 };
+
+OSM.displayLoadError = function (message, close) {
+  $("#browse_status").html(
+    $("<div class='p-3'>").append(
+      $("<div class='d-flex'>").append(
+        $("<h2 class='flex-grow-1 text-break'>")
+          .text(OSM.i18n.t("browse.start_rjs.load_data")),
+        $("<div>").append(
+          $("<button type='button' class='btn-close'>")
+            .attr("aria-label", OSM.i18n.t("javascripts.close"))
+            .on("click", close))),
+      $("<p class='alert alert-warning'>")
+        .text(OSM.i18n.t("browse.start_rjs.feature_error", { message: message }))));
+};
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -1,3 +1,4 @@
+//= require download_util
 L.extend(L.LatLngBounds.prototype, {
   getSize: function () {
     return (this._northEast.lat - this._southWest.lat) *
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/download_util.js
```diff
@@ -54,3 +54,17 @@ OSM.turboHtmlResponseHandler = function (event) {
...
+OSM.displayLoadError = function (message, close) {
+  $("#browse_status").html(
+    $("<div class='p-3'>").append(
+      $("<div class='d-flex'>").append(
+        $("<h2 class='flex-grow-1 text-break'>")
+          .text(OSM.i18n.t("browse.start_rjs.load_data")),
+        $("<div>").append(
+          $("<button type='button' class='btn-close'>")
+            .attr("aria-label", OSM.i18n.t("javascripts.close"))
+            .click(close))),
```
hlfan: that should be written with on("click", close)

mmd-osm: Initially, I wanted to move this function "as is" to a new place, but I guess we can address this issue.

app/assets/javascripts/leaflet.map.js
```diff
@@ -300,8 +312,28 @@ L.OSM.Map = L.Map.extend({
...
+          $("<div class='d-flex'>").append(
+            $("<h2 class='flex-grow-1 text-break'>")
+              .text(OSM.i18n.t("browse.start_rjs.load_data")),
+            $("<div>").append(
+              $("<button type='button' class='btn-close'>")
+                .attr("aria-label", OSM.i18n.t("javascripts.close"))
+                .click(close))),
+          $("<p class='alert alert-warning'>")
+            .text(OSM.i18n.t("browse.start_rjs.feature_error", { message: message }))));
     }
```
AntonKhorev: Why don't you first introduce and use `OSM.displayLoadError` instead of adding this function and then deleting it?

mmd-osm: fixed by squashing commits

tomhughes:
Looks good to me, thanks.

deevroman:
There seems to be no need to show an error message for deleted objects



https://osm.org/node/13100171302

app/assets/javascripts/leaflet.map.js
```diff
@@ -277,7 +278,19 @@ L.OSM.Map = L.Map.extend({
...
         signal: this._objectLoader.signal
       })
-        .then(response => response.json())
+        .then(async response => {
+          if (response.ok) {
+            return response.json();
+          }
+
+          const status = response.statusText || response.status;
+          if (response.status !== 400 && response.status !== 509) {
```
hlfan, line 287: To ignore deleted element response errors:
```suggestion
          if (![400, 410, 509].includes(response.status)) {
```

mmd-osm: No, this will not work and only change the error text itself. Closing here, follow up in #6363.

PR #6354, opened 2025-08-26
Pull docker ruby version into arg

rkoeze:
### Description
Pull the Dockerfile Ruby version in to an ARG so we can override it. This will make it easier to run Ruby 3.3 and 3.4, [which we also support](https://github.com/openstreetmap/openstreetmap-website/blob/master/INSTALL.md#prerequisites). Now building the Dockerfile with a different Ruby version can be done like this:

```
docker-compose build --build-arg RUBY_VERSION=3.3
```

### How has this been tested?
The tests pass and I can run the app locally. I was able to confirm the correct Ruby version is still being pulled after the change by looking at the Docker build logs:

```
 => [web  1/10] FROM docker.io/library/ruby:3.2-bookworm@sha256:4bcc61f192db4d9f72c87998fa77f97deaaeb0c1021951632f4cb986b3c779de                                                                                                               33.8s
```

And this is with the RUBY_VERSION set by the command line:

```
=> [web  1/10] FROM docker.io/library/ruby:3.3-bookworm@sha256:49ee09d4811028286bb9510599999622fb7e6264190ebe645e17b3b078d633a6                                                                                                                3.7s
```


Firefishy:
Would it not be better to have the whole image as an ARG? eg: `FROM $RUBY_IMAGE` or similar, as this would allow testing different upstream images eg: `ghcr.io/ruby/ruby:3.3.5-noble`

rkoeze:
I am not completely sure. This change was motivated by the fact we support Ruby 3.2+ and so we might reasonably expect a contributor to want to develop against a different version. That seems like a relatively safe use case to support. Since we also expect that version to change over time, it also seemed like a decent candidate for parameterization.

Allowing users to pass entire images, however, feels a bit different. Our Dockerfile makes certain assumptions that parameterizing the base image may break. Depending on the image passed some of the commands in our Dockerfile may not work. The names of packages may also be different. My instinct is to avoid having to deal with that.

rkoeze:
Here are a couple of examples of how parameterizing the entire Docker image will break the build: 
* rubylang/ruby:3.2-jammy
```
6.690 E: Package 'firefox-esr' has no installation candidate
------
failed to solve: process "/bin/sh -c apt-get update   && apt-get install --no-install-recommends -y   build-essential   curl   default-jre-headless   file   git-core   gpg-agent   libarchive-dev   libffi-dev   libgd-dev   libpq-dev   libsasl2-dev   libvips-dev   libxml2-dev   libxslt1-dev   libyaml-dev   locales   postgresql-client   tzdata   unzip   nodejs   npm   osmosis   ca-certificates   firefox-esr" did not complete successfully: exit code: 100
```
* ruby:3.2-alpine3.2
```
 > [web  2/10] RUN apt-get update   && apt-get install --no-install-recommends -y   build-essential   curl   default-jre-headless   file   git-core   gpg-agent   libarchive-dev   libffi-dev   libgd-dev   libpq-dev   libsasl2-dev   libvips-dev   libxml2-dev   libxslt1-dev   libyaml-dev   locales   postgresql-client   tzdata   unzip   nodejs   npm   osmosis   ca-certificates   firefox-esr:
0.083 /bin/sh: apt-get: not found
```

gravitystorm:
Merged, thanks!

PR #6356, opened 2025-08-26
Use @elements instead of @elems in api element history

AntonKhorev:
Eliminates `@elems` instance variable. Otherwise there are both `@elements` and `@elems` and the difference is that one is filtered by redactions and the other is not, which you can't guess from their names.


tomhughes:
Looks good to me, thanks.

PR #6323, opened 2025-08-14
Forward preferred color scheme to id-container

hlfan:
Companion PR for https://github.com/openstreetmap/iD/pull/11308.

PR #6359, opened 2025-08-27
Update to iD v2.36.0

tyrasd:
(changelog: https://github.com/openstreetmap/iD/releases/tag/v2.36.0)


tomhughes:
Does this mean we're ready to take #6323 as well?

tyrasd:
not quite yet, the dark-mode support is planned for v2.37 :blush:

hlfan:
As the [commit names](https://github.com/openstreetmap/iD/pull/11308/commits) would suggest, this is not something to throw in at the end of a release cycle...
I'll update #6323 soon enough.

PR #6361, opened 2025-08-27
Update num_changes when backfilling changesets

tomhughes:
It seems that at least among the early changesets there are a number where `num_changes` is not correct so they still appear to have out of date statistics after running the backfill.

So far I see about 25000 such changesets in the first 3 million.

This PR adjusts the backfill to update `num_changes` as well as the individual components.

### DIFF
--- a/lib/tasks/backfill_changeset_stats.rake
+++ b/lib/tasks/backfill_changeset_stats.rake
@@ -91,7 +91,16 @@ namespace :db do
             num_deleted_ways       = total.num_deleted_ways,
             num_created_relations  = total.num_created_relations,
             num_modified_relations = total.num_modified_relations,
-            num_deleted_relations  = total.num_deleted_relations
+            num_deleted_relations  = total.num_deleted_relations,
+            num_changes            = total.num_created_nodes +
+                                     total.num_modified_nodes +
+                                     total.num_deleted_nodes +
+                                     total.num_created_ways +
+                                     total.num_modified_ways +
+                                     total.num_deleted_ways +
+                                     total.num_created_relations +
+                                     total.num_modified_relations +
+                                     total.num_deleted_relations
         FROM total
         WHERE changesets.id = total.changeset_id
       SQL


tomhughes:
There appear to be 56331 changesets in total that are still invalid after backfilling.

Most are early on but some like https://www.openstreetmap.org/changeset/168981909 are only from a few weeks ago - that has num_changes as 1 but it clearly has 2 changes.

mmd-osm:
I don't know if it's feasible to identify the root cause for these deviations, given that it's 56k out of 171 million changesets. As a starting point, a list of relevant changesets for the last year or so would be good. Maybe there's some pattern in the data, like, it's always happening for some app or either on Rails or cgimap.

tomhughes:
This is the most recent 50 where num_changes seems to be wrong, with the current and correct values:

```
    id     | current_num_changes | correct_num_changes 
-----------+---------------------+---------------------
 168981909 |                   1 |                   2
 168900195 |                   1 |                  10
 168890737 |                   1 |                   6
 168890675 |                   1 |                   3
 168886428 |                   1 |                   9
 168883250 |                   1 |                   9
 168853665 |                   1 |                   2
 168812239 |                   2 |                  10
 168617198 |                   1 |                   2
 168615876 |                   1 |                   2
 168615874 |                   1 |                   2
 168615495 |                   1 |                   7
 168615493 |                   1 |                   7
 168615490 |                   1 |                   7
 168615489 |                   1 |                   7
 168615486 |                   1 |                   7
 168615483 |                   1 |                   7
 168615477 |                   2 |                   7
 168615467 |                   1 |                   7
 168615463 |                   1 |                   7
 168615461 |                   1 |                   7
 168615457 |                   2 |                   7
 168539640 |                   1 |                   2
 168535456 |                   6 |                  52
 168535319 |                   3 |                  45
 168533944 |                   1 |                   6
 168497084 |                   1 |                   2
 168497011 |                   1 |                   4
 168482383 |                   1 |                   4
 168480002 |                   1 |                   2
 168458124 |                   1 |                   2
 168457857 |                   2 |                   9
 168457710 |                   5 |                  16
 168457479 |                   1 |                   2
 168457454 |                   1 |                   7
 168457404 |                   1 |                   2
 168457367 |                   2 |                   3
 168457357 |                   1 |                   3
 168457332 |                   4 |                  27
 168457088 |                   2 |                   9
 168448455 |                   2 |                   9
 168446773 |                   3 |                  18
 168445364 |                   3 |                  31
 168319929 |                   1 |                   4
 168319848 |                   2 |                  16
 168319369 |                   2 |                  20
 168266745 |                   1 |                   3
 168266722 |                   1 |                   2
 168266611 |                   1 |                   4
 168254829 |                   2 |                  23
(50 rows)
```

Interestingly all the ones I've looked at are OsmAPP 1.7.0 so that may be a clue.

tomhughes:
Looks like it uses [diff upload](https://github.com/zbycz/osmapp/blob/a813f394d99c680e0cd4a991c4752397f7967f9e/src/services/osm/auth/osmApiAuth.ts#L118).

mmd-osm:
Somehow it suddenly stopped on 2025-07-15, which is about 1.5 weeks before the cgimap 2.1.0 upgrade. According to OSMCha, the app has created about 1800 changesets since then.

mmd-osm:
I believe these version numbers are a bit meaningless. v1.7.0 was released on June 26. A Chinese translation was added about a month later in https://github.com/zbycz/osmapp/commit/82fe653a4e385aa58fbc55aea2f1b00148bfd308, and this translation is already available on https://osmapp.org.

I think it's best to ask the author @zbycz, if a change to the OSM upload mechanism has been deployed to the site on July 15. 

https://github.com/zbycz/osmapp/blob/master/src/services/osm/auth/api.ts still has some single element operations in place, like createNodeItem. Maybe back then until mid-July, OsmAPP was flooding the server with lots of single element async requests.

I don't remember if Rails checks the current number of changes while holding a lock on Changesets. If not, then there's a chance of a race condition. On cgimap, there's an exclusive lock on the changeset id in place, which prevents this sort of issue.

By the way, the bounding box is also incorrect in this case:

https://www.openstreetmap.org/node/12980345238 which is outside of https://www.openstreetmap.org/changeset/168535456

zbycz:
Hi - yes, the versioning is only for summarizing changes once in a while. Otherwise osmapp is upgraded continously with every commit. Should I maybe include the commit hash in the version? 

We deployed the xml diff upload on 29/4 (https://github.com/zbycz/osmapp/pull/1009) - until then the app may have been sending more single requests in one changests. This was happening only when editing relation and its members - the multi-editor was added on 3rd of Dec (https://github.com/zbycz/osmapp/pull/831).

I am not aware of any such changes mid-july - since end of April it should be just the diff.

One idea - maybe someone could have forked and deployed some older version? but thats not very probable.

zbycz:
I checked the changesets and I think I have an idea what is happening. If user opens more items and doesn't make any changes - they are all submitted as `action='modify'` even though the properites are the same. 

I wanted to fix that some time ago, but haven't done it yet. Added issue: https://github.com/zbycz/osmapp/issues/1319

mmd-osm:
Thanks a lot for the quick reply. Here's another hint based on the list of changesets above [link](https://api.openstreetmap.org/api/0.6/changesets?changesets=168981909,168900195,168890737,168890675,168886428,168883250,168853665,168812239,168617198,168615876,168615874,168615495,168615493,168615490,168615489,168615486,168615483,168615477,168615467,168615463,168615461,168615457,168539640,168535456,168535319,168533944,168497084,168497011,168482383,168480002,168458124,168457857,168457710,168457479,168457454,168457404,168457367,168457357,168457332,168457088,168448455,168446773,168445364,168319929,168319848,168319369,168266745,168266722,168266611,168254829,):

All changesets have this tag:

```
<tag k="submitted_from" v="https://openclimbing.org/relation/xxx"/>
```

Is this site running the same code as osmapp.org?

zbycz:
Yes, it is the same app/deploy only different domain. 

See my comment above - do you think this is the root cause?

mmd-osm:
I had a quick look at https://www.openstreetmap.org/changeset/168981909, it seems both nodes have been changed and we still see a deviation for the number of changes. Do you happen to have an example where nodes haven't been changed at all?

I haven't really looked much into the code, but I'm assuming that you would upload all changes in a single request, like everyone else does, right?

mmd-osm:
@tomhughes : in https://www.openstreetmap.org/api/0.6/changeset/168535456/download, I've noticed that not all nodes have the same timestamp, even though the changeset is modifying nodes only. When uploading using cgimap, my expectation is that all nodes have the exact same timestamp, down to the microsecond.

It would be interesting to see the exact microsecond level timestamps for all nodes in changeset 168535456. Are they all slightly different?

zbycz:
@mmd-osm I tried a test against the Test API. I created a new node + its relation  ([changeset](https://master.apis.dev.openstreetmap.org/changeset/437816)), and then tried to change only the relation, but "open" the node in the UI as well and make no changes to it – https://master.apis.dev.openstreetmap.org/changeset/437817. Due to https://github.com/zbycz/osmapp/issues/1319 it sends also the node. There no changes happened, but the version increased nevertheless.

The exact osc sent in this case was:
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<osmChange generator="OsmAPP" version="0.6">
    <create/>
    <modify>
        <node id="4360282798" lon="14.25774" lat="49.661" version="1" changeset="437817">
            <tag k="climbing" v="route_bottom"/>
            <tag k="name" v="new node"/>
            <tag k="sport" v="climbing"/>
        </node>
        <relation id="4305234299" version="1" changeset="437817">
            <tag k="climbing" v="crag"/>
            <tag k="name" v="new relation - changed (but node unchanged)"/>
            <tag k="site" v="climbing"/>
            <tag k="sport" v="climbing"/>
            <tag k="type" v="site"/>
            <member type="node" ref="4360282798" role=""/>
        </relation>
    </modify>
    <delete if-unused="true"/>
</osmChange>
```


Could the issue be, that `current_num_changes` counts only the objects where some values (tags/location/members) had really changed, but the backfill (`correct_num_changes`) counts merely the number of objects in the changeset? So it counts all the submitted items, even though some were not changed at all? (only version increased)

AntonKhorev:
> I don't remember if Rails checks the current number of changes while holding a lock on Changesets. If not, then there's a chance of a race condition.

There's no changeset lock for single element writes.

> On cgimap, there's an exclusive lock on the changeset id in place, which prevents this sort of issue.

cgimap doesn't have single element writes.

mmd-osm:
> Could the issue be, that current_num_changes counts only the objects where some values (tags/location/members) had really changed

Let's take a look at your second changeset: https://master.apis.dev.openstreetmap.org/changeset/437817

Since you have uploaded both relation and node in the osmChange message, both of them end up with a new version in that changeset:

image

Now, in the changeset metadata https://master.apis.dev.openstreetmap.org/api/0.6/changeset/437817 , we can see the number of changes as expected:

```
changes_count="2"
```

So the number of changes reflects any new/modified/deleted object version in that changeset. It doesn't really matter if you're uploading the exact same object again. The issue is non-deterministic and timing dependent, otherwise we would see this for many more changesets.

mmd-osm:
> >  On cgimap, there's an exclusive lock on the changeset id in place, which prevents this sort of issue.
> cgimap doesn't have single element writes.

Yes sure, what I meant is if you're sending multiple diff uploads in parallel.

tomhughes:
Here's the timestamp data for changeset 168535456:

```
   node_id   | version |         timestamp          
-------------+---------+----------------------------
 12980345202 |       3 | 2025-07-05 19:02:22.233214
 12979603740 |       4 | 2025-07-05 19:02:22.251147
 12980345203 |       3 | 2025-07-05 19:02:22.251773
 12979690624 |       4 | 2025-07-05 19:02:22.252439
 12979647498 |       4 | 2025-07-05 19:02:22.258134
 12979683712 |       4 | 2025-07-05 19:02:22.259855
 12980345205 |       3 | 2025-07-05 19:02:22.269461
 12979647497 |       4 | 2025-07-05 19:02:22.271039
 12980345207 |       3 | 2025-07-05 19:02:22.274816
 12980345206 |       3 | 2025-07-05 19:02:22.292598
 12979647499 |       4 | 2025-07-05 19:02:22.299024
 12980345204 |       3 | 2025-07-05 19:02:22.330371
 12980345201 |       3 | 2025-07-05 19:02:22.402109
 12980345216 |       3 | 2025-07-05 19:02:22.403848
 12979507064 |       5 | 2025-07-05 19:02:22.411536
 12979461759 |       5 | 2025-07-05 19:02:22.456349
 12980345213 |       3 | 2025-07-05 19:02:22.465104
 12980345226 |       3 | 2025-07-05 19:02:22.46558
 12980345219 |       3 | 2025-07-05 19:02:22.471514
 12980345214 |       3 | 2025-07-05 19:02:22.474112
 12980345208 |       3 | 2025-07-05 19:02:22.474654
 12980345220 |       3 | 2025-07-05 19:02:22.487285
 12980345222 |       3 | 2025-07-05 19:02:22.493207
 12980345218 |       3 | 2025-07-05 19:02:22.540268
 12980345225 |       3 | 2025-07-05 19:02:22.55978
 12980345224 |       3 | 2025-07-05 19:02:22.562974
 12980345230 |       3 | 2025-07-05 19:02:22.592606
 12980345211 |       3 | 2025-07-05 19:02:22.602377
 12980345215 |       3 | 2025-07-05 19:02:22.609225
 12980345227 |       3 | 2025-07-05 19:02:22.635611
 12980345229 |       3 | 2025-07-05 19:02:22.645984
 12980345223 |       3 | 2025-07-05 19:02:22.65584
 12979569450 |       5 | 2025-07-05 19:02:22.662778
 12980345235 |       3 | 2025-07-05 19:02:22.667434
 12979558472 |       5 | 2025-07-05 19:02:22.690501
 12980345231 |       3 | 2025-07-05 19:02:22.693952
 12980345232 |       3 | 2025-07-05 19:02:22.710277
 12980345221 |       3 | 2025-07-05 19:02:22.755382
 12979605020 |       4 | 2025-07-05 19:02:22.760826
 12980345236 |       3 | 2025-07-05 19:02:22.761379
 12979602450 |       5 | 2025-07-05 19:02:22.779572
 12980345234 |       3 | 2025-07-05 19:02:22.808984
 12980345228 |       3 | 2025-07-05 19:02:22.880571
 12979588716 |       3 | 2025-07-05 19:02:22.911363
 12980345239 |       2 | 2025-07-05 19:02:22.917784
 12979481737 |       5 | 2025-07-05 19:02:22.924431
 12980345238 |       2 | 2025-07-05 19:02:22.925123
 12980345237 |       2 | 2025-07-05 19:02:22.931746
 12979461757 |       4 | 2025-07-05 19:02:22.93335
 12979553510 |       3 | 2025-07-05 19:02:22.948512
 12980345212 |       3 | 2025-07-05 19:02:23.010192
 12980345233 |       3 | 2025-07-05 19:02:23.330984
(52 rows)
```

mmd-osm:
Thank you for the timestamp data. This clearly looks like a non-"cgimap single diff upload", otherwise all timestamps would be exactly the same.

Looking again at the source code, [this section](https://github.com/zbycz/osmapp/blob/master/src/services/osm/auth/editCrag.ts#L49-L54) may be relevant. It seems to be in place to update crag data, and the changeset description exactly matches [this code](https://github.com/zbycz/osmapp/blob/master/src/services/osm/auth/editCrag.ts#L43-L47):

```javascript
  const changesetId = await api.putChangeset(changesetXml);

  await Promise.all(
    changes.map((change) => saveCragChange(changesetId, change)),   // << here's where the server is flooded with requests
  );
  await api.putChangesetClose(changesetId);
```

saveCragChange in the end calls api.putOrDeleteItem, which is then performing [single object](https://github.com/zbycz/osmapp/blob/master/src/services/osm/auth/api.ts#L36-L50) operations, handled by Rails. So #6366 is spot on.

---

tomhughes:
It has to be said that parallel firing single object calls for every object you're changing is a pretty obnoxious thing to do and probably not something we anticipated but for a change of any size it will have seriously risked overwhelming the servers briefly.

mmd-osm:
By the way, the issue didn't really stop on July 15. This OSMCha query ([link](https://osmcha.org/?filters=%7B%22comment%22%3A%5B%7B%22label%22%3A%22routes%20%E2%80%A2%20%22%2C%22value%22%3A%22routes%20%E2%80%A2%20%22%7D%5D%2C%22editor%22%3A%5B%7B%22label%22%3A%22OsmAPP%201.7.0%22%2C%22value%22%3A%22OsmAPP%201.7.0%22%7D%5D%2C%22date__gte%22%3A%5B%7B%22label%22%3A%222025-07-01%22%2C%22value%22%3A%222025-07-01%22%7D%5D%7D)) shows relevant changesets, some of which are only a few hours old: https://www.openstreetmap.org/api/0.6/changeset/171025698

Since OSMCha filtering  tends to be a bit slow at times, https://www.openstreetmap.org/user/jvaclavik/history would also highlight the issue. The changeset description follows the pattern  "... routes • Edited ..."

zbycz:
Thanks for the investigation. You are right – OsmAPP still sends parallel requests in that specific scenario of uploading the crag data.

I had overlooked this part of code. I will fix it and keep you posted.

zbycz:
Fixed in https://github.com/zbycz/osmapp/pull/1329 - I have also bumped OsmAPP version to `1.7.1`, so we can debug it more easily if anything else occurs with the new code. There is one fork running at [map.et](https://map.et/), but AFAIK they are not using this part of the code.

All api calls to put/delete single items removed altogether, it was legacy code - now superseded by uploadDiff.

I also updated the wiki: https://wiki.openstreetmap.org/w/index.php?diff=2892499

gravitystorm:
So as I understand it, these counter discrepancies were (and still are, see #6379) coming from race conditions when handling changesets, which we are fixing separately.

I'm not super keen on huge SQL queries being run in production, without any unit tests! But at least this is only affecting the counter attributes, so the data can be recreated if something goes wrong. And given I'm not running it in production myself, then technically it's not my responsibility!

In future, I would be more comfortable with the queries and/or associated code being elsewhere (e.g. in a file in `lib/`) where it can be more easily tested (even just to catch syntax errors, or document expected behaviour) and the rake task just being a one-liner to call that code.

I'm going to merge this, but I haven't actually run it myself or verified that it correctly fixes incorrect `num_changes`, so caveat emptor!

PR #6363, opened 2025-08-28
Map display: handle deleted elements

### LINKED ISSUE #6362: Features could not be loaded: Cannot read properties of undefined
When I browse a deleted version of a node, I always have this error:
```
Load Data

Features could not be loaded: Cannot read properties of undefined (reading 'lat')
```

https://www.openstreetmap.org/node/13098917888/history/3?locale=en

Image

mmd-osm:
Fixes two map display issues:

* Exclude objects with visible=false
* Don't show error message if object is gone (HTTP 410).

Fixes #6362


tomhughes:
Looks good to me, thanks.

PR #6365, opened 2025-08-28
Move English language names to en.yml

### LINKED ISSUE #6319: Localize English captions in language picker
#6313 modified the language picker so that every language option has the English name below the native name. It’s always in English regardless of the user’s preferred language. This can be a very useful feature: you can easily see what the map looks like in another language like Greek, even if you don’t know that “Ελληνικά” means Greek in Greek. Unfortunately, this feature mostly only benefits users who speak some English. Speakers of other languages will only recognize the words if it’s similar to their own language.

Similar to #6127, we could instead display each language’s name in both the native name and the current locale using [`Intl.DisplayNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames). Different browsers have translations of different language names into different languages, so we’d need to fall back to either the English name or the raw language code:

```js
const displayNames = new Intl.DisplayNames(OSM.preferred_languages, { type: "language" });
let displayName = displayNames.of(:code);
if (displayName === :code) {
  displayName = language[:english_name];
}
```

As discussed in https://github.com/openstreetmap/openstreetmap-website/pull/6313#discussion_r2268803424, in the case where the user lacks a font for the current interface language, they probably will have a font for the language they speak. The dialog would continue to label that language with the native name, which they can still read.

AntonKhorev:
With this PR the language names in *Select Language* dialog can be translated, that would fix #6319.

Using CLDR or anything derived from it is not going to work because it's not complete enough. If you try, you'll have to "fall back to ... the raw language code" as #6319 says and why would you want that when we already have English names.

The problem this PR may introduce is that we won't be able to control how wide the names are. Some of them will likely won't fit on one line which will make the list look worse.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -195,7 +195,7 @@ body.small-nav {
 /* Rules for language selector */
 
 .select_language_list {
-  column-width: 160px;
+  column-width: 165px;
 
   a {
     .native_name small {
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -202,7 +202,7 @@ body.small-nav {
       font-size: 10px;
     }
 
-    &:hover .english_name {
+    &:hover .current_locale_name {
       text-decoration: underline;
     }
   }
--- a/app/views/layouts/_select_language_list.html.erb
+++ b/app/views/layouts/_select_language_list.html.erb
@@ -16,8 +16,8 @@
             <%= language[:native_name] %>
           <% end %>
         </span>
-        <small lang="en" class="english_name d-block text-secondary">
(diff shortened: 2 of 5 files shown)


1ec5:
Fortunately, as far as I can tell, every one of these strings has been translated for some project already on Translatewiki.net, so there should be translation memory for it.

> The problem this PR may introduce is that we won't be able to control how wide the names are. Some of them will likely won't fit on one line which will make the list look worse.

Is there a rule of thumb about how many characters long a translation should be (imprecise of course)? I can add that as context for translators in the /qqq messages on Translatewiki.net.

tomhughes:
Looks good to me, or as good as it can be given the potential width issues. Thanks.

PR #6366, opened 2025-08-28
Lock changeset in api single element changes

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/6361#issuecomment-3232959610.

I was easily able to trigger race conditions when creating nodes and ways but not when creating relations, modifying nodes and ways. That's why there are race condition tests only for create actions.

I assume we're locking the changeset first and the elements later, that's what cgimap seems to do.


tomhughes:
Presumably the diff upload should also lock the changeset?

tomhughes:
Let's merge this and we can always deal with the diff upload later.

PR #6373, opened 2025-09-01
Map API requests with Bearer token

mmd-osm:
In this PR, we're adding a Bearer token for Map API requests, if available.

Benefits:
- Users behind a shared IP address are less likely to hit rate limiting.
- Higher rate limits for moderators are taken into account.

Tested with anonymous / known user for "Map Data" and single object map display. Map requests were  served by a cgimap backend.


tomhughes:
Looks good to me, thanks.

PR #6377, opened 2025-09-03
Upgrade maplibre-gl dependency, fixing CJK labels

mapmeld:
Chinese map labels with the new vector tile / MapLibre GL layers haven't been rendering properly in the Chrome browser.
When the user sets their UI language to Traditional Chinese, these two labels should have these characters rendered as <span lang="zh">门</span> and <span lang="zh-Hant">北</span>:

image

In https://github.com/mapbox/tiny-sdf/pull/58 and  https://github.com/maplibre/maplibre-gl-js/pull/6223 we added support for MapLibre to pass the document language to Tiny-SDF, and Tiny-SDF sets the language in the canvas `ctx.lang` to render CJK characters.  So I think all that's left is upgrading `maplibre-gl` here. If we need to update yarn.lock too please let me know.

Note: in the MapLibre GL layers, switching the UI language changes which `name` tag we prefer. So we're rendering CJK characters in the user's language and NOT trying to render Hong Kong in Traditional Chinese, Beijing in Simplified Chinese, Tokyo in Japanese, etc.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -670,7 +670,7 @@ make-plural@*, make-plural@^7.4.0:
   resolved "https://registry.yarnpkg.com/make-plural/-/make-plural-7.4.0.tgz#fa6990dd550dea4de6b20163f74e5ed83d8a8d6d"
   integrity sha512-4/gC9KVNTV6pvYg2gFeQYTW3mWaoJt7WZE5vrp1KnQDgW92JtYZnzmZT81oj/dUTqAIu0ufI2x3dkgu3bB1tYg==
 
-maplibre-gl@^5.6.0:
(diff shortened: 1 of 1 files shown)


tomhughes:
I wonder why dependabot hasn't already suggested this? In any case you need to fix the lock file...

mapmeld:
oh! It looks like yarn.lock was mostly on 5.7.0 already and we just didn't have an upgrade in package.json

I had this PR queued up for when maplibre-gl was released and didn't notice it is fixed on the OSM main site already.

tomhughes:
Ah yes dependabot already bumped it in #6357.

PR #6379, opened 2025-09-04
Lock changesets during a diff upload

tomhughes:
Continues from #6366 by adding locking for diff uploads.

### DIFF
--- a/app/controllers/api/changesets/uploads_controller.rb
+++ b/app/controllers/api/changesets/uploads_controller.rb
@@ -28,11 +28,11 @@ module Api
       # Returns: a diffResult document, as described in
       # http://wiki.openstreetmap.org/wiki/OSM_Protocol_Version_0.6
       def create
-        changeset = Changeset.find(params[:changeset_id])
-        check_changeset_consistency(changeset, current_user)
-
-        diff_reader = DiffReader.new(request.raw_post, changeset)
         Changeset.transaction do
+          changeset = Changeset.lock.find(params[:changeset_id])
+          check_changeset_consistency(changeset, current_user)
+
+          diff_reader = DiffReader.new(request.raw_post, changeset)
           result = diff_reader.commit
           # the number of changes in this changeset has already been
           # updated and is visible in this transaction so we don't need
--- a/test/controllers/api/changesets/uploads_controller_test.rb
+++ b/test/controllers/api/changesets/uploads_controller_test.rb
@@ -198,6 +198,35 @@ module Api
         assert_dom "osmError>message", 1
       end
 
+      def test_upload_race_condition
+        user = create(:user)
+        changeset = create(:changeset, :user => user)
+
+        diff = <<~CHANGESET
+          <osmChange>
+            <create>
+              <node id='-1' lon='1' lat='2' changeset='#{changeset.id}'/>
+            </create>
+          </osmChange>
+        CHANGESET
+
+        auth_header = bearer_authorization_header user, :scopes => %w[write_api]
+        path = api_changeset_upload_path(changeset)
+        concurrency_level = 16
+
+        threads = Array.new(concurrency_level) do
+          Thread.new do
+            post path, :params => diff, :headers => auth_header
+          end
+        end
+        threads.each(&:join)
+
+        changeset.reload
+        assert_equal concurrency_level, changeset.num_changes
+        assert_predicate changeset, :num_type_changes_in_sync?
+        assert_equal concurrency_level, changeset.num_created_nodes
+      end
+
       # -------------------------------------
       # Test creating elements.
       # -------------------------------------


mmd-osm:
I think some more changes are needed here, because https://github.com/openstreetmap/openstreetmap-website/blob/059df79b21f8140bee9df5533c1825004a98e2b7/lib/diff_reader.rb#L20 reads the changeset without holding a lock (iirc, I can’t check the exact location right now).

Can you check in the rails logs if reading from changesets is always happening after a “select … for update”?

tomhughes:
Actually I think it's fine because `lock!` rereads the record (with `FOR UPDATE`) if it's not already part of a read so if `num_changes` has been changed in the database it will be updated at that point before we start incrementing it.

Possibly it would be better to move the transaction start earlier in the controller and take the lock as part of the find to avoid the re-read. It meant the diff object construction is inside the transaction but that's actually cheap as all the real work is in the `commit` method anyway.

tomhughes:
Then again arguably the `check_changeset_consistency` should be done after the lock is taken or it might get closed after the check and before we take the lock...

mmd-osm:
Here's the kind of change I've been testing locally. It moves both changeset consistency checking as well as DiffReader inside the exclusive changeset lock.

```ruby
diff --git a/app/controllers/api/changesets/uploads_controller.rb b/app/controllers/api/changesets/uploads_controller.rb
index ee9406d91..2afdd5e63 100644
--- a/app/controllers/api/changesets/uploads_controller.rb
+++ b/app/controllers/api/changesets/uploads_controller.rb
@@ -27,10 +27,10 @@ module Api
       # http://wiki.openstreetmap.org/wiki/OSM_Protocol_Version_0.6
       def create
         changeset = Changeset.find(params[:changeset_id])
-        check_changeset_consistency(changeset, current_user)
+        changeset.with_lock do
+          check_changeset_consistency(changeset, current_user)
 
-        diff_reader = DiffReader.new(request.raw_post, changeset)
-        Changeset.transaction do
+          diff_reader = DiffReader.new(request.raw_post, changeset)
           result = diff_reader.commit
           # the number of changes in this changeset has already been
           # updated and is visible in this transaction so we don't need
```

tomhughes:
I've updated it along those lines, but using `lock.find` inside an explicit transaction rather than `with_lock` to avoid reading the changeset twice.

pablobm:
Thoughts on including a test, similar to those in https://github.com/openstreetmap/openstreetmap-website/pull/6366?

I just had a look and the tests seem useful. Initially I couldn't get them to reproduce the issue on my Apple M1 Pro, but they do reproduce it consistently on my older, slower Thinkpad.

tomhughes:
I've added a test though I'm having trouble getting it to fail with the old code in the controller - possibly my machine is also too fast ;-)

pablobm:
I'm not convinced that this change fixes the issue :thinking:

I have been abusing the test (thank you for providing it!) and I can't get it to fail without the change. For example, I expected that adding a random sleep after `Changeset.find` (pre-change) would make the test fail, but it doesn't.

In deeper inspection, I see that the changeset **object** passed to `DiffReader` is not the same one whose counts are increased. As in: it represents the same changeset _conceptually_, but it is not the same object in memory.

When `DiffReader#commit` calls `Node.from_xml_node`, the created node gets `node.changeset_id = pt["changeset"].to_i`. ActiveRecord creates a new `Changeset` in memory at this point. This one has the correct counts, and they remain correct because we are already inside a transaction. When we get to `changeset.num_created_nodes += 1` (in `Node#create_with_history`), the increase will be correct.

Admittedly, my understanding of ACID is not great so I might be missing something. Thoughts @tomhughes?

It might still be worth merging this and see if it helps, but I suspect we'll continue seeing the issue.

mmd-osm:
I don't know what Rails is doing with different object instances, but from a database point of view this looks ok:

The important thing is that we start with SELECT ... FOR UPDATE on the changesets table before reading anything from that table without holding a row level db lock.

image

tomhughes:
Yes @mmd-osm says it doesn't really matter how many times the changeset is loaded at the ruby level - what matters is that we take a lock in the database before making any changes to it in the database.

Each object change will then read the changeset, update it and write it back, but as that's all inside the transaction each subsequent read will see the dirty data from previous writes inside the uncommitted transaction so as those events are sequential there is no risk there.

The lock will be released when the transaction is committed and any other diff upload (or single object change) that tries to run will have to wait for the lock to release before it can get it's own lock and start making changes.

pablobm:
Thank you, that makes sense :+1: 

One question: does the `FOR UPDATE` need to be in a transaction? Reading the Postgres docs I'm understanding yes, and that's why the `.lock` is put within `Changeset.transaction`. But if this wasn't within a transaction, what would be the behaviour? I see that `FOR UPDATE` still happens, but I wonder how log the lock lasts for in that case.

tomhughes:
If you're not in an explicitly started transaction then each statement is executed in a transaction that is then immediately committed so the lock would release at the end of the statement. See https://www.postgresql.org/docs/17/sql-begin.html:

> By default (without BEGIN), PostgreSQL executes transactions in “autocommit” mode, that is, each statement is executed in its own transaction and a commit is implicitly performed at the end of the statement (if execution was successful, otherwise a rollback is done).

pablobm (approved):
Sounds good! Thank you all for the explanations.

gravitystorm:
I'm not very good with database locks, but I have two questions:

```ruby
changeset = Changeset.lock.find(params[:changeset_id])
```

Is this different from `changeset = Changeset.find(params[:changeset_id]).lock!` ? As far as I can tell from the docs, it's the same thing, but we use the `.lock!` variant elsewhere. If they are indeed identical, then I'd prefer to use the `.lock!` variant here, since that makes slightly easier to understand that it's the changeset being locked, and not e.g. the whole `Changeset` table.

Second question, in #6366 the "safe navigation operator" was used, e.g. `node.changeset&.lock!`. I found that surprising, since taking a lock is generally not optional! Is this deliberate, or is it a side-effect of taking the lock too early on e.g. should the lock be taken after the code that validates the changeset exists? Or is there something else there about locks that I don't understand?

tomhughes:
Putting `lock` before `find` avoids doing two reads because `find` realises the query and returns a changeset so calling `lock!` then has to do a new select to acquire the lock. Putting `lock` before the `find` just adds it as a modifier on the AREL object that is then executed by `find`.

I think the point of the safe navigation operator in the other cases is to avoid raising an exception if there was a bogus changeset ID in the XML that doesn't exist - that will then get properly detected when creating/updating the object by the consistency checks.

gravitystorm:
Thanks for the extra details! Merged.

PR #6384, opened 2025-09-07
Remove kgio gem

rkoeze:
### Description
The kgio gem was added [in this commit](https://github.com/openstreetmap/openstreetmap-website/commit/255b0b94252b48a9fe4326bb9d9fc4f184308460) as a dependency of dalli. However, [dalli no longer needs it](https://github.com/petergoldstein/dalli/blob/c6d85e4e4b440b22bf4a784eca77b234798553e5/3.0-Upgrade.md?plain=1#L17). See [this PR](https://github.com/petergoldstein/dalli/pull/721) for removal.

### How has this been tested?
I was able to build and run the app locally.  The tests pass.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -109,7 +109,6 @@ gem "rotp"
 # Load memcache client in case we are using it
 gem "connection_pool"
 gem "dalli"
-gem "kgio"
 
 # Load canonical-rails to generate canonical URLs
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #6360, opened 2025-08-27
Make GitHub auth button invertible

AntonKhorev:
Continuing #6351 with GitHub logo.

image
image

PR #6385, opened 2025-09-07
Safari-compatible auth provider images

mmd-osm:
As discussed in #6360, Github and Apple icons don't render properly on Safari, because `filter="invert(100%)" ` and `filter="contrast(82%)"` are not supported. As a result, the icons are shown as black (or in dark mode: white) boxes.

image

In this PR, I'm proposing a new approach, which renders both icons as expected on Safari. Also tested on Firefox, Chrome, Edge, Librewolf:

image

image

I added an inline style element to both svg files, which, due to CSP, requires a nonce to be passed along. The only purpose of the style is to define background and path colors for both color schemes.

```
  <style>
    .apple-logo rect { fill: black; }
    .apple-logo path { fill: white; }
    @media (prefers-color-scheme: dark) {
      .apple-logo rect { fill: white; }
      .apple-logo path { fill: black; }
    }
  </style>
```

I think moving the style definition to common.scss should also work, if inline styles are not the preferred approach.


tomhughes:
I think a class and a rule in the stylesheet would be preferable really.

mmd-osm:
Agree, it's probably easier to manage everything in the stylesheet. I moved the styling to common.scss, and added respective class names to the svg files.

tomhughes:
Did you find some documentation somewhere that those filters don't work in Safari? Only caniuse seems to suggest they should:

https://caniuse.com/mdn-css_types_filter-function_invert
https://caniuse.com/mdn-css_types_filter-function_contrast

Possibly it's because this is SVG though? Is that filter there a CSS filter? I have no idea...

mmd-osm:
I found an open WebKit issue which suggests that CSS filters are only working when applied to an SVG root element. In our case, we were applying them to path/rect elements, which doesn't seem to be supported.

https://bugs.webkit.org/show_bug.cgi?id=246106  (in particular comments 4 + 5).

app/assets/images/auth_providers/github.svg
```diff
@@ -1,4 +1,4 @@
...
+<svg xmlns="http://www.w3.org/2000/svg" width="36" height="36" viewBox="0 0 36 36" fill="currentColor" fill-rule="evenodd" class="github-logo">
+  <rect width="36" height="36"/>
+  <path d="M0,0 V36 H36 V0 Z M18,4a14,14,0,0,0-4.41,27.23c.7.13.93-.3.93-.66V28.23c-3.88.85-4.66-1.86-4.66-1.86a3.78,3.78,0,0,0-1.57-2.05c-1.26-.87.11-.85.11-.85a2.92,2.92,0,0,1,2.13,1.45,3,3,0,0,0,4,1.17h0a3,3,0,0,1,.89-1.87c-3.11-.35-6.37-1.55-6.37-6.91a5.43,5.43,0,0,1,1.43-3.72,5.06,5.06,0,0,1,.14-3.73s1.17-.37,3.85,1.43a13.2,13.2,0,0,1,7,0c2.67-1.85,3.85-1.46,3.85-1.46a5.09,5.09,0,0,1,.14,3.74,5.44,5.44,0,0,1,1.44,3.72c0,5.36-3.28,6.52-6.38,6.89a3.37,3.37,0,0,1,.94,2.56v3.83c0,.45.26.81.93.67A14,14,0,0,0,18,4Z"/>
```
hlfan, line 3: Why reinvent the wheel?
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" width="36" height="36" viewBox="0 0 36 36" fill-rule="evenodd">
  <rect width="36" height="36" fill="var(--bs-carousel-indicator-active-bg)"/>
  <path d="M0,0 V36 H36 V0 Z M18,4a14,14,0,0,0-4.41,27.23c.7.13.93-.3.93-.66V28.23c-3.88.85-4.66-1.86-4.66-1.86a3.78,3.78,0,0,0-1.57-2.05c-1.26-.87.11-.85.11-.85a2.92,2.92,0,0,1,2.13,1.45,3,3,0,0,0,4,1.17h0a3,3,0,0,1,.89-1.87c-3.11-.35-6.37-1.55-6.37-6.91a5.43,5.43,0,0,1,1.43-3.72,5.06,5.06,0,0,1,.14-3.73s1.17-.37,3.85,1.43a13.2,13.2,0,0,1,7,0c2.67-1.85,3.85-1.46,3.85-1.46a5.09,5.09,0,0,1,.14,3.74,5.44,5.44,0,0,1,1.44,3.72c0,5.36-3.28,6.52-6.38,6.89a3.37,3.37,0,0,1,.94,2.56v3.83c0,.45.26.81.93.67A14,14,0,0,0,18,4Z" fill="var(--bs-emphasis-color)"/>
```

mmd-osm: Have you tried if this is working on WebKit?

hlfan: Yup, via [browserling](https://browserling.com/), so I only tested it on Safari 18 on macOS 15. But I don't see any reason for this not to work on other systems.

mmd-osm: Ok, sounds good. Let's go for the easier way. I will update the Apple logo later on and remove the obsolete scss changes.

tomhughes:
Is tying this to some bootstrap colour variables that just happen to be the colour we want but which don't logically have anything to do with this really a good idea? What happens if bootstrap changes the emphasis colour in a future release...

mmd-osm:
Good point. Applying some bootstrap themes would probably also screw up logo colors. Reverted to previous approach for now.

tomhughes:
Thanks - lets go with that for now at least!

PR #6391, opened 2025-09-12
Add shn to UI languages

### LINKED ISSUE #6390: A new locale has broken the build
[An update from translatwiki](https://github.com/openstreetmap/openstreetmap-website/commit/d105511dbd299740bf8c2f97f008db60611af73f) yesterday included a new locale, [`shn`](https://github.com/openstreetmap/openstreetmap-website/blob/master/config/locales/shn.yml).

We have [a test](https://github.com/openstreetmap/openstreetmap-website/blob/d618907d6237e940dc092df86856f25e8ce58048/test/lib/i18n_test.rb#L74-L76) that checks that every locale in `config/locales` is also in `config/ui_languages.yml`, and this therefore failed on the new locale.

We should
* Add `shn` to `config/ui_languages.yml`
* (optional) Add a longer description to the test output, to explain how to resolve the issue and in particular, where to find the correct values for the `native_name` attribute.

tomhughes:
Fixes #6390.

Name is taken from https://en.wikipedia.org/wiki/Shan_language#Names.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1863,6 +1863,7 @@ en:
         kn: Kannada
         th: Thai
         my: Burmese
+        shn: Shan
         xmf: Mingrelian
         ka: Georgian
         km: Khmer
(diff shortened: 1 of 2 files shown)


gravitystorm:
Merged, thanks!

PR #6392, opened 2025-09-13
Fix some language names

amire80:
I'm a translatewiki administrator, and I noticed that some language names should be fixed.

* arz: The name "Arabic (Egypt)" looks more like a local variant of Standard Arabic for Egypt, but arz is actually a different language and not Standard Arabic.
* skr-arab: Saraiki is always written in the Arabic script. It uses the code "skr-arab" on translatewiki for historical and technical reasons, and some day it may actually move to just "skr". There is no reason to mention "(Arabic)" in parentheses.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1846,9 +1846,9 @@ en:
         yi: Yiddish
         he: Hebrew
         ar: Arabic
-        skr-arab: Saraiki (Arabic script)
+        skr-arab: Saraiki
         fa: Persian
-        arz: Arabic (Egypt)
+        arz: Egyptian Arabic
         pnb: Western Punjabi
         ps: Pashto
         nqo: N’Ko


tomhughes:
The reason for abbreviation will probably be to avoid the name being too long which in turn causes problems with the multi-column formatting of the language selector.

amire80:
> The reason for abbreviation will probably be to avoid the name being too long which in turn causes problems with the multi-column formatting of the language selector.

OK, no problem, I restored it.

tomhughes:
Thanks - can you squash that fix into the original commit please and then we can merge this.

amire80:
I can, but isn't it easier to squash and merge from the GitHub UI?

tomhughes:
We like to record all merges so we don't normally use rebase type merges that lead to a fast forward - in fact I normally do all my merges offline and not using the github UI at all.

Also in this case the commit message really needing changing to reflect the change.

I've one the squash now anyway, and merged this. Thanks for the work.

PR #6393, opened 2025-09-14
Remove logstash support

tomhughes:
This is no longer used in production...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -115,9 +115,6 @@ gem "dalli"
 # Load canonical-rails to generate canonical URLs
 gem "canonical-rails"
 
-# Used to generate logstash friendly log files
-gem "logstasher"
-
 # Used to generate images for traces
 gem "bzip2-ffi"
 gem "ffi-libarchive"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -383,9 +383,6 @@ GEM
       rb-fsevent (~> 0.10, >= 0.10.3)
       rb-inotify (~> 0.9, >= 0.9.10)
     logger (1.7.0)
-    logstasher (2.1.5)
-      activesupport (>= 5.2)
-      request_store
     loofah (2.24.1)
       crass (~> 1.0.2)
       nokogiri (>= 1.12.0)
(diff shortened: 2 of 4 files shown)


gravitystorm:
Happy to merge, but there's a Gemfile.lock conflict now. You'll either need to rebase this, or teach me how to handle these situations cleanly.

(I don't like merging and and fixing things up locally, in case I mess up the lockfile, but if there's a way to do it that updates the PR, keeps credit for the original author, and also runs CI, then I'm all ears :smile: )

tomhughes:
Updated... Generally fixing `Gemfile.lock` conflicts just means deleting the conflicts and running `bundle install` to fix things I think?

gravitystorm:
Thanks, merged.

> Generally fixing `Gemfile.lock` conflicts just means deleting the conflicts and running `bundle install` to fix things I think?

I mean yes, that's how I've done it before (roughly `git merge`, manually fix conflicts, run `bundle install` again just in case, `git add`, `git merge --continue`, `git push upstream master`, which means my lockfile fixes are hidden in the merge commit. But it also means that I push the merge commit to master without that having gone through CI, and that makes me nervous.

Whereas if there's a way to check out the branch locally, twiddle stuff (e.g. `git rebase` + lockfile cleanup), and then push the new commit back to the PR, I could let CI run and then use the Github UI to do the merge when CI is green. I *think* I've done this before (and managed to preserve the commit authorship). But I also know it doesn't work if the PR submitter has unticked the "let maintainers write to this branch" option, and it even if that's ticked it can be a faff to set up the correct remote to push to since you can't (afaik) push to the `pull/1234` branch on origin, you have to set up and push to the `someone` remote, with whatever branch they've used to make the PR. And I've messed that up too in the past, by pushing the wrong thing to the branch on someone else's repo, which is a bit embarrassing.

So if there's an easier way to do option 2, with less faffing with the various `remotes`, or a different way entirely, then that would be great.

tomhughes:
You don't actually have to create a remote you can just do `git push -f <url> <local-branch>:<remote-branch>`.

HolgerJeromin:
Another possibility would be:
* Create locally a new branch from master.
* Cherry pick the PRs commit into it.
* Fix the code.
A) create an own fixing commit or
B) Amend the changes to the last (cherry picked) commit.
* Create a new own PR. Close original PR.

Version A pretends the original author did no mistake
while B keeps you the autorship of the fix.

tomhughes:
There was no "mistake" here, just a merge conflict. It's completely normal for the person merging to fix those as part of the merge and the merge commit (where the fix is) will already be attributed to the person doing the merge.

PR #6395, opened 2025-09-17
we are mapping not only significant buildings

matkoniecz:
### Description

fixed legend - we are mapping not only significant buildings

triggered by https://community.openstreetmap.org/t/the-simpler-the-better/135578/29

### How has this been tested?

untested change, but hopefully text-only change is trivial enough

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2868,7 +2868,7 @@ en:
         school: "School"
         university: "University"
         hospital: "Hospital"
-        building: "Significant building"
+        building: "Building"
         station: "Railway station"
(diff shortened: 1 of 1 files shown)


1ec5:
Oh, I should’ve clarified in my post that the label “Significant building” remains from the old days, but the legend entry these days is actually about the darker fill that OSM Carto assigns to certain `building=church`/`train_station`/etc., so it’s accurate. I guess the purpose of this entry changed in #4396, which updated the legend from Carto’s previous color scheme that gave all buildings a more reddish color.

matkoniecz:
that is what I got for not testing

Issue #2025, opened 2018-10-12
Increase may key and value max length


westnordost:
Some OSM tags do not (always) fit into the 255-character-limit imposed by the OSM API. This is a feature request to either increase the limit or remove it altogether.

### Discussion

There is a [discussion in the tagging mailing list right now](https://lists.openstreetmap.org/pipermail/tagging/2018-October/039889.html) about how to deal with a very complicated opening hours which is too long to fit into one value.

Citing @simonpoole :
> We have a number of keys for which the values can easily exceed 255 chars besides opening_hours, lane destinations and conditional restrictions are good candidates. Not to mention changeset tags. With other words it is a general problem which should be tackled with a general solution.

So, we passed the point where value-lengths above 255 characters are unrealistic for structured data, data that is not prose.

### Solutions

In the above discussion, the general opinion seemed to be that the OSM API will not change (until the unicorny API 0.7) and thus we have to workaround this.

There have been some suggestions towards a general-purpose syntax splitting up the actual value in different keys, like `opening_hours`, `opening_hours#2`, ... I think this is a very bad solution, because it requires **each and every** application and library that works with OSM data to implement this kind of tag concatenation - for any key, not just `opening_hours`, because this kind of enumeration scheme is supposed to be a generic solution

In #1593, I read that the limitation to 255 characters is actually a legacy from back when the OSM database was running on MySQL. So, removing that limitation, or increasing it somewhat, seems to be the swiftest solution because a weird workaround is avoided and it doesn't require a change in applications that use OSM data.
The only change that should be done would be to remove/change the client-side checks for tag lengths (editors).

tomhughes:
I think the answer to how to store complicated opening hours is not to store opening hours in OSM because it's insane to store something as volatile and non-geo related as that ;-)

bhousel:
From my perspective, it would be super to just increase the tagvalue length in the database to something like varchar(2000), bump the API minor version, and move on.  Inventing tags to workaround the existing limit is just asking for trouble.

tomhughes:
Well you might as well get rid of it altogether because any length limit in postgres is entirely arbitrary - the storage for `char(N)`, `varchar(N)` and `text` is all identical - the first two just do a length check on insert is all.

That said such large values is really stretching the basic concept of tags if you ask me, especially for the name.

bhousel:
[...]

Issue #1593, opened 2017-07-23
Add tag key and value max length to capabilities API call


simonpoole:
See https://github.com/MarcusWolschon/osmeditor4android/issues/590 will document and create a PR if nobody beats me to it when i'm back from vacation.

simonpoole:
Joting this down mainly for my benefit.

The root cause here seems to be a railsism that the default string length in rails is 255 characters that is mapped to character varying(255) (in PGSQL) , the down side of this is that the code is peppered with explicit references to 255 (range checking etc). From an API pov this effects at least tag key, value lengths on objects and changesets and role values for relation members. 

IM(very)HO I would refactor the code to reference a configuration value for the string length which means adding explicit limit values where necessary, this would touch a fair amount of code. One could argue that this a lot of work just so that we can actually return a max value that could in principle be set different per rails-port instance. 

@gravitystorm and @tomhughes any opinion?

tomhughes:
I don't understand what your goal is - are you running an alternate installation of the code that you want to have a different limit?

The limit is almost certainly accidental and/or arbitrary. Certainly with postgres there is really no advantage to specifying a limit in the type - the on disk representation for `text` and `varchar(N)` is identical. All the latter form does is make the database police a maximum length.

smsm1:
The history of the limit is from the MySQL days, and it was decided that keeping the limit would be best for consistency, and also to prevent people from writing an essay in a key/value pair. 

It's reasonable to hard code this value in the apps as it's not a config that will very unlikely change, it's been the same value since probably the beginning of OSM, so the likelihood of it changing is low.

simonpoole:
@smsm1 even if the value never changes there is (well in every programming language that I'm reasonably fluent in, which doesn't include rails) a programming style issue in that there are numerous references to 255 sprinkled around the code instead of it being a constant.

In any case I've documented the limit now: http://wiki.openstreetmap.org/wiki/API_v0.6#Maximum_string_lengths and while 255 would seem to be shortish for some of the strings (changeset tags at least), if nobody feels particularly strong about the issue, there is no point in undertaking the effort.

smsm1:
I agree a constant of some form would be a more consistent, and if I was implementing the code from scratch, would use a constant instead. 

I thought the max string length had already been documented on the API pages.

HarelM:
Continuing from issue #1702, any chance to change this limit?
[...]

PR #6399, opened 2025-09-19
Allow `user/preferences` values up to 10_000

### LINKED ISSUE #6297: Increase character limit for OSM user preference API values
### Problem

More and more OSM applications use the OSM user preferences API to store user specific settings.
Docs: https://wiki.openstreetmap.org/wiki/API_v0.6#Preferences_of_the_logged-in_user:_GET|PUT|DELETE_/api/0.6/user/preferences

This API allows to store key-value pairs that can be accessed by all editors that are allowed to read the preferences.

The problem is, that the character limits for this preferences values is the same as for OSM tags (255 characters).

This makes storing preferences harder than it needs to be, because apps need to split the values into multiple "tags" and then come up with a good key-naming schema to later merge those keys again.

Apps that – to my knowledge – struggle with the current setup are…
- [MapComplete](https://github.com/pietervdvn/MapComplete) @pietervdvn
- [Overpass turbo](https://github.com/tyrasd/overpass-turbo) @tyrasd 
- And iD in the near future

Given that the user preferences API is not directly related to the map data key-value system, I propose to increase the limit for this API and allow a lot more characters.

IMO this can be done without a version increase given that we can expect apps to only load their own keys and they will know what they put into them.

---

Just to be clear, this issue is only about the user/preferences endpoint, not a general increase in the limit, which is discussed in https://github.com/openstreetmap/openstreetmap-website/issues/2025 

### Description

_No response_

### Screenshots

_No response_

tordans:
### Description

Fixes https://github.com/openstreetmap/openstreetmap-website/issues/6297

The goal is to increase the limit of the `user/preferences` API values.
~I picket 10k which is what we use for user blocks and should be enough to add more complex data.~
_Update:_ [After more discussion](https://github.com/osmlab/osm-api-js/issues/32#issuecomment-3660625648), I change the limit to 1_000_000. A lower limit would still mean that many of the use cases we have in mind would require workarounds to split the data into chunks which would make this unreasonable hard to work with.

<details><summary>Other limits</summary>
<p>

User descriptions can be up to 65,536 characters
Diary comments can be up to 65,536 characters
User block reasons can be up to 10,000 characters
Note comments can be up to 2,000 characters

</p>
</details> 

### How has this been tested?

I tested it with the demo page from https://gist.github.com/tordans/94481e6336e8c34f258c677350fc2e77 which uses https://github.com/osmlab/osm-api-js to authenticate against localhost 3001 (hard coded) with a local oAuth App (hardcoded) and then runs two test cases against the preferences endpoint.

### DIFF
--- a/app/models/user_preference.rb
+++ b/app/models/user_preference.rb
@@ -17,7 +17,8 @@ class UserPreference < ApplicationRecord
   belongs_to :user
 
   validates :user, :associated => true
-  validates :k, :v, :length => 1..255, :characters => true, :presence => true
+  validates :k, :length => 1..255, :characters => true, :presence => true
+  validates :v, :length => 1..1_000_000, :characters => true, :presence => true
 
   scope :color_schemes, -> { where("k LIKE '%.color_scheme'") }
 end
--- a/app/views/api/capabilities/show.builder
+++ b/app/views/api/capabilities/show.builder
@@ -14,6 +14,8 @@ xml.osm(OSM::API.new.xml_root_attributes) do |osm|
                    :maximum_query_limit => Settings.max_changeset_query_limit)
     api.notes(:default_query_limit => Settings.default_note_query_limit,
               :maximum_query_limit => Settings.max_note_query_limit)
+    api.user_preferences(:key_maximum_length => 255,
+                         :value_maximum_length => 1_000_000)
     api.timeout(:seconds => Settings.api_timeout)
     api.status(:database => @database_status,
                :api => @api_status,
--- a/app/views/api/capabilities/show.json.jbuilder
+++ b/app/views/api/capabilities/show.json.jbuilder
@@ -31,6 +31,10 @@ json.api do
     json.default_query_limit Settings.default_note_query_limit
     json.maximum_query_limit Settings.max_note_query_limit
   end
+  json.user_preferences do
+    json.key_maximum_length 255
+    json.value_maximum_length 1_000_000
+  end
   json.timeout do
     json.seconds Settings.api_timeout
   end
--- a/test/controllers/api/capabilities_controller_test.rb
+++ b/test/controllers/api/capabilities_controller_test.rb
@@ -42,6 +42,7 @@ module Api
           assert_select "notes" \
                         "[default_query_limit='#{Settings.default_note_query_limit}']" \
                         "[maximum_query_limit='#{Settings.max_note_query_limit}']", :count => 1
+          assert_select "user_preferences[key_maximum_length='255'][value_maximum_length='1000000']", :count => 1
           assert_select "status[database='online']", :count => 1
           assert_select "status[api='online']", :count => 1
           assert_select "status[gpx='online']", :count => 1
@@ -68,6 +69,8 @@ module Api
       assert_equal Settings.max_number_of_relation_members, js["api"]["relationmembers"]["maximum"]
       assert_equal Settings.default_note_query_limit, js["api"]["notes"]["default_query_limit"]
       assert_equal Settings.max_note_query_limit, js["api"]["notes"]["maximum_query_limit"]
+      assert_equal 255, js["api"]["user_preferences"]["key_maximum_length"]
+      assert_equal 1_000_000, js["api"]["user_preferences"]["value_maximum_length"]
       assert_equal "online", js["api"]["status"]["database"]
       assert_equal "online", js["api"]["status"]["api"]
       assert_equal "online", js["api"]["status"]["gpx"]
--- a/test/models/user_preference_test.rb
+++ b/test/models/user_preference_test.rb
@@ -14,32 +14,45 @@ class UserPreferenceTest < ActiveSupport::TestCase
     assert_raise(ActiveRecord::RecordNotUnique) { new_up.save }
   end
 
-  def test_check_valid_length
-    key = "k"
-    val = "v"
-    [1, 255].each do |i|
-      up = build(:user_preference)
-      up.user = create(:user)
-      up.k = key * i
-      up.v = val * i
-      assert_predicate up, :valid?
-      assert up.save!
-      resp = UserPreference.find(up.id)
-      assert_equal key * i, resp.k, "User preference with #{i} #{key} chars (i.e. #{key.length * i} bytes) fails"
-      assert_equal val * i, resp.v, "User preference with #{i} #{val} chars (i.e. #{val.length * i} bytes) fails"
-    end
+  def test_key_length_valid
+    up = build(:user_preference)
+    up.user = create(:user)
+    up.k = "k" * 255
+    up.v = "v"
+    assert_predicate up, :valid?
+    assert up.save!
+    resp = UserPreference.find(up.id)
+    assert_equal "k" * 255, resp.k, "User preference with 255 k chars fails"
+    assert_equal "v", resp.v
   end
 
-  def test_check_invalid_length
-    key = "k"
-    val = "v"
-    [0, 256].each do |i|
-      up = build(:user_preference)
-      up.user = create(:user)
-      up.k = key * i
-      up.v = val * i
-      assert_not_predicate up, :valid?
-      assert_raise(ActiveRecord::RecordInvalid) { up.save! }
-    end
+  def test_key_length_invalid
+    up = build(:user_preference)
+    up.user = create(:user)
+    up.k = "k" * 256
+    up.v = "v"
+    assert_not_predicate up, :valid?, "Key should be too long"
+    assert_predicate up.errors[:k], :any?
+  end
+
+  def test_value_length_valid
+    up = build(:user_preference)
+    up.user = create(:user)
+    up.k = "k"
+    up.v = "v" * 1_000_000
+    assert_predicate up, :valid?
+    assert up.save!
+    resp = UserPreference.find(up.id)
+    assert_equal "k", resp.k
+    assert_equal "v" * 1_000_000, resp.v, "User preference with 1000000 v chars fails"
+  end
+
+  def test_value_length_invalid
+    up = build(:user_preference)
+    up.user = create(:user)
+    up.k = "k"
+    up.v = "v" * 1_000_001
+    assert_not_predicate up, :valid?, "Value should be too long"
+    assert_predicate up.errors[:v], :any?
   end
 end


app/models/user_preference.rb
```diff
@@ -17,5 +17,6 @@ class UserPreference < ApplicationRecord
   belongs_to :user
 
   validates :user, :associated => true
-  validates :k, :v, :length => 1..255, :characters => true
+  validates :k, :length => 1..255, :characters => true
+  validates :v, :length => 1..10_000, :characters => true
```
pablobm: Since this is intended to store arbitrary, computer-readable data: does it make sense to have this restriction to `characters`? Would it make sense to allow any binary bytes?

pnorman: Preferences have to be valid Unicode or else they can't be represented in the XML sent to the client

pablobm: Got it. Thank you! 👍

simonpoole:
@tordans this should definitely be accompanied by a change to the capabilities api call that reports the max value (and given that it is related, should probably report the standard 255 limit too, see https://github.com/openstreetmap/openstreetmap-website/issues/1593).

tordans:
> this should definitely be accompanied by a change to the capabilities api call that reports the max value

Good point. I added changes that expose this limit like so…
`http://localhost:3001/api/0.6/capabilities.json`

```
{
  "version": "0.6",
  "api": {
    "version": {
      "minimum": "0.6",
      "maximum": "0.6"
    },
    "area": {
      "maximum": 0.25
    },
    "note_area": {
      "maximum": 25
    },
    "tracepoints": {
      "per_page": 5000
    },
    "waynodes": {
      "maximum": 2000
    },
    "relationmembers": {
      "maximum": 32000
    },
    "changesets": {
      "maximum_elements": 10000,
      "default_query_limit": 100,
      "maximum_query_limit": 100
    },
    "notes": {
      "default_query_limit": 100,
      "maximum_query_limit": 10000
    },
    "user_preferences": {
      "key_maximum_length": 255,
      "value_maximum_length": 10000
    },
    "timeout": {
      "seconds": 300
    },
    "status": {
      "database": "online",
      "api": "online",
      "gpx": "online"
    }
  },
  "policy": {
    "imagery": {
      "blacklist": []
    }
  }
}
```

> (and given that it is related, should probably report the standard 255 limit too, see https://github.com/openstreetmap/openstreetmap-website/issues/1593).

Lets tackle this separate to keep this PR as small as possible. What would a good format be? Just `defaults: { key_max…, value_max…}`?

1ec5:
> What would a good format be? Just `defaults: { key_max…, value_max…}`?

Seems reasonable to me. Have we ever documented the 255-character limit as a hard limit? That could influence whether we need to expose the capabilities info now or if it can wait as tail work.

simonpoole:
> Have we ever documented

_Have_, _we_, _ever_ and _documented_ are doing some very heavy lifting there, it's _yes_ as in the only official documentation is the code, or/and _no_ as in it has never been exposed in the capabilities call, see the issue I linked above https://github.com/openstreetmap/openstreetmap-website/issues/1593

https://github.com/openstreetmap/openstreetmap-website/issues/2025 is likely at least related.

tordans:
Update: I rebased the PR and change the limit to 1_000_000. See the updated PR description for more. Ping https://github.com/osmlab/osm-api-js/issues/32#issuecomment-3660625648

simonpoole:
Why are we considering substantially changing the API for a project that hasn't even been discussed with other editing app devs outside of @tordans inner circle?,

gravitystorm:
> I rebased the PR

Thanks.

> change the limit to 1_000_000

This needs some substantial and convincing supporting use-cases. A million UTF-8 characters is an awful lot of text.

The UserPreferences system is designed for storing key-value pairs, not for storing either entire works of Shakespeare or arbitrary binary blobs masquerading in a pseudo UTF-8 character encoding.

Storing ~1MB of data in a single value has some serious knock-on effects. You can't download part of a value, so every client that needs any iota of the information stored in that value needs to download the entire value (think e.g. mobile connectivity). Every upload has the same problem - you can't upload part of a value. If we start storing ~MB values, someone will then ask for range requests within the value, or to paginate the values, or something similar. 

```
GET /api/0.6/user/preferences/my_key_name?character_range=600000-600010 # n.b. not a serious proposal
```

Also, if you need ~MB values in keys, and multiple apps using multiple keys start storing multi-megabytes of "preferences", then the base `GET /api/0.6/user/preferences` call becomes unusable for most clients.

If you're storing structured data hidden within a value of a key, which is itself part of a structured document, that's some inception-nesting and again suggests this isn't a good idea.

So let's see some use-cases, and figure out a better option. I don't want to say that the current 255 limit is unchangeable, but the limit exists for various reasons, and aren't just there to get in the way or to be worked around.

tordans:
Thanks for your reply, @gravitystorm!

> So let's see some use-cases, and figure out a better option.

The idea behind this change is to make it very easy for editors and OSM apps to **share mapping-related preferences**.

The current limit of 255 makes it very hard to use the API for this use case. There are a few apps that work around this limitation by splitting values into multiple keys. But this is very error-prone and only works well for one app. The advantage of the shared user/preferences data is, however, that all apps can access them, so sharing should be easy.

**Use Case Examples:**

* **Shared custom background layers:**  
   In GoMap I can set up custom background layers. In iD I can do the same. In MapRoulette I have settings to set up custom background layers as well. All those are stored separately. (And most editors have similar features.)  
   My goal is to create a schema that apps can use to read and write to this user-specific list of background layers. I started with a draft on this in https://github.com/tordans/editor-settings-schema under "layers-schema".  
   The character limit should be high enough to allow for at least 10 of those, so a limit of ~5k or ~10k would be ok.

* **Shared custom presets:**  
   In GoMap I can create custom presets, and iD is planning to add those in the future. Other editors already have custom presets. The user/preferences could be the place where we share those between editors.  
   I started with a draft on this in https://github.com/tordans/editor-settings-schema under "presets-schema".  
   The character limit should be high enough to allow for at least 50 of those, so a limit of ~30k or ~50k would be ok.

* **Shared custom fields:**  
   This is similar to the shared presets but adds fields to presets; the story is the same as above.

* **Shared custom data layers:**  
   In GoMap I can upload custom data (GPX, GeoJSON) that I can use to review OSM data. For example, a list of custom TODOs of places that I need to survey and what to look at. In iD I can use similar files as an external data source.  
   This feature could evolve in two ways:  
   Either as a list of sources where the data needs to be stored somewhere and the apps download them. This is what is described in https://github.com/tordans/editor-settings-schema/blob/main/data-layers-schema/README.md. The character limits should be similar to "Shared custom background layers".  
   But it might be better to think about it differently and use the preferences to store the actual GeoJSON or GPX file that the editors can then load right from the API. In this case, being able to store a GeoJSON of 1MB would be nice, and editors would only download it when the user requests it—same as any other external data store / URL.

Those use cases are only examples, and the linked drafts I wrote are just to get the discussion started.  
The potential and strength of this API is that it is shared between apps, and I am looking for ways to make using it easy and start the collaboration between editors to use it to share settings.

I talked with @k-yle about how we could create a wrapper for the current API to make it easier now for at least JavaScript-based apps to use the API with the current limits: https://github.com/osmlab/osm-api-js/issues/32#issuecomment-3660625648. It would be possible. However, before we go down this road, I wanted to first try this PR.

**What is the right limit?**

* I started this PR with 10k because that is what we use elsewhere (e.g. blog posts).
* I then updated it to 1,000,000 after talking to Kyle.
* Above are some other estimates, if we want to look into small numbers. The max is ~50k. So maybe setting it to 100k is a compromise?

The important part is: Even a bigger limit that still might be too small will require all apps to handle edge cases and communicate them to the user. So either we all need to implement an edge-case splitting approach, or communicate some hard-to-understand "you have too many presets, we cannot store this much, delete some…" message to users, which will be bad UX and a lot of work.

The idea of the 1,000,000 is not that editors will store this much data, but that we have a limit that we will never hit, so it does not get in the way … while still setting some ground rules.

I am very open to other numbers here!

@tyrasd mentioned that he might be able to get some data from his Overpass Turbo Short URL service, which would be a real-world use case that uses the workaround ATM.

**What apps could use this feature or use it already?**

The user/preferences API is used for different purposes:
* Store UI state (for a specific app)
* Store mapping-related preferences (but for now just for one specific app)
* Other

From @1ec5 in https://github.com/openstreetmap/openstreetmap-website/issues/6297#issuecomment-3223310244

> An incomplete list of current usage based on a cursory GitHub search:
>
> - MapComplete: favorites, themes, some other preferences?
> - Overpass Turbo: [saved queries](https://github.com/tyrasd/overpass-turbo/blob/3abf31b358eead177d3c130f85e696f5306c0569/js/sync-with-osm.ts).
> - Go Map!!: https://github.com/bryceco/GoMap/issues/599, replaced with iCloud in https://github.com/bryceco/GoMap/commit/7bfcb5f08e0b922038901bafe80efb050d4de778
> - Also, [Rapid](https://github.com/facebook/Rapid/blob/d4254643373b2ebd1076e10bcba11fed380fa3c4/modules/services/OsmService.js#L713) and [Vespucci](https://github.com/ltog/osmeditor4android/blob/92602d243a8debb703eb6ab8a0c96d3fe55a0d15/src/main/java/de/blau/android/tasks/MapRouletteApiKey.java#L57) use this API to communicate with the [MapRoulette backend](https://github.com/maproulette/maproulette-backend/blob/c888df5e045b83f3d8f4210b0e2ade2905b2c343/app/controllers/AuthController.scala#L250) via a MapRoulette API key.

Some of this data is not relevant for my idea of sharing data between editors. Settings like "is the Mapillary layer on" in iD are likely something that only iD will read and write (this feature is on the roadmap), given that I might not want to see this layer in Go Map all the time. But in the end, those decisions are up to the editors.

Classic preferences like this will be easy to store within the current key-value limits. However, apps like iD would have an easier life if iD could just write a full config object with all maybe 50 settings instead of handling 50 separate key-value tags.

**How do those apps handle the current limitation?**

AFAIK, MapComplete and Overpass Turbo are the only apps that write more than simple key-value pairs to the API.  
AFAIK, both use a custom way to split values into separate key-value pairs and re-join them later.  
(My idea in https://github.com/osmlab/osm-api-js/issues/32 was to extract this logic into a shared helper.)

**Why is this relevant now?**

Go Map recently added more features that I would love to be able to reuse in other editors. iD has features on the roadmap for all the described use cases, and it would be great to be able to store them using this API. StreetComplete and Every Door both use (part of) the iD Tagging Schema, and it would be great to be able to share custom presets with users of those apps. This will remove some pressure from the tagging-schema project to support presets that are relevant for only a few users or only relevant for a local community.

So IMO it is the right time now to make it easier for all those apps to collaborate.

mmd-osm:
Have you ever considered creating a proper dedicated API according to your schema definition, rather than trying to squeeze all the data into some generic key-value store that was never really designed for highly structured data?

tyrasd:
I was just about to post the potential use cases from an OSM editor application (custom background imagery URLs, custom presets data in JSON format, etc.), but it seems @tordans was faster. :sweat_smile: 

Not having looked at @tordans's [editor-settings-schema](https://github.com/tordans/editor-settings-schema) in detail yet, my naive initial idea was also to implement this using the _preferences_ API: e.g. by storing the number of to be saved custom background layer entries as the value of one preference, and then one additional entry per to be stored URL. For presets this would be similar, except that the payload would be a JSON representation of the preset instead.

Would introducing a new API for this really be a big benefit? I see the downside of having to implement and maintain of additional endpoints. And there we would again have the same question of how much data one would allow a user to store in that extra API, wouldn't we?

PS: here are some numbers for reference for some of the mentioned use cases:
* currently, an median length of a URL of an imagery entry in ELI is 170 characters, but 59 entries are longer than 255 characters
* the average file size of a preset in the id-tagging-schema is 433 bytes. that might be compressible a bit by stripping unnecessary whitespace, but it would still be often over the 255 bytes limit, I assume
* for how overpass-turbo currently stores queries (which admittedly is kind of questionable of whether its use of the _preferences_ API is intended or not), the average size of a stored query should be in the ballpark of around 550 characters (based on the average size of queries shared publicly through overpass-turbo.eu's short URL service)

simonpoole:
It is very unclear which "sharing" we are even discussing here:

- sharing of configuration between editing apps operated by the same contributor.
- sharing of configuration between contributors.
- or a mix of both.

In any case I would note that making it easier to discover data/imagery layer availability is an old idea and such information should as a rule not be hidden in private configuration and always be available for public inspection.

See https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2017/Project_Ideas#Third_Party_Source_Management_System_2

tyrasd:
I assumed we were talking about 

> sharing of configuration between editing apps operated by the same contributor.

But the same issues would arise even for the case of 

> saving of configuration of a single editing app operated by a single user across multiple devices / editing sessions

pablobm:
Is there potential for malicious use of this feature? Say... using these settings as storage backend for a website distributing pirated content.

simonpoole:
> Is there potential for malicious use of this feature? Say... using these settings as storage backend for a website distributing pirated content.

Well there currently isn't any limit on the _number_ of preferences, so, theoretically, there's misuse potential even now, though a block size of 255 bytes wouldn't be particularly convenient (yes I know it is actually more bytes ....).

PR #6400, opened 2025-09-19
Map: Add directional arrows to ways

tordans:
### Description

Whenever I review OSM data that has some kind of direction tag eg `cycleway:right=lane` I have to switch to iD in order to see the direction of the way. This DRAFT shows how we could add direction indicators on the map using the existing leaflet setup with two leaflet plugins.
* https://github.com/slutske22/leaflet-arrowheads
* which also requires https://github.com/makinacorpus/Leaflet.GeometryUtil

I am pretty sure this PR is not wat we want from a code point of view but rather have this function added to https://github.com/openstreetmap/leaflet-osm (right?)

So for now this is just a draft to get directions on if this is something that is worth continuing on and if it actually belongs in `leaflet-osm`.

### How has this been tested?

This is what it looks like with those changes:
- http://localhost:3001/way/29892612
   way
- http://localhost:3001/way/289093405
   area
- http://localhost:3001/node/1670728448
    point
- http://localhost:3001/relation/11612537
    relation

### DIFF
--- a/Vendorfile
+++ b/Vendorfile
@@ -22,6 +22,15 @@ folder 'vendor/assets' do
     from 'https://github.com/jieter/Leaflet.encoded.git', :tag => '0.0.9' do
       file 'leaflet.polyline.js', 'Polyline.encoded.js'
     end
+
+    from 'https://github.com/slutske22/leaflet-arrowheads.git' do
+      file 'leaflet.arrowheads.js', 'src/leaflet-arrowheads.js'
+    end
+
+    from 'https://github.com/makinacorpus/Leaflet.GeometryUtil.git' do
+      # Needed for leaflet.arrowheads.js
+      file 'leaflet.geometryutil.js', 'src/leaflet.geometryutil.js'
+    end
   end
 
   folder 'iD' do
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -8,6 +8,8 @@
 //= require leaflet.note
 //= require leaflet.share
 //= require leaflet.polyline
+//= require leaflet.geometryutil
+//= require leaflet.arrowheads
 //= require leaflet.query
 //= require leaflet.contextmenu
 //= require index/contextmenu
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -313,7 +313,17 @@ L.OSM.Map = L.Map.extend({
           map._objectLayer = new L.OSM.DataLayer(null, {
             styles: {
               node: objectStyle,
-              way: objectStyle,
+              way: {
+                ...objectStyle,
+                arrowheads: {
+                  size: '6px',
+                  frequency: 'allvertices',
+                  yawn: 60,
+                  fill: true,
+                  fillColor: objectStyle.color,
+                  color: objectStyle.color
+                }
+              },
               area: objectStyle,
               changeset: changesetStyle
             }
--- a/vendor/assets/leaflet/leaflet.arrowheads.js
+++ b/vendor/assets/leaflet/leaflet.arrowheads.js
@@ -1,0 +1,691 @@
+function modulus(i, n) {
+	return ((i % n) + n) % n;
+}
+
+function definedProps(obj) {
+	return Object.fromEntries(
+		Object.entries(obj).filter(([k, v]) => v !== undefined)
+	);
+}
+
+/**
+ * Whether or not a string is in the format '<number>m'
+ * @param {string} value
+ * @returns Boolean
+ */
+function isInMeters(value) {
+	return (
+		value
+			.toString()
+			.trim()
+			.slice(value.toString().length - 1, value.toString().length) === 'm'
+	);
+}
+
+/**
+ * Whether or not a string is in the format '<number>%'
+ * @param {string} value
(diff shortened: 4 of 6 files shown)


1ec5:
I’m surprised no one has requested this feature before (as far as I can tell). The arrows would definitely be more convenient than opening both endpoint nodes and comparing their locations manually. Some users might find the arrows annoying for the purpose of screenshotting something for a lay audience, but that isn’t really what the site is for anyways.

leaflet-osm might have [other clients](https://github.com/openstreetmap/leaflet-osm/network/dependents) besides osm-website that are less mapper-focused. If the arrows would be distracting for laypeople, maybe it could be an option that we enable here.

tomhughes (commented):
As `leaflet.arrowheads` is on NPM it would be better to add it to `package.json` rather than vendoring it - that will also pull in `leaflet.geometryutils` as NPM has a dependency there.

tomhughes:
> leaflet-osm might have [other clients](https://github.com/openstreetmap/leaflet-osm/network/dependents) besides osm-website that are less mapper-focused. If the arrows would be distracting for laypeople, maybe it could be an option that we enable here.

The other problem with putting it in `leaflet-osm` is that it would require adding a dependency on `leaflet-arrowheads` which really means publishing an update to NPM but I have no idea who has the power to do that... Then again fixing that so we could use the NPM version would be good.

pablobm:
One thing I don't understand: how is the direction encoded in the data? Is it just the "direction" in which the ways happen to be stored in DB (in other words: the order of the nodes)?

When it comes to showing the direction of a path, lane, etc, is that enough, or is additional tagging required?

tomhughes:
Yes the direction is just determined by the order of the nodes in a way.

pablobm:
OK, cool. Glad to know I'm not entirely losing it 😄

So if I understand correctly: rendering the arrows to show the direction of a cycle lane, a bus route, etc would only make sense if the way was entered in the right order.

In this general idea, I'm guessing:
- This could be helped by editor tools with the ability to flip the order of nodes in a way.
- To avoid showing irrelevant directions, we would need tags. Perhaps the opposite: a tag to mark that the direction of the way is relevant (since in most ways it won't be).

tomhughes:
So tagging can sort of override the natural direction - so `oneway=-1` is understood to mean it's one way in the opposite direction to the order of the nodes...

I think what is being suggested here is just to show the "natural" order though.

tordans:
> I think what is being suggested here is just to show the "natural" order though.

Correct. @pablobm Showing the direction of the line is was iD and other editors do in order to provide feedback to mappers how to interpret directional tagging like `*:forward`, `*:backward` and `*:left`, `*:right` ([Wiki](https://wiki.openstreetmap.org/wiki/Forward_%26_backward,_left_%26_right)). A prominent example would be [`cycleway:right`](https://taginfo.openstreetmap.org/keys/cycleway%3Aright).

---

Given the feedback at https://github.com/openstreetmap/openstreetmap-website/pull/6400#issuecomment-3314985776 I consider this PR abandoned and will look into adding the feature to https://github.com/openstreetmap/leaflet-osm

pnorman:
@jfirebaugh has NPM access. Do we have an account to transfer it to?

pablobm:
Thank you all for clarifying. One more thing: who are the public for this feature and what do they want to be shown?

1. For editors, seeing the direction of ways is useful, but that's already shown on iD which is the place for editing on the website (I don't know other editors).
2. For readers, the direction of physical things in the real world (bus routes, cycle paths, etc) is the useful thing to see. For this, the arrows would need to render based on both the storage order and the value of the relevant tags (`*:forward`, `*:backward`, etc).

@tordans does say "Whenever I review OSM data", so I'm guessing there's a workflow here that I'm not familiar with. Which of the above is more useful for this workflow? The assumption here appears to be option 1, but perhaps option 2 should be considered as that could help reviewing both aspects (way structure and tags) together.

tordans:
@pablobm my understanding of the goal of the OSM website is to support mappers with their work on OSM. I understand "readers" in your comment more like end users that are looking for an alternative to commercial map and routing services. Those are _not_ the audience we are building for. This is mentioned from time to time in comments when issues are closed and also conference talks (as far as I remember) but I don't remember if this was ever written down in some more formal "target audience" / "persona" (or similar) kind of way.

pablobm:
@tordans - You understand me correctly, thank you 🙂 

I do think that the audience is not well communicated. I admit that I only understood it indirectly after some time. I'm ok with that, but in terms of comms it's something to work on.

Having said that: is the general public ever considered as a secondary audience? Definitely build for the mappers, but at times there will be cases when some allowances can be done for the wider public. The website is not intended as an alternative to commercial maps, but there are opportunities for it to be built as an attractive way to interest new people, who then can become mappers and spread the word.

Sorry, I realise that I'm making this quite long with all these questions 😓 I'm trying to get a better understanding of all the thinking, and I'll be using these opportunities if y'all will have the patience, for which I very much thank you! 🙏

tordans:
@pablobm very happy to voice/video chat about this more to share my opinion. There is a lot of history… and then some opportunities for the future on this topic. If you want, send me an email (see https://github.com/tordans) so we can talk more. — Unfortunately I don't have the time to write it all out.

PR #6404, opened 2025-09-20
Move markers out of the map div for the embed view

### LINKED ISSUE #6045: Shared osm link doesn't show a marker
### URL

https://www.openstreetmap.org/note/4678312

For example, the shared osm link within this [note](https://www.openstreetmap.org/note/4678312) doesn't show a blue marker on the map. The [same link](https://www.openstreetmap.org/?mlat=46.767712&mlon=15.871360#map=17/46.767716/15.871360&layers=N) outside the osm website works as expected.

_No response_

tomhughes:
Having the markers inside the map div means they get destroyed when the map is created.

Fixes #6045.

### DIFF
--- a/app/views/export/embed.html.erb
+++ b/app/views/export/embed.html.erb
@@ -7,10 +7,10 @@
     <%= javascript_include_tag "embed" %>
   </head>
   <body>
+    <% if params[:marker].present? %>
+      <%= render :partial => "layouts/markers", :locals => { :types => %w[dot] } %>
+    <% end %>
     <div id="map">
-      <% if params[:marker].present? %>
-        <%= render :partial => "layouts/markers", :locals => { :types => %w[dot] } %>
-      <% end %>
     </div>
   </body>
 </html>


gravitystorm:
I've spent a while reviewing this, but I can't recreate the original problem (so I can't be sure that the changes fix it). For me, if I make a simple page with the current live embed widget (copy+pasted from https://github.com/openstreetmap/openstreetmap-website/issues/6045#issuecomment-3277248599 ), the marker shows up fine.

Screenshot from 2025-09-24 15-35-11

Is this a browser-specific problem, or is there something else going on that I'm not realising? Do I need to change something about my testcase to trigger the problem?

tomhughes:
It was reproducible for me in firefox with:

```html
<html>
  <head>
    <title>Map</title>
  </head>
  <body>
    <iframe width="425" height="350" src="https://www.openstreetmap.org/export/embed.html?bbox=15.857627391815187%2C46.76306775754699%2C15.882411003112795%2C46.772356640531214&amp;layer=mapnik&amp;marker=46.76771239926923%2C15.87001919746399" style="border: 1px solid black"></iframe><br/><small><a href="https://www.openstreetmap.org/?mlat=46.767712&amp;mlon=15.870019#map=17/46.767712/15.870019&amp;layers=N">View Larger Map</a></small>
  </body>
</html>
```

The marker does get created but as an SVG that refers to symbols that have been deleted from the DOM.

gravitystorm:
> It was reproducible for me in firefox with:

This is so strange, with your html I still can't reproduce it - the marker shows for me (Firefox 142.0.1). I was this close to just merging this PR anyway, but then I tested changing back and forward between the branch and master, and noticed that on the branch the map div is moved down slightly for me.

Screenshot from 2025-09-24 16-43-41
Screenshot from 2025-09-24 16-43-15

So I don't feel confident about merging this, and I also don't know what to do next. Hmm.

tomhughes:
I can't reproduce it now either, but I think I know what's really going on.

That view is cached in production, but it not does different things depending on whether `params[:marker]` is set when the request is made, so for each of the three servers the symbols will be there if the first request asked for a marker and not if it didn't.

I did a deployment last night which will have flushed the caches and caused the page to regenerate hence causing a change in behaviour.

tomhughes:
I've updated the fix now to move the symbols back into the div but make them unconditional so they're always present in case the javascript needs them for a given load.

gravitystorm:
Great, thanks for figuring that out!

MLNW:
Thanks, that fixed the issue!

PR #6405, opened 2025-09-21
Add support for OpenTelemetry reporting

tomhughes:
This will be necessary for https://github.com/openstreetmap/operations/issues/1284 to allow OpenTelemetry data collection.

Note that it does nothing unless an exporter is configured in the environment.


Gemfile
```diff
@@ -115,6 +115,11 @@ gem "dalli"
 # Load canonical-rails to generate canonical URLs
 gem "canonical-rails"
 
+# Use to generate telemetry
+gem "opentelemetry-exporter-otlp"
+gem "opentelemetry-instrumentation-all"
+gem "opentelemetry-sdk"
```
pablobm: These are `require`d by default. If we are going to `require` them only when `OTEL_TRACES_EXPORTER` is defined, then we should add `require: false` to each:

```suggestion
gem "opentelemetry-exporter-otlp", require: false
gem "opentelemetry-instrumentation-all", require: false
gem "opentelemetry-sdk", require: false
```

You can verify this by running: `bin/rails runner OpenTelemetry::SDK`. With the code in this PR, it runs without a complaint. If you add the `require: false` you'll see it end with `uninitialized constant OpenTelemetry`.

tomhughes: Thanks. I've made those changes now.

pablobm (approved):

gravitystorm:
* Should we add the env key to settings.yml ? I think (not sure though) that all our other env vars are documented there.
* Are we happy with the `opentelemetry-instrumentation-all` pulling in 36 dependencies? Is it easier to stick with that for now, or should be more selective (e.g. `-rails`, `-pg` and any others we need)? I don't know how often these dependencies churn, or increase application size, etc. Maybe it's not important, it just looks like a lot of stuff in the diff.

tomhughes:
So I wanted to add it to the settings but the opentelemetry gems really, really don't help with that and want you to use the environment :cry: 

I did eventually manage to make it work after digging through the source to find out what the gems did with the environment variables but it wasn't pretty - there's no API call that is a direct equivalent so you wind up having to create several objects yourself and wire them together.

Much the same goes for the `-all` thing - it is possible, and it's not quite as bad and is at least documented but it's quite a bit more verbose.

In particular there's no way to enable `opentelemetry-instrumentation-rails` and have it do everything so you have to do all the subcomponents of that separately and I wound up with something like this:

```ruby
require "opentelemetry/sdk"
require "opentelemetry/instrumentation/action_mailer"
require "opentelemetry/instrumentation/action_pack"
require "opentelemetry/instrumentation/action_view"
require "opentelemetry/instrumentation/active_job"
require "opentelemetry/instrumentation/active_record"
require "opentelemetry/instrumentation/active_storage"
require "opentelemetry/instrumentation/active_support"
require "opentelemetry/instrumentation/dalli"
require "opentelemetry/instrumentation/delayed_job"
require "opentelemetry/instrumentation/faraday"
require "opentelemetry/instrumentation/pg"
require "opentelemetry/instrumentation/rack"
require "opentelemetry/instrumentation/rails"

OpenTelemetry::SDK.configure do |c|
  c.use "OpenTelemetry::Instrumentation::ActionMailer"
  c.use "OpenTelemetry::Instrumentation::ActionPack"
  c.use "OpenTelemetry::Instrumentation::ActionView"
  c.use "OpenTelemetry::Instrumentation::ActiveJob"
  c.use "OpenTelemetry::Instrumentation::ActiveRecord"
  c.use "OpenTelemetry::Instrumentation::ActiveStorage"
  c.use "OpenTelemetry::Instrumentation::ActiveSupport"
  c.use "OpenTelemetry::Instrumentation::Dalli"
  c.use "OpenTelemetry::Instrumentation::DelayedJob"
  c.use "OpenTelemetry::Instrumentation::Faraday"
  c.use "OpenTelemetry::Instrumentation::PG"
  c.use "OpenTelemetry::Instrumentation::Rack"
  c.use "OpenTelemetry::Instrumentation::Rails"
end
```

gravitystorm:
Alright, let's go with what we have. Thanks for investigating!

Issue #6287, opened 2025-08-04
Errors on Docker + Apple Silicon


pablobm:
At the moment, the test suite fails when running in a Docker container on Apple Silicon.

I cloned the repo for the first time today and setup it up using the Docker instructions. I am using a MacBook Pro on Apple silicon. In my environment, the test suite is showing a few of these:
```
Error:
TraceTest#test_import_handles_gz:
NoMethodError: undefined method `gdImageCreate' for GD2::GD2FFI:Module
    lib/gpx.rb:81:in `block in picture'
    lib/gpx.rb:80:in `picture'
    app/models/trace.rb:258:in `block in import'
    app/models/trace.rb:205:in `import'
    test/models/trace_test.rb:255:in `test_import_handles_gz'
```

After looking a bit into it, it appears that the `gd2-ffij` gem doesn't handle the possibility of `libgd` living under `/usr/lib/aarch64-linux-gnu`. This is the offending code: [/lib/gd2-ffij.rb](https://github.com/dark-panda/gd2-ffij/blob/039480b72246463d9ad8acc6aed0536a1a31ec50/lib/gd2-ffij.rb#L35)

I'm not sure there is a non-horrible way to solve this. There's a recent PR in the `gd2-ffij` repo that would help (https://github.com/dark-panda/gd2-ffij/pull/28) but the repo hasn't had new commits since 2021, so I'm not hopeful that it will be merged any time soon.

For the moment, anyone with a setup similar to mine may have to resort to hacks. For example, I can get things going by adding the following environment variable in `docker-compose.yml`. However this not portable at all so I can't just create a PR:
```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index e98824e50..d8293a0f9 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
       context: .
     environment:
       PIDFILE: /tmp/pids/server.pid
+      GD2_LIBRARY_PATH: /usr/lib/aarch64-linux-gnu
     volumes:
       - .:/app
       # Prevent these directories from mounting so they're not shared between host OS and Docker
```
Thoughts? At least I guess this issue may help as a guide to others who find the same issue.

By the way, the "horrible ideas" that cross my mind are:
1. Fork `gd2-ffij`.
2. Monkey-patch `GD2::GD2FFI.gd_library_name`.

kcne:
[...]

PR #6408, opened 2025-09-22
Fix tests for Docker on Apple silicon

pablobm:
Fixes https://github.com/openstreetmap/openstreetmap-website/issues/6287
### Context
Currently the test suite fails when run with Docker on Apple post-Intel machines (aka Apple silicon, AArch64, ARM64). The issue is with the gem `gd2-ffij`, which hard-codes possible locations for its underlying C library. Unfortunately it doesn't allow for the path that this library uses in an AArch64 machine, and doesn't offer a configuration option that would serve our use case.
### Description
This PR implements a proposal by @firefishy (https://github.com/openstreetmap/openstreetmap-website/issues/6287#issuecomment-3156583006): it changes the gem dependency to use a fork that implements the required change. Since it's bound to the specific SHA of the commit, it should be safe to use.

This fix is based on an outstanding PR for the `gd2-ffij` repository, awaiting since last June. Given the low level of activity in [the `gd2-ffij` repo](https://github.com/dark-panda/gd2-ffij) (last commit September 2021), it's unlikely that it will be merged.
### How has this been tested?
I have tested this on my MacBook M1 Pro machine.
### Caveat
If the forked repo or the relevant commit are removed from GitHub, that will break our builds. A fix for this would be for OSM to host this fork instead.

 Thoughts?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -118,7 +118,8 @@ gem "canonical-rails"
 # Used to generate images for traces
 gem "bzip2-ffi"
 gem "ffi-libarchive"
-gem "gd2-ffij", ">= 0.4.0"
+# Use https://github.com/dark-panda/gd2-ffij/pull/28 for Docker/macOS compatibility
+gem "gd2-ffij", :github => "rkoeze/gd2-ffij", :ref => "a203a8d5ef004a4198950e86329228fe3f331d06"
 gem "marcel"
 
 # Used for S3 object storage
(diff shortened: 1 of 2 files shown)


tomhughes:
I don't really like doing this sort of thing, especially where there seems little chance of it being temporary, but I guess we don't have any better options.

We probably need to be considering if there's some way we can replace this gem if it's not being maintained...

pablobm:
It's indeed a smell. I have opened an issue to look into whether we could remove `gd2-ffij` altogether: https://github.com/openstreetmap/openstreetmap-website/issues/6415

PR #6410, opened 2025-09-22
Bootstrap contextmenu

### LINKED ISSUE #6383: Right click menu is unnecessarily small
Note the size of the Chrome three-dot menu items:

Image

Now compare this with the size of each item in OSM.ORG's right click menu:

Image

No wonder many times when we intend to click "Show address" we end up clicking "Add a note here".

mmd-osm:
This PR replaces the current Leaflet.contextmenu based context menu by a Bootstrap based version. Adds a new dependency on [Bootstrap icons](https://icons.getbootstrap.com/)

image

Fixes #6383

### How has this been tested?
* Local tests on desktop / mobile browsers, checking if all 6 actions are still working
* Test all four corners of the screen, dropdown is supposed to stay within map boundaries
* LTR and RTL mode
* Normal/Dark mode, including currently highlighted menu item
* Depending on zoom level, "Query features" and "Add a note here" are disabled
* Dropdown is hidden when: (1) clicking on an action, (2) clicking anywhere in the window, (3) panning the map.


config/locales/en.yml
```diff
@@ -591,6 +591,14 @@ en:
     no_such_entry:
       heading: "No entry with the id: %{id}"
       body: "Sorry, there is no changeset with the id %{id}. Please check your spelling, or maybe the link you clicked is wrong."
+  context:
```
tomhughes: If you moved this to `layouts.contextmenu` instead of here then you could just use `.directions_from` etc in the view to reference them.

app/assets/stylesheets/common.scss
```diff
@@ -4,6 +4,9 @@
 @import "rails_bootstrap_forms";
 @import "colors";
 
+$bootstrap-icons-font-dir: "bootstrap-icons/font/fonts";
+@import "bootstrap-icons/font/bootstrap-icons";
```
tomhughes, line 8: Am I right in thinking that we've used icons from bootstrap icons before, but by manually importing them to our assets instead of like this?

tomhughes (commented):
I think the structure of the commits here is wrong - rather than removing the old assets and commenting out the code in the first commit and then rewriting things and re-enabling the code in the second I would reverse things and have the first commit rework the code and the the second one clean up the now unused assets, style rules etc.

On a broader level I'm not sure what I think about this... Getting rid of some non-npm vendored js is nice but I on the flip side we're moving from having something do the work for us to seemingly doing everything from scratch ourselves.

I think I thought bootstrap was going to be doing more for us then it actually is - it's just providing styling and not actually any sort of popup menu support which is more what I had assumed we were talking about.

config/locales/en.yml
mmd-osm: Good point! Fixed.

app/assets/stylesheets/common.scss
mmd-osm: We seem to have a few icons by the same name in different locations. The ones I quickly checked turned out to have a different file contents. I haven't found any larger scale copying of bootstrap icons. 

I was following the docs on https://icons.getbootstrap.com/ here, which has some details on how to use bootstrap icons along with Sass.

```
./app/assets/images/social_link_icons/tiktok.svg
./node_modules/osm-community-index/dist/img/tiktok.svg
./node_modules/bootstrap-icons/icons/tiktok.svg
```

tomhughes: I thought a lot of the icons in `app/assets/images/icons` came from the bootstrap set but I might be wrong.

mmd-osm: A few of them, like calendar.svg, chat.svg, chat-fill.svg are indeed almost identical to their bootstrap icons counterpart, except that the class attribute has been removed from the svg element.

mmd-osm:
> we're moving from having something do the work for us to seemingly doing everything from scratch ourselves

Bootstrap offers some more options which I haven't fully explored yet. Popper.js which is bundled with Bootstrap has some way to automatically flip contextmenus. I think this way, we might get rid of the two utility functions. I suppose the other bits in this PR are ok, and within the expected scope of how Bootstrap works.

mmd-osm:
There's still quite a bit of boilerplate code in app/views/layouts/_contextmenu.html.erb, which could probably be generated based on the action definition. I will take a look at what is feasible here.

config/locales/en.yml
```diff
@@ -1878,6 +1878,14 @@ en:
...
       expected_restore_html: "Services are expected to be restored in %{time}."
       announcement: "You can read the announcement here."
+    contextmenu:
+      directions_from: Directions from here
+      directions_to: Directions to here
+      add_note: Add a note here
+      show_address: Show address
+      query_features: Query features
+      centre_map: Centre map here
+      scroll_to_changeset: Scroll to changeset
```
danieldegroot2: 'Scroll to changeset' does not seem to have an icon yet, though it could be left empty.

S: There is '[eye](https://icons.getbootstrap.com/icons/eye/)' icon. ('View')
(of course user is viewing only its listing, not any kind of details)

Alternatively, '[search](https://icons.getbootstrap.com/icons/search/)' icon could be (re)used (but might be confused with query feature)

Otherwise, there is '[align middle](https://icons.getbootstrap.com/icons/align-middle/)' or '[align top](https://icons.getbootstrap.com/icons/align-top/)' depending on interpretation.
(align element 'box' to 'bar' in vertical direction shown by orientation of element 'box').
Maybe not comprehensive for the majority, as it does not appear often online.

(sidenote: it technically aligns the changeset to the of the sidebar, but it is not the top of the changeset list until going to the changeset and back which locks the changeset list view. even then it could be confusing.)

pablobm: At the risk of overdoing this, here are some more that could work:

- Clipboard: https://icons.getbootstrap.com/icons/clipboard2/
- File break: https://icons.getbootstrap.com/icons/file-break/
- Journal bookmark: https://icons.getbootstrap.com/icons/journal-bookmark-fill/

Also, unless I'm missing something, "Scroll to changeset" has not been added yet at all.

mmd-osm: >  "Scroll to changeset" has not been added yet at all.

Right, this one isn't covered by system tests at the moment, and I seem to have missed it.

app/assets/javascripts/index/history-changesets-layer.js -> addChangesetLayer

pablobm: Am I right in thinking that only "Show address" is covered by system tests currently? (at [test/system/search_test.rb](https://github.com/openstreetmap/openstreetmap-website/blob/d09d922a9107a854d5778e5eccc2b3c75942e9e7/test/system/search_test.rb#L50)). This is something I could work on, if desirable.

mmd-osm: I haven't checked all the other actions. My comment was more related to the observation that the system tests didn't break even though "Scroll to changeset" didn't work anymore. For sure I won't hurt to add some more tests in another PR to have better coverage of the status quo.

app/assets/javascripts/index/contextmenu.js
```diff
@@ -1,80 +1,200 @@
+/* global Popper */
```
tomhughes: Other global declarations are in `config/eslint.config.mjs` rather than being done on a per-file basis like this?

app/assets/javascripts/index/contextmenu.js
```diff
@@ -1,80 +1,200 @@
...
+  ];
 
-  const updateMenu = function updateMenu() {
-    map.contextmenu.setDisabled(2, map.getZoom() < 12);
-    map.contextmenu.setDisabled(4, map.getZoom() < 14);
-  };
+  OSM.renderContextMenu($contextMenu, contextmenuItems);
+};
+
+OSM.createSeparator = function () {
```
tomhughes: This (and `createMenuItem` and `renderContextMenu`) are only used from `initializeContextMenu` at the moment so they could just be declared in there rather than being exported in the `OSM` namespace unless you're expecting to be using them somewhere else?

Though I guess they may go away if you switch to rendering the menu on the server.

mmd-osm: They're meant to be used in app/assets/javascripts/index/history-changesets-layer.js -> addChangesetLayer later on, which needs some dynamic changes to the contextmenu.

app/assets/javascripts/index/contextmenu.js
mmd-osm: done

config/locales/en.yml
mmd-osm: I've added a bootstrap popup for the changeset bbox now. At the moment, I'm only adding the "Scroll to Changeset" action, like shown here:

image

Previously, this "scroll to changeset" action and a separator were merged with the existing context menu. I have some doubts if all these default actions really make sense in this scenario. If you happen to have an opinion on the topic, please let me know.

image

Code is a bit of a mess right now, and will be cleaned up later. I'm trying to get the functionality right first.

pablobm: You make a good point questioning what should appear on this menu. Given that a changeset can cover a wide area, I think that directions, note, and address don't make a lot of sense here. Perhaps "Query features" and "centre map here" do make sense?

tomhughes: Centre map maybe but if we now view that as a menu related to the changeset rather than a general menu then I don't see query features as relevant.

pablobm: Re: the choice of icon - It's also legitimate to decide that there's no icon suitable for a specific action, and leave that action icon-less.

gravitystorm:
This looks great, and I'm happy to merge it as-is.

I'm happy to see the bootstrap-icons dependency, and I hope this makes it easier for other developers to add icons elsewhere in the site, where they are useful.

I'm going to create a follow-up issue about making our icons more consistent, since several items on the menu have icons which don't match the icons elsewhere in the UI. However, I didn't feel like these inconsistencies should block the merging of this PR, which was substantially concerned with changing the implementation. We can discuss and iterate the icons separately.

I also agree with Tom that it would be good to make the html generated in a view, where it's easier for most developers to find and adjust, but that can also be tackled separately.

Thanks @mmd-osm for your work on this!

PR #6411, opened 2025-09-22
Confirmation popup for Overpass API export

mmd-osm:
### Description

The list below is a bit confusing from a UX point of view. While 3 out of 4 links would take a user to another page with more details about download options, the Overpass API link immediately starts a potentially huge download. The link is followed by a harmless-looking explanatory text that most users will probably overlook.

image

To make things a bit clearer, I'm adding a confirmation pop-up for now.

image

Some UX redesign might be the better approach in the long run.

### How has this been tested?
Local test

### DIFF
--- a/app/assets/javascripts/index/export.js
+++ b/app/assets/javascripts/index/export.js
@@ -37,6 +37,28 @@ OSM.Export = function (map) {
     validateControls();
   }
 
+  async function showConfirmationModal() {
+    const $modal = $("#export_confirmation");
+    const $downloadButton = $modal.find("[data-action=\"download\"]");
+    $modal.appendTo("body").modal("show");
+
+    return new Promise(resolve => {
+      const onOkClick = () => {
+        resolve(true);
+        $modal.modal("hide");
+      };
+
+      const onModalHidden = () => {
+        $downloadButton.off("click", onOkClick);
+        $modal.off("hidden.bs.modal", onModalHidden);
+        resolve(false);
+      };
+
+      $downloadButton.on("click", onOkClick);
+      $modal.on("hidden.bs.modal", onModalHidden);
+    });
+  }
+
   function setBounds(bounds) {
     const truncated = [bounds.getSouthWest(), bounds.getNorthEast()]
       .map(c => OSM.cropLocation(c, map.getZoom()));
@@ -82,6 +104,15 @@ OSM.Export = function (map) {
       .addEventListener("turbo:before-fetch-request", function (event) {
         event.detail.fetchOptions.headers.Accept = "application/xml";
       });
+
+    $("#export_overpass").on("click", async function (event) {
+      event.preventDefault();
+      const downloadUrl = $(this).attr("href");
+      const confirmed = await showConfirmationModal();
+      if (confirmed) {
+        window.location.href = downloadUrl;
+      }
+    });
 
     update();
     return map.getState();
--- a/app/views/site/export.html.erb
+++ b/app/views/site/export.html.erb
@@ -48,3 +48,25 @@
     <dd><%= t ".too_large.other.description" %></dd>
   </dl>
 <% end %>
+
+<div class="modal fade" tabindex="-1" id="export_confirmation" aria-hidden="true">
+  <div class="modal-dialog modal-dialog-centered">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title"><%= t ".confirmation.header" %></h5>
+        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+      </div>
+      <div class="modal-body">
+        <p><%= t ".confirmation.body" %></p>
+      </div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">
+          <%= t ".confirmation.cancel" %>
+        </button>
+        <button type="button" class="btn btn-primary" data-action="download">
+          <%= t ".confirmation.download" %>
+        </button>
+      </div>
+    </div>
+  </div>
+</div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2623,6 +2623,11 @@ en:
           description: "Additional sources listed on the OpenStreetMap Wiki"
       export_button: "Export"
       export_download: "Downloading..."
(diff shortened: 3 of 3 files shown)


app/assets/javascripts/index/export.js
```diff
@@ -37,6 +37,55 @@ OSM.Export = function (map) {
     validateControls();
   }
 
+  async function showConfirmationModal(message) {
+    return new Promise((resolve) => {
+      const modalHtml = `
```
gravitystorm: This block of html should be in a erb view template. It might make sense to put it in the same view template as the overpass api link, since they are closely related.

config/locales/en.yml
```diff
@@ -3531,6 +3531,11 @@ en:
       filename: "map"
     embed:
       report_problem: "Report a problem"
+    export:
+      header: Confirm
```
gravitystorm: This should describe the subject of the body, for example "Download XML Data" or something similar

config/locales/en.yml
```diff
@@ -3531,6 +3531,11 @@ en:
...
+      body: Do you want to download map data for this area in XML format? Note that large areas may take some time to download.
+      cancel: Cancel
+      ok: OK
```
gravitystorm: Please avoid using "OK" as button text. Instead we should use a verb for the action being taken, e.g. "Download" - see e.g. https://ux.stackexchange.com/a/9960

gravitystorm (changes requested):
I agree it's worth having a warning for this link. It would be even better if the link was to the third-party website, for example pre-populating a form or something similar. That website could then take care of things like formats, warnings, etc.

app/assets/javascripts/index/export.js
pablobm: Would it make sense to make this into a Turbo frame?

gravitystorm: I'm not sure the complexity is worth it, it's only a small amount of html so loading it inline with the rest of the export panel seems reasonable to me.

mmd-osm: > It might make sense to put it in the same view template as the overpass api link, since they are closely related.

Lots of fun, the backdrop is hiding the modal dialog underneath, making it unusable. Moving it outside the "sidebar" div immediately fixes it.

image

config/locales/en.yml
mmd-osm: done

config/locales/en.yml
mmd-osm: fixed

app/assets/javascripts/index/export.js
mmd-osm: This seems to be a fairly well known issue: https://stackoverflow.com/questions/10636667/bootstrap-modal-appearing-under-background has tons of upvotes. I'm using this trick now: https://stackoverflow.com/a/15780841/3501889

app/views/site/export.html.erb
```diff
@@ -48,3 +48,25 @@
...
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title"><%= t ".confirmation.header" %></h5>
+          <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+        </div>
+        <div class="modal-body">
+          <p><%= t ".confirmation.body" %></p>
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">
```
hlfan: How about rejecting the Promise on dismissal?

config/locales/en.yml
```diff
@@ -2623,6 +2620,11 @@ en:
           description: "Additional sources listed on the OpenStreetMap Wiki"
       export_button: "Export"
       export_download: "Downloading..."
+      confirmation:
+        header: Download XML Data
+        body: Do you want to download map data for this area in XML format? Note that large areas may take some time to download.
+        cancel: Cancel
+        ok: Download
```
pablobm: Should the key also be `download`, to better reflect the intent? And also the id, variable, function, etc. Nitpicky, but I think it's worth having the code describe the action, adding clarity.

app/assets/javascripts/index/export.js
```diff
@@ -37,6 +37,28 @@ OSM.Export = function (map) {
     validateControls();
   }
 
+  async function showConfirmationModal() {
+    const $modal = $("#export_confirmation");
+    const $okButton = $("#ok_button");
```
pablobm: I'm thinking that this is a bit generic to get an id on its own. A different modal could be added in the future that also has `id="ok_button"` (or whatever name it receives, as per my other comment). Perhaps this should be handled with a class? Like this:


```suggestion
    const $okButton = $modal.find(".ok_button");
```

app/views/site/export.html.erb
mmd-osm: I've used resolve(false) instead now.

config/locales/en.yml
mmd-osm: done

app/assets/javascripts/index/export.js
mmd-osm: `.find` checks the DOM subtree of $modal here, so I don't think we need a class here.

pablobm: What I mean is that, technically, an id should be unique to the whole page. So if there's another `id="download_button"` in the future, that will be against the HTML standard and perhaps malfunction with some DOM implementation. With a class, we can have as many `.download_button` as we wish in the same page.

mmd-osm: While I would argue that future users should really take care of it, we can add something similar to a class now: `data-action="download"`

pablobm: Attribute works for me :+1:

Future users are any contributors to this project. One day someone may add a duplicate id and not notice. Perhaps something stops working in some weird way. It might not be immediately obvious what it is.

mmd-osm: I think we're already past this point:

```
There are 7 duplicate ids in DOM

query
modes_car
modes_bicycle
modes_foot
routing_engines
route_from
route_to
```

pablobm: :scream:

gravitystorm:
Looks great! Thanks for the refactoring.

PR #6412, opened 2025-09-22
Language selection enhancements

mmd-osm:
### Description

This PR includes the following two changes:

- Dynamic loading using Turbo: instead of rendering the language selection for every page, this is now done on demand only, when clicking on the language selection button.
   + Some sidebar navigation steps don't automatically reload the header section, which is why I have to evaluate and possibly update the turbo frame "src" link in Javascript. This happens e.g. when navigating from a changeset to an element.
   + Rails controller name and path names may be confusing. I'm open for better ideas.

- Search field: since the list of languages is a bit long, I have added a type ahead search at the top of popup window. It's doing some case-insensitive sub-string matching, and filters out any non-matching entries.

Here's how the overall result looks like:

image

image

or as screen recording (with the example used in system tests):

https://github.com/user-attachments/assets/4deaf116-a7c0-4454-962f-0a17a10d5e7d

### How has this been tested?
Local test


app/controllers/languages_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class LanguagesController < ApplicationController
+  before_action :authorize_web
+  before_action :set_locale
+  authorize_resource :class => false
+
+  def show
+    @source_page = params[:source]
+    render :partial => "layouts/select_language_list_frame"
```
tomhughes: Why not use the standard rails view name of `languages/show` for this, then you would just need `render :layout => false` like the other panes?

config/routes.rb
```diff
@@ -208,6 +208,7 @@
   post "/login" => "sessions#create"
   match "/logout" => "sessions#destroy", :via => [:get, :post]
   get "/offline" => "site#offline"
+  resource :language, :path => "/select_language_list", :only => :show
```
tomhughes: Maybe we should use the same naming as the other panes?
```suggestion
  resource :languages_pane, :path => "/panes/languages", :only => :show
```

app/views/layouts/_header.html.erb
```diff
@@ -116,15 +116,21 @@
...
-        <% if current_user&.id %>
-          <%= form_tag basic_preferences_path, :method => "PUT" do %>
-            <%= hidden_field_tag "referer", request.fullpath %>
-            <%= hidden_field_tag "language", I18n.locale %>
-            <%= render "layouts/select_language_list" %>
-          <% end %>
-        <% else %>
-          <%= render "layouts/select_language_list" %>
-        <% end %>
+        <div class="mb-3 px-2">
```
tomhughes: What's the logic behind rendering the search form in the main page instead of as part of the turbo frame?

mmd-osm: That's exactly how I started out, then I thought if the /panes/languages list is supposed to return a clean list of language links, why should the search field belong in there. I'm really undecided here which option makes more sense.

config/routes.rb
mmd-osm: This is fixed now.

app/controllers/languages_controller.rb
mmd-osm: Good idea, I'm following the layers_panes pattern now.

app/views/layouts/_header.html.erb
tomhughes: I see it as returning the language selector dialog I guess..

app/controllers/languages_controller.rb
tomhughes: The `select_language_list` partial should move as well as `languages/show` is the only thing that uses it.

mmd-osm: Done... I had to move some of the translations as well.

app/assets/javascripts/application.js
```diff
@@ -201,4 +201,34 @@ $(function () {
...
+            const text = $(this).text().toLowerCase();
+            $(this).toggle(text.indexOf(query) > -1);
+          });
+        });
+      }
+    }
+  });
+
+  $("#select_language_dialog").on("shown.bs.modal", function () {
+    $("#language_search").val("").trigger("input");
```
pablobm: Would it make sense to grab the focus so that the user can start typing straightaway?

```suggestion
    $("#language_search")
      .focus()
      .val("").trigger("input");
```

(I'm not an accessibility expert, so I might be proposing an a11y sin here).

app/views/layouts/_header.html.erb
```diff
@@ -112,19 +112,25 @@
...
+                             params[:language_search],
+                             :placeholder => t("languages_panes.select_language_list.search_language", :default => "Search..."),
+                             :autocomplete => "off",
+                             :dir => "auto",
+                             :class => "form-control py-2 px-3" %>
+        </div>
+        <turbo-frame id="select_language_list" src="<%= languages_pane_path %>?source=<%= ERB::Util.url_encode(request.fullpath) %>" loading="lazy">
+          <div class="text-center py-3">
+            <div class="spinner-border text-secondary" role="status">
+              <span class="visually-hidden">Loading...</span>
```
tomhughes: Presumably `Loading...` here should be translatable...

app/views/layouts/_header.html.erb
```diff
@@ -112,19 +112,25 @@
   <div class="modal-dialog modal-lg modal-dialog-scrollable">
     <div class="modal-content">
       <div class="modal-header">
-        <h1 class="modal-title fs-5" id="select_language_dialog_label"><%= t ".select_language" %></h1>
+        <h1 class="modal-title fs-5" id="select_language_dialog_label"><%= t "languages_panes.select_language_list.header.select_language" %></h1>
```
tomhughes: What was the logic of moving this string so that it then needed a fully qualified name? I don't it's used in the partial whose scope you've now put it is it?

app/views/layouts/_header.html.erb
```diff
@@ -112,19 +112,25 @@
...
-            <%= hidden_field_tag "language", I18n.locale %>
-            <%= render "layouts/select_language_list" %>
-          <% end %>
-        <% else %>
-          <%= render "layouts/select_language_list" %>
-        <% end %>
+        <div class="mb-3 px-2">
+          <%= text_field_tag :language_search,
+                             params[:language_search],
+                             :placeholder => t("languages_panes.select_language_list.search_language", :default => "Search..."),
```
tomhughes: Same question here about the scope of the resource string, though obviously in this case it depends on if the form stays here or moves but in general we put translations in the scope where they're used so we can refer to them with a relative name.

app/views/layouts/_header.html.erb
mmd-osm: We have something similar in app/views/browse/_relation.html.erb. I think the text would be relevant for screen readers only, and it not visible otherwise.

Image


Image

This pattern is following the example here: https://getbootstrap.com/docs/5.0/components/spinners/#border-spinner

app/views/layouts/_header.html.erb
mmd-osm: >  it depends on if the form stays here

Right, I wanted to keep all translations in the same place, also anticipating that we will move the form again.

app/views/layouts/_header.html.erb
1ec5: Screenreader users might be using a locale other than English. We also localize the [hidden attribution text](https://github.com/openstreetmap/openstreetmap-website/blob/801aeaa43a7bc5ec3b6f9fba530980cf55b33621/app/views/layouts/map.html.erb#L77) on the homepage and `<title>` tags inside SVGs, so localizing the “Loading” indicators would be consistent. Granted, this would probably affect a small segment of the user base.

app/assets/javascripts/application.js
mmd-osm: Good idea, I also thought about this before. `.focus()` is deprecated,  I'm using `.trigger("focus")` instead now.

app/views/layouts/_header.html.erb
mmd-osm: Form has moved, this should be fixed now.

app/views/layouts/_header.html.erb
mmd-osm: Moved back---> resolved.

app/views/layouts/_header.html.erb
mmd-osm: added now...

app/views/layouts/_header.html.erb
mmd-osm: fixed

tomhughes:
This is looking very good now - the only thing I've noticed is that sometimes (but not always) on first view the `shown.bs.modal` event can fire before the turbo frame has finished loading meaning that it doesn't find the elements it is looking for.

mmd-osm:
When adding a 5 second sleep in LanguagesPanesController, it was easy to spot that the search field wasn't properly focused. I've added another `$search .val("") .trigger("input") .trigger("focus");` to the turbo:frame-load event, to cover this case.

Since the turbo event doesn't fire if the data is still cached (and src hasn't changed), we have the exact same code in shown.bs.modal as well.

pablobm:
Very edge-case behaviour I noticed (Firefox and Chrome). When the modal is closed with the mouse, the focus returns to the "change language" button, but it's not `:focus-visible`. The focus is visible when closing with the keyboard (eg: escape, or tab to the close icon). I guess it's some browser heuristic at play. I can't force it with `focus({focusVisible: true})`.

Not sure it's a big deal, but I thought I'd mention it.

tomhughes:
I think this is probably good to merge now - thanks for all the work.

PR #6417, opened 2025-09-23
State of the Map Europe Banner

nomoregrapes:
This is a banner for the SotM Europe conference.

The banner should start being shown on the 6th October, as the current banner will be gone then (otherwise I would have requested it sooner).
With the config the banner will be shown for 1 month, as per the Banner Policy. If a delay impacts this, the latest the banner should show would be the 9th November.

Thank you.

### DIFF
--- a/config/banners.yml
+++ b/config/banners.yml
@@ -15,5 +15,5 @@ sotm_eu_2025:
   img: banners/sotm_eu_2025.png
   srcset:
     - [banners/sotm_eu_2025.png, 1x]
-  startdate: 2025-oct-06
-  enddate: 2025-nov-05
+  startdate: 2025-sep-24
+  enddate: 2025-oct-23


tomhughes:
Banners can have overlapping dates if you want - any banners that haven't been dismissed yet will be shown in turn in that case.

nomoregrapes:
@tomhughes that's a nice way of it working actually.
In which case I would like it to show ASAP and for a month. Will that need to be 1st October, so it meets the 7-day-notice requirement? Or can it be earlier if someone is able to merge it?

tomhughes:
The seven days isn't actually part of the OWG policy but something that is just in the FAQ here which is a bit odd... It's probably better read as that we can't guarantee to merge and deploy things immediately.

So I say set it to what you'd like and I can get it merged.

nomoregrapes:
Thank you Tom, I've changed the dates.

I appreciate the reason for some time frame in the FAQ, and the willingness to get it done quicker when possible.

PR #6418, opened 2025-09-24
Use Close as the button label to close modal alerts

gravitystorm:
The avoids using "OK" as a button label, and aligns with the labelling of the cross in the corner of the modal.

I considered alternative labels (like "dismiss") but since we have two ways to close the dialog, the cross is already labelled as "Close", so let's use "Close" for the button too.

Before:

Screenshot from 2025-09-24 14-22-16

After:
Screenshot from 2025-09-24 14-24-31


HolgerJeromin (approved):

tomhughes:
Looks good to me, thanks.

PR #6421, opened 2025-09-28
Send the user's preferred languages to Nominatim

### LINKED ISSUE #6420: Use user language preferences for queries to Nominatim
### Problem

When searching for a place, the website currently forwards the language preferences from the browser to Nominatim, which then returns a result which is localised according to these preferences. This seems to confuse users when they have some rather exotic fallbacks in their preferences. I had [two](https://github.com/osm-search/Nominatim/issues/3845) [complaints](https://github.com/osm-search/Nominatim/issues/3843) about Japanese in results within a week.

### Description

When the user is logged in, then it would be more consistent to just send the configured language to Nominatim.

### Screenshots

_No response_

tomhughes:
This sends the user's preferred languages to Nominatim instead of sending the browser languages - it will still fallback to the browser languages that were being sent before if we have no other source for preferred languages.

I also did (in a separate commit) some refactoring to improve how we build the URL for nominatim.

Fixes #6420.

### DIFF
--- a/app/controllers/concerns/nominatim_methods.rb
+++ b/app/controllers/concerns/nominatim_methods.rb
@@ -7,7 +7,7 @@ module NominatimMethods
 
   def nominatim_url(method, parameters)
     url = URI.join(Settings.nominatim_url, method)
-    url.query = parameters.merge("accept-language" => http_accept_language.user_preferred_languages.join(",")).to_query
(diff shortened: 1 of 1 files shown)


pablobm (approved):

gravitystorm:
Merged, thanks!

PR #6424, opened 2025-10-01
WIP: devcontainer configuration

pablobm:
Fixes https://github.com/openstreetmap/openstreetmap-website/issues/5264

Based on the conversation at https://github.com/openstreetmap/openstreetmap-website/issues/2344 (re: what to do with Vagrant and Docker), I have been playing with devcontainer a bit. This is not trivial stuff and requires a lot of trial&error (or perhaps I'm a dinosaur unable to understand these newfangled doohickeys), so for now I'm going to publish this draft for others to see and gather opinions and expertise.

To the extent I have been able to, I have followed "the Rails way". Parting from `bin/rails devcontainer` and iterating from there. This sort of works now: the dev server works and most test pass (need to look into the last few). There's a few caveats though that still need to be looked into:

- I can't copy files or run user commands (rails, bundler, yarn) in the Dockerfile, so these don't get committed in the image. I have to do it in `bin/setup` instead. As a result:
	- It installs Bundler dependencies every time I start the devcontainer.
		- Perhaps this can be avoided by installing gems to `.vendor` within the project (and gitignore'ing it)?
	- It copies config files such as `database.yml` or `settings.local.yml` both inside and outside the container. Therefore you can only configure the project to run with or without devcontainer, but not both.
- No idea what's different between the local and the remote ways to run Selenium (perhaps just a version number), but it appears that Selenium doesn't like posting forms where something has a value `"test"`. As a result I had to change all instances of `:password => "test"` to something else.
- So far I have tried only with VSCode. I need to test this with something else.
- I had to disable parallel test as they don't play well with this remote Selenium setup.
	- Perhaps if I was able to run a few Selenium containers in parallel...? No idea of what is feasible.
- The DB seed script doesn't quite work: users get the annoying error about scopes after log in. I took it from mmd's at https://gist.github.com/mmd-osm/a3a0b5a8799a333e7fae578038aa7a93. I need to look a bit more into it.

Good news:
- This appears to be running well on CI (still need to look into those failing tests).
- The number of required apt packages has been reduced to 3.

I'm sure I'm leaving something out... Will update. In the meantime: thoughts?


tomhughes:
Is it possible to mount a persistent volume in the container for gems, npms etc so that it will only have to install new ones when it started?

tomhughes:
Looking at it in more detail how exactly does the host/container split work? Is it mounting the checkout from the host to the `/workspaces/...` path in the container?

In which case I guess the gems and npms are already persistent? Certainly npms but where exactly is it installing gems to at the moment?

We'll need to replace `DOCKER.md` with some instructions on using this for sure.

The commit to get rid of `test` as a password can probably be pulled out to a separate PR as preparation for this then we probably want to squash this down to just a couple of commits once we're happy with it - maybe one to add devcontainer and the one to remove docker?

pablobm:
NPM dependencies should be ok as they live in `node_modules` within the work directory.

Ruby dependencies are handled with https://mise.jdx.dev and go in `/home/vscode/.local/share/mise/installs/ruby/`. I have now added configuration for a persistent volume and it seems to work. Thank you for mentioning, I do forget these things exist 😳

I have added a todo list to the description. Will go through these items in the next few days, a bit at a time. If others are able to test that this works in different environments, that'd be grand!

pablobm:
Found something very annoying while fixing `test/system/embed_test.rb`. I have left the relevant section of the PR untidy on purpose: discussing here before I clean it up properly, to see if people have thoughts. Commit: https://github.com/openstreetmap/openstreetmap-website/pull/6424/commits/7d992b9560cddbd1e1a0cb7e56ae2c04ec9e7c0e

From what I can tell, Rails+Capybara don't properly "reset" the remote Selenium when running a test with a different `driven_by` declaration. After trying a few things, what worked for me was to have different Selenium containers, one to be used for each configuration. This way we don't need to reset anything.

First issue I see is that this is not scalable. Not a problem right now as there are only three configurations. Could be a problem in the future if for any reason we want to test more.

pablobm:
Another funny one. Relevant commit: https://github.com/openstreetmap/openstreetmap-website/pull/6424/commits/cf421c07312452af6e28f6924dc019ad4c3c33db. The test [test/system/select_language_test.rb](https://github.com/openstreetmap/openstreetmap-website/blob/4191555f2846ac6e6f353b29c7d48f9ef9261efb/test/system/select_language_test.rb) fails with remote Selenium because there is no HTTPS, and therefore it can't set "secure" cookies.

I'm ignorant on this topic: should we bother with secure cookies? We force HTTPS anyway 🤔 Alternatively, we could have a setting in the JS that announces whether we are in production mode, and then set `secure: true` or not depending on that.

tomhughes:
The point of the `secure` flag is to ensure that the browser will never send the cookie over an insecure connection.

tomhughes:
That code should probably just be `secure: location.protocol === 'https'` or something.

pablobm:
I'm reading a bit more. I think the explanation (or at least one explanation) is that an attacker could impersonate the HTTP version of the site before the redirection to HTTPS. Hence we can't be sure that we are setting the cookie securely.

I think `location.protocol === 'https'` would not work as it would not protect us from those edge cases that `secure` is supposed to be about. I'll put a variable to signal that we are in production and use that.

tomhughes:
We already send `Settings.server_protocol` to the client as `OSM.SERVER_PROTOCOL` if you prefer to use that.

pablobm:
Right, that's all tests passing under the devcontainer, on my Mac and on CI. Other environments yet to be tested.

I solved the issue with the cookies using `OSM.SERVER_PROTOCOL` as suggested. I also did a refactor to DRY up the cookie handling. Another candidate for a separate PR? Commit: https://github.com/openstreetmap/openstreetmap-website/pull/6424/commits/60af1a16111da236fadeba9fa432408479ddc6fb

Still a few todo items to go through, plus the open questions. Non-devcontainers tests are failing because I'm missing a cleanup: I didn't want to commit too much time on that, just in case there are concerns about the caveats.

If you have opinions, I'd love to hear them!

pablobm:
Correction: a bunch of errors on CI (devcontainer workflow), after a flawless one earlier... Oh my.

pablobm:
Crisis averted. I had committed something silly.

.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,17 @@
+name: devcontainer
+
+on:
+  - push
+  - pull_request
+
```
hlfan, line 6: Could the superseded workflow cancellation work here too?

app/assets/javascripts/osm.js.erb
```diff
@@ -27,6 +27,8 @@ OSM = {
       end.to_json
   %>,
 
+  COOKIES: Cookies.withAttributes({ path: "/", samesite: "lax", secure: <%= (Settings.server_protocol == "https").to_json %> }),
```
hlfan: Since this isn't a constant, I'd like it to be in lowercase.

hlfan (commented):
I think the cookies clean up doesn't need that much of investigation ahead of pulling it into its own PR.

.github/workflows/devcontainer.yml
pablobm: Worth trying :-) Didn't add it initially in order to keep it simple while figuring it out.

Added now. Seems to work, I just pushed the cookies change and the previous run was cancelled.

app/assets/javascripts/osm.js.erb
pablobm: Done 👍

pablobm:
Cookies clean up extracted to https://github.com/openstreetmap/openstreetmap-website/pull/6430. Will remove from here when that is merged, as it's necessary for CI to pass.

pablobm:
Selenium `test`/`s3cr3t` workaround extracted to https://github.com/openstreetmap/openstreetmap-website/pull/6431. Again, will remove from here when that is merged.

.github/workflows/devcontainer.yml
hlfan: Does git figure out the file migration if the newlines are removed?

pablobm: I don't understand: what file migration? What newlines?

hlfan: If you remove the empty lines 2, 6, 10, and 17, does git think you renamed and edited docker.yml?

pablobm: Oh, right, no idea. I have never known what heuristics Git applies to identify files as "moved".

However, on reflection I prefer not to do it. They are different workflows, even if ultimately they are about running Docker. I think it makes more sense to accept them as separate.

hlfan: So I've now tried matching the style of `docker.yml`, `lint.yml`, and `tests.yml`, and, locally, Git does NOT conflate the history of the workflows then either. So if you've been keeping these around for that reason, it's unnecessary. And as Tom already suggested, rebasing the commits as "one to add devcontainer and the one to remove docker" would eliminate the chance of this happening accidentally anyway.

.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,21 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout (GitHub)
+        uses: actions/checkout@v3
```
hlfan: #6332 locked the dependencies to SHAs. I think this should be done here, too.

pablobm:
I'm going to change tack slightly and not remove the "manual" Docker config with this PR yet. My concern is about how widespread devcontainers actually are, as they require certain tool support and the manual methods to work with them are not trivial. I want to make sure we don't remove something that is working and useful and replace it with something less well understood that might be trickier to use.

To clarify: I don't mean that devcontainers are definitely more complex, just that I don't understand whether they are or not. I would like to hear opinions before removing the old way of using Docker.

DEVCONTAINER.md
```diff
@@ -0,0 +1,52 @@
...
+
+![VS Code: panel to install extensions](./docs/assets/vscode-devcontainers-extension.png)
+
+## Open the project in a container
+
+If everything is correct, this should make a few new commands available to you within VS Code. Find and run the command "Dev Containers: Reopen in Container".
+
+![VS Code: command to open in a devcontainer](./docs/assets/vscode-dev-reopen.png)
+
+The first time you do this, it will go into a bit of a process. It will create the devcontainer, pull the Docker images, install the dependencies, etc. Go drink some water while this runs.
```
hlfan, line 95: I found the part of the setup time for downloading the images and dependencies significant enough that I'd like an approximation of how many GB this downloads in the text once this PR is more finalized.

pablobm: I'll have a look, but it's going to be in the same ballpark as the Docker install that we already document.

pablobm: Just had a look. It appears to take up about 6GB of space :dizzy_face:

pablobm: For completeness, I don't know how much is actually downloaded. It could be the downloads are compressed to a fraction of the size.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Make sure RUBY_VERSION matches the Ruby version in .ruby-version
+ARG RUBY_VERSION=3.4.4
+FROM ghcr.io/rails/devcontainer/images/ruby:$RUBY_VERSION
+
+USER root
+ENV DEBIAN_FRONTEND=noninteractive
+
```
hlfan, line 7: To make it run on my Windows machine, I need to set additional permissions.
This got it to work, but it doesn't seem like the best solution.

```suggestion

RUN mkdir -p /workspaces/openstreetmap-website \
    && chown -R vscode:vscode /workspaces/openstreetmap-website
```

pablobm: Thank you so much for testing on Windows :slightly_smiling_face: There was a time when Microsoft provided virtual machines for Windows development, but that doesn't seem to be the case any more so I can't do it myself.

What is your environment? Windows version, WSL2 installed or not... not sure if there are other relevant questions.

hlfan: Win 11 with WSL2 and, as you may have guessed, VSCode.
Binding the permissions to that user could be meaningless for other IDEs.

pablobm:
@hlfan - Would you be able to write a short section on how to install Git on Windows? :sweat_smile: I copied the Docker instructions from `DOCKER.md`, but the Git ones are conspicuously absent.

DEVCONTAINER.md
```diff
@@ -0,0 +1,112 @@
...
+### Fedora
+
+```bash
+sudo dnf install git
+```
+
+### macOS
+
+For macOS, you will need [Xcode Command Line Tools](https://mac.install.guide/commandlinetools/); macOS 14 (Sonoma) or later; and some familiarity with Unix development via the Terminal.
+
```
hlfan, line 29: Do the Windows instructions need to be more than this?
```markdown
### Windows

With `winget`, you can install Git with this command:
```bash
winget install --id Git.Git -e --source winget
``ˋ

Alternatively, you can check out [other install options](https://git-scm.com/install/windows).
```

pablobm: I think that's enough. Thank you! ❤️

pablobm:
OK, I think this is now ready for review.

DEVCONTAINER.md
```diff
@@ -0,0 +1,139 @@
...
+If everything went well, you are done! For example, now you can open a shell in this environment using the VS Studio command "Create New Terminal (With Profile)":
+
+![VS Code: command to open a terminal](./docs/assets/vscode-create-terminal.png)
+
+From this terminal, you can run the test suite with `bin/rails test:all`:
+
+![Running the test suite in the terminal](./docs/assets/vscode-rails-test-all.png)
+
+Hopefully it should be all green? 🤞 You can also start a development server with `bin/rails s`:
```
hlfan: I found specifying the bin directory not to be necessary:
```suggestion
From this terminal, you can run the test suite with `rails test:all`:

![Running the test suite in the terminal](./docs/assets/vscode-rails-test-all.png)

Hopefully it should be all green? 🤞 You can also start a development server with `rails s`:
```

pablobm: Hm, that's a tricky one. I used `bin/rails` because it's my preference, but the rest of the documentation uses `bundle exec rails` which should be equivalent but is longer. In general I prefer not to use `rails` alone because in some environments that may pick up the wrong Rails. This typically happens if you have a mix of installs in your system, where there's a system Rails but the Gemfile specifies a different Rails (for example).

I have changed this to be `bundle exec rails` because it's the same as the rest of the documentation. Deep inside I'd rather we standardised to `bin/rails`, but I can live with it.

hlfan: Yeah that can be a follow up thing as long as it's consistent. Thanks.

config/devcontainer.database.yml
```diff
@@ -0,0 +1,103 @@
...
+# the values provided in this file. Alternatively, you can specify a connection
+# URL environment variable explicitly:
+#
+#   production:
+#     url: <%= ENV["MY_APP_DATABASE_URL"] %>
+#
+# Read https://guides.rubyonrails.org/configuring.html#configuring-a-database
+# for a full overview on how database connection configuration can be specified.
+#
+production:
```
gravitystorm: I don't think this section is necessary, and it's a bit confusing having references to migration paths that don't exist.

config/devcontainer.database.yml
```diff
@@ -0,0 +1,103 @@
...
+  #   log, notice, warning, error, fatal, and panic
+  # Defaults to warning.
+  #min_messages: notice
+
+# Warning: The database defined as "test" will be erased and
+# re-generated from your development database when you run "rake".
+# Do not set this db to the same as development or production.
+test:
+  <<: *default
+  database: open_street_map_test
```
gravitystorm: To reduce confusion, I think we should keep the same database names that are used in `docker.database.yml` and `example.database.yml`

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,34 @@
...
+    "ghcr.io/rails/devcontainer/features/postgres-client": {}
+  },
+
+  "containerEnv": {
+    "CAPYBARA_SERVER_PORT": "45678",
+    "DB_HOST": "postgres"
+  },
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
+  "forwardPorts": [3000, 5432, 6379],
```
gravitystorm: 3000 = rails app, 5432 = postgres, those seem reasonable. But as I understand it, 6379 is redis, and is that used at all?

Also just to check - do the ports get forwarded to the correct containers - specifically does 5432 get forwarded to the postgres container? I'm no expert on container routing.

DEVCONTAINER.md
```diff
@@ -0,0 +1,139 @@
...
+- [DevPod](https://devpod.sh): a CLI tool to work with devcontainers.
+
+## Troubleshooting
+
+### `‘ruby’: No such file or directory`
+
+In some cases Ruby may not install correctly. If you see this message, run these two commands:
+
+```
+mise install
```
gravitystorm, line 137: I don't understand what mise is used for, particularly when it's referring to installing ruby. Is the ruby binary not already installed in the container (`FROM ghcr.io/rails/devcontainer/images/ruby:$RUBY_VERSION`)?

I can see the mise cache being configured, but I can see any mise commands being run here (other than mise trust, see separate comment).

.devcontainer/start
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env sh
+
+# This setup uses https://mise.jdx.dev to manage language versions.
+# Mark the config file as trusted to avoid an interactive prompt.
+mise trust /workspaces/mise.local.toml
```
gravitystorm, line 5: I don't see a `mise.local.toml` file anywhere - where does it come from? Is it in one of the container images?

config/devcontainer.database.yml
pablobm: Makes sense. Removed this.

config/devcontainer.database.yml
pablobm: I have changed this file now to follow the existing ones closely :+1:

.devcontainer/devcontainer.json
pablobm: Good catch. The Redis port is a leftover that I failed to remove. The initial command `bin/rails devcontainer` generates a configuration that includes Redis, which I removed except for that detail.

As for the use of the list of ports... actually I have no idea. I'm no expert either, and I was playing with it now and I can't make any sense of it. I can't access Postgres from outside the container, and if I remove 3000 I can still access the web server with my browser. I'm now wondering if we should remove this line altogether as I don't know what it does at all :thinking: The documentation is not very clear.

DEVCONTAINER.md
pablobm: It's in the container image (you already figured this out by the next comment :wink:) which instead of providing a system Ruby and Node.js, uses mise to provide them. I guess it's so that multiple versions can be tested?

The readme (https://github.com/rails/devcontainer) does say that the image "includes an installation of Ruby" along with mise, but I can't find it anywhere (tried some `find /` lines and no dice). I wonder if it's outdated.

pablobm:
OK... I tried with VSCodium this afternoon and it definitely was not as straightforward as with VS Code, having to install DevPod manually and all. Also depending on the setup it may require you to manually ssh into the container.

Tried VSCodium with virtual machines:
- With Fedora I couldn't get it to work properly as I couldn't access the Internet from within the container and therefore couldn't run `bundle install` and so on. Perhaps a firewall setting in Fedora?
- With Debian I got it working.

I'm pasting a script below that I used to spare me repetitive installs in the VMs. It may help others get up and running:
```
# Debian/Ubuntu

for pkg in docker.io docker-doc docker-compose podman-docker containerd runc; do sudo apt remove -y $pkg; done

sudo apt-get update
sudo apt-get install ca-certificates curl
sudo install -m 0755 -d /etc/apt/keyrings
sudo curl -fsSL https://download.docker.com/linux/debian/gpg -o /etc/apt/keyrings/docker.asc
sudo chmod a+r /etc/apt/keyrings/docker.asc

# Add the repository to Apt sources:
echo \
  "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/debian \
  $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
  sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
sudo apt-get update

sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin

sudo docker run hello-world

wget -qO - https://gitlab.com/paulcarroty/vscodium-deb-rpm-repo/raw/master/pub.gpg \
  | gpg --dearmor \
  | sudo dd of=/usr/share/keyrings/vscodium-archive-keyring.gpg

echo -e 'Types: deb\nURIs: https://download.vscodium.com/debs\nSuites: vscodium\nComponents: main\nArchitectures: amd64 arm64\nSigned-by: /usr/share/keyrings/vscodium-archive-keyring.gpg' \
  | sudo tee /etc/apt/sources.list.d/vscodium.sources

sudo apt update && sudo apt install codium


sudo apt install -y fuse libfuse2 libopengl0 libfribidi0 libgles2-mesa-dev

# Fedora

sudo dnf remove docker \
                  docker-client \
                  docker-client-latest \
                  docker-common \
                  docker-latest \
                  docker-latest-logrotate \
                  docker-logrotate \
                  docker-selinux \
                  docker-engine-selinux \
                  docker-engine

sudo dnf -y install dnf-plugins-core
sudo dnf-3 config-manager --add-repo https://download.docker.com/linux/fedora/docker-ce.repo
sudo dnf install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
sudo systemctl enable --now docker

sudo rpmkeys --import https://gitlab.com/paulcarroty/vscodium-deb-rpm-repo/-/raw/master/pub.gpg
printf "[gitlab.com_paulcarroty_vscodium_repo]\nname=download.vscodium.com\nbaseurl=https://download.vscodium.com/rpms/\nenabled=1\ngpgcheck=1\nrepo_gpgcheck=1\ngpgkey=https://gitlab.com/paulcarroty/vscodium-deb-rpm-repo/-/raw/master/pub.gpg\nmetadata_expire=1h\n" | sudo tee -a /etc/yum.repos.d/vscodium.repo
sudo dnf install codium
sudo dnf install fuse fuse-libs libglvnd-egl libglvnd-opengl libglvnd-glx harfbuzz fontconfig fribidi libthai 

# Common

sudo usermod --append --group docker $(whoami)

sudo reboot

mkdir -p ~/.local/bin
curl -L -o devpod "https://github.com/loft-sh/devpod/releases/latest/download/devpod-linux-amd64" && sudo install -c -m 0755 devpod ~/.local/bin && rm -f devpod
devpod provider add docker
devpod provider use docker

sudo reboot

cd Documents
git clone https://github.com/pablobm/openstreetmap-website
cd openstreetmap-website
git checkout devcontainer

codium .

# Manually
# 1. Install "DevPod Containers" extension
# 2. A toast notification may invite to "Reopen in Container". Go for it. If not, find the command "DevPod: Reopen in Container".
```

I'm sure I don't need to say not to run it all at once! :wink: Apart from not being thoroughly tested, it includes steps for Debian/Ubuntu and Fedora, as well as two `reboot` lines. You may need to install Git first too (I used VM snapshots that already had it installed).

pablobm:
Tested with GitHub Codespaces. Mostly successful, but the `postCreateCommand` failed for reasons I couldn't quite tell, and I had to run `bin/rails db:create` manually and continue from there. Then tests ran successfully.

Also now VS Code is giving me trouble, tells me it can't authenticate into the container DB? My patience with this thing is wearing thin...

DEVCONTAINER.md
```diff
@@ -0,0 +1,141 @@
+# Using Development Containers for development and testing
+
+You can set up a development environment for this project using [Development Containers](https://containers.dev/), aka. devcontainers.
```
waldyrious: Since this is an acronym rather than an abbreviation, either use "a.k.a." or no punctuation at all :)

```suggestion
You can set up a development environment for this project using [Development Containers](https://containers.dev/), aka devcontainers.
```

DEVCONTAINER.md
```diff
@@ -0,0 +1,141 @@
...
+
+There are different ways of working with devcontainers, some are automated and integrated in IDE applications, while others are more manual and require CLI tools. This guide will use [Visual Studio Code](https://code.visualstudio.com) and assumes that it's installed already.
```
waldyrious:
```suggestion
There are different ways of working with devcontainers. Some are automated and integrated in IDE applications, while others are more manual and require CLI tools. This guide will use [Visual Studio Code](https://code.visualstudio.com) and assumes that it's installed already.
```

pablobm:
@waldyrious - Thank you :+1: Incorporated both edits now.

pablobm:
@hlfan - What else would you like to see in the docs? The size is already mentioned: "Be aware that the containers will take significant space in your hard drive, in the region of 6GB".

Currently I don't think this is good to merge as there's a test that is failing consistently. Also there's that suddenly, with no changes on my side, VS Code is behaving differently to what it did a month ago. Did they change something? See https://github.com/openstreetmap/openstreetmap-website/pull/6424#issuecomment-3553750993

I want to reserve some time to look into this again, but won't immediately. I want to look a bit more into how other Rails projects do this. I have already seen a couple that might be of interest.

I'll move this back to draft to make this explicit.

hlfan:
> there's a test that is failing consistently.

I already want to take a look at that one anyway, maybe it's just general docker weirdness.

pablobm:
@hlfan - Does it work without trouble for you? Remind me: are you using VS Code? Do you have to perform any manual steps for things to work, beyond what's in the docs?

hlfan:
Works without trouble or anything beyond the docs in VSCode. Since I couldn't be bothered to dig into why the recent DB migration confused Mise or RubyLSP (just some warnings), I let the setup download everything again.
2.28 GB were downloaded all in all (40 MB up); perhaps that could land here:

https://github.com/openstreetmap/openstreetmap-website/blob/e29b829375b5da10f579c7e1e5b047742525bfa2/DEVCONTAINER.md#L95

I don't have much of a sample size to see how #6589 affects the devcontainer CI runs, but Docker seems to get along better with it.
Otherwise, this is ready for me.

hlfan:
Well, now we know how #6589 doesn't affect the test failure.

pablobm:
Well, so much for a hope with a sample size of N=2 :disappointed: We should look into the "event not triggering" hypothesis.

pablobm:
At https://github.com/thoughtbot/administrate/pull/2948, I'm learning of the possibility to have a separate devcontainer configs to use in different contexts, such as quickly running on a GitHub Codespace as a "review app" without the need of pulling into your own repo, etc.

DEVCONTAINER.md
```diff
@@ -0,0 +1,141 @@
...
+
+```bash
+winget install --id Git.Git -e --source winget
+```
+
+Alternatively, see [other install options](https://git-scm.com/install/windows).
+
+## Install Docker
+
+Devcontainers run on top of Docker, which needs to be installed. Again, the methods will depend on your operating system.
```
danieldegroot2: S: Change grammar
```suggestion
You will need to install Docker in order to run devcontainers. Again, the methods will depend on your operating system.
```

or
```suggestion
Installing Docker is required in order to run devcontainers. Again, the methods will depend on your operating system.
```

or alternatively
```suggestion
Devcontainers run on top of Docker, which needs to be installed on your operating system. Again, the methods will depend on your operating system.
```
(though this introduces duplicate text 'operating system'.)

DEVCONTAINER.md
```diff
@@ -0,0 +1,141 @@
...
+
+In general: if you get an error about not having permission to use Docker, you may need to add your user to the `docker` group. The details vary, but often this is a case of running the following command:
+
+```bash
+sudo usermod --append --group docker $USERNAME
+```
+
+Where you must substitute `$USERNAME` with your actual username. After that, reboot your computer to ensure that this change takes effect.
```
danieldegroot2: S: Change capital letter to lowercase letter as it elaborates on the example ('[example], where ...')
```suggestion
where you must substitute `$USERNAME` with your actual username. After that, reboot your computer to ensure that this change takes effect.
```

or alternatively
```suggestion
You must substitute `$USERNAME` with your actual username. After that, reboot your computer to ensure that this change takes effect.
```
(if starting a new sentence is preferred.)

pablobm: Thank you. I went with "Substitute `$USERNAME` with your actual username in the command above".

INSTALL.md
```diff
@@ -8,14 +8,17 @@ There is more than one way to set up a development environment.
 
 ### Containerized Installation
 
-We offer a containerized environment with Docker which may avoid installation difficulties. See [DOCKER.md](DOCKER.md) for complete instructions.
+We offer containerized environments with Docker which may avoid installation difficulties:
+
+- To use Docker manually, see [DOCKER](DOCKER.md).
+- Do use Docker via [Development Containers](https://containers.dev) (devcontainers), see [DEVCONTAINER.md](DEVCONTAINER.md).
```
danieldegroot2:
```suggestion
- To use Docker via [Development Containers](https://containers.dev) (devcontainers), see [DEVCONTAINER.md](DEVCONTAINER.md).
```
Change spelling (typo).

pablobm:
@waldyrious @danieldegroot2 - Since you left reviews... :wink: Have you tried this setup? Did you encounter any issues?

pablobm:
@hlfan - Potential clue of what's going on with the failing test for the language selector. I tested the devcontainer again today and it failed. This is an excerpt of the screenshot I got:

osmw-language-selector-focus

This suggests that the browser gives the focus to the first element in the list (Afrikaans) instead of leaving it on the input box. If you try this on your browser (open the selector, press tab to move the focus) then you won't be able to type in the box any more, which is consistent with the test failure.

As to what could possibly be moving the focus, I have no idea.

danieldegroot2:
> Have you tried this setup?

No, I personally don't use code.

waldyrious:
I haven't used this either 😅

gravitystorm:
I've tried this out today, using [`@devcontainers/cli`](https://www.npmjs.com/package/@devcontainers/cli) and it worked for me. I successfully ran the entire test suite, and I got `rails server` working after a couple of small tweaks.

<details><summary>devcontainers-cli specific notes</summary>
<p>
The port forwarding is not supported by @devcontainers/cli ( see https://github.com/devcontainers/cli/issues/186 and https://github.com/devcontainers/cli/issues/22) but this can be worked around. First, launch the server bound to all IP addresses, not just localhost within the container, i.e. `bundle exec rails server -b 0.0.0.0`. Secondly, add the port forwarding to the compose file, i.e.

```diff
--- a/.devcontainer/compose.yaml
+++ b/.devcontainer/compose.yaml
@@ -6,6 +6,9 @@ services:
       context: ..
       dockerfile: .devcontainer/Dockerfile
 
+    ports:
+      - "3000:3000"
+
     volumes:
     - ../..:/workspaces:cached
     - mise-cache:/home/vscode/.local/share/mise
```

It would be great if `cli` had the port forwarding built in, since it would then be an easy substitute for having our standalone docker config - it will suit many people who might want to use Docker but who aren't interested in the IDE side of things.

</p>
</details> 

My next plan is to test with [DevPod](https://devpod.sh/) to see how the IDE integration side of things works.

But from my side, this looks good to me. What are the current blockers, if any? Are there any other specific things you want tested? 

Since this PR is additive, should we merge this and iron out any kinks later?

hlfan:
> What are the current blockers, if any?

> Currently I don't think this is good to merge as there's a test that is failing consistently. 

#6585 extends here as well, but I don't necessarily see that as blocking.

pablobm:
Agreed with @hlfan - The test `test/system/select_language_test.rb` appeared to fail more consistently with the devcontainer than with other workflows. (However it appears to work better now? Go figure). For me that's the blocker.

Regarding the actual development environment: it did something weird for me last month (https://github.com/openstreetmap/openstreetmap-website/pull/6424#issuecomment-3553750993), then it's working again this week. I think we can live with that and improve it over time.

Something we could do is merge this, but disabling the workflow until we figure out what's going on with the language selector. I can create a separate issue to track that, including the detail I discovered at https://github.com/openstreetmap/openstreetmap-website/pull/6424#issuecomment-3714563968

gravitystorm:
> My next plan is to test with [DevPod](https://devpod.sh/) to see how the IDE integration side of things works.

I've not made much progress with this today, but that's to do with things on the DevPod side of things, nothing in our repo.



> Something we could do is merge this, but disabling the workflow until we figure out what's going on with the language selector.

Yes please. I'd like to see if/how things work with CodeSpaces at least, in the meantime.

pablobm:
Created new issue to track the flaky test for the language selector: https://github.com/openstreetmap/openstreetmap-website/issues/6712

pablobm:
I have brought this out of draft now.

@gravitystorm: should the port forwarding config you shared above be included? I just played a bit with @devcontainers/cli and couldn't get the server to work despite your instructions. I stopped before I sank too much time on it. The test suite ran well.

If that's not an issue, then this might be ready to merge.

Re: disabling the GHA workflow, I don't know that there's a way to do it with code. Would a maintainer disable it on the GitHub UI please? https://docs.github.com/en/actions/how-tos/manage-workflow-runs/disable-and-enable-workflows

gravitystorm:
> @gravitystorm: should the port forwarding config you shared above be included?

No, I don't think so. The lack of port forwarding (and/or the subtly different "port publishing") is a devcontainers-cli limitation, and I'm not sure what else might break if we start including that stuff as a workaround. So best leave it aside for now.

hlfan:
> Re: disabling the GHA workflow, I don't know that there's a way to do it with code. Would a maintainer disable it on the GitHub UI please? https://docs.github.com/en/actions/how-tos/manage-workflow-runs/disable-and-enable-workflows

Done.

tomhughes:
I haven't had any much luck trying to use this with codespaces...

First it took forever to startup and then when it did it said there were migrations pending and offered to run them but that failed without telling me why - running them manually revealed that they had run but the schema dump at the end was failing as it was using pg_dump 15 against a postgres 16 server.

After all that it's still not functional as far as I can see because there's no server running and no way that I can find to start one.

mmd-osm:
Tried DevPod today w/ Docker as Provider. It took around 5 minutes to build and start up. Test cases were running ok:

image

pablobm:
> I haven't had any much luck trying to use this with codespaces

My opinion is that we merge this, for the purposes of development. Then we can have a separate PR with a different config for the purposes of using with codespaces, which is apparently a thing (example at https://github.com/thoughtbot/administrate/pull/2948).

There could be a discussion along the lines of: "well if the codespaces config works better, why not use that". My answer at the moment is that I have no idea of what works better in which cases. I'm currently advocating to go with this PR because it follows Rails convention, in the sense that it works on top of what is provided by the generator `bin/rails devcontainer`. If after releasing this and gathering more feedback we think we'd be best served by a different approach, then we can change.

tomhughes:
I was basically doing it because it was the easy way for me to test it but it sounds like I'll have to bite the bullet and see if I can get something working locally :cry:

PR #6429, opened 2025-10-05
Include preferred_username claim in OIDC ID token

ENT8R:
The id_token that is issued when the `openid` scope is set does not contain the `preferred_username` claim which would require external applications using OIDC to make another request to the API to be able to get the (current) user name to address users more personally.
I did not found any information if the decision to not include this claim in the token was deliberate? Otherwise this pull request would change the Doorkeeper configuration so the `preferred_username` is also included in the id_token

### DIFF
--- a/test/integration/oauth2_test.rb
+++ b/test/integration/oauth2_test.rb
@@ -110,7 +110,7 @@ class OAuth2Test < ActionDispatch::IntegrationTest
     end
 
     assert_equal user.id.to_s, data["sub"]
-    assert_not data.key?("preferred_username")
+    assert_equal user.display_name, data["preferred_username"]
 
     get oauth_userinfo_path
     assert_response :unauthorized
--- a/config/initializers/doorkeeper_openid_connect.rb
+++ b/config/initializers/doorkeeper_openid_connect.rb
@@ -27,7 +27,7 @@ Doorkeeper::OpenidConnect.configure do
   end
 
   claims do
-    claim :preferred_username, :scope => :openid do |resource_owner, _scopes, _access_token|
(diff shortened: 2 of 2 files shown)


test/integration/oauth2_test.rb
```diff
@@ -110,7 +110,7 @@ def test_openid_connect
     end
 
     assert_equal user.id.to_s, data["sub"]
-    assert_not data.key?("preferred_username")
```
pablobm, line 113: This assertion gives me pause, explicitly checking that the key is not present. Waiting to see what others say. If this PR is rejected, at least we should add an explanatory comment in this test.

tomhughes: Well I think it was always meant to be there - what else is that line there for?

The discussion in #4226 all seems to suggest it was intended to be there and the commit message for 64bcf7652bc1053741ab58f3ff54505b3d7820ad says it's returned as a claim for the openid scope...

Quite why that assertion wound up asserting the opposite I'm not sure!

ENT8R: I was also confused why the preferred_username is not in the id token but apparently all custom claims defined in the configuration are only returned by the userinfo endpoint and not added to the id token, if not specified explicitly otherwise:
> By default all custom claims are only returned from the UserInfo endpoint and not included in the ID token. You can optionally pass a response: keyword with one or both of the symbols :id_token or :user_info to specify where the claim should be returned.

<sup><i>https://github.com/doorkeeper-gem/doorkeeper-openid_connect#claims</i></sup>

tomhughes: Right so it's always been available in userinfo so there's really no point not including it in the token then.

tomhughes:

PR #6430, opened 2025-10-06
Use secure cookies only when HTTPS is expected

pablobm:
Two changes:
- DRY up the code handling JS cookies.
	- It's very repetitive as every single instance repeats the arguments `{ secure: true, path: "/", samesite: "lax" }`.
	- We can avoid this using `Cookies.withAttributes`, creating a preset that I have placed at `OSM.cookies`.
	- Argument `expires` left in place as it does change in each instance.
- Change `secure: true` to only apply when HTTPS is expected. This is, when `Settings.server_protocol == "https"`.

This change is extracted from https://github.com/openstreetmap/openstreetmap-website/pull/6424. Over there it was done so that system tests work with remote Selenium.

### DIFF
--- a/app/assets/javascripts/index.js
+++ b/app/assets/javascripts/index.js
@@ -171,16 +171,16 @@ $(function () {
       map.getLayersCode(),
       map._object);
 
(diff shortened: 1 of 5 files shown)


tomhughes:
Looks good to me, thanks.

Issue #6092, opened 2025-06-07
changesets: addresses buildings on the list


maro-21:
In changesets, the recently added display of the building address instead of the way number itself is using the `addr:housenumber` `addr:street` scheme. In my language the order is reversed - first there is the street name and then the house number. It should be possible for translators to change this.

hlfan:
This has been an issue for the query tool since its introduction.
It is relatively easy to let i18n do the templating, my preferred method and something I already implemented in cfb17435d1b11cd94aa865a6448e487b1a0cca17 before porting that logic to the browse helper.
While that attempt was shut down, see https://github.com/openstreetmap/openstreetmap-website/pull/6037#discussion_r2102834913, I'm delighted to see this issue gaining traction.

maro-21:
> something I already implemented in [cfb1743](https://github.com/openstreetmap/openstreetmap-website/commit/cfb17435d1b11cd94aa865a6448e487b1a0cca17)

So can it be merged too?

hlfan:
If you want to, you can copy this commit into a new PR, I'm not quite ready for another round of this.

pablobm:
Is this a locale issue or a location issue? 🤔

My personal experience is with Spain (street-number) and the UK (number-street). For me, referring in English to a Spanish address using number-street sounds odd, and vice-versa. Taking this to a Polish example, in English I would say "Długi Targ 43-44" rather than "43-44 Długi Targ". Does that make sense? This might just be me though, I wonder how other people do it.

If I'm correct, then we would need a different mechanism which takes into account the location instead of the locale... but wait I'm sure there are areas with more than one style 😓 (I don't know which though).

1ec5:
Address formats depend primarily on region and to a lesser extent on language. For example, [addresses in China](https://en.wikipedia.org/wiki/Address_format_by_country_and_area#China) are ordered from biggest to smallest in Chinese but smallest to biggest in English. To be totally correct, we would need some mechanism for determining the location of each address point in the changeset. Can we instead find a way to make this look like not quite a properly formatted address but just a human-readable representation of the raw tags?

hlfan:
> Can we instead find a way to make this look like not quite a properly formatted address but just a human-readable representation of the raw tags?

Isn't that what #721 tried to do? Remember that this also has to work in JS to be consistent.

1ec5:
[...]

PR #6435, opened 2025-10-08
Remove no-warning-comments eslint rule

### LINKED ISSUE #6434: ESLint: Fix "Unexpected todo" warnings
I find it distracting that on every PR I review, there's a section at the bottom that shows "Unchanged files with check annotations", and these are for the "Unexpected todo comment" checks.

Image

We should resolve this, with whatever works best from these options:

* Fix whatever these comments are referring to
* Remove the comments, if they are unnecessary
* Put these into some equivalent of `rubocop_todo`, but for eslint
* Disable this check
* ... or something else.

HolgerJeromin:
Fixes #6434

This is the fast fix which removes the warning, but does not fix the TODOs.

### DIFF
--- a/config/eslint.config.mjs
+++ b/config/eslint.config.mjs
@@ -146,7 +146,6 @@ export default [
       "no-useless-return": "error",
       "no-var": "error",
       "no-void": "error",
-      "no-warning-comments": "warn",
       "operator-assignment": "error",
       "prefer-const": "error",
       "prefer-object-spread": "error",


pablobm:
How about instead of hiding the warnings, we file issues for them, then remove the comments?

In this case, we have the following TODO comments:

- Localize format to country of address: in discussion at https://github.com/openstreetmap/openstreetmap-website/issues/6092
- Collapse width of sidebar back to previous: I have filed a new issue at https://github.com/openstreetmap/openstreetmap-website/issues/6442 (at least what I understand of it).

Now that we have issues we can remove the comments, leaving eslint to warn us again next time one appears.

PR #6438, opened 2025-10-09
Remove support for development environment with Vagrant

pablobm:
From the discussion at https://github.com/openstreetmap/openstreetmap-website/issues/2344, I think we agree that we can drop support for development with Vagrant virtual machines.

This PR does exactly so.


tomhughes:
Looks good to me, thanks.

PR #6443, opened 2025-10-13
Remove TODOs covered by GitHub issues

### LINKED ISSUE #6434: ESLint: Fix "Unexpected todo" warnings

hlfan:
Fixes #6434, closes #6435, keeps the ESLint rule.


tomhughes:
Looks good to me, thanks.

Issue #6442, opened 2025-10-13
Directions panel can't be brought back after collapse


pablobm:
### URL

https://www.openstreetmap.org/directions?engine=fossgis_osrm_foot&route=37.3908378%2C-5.9740888%3B37.3967759%2C-5.9941046#map=16/37.39367/-5.98409

### How to reproduce the issue?

On Desktop

1. Right-click anywhere on the map. Select "Directions from here".
    - A panel appears on the left (assuming desktop) with the To/From addresses.
2. Type a destination, press Enter.
    - Directions should be calculated and will appear on a new side panel under the previous one.

<img width="1162" height="815" alt="Image" src="https://github.com/user-attachments/assets/2e150b95-67f5-4c01-80c2-fbda045e5486" />

3. Click on the "X" icon to close the Directions panel (Not the panel with the To/From addresses).
    - The Directions panel will close. The To/From panel will stay open.

<img width="1162" height="815" alt="Image" src="https://github.com/user-attachments/assets/38a7d9c9-dc07-4d98-9285-6c96b135b46e" />

The issue: at this point it's not trivially possible to bring back the directions. Bringing the keyboard focus to a To/From input and pressing enter won't be enough to do this. Instead I have to delete a bit of an address and type it again to "wake up" the search facility and reset the behaviour.

What the exact behaviour should be, I can't tell yet. Just that the current one does not feel desirable to me.

PR #6446, opened 2025-10-14
Set getRoute default function parameters

hlfan:
An idea for closing #6442:
Use the selected mode of the mode group to trigger showing the route again after hiding the sidebar.
This does NOT request the route again.

Tested locally.

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -19,13 +19,10 @@ OSM.Directions = function (map) {
 
     getRoute(false, !dragging);
   };
-  const endpointChangeCallback = function () {
-    getRoute(true, true);
-  };
 
   const endpoints = [
-    OSM.DirectionsEndpoint(map, $("input[name='route_from']"), { icon: "start", color: "var(--marker-green)" }, endpointDragCallback, endpointChangeCallback),
-    OSM.DirectionsEndpoint(map, $("input[name='route_to']"), { icon: "destination", color: "var(--marker-red)" }, endpointDragCallback, endpointChangeCallback)
+    OSM.DirectionsEndpoint(map, $("input[name='route_from']"), { icon: "start", color: "var(--marker-green)" }, endpointDragCallback, getRoute),
+    OSM.DirectionsEndpoint(map, $("input[name='route_to']"), { icon: "destination", color: "var(--marker-red)" }, endpointDragCallback, getRoute)
   ];
 
   const expires = new Date();
@@ -95,7 +92,7 @@ OSM.Directions = function (map) {
     select.val(chosenEngine.provider);
   }
 
-  function getRoute(fitRoute, reportErrors) {
+  function getRoute(fitRoute = true, reportErrors = true) {
     // Cancel any route that is already in progress
     if (controller) controller.abort();
 
@@ -153,18 +150,18 @@ OSM.Directions = function (map) {
   modeGroup.on("change", "input[name='modes']", function (e) {
     setEngine(chosenEngine.provider + "_" + e.target.value);
     OSM.cookies.set("_osm_directions_engine", chosenEngine.id, { expires });
-    getRoute(true, true);
+    getRoute();
   });
 
   select.on("change", function (e) {
     setEngine(e.target.value + "_" + chosenEngine.mode);
     OSM.cookies.set("_osm_directions_engine", chosenEngine.id, { expires });
-    getRoute(true, true);
+    getRoute();
   });
 
   $(".directions_form").on("submit", function (e) {
     e.preventDefault();
-    getRoute(true, true);
+    getRoute();
   });
 
   $(".routing_marker_column span").on("dragstart", function (e) {


tomhughes:
I'm not sure we need to go to a lot of effort to avoid requesting the route again - if the user has closed the results and then requests them again is refetching them wrong?

hlfan:
Just that the button is not spammable, but if re-requesting the route should be the standard, then adding and removing the listener would be better.
I don't even think the push- & popstate logic needs to be touched.

pablobm:
I think this is a slight improvement, but doesn't solve the problem. As a user it's not obvious to me that, if I'm looking at a walking route and close the panel, clicking on "walking" again will bring it back.

I'm thinking of an alternative:
1. The X should close both the form and the results panels.
2. The X in the results panel should only show in narrow screens, when the form doesn't show the form panel (or other panels such as the announcement banners).

The rationale is the following:
1. On desktop, we can only close form+results together, avoiding this issue.
2. On mobile, the form doesn't appear so there's no X to close the whole thing. Show an X in the results to achieve the same effect.
3. In the edge case of someone resizing the window from small to large, having closed the results means they won't show up again when enlarging the window.

That's after a cursory test resizing windows myself, so I might be missing some use cases.

hlfan:
Nice idea! I'm curious what @AntonKhorev's opinion on this is as this is a different solution than the one proposed in https://github.com/openstreetmap/openstreetmap-website/pull/5960#issuecomment-2845105935.

pablobm:
Another dubious behaviour I have noticed: if you close the results panel, the route line disappears from the map, but the "start" and "end" markers remain:

Map, overlaid by the inputs for start and end of a route, showing no results, and showing the markers for start and end, but with no line joining them

hlfan:
Yeah, this kinda "configuration screen" could maybe be merged with an "overview" screen with only 1-3 lines of route instructions from https://c.osm.org/t/138969.
But I don't have an idea how to incorporate all that in one understandable way.
That may include reverting a part of #5960.
@mjourdan, do you have some ideas?

mjourdan:
Is it actually of any use to hide the directions while keeping the directions form visible? Hiding both at once sounds like the right thing to do, both on desktop and on mobile. 

On mobile the directions form beeing held in the burger menu, and then the results covering the burger menu, are issues in themselves. I'll do some mockups.

mjourdan:
Here is an example of a directions form on mobile. The form could slide in from the right when tapping the "directions" icon. It would hide out when tapping the "right pane" icon. Both the form and results would reopen at once when tapping again the "directions" icon. Clearing the results requires to clear either the start or the end field.

image

What do you think?

pablobm:
Thank you! Let's see if we can move this forward.

If these elements slide from the right, that'll conflict with the widescreen/desktop view where they live on the left-hand side. Technically it could be different on each case by use of a media query, but adds complexity.

On your second screen, the form is showing as well as the route line, but not the list of directions. How do users re-open the directions? Or I guess how do they open them in the first place, as there's no "submit".

Having the directions completely obscure the map on mobile... might actually work? Assuming there's a way to hide/show at will. I think I like it better than the current solution where it takes exactly half the screen.

mjourdan:
> Thank you! Let's see if we can move this forward.
> 
> If these elements slide from the right, that'll conflict with the widescreen/desktop view where they live on the left-hand side. Technically it could be different on each case by use of a media query, but adds complexity.

I don't feel like it adds complexity over the current situation, as the search and direction forms are hidden behind the burger menu.

> On your second screen, the form is showing as well as the route line, but not the list of directions. How do users re-open the directions? Or I guess how do they open them in the first place, as there's no "submit".

Yeah, I updated the mockups with some hints, to state that this second screen was meant as a transition. No direct way to go back to it from the fully expanded directions. Users would hide the directions by tapping the "right pane" icon on the second and third screen. They would then be able to reopen the directions by tapping the routing icon from first screen.

> Having the directions completely obscure the map on mobile... might actually work? Assuming there's a way to hide/show at will. I think I like it better than the current solution where it takes exactly half the screen.

As an alternative, using a bottom sheet rather than a popover would allow to keep both the form and the map visible:

image

pablobm:
Sorry, I meant complexity in the implementation. It would need a set of rules (and probably perhaps JS) for the wide version (desktop) and another set for the narrow version (mobile). I like the proposed bottom sheet, but it's a whole different UI to implement and maintain in parallel to the desktop one.

mjourdan:
Ok. Exploring the idea further with search, while keeping the buttons on the left:

image

pablobm:
Thank you @mjourdan - In terms of the original problem, what I take away from your proposals is to put the search form and the results in the same panel, and hide/show/resize the results with a "pull" interface. Does that make sense?

mjourdan:
Yes it does, though that's not all, so I'll try to sum up the changes.

Strictly related to search and directions:
- (mobile) search and directions forms are moved out of the navigation menu, and accessible through buttons on the map instead
- (mobile) search and directions button reveal a popover, which shows both forms and results
- a "collapse" button is used for hiding the popover
- resizing would be a thing only a thing if we were going to implement the bottom sheet pattern rather than the popover
- location results (bot for search and directions) appear in dropdown lists connected to their input fields. This allows people to select the most relevant location in the directions form, rather than being force to use the first matches.
- (not shown on mockups) selecting a location from the search results offers a mean to get to the location details
- when showing details about a location, the popover offers to navigate back to the results

As we add two more buttons on the map, the changes above depend on some more changes, mainly to avoid cluttering on mobile (these can perhaps be tracked through a separate issue, but here they are):
- move language selection to the navigation menu
- move layers and legend buttons to the navigation bar
- reorganize the navigation menu (not shown on mockups)
- move the zoom controls near to the scale line, as the former affect directly the latter
- move the scale line above the copyright notice, so it is not hidden behind the bottom leaflet on mobile

pablobm:
Thank you! This looks good but it has strayed a bit from the original topic. Perhaps we should open a new, separate issue to discuss a possible redesign of all those interface elements? @hlfan, what do you think?

As for this specific issue, I do like the pull interface.

tomhughes:
So I've been asked to review this but having tried to read back through the discussion I'm totally confused about what exactly it's supposed to be doing.

It was titled "Add functionality to reopen directions" and most of the discussion was about that but now the title has been changed to "Set getRoute default function parameters" which suggests it's now just a technical refactoring of some sort but I don't really see any explanation of that change in this discussion?

tomhughes:
OK it looks like it is now just a technical refactoring to simplify the code, and on that basis it looks fine to me.

hlfan:
> OK it looks like it is now just a technical refactoring to simplify the code, and on that basis it looks fine to me.

Yeah this was originally mostly a technical refactoring with some route retriggering mechanisms added, but doing it the right way is more involved. That's why I removed the second commit, as it isn't straightforward for users at all. See e.g. [this comment](https://github.com/openstreetmap/openstreetmap-website/pull/6446#issuecomment-3405757362):

> I think this is a slight improvement, but doesn't solve the problem.

Issue #1253, opened 2016-07-18
improve browse and history "pages" by highlighting changes


dieterdreist:
It would be convenient to show also summaries in object history and browse pages and highlight the kind of change, for example:
### for nodes+ways+rels:

amount of tags (useful for objects with lots of tags)
### for ways:

amount of nodes
### for rels:

amount of members (or nodes and ways and rels members)
### generally

highlight the kind of action by formatting (similar to how it is already done for deleted objects in changesets):
e.g. show new objects / tags bold (and or e.g. green)
show deleted tags (that are in version n but not n+1) as strike through (and or red) in version n+1
### for nodes+ways+rels:

amount of tags (useful for objects with lots of tags), separated into amount total+added+deleted (compared to previous version)
### for nodes:

position changed?
### for ways:

amount of nodes total,removed and added (compared to previous version)
### for rels:

type changed (only displayed if yes, e.g. type changed from x to y)
amount of node members total,removed and added (compared to previous version)
amount of way members total,removed and added (compared to previous version)
amount of role changes

HolgerJeromin:
see #738 
http://osm.mapki.com/history/way.php?id=4239284 has a nice interface for that. Nobody is currently working on including such a thing in the main page.

tomhughes:
Whilst useful, that's a bit "tech geeky" and I don't think it would really fit in the main page as is.

I'm also a bit reluctant to try and do even more analysis of the data because of the extra load it may add to the database - in some ways having these things done by third party services with their own copy of he data is better.

iandees:
I've been meaning to put together a pull request that does a simple version of what I did with osm.mapki.com/history. Basically, to highlight what changed in the HTML of the osm.org history browse page.

It's good to know that others might find that useful, too. I'm happy to work with someone on that if there's interest.

kleenkanteen:
@gravitystorm Would it make sense to close issues like this that have gone 1+ years with no activity?

danieldegroot2:
Related #1376

1ec5:
> I've been meaning to put together a pull request that does a simple version of what I did with osm.mapki.com/history. Basically, to highlight what changed in the HTML of the osm.org history browse page.

This is happening in #6448. I think it covers most of this feature request, so I’ll forward-duplicate this issue to #738, which will be closed by that PR.

> position changed?

[...]

PR #6448, opened 2025-10-15
Show tag changes in object history view

### LINKED ISSUE #738: Better history diff by adding colorization of changes
Spotting differences in an object's history can be really challenging, especially if there are lots of tags or only minor differences like changes in a value, spelling corrections or additions/removals of way or relation members.

This situation can be improved by adding colorization similar to [Mapki's Deep Diff](http://osm.mapki.com/history/). Example: 
history_diff

A similar colorization could be realized for changes in members of ways and relations as well as the object's coordinates.

iandees:
### Description

This adds highlighting of tag changes in each version to the history view for primitives:
- Tags that are **new** will be highlighted in green
- Tags that have **changed value** will be highlighted in yellow, with the value change described using an arrow like `old → new`
- Tags that have been **removed** will be highlighted in red, with the value `deleted`

Screenshot:
image

This probably closes https://github.com/openstreetmap/openstreetmap-website/issues/738. I know there are other discussions and tickets related to this idea, but I couldn't find them in my few searches.

Ruby is not my native language, so please point out where I can improve the change.

### How has this been tested?

I ran the site locally in Docker, added some fake data to the database, and inspected it on my browser locally.


iandees:
Apologies, I thought I had run linting successfully. I'll tackle the failing CI checks.

pnorman:
Unchanged tags look slightly odd with the white bar of a different width
image

iandees:
> Unchanged tags look slightly odd with the white bar of a different width

Agreed. What do you think about adding a grey bar of the same width?

image

For the record, I suggested the change to the border color (rather than the whole cell or row) in an attempt to avoid having to deal with text on odd color backgrounds.

app/views/browse/_tag_details_with_changes.html.erb
```diff
@@ -0,0 +1,18 @@
+<% unless tag_details_with_changes.empty? %>
+  <h4><%= t ".tags" %></h4>
+  <div class='mb-3 border border-secondary-subtle rounded overflow-hidden'>
+    <table class='mb-0 browse-tag-list table align-middle'>
+      <% if defined?(@old_features) && @old_features %>
+        <% tag_changes = tag_changes_for_version(tag_details_object, @old_features) %>
+        <% tag_changes.sort.each do |key, change_info| %>
```
pablobm: I think it's fine to have these in a single line:


```suggestion
        <% tag_changes_for_version(tag_details_object, @old_features).sort.each do |key, change_info| %>
```

On a quick search, I've seen two existing examples of what you are doing, but both look like something that should be moved to a helper instead ([app/views/layouts/_control_icons.html.erb](https://github.com/openstreetmap/openstreetmap-website/blob/6ff5918bd07673fb8db633701f77abeb16c5a1bd/app/views/layouts/_control_icons.html.erb#L1), [app/views/site/help.html.erb](https://github.com/openstreetmap/openstreetmap-website/blob/6ff5918bd07673fb8db633701f77abeb16c5a1bd/app/views/site/help.html.erb#L7-L8)).

app/views/browse/_common_details.html.erb
```diff
@@ -38,4 +38,8 @@
   <% end %>
 </div>
 
-<%= render :partial => "browse/tag_details", :object => common_details.tags %>
+<%= if defined?(@old_features) && @old_features
```
pablobm: This shouldn't be necessary in Ruby. Instance variables (starting with `@`) resolve to `nil` if not defined.


```suggestion
<%= if @old_features
```

app/views/browse/_tag_details_with_changes.html.erb
```diff
@@ -0,0 +1,18 @@
+<% unless tag_details_with_changes.empty? %>
+  <h4><%= t ".tags" %></h4>
+  <div class='mb-3 border border-secondary-subtle rounded overflow-hidden'>
+    <table class='mb-0 browse-tag-list table align-middle'>
+      <% if defined?(@old_features) && @old_features %>
```
pablobm: Same as above:

```suggestion
      <% if @old_features %>
```

app/views/browse/_tag_with_changes.html.erb
pablobm, line 1: There's a lot of duplication across `_tag`/`_tag_with_changes` and `_tag_details`/`_tag_details_with_changes`. I wonder if they can be merged into only `_tag` and `_tag_details`, with conditionals to show the changes only when available. In this case you will need `defined?` as `tag_details_object` is a local variable, as opposed to an instance variable that will be nil if undefined.

app/helpers/browse_helper.rb
```diff
@@ -72,6 +72,67 @@ def link_follow(object)
...
+    # Check for added and modified tags
+    current_tags.each do |key, value|
+      if !previous_tags.key?(key)
+        changes[key] = { type: :added, current: value }
+      elsif previous_tags[key] != value
+        changes[key] = { type: :modified, current: value, previous: previous_tags[key] }
+      else
+        changes[key] = { type: :unchanged, current: value }
+      end
+    end
```
pablobm: I try to avoid `each` for these things because it often means "anything could happen here!". Using idiomatic methods like `each_with_object` helps communicating that, eg: "we are going to build a hash step by step":

```suggestion
    changes = current_tags.each_with_object({}) do |(key, value), memo|
      if !previous_tags.key?(key)
        memo[key] = { type: :added, current: value }
      elsif previous_tags[key] != value
        memo[key] = { type: :modified, current: value, previous: previous_tags[key] }
      else
        memo[key] = { type: :unchanged, current: value }
      end
    end
```

app/helpers/browse_helper.rb
```diff
@@ -72,6 +72,67 @@ def link_follow(object)
...
+
+    # Check for deleted tags
+    previous_tags.each do |key, value|
+      unless current_tags.key?(key)
+        changes[key] = { type: :deleted, previous: value }
+      end
+    end
```
pablobm: Another way to do this. In Ruby you can do Array difference with `-`. Also you are not using the `previous` for deleted tags, so it can be removed:

```suggestion
    (previous_tags.keys - current_tags.keys).each do |key|
      changes[key] = { type: :deleted }
    end
```

I have to admit that in this instance I'm not 100% sure if my alternative is too clever. Thoughts?

pablobm (commented):
I love this ❤️ Some notes:
- The border in the middle (`border-start` of value cell) looks perhaps a bit too thin to me, particularly when contrasted with the one in the left (tag name cell). Probably not a big deal, but I wonder if other variants can be experimented.
- I see you added a background color for dark mode. To me it looks subtle enough not to clash with the text, so that's good. Still I wonder: is there a rationale for this difference?
- I wonder if there's a way to make this work with assistive technologies too. An `aria-*` tag or something, but I can't find anything right now and it's a very tricky subject anyway. Perhaps for a future PR.

app/helpers/browse_helper.rb
tomhughes: Maybe `previous_tags.keys.difference(current_tags.keys).each` is better?

app/views/browse/_tag_details_with_changes.html.erb
iandees: Made this to a single line in 8b1fb0bf6.

app/views/browse/_common_details.html.erb
iandees: Suggestion taken in 8b1fb0bf6.

app/helpers/browse_helper.rb
iandees: This is nicer, thanks. Improved in 8b1fb0bf6.

app/helpers/browse_helper.rb
iandees: I went with Tom's suggestion in 8b1fb0bf6 as it felt more descriptive.

app/views/browse/_tag_with_changes.html.erb
iandees: I tried this out in f56c397d1. It feels cleaner overall, but the partials look a bit messier. Let me know what you think.

iandees:
> I see you added a background color for dark mode. To me it looks subtle enough not to clash with the text, so that's good. Still I wonder: is there a rationale for this difference?

I was experimenting but didn't mean to include this in the PR. I'll remove it.

> I wonder if there's a way to make this work with assistive technologies too. An aria-* tag or something, but I can't find anything right now and it's a very tricky subject anyway. Perhaps for a future PR.

I'm happy to explore this, but I agree that it should probably be another PR.

> The border in the middle (border-start of value cell) looks perhaps a bit too thin to me, particularly when contrasted with the one in the left (tag name cell). Probably not a big deal, but I wonder if other variants can be experimented.

Here's a variant where the whole row is highlighted in the color:

image

Highlight the row with a stronger left border:

image

Here's one where just the value part of the table is highlighted:

image

1ec5:
This is a smart design, makes it a lot easier to find changes at a glance. My first impression is that we could probably even diff within a changed tag, word by word like MediaWiki does, for more clarity. But we can always fine-tune the presentation separately.

> I wonder if there's a way to make this work with assistive technologies too. An aria-* tag or something, but I can't find anything right now and it's a very tricky subject anyway. Perhaps for a future PR.

Along these lines, we should probably supplement the colors with some other visual distinction, because the traffic light colors aren’t great for colorblind friendliness. We already use these colors in the change counts in a changeset listing, but the user can depend on the position to distinguish between additions and deletions. Here the user can depend on deletions to be italicized and changes to have a “→”. But nothing distinguishes an added tag from an unchanged value. Not sure if that’s a problem with these particular colors. We should track colorblind friendliness in a separate issue if it isn’t trivial to deal with in this PR.

iandees:
Red/green colorblindness is definitely a thing, so what I have is suboptimal from that standpoint.

What about using the arrow to show something like "not present → <current value>" and "<current value> → not present", with "not present" in italics? It looks quite a bit visually noisier though:

image

HolgerJeromin:
We could use the git/diff wording.
So add `+` and `-` In the first column to tackle color blindness. For changed values this could be empty but there we have the arrow.

We even could get rid of the arrow and be more like diff:
```diff
  amenity cafe
- name oldname
+ name newname
+ opening_hours sometimes
- fixme add opening time
```
But please keep the color (green/yellow/red or (green/red depending on implementation).

gravitystorm:
I like the idea overall! Some small notes

> Red/green colorblindness is definitely a thing, so what I have is suboptimal from that standpoint.

Yeah, we will need some non-colour indications here

> What about using the arrow to show something like "not present → " and " → not present", with "not present" in italics? It looks quite a bit visually noisier though:

It's worth noting that not all languages have italics (e.g. Chinese) so we can't rely on italics to differentiate between user-generated tag values and system-generated UI text.

And I like the variations with the strong border and more gentle backgrounds. I wouldn't be surprised if the strong red background is too close to the blue link colour to pass the accessibility contrast checks, but I haven't actually checked the numbers on this.

pablobm:
I like the gentle backgrounds too :+1:

For colour-blindness, another option would be using <s>strikethrough</s>. For new tags we'd still need something else, here I'm using _(added)_:

screenshot using the style described above

app/views/browse/_common_details.html.erb
```diff
@@ -38,4 +38,4 @@
   <% end %>
 </div>
 
-<%= render :partial => "browse/tag_details", :object => common_details.tags %>
+<%= render :partial => "browse/tag_details", :object => common_details.tags, :locals => { :tag_details_object => common_details } %>
```
pablobm: How about renaming `tag_details_object` to `current_feature`? I had some difficulty reasoning about what it represented, due to the naming.

app/views/browse/_tag.html.erb
```diff
@@ -1,4 +1,12 @@
...
-  <th class='py-1 border-secondary-subtle table-secondary fw-normal' dir='auto'><%= format_key(tag[0]) %></th>
-  <td class='py-1 border-secondary-subtle border-start' dir='auto'><%= format_value(tag[0], tag[1]) %></td>
-</tr>
+<% if defined?(tag_with_changes) %>
+  <% change_info = tag_with_changes %>
+  <tr class="<%= tag_change_class(change_info[:type]) %>">
+    <th class='py-1 border-secondary-subtle table-secondary fw-normal' dir='auto'><%= format_key(change_info[:key]) %></th>
+    <td class='py-1 border-secondary-subtle border-start' dir='auto'><%= format_tag_value_with_change(change_info) %></td>
+  </tr>
+<% else %>
```
pablobm: I think we can remove this whole branch, as `tag_with_changes` is no more.

app/views/browse/_tag_details.html.erb
```diff
@@ -2,7 +2,16 @@
   <h4><%= t ".tags" %></h4>
   <div class='mb-3 border border-secondary-subtle rounded overflow-hidden'>
     <table class='mb-0 browse-tag-list table align-middle'>
-      <%= render :partial => "browse/tag", :collection => tag_details.sort %>
+      <% if defined?(tag_details_object) && @old_features %>
```
pablobm: I remembered now a better way to do this than using `define?`. You can put a `<% tag_details_object ||= nil %>` at the top of the file. If `tag_details_object` is not defined, it will be defined with a `nil` value.

As usual `a ||= b` expands to `a = a || b`. On any assignment, Ruby causes a definition of any variables on the right-hand side that didn't exist yet, so if `a` doesn't exist, it gets `a = nil` before entering the expression.

(But of course it would be even better if `tag_details_object` was also renamed to `current_feature` as mentioned above 😝).

pablobm (commented):
Still looking into how to better resolve the duplication in `browse/_tag_details`, but these are some quick notes.

1ec5:
<del>Strikethrough</del> is the default user agent styling for `<del>` tags. The corresponding `<ins>` tag gets <ins>underlined</ins>. Unfortunately, underlining gets confusing when text can be linked, and especially with `camel_case` keywords. We aren’t using **bold** for anything yet, and it would make it easier to see at a glance when something has been introduced in the element history.

app/helpers/browse_helper.rb
```diff
@@ -72,6 +72,63 @@ def link_follow(object)
...
+
+    current_tags = current_version.tags || {}
+
+    # Find the previous version by sorting all versions and finding the one before current
+    sorted_versions = all_versions.sort_by(&:version)
+    current_index = sorted_versions.find_index { |v| v.version == current_version.version }
+    previous_version = current_index&.positive? ? sorted_versions[current_index - 1] : nil
+
+    # Don't compare with redacted versions unless we're showing redactions
+    previous_tags = if previous_version && (!previous_version.redacted? || params[:show_redactions])
```
pablobm: Would it make sense to filter out the redacted versions when finding the index above, and show changes bridging versions, pretending that the redacted ones don't exist?

So for example: if we have v1, v2(redacted), v3, then show changes between v1 and v3

pablobm:
I gave it an attempt at refactoring: https://github.com/iandees/openstreetmap-website/compare/highlight-tag-changes-in-history...pablobm:openstreetmap-website:highlight-tag-changes-in-history History is clean and you can read each commit separately. Thoughts?

iandees:
> I gave it an attempt at refactoring: [iandees/openstreetmap-website@highlight-tag-changes-in-history...pablobm:openstreetmap-website:highlight-tag-changes-in-history](https://github.com/iandees/openstreetmap-website/compare/highlight-tag-changes-in-history...pablobm:openstreetmap-website:highlight-tag-changes-in-history) History is clean and you can read each commit separately. Thoughts?

That does look significantly better, thanks for taking the time to do that. Mind if I cherry pick your suggestions onto my branch?

pablobm:
Definitely, do as you please :+1: My branch was only an experiment.

app/helpers/browse_helper.rb
iandees: Yep, definitely. I tried to do this with 3530bce.

iandees:
Ok, I think the code is probably in a decent spot now. I'm going to fiddle with trying to improve the visualization a bit now.

iandees:
I added +/- for value changes. It's getting a little tight in there, especially when the value is a long string like a URL. It think it helps with vision impairments though:

image

HolgerJeromin:
> I added +/- for value changes. It's getting a little tight in there, especially when the value is a long string like a URL. It think it helps with vision impairments though:

But why adding the +- to the value cell?
I am sure this is pretty not intuitive that this is not part of the value.
At least visually the colored border is already a table column.

iandees:
> But why adding the +- to the value cell?

See conversation above. We are looking for something to show the change without relying solely on color.

HolgerJeromin:
> See conversation above. We are looking for something to show the change without relying solely on color.

hacked a bit in devTools:
image

iandees:
That looks better, but it's harder to distinguish between a tag's value changing and a tag getting added.

Maybe I'll change it so the key stays yellow but the value gets the red/green -/+ treatment.

1ec5:
> Maybe I'll change it so the key stays yellow but the value gets the red/green -/+ treatment.

In that case, I’ve seen some sites like Apple Developer use `~` as the symbol for a changed entry, as opposed to an unchanged one. For those who don’t know what these symbols mean, we could wrap each symbol in an `<abbr title>` tag. This will provide a tooltip, which screenreaders will read instead of the symbol itself.

app/helpers/browse_helper.rb
```diff
@@ -72,6 +72,71 @@ def link_follow(object)
...
+
+  def format_tag_value_with_change(key, change_info)
+    case change_info[:type]
+    when :added
+      tag.div(safe_join(["+", format_value(key, change_info[:current])], " "), :class => "diff-new")
+    when :unmodified
+      tag.div(safe_join([tag.nbsp, format_value(key, change_info[:current])], " "), :class => "diff-unchanged")
+    when :modified
+      safe_join([
+                  tag.div(safe_join(["-", format_value(key, change_info[:previous])], " "), :class => "diff-old"),
```
1ec5: Super super nitpicky, but the proper Unicode minus sign matches the width of the plus sign in most fonts, so it looks a bit tidier:

```suggestion
                  tag.div(safe_join(["\u2212", format_value(key, change_info[:previous])], " "), :class => "diff-old"),
```

iandees:
I've spent the evening stretching the limits of my CSS knowledge and getting fairly close to what I'm going for:

image

I'm not happy with this yet.
1. The +/- indicators should expand to fill the entire vertical space of the cell they're in
2. Something changed the vertical padding from 4 pixels to 8 pixels even though the style specifies the same 0.5 rem value

More later!

iandees:
I switched to using an extra `<td>` for the +/- indicators and that got it to look the way I wanted:

image

Happy to take feedback on this before spending much more time fixing tests.

Dimitar5555:
It looks great! But the plus and minus signs for value change looks confusing to me. Wouldn't it be better to use tilde (`~`) on yellow/orange background with `rowspan=2`?

HolgerJeromin:
> Wouldn't it be better to use tilde (~) on yellow/orange background with rowspan=2?

Than you would loose the info which is old an which is new.

app/helpers/browse_tag_changes_helper.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def format_tag_row_with_change(key, change_info)
+    case change_info[:type]
+    when :modified
+      # Generate two rows for modified tags
+      value_cells = format_tag_value_with_change(key, change_info)
+      [
+        tag.tr(:class => tag_change_class(change_info[:type])) do
+          safe_join([
+                      tag.th(format_key(key), :class => "diff-key-modified", :rowspan => 2),
```
1ec5: How does the rowspan interact with the Wikidata preview, which normally adds a row below the existing row?

app/assets/stylesheets/common.scss
```diff
@@ -1006,3 +1006,93 @@ img.trace_image {
...
+  color: var(--bs-body-color);
+}
+
+.diff-cell {
+  padding: 0.25rem 0.5rem !important;
+}
+
+/* Modified tag key cell styling */
+.diff-key-modified {
+  color: var(--bs-body-color) !important;
```
1ec5: What if the rightmost column still has a red or green background like you originally had, while the key column remains yellow? It would make the relationship between the ± column and this column a little clearer. (I realize this is going back and forth to an older design, so feel free to ignore this suggestion if it isn’t a good tradeoff.)

app/helpers/browse_tag_changes_helper.rb
iandees: Good question, I didn't know that was visible in the history view. I would propose removing it from the history view to reduce the complexity of this view. It would still be visible in the view of a single version.

app/assets/stylesheets/common.scss
iandees: How about this?

image

1ec5: Yes, that’s exactly what I had in mind, thanks!

1ec5:
> This probably closes https://github.com/openstreetmap/openstreetmap-website/issues/738. I know there are other discussions and tickets related to this idea, but I couldn't find them in my few searches.

Also partially addresses #1253, which describes some further ideas for turning the history page into a series of diffs.

app/helpers/browse_tag_changes_helper.rb
iandees: I made a change in fa4310f to remove the wikidata preview. It seems to work ... usually ... but it definitely makes the whole tag diff situation messier.

pablobm:
@iandees: what's the current status of this PR? Given the discussion about the UI, I was thinking of it as a draft, but it's not marked as such. Will you be iterating it more or should we consider it ready for final review? If not, perhaps best to mark it as draft explicitly :slightly_smiling_face:

1ec5:
I think @iandees has already responded to the outstanding visual feedback (see resolved review comments).

The only thing that might be debatable from a user standpoint would be the removal of Wikidata previews from history pages to keep the row-spanning key cell from getting mangled. If we think that’s a problem, it might be possible (if annoying) to split the key cell in half when showing the preview. Or the preview could occupy only the right-hand column on the history page. That said, the preview remains on version pages, so no information is actually lost, just slightly less convenient. We could ticket out restoring this functionality as tail work.

iandees:
Yep, the status of the PR is that the visual representation of tag history looks the same as the screenshot in [this comment](https://github.com/openstreetmap/openstreetmap-website/pull/6448#issuecomment-3443928967). The work that happened since is to fix lint/test/etc.

I'm happy to take more feedback, but it seems to be in a pretty good spot.

I think the wikidata lookup would probably be fairly easy to add back if that's important in this view.

app/assets/stylesheets/common.scss
```diff
@@ -1033,3 +1033,91 @@ img.trace_image {
...
+    background-color: rgba(var(--bs-success-rgb), 0.08) !important;
+    border-left: 4px solid var(--bs-success);
+
+    th, td {
+      background-color: inherit !important;
+    }
+  }
+
+  .tag-modified {
+    background-color: rgba(var(--bs-warning-rgb), 0.08) !important;
```
pablobm: Are all these `important!` declarations necessary? I just tried removing all of them, and the only change I could notice was taller column rows:

Current site | This PR | Removed `important!`
--- | --- | ---
Tags in changeset, as currently on production | Tags in changeset as in this PR, with differences highlighted | Tags in changeset, as in this PR, but columns are a bit taller

Mentioning because `important!` is frowned upon in CSS circles.

pablobm (commented):
I'm seeing two things when comparing the production and the PR renders side by side. This is for the normal list of current tags, as opposed to the history view with diffs:

Production | PR
--- | ---
Tags show with a grey background behind the keys, while values have a white background. Wikidata row has a 'barcode' icon to show further info | Tags show with an all-white background. There's no 'barcode' icon in Wikidata rows

On production, keys have a grey background and values a white background. On this PR it's an all-white background. (I'm not too bothered by this one, but I don't know what others think and whether they are aware).

On production, `wikidata` tags show the "barcode" icon to trigger an expand for more info. This is missing on the PR version.

app/assets/stylesheets/common.scss
tomhughes: Well one way or another I think we want the height to be consistent with the current behaviour, or at least with the non-history tag view, so if we're going to change it then it needs to change there too.

tomhughes:
> On production, keys have a grey background and values a white background. On this PR it's an all-white background. (I'm not too bothered by this one, but I don't know what others think and whether they are aware).

I think we definitely want some sort of distinction between keys and values and as with the height whatever it is needs to be consistent with the non-history view.

iandees:
> On production, keys have a grey background and values a white background. On this PR it's an all-white background. (I'm not too bothered by this one, but I don't know what others think and whether they are aware).

I've been checking the history view but didn't think to check how these changes affect the standard tag view. I'll look at that.

> On production, `wikidata` tags show the "barcode" icon to trigger an expand for more info. This is missing on the PR version.

See the discussion above about this – I decided that it would be too confusing to have in this already-complex view and took it out to simplify. The functionality exists in the standard tag view, though. I can try to add it back in if we think it's important in the history view.

> the height whatever it is needs to be consistent with the non-history view

The rendered height of the tag rows is consistent. I'll take a look at what I can do to get rid of the `!important` flags in the CSS, though.

app/assets/stylesheets/common.scss
pablobm: That's fine. My question is more: do we need *all* of the `important!` tags or can we lose a few of them?

pablobm:
Re: Wikidata references, what I mean they are gone from the standard tag view too, unless I'm misunderstanding something. Eg: the tags that appear on the sidebar when I visit `/way/:id`

pablobm:
> The rendered height of the tag rows is consistent

If I'm seeing this correctly, they are consistent internally (with your code changes, they are all the same), but not consistent with the current one on production (difference between production and PR).

My view is that this is not necessarily bad. I think I find the additional spacing nicer in the eyes. If "correcting" this results in more complex code, it might not be worth it. I'm only highlighting it to make sure everyone is ok with it.

tomhughes:
Yes I'm not necessarily concerned about changing it (though I'd need to look at the result first) but it does need to be consistent and I was assuming that this would break consistency as I thought it was only touching the history view.

iandees:
I've removed the `!important` flags and replaced them with some more specific CSS. This also means the other tag views aren't affected by the change to the tag history view.

I also made a small change to the Wikidata Javascript so that it works with the tag diff, so I added that back.

The row heights are fixed as well. This one was a bit confusing to me because it looked great locally, but I discovered I have to rebuild CSS and restart the docker compose system for every change to the CSS to see all changes.

This is where we're at:

image

pablobm:
I'm seeing issues with the wikidata tag :disappointed: 

Expanding the "removed" value breaks the "added" value:

changeset-diff-expanded-removed

Also expanding the "added" value places the expansion in the key cell when I would expect it in the value cell:

changeset-diff-expanded-added

To clarify: I'm ok with the wikidata info not expanding in the diff view. It was just that it had disappeared in the normal view.

iandees:
Ah yep, I didn't try this when the wikidata tag changes, only when added or removed. I don't think there's a great option here. The value cell isn't enough room to show the Wikidata preview, and adding it as another row below the change row is a bit confusing. I'll probably go back to removing it in the history view.

1ec5:
Yeah, the only way to make the layout work would be to temporarily (?) split and duplicate the key cell so that it no longer spans two rows, then insert the preview row immediately beneath the row that the user clicked on. But that would break the visual link between the old and new values for that key. I don’t think getting this feature to work on the history page should be a blocker, since the preview is only a convenience and one that hasn’t been around for very long. The user can still access all the same information on Wikidata by clicking on the QID. We can ticket out restoring the feature to this page.

iandees:
Alrite, I removed the Wikidata preview in the history/tag diff view while keeping it in the other tag views.

Also, I believe I found the missing class that was causing the table cells to get a couple extra pixels of top and bottom padding.

pablobm:
I think this is good? The one thing I'd comment is that the unchanged values appear with a grey background (which was the case in a previous version but I guess I didn't notice):



This is not necessarily bad. If we are comparing changes, then the grey helps highlighting the actual changes. This may not be great for accessibility though, poor contrast and all, but the normal view does show the keys with that same background, so I'm not sure of how big a deal it is.

waldyrious:
> ~Strikethrough~ is the default user agent styling for `<del>` tags. The corresponding `<ins>` tag gets underlined. Unfortunately, underlining gets confusing when text can be linked, and especially with `camel_case` keywords. We aren’t using **bold** for anything yet, and it would make it easier to see at a glance when something has been introduced in the element history.

Apologies for the late comment but I don't see additional discussion about this. I am confused about the `<del>`/`<ins>` approach having been abandoned, since the removed parts are pretty evident with the strikethrough formatting, and as for the new parts, CSS could be used to omit the underlining, since the absence of strikethrough should be enough to distinguish the old vs. new values.

I'm especially puzzled that this additional marking of the values (besides the coloring), which is being considered due to accessibility concerns, has moved away from the semantic HTML tags that would make the result universally unambiguous not just for those with color vision limitations, but also for the blind users relying on screen readers. (And it would even be machine-readable!)

iandees:
> I am confused about the `<del>`/`<ins>` approach having been abandoned ... moved away from the semantic HTML tags that would make the result universally unambiguous not just for those with color vision limitations, but also for the blind users relying on screen readers

This option wasn't abandoned or moved away from. It was never attempted.

Can you suggest a specific change I should try?

1ec5:
> the unchanged values appear with a grey background

This means a version that only changes geometry will have a completely grayed-out table. I think it could help visually differentiate the history page from the version page.

> Apologies for the late comment but I don't see additional discussion about this. I am confused about the `<del>`/`<ins>` approach having been abandoned, since the removed parts are pretty evident with the strikethrough formatting, and as for the new parts, CSS could be used to omit the underlining, since the absence of strikethrough should be enough to distinguish the old vs. new values.

Do you suggest using `<del>` and `<ins>` tags in addition to the + and − indicators? Or do you mean we should return to inlining a changed tag, using `<del>` and `<ins>` to distinguish old and new values? Inlining could be useful if we later introduce per-word or per-character diffing, to make subtle changes easier to spot. If so, I agree with using CSS to nix the underline.

Without that more granular diffing, this more structured design would be more effective at highlighting subtle changes. If we keep it, we should still make the + and − indicators more screen reader accessible: wrap each of these symbols in a `<abbr title>` tag that the screen reader can expand to something like “Removed” or “Added”. (The tooltip text should be localizable.)

waldyrious:
> This option wasn't abandoned or moved away from. It was never attempted.

Ah, my bad — I totally overlooked the author of [this comment](https://github.com/openstreetmap/openstreetmap-website/pull/6448#issuecomment-3410364736) :sweat_smile:

(And responding to your question, my suggestion would be precisely that one — not sure if implementing it would bring more value than the mock-up in that comment, though, at least until a decision is made about whether it's desirable, or concrete use cases where its usage may not be optimal are identified.)

> Do you suggest using `<del>` and `<ins>` tags in addition to the + and − indicators? Or do you mean we should return to inlining a changed tag, using `<del>` and `<ins>` to distinguish old and new values? 

IMHO the latter should be sufficient, since for regular vision users the colors alone would convey the necessary information, and that way we'd avoid the extra visual complexity.

> Inlining could be useful if we later introduce per-word or per-character diffing, to make subtle changes easier to spot.

Agreed, but to be clear I'd argue it'd still be useful with line-based diffing.

> Without that more granular diffing, this more structured design would be more effective at highlighting subtle changes.

By "more structured design" do you mean the +/- indicators? And if so, can you explain what is it that makes it more effective at highlighting subtle changes? I'm not sure I follow.

1ec5:
> By "more structured design" do you mean the +/- indicators? And if so, can you explain what is it that makes it more effective at highlighting subtle changes? I'm not sure I follow.

Because the value cells are left-aligned (or right-aligned in a right-to-left environment), showing the new value under the old value can help the user to spot any minor changes just from the characters being misaligned between the two cells. For example, [this relation](https://www.openstreetmap.org/relation/18322858/history/10)’s fairly long `description=*` is edited to replace “shift” with a naughty word one letter shorter, then showing the full new value under the full old value would make that change more apparent. But showing both the old and new values together on one line in their entirety, separated by a →, would make it more difficult to spot that one-letter change. In the future, an inline diff word by word would also make the change apparent, so we could go back to combining the old and new values in a single cell.

HolgerJeromin:
> since the removed parts are pretty evident with the strikethrough formatting

Is it? in all languages the tags contains?
Thinking about asian fonts here...

https://en.wikipedia.org/wiki/Strikethrough#:~:text=In%20Japan%2C%20double%20strikethrough%20is%20conventionally%20used%20because%20in%20complex%20kanji%2C%20a%20single%20strikethrough%20may%20be%20missed%20or%20confused%20with%20a%20stroke%20in%20the%20character.
says:
> In Japan, double strikethrough is conventionally used because in complex kanji, a single strikethrough may be missed or confused with a stroke in the character.

waldyrious:
@1ec5, I wouldn't tie the `<ins>`/`<del>` approach to showing things in one line with a →, as opposed to in separate lines. I do agree that having the changes in separate lines makes it easier to spot differences, and indeed @pablobm's mockup [above](https://github.com/openstreetmap/openstreetmap-website/pull/6448#issuecomment-3410364736) shows how this would look like, in the `leaf_type` row. I certainly prefer that mode to the inline display that can be seen in the `circumference` row.

@HolgerJeromin just briefly responding since you quoted me, but with the caveat that I don't know Japanese, so I'll leave it to others who might have a more informed opinion to properly respond to the concern you raise. I'll just say I would assume that, given the standard rendering of the `<del>` tag in browsers as strikethrough, Japanese users would at least be used to its effect, and so I suspect it might be inconvenient, but not completely unreadable.

1ec5:
Strikethrough can somewhat obscure a <del>- hyphen</del>, <del>– dash</del>, or <del>− minus</del> just like a Chinese <del>一 one</del> or Japanese <del>ー long vowel mark</del>. It might be more of a problem on 1× displays than 2× displays, where the strikethrough can be half as thick. I’m less concerned about it than underlines, since an <ins>_ underscore</ins> has a special role in OSM tagging and we tend to [`link_values`](https://github.com/openstreetmap/openstreetmap-website/pull/6448#issuecomment-3556958546) with underscores in them.

I guess there’s a potential caveat if someone tries to print out the page, since we’re relying on the cell backgrounds to provide contrast against the ± indicators. Strikethrough could help with that and the case where someone disables the stylesheet. But I’m pretty sure we haven’t given these use cases much consideration lately. If the only concern is about screen reader behavior, then we could wrap the old value in `<del>` and the new value in `<ins>` and use CSS to eliminate any styling of both tags in this table. Otherwise, I would view the suggestion for strikethrough as an alternative to the ± indicators, because if we do both then we’re starting to really overengineer the table with four visual differences between old and new.

pablobm:
For completeness, I just realised that we use strikethrough when showing the features removed in a changeset:

List of ways and nodes affected in a changeset, where a way is rendered striken-through (is that correct English?)

Differences:
- From my experiments, it appears to only show the feature id number and version. This avoid the issue with Chinese/etc scripts.
- Strikethrough is done with an `<s>` tag rather than a `<del>`. I don't see a counterpart `<ins>` or equivalent in added features.

waldyrious:
> If the only concern is about screen reader behavior, then we could wrap the old value in `<del>` and the new value in `<ins>` and use CSS to eliminate any styling of both tags in this table.

I do agree that using the semantic tags with the usual CSS striped out would be desirable, since their meaning would be conveyed directly in screen readers, and by the colors (as progressive enhancement) for those with normal vision.

> Otherwise, I would view the suggestion for strikethrough as an alternative to the ± indicators, because if we do both then we’re starting to really overengineer the table with four visual differences between old and new.

I'd be inclined to share that view, but I'd like someone who's actually affected by color issues to comment on whether this is an acceptable solution. I mean, if we do hide the visual signals of `<ins>` and `<del>` in normal rendering, then we'd be leaving them with nothing but color to distinguish the nature of the changes. We could assume that people who are affected by red/green styling would be likely to have custom user stylesheets or browser extensions to enhance color-only encoding of diffs, since that's a pretty common pattern; but I wouldn't want to take that decision (to deem color-only distinction sufficient in non-screen-reader contexts) on their behalf.

By the way, you raised a good point about e.g. printing the page — I'd suggest that matching rules be set up in [`print.scss`](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/stylesheets/print.scss) to either restore the strikethrough styling (and possibly also underline, though maybe using an alternative CSS property like `border-bottom` to shift it downwards from the text baseline), or use pseudo-elements to prepend `+` and `-` indicators, when color is absent.

Anyway, we're pretty much veering towards bikeshedding now, so just to be clear, I'd be pretty happy with something like [this](https://github.com/openstreetmap/openstreetmap-website/pull/6448#issuecomment-3489346156) (with grey for unchganged values, as shown [here](https://github.com/openstreetmap/openstreetmap-website/pull/6448#issuecomment-3516849705)), as long as the changed values are wrapped in `<ins>`/`<del>` (regardless of their visibility).

1ec5:
> For completeness, I just realised that we use strikethrough when showing the features removed in a changeset:

The browse view currently <del>crosses out</del> any element that today is deleted or redacted, regardless of whether the changeset you’re viewing was responsible for deleting it. Otherwise, you wouldn’t see any crossed-out elements, since this PR only deals with tag changes. Eventually it would be just as useful to highlight changes to members or coordinates in the same manner. If so, we probably wouldn’t be able to rely on strikethrough (or the aural equivalent) to indicate both “this changeset removed this member” and “this member has since been deleted or redacted”.

> From my experiments, it appears to only show the feature id number and version. This avoid the issue with Chinese/etc scripts.

This is because most editors these days automatically remove all the tags when deleting an object (which just sets its visibility attribute). It remains possible to delete an object without deleting the tags using the raw API or [Level0](https://wiki.openstreetmap.org/wiki/Level0). Older editors like Potlatch used to keep the tags too, so one can still encounter crossed-out names and such. However, this is an edge case that I don’t think we need to be concerned about.

> Strikethrough is done with an `<s>` tag rather than a `<del>`.

#6558

mmd-osm:
> It remains possible to delete an object without deleting the tags using the raw API

That’s not correct, the API always deletes tags in this case. The client has no way to influence this behavior. I don’t remember what the old AMF code for Potlatch was doing, but since it’s gone for a long time already that shouldn’t really matter anymore.

1ec5:
Ah OK, I was confusing the ability to specify tags on a deleted element in Level0L with it actually working that way. 😅 I do occasionally stumble upon a long-deleted element with tags, but it would be difficult to find one now.

Edit: [Found one](https://www.openstreetmap.org/node/200288623), and it was indeed deleted using Potlatch 1.

mmd-osm:
You can add these tags in osmChange, which is more of a curtesy to human readers to make clear what is being deleted. The API will ignore it, though. —> https://wiki.openstreetmap.org/wiki/OsmChange

iandees:
Just to clarify, the only concrete feedback since I asked about it (I don't mean to sound negative – the conversation was good to have) was @waldyrious's comment [here](https://github.com/openstreetmap/openstreetmap-website/pull/6448#issuecomment-3564636673) that they would prefer the changes be wrapped in `<ins>` and `<del>` so that user agents can better understand what's going on.

I'll make this change later today and we can go from there.

iandees:
The change to `BrowseTagsHelper` to remove wikidata link previews made that module too long for rubocop, so I moved a bunch of that code out into a separate module. I'm happy to undo that change if we can adjust the rubocop "module too long" failure.

iandees:
Ok, the tag changes are wrapped in `<ins>` and `<del>` HTML tags now. I removed styling for those specific tags so they don't change the visual look from the screenshot I posted above.

tomhughes:
Could we try and clean up the commit history here... I've pretty sure there aren't really 33 separate units of change here and some of them are very obviously fixup commits that should be merged into the commit they are fixing.

iandees:
> Could we try and clean up the commit history here

Yep, absolutely.

pablobm (commented):
I think this is ok. The initial refactor and the markup-generating code make things difficult to follow, but it can be followed a bit better when going commit-by-commit.

If anything, the markup generation could be untangled a bit. I gave it a shot at https://github.com/iandees/openstreetmap-website/compare/highlight-tag-changes-in-history...pablobm:openstreetmap-website:highlight-tag-changes-in-history and I think it's an improvement there. Not sure if it worth pursuing elsewhere, after playing with it a bit more.

app/helpers/browse_tag_changes_helper.rb
```diff
@@ -0,0 +1,134 @@
...
+
+module BrowseTagChangesHelper
+  include BrowseTagsHelper
+
+  def wrap_tags_with_version_changes(tags_to_values, current_version = nil, all_versions = [])
+    # Find the previous usable version by looking backwards through all versions
+    previous_version = all_versions
+                       .find_index { |v| v.version == current_version }
+                       &.then { |index| index.positive? ? all_versions[0...index].reverse : nil }
+                       &.find { |v| !v.redacted? || params[:show_redactions] }
```
tomhughes: This is trying to cope with `current_version` not existing in `all_versions` but when would that happen? Not only would I expect it to always exist but I'd expect it to always be at the end? In which case this should work:

```suggestion
    previous_version = all_versions
                       .reverse
                       .drop(1)
                       .find { |v| !v.redacted? || params[:show_redactions] }
```

app/helpers/browse_tag_changes_helper.rb
```diff
@@ -0,0 +1,134 @@
...
+      # For added and unmodified show the current value without the wikidata preview
+      format_value(key, change_info[:current], :skip_wikidata_preview => true)
+    when :modified
+      # Return array of two values for the two rows that will be created
+      [
+        tag.del(format_value(key, change_info[:previous], :skip_wikidata_preview => true)),
+        tag.ins(format_value(key, change_info[:current], :skip_wikidata_preview => true))
+      ]
+    when :removed
+      tag.del(format_value(key, change_info[:previous] || "", :skip_wikidata_preview => true))
```
tomhughes: When will a removed key not have a previous value? As far as I can see `wrap_tags_with_version_changes` always includes it/

app/helpers/browse_tag_changes_helper.rb
```diff
@@ -0,0 +1,134 @@
...
+    when nil
+      # For unknown versioning show the current value with the wikidata preview
+      format_value(key, change_info[:current] || "", :skip_wikidata_preview => false)
```
tomhughes: Same here for the current value? As far as I can see it should always have one?

app/helpers/browse_tag_changes_helper.rb
```diff
@@ -0,0 +1,134 @@
...
+        end
+      ]
+    else
+      # Generate single row for other change types
+      cells = [tag.th(format_key(key), :class => "py-1 border-secondary-subtle table-secondary fw-normal")]
+
+      # Only add diff indicator cell if we're in a history/diff context
+      cells << tag.td(get_change_indicator_text(change_info[:type]), :class => get_indicator_cell_class(change_info[:type])) if change_info[:type]
+
+      cells << tag.td(format_tag_value_with_change(key, change_info), :class => "py-1")
```
tomhughes: Why break this down into multiple statements that append to a temporary array when the modified version just does it all inline?

tomhughes (commented):
In addition to the specific comments could you rebase this please as there are apparently some conflicts.

Having now tried this after reviewing the code I think a lot of my comments may be explained by the fact that this only works in the history list, so I guess `current_version` is not the current version in the normal sense but is the version currently being rendered?

I have to say I find it a bit odd that the changes aren't shown when looking at a specific version - it actually took me a while to find out how to trigger this to the extent that I was putting in debugging code to try and find out why it didn't seem to be working!

1ec5:
Some musing about next steps following this PR:

> I have to say I find it a bit odd that the changes aren't shown when looking at a specific version - it actually took me a while to find out how to trigger this to the extent that I was putting in debugging code to try and find out why it didn't seem to be working!

I experienced this too, but I think it’s a symptom of a broader issue. The history and version pages have distinct headings, but the tops of the pages still look roughly the same. I think moving the breadcrumb and pagination bar to the top of the page would address this confusion. It would be more convenient now that the two pages contain substantially different content (diffs on one, wiki previews on the other).

I also wonder if we should have a separate page for diffing two specific versions. Most of the time, when I link to a single version, the snapshot in time is more interesting than the changes from a previous version. But sometimes I want to call attention to the changes affecting a particular element in a particular changeset. Or I want to compare the cumulative effect of multiple revisions in a row, for example to determine whether a revert was complete or not.

To be clear, I don’t think these ideas would block the current PR by any means.

1ec5:
> I think moving the breadcrumb and pagination bar to the top of the page would address this confusion.

#6759

> I also wonder if we should have a separate page for diffing two specific versions.

#6761

iandees:
Alrite, I think I addressed review comments and this is ready for another look.

test/helpers/browse_tags_helper_test.rb
```diff
@@ -90,6 +90,7 @@ def test_format_value
     html = format_value("website", "example.com/page")
     assert_dom_equal "example.com/page", html
   end
+<<<<<<< HEAD
```
hlfan, line 93: This looks like a messed-up merge?

iandees: Damnit, yes.

iandees: This is an outdated revision and isn't in the resulting diff.

hlfan: Still, that should be cleaned up for git history/blame sake.

iandees: That makes sense. I'll do that along with any other feedback I get.

app/views/browse/_tag.html.erb
```diff
@@ -1,4 +1,6 @@
-<tr>
-  <th class="py-1 border-secondary-subtle table-secondary fw-normal" dir="auto"><%= format_key(tag[0]) %></th>
-  <td class="py-1 border-secondary-subtle border-start" dir="auto"><%= format_value(tag[0], tag[1]) %></td>
-</tr>
+<% key = tag[0] %>
+<% version_details = tag[1] %>
+<% rows = format_tag_row_with_change(key, version_details) %>
+<% rows.each do |row| %>
+  <%= row %>
+<% end %>
```
tomhughes, line 6: You kind of have to ask, once every line in your view is a ruby interpolation, is it serving any purpose as a view partial?

Given this is the only use of `format_tag_row_with_change` maybe that function should incorporate the rest of the logic here so there's just one interpolation?

Then once you've done that maybe the partial goes away altogether and we just call that helper in the one partial that currently invokes this partial, either iterating the collection there or changing the helper again to process the whole collection?

tomhughes (commented):
A couple of overall comments...

Firstly, please don't create "address review feedback" commits but instead fold the changes into the original commit(s) where the code being changed was introduced.

Secondly, maybe it's just me but to me the way this shows changes seems to be backwards to me... Consider this example:

Image

That way was `leisure=park` in v1 and changed `amenity=school` to in v2 but it looks like the change is happening in v1, which is logically impossible and hence why I spotted it. Then you realise it's actually showing the change backwards as well at which point it makes a bit more sense but is assuming you're reading backwards in time from the present version.

Maybe it's just the programmer in me and normal people expect it to work like that but it's not what a programmer used to reading diffs expects to see!

app/helpers/browse_tag_changes_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+  def wrap_tags_with_version_changes(tags_to_values, _current_version = nil, all_versions = [])
+    # Find the previous usable version by looking backwards from the end
+    previous_version = all_versions
+                       .reverse
+                       .drop(1)
+                       .find { |v| !v.redacted? || params[:show_redactions] }
+
+    previous_tags = previous_version&.tags || {}
+
+    tags_added = tags_modified = tags_unmodified = tags_removed = tags_with_unknown_versioning = {}
```
pablobm, line 15: I should have realised this earlier:

```ruby
a = b = {}
a[:foo] = 1
a # => {foo: 1}
b # => {foo: 1}
```

(By the way, this is the same in other languages. I just checked in Python and JS).

Therefore, the assignment needs to be split:

```suggestion
    tags_added = {}
    tags_modified = {}
    tags_unmodified = {}
    tags_removed = {}
    tags_with_unknown_versioning = {}
```

This doesn't solve the issue that @tomhughes found in a separate comment (incorrect changes assigned to each versions). Still looking into that one.

pablobm: In fact, I think this doesn't really matter in the end because each variable is re-assigned rather than mutated. But still good practice.

app/helpers/browse_tag_changes_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+
+module BrowseTagChangesHelper
+  include BrowseTagsHelper
+
+  def wrap_tags_with_version_changes(tags_to_values, _current_version = nil, all_versions = [])
+    # Find the previous usable version by looking backwards from the end
+    previous_version = all_versions
+                       .reverse
+                       .drop(1)
+                       .find { |v| !v.redacted? || params[:show_redactions] }
```
pablobm, line 11: I think the problem (as mentioned by Tom) is here. We never use `_current_version`, so we always get the same version as `previous_version`. All comparisons are to... is it the second to latest version? Whichever.

I think this is the correct code:

```suggestion
  def wrap_tags_with_version_changes(tags_to_values, current_version = nil, all_versions = [])
    # Find the previous usable version by looking backwards from the end
    usable_versions_in_order =
      all_versions
      .reverse
      .filter { |v| !v.redacted? || params[:show_redactions] }

    current_version_index = usable_versions_in_order.find_index { |v| v.version == current_version }
    previous_version = current_version_index && usable_versions_in_order[current_version_index + 1]
```

stillhart:
I wanted to mention that I tried to solve the same but went with a different approach, see #6995. If mine doesn't make it, it might be of interested anyway.

stillhart:
Honestly, I prefer my approach a lot more. It's way simpler, clear, doesn't break and especially doesn't confuse with existing. The diff shouldn't be in the state. The diff is the diff in between states. It's whats changed, not what is. Having the diff in the state at a given time makes it more ambiguous whether we look forward or backward. It's something to look into, not something that must be visible at all times. 

Further, having just both values above/below each other makes it clearly comparable by eye. 

Lastly, the KISS bootstrap colors fit all existing schemas. 

It looks like this in the existing view
image

and once expanded using no js it's all in there. 

image

PR #6449, opened 2025-10-15
Enable ActionMailer Previews

gravitystorm:
Previewing email notifications during development is hard. Previously I've set up the [mailcatcher](https://mailcatcher.me/) gem and configured the app to send emails to the right place. But nowadays rails has a built-in option - [ActionMailer::Preview](https://guides.rubyonrails.org/action_mailer_basics.html#previewing-and-testing-mailers), so let's use that.

The previews run using the development database, which may or may not already contain sufficient records (e.g. diary comments). So to avoid failures, we generate content using factory bot, and to avoid filling your development database with duplicate data, we run the previews inside a transaction.

If you manually create records in the database that collide with factory bot (e.g. users with the same email pattern as the factory) then you'll need to update your manual records to avoid collisions.

My hope is that this will make it much easier for new contributors to work on the email notifications. If this PR is merged, then I intend to add an example of each outbound email to the previews.

### DIFF
--- a/test/mailers/previews/user_mailer_preview.rb
+++ b/test/mailers/previews/user_mailer_preview.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+require "factory_bot_rails"
+
+class UserMailerPreview < ActionMailer::Preview
+  include FactoryBot::Syntax::Methods
+
+  # Wraps the preview in a transaction, so that no changes
+  # are persisted to the development db
+  def self.call(...)
+    preview = nil
+    ActiveRecord::Base.transaction do
+      preview = super(...)
+      raise ActiveRecord::Rollback
+    end
+    preview
+  end
+
+  def diary_comment_notification
+    recipient = create(:user, :languages => [I18n.locale])
+    diary_entry = create(:diary_entry)
+    diary_comment = create(:diary_comment, :diary_entry => diary_entry)
+    UserMailer.diary_comment_notification(diary_comment, recipient)
+  end
+end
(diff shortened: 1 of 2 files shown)


test/mailers/previews/user_mailer_preview.rb
```diff
@@ -0,0 +1,25 @@
...
+    preview = nil
+    ActiveRecord::Base.transaction do
+      preview = super(...)
+      raise ActiveRecord::Rollback
+    end
+    preview
+  end
+
+  def diary_comment_notification
+    recipient = create(:user)
```
tomhughes: This should probably set the user language so that the local selector in the preview works - something like:

```suggestion
    recipient = create(:user, :languages => [I18n.locale])
```

gravitystorm: Oh that's awesome, thanks for that!

gravitystorm:
I'd marked this as draft after I opened it, because I suddenly thought of some simplification that I wanted to try. But that didn't work out. So this PR is now ready for review as-is.

test/mailers/previews/user_mailer_preview.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "factory_bot_rails"
+
+class UserMailerPreview < ActionMailer::Preview
+  include FactoryBot::Syntax::Methods
+
+  # Wraps the preview in a transaction, so that no changes
+  # are persisted to the development db
+  def self.call(...)
```
pablobm, line 10: 6 years after being made available in the language, TIL about argument forwarding with `...` 😳 I have always used `*` and variants, but this seems easier.

tomhughes:
Looks good to me, thanks.

PR #6450, opened 2025-10-16
Rename "Map Key" to "Legend" in UI and tests for consistency

hlfan:
Following #6063, this PR aligns the labels with the terminology used everywhere else in this repo.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3539,9 +3539,9 @@ en:
     embed:
       report_problem: "Report a problem"
     legend:
-      title: "Map Key"
-      tooltip: "Map Key"
-      tooltip_disabled: "Map Key not available for this layer"
+      title: "Legend"
+      tooltip: "Legend"
+      tooltip_disabled: "Legend not available for this layer"
     map:
       zoom:
         in: Zoom In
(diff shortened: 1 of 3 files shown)


HolgerJeromin (approved):

config/locales/en.yml
pablobm, line 1: Would it be ok to add the `en-GB` translation as part of this PR too? Asking purely out of self-interest: I had it on and couldn't see the change until I realised and changed to `en`.

pablobm (commented):
Quick screenshot for future readers who might not know where things are (like I was myself :sweat_smile:).

Partial screenshot of the OSM Website, showing the Map Key, now renamed Legend. The toolbar is visible and the Legend icon is highlighted

config/locales/en.yml
tomhughes: I'm afraid not, no, as it causes problems for Translatewiki.

pablobm:
By the way, by "terminology used everywhere else", where do you mean? I see it used on the source code, is that it? Or do you mean outside the website?

hlfan:
Yeah, in this repo. That part I left out initially in #6113 and then forgot about it.

tomhughes:
Looks good to me, thanks.

PR #6452, opened 2025-10-17
feat: add social links and company to user api endpoint

### LINKED ISSUE #6451: Return `social_links` and `company`on `/api/user` endpoint
### Problem

It seems to me that social link were only added in 2025. However it seems nobody modified the `api/user` endpoint accordingly. The endpoint does unfortunately not return social_links. Would be nice if the API returned everything that' visible on the public profile pages.

### Description

_No response_

### Screenshots

_No response_

pReya:
### Description
- add missing `company` property to api user endpoint
- add missing `social_links` property to api user endpoint

closes #6451 

### How has this been tested?
Manual, local testing. Could not find any automated tests for API endpoint properties.

### DIFF
--- a/app/views/api/users/_user.xml.builder
+++ b/app/views/api/users/_user.xml.builder
@@ -4,6 +4,15 @@ xml.tag! "user", :id => user.id,
                  :display_name => user.display_name,
                  :account_created => user.created_at.xmlschema do
   xml.tag! "description", user.description if user.description
+  xml.tag! "company", user.company if user.company
+
+  xml.tag! "social-links" do
+    user.social_links.each do |link|
+      details = link.parsed
+      xml.tag! "link", details[:url], :platform => details[:platform]
+    end
+  end
+
   if current_user && current_user == user && can?(:details, User)
     xml.tag! "contributor-terms", :agreed => user.terms_agreed.present?,
                                   :pd => user.consider_pd
--- a/test/controllers/api/users_controller_test.rb
+++ b/test/controllers/api/users_controller_test.rb
@@ -38,7 +38,10 @@ module Api
                     :description => "test",
                     :terms_agreed => Date.yesterday,
                     :home_lat => 12.1, :home_lon => 23.4,
-                    :languages => ["en"])
+                    :languages => ["en"]) do |user|
+                      create(:social_link, :user => user, :url => "https://twitter.com/testuser")
+                      create(:social_link, :user => user, :url => "https://github.com/testuser")
+                    end
 
       # check that a visible user is returned properly
       get api_user_path(:id => user.id)
@@ -408,6 +411,18 @@ module Api
     def check_xml_details(user, include_private, include_email)
       assert_select "user[id='#{user.id}']", :count => 1 do
         assert_select "description", :count => 1, :text => user.description
+        assert_select "company", :count => 1, :text => user.company if user.company
+
+        if user.social_links.present?
+          assert_select "social-links link", :count => user.social_links.count do
+            user.social_links.each do |link|
+              details = link.parsed
+              assert_select "link[platform='#{details[:platform]}']", :count => 1, :text => details[:url]
+            end
+          end
+        else
+          assert_select "social-links link", :count => 0
+        end
 
         assert_select "contributor-terms", :count => 1 do
           if user.terms_agreed.present?
@@ -487,6 +502,26 @@ module Api
     def check_json_details(js, user, include_private, include_email)
       assert_equal user.id, js["user"]["id"]
       assert_equal user.description, js["user"]["description"]
+      if user.company
+        assert_equal user.company, js["user"]["company"]
+      else
+        assert_nil js["user"]["company"]
+      end
+
+      if user.social_links.present?
+        ordered_js_links = js.dig("user", "social_links").sort_by { |link| link["url"] }
+        ordered_user_links = user.social_links.sort_by(&:url).map(&:parsed)
+
+        assert_equal ordered_user_links.count, ordered_js_links.count
+
+        ordered_user_links.zip(ordered_js_links).each do |user_link, js_link|
+          assert_equal user_link[:url], js_link["url"]
+          assert_equal user_link[:platform], js_link["platform"]
+        end
+      else
+        assert_empty js["user"]["social_links"]
+      end
+
       assert_operator js["user"]["contributor_terms"], :[], "agreed"
 
       if include_private
--- a/app/views/api/users/_user.json.jbuilder
+++ b/app/views/api/users/_user.json.jbuilder
@@ -5,6 +5,14 @@ json.user do
   json.display_name user.display_name
   json.account_created user.created_at.xmlschema
   json.description user.description if user.description
+  json.company user.company if user.company
+  json.social_links do
+    json.array! user.social_links do |link|
+      details = link.parsed
+      json.url details[:url]
+      json.platform details[:platform]
+    end
+  end
 
   if current_user && current_user == user && can?(:details, User)
     json.contributor_terms do


app/views/api/users/_user.xml.builder
```diff
@@ -4,6 +4,14 @@ xml.tag! "user", :id => user.id,
                  :display_name => user.display_name,
                  :account_created => user.created_at.xmlschema do
   xml.tag! "description", user.description if user.description
+  xml.tag! "company", user.company if user.company
+  if user.social_links
```
pablobm: This will be truthy when there are no links, since in Ruby an empty array is truthy. If you want it not to show in the "no links" case, then do something like this:


```suggestion
  if user.social_links.present?
```

app/views/api/users/_user.xml.builder
```diff
@@ -4,6 +4,14 @@ xml.tag! "user", :id => user.id,
...
+    xml.tag! "social-links" do
+      user.social_links.each do |link|
+        xml.tag! "link", link["url"]
```
pablobm: Is it worth including the platform?

```suggestion
        details = link.parsed
        xml.tag! "link", details[:url], :platform => details[:platform]
```

And something similar for the JSON case.

app/views/api/users/_user.xml.builder
pReya: Since other arrays seem to be returned as empty arrays, too (e.g. "roles"), I did the same thing, and just removed the check.

app/views/api/users/_user.xml.builder
pReya: I've added it now.

pReya:
> This should be tested in `test/controllers/api/users_controller_test.rb` by extending `check_xml_details` and `check_json_details` and adding appropriate attributes to at least some of the users created by the test methods.
> 
> There's also a lint warning that needs resolving.

Thanks for the feedback and the hint regarding the tests. I tried to fix it, to the best of my ability. However, I'm not a Ruby programmer, so be gentle with me, please.

Full Disclosure: I've used AI/LLM Support for these changes (as much as I hate LLMs, for these kind of use cases (e.g. unfamiliar language) I find them rather useful), as my Ruby knowledge is very very limited.

pablobm:
@pReya: I had a look and improved the tests a bit. See my branch at https://github.com/openstreetmap/openstreetmap-website/compare/master...pablobm:openstreetmap-website:pablobm/feat/add-social-links-company-to-user-endpoint Feel free to incorporate to your branch.

test/controllers/api/users_controller_test.rb
```diff
@@ -487,6 +501,20 @@ def check_xml_details(user, include_private, include_email)
     def check_json_details(js, user, include_private, include_email)
       assert_equal user.id, js["user"]["id"]
       assert_equal user.description, js["user"]["description"]
+      assert_equal user.company, js["user"]["company"]
+
+      if user.social_links.present?
+        assert_not_nil js["user"]["social_links"]
+        assert_equal user.social_links.length, js["user"]["social_links"].length
+        user.social_links.each_with_index do |link, index|
```
tomhughes: This is assuming that `user.social_links` here will return the links in the same order as when the template was rendered.

Whilst that is very likely to happen there's no guarantee of it as the links are not fetched in any particular order so the test should probably just check that each expected link exists somewhere in the returned JSON rather than expecting the order to match.

tomhughes (changes requested):
Other than the test comment I left, could you squash the fixups down into the original commit please - see the tips at https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#pull-requests for advice.

pReya:
> @pReya: I had a look and improved the tests a bit. See my branch at [master...pablobm:openstreetmap-website:pablobm/feat/add-social-links-company-to-user-endpoint](https://github.com/openstreetmap/openstreetmap-website/compare/master...pablobm:openstreetmap-website:pablobm/feat/add-social-links-company-to-user-endpoint) Feel free to incorporate to your branch.

Thanks! I've incorporated all your changes. Very much appreciated.

test/controllers/api/users_controller_test.rb
pReya: I've modified it to be order-independent, now.

test/controllers/api/users_controller_test.rb
```diff
@@ -487,6 +502,26 @@ def check_xml_details(user, include_private, include_email)
...
+      if user.social_links.present?
+        js_links = js.dig("user", "social_links")
+        assert_equal user.social_links.count, js_links.count
+        user.social_links.each do |user_link|
+          details = user_link.parsed
+          matching_js_link = js_links.find { |js_link| 
+            js_link["url"] == details[:url] && js_link["platform"] == details[:platform] 
+          }
+          assert_not_nil matching_js_link
+        end
```
pablobm: Alternatively, you can sort them first and avoid the nested loop:

```suggestion
        ordered_js_links = js.dig("user", "social_links").sort_by { |link| link["url"] }
        ordered_user_links = user.social_links.sort_by(&:url).map(&:parsed)

        assert_equal ordered_user_links.count, ordered_js_links.count

        ordered_user_links.zip(ordered_js_links).each do |user_link, js_link|
          assert_equal user_link[:url], js_link["url"]
          assert_equal user_link[:platform], js_link["platform"]
        end
```

pReya:
@tomhughes Any chance this could get merged soon? Or are there any remaining tasks?

test/controllers/api/users_controller_test.rb
```diff
@@ -487,6 +502,26 @@ def check_xml_details(user, include_private, include_email)
     def check_json_details(js, user, include_private, include_email)
       assert_equal user.id, js["user"]["id"]
       assert_equal user.description, js["user"]["description"]
+      if user.company
+        assert_equal user.company, js["user"]["company"]
+      else
+        assert_nil js["user"]["company"]
+      end
```
tomhughes, line 509: This should probably have been a separate commit, as it's fixing an existing issue in the test rather than being anything to do with the main purpose of this PR but let's not worry about it this time.

pReya:
@tomhughes Any idea when this will be deployed on the live instance (https://api.openstreetmap.org/api/0.6/)? I assume there's a manual deployment step? So far I don't get either of these new fields on the public API.

pablobm:
@pReya - I think it should be live now. Deployment is not automatic on merge unfortunately. You can see what was last deployed by looking at https://git.openstreetmap.org/rails.git. Look for the tag `live`.

1ec5:
This is a public API change, right? We should update the API documentation:

https://wiki.openstreetmap.org/wiki/API_v0.6

pReya:
> This is a public API change, right? We should update the API documentation:
> 
> https://wiki.openstreetmap.org/wiki/API_v0.6

Done

simonpoole:
> This is a public API change, right? We should update the API documentation:
> 
> https://wiki.openstreetmap.org/wiki/API_v0.6

And changes to all the relevant privacy related documents, from the privacy policy to the GDPR related work lists?

1ec5:
At a glance, the [GDPR worklist](https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services#Methods_for_user_data) lists this API method overall as needing to be gated by ToU acceptance. I’ll follow up with the LWG about the social links feature as it relates to the privacy policy (now with the API field, but even without).

simonpoole:
> At a glance, the [GDPR worklist](https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services#Methods_for_user_data) lists this API method overall as needing to be gated by ToU acceptance. I’ll follow up with the LWG about the social links feature as it relates to the privacy policy (now with the API field, but even without).

Looking at the response from the LWG to the heat map issue, I'm fairly sure that they are not aware of the representations the policy makes (because all of the automatically generated content on the profile page is at odds with them). In any case I don't think there is a way around redoing the analytical work on the API and website if there was an attempt at compliance in earnest.

pReya:
> At a glance, the [GDPR worklist](https://wiki.openstreetmap.org/wiki/GDPR/Affected_Services#Methods_for_user_data) lists this API method overall as needing to be gated by ToU acceptance. I’ll follow up with the LWG about the social links feature as it relates to the privacy policy (now with the API field, but even without).

Not sure I understand the background here. Just wanted to make sure to point out, that these two new properties are already visible (without any consent or ToU acceptance) on the public profile page of any user. So this is just a new method of accessing this data. It's not new data.

simonpoole:
> Just wanted to make sure to point out, that these two new properties are already visible (without any consent or ToU acceptance) on the public profile page of any user.

We are all aware of that, the problem is that displaying that information on the profile page was never a fsvo "legal" change to start with (in more than one way).

tomhughes:
It's information that people voluntarily choose to disclose - if they don't want to disclose then just don't add it!

If they don't want to disclose it then what on earth are they adding it for?

At some point there needs to be some actual sanity in the debate...

simonpoole:
> It's information that people voluntarily choose to disclose - if they don't want to disclose then just don't add it!

That's not the problem, it needs to be a) disclosed to the users in the privacy policy what the OSMF does with the information, aka stored in the database etc. and b) disclosed that it is displayed publicly (including via the public API), aka not just to other logged in OSM users (which is likely why most people added it in the first place, except for companies doing SEO). 

Or put differently, the OSMF needs to provide the users with enough information that they can make an informed decision on if it makes sense for them to add the information.

PS: yes this is different from the heatmap issue in that it is at least nominally opt-in, which the heatmap isn't.

PR #6455, opened 2025-10-19
Bump minitest

rkoeze:
### Description

This PR removes the version constraint for the minitest gem. A version constraint [was first added back in 2013](https://github.com/openstreetmap/openstreetmap-website/commit/61dd2af0a0d952c9aabea92426e961cfc09b838b), but I don't see a reason to keep it around at this point. If the gem receives a breaking update, we'll find out when the tests stop passing and can decide to ship a fix or pin a version as needed.

### How has this been tested?
The tests pass.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -159,7 +159,7 @@ group :test do
   gem "erb_lint", :require => false
   gem "factory_bot_rails"
   gem "jwt"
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #6456, opened 2025-10-20
Remove unused initializer

rkoeze:
### Description
The `application_controller_renderer.rb` file is no longer included in newly generated Rails apps. The [rationale](https://github.com/rails/rails/pull/42538):

> I don't think this is something most Rails apps will need (certainly not on day 1); users can configure this feature when they need it.

That reasoning applies here too. Let's remove something we don't need. And if we ever do, we can add it back without issue. This change also keeps us aligned with Rails defaults.

### How has this been tested?

I was able to build and navigate the app locally. That and the tests pass.


gravitystorm:
Merged, thanks!

PR #6458, opened 2025-10-20
DRY docs

pablobm:
As part of #6424, new documentation was added that duplicated the existing one, which in turn had substantial duplication already.

I propose that we DRY up the existing docs a little bit, splitting it as follows:

```
CONTRIBUTING.md
README.md
SECURITY.md
doc
├── CONFIGURE.md
├── DEVCONTAINER.md
├── DOCKER.md
├── FAQ.md
├── INSTALL.md
└── MANUAL_INSTALL.md
```

Details:
- I'm leaving in the root dir the files that I think are standard to find there, at least in the sense that GitHub picks up on them.
- From `CONTRIBUTING.md` we link to `INSTALL.md`, which in turn gives options to the reader to continue with either `DOCKER.md`, `DEVCONTAINER.md`, or `MANUAL_INSTALL.md`. We could also separate the `rbenv` stuff into its own file.
- Regardless of the chosen option, the user is sent to `CONFIGURE.md` after the install. All this stuff is common to all methods.


waldyrious:
I normally interpret "install" as something an end-user would do to have a software ready to use in their system. These `INSTALL*.md` files feel more about something I'd call `DEV_SETUP*.md` or similar. But I didn't read the contents exhaustively, so I might be missing something.

pnorman:
> I normally interpret "install" as something an end-user would do to have a software ready to use in their system.

In the context of this codebase an end-user is someone visiting the site. They don't have to install anything and should never need to look at the files in git.

INSTALL.md is targeted at people using the code, rather than end-users who interact with the result of the code (the website).

Most of the users of the code will be developers running it locally or non-OSM sites running a variant of the code in production.

pablobm:
Re: `INSTALL.md`, I think in this case it doesn't matter too much, as it's out of sight in the `docs/` folder, as opposed to in the root dir (where it is in `master` at the moment).

Note that (at least in my understanding) the `docs/` folder is still special. GitHub takes it into account when generating autolinks that appear in the interface (as it happens with the license, etc). However I believe that `INSTALL.md` is not one such "conventional" file, so GitHub won't do anything special about it.

Having said that, perhaps there's a better, more conventional name that can be used. I'm not sure where to find the full list. The closest I've found is https://github.com/danpoynor/special-github-files, but it's unofficial.

pnorman:
> Having said that, perhaps there's a better, more conventional name that can be used

INSTALL.md comes from the long-standing [GNU standards](https://www.gnu.org/prep/standards/standards.html#Releases) and is what is commonly used. As far as I know it has no special significance to Github.

doc/DOCKER_INSTALL.md
```diff
@@ -101,6 +101,12 @@ Run the test suite:
 docker compose run --rm web bundle exec rails test:all
 ```
 
+If you encounter errors about missing assets, precompile the assets:
+
+```bash
+docker compose run --rm web bundle exec rake assets:precompile
```
hlfan: That section has duplicated itself.

pablobm: I just noticed 🙂

doc/INSTALL.md
```diff
@@ -0,0 +1,41 @@
...
+### Containerized Installation
+
+We offer containerized environments with Docker which may avoid installation difficulties:
+
+- To use Docker manually, see [DOCKER](DOCKER.md).
+- To use Docker via [Development Containers](https://containers.dev) (devcontainers), see [DEVCONTAINER.md](DEVCONTAINER.md).
+
+### Manual Installation
+
+Install dependencies directly on your machine (traditional approach, covered in this guide). This gives you the most control and is often preferred by experienced developers on Linux systems.
```
hlfan: The "covered in this guide" leftover is somewhat misleading

doc/MANUAL_INSTALL.md
```diff
@@ -276,28 +254,10 @@ After installing a version of Ruby with `rbenv` (the latest stable version is a
 rbenv local $VERSION
 ```
 
-Where `$VERSION` is the version you installed. You can see a list of available versions by running `rbenv versions`. Then install bundler:
```
hlfan, line 279: I don't think you changed your mind on that line, did you?

pablobm: I'm rebasing my original branch from November and wishing I had just started from scratch :sweat:

doc/INSTALL.md
```diff
@@ -0,0 +1,41 @@
...
+
+- To use Docker manually, see [DOCKER](DOCKER.md).
+- To use Docker via [Development Containers](https://containers.dev) (devcontainers), see [DEVCONTAINER.md](DEVCONTAINER.md).
+
+### Manual Installation
+
+Install dependencies directly on your machine. This gives you the most control and is often preferred by experienced developers on Linux systems.
+
+> [!WARNING]
+> **Windows Note:** We don't recommend using this approach on Windows, as some Ruby gems may not be supported. If you are using Windows, we recommend containerized setup using [Docker](DOCKER.md).
```
hlfan:
```suggestion
> **Windows Note:** We don't recommend using this approach on Windows, as some Ruby gems may not be supported. If you are using Windows, we recommend a containerized setup as mentioned above.
```

doc/DOCKER_INSTALL.md
hlfan: Is this by any chance now an older version of the docker guide?

pablobm: I'm currently starting from scratch. I'll leave a comment when done :wink:

pablobm:
I have now rebased (actually redone from scratch) this PR, after the merge of https://github.com/openstreetmap/openstreetmap-website/pull/6424. It's ready to be reviewed.

doc/DEVCONTAINER.md
```diff
@@ -84,39 +84,48 @@ If you want to add in the full history later on, perhaps to run `git blame` or `
...
 Hopefully it should be all green? 🤞 You can also start a development server with `bundle exec rails s`:
 
-![Running the dev server in the terminal](./docs/assets/vscode-rails-server.png)
+![Running the dev server in the terminal](./assets/vscode-rails-server.png)
 
 It will take a moment to start, after which you will be able to take a browser to http://localhost:3000 and visit your own local version of the site.
 
-## Other useful configuration
+> [!NOTE]
+> The OSM map tiles you see aren't created from your local database - they are the production map tiles, served from a separate service over the Internet.
```
hlfan, line 118: As said in presentations like https://youtu.be/XtCMumUX89o?t=210, the map tiles are just one of a few features that are coming over the web.

pablobm: For now I just wanted this to be consistent with the other documentations.

doc/DOCKER.md
```diff
@@ -79,105 +79,65 @@ This will launch one Docker container for each 'service' specified in `docker-co
...
-Similarly, if you want to `exec` in the db container use:
+## What's next?
 
-```bash
-docker compose exec db bash
-```
+🎉 **Congratulations!** You have successfully installed the OpenStreetMap website.
+
+**Next steps:**
+* **Configuration:** See [CONFIGURE.md](CONFIGURE.md) for populating the database with data, creating users, setting up OAuth, and other configuration tasks.
```
tomhughes, line 140: The link here is broken because `CONFIGURE.md` is still in the top level - the same applies to the `What's next?` section in the other files as well.

tomhughes (changes requested):

doc/DOCKER.md
pablobm: I think you mean `CONTRIBUTING.md`? I have fixed all instances now, thank you!

pablobm:
Apologies for the merge commit. After some head scratching wondering where that had come from, I realised that I had clicked a button that GitHub makes all too easy to click by mistake. It's fixed now.

doc/DOCKER.md
tomhughes: Sorry yes that was what I meant ;-)

tomhughes (approved):

PR #6461, opened 2025-10-21
Remove unnecessary requires

pablobm:
This PR removes a number of `require` calls that are unnecessary or discouraged.

In general, it's discouraged to call `require` from within `app/` as it can interfere with the autoloader.

These are the individual rationales for each `require`:
- `timeout`: already required by ActiveRecord, but added to the gemfile to be explicit.
- `rexml/document`: required by default by `rexml`, added to the gemfile to be explicit.
- `ostruct`: not being used any more.
- `uri`: loaded as a gem.
- `xml/libxml`: required automatically by `libxml-ruby`.
- `rtlcss`: loaded as a gem.
- `open3`: loaded as a gem.
- `digest`: already required by... Bundler I think, but added to the gemfile to be explicit
- `addressable/uri`: required by default by `addressable`

Note that some gems like `timeout` are "standard gems" which are present in Ruby by default but are also offered as separate gems (see https://stdgems.org/). The point here is that I think it's better to add them to the gemfile to be explicit, which also avoids these discouraged `require`s.

Additionally, there's no need to add `:require => "libxml"` to the gem `libxml-ruby` as that happens automatically too.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -93,7 +93,7 @@ gem "kramdown"
 gem "aasm"
 
 # XML parsing and generation
-gem "libxml-ruby", ">= 2.0.5", :require => "libxml"
+gem "libxml-ruby", ">= 2.0.5"
 gem "rexml"
 
 # Use for HTML sanitisation


tomhughes:
Looks good to me, thanks.

PR #6464, opened 2025-10-22
Docker: clear out apt repo after installing packages

pablobm:
At https://github.com/openstreetmap/openstreetmap-website/pull/5157, this `apt-get clean` was removed. The man page describes this command as follows:

> clean clears out the local repository of retrieved package files. It removes everything but the lock file from /var/cache/apt/archives/ and /var/cache/apt/archives/partial/.

Normally I wouldn't have thought much about it, but what tickled me was that the comment at the top still says "then clean up to minimize image size". So this left me wondering whether I should update the comment or re-introduce the command. I couldn't see in the PR conversation anything about why it was removed, so I thought I'd start by proposing the re-introduction.

I didn't go as far as adding the `rm -rf /var/lib/apt/lists/*` line as I don't feel comfortable messing up with APT internals.

Thoughts? Perhaps @kcne remembers?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,30 +6,32 @@ ENV DEBIAN_FRONTEND=noninteractive
 # Install system packages then clean up to minimize image size
 RUN apt-get update \
   && apt-get install --no-install-recommends -y \
-  build-essential \
-  curl \
-  default-jre-headless \
-  file \
-  git-core \
-  gpg-agent \
-  libarchive-dev \
-  libffi-dev \
-  libgd-dev \
-  libpq-dev \
-  libsasl2-dev \
-  libvips-dev \
-  libxml2-dev \
-  libxslt1-dev \
-  libyaml-dev \
-  locales \
-  postgresql-client \
-  tzdata \
-  unzip \
-  nodejs \
-  npm \
-  osmosis \
-  ca-certificates \
-  firefox-esr
+    build-essential \
+    curl \
+    default-jre-headless \
+    file \
+    git-core \
+    gpg-agent \
+    libarchive-dev \
+    libffi-dev \
+    libgd-dev \
+    libpq-dev \
+    libsasl2-dev \
+    libvips-dev \
+    libxml2-dev \
+    libxslt1-dev \
+    libyaml-dev \
+    locales \
+    postgresql-client \
+    tzdata \
+    unzip \
+    nodejs \
+    npm \
+    osmosis \
+    ca-certificates \
+    firefox-esr \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/* # Replace with `apt-get dist-clean` after upgrading to Debian 13 (Trixie)
 
 # Install yarn globally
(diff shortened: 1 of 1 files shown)


gravitystorm:
Good catch, this should be re-added. 

I’m also happy to see the apt/lists removal re-added - it’s in every guide I’ve seen, and although I don’t know exactly what it does (or more specifically, whether their removal actually saves much space when layering is considered) it’s low risk since `apt` isn’t subsequently used.

Then again, if there’s even a small chance that a dev might want to apt-install some tool into their container (e.g. for debugging) perhaps we should leave the apt in a fully working state?

I’ll leave this in case anyone has further comment on the lists topic, otherwise I plan to merge this as-is.

pablobm:
>  if there’s even a small chance that a dev might want to apt-install some tool into their container (e.g. for debugging) perhaps we should leave the apt in a fully working state?

That's my thought. I may want to install packages for some punctual need, so I'd rather that we leave APT in a working state.

However, I just had a look at  `/var/lib/apt/lists/` and realised that it takes 164M in my Debian machine. Let me have a look at how broken things can be if removed. If APT recovers gracefully, I may add the rimraf line.

pablobm:
Weirdly, when I tried there appeared to be less space available when deleting the lists than when not. I'm guessing it was a fluke as a result of... something container image volume something. No idea.

Before removal:
```
# du -sh /var/lib/apt/lists/
19M	/var/lib/apt/lists/

# df -h /
Filesystem      Size  Used Avail Use% Mounted on
overlay         220G  121G   89G  58% /
```
After removal:
```
# df -h /
Filesystem      Size  Used Avail Use% Mounted on
overlay         220G  123G   87G  59% /
```

But APT works. It just needs running `apt update` and then you can install packages again. So I added the line.

pnorman:
the apt-get update, install, clean, and rm -rf all need to be in the same command. It'll save the full size of the temporary files because they aren't present on any other layers. It doesn't break anything since it's just restoring parts of the state to how they were before temporary files were downloaded with apt-get update.

`apt-get distclean` is probably the right command to use instead of the rm -rf. I'm not sure if distclean implies clean

pablobm:
> all need to be in the same command

What do you mean by this, @pnorman? They are currently all in the same command, in the sense that they are joined by `&&`. Am I misunderstanding?

As for `apt-get distclean`, it looks like it was introduced in Debian 13 (Trixie), so it's not yet available here as this runs Debian 12 (Bookworm). We'll have to leave it to when that happens. I have left a comment to remind anyone who looks at the file in the future.

tomhughes:
Looks good to me, thanks.

pnorman:
> What do you mean by this, @pnorman? They are currently all in the same command, in the sense that they are joined by `&&`. Am I misunderstanding?

I recall seeing something else on my phone, but the phone isn't great for reviewing diffs so I probably misread it.

pablobm:
@pnorman - No worries :slightly_smiling_face:

PR #6466, opened 2025-10-22
Order tracepoints before asking for the first one

tomhughes:
This fixes a new warning in rails 8.1.0 about calling first on a query that does not have a defined order.

We've been getting away with because the trace only has one point.


gravitystorm:
Merged, thanks!

PR #6467, opened 2025-10-22
Update to rails 8.1.0

tomhughes:
First draft of an update to rails 8.1.0 which does actually work but has a number of new deprecation warnings:

* One from our code, fixed by #6466
* One from `omniauth-rails_csrf_protection` already reported upstream in https://github.com/cookpad/omniauth-rails_csrf_protection/issues/23
* One (repeated many times in the tests) from `bootstrap_form` where upstream doesn't look very active :-(

There's also a pending dependency fix for `canonical-rails` which is open as https://github.com/jumph4x/canonical-rails/pull/96 upstream which I've pulled it as a github reference for now.


pablobm:
Looks like `bootstrap_form` is fixed in main, but there has been no release since 2023. I have opened an issue at https://github.com/bootstrap-ruby/bootstrap_form/issues/765, but as you say there hasn't been any activity in the last 4 months.

So a workaround would be to use the main version. I can confirm it removes the warning:

```
gem "bootstrap_form", :github => "bootstrap-ruby/bootstrap_form", :ref => "main"
```

pablobm:
There's new activity over at `bootstrap_form`, and the maintainer has self-assigned the issue I opened. This may solve itself :crossed_fingers:

tomhughes:
The `bootstrap_form` warning has been resolved by the latest upstream release.

gravitystorm:
Looks good to me. I'm happy enough with the single deprecation warning for `omniauth-rails_csrf_protection`. We could wait for it to be fixed first, but we might be waiting a while. Same thought process for `canonical_rails`, at least there's no warning from that.

This is currently a draft PR, so whenever you are happy to mark it as ready, I'm happy to merge.

tomhughes:
Sounds good to me... I have rebased and marked this ready.

gravitystorm:
Merged, thanks!

Issue #6419, opened 2025-09-26
Handle whitespace in social links username display


uw-amy-bordenave:
### URL

https://www.openstreetmap.org/user/UW%20Amy%20Bordenave

### How to reproduce the issue?

Display username is truncated at whitespace in HDYC link.
Link functionality itself is not impacted and works as expected. 

### Screenshot(s) or anything else?

<img width="231" height="230" alt="Image" src="https://github.com/user-attachments/assets/2f45e41a-4adb-4d6a-a27c-2e8bf6194542" />

pablobm:
Thank you for the report! 🙂 

The issue seems to be with the regular expressions used to parse the social links. The one for HDYC should accept spaces, but it doesn't. See [app/models/social_link.rb](https://github.com/openstreetmap/openstreetmap-website/blob/b5800a57d59df9b41a61eab9da3f91faab71a032/app/models/social_link.rb#L34).

This could be a great first issue for new contributors to work on. Anyone up for it? Ideally a PR should also include a unit test, see [test/models/social_link_test.rb](https://github.com/openstreetmap/openstreetmap-website/blob/b5800a57d59df9b41a61eab9da3f91faab71a032/test/models/social_link_test.rb).

Yatharths11:
Hey @pablobm Is it possible that the issue is not with the Regex parsing social links but with text truncation class provided in
[views/social_links/_show.html.rb](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/views/social_links/_show.html.erb)

pablobm:
I think not. My understanding is that the CSS class would truncate with an ellipsis: https://getbootstrap.com/docs/5.3/helpers/text-truncation/ instead of disappearing the text without a trace. Also, the regexp is missing a space.

Openstreetmapler:
This bug also occurs with the link to my Wikidata user page, which is [Keepright! ler](https://www.wikidata.org/wiki/User:Keepright!_ler), but shows up as [Keepright](https://www.wikidata.org/wiki/User:Keepright!_ler) on my [OSM website profile](https://www.openstreetmap.org/user/keepright!%20ler) – note that the **exclamation point is cut off** as well as the space.

pablobm:
[...]

PR #6468, opened 2025-10-23
Modified hdyc regex to accept spaces and percent-encoded spaces

Yatharths11:
Modified hdyc regex to accept spaces and percent-encoded spaces
For Issue #6419

### DIFF
--- a/app/models/social_link.rb
+++ b/app/models/social_link.rb
@@ -31,7 +31,7 @@ class SocialLink < ApplicationRecord
     :flickr => %r{\Ahttps?://(?:www\.)?flickr\.com/people/([a-zA-Z0-9@._-]+)},
     :github => %r{\Ahttps?://(?:www\.)?github\.com/([a-zA-Z0-9_-]+)},
     :gitlab => %r{\Ahttps?://(?:www\.)?gitlab\.com/([a-zA-Z0-9_-]+)},
-    :hdyc => %r{\Ahttps?://(?:www\.)?hdyc\.neis-one\.org/\?([a-zA-Z0-9_-]+)},
+    :hdyc => %r{\Ahttps?://(?:www\.)?hdyc\.neis-one\.org/\?([a-zA-Z0-9\s%20_-]+)},
     :hot => %r{\Ahttps?://tasks\.hotosm\.org/users/([a-zA-Z0-9_-]+)},
     :instagram => %r{\Ahttps?://(?:www\.)?instagram\.com/([a-zA-Z0-9._]+)},
     :linkedin => %r{\Ahttps?://(?:www\.)?linkedin\.com/in/([a-zA-Z0-9_-]+)},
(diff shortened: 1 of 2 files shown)


app/models/social_link.rb
```diff
@@ -31,7 +31,7 @@ class SocialLink < ApplicationRecord
     :flickr => %r{\Ahttps?://(?:www\.)?flickr\.com/people/([a-zA-Z0-9@._-]+)},
     :github => %r{\Ahttps?://(?:www\.)?github\.com/([a-zA-Z0-9_-]+)},
     :gitlab => %r{\Ahttps?://(?:www\.)?gitlab\.com/([a-zA-Z0-9_-]+)},
-    :hdyc => %r{\Ahttps?://(?:www\.)?hdyc\.neis-one\.org/\?([a-zA-Z0-9_-]+)},
+    :hdyc => %r{\Ahttps?://(?:www\.)?hdyc\.neis-one\.org/\?([a-zA-Z0-9\s%20_-]+)},
```
tomhughes, line 34: You don't need to add `20` there as that's a character class so it's matching each character individually and digits are already allowed.

I'm not really keen to handling escaping like this anyway - it would probably be better to canonicalise the URLs before matching them so we don't have to handle both escaped and unescaped versions in the regular expressions.

tomhughes: Also do we want `\s` or just ` ` allowed? Are other forms of whitespace like tab really valid?

Yatharths11: I observed for #6419 the view uses the text truncate class, can that be the reason for the issue ?
If it is, there will be no reason to modify the regex here and just fix the view.

pablobm: Good question on what exact characters are allowed. HDYC references our display names, so anything valid on our website will be valid for HDYC. This seems to be the relevant declaration in our user validations:

https://github.com/openstreetmap/openstreetmap-website/blob/8738786e21b8c925e2d33078a6d76e5c9333db1c/app/models/user.rb#L106-L109

Which in turn uses a custom validator that excludes these characters:

https://github.com/openstreetmap/openstreetmap-website/blob/8738786e21b8c925e2d33078a6d76e5c9333db1c/app/validators/characters_validator.rb#L4-L5

So perhaps those constants should be used in the regexp? Similar to what the validator does when it goes `/[#{INVALID_CHARS}]/o.match?(value)`.

pablobm: Also, generally on regular expressions: "character classes" (lists of characters surrounded by square brackets `[...]`) take each character individually, so `%20` is `%`, `2`, or `0` individually, but not the three together. The `\s` does work because `\` is a escape character which is a special case. `%` is a escape character in URLs, but not in regular expressions.

This is a great website to test out regular expressions: https://rubular.com/

Yatharths11: In that case, the changes made are valid right?

tomhughes: Well they don't actually break anything but they also don't really match % escape sequence and it definitely has redundant duplicate characters in the character class.

pablobm: Also note that the tests are failing, as `%20` is not translated to ` ` as they expect.

pnorman: > In that case, the changes made are valid right?

It would allow stuff like `%22`, `%_20`, and other invalid

PR #6470, opened 2025-10-23
Increase tag2link item leniency

hlfan:
The tag2link initialization now handles value conflicts between the two sources more effectively and gives the claims from the OSM wiki a slight preference.

This improved URL selection logic for ambiguous cases results in more keys receiving formatters:

For v2025.9.21, this adds these:
- `ref:INEP`
- `wikimedia_commons`
- `woeid`

For v2025.10.21, this adds these:
- `mapillary`
- `ref:INEP`
- `ref:US:EIA`
- `wikimedia_commons`
- `woeid`

This unblocks #6463.

### DIFF
--- a/lib/tag2link.rb
+++ b/lib/tag2link.rb
@@ -19,13 +19,36 @@ module Tag2link
     data
       # exclude deprecated and third-party URLs
       .reject { |item| item["rank"] == "deprecated" || item["source"] == "wikidata:P3303" }
-      .group_by { |item| item["key"] }
-      .transform_keys { |key| key.sub(/^Key:/, "") }
-      # move preferred to the start of the array
-      .transform_values { |items| items.sort_by { |item| item["rank"] == "preferred" ? 0 : 1 }.uniq { |item| item["url"] } }
-      # exclude any that are ambiguous, i.e. the best and second-best have the same rank
-      .reject { |_key, value| value[1] && value[0]["rank"] == value[1]["rank"] }
-      # keep only the best match
-      .transform_values { |items| items[0]["url"] }
+      .group_by { |item| item["key"].sub(/^Key:/, "") }
+      .transform_values { |items| choose_best_item(items) }
+      .compact
+      .transform_values { |items| items["url"] }
   end
+
+  def self.choose_best_item(items)
+    return nil if items.blank?
+
+    return items.first if items.size == 1
+
+    # move preferred to the start of the array
+    ranked = items.sort_by { |item| item["rank"] == "preferred" ? 0 : 1 }.uniq { |item| item["url"] }
+    top_rank = ranked.first["rank"]
+    top_items = ranked.select { |i| i["rank"] == top_rank }
+
+    # if only one top-ranked item, prefer that
+    return top_items.first if top_items.size == 1
+
+    grouped = top_items.group_by { |i| i["source"] }
+    return nil if grouped.size > 2
+
+    # if both sources have exactly one preferred, prefer osmwiki
+    return grouped["osmwiki:P8"]&.first || grouped.values.flatten.first if grouped.all? { |_s, vals| vals.size == 1 }
+
+    # if one source has multiple preferreds and the other has one, prefer the single one
+    return grouped.min_by { |_s, vals| vals.size }.last.first if grouped.any? { |_s, vals| vals.size == 1 }
+
+    # exclude any that are ambiguous
+    nil
+  end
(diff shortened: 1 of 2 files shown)


tomhughes:
What are the actual changes though?

hlfan:
Don't reject a key, if:
- one source (Wikidata item or OSM Wiki data item) has a single best URL.
- both sources have only one URL, then take the one from the OSM Wiki, since it usually more accurately represents the identifier.

hlfan:
e. g. wikimedia_commons=* is linked by multiple wikidata properties, which don't exactly represent the key's scope.

pablobm:
I think we should have some tests about this. It's all very obscure at the moment. With tests in place, we can also refactor this logic with more confidence, as it's not pretty at the moment.

It doesn't need to be comprehensive from the start, but at least stub out a test module and test some cases, like the ones introduced here.

hlfan:
No initializers have dedicated tests at the moment. I was partially working off the test that failed in #6463.

pablobm:
Right. Yes, in this case I think the code should be moved to a module, instead of depending on a global constant.

config/initializers/tag2link.rb
```diff
@@ -1,17 +1,5 @@
...
-               .reject { |item| item["rank"] == "deprecated" || item["source"] == "wikidata:P3303" }
-               .group_by { |item| item["key"] }
-               .transform_keys { |key| key.sub(/^Key:/, "") }
-               # move preferred to the start of the array
-               .transform_values { |items| items.sort_by { |item| item["rank"] == "preferred" ? 0 : 1 }.uniq { |item| item["url"] } }
-               # exclude any that are ambiguous, i.e. the best and second-best have the same rank
-               .reject { |_key, value| value[1] && value[0]["rank"] == value[1]["rank"] }
-               # keep only the best match
-               .transform_values { |items| items[0]["url"] }
+require Rails.root.join("lib/tag2link")
```
pablobm: No need for the absolute path. `lib/` is automatically in the library load path:

```suggestion
require "tag2link"
```

lib/tag2link.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module Tag2link
+  class << self
```
tomhughes: Is there a reason for using this relatively obscure syntax rather than just using `def self.` to define class methods? I find it easier to understand and it's how we generally do it in other places?

test/lib/tag2link_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    assert_equal "https://www.wikidata.org/entity/Q936", url
+  end
+
+  def test_build_dict_rejects_deprecated_and_third_party
+    data = [
+      { "key" => "Key:example", "url" => "http://example.com/$1", "rank" => "deprecated", "source" => "osmwiki:P8" },
+      { "key" => "Key:example2", "url" => "http://example2.com/$1", "rank" => "preferred", "source" => "wikidata:P3303" },
+      { "key" => "Key:example3", "url" => "http://example3.com/$1", "rank" => "preferred", "source" => "osmwiki:P8" }
+    ]
+    dict = Tag2link.send(:build_dict, data)
```
tomhughes: Why the need to use `send` here? It's not a private method?

lib/tag2link.rb
hlfan: Oh yeah that's a remnant of a previous structure. Taking a look.

hlfan: RuboCop complained that
> `private` (on line 28) does not make singleton methods private. Use `private_class_method` or `private` inside a `class << self` block instead.

When I tried `private_class_method` the first time, it didn't work out for me, but it seems to work now.

pablobm: Yeah, long story but `private` doesn't apply to singleton methods, which is what those `def self.method_name` do and allows us to get class methods. The `class << self` thing is common in this type of function-wrapper modules, but I agree it may not be appropriate in a project like this one, where we want to encourage contributions from non-Rubyists.

A slight improvement would be to do `private_class_method def self.choose_best_item(items)`, all in one line, which avoids splitting this declaration into two lines.

tomhughes: The other option would be to use `class << self` just for the private methods as things like `lib/bounding_box.rb` do.

hlfan: Or just leave `choose_best_item` public. `build_dict` also doesn't need to be public but I left it for testing.

PR #6476, opened 2025-10-25
Fix naming in abstract adapter initializer

rkoeze:
### Description
Fix module name referenced in AbstractAdapter.  Changes `ConnectionAdaptors` to `ConnectionAdapters`. 

### How has this been tested?
TBD

### DIFF
--- a/config/initializers/abstract_adapter.rb
+++ b/config/initializers/abstract_adapter.rb
@@ -4,7 +4,7 @@ if defined?(ActiveRecord::ConnectionAdapters::AbstractAdapter)
   module OpenStreetMap
     module AbstractAdapter
       module PropagateTimeouts
-        def translate_exception_class(e, sql)
+        def translate_exception_class(e, sql, binds)
           if e.is_a?(Timeout::Error) || e.is_a?(OSM::APITimeoutError)
             e
           else


pablobm:
Ooopsie, so I gather that this never worked? :cold_sweat: 

Questions:
- What led to including this monkeypatch?
- Is this necessary any more? The upstream implementation looks ok to me (start at https://github.com/rails/rails/blob/dcb29df5b25006dd92f4fc10295f0177a33db93c/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L1193).
- If necessary, is the big conditional still required?

tomhughes:
Yes apparently it never worked and it may well be at least part of a problem that is on my list to look at where timeouts don't seem to be working properly.

rkoeze:
Sorry for the late reply here. I think the idea behind the monkey patching is that it allows timeout exceptions to be propagated without wrapping the exception (ActiveRecord's default behavior) so we [can handle the timeout case explicitly](https://github.com/openstreetmap/openstreetmap-website/blob/1515df8ce98e2de3e1cfdbd20681ecbe86cd1417/app/controllers/application_controller.rb#L228). 

With that said, `Timeout::Error` is a runtime error, so it looks like [ActiveRecord should let that bubble up](https://github.com/rails/rails/blob/dcb29df5b25006dd92f4fc10295f0177a33db93c/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L1226) anyway. 

Generally using Timeout to wrap critical operations [is not recommended](https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/?utm_source=chatgpt.com). With that in mind, how do we feel about the following?
1. We remove this initializer. It's never worked. And fixing it would lean into a pattern that is not recommended.
2. We start exploring an alternative to using Timeout for these longer running queries (a DB timeout perhaps).

I'm sure I'm missing some historical context here so @tomhughes please feel free to correct me on what I'm missing.

tomhughes:
The historical context is that we need to ensure that people can't kill the site by finding some way to trigger a very slow database query.

This is an ongoing problem right now with trace point queries which is what I was hinting at above because at least sometimes the timeouts are not working a the moment which is causing operational problems - the whole thing has long been tricky to get right because rails has a habit of swallowing the exceptions or transforming them into different sorts of exceptions hence things like this and some of the logic in the timeout methods in the controllers.

Unfortunately it's very hard to debug and I suspect it depends on exactly when the timeout fires as I haven't so far been able to reproduce the problem locally I just know that we sometimes have database queries that have been running for 30 minutes despite the timeout.

A database timeout doesn't quite do the same thing of course as this timeout in principle also protects against the rails code itself taking a long time for some reason.

pablobm:
Do we know which non-DB code causes exceptions? Do I suspect correctly that our absence of error reporting is making this tricky to figure out?

Normally I would rely on DB timeouts for DB-related code while manually tackling other sources of timeouts. For example with manual deadline checks inside long-running loops. Is that something that is feasible here?

tomhughes:
I'm not sure any absence of error reporting is an issue because the whole point is that errors (if you mean timeout exceptions) aren't triggering or are being swallowed somewhere so the request continues being processed.

rkoeze:
@pablobm My instinct would be to do what you're describing: let timeouts be configuration details when possible and then fill in any gaps with more targeted/manual solutions where needed.

I think the issue with Timeout is that it can inject an exception into a thread at essentially any point. That kind of behavior is hard to reason about and makes it very difficult to ship a single fix. We’ll always be playing whackamole as exceptions surface in different parts of the stack.

If there isn’t already, I’ll create an issue to centralize this discussion.

tomhughes:
I think at this point from my point of view probably the best thing to do as a matter of some urgency is to set a database timeout to match the overall timeout - after all if the whole controller action is meant to be limited to 30s then no database can ever take longer than that.

What I'm not sure of is how easy it is to set the database timeout dynamically - you can set it in the database configuration but that's a one shot thing.

mmd-osm:
I haven’t tested this code, but it looks like it could change the timeout for a single statement: https://stackoverflow.com/a/50325976

tomhughes:
Yes you can do a bit of low level poking like that - just wondering if there were better options.

I've set a global 5 minute timeout in https://github.com/openstreetmap/chef/commit/2943e2cb7e273b139f796fee86323b4882461340 for now which should hopefully help deal with the worst issues.

rkoeze:
It might also make sense to set a value for `lock_timeout`. That way long lock waits fail fast and show up as `ActiveRecord::LockWaitTimeout` instead of hanging until the statement timeout.

rkoeze:
Added an issue [here](https://github.com/openstreetmap/openstreetmap-website/issues/6572).

pablobm:
Regarding the content of this PR: should `config/initializers/abstract_adapter.rb` be removed entirely? If it never worked, its existence only adds confusion to the codebase.

rkoeze:
@pablobm planning to do that in a different PR

PR #6478, opened 2025-10-27
Add explicit requirement for Node.js 22.2+

pablobm:
As noted by @tomhughes at https://github.com/openstreetmap/openstreetmap-website/pull/6462#issuecomment-3435496014, this is the minimum required by osm-community-index, which is one of our dependencies. The change was introduced at https://github.com/osmlab/osm-community-index/commit/c3738945eeda5c43d6130d0139dc21d36352d39f

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,6 +27,7 @@ Many of the dependencies are managed through the standard Ruby on Rails mechanis
 
 **Minimum requirements:**
 * Ruby 3.2+
+* Node.js 22.2+
 * PostgreSQL 13+
 * Bundler (see note below about [developer Ruby setup](#ruby-version-manager-optional))
 * JavaScript Runtime
--- a/package.json
+++ b/package.json
@@ -21,5 +21,8 @@
     "eslint": "^9.0.0",
     "eslint-formatter-compact": "^9.0.1",
     "eslint-plugin-erb": "^2.1.0"
+  },
+  "engines": {
+    "node": ">=22.2"
   }
 }


gravitystorm:
It's a shame that we have to follow the minimum node version from a package that we only need static `.json` files from! 

It would be great to have as wide a supported version as possible, so that  our developers can use OS distribution packages wherever possible.

Given that [we use bookworm](https://github.com/openstreetmap/openstreetmap-website/blob/5fdeb3c68a906094f5ad5e0e722f5b81ad871d02/Dockerfile#L2) in the Dockerfile, and [bookworm ships with node 18](https://packages.debian.org/bookworm/nodejs), this change would be expected to fail CI.

However, since [we've patched bin/yarn to use `--ignore-engines`](https://github.com/openstreetmap/openstreetmap-website/blob/master/bin/yarn#L13) it still works. This then leads to a confusing situation where we will both specify a minimum node version (22) and also go out of our way to ignore it, and I'm not sure that's ideal.

Any thoughts? Is there any trickery that can allow us to set a minimum node version (so that dependabot doesn't unexpectedly introduce a new stricter minimum) while also allow per-package overrides (e.g. for packages where we don't execute any javascript code)?

pablobm:
I have put the question at https://github.com/osmlab/osm-community-index/commit/c3738945eeda5c43d6130d0139dc21d36352d39f#r168978734.

I was also looking into the possibility of vendorising the package, but I see that the files we want are built and not present in the repo... that makes things hairier.

A left-field option (bear with me for this one) would be to have two separate `package.json` files. One for "proper" dependencies, and another in `vendor/` or similar for things like this, which we run with `--ignore-engines`. Sounds gnarly and I haven't looked into it.

tomhughes:
Interestingly our production deployments use `bundle exec rails yarn:install` which apparently doesn't go through `bin/yarn` and hence doesn't get `--ignore-engines` because that's how I discovered this when the dev server deployment failed due to the engine version.

I upgraded production to use node 22 as a workaround, having long since forgotten about the `--ignore-engines` thign which comes from a previous time when something similar happened!

bhousel:
> It's a shame that we have to follow the minimum node version from a package that we only need static `.json` files from!

Yes - I agree you shouldn't need to do that! I commented in osm-community-index:

The engines field is really for specifying what version is required to build the project, but downstream projects sometimes treat it as the version required to install the project. You won't need to build anything if you are just installing this as a dependency and pulling .json files from it.

Even node treats this field as "advisory" unless you set the engineStrict flag, which is off by default.
https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines

Anyway, since we are using Bun now, I think I'll just remove the Node version from here. I didn't realize it was causing issues for installing the package.

hlfan:
Could this PR be marked as draft in the meantime?

pablobm:
Done :+1:

bhousel:
I just published a [v6.0.0](https://github.com/osmlab/osm-community-index/releases/tag/v6.0.0) a few hours ago where I removed the `engines` property - let me know if this one installs without trouble!

hlfan:
This PR seems indeed to be superseded by #6482.

PR #6479, opened 2025-10-27
Update to iD v2.37

tyrasd:
Most visible change: support of a dark theme. This PR also includes the necessary glue code from #6323 (thanks @hlfan).

Full changelog: https://github.com/openstreetmap/iD/releases/tag/v2.37.0 + https://github.com/openstreetmap/iD/releases/tag/v2.37.1 + https://github.com/openstreetmap/iD/releases/tag/v2.37.2


hlfan:
If you rebase my commit onto the translatewiki one (5fdeb3c68a906094f5ad5e0e722f5b81ad871d02) to get rid of the merge commit, does it get signed and "verified"?

tyrasd:
> does it get signed and "verified"

yes :innocent:

app/views/site/id.html.erb
```diff
@@ -3,9 +3,7 @@
 <head>
   <meta charset='utf-8'>
   <%= stylesheet_link_tag "id" %>
-  <!--[if !IE || gte IE 9]><!-->
   <%= javascript_include_tag "id" %>
-  <!-- <![endif]-->
```
pablobm: Does removing this conditional make sense as part of this PR? Looks like something that should be removed, but perhaps in a separate PR.

hlfan: It's already a separate commit and IE<9 support is already basically nonexistent.

pablobm: Definitely remove the support, it's just that this PR is about an iD update which is different. If this was merged, something went wrong, and we had to revert this merge, then this unrelated change would be reverted too.

tyrasd: Ok, I extracted the obsolete IE code and will open a separate PR for it. -> #6489

tyrasd:
> How can I test this? Poking at it on local and I don't know what I should be looking at.

It only affects the view http://127.0.0.1:3000/id (and http://127.0.0.1:3000/edit if iD is configured as the user's editor). So, it's probably best to test that the embedded editor loads as expected, including assets like icons, background layers data and presets, and that changeset upload works by doing a quick test edit.

Maybe @tomhughes can give some hints about what else there is to look at.

tyrasd:
For this particular version, some extra reviewing effort should probably be at the taken at the new dark mode integration: does iD properly adjust to the preference in the user's profile, or the browser's setting if left on `auto`.

pablobm:
Thank you, I should have thought of that. I'm not a dark mode user and I forget how it works.

I can report it works well on Firefox 140.4.0esr + Debian 13 + GNOME 48 + Wayland, running on fairly old hardware. I tested by changing the dark mode setting at the Firefox level and the GNOME level. Also uploaded a simple changeset.

pablobm:
The CI failure looks like a flake, but it might be interesting to @tomhughes outside of this context as it's at `Api::Changesets::UploadsControllerTest#test_upload_race_condition` https://github.com/openstreetmap/openstreetmap-website/actions/runs/18937363715/job/54067445794?pr=6479

hlfan:
I feel like I've seen these "Request a page first." errors a bit more often in CI recently, but I wasn't able to reproduce these locally.

tyrasd:
PS: I've [updated](https://github.com/openstreetmap/openstreetmap-website/compare/4f0d52e4535f1fba9264bd5e3149b7d127e5aec7..9311074691552934c6d2b8101af1cf7b30c9a7b6) this branch to version [2.37.2](https://github.com/openstreetmap/iD/releases/tag/v2.37.2) of iD (which includes another small bugfix).

tomhughes:
All looks good to me, thanks.

andrewharvey:
There appears to be a critical issue related to this release https://github.com/openstreetmap/iD/issues/11532 with a number of people reporting the same.

PR #6484, opened 2025-10-29
Remove old google webmaster files

gravitystorm:
These proof-of-control files are probably not needed any more, since the newest one dates from 13 years ago. In future, they would be more likely controlled from the deployment-specific chef configuration, rather than included for all deployments in this codebase.


tomhughes:
Agreed if they're anywhere they should be in chef.

@Firefishy before we remove these do you know if any of them are needed? If so I can move them to chef...

Firefishy:
Feel free to remove. I will add as required via chef.

PR #6486, opened 2025-10-29
Check that commenters don't receive notifications about their own comments

pablobm:
I noticed that the following code has a `if current_user != user` that is not tested anywhere:

https://github.com/openstreetmap/openstreetmap-website/blob/8e97f3b1017975fb51104a22175e6d4cd8d23eec/app/controllers/api/changeset_comments_controller.rb#L47-L49

I went to add a test, ended up doing a bit of refactoring. Two commits here:
1. Refactor checks of whether emails are being received by the expected recipients, using a custom assertion. I also use this at `test/controllers/api/notes_controller_test.rb` which is unrelated to my initial intention.
2. Add an assertion that an unsubscribed commenter doesn't receive a notification email about their own comment, plus a new test checking the same thing for a subscribed commenter. Also testing an unrelated user, since we are at it.

### DIFF
--- a/test/controllers/api/changeset_comments_controller_test.rb
+++ b/test/controllers/api/changeset_comments_controller_test.rb
@@ -228,10 +228,11 @@ module Api
       creator_user = create(:user)
       changeset = create(:changeset, :closed, :user => creator_user)
       changeset.subscribers << creator_user
-      other_user = create(:user)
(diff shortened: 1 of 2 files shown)


tomhughes:
Looks good to me, thanks.

PR #6487, opened 2025-10-29
Make use of the linked Bootstrap icons font

hlfan:
So we don't need to save the icons in this repo, too.

### DIFF
--- a/app/assets/javascripts/index/contextmenu.js
+++ b/app/assets/javascripts/index/contextmenu.js
@@ -190,7 +190,7 @@ class ContextMenu {
   }
 
   _createMenuItem(item) {
-    const $icon = $("<i>").addClass(`bi ${item.icon}`);
+    const $icon = $("<i>").addClass(`bi ${item.icon}`).prop("ariaHidden", true);
     const $label = $("<span>").text(item.text);
 
     const $link = $("<a>")


test/system/profile_company_change_test.rb
```diff
@@ -43,7 +43,7 @@ class ProfileCompanyChangeTest < ApplicationSystemTestCase
     assert_text "Profile company updated."
 
     within_content_body do
-      assert_text :all, "Company Test Co."
```
pablobm, line 46: I was scratching my head about these changes for a minute. Now I see that the `title` argument was being injected into the SVG to serve as `<title>` tag within it.

Is there an accessibility impact? Do user agents read the `<title>` in the SVG? If so, are we losing an accessibility aid? Should these be replaced with an alt text? Not sure how to do that with Bootstrap icons, a quick search seems to suggest to use `.visually-hidden` within the `<i>`: https://getbootstrap.com/docs/5.3/getting-started/accessibility/#visually-hidden-content

pablobm (commented):
I'm generally in favour of this change, although I want to highlight that there will be subtle changes to some icons. In particular some appear to be a bit smaller. Just so that we all know what this exactly entails.

This is three of them. Note that I have also seen some that appear completely unchanged:

| Where | Current | Proposed |
| ---------- |  ------------- | ------------- |
| New Diary Entry | Image |  Image |
| Latest Diaries | Image | Image |
| Company icon | Image | Image |

That's as rendered by Firefox 140.4.0esr with Debian 13 on old hardware.

test/system/profile_company_change_test.rb
hlfan: From my quick test, the title HTML attribute gave me the same behavior as the SVG element, like the pop-up on hover.
I haven't taken a closer look at the accessibility since I noticed the failing tests, though.

hlfan:
What do you think of using the non-filled rss and diary+ icons instead? These would be thicker.

hlfan:
@pablobm would you like to take another look at the icons? I think Tom is waiting for your verdict.

pablobm:
Ooops, thank you for the ping :sweat_smile: I had a quick look now. I wasn't very comprehensive, but I think they look good. There are still some small differences, but I don't think they are necessarily for the worse.

tomhughes:
I guess this is good to merge then, thanks.

PR #6489, opened 2025-10-30
drop obsolete IE related code

tyrasd:
iD fully dropped support for IE [a while ago](https://github.com/openstreetmap/iD/blob/develop/CHANGELOG.md#2210). This comments were triggered only by the very oldest IE versions and prevented it from loading. As those old versions of IE are not able to parse the js of the modern versions of iD, it had no effect. And of course no mapper will want to use that ancient browser anyway.

### DIFF
--- a/app/views/site/id.html.erb
+++ b/app/views/site/id.html.erb
@@ -3,9 +3,7 @@
 <head>
   <meta charset='utf-8'>
   <%= stylesheet_link_tag "id" %>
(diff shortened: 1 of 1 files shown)


pablobm (approved):

gravitystorm:
Merged, thanks. Thanks also to @pablobm and @hlfan for the reviews!

PR #6490, opened 2025-10-31
Upgrade iD to v2.37.3

tyrasd:
Hotfix for https://github.com/openstreetmap/iD/issues/11532


tyrasd:
Yeah, the upstream issue is only triggered if you received the broken id-tagging-schema version during the couple hours the broken version was out before the fix was released, and that version was cached by your browser. So, it will not affect too many users, but I'd recommend to still deploy this to minimize the disruption for affected mappers.

PR #6492, opened 2025-11-03
Editor color mode preference

hlfan:
Like #5362, but for openstreetmap/iD#11531. See reasoning there.

### DIFF
--- a/app/controllers/preferences/basic_preferences_controller.rb
+++ b/app/controllers/preferences/basic_preferences_controller.rb
@@ -35,6 +35,11 @@ module Preferences
         success &= map_color_scheme_preference.update(:v => params[:map_color_scheme])
       end
 
+      if params[:editor_color_scheme]
+        editor_color_scheme_preference = current_user.preferences.find_or_create_by(:k => "editor.color_scheme")
+        success &= editor_color_scheme_preference.update(:v => params[:editor_color_scheme])
+      end
+
       success
     end
   end
--- a/test/controllers/preferences/basic_preferences_controller_test.rb
+++ b/test/controllers/preferences/basic_preferences_controller_test.rb
@@ -27,6 +27,7 @@ module Preferences
       user = create(:user, :languages => [])
       user.preferences.create(:k => "site.color_scheme", :v => "light")
       user.preferences.create(:k => "map.color_scheme", :v => "light")
+      user.preferences.create(:k => "editor.color_scheme", :v => "light")
       session_for(user)
 
       # Changing to a invalid editor should fail
@@ -39,6 +40,7 @@ module Preferences
       assert_nil user.preferred_editor
       assert_equal "light", user.preferences.find_by(:k => "site.color_scheme")&.v
       assert_equal "light", user.preferences.find_by(:k => "map.color_scheme")&.v
+      assert_equal "light", user.preferences.find_by(:k => "editor.color_scheme")&.v
 
       # Changing to a valid editor should work
       put basic_preferences_path, :params => { :user => { :preferred_editor => "id" } }
@@ -50,6 +52,7 @@ module Preferences
       assert_equal "id", user.preferred_editor
       assert_equal "light", user.preferences.find_by(:k => "site.color_scheme")&.v
       assert_equal "light", user.preferences.find_by(:k => "map.color_scheme")&.v
+      assert_equal "light", user.preferences.find_by(:k => "editor.color_scheme")&.v
 
       # Changing to the default editor should work
       put basic_preferences_path, :params => { :user => { :preferred_editor => "default" } }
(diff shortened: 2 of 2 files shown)


app/views/site/id.html.erb
```diff
@@ -13,7 +13,7 @@
      data[:token] = token.token
    end
    data[:locale] = ID::LOCALES.preferred(preferred_languages).to_s
-   data[:theme] = (preferred_color_scheme(:site) if preferred_color_scheme(:site) != "auto")
+   data[:theme] = (preferred_color_scheme(:editor) if preferred_color_scheme(:editor) != "auto") || (preferred_color_scheme(:site) if preferred_color_scheme(:site) != "auto")
```
pablobm: This got too complex here. Let's move it to a helper, preferably with a test.

app/views/preferences/basic_preferences/show.html.erb
```diff
@@ -32,5 +32,13 @@
                    :class => "form-select" %>
   </div>
 
+  <div class="mb-3">
+    <%= label_tag "editor_color_scheme", t(".preferred_editor_color_scheme"), :class => "form-label" %>
+    <%= select_tag "editor_color_scheme",
+                   options_for_select(%w[auto light dark].map { |scheme| [t(".editor_color_schemes.#{scheme}"), scheme] },
```
pablobm, line 38: The user-facing label "auto" is not clear enough here. There can be confusion as to how three dropdowns relate to each other. How about labelling it as "same as Website" on both `:editor` and `:map`?

app/views/site/id.html.erb
hlfan: What do you think of adding this to the preferred_color_scheme helper? (with tests)

pablobm: Yes, that might work :+1:

hlfan: I was able to reduce quite a bit of complexity by rewriting preferred_color_scheme, but now I'm unsure where to put a test. If you have any ideas...

app/views/layouts/_head.html.erb
```diff
@@ -3,7 +3,7 @@
   <%= javascript_include_tag "turbo", :type => "module" %>
   <%= javascript_include_tag "application" %>
   <%= javascript_include_tag "i18n/#{I18n.locale}" %>
-  <% if preferred_color_scheme(:site) == "auto" %>
+  <% if preferred_color_scheme(:site).nil? %>
```
tomhughes: If you reversed the branches here then you wouldn't need the `.nil?` which might be a bit clearer?

app/views/preferences/basic_preferences/show.html.erb
```diff
@@ -20,15 +20,23 @@
...
+                                      preferred_color_scheme(:site) || "auto"),
                    :class => "form-select" %>
   </div>
 
   <div class="mb-3">
     <%= label_tag "map_color_scheme", t(".preferred_map_color_scheme"), :class => "form-label" %>
     <%= select_tag "map_color_scheme",
                    options_for_select(%w[auto light dark].map { |scheme| [t(".map_color_schemes.#{scheme}"), scheme] },
-                                      preferred_color_scheme(:map)),
+                                      current_user.preferences.find_by(:k => "map.color_scheme")&.v || "auto"),
```
tomhughes: This seems like a step backwards but see my suggestion for changes to `preferred_color_scheme` for how we might avoid it.

app/controllers/application_controller.rb
```diff
@@ -275,11 +275,12 @@ def preferred_editor
   end
 
   def preferred_color_scheme(subject)
-    if current_user
-      current_user.preferences.find_by(:k => "#{subject}.color_scheme")&.v || "auto"
-    else
-      "auto"
-    end
+    scheme = current_user ? current_user.preferences.find_by(:k => "#{subject}.color_scheme")&.v : nil
```
tomhughes: Can this not use safe navigation like this?
```suggestion
    scheme = current_user&.preferences.find_by(:k => "#{subject}.color_scheme")&.v
```

app/controllers/application_controller.rb
```diff
@@ -275,11 +275,12 @@ def preferred_editor
   end
 
   def preferred_color_scheme(subject)
```
tomhughes: I think that rather than taking a single subject here and then having to special case `:site` when deciding whether to fallback this could take `*subjects` and then try each one in turn until it finds one that is set to something other that `auto`.

Then you could do a call like `preferred_color_scheme(:map, :site)` when you want fallback or `preferred_color_scheme(:map)` when you don't?

Possibly it could also take a default value to return as a last resort instead of `nil` but that would probably need a keyword argument so it might be easier to stick with `||` at the call site unless we just go back to returning `auto` in that case.

app/controllers/application_controller.rb
hlfan: Only with offending RuboCop. But not going into the subjects loop seems even better.

pablobm:
For things like this, I'd suggest moving code to models (or dedicated Ruby classes) where they can be tested thoroughly without too much ceremony.

Here's an example (feel free to steal): https://github.com/hlfan/openstreetmap-website/compare/editor-theme...pablobm:openstreetmap-website:editor-theme The second commit is refactor idea, take it or leave it.

Issue #6493, opened 2025-11-03
Align the button in profile page.


daishu0000:
The "My Profile" page of osm has two buttons: "Edit Profile Details" and "Edit Description". However, these two buttons  have different size and are not aligned, which looks unattractive. I think it's better to align them. 

<img width="2828" height="1482" alt="Image" src="https://github.com/user-attachments/assets/42d89ddc-2ae9-43d6-93d5-355d16b9dc31" />

pablobm:
Thank you @daishu0000 for the report. I had noticed this too recently 😅  Do you have knowledge of web development? Would you be able to contribute a fix? I think this one needs only HTML and CSS knowledge, no Ruby.

It still needs to be able to run the development server in order to check it locally, so it's not trivial. However it might be easier than it looks. I'm happy to provide guidance and even pair on it, if it helps. Same for anyone else who volunteers on this thread.

tomhughes:
I think it might be more complicated than that depending on which one you want to standardise on...

The smaller one has the `btm-sm` class and removing that makes it the same size as the larger one but adding that to the larger one doesn't seem to work to make it smaller, possibly because that one isn't a real button but is a link styled as a button.

daishu0000:
> Would you be able to contribute a fix

I'm trying to fix it.

Hiteshpy07:
hey @daishu0000 , i noticed the issue you listed ,it is a simple UI fix , can i fix it ?

tomhughes:
@daishu0000 has already opened a PR to fix it?

daishu0000:
> hey [@daishu0000](https://github.com/daishu0000) , i noticed the issue you listed ,it is a simple UI fix , can i fix it ?

> [@daishu0000](https://github.com/daishu0000) has already opened a PR to fix it?

see #6494, I have opened a PR and alreadey fixed it

PR #6494, opened 2025-11-03
Fix alignment issues on the profile page

daishu0000:
about #6493
fix the problem that "Edit Profile Details" and "Edit Description" different size and are not aligned.
old version
image
new version
image

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -288,8 +288,8 @@
     <div class="col-sm-9">
       <div class="richtext text-break"><%= @user.description.to_html %></div>
       <% if owned %>
-        <div class="mb-3">
-          <%= link_to t(".edit_description"), profile_description_path, :class => "btn btn-outline-primary" %>
+        <div class="mt-2 mb-3 col-sm-4">
+          <%= link_to t(".edit_description"), profile_description_path, :class => "btn btn-sm w-100 btn-outline-primary" %>
         </div>
       <% end %>
     </div>


tomhughes:
Please don't do merges onto a PR branch - either rebase or just leave it as we will do a merge back anyway.

daishu0000:
> Please don't do merges onto a PR branch - either rebase or just leave it as we will do a merge back anyway.

Ok, I have reseted.

app/views/users/show.html.erb
```diff
@@ -285,11 +285,11 @@
         </div>
       <% end %>
     </div>
-    <div class="col-sm-9">
+    <div class="col-sm-3">
```
tomhughes: What was the purpose of this change? It causes the user description to be crushed into a narrow column like this:

Image

instead of being full width as it was before:

Image

daishu0000: See commit [af1b188](https://github.com/openstreetmap/openstreetmap-website/pull/6494/commits/af1b1886d17e058d37d957aaf968594d35687d1c), I have resolved it, making sure text is full width and two buttons have the same width.

daishu0000:
Now the text can be shown as before @tomhughes 
image

pablobm:
Would you be able to also align the buttons that appear after clicking on "Edit profile details"?

Different buttons to edit profile details, with slightly different aspects

I appreciate that those are links instead of buttons, so they may require different techniques and not be trivial. Let us know how it goes.

daishu0000:
> Would you be able to also align the buttons that appear after clicking on "Edit profile details"?

@pablobm Technically, it’s possible to align those buttons, but I’m not sure if that is a better design choice. Making the buttons slightly smaller might better convey their subordinate relationship to “Edit profile details.”

Here’s what it looks like when they are aligned:

image

For comparison, here’s the original version:

image

pablobm:
That's a good point. However they are still jarring when open. Perhaps the solution should be different?

My first thought was about not doing the collapse/expand, and instead show placeholders, each with an "edit" link. However this may get complicated again as it's not trivial to have placeholders for links and image... But wait, why a button for the image? It can be edited by clicking on the image already! :thinking:

How about this: let's accept this PR as it solves an immediate issue, and we can see about redesigning that section in a different PR?

pablobm (approved):

tomhughes (changes requested):
This looks good now - if you could just squash the fixup commit into the original then we can get it merged.

daishu0000:
> if you could just squash the fixup commit into the original then we can get it merged

@tomhughes OK, I have squashed the commit.

tomhughes (approved):
This looks good now, thanks.

PR #6501, opened 2025-11-05
Adjust Bootstrap icon usage

hlfan:
Addendum to #6487. Touches some adjustments from #6215.


app/views/diary_entries/_profile_diaries.html.erb
```diff
@@ -6,20 +6,20 @@
...
+              <i class="bi bi-journal-text fs-2 my-n2 flex-shrink-0" title="<%= t ".title_label" %>" aria-hidden="true"></i>
               <span class="visually-hidden"><%= t ".title_label" %></span>
               <%= link_to entry.title, diary_entry_path(@user, entry), :class => "align-self-center lh-sm" %>
             </p>
             <p class="card-text flex-grow-1"><%= truncate(strip_tags(entry.body.to_html), :length => 150) %></p>
 
             <div class="card-text d-flex justify-content-between align-items-center mt-auto">
               <small class="text-body-secondary d-flex gap-1 align-items-center">
-                <i class="bi bi-chat fs-5 my-n1" aria-hidden="true"></i>
+                <i class="bi bi-chat fs-5 my-n1" title="<%= t ".comments_label" %>" aria-hidden="true"></i>
```
pablobm: Is this specific `title` needed? Also the `span` right after. They are followed by an "X comments"  line, so I feel they might be redundant.

app/views/diary_entries/_profile_diaries.html.erb
```diff
@@ -6,20 +6,20 @@
...
                 <span class="visually-hidden"><%= t ".comments_label" %></span>
                 <%= link_to t(".comments", :count => entry.comments.size), diary_entry_path(@user, entry, :anchor => "comments"), :class => "text-body-secondary" %>
               </small>
               <small class="text-body-secondary d-flex gap-1 align-items-center">
-                <i class="bi bi-calendar fs-5 my-n1" aria-hidden="true"></i>
+                <i class="bi bi-calendar fs-5 my-n1" title="<%= t ".date_label" %>" aria-hidden="true"></i>
```
pablobm: Similarly here: perhaps the `title` and `span` are not necessary as the content is self-explanatory.

app/views/users/_role_icons.html.erb
```diff
@@ -1,20 +1,22 @@
...
     <% if @user.role?(role) %>
-      <%= link_to inline_svg_tag("icons/role-star.svg", :class => "role-icon #{role}",
-                                                        :title => t(".revoke.#{role}")),
-                  user_role_path(@user, role),
+      <%= link_to user_role_path(@user, role),
                   :method => :delete,
-                  :data => { :confirm => t(".revoke.are_you_sure", :name => @user.display_name, :role => role) } %>
+                  :data => { :confirm => t(".revoke.are_you_sure", :name => @user.display_name, :role => role) } do %>
+        <i class="bi bi-star-fill fs-5 role-<%= role %> align-middle" title="<%= t ".revoke.#{role}" %>" aria-hidden="true"></i>
+        <span class="visually-hidden"><%= t ".revoke.#{role}" %></span>
```
pablobm: Related to the above: do we need both the `title` and the `span`? I suspect a single one will suffice. As for which one... any a11y experts in the room? 😅

app/views/users/_role_icons.html.erb
```diff
@@ -1,20 +1,22 @@
...
     <% else %>
-      <%= link_to inline_svg_tag("icons/role-star.svg", :class => "role-icon #{role} inactive",
-                                                        :title => t(".grant.#{role}")),
-                  user_role_path(@user, role),
+      <%= link_to user_role_path(@user, role),
                   :method => :post,
-                  :data => { :confirm => t(".grant.are_you_sure", :name => @user.display_name, :role => role) } %>
+                  :data => { :confirm => t(".grant.are_you_sure", :name => @user.display_name, :role => role) } do %>
+        <i class="bi bi-star fs-5 role-<%= role %> align-middle" title="<%= t ".grant.#{role}" %>" aria-hidden="true"></i>
+        <span class="visually-hidden"><%= t ".grant.#{role}" %></span>
```
pablobm: Same here.

app/views/users/show.html.erb
```diff
@@ -253,7 +253,7 @@
                                     :edit_title => t(".edit_location"),
                                     :edit_path => profile_location_path do %>
         <div class="icon-link mw-100">
-          <i class="align-self-baseline bi bi-geo-alt-fill flex-shrink-0 fs-6" aria-hidden="true"></i>
+          <i class="align-self-baseline bi bi-geo-alt-fill flex-shrink-0 fs-6" title="<%= t ".home_location" %>" aria-hidden="true"></i>
           <span class="visually-hidden"><%= t ".home_location" %></span>
```
pablobm: Same: probably one is enough.

app/views/users/show.html.erb
```diff
@@ -265,7 +265,7 @@
                                     :edit_title => t(".edit_company"),
                                     :edit_path => profile_company_path do %>
         <div class="icon-link mw-100">
-          <i class="align-self-baseline bi bi-buildings-fill flex-shrink-0 fs-6" aria-hidden="true"></i>
+          <i class="align-self-baseline bi bi-buildings-fill flex-shrink-0 fs-6" title="<%= t ".company" %>" aria-hidden="true"></i>
           <span class="visually-hidden"><%= t ".company" %></span>
```
pablobm: And same.

app/views/diary_entries/_profile_diaries.html.erb
hlfan: It only occurred to me once I was testing something else on https://master.apis.dev.openstreetmap.org/ that the hover tooltips need a title attribute.

And since the Bootstrap icons use special characters in the pseudo-elements, I'm not happy with having this maybe appear on screen readers.

Hence twice.

pablobm: In this specific case, what I mean is that I think it's not necessary to have anything working as a title (neither icon title nor `span`) because they are followed by "X comments" which is self-explanatory.

I just tried with iOS VoiceOver and with the [Read Aloud](https://addons.mozilla.org/en-US/firefox/addon/read-aloud/) extension on Firefox. In this specific instance both say "comments, zero comments". I think it can do with just saying "zero comments", with no prologue.

Also in my test, it appears to be the `span` that is read by these tools, so perhaps it can be removed and the icon title left in place, although I don't know that the title serves any purpose. If it's any like the `alt` attribute in images, then it's best not to use it unless the image adds significant meaning, which is not the case here.

app/views/users/_role_icons.html.erb
pablobm: Following my experiment above, I think in these cases we should leave the span in place and remove the icon title.

app/views/diary_entries/_profile_diaries.html.erb
hlfan: What do you think of removing the span elements and adding role=img, aria-label=* AND title=* instead of the aria-hidden?
Then there's just one element, even though there's still two translated texts specified.

app/views/diary_entries/_profile_diaries.html.erb
hlfan: In the actual blog entry, the date is put in a sentence, so I think both can be dropped.

app/views/users/_role_icons.html.erb
hlfan: Since this logic branch is only visible for admins, I'd drop the span and leave the title for the tooltips.

app/views/diary_entries/_profile_diaries.html.erb
pablobm: In this specific case I would remove both the title and the span. I don't think they add anything. The i18n texts can be removed too then.

app/views/users/_role_icons.html.erb
pablobm: I tested iOS VoiceOver and Read Aloud (on Firefox) again. They have slightly different behaviours that are driving me a bit crazy, particularly how Read Aloud doesn't always reload the page properly :angry: 

The following seems to work with both:

```
<i class="bi bi-star-fill fs-5 role-<%= role %> align-middle" title="<%= t ".revoke.#{role}" %>" role="button" aria-pressed="true"><span class="visually-hidden"><%= t ".revoke.#{role}" %></span></i>
```

On iOS the text is read twice, annoyingly. Not sure if this is best. Read Aloud appears to read what's in the browser's Accessibility Tree so I would think that's relevant...? No idea :sob:

gravitystorm:
@hlfan do you consider this ready, or are you planning further tweaks?

hlfan:
Haven't gotten around to play with the ARIA details...

app/views/shared/_loader.html.erb
```diff
@@ -0,0 +1,5 @@
+<%# locals: () %>
+
+<div class="spinner-border" role="status">
+  <span class="visually-hidden"><%= t("browse.start_rjs.loading") %></span>
```
tomhughes: Should we move that string to the `shared` scope, or maybe create a duplicate of it there? Note that if we move it and the js still uses it then we need to make sure it's still exported.

app/helpers/application_helper.rb
```diff
@@ -11,7 +11,7 @@ def linkify(text)
   end
 
   def rss_link_to(args = {})
-    link_to tag.i(:class => "bi bi-rss-fill fs-6", :aria => { :hidden => "true" }), args
+    link_to tag.i(:class => "bi bi-rss-fill fs-6 role-administrator", :aria => { :hidden => "true" }), args
```
tomhughes: I'm not sure I'm keen on abusing the `role-administrator` class like this just because it happens to be the same colour as the RSS icon...

app/views/shared/_loader.html.erb
hlfan: I wouldn't move it with this pr, otherwise translatewiki might move the lesser translated language selector one instead.

app/helpers/authorization_helper.rb
```diff
@@ -8,8 +8,8 @@ def authorization_scope(scope)
     html << t("oauth.scopes.#{scope}")
     if Oauth::MODERATOR_SCOPES.include? scope
       html << " "
-      html << inline_svg_tag("icons/role-star.svg", :class => "role-icon moderator align-text-bottom",
-                                                    :title => t("oauth.for_roles.moderator"))
```
pablobm, line 12: How can I see this page in local? 😄

tomhughes: There's a few places it is used but the easiest one to see is to go to `/oauth2/applications` and add an application that requests a moderator scope  (either `write_redactions` or `write_blocks`) and then go back to the list to see the star.

app/helpers/authorization_helper.rb
```diff
@@ -8,8 +8,8 @@ def authorization_scope(scope)
...
+      html << tag.i(tag.span(t("oauth.for_roles.moderator"), :class => "visually-hidden"),
+                    :class => "bi bi-star-fill fs-5 role-moderator align-middle")
```
pablobm: Played a bit with this. The one thing that bothered me slightly was on iOS VoiceOver, where it would stop for a moment on the star icon without saying anything, before moving on to the text description. I'm not an expert, but I think this may hint to users that there's something but it's hidden from them, which shouldn't be the message here.

Instead, if we separate these two elements and add an `aria-hidden="true"` on the icon, this disappears from the accessibility tree and VoiceOver ignores it, while still reading out the text description:

```suggestion
      html << tag.i(:class => "bi bi-star-fill fs-5 role-moderator align-middle", "aria-hidden" => true)
      html << tag.span(t("oauth.for_roles.moderator"), :class => "visually-hidden")
```

hlfan: Huh, that was where I started before https://github.com/openstreetmap/openstreetmap-website/pull/6501#discussion_r2504458134 (but without the title attribute)...

hlfan:
I've replaced two additional SVGs:

|reverse_directions|feed|
|-|-|
|||

hlfan:
> do you consider this ready[...]?

Now I do, @gravitystorm.

tomhughes:

Issue #6234, opened 2025-07-23
Migrate new vector layers from maplibre-gl-leaflet to “unwrapped” maplibre-gl-js


controversial:
### Problem

It’s really exciting to have vector tiles hosted for the first time on openstreetmap.org!

Currently, vector tiles are still displayed using `maplibre-gl-leaflet`, which wraps `maplibre-gl-js` in the same Leaflet API that we use to display raster maps like Carto.

Using the leaflet wrapper for the vector maps holds the vector maps back in a container that’s designed around the constraints/limitations of raster maps:
- No “smooth zooming” between integer zoom levels
- No ability to change map bearing (rotation) / pitch

### Description

It would be great to migrate from Leaflet to mapbox-gl-js. There are several pieces that would need to be migrated to different APIs, some of which are documented by @tordans in https://github.com/gravitystorm/openstreetmap-website/issues/289.

Theoretically, using maplibre-gl-js directly we could:
- support smooth zooming, a big user experience improvement that will help the map feel more modern and responsive.
- support rotating the map
- support _potential_ future features, including but not limited to:
    - clickable POIs and features that can surface simple data without a round-trip to overpass.
    - a 3D [globe projection](https://maplibre.org/maplibre-gl-js/docs/examples/add-a-simple-custom-layer-on-a-globe/) at low zoom levels
    - Interactive map features that are _inside_ the map instead of just on top, for example a routing line that appears faded when it passes under a tunnel or through a building
    - extruded 3D buildings based on building height
    -  the ones listed [here](https://github.com/gravitystorm/openstreetmap-website/issues/289#issuecomment-2878035396) by @1ec5
- “modernize” the front-end a bit by using a more contemporary map renderer / javascript library

tomhughes:
As I explained in the other ticket while we'd love to do this I strongly suspect that maplibre-gl-js is missing many of the features we need at the moment. I did a test re-implemenation of my personal site at maps.compton.nu a while ago and found it very hard to get even close to feature parity.

controversial:
`maplibre-gl-leaflet` is a wrapper around `maplibre-gl-js`, so to my understanding any features it implements are supported by `maplibre-gl-js`.

do you have any examples of the missing features we need?

tomhughes:
A not necessarily complete list of things we need:

* Ability to overlay markers with optional popups, including draggable markers for note addition and route endpoints
* Ability to overlay arbitrary geometries for routes and data visualisation
[...]

PR #6504, opened 2025-11-06
Migrate the dashboard entry page from leaflet to maplibre

CommanderStorm:
### Description

As discussed in the following thread, this PR is the first PR to migrate leaflet -> maplibre
- https://github.com/openstreetmap/openstreetmap-website/issues/6234

Longer term, this migration has more than the immediate benefits as discussed in the issue.

### How has this been tested?

- Manually by interacting with the map
- automated testing

image
image

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -9,9 +9,9 @@ $(function () {
   if ($("#map").length) {
     map = new maplibregl.Map({
       container: "map",
-      style: OSM.Mapnik,
+      style: OSM.MapLibre.Styles.Mapnik,
       attributionControl: false,
-      locale: OSM.MaplibreLocale,
+      locale: OSM.MapLibre.Locale,
       center: OSM.home ? [OSM.home.lon, OSM.home.lat] : [0, 0],
       zoom: OSM.home ? defaultHomeZoom : 0
     });
@@ -24,14 +24,14 @@ $(function () {
       },
       trackUserLocation: true
     });
-    map.addControl(new OSM.CombinedControlGroup([navigationControl, geolocateControl]), position);
+    map.addControl(new OSM.MapLibre.CombinedControlGroup([navigationControl, geolocateControl]), position);
 
     $("[data-user]").each(function () {
       const user = $(this).data("user");
       if (user.lon && user.lat) {
-        OSM.createMapLibreMarker({ icon: "dot", color: user.color })
+        OSM.MapLibre.getMarker({ icon: "dot", color: user.color })
           .setLngLat([user.lon, user.lat])
-          .setPopup(OSM.createMapLibrePopup(user.description))
+          .setPopup(OSM.MapLibre.getPopup(user.description))
           .addTo(map);
       }
     });
--- a/app/assets/javascripts/maplibre.combinedcontrolgroup.js
+++ b/app/assets/javascripts/maplibre.combinedcontrolgroup.js
@@ -31,4 +31,4 @@ class CombinedControlGroup {
   }
 }
 
-OSM.CombinedControlGroup = CombinedControlGroup;
+OSM.MapLibre.CombinedControlGroup = CombinedControlGroup;
--- a/app/assets/javascripts/maplibre.i18n.js
+++ b/app/assets/javascripts/maplibre.i18n.js
@@ -1,6 +1,6 @@
 //= require i18n
 
-OSM.MaplibreLocale = {
+OSM.MapLibre.Locale = {
   "GeolocateControl.FindMyLocation": OSM.i18n.t("javascripts.map.locate.title"),
   "GeolocateControl.LocationNotAvailable": OSM.i18n.t("javascripts.map.locate.not_available"),
   "NavigationControl.ZoomIn": OSM.i18n.t("javascripts.map.zoom.in"),
--- a/app/assets/javascripts/maplibre.map.js
+++ b/app/assets/javascripts/maplibre.map.js
@@ -1,6 +1,6 @@
 //= require maplibre-gl/dist/maplibre-gl
 
-OSM.Mapnik = {
+OSM.MapLibre.Styles.Mapnik = {
   version: 8,
   sources: {
     osm: {
@@ -23,7 +23,7 @@ OSM.Mapnik = {
 
 // Helper function to create Leaflet style (SVG comes from Leaflet) markers for MapLibre
 // new maplibregl.Marker({ color: color }) is simpler, but does not have the exact same gradient
-OSM.createMapLibreMarker = function ({ icon = "dot", color = "var(--marker-red)", ...options }) {
+OSM.MapLibre.getMarker = function ({ icon = "dot", color = "var(--marker-red)", ...options }) {
   const el = document.createElement("div");
   el.className = "maplibre-gl-marker";
   el.style.width = "25px";
@@ -58,7 +58,7 @@ OSM.createMapLibreMarker = function ({ icon = "dot", color = "var(--marker-red)"
 };
 
 // Helper function to create MapLibre popups that don't overlap with Leaflets' markers
-OSM.createMapLibrePopup = function (content) {
+OSM.MapLibre.getPopup = function (content) {
   // General offset 5px for each side, but the offset depends on the popup position:
   // Popup above the marker -> lift it by height + 5px = 45px
   // Popup left the marker -> lift it by width/2 + 5px = 22.5px ~= 17px
--- a/app/assets/javascripts/osm.js.erb
+++ b/app/assets/javascripts/osm.js.erb
@@ -228,5 +228,9 @@ OSM = {
     modalBody.textContent = message;
     const alertModal = new bootstrap.Modal(document.getElementById("osm_alert_modal"));
     alertModal.show();
+  },
+
+  MapLibre: {
+    Styles: {}
   }
 };


app/views/layouts/_head.html.erb
```diff
@@ -10,6 +10,7 @@
   <% end %>
   <%= stylesheet_link_tag "print-#{dir}", :media => "print" %>
   <%= stylesheet_link_tag "leaflet-all", :media => "screen, print" %>
+  <link href="https://unpkg.com/maplibre-gl@^5.11.0/dist/maplibre-gl.css" rel="stylesheet" />
```
CommanderStorm: @pablobm
I am taking your comment as an OK to ask you questions.

I have a bit of trouble with this system and have no clue how to fix it.
THis right here obviously is not production quality and needs to be fixed before the PR can move into review.

**How do I do this in the rails?** 

I have already tried adding below lines, but that does not seem the way to handle this:

https://github.com/OpenHistoricalMap/ohm-website/blob/3806c1b5c294e0656ee82cc5840ddccde67068dd/config/initializers/assets.rb#L22

https://github.com/OpenHistoricalMap/ohm-website/blob/3806c1b5c294e0656ee82cc5840ddccde67068dd/app/views/layouts/_head.html.erb#L16

The error that this is causing is the following:

image

app/assets/javascripts/dashboard.js
```diff
@@ -1,32 +1,52 @@
...
+          }
+        },
+        "layers": [
+          {
+            "id": "osm",
+            "type": "raster",
+            "source": "osm",
+          }
+        ],
+        "projection": { "type": "globe" },
```
CommanderStorm: The globe is the only only thing which should be not precicely as before for users.

If osm prefers a flat eath (pun intended), this can be removed to use webmercator instead

```suggestion
```

CommanderStorm (commented):
I am having a bit of trouble with the asset system that I did no have when I was doing the PR for OHM. There seems to be a diff, but I have no clue which part of the diff is the relevant part

app/assets/javascripts/dashboard.js
gravitystorm: In the interests of making the PR as small/incremental as possible (i.e. with as few things for reviewers to ponder/worry about as possible) I'd suggest sticking with Web Mercator for now, and leave switching to a globe projection for a separate PR later.

app/views/layouts/_head.html.erb
tomhughes: Well we don't want to be pulling it in from a third party site to start with - you should be adding it to `package.json` and serving it locally.

tomhughes: Surely it also needs the javascript anyway? The style file on it's own isn't going to do anything...

1ec5: We’re already pulling in MapLibre GL JS elsewhere in the codebase, so you’ll want to look into replicating that for the profile page. This looks like a JavaScript comment, but we have a Rails-related build step that actually includes the library somehow. I think it ends up pulling in the stylesheet as well, because it’s part of the NPM package.

https://github.com/openstreetmap/openstreetmap-website/blob/8809986357776af595946d5878e57d1e7352da47/app/assets/javascripts/leaflet.maplibre.js#L1

app/assets/javascripts/dashboard.js
tomhughes: Well I doubt we would switch to it as the default anyway... I thought the point here was to make it possible to offer it as an option not to change the default.

app/views/layouts/_head.html.erb
tomhughes: No that won't pull in any stylesheets, but a similar comment in one of the stylesheet files could.

app/assets/javascripts/dashboard.js
1ec5: The profile map probably doesn’t benefit from globe view, even if the slippy map on the homepage might. If the map is zoomed out enough that the globe projection is discernible, chances are that it’s zoomed out too much to see the pins near your home location. That said, it’s nice to see that the feature is trivial to enable if warranted.

app/assets/stylesheets/maplibre-gl-all.scss
CommanderStorm, line 1: not sure if the css should be moved from common.scss here.

I did it this way, because OHM does it like this.

app/assets/stylesheets/common.scss
```diff
@@ -219,6 +219,54 @@ body.small-nav {
...
+}
+.maplibregl-ctrl button.maplibregl-ctrl-zoom-out .maplibregl-ctrl-icon {
+  background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='20' height='20' fill='%23ffffff' viewBox='0 0 20 20'%3E%3Cpath d='M4 9v1l1 1H16l1-1V9L16 8H5Z'/%3E%3C/svg%3E") !important;
+}
+.maplibregl-ctrl button.maplibregl-ctrl-geolocate .maplibregl-ctrl-icon {
+  background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='20' height='20' fill='%23ffffff' viewBox='0 0 20 20'%3E%3Cpath d='M10 10v6h2L16 6V4H14L4 8v2Z'/%3E%3C/svg%3E") !important;
+}
+.maplibregl-ctrl button.maplibregl-ctrl-geolocate.maplibregl-ctrl-geolocate-active .maplibregl-ctrl-icon {
+  background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='20' height='20' fill='%2376c551' viewBox='0 0 20 20'%3E%3Cpath d='M10 10v6h2L16 6V4H14L4 8v2Z'/%3E%3C/svg%3E") !important;
+}
```
hlfan: This feels like a big https://github.com/openstreetmap/openstreetmap-website/labels/dx regression from #6200.

test/system/dashboard_test.rb
```diff
@@ -62,7 +62,7 @@ class DashboardSystemTest < ApplicationSystemTestCase
       assert_no_text "Your location"
       assert_no_link "Fred Tester"
 
-      find(".leaflet-marker-icon").click
+      find(".maplibregl-marker", :wait => 60).click
```
CommanderStorm: I have no clue why this does not find the marker.
he marker defintively exists.

image

app/assets/stylesheets/maplibre-gl-all.scss
pablobm: Good question. Personally I would expect the MapLibre styles to live here, but I don't know if there are any historical reasons why it's done differently. It was done like this in the original PR (https://github.com/openstreetmap/openstreetmap-website/pull/121) and I can't see an explanation as to why.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -0,0 +1,6 @@
+
+/*
+ *= require maplibre-gl
+ */
+
+/* BASED ON leaflet-all.scss */
```
pablobm: This comment probably can go. (I see it's like this in OHM.)

test/system/dashboard_test.rb
pablobm: Was this giving you trouble? 60 seconds sounds a bit excessive to me.

pablobm (commented):
Tested in my machine and looks great. Even gave the globe view a whirl 🙂 I'll leave to others to review though as I'm sure there's tricky, non-obvious stuff.

app/assets/stylesheets/common.scss
CommanderStorm: Not sure which alternative you are suggesting.

I tried to do the following, but the ruby part does not seem to allow this, at least the `inline-image` does not seem to do anything.

```css
   background-image: inline-image('app/assets/images/map-controls/zoomin.svg') !important;
```

https://stackoverflow.com/questions/33299453/inline-svg-from-a-file-with-sass/33302941#33302941

How do I enable this functionality in the asset pipeline?

As I said previusly, I know JS but ruby is not taught in school, so I would need help with the magic there.

I ruled out https://github.com/zemax/node-sass-svg as it does not seem very active.

app/assets/stylesheets/maplibre-gl-all.scss
CommanderStorm:
```suggestion
```

test/system/dashboard_test.rb
CommanderStorm: See my comment above:
- https://github.com/openstreetmap/openstreetmap-website/pull/6504#discussion_r2499069366

Not sure why this does not work or how to make this testcase run.

It works in every environment I can test.

The 60s was just to make reeeely sure that it is not the timeout

tomhughes: Well in general we need to avoid arbitrary timeouts because they will be too long, until they aren't and a test fails because a GHA runner was unusually slow.

CommanderStorm: here is the CI run for reference:
https://github.com/CommanderStorm/ohm-website/actions/runs/19137661729

tomhughes:
Of course one problem of tackling this by starting with this one use case is that it doesn't actually help prove that maplibre is actually capable of doing everything we need for the main map - what if we start down this path and then get stuck later on?

I only ask because I know when I was trying to move my own map site to maplibre that I found it was generally lacking in features compared to leaflet and I eventually put the conversion on hold.

CommanderStorm:
> I only ask because I know when I was trying to move my own map site to maplibre that I found it was generally lacking in features compared to leaflet and I eventually put the conversion on hold.

which features are you missing?

And more broadly, is there an alternative approach that you would like to see?
Starting with another part is also fine, but doing everything in one PR will get too messy.
We can always revert if this path does not work (I don't think that will be nessesary)

app/assets/stylesheets/common.scss
hlfan: I'd suggest using the return value of [`geolocate.onAdd(map)`](https://maplibre.org/maplibre-gl-js/docs/API/interfaces/IControl/#onadd), which is the HTML container element.
That can be manipulated like was done in #5752, where the same issue was brought up.

tomhughes:
It was nearly two years ago now so it's hard to remember and things may well have improved - looking at my branch it looks like I struggled with layer (or should I say style) switching and I never got geocoding working though I may have given up before attempting that.

I seem to recall that drawing tools were hard as well, or rather than they all wanted to add geometries and missed the ability of the leaflet ones to be able to "draw" a marker instead - not sure if we use that ourselves though.

I think the final straw (which definitely doesn't affect us) was probably when I realised that I was probably never going to be able to include the google tile layers.

I did attempt in https://github.com/openstreetmap/openstreetmap-website/issues/6234#issuecomment-3109509258 to enumerate what features we needed for osm.org.

CommanderStorm:
> It was nearly two years ago now so it's hard to remember and things may well have improved - looking at my branch it looks like I struggled with layer (or should I say style) switching and 

Not sure what was the problem there.

Style switching works as demonstrated my https://maputnik.maplibre.org or https://map-gl-style-switcher.netlify.app/

Finer grained interactivity is also possible, no need to switch out everything.
See https://maputnik.maplibre.org or https://nav.tum.de/view/mi for examples of this.

> I never got geocoding working though I may have given up before attempting that.

Geocoding is not something that maplibre does. We are a renderer, but nominatim has this functionality.

If you are refering to geolocation, that works quite well. IMO better than leaflet given the animation ^^

```js
maplibregl.GeolocateControl({
  positionOptions: {
    enableHighAccuracy: true
  },
  trackUserLocation: true
});
```
 
> I seem to recall that drawing tools were hard as well, or rather than they all wanted to add geometries and missed the ability of the leaflet ones to be able to "draw" a marker instead - not sure if we use that ourselves though.

The tooling in this area certainly has improved.

Back 2y ago, we were still quite new and plugins like https://terradraw.io/ (among others) did not exist (CC @JinIgarashi)

> I think the final straw (which definitely doesn't affect us) was probably when I realised that I was probably never going to be able to include the google tile layers.

That actually is possible, but implementations of this [are not open source](https://linkedin.com/posts/yurik_electric-vehicle-charging-stations-rivian-activity-7353477614587506688-6OnL).
Not a concearn for OSM though 😉

> I did attempt in [#6234 (comment)](https://github.com/openstreetmap/openstreetmap-website/issues/6234#issuecomment-3109509258) to enumerate what features we needed for osm.org.

I think @controversial already went into all the cases.
We support all of them. I see no issues on that side.

test/system/dashboard_test.rb
pablobm: I suspect that WebGL is not enabled in the browser in CI.

How's the CI browser set up anyway? :thinking:

tomhughes: I mean it's firefox in it's default configuration but webgl is likely disabled in headless mode, or rather when it's not running on a display with GL support?

CommanderStorm: Yes, issue is likely firefox-headless-related.

It is configured here:

https://github.com/openstreetmap/openstreetmap-website/blob/9ff2103e60832de0f45c993fda22e4ca51e5303b/test/teaspoon_env.rb#L109-L117

Firefox is being difficult with regards to webgl in `--headless` mode for some reason:
https://bugzilla.mozilla.org/show_bug.cgi?id=1375585#c27

I have some ideas how to workaound this upstream issue.
Need some time to figure out which of them works and which does not. 😉

The error message is just misleading.

pablobm: I don't think it's headless mode alone, as that's what I run locally and it works for me.

Now, if I understand this correctly, headless or not it runs the browser installed in the machine. So I guess those CI machines don't have GL support or it's not configured.

Having said that: there's people still reporting this issue at https://bugzilla.mozilla.org/show_bug.cgi?id=1375585... so I don't understand why it works in my machine.

pablobm: Someone else reporting this issue on GH Actions: https://github.com/orgs/community/discussions/60273. According to them it's random, some machines have it, some don't.

1ec5: Does this `click` method call a standard DOM method? I wonder if `maplibregl.Marker` might be expecting a custom `click` event, based on [this migration guide](https://maplibre.org/maplibre-gl-js/docs/guides/leaflet-migration-guide/#handling-click-events).

app/assets/javascripts/dashboard.js
```diff
@@ -1,32 +1,57 @@
...
+          }
+        },
+        layers: [
+          {
+            id: "osm",
+            type: "raster",
+            source: "osm"
+          }
+        ]
+      },
```
1ec5: Long-term, we’ll probably want a system for centralizing style configuration like we have for layers on the main slippy map:

https://github.com/openstreetmap/openstreetmap-website/blob/9ff2103e60832de0f45c993fda22e4ca51e5303b/app/assets/javascripts/leaflet.maptiler.js#L4

test/system/dashboard_test.rb
CommanderStorm: So CI is now green on the GHA side.
Have not touched docker, so that still needs a bit of work.

app/assets/stylesheets/common.scss
CommanderStorm: I am not sure if I follow your rationale fully :sweat_smile:.

What kind of https://github.com/openstreetmap/openstreetmap-website/labels/dx are we optimising for?
Neither [`sprite.svg`](https://github.com/openstreetmap/openstreetmap-website/blob/59d834114145771d228dd0a728b026ed2eb13bd3/app/assets/images/sprite.svg), nor item in the new [`map-controls`](https://github.com/hlfan/openstreetmap-website/tree/master/app/assets/images/map-controls) directory were changed in the last 2 years, so I don't expect this to materially change.
The icons seem to be working.

Your approach means:
- having every control you add to the map has to explicitely change its appearance via DOM interaction
- having to have the svg in scope via adding the following (kind of magic unless you know that these are map icons)
  ```html
  <svg class="d-none"><defs><%= render :partial => "layouts/control_icons", :locals => { :icons => %w[zoomin zoomout geolocate] } %></defs></svg>
  ```
- having to know that the indirect linking of the svg does what it does.
- having to subclass every Control to change the DOM element. 
 `geolocate.onAdd(map)` is not what we want, we want control over placement, i.e. top right in this instance, too.

The existing solution has the cost of this is a lot of abstraction "magic" and code duplication.
In the definting it in CSS way, this is only defined once in the CSS (where the rest of the styling code lives), not on every map control adding to the map.

**Are you sure that I should do this refactoring?**

After a bit of searching, I found `image-url()` which works as expected and solves this too.
Downside: no inlined SVGs.

Given that the map takes some time to intalise, the few ms to the CDN are likely fine.
With caching, likely not an issue.

Alternatively, I could do a custom scss function, but that also is not actually simpler, given how often it would be used.

app/assets/stylesheets/maplibre-gl-all.scss
CommanderStorm: I moved the code which is map specific.

Not sure if I should move the popup code too.

app/assets/javascripts/dashboard.js
CommanderStorm: Agree, but lets move this to another PR. There is already enough here to review

test/system/dashboard_test.rb
CommanderStorm: CI is now green on the docker side as well.
https://github.com/CommanderStorm/ohm-website/actions

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -0,0 +1,62 @@
...
+
+.maplibregl-ctrl button.maplibregl-ctrl-zoom-out .maplibregl-ctrl-icon {
+  mask-image: image-url("map-controls/zoomout.svg");
+  -webkit-mask-image: image-url("map-controls/zoomout.svg");
+}
+
+.maplibregl-ctrl button.maplibregl-ctrl-geolocate .maplibregl-ctrl-icon {
+  mask-image: image-url("map-controls/geolocate.svg");
+  -webkit-mask-image: image-url("map-controls/geolocate.svg");
+}
```
hlfan: https://github.com/openstreetmap/openstreetmap-website/blob/713dc3285a54ff20a5051c055706f0b3d355242e/app/views/dashboards/show.html.erb#L5

Now the icons are requested twice each.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -0,0 +1,62 @@
...
+
+.maplibregl-ctrl button .maplibregl-ctrl-icon {
+  background-image: none !important;
+  background-color: white;
+}
+
+.maplibregl-ctrl button.maplibregl-ctrl-zoom-in .maplibregl-ctrl-icon {
+  mask-image: image-url("map-controls/zoomin.svg");
+  -webkit-mask-image: image-url("map-controls/zoomin.svg");
```
CommanderStorm: the `-webkit` aliase are here to support Opera and UC (sigh..)

https://caniuse.com/mdn-css_properties_mask-image

tomhughes: There's no need to add them manually though as we include `autoprefixer-rails` in the asset pipeline to generate any necessary prefixes.

app/assets/javascripts/dashboard.js
```diff
@@ -1,32 +1,57 @@
...
+        name: "OSM Raster",
+        sources: {
+          osm: {
+            type: "raster",
+            tiles: [
+              "https://tile.openstreetmap.org/{z}/{x}/{y}.png"
+            ],
+            tileSize: 256,
+            maxzoom: 19,
+            attribution: "© OpenStreetMap contributors"
```
1ec5: Do we need to set this property if `attributionControl` is going to be `false` anyways?

app/assets/javascripts/dashboard.js
```diff
@@ -1,32 +1,57 @@
...
   const defaultHomeZoom = 12;
   let map;
 
   if ($("#map").length) {
-    map = L.map("map", {
+    map = new maplibregl.Map({
+      container: "map",
+      style: {
+        version: 8,
+        name: "OSM Raster",
```
1ec5: This property is [optional](https://maplibre.org/maplibre-style-spec/root/#name). If anything surfaces it to the user somehow, we’d need to make it localizable, but I think it would be easier to omit it.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -0,0 +1,59 @@
...
+    border-start-start-radius: 4px;
+  }
+
+  &:last-child {
+    border-end-start-radius: 4px;
+  }
+}
+
+.maplibregl-user-location-accuracy-circle {
+  background-color: $green !important;
```
1ec5, line 42: This is a nice touch. It’s inconsistent with the accuracy circle’s appearance on the homepage, but I suspect most users won’t notice the inconsistency. (For that matter, I had forgotten we even had a geolocation control on this map.)

1ec5 (approved):

app/assets/javascripts/dashboard.js
CommanderStorm: Does not matter, just giving it nicer debug info.

```suggestion
```

app/assets/javascripts/dashboard.js
CommanderStorm: I guess for now this can be removed, but it will be re-added in the next PR where this is generalised.

I don't think this matters.

```suggestion
```

pablobm (approved):
I think this is good 👍 I don't know MapLibre and I might be missing stuff, but the point here is to deploy somewhere harmless and see what happens, so let's do just that.

app/assets/javascripts/dashboard.js
1ec5: Agreed, though when that happens, we’ll need to make it localizable anyhow. We already have a similar string for the Leaflet integration.

.github/workflows/tests.yml
```diff
@@ -19,13 +19,20 @@ jobs:
...
     - name: Setup ruby
-      uses: ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71 # v1.268.0
+      uses: ruby/setup-ruby@d5126b9b3579e429dd52e51e68624dda2e05be25 # v1.267.0
       with:
         ruby-version: ${{ matrix.ruby }}
         rubygems: 3.4.10
         bundler-cache: true
+    - name: Enable virtual display & WebGL
+      run: |
+        sudo apt-get install -y xvfb mesa-utils libgl1-mesa-dri
```
tomhughes: This should just be added to the existing package install rather than running `apt-get` twice.

.github/workflows/tests.yml
```diff
@@ -19,13 +19,20 @@ jobs:
...
+        Xvfb :99 -screen 0 1024x768x24 &
+        export DISPLAY=:99
+        echo "Using virtual display $DISPLAY"
+        glxinfo | grep "OpenGL renderer" || echo "GL info not available"
```
tomhughes: These three lines are just debugging and should just be removed now you've got it working.

This step, which will then be just the one line to start `Xvfb` should be moved down just before the test steps that use it, and after the package install.

tomhughes (changes requested):
In addition to the inline comments, please reorder the commits so we have Xvfb support before the commit where we start needed it.

tomhughes (changes requested):
This needs to add `tile.openstreetmap.org` to the `connect_src` rule in `config/initializers/content_security_policy.rb` as maplibre-gl fetches the tiles with XHR rather than as images.

Personally I'd suggest removing the 1 pixel border between controls in a group so the the zoom in/out don't have a gap between them like the current map but I'm open to argument on that.

Whether to put the geolocate control in the same group (if that's even possible?) is a more interesting question - it looks OK separate here but it might become more of an issue when we get to the main page where there are more controls?

Finally, could we tone down the colour of the marker a bit? To get it a bit closer to the current one - it's not so much that I want it to be identical but the new one is very bright and a bit in your face?

CommanderStorm:
> Personally I'd suggest removing the 1 pixel border between controls in a group so the the zoom in/out don't have a gap between them like the current map but I'm open to argument on that.

<details><summary>very simple to do (click to open)</summary>
<p>

```diff
diff --git i/app/assets/stylesheets/maplibre-gl-all.scss w/app/assets/stylesheets/maplibre-gl-all.scss
index c1426ef5e..3d10304aa 100644
--- i/app/assets/stylesheets/maplibre-gl-all.scss
+++ w/app/assets/stylesheets/maplibre-gl-all.scss
@@ -42,6 +42,10 @@
   background-color: white;
 }
 
+.maplibregl-ctrl-group button+button {
+  border-top: 0;
+}
+
 .maplibregl-ctrl button.maplibregl-ctrl-zoom-in .maplibregl-ctrl-icon {
   mask-image: image-url("map-controls/zoomin.svg");
 }
```

</p>
</details> 

Same as the choice of icon for the markers though this did not seem like a concious decision.
I think following the general trend and having buttons be a bit separated makes sense, but if you want I can dd the 4 lines.

I think it makes is simpler for people hard of seeing to differentiate buttons, but then I have 10/10 vision, so don't actually know what it is like.

> could we tone down the colour of the marker a bit? To get it a bit closer to the current one - it's not so much that I want it to be identical but the new one is very bright and a bit in your face?

Obviously, I could make the marker look the same way leaflet defaults.

Thing is: Do you actually want this?
For this feature, being obvious where something is locaed is kind of the point. In the end we don't want to hide this information.
(More contrast is also more accessible, not a big problem here though 😉)

| default maplibre | default leaflet |
|--|--|
| image | image |

> Whether to put the geolocate control in the same group (if that's even possible?) is a more interesting question - it looks OK separate here but it might become more of an issue when we get to the main page where there are more controls?

It is *possible* to build a custom combined control, but that means more code, and given the feedback on keeping this PR minimal, I’m *guessing* that’s not preferred.

tomhughes:
Let's give it a day or two to see what other people think... So people can see what we're talking about here's the old version:

image 

and the new:

image

1ec5:
I figure that we’ll eventually want consistency between this marker and the markers on maps elsewhere on the site. If we’re OK with the blue markers elsewhere being brighter, then these red and green ones can be brighter too. On the other hand, there’s the rainbow colors when we plot search results on the map – some of those are quite bright, so there’s precedent for either appearance.

hlfan:
Is there an option to use the common marker with the non-rounded tip with the oval shadow? That feels to me like the most precise presentation.

pablobm:
To me the colour of the marker is significantly brighter than the existing. To compare with the search results:

A few markers for search results, in a variety of colours

Disclaimer: colour theory is not my forte, and my main screen is a bit ancient. With that in mind, what I do perceive is any of those colours as toned down in comparison to the new red.

I don't think I mind if the new selection of colours was brighter overall, as long as in the long term we have consistency. If the move to MapLibre comes with a new choice of colours, while there's a temporary inconsistency as MapLibre is applied throughout the site, then that's ok with me.

Could this be used as an opportunity when presenting the library transition to the community? Mappers can be let know that the maps with brighter markers are the ones with MapLibre, and build some sort of anticipation for the change to come. (Assuming it doesn't take years!).

gravitystorm:
> Let's give it a day or two to see what other people think... So people can see what we're talking about here's the old version:

* There's an off-by-one error on the zoom level shown. This is a common problem when converting things to use vector maps, since Maplibre etc have a different idea of what counts as zoom level X.
* The map marker should be changed to match the original leaflet marker. This is because we have other marker (e.g. notes, search results) that are based on that, and so this map should retain consistency.
* I'd like to see the button groups stay the same as other maps, again for consistency.

> Could this be used as an opportunity when presenting the library transition to the community? Mappers can be let know that the maps with brighter markers are the ones with MapLibre, and build some sort of anticipation for the change to come. 

I don't think this is worth doing. Let's wait until there's something user-interesting (e.g. continuous zooms) before trying to make anyone aware of what's going on.

> (Assuming it doesn't take years!).

It was my intention that something relatively simple like this would be the first PR to be merged / reviewed, but my hope was also that someone would made a reasonable stab at ensuring the more difficult things are actually possible before we start down this route! If so, it shouldn't take too long, but there's always that risk.

hlfan:
> With that in mind, what I do perceive is any of those colours as toned down in comparison to the new red.

And also in comparison to the colored search result border. https://github.com/openstreetmap/openstreetmap-website/pull/6235#issuecomment-3153800840
That's because the pin fill is behind a partially transparent gradient, so it doesn't look as flat as the maplibre marker.

CommanderStorm:
here is the updated screenshot.

image

I don't quite like that hack that is used for leaflet-markers in maplibre, but that is the sma ehack that is used for leaflet markers in leaflet, so I guess this is the "accepted" way.

CommanderStorm:
Now the only thing that is different is that the corners are not edgy, but I consider this a bug in the previous impl.

app/assets/javascripts/maplibre.combinedcontrolgroup.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable no-unused-vars */
```
CommanderStorm: not sure if this is THE correct way to handle this.
Eslint does not seem to know about the the (a bit alien to me) `=require` ruby style imports.

There are other occurances of this pattern, but not as many as I would expect, so not sure if this is correct.

app/assets/javascripts/maplibre.map.js
```diff
@@ -0,0 +1,58 @@
...
+      id: "osm",
+      type: "raster",
+      source: "osm"
+    }
+  ]
+};
+
+// Helper function to create Leaflet style (SVG comes from Leaflet) markers for MapLibre
+// new maplibregl.Marker({ color: color }) is simpler, but does not have the exact same gradient
+OSM.createMaplibreMarker = function ({ icon = "dot", color = "var(--marker-red)", ...options }) {
```
CommanderStorm: this is the analogue to `OSM.getMarker` which is how to create a leaflet marker.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -0,0 +1,67 @@
...
+    background-color: black !important;
+  }
+
+  &:first-child {
+    border-start-start-radius: 4px;
+  }
+
+  &:last-child {
+    border-end-start-radius: 4px;
+  }
```
hlfan: This needs to override the `button:focus` state as well:

app/assets/javascripts/maplibre.map.js
```diff
@@ -0,0 +1,58 @@
...
+  svg.setAttribute("height", "40");
+  svg.classList.add("pe-none");
+  svg.style.overflow = "visible";
+
+  // Use the marker icons from NoteMarker.svg
+  const use1 = document.createElementNS("http://www.w3.org/2000/svg", "use");
+  use1.setAttributeNS("http://www.w3.org/1999/xlink", "xlink:href", "#pin-shadow");
+
+  const use2 = document.createElementNS("http://www.w3.org/2000/svg", "use");
+  use2.setAttributeNS("http://www.w3.org/1999/xlink", "xlink:href", `#pin-${icon}`);
```
hlfan:
```suggestion
  const use1 = document.createElementNS("http://www.w3.org/2000/svg", "use");
  use1.setAttribute("href", "#pin-shadow");

  const use2 = document.createElementNS("http://www.w3.org/2000/svg", "use");
  use2.setAttribute("href", `#pin-${icon}`);
```

Others might further suggest using more jQuery here, but since there won't be a shorthand for `document.createElementNS` without Leaflet, I'd leave it as is.

app/assets/javascripts/dashboard.js
```diff
@@ -1,32 +1,41 @@
...
+    });
+    map.addControl(new OSM.CombinedControlGroup([navigationControl, geolocateControl]), position);
 
     $("[data-user]").each(function () {
       const user = $(this).data("user");
       if (user.lon && user.lat) {
-        L.marker([user.lat, user.lon], { icon: OSM.getMarker({ color: user.color }) }).addTo(map)
-          .bindPopup(user.description, { minWidth: 200 });
+        const popup = new maplibregl.Popup()
+          .setHTML(user.description);
```
hlfan: Right now, this makes the pop-up tip cover the pin. In the Leaflet map, the pop-up is positioned higher, 10px above the center of the white pin icon.

app/assets/javascripts/maplibre.combinedcontrolgroup.js
```diff
@@ -0,0 +1,38 @@
...
+
+    return this._container;
+  }
+
+  onRemove() {
+    for (const ctrl of this.controls) {
+      if (typeof ctrl.onRemove === "function") {
+        ctrl.onRemove();
+      }
+    }
```
hlfan:
```suggestion
    for (const ctrl of this.controls) ctrl.onRemove?.();
```

app/assets/javascripts/maplibre.i18n.js
hlfan, line 1: Eventually, we'd probably wanna overwrite `map._getUIString`, but I guess this is fine for now.

app/assets/stylesheets/common.scss
```diff
@@ -419,6 +419,11 @@ body.small-nav {
   border-top: 0px !important;
 }
 
+.maplibregl-popup-tip {
+  border-top-color: rgb(var(--bs-body-bg-rgb)) !important;
+}
```
hlfan: Since maplibre tries to keep the pop-up in the map window, all directions are necessary:

```suggestion
@each $anchor, $border in (
  left: right,
  right: left,
  top: bottom,
  bottom: top
) {
  .maplibregl-popup#{'[class*="anchor-#{$anchor}"]'} .maplibregl-popup-tip {
    border-#{$border}-color: var(--bs-body-bg);
  }
}
```

app/assets/javascripts/dashboard.js
CommanderStorm: Good catch.
When dragged to the corners, this now looks like this

app/assets/javascripts/maplibre.map.js
```diff
@@ -0,0 +1,77 @@
...
+    'bottom-right': [0, -45],
+    'top': [0, 5],
+    'top-left': [0, 5],
+    'top-right': [0, 5],
+    // our marker is bigger at the top, but this does not attach there -> tucked 2px more
+    'right': [-15, -10],
+    'left': [15, -10]
+  };
+  return new maplibregl.Popup({ offset }).setHTML(content);
+}
```
hlfan:
```suggestion
  const offset = {
    "bottom": [0, -45],
    "bottom-left": [0, -45],
    "bottom-right": [0, -45],
    "top": [0, 5],
    "top-left": [0, 5],
    "top-right": [0, 5],
    // our marker is bigger at the top, but this does not attach there -> tucked 2px more
    "right": [-15, -10],
    "left": [15, -10]
  };
  return new maplibregl.Popup({ offset }).setHTML(content);
};
```

app/assets/javascripts/maplibre.map.js
```diff
@@ -0,0 +1,77 @@
+//= require maplibre-gl/dist/maplibre-gl
+
+OSM.Mapnik = {
```
tomhughes: Do we maybe want to put this in a sub-namespace of OSM in anticipation of having additional layers there?

Maybe even `OSM.Maplibre` that this file can own and which contains `Styles.Mapnik` and `getMarker` and `getPopup` functions and the `Locale` and `CombinedControlGroup` names as well?

app/assets/javascripts/maplibre.map.js
```diff
@@ -0,0 +1,77 @@
...
+  version: 8,
```
tomhughes, line 4: Is there some significance to that choice of version?

1ec5: MapLibre only supports version 8 of the MapLibre Style Specification. It’s an error not to specify version 8. Mapbox held out some hope of a backwards incompatible version 9 at some point (mapbox/mapbox-gl-js#6584), but it’s a remote possibility even for MapLibre because of all the tooling that would have to change.

tomhughes: Oh sorry I thought that was a version number for this style, not for the style syntax ;-)

app/assets/javascripts/maplibre.map.js
CommanderStorm: It is not pretty or straigtforward (inconsistent with `OSM.getMarker`, ...), but if you want to have it this way, shure

tomhughes: It was a suggestion but if there's some reason why you think it's harder than I thought it would be then feel free to say so.

I don't disagree that it may be different - was just trying to see if we could do things better.

hlfan: It should probably be namespaced. If you don't like the inconsistency you can put getMarker & co. into L.OSM (which could already have been done when replacing OSM.getUserIcon).

CommanderStorm: Lets not waste more time on the leaflet code than nessesary.

Gettting reviews is already extremely hard

app/assets/javascripts/maplibre.combinedcontrolgroup.js
hlfan, line 1: Can this be wrapped to not pollute the global scope?

app/assets/javascripts/osm.js.erb
```diff
@@ -228,5 +228,9 @@ OSM = {
     modalBody.textContent = message;
     const alertModal = new bootstrap.Modal(document.getElementById("osm_alert_modal"));
     alertModal.show();
+  },
+
+  MapLibre: {
+    Styles: {}
```
hlfan, line 234: Could this also use PascalCase only for instantiable objects?

CommanderStorm: this is the naming-change that was requested in https://github.com/openstreetmap/openstreetmap-website/pull/6504#discussion_r2611316618 by @tomhughes 

I won't go against mainteiner advice on this one.

app/assets/javascripts/maplibre.combinedcontrolgroup.js
CommanderStorm: The way how I did this is how it is done in other places such as

https://github.com/openstreetmap/openstreetmap-website/blob/2f01fd9596677fd750fbe34eaf07e47269a28f39/app/assets/javascripts/index/contextmenu.js#L104

If you can point me to how you want this done, I can do it.
Currently, this follows the established best pracitce, so I would like to not deviate from that.
All other class definitions are done this way.

app/assets/javascripts/osm.js.erb
hlfan: I think that was only about namespacing these things, the rest is cosmetic enough to dismiss here.

app/assets/javascripts/maplibre.combinedcontrolgroup.js
hlfan: Yeah I see the clean-up maintainer has been a bit inactive...
But then that can also wait and be tackled together.

tomhughes (approved):
Thanks for all the work on this and sorry if it's been a bit painful but I think this is good to merge now.

Hopefully it also provides us with a good base to build on for moving the rest of the site to MapLibre.

CommanderStorm:
🥳 being upfront: next PR in this series won't be before Christmas, but after

imagico:
Just a quick note for information:

What this change introduces is not considered a suitable application of OpenStreetMap Carto by the OSM-Carto designers.  Guidelines for this can be found on https://github.com/openstreetmap-carto/openstreetmap-carto/issues/5117.

This change does two things that - as described in the linked to document - we do not consider suitable applications of the style.  That is:

* display of a map based on the style at lower than native resolution.
* using raster post processing after rendering in the form of resampling.

To be clear: This has no immediate practical effect - OpenStreetMap Carto can be freely used by anyone even for applications that the OSM-Carto designers do not consider suitable uses of the style.  And, since you don't credit OpenStreetMap Carto anyway, there is no point in asking not to be associated with this use.  And finally - this dashboard has never been a significant use case for OpenStreetMap Carto style development, so this change is not going to affect OSM-Carto style development in a measurable way.

Bottom line: OpenStreetMap Carto does not really take substantial issue with this change, but i thought it is prudent to point out that our work is neither intended to be used this way nor do we consider it a suitable application to use it this way.

tomhughes:
Is the resampling just because it's allowing zoom beyond z19? I thought we didn't have variable zoom enabled yet so I wouldn't expect any resampling otherwise?

hlfan:
With zooming via touch, the map does not snap to zoom levels. But the resampling issue there is just an addition to not having retina tiles.

imagico:
Overzooming beyond z19 is not an issue - that is a natural way to display tile services that only go to z19 on sites that go beyond z19 in their map display otherwise.  We have that in plenty of use cases of OSM-Carto tiles in the OSM community.  Overpass turbo might be the most prominent one.

What i see (desktop, chromium) on the dashboard is this:

dashboard

compared to native rendering:

native

1ec5:
The map on the dashboard now defaults to z11, which [looks the same](https://wiki.openstreetmap.org/wiki/Zoom_levels#Mapbox_GL) to me as z12 on the homepage. @imagico, did you zoom in or out before taking the screenshot (using either a scroll wheel or two-finger trackpad gesture, as opposed to the ± buttons)? I think this might be an artifact of smooth zooming (fractional zooming). Leaflet has this capability too, but we never enabled it on this site.

By default, both libraries assume a raster tile layer is showing raster data (i.e., aerial or satellite imagery), so they resolve the `zoom` in the tile URL by rounding the current zoom level. This ensures the tiles are never over- or underzoomed by more than half a zoom level. The alternative is to disable the [`roundZoom`](https://maplibre.org/maplibre-gl-js/docs/API/classes/RasterTileSource/#roundzoom) option, which would floor the zoom level, so the tiles will be overzoomed by as much as one zoom level unless you set the zoom level precisely to an integer using the ± buttons. There’s a diagram of the tradeoffs in https://github.com/mapbox/mapbox-gl-js/issues/1030#issuecomment-76043779.

The other alternative would be to disable smooth zooming when viewing raster tiles. MapLibre doesn’t have a built-in option for that, but there’s a workaround in https://github.com/maplibre/maplibre-gl-js/issues/1591#issuecomment-2748319483. This could probably also be used to force a ceiling zoom strategy if desired.

imagico:
The shown screenshot can be obtained both through zooming in or zooming out, it is clearly an intended result, not a rare freak result after a specific sequence of unusual operations.  If that map was not meant to be zoomed then it seems odd to me that it can be zoomed.

I don't really want to and neither am i qualified to participate in a discussion on *how* to properly display OSM-Carto tiles on a website.  I am pretty confident that all decent quality frameworks for displaying tiled maps are capable of interactively displaying OSM-Carto in a suitable way - there are literally thousands of community websites in the OSM world that do so.  But - if that is somehow not deemed possible/desirable here, then it would likewise be an option not to use OSM-Carto on the dashboard of course.

1ec5:
> The shown screenshot can be obtained both through zooming in or zooming out, it is clearly an intended result, not a rare freak result after a specific sequence of unusual operations. If that map was not meant to be zoomed then it seems odd to me that it can be zoomed.

That wasn’t my suggestion. I was just confirming the steps to reproduce what you were seeing, before segueing into a discussion about how to make it better. Now that we have enough information, I’ve opened #6618 to track your feedback. Note that this isn’t just about OSM Carto; in the context of #6234, other raster featured layers would also be affected. The point of starting with the dashboard map was precisely to identify potential issues before they come up in retrofitting more prominent maps elsewhere on the site.

PR #6505, opened 2025-11-06
Add db file diff when running migrations

CommanderStorm:
### Description

When running the migrations [as documented here](https://github.com/openstreetmap/openstreetmap-website/blob/master/DOCKER.md), these file changes appear.
I have no clue what they mean or what `consider_pd` is and why the default gets auto-generated to a different value.

### How has this been tested?

The site works if I run it manually? Not sure where this diff is coming from as there is quite a lot of "things" going on in the website and the underlying architecture is not yet quite clear to me.


CommanderStorm:
does not seem to be reproducible in CI. Weird.

tomhughes:
It's entirely dependent on the history of your local database so please don't worry about changes to that file unless you've actually added new migrations.

PR #6518, opened 2025-11-11
Add URL prettification for plain texts

### LINKED ISSUE #5780: pretty formatting of OSM URLS in notes, changeset comment, and comments

hlfan:
This mostly resolves #5780 by extending `linkify` to recognize and normalize common OSM and wiki references, automatically linking them to their canonical URLs.
It keeps the general URL-shortening improvements from #5844 and, as a successor to #5861, ensures that links remain stable and valid even when the text is reprocessed without rich text links.

The new `linkify` flow is now a three-stage pipeline:
#### Understand the reference 🧭
`expand_link_shorthands` pattern-matches common reference shorthands and expands them into full canonical URLs.
#### Add a hyperlink 🔗
This step, done by `Rinku.auto_link`, remains unchanged. It detects URLs and wraps them in `<a>` tags.
#### Make it look pretty 💅
`format_link_text` builds on #5844 and extends it to shorten link texts that match predefined patterns further. The `href` always remains the canonical URL that Rinku found.

---

The pre-processing and post-processing stages use a centralized rulebook defined in `Settings.linkify`. #5862 could be adapted to use this block as well.

I turned the table from https://github.com/openstreetmap/openstreetmap-website/issues/5780#issue-2905009904 into an end-to-end `test_text_to_html_linkify_openstreetmap_links`, dropping only a few enhancements not yet implemented.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -177,6 +177,15 @@ linkify:
       - "(?<key>[^\"?#<>/\\s]+)=(?<value>[^\"?#<>\\s]+)"
       path_template: "wiki/Tag:\\k<key>=\\k<value>"
       host: "https://wiki.openstreetmap.org"
+  display_rules:
+    - pattern: "osm\\.org/user/"
+      replacement: "@"
+    - pattern: "osm\\.org/(?<type>node|way|relation|changeset|note)/"
+      replacement: "\\k<type>/"
+    - pattern: "osm\\.wiki/Key:(?<key>[^\"?#<>/\\s]+)"
+      replacement: "\\k<key>=*"
+    - pattern: "osm\\.wiki/Tag:(?<key>[^\"?#<>/\\s]+)(?:=|%3D)(?<value>[^\"?#<>\\s]+)"
+      replacement: "\\k<key>=\\k<value>"
 # Main website hosts to match in linkify
 linkify_hosts: ["www.openstreetmap.org", "www.osm.org", "www.openstreetmap.com", "openstreetmap.org", "osm.org", "openstreetmap.com"]
 # Shorter host to replace main hosts
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -139,6 +139,44 @@ fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
 # Endpoints for Wikimedia integration
 wikidata_api_url: "https://www.wikidata.org/w/api.php"
 wikimedia_commons_url: "https://commons.wikimedia.org/wiki/"
+linkify:
+  detection_rules:
+    - patterns:
+      - "node/(?<id>\\d+)"
+      - "node (?<id>\\d{5,})"
+      - "n ?(?<id>\\d{5,})"
+      path_template: "node/\\k<id>"
+    - patterns:
+      - "way/(?<id>\\d+)"
+      - "way (?<id>\\d{5,})"
+      - "w ?(?<id>\\d{5,})"
+      path_template: "way/\\k<id>"
+    - patterns:
+      - "relation/(?<id>\\d+)"
+      - "relation (?<id>\\d{5,})"
+      - "r ?(?<id>\\d{5,})"
+      path_template: "relation/\\k<id>"
+    - patterns:
+      - "changeset/(?<id>\\d+)"
+      - "changeset (?<id>\\d{5,})"
+      - "cs ?(?<id>\\d{5,})"
+      path_template: "changeset/\\k<id>"
+    - patterns:
+      - "note/(?<id>\\d+)"
+      - "note (?<id>\\d{5,})"
+      path_template: "note/\\k<id>"
+    - patterns:
+      - "user/(?<username>\\S+)"
+      - "@(?<username>\\S+)"
+      path_template: "user/\\k<username>"
+    - patterns:
+      - "(?<key>[^\"?#<>/\\s]+)=\\*"
+      path_template: "wiki/Key:\\k<key>"
+      host: "https://wiki.openstreetmap.org"
+    - patterns:
+      - "(?<key>[^\"?#<>/\\s]+)=(?<value>[^\"?#<>\\s]+)"
+      path_template: "wiki/Tag:\\k<key>=\\k<value>"
+      host: "https://wiki.openstreetmap.org"
 # Main website hosts to match in linkify
 linkify_hosts: ["www.openstreetmap.org", "www.osm.org", "www.openstreetmap.com", "openstreetmap.org", "osm.org", "openstreetmap.com"]
 # Shorter host to replace main hosts
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -89,12 +89,30 @@ module RichText
 
     def linkify(text, mode = :urls)
       ERB::Util.html_escape(text)
+               .then { |html| expand_link_shorthands(html) }
                .then { |html| expand_host_shorthands(html) }
                .then { |html| auto_link(html, mode) }
                .html_safe
     end
 
     private
+
+    def gsub_pairs_for_linkify_detection
+      Array
+        .wrap(Settings.linkify&.detection_rules)
+        .select { |rule| rule.path_template && rule.patterns.is_a?(Array) }
+        .flat_map do |rule|
+          expanded_path = "#{rule.host || "#{Settings.server_protocol}://#{Settings.server_url}"}/#{rule.path_template}"
+          rule.patterns
+              .select { |pattern| pattern.is_a?(String) }
+              .map { |pattern| [Regexp.new("(?<=^|#{URL_UNSAFE_CHARS})#{pattern}", Regexp::IGNORECASE), expanded_path] }
+        end
+    end
+
+    def expand_link_shorthands(text)
+      gsub_pairs_for_linkify_detection
+        .reduce(text) { |text, (pattern, replacement)| text.gsub(pattern, replacement) }
+    end
 
     def expand_host_shorthands(text)
       [
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -134,9 +134,12 @@ module RichText
 
     def format_link_text(url)
       url = shorten_host(url, Settings.linkify_hosts, Settings.linkify_hosts_replacement)
-      shorten_host(url, Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement) do |path|
+      url = shorten_host(url, Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement) do |path|
         path.sub(Regexp.new(Settings.linkify_wiki_optional_path_prefix || ""), "")
       end
+      Array.wrap(Settings.linkify&.display_rules)
+           .select { |rule| rule.pattern && rule.replacement }
+           .reduce(url) { |url, rule| url.sub(Regexp.new(rule.pattern), rule.replacement) }
     end
 
     def shorten_host(url, hosts, hosts_replacement)
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -87,16 +87,24 @@ module RichText
     end
 
     def linkify(text, mode = :urls)
-      link_attr = 'rel="nofollow noopener noreferrer" dir="auto"'
-      Rinku.auto_link(ERB::Util.html_escape(text), mode, link_attr) do |url|
-        url = shorten_host(url, Settings.linkify_hosts, Settings.linkify_hosts_replacement)
-        shorten_host(url, Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement) do |path|
-          path.sub(Regexp.new(Settings.linkify_wiki_optional_path_prefix || ""), "")
-        end
-      end.html_safe
+      ERB::Util.html_escape(text)
+               .then { |html| auto_link(html, mode) }
+               .html_safe
     end
 
     private
+
+    def auto_link(text, mode)
+      link_attr = 'rel="nofollow noopener noreferrer" dir="auto"'
+      Rinku.auto_link(text, mode, link_attr) { |url| format_link_text(url) }
+    end
+
+    def format_link_text(url)
+      url = shorten_host(url, Settings.linkify_hosts, Settings.linkify_hosts_replacement)
+      shorten_host(url, Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement) do |path|
+        path.sub(Regexp.new(Settings.linkify_wiki_optional_path_prefix || ""), "")
+      end
+    end
 
     def shorten_host(url, hosts, hosts_replacement)
       %r{^(https?://([^/]*))(.*)$}.match(url) do |m|
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -3,6 +3,7 @@
 module RichText
   DESCRIPTION_MAX_LENGTH = 500
   DESCRIPTION_WORD_BREAK_THRESHOLD_LENGTH = 450
+  URL_UNSAFE_CHARS = "[^\\w!#$%&'*+,./:;=?@_~^\\-]"
 
   def self.new(format, text)
     case format
@@ -88,11 +89,25 @@ module RichText
 
     def linkify(text, mode = :urls)
       ERB::Util.html_escape(text)
+               .then { |html| expand_host_shorthands(html) }
                .then { |html| auto_link(html, mode) }
                .html_safe
     end
 
     private
+
+    def expand_host_shorthands(text)
+      [
+        [Settings.linkify_hosts, Settings.linkify_hosts_replacement],
+        [Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement]
+      ]
+        .select { |hosts, replacement| replacement && hosts&.any? }
+        .reduce(text) do |text, (hosts, replacement)|
+          text.gsub(/(?<=^|#{URL_UNSAFE_CHARS})\b#{Regexp.escape(replacement)}/) do
+            "#{Settings.server_protocol}://#{hosts[0]}"
+          end
+        end
+    end
 
     def auto_link(text, mode)
       link_attr = 'rel="nofollow noopener noreferrer" dir="auto"'
--- a/test/lib/rich_text_test.rb
+++ b/test/lib/rich_text_test.rb
@@ -352,6 +352,55 @@ class RichTextTest < ActiveSupport::TestCase
     end
   end
 
+  def test_text_to_html_linkify_recognize_path
+    with_settings(:linkify => { :detection_rules => [{ :patterns => ["@(?<username>\\w+)"], :path_template => "user/\\k<username>" }] }) do
+      r = RichText.new("text", "foo @example bar")
+      assert_html r do
+        assert_dom "a", :count => 1, :text => "http://test.host/user/example" do
+          assert_dom "> @href", "http://test.host/user/example"
+          assert_dom "> @rel", "nofollow noopener noreferrer"
+        end
+      end
+    end
+  end
+
+  def test_text_to_html_linkify_recognize_path_no_partial_match
+    with_settings(:linkify => { :detection_rules => [{ :patterns => ["@(?<username>\\w+)"], :path_template => "user/\\k<username>" }] }) do
+      r = RichText.new("text", "foo example@example.com bar")
+      assert_html r do
+        assert_select "a", 0
+      end
+    end
+  end
+
+  def test_text_to_html_linkify_recognize_wiki_path
+    with_settings(:linkify => { :detection_rules => [{ :patterns => ["(?<key>[^\"?#<>/\\s]+)=(?<value>[^\"?#<>\\s]+)"], :path_template => "Tag:\\k<key>=\\k<value>", :host => "http://example.wiki" }] }) do
+      r = RichText.new("text", "foo surface=metal bar")
+      assert_html r do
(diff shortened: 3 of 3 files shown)


pablobm:
Looking good :eyes: How about a short test that explicitly shows that the reason for 5 digits minimum is to avoid expanding years accidentally? Something like `test_years_not_expanded`. If in the future someone experiments with that number, the test failure should be explanatory.

lib/rich_text.rb
```diff
@@ -118,16 +118,36 @@ def sanitize(text)
...
-        end
+      Rinku.auto_link(unshorten_links(ERB::Util.html_escape(text)), mode, link_attr) do |url|
+        shorten_link(url)
       end.html_safe
     end
 
     private
 
+    def unshorten_links(text)
+      Settings.linkify.paths.each do |path_rule|
```
pablobm: Probably worth providing for the case where this setting is not present. This appears to be done for `Settings.linkify_hosts`, as `shorten_host` does `hosts&.include(host)`.

```suggestion
      linkify_paths = Array.wrap(Settings.linkify&.paths)
      linkify_paths.each do |path_rule|
```

lib/rich_text.rb
```diff
@@ -118,16 +118,36 @@ def sanitize(text)
...
+      end
+      text
+    end
+
+    def shorten_link(url)
+      url = shorten_host(url, Settings.linkify_hosts, Settings.linkify_hosts_replacement)
+      url = shorten_host(url, Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement) do |path|
+        path.sub(Regexp.new(Settings.linkify_wiki_optional_path_prefix || ""), "")
+      end
+      Settings.linkify.print.each do |print_rule|
```
pablobm: Similarly:

```suggestion
      linkify_print = Array.wrap(Settings.linkify&.print)
      linkify_print.each do |print_rule|
```

config/settings.yml
```diff
@@ -139,6 +139,35 @@ fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
...
+      replacement: "note/\\k<id>"
+    - list: ["user/(?<username>[^\\s]+)", "@(?<username>[^\\s]+)"]
+      replacement: "user/\\k<username>"
+    - list: ["(?<key>[^\"?#<>/\\s]+)=\\*?"]
+      replacement: "wiki/Key:\\k<key>"
+      host: "wiki.openstreetmap.org"
+    - list: ["(?<key>[^\"?#<>/\\s]+)=(?<value>[^\"?#<>\\s]+)"]
+      replacement: "wiki/Tag:\\k<key>=\\k<value>"
+      host: "wiki.openstreetmap.org"
+  print:
```
pablobm: I was a bit confused about the word `print` here, but now that I've looked into this in a bit more detail, I'm very confused 😅 with all this `shorten`, `unshorten`, `print` and `paths`.

Would you be able to explain the pipeline for me to understand? An example of how the transformations take place.

hlfan: I updated the PR description to describe the pipeline, but as an example, `way/123` first gets expanded to `https://www.openstreetmap.org/way/123`, which is then converted into a hyperlink by Rinku, and following that, the text inside the <a> is shortened back to `way/123`.

lib/rich_text.rb
```diff
@@ -118,16 +118,41 @@ def sanitize(text)
...
+        next unless rule.path_template && rule.patterns.is_a?(Array)
+
+        rule.patterns.each do |pattern|
+          expanded_path = "#{Settings.server_protocol}://#{rule.host || Settings.server_url}/#{rule.path_template}"
+          text.gsub!(Regexp.new("(?<before>\\s|^|>)#{pattern}(?<after>\\s|$|<)", Regexp::IGNORECASE), "\\k<before>#{expanded_path}\\k<after>")
+        end
+      end
+      [[Settings.linkify_hosts, Settings.linkify_hosts_replacement], [Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement]].each do |hosts, replacement|
+        text.gsub!(/(\s)#{Regexp.escape(replacement)}/, "\\1#{Settings.server_protocol}://#{hosts[0]}") if replacement && hosts&.any?
+      end
```
tomhughes: Can you explain the purpose of this? It comes from a commit that says "Fix copy-pasting breaking links in linkify" but what exactly is breaking the links and why does this fix it?

The PR talks about it rewriting the URL to use a canonical name but wasn't the short name (which you're rewriting from here) supposed to be the canonical name?

config/settings.yml
```diff
@@ -139,6 +139,35 @@ fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
...
 wikimedia_commons_url: "https://commons.wikimedia.org/wiki/"
+linkify:
+  detection_rules:
+    - patterns: ["node/(?<id>\\d+)", "node (?<id>\\d{5,})", "n(?<id>\\d+)"]
+      path_template: "node/\\k<id>"
+    - patterns: ["way/(?<id>\\d+)", "way (?<id>\\d{5,})", "w(?<id>\\d+)"]
+      path_template: "way/\\k<id>"
+    - patterns: ["relation/(?<id>\\d+)", "relation (?<id>\\d{5,})", "r(?<id>\\d+)"]
+      path_template: "relation/\\k<id>"
+    - patterns: ["changeset/(?<id>\\d+)", "changeset (?<id>\\d{5,})", "cs ?(?<id>\\d{5,})"]
```
tomhughes: Why do we allow a space after `cs` but not after object object types like `n` or `w`?

lib/rich_text.rb
```diff
@@ -118,16 +118,41 @@ def sanitize(text)
...
 
     private
 
+    def expand_link_shorthands(text)
+      linkify_detection_rules = Array.wrap(Settings.linkify&.detection_rules)
+      linkify_detection_rules.each do |rule|
+        next unless rule.path_template && rule.patterns.is_a?(Array)
+
+        rule.patterns.each do |pattern|
+          expanded_path = "#{Settings.server_protocol}://#{rule.host || Settings.server_url}/#{rule.path_template}"
```
tomhughes: It makes sense to combine `Settings.server_protocol` and `Settings.server_url` but there's no logical reason it's right combine it with `rule.host`? Should the rules have a base URL rather than just a host?

lib/rich_text.rb
hlfan: See https://github.com/openstreetmap/openstreetmap-website/issues/5780#issuecomment-2760413875 and onward for the problem.

By canonical I mean the URL that is the target of the redirects which is on the `www.openstreetmap.org` domain. The path is still canonical on the shortened domains.

hlfan: I just stumbled over this again here:

https://www.openstreetmap.org/user_blocks/18933
https://www.openstreetmap.org/user_blocks/18934

The first description has the import guidelines linked.
The second description, which was presumably copied and pasted, doesn't have that link.

pablobm:
Thank you, that explanation and the changes to the config are helpful. I had a more in-depth look this morning, and I've been playing with how to make this better. The code is difficult to follow, which is common with this sort of thing, but it's not unavoidable. For example, I rewrote the `linkify` method to show the steps more clearly, and moved the `Rinku.autolink_block` into a separate method. See: https://github.com/hlfan/openstreetmap-website/compare/prettify-urls...pablobm:openstreetmap-website:prettify-urls (Feel free to take from there).

Would you be able to do something similar with `expand_link_shorthands` and `format_link_text`? They are not the easiest to read currently.

(Oh, and also I added a test for something that wasn't tested).

config/settings.yml
```diff
@@ -139,6 +139,35 @@ fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
 # Endpoints for Wikimedia integration
 wikidata_api_url: "https://www.wikidata.org/w/api.php"
 wikimedia_commons_url: "https://commons.wikimedia.org/wiki/"
+linkify:
+  detection_rules:
+    - patterns: ["node/(?<id>\\d+)", "node (?<id>\\d{5,})", "n ?(?<id>\\d{5,})"]
```
pablobm: I think these would be more readable if formatted as multiline YAML arrays:

```suggestion
    - patterns:
      - "node/(?<id>\\d+)"
      - "node (?<id>\\d{5,})"
      - "n ?(?<id>\\d{5,})"
```

lib/rich_text.rb
```diff
@@ -87,17 +87,60 @@ def sanitize(text)
...
     private
 
+    def expand_link_shorthands(text)
+      Array.wrap(Settings.linkify&.detection_rules)
+           .select { |rule| rule.path_template && rule.patterns.is_a?(Array) }
+           .flat_map { |rule| linkify_detection_rule_to_gsub(rule) }
+           .reduce(text) { |text, (pattern, replacement)| text.gsub(pattern, replacement) }
+    end
+
+    def linkify_detection_rule_to_gsub(rule)
```
pablobm: Right, so this turns one of `linkify.detection_rules` into a `(pattern, replacement)` pair. 

I had no idea of what the method name meant and had to read the whole thing 😄 I was reading it as "linkify X to Y", with "linkify" as a verb and "to" linking it to a result. Instead it should be read as "X to Y", with "linkify" as part of a noun that gets turned into a different noun.

Can we come up with a better name? How about: `parse_linkify_detection_rule_into_gsub_pair`? I think that clarifies the verb and the result. Open to other options!

lib/rich_text.rb
```diff
@@ -87,17 +87,60 @@ def sanitize(text)
...
+               .then { |html| auto_link(html, mode) }
+               .html_safe
     end
 
     private
 
+    def expand_link_shorthands(text)
+      Array.wrap(Settings.linkify&.detection_rules)
+           .select { |rule| rule.path_template && rule.patterns.is_a?(Array) }
+           .flat_map { |rule| linkify_detection_rule_to_gsub(rule) }
```
pablobm: Interestingly, this whole thing does not depend on the argument `text`. I would separate it into a different method. Something like this:


```ruby
    def self.gsub_patterns_for_linkify
      Array
        .wrap(Settings.linkify&.detection_rules)
        .select { |rule| rule.path_template && rule.patterns.is_a?(Array) }
        .flat_map { |rule| linkify_detection_rule_to_gsub(rule) }
     end
```

Perhaps should be memoised? It does feel like waste, having to do all this work every time. But memoisation can bring its own issues.

pablobm (commented):
Why, doesn't this read so much better! 🎉 

The good thing is that I can review the code more thoroughly.

The bad thing is that I find new details to comment on 😝

lib/rich_text.rb
hlfan: I thought about that too when I broke the flat_mapping out into `linkify_detection_rule_to_gsub(rule)` to make RuboCop happy. What issues were you thinking about for memoisation?

pablobm: The issue is with the state it creates (the class variable). It lingers on, sometimes in places where it's not expected.

A couple of examples that come to mind:
- State remains across tests, introducing flaky failures if not careful.
- Values generated in a non-threadsafe manner, causing unexpected behaviour.

In this case I think the tests are affected, as they set things up with `with_settings` and the memoization will not be reset across tests. In the past I have worked around that with an explicit `RichText.reset_linkify_settings!` (for example) called on `setup` of affected tests, but it adds complication.

As for the thread safety, building the list in an initializer might be enough.

hlfan: I see. Let's leave this implementation detail of `get_linkify_detection_gsub_patterns` for another time!

lib/rich_text.rb
tomhughes: > See [#5780 (comment)](https://github.com/openstreetmap/openstreetmap-website/issues/5780#issuecomment-2760413875) and onward for the problem.

That explains (or at least claims) that copy paste breaks something but I don't see any analysis there of how/why it breaks or why this code would fix the problem?

Basically I'm reviewing this again and I'm staring at the second commit and I can't see anything technically wrong with it but I still have no understanding of why it's needed.

config/settings.yml
```diff
@@ -139,6 +139,35 @@ fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
...
+      host: "https://wiki.openstreetmap.org"
+  display_rules:
+    - pattern: "osm\\.org/user/"
+      replacement: "@"
+    - pattern: "osm\\.org/(?<type>node|way|relation|changeset|note)/"
+      replacement: "\\k<type>/"
+    - pattern: "osm\\.wiki/Key:(?<key>[^\"?#<>/\\s]+)"
+      replacement: "\\k<key>=*"
+    - pattern: "osm\\.wiki/Tag:(?<key>[^\"?#<>/\\s]+)(?:=|%3D)(?<value>[^\"?#<>\\s]+)"
+      replacement: "\\k<key>=\\k<value>"
```
tomhughes, line 188: Would taginfo be a better target here? Many keys and/or key-value pairs won't actually have a wiki page?

lib/rich_text.rb
```diff
@@ -87,17 +87,60 @@ def sanitize(text)
...
+           .reduce(text) { |text, (pattern, replacement)| text.gsub(pattern, replacement) }
+    end
+
+    def linkify_detection_rule_to_gsub(rule)
+      expanded_path = "#{rule.host || "#{Settings.server_protocol}://#{Settings.server_url}"}/#{rule.path_template}"
+      rule.patterns
+          .select { |pattern| pattern.is_a?(String) }
+          .map do |pattern|
+        [
+          Regexp.new("(?<before>\\s|^|>)#{pattern}(?<after>\\s|$|<)", Regexp::IGNORECASE),
```
tomhughes: Would  looking for a word boundary with `\b` before and after work here and make this a bit simpler?

lib/rich_text.rb
tomhughes: Is `flat_map` correct here? We want a list of pattern and replacement pairs and `linkify_detection_rule_to_gsub` returns a pattern/replacement pair so running it on each rule will give us what we want and flattening will destroy that?

lib/rich_text.rb
hlfan: Rinku only adds links to URLs with a protocol or the www subdomain: https://github.com/vmg/rinku?tab=readme-ov-file#rinku-is-a-standalone-library

However, the text representation of the shortened links doesn't have either. So when the plain text output gets run through it again, the osm.org link doesn't get recognized anymore.

hlfan: I think molding the requirement in a test would clear things up, but I don't know how to translate...

lib/rich_text.rb
hlfan: flat_map turns a tuple[][] into a tuple[] here as it's called on the top array.

lib/rich_text.rb
tomhughes: Ah that explanation helps a lot to explain things!

config/settings.yml
hlfan: The bulk of the tagging documentation is in the wiki, and even from non-existing pages, you can reach the taginfo site in one click. The other way around takes longer and is less straightforward.
Also, how often do you discuss a key or tag without a wiki page on the OSM site?

lib/rich_text.rb
hlfan: `\b` only matches after the `@` for user links or before the `=*` for key links, which sadly makes it unusable here.
I still tried to make it more easily readable now.

config/settings.yml
1ec5: That reminds me, we’ve only customized the wiki’s 404 page to include the data item–powered infobox and taginfo box when the interface is in English and a few other languages. It needs more translations. Maybe we can follow up separately on the wiki.

lib/rich_text.rb
```diff
@@ -87,17 +88,65 @@ def sanitize(text)
...
+        .wrap(Settings.linkify&.detection_rules)
+        .select { |rule| rule.path_template && rule.patterns.is_a?(Array) }
+        .flat_map do |rule|
+          expanded_path = "#{rule.host || "#{Settings.server_protocol}://#{Settings.server_url}"}/#{rule.path_template}"
+          rule.patterns
+              .select { |pattern| pattern.is_a?(String) }
+              .map do |pattern|
+            [
+              Regexp.new("(?<before>^|#{URL_UNSAFE_CHARS})#{pattern}", Regexp::IGNORECASE),
+              "\\k<before>#{expanded_path}"
```
tomhughes: I think you could use a lookbehind assertion here and then you wouldn't need to capture the prefix and add it back, like this:

```suggestion
              Regexp.new("(?<=^|#{URL_UNSAFE_CHARS})#{pattern}", Regexp::IGNORECASE),
              expanded_path
```

lib/rich_text.rb
```diff
@@ -87,17 +88,65 @@ def sanitize(text)
...
+
+    def expand_host_shorthands(text)
+      [
+        [Settings.linkify_hosts, Settings.linkify_hosts_replacement],
+        [Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement]
+      ]
+        .select { |hosts, replacement| replacement && hosts&.any? }
+        .reduce(text) do |text, (hosts, replacement)|
+          text.gsub(/(^|#{URL_UNSAFE_CHARS})\b#{Regexp.escape(replacement)}/) do
+            "#{Regexp.last_match(1)}#{Settings.server_protocol}://#{hosts[0]}"
```
tomhughes: Another candidate for an assertion?

```suggestion
          text.gsub(/(?<=^|#{URL_UNSAFE_CHARS})\b#{Regexp.escape(replacement)}/) do
            "#{Settings.server_protocol}://#{hosts[0]}"
```

tomhughes:
Thanks for all the work. I think this looks good now.

mmd-osm:
On the dev instance, I've added a comment with "node 4361366951" (see [link](https://master.apis.dev.openstreetmap.org/api/0.6/changeset/514343?include_discussion)), which is rendered as ([link](https://master.apis.dev.openstreetmap.org/changeset/514343))

image


I didn't expect the full hostname to appear as text. Is this works as designed, some bug or a config issue on the dev instance?

1ec5:
In production, it would turn into “osm.org/node/4361366951” based on the `linkify_hosts_replacement` setting. Is the dev instance overriding the default?

https://github.com/openstreetmap/openstreetmap-website/blob/56b153f62b77543a24708e4958d6097e189d6eeb/config/settings.yml#L191-L192

mmd-osm:
I haven't followed the discussion in all detail, apologies if this has been discussed before. My expectation would be something along these lines: the text "node 4361366951" with a hyperlink taking me to the respective node. I don't even expect the short version "osm.org" to appear as text in this case.

For clarity, here's the rendered result and HTML side by side, according to my expectation:

image

1ec5:
You have a point. Analogously, we linkify user name and key=value references without formatting the link text as a URL fragment.

mmd-osm:
So this appears to be some sort of configuration issue: [this discussion](https://www.openstreetmap.org/changeset/101585268) on osm.org looks fine given the following [raw data](https://api.openstreetmap.org/api/0.6/changeset/101585268?include_discussion).

pablobm:
Is there a reason why this was implemented for `"text"` types, but not for `"markdown"`? Someone just asked me why this was not working in direct messages, and now I realise it only works (if I have this right) for note comments and changeset comments.

hlfan:
Funnily you already questioned this a while ago in https://github.com/openstreetmap/openstreetmap-website/pull/6674#pullrequestreview-3641374364. If you have new ideas, an issue for that might be a good stepping stone.

pablobm:
😳 

New issue created at https://github.com/openstreetmap/openstreetmap-website/issues/7122

PR #6519, opened 2025-11-11
Correct MIME type of exported directions GeoJSON

### LINKED ISSUE #6388: Directions export should use GeoJSON MIME type
#5694 added a link to download directions as GeoJSON. However, when I click this link, the browser gives it a .json extension, which is incompatible with some GeoJSON consumers. This appears to be because we’re setting the blob’s MIME type to JSON, which browsers map to the .json extension:

https://github.com/openstreetmap/openstreetmap-website/blob/c97fff3965bb12d3588e98656f21b15ae6a322db/app/assets/javascripts/index/directions-route-output.js#L133

The official MIME type for GeoJSON is `application/geo+json`, which results in a .geojson extension (at least in Firefox 143 for macOS). Migrating to that MIME type would help the user better keep track of what they’ve downloaded and use it more easily. The downside is that maybe fewer applications would recognize the .geojson extension, but these applications are less likely to do anything special for GeoJSON versus regular JSON.

Originally reported [on the forum](https://community.openstreetmap.org/t/suggestion-add-download-route-as-gpx-to-osm-org/129580/9).

1ec5:
When exporting directions as GeoJSON, create the file using a .geojson file extension based on the more specific MIME type for GeoJSON.

Fixes #6388.

### DIFF
--- a/app/assets/javascripts/index/directions-route-output.js
+++ b/app/assets/javascripts/index/directions-route-output.js
@@ -130,7 +130,7 @@ OSM.DirectionsRouteOutput = function (map) {
       writeSteps(route);
     });
 
-    const blob = new Blob([JSON.stringify(polyline.toGeoJSON())], { type: "application/json" });
+    const blob = new Blob([JSON.stringify(polyline.toGeoJSON())], { type: "application/geo+json" });
     URL.revokeObjectURL(downloadURL);
     downloadURL = URL.createObjectURL(blob);
     $("#directions_route_download").prop("href", downloadURL);


HolgerJeromin:
Perhaps we could create a `download` attibute with a nice name, too.

hlfan:
@HolgerJeromin that's already the case:

https://github.com/openstreetmap/openstreetmap-website/blob/9fb6b21aca893ce30b9c03271142c3ad71efdd09/app/views/directions/show.html.erb#L68

HolgerJeromin:
`<a id="directions_route_download" download="route" href="blob:..." style="">Route als GeoJSON herunterladen</a>`

IMO this would be much better to be
`<a id="directions_route_download" download="route.geojson" href="blob:..." style="">Route als GeoJSON herunterladen</a>`

tomhughes:
Looks good to me, thanks - we can always improve the filename in a separate PR.

1ec5:
Firefox automatically chooses a .geojson file name based on the MIME type. (That’s why it previously chose .json.) I assumed that’s the case in other browsers; @HolgerJeromin, are you seeing differently?

HolgerJeromin:
> Firefox automatically chooses a .geojson file name based on the MIME type. (That’s why it previously chose .json.) I assumed that’s the case in other browsers; @HolgerJeromin, are you seeing differently?

Vivaldi and Chrome on desktop (windows) and mobile save a file `route` without extension. 
So does firefox on windows

Issue #6509, opened 2025-11-08
Map zooms out to minimum level when closing Changeset History during zoom animation


daishu0000:
### How to reproduce the issue?

1. Go to [OpenStreetMap.org](https://www.openstreetmap.org/)

2. Click "history" button, open changeset history (for example: https://www.openstreetmap.org/history#map=11/31.1179/121.6056)

3. With the Changeset History panel open, start zooming the map (using the mouse wheel)

4. Before the zoom animation completes, close the Changeset History panel (should be quick)

5. Observe the map behavior, you can see the map change to the minimum level.

### Screenshot

https://github.com/user-attachments/assets/c6253d8b-5a58-4d3f-8a84-0208a24b3c34

LaoshuBaby:
I also meet with this and can reproduce this on firefox and chrome

daishu0000:
Hi @tomhughes , could you please take a look at this issue when you have a moment? 🙏

tomhughes:
What is there to look at? A bug has been reported but there's no fix that needs review yet?

If you look at the open issues here you'll see there are loads of bugs around odd behaviour of the map when the sidebar opens/closes and for the most part they're just impossible to diagnose/fix. Certainly I wouldn't have a clue where to start.

daishu0000:
> A bug has been reported but there's no fix that needs review yet?

@tomhughes Now I have found the cause and solve it in #6520

PR #6520, opened 2025-11-11
Fix map zooming out when closing History sidebar during zoom animation

daishu0000:
### Issue:

See #6509

Map zooms out unexpectedly when closing Changeset History during zoom animation 

### Cause:

When zooming the map, or opening the History sidebar, the moveEndListener is triggered.

Inside moveEndListener, location.pathname is /history, which causes loadFirstChangesets to be called.

loadFirstChangesets then calls updateMap().

Since loadFirstChangesets is asynchronous (due to fetch), if the History sidebar is closed while it is still executing, location.pathname changes to /.

This leads to map.fitBounds() being triggered, causing the map to zoom unexpectedly.

### Solution:

Store whether the current page is /history in a variable isHistory before executing the fetch in loadFirstChangesets() and loadMoreChangesets().

This prevents changes to pathname during the fetch from causing unintended calls to map.fitBounds() and ensures the map does not zoom unexpectedly.
### Screenshots

old version: 

https://github.com/user-attachments/assets/16283fd0-7abb-4b93-ae29-4b356b148b23

new version:

https://github.com/user-attachments/assets/58acd6bf-7c2d-42bc-bba1-c860a1e38bfe

### DIFF
--- a/app/assets/javascripts/index/history.js
+++ b/app/assets/javascripts/index/history.js
@@ -168,10 +168,11 @@ OSM.History = function (map) {
 
   function loadFirstChangesets() {
     const data = new URLSearchParams();
+    const isHistory = location.pathname === "/history";
 
     disableChangesetIntersectionObserver();
 
-    if (location.pathname === "/history") {
+    if (isHistory) {
       setBboxFetchData(data);
       const feedLink = $("link[type=\"application/atom+xml\"]"),
             feedHref = feedLink.attr("href").split("?")[0];
@@ -197,7 +198,7 @@ OSM.History = function (map) {
           sidebar.scrollTop = 0;
         }
 
-        updateMap();
+        updateMap(isHistory);
       });
   }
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/index/history.js
```diff
@@ -269,7 +271,7 @@ OSM.History = function (map) {
     changesetsLayer.updateChangesetsGeometry(map);
   }
 
-  function updateMap() {
+  function updateMap(isHistory = false) {
```
tomhughes: We don't really need a default here because both callers have been changed to pass the argument.

daishu0000: OK, I have deleted the default value

daishu0000:
> cancelling any fetch

I've also considered this approach, but I'm not sure if it could introduce new bugs.

app/assets/javascripts/index/history.js
tomhughes: Thanks, but can you squash that into the previous commit please - we prefer to avoid fixup commits.

daishu0000: OK, I have squashed.

tomhughes:
Looks good to me now, thanks.

PR #6521, opened 2025-11-11
Append '.geojson' to download filename of direction

HolgerJeromin:
This is for more broad Browser support in generated download name. 

Related #6388 with discussion in #6519


1ec5:
Tested and verified that it still works in Firefox 144 on macOS.

HolgerJeromin:
@tomhughes Done

tomhughes (approved):
Looks good to me now, thanks.

PR #6524, opened 2025-11-13
Fix Wikipedia tag handling to support multiple links

### LINKED ISSUE #6397: Fix multiple *:wikipedia links
### URL

https://www.openstreetmap.org/node/5328066229#map=19/55.155481/8.773159

There is issue with displaying of
multiple subject:wikipedia links
https://www.openstreetmap.org/node/5328066229

Image

Image

Same with brand:wikipedia
https://www.openstreetmap.org/node/373744329

operator:wikipedia
https://www.openstreetmap.org/node/5167657812

And probably other *:wikipedia tags.

_No response_

gergelypap:
Closes #6397

### DIFF
--- a/test/helpers/browse_tags_helper_test.rb
+++ b/test/helpers/browse_tags_helper_test.rb
@@ -42,6 +42,10 @@ class BrowseTagsHelperTest < ActionView::TestCase
     html = format_value("wikipedia", "Test")
     assert_dom_equal "<a title=\"The Test article on Wikipedia\" href=\"https://en.wikipedia.org/wiki/Test?uselang=en\">Test</a>", html
 
+    html = format_value("wikipedia", "de:Berlin;en:London")
+    assert_dom_equal "<a title=\"The de:Berlin article on Wikipedia\" href=\"https://de.wikipedia.org/wiki/Berlin?uselang=en\">de:Berlin</a>;<a title=\"The en:London article on Wikipedia\" href=\"https://en.wikipedia.org/wiki/London?uselang=en\">en:London</a>",
+                     html
+
     html = format_value("wikidata", "Q42")
     dom = Rails::Dom::Testing.html_document_fragment.parse html
     assert_select dom, "a[title='The Q42 item on Wikidata'][href$='www.wikidata.org/entity/Q42?uselang=en']", :text => "Q42"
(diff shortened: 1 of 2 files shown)


tomhughes (changes requested):
Can you fix the lint failure please, then we should be able to get this merged...

gergelypap:
@tomhughes Oops, missed that one. I fixed it now.

test/helpers/browse_tags_helper_test.rb
```diff
@@ -186,57 +190,99 @@ def test_wikidata_links
...
+    assert_equal 1, links.length
+    assert_equal "https://en.wikipedia.org/wiki/Pablo_Picasso?uselang=en", links[0][:url]
+    assert_equal "en:Pablo Picasso", links[0][:title]
+
+    links = wikipedia_links("architect:wikipedia", "en:Frank Lloyd Wright")
+    assert_equal 1, links.length
+    assert_equal "https://en.wikipedia.org/wiki/Frank_Lloyd_Wright?uselang=en", links[0][:url]
+    assert_equal "en:Frank Lloyd Wright", links[0][:title]
+
+    links = wikipedia_links("buried:wikipedia", "en:Westminster Abbey")
```
pablobm: Suuuuper-nitpicky, but this example is a bit weird 😅 The value for this tag should be a person (or generally a formerly-living entity).

pablobm (commented):
Apart from my extremely nitpicky comment, the code looks good to me.

There's the question I posed at https://github.com/openstreetmap/openstreetmap-website/issues/6397#issuecomment-3328314366 of whether this should be allowed at all. I think that it should be allowed: if this is fair for Wikidata tags, then it should be fair for Wikipedia tags. There are good examples out there such as monuments depicting several notable people (like this [statue of the Obamas in Ireland](https://www.openstreetmap.org/node/10074619754#map=18/52.885049/-7.946060)), and probably other examples that my imagination can't reach for right now.

test/helpers/browse_tags_helper_test.rb
gergelypap: You're right. This is AI generated code. However I don't think it matters much, it's just a string match.

gergelypap:
I think it should be allowed. There are a lot of examples where buildings have more than one architects, or statues have multiple subjects.

test/helpers/browse_tags_helper_test.rb
```diff
@@ -186,57 +190,99 @@ def test_wikidata_links
...
+    assert_equal 1, links.length
+    assert_equal "https://en.wikipedia.org/wiki/Westminster_Abbey?uselang=en", links[0][:url]
+    assert_equal "en:Westminster Abbey", links[0][:title]
```
pablobm: This is a bit silly, but I'd prefer it to be corrected. When it comes to the code, it doesn't matter as it's just a text match. However for nerds like me the lack of consistency creaks. If I read this, later I may have an incorrect understanding of what the tag means when editing the map.

```suggestion
    links = wikipedia_links("buried:wikipedia", "en:Stephen Hawking")
    assert_equal 1, links.length
    assert_equal "https://en.wikipedia.org/wiki/Stephen_Hawking?uselang=en", links[0][:url]
    assert_equal "en:Stephen Hawking", links[0][:title]
```

gergelypap: Thanks for fixing! I'm all for the attention to detail mindset.

pablobm: Thank you. When you incorporate the changes and squash them into a single commit, I'll approve the PR :slightly_smiling_face:

tomhughes (approved):

test/helpers/browse_tags_helper_test.rb
pablobm: Thank you for indulging my nitpicks here :heart:

PR #6531, opened 2025-11-16
Improve dependabot configuration

tomhughes:
As discussed at SOTM-EU this updates the dependabot configuration to process non urgent update once a week and group them into a smaller number of PRs and enables that for ruby as well as npm and actions.

Updates will be grouped into one PR for each ecosystem except ruby where rails and rubocop are separated out.

Updates to rails other than patch updates are excluded altogether.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,27 @@
 version: 2
 updates:
+  - package-ecosystem: "bundler"
+    versioning-strategy: "increase-if-necessary"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


tomhughes:
I was wanting to test it but I'm not really sure how to do so!

HolgerJeromin:
Just merge and set a calendar entry to see if there was a dependabot PR and call it a day...

gravitystorm:
Looks good to me, let's see how it goes.

PR #6532, opened 2025-11-16
Resolve any open issues for a user when they are suspended

tomhughes:
As proposed by @Firefishy this automatically closes any issues that are open against a user when they get suspended.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -66,7 +66,7 @@ Metrics/BlockNesting:
 # Offense count: 26
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 340
+  Max: 350
 
 # Offense count: 58
 # Configuration parameters: AllowedMethods, AllowedPatterns.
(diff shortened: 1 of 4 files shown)


tomhughes:
This now resolves issues when a user is deleted as well as there doesn't seem to be much point keeping them open.

.rubocop_todo.yml
```diff
@@ -66,7 +66,7 @@ Metrics/BlockNesting:
 # Offense count: 26
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 340
+  Max: 350
```
pablobm, line 69: Tangentially to this PR: should we have this rule at all? It's the second change to this limit this month.

tomhughes: Well that, like most metrics cops, is in `.rubocop_todo.yml` because we never decided what we wanted to do with them...

pablobm: Oh, I see...

gravitystorm:
Merged, thanks!

PR #6533, opened 2025-11-16
Add Nominatim description terms

spixi:
Adds the following Nominatim terms to make them translatable

* boundary.low_emission_zone
* boundary.postal_code
* boundary.public_transport
* boundary.timezone
* landuse.highway (this refers to "highway" with the meaning of roadway in general (German: "Fahrbahn"), not to freeway/speedway (German: "Autobahn"))
* railway.crossing (not to be confused with highway.crossing, which may be translated completely differently, e. g. German highway.crossing = "Kreuzung", but railway.crossing = "Bahnübergang")
* historic:monastery


config/locales/en.yml
```diff
@@ -903,9 +903,13 @@ en:
           aboriginal_lands: "Aboriginal Lands"
           administrative: "Administrative Boundary"
           census: "Census Boundary"
+          low_emission_zone: "Low-emission Zone"
```
1ec5: Nit: Title case typically capitalizes each part of a hyphenated compound word:

```suggestion
          low_emission_zone: "Low-Emission Zone"
```

config/locales/en.yml
```diff
@@ -903,9 +903,13 @@ en:
...
           national_park: "National Park"
           political: "Electoral Boundary"
+          postal_code: "Postal Code"
           protected_area: "Protected Area"
+          public_transport: "Public Transport"
```
1ec5: This string will appear in front of the name of the feature without any additional modifications, so it should generally be a count noun. “A public transport” would be nonsensical. From the [`boundary=public_transport`](https://wiki.openstreetmap.org/wiki/Tag:boundary=public_transport) documentation, it sounds like “Public Transportation District” would be more descriptive?

config/locales/en.yml
```diff
@@ -1119,6 +1123,7 @@ en:
           garages: "Garages"
           grass: "Grass"
           greenfield: "Greenfield Land"
+          highway: "Highway"
```
1ec5: > this refers to "highway" with the meaning of roadway in general (German: "Fahrbahn"), not to freeway/speedway (German: "Autobahn")

This could be confusing in English too. It suggests that the feature is the highway (or roadway) itself, rather than the land around it. The closest real-world term in American English would be “Road Right of Way”, though it’s legal jargon.

config/locales/en.yml
```diff
@@ -1348,6 +1353,7 @@ en:
           abandoned: "Abandoned Railway"
           buffer_stop: "Buffer Stop"
           construction: "Railway under Construction"
+          crossing: "Crossing"
```
1ec5: > not to be confused with highway.crossing, which may be translated completely differently, e. g. German highway.crossing = "Kreuzung", but railway.crossing = "Bahnübergang"

At the same time, this could be confused with `railway=level_crossing`. [The wiki](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dcrossing) describes it as a pedestrian crossing. iD calls this a “Railway–Path Crossing”.

tomhughes: To be honest I'd have been inclined to call both a level crossing anyway! Certainly "Railway-Path Crossing" is quite a tortured phrase.

config/locales/en.yml
tomhughes: Well "Road Right of Way" is tortured and to me seems legally wrong because to me the right of way is the bit you have the right to travel along, not land that surrounds it that happens to belong to the highway department. Maybe something like "Highway Land" would do?

1ec5: Some legal systems use “road right of way” to mean the whole thing, even the landscaped portions, but we should try to find a better term that’s less prone to regional legal differences. “Road reservation” also comes to mind, but dialects that don’t speak of central reservations won’t have a clue about road reservations.

I guess I can’t think of anything better than “Highway Land”, as weird as it sounds.

config/locales/en.yml
1ec5: Fair enough. “Pedestrian level crossing” would be ideal, but “level crossing” is shorter. Unlike with presets, we can afford to reuse labels between multiple tags without necessarily warping tag usage.

pablobm:
@spixi - I'm not sure if reviewers received notifications after your latest changes, so I'm going to use this comment for a ping and a summary.

After the latest changes, the current version of this PR offers the following contested values:

```
public_transport: "Public Transportation District"
highway: "Road Right of Way"
crossing: "Railway-Path Crossing"
```

I'm going to leave this discussion to the experts :slightly_smiling_face:

1ec5:
There’s some further discussion in the threads above. Coming up with a good label for `highway` is proving more difficult than I thought at first, but the `public_transport` label seems OK to me.

gravitystorm:
```
public_transport: "Public Transportation District"
highway: "Road Right of Way"
```
These two are both fine for American-English, but sound very weird to British-English ears. We don't have an equivalent term to "Right of Way" as used by Americans. Additionally, "Right of Way" has a distinct meaning in the UK, which is related to access rights and not landuse.

```
crossing: "Railway-Path Crossing"
```
I'd either accept this as-is, or use "Pedestrian Railway Crossing" (i.e. using Pedestrian as a modifier to the term general term Railway Crossing).

In summary, I'm fine to merge these as `en.yml` can tolerate Americanisms and us Brits just need to rely on `en-GB.yml` fixing such things :smile:

gravitystorm (approved):

lonvia:
highway: "Road Right of Way"

I'd like to make a plead again here for using OSM English for the default 'en' terms. If Americanisms are required, they should go into 'en-US'. As a fairly fluent non-native speaker I wouldn't have the faintest idea what  a "Road Right Of Way" might be. As a fairly fluent OSMer, I wouldn't even start to guess that this might relate to a `landuse`. Something to do with access tags seems more likely. 

The equivalent `landuse=railway` seems to go with just 'Railway'. So for consistency, `landuse=highway` should be just 'Highway'? It's the same concept. A cleared but more verbose version would be 'Land used by Roads/Railways'.

1ec5:
As I said above, I’m fine with using an alternative “OSM English” term for `landuse=highway`. Also fine with putting “right of way” in an en-US localization. But that’s two things that don’t exist yet…

If we’re OK with it saying “Highway”, then that’s already the behavior in English, though this would give other localizations an opportunity to translate it to OSM Italian, OSM Thai, etc. Most likely they’d translate it literally. It would likely appear next to the result for the roadway, so a search for a street or an actual highway (e.g., Autobahn) could get confusing.

Highway Tudor Close, Residential Road Tudor Close Highway Tseung Lam Highway, Primary Road Tseung Lam Highway

“Road Reservation” is the other option I floated. It sounds very British, but it’s actually an Americanism. 🤔

gravitystorm:
@spixi thanks for your patience with this one. It seems like there is not yet consensus on what to use for the "highway" value, but the discussion has ground to a halt for now.

Could you please remove that key from the PR (and combine your changes into just one commit)? That way we can merge the uncontroversial changes straight away, and discuss the highway one separately.

gravitystorm:
I've merged this, after having removed the `landuse=highway` term. That can be handled in a separate PR, if desired.

@spixi thanks for your PR!

Issue #6538, opened 2025-11-18
Add info about using screenshots from OSM to the About page


aawiseman:
### Problem

The Communications Working Group has gotten a lot of questions lately about if people can use a screenshot of OSM in their publication or TV show etc. The answer is yes, just attribute us. We should make that info easier to find, such as on the About page.

For example if you go to [osm.org](http://osm.org/) and click About there is this page https://www.openstreetmap.org/about with some good info, but mostly it says reach out to us with legal questions. It would be good to have the screenshot issue on there since it is a common question and we want more people to be using and attributing OSM without having to reach out and ask us, because the answer is yes.

### Description

Such as something like this 

> 
> Legal
> 
> This site and many other related services are formally operated by the [OpenStreetMap Foundation](https://osmfoundation.org/) (OSMF) on behalf of the community. Use of all OSMF operated services is subject to our [Terms of Use](https://osmfoundation.org/wiki/Terms_of_Use), [Acceptable Use Policies](https://wiki.openstreetmap.org/wiki/Acceptable_Use_Policy) and our [Privacy Policy](https://osmfoundation.org/wiki/Privacy_Policy).
> 
> If you would like to use a screenshot directly from OpenStreetMap in your book, publication, film or TV show, you do not need to ask for permission as it is open source. Please make sure to attribute OpenStreetMap. Review these guidelines https://wiki.osmfoundation.org/wiki/Licence/Attribution_Guidelines and see examples at  https://wiki.openstreetmap.org/wiki/Films and https://wiki.openstreetmap.org/wiki/TV_series.
> 
> Please [contact the OSMF](https://osmfoundation.org/Contact) if you have any licensing, copyright or other legal questions.

tomhughes:
Surely rather than calling out one specific, and slightly odd, use case we should simply add pointers to licensing information and only invite people to email if that doesn't answer their question.

To be honest I'm surprised that's not there already... I mean it is, on the copyright page, but as we like to repeat everything on a dozen different similar pages I guess we should repeat it on the about page.

aawiseman:
Odd or not we get that request frequently, a few times the last few weeks for example. And I did include links to the guidelines in my proposed text, but it is quite verbose which is why I also included the shorter part for the screenshots. If you have a suggestion for a better way to word it or other links to include, feel free.

aawiseman:
Maybe something like 

[...]

PR #6541, opened 2025-11-19
Answer a very frequent question about licensing

pablobm:
As discussed at https://github.com/openstreetmap/openstreetmap-website/issues/6538, the Licensing WG get this question often enough that it's worth including in the About page.

I have taken a couple of liberties with the phrasing. Also please someone let me know if the use of an em dash is correct as I'm not a user myself.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2344,9 +2344,19 @@ en:
       legal_1_1_aup_url: https://wiki.openstreetmap.org/wiki/Acceptable_Use_Policy
       legal_1_1_privacy_policy: Privacy Policy
       legal_1_1_privacy_policy_url: https://osmfoundation.org/wiki/Privacy_Policy
+      legal_1_2_html: |
+        If you would like to use OpenStreetMap maps or data please make sure to attribute OpenStreetMap—refer to %{attribution_guidelines_link} for more information.
+      legal_attribution_guidelines_title: Attribution Guidelines
+      legal_attribution_guidelines_url: https://wiki.osmfoundation.org/wiki/Licence/Attribution_Guidelines
+      legal_1_3_html: |
+        Specifically for using a screenshot from OpenStreetMap directly in your book, publication, film or TV show, you do not need to ask for permission as data in OpenStreetMap is open for all to use. Refer to the %{attribution_guidelines_link} and see existing examples in %{attribution_example_film} and %{attribution_example_tv}
+      legal_attribution_example_film_title: Film
+      legal_attribution_example_film_url: https://wiki.openstreetmap.org/wiki/Films
+      legal_attribution_example_tv_title: TV
+      legal_attribution_example_tv_url: https://wiki.openstreetmap.org/wiki/TV_series
       legal_2_1_html: |
         Please %{contact_the_osmf_link}
-        if you have licensing, copyright or other legal questions.
+        for any other licensing, copyright or other legal questions.
       legal_2_1_contact_the_osmf: contact the OSMF
       legal_2_1_contact_the_osmf_url: https://osmfoundation.org/Contact
       legal_2_2_html: |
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -2344,9 +2344,17 @@ en:
       legal_1_1_aup_url: https://wiki.openstreetmap.org/wiki/Acceptable_Use_Policy
       legal_1_1_privacy_policy: Privacy Policy
       legal_1_1_privacy_policy_url: https://osmfoundation.org/wiki/Privacy_Policy
+      legal_1_2_html: |
+        If you would like to use OpenStreetMap maps or data please make sure to attribute OpenStreetMap—refer to %{attribution_guidelines_link} for more information.
+      legal_attribution_guidelines_title: Attribution Guidelines
+      legal_attribution_guidelines_url: https://wiki.osmfoundation.org/wiki/Licence/Attribution_Guidelines
+      legal_1_3_html: |
+        Specifically for using a screenshot from OpenStreetMap directly in your book, publication, film or TV show, you do not need to ask for permission as OpenStreetMap is open source. Refer to the %{attribution_guidelines_link} as well as the examples at %{attribution_example_1_link} and %{attribution_example_2_link}
```
tomhughes: I'm not sure "open source" is the best phrase? or do we use that elsewhere?

pablobm: Perhaps @aawiseman can comment on this?

1ec5: We normally describe OSM as “open data”. Would that have the intended effect for authors and artists who might be unfamiliar with that term, or “open source” for that matter?

1ec5: Can we avoid embedding a raw URL in the text? A link label would be a lot more readable. We could link the “film” and “TV show” in the first sentence to these wiki pages to avoid repetition.

aawiseman: I think open data is fine, or maybe something like "has an open license" - will defer to you all

aawiseman: yeah I agree, I just pasted it quickly as a draft. would be good to have the proper wiki url formatting

aawiseman: or we could remove that altogether, just "you do not need to ask for permission. Refer to..." I'm fine either way, no strong preference.

pablobm: I have updated it to "open for all to use". There's the valid concern that people may stop reading there and just use without attribution, but I'm hoping that it being followed immediately by "Refer to the Attribution Guidelines..." should be enough to avoid this.

pablobm: Reformatted with proper, titled links, plus some rephrasing.

tomhughes:

config/locales/en.yml
aawiseman: Sounds good! Thanks

config/locales/en.yml
```diff
@@ -2344,9 +2344,19 @@ en:
       legal_1_1_aup_url: https://wiki.openstreetmap.org/wiki/Acceptable_Use_Policy
       legal_1_1_privacy_policy: Privacy Policy
       legal_1_1_privacy_policy_url: https://osmfoundation.org/wiki/Privacy_Policy
+      legal_1_2_html: |
+        If you would like to use OpenStreetMap maps or data please make sure to attribute OpenStreetMap—refer to %{attribution_guidelines_link} for more information.
+      legal_attribution_guidelines_title: Attribution Guidelines
+      legal_attribution_guidelines_url: https://wiki.osmfoundation.org/wiki/Licence/Attribution_Guidelines
+      legal_1_3_html: |
+        Specifically for using a screenshot from OpenStreetMap directly in your book, publication, film or TV show, you do not need to ask for permission as data in OpenStreetMap is open for all to use. Refer to the %{attribution_guidelines_link} and see existing examples in %{attribution_example_film} and %{attribution_example_tv}
```
1ec5, line 2352: This is missing a period at the end.

pablobm: D'oh! PR at https://github.com/openstreetmap/openstreetmap-website/pull/6556

PR #6542, opened 2025-11-19
Fix context menu link color in light mode

### LINKED ISSUE #6537: Poor contrast on context menu in light theme
### URL

_No response_

When clicking on the new context menu, the link text becomes white; this is correct in the dark theme but produces an unreadable white-on-grey in the light theme.

Image

jleedev:
Fixes #6537

### Description
It appears that the intent was to make the 'active' state the same as the 'hover' state for the map context menu. Make it so.

### How has this been tested?
Manual test

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1014,6 +1014,7 @@ img.trace_image {
   .dropdown-item {
     &:hover, &:active {
       background-color: $gray-200;
+      color: var(--bs-dropdown-link-hover-color);
     }
   }
   &.cm_dropdown {


app/assets/stylesheets/common.scss
hlfan, line 1039:
```suggestion
#map-context-menu {
  .dropdown-item:active {
    @extend .dropdown-item:focus;
  }
  &.cm_dropdown {
    position: absolute;
    z-index: 1000;
  }
  &.cm_dropdown_menu {
    position: absolute;
    display: block;
    min-width: 200px;
  }
}
```

mmd-osm:
> I'm not really sure why we override anything to begin with.

I have adjusted contrast/colors so that the Firefox accessibility tool would report an "AAA" rating for _WCAG standards for accessible texts_.  Without adjustment, it was barely readable, similar to https://github.com/openstreetmap/openstreetmap-website/issues/6537. 

By the way, I have noticed the poor contrast when clicking on an entry, but that had lower priority for me. Good that there's some follow up activity here to tackle this bit as well.

gravitystorm:
@jleedev Thanks again for this pull request. We've gone with a slightly different approach, see #6576.

PR #6543, opened 2025-11-19
Point potential contributors to the Roadmap

pablobm:
With the publication of the [Core Software Roadmap](https://github.com/openstreetmap/software-roadmap) ([announcement](https://community.openstreetmap.org/t/core-software-roadmap/138102)), perhaps we should consider including a link to it in our contribution guidelines.

I am doing exactly that here, in addition to expanding with additional wording to help manage the expectations of potential contributors.

Thoughts?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,12 +18,62 @@ This guide covers our development workflow, coding standards, and how to get you
 * https://www.ruby-lang.org/ - The homepage of Ruby which has more links and some great tutorials.
 * https://rubyonrails.org/ - The homepage of Rails, also has links and tutorials.
 
-### Finding Issues to Work On
+### Finding Opportunities to Contribute
 
 > [!NOTE]
 > We don't assign issues to individual contributors. You are welcome to work on any issue, and there's no need to ask first.
+> For more details see [our FAQ](FAQ.md)
 
-For more details see [our FAQ](FAQ.md)
+We welcome the community to contribute to this repository in any form:
+
+- **Documentation:** are the docs clear? Are they correct? Did you come across
+  any issues when trying to follow them? Could something be explained more
+  clearly?
+- **Bug reporting:** something is broken on the website? Something doesn't work
+  as expected? Create an issue and give as much detail as possible, so that
+  others can understand the context, the conditions in which it happens,
+  and your use case. Or comment on an existing report if one already exists.
+- **Bug triage:** we receive many issue reports. Can you take one and reproduce
+  it? Can you add more detail that could help others fix the issue more
+  easily? Or perhaps you think it's not really an issue and should be left
+  as is? Come to the issue tracker and let us know your thoughts.
+- **Feature discussion:** is there a use case that is not covered? Is there
+  something that you wish that the website did, or did differently? Create
+  an issue to discuss it, or comment on an existing issue if it's already
+  being discussed. Try to express your use case in a way that helps understand
+  your needs well, and frames them in the context of the wider community.
+- **Code reviews:** at any given point, there will be pending PRs, waiting for
+  reviews. Can you take on one, understand what it's trying to do, and
+  provide actionable feedback? Is the code clear, maintainable, and readable?
+  Would you do something differently? Are useful, clear tests provided?
+- **Code:** take an existing issue and try to fix it, or try to implement
+  an idea.
+- And probably other ideas not captured here.
+
+Bear in mind that OSM attracts very diverse contributors with very different
+needs. Others may have needs different from yours, and reaching a consensus
+is sometimes difficult.
+
+If you want to code a feature to this repository, we recommend that you ask
+for feedback early and often. Create an issue to discuss it, or start with a
+["draft" PR](https://github.blog/news-insights/product-news/introducing-draft-pull-requests/)
+that shows your intention clearly and can be used to provide early
+feedback. We are not strict on how you communicate, but it's important that
+you do communicate, early and often, so that you can get feedback quickly.
+This will help you get buy-in from the maintainers, which will translate into
+less waste for everyone and a much easier time getting your code merged.
+
+Bug fixes should be more straighforward than new features, but the same
+guidance applies. If it turns out to be more complex than initially expected,
+stop for a moment and seek feedback, be it in an issue or in a draft PR.
+
+If you are looking for some additional inspiration, you may want to have
+a look at the [Roadmap](https://github.com/openstreetmap/software-roadmap)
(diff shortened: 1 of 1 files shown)


tomhughes:
I mean we should check we agree with it first didn't it didn't originate with us.

CONTRIBUTING.md
```diff
@@ -22,8 +22,22 @@ This guide covers our development workflow, coding standards, and how to get you
...
+
+In 2025 we published a [Roadmap](https://github.com/openstreetmap/software-roadmap)
+that can be helpful to understand the desired direction of the project. Some
+items in it are already being discussed in issues. You may want to make sure
+that you are up to date with discussions before moving to implementation.
+
+Before spending any significant time, we recommend that you create an issue
+or a draft PR that communicates your intention clearly.
+Make sure that you have buy-in from the maintainers and you will have a
+much easier time getting your code merged.
```
1ec5: Some people keep thinking have to ask for an issue to get assigned to them. We have a note about that just above here, but it probably needs to have more nuance. Basically, most bug reports are pretty straightforward, but feature requests really need buy-in about the concept and design before diving in too deep.

pablobm: I have reworded the whole thing now. I tried to make some distinction between features (which can find more resistance if unrequested) and bug fixes (which _should_ be straightforward). All while not losing sight that both can end up being more effort than initially planned and therefore should lean on communicating early and often.

gravitystorm:
When I read the PR I felt it didn't quite hit the spot in a way that I found hard to articulate, but let me try.

Firstly, the roadmap has a much wider scope than just this project, because it covers "core software" in general and not just the stuff we maintain here. This can be confusing if someone reads the roadmap and thinks that any particular problem mentioned there is therefore related to this project (e.g. osmdbt). So this difference in scope should be made clear in this PR.

The proposed documentation in this PR is also slightly confusing when it mentions "the project". At first it seems to refer to openstreetmap-website, but then it makes more sense in places for it to mean the wider "OpenStreetMap project" when referring to the roadmap, because "we" (as developers of this project) certainly didn't propose changes or features for other projects!

The roadmap also contains a completely different definition of Developer Experience than what we use here. In the roadmap, it's all about developers using the OSM data and building things on top of that (hence discussion of switch2osm, welcome mat, extract formats etc) whereas for us, Developer Experience is about making it easier for people to contribute to this codebase, e.g. installing dependencies, writing code, running tests. This again can be confusing when the roadmap is read in terms of openstreetmap-website, instead of core openstreetmap software in more general terms.

Overall the roadmap feels more like a description of what OSMF wants to focus on, both to indicate what they intend to work on and to manage expectations of what they won't be working on. Which makes complete sense, and with paid staff it's better to have a rough plan! Thankfully the bits that are relevant to this project are (mostly) aligned with what I want to see too (I'll let other maintainers/developers chime in if they feel differently). 

However, both the roadmap and this PR have the issue that it really sounds like if it's not on the roadmap, you have to make a suggestion to have it considered for being added. That's fine if you expect the OSMF team to focus on your topic, but that's really not the case for contributing to this repo. If a contributor comes along out of the blue with an entirely different feature that's useful and meets our general criteria, but that feature is not even hinted at in the roadmap, I don't want them to feel they need to persuade anyone to add it to the roadmap, instead of just adding the feature directly here as normal. All four paragraphs of this PR are phrased in a way that discourages contributions that don't align with the roadmap, but the roadmap is to set expectations of what other people are working on, and not meant to discourage any other non-roadmap contributions.

Wow, that was more words than I intended. But to sum up - yes, I think it's reasonable to link to the roadmap, but let's figure out a better way to phrase things and add a bit more clarity to exactly how it relates to this project.

pablobm:
Thank you for your thoughtful review @gravitystorm. I have reworked the text heavily, with these aims:

- De-emphasise the roadmap to make sure it's not presented as the only (or even primary) source of inspiration.
- Avoid the use of the word "project" as it can be ambiguous as to whether it refers to this website or the wider OSM ecosystem.
- Illustrate the many possible forms of contribution.
- Manage expectations and encourage frequent, productive communication.

For which I have used maaaany words though. Not sure if the resulting wall of text is going to end up discouraging to readers :sweat_smile:

1ec5:
> The roadmap also contains a completely different definition of Developer Experience than what we use here. In the roadmap, it's all about developers using the OSM data and building things on top of that (hence discussion of switch2osm, welcome mat, extract formats etc) whereas for us, Developer Experience is about making it easier for people to contribute to this codebase, e.g. installing dependencies, writing code, running tests. This again can be confusing when the roadmap is read in terms of openstreetmap-website, instead of core openstreetmap software in more general terms.

Yes, as I put together the roadmap, I heard a lot about both needs. Ultimately I ended up putting the ideas around fostering contributions to this repository under the “sustainability” banner, because I view it as even more central to keeping the lights on than the complementary goal of making OSM more approachable as a platform for other projects. I appreciate that this approach doesn’t come through very clearly though. It’s a lot easier to talk about how we want to improve the contributor experience directly in the context of this specific repository, so, here in the contributing guide. @pablobm and I have revisited this topic many times just in the last couple months and expect to in the future.

CONTRIBUTING.md
```diff
@@ -18,12 +18,61 @@ This guide covers our development workflow, coding standards, and how to get you
...
+  an idea.
+- And probably other ideas not captured here.
+
+Bear in mind that OSM attracts very diverse contributors with very different
+needs. Others may have needs different from yours, and reaching a consensus
+is sometimes difficult.
+
+If you want to code a feature to this repository, we recommend that you ask
+for feedback early and often. Create an issue to discuss it, or start with a
+Draft PR that shows your intention clearly and can be used to provide early
```
tomhughes: Should `Draft` be capitalised here? There's another use of the phrase later that isn't...

tomhughes (commented):
This mostly looks pretty good to me - possibly we should add something at some point about what we consider "in scope" to try and avoid people spending time on out of scope features though.

CONTRIBUTING.md
pablobm: Good catch. Initially I capitalised it on purpose as I am referring to the specific feature to create "draft" PRs.

On second thought, that might be too implicit for new contributors. I have now changed style:

- All instances are de-capitalised.
- The first instance is in quotation marks and links to an explanation of what it refers to.

pablobm:
Yeah, a definition of scope would be great... but if there's one I don't know what it is :sweat_smile:

tomhughes:
I think this is good to merge now, thanks.

We can address the question of scope later - the main thing I can think of is the fact that we're targeting mappers rather than end users.

Issue #6510, opened 2025-11-10
Move SPAMMY_PHRASES to the database


pablobm:
### Problem

User profiles are a vector for spam in the website. Some instances are caught using a hard-coded list of "spammy phrases" that can be found at [`lib/rich_text.rb`](https://github.com/openstreetmap/openstreetmap-website/blob/b715c12a0aae54631addfff1b93600e0337df3d3/lib/rich_text.rb#L4).

Being hard-coded, this list is difficult to manage. See https://github.com/openstreetmap/openstreetmap-website/pull/6044

### Description

Move the list of spammy phrases to the database. For the moment don't include an UI. Admins can add/remove with SQL manually. An UI will be added separately.

tomhughes:
How is this actually going to work though?

Matching text against the database every time we compute a spam score would be very costly I think so a re we just going to load the list once? But then how do we know to reload it if the database changes?

pablobm:
That'll depend on the size of the list and the frequency of signups. If both are small enough, the current code should do the job. I just had a quick look and I'm seeing 13 SELECT queries and a single INSERT query generated as a result of a successful `POST /users` request, so initially a new SELECT with a reasonable payload shouldn't be much trouble. Also, from the numbers I've seen, I get the impression that although we have many users, we are not in the "hundreds per minute" signups territory at the moment.

If these parameters change (more signups, more phrases) then we can adapt. For example we could load the list from DB into memory only when the application restarts. Tradeoff that phrases won't be effective immediately, but saves us a potentially costly SELECT every time. Forked processes can share the list thanks to the magic of copy-on-write. (I don't know how our application servers work so this might not apply).

tomhughes:
It's not just signups that evaluate the spam score though - currently it's also every time they post a diary entry or comment or somebody reports them and every time an administrator views their profile.

Overall that's probably not too bad - it's less than I thought it would be but maybe that means we're not including enough things in the score ;-)

pablobm:
Ah, thanks for the info. I had missed that indeed. I'm going to take that as "good to go" and will look into it soon 🙂 We can tweak as we go depending on production behaviour.

This reminds me: the OpenTelemetry integration would be amazing for these things. No idea what the roadmap is for that though?

pablobm:
PR ready at https://github.com/openstreetmap/openstreetmap-website/pull/6549

Issue #6511, opened 2025-11-10
Create UI to manage spammy phrases


pablobm:
### Problem

Following https://github.com/openstreetmap/openstreetmap-website/issues/6510, the only way to manage the list of "spammy phrases" will be by directly editing the DB. This requires special access (which limits who can do it), and is potentially dangerous (you don't want to mess up that `DELETE` query).

### Description

Create a UI that can be used to manage the list of spammy phrases.

Initially only admins will have access to this. In the future this might be extended to other roles.

Alternative proposal at https://github.com/openstreetmap/openstreetmap-website/issues/6566

Rustix69:
@pablobm Is their any figma to give a sense like how the UI will Look. Can I start to work on this PR ?

pablobm:
I think this is not a good issue to work on at the moment. It's not clear that admins actually need it, and instead it might be enough with https://github.com/openstreetmap/openstreetmap-website/issues/6566 instead

pablobm:
Closing as this doesn't seem necessary anymore. Admins are loading lists of phrases manually in the DB.

PR #6549, opened 2025-11-20
Move spammy phrases to DB

pablobm:
Closes https://github.com/openstreetmap/openstreetmap-website/issues/6510

This moves `SPAMMY_PHRASES` to DB, where admins can add/remove phrases at their pleasure, and the secret sauce of our spam scoring is not open for all to see. In the process I also:

- Refactored the existing spam scoring into its own module.
- Changed the matching of spammy phrases to be case-insensitive.
- Added an actual test for the phrase matching, as this was missing.

Note that this doesn't add any UI. Admins will have to manage phrases by manually inserting/deleting from DB. If a UI is desired, admins are invited to express their interest at https://github.com/openstreetmap/openstreetmap-website/issues/6511 :smile:

### DIFF
--- a/lib/spam_scorer/rich_text.rb
+++ b/lib/spam_scorer/rich_text.rb
@@ -23,8 +23,9 @@ module SpamScorer
         link_proportion = link_size.to_f / doc.content.length
       end
 
+      comparable_content = to_comparable_form(doc.content)
       spammy_phrases = SpammyPhrase.pluck(:phrase).count do |phrase|
-        doc.content.include?(phrase)
+        comparable_content.include?(to_comparable_form(phrase))
       end
 
       ([link_proportion - 0.2, 0.0].max * 200) +
@@ -35,5 +36,9 @@ module SpamScorer
     private
 
     attr_reader :text
+
+    def to_comparable_form(str)
+      str.downcase(:fold).unicode_normalize(:nfkc)
+    end
   end
 end
--- a/test/lib/spam_scorer_test.rb
+++ b/test/lib/spam_scorer_test.rb
@@ -30,8 +30,18 @@ class SpamScorerTest < ActiveSupport::TestCase
   def test_spammy_phrases
     create(:spammy_phrase, :phrase => "Business Description:")
     create(:spammy_phrase, :phrase => "Additional Keywords:")
-    r = RichText.new("markdown", "Business Description: totally legit beesknees. Additional Keywords: apiary joints")
+    create(:spammy_phrase, :phrase => "Große Preise")
+    create(:spammy_phrase, :phrase => "Relevant Experience:")
+    r = RichText.new(
+      "markdown",
+      <<~SPAM
+        BUSINESS DESCRIPTION: totally legit beesknees.
+        Additional Keywords: apiary joints.
+        GROSSE PREISE: VIEL GELD.
+        RELEVANT EⅩPERIENCE: disguising latin characters as roman numerals.
+      SPAM
+    )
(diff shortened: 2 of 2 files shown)


lib/spam_scorer/rich_text.rb
```diff
@@ -0,0 +1,40 @@
...
+      else
+        doc.xpath("//a").each do |link|
+          link_count += 1
+          link_size += link.content.length
+        end
+
+        link_proportion = link_size.to_f / doc.content.length
+      end
+
+      downcased_content = doc.content.downcase
```
pnorman: 1. We should apply a Unicode NFKC normalization before lowercase
2. Downcasing should probably case folding, not just case mapping
3. The same transformations need to be applied to the phrases retrieved from the DB

pablobm: Good points, thank you! I have implemented something now. Potentially very naive approach, no idea. Let me know what you think.

tomhughes: Really you should use the unicode case folding transformation if you want to do case insensitive comparison rather than using upcase or downcase but it doesn't look like ruby has that as a standard string method :disappointed:

tomhughes: Ah it is available but confusingly as `downcase(:fold)` and that is what we've actually using now.

pablobm:
@tomhughes - How are DB migrations handled in prod? Is the existence of a migration here ok?

Firefishy:
Maybe a rake job to import a text file? It then would be a small job to automate import using chef.

tomhughes:
Yes having migrations is fine - they don't run automatically in prod but I'll run it manually as part of merging anything with a migration.

I had a quick look at this last night but I'll try and do a proper review soon - hopefully at the weekend or something.

pablobm:
> Maybe a rake job to import a text file?

I think that's a great idea :+1: I'll do that separately though as even for something that simple there could be some discussion. Issue created at https://github.com/openstreetmap/openstreetmap-website/issues/6566

tomhughes:
> Maybe a rake job to import a text file? It then would be a small job to automate import using chef.

Hard to make a rake job that can apply updates though, and if we constantly reload the whole table we'll be chewing up record ids at a rate of knots and possibly bloating the table and/or index badly as well.

pablobm:
That's fair, I wonder about that sometimes. This specific example is very simple and could have been done with modules and functions.

I stick to this practice because, when things grow a bit, I end up wanting a bit more. To hold a bit more of state, or to have the ability to inject options or a dependency for a test or reuse. Things like that, which are a bit easier with classes, avoiding a method with too many arguments. Also Ruby does nudge you in the direction of using classes and I let myself be nudged.

By the way, I'm realising that there's something I normally do that I didn't do here: I like putting all this stuff in `app/models/`! I'm of the opinion that `lib/` should be used sparingly and all business logic should go in `app/models/`. But I can live with this for now.

(I'm also of the opinion that all ORM classes should be confined to a separate `app/records/` with names like `UserRecord`, `NodeRecord`, etc, but I won't go that far here, and it's not conventional anyway).

tomhughes:
I'm not sure about putting things in `app/models` that aren't actually models, or concerns that are going to get mixed into models but we don't need to debate that now ;-)

tomhughes:
I think this one is good to merge now, thanks.

Firefishy:
Initial feedback: We've populated the list with 89 words or phrases and there has been a nice improvement in the number of SEO/Spam accounts automatically being suspended (Up to ~10 per hour).

PR #6551, opened 2025-11-20
Rename export controller actions

gravitystorm:
Refs #5539

Although I've renamed the actions, I've left the urls alone for now. Further refactoring might include things like:

* Reconsider whether the embedded html source, data download and image export functionality should all really be in the same controller
* Make resourceful routes for whatever gets decided
* Remove hardcoded form target urls from forms and use route helpers instead


tomhughes:
Looks good to me, thanks.

pablobm:
> I've left the urls alone for now

Would changing URLs be desired? It would break the API contract. Or am I misunderstanding?

tomhughes:
Those URLs aren't API as such - normally anything that isn't under `/api` can be expected to change at any time.

The `/export/embed` one is a bit special though because the whole point is that we're giving people something to put on their web site so if we changed that we would at least need to keep a redirect or continue recognising the old route as an alias for a long time.

There's no such issue with `/export/finish` though.

pablobm:
Ah, makes sense. Should have paid a bit more attention, thank you :+1:

PR #6552, opened 2025-11-20
Reconfigure i18n-js to avoid top-level i18n directory

### LINKED ISSUE #6550: Reconfigure i18n-js to avoid top-level i18n directory
As part of our [i18n-js configuration](https://github.com/openstreetmap/openstreetmap-website/blob/74a41d95fa146523bd856cd63452e1ed3c462696/config/i18n.yml), we currently create a top-level `i18n` directory. It's used to hold [a single template file under source control](https://github.com/openstreetmap/openstreetmap-website/tree/74a41d95fa146523bd856cd63452e1ed3c462696/i18n/templates), and a number of intermediate `.json` files created while generating the corresonding `.js` files in `app/assets/javascripts/i18n/`

From a developer experience point of view, it would be great to move these elsewhere, because:
* we've already got a lot of top-level directories, that e.g. push our readme far down the page
* these files aren't important enough to warrant a top-level directory of their own
* more importantly, it can mislead developers to think that this is where to look for translation-related matters, when the vast majority of the time they should look elsewhere e.g. `config/locales/`.

As far as I can tell we get free choice over the directory used, but I haven't tested to see if i18n-js has any built-in reasons to prefer that specific directory.

tomhughes:
Fixes #6550.


gravitystorm:
Thanks! Merged.

PR #6555, opened 2025-11-21
#3532 Created a 404 Page

anant2526:
I made a beautiful / minimal 404 page which is made as I was told .

### DIFF
--- a/404 Page
+++ b/404 Page
@@ -1,0 +1,65 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8" />
+<meta name="viewport" content="width=device-width,initial-scale=1" />
+<title>404 — Page not found</title>
+<head>
+  <style>
+    .container {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    padding: 40px;
+    gap: 40px;
+    margin-top: 150px;
+}
+
+.image-wrapper img {
+    max-width: 420px;
+    height: auto;
+}
+
+.text-wrapper {
+    max-width: 500px;
+}
+
+.text-wrapper h1 {
+    font-size: 80px;
+    font-weight: 700;
+    margin-bottom: 20px;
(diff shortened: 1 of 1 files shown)


tomhughes:
Told by who? Is there some background to this that I'm missing?

pablobm:
I gather that this is about the discussion at https://github.com/openstreetmap/openstreetmap-website/issues/3532. @anant2526, please make sure to reference previous discussions so that we can tell the context better. I see that you put the issue number in the PR title, but that's easy to miss.

You are providing an HTML document, but this doesn't automatically work. It needs to be integrated in the application correctly. I did a quick fix at https://github.com/openstreetmap/openstreetmap-website/compare/master...pablobm:openstreetmap-website:404-page, but it's still not good enough. It needs to follow the template and layout conventions, like the current error page does. See [app/views/errors/not_found.html.erb](https://github.com/openstreetmap/openstreetmap-website/blob/a2d0762f2a4499e7ddcbc6dde29bac847c249024/app/views/errors/not_found.html.erb). If you are new to Ruby on Rails, I recommend you read the documentation (eg: https://guides.rubyonrails.org/layouts_and_rendering.html).

As for the design itself, it looks nice but it has issues. For example the text is not visible in small screens:

Current | This PR
--- | ---
404-mobile-current | 404-mobile-pr

I recommend that you look into how we use Bootstrap in this application, as it has features to help with responsive design.

tomhughes:
Yes we definitely want something that uses rails and bootstrap and our stylesheets rather than completely separate HTML with inline styles.

Also whatever design we use for the 404 page should be applied to other error pages as well.

gravitystorm:
Hi @anant2526, thanks for making this pull request. There are several large changes that would need to be made before we could merge this, as other people have described above.

I'm going to close this PR, but you are welcome to open a fresh one when you are ready.

PR #6556, opened 2025-11-21
Add missing period

pablobm:
As caught by @1ec5 at https://github.com/openstreetmap/openstreetmap-website/pull/6541#discussion_r2547655721


gravitystorm:
Thanks both for the catch + fix.

PR #6559, opened 2025-11-22
Use a del element for deleted objects in a changeset

### LINKED ISSUE #6558: Cross out deleted elements using `<del>`
https://github.com/openstreetmap/openstreetmap-website/pull/6448#issuecomment-3558167981 points out that deleted and redacted elements are crossed out using a presentational `<s>` tag. We should use a semantic `<del>` tag instead for the benefit of screen readers.

https://github.com/openstreetmap/openstreetmap-website/blob/7c47232da8c3b9ea6447e4719eab2a65cb72e3f4/app/helpers/browse_helper.rb#L63-L69

tomhughes:
Fixes #6558.

### DIFF
--- a/app/helpers/browse_helper.rb
+++ b/app/helpers/browse_helper.rb
@@ -62,7 +62,7 @@ module BrowseHelper
 
   def element_strikethrough(object, &)
     if object.redacted? || !object.visible?
-      tag.s(&)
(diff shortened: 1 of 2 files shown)


gravitystorm:
Merged, thanks!

PR #6560, opened 2025-11-22
Workaround for ruby 3.5.0 compatibility with delayed_job

tomhughes:
This works around https://github.com/collectiveidea/delayed_job/issues/1239 which is causing a deprecation warning with ruby 3.4.7 that will become an error with 3.5.0:

```
/home/tom/.bundle/ruby/3.4.0/gems/delayed_job-4.1.13/lib/delayed/worker.rb:9: warning: /usr/share/ruby/benchmark.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add benchmark to your Gemfile or gemspec to silence this warning.
Also please contact the author of delayed_job-4.1.13 to request adding benchmark into its gemspec.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,6 +44,10 @@ gem "argon2"
 # Support brotli compression for assets
 gem "sprockets-exporters_pack"
 
+# Needed for ruby 3.5.0 compatibility with delayed_job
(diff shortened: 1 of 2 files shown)


gravitystorm:
Merged, thanks!

Issue #6546, opened 2025-11-19
Improvements to UI to suspend/delete users


pablobm:
### Problem

Recently when reviewing the facilities to suspend/delete users with @firefishy, we agreed that these were a bit confusing and could be improved. Aim:
- Better express what the actions do: currently "Hide this User" marks the uses as deleted, while "Delete this User" additionally scrubs their data.
- Move dangerous actions out of the way: the "delete" action cannot be undone and sits too close to "hide".
- Make it easier to interpret data in the future: having admins being able to delete users obscures from us who self-deleted vs who was removed by an admin.

### Proposal

- Remove "hide" and "delete" actions.
    - Users should only be marked as "deleted" when they carry out a self-destroy
- Add a "suspend" action.
- Existing "hidden" users will be left as-is initially. At a later date these users can be moved to "suspended" state by the Ops team with a script, but it's not a priority.

Edits:
- 2025-11-27 Existing "hidden" users will be left untouched.
- 2025-11-25 Corrected problem description represent my current understanding.
- 2025-11-24 Reworked proposal, incorporating feedback received at the related https://github.com/openstreetmap/openstreetmap-website/issues/6544

pablobm:
Reworked description based on discussion at https://github.com/openstreetmap/openstreetmap-website/issues/6544.

pablobm:
@tomhughes @Firefishy - Does this look reasonable to you?

tomhughes:
It's mostly fine I think.

I mean I do occasionally delete a user at user request but I usually only do that if they insist on having the email address removed (and they have no edits) and in that case I have to do it from the CLI anyway so removing the delete option here probably doesn't matter.

The third part is tricky especially for accounts deleted some time ago (basically before we had self deletion) as it may be hard to distinguish user requested deletions from admin actions.

pablobm:
Thank you @tomhughes - Do you see that as a blocker? My thinking, and from conversation with @Firefishy, is that a heuristic to get 90% cases right should be "good enough"... but I might be wrong! 🙂

If that's ok, PR ready at https://github.com/openstreetmap/openstreetmap-website/pull/6564

tomhughes:
Working how to update old accounts certainly isn't a blocker, no.

PR #6564, opened 2025-11-25
Rework UI to hide/delete/suspend users

pablobm:
Closes https://github.com/openstreetmap/openstreetmap-website/issues/6546

As recently discussed with @firefishy, this change clarifies the admin interface:

- Remove "hide" and "delete" actions. Users should only be marked as "deleted" when they carry out a self-delete.
- Add a "suspend" action. This is now preferred over a deletion.
- In the code, renamed "unhide" as "undelete", which is more descriptive. Admins perform this action when a user regrets their self-deletion, which apparently happens every now and then.

Note that we'll be left with pre-existing "hidden" users. After discussing with @firefishy, these users can be moved to the "suspended" state by the Ops team with a script, but it's not a priority.

### Bulk suspension
In the "user list" page, admins were able to "hide" (mark as deleted) a list of users. Now they can instead suspend them.

The "hide" was done with a single `UPDATE` query: `User.where(:id => ids).update_all(:status => "deleted")`. However this is not so simple with suspending. When a user is suspended, we call `close_open_issues` on them, so we need to make sure that this is happening. Therefore, we bulk-suspend users with `User.where(:id => ids).each(&:suspend_if_possible!)` instead. This is slower, but shouldn't be an issue unless admins decide to suspend hundreds of accounts in one fell swoop.

This `#suspend_if_possible!` API is also new. The possibility exists that admins erroneously try to suspend an already-suspended user. This would raise an error, so this new API avoids this.

### Suspending "confirmed" users
Currently we are not allowing suspension of "confirmed" users. There seems to be no reason for this, and to quote @tomhughes [in the comments of this PR](https://github.com/openstreetmap/openstreetmap-website/pull/6564#pullrequestreview-3531723379):

> if we change our minds I don't see why we should not be able to suspend them

So I have changed the constraint in the state machine, allowing confirmed users to be suspended.

In addition, this requires observing a special case: we manually must guard from confirmed users not being suspended by a spam check. Otherwise they will be immediately suspended the moment we confirm them, as the spam check will run again on save.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -229,7 +229,9 @@ class User < ApplicationRecord
     end
 
     # Mark the account as deleted but keep all data intact
-    event :hide do
+    # Only to be used in tests. There's currently no production
+    # use for this transition.
+    event :mark_deleted do
       before do
         close_open_issues
       end
--- a/test/controllers/confirmations_controller_test.rb
+++ b/test/controllers/confirmations_controller_test.rb
@@ -177,7 +177,7 @@ class ConfirmationsControllerTest < ActionDispatch::IntegrationTest
     post users_path, :params => { :user => user.attributes }
     confirm_string = User.find_by(:email => user.email).generate_token_for(:new_user)
 
-    User.find_by(:display_name => user.display_name).hide!
+    User.find_by(:display_name => user.display_name).mark_deleted!
 
     # Get the confirmation page
     get user_confirm_path, :params => { :display_name => user.display_name, :confirm_string => confirm_string }
@@ -212,7 +212,7 @@ class ConfirmationsControllerTest < ActionDispatch::IntegrationTest
     user = build(:user, :pending)
     post users_path, :params => { :user => user.attributes }
 
-    User.find_by(:display_name => user.display_name).hide!
+    User.find_by(:display_name => user.display_name).mark_deleted!
 
     assert_no_difference "ActionMailer::Base.deliveries.size" do
       perform_enqueued_jobs do
--- a/test/integration/oauth2_test.rb
+++ b/test/integration/oauth2_test.rb
@@ -246,7 +246,7 @@ class OAuth2Test < ActionDispatch::IntegrationTest
     get api_user_preferences_path, :headers => auth_header
     assert_response :forbidden
 
-    user.hide!
+    user.mark_deleted!
 
     get api_user_preferences_path, :headers => auth_header
     assert_response :forbidden
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -432,16 +432,16 @@ class UserTest < ActiveSupport::TestCase
     assert_equal "resolved", issue.reload.status
   end
 
-  def test_hide
+  def test_mark_deleted
     user = create(:user)
-    user.hide
+    user.mark_deleted
     assert_equal "deleted", user.status
   end
 
-  def test_hide_closes_issues
+  def test_mark_deleted_closes_issues
     user = create(:user)
     issue = create(:issue, :reportable => user)
-    user.hide
+    user.mark_deleted
     assert_equal "deleted", user.status
     assert_equal "resolved", issue.reload.status
   end


pablobm:
@tomhughes - Re: ability to suspend confirmed users, that makes sense. I have added a commit at the start to remove this restriction. A couple of notes:

- As a side effect, a test was creating 50 notes for a user, then confirming them. The removal of the restriction triggers a spam suspension and resulting in failure. I have worked around this by changing the factory, to avoid this happening in other cases. But perhaps this is a bit heavy handed and implicit? Let me know what you think.
- I have removed `suspend_if_possible!`, but now I'm thinking it might still be necessary. The user listing offers a couple of footguns to suspend already suspended users, which would raise an exception. I'm thinking of reinstating `suspend_if_possible!`. Thoughts?

pablobm:
As for the workaround for the heatmap issue, I'm going to look into that a bit more now.

pablobm:
> I have worked around this by changing the factory, to avoid this happening in other cases.

Actually, scratch that. It caused something else to fail. Ended up going down the more explicit, specific route.

pablobm:
I have been looking a bit more into the heatmap issue. I think it's unrelated to this PR, a pre-existing lingering issue. Specifically I'm observing the following sequence of events on test failures:

1. The test ends and calls `teardown`.
2. [`Capybara.reset_session!` is called](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/system_testing/test_helpers/setup_and_teardown.rb#L16).
3. `Users::HeatmapsController#show` begins.

Hypothesis: the Ajax request for the heatmap fires late, at a point when the cookie jar has been discarded. Therefore there's no session when the controller action runs, and `@heatmap` is not set.

I also noticed that [`HeatmapsController#show` checks for `@user` to be set](https://github.com/openstreetmap/openstreetmap-website/blob/6ca74538293c56da5309fdde0a2b956207126de2/app/controllers/users/heatmaps_controller.rb#L16). If it's not set, then `@heatmap_data` won't be set and the view will fail anyway. From here, I think it makes sense to check for `@heatmap_data` in the view, as the controller does allow for the possibility of not setting it.

Summary: PR good, plz approve :pleading_face: (Happy to be contradicted!)

tomhughes:
> I also noticed that [`HeatmapsController#show` checks for `@user` to be set](https://github.com/openstreetmap/openstreetmap-website/blob/6ca74538293c56da5309fdde0a2b956207126de2/app/controllers/users/heatmaps_controller.rb#L16). If it's not set, then `@heatmap_data` won't be set and the view will fail anyway. From here, I think it makes sense to check for `@heatmap_data` in the view, as the controller does allow for the possibility of not setting it.

Or should the controller error if asked for a heatmap when there's no user?

Note that `@user` there is derived from the arguments to the controller - it's not the logged in user. In fact there's no need to be logged in at all to call that URL so I'm not sure your theory about the missing session works?

pablobm:
Ah, you are right about the provenance of `@user`, I got a bit mixed up when putting together the writeup. But I also missed to include a detail: it's `current_user&.administrator?` that is the problem. In this test, the administrator is the current user. The target user is suspended, so it's not `visible?` and only an administrator can see the heatmap, therefore a `current_user` is required, missing, and `@heatmap_data` is not set.

Also, either way the controller allows for `@heatmap_data` not to be set.

tomhughes:
> it's `current_user&.administrator?` that is the problem. In this test, the administrator is the current user. The target user is suspended, so it's not `visible?` and only an administrator can see the heatmap, therefore a `current_user` is required

That uses safe navigation so it won't fail in itself if `current_user` is not defined though. It will however fail while rendering the view as does anything else that doesn't generate the heatmap data.

I've now opened #6591 to deal with that properly by 404ing requests for unavailable heatmaps.

Given that can we remove the attempted fix from this PR and then I'm happy to merge it.

pablobm:
@tomhughes - Will do :+1: I'm also going to bring back `suspend_if_possible!`, as this change makes it possible to suspend suspended users, which would raise an error.

pablobm:
Ready now. Incidentally, both the [Docker](https://github.com/openstreetmap/openstreetmap-website/actions/runs/19964725676/job/57253460582?pr=6564) and the [Ruby 3.3](https://github.com/openstreetmap/openstreetmap-website/actions/runs/19964725670/job/57253461847?pr=6564) workflows fell prey to the heatmap issue this time.

hlfan:
@gravitystorm these workflows should now run through with the merge of #6591.

test/controllers/diary_entries_controller_test.rb
```diff
@@ -476,8 +476,11 @@ def test_index_paged
 
   def test_index_user_paged
     user = create(:user)
-    create_list(:diary_entry, 50, :user => user)
-    user.confirm!
+    num_entries = 50
+    build_list(:diary_entry, num_entries, :user => user) do |entry, i|
+      entry.created_at = (num_entries - i).days.ago
+      entry.save!
+    end
```
tomhughes: What's this about, and what does it have to do with the commit it's part of?

pablobm: Hm, there's a reason, but on second thought it was a bad idea.

This is here because the test was failing once we allowed transitioning from `confirmed` to `suspended`. I didn't think it through and filed it under "issue with the test", but instead the test is showing why this can be a problem.

I'll elaborate on the main thread so that it's more visible for future readers.

pablobm:
There's a problem with allowing `confirmed` users to transition to `suspended`.

Currently, there's a `User.after_save :spam_check`. This in turn looks like follows:

https://github.com/openstreetmap/openstreetmap-website/blob/398660d1fa0007c9a0e65d26fb6b4a8e4a422da8/app/models/user.rb#L404-L406

So we enter this problem:
1. A legit user is suspended automatically by the spam check (a false positive).
2. An admin confirms the user to reactivate them and protect them from future spam checks.
3. As the user is confirmed, they are saved... and immediately suspended again as the spam check returns a new false positive.

Ways around this issue:
1. We add an additional `&& !confirmed?` to the spam check condition, so that it specifically doesn't trigger on confirmed users. They can still be suspended/unconfirmed manually.
2. We return to disallowing the transition `confirmed -> suspended`.
3. We change the logic completely.
4. Other?

I'm going to go for option 1, but please let me know if you think differently.

tomhughes:
So the intention was that a confirmed use was exempt from auto-suspension which was originally implemented by only considering users in state `active`.

When @gravitystorm introduced the aasm state machine in 1a11c4dc19 he changed that to use `may_suspend?` relying on the fact that the state machine didn't allow the transition from confirmed to suspended.

app/views/users/show.html.erb
```diff
@@ -197,24 +203,6 @@
...
-              <li>
-                <%= link_to t(".hide_user"), user_status_path(@user, :event => "hide"), :method => :put, :data => { :confirm => t(".confirm") } %>
-              </li>
-            <% end %>
-
-            <% if @user.may_unhide? %>
-              <li>
-                <%= link_to t(".unhide_user"), user_status_path(@user, :event => "unhide"), :method => :put, :data => { :confirm => t(".confirm") } %>
-              </li>
-            <% end %>
```
tomhughes, line 211: I was about to merge this but was just doing some final testing and realised that removing this is a problem as it means there is no easy way for me to reopen an account that has been self-deleted by a user....

pablobm: Do you mean self-deletion (which scrubs personal data), or "hiding" by an admin (which leaves data intact)?

If actually talking about self-deletion, I didn't know that was a use case. When is that desired?

If talking about "hiding", I was going with admins manually suspending all hidden users directly on the DB, as per the PR description:

> Note that we'll be left with pre-existing "hidden" users. After discussing with @Firefishy, these users can be moved to the "suspended" state by the Ops team with a script, but it's not a priority.

You did raise a question about the feasibility of this at https://github.com/openstreetmap/openstreetmap-website/issues/6546#issuecomment-3586714352, but then accepted it as non-blocking, if I understood correctly.

If however you'd still like to be able to do this manually, I can bring back the "unhide". Perhaps only for the user list, to reduce the number of options in user/show?

Or am I missing something?

tomhughes: I'm talking about self-deletion typically - where we get an email from somebody complaining that they can't get into their account and/or can't create an account because the email is taken and on investigation we find they have an account that has been self-deleted and we want to reopen it.

pablobm: Happy to restore the "unhide" action, but I'm still a bit confused. When a user self-deletes, their data is scrubbed. How do admins know that the scrubbed users is the same?

Although... I'm seeing now that `email` is not scrubbed (`new_email` is). So I'm guessing that's how these accounts are identified? Do admins "unhide" these users, leaving them to log in and set their (hitherto scrubbed) username, etc themselves?

If the above is right, I would have thought this would be a GDPR issue, not scrubbing their email. I'm assuming this has been discussed and deemed acceptable?

Are you ok with me restoring only the "unhide" in user/list, leaving the action out of user/show?

tomhughes: Yes the email is currently preserved (see #3488 for more on that) so that is how I find them and also what stops them just creating a new account. I then reopen the account and tell them to do a password reset.

I think we should restore it though probably not using the `unhide` name.

pablobm: Restored "unhide" to users/lists. More detail in the main thread.

pablobm:
I have added an "undelete" action in the users/lists page
- This is the same as the pre-existing "unhide", but moved out of users/show to be in the list instead. The intention is to unclutter the users/show page. Also I assume that this is the most useful place, as admins find self-deleted through this page (assumption danger here, please let me know if not).
- The action filters users by `:status => "deleted"` before applying the change, as a precaution.
- The rename from "unhide" to "undelete" is as suggested by @tomhughes at https://github.com/openstreetmap/openstreetmap-website/pull/6564/changes#r2618873980. I chose that name as seems the most descriptive, but open to others.

pablobm:
Hm, broke something. Looking into it 👀

app/views/users/show.html.erb
tomhughes: I think there's been a horrible misunderstanding here... You've added undelete to the bulk interface which never had it before, rather than restoring it to the use profile which was what I intended to happen?

pablobm: Well, hopefully not "horrible" as no lives or data were lost :slightly_smiling_face: 

I did ask twice about moving this to users/lists (see earlier in this thread). The logic is:
- The intent of this PR is to clarify and unclutter the admin options in users/show.
- When a user asks to revert a deletion, I can only guess that admin find them via users/lists, searching by email and potentially filtering by status=Deleted.

Therefore users/lists might be a better way to place this. Of course this is based in assumptions, and I'll be happy to move this back to users/show.

pablobm:
Ready for a new review:
- Moved the "undelete" action out of users/lists, back to users/show, as requested by @tomhughes.
- Added a test for "unhide" (first commit, before renaming to "undelete") as it was missing before.
- Renamed `hide` as `mark_deleted` in the code, which is more descriptive.

Question:  if this PR is merged, the status transition `mark_deleted` will have no production use left. It'll be there only for the benefit of some tests that I don't think would represent reality either. Should transition and tests be removed altogether?

~CI error appears to be a flaky test (see https://github.com/openstreetmap/openstreetmap-website/actions/runs/20302680370/job/58311962725?pr=6564). Rerunning~

tomhughes:
Yes I think I was wondering about `mark_deleted` when I looked at it the other day but I didn't investigate fully.

tomhughes:
Thanks for all the work on this. I think you're probably right that `mark_deleted` should be removed but I'm going to merge this and we can deal with that later.

PR #6565, opened 2025-11-25
include iD as npm package

tyrasd:
Instead of pushing static builds of iD into this repository for every release of iD, this loads the specified build of iD from npm [`@openstreetmap/id`](https://www.npmjs.com/@openstreetmap/id). In the future, updating iD will only require bumping the version number in the `package.json` file and a `yarn install` on the deployment.

As this is including the exact same version of iD as the one that is currently deployed, this PR should not change anything for the user of iD on the website.

One thing that I think is not very elegant is how I ported over that one postprocessing step which was [previously](https://github.com/openstreetmap/openstreetmap-website/pull/6565/files#diff-97d92457f6d9fdf1f937ddfe97aad5f7313ac0cee8894400f984d86f9dfa462cL30-L34) done in the Vendorfile: The URLs of iD's cursors are defined in the CSS and need to be rewritten to use the correct respective `asset_path`.[^1] I don't know enough about rails to say whether [my solution here](https://github.com/openstreetmap/openstreetmap-website/pull/6565/files#diff-0db45d17b7f6cb046a18fef25dd3eab756c833bf4086b456030715832534b9afR2-R4) makes any sense at all, but my gut feeling says there must be a way to do this a bit less hackily. Please let me know if you have suggestions for that. Potentially, I would also looking into maybe addressing this on the iD side in one of the upcoming releases (e.g. by replacing the cursor URLs with static inline data-URLs during iD's own build scripts).

[^1]: unfortunately, in the current version of iD, those URLs are kind of hardcoded and not affected by iD's mechanism to be able to read the `assetMap()` to resolve the asset URLs dynamically at runtime.

PS: Thanks @pablobm for helping me navigating the rails asset pipeline. :relaxed: 

PSS: Note to self:
* [x] update [release process documentation](https://github.com/openstreetmap/iD/blob/develop/RELEASING.md) on iD's repo.


tomhughes:
The alternative to trying to edit the CSS is to put the iD images on the asset search path by editing `config/initializers/assets.rb` to add this at the end:

```ruby
Rails.application.config.assets.paths << Rails.root.join("node_modules/@openstreetmap/id/dist")
```

Then sprockets will manage to find them from the paths in the iD CSS and rewrite the URLs automatically.

It's a bit nasty because they're on the path for everything, but probably more robust than trying to do non-syntax aware text replacement on the CSS data.

hlfan:
Independent of that, I'd like the asset management to be its own commit with ‎Vendorfile‎ and CSS adaptation before the main vendor -> npm migration.

tomhughes:
That doesn't make any sense with the current scheme as the two are not independent - neither works without the other.

It probably is OK with my proposed solution.

app/assets/javascripts/id.js
```diff
@@ -22,7 +22,7 @@ document.addEventListener("DOMContentLoaded", function () {
 
     const id = idContext
       .embed(true)
-      .assetPath("iD/")
+      .assetPath("@openstreetmap/id/dist/")
```
pablobm, line 25: Wrong path here, at least in my machine where the correct one appears to be the following:

```suggestion
      .assetPath("assets/@openstreetmap/id/dist/")
```

tyrasd:
@tomhughes thanks, that works. I've updated the PR accordingly.

app/assets/javascripts/id.js
tomhughes: Really? It seemed to be working for me when I tried it last night and I didn't notice anything failing to load...

tyrasd: That's odd. For me it only works as expected if I omit the `assets/` part here. This path needs to match the keys of the dictionary provided in `data-asset-map` (see screenshot below), which comes from the `asset_helper.rb` via `id.html.erb`. That one does have the assets listed without the `assets/` "prefix". :thinking: 

image

If I use `.assetPath("assets/…`, iD still appears to run fine, but it will load the un-hashed assets[^1], which is not what we want, is it?

[^1]: e.g. in a local dev environment `http://127.0.0.1:3000/assets/@openstreetmap/id/dist/img/pattern/farmland.png` instead of `http://127.0.0.1:3000/assets/@openstreetmap/id/dist/img/pattern/farmland-a59cbecf2b7bb338812ee0b4e25dbf79183ac809f5818451850922142121c3d9.png`

pablobm: That's why I qualify with "at least in my machine" as I'm always a bit worried that it might be _one of those things_ :sweat_smile: Would you be able to test in the dev server?

@tyrasd - How about your machine?

tyrasd: @pablobm see my reply above (I think we have posted almost at the same moment). I _think_ the current version is correct. It works on my machine. :blush:

pablobm: Looking into it, as it's indeed different for me. For the record, this is the HTML I get served with the code as in this PR, with an empty asset map:

```
<div id="id-container" data-token="9PXRfYK_vnT-6CZw4InO1XTwjnxqPnqzQkfQFvKw2kM" data-locale="en" data-asset-map="{}"></div>
```

And visibly broken UI:

Visibly broken iD editor, with missing translation and no aerial imagery

tomhughes: If the asset map isn't being built than that means that `AssetHelper#assets` is failing - it is looking for things in `Rails.application.assets_manifest.assets` whose key starts with the specified directory...

tomhughes: Have you built production assets at some point? Try cleaning them out as mixing dev mode and production assets can do weird things...

pablobm: Er... now it works? It might have been the precompiled assets thing, but I'm not 100% sure because of the order of events here. Anyway, success :stuck_out_tongue_closed_eyes:

pablobm:
There's a symlink at `public/iD` that I think can be removed now.

lib/id.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module ID
-  LOCALES = Locale.list(Rails.root.join("vendor/assets/iD/iD/locales").entries.filter_map { |p| p.basename.to_s[/(.*)\.min\.json/] && Regexp.last_match(1) })
+  LOCALES = Locale.list(Rails.root.join("node_modules/@openstreetmap/id/dist/locales").entries.filter_map { |p| p.basename.to_s[/(.*)\.min\.json/] && Regexp.last_match(1) })
```
pablobm: I know this is just the same what was there before but, since we are at it, I think it's a good opportunity to refactor this a bit:

```suggestion
  ID_PATH = Rails.root.join("node_modules/@openstreetmap/id")
  LOCALES = Locale.list(Rails.root.join(ID_PATH, "dist/locales").entries.filter_map { |p| p.basename.to_s[/(.*)\.min\.json/] && Regexp.last_match(1) })
```

I would go further than that, and perhaps will, but for the purposes of this PR I think this would be sufficient.

tyrasd:
(updated to remove the dead symlink and split the variables into ID_PATH / LOCALES.)

tomhughes:
It think this looks good now, thanks.

Once it's merged I'll let it go to https://master.apis.dev.openstreetmap.org/ first to make sure a production deployment is good before deploying it to the main site.

pablobm:
@tyrasd - Remember to update the iD release docs :wink: Mentioning because it appears as a checkbox in the description here.

gravitystorm:
Thanks @tyrasd , this is great and I don't envy you the work that went into the asset pipeline changes!

Firefishy:
https://en.osm.town/@osm_tech/115623771973315977 to celebrate the great work! Thank you.

PR #6569, opened 2025-11-27
Allow meta description and search title to be translated

### LINKED ISSUE #6527: Search and description meta tags are unlocalizable
The `search` and `description` meta tags on the homepage are hard-coded in English:

https://github.com/openstreetmap/openstreetmap-website/blob/052f0bc356806c56d3b5779c97ede17c40909a5b/app/views/layouts/_meta.html.erb#L23-L24

This results in browsers like Firefox associating the OpenSearch description file with an English title:

Thêm công cụ tìm kiếm “OpenStreetMap Search”

Sometimes search engines like Google turn up URLs that are explicitly non-English, such as with the `?locale=` parameter set, but still have the description in English:

[Trazas GPS públicas: OpenStreetMap is the free wiki world map.](https://www.google.com/search?q=https%3A%2F%2Fwww.openstreetmap.org%2F%3Flocale%3Des&ie=utf-8&oe=utf-8&hl=es)

gravitystorm:
Fixes #6527

This uses the same description as the default "og:description" for the site, to reduce the number of translations. Of the two alternative previous descriptions, the one in "og:description" more closely matches the language used on the about page, so let's go with that.

To be more explicit, here's the change in the html:

```diff
<link rel="search" type="application/opensearchdescription+xml" title="OpenStreetMap Search" href="/assets/osm-42b7b3fbcee2193e455a773db6cd3d34a2f48ca94547fed54901dd9d8307b02b.xml">
- <meta name="description" content="OpenStreetMap is the free wiki world map.">
+ <meta name="description" content="OpenStreetMap is a map of the world, created by people like you and free to use under an open licence.">
  <meta property="og:site_name" content="OpenStreetMap">
  <meta property="og:title" content="OpenStreetMap">
  <meta property="og:type" content="website">
  <meta property="og:url" content="http://localhost:3000/">
  <meta property="og:description" content="OpenStreetMap is a map of the world, created by people like you and free to use under an open licence.">
```
Note that this makes `<meta name="description">` become the same as `<meta property="og:description">` which I think is reasonable. The `<meta property="og:description">` has been translatable for a long time.

### DIFF
--- a/app/views/layouts/_meta.html.erb
+++ b/app/views/layouts/_meta.html.erb
@@ -20,8 +20,8 @@
 <% if Settings.key?(:publisher_url) -%>
 <%= tag.link :rel => "publisher", :href => Settings.publisher_url %>
 <% end -%>
-<%= tag.link :rel => "search", :type => "application/opensearchdescription+xml", :title => "OpenStreetMap Search", :href => asset_path("osm.xml") %>
-<%= tag.meta :name => "description", :content => "OpenStreetMap is the free wiki world map." %>
+<%= tag.link :rel => "search", :type => "application/opensearchdescription+xml", :title => t(".openstreetmap_search"), :href => asset_path("osm.xml") %>
+<%= tag.meta :name => "description", :content => t("layouts.intro_text") %>
 <%= opengraph_tags(@title, @opengraph_properties || {}) %>
 <% if flash[:matomo_goal] -%>
 <%= tag.meta :name => "matomo-goal", :content => flash[:matomo_goal] %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1874,6 +1874,8 @@ en:
     header:
       select_language: Select Language
       loading: Loading...
+    meta:
+      openstreetmap_search: OpenStreetMap Search
     select_language_button:
       title: Select Language
     offline_flash:


tomhughes:
Looks good to me, thanks.

hlfan:
I think this is why the tests fail now also on the master:

https://github.com/openstreetmap/openstreetmap-website/blob/b14ac16672c96e5b764fbc2a9c36a40307347ed8/.github/workflows/docker.yml#L37

1ec5:
Yep, that’ll do it. I think this line is just verifying that the page loads and is the homepage. I think we can update the line to grep for the new text or even just “OpenStreetMap”.

tomhughes:
Oh right. I generally just ignore docker failures because it's so unreliable.

1ec5:
#6577

gravitystorm:
Thanks @hlfan and @1ec5 for the fix, and I'm sorry for breaking this in the first place. I hadn't even noticed the failure (and given the work that I was doing, it never crossed my mind that it would affect the Docker build!). 

I agree with @tomhughes , I find the docker build is a bit unreliable, and I'm often manually re-running it on PRs after finding out that it's just some generic CI failure rather than something we can act on. I would say most weeks I'm either re-running it on master or on a PR.

pablobm:
Re: flakiness of the Docker workflow. Is this normally triggered by the Selenium-driven tests? If so, https://github.com/openstreetmap/openstreetmap-website/pull/6497 could help alleviate the issue.

Issue #6485, opened 2025-10-29
Context menu icons are inconsistent with icons used elsewhere


gravitystorm:
Our [updated context menu](https://github.com/openstreetmap/openstreetmap-website/pull/6410) ships with icons for the various menu items. However, these are inconsistent with the icons used elsewhere in the UI. We should consider whether they can be unified - either by changing the menu, or changing the icons used elsewhere, or perhaps both. Here's the list of inconsistencies:

## Directions

The directions icons used in the sidebar and on the map show pins with play and stop icons. The context menu has a pin with a circle for the start, and a flag for the finish.

<img width="672" height="355" alt="Image" src="https://github.com/user-attachments/assets/515c9a6d-b5cb-472d-8e2a-39906389c69d" />

## Add a Note

The Add a Note menu item has a pencil, the map button has a speech bubble with a plus sign, and the map icon is a blue pin with a plus sign

<img width="672" height="123" alt="Image" src="https://github.com/user-attachments/assets/179f6cbd-d28a-4959-8d62-f3228b005df0" />

## Query features

The Query features menu item is a question mark in a circle, the map button is an arrow with a small question mark. Note that the cursor also changes (using CSS `cursor: help`) but this is platform specific - some platforms show the arrow and small question mark, others show different icons.

<img width="483" height="79" alt="Image" src="https://github.com/user-attachments/assets/257710c4-1782-4715-afb2-a0be81f2a54b" />

Hiteshpy07:
hey @gravitystorm , i noticed all the inconsitancys you have listed , but now it is resolved i guess, i cannot spot them any more , if there is any more , i can fix it

pablobm:
@Hiteshpy07 - What did you check? There has been no change and the inconsistencies are still present as listed in the issue description. Therefore you can look into them 🙂

aNsHuL5217:
Hi @gravitystorm  and @pablobm ,

I have checked the website and confirmed that the icon inconsistencies are still present as described in the issue.

Since there hasn't been any update from the previous commenter for a month, I would like to pick this up. I plan to update the context menu icons to match the standard icons used in the Sidebar and Map buttons for consistency. Could you please assign this to me?

HolgerJeromin:
> Could you please assign this to me?

We do not work with assignments.
Just go ahead

pablobm:
Specific questions raised at https://github.com/openstreetmap/openstreetmap-website/pull/6571:

[...]

PR #6221, opened 2025-07-22
Add play and stop markers for directions

hlfan:
Closes #2787 by adding media control based marker variants for the directions.
This is inspired by the markers from https://brouter.de/brouter-web/.

PR #6571, opened 2025-11-30
Fix inconsistent context menu icons (Issue #6485)

aNsHuL5217:
As described at #6485

This PR updates the context menu icons to match the icons used elsewhere in the UI (Sidebar and Map buttons), resolving the inconsistencies described in the issue.

**Changes:**
- **Directions from here:** Changed from `bi-geo-alt` (Pin) to `bi-play-fill` (Play Triangle) to match the Sidebar.
- **Directions to here:** Changed from `bi-flag` (Flag) to `bi-stop-fill` (Stop Square) to match the Sidebar.
- **Add a note here:** Changed from `bi-pencil` (Pencil) to `bi-chat-text-fill` (Speech Bubble) to match the Sidebar note button metaphor.
- **Query features:** Changed from `bi-question-circle` to `bi-question-lg` (Plain Question Mark) to match the Map control button style.
- BEFORE: 
507158610-515c9a6d-b5cb-472d-8e2a-39906389c69d

-AFTER:
Screenshot 2025-11-30 at 8 03 21 PM

### DIFF
--- a/app/assets/javascripts/index/contextmenu.js
+++ b/app/assets/javascripts/index/contextmenu.js
@@ -34,7 +34,7 @@ OSM.initializations.push(function (map) {
   const contextmenuItems = [
     {
       id: "menu-action-directions-from",
-      icon: "bi-geo-alt",
+      icon: "bi-play",
       text: OSM.i18n.t("javascripts.context.directions_from"),
       callback: () => {
         const params = new URLSearchParams({
@@ -46,7 +46,7 @@ OSM.initializations.push(function (map) {
     },
     {
       id: "menu-action-directions-to",
-      icon: "bi-flag",
+      icon: "bi-stop",
       text: OSM.i18n.t("javascripts.context.directions_to"),
       callback: () => {
         const params = new URLSearchParams({
@@ -61,7 +61,7 @@ OSM.initializations.push(function (map) {
     },
     {
       id: "menu-action-add-note",
-      icon: "bi-pencil",
+      icon: "bi-chat-text",
       text: OSM.i18n.t("javascripts.context.add_note"),
       callback: () => routeWithLatLon("/note/new")
     },
@@ -76,7 +76,7 @@ OSM.initializations.push(function (map) {
     },
     {
       id: "menu-action-query-features",
-      icon: "bi-question-circle",
+      icon: "bi-question",
       text: OSM.i18n.t("javascripts.context.query_features"),
       callback: () => routeWithLatLon("/query")
     },


pablobm:
Thank you for the PR, @aNsHuL5217. It gives us a starting point to discuss solutions. For completeness, here's a screenshot with all the icons, in context, as they appear in production at the moment:

Navigation menu, context menu, and right-hand sidebar menu, with icons as per currently on production

And this is the suggestion offered by your PR:
Navigation menu, context menu, and right-hand sidebar menu, with icons as suggested by this PR

Below are my thoughts. Please don't interpret them as "make these changes", but rather as an invitation for everyone to give their opinion.

### Fill vs no fill
The first thing that catches my eye is the choice of `*-fill` icons in an otherwise sparse context. I feel they grab attention too much.

This is an alternative without the "no fill" version of the icons. In my view, the icons are better integrated in their context like this:
Navigation menu, context menu, and right-hand sidebar menu, with icons similar to those in this PR, but using the no-fill version for three of them

### "Stop" icon
The second thing is that I'm not sure about the choice of the square "stop" icon for "Directions to here". My feeling is that it doesn't convey the meaning and instead invites to "stop" something (I don't know what).

I understand that you wanted to match the "stop" in the red marker. This is one of those things where perhaps both icons (in the context menu and in the red marker) should change. To what I don't know.

### Different note icon
This PR uses a speech bubble icon for the "note" action. This is an improvement over the pencil icon, in that it's closer to the icon in the side menu. However it's still quite different and it may still lead to confusion.

The one used in the sidebar lives at [`app/assets/images/map-controls/note.svg`](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/images/map-controls/note.svg). Can this one be used there instead? Or is this another case of changing both icons?

### Different query icon
A bit similar to the previous point, but in this case I'm not sure if the new choice of icon is an improvement. This goes from a question mark enclosed in a circle, to a freestanding question mark. I don't see either better than the other, but in both cases they are different from the one in the sidebar, which is [app/assets/images/map-controls/query.svg](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/images/map-controls/query.svg). Again: should we use the same one, or should we change both?

aNsHuL5217:
@pablobm  Thank you for the detailed feedback!

I agree that the filled icons were visually too heavy for this context. I have just pushed an update that switches all icons to their **outline ("no fill") versions** (`bi-play`, `bi-stop`, `bi-chat-text`, `bi-question`). This should make them blend in much better with the existing UI.

**Regarding the custom SVGs (`note.svg` and `query.svg`):**
The current context menu code is designed to use Bootstrap Icon CSS classes (`bi-*`). To use the actual SVG files from `app/assets/images/`, I would likely need to refactor `_createMenuItem` to support `` tags or inline SVGs.

Would you prefer I attempt that refactor now, or are these outline icons a close enough match to proceed for this PR?

tomhughes:
Given we have recently been trying to move towards using bootstrap icons I'm a bit reluctant to start going in the other direction but equally it's probably hard, if not impossible, to use the bootstrap icons in the context of actual map markers so I'm not sure what the right answer is - it might be somewhere where we have to accept duplication and just try and look close to each other.

The question of whether to change what markers we use for the start/end of a route is maybe something we should leave and deal with separately if somebody has a good idea for something better?

aNsHuL5217:
@tomhughes Thanks for the guidance!

That makes perfect sense. I will stick with the Bootstrap icons to keep the codebase consistent.

I have already updated the PR to use the **outline** versions of the icons (e.g., `bi-play` instead of `bi-play-fill`) as discussed with @pablobm. This keeps them visually lighter and "close enough" to the map markers without adding complexity.

I believe this is ready for a final review.

pablobm:
Thank you for the update @aNsHuL5217. I think those icons look better now.

I am ok with the current result. I think this is a step in the right direction, and the questions I raised are each complex and can take time to resolve so they are better worked separately.

A couple of details:
- There are a number of changes that are unrelated to the substance of the PR. I think you ran a linter and it changed more things than needed? Please remove those.
- After that, can you please squash all the commits into a single one? I'm realising this is not that clear in the contribution guidelines, we'll have to update that.

Also a general piece of advice for future contributions in any project: avoid using `master` (or `main` or whatever a project uses) as your working branch, as that makes it a bit more difficult to work with. Don't worry about that for this PR!

Finally, just out of curiosity if it's ok: how are you running the project in your computer? Docker, rbenv, other? Just to have an idea of what people are using, particularly those new to the project.

pablobm:
Just realised, the PR description says it "fixes" the original issue, but it doesn't completely. @aNsHuL5217, can you please remove that line? Perhaps replace it with "As described at #6485". When this PR is merged, the original issue should not be closed automatically as there will still be open questions.

pablobm:
@aNsHuL5217 - I think you pushed the wrong thing.

aNsHuL5217:
@pablobm Sorry about that! I had a sync issue that accidentally included upstream changes.

I have performed a hard reset and force-pushed the clean version. The PR now contains exactly **1 commit** and changes **1 file** (just the icon updates), as requested.

To answer your question: I am running the project via a **manual installation on macOS** (Apple Silicon), following the standard guide.

hlfan:
> My feeling is that it doesn't convey the meaning and instead invites to "stop" something (I don't know what). ... This is one of those things where perhaps both icons (in the context menu and in the red marker) should change. To what I don't know.

My original draft for #5764 had the navigation and flag icons for the routing markers. But since we're also using the navigation icon for the location button (instead of a crosshair) this was also irritating. Eventually I pivoted to the media control icons in #6221, as in combination these feel more consistent across locales. But the play and stop icons alone do seem weird without context.

tomhughes:
I think this looks good now - thanks for all the work.

PR #6574, opened 2025-12-01
Restrict building editing until zoom level 19

ariffaisal-github:
## Summary
Adds a zoom-level restriction for building editing in the iD editor. When the user is zoomed out (typically <19), building features are auto-hidden via the existing rendererFeatures culling system. Roads remain visible and editable at all zoom levels.

## What’s Changed
- Uses the iD renderer’s feature rules to restrict building visibility when feature density is high.
- Behavior:
  - zoom < 19 → buildings hidden & non-editable
  - zoom ≥ 19 → buildings visible & editable

## Why
- Improves editing precision for buildings.
- Reduces accidental building edits when zoomed out.
- Aligns editable behavior with intended UX for detailed geometry.

## Testing
- Zoomed out below 19: buildings disappear, roads still editable.
- Zoomed in to ≥19: buildings appear and are editable again.


tomhughes:
Changes to iD should be proposed at https://github.com/openstreetmap/iD/ not here.

Nothing in the `vendor` directory is ever edited here and in fact since last week iD isn't even copied into that directory.

PR #6576, opened 2025-12-01
Fix some dark mode colors desyncing

### LINKED ISSUE #6537: Poor contrast on context menu in light theme

hlfan:
- Removed the "cut-out" part from the GitHub icon, so the Invertocat is the `path` like the apple.
- Synced the used preferred color scheme to Bootstrap to prevent it from looking like this when the OSM user preference isn't the same as the browser one:
  image
- Used the dark grey that [GitHub provides](https://brand.github.com/foundations/logo) in https://brand.github.com/GitHub_Logos.zip
- Removed the overrides to invert the `.dropdown-item:active` style. This closes #6542 and fixes #6537.

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1011,11 +1011,6 @@ img.trace_image {
 /* Rules for contextmenu */
 
 #map-context-menu {
-  .dropdown-item {
-    &:hover, &:active {
-      background-color: $gray-200;
-    }
-  }
   &.cm_dropdown {
     position: absolute;
     z-index: 1000;
@@ -1024,15 +1019,5 @@ img.trace_image {
     position: absolute;
     display: block;
     min-width: 200px;
-  }
-}
-
-@media (prefers-color-scheme: dark) {
-  #map-context-menu {
-    .dropdown-item {
-      &:hover, &:active {
-        background-color: $gray-700;
-      }
-    }
   }
 }


tomhughes (commented):
As fast as I can tell the icon related changes seem fine though it would have been helpful to have a bit more information about what you were doing as the linked issue is only about the menu.

The menu change does fix the current issues in dark mode, but the result has very little contrast between the background and the highlight, at least on this laptop, and the contrast in light mode is also reduced to the point where you can hardly see the bar.

hlfan:
I updated the JS to increase the contrast.

pablobm:
Where did you see about the special grey for the GitHub log? I can't see it on the page you linked. All I see is the phrase "The Invertocat and our wordmark should only appear in white, black, or in few cases grey".

On the topic of the contrast in the context menu, I see it diminished, before even hovering:

Production | This PR
--- | ---
Context menu as in production | Context menu as in this PR, with slightly but noticeably decreased contrast

app/assets/javascripts/index/contextmenu.js
```diff
@@ -177,7 +177,7 @@ class ContextMenu {
   }
 
   _render(items) {
-    const $menuList = $("<ul>").addClass("dropdown-menu show shadow cm_dropdown_menu");
+    const $menuList = $("<ul>").addClass("dropdown-menu list-group show shadow cm_dropdown_menu");
```
gravitystorm: I don't think it's a good idea to try to make this styled as *both* a dropdown menu and also as a list-group. It shouldn't be both, these are two different bootstrap components.

```diff
-    const $menuList = $("<ul>").addClass("dropdown-menu list-group show shadow cm_dropdown_menu");
+    const $menuList = $("<ul>").addClass("dropdown-menu show shadow cm_dropdown_menu");
```

app/assets/stylesheets/common.scss
```diff
@@ -995,25 +995,25 @@ img.trace_image {
...
+  .github-logo path { fill: #1b1f24; }
 }
 
 /* Rules for contextmenu */
 
 #map-context-menu {
   .dropdown-item {
     &:hover, &:active {
-      background-color: $gray-200;
+      @extend :focus;
```
gravitystorm: How about we just don't override the hover and active states at all, and let bootstrap deal with them? If we leave it alone, it works fine in both dark and light modes, as far as I can tell.

```diff
 #map-context-menu {
-  .dropdown-item {
-    &:hover, &:active {
-      @extend :focus;
-    }
-  }
   &.cm_dropdown {
```

app/assets/javascripts/index/contextmenu.js
```diff
@@ -194,7 +194,7 @@ class ContextMenu {
     const $label = $("<span>").text(item.text);
 
     const $link = $("<a>")
-      .addClass("dropdown-item d-flex align-items-center gap-3")
+      .addClass("dropdown-item list-group-item-action d-flex align-items-center gap-3")
```
gravitystorm:
```diff
-      .addClass("dropdown-item list-group-item-action d-flex align-items-center gap-3")
+      .addClass("dropdown-item d-flex align-items-center gap-3")
```

hlfan:
> Where did you see about the special grey for the GitHub log? I can't see it on the page you linked. All I see is the phrase "The Invertocat and our wordmark should only appear in white, black, or in few cases grey".

That's what is provided in:
- the logo files zip download: https://brand.github.com/GitHub_Logos.zip
- the logo color scheme demonstration below the text you're citing: https://brand.github.com/foundations/logo#coloring-the-logo

gravitystorm:
Merged, thanks!

mmd-osm:
Regarding the context menu, I think the hovering is barely visible now, both in light and dark mode. What was the reason again to make this worse than before?

This is how it looks like here. I'm hovering over the second item.

{42959EEB-A53D-4697-8131-2D2E0045C4EB}

tomhughes:
Which I did point out and @hlfan said he had increased the contrast though I admit I don't see much difference myself from the original version.

gravitystorm:
We're talking about several different things here when it comes to contrast. There's the foreground-background contrast levels of the text (in various states, hover, active, dark mode etc), and the background-background contrast between hovered and unhovered states.

@mmd-osm what this PR did was remove all the attempts to twiddle things. We were starting to layer fixes on top of fixes. Instead, we now just use the [bootstrap dropdown component defaults](https://getbootstrap.com/docs/5.3/components/dropdowns/). 

The background hover highlighting is subtle, but that's coming from upstream. I wouldn't call the end result "worse" because it fixed various other problems.

If we want to change the backgrounds of hovered items in dropdown menus, i.e. override the bootstrap defaults, then let's discuss that and make it work for all the dropdown menus consistently, rather than trying to adjust this one separately.

mmd-osm:
> We were starting to layer fixes on top of fixes. Instead, we now just use the [bootstrap dropdown component defaults](https://getbootstrap.com/docs/5.3/components/dropdowns/).

I'm getting your idea to get rid of custom css logic, that's all fine. Still, if the resulting UI is subpar, these defaults may not be ideal.

>  then let's discuss that and make it work for all the dropdown menus consistently, rather than trying to adjust this one separately.

For real consistency, you would have to make the coloring match the other two places where we have popups today. They're using some green (`#7ebc6f`) as background color. I was a bit hesitant to go for that color for the new popup, but maybe that's an option? In any case, it has a significantly better background-background contrast than what we have now.

{B290971F-BD91-4AEB-943D-E2E10B215376}

{95110FEC-CCFD-4FB8-86D8-BDEBFA886EC2}

PR #6577, opened 2025-12-02
Update Docker test of homepage description

1ec5:
Fixed the Docker test failure mentioned in https://github.com/openstreetmap/openstreetmap-website/pull/6569#issuecomment-3598543256 by updating the string we’re grepping for.


gravitystorm:
Thanks! Merged.

PR #6583, opened 2025-12-03
Allow various hard-coded strings to be translated

gravitystorm:
This is part of a larger project to enable the ["No Hardcoded Strings"](https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#hardcodedstring) linter in erb_lint. Before I resolve some edge cases with that linter, I'm submitting this PR to fix some hardcoded strings that it found in our views.

For the `layouts.project_name.h1` key (see first commit) I checked every translation and validated that, if the translation existed, it matched the `layouts.project_name.title` key. This is true for all locales. I don't think there was ever a good reason to have two translations available for this.

### DIFF
--- a/app/views/browse/_relation.html.erb
+++ b/app/views/browse/_relation.html.erb
@@ -34,7 +34,7 @@
             <% end %>
             <div class="text-center my-2">
               <div class="spinner-border spinner-border-sm text-primary" role="status">
-                <span class="visually-hidden">Loading...</span>
+                <span class="visually-hidden"><%= t("browse.start_rjs.loading") %></span>
               </div>
             </div>
           </turbo-frame>
--- a/app/views/export/show.html.erb
+++ b/app/views/export/show.html.erb
@@ -2,7 +2,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
   <head>
     <meta charset="utf-8">
-    <title>OpenStreetMap Embedded</title>
+    <title><%= t ".title" %></title>
     <%= stylesheet_link_tag "embed", :media => "screen" %>
     <%= javascript_include_tag "embed" %>
   </head>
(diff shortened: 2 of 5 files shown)


pablobm (approved):
Looks good to me 👍 What are the edge cases?

gravitystorm:
> Looks good to me 👍 What are the edge cases?

* We use `&middot;` in various places, but that's not in the erb_lint entity ignore list. See also https://github.com/Shopify/erb_lint/pull/397 
* erb_lint gets confused when attempting to parse https://github.com/openstreetmap/openstreetmap-website/blob/acb1e73fa98e6cb7adf67f99e0be5891dcd4f6ea/app/views/browse/_relation.html.erb#L27-L35 - it ends up thinking that `src=` is text content instead of an html attribute. It might be possible to refactor that to be more parser friendly (and maybe easier for me to read too!), or we just ignore that file.

1ec5:
> For the `layouts.project_name.h1` key (see first commit) I checked every translation and validated that, if the translation existed, it matched the `layouts.project_name.title` key. This is true for all locales. I don't think there was ever a good reason to have two translations available for this.

Comparing the [h1](https://translatewiki.net/wiki/Special:Translations/Osm:Layouts.project_name.h1) and [title](https://translatewiki.net/wiki/Special:Translations/Osm:Layouts.project_name.title) messages over in Translatewiki.net, only [Persian](https://translatewiki.net/wiki/Osm:Layouts.project_name.h1/fa) has any difference between the two, leaving the title untranslated. Combining the messages should be fine.

pablobm:
The relation template looks like it could use a refactor :+1: 

Re: the middot/interpunct - Would it make sense to allow it to be translateable after all? To my eyes it does look good as the separator that I understand it's intended as. However I wonder if it might be confusing to readers from other cultures.

Looking at the relevant [Wikipedia page](https://en.wikipedia.org/wiki/Interpunct), I see it's commonly used across languages to separate words or parts of words. Some uses stand out to me, like the Japanese: "often used to separate transcribed foreign names or words written in katakana", or the Korean: "to denote a list of two or more words, similarly to how a slash (/) is used to juxtapose words in many other languages".

tomhughes:

PR #6587, opened 2025-12-04
Exclude non-http URLs from tag2link

### LINKED ISSUE #6407: website key without "http" links to the OSM website
When browsing a node or way, with the key "website" if the URL doesn't contain "http", but just "www", then the link directs to https://www.openstreetmap.org/node/555555/www.example.com
From what I remember, I think it used to not be like that and the links worked correctly.

### LINKED ISSUE #6586: linkifying should not indiscriminately split values at semicolons
### URL

https://master.apis.dev.openstreetmap.org/node/4295725477

See discussion at the bottom of https://github.com/MarcusWolschon/osmeditor4android/issues/3045

The feature was added in https://github.com/openstreetmap/openstreetmap-website/pull/6197/files#diff-f787e682da0df0dd52c310ccf1d92b90b1245e4a8951f6f677ed90c09a38a9b2L43

_No response_

hlfan:
Change back to the Rinku linkifier instead of indiscriminately linkifying via tag2link, which closes #6586 and closes #6407.

Forcing the HTTP also removes the unused telephone link template covered by `telephone_links` and wrongly relative templates like for `ref:wikimaginot` in https://osm.org/relation/19757246

### DIFF
--- a/test/helpers/browse_tags_helper_test.rb
+++ b/test/helpers/browse_tags_helper_test.rb
@@ -77,11 +77,18 @@ class BrowseTagsHelperTest < ActionView::TestCase
     html = format_value("email", "foo@example.com")
     assert_dom_equal "<a title=\"Email foo@example.com\" href=\"mailto:foo@example.com\">foo@example.com</a>", html
 
-    html = format_value("source", "https://example.com")
+    html = format_value("website", "https://example.com")
     assert_dom_equal "<a href=\"https://example.com\" rel=\"nofollow\" dir=\"auto\">https://example.com</a>", html
 
-    html = format_value("source", "https://example.com;hello;https://example.net")
+    html = format_value("website", "https://example.com;hello;https://example.net")
     assert_dom_equal "<a href=\"https://example.com\" rel=\"nofollow\" dir=\"auto\">https://example.com</a>;hello;<a href=\"https://example.net\" rel=\"nofollow\" dir=\"auto\">https://example.net</a>", html
+
+    html = format_value("website", "https://routing.openstreetmap.de/routed-car/route/v1/driving/-3.68,57.63;-3.68,57.61")
+    dom = Rails::Dom::Testing.html_document_fragment.parse html
+    assert_select dom, "a", 1
+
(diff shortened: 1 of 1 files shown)


1ec5 (approved):

tomhughes:
Should we still allow the `tel` links? What about the one `tg` link for telegram? Both of those likely have some utility on mobile devices in particular?

tomhughes:
Also I'm not sure what the `telephone_links` template you refer to is, as I can't see that anywhere in the current rules?

tomhughes:
Maybe add some tests for those two issues as well?

hlfan:
The tg:// link template is dropped anyway, as it's third-party, and the tel links are caught before tag2link here:

https://github.com/openstreetmap/openstreetmap-website/blob/cc6413101fc00bd5261ab5fb7ef8053aee58b1c7/app/helpers/browse_tags_helper.rb#L173-L196

I'll add tests when the rest is clear.

tomhughes:
Ah right that all makes sense then!

simonpoole:
WRT this closing https://github.com/openstreetmap/openstreetmap-website/issues/6586

In general all the tags processed by tag2link/the linkifying code are external values that we don't have any control over. Are we sure none of them besides URLs contain semi-colons as a legal (non-OSM-list) value?

tomhughes:

1ec5:
> In general all the tags processed by tag2link/the linkifying code are external values that we don't have any control over. Are we sure none of them besides URLs contain semi-colons as a legal (non-OSM-list) value?

The reason for special-casing URLs is that we know a URL can validly contain a semicolon and can parse it both ways without any external lookup. By contrast, with a key like `wikipedia=*` or `wikimedia_commons=*`, we technically wouldn’t know if the semicolon serves as a delimiter without calling out to the MediaWiki API. We could employ some heuristics like looking for `;[-a-z]{2,}:`, but this could become a rabbit hole fast. If tag2link provides more information on the heuristics to use, we could adopt a two-pass approach for more keys.

Issue #6585, opened 2025-12-03
Flakey Docker workflow


gravitystorm:
As mentioned in #6569 I'm finding the Docker workflow to be a bit flakey. I find myself re-running the workflow once or twice per week, usually either on master or on a PR where it has failed.

I haven't really kept track of what the failure causes are, since it's easiest just to re-run the workflow and concentrate on other tasks. Perhaps other people can get to the bottom of the different causes. So I'll post here from time to time when it fails.

gravitystorm:
https://github.com/openstreetmap/openstreetmap-website/actions/runs/19893059849/job/57016425381

```Error:
Api::RelationsControllerTest#test_create_race_condition:
Rack::Test::Error: No response yet. Request a page first.
    test/controllers/api/relations_controller_test.rb:519:in `block (2 levels) in test_create_race_condition'

bin/rails test test/controllers/api/relations_controller_test.rb:502
```

pablobm:
Re: `Api::RelationsControllerTest#test_create_race_condition`, I'm not sure that's a Docker thing. It may fail a bit more often on Docker, but I've seen it in the "normal" Ruby workflows too. Eg: https://github.com/openstreetmap/openstreetmap-website/actions/runs/18480327748/job/52653654041
```
Api::WaysControllerTest#test_create_race_condition:
Rack::Test::Error: No response yet. Request a page first.
    test/controllers/api/ways_controller_test.rb:457:in `block (2 levels) in test_create_race_condition'

bin/rails test test/controllers/api/ways_controller_test.rb:440
```

pablobm:
Are these failures normally triggered by the Selenium-driven tests? If so, #6497 (Reduce use of Selenium in favour of rack-test) could help alleviate the issue.

gravitystorm:
https://github.com/openstreetmap/openstreetmap-website/actions/runs/20306972690/job/58326991924

```
SiteTest#test_tooltip_shows_for_Legend_button_on_a_layer_without_a_legend_provided:
Net::ReadTimeout: Net::ReadTimeout with "Net::ReadTimeout with #<TCPSocket:(closed)>"
    test/system/site_test.rb:33:in 'block in <class:SiteTest>'

bin/rails test test/system/site_test.rb:32
```

This was on a ruby 3.4 job, so not a Docker workflow. Let's see which ones come up most often.

gravitystorm:
Again, not docker this time:

```
Error:
[...]

PR #6588, opened 2025-12-04
fixes: #6585 add Postgres healthcheck and make Docker workflow wait + retries

manik3160:
fixes:#6585
This PR stabilizes the Docker-based CI workflow by ensuring that Postgres is fully ready before Rails tests begin, and by adding lightweight retry logic around the test command. The root cause of the flakiness was that tests could start before Postgres had finished initializing, causing intermittent failures.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -28,16 +28,6 @@ jobs:
         set -euo pipefail
         retry() { n=0; until [ $n -ge 3 ]; do "$@" && break || n=$((n+1)); sleep 5; done; if [ $n -ge 3 ]; then echo "Command failed: $*"; return 1; fi }
         retry docker compose up -d
-        # Wait for Postgres healthcheck to report ready (uses the service `db` healthcheck)
-        # Fallback to polling pg_isready inside the container if healthcheck isn't exposed yet
-        for i in {1..60}; do
-          if docker compose exec db pg_isready -U openstreetmap -d openstreetmap >/dev/null 2>&1; then
-            echo "Postgres is ready"
-            break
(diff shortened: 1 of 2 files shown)


Firefishy:
Wouldn't it be better to change the `docker-compose.yml` to use a `condition: service_healthy` for the app rather than the custom logic in `.github/workflows/docker.yml`?:

```
    depends_on:
      db:
        condition: service_healthy
```

manik3160:
You're absolutely right! Using condition: service_healthy  would be much cleaner. Plz let me know more as it is my first Pr .

tomhughes:
So the database healthcheck thing doesn't look unreasonable but I can't actually find any failing cases that look like they are caused by the database being ready so I'd be interested to know what made you think it was an issue?

You've also extended the (recently introduced) timeout for the job from 20 minutes to 40 minutes - do you have some cases in mind where the current timeout was too short?

Finally you've added a couple of retry loops with no real explanation of why you think they're a good idea - if `docker build` fails once why would think you think running it a second time would help? That suggests that transient failures of some sort are common - do you have an example of that? Ideally the solution to such things is to try and fix the transient failures not to work around them with a retry loop.

gravitystorm:
I've extracted the db healthcheck into a separate PR, and tweaked some aspects of it - see #6626.

I don't believe the other retry loops are required, and I think the custom bash retry logic is far too complex for both the situation and the (un)likely benefits. It definitely won't fix #6585 since the workflows are failing at unrelated locations.

@manik3160 thanks for making this PR! We've used some of it, but we aren't going to use the rest, so I'm going to close this now.

PR #6590, opened 2025-12-04
Change Valhalla credit link to working URL

simonpoole:
The current URL points to a non-existing site/url, at least as a temporary solution this should be better.


tomhughes:
Looks good to me, thanks.

PR #6591, opened 2025-12-04
Render a not_found error for unavailable heatmaps

tomhughes:
Render a not found error if we are used for a heatmap for a user that doesn't exist or isn't visible to the requesting party instead of trying to render the view and throwing an exception.

As discussed in https://github.com/openstreetmap/openstreetmap-website/pull/6564#issuecomment-3612295106 and subsequent comment this is thought to be the root cause behind some intermittent CI failures.

### DIFF
--- a/test/controllers/users/heatmaps_controller_test.rb
+++ b/test/controllers/users/heatmaps_controller_test.rb
@@ -87,6 +87,76 @@ module Users
       assert_select ".heatmap", :count => 0
     end
 
+    def test_show_data_suspended_user
+      user = create(:user, :suspended)
+      # Create two changesets
+      create(:changeset, :user => user, :created_at => 6.months.ago, :num_changes => 10)
+      create(:changeset, :user => user, :created_at => 3.months.ago, :num_changes => 20)
+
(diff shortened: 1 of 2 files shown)


test/controllers/users/heatmaps_controller_test.rb
```diff
@@ -87,6 +87,76 @@ def test_show_data_no_changesets
...
+      heatmap_data[:data].each_value do |entry|
+        assert_equal [:date, :max_id, :total_changes], entry.keys.sort, "Heatmap data entries should have expected keys"
+      end
+      assert_equal 30, heatmap_data[:count]
+    end
+
+    def test_show_data_unknown_user
+      get user_heatmap_path(:user_display_name => "unknown_user")
+
+      # Should fail for deleted users
```
pablobm:
```suggestion
      # Should fail for unknown users
```

tomhughes: Fixed now...

hlfan (approved):

gravitystorm:
Merged, thanks

Issue #5934, opened 2025-04-15
Block anonymous notes above a given threshold


etienneJr:
### Problem

Following PR #5468, we still have a user in North of France who posts a huge amount of anonymous notes. Many of them are cryptic or seem to be copy paste from gmaps. He does not reply to our messages on these notes. And these useless notes make difficult the treatment of the other standard notes.

### Description

In these conditions, can I propose a new PR to block the possibility of posting an anonymous note after, let's say, 100 anonymous notes have already been posted? (in practice I would hide the form)

matkoniecz:
> after, let's say, 100 anonymous notes have already been posted?

globally? by given IP address?

> He does not reply to our messages on these notes

only users can comment on notes

> Following PR https://github.com/openstreetmap/openstreetmap-website/pull/5468, we still have a user in North of France who posts a huge amount of anonymous notes

another option is bot closing all anonymous notes in a given area (or reviving idea of dropping anonymous notes altogether)

etienneJr:
> globally? by given IP address?

I propose to use the cookie `_osm_anonymous_notes_count` introduced [here](https://github.com/etienneJr/openstreetmap-website/blob/0f2df0b9efcfff1d6bb0f7b2d16fa23f85d8708d/app/assets/javascripts/index/new_note.js#L167) in #5468. It's not perfect because deleting the cookie will release the blockage, but it's better than nothing I think, and it's simpler than a system based on IP address, right ?

> only users can comment on notes

Of course, the message was `please create a account to discuss with us`

> another option is bot closing all anonymous notes in a given area

not satisfactory since other anonymous notes from other people are interesting in this dense area

> or reviving idea of dropping anonymous notes altogether

it seems difficult to convince everyone. See [this discussion](https://community.openstreetmap.org/t/we-dont-need-anonymous-notes/105335). For many users, posting an anonymous note was the first contribution ever. Personally, I like the idea of keeping this feature but only allowing a limited number of anonymous notes.

pablobm:
I think this can be closed after the merge of https://github.com/openstreetmap/openstreetmap-website/pull/6593

etienneJr:
Yes! Thanks!

PR #6593, opened 2025-12-06
Hide the anonymous note creation form when the limit is exceeded

deevroman:
### Description

I am justifying this PR based on renewed interest in forum posts on this topic starting in the fall of this year https://community.openstreetmap.org/t/we-dont-need-anonymous-notes/105335/128 as well as from personal experience of resolving notes

1. After 10 notes, the create notes form hides and the warning is repainted red

2. The previous warning is displayed after 5 anonymous notes

3. I reduced the lifetime of cookies that count anonymous notes to two weeks to compensate for the limit a little

### How has this been tested?

Manual testing, as well as adding a new autotest 

### Explanation

Yes, the API is still not protected. Yes, you can open a private window. 

But at least it will stop stupid robots, and it will also stop some users who use notes for other purposes faster. It is also a small stick in the wheels of lazy vandals.

### DIFF
--- a/app/helpers/note_helper.rb
+++ b/app/helpers/note_helper.rb
@@ -34,4 +34,12 @@ module NoteHelper
               :class => "mw-100 d-inline-block align-bottom text-truncate text-wrap", :dir => "auto"
     end
   end
+
+  def soft_anonymous_notes_limit_reached?(anonymous_notes_count)
+    !current_user && anonymous_notes_count >= 5
+  end
+
+  def hard_anonymous_notes_limit_reached?(anonymous_notes_count)
+    !current_user && anonymous_notes_count >= 10
+  end
 end
--- a/app/views/notes/new.html.erb
+++ b/app/views/notes/new.html.erb
@@ -1,15 +1,14 @@
 <% set_title(t(".title")) %>
 
 <%= render "sidebar_header", :title => t(".title") %>
-
 <div class="note">
   <p class="alert alert-info"><%= t(".intro") %></p>
   <% if !current_user %>
-    <div class="alert alert-warning pb-0">
+    <div class="alert <%= hard_anonymous_notes_limit_reached?(@anonymous_notes_count) ? "alert-danger" : "alert-warning" %> pb-0">
       <p><%= t ".anonymous_warning_html",
                :log_in => link_to(t(".anonymous_warning_log_in"), login_path(:referer => new_note_path)),
                :sign_up => link_to(t(".anonymous_warning_sign_up"), new_user_path) %></p>
-    <% if @anonymous_notes_count >= 10 %>
+    <% if soft_anonymous_notes_limit_reached?(@anonymous_notes_count) %>
       <p><%= t ".counter_warning_html",
                :x_anonymous_notes => t(".x_anonymous_notes", :count => @anonymous_notes_count),
                :contribute_by_yourself => link_to(t(".counter_warning_guide_link.text"), t(".counter_warning_guide_link.url")),
@@ -17,6 +16,7 @@
     <% end %>
     </div>
   <% end %>
+  <% if !hard_anonymous_notes_limit_reached?(@anonymous_notes_count) %>
   <p class="alert alert-warning" id="new-note-zoom-warning" hidden><%= t "javascripts.site.createnote_disabled_tooltip" %></p>
   <form action="#">
     <input type="hidden" name="lon" autocomplete="off">
@@ -28,4 +28,5 @@
       <%= submit_tag t(".add"), :name => "add", :disabled => 1, :class => "btn btn-primary" %>
     </div>
   </form>
+  <% end %>
 </div>
--- a/app/assets/javascripts/index/new_note.js
+++ b/app/assets/javascripts/index/new_note.js
@@ -149,7 +149,7 @@ OSM.NewNote = function (map) {
       createNote(location, text, (feature) => {
         if (typeof OSM.user === "undefined") {
           const anonymousNotesCount = Number(OSM.cookies.get("_osm_anonymous_notes_count")) || 0;
-          OSM.cookies.set("_osm_anonymous_notes_count", anonymousNotesCount + 1, { expires: 30 });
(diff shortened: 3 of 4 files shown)


tomhughes:
I thought we had already done this? How does it differ to what we have now?

deevroman:
Currently, a suggestion to log in is shown after 10 notes:



You can still ignore it and continue creating notes

I propose to show this warning after 5 notes.

qugebert:
I doubt that this is the right approach. Anonymous notes can also be useful to us, they just take more time to process.
I admit that I myself am sometimes annoyed by too many notes where it is not possible to ask the person who submitted them for more information, but I still think that we should keep the barriers to reporting map errors as low as possible.

deevroman:
It's not (yet) about complete banning anonymous notes. If someone has created more than 10 anonymous notes through a website, then in practice this person is doing something wrong. Either these notes will contain little useful information, or it is from a dubious source, or he adds personal notes to his journey, or it will be easier and faster for the user to register and add objects own.

The reverse side of this restriction is the registration of a new mapper. This is potentially more useful.

pablobm:
For completeness, this PR would close https://github.com/openstreetmap/openstreetmap-website/issues/5934 (if I'm reading correctly).

deevroman:
> For completeness, this PR would close https://github.com/openstreetmap/openstreetmap-website/issues/5934 (if I'm reading correctly).

No. This is client-side protection. It can be bypassed, for example, using third-party apps. However, third-party app users are far fewer, and most apps add the "user agent" to the note text. Thanks to the user agent, at least some grouping of notes will appear, which will simplify the elimination of vandalism.

---

However, perhaps this will help for most cases of vandalism and problem #5934 will be solved. After all, the restrictions on the size of changesets are also imperfect and easy to bypass, but they helped.

app/views/notes/new.html.erb
```diff
@@ -1,22 +1,25 @@
 <% set_title(t(".title")) %>
 
 <%= render "sidebar_header", :title => t(".title") %>
-
+<% soft_anonymous_notes_limit_reached = !current_user && @anonymous_notes_count >= 5
+   anonymous_notes_limit_reached = !current_user && @anonymous_notes_count >= 10
+   warn_class = anonymous_notes_limit_reached ? "alert-danger" : "alert-warning" %>
```
pablobm: I think this section is a bit too much code for a template. How about moving these to helpers? Also I would rename the "normal" limit as `hard_` for a better contrast with the soft limit, and `warn_class` to something that explains what it warns about:

```
    def soft_anonymous_notes_limit_reached
      !current_user && @anonymous_notes_count >= 5
    end

    def hard_anonymous_notes_limit_reached
      !current_user && @anonymous_notes_count >= 10
    end

    def anonymous_notes_limit_warn_class
      anonymous_notes_limit_reached ? "alert-danger" : "alert-warning"
    end
```

pablobm:
> No. This is client-side protection. It can be bypassed

Right, I understand the initial description of https://github.com/openstreetmap/openstreetmap-website/issues/5934 sounds like it expects something more reliable, but ultimately a later comment (https://github.com/openstreetmap/openstreetmap-website/issues/5934#issuecomment-2830534198) refines it to pretty much what this PR offers. Which is understandable, because any other option sounds too heavy-handed.

app/views/notes/new.html.erb
deevroman: > Also I would rename the "normal" limit as hard_

Agreed. Corrected.

> I think this section is a bit too much code for a template. How about moving these to helpers?

I'm not a Ruby developer, so the code organization example was helpful. I suspect you're talking about creating an additional file, and it seems to me there's not much logic that can be moved to a separate file yet.

p.s. by the way, the template code has become more aligned. Maybe three lines in the template now not so terrible? :)

app/views/notes/new.html.erb
```diff
@@ -1,22 +1,25 @@
...
 <%= render "sidebar_header", :title => t(".title") %>
-
+<% soft_anonymous_notes_limit_reached = !current_user && @anonymous_notes_count >= 5
+   hard_anonymous_notes_limit_reached = !current_user && @anonymous_notes_count >= 10
+   warn_class = hard_anonymous_notes_limit_reached ? "alert-danger" : "alert-warning" %>
 <div class="note">
   <p class="alert alert-info"><%= t(".intro") %></p>
   <% if !current_user %>
-    <div class="alert alert-warning pb-0">
+    <div class="alert <%= warn_class %> pb-0">
```
pablobm: I think I can work with `warn_class` being inlined, and the other two being made into helpers.

```suggestion
    <div class="alert <%= hard_anonymous_notes_limit_reached ? "alert-danger" : "alert-warning" %> pb-0">
```

You don't need to create a new file for the helpers, you can put them in `app/helpers/note_helper.rb`.

app/views/notes/new.html.erb
deevroman: Fixed it. I had to change the helpers a little according to the linter's hints

app/helpers/note_helper.rb
```diff
@@ -34,4 +34,12 @@ def note_author(author, link_options = {})
...
     end
   end
+
+  def soft_anonymous_notes_limit_reached?(anonymous_notes_count)
+    !current_user && anonymous_notes_count >= 5
+  end
+
+  def hard_anonymous_notes_limit_reached?(anonymous_notes_count)
+    !current_user && anonymous_notes_count >= 10
+  end
```
pablobm, line 44: I think you are correct here, making these into pure functions instead of expecting `@anonymous_notes_count` to exist 👍 (which was what I initially suggested).

pablobm (approved):
I say let's try this 🤞 Thank you @deevroman!

pablobm:
@tomhughes - Would you be ok to merge this? I appreciate it's not a "solution", but I'm curious as to whether it makes a difference.

gravitystorm:
I'm happy to merge this. We can adjust the three config options (warning, limit, and cookie duration) if required.

gravitystorm:
Merged, thanks @deevroman , and thanks @pablobm for the reviews.

PR #6596, opened 2025-12-08
Shorten cookie expiry assignment

hlfan:
Continue the DRYing from #6430 and replace the deprecated `setYear()` method.


tomhughes:
Looks good to me, thanks.

PR #6597, opened 2025-12-08
Use RichText for changeset comments

hlfan:
So the changeset comment gets the same #6518 treatment as requested in #5780.

### DIFF
--- a/app/views/browse/_common_details.html.erb
+++ b/app/views/browse/_common_details.html.erb
@@ -8,7 +8,7 @@
 </h4>
 
 <p class="fs-6 overflow-x-auto mb-2" dir="auto">
-  <%= linkify(common_details.changeset.comment || t("browse.no_comment")) %>
+  <%= RichText.new("text", common_details.changeset.comment || t("browse.no_comment")) %>
 </p>
 
 <div class="mb-3">
--- a/app/views/changesets/show.html.erb
+++ b/app/views/changesets/show.html.erb
@@ -4,7 +4,7 @@
 
 <div class="mb-3 border-bottom border-secondary-subtle pb-3">
   <p class="fs-6 overflow-x-auto" dir="auto">
-    <%= linkify(@changeset.comment || t("browse.no_comment")) %>
+    <%= RichText.new("text", @changeset.comment || t("browse.no_comment")) %>
   </p>
   <%= tag.p :class => "details", :data => { :changeset => changeset_data(@changeset) } do %>
     <%= changeset_details(@changeset) %>


app/views/changesets/show.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <div class="mb-3 border-bottom border-secondary-subtle pb-3">
   <p class="fs-6 overflow-x-auto" dir="auto">
-    <%= linkify(@changeset.tags["comment"].to_s.presence || t("browse.no_comment")) %>
+    <%= RichText.new("text", @changeset.comment || t("browse.no_comment")) %>
```
tomhughes, line 7: The previous question about `browse.no_comment` applies here as well.

app/views/browse/_common_details.html.erb
```diff
@@ -8,11 +8,7 @@
 </h4>
 
 <p class="fs-6 overflow-x-auto mb-2" dir="auto">
-  <% if common_details.changeset.tags["comment"].present? %>
-    <%= linkify(common_details.changeset.tags["comment"]) %>
-  <% else %>
-    <%= t "browse.no_comment" %>
-  <% end %>
+  <%= RichText.new("text", common_details.changeset.comment || t("browse.no_comment")) %>
```
tomhughes, line 11: Do we need to put the `browse.no_comment` resource through the rich text processor, or through `linkify` as the initial commit does?

hlfan: I just took the structure from `changesets/show.html.erb` to not have to worry with the nullable comment and avoid code that's more difficult to read:
```suggestion
  <%= common_details.changeset.comment&.then { |t| RichText.new("text", t) } || t("browse.no_comment") %>
```

app/views/changesets/index.atom.builder
```diff
@@ -25,7 +25,7 @@ atom_feed(:language => I18n.locale, :schema_date => 2009,
                  :type => "application/osmChange+xml"
 
       if !changeset.tags.empty? && changeset.tags.key?("comment")
```
pablobm: I think an additional API change would be good here, to completely abstract the comment living in a tag:

```suggestion
      if changeset.comment?
```

What I'm not sure is why the empty check on `tags` was necessary :thinking:

app/models/changeset.rb
```diff
@@ -200,6 +200,8 @@ def save_with_tags!
     end
   end
 
+  def comment = tags["comment"].to_s.presence
```
pablobm: Is the `.to_s` required here? Just wondering, I'm not sure. If the value is nil, `#presence` will have the same behaviour, unless I'm much mistaken.

I appreciate this is just copied from the existing code, but since we are improving this, we might as well consider this question.

app/views/changesets/index.atom.builder
hlfan: How about no additional API change?
```suggestion
      if changeset.comment.present?
```

pablobm: Works for me! :+1:

tomhughes (approved):
Looks good to me now, thanks.

Issue #6517, opened 2025-11-11
Donate button is hard to find


1ec5:
There’s a “Make a Donation” link in fine print in the lower-right corner of the slippy map on the homepage. [A recent forum post](https://community.openstreetmap.org/t/followup-questions-to-2025-gm/135628/41) points out that it’s difficult to find and even difficult for some to see. I don’t have specific traffic statistics, but I guess I’d agree with that anecdotally. Can we find room for a Donate link in the navigation bar at the top, maybe between Help and About?

In the past month, as I’ve introduced OSM mapping to laypeople and GIS people, several times someone has asked something along the lines of: Are you a nonprofit? If so, how come you don’t have a Donate button on the homepage or any mention of it on the About page? They seem to expect it in the navigation bar, since that’s where most nonprofit sites have it. One of the pages in the navigation bar, Copyright, does have a donation link as of #5132, but it only helps people who land on the copyright page from map attribution. No one else would expect it to be on that page.

A Donate link wouldn’t need to be any flashier than the other links in the navigation bar. It wouldn’t be a donation drive, just something to get users where they want to go. We can probably remove the existing “Make a Donation” link to avoid redundancy. That would also avoid suggesting that the donations go directly to “OpenStreetMap contributors”, which appears directly to the left.

CommanderStorm:
<img width="1909" height="459" alt="Image" src="https://github.com/user-attachments/assets/4a17ee16-3d88-472c-9b5a-8dffa6db185e" />

Thanks for bringing this up here as well - I hadn’t thought to post it in this repository too.

To make space for adding a new option in the top bar, some existing items could be consolidated:

1. The Export feature could be integrated into the Share menu, as that’s where most users would likely expect it
1. GPS Traces could be hidden, since they’re not very useful for most regular users in the current state (or is it?)
1. The number of “About Us”–type pages could perhaps be reduced from four to three.

1ec5:
For what it’s worth, the site can technically accommodate more items in the navigation bar. If the window is too narrow to fit all of the items, the rest go in an overflow menu, starting from the right:

<img width="536" height="207" alt="Communities, Copyright, Help, and About relegated to a More menu." src="https://github.com/user-attachments/assets/0f452c6e-d939-4365-9644-d4616e346afd" />

On a mobile screen, the panel that contains the navigation bar can wrap onto another line as needed without interfering with normal site usage:

[...]

PR #6601, opened 2025-12-08
Use red heart in donate attribution

deevroman:
### Description

In most fonts, the current heart is monochrome. As noted in #6517, the Make a Donation button is hard to find. This is one way to make it more visible.

### How has this been tested?

Снимок экрана 2025-12-09 в 00 23 27
Снимок экрана 2025-12-09 в 00 23 05

### DIFF
--- a/app/assets/javascripts/leaflet.map.js
+++ b/app/assets/javascripts/leaflet.map.js
@@ -59,7 +59,7 @@ L.OSM.Map = L.Map.extend({
         }).prop("outerHTML")
       });
 
-      attribution += credit.donate ? " &hearts; " : ". ";
+      attribution += credit.donate ? " ♥️ " : ". ";
       attribution += makeCredit(credit);
       attribution += ". ";
 


app/assets/javascripts/leaflet.map.js
```diff
@@ -59,7 +59,7 @@ L.OSM.Map = L.Map.extend({
         }).prop("outerHTML")
       });
 
-      attribution += credit.donate ? " &hearts; " : ". ";
+      attribution += credit.donate ? " ♥️ " : ". ";
```
CommanderStorm, line 62: Personal Preference: this punctuation looks a bit out of place.

Unrelated to your PR though, so lets not change it ^^

```suggestion
      attribution += credit.donate ? " ♥️ " : " ";
```

1ec5: The period is there because the attribution lines for each of the third-party layers’ attribution lines lack a period at the end. Without it, the attribution would run right into “Website and API terms”. However, this is “Lego localization” that doesn’t work as well in some languages that use other punctuation. The period should be moved to each of the attribution strings, such as:

https://github.com/openstreetmap/openstreetmap-website/blob/7fdee463c76c1cbb07f9392986950674bc05edb8/config/locales/en.yml#L3595

deevroman: I suggest we think about it in a separate PR:)

CommanderStorm, line 62: Sorry to bikeshed on this one.

🤔 actually, would green not be the color that this should be? This would match the rest of the donation text

```suggestion
      attribution += credit.donate ? " 💚 " : ". ";
```

deevroman: I specifically chose the red color so that it would be visible, and not disguised as a green map and inscription

CommanderStorm: Fair

gravitystorm:
Merged, thanks!

PR #6606, opened 2025-12-10
Allow users to hide the heatmap

pablobm:
Closes https://github.com/openstreetmap/openstreetmap-website/issues/5804, https://github.com/openstreetmap/openstreetmap-website/pull/5853.

This is a replacement for https://github.com/openstreetmap/openstreetmap-website/pull/5853. That area of the code has changed enough that I think the original PR cannot be salvaged.

This PR adds a preference that allows users to hide the heatmap from their profiles, if they so choose. There was a question of whether this was necessary, privacy implications, LWG discussion, etc. As communicated to me, this is the decision:
- OSM have a legitimate interest to show the heatmap, so it's ok that users get it by default.
- The heatmap is not strictly necessary in order to use OSM, so users should have the ability to hide it.

Notes:
- The setting can be accessed via a big button under the heatmap:

The heatmap on a user's profile page. A long, horizontal button underneath, spanning the whole width of the heatmap above, reads 'Edit Heatmap'

- When a user chooses to hide their heatmap, absolutely nobody can see it. No moderators, no admins, not even the user themselves. This is to align with the goal of making it obvious to users what is showing and what is not on their profiles.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -300,7 +300,7 @@
 <% end %>
 <% if owned %>
   <div class="mt-2">
-    <%= link_to t(".edit_heatmap"), profile_heatmap_path, :class => "btn btn-sm w-100 btn-outline-primary" %>
+    <%= link_to t(".edit_heatmap"), profile_heatmap_path, :class => "btn btn-sm w-100 btn-outline-primary edit_heatmap_button" %>
   </div>
 <% end %>
 
--- a/test/controllers/users/heatmaps_controller_test.rb
+++ b/test/controllers/users/heatmaps_controller_test.rb
@@ -157,6 +157,26 @@ module Users
       assert_response :not_found
     end
 
+    def test_show_not_public
+      user = create(:user)
+      user.update(:public_heatmap => false)
+
+      get user_heatmap_path(user)
+      assert_response :not_found
+
+      session_for(create(:user))
+      get user_heatmap_path(user)
+      assert_response :not_found
+
+      session_for(create(:moderator_user))
+      get user_heatmap_path(user)
+      assert_response :not_found
+
+      session_for(create(:administrator_user))
+      get user_heatmap_path(user)
+      assert_response :not_found
+    end
+
     def test_show_rendering_of_user_with_no_changesets
       user_without_changesets = create(:user)
 
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -375,4 +375,73 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
     assert_select ".profile-diary-card", 0
   end
+
+  def test_heatmap_visibility
+    edit_heatmap_button_selector = ".edit_heatmap_button"
+    heatmap_selector = ".heatmap_frame"
+    protagonist = create(:user)
+
+    # Testing unauthenticated user first
+
+    protagonist.update(:public_heatmap => true)
+    get user_path(protagonist)
+    assert_select heatmap_selector
+    refute_select edit_heatmap_button_selector
+
+    protagonist.update(:public_heatmap => false)
+    get user_path(protagonist)
+    refute_select heatmap_selector
+    refute_select edit_heatmap_button_selector
+
+    session_for(protagonist)
+
+    protagonist.update(:public_heatmap => true)
+    get user_path(protagonist)
+    assert_select heatmap_selector
+    assert_select edit_heatmap_button_selector
+
+    protagonist.update(:public_heatmap => false)
+    get user_path(protagonist)
+    refute_select heatmap_selector
+    assert_select edit_heatmap_button_selector
+
+    antagonist = create(:user)
+    session_for(antagonist)
+
+    protagonist.update(:public_heatmap => true)
+    get user_path(protagonist)
+    assert_select heatmap_selector
+    refute_select edit_heatmap_button_selector
+
+    protagonist.update(:public_heatmap => false)
+    get user_path(protagonist)
+    refute_select heatmap_selector
+    refute_select edit_heatmap_button_selector
+
+    moderator = create(:moderator_user)
+    session_for(moderator)
+
+    protagonist.update(:public_heatmap => true)
+    get user_path(protagonist)
+    assert_select heatmap_selector
+    refute_select edit_heatmap_button_selector
+
+    protagonist.update(:public_heatmap => false)
+    get user_path(protagonist)
+    refute_select heatmap_selector
+    refute_select edit_heatmap_button_selector
+
+    admin = create(:administrator_user)
+    session_for(admin)
+
+    protagonist.update(:public_heatmap => true)
+    get user_path(protagonist)
+    assert_select heatmap_selector
+    refute_select edit_heatmap_button_selector
+
+    protagonist.update(:public_heatmap => false)
+    get user_path(protagonist)
+    refute_select heatmap_selector
+    refute_select edit_heatmap_button_selector
+  end
 end


gravitystorm:
I'm happy to see this as an option.

The control should be on the "edit profile" page, not on the profile itself. I want to minimise the differences between what you see when you view your own profile, compared to what other people see. We have a page to control what is or is not shown on your own profile, so that's where the control should be.

I don't think it's worth storing a UserPreference. The UserPreference system is really just there as an API, for editors etc to store user-specific state between instances, or perhaps to read website preferences that might be useful for other apps to know via the API. But I admit that's not really borne out in practice, since we've been putting random website preferences in their regardless of whether they need to be exposed via the API.

I think there's no need for anything other than the website to read/write this particular option, so I don't think it needs to go into the UserPreference system, but I'd ask for feedback from other people on this point.

For admins - why do they need to see it? Should this be visible to moderators too? I would think moderators are more focussed on mapping behaviour than admins, but I don't know if either group use the heatmap as part of their tasks.

1ec5:
> There was a question of whether this was necessary, privacy implications, LWG discussion, etc. As communicated to me, this is the decision:
> 
> * OSM have a legitimate interest to show the heatmap, so it's ok that users get it by default.
> * The heatmap is not strictly necessary in order to use OSM, so users should have the ability to hide it.

Correct, this is an accurate description of what the LWG decided. I’ve updated https://github.com/openstreetmap/openstreetmap-website/issues/5804#issuecomment-3638311030 with more details.

> The option appears as a checkbox under the heatmap. At the moment, any preferences that affect the profile can be reached from the profile page, so filing this under Preferences didn't feel right.

I’d agree with @gravitystorm that the actual preference should be under /profile somewhere. We don’t have a separate page for previewing your own profile as it appears to other users, so additional controls make the page pretty unwieldy. For consistency, I think the user would expect to see a button that takes them to the specific tab of the Edit Profile page. Unfortunately, this means yet another tab so we have something to link to.

(I wonder if we need to revisit the decision in #6084 to tie each of these tabs to a path rather than a named anchor and URL hash. Eventually we’ll have enough preferences that the tab bar will get pretty unwieldy. Or maybe we need to have a dedicated page for previewing the profile so we can make the default view more interactive…)

> For admins - why do they need to see it? Should this be visible to moderators too? I would think moderators are more focussed on mapping behaviour than admins, but I don't know if either group use the heatmap as part of their tasks.

I vaguely recall comments that the heatmap would be useful to DWG members, but I can’t find them anymore, or maybe it was just speculation. In any case, if the administrators don’t see a need to always see the heatmap, then let’s hide it for them too. We can always revisit if the need arises.

tomhughes:
> > For admins - why do they need to see it? Should this be visible to moderators too? I would think moderators are more focussed on mapping behaviour than admins, but I don't know if either group use the heatmap as part of their tasks.
> 
> I vaguely recall comments that the heatmap would be useful to DWG members, but I can’t find them anymore, or maybe it was just speculation. In any case, if the administrators don’t see a need to always see the heatmap, then let’s hide it for them too. We can always revisit if the need arises.

It's probably more useful for moderators that administrators but if we do enable it for moderators then we might as well include administrators as well.

pablobm:
> The control should be on the "edit profile" page, not on the profile itself. 

That makes sense, but I'm not sure what this would look like. Currently users arrive at the edit section via:
- "Edit Profile Details" which expands buttons that link to the respective edit interfaces.
- "Edit Description" which is in place of the description and also leads to the corresponding edit interface.

What would be a good way to adapt this convention to the heatmap? @1ec5 suggests a button. Perhaps a new "Edit heatmap" (?) button under the heatmap, linking to the a page showing this checkbox+submit?

> The UserPreference system is really just there as an API, (...) (b)ut I admit that's not really borne out in practice

What I had in mind was the "dark mode" preferences as introduced at https://github.com/openstreetmap/openstreetmap-website/pull/6492. Do you think we should be introducing a new column in the `users` table? I guess columns are cheap but it feels strange to me, particularly if over time we continue adding/removing preferences.

> I think there's no need for anything other than the website to read/write this particular option

There has been discussion on GDPR-adjacent topics about how the API is not able to provide access to certain things like confirmation of Terms of Use. I wonder if we should start designing things like this in a way that allows for API clients to access these in the future?

> Should this be visible to moderators too?

~That makes sense. I'm going to change this to admins+moderators as per @tomhughes comment below.~ (Changed my mind, see next comment).

pablobm:
Correction. I'm going to simplify and make it so that admins/mods can't see a hidden heatmap either, as per @1ec5's comment:

> let’s hide it for them too. We can always revisit if the need arises.

pablobm:
Currently I have it as a big button under the heatmap:

The heatmap on a user's profile page. A long, horizontal button underneath, spanning the whole width of the heatmap above, reads 'Edit Heatmap'

pablobm:
I think this is now in a presentable state, even if objections could still appear. Notes:
- The setting is still a `UserPreference`. Again, happy to move this to a new column `users.public_heatmap`, but wanted to make sure first.
	- In particular, I seem to recall there were some concerns about things like accepting ToUs that can be only done on the website, so perhaps things like that should progressively be moved to locations accessible by the API? Just spitballing here. (/cc @1ec5)
- It occurred to me that the routes to edit the profile are under `show` actions, when the convention would be to have them under `edit` actions. I've stuck to `show`, but perhaps this should be changed in a different PR?
- I have not added system tests. I noticed these exist for the other profile settings (eg: [`test/system/profile_company_change_test.rb`](https://github.com/openstreetmap/openstreetmap-website/blob/master/test/system/profile_company_change_test.rb), but I think they are redundant (we already cover them with controller tests) and slow down the test suite. I would remove all of them (in a separate PR). Thoughts?

1ec5:
> There has been discussion on GDPR-adjacent topics about how the API is not able to provide access to certain things like confirmation of Terms of Use. I wonder if we should start designing things like this in a way that allows for API clients to access these in the future?

We’re tracking this confirmation in #6610. It’s premature to assume that everyone has confirmed the ToU, but I haven’t seen any indication that we need to expose this information via the API. If so, it would probably be similar to the “consider public domain” flag, which a client can [discover about the currently authenticated user](https://wiki.openstreetmap.org/wiki/API_v0.6#Details_of_the_logged-in_user:_GET_/api/0.6/user/details) but not about other users.

> Currently I have it as a big button under the heatmap:

Sensible, though kind of amusing. Maybe you could stick it next to the heading, so it’s less obtrusive. It feels like we’re eventually going to have an overarching discussion about restyling the editing options anyways.

pablobm:
I was just playing with the location of the "Edit Heatmap" button, to put it next to the heading as suggested by @1ec5. I found it to be trickier than it seems, let me explain:
- If  the heatmap is enabled, this is retrieved via an Ajax request (with a Turbo Frame). This will include the heading.
- There are three use cases:
	- Disabled: no Ajax request.
	- Enabled, but no contributions: the Ajax request takes place, renders a mostly empty result, with no heading.
	- Enabled, there are contributions: the Ajax request takes place, renders the heatmap with the heading.
- The check of whether there are contributions takes place when rendering the HTML retrieved via Ajax.
- To place the button, visually:
	- When the heatmap is rendered, we can put the button next to the heading, a-ok.
	- If not rendered (whatever the reason), we'll have to put a button roughly like the one I showed above. Not necessarily as wide, but a button in that general area nonetheless.
- To place the button, in terms of code:
	- No issue if the heatmap is rendering.
	- If enabled but no contributions available, we'll only know when responding to the Ajax request. The standalone button will have to be rendered as part of the "heatmap" template.
	- If disabled, we know at users/show time and the Ajax request doesn't take place. The button will have to be rendered as part of users/show.

Technically all this could be worked around, but I don't think it's worth the effort.

config/locales/en.yml
```diff
@@ -2214,6 +2215,15 @@ en:
       update:
         success: Profile location updated.
         failure: Couldn't update profile location.
+    heatmaps:
+      show:
+        title: Edit Heatmap
```
hlfan: This doesn't really "edit" the heatmap, it only toggles its visibility.

pablobm: It's complicated. It leads to a whole "Heatmap settings" page where yes, the only option is to toggle it. However if it said "Toggle Heatmap Visibility" I think it would be misleading, giving the impression that a click of the button is all you need.

hlfan: Maybe _configure heatmap_? I don't know which other options to add tho.

1ec5: We do have enough room to say “Edit Heatmap Settings”. Anyhow, I’m of half a mind to suggest replacing all these edit buttons with more discreet ✏️ buttons, like you often see on templates all over the wiki. But that should be a separate discussion. Don’t want to distract from this more important feature.

pablobm:
Moved the setting to its own DB column `users.public_heatmap` and renamed the button as "Configure Heatmap" (plus renames to section heading and update button).

app/controllers/users/heatmaps_controller.rb
```diff
@@ -13,7 +13,7 @@ class HeatmapsController < ApplicationController
     def show
       @user = User.find_by(:display_name => params[:user_display_name])
 
-      if @user && (@user.visible? || current_user&.administrator?)
+      if @user&.public_heatmap? && (@user.visible? || current_user&.administrator?)
```
tomhughes, line 16: So we decided against always showing it to moderators and/or administrators?

In fact as it stands it isn't even shown to the user themselves?

1ec5: I mentioned to @pablobm early on that we could take a conservative approach and hide it from both groups unless someone from either group said they needed it for something. (As long as it’s shown to moderators, it would only make sense for it to be shown to administrators too.)

We could reasonably show the user’s own heatmap even when they opt out, but it really works against the goal of making the profile look similar to what the public sees. Unless we hide it in response to the setting, the setting will appear to have no effect unless the user opens up their profile in private browsing or something. Maybe as tail work, we could consider moving it to the dashboard in that case, so the user can at least still navigate to their own contributions by date?

pablobm: What @1ec5 said. I have updated the PR description to make this explicit.

tomhughes: In that case I think this is good to merge, thanks.

PR #6607, opened 2025-12-10
Ensure consistent naming for all factory files

pablobm:
I noticed that some factory files were named in the singular and some in the plural. This PR makes it all consistent by using the singular for all.

I don't think there's a clear convention one whether to use either. To me, it makes sense to use the singular as the factories themselves are in the singular: `create(:user)`.


gravitystorm:
I agree this should be consistent.

It's exactly the kind of thing we need a linter for, otherwise it'll just drift again in the future - it's clearly not an easy thing to spot during reviews! I've checked the `rubocop-factory_bot` repo, and found https://github.com/rubocop/rubocop-factory_bot/issues/57 . Ideally, that would be implemented upstream, or perhaps we would implement it ourselves as a [custom cop](https://github.com/openstreetmap/openstreetmap-website/tree/4c97ee96df281efd0e4129e3a841d12fbdb898f7/.rubocop).

As for which name - it seems like there's three conventions:
* singular model name
* plural model name
* singular model name plus "_factory" suffix

I've check the factory_bot docs, and they don't mention anything. [`annotate_models`](https://github.com/ctran/annotate_models/) (mentioned in the issue above) supports the three alternatives (while referring to "_factory" suffix as ["old style"](https://github.com/ctran/annotate_models/blob/develop/lib/annotate/annotate_models/file_patterns.rb#L107)). Neither Discourse nor Mastodon use factory_bot. I can't find a style guide with any suggestions.

Personally I prefer the plural option, since my IDE then shows "user.rb" and "users.rb" and I know which is which :smile: But I'm eager to find some better reasoning than that.

pablobm:
From those options, I prefer `<singular>_factory.rb`, because:

- The factory within is defined in the singular. It feels weird to me that `factories/users.rb` defines internally a `factory :user`.
- The suffix helps telling the file from others in an IDE (in a list of tabs or similar).

But I have worked in projects with the plural before and I can live with it. In fact I think plural might be more widespread...

gravitystorm:
I've just been digging further, and I found that factory_bot_rails [generates them as plural model names](https://github.com/thoughtbot/factory_bot_rails/blob/5ef69711d72dc269d363505744e72c3b1f213187/lib/generators/factory_bot/model/model_generator.rb#L74) by default. (See also #6625)

So let's standardise on the plural option, since we tend to stick to defaults where possible.

pablobm:
All plural now.

pluralize_file_names.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
```
gravitystorm: I don't think we need this file in the repo, since if we miss anything during future PRs we're likely to only need to rename files one at a time. Also, it's probably not intentional to include this at the top level of the repo.

pablobm: :flushed: An accident indeed. Removed now

gravitystorm:
Merged, thanks!

PR #6608, opened 2025-12-10
Put `require` calls outside modules unless necessary

pablobm:
Very minor thing, but I'm not sure about `require` calls within modules, unless there's a specific reason for them which I don't think is the case here.

These could probably be removed entirely, similar to https://github.com/openstreetmap/openstreetmap-website/pull/6461, but since it's `lib/` I'm less bothered about that.

Another discussion could be had about moving everything in `lib/` to `app/lib/` as I believe that's the recommendation these days (see https://github.com/rails/rails/issues/37835#issuecomment-782843222). However perhaps some items do belong under `lib/`, if they are considered to be something that can be extracted to a gem and used by other projects. Let me know if you have thoughts.

### DIFF
--- a/lib/osm.rb
+++ b/lib/osm.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
+require "time"
+require "rexml/parsers/sax2parser"
+require "rexml/text"
+
 # The OSM module provides support functions for OSM.
 module OSM
-  require "time"
-  require "rexml/parsers/sax2parser"
-  require "rexml/text"
-
   # The base class for API Errors.
   class APIError < RuntimeError
     def initialize(message = "Generic API Error")


gravitystorm:
Merged, thanks.

I think that everything (except tasks) could be moved to app/lib, and that's what I've done in other projects too. 

I thought we had to specifically ask for `lib/` to be autoloaded, but I'm not sure if that's actually the case at the moment - I can't find that with a cursory search.

pablobm:
Last time I checked `lib/` wasn't autoloaded by default, which I think is the main reason to move things to `app/lib/` instead.

tomhughes:
https://github.com/openstreetmap/openstreetmap-website/blob/4c97ee96df281efd0e4129e3a841d12fbdb898f7/config/application.rb#L19 is what enables it for us.

pablobm:
> is what enables it for us.

Right. So I think that's another reason to move to `app/lib/`: if we are doing something similar already, better if we follow the convention and drop the configuration :+1:

Issue #6611, opened 2025-12-11
Backquotes around tags in CS comments are included when auto-linking


Ltrlg:
### URL

https://www.openstreetmap.org/changeset/175766995

### How to reproduce the issue?

I quite like the convention to surround tags in backquotes in comments, as a quite standard technical string indicator. However recently introduced tags auto-linking in comments includes the backquotes in the key/value pair, which means that ``` `access=private` ``` links to [Tag:\`access=private\`](https://wiki.openstreetmap.org/wiki/Tag:%60access=private%60) instead of [Tag:access=private](https://wiki.openstreetmap.org/wiki/Tag:access=private).

tomhughes:
There is no such convention that I'm aware of - changeset comments are not, and never have been, handled as markdown so markdown syntax is not valid there.

In addition if the change you're referring to is #6597 then that isn't even live on the production site yet!

tomhughes:
I guess #6518 is actually the change you mean? That is live.

hlfan:
Yeah obviously the settings are currently set to rather over-link than to under-link, but backticks can be added to the ignore-regex to account for markdown text being processed as plain text.
And cases like this will come up, but since it's all in the rendering pipeline it will automatically be fixed retroactively.

Ltrlg:
Sorry, I meant “changeset discussion comment”, forgot that the term is used for both.

@tomhughes Convention is not syntax. Markdown is a (somewhat) formalized syntax that was first inspired by preexisting conventions for text-only writing and then influenced conventions in return. Granted, it might not be the most common convention, but I did use it in comments before and retroactively changing rendering to make it look bad seems wrong.

@hlfan Thanks! The set of characters used in common OSM keys is pretty limited¹, maybe that part should use a positive set instead of a negative one?

¹ Arguably even the values are limited, since this is linking to wiki pages and values with more unusual characters are probably not much documented. The proper thing to do imho would be to use the same logic as in tags tables to choose linking to full tag / key only / not at all, but this would be more work and your quick fix is a good step forward.

pnorman:
> I did use [markdown] in comments before and retroactively changing rendering to make it look bad seems wrong

I don't believe Markdown has ever been supported in changeset discussions.

hlfan:
So turns out Rinku is the one including the backticks in the links:
<img src="https://github.com/user-attachments/assets/73904f33-2992-467a-a26e-62ca0c756827" />
😞

[...]

PR #6612, opened 2025-12-11
Scope wiki linkification characters using POSIX words

hlfan:
Fixes the Rinku-independent part of #6611

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -283,7 +283,7 @@ module RichText
           expanded_path = "#{rule.host || "#{Settings.server_protocol}://#{Settings.server_url}"}/#{rule.path_template}"
           rule.patterns
               .select { |pattern| pattern.is_a?(String) }
-              .map { |pattern| [Regexp.new("(?<=^|#{URL_UNSAFE_CHARS})#{pattern}", Regexp::IGNORECASE), expanded_path] }
+              .map { |pattern| [Regexp.new("(?<=^|#{URL_UNSAFE_CHARS})#{pattern}", Regexp::IGNORECASE, :timeout => 0.01), expanded_path] }
         end
     end
   end
--- a/test/lib/rich_text_test.rb
+++ b/test/lib/rich_text_test.rb
@@ -496,6 +496,33 @@ class RichTextTest < ActiveSupport::TestCase
     assert_equal t0.delete("`"), t1.strip
   end
 
+  def test_text_to_html_linkify_trims_punctuation
+    r = RichText.new("text", "foo `surface=metal) bar inscription=🙂 baz")
+    assert_html r do
+      assert_dom "a", :count => 1
+      assert_dom "a[href$='Tag:surface=metal']", :text => "surface=metal"
+    end
+  end
+
+  def test_text_to_html_linkify_recognizes_non_standard_wiki_pages
+    r_paren = RichText.new("text", "foo source=Isle_of_Man_Government_1:25000_map_(2007) bar")
+    assert_html r_paren do
+      assert_dom "a[href*='Tag:source']", :text => "source=Isle_of_Man_Government_1:25000_map_(2007)"
+    end
+    r_latin_ext = RichText.new("text", "foo cuisine=açaí bar")
+    assert_html r_latin_ext do
+      assert_dom "a[href*='Tag:cuisine']", :text => "cuisine=açaí"
+    end
+    r_cyrillic = RichText.new("text", "foo name=Продукты bar")
+    assert_html r_cyrillic do
+      assert_dom "a[href*='Tag:name']", :text => "name=Продукты"
+    end
+    r_cjk = RichText.new("text", "foo shop=園芸店 bar")
+    assert_html r_cjk do
+      assert_dom "a[href*='Tag:shop']", :text => "shop=園芸店"
+    end
+  end
+
   def test_text_to_html_email
     r = RichText.new("text", "foo example@example.com bar")
     assert_html r do
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -173,11 +173,11 @@ linkify:
       - "@(?<username>\\S+)"
       path_template: "user/\\k<username>"
     - patterns:
-      - "(?<key>[^\"?#<>/\\s]+)=\\*"
+      - "(?<key>[-+,.:;'%()\\p{Word}]+)=\\*"
       path_template: "wiki/Key:\\k<key>"
       host: "https://wiki.openstreetmap.org"
     - patterns:
-      - "(?<key>[^\"?#<>/\\s]+)=(?<value>[^\"?#<>\\s]+)"
+      - "(?<key>[-+,.:;'%()\\p{Word}]+)=(?<value>[-+,.:;'%()/\\p{Word}]+)"
       path_template: "wiki/Tag:\\k<key>=\\k<value>"
       host: "https://wiki.openstreetmap.org"
   display_rules:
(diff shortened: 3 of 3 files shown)


pablobm:
Hm, I wish we tested those too. I understand we present them as a configuration detail, but as they get more complex their logic is going to be forgotten.

tomhughes:
There's no reason we can't test the behaviour of the default though?

We do have a load of tests in `test/lib/rich_text.rb` though most of them override the actual linkification rules though I don't really know why as testing with the default seems at least as useful and probably more so.

pablobm:
Haven't tried, but by looking at the code I think if we don't call `with_settings` (the test helper) then the defaults will apply, making them testable :+1:

hlfan:
The default test settings are a bit different from the default settings:

https://github.com/openstreetmap/openstreetmap-website/blob/a30c8d037e23b2ee48e0aabb9a0e555f27ef8be3/test/lib/rich_text_test.rb#L398

That's what I needed to override to paste in the cases from #5780.

pablobm:
@hlfan - I think we shouldn't try to fix https://github.com/openstreetmap/openstreetmap-website/pull/6616 (mentions with spaces) with a regexp or similar. I think we should implement a selector of the sort you see in GitHub or Slack, where you get a dropdown the moment you enter an "@" character, and then this gets stored as `@[user_id:$USER_ID]` (or something to that effect) that is interpreted and replaced as appropriate when renderding. This would also fix the issue of users changing usernames and leaving old usernames behind in comments, sometimes maliciously.

test/lib/rich_text_test.rb
```diff
@@ -491,6 +518,38 @@ def test_text_to_html_linkify_no_year_misinterpretation
...
+    key_links = WIKI_PAGES.values
+                          .flat_map { |l| l["key"]&.keys }
+                          .filter { |key| key && !%r{\*:|/|\*$}.match?(key) }
+                          .map { |l| "#{l}=*" }
+                          .uniq
+                          .filter_map do |target|
+      r = RichText.new("text", "foo #{target} bar")
+      links = Nokogiri::HTML::DocumentFragment.parse(r.to_html).css("a")
+      target if links.empty?
+    end
```
pablobm: I understand you are being thorough with these tests, but I have two problems with this approach:
1. It's not readable. There's too much transformation going on to easily see exactly what we are testing.
2. It's a long test to run, clocking at over 15s in my computer and causing the fan to kick off.

Instead I would test a handful of hand-picked examples, so that it's clear what it does and doesn't spend too much effort on it.

gravitystorm:
> gets stored as `@[user_id:$USER_ID]` (or something to that effect)

We could use ActionText Attachments (see [this blog](https://theleafnode.com/embed-active-record-models-in-action-text/) and the [ActionText docs](https://edgeguides.rubyonrails.org/action_text_overview.html#signed-globalid)) for this.

hlfan:
Do you wanna make a separate issue for the user dropdown?

tomhughes:
Yes that's a technical way to store it but it really requires that you have a proper rich text editor that is able to generate attachment references using some user friendly user interface which is of course exactly what ActionText does with it's trix editor I assume.

test/lib/rich_text_test.rb
hlfan: > causing the fan to kick off

unacceptable!
Would you say one assertion per selectively allowed Unicode code block is enough?

pablobm:
> Do you wanna make a separate issue for the user dropdown?

Probably best, yes. Thank you.

test/lib/rich_text_test.rb
pablobm: Sorry, what do you mean by Unicode code block in this context? I'm looking at `config/wiki_pages.yml` and I'm confused.

hlfan: https://github.com/openstreetmap/openstreetmap-website/blob/07c068dc5ff4ed1e53ee6f1bb0367b643907ad4d/config/settings.yml#L173

The three `\u%%%%-\u%%%%` parts are roughly based on these blocks:
- Latin extensions
- Cyrillic
- Unified CJK

hlfan: ...actually, the one CJK tag `shop=園芸店` is already redirected to https://osm.wiki/wiki/JA:Tag:shop=agrarian, might as well drop it here too.

tomhughes: Would it be better to use `\p{Word}` to match unicode word characters rather than using `\w` to match ascii word characters and then trying to tag on chunks of the unicode code space? See https://docs.ruby-lang.org/en/3.4/Regexp.html#class-Regexp-label-Unicode+Properties for details.

hlfan:
@tomhughes `\p{Word}` is causing timeouts... is there a more performant option?

pablobm:
Not sure it's enough for a timeout, but I can confirm that in the `key=value` patterns, the `\p{Word}` ones are running about 10x slower than similar `\w` ones (just as a comparison, I know they are not the same). These are some numbers, taken from running `test/models/diary_entry_test.rb` in my computer:

```
 11us /(?<=^|[^\w!#$%&'*+,.\/:;=?@_~^\-])(?<key>[-+,.:;'%()\w]+)=(?<value>[-+,.:;'%()\/\w]+)/i
116us /(?<=^|[^\w!#$%&'*+,.\/:;=?@_~^\-])(?<key>[-+,.:;'%()\p{Word}]+)=(?<value>[-+,.:;'%()\/\p{Word}]+)/i

 11us /(?<=^|[^\w!#$%&'*+,.\/:;=?@_~^\-])(?<key>[-+,.:;'%()\w]+)=(?<value>[-+,.:;'%()\/\w]+)/i
116us /(?<=^|[^\w!#$%&'*+,.\/:;=?@_~^\-])(?<key>[-+,.:;'%()\p{Word}]+)=(?<value>[-+,.:;'%()\/\p{Word}]+)/i

 24us /(?<=^|[^\w!#$%&'*+,.\/:;=?@_~^\-])(?<key>[-+,.:;'%()\w]+)=(?<value>[-+,.:;'%()\/\w]+)/i
236us /(?<=^|[^\w!#$%&'*+,.\/:;=?@_~^\-])(?<key>[-+,.:;'%()\p{Word}]+)=(?<value>[-+,.:;'%()\/\p{Word}]+)/i

 32us /(?<=^|[^\w!#$%&'*+,.\/:;=?@_~^\-])(?<key>[-+,.:;'%()\w]+)=(?<value>[-+,.:;'%()\/\w]+)/i
302us /(?<=^|[^\w!#$%&'*+,.\/:;=?@_~^\-])(?<key>[-+,.:;'%()\p{Word}]+)=(?<value>[-+,.:;'%()\/\p{Word}]+)/i
```

pablobm:
> Do you wanna make a separate issue for the user dropdown?

@hlfan - Sorry, for a moment I thought you were saying you were going to create the issue yourself. Now I re-read the comment and realised you hadn't. Done now! https://github.com/openstreetmap/openstreetmap-website/issues/6635

tomhughes:
The default timeout is 1s which seems a world away from the times @pablobm is reporting unless the github actions runners are really slow...

lib/rich_text.rb
```diff
@@ -283,7 +282,7 @@ def gsub_pairs_for_linkify_detection
           expanded_path = "#{rule.host || "#{Settings.server_protocol}://#{Settings.server_url}"}/#{rule.path_template}"
           rule.patterns
               .select { |pattern| pattern.is_a?(String) }
-              .map { |pattern| [Regexp.new("(?<=^|#{URL_UNSAFE_CHARS})#{pattern}", Regexp::IGNORECASE), expanded_path] }
+              .map { |pattern| [Regexp.new("(?<=^|#{URL_UNSAFE_CHARS})#{pattern}", Regexp::IGNORECASE, :timeout => 1), expanded_path] }
```
hlfan: Could the default timeout be something else than 1 s here, @tomhughes? Otherwise I don't see a reason this would work.

tomhughes: I don't believe so, at least that seems to be the default for me:
```
bericote [~/src/osm/rails] % bundle exec rails c                
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
Loading development environment (Rails 8.1.2)
open-street-map(dev):001> Regexp.timeout
=> 1.0
```

hlfan: interesting...

hlfan:
> I don't understand the timeout thing but I guess this is good now, thanks.

What's the highest timeout you'd go for? Since the tests keep being flakey I'd raise the timeout to 0.02 or 0.05...

PR #6613, opened 2025-12-11
Stop arguments in path or query from being interpreted as hashes

### LINKED ISSUE #6508: Change Hashtag links to neis-one.org
### Problem

Hi, in my opinion linking hashtags in changesets to https://hashtags-hub.toolforge.org/ is not very helpful. Hashtags on social media and hashtags on OSM serve different purposes.

### Description

In my opinion it would be more helpful to use https://resultmaps.neis-one.org/osm-changesets because it gives you statistics about the usage of the hashtag on OSM.

### Screenshots

_No response_

hlfan:
Remove leading '#' from value in URL template replacement.
Closes #6508.

### DIFF
--- a/lib/tag2link.rb
+++ b/lib/tag2link.rb
@@ -12,7 +12,7 @@ module Tag2link
     url_template = @dict[key]
     return nil unless url_template
 
-    url_template.gsub("$1", value)
+    url_template.gsub("$1", value.sub(/^#/, ""))
   end
 
   def self.build_dict(data)
--- a/test/lib/tag2link_test.rb
+++ b/test/lib/tag2link_test.rb
@@ -16,6 +16,11 @@ class Tag2linkTest < ActiveSupport::TestCase
     assert_equal "https://www.wikidata.org/entity/Q936", url
   end
 
+  def test_link_strips_path_terminators
+    url = Tag2link.link("hashtags", "#maproulette")
+    assert_equal "https://resultmaps.neis-one.org/osm-changesets?comment=maproulette", url
+  end
+
   def test_build_dict_rejects_deprecated_and_third_party
     data = [
       { "key" => "Key:example", "url" => "http://example.com/$1", "rank" => "deprecated", "source" => "osmwiki:P8" },


pablobm:
Test plz! :smile: (Now I'm so glad I made you create `test/lib/tag2link_test.rb` back in https://github.com/openstreetmap/openstreetmap-website/pull/6470!)

tomhughes:
Yes definitely a test, and some more explanation because I'm currently baffled as to what this has to do with how the target for a hashtag link is chosen.

hlfan:
> some more explanation

A #hashtag is led by the eponymous hash, which has a special meaning in URLs.

Because the value is not expected in the hash of the URL, links like
- https://www.color-hex.com/color/#76c551 and
- https://resultmaps.neis-one.org/osm-changesets?comment=#maproulette don't work

With the hash removed, the links to
- https://www.color-hex.com/color/76c551 and
- https://resultmaps.neis-one.org/osm-changesets?comment=maproulette work.

Why? Because in that case, the tag part of the hashtag actually lands where it's supposed to be, in the pathname or the query, not in the hash.

pablobm:
I hoped the test would explain, but it didn't :sweat_smile: I went and had a look. I think I understand now:
- Changeset tag values go through tag2link.
- tag2link feeds from this list: https://github.com/JOSM/tag2link/blob/master/index.json
- Each separate tag in a `hashtags` key in a changeset gets applied this substitution rule:
```
  {
    "key": "Key:hashtags",
    "url": "https://resultmaps.neis-one.org/osm-changesets?comment=$1",
    "source": "osmwiki:P8",
    "rank": "normal"
  },
```
- That link expects the text to replace `$1` not to have a hashtag.
- But we apply the substitution verbatim, passing the hashtag.
- We need to remove the hashtag for that substitution to work, which is what you are doing here.

Your solution would fail if we had a similar tag2link rule that did expect the `#`, as it applies to all rules, not only `Key:hashtags`. Do we know of any such rule? I'm guessing the probability is low, as `#` has a special meaning in URLs as you say.

tomhughes:
Is removing the `#` correct though, or should it be escaped as `%23`?

The answer seems to be that for color-hex.com it's not wanted but for maproulette I would argue that it is, or you're going to match every changeset with that word not just the ones with the hash tag?

hlfan:
Looking at the third-party hashtag links, if a template needs an escaped `#`, it's included in the template. Like here:
[`https://bsky.app/search?q=%23$1`](https://github.com/JOSM/tag2link/blob/40f61430734bd36d40d194d3dd385f29fc0128c4/index.json#L802)

tomhughes:
So arguably it's an error in tag2link that it doesn't preserve the hash in the resultmaps link... But this is good then by the looks of it.

hlfan:
I think technically it preserves the hash in the links, but for the resultmaps link it instantly gets overwritten.

PR #6616, opened 2025-12-14
Fix issue #6614: Support spaces in @mentions using quotes

### LINKED ISSUE #6614: @ syntax with space in user name fails on changeset comment
On https://www.openstreetmap.org/changeset/175272266
we see there is no way to use "@" syntax on
people with space in their usernames.

Image

Apparently the trick is to use `%20` for space.

But alas, there is no GitHub-style @ list to pick from, so it is easy to misspell other peoples names, causing them to not get notified (if indeed they are especially notified,  via these "@"s, in the first place, even if not subscribed.)

aNsHuL5217:
### Description
Fixes #6614.

This PR updates the `RichText#linkify_users` method to support user mentions containing spaces by allowing quoted syntax (e.g., `@"Open Mapper"`).

**Key changes:**
* **Updated Regex:** Now detects mentions enclosed in quotes.
* **Quote Handling:** Supports straight quotes (`"`), smart quotes (`“”`), and escaped entities (`&quot;`), ensuring compatibility with the Markdown renderer.
* **Email Protection:** Added a negative lookbehind `(?<!\w)` to ensure that email addresses (e.g., `user@example.com`) are not accidentally parsed as mentions.
* **Prevented Double-Linking:** The visible `@` symbol is replaced with the HTML entity `&#64;` to prevent subsequent parsers (like `expand_link_shorthands`) from detecting the mention a second time.

### How has this been tested?
* **New Test Case:** Added `test_linkify_username_with_space` to `test/lib/rich_text_test.rb` to verify that quoted mentions are correctly parsed into relative links with safe HTML output.
* **Regression Testing:** Ran the full `RichTextTest` suite (`bundle exec rails test test/lib/rich_text_test.rb`) to confirm that standard mentions (`@user`) and email addresses still function correctly.
* **Linting:** Verified code style compliance using `rubocop`.

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -4,6 +4,7 @@ module RichText
   DESCRIPTION_MAX_LENGTH = 500
   DESCRIPTION_WORD_BREAK_THRESHOLD_LENGTH = 450
   URL_UNSAFE_CHARS = "[^\\w!#$%&'*+,./:;=?@_~^\\-]"
+  LINK_ATTRIBUTES = 'rel="nofollow noopener noreferrer" dir="auto"'
 
   def self.new(format, text)
     case format
@@ -99,12 +100,18 @@ module RichText
     private
 
     def linkify_users(text)
-      text.gsub(/(?<!\w)@(?:(?:&quot;|["“”])(.+?)(?:&quot;|["“”])|(\w+))/) do |match|
-        name = ::Regexp.last_match(1) || ::Regexp.last_match(2)
-        slug = ERB::Util.url_encode(name)
-        safe_match = match.sub("@", "&#64;")
+      regex = /(?<!\w)@(?:“([^”]+?)”|”([^”]+?)”|‘([^’]+?)’|’([^’]+?)’|"([^"]+?)"|'([^']+?)'|(\w+))/
 
-        "<a href=\"/user/#{slug}\" rel=\"nofollow noopener noreferrer\">#{safe_match}</a>"
+      text.gsub(regex) do |match|
+        name = ::Regexp.last_match.captures.compact.first
+        next match unless name
+
+        slug_name = name.tr("“”‘’", %q(""''
+))
+        slug = ERB::Util.url_encode(slug_name)
+
+        safe_match = match.sub("@", "&#64;")
+        "<a href=\"/user/#{slug}\" #{LINK_ATTRIBUTES}>#{safe_match}</a>"
       end
     end
 
@@ -139,8 +146,7 @@ module RichText
     end
 
     def auto_link(text, mode)
-      link_attr = 'rel="nofollow noopener noreferrer" dir="auto"'
-      Rinku.auto_link(text, mode, link_attr) { |url| format_link_text(url) }
+      Rinku.auto_link(text, mode, LINK_ATTRIBUTES) { |url| format_link_text(url) }
     end
 
     def format_link_text(url)
--- a/test/lib/rich_text_test.rb
+++ b/test/lib/rich_text_test.rb
@@ -193,7 +193,6 @@ class RichTextTest < ActiveSupport::TestCase
   end
 
   def test_markdown_table_alignment
-    # Ensure that kramdown table alignment styles are converted to bootstrap classes
     markdown_table = <<~MARKDOWN
       | foo  | bar |
       |:----:|----:|
@@ -226,9 +225,11 @@ class RichTextTest < ActiveSupport::TestCase
   end
 
   def test_linkify_username_with_space
-    text = 'Hello @"Open Mapper"'
-    html = RichText.new("markdown", text).to_html
-    assert_match %r{<a href="/user/Open%20Mapper" rel="nofollow noopener noreferrer">&#64;”Open Mapper”</a>}, html
+    with_settings(:linkify => { :detection_rules => [{ :patterns => ["@(?<username>\\w+)"], :path_template => "user/\\k<username>" }] }) do
+      text = 'Hello @"Open Mapper"'
+      html = RichText.new("markdown", text).to_html
+      assert_match %r{<a href="/user/Open%20Mapper" #{RichText::LINK_ATTRIBUTES}>&#64;[“”]Open Mapper[“”]</a>}o, html
+    end
   end
 
   def test_text_to_html_linkify_replace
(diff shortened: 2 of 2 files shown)


lib/rich_text.rb
```diff
@@ -97,6 +98,16 @@ def linkify(text, mode = :urls)
 
     private
 
+    def linkify_users(text)
+      text.gsub(/(?<!\w)@(?:(?:&quot;|["“”])(.+?)(?:&quot;|["“”])|(\w+))/) do |match|
+        name = ::Regexp.last_match(1) || ::Regexp.last_match(2)
+        slug = ERB::Util.url_encode(name)
+        safe_match = match.sub("@", "&#64;")
+
+        "<a href=\"/user/#{slug}\" rel=\"nofollow noopener noreferrer\">#{safe_match}</a>"
```
hlfan: This should go through `Rinku.auto_link` as well to keep it DRY

aNsHuL5217: Thanks for pointing that out! While Rinku doesn’t provide a builder for arbitrary tags, I noticed I was duplicating link attributes and had also missed dir="auto". I’ve refactored those into a shared LINK_ATTRIBUTES constant used by both linkify_users and auto_link to keep things DRY and consistent.

hlfan:
And how am I supposed to tag this user named ["Yo"](https://www.openstreetmap.org/user/%22Yo%22)?

aNsHuL5217:
Thanks for pointing that out @hlfan ! I realized the original regex didn't account for names containing quotes like "Yo".

I've updated the logic to support single quotes as wrappers, so you can now tag that user simply as @'Yo'.

This update also handles cases where Markdown converts standard quotes into "smart" curly quotes, preventing broken links. I've added a test case to cover this scenario as well.

hlfan (changes requested):

aNsHuL5217:
Sorry for that @hlfan !
To fix the @'tis false positive, would you prefer I restrict the regex (e.g. max 50 chars, no newlines) or implement database validation?

hlfan:
Well, that's the crux of the issue. We couldn't even deny newlines, as there's already one user with \r\n in the name: https://www.openstreetmap.org/user/Mohamedkaizen%0D%0AMohamedkaizen

I think the way to reduce this risk as much as possible is to use a different delimitation character that is more commonly related to URLs than casual text. Right now, the longest name of a contributor is [bwdlqwbdjlbqlwdjbqlwjdbqljwbdlqwjbdlqwjdblqjwbdlqjwdbqlwjdblqwdblqjbdlqjwbdljqbwdljqwbdlqbwdbqwdblqwdlqwbdlbdlqbwldbqlwbdlqbwdlqbdlqbdljwbdlqwbdwlbqlwdblqwbdjlqwbdjlqwbdljqwbdljqbwdljqbdlqwbdlqbdljqbdljqbwdljbqldbqlwdbljqwbdljqwdbqlwdbqlwjdbqlwdbqlw](https://www.openstreetmap.org/user/bwdlqwbdjlbqlwdjbqlwjdbqljwbdlqwjbdlqwjdblqjwbdlqjwdbqlwjdblqwdblqjbdlqjwbdljqbwdljqwbdlqbwdbqwdblqwdlqwbdlbdlqbwldbqlwbdlqbwdlqbdlqbdljwbdlqwbdwlbqlwdblqwbdjlqwbdjlqwbdljqwbdljqbwdljqbdlqwbdlqbdljqbdljqbwdljbqldbqlwdbljqwbdljqwdbqlwdbqlwjdbqlwdbqlw) at 249 chars, so none of these suggestions would be inclusive.

And database validation is probably too much of a slowdown for how often this library is called.

HolgerJeromin:
Do we really have to fix the issue for **all**? 
Or it is OK to break the three funny users which have ", linebreaks and other stuff in it?
I would keep it simple and ignore them.

we could keep them in (negative) unit test (with explaining comments) to declare the explicit simplicity.

hlfan:
But where do you draw the line? Do you want users to need to escape whitespace or not?

lib/rich_text.rb
```diff
@@ -97,6 +99,22 @@ def linkify(text, mode = :urls)
 
     private
 
+    def linkify_users(text)
+      regex = /(?<!\w)@(?:“([^”]+?)”|”([^”]+?)”|‘([^’]+?)’|’([^’]+?)’|"([^"]+?)"|'([^']+?)'|(\w+))/
```
pnorman, line 103: Is there some way to match that doesn't involve an unreadable regular expression?

pnorman: There's also a lot of logic in this regular expression that is not tested. It looks like it matches
- `“”`
- `””`
- `‘’`
- `’’`
- `""`
- `''`

Should we have a test that shows that it doesn't linkify something like `@"foo'`? What happens if there are multiple potential matches? e.g. `@“foo@”bar”`

HolgerJeromin:
> But where do you draw the line? Do you want users to need to escape whitespace or not?

IMO spaces could be supported. But we should ignore other white space or `"` (which will produce wrong links, yes)

pablobm:
I think we should go in a different direction. This solution would require users to learn a new syntax that is not obvious, particularly as it's not a common convention that they may have learned at other apps. And that's before going into the side-effects that are being discussed above.

I have created an issue at https://github.com/openstreetmap/openstreetmap-website/issues/6635, describing what I think we should do instead.

aNsHuL5217:
@pnorman I gave the `@"foo'` edge case a shot, but getting it to work safely (without breaking URLs or HTML) made the regex extremely complex and unreadable.

I'm thinking of sticking with the simpler implementation. In my opinion, the trade-off of having hard-to-maintain code isn't worth it for such a rare edge case. The current version handles all standard and smart quotes correctly.

I've reverted to the cleaner version—let me know if you're okay with this approach!

aNsHuL5217:
@pablobm Thanks for linking that issue! I just read through the discussion in #6635.

Honestly, while a proper dropdown/User-ID system would be the ideal solution eventually, it looks like it's facing some pretty massive blockers right now. As 1ec5 and TomHughes pointed out, we don't have a User Search API, and moving away from plain-text comments to support hidden IDs would likely require a full Rich Text Editor overhaul.

I think this PR is the best path forward for now because it solves the bug immediately without needing a massive architectural rewrite. We can always upgrade to the "perfect" system later if those technical hurdles get cleared, but I don't think we should block this fix while waiting for that to happen.

PR #6600, opened 2025-12-08
Add a donate button to the navbar

Robot8A:
Would fix #6517

I went for an embed of https://supporting.openstreetmap.org/ on /donate. This way, the user can always return to the map or other parts of the website.

We can discuss whether this is the most appropriate way to do it or whether others would be better. Also, it's my first time programming with Ruby on Rails, so I hope I did everything in a way that makes sense :)

PR #6617, opened 2025-12-15
Add Donate to navigation bar

### LINKED ISSUE #6517: Donate button is hard to find

1ec5:
Added a Donate link to the sitewide navigation bar. Similar to #6600, but supporting.openstreetmap.org opens in a new tab instead of an iframe. Pulled the URL to the donation site into a configuration variable since it doesn’t need to be localized per occurrence.

Before | After
----|----
History, Export, GPS Traces, User Diaries, Communities, Copyright, Help, and About | History, Export, GPS Traces, User Diaries, Communities, Copyright, and Help; Donate and About in More overflow menu

I considered creating a landing page as suggested in https://github.com/openstreetmap/openstreetmap-website/pull/6600#issuecomment-3638068481, but it wasn’t clear what the landing page should say that isn’t already on the landing page at supporting.openstreetmap.org. So I just kept things as simple as possible.

Fixes #6517.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -139,6 +139,8 @@ fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
 # Endpoints for Wikimedia integration
 wikidata_api_url: "https://www.wikidata.org/w/api.php"
 wikimedia_commons_url: "https://commons.wikimedia.org/wiki/"
+# URL to donation form
+donation_url: "https://supporting.openstreetmap.org/donate/"
 # Linkification rules
 linkify:
   detection_rules:
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -62,6 +62,9 @@
         <%= link_to t("layouts.help"), help_path, :class => header_nav_link_class(help_path) %>
       </li>
       <li class="nav-item">
+        <%= link_to t("layouts.donate"), Settings.donation_url, :class => header_nav_link_class(Settings.donation_url), :target => :new %>
+      </li>
+      <li class="nav-item">
         <%= link_to t("layouts.about"), about_path, :class => header_nav_link_class(about_path) %>
       </li>
       <li id="compact-secondary-nav" class="dropdown nav-item ms-auto">
--- a/app/views/site/copyright.html.erb
+++ b/app/views/site/copyright.html.erb
@@ -43,9 +43,10 @@
       <%= legal_babble_paragraph "lead_1" %>
     </p>
     <p class="lead">
-      <%= legal_babble_paragraph "lead_2", :links => %w[learn_more_about_osm osm_blog weeklyosm osm_foundation making_donation],
+      <%= legal_babble_paragraph "lead_2", :links => %w[learn_more_about_osm osm_blog weeklyosm osm_foundation],
                                            :local_links => { :get_started_mapping => new_user_path,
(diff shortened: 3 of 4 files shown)


1ec5:
If we do need a landing page, as @gravitystorm suggested in https://github.com/openstreetmap/openstreetmap-website/pull/6600#issuecomment-3638068481, then #6600 would be a better starting point (which is why I haven’t closed it), but we’d still need to clean it up a bit. Otherwise, this PR would settle for a simple link for now. I think we could start with this approach first and adopt something more full-fledged later on without breaking any URLs.

config/settings.yml
```diff
@@ -140,7 +140,7 @@ fossgis_valhalla_url: "https://valhalla1.openstreetmap.de/route"
 wikidata_api_url: "https://www.wikidata.org/w/api.php"
 wikimedia_commons_url: "https://commons.wikimedia.org/wiki/"
 # URL to donation form
-donation_url: "https://supporting.openstreetmap.org/"
+donation_url: "https://supporting.openstreetmap.org/donate/"
```
1ec5, line 143: I changed the link to go directly to [the donation form](https://supporting.openstreetmap.org/donate/) instead of [the landing page](https://supporting.openstreetmap.org/). I think this is a better match for the user’s intent when they click the Donate button in the navigation bar. In general, any extraneous step standing in the way of a signup or donation flow will tend to lose a certain share of users, which we don’t want. The donation form also mentions opportunities to become a member, but making a donation is the focus.

pablobm (approved):

CommanderStorm (approved):
This seems to work as expected, mostly (I am having a few setup issues, but I think those are due to https://community.openstreetmap.org/t/more-technical-updates-to-the-tile-openstreetmap-org-service-openstreetmap-org-standard-layer/141293).

tomhughes (changes requested):
Can we rebase this and squash the commits together please? or maybe rework the commits so the first one makes the URL configurable and a second one adds the button to the navigation bar?

1ec5:
> or maybe rework the commits so the first one makes the URL configurable and a second one adds the button to the navigation bar?

Done.

tomhughes (approved):
I this looks good now, thanks.

PR #6623, opened 2025-12-16
Pull changeset comment rich text into separate method

### LINKED ISSUE #6615: Changeset URLs in edit comments are no longer linked
### URL

https://www.openstreetmap.org/changeset/175863007

I'm fairly certain that previously when I posted changeset URLs into my edit comments, they were transformed into clickable links on the website, but this seems to be no longer the case:
Image

In changeset comments and Map note discussions, they are still linkified as usual, and now also shortened, see for example <https://www.openstreetmap.org/changeset/173737130> :
Image

BTW thanks for the recent linkification improvements regarding OSM URLs and user names, I think this makes discussions much more readable! 🙂

_No response_

hlfan:
Addresses https://github.com/openstreetmap/openstreetmap-website/pull/6597#discussion_r2601656153 and fixes #6615 by adding a separate method to the changeset model that _actually_ outputs HTML.

I also added a test checking for changeset comment linkification to prevent repeats of #6615.

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -201,6 +201,7 @@ class Changeset < ApplicationRecord
   end
 
   def comment = tags["comment"].presence
+  def comment_html = comment ? RichText.new("text", comment).to_html : nil
 
   ##
   # set the auto-close time to be one hour in the future unless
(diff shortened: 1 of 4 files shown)


deevroman:
Is an additional paragraph expected to appear before the comment?



(if I'm not mistaken, this is a consequence of this PR)

hlfan:
And another one after it because they are [block-level elements](https://developer.mozilla.org/en-US/docs/Glossary/Block-level_content) and I totally didn't think about this.

PR #6625, opened 2025-12-17
Automatically generate factories instead of fixtures

gravitystorm:
From the factory_bot_rails README:

> Including factory_bot_rails in the development group of your Gemfile
> will cause Rails to generate factories instead of fixtures.

This is helpful for any developers running "rails generate model {modelname}".

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -171,7 +171,6 @@ group :test do
   gem "brakeman"
   gem "capybara", ">= 2.15"
   gem "erb_lint", :require => false
-  gem "factory_bot_rails"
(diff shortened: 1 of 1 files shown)


gravitystorm:
> Gemfile was updated, but Gemfile.lock wasn't updated. Usually, when Gemfile is updated, you should run `bundle install` to update Gemfile.lock.

In this case, this `Gemfile.lock` warning can be ignored, because there are no gems being added or removed from the overall bundle.

tomhughes:
Looks good to me, thanks.

PR #6627, opened 2025-12-17
Refactor turbo-frame element to avoid multiline conditionals inside html attribute

gravitystorm:
While the original is perfectly valid erb, it's hard to read and hard for erblint to parse.

This refactor moves the multi-line conditional logic outside of the turbo-frame `src` html attribute.

This is needed before we can introduce the erblint [HardCodedString](https://github.com/Shopify/erb_lint/?tab=readme-ov-file#hardcodedstring) linter

### How has this been tested?

I checked with a local dev instance to make sure that the turbo frame worked with a relation with > 10 members


tomhughes:
That looks nicer, thanks.

PR #6629, opened 2025-12-17
Send bulk user confirmations through the state machine

tomhughes:
This ensure that bulk actions performed on users from the list screen go through the state machine and trigger any additional actions (not that there are any currently for confirm) rather than updating the database directly.


pablobm (approved):

gravitystorm:
Merged, thanks!

PR #6631, opened 2025-12-18
Swap confirm and suspend actions

pablobm:
As requested by @Firefishy at #osmf-operations. Also coloured the "suspend" action in danger red.

Links to Confirm or Suspend a user. The second one in red to express danger

Something that annoys me of these action links: the confirmation modal is common to both (and other links that can appear here) and simply reads "Confirm" in English. I would have preferred a descriptive "Suspend user %{user_name}" (etc) or similar, to further help prevent an accident. Thoughts? It's a bit of a faff, with all the i18n strings, wouldn't do it as part of this PR anyway.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -194,7 +194,7 @@
 
             <% if @user.may_suspend? %>
               <li>
-                <%= link_to t(".suspend_user"), user_status_path(@user, :event => "suspend"), :method => :put, :data => { :confirm => t(".confirm") } %>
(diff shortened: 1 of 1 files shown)


hlfan:
What do you think of the `link-success` class for the unsuspend action?

pablobm:
It crossed my mind, but I'm not sure if it would be appropriate. It might be, but I can't quite decide.

Firefishy:
I'd prefer to see this go live soon, any enhancements can come later.

PR #6632, opened 2025-12-18
Fix invalid <p> nesting in comment rendering

hlfan:
Fixes https://github.com/openstreetmap/openstreetmap-website/pull/6623#issuecomment-3669442629 by not using paragraphs to wrap the rich text content invalidly. Thanks @deevroman for pointing this out!


tomhughes:
Looks good to me, thanks.

PR #6654, opened 2025-12-25
Fix incorrect z-order of dashboard map markers (fixes #6639)

### LINKED ISSUE #6639: Markers on dashboard map have incorrect z-order
### URL

https://www.openstreetmap.org/dashboard

As of #6504, the markers on the dashboard map are sorted in z-order based on when they were added to the map. The markers mimic physical pins popping up out of the map, replete with drop shadow. This illusion breaks when a marker is located to the north of another marker but overlaps it.

Originally [reported](https://en.osm.town/@rjw62/115725913344089357) by @rjw62.

aNsHuL5217:
### Description
Fixes #6639

The dashboard map markers were previously rendered in the order they appeared in the DOM (which matches the text list order). This caused visual z-order issues where "Northern" markers (background) were sometimes drawn on top of "Southern" markers (foreground) if they appeared later in the list.

This PR fixes the issue by:
1. Collecting all user data from the DOM first.
2. Sorting the users by latitude in descending order (North to South).
3. Adding the markers to the map in this sorted order.

This ensures that Northern markers are always rendered first (in the background) and Southern markers are rendered last (in the foreground), preserving the correct visual perspective.

**Screenshot (After Fix):**
image (2)

### How has this been tested?
I tested this locally in the development environment:
1. Created two users with overlapping coordinates using the Rails console:
   * **User A (Friend):** Latitude `51.501` (North)
   * **User B (Me):** Latitude `51.5` (South)
2. Verified on the Dashboard map that **User B (South)** is correctly drawn **in front of** User A (North).

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -12,6 +12,11 @@ $(function () {
       style: OSM.MapLibre.Styles.Mapnik,
       attributionControl: false,
       locale: OSM.MapLibre.Locale,
+      rollEnabled: false,
+      dragRotate: false,
+      pitchWithRotate: false,
+      bearingSnap: 180,
+      maxPitch: 0,
       center: OSM.home ? [OSM.home.lon, OSM.home.lat] : [0, 0],
       zoom: OSM.home ? defaultHomeZoom : 0
     });
@@ -25,15 +30,46 @@ $(function () {
       trackUserLocation: true
     });
     map.addControl(new OSM.MapLibre.CombinedControlGroup([navigationControl, geolocateControl]), position);
+    map.touchZoomRotate.disableRotation();
+    map.keyboard.disableRotation();
 
-    $("[data-user]").each(function () {
-      const user = $(this).data("user");
-      if (user.lon && user.lat) {
-        OSM.MapLibre.getMarker({ icon: "dot", color: user.color })
-          .setLngLat([user.lon, user.lat])
-          .setPopup(OSM.MapLibre.getPopup(user.description))
-          .addTo(map);
+    const markerObjects = $("[data-user]")
+      .filter(function () {
+        const { lat, lon } = $(this).data("user");
+        return lat && lon;
+      })
+      .map(function () {
+        const { lat, lon, color, description } = $(this).data("user");
+
+        const marker = OSM.MapLibre.getMarker({ icon: "dot", color })
+          .setLngLat([lon, lat])
+          .setPopup(OSM.MapLibre.getPopup(description));
+
+        return { marker, lat, lon };
+      })
+      .get();
+
+    for (const item of markerObjects) {
+      item.marker.addTo(map);
+    }
+
+    const updateZIndex = () => {
+      for (const item of markerObjects) {
+        item.currentY = map.project([item.lon, item.lat]).y;
       }
-    });
+
+      markerObjects.sort((a, b) => a.currentY - b.currentY);
+
+      for (const [index, item] of markerObjects.entries()) {
+        item.marker.getElement().style.zIndex = index;
+      }
+    };
+
+    if (markerObjects.length > 0) {
+      map.on("move", updateZIndex);
+      map.on("rotate", updateZIndex);
+      map.on("pitch", updateZIndex);
+      updateZIndex();
+    }
   }
 });


hlfan:
> Sorting the users by latitude in descending order

If we need a solution that works with a rotated map for notes, shouldn't we try to use the same system here even if we don't need rotation here?
The main map PR will be large enough anyway.

aNsHuL5217:
> > Sorting the users by latitude in descending order
> 
> If we need a solution that works with a rotated map for notes, shouldn't we try to use the same system here even if we don't need rotation here? The main map PR will be large enough anyway.

Good point. I hadn't considered rotation since it's locked here, but using the Notes system is a better long-term fix. I'll investigate how that works and update this to match

aNsHuL5217:
I’ve updated the code to handle the Z-ordering correctly.

To solve the sorting problem, I added a listener that updates the `zIndex` based on the marker's Y-position on the screen whenever the map moves or rotates. This keeps the "closest" markers on top without breaking the visuals.

Here is a quick video showing the rotation behavior:

https://github.com/user-attachments/assets/c0b4c048-3c75-4ec8-bfdd-5f877eb41d7e

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,40 @@ $(function () {
...
 
+    const markerObjects = [];
+
     $("[data-user]").each(function () {
       const user = $(this).data("user");
       if (user.lon && user.lat) {
-        OSM.MapLibre.getMarker({ icon: "dot", color: user.color })
-          .setLngLat([user.lon, user.lat])
+        const lat = parseFloat(user.lat);
+        const lon = parseFloat(user.lon);
```
hlfan: We shouldn't need to parse here as MapLibre casts again anyway.

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,40 @@ $(function () {
     map.touchZoomRotate.disableRotation();
     map.keyboard.disableRotation();
 
+    const markerObjects = [];
+
     $("[data-user]").each(function () {
```
hlfan: Could we make this more pipeliney?

```javascript
const markerObjects = $("[data-user]")
  .filter(...)
  .map(...)
  .get();
```

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,40 @@ $(function () {
...
+        markerObjects.push({ marker: marker, lat: lat, lon: lon });
       }
     });
+
+    const updateZIndex = () => {
+      markerObjects.forEach((item) => {
+        const point = map.project([item.lon, item.lat]);
+        const zIndex = Math.round(point.y);
+        item.marker.getElement().style.zIndex = zIndex;
+      });
```
hlfan: I don't really like that this assigns values in the hundreds for a few dozen markers.
Some method of ending up with a more sequential z-index list would be great.

app/assets/javascripts/dashboard.js
aNsHuL5217: I dropped the parseFloat since MapLibre handles the strings fine on its own. Passing them raw now.

app/assets/javascripts/dashboard.js
aNsHuL5217: Done. I switched the loop over to a `.map(...).get()` chain. It's definitely cleaner to build the array in one pass rather than pushing to a list manually.

app/assets/javascripts/dashboard.js
aNsHuL5217: That makes sense. I updated the logic to sort the markers by Y-position first, then just assigned a simple index (0, 1, 2...) based on that order.

hlfan: That's actually better than the idea I had, since mutating the array keeps the list at least nearly sorted.

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,38 @@ $(function () {
...
+          icon: "dot",
+          color: user.color
+        })
           .setLngLat([user.lon, user.lat])
           .setPopup(OSM.MapLibre.getPopup(user.description))
           .addTo(map);
-      }
-    });
+
+        return { marker: marker, lat: user.lat, lon: user.lon };
```
hlfan: While we're at it, we could destructure `user` to make this call look like this:

```suggestion
        return { marker, lat, lon };
```

aNsHuL5217: Agreed, that is much cleaner. I've destructured `lat` and `lon` from the user object so we can use the shorthand syntax in the return statement.

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,40 @@ $(function () {
...
+        return { marker, lat, lon };
+      })
+      .get();
+
+    const updateZIndex = () => {
+      markerObjects.forEach((item) => {
+        item.currentY = map.project([item.lon, item.lat]).y;
+      });
+      markerObjects.sort((a, b) => a.currentY - b.currentY);
+      markerObjects.forEach((item, index) => {
```
1ec5: Nit: It looks like other files are using for-of loops instead of `forEach`. Maybe we should be consistent for readability.

aNsHuL5217: I have replaced the `forEach` with `for-of `loops.

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,44 @@ $(function () {
...
+          .setLngLat([lon, lat])
           .setPopup(OSM.MapLibre.getPopup(user.description))
           .addTo(map);
+
+        return { marker, lat, lon };
+      })
+      .get();
+
+    const updateZIndex = () => {
+      // REVIEW FIX: Use for-of loop instead of forEach
```
HolgerJeromin:
```suggestion
```

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,44 @@ $(function () {
...
+      for (const item of markerObjects) {
+        item.currentY = map.project([item.lon, item.lat]).y;
       }
-    });
+
+      markerObjects.sort((a, b) => a.currentY - b.currentY);
+
+      // REVIEW FIX: Use for-of with entries() to get index
```
HolgerJeromin:
```suggestion
```

aNsHuL5217: I just left those in to help track the specific changes during the review back-and-forth. I've cleared them.

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,42 @@ $(function () {
...
+        const { lat, lon } = user;
+
+        if (!lat || !lon) return null;
+
+        const marker = OSM.MapLibre.getMarker({
+          icon: "dot",
+          color: user.color
+        })
+          .setLngLat([lon, lat])
           .setPopup(OSM.MapLibre.getPopup(user.description))
```
hlfan:
```suggestion
        const { lat, lon, color, description } = $(this).data("user");

        if (!lat || !lon) return null;

        const marker = OSM.MapLibre
          .getMarker({ icon: "dot", color })
          .setLngLat([lon, lat])
          .setPopup(OSM.MapLibre.getPopup(description))
```

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,37 @@ $(function () {
...
+
+        if (!lat || !lon) return null;
+        const marker = OSM.MapLibre.getMarker({ icon: "dot", color })
+          .setLngLat([lon, lat])
+          .setPopup(OSM.MapLibre.getPopup(description))
           .addTo(map);
+
+        return { marker, lat, lon };
+      })
+      .get();
```
pablobm, line 50: I like that you use `map` to collect the markers and their lat/lon. However I expect `map` not to have side effects, and here it has one (the `addTo` call).

How about first creating and collecting the markers, as you do here, and then have a separate loop to call `addTo` on each of them?

app/assets/javascripts/dashboard.js
```diff
@@ -33,14 +33,37 @@ $(function () {
...
-      const user = $(this).data("user");
-      if (user.lon && user.lat) {
-        OSM.MapLibre.getMarker({ icon: "dot", color: user.color })
-          .setLngLat([user.lon, user.lat])
-          .setPopup(OSM.MapLibre.getPopup(user.description))
+    const markerObjects = $("[data-user]")
+      .map(function () {
+        const { lat, lon, color, description } = $(this).data("user");
+
+        if (!lat || !lon) return null;
```
pablobm: I think this is better done as a `filter()` before the `map()`.

app/assets/javascripts/dashboard.js
aNsHuL5217: I agree that keeping `.map()` pure is better for maintainability. I've refactored this to strictly create the marker objects in the map, and added a separate loop afterwards to handle the `.addTo(map)` side effect.

app/assets/javascripts/dashboard.js
aNsHuL5217: I added a `.filter()` step before the map to check for valid `lat` and `lon` data.

hlfan (approved):
Thanks for your work on this.

PR #6657, opened 2025-12-26
Fix heatmap tooltip date offset

### LINKED ISSUE #6378: Timezone mismatch between heatmap data and tooltips
### URL

_No response_

On the contributions heatmap shown on user profile pages, the rows and columns appear to be aligned to UTC. However the date shown in the hover tooltip is localized to the browser's time zone. This can lead to a mismatch between the data being visualized and the date being displayed.

For example, today is Thursday September 4, but when I hover over the Thursday row for this week on my heatmap, the tooltip says "September 3" when I view it from UTC-4:00.

Image

I think the tooltip dates should use UTC to match the grid. Alternatively the grid could be aligned to the browser's local time zone, but I assume that would be a much more complicated solution.

_No response_

hlfan:
Closes #6378 by letting i18n do the date conversion by itself.

### DIFF
--- a/app/assets/javascripts/heatmap.js
+++ b/app/assets/javascripts/heatmap.js
@@ -39,7 +39,7 @@ $(document).on("turbo:frame-load", function () {
       continue;
     }
     const count = $day.data("count") ?? 0;
-    const tooltipText = getTooltipText(date, count);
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #6661, opened 2025-12-30
SpamScorer: normalise whitespace in comparable form

Firefishy:
### Description
Collapse Unicode whitespace runs to a single space in SpamScorer::RichText#to_comparable_form to improve SpammyPhrase matching across newlines/tabs.

### How has this been tested?
Suitable tests have been added. Will be run as part of CI.

### DIFF
--- a/lib/spam_scorer/rich_text.rb
+++ b/lib/spam_scorer/rich_text.rb
@@ -38,7 +38,7 @@ module SpamScorer
     attr_reader :text
 
     def to_comparable_form(str)
-      str.downcase(:fold).unicode_normalize(:nfkc)
+      str.downcase(:fold).unicode_normalize(:nfkc).gsub(/\s+/u, " ")
     end
   end
 end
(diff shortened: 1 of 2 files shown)


Firefishy:
I've removed the `.strip` because we may want to use spaces in spammy_phrase to avoid "scunthorpe problem" matches.

tomhughes:
WTF is copilot doing reviewing PRs here? I made sure never to turn any of that on I thought...

Firefishy:
> WTF is copilot doing reviewing PRs here? I made sure never to turn any of that on I thought...

I clicked the "Copilot" reviewer top right of this PR. I was surprised it was available.

pablobm:
> I was surprised it was available.

My understanding is that [there is no way to disable the option to request a Copilot review](https://github.com/orgs/community/discussions/169148), only to [disable Copilot from automatically reviewing every PR](https://github.com/orgs/community/discussions/155923), which I guess is what is disabled here.

tomhughes:
What you can do is disable copilot's access to a repository which is supposed to effectively make it unable to do reviews but it doesn't seem to have worked.

Blocking the copilot review user also seems to error out.

gravitystorm:
Merged, thanks!

PR #6669, opened 2026-01-02
Add info icon with Wiki links to map layers menu

### LINKED ISSUE #6256: Add "info" icon to layers in Map Layers menu
### Problem

Now we have 8 layers being presented in the May Layers menu, it would be nice to have readily available information about the layers. For example: main aim of the layer, raster vs vector, update frequency, link to the layer's GitHub repo etc.

### Description

If possible, this could be via an "info" button on each layer's preview in the Map Layers menu, which when hovered on or clicked opens up a the info message for that layer.

### Screenshots

Image

(Apologies, this was just hand drawn but I'm sure you get the idea already anyway)

SinghSwayam:
### Description
This PR Fixes #6256 by adding an "info" icon next to the layer names in the "Map Layers" sidebar. This allows users to access the wiki page for each layer to understand its purpose. 

### Implementation Details
- **Inline SVG:** Injected a lightweight SVG info icon directly into the layer control in `leaflet.layers.js`.
- **Wiki Linking:** Mapped all 8 production layer IDs (including `cyclemap`, `transportmap`, and `openmaptiles_osm`) to their respective OpenStreetMap Wiki URLs.
- **Event Handling:** Added `e.stopPropagation()` to ensure clicking the info icon opens the wiki link without triggering a map layer switch.
- **UI/UX Polish:**
    - Positioned the icon in the top-right corner of the layer preview button.
    - Used `mix-blend-mode: difference` CSS. This ensures the icon automatically inverts colors to be visible on both light backgrounds (Standard) and dark backgrounds (Shortbread) without needing complex logic.

### Verification & Testing
I verified the implementation locally with the following steps:
1. **Visual Check:** Confirmed the icon is visible and aligned correctly on the Standard layer.
2. **Contrast Check:** Confirmed the icon remains visible on the dark "Shortbread" layer due to the `mix-blend-mode` setting.
3. **Functionality:** Verified that clicking the icon opens the correct Wiki page in a new tab, while clicking the text label still switches the map layer.
4. Third-Party Layers: I configured local API keys to verify that icons appear and link correctly for third-party layers (Cycle Map, Transport Map, MapTiler) that are not enabled by default.

### Screenshots
Before:
image

After:
image

### DIFF
--- a/app/views/layers_panes/show.html.erb
+++ b/app/views/layers_panes/show.html.erb
@@ -7,6 +7,11 @@
         <span class="badge position-absolute top-0 start-0 rounded-top-0 rounded-start-0 py-1 px-2 bg-body bg-opacity-75 text-body text-wrap text-start fs-6 lh-base">
           <%= t "javascripts.map.base.#{layer['nameId']}" %>
         </span>
+        <% if layer["wiki"] %>
+          <%= link_to layer["wiki"], :target => "_blank", :rel => "noopener", :title => t("layouts.layer_info"), :class => "layer-info-toggle position-absolute z-3 p-1 end-0 text-white text-decoration-none lh-1" do %>
+            <i class="bi bi-info-circle"></i>
+          <% end %>
+        <% end %>
       </label>
     </div>
   <% end %>
--- a/config/layers.yml
+++ b/config/layers.yml
@@ -2,6 +2,7 @@
   code: "M"
   layerId: "mapnik"
   nameId: "standard"
+  wiki: "https://wiki.openstreetmap.org/wiki/OpenStreetMap_Carto"
   canEmbed: true
   canDownloadImage: true
   credit:
@@ -13,6 +14,7 @@
   code: "Y"
   layerId: "cyclosm"
   nameId: "cyclosm"
+  wiki: "https://wiki.openstreetmap.org/wiki/CyclOSM"
   canEmbed: true
   credit:
     id: "cyclosm_credit"
@@ -28,6 +30,7 @@
   code: "C"
   layerId: "cyclemap"
   nameId: "cycle_map"
+  wiki: "https://wiki.openstreetmap.org/wiki/OpenCycleMap"
   apiKeyId: "thunderforest_key"
   canEmbed: true
   canDownloadImage: true
@@ -43,6 +46,7 @@
   code: "T"
   layerId: "transportmap"
   nameId: "transport_map"
+  wiki: "https://wiki.openstreetmap.org/wiki/Transport_Map"
   apiKeyId: "thunderforest_key"
   canEmbed: true
   canDownloadImage: true
@@ -57,6 +61,7 @@
   code: "P"
   layerId: "tracestracktopo"
   nameId: "tracestracktop_topo"
+  wiki: "https://wiki.openstreetmap.org/wiki/Tracestrack"
   apiKeyId: "tracestrack_key"
   credit:
     id: "tracestrack_credit"
@@ -69,6 +74,7 @@
   code: "H"
   layerId: "hot"
   nameId: "hot"
+  wiki: "https://wiki.openstreetmap.org/wiki/Humanitarian"
   canEmbed: true
   credit:
     id: "hotosm_credit"
@@ -85,6 +91,7 @@
   code: "S"
   layerId: "shortbread"
   nameId: "shortbread"
+  wiki: "https://wiki.openstreetmap.org/wiki/Shortbread"
   maxZoom: 23
   credit:
     id: "make_a_donation"
@@ -95,6 +102,7 @@
   code: "V"
   layerId: "openmaptiles_osm"
   nameId: "openmaptiles_osm"
+  wiki: "https://wiki.openstreetmap.org/wiki/MapTiler"
   apiKeyId: "maptiler_key"
   maxZoom: 23
   credit:
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -1033,3 +1033,9 @@ img.trace_image {
     min-width: 200px;
   }
 }
+
+/* Map Layers Menu - Info Icon */
+
+.layer-info-toggle {
+  mix-blend-mode: difference;
+}
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1846,6 +1846,7 @@ en:
       title: OpenStreetMap
     logo:
       alt_text: OpenStreetMap logo
+    layer_info: "Layer Info"
     home: Go to Home Location
     logout: Log Out
     log_in: Log In


mmd-osm:
I suppose  layerLinks would have to be managed in https://github.com/openstreetmap/openstreetmap-website/blob/master/config/layers.yml, rather than hardcoding them in Javascript.

There are a few other issues here, so best wait for more feedback before changing anything.

app/assets/javascripts/leaflet.layers.js
hlfan, line 1: This shouldn't need to involve JS at all. Start here instead:

https://github.com/openstreetmap/openstreetmap-website/blob/65936c9e5b87a47e7a819a59e605f0a3e6bfffd5/app/views/layers_panes/show.html.erb#L6-L10

app/assets/javascripts/leaflet.layers.js
```diff
@@ -15,6 +15,47 @@ L.OSM.layers = function (options) {
...
+      };
+
+      const url = layerLinks[layer.options.layerId];
+      
+      if (url) {
+        const link = document.createElement("a");
+        link.href = url;
+        link.target = "_blank";
+        link.rel = "noopener";
+        link.title = "Layer Info";
```
HolgerJeromin: Must be translatable

SinghSwayam:
Thank you @mmd-osm, @HolgerJeromin, and @hlfan for the detailed feedback!
I agree that handling this in the View layer is much cleaner than using JavaScript. I will rework the PR to:
* Move the Wiki URLs into config/layers.yml.
* Render the icon directly in app/views/layers_panes/show.html.erb using Rails tags.
* Add the translation key to config/locales for accessibility.

I am working on these changes now and will push an update shortly.

SinghSwayam:
@mmd-osm @HolgerJeromin @hlfan Thank you for the guidance! I have refactored the implementation as requested.

**Changes in this update:**
1.  Moved all Wiki URLs into `config/layers.yml`.
2.  Added the translation key `layer_info` to `config/locales/en.yml`.
3.  Implemented the icon rendering directly in `app/views/layers_panes/show.html.erb`.
4.  Completely removed the JavaScript logic from `leaflet.layers.js`.

It is ready for a re-review when you have a moment. Thanks!

mmd-osm:
Instead of using your own svg, you could check out https://icons.getbootstrap.com/icons/info-circle/ as `class="bi bi-info-circle"`

app/views/layers_panes/show.html.erb
```diff
@@ -7,6 +7,20 @@
         <span class="badge position-absolute top-0 start-0 rounded-top-0 rounded-start-0 py-1 px-2 bg-body bg-opacity-75 text-body text-wrap text-start fs-6 lh-base">
           <%= t "javascripts.map.base.#{layer['nameId']}" %>
         </span>
+
+        <% if layer['wiki'] %>
+          <a href="<%= layer['wiki'] %>"
```
hlfan: This should use the link to helper for consistency.

app/views/layers_panes/show.html.erb
```diff
@@ -7,6 +7,20 @@
...
+             target="_blank"
+             rel="noopener"
+             title="<%= t 'layouts.layer_info' %>"
+             class="position-absolute z-3"
+             style="top: 4px; right: 4px; color: #ffffff90; mix-blend-mode: difference; text-decoration: none; line-height: 1;"
```
hlfan: This has to come from a stylesheet to apply outside of development.

app/views/layers_panes/show.html.erb
```diff
@@ -7,6 +7,20 @@
...
+             onclick="event.stopPropagation()">
```
hlfan: 1. Is this really needed? Shouldn't the link clicking action consume the event already?
2. This would also be blocked by the CSP in production, so this handler would need to be added in the pane JS file.

app/views/layers_panes/show.html.erb
```diff
@@ -7,6 +7,20 @@
...
+             <svg width="20" height="20" viewBox="0 0 512 512" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
hlfan: You can use `i` elements for bootstrap icons without needing to copy the SVG.

SinghSwayam:
@hlfan You were right—I tested it locally, and the link click consumes the event automatically, so the map layer does not switch when clicking the icon.

I have removed the JavaScript entirely. I also moved the styles to `common.scss` and switched to the standard Bootstrap `bi-info-circle` icon as requested.

The implementation is now just the View change and the CSS. Ready for re-review!

hlfan:
I think rtl layout support still needs some consideration; other than that, no more glaring things.

app/views/layers_panes/show.html.erb
SinghSwayam: Implemented the bootstrap `bi-info-circle`

app/views/layers_panes/show.html.erb
SinghSwayam: yes i have checked the link clicking action and it consumes the event

app/views/layers_panes/show.html.erb
SinghSwayam: Implemeted within the stylesheet

app/views/layers_panes/show.html.erb
SinghSwayam: implemented link_to

SinghSwayam:
@hlfan I've updated the CSS to use the logical property `inset-inline-end: 4px` instead of `right: 4px`.

HolgerJeromin (approved):

config/layers.yml
```diff
@@ -13,6 +14,7 @@
   code: "Y"
   layerId: "cyclosm"
   nameId: "cyclosm"
+  wiki: "https://wiki.openstreetmap.org/wiki/CyclOSM"
```
HolgerJeromin, line 17: > we could change the link to `http://osm.wiki/xy`

This would need another DNS resolve and hop for the browser to visit.

pnorman: Is there a need to make it less verbose? Anyone clicking on it will end up in the same place

app/views/layers_panes/show.html.erb
```diff
@@ -7,6 +7,11 @@
         <span class="badge position-absolute top-0 start-0 rounded-top-0 rounded-start-0 py-1 px-2 bg-body bg-opacity-75 text-body text-wrap text-start fs-6 lh-base">
           <%= t "javascripts.map.base.#{layer['nameId']}" %>
         </span>
+        <% if layer["wiki"] %>
+          <%= link_to layer["wiki"], :target => "_blank", :rel => "noopener", :title => t("layouts.layer_info"), :class => "layer-info-toggle position-absolute z-3 text-white text-decoration-none lh-1" do %>
```
hlfan: To reduce the required custom CSS to just the `mix-blend-mode`:

```suggestion
          <%= link_to layer["wiki"], :target => "_blank", :rel => "noopener", :title => t("layouts.layer_info"), :class => "layer-info-toggle position-absolute z-3 p-1 end-0 text-white text-decoration-none lh-1" do %>
```

SinghSwayam: I have made the necessary changes.

PR #6674, opened 2026-01-04
Restrict linkify shorthand expansion to plain text

### LINKED ISSUE #6643: Linkify breaks Overpass QL source code blocks
### URL

~https://www.openstreetmap.org/user/bmaczero/diary/407962~

**Edit: User has tweaked the query in the meantime, so that the issue is no longer visible in the original blog post. It can still be seen here: https://master.apis.dev.openstreetmap.org/user/mmd2mod/diary/44**

Another diary comment with similar issue: https://www.openstreetmap.org/user/SK53/diary/401423#comment55322

A user has posted an Overpass QL query in  https://www.openstreetmap.org/user/bmaczero/diary/407962 as:

```
{{geocodeArea:Colorado}}->.searchArea;
way["leisure"="pitch"]["sport"="tennis"](area.searchArea);
foreach (way._(if:count_members()==5); out geom;);
```

Linkify rules have turned the query into this code, which is no longer valid:

```
{{geocodeArea:Colorado}}->.searchArea;
way["leisure"="pitch"]["sport"="tennis"](area.searchArea);
foreach https://wiki.openstreetmap.org/wiki/Tag:(way._(if:count_members()==5); out geom;);
```

Linkify rules should not be applied to source code blocks. They're supposed to be left as is.

_No response_

### LINKED ISSUE #6673: Diary comments expanded to include text I did not write
### URL

https://www.openstreetmap.org/user/chris_debian/diary/408003#comment60474

In a diary comment, write `disused:military=bunker`.

Notice that the comment is expanded to `https://wiki.openstreetmap.org/wiki/Tag:disused:military=bunker` which is (a) not what I wrote** and (b) not a clickable link.

** (if I'd wanted to link to the wiki I'd have done so; had I wanted to use a link I'd have been far more linkely to link to taginfo

n/a

### LINKED ISSUE #6692: Undocumented changing of message content
### URL

https://www.openstreetmap.org/messages/new/*

In the body of the message, I tried to format a tag using backticks (\`) for the standard markdown code formatting, this changes the message content to be a link to the wiki instead, even though that was not my intention with the message nor is it documented on the page that this will happen.

Image

Image

Here I would have expected the tags being put in the standard markdown code formatting to show the original content the same way \`test\` below does.

I'm not against this feature, it could be quite nice if someone is expecting it. I just think it either needs to be documented with the rest of the formatting options on the right, or it needs to be opt-in somehow.

hlfan:
Define linkify link detection/expansion rules only in the Text submodule and add a test against it messing up markdown content, which
* fixes #6643,
* fixes #6673, and
* fixes #6692.

### DIFF
--- a/test/lib/rich_text_test.rb
+++ b/test/lib/rich_text_test.rb
@@ -491,6 +491,15 @@ class RichTextTest < ActiveSupport::TestCase
     end
   end
 
+  def test_deactivated_linkify_expansion_in_markdown
+    t0 = "foo `surface=metal` bar"
+
+    r1 = RichText.new("markdown", t0)
+    t1 = Nokogiri::HTML.fragment(r1.to_html).text
+
+    assert_equal t0.delete("`"), t1.strip
+  end
+
   def test_text_to_html_email
     r = RichText.new("text", "foo example@example.com bar")
     assert_html r do


lib/rich_text.rb
```diff
@@ -88,25 +88,19 @@ def sanitize(text)
...
 
     def linkify(text, mode = :urls)
-      ERB::Util.html_escape(text)
-               .then { |html| expand_link_shorthands(html) }
-               .then { |html| expand_host_shorthands(html) }
-               .then { |html| auto_link(html, mode) }
-               .html_safe
+      link_attr = 'rel="nofollow noopener noreferrer" dir="auto"'
+      ERB::Util
+        .html_escape(text)
```
pablobm: Sane indentation! Yes, please, more of this 😄

lib/rich_text.rb
```diff
@@ -88,25 +88,19 @@ def sanitize(text)
...
+        .then { |html| expand_link_shorthands(html) }
+        .then { |html| expand_host_shorthands(html) }
+        .then { |html| Rinku.auto_link(html, mode, link_attr) { |url| shorten_link(shorten_hosts(url)) } }
```
pablobm: Why did you decide to move this back out of the method to here?

lib/rich_text.rb
```diff
@@ -119,41 +113,40 @@ def expand_host_shorthands(text)
         [Settings.linkify_hosts, Settings.linkify_hosts_replacement],
         [Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement]
       ]
-        .select { |hosts, replacement| replacement && hosts&.any? }
-        .reduce(text) do |text, (hosts, replacement)|
-          text.gsub(/(?<=^|#{URL_UNSAFE_CHARS})\b#{Regexp.escape(replacement)}/) do
-            "#{Settings.server_protocol}://#{hosts[0]}"
+        .map { |hosts, replacement| Struct.new(:hosts, :host_replacement).new(hosts, replacement) }
```
pablobm: I don't think that these structs buy us anything. They wrap data that then is only used in the same method, with the same names. If this data was passed on to other methods, then perhaps it would make sense, but I don't see it here.

Also, I did some cursory benchmarking and I'm seeing the struct versions running at half the speed of the non-struct ones. However this is probably not a big deal in context, and solvable by not creating the struct class (`Struct.new`) every time, I think.

lib/rich_text.rb
```diff
@@ -119,41 +113,40 @@ def expand_host_shorthands(text)
...
-            "#{scheme_host}#{path}"
-          end
-        end || url
-      end || url
+    def shorten_hosts(url)
+      [
+        [Settings.linkify_hosts, Settings.linkify_hosts_replacement],
+        [Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement, Settings.linkify_wiki_optional_path_prefix]
+      ]
+        .map { |hosts, replacement, prefix| Struct.new(:hosts, :host_replacement, :optional_path_prefix).new(hosts, replacement, prefix) }
```
pablobm: Same here about the structs.

lib/rich_text.rb
```diff
@@ -119,41 +113,40 @@ def expand_host_shorthands(text)
         [Settings.linkify_hosts, Settings.linkify_hosts_replacement],
         [Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement]
       ]
```
pablobm, line 113: Not really about this PR, but I don't like how we do this processing chain for a static array with only two elements. I think it would be more readable by moving the chain to a method and then doing something like this:

```ruby
[
  expand_shorthands(Settings.linkify_hosts, Settings.linkify_hosts_replacement),
  expand_shorthands(Settings.linkify_wiki_hosts, Settings.linkify_wiki_hosts_replacement)
]
```

lib/rich_text.rb
```diff
@@ -119,41 +113,40 @@ def expand_host_shorthands(text)
...
+            scheme_host, host, path = m.captures
+            if rule.hosts&.include?(host)
+              path = path.sub(Regexp.new(rule.optional_path_prefix || ""), "")
+              if rule.host_replacement
+                "#{rule.host_replacement}#{path}"
+              else
+                "#{scheme_host}#{path}"
+              end
+            end || url
+          end || url
```
pablobm: And similarly to `auto_link`: why remove `shorten_host` and move the code here? Chain methods with blocks that are in turn multi-level are difficult to follow.

pablobm (commented):
I wonder if the Markdown versions should still linkify, just not within code blocks. More difficult to implement though, so not for this PR.

lib/rich_text.rb
hlfan: You already mentioned re-enabling it for markdown later on.
In that case, rather than overwriting the gsub pair generator, I'd like to add flags to this linkify function to toggle the pairs of expanding & shortening. Then having both in the same function makes things easier.

Also, I didn't like having these pairs split in different places.

hlfan:
You know what? I'm splitting this PR. I originally just wanted to get some ideas down while in that headspace, but this is holding up the bug fix.

lib/rich_text.rb
hlfan: Lucky that this is already getting removed in #5862.

lib/rich_text.rb
hlfan: Another option is to not move the path replacement definition but to make it toggleable:

https://github.com/openstreetmap/openstreetmap-website/compare/master...hlfan:openstreetmap-website:structured-linkify

pablobm (approved):

tomhughes (approved):
The new cut down version of this looks good to me, thanks.

PR #6675, opened 2026-01-04
Migrate the edit location page from leaflet to Maplibre

CommanderStorm:
### Description

This PR migrates the edit location page from leaflet to maplibre.

This has two intentional changes:
- When clicking on the "Show" button, we now "fly" there instead of "fast-zoom-jump" (does not look very polished).
  I can change this to the previous behaviour, but that did not seem entirely intentional.
- the zoom buttons have rounded corners like on the main map

I disabled rotation, pich and yaw for this one, since for selecting a location you likely want this.

### How has this been tested?

Via docker, manually starting the map, interactig with it.
Locations get set appropriately, all buttons work.

### DIFF
--- a/app/assets/javascripts/user.js
+++ b/app/assets/javascripts/user.js
@@ -1,10 +1,8 @@
 //= require maplibre.map
-//= require maplibre.i18n
 //= require maplibre.combinedcontrolgroup
 //= require ./home_location_name-endpoint
 
 $(function () {
-  const defaultHomeZoom = 11;
   let map, marker, deleted_lat, deleted_lon, deleted_home_name, homeLocationNameGeocoder, savedLat, savedLon;
 
   if ($("#social_links").length) {
@@ -52,19 +50,7 @@ $(function () {
   }
 
   if ($("#map").length) {
-    map = new maplibregl.Map({
-      container: "map",
-      style: OSM.MapLibre.Styles.Mapnik,
-      attributionControl: false,
-      locale: OSM.MapLibre.Locale,
-      rollEnabled: false,
-      dragRotate: false,
-      pitchWithRotate: false,
-      bearingSnap: 180,
-      maxPitch: 0,
-      center: OSM.home ? [OSM.home.lon, OSM.home.lat] : [0, 0],
-      zoom: OSM.home ? defaultHomeZoom : 0
-    });
+    map = new maplibregl.Map(OSM.MapLibre.defaultSecondaryMapOptions);
 
     savedLat = $("#home_lat").val();
     savedLon = $("#home_lon").val();
@@ -132,7 +118,7 @@ $(function () {
       const lat = $("#home_lat").val(),
             lon = $("#home_lon").val();
 
-      map.flyTo({ center: [lon, lat], zoom: defaultHomeZoom });
+      map.flyTo({ center: [lon, lat], zoom: OSM.MapLibre.defaultHomeZoom });
     });
 
     $("#home_delete").click(function () {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/user.js
```diff
@@ -187,19 +195,19 @@ $(function () {
       ((deleted_lat && deleted_lon) || deleted_home_name)
     ));
     if (location) {
-      marker.setLatLng([lat, lon]);
+      marker.setLngLat([lon, lat]);
       marker.addTo(map);
-      map.panTo([lat, lon]);
+      map.panTo([lon, lat]);
```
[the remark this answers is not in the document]
CommanderStorm: We do!
https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#panto

but maybe `easeTo` would be better UX, not sure.

app/assets/javascripts/user.js
```diff
@@ -50,56 +52,62 @@ $(function () {
...
-      marker.setLatLng([OSM.home.lat, OSM.home.lon]);
+      marker.setLngLat([OSM.home.lon, OSM.home.lat]);
       marker.addTo(map);
     }
 
     map.on("click", function (e) {
       if (!$("#updatehome").is(":checked")) return;
 
-      const [lat, lon] = OSM.cropLocation(e.latlng, map.getZoom());
+      const [lat, lon] = OSM.cropLocation(e.lngLat, map.getZoom() + 1);
```
CommanderStorm, line 81: this might look like it needs more changing, but actually does not.

https://github.com/openstreetmap/openstreetmap-website/blob/65936c9e5b87a47e7a819a59e605f0a3e6bfffd5/app/assets/javascripts/osm.js.erb#L212-L216

app/assets/javascripts/user.js
```diff
@@ -50,56 +52,62 @@ $(function () {
...
-    }).addLayer(new L.OSM.Mapnik());
+      locale: OSM.MapLibre.Locale,
+      rollEnabled: false,
+      dragRotate: false,
+      pitchWithRotate: false,
+      bearingSnap: 180,
+      maxPitch: 0,
+      center: OSM.home ? [OSM.home.lon, OSM.home.lat] : [0, 0],
+      zoom: OSM.home ? defaultHomeZoom : 0
+    });
```
1ec5: Between #6504 and this and #6676, we should factor out common settings and controls for locator maps on the site, perhaps as a subclass.

CommanderStorm: If there are two item of something I don't think the added complexity of reasoning about that is worth it.
Reasoning how this `OSM`-global and the weird `//=` imports work is already a large overhead.

Lets not add more overhead by adding another function that needs resoning when it is intialised and what it does.

hlfan: An extra class for configuration seems overkill, there are simpler ways to keep it DRY:

```suggestion
    map = new maplibregl.Map(OSM.MapLibre.defaultSecondaryMapOptions);
```

CommanderStorm: if you think, you can reason better this way, shure.
I'd just like to avoid imports where possible after

- https://github.com/openstreetmap/openstreetmap-website/pull/6630/

app/assets/javascripts/user.js
```diff
@@ -187,19 +181,19 @@ $(function () {
...
   }
 
   function isCloseEnoughToMapCenter(location) {
-    const inputPt = map.latLngToContainerPoint(location),
-          centerPt = map.latLngToContainerPoint(map.getCenter());
+    const inputPt = map.project(location),
+          centerPt = map.project(map.getCenter());
 
-    return centerPt.distanceTo(inputPt) < 10;
+    return Math.sqrt(Math.pow(centerPt.x - inputPt.x, 2) + Math.pow(centerPt.y - inputPt.y, 2)) < 10;
```
hlfan: As of https://github.com/mapbox/point-geometry:
```suggestion
    centerPt.dist(inputPt) < 10;
```

tomhughes:
For crying out loud I thought I had turned off copilot access to this repository!

If anybody can navigate the github documentation better than me to find a way of getting rid of it please let me know...

PR #6676, opened 2026-01-04
Migrate the user diary to be maplibre based

CommanderStorm:
### Description

This Pull request migrates the user diary page to be maplibre based.

Same as https://github.com/openstreetmap/openstreetmap-website/pull/6675 this is map does not have rotation, yaw and pich.

It fixes the the following bugs:
- when i add the coordinates in the fields and then click "use map", this previously did not set the center correctly
- rounding of the zoom buttons

I did not do these changes, even though they seem sensible to me:
- adding attribution to the map (it was not there beforehand..)
- adding the "locate me"-GPS button (it was not there beforehand..)

### How has this been tested?

Manually, by all the 3 expected interraction modes:
- enter into fields, click "use map"
- click to set marker
- drag marker

### DIFF
--- a/app/assets/javascripts/diary_entry.js
+++ b/app/assets/javascripts/diary_entry.js
@@ -1,18 +1,26 @@
+//= require maplibre.map
+
 $(function () {
   let marker, map;
 
+  function updateFormFieldsFromMarkerPosition() {
+    const lngLat = marker.getLngLat();
+    $("#latitude").val(lngLat.lat);
+    $("#longitude").val(lngLat.lng);
+  }
+
   function setLocation(e) {
-    const latlng = e.latlng.wrap();
-
-    $("#latitude").val(latlng.lat);
-    $("#longitude").val(latlng.lng);
-
+    const coords = e.lngLat.wrap();
     if (marker) {
-      map.removeLayer(marker);
+      marker.setLngLat(coords);
+    } else {
+      marker = OSM.MapLibre.getMarker({ draggable: true })
+        .setLngLat(coords)
+        .setPopup(OSM.MapLibre.getPopup(OSM.i18n.t("diary_entries.edit.marker_text")))
+        .addTo(map);
+      marker.on("dragend", updateFormFieldsFromMarkerPosition);
     }
-
-    marker = L.marker(e.latlng, { icon: OSM.getMarker({}) }).addTo(map)
-      .bindPopup(OSM.i18n.t("diary_entries.edit.marker_text"));
+    updateFormFieldsFromMarkerPosition();
   }
 
   $("#usemap").click(function (e) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/diary_entry.js
```diff
@@ -22,23 +31,33 @@ $(function () {
...
+    map.keyboard.disableRotation();
+    map.on("click", setLocation);
 
+    // Create marker if coordinates exist when map is shown
     if ($("#latitude").val() && $("#longitude").val()) {
-      marker = L.marker(centre, { icon: OSM.getMarker({}) }).addTo(map)
-        .bindPopup(OSM.i18n.t("diary_entries.edit.marker_text"));
+      const lngLat = new maplibregl.LngLat($("#longitude").val(), $("#latitude").val());
+      setLocation({ lngLat });
+      map.setCenter(lngLat);
```
CommanderStorm, line 49: Previous behaviour, but I think this feels buggy though.
I would like this to remain as I am submitting in the PR and not as suggested below

```suggestion
```

app/assets/javascripts/diary_entry.js
```diff
@@ -22,23 +31,33 @@ $(function () {
...
+      pitchWithRotate: false,
+      bearingSnap: 180,
+      maxPitch: 0,
+      center: [params.lon, params.lat],
+      zoom: params.zoom - 1
+    });
+
+    const position = $("html").attr("dir") === "rtl" ? "top-left" : "top-right";
+    const navigationControl = new maplibregl.NavigationControl({ showCompass: false });
+    map.addControl(navigationControl, position);
```
CommanderStorm, line 41: if you want the GPS component on this page too

```suggestion
    const navigationControl = new maplibregl.NavigationControl({ showCompass: false });
    const geolocateControl = new maplibregl.GeolocateControl({
      positionOptions: {
        enableHighAccuracy: true
      },
      trackUserLocation: true
    });
    map.addControl(new OSM.MapLibre.CombinedControlGroup([navigationControl, geolocateControl]), position);
```

app/assets/javascripts/diary_entry.js
```diff
@@ -1,18 +1,27 @@
+//= require maplibre.map
+//= require maplibre.i18n
```
CommanderStorm: if you want the GPS component on this page too

```suggestion
//= require maplibre.i18n
//= require maplibre.combinedcontrolgroup
```

app/assets/javascripts/diary_entry.js
```diff
@@ -22,23 +31,33 @@ $(function () {
...
+      bearingSnap: 180,
+      maxPitch: 0,
+      center: [params.lon, params.lat],
+      zoom: params.zoom - 1
+    });
+
+    const position = $("html").attr("dir") === "rtl" ? "top-left" : "top-right";
+    const navigationControl = new maplibregl.NavigationControl({ showCompass: false });
+    map.addControl(navigationControl, position);
+    map.touchZoomRotate.disableRotation();
```
hlfan, line 42: Could we keep the three lines of event handling modifications to the end of the function?

CommanderStorm: I only see one `map.on("click", setLocation);`

PR #6682, opened 2026-01-06
Avoid language selector breaking on submit

pablobm:
Fixes https://github.com/openstreetmap/openstreetmap-website/issues/6598, closes https://github.com/openstreetmap/openstreetmap-website/pull/6599

Avoid submission of the language selection form as it is intended only as a filter for the options below. Submission not only doesn't make sense, it also breaks the form.

This is extracted from my comment at https://github.com/openstreetmap/openstreetmap-website/pull/6599#issuecomment-3636677922. I'm not adding tests because this is not trivial to test with Capybara. I have only tested manually, please others confirm that it fixes the problem for you.

### DIFF
--- a/app/assets/javascripts/language_selector.js
+++ b/app/assets/javascripts/language_selector.js
@@ -12,3 +12,7 @@ $(document).on("click", "#select_language_dialog [data-language-code]", function
     location.reload();
   }
 });
+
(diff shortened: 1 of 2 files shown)


hlfan (approved):
Thanks for this clean take on #6599.

PR #6683, opened 2026-01-07
Minimap migration from Leaflet to MapLibre

### LINKED ISSUE #6638: Migrate layer minimaps to MapLibre
The homepage’s Layers pane shows a series of live “minimaps” of the current viewport, one per featured layer. These minimaps are currently implemented as Leaflet maps distinct from the main slippy map. We don’t do much with the minimaps other than to display a particular layer noninteractively and synchronize it with the viewport. We should be able to migrate these maps over to MapLibre GL&nbsp;JS (#6234) before the main slippy map. This will help us catch device compatibility issues before they affect the more critical main map. It’ll also be the first time that we render layers other than Standard without the [maplibre-gl-leaflet](https://github.com/maplibre/maplibre-gl-leaflet/) compatibility shim.

https://github.com/openstreetmap/openstreetmap-website/blob/56b153f62b77543a24708e4958d6097e189d6eeb/app/assets/javascripts/leaflet.layers.js#L19-L20

Ironically, this will result in embedding a MapLibre map inside a Leaflet control, since we implement the Layers pane as an `L.OSM.SidebarPane`. I’d expect it to be fine from a performance and code management standpoint. The Shortbread and MapTiler OMT minimaps are already being rendered by MapLibre behind the scenes.

CommanderStorm:
### Description

This PR migrates the leaflet based minimaps to maplibre.

I did fix the bug, that moving the main vs the minimaps is not synced nicely (aka, just as on osm.org) but rather rubber-string-elasic.
This is now fixed in the xy-axis, but not in z axis, as leaflet does not have an onZoomIn/.. handler (`zoom` does not seem to work well for this).
Being snappy is a better experince likely and not much performance impact.

- Closes #6638

### How has this been tested?

Manually, everything works approximately the same.

When I run the development version, I tested with these styles:
- Standard
- CycleOSM
- HOT
- Shortbread
- OMT (with a freee api key from their website)

Others are hard to test without API keys.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -7,6 +7,8 @@
 //= require osm
 //= require leaflet/dist/leaflet-src
 //= require leaflet.osm
+//= require maplibre-gl/dist/maplibre-gl
+//= require leaflet.maplibre
 //= require leaflet.shortbread
 //= require i18n
 //= require leaflet.maptiler
--- a/app/assets/javascripts/leaflet.maptiler.js
+++ b/app/assets/javascripts/leaflet.maptiler.js
@@ -1,4 +1,3 @@
-//= require leaflet.maplibre
 //= require maplibre/i18n
 //= require @maptiler/maplibre-gl-omt-language
 
--- a/app/assets/javascripts/maplibre/map.js
+++ b/app/assets/javascripts/maplibre/map.js
@@ -1,4 +1,3 @@
-//= require maplibre-gl/dist/maplibre-gl
 //= require maplibre/controls
 //= require maplibre/dom_util
 //= require maplibre/styles
--- a/app/assets/javascripts/leaflet.maplibre.js
+++ b/app/assets/javascripts/leaflet.maplibre.js
@@ -1,4 +1,3 @@
-//= require maplibre-gl
 //= require @maplibre/maplibre-gl-leaflet
 
 maplibregl.setRTLTextPlugin(OSM.RTL_TEXT_PLUGIN, true);
--- a/app/assets/javascripts/leaflet.shortbread.js
+++ b/app/assets/javascripts/leaflet.shortbread.js
@@ -1,5 +1,3 @@
-//= require leaflet.maplibre
-
 L.OSM.Shortbread = L.OSM.MaplibreGL.extend({
   onAdd: function (map) {
     L.OSM.MaplibreGL.prototype.onAdd.call(this, map);


app/assets/javascripts/leaflet.layers.js
```diff
@@ -16,21 +20,27 @@ L.OSM.layers = function (options) {
...
-        const miniMap = L.map(container, { attributionControl: false, zoomControl: false, keyboard: false })
-          .addLayer(new layer.constructor(layer.options));
+        const styleId = layer.options.leafletOsmId;
+        const style = OSM.MapLibre.Styles[styleId](layer.options);
+        const miniMap = new maplibregl.Map({
+          container,
+          style,
+          interactive: false,
+          attributionControl: false,
+          fadeDuration: 0
```
CommanderStorm: In production, we would want this for better performance and maybe (depending how bundling works, I don't understand these imports) slightly smaller size.
How do I access if we are in prod or dev?

https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapOptions/#validatestyle

```suggestion
          fadeDuration: 0,
          validatestyle: false
```

CommanderStorm:
I think the copilot reviews that @tomhughes are a setting that I just disabled under the Copilot settings.
Will need to keep in mind to toggle this every time I create a PR for this effort, which is anoying but tolerable in the short term.

app/assets/javascripts/maplibre.i18n.js
```diff
@@ -1,7 +1,32 @@
+//= require i18n
```
CommanderStorm: I really don't understand how these imports actually work.
I seem to both require i18n, but then also not require it.

For this PR, I seem to need it, otherwise `OSM.i18n.t` does not seem to exist.

- https://github.com/openstreetmap/openstreetmap-website/pull/6630

Can someone with more rails knowlege give me a better mental model how this works?
- At first I was "C-Style-Preprocessor", that does not seem the case,
- then I was thinking this works in the "we concat things into a few giantic files, this specifes the order" but that also does not seem to be the case

So confusing

gravitystorm:
> I think how this works is that I now need to find the production `settigs.yml` yoink me the keys from there and test that.
> How is the recomend workflow there?

I generally don't bother testing the layers with apikeys, since they behave the same way as the other layers. This is definitely true for the raster layers, but I don't think there's much difference between the two vector layers either (in terms of how they integrate into the site that is).

CommanderStorm:
I figured out how to do it, you can just get a free account:
- https://github.com/openstreetmap/openstreetmap-website/pull/4042#issuecomment-3068587943

OMT works too, now just that weird import problem left

CommanderStorm:
seems that `i18n`, `maplibre.i18n` and `maplibre.setlanguage` need to be separate.
I don't get why, if someone has a poineter why rails behave so strangely.

(I purposely did not squash to make apparent what I tried and how weirdly this behaves, will do that once we have decided if I did something fundamentally broken or if rails is broken)

I think `maplibre.setlanguage` maybe could be integrated into i18n, but I am not sure how.
All ways that I tried seemed to require including `i18n` (which is logically required, but bricks the system if it is incuded), which does not work.
Maybe there is a magic way that I just don't know to make it prettier.

app/assets/javascripts/maplibre.i18n.js
tomhughes: It should be included as part of application.js on every page already and must not be imported a second time or it will clear all the translations.

CommanderStorm: For some reason this does not seem to work.
I think this is an import order problem where the application.js code gets imported after this, but not sure how to resolve it properly.

tomhughes: I can have a look when I get back next week.

CommanderStorm: thank you ❤️

app/assets/javascripts/application.js
```diff
@@ -8,11 +8,11 @@
 //= require leaflet/dist/leaflet-src
 //= require leaflet.osm
 //= require leaflet.shortbread
+//= require i18n
```
CommanderStorm, line 13: not a bug, nessesary to not crash the map... 😒

app/assets/javascripts/maplibre/styles.js
```diff
@@ -0,0 +1,36 @@
+const createRasterStyle = function (maxzoom, tileUrl) {
```
CommanderStorm: I tried a bit to load the yml file here, but have not figured out how to actually do this, since the leaflet code also did not figure this out, I think this code is fine.

It is a bit dense, bu not sure how to make it less dense without impacting on the dry-ness that was requested in previous reviews.

app/assets/javascripts/maplibre.i18n.js
CommanderStorm: I think I figured out where this import-issue was coming from. Not sure if my solution is the correct one though.

app/assets/javascripts/maplibre.styles.js
```diff
@@ -0,0 +1,36 @@
...
+      {
+        id: "raster",
+        type: "raster",
+        source: "raster"
+      }
+    ]
+  };
+};
+
+OSM.MapLibre.Styles.Mapnik = () => createRasterStyle(19, "https://tile.openstreetmap.org/{z}/{x}/{y}.png");
```
HolgerJeromin: I find the usage super surprising

```js
const style = OSM.MapLibre.Styles[styleId](layer.options);
```

Perhaps we should rename it
```js
const style = OSM.MapLibre.createStyles[styleId](layer.options);
```

Or make a helper
```js
const style = OSM.MapLibre.createStyles(styleId, layer.options);
```

CommanderStorm: None of this is dynamic code.
It should be static, assuming that I could figure out the yaml loading issue.

That is where I am currently a bit at stuck.

HolgerJeromin: My point is: 
Having a function call on a property named `.Styles[styleId]` was a bit confusing.

CommanderStorm: In normal software, yes I agree.

In this software, I think this breaks the existing convention that `OSM.PascalCase` is usually a constructor-ish thing. This is the equivalent to `L.OSM[stileId]`.

app/assets/javascripts/application.js
pablobm: True, but that doesn't clarify anything 🙂  How about an explanation:

- Before this PR, `OSM.i18n` was used later, when specific secondary maps (dashboard, diary) were loaded.
- With this PR, `leaftlet.maptiler` loads `maplibre.i18n`, which requires this library. Therefore we need to make sure it's available beforehand.

app/assets/javascripts/maplibre/styles.js
pablobm: Sorry, I'm very ignorant about this code 😓 What yml are you referring to here?

CommanderStorm:
> Had a quick look: the maps in the dashboard and the diary pages have stopped working

Fixed, was a stupid mistake

```diff
- style: OSM.MapLibre.Styles.Mapnik,
+ style: OSM.MapLibre.Styles.Mapnik(),

File: app/assets/javascripts/maplibre/styles.js (hunk 4 above)
Comment (CommanderStorm) [reply app/assets/javascripts/maplibre/styles.js]: We currently have these 3 definitions of which styles exist.
Would it not be nice to have 2 (and only one, once leaflet gets removed):

- [`layers.yml`](https://github.com/openstreetmap/openstreetmap-website/blob/master/config/layers.yml)
- [`leaflet.osm.js`](vendor/assets/leaflet/leaflet.osm.js)
- the place which I have added

Yes, they don't change all that often, but with 3 places, it is simple to miss one.
I think given that this is simpler once there are only 2 places, moving this to the future is likely simpler.

File: app/assets/javascripts/application.js (hunk 3 above)
Comment (tomhughes) [reply app/assets/javascripts/application.js:13]: Also as I said it will break other things when the second require happens and deletes all the translations - that's why I had to get rid of the duplicate require last time.

So we will need to find a solution but I will look at it later this week.

Comment (CommanderStorm) [reply app/assets/javascripts/application.js:13]: @tomhughes I just moved the import a few lines up. no digging nessesary

File: app/assets/javascripts/maplibre/styles.js (hunk 4 above)
Comment (pablobm) [reply app/assets/javascripts/maplibre/styles.js]: (Ooops, I left this comment here last week, but I didn't properly post the review so it didn't publish).

Thank you, one day I might understand this stuff 😅 I'll try to explain based on my Rails experience (in absence of relevant OSM experience):

Data from yml files typically will be loaded in `.js.erb` files as it needs to be read from filesystem into a template. I searched for `layers.yml`, and found it at [app/assets/javascripts/osm.js.erb](https://github.com/openstreetmap/openstreetmap-website/blob/deb5277ca5fef3321bd77652aa9bb03331bacea3/app/assets/javascripts/osm.js.erb#L34), loaded as `OSM.LAYER_DEFINITIONS`, which you can reference from this file. The specific definitions to load are selected by [lib/map_layers.rb](https://github.com/openstreetmap/openstreetmap-website/blob/deb5277ca5fef3321bd77652aa9bb03331bacea3/lib/map_layers.rb).

Still, the yml is missing the tile URLs, so either they should be added to the yml, or we add them manually like you do there. I would normally go for the first option, but I don't know the history here so there might be other things I'm missing.

File: app/assets/javascripts/maplibre/styles.js [hunk 6]
```diff
@@ -0,0 +1,36 @@
...
+OSM.MapLibre.Styles.TransportDarkMap = (options) => createRasterStyle(21, `https://{abc}.tile.thunderforest.com/transport-dark/{z}/{x}/{y}.png?apikey=${options.apikey}`);
+OSM.MapLibre.Styles.OPNVKarte = () => createRasterStyle(18, "https://tileserver.memomaps.de/tilegen/{z}/{x}/{y}.png");
+OSM.MapLibre.Styles.HOT = () => createRasterStyle(20, "https://tile-{abc}.openstreetmap.fr/hot/{z}/{x}/{y}.png");
+OSM.MapLibre.Styles.TracestrackTopo = (options) => createRasterStyle(19, `https://tile.tracestrack.com/topo__/{z}/{x}/{y}.webp?key=${options.apikey}`);
+OSM.MapLibre.Styles.GPS = () => createRasterStyle(21, "https://gps.tile.openstreetmap.org/lines/{z}/{x}/{y}.png");
+OSM.MapLibre.Styles.ShortbreadColorful = (options) => `https://vector.openstreetmap.org/styles/shortbread/${options.styleName}`;
+OSM.MapLibre.Styles.ShortbreadEclipse = (options) => `https://vector.openstreetmap.org/styles/shortbread/${options.styleName}`;
+OSM.MapLibre.Styles.OpenMapTiles = (options) => `https://api.maptiler.com/maps/openstreetmap/style.json?key=${options.apikey}`;
```
tomhughes: Isn't this creating these as raster layers rather than vector layers?

app/assets/javascripts/leaflet.layers.js
```diff
@@ -16,34 +19,50 @@ L.OSM.layers = function (options) {
...
+          validateStyle: OSM.RAILS_ENV !== "production"
+        });
 
-        miniMap.dragging.disable();
-        miniMap.touchZoom.disable();
-        miniMap.doubleClickZoom.disable();
-        miniMap.scrollWheelZoom.disable();
+        if (layer.options.leafletOsmId === "OpenMapTiles") {
+          OSM.MapLibre.setMapLanguage(miniMap);
+        }
```
tomhughes: So this is setting the language (which currently only affects the OpenMapTiles layer) but only if the main map is showing OpenMapTiles which seems wrong as the behaviour of the minimap for OpenMapTiles will depend on what main map is doing?

Plus it's changing global state anyway, and once at page load time (once the main map is initialised) so it depends on the initial layer.

As the language is global state I suspect it should just be set once when MapLibre is initialised and regardless of what layers are active?

app/assets/javascripts/leaflet.layers.js
```diff
@@ -16,34 +19,50 @@ L.OSM.layers = function (options) {
...
-
-        function setView(options) {
-          miniMap.setView(map.getCenter(), Math.max(map.getZoom() - 2, 0), options);
+        function animateZoomToMatchMainMap() {
+          miniMap.zoomTo(getMiniMapZoom(), { duration: 300 });
+        }
+        function setZoomToMatchMainMap() {
+          miniMap.setZoom(getMiniMapZoom());
+        }
+        function getMiniMapZoom() {
```
tomhughes: Can add a blank line before each of these functions please, for consistency with the other functions here if nothing else.

app/assets/javascripts/maplibre/styles.js
CommanderStorm: I don't get what you mean here.

These are the URLs under which the styles can be found.

app/assets/javascripts/application.js
tomhughes: Ah right that makes sense and is no problem.

app/assets/javascripts/maplibre/styles.js
tomhughes: Oh sorry I didn't realise those ones weren't actually calling the `createRasterStyle` function and were just returning a string!

app/assets/javascripts/leaflet.layers.js
CommanderStorm: > So this is setting the language (which currently only affects the OpenMapTiles layer) but only if the main map...

> it's changing global state anyway, and once at page load time (once the main map is initialised) so it depends on the initial layer

No, this is setting the OpenMapTiles minimap's language.
It is not changing global state or affected by the main map.

It is affected by the
- global state of what the prefered language is and
- the local state which languages the style supports.

Since there is no way to change the language without a reload cycle, no further action is required.
Am I missing something?

> As the language is global state I suspect it should just be set once when MapLibre is initialised and regardless of what layers are active?

Since currently only OMT supports this, I think the current approach is fine.
Raster layers won't support languages, but Shortbread might.
I can remove the if statement if you want.

I am unusure if the current approach is compatible

CC @pnorman 

Should I enable this for shortbread too, or is the following more of a far-future?

- https://github.com/shortbread-tiles/shortbread-docs/pull/106

app/assets/javascripts/maplibre.i18n.js
```diff
@@ -5,3 +5,26 @@ OSM.MapLibre.Locale = {
...
   "Popup.Close": OSM.i18n.t("javascripts.map.popup.close")
 };
+
+OSM.MapLibre.setMapLanguage = function (map) {
+  if (!map.style.loaded()) {
+    map.once("load", () => OSM.MapLibre.setMapLanguage(map));
+    return;
+  }
+
+  const supportedLanguages = map.style.getLanguage();
```
tomhughes: There is no `getLanguage` function on the style so this is erroring?

The old code used `maplibregl.Map.prototype.supportedLanguages` to get the list of supported languages.

app/assets/javascripts/leaflet.layers.js
tomhughes: Looks like I misunderstood this though I am interested in what he effect (if any) is of setting the language for a layer that doesn't support it - the list of supported languages at least appears to be fixed and not tied to the current style?

CommanderStorm: Yea, the map language not updating with the language selector is defintively a bug, but at least if I don't fix it here it is consistent 😉

app/assets/javascripts/maplibre.i18n.js
CommanderStorm: I think I did not fully understand what this code was doing beforehand.

This is maptilers code for setting the language in OMT -> https://www.npmjs.com/package/@maptiler/maplibre-gl-omt-language?activeTab=code

app/assets/javascripts/maplibre/styles.js
tomhughes: I think mostly the URLs aren't in `layers.yml` because with leaflet we were using the constructors from `leaflet-osm` which had the URLs embedded in it so we didn't need them directly in our code.

app/assets/javascripts/leaflet.layers.js
1ec5: > the list of supported languages at least appears to be fixed and not tied to the current style?

Correct, this is a limitation of mapbox-gl-language and the fork of it that MapTiler made for us. I’m working on [a new library](https://github.com/osm-americana/diplomat/) that’s more generic, among other things.

app/assets/javascripts/osm.js.erb
```diff
@@ -4,6 +4,7 @@
 //= depend_on legend.yml
 
 OSM = {
+  RAILS_ENV: "<%= Rails.env %>",
```
hlfan: Could the boolean conversion be done on the server side?
```ruby
OSM.validateStyle = Rails.env != "production"
```

app/assets/javascripts/maplibre.styles.js
```diff
@@ -0,0 +1,36 @@
+const createRasterStyle = function (maxzoom, tileUrl) {
+  const expandedUrls = tileUrl.includes("{abc}") ?
```
hlfan: Could this use the iD style definition to be more flexible about subdomains?

app/assets/javascripts/leaflet.maptiler.js
```diff
@@ -12,20 +13,7 @@ L.OSM.OpenMapTiles = L.OSM.MaplibreGL.extend({
...
-        .replace("-", "_");
-      const matchedLanguage = supportedLanguages.find(
-        (supported) => supported.toLowerCase() === normalizedPreferredLanguage
-      );
-      if (matchedLanguage) {
-        maplibreMap.setLanguage(matchedLanguage);
-        break;
-      }
-    }
+    OSM.MapLibre.setOMTMapLanguage(this.getMaplibreMap());
```
hlfan, line 15: Overriding `L.MaplibreGL._initGL()` is unreasonable now, but when Leaflet is dropped, we could make this a part of `OSM.MapLibre.Map`.

CommanderStorm:
Will rework this and rebase onto master over the week, am just coming back from a conference and have a bit of a backlog

app/assets/javascripts/maplibre.styles.js
CommanderStorm: Where does iD store their style defintions?
I am not sure if I found this.

I only found things like this, which seems fairly similar.

https://github.com/openstreetmap/iD/blob/585921e43fe5982d7d220514313fd64158b525ce/modules/services/mapilio.js#L90-L92

hlfan: I'd just go with the definition syntax from the editor-layer-index:

https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md

https://github.com/openstreetmap/iD/blob/585921e43fe5982d7d220514313fd64158b525ce/modules/renderer/background_source.js#L222-L225

app/assets/javascripts/osm.js.erb
CommanderStorm: I can change it to something like `OSM.PRODUCTION_ENVIRONMENT` (`OSM.validateStyle` is a bit too specific imo and does not match the captalisation for constants).

Readability of `OSM.RAILS_ENV === 'production'` is in my oppinon a bit higher, as this way it is clear tht the environment comes from rails.

app/assets/javascripts/leaflet.maptiler.js
CommanderStorm: We need to rework this system anyhow as @1ec5 noted.
I think pushing this into the future is a good call though, that would get a bit large maybe.

app/assets/javascripts/osm.js.erb
CommanderStorm: moved to another PR

app/assets/javascripts/leaflet.layers.js
```diff
@@ -16,34 +20,53 @@ L.OSM.layers = function (options) {
...
-          .addLayer(new layer.constructor(layer.options));
+        const styleId = layer.options.leafletOsmId;
+        const style = OSM.MapLibre.Styles[styleId](layer.options);
+        const miniMap = new OSM.MapLibre.Map({
+          container,
+          style,
+          interactive: false,
+          attributionControl: false,
+          fadeDuration: 0,
+          zoomSnap: styleId === "OpenMapTiles" || styleId === "ShortBread" ? 0.0 : 1.0
```
hlfan: This surely won't be the last time we need this distinction.
I think we can no longer postpone adding a flag for vector layers in the `layers.yml`.

app/assets/javascripts/maplibre/styles.js
```diff
@@ -1,20 +1,50 @@
...
+      raster: {
+        type: "raster",
+        tiles: convertIDUrlToMapLibreUrls(tileUrl),
+        tileSize: 256,
+        maxzoom: maxzoom
+      }
+    },
+    layers: [
+      {
+        id: "raster",
```
hlfan: Won't reusing ids cause problems later?

app/assets/javascripts/maplibre/styles.js
```diff
@@ -1,20 +1,50 @@
...
+    return [tileUrl];
+  }
+
+  const subdomains = match[1].split(",");
+  return subdomains.map(subdomain =>
+    tileUrl.replace(switchRegex, subdomain)
+  );
 };
+
+const createRasterStyle = function (maxzoom, tileUrl) {
```
hlfan: Could these functions not pollute the global scope?

CommanderStorm: I made sure to not include the styles twice and I am assigning to a variable to avoid the eslint lint for that.
If it is included twice, it breaks.

Not sure how to do this without since the rails import system seems very hard to work with. I tired using imports as suggested previously, but that did not entirely work out, so not sure how to solve this.

app/assets/javascripts/maplibre/styles.js
CommanderStorm: No, they are map specific IDs to identify the layer.

app/assets/javascripts/maplibre/styles.js
CommanderStorm: I can either have helper functions or include the result of running those helper functions.

Both are equally fine by me.

HolgerJeromin: https://developer.mozilla.org/en-US/docs/Glossary/IIFE
Is often used for this

app/assets/javascripts/maplibre/styles.js
hlfan, line 1: Actually, this doesn't need to be in JS at all.
What do you think of moving this to the MapLayers library?

CommanderStorm: On that aspect I don't think I am the right person to contribute.

I can do JS work, but on the ruby front I really don't have any experience.
Moving the way styles are constructed to unify on `layers.yml`, just not one where I can be of a lot of use I think.

config/layers.yml
```diff
@@ -50,6 +53,7 @@
   apiKeyId: "thunderforest_key"
   canEmbed: true
   canDownloadImage: true
+  isVectorStyle: false
```
hlfan: `canEmbed` and `canDownload` are only set if true to keep the line count low.

CommanderStorm: Very confusing to do it this way, but sure.

config/layers.yml
```diff
@@ -93,6 +106,9 @@
   nameId: "shortbread"
   wiki: "https://wiki.openstreetmap.org/wiki/Shortbread"
   maxZoom: 23
+  isVectorStyle: true
+  stileUrl: "https://vector.openstreetmap.org/styles/shortbread/colorful.json"
+  stileUrlDark: "https://vector.openstreetmap.org/styles/shortbread/eclipse.json"
```
hlfan: Do you mean style or something else?

CommanderStorm: yes, typo

app/assets/javascripts/maplibre/styles.js
hlfan: Then let's leave this JS-heavy; integration testing gets easier that way anyhow.

hlfan:
`@maptiler/maplibre-gl-omt-language` needs to overwrite the `maplibre.Map` prototype before `OSM.MapLibre` extends it for `setOMTMapLanguage` to work in the minimap.

hlfan:
I'd split off your last commit _"[Made sure that the minimap updates its center continuously, but zooms with on-animation](https://github.com/openstreetmap/openstreetmap-website/pull/6683/changes/3b1e037178474a0049fc916cbe706d169256eda1)" 3b1e037178474a0049fc916cbe706d169256eda1_ into another PR for later in the migration process because off-centre zooming makes the transition look much worse than when both transition smoothly together.

CommanderStorm:
> `@maptiler/maplibre-gl-omt-language` needs to overwrite the `maplibre.Map` prototype before `OSM.MapLibre` extends it for `setOMTMapLanguage` to work in the minimap.

I have a few bad words for this sprocket thing. This is very very hard to work with due to how the globals and prototypes and everything about this system works.

I have spend the last hour trying to trace where we are importing what why and how to fix this and I think I have an import order that fixes this .. for now.

app/assets/javascripts/maplibre/map.js
```diff
@@ -1,4 +1,5 @@
 //= require maplibre-gl/dist/maplibre-gl
+//= require @maptiler/maplibre-gl-omt-language
```
CommanderStorm: this is nessesary since this needs maplibre imported, but needs to be imported before styles are imported below.

For this file, it is not required but for the other files it is.

app/assets/javascripts/leaflet.maptiler.js
```diff
@@ -1,4 +1,5 @@
 //= require leaflet.maplibre
+//= require maplibre/i18n
 //= require @maptiler/maplibre-gl-omt-language
```
CommanderStorm, line 2: this is still required here since we need this for the main map intialisation sequence

app/assets/javascripts/application.js
```diff
@@ -7,12 +7,14 @@
 //= require osm
 //= require leaflet/dist/leaflet-src
 //= require leaflet.osm
+//= require maplibre-gl/dist/maplibre-gl
```
CommanderStorm, line 10: maplibre needs to be imported here to not be imported by the minimap, shortbread and OMT.

Currently this is done by both shortbread and OMT.

I cannot leave this in leaflet.maplibre since otherwise the dependency with maplibre/map stops working

app/assets/javascripts/leaflet.layers.js
```diff
@@ -1,3 +1,8 @@
+//= require @maptiler/maplibre-gl-omt-language
```
CommanderStorm, line 1: required for the OMT style

app/assets/javascripts/application.js
hlfan: > not be imported by the minimap, shortbread and OMT

Which Sprockets does not, iirc. But I'm not a Sprockets expert either.

CommanderStorm: If I add a duplicate require of some global `const foo = 1;`, this will cause an error.

This leads me to beleve that sprocket does not deduplicate.

Maybe @pablobm can chime in 😉

tomhughes: Sprockets is entirely text based inclusion - there is no deduplication or tree shaking or anything like that. It's up to you to make sure that each file is only included once.

hlfan (approved):
Thanks for your work on this; it looks good now!

PR #6685, opened 2026-01-07
Clarify how precompilation works in the docker setup guide

### LINKED ISSUE #6677: How to develop with decent velocity
### Problem

what is the recomended developement cycle for editing the JS parts of the project?

### Description

I tried to use the ruby things that are in the guide, but this does not seem to work well on my system (archlinux).

So the magic incarnation to get any changes to the JS to show up seem to be 

```bash
docker compose stop web
docker compose rm web
docker volume ls # to list the unnamed volume
docker volume rm ohm-website_web-storage ohm-website_web-tmp  0b494c14f2ce7d8b61bdcc02ce3de0b0840b9f311929f3f0a8d1da8b91d38142
docker compose up web --build
```

@1ec5 told me that he uses `docker-compose run --rm web bundle exec rake assets:precompile`, but this does not seem to work for me reliably.

The problem with this takes a bunch of time (~5 minutes) and is also not super intuitive to do.
How are others actually developing changes?
This seems like a pretty non-intitive workflow coming from the web world.

### Screenshots

_No response_

CommanderStorm:
Resolves #6677 

I did likely execute this every time I tested my code which caused this issue

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -24,13 +24,13 @@ Use [Docker Engine](https://docs.docker.com/engine/install/ubuntu/) with the [do
 
 The first step is to fork/clone the repo to your local machine:
 
-```
+```bash
 git clone https://github.com/openstreetmap/openstreetmap-website.git
 ```
 
 Now change working directory to the `openstreetmap-website`:
 
-```
+```bash
 cd openstreetmap-website
 ```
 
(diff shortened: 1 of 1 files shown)


gravitystorm:
Merged, thanks!

PR #6686, opened 2026-01-07
Enable erb_lint hard coded string linter

gravitystorm:
The implementation uses a customised linter, to add `&middot;` to the list of punctuation that doesn't need translating.

We can remove our customised linter when this list becomes a config option, see https://github.com/Shopify/erb_lint/pull/397

This follows on from #6569 and #6583 (see also #6627) and will prevent these types of problems from recurring.

### DIFF
--- a/.erb_linters/custom_hard_coded_string.rb
+++ b/.erb_linters/custom_hard_coded_string.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+require "better_html/tree/tag"
+require "active_support/core_ext/string/inflections"
+
+module ERBLint
+  module Linters
+    # A customised version of the HardCodedString linter, adding `&middot;` as not needing translation
+    # Checks for hardcoded strings. Useful if you want to ensure a string can be translated using i18n.
+    class CustomHardCodedString < HardCodedString
+      include LinterRegistry
+
+      NO_TRANSLATION_NEEDED = Set.new([
+                                        "&nbsp;",
+                                        "&amp;",
+                                        "&lt;",
+                                        "&gt;",
+                                        "&quot;",
+                                        "&copy;",
+                                        "&reg;",
+                                        "&trade;",
+                                        "&hellip;",
+                                        "&mdash;",
+                                        "&bull;",
+                                        "&ldquo;",
+                                        "&rdquo;",
+                                        "&lsquo;",
+                                        "&rsquo;",
+                                        "&larr;",
+                                        "&rarr;",
+                                        "&darr;",
+                                        "&uarr;",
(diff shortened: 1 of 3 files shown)


.erb_linters/custom_hard_coded_string.rb
```diff
@@ -0,0 +1,50 @@
...
+                                        "&rdquo;",
+                                        "&lsquo;",
+                                        "&rsquo;",
+                                        "&larr;",
+                                        "&rarr;",
+                                        "&darr;",
+                                        "&uarr;",
+                                        "&ensp;",
+                                        "&emsp;",
+                                        "&thinsp;",
```
hlfan, line 35: Don't these and `&nbsp;` get picked up by `/\s/`?

gravitystorm: I don't think so. The regexp is running against the erb source code, and so the `/\s/` removes all spaces from the source code. But if the source code contains the literal 6-character sequence `&nbsp;` then it will ignore that sequence.

Anyway, all this code is copy-pasted from https://github.com/Shopify/erb_lint/blob/52b6a4df2f83e4cff18738dea2c5b4bf2f59f086/lib/erb_lint/linters/hard_coded_string.rb . I changed the class to inherit from the original i.e. `class CustomHardCodedString < HardCodedString` and then removed everything I could without breaking it. Ideally our linter would just be something like:

```ruby
class CustomHardCodedString < HardCodedString
  NO_TRANSLATION_NEEDED << "&middot;"
end
```
...but I couldn't quite figure out how to get it that minimal! So we end up with the constant in full, plus our one-line addition, plus one method that couldn't be inherited cleanly and the `include` without which it breaks too.

hlfan: Oh right, the escape sequences wouldn't get picked up.

.erb_linters/custom_hard_coded_string.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require "better_html/tree/tag"
+require "active_support/core_ext/string/inflections"
+
+module ERBLint
+  module Linters
+    # A customised version of the HardCodedString linter, adding `&middot;` as not needing translation
```
hlfan, line 8: We're not just adding `&middot;` here, but more HTML escape sequences.
```suggestion
    # A customised version of the HardCodedString linter, adding HTML escape sequences as not needing translation
```

gravitystorm: We're only adding `&middot;` as the only customisation of this linter, the rest are from the upstream source code.

PR #6691, opened 2026-01-08
Fix keyboard navigation for richtext field tabs in New Diary Entry

### LINKED ISSUE #6561: Edit/Preview tab bar gets stuck when using arrow keys
### URL

https://www.openstreetmap.org/diary/new

1. Focus the big text field for the body of the diary entry.
2. Press <kbd>Shift</kbd><kbd>Tab</kbd> to focus the tab bar above the text field.
3. Press <kbd>→</kbd> or <kbd>↓</kbd> twice.

At this point, neither tab is selected, and the area below the tab bar is blank. It is no longer possible to select either tab using the keyboard.

Neither Edit tab nor Preview tab is selected.

Rustix69:
### Description

Fixes keyboard navigation issue on richtext field tabs (Edit/Preview) where pressing arrow keys twice would cause focus to escape the tab bar, leaving no tab selected and blank content displayed.

- Fixes : #6561 

**Problem:**
When users focused a tab button and pressed arrow keys (left/right/up/down), Bootstrap's built-in keyboard navigation would eventually cause focus to leave the tab list entirely. This made keyboard navigation impossible and left the UI in a broken state.

**Root Cause:**
1. Bootstrap's tab component has buggy arrow key handling that causes focus loss
2. Bootstrap automatically sets `tabindex="-1"` on inactive tabs, removing them from the Tab order

**Solution:**
1. Block arrow keys on richtext tab buttons using a capture-phase event listener that intercepts before Bootstrap handles the event
2. Override Bootstrap's tabindex management by ensuring all visible tabs have `tabindex="0"` so Tab/Shift+Tab navigation works correctly

**Changes:**
- `app/assets/javascripts/richtext.js`: Added arrow key blocking and tabindex override logic
- `app/views/shared/_richtext_field.html.erb`: Removed hardcoded tabindex attributes (now managed by JS)

**Navigation after fix:**
- Arrow keys: Blocked (do nothing)
- Tab: Edit -> Preview -> Body textarea
- Shift+Tab: Body textarea -> Preview -> Edit
- Click/Enter/Space: Activates the focused tab

### How has this been tested?

Tested on Docker environment at `/diary/new`:

1. Focused the body textarea
2. Pressed Shift+Tab - focus moved to Preview tab
3. Pressed Shift+Tab - focus moved to Edit tab
4. Pressed Tab - focus moved to Preview tab
5. Pressed Tab - focus moved to body textarea
6. Pressed arrow keys while on tab buttons - nothing happened (blocked as expected)
7. Clicked tabs and pressed Enter/Space - tabs activated correctly
8. Verified no focus loss or blank content issues

All keyboard navigation now works correctly without the original bug.


Rustix69:
**TCs Snapshot:** 
image

app/assets/javascripts/richtext.js
```diff
@@ -12,6 +12,42 @@
...
+   * Uses capture phase to intercept before Bootstrap handles the event.
+   */
+  document.addEventListener("keydown", function (e) {
+    if (!["ArrowLeft", "ArrowRight", "ArrowUp", "ArrowDown"].includes(e.key)) return;
+    if (!e.target.matches?.(".richtext_container button[data-bs-toggle='tab']")) return;
+
+    e.preventDefault();
+    e.stopPropagation();
+    e.stopImmediatePropagation();
+  }, true);
```
HolgerJeromin: Make this more clear:
```suggestion
  }, {capture: true});
```

HolgerJeromin:
Calling `ensureAllTabsFocusable` at three places is a bit cumbersome.
This (loosing focus) should be best fixed at the root level. But I do not know where this is.

tomhughes (changes requested):
The lint failures need to be resolved please.

pablobm:
There's one thing I don't understand: why does this bug exist in the first place? It looks like something we might have introduced ourselves, rather than a Bootstrap behaviour. I am concerned that adding more JS is only going to introduce complexity without fixing the issue at source.

Also I don't understand what the default behaviour in Bootstrap is. The examples at https://getbootstrap.com/docs/5.0/components/navs-tabs/#javascript-behavior use tab to navigate between tabs. However [a quick jsbin with the same code](https://jsbin.com/sisexiy/edit?html,output) has it working with the arrow keys instead... although at least it doesn't "overflow" like it happens in our bug.

Whichever the correct behaviour is, I would expect that the fix would be one of removing code, rather than adding more JS, and particularly JS that checks on specific key codes.

tomhughes:
Why ask for a new review when you haven't answered any questions people asked before?

Rustix69:
@tomhughes Sorry I didnt get to the question. Previously we were complicating this thing too much but after figuring out the exact issue it can be reduced and solved

1. The bug was mostly introduced by our own component setup, not by Bootstrap alone. Bootstrap’s default tab behavior is to support arrow-key navigation between tabs, and inactive tabs using `tabindex="-1"` is expected ARIA behavior.  
2. In our case, the richtext tablist includes a Help tab that is hidden on desktop (`d-sm-none`), while its panel content is moved into the sidebar. That leaves a tab control present in keyboard navigation but not actually visible/usable in that layout, which is what caused focus/selection to end up in a broken state.


I changed this to a minimal source fix:
- removed the keydown interception code
- removed the global tabindex override code
- kept Bootstrap’s default keyboard behavior
- only disabled the hidden Help tab when desktop sidebar mode is active, so keyboard navigation can’t land on an invisible tab target

This keeps behavior aligned with Bootstrap and reduces custom JS complexity.

app/views/shared/_richtext_field.html.erb
```diff
@@ -1,27 +1,31 @@
...
-      <div id="<%= id %>_edit" class="tab-pane show active">
+      <div id="<%= id %>_edit" class="tab-pane show active" role="tabpanel">
         <%= builder.text_area(attribute, options.merge(:wrapper => false, "data-preview-url" => preview_url(:type => type))) %>
       </div>
-      <div id="<%= id %>_preview" class="tab-pane">
-        <div class="richtext_placeholder text-center py-5" hidden><%= render "shared/loader" %></div>
+      <div id="<%= id %>_preview" class="tab-pane" role="tabpanel">
+        <div class="richtext_placeholder text-center py-5" hidden>
+          <div class="spinner-border" role="status">
+            <span class="visually-hidden"><%= t("browse.start_rjs.loading") %></span>
```
hlfan, line 23: Could you rebase your code in a way that keeps the shared loader partial?

PR #6693, opened 2026-01-09
Add demo links for routing engine directions Fixes #6387

### LINKED ISSUE #6387: Link directions to routing engine demo
[GraphHopper](https://graphhopper.com/maps/), [OSRM](https://map.project-osrm.org/), and [Valhalla](https://valhalla.openstreetmap.de/) all have their own first-party demo page that exposes more options and details than the [Leaflet Routing Machine](https://github.com/perliedman/leaflet-routing-machine/) plugin we’re using, such as via points (A to B to C), turn lanes, elevation profiles, time-sensitive routing, truck routing, and isochrones. Some of these features would be useful to mappers using the OSM website, but others are mainly useful to end users. The end user experience for directions is much more open-ended. We could address their needs by linking out to the selected routing engine’s demo page, adding any supported parameters that can get the page to show something equivalent to the current directions. Then it’s up to the routing engine projects whether they want to facilitate the end user experience to that degree. For example, the GraphHopper and OSRM demo pages both have a convenient button to [export GPX](https://community.openstreetmap.org/t/suggestion-add-download-route-as-gpx-to-osm-org/129580) (#5224), so there’s less need for us to reimplement that functionality ourselves.

SykoDaedalus:
Screenshot 2026-01-09 094808

issue #6387 

### Description
Changes:
- Extend the routes variable to include the selected routing mode and routing engine
- Generate routing engine demo URLs using URLSearchParams with engine-specific parameters
- Add a demo link to the directions view so routes can be opened in external routing engine demos
- The changes have been implemented as suggested in [this](https://github.com/openstreetmap/openstreetmap-website/issues/5224#issuecomment-2365144537) comment

### How has this been tested?
Tested locally using the Docker development environment.

- Verified directions page renders correctly for different routing modes and engines
- Confirmed the generated demo link opens the selected routing engine’s demo page with equivalent route parameters
- Ran Rails test suite with system tests skipped
- Ran RuboCop, ESLint, and ERB lint checks and fixed all reported issues

### DIFF
--- a/app/assets/javascripts/index/directions/fossgis_osrm.js
+++ b/app/assets/javascripts/index/directions/fossgis_osrm.js
@@ -34,6 +34,7 @@
         "depart": "start",
         "arrive": "destination"
       };
+
       function numToWord(num) {
         return ["first", "second", "third", "fourth", "fifth", "sixth", "seventh", "eighth", "ninth", "tenth"][num - 1];
       }
@@ -59,25 +60,33 @@
       if (step.maneuver.type.match(/^exit (rotary|roundabout)$/)) {
         return OSM.i18n.t(template, { name: name });
       }
+
       if (step.maneuver.type.match(/^(rotary|roundabout)$/)) {
         if (!step.maneuver.exit) {
           return OSM.i18n.t(template + "_without_exit", { name: name });
         }
+
         if (step.maneuver.exit > 10) {
           return OSM.i18n.t(template + "_with_exit", { exit: step.maneuver.exit, name: name });
         }
+
         return OSM.i18n.t(template + "_with_exit_ordinal", { exit: OSM.i18n.t(instrPrefix + "exit_counts." + numToWord(step.maneuver.exit)), name: name });
       }
+
       if (!step.maneuver.type.match(/^(on ramp|off ramp)$/)) {
         return OSM.i18n.t(template + "_without_exit", { name: name });
       }
+
       const params = {};
+
       if (step.exits && step.maneuver.type.match(/^(off ramp)$/)) params.exit = step.exits;
       if (step.destinations) params.directions = destinations;
       if (namedRoad) params.directions = name;
+
       if (Object.keys(params).length > 0) {
         template = template + "_with_" + Object.keys(params).join("_");
       }
+
       return OSM.i18n.t(template, params);
     }
 
@@ -86,6 +95,7 @@
         // special case handling
         if (mode === "ferry") return "ferry";
         if (intersections.some(i => i.classes?.includes("ferry"))) return "ferry";
+
         switch (maneuver.type) {
           case "on ramp":
           case "off ramp":
@@ -108,6 +118,7 @@
             return "turn " + maneuver.modifier;
         }
       }
+
       const ICON_MAP = {
         "continue": "straight",
         "merge right": "merge-right",
--- a/app/assets/javascripts/index/directions/graphhopper.js
+++ b/app/assets/javascripts/index/directions/graphhopper.js
@@ -27,6 +27,7 @@
         instr.distance,
         line.slice(instr.interval[0], instr.interval[1] + 1)
       ]);
+
       steps.at(-1)[0] = "destination";
 
       return {
@@ -71,7 +72,7 @@
           .then(({ paths }) => {
             if (!paths || paths.length === 0) throw new Error();
 
-            return { ... _processDirections(paths[0]), ...meta };
+            return { ..._processDirections(paths[0]), ...meta };
           });
       }
     };
--- a/app/assets/javascripts/index/directions-route-output.js
+++ b/app/assets/javascripts/index/directions-route-output.js
@@ -63,13 +63,16 @@ OSM.DirectionsRouteOutput = function (map) {
   function formatTime(s) {
     let m = Math.round(s / 60);
     const h = Math.floor(m / 60);
+
     m -= h * 60;
+
     return h + ":" + (m < 10 ? "0" : "") + m;
   }
 
   function writeSummary(route) {
     $("#directions_route_distance").val(formatTotalDistance(...translateDistanceUnits(route.distance)));
     $("#directions_route_time").val(formatTime(route.time));
+
     if (typeof route.ascend !== "undefined" && typeof route.descend !== "undefined") {
       $("#directions_route_ascend_descend").prop("hidden", false);
       $("#directions_route_ascend").val(formatHeight(...translateDistanceUnits(route.ascend)));
@@ -92,6 +95,7 @@ OSM.DirectionsRouteOutput = function (map) {
       } else {
         row.append("<td class='ps-3'>");
       }
+
       row.append(`<td><b>${i + 1}.</b> ${instruction}`);
       row.append("<td class='pe-3 distance text-body-secondary text-end'>" + formatStepDistance(...translateDistanceUnits(dist)));
 
@@ -131,6 +135,7 @@ OSM.DirectionsRouteOutput = function (map) {
     });
 
     const blob = new Blob([JSON.stringify(polyline.toGeoJSON())], { type: "application/geo+json" });
+
     URL.revokeObjectURL(downloadURL);
     downloadURL = URL.createObjectURL(blob);
     $("#directions_route_download").prop("href", downloadURL);
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -29,6 +29,7 @@ OSM.Directions = function (map) {
   ];
 
   const expires = new Date();
+
   expires.setFullYear(expires.getFullYear() + 10);
 
   const modeGroup = $(".routing_modes");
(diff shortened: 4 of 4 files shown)


hlfan (changes requested):
Could this be done similar to #5914?

`{ ...meta, ..._processDirections(foo) }` together with bundling the demo link and credit in metadata would make things less convoluted.

SykoDaedalus:
Sure, I will make the changes

app/views/directions/show.html.erb
```diff
@@ -71,4 +71,7 @@
   <p class="text-center">
     <%= t ".directions_courtesy_html", :link => tag.a("", :id => "directions_route_credit", :target => "_blank") %>
   </p>
+  <p class="text-center">
+  <%= tag.a "Check demo", :id => "directions_route_demo", :target => "_blank" %>
```
hlfan: That text would need to go through i18n.

app/assets/javascripts/index/directions/graphhopper.js
```diff
@@ -60,7 +69,7 @@
           .then(response => response.json())
           .then(({ paths }) => {
             if (!paths || paths.length === 0) throw new Error();
-            return _processDirections(paths[0]);
+            return _processDirections(paths[0], points, vehicleType);
```
hlfan: I'd use this opportunity to pull the credit out of the private function too:

```javascript
const meta = {
  credit: "GraphHopper",
  creditlink: "https://www.graphhopper.com/",
  demolink: demoLink
};
```

```suggestion
            return { ..._processDirections(paths[0]), ...meta };
```

That's what I originally was thinking.

SykoDaedalus:
Removed the creditlink and demolink from `_processDirection() `to `getRoute`

app/assets/javascripts/index/directions/graphhopper.js
```diff
@@ -55,12 +56,17 @@
           elevation: false,
           instructions: true
         });
+        const meta = {
+          credit: "GraphHopper",
+          creditlink: "https://www.graphhopper.com/",
+          demolink: `https://graphhopper.com/maps/?point=${points[0].lat}%2C${points[0].lng}&point=${points[1].lat}%2C${points[1].lng}&profile=${DEMO_PARAMS[modeId]}`
+        };
         points.forEach(p => query.append("point", p.lat + "," + p.lng));
```
hlfan: Using the `URLSearchParams` API would make this a bit more readable:

```suggestion
        const demoQuery = new URLSearchParams({ profile: vehicleType });
        for (const { lat, lng } of points) {
          query.append("point", [lat, lng]);
          demoQuery.append("point", [lat, lng]);
        }
        const meta = {
          credit: "GraphHopper",
          creditlink: "https://www.graphhopper.com/",
          demolink: "https://graphhopper.com/maps/?" + demoQuery
        };
```

SykoDaedalus:
Screenshot 2026-01-12 194139

app/assets/javascripts/index/directions/fossgis_valhalla.js
```diff
@@ -80,11 +80,20 @@
             }
           })
         });
+        const demoQuery = new URLSearchParams({ profile: DEMO_PARAMS[modeId] });
```
hlfan:
```suggestion
        const demoQuery = new URLSearchParams({
          profile: DEMO_PARAMS[modeId],
          wps: points.flatMap(({ lat, lng }) => [lng, lat])
        });
```

app/assets/javascripts/index/directions/fossgis_osrm.js
```diff
@@ -174,15 +175,23 @@
           // invalidate cache
           cachedHints = [];
         }
-
+        const demoQuery = new URLSearchParams({ srv: DEMO_PARAMS[modeId] });
```
hlfan: Could this be done with Array.indexOf?

hlfan (commented):
It might be worth looking into the Valhalla API if `car` could replace `auto` as the `vehicleType`, so we don't need an extra demo params object there.

app/assets/javascripts/index/directions/fossgis_valhalla.js
SykoDaedalus: I tried using both modeId and costing, but neither worked across all engines. modeId worked for Valhalla but not OSRM, which needs numeric IDs. costing worked for OSRM but broke Valhalla. Using auto in URL params didn't work either. Each engine expects different formats, so no single approach works for both.

I used the extra var to keep the style consistent across all engines.

app/assets/javascripts/index/directions/fossgis_osrm.js
SykoDaedalus: > Could this be done with Array.indexOf?

 I could use Array.indexOf(), but I think keeping the mapping makes it clearer what's happening and keeps the style consistent with other engines. It might help in future to understand the vehicle-to-parameter relationship at a glance?

Or maybe i use Array.indexof and add a comment why it has been done?

SykoDaedalus:
I also tried changing ` OSM.Directions.addEngine(new FOSSGISValhallaEngine("car", "auto"), true);` to  `OSM.Directions.addEngine(new FOSSGISValhallaEngine("car", "car"), true);`. Did not worked

app/assets/javascripts/index/directions/fossgis_valhalla.js
hlfan: Then I'd rather add this config to the `OSM.Directions.addEngine` lines than having a separate dict in different places.

hlfan (commented):
Could you squash this down to a single commit and remove some newline-only changes?

app/assets/javascripts/index/directions/fossgis_valhalla.js
SykoDaedalus: > Then I'd rather add this config to the `OSM.Directions.addEngine` lines than having a separate dict in different places.

something like `OSM.Directions.addEngine(new FOSSGISValhallaEngine("car", "auto","car"), true)`?

SykoDaedalus:
changed function header from `Engine(modeId,costing)`(different for each engine) to `Engine(modeId,costing,profile)`(needed for demo page url)

removed new-line only commits

SykoDaedalus:
> I also tried changing ` OSM.Directions.addEngine(new FOSSGISValhallaEngine("car", "auto"), true);` to `OSM.Directions.addEngine(new FOSSGISValhallaEngine("car", "car"), true);`. Did not worked

I used string for 0/1/2 in osrm because everything was in string. works also with int

app/assets/javascripts/index/directions/graphhopper.js
```diff
@@ -17,7 +17,6 @@
       "-8": "u-turn-left", // left u-turn
       "8": "u-turn-right" // right u-turn
     };
-
```
hlfan, line 20: imo these newlines quite improve the readability by breaking up the wall of text. What do you think?

app/assets/javascripts/index/directions/graphhopper.js
```diff
@@ -55,18 +51,28 @@
           elevation: false,
           instructions: true
         });
-        points.forEach(p => query.append("point", p.lat + "," + p.lng));
+        const demoQuery = new URLSearchParams({ profile: profile });
```
hlfan:
```suggestion
        const demoQuery = new URLSearchParams({ profile });
```

This `profile` variable could be used by both queries as query parameter and content are always the same here.

app/assets/javascripts/index/directions/graphhopper.js
hlfan: Like not just this one... Most of them.

SykoDaedalus: I think linebreaks improve readability but it is quite subjective. should I add linebreaks according to my preferences if you want?

SykoDaedalus: > Like not just this one... Most of them.

Added new lines between code blocks (if/else , function defs). Probably added too many😅

hlfan: I'll have to take another look at this on a bigger screen but this looks fine to me.

config/locales/en.yml
```diff
@@ -1612,6 +1612,7 @@ en:
       download: "Download route as GeoJSON"
       filename: "route"
       directions_courtesy_html: "Directions courtesy of %{link}"
+      check_demo_html: "Check demo on %{link}"
```
1ec5: “Check demo on GraphHopper” doesn’t really give the user an idea of why they might feel a need to visit the demo site. Maybe something like “More details from GraphHopper” would get the point across.

SykoDaedalus:
> Ok, the changes overall look good to me, but the commit message is a bit vague.
> 
> I think it would be clearer with
> 
> * all the feature changes in one commit (maybe titled alike the PR)
> * and the added newlines for readability in another.
> 
> Final nit while you're rebasing: the JS spread syntax is usually written without a trailing space, aka `"..._processDirections"` instead of `"... _processDirections"`.

separated features and newlines into 2 different commits. Fixed the trailing space issue

PR #6696, opened 2026-01-11
Update wiki_pages.yml

mmd-osm:
I ran the Wiki pages update script again, following a [request on the Wiki Talk page](https://wiki.openstreetmap.org/wiki/Talk:Wiki#Run_script_to_update_links_from_map).

The unit test fix was needed because someone added a Turkish translation for highway=primary.

### DIFF
--- a/test/helpers/browse_tags_helper_test.rb
+++ b/test/helpers/browse_tags_helper_test.rb
@@ -110,8 +110,8 @@ class BrowseTagsHelperTest < ActionView::TestCase
       link = wiki_link("key", "highway")
       assert_equal "https://wiki.openstreetmap.org/wiki/Tr:Key:highway?uselang=tr", link
 
-      link = wiki_link("tag", "highway=primary")
-      assert_equal "https://wiki.openstreetmap.org/wiki/Tag:highway=primary?uselang=tr", link
+      link = wiki_link("tag", "highway=path")
+      assert_equal "https://wiki.openstreetmap.org/wiki/Tag:highway=path?uselang=tr", link
     end
   end
 


test/helpers/browse_tags_helper_test.rb
```diff
@@ -111,7 +111,7 @@ def test_wiki_link
       assert_equal "https://wiki.openstreetmap.org/wiki/Tr:Key:highway?uselang=tr", link
 
       link = wiki_link("tag", "highway=primary")
-      assert_equal "https://wiki.openstreetmap.org/wiki/Tag:highway=primary?uselang=tr", link
+      assert_equal "https://wiki.openstreetmap.org/wiki/Tr:Tag:highway=primary?uselang=tr", link
```
hlfan: I think this should test the fallback for non-existing pages; the regular case is already covered by `de`.

mmd-osm: Ok I’ll change it to highway=path instead

pablobm:
@mmd-osm  - The wiki page has a "Last update" date. Is that still maintained? If so, mind updating it? (With year please!).

(Edit: pinged the wrong mmd initially).

mmd-osm:
Sorry, I’m a bit lost which last update date you mean here. I’ve updated this file a few times in the past and never updated any date that I’m aware of…

pablobm:
Sorry, I could have explained myself better. I mean at https://wiki.openstreetmap.org/wiki/Talk:Wiki#Run_script_to_update_links_from_map. There was a line reading as follows:

```
: Last updated of '''wiki_pages.yml''' was on '''March 19'''. Update is '''irregular'''.
```

Without even a year. I just changed it myself to 12 January 2026

mmd-osm:
Ah that’s a recent invention by a wiki user. That page is mainly used for discussion and contents will be archived after a while. I don’t think it’s a good place to keep such status information. In any case, the truth is in this repo, so I wouldn't worry too much about the wiki.

PR #6707, opened 2026-01-14
Update rubocop todos

gravitystorm:
This PR makes 3 changes to our rubocop todos

* A general `rubocop --auto-gen-config` update, since it's been a while since we last ran that, and some values have changed. Additionally, some of the comments have changed in newer versions of rubocop.
* Move Rails/SpecificActionNames todo list to be auto generated, now that we have fewer than 15 files remaining.
* Remove an exclusion from `rubocop.yml` that's no longer necessary.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -70,7 +70,6 @@ Rails/ReflectionClassName:
 Rails/SkipsModelValidations:
   Exclude:
     - 'db/migrate/*.rb'
-    - 'app/controllers/users/lists_controller.rb'
 
 Style/CollectionQuerying:
   Exclude:


tomhughes:
Well we want to tackle it in the sense of either fixing the issues, or agreeing that we don't want to enforce those rules and moving the disablement to the main rules file.

gravitystorm:
> Is the todo list something that we want to tackle? As in: getting some PRs in to reduce it. Has there been any effort towards this lately?

It's a "slowly chip away at it" kind of thing, and as Tom says there's a mixture of changing the code and changing the rubocop config - usually the code to be fair.

hlfan (approved):
Thanks!

PR #6708, opened 2026-01-14
Avoid suggesting support can create accounts

tomhughes:
The original plan was that administrator would manually create accounts for people that were blocked by an ACL but this is not actually what we do because we need people to read and agree the terms.

The error still suggests that we will create the account though, so let's fix that...


gravitystorm:
Merged, thanks

PR #6713, opened 2026-01-15
Add moderation zones where anonymous notes are not allowed

### LINKED ISSUE #6570: Enable PostGIS
(Moved from https://github.com/openstreetmap/operations/issues/1317).

As part of the conversation at https://github.com/openstreetmap/openstreetmap-website/issues/6512, I would like PostGIS to be enabled in the OSM website.

Currently the absence of PostGIS has led to some features being implemented with ugly workarounds, sometimes not functional at all. These are some examples that have been mentioned, although whether PostGIS is appropriate for all of them is still being debated:
- https://github.com/openstreetmap/openstreetmap-website/issues/2396
- https://github.com/openstreetmap/openstreetmap-website/issues/6567
- https://github.com/openstreetmap/openstreetmap-website/issues/6568
- List of nearby users.

To clarify: this is not about using PostGIS generally for everything in OSM. It's for use in small, ancillary features, which over time can be expanded.

pablobm:
Fixes https://github.com/openstreetmap/openstreetmap-website/issues/6570 (Enable PostGIS).

This PR introduces "moderation zones" where anonymous notes are not allowed. It also enables PostGIS in the process.

This addresses https://github.com/openstreetmap/operations/issues/1362 partially:
- Within the moderation zones, only unauthenticated users are blocked, not "recent" ones. We can measure the success of this one before moving on to recent accounts which are a bit more complex.
- No UI for moderators to define the zones. This will be implemented in a future PR.

### Interacting with PostGIS
In PostGIS terms, these zones are defined as geometries as opposed to geographies. This is consistent with the rest of the API.

Two gem options are available:
- [activerecord-postgis-adapter](https://github.com/rgeo/activerecord-postgis-adapter):
	- "Classic" gem, 15 years old.
	- Requires setting `adapter: postgis` which can be a headache when moving between branches.
	- Doesn't work well when using geography-enabled functions such as `ST_Covers`. Probably can be added by implementing the operations with Arel.
	- Does work well with `ST_Contains`.
	- Works with Ruby 3.2, probably lower but I haven't checked.
- [activerecord-postgis](https://github.com/seuros/activerecord-postgis) (current choice):
	- Very recent gem, 1 year old, I only discovered it by accident.
	- Doesn't require changing `config/database.yml`.
	- Works with geography-enabled functions such as `ST_Covers` by interpolating params into `?`, but not with Arel (can be added though).
	- Requires Ruby 3.3 minimum (although as mentioned Ruby 3.2 is about to enter EOL).
	- Claims not to be based on monkeypatching in comparison to the competition (I haven't looked into this).

In common to both:
- They are drop-in replacements for each other, at least as long as you stick to geometry functions.
- Both support Arel queries for geometry, which is the way I have implemented the query.
- They use [RGeo](https://github.com/rgeo/rgeo) behind the scenes to encapsulate GIS objects. This requires creating them with one of [several available factories](https://github.com/rgeo/rgeo/blob/main/doc/Which-factory-should-I-use.md). I have gone for `RGeo::Cartesian.simple_factory` as it appears to just work. If we changed to one of the others, we'd need the gem [ffi-geos](https://github.com/dark-panda/ffi-geos). This is not made clear in the documentation.
 
I haven't found a big argument for either, but I have gone for the newer one if only because of the simplicity of not having to update `config/database.yml`.

### Future work
- If an unauthenticated user writes a long, thoughtful note, submits it, but then it fails because it was a moderation zone... they have wasted their time, if not lost their work. Therefore we may want a way for clients to tell ahead of time if a specific location has a block. Probably an API such as `GET /api/0.6/notes/allowed?lat=X&lon=Y`.
- There's no UI to define the zones. We'll need a CRUD for moderators to use.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -123,6 +123,10 @@ jobs:
         cache: yarn
     - name: Install node modules
       run: bundle exec bin/yarn install
+    - name: Install packages
+      run: |
+        sudo apt-get -yqq update
+        sudo apt-get -yqq install postgresql-postgis
     - name: Setup database
       run: |
         sudo systemctl start postgresql
@@ -157,6 +161,10 @@ jobs:
         cache: yarn
     - name: Install node modules
       run: bundle exec bin/yarn install
+    - name: Install packages
+      run: |
+        sudo apt-get -yqq update
+        sudo apt-get -yqq install postgresql-postgis
     - name: Setup database
       run: |
         sudo systemctl start postgresql
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,8 @@ source "https://rubygems.org"
 gem "rails", "~> 8.1.0"
 gem "turbo-rails"
 
-# Use postgres as the database
+# Use postgres+postgis as the database
+gem "activerecord-postgis"
 gem "pg"
 
 # Use SCSS for stylesheets
--- a/app/controllers/api/notes_controller.rb
+++ b/app/controllers/api/notes_controller.rb
@@ -89,6 +89,8 @@ module Api
       lat = OSM.parse_float(params[:lat], OSM::APIBadUserInput, "lat was not a number")
       description = params[:text]
 
+      raise OSM::APIModerationZoneError if current_user.nil? && ModerationZone.falls_within_any?(:lon => lon, :lat => lat)
+
       # Get note's author info (for logged in users - user_id, for logged out users - IP address)
       note_author_info = author_info
 
--- a/app/models/moderation_zone.rb
+++ b/app/models/moderation_zone.rb
@@ -32,4 +32,13 @@ class ModerationZone < ApplicationRecord
   validates :name, :presence => true
   validates :reason, :presence => true
   validates :zone, :presence => true
+
+  def self.falls_within_any?(lon:, lat:)
+    factory = RGeo::Cartesian.simple_factory(:srid => 4326)
+    point = factory.point(lon, lat)
+
+    where(
+      arel_table[:zone].st_contains(point)
+    ).exists?
+  end
 end
--- a/app/models/moderation_zone.rb
+++ b/app/models/moderation_zone.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: moderation_zones
+#
+#  id            :bigint           not null, primary key
+#  name          :string           not null
+#  reason        :string           not null
+#  reason_format :enum             default("markdown")
+#  zone          :st_polygon       not null, polygon, 4326
+#  ends_at       :datetime
+#  creator_id    :bigint           not null
+#  revoker_id    :bigint
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+# Indexes
+#
+#  index_moderation_zones_on_creator_id  (creator_id)
+#  index_moderation_zones_on_revoker_id  (revoker_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (creator_id => users.id)
+#  fk_rails_...  (revoker_id => users.id)
+#
+class ModerationZone < ApplicationRecord
+  belongs_to :creator, :class_name => "User"
+  belongs_to :revoker, :class_name => "User", :optional => true
+
+  validates :name, :presence => true
+  validates :reason, :presence => true
+  validates :zone, :presence => true
+end
--- a/db/migrate/20260113142804_enable_postgis.rb
+++ b/db/migrate/20260113142804_enable_postgis.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class EnablePostgis < ActiveRecord::Migration[8.1]
+  def change
+    enable_extension "postgis"
+  end
+end
--- a/lib/osm.rb
+++ b/lib/osm.rb
@@ -19,8 +19,18 @@ module OSM
 
   # Raised when access is denied.
   class APIAccessDenied < APIError
-    def initialize
-      super("Access denied")
+    def initialize(message = "Access denied")
+      super
+    end
+
+    def status
+      :forbidden
+    end
+  end
+
+  class APIModerationZoneError < APIAccessDenied
+    def initialize(message = "You don't have permissions to make changes in this zone, as it is currently protected by moderators")
+      super
     end
 
     def status
--- a/test/controllers/api/notes_controller_test.rb
+++ b/test/controllers/api/notes_controller_test.rb
@@ -204,6 +204,18 @@ module Api
       assert_response :bad_request
     end
 
+    def test_create_anonymous_in_moderation_zone
+      point = coordinates_inside_seville_cathedral
+      create(:moderation_zone, :seville_cathedral)
+      assert_no_difference "Note.count" do
+        assert_no_difference "NoteComment.count" do
+          post api_notes_path(:lat => point.lat, :lon => point.lon, :text => "Down with this sort of thing")
+        end
+      end
+      assert_response :forbidden
+      assert_equal "You don't have permissions to make changes in this zone, as it is currently protected by moderators", response.headers["Error"]
+    end
+
     def test_create_success
       user = create(:user)
       auth_header = bearer_authorization_header user
@@ -230,6 +242,21 @@ module Api
       subscription = NoteSubscription.last
       assert_equal user, subscription.user
       assert_equal note, subscription.note
+    end
+
+    def test_create_success_in_moderation_zone
+      point = coordinates_inside_seville_cathedral
+      create(:moderation_zone, :seville_cathedral)
+      user = create(:user)
+      auth_header = bearer_authorization_header user
+      assert_difference "Note.count", 1 do
+        assert_difference "NoteComment.count", 1 do
+          assert_difference "NoteSubscription.count", 1 do
+            post api_notes_path(:lat => point.lat, :lon => point.lon, :text => "This is a comment", :format => "json"), :headers => auth_header
+          end
+        end
+      end
+      assert_response :success
     end
 
     def test_create_no_scope_fail
@@ -1253,6 +1280,10 @@ module Api
       notes.each_with_index do |note, index|
         assert_select "osm>note:nth-child(#{index + 1})>id", :text => note.id.to_s, :count => 1
       end
+    end
+
+    def coordinates_inside_seville_cathedral
+      Struct.new(:lat, :lon).new(37.385972, -5.993149)
     end
   end
 end
--- a/.devcontainer/Dockerfile.postgres
+++ b/.devcontainer/Dockerfile.postgres
@@ -1,0 +1,8 @@
+# Would use postgis:15-3.5, but it doesn't work with Apple Silicon
+# (https://github.com/postgis/docker-postgis/issues/216).
+# Therefore we use postgres:15 and apt-install Postgis ourselves.
+FROM postgres:15
+
+RUN apt-get update \
+  && apt-get install --no-install-recommends -y \
+    postgresql-15-postgis-3
--- a/.devcontainer/compose.yaml
+++ b/.devcontainer/compose.yaml
@@ -36,7 +36,9 @@ services:
     restart: unless-stopped
 
   postgres:
-    image: postgres:15
+    build:
+      context: .
+      dockerfile: ./Dockerfile.postgres
     restart: unless-stopped
     networks:
     - default
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,7 +36,7 @@ jobs:
     - name: Install packages
       run: |
         sudo apt-get -yqq update
-        sudo apt-get -yqq install memcached libvips-dev xvfb mesa-utils libgl1-mesa-dri
+        sudo apt-get -yqq install memcached libvips-dev xvfb mesa-utils libgl1-mesa-dri postgresql-postgis
     - name: Create database
       run: |
         sudo systemctl start postgresql
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -92,6 +92,10 @@ GEM
       timeout (>= 0.4.0)
     activerecord-import (2.2.0)
       activerecord (>= 4.2)
+    activerecord-postgis (0.5.1)
+      activerecord (>= 8.1.0, < 8.2)
+      pg
+      rgeo-activerecord (>= 8.0)
     activestorage (8.1.3)
       actionpack (= 8.1.3)
       activejob (= 8.1.3)
@@ -784,6 +788,10 @@ GEM
     reline (0.6.3)
       io-console (~> 0.5)
     rexml (3.4.4)
+    rgeo (3.1.0)
+    rgeo-activerecord (8.1.0)
+      activerecord (>= 8.1, < 8.2)
+      rgeo (>= 3.0)
     rinku (2.0.6)
     rotp (6.3.0)
     rouge (4.7.0)
@@ -922,6 +930,7 @@ DEPENDENCIES
   actionpack-page_caching (>= 1.2.0)
   active_record_union
   activerecord-import
+  activerecord-postgis
   addressable (~> 2.8)
   annotaterb
   argon2
--- a/db/migrate/20260113144310_create_moderation_zones.rb
+++ b/db/migrate/20260113144310_create_moderation_zones.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class CreateModerationZones < ActiveRecord::Migration[8.1]
+  def change
+    create_table :moderation_zones do |t|
+      t.string :name, :null => false
+      t.string :reason, :null => false
+      t.column :reason_format, :format_enum, :default => "markdown"
+      t.st_polygon :zone, :srid => 4326, :null => false
+      t.datetime :ends_at
+
+      t.references :creator, :null => false, :foreign_key => { :to_table => :users }
+      t.references :revoker, :foreign_key => { :to_table => :users }
+
+      t.timestamps
+    end
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1018,6 +1018,43 @@ ALTER SEQUENCE public.messages_id_seq OWNED BY public.messages.id;
 
 
 --
+-- Name: moderation_zones; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.moderation_zones (
+    id bigint NOT NULL,
+    name character varying NOT NULL,
+    reason character varying NOT NULL,
+    reason_format public.format_enum DEFAULT 'markdown'::public.format_enum,
+    zone public.geometry(Polygon,4326) NOT NULL,
+    ends_at timestamp(6) without time zone,
+    creator_id bigint NOT NULL,
+    revoker_id bigint,
+    created_at timestamp(6) without time zone NOT NULL,
+    updated_at timestamp(6) without time zone NOT NULL
+);
+
+
+--
+-- Name: moderation_zones_id_seq; Type: SEQUENCE; Schema: public; Owner: -
+--
+
+CREATE SEQUENCE public.moderation_zones_id_seq
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
+
+
+--
+-- Name: moderation_zones_id_seq; Type: SEQUENCE OWNED BY; Schema: public; Owner: -
+--
+
+ALTER SEQUENCE public.moderation_zones_id_seq OWNED BY public.moderation_zones.id;
+
+
+--
 -- Name: node_tags; Type: TABLE; Schema: public; Owner: -
 --
 
@@ -1867,6 +1904,13 @@ ALTER TABLE ONLY public.messages ALTER COLUMN id SET DEFAULT nextval('public.mes
 
 
 --
+-- Name: moderation_zones id; Type: DEFAULT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.moderation_zones ALTER COLUMN id SET DEFAULT nextval('public.moderation_zones_id_seq'::regclass);
+
+
+--
 -- Name: note_comments id; Type: DEFAULT; Schema: public; Owner: -
 --
 
@@ -2192,6 +2236,14 @@ ALTER TABLE ONLY public.languages
 
 ALTER TABLE ONLY public.messages
     ADD CONSTRAINT messages_pkey PRIMARY KEY (id);
+
+
+--
+-- Name: moderation_zones moderation_zones_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.moderation_zones
+    ADD CONSTRAINT moderation_zones_pkey PRIMARY KEY (id);
 
 
 --
@@ -2737,6 +2789,20 @@ CREATE INDEX index_issues_on_status ON public.issues USING btree (status);
 --
 
 CREATE INDEX index_issues_on_updated_by ON public.issues USING btree (updated_by);
+
+
+--
+-- Name: index_moderation_zones_on_creator_id; Type: INDEX; Schema: public; Owner: -
+--
+
+CREATE INDEX index_moderation_zones_on_creator_id ON public.moderation_zones USING btree (creator_id);
+
+
+--
+-- Name: index_moderation_zones_on_revoker_id; Type: INDEX; Schema: public; Owner: -
+--
+
+CREATE INDEX index_moderation_zones_on_revoker_id ON public.moderation_zones USING btree (revoker_id);
 
 
 --
@@ -3297,6 +3363,14 @@ ALTER TABLE ONLY public.social_links
 
 
 --
+-- Name: moderation_zones fk_rails_6a0b70e3da; Type: FK CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.moderation_zones
+    ADD CONSTRAINT fk_rails_6a0b70e3da FOREIGN KEY (creator_id) REFERENCES public.users(id);
+
+
+--
 -- Name: oauth_access_tokens fk_rails_732cb83ab7; Type: FK CONSTRAINT; Schema: public; Owner: -
 --
 
@@ -3366,6 +3440,14 @@ ALTER TABLE ONLY public.user_mutes
 
 ALTER TABLE ONLY public.oauth_access_tokens
     ADD CONSTRAINT fk_rails_ee63f25419 FOREIGN KEY (resource_owner_id) REFERENCES public.users(id) NOT VALID;
+
+
+--
+-- Name: moderation_zones fk_rails_f2132b7340; Type: FK CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.moderation_zones
+    ADD CONSTRAINT fk_rails_f2132b7340 FOREIGN KEY (revoker_id) REFERENCES public.users(id);
 
 
 --
@@ -3701,6 +3783,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('21'),
 ('20260223105922'),
 ('20260218183352'),
+('20260113144310'),
 ('20260113142804'),
 ('20251218105716'),
 ('20251121134648'),
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -31,6 +31,20 @@ COMMENT ON EXTENSION btree_gist IS 'support for indexing common datatypes in GiS
 
 
 --
+-- Name: postgis; Type: EXTENSION; Schema: -; Owner: -
+--
+
+CREATE EXTENSION IF NOT EXISTS postgis WITH SCHEMA public;
+
+
+--
+-- Name: EXTENSION postgis; Type: COMMENT; Schema: -; Owner: -
+--
+
+COMMENT ON EXTENSION postgis IS 'PostGIS geometry and geography spatial types and functions';
+
+
+--
 -- Name: format_enum; Type: TYPE; Schema: public; Owner: -
 --
 
@@ -3687,6 +3701,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('21'),
 ('20260223105922'),
 ('20260218183352'),
+('20260113142804'),
 ('20251218105716'),
 ('20251121134648'),
 ('20250704143751'),
--- a/doc/MANUAL_INSTALL.md
+++ b/doc/MANUAL_INSTALL.md
@@ -21,7 +21,7 @@ sudo apt-get update
 sudo apt-get install ruby ruby-dev ruby-bundler \
                      libvips-dev libxml2-dev libxslt1-dev \
                      nodejs build-essential git-core \
-                     postgresql postgresql-contrib libpq-dev \
+                     postgresql postgresql-contrib postgresql-postgis libpq-dev \
                      libsasl2-dev libffi-dev libgd-dev \
                      libarchive-dev libyaml-dev libbz2-dev npm
 sudo npm install --global yarn
@@ -40,7 +40,7 @@ sudo npm install --global yarn
 sudo dnf install ruby ruby-devel rubygem-rdoc rubygem-bundler \
                  rubygems libxml2-devel nodejs gcc gcc-c++ git \
                  postgresql postgresql-server \
-                 postgresql-contrib libpq-devel \
+                 postgresql-contrib postgis libpq-devel \
                  perl-podlators libffi-devel gd-devel \
                  libarchive-devel libyaml-devel bzip2-devel \
                  nodejs-yarn vips-devel
--- a/docker/postgres/Dockerfile
+++ b/docker/postgres/Dockerfile
@@ -1,4 +1,11 @@
+# Would use postgis:14-3.5, but it doesn't work with Apple Silicon
+# (https://github.com/postgis/docker-postgis/issues/216).
+# Therefore we use postgres:14 and apt-install Postgis ourselves.
 FROM postgres:14
+
+RUN apt-get update \
+  && apt-get install --no-install-recommends -y \
+    postgresql-14-postgis-3
 
 # Add db init script to install OSM-specific Postgres user.
 ADD docker/postgres/openstreetmap-postgres-init.sh /docker-entrypoint-initdb.d/
--- a/test/factories/moderation_zones.rb
+++ b/test/factories/moderation_zones.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :moderation_zone do
+    sequence(:name) { |n| "Moderation Zone #{n}" }
+    sequence(:reason) { |n| "Reason #{n}" }
+    creator { association :user }
+
+    trait :seville_cathedral do
+      zone do
+        <<~GEOMETRY
+          POLYGON((
+            -5.994110 37.386714,
+            -5.993954 37.385371,
+            -5.993621 37.385124,
+            -5.992162 37.385239,
+            -5.992264 37.386309,
+            -5.992506 37.386467,
+            -5.992559 37.386842,
+            -5.994110 37.386714
+          ))
+        GEOMETRY
+      end
+    end
+  end
+end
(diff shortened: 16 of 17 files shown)


mmd-osm:
[Geoblocking](https://en.wikipedia.org/wiki/Geo-blocking) usually refers to a different concept, i.e. a block based on the geographical location of a user, rather than an area they aren't supposed to edit in. I believe it would avoid confusion if this feature were given a different name.

pablobm:
> it would avoid confusion if this feature were given a different name.

Good call. Happy to hear suggestions :slightly_smiling_face:

app/controllers/api/notes_controller.rb
```diff
@@ -89,6 +89,8 @@ def create
       lat = OSM.parse_float(params[:lat], OSM::APIBadUserInput, "lat was not a number")
       description = params[:text]
 
+      raise OSM::APIAccessDenied if current_user.nil? && GeoblockZone.falls_within_any?(:lon => lon, :lat => lat)
```
gravitystorm: I suspect we'll need a separate error. Otherwise it's hard for data consumers to know what sort of error message to present.

I note here that in future we'll probably want to narrow the exemption list to moderators and admins rather than all signed in users.

Gemfile
```diff
@@ -6,7 +6,8 @@ source "https://rubygems.org"
 gem "rails", "~> 8.1.0"
 gem "turbo-rails"
 
-# Use postgres as the database
+# Use postgres+postgis as the database
+gem "activerecord-postgis-adapter"
```
gravitystorm: You asked about which gem to use, but I don't have a strong opinion on this, other people might have more experience with the alternative gem. It would be good to check if they are easily interchangeable - i.e. is the reason for their difference more to do with monkeypatching, or do they also have e.g. differences in how to write the queries?

app/models/geoblock_zone.rb
```diff
@@ -0,0 +1,30 @@
...
+#  updated_at :datetime         not null
+#
+class GeoblockZone < ApplicationRecord
+  def self.falls_within_any?(opts = {})
+    lon = opts.fetch(:lon)
+    lat = opts.fetch(:lat)
+
+    GeoblockZone.exists?(
+      [
+        # FIXME: do proper interpolation of lon/lat
```
gravitystorm: I think the fixme would be to write the queries using arel. If the st_covers isn't part of either gem, then now would be a good time to report that upstream and see if that can be fixed - maybe it will be available before we go further.

app/models/geoblock_zone.rb
```diff
@@ -0,0 +1,30 @@
...
+        <<~SQL.squish
+          ST_Covers(
```
gravitystorm: I don't have strong opinions on st_covers vs st_contains in our context - the difference between the two is not relevant here. The bigger difference is geometry vs geography support, see notes elsewhere.

db/migrate/20260113142804_enable_postgis.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class EnablePostgis < ActiveRecord::Migration[8.1]
+  def change
+    enable_extension "postgis"
```
gravitystorm, line 5: This will need the relevant "postgis" packages installed, both in CI but also in Docker, devcontainers and the Install notes. I don't think we need a full-blown extra github action for it though, especially because the one you mention does things to the db (e.g. enabling the extension) that we don't want to be done externally.

db/migrate/20260113144310_create_geoblock_zones.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateGeoblockZones < ActiveRecord::Migration[8.1]
+  def change
+    create_table :geoblock_zones do |t|
+      t.string :name
+      t.geometry :zone, :geographic => true
```
gravitystorm: We want to get geography vs geometry sorted out here, and this is something I'd like feedback from other people on.

I lean slightly towards geometry, for both theoretical and practical reasons. For example, let's say we draw a rectangular geoblock_zone, with four corners and a top corners at 70°N. For a geometry, the top edge is along the 70°N parallel. But for a geography, the edges of shapes are defined by great-circle arcs between the corners, and this might be unexpected behaviour. Halfway between the top corners, the arc could include 72°N, for example, or go over the pole if the corners are 180° longitude apart. Is this way moderators and mappers would expect?

We (generally?) treat OSM coordinates as a rectangular grid, and ways are (typically?) drawn as straight lines on a plane, rather than great-circle arcs.

So geography is a valid option, but I'm not sure if the extra complexity is worth it. One example of complexity is that many fewer PostGIS functions are implemented for geography compared to geometry.

db/migrate/20260113144310_create_geoblock_zones.rb
```diff
@@ -0,0 +1,12 @@
...
+
```
gravitystorm: I would consider mirroring the `user_blocks` table here, and add creator_id, and revoker_id here. We will want to have them in at the start, since we will want to "not null" those columns at some point.

We have "reason" in user_blocks, should we use name, reason, or both for geoblocks? We also have ends_at with user_blocks, which makes sure that all blocks automatically end, without moderators having to revisit them to end them manually.

test/controllers/api/notes_controller_test.rb
```diff
@@ -202,6 +202,14 @@ def test_create_anonymous_fail
         end
       end
       assert_response :bad_request
+
+      create_geoblock_zone
```
gravitystorm: Should be a different test, rather than adding to a megatest.

test/controllers/api/notes_controller_test.rb
```diff
@@ -1254,5 +1276,28 @@ def assert_notes_in_order(notes)
         assert_select "osm>note:nth-child(#{index + 1})>id", :text => note.id.to_s, :count => 1
       end
     end
+
+    def create_geoblock_zone
```
gravitystorm: Should be part of the factory, perhaps as a named trait

test/models/geoblock_zone_test.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class GeoblockZoneTest < ActiveSupport::TestCase
+  def test_falls_within_any
+    GeoblockZone.create(
```
gravitystorm: needs a factory

gravitystorm (changes requested):
Initial thoughts inline.

On a general overview, I'm happy with the scope of this initial PR, and then following up with a separate PR to sort out the "detection api" topic, then actually creating the zones via a UI.

db/migrate/20260113144310_create_geoblock_zones.rb
pnorman: We want geometry to have behavior consistent with the rest of the API

db/migrate/20260113142804_enable_postgis.rb
pablobm: Added now in individual workflows as appropriate.

db/migrate/20260113144310_create_geoblock_zones.rb
pablobm: Done now.

app/controllers/api/notes_controller.rb
pablobm: Done now.

Gemfile
pablobm: This is what I have so far:
1. `activerecord-postgis-adapter`: this seems to be the popular choice, but I'm having issues with it and barely can make it work. Also I find it annoying that I need to change the DB config to `adapter: postgis`, as then I have to change it back when I move branches to look at something else.
2. `activerecord-postgis`: it works and doesn't have the `adapter:` annoyance. However it requires Ruby 3.3 minimum.

For now I have gone with option 2. I need to research what's wrong with option 1 and whether I'm doing something wrong.

The differences appear to be with the implementation, but I haven't looked much into it yet. They seem to be interchangeable, with queries written pretty much the seem way and both relying on `rgeo-activerecord`.

app/models/geoblock_zone.rb
pablobm: This is fixed now, although may change depending on the final choice of gem, or preference of style.

app/models/geoblock_zone.rb
pablobm: I have gone back to `ST_Contains` following @pnorman's comment re: geography vs geometry.

db/migrate/20260113144310_create_geoblock_zones.rb
pablobm: Any preference of factory to use? https://github.com/rgeo/rgeo/blob/main/doc/Which-factory-should-I-use.md I am using `RGeo::Cartesian.simple_factory` for the moment as it appears to "just work". If we changed to one of the others, we'd need the gem `ffi-geos`.

test/controllers/api/notes_controller_test.rb
pablobm: Moved out of the megatest now.

test/controllers/api/notes_controller_test.rb
pablobm: Moved to a factory: `create(:geoblock_zone, :seville_cathedral)`.

test/models/geoblock_zone_test.rb
pablobm: Done.

test/controllers/api/notes_controller_test.rb
pablobm: There's also `build(:coordinates, :inside_seville_cathedral)`.

pablobm:
Any ideas on naming? If the current `geoblock_zones` is not good, options could be: `zone_blocks`, `protected_zones`. Also `areas` instead of `zones`. Thoughts?

I'm leaning towards `protected_zones` as "block" could be understood as a noun and cause confusion due to its other meanings.

Gemfile
tomhughes: Ruby 3.2 is EOL at the end of March in any case so we probably need to be thinking about moving to 3.2 as a baseline.

gravitystorm:
> Also `areas` instead of `zones`

Probably not `areas`, since they  [have a meaning already in OSM](https://wiki.openstreetmap.org/wiki/Area). `protected_area` also [has a specific meaning](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dprotected_area).

mmd-osm:
On the proposed API 0.7 page, someone called them "lock regions". I think that's not too bad. The other proposal "Quarantine" sounds a bit too strong for what we want to achieve here.

https://wiki.openstreetmap.org/wiki/API_v0.7#New_Features

db/migrate/20260113144310_create_geoblock_zones.rb
pablobm: Forgot to mention: I have changed this to be geometry.

pablobm:
> "lock regions". I think that's not too bad

"Locked regions" (if I may change the grammar slightly) could work. Having said that: even if the feature proposed in the wiki may never exist, I'm conscious that we might be overloading the term. I'm thinking whether the concept could be used for both this and what's described in the wiki and I think yes it could, but I could be missing something.

Gemfile
pablobm: Trying to debug this. If I use `activerecord-postgis-adapter`, with this very same code, I get this warning and error:

```
$ bin/rails test test/models/geoblock_zone_test.rb
Running 1 tests in a single process (parallelization threshold is 50)
Run options: --seed 64084

# Running:

unknown OID 147959: failed to recognize type of 'zone'. It will be treated as String.
E

Error:
GeoblockZoneTest#test_falls_within_any:
TypeError: Cannot visit RGeo::Cartesian::PointImpl
    app/models/geoblock_zone.rb:38:in 'GeoblockZone.falls_within_any?'
    test/models/geoblock_zone_test.rb:10:in 'GeoblockZoneTest#test_falls_within_any'
```

However if I don't have this problem in a new Rails app, with same gem versions, model, migration, and test.

So there's something in our gems or config that is interfering with `activerecord-postgis-adapter` but has no issues with `activerecord-postgis`.

pablobm: Trying to bisect this a bit. It's tricky. For now I know that revision aad81eb7 shows the same behaviour. I can't compare it with the alternative gem because it requires Ruby 3.3.0, which I can't get working there.

simonpoole:
May I suggest that doing this with postgis would only make sense if you are considering arbitrary boundaries. If you are just considering blocking at a country/state level there are a number of super fast geocoders (including one used by iD) that do this without all the complexity.

I would note that I'm very much for moving to postgis for other functionality, but for blocking notes it seems a tiny bit of overkill I might say.

1ec5:
I think the discussions that led us here were focused on arbitrary regions, or at least proximity to a certain location. Blocking an administrative boundary per se might be of less interest for notes specifically, but I could see that being potentially relevant for other kinds of blocking.

> super fast geocoders (including one used by iD)

To elaborate, iD uses [country-coder](https://github.com/rapideditor/country-coder/), which stores simplified country boundaries in a [which-polygon](https://github.com/mapbox/which-polygon/) spatial index, which in turn is based on [RBush](https://github.com/mourner/rbush/), which is an R-tree implementation, like PostGIS.

Gemfile
pablobm: The last issue was probably me. I think I had forgotten at one point to set `adapter: postgis`. Summary:

- `activerecord-postgis-adapter`: "classic" gem, 15 years old. Requires setting `adapter: postgis` which as I just illustrated can be a headache. Doesn't work well when using geography-enabled functions such as `ST_Covers`. Does work well with `ST_Contains`. Works with Ruby 3.2, probably lower but I haven't checked.
- `activerecord-postgis`: very recent gem, 1 year old, only discovered it by accident. Doesn't require changing `config/database.yml`. Works with geography-enabled functions such as `ST_Covers` interpolating `?`, but not with Arel (can be added though). Requires Ruby 3.3 minimum (although as mentioned Ruby 3.2 is about to enter EOL). Claims not to be based on monkeypatching in comparison to the competition (I haven't looked into this).

My preference is for the newer gem, if only because of the `database.yml` thing that does bother me, but I appreciate is only a short term thing. I'm sensing that dropping Ruby 3.2 would be on the cards for maintainers?

simonpoole:
> To elaborate, iD uses [country-coder](https://github.com/rapideditor/country-coder/), which stores simplified country boundaries in a [which-polygon](https://github.com/mapbox/which-polygon/) spatial index, which in turn is based on [RBush](https://github.com/mourner/rbush/), which is an R-tree implementation, like PostGIS.

I wasn't implying that the iD implementation was the best in class (it's the worst), just that it isn't a novel idea even for OSM central infrastructure.

pablobm:
> only make sense if you are considering arbitrary boundaries

It's for arbitrary boundaries indeed. For the sake of simplicity and separation of concerns, this PR doesn't include a tool to edit the boundaries, but the plan is to add one at a later stage.

Additionally, adding PostGIS support now would break a psychological barrier (if you will) that will encourage the creation of features that benefit from it.

pablobm:
Another potential name: moderation zones.

pablobm:
There's at least the question of naming open, but I think this advanced enough now to be out of draft status.

.github/workflows/lint.yml
```diff
@@ -103,6 +103,10 @@ jobs:
         cache: yarn
     - name: Install node modules
       run: bundle exec bin/yarn install
+    - name: Install packages
+      run: |
+        sudo apt-get -yqq update
+        sudo apt-get -yqq install postgis
```
pablobm: This may seem overkill for a lint workflow, but it is necessary. Otherwise we get this:
```
 psql:/home/runner/work/openstreetmap-website/openstreetmap-website/db/structure.sql:37: ERROR:  could not open extension control file "/usr/share/postgresql/14/extension/postgis.control": No such file or directory
bin/rails aborted!
failed to execute:
psql --set ON_ERROR_STOP=1 --quiet --no-psqlrc --output /dev/null --file /home/runner/work/openstreetmap-website/openstreetmap-website/db/structure.sql openstreetmap
```

pablobm (commented):
I have renamed the model to `ModerationZone` and tweaked a couple of things.

db/migrate/20260113142804_enable_postgis.rb
tomhughes: A bigger problem is that enabling postgis requires a postgres superuser which means that, for example, this would fail in our production environment. Indeed it did just fail to deploy this branch to the dev server and I had to manually hack it.

.github/workflows/lint.yml
gravitystorm: That's fair enough, it's expected that annotate_models needs access to the database to check the model file comments are up to date.

lib/osm.rb
```diff
@@ -19,8 +19,18 @@ def status
...
+      super
+    end
+
+    def status
+      :forbidden
+    end
+  end
+
+  class APIModerationZoneError < APIAccessDenied
+    def initialize(message = "Attempted to edit a Moderation Zone without enough trust")
```
gravitystorm: I think the message needs rephrasing:

* The user did not "attempt to edit a moderation zone", it will be moderators who edit the zones
* I know what you mean by "without enough trust" but perhaps it's a little cryptic?

test/factories/coordinates.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :coordinates, :class => Struct.new(:lat, :lon) do
+    trait :inside_seville_cathedral do
```
gravitystorm: It's a little unusual to have a factory that doesn't represent a model. Would it be more clear if it was just a helper method within the relevant test file?

test/factories/geoblock_zones.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :moderation_zone do
+    sequence(:name) { |n| "Zone block #{n}" }
+    sequence(:reason) { |n| "Reason for block #{n}" }
```
gravitystorm: "reason for moderation zone"?

app/models/moderation_zone.rb
```diff
@@ -0,0 +1,40 @@
...
+
+# == Schema Information
+#
+# Table name: moderation_zones
+#
+#  id            :bigint           not null, primary key
+#  name          :string           not null
+#  reason        :string           not null
+#  reason_format :enum             default("markdown")
+#  zone          :st_geometry      not null, geometry, 0
```
gravitystorm: The 0 at the end makes me think about projections. I'm not sure here whether an srid of 0 or 4326 would be better here.

gravitystorm (changes requested):
This looks good to me, just a few minor comments inline.

The big topic here for me now is the ruby 3.3+ requirement. This means that developers can't install the codebase on the latest Ubuntu LTS, which we've traditionally supported for DX reasons. While experienced devs might be fine with ruby version managers, for many people it's a confusing nightmare of unclear choices, particularly if they aren't familiar with ruby. (I even had to change my local ruby version to test this PR!)

I can think of a few options:
* review the upstream gem to see if ruby 3.2 support is trivial? Long shot, but might be an easy option. I'd even consider using a fork for a few months if it's a minor delta.
* Move the ruby 3.3+ stuff into a separate PR, so that we can discuss/merge that bit first
* Wait for a few months until the next LTS is out and some of this stuff resolves itself!

Aside from that, I'm keen to hear if anyone else has review comments.

app/models/moderation_zone.rb
pablobm: I don't think I have much choice here, as it's automatically generated. However: doesn't 4326 imply geography? I'm totally ignorant on this topic, but that was the impression I got after switching between the two options a few times.

test/factories/geoblock_zones.rb
pablobm: Going for "Reason #{n}".

test/factories/coordinates.rb
pablobm: Since it's used only in one test, I'm ok with either option. I do use FactoryBot a bit liberally and this showed here :slightly_smiling_face: Moved to a helper now.

lib/osm.rb
pablobm: Going with "You don't have permissions to make changes in this zone, as it is currently protected by moderators".

pablobm (commented):
Addressed the simpler pieces of feedback. I'm going to look into the Ruby 3.2 question now and will report back.

Incidentally, I can't edit the title of the PR any more and I have now seen the typo! My eyes! :sob: Judging from a different comment elsewhere, this seems to be a change in GitHub.

app/models/moderation_zone.rb
pnorman: You want it in a a SRID of 4326. Geometry or geography is determined by the type, not the SRID.

gravitystorm: > I don't think I have much choice here, as it's automatically generated.

I've just checked, and the migration syntax supports adding `srid: 4326`

> However: doesn't 4326 imply geography?

No, the geometry vs geography stuff ("straight lines vs great circles" or Cartesian vs Spherical coords) is separate from the projection stuff ("what exactly does 10.7, 5.0 mean"). But the question is only whether we want a srid of 0 ("unknown") or 4326 (lat/lon roughly speaking). Using 0 is often OK but can cause issues if we do more complex things in future, so perhaps we should stick with 4326.

One final note - do we want to constrain the types of objects in the table? At the moment we have "geometry", which supports all the postgis types (e.g. point, linestring, polygon and their multi- variants). Should we make this instead a POLYGON (or MULTIPOLYGON?) column? @pnorman any thoughts?

pablobm:
I had a look. The author of the activerecord-postgis explains their decision at https://github.com/seuros/activerecord-postgis/issues/13. I quote:

> Ruby 3.2 has some weird bug in Async and fiber.
> (...)
> [If you want to use Ruby 3.2,] you fork, change the requirement, and use your fork until you upgrade. I cannot guarantee you compatibility in a version i don't even have installed in my machines.

I interpret that (plus a link they provide) specifically as "a bug when doing concurrency with fibers". I have asked to clarify in any case.

I think that this problem doesn't affect us. Unless I'm mistaken our app doesn't use fibers, neither does Rails/ActiveRecord (unless specifically configured), and I don't think Phusion Passenger does either. Caveat: I have never used fibers myself, I might be missing something.

If I'm correct, then we can create a fork that allows for Ruby 3.2 for a while, at least until the next Ubuntu LTS which should be published in April.

If not... well, I'd rather not having to park this for another 2-3 months! :sweat_smile: But the alternative wouldn't really be great: if we used activerecord-postgis-adapter, everyone would have to set `adapter: postgis`, and then change it back to `adapter: postgres` in the future if we decide to actually change gems for any reason. It doesn't sound like much, but I can see it becoming a DX problem for many inexperienced contributors.

app/models/moderation_zone.rb
pnorman: Embedded in that question is if geoblock zones can be disjoint areas. If they can't then it should be `geometry(Polygon,4326)`. I'd have to double-check what you need if disjoint areas are allowed.

pablobm: Thank you for the explanation, team. I have changed it to 4326.

Good point on the type question, I was too liberal there. Again from ignorance: my feeling is that changing to polygon would make sense. Allowing users to define multipolygons would be adding complexity (eg: to the upcoming shapes editor). And if multipolygons are even needed... I'm guessing it can be changed from polygon to multipolygon then?

pablobm: @pnorman - My thought is that there's no reason for them to be disjointed areas. If moderators want disjointed areas, they can create several zones.

pnorman: Forcing polygons also improves indexing. If you represent the US as a multipolygon most of Europe is within its bounding box. If you split it into 2-4 the bounding boxes become sensible.

pablobm:
> I interpret that (plus a link they provide) specifically as "a bug when doing concurrency with fibers". I have asked to clarify in any case.

My understanding has been confirmed at https://github.com/seuros/activerecord-postgis/issues/13#issuecomment-3854312930

.rubocop.yml
```diff
@@ -12,7 +12,7 @@ require:
   - ./.rubocop/specific_action_names.rb
 
 AllCops:
-  TargetRubyVersion: 3.2
+  TargetRubyVersion: 3.3
```
seuros, line 15: If you can use 3.4 , that more future prone.   The Cops did not change between 3.3 and 4.0

pablobm:
More info from the creator of the alternative gem: https://github.com/seuros/activerecord-postgis/issues/13#issuecomment-3855009841 So the gem is intended as a testing ground for new APIs in Rails that allow for cleaner extension. It might even be transferred to rgeo in the future.

pablobm:
Changed the type to `st_polygon` as suggested, since it does make a lot of sense.

@tomhughes: in terms of enabling the postgis extension, do you have any suggestions on what is best to do? I'm aware that simply putting it in a migration may not work.

pablobm:
There's a snag. After a few weeks working on my Linux/amd64 machine, I switched to my MacBook Pro M1 and found that PostGIS wasn't working on the Docker and Dev Container setups. Turns out that the PostGIS team don't supply Docker images for arm64.

For the "classic" Docker setup, I have added a commit to make it work. It leverages the Postgres container having its own Dockerfile, using `apt install` to get the appropriate package. An alternative would have been adding `platform: linux/amd64` to this container in `docker-compose.yml`, but I don't know what the side effects would be. Eg: would it work on Windows?

For the Dev Container setup... I'm stumped. The setup has stopped working for me on macOS (it works on Linux/amd64). Therefore I'm not able to test this. I have spent too long today looking into this and I'm parking it for a moment. Is anyone able to test? Any ideas?

seuros:
Docker cares only about the architecture of the host.

Check postgis-with-extension repo i have my github.

It has arm and amd support. Fell free to take anything or use it. ( it always pointing to lastest stable version )

pablobm:
I have a fix at https://github.com/openstreetmap/openstreetmap-website/pull/6810 that hopefully will unblock this.

pablobm:
I have rebased after merging the PR above. This works well now.
- I have added a file `.devcontainer/Dockerfile.postgres`, similar to the existing `docker/postgres/Dockerfile`, which installs Postgis using Apt. This works around Postgis not having images for arm64 (at least not official ones).
- Both containerised setups work on my arm64 and amd64 machines.

gravitystorm:
Great. I think it's just the ruby min version thing that's blocking this now for me. [Ways forward are discussed above](https://github.com/openstreetmap/openstreetmap-website/pull/6713#pullrequestreview-3751759307).

tomhughes:
The fact that it will not be deployable in production is not a blocker?

gravitystorm:
> The fact that it will not be deployable in production is not a blocker?

Do you mean that nobody has added `package "postgresql-15-postgis-3"` to the OSMF chef repo, or are you warning us that there's some more deep-seated reason that this cannot be deployed?

tomhughes:
I mean, as previously explained, that a migration that tries to enable postgis needs to run as a superuser and we do not run migrations as a superuser currently and it's not clear that we easily can if we assume that we don't want the production rails code to be using a superuser account!

gravitystorm:
> a migration that tries to enable postgis needs to run as a superuser

I've tested locally that "CREATE EXTENSION IF NOT EXISTS" as a non-superuser will succeed, if the extension already exists. Could this be a way forward - i.e. to create the extension via chef as a superuser, and then run the migrations in this PR as normal?

tomhughes:
Maybe, it's sort of what I did to get the test instance deployed on dev except there I also inserted the schema migration record manually.

The other option would be to create a separate user for running migrations but that would need a bit of work especially in the dev environment and would require some way to override the database connection details when running the migrations.

There's also the question of how it impacts developers and third party users of the code.

pablobm:
> Great. I think it's just the ruby min version thing that's blocking this now for me. Ways forward are discussed above.

Sounds good, I have now implemented my preferred of [the ways forward I proposed above](https://github.com/openstreetmap/openstreetmap-website/pull/6713#issuecomment-3854041331). Specifically: I have created a fork of the gem, allowing for Ruby 3.2. This is following the conversation with the author (see comment linked above and subsequents) where it seems unlikely that our app will be affected by the Ruby 3.2 incompatibility.

> The fact that it will not be deployable in production is not a blocker?

Happy to discuss what can be done here. I did ask at https://github.com/openstreetmap/openstreetmap-website/pull/6713#issuecomment-3878016798

tomhughes:
If I had a good solution I would have offered it when you asked - the problem is that I don't.

woodpeck:
Note, DWG currently runs a script that auto-hides anonymous notes in certain regions, which would be superseded by an API functionality like the one described here. The current DWG implementation can only deal with rectangular areas for which the point-in-polygon test is trivial and does not require PostGIS. Not saying that you shouldn't use PostGIS but if it turns out to be a pain, rectangles are your friend.

pnorman:
I know a comment of this nature is best done early into the review process, but I've been looking over this issue and openstreetmap/openstreetmap-website#6570 and can't find where PostGIS was decided as the best way to do this. Are we satisfied that it's the best option?

pablobm:
Still better asked now than after merge! :smile: 

My view is that we are introducing PostGIS here as a strategic move. This PR adds an ancillary feature that can serve as a first, localised use case for PostGIS without affecting OSM at large.

From conversations, I understand that over the years features have been proposed that either required PostGIS or have received a sub-par implementation due to the lack of it. Introducing PostGIS at last breaks a psychological barrier that will allow the creation of new and more performant features in the future.

pablobm:
> There's also the question of how it impacts developers and third party users of the code.

This is an interesting point. I have not had issues in any of the environments I've tried, but I understand there will be use cases that escape me.

One specific case I'm wondering about: @hlfan, I believe you work on Windows? Would you be able to test this PR and tell us if you had any issues getting PostGIS working?

@tomhughes - Are there any specific use cases that you have in mind?

pablobm:
I see there's been some movement on the operations side (eg: https://github.com/openstreetmap/chef/pull/839). Does the team have an idea of what the current blockers are?

pablobm:
As of https://github.com/openstreetmap/openstreetmap-website/pull/6994 the minimum Ruby in this project is v3.3, so I have rebased&updated this PR to drop the custom branch that allowed for compatibility of activerecord-postgis with Ruby 3.2 .

Gemfile
pablobm: For posterity: the Ruby version question is now moot as this project has changed the minimum to be Ruby 3.3.

.github/workflows/lint.yml
```diff
@@ -157,6 +161,10 @@ jobs:
...
```
gravitystorm: I want to just quibble slightly over the package that we want to install:

* `postgis` - this only contains client-side stuff, i.e. "PostGIS userland binaries for importing and
 exporting shape and raster files: pgsql2shp, raster2pgsql, and shp2pgsql." We don't need these. However, it also "recommends" the metapackage `postgresql-postgis`, which is more useful...
* `postgresql-postgis` is a metapackage that points at whichever `postgresql-X-postgis-Y` is default on the distribution
* `postgresql-X-postgis-Y` contains the actual server-side extension (and depends on some other minor packages).

In some places we're installing `postgis` but relying on (hoping?) the "recommends" packages get installed too. In other places we're installing the `postgresql-X-postgis-Y` package explicitly, but that will need to kept in sync whenever the underlying distro/image is updated. 

Is it preferable to standardise on the `postgresql-postgis` package throughout?

gravitystorm (approved):
I don't have any blockers, I'm happy to merge this unless someone else spots something.

Minor comment about package names inline.

.github/workflows/lint.yml
seuros: The question i have, why do you need postgis here ? This is a linting workflow.

.rubocop.yml
tomhughes: Yes but 3.3 is our baseline minimum supported ruby so that is what we test with.

.github/workflows/lint.yml
tomhughes: Because that linter is the database consistency check, so it needs to run the migrations to build the database, which means it needs postgis.

gravitystorm: It's in the database_consistency linter, which needs to compare the code against what's in the database.

tomhughes (commented):
I'm a bit confused about the intended scope of "moderation zones" to be honest... I hadn't looked at this in much detail before but my understanding was that this was mostly creating a framework to define zones where various activities could be restricted, a bit like the existing ACL model but for geographic areas.

It appears that despite the generic name these zones are in fact tied explicitly to notes, so should they have a more explicit name? or alternatively something like a type field that indicates what is being moderated and then the note moderation is only applied to zones of the right type, or maybe even a separate one-many table that defines what restrictions are applied to each zone?

Other than that my main comments are that the commit structure doesn't look very good - quite a lot of the last three commits probably needs to be merged into the earlier commit where postgis is added to the database I think so that we don't have commits where, for example, lint can't possibly work.

I'd also question the split with most of the model being introduced in one commit and then the `belongs_to` rules being added in the next one and would move those into the first commit.

Finally "proof of concept" is a horribly generic commit message that doesn't tell us what it's doing at all.

.github/workflows/lint.yml
pablobm: OK, so I had a look and this gets a bit tricky...

- The Debian package `postgresql-postgis` assumes the Postgres version that would be default in the distro. So for example, Debian 13 defaults to Postgres 17. Therefore `postgresql-postgis` will install `postgresql-17-postgis-3`, even if we have `postgresql-15` installed.
- The Docker images `postgres:14`, etc, install the requested version of Postgres from source, on Debian. If we then install `postgresql-postgis`, it will install the wrong flavour of Postgis, or in the wrong library dir or whatever. Doesn't matter, doesn't work.
- In the `lint` and `tests` CI workflows we can use `postgresql-postgis` because they use the default Postgres... which I can't tell where it exactly comes from, but whatever.
- Same for `MANUAL_INSTALL.md`, where at least I can see where Postgres comes from. For the Fedora instructions, I'm leaving `postgis` as that appears to [the simplest option there](https://packages.fedoraproject.org/search?query=postgis).
- For the Dev Container, we are using Postgres 15 (because [reasons](https://github.com/openstreetmap/openstreetmap-website/pull/6810)) so that requires explicitly stating `postgresql-15-postgis-3`. Similar in the "normal" Docker setup, with `postgresql-14-postgis-3`.

gravitystorm: > * The Docker images `postgres:14`, etc, install the requested version of Postgres from source, on Debian. If we then install `postgresql-postgis`, it will install the wrong flavour of Postgis, or in the wrong library dir or whatever. Doesn't matter, doesn't work.

OK, thanks for looking into this. I assumed that on `postgres:14` the package `postgresql-postgis` would depend on `postgres-14-postgis-3` but from the sound of it, that's not necessarily true and the `postgresql-postgis` package might point at something else. I'm not sure that it's the most helpful thing from whoever makes those images, but hey-ho - we can deal with it ourselves by being specific where we need to.

gravitystorm:
> It appears that despite the generic name these zones are in fact tied explicitly to notes, so should they have a more explicit name?

I would say that they are being implemented for notes *first*, with the intention that when that is fully implemented we can add other types in future e.g. changesets.

> alternatively something like a type field that indicates what is being moderated and then the note moderation is only applied to zones of the right type, or maybe even a separate one-many table that defines what restrictions are applied to each zone?

Until now I haven't considered having that level of selectivity. I assumed that a zone would apply to both notes and also whatever else is implemented at that stage. I think we can implement what we have (i.e. a zone applies to everything (currently only notes)) and revisit it if moderators need more fine-grained controls in future.



> Other than that my main comments are that the commit structure doesn't look very good

That's a good point.

pablobm:
> something like a type field that indicates what is being moderated (...) or maybe even a separate one-many table that defines what restrictions are applied to each zone

This crossed my mind, but I think it's premature. When :crossed_fingers: we extend moderation zones to new types of data, then we can see how we handle that. I don't see a reason to do it now.

> the commit structure doesn't look very good

I'll have a look.

pablobm:
Commits should now be better structured :card_index_dividers:

tomhughes:
> Until now I haven't considered having that level of selectivity. I assumed that a zone would apply to both notes and also whatever else is implemented at that stage. I think we can implement what we have (i.e. a zone applies to everything (currently only notes)) and revisit it if moderators need more fine-grained controls in future.

I'd accept that argument more if what we had here applied to all notes, but it doesn't, it only applies to notes by people that aren't logged in.

So are you saying that if we extended it to map edits then it would also have to apply to non-anonymous notes at that point, or that it would to anonymous notes but to all map edits which seems very confusing to me.

tomhughes:
The structure of this looks good now, and my only remaining concerns are around the naming and scope of the zones as discussed.

As that's something that can be resolved as and when we want to expand the use of the zones to other things I think this is good to merge now - thanks for all the work on it.

Issue #6699, opened 2026-01-13
OAuth2: Newly created and email-confirmed accounts not redirected to authorization page


bhatganeshdarshan:
### URL

https://hdyc.neis-one.org/

### How to reproduce the issue?

# OAuth2 Authorization Redirect Bug for Newly Created Users

---

### 1. Start OAuth2 Authorization Flow

Visit:

https://resultmaps.neis-one.org/oauth2/?login&myclient=HDYC&origin=https://hdyc.neis-one.org

Expected:  
You should be redirected to the login page since you're not authenticated.

---

### 2. Create a New Account

1. Click **"Sign up"** on the login page.  
2. Fill in the registration form:
   - Display Name
   - Email
   - Password
3. Click **"Sign up"**.

Confirmation email sent to your email 

---

### 3. Confirm Email

Option A :  
- Check your email inbox.  
- Click the confirmation link.
- Come back and login with confirmed email and password

Option B :  
- No need to confirm the mail 
- Just use already existing account email and password to login 

---

### 4. Observe the Bug

Expected Behavior:  
After Option A / Option B 
- Application name requesting access  
- Requested permissions/scopes  
- **Authorize** and **Deny** buttons  

Actual Behavior:  
You are redirected to the OpenStreetMap home page (`/`) instead of the OAuth2 authorization page.

---

### 5. Verify the Workaround (Existing Accounts)

1. Go back to the login page:

https://resultmaps.neis-one.org/oauth2/?login&myclient=HDYC&origin=https://hdyc.neis-one.org

2. Log in using an existing account (not newly created).

Result:  
You are correctly redirected to the OAuth2 authorization page.

### 

### Screenshot(s) or anything else?
Expected page : 
<img width="1911" height="906" alt="Image" src="https://github.com/user-attachments/assets/bda969e4-ed31-43b1-9bbe-a243ff759b4f" />

Here is more details related to this issue that i have found in  https://github.com/streetcomplete/StreetComplete/ repo : 

https://github.com/streetcomplete/StreetComplete/issues/6688#issuecomment-3739134513

tomhughes:
If you're going to give example URLs please base them on the live site, not your own development environment, so that people can easily test them.

There's also no need to open a bug if you're going to submit a PR as the PR can stand on it's own and having a bug as well just splits the discussion in two.

I'll need to look later but I suspect there may be something deliberate (if unfortunate) about this due to the desire to send new users to the welcome page.

[...]

PR #6714, opened 2026-01-15
Fix: Preserve OAuth referer when resending confirmation email 

bhatganeshdarshan:
# Description 

- When users sign up through an OAuth2 authorization flow and click "Resend confirmation email", the OAuth referer information is lost. The resent confirmation email does NOT contain the referer parameter with the oauth_return_url, causing the "Continue authorization" button to disappear from the welcome page after email confirmation.

- Note : After signup if user doesnt click on "resend confirmation" , the initial confirmation link which was sent to email properly redirects to welcome page with "Continue authorization" , the above issue occurs only when user clicks "Resend confirmation mail"

Related to #6699 

### Fix 
- Store referer in session during user signup
- Pass stored referer when resending confirmation email
- Include referer in redirect after resending confirmation

This ensures that users who sign up via OAuth2 and click 'Resend confirmation email' will still see the 'Continue authorization' button on the welcome page after confirming their account.

Added unit tests:
- test_confirm_success_with_oauth_referer: Verifies initial signup confirmation preserves OAuth referer
- test_confirm_resend_preserves_oauth_referer: Verifies resending confirmation also preserves OAuth referer (this test fails without the fix, proving the bug exists)

### DIFF
--- a/app/controllers/confirmations_controller.rb
+++ b/app/controllers/confirmations_controller.rb
@@ -60,11 +60,12 @@ class ConfirmationsController < ApplicationController
     if user.nil? || user.id != session[:pending_user]
       flash[:error] = t ".failure", :name => params[:display_name]
     else
-      UserMailer.signup_confirm(user, user.generate_token_for(:new_user)).deliver_later
+      referer = session[:referer]
+      UserMailer.signup_confirm(user, user.generate_token_for(:new_user), referer).deliver_later
       flash[:notice] = { :partial => "confirmations/resend_success_flash", :locals => { :email => user.email, :sender => Settings.email_from } }
     end
 
-    redirect_to login_path
+    redirect_to login_path(:referer => session[:referer])
   end
 
   def confirm_email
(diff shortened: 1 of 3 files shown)


bhatganeshdarshan:
Hi! Just a gentle ping to check if someone has time to review this PR when convenient. Thanks!

tomhughes:
So this needs to delete `:referer` from the session in all the places that currently delete `:pending_user` from the session, or it will stay around forever.

I'm also not sure about the way the tests are written - they seem to be testing more than is probably necessary and they look more like integration or system tests than controller tests at the moment but I think that may mean they're doing more than they need to rather than that they should be moved.

The first test is probably redundant - we already have tests further up that check that the referer is preserved normally I think and OAuth is not special, it's just another example of a referer. It's also not truly following through each stop as demonstrated by the fact that at the end it just makes up the referer URL for the final confirm rather than extracting it from the email.

The second test is mostly modelled on the first one but oddly stops sooner without running the final test that the confirm redirects as expected.

The critical thing that should be tested is the thing that's being fixed here - that if you visit the signup page with a referer and then visit the resend confirmation page that the subsequent email preserves the referer and that visiting the confirmation link from the email redirects to the expected referer from the original signup.

Issue #1494, opened 2017-03-21
Navigation icon being used for geolocation feature


xenotropic:
The icon that is used for geolocation on the OSM site is currently this:

![osm-geoloc-icon](https://cloud.githubusercontent.com/assets/4050617/24165769/8e0dd1fe-0e2e-11e7-8ce9-dbaffbe9ff95.PNG)

That icon more typically used for navigation. I would expect that clicking on such an arrow would provide me with turn-by-turn directions. There are other icons that much more clearly mean "show me my location".

While there is not one definitive agreed-upon standard set of icons, [The Noun Project](https://thenounproject.com/) includes many cc-licensed icons that are tagged and organized. If you search for "navigation", then you get a lot of arrows that are similar to the one above. 

If you search for "current location" on the Noun Project, or just on Google Image Search, it yields many crosshair-like icons; such as:
![noun_814578_cc](https://cloud.githubusercontent.com/assets/4050617/24166587/f8bc9d58-0e30-11e7-8bc8-976d70d588b6.png)

(cc-by shashank singh, [source](https://thenounproject.com/search/?q=current%20location&i=814578))

Alternatively, if you search for "my location" on the Noun Project it gives many icons of a "dropped pin" with a human associated with it, e.g.: 

![noun_607182_cc](https://cloud.githubusercontent.com/assets/4050617/24166645/2dcd6a36-0e31-11e7-9d0d-4005d1ee7249.png)
 (cc-by, Ismael Ruiz, [source](https://thenounproject.com/term/my-location/607182/))

Either of those would be more clearly indicative of what the location button actually does. Please consider using one of those alternatives.  I do not mean that you necessarily have to use one of these specific icons; I am just using them as representative of icons that, in general, are used to indicate the user's current position, so some other "crosshair" or "person-with-pin" icon would work as well. If cc-by is insufficient, I volunteer to go hunting for one that is public domain/cc0.

pnorman:
Looking at Google Image results, I'm in agreement with changing. The arrow-like icon is most associated with navigation, and crosshairs of some kind with **my/current** location. Pins are better associated with other locations which aren't necessarily the user's, which is how we currently use them.

dieterdreist:
2017-03-23 0:37 GMT+01:00 Paul Norman <notifications@github.com>:

> Looking at Google Image results, I'm in agreement with changing. The
> arrow-like icon is most associated with navigation, and crosshairs of some
[...]

PR #6716, opened 2026-01-16
Update route markers and context menu icons to cursor/flag metaphor (…

### LINKED ISSUE #929: Adjust colouring of icon when sharing with position to avoid route icon conflict
WIth the new routing now available, viewing a shared map with an icon appears to indicate that the 'end point' is already set, though the location is not set in the 'end point' box.

I'd suggest changing the colour of the icon which appears on a shared map to be eg. blue, so that it does not imply that it is the start or end of the trip.

On completion of this, another issue might track eg. having a popup on a shared icon to make it the start or end of a route.

aNsHuL5217:
Resolves #929.

### Description
This PR updates the route markers and context menu icons to use standard navigation metaphors, replacing the previous "media player" style icons (triangle/square) that were causing user confusion.

* **Map Markers (`_markers.html.erb`):** Updated the internal SVG paths for the start and end markers. The start marker now uses a "navigation arrow" (cursor) and the end marker uses a "flag".
* **Context Menu (`contextmenu.js`):** Updated the Bootstrap icons to `bi-cursor` and `bi-flag` to ensure the menu options visually match the new map markers.

**Screenshots:**

Screenshot 2026-01-16 at 2 27 15 PM

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -24,8 +24,8 @@ OSM.Directions = function (map) {
   };
 
   const endpoints = [
-    OSM.DirectionsEndpoint(map, $("input[name='route_from']"), { icon: "play", color: "var(--marker-green)" }, endpointDragCallback, endpointChangeCallback),
-    OSM.DirectionsEndpoint(map, $("input[name='route_to']"), { icon: "stop", color: "var(--marker-red)" }, endpointDragCallback, endpointChangeCallback)
+    OSM.DirectionsEndpoint(map, $("input[name='route_from']"), { icon: "start", color: "var(--marker-green)" }, endpointDragCallback, endpointChangeCallback),
+    OSM.DirectionsEndpoint(map, $("input[name='route_to']"), { icon: "destination", color: "var(--marker-red)" }, endpointDragCallback, endpointChangeCallback)
   ];
 
   const expires = new Date();
--- a/app/views/layouts/_markers.html.erb
+++ b/app/views/layouts/_markers.html.erb
@@ -18,8 +18,8 @@
          "tick" => { :"stroke-linecap" => "round", :"stroke-linejoin" => "round", :fill => "none", :d => "M7.157 14.649Q8.9 16 11.22 18.761 14.7 11.7 17.843 8.239" },
          "cross" => { :"stroke-linecap" => "round", :d => "m7.5 8 10 10m0-10-10 10" },
          "minus" => { :"stroke-linecap" => "round", :d => "M5.75 13h13.5" },
-         "play" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M10 17.5v-9l7 4.5z" },
-         "stop" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M9 9.5h7v7H9z" },
+         "start" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M5.5 13h7v7l5-12z" },
+         "destination" => { :"stroke-linecap" => "round", :"stroke-linejoin" => "round", :fill => "#fff", :d => "M10 21 7 9c5-3 6 2 11 0l1 4c-5 2-6-3-11 0" },
          "dot" => { :"stroke-linecap" => "round", :fill => "#fff", :d => "M11.5 10a1 1 0 0 0 2 5 1 1 0 0 0-2-5" }
        } %>
 
--- a/app/assets/javascripts/index/contextmenu.js
+++ b/app/assets/javascripts/index/contextmenu.js
@@ -34,7 +34,7 @@ OSM.initializations.push(function (map) {
   const contextmenuItems = [
     {
       id: "menu-action-directions-from",
-      icon: "bi-play",
+      icon: "bi-cursor",
       text: OSM.i18n.t("javascripts.context.directions_from"),
       callback: () => {
         const params = new URLSearchParams({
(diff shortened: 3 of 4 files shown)


app/views/layouts/_markers.html.erb
```diff
@@ -18,8 +18,8 @@
          "tick" => { :"stroke-linecap" => "round", :"stroke-linejoin" => "round", :fill => "none", :d => "M7.157 14.649Q8.9 16 11.22 18.761 14.7 11.7 17.843 8.239" },
          "cross" => { :"stroke-linecap" => "round", :d => "m7.5 8 10 10m0-10-10 10" },
          "minus" => { :"stroke-linecap" => "round", :d => "M5.75 13h13.5" },
-         "play" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M10 17.5v-9l7 4.5z" },
-         "stop" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M9 9.5h7v7H9z" },
+         "play" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M5.5 13h7v7l5-12z" },
+         "stop" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M10 21 7 9c5-3 6 2 11 0l1 4c-5 2-6-3-11 0" },
```
hlfan: This should set the stroke linecap too for the outline to go around the whole pole.

app/views/layouts/_markers.html.erb
```diff
@@ -18,8 +18,8 @@
          "tick" => { :"stroke-linecap" => "round", :"stroke-linejoin" => "round", :fill => "none", :d => "M7.157 14.649Q8.9 16 11.22 18.761 14.7 11.7 17.843 8.239" },
          "cross" => { :"stroke-linecap" => "round", :d => "m7.5 8 10 10m0-10-10 10" },
          "minus" => { :"stroke-linecap" => "round", :d => "M5.75 13h13.5" },
-         "play" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M10 17.5v-9l7 4.5z" },
-         "stop" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M9 9.5h7v7H9z" },
+         "play" => { :"stroke-linejoin" => "round", :fill => "#fff", :d => "M5.5 13h7v7l5-12z" },
```
hlfan: Now that we don't use the media control metaphors anymore "play" and "stop" should probably be renamed too.

app/views/layouts/_markers.html.erb
aNsHuL5217: I have added `stroke-linecap: "round"` to the destination marker definition. The outline now wraps around the pole correctly.

app/views/layouts/_markers.html.erb
aNsHuL5217: Agreed. Since the media metaphors no longer fit, I have renamed the keys `play` → `start` and `stop` → `destination` across the codebase (including `directions.js`, the markers partial, and the map layout) to match the new semantics.

app/assets/javascripts/index/directions.js
```diff
@@ -274,3 +274,4 @@ OSM.Directions.addEngine = function (engine, supportsHTTPS) {
     OSM.Directions.engines.push(engine);
   }
 };
+
```
hlfan: Could you drop this last newline and squash these commits down to one?

aNsHuL5217: Done! I removed the extra newline and squashed everything into a single commit.

pablobm:
I like the choice of icons (thank you!) although now there's the problem that "Show My Location" looks like "Start Route". Should we change the latter to something else? Perhaps the classic [crosshair](https://icons.getbootstrap.com/icons/crosshair/)?

(I'm going through my GH notifications now, so I might be missing a later conversation where this is brought up).

gravitystorm:
See https://github.com/openstreetmap/openstreetmap-website/issues/1494 although there's definitely some platform-related conventions.

hlfan:
> Perhaps the classic [crosshair](https://icons.getbootstrap.com/icons/crosshair/)?

That's already used by the "centre map here" context menu item.

aNsHuL5217:
You are right, `bi-crosshair` is already taken by the context menu.

Since the "Show My Location" button shouldn't look like the navigation arrow (start point), perhaps we could use `bi-record-circle` (GPS dot style) or `bi-bullseye`? These are standard for geolocation but distinct from the "Centre Map" crosshair.

pablobm:
Idea: "start route" to use the same icon as the selected mode of transport. So it would be dynamic and be either a car, a bicycle, or a walking person. The default (shown in the contextual menu) would be the walking person.

hlfan:
I like this idea. But do you also mean to change it in the marker pin?
That would require a restructuring of the markers.
Otherwise we once again have non-matching icons.

pablobm:
Yes: marker pin would reflect the mode of transport, as it basically says "you are here" so it makes sense to show "your self/car/bicycle is here". We don't have that info in the contextual menu, so default to walking person.

I left the suggestion without looking at the code, so it might not be practical :sweat_smile:

hlfan:
Hmmm... the marker stroke width is a bit high for such details, 


```
M12.5 5.5v0m-.5 4.5v5a.5.5 0 001 0V10M9.5 21l2-3 1-3 2 3 1 3m-7-6 .9-3L12 10a.7.7 0 011 0l2.5 2.5L18 13
```
And the other icons have even more details.

PR #6720, opened 2026-01-18
Deconvolute linkify logic

hlfan:
This PR adds flags to the linkify method to more clearly indicate what is being executed.

This is built on dropped parts from #5862 and #6674.

### DIFF
--- a/lib/rich_text.rb
+++ b/lib/rich_text.rb
@@ -105,7 +105,15 @@ module RichText
     private
 
     def gsub_pairs_for_linkify_detection
-      []
+      Array
+        .wrap(Settings.linkify&.detection_rules)
+        .select { |rule| rule.path_template && rule.patterns.is_a?(Array) }
+        .flat_map do |rule|
+          expanded_path = "#{rule.host || "#{Settings.server_protocol}://#{Settings.server_url}"}/#{rule.path_template}"
+          rule.patterns
+              .select { |pattern| pattern.is_a?(String) }
+              .map { |pattern| [Regexp.new("(?<=^|#{URL_UNSAFE_CHARS})#{pattern}", Regexp::IGNORECASE), expanded_path] }
+        end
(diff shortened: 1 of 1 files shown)


lib/rich_text.rb
```diff
@@ -87,18 +87,33 @@ def sanitize(text)
...
-               .then { |html| expand_link_shorthands(html) }
-               .then { |html| expand_host_shorthands(html) }
-               .then { |html| auto_link(html, mode) }
-               .html_safe
+    def linkify(text, mode = :urls, hosts: true, paths: true)
+      link_attr = 'rel="nofollow noopener noreferrer" dir="auto"'
+      html = ERB::Util.html_escape(text)
+
+      html = expand_link_shorthands(html) if paths
+      html = expand_host_shorthands(html) if hosts
```
pablobm, line 95: Is there any instance when `hosts` will be false? If not, should we remove the option altogether?

hlfan: Arguably, html shouldn't have osm link hosts shortened, but I don't know if html richtext is even used here.

pablobm: Do I understand correctly that "text" converts text to HTML, "markdown" converts markdown to HTML, and "html"... leaves things as they are? Should the "html" one be removed completely?

hlfan: The input is still sanitized, and the Richtext.to_text function goes above leaving things as they are too.
@tomhughes might know more about why this exists.

Coming back to the `hosts` flag in `linkify`, I also think the symmetry makes even an underused option worthwhile here.

tomhughes: It looks like `to_text` is mostly used for things like plain text emails though obviously it actually includes the formatting rather than trying to remove it though that's not terrible for markdown at least.

tomhughes:

PR #6721, opened 2026-01-18
Load make-plural as ES module

HolgerJeromin:
### Description

ref #6664
### How has this been tested?

Not tested. Will most probably will not work but give an example for the rails experts.


tomhughes:
I already have a working version of this... Just running final tests now.

HolgerJeromin:
Ref #6722

PR #6722, opened 2026-01-18
Update to make-plural 8.x

tomhughes:
This updates make-plural to 8.x and makes it load as a module.

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -20,3 +20,5 @@
 //= link leaflet/dist/images/marker-shadow.png
 
 //= link @mapbox/mapbox-gl-rtl-text/dist/mapbox-gl-rtl-text.js
+
+//= link make-plural/cardinals.js
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -13,7 +13,6 @@
 //= require leaflet.zoom
 //= require leaflet.locationfilter
 //= require i18n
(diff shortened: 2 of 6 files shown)


HolgerJeromin:
`‎config/eslint.config.mjs`‎ still has a reference as a global.

tomhughes:
> `‎config/eslint.config.mjs`‎ still has a reference as a global.

That should be fixed now.

pablobm:
Seems ok in production, but in development I'm getting this error:

```
Uncaught (in promise) TypeError: can't access property "plurals", root is undefined
```

tomhughes:
Did the load fail? Though presumably it won't resolve the promise in that case, or will it?

pablobm:
It loads fine and I'm not seeing any other load fails. The error comes from within the module. Excerpt:

```
(function (root, plurals) {
  Object.defineProperty(plurals, '__esModule', { value: true });
  if (typeof define === 'function' && define.amd) define(plurals);
  else if (typeof exports === 'object') module.exports = plurals;
  else root.plurals = plurals;
}(this, {
af: a,

ak: b,

am: c,
```

Where `root.plurals = plurals` appears to be the issue. This is JS import land, which I'm very out of touch with...

tomhughes:
That's the old version of make-plural though... Have you precompiled your assets (and not updated them since the change)?

pablobm:
:flushed: Argh, assets... Thank you, fixed now.

PR #6723, opened 2026-01-18
add copilot instructions

stillhart:
### Description

`.github/copilot-instructions.md` seems to be the default AI/LLM base instruction file to guide an LLM through a repo.
I suggest to add one to comfort the next generation of vibe coders. 

### How has this been tested?
I use it. 😅

### Did I make this?
No, an LLM did that for me. 🤷‍♂️

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,110 @@
+<!-- Copilot / AI agent instructions for contributors working on openstreetmap-website -->
+
+# Purpose
+
+Short, actionable guidance to help an AI coding agent be productive in this codebase.
+
+# Quick start (commands)
+- Bootstrap a dev environment: `bin/setup` (idempotent; runs `bin/dev` unless `--skip-server`).
+- Start the app: `bin/dev` (executes `bin/rails server`).
+- Install JS deps: `yarn install` (see `package.json`).
+- Install Ruby gems: `bundle install` (the repo uses `Gemfile`).
+- Database prepare: `bin/rails db:prepare` (or `bin/rails db:reset` for a full reset).
+- Run tests: `bin/rails test` or `bundle exec rails test`.
+- Lint/security checks: `bundle exec rubocop`, `bundle exec brakeman`.
+
+# Big-picture architecture (what to know first)
+- This is a Ruby on Rails application (Rails ~8.1) that contains:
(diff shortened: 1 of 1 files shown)


tomhughes:
So long as the instruction is "do not touch this repo and certainly do not 'vibe' code anything" then fine...

gravitystorm:
Thanks for the suggestion, but I'm not convinced about merging this, for a number of reasons. I speak with experience of using LLMs (specifically, Claude) in multiple projects.

* Each LLM/agent has its own preference for where to put instructions, e.g. CLAUDE.md or AGENTS.md. I don't think there's really any conventions yet.
* The contents of the agents files are usually just regurgitated versions of the instructions found elsewhere in the repo, without much original content. This makes them more of a "generated" or "compiled" source of knowledge, and such artefacts aren't generally stored under source control.
* If the agents file contained original content (e.g. telling it how to do something, but we had no other documentation about how to do that) then it would be a candidate for source control. But I'd be really really curious why we would be telling the agents about something that we're not telling everyone else about!
* I don't want to end up with 10 different agents files, one for each LLM/agent.
* However, it's not clear what value there is giving them all a common agents file, rather than each agent generating their own files themselves, which will (perhaps) match their specific instructional needs more closely.
* Since everything moves so quickly, it's not even clear how to maintain these files. Would we have to regenerate them every month or two when there's new agent releases? Do we treat them as diffable text, or will the output each regeneration be completely different making diffs meaningless? Or do we edit them manually e.g. when we add a new linter?

It's different for organizations that have many developers, all using the same version of the same agent, where it saves each developer a few seconds (e.g. waiting on `/init` run) when working on a common codebase. 

I'm not even really convinced that the concept of an agents file has any longevity, as LLMs develop. Between internal knowledge (e.g. most LLMs will know how to run test suites) and human-orientated documentation (e.g. INSTALL.md) then it seems to me like the agents files are just a minor implementation detail and could in most cases (and very likely so in ours) be hidden from developers. 

But in summary, since this stuff is all so new, I'm happy to be persuaded that I'm wrong on any of the above!

gravitystorm:
Since there haven't been any responses in this thread, I'm going to decline this PR, based on my concerns above.

I'll reiterate that this stuff is all so new, so I'm happy to be persuaded that I'm wrong on any of this.

Thanks @stillhart for the suggestion!

PR #6724, opened 2026-01-18
Check response.ok before parsing JSON in query tool

### LINKED ISSUE #6667: Query tool fails - likely due to data change
### URL

_No response_

I used the query tool to check some data in a municipality that I just (as in perhaps 5 minutes back) merged with a further one and got the error message below. Probably not actionable but at least FYI.

Image

SykoDaedalus:
Added a if-else to check response.ok before parsing JSON in query tool

### Description
This PR fixes issue #6667.
The query tool no longer crashes when the network or server returns an error.

## Problem
The query tool tried to read every response as JSON.
When the Overpass API returned an error page or the request failed, this caused unclear errors like:

`JSON.parse: unexpected character at line 1 column 1 of the JSON data`

## Solution
The tool now checks the response status before reading JSON.
If the request fails, it shows a clear message, for example:

`Error contacting https://overpass-api.de/api/interpreter: Failed to fetch`

### How has this been tested?

- Linting passes (eslint, erb_lint, rubocop)

- Manually testing network failure using browser offline mode
 The tool now shows “Failed to fetch”

- Normal queries still work as expected

Fixes #6667

### DIFF
--- a/app/assets/javascripts/index/query.js
+++ b/app/assets/javascripts/index/query.js
@@ -171,7 +171,12 @@ OSM.Query = function (map) {
       credentials: OSM.OVERPASS_CREDENTIALS ? "include" : "same-origin",
       signal: $section.data("ajax").signal
     })
-      .then(response => response.json())
+      .then(response => {
+        if (response.ok) {
+          return response.json();
+        }
+        throw new Error(response.statusText || response.status);
+      })
       .then(function (results) {
         let elements = results.elements;
 
--- a/test/system/query_features_test.rb
+++ b/test/system/query_features_test.rb
@@ -214,6 +214,7 @@ class QueryFeaturesSystemTest < ApplicationSystemTestCase
                 elements = data.includes("is_in") ? #{enclosing_elements.to_json} : #{nearby_elements.to_json};
 
           return Promise.resolve({
+            ok: true,
             json: () => Promise.resolve({ elements })
           });
         }


mmd-osm:
We might need to fine tune the exact HTTP status codes to check. Overpass API uses slightly different ones: [link](https://github.com/drolbr/Overpass-API/blob/a0db4f392f744d5e1304331edbf542ef6d6ce2fa/src/overpass_api/frontend/basic_formats.cc#L38-L49). In any case, I would remove HTTP 509 from the list.

As a heads up, sometimes Overpass API generates fairly large HTML/XML error responses, which could look a bit odd on the screen when rendered as plain text. Since this is happening rather infrequently, I would wait for more user feedback.

Edit: so adding 429 would exactly exhibit the issue, where a user would see this:

```
Enclosing features

    Error contacting https://overpass-api.de/api/interpreter: <?xml version="1.0" encoding="UTF-8"?> <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"> <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en"> <head> <meta http-equiv="content-type" content="text/html; charset=utf-8" lang="en"/> <title>OSM3S Response</title> </head> <body> <p>The data included in this document is from www.openstreetmap.org. The data is made available under ODbL.</p> <p><strong style="color:#FF0000">Error</strong>: runtime error: open64: 0 Success /osm3s_osm_base Dispatcher_Client::request_read_and_idx::rate_limited. Please check /api/status for the quota of your IP address. </p> </body> </html> 
```

whereas, when using:

```
        if (response.status !== 400) {
          throw new Error(status);
        }
```

it looks a bit saner:

```

Enclosing features

    Error contacting https://overpass-api.de/api/interpreter: Too Many Requests
```

pablobm:
Ooof. The tests are failing because Overpass queries are stubbed for testing at:

https://github.com/openstreetmap/openstreetmap-website/blob/5b3210c5f637596ed65e75c2173990055b11fe5c/test/system/query_features_test.rb#L216-L218

Only `json` is present in the stub, so you'll have to add `ok` and that should be enough.

SykoDaedalus:
> Ooof. The tests are failing because Overpass queries are stubbed for testing at:
> 
> https://github.com/openstreetmap/openstreetmap-website/blob/5b3210c5f637596ed65e75c2173990055b11fe5c/test/system/query_features_test.rb#L216-L218
> 
> Only `json` is present in the stub, so you'll have to add `ok` and that should be enough.

Thanks for the help, it worked.

test/system/query_features_test.rb
```diff
@@ -214,6 +214,7 @@ def stub_overpass(nearby_elements: [], enclosing_elements: [])
                 elements = data.includes("is_in") ? #{enclosing_elements.to_json} : #{nearby_elements.to_json};
 
           return Promise.resolve({
+            ok:true,
```
tomhughes: Space after the colon please, for consistency:

```suggestion
            ok: true,
```

app/assets/javascripts/index/query.js
```diff
@@ -224,9 +238,7 @@ OSM.Query = function (map) {
       })
       .catch(function (error) {
         if (error.name === "AbortError") return;
-
```
tomhughes, line 227: Why are we getting rid of these blank lines?

app/assets/javascripts/index/query.js
```diff
@@ -171,7 +171,21 @@ OSM.Query = function (map) {
...
+        const status = response.statusText || response.status;
+
+        if (response.status !== 400) {
+          throw new Error(status);
+        }
+
+        const text = await response.text();
+        throw new Error(text || status);
+      })
+
```
tomhughes: Get rid of this blank line please - it needs to be clear that there's another operation chained after this call.

SykoDaedalus:
> Perhaps we should always read the response text if it isn't OK and pass it together with the status and status text down to the `.catch` section?

I was trying to do that before changing tests. But I guess reading the response if itsn't ok and converting to json will again cause the wierd parsing error? Or maybe I am understanding it wrong 🤷‍♂️

SykoDaedalus:
> > Perhaps we should always read the response text if it isn't OK and pass it together with the status and status text down to the `.catch` section?
> 
> I was trying to do that before changing tests. But I guess reading the response if itsn't ok and converting to json will again cause the wierd parsing error? Or maybe I am understanding it wrong 🤷‍♂️

Maybe I understand what you are trying yo say, but I have some confusion. Like will overpass send "meaningful" data if request is not ok and if yes then probably we will have to make a separate render function to display it which we can also use it in the `then` block?

hlfan:
I was thinking of something like

```js
        throw new Error("Response not OK", { cause: {
          status: response.status,
          statusText: response.statusText,
          text: await response.text()
        } });
```

and handle that in the `catch` block in the vicinity of this code

https://github.com/openstreetmap/openstreetmap-website/blob/70c2269f5b9c7112998d6443c038591f23763425/app/assets/javascripts/index/query.js#L243

SykoDaedalus:
> I was thinking of something like
> 
> ```js
>         throw new Error("Response not OK", { cause: {
>           status: response.status,
>           statusText: response.statusText,
>           text: await response.text()
>         } });
> ```
> 
> and handle that in the `catch` block in the vicinity of this code
> 
> https://github.com/openstreetmap/openstreetmap-website/blob/70c2269f5b9c7112998d6443c038591f23763425/app/assets/javascripts/index/query.js#L243

Respone text is something like - `<p><strong style="color:#FF0000">Error</strong>: runtime error: open64: 0 Success /osm3s_osm_base Dispatcher_Client::request_read_and_idx::timeout. The server is probably too busy to handle your request. </p>`
While there is useful information in the response, it's wrapped in HTML and would need parsing to extract. `statusText` already provides a reasonable error message, I feel the added complexity of HTML parsing might not be worth it? However, I can implement it if it can help in debugging the code.

hlfan:
Or just drop the whole waiting for the response text for _bad requests_ too and make the whole function synchronous?

app/assets/javascripts/index/query.js
```diff
@@ -171,7 +171,15 @@ OSM.Query = function (map) {
...
     })
-      .then(response => response.json())
+      .then(response => {
+        if (response.ok) {
+          return response.json();
+        }
+
+        const status = response.statusText || response.status;
+
+        throw new Error(status);
```
hlfan: I don't think there's a need for making an extra var for that then.

```suggestion
        throw new Error(response.statusText || response.status);
```

tomhughes (approved):
This looks good to me now, thanks for the contribution.

PR #6725, opened 2026-01-18
Fix month label index desync causing missing labels

hlfan:
Since JS trims the heatmap data to align with the calendar-dependent start of the week, the point where monotonic month indices snap back a year can differ between calendars.

This misalignment desynchronizes month label indices and can cause month labels to be skipped.

The fix makes the label handling logic agnostic to yearly offsets.

### DIFF
--- a/app/assets/javascripts/heatmap.js
+++ b/app/assets/javascripts/heatmap.js
@@ -3,6 +3,7 @@ $(document).on("turbo:frame-load", function () {
   const weekInfo = getWeekInfo();
   const maxPerDay = heatmap.data("max-per-day");
   const weekdayLabels = heatmap.find("[data-weekday]");
+  const monthLabelStartIndex = Math.min(...heatmap.find("[data-month]").get().map(l => l.dataset.month));
   let weekColumn = 1;
   let previousMonth = null;
 
@@ -24,7 +25,7 @@ $(document).on("turbo:frame-load", function () {
       weekColumn++;
       const currentMonth = getMonthOfThisWeek(date);
       if (previousMonth === null) {
-        previousMonth = currentMonth;
+        previousMonth = currentMonth + (Math.round((monthLabelStartIndex - currentMonth) / 12) * 12);
         heatmap.find(`[data-month]:has( ~ [data-month="${previousMonth}"])`).remove();
         heatmap.find("[data-month]").first().css("grid-column-start", 2);
       }


app/assets/javascripts/heatmap.js
```diff
@@ -22,8 +23,9 @@ $(document).on("turbo:frame-load", function () {
     const date = new Date($day.data("date"));
     if (date.getUTCDay() === weekInfo.firstDay % 7) {
       weekColumn++;
-      const currentMonth = getMonthOfThisWeek(date);
+      let currentMonth = getMonthOfThisWeek(date);
       if (previousMonth === null) {
+        currentMonth += Math.round((monthLabelStartIndex - getMonthOfThisWeek(date)) / 12) * 12;
```
tomhughes: Can we not use `currentMonth` here rather than calling the function a second time?
```suggestion
        currentMonth += Math.round((monthLabelStartIndex - currentMonth) / 12) * 12;
```

hlfan: Yes. Also, this could be part of the `previousMonth` initialization all together.

hlfan:
These before/after images are from applying the JS patch to the current https://master.apis.dev.openstreetmap.org/user/OrganicMapsTestUser asset.

app/assets/javascripts/heatmap.js
```diff
@@ -24,7 +25,7 @@ $(document).on("turbo:frame-load", function () {
       weekColumn++;
       const currentMonth = getMonthOfThisWeek(date);
       if (previousMonth === null) {
-        previousMonth = currentMonth;
+        previousMonth = currentMonth + (Math.round((monthLabelStartIndex - currentMonth) / 12) * 12);
         heatmap.find(`[data-month]:has( ~ [data-month="${previousMonth}"])`).remove();
```
pablobm, line 29: Incidentally: what does this line do?

tomhughes:
I didn't completely figure things out last night but I think this is the root cause of the problem:

 https://github.com/openstreetmap/openstreetmap-website/blob/5b3210c5f637596ed65e75c2173990055b11fe5c/app/helpers/heatmap_helper.rb#L17

The intention is that if you start with say February then each time you get to January you add 12 so that the numbers carry on going up but in this case we're starting in January as the first value and unnecessarily add 12 so the values start from 13 as the first "month".

hlfan:
Trying a less compressed explanation:

Rails is padding the month label range here:

https://github.com/openstreetmap/openstreetmap-website/blob/5b3210c5f637596ed65e75c2173990055b11fe5c/app/helpers/heatmap_helper.rb#L12

This is to have the "outgoing" month labeled even if the date range doesn't include it.

To give the months an unique identifier also across years, every January the index is additionally incremented by 12.

However, that means that even if

1. Rails correctly starts the padded month label indices with 12 in December and
2. JS correctly deuces the month of the first full week to be January with index 1

then JS incorrectly assumed the labels to start at 1 or less.

While this logic is needed to drop labels if the previous month barely reaches into the new new month's first week, it shouldn't drop a year worth of labels due to desynced indices.

tomhughes:
As I suspected, changing the server code to this:

```ruby
    # Get unique months with repeating months and count into the next year with numbers over 12
    month_offset = 0
    months = ((from - 2.weeks).to_date..(to + 1.week).to_date)
             .map(&:month)
             .chunk_while { |before, after| before == after }
             .map(&:first)
    months = months.take(1) + months.drop(1).map do |month|
      month_offset += 12 if month == 1
      month + month_offset
    end
```

which only adds the extra offset when we see January after another month fixes the problem.

That code's not very nice though and so far I haven't come up with a cleaner way to do it...

mmd-osm:
Maybe you could use map_with_index on months, and only increase the month_offset, if the index is positive (January is not the first entry).

mmd-osm:
This might be a bit easier to read:

```ruby
    months = months.map.with_index do |month, index|
      month_offset += 12 if index.positive? && month == 1
      month + month_offset
    end
```

as minimal change to the original code:

```ruby
    month_offset = 0
    months = ((from - 2.weeks).to_date..(to + 1.week).to_date)
             .map(&:month)
             .chunk_while { |before, after| before == after }
             .map(&:first)
             .map.with_index do |month, index|
      month_offset += 12 if index.positive? && month == 1
      month + month_offset
    end
```

hlfan:
I was contemplating doing it on the server side, too, but this can also happen when December is the first month that a label is sent for, and the first labeled week column is in January, as the month of the first full week is dependent on the user's calendar. Hence, the JS side change.

tomhughes:
I'm not sure it's worth spending a lot more time debating this - the whole heatmap code is already pretty hard to understand and I'm not sure this makes it appreciably worse and it definitely fixes the bug so I think we should take it and we can always revisit things later if we want.

PR #6726, opened 2026-01-19
Use MapLibre class extensions to DRY code

hlfan:
I'd like to seize the opportunity of the API migration to restructure where and how our customisations of it reside.

### DIFF
--- a/app/assets/javascripts/maplibre/map.js
+++ b/app/assets/javascripts/maplibre/map.js
@@ -1,8 +1,12 @@
 //= require maplibre-gl/dist/maplibre-gl
-//= require maplibre.i18n
 //= require maplibre/controls
 //= require maplibre/dom_util
 //= require maplibre/styles
+
+maplibregl.Map.prototype._getUIString = function (key) {
+  const snakeCaseKey = key.replaceAll(/(?<=\w)[A-Z]/g, "_$&").toLowerCase();
+  return OSM.i18n.t(`javascripts.map.${snakeCaseKey}`);
+};
 
 OSM.MapLibre.Map = class extends maplibregl.Map {
   constructor({ allowRotation, ...options } = {}) {
@@ -34,7 +38,6 @@ OSM.MapLibre.SecondaryMap = class extends OSM.MapLibre.Map {
       container: "map",
       style: OSM.MapLibre.Styles.Mapnik,
       attributionControl: false,
-      locale: OSM.MapLibre.Locale,
       allowRotation: false,
       maxPitch: 0,
       center: OSM.home ? [OSM.home.lon, OSM.home.lat] : [0, 0],
--- a/app/assets/javascripts/leaflet.locate.js
+++ b/app/assets/javascripts/leaflet.locate.js
@@ -3,9 +3,9 @@
 L.OSM.locate = function (options) {
   const control = L.control.locate({
     strings: {
-      title: OSM.i18n.t("javascripts.map.locate.title"),
+      title: OSM.i18n.t("javascripts.map.geolocate_control.find_my_location"),
       popup: function (options) {
-        return OSM.i18n.t("javascripts.map.locate." + options.unit + "Popup", { count: options.distance });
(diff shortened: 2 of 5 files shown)


hlfan:
@CommanderStorm, I can't yet request a review in the GitHub-integrated way, so this has to do.

app/assets/javascripts/maplibre/map.js
```diff
@@ -0,0 +1,48 @@
...
+OSM.MapLibre.Map = class extends maplibregl.Map {
+  constructor({ allowRotation, ...options } = {}) {
+    const rotationOptions = {};
+    if (allowRotation === false) {
+      Object.assign(rotationOptions, {
+        rollEnabled: false,
+        dragRotate: false,
+        pitchWithRotate: false,
+        bearingSnap: 180
+      });
```
CommanderStorm, line 20: I would argue against API abstractions like this, since without any documentation how the code works this is otherwise reasonably hard to think about what is actually happening.
This rails thingy with the weird not quite working imports where one never knows what code runs where and where it comes from is already a high bar, no need to be even more weird.

Your call though, if you want your world to look like this, I am the last person to stop you.

app/assets/javascripts/maplibre/map.js
```diff
@@ -0,0 +1,48 @@
+//= require maplibre-gl/dist/maplibre-gl
+//= require maplibre/controls
+//= require maplibre/dom_util
+//= require maplibre/styles
+
+maplibregl.Map.prototype._getUIString = function (key) {
```
tomhughes, line 6: As we're already extending the map class can we not just put a replacement `_getUIString` there rather than monkey patching maplibregl? or would that not work?

CommanderStorm (approved):
If you want to abstract this away, I can adjust the rest of the code I wrote.

Do note that for the abstractions you are writing there are no docs anywhere, which will be harder for other developers without reading your code over time.
There is a cost to doing it this way, but if you are willing to pay that "weirdness" tax, then that is fine.

tomhughes:
> There is a cost to doing it this way, but if you are willing to pay that "weirdness" tax, then that is fine.

There's a cost to not doing it as well, which is having to repeat the same thing multiple times... Implementing local wrappers/extensions for an API to encapsulate how that API is used by a particular project is a pretty normal thing to do.

app/assets/javascripts/maplibre/map.js
tomhughes: > I would argue against API abstractions like this, since without any documentation how the code works this is otherwise reasonably hard to think about what is actually happening.

We can argue about the details of how it is implemented but I think the basic principle of not repeating the same code to initialise a map multiple times is good.

> This rails thingy with the weird not quite working imports where one never knows what code runs where and where it comes from is already a high bar, no need to be even more weird.

To be clear that's because we're using quite an old way of managing assets that works by textual inclusion/composition and a modern rails way of doing things would use ES6 imports and avoid the problems you're referring to.

app/assets/javascripts/maplibre/map.js
hlfan: `_getUIString` is already called in the Map constructor to set the ARIA label for the container. A replacement can't be assigned in time for that. I think it's more worth not having an edge case here than replacing it cleanly.

hlfan:
> I think it's probably a bit silly moving `CombinedControlGroup` to `maplibre/dom_util` in one commit and then moving it again to `maplibre/controls` in the next one - why not just move it directly to the final location in the first commit?

I wanted to get reviews in early; I also felt it was silly when committing this but figured I'd clean it up together if other restructures were suggested.

app/assets/javascripts/maplibre/map.js
hlfan: This was more caused by wanting to get rid of the `disableRotation` calls. Abstracting the options away is more a "do it with the same sweep" kinda thing.

hlfan:
> I'm not super keen on the `dom_utils` name

Me neither, but I expect that we'll find a better term as the API migration continues.

PR #6727, opened 2026-01-20
Use media queries to hide menus instead of JavaScript

### LINKED ISSUE #6523: A screamer consisting of a banner and collapsed panel elements on mobile devices
### URL

https://www.openstreetmap.org

Open the OSM website

Firefox 147 / Chrome 142

<table>

<tr>

<td>

https://github.com/user-attachments/assets/6522c80f-b557-4f43-8610-6966fe36a420

<td>

https://github.com/user-attachments/assets/3192691d-2ae7-4c0a-9f68-1b41097f6264

</table>

deevroman:
fixes #6523

### Description

Currently, on mobile devices, you can see opened menu items and banners within a few milliseconds.

https://github.com/user-attachments/assets/6522c80f-b557-4f43-8610-6966fe36a420

This is because the style that hides them is applied using JavaScript by setting the `small-nav` class to `<body>`. In the code, it looks like:

https://github.com/openstreetmap/openstreetmap-website/blob/5b3210c5f637596ed65e75c2173990055b11fe5c/app/assets/javascripts/application.js#L115-L119

This can be replaced with a media query that will be applied without waiting for JavaScript to be applied.

However, now you need to fill in the `secondaryMenuItems` array only if the screen size is mobile. Otherwise, the width of the menu items will always remain 0. This is important if you had a narrow window when loading the site, but then you expanded it.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -112,10 +112,14 @@ $(function () {
     const windowWidth = $(window).width();
 
     if (windowWidth < breakpointWidth) {
-      $("body").addClass("small-nav");
       expandAllSecondaryMenuItems();
     } else {
-      $("body").removeClass("small-nav");
+      if (secondaryMenuItems.length === 0) {
+        $expandedSecondaryMenu.find("li:not(#compact-secondary-nav)").each(function () {
+          secondaryMenuItems.push([this, $(this).width()]);
+        });
+        moreItemWidth = $("#compact-secondary-nav").width();
+      }
       const availableWidth = $expandedSecondaryMenu.width();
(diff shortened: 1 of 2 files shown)


deevroman:
I see here is a potential problem that something might happen with SotM banners. The welcome banner for beginners works, but I can't check whether others will work.

tomhughes:
Banners seem to be working fine as far as I can tell and I don't see any other problems so I think this is good, thanks.

There is actually a similar problem with intermediate widths where some items need to be collapsed into the secondary menu and they briefly flash up before the width calculation is done and they're moved to the secondary menu but that's an existing problem that is neither better nor worse with this fix.

PR #6728, opened 2026-01-21
Document project roles, permissions, and eligibility criteria

gravitystorm:
The aim of this is both transparency around these roles, but also some support to contributors so that they know what's involved.

Most of this comes from things I've said previously during presentations (e.g. [SotM-EU 2023 - Maintaining OpenStreetMap.org](https://www.youtube.com/watch?v=99_ecrNOMPs)) and during individual issue and PR comment threads. But I think it's worth putting these down in writing in one place.

### DIFF
--- a/doc/ROLES.md
+++ b/doc/ROLES.md
@@ -1,0 +1,36 @@
+# Project Roles
+
+Everyone can contribute to the project. Everyone can create and comment on issues and pull requests, review PRs, and generally help out.
+
+There are only two roles that grant extra permissions to contributors, and those extra permissions are quite small.
+
+## Maintainers
+
+The primary goal of the group of maintainers is to ensure the long-term success of the project.
+
+This goal can be achieved by ensuring the project meets the needs of the mapping community, maintaining code quality, making it easy to contribute, and building the team of people involved in development.
+
+The additional permissions granted allow maintainers to:
+
+* Merge pull requests. Note that anyone can review PRs, not just maintainers. It's only the final merge that is restricted.
+* Approve Github Actions runs for first-time contributors
+* View and handle security reports
+* Grant roles to contributors. Nominations can be from contributors and/or current maintainers, and are discussed and agreed upon by the existing maintainers
+* Other administrative work, including changing some project settings (like enabling CI workflows, creating new labels)
+
+To become eligible for the maintainer role, contributors should show a consistent history of quality in their pull requests, a history of helping other people when reviewing code, and a history of providing actionable guidance to other contributors.
+
+## Triagers
+
+The primary goal of the triagers is to reduce the need for maintainers to manage issues and pull requests.
+
+This goal can be achieved by keeping the issues list small, each issue on topic, and discussions focussed and productive. Anyone in the community can help with most of this work. However, triagers have extra permissions that can help in certain situations.
+
+These additional permissions allow triagers to:
+
+* Close any issue. This can include duplicate issues, off-topic issues, spam, etc
+* Add labels to any issue. This can help highlight bugs, pull requests that need changes, etc
+* Rename or edit any issue description. This can be useful when new contributors use the wrong terminology, etc
+* Edit issue relationships and create subissues
+
+To become eligible for the triager role, contributors should show a wide knowledge of both this project and the wider OSM ecosystem, along with a consistent history of helpful comments in the repo.


hlfan:
Some link to this explanation from the readme would be nice.

gravitystorm:
> Some link to this explanation from the readme would be nice.

I feel it might be a bit of a niche topic though, perhaps too much "[insider baseball](https://en.wikipedia.org/wiki/Inside_baseball_(metaphor))"? I don't expect 90% of developers to need to read any of this, and so it might be a bit of a distraction if we link it from the README.

hlfan:
I was thinking of something like this after the list:

```markdown
[what this means](ROLES.md#maintainers)
```

I certainly don't wanna imply it's necessary reading material.

ROLES.md
```diff
@@ -0,0 +1,35 @@
...
+## Triagers
+
+The primary goal of the triagers is to reduce the need for maintainers to manage issues and pull requests.
+
+This goal can be achieved by keeping the issues list small, each issue on topic, and discussions focussed and productive. Anyone in the community can help with most of this work. However, triagers have extra permissions that can help in certain situations.
+
+These additional permissions allow triagers to:
+
+* Close any issue. This can include duplicate issues, off-topic issues, spam, etc
+* Add labels to any issue. This can help highlight bugs, pull requests that need changes, etc
```
1ec5: For awareness, the triager role apparently doesn’t come with the ability to edit the labels themselves, just to apply them. I think this is fine but just wanted to head off any potential future misunderstanding.

doc/ROLES.md
```diff
@@ -0,0 +1,35 @@
...
+The primary goal of the triagers is to reduce the need for maintainers to manage issues and pull requests.
+
+This goal can be achieved by keeping the issues list small, each issue on topic, and discussions focussed and productive. Anyone in the community can help with most of this work. However, triagers have extra permissions that can help in certain situations.
+
+These additional permissions allow triagers to:
+
+* Close any issue. This can include duplicate issues, off-topic issues, spam, etc
+* Add labels to any issue. This can help highlight bugs, pull requests that need changes, etc
+* Hide any comment. This can be useful for off-topic conversations like issue assignment
+* Rename or edit any issue description. This can be useful when new contributors use the wrong terminology, etc
```
1ec5, line 33: Triagers can also edit issue relationships and create subissues, which can be helpful for making sense of the backlog.

mmd-osm:
I would add a comment who can trigger/approve a full Github actions run beyond the minimal Danger workflow. That's an issue for every newcomer.

tomhughes:
The really annoying thing is that although you only have to do it for a user's first PR you have to do it every time they change that PR as it only defaults to allowed for future PRs after you're approved them once.

ROLES.md
```diff
@@ -0,0 +1,35 @@
...
+
+The primary goal of the group of maintainers is to ensure the long-term success of the project.
+
+This goal can be achieved by ensuring the project meets the needs of the mapping community, maintaining code quality, making it easy to contribute, and building the team of people involved in development.
+
+The additional permissions granted allow maintainers to:
+
+* Merge pull requests. Note that anyone can review PRs, not just maintainers. It's only the final merge that is restricted.
+* View and handle security reports
+* Grant maintainer and triager roles to contributors. Nominations can be from contributors and/or current maintainers, and are discussed and agreed upon by the existing maintainers
```
tomhughes: Strictly speaking at the github level you need `admin` access (currently me and @gravitystorm) to do this rather than `maintain` access but I guess this is more about logical roles than how they're implemented?

ROLES.md
```diff
@@ -0,0 +1,35 @@
...
+
+The primary goal of the triagers is to reduce the need for maintainers to manage issues and pull requests.
+
+This goal can be achieved by keeping the issues list small, each issue on topic, and discussions focussed and productive. Anyone in the community can help with most of this work. However, triagers have extra permissions that can help in certain situations.
+
+These additional permissions allow triagers to:
+
+* Close any issue. This can include duplicate issues, off-topic issues, spam, etc
+* Add labels to any issue. This can help highlight bugs, pull requests that need changes, etc
+* Hide any comment. This can be useful for off-topic conversations like issue assignment
```
HolgerJeromin: No, i can not hide all comments (only mine).
Perhaps this is a bug/misconfig, 
Same with edit subject or issue content.

Edit:
image

tomhughes: You should be able to.. Per https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization you need `write` to edit or delete comments but only `triage` to hide them.

1ec5: I wonder if that’s outdated. [The documentation about this functionality](https://docs.github.com/en/communities/moderating-comments-and-conversations/managing-disruptive-comments) says it’s only for organization moderators and anyone with write access.

gravitystorm:
> I would add a comment who can trigger/approve a full Github actions run beyond the minimal Danger workflow. That's an issue for every newcomer.

That's a good idea. It's not explicit at https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization for the "first-time approval" but I strongly suspect that's only available to maintainers.

ROLES.md
gravitystorm: Eugh. I sort of just assumed this would be available for repository moderators (and I certainly didn't expect there to be any difference between "repository moderators" and "organisation moderators").

gravitystorm: It does make me wonder about the tradeoffs in giving people the somewhat-restricted "github triage" role, compared with what we'd be happy with our triagers being able to do. But maybe that's a discussion for another time, and this PR should just reflect the status-quo

ROLES.md
pablobm:
```suggestion
* Grant roles to contributors. Nominations can be from contributors and/or current maintainers, and are discussed and agreed upon by the existing maintainers
```

Nitpick: if the roles change in the future, this line would need to be updated, unless it doesn't list the roles.

ROLES.md
gravitystorm: I've updated the maintainers section, it was awkward to clarify here since "add existing labels" can be understood in two different ways.

doc/ROLES.md
gravitystorm: Added

ROLES.md
gravitystorm: Yeah, the doc here is on logical roles and the terminology that we use (i.e. "maintainer" rather than "admin" or "owner" which are github concepts)

ROLES.md
gravitystorm: I've updated this with the status quo

ROLES.md
gravitystorm: Done

gravitystorm:
Thanks for all the feedback, I've updated and rebased onto master to pick up the changes in #6458

tomhughes:

PR #6730, opened 2026-01-21
More padding for menu items on mobile devices

deevroman:
### Description

On mobile devices, it is quite inconvenient to click on menu items. Especially in the Russian localization, where as many as three lines appear. I suggest slightly increasing the line margins:

<table>
<tr>
 <td>
Before
 <td>
After
<tr>
 <td>

 <td>

<tr>
	<td>

	<td>

</table>

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -33,12 +33,12 @@
   <nav class="secondary d-flex flex-column flex-md-row gap-2 flex-grow-1 align-items-center">
     <ul id="secondary-nav-menu" class="nav flex-grow-1 justify-content-center justify-content-md-start" data-turbo-permanent>
       <% secondary_nav_items.each do |path, label, options| %>
-        <li class="nav-item">
-          <%= link_to label, path, :class => ["nav-link", current_page?(path) ? "active text-secondary-emphasis" : "text-secondary"], **(options || {}) %>
+        <li class="nav-item pb-2 pb-md-0">
+          <%= link_to label, path, :class => ["nav-link", "px-1", "py-0", current_page?(path) ? "active text-secondary-emphasis" : "text-secondary"], **(options || {}) %>
         </li>
       <% end %>
-      <li id="compact-secondary-nav" class="dropdown nav-item ms-auto collapse">
-        <button class="dropdown-toggle nav-link btn btn-outline-secondary border-0 bg-body text-secondary" type="button" data-bs-toggle="dropdown"><%= t ".more" %></button>
+      <li id="compact-secondary-nav" class="dropdown nav-item ms-auto pb-2 pb-md-0 collapse">
+        <button class="dropdown-toggle nav-link btn btn-outline-secondary px-1 py-0 border-0 bg-body text-secondary" type="button" data-bs-toggle="dropdown"><%= t ".more" %></button>
         <ul class="dropdown-menu">
         </ul>
       </li>
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -157,7 +157,7 @@ $(function () {
   function expandSecondaryMenuItem($item) {
     $item.children("a")
       .removeClass("dropdown-item")
-      .addClass("nav-link")
+      .addClass("nav-link px-1 py-0")
       .addClass(function () {
         return $(this).hasClass("active") ? "text-secondary-emphasis" : "text-secondary";
       });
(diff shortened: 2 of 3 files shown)


hlfan:
Could this be done with bootstrap?

deevroman:
Do you mean to use something like the `pb-2` class for each `<li>`? Possible, if it doesn't affect the desktop layout

hlfan:
Something like `pb-md-2` to parallel #6727?

deevroman:
> -md-

Interesting opportunity. But so far I don't understand why I have `pb-2` working but `pb-md-2` not working. (Of course I changed the width of the window)

deevroman:
Ah, I get it. `pb-md-2` is for ≥768px. But we need the opposite. 

So far, I prefer the solution via media query :)

hlfan:
Well, in combination with getting rid of this custom styling:

https://github.com/openstreetmap/openstreetmap-website/blob/bf4a8f554a4a45f3a79648f9bde2fb9a2b32d8f3/app/assets/stylesheets/common.scss#L131-L134

Perhaps set `.nav-item` to extend `px-1 pt-0 pb-2 pb-md-0`?

deevroman:
> px-1 pt-0 pb-2 pb-md-0

mmm, do you really want to see this set of weird classes on each of these `<li>`?

https://github.com/openstreetmap/openstreetmap-website/blob/d863f1628a4789401401cf4b1821c1eebb00e876/app/views/layouts/_header.html.erb#L43-L66

tomhughes:
There are plenty of other places that have multiple bootstrap classes applied to things, but maybe it would be worth adding a helper to `app/helpers/application_helper.rb` for generating those items?

There's already a helper there used to generate the class on the link element but maybe the helper should generate the whole li element and it's link?

deevroman:
While I was trying to add the classes, I realized that it was actually only adding two classes to the `<li>`, which looks better than something like:

```erb
<%= tag.li :class => header_nav_item_class do %>
  ... 
<% end %>
```

Additionally, `header_nav_link_class` can return different classes depending on the current page. `header_nav_item_class` would return a constant.

---

Since we are removing the custom style for `.nav-link`, `"px-1", "py-0"` needs to be added to `header_nav_link_class`.

---

One more note, I had to add manipulation with new classes in JS, the links in the dropdown and in the header should have different paddings

---

Finally, `.3rem` has no equivalent in Bootstrap, so the indentation between links has become slightly smaller:


https://github.com/user-attachments/assets/eed6520c-d3a4-4d59-ba4d-3af64e8abee3

app/views/layouts/_header.html.erb
```diff
@@ -33,39 +33,39 @@
...
         <%= link_to t("layouts.copyright"), copyright_path, :class => header_nav_link_class(copyright_path) %>
       </li>
-      <li class="nav-item">
+      <li class="nav-item pb-2 pb-md-0">
         <%= link_to t("layouts.help"), help_path, :class => header_nav_link_class(help_path) %>
       </li>
-      <li class="nav-item">
+      <li class="nav-item pb-2 pb-md-0">
         <%= link_to t("layouts.about"), about_path, :class => header_nav_link_class(about_path) %>
       </li>
```
hlfan: > There's already a helper there used to generate the class on the link element but maybe the helper should generate the whole li element and it's link?

I don't like the repetition either; there may be a less roundabout way to do that, though. Perhaps something similar to this:

```suggestion
      <% secondary_nav_items = [
        [history_path, t("layouts.history")],
        [export_path, t("layouts.export")],
        [traces_path, t("layouts.gps_traces")],
        [diary_entries_path, t("layouts.user_diaries")],
        [communities_path, t("layouts.communities")],
        [copyright_path, t("layouts.copyright")],
        [help_path, t("layouts.help")],
        [about_path, t("layouts.about")]
      ]

      secondary_nav_items.prepend([
        issues_path(:status => "open"),
        safe_join([t("layouts.issues"), open_issues_count], " ")
      ]) if Settings.status != "database_offline" && can?(:index, Issue)

      secondary_nav_items.each do |path, label| %>
        <li class="nav-item pb-2 pb-md-0">
          <%= link_to label, path, :class => ["nav-link", "px-1", "py-0", current_page?(path) ? "active text-secondary-emphasis" : "text-secondary"] %>
        </li>
      <% end %>
```

pablobm: Not a fan of so much code in the template, but perhaps if we move `secondary_nav_items` into a helper, it can work.

hlfan: Absolutely. Move the nav config to the helper and the styling from helper and SCSS to the template.

hlfan:
@deevroman could you please rebase this PR?

deevroman:
It is not difficult to make a rebase (did this), but it is difficult for me to make a correction to the template <b>and test it</b> in the next 2-4 weeks. 

p.s. this is one of the reasons why I don't answer in my other PRs yet, which I will definitely do as soon as I solve problems in life

Issue #6712, opened 2026-01-15
System test for language selector is flaky


pablobm:
The system test at `test/system/select_language_test.rb` fails randomly. The causes are unclear. It should be hardened so that it runs consistently.

There was an attempt to rectify this at https://github.com/openstreetmap/openstreetmap-website/pull/6589, but it didn't solve the problem.

While working on https://github.com/openstreetmap/openstreetmap-website/pull/6424, a potential clue was observed: the test failed when running on a local devcontainer, leaving the following screenshot:

<img width="815" height="284" alt="Language selection panel. The input box is empty, and the first language in the list (Afrikaans) has an outline that suggests it has the focus" src="https://github.com/user-attachments/assets/7561d296-eddc-492d-b629-bc93b2d08275" />

This suggests that the browser gives the focus to the first element in the list (Afrikaans) instead of leaving it on the input box. If you try this on your browser (open the selector, press tab to move the focus) then you won't be able to type in the box any more, which is consistent with the test failure.

As to what could possibly be moving the focus, no idea.

mmd-osm:
To generate some fresh ideas, I have thrown all of the relevant code in a chat prompt, and let it figure out why the test case is flaky. I'm posting it here without adding my own comments, so you can make up your own mind. I would encourage further commenting which of the options make sense (or are simply hallucination): https://gist.github.com/mmd-osm/61e02dee826af4ddcb28c29e621dca1e

mmd-osm:
Another option I've seen elsewhere is to run some tests slower. I've only tested it on Chrome, the snap based firefox is causing some major drama (as always). When disabling the headless mode, you can even follow along the steps. Maybe give it a try as well.

<details>

Then [this page](https://evilmartians.com/chronicles/flaky-tests-be-gone-long-lasting-relief-chronic-ci-retry-irritation) has lots of other recommendations on flaky unit tests...

By the way, in Chrome Headless, other tests tend to randomly fail as well with errors such as:

```
Failure:
ReportUserTest#test_it_promotes_issues [test/system/report_user_test.rb:46]:
`Issue.count` didn't change by 1, but by 0.
Expected: 1
  Actual: 0
```

```
Error:
AccountDeletionTest#test_the_user_is_signed_out_after_deletion:
Capybara::ElementNotFound: Unable to find link or button "Delete Account..."
    test/system/account_deletion_test.rb:28:in 'block in <class:AccountDeletionTest>'
```
</details>

[...]

PR #6731, opened 2026-01-21
Fix vector map route offset and marker sync race condition

### LINKED ISSUE #6684: Vector maps with navigation displayed break on reload (or pasting URL)
### URL

https://www.openstreetmap.org/directions?engine=fossgis_valhalla_car&route=45.47442%2C15.54199%3B45.475277%2C15.542414#map=18/45.475158/15.542361&layers=VN

1. have a "Shortbread" or "MapTiler OMT" layer opened
2. choose a navigation from one point to another (e.g. URL above, but any really)
3. note that route is correctly display
4. press F5 to reload (or copy/paste URL to another browser)
5. note that the route is offset from correct location
6. opening the layer icon fixes the display

Image

Image

happens every time to me in both Firefox and Chromium

SinghSwayam:
Fixes #6684

### The Issue
When using Vector Tile layers (e.g., Shortbread, MapTiler), reloading the page with an active route caused the route line to appear significantly offset from the map features.

### The Cause
The issue is caused by a race condition between the Sidebar CSS animation (approx. 300ms) and the WebGL Vector Tile renderer. When the sidebar opens, it resizes the map container. While Leaflet detects this, the internal MapLibre/GL instance often fails to update its viewport coordinates in time, causing the SVG route layer and the WebGL background to desynchronize.

### The Fix
This PR modifies `getRoute` to ensure strict synchronization without relying on fragile fixed timeouts:

1. **Event-Driven Synchronization:** Replaced hardcoded `setTimeout` delays with a `Promise` that listens for the `transitionend` event on the map container.
2. **Guaranteed Stability:** The route drawing logic now waits until the sidebar animation has physically completed before forcing a map resize.
3. **Vector Layer Resizing:** Iterates through layers to manually trigger `_glMap.resize()` on any vector layers to ensure the WebGL context matches the Leaflet coordinate system.

This approach ensures the map is stable before drawing, regardless of the user's machine speed or animation timing.

### Screen Recording
Before : 

https://github.com/user-attachments/assets/7c051230-4bbc-488c-b339-e48734bdb56c

After : 

https://github.com/user-attachments/assets/77667111-6ce5-4c16-961e-7c482ecae1e1

### DIFF
--- a/app/assets/javascripts/index/directions.js
+++ b/app/assets/javascripts/index/directions.js
@@ -113,39 +113,47 @@ OSM.Directions = function (map) {
     $("#directions_loader").prop("hidden", false);
     $("#directions_error").prop("hidden", true).empty();
     $("#directions_route").prop("hidden", true);
-    map.setSidebarOverlaid(false);
 
-    // Update map size after sidebar animation (approx 300ms) to snap pins to correct location
-    setTimeout(function () {
-      map.invalidateSize();
-      map.eachLayer(l => { if (l._glMap) l._glMap.resize(); });
-    }, 350);
+    const sidebarAnimation = new Promise((resolve) => {
+      const container = map.getContainer();
+
+      const onEnd = (e) => {
+        if (e.target === container && (e.propertyName === "width" || e.propertyName === "size")) {
+          container.removeEventListener("transitionend", onEnd);
+          resolve();
+        }
+      };
+
+      container.addEventListener("transitionend", onEnd);
+
+      // Failsafe for reduced motion or missed events
+      setTimeout(() => {
+        container.removeEventListener("transitionend", onEnd);
+        resolve();
+      }, 400);
+    });
+
+    map.setSidebarOverlaid(false);
 
     controller = new AbortController();
     chosenEngine.getRoute(points, controller.signal).then(async function (route) {
       await sidebarLoaded();
       $("#directions_route").prop("hidden", false);
 
+      await sidebarAnimation;
       routeOutput.remove();
-
-      // Wait for map stability to prevent route offset glitches
-      setTimeout(function () {
-        // Force Vector Layer (WebGL) resize to sync with Leaflet
-        map.eachLayer(function (layer) {
-          if (layer._glMap) {
-            layer._glMap.resize();
-          }
-        });
-
-        map.invalidateSize({ animate: false });
-
-        routeOutput.write(route);
-
-        if (fitRoute) {
-          routeOutput.fit();
+      map.eachLayer((layer) => {
+        if (layer._glMap) {
+          layer._glMap.resize();
         }
-      }, 500);
+      });
+      map.invalidateSize({ animate: false });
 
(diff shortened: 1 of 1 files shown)


tomhughes:
To echo what @gravitystorm said earlier in https://github.com/openstreetmap/openstreetmap-website/issues/6712#issuecomment-3780090265 adding fixed delays is rarely a good solution and that goes even more for the main code than it does for the tests.

I'm sure those delays have worked for you, on your machine today, for however many times you've tried it but I'm equally sure there will be circumstances where they don't work.

What is needed is a way to actually synchronise things properly rather than adding delays to try and make things work.

SinghSwayam:
@tomhughes Thanks for the feedback. I've updated the PR to replace the `setTimeout` with a transitionend event listener, so it now synchronizes perfectly with the map container's animation state. I also fixed the linting errors.

app/assets/javascripts/index/directions.js
```diff
@@ -113,11 +113,42 @@ OSM.Directions = function (map) {
...
       $("#directions_route").prop("hidden", false);
+
+      await sidebarAnimation;
+      routeOutput.remove();
+      map.eachLayer((layer) => {
+        if (layer._glMap) {
+          layer._glMap.resize();
+        }
+      });
+      map.invalidateSize({ animate: false });
```
tomhughes, line 150: So the maplibre-gl plugin for leaflet doesn't respond correctly to size invalidation I assume, hence the need to manually poke those layers? Has that been reported to maplibre-gl-leaflet upstream?

SinghSwayam: @tomhughes

**Re: Upstream Issue**
You are spot on. Ideally, the plugin should listen for `invalidateSize` and handle the context resize automatically. I saw that **[PR #76 (Fix resize issue of vector maps)](https://github.com/maplibre/maplibre-gl-leaflet/pull/76)** was merged upstream, but it clearly didn't cover this specific race condition involving CSS transitions. I haven't found an open issue tracking this specific "transition desync" scenario, so I can certainly open one to track it.

**Re: Generalization**
That sounds like a much more robust approach. Moving the `transitionend` listener into `OSM.loadSidebarContent` (or the sidebar controller) would solve this class of race condition for all sidebar interactions, not just routing.

Would you prefer I update this PR to refactor `OSM.loadSidebarContent` to fire a "sidebarReady" event that we can listen to? Or should we keep this PR focused on the routing fix and handle the architecture change separately?

matkoniecz: @SinghSwayam are you reviewing and verifying LLM content before posting it as comments or PR code?

SinghSwayam: > @SinghSwayam are you reviewing and verifying LLM content before posting it as comments or PR code?

@matkoniecz I use tools to assist with drafting my comments, but I verify the content myself. The code changes here were manually written and tested I experienced locally.

PR #6732, opened 2026-01-21
Add Zed to docs about Dev Containers

deevroman:
Zed [added](https://zed.dev/blog/dev-containers) support for Dev Containers in January of 2026. I tested the work of Dev Containers with this repository and the only remark is that Zed opens the terminal from under the `root` user, which is why you get an error:

`env: ‘ruby’: No such file or directory`

and `mise: not found`

Added information about this to Troubleshooting

### DIFF
--- a/doc/DEVCONTAINER.md
+++ b/doc/DEVCONTAINER.md
@@ -135,10 +135,13 @@ VS Code is not the only way to work with devcontainers. Other options include:
 - [DevPod](https://devpod.sh): a CLI tool to work with devcontainers.
 - [VSCodium](https://vscodium.com): a Free/Libre alternative to VS Code.
 - [GitHub Codespaces](https://github.com/codespaces): GitHub's hosted IDE.
+- [Zed](https://zed.dev): a code editor
 
 ## Troubleshooting
 
 ### `‘ruby’: No such file or directory`
+
+Make sure that you are running commands from under the `vscode` user. Use `su vscode` to switch to this user.
 
 In some cases Ruby may not install correctly. If you see this message, run these two commands:
 


tomhughes:
The whole point of dev containers was supposed to be that they provide a portable description of how to run things that can work across environments but between apparently needing an entirely different definition for codespaces and starting to fill up the documentation with a series of environment specific hacks it seems they are failing to fulfill that :disappointed:

deevroman:
It seems to me that the problem is either somewhere near this setting 

https://github.com/openstreetmap/openstreetmap-website/blob/faf250abbd9ce279d6774ec9f1359cf6e6593e75/.devcontainer/devcontainer.json#L28-L29

Either in .devcontainer/Dockerfile, where the root user is specified by default

https://github.com/openstreetmap/openstreetmap-website/blob/faf250abbd9ce279d6774ec9f1359cf6e6593e75/.devcontainer/Dockerfile#L5

However, I'm not good at Dev Containers configuration. Maybe even a feature freshness issue in Zed

---

Anyway it's a small thing compared to the fact that this is the first time I've been able to get fully green tests on a Mac with Arm, without using Rosetta:

deevroman:
Well, I was glad early. There is also a difference between dev containers in VSCode and Zed.

The documentation suggests using `bundle exec rails s` to start the web server, and it starts on `http://127.0.0.1:3000`. At the same time, in `compose.yaml` port forwarding is not configured. It is configured via `devcontainer.json` using `"forwardPorts": [3000, 5432]`. VSCode does some magic, and at the moment when you start the server insists on forwarding the port.

That is, for Zed, I had to add it to `compose.yaml`
```yaml
ports:
      - "3000:3000"
```
And start the server with the `-b '0.0.0.0'` flag

pablobm:
OK, "VSCode does some magic" might be the issue here. It could be that different IDEs are trying to be smart and do things that are neither on the common spec nor advertised by themselves. So if we slowly get things to work in different tools, it might solve things for others.

I'm trying to test with Zed in a machine where I have it installed... but now VSCode is not working there :rage: I'll come back to this in a few days when my blood pressure is lower.

pablobm:
Had another look. It appears that VSCode creates a `vscode` user behind the scenes, also making it into a sudoer. I haven't tried, but seems like we could replicate this for other IDEs by doing it explicitly in the Dockerfile: https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user#_creating-a-nonroot-user

pablobm:
@deevroman - I'm trying with Zed on macOS now. It appears to start the `rails-app` service but nothing else. Eg: no `postgres` service. Is that working for you?

deevroman:
Now Zed can't launch the project at all :(

```
2026-05-15T17:10:15+03:00 ERROR [dev_container::docker] Non-success result from docker pull: invalid reference format

2026-05-15T17:10:15+03:00 ERROR [dev_container::command_json] Error running command Command { program: "docker", args: ["inspect", "--format={{json . }}", "ghcr.io/rails/devcontainer/images/ruby:$RUBY_VERSION"], envs: {}, env_clear: false, current_dir: None, stdin_cfg: Some(Null), stdout_cfg: Some(Piped), stderr_cfg: Some(Piped), kill_on_drop: false }: Sent non-successful output; cannot deserialize. StdErr: Error: No such object: ghcr.io/rails/devcontainer/images/ruby:$RUBY_VERSION
```

I heard that they recently rewrote the implementation of dev containers, maybe they broke something. In this case, the RUBY_VERSION variable was not substituted.

upd: probably need to write a substitution in curly braces in the Dockerfile:

```
ARG RUBY_VERSION=3.4.7
FROM ghcr.io/rails/devcontainer/images/ruby:${RUBY_VERSION} 
```

---

Also note that on ARM MacBooks you won't be able to perform database migration because PostGIS doesn't support ARM: https://github.com/postgis/docker-postgis/issues/216 (In my local environment, I had to disable migrations that activate PostGIS)

deevroman:
Zed apparently learned to forward ports, but for some reason forward a port for the rails-app container and not the database ¯\\\_(ツ)_/¯ 



upd: or he hasn't learned it yet, because `-b '0.0.0.0'` still has to be specified.

upd2: `su vscode` is no longer required, Zed now logs in automatically as this user

pablobm:
> Also note that on ARM MacBooks you won't be able to perform database migration because PostGIS doesn't support ARM: postgis/docker-postgis#216 (In my local environment, I had to disable migrations that activate PostGIS)

As I understand it, the actual issue is that PostGIS don't publish an arm64 image. However I was able to make it work at https://github.com/openstreetmap/openstreetmap-website/pull/6713. See for example `.devcontainer/Dockerfile.postgres`, which uses the normal Postgres image but then installs Postgis. This works on my Apple M1 laptop.

PR #6738, opened 2026-01-23
Make share buttons use Bootstrap icons

hlfan:
Switching from custom SVGs to Bootstrap icons removes the inconsistency of logo sizes in the buttons:

Also, this quite considerably slims the codebase.

The `SHARE_BUTTONS_CONFIG` could be put in a YAML file in another PR if requested.


app/helpers/share_buttons_helper.rb
```diff
@@ -1,75 +1,37 @@
...
-      "https://www.facebook.com/sharer/sharer.php?u=#{url}&t=#{title}"
-    when :mastodon
-      "https://mastodonshare.com/?text=#{title}&url=#{url}"
-    when :telegram
-      "https://t.me/share/url?url=#{url}&text=#{title}"
-    when :bluesky
-      "https://bsky.app/intent/compose?text=#{title}+#{url}"
-    else
-      raise ArgumentError, "Unsupported platform: #{platform}"
+        link_to tag.i(:class => "bi bi-#{share[:icon] || share[:site]}", :aria => { :label => I18n.t("application.share.#{share[:site]}.alt") }), share_url, link_options
```
tomhughes: I think this would be easier to read if we kept the icon creation split from the link as it was before?

```suggestion
        link_to share_url, link_options do
          tag.i(:class => "bi bi-#{share[:icon] || share[:site]}", :aria => { :label => I18n.t("application.share.#{share[:site]}.alt") })
        end
```

hlfan: Agree, this even makes the diff nicer.

tomhughes:
Thanks - this looks good now.

PR #6744, opened 2026-01-27
Indicate content state in header

hlfan:
Mirror the state indicating styling of other buttons on the site.
@jidanni, does this resolve https://github.com/openstreetmap/iD/issues/11623?

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -14,7 +14,7 @@
     <div id="edit_tab" class="btn-group w-100">
       <%= link_to t("layouts.edit"),
                   edit_path,
-                  :class => "btn btn-outline-primary geolink editlink",
+                  :class => ["btn btn-outline-primary geolink editlink", { "active" => current_page?(edit_path) }],
                   :id => "editanchor",
                   :data => { :editor => preferred_editor } %>
       <button class='btn btn-outline-primary dropdown-toggle dropdown-toggle-split flex-grow-0' type='button' data-bs-toggle='dropdown'></button>


HolgerJeromin:
Can you add a screenshot?

HolgerJeromin:
We could change the page title from "iD | Openstreetmap" to "Edit | Openstreetmap"

hlfan:
> We could change the page title from "iD | Openstreetmap" to "Edit | Openstreetmap"

The page title is apparently not enough as mentioned in https://github.com/openstreetmap/iD/issues/11623.

The button appearance is exactly the same as when it is pressed on any page:

Screenshot_2026-01-28-09-20-41-26_40deb401b9ffe8e1df2f1cc5ba480b12

HolgerJeromin:
> The page title is apparently not enough as mentioned in openstreetmap/iD#11623.

I would say that "Edit" / "Bearbeiten" is a better title as perhaps many people do not remember the submenu entry name.
🤷‍♂️

hlfan:
We only forward the title of the editor.

https://github.com/openstreetmap/openstreetmap-website/blob/c7c0119944ff375037624cc3463402cc26b2a77a/app/assets/javascripts/id.js#L83

And I don't think there is much precedent for including an action in the page title.

tomhughes:
Is there really not a better way of doing this than having CSS that looks at what scripts are being loaded?

hlfan:
> Is there really not a better way of doing this than having CSS that looks at what scripts are being loaded?

Well yes, I just wasn't quite in rails mode when transferring that issue mentally.

tomhughes:
Thanks looks much better now, thanks!

PR #6748, opened 2026-01-28
Fix geolocate `background-color` not being properly applied and improve css maintainability

CommanderStorm:
### Description

not sure when this broke or why I did it like this before..

| before | after |
|--------|--------|
| image | image | 

### How has this been tested?

Manually

### DIFF
--- a/app/assets/stylesheets/maplibre-gl-all.scss
+++ b/app/assets/stylesheets/maplibre-gl-all.scss
@@ -35,8 +35,8 @@
         background-image: none;
       }
 
-      &.maplibregl-ctrl-geolocate-active .maplibregl-ctrl-icon {
-        color: $vibrant-green;
+      &.maplibregl-ctrl-geolocate-active {
+        background-color: $vibrant-green;
       }
     }
   }


hlfan:
I like the formatting changes, but can they please be put in another commit?

CommanderStorm:
Shure, done.

Those formatting changes were my editor, not me.

hlfan:
Care about the backgroudn typo?

CommanderStorm:
I guess..

hlfan:
> not sure when this broke or why I did it like this before

https://github.com/openstreetmap/openstreetmap-website/blob/2cda07b62e19c8c5171fc51e9790fae59a1d71f6/app/assets/stylesheets/maplibre-gl-all.scss#L21

Probably to work around this. I don't think the `important` is necessary there.

CommanderStorm:
Na, I think I remember:
Originally, the icons were done differently.
Now, they are bootstrap icons, so this also is a bit different.

We should have changed this in the DRY PR, but wer missed it, which is fair.

To match what was there before the migration to the bootstrap icons here, this should be the color actually.. 😉

CommanderStorm:
I guess I can increase the scss nesting to achive the same thing, yes.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -4,54 +4,58 @@
...
+    opacity: 0.2;
   }
 
-  &:last-child,
-  &:last-child:focus {
-    border-radius: 0 0 0 4px;
+  .maplibregl-user-location-dot,
+  .maplibregl-user-location-dot:before {
+    background-color: $vibrant-green;
   }
```
hlfan, line 62: These aren't in the control container. Hence, the circle is blue.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -4,54 +4,58 @@
...
+        border-radius: 0 0 0 4px;
+      }
+
+      .maplibregl-ctrl-icon {
+        background-image: none;
+      }
+
+      &.maplibregl-ctrl-geolocate-active .maplibregl-ctrl-icon {
+        color: $vibrant-green;
+      }
```
hlfan: To mirror the togglability of the button and to be consistent with the other leaflet sidebar buttons:

```suggestion
      &.maplibregl-ctrl-geolocate-active {
        background-color: $vibrant-green;
      }
```

hlfan:
Now with a bit tidier commit history, I think we're golden.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -4,54 +4,62 @@
...
-}
+  .maplibregl-ctrl button {
+    .maplibregl-ctrl-icon {
+      background-image: none;
+    }
 
-.maplibregl-ctrl button.maplibregl-ctrl-geolocate.maplibregl-ctrl-geolocate-active .maplibregl-ctrl-icon {
-  background-color: $vibrant-green;
+    &.maplibregl-ctrl-geolocate-active .maplibregl-ctrl-icon {
+      color: $vibrant-green;
```
hlfan, line 62: You changed this from `background-color` to `color` back to `background-color` for what reason?

CommanderStorm: The first change fixes it to how this worked before the DRY PR.
The second change changes this to work like the leaflet map.

hlfan: But before the switch to the bootstrap icons it already worked like the leaflet map.

CommanderStorm: Cannot reproduce.

In 7f77c192d26756bf29444e0c0f58c4631252efa5 which was the commit right before that switch it looked like this:

image

hlfan: So it's the fix of *that* change. I understand what you mean now.

hlfan (approved):

PR #6749, opened 2026-01-28
Fix map controls in user diary form

pablobm:
On the "New Diary Entry", the map controls are not showing. There's only a empty control group:

Illustration of the problem

I'm guessing we want this to be the same as the "Edit Location" controls in the user profile, so I have copied a solution from there.

### DIFF
--- a/app/assets/javascripts/diary_entry.js
+++ b/app/assets/javascripts/diary_entry.js
@@ -37,7 +37,8 @@ $(function () {
 
     const position = $("html").attr("dir") === "rtl" ? "top-left" : "top-right";
     const navigationControl = new OSM.MapLibre.NavigationControl();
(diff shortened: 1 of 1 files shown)


pablobm:
@CommanderStorm might be a good one to have a quick look :slightly_smiling_face:

hlfan:
Gosh darn...
I absolutely forgot we have a control without a control group.

While that does fix this, putting the icons directly in the controls instead of the control group in `maplibre/controls.js` would be where they should go.

CommanderStorm (approved):
Since the refactoring moved the icon-handling to `OSM.MapLibre.CombinedControlGroup`, I think this is a good fix.
Having the geolocator here is nice to have anyhow likely, maybe.

I have never written a diary, so might be wrong on that one.

PR #6754, opened 2026-01-29
Disable `validateStyle` in production for minor initalisation performance gains on slower hardware

CommanderStorm:
### Description

This PR pulls the `validateStyle` improvementation from the minimap.
This currently is not a noticable optimisation for the minor map views, since their style is very light.

Some context: https://github.com/maplibre/maplibre-gl-js/pull/2390
It also allows the tree-shaker (assuming you have that, I have not tested how good this works) to tree shake some of maplibres code reducing bundle size.

### How has this been tested?

Running the minor map views.

### DIFF
--- a/app/assets/javascripts/maplibre/map.js
+++ b/app/assets/javascripts/maplibre/map.js
@@ -20,6 +20,9 @@ OSM.MapLibre.Map = class extends maplibregl.Map {
       });
     }
     const map = super({
+      // Style validation only affects debug output.
+      // Style errors are usually reported to authors, who should validate the style in CI for better error messages.
+      validateStyle: false,
       ...rotationOptions,
       ...options
     });


gravitystorm:
I don't think it's a good idea to base this on the environment.

We use the `style.json` files directly from the upstream projects (e.g. [here](https://github.com/openstreetmap/openstreetmap-website/blob/cfae67036dcbeb66e5358d5cf670df915220d885/app/assets/javascripts/leaflet.maptiler.js#L8)). So they can be valid or invalid at any moment. There's nothing in our application that locks the stylesheet to a specific version, and there's nothing in our application that e.g. ensures it is valid during the tests and then that guarantees it will remain valid during production.

So the question is how should we handle the stylesheets if, at any moment, they become invalid. I don't think it's worth having different behaviour in production and development in this case. Either the app handles the invalid stylesheet cleanly, or it doesn't.

As for turning off the validation for performance reasons - I can be persuaded, but we need to describe exactly what happens in production if the upstream project deploys an invalid stylesheet. Just a blank map? Or does the whole map javascript fail?

CommanderStorm:
The effect is basically that the error in the console is less usefull.

image

For the two projects that actually give out styles (shortbread and OMT) I think we can just add this to their respective CI to never give out a broken style and be fine.
The map would error out in both cases the same way anyhow, so undesirable.

hlfan:
So is there any reason to not disable it all the time if it can easily be overridden if necessary?

CommanderStorm:
Yea, but will you or other devs remember where this tuning knob is in the future?

gravitystorm:
> Yea, but will you or other devs remember where this tuning knob is in the future?

I can see two choices here:

* One, make it an option, that developers can control, either in settings.yml or somewhere closer to the source
* Two, just disable it entirely, and either developers can edit the source code during debugging or leave debugging to the style providers.

I'm happy with either, but I would lean towards option 2. In the past, if there's been a problem with one of the styles, we tend to just report it upstream rather than investigating it ourselves.

Does that seem reasonable to everyone?

CommanderStorm:
Ok, changed as such.

PR #6756, opened 2026-01-29
Remove unused OPNVKarte TileLayer

CommanderStorm:
### Description

This seems to be dead code which was left over in https://github.com/openstreetmap/openstreetmap-website/pull/5196/

### How has this been tested?

The code is dead, so hard to test 😉


gravitystorm:
This is vendored code, so it shouldn't be changed in this project.

See https://github.com/openstreetmap/openstreetmap-website/blob/master/Vendorfile for the upstream source locations.

hlfan:
Gonna test #6760 real quick...

Issue #3611, opened 2022-07-21
Simple tag editor in the object view interface


deevroman:
### Description

In fact, this is a re-opening https://github.com/openstreetmap/trac-tickets/issues/4944

**What for?**

To fix an error in the tag, add the name of the POI does not need a complex iD-type editor.

**Interface variants:**

<details>
<summary>iD style</summary>

In essence, you need to take this editor out of iD
<img width="355" alt="Снимок экрана 2022-07-21 в 14 07 25" src="https://user-images.githubusercontent.com/25708359/180199764-eae80780-c4a5-4c48-aa07-b3e23327e33b.png">

</details>

<details>
<summary>Osm.Org Tags Editor</summary>

An extension that has essentially already implemented this idea
https://shtosm.ru/all/samy-bystry-redaktor/
![изображение](https://user-images.githubusercontent.com/25708359/180199028-621239cf-4029-4c59-a3e7-0eaa1a44f2cc.png)

</details>

**Tags are too low-level to give them so easy to edit.**

Users can still open the iD and edit the tags there. Newcomers who come to osm.org they won't see this new editor right away, so don't worry about newbie edits. The only thing is that there is some tag validation in the iD.

**More pros?**
Such a simple editor will be able to work on the phone in the browser!

**Is this adding Level 0?** https://github.com/openstreetmap/openstreetmap-website/issues/2391

No. Tag editor only.

matkoniecz:
I seem to remember someone making extension that can achieve this.

> they won't see this new editor right away, so don't worry about newbie edits

At least some people will be confused and break some stuff. This needs some handling, at least making clear that actual data is getting edited.

mmd-osm:
Data point: Firefox/Chrome report about 100 users for the extension in total so far.

deevroman:
You can also estimate to see how many edits are being made https://resultmaps.neis-one.org/osm-changesets?comment=Osm.Org%20Tags%20Editor#1/14/1

maro-21:
This would be perfect, as most of my edits are editing tags. I could save a lot of time loading the iD editor.

Issue #2391, opened 2019-10-15
Include Level0 in the edit menu


grischard:
Level0 is perfect for quick small corrections. In the edit drop-down, Level0 should be included. If that's too dangerous for newbies, maybe it can be a checkbox in the settings.

tomhughes:
Well there are two issues, firstly the fact that it's not exactly what you'd call user friendly and secondly that of baking in dependencies on third party sites.

woodpeck:
In my opinion, someone who is knowledgeable and mature enough to use Level0 without making an undiscussed mechanical edit will also be able to add themselves a Level0 bookmarklet or Greasemonkey script. Actually "looking at the map" while editing is something that avoids a ton of pitfalls, and I would really like to avoid officially suggesting that Level0 is a good way to make edits.

dieterdreist:
or an edit button, so you could add or modify tags of existing objects „without an editor“, right in the osm map sidebar/detail view. For  just-fixing-a-typo edits this would often be useful.

level0 actually does provide a map view, although it didn’t work for me on a mobile phone (edited a road in Italy and the marker was somewhere in Algeria). I just used level0 and agree it is kind of cumbersome to copy the id, prepend a w/n/r before pasting, rather than using the editor mechanism on the webpage, but I guess it won’t be disputed that level0 is generally only suitable for experienced mappers, so maybe don’t bother beginners with this choice?

mmd-osm:
@grischard : I'm curious what particular kind of correction you had in mind, and how you would have done it in Id or JOSM.

Zverik:
@dieterdreist you don't have to copy and edit an id — the full URL would work.

If we decide on somehow referencing Level0 from the website, I'd prefer it to be a like near "download XML" on node/way/relation panels.

tordans:
> Should we close this in favor of #3611?

I don't think so. They are very different things, IMO. We will have a policy on how to add new entries to the edit menu in the near future. But planning and maintaining features like an editor inside the website is something else entirely.

mmd-osm:
I thought more about the use case of doing "quick small corrections", rather than the technical details of where this is implemented (part of the website or linking to external sites). If these small corrections could be covered by a simple tag editor that is less difficult to work with for "normal" users, then our requirement may be sufficiently covered already.

Issue #1254, opened 2016-07-18
allow width change of browse/history pages


dieterdreist:
It would be nice if the width of browse and history pages (on the left) could be enlarged (by the user). On big screens you could need the available space for a better analysis and don't need a huge map.

satyammittal:
i would like to work on this

satyammittal:
On mouse pushing the region of history to right , width increases.
![screenshot from 2017-02-01 01-46-56](https://cloud.githubusercontent.com/assets/15922592/22482802/8c5c4946-e820-11e6-82b5-1a32318227c0.png)
![screenshot from 2017-02-01 01-46-52](https://cloud.githubusercontent.com/assets/15922592/22482801/8c556d6a-e820-11e6-93d8-05d9b7e0c293.png)
![screenshot from 2017-02-01 01-46-47](https://cloud.githubusercontent.com/assets/15922592/22482800/8c537a00-e820-11e6-81a9-39b9b65e7c6a.png)
![screenshot from 2017-02-01 01-46-42](https://cloud.githubusercontent.com/assets/15922592/22482799/8c52bb56-e820-11e6-9fad-89c27a2b0603.png)

1ec5:
The fixed sidebar width sometimes comes up as a frustration among mappers who get involved in long changeset discussions. On the other hand, I guess that can be an extra incentive to take a long back-and-forth conversation to a direct message or the forum.

An inline tag editor would also motivate users to resize the sidebar: https://github.com/openstreetmap/openstreetmap-website/pull/6758#discussion_r2743934490.

deevroman:
For note, better-osm-org expands the sidebar to 450px instead of 350px. Sometimes pages with objects that have few tags and whether they are short look a little empty, but mostly no one complained about this. (only the search bar becomes unusually wide)

You can also add `resize: horizontal;` for `#sidebar` and get extensible sidebar without JS

<img width="1846" height="1624" alt="Image" src="https://github.com/user-attachments/assets/39a02e2d-68ac-4e99-8414-a8f1889e1764" />

A little inconspicuous element, but it is a low-hanging fruit

1ec5:
> You can also add `resize: horizontal;` for `#sidebar` and get extensible sidebar without JS

This would work as a first iteration, though at a minimum we’d also want to persist the width across page loads and constrain the width so it doesn’t break other parts of the layout.

[...]

PR #6758, opened 2026-01-29
Tags editor

deevroman:
Implements #3611

### Description

<details>
<summary>first vesrion</summary>

https://github.com/user-attachments/assets/726ba7a1-7d48-49ce-a5f2-6fd7f1c05f94

</details>

https://github.com/user-attachments/assets/8cefff04-ed50-43a9-ac4b-cf36c50739a5

The interface now replicates https://github.com/Zverik/osmtags-editor

This is still a draft (it's fully functional, but the code could use some polishing here and there), so I invite you to test it and share your impressions.

## Do we need such an editor?

Yes. See the [issue's](https://github.com/openstreetmap/openstreetmap-website/issues/3611) likes and [popularity](https://resultmaps.neis-one.org/osm-changesets?comment=Osm.Org%20Tags%20Editor) in the osmtags-editor community.

## Are we lowering the barrier to entry too much?

We'll likely only find out about this once this feature is rolled out. Currently, the button only appears for authorized users.

I see risks in making it easier to vandalize important objects (for example, those with Wikidata), which iD only allows editing in the tag editor.

Well, I think if the situation gets out of control, we can check the `walkthrough_progress` value that the iD sets in Local Storage, or check the user's registration date on the server. I think we can find the optimal limitation.

After all, we don't know how many users click Edit on the object page. The new button is less noticeable and is located at the bottom of the page.

## Maybe make it in the form of a table?

Possibly, at least there won't be any unexpected = symbols for newbies. But I don't think this is a blocker to implementing the editor.

## created_by=?

I'm waiting for your suggestions. Now I have put "Tags editor on osm.org"

## What's wrong with iD?

He's slow and has a slow workflow for simple fixes. It's also terrible for mobile devices.

## Tests?

It seems that writing unit tests is problematic in the current repository, but it is probably possible to write E2E tests.

### DIFF
--- a/app/assets/javascripts/object_tags_editor.js
+++ b/app/assets/javascripts/object_tags_editor.js
@@ -8,23 +8,23 @@
         continue;
       }
       const eqPos = line.indexOf("=");
-      if (eqPos <= 0 || eqPos === line.length - 1) {
-        throw new Error(`= not found on ${i + 1} line`);
+      if (eqPos === -1) {
+        throw new Error(OSM.i18n.t("javascripts.object_tags_editor.errors.equals_not_found", { line: i + 1 }));
       }
       const k = line.substring(0, eqPos).trim();
       if (k === "") {
-        throw new Error(`empty key on ${i + 1} line`);
+        throw new Error(OSM.i18n.t("javascripts.object_tags_editor.errors.empty_key", { line: i + 1 }));
       }
       const v = line.substring(eqPos + 1).trim();
       if (v === "") {
-        throw new Error(`empty value on ${i + 1} line`);
+        throw new Error(OSM.i18n.t("javascripts.object_tags_editor.errors.empty_value", { line: i + 1 }));
       }
       tags.set(k, v.replaceAll("\\\\", "\n"));
     }
     return tags;
   }
 
-  function makeChangesSummary(objectType, objectId, prevTags, newTags) {
+  function validateChanges(prevTags, newTags) {
     const addedTags = new Map();
     const changedTags = new Map();
     const removedTags = new Map();
@@ -37,83 +37,13 @@
     });
     newTags.entries().forEach(([k, v]) => {
       if (!prevTags.has(k)) {
+        if (!k.match(/^[0-9a-zA-Z:_]+$/)) {
+          throw new Error(OSM.i18n.t("javascripts.object_tags_editor.errors.added_invalid_key", { key: k }));
+        }
         addedTags.set(k, v);
       }
     });
-    if (addedTags.size + changedTags.size + removedTags.size === 0) {
-      return;
-    }
-
-    function makeDetailedPrefix() {
-      let res = "";
-      if (addedTags.size) {
-        res += "Add " + addedTags.entries().map(([k, v]) => `${k}=${v}`).toArray().join(", ") + "; ";
-      }
-      if (changedTags.size) {
-        res += "Changed " + changedTags.entries().map(([k, v]) => `${k}=${v}\u200b→\u200b${newTags.get(k)}`).toArray().join(", ") + "; ";
-      }
-      if (removedTags.size) {
-        res += "Removed " + removedTags.entries().map(([k, v]) => `${k}=${v}`).toArray().join(", ");
-      }
-      return res;
-    }
-
-    function makeOnlyKeysPrefix() {
-      let text = "";
-      if (addedTags.size) {
-        text += "Add " + addedTags.map(([k]) => k).join(", ") + "; ";
-      }
-      if (changedTags.size) {
-        text += "Changed " + changedTags.map(([k]) => k).join(", ") + "; ";
-      }
-      if (removedTags.size) {
-        text += "Removed " + removedTags.map(([k]) => k).join(", ");
-      }
-      return text;
-    }
-
-    function makeUniversalSummary() {
-      return `Update tags of ${objectType}/${objectId}`;
-    }
-
-    function makeMainTagsHint() {
-      const mainTags = new Set([
-        "shop", "building", "amenity", "man_made", "highway", "natural",
-        "aeroway", "historic", "railway", "tourism", "landuse", "leisure"
-      ]);
-      let mainTagsHint = "";
-
-      for (const [k, v] of prevTags.entries()) {
-        if (mainTags.has(k) && !removedTags.has(k) && !changedTags.has(k)) {
-          mainTagsHint += ` ${k}=${v}`;
-          break;
-        }
-      }
-      for (const [k, v] of prevTags.entries()) {
-        if (k === "name" && !removedTags.has("name") && !changedTags.has("name")) {
-          mainTagsHint += ` ${k}=${v}`;
-          break;
-        }
-      }
-      return mainTagsHint;
-    }
-
-    let summary = makeDetailedPrefix();
-    const mainTagsHint = makeMainTagsHint();
-    if (summary.length > 200 || changedTags.size > 1) {
-      summary = makeOnlyKeysPrefix();
-    }
-    summary = summary.replace(/; $/, "");
-    if (mainTagsHint === "") {
-      summary += ` for ${objectType}/${objectId}`;
-    } else if (removedTags.size) {
-      summary += " from" + mainTagsHint;
-    } else if (changedTags.size) {
-      summary += " of" + mainTagsHint;
-    } else if (addedTags.size) {
-      summary += " to" + mainTagsHint;
-    }
-    return summary && summary.length < 255 ? summary : makeUniversalSummary();
+    return addedTags.size + changedTags.size + removedTags.size !== 0;
   }
 
   function makeAuthHeaders() {
@@ -141,7 +71,7 @@
       body: new XMLSerializer().serializeToString(changesetPayload)
     });
     if (!res.ok) {
-      throw new Error("Failed changeset creation");
+      throw new Error(OSM.i18n.t("javascripts.object_tags_editor.errors.failed_changeset_creation"));
     }
     return await res.text();
   }
@@ -156,7 +86,7 @@
         body: objectInfoStr
       });
       if (!res.ok) {
-        throw new Error(`Error when uploading changes: HTTP ${res.status}`);
+        throw new Error(OSM.i18n.t("javascripts.object_tags_editor.errors.upload_failed", { status: res.status }));
       }
     } finally {
       await fetch(`/api/0.6/changeset/${changesetId}/close`, {
@@ -208,6 +138,13 @@
 
     const $errorBox = $("<p>");
 
+    const $commentInput = $("<input>")
+      .prop({
+        type: "text",
+        placeholder: OSM.i18n.t("javascripts.object_tags_editor.changeset_comment_placeholder")
+      })
+      .addClass("form-control mb-2");
+
     const $editorTextarea = $("<textarea>")
       .addClass("form-control font-monospace mb-3")
       .prop({
@@ -225,11 +162,10 @@
 
     async function submitTags() {
       const newTags = parseTags($editorTextarea.val());
-      const summary = makeChangesSummary(type, id, currentTags, newTags);
-      if (!summary) {
-        throw OSM.i18n.t("javascripts.object_tags_editor.no_changes");
+      if (!validateChanges(currentTags, newTags)) {
+        throw new Error(OSM.i18n.t("javascripts.object_tags_editor.errors.no_changes"));
       }
-      const changesetId = await openOsmChangeset(summary);
+      const changesetId = await openOsmChangeset($commentInput.val().trim());
       applyNewTags(objectInfo, newTags);
       await uploadNewVersion(type, id, objectInfo, changesetId);
     }
@@ -242,7 +178,7 @@
           await submitTags();
           location.reload();
         } catch (e) {
-          $errorBox.text(e);
+          $errorBox.text(e.message ?? String(e));
         }
       });
 
@@ -250,9 +186,12 @@
       .prop({
         type: "submit",
         name: "save",
-        value: OSM.i18n.t("javascripts.object_tags_editor.save")
+        value: OSM.i18n.t("javascripts.object_tags_editor.save"),
+        disabled: true
       })
       .addClass("btn btn-primary");
+
+    $commentInput.on("input", () => $saveButton.prop("disabled", $commentInput.val().trim() === ""));
 
     const $cancelButton = $("<input>")
       .prop({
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/object_tags_editor.js
```diff
@@ -0,0 +1,278 @@
...
+    summary = summary.replace(/; $/, "");
+    if (mainTagsHint === "") {
+      summary += ` for ${objectType}/${objectId}`;
+    } else if (removedTags.size) {
+      summary += " from" + mainTagsHint;
+    } else if (changedTags.size) {
+      summary += " of" + mainTagsHint;
+    } else if (addedTags.size) {
+      summary += " to" + mainTagsHint;
+    }
```
1ec5: This Lego string-building would need to be replaced by a series of format strings, one for each scenario.

Could we add an input box to at least give the user an option to supply [a more descriptive changeset comment](https://wiki.openstreetmap.org/wiki/Good_changeset_comments)? It could be prefilled with something formulaic for someone who’s really in a hurry.

app/assets/javascripts/object_tags_editor.js
```diff
@@ -0,0 +1,278 @@
...
+
+    const [, type, id] = location.pathname.match(/\/(node|way|relation)\/([0-9]+)/);
+    const objectInfo = await downloadObjectInfo(type, id);
+    const currentTags = extractTagsFromObjectInfo(objectInfo);
+
+    const $browseSection = $("#sidebar_content h2 + div").first();
+
+    const $errorBox = $("<p>");
+
+    const $editorTextarea = $("<textarea>")
```
1ec5, line 148: The big question: should this be a simple text area or something more structured? If it’s a simple text area that depends on a format of some sort, we need to provide hard-to-miss access to documentation about the format. It isn’t quite [Level0L](https://wiki.openstreetmap.org/wiki/Level0L), it’s the [multiline key-value syntax](https://wiki.openstreetmap.org/wiki/Tags#key=value_syntax) that iD exposes in its text view.

Most of the mappers you know already know the syntax by heart or will it figure on their own, but this PR would put “Edit Tags” in front of a broader user base, including users who’ve never edited the map before. Maybe this text area could be a steppingstone to something more structured like iD’s default raw tags view, a table of editable fields. After all, unlike Level0 (#2391), this inline text editor doesn’t really need to serve a mail-merge-like use case where you copy-paste the result of some text processing on the command line.

app/views/elements/show.html.erb
```diff
@@ -7,6 +7,10 @@
 <% if @feature.visible? %>
   <div class='secondary-actions mb-3'>
     <%= link_to(t("browse.download_xml"), :controller => "api/#{@type.pluralize}", :action => :show) %>
+    <% if current_user %>
+      &middot;
+      <%= tag.a t("browse.edit_object_tags"), :href => "", :class => "edit_object_tags p-0 border-0 bg-transparent shadow-none" %>
```
1ec5, line 12: From my perspective as an iD developer, I’d expect some volume of well-meaning but misguided contributions from people who don’t speak English or OSM English, as well as even messier contributions from SEOs. Think whole sentences as keys, that sort of thing.

If we’re concerned about this possibility, we could make inline editing an option alongside the [Preferred Editor](https://www.openstreetmap.org/preferences/basic) preference. It would still be more discoverable than the current browser extension, but just high enough of a hurdle that we’d have a better idea of who’s using the feature initially.

app/assets/javascripts/object_tags_editor.js
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    function makeUniversalSummary() {
+      return `Update tags of ${objectType}/${objectId}`;
+    }
+
+    function makeMainTagsHint() {
+      const mainTags = new Set([
+        "shop", "building", "amenity", "man_made", "highway", "natural",
+        "aeroway", "historic", "railway", "tourism", "landuse", "leisure"
```
1ec5: Can we derive this set from [browse_icons.yml](https://github.com/openstreetmap/openstreetmap-website/blob/b8c6ab2b3d3bafd2c6314c4d6d37f99fd9c91219/config/browse_icons.yml), which also tracks primary feature tags?

app/views/elements/show.html.erb
deevroman: > Think whole sentences as keys, that sort of thing.

If we look at keys specifically, we can implement strict validation and prohibit spaces in keys. Or even prohibit adding anything but `0-9a-zA-Z:_` to a new keys.

app/assets/javascripts/object_tags_editor.js
deevroman: I don't have any good ideas yet. Either we need some kind of templating for the JS file. Or we need to fetch it from the user's browser and parse the yml file. Or we need to move comment generation to the server.

upd: or opt out of generating comments

app/assets/javascripts/object_tags_editor.js
deevroman: A text field definitely has advantages over something like a table (it's not for nothing that iD has two editing modes)

We might need something like a Help button with a tooltip

app/views/elements/show.html.erb
deevroman: Moreover, it is not difficult to implement autocomplete and validation based on Taginfo, but we can think about this later.

app/assets/javascripts/object_tags_editor.js
deevroman: > Could we add an input box to at least give the user an option to supply a more descriptive changeset comment? 

While the extra screen might be annoying for me as an advanced user, it's a really useful step for beginners editing. Perhaps we could even skip pre-filling the comment to avoid confusing non-English speakers. 

I'll experiment with this workflow

1ec5: It doesn’t have to be a separate screen. It can go right above the Save button, similar to Level0 or the basic wikitext editor in MediaWiki.

app/assets/javascripts/object_tags_editor.js
1ec5: > it's not for nothing that iD has two editing modes

Right, the main motivation was to enable mappers to copy-paste tags between elements: openstreetmap/iD#6302. A JOSM-like UI for that would’ve been inconvenient because the sidebar doesn’t have a custom context menu. As a side benefit, the single text area makes it easier to copy-paste tag combinations from the wiki or nsi.guide.

I wouldn’t say the single text area is a showstopper, but it would be pretty explicitly an advanced thing that we wouldn’t want newer users to think of as the main tool for editing tags. It’s essentially a code editor (which I guess implies a future enhancement for syntax highlighting).

deevroman: Maybe the motivation was really like that, but in my practice there are still a few advantages:

1. A wider field. There are no additional buttons, and the keys column does not take up a fixed width
2. Works arrow navigation on the keyboard

> future enhancement for syntax highlighting

well, so far there is no CodeMirror even in the iD :)

1ec5: > A wider field. There are no additional buttons, and the keys column does not take up a fixed width

Yeah, the sidebar width can be constraining since it isn’t resizable: #1254.

> Works arrow navigation on the keyboard

Personally I am worried that, the third time I use this feature, I’m going to press the adjacent <kbd>↑</kbd> and <kbd>⇧</kbd> buttons on my laptop keyboard and eject a tag by accident. But I probably should be more worried about what’ll happen when I use this editor on my phone, since you mentioned that as a motivating factor. MediaWiki presents mobile users with an extra mandatory preview due to all the possible accidents that can occur, like accidentally tapping on the Save button while trying to select text. Maybe that confirmation screen isn’t such a bad idea after all.

deevroman: > Yeah, the sidebar width can be constraining since it isn’t resizable

In fact, it's high time we increased the width of the sidebar. And you can make it dynamic with a single line in CSS. But this is a separate discussion.

app/assets/javascripts/object_tags_editor.js
tomhughes: > This Lego string-building would need to be replaced by a series of format strings, one for each scenario.

Actually it should probably be done properly in rails views using turbo - we're trying to get away from doing this sort of thing on the client side and we certainly don't want to be adding more.

app/assets/javascripts/object_tags_editor.js
tomhughes: > The big question: should this be a simple text area or something more structured? If it’s a simple text area that depends on a format of some sort, we need to provide hard-to-miss access to documentation about the format. It isn’t quite [Level0L](https://wiki.openstreetmap.org/wiki/Level0L), it’s the [multiline key-value syntax](https://wiki.openstreetmap.org/wiki/Tags#key=value_syntax) that iD exposes in its text view.

Even if we do offer a simple text area (and I suspect we shouldn't) then it definitely shouldn't be the default.

tomhughes:
> Well, I think if the situation gets out of control, we can check the `walkthrough_progress` value that the iD sets in Local Storage, or check the user's registration date on the server. I think we can find the optimal limitation.

I'm not sure what that would have to do with anything? I don't think we can condition features on the main web site on user's having used iD.

deevroman:
> I don't think we can condition features on the main web site on user's having used iD.

It could be logic like this: we show the button, either if the user is more than a week old, or they were trained in iD. Yes, there is an undesirable connection between components, and I hope we don't need it. But as an option, if it turns out to be something like MAPS.ME 2.0, we can think about it.

deevroman:
Okay, I threw it away automatic comment generation. Now everything is much simpler:


https://github.com/user-attachments/assets/8cefff04-ed50-43a9-ac4b-cf36c50739a5

Also added key validation for allowed characters and error translations
p.s. I won't forget to squash commits

HolgerJeromin:
> It could be logic like this: we show the button, either if the user is more than a week old

I would go for: needs contribution of more than x edits.

pablobm:
I agree with @tomhughes (https://github.com/openstreetmap/openstreetmap-website/pull/6758#discussion_r2744036740) that this should be moved to Rails views as much as possible. The code that puts together the content using jQuery can be implemented with a template. Not sure about [Turbo](https://turbo.hotwired.dev/) purely because I don't know what the current situation is and how complex it would be to use for sidebar content (is there any prior art?), but at the very least integrated with [the current JS router](https://github.com/openstreetmap/openstreetmap-website/blob/e73834f7756d618a4f6eee2378cedb1042addcd6/app/assets/javascripts/index.js#L292) (see for example [how the sidebar is loaded for map elements](https://github.com/openstreetmap/openstreetmap-website/blob/e73834f7756d618a4f6eee2378cedb1042addcd6/app/assets/javascripts/index/element.js#L13)).

There's also the validation, which here is happening fully on the client side. Ideally this should happen in the backend... but from what I'm seeing there's no validation applied to tags anywhere at the moment? So I don't know what the expectations are here.

simonpoole:
- How does this handle tag changes with geometry impact/ties?  
- How does it handle relation memberships (changing a tag might make the element in question an invalid member)?

More general observation: it has never been as easy to edit OSM as right now, however that development has not had any noticeable impact on sign-ups, editor activity and retention. There is no reason to believe that adding this facility will have any other effect that to cannibalise edits from existing tools with (currently, see below) better safeguards and user support.

And while I realise that everybody reading this already knows it: in the long run you are essentially adding yet another fully fledged editor with all the work and issues that implies. I'm sure that everybody has the good intention of limiting the scope, but irl that just isn't a thing.

simonpoole:
> There's also the validation, which here is happening fully on the client side. Ideally this should happen in the backend... but from what I'm seeing there's no validation applied to tags anywhere at the moment? So I don't know what the expectations are here.

Maybe https://youtu.be/3wg0P32QhO4?si=tUww0jdExqbWsk6Y is of interest.

pablobm:
Thank you @simonpoole, that was illuminating.

So any editor, including this one, is responsible for validation. The API won't do it as it can't be an arbiter of the evolving conventions and schemas in the community. This leads to complications even in cases like this one, as certain tags can have complex effects.

Now, as for this PR, I see the following:
1. Explicit validation of the format of any new keys introduced in the object.
	- Is this established enough a convention that could be implemented in the backend?
	- In particular I wonder if any local communities maintain tags whose keys have non-latin characters. I'm no expert in any of the many tools that could answer this, so no idea.
2. A check that the submission has any changes at all.
	- Makes sense to keep in the frontend.
3. A lack of validation for unique keys.
	- Currently this code parses tags in order, stuffing them in a `Map` as they come. This means that the last tag will win in case of duplication.
	- Makes sense to implement in the frontend.

1ec5:
> In particular I wonder if any local communities maintain tags whose keys have non-latin characters. I'm no expert in any of the many tools that could answer this, so no idea.

I’m unaware of any tagging convention that explicitly calls for non-Latin characters in keys, but some tagging schemes technically allow relatively freeform names as subkeys, such as `payment:*=*` and `ref:*=*`. [This report](https://taginfo.openstreetmap.org/reports/characters_in_keys#letters) shows that non-ASCII keys are relatively rare, the most common being [`agglomération=*`](https://taginfo.openstreetmap.org/keys/agglom%C3%A9ration) (8,814 occurrences) and [`railway:position:exact:京通`](https://taginfo.openstreetmap.org/keys/railway%3Aposition%3Aexact%3A%E4%BA%AC%E9%80%9A) (120).

Formalizing a character set requirement would make the API more opinionated about tagging schemes than it usually is. Mappers using QA tools have historically been responsible for cleaning up exotic characters in keys. It’s quite easy to introduce them intentionally using other mainstream editors. The only difference here is that the text box is so accessible that a cat on the keyboard could plausibly mess up a feature directly on the homepage.

tomhughes:
At the end of the day even iD and JOSM will let you enter any tags you want - they will try and guide you to certain ways of tagging but they won't stop you entering raw values.

Other editors are probably do even less in the way of validation.

So you certainly could try and extend this to do some validation, but it would probably make this PR at least ten times bigger at a guess, and we'd basically be getting into the job of writing and maintaining a partial editor as part of the site which we've avoided until now.

simonpoole:
> At the end of the day even iD and JOSM will let you enter any tags you want - they will try and guide you to certain ways of tagging but they won't stop you entering raw values.
> 
> Other editors are probably do even less in the way of validation.

Both iD and JOSM do present warnings for problematic things, and more general nearly all editors employ a combination of not allowing an operation in the first place and after the fact validation that will display a warning in some form.

> 
> So you certainly could try and extend this to do some validation, but it would probably make this PR at least ten times bigger at a guess, and we'd basically be getting into the job of writing and maintaining a partial editor as part of the site which we've avoided until now.

Just wait for the first requests for adding a "simple button to reverse a way" and  "I need to change this maxspeed tag, but just for part of the way, you already have a list of all the nodes displayed so you just need to add a button to split there" :-).

PR #6760, opened 2026-01-29
Add a danger warning for changes to vendored files

tomhughes:
This adds a danger warning to any PR which modifies files in the vendor tree.


gravitystorm:
I'm happy with this. Obviously it will flag up any legitimate changes to these files (e.g. when the upstream code changes are being updated into our vendored copy) but in those circumstances they can be ignored.

PR #6765, opened 2026-02-01
Add datalist to display multiple search results

### LINKED ISSUE #6368: Let the user select search results from the origin and destination fields
Currently, the directions panel’s origin and destination fields allow the user to input a Nominatim query, but pressing Enter simply resolves to the top result; the user never gets a chance to choose any other results. A typical directions interface would pop open a minimalistic list of search results (such as what the browser would provide with a [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist)) for the user to choose from. The [Valhalla](https://valhalla.openstreetmap.de/) and [Openrouteservice](https://maps.openrouteservice.org/) demo applications already have this functionality. Since Nominatim doesn’t provide autocompletion out of the box, this list would only show up when pressing Enter if there are multiple results.

SykoDaedalus:
### Description
Fixes #6368 

This PR adds a suggestion menu similar to Valhalla to give users options to choose from.
- Increase query limit from 1 to 5 to show more options
- Populate datalist with search results for user selection

Screenshot 2026-02-01 004347

It uses `<datalist>` that are not the best in appearance but is getting the job done. Problem with `<ul>`  and `<div>` was that options menu was getting clipped off by directions-output sidebar.

### How has this been tested?
- Manually tested in the browser by entering locations with multiple matches and selecting options  from the suggestion list
- Linting test

### DIFF
--- a/app/assets/javascripts/index/directions-endpoint.js
+++ b/app/assets/javascripts/index/directions-endpoint.js
@@ -1,5 +1,6 @@
 OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, changeCallback) {
   const endpoint = {};
+  const datalist = $("#" + input.attr("list"));
 
   endpoint.marker = L.marker([0, 0], {
     icon: OSM.getMarker(marker),
@@ -37,19 +38,33 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
   }
 
   function inputChangeListener(e) {
-    // make text the same in both text boxes
     const value = e.target.value;
-    endpoint.setValue(value);
+    const selectedOption = datalist
+      .find("option")
+      .get()
+      .find(option => option.value === value);
+
+    if (selectedOption) {
+      const lat = selectedOption.getAttribute("data-lat");
+      const lon = selectedOption.getAttribute("data-lon");
+      endpoint.setValue(value, { lat, lon });
+    } else {
+      endpoint.setValue(value);
+    }
   }
 
-  endpoint.setValue = function (value) {
+  endpoint.setValue = function (value, latlngObj) {
     if (endpoint.geocodeRequest) endpoint.geocodeRequest.abort();
     delete endpoint.geocodeRequest;
     input.removeClass("is-invalid");
 
-    const coordinatesMatch = value.match(/^\s*([+-]?\d+(?:\.\d*)?)(?:\s+|\s*[/,]\s*)([+-]?\d+(?:\.\d*)?)\s*$/);
-    const latlng = coordinatesMatch && L.latLng(coordinatesMatch[1], coordinatesMatch[2]);
-
+    let latlng;
+    if (latlngObj) {
+      latlng = L.latLng(latlngObj.lat, latlngObj.lon);
+    } else {
+      const coordinatesMatch = value.match(/^\s*([+-]?\d+(?:\.\d*)?)(?:\s+|\s*[/,]\s*)([+-]?\d+(?:\.\d*)?)\s*$/);
+      latlng = coordinatesMatch && L.latLng(coordinatesMatch[1], coordinatesMatch[2]);
+    }
     if (latlng && endpoint.cachedReverseGeocode && endpoint.cachedReverseGeocode.latlng.equals(latlng)) {
       setLatLng(latlng);
       if (endpoint.cachedReverseGeocode.notFound) {
@@ -97,7 +112,7 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
 
   function getGeocode() {
     const viewbox = map.getBounds().toBBoxString(), // <sw lon>,<sw lat>,<ne lon>,<ne lat>
-          geocodeUrl = OSM.NOMINATIM_URL + "search?" + new URLSearchParams({ q: endpoint.value, format: "json", viewbox, limit: 1 });
+          geocodeUrl = OSM.NOMINATIM_URL + "search?" + new URLSearchParams({ q: endpoint.value, format: "json", viewbox, limit: 5 });
 
     endpoint.geocodeRequest = new AbortController();
     fetch(geocodeUrl, { signal: endpoint.geocodeRequest.signal })
@@ -114,13 +129,20 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
         return;
       }
 
-      setLatLng(L.latLng(json[0]));
-
-      endpoint.value = json[0].display_name;
-      input.val(json[0].display_name);
-
-      changeCallback();
+      populateDatalist(json);
     }
+  }
+
+  function populateDatalist(results) {
+    datalist.empty();
+    results.forEach(result => {
+      const option = $("<option></option>")
+        .val(result.display_name)
+        .attr("data-lat", result.lat)
+        .attr("data-lon", result.lon);
+
+      datalist.append(option);
+    });
   }
 
   function getReverseGeocode() {
--- a/app/views/layouts/_search.html.erb
+++ b/app/views/layouts/_search.html.erb
@@ -53,7 +53,8 @@
                 </svg>
               </span>
             </div>
-            <%= text_field_tag "route_from", params[:from], :placeholder => t("site.search.from"), :autocomplete => "on", :class => "form-control py-1 px-2 ps-4", :dir => "auto" %>
+            <%= text_field_tag "route_from", params[:from], :placeholder => t("site.search.from"), :autocomplete => "off", :class => "form-control py-1 px-2 ps-4", :dir => "auto", :list => "from_suggestions" %>
+            <datalist id="from_suggestions"></datalist>
           </div>
           <div class="d-flex align-items-center">
             <div class="routing_marker_column position-absolute">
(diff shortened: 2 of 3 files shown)


SykoDaedalus:
Valhalla has better styling for their options menu. I was facing difficulty to achieve something similar because the menu was getting clipped off and I was unable to discover its cause.
@1ec5 I have raise the draft PR because I would like to make styling better but I can't figure out whats causing the clipping. I will look more into it

app/assets/javascripts/index/directions-endpoint.js
```diff
@@ -37,11 +37,36 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
...
+      input.val(value);
+      changeCallback();
+      return;
+    }
+
     endpoint.setValue(value);
   }
 
+  function findDatalistOption(value) {
+    if (!datalist || !datalist.length) return null;
```
tomhughes: As we always provide a datalist I don't think this is required? The value should never be null and if the length is zero then we just won't match in the loop?

app/assets/javascripts/index/directions-endpoint.js
```diff
@@ -84,6 +109,10 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
     delete endpoint.value;
     input.val("");
     map.removeLayer(endpoint.marker);
+
+    if (datalist && datalist.length) {
```
tomhughes: Same goes here - datalist should always be set and calling empty on an empty list should be safe?

app/assets/javascripts/index/directions-endpoint.js
```diff
@@ -37,11 +37,36 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
...
+    const selectedOption = findDatalistOption(value);
+
+    if (selectedOption) {
+      const lat = selectedOption.getAttribute("data-lat");
+      const lon = selectedOption.getAttribute("data-lon");
+
+      setLatLng(L.latLng(lat, lon));
+      endpoint.value = value;
+      input.val(value);
+      changeCallback();
```
tomhughes: This is duplicating code in `setValue` so I think we need to share that - could we not just use `setValue` here if we passed in the lat/lon object? Obviously it would then only look for lat/lon in the text if that argument wasn't passed...

app/assets/javascripts/index/directions-endpoint.js
```diff
@@ -37,19 +37,40 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
...
 
-  endpoint.setValue = function (value) {
+  function findDatalistOption(value) {
+    const options = datalist[0].querySelectorAll("option");
+    for (const option of options) {
+      if (option.value === value) {
+        return option;
+      }
+    }
+    return null;
```
pablobm: Unnecessary line.

app/assets/javascripts/index/directions.js
```diff
@@ -24,8 +24,8 @@ OSM.Directions = function (map) {
   };
 
   const endpoints = [
-    OSM.DirectionsEndpoint(map, $("input[name='route_from']"), { icon: "start", color: "var(--marker-green)" }, endpointDragCallback, endpointChangeCallback),
-    OSM.DirectionsEndpoint(map, $("input[name='route_to']"), { icon: "destination", color: "var(--marker-red)" }, endpointDragCallback, endpointChangeCallback)
+    OSM.DirectionsEndpoint(map, $("input[name='route_from']"), { icon: "start", color: "var(--marker-green)" }, endpointDragCallback, endpointChangeCallback, $("datalist[id='from_suggestions']")),
+    OSM.DirectionsEndpoint(map, $("input[name='route_to']"), { icon: "destination", color: "var(--marker-red)" }, endpointDragCallback, endpointChangeCallback, $("datalist[id='to_suggestions']"))
```
pablobm: Since the `datalist` is referenced from the input via the `list` attribute, I think it's to to use that from within `OSM.DirectionsEndpoint` instead of passing it explicitly here.

app/views/layouts/_search.html.erb
```diff
@@ -63,7 +64,8 @@
                 </svg>
               </span>
             </div>
-            <%= text_field_tag "route_to", params[:to], :placeholder => t("site.search.to"), :autocomplete => "on", :class => "form-control py-1 px-2 ps-4", :dir => "auto" %>
+            <%= text_field_tag "route_to", params[:to], :placeholder => t("site.search.to"), :autocomplete => "off", :class => "form-control py-1 px-2 ps-4", :dir => "auto", :list => "to_suggestions" %>
+          <datalist id="to_suggestions"></datalist>
```
pablobm: The indent is slightly off here:


```suggestion
            <datalist id="to_suggestions"></datalist>
```

app/assets/javascripts/index/directions-endpoint.js
pablobm: Unnecessary line.

app/assets/javascripts/index/directions-endpoint.js
```diff
@@ -37,19 +37,40 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
...
     const value = e.target.value;
+    const selectedOption = findDatalistOption(value);
+
+    if (selectedOption) {
+      const lat = selectedOption.getAttribute("data-lat");
+      const lon = selectedOption.getAttribute("data-lon");
+      endpoint.setValue(value, { lat, lon });
+      return;
+    }
     endpoint.setValue(value);
```
pablobm: In my opinion an explicit `else` would read better:

```suggestion
    } else {
      endpoint.setValue(value);
    }
```

app/assets/javascripts/index/directions-endpoint.js
hlfan: Or if we're already using jquery:
```javascript
return datalist
  .find("option")
  .get()
  .find(option => option.value === value)
```

`selectedOption` could also be set without that function in this case.

SykoDaedalus:
I was able to remove the clipping but now directions-sidebar is broken. I think I will stick to datalist 😅

https://github.com/user-attachments/assets/062f1361-6fa4-4da0-b6b7-ebd69c548666

tomhughes:
Does this actually work for you then @pablobm? Because it doesn't work for me in firefox or chrome... The callback is populating the datalist but it never appears so I never get to choose a value.

tomhughes:
Ah the list does come if you wait long enough, or hit enter, but if you tab to the next field (as I'm used to doing) then it doesn't appear and also doesn't complete the value in the way it used to.

lonvia:
> Ah the list does come if you wait long enough,

I certainly hope there isn't a form of search-as-you-type against Nominatim in this PR.

tomhughes:
I don't believe it's changed when geocode requests are sent, only how many results are requested.

That request, as now, is made on a `change` event which should only fire when the element loses focus (eg by tabbing to a new element) or is committed by the user (eg by pressing enter) and trying again this morning that seems to be true.

I think what confused me last night was that if you go back to the field and type the same thing again it still has the old data list so as soon as you type a couple of matching characters the old list reappears so you think it did a request because the typing but it didn't really.

SykoDaedalus:
@pablobm This is my first time writing tests, so I just gave it a try based on how input tag was tested. Please take a look if its correct.

app/assets/javascripts/index/directions-endpoint.js
```diff
@@ -37,19 +38,33 @@ OSM.DirectionsEndpoint = function Endpoint(map, input, marker, dragCallback, cha
...
+      .get()
+      .find(option => option.value === value);
+
+    if (selectedOption) {
+      const lat = selectedOption.getAttribute("data-lat");
+      const lon = selectedOption.getAttribute("data-lon");
+      endpoint.setValue(value, { lat, lon });
+    } else {
+      endpoint.setValue(value);
+    }
```
hlfan, line 53: Since `.data()` returns `undefined` if there's no match, this can be shortened.

```suggestion
    const latlngObj = datalist
      .find("option")
      .filter((i, option) => option.value === value)
      .first()
      .data();
    endpoint.setValue(value, latlngObj);
```

test/system/directions_test.rb
```diff
@@ -65,6 +65,28 @@ class DirectionsSystemTest < ApplicationSystemTestCase
...
     end
   end
+  test "selects from datalist and shows route" do
+    visit directions_path
+    stub_nominatim_search
+    stub_straight_routing(:start_instruction => "Head north")
+
+    find_by_id("route_from").set("Berlin").send_keys :enter
+
+    find_by_id("route_from").set("Berlin, Germany").send_keys :enter
```
hlfan, line 75: Shouldn't you be able to select that with arrow keys + enter?

pablobm:
There's a very confusing behaviour in Firefox: the datalist items are filtered using the text input as a strict substring.

So for example, if I search for "london texas", several results arrive from Nominatim:
- In Firefox, no results are displayed. This is because none of the results has "london texas" as an exact substring.
- In Chrome I do see results. Looks like Chrome takes the strings "london" and "texas" separately, allowing any result that includes both strings in any way. So for example "London, Rusk County, Texas" appears.

This means that the form won't work on Firefox in these cases, at all. The current behaviour takes the user to the first result, but with this new behaviour no result is used and the user is not allowed to pick one. I don't think that we can detect this behaviour, short of detecting the browser which would be a bad idea.

SykoDaedalus:
Maybe the only way to make it work without a 'hack' is to create a dropdown with ul. It might take some time so I'm closing this PR. 
The current method does not work for firefox(all other chromium based browsers-brave, edge, etc are working)

PR #6770, opened 2026-02-03
Suppress install prompt on non-SPA-like pages

### LINKED ISSUE #6406: Avoid suggesting to add OSM to home screen in the login flow
### Problem

When using OSM as a third party OAuth popup, OSM suggest to add itself to the home screen on Android devices.

### Description

It would be great to avoid this suggestion in the OAuth login flow in order to avoid confusing the users of the original app/website.
Israel hiking map, in our case.

- I believe it is related to the following feature if I needed to guess: #2361

CC: @1ec5 

### Screenshots

Image

hlfan:
Fixes #6406 by denying the prompts on all pages not using `router.js`. This includes pages of the sign up and log in flow, copyright, about, trace, diary, user and other lesser used pages (https://github.com/openstreetmap/openstreetmap-website/pull/1558#issuecomment-306334341).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -97,6 +97,10 @@ window.updateLinks = function (loc, zoom, layers, object) {
     .toggleClass("disabled", editDisabled);
 };
 
+window.addEventListener("beforeinstallprompt", e => {
+  if (!OSM.Router) e.preventDefault();
+});
+
 $(function () {
   // NB: Turns Turbo Drive off by default. Turbo Drive must be opt-in on a per-link and per-form basis
   // See https://turbo.hotwired.dev/reference/drive#turbo.session.drive


tomhughes (changes requested):
The lint failure needs fixing...

hlfan:
Maybe I should look into apps to rebase from on the go...

gravitystorm:
Whatever the problem is, I don't think this is the right solution.

OSM.Router is an implementation detail - it's just there to allow page navigation without full page reloads. These pages aren't special or fundamentally different from the pages that don't use it. So it can't be used as "these pages should do something special on Android and these ones shouldn't".

And in any case, it's a piece of functionality that we're replacing with Turbo Drive, and every page eventually will end up using turbo drive for page navigation, even clicking between the About and Help pages, for example.

If there are pages that should have this "install prompt" suppressed, then we need some other logic to decide which ones these should be.

PR #6771, opened 2026-02-03
Fix unclosed HTML elements

hlfan:
Trying out [HERB](https://github.com/marcoroth/herb), I found some unclosed tags. I closed these; for the self-closing `meta` and `input` elements, I followed ERBLint.


tomhughes:
Looks good to me, thanks.

PR #6772, opened 2026-02-03
Generate maplibre styles from `layers.yml`

hlfan:
I gave moving the style definitions to the maplayers lib a shot.


hlfan:
@pablobm given MapLayers.rb doesn't have any test at the moment, how should the layer definitions be tested at all?

pablobm:
In a case like this, I would take the opportunity to write some light tests for `MapLayers`. Just enough to have minimal guardrails and encourage writing more tests in the future, as well as refactor safely.

lib/map_layers.rb
```diff
@@ -20,4 +31,45 @@ def self.embed_definitions(layers_filename)
...
       .to_h { |entry| [entry["layerId"], entry.slice("leafletOsmId", "leafletOsmDarkId", "apikey").compact] }
   end
+
+  def self.style_from_style_url(layer, key)
+    layer[key].sub("{apikey}", layer["apikey"] || "")
+  end
+
+  def self.style_from_tile_url(layer, key)
+    url_template = layer[key]
+                   .sub("{r}", "{ratio}")
```
tomhughes: Why not just use `{ratio}` in the template? You already changed `{z}` to `{zoom}` to avoid having to rename it but then did the same thing again with `{r}`?

app/assets/javascripts/leaflet.map.js
```diff
@@ -11,25 +11,21 @@ L.OSM.Map = L.Map.extend({
...
-        layerOptions.leafletOsmId = leafletOsmId;
-      } else {
-        layerOptions.leafletOsmId = "TileLayer";
-      }
-
-      const layerConstructor = L.OSM[layerOptions.leafletOsmId];
+      const layerConstructor =
+        (OSM.isDarkMap() && L.OSM[leafletOsmDarkId]) ||
+        L.OSM[leafletOsmId] ||
+        L.OSM.TileLayer;
```
tomhughes: The back and forth on this is getting a bit silly... IIRC this was originally written as if/then/else then you changed it to an &&/|| chain then the maplibre conversion changed it back to if/then/else and now you're changing it back again...

Personally I have to say that as clever as your version might be I find the other one easier to read and reason about.

config/layers.yml
```diff
@@ -6,7 +6,7 @@
   canEmbed: true
   canDownloadImage: true
   maxZoom: 19
-  tileUrl: "https://tile.openstreetmap.org/{zoom}/{x}/{y}.png"
+  tileUrl: "https://tile.openstreetmap.org/{z}/{x}/{y}.png"
```
CommanderStorm, line 9: To explain:
the ID scheme uses `{zoom}`, which is why I did it this way.
I don't care if it is `{zoom}` or `{z}`.

app/assets/javascripts/leaflet.map.js
CommanderStorm: In the maplibre change, I changed this since I NEEEDED leafletOsmId to be set to the exact style that is chosen on the main map (i.e. for the dark mode style, the dark mode).

Since this is now moved upstream, the previous version is fine again.

config/layers.yml
hlfan: iD also accepts `{z}`, I'm unaware of other editors using the ELI that don't support both.

app/assets/javascripts/leaflet.map.js
hlfan: I've pondered preemptively ditching the L.OSM layers for a single source of truth. What do you think of that?

CommanderStorm: If it is possible with resonable effort, that would be another step into the direction where "we" want to go.

Removing this is on my TODO list for the second-to-last MapLibre migration PR, but if we can pull this out this again makes the diff smaller.

lib/map_layers.rb
hlfan: It's probably better to replace `{ratio}` with `{r}` only in the leaflet layer options for now so we don't need need to change the YAML file again in the migration.

lib/map_layers.rb
```diff
@@ -4,20 +4,75 @@ module MapLayers
...
+            end
+
+    {
+      :version => 8,
+      :sources => {
+        "raster-tiles-#{layer['layerId']}" => {
+          :type => "raster",
+          :tiles => tiles,
+          :tileSize => 256,
+          :maxZoom => layer["maxZoom"]
```
hlfan: @CommanderStorm which one is the canonical property Maplibre expects?

```suggestion
          :maxzoom => layer["maxZoom"]
```

CommanderStorm: `maxzoom` => https://maplibre.org/maplibre-style-spec/sources/#maxzoom_1

test/lib/map_layers_test.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class MapLayersTest < ActiveSupport::TestCase
+  def test_full_definitions_returns_well_formed_layers_collection
+    layers = MapLayers.full_definitions("config/layers.yml")
+
+    assert_kind_of Array, layers, "Expected full_definitions to return an array"
```
pablobm, line 9: I would add a check that there are some layers present at all. Otherwise, if there are no layers (count == 0) the `layers.each` loop below will appear to pass when it's actually not testing anything.

```suggestion
    assert_kind_of Array, layers, "Expected full_definitions to return an array"
    assert_operator layers.count, :>, 0, "Expected some layers available"
```

The next test does test this when it does `layers.first["nameId"]`, but it's indirect and I'd prefer it explicit.

PR #6773, opened 2026-02-04
Remove CI guard on minitest/focus require

gravitystorm:
There is now a [Minitest/Focus cop](https://docs.rubocop.org/rubocop-minitest/cops_minitest.html#minitestfocus) in rubocop-minitest, which will catch any inadvertent usage of focus in a commit.

### How has this been tested?

I changed one test from `def test_show` to `focus def test_show` and it was flagged by rubocop.


tomhughes:
Looks good to me, thanks.

PR #6780, opened 2026-02-07
Expose the enhanced changeset stats via the API

### LINKED ISSUE #6719: Make the enhanced changeset stats available via the API
### Problem

The Image statistics that were added in #5758 are not available via the API at the moment.

### Description

I propose adding this attribute to all APIs that return changesets (`/changeset/{id}`, `/changesets`, etc.)

Happy to send a PR if this idea is accepted

### Screenshots

_No response_

k-yle:
Closes #6719

### Description

Adds 3 new properties to APIs that return changesets, for the enhanced changeset stats.

### How has this been tested?

unit tests and [localhost:3000](http://localhost:3000) using 
- `/api/0.6/changeset/3`
- `/api/0.6/changeset/3.json`
- `/api/0.6/changesets`
- `/api/0.6/changesets.json`

### DIFF
--- a/app/views/api/changesets/_changeset.json.jbuilder
+++ b/app/views/api/changesets/_changeset.json.jbuilder
@@ -6,6 +6,9 @@ json.created_at changeset.created_at.xmlschema
 json.open changeset.open?
 json.comments_count changeset.comments.length
 json.changes_count changeset.num_changes
+json.created_count changeset.num_created_elements
+json.modified_count changeset.num_modified_elements
+json.deleted_count changeset.num_deleted_elements
 
 json.closed_at changeset.closed_at.xmlschema unless changeset.open?
 if changeset.bbox.complete?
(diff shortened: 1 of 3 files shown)


mmd-osm:
Why are you exposing 9 attributes while the UI exposes only 3? What is the use case to add such fine granular attributes?

I would offer the same data as the UI only, for consistency reasons.

k-yle:
agreed - my only hesitation with adding 3 derived properties, is that someone might ask for the 9 original properties in the future - and then we'll end up with 12, to avoid a breaking change 😆

but if this concern is irrelevant, then i'll change it to the 3 derived properties

hlfan:
I think in that case the metadata would only be fetched supplementary to the osmChange, which would then have more than enough info.

k-yle:
cool, changed to just the 3 properties

tomhughes:

mmd-osm:
@k-yle maybe you could update https://wiki.openstreetmap.org/wiki/API_v0.6 as well…

k-yle:
sure, [done](https://wiki.openstreetmap.org/w/index.php?title=API_v0.6&diff=prev&oldid=2957643). feel free to change anything obviously

Issue #6240, opened 2025-07-24
JS crash when WebGL is disabled


danielrainer:
### How to reproduce the issue?

The new vector tile layers (Shortbread, MapTiler OMT) use WebGL. WebGL is not available on all browsers (e.g. disabled for privacy). The code does not seem to be able to handle this and crashes with

```text
Uncaught Error: {"requestedAttributes":{"antialias":false,"preserveDrawingBuffer":false,"powerPreference":"high-performance","failIfMajorPerformanceCaveat":false,"desynchronized":false,"alpha":true,"depth":true,"stencil":true,"premultipliedAlpha":true},"statusMessage":"WebGL is currently disabled.","type":"webglcontextcreationerror","message":"Failed to initialize WebGL"}
```

This happens as soon as the "Layers" button is clicked, since that tries to render a bit of the map using each style as a preview. The previews of the Shortbread and MapTiler OMT layers just appear gray, and so does the main map if either of them is selected. The main map stays gray even when switching back to a raster layer, which needs a page reload to fix.

Tested on Librewolf 140 (with WebGL disabled, the default for this browser).

### Screenshot(s) or anything else?

Shortbread and MapTiler OMT in gray:
<img width="3598" height="1794" alt="Image" src="https://github.com/user-attachments/assets/e5f387b8-98e8-4379-a7ce-d3061ac01330" />

Whole map gray if Shortbread is selected:
<img width="3590" height="1794" alt="Image" src="https://github.com/user-attachments/assets/f609ae63-98f7-4b58-a496-4877484c89b2" />

Map stays gray even when switching back to the Standard layer:
<img width="3602" height="1786" alt="Image" src="https://github.com/user-attachments/assets/73c9ee82-abf4-42fd-b5cd-c016a3126599" />

ProgrammerWhoPrograms:
I seem to be having this problem as well. Also a Librewolf user (141.0.3-1).
With WebGl disabled map is blank. Shows tiles otherwise.

jtn20:
I've run into a more severe symptom, with (I suspect) the same root cause: in a browser with WebGL disabled, after switching to one of the WebGL-based layers, the map on the openstreetmap.org website can end up entirely unusable (even after reloading the page or restarting the browser), until I manually delete a cookie; there's no way to switch back to a tile-based layer.

[...]

PR #6782, opened 2026-02-08
Add Human facing error messages when users without WebGL use 

CommanderStorm:
### Description

There is no good way to fallback when a user without webgl comes, but we can add a better error message.

Basicaly fixes (except on the main map) https://github.com/openstreetmap/openstreetmap-website/issues/6240

### How has this been tested?

Using Firefox, in `about:config` set `webgl.disabled=true`.

| desktop | square | mobile | minimap |
|--------|--------|--------|-----|
| image | image | image | image |

### DIFF
--- a/app/assets/javascripts/maplibre/map.js
+++ b/app/assets/javascripts/maplibre/map.js
@@ -6,6 +6,17 @@ maplibregl.Map.prototype._getUIString = function (key) {
   return OSM.i18n.t(`javascripts.map.${snakeCaseKey}`);
 };
 
+OSM.MapLibre.showWebGLError = function (container) {
+  const containerElement =
+    typeof container === "string" ? document.getElementById(container) : container;
+
+  if (containerElement) {
+    fetch("/panes/webgl_error")
+      .then(response => response.text())
+      .then(html => containerElement.innerHTML = html);
+  }
+};
+
 OSM.MapLibre.Map = class extends maplibregl.Map {
   constructor({ allowRotation, ...options } = {}) {
     const rotationOptions = {};
@@ -17,13 +28,24 @@ OSM.MapLibre.Map = class extends maplibregl.Map {
         bearingSnap: 180
       });
     }
-    const map = super({
-      // Style validation only affects debug output.
-      // Style errors are usually reported to authors, who should validate the style in CI for better error messages.
-      validateStyle: false,
-      ...rotationOptions,
-      ...options
-    });
+
+    let map;
+    try {
+      map = super({
+        // Style validation only affects debug output.
+        // Style errors are usually reported to authors, who should validate the style in CI for better error messages.
+        validateStyle: false,
+        ...rotationOptions,
+        ...options
+      });
+    } catch (error) {
+      const structuredError = JSON.parse(error.message);
+      if (structuredError.type === "webglcontextcreationerror") {
+        OSM.MapLibre.showWebGLError(options.container);
+      }
+      // the constructor panicked => we need to re-throw
+      throw error;
+    }
     if (allowRotation === false) {
       map.touchZoomRotate.disableRotation();
       map.keyboard.disableRotation();
--- a/app/assets/stylesheets/maplibre-gl-all.scss
+++ b/app/assets/stylesheets/maplibre-gl-all.scss
@@ -61,3 +61,26 @@
     background-color: $vibrant-green;
   }
 }
+
+.maplibre-error {
+  container-type: size;
+
+  p {
+    max-width: 600px;
+  }
+
+  /*
+   Show compact message for small-height containers.
+   For width-based container there is bootstrap
+   */
+  @container (max-height: 60px) {
+    p:not(.maplibre-error-compact) {
+      display: none !important;
+    }
+
+    .maplibre-error-compact {
+      display: block !important;
+      font-size: 0.75rem;
+    }
+  }
+}
--- a/app/views/webgl_error_panes/show.html.erb
+++ b/app/views/webgl_error_panes/show.html.erb
@@ -1,0 +1,17 @@
+<div class="maplibre-error w-100 h-100 d-flex flex-column align-items-center justify-content-center text-center p-3 fs-5 lh-base bg-info-subtle text-info-emphasis" data-bs-theme="light">
+  <p class="maplibre-error-compact d-none position-absolute bottom-0 start-0 end-0 py-1 px-2 m-0 fw-bold">
+    <%= t(".compact_message") %>
+  </p>
+  <p class="d-none d-md-block">
+    <%= t(".description", :your_browser_does_not_support_webgl => content_tag(:b, t(".your_browser_does_not_support_webgl"))).html_safe %>
+  </p>
+  <p>
+    <%= link_to t(".read_more_about_webgl_url"), :target => "_blank", :rel => "noopener", :class => "btn btn-link text-info-emphasis fw-bold" do %>
+      <b><%= t(".webgl_is_required_for_this_map") %></b><br>
+      <small>
+        <i class="bi bi-info-circle-fill" aria-hidden="true"></i>
+        <%= t(".read_more_about_webgl") %>
+      </small>
+    <% end %>
+  </p>
+</div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2930,6 +2930,14 @@ en:
       image_dimensions_html: "Image will show the %{layer} layer at %{width} x %{height}"
       download: "Download"
       downloading: "Downloading..."
+  webgl_error_panes:
+    show:
+      compact_message: WebGL is required for this map.
+      description: "We are sorry, but it seems that %{your_browser_does_not_support_webgl}, a technology for rendering 3D graphics on the web."
+      your_browser_does_not_support_webgl: your browser does not support WebGL
+      webgl_is_required_for_this_map: WebGL is required to display this map.
+      read_more_about_webgl: "Read more"
+      read_more_about_webgl_url: https://wiki.openstreetmap.org/wiki/This_map_requires_WebGL
   traces:
     visibility:
       private: "Private (only shared as anonymous, unordered points)"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -212,6 +212,7 @@ OpenStreetMap::Application.routes.draw do
   resource :layers_pane, :path => "/panes/layers", :only => :show
   resource :legend_pane, :path => "/panes/legend", :only => :show
   resource :share_pane, :path => "/panes/share", :only => :show
+  resource :webgl_error_pane, :path => "/panes/webgl_error", :only => :show
   get "/id" => "site#id"
   resource :feature_query, :path => "query", :only => :show
   post "/user/:display_name/confirm/resend" => "confirmations#confirm_resend", :as => :user_confirm_resend
--- a/app/abilities/ability.rb
+++ b/app/abilities/ability.rb
@@ -4,7 +4,7 @@ class Ability
   include CanCan::Ability
 
   def initialize(user)
-    can :read, [:feature_query, :layers_pane, :legend_pane, :share_pane, :languages_pane]
+    can :read, [:feature_query, :layers_pane, :legend_pane, :share_pane, :languages_pane, :webgl_error_pane]
     can :read, [Node, Way, Relation, OldNode, OldWay, OldRelation]
     can :read, [RelationMember, OldRelationMember]
     can [:show, :create], Note
--- a/app/assets/javascripts/leaflet.layers.js
+++ b/app/assets/javascripts/leaflet.layers.js
@@ -27,16 +27,20 @@ L.OSM.layers = function (options) {
 
         function shown() {
           const center = map.getCenter();
-          miniMap = new OSM.MapLibre.Map({
-            container,
-            style: layer.options.style,
-            interactive: false,
-            attributionControl: false,
-            fadeDuration: 0,
-            zoomSnap: layer.options.isVectorStyle ? 0 : 1,
-            center: [center.lng, center.lat],
-            zoom: getZoomForMiniMap()
-          });
+          try {
+            miniMap = new OSM.MapLibre.Map({
+              container,
+              style: layer.options.style,
+              interactive: false,
+              attributionControl: false,
+              fadeDuration: 0,
+              zoomSnap: layer.options.isVectorStyle ? 0 : 1,
+              center: [center.lng, center.lat],
+              zoom: getZoomForMiniMap()
+            });
+          } catch (error) {
+            return;
+          }
 
           if (layer.options.layerId === "openmaptiles_osm") {
             OSM.MapLibre.setOMTMapLanguage(miniMap);
@@ -46,8 +50,11 @@ L.OSM.layers = function (options) {
         }
 
         function hide() {
-          map.off("moveend", moved);
-          miniMap.remove();
+          // miniMap can be falsy if webgl is not supported
+          if (miniMap) {
+            map.off("moveend", moved);
+            miniMap.remove();
+          }
         }
 
         function moved() {
--- a/app/controllers/webgl_error_panes_controller.rb
+++ b/app/controllers/webgl_error_panes_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class WebglErrorPanesController < ApplicationController
+  before_action :authorize_web
+  before_action :set_locale
+  authorize_resource :class => false
+
+  def show
+    render :layout => false
+  end
+end
--- a/test/controllers/webgl_error_panes_controller_test.rb
+++ b/test/controllers/webgl_error_panes_controller_test.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class WebglErrorPanesControllerTest < ActionDispatch::IntegrationTest
+  ##
+  # test all routes which lead to this controller
+  def test_routes
+    assert_routing(
+      { :path => "/panes/webgl_error", :method => :get },
+      { :controller => "webgl_error_panes", :action => "show" }
+    )
+  end
+
+  def test_show
+    get webgl_error_pane_path, :xhr => true
+
+    assert_response :success
+    assert_template "webgl_error_panes/show"
+    assert_template :layout => false
+    assert_select "div.maplibre-error"
+  end
+end


app/assets/javascripts/maplibre/map.js
```diff
@@ -7,6 +7,34 @@ maplibregl.Map.prototype._getUIString = function (key) {
...
+      : container;
+
+  if (containerElement) {
+    const errorDiv = document.createElement("div");
+    errorDiv.className = "maplibre-error";
+    errorDiv.setAttribute("data-compact-message", "WebGL is required for this map.");
+    errorDiv.innerHTML = `
+      <p>
+        We are sorry, but it seems that your browser does not support
+        <abbr title="Web Graphics Library is a JavaScript API for displaying interactive 2D and 3D graphics in web browsers without the need for plugins">WebGL</abbr>,
```
1ec5: In general, an `<abbr title>` should be just an expansion of the abbreviation and nothing else. A screen reader would have a mouthful to say instead of “WebGL”.

app/assets/javascripts/maplibre/map.js
```diff
@@ -7,6 +7,34 @@ maplibregl.Map.prototype._getUIString = function (key) {
...
+      ? document.getElementById(container)
+      : container;
+
+  if (containerElement) {
+    const errorDiv = document.createElement("div");
+    errorDiv.className = "maplibre-error";
+    errorDiv.setAttribute("data-compact-message", "WebGL is required for this map.");
+    errorDiv.innerHTML = `
+      <p>
+        We are sorry, but it seems that your browser does not support
```
1ec5: Would we even get here if the browser lacks WebGL support? In that case, I would’ve assumed it would also lack support for the [`webglcontextcreationerror`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/webglcontextcreationerror_event) event that triggers this function.

My understanding is that some browsers let the user disable WebGL, and some devices are incapable of WebGL regardless of the browser. But I don’t know if this is common enough to worry about.

CommanderStorm: Shure, adding a hint for settings is resonable.

image

If the device is broken, this is not something that is easily detectable for us or fixable for the user => I don't think that this is a good hint.. maybe you have a concrete wording in mind

1ec5: This is potentially a lot to put into an embedded iframe. Does MapLibre have a dedicated page explaining the situation and options for end users that this message can link to for more detail? If not, we could start a page on the OSM Wiki to that effect.

CommanderStorm: We don't currently sice users without working webgl is very much an edgecase of an edgecase.

I don't know what exactly I would put on such a page tbh.

app/assets/javascripts/maplibre/map.js
```diff
@@ -7,6 +7,31 @@ maplibregl.Map.prototype._getUIString = function (key) {
...
+  if (containerElement) {
+    const errorDiv = document.createElement("div");
+    errorDiv.className = "maplibre-error";
+    errorDiv.setAttribute("data-compact-message", OSM.i18n.t("javascripts.map.webgl_error.compact_message"));
+    errorDiv.innerHTML = `
+      <p class="d-none d-md-block">
+        ${OSM.i18n.t("javascripts.map.webgl_error.description", { browser_does_not_support_webgl: `<b>${OSM.i18n.t("javascripts.map.webgl_error.browser_does_not_support_webgl")}</b>` })}
+      </p>
+      <p>
+        <a class="btn btn-link" href="https://wiki.openstreetmap.org/wiki/Using_OpenStreetMap/Why_is_WebGL_Required" target="_blank">
```
CommanderStorm: @1ec5 I am assuming that OHM is fine with not making this configurable?

1ec5: Thank you for putting together this article! This is quite fine for OHM. But the URL still needs to be localizable, because the article will get translated into other languages, each at a slightly different URL.

Incidentally, I renamed the article to  https://wiki.openstreetmap.org/wiki/This_map_requires_WebGL So that the URL will be more understandable to end users.

CommanderStorm: Ok, I added this as a translation key.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -61,3 +61,46 @@
     background-color: $vibrant-green;
   }
 }
+
+.maplibre-error {
+  width: 100%;
```
hlfan: I get that container queries aren't a thing in Bootstrap, but I'd replace at least some properties with Bootstrap classes.

config/locales/en.yml
```diff
@@ -3561,6 +3561,13 @@ en:
       tooltip: "Legend"
       tooltip_disabled: "Legend not available for this layer"
     map:
+      webgl_error:
+        compact_message: WebGL is required for this map.
+        description: "We are sorry, but it seems that your %{browser_does_not_support_webgl}, a technology for rendering 3D graphics in your browser."
```
1ec5: We’re probably going to get a complaint from a translator about “your” being separated from “browser”. Let’s fold that word into the `browser_does_not_support_webgl` string too.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -61,3 +61,38 @@
     background-color: $vibrant-green;
   }
 }
+
+.maplibre-error {
+  background-color: lighten($green, 35%);
+  color: darken($green, 40%);
```
hlfan: This is basically `bg-success-subtle text-success-emphasis`, the link `text-success`.
Branding a notice as `success` is definitely a curious choice.

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -61,3 +61,38 @@
...
+   Show compact message for small containers (only height).
+   For width there are bootstrap adjustments 
+   */
+  @container (max-height: 60px) {
+    p {
+      display: none !important;
+    }
+
+    &::before {
+      content: attr(data-compact-message);
```
hlfan: Is there a reason the compact message has to be in a pseudo-element?

app/assets/stylesheets/maplibre-gl-all.scss
```diff
@@ -61,3 +61,38 @@
...
+      text-align: center;
```
hlfan: You can drop this line if you put `text-center` on `maplibre-error`.

app/assets/stylesheets/maplibre-gl-all.scss
CommanderStorm: Well.. my thinking was that for the users that will expericen this most (the few LibreWolf users) likely this is not an error, but things working as expected, so I did not make it yellow.

If you think this should be differerent I definively can change this, design in the end is very subjective

hlfan: That's a solid argument.
In that case, I'd adapt the wording to something in the vicinity of _we're unable to use WebGL here_, to emphasize more that this may be a choice of the user.

CommanderStorm: `..success..` looks very wrong (not the light green that I had but a very dark one).
`..warning..` looks much better:

image

Still a bit too alarmy, but much better than the laternative

image
image
image

hlfan: This is with `..success..` and `data-bs-theme="light"`.

app/assets/stylesheets/maplibre-gl-all.scss
CommanderStorm: it seemed the simplest way to have everything be in the css.
If bootstrap is prefered, the other way around obviously makes more sense.

hlfan (commented):
I'm still a bit unhappy with the amount of HTML in the function, but I don't have a better idea.

I also wouldn't say this closes #6240, as the leaflet wrapper doesn't use our maplibre extension.

app/assets/stylesheets/maplibre-gl-all.scss
CommanderStorm: which of these options do you prefer? warning or success?

CommanderStorm:
We can either have some CSS or some HTML..
We can make it less pretty or actionable, but that is likely not the best tradeoff

app/assets/stylesheets/maplibre-gl-all.scss
hlfan: I'd personally rate them `info` > `success` > `warning`.

CommanderStorm: I don't think the `info` fits the rest of the UI (at least it is nowere else, I think).

Maybe the light version works better, will have to see tomorrow.

hlfan:
What do you think of letting turbo fetch this as a view? Putting the markup in a .html.erb file seems more sensible than the current version.

CommanderStorm:
I have no clue what this turbofetch is. That must be something ruby-specific.
I need to look into this if you want me to do this.

Good night ^^

hlfan:
I don't necessarily see that as blocking, it just goes a bit against #5616.

app/assets/javascripts/maplibre/map.js
```diff
@@ -1,11 +1,26 @@
...
+  const containerElement =
+    typeof container === "string" ? document.getElementById(container) : container;
+
+  if (containerElement) {
+    fetch("/panes/maplibre/webgl_error")
+      .then(response => response.text())
+      .then(html => {
+        containerElement.innerHTML = html;
+      })
+      .catch(() => containerElement.innerHTML = OSM.i18n.t("javascripts.map.webgl_error.webgl_is_required_for_this_map"));
```
CommanderStorm: I could not quite get the turbofetch thingy to work correctly, but putting this behind one of these controller-vew-route-abilitys seems to work.

Honestly, not what I would call developer friendly or simpler, but if you like it better, I am fine with it.

CommanderStorm:
(testcases are not failing and just flaky)

pnorman:
> I also wouldn't say this closes #6240, as the leaflet wrapper doesn't use our maplibre extension

I wouldn't worry about the leaflet wrapper because we'll be getting rid of it

app/assets/stylesheets/maplibre-gl-all.scss
CommanderStorm: I changed my mind, the light version somehow works quite nicely

image

1ec5 (approved):
I haven’t tried this locally, but the changes seem reasonable to me. Would appreciate a review from someone more familiar with the Rails side of this.

app/assets/javascripts/maplibre/map.js
```diff
@@ -6,6 +6,18 @@ maplibregl.Map.prototype._getUIString = function (key) {
...
 
+OSM.MapLibre.showWebGLError = function (container) {
+  const containerElement =
+    typeof container === "string" ? document.getElementById(container) : container;
+
+  if (containerElement) {
+    fetch("/panes/maplibre/webgl_error")
+      .then(response => response.text())
+      .then(html => containerElement.innerHTML = html)
+      .catch(() => containerElement.innerHTML = OSM.i18n.t("javascripts.map.webgl_error.webgl_is_required_for_this_map"));
```
tomhughes: I wonder if we really need to have special handling for our own HTML page failing?

app/assets/javascripts/maplibre/map.js
tomhughes: What sort of problems did you have with turbo?

config/routes.rb
```diff
@@ -212,6 +212,7 @@
   resource :layers_pane, :path => "/panes/layers", :only => :show
   resource :legend_pane, :path => "/panes/legend", :only => :show
   resource :share_pane, :path => "/panes/share", :only => :show
+  resource :maplibre_pane, :path => "/panes/maplibre/webgl_error", :only => :show
```
tomhughes: I don't really like having this URL not map to the resource name in the same way as the others, and it strikes me that "your browser doesn't support webgl" isn't really a maplibre thing as such should this just be something like:

```suggestion
  resource :webgl_error_pane, :path => "/panes/webgl_error", :only => :show
```

or

```suggestion
  resource :nowebgl_pane, :path => "/panes/nowebgl", :only => :show
```

app/assets/javascripts/maplibre/map.js
CommanderStorm: It is more about "what if the internet goes bad in that exact point in time".

app/assets/javascripts/maplibre/map.js
CommanderStorm: I have no clue, since it just does not do anyhting.

Below does do the fetch correctly, but does not do the replacing of the `innerHTML`.
There is no error message or other information to go on and the docs on this are.. not very clear -> https://turbo.hotwired.dev/reference/frames

```ruby
Turbo.fetch("/panes/webgl_error", { frame: container });
```

CommanderStorm: regular fetch? -> Works
turbo.fetch? No clue how that is supposed to work

config/routes.rb
CommanderStorm: naming is hard

app/assets/javascripts/maplibre/map.js
hlfan: When tiles can't be loaded maplibre also doesn't show a warning?

CommanderStorm: Tiles not loading is different and I would need to look into how we handle this. Not sure.
Definitely different from this line.

Which requests can be cached is something we have quite a bit of logic around.

hlfan: I just mean the moved map looks like this when offline:



So the fallback here seems overkill compared to that.
I think for most browser map apps, the offline capability can be disregarded.

CommanderStorm: fair, removed it

app/views/webgl_error_panes/show.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="maplibre-error w-100 h-100 d-flex flex-column align-items-center justify-content-center text-center p-3 fs-5 lh-base bg-info-subtle text-info-emphasis" data-bs-theme="light">
+  <p class="maplibre-error-compact d-none position-absolute bottom-0 start-0 end-0 py-1 px-2 m-0 fw-bold">
+    <%= t("javascripts.map.webgl_error.compact_message") %>
+  </p>
+  <p class="d-none d-md-block">
+    <%= t("javascripts.map.webgl_error.description",
```
hlfan: Does this need to be namespaced with the js or can these be moved to webgl_error_pane?
```suggestion
    <%= t(".description",
```

CommanderStorm: I thought that thinks are likely simpler to translate if we pretend that this is part of the map.
This way, you know from the key where it is used.

hlfan: This can be done more clearly with the `qqq` language for "message documentation".
The key location structure is significantly less prominent for translators.

CommanderStorm: Ok, but lets do it like this instead.

Reason: `t(".` does not work (rendering as "Description") with complex keys such as `webgl_error_pane.description`, only simple keys.

No clue why and also no clue how the translation system works truely.
That seems to be coming from the ruby parts.

```suggestion
    <%= t("webgl_error_pane.description",
```

Side note: I was unable to find qqq in the codebase, must be somewhere external.

tomhughes: I believe `qqq` is just on TranslateWiki - it's dummy language where any strings you add are then displayed as advice to people translating other languages.

tomhughes (changes requested):
If you have webgl disabled and you open the layer switcher then it displays nice error messages as expected but if you then try and close the sidebar then a javascript exception is thrown as it tries to destroy the maps that were never created and the sidebar contents vanish but it never closes.

config/locales/en.yml
```diff
@@ -3763,3 +3763,10 @@ en:
     trailing_whitespace: "has trailing whitespace"
     invalid_characters: "contains invalid characters"
     url_characters: "contains special URL characters (%{characters})"
+  webgl_error_pane:
```
tomhughes: This should be `webgl_error_panes` and then you should have `show` underneath it with the resources underneath that.


```suggestion
  webgl_error_panes:
    show:
```

Then you will be able to use relative lookups in the view.

app/views/webgl_error_panes/show.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="maplibre-error w-100 h-100 d-flex flex-column align-items-center justify-content-center text-center p-3 fs-5 lh-base bg-info-subtle text-info-emphasis" data-bs-theme="light">
+  <p class="maplibre-error-compact d-none position-absolute bottom-0 start-0 end-0 py-1 px-2 m-0 fw-bold">
+    <%= t("webgl_error_pane.compact_message") %>
```
tomhughes: This, and other translation lookups in here, can just be `.compact_message` if you use the correct structure in the locale file.

tomhughes (changes requested):

CommanderStorm:
I have no clue why plural and this show key you suggested works, but it does.
Are there any resources to read up on how this system works, because it does not seem entirely understood by me.

tomhughes:
When you use a leading `.` to indicate a relative name it will construct a prefix based on the path to the view file, hence why the leading labels in the yml need to match the path.

See https://guides.rubyonrails.org/i18n.html#lazy-lookup for more.

CommanderStorm:
I think there is a bit more than is documented there.
If I do `.description` (without putting translations under `show`) this is not working, but the other keys are working...

Anyhow, offtopic.

app/assets/javascripts/maplibre/map.js
```diff
@@ -6,6 +6,17 @@ maplibregl.Map.prototype._getUIString = function (key) {
   return OSM.i18n.t(`javascripts.map.${snakeCaseKey}`);
 };
 
+OSM.MapLibre.showWebGLError = function (container) {
+  const containerElement =
+    typeof container === "string" ? document.getElementById(container) : container;
+
+  if (containerElement) {
+    fetch("/panes/webgl_error")
```
pablobm, line 14: Re: the conversation about Turbo. I'm not really super knowledgeable on the topic. However I'm not sure this is quite the use case for it. Normally I've seen it in action responding to DOM events automatically, as opposed to programmatically like this `fetch` here.

Having said that, I can make this fetch work with `Turbo.fetch` as a drop-in replacement:

```suggestion
    Turbo.fetch("/panes/webgl_error")
```

All this appears to do is adding a header `X-Turbo-Request-Id` to the request. (See [code](https://github.com/hotwired/turbo/blob/4481af6b6d7483363a44ee9f4af2e0f879c83b3c/src/http/fetch.js#L6)). No idea what this is for of whether we should be doing it in the first place.

pablobm (commented):
Something I'm seeing is that the `fetch` is issued once for every minimap. On one hand it's not a big issue because the server side knows to respond with a 304 status, but it's still additional, unnecessary requests that we could avoid by caching on our side.

app/assets/javascripts/maplibre/map.js
hlfan: I've also thought about this, but since this should also work on the embed page I'd leave it as is. Requiring Turbo there would really bloat the resource for what we get out.

hlfan, line 14: We could set `OSM.MapLibre.webGLErrorPane` to the response HTML to avoid fetching the same content multiple times.

CommanderStorm: I don't think this is actualyl an issue worth fixing.
Lets rembember how many users this actualy affects (and how much the additional 304 are an issue).. not that many.

Doing multiple reaquests was one of the choices that comes with pushing this to these ruby pages that are fetched.
I am not sure why chrome is not allowed to deduplicate the requests properly 🤷🏻 

I can either have this in ruby and then do web requests, or have this on the js side and not.

hlfan (approved):
Thanks, I think this looks good now.

CommanderStorm:
🎉 finally!

ell1e:
Perhaps it would be useful to actually have a non-WebGL render path, in the long term. Rather than just showing an error, even though that is admittedly better than a silent crash.

CommanderStorm:
I'm not really sure what a non-WebGL render path would look like in practice.
All modern browsers have webgl enabled, so it should not be an issue for users.

MapLibre GL is fundamentally built around a GPU rendering pipeline.
Without WebGL, we'd effectively need to implement and maintain a second renderer (or significantly reduce functionality), which is a very large undertaking.

For the current architecture, if we can't obtain a GPU-backed canvas, failing early with a clear error is the only sensible option.

ell1e:
Librewolf is a modern browser and has it disabled by default. Other security-focused browsers may also have it disabled or have an easy option to do so. Sadly, WebGL isn't particularly safe to leave enabled due to how e.g. on Android, GPU drivers are commonly left outdated for years.

I imagine an alternative renderer would be required, that is true. I don't want to diminish the effort required, however there might be useful libraries available for the task that could make this easier. (E.g. including some that can draw vector shapes with both a WebGL and software path.)

CommanderStorm:
If a CPU side software rendering path were a practical solution, I'd expect that such rather niche browsers such as LibreWolf to provide one transparently.
In the end, if compatability with web standards is a goal, you should adhere to them.
Since they don't, showing an error that this does not work is a good solution in my book.
You can still enable this for the website I assume, so no biggie.

ell1e:
In fact there used to be one for Chromium. I got removed because as far as I understand it ran outside of the sandbox, and it was too error-prone. The same wouldn't apply to using some library that draws vector shapes on the canvas and that is implemented in JS.

Some devices also simply don't have 3D acceleration. This is especially common in the Linux world, where there is no OS-provided easy fallback to a fake 3D renderer.

Anyway, I'm not saying this is a vital feature, but other big map sites that I'm aware of don't require WebGL. In fact, almost no site outside of games **requires** it. So OSM apparently planning to go for mandatory WebGL in the future (correct me if this is wrong) seems a bit of an odd one out, and in terms of browser security and access from more niche devices it seems suboptimal.

hlfan:
Client-side map rendering relies on some kind of hardware acceleration to make the map snappy enough.
With the spread of vector maps, the non-WebGL clients have been rejected more and more, even on commercial map services, e.g., in:

- [ArcGIS](https://www.arcgis.com/apps/mapviewer/index.html)
- [Apple Maps](https://maps.apple.com)
- [Bing Maps](https://www.bing.com/maps)
- [HERE WeGo](https://wego.here.com)
- [MapQuest](https://www.mapquest.com)
- [Stadia Maps](https://stadiamaps.com/explore-the-map)
- [TomTom](https://plan.tomtom.com)

Once there are **vector-map-capable** frameworks around that do not require WebGL, this requirement will be revisited for sure.

ell1e:
I hope I'm not seeming too pushy, it's just I realized there's even a vector example based on OSM that works fine without 3D acceleration: [CoMaps](https://codeberg.org/comaps/comaps/). Edit: removed irrelevant text.

ell1e:
Sorry, I guess I just brought this up in the wrong place. I suggested it directly to MapLibre GL now: https://github.com/maplibre/maplibre-gl-js/issues/7888

CommanderStorm:
Re: your “businesses using the embed” argument

From a product perspective, it would still be useful to look at how large the user base actually is that has WebGL disabled or unavailable. Based on what I’ve seen, this seems to be a fairly small segment (for example, LibreWolf doesn’t even appear on caniuse.com), so I’m not sure this would typically be a decision factor for businesses. It might be worth grounding this discussion in concrete numbers rather than assuming its a significant share.

> The performance argument doesn't quite hold up in my opinion

My current understanding is that implementing a CPU-based fallback with a comparable feature set and acceptable performance and bundle size would be very challenging, especially given typical constraints around maintenance and developer capacity.

If there are existing prototypes, libraries, or benchmarks that suggest a viable path here, I’d be very interested to take a look — that would likely be the most constructive next step.

More generally, I think it would help to separate what is feasible for smaller teams to maintain from what large organizations like Google can do with substantially different resources, so we can keep the discussion focused on realistic options for this project.

tomhughes:
Matomo says it's about 94% with WebGL for month to date.

ell1e:
6% is higher than I expected :exploding_head:  (I realize most normal users would have it enabled)

CommanderStorm:
Is the number without or with bots like google or chatgpt and scrapers?

PR #6783, opened 2026-02-08
Clean view markup further

hlfan:
Continuing #6771, I have furthered the use of tag helpers and removed self-closing tags for void elements

### DIFF
--- a/app/views/directions/show.html.erb
+++ b/app/views/directions/show.html.erb
@@ -4,13 +4,13 @@
   <% Rails.root.glob("app/assets/images/routing/*.svg")
           .map { |path| File.basename(path, ".svg") }
           .each do |icon| %>
-    <%= inline_svg_tag "routing/#{icon}.svg", :id => "routing-sprite-#{icon}", :to_symbol => true %>
-    <% if icon.end_with?("right") %>
-      <symbol id="routing-sprite-<%= icon.sub("right", "left") %>">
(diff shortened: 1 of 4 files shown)


tomhughes:
Looks good to me, thanks.

app/views/layouts/_markers.html.erb
```diff
@@ -24,7 +24,7 @@
        } %>
 
     <% types.filter { |type| markers[type] }.each do |type| %>
-      <path id="<%= type %>-path" <%= tag.attributes(markers[type]) %> />
+      <%= tag.path :id => "#{type}-path", :attributes => markers[type] %>
```
hlfan, line 27: I just noticed I left that in there, which breaks the markers...

tomhughes: Oops... That was one of the few bits I didn't actually check in the running site after merging. I won't deploy yet then ;-)

hlfan: You can if this looks good to you too:

```suggestion
      <%= tag.path :id => "#{type}-path", **markers[type] %>
```

I can't push 3c138206bd8be677e56bfe7a5cb4d5e44fba4230 to [git.osm.org](//git.osm.org), tho.

tomhughes: That looks like it should work, yes.

PR #6784, opened 2026-02-08
Add Herb parser and linter

hlfan:
Add additional view linting from https://github.com/marcoroth/herb and a CI workflow too.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -40,6 +40,31 @@ jobs:
         bundler-cache: true
     - name: Run erblint
       run: bundle exec erb_lint .
+  herb:
+    name: Herb
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+    - name: Check out code
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    - name: Setup ruby
+      uses: ruby/setup-ruby@09a7688d3b55cf0e976497ff046b70949eeaccfd # v1.288.0
+      with:
+        ruby-version: ${{ env.ruby }}
+        rubygems: 3.4.10
+        bundler-cache: true
+    - name: Run herb parser
+      run: bundle exec herb analyze app/ --no-log-file --non-interactive
+    - name: Cache node modules
+      uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
+      with:
+        cache: yarn
+    - name: Install node modules
+      run: bundle exec bin/yarn install
+    - name: Create dummy database configuration
+      run: cp config/example.database.yml config/database.yml
+    - name: Run herb linter
+      run: bundle exec bin/yarn run herb-lint
   eslint:
     name: ESLint
     runs-on: ubuntu-22.04
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .ruby-gemset
 .ruby-version
 .vagrant
+*erb_parsing_result*.log
 app/assets/javascripts/i18n/*.js
 config/credentials.yml.enc
 config/master.key
--- a/.herb.yml
+++ b/.herb.yml
@@ -1,0 +1,8 @@
+version: 0.8.9
+
+linter:
+  rules:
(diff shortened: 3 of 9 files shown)


.github/workflows/lint.yml
```diff
@@ -40,6 +40,31 @@ jobs:
         bundler-cache: true
     - name: Run erblint
       run: bundle exec erb_lint .
+  herb:
+    name: Herb
+    runs-on: ubuntu-latest
```
tomhughes, line 45: I was going to say why `latest` when the others use `22.04` but I see rubocop also uses `latest` so it seems we're already inconsistent...

.github/workflows/lint.yml
```diff
@@ -40,6 +40,31 @@ jobs:
...
+    timeout-minutes: 10
+    steps:
+    - name: Check out code
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    - name: Setup ruby
+      uses: ruby/setup-ruby@90be1154f987f4dc0fe0dd0feedac9e473aa4ba8 # v1.286.0
```
tomhughes: That seems to be an older version than other linters in here are using?

.github/workflows/lint.yml
```diff
@@ -40,6 +40,31 @@ jobs:
...
+    - name: Cache node modules
+      uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
+      with:
+        cache: yarn
+    - name: Install node modules
+      run: bundle exec bin/yarn install
+    - name: Create dummy database configuration
+      run: cp config/example.database.yml config/database.yml
+    - name: Run herb linter
+      run: yarn herb-lint
```
tomhughes: Maybe make this consistent with how we run installs? Also I think using `yarn run` is better than relying on yarn treating unknown commands as scripts to run...

```suggestion
      run: bundle exec bin/yarn run herb-lint
```

tomhughes (commented):
How does this overlap with what `erb_lint` is already doing?

.github/workflows/lint.yml
hlfan: Could that have been missed in #5692? That's the reason I chose `latest`.

.github/workflows/lint.yml
hlfan: Yeah, I didn't notice #6775 made that inconsistent.

hlfan:
> How does this overlap with what `erb_lint` is already doing?

In [their readme](https://github.com/marcoroth/herb?tab=readme-ov-file#prior-art--inspiration), `erb_lint` is listed as tooling Herb was built upon. I have not investigated further.

marcoroth:
Hey friends, thanks for adopting Herb into OSM! 🙏🏼 

The Herb Linter is intended to be the spiritual successor to ERB Lint. We don't have all ERB Lint rules in the Herb Linter yet, but that's the goal.

You can follow this issue for progress on this: See https://github.com/marcoroth/herb/issues/537 and https://github.com/marcoroth/herb/issues?q=is%3Aissue%20state%3Aopen%20erb%20lint%20label%3Aerb-lint

The [`better_html` gem was already deprecated in favor of `herb`](https://github.com/shopify/better-html?tab=readme-ov-file#deprecation-notice), and given the current trajectory, I could also see them deprecating `erb_lint` in favor of the Herb Linter eventually.

And yes, the Herb Linter also supports custom linter rules, check the docs for that here: https://herb-tools.dev/projects/linter#custom-rules

Please feel free to open any issues or weird behavior you might see with Herb. I'd love to make this as robust as possible so it can become the de-facto standard. Thank for trusting the project! 🙏🏼

gravitystorm:
> You can follow this issue for progress on this: See [marcoroth/herb#537](https://github.com/marcoroth/herb/issues/537) and https://github.com/marcoroth/herb/issues?q=is%3Aissue%20state%3Aopen%20erb%20lint%20label%3Aerb-lint

Thanks for the links @marcoroth - very helpful! So we'll need to keep erb_lint enabled here, at least for the short term.

There's a few things that erb_lint hasn't yet implemented that I would also like to see, so I'll open issues on `herb` when I get the chance.

PR #6787, opened 2026-02-10
Deduplicate social link icon assets

hlfan:
There's no need to store the OSM logo four times with basically no difference.

|before|after|
|-|-|
|||

This reduces the resource size of some user pages **drastically**.
For [mine](https://osm.org/user/hlfan), this **strips about 90%** of the page size (276k down to 28k).

### DIFF
--- a/test/helpers/svg_transform_test.rb
+++ b/test/helpers/svg_transform_test.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class SvgTransformsTest < ActionView::TestCase
+  def test_asset_routing_transform
+    context = Object.new
+    context.extend(InlineSvg::ActionView::Helpers)
+
+    svg = context.inline_svg_tag("social_link_icons/ohm.svg")
+    assert_no_match "../osm_logo.svg", svg
+    assert_match %r{assets/osm_logo-[0-9a-f]+.svg}, svg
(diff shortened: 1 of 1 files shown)


pablobm (commented):
Is it possible to add a test for this?

hlfan:
Possible?
Definitely.

Straightforward and without overhauling how the social links are tested?
Probably only if you accept the tests to be in ‎test/controllers/profiles_controller_test.rb‎

pablobm:
Rails doesn't make it straightforward, but I think we can reach an acceptable compromise: https://github.com/openstreetmap/openstreetmap-website/compare/master...pablobm:openstreetmap-website:deduplicated-social-link-assets-tested

hlfan:
@1ec5 as you said in https://github.com/openstreetmap/openstreetmap-website/pull/5439#discussion_r2091884846, the OHM logo could be more adaptable.

With this simplification, that logo changes the most. What's your view with these hats on (like in https://github.com/openstreetmap/openstreetmap-website/pull/6065#issuecomment-2923493824)?

1ec5:
The changes to the OHM logo are an improvement, actually. I don’t think the artifacts, like a harder shadow on the magnifying glass, were ever intentional.

1ec5 (approved):
Fancy.

Issue #6788, opened 2026-02-10
Incorrect english term for `leisure=track`


Oivo35:
There is incorrect english term for leisure=track - `track: "Running Track"`
[leisure=track](https://wiki.openstreetmap.org/wiki/Tag:leisure=track) is used also for other non-motorised sports, e.g. bobsleigh.

<img width="500" height="282" alt="Image" src="https://github.com/user-attachments/assets/728d8b84-1ee6-48d5-917f-1d22a1f82f76" />
<br>
<br>
The term in iD seems correct.
<br>
<br>
<img width="500" height="261" alt="Image" src="https://github.com/user-attachments/assets/8871e70e-c542-4071-9b34-04532275fa7d" />

tomhughes:
Please see https://github.com/openstreetmap/openstreetmap-website/blob/master/CONTRIBUTING.md#localization-i18n for advice on correcting translations.

tomhughes:
Oh sorry I missed that this was about the english version of the translation when I closed it.

Oivo35:
I'm sorry for the inaccurate title about the incorrect translation.

gravitystorm:
Fixed by #6789

PR #6789, opened 2026-02-10
Incorrect english term for `leisure=track`

Oivo35:
Changed english term for `leisure=track`
See #6788

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1161,7 +1161,7 @@ en:
           sports_centre: "Sports Centre"
           stadium: "Stadium"
           swimming_pool: "Swimming Pool"
-          track: "Running Track"
+          track: "Racetrack (Non-Motorsport)"
           water_park: "Water Park"
           "yes": "Leisure"
         lock:


tomhughes:
I'm not sure `Racetrack (Non-Motorsport)` is very nice but I'll defer to people that know more about the tagging details...

pnorman:
Most of the leisure=track around here would be more correctly described as Running Track than any form of race track.

Oivo35:
> Most of the leisure=track around here would be more correctly described as Running Track than any form of race track.

Of course, most `leisure=track` elements in OSM are running (athletics) tracks, but there are also tracks intended only for cycling, roller skating, skateboard, wakeboarding, ice skating, bobsleigh, toboggan, etc., which are tagged with `leisure=track`.
The term 'running track' is, therefore, confusing. and in addition, there is a tag [`sport=running`](https://wiki.openstreetmap.org/wiki/Tag:sport%3Drunning) that describes running as a specific discipline.

I know that the term "Racetrack (Non-Motorsport)" is not used much for this type of track, but in iD editor the term for `leisure=track` is "Racetrack (Non-Motorsport)" and in JOSM "Racetrack".

The term for highway=raceway on OSM website is "Raceway", in iD "Motorsport Racetrack" and in Josm "Raceway".

gravitystorm:
It's not an ideal translation, but I'm happy to follow the lead of iD and JOSM with this one.

@Oivo35 thanks for the PR!

SomeoneElseOSM:
> Most of the leisure=track around here would be more correctly described as Running Track than any form of race track.

FWIW In a UK/IE database I see 575 cycling, 423 equestrian, 366 running, 209 bmx, 203 horse_racing, 133 blank and 127 athletics.  "Racetrack (Non-Motorsport)" at least avoids mentioning one particular sport.

Issue #6790, opened 2026-02-10
Flooding tile server to 1/[0-1]/[0-1] to all featured raster map services


quinncnl:
### How to reproduce the issue?

Go to any page for example: https://www.openstreetmap.org/#map=12/53.1700/10.1474

Then it automatically requests to all raster tile providers. 

It happens since Feb 2. Please fix it asap.

### Screenshot(s) or anything else?

<img width="543" height="526" alt="Image" src="https://github.com/user-attachments/assets/ff5bca76-0adf-4271-9f44-a40eb06424b7" />

_No response_

tomhughes:
Are you a tile server operator that is affected?

tomhughes:
In any case I can't reproduce this... I went to the URL you gave and panned around with the debugger open and only the selected layer was being loaded.

gravitystorm:
I can reproduce this. Given that the page is loading z12, but the extra tiles are z1 and are from all available layers, I suspect it's related to the layer picker mini maps during initialisation.

tomhughes:
I see it load a few tiles initially yes, which is likely for the mini maps, but it doesn't carry on loading them as you pan around when I do it?

1ec5:
#6683 initializes the map as soon as the container element is available. I guess the difference from Leaflet is that, since GL&nbsp;JS renders to a background context, it would potentially start rendering before the element is visible. But I’m surprised that wasn’t happening with the maplibre-gl-leaflet compatibility shim already.

The requests for z1 tiles are because the maps don’t zoom and pan until the user opens the layer switcher. From code inspection, I don’t see how this code would run before that point.

https://github.com/openstreetmap/openstreetmap-website/blob/2b9ef12d372bdac871322e2b963d51c1daa865d3/app/assets/javascripts/leaflet.layers.js#L46

In principle, we could set no style during initialization and defer setting the style until the first time the layer switcher appears.

gravitystorm:
> I see it load a few tiles initially yes, which is likely for the mini maps, but it doesn't carry on loading them as you pan around when I do it?

That's what I see too. 

But I think we shouldn't load z1 tiles when they aren't going to be viewed. That's what feels most like a bug to me. It's just wasted traffic for every visitor to the front page.

quinncnl:
> Are you a tile server operator that is affected?

[...]

PR #6792, opened 2026-02-10
Initalise the MiniMaps only when the style is acually needed

CommanderStorm:
### Description

I can't percive a mesurable performance difference, most difference comes from tiles being in cache or not.
Seems to be very similar to OSM (cannot tell a difference).

### How has this been tested?

Manually, loading the map with either regular internet, 3G or caching disabled.

Resolves https://github.com/openstreetmap/openstreetmap-website/issues/6790

### DIFF
--- a/app/assets/javascripts/leaflet.layers.js
+++ b/app/assets/javascripts/leaflet.layers.js
@@ -20,45 +20,44 @@ L.OSM.layers = function (options) {
       input.checked = map.hasLayer(layer);
 
       map.whenReady(function () {
-        const miniMap = new OSM.MapLibre.Map({
-          container,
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

gravitystorm:
Thanks for the quick fix @CommanderStorm !

CommanderStorm:
I think this should be obvious, but in case it is not:

Please ping me on anything even vaguely related to maplibre. I will handle it as soon as I can.
I don't subscribe to the issue-tracker though, because that is largely not very relevant for my work.
Not sure if there is a better way to handle this.

PR #6797, opened 2026-02-11
Use double quotes for HTML atttribute values

hlfan:
A pretty straightforward linting change, done with `bundle exec bin/yarn run herb-lint --fix`.

### DIFF
--- a/.herb.yml
+++ b/.herb.yml
@@ -2,7 +2,5 @@ version: 0.8.9
 
 linter:
   rules:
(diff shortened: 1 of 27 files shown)


gravitystorm:
I'll merge this before it bit-rots, but I expect there will be some other PRs which will need a rebase. 

It's never painless, but with widespread refactoring, the quicker we get it merged, the more time we have to enjoy the long-term benefits. We had similar situations when we introduced rubocop.

Thanks!

PR #6798, opened 2026-02-11
Enable some Herb HTML linting rules

hlfan:
Some additional rules make it clearer which way elements should be created.

### DIFF
--- a/app/views/users/changeset_comments/_page.html.erb
+++ b/app/views/users/changeset_comments/_page.html.erb
@@ -14,7 +14,7 @@
     <% @comments.each do |comment| -%>
     <tr>
       <td width="25%" class="<%= "text-muted" unless comment.visible? %>"><%= link_to comment.changeset.id, changeset_path(comment.changeset) %></td>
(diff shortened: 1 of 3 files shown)


app/views/users/changeset_comments/_page.html.erb
```diff
@@ -14,7 +14,7 @@
     <% @comments.each do |comment| -%>
     <tr>
       <td width="25%" class="<%= "text-muted" unless comment.visible? %>"><%= link_to comment.changeset.id, changeset_path(comment.changeset) %></td>
-      <td width="25%" class="<%= "text-muted" unless comment.visible? %>"><span title="<%= l comment.created_at, :format => :friendly %>"><%= time_ago_in_words(comment.created_at, :scope => :"datetime.distance_in_words_ago") %></span></td>
+      <td width="25%" class="<%= "text-muted" unless comment.visible? %>"><%= friendly_date_ago comment.created_at %></td>
```
tomhughes, line 17: I mean that is still using `title` you're stopping herb from seeing it...

hlfan: I'm aware of how backwards this looks. But I'd argue that since helpers setting the title attribute are more specialized, this setting is more to spot inconsistencies of when what should be used.

While I see not enabling the rule as an option, lowering its severity may be a more suitable way.

tomhughes:
The changes both look like an improvement to me anyway, so I think we can call this good, thanks.

Issue #6791, opened 2026-02-10
Communicate Terms of Use to unauthenticated users


pablobm:
After discussion with @1ec5 and a representative of the OSMF board, I understand that the OSMF would like unauthenticated users to be considered as having accepted the Terms of Use. The "Welcome" banner can be used to communicate this, providing a link to the terms.

This is not intended as a definitive solution to the many questions around acceptance of the ToUs, but as a first step towards a larger solution.

1ec5:
The purpose of this feature request is to give notice about the ToU to logged-out users, so that the terms have more legal standing. This is an idea for chipping away at #6610, since we don’t track logged-out users’ acceptance of the ToU and wouldn’t consider such tracking to be feasible anyways. OSMF doesn’t intend to impose this specific solution on the project, but this is part of our best understanding of what the LWG asked for.

The banner in question looks like this:

<img width="350" height="307" alt="Welcome to OpenStreetMap! OpenStreetMap is a map of the world, created by people like you and free to use under an open license. Hosting is supported by Fastly, OSMF corporate members, and other partners." src="https://github.com/user-attachments/assets/cf425579-abf8-4a87-856e-db0649a5fc7e" />

Based on code inspection, I believe the welcome banner only ever appears to logged-out users. If they click the ✖️ button, we set the `_osm_welcome` cookie to `hide`, which prevents it from appearing the next time. This is sufficient for the purpose of notifying users of the ToU.

Sometimes we configure a special banner to appear to all users. Logged-out users see that banner underneath the welcome banner. openhistoricalmap.org is currently running a special banner, if you’d like to see how that looks in different situations.

Here’s where we populate the banner:

https://github.com/openstreetmap/openstreetmap-website/blob/bcded99f5c9fc4f733ed5caedce6d1b5921fc3ab/app/views/layouts/map.html.erb#L37-L41 https://github.com/openstreetmap/openstreetmap-website/blob/bcded99f5c9fc4f733ed5caedce6d1b5921fc3ab/config/locales/en.yml#L1852-L1855

We’d append the following text from #6610, which was suggested by the LWG:

> By using this website or other infrastructure provided by the OpenStreetMap Foundation, you agree to the [OpenStreetMap Terms of Use](https://osmfoundation.org/wiki/Terms_of_Use).

To avoid making the banner too large, especially on mobile devices, we could make the second paragraph smaller (like it used to be, I think) and append the new text to that paragraph.

[...]

Issue #6610, opened 2025-12-10
Prompt older users to acknowledge terms of use


1ec5:
The LWG has determined that users who joined before we introduced the Terms of Use should be actively prompted to review and accept the terms. #5706 added a setting to indicate ToU acceptance, but it’s hidden away on the Account Settings screen where relatively few users will see it. As I understand it, the suggestion is to also present these users with a toast or popup pointing to https://www.openstreetmap.org/account/terms after they successfully log in. The precise design doesn’t matter from a compliance standpoint; it could just as well take over the full page if it would be easier to implement. This would only affect older users, as most users were already required to accept the terms as part of signing up.

The LWG is aware that there isn’t an airtight enforcement mechanism. Many users only access OSM through the REST API, some of them anonymous, yet we also want to consider them subject to the terms. Instead, this would be a best effort to put older users on par with newer ones, so that they can’t claim to have been uninformed about the change in terms. This feature could also be useful in the future in case OSMF needs to make any substantive changes to the privacy policy.

simonpoole:
Just a general note on this: if it had ever just been a question of the technical implementation I would have done that many years ago. The issue is that the OSMF board needs to communicate this to the OSM contributors and it  has never been willing to actually do that.

PR #6804, opened 2026-02-12
Migrate `/export/embed.html` from leaflet to maplibre

CommanderStorm:
### Description

Migrate /export/embed.html from leaflet to maplibre
Part of the following mega-issue:
- https://github.com/openstreetmap/openstreetmap-website/issues/6234

### How has this been tested?

Manually, running the code
- adding markers
- setting different styles
- setting different light/dark preferences

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -387,12 +387,14 @@ nav.primary, nav.secondary {
   }
 }
 
+.maplibregl-ctrl.maplibregl-ctrl-attrib,
 .maplibregl-popup-content,
 .leaflet-popup-content-wrapper, .leaflet-popup-tip,
 .leaflet-control-attribution, .leaflet-control-scale-line {
   @extend .bg-body, .text-body;
 }
 
+.maplibregl-ctrl.maplibregl-ctrl-attrib,
 .leaflet-control-attribution, .leaflet-control-scale-line {
   @extend .bg-opacity-75;
   text-shadow: none !important;
@@ -407,6 +409,7 @@ nav.primary, nav.secondary {
 }
 
 @include color-mode(dark) {
+  .maplibregl-ctrl.maplibregl-ctrl-attrib a,
   .leaflet-container .leaflet-control-attribution a {
     color: var(--bs-link-color);
   }


app/assets/javascripts/maplibre/attribution.js
```diff
@@ -0,0 +1,123 @@
+OSM.MapLibre.AttributionControl = class extends maplibregl.AttributionControl {
```
CommanderStorm, line 1: by default, maplibre does have a hidable attribution on small mobile viewports once the user interacts with the map.
This is there to have more of the map visible on viewports that are already small.

Is this what is wanted or should I stick with the no changes ethos and change `compact:false` (or more rather remove the extends `maplibregl.AttributionControl`)

app/assets/javascripts/maplibre/attribution.js
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    attribHTML += ". ";
+
+    const termsLink = document.createElement("a");
+    termsLink.href = "https://wiki.osmfoundation.org/wiki/Terms_of_Use";
+    termsLink.target = "_blank";
+    termsLink.rel = "noopener noreferrer";
+    termsLink.textContent = OSM.i18n.t("javascripts.map.website_and_api_terms");
+    attribHTML += termsLink.outerHTML;
```
CommanderStorm, line 49: Credits and TOS are currently not used on the embed (it is sort of partial, for example no donate) on osm.org

Is this intentional or a bug?

app/assets/stylesheets/embed.scss
```diff
@@ -1,7 +1,11 @@
 /*
- *= require leaflet/dist/leaflet
+ *= require maplibre-gl-all
```
CommanderStorm: Previously, this had different styling from the non-embedded case.

Should I stick to:
- the old button design?
- the default maplibre design?
- the design of the buttons on the non-embedded case?

app/assets/stylesheets/embed.scss
```diff
@@ -1,7 +1,11 @@
...
  */
 
+@use "common" with (
+  $color-mode-type: media-query
+);
```
CommanderStorm: This imports a few styles, among them bootstrap, bootstrap icons, ...

Most of these are not nessesary.
Should I:
- stick with the previous approach and duplicate the styles that are needed here to get the exact old look
- accept that importing a few KB of styles is likely not that big of a deal

app/assets/javascripts/embed.js.erb
```diff
@@ -18,61 +22,41 @@ OSM.i18n.enableFallback = true;
...
-
-  const map = L.map("map");
-  map.attributionControl.setPrefix("");
-  map.removeControl(map.attributionControl);
-
   const isDarkTheme = args.get("theme") === "dark" || (args.get("theme") !== "light" && window.matchMedia("(prefers-color-scheme: dark)").matches);
   const layers = <%= MapLayers::embed_definitions("config/layers.yml").to_json %>;
+  <% if Settings.key?(:tile_cdn_url) %>
+    layers.mapnik.style.sources.raster-tiles-mapnik.tiles = [<%= Settings.tile_cdn_url.to_json %>];
+  <% end %>
```
CommanderStorm, line 28: I have no clue if this works as intened.

osm.org does not seem to set this variable.
Is it still needed?
What query placeholders should I expect here? (Do I need to `.replace(..)` some of them for compatability?)

app/assets/javascripts/embed.js.erb
```diff
@@ -18,61 +22,41 @@ OSM.i18n.enableFallback = true;
...
+  map.addControl(new OSM.MapLibre.CombinedControlGroup([navigationControl]), position);
 
-    this._container.innerHTML =
-      this._container.innerHTML
-        .replace("{x}", this._map.getCenter().lat)
-        .replace("{y}", this._map.getCenter().lng)
-        .replace("{z}", this._map.getZoom());
+  if (args.has("marker")) {
+    const markerCoords = args.get("marker").split(",");
+    new OSM.MapLibre.Marker({color: "var(--marker-blue)"})
```
CommanderStorm: usually, the markers on the site are red by default.

Here it is blue.
Intentional?

app/assets/javascripts/embed.js.erb
```diff
@@ -18,61 +22,41 @@ OSM.i18n.enableFallback = true;
...
+    credit: layerConfig.credit,
+    includeReportLink: true
+  });
+  map.addControl(attribution, 'bottom-right');
 
-  _update: function () {
-    L.Control.Attribution.prototype._update.call(this);
+  const position = document.documentElement.dir === "rtl" ? "top-left" : "top-right";
+  const navigationControl = new OSM.MapLibre.NavigationControl();
+  map.addControl(new OSM.MapLibre.CombinedControlGroup([navigationControl]), position);
```
CommanderStorm: Is this RTL handling correct? (previously there was no RTL handling)
- attribution always on the right
- navigation on right for ltr and left for rtl

I am not an RTL expert, since germany is firmly LTR

app/assets/javascripts/maplibre/attribution.js
```diff
@@ -0,0 +1,123 @@
...
+    // Add report link if enabled
+    if (this._includeReportLink) {
+      const reportLink = document.createElement("a");
+      reportLink.href = "/fixthemap";
+      reportLink.target = "_blank";
+      reportLink.rel = "noopener noreferrer";
+      reportLink.className = "maplibregl-ctrl-attrib-report-link";
+      reportLink.textContent = OSM.i18n.t("javascripts.embed.report_problem");
+      attribHTML += reportLink.outerHTML + " | ";
+    }
```
CommanderStorm: should `/fixthemap` be visible during style loading or only afterwards?

On slow internet, this might be a few seconds..

app/assets/javascripts/maplibre/attribution.js
```diff
@@ -0,0 +1,123 @@
...
+    const copyrightLink = document.createElement("a");
+    copyrightLink.href = "/copyright";
+    copyrightLink.textContent = OSM.i18n.t("javascripts.map.openstreetmap_contributors");
+
+    attribHTML += OSM.i18n.t("javascripts.map.copyright_text", {
+      copyright_link: copyrightLink.outerHTML
+    });
+
+    if (this._credit) {
+      attribHTML += this._credit.donate ? " ♥️ " : ". ";
```
CommanderStorm, line 38: Why is there a dot?
Does this make sense (in german and english I don't think this does)?
Is this intentional?

I am asking because the separator for fixthemap is `|`

app/assets/javascripts/maplibre/attribution.js
1ec5: We should make sure the attribution follows the spirit and letter of the [attribution guidelines](https://www.osmfoundation.org/wiki/Licence/Attribution_Guidelines) and [Copyright page](https://www.openstreetmap.org/copyright). My understanding is that the MapLibre default does adhere to the guidelines following maplibre/maplibre-gl-js#795, but definitely double-check that there are no situations (like maps above a certain size) where this is not the case.

app/assets/javascripts/maplibre/attribution.js
1ec5: It’s an edge case, but I guess we wouldn’t want people prematurely leaving notes or whatever if the styles or tiles don’t load for some reason.

app/assets/javascripts/maplibre/attribution.js
1ec5: Same reason as in https://github.com/openstreetmap/openstreetmap-website/pull/6601#discussion_r2604963172, probably premature optimization. If the attribution never appears without the trailing period, then we should fold the period into the attribution strings. Otherwise, a different separator like a middle dot, bullet, or vertical bar might be more appropriate than a period anyways. (It’s weird to see a period here but not at the end of the line.) But this should probably be a separate PR.

app/assets/javascripts/maplibre/attribution.js
1ec5: I think this is pretty much the reason the Copyright page ends up repeating a lot of the About content. How well would the extra links fit in the default 425-pixel-wide iframe?

app/assets/javascripts/embed.js.erb
1ec5: Does this actually flip the direction if your browser and/or site preferences specify a right-to-left language like Arabic or Hebrew? In cursory testing, I’m finding that the embed _code_ respects the site preference as far as the hard-coded “View Larger Map” link is concerned, but the iframe within it only respects the browser preference. And the page never sets a `lang` or `dir` tag or localizes the “OpenStreetMap Embedded” title. Only the attribution string gets localized through JavaScript.

app/assets/javascripts/maplibre/attribution.js
CommanderStorm: > You may use a mechanism to fade/collapse the attribution under certain conditions:
> - automatically on map interaction such as panning, clicking, or zoomin
>
> If the attribution has been collapsed, the user must still be able to find the licence information if they look for it, for example from an ‘(i)’ button in the corner of the map or an ‘About’ option in a menu.

Yes, our default obviously complies. It is more a question of what option you want.

app/assets/javascripts/maplibre/attribution.js
CommanderStorm: They fit alright. Wraps to a second line on some viewports, but since the default is to hide on-interaction that also works.

Since we are already touching this area, this would be the time to rethink what is intentional and what is a bug.

app/assets/javascripts/maplibre/attribution.js
CommanderStorm: will move this to a future PR since the ettribution is currently a bit hard to change (because under leaflet it is in a few different places)

app/assets/javascripts/embed.js.erb
hlfan: I think it's just to have the option. I wouldn't worry much about the format as there hasn't been any spec since its introduction in 685561687022262a1af5fc09c5617757d8529f3c.

app/assets/stylesheets/embed.scss
hlfan: I think the idea of this page is to use the default design. If the map library changes, the styling can too. I wouldn't even alter the default navigation control icons.

app/assets/javascripts/embed.js.erb
hlfan: It's the default Leaflet marker color. It was just replaced to consistently use vector images.

app/assets/javascripts/maplibre/attribution.js
1ec5: At the embed’s default size of 425×350, the ℹ️ will appear and automatically collapse by default. This will be a change from the existing Leaflet implementation. Personally, I’d be OK with that because we’d still comply with the guidelines by expanding the control by default. However, we also rely on the same control to show the equally important link to /fixthemap, so I’d lean towards expanding the control persistently. We can always revisit this choice if there’s a compelling reason for the ℹ️.

app/assets/javascripts/maplibre/attribution.js
1ec5: If I’m not mistaken, with the expanded version of the attribution control, if the attribution is too long to fit on a single line, it’ll truncate (…) by default until you hover over it, at which point it’ll wrap onto multiple lines. That’s reasonable behavior and will probably only affect a small number of locales.

Our attribution guidelines only require attribution. My current assumption is that we don’t need to link to the Terms of Use here. Someone viewing the embed on a third-party site is only minimally using the OSM site and tile server and not using the API at all. As soon as they do anything more substantial by visiting the main site, they’ll be covered by #6610, #6791, or any existing notice about the terms. The LWG hasn’t communicated to me that we need to extend this to the embed.

I’d also hold off on introducing a ❤️ Donate link into third-party sites. It would be a power move for sure. We can always introduce it later if we change our minds, for example to make the Copyright page less redundant. I wouldn’t want that to hold up the MapLibre migration.

app/assets/javascripts/maplibre/attribution.js
CommanderStorm: Makes sense, will rework this accordingly

app/assets/javascripts/maplibre/attribution.js
Robot8A: Third-party users are using OSMF resources when they use the embed, so why hide it? :wink:
I don't know how well used it is, given that it's difficult, if not impossible to find this embed function from the Export menu on the website, although a quick GitHub search tells me that quite a few repos use it (https://github.com/search?q=openstreetmap.org%2Fexport%2Fembed&type=code&ref=advsearch).

On the main website itself, I can understand hiding it if the new donate button on the top bar works well, which I'd say is still too early to measure.

CommanderStorm:
Sorry for the long wait.
I am currently getting figuratively killed by my master thesis.
I worked in all the feedback that I collected above on how this SHOULD work and this is what I came up with:

| regular case | no webgl | previous |
|--------|--------|--|
| image | image | image |


So the TL;DR on the changes compared to existing:
- [maplibre default button design](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2799134667), but the Marker is OSM-green 😉
- [showing the donate and T&C link](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2799134667) 
- [Zoom/Attribution are placed on right/left depending on iframe-RTL](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2796600764)
- [/fixthemap is not shown until the style has loaded (for raster not noticable, for vector depends on network speed and location, given the roundtrip for url -> style -> fonts/sprites)](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2796603791)

I did not change:
- The [separator inconsistency (`|` vs `.`)](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2796623546)
- [`Settings.tile_cdn_url` being unused](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2796594562)
- [the attribution does not hide, even if there is a lot of text and the viewport is small, because `/fixthemap`](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2796575928)

config/locales/en.yml
```diff
@@ -3573,6 +3573,10 @@ en:
       tooltip: "Legend"
       tooltip_disabled: "Legend not available for this layer"
     map:
+      map:
+        title: Map
+        marker:
+          title: Map marker
```
CommanderStorm, line 3574: both show up in `aria-labels`

app/assets/stylesheets/common.scss
```diff
@@ -387,12 +387,14 @@ nav.primary, nav.secondary {
   }
 }
 
+.maplibregl-ctrl.maplibregl-ctrl-attrib,
```
CommanderStorm, line 390: not precicly nesseary, but since I am adding the attrib controll here, integrating it so that it "works as expected" if used with the common style as well makes sense likely

app/assets/javascripts/osm_embed.js
```diff
@@ -0,0 +1,3 @@
+OSM = {
+  MapLibre: {}
+};
```
CommanderStorm, line 3: nesseary because I cannot have code before the first require, but the requires need this code to have been executed

app/assets/javascripts/maplibre/controls.js
```diff
@@ -30,9 +30,9 @@ OSM.MapLibre.CombinedControlGroup = class CombinedControlGroup {
           const icon = iconMap[type];
 
           if (icon) {
-            $(button)
-              .empty()
-              .append($("<i>").addClass(`maplibregl-ctrl-icon fs-5 bi bi-${icon}`));
+            const iconElement = document.createElement("i");
```
CommanderStorm, line 33: because otherwise the embed would need JQuery, which seems unnessary

app/assets/javascripts/maplibre/attribution.js
```diff
@@ -0,0 +1,128 @@
...
+    this._attribHTML = attribHTML;
+    this._updateCompact();
+
+    // Update report link href after initial render
+    if (this._includeReportLink) {
+      this._updateReportLink();
+    }
+  }
+
+  _buildCreditHtml(credit) {
```
hlfan, line 64: This function seems to be unavailable for third-party layers?

config/layers.yml
```diff
@@ -107,6 +107,7 @@
   layerId: "shortbread"
   nameId: "shortbread"
   wiki: "https://wiki.openstreetmap.org/wiki/Shortbread"
+  canEmbed: true
```
hlfan, line 110: Shouldn't the Maptiler OMT layer also be embeddable now?

app/assets/javascripts/embed.js.erb
```diff
@@ -18,61 +21,41 @@ OSM.i18n.enableFallback = true;
...
+  map.addControl(new OSM.MapLibre.CombinedControlGroup([navigationControl]), `top-${position}`);
 
-    this._container.innerHTML =
-      this._container.innerHTML
-        .replace("{x}", this._map.getCenter().lat)
-        .replace("{y}", this._map.getCenter().lng)
-        .replace("{z}", this._map.getZoom());
+  if (args.has("marker")) {
+    const markerCoords = args.get("marker").split(",").map(parseFloat);
+    new maplibregl.Marker({ color: "#7ebc6f" })
```
hlfan: We already include the "agreed-upon" marker?
```suggestion
    new OSM.MapLibre.Marker({ color: "#9cef11" })
```

app/assets/javascripts/maplibre/attribution.js
CommanderStorm: third party layers?
Yes, it is only used in this class, since giving credit is very attribution specific

config/layers.yml
CommanderStorm: not without more work since the API key that MapTiler gave won't work.

app/assets/javascripts/embed.js.erb
CommanderStorm: I left that part out because I understood from your comments ([design](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2796583839), [colors](https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2796595865)) that the page should follow the default MapLibre design.

To make sure I implement this as intended, which parts should use the Leaflet, OSM design, or MapLibre styling?

Here are the options for clairity:
- All maplibre with OSM-green marker color
- All maplibre with leaflet default marker color
- Marker the leaflet marker with default leaflet color and buttons the MapLibre one
- Full OSM style, incl. how the buttons look

I'm happy to adjust things either way - just want to communicate the "SHOULD" state in terms of direction to avoid unnecessary churn for both of us 😉

app/assets/javascripts/maplibre/attribution.js
hlfan: So the exceptions in the recursion for thunderforest's, hot, or CyclOSM layers are intended?

app/assets/javascripts/embed.js.erb
hlfan: Ah, well I think the slippy map parts should be consistent across the site. The minimal alterations argument was more for the UI. If the attribution container consistently mirrors the Leaflet style, that's fine too. But I'd leave most design changes for when leaflet is fully migrated away.

I'd go for the combination of native maplibre nav controls together with the osm green accent and OSM.MapLibre attribution & marker.

CommanderStorm: Ok, is adjusted.

image

app/assets/javascripts/maplibre/attribution.js
```diff
@@ -0,0 +1,128 @@
+OSM.MapLibre.AttributionControl = class extends maplibregl.AttributionControl {
+  constructor(options = {}) {
+    const { includeReportLink, credit, ...maplibreOptions } = options;
+    if (includeReportLink) {
+      maplibreOptions.compact = false;
+    }
+    super(maplibreOptions);
```
hlfan: Let's shorten this
```suggestion
  constructor({ includeReportLink, credit, ...options } = {}) {
    if (includeReportLink) {
      options.compact = false;
    }
    super(options);
```

app/assets/javascripts/embed.js.erb
```diff
@@ -18,61 +21,41 @@ OSM.i18n.enableFallback = true;
...
+    container: "map",
+    style,
+    attributionControl: false,
+    allowRotation: layerConfig.isVectorStyle,
+    zoomSnap: layerConfig.isVectorStyle ? 0 : 1
+  });
 
-    map.on("moveend", this._update, this);
+  const bbox = (args.get("bbox") || "-180,-90,180,90").split(",");
+  map.fitBounds([[bbox[0], bbox[1]], [bbox[2], bbox[3]]], { animate: false });
```
hlfan, line 43: For raster layers, one has to zoom out to get to the equivalent 1-to-1-pixel scaling.

app/assets/javascripts/maplibre/attribution.js
```diff
@@ -0,0 +1,128 @@
...
+        link.className = "donate-attr";
+      } else {
+        link.target = "_blank";
+        link.rel = "noopener noreferrer";
+      }
+
+      return link.outerHTML;
+    }
+
+    return text;
```
hlfan: I think this would be more readable with an early return.

app/assets/javascripts/embed.js.erb
CommanderStorm: Not sure I am following.

For raster layers, the requested bbox bounds are the same..
The bbox does not change...

Yes, I still have not gotten around to implemeting the touchpad `snapZoom` if that is what you are refering too. Will have to postpone implementing this by about a month.
Not relevant to the code you are pointing to though..

app/assets/javascripts/maplibre/attribution.js
CommanderStorm: ACK, adjusted both leaflet and maplibre 👍🏻

app/assets/javascripts/maplibre/attribution.js
CommanderStorm: fair

app/assets/javascripts/embed.js.erb
hlfan: leaflet automatically snaps to the highest integer zoom level that fits the bbox, maplibre usually sets a fractional zoom.

CommanderStorm: Yes, that is what I meant with snapZoom not being entirely implemented.

I can add a quick "hack" if clause on the zoom here and correct it or we handle this via zoomSnap upstream.
The first is something I can do now, the second is something which I still have to implement when I am a little less busy..
I think that users likely won't notice either way..

Being transparent with timelines:
I am willing to do that work upstream but cannot promise that I will get to it before the 30.5. (when I have to submit my thesis).

hlfan: I see. A zoom level flooring conditional in OSM.MapLibre.Map.fitBounds is clean enough for now and presumably easier to remove with the upstream fix in Q3.

app/assets/javascripts/maplibre/attribution.js
```diff
@@ -0,0 +1,126 @@
...
+
+    // Update report link href after initial render
+    if (this._includeReportLink) {
+      this._updateReportLink();
+    }
+  }
+
+  _buildCreditHtml(credit) {
+    const children = {};
+    if (credit.children) {
```
hlfan: Is this condition actually guarding against anything the loop cares about?

app/assets/javascripts/embed.js.erb
CommanderStorm: Ok, I implemented it -> https://github.com/maplibre/maplibre-gl-js/pull/7332
Will be released in our next, sheduled release.

tomhughes:
What exactly is it that we're waiting for from upstream?

hlfan:
> What exactly is it that we're waiting for from upstream?

new maplibre release - https://github.com/openstreetmap/openstreetmap-website/pull/6804#discussion_r2995085729

app/assets/javascripts/maplibre/attribution.js
CommanderStorm: Yes, please see `leafletOsmId: "Mapnik"` -> no `credit.children`

hlfan: And in which environment does a for..in loop over an undefined collection cause errors?

CommanderStorm: Ah, did not know that for..in does behave differently to for..of

I would have expected that if one errors the other also should.

CommanderStorm:
image

Bumped to ml@5.17 -> ml@5.22, which (only osm relevant features) among others comes with
- `snapZoom` for all programatic controls (`fitBounds`, `fitScreenCoordinates`, `jumpTo`, `easeTo`, and flyTo`)
- a nice rendering performance upgrade for symbols on the vector layer
- better ETAG support,
- `prefers-reduced-motion` makes the pulsing GeoLocation anmiation not appear
- click events on markers

I don't know of any (non-intentional, discussed) diff, this should be ready to ship

app/assets/javascripts/maplibre/attribution.js
hlfan: So?

app/assets/javascripts/maplibre/attribution.js
hlfan: Did ya?

CommanderStorm: Grr, I hate this force-pushing over my own changes.

CommanderStorm: force pushed again, now with this change.

I really hate this rebasing and force pushing, this is such an confusing and error-prown workflow as can be seen here..

hlfan:
Thanks! I've fixed an inverted condition that vanished in the last push:

https://github.com/openstreetmap/openstreetmap-website/blob/0f246698a75d9b26fe34b9ed854d33ba2e1f03b0/app/assets/javascripts/maplibre/attribution.js#L72

CommanderStorm:
Thanks ❤️

PR #6810, opened 2026-02-12
Fixes for dev container environment under macOS

pablobm:
Following my tribulations at https://github.com/openstreetmap/openstreetmap-website/pull/6713, this helped me making the dev container work on both macOS/arm64 and Linux/amd64.

The main problem appeared to be with the way I set up a volume to cache the packages downloaded by Bundler. At the time this seemed like a good idea and worked. In fact I do think otherwise it would install the packages every time? Either I misinterpreted something or something has changed. Whatever. This works.

Another issue I ran into had to do with Postgres versions. The Docker image used by the dev container setup includes tools such as `pg_dump` which is used to generate `db/structure.sql`... but these tools expect Postgres v15. Since `compose.yaml` declares `postgres:16`, there's an incompatibility. How this worked for me on Linux and not on macOS is beyond my understanding, but it fixes things.

Additionally I have included a section in the docs explaining how to reset the DB when this happens.

### DIFF
--- a/doc/DEVCONTAINER.md
+++ b/doc/DEVCONTAINER.md
@@ -148,3 +148,29 @@ bundle install
 ```
 
 This has been observed particularly when using RubyMine.
+
+### Issues connecting to the database
+
+For example:
+
+```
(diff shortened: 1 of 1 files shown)


tomhughes:
Looks good to me, thanks.

PR #6811, opened 2026-02-13
Remove jQuery throttle-debounce dependency

### LINKED ISSUE #7142: Remove jQuery
Is jQuery required in this project? The codebase is old so I'm sure it made sense for a long time, but I have the impression that now it's used for things already provided by browsers at very high level of compatibility.

Trying now, my browser loads 2.68 MB of uncompressed JS (617.35 kB compressed/transferred). Of that 285k (uncompressed) appears to be jQuery.

Dependencies that require JS:
- jquery-ujs (22kB uncompressed): can't remember exactly what this includes. One thing is support for `method="delete"`, etc which I think nowadays is implemented with Turbo, which we also load.
- ~jquery.throttle-debounce: throttling/debouncing is old hat in JS. There must be a different library we can use that doesn't require jQuery.~ (Removed at https://github.com/openstreetmap/openstreetmap-website/pull/6811).

Something potentially important that we'd be missing: delegated events. jQuery makes it easy to set events on elements that don't exist on the page yet, and these magically work when the element appears. Some refactoring might be required. Other, smaller libraries might help here.

If we were to decide to remove jQuery, I wouldn't expect it to happen all in one fell swoop, but rather incrementally: discouraging it from new code while slowly removing existing usage module by module.

hlfan:
iD already throttles its hash updates. Hooking into this allows dropping the vendored dependency, leaving only leaflet-related ones.


app/assets/javascripts/id.js
```diff
@@ -35,7 +35,7 @@ document.addEventListener("DOMContentLoaded", function () {
     }
 
     let hashChangedAutomatically = false;
-    id.map().on("move.embed", parent.$.throttle(250, function () {
+    window.history.replaceState = function () {
```
HolgerJeromin: You do not want to overwrite basic Browser APIs.
https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState

hlfan: I've now made it fire a custom event. Is that override acceptable?

HolgerJeromin:
Can we have instead a (throttled) event from iD on its parent like?
```ts
window.parent?.postMessage('id-hash-update', '*' /* This is no sensitive data, so ok to allow all targets */);
```

HolgerJeromin:
Now we have `postmessage` and hooking into `replaceState`. 
I wanted to suggest replacing `replaceState` hook with the postmessage. 
Yeah, that needs adjustment in iD (and not the id-embedder) probably.

tomhughes:
Does the second change actually make anything simpler? It certainly increases the amount of code...

hlfan:
> Does the second change actually make anything simpler?

It decreases the coupling between the parent and the iframe.
With it, the parent is fully in control of its listeners and doesn't have that amount of cross-context calls. Each side now owns its own responsibilities and sends events to the counterpart if it needs something.

HolgerJeromin:
My favorite solution would be here in this repo in one file `window.addEventListener("message", ...` and in the id repo `window.parent?.postMessage('id-hash-update', '*');`

hlfan:
Yeah that would be nice, or exposing iD's lodash utilities similarly to d3.

@tyrasd which option do you think would be more in-scope for the iD repo?

hlfan:
> Does the second change actually make anything simpler? It certainly increases the amount of code...

I've dropped this change for now.

app/assets/javascripts/id.js
```diff
@@ -2,6 +2,15 @@
...
 /* globals iD */
 
+(() => {
+  const originalReplaceState = window.history.replaceState;
+  window.history.replaceState = function (...args) {
+    const result = originalReplaceState.apply(this, args);
+    window.dispatchEvent(new CustomEvent("replaceHistoryState", { detail: args }));
+    return result;
+  };
+})();
```
pablobm: This makes me very uneasy. Is this overwrite common? Did @HolgerJeromin have an opinion on this? I saw the comment was resolved.

HolgerJeromin: I do not like this approach as you know, but it is not unusual in JS world.

hlfan:
> Yeah that would be nice, or exposing iD's lodash utilities similarly to d3.
> 
> which option do you think would be more in-scope for the iD repo?

Or @k-yle, maybe you have an idea?

k-yle:
> Or @k-yle, maybe you have an idea?

I would suggest adding [a new event](https://github.com/openstreetmap/iD/blob/c2c1d769da9becbde59fa0c38fcf07e734e304a1/modules/core/context.js#L26) to iD's `coreContext`, and emitting that event whenever we call `history.replaceState()`[^1].

Then you don't need to override `history.replaceState` in this repo, and you can remove jQuery.throttle since iD [already thottles `history.replaceState`](https://github.com/openstreetmap/iD/blob/c2c1d769da9becbde59fa0c38fcf07e734e304a1/modules/behavior/hash.js#L125) by 500ms.

[^1]: which is unfortunately copy-pasted many times in iD, that should probably be cleaning up at some point 🤷

HolgerJeromin:
> I would suggest adding [a new event](https://github.com/openstreetmap/iD/blob/c2c1d769da9becbde59fa0c38fcf07e734e304a1/modules/core/context.js#L26) to iD's coreContext, and emitting that event whenever we call history.replaceState()

I have no intern view what this event is, but custom events are not always walk through the iframe barrier, while `postMessage` do.

hlfan:
We are currently using `parent.` for that anyway.

hlfan:
> Does the second change actually make anything simpler? It certainly increases the amount of code...

I'll revisit this change in a separate pr to sandbox the iD frame as per the security report.

HolgerJeromin (approved):

pablobm:
@HolgerJeromin - Unfortunately it doesn't work: `Uncaught TypeError: can't access property "on", id.ui().hash is undefined`.

Here's an alternative implementation: https://github.com/openstreetmap/openstreetmap-website/compare/master...pablobm:remove-throttle-dep

hlfan:
> Unfortunately it doesn't work

It does with 2.41, which hasn't been released yet; hence, it stays as a draft for now.

> `Uncaught TypeError: can't access property "on", id.ui().hash is undefined`

That just shows `hash.on` hasn't yet been rebound to the event dispatcher.

> Here's an alternative implementation

That could break without notice if iD moves further with simplifying their dependencies.

app/assets/javascripts/id.js
```diff
@@ -39,14 +38,17 @@ document.addEventListener("DOMContentLoaded", function () {
       parent.postMessage({ type, data }, location.origin);
     }
 
-    id.map().on("move.embed", window.Cowboy.throttle(250, function () {
-      if (id.inIntro()) return;
-      const zoom = ~~id.map().zoom(),
-            center = id.map().center(),
-            llz = { lon: center[0], lat: center[1], zoom: zoom };
+    id.map().on("move.embed-init", function () {
+      id.map().on("move.embed-init", null);
```
pablobm: That's an odd line. I'm guessing it's to avoid multiple initialisation? But to me that would suggest that there's only one possible handler for this event, is that the case?

hlfan: `id.ui().hash` isn't available at the time this code runs, so I also made this listen lazily. `d3.dispatch` doesn't have a `once` listener method like jQuery or Leaflet.

pablobm: Can you only place one single handler on a given event then? In other frameworks, I would expect something like this:

```js
function eventHandler() {
   id.map().off("move.embed-init", eventHandler);
   // ...
}
id.map().on("move.embed-init", eventHandler);
```

Trying to look up `off` and it doesn't appear to exist, so I'm guessing that's the case. Probably the reason why `once` doesn't exist either!

How about a comment?

```suggestion
      // D3.js doesn't give us a better way to ensure
      // that this handler runs only once.
      id.map().on("move.embed-init", null);
```

hlfan: > Can you only place one single handler on a given event then?

I can place many on e.g. move, just with different names. That way I can tweak the embed-init _move_-listener without having to keep references to the callbacks.

tyrasd: yeah, d3's [event handling](https://d3js.org/d3-selection/events) is pretty bare-bones, so you do have to do some stuff the manual way. :sweat_smile: 

> `id.ui().hash` isn't available

Hmm, sorry I didn't think about this when adding the event. As far as I can see, there's nothing speaking against making it available right after `.init()` is finished. Let's change this (probably has to wait for the next release v2.42, though)

pablobm: Thank you for the explanations, team :heart: I didn't realise that `embed-init` was a label; I thought it was part of the event name. I couldn't find it anywhere in iD's or D3's codebase and assumed it was part of some other dependency. All clear now :+1:

tomhughes:
Looks good to me, thanks.

PR #6818, opened 2026-02-18
Fix traces going off screen for very long filenames

### LINKED ISSUE #6817: GPS page can go off the right on the screen if the name is long
### URL

https://www.openstreetmap.org/traces

When I visit the Traces, they seem to be able to shift off the screen quite significantly.

This is becauset there is no `<div class="table-responsive">` around the table and changing this

Image

Image

Because 

Image

CommanderStorm:
### Description

Resolves #6817 

diff looks a bit large, it is just indentation. see the disable whitespace option in the the files tab.

### How has this been tested?

Manually, in the UI

image

### DIFF
--- a/app/views/traces/_trace.html.erb
+++ b/app/views/traces/_trace.html.erb
@@ -12,7 +12,7 @@
   </td>
   <td>
     <ul class="list-inline mb-0">
-      <li class="list-inline-item">
+      <li class="text-break list-inline-item">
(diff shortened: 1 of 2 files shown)


hlfan:
Could the `table-responsive` class also be added directly onto the table element?

CommanderStorm:
No, this needs to be in the container above

image

PR #6820, opened 2026-02-18
Avoid zombie Hardware Upgrade Fund translations

### LINKED ISSUE #6819: Remove outdated translations from Translatewiki
In #6617 we added the `layouts.donate` translation key. It turns out that we had used this previously (https://github.com/openstreetmap/openstreetmap-website/pull/5169 and references), and Translatewiki has kept the old translations for this key. These old translations have now resurfaced.

The old translations contain an interpolation variable, which causes errors. We've worked around this issue for the last couple of weeks by repeatedly removing the outdated translations ( https://github.com/openstreetmap/openstreetmap-website/commit/a9c6d3b7bf91a2a5df650805d6175d115f6821aa , https://github.com/openstreetmap/openstreetmap-website/commit/ec7af603382db0067022b6fb853900aecc7087ea ) but we don't want to keep doing so indefinitely.

I would like to blank these translations on Translatewiki, to allow them to be re-translated in due course. However, Translatewiki won't accept my attempts at blanking these translations, and I don't feel like I can just "pick something" for all the affected languages.

So we have three ways forward:
* Keep deleting them every week, until eventually everything is translated
* Use a different key, one that hasn't been used before
* Arrange for a translatewiki moderator to remove the affected translations

@translatewiki are you able to help with the third option?

1ec5:
Renamed the string for the Donate navigation bar link introduced in #6617 to prevent ancient Hardware Upgrade Fund translations from reappearing via Translatewiki.net.

Fixes #6819.

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -62,7 +62,7 @@
         <%= link_to t("layouts.help"), help_path, :class => header_nav_link_class(help_path) %>
       </li>
       <li class="nav-item">
-        <%= link_to t("layouts.donate"), Settings.donation_url, :class => header_nav_link_class(Settings.donation_url), :target => :new %>
+        <%= link_to t("layouts.donate_link"), Settings.donation_url, :class => header_nav_link_class(Settings.donation_url), :target => :new %>
       </li>
       <li class="nav-item">
         <%= link_to t("layouts.about"), about_path, :class => header_nav_link_class(about_path) %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1859,7 +1859,7 @@ en:
     about: About
     copyright: Copyright
     communities: Communities
-    donate: Donate
+    donate_link: Donate
     learn_more: "Learn More"
     more: More
     header:


gravitystorm:
Thanks!

As the Danger report indicates, you need to only change en.yml, and then we wait for the translations to come through from Translatewiki. Could you update this PR accordingly please.

I'm also very interested to see how this will interact with the @translatewiki key-renaming-detection, which might just copy all the existing translations (both good and bad, potentially) over to the new key! Let's see.

Nikerabbit:
> I'm also very interested to see how this will interact with the @translatewiki key-renaming-detection, which might just copy all the existing translations (both good and bad, potentially) over to the new key! Let's see.

If I get to review the incoming changes (someone else might do it before me), I will instruct not to rename.

1ec5:
> As the Danger report indicates, you need to only change en.yml, and then we wait for the translations to come through from Translatewiki. Could you update this PR accordingly please.

OK, I pared it back.

hlfan:
> then we wait for the translations to come through from Translatewiki.

I've accelerated this a bit: https://translatewiki.net/w/i.php?title=Special:Log&type=move&dir=prev&offset=20260218

